PR #2479, opened 2017-12-08
Do not check library ownership in libexec

### LINKED ISSUE #2473: xccdf_org.ssgproject.content_rule_file_ownership_library_dirs failes because of /usr/libexec/ file.
#### Description of problem:

In "xccdf_org.ssgproject.content_rule_file_ownership_library_dirs",
library_files_uid_root failes because /usr/libexec/abrt-action-install-debuginfo-to-abrt-cache is not root:root. The rule should not check /usr/libexec/* files.

scap-security-guide-0.1.36-1.fc27.noarch

Fedora 27

#### Steps to Reproduce:

1. Install abrt-addon-ccpp on Fedora
2. enable xccdf_org.ssgproject.content_rule_file_ownership_library_dirs
3. run oscap check

#### Actual Results:

library files uid root  failed because of: /usr/libexec/abrt-action-install-debuginfo-to-abrt-cache 

#### Expected Results:

check goes well without fail.

redhatrises:
#### Description:

- Do not check library ownership in /usr/libexec since scripts reside in that directory.

#### Rationale:

- /usr/libexec is script directory that does not contain library files.
- Fixes #2473

### DIFF
--- a/shared/checks/oval/file_ownership_library_dirs.xml
+++ b/shared/checks/oval/file_ownership_library_dirs.xml
@@ -34,7 +34,7 @@
 
   <unix:file_object comment="library files" id="object_file_ownership_lib_files" version="1">
     <!-- Check that files within /lib, /lib64, /usr/lib, and /usr/lib64 directories belong to user with uid 0 (root) -->
-    <unix:path operation="pattern match">^\/lib(|64)|^\/usr\/lib(|64)</unix:path>
+    <unix:path operation="pattern match">^\/lib(|64)\/|^\/usr\/lib(|64)\/</unix:path>
     <unix:filename operation="pattern match">^.*$</unix:filename>
    <filter action="include">state_owner_libraries_not_root</filter>
   </unix:file_object>


mpreisler:
IMO a better fix would be to append a slash at the end of both the regex parts.

`/usr/lib64blablabla` should also not be checked.

redhatrises:
> IMO a better fix would be to append a slash at the end of both the regex parts.
> 
> `/usr/lib64blablabla` should also not be checked.

Good point. Updated PR.

mpreisler:
Dont want to be a nitpicker but NACK.

`^\/usr\/lib(\/|64\/)` is equivalent to ^\/usr\/lib(|64)\/`. Do not repeat yourself.

Furthermore, please fix it for the first regex part as well `/libblablabla` should also not be matched.

redhatrises:
> Dont want to be a nitpicker but NACK.

@mpreisler you nitpicker you ;)

> `^\/usr\/lib(\/|64\/)` is equivalent to `^\/usr\/lib(|64)\/`. Do not repeat yourself.

I originally did that, but it didn't work. But now it does.... huh.... maybe a case of the Fridays.

> Furthermore, please fix it for the first regex part as well `/libblablabla` should also not be matched.

Fixed.

mpreisler:
ACK

PR #2481, opened 2017-12-08
Fix title of DISA STIG profile in RHEL6 DS.

dahaic:
#### Description:

- Fixes DISA STIG profile in RHEL6 DataStream

#### Rationale:

- DISA STIG profile extends common profile, but title is completely different, thus needs `override` attribute.
- Fixes: RHBZ#1521081


mpreisler:
ACK

PR #2485, opened 2017-12-12
Add test for accounts_umask_etc_login_defs

yuumasato:
#### Description:

- Adds tests for correct and wrong UMASK for OSPP-RHEL7 profile.

#### Rationale:

- Test scenarios are useful.
  In this particular case, it demonstrated that bash remediation for `accounts_umask_etc_login_defs`  can't deal with `UMASK`, just `umask`.


dahaic:
@yuumasato so, can you also add scenario that use `umask`? And super harden? (UMASK=177)? Could be also useful to define what is actually expected in this case (as one of the goals of the test suite is also to set expectations).

yuumasato:
Oh, I have proposed this PR directly, I'm closing, this. I will open a new one from my fork.

PR #2488, opened 2017-12-12
Fix accounts_umask_etc_login_defs remediation

### LINKED ISSUE #2486: Bash remediation for accounts_umask_etc_login_defs can't handle umask in lower case
#### Description of problem:
Bash remediation for `accounts_umask_etc_login_defs` adds another line defining umask with `umask` when `/etc/login.defs` already contains value defined by `UMASK`, the remediation is case sensitive.

0.1.36

#### Steps to Reproduce:

1. Remediate Rule `accounts_umask_etc_login_defs` for OSPP-RHEL7 Profile with `UMASK 027` in `/etc/login.defs`

#### Expected Results:
Remediation should replace line with umask value, not add another one.

redhatrises:
#### Description:

- Fix accounts_umask_etc_login_defs remediation

#### Rationale:

- Current remediation is incorrect and looks for the wrong case in login.defs
- Fixes #2486

### DIFF
--- a/shared/fixes/bash/accounts_umask_etc_login_defs.sh
+++ b/shared/fixes/bash/accounts_umask_etc_login_defs.sh
@@ -2,8 +2,4 @@
 . /usr/share/scap-security-guide/remediation_functions
 populate var_accounts_user_umask
 
-grep -q umask /etc/login.defs && \
-  sed -i "s/umask.*/umask $var_accounts_user_umask/g" /etc/login.defs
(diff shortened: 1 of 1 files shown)


yuumasato:
I wonder if we should also support `umask`. I see lots of guides mentioning it in lower case.

yuumasato:
Just tested, and it errors when all lower case:
```
configuration error - unknown item 'umask' (notify administrator)
```

Ack, thanks.

PR #2491, opened 2017-12-13
Check state of openssh-server package when sshd_required is unset

yuumasato:
#### Description:

- Make definitions `sshd_required_or_unset` and `sshd_not_required_or_unset` do only one job, which is to check value of `sshd_required`.
- Explicitly check state of openssh-server package in definitions, so that when `sshd_required` is unset we check for state of openssh-server package, thus checking configuration of system if it is installed, and just passing if not installed
- Small fix in test scenario for `sshd_use_approved_ciphers`.

#### Rationale:

- Definitions `sshd_required_or_unset` and `sshd_not_required_or_unset` cannot be generic and know whether the system is ok or not when `sshd_required` is unset, just checking whether `openssh-server` is installed or not is not enough, both states are valid.


yuumasato:
I have fixed test scenarios for SSH.
Scenarios for `firewalld_sshd_port_enabled` are reporting error because there is no remediation for it, see https://github.com/OpenSCAP/scap-security-guide/pull/2328. They are valid and should be ok to have them there.

mpreisler:
ACK

PR #2497, opened 2017-12-19
Better bash remediations tests.

matejak:
The ultimate goal of this PR is to test that bash remediations can be applied individually, and they will still succeed.
~Depends on https://github.com/OpenSCAP/scap-security-guide/pull/2468~

The PR:
- Implements checks for remediations done in separate steps.
- Adds checks for separate-step bash remediations.
- Reduces code duplication.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -21,7 +21,7 @@ accessible via ssh on the libvirt domain and have Ansible installed on the host 
 
 ### Domain preparation (example CentOS)
 1. Install domain, using `kickstarts/rhel_centos_7.cfg`
-   Typically, you supply a boot option `inst.ks=https://raw.githubusercontent.com/matejak/scap-security-guide/master/tests/kickstarts/rhel_centos_7.cfg`.
+   Typically, you supply a boot option `inst.ks=https://raw.githubusercontent.com/OpenSCAP/scap-security-guide/master/tests/kickstarts/rhel_centos_7.cfg`.
 1. Import ssh key to the machine and make sure that you can use them to log in as the `root` superuser.
 1. Setup repo, so machine can install and uninstall packages.
 
--- a/tests/README.md
+++ b/tests/README.md
@@ -11,7 +11,7 @@ remediation works.
 For the Test Suite to work, you need to have libvirt domains prepared for testing.
 SSG Test Suite currently does not provide automated provisioning of domains.
 You can use kickstart usable for Red Hat Enterprise Linux 7 (and of course CentOS 7) in
-```kickstarts``` directory, which installs machine to be capable of building openscap
+`kickstarts` directory, which installs machine to be capable of building openscap
 and builds and installs the latest upstream code.
 
 If you want to use your own domain, make sure `openscap-1.2.15` and `qemu-guest-agent`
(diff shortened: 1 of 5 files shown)


tests/README.md
```diff
@@ -11,38 +11,45 @@ remediation works.
...
 If you want to use your own domain, make sure `openscap-1.2.15` and `qemu-quest-agent`
 are installed there, `root` is accessible via ssh and that `yum` command is able
 to install packages.
 
 ### Domain preparation (example CentOS)
-1. Install domain, using kickstarts/rhel_centos_7.cfg
-1. Import ssh key to the machine
-1. Setup repo, so machine can install and uninstall packages
+1. Install domain, using `kickstarts/rhel_centos_7.cfg`
+   Typically, you supply a boot option `inst.ks=https://raw.githubusercontent.com/matejak/scap-security-guide/master/tests/kickstarts/rhel_centos_7.cfg`.
```
dahaic: Please refer to core project, not your fork :)

tests/test_suite.py
```diff
@@ -55,11 +54,11 @@
...
-common_parser.add_argument("--ansible",
-                           dest="ansible",
-                           action="store_true",
-                           default=False,
-                           help="Use ansible instead of bash remediations")
+common_parser.add_argument("--remediate-using",
+                           dest="remediate_using",
+                           default="oscap",
+                           choices=ssg_test_suite.oscap.REMEDIATION_RULE_RUNNERS.keys(),
+                           help="What type of remediations to use.")
```
dahaic: It would help to explain what each choice means, at least what is a difference between `oscap` and `bash`

tests/analyze_test.py
```diff
@@ -0,0 +1,190 @@
...
+import argparse
+
+NO_STATE_KEYS = ["name",]
+
+
+def get_table(records):
+    columns = tuple(sorted(get_columns(records)))
+    counts = []
+    for record in records:
+        counts.append((record.get('name', ''),) + tuple(get_values(record, columns)))
```
mpreisler: needs some comments, this is pretty hard to read

tests/test-analyze_test.py
```diff
@@ -0,0 +1,159 @@
+import json
+import os
+
+import pytest
+
+import analyze_test as analyze
+
+
+def transform_table(table):
```
mpreisler: seems like a waste to have this delegate defined on 3 lines

mpreisler:
Didn't have time to do a full review but one common theme here is that the code has no comments even in areas where I would expect some. There is some code bloat where you define an analyze module that you only use once. Perhaps define it in the test itself and don't have multiple modules for it? If you want to have a module for it it needs docs and justification, is it going to be used anywhere else? Same with some of the delegates, python has lambdas so you don't have to define a function that just delegates to another function binding in an arg.

So tl;dr:
1) provide docs, when you manipulate tables either use an API that makes it obvious or show how you use lists of tuples of tuples as a table
2) make the code shorter, avoid delegates, avoid modules for things only used once
3) fix pep8 issues
4) provide a PR description so that it's easy for a reviewer to test this

matejak:
I will make the code more understandable and comment it where appropriate. I don't get the comment about the `analyze` module - do you suggest to move it under the `ssg_test_suite` package?

To put this in a context, this PR is not a candidate for mergning. Although the code it is supposed to work, the essential pre-merge polish will be applied after @dahaic confirms that the functionality meets expectations. However, all comments are welcomed.

mpreisler:
> To put this in a context, this PR is not a candidate for mergning

This absolutely has to be in the title then. It was not obvious to me that this is not for merging.

mpreisler:
> about the analyze module - do you suggest to move it under the ssg_test_suite package?

Yeah, if it's a module that is only used at one place and the place where it's used is a small delegate executable I think we can just have it all there. Would be simpler and more readable. I could of course be missing something :)

mpreisler:
> A separate script analyze_test has been introduced together with its tests test-analyze_test.py. It is supposed to display and compare results of earlier tests by reading those .json files. The ultimate goal is to make discovery of various irregularities or regressions easy and quickly provide the user with essential data.

I really wish that was a separate PR.

dahaic:
sudo ./test_suite.py rule --hypervisor 'qemu:///system' --domain ssg-test-suite --datastream ../build/ssg-rhel7-ds.xml --benchmark xccdf_org.ssgproject.content_benchmark_RHEL-7 --remediate-using bash sshd

matejak:
I have rebased the branch against recent master and took out the script for analysis.
The requested missing functionality was added, so if it passes, I will focus on formal aspects of the PR to prepare it for merging.

matejak:
The PR is now considered a merge candidate. I have not fixed all line-length issues though. Although I agree that most lines should fit under 80 chars, I think that enforcing the limit for every line is counter-productive. However, if the general consensus is that we want to keep this restriction, I will fix the remaining lines.

Aside from implementing some new functionality, the existing code was heavily reorganized, so the diff doesn't make so much sense. I have updated the description with new highlights.

dahaic:
Great job Matej!!! Test Suite is in great shape now.

Issue #2495, opened 2017-12-15
aide remediations - aide_scan_notification needs update for regular grep syntax


nathanpeters008:
#### Description of problem:
The remediation script for aide_scan_notification uses the following regular expression to determine if the remediation needs to be applied: `'^.*\/usr\/sbin\/aide\s*\-\-check.*\|.*\/bin\/mail\s*-s\s*".*"\s*root@.*$'`

This results in the following command being run in the fix script: `grep -qR '^.*\/usr\/sbin\/aide\s*\-\-check.*\|.*\/bin\/mail\s*-s\s*".*"\s*root@.*$' /etc/crontab /etc/cron.d /etc/cron.daily /etc/cron.weekly /etc/cron.monthly`

However, since the grep command line does not include the -e flag, it uses regular grep syntax, which interprets the `\|` as the OR operator, instead of the literal pipe, causing the check to pass even if everything after the `|` in the regular expression is missing.

#### SCAP Security Guide Version:
0.1.36-1.el7.centos

#### Operating System Version:
CentOS 7.4.1708

#### Steps to Reproduce:

1. Run remediations against a system using a profile that includes both aide_periodic_cron_checking and aide_scan_notification (such as NIST 800-171)

#### Actual Results:
The remediation for aide_periodic_cron_checking is applied, but the remediation for aide_scan_notification is not

#### Expected Results:
Both remediations are applied

#### Addition Information/Debugging Steps:
The remediation for aide_periodic_cron_checking adds the following line to /etc/crontab:
`0 5 * * * /usr/sbin/aide  --check`

When the aide_scan_notification remediation is run, because the pipe is escaped, it is literally interpreted as match if you find either of these values `'0 5 * * * /usr/sbin/aide  --check' OR '/bin/mail -s "$(hostname) - AIDE Integrity Check" root@localhost'`

This causes the script to not bother applying the remediation because the first half of the OR expression already exists.

The easiest fix is to just remove the `\` character before any pipe in the regular expression since basic syntax already treats the pipe as a literal and there is no need to escape it.

yuumasato:
Hello @nathanpeters008, thank you for the detailed report.

Would you like to implement and propose the fix?
This will also fix https://github.com/OpenSCAP/scap-security-guide/issues/2112

nathanpeters008:
I have included this in https://github.com/OpenSCAP/scap-security-guide/pull/2500/

Issue #2112, opened 2017-06-23
rule_aide_scan_notification fails to remediate


yuumasato:
I could not find out why this regex 
```
^.*\/usr\/sbin\/aide\s*\-\-check.*\|.*\/bin\/mail\s*-s\s*".*"\s*root@.*$ 
```
from [aide_scan_notification.sh](https://github.com/OpenSCAP/scap-security-guide/blob/master/shared/templates/static/bash/aide_scan_notification.sh#L9)

matches 
```
05 4 * * * root /usr/sbin/aide --check
```
which is added by [aide_periodic_cron_checking.sh](https://github.com/OpenSCAP/scap-security-guide/blob/master/shared/templates/static/bash/aide_periodic_cron_checking.sh#L3) remediation.

The string should not match and remediation action should happen.

Also, remediation is missing username in [line10](https://github.com/OpenSCAP/scap-security-guide/blob/master/shared/templates/static/bash/aide_periodic_cron_checking.sh#L3)

PR #2500, opened 2017-12-20
Updated aide_scan_notification remediation to run cron job as root

### LINKED ISSUE #2496: aide_periodic_cron_checking and aide_scan_notification are not compatible
#### Description of problem:
A system that has the aide_scan_notification remediation applied should satisfy both the aide_scan_notification check and the aide_periodic_cron_checking check; however, this is not the case due to some slight differences in syntax between the checks.

Because of this error in syntax, the /etc/crontab file ends up with 2 aide cron jobs

0.1.36-1.el7.centos

CentOS 7.4.1708

#### Steps to Reproduce:

1. Perform remediation on a system and include both the aide_periodic_cron_checking and aide_scan_notification remediations

#### Actual Results:
The following 2 lines are added to the /etc/crontab, causing the aide scan to be run twice every day
05 4 * * * root /usr/sbin/aide --check
0 5 * * * /usr/sbin/aide  --check | /bin/mail -s "$(hostname) - AIDE Integrity Check" root@localhost

#### Expected Results:
Only the following line is added, which should satisfy both checks:
0 5 * * * /usr/sbin/aide  --check | /bin/mail -s "$(hostname) - AIDE Integrity Check" root@localhost

The check for aide_periodic_cron_checking expects the word 'root' in /etc/crontab between the time and the command, so the line added by aide_scan_notification will not satisfy it because it does not contain the word 'root' in that location.

The check for aide_scan_notification needs mail to be sent at the end of the check and so it will not be satisfied by the remediation for aide_periodic_cron_checking.

#### Suggested Fix
1. Adjust the remediation for aide_scan_notifcation to include the word root before the command in /etc/crontab which will cause both checks to pass
or
2. Remove the requirement for the word root to appear in the aide_periodic_cron_checking check, which will allow it to recognize the remediation applied by aide_scan_notification as valid

nathanpeters008:
#### Description:

Updated aide_scan_notification remediation to run cron job as root and removed '\\' before the '|' to make the grep check compatible with regular grep syntax.

#### Rationale:

The check for aide_periodic_cron_checking expects the word 'root' in /etc/crontab between the time and the command, so the line added by aide_scan_notification will not satisfy it because it does not contain the word 'root' in that location.

The check for aide_scan_notification needs mail to be sent at the end of the check and so it will not be satisfied by the remediation for aide_periodic_cron_checking.

By adding root to the cron remediation for aide_scan_notification, the line added to /etc/crontab will satisfy both checks and you will not end up with 2 cron jobs for aide running every day.

The remediation for aide_periodic_cron_checking adds the following line to /etc/crontab:
`0 5 * * * /usr/sbin/aide  --check`

When the aide_scan_notification remediation is run, because the pipe is escaped, it is literally interpreted as match if you find either of these values `'0 5 * * * /usr/sbin/aide  --check' OR '/bin/mail -s "$(hostname) - AIDE Integrity Check" root@localhost'`

Removing '\' before the '|' makes the grep check compatible with regular grep syntax.

Fixes #2496 #2495 #2112

### DIFF
--- a/shared/fixes/bash/aide_scan_notification.sh
+++ b/shared/fixes/bash/aide_scan_notification.sh
@@ -10,7 +10,7 @@ if [ -f /var/spool/cron/root ]; then
 	VARSPOOL=/var/spool/cron/root
 fi
 
-if ! grep -qR '^.*\/usr\/sbin\/aide\s*\-\-check.*\|.*\/bin\/mail\s*-s\s*".*"\s*root@.*$' $CRONTAB $VARSPOOL $CRONDIRS; then
+if ! grep -qR '^.*\/usr\/sbin\/aide\s*\-\-check.*|.*\/bin\/mail\s*-s\s*".*"\s*root@.*$' $CRONTAB $VARSPOOL $CRONDIRS; then
 	echo '0 5 * * * root /usr/sbin/aide  --check | /bin/mail -s "$(hostname) - AIDE Integrity Check" root@localhost' >> $CRONTAB
 fi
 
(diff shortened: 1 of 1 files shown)


redhatrises:
@openscap-ci add to whitelist

nathanpeters008:
I realized after I submitted this that I had 2 fixes for the same file so I added the 2nd commit after the previous comments.

redhatrises:
Thanks @nathanpeters008 for the fixes! Most appreciated. Ack.

PR #2502, opened 2017-12-21
Added --follow-symlinks to sed commands in display_login_attempts.sh

### LINKED ISSUE #2482: pam remediations - sed is missing --follow-symlinks
This issue was reported by @nathanpeters008 [in the openscap project](https://github.com/OpenSCAP/openscap/issues/924), but it is an SSG bug. Here is the original text for this part of the issue:

`sed` is missing `--follow-symlinks`
Example:
```
if ! `grep -q ^[^#].*pam_succeed_if.*showfailed /etc/pam.d/postlogin` ; then
  if ! grep `^session.*pam_succeed_if.so /etc/pam.d/postlogin` ; then
    echo "session     [default=1]   pam_lastlog.so nowtmp showfailed" >> /etc/pam.d/postlogin
    echo "session     optional      pam_lastlog.so silent noupdate showfailed" >> /etc/pam.d/postlogin
  else
    sed -i '/^session.*pam_succeed_if.so/a session\t    optional\t  pam_lastlog.so silent noupdate showfailed' /etc/pam.d/postlogin
    sed -i '/^session.*pam_succeed_if.so/a session\t    [default=1]\t  pam_lastlog.so nowtmp showfailed' /etc/pam.d/postlogin
  fi
else
  sed -i "s/session[ ]*\[default=1][ ]*pam_lastlog.so.*/session     [default=1]   pam_lastlog.so nowtmp showfailed/g" /etc/pam.d/postlogin
  sed -i "s/session[ ]*optional[ ]*pam_lastlog.so.*/session     optional      pam_lastlog.so silent noupdate showfailed/g" /etc/pam.d/postlogin
fi
```

That code above will always replace replace the symlink from `/etc/pam.d/postlogin` that is supposed to point to `/etc/pam.d/postlogin-ac`.  At that point, postlogin is now a file with actual contents instead of a symlink.  The sed lines should have `--follow-symlinks` added to them to avoid this.

nathanpeters008:
#### Description:

Added --follow-symlinks to sed commands in shared/fixes/bash/display_login_attempts.sh

#### Rationale:

The remediation in shared/fixes/bash/display_login_attempts.sh will always replace replace the symlink from /etc/pam.d/postlogin that is supposed to point to /etc/pam.d/postlogin-ac. At that point, /etc/pam.d/postlogin is now a file with actual contents instead of a symlink. 

The sed lines should have --follow-symlinks added to them to avoid this.

- Fixes #2482

### DIFF
--- a/shared/fixes/bash/display_login_attempts.sh
+++ b/shared/fixes/bash/display_login_attempts.sh
@@ -5,10 +5,10 @@ if ! `grep -q ^[^#].*pam_succeed_if.*showfailed /etc/pam.d/postlogin` ; then
     echo "session     [default=1]   pam_lastlog.so nowtmp showfailed" >> /etc/pam.d/postlogin
     echo "session     optional      pam_lastlog.so silent noupdate showfailed" >> /etc/pam.d/postlogin
   else
-    sed -i '/^session.*pam_succeed_if.so/a session\t    optional\t  pam_lastlog.so silent noupdate showfailed' /etc/pam.d/postlogin
-    sed -i '/^session.*pam_succeed_if.so/a session\t    [default=1]\t  pam_lastlog.so nowtmp showfailed' /etc/pam.d/postlogin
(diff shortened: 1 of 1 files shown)


redhatrises:
Ack. LGTM.

PR #2504, opened 2017-12-22
Ansible: don't use spaces in custom.conf

redhatrises:
#### Description:

- When using Ansible, don't use spaces in between = in custom.conf

#### Rationale:

- Checks were failing when fixed with ansible


yuumasato:
Ack, thanks.

PR #2509, opened 2017-12-30
Add OVAL check and fix for RHEL-07-041001 rule.

### LINKED ISSUE #1861: needs mapping/rule: SRG-OS-000375-GPOS-00160, SV-87041r2_rule, RHEL-07-041001 (pki required packages)

ptitoliv:
Add check and fixes for packages installation needed for smartcard use.

#### Description:
- First contribution proposal for and oval check and the matching fix. 

#### Rationale:

- Fixes #1861


dahaic:
Looks good, ack.

Thank you for the patch!

PR #2514, opened 2018-01-03
Add scenario remediation parameter and fix sshd test scenarios

dahaic:
#### Description:

- Fixes multiple scenarios, and includes new parameter that can be used within scenarios to define which remediations are expected. This fixes error type "expected 'fixed' but result was 'fail'"

#### Rationale:

- When I run sshd related test scenarios on RHEL7, there were multiple errors reported. This PR should fix these errors.


yuumasato:
LGTM, thanks.

PR #2519, opened 2018-01-05
fixed template_BASH_sebool_var with valid bash syntax

### LINKED ISSUE #2513: xccdf_org.ssgproject.content_rule_sebool* bash fix scripts are broken
#### Description of problem:
The generated bash fix scripts for xccdf_org.ssgproject.content_rule_sebool* produce non functional bashcode.

The error received when running these is: `setsebool: illegal value var_abrt_upload_watch_anon_write for boolean abrt_upload_watch_anon_write`

0.1.36-1.el7.centos

#### Steps to Reproduce:

1. Run any of the remediation scripts for xccdf_org.ssgproject.content_rule_sebool*

#### Actual Results:
`setsebool: illegal value var_auditadm_exec_content for boolean auditadm_exec_content`

#### Expected Results:
SELinux boolean is successfully set

The generated scripts contain the following logic:
```
var_abrt_anon_write="false"

setsebool -P abrt_anon_write var_abrt_anon_write
```

The variable in the setsebool command should be prefixed with a `$` in order to be considered a variable by bash, but since it isn't, the error occurs.

#### Suggested Fix:
I think the issue is with the following script : shared/templates/template_BASH_sebool_var
Specifically line 11: `setsebool -P %SEBOOLID% var_%SEBOOLID%`

Assuming that `$` does not have any special meaning to the template mechanism that generates the remediation scripts, the fix should be as simple as replacing the line above with: `setsebool -P %SEBOOLID% $var_%SEBOOLID%`

nathanpeters008:
#### Description:

Added a `$` to var_%SEBOOLID%

#### Rationale:

The current template is broken, and generates bash fix scripts that result in the following error because bash variables must be prefixed with a `$`: `setsebool: illegal value var_abrt_upload_watch_anon_write for boolean abrt_upload_watch_anon_write`

- Fixes #2513

### DIFF
--- a/shared/templates/template_BASH_sebool_var
+++ b/shared/templates/template_BASH_sebool_var
@@ -8,4 +8,4 @@
 . /usr/share/scap-security-guide/remediation_functions
 populate var_%SEBOOLID%
 
-setsebool -P %SEBOOLID% var_%SEBOOLID%
+setsebool -P %SEBOOLID% $var_%SEBOOLID%


yuumasato:
Ack, thanks for catching this.

PR #2521, opened 2018-01-05
Fixing reference to outdated PAM configuration manual

dahaic:
#### Description:

- Fixed reference to PAM configuration manual

#### Rationale:

- PAM configuration manual probably moved to different location, thus URL in our guides https://www.kernel.org/pub/linux/libs/pam/Linux-PAM-html/ no longer works


mpreisler:
ACK

PR #2522, opened 2018-01-06
fix kernel_module_loading checks to work properly on 64 bit systems

### LINKED ISSUE #2227: False failures for audit_rules_kernel_module_loading init and delete?
Testing the `stig-rhel7-disa` profile, and it is reporting an error when remediating `audit_rules_kernel_module_loading_delete` and `audit_rules_kernel_module_loading_init`. The report says:
> Failed to verify applied fix: Checking engine returns: fail

However, when I inspect the system, the rules are present.

```
# grep init_module /etc/audit/audit.rules /etc/audit/rules.d/*
/etc/audit/audit.rules:-a always,exit -F arch=b64 -S init_module -k modules
/etc/audit/rules.d/modules.rules:-a always,exit -F arch=b64 -S init_module -k modules
# grep delete_module /etc/audit/audit.rules /etc/audit/rules.d/*
/etc/audit/audit.rules:-a always,exit -F arch=b64 -S delete_module -k modules
/etc/audit/rules.d/modules.rules:-a always,exit -F arch=b64 -S delete_module -k modules
```

Looking at the OVAL check, and I believe the logic results in requiring **both** b64 and b32 rules:

- https://github.com/OpenSCAP/scap-security-guide/blob/master/shared/templates/static/oval/audit_rules_kernel_module_loading_init.xml#L12-L38
- https://github.com/OpenSCAP/scap-security-guide/blob/master/shared/templates/static/oval/audit_rules_kernel_module_loading_delete.xml#L12-L38

I was under the impression that only the rules matching the system arch needed to be present, and the remediation scripts state the same thing:

- https://github.com/OpenSCAP/scap-security-guide/blob/master/shared/templates/static/bash/audit_rules_kernel_module_loading_init.sh
- https://github.com/OpenSCAP/scap-security-guide/blob/master/shared/templates/static/bash/audit_rules_kernel_module_loading_delete.sh

So, what's the fix? Patch the OVAL files, or the remediation scripts? Or am I just wrong about the OVAL logic?

nathanpeters008:
#### Description:

The OVAL checks for audit_rules_kernel_module_loading_init and audit_rules_kernel_module_loading_delete will fail if you don't have an audit rule for 32 bit kernel init and delete on 64 bit systems.

This PR fixes those checks to properly only require 32 bit rules on 32 bit systems and 64 bit rules on 64 bit systems.

#### Rationale:

For discussion see : https://github.com/OpenSCAP/scap-security-guide/issues/2227

32 bit kernel modules cannot be loaded on 64 bit systems so auditing loading of 32 bit kernel modules on 64 bit systems is not necessary.

- Fixes #2227

### DIFF
--- a/shared/checks/oval/audit_rules_kernel_module_loading_delete.xml
+++ b/shared/checks/oval/audit_rules_kernel_module_loading_delete.xml
@@ -8,44 +8,44 @@
       </affected>
       <description>The audit rules should be configured to log information about kernel module loading and unloading.</description>
     </metadata>
-      <criteria operator="OR">
 
-        <!-- Test the augenrules case -->
-        <criteria operator="AND">
-          <extend_definition comment="audit augenrules" definition_ref="audit_rules_augenrules" />
-          <criteria operator="OR">
-            <!-- System isn't 64-bit => we check presence of 32-bit version of delete_module audit DAC rule -->
-            <criteria operator="AND">
-              <extend_definition comment="32-bit system" definition_ref="system_info_architecture_x86" />
-              <criterion comment="audit augenrules 32-bit delete_module" test_ref="test_32bit_ardm_delete_module_augenrules" />
-            </criteria>
-            <!-- System is 64-bit => we check presence of 64-bit version of delete_module audit DAC rule -->
-            <criteria operator="AND">
-              <extend_definition comment="64-bit system" definition_ref="system_info_architecture_64bit" />
-              <criterion comment="audit augenrules 64-bit delete_module" test_ref="test_64bit_ardm_delete_module_augenrules" />
-            </criteria>
+    <criteria operator="OR">
+
+      <!-- Test the augenrules case -->
+      <criteria operator="AND">
+        <extend_definition comment="audit augenrules" definition_ref="audit_rules_augenrules" />
+        <criteria operator="OR">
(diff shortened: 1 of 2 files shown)


shared/checks/oval/audit_rules_kernel_module_loading_delete.xml
```diff
@@ -8,34 +8,44 @@
       </affected>
       <description>The audit rules should be configured to log information about kernel module loading and unloading.</description>
     </metadata>
+	  
+      <criteria operator="OR">
```
mpreisler: minor nitpick, the indents are borked here

nathanpeters008:
@mpreisler good catch, thanks.  I had been editing this in multiple editors and didn't realize one was set to tabs instead of spaces.

It should be fixed now.

mpreisler:
https://github.com/GovReady/govready/issues/65#issuecomment-68658479 might be relevant here (thanks @redhatrises for the reference)

nathanpeters008:
@mpreisler I have created https://github.com/OpenSCAP/scap-security-guide/pull/2532

If you decide that the logic that you can actually boot a 64 bit system into a 32 bit kernel is actually correct, then please merge that branch instead and reject this PR.

nathanpeters008:
Closing this request as it was decided that https://github.com/OpenSCAP/scap-security-guide/pull/2532 was the correct fix.

PR #2524, opened 2018-01-06
Add check and remediation for RHEL-07-040540

### LINKED ISSUE #1855: needs mapping/rule: SRG-OS-000480-GPOS-00227, SV-86901r1_rule, RHEL-07-040540 (no .shosts)

ptitoliv:
#### Description:

- OVAL check if there is any .shosts files on the file system.
- Remediation removes all the .shosts files if there is a finding.

#### Rationale:

- Fixes #1855

### DIFF
--- a/shared/checks/oval/no_user_host_based_files.xml
+++ b/shared/checks/oval/no_user_host_based_files.xml
@@ -1,0 +1,22 @@
+<def-group oval_version="5.11">
+  <definition class="compliance" id="no_user_host_based_files" version="1">
+    <metadata>
+      <title>No .shosts file deployed on the system</title>
+      <affected family="unix">
+        <platform>multi_platform_rhel</platform>
+      </affected>
+      <description>There should not be any .shosts files on the system.</description>
+    </metadata>
+    <criteria operator="AND">
+      <criterion test_ref="test_no_shosts" negate="true" />
+    </criteria>
+  </definition>
+  <unix:file_test check="all" check_existence="at_least_one_exists" comment="look for .shosts in /" id="test_no_shosts" version="1">
+    <unix:object object_ref="object_no_shosts_files_root" />
+  </unix:file_test>
+  <unix:file_object comment="look for any .shosts file on the system" id="object_no_shosts_files_root" version="1">
+    <unix:behaviors recurse="directories" recurse_direction="down" max_depth="-1" recurse_file_system="local" />
+    <unix:path operation="equals">/</unix:path>
+    <unix:filename operation="pattern match">\.shosts</unix:filename>
+  </unix:file_object>
+</def-group>
--- a/shared/checks/oval/no_user_host_based_files.xml
+++ b/shared/checks/oval/no_user_host_based_files.xml
@@ -1,4 +1,4 @@
-<def-group oval_version="5.11">
+<def-group oval_version="5.10">
   <definition class="compliance" id="no_user_host_based_files" version="1">
     <metadata>
       <title>No .shosts file deployed on the system</title>
@@ -7,11 +7,11 @@
       </affected>
       <description>There should not be any .shosts files on the system.</description>
     </metadata>
-    <criteria operator="AND">
-      <criterion test_ref="test_no_shosts" negate="true" />
+    <criteria>
+      <criterion test_ref="test_no_shosts" />
     </criteria>
   </definition>
-  <unix:file_test check="all" check_existence="at_least_one_exists" comment="look for .shosts in /" id="test_no_shosts" version="1">
+  <unix:file_test check="all" check_existence="none_exist" comment="look for .shosts in /" id="test_no_shosts" version="1">
     <unix:object object_ref="object_no_shosts_files_root" />
   </unix:file_test>
   <unix:file_object comment="look for any .shosts file on the system" id="object_no_shosts_files_root" version="1">
--- a/shared/fixes/bash/no_user_host_based_files.sh
+++ b/shared/fixes/bash/no_user_host_based_files.sh
@@ -1,2 +1,10 @@
 # platform = multi_platform_rhel
-find / -type f -name .shosts -fstype local -exec rm -f '{}' \;
+
+# Identify local mounts
+MOUNT_LIST=$(df | grep "^/dev" | awk '{ print $6 }') 
+
+# Find file on each listed mount point
+for cur_mount in ${MOUNT_LIST}
+do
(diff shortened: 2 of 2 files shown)


shared/checks/oval/no_user_host_based_files.xml
```diff
@@ -0,0 +1,22 @@
...
+    </metadata>
+    <criteria operator="AND">
+      <criterion test_ref="test_no_shosts" negate="true" />
+    </criteria>
+  </definition>
+  <unix:file_test check="all" check_existence="at_least_one_exists" comment="look for .shosts in /" id="test_no_shosts" version="1">
+    <unix:object object_ref="object_no_shosts_files_root" />
+  </unix:file_test>
+  <unix:file_object comment="look for any .shosts file on the system" id="object_no_shosts_files_root" version="1">
+    <unix:behaviors recurse="directories" recurse_direction="down" max_depth="-1" recurse_file_system="all" />
```
redhatrises: Can you change `recurse_file_system="all"` to `recurse_file_system="local"`?

ptitoliv:
@dahaic. I just rebased the branch including the changes you requested. About the remediation script you proposed, i can apply that script but in that case the remediation will be different from the check since the check is about the complete local FS whereas the remediation is only for home directories. Is it a thing allowed ?

dahaic:
@ptitoliv I am afraid that `-fstype local` does not work for me on rhel7. It won't find anything. Specification says remove this file anywhere on the system. I don't see necessity for this, but I don't know how much leeway there is. Also I am afraid I made mistake, it's not `-fstype local` which actually does not work for me on RHEL7, but `-mount`.

But again - we will have to accept that for at least DISA STIG and C2S profiles, this remediation won't work quite often (as these require `/home` on separate partition) and for CJIS, common, and RH CCP it won't work if this file exists on some mounted partition. (`/tmp` or `/var/` usually)

@redhatrises @mpreisler guys, can you take a look? What do you think?

dahaic:
Also I have noticed DISA STIG requires these files to not exist anywhere on the system. Not just in `~/.shosts` where I would expect it. I don't know how much we can diverge from it. Again - help of more experienced will be appreciated :)

shawndwells:
Could create two rules.... one that aligns w/exactly the DISA language, to be included in the disa-stig profile.

The other rule, which is a bit more sane, to be included in the other profiles like ospp.

shawndwells:
(meanwhile, +cc @tbrunell for his awareness this may be something to have disa update in their content)

ptitoliv:
@dahaic : Yeah indeed, the `-fstype local` does not work on GNU find. I think I did the PR without testing enough in blind mode sorry for that :(

Maybe we could do a smarter remediation script with a find instruction excluding known remotefs like nfs and SMBFS/CIFS and maybe also sshfs and other remote fs you could suggest :)

Or maybe a better idea : parsing /proc/mounts of df output and then just considering lines starting with /dev which normally should match only local bloc devices. With such a script, the remediation would work even when /home is mounted on a separated partition. What do you think ?

Maybe a last final ultimate solution : Do the same than the oscap OVAL probe like here ? https://github.com/OpenSCAP/openscap/blob/maint-1.2/src/OVAL/probes/oval_fts.c (Method valid_local_fs)

redhatrises:
Haven't fully tested using multiple partitions, but how about using `-xdev` or `-mount`?

shared/checks/oval/no_user_host_based_files.xml
```diff
@@ -0,0 +1,22 @@
+<def-group oval_version="5.11">
```
redhatrises: There is nothing inherently version `5.11` here, so make this `5.10`

shared/checks/oval/no_user_host_based_files.xml
```diff
@@ -0,0 +1,22 @@
...
+  <definition class="compliance" id="no_user_host_based_files" version="1">
+    <metadata>
+      <title>No .shosts file deployed on the system</title>
+      <affected family="unix">
+        <platform>multi_platform_rhel</platform>
+      </affected>
+      <description>There should not be any .shosts files on the system.</description>
+    </metadata>
+    <criteria operator="AND">
```
redhatrises: remove `operator="AND"`

shared/checks/oval/no_user_host_based_files.xml
```diff
@@ -0,0 +1,22 @@
...
+      <criterion test_ref="test_no_shosts" negate="true" />
```
redhatrises: remove `negate="true"`

shared/checks/oval/no_user_host_based_files.xml
```diff
@@ -0,0 +1,22 @@
...
+    </criteria>
+  </definition>
+  <unix:file_test check="all" check_existence="at_least_one_exists" comment="look for .shosts in /" id="test_no_shosts" version="1">
```
redhatrises: change `check_existence="at_least_one_exists"` to `check_existence="none_exist"`

dahaic:
@redhatrises neither `-mount` nor `-xdev` work with partitions :(

dahaic:
@ptitoliv inspired by your idea with '/dev/' something (which by itself won't work, because for example `/` is local, but contains all the problematic network mounts, I propose this approach:
```
# identify mounts that does not start with /dev
MOUNT_LIST=$(df | grep -v "^/dev" | awk '{ print $6 }' | tail -n +2 | tr '\n' " ")
# build find command that will skip those
PRUNE_PART=" -path $( echo ${MOUNT_LIST%?} | sed 's/ / -o -path /g')"
find / -type d \( $PRUNE_PART \) -prune -o -name ".shosts" -print
```

ptitoliv:
@dahaic. I did a little implementation with my idea actually using -xdev but for each mount point :)

It gives something like that : 

`# identify mounts that does start with /dev
MOUNT_LIST=$(df | grep "^/dev" | awk '{ print $6 }')

for cur_mount in ${MOUNT_LIST}
do
        find / -xdev -type f -name ".shosts.equiv"
done
`

I tried with an nfs server here and it seems to work for me. Can you test in your environment ?

dahaic:
@ptitoliv cleaner solution for sure :) Just a small typo - you actually don't use `cur_mount` in the example. When fixed, the snipped works perfectly for me.

ptitoliv:
@dahaic Sorry for that newbie error. I updated the script with the use of cur_mount and I directly updated the PR.

dahaic:
@ptitoliv thank you, great patch! ACK

PR #2526, opened 2018-01-09
Fix ID of RHEL6 DISA STIG Profile

yuumasato:
#### Description:

- Fix ID of RHEL6 STIG DISA Profile

#### Rationale:

- In https://github.com/OpenSCAP/scap-security-guide/pull/2376 STIG Profiles `stig-rhel6-workstation-upstream`, `stig-rhel6-server-gui-upstream`, and `stig-rhel6-server-upstream` for RHEL6 were removed and Profile `stig-rhel6-disa was introduced`. But ID for this profile remained the `stig-rhel6-server-upstream`.

Edit: actually it is the ID that remained old, not filename.

### DIFF
--- a/rhel6/CMakeLists.txt
+++ b/rhel6/CMakeLists.txt
@@ -21,7 +21,7 @@ ssg_build_html_cce_table(${PRODUCT})
 
 ssg_build_html_srgmap_tables(${PRODUCT} ${DISA_SRG_TYPE} ${DISA_SRG_VERSION})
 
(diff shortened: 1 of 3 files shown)


mpreisler:
> Profile filename was renamed but ID remained the same.

reference please

yuumasato:
PR description updated with reference.

dahaic:
Thanks Yuuma!

PR #2533, opened 2018-01-11
Added "anssi" profile to the RHEL7 product

mpreisler:
#### Description:

There are only 3 rules for now but this will enable us to collaborate on
the profile.

#### Rationale:

ANSSI is requested more and more by our users.


redhatrises:
@mpreisler there are 4 different profiles for Debian and Ubuntu. Should we also account for that here rather than just having a single ANSSI profile?

mpreisler:
@redhatrises good point, I think we should, I wasn't aware of this. I will adapt this PR accordingly.

redhatrises:
Looks good for starters. More to come I am sure. Ack.

shawndwells:
(ties to https://bugzilla.redhat.com/show_bug.cgi?id=1497415)

PR #2534, opened 2018-01-12
Small test suite tweaks

dahaic:
#### Description:

- Small tweaks of latest Test Suite update.

#### Rationale:

- Result files in profile runs were missing stage identifier
- pep8: W503 line break before binary operator

### DIFF
--- a/tests/ssg_test_suite/rule.py
+++ b/tests/ssg_test_suite/rule.py
@@ -247,8 +247,8 @@ def run_rule_checks(
     supported_and_available_remediations = set(
         script_params['remediation']).intersection(is_supported)
 
-    if (script_context not in ['fail', 'error']
-            or len(supported_and_available_remediations) == 0):
+    if (script_context not in ['fail', 'error'] or
+            len(supported_and_available_remediations) == 0):
         return success
 
     success = oscap_run_rule('remediation', 'fixed')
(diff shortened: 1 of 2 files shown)


tests/ssg_test_suite/rule.py
```diff
@@ -247,8 +247,8 @@ def oscap_run_rule(stage, context):
     supported_and_available_remediations = set(
         script_params['remediation']).intersection(is_supported)
 
-    if (script_context not in ['fail', 'error']
-            or len(supported_and_available_remediations) == 0):
+    if (script_context not in ['fail', 'error'] or
```
matejak: It is not a big thing, but PEP8 lists reasons why to break lines before operators, which I also like:
https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

matejak: I read the `if` statement like this: I see the first condition, then the other one line below. It is easy for me to see the relationship between those two if the `or` is at the beginning of the second line.
In this sense, `or` seems to be the same to me as another binary operators.

mpreisler: Either way works for me, this is probably personal preference. I would prefer whatever doesn't trigger a pep8 error unless it decreases readability. In this case I don't think it's worth it to go against pep8.

matejak: OK, it is a bikeshed-class of problem anyway.

PR #2536, opened 2018-01-12
Changed test suite benchmark specification to use Ref-Id.

matejak:
This PR allows to specify the benchmark operated on by the test suite using a component reference ID. This is more convenient as the ID can be queried by 'oscap info'. 

The change makes the command-line interface backwards-incompatible, but backward compatibility can be implemented at minor cost.

Aside from that, a small refactoring was performed, and a XML parsing functionality was moved to a new module, so they can share namespace definitions.

### DIFF
--- a/tests/ssg_test_suite/xml_operations.py
+++ b/tests/ssg_test_suite/xml_operations.py
@@ -1,0 +1,57 @@
+#!/usr/bin/env python2
+import xml.etree.cElementTree as ET
+
+import logging
+
+NAMESPACES = {
+    'xccdf': "http://checklists.nist.gov/xccdf/1.2",
+    'ds': "http://scap.nist.gov/schema/scap/source/1.2",
+    'xlink': "http://www.w3.org/1999/xlink",
+}
+
+
+logging.getLogger(__name__).addHandler(logging.NullHandler())
+
+
+def infer_benchmark_id_from_component_ref_id(datastream, ref_id):
+    root = ET.parse(datastream).getroot()
+    component_ref_node = root.find("*//ds:component-ref[@id='{0}']"
+                                   .format(ref_id), NAMESPACES)
+    if component_ref_node is None:
+        msg = (
+            'Component reference of Ref-Id {} not found within datastream'
+            .format(ref_id))
+        raise RuntimeError(msg)
+
+    comp_id = component_ref_node.get('{%s}href' % NAMESPACES['xlink'])
+    comp_id = comp_id.lstrip('#')
+    component_node = root.find("ds:component[@id='{0}']"
+                               .format(comp_id), NAMESPACES)
+    if component_node is None:
+        msg = 'Component of Id {} not found within datastream'.format(comp_id)
+        raise RuntimeError(msg)
+
+    benchmark_node = component_node.find("xccdf:Benchmark", NAMESPACES)
+    if benchmark_node is None:
+        msg = (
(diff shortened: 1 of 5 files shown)


tests/test_suite.py
```diff
@@ -109,6 +111,16 @@
 # debug otherwise it cuts silently everything
 log.setLevel(logging.DEBUG)
 
+try:
+    bench_id = xml_operations.infer_benchmark_id_from_component_ref_id(
```
mpreisler: why not use the component ref directly? oscap supports it:

```
--xccdf-id ID
                     Takes component ref with given ID from checklists. This allows to select a particular XCCDF component even in cases where there are 2 XCCDFs in  one  datastream.  If
                     none is given, the first component from the checklists element is used.
```

tests/test_suite.py
```diff
@@ -38,12 +40,12 @@
                            required=True,
                            help=("Path to the Source DataStream on this "
                                  "machine which is going to be tested"))
-common_parser.add_argument("--benchmark-id",
-                           dest="benchmark_id",
-                           metavar="BENCHMARK",
+common_parser.add_argument("--ref-id",
```
mpreisler: we use `xccdf-id` in oscap for this, it's not the best name but it would make sense to be consistent

dahaic:
Jenkins test this please.

tests/test_suite.py
mpreisler: It seems to me that you can refactor the code to avoid the whole infer id code. Instead of looking for //cdf:Benchmark you can look for ds:component[@id=...]/cdf:Benchmark. Or am I missing something? Would make the code way shorter.

tests/ssg_test_suite/xml_operations.py
```diff
@@ -0,0 +1,57 @@
...
+
+    benchmark_node = component_node.find("xccdf:Benchmark", NAMESPACES)
+    if benchmark_node is None:
+        msg = (
+            'Benchmark not found within comonent of Id {}'
+            .format(comp_id)
+        )
+        raise RuntimeError(msg)
+
+    return benchmark_node.get('id')
```
mpreisler: I am not 100% sure whether ElementTree supports it but this can be done iwth just one XPath on maybe 3 lines of code.

```//cdf:Benchmark[parent::ds:component/@id=//ds:component-ref[@id=$ref_id]]/@id```

Untested and probably needs changes but I think it gets the idea across.

That said, I think this whole code can be avoided completely by changing the profile scraping code.

matejak: I didn't know what is possible to achieve with the xpath language. This one-liner would be very impressive, but the xpath from python standard library doesn't support this as I get errors that indicate that `//` is not supported properly and then it throws `prefix 'parent' not found in prefix map` error (i.e. it doesn't understand XPath Axes).

tests/test_suite.py
matejak: I still see need for a ref-id -> benchmark function, although it has potential be a one-liner (if one knows how to write it and the module supports the required part of the XPath language).

tests/ssg_test_suite/xml_operations.py
mpreisler: ACK. I looked into it and the python2 stdlib xpath is very limited.

mpreisler:
ACK

PR #2540, opened 2018-01-15
Fatal error if user attempts in-source build

mpreisler:
Previously we would show an error but continue configuring. That makes it harder to figure out what the error was.

Thanks goes to @matejak for this suggestion.


redhatrises:
Ack.

PR #2541, opened 2018-01-15
Add test scenarios of rule_umask_for_daemons.

dahaic:
#### Description:

- Test scenarios to cover rule_umask_for_daemons. Profile is deliberately NOTFOUND, because there is no profile in RHEL7 with this rule enabled.

#### Rationale:

- Extending test coverage.

- Covers RHBZ#1520493


jan-cerny:
This is great! I really like that you covered a bug in the upstream test suite.

PR #2544, opened 2018-01-16
Reference real files instead of procfs and sysfs files

### LINKED ISSUE #810: [RHEL/6] CCE-26910-0, DISA FSO RHEL-06-000282, might need to explain that we are talking about real files and not proc or sysfs files
CCE-26910-0, DISA FSO RHEL-06-000282, might need to explain that we are talking about real files and not proc or sysfs files.

redhatrises:
#### Description:

- Reference real files instead of procfs and sysfs files when checking for world-writeable permissions

#### Rationale:

- procfs and sysfs files aren't real files, and configuring permissions on files that are handled by the kernel don't make sense.

- Fixes #810


mpreisler:
ACK

PR #2548, opened 2018-01-17
Small changes to RHEL7 checks and fixes to comply with STIG

chrisle925:
#### Description:

Small changes to properly comply with the STIG

#### Rationale:

**sshd_disable_kerb_auth.xml**: should only pass if the pattern exists (otherwise it should fail and the fix should be applied)
**sshd_enable_strictmodes.xml**: should only pass if the pattern exists (otherwise it should fail and the fix should be applied)
**audit_rules_sysadmin_actions.sh**: missing /etc/sudoers.d/ and changing actions to privileged-actions
**sshd_use_approved_ciphers.sh**: removing unapproved ciphers
**disable_host_auth.xml**: should only pass if the pattern exists (otherwise it should fail and the fix should be applied)
**partition_for_tmp.xml**: previous check would never not pass
**sshd_disable_rhosts.xml**: should only pass if the pattern exists (otherwise it should fail and the fix should be applied)
**audit_rules_kernel_module_loading_delete.sh**: forcing a single architecture and changing modules to module-change
**audit_rules_kernel_module_loading_init.sh**: forcing a single architecture and changing modules to module-change
**audit_rules_kernel_module_loading_insmod.sh**: adding -F option and changing modules to module-change
**audit_rules_kernel_module_loading_modprobe.sh**: adding -F option and changing modules to module-change
**audit_rules_kernel_module_loading_rmmod.sh**: adding -F option and changing modules to module-change
**rpm_verify_permissions.sh**: adding setugids to loop

### DIFF
--- a/rhel7/checks/oval/sshd_disable_kerb_auth.xml
+++ b/rhel7/checks/oval/sshd_disable_kerb_auth.xml
@@ -26,14 +26,14 @@ the SSH Server.</description>
       </criteria>
     </criteria>
   </definition>
-  <ind:textfilecontent54_test check="all" check_existence="none_exist"
+  <ind:textfilecontent54_test check="all" check_existence="all_exist"
   comment="tests the value of KerberosAuthentication setting in the /etc/ssh/sshd_config file"
   id="test_sshd_disable_kerb_auth" version="1">
     <ind:object object_ref="obj_sshd_disable_kerb_auth" />
   </ind:textfilecontent54_test>
   <ind:textfilecontent54_object id="obj_sshd_disable_kerb_auth" version="1">
     <ind:filepath>/etc/ssh/sshd_config</ind:filepath>
-    <ind:pattern operation="pattern match">^[\s]*(?i)KerberosAuthentication(?-i)[\s]+yes[\s]*(?:|(?:#.*))?$</ind:pattern>
+    <ind:pattern operation="pattern match">^[\s]*(?i)KerberosAuthentication(?-i)[\s]+no[\s]*(?:|(?:#.*))?$</ind:pattern>
     <ind:instance datatype="int">1</ind:instance>
   </ind:textfilecontent54_object>
 </def-group>
--- a/rhel7/checks/oval/sshd_enable_strictmodes.xml
+++ b/rhel7/checks/oval/sshd_enable_strictmodes.xml
@@ -26,14 +26,14 @@ and configurations.</description>
       </criteria>
     </criteria>
   </definition>
-  <ind:textfilecontent54_test check="all" check_existence="none_exist"
+  <ind:textfilecontent54_test check="all" check_existence="all_exist"
   comment="tests the value of StrictModes setting in the /etc/ssh/sshd_config file"
   id="test_sshd_enable_strictmodes" version="1">
     <ind:object object_ref="obj_sshd_enable_strictmodes" />
   </ind:textfilecontent54_test>
   <ind:textfilecontent54_object id="obj_sshd_enable_strictmodes" version="1">
     <ind:filepath>/etc/ssh/sshd_config</ind:filepath>
-    <ind:pattern operation="pattern match">^[\s]*(?i)StrictModes(?-i)[\s]+no[\s]*(?:|(?:#.*))?$</ind:pattern>
+    <ind:pattern operation="pattern match">^[\s]*(?i)StrictModes(?-i)[\s]+yes[\s]*(?:|(?:#.*))?$</ind:pattern>
     <ind:instance datatype="int">1</ind:instance>
   </ind:textfilecontent54_object>
 </def-group>
--- a/rhel7/fixes/bash/audit_rules_sysadmin_actions.sh
+++ b/rhel7/fixes/bash/audit_rules_sysadmin_actions.sh
@@ -4,5 +4,8 @@
 . /usr/share/scap-security-guide/remediation_functions
 
 # Perform the remediation for both possible tools: 'auditctl' and 'augenrules'
-fix_audit_watch_rule "auditctl" "/etc/sudoers" "wa" "actions"
-fix_audit_watch_rule "augenrules" "/etc/sudoers" "wa" "actions"
+fix_audit_watch_rule "auditctl" "/etc/sudoers" "wa" "privileged-actions"
+fix_audit_watch_rule "augenrules" "/etc/sudoers" "wa" "privileged-actions"
+
+fix_audit_watch_rule "auditctl" "/etc/sudoers.d/" "wa" "privileged-actions"
+fix_audit_watch_rule "augenrules" "/etc/sudoers.d/" "wa" "privileged-actions"
--- a/rhel7/fixes/bash/sshd_use_approved_ciphers.sh
+++ b/rhel7/fixes/bash/sshd_use_approved_ciphers.sh
@@ -3,4 +3,4 @@
 # Include source function library.
 . /usr/share/scap-security-guide/remediation_functions
 
-replace_or_append '/etc/ssh/sshd_config' '^Ciphers' 'aes128-ctr,aes192-ctr,aes256-ctr,aes128-cbc,3des-cbc,aes192-cbc,aes256-cbc' '@CCENUM@' '%s %s'
+replace_or_append '/etc/ssh/sshd_config' '^Ciphers' 'aes128-ctr,aes192-ctr,aes256-ctr' '@CCENUM@' '%s %s'
--- a/shared/checks/oval/disable_host_auth.xml
+++ b/shared/checks/oval/disable_host_auth.xml
@@ -25,14 +25,14 @@
       </criteria>
     </criteria>
   </definition>
-  <ind:textfilecontent54_test check="all" check_existence="none_exist"
+  <ind:textfilecontent54_test check="all" check_existence="all_exist"
   comment="sshd HostbasedAuthentication" id="test_sshd_hostbasedauthentication"
   version="1">
     <ind:object object_ref="object_sshd_hostbasedauthentication" />
   </ind:textfilecontent54_test>
   <ind:textfilecontent54_object id="object_sshd_hostbasedauthentication" version="2">
     <ind:filepath>/etc/ssh/sshd_config</ind:filepath>
-    <ind:pattern operation="pattern match">^[\s]*(?i)HostbasedAuthentication(?-i)[\s]+yes[\s]*(?:|(?:#.*))?$</ind:pattern>
+    <ind:pattern operation="pattern match">^[\s]*(?i)HostbasedAuthentication(?-i)[\s]+no[\s]*(?:|(?:#.*))?$</ind:pattern>
     <ind:instance datatype="int">1</ind:instance>
   </ind:textfilecontent54_object>
 </def-group>
--- a/shared/checks/oval/partition_for_tmp.xml
+++ b/shared/checks/oval/partition_for_tmp.xml
@@ -13,7 +13,7 @@
       <criterion test_ref="test_tmp_partition" comment="/tmp on own partition" />
     </criteria>
   </definition>
-  <linux:partition_test check="all" check_existence="all_exist"
+  <linux:partition_test check="all" check_existence="none_exist"
   id="test_tmp_partition" version="1" comment="/tmp on own partition">
     <linux:object object_ref="object_own_tmp_partition" />
   </linux:partition_test>
--- a/shared/checks/oval/sshd_disable_rhosts.xml
+++ b/shared/checks/oval/sshd_disable_rhosts.xml
@@ -26,7 +26,7 @@
       </criteria>
     </criteria>
   </definition>
-  <ind:textfilecontent54_test check="all" check_existence="none_exist"
+  <ind:textfilecontent54_test check="all" check_existence="all_exist"
   comment="Tests the value of the IgnoreRhosts[\s]*(&lt;:nocomment:&gt;*) setting in the /etc/ssh/sshd_config file"
   id="test_sshd_rsh_emulation_disabled" version="1">
     <ind:object object_ref="obj_sshd_rsh_emulation_disabled" />
@@ -34,7 +34,7 @@
   <ind:textfilecontent54_object id="obj_sshd_rsh_emulation_disabled"
   version="2">
     <ind:filepath>/etc/ssh/sshd_config</ind:filepath>
-    <ind:pattern operation="pattern match">^[\s]*(?i)IgnoreRhosts(?-i)[\s]+no[\s]*(?:|(?:#.*))?$</ind:pattern>
+    <ind:pattern operation="pattern match">^[\s]*(?i)IgnoreRhosts(?-i)[\s]+yes[\s]*(?:|(?:#.*))?$</ind:pattern>
     <ind:instance datatype="int">1</ind:instance>
   </ind:textfilecontent54_object>
 </def-group>
--- a/shared/fixes/bash/audit_rules_kernel_module_loading_delete.sh
+++ b/shared/fixes/bash/audit_rules_kernel_module_loading_delete.sh
@@ -5,17 +5,17 @@
 
 # First perform the remediation of the syscall rule
 # Retrieve hardware architecture of the underlying system
-# If the system has a 32-bit processor, only the 32-bit rule is needed.
-# If the system has a 64-bit processor, both arch 32 and 64 need to be included in
-# the audit file because it is not possible to know if the computer will be booted
-# in 64 or 32 bit mode or for which architecture a binary is compiled.
-[ "$(getconf LONG_BIT)" = "32" ] && RULE_ARCHS=("b32") || RULE_ARCHS=("b32" "b64")
+# Note: 32-bit kernel modules can't be loaded / unloaded on 64-bit kernel =>
+#       it's not required on a 64-bit system to check also for the presence
+#       of 32-bit's equivalent of the corresponding rule. Therefore for
+#       each system it's enought to check presence of system's native rule form.
+[ "$(getconf LONG_BIT)" = "32" ] && RULE_ARCHS=("b32") || RULE_ARCHS=("b64")
 
 for ARCH in "${RULE_ARCHS[@]}"
 do
 	PATTERN="-a always,exit -F arch=$ARCH -S delete_module \(-F key=\|-k \).*"
-	GROUP="modules"
-	FULL_RULE="-a always,exit -F arch=$ARCH -S delete_module -k modules"
+	GROUP="module-change"
+	FULL_RULE="-a always,exit -F arch=$ARCH -S delete_module -k module-change"
 	# Perform the remediation for both possible tools: 'auditctl' and 'augenrules'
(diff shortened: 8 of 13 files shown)


mpreisler:
@openscap-ci test this please

rhel7/checks/oval/sshd_disable_kerb_auth.xml
```diff
@@ -26,14 +26,14 @@ the SSH Server.</description>
...
-  <ind:textfilecontent54_test check="all" check_existence="none_exist"
+  <ind:textfilecontent54_test check="all" check_existence="all_exist"
   comment="tests the value of KerberosAuthentication setting in the /etc/ssh/sshd_config file"
   id="test_sshd_disable_kerb_auth" version="1">
     <ind:object object_ref="obj_sshd_disable_kerb_auth" />
   </ind:textfilecontent54_test>
   <ind:textfilecontent54_object id="obj_sshd_disable_kerb_auth" version="1">
     <ind:filepath>/etc/ssh/sshd_config</ind:filepath>
-    <ind:pattern operation="pattern match">^[\s]*(?i)KerberosAuthentication(?-i)[\s]+yes[\s]*(?:|(?:#.*))?$</ind:pattern>
+    <ind:pattern operation="pattern match">^[\s]*(?i)KerberosAuthentication(?-i)[\s]+no[\s]*(?:|(?:#.*))?$</ind:pattern>
```
shawndwells: Nack. In RHEL, the default value of ``KerberosAuthentication`` is ``no``. The current OVAL check evaluates for existence of ``yes``, which is the insecure setting.

The reason for this approach is to prevent sysadmins from having to explicitly add ``KerberosAuthentication no`` to their config files when that's the default behavior.

This reflects a bug in the DISA content and should be fixed downstream by them.

rhel7/checks/oval/sshd_enable_strictmodes.xml
```diff
@@ -26,14 +26,14 @@ and configurations.</description>
...
-  <ind:textfilecontent54_test check="all" check_existence="none_exist"
+  <ind:textfilecontent54_test check="all" check_existence="all_exist"
   comment="tests the value of StrictModes setting in the /etc/ssh/sshd_config file"
   id="test_sshd_enable_strictmodes" version="1">
     <ind:object object_ref="obj_sshd_enable_strictmodes" />
   </ind:textfilecontent54_test>
   <ind:textfilecontent54_object id="obj_sshd_enable_strictmodes" version="1">
     <ind:filepath>/etc/ssh/sshd_config</ind:filepath>
-    <ind:pattern operation="pattern match">^[\s]*(?i)StrictModes(?-i)[\s]+no[\s]*(?:|(?:#.*))?$</ind:pattern>
+    <ind:pattern operation="pattern match">^[\s]*(?i)StrictModes(?-i)[\s]+yes[\s]*(?:|(?:#.*))?$</ind:pattern>
```
shawndwells: Nack. Same reason as KerbAuth above.

rhel7/fixes/bash/audit_rules_sysadmin_actions.sh
```diff
@@ -4,5 +4,8 @@
 . /usr/share/scap-security-guide/remediation_functions
 
 # Perform the remediation for both possible tools: 'auditctl' and 'augenrules'
-fix_audit_watch_rule "auditctl" "/etc/sudoers" "wa" "actions"
-fix_audit_watch_rule "augenrules" "/etc/sudoers" "wa" "actions"
+fix_audit_watch_rule "auditctl" "/etc/sudoers" "wa" "privileged-actions"
+fix_audit_watch_rule "augenrules" "/etc/sudoers" "wa" "privileged-actions"
+
+fix_audit_watch_rule "auditctl" "/etc/sudoers.d/" "wa" "privileged-actions"
+fix_audit_watch_rule "augenrules" "/etc/sudoers.d/" "wa" "privileged-actions"
```
shawndwells: ack to these changes.

rhel7/fixes/bash/sshd_use_approved_ciphers.sh
```diff
@@ -3,4 +3,4 @@
 # Include source function library.
 . /usr/share/scap-security-guide/remediation_functions
 
-replace_or_append '/etc/ssh/sshd_config' '^Ciphers' 'aes128-ctr,aes192-ctr,aes256-ctr,aes128-cbc,3des-cbc,aes192-cbc,aes256-cbc' '@CCENUM@' '%s %s'
```
redhatrises: Can you make this a variable? DISA feels only a subset of FIPS cipbers should be used; however, other gov agencies will use any FIPS cipher.

shared/fixes/bash/audit_rules_kernel_module_loading_rmmod.sh
```diff
@@ -5,5 +5,5 @@
 
 # Perform the remediation for both possible tools: 'auditctl' and 'augenrules'
 
-fix_audit_watch_rule "auditctl" "/usr/sbin/rmmod" "x" "modules"
-fix_audit_watch_rule "augenrules" "/usr/sbin/rmmod" "x" "modules"
+fix_audit_watch_rule "auditctl" "/sbin/rmmod" "x -F auid!=4294967295" "module-change"
+fix_audit_watch_rule "augenrules" "/sbin/rmmod" "x -F auid!=4294967295" "module-change"
```
shawndwells: Seems like both should be audited, actually.

shared/checks/oval/disable_host_auth.xml
```diff
@@ -25,14 +25,14 @@
...
-  <ind:textfilecontent54_test check="all" check_existence="none_exist"
+  <ind:textfilecontent54_test check="all" check_existence="all_exist"
   comment="sshd HostbasedAuthentication" id="test_sshd_hostbasedauthentication"
   version="1">
     <ind:object object_ref="object_sshd_hostbasedauthentication" />
   </ind:textfilecontent54_test>
   <ind:textfilecontent54_object id="object_sshd_hostbasedauthentication" version="2">
     <ind:filepath>/etc/ssh/sshd_config</ind:filepath>
-    <ind:pattern operation="pattern match">^[\s]*(?i)HostbasedAuthentication(?-i)[\s]+yes[\s]*(?:|(?:#.*))?$</ind:pattern>
+    <ind:pattern operation="pattern match">^[\s]*(?i)HostbasedAuthentication(?-i)[\s]+no[\s]*(?:|(?:#.*))?$</ind:pattern>
```
redhatrises: Nack. This is already the default. No need to set it.

shared/fixes/bash/audit_rules_kernel_module_loading_rmmod.sh
shawndwells: The XCCDF and OVAL should be updated too. Is this something you could do, @chrisle925 ?

shared/checks/oval/partition_for_tmp.xml
```diff
@@ -13,7 +13,7 @@
       <criterion test_ref="test_tmp_partition" comment="/tmp on own partition" />
     </criteria>
   </definition>
-  <linux:partition_test check="all" check_existence="all_exist"
+  <linux:partition_test check="all" check_existence="none_exist"
```
redhatrises: Can you provide more detail on why this change? Check will pass now if `/tmp` does not exist.

shared/checks/oval/sshd_disable_rhosts.xml
```diff
@@ -26,15 +26,15 @@
...
+  <ind:textfilecontent54_test check="all" check_existence="all_exist"
   comment="Tests the value of the IgnoreRhosts[\s]*(&lt;:nocomment:&gt;*) setting in the /etc/ssh/sshd_config file"
   id="test_sshd_rsh_emulation_disabled" version="1">
     <ind:object object_ref="obj_sshd_rsh_emulation_disabled" />
   </ind:textfilecontent54_test>
   <ind:textfilecontent54_object id="obj_sshd_rsh_emulation_disabled"
   version="2">
     <ind:filepath>/etc/ssh/sshd_config</ind:filepath>
-    <ind:pattern operation="pattern match">^[\s]*(?i)IgnoreRhosts(?-i)[\s]+no[\s]*(?:|(?:#.*))?$</ind:pattern>
+    <ind:pattern operation="pattern match">^[\s]*(?i)IgnoreRhosts(?-i)[\s]+yes[\s]*(?:|(?:#.*))?$</ind:pattern>
```
redhatrises: Nack. This is already the default. No need to set it.

shared/fixes/bash/audit_rules_kernel_module_loading_delete.sh
```diff
@@ -5,17 +5,17 @@
 
 # First perform the remediation of the syscall rule
 # Retrieve hardware architecture of the underlying system
-# If the system has a 32-bit processor, only the 32-bit rule is needed.
-# If the system has a 64-bit processor, both arch 32 and 64 need to be included in
-# the audit file because it is not possible to know if the computer will be booted
-# in 64 or 32 bit mode or for which architecture a binary is compiled.
-[ "$(getconf LONG_BIT)" = "32" ] && RULE_ARCHS=("b32") || RULE_ARCHS=("b32" "b64")
+# Note: 32-bit kernel modules can't be loaded / unloaded on 64-bit kernel =>
```
redhatrises: Nack. A 32bit kernel however can be used, and now you are no longer auditing.

shared/fixes/bash/audit_rules_kernel_module_loading_init.sh
```diff
@@ -5,17 +5,17 @@
 
 # First perform the remediation of the syscall rule
 # Retrieve hardware architecture of the underlying system
-# If the system has a 32-bit processor, only the 32-bit rule is needed.
-# If the system has a 64-bit processor, both arch 32 and 64 need to be included in
-# the audit file because it is not possible to know if the computer will be booted
-# in 64 or 32 bit mode or for which architecture a binary is compiled.
-[ "$(getconf LONG_BIT)" = "32" ] && RULE_ARCHS=("b32") || RULE_ARCHS=("b32" "b64")
+# Note: 32-bit kernel modules can't be loaded / unloaded on 64-bit kernel =>
```
redhatrises: Nack. A 32bit kernel however can be used, and now you are no longer auditing.

shared/fixes/bash/rpm_verify_permissions.sh
```diff
@@ -9,7 +9,7 @@ declare -a SETPERMS_RPM_LIST
 
 # Create a list of files on the system having permissions different from what
 # is expected by the RPM database
-FILES_WITH_INCORRECT_PERMS=($(rpm -Va --nofiledigest | grep '^.M' | cut -d ' ' -f4-))
+FILES_WITH_INCORRECT_PERMS=($(rpm -Va --nofiledigest | grep '^.M' | cut -d ' ' -f5-))
```
redhatrises: Nack. This breaks things. See https://github.com/OpenSCAP/scap-security-guide/pull/2531

shared/fixes/bash/rpm_verify_permissions.sh
```diff
@@ -29,4 +29,5 @@ SETPERMS_RPM_LIST=( $(echo "${SETPERMS_RPM_LIST[@]}" | sort -n | uniq) )
 for RPM_PACKAGE in "${SETPERMS_RPM_LIST[@]}"
 do
 	rpm --setperms "${RPM_PACKAGE}"
+	rpm --setugids "${RPM_PACKAGE}"
```
redhatrises: Nack. This is now a separate rule. See the `rpm_verify_ownership` rule.

redhatrises:
@chrisle925 thanks for submitting your PR! We appreciate really your contribution. We have a few comments that need to be fixed before merging. Thanks!

rhel7/fixes/bash/sshd_use_approved_ciphers.sh
redhatrises: Both the OVAL, XCCDF, and profile will have to be updated for the variable. If you need help on that, please let us know.

rhel7/checks/oval/sshd_disable_kerb_auth.xml
thenefield: I would disagree with this position. The point of having the STIG profile is to closely match, as much as possible, the specific requirements outlined in the STIG. While I fully understand your point of view, customers looking to match our system configuration with the text indicated in the STIG would mark it as an open finding.

I think that if there is a disagreement with the text in the STIG, this should be brought forth to DISA. Then when they update the STIG to match your guidance, it may be appropriate to revert back.

Chris has worked quite a bit aligning the current content to match the STIG content. I'm sure others would appreciate changes that further align their systems to the STIG requirements, as they are currently defined.

shawndwells: Before going into this further I want to clearly separate my words and opinions about DISA STIG content from any perceptions of the DISA staff. This isn't a reflection of anything you (@thenefield, @chrisle925) said. Rather I'm trying to be mindful that my own public comments are often... strong... about this topic. DISA did truly attempt to publish the best guidance they could given organizational, process, and time constraints. They even invited @tbrunell and myself for multiple in-person meetings. Lead by @jmackani, the DISA team proactively sent out drafts of their baseline MANY times and proactively worked with gov, industry, and Red Hat, to resolve as many comments as possible. In an effort to increase quality DISA actually delayed publication to get another round of feedback resolved. I understand that some of their government team unfortunately received blowback on this decision -- even though it was the right call. Opinions on their process and final product may be strong, but that's not a reflection of views on their people who have repeatedly demonstrated attempts to be collaborative and communicative. Heck, just last week DISA chirped in on some RHEL 7.5 bugs (https://github.com/OpenSCAP/scap-security-guide/issues/2568).

That said, DISA deviated from NSA guidance, DoD guidance, US Government configuration guidance, Red Hat guidance, and commercially accepted practices, with their baseline. The DISA content was published without acceptance or review from the DoD consensus members (DAA reps from DoD commands, NSA Information Assurance, civilian DAA staff, Linux vendors). This should not be taken as gross incompetence of the DISA team, but rather a reflection of ambitious time schedules imposed on the team actually doing the work and immense pressures from across DoD to publish *something*. 

Intentions aside, is it reasonable to expect all these varying entities to introduce bugs and regressions into their content in order to align with what DISA published? The SCAP Security Guide project has relationships with downstream commercial distributions like RHEL and Ubuntu LTS. Even Oracle uses content from the OpenSCAP and SCAP Security Guide projects in OEL. Should these companies be expected to ship DISA's bugs and regressions to literally *millions* of Linux endpoints across the globe?

Open source projects live and die by their upstream<->downstream relationships. When RHEL or Ubuntu have kernel bugs, those fixes are first addressed upstream and then those patches are re-integrated into downstream distributions. At the end of the day DISA is a *downstream* of SCAP Security Guide. 

Even more important than upstream<->downstream relationships is actual users and usability. To that end, like you (@thenefield) mentioned, SCAP Security Guide does have a profile (``stig-rhel7-disa``) intended to align as closely as possible to DISA's content. Kind of a recognition that many users find themselves bound to DISA's STIG content instead of the DoD and Government recommendations. 

The ``stig-rhel7-disa`` profile can be used to house closely aligned configuration checks and remediations. This would require temporarily duplicating existing XCCDF, OVAL, and remediation such as ``sshd_disable_kerb_auth``, to something like ``disa_temp_sshd_disable_kerb_auth``, of which would contain the deltas. Those rules would ONLY be selected in the ``stig-rhel7-disa`` profile while we wait for DISA STIG content to be updated.

If someone cares enough to actually do all that, they'll likely find other community members care about them as fellow collaborators and would provide any help needed. 

The issue with this pull request is many of the DISA STIG bugs and regressions are being introduced into "core" rules that are selected in other profiles. If we can temporarily isolate the changes to just the ``stig-rhel7-disa`` profile, a PR could be merged. And crap, in hindsight, expressing things that way would have saved a whole lot of typing just now.... =/ 

If you (@thenefield or @chrisle925) are interested in such a PR, we can first double check with DISA about their recommended approach. We may find they have bug fixes coming soon, or deem the deltas "close enough" not to impact a CCRI or STIG compliance.

thenefield: I greatly appreciate your feedback and completely understand where your coming from. We have also found some STIG mitigations to break the system when applied. But nobody is perfect. We just have to be able to demonstrate compliance as best possible. But as you said there are other profiles that may not require the system to be configured in the manner the STIG is requiring currently. While the logic you provided makes perfect sense, individuals evaluating a system against the STIG would still trigger it as a finding. I like your idea of creating a similar remediation and xml check for those that differ in how DISA requires them to be configured and prepending them with "disa_". I think that's the approach we will take moving forward, at least for the SSH configurations.

Thanks for the suggestion!

redhatrises: I have a hard time with this for a couple of reasons:

1. Setting configuration values in configuration files that are already default values that cannot be changed by a non-admin user is bad security practice no matter who it is that is requesting it. We need to move away from this as an industry and community. It also smells of Windows registry configuration practices where a registry setting is considered default and can be changed by a user unless set explicitly through a GPO by an admin which does not apply on Linux systems. This same line of thinking is pervasive and frustrating to deal with when working with auditors. Just like security through obscurity doesn't work, setting default system values that require privilege users to change does not make your system more secure.

2. Having temp files (oval, remediation, etc.) such as prepending with `disa_temp` opens the door to other for other compliance profiles e.g. HIPAA, C2S, etc. to follow suit which now increases maintenance costs and complexity. Maybe in SSG we need to make it highly configurable compared to what it is today.

3. If SSG is considered upstream to DISA, we need a better workflow and/or process to open tickets with DISA (and DISA with us) to help address and understand as well as sync up content. 
Maybe in the build system we have a built target that generates the content into DISA format for them to easily take, use, and polish for their uses. This in my mind is the best scenario that we should work towards: having DISA and community come to the table to make better security guidance for everyone, the integrators, and even more importantly, the warfighter. In my mind, we need to move forward together to remove some legacy entrenched security practices and replace those with modern best practices and thinking.

thenefield: 1. If software is already enforcing a required value for a setting by default, I can understand how it might seem silly to explicitly define the value for the setting in the configuration. At the same time, it doesn't do any harm explicitly defining the value for the setting, yet it provides assurance that the required value is being enforced.

2. This may not be the most desirable solution. But I think Shawn's suggestion of doing this makes sense and enables us to refine the checks and remediations to what the STIG is currently requiring without impacting other profiles.

3. While the SSG may be considered upstream, bear in mind that DoD auditors specifically look at what the STIG dictates should be configured. Unless there is a compelling reason, such as a configuration breaks functionality, all STIG requirements are expected to be addressed in the way the STIG requires them to be configured. While Red Hat's guidance is very valuable to the process, DoD users are required to follow guidance published by DISA.

I can certainly appreciate Red Hat's initiative to sync up with DISA in developing security guidance for it's software. We could only hope for better coordination from DISA. They do have a tracking system. I have submitted issues with a STIG before. You just send an email to disa.stig_spt@mail.mil describing your concerns.

yuumasato:
Closing this for inactivity.

PR #2550, opened 2018-01-18
Ensure libsemanage-python is installed or Ansible SELinux boolean tas…

rchayes:
#### Description:

- Installs libsemanage-python if package is not present on the system.  Centos7 minimal install doesn't include this package.

#### Rationale:

- All ansible seboolean tasks require this package.  Ansible playbooks halts with error when package is not available.

### DIFF
--- a/shared/templates/template_ANSIBLE_sebool_var
+++ b/shared/templates/template_ANSIBLE_sebool_var
@@ -5,6 +5,11 @@
 # disruption = low
 - (xccdf-var var_%SEBOOLID%)
 
+- name: Ensure libsemanage-python installed
+  package:
+    name: libsemanage-python
+    state: latest
+
 - name: Set SELinux boolean %SEBOOLID% accordingly
   seboolean:
     name: %SEBOOLID%


yuumasato:
LGMT, thanks.

mpreisler:
Doesn't this add this task to every single boolean task? Won't we end up with hundreds of them in the role?

yuumasato:
> Doesn't this add this task to every single boolean task? Won't we end up with hundreds of them in the role?

Yeah, it will. Do we have some kind of `#pragma once` mechanism?

rchayes:
I'm good with fixing small problems in the code but how it's compiled and built not so much.  Is there any similar instances already there I could reference?  If not, I'm afraid this was the best I could come up with.

rchayes:
Let me try that again, yes it adds it to every boolean task.  I looked for an existing alternative in the code where something was created once if any of task X were created but missed anything similar if it exists.  If there is something similar I could use that as a reference.

advocateddrummer:
This affecting me too; am I missing something or does this still happen? The "Ensure libsemanage-python" tasks executes over and over when I run this role/playbook. It clearly still works, but it slows things down. Is there a fix that I missed?

Issue #2467, opened 2017-11-22
Add requirement on ansible version needed for SSG remediations


ptitoliv:
Hello again,

I identified that some generated playbooks are not working because when we use an old ansible version. On my test machine which was a little bit outdated I had ansible 2.2 and it seems that for some playbooks we need at least ansible 2.3.

I think it could be nice to add in the documentation (Developer and User guides) the minimum ansible version needed. And maybe that could be interesting to have such a requirement directly into the specfile ?

Regards,

dahaic:
The documentation is something I agree with. What do you suggest for specfile? Even weak dependency is too strong I would say (as we don't want ansible to be installed because of SSG).

ptitoliv:
For SPECFILE, I was actually thinking about Requires for me since there is more and more remediations using ansible, I thought that scap-security-guide is now linked in a strong way to ansible. Today on a RHEL if you don't have ansible installed, lots of remediation won't be applied.

Anyway, if it is not wanted to create such a strong dependency, the important thing for a newcomer is to understand immediately that ansible has to be used for applying remediation. So, I think the two interesting places for saying that is of course the documentation and also in the generated guide and report introduction.

redhatrises:
@ptitoliv all of this depends on how you are using `scap-security-guide`. `scap-security-guide` has no means of running any sort of remediation whether it be `bash`, `ansible`, `puppet`, etc. so the link to `ansible` is a weak dependency as this all depends on an operational environment.  `scap-security-guide` only provides the scripts for you to determine which ones you want to run and use. Also last time I checked, `oscap` does not run ansible and defaults to `bash`. 

Adding a `Requires` to the specfile doesn't make sense from an infrastructure architecture standpoint, because while you may use ansible, others may use `puppet`, `bash`, or some other language. Therefore, adding every script that `scap-security-guide` can support would have to be added as a `Requires` to the specfile which means all those languages would have to be installed when `scap-security-guide` is installed.

> Today on a RHEL if you don't have ansible installed, lots of remediation won't be applied.

This is not true with `oscap` defaulting to bash.

ptitoliv:
[...]

PR #2560, opened 2018-01-22
Add Ansible pre-task version checking for Ansible roles

redhatrises:
#### Description:

- Add Ansible pre-task version checking for Ansible roles

#### Rationale:

- #2467 asks for a minimum version of Ansible 2.3 for remediation scripts. This adds a pre-task to check Ansible version in the roles that are generated.

### DIFF
--- a/shared/utils/build-all-remediation-roles.py
+++ b/shared/utils/build-all-remediation-roles.py
@@ -79,6 +79,19 @@ def generate_role_for_input_content(input_content, benchmark_id, profile_id, tem
     return ret
 
 
+def add_minimum_ansible_version(ansible_src):
+    pre_task = (""" - hosts: all
+   pre_tasks:
+     - name: Verify Ansible meets SCAP-Security-Guide version requirements.
+       assert:
+         that: "ansible_version.full | version_compare('2.3', '>=')"
+         msg: >
+           "You must update Ansible to at least version 2.3 to use this role."
+          """)
+
+    return ansible_src.replace(" - hosts: all", pre_task, 1)
+
+
 def get_cpu_count():
     try:
         return max(1, multiprocessing.cpu_count())
@@ -233,6 +246,10 @@ def main():
                 role_src = generate_role_for_input_content(
                     input_path, benchmark_id, profile_id, args.template
                 )
+
+                if args.extension == "yml" and \
+                   args.template == "urn:xccdf:fix:script:ansible":
+                    role_src = add_minimum_ansible_version(role_src)
                 with open(role_path, "w") as f:
                     f.write(role_src.encode("utf-8"))
 


shared/utils/build-all-remediation-roles.py
```diff
@@ -233,6 +250,9 @@ def builder():
                 role_src = generate_role_for_input_content(
                     input_path, benchmark_id, profile_id, args.template
                 )
+
+                if args.extension.endswith("yml"):
```
jan-cerny: I think it should check for `args.template == "urn:xccdf:fix:script:ansible"`  as well.

shared/utils/build-all-remediation-roles.py
```diff
@@ -79,6 +79,23 @@ def generate_role_for_input_content(input_content, benchmark_id, profile_id, tem
     return ret
 
 
+def add_minimum_ansible_version(ansible_src):
+    content = ""
+
+    for lines in ansible_src.split("\n"):
+        if lines == " - hosts: all":
```
jan-cerny: It splits a very big file to lines and then merges the lines together, line by line, using concatenation. That looks ineffective. Is it possible to just replace a substring by another substring in the playboook?

redhatrises:
Thanks for the review @jan-cerny! Made suggested changes.

shared/utils/build-all-remediation-roles.py
```diff
@@ -233,6 +248,10 @@ def builder():
                 role_src = generate_role_for_input_content(
                     input_path, benchmark_id, profile_id, args.template
                 )
+
+                if args.extension.endswith("yml") and \
```
mpreisler: if it's args.extension shouldn't we test for equality rather than endswith?

shared/utils/build-all-remediation-roles.py
```diff
@@ -79,6 +79,21 @@ def generate_role_for_input_content(input_content, benchmark_id, profile_id, tem
...
+    pre_task = (""" - hosts: all
+   pre_tasks:
+     - name: Verify Ansible meets SCAP-Security-Guide version requirements.
+       assert:
+         that: "ansible_version.full | version_compare('2.3', '>=')"
+         msg: >
+           "You must update Ansible to at least version 2.3 to use this role."
+          """)
+
+    content = ansible_src.replace(" - hosts: all", pre_task, 1)
```
mpreisler: just return...

shared/utils/build-all-remediation-roles.py
redhatrises: @mpreisler why wouldn't endswith work?

mpreisler: it would but equality is easier to read and makes more sense IMO.

It's sort of like `name="Gahbay"` vs `name.endswith("Gahbay")`. Both work but one makes more sense :-)

redhatrises:
Thanks for the review @mpreisler! Made the suggested changes.

I do agree that the Developer Guide needs to be updated for ansible remediations.
IMO, that should be handled fully in #2467. This is just a partial fix for generated remediations only.

mpreisler:
ACK

PR #2562, opened 2018-01-22
HTML table sanity check

mpreisler:
#### Description:

- Sanity check to make sure none of the generated HTML tables are empty

#### Rationale:

- It happened to us in the past that changes to an XSLT or XML input made some of the tables empty. Sometimes it takes a few weeks (or more) to notice that because nobody is checking all of the tables. This automated test should help with that.

Credit goes to @matusmarhefka , this is just a port of his downstream test.

### DIFF
--- a/cmake/SSGCommon.cmake
+++ b/cmake/SSGCommon.cmake
@@ -64,8 +64,8 @@ else()
     set(OSCAP_OVAL_SCHEMATRON_OPTION "")
 endif()
 
-set(SSG_LINKCHECKER_GUIDE_FILE_LIST "")
-set(SSG_LINKCHECKER_TABLE_FILE_LIST "")
+set(SSG_HTML_GUIDE_FILE_LIST "")
+set(SSG_HTML_TABLE_FILE_LIST "")
 
 macro(ssg_build_bash_remediation_functions)
     file(GLOB BASH_REMEDIATION_FUNCTIONS "${CMAKE_SOURCE_DIR}/shared/bash_remediation_functions/*.sh")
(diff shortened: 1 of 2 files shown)


cmake/SSGCommon.cmake
```diff
@@ -1004,27 +1003,39 @@ macro(ssg_build_html_stig_tables PRODUCT STIG_PROFILE DISA_STIG_VERSION)
...
+            COMMAND "${LINKCHECKER_EXECUTABLE}" --check-extern ${SSG_HTML_TABLE_FILE_LIST}
         )
     endif()
+
+    if (GREP_EXECUTABLE)
+        foreach(TABLE_FILE ${SSG_HTML_TABLE_FILE_LIST})
+            string(REPLACE "${CMAKE_BINARY_DIR}/tables/" "" TEST_NAME "${TABLE_FILE}")
+            # -z treats newlines as regular chars so we can match multi-line
+            # -v inverts the match, we are trying to make sure the tables don't
+            #    match this pattern
```
jan-cerny: This explanation is really perfect! I like it very much ❤️

cmake/SSGCommon.cmake
```diff
@@ -1004,27 +1003,39 @@ macro(ssg_build_html_stig_tables PRODUCT STIG_PROFILE DISA_STIG_VERSION)
...
     install(FILES "${CMAKE_BINARY_DIR}/tables/table-${PRODUCT}-stig.html"
         DESTINATION "${SSG_TABLE_INSTALL_DIR}")
     install(FILES "${CMAKE_BINARY_DIR}/tables/table-${PRODUCT}-stig-testinfo.html"
         DESTINATION "${SSG_TABLE_INSTALL_DIR}")
 endmacro()
 
-macro(ssg_define_linkchecker_tests)
-    if (SSG_LINKCHECKER_VALIDATION_ENABLED AND LINKCHECKER_EXECUTABLE)
+macro(ssg_define_guide_and_table_tests)
+    if (SSG_HTML_VALIDATION_ENABLED AND LINKCHECKER_EXECUTABLE)
```
jan-cerny: Where is `SSG_HTML_VALIDATION_ENABLED` defined?

mpreisler: Good catch, fixed!

cmake/SSGCommon.cmake
```diff
@@ -1004,27 +1003,39 @@ macro(ssg_build_html_stig_tables PRODUCT STIG_PROFILE DISA_STIG_VERSION)
...
+
+    if (GREP_EXECUTABLE)
+        foreach(TABLE_FILE ${SSG_HTML_TABLE_FILE_LIST})
+            string(REPLACE "${CMAKE_BINARY_DIR}/tables/" "" TEST_NAME "${TABLE_FILE}")
+            # -z treats newlines as regular chars so we can match multi-line
+            # -v inverts the match, we are trying to make sure the tables don't
+            #    match this pattern
+            add_test(
+                NAME "sanity-ssg-tables-${TEST_NAME}"
+                COMMAND "${GREP_EXECUTABLE}" "-zv" "</thead>[[:space:]]*</table>" "${TABLE_FILE}"
```
jan-cerny: It would be nice to redirect the grep output to /dev/null, now it prints the whole file on stdout, which pollutes ctest output.

PR #2565, opened 2018-01-22
Update rule_sshd_use_priv_separation test to check for sandbox value

dahaic:
#### Description:

- Correct value for `UsePrivilegeSeparation` is `sandbox`, not `yes`

#### Rationale:

- Few releases ago, OVAL and remediations has changed to accept `sandbox` and not `yes`. Tests has not been updated at that time.

- Test coverage for https://bugzilla.redhat.com/show_bug.cgi?id=1490792


jan-cerny:
This sounds to be reasonable.

PR #2566, opened 2018-01-23
Avoid 3 copy paste definitions of subprocess_check_output

mpreisler:
#### Description:

- I was trying to resolve all licensing loose ends and found this function with a LGPL compatibility comment. Turns out it was defined 3 times in the code-base.

#### Rationale:

- Comment about compatibility with BSD-3 makes more sense
- More definitions make it more difficult to maintain the code and fix it


jan-cerny:
Nice catch! Thank you.

PR #2567, opened 2018-01-24
Fixed the banner enablement option name.

matejak:
#### Description:

The ansible directive is wrong, so applying the playbook doesn't yield the expected result.

#### Rationale:

The directive is wrong according to https://help.gnome.org/admin/system-admin-guide/stable/login-banner.html.en and according to a test on a VM.


redhatrises:
Ack.

PR #2592, opened 2018-01-30
Make SSG easier to edit (the yaml project)

mpreisler:
#### Description:
Let's switch from the shorthand XML format to something similar to OpenControl's yaml format. This will make it easier for contributors to fix up rules or even add brand new rules. In this phase we will only switch to yaml and keep the XHTML descriptions and the other markup.

Future plans include moving to markdown and jinja2 templating.

Example benchmark format:
```
---
documentation_complete: true

title: Guide to the Secure Configuration of <product-name-macro/>

status: draft

description: |
    This guide presents a catalog of security-relevant
    configuration settings for <product-name-macro/>. It is a rendering of
    content structured in the eXtensible Configuration Checklist Description Format (XCCDF)
    in order to support security automation.  The SCAP content is
    is available in the config_item="scap-security-guide" package which is developed at
    <weblink-macro link="https://www.open-scap.org/security-policies/scap-security-guide" />.
    <br/><br/>
    Providing system administrators with such guidance informs them how to securely
    configure systems under their control in a variety of network roles. Policy
    makers and baseline creators can use this catalog of settings, with its
    associated references to higher-level security control catalogs, in order to
    assist them in security baseline creation. This guide is a italics="catalog, not a
    checklist," and satisfaction of every item is not likely to be possible or
    sensible in many operational scenarios. However, the XCCDF format enables
    granular selection and adjustment of settings, and their association with OVAL
    and OCIL content provides an automated checking capability. Transformations of
    this document, and its associated automated checking content, are capable of
    providing baselines that meet a diverse set of policy objectives. Some example
    XCCDF italics="Profiles", which are selections of items that form checklists and
    can be used as baselines, are available with this guide. They can be
    processed, in an automated fashion, with tools that support the Security
    Content Automation Protocol (SCAP). The DISA STIG for <product-name-macro/>,
    which provides required settings for US Department of Defense systems, is
    one example of a baseline created from this guidance.

notice:
    id: terms_of_use
    description: |
        Do not attempt to implement any of the settings in
        this guide without first testing them in a non-operational environment. The
        creators of this guidance assume no responsibility whatsoever for its use by
        other parties, and makes no guarantees, expressed or implied, about its
        quality, reliability, or any other characteristic.

front-matter: |
    The SCAP Security Guide Project<br/>
    <weblink-macro link="https://www.open-scap.org/security-policies/scap-security-guide" />

rear-matter: |
    Red Hat and Red Hat Enterprise Linux are either registered
[...]

PR #2596, opened 2018-01-31
[Do not merge] [WIP] Add XCCDF from YAML generator

jan-cerny:
#### Description:
This is @redhatrises 's script, but improved to handle the new format.
I worked on it with @matejak yesterday.
It doesn't work yet, the code needs a lot of refactoring and logic must be adjusted to the YAML format.

#### Rationale:

We need a script that will use the YAML format, that is proposed in https://github.com/OpenSCAP/scap-security-guide/pull/2592
This will allow us to use the new format in SSG build system.

Please this will change a lot.

### DIFF
--- a/shared/modules/xccdf_generator.py
+++ b/shared/modules/xccdf_generator.py
@@ -1,0 +1,348 @@
+import os
+import os.path
+import re
+import sys
+import argparse
+import datetime
+import yaml
+
+from xccdf_macros import *
+
+from collections import defaultdict
+
+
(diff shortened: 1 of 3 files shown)


mpreisler:
Why not push it to the yaml branch I created?

mpreisler:
Please push this commit to #2592 . We need to keep testing it there. This can never be merged as a separate PR.

jan-cerny:
Aha I have pushed it there.

PR #2598, opened 2018-01-31
cmake/SSGCommon.cmake: added check for override attribute

matusmarhefka:
#### Description:

- Checks that all profile titles and descriptions in generated XCCDF files contain `override="true"` attribute.

#### Rationale:

- This check should prevent us from releasing SSG with profiles which are extending other profile but have no `override` attribute set for their `title` and `description` tags.

### DIFF
--- a/chromium/profiles/stig-chromium-upstream.xml
+++ b/chromium/profiles/stig-chromium-upstream.xml
@@ -1,6 +1,6 @@
 <Profile id="stig-chromium-upstream" xmlns="http://checklists.nist.gov/xccdf/1.1" >
-<title>Upstream STIG for Google Chromium</title>
-<description>This profile is developed under the DoD consensus model and DISA FSO Vendor STIG process,
+<title override="true">Upstream STIG for Google Chromium</title>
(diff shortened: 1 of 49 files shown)


shawndwells:
@openscap-ci ok to test

shawndwells:
@openscap-ci add to whitelist

shawndwells:
Jenkins may need to be ignored on this one, as the patchset properly fails the unit tests.

The tests only fail if a full ``make`` is executed. Should/can the unit tests be updated to also fail builds when doing a targeted build (e.g. ``make rhel7``)?

dahaic:
@matusmarhefka as the tests are failing because of missing override tags on many profiles, can you fix it across whole SSG? I know the PR will get bigger, but that way, we won't introduce always failing check. No need to create new PR, just make this PR passing.

mpreisler:
I agree with @dahaic , we should fix this in all profiles before merging.

matusmarhefka:
The commit ae91818 sets `override="true"` for all profile titles and descriptions.

mpreisler:
Great stuff! This was a problem in multiple releases and now it can't happen any more!

PR #2599, opened 2018-02-01
Do not fail aide_scan_notification with other email adresses

jan-cerny:
#### Description:

The rule aide_scan_notification says that AIDE should notify appropriate
personnel of the details of an AIDE scan. The check currently requires
that the email address of the appropriate personnel starts with 'root@'.
In practice, the email address could be any email address. The check
should match any email address.

#### Rationale:
Fixes RHBZ#1540505


mpreisler:
`.*` will match 0 chars, shouldn't we be using `.+` at least?

jan-cerny:
@mpreisler aha, that sounds useful

matejak:
Great, I think that this is it.
Just for reference - writing a more specific validator would be highly demanding, because an e-mail address can have many many forms: https://en.wikipedia.org/wiki/Email_address

Issue #2604, opened 2018-02-12
Unanchored regular expression patterns in SCAP


solind:
#### Description of problem:

There are unanchored regular expression patterns in the SCAP content.  Specifically in the following objects:
- oval:ssg-obj_unauthenticated_apt_conf:obj:1
- oval:ssg-obj_unauthenticated_apt_conf_d:obj:1

And in the following variables:
- oval:ssg-var_rfp_all_log_files_as_string:var:1
- oval:ssg-var_rfg_all_log_files_as_string:var:1
- oval:ssg-var_rfo_all_log_files_as_string:var:1

#### SCAP Security Guide Version:

0.1.37

#### Operating System Version:

Ubuntu 14.04 (and probably also 16.04)

matejak:
Could you please elaborate what do you mean by unanchored? The `oval:ssg-obj_unauthenticated_apt_conf:obj` regexp (that can be found e.g. in the Ubuntu oval or datastream) is `^[\s]*APT::Get::AllowUnauthenticated(=|[\s]+)(yes|true|True);.*$`, which looks anchored to me.

solind:
Hi, look at the pattern in the filepath entity -- this is what an SCAP scanner must interpret to find the file location, and if it's not anchored, technically the entire filesystem would have to be searched (which is very expensive and time-consuming).

matejak:
Thank you for your feedback!
I see, it makes sense to add the beginning anchor e.g. to the `<ind:filepath operation="pattern match">/etc/apt/apt/apt.conf</ind:filepath>`, so it becomes  `<ind:filepath operation="pattern match">^/etc/apt/apt/apt.conf</ind:filepath>`, or even better `<ind:filepath operation="equals">/etc/apt/apt/apt.conf</ind:filepath>` as this one doesn't use the regex syntax.
This is altogether a little bit bizzare, as [the OVAL docs](https://oval.mitre.org/language/version5.11/ovaldefinition/documentation/independent-definitions-schema.html#textfilecontent54_object) suggest that `filepath` is the *absolute* path to a filename.
However, what about the situation with variables? The `<literal_component datatype="string">%/etc/rsyslog.conf</literal_component>` entry doesn't indicate that it is treated like a regexp (`%` seems to be just a delimiter). Do you have any indication that it actually is a regexp?

mpreisler:
Shouldn't it rather avoid using pattern match at all? shouldn't operation be "equal"?

[...]

PR #2605, opened 2018-02-13
Improve OVAL filepath expressions.

matejak:
This PR addresses #2604 .
The `filepath` can be either considered a
- regexp, in which case it should be anchored at the beginning, or a
- string, which should be treated as an absolute path specification.

Before this PR, `filepath` was considered a regexp and quite often lacked anchorage.


mpreisler:
@matejak please rebase to get rid of the merge commit.

matejak:
I have performed the rebase.

shawndwells:
On 2/13/18 7:58 PM, Matěj Týč wrote:
>
> This PR addresses #2604
> <https://github.com/OpenSCAP/scap-security-guide/issues/2604> .
> The |filepath| can be either considered a
>
>   * regexp, in which case it should be anchored at the beginning, or a
>   * string, which should be treated as an absolute path specification.
>
> Before this PR, |filepath| was considered a regexp and quite often
> lacked anchorage.
>

Thanks for doing this! Helps mature the content and was a long time coming.

(Catching up after 2 weeks PTO, hence the delayed thank you)

PR #2608, opened 2018-02-20
Revert "Grub2 password fix"

jan-cerny:
Reverts OpenSCAP/scap-security-guide#2606


tbrunell:
Is there something I can do to fix this PR?  Maybe not change the reference
file for the STIG setting?  Please let me know - I am trying to learn how
to be a good contributor.

Thanks

Ted

mpreisler:
@tbrunell Could you please open another PR with the same change but without the shared references change? We missed it in the review and now we are trying to straighten it all out :-) Thanks!

PR #2611, opened 2018-02-21
It's HIPAA, not HIPPA

mpreisler:
My OCD senses were tingling...


redhatrises:
Ack. Thanks for fixing my dyslexia ;)

PR #2612, opened 2018-02-21
update to match all supported EAP 6 releases

rlucente-se-jboss:
#### Description:

- Updated the product CPE lists to match all the released and supported product versions.

#### Rationale:

- I did this in an automated way to keep the bookkeeping straight.  Should be all cross-referenced correctly.  The official CPE list only goes up to EAP 6.4.18 since the 19 patch came out at the beginning of the month.


redhatrises:
openscap-ci add to whitelist

mpreisler:
Thank you @rlucente-se-jboss !

ACK

shawndwells:
@rlucente-se-jboss you mentioned this was generated w/automation. Is that automation something we could build into our CI/CD pipeline to keep these up to date?

rlucente-se-jboss:
This command would pull all the EAP 6.x release numbers in a way that works for anyone without passing through the paywall:

`curl -l ftp://ftp.redhat.com/redhat/jbeap/ | grep -E '6\.[0-9]+\.[0-9]+$'`

Then it's a matter of how to insert the right boilerplate into the different files.  Completely doable.  My approach was to have a simple text file with all the version numbers and then grep/sed/awk that into the files as needed.  Not fully automated but I avoided handjamming and potentially screwing up the version numbers.  This could be more automated.

rlucente-se-jboss:
This would eliminate the progress meter so cleaner:

`curl -ls ftp://ftp.redhat.com/redhat/jbeap/ | grep -E '6\.[0-9]+\.[0-9]+$'`

PR #2613, opened 2018-02-22
Clarify references to make it easier to understand which references map to the difference compliance organizations

redhatrises:
#### Description:

- Clarify references to make it easier to understand which references map to the difference compliance organizations

#### Rationale:

- A lot of compliance organizations use a similar number/decimal scheme e.g. 1.1.1 or 2.1.3 which makes it confusing on which benchmark (HIPAA, NIST 800-171, etc.) reference the rule applies. This hopefully fixes that issue.

### DIFF
--- a/shared/transforms/shared_shorthand2xccdf.xslt
+++ b/shared/transforms/shared_shorthand2xccdf.xslt
@@ -504,6 +504,19 @@
           <xsl:when test="name() = 'stigid'">
             <xsl:value-of select="normalize-space(concat($os-stigid-concat, .))" />
           </xsl:when>
+          <xsl:when test="name() = 'cis'">
+            <xsl:value-of select="normalize-space(concat('CIS ', $refitem))" />
+          </xsl:when>
+          <xsl:when test="name() = 'cjis'">
+            <xsl:value-of select="normalize-space(concat('CJIS ', $refitem))" />
+          </xsl:when>
+          <xsl:when test="name() = 'cui'">
+            <xsl:value-of select="normalize-space(concat('NIST 800-171 ', $refitem))" />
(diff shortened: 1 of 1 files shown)


redhatrises:
@shawndwells thoughts on this? Want to make sure that this does not break reference mapping in some way.

redhatrises:
@mpreisler does this break something somewhere else in the OpenSCAP ecosystem?

mpreisler:
> HIPPA

:beetle: :bug: :bug: :beetle:

mpreisler:
> @mpreisler does this break something somewhere else in the OpenSCAP ecosystem?

probably not but needs testing. Does this work well with the report's group by feature?

redhatrises:
@mpreisler luckily I only spelled it wrong in the PR description.

redhatrises:
@mpreisler it doesn't appear to, but it does show this in the Grouping by 

NIST SP 800-171 = **NIST 800-171 3.1.13**

mpreisler:
> @mpreisler it doesn't appear to, but it does show this in the Grouping by
> 
> NIST SP 800-171 = NIST 800-171 3.1.13

that sucks IMO. we may have similar assumptions in other GUIs

mpreisler:
@redhatrises what do you think about instead improving the UX of the HTML guide in openscap to show the type of reference more clearly?

redhatrises:
> @redhatrises what do you think about instead improving the UX of the HTML guide in openscap to show the type of reference more clearly?

@mpreisler I was thinking about that.

mpreisler:
This shouldn't go into SSG, this should be fixed in the tools.

PR #2614, opened 2018-02-22
Fix kernel module loading and unloading rules

dahaic:
#### Description:

- Fixed rule descriptions to make more sense, and remediation of general rule to set auditing of both b32 and b64 syscalls on 64-bit systems.

#### Rationale:

- Justification why 64-bit system needs both was not present in the descriptions, and everything was overly complicated because of furious copy-pasting.

- Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1547694


yuumasato:
As noted by @matejak, RHEL 6 should also be updated as it relies on Rule `audit_rules_kernel_module_loading`

shared/fixes/bash/audit_rules_kernel_module_loading.sh
```diff
@@ -5,11 +5,11 @@
...
 # Retrieve hardware architecture of the underlying system
-# Note: 32-bit kernel modules can't be loaded / unloaded on 64-bit kernel =>
-#       it's not required on a 64-bit system to check also for the presence
-#       of 32-bit's equivalent of the corresponding rule. Therefore for
-#       each system it's enought to check presence of system's native rule form.
-[ "$(getconf LONG_BIT)" = "32" ] && RULE_ARCHS=("b32") || RULE_ARCHS=("b64")
+# Note: 32-bit and 64-bit kernel syscall numbers not always line up =>
+#       it's required on a 64-bit system to check also for the presence
+#       of 32-bit's equivalent of the corresponding rule.
+#       (See `man 7 audit.rules` for details )
```
yuumasato: I really like the reference to audit manual.

dahaic:
Ah, sorry, missed the comment about RHEL6...

PR #2617, opened 2018-02-23
Fixes #2508

### LINKED ISSUE #2508: syntax errors in audit_rules_privileged_commands bash fix script cause the script to be non idempotent
#### Description of problem:
The remediation script for audit_rules_privileged_commands contains 2 syntax errors that cause it to be non idempotent and result in rules being added regardless of whether they already exist.

The issue is with this block of code:
```
base_search=$(sed -e "/-a always,exit/!d" -e "/-F path=${sbinary_esc}$/!d"   \
                                  -e "/-F path=[^[:space:]]\+/!d" -e "/-F perm=.*/!d"       \
                                  -e "/-F auid>=${min_auid}/!d" -e "/-F auid!=4294967295/!d"  \
                                  -e "/-k privileged/!d" $afile)
```
Problems:
1. `${sbinary_esc}$` does not require the 2nd dollar sign
2. enclosing `!d` in double quotes causes bash to see those characters as an attempt at history expansion, which attempts to replay the last command that started with d.

0.1.36-1.el7.centos

#### Steps to Reproduce:

1. Run the audit_rules_privileged_commands remediation bash script on a system which already has one or more audit rules for suid/sgid binaries

#### Actual Results:
Rules are added to /etc/audit/audit.rules for all suid/sgid binaries

#### Expected Results:
Rules are only added for the suid/sgid binaries that do not already contain rules in /etc/audit/audit.rules

### Suggested Fix:

1. Add `set +H` at the top of the script, which disables history expansion
2. Update the base_search expression to remove the unnecessary dollar sign after the sbinary_esc variable:
```
base_search=$(sed -e "/-a always,exit/!d" -e "/-F path=${sbinary_esc}/!d"   \
                                  -e "/-F path=[^[:space:]]\+/!d" -e "/-F perm=.*/!d"       \
                                  -e "/-F auid>=${min_auid}/!d" -e "/-F auid!=4294967295/!d"  \
                                  -e "/-k privileged/!d" $afile)
```

OnceUponALoop:
#### Description:

Updated sed logic to avoid bash history expansion issue. Use single quotes and expand only on variables.

#### Rationale:

Fixes #2508

### DIFF
--- a/shared/bash_remediation_functions/perform_audit_rules_privileged_commands_remediation.sh
+++ b/shared/bash_remediation_functions/perform_audit_rules_privileged_commands_remediation.sh
@@ -90,10 +90,10 @@ do
 		# * existing rule contains all arguments from expected rule form (though can contain
 		#   them in arbitrary order)
 	
-		base_search=$(sed -e "/-a always,exit/!d" -e "/-F path=${sbinary_esc}$/!d"   \
-		    		  -e "/-F path=[^[:space:]]\+/!d" -e "/-F perm=.*/!d"       \
-				  -e "/-F auid>=${min_auid}/!d" -e "/-F auid!=4294967295/!d"  \
-				  -e "/-k privileged/!d" $afile)
+		base_search=$(sed -e '/-a always,exit/!d'   -e '/-F path='"${sbinary_esc}"'/!d' \
+		    	  -e '/-F path=[^[:space:]]\+/!d'   -e '/-F perm=.*/!d'                 \
+				  -e '/-F auid>='"${min_auid}"'/!d' -e '/-F auid!=4294967295/!d'        \
+				  -e '/-k privileged/!d' $afile)
 
 		# Increase the count of inspected files for this sbinary
 		count_of_inspected_files=$((count_of_inspected_files + 1))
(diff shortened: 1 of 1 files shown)


jan-cerny:
@openscap-ci ok to test

matejak:
Just a minor remark: I see that the line beginning with `e '/-F path=[^[:space:]]` that lost the leading indentation, and the next line got some whitespaces before backslashes at the end of the line, could you please check this out? This was messed up before you PR - as rest of the file is tab-based, I feel that this part also shouldn't mix tabs with spaces, and I don't think that we need those whitespaces by the line ends.

OnceUponALoop:
@matejak I'll fix it up soon. Since you brought it up, is there a project standard or contributor documentation?
I had initially converted it all from tabs to spaces because the different usage was driving me nuts but since i'm a new contributor I didn't want the PR to be a mess and restricted my change to the bug fix.

OnceUponALoop:
@matejak I fixed the indentation but kept the trailing spaces. It's not presented correctly in the github preview but it's just to align the bash line continuation slashes

should look like this
``` bash
        base_search=$(sed -e '/-a always,exit/!d' -e '/-F path='"${sbinary_esc}"'/!d' \
                -e '/-F path=[^[:space:]]\+/!d'   -e '/-F perm=.*/!d'                 \
                -e '/-F auid>='"${min_auid}"'/!d' -e '/-F auid!=4294967295/!d'        \
                -e '/-k privileged/!d' $afile)
```

matejak:
Thank you for the fix! Regarding the style, there is no agreed-upon style for bash scripts, so the most meaningful thing to do is to respect the local style. Therefore, if you edit a script in a place where there is a 4-spaces indentation, use that one and if another script uses tabs, use tabs.
However, keep in mind that different editors may have different default tab width, so the text may look differently.

PR #2619, opened 2018-02-23
Fix #2618

### LINKED ISSUE #2618: bootloader_password check - two valid states after documentation update
#### Description of problem:
Failed rule : xccdf_org.ssgproject.content_rule_bootloader_password

Git master (bb897175d0695dd6b6ec434e727f66fb6eb13eea)

RHEL 7.4 

#### Steps to Reproduce:

1. Set grub password using the `grub2-password` tool
2. Rename grub root user `sed -i s/root/ricksanchez/g /etc/grub.d/01_users`
3. Evaluate rule

#### Actual Results:
```
[root@rh-test build]# oscap oval eval --verbose INFO --id oval:ssg-bootloader_password:def:1 ssg-rhel7-oval.xml
--- Relevant output ----
I: probe_textfilecontent54: Opening file '/boot/grub2/user.cfg'.
I: probe_textfilecontent54: Extracting item from the cache queue: cnt=1, beg=0
I: oscap:     Test 'oval:ssg-test_bootloader_password:tst:1' requires that every object defined by 'oval:ssg-object_bootloader_password:obj:1' exists on the system.
I: oscap:     0 objects defined by 'oval:ssg-object_bootloader_password:obj:1' exist on the system.
I: oscap:     Test 'oval:ssg-test_bootloader_password:tst:1' does not contain any state to compare object with.
I: oscap:     No item matching object 'oval:ssg-object_bootloader_password:obj:1' was found on the system. (flag=does not exist)
I: oscap:   Test 'oval:ssg-test_bootloader_password:tst:1' evaluated as false.
```

#### Expected Results:
Test should pass

The oval check `bootloader_password` has an incorrect pattern match that doesn't match the file contents.

`/boot/grub2/user.cfg` contents after calling `grub2-password` tool
```
GRUB2_PASSWORD=grub.pbkdf2.sha512.<<TRUNCATED-HASH>>
```

Pattern match being used will not match the file content
```
^[\s]*password_pbkdf2[\s]+.*[\s]+grub\.pbkdf2\.sha512.*$
```

Should be updated to
```
^[\s]*GRUB2_PASSWORD=grub\.pbkdf2\.sha512.*$
```

This seems to be a result of a documentation update without updating the corresponding pattern match.

This leaves us with 2 valid states
- `/etc/grub.d/01_users` with the `password_pbkdf2 <USER> grub.pbkdf2.sha512<HASH>` format
- `/boot/grub2/user.cfg` with the `GRUB2_PASSWORD=grub.pbkdf2.sha512<HASH>` format

Each state will need it's own checks unless we want to invalidate the previous recommendation but that would require manual intervention on previously deployed systems.

OnceUponALoop:
#### Description:

Update the check to verify if password exists in either `/boot/grub2/grub.cfg` or `/boot/grub2/user.cfg`.
They're both valid configurations, the documented remediation was recently updated from modifying `/boot/grub.d/01_users` to using the `grub2-setpassword` tool.

#### Rationale:

Documented remediation process conflicts with previously recommended remediation. This left us with 2 valid configurations. 

Fixes #2618

### DIFF
--- a/shared/checks/oval/bootloader_password.xml
+++ b/shared/checks/oval/bootloader_password.xml
@@ -11,7 +11,10 @@
     <criteria operator="OR">
       <criterion comment="Pass if /boot/grub2/grub.cfg does not exist" test_ref="test_bootloader_grub_cfg" />
       <criteria operator="AND">
-        <criterion comment="make sure a password is defined in /boot/grub2/user.cfg" test_ref="test_bootloader_password" />
+        <criteria comment="check both files to account for procedure change in documenation" operator="OR">
+            <criterion comment="make sure a password is defined in /boot/grub2/user.cfg" test_ref="test_bootloader_password_usercfg" />
+            <criterion comment="make sure a password is defined in /boot/grub2/grub.cfg" test_ref="test_bootloader_password_grubcfg" />
+        </criteria>
         <criterion comment="make sure a superuser is defined in /boot/grub2/grub.cfg" test_ref="test_bootloader_superuser"/>
       </criteria>
     </criteria>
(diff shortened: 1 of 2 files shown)


mpreisler:
@openscap-ci test this please

shared/checks/oval/bootloader_password.xml
```diff
@@ -11,7 +11,10 @@
     <criteria operator="OR">
       <criterion comment="Pass if /boot/grub2/grub.cfg does not exist" test_ref="test_bootloader_grub_cfg" />
       <criteria operator="AND">
-        <criterion comment="make sure a password is defined in /boot/grub2/user.cfg" test_ref="test_bootloader_password" />
+        <criteria comment="check both files to account for procedure change in documenation" operator="OR">
```
yuumasato: The password should be configured in both places, right?
So operator should be `AND`.

yuumasato: I was confused, never mind.

yuumasato:
@OnceUponALoop Should the same fix be applied to Rule `bootloader_uefi_password`?

OnceUponALoop:
@yuumasato Good catch, It's likely, i overlooked it since i was working with a bios booted VM.

I'll verify and update the PR as needed.

yuumasato:
Ack, thanks for updating the PR.

Issue #2589, opened 2018-01-28
Gnome3 banner : Incorrect ansible remediation and OVAL check


ptitoliv:
When trying to add the DoD banner as login banner text on GDM, I identified the following issues : 
 - ansible remediation is not working because the login banner text is the regexp stored in the XCCDF file and GDM doesn't handle such a format. So when, the remediation is applied, the result is an ugly display like that : ![gnome_banner_ansible_remediation](https://user-images.githubusercontent.com/567505/35482115-bd1307aa-0430-11e8-9105-a64071866a95.png)
 - I wrote a bash remediation (https://gist.github.com/ptitoliv/bc7f2aff8c550edb37268cc8ecc9b6f5) taking the banner stored into the XCCDF doing a small transformation on the regexp in order to have a correct display. The display seems fine with that transformation : 
![gnome_banner_bash_remediation](https://user-images.githubusercontent.com/567505/35482133-036ec162-0431-11e8-8ca9-c723fe056a88.png). However, the thing is that the OVAL check is not working any more because of the mandatory \n we have to put in the GDM banner message text in order to have a better display. 

For the last point I have two questions : 
 - Should we have to create a specific variable with the correct regexp format in order to handle this specific use case.
 - I tried to make some tests but they were unsuccessful. The thing is that for testing the regexp, I didn't find any other solution than rebuild the XCCDF and do a check which is very long. Is there some tricks in order to validate these regexp without such a long process ?

Context : SSG master branch on CentOS 7.

ptitoliv:
Just started some WIP on that issue on branch https://github.com/ptitoliv/scap-security-guide/tree/gnome_banner. It seems I have now a correct bash and remediation for that issue. I checked only for the dod_banners right now. What do you think ?

redhatrises:
Duplicating the banner text is not the right way to go as it not only creates confusion but it increases maintenance costs as well as duplicates content which is not what we want to do.

ptitoliv:
@redhatrises : Just updated the branch with a common banner usable for both checks and remediations. I had to update /etc/issue remediation in order to handle that new regexp.

ptitoliv:
Hello team,
Any news about that review ? Would you prefer me to open a PR in WIP mode ?

matejak:
[...]

PR #2626, opened 2018-02-24
Correct display for Gnome Banner

ptitoliv:
#### Description:

This PR is a work in progress in order to propose a correct display of the banners on the GDM login screen.  On this PR, I update the remediation script which configures GDM in order to display the banner and I also update the dod_banner regexp in order to have a compliant banner which works for /etc/issue and also GDM.

Warning : For this preview, the work has only be done for the dod_banner.

#### Rationale:

Propose a fix for #2589.


matejak:
It may take some time before we merge it, as we are not sure if we can change the `dod_banners` as it is mandated by regulations. @redhatrises any thoughts of that?
Next, although operations with regular expressions may be technically correct, they are mind-bending. Understanding them is not trivial, which increases maintenance costs. So could you please provide comments concerning what the individual `sed` commands in the "expansion" line in Bash do?
And as you seem to have a good insight into the state of things prior to your contribution, could you please add explanation comments into the `banners.xml` file that would explain why the pattern was quite complicated already and what is the intention behind your changes?

shawndwells:
Regarding to banner text: It might be time to revisit/clarify what DISA wants for their DoD banner. The DISA-provided text in the STIG conflicts with the DISA-provided guidance and DSS mandated text.

DISA published the following guidance relating to banners:
https://iase.disa.mil/Documents/unclass-faq-dod_notice_and_consent_banner.docx

Which points to this URL to get the banner text from:
http://www.dss.mil/documents/odaa/dod_warning_banner.pdf

... and that language is different than what we've been told to use in the STIG.

@tbrunell could you run this down for us? Which banner should we use?

ptitoliv:
@matejak : Thanks for your review. Let me try to answer all the points you raised in your comment. 

> It may take some time before we merge it, as we are not sure if we can change the dod_banners as it is mandated by regulations. 

Actually, I didn't modify the banner text it self. I just updated the regular expression stored into the login_banner variable in banners.xml. The generated text when the remediation is applied using the SED command is normally the one provided in the STIG Guide.  So I guess, we're still compliant with the STIG.

> Next, although operations with regular expressions may be technically correct, they are mind-bending. Understanding them is not trivial, which increases maintenance costs. So could you please provide comments concerning what the individual sed commands in the "expansion" line in Bash do?

I agree with you : Regular expression can be very difficult to understand and i'm not the best expert for writing the beautifulest regular expression.  Before explaining what my SED command does, let me explain how and why I rewrote the regular expression for the banner. 

Actually, it is explained on the ticket #2589 I opened : The current regular expression used to generated the banner can't be used for the GDM login banner because it is not possible to have a string with newlines in the GDM configuration files ==> You have to describe the newline using \n\n. So I updated the banner text adding `(\\n)*(\n)*` where a new line should be created in order to have a nice display for the banner in both mode (/etc/issue and GDM display).

Then I created the remediation `script rhel7/fixes/bash/dconf_gnome_login_banner_text.sh` with that SED expression which transforms the `(\\n)*(\n)*` into `\n\n` string compliant with the string format used in GDM configuration files.  This is done by the code `;s/\\//g;s/(n)\*/\\n/g'` at the end of the REGEXP. 

To finish I updated the regular expression in the file `rhel7/fixes/bash/banner_etc_issue.sh` in order to generate the same /etc/issue banner than before replacing  `(\\n)*(\n)*` by real `\n\n' which is the allowed format for /etc/issue file.

> And as you seem to have a good insight into the state of things prior to your contribution, could you please add explanation comments into the banners.xml file that would explain why the pattern was quite complicated already and what is the intention behind your changes?

My first intention was to add the missing remediation for dconf_gnome_login_banner_text check which wasn't implemented in SSG in bash mode. There is only an ansible remediation which produces the ugly displayed as shown in #2589.  So I wanted to have a real bash remediation in order to have the correct display with GDM login. My first idea was to have a dedicated banner but @redhatries told me on the ticket that it wasn't the good way to proceed so that's why I changed the banner regular expression in order to have one usable by the both bash remediation script. For me the banner regexp is complicated because it considers we can have a newline after each word. I wonder if it wouldn't be simplier to have a classical sentence adding just newline characters (\n) on the right places for a good display. But maybe there is some regulation stuff I'm not aware about. 

Hope, I have been clear. Don't hesitate to ping me if you have other questions.

matejak:
@ptitoliv Thank you for your explanation. I get it and I think that we can adopt your fix, although we will probably have to approach the problem differently later, when we will want the Ansible remediation to work.
I have just one question - why you have used `(\\n)*(\n)*` to mark the place where `\n\n` will appear later in the process? I see that it works, but I also think that `(\n)*(\n)` would work in the same way. Is there some reason behind it?

ptitoliv:
@matejak `(\\n)*(\n)*` is because I want to use the same banner for /etc/issue and GDM banner as @redhatrises requested. So I must use `\\n` in stead of `\n` only for the GDM banner check ==> If I don't do that, regexp check will try to search for a real newline instead of the `\n`character mandatory for the GDM banner.

If you're okay with my explainations, does that mean I can do the same work for the two others banners considering that PR is going to be merged ?

matejak:
@ptitoliv Thank you for your explanation, this makes perfect sense. I am pretty sure that you can go ahead with the other banners. As your PR doesn't break anything and it fixes the bash remediation, I am sure that it will get merged soon after it leaves the WIP status.

ptitoliv:
@matejak : Here is the first version with updated banners but also remediations scripts with new transformations in order to handle correctly all the cases considering using a unique banner. You can also find with that PR a ZIP containing screenshot about the produced banners with the scripts available on this branch.
[banners_screenshot.zip](https://github.com/OpenSCAP/scap-security-guide/files/1791399/banners_screenshot.zip)

matejak:
I have verified that it works on a RHEL7 system. Olivier, thank you very much for your awesome contribution!

PR #2629, opened 2018-02-27
Improved bash syntax of bash remediations

matejak:
Notes on all changes:
* Implemented fixes suggested by `shellcheck`.

Notes on the audit rules-related remediation:
* Used custom sed delimiters to get rid of escaping forward slashes.
* ~Can't be merged before #2617 is dealt with.~

Notes on the NTP-related and accounts_passwords_pam_faillock_...-related parts:
* Some functionality has been deduplicated.
* Functionality that could be shared between remediations has been isolated into the `handle_ntp_like_file` and `substitute_in_pam_file` functions and moved to shared remediation functions.


matejak:
I have moved some functions to the `shared/bash_remediation_functions`. However, I have encountered these issues:
* `substitute_in_pam_file` function is placed immediately before its invocation, which is inside a `for` loop.
* `rhel7_handle_ntp_like_file` function is not placed into the remediation script at all. I have examined how the `rhel7_fedora_perform_audit_adjtimex_settimeofday_stime_remediation` function is handled, but although I believe that I have done everything the same way, definition of `rhel7_handle_ntp_like_file` is still missing from the built remediation role.

yuumasato:
> substitute_in_pam_file function is placed immediately before its invocation, which is inside a for loop.

At https://github.com/OpenSCAP/scap-security-guide/blob/master/shared/utils/combine-remediations.py#L349 we append a `<sub>` element which is substituted later on by build system.

> definition of rhel7_handle_ntp_like_file is still missing from the built remediation role.

Regex in https://github.com/OpenSCAP/scap-security-guide/blob/master/shared/utils/combine-remediations.py#L260 only expects newline or whitespace before remediation function.

matejak:
Awesome, thank you @yuumasato for explanation. Applying workarounds is quite easy.

shared/bash_remediation_functions/substitute_in_pam_file.sh
```diff
@@ -0,0 +1,32 @@
+function substitute_in_pam_file {
+    # If invoked with no arguments, exit. This is an intentional behavior.
+    [ $# -gt 1 ] || return 0
+    [ $# = 3 ] || die "$0 requires zero or exactly three arguments"
+	local _pamfile="$1" _directive="$2" _value="$3"
```
yuumasato: From `pam_faillock` manual, the `directives` are called `options`.

shared/bash_remediation_functions/substitute_in_pam_file.sh
```diff
@@ -0,0 +1,32 @@
+function substitute_in_pam_file {
```
yuumasato: This function is specific for configuring `pam_faillock` module, it should be named according to it.

shared/bash_remediation_functions/rhel7_handle_ntp_like_file.sh
```diff
@@ -0,0 +1,16 @@
+# Function ensures that the ntp/chrony config file contains valid server entries
+# $1: Path to the config file
+# $2: Comma-separated list of servers
+function rhel7_handle_ntp_like_file {
```
yuumasato: The function name should be more specific to what it does. Like adding or enabling the list of servers in ntp like files.

shared/bash_remediation_functions/rhel7_handle_ntp_like_file.sh
```diff
@@ -0,0 +1,16 @@
...
+    # If invoked with no arguments, exit. This is an intentional behavior.
+    [ $# -gt 1 ] || return 0
+    [ $# = 2 ] || die "$0 requires zero or exactly two arguments"
+	local _config_file="$1" _servers_list="$2"
+	if ! grep -q '#[[:space:]]*server' "$_config_file"; then
+		for server in $(echo "$_servers_list" | tr ',' '\n') ; do
+			printf '\nserver %s iburst' "$server" >> "$_config_file"
+		done
+	else
+		sed -i 's/#[ \t]*server/server/g' "$_config_file"
```
yuumasato: This remediation looks misleading:
If the file doesn't have any `server` entry, the list of servers from variable received is added.
If there are commented `server` entries, the remediation just uncomments it, and no check is done regarding the servers itself.

Also, OVAL definition doesn't care about which servers are configured.
If we set the variable for the list of servers, we might as well be checking for them.
For this PR we can keep original behavior, but we can improve it according to expectations on variable.

mpreisler:
@matejak It's not a big deal but please rebase instead of merging master into the PR. It makes the history less tangled up.

matejak:
I have rebased, deduplicated one more low hanging fruit and renamed some of the functions and variables.

yuumasato:
I've tested rules `pam_faillock_deny` and `chronyd_or_ntpd` with SSG Test Suite and everything works.
Thank you, for improving our bash remediations.

@matejak I'm afraid you'll need to rebase again.

matejak:
Thank you for review, I have performed the rebase.

redhatrises:
Great job @matejak!

PR #2633, opened 2018-02-28
Introduced draft of SSG Bash scripting guidelines.

matejak:
As @OnceUponALoop  in #2617 noticed, it is not clear what style and approach to prefer when contributing to Bash remediations.
I have went to the existing remediations in the `shared/fixes/bash` folder and observed the prevalent existing style. I have found out two points which I believe that are not controversial at all:
* most scripts use tabs for indentation, so let's use tabs.
* usage of `sed` greatly surpasses `awk`, so let's encourage `sed`.

Next, I propose these:
* Use the `die` function [idiom](https://stackoverflow.com/q/7868818/592892) to handle possible errors (I have defined it and checked that it can be actually used).
* Run `shellcheck` over the snippet and try to fix all warnings.
* Avoid double negatives s.a. `if ! [ $x != 1 ]; then ...`

Finally, some points may be controversial:
* I have slightly encouraged usage of compound statements. I think that if they are used in moderation, they can make the code much shorter, without obscuring anything.
* I have suggested to write scripts in a way that they work if invoked in the "strict mode" e.g. by running them using `bash -e` or by writing `set -e` to the top of the script. This imposes greater requirements on error handling inside the script and therefore increases its robustness.

### DIFF
--- a/docs/manual/developer_guide.adoc
+++ b/docs/manual/developer_guide.adoc
@@ -730,11 +730,21 @@ Following, you can see an example of a bash remediation that sets the maximum nu
 populate var_accounts_maximum_age_login_defs
 
 grep -q ^PASS_MAX_DAYS /etc/login.defs && \
-  sed -i "s/PASS_MAX_DAYS.*/PASS_MAX_DAYS     $var_accounts_maximum_age_login_defs/g" /etc/login.defs
-if ! [ $? -eq 0 ]; then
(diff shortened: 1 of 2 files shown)


mpreisler:
ACK, great stuff!

shawndwells:
Agreed, thank you for doing this @matejak!

PR #2641, opened 2018-03-02
Add override to C2S-docker Profile

yuumasato:
#### Description:

- Override title and description of C2S Docker Profile.

#### Rationale:

- Default is to override these fields in all Profiles, so that Profiles that extend another don't forget to do so.
- There is no harm in overriding even if not extending any Profile.


redhatrises:
Ack

yuumasato:
That was fast, thanks!

yuumasato:
Self-merging for release, this has been acked by @redhatrises.

PR #2644, opened 2018-03-05
Fix a mismatched tag issue in RHEL6 sudo.xml

mpreisler:
The mismatched tag prevents the yaml dump script from working.

This is clearly a typo. I am puzzled how no part of our build system / test system spotted this until now.


redhatrises:
Ack.

PR #2646, opened 2018-03-05
Chromium XCCDF was in shared even though it uses nothing else from sh…

mpreisler:
…ared

And nothing else uses the "shared" chromium xml.


redhatrises:
Ack.

PR #2647, opened 2018-03-05
Moved firefox shorthand XML to the firefox product folder from shared

mpreisler:
Nothing in it is actually shared...


redhatrises:
Ack.

PR #2648, opened 2018-03-05
Untangle shared shorthands

mpreisler:
Moved some of the XMLs into right places. This is just the easy stuff unfortunately. The difficult stuff will be a huge pain - ubuntus, fedora, rhel6, wrlinux are all total mess with mixed bits from all over the repo.


redhatrises:
ack.

PR #2649, opened 2018-03-05
Remove the empty RHEVM3 benchmark

mpreisler:
It's broken and the cost of trying to make it work is too high. It only
has a sample group and a sample rule. If somebody wants to reintroduce
it please make it contain something, otherwise it can't be accepted into
SSG.

RHEVM3 has been disabled ever since the cmake port which was over a year ago. Even before it was not working and didn't have any rules in it.

This may be controversial so please leave this open until at least 2 people ack it.

(this is blocking the yaml work)


redhatrises:
ack

mpreisler:
@openscap-ci test this please

yuumasato:
Ack.

PR #2652, opened 2018-03-06
eaccess should actually be eacces

sean797:
#### Description:

eaccess should be eacces. See `man audit.rules`

#### Rationale:

I guess this was a typo :wink:


mpreisler:
Thank you! ACK

PR #2656, opened 2018-03-07
Remove Webmin content

redhatrises:
- Remove Webmin content form SSG
- The Webmin content is not really being developed, is out-of-date, and in varying states of completeness.

This may be controversial so please leave this open until at least 2 people ack it.


mpreisler:
ACK from me

matejak:
ACK, we have agreed on removing the module, and there seems to be no trace of `webmin` left in the branch.

PR #2661, opened 2018-03-08
Source only local shorthand XCCDF to build debian8 content

yuumasato:
#### Description:

- Copy shared xccdf content used by debian8  to its local folders
- Build shorthand using local content only

#### Rationale:

- This will help untangle where the shortand XCCDF are sourced from to continue yaml  work.
- Later debian8 content can be ported to shared.


redhatrises:
Ack.

PR #2662, opened 2018-03-08
Move debian8 and rhel6 system and services locally

yuumasato:
#### Description:

- Copy shared system and services xccdf content used by debian8 and rhel6 to its local folders
- Build shorthand using local content only

#### Rationale:
- This will help untangle where the shortand XCCDF are sourced from to continue yaml work.

### DIFF
--- a/debian8/guide.xslt
+++ b/debian8/guide.xslt
@@ -35,8 +35,8 @@
       <xsl:apply-templates select="document(concat($BUILD_RP, '/bash-remediation-functions.xml'))" />
 
       <xsl:apply-templates select="document(concat($SHARED_RP, '/xccdf/intro/shared_intro_os.xml'))" />
-      <xsl:apply-templates select="document(concat($SHARED_RP, '/xccdf/system/system.xml'))" />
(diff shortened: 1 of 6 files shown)


debian8/guide.xslt
```diff
@@ -34,9 +34,9 @@
 
       <xsl:apply-templates select="document(concat($BUILD_RP, '/bash-remediation-functions.xml'))" />
 
-      <xsl:apply-templates select="document(concat($SHARED_RP, '/xccdf/intro/shared_intro_os.xml'))" />
```
redhatrises: @yuumasato this is the only thing to not change.

yuumasato: @redhatrises Can you explain me why?

redhatrises: We need the XCCDF that contains the shorthand groups which contain the rules. The shared intro has been migrated for all platforms and isn’t the issue that we are coming across in the yaml migration.

mpreisler: +1, at this point don't move the shared intro, I will either deal with it in the yaml dumper script or I will copy it later.

yuumasato: Okay, updated.

mpreisler:
ACK

PR #2663, opened 2018-03-09
[WIP] Jre update and fix

dahaic:
#### Description:

- Slight update of Java JRE profile, updating reference and making fixes in line with rest of the project.

#### Rationale:

- Triggered by bugzilla, there was some misleading content, outdated remediations, some broken.

- Fixes RHBZ#1553378


yuumasato:
I'm a bit confused. Both `disa-stig-jre7-unix-v1r4-xccdf-manual.xml` and `disa-stig-jre7-unix-v1r6-xccdf-manual.xml` say at [line125](https://github.com/OpenSCAP/scap-security-guide/blob/master/shared/references/disa-stig-jre7-unix-v1r4-xccdf-manual.xml#L125) that config should be `deployment.system.config.mandatory=false`.

jre/xccdf/java.xml
```diff
@@ -90,6 +90,11 @@ To configure the Java mandatory deployment setting, add or modify
 <tt>deployment.system.config.mandatory</tt> to equal <tt>true</tt>
 in <tt>/etc/.java/deployment/deployment.config</tt>.
 </description>
+<warning category="general">This recommendation is opposite of what is in DISA
```
redhatrises: @dahaic rather than add this comment. Let's do a complete update of the JRE content and remove STIG deprecated products such as JRE6 and JRE7.

dahaic: @redhatrises JRE8 STIG is quite different so the changes will be (relatively) extensive - I'll do it on top of these commits, though, so I can reference fix in BZ. Is it ok?

dahaic:
I am going to update to JRE8, until then, this is not to be merged :) (It might take few days)

yuumasato:
@dahaic This will generate conflicts with the yaml port effort.
Do you plan to finish this before conversion happens, or are you ok with reworking when port happens?

dahaic:
@yuumasato right, it does not make sense to keep it up, when I need to do complete rework anyway. I am closing the PR for now.

PR #2664, opened 2018-03-12
Fix Rule ensure logrotate activated

yuumasato:
#### Description:

- Add test scenarios for Rule ensure_logrotate_activated
- Update OVAL definition to:
  - Allow other arguments to logrotate command
  - Ensure that `/etc/logrotate.conf` is the last config file
  - Check for both `logrotate.conf` AND `cron.daily`
- Add bash remediation


shared/fixes/bash/ensure_logrotate_activated.sh
```diff
@@ -0,0 +1,7 @@
+# platform = multi_platform_rhel
+
+# daily rotation is configured
+grep -q "^daily$" /etc/logrotate.conf || echo "daily" >> /etc/logrotate.conf
+
+# remove any line configuring weekly, monthly or yearly rotation
+sed -i -r "/^(weekly|monthly|yearly)$/d" /etc/logrotate.conf
```
mpreisler: this is not adding logrotate to crontab

yuumasato: Hmm, so just configuring `logrotate.conf` is not enough?

mpreisler: I don't know but I thought you have to call `/sbin/logrotate` periodically for the config file to even be used.

yuumasato: It makes sense, updated everything to account for `logrotate.conf` and `cron.daily`.

mpreisler: Wait, did you verify that? Because I am not sure about what I said.

yuumasato: Yes, `man logrotate` states the following
> Normally, logrotate is run as a daily cron job.

yuumasato: Also from man page:
> hourly       Log files are rotated every hour. Note that usually logrotate is configured to be run by cron daily. You have to change this configuration and run logrotate hourly to be able to really rotate logs hourly.

Configuring frequency of logrotation in `logrotate.conf` doesn't account for frequency of `logrotate` runs. The cron job also needs to be configured.

mpreisler:
ACK

PR #2665, opened 2018-03-14
Add support for Oracle Linux 7

iokomin:
#### Description:

Suggested initial support for Oracle Linux 7 product.

Existing rhel7 and shared content reused as much as possible to cover standard profile content and make ol7 project to pass build and build tests.

Affected changes:
- Added ol7 product build structure and Oracle Linux 7 cpe content
- Created ol7 'standard' profile reflecting rhel7 'standard' profile rules set
- Added ensure_oracle_gpgkey_installed oval, xccdf and fix content, other definitions and templates copied from rhel7 product without changes due to RHEL7 and OL7 compatibility
- Added ol7 product platform to affected shared oval, xccfd rules and fix scripts
- Checked build and tests to pass, using oscap tool validated generated ssg-ol7-* xml files, also checked generated guide content and evaluation results for 'standard' profile.

#### Rationale:

- Add Oracle Linux platform specific SCAP profiles support.

### DIFF
--- a/ol7/checks/oval/file_permissions_unauthorized_sgid.xml
+++ b/ol7/checks/oval/file_permissions_unauthorized_sgid.xml
@@ -3,7 +3,6 @@
     <metadata>
       <title>Find setgid files system packages</title>
       <affected family="unix">
-        <platform>Red Hat Enterprise Linux 7</platform>
         <platform>Oracle Linux 7</platform>
       </affected>
       <description>All files with setgid should be owned by a base system package</description>
@@ -29,7 +28,7 @@
     <unix:sgid datatype="boolean">true</unix:sgid>
   </unix:file_state>
 
-  <!-- list of all setgid files included with base RHEL7 system -->
+  <!-- list of all setgid files included with base OL7 system -->
   <unix:file_state id="state_sgid_whitelist" version="1">
     <unix:filepath var_ref="var_sgid_whitelist" var_check="at least one" />
   </unix:file_state>
--- a/ol7/checks/oval/file_permissions_unauthorized_sgid.xml
+++ b/ol7/checks/oval/file_permissions_unauthorized_sgid.xml
@@ -1,0 +1,79 @@
+<def-group>
+  <definition class="compliance" id="file_permissions_unauthorized_sgid" version="2">
+    <metadata>
+      <title>Find setgid files system packages</title>
+      <affected family="unix">
+        <platform>Red Hat Enterprise Linux 7</platform>
+        <platform>Oracle Linux 7</platform>
+      </affected>
+      <description>All files with setgid should be owned by a base system package</description>
+    </metadata>
+    <criteria>
+      <criterion comment="Check all setgid files" test_ref="check_setgid_files" />
+    </criteria>
+  </definition>
+
+  <unix:file_test check="all" check_existence="none_exist" comment="setgid files outside system RPMs" id="check_setgid_files" version="1">
+    <unix:object object_ref="object_file_permissions_unauthorized_sgid" />
+  </unix:file_test>
+
+  <unix:file_object comment="files with sgid set" id="object_file_permissions_unauthorized_sgid" version="1">
+    <unix:behaviors recurse="directories" recurse_direction="down" max_depth="-1" recurse_file_system="local" />
+    <unix:path operation="equals">/</unix:path>
+    <unix:filename operation="pattern match">^.*$</unix:filename>
+    <filter action="include">state_file_permissions_unauthorized_sgid</filter>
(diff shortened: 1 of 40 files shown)


mpreisler:
@openscap-ci add to whitelist

redhatrises:
@iokomin thanks for your PR! Great stuff. Do you plan on adding the Oracle STIG reference file and Oracle Linux STIGIDs?

ol7/checks/oval/file_permissions_unauthorized_sgid.xml
```diff
@@ -0,0 +1,78 @@
+<def-group>
+  <definition class="compliance" id="file_permissions_unauthorized_sgid" version="2">
+    <metadata>
+      <title>Find setgid files system packages</title>
+      <affected family="unix">
+        <platform>Oracle Linux 7</platform>
```
shawndwells: An OVAL check by the same name exists (ref: shared/xccdf/system/permission/files.xml). I'm not really familiar with Oracle Linux. Is something unique about OL7 that the shared content doesn't handle?

ol7/checks/oval/file_permissions_unauthorized_suid.xml
```diff
@@ -0,0 +1,93 @@
+<def-group>
+  <definition class="compliance" id="file_permissions_unauthorized_suid" version="1">
```
shawndwells: Same comment - could we use shared here?

ol7/templates/oval_5.11_templates/template_OVAL_service_disabled
```diff
@@ -0,0 +1,53 @@
...
+        <criterion comment="%SERVICENAME% not wanted by multi-user.target" test_ref="test_%SERVICENAME%_not_wanted_by_multi_user_target" />
+        <criterion comment="%SERVICENAME% socket not wanted by multi-user.target" test_ref="test_%SERVICENAME%_socket_not_wanted_by_multi_user_target" />
+        <criterion comment="%SERVICENAME% is not running" test_ref="test_service_not_running_%SERVICENAME%" />
+      </criteria>
+    </criteria>
+  </definition>
+
+  <linux:systemdunitdependency_test check="all" check_existence="any_exist" comment="systemd test" id="test_%SERVICENAME%_not_wanted_by_multi_user_target" version="1">
+    <linux:object object_ref="object_multi_user_target_for_%SERVICENAME%_disabled" />
+    <linux:state state_ref="state_systemd_%SERVICENAME%_off"/>
```
shawndwells: Because OL7 uses systemd, could the shared checks be used here as well? They look to be the same as RHEL -- which means we could gain efficiency by using one template.

ol7/checks/oval/file_permissions_unauthorized_sgid.xml
iokomin: Shawn, in general Oracle Linux (OL) mostly based on RHEL userspace packages source code. In addition to what is released in RHEL these packages might also contain additional fixes or improvements for Oracle products while some of packages might be excluded from OL distribution as not supported. OL also includes Unbreakable Enterprise Kernel, Ksplice and some other components developed in Oracle. Thus I would say from STIG perspective OL is 95% compatible with RHEL and majority of existing rhel content could be just reused. Though in some cases either guide content or OVAL checks might require platform specific changes.

Speaking of file_permissions_unauthorized_sgid.xml it is just copied from rhel7 content and adopted to OL as part of initial patch. Unfortunately I couldn't get it in shared oval folder despite it is referenced from shared xccdf. Further it will be reviewed and aligned with OL7 STIG, thus minor modifications could be considered. 
In theory it could be shared if 'var_suid_whitelist' which is product specific defined outside of this rule.

ol7/checks/oval/file_permissions_unauthorized_suid.xml
iokomin: No shared corresponding OVAL check observed at the moment. Could be shared in case  'var_suid_whitelist' which is product specific defined outside of this rule.

ol7/templates/oval_5.11_templates/template_OVAL_service_disabled
iokomin: OL manages services exactly in the same manner as RHEL does. As future improvement this templated could be moved to shared templates folder to be utilized at least by ol7 and rhel7 products.

iokomin:
@redhatrises, thanks for your feedback! 
The final goal is OL{6,7} STIG profiles implementation. First it was decided to adopt rhel7 'standard' and 'common' profiles for ol7 to populate product with oval, xccdf and remediation content.

redhatrises:
Ack. Need this to be merged for the conversion project.

PR #2667, opened 2018-03-14
Some audit tests and fixes

yuumasato:
#### Description:

- Add tests for `audit_rules_login_events`
- Add tests for `auditd_data_retention_space_left_action`
- Add tests for `audit_data_retention_space_left`
- Add tests for `audit_rules_privileged_commands`
- Fix remediation for `audit_data_retention_space_left`
- Fix bash remediation function  `perform_audit_rules_privileged_commands_remediation` to be able to deal with empty `/etc/audit/rules.d/`.

#### Notes
- Test scenario `own_key` for rule `audit_rules_privileged_commands` is failing for now. This is bug in OVAL definition that checks for specific key `privileged`.


shared/fixes/bash/auditd_data_retention_space_left.sh
```diff
@@ -2,7 +2,7 @@
 . /usr/share/scap-security-guide/remediation_functions
 populate var_auditd_space_left
 
-grep -q ^space_left /etc/audit/auditd.conf && \
+grep -q "^space_left[[:space:]]*=" /etc/audit/auditd.conf && \
```
matejak: We can use `\s` in grep, do we want to do it, or do we want to rather go with `[[:space:]]`? I feel that it is a good idea to decide for one and declare its preference in the bash remediation guideline.

yuumasato: Good point.

I'd rather stick with the POSIX standard, witch means using `[[:space:]]`.
Then, expressions as in https://github.com/OpenSCAP/scap-security-guide/blob/master/shared/fixes/bash/ensure_logrotate_activated.sh#L13 are more readable.

Also, from a quick `grep` in `shared/fixes/bash` it looks like `[[:space:]]` is used more than `\s`.

tests/data/group_system/group_auditing/group_auditd_data_retention/rule_auditd_data_retention_space_left/space_left.pass.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+# profiles = xccdf_org.ssgproject.content_profile_stig-rhel7-disa
+# remediation = bash
+
+if grep -q "^space_left[[:space:]]*= " /etc/audit/auditd.conf; then
+	sed -i "s/^space_left = .*$/space_left = 100/" /etc/audit/auditd.conf
```
matejak: It seems that you make sure that there is a line `^space_left = ...` and you make a substitution of the whole line to get 100 as right-hand side value. However, the grep regexp has exactly one space between `space_left` and equals. I think that it should be the same as the grep regexp - maybe have it in a variable?

tests/data/group_system/group_auditing/group_auditd_data_retention/rule_auditd_data_retention_space_left/space_left_not_enough.fail.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+# profiles = xccdf_org.ssgproject.content_profile_stig-rhel7-disa
+# remediation = bash
+
+sed -i "s/^space_left = .*$/space_left = 15/" /etc/audit/auditd.conf
```
matejak: ditto for this sed regexp and ones that follow, but I am not sure - maybe it is safe to assume that there will be exactly one space around the `=` operator?

shared/fixes/bash/auditd_data_retention_space_left.sh
```diff
@@ -2,7 +2,7 @@
...
   sed -i "s/^space_left[[:space:]]*=.*/space_left = $var_auditd_space_left/g" /etc/audit/auditd.conf
 if ! [ $? -eq 0 ]; then
     echo "space_left = $var_auditd_space_left" >> /etc/audit/auditd.conf
```
matejak: The if condition can be avoided by chaining the `echo` to the previous compound expression using `&&`.

matejak: This is a mistaken comment and you can either disregard it, or consider chaining using the `||` operator. Bash logical operators have the same precedence and are [left-associative](https://stackoverflow.com/a/930505).

yuumasato: I have changed the `if` to `||` approach.

tests/data/group_system/group_auditing/group_auditd_data_retention/rule_auditd_data_retention_space_left/space_left.pass.sh
yuumasato: I have made the regular expression consistent across tests and remediation.

yuumasato:
@matejak Thanks for your input
I have made the regular expressions consistent and chained the `echo` command with `||`.

matejak:
Great, thank you for making SSG better!

PR #2668, opened 2018-03-15
Made Ubuntu14 and Ubuntu16 to use local content.

matejak:
Similar to #2662, the Ubuntu local files couldn't be moved to shared, so shared functionality had to be moved local.
For future reference, after I have determined what shared files to transfer, this shell snippet assisted me in copyting and reviewing it:
```
STEMS=software permissions accounts accounts/restrictions
for stem in $STEMS
do
mkdir -p xccdf/system/$stem \
  && cp ../shared/xccdf/system/$stem/${stem##*/}.xml xccdf/system/$stem/ \
  && vim xccdf/system/$stem/${stem##*/}.xml
done
```

### DIFF
--- a/ubuntu14/guide.xslt
+++ b/ubuntu14/guide.xslt
@@ -34,9 +34,9 @@
       <xsl:apply-templates select="document(concat($BUILD_RP, '/bash-remediation-functions.xml'))" />
 
       <xsl:apply-templates select="document(concat($SHARED_RP, '/xccdf/intro/shared_intro_os.xml'))" />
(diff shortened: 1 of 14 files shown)


matejak:
The Ubuntu 16 build fails with some strange error `oval-linked` error. Ubuntu 14 doesn't fail, but I have verified that the Ubuntu14 and 16 changeset are the same, so the cause for it is probably not some typo.
I am investigating this issue.

mpreisler:
Good stuffm, thx! ACK.

PR #2597, opened 2018-01-31
[wip] initial OSPP v4.2 draft profile

shawndwells:
#### Description:
Add new profile that aligns to draft OSPP 4.1 configuration annex:
https://commoncriteria.github.io/pp/operatingsystem/configannex.html

#### Rationale:

Eventually this will be the "base" profile for all US Gov. Will need to refactor existing ospp profile.

PR #2670, opened 2018-03-15
Initial OSPP v4.2 draft profile

mpreisler:
See #2597 for info and original discussion. This is just a squashed version with conflicts resolved and the stigid build error fixed. Other than that I haven't done any changes.

author: @shawndwells


shawndwells:
There's still work to do, however the profile isn't linked into the build system. Won't impact releases.

rhel7/guide.xslt
```diff
@@ -28,6 +28,7 @@
       <xsl:apply-templates select="document('profiles/stig-rhel7-disa.xml')" />
       <xsl:apply-templates select="document('profiles/stig-rhvh-upstream.xml')" />
       <xsl:apply-templates select="document('profiles/ospp-rhel7.xml')" />
+      <xsl:apply-templates select="document('profiles/ospp42-draft-rhel7.xml')" />
```
mpreisler: couple things I don't like about the name. I wish it didn't contain rhel7 in it. That is already implied by bechmark file name - ssg-rhel7-ds.xml, why specify it again?

"draft" is good because it means people shouldnt' rely on it. I wish the version name wasn't there.

I guess what I am trying to get at is that in the future let's please rename profiles to small descriptive names that are stable across releases. And avoid all duplicates.

redhatrises: @mpreisler how would you suggest having version names that the profiles correspond to? ospp42 is clearly different than ospp.

mpreisler: ok nvm about the version then

repeating the product name in profile IDs is not useful though and we do it everywhere... It's maybe useful for improving typing speed of our users :-)

shawndwells:
On 3/15/18 5:18 PM, Martin Preisler wrote:
> ------------------------------------------------------------------------
>
> In rhel7/guide.xslt
> <https://github.com/OpenSCAP/scap-security-guide/pull/2670#discussion_r174935708>:
>
> > @@ -28,6 +28,7 @@
>        <xsl:apply-templates select="document('profiles/stig-rhel7-disa.xml')" />
>        <xsl:apply-templates select="document('profiles/stig-rhvh-upstream.xml')" />
>        <xsl:apply-templates select="document('profiles/ospp-rhel7.xml')" />
> +      <xsl:apply-templates select="document('profiles/ospp42-draft-rhel7.xml')" />
>
> couple things I don't like about the name. I wish it didn't contain
> rhel7 in it. That is already implied by bechmark file name -
> ssg-rhel7-ds.xml, why specify it again?
>
> "draft" is good because it means people shouldnt' rely on it. I wish
> the version name wasn't there.
>
> I guess what I am trying to get at is that in the future let's please
> rename profiles to small descriptive names that are stable across
> releases. And avoid all duplicates.
>

No objection. Not sure why the profile names have the OS in them... have
been that way for years and everyone just kept doing it ;)

We can clean the profile names up in a future release.

redhatrises:
Ack.

PR #2672, opened 2018-03-16
Updated parts of the project documentation.

matejak:
This PR has two distinct parts:
- Rewrite of a section of the SSG test suite README. Rationale about `.pass` and `.fail` scripts has been added.
- Regexp recommendation. As @yuumasato found out, POSIX forms of char classes are more common and more universal. There is just tradeoff of readability vs clutter.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -27,64 +27,65 @@ accessible via ssh on the libvirt domain and have Ansible installed on the host 
 
 *NOTE*: Create snapshot after all these steps, to manually revert in case the
 test suite breaks something and fails to revert. Do not use snapshot names
-starting with `ssg\_`
+starting with `ssg_`
 
-## Two modes of operation
-SSG Test Suite currently supports two ways of DataStream evaluation. Simpler
-one to use is `profile`, which scans and remedies target domain based on
-particular profile.
+## Executing the test suite
 
-Second, more complex, is `rule`, which runs set of validation scenarios
-for particular subset of rules. Each rule has its own scenarios, and each
-scenario is run separately to eliminate need for cleanup.
+### Common options
 
-## Running test suite in profile mode
-Example of evaluation of default profile (common):
+- `--hypervisor`: Typically, you will use the `qemu:///system` value.
+- `--domain`: `libvirt` domain, which is basically name of the virtual machine.
+- `--datastream`: Path to the datastream that you want to use for scanning.
+  It will be transferred to the scanned VM via SSH.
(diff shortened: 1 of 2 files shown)


tests/README.md
```diff
@@ -27,64 +27,65 @@ accessible via ssh on the libvirt domain and have Ansible installed on the host
...
+- `--hypervisor`: Typically, you will use the `qemu:///system` value.
+- `--domain`: `libvirt` domain, which is basically name of the virtual machine.
+- `--datastream`: Path to the datastream that you want to use for scanning.
+  It will be transferred to the scanned VM via SSH.
+- `--xccdf-id`: Also known as `Ref-ID`, it is the identifier of benchmark within the datastream.
+  It can be obtained by running `oscap info` over the datastream XML and searching for `Ref-ID` strings.
+- `--help`: This will get you the invocation help.
+
+### Profile-based testing
+In this operation mode, you specify the `profile` command and you supply the profile ID as a positional argument.
```
yuumasato: > `profile` command

The `profile` command is the example command below?

tests/README.md
```diff
@@ -27,64 +27,65 @@ accessible via ssh on the libvirt domain and have Ansible installed on the host
...
-```
-./test_suite.py profile --help
-```
+Rule-based testing enables to perform two kinds of tests:
 
+- Check tests: The system is set up into a compliant, or a non-compliant state.
 
-## Running test suite in rule mode
-Example of evaluation of rule ```rule_sysctl_net_ipv4_conf_default_secure_redirects```:
+  Typically, the compliant state is different from the default or post-remediation state.
```
yuumasato: Isn't the post-remediation state after the run of remediation? Shoudln't the system be compliant then?

mpreisler:
ACK

tests/README.md
matejak: What I wanted to say was that the compliant state induced by the script is typically different than compliant state induced by a remediation. Do you think that it is worth to clarify this further?

tests/README.md
matejak: Yes, that argument is commonly referred to as command or subcommand (e.g. `commit` command of `git`). I agree that subcommand could be easier to understand.

tests/README.md
yuumasato: I see, since an OVAL definition can return `pass` for multiple states, and while a remediation script only brings the system into a compliant state, it is good to test these other states in which result is `pass`.

I'm just not sure that it is that common for OVAL definitions have multiple states in which the system pass. Anyway, I think that clarification of why a compliant state can be different of a post-remediation state is enough.

PR #2673, opened 2018-03-16
Change ID of Rule that checks for IPV6 disabled

### LINKED ISSUE #2515: fix for xccdf_org.ssgproject.content_rule_sysctl_kernel_ipv6_disable is missing with no explanation
#### Description of problem:
Generating the bash fix for sysctl_kernel_ipv6_disable results in the following fix:
`# FIX FOR THIS RULE IS MISSING`

0.1.36-1.el7.centos

#### Steps to Reproduce:

1. Generate bash fixes for NIST 800-171

#### Actual Results:
```
###############################################################################
# BEGIN fix (204 / 359) for 'xccdf_org.ssgproject.content_rule_sysctl_kernel_ipv6_disable'
###############################################################################
(>&2 echo "Remediating rule 204/359: 'xccdf_org.ssgproject.content_rule_sysctl_kernel_ipv6_disable'")
# FIX FOR THIS RULE IS MISSING
# END fix for 'xccdf_org.ssgproject.content_rule_sysctl_kernel_ipv6_disable'
```

#### Expected Results:
```
function replace_or_append {
  local config_file=$1
  local key=$2
  local value=$3
  local cce=$4
  local format=$5

  # Check sanity of the input
  if [ $# -lt "3" ] 
  then
        echo "Usage: replace_or_append 'config_file_location' 'key_to_search' 'new_value'"
        echo  
        echo "If symlinks need to be taken into account, add yes/no to the last argument"
        echo "to allow to 'follow_symlinks'."
        echo "Aborting."
        exit 1
  fi  

  # Test if the config_file is a symbolic link. If so, use --follow-symlinks with sed.
  # Otherwise, regular sed command will do.
  if test -L $config_file; then
    sed_command="sed -i --follow-symlinks"
  else  
    sed_command="sed -i"
  fi    

  # Test that the cce arg is not empty or does not equal CCENUM.
  # If CCENUM exists, it means that there is no CCE assigned.
  if ! [ "x$cce" = x ] && [ "$cce" != 'CCENUM' ]; then
    cce="CCE-${cce}"
  else  
    cce="CCE"
  fi    

  # Strip any search characters in the key arg so that the key can be replaced without
  # adding any search characters to the config file.
  stripped_key=${key//[!a-zA-Z0-9_.]/}

  # If there is no print format specified in the last arg, use the default format.
  if ! [ "x$format" = x ] ; then
    printf -v formatted_output "$format" $stripped_key $value
  else  
    formatted_output="$stripped_key = $value"
  fi    

  # If the key exists, change it. Otherwise, add it to the config_file.
  if `grep -qi $key $config_file` ; then
    $sed_command "s/$key.*/$formatted_output/g" $config_file
  else  
    echo -ne "\n# Per $cce: Set $formatted_output in $config_file" >> $config_file
    echo -ne "\n$formatted_output" >> $config_file
  fi    

}

replace_or_append '/etc/sysctl.conf' 'net.ipv6.conf.all.disable_ipv6' '1' 'CCENUM'
```

This fix existed in a previous version of the scap-security-guide.  Usually when a fix is not enabled, there is a comment in the html results listing why.  For example, some missing fixes reference an outstanding RHEL bug (like the fix for enabling repo metadata gpg check).

Since scanning a system will list this as a fail, there should be an explanation as to why there is no fix currently available.

### LINKED ISSUE #2576: rhel 7.5 beta: sysctl_kernel_ipv6_disable still fails after remediation

yuumasato:
#### Description:

- Change ID of Rule `sysctl_kernel_ipv6_disable` to `sysctl_net_ipv6_conf_all_disable_ipv6`.

#### Rationale:
* The Rule is not following expected pattern. The sysctrl in question is
`net.ipv6.conf.all.disable_ipv6`, so exptected Rule id is
`syctl_net_ivp6_conf_all_disable_ipv6`.
* This is causing template generated remediation to not be picked up by
build system.
* The used pattern would be for a sysctl with name  `kernel.ipv6_disabled`, witch
doesn't exist.

- Changing ID of a Rule can break existing tailoring, but I'm not fan of keeping static remediation for this Rule.

* Fixes #2515 
* Fixes #2576

### DIFF
--- a/rhel7/profiles/C2S.xml
+++ b/rhel7/profiles/C2S.xml
@@ -341,7 +341,7 @@ baseline.
 <select idref="sysctl_net_ipv6_conf_default_accept_redirects" selected="true" />
 
 <!-- 3.3.3 Ensure IPv6 is disabled (Not Scored) -->
-<select idref="sysctl_kernel_ipv6_disable" selected="true" />
+<select idref="sysctl_net_ipv6_conf_all_disable_ipv6" selected="true" />
 
 <!-- 3.4 TCP Wrappers -->
 <!-- 3.4.1 Ensure TCP Wrappers is installed (Scored) -->
--- a/rhel7/profiles/ospp-rhel7.xml
+++ b/rhel7/profiles/ospp-rhel7.xml
@@ -115,7 +115,7 @@ the consensus process.
 <select idref="service_firewalld_enabled" selected="true" />
 <select idref="set_firewalld_default_zone" selected="true" />
 <select idref="firewalld_sshd_port_enabled" selected="true"/>
-<select idref="sysctl_kernel_ipv6_disable" selected="true" />
+<select idref="sysctl_net_ipv6_conf_all_disable_ipv6" selected="true" />
 <select idref="sysctl_net_ipv4_conf_all_accept_redirects" selected="true" />
 <select idref="sysctl_net_ipv4_conf_all_accept_source_route" selected="true" />
 <select idref="sysctl_net_ipv4_conf_all_log_martians" selected="true" />
--- a/rhel7/profiles/rht-ccp.xml
+++ b/rhel7/profiles/rht-ccp.xml
@@ -95,7 +95,7 @@
 
 <!-- <select idref="sysctl_kernel_randomize_va_space" selected="true"/>
 <select idref="enable_execshield" selected="true"/>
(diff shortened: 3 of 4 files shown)


yuumasato:
I'd like feedback on changing ID of this Rule.
I still plan to add test scenarios for this.

mpreisler:
I agree with this change.

redhatrises:
Ack

PR #2677, opened 2018-03-16
Use the shared benchmark for opensuse

mpreisler:
We are already de-facto doing that because none of the local XCCDF XMLs
exist.

Needed for yaml port.


redhatrises:
Ack.

PR #2678, opened 2018-03-16
RHOSP7 now uses the shared guide

mpreisler:
The prod_type filtering was already in the groups.

One last thing (yeah, right....) needed for yaml.


redhatrises:
Ack.

PR #2679, opened 2018-03-19
Partition remediations

dahaic:
#### Description:

- Creates mount point template to cover separate partitions for directories.

#### Rationale:

- Anaconda remediations for separate partitions requirements were missing, so this fixes the omission. 

- Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1462944


dahaic:
Jenkins test this please.

yuumasato:
@dahaic Great stuff!

Question though. Is it worth to keep separate `csv` files for each product?
As this about linux OVAL definition, and they all use same OVAL template, I think all `csv` files can be merged in shared?

debian8/templates/csv/mounts.csv
```diff
@@ -1,4 +1,10 @@
 # mountpoint options that should be set
```
yuumasato: It would be good make it clear that when there is no mount point option defined it is just about mount points in separate partition. Not that we are checking that the mount point option has no options.

Would be good to have it here as a comment, but also somewhere as documentation.

dahaic:
@yuumasato there are two lines, that are common between all mounts:
```
/tmp,nodev
/tmp,nosuid
```
I don't think it's worth splitting.

debian8/templates/csv/mounts.csv
dahaic: I will look into that, good point.

dahaic: Updated comments, hope it's ok now

dahaic:
Jenkins test this please

yuumasato:
Sorry that I only compared ovals and remediations only now. By doing that I noticed that Ansible remediations for mountpoint options were not generated.

The following patch fiexes the issue: https://gist.github.com/yuumasato/eb04360099345d2b05029a204d6948df

But I think this patch breaks an assumption that a `csv` has the same target language (OVAL, ansible, bash, puppet, anaconda) capability.
The patch allows for a `csv` file to have lines that works different sets of languages. 
But then, before, when first line if csv file failed to generate for a given language, whole file would be skipped.
Now, it will keep trying for all lines to generate for a language that doesn't make sense, and increase our build time.

Adding a new generator will keep the assumption that a `csv` file has the capability to generate for the same languages in whole file.

dahaic:

yuumasato:
Ack. Thank you!

PR #2684, opened 2018-03-22
Improved replace_or_append.

matejak:
* Removed misleading usage message when not enough args were specified.
* Made the sed command an array, avoiding usage of eval.
* Added double quotes when applicable.
* Removed echo -e in [favor of printf](https://unix.stackexchange.com/questions/65803/why-is-printf-better-than-echo).
* Achieved function compliance with shellcheck.


mpreisler:
This is awesome, thank you!

ACK

PR #2686, opened 2018-03-22
Generalized the 'not running' state of a service in OVAL check.

matejak:
Before, a service was acknowledged as "not running" only if it was in "inactive" state. However, some services tend to end up in the "failed" state, which is also a "not running" state.

For example, the `kdump` service fails on a default RHEL7 installation on a VM with this error:

```
... kdumpctl[1102]: No memory reserved for crash kernel
... kdumpctl[1102]: Starting kdump: [FAILED]
```

### DIFF
--- a/debian8/templates/oval_5.11_templates/template_OVAL_service_disabled
+++ b/debian8/templates/oval_5.11_templates/template_OVAL_service_disabled
@@ -47,7 +47,7 @@
     <linux:unit operation="pattern match">%DAEMONNAME%\.(service|socket)</linux:unit>
     <linux:property>ActiveState</linux:property>
   </linux:systemdunitproperty_object>
-  <linux:systemdunitproperty_state id="state_service_not_running_%SERVICENAME%" version="1" comment="%SERVICENAME% is not running">
-      <linux:value>inactive</linux:value>
+  <linux:systemdunitproperty_state id="state_service_not_running_%SERVICENAME%" version="1" comment="%SERVICENAME% is not running, either because nobody tried, service has been stopped, or because it terminated unexpectedly">
+      <linux:value operation="pattern match">(inactive|failed)</linux:value>
   </linux:systemdunitproperty_state>
 </def-group>
--- a/fedora/templates/oval_5.11_templates/template_OVAL_service_disabled
+++ b/fedora/templates/oval_5.11_templates/template_OVAL_service_disabled
@@ -47,7 +47,7 @@
     <linux:unit operation="pattern match">%SERVICENAME%\.(service|socket)</linux:unit>
     <linux:property>ActiveState</linux:property>
   </linux:systemdunitproperty_object>
-  <linux:systemdunitproperty_state id="state_service_not_running_%SERVICENAME%" version="1" comment="%SERVICENAME% is not running">
-      <linux:value>inactive</linux:value>
+  <linux:systemdunitproperty_state id="state_service_not_running_%SERVICENAME%" version="1" comment="%SERVICENAME% is not running, either because nobody tried, service has been stopped, or because it terminated unexpectedly">
+      <linux:value operation="pattern match">(inactive|failed)</linux:value>
   </linux:systemdunitproperty_state>
 </def-group>
--- a/ol7/templates/oval_5.11_templates/template_OVAL_service_disabled
+++ b/ol7/templates/oval_5.11_templates/template_OVAL_service_disabled
@@ -47,7 +47,7 @@
     <linux:unit operation="pattern match">%SERVICENAME%\.(service|socket)</linux:unit>
     <linux:property>ActiveState</linux:property>
   </linux:systemdunitproperty_object>
-  <linux:systemdunitproperty_state id="state_service_not_running_%SERVICENAME%" version="1" comment="%SERVICENAME% is not running">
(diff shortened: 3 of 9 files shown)


redhatrises:
Thanks for this @matejak. Should we perform process58_test checks as well?

mpreisler:
This particular template is all over the place, could you please make sure the fix is applied everywhere?

```
$ find . -name template_OVAL_service_disabled
./fedora/templates/oval_5.11_templates/template_OVAL_service_disabled
./ubuntu14/templates/template_OVAL_service_disabled
./ubuntu14/templates/oval_5.11_templates/template_OVAL_service_disabled
./rhel7/templates/oval_5.11_templates/template_OVAL_service_disabled
./rhosp7/templates/oval_5.11_templates/template_OVAL_service_disabled
./ubuntu16/templates/template_OVAL_service_disabled
./ubuntu16/templates/oval_5.11_templates/template_OVAL_service_disabled
./rhel6/templates/template_OVAL_service_disabled
./opensuse/templates/oval_5.11_templates/template_OVAL_service_disabled
./sle12/templates/oval_5.11_templates/template_OVAL_service_disabled
./ol7/templates/oval_5.11_templates/template_OVAL_service_disabled
./debian8/templates/template_OVAL_service_disabled
./debian8/templates/oval_5.11_templates/template_OVAL_service_disabled
```

matejak:
I will put this to another templates as well as suggested by @mpreisler.
@redhatrises I think that as those enable/disable checks use `systemd`, it is enough to make sure that the `systemd` output is interpreted properly. Plus it seems to me that the exact name of the running process may not match the basename of the unit file.

redhatrises:
> Plus it seems to me that the exact name of the running process may not match the basename of the unit file.

Good point.

shawndwells:
The intent of the check is to ensure services are not configured to automatically start when the OS is booted. 

Is there a way with systemd to evaluate if a service is configured to start on boot, vs how we attest this now (through ``inactive`` state)?

If a service state is ``failed``, then it was started. How can we tell if that was because of system initialization or an admin running systemctl?

matejak:
The OVAL check is quite strict. It wants to make sure that the service is disabled AND that it is not even running. The PR makes changes to the second part of the check, the service still has to be disabled in order for the check to pass. The complete OVAL check is in those files that were edited, so you can check them out.

redhatrises:
So, I reread this the PR description here, and I believe that if a service has failed that this check should fail. This is OVAL is quite strict by design.

shawndwells:
On 3/22/18 12:48 PM, Matěj Týč wrote:
> The OVAL check is quite strict. It wants to make sure that the service
> is disabled AND that it is not even running. The PR makes changes to
> the second part of the check, the service still has to be disabled in
> order for the check to pass. The complete OVAL check is in those files
> that were edited, so you can check them out.

This PR exposes a good conversation. Evaluating by "Legislative Intent"
may be appropriate.

``failed`` services are clearly not running, but does indicate service
initialization happened. Who knows what was executed between
initialization and the failed state.

The intent of the check is to ensure the service was disabled and not
initialized. With these changes, a service could be disabled but
initialized and still pass. That goes against the intent of the check.

yuumasato:
Discussing with @matejak, the story is as follows:

kdump service attempted to start, but it failed for reasons not important here. kdump service was clearly enabled, because it attempted to start.
Evaluation happened, followed by remediation, which disabled and stopped the service.
Evaluation happened again, and found service as disabled and failed, as apparently the service cannot go from failed to inactive by stop command.
From sysctl man page:
```
reset-failed 
...
When a unit fails in some way (i.e. process exiting with non-zero error code, terminating
abnormally or timing out), it will automatically enter the "failed" state and its exit code
 and status is recorded for introspection by the administrator until the service is restarted 
or reset with this command.
```
After reboot, the status of service was disabled and inactive.

>  Who knows what was executed between initialization and the failed state.

Maybe the service should be masked? Then it is assured it cannot be started manually, nor as a dependency.

> With these changes, a service could be disabled but initialized and still pass.

No, in this case it will fail. Only if something went wrong and state goes to fail, the rule would pass.

redhatrises:
Kdump seems like a special case considering it is an Anaconda Addon, and is configured in kickstart like:
```
%addon com_redhat_kdump --disable
%end
```
I wonder if we can add this functionality to OAA and the anaconda templates?

dahaic:
@matejak so if I get it right, we can, instead of generalization of OVAL, include `reset-failed` to all disable remediations?

yuumasato:
Proposal of adding `reset-failed` to remediation is interesting.

>I wonder if we can add this functionality to OAA and the anaconda templates?

I don't know of a command for OAA to disable an addon. 
But if we consider that OAA can add to the kickstart file, we may be able to drop the snippet in `ks` file and disable kdump addon.

matejak:
@dahaic What exactly do you mean by `reset-failed`?

matejak:
My apologies. I like the approach of running `systemctl reset failed` after disabling services. I will propose this in another PR.

mpreisler:
@matejak Needs a rebase

redhatrises:
Nack on this PR. The strictness of the OVAL check is correct and desired by many of our customers.

matejak:
The alternative fix has been merged in #2698 , so I am closing this.

PR #2687, opened 2018-03-22
Improve remediation of auditd_data_disk_full_action

matejak:
So far, I have made the remediation to use our shared remediation functionality. However, the remediation is not included in the datastream. I have no idea why, renaming the bash file to `content_rule_auditd_data_disk_full_action.sh` didn't help. 
Any ideas?

### DIFF
--- a/shared/fixes/bash/auditd_data_disk_full_action.sh
+++ b/shared/fixes/bash/auditd_data_disk_full_action.sh
@@ -1,16 +1,8 @@
 # platform = multi_platform_rhel
+
+# Include source function library.
 . /usr/share/scap-security-guide/remediation_functions
+
 populate var_auditd_disk_full_action
 
-#
-# If disk_full_action present in /etc/audit/auditd.conf, change value
-# to var_auditd_disk_full_action, else
-# add "disk_full_action = $var_auditd_disk_full_action" to /etc/audit/auditd.conf
-#
-
-if grep --silent ^disk_full_action /etc/audit/auditd.conf ; then
-        sed -i 's/^disk_full_action.*/disk_full_action = '"$var_auditd_disk_full_action"'/g' /etc/audit/auditd.conf
-else
-        echo -e "\n# Set disk_full_action to $var_auditd_disk_full_action per security requirements" >> /etc/audit/auditd.conf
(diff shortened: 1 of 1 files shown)


redhatrises:
```
```

matejak:
I have amended my commit, so the script contains the exact comment. The inclusion of remediation functions was there even before.
Next, I have decided to remove other comments that just stated the obvious.

shared/fixes/bash/auditd_data_disk_full_action.sh
```diff
@@ -1,16 +1,10 @@
...
 . /usr/share/scap-security-guide/remediation_functions
+
 populate var_auditd_disk_full_action
 
-#
-# If disk_full_action present in /etc/audit/auditd.conf, change value
-# to var_auditd_disk_full_action, else
-# add "disk_full_action = $var_auditd_disk_full_action" to /etc/audit/auditd.conf
-#
+cpe_number='27370-6'
```
redhatrises: why this change?

matejak: I think that it increases readability at no cost.

redhatrises: In this shorthand script, sure. But in HTML output, I don't think that it does.

mpreisler: we have a macro for CCEs, they must not be hardcoded in the scripts

mpreisler: and it's cce, not cpe, so this makes it super confusing

redhatrises: If I recall correctly, making this change would break that macro, but I haven't tried it to confirm.

matejak: OK, I have rebased and put the `@CCENUM@` macro back there.

matejak:
To explain what went wrong, I have accidentally mixed up this remediation with remediation for another rule, therefore I couldn't find it when I was looking for it (this is a RHEL6 remediation, whereas I was looking for it in RHEL7).
As of now, nothing seems to be wrong with it.

yuumasato:
Ack.

PR #2688, opened 2018-03-23
Tests and OVAL fix for Rule sssd_enable_pam_services

yuumasato:
#### Description:

- Add tests for `sssd_enable_pam_services`
  - sssd_pam_services - pass - expected config line is present
  - services_pam_missing - fail - expected config line is missing
  - services_pam_wrong_section - fail - expected config line is put in wrong section file
- Regex in OVAL check made stricter, motivated by test that puts config line in wrong section.
- Add bash remediation for `sssd_enable_pam_services`

### DIFF
--- a/shared/checks/oval/sssd_enable_pam_services.xml
+++ b/shared/checks/oval/sssd_enable_pam_services.xml
@@ -20,7 +20,7 @@
   </ind:textfilecontent54_test>
   <ind:textfilecontent54_object id="obj_sssd_enable_pam_services" version="1">
     <ind:filepath>/etc/sssd/sssd.conf</ind:filepath>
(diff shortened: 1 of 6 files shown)


yuumasato:
Remediation for this rule would be to add following snippet to `/etc/sssd/sssd.conf`.
```
[sssd]
services = pam
```

But the config file is not there after `sssd` is installed.
I can find this file `/usr/lib64/sssd/conf/sssd.conf`, it contains the snippet, but it seems to be a sample.
Rationale for this seems to be that it should be generated by some humans, or a human assisted by tools.

Remediation is easy, but I don't feel confident that it should add the configuration if it doesn't exist.
Opinions on these?

dahaic:
@yuumasato I would say just create the file. That way, we are sure not to introduce more behaviour changes than is required by the policy we harden to.

yuumasato:
Bash remediation added.

mpreisler:
Great stuff! ACK

PR #2689, opened 2018-03-23
Further improved replace_or_append

matejak:
Further improvements to the function:

- Harmless improvement: Synchronized the case sensitivity option of `grep` and `sed` commands. In cases where the provided and actual keys differed by case, the substitution would fail.
The modification is backwards-compatible (c-insensitivity is on by default) and controlled from a single point.

- Potentially controversial, but needed improvement: I have appended the word boundary right after the key for the matching.
Although this change alters behavior of the function, remediations that are not broken already shouldn't break, as this change just prevents longer strings that begin with the substring that happens to be the ky to be mistaken for it. For example, `space_left_action` begins with substring `space_left`, so remediation for `auditd_data_retention_space_left` rule could break the `auditd_data_retention_space_left_action` rule.

### DIFF
--- a/shared/bash_remediation_functions/replace_or_append.sh
+++ b/shared/bash_remediation_functions/replace_or_append.sh
@@ -65,8 +65,10 @@ function replace_or_append {
   printf -v formatted_output "$format" "$stripped_key" "$value"
 
   # If the key exists, change it. Otherwise, add it to the config_file.
-  if grep -q $grep_case_insensitive_option "${key}" "$config_file"; then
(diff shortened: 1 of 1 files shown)


yuumasato:
I'm totally in favor of the potentially controversial improvement, although I don't see it can be controversial.

mpreisler:
I think the "controversial" improvement is a great improvement for additional sanity of the fixes.

matejak:
While I also think that it is a good idea, it changes the behavior of the function. Although I think that it shouldn't break anything, somebody who is very cautious might oppose this change.

mpreisler:
ACK

PR #2690, opened 2018-03-23
Remove the dumper scripts

mpreisler:
The shorthand data has already been moved to yaml.
These scripts are no longer needed.

See #2592


mpreisler:
Also gets rid of warnings such as:
```
Scanning for dependencies of opensuse checks (OVAL)...
Encountered file 'dump_shorthand.py' while recursing, extension '.py' is unknown. Skipping..
```

matejak:
Seems a good time to do it, if there is need for them, they will be in the git history.
Thumbs up for cleaning up!

PR #2691, opened 2018-03-23
Moved stuff around so that the folder matches the Makefile target

mpreisler:
Makes it a little less confusing when you want to build just one
product. There is no impact on build outputs or what gets installed.


CMakeLists.txt
```diff
@@ -208,7 +208,7 @@ if (SSG_PRODUCT_OPENSUSE)
     add_subdirectory("opensuse")
 endif()
 if (SSG_PRODUCT_OSP7)
-    add_subdirectory("rhosp7")
+    add_subdirectory("rhel-osp7")
```
redhatrises: Can we actually have this as `rhosp7`?

CMakeLists.txt
```diff
@@ -223,10 +223,10 @@ if (SSG_PRODUCT_SUSE12)
     add_subdirectory("sle12")
 endif()
 if (SSG_PRODUCT_UBUNTU14)
-    add_subdirectory("ubuntu14")
+    add_subdirectory("ubuntu1404")
```
redhatrises: Can we actually have this as ubuntu14?

CMakeLists.txt
```diff
@@ -223,10 +223,10 @@ if (SSG_PRODUCT_SUSE12)
...
 endif()
 if (SSG_PRODUCT_UBUNTU16)
-    add_subdirectory("ubuntu16")
+    add_subdirectory("ubuntu1604")
```
redhatrises: Can we actually have this as `ubuntu16`?

mpreisler:
Changing product names changes names of the datastreams we output so it's potentially breaking use-cases for users.

redhatrises:
> Changing product names changes names of the datastreams we output so it's potentially breaking use-cases for users.

You're right. We just changed directory naming.

Still would like to simply the naming.

redhatrises:
Why not fix the naming here and then forbid it in a separate PR? :D

mpreisler:
> Why not fix the naming here and then forbid it in a separate PR? :D

If we change naming we break use-cases, rpms, deb packages, tests, installers. Exposing what the names are is the first step. If you want to break and change names, let's do that as a separate PR, this is just refactoring, no intention to change any outputs.

redhatrises:
Ack.

PR #2694, opened 2018-03-23
shorthand.xml target should depend on the yaml-to-shorthand script

mpreisler:
That way, when you change the script it will auto-rebuild the shorthand
and everything depending on it.

Not a big deal but we do it for every other target and it makes sense IMO.


yuumasato:
Ack.

PR #2696, opened 2018-03-26
Added support for checks and remediation for mount_options.

matejak:
* Introduced templates for bash remediations of mount_options.
* Enabled templates for profiles that contain respective rules in XCCDF.
* Introduced shared bash remediation based on existing remediations.

There is a change in behavior - before, the remediations have removed `defaults,` from mount options. I don't think that this is a good idea, so I have removed it. However, there is a function `remove_defaults_from_fstab_if_overriden` available that will do just that if passed the mount point as its first argument.


mpreisler:
This is awesome. Could you please remove all non-templated versions of this?
```
# already removed:
./rhel7/fixes/bash/mount_option_tmp_noexec.sh
./rhel7/fixes/bash/mount_option_tmp_nodev.sh
./rhel7/fixes/bash/mount_option_tmp_nosuid.sh
# still in repo:
./rhel6/fixes/bash/mount_option_dev_shm_noexec.sh
./rhel6/fixes/bash/mount_option_dev_shm_nodev.sh
./rhel6/fixes/bash/mount_option_dev_shm_nosuid.sh
./rhel6/fixes/bash/mount_option_nodev_nonroot_local_partitions.sh
./shared/fixes/bash/mount_option_noexec_removable_partitions.sh
./shared/fixes/bash/mount_option_var_tmp_bind.sh
./shared/fixes/bash/mount_option_nodev_removable_partitions.sh
./shared/fixes/bash/mount_option_nosuid_removable_partitions.sh
```

matejak:
Good point, however, the `./rhel6/fixes/bash/mount_option_nodev_nonroot_local_partitions.sh` is quite special, so either we leave it like it its, or further investigation is needed.
I was not sure how the value substitution would work on those CSVs, so I have kept the `shared/fixes/bash/mount_option_<option>_removable_partitions.sh` and just changed their implementation.

mpreisler:
@matejak For variables we usually create a template with the suffix `_var`. It's just a slightly altered version of the origianl template. For examples see
- template_ANSIBLE_sebool and template_ANSIBLE_sebool_var
- template_ANSIBLE_sysctl and template_ANSIBLE_sysctl_var
- ...

matejak:
The removable partition remediations have been taken care of, thanks @dahaic for helping me out. OVAL can't be made into templates easily as it contains quite a lot of human-readable text that is, as a whole, unique for each mount point / mount option combination.

matejak:
@openscap-ci test this pls.

matejak:
@openscap-ci jenkins test this please.

dahaic:
@matejak when I templated `partition_for_` rules, I have reduced oval descriptions to the templated text as well. I do not consider OVAL file to be separately consumed product of SSG, it's just part of DataStream, which has descriptions from the XCCDF component. So there is no point in having bespoke oval descriptions there.

mpreisler:
@openscap-ci test this please

wrlinux/templates/template_BASH_mount_options
```diff
@@ -0,0 +1,9 @@
+# platform = multi_platform_wrlinux
```
dahaic: Again, do we need separate template, if the contents are the same?

mpreisler:
@matejak needs a rebase

matejak:
I have no idea what is wrong with the build.

mpreisler:
@matejak 
```
ERROR: Invalid OVAL definition referenced by XCCDF Rule: mount_option_nosuid_removable_partitions
ERROR: Invalid OVAL definition referenced by XCCDF Rule: mount_option_nodev_removable_partitions
ERROR: Invalid OVAL definition referenced by XCCDF Rule: mount_option_noexec_removable_partitions
```

This error message is pretty bad but IIRC it means the XCCDF check-content-ref is referencing an OVAL def that is not in that product's OVAL file. Make sure the OVAL defs are applicable to the products that error out, make sure they end up in the OVAL.

matejak:
I also thought this way, but it doesn't seem to work that way. Now, when I grep the OVAL templates, I get:
```
$ grep -R platform shared/templates/template_OVAL_mount*
shared/templates/template_OVAL_mount:        <platform>multi_platform_all</platform>
shared/templates/template_OVAL_mount_option:        <platform>multi_platform_all</platform>
shared/templates/template_OVAL_mount_option_remote_filesystems:        <platform>multi_platform_all</platform>
shared/templates/template_OVAL_mount_option_removable_partitions:        <platform>multi_platform_all</platform>
```
and the `mount_options.csv` in `shared` contains
```
var_removable_partition,nodev
var_removable_partition,nosuid
var_removable_partition,noexec
```
which works for e.g. RHEL7. It seems that it is not picked up by the Fedora and wrlinux.

mpreisler:
@matejak go to build/fedora/checks and run `find` in that folder recursively, look whether these files get generated there.

matejak:
@mpreisler Those files are there. Even if I declare the template for the Fedora platform, the check of built Fedora stuff fails with the same error.
Strangely, there is a corresponding file `fedora/checks/shared/oval/mount_option_nodev_removable_partitions.xml` and the content is included in `fedora/oval-unlinked.xml`. However, nothing concerning removable partitions makes it to the `fedora/oval-linked.xml`, except variables.

mpreisler:
@matejak In these cases I usually just debug the scripts to figure it out.

One possible explanation is that the defs that are removed from the linked OVAL are extending defs that are not available in the Fedora OVAL.

matejak:
OK, it is probably the `relabel-ids` script that is responsible for it, so I will look into it.

matejak:
The tests now work - thanks to @mpreisler to preciesly point the cause and to @yuumasato who made sure that I understand it correctly.

matejak:
There are still flaws in the mount options checks (e.g. the OVAL check of removable partitions is really strange), but it is a huge batch of work, so I propose to leave that for another PRs.

matejak:
@openscap-ci jenkinst test

shared/templates/template_BASH_mount_option_var
```diff
@@ -0,0 +1,17 @@
...
+
+include_mount_options_functions
+
+# test "$mount_has_to_exist" = 'yes'
+test "%MOUNT_HAS_TO_EXIST%" = 'yes' && assert_mount_point_in_fstab %MOUNTPOINT% \
+	|| { echo "Not remediating, because there is no record of %MOUNTPOINT% in /etc/fstab" >&2; exit 1; }
+
+ensure_mount_option_in_fstab "$%MOUNTPOINT%" "%MOUNTOPTION%"
+
+ensure_partition_is_mounted "%MOUNTPOINT%"
```
yuumasato: As `%MOUNTPOINT%` can be `var_removable_partition`, I believe it needs to be expanded in lines, 12, 13 and 17.

matejak: Nice catch!

matejak:
@yuumasato Yes, the `/dev/shm` OVAL check is also strange in a similar way - it passes, but the HTML report shows errors. I guess that those checks will have to be seriously refactored.

matejak:
Yes, OVAL definitions were dropped, because there is a shared template. I have started to use it, and I have also refactored the template instantiation code again, so it can support that.

shared/templates/create_mount_options.py
```diff
@@ -8,83 +8,124 @@
...
-        return self.process(mount_point, mount_option, point_id,
-                            assert_mount_exists, template_file, stem)
+    def _set_correct_values(self, mount_point, mount_option, point_id):
+        super(RemediationTarget, self)._set_correct_values(mount_point, mount_option, point_id)
+        if mount_point.startswith("var_"):
+            self._point_id = re.sub(r"^var_(.*)", r"\1s", mount_point)
+            self.template_file = "{0}_var".format(self.TEMPLATE_FILE_BASE)
+            self._output_id_template = "{mount_option}_{point_id}"
+        elif not mount_point.startswith("/"):  # no path, but not a variable either
+            raise Skipped("No template available yet")
```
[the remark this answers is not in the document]
matejak: I can't inform anybody here, because it would make the build process a UX disaster. So I just left some comments, so it will help when the build system will be refactored.

shared/templates/create_mount_options.py
```diff
@@ -8,44 +8,143 @@
...
+
+
+class OvalTarget(MountOptionTarget):
+    TEMPLATE_FILE_BASE = "./template_OVAL_mount_option"
+
+    def __init__(self, generator):
+        super(OvalTarget, self).__init__(
+            generator, "oval")
+
+    def _set_correct_values(self, mount_point, mount_option, point_id):
```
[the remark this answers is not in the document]
matejak: I have renamed it, but anyone please help me if you have a better name idea. I think that the method does one thing, so it justified, but finding a not-too-long name for it may be a challenge.

matejak:
I have forgot to commit the shared template, thank you for pointing that out so clearly :-)

PR #2697, opened 2018-03-26
Make it possible to build SSG with python3

mpreisler:
We need to be careful about treating text as bytes and vice-versa.

It's now possible to build SSG with python2 or python3. The port was pretty easy.

### DIFF
--- a/shared/utils/enable-derivatives.py
+++ b/shared/utils/enable-derivatives.py
@@ -208,13 +208,16 @@ def main():
     parser.add_option("--enable-sl", dest="sl", default=False,
                       action="store_true", help="Enable Scientific Linux")
     parser.add_option("-i", "--input", dest="input_content", default=False,
-                      action="store", help="INPUT can be XCCDF or Source DataStream")
+                      action="store",
+                      help="INPUT can be XCCDF or Source DataStream")
     parser.add_option("-o", "--output", dest="output", default=False,
                       action="store", help="XML Tree content")
     (options, args) = parser.parse_args()
 
     if options.centos and options.sl:
-        print "Cannot enable two derivative OS(s) at the same time"
+        sys.stderr.write(
+            "Cannot enable two derivative OS(s) at the same time\n"
+        )
         parser.print_help()
         sys.exit(1)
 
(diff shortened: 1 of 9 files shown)


jan-cerny:
@mpreisler This is really great news! :+1: 

What is your opinion on building Fedora package using Python 3?

shared/utils/enable-derivatives.py
```diff
@@ -208,13 +208,16 @@ def main():
...
     parser.add_option("-i", "--input", dest="input_content", default=False,
-                      action="store", help="INPUT can be XCCDF or Source DataStream")
+                      action="store",
+                      help="INPUT can be XCCDF or Source DataStream")
     parser.add_option("-o", "--output", dest="output", default=False,
                       action="store", help="XML Tree content")
     (options, args) = parser.parse_args()
 
     if options.centos and options.sl:
-        print "Cannot enable two derivative OS(s) at the same time"
```
matejak: FYI, you could have used `print("...", file=sys.stderr)`

shared/utils/enable-derivatives.py
```diff
@@ -261,16 +264,18 @@ def scrape_benchmarks(root_element, namespace, dest):
     for namespace, benchmark in benchmarks:
         if not add_derivative_cpes(benchmark, namespace, mapping):
             raise RuntimeError(
-                "Could not add derivative OS CPEs to Benchmark '%s'." % (benchmark)
```
matejak: I suggest raising the line length limit from 80 to 100. Although majority of lines can and should be under 80 chars long, I think that enforcing this rule upon every line is more like self-torture. PEP8 actually states that it is OK to increase to width of 99: https://www.python.org/dev/peps/pep-0008/#maximum-line-length

mpreisler: sure, I'd be all for that but it's unrelated to this PR. Let's make a separate PR that changes the pep8 settings.

mpreisler:
> What is your opinion on building Fedora package using Python 3?

IMO we should start doing that.

shared/utils/enable-derivatives.py
mpreisler: good point, do you want me to change that? we use sys.stderr.write all over the place in SSG, if one approach is preferable we should switch everywhere...

matejak: As you wish, I am fine either way. However, I have noticed that we don't use the "future imports", which is probably wrong - that way, print usage across Python2 and Python3 may be inconsistent.
Therefore, I suggest putting `from __future__ import print_function` at the top of modules that use `print`. This is supported from Python 2.6 and it will ensure that e.g. `print(1, 2)` will behave the same and also ensure the `file` kwarg availability.

shared/utils/enable-derivatives.py
matejak: I agree, fair enough.

PR #2698, opened 2018-03-27
Made the service disable command more complete.

matejak:
As discussed in #2686, this PR changes the remediation, so the discussed use cases work without modifying OVAL.

I have also removed some excessive negations that were present.

### DIFF
--- a/shared/bash_remediation_functions/service_command.sh
+++ b/shared/bash_remediation_functions/service_command.sh
@@ -47,20 +47,24 @@ else
 fi
 
 # If chkconfig_util is not empty, use chkconfig/service commands.
-if ! [ "x$chkconfig_util" = x ] ; then
+if [ "x$chkconfig_util" != x ] ; then
   $service_util $service $service_operation
   $chkconfig_util --level 0123456 $service $chkconfig_state
 else
   $service_util $service_operation $service
   $service_util $service_state $service
+  # The service may not be running because it has been started and failed,
+  # so let's reset the state so OVAL checks pass.
+  # Service should be 'inactive', not 'failed' after reboot though.
+  $service_util reset-failed $service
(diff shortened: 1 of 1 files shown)


shared/bash_remediation_functions/service_command.sh
```diff
@@ -47,20 +47,24 @@ else
...
+if [ "x$chkconfig_util" != x ] ; then
   $service_util $service $service_operation
   $chkconfig_util --level 0123456 $service $chkconfig_state
 else
   $service_util $service_operation $service
   $service_util $service_state $service
+  # The service may not be running because it has been started and failed,
+  # so let's reset the state so OVAL checks pass.
+  # Service should be 'inactive', not 'failed' after reboot though.
+  $service_util reset-failed $service
```
mpreisler: does this work with the "service" command? this looks like something that is systemctl only.

matejak: The shell is not clear, at first sight, but I believe that the branch of the if statement is the exclusive `systemctl` branch.

redhatrises: Haven't tested but does OVAL really report pass/fail on this?

matejak: @redhatrises The PR #2686 exposes that the OVAL check on whether a service is disabled will fail is the service is indeed disabled, but in a `failed` state. Or you meant something else?

redhatrises: @matejak no didn't mean something else... I think I need more coffee as I am confusing the difference between the two PRs.

mpreisler:
@openscap-ci test this please

mpreisler:
ACK

PR #2699, opened 2018-03-27
Added print function support for Python2 where applicable.

matejak:
This is needed for print statement consistency, details here:
https://docs.python.org/2/reference/simple_stmts.html#future

This small change didn't get into PR #2697.


mpreisler:
ACK

PR #2704, opened 2018-04-03
Refactoring of relabel-ids.py

matejak:
The goal of this PR is to refactor the code - functionality should be preserved, but the internals may change.

Goals to reach:

* Remove unhelpful comments.
* Streamline conditional statements.
* Deduplicate code into functions.
* Reduce indentation level of nested loops/if-statements.
* Add argparse interface.

Goals in question:

* Raise and handle exceptions when some conditions are not met - make the operation more strict?

### DIFF
--- a/shared/utils/relabel-ids.py
+++ b/shared/utils/relabel-ids.py
@@ -2,22 +2,10 @@
 
 from __future__ import print_function
 
-import re
 import sys
 import os
 
 import argparse
-
-try:
-    from xml.etree import cElementTree as ElementTree
-except ImportError:
-    import cElementTree as ElementTree
-
-try:
-    from configparser import SafeConfigParser
-except ImportError:
-    # for python2
-    from ConfigParser import SafeConfigParser
 
 # Put shared python modules in path
 sys.path.insert(0, os.path.join(
@@ -52,16 +40,6 @@ OVAL_TO_XCCDF_DATATYPE_CONSTRAINTS = {
 }
 
 
-def get_check_content_ref_if_exists_and_not_remote(check):
-    checkcontentref = check.find("./{%s}check-content-ref" % xccdf_ns)
-    if checkcontentref is None:
-        return None
-    if check_content_href_is_remote(checkcontentref):
-        return None
-    else:
-        return checkcontentref
-
-
 class FileLinker(object):
     CHECK_SYSTEM = None
     CHECK_NAMESPACE = None
@@ -83,7 +61,7 @@ class FileLinker(object):
             # Include the file in the particular check system only if it's NOT
             # a remotely located file (to allow OVAL checks to reference http://
             # and https:// formatted URLs)
-            checkcontentref = get_check_content_ref_if_exists_and_not_remote(check)
+            checkcontentref = ssgcommon.get_check_content_ref_if_exists_and_not_remote(check)
             if checkcontentref is not None:
                 checkfiles.add(checkcontentref.get("href"))
         return checkfiles
@@ -99,14 +77,14 @@ class FileLinker(object):
     def save_linked_tree(self):
         assert self.tree is not None, \
             "There is no tree to save, you have probably skipped the linking phase"
-        ElementTree.ElementTree(self.tree).write(self.linked_fname)
+        ssgcommon.ElementTree.ElementTree(self.tree).write(self.linked_fname)
 
     def _get_checkid_string(self):
         raise NotImplementedError()
 
     def link_xccdf(self):
         for check in self.checks_related_to_us:
-            checkcontentref = get_check_content_ref_if_exists_and_not_remote(check)
+            checkcontentref = ssgcommon.get_check_content_ref_if_exists_and_not_remote(check)
             if checkcontentref is None:
                 continue
 
@@ -135,7 +113,7 @@ class OvalFileLinker(FileLinker):
         return "{%s}definition" % self.CHECK_NAMESPACE
 
     def link(self):
-        self.tree = parse_xml_file(self.fname)
+        self.tree = ssgcommon.parse_xml_file(self.fname)
         try:
             self._link_oval_tree()
 
@@ -151,7 +129,7 @@ class OvalFileLinker(FileLinker):
     def _link_oval_tree(self):
         xccdf_to_cce_id_mapping = create_xccdf_id_to_cce_id_mapping(self.xccdftree)
 
-        indexed_oval_defs = aggregate_elements_by_id(
+        indexed_oval_defs = ssgcommon.map_elements_to_their_ids(
             self.tree, ".//{0}".format(self._get_checkid_string()))
 
         drop_oval_checks_extending_non_existing_checks(
@@ -200,7 +178,7 @@ class OvalFileLinker(FileLinker):
 
             # Then append the <reference source="CCE" ref_id="CCE-ID" /> element right
             # after <description> element of specific OVAL check
-            ccerefelem = ElementTree.Element(
+            ccerefelem = ssgcommon.ElementTree.Element(
                 'reference', ref_id=xccdfcceid, source="CCE")
             ovaldesc.addnext(ccerefelem)
             # Sanity check if appending succeeded
@@ -227,7 +205,7 @@ class OvalFileLinker(FileLinker):
                 if check.get("system") != oval_cs:
                     continue
 
-                if get_check_content_ref_if_exists_and_not_remote(check) is None:
+                if ssgcommon.get_check_content_ref_if_exists_and_not_remote(check) is None:
                     continue
 
                 # For local OVAL drop the reference to OVAL definition from XCCDF document
@@ -248,16 +226,8 @@ class OcilFileLinker(FileLinker):
         return "{%s}questionnaire" % self.CHECK_NAMESPACE
 
     def link(self):
-        self.tree = parse_xml_file(self.fname)
+        self.tree = ssgcommon.parse_xml_file(self.fname)
         self.tree = self.translator.translate(self.tree, store_defname=True)
-
-
-def parse_xml_file(xmlfile):
-    with open(xmlfile, 'r') as xml_file:
-        filestring = xml_file.read()
-        tree = ElementTree.fromstring(filestring)
-        # print filestring
-    return tree
 
 
 def _find_identcce(rule):
@@ -296,14 +266,9 @@ def create_xccdf_id_to_cce_id_mapping(xccdftree):
     return xccdftocce_idmapping
 
 
-def cce_is_valid(cceid):
-    # IF CCE ID IS IN VALID FORM (either 'CCE-XXXX-X' or 'CCE-XXXXX-X'
-    # where each X is a digit, and the final X is a check-digit)
-    match = re.search(r'CCE-\d{4,5}-\d', cceid)
-    return match is not None
-
-
 def definition_extends_nonexisting_checks(definition, indexed_oval_defs):
+    # TODO: handle multiple levels of referrals.
+    # OVAL checks that go beyond one level of extend_definition won't be properly identified
     for extdefinition in definition.findall(".//{%s}extend_definition" % oval_ns):
         # Verify each extend_definition in the definition
         extdefinitionref = extdefinition.get("definition_ref")
@@ -324,9 +289,6 @@ def definition_extends_nonexisting_checks(definition, indexed_oval_defs):
 def drop_oval_checks_extending_non_existing_checks(ovaltree, indexed_oval_defs):
     # Incomplete OVAL checks are as useful as non existing checks
     # Here we check if all extend_definition refs from a definition exists in local OVAL file
-    #
-    # TODO: handle multiple levels of referrals.
-    # OVAL checks that go beyond one level of extend_definition won't be completely removed
     definitions = ovaltree.find(".//{%s}definitions" % oval_ns)
     defstoremove = set()
     for definition in definitions:
@@ -355,7 +317,7 @@ def check_and_correct_xccdf_to_oval_data_export_matching_constraints(xccdftree, 
     # as specified in Table 16 of XCCDF v1.2 standard:
     # http://csrc.nist.gov/publications/nistpubs/800-126-rev2/SP800-126r2.pdf#page=30&zoom=auto,69,313
     #
-    indexed_xccdf_values = aggregate_elements_by_id(xccdftree, ".//{%s}Value" % (xccdf_ns))
+    indexed_xccdf_values = ssgcommon.map_elements_to_their_ids(xccdftree, ".//{%s}Value" % (xccdf_ns))
 
     # Loop through all <external_variables> in the OVAL document
     ovalextvars = ovaltree.findall(".//{%s}external_variable" % oval_ns)
@@ -425,7 +387,7 @@ def verify_correct_form_of_referenced_cce_identifiers(xccdftree):
             continue
 
         cceid = identcce.text
-        if not cce_is_valid(cceid):
+        if not ssgcommon.cce_is_valid(cceid):
             msg = "The CCE ID {0} is not in valid form.".format(cceid)
             raise RuntimeError(msg)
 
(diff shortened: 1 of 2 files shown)


jan-cerny:
@matejak I really like the refactoring idea and I appreciate using argparse.

shared/utils/relabel-ids.py
```diff
@@ -6,6 +6,8 @@
 import sys
 import os
 
+import argparse as ap
```
jan-cerny: I am not sure if it's needed to rename `argparse` to `ap`.

shared/utils/relabel-ids.py
```diff
@@ -339,6 +337,15 @@ def verify_correct_form_of_referenced_cce_identifiers(xccdftree):
...
+    parser = ap.ArgumentParser(
+        description="This script finds check-content files (currently, "
+        "OVAL and OCIL) referenced from XCCDF and synchronizes all IDs.")
+    parser.add_argument("xccdf_file")
+    parser.add_argument("id_name", help="ID naming scheme")
+    return parser
+
+
 def main():
     if len(sys.argv) < 3:
```
jan-cerny: I think you can move the responsibility for this to argparse as well if you like.

shared/utils/relabel-ids.py
```diff
@@ -51,18 +66,27 @@ def get_checkfiles(checks, checksystem):
...
+        if not checkcontent_hrefarttr_startswith_http_https(checkcontentref):
+            checkfiles.add(checkcontentref.get("href"))
     return checkfiles
 
 
+def _find_identcce(rule):
+    identcce = None
+    for ident in rule.findall("./{%s}ident" % xccdf_ns):
+        if ident.get("system") == cce_uri:
+            identcce = ident
```
mpreisler: this can be much simpler, just return ident here and return None after the loop

shared/utils/relabel-ids.py
```diff
@@ -76,17 +100,15 @@ def create_xccdf_id_to_cce_id_mapping(xccdftree):
...
-                xccdftocce_idmapping[xccdfid] = cceid
+        if xccdfid is None:
+            continue
+
+        identcce = _find_identcce(rule)
+        if identcce is None:
+            continue
+
+        cceid = identcce.text
+        xccdftocce_idmapping[xccdfid] = cceid
```
mpreisler: the extra variable is not necessary here and doesn't make it any more readable

shared/utils/relabel-ids.py
```diff
@@ -102,45 +124,42 @@ def add_cce_id_refs_to_oval_checks(ovaltree, idmappingdict):
...
+        if (ovalid is None
+                or ovaldesc is None
+                or ovalid not in idmappingdict):
+            continue
+
+        xccdfcceid = idmappingdict[ovalid]
+        # IF CCE ID IS IN VALID FORM (either 'CCE-XXXX-X' or 'CCE-XXXXX-X'
+        # where each X is a digit, and the final X is a check-digit)
+        if re.search(r'CCE-\d{4,5}-\d', xccdfcceid) is None:
+            continue
```
mpreisler: I don't like that if I write a wrong CCE it gets silently ignored. is there any step in the build that tells me that I put in an invalid CCE?

shared/utils/relabel-ids.py
```diff
@@ -170,27 +189,19 @@ def ensure_by_xccdf_referenced_oval_def_is_defined_in_oval_file(xccdftree, ovalt
...
         # For local OVAL drop the reference to OVAL definition from XCCDF document
         # in the case:
         # * OVAL definition is referenced from XCCDF file,
         # * But not defined in OVAL file
-        sys.stderr.write("WARNING: OVAL check '%s' was not found, removing "
-                         "<check-content> element from the XCCDF rule.\n"
-                         % xccdfid)
+        print("WARNING: OVAL check '{0}' was not found, removing "
+              "<check-content> element from the XCCDF rule."
+              .format(xccdfid))
```
mpreisler: stderr

shared/utils/relabel-ids.py
```diff
@@ -339,16 +330,53 @@ def verify_correct_form_of_referenced_cce_identifiers(xccdftree):
...
+        description="This script finds check-content files (currently, "
+        "OVAL and OCIL) referenced from XCCDF and synchronizes all IDs.")
+    parser.add_argument("xccdf_file")
+    parser.add_argument("id_name", help="ID naming scheme")
+    return parser
+
 
-    xccdffile = sys.argv[1]
-    idname = sys.argv[2]
+def checkcontent_hrefarttr_startswith_http_https(check_content_ref):
```
mpreisler: super long function name

shared/utils/relabel-ids.py
```diff
@@ -339,16 +330,53 @@ def verify_correct_form_of_referenced_cce_identifiers(xccdftree):
...
+        # @href attribute of <check-content-ref> is required by XCCDF standard
+        msg = "Invalid OVAL <check-content-ref> detected!"
+        raise RuntimeError(msg)
+
+    if hrefattr.startswith("http://") or hrefattr.startswith("https://"):
+        return True
+    return False
+
+
+def assert_that_oxxl_ids_match_rule_id(checks, xccdf_rule):
```
mpreisler: use "check" instead of oxxl, it should be easier to understand

redhatrises:
@matejak any code that can be added to `ssgcommon` that can be reused for other scripts?

shared/utils/relabel-ids.py
matejak: Yes, but do you have any suggestions how to change it? I am somehow OK with it.

matejak:
@redhatrises Good point, some of the validation functions could end up there, but maybe even more will come up.

shared/utils/relabel-ids.py
mpreisler: `check_content_href_is_remote`?

shared/utils/relabel-ids.py
```diff
@@ -483,13 +497,14 @@ def main():
             checkcontentref.set("href", newocilfile)
             checkexport = check.find("./{%s}check-export" % xccdf_ns)
             if checkexport is not None:
-                newexportname = translator.generate_id("{" + oval_ns + "}variable",
+                newexportname = translator.generate_id("{" + ocil_ns + "}variable",
                                                        checkexport.get("export-name"))
```
matejak: This code was broken, maybe it was not used at all? BTW does `check.get("system") != oval_cs` imply `check.get("system") == oval_cs`

shared/utils/relabel-ids.py
mpreisler: If you insist on the long name at least remove the typo ;-)

shared/utils/relabel-ids.py
```diff
@@ -328,32 +324,58 @@ def verify_correct_form_of_referenced_cce_identifiers(xccdftree):
...
-        sys.exit(1)
+def checkcontent_hrefarttr_startswith_http_https(check_content_ref):
+    # Obtain the value of the 'href' attribute of particular <check-content-ref> element
+    hrefattr = check_content_ref.get("href")
+    if hrefattr is None:
+        # @href attribute of <check-content-ref> is required by XCCDF standard
+        msg = "Invalid OVAL <check-content-ref> detected!"
+        raise RuntimeError(msg)
+
+    if hrefattr.startswith("http://") or hrefattr.startswith("https://"):
```
mpreisler: just return the result of this expression

shared/utils/relabel-ids.py
```diff
@@ -375,136 +455,94 @@ def assert_that_check_ids_match_rule_id(checks, xccdf_rule):
...
-                checkexport.set("export-name", newexportname)
+    oval_linker = OvalFileLinker(translator, xccdftree, checks)
+    oval_linker.link()
+    oval_linker.save_linked_tree()
+    oval_linker.link_xccdf()
+
+    ocil_linker = OcilFileLinker(translator, xccdftree, checks)
+    ocil_linker.link()
+    ocil_linker.save_linked_tree()
+    ocil_linker.link_xccdf()
```
mpreisler: so we link the xccdf tree twice?

mpreisler:
This looks awesome, btw. Love the OOP additions. Actually makes it easy to read and reason about.

shared/utils/relabel-ids.py
```diff
@@ -39,6 +41,121 @@
...
+            # Don't attempt to relabel ID on <check-content-ref> element having
+            # its "href" attribute set either to "http://" or to "https://" values
+            if check_content_href_is_remote(checkcontentref):
+                continue
+
+            checkexports = check.findall("./{%s}check-export" % xccdf_ns)
+
+            self._link_xccdf(checkcontentref, checkexports)
+
+    def _link_xccdf(self, checkcontentref, checkexports):
```
mpreisler: rename this to _link_xccdf_checkcontentref

shared/utils/relabel-ids.py
```diff
@@ -317,179 +402,147 @@ def verify_correct_form_of_referenced_cce_identifiers(xccdftree):
...
+    aggregated = {}
+    for element in tree.findall(xpath_expr):
+        element_id = element.get("id")
+        assert element_id is not None
+        aggregated[element_id] = element
+    return aggregated
 
-    # Create XCCDF rule ID to assigned CCE ID mapping
+
+def get_linked_oval_tree(ovaltree, translator, xccdftree):
```
mpreisler: shouldn't this be in the OVAL linker class?

shared/utils/relabel-ids.py
```diff
@@ -76,21 +184,25 @@ def create_xccdf_id_to_cce_id_mapping(xccdftree):
...
+        identcce = _find_identcce(rule)
+        if identcce is None:
+            continue
+
+        xccdftocce_idmapping[xccdfid] = identcce.text
 
     return xccdftocce_idmapping
 
 
+def cce_is_valid(cceid):
```
mpreisler: this might be a good candidate to move to ssgcommon and use in other scripts.

shared/utils/relabel-ids.py
```diff
@@ -73,123 +287,22 @@ def create_xccdf_id_to_cce_id_mapping(xccdftree):
...
-            # The OVAL check was found, we can continue
+    for xccdfid, rule in rules_with_ids_generator(xccdftree):
+        identcce = _find_identcce(rule)
+        if identcce is None:
             continue
 
-        # Search same ID in XCCDF document
-        check = None
-        for el in rule.findall(".//{%s}check" % (xccdf_ns)):
-            if el.get("system") != oval_ns:
```
matejak: I believe that this is a typo and `oval_cs` should be used instead of `oval_ns`. However, this functionality is almost a duplicate of the former `get_checkfiles` function.

shared/utils/relabel-ids.py
```diff
@@ -304,32 +303,35 @@ def cce_is_valid(cceid):
...
-                # There is no oval satisfying the extend_definition referal
-                incomplete = True
-                print("WARNING: OVAL definition '{0}' extends non-existing '{1}', "
-                      "removing it from OVAL definitions."
-                      .format(definition.get("id"), extdefinitionref),
-                      file=sys.stderr)
-        if incomplete:
-            defstoremove.append(definition)
+        if definition_extends_nonexisting_checks(definition, indexed_oval_defs):
+            defstoremove.add(definition)
```
mpreisler: I would expect the warning to be printed here and not in the function. If I am calling a function asking about some state I don't want it to write out messages.

shared/utils/relabel-ids.py
```diff
@@ -223,22 +223,21 @@ def _ensure_by_xccdf_referenced_oval_def_is_defined_in_oval_file(
...
-                # Skip XCCDF rules not referencing OVAL checks
-                continue
-            check = self.checks_related_to_us.pop()
+            for check in rule.findall(".//{%s}check" % (xccdf_ns)):
+                if check.get("system") != oval_cs:
+                    continue
 
-            if get_check_content_ref_if_exists_and_not_remote(check) is not None:
-                continue
+                if get_check_content_ref_if_exists_and_not_remote(check) is None:
```
mpreisler: I am not a big fan of the name of this function. so it returns None if it doesn't exist, it returns None if it's remote, does the logic around it work then?

shared/utils/relabel-ids.py
```diff
@@ -137,60 +137,60 @@ def _get_checkid_string(self):
...
+                "An OVAL rule doesn't have an ID"
+
             ovaldesc = rule.find(".//{%s}description" % self.CHECK_NAMESPACE)
-            # Assign <reference> CCE elements only to those OVAL checks whose
-            # XCCDF rule ID has <ident> CCE set
-            if (ovalid is None
-                    or ovaldesc is None
-                    or ovalid not in idmappingdict):
+            assert ovaldesc is not None, \
+                "Oval rule '{0}' doesn't have a description, which is mandatory".format(ovalid)
```
mpreisler: OVAL

shared/utils/relabel-ids.py
```diff
@@ -33,34 +23,227 @@
...
+        checkcontentref.set("href", self.linked_fname)
+
+        variable_str = "{%s}variable" % self.CHECK_NAMESPACE
+        for checkexport in checkexports:
+            newexportname = self.translator.generate_id(
+                variable_str, checkexport.get("export-name"))
+            checkexport.set("export-name", newexportname)
+
+
+class OvalFileLinker(FileLinker):
```
mpreisler: it's a nitpick but please use "OVAL" not "Oval", it's an acronym

shared/utils/relabel-ids.py
```diff
@@ -33,34 +23,227 @@
...
+                # in the case:
+                # * OVAL definition is referenced from XCCDF file,
+                # * But not defined in OVAL file
+                print("WARNING: OVAL check '{0}' was not found, removing "
+                      "<check-content> element from the XCCDF rule."
+                      .format(xccdfid), file=sys.stderr)
+                rule.remove(check)
+
+
+class OcilFileLinker(FileLinker):
```
mpreisler: same with OCIL

shared/utils/relabel-ids.py
```diff
@@ -73,236 +256,115 @@ def create_xccdf_id_to_cce_id_mapping(xccdftree):
...
+
+    # Loop through all <external_variables> in the OVAL document
+    ovalextvars = ovaltree.findall(".//{%s}external_variable" % oval_ns)
+    if ovalextvars is None:
+        return
+
+    for ovalextvar in ovalextvars:
+        # Verify the found external variable has both 'id' and 'datatype' set
+        if 'id' not in ovalextvar.attrib or 'datatype' not in ovalextvar.attrib:
+            msg = "Invalid OVAL <external_variable> found."
```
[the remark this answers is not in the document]
matejak: Some error messages can't be very specific, because not enough info is available. However, higher level of the code adds info about the OVAL file, where the error originates.

shared/utils/relabel-ids.py
```diff
@@ -33,34 +23,231 @@
...
+        """
+        ovalrules = self.tree.findall(".//{0}".format(self._get_checkid_string()))
+        for rule in ovalrules:
+            ovalid = rule.get("id")
+            assert ovalid is not None, \
+                "An OVAL rule doesn't have an ID"
+
+            ovaldesc = rule.find(".//{%s}description" % self.CHECK_NAMESPACE)
+            assert ovaldesc is not None, \
+                "Oval rule '{0}' doesn't have a description, which is mandatory".format(ovalid)
```
mpreisler: OVAL

shared/utils/relabel-ids.py
```diff
@@ -73,236 +260,118 @@ def create_xccdf_id_to_cce_id_mapping(xccdftree):
...
+    if ovalextvars is None:
+        return
+
+    for ovalextvar in ovalextvars:
+        # Verify the found external variable has both 'id' and 'datatype' set
+        if 'id' not in ovalextvar.attrib or 'datatype' not in ovalextvar.attrib:
+            msg = "Invalid OVAL <external_variable> found - either without 'id' or 'datatype'."
+            raise HighLevelError(msg)
+        # Obtain the 'id' and 'datatype attribute values
+        if 'id' in ovalextvar.attrib and 'datatype' in ovalextvar.attrib:
```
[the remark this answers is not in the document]
matejak: I agree.

matejak:
The code is now in a better shape, but it still has to be refactored in a non-trivial way to attain reachable perfection. For example, not all errors produce really helpful messages, so I have decided to re-raise all exceptions for the time being.

mpreisler:
ACK

PR #2705, opened 2018-04-04
Add tests for verifying file permissions and hashes with RPM

mildass:
#### Description:
Add two tests, that should pass on fresh system, and one script, after which should test fail.

Tests for:
- Verify and Correct File Permissions with RPM rule
- Verify File hashes with RPM rule

Script for:
- Verify and Correct File Permissions with RPM rule

### DIFF
--- a/tests/data/group_system/group_software/group_integrity/group_software-integrity/group_rpm_verification/rule_rpm_verify_hashes/fresh_system.pass.sh
+++ b/tests/data/group_system/group_software/group_integrity/group_software-integrity/group_rpm_verification/rule_rpm_verify_hashes/fresh_system.pass.sh
@@ -2,6 +2,8 @@
 #
 # profiles = xccdf_org.ssgproject.content_profile_pci-dss
 
-# we expect the system is in good shape
-# TODO: add check here, so this script fails when not true
-true
+# verify (-V) all installed (-a) packages digests,
+# if digest differs from package metadata, then attribute '5' is printed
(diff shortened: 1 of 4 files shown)


jan-cerny:
@openscap-ci add to whitelist

mpreisler:
NACK. Fix your git author in the commit:
```
From: Milan Lysonek <mlysonek@localhost.localdomain>
```

tests/data/group_system/group_software/group_integrity/group_software-integrity/group_rpm_verification/rule_rpm_verify_hashes/fresh_system.pass.sh
```diff
@@ -2,6 +2,11 @@
 #
 # profiles = xccdf_org.ssgproject.content_profile_pci-dss
 
-# we expect the system is in good shape
-# TODO: add check here, so this script fails when not true
-true
+# verify all packages from rpm 
+result=$(rpm -Va | grep '^..5')
+
+if [ -z "$result" ]; then
```
mpreisler: this is a bash anti-pattern. you are running `[` which returns 0 exit code or 1, 0 for true. then you check that it's 0 (true) and then you run true wihch again exists with 0. you can simplify this to just running `[` and that's all.

dahaic:
Agree, tester should prepare the machine to pass. It's more efficient that way.

PR #2710, opened 2018-04-06
OVAL: audit_rules_login_events optional trailing slash

### LINKED ISSUE #2709: RHEL-07-030610 audit_rules_login_events_faillock fails after remediation
#### Description of problem:
Apply remediation in check content of RHEL-07-030610 (V1R4), check still fails.

0.1.38

RHEL 7.4

#### Steps to Reproduce:

1.
2.
3.
4.

#### Actual Results:
xccdf_org.ssgproject.content_rule_audit_rules_login_events_faillock: fail

#### Expected Results:
xccdf_org.ssgproject.content_rule_audit_rules_login_events_faillock: success

The STIG does not specify a trailing slash.  The check should make it optional.

jamescassell:
#### Description:

Makes the trailing slash optional for the directory watch.

Fixes #2709 

#### Rationale:

It should not matter whether the audit watch on a directory is specified with or without a trailing slash.  The RHEL 7 STIG V1R4 shows it without a trailing slash.  This change makes both options pass.

### DIFF
--- a/shared/templates/create_audit_rules_login_events.py
+++ b/shared/templates/create_audit_rules_login_events.py
@@ -20,7 +20,7 @@ class AuditRulesLoginEventsGenerator(FilesGenerator):
                 "./template_OVAL_audit_rules_login_events",
                 {
                     "%NAME%":	name,
-                    "%PATH%":	path.replace("/", "\\/")
+                    "%PATH%":	re.sub('/$', '/?', path).replace("/", "\\/")
                 },
                 "./oval/audit_rules_login_events_{0}.xml", name
             )


jan-cerny:
@openscap-ci OK to test

redhatrises:
@jamescassell thanks for your PR! Really appreciate it. Unfortunately, the trailing slash is actually required as it will have auditd actually audit the events in the directory and not treat the directory itself as a file. Originally, the fixtext was incorrect in V1R1, but looks like V1R4 is now incorrect of how this event should be handled.

jamescassell:
Should this be reported as a bug in DISA content, then?

redhatrises:
@jamescassell yes. Without having the trailing slash, you are blind to anything that happens in the directory.

redhatrises:
@jamescassell I was surprised when I learned this as well.

jamescassell:
@redhatrises I opened a case here: https://access.redhat.com/support/cases/#/case/02072096

I'll close this PR now.

jamescassell:
I've now tested this.  Without the trailing slash, audit.log still ends up with an entry whenever a user tries to log in, both for users who have never logged in, and for users who have successfully logged in.  faillock does not, as far as I can tell, write directories to /var/run/faillock, only files with the name of the user.

This seems to meet the intention of RHEL-07-030610.  Re-opening.

jamescassell:
@redhatrises , I did more testing, and the audit system does behave as the man page describes.  Specifically, with a rule of:
```
-w /var/run/faillock -p wa -k identity
```
changes to arbitrarily deep directories and files under the /var/run/faillock directory will be logged to the audit log.  This is on RHEL 7.4.  Is there another platform where this is not the case?  If so, could we make this change specific to RHEL 7.

jamescassell:
That is to say, the trailing slash is not required on RHEL 7.4, per my testing.

jamescassell:
@redhatrises can you reconsider this PR?

Per Red Hat support on the above ticket:
```
Yes, "auditctl -w /etc -p wa" and "auditctl -w /etc/ -p wa" are having same functionality.
```
and
```
it doesn't matter for "trailing slash"
```

redhatrises:
@jamescassell totally will be looking at it. I just haven't been able to review it yet.

redhatrises:
@jamescassell please make the changes in `audit_rules_login_events.csv` and not `create_audit_rules_login_events.py`

jamescassell:
@redhatrises wouldn't that break the basename transformation?

jamescassell:
...   I suppose I could rather just strip the trailing slash from the .csv file, but then check would fail compliant configurations that do have the trailing slash.

jamescassell:
tested it with adding a trailing `?` after the `/`.  oscap complains of an invalid datastream, though the build did not fail.  Do you prefer to simply strip the tailing slash from the csv?

redhatrises:
@jamescassell yes. Please remove from csv.

PR #2712, opened 2018-04-06
Remove ignore option for auditing configuration

redhatrises:
#### Description:

- Remove ignore option for auditing configuration

#### Rationale:

- The ignore option is not really an acceptable security practice


shawndwells:
good catch

PR #2713, opened 2018-04-06
[issue 2571] update OSPP profile name and description

shawndwells:
Update profile name to reflect national checklist program submission /
remove draft.

Updated profile description language and formatting to ease readability.

At some time this profile name should be removed -- but that will break some RHEL7 scripts.
Purposefully leaving the name consistent for now.

resolves
https://github.com/OpenSCAP/scap-security-guide/issues/2571

### DIFF
--- a/rhel7/profiles/ospp-rhel7.profile
+++ b/rhel7/profiles/ospp-rhel7.profile
@@ -1,20 +1,35 @@
 documentation_complete: true
 
-title: 'United States Government Configuration Baseline (USGCB / STIG) - DRAFT'
+title: 'United States Government Configuration Baseline'
 
-description: "This profile is developed in partnership with the \nU.S. National Institute of Standards and Technology (NIST),\
-    \ U.S. Department of \nDefense, the National Security Agency, and Red Hat. The USGCB is intended\nto be the core set of\
-    \ security related configuration settings by which all\nfederal agencies should comply.\n\nThis baseline implements configuration\
-    \ requirements from the following documents:\n\n- Committee on National Security Systems Instruction No. 1253 (CNSSI 1253)\n\
-    - NIST Controlled Unclassified Information (NIST 800-171)\n- NIST 800-53 control selections for MODERATE impact systems\
-    \ (NIST 800-53)\n- U.S. Government Configuration Baseline (USGCB)\n- NIAP Protection Profile for General Purpose Operating\
-    \ Systems v4.0 (OSPP v4.0)\n- DISA Operating System Security Requirements Guide (OS SRG)\n\nFor any differing configuration\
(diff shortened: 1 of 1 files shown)


shawndwells:
@mpreisler yes, agreed the ``-rhel7`` is redundant. If we change in 0.1.39, will that effect downstream releases? Trying to be sensitive to people who setup RHEL7 scripts and may find them broken with RHEL 7.5.1 / RHEL 7.future.

shawndwells:
aka change now or create tracking issue for mass rename in something like 0.1.40

mpreisler:
Unfortunately it will, one way to fix it is to create a profile with the old name that extends the profile with the new name with no other changes. It's ugly though. We have to decide whether we are OK with breaking downstream.

shawndwells:
Was reminded that downstreams follow a different release process than mass rebuild. 

Agreed with @mpreisler on the profile renames. Will open separate PR for that.

mpreisler:
@openscap-ci test this please

shawndwells:
note - this needs to be merged prior to https://github.com/OpenSCAP/scap-security-guide/pull/2714 to help reduce conflicts

mpreisler:
@openscap-ci test this please

rhel7/profiles/ospp-rhel7.profile
```diff
@@ -1,20 +1,35 @@
...
+    \ security setting was chosen. Security Requirement Traceability Guides (RTMs) and \n
+    \ sample System Security Configuration Guides are provided via the \n
+    \ scap-security-guide-docs package. \n
+    \ \n
+    \ This profile reflects U.S. Government consensus content and is developed through \n
+    \ the OpenSCAP/SCAP Security Guide initiative, championed by the National \n
+    \ Security Agency. Except for differences in formatting to accommodate \n
+    \ publishing processes, this profile mirrors OpenSCAP/SCAP Security Guide \n
+    \ content as minor divergences, such as bugfixes, work through the \n 
+    \ consensus and release processes."
```
mpreisler: could you please reflow this to the `|-` yaml syntax? We use it everywhere else, makes it easier to edit because you don't have to escape newlines. I think the prefix escapes are there because the original description contains superfluous whitespace there.

shawndwells: These PRs use the existing format\line breaks. Converting them to a new format be another PR, where we can handle this across multiple files in the repo.

mpreisler: ok

mpreisler:
ACK

shawndwells:
thanks, @mpreisler !

PR #2727, opened 2018-04-09
Mark draft profiles as "documentation_complete: false"

shawndwells:
Resolves https://github.com/OpenSCAP/scap-security-guide/issues/2719


redhatrises:
Ack

PR #2731, opened 2018-04-10
[DO NOT MERGE] Example OVAL YAML file

redhatrises:
This is an example OVAL YAML and not intended to be merged. 

Look it over. Provide feedback. What can be made easier? Can test, object, and state be made
into something more understandable?

### DIFF
--- a/root_path_no_dot.oval
+++ b/root_path_no_dot.oval
@@ -1,0 +1,49 @@
+---
+documentation_complete: true
+title: Ensure that No Dangerous Directories Exist in Root's Path
+primary_id: root_path_no_dot
+check_type: compliance
+family: unix
+version: 1
+platforms:
+  - Red Hat Enterprise Linux 7
+  - multi_platform_fedora
+
+description: |
+  The environment variable PATH should be set correctly for the root user.
+
+operation:
+  - and: begins_colon_period, begins_doublecolon
+
+test:
+  - id: begins_colon_period
+  - type: environmentvariable
+  - comment:
+  - check: none satisfy
+
+  state:
+    - comment:
+    - regex: ^[:\.]
+    - operation: match
+
+object:
+    - type: environmentvariable
+    - id: root_path_no_dot
+    - map_to: begins_colon_period, begins_doublecolon
+    - comment:
+    - pid: nil
+      type: int
+    - name: PATH
+
+test:
+  - id: begins_doublecolon
+  - type: environmentvariable
+  - comment:
+  - check: none satisfy
+
+  state:
+    - type: environmentvariable
+    - comment:
+    - regex: ::
+    - operation: match
+


root_path_no_dot.oval
```diff
@@ -0,0 +1,49 @@
...
+version: 1
+platforms:
+  - Red Hat Enterprise Linux 7
+  - multi_platform_fedora
+
+description: |
+  The environment variable PATH should be set correctly for the root user.
+
+operation:
+  - and: begins_colon_period, begins_doublecolon
```
jamescassell: Make this into a list?

root_path_no_dot.oval
```diff
@@ -0,0 +1,49 @@
...
+title: Ensure that No Dangerous Directories Exist in Root's Path
+primary_id: root_path_no_dot
+check_type: compliance
+family: unix
+version: 1
+platforms:
+  - Red Hat Enterprise Linux 7
+  - multi_platform_fedora
+
+description: |
```
jamescassell: Do we want to preserve all newlines in the description?

root_path_no_dot.oval
```diff
@@ -0,0 +1,49 @@
...
+  The environment variable PATH should be set correctly for the root user.
+
+operation:
+  - and: begins_colon_period, begins_doublecolon
+
+test:
```
jamescassell: make it plural?

root_path_no_dot.oval
```diff
@@ -0,0 +1,49 @@
...
+  - id: begins_colon_period
+  - type: environmentvariable
```
jamescassell: I'd only put the list marker on the first line of each list item.

root_path_no_dot.oval
```diff
@@ -0,0 +1,49 @@
...
+  - comment:
+  - check: none satisfy
+
+  state:
+    - comment:
+    - regex: ^[:\.]
+    - operation: match
+
+object:
```
jamescassell: Duplicate keys are prohibited by YAML.

root_path_no_dot.oval
```diff
@@ -0,0 +1,49 @@
...
+    - type: environmentvariable
+    - id: root_path_no_dot
+    - map_to: begins_colon_period, begins_doublecolon
+    - comment:
+    - pid: nil
+      type: int
+    - name: PATH
+
+test:
```
jamescassell: duplicate key.

root_path_no_dot.oval
jamescassell: or could we use `all:` and `any:`?

root_path_no_dot.oval
jamescassell: this one wasn't duplicate, but the `test:` one was.

root_path_no_dot.oval
jamescassell: also it's a duplicate key.

root_path_no_dot.oval
redhatrises: @jamescassell probably would have something like:
```complex: (something and something) or something else```

@jamescassell is your vision for all and any?

Comment (dahaic) [reply root_path_no_dot.oval]: Do we have to reinvent this part at all? I think it's better to stick to OVAL nomenclature to make it easier to translate ideas. What about this?

```
criteria:
  - and:
    - a
    - b
    - or:
      - c
      - d
  - c
```

**Comment by jamescassell** [comment]
Unfortunately, I'm not super familiar with oval checks, so my comments may not be fully informed.

**Comment by jan-cerny** [comment]
I think it should more follow more the OVAL structure and nomenclature, because we need to be able to convert OVALs to the new format automatically. And generate OVALs from YAML as well.

I think there needs to be `tests`, `objects`, `variables` and `states` sections, because single OVAL definition can consist of multiple tests. The tests can share objects. States doesn't have to be used as states, but as filters., ...

**Comment by mpreisler** [comment]
I am undecided about this but I feel that with the move of OVAL to yaml we lose the ability to direct people to OVAL code samples, specification and references. We will have to write these ourselves because none of the existing XML OVAL docs will apply.

Not sure if that's worth it.

PS: I also have a feeling that in this PR we will discover that YAML sucks more than we thought it did ;-)

**Comment by redhatrises** [comment]
> I think it should more follow more the OVAL structure and nomenclature, because we need to be able to convert OVALs to the new format automatically. And generate OVALs from YAML as well.
> 
> I think there needs to be tests, objects, variables and states sections, because single OVAL definition can consist of multiple tests. The tests can share objects. States doesn't have to be used as states, but as filters., ...

@jan-cerny true, but should they be called tests/objects/states, or is there something easier that non-experts (such as ourselves) would understand?

> PS: I also have a feeling that in this PR we will discover that YAML sucks more than we thought it did ;-)

hahaha @mpreisler I feel as if we have had this conversation before especially since we could find issues with many differing formats. :)

File: root_path_no_dot.oval (hunk 1 above)
Comment (redhatrises) [reply root_path_no_dot.oval]: @dahaic how would you handle the following then:
```
- and:
    - a
    - or:
      - c
      - d
    - or:
      - e
      - f
```

Comment (dahaic) [reply root_path_no_dot.oval]: @redhatrises I am confused, yaml can handle that just fine:

```
[{'and': ['a', {'or': ['c', 'd']}, {'or': ['e', 'f']}]}]
```

dahaic:
What about having multi document yaml files? With every test/object/state/value being different documents (separated by `---`). That would make processing of the yaml file simpler, in my opinion.

Also having `state` within `test` doesn't scale, in my opinion - what is the use case?

matejak:
My understanding is that OVAL XML is hierarchical data, which can be represented in another forms, and YAML also has the hierarchical nature, which would make it a good candidate. However, there is a risk that a representation that looks good may have some flaws that are not obvious right away.
Therefore, I think that it is more practical to start the efforts from another direction - by having an intermediate representation of OVAL data in form of a Python data structure. I am aware that this is quite difficult though, as the OVAL schema is huge. However, it would enable parsers and/or formatters of this data structure from/to other formats - for instance YAML.
In that mode of development, the final format of YAML would be guided by the implementation process, especially by the requirement that the format has to be isomorphic with the data structure it represents.

jan-cerny:
Thank you for your proposal. We have considered this change. We think that the benefit is small, because it would not add any new high-level concept and the work on converting all OVALs to a different format is extensive.

PR #2732, opened 2018-04-10
Remove common profile and use standard profile instead

### LINKED ISSUE #2569: rhel 7.5 beta: difference between "standard security profile" and "common profile" ?
they appear to be the same by description. can one be deleted?

image

redhatrises:
#### Description:

- Remove common profile and use standard profile instead

#### Rationale:

- Having the common and standard profile doesn't make much sense as there really isn't anything
  uniquely different about them besides the profile titles and descriptions. This standarizes the general
  profile to the "standard" profile which makes more sense as common is confusing and really a
  duplicate of the "standard" profile.

- Fixes #2569


shawndwells:
Long time coming. Thanks!

PR #2735, opened 2018-04-11
Simplify profile ids

mpreisler:
#### Description:

- Some of the profile IDs in RHEL7 are unnecessarily long. They contain "rhel7" in the ID which is not necessary because the user is using the RHEL7 benchmark, what else are the profiles gonna be for... It's just bad UX to force users to type out long IDs.

If the community agrees we can rename more profiles.


mpreisler:
> This is nice. Does it affect our documentation on Portal, tutorials, user manuals, man pages of other components?

yes, most likely. this also breaks old docs, code samples, talk slides and stuff like that

shawndwells:
Had to happen some time. 

How do we properly tip off downstream packagers? Not sure who maintains the Ubuntu spins. The OEL maintainers are active (@iokomin). As are partners like @trevor-vaughan / SIMP.

Updating the README with a temporary notice might be appropriate.

trevor-vaughan:
@shawndwells Thanks for the heads up! I *think* we're OK but I'm not 100% sure since we might have to use this to reference the scans that we want to run. We're actually trying to work around some other broken profile names that seemed to happen mid-EL7. It's a bit difficult to automate when things keep getting renamed.

shawndwells:
Updating docs can be another PR. 

Thanks for doing this @mpreisler

PR #2736, opened 2018-04-11
Don't add profiles if documentation_complete is false

redhatrises:
`yaml-to-shorthand.py` was not properly ignoring profiles when `documentation_complete` was set to `false`. This fixes that.


redhatrises:
Closing and opening a new PR.

PR #2737, opened 2018-04-11
Don't load yaml booleans as python booleans

redhatrises:
#### Description:

- Don't load yaml booleans as python booleans

#### Rationale:

- Booleans were being loaded as `True`, `False`, etc.

### DIFF
--- a/shared/utils/yaml-to-shorthand.py
+++ b/shared/utils/yaml-to-shorthand.py
@@ -16,7 +16,17 @@ except ImportError:
     import cElementTree as ET
 
 
+def bool_constructor(self, node):
+    value = self.construct_yaml_bool(node)
+    if value == False:
+        return 'false'
+    else:
+        return 'true'
+
 def open_yaml(yaml_file):
+    # Don't follow python bool case
+    yaml.Loader.add_constructor(u'tag:yaml.org,2002:bool', bool_constructor)
+
     with codecs.open(yaml_file, "r", "utf8") as stream:
         yaml_contents = yaml.load(stream)
         if "documentation_complete" in yaml_contents and \
--- a/shared/utils/yaml-to-shorthand.py
+++ b/shared/utils/yaml-to-shorthand.py
@@ -17,16 +17,8 @@ except ImportError:
     import cElementTree as ET
 
 
-def construct_yaml_bool(self, node):
-    value = self.construct_scalar(node)
-
-    return value
-
-
 def bool_constructor(self, node):
-    value = construct_yaml_bool(self, node)
-
-    return value
+    return self.construct_scalar(node)
 
 
 def open_yaml(yaml_file):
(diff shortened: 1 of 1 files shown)


matejak:
What is the reason for this change? Do we have `true`/`false` values in a non-boolean context? I am uncomfortable with silent mass conversion of any passed boolean value to a string.

matejak:
@openscap-ci test this please

redhatrises:
@openscap-ci test this please

matejak:
I don't think that we have to hack the yaml format parsing - those `true`, `on` etc. values [more on those here](http://yaml.org/type/bool.html) can be protected by double quotes, so `"true"` will be parsed as strings.
I think that raising an exception when an input value is expected to be string, but it is boolean would lead to a more strict requirements on input, which could be beneficial for the project.

redhatrises:
@matejak I would like to not have to have the quotes if possible and not to have that nuance that could be confusing to some people. It would also make it easier ignore case for whatever type of boolean and not have to remember to have quotes.

matejak:
OK, let's collect opinions of more people than just two of us, this is quite a substantial change.

redhatrises:
Agreed. I updated the code to handle any case of booleans e.g. `true`, `True`, `false`, `on`, `Off`, etc.

shared/utils/yaml-to-shorthand.py
```diff
@@ -10,13 +10,29 @@
...
+def construct_yaml_bool(self, node):
+    value = self.construct_scalar(node)
+
+    return value
+
+
+def bool_constructor(self, node):
+    value = construct_yaml_bool(self, node)
+
+    return value
```
mpreisler: this is python code, not a Tolkien novel :-P Please do this on just 2 lines.

redhatrises: But @mpreisler I like Tolkien novels... maybe that is why it is written as such. :D

redhatrises: @mpreisler Tolkien novel reduced to Ernest Hemingway's shortest novel.

dahaic:
@redhatrises, I agree with @matejak. If we say we use yaml, we should stick to the specification. It might be less user friendly, but then we can just recommend to quote everything and be done with it.

redhatrises:
Remember that we are doing a YAML-like conversion, so we don't actually have to stick 100% with yaml standards. In fact, we may not fully be able to. Plus, this is going to be documented in the developer guide. Remember the easier that we can make this, without having all the nuances, the better for everyone.

dahaic:
I have changed my mind. This YAML "feature" is quite obscure anyway, and has probably zero usage in SSG. I am for accepting this :)

shawndwells:
I don't care as long as it's documented somewhere

dahaic:
Thanks @matejak for challenging the idea :) Good stuff, @redhatrises !! Ack.

redhatrises:
Yes. Thanks @matejak for challenging the idea as well.

PR #2740, opened 2018-04-12
Minor in size, but substantial test suite improvements.

matejak:
* Don't clean HTML files by default - Fixes the issue when HTML reports of profile runs were wiped out.
* Match profile IDs by suffix as the rest of the ecosystem. - Helps when the ...C2S profile is selected, because the ...C2S-docker profile is the superstring of the former.


matejak:
@openscap-ci jenkins test now

yuumasato:
Ack, thank you!

PR #2743, opened 2018-04-13
Fix typo in set_faillock_option_to_value_in_pam_file.sh

redhatrises:
#### Description:

- `_pamfile` should be `_pamFile`

#### Rationale:

- Causing faillock remediation scripts to error.


shawndwells:
Good find. Thanks!

PR #2744, opened 2018-04-13
Not escaping / character breaks perform_audit_rules_privileged_commands_remediation.sh

### LINKED ISSUE #2582: rhel 7.5 beta: audit_rules_privileged_commands still fails after remediation
side note: why is this still selected in profiles? wasn't this broken out into smaller rules?

redhatrises:
#### Description:

- Not escaping / character breaks perform_audit_rules_privileged_commands_remediation.sh

#### Rationale:

- See above.

- Fixes #2582


redhatrises:
@matejak might have a better fix for this.

shawndwells:
This PR does the job. If there's a more elegant fix, another PR can be opened :)

PR #2745, opened 2018-04-13
Simplify display_login_attempts remediation script logic

### LINKED ISSUE #2483: pam remediations - bad fix logic
This issue was reported by @nathanpeters008 [in the openscap project](https://github.com/OpenSCAP/openscap/issues/924), but it is an SSG bug. Here is the original text for this part of the issue:

Consider the fix code:

```
if ! `grep -q ^[^#].*pam_succeed_if.*showfailed /etc/pam.d/postlogin` ; then
  if ! grep `^session.*pam_succeed_if.so /etc/pam.d/postlogin` ; then
    echo "session     [default=1]   pam_lastlog.so nowtmp showfailed" >> /etc/pam.d/postlogin
    echo "session     optional      pam_lastlog.so silent noupdate showfailed" >> /etc/pam.d/postlogin
  else
    sed -i '/^session.*pam_succeed_if.so/a session\t    optional\t  pam_lastlog.so silent noupdate showfailed' /etc/pam.d/postlogin
    sed -i '/^session.*pam_succeed_if.so/a session\t    [default=1]\t  pam_lastlog.so nowtmp showfailed' /etc/pam.d/postlogin
  fi
else
  sed -i "s/session[ ]*\[default=1][ ]*pam_lastlog.so.*/session     [default=1]   pam_lastlog.so nowtmp showfailed/g" /etc/pam.d/postlogin
  sed -i "s/session[ ]*optional[ ]*pam_lastlog.so.*/session     optional      pam_lastlog.so silent noupdate showfailed/g" /etc/pam.d/postlogin
fi
```

In the above code, the if-then logic is faulty. 
Here are the contents of a default pam `postlogin-ac` file:
```
#%PAM-1.0
# This file is auto-generated.
# User changes will be destroyed the next time authconfig is run.


session     [success=1 default=ignore] pam_succeed_if.so service !~ gdm* service !~ su* quiet
session	    [default=1]	  pam_lastlog.so nowtmp showfailed
session	    optional	  pam_lastlog.so silent noupdate showfailed
```
Using the above logic, line 1 will never match because the showfailed directive is on lines 2 and 3, not on the pam_succeed_if line.
Line 2, when it finds pam_succeed_if, will always append the 2 lines, without bothering to check if they already exist in the file.
This means the file will have 2 copies of each directive, like so:
```
session     [success=1 default=ignore] pam_succeed_if.so service !~ gdm* service !~ su* quiet
session	    [default=1]	  pam_lastlog.so nowtmp showfailed
session	    optional	  pam_lastlog.so silent noupdate showfailed
session	    [default=1]	  pam_lastlog.so nowtmp showfailed
session	    optional	  pam_lastlog.so silent noupdate showfailed
```

As a result of these issues, a user will always see the last login time twice due to the extra and unnecessary copies of `pam_lastlog.so` in this file.

redhatrises:
#### Description:

- Simplify display_login_attempts remediation script logic

#### Rationale:

- Logic was for an outdated version of RHEL

- Fixes #2483

### DIFF
--- a/shared/fixes/bash/display_login_attempts.sh
+++ b/shared/fixes/bash/display_login_attempts.sh
@@ -1,10 +1,8 @@
 # platform = Red Hat Enterprise Linux 7
 
-if ! `grep -q "^session.*pam_lastlog.so" /etc/pam.d/postlogin` ; then
-	echo "session     [default=1]   pam_lastlog.so nowtmp showfailed" >> /etc/pam.d/postlogin
-	echo "session     optional      pam_lastlog.so silent noupdate showfailed" >> /etc/pam.d/postlogin
-else
+if $(grep -q "^session.*pam_lastlog.so" /etc/pam.d/postlogin) ; then
 	sed -i --follow-symlinks "/pam_lastlog.so/d" /etc/pam.d/postlogin
-	echo "session     [default=1]   pam_lastlog.so nowtmp showfailed" >> /etc/pam.d/postlogin
-        echo "session     optional      pam_lastlog.so silent noupdate showfailed" >> /etc/pam.d/postlogin
 fi
+
+echo "session     [default=1]   pam_lastlog.so nowtmp showfailed" >> /etc/pam.d/postlogin
+echo "session     optional      pam_lastlog.so silent noupdate showfailed" >> /etc/pam.d/postlogin
--- a/shared/fixes/bash/display_login_attempts.sh
+++ b/shared/fixes/bash/display_login_attempts.sh
@@ -1,14 +1,10 @@
 # platform = Red Hat Enterprise Linux 7
 
-if ! `grep -q ^[^#].*pam_succeed_if.*showfailed /etc/pam.d/postlogin` ; then
-  if ! grep `^session.*pam_succeed_if.so /etc/pam.d/postlogin` ; then
-    echo "session     [default=1]   pam_lastlog.so nowtmp showfailed" >> /etc/pam.d/postlogin
-    echo "session     optional      pam_lastlog.so silent noupdate showfailed" >> /etc/pam.d/postlogin
-  else
-    sed -i --follow-symlinks '/^session.*pam_succeed_if.so/a session\t    optional\t  pam_lastlog.so silent noupdate showfailed' /etc/pam.d/postlogin
-    sed -i --follow-symlinks '/^session.*pam_succeed_if.so/a session\t    [default=1]\t  pam_lastlog.so nowtmp showfailed' /etc/pam.d/postlogin
-  fi
+if ! `grep -q "^session.*pam_lastlog.so" /etc/pam.d/postlogin` ; then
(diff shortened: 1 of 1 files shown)


shared/fixes/bash/display_login_attempts.sh
```diff
@@ -1,14 +1,10 @@
...
 
-if ! `grep -q ^[^#].*pam_succeed_if.*showfailed /etc/pam.d/postlogin` ; then
-  if ! grep `^session.*pam_succeed_if.so /etc/pam.d/postlogin` ; then
-    echo "session     [default=1]   pam_lastlog.so nowtmp showfailed" >> /etc/pam.d/postlogin
-    echo "session     optional      pam_lastlog.so silent noupdate showfailed" >> /etc/pam.d/postlogin
-  else
-    sed -i --follow-symlinks '/^session.*pam_succeed_if.so/a session\t    optional\t  pam_lastlog.so silent noupdate showfailed' /etc/pam.d/postlogin
-    sed -i --follow-symlinks '/^session.*pam_succeed_if.so/a session\t    [default=1]\t  pam_lastlog.so nowtmp showfailed' /etc/pam.d/postlogin
-  fi
+if ! `grep -q "^session.*pam_lastlog.so" /etc/pam.d/postlogin` ; then
```
matejak: I suggest getting rid of the negation - you can remove the negating `!` and swap the `then` and `else` bodies.
Next, using `$(...)` instead of `` `...` `` is considered a better practice.

shared/fixes/bash/display_login_attempts.sh
```diff
@@ -1,14 +1,10 @@
...
+	echo "session     [default=1]   pam_lastlog.so nowtmp showfailed" >> /etc/pam.d/postlogin
+	echo "session     optional      pam_lastlog.so silent noupdate showfailed" >> /etc/pam.d/postlogin
 else
-  sed -i --follow-symlinks "s/session[ ]*\[default=1][ ]*pam_lastlog.so.*/session     [default=1]   pam_lastlog.so nowtmp showfailed/g" /etc/pam.d/postlogin
-  sed -i --follow-symlinks "s/session[ ]*optional[ ]*pam_lastlog.so.*/session     optional      pam_lastlog.so silent noupdate showfailed/g" /etc/pam.d/postlogin
+	sed -i --follow-symlinks "/pam_lastlog.so/d" /etc/pam.d/postlogin
+	echo "session     [default=1]   pam_lastlog.so nowtmp showfailed" >> /etc/pam.d/postlogin
+        echo "session     optional      pam_lastlog.so silent noupdate showfailed" >> /etc/pam.d/postlogin
```
matejak: This line is space-indented, unlike others that have tabs.
Moreover, those two lines are identical, so you can consider defining a function that appends those two lines to `postlogin`.
Either way, you can have one `if` with sed delete (i.e. without `else`), and just append those lines after that `if` block.

redhatrises:
Thanks @matejak. Changes made.

matejak:
Great, thank you!

PR #2748, opened 2018-04-16
Fixed OVAL check exports.

matejak:
* OVAL variables needed for checks are autodetected from OVAL definitions
  and added to as check-exports to OVAL checks.
  Therefore, those exports can end up in output XCCDF files.

* Added oval-parsing functionality.

This is an alternative and hopefully better approach to fix #2703.


shared/utils/parse_oval.py
```diff
@@ -0,0 +1,141 @@
+
```
redhatrises: Please move this module to `shared/modules`.

dahaic:
I have checked results, and variables are (looks ;) ) fixed. Thanks @matejak !!!

PR #2749, opened 2018-04-16
Fix accounts_passwords_pam_faillock_deny.sh script

redhatrises:
#### Description:

- Fix accounts_passwords_pam_faillock_deny.sh script

#### Rationale:

- Script was referencing wrong variable and setting.


shawndwells:
Lovely.

PR #2764, opened 2018-04-17
Add Anaconda Kdump disable script

redhatrises:


shared/fixes/anaconda/service_kdump_disable.anaconda
```diff
@@ -0,0 +1,3 @@
+# platform = multi_platform_rhel
```
shawndwells: fedora too?

redhatrises:
@shawndwells OAA has to be updated to version 0.9... same as RHEL for this to work.

PR #2770, opened 2018-04-18
Added tests for checking for bootloader password protection.

matejak:
The tests work as expected, except that the remediaion is not implemented (which is done on purpose).


yuumasato:
If it is expected there won't be remediation you can add following line into test scenarios
```
# remediation = none
```
This will tell SSGTestSuite to expect rule to not be fixed.

matejak:
Great, it works like a charm now.

yuumasato:
Ack, thank you!

PR #2771, opened 2018-04-18
Document minimum Ansible version in User/Developer Guides

### LINKED ISSUE #2467: Add requirement on ansible version needed for SSG remediations

redhatrises:
#### Description:

- Document minimum Ansible version in User/Developer Guides

#### Rationale:

- Need to document minimum ansible version for users and developers as some users have used
  versions of Ansible that do not work with the playbooks that are generated.

- Fixes #2467

### DIFF
--- a/docs/manual/developer_guide.adoc
+++ b/docs/manual/developer_guide.adoc
@@ -689,6 +689,8 @@ Remediations also carry metadata that should be present at the beginning of the 
 
 ==== Ansible
 
+IMPORTANT: The minimum version of Ansible must be at least version 2.3
+
 Ansible remediations are stored as yml files in directory _/template/static/ansible_ under the targeted platform. They are meant to be executed by Ansible itself when requested by openscap, so they are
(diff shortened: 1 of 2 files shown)


dirtyharrycallahan:
For end users it is more important to know which Ansible versions work rather than the minimum version. As someone who has used Ansible for a few years Ansible is a moving target and regressions are not uncommon.

For example Red Hat 7:
ansible-2.3.1.0-3.el7.noarch.rpm
ansible-2.3.2.0-2.el7.noarch.rpm
ansible-2.4.0.0-5.el7.noarch.rpm
ansible-2.4.1.0-1.el7.noarch.rpm
ansible-2.4.2.0-2.el7.noarch.rpm

and the latest from EPEL7:
ansible-2.5.1-1.el7.noarch.rpm

PR #2772, opened 2018-04-18
Extended definitions contribute their variables

dahaic:
#### Description:

- To identify which variables are needed within particular xccdf rule, we have to delve into depths of extended definitions as well.

#### Rationale:

- PR #2748 missed one detail - extended definitions.

### DIFF
--- a/shared/modules/parse_oval.py
+++ b/shared/modules/parse_oval.py
@@ -25,16 +25,18 @@ class ElementFinder(object):
     def __init__(self, oval_groups):
         self.oval_groups = oval_groups
         self.target = None
+        self.attrib = None
         self.result = set()
 
-    def find_element(self, start_element, target_element_name):
+    def find_element(self, start_element, target_element_name, sought_attrib):
         self.target = target_element_name
+        self.attrib = sought_attrib
         self.result = set()
         self._recurse(start_element)
 
     def _recurse(self, element):
         if element.tag.endswith(self.target):
-            self.result.add(element.attrib["id"])
+            self.result.add(element.attrib[self.attrib])
             return
 
         self._examine_element(element)
@@ -83,7 +85,13 @@ def search_element_for_reference_attributes(element):
 
 def resolve_definition(oval_groups, defn):
     finder = ElementFinder(oval_groups)
-    finder.find_element(defn, "external_variable")
+    finder.find_element(defn, "external_variable", "id")
+    return finder.result
+
+
+def find_extending_defs(oval_groups, defn):
+    finder = ElementFinder(oval_groups)
+    finder.find_element(defn, "extend_definition", "definition_ref")
     return finder.result
 
 
--- a/shared/utils/relabel-ids.py
+++ b/shared/utils/relabel-ids.py
@@ -2,10 +2,10 @@
 
 from __future__ import print_function
 
+import argparse
 import sys
 import os
 
-import argparse
 
 # Put shared python modules in path
 sys.path.insert(0, os.path.join(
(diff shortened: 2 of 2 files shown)


shared/modules/parse_oval.py
```diff
@@ -25,16 +25,18 @@ class ElementFinder(object):
     def __init__(self, oval_groups):
         self.oval_groups = oval_groups
         self.target = None
+        self.attrib = None
         self.result = set()
 
-    def find_element(self, start_element, target_element_name):
+    def find_element(self, start_element, target_element_name, seeked_attrib):
```
mpreisler: isn't this supposed to be "sought" or something like that? I dunno but seeked looks weird.

shared/utils/relabel-ids.py
```diff
@@ -201,15 +202,36 @@ def _add_cce_id_refs_to_oval_checks(self, idmappingdict):
                 msg = "Failed to add CCE ID to {0}.".format(ovalid)
                 raise ssgcommon.SSGError(msg)
 
+    def get_nested_definitions(self, oval_def_id):
+        processed_def_ids = set([])
```
mpreisler: isn't just `set()` the same thing?

shared/utils/relabel-ids.py
```diff
@@ -2,6 +2,7 @@
 
 from __future__ import print_function
 
+import copy
```
mpreisler: this doesn't seem to be used

shared/modules/parse_oval.py
dahaic: Yeah,  you are probably right. I do not connect technical "seek" with the proper English word. Fixed ;)

shared/utils/relabel-ids.py
dahaic: Huh, it never occurred to me :D

shared/utils/relabel-ids.py
dahaic: Good catch, fixed :)

shared/utils/relabel-ids.py
```diff
@@ -201,15 +202,36 @@ def _add_cce_id_refs_to_oval_checks(self, idmappingdict):
...
+                definition = queue.pop()
+            except KeyError:
+                break
+            processed_def_ids.add(definition)
+            definition_tree = self.oval_groups["definitions"].get(definition)
+            extensions = parse_oval.find_extending_defs(self.oval_groups,
+                                                        definition_tree)
+            if not extensions:
+                continue
+            queue |= extensions - processed_def_ids
```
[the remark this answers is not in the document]
dahaic:
```
queue = queue.union(extensions.difference(processed_def_ids))
```
I would rather not use this ^^, it takes me much longer to realize what's going on.

shared/utils/relabel-ids.py
```diff
@@ -201,15 +201,36 @@ def _add_cce_id_refs_to_oval_checks(self, idmappingdict):
                 msg = "Failed to add CCE ID to {0}.".format(ovalid)
                 raise ssgcommon.SSGError(msg)
 
+    def get_nested_definitions(self, oval_def_id):
+        processed_def_ids = set()
+        queue = set([oval_def_id])
+        while True:
```
matejak: You can have `while queue:` here, and get rid of the try/except when popping.

shared/utils/relabel-ids.py
```diff
@@ -201,15 +201,36 @@ def _add_cce_id_refs_to_oval_checks(self, idmappingdict):
...
+            try:
+                definition = queue.pop()
```
matejak: I would call that `definition_id`, or `def_id`, as you have the former on other places.

shared/utils/relabel-ids.py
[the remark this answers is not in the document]
matejak: A compromise may be `queue.update(extensions - processed_def_ids)`.

shared/utils/relabel-ids.py
dahaic: Thanks, looks much better.

shared/utils/relabel-ids.py
dahaic: Good point

shared/utils/relabel-ids.py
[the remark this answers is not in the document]
dahaic: It's just that with my line, I know immediately it's `set` operation. `update` is a bit vague. If you both think that either of your ways is more readable for general public, I will change it. But it's going to be a downgrade for me personally.

matejak:
Great contribution!

PR #2728, opened 2018-04-09
RHEL 7 STIG V1R4

jamescassell:
#### Description:

- added DISA RHEL 7 STIG v1r3 and v1r4 to history and shared/references
- updated stig_overlay.xml (manually, with help from the script
- added/updated xccdf rules
- added/updated oval
- rewrite of #2721

#### Rationale:

- STIG out of date

- Fixes #1891, #2721

PR #2773, opened 2018-04-19
Add create_module and finit_module scripts

redhatrises:
Remediation scripts for #2728


yuumasato:
They look good, Ack.
They are very similar to other remediations, like `  audit_rules_kernel_module_loading_init` and `  audit_rules_kernel_module_loading_delete`. We should consider templating them if other modules are added in future.

PR #2775, opened 2018-04-19
Require the python yaml module, fatal error if it's not found

mpreisler:
`yaml` is a critical requirement since our move to yaml. It is not included in python's standard distribution. It only makes sense to look for it at configure time and tell the user at that time before the build even starts. I plan to use the same mechanism for `jinja2` module.


mpreisler:
> I think it should somehow fail before it tries to scan for dependencies.

Good catch, there was a mistake in the REQUIRE variable it was setting. It's fixed now.

PR #2781, opened 2018-04-20
Unify sshd disable empty passwords

mpreisler:
We had 3 versions of a bash script and they were all doing pretty much the same things.

This is somewhat related to the jinja2 work but doesn't depend on it.


redhatrises:
Ack

PR #2783, opened 2018-04-20
jinja2 for fixes, checks and the opencontrol yaml

mpreisler:
As SSG grows bigger we feel the need for a more advanced and unified templating system. I am hoping jinja2 will become that. It's standard, has great docs and a lot of examples online.

This PR processes every check, fix and opencontrol yaml with jinja2. It provides 2 examples, one for OVAL checks and one for opencontrol yaml. So far we have no fix using the jinja2 capabilities but I am hoping that will change in time once this PR gets merged.


mpreisler:
I was trying to find a good example to showcase the capabilities of this feature. Found some issues in SSG instead: #2786 and #2787

So far I can't find any good example for jinja2 in fixes. There definitely are cases where jinja2 would be great for OVALs but with fixes you can always use the language to do conditional execution - bash or ansible.

mpreisler:
(the jenkins failures are expected because jinja2 package is missing on our jenkins slaves, please don't let that keep you from looking at this PR :-))

mpreisler:
I will proceed to implement this for checks as well because that's where it will help more at the moment. I am sure we can use it later for fixes as well.

mpreisler:
> @mpreisler I assume that {% if product == 'rhel7' %} would be how the do product handling?

yes but I am hoping this will be rare. something like {{% if rpm_based %}} should be more common.

mpreisler:
@openscap-ci test this please

mpreisler:
TODO:
- provide an example of an OVAL check using jinja2
- provide an example of a bash fix using jinja2
- provide an example of opencontrol yaml using jinja2 - remote OVAL is a good candidate

Then I think it can be merged. Error reporting really sucks right now but that can be improved later.

mpreisler:
I have made sure jenkins workers have python-jinja2 installed. See https://github.com/OpenSCAP/jenkins/commit/f452139d3eadf058e44823c4a63b27c498bb053e

shared/guide/system/software/updating/security_patches_up_to_date.rule
```diff
@@ -1,15 +1,23 @@
 documentation_complete: true
 
-prodtype: rhel7
+prodtype: rhel6,rhel7,ubuntu1404,ubuntu1604,ol7
 
 title: 'Ensure Software Patches Installed'
 
 description: |-
+{{% if product in ["rhel6", "rhel7"] %}}
```
yuumasato: Syntax here looks good to me. I like that we can use this kind of construct.

shared/templates/template_OVAL_package_installed
```diff
@@ -4,15 +4,25 @@
...
+{{% if pkg_system == "rpm" %}}
+  <linux:rpminfo_test check="all" check_existence="all_exist"
+  id="test_package_%PKGNAME%_installed" version="1"
+  comment="package %PKGNAME% is installed">
+    <linux:object object_ref="obj_package_%PKGNAME%_installed" />
+  </linux:rpminfo_test>
+  <linux:rpminfo_object id="obj_package_%PKGNAME%_installed" version="1">
+    <linux:name>%PKGNAME%</linux:name>
+  </linux:rpminfo_object>
+{{% elif pkg_system == "dpkg" %}}
```
yuumasato: Shouldn't this be `deb`?

mpreisler: no. see https://en.wikipedia.org/wiki/Dpkg

deb is short for debian, dpkg is the packaging system

yuumasato: Ok.

shared/templates/template_OVAL_package_installed
```diff
@@ -4,15 +4,25 @@
     <metadata>
       <title>Package %PKGNAME% Installed</title>
       <affected family="unix">
-        <platform>Debian 8</platform>
+        <platform>multi_platform_all</platform>
       </affected>
-      <description>The DEB package %PKGNAME% should be installed.</description>
+      <description>The {{{ pkg_system|upper }}} package %PKGNAME% should be installed.</description>
```
yuumasato: But then here it be expanded to `DPKG package`, not `DEB package`.

yuumasato: It may actually be `DPKG package`, but I think I never saw like this.

mpreisler: is that a problem? if anything DPKG package is more correct because DEB is short for debian so it doesn't apply on Ubuntu.

shared/utils/combine-ovals.py
```diff
@@ -278,18 +278,21 @@ def checks(product, oval_version, oval_dirs):
...
-                        body.append(xml_content)
-                        included_checks_count += 1
-                        already_loaded[filename] = oval_version
+                    xml_content = ssgcommon.process_file_with_jinja(
+                        os.path.join(oval_dir, filename), product_yaml
+                    )
+                    if not check_is_applicable_for_product(
+                        xml_content,
+                        ssgcommon.required_yaml_key(product_yaml, "product")
+                    ):
```
yuumasato: Do you see jinja being used to determine applicability? Otherwise we can check for applicability before processing.

mpreisler: It's not used for applicability right now but I am hoping it will replace the applicability checking in the future. We have to process first because the other order breaks parsing error functionality. It's doable but not without loading the file twice.

yuumasato: Thanks for clarification.

mpreisler:
I have removed the weird relpath hack and implemented an absolute file path loader for jinja2. Submitted it as a PR, see https://github.com/pallets/jinja/pull/846

mpreisler:
I don't want to do that as part of this PR but in another PR I want to start using jinja2 macros to replace some of the templating and shared bash functionality. See http://davedash.com/tutorial/jinja-macros/

We could replace all package install and remove with jinja2 code that always puts in the correct minimal snippet into fixes.

PR #2784, opened 2018-04-20
[DO NOT MERGE] Horrible hack fix for audit_rules_privileged_commands OVAL

redhatrises:

### DIFF
--- a/shared/checks/oval/audit_rules_privileged_commands.xml
+++ b/shared/checks/oval/audit_rules_privileged_commands.xml
@@ -31,20 +31,13 @@
   <!-- First define OVAL entities that can be reused across tests below -->
   <unix:file_object id="object_system_privileged_commands" comment="system files with setuid or setgid permission set" version="1">
     <unix:behaviors recurse="directories" recurse_direction="down" recurse_file_system="local" max_depth="-1" />
-    <unix:path operation="equals">/</unix:path>
-    <!-- [a-z]+ regex below is a workaround for OpenSCAP https://fedorahosted.org/openscap/ticket/457 bug -->
-    <unix:filename operation="pattern match">[a-z]+</unix:filename>
+    <unix:filepath operation="pattern match">^\/(?!proc|sbin|lib|lib64|dev|sys|run|var\/lib\/nfs|var\/run\/user\/1000\/gvfs).*\/[a-z]+</unix:filepath>
     <filter action="include">state_setuid_or_setgid_set</filter>
-    <filter action="exclude">state_dev_proc_sys_dirs</filter>
   </unix:file_object>
 
   <unix:file_state id="state_setuid_or_setgid_set" version="1" operator="OR">
     <unix:suid datatype="boolean">true</unix:suid>
     <unix:sgid datatype="boolean">true</unix:sgid>
-  </unix:file_state>
-
-  <unix:file_state id="state_dev_proc_sys_dirs" version="1">
-    <unix:filepath operation="pattern match">^\/(dev|proc|sys)\/.*$</unix:filepath>
   </unix:file_state>
 
   <local_variable id="variable_all_privileged_commands" comment="All privileged commands" datatype="string" version="1">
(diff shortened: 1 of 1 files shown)


shared/checks/oval/audit_rules_privileged_commands.xml
```diff
@@ -31,22 +31,15 @@
   <!-- First define OVAL entities that can be reused across tests below -->
   <unix:file_object id="object_system_privileged_commands" comment="system files with setuid or setgid permission set" version="1">
     <unix:behaviors recurse="directories" recurse_direction="down" recurse_file_system="local" max_depth="-1" />
-    <unix:path operation="equals">/</unix:path>
-    <!-- [a-z]+ regex below is a workaround for OpenSCAP https://fedorahosted.org/openscap/ticket/457 bug -->
-    <unix:filename operation="pattern match">[a-z]+</unix:filename>
+    <unix:filepath operation="pattern match">^\/(?!proc|sbin|lib|lib64|dev|sys|run|var\/lib\/nfs|var\/run\/user\/1000\/gvfs).*\/[a-z]+</unix:filepath>
```
redhatrises: Have to exclude the search of `/dev, /proc, /sys` here for some reason as the exclude wasn't working which was helping to cause this to fail.

shared/checks/oval/audit_rules_privileged_commands.xml
```diff
@@ -31,22 +31,15 @@
...
     <filter action="include">state_setuid_or_setgid_set</filter>
-    <filter action="exclude">state_dev_proc_sys_dirs</filter>
```
redhatrises: This excludes filter doesn't appear to be working. Any excludes filter for that matter.

shared/checks/oval/audit_rules_privileged_commands.xml
```diff
@@ -31,22 +31,15 @@
...
   </unix:file_object>
 
   <unix:file_state id="state_setuid_or_setgid_set" version="1" operator="OR">
     <unix:suid datatype="boolean">true</unix:suid>
     <unix:sgid datatype="boolean">true</unix:sgid>
   </unix:file_state>
 
-  <unix:file_state id="state_dev_proc_sys_dirs" version="1">
-    <unix:filepath operation="pattern match">^\/(dev|proc|sys)\/.*$</unix:filepath>
```
redhatrises: `dev`, `proc`, and `sys` are still included even though they are filtered out?

shared/checks/oval/audit_rules_privileged_commands.xml
yuumasato: This is very likely a bug in openscap, see https://www.redhat.com/archives/open-scap-list/2018-April/msg00001.html

matejak:
Added the DO NOT MERGE label, `openscap` should be investigated first as it is the primary source of the problem.

jan-cerny:
The problem with OpenSCAP should be reported to OpenSCAP as an upstream issue. @yuumasato Please file an issue against OpenSCAP.

A good idea is to write a test in OpenSCAP upstream test suite that would trigger the fail. @matejak Could you please write such a test?

shared/checks/oval/audit_rules_privileged_commands.xml
```diff
@@ -75,7 +68,7 @@
   </ind:textfilecontent54_test>
   <ind:textfilecontent54_object id="object_arpc_suid_sgid_augenrules" version="1">
     <ind:filepath operation="pattern match">^/etc/audit/rules\.d/.*\.rules</ind:filepath>
-    <ind:pattern operation="pattern match">^[\s]*-a always,exit (?:-F path=([\S]+) )+-F perm=[r|w]?x -F auid>=1000 -F auid!=4294967295[\s]+(-k[\s]+|-F[\s]+key=)[\S]+[\s]*$</ind:pattern>
```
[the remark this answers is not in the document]
yuumasato: This was fixed in https://github.com/OpenSCAP/scap-security-guide/pull/3134/commits/6e78390ff721f1e38eef453a74d3a0280f099384

jan-cerny:
It looks like there is no intention to merge this PR and therefore we lean towards closing it.
We don't have any reproducer so far.
@redhatrises Do you have any reproducer to trigger the bug that you're trying to fix?

redhatrises:
@jan-cerny let me check if the latest changes have fixed this issue.

redhatrises:
@jan-cerny this is still not fixed! Still results in an error. 

[audit_privileged_commands.tar.gz](https://github.com/OpenSCAP/scap-security-guide/files/2328722/audit_privileged_commands.tar.gz)

We need to fix this asap as this has been a problem for a while.

yuumasato:
@redhatrises How did you run it?

When I do this, I see result false, not error:
```
$ oscap oval eval --verbose=DEVEL --verbose-log-file privileged_commands.log --results results.xml audit_rules_privileged_commands21dfdF.xml 
Definition oval:scap-security-guide.testing-audit_rules_privileged_commands:def:1: false
Definition oval:scap-security-guide.testing-audit_rules_augenrules:def:1: true
Definition oval:scap-security-guide.testing-audit_rules_auditctl:def:1: false
Evaluation done.
```

Also, I see that  
`oval:scap-security-guide.testing-test_arpc_bin_count_equals_rules_count_augenrules:tst:1` results in error, and I think this should cause whole OVAL test to result in `error`, but I see `fail` there.

And the reason it is result in error, is that there is no rule setup for any privileged command. At some point the definition tries to count the number of rules that have `path` to a privileged commands. And I think is trying to count number of things which don't exist, and unfortunately I think it doesn't result in zero, it result in `no values`. And comparing `41` with `no values` is an error.

Could you try setting up at least one rule in any of the audit rules file checked, matching any privileged command?

redhatrises:
@yuumasato it's actually passing with the latest master... sooooo.... nevermind. I tried it also off of the latest release, and it was failing but nevermind.

PR #2786, opened 2018-04-23
Package command apt get

mpreisler:
#### Description:

Split package_command into package_install and package_remove. Added apt-get support. Simplified a bunch of templates.

There is more work to be done, ubuntus are basically copies of each other and they each have a ton of non templated package install/remove fixes.

This is somewhat related to the jinja2 work but doesn't depend on it.


shared/bash_remediation_functions/package_install.sh
```diff
@@ -0,0 +1,36 @@
...
+  echo "Aborting."
+  exit 1
+fi
+
+if which dnf ; then
+  if ! rpm -q --quiet "$package"; then
+    dnf install -y "$package"
+  fi
+elif which yum ; then
+  if ! rpm -q --quiet "$package"; then
```
jan-cerny: I think it is possible to have `yum` on Debian.
https://packages.debian.org/jessie/yum
Will this work in that case?

mpreisler: it should still work

shawndwells: YUM would be on debian, but not the RPM command.

yuumasato: `rpm` is listed as dependency of `yum`, so both will be there even on debian.

yuumasato:
Generated content looks good, Ack!

PR #2788, opened 2018-04-23
Fix CCE URI

redhatrises:
#### Description:

- Fix CCE URI

#### Rationale:

- URI was pointing to an outdated page.


shared/utils/relabel-ids.py
```diff
@@ -187,19 +187,23 @@ def _add_cce_id_refs_to_oval_checks(self, idmappingdict):
...
+            if ssgcommon.cce_is_valid(xccdfcceid):
+                # Then append the <reference source="CCE" ref_id="CCE-ID" /> element right
+                # after <description> element of specific OVAL check
+                ccerefelem = ssgcommon.ElementTree.Element(
+                    'reference', ref_id=xccdfcceid, source="CCE")
+                metadata = rule.find(".//{%s}metadata" % self.CHECK_NAMESPACE)
+                metadata.append(ccerefelem)
+                # Sanity check if appending succeeded
+                for ref in metadata.findall(".//reference"):
+                    if ref.attrib.get("ref_id") == xccdfcceid:
```
mpreisler: you can do this in a pythonic way on just one line:
`cce_ref = ref if ref.attrib.get("ref_id" == xccdfcceid else None`

shared/utils/relabel-ids.py
```diff
@@ -429,13 +433,10 @@ def verify_correct_form_of_referenced_cce_identifiers(xccdftree):
...
-            continue
-
-        cceid = identcce.text
-        if not ssgcommon.cce_is_valid(cceid):
-            msg = "The CCE ID {0} is not in valid form.".format(cceid)
-            raise ssgcommon.SSGError(msg)
+        if identcce is not None:
+            cceid = identcce.text
+            if not ssgcommon.cce_is_valid(cceid):
+                rule.remove(identcce)
```
mpreisler: I strongly prefer when we fail build than silently remove. This PR changes behavior in a way that invalid CCEs will be ignored and only spotted by users and customers.

shared/utils/relabel-ids.py
redhatrises: @mpreisler always forget the more pythonic way.... guess it shows. :grin:

shared/utils/relabel-ids.py
redhatrises: Actually, it removes the idents with invalid CCEs. Failing on invalid CCEs is not a good way to go IMO as we would also need do to for all references to be consistent which means that we would be getting bogged down in build failures which we are actually getting right now with invalid CCEs. The code just never worked to show that in the first place.

redhatrises: @mpreisler ^

shared/utils/relabel-ids.py
```diff
@@ -429,13 +430,11 @@ def verify_correct_form_of_referenced_cce_identifiers(xccdftree):
...
-            continue
-
-        cceid = identcce.text
-        if not ssgcommon.cce_is_valid(cceid):
-            msg = "The CCE ID {0} is not in valid form.".format(cceid)
-            raise ssgcommon.SSGError(msg)
+        if identcce is not None:
+            cceid = identcce.text
+            if not ssgcommon.cce_is_valid(cceid):
+                print("Warning: CCE '{0}' is invalid for rule '{1}'. Removing CCE...".format(cceid, rule.get("id")))
```
mpreisler: write this to stderr

redhatrises: @mpreisler done

shared/utils/relabel-ids.py
```diff
@@ -187,19 +187,20 @@ def _add_cce_id_refs_to_oval_checks(self, idmappingdict):
...
+            if ssgcommon.cce_is_valid(xccdfcceid):
+                # Then append the <reference source="CCE" ref_id="CCE-ID" /> element right
+                # after <description> element of specific OVAL check
+                ccerefelem = ssgcommon.ElementTree.Element(
+                    'reference', ref_id=xccdfcceid, source="CCE")
+                metadata = rule.find(".//{%s}metadata" % self.CHECK_NAMESPACE)
+                metadata.append(ccerefelem)
+                # Sanity check if appending succeeded
+                for ref in metadata.findall(".//reference"):
+                    cce_ref = ref if ref.attrib.get("ref_id") == xccdfcceid else None
```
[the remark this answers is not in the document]
redhatrises: @mpreisler `cce-ref` is now set to nothing before the `for` loop.

shared/utils/relabel-ids.py
```diff
@@ -187,19 +187,21 @@ def _add_cce_id_refs_to_oval_checks(self, idmappingdict):
...
-                raise ssgcommon.SSGError(msg)
+            if ssgcommon.cce_is_valid(xccdfcceid):
+                # Then append the <reference source="CCE" ref_id="CCE-ID" /> element right
+                # after <description> element of specific OVAL check
+                ccerefelem = ssgcommon.ElementTree.Element(
+                    'reference', ref_id=xccdfcceid, source="CCE")
+                metadata = rule.find(".//{%s}metadata" % self.CHECK_NAMESPACE)
+                metadata.append(ccerefelem)
+                # Sanity check if appending succeeded
+                cce_ref = ""
```
mpreisler: you reset cce_ref to empty string but then set it to None in the loop in case it's not defined. choose just one

redhatrises: @mpreisler a little confused here. In a previous comment, you say that it is dangerous that cce_ref is never reset, so what should be done here?

mpreisler: either reset cce_ref to None or set it to "" in the code below. Don't use None and "" interchangeably, they are different. In this case it's fine and doesn't do anything bad but it's confusing to read.

redhatrises: Thanks @mpreisler that makes sense. Set to None.

shared/utils/relabel-ids.py
```diff
@@ -187,19 +187,21 @@ def _add_cce_id_refs_to_oval_checks(self, idmappingdict):
...
+                # Then append the <reference source="CCE" ref_id="CCE-ID" /> element right
+                # after <description> element of specific OVAL check
+                ccerefelem = ssgcommon.ElementTree.Element(
+                    'reference', ref_id=xccdfcceid, source="CCE")
+                metadata = rule.find(".//{%s}metadata" % self.CHECK_NAMESPACE)
+                metadata.append(ccerefelem)
+                # Sanity check if appending succeeded
+                cce_ref = None
+                for ref in metadata.findall(".//reference"):
+                    cce_ref = ref if ref.attrib.get("ref_id") == xccdfcceid else None
```
mpreisler: I am sorry for endlessly nitpicking around this loop but IMO the logic is still flawed.

take this case:
first reference is non-cce
second reference is cce
third reference is non cce

the final cce_ref will be None, this certainly is not what you are looking for.

you need to set cce_ref **if and only if** the ref_id is cceid

mpreisler: actually I think this whole block is quite ridiculous, we are checking that elementtree did what it was supposed to do...

redhatrises: @mpreisler I agree with that statement as well. I thought it was really odd. Removed the sanity checking block as it seems.... unneeded.

mpreisler:
ACK, sorry for delays

PR #2789, opened 2018-04-23
Add external content to yaml

redhatrises:
#### Description:

- Adds capability to get external SCAP content when specified in YAML

#### Rationale:

- Regression from YAML conversion as capability existed prior.

### DIFF
--- a/shared/guide/system/software/updating/security_patches_up_to_date.rule
+++ b/shared/guide/system/software/updating/security_patches_up_to_date.rule
@@ -36,6 +36,8 @@ references:
     srg: SRG-OS-000480-GPOS-00227
     stigid@rhel7: "020260"
 
+oval_external_content: https://www.redhat.com/security/data/oval/com.redhat.rhsa-RHEL7.xml.bz2
+
 ocil_clause: 'updates are not installed'
 
 ocil: |-
--- a/shared/utils/yaml-to-shorthand.py
+++ b/shared/utils/yaml-to-shorthand.py
@@ -430,8 +430,7 @@ class Rule(object):
             check = ET.SubElement(rule, 'check')
             check.set("system", "http://oval.mitre.org/XMLSchema/oval-definitions-5")
             external_content = ET.SubElement(check, "check-content-ref")
-            if self.external_oval:
-                external_content.set("href", self.external_oval)
+            external_content.set("href", self.external_oval)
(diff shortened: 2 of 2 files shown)


shared/guide/system/software/updating/security_patches_up_to_date.rule
```diff
@@ -36,6 +36,8 @@ references:
     srg: SRG-OS-000480-GPOS-00227
     stigid@rhel7: "020260"
 
+oval_external_content: https://www.redhat.com/security/data/oval/com.redhat.rhsa-RHEL7.xml.bz2
```
shawndwells: This seems Red Hat specific. What happens if/when someone adds content for other communities, e.g. the Ubuntu OVAL feed @ https://people.canonical.com/~ubuntu-security/oval/

redhatrises: @shawndwells true. This is being worked as a separate effort with jinja templating.

shared/utils/yaml-to-shorthand.py
```diff
@@ -425,11 +426,18 @@ def to_xml_element(self):
...
-        # TODO: This is pretty much a hack, oval ID will be the same as rule ID
-        #       and we don't want the developers to have to keep them in sync.
-        #       Therefore let's just add an OVAL ref of that ID.
-        oval_ref = ET.SubElement(rule, "oval")
-        oval_ref.set("id", self.id_)
+        if self.external_oval:
+            check = ET.SubElement(rule, 'check')
+            check.set("system", "http://oval.mitre.org/XMLSchema/oval-definitions-5")
+            external_content = ET.SubElement(check, "check-content-ref")
+            if self.external_oval:
```
jan-cerny: this condition is true here, it is the same condition as line 429

shared/guide/system/software/updating/security_patches_up_to_date.rule
mpreisler: This is a perfect case for a jinja example. I will get that working in #2783 once this is merged.

shared/utils/yaml-to-shorthand.py
redhatrises: hahahaha thanks @jan-cerny! fixed.

PR #2790, opened 2018-04-24
Don't exit with 0 when product.yml loading fails

mpreisler:
Previously this was done to slowly port everything over to yaml, now
that it is ported we want to fail hard in case product.yml loading
fails.

Oh well..


redhatrises:
Ack

PR #2791, opened 2018-04-24
Use new dconf bash functions for bash scripts and add some missing dconf scripts

redhatrises:
#### Description:

- Use new dconf bash functions for bash scripts and add some missing dconf scripts


shared/fixes/bash/dconf_gnome_banner_enabled.sh
```diff
@@ -0,0 +1,5 @@
+# platform = Red Hat Enterprise Linux 7
+. /usr/share/scap-security-guide/remediation_functions
+
+dconf_settings 'org/gnome/login-screen' 'banner-message-enable' "true" 'gdm.d' '00-security-settings'
```
shawndwells: double vs single quote on ``"true"``

redhatrises: @shawndwells nitpicker ;)

Fixed your nitpicking.

shawndwells:
just waiting on jenkins

shared/fixes/bash/dconf_gnome_banner_enabled.sh
shawndwells: "nickpicking'

redhatrises: hahaha

PR #2792, opened 2018-04-24
Make sure that dconf dirs exist

redhatrises:
#### Description:

- Make sure that dconf dirs exist

#### Rationale:

- dconf directories may not exist and could cause a remediation failure.

### DIFF
--- a/shared/bash_remediation_functions/dconf_lock.sh
+++ b/shared/bash_remediation_functions/dconf_lock.sh
@@ -19,10 +19,7 @@ function dconf_lock {
 	LOCKFILES=$(grep -r "^/${_key}/${_setting}$" "/etc/dconf/db/" | grep -v "distro\|ibus" | cut -d":" -f1)
 	LOCKSFOLDER="/etc/dconf/db/${_db}/locks"
 
-	if [ ! -d ${LOCKSFOLDER} ]
-	then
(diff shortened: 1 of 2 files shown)


shared/bash_remediation_functions/dconf_lock.sh
```diff
@@ -17,6 +17,12 @@ function dconf_lock {
 
 	# Check for setting in any of the DConf db directories
 	LOCKFILES=$(grep -r "^/${_key}/${_setting}$" "/etc/dconf/db/" | grep -v "distro\|ibus" | cut -d":" -f1)
+	LOCKSFOLDER="/etc/dconf/db/${_db}/locks"
+
+	if [ ! -d ${LOCKSFOLDER} ]
```
mpreisler: There is no need to check, just run `mkdir -p`, it will succeed even if the dir exists.

shared/bash_remediation_functions/dconf_lock.sh
```diff
@@ -17,6 +17,12 @@ function dconf_lock {
...
+	then
+		mkdir -p ${LOCKSFOLDER}
```
mpreisler: use quotes properly here in case LOCKSFOLDER has spaces.

redhatrises:
@mpreisler changes made

PR #2794, opened 2018-04-24
Fix sudo audit rules

### LINKED ISSUE #2507: audit_rules_privileged_commands_su* scripts not compatible with each other
#### Description of problem:
The following 3 bash fix scripts are not compatible with each other and must be run in a certain order or they will overwrite each other's fixes:
1. audit_rules_privileged_commands_su
2. audit_rules_privileged_commands_sudo
3. audit_rules_privileged_commands_sudoedit

The issue is with the following lines:
`PATTERN="-a always,exit -F path=/usr/bin/su.*"`
`PATTERN="-a always,exit -F path=/usr/bin/sudo.*"`
`PATTERN="-a always,exit -F path=/usr/bin/sudoedit.*"`

If you run the scripts in the order (su, sudo, sudoedit) they will likely work, however running them in the order (sudoedit, sudo, su) will result in the sudo script overwriting the audit rule for sudoedit, and the su rule overwriting both sudo and sudoedit if either of them exist.

0.1.36-1.el7.centos

#### Steps to Reproduce:

1. Run the individual fixes in the order : sudoedit, sudo, su

#### Actual Results:
The sudo fix overwrites the sudoedit rules, and then the su fix overwrites the sudo rules

#### Expected Results:
All 3 rules are successfully applied

This should be pretty easy to fix by simply adding `[[:space:]]` or some word boundary delimiter after the executable and before the `.*`

redhatrises:
#### Description:

- Handle space after word in template_BASH_audit_rules_privileged_commands

#### Rationale:

- Fix was overwriting expanded words e.g. sudo with sudoedit in audit rules 
- Fixes #2507

### DIFF
--- a/shared/templates/template_BASH_audit_rules_privileged_commands
+++ b/shared/templates/template_BASH_audit_rules_privileged_commands
@@ -3,7 +3,7 @@
 # Include source function library.
 . /usr/share/scap-security-guide/remediation_functions
 
-PATTERN="-a always,exit -F path=%PATH%.*"
+PATTERN="-a always,exit -F path=%PATH%\\s*.*"
 GROUP="privileged"
 FULL_RULE="-a always,exit -F path=%PATH% -F perm=x -F auid>=1000 -F auid!=4294967295 -F key=privileged"
 # Perform the remediation for both possible tools: 'auditctl' and 'augenrules'


shared/templates/template_BASH_audit_rules_privileged_commands
```diff
@@ -3,7 +3,7 @@
 # Include source function library.
 . /usr/share/scap-security-guide/remediation_functions
 
-PATTERN="-a always,exit -F path=%PATH%.*"
+PATTERN="-a always,exit -F path=%PATH%\\s*.*"
```
matejak: Is there a special reason why to use double backslash? I think that it is correct, but if you replace double quotes with single quotes, you can use single backslash.

redhatrises: @matejak the regex wasn't properly handled without the double backslash. 

Should we replace the double quotes with single quotes? Not sure if we would come across other problems that we would then have to figure out.

matejak: I have reviewed the called bash function and it is full of bugs. Fixing that one is out of scope of this PR, so I merge it as it works in this current form, and we can fix everything later, during a refactoring phase.
However, by looking at the function code, I am not able to explain why single backslash doesn't work and double does.

matejak:
Thanks @redhatrises for finding the way how to fix the issue!

redhatrises:
Thanks for reviewing @matejak!

PR #2795, opened 2018-04-25
Add Missing DISA RHEL7 STIG XCCDF rules

### LINKED ISSUE #2637: Missing RHEL7 STIG XCCDF Rules
#### Description of problem:
The DISA RHEL7 STIG was updated on Jan 26 2018.  There were new rules added that have not been added to the SSG repository yet.

Current repository snapshot

RHEL 7.4

Missing Rules:
RHEL-07-010062
```
Verify the operating system prevents a user from overriding the screensaver lock-enabled setting for the graphical user interface. 

Note: If the system does not have GNOME installed, this requirement is Not Applicable. The screen program must be installed to lock sessions on the console.

Determine which profile the system database is using with the following command:
# grep system-db /etc/dconf/profile/user

system-db:local

Check for the lock-enabled setting with the following command:

Note: The example below is using the database "local" for the system, so the path is "/etc/dconf/db/local.d". This path must be modified if a database other than "local" is being used.

# grep -i lock-enabled /etc/dconf/db/local.d/locks/*

/org/gnome/desktop/screensaver/lock-enabled

If the command does not return a result, this is a finding.
```

RHEL-07-010101
```
Verify the operating system prevents a user from overriding the screensaver idle-activation-enabled setting for the graphical user interface. 

Note: If the system does not have GNOME installed, this requirement is Not Applicable. The screen program must be installed to lock sessions on the console.

Determine which profile the system database is using with the following command:
# grep system-db /etc/dconf/profile/user

system-db:local

Check for the idle-activation-enabled setting with the following command:

Note: The example below is using the database "local" for the system, so the path is "/etc/dconf/db/local.d". This path must be modified if a database other than "local" is being used.

# grep -i idle-activation-enabled /etc/dconf/db/local.d/locks/*

/org/gnome/desktop/screensaver/idle-activation-enabled

If the command does not return a result, this is a finding.
```

RHEL-07-030819
```
Verify the operating system generates audit records when successful/unsuccessful attempts to use the "create_module" command occur. 

Check the auditing rules in "/etc/audit/audit.rules" with the following command:

Note: The output lines of the command are duplicated to cover both 32-bit and 64-bit architectures. Only the line appropriate for the system architecture must be present.

# grep -iw create_module /etc/audit/audit.rules

If the command does not return the following output (appropriate to the architecture), this is a finding. 

-a always,exit -F arch=b32 -S create_module -k module-change

-a always,exit -F arch=b64 -S create_module -k module-change

If there are no audit rules defined for “create_module”, this is a finding.
```

RHEL-07-030821
```
Verify the operating system generates audit records when successful/unsuccessful attempts to use the "finit_module" command occur. 

Check the auditing rules in "/etc/audit/audit.rules" with the following command:

Note: The output lines of the command are duplicated to cover both 32-bit and 64-bit architectures. Only the line appropriate for the system architecture must be present.

# grep -iw finit_module /etc/audit/audit.rules

If the command does not return the following output (appropriate to the architecture), this is a finding. 

-a always,exit -F arch=b32 -S finit_module -k module-change

-a always,exit -F arch=b64 -S finit_module -k module-change

If there are no audit rules defined for “finit_module”, this is a finding.
```

redhatrises:
- Fixes #2637


mpreisler:
ACK

PR #2796, opened 2018-04-25
Fix missing elements and description in var_auditd_admin_space_left_action and var_auditd_space_left_action

### LINKED ISSUE #2700: auditd_data_retention_space_left_action rule description missing elements
#### Description of problem:
It miss an element in the description of the xccdf_org.ssgproject.content_rule_auditd_data_retention_space_left_action rule on the file /usr/share/xml/scap/ssg/content/ssg-centos7-xccdf.xml.

The Description of this rule enumerate all the possible ACTION described in the auditd.conf man page.
The list of action given on the description is the following :
* ignore
* syslog
* email
* exec
* suspend
* single
* halt

But, in the auditd.conf man pages, it given one more possible action : rotate.
This is the corresponding portion of the man page 

> admin_space_left_action
              This  parameter  tells  the  system  what  action  to take when the system has
              detected that it is low on disk  space.   Valid  values  are  ignore,  syslog,
              rotate,  email,  exec, suspend, single, and halt.  If set to ignore, the audit
              daemon does nothing.  Syslog means that it will issue  a  warning  to  syslog.
              rotate will rotate logs, losing the oldest to free up space.

Furthermore, I think this value is also an acceptable value and must be add to the OVAL acceptance tests.

0.1.30-3.el70.centos.0.3

CentOS 7.2 x86_34
CentOS 7.3 x86_64

#### Steps to Reproduce:

1. Build the CentOS 7 ssg.

add the line : 
` <li><tt>rotate</tt></li>`
to the 23rd line of the scap-security-guide/shared/guide/system/auditing/configure_auditd_data_retention/ file

mildass:
#### Description:

Add missing element `rotate` to `var_auditd_admin_space_left_action` and `var_auditd_space_left_action`.
Change description in `var_auditd_admin_space_left_action`, which contained `space_left_action` instead of `admin_space_left_action`.

- Fixes #2700

### DIFF
--- a/shared/guide/system/auditing/configure_auditd_data_retention/var_auditd_admin_space_left_action.var
+++ b/shared/guide/system/auditing/configure_auditd_data_retention/var_auditd_admin_space_left_action.var
@@ -2,7 +2,7 @@ documentation_complete: true
 
 title: 'Action for auditd to take when disk space just starts to run low'
 
-description: 'The setting for space_left_action in /etc/audit/auditd.conf'
+description: 'The setting for admin_space_left_action in /etc/audit/auditd.conf'
 
 type: string
 
@@ -18,3 +18,4 @@ options:
     single: single
     suspend: suspend
     syslog: syslog
+    rotate: rotate
--- a/shared/guide/system/auditing/configure_auditd_data_retention/var_auditd_space_left_action.var
+++ b/shared/guide/system/auditing/configure_auditd_data_retention/var_auditd_space_left_action.var
@@ -18,3 +18,4 @@ options:
     single: single
     suspend: suspend
     syslog: syslog
+    rotate: rotate


redhatrises:
The description change is good, but we may be removing `rotate` as `exec` is more acceptable handling.

redhatrises:
The `rotate` option is not in the STIG or CIS benchmarks. Please remove it from this PR.

dahaic:
I don't understand, @redhatrises - `rotate` is only adding option for customization. Not changing any existing profile?

shawndwells:
The associated OVAL will use ``var_auditd_admin_space_left_action`` in the pass/fail evaluation. Adding ``rotate`` will mean the OVAL will pass when ``admin_space_left_action = rotate``, which will go against existing policies like STIG and CIS (as @redhatrises noted).

... but that also means someone tailored the profile away from the pre-defined defaults. If this was done, that's the user acknowledgement of drifting away from accepted values from whatever regulatory baseline the profile started from.

I'm ok with merging this, having thought through it a bit. @redhatrises what do you think?

redhatrises:
@shawndwells you could make that same argument for using `ignore` as an option for customization which clearly is a no-go.

mpreisler:
> @shawndwells you could make that same argument for using ignore as an option for customization which clearly is a no-go.

Adding more options doesn't make things less or more secure. You can still override it to anything via SCAP Workbench or via set-value. I think having "rotate" there makes sense. I wouldn't add "ignore" because that's a bad recommendation but rotate can make sense in some workloads.

dahaic:
As Martin said, we can leave out obviously bad options, but if the option can in some context be used successfully, why not to leave them the convenience of predefined value? Even if the context is "home computer with mild level of hardening".

yuumasato:
Agree with @shawndwells, that user customization can be treated as acknowledgment of drifting away from compliance with a policy.

IMO `rotate` option can be added for convenience of customization, even if no policy selects it.

mpreisler:
I think we have consensus, let's merge this.

PR #2797, opened 2018-04-25
Add regex handling for SRG and STIG reference versions in CMake

redhatrises:
#### Description:

- Add regex handling for SRG and STIG versions in CMake

#### Rationale:

- While SRG and STIG references may not be updated often, this makes it easier to update the reference file. If a specific version is needed, the version can still me manually entered.

### DIFF
--- a/cmake/SSGCommon.cmake
+++ b/cmake/SSGCommon.cmake
@@ -922,8 +922,8 @@ macro(ssg_build_html_cce_table PRODUCT)
         DESTINATION "${SSG_TABLE_INSTALL_DIR}")
 endmacro()
 
-macro(ssg_build_html_srgmap_tables PRODUCT DISA_SRG_TYPE DISA_SRG_VERSION)
-    FILE(GLOB DISA_SRG_REF "${SSG_SHARED_REFS}/disa-${DISA_SRG_TYPE}-srg-${DISA_SRG_VERSION}.xml")
+macro(ssg_build_html_srgmap_tables PRODUCT DISA_SRG_TYPE)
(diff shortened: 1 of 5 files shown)


cmake/SSGCommon.cmake
```diff
@@ -923,27 +923,28 @@ macro(ssg_build_html_cce_table PRODUCT)
 endmacro()
 
 macro(ssg_build_html_srgmap_tables PRODUCT DISA_SRG_TYPE DISA_SRG_VERSION)
+    FILE(GLOB DISA_SRG_REF "${SSG_SHARED_REFS}/disa-${DISA_SRG_TYPE}-srg-${DISA_SRG_VERSION}.xml")
```
mpreisler: cmake style, file should be lowercase
since the regex is the same for all products, shouldn't we just set it here? why set it in the CMakeLists.txt in each product to exactly the same thing?

redhatrises: While it isn't the case today, I was planning for if there was a need in case you have `disa-os-srg-v1r4.xml` and `disa-os-srg-v2r10.xml` where one product had to have a specific version that needed to be tied to it. Or should we just not care about that?

mpreisler: If that is the case couldn't we just remove one of the files and keep only the one that should be tied to the product?

redhatrises: @mpreisler what if the other product(s) require the new versions?

mpreisler: every product has their own reference doc in CMAKE_CURRENT_SOURCE_DIR, right? so they can all use different versions, that's fine. As long as every product is using just one version.

PR #2800, opened 2018-04-26
Use yaml.safe_load for build related yaml files

mpreisler:
There is no real danger of somebody trying to exploit their own machine
but on the other hand we don't use any of the unsafe facilities of
yaml.load so we might as well use safe_load.


jan-cerny:
This sounds very reasonable for me. The SSG build with this patch works OK on my machine.

PR #2802, opened 2018-04-27
Removed package_*_{installed,removed} from shared that are templated

mpreisler:
These files are not needed in the git repo because these checks are
templated.


mpreisler:
I will also template the rest of them to get rid of all bash and oval files in shared that can be templated.

mpreisler:
> I will also template the rest of them to get rid of all bash and oval files in shared that can be templated.

DONE

matejak:
Great cleanup work! However, I have noticed that we don't have any tests for installed package. Although I have verified that this PR works, it would be nice to have a test as part of the SSG test suite. Would you like to add one?

mpreisler:
> Would you like to add one?

not particularly

matejak:
OK, that can be done later.

PR #2805, opened 2018-04-27
openSUSE CPE update and clean-up

abergmann:
#### Description:

- openSUSE CPE entries where updated inside the official NIST CPE dictionary.
- Distribution version 42.2 and 42.3 where added.

#### Rationale:

- Update the openSUSE entries to make then usable with the current openSUSE release.


redhatrises:
@openscap-ci add to whitelist

PR #2812, opened 2018-04-30
remove kernel IPv6 from RHEL6 STIG

shawndwells:
Removes rhel6 IPv6 kernel module check to align with DISA content.

Resolves https://github.com/OpenSCAP/scap-security-guide/issues/2758


shawndwells:
Also removes from RHEL6 USGCB, which was just a carryover from DISA RHEL6 STIG

redhatrises:
Ack

PR #2813, opened 2018-05-01
Removed an accidentally committed file in shared/fixes/bash

mpreisler:
Keep it clean, people :-D


redhatrises:
Huh.... How did that happen? :D

redhatrises:
Ack

PR #2814, opened 2018-05-02
Add benchmark metadata element to shorthand

yuumasato:
#### Description:

- Add benchmark metadata element to shorthand.

#### Rationale:

- Script that generates shorthand from yaml was not adding metadata
element. This element is expanded to XCCDF metadata by shorthand
transformation.
- NIST SCAPVAL tool errors if there is no metadata in the Benchmark.

### DIFF
--- a/shared/utils/yaml-to-shorthand.py
+++ b/shared/utils/yaml-to-shorthand.py
@@ -227,6 +227,7 @@ class Benchmark(object):
                 add_sub_element(root, "platform", cpe)
         version = ET.SubElement(root, 'version')
         version.text = self.version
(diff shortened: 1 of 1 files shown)


yuumasato:
Unfortunately the SCAP Content Validation Tool 1.2.16 cannot be found on official page anymore, and Jenkins NIST job is failing.
This PR is based on tool I had on my machine.

There is SCAP Content Validation Tool 1.3.2 available, but it returns thousands of errors when validation SCAP 1.2. Still need to figure out what are there errors and if I'm using it correctly.
I plan to update Jenkins with `scapval 1.3.2` when I get to know how to use it.

PR #2815, opened 2018-05-02
Stop shadowing 'package_prelink_removed'

jan-cerny:
#### Description:
The shared/checks/oval/package_prelink_removed.xml file was shadowed
in RHEL7 and SLE12 by OVAL generated from template. In Fedora and
RHEL6, the shared version was used. We remove the OVAL from shared
and generate it in Fedora and RHEL6 from template as well. We remove
it from SLE12, because there is no rule in SLE12 checking for prelink
package.

#### Rationale:
In https://github.com/OpenSCAP/scap-security-guide/pull/2804 many cases of shadowing and duplicating have been found. With templating and Jinja we can clean up them.


jan-cerny:
It probably can't go to shared template, because there are products the do not use this OVAL, eg. Fedora.

mpreisler:
> It probably can't go to shared template, because there are products the do not use this OVAL, eg. Fedora.

yes, IMO have it in the products but remove the genertaed file from shared

jan-cerny:
Actually, the `shared/checks/oval/package_prelink_removed.xml` is in shared in current master. That means the OVAL definition is now part of all the product.

And I was wrong with Fedora, because Fedora doesn't list `prelink` in the templates, but uses the `shared/checks/oval/package_prelink_removed.xml`

This is a mess :smile: 

I think the correct approach would be:
Remove the file `shared/checks/oval/package_prelink_removed.xml`, do not add anything to `shared/templates/csv/packages_removed.csv`, add `prelink` to CSVs in every product that has rule "Disable prelink" in any of their profiles.
Am I correct?

jan-cerny:
I have amended the commit. I think now it is OK.

yuumasato:
Looks good, Ack.
Thank you for helping cleaning SSG.

PR #2817, opened 2018-05-02
Change Value sshd_required selector to True False

yuumasato:
#### Description:

- Changed selector for Value sshd_required default / True / False

#### Rationale:

- Selector no/yes for Value sshd_required in yaml file was transformed to
True/False by changes in yaml loader, while selector in refine-value for
the Value remained as yes/no.
- SSHD rules using `sshd_required` value were resulting in `unknown`.


mpreisler:
Seems like a workaround rather than a fix, can't you change the yaml type instead iwth the yaml escape syntax?

redhatrises:
Closing this in favor of #2818

PR #2818, opened 2018-05-02
Change yaml.Loader to yaml.SafeLoader

redhatrises:
#### Description:

- Change yaml.Loader to yaml.SafeLoader

#### Rationale:

- Changed was missed in #2800 
- Fixes #2817

### DIFF
--- a/shared/modules/ssgcommon.py
+++ b/shared/modules/ssgcommon.py
@@ -249,7 +249,7 @@ def open_yaml(yaml_file, product_yaml=None):
         return self.construct_scalar(node)
 
     # Don't follow python bool case
-    yaml.Loader.add_constructor(u'tag:yaml.org,2002:bool', bool_constructor)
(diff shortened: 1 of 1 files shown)


shared/modules/ssgcommon.py
```diff
@@ -249,7 +249,7 @@ def bool_constructor(self, node):
         return self.construct_scalar(node)
 
     # Don't follow python bool case
-    yaml.Loader.add_constructor(u'tag:yaml.org,2002:bool', bool_constructor)
+    yaml.SafeLoader.add_constructor(u'tag:yaml.org,2002:bool', bool_constructor)
```
mpreisler: wait... every time we load a yaml we add a constructor to the static class?

mpreisler: shouldn't this be called right after we import "yaml" module?

redhatrises: @mpreisler it doesn't have to. I moved it to the top after the imports.

PR #2819, opened 2018-05-02
Add mcafee_antivirus_definitions_updated OVAL and XCCDF variable

redhatrises:
- Add mcafee_antivirus_definitions_updated OVAL and XCCDF variable
- Set the RHEL7 DISA profile to 1 week

### DIFF
--- a/shared/checks/oval/mcafee_antivirus_definitions_updated.xml
+++ b/shared/checks/oval/mcafee_antivirus_definitions_updated.xml
@@ -1,7 +1,7 @@
 <def-group oval_version="5.10">
   <definition class="compliance" id="mcafee_antivirus_definitions_updated" version="1">
     <metadata>
-      <title>Confirm Existence and Permissions of System Log Files</title>
+      <title>McAfee AntiVirus Definitions Updated</title>
       <affected family="unix">
         <platform>multi_platform_all</platform>
       </affected>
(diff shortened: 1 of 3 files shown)


shared/checks/oval/mcafee_antivirus_definitions_updated.xml
```diff
@@ -0,0 +1,49 @@
+<def-group oval_version="5.10">
+  <definition class="compliance" id="mcafee_antivirus_definitions_updated" version="1">
+    <metadata>
+      <title>Confirm Existence and Permissions of System Log Files</title>
```
yuumasato: Title of this check is not related to antiviruses.

shared/checks/oval/mcafee_antivirus_definitions_updated.xml
```diff
@@ -0,0 +1,49 @@
...
+
+  <local_variable id="variable_mcafee_dat_files_mtime" version="1" comment="Age of file for each McAfee definition file." datatype="int">
+      <time_difference format_2="seconds_since_epoch">
+        <object_component object_ref="mcafee_dat_files_mtime" item_field="m_time"/> 
+      </time_difference>
+  </local_variable>
+
+  <ind:variable_test check="all" check_existence="all_exist" id="test_mcafee_antivirus_definitions_updated" version="1" comment="McAfee AntiVirus definitions have been updated">
+    <ind:object object_ref="object_mcafee_antivirus_definitions_updated" />
+    <ind:state state_ref="state_mcafee_antivirus_definitions_updated" />
```
yuumasato: Are the objects and states always named after the name of tests?

Nitpick:
Would be more clear if this object was called something like "antivirus definitions modified time".
And state called "antivirus definition timeout", or "max modified time".

shared/checks/oval/mcafee_antivirus_definitions_updated.xml
redhatrises: hahaha... thanks @yuumasato

shared/checks/oval/mcafee_antivirus_definitions_updated.xml
redhatrises: > Are the objects and states always named after the name of tests?

They are done more often than not, but updated them per your suggestion.

yuumasato:
Ack.

PR #2820, opened 2018-05-02
Fix dconf_gnome_disable_geolocation script and add missing dconf remedation scripts

redhatrises:
#### Description:

- Fix dconf_gnome_disable_geolocation script and add missing dconf remediation scripts

#### Rationale:

- Script was erroring out on remediation and a couple of remediation scripts were missing

### DIFF
--- a/shared/fixes/bash/dconf_gnome_disable_wifi_create.sh
+++ b/shared/fixes/bash/dconf_gnome_disable_wifi_create.sh
@@ -1,4 +1,4 @@
-# platform = Red Hat Enterprise Linux 7
+# platform = Red Hat Enterprise Linux 7, multi_platform_fedora
 . /usr/share/scap-security-guide/remediation_functions
 
 include_dconf_settings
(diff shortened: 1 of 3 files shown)


shared/fixes/bash/dconf_gnome_disable_wifi_create.sh
```diff
@@ -0,0 +1,7 @@
+# platform = Red Hat Enterprise Linux 7
```
mpreisler: if it's rhel7 only don't put it in shared, if it's universal don't make it rhel7 only

redhatrises: The majority of rules in shared are rhel7 only.

redhatrises:
@yuumasato if still possible need this in 0.1.39

shared/fixes/bash/dconf_gnome_disable_wifi_create.sh
redhatrises: @mpreisler added to Fedora as well.

yuumasato:
@redhatrises sure, just this one.

redhatrises:
yup @yuumasato just this one.

yuumasato:
Looks good, Ack.

shared/fixes/bash/dconf_gnome_disable_wifi_create.sh
mpreisler: > The majority of rules in shared are rhel7 only.

yes and that sucks, let's not make it even worse :D

PR #2825, opened 2018-05-03
Remove duplicated openSUSE CPE UNIX family test

abergmann:
Remove duplicated UNIX family test `test_opensuse_unix_family` and link
against the more generic UNIX family test via `extend_definition`.


mpreisler:
ACK. Thx!

PR #2835, opened 2018-05-15
Convert XSLT macros to jinja

matejak:
Rationale: XSLT is difficult to work with, jinja is not, while being equally powerful.

Important points about this PR:
- The macro definitions location is `shared/modules/macros.jinja`. Maybe those should be elsewhere?
- Jinja is designed to use include and import statements in templates in order to read definitions of macros and use them afterwards. We want something else - the template without include or imports, and the template interpreter should be able to expand all macros. Check out `open_and_macro_expand_yaml` for implementation.

### DIFF
--- a/shared/modules/macros.jinja
+++ b/shared/modules/macros.jinja
@@ -67,3 +67,24 @@ ocil: |
 
 ocil_clause: "no line is returned"
 {{%- endmacro %}}
+
+
+{{% macro audit_syscall_check(syscall) -%}}
+    To determine if the system is configured to audit calls to the
+    <code>{{{ syscall }}}</code> system call, run the following command:
+    <pre space="preserve">$ sudo grep "{{{ syscall }}}" /etc/audit/audit.rules</pre>
+    If the system is configured to audit this activity, it will return a line.
+{{%- endmacro %}}
+
+
+{{% macro audit_syscall_check_ocil_clause_entry() -%}}
+ocil_clause: "no line is returned"
+{{%- endmacro %}}
+
+
+{{% macro audit_syscall_check_complete_ocil_entry(syscall) -%}}
+ocil: |
+    {{{ audit_syscall_check(syscall) }}}
+
+{{{ audit_syscall_check_ocil_clause_entry() }}}
+{{%- endmacro %}}
--- a/shared/transforms/shared_shorthand2xccdf.xslt
+++ b/shared/transforms/shared_shorthand2xccdf.xslt
@@ -401,7 +401,6 @@
               <xsl:if test="service-enable-check-macro">the service is not running</xsl:if>
               <xsl:if test="package-check-macro">the package is installed</xsl:if>
               <xsl:if test="module-disable-check-macro">no line is returned</xsl:if>
-              <xsl:if test="audit-syscall-check-macro">no line is returned</xsl:if>
             </xsl:attribute>
 
             <!-- add clause if explicitly specified (and also override any above) -->
@@ -854,14 +853,6 @@ These lines instruct the module loading system to run another program (such as
 Run the following command to search for such lines in all files in <xhtml:code>/etc/modprobe.d</xhtml:code>
 and the deprecated <xhtml:code>/etc/modprobe.conf</xhtml:code>:
 <xhtml:pre xml:space="preserve">$ grep -r <xsl:value-of select="@module"/> /etc/modprobe.conf /etc/modprobe.d</xhtml:pre>
-  </xsl:template>
-
-  <xsl:template match="audit-syscall-check-macro">
-To determine if the system is configured to audit calls to
-the <xhtml:code><xsl:value-of select="@syscall"/></xhtml:code>
-system call, run the following command:
-<xhtml:pre xml:space="preserve">$ sudo grep "<xsl:value-of select="@syscall"/>" /etc/audit/audit.rules</xhtml:pre>
-If the system is configured to audit this activity, it will return a line.
   </xsl:template>
 
 <xsl:template match="auditctl-syscall-check-macro">
--- a/rhel6/guide/system/auditing/auditd_configure_rules/audit_dac_actions/audit_rules_dac_modification_chmod.rule
+++ b/rhel6/guide/system/auditing/auditd_configure_rules/audit_dac_actions/audit_rules_dac_modification_chmod.rule
@@ -30,7 +30,7 @@ references:
 
 ocil_clause: 'the system is not configured to audit permission changes'
 
-ocil: '<audit-syscall-check-macro syscall="chmod" />'
+{{{ audit_syscall_check_complete_ocil_entry(syscall="chmod") }}}
 
 warnings:
     - general: |-
--- a/rhel6/guide/system/auditing/auditd_configure_rules/audit_dac_actions/audit_rules_dac_modification_chown.rule
+++ b/rhel6/guide/system/auditing/auditd_configure_rules/audit_dac_actions/audit_rules_dac_modification_chown.rule
@@ -28,7 +28,7 @@ references:
     srg: SRG-OS-000064
     stigid: RHEL-06-000185
 
-ocil: '<audit-syscall-check-macro syscall="chown" />'
+{{{ audit_syscall_check_complete_ocil_entry(syscall="chown") }}}
 
 warnings:
     - general: |-
--- a/rhel6/guide/system/auditing/auditd_configure_rules/audit_dac_actions/audit_rules_dac_modification_fchmod.rule
+++ b/rhel6/guide/system/auditing/auditd_configure_rules/audit_dac_actions/audit_rules_dac_modification_fchmod.rule
@@ -28,7 +28,7 @@ references:
     srg: SRG-OS-000064
     stigid: RHEL-06-000186
 
-ocil: '<audit-syscall-check-macro syscall="fchmod" />'
+{{{ audit_syscall_check_complete_ocil_entry(syscall="fchmod") }}}
 
 warnings:
     - general: |-
--- a/rhel6/guide/system/auditing/auditd_configure_rules/audit_dac_actions/audit_rules_dac_modification_fchmodat.rule
+++ b/rhel6/guide/system/auditing/auditd_configure_rules/audit_dac_actions/audit_rules_dac_modification_fchmodat.rule
@@ -28,7 +28,7 @@ references:
     srg: SRG-OS-000064
     stigid: RHEL-06-000187
 
-ocil: '<audit-syscall-check-macro syscall="fchmodat" />'
+{{{ audit_syscall_check_complete_ocil_entry(syscall="fchmodat") }}}
 
 warnings:
     - general: |-
--- a/rhel6/guide/system/auditing/auditd_configure_rules/audit_dac_actions/audit_rules_dac_modification_fchown.rule
+++ b/rhel6/guide/system/auditing/auditd_configure_rules/audit_dac_actions/audit_rules_dac_modification_fchown.rule
@@ -28,7 +28,7 @@ references:
     srg: SRG-OS-000064
     stigid: RHEL-06-000188
 
-ocil: '<audit-syscall-check-macro syscall="fchown" />'
+{{{ audit_syscall_check_complete_ocil_entry(syscall="fchown") }}}
 
 warnings:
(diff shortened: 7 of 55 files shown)


matejak:
@openscap-ci test this please

debian8/guide/system/fs-part/installation-storage-partitioning/partition_for_home.rule
```diff
@@ -21,4 +21,4 @@ references:
     disa: 1208
     nist: SC-32
 
-ocil: '<partition-check-macro part="/home " />'
+ocil: '{{{ partition_check(part="/home ") }}}'
```
redhatrises: @matejak any reason for the space in all of the partition_checks e.g. ```"/home "``` instead of ```"/home"```

debian8/guide/system/fs-part/installation-storage-partitioning/partition_for_srv.rule
```diff
@@ -20,4 +20,4 @@ severity: unknown
 references:
     anssi: NT28(R12)
 
-ocil: '<partition-check-macro part="/home " />'
+ocil: '{{{ partition_check(part="/home ") }}}'
```
redhatrises: @matejak we should probably be consistent and either use single quotes or remove them.

mpreisler:
Shouldn't ocil be in the name of the macros? It is generating ocil and just ocil.

redhatrises:
> Shouldn't ocil be in the name of the macros? It is generating ocil and just ocil.

@mpreisler I wouldn't think so as a the macro could be used in the description.

shared/modules/ssgcommon.py
```diff
@@ -242,29 +244,71 @@ def get_jinja_environment():
...
-        with codecs.open(yaml_file, "r", "utf8") as stream:
-            yaml_contents = yaml.safe_load(stream)
-    else:
-        yaml_contents = yaml.safe_load(
-            process_file_with_jinja(yaml_file, product_yaml)
-        )
 
-    if "documentation_complete" in yaml_contents and \
-            yaml_contents["documentation_complete"] == "false":
+def _extract_substitutions_dict_from_template(filename):
```
jan-cerny: What if `macros.jinja` does not exist or is invalid? Could it happen?

shared/modules/macros.jinja
```diff
@@ -0,0 +1,51 @@
...
+	port=port,
+) }}}
+{{%- endmacro %}}
+
+
+{{% macro iptables_desc(traffic_action, how_to_do, proto, port) -%}}
+    To configure <tt>iptables</tt> {{{ traffic_action }}} port {{{ port }}} traffic, one must edit
+    <tt>/etc/sysconfig/iptables</tt> and
+    <tt>/etc/sysconfig/ip6tables</tt> (if IPv6 is in use).
+    {{{ how_to_do }}}:
```
jan-cerny: This produces 2 semicolons, because both how_to_do options contain semicolon.

shared/modules/macros.jinja
```diff
@@ -0,0 +1,51 @@
+
+{{# Example usage: sshd_check(default="no", option="Banner", value="/etc/issue") #}}
+{{% macro sshd_check(default, option, value) -%}}
+|-
```
jan-cerny: This `|-` appears in `ssg-rhel7-ocil.xml`

shared/transforms/shared_shorthand2xccdf.xslt
```diff
@@ -394,7 +394,6 @@
               <!-- add clauses if specific macros are found within -->
               <xsl:if test="sysctl-check-macro">the correct value is not returned</xsl:if>
               <xsl:if test="fileperms-check-macro or fileowner-check-macro or filegroupowner-check-macro">it does not</xsl:if>
-              <xsl:if test="partition-check-macro">no line is returned</xsl:if>
```
jan-cerny: These strings will be missing in OCIL. It creates  a weird question in OCIL: `Is it the case that ?`

shared/modules/macros.jinja
matejak: Nice catch, I have replaced it with a pipe alone as indicated here: https://stackoverflow.com/questions/15540635/what-is-the-use-of-pipe-symbol-in-yaml

shared/modules/macros.jinja
matejak: Nice catch, thanks!

debian8/guide/system/fs-part/installation-storage-partitioning/partition_for_srv.rule
matejak: I agree.

shared/transforms/shared_shorthand2xccdf.xslt
matejak: Indeed, this is an issue that has to be solved somehow before merging this PR.

shared/modules/macros.jinja
```diff
@@ -0,0 +1,51 @@
...
+    {{% if default == "yes" -%}}
+    If no line, a commented line, or a line indicating the value <tt>{{{ value }}}</tt> is returned, then the required value is set.
+    {{%- endif %}}
+    {{% if default == "no" or default == '' -%}}
+    If a line indicating <tt>{{{ value }}}</tt> is returned, then the required value is set.
+    {{%- endif %}}
+{{%- endmacro %}}
+
+
+{{# Example usage: iptables_describe_allow(proto="tcp", port=80) #}}
```
yuumasato: Example usage is for `allow` instead of `disallow`

shared/modules/macros.jinja
```diff
@@ -10,3 +11,33 @@
     If a line indicating {{{ value }}} is returned, then the required value is set.
     {{%- endif %}}
 {{%- endmacro %}}
+
+
+{{# Example usage: iptables_describe_allow(proto="tcp", port=80) #}}
+{{% macro iptables_describe_disallow(proto, port) -%}}
+{{{ iptables_desc(
+	traffic_action="to block",
+	how_to_do="Remove the following line, ensuring that it does not appear the INPUT chain:",
```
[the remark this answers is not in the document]
mpreisler: use block or disallow but not both

mpreisler:
@openscap-ci test this please

debian8/guide/system/fs-part/installation-storage-partitioning/partition_for_home.rule
matejak: It was just backwards compatibility :-)

shared/transforms/shared_shorthand2xccdf.xslt
matejak: Resolved. I have rebased to keep the mapping commit <-> template converted.

shared/modules/ssgcommon.py
```diff
@@ -242,29 +244,75 @@ def get_jinja_environment():
...
+
+def open_yaml(yaml_file):
+    """
+    Open given file-like object and parse it as YAML
+    without performing any kind of template processing
+
+    See also: _open_yaml
+    """
+    with codecs.open(yaml_file, "r", "utf8") as stream:
+        yaml_contents = _open_yaml(stream)
```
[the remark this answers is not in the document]
matejak: Why not, I have done so in the rebase.

matejak:
This PR is no more WIP, more macros will be converted in other PRs. I propose using the following snippet to make conversions more automated:

```
#!/bin/bash

set -eo pipefail

dirs_to_examine=( chromium debian8 docs eap6 fedora firefox fuse6 jboss_eap6 jboss_fuse6 jre ocp3 ol7 opensuse rhel6 rhel7 rhel-osp7 rhosp7 sle11 sle12 ubuntu14 ubuntu1404 ubuntu16 ubuntu1604 webmin wrlinux )
# TODO: What other shared dirs to search in?
dirs_to_examine+=( shared/guide )

get_files_to_subst() {
        local _old="$1" IFS=$'\n'
        files_to_subst=( $(grep -Rl -e "$_old" "${dirs_to_examine[@]}") )
}


macro2jinja() {
        local _old="$1" _new="$2"
        get_files_to_subst "$1"
        for fname in "${files_to_subst[@]}"; do
                test $(grep -c -e "$_old" "$fname") -eq 1 || { printf "%s\n" "File $fname contains more than one occurence of the macro call. You need to process that manually" >&2; continue; }
                # good = find macro invocation -> extract macro arguments -> put ',' after double quotes followed by space -> get rid of the trailing comma and put args into the jinja macro invocation.
                good="$(grep -e "$_old" "$fname" | sed -e "s/.*<$_old"'\s\+\(.*\)\s*\/.*/\1/' | sed -e 's/"\s\+/", /g' | sed -e 's/\(.*\),\s*$/{{{ '"$_new"'(\1) }}}/')"
                sed -i "s~<$_old.*>~$good~" "$fname"
        done
}


# Run e.g. bash macro2jinja firewalld-desc-macro firewalld_desc
if test "$#" = 2; then
        macro2jinja "$1" "$2"
fi
```

debian8/guide/services/ssh/ssh_server/sshd_disable_root_login.rule
```diff
@@ -23,4 +23,5 @@ references:
     anssi: NT007(R21)
     nist: AC-3,AC-6(2),IA-2(1)
 
-ocil: '<sshd-check-macro default="no" option="PermitRootLogin" value="no" />'
+
+{{{ sshd_check_complete_ocil_entry(default="no", option="PermitRootLogin", value="no") }}}
```
redhatrises: This seems odd to me. I don't think that people who are used to Ansible and OpenControl would find this familiar either. I understand the purpose and reasoning, but I would expect this:

```
ocil: {{{ sshd_check_complete_ocil_entry(default="no", option="PermitRootLogin", value="no") }}}
```

redhatrises: Also there is an added whitespace for no reason.

matejak: Have you checked out the full definition of the macro? There was an issue that an ocil clause has to be added as well, and the `complete_ocil_entry` refers to adding `ocil` as well as the corresponding `ocil_clause`. As the clause should be added every time, it made sense to me to make a encompassing macro, so there are no omissions.

shared/modules/macros.jinja
```diff
@@ -0,0 +1,75 @@
...
+{{% macro partition_check_complete_ocil_entry(part) -%}}
+ocil: |
+    {{{ partition_check(part) }}}
+
+{{{ partition_check_ocil_clause_entry() }}}
+{{%- endmacro %}}
+
+
+{{% macro partition_check_ocil_clause_entry() -%}}
+ocil_clause: "no line is returned"
```
mpreisler: so this is defined after it's used which is bad for readability, I also don't see why it's even necessary? Why not just include it in the first macro?

shared/utils/yaml-to-shorthand.py
```diff
@@ -19,6 +19,7 @@
         os.path.dirname(os.path.dirname(os.path.realpath(__file__))),
         "modules"))
 from ssgcommon import open_yaml, required_yaml_key
```
mpreisler: the open_yaml import is no longer necessary

shared/modules/macros.jinja
matejak: It was probably for the sake of symmetry, but I have inlined it.

debian8/guide/services/ssh/ssh_server/sshd_disable_root_login.rule
matejak: I have also removed those whitespaces.

shared/modules/macros.jinja
```diff
@@ -0,0 +1,90 @@
...
+{{{ iptables_desc(
+	traffic_action="to allow",
+	how_to_do="Add the following line, ensuring that it appears before the final LOG and DROP lines for the INPUT chain:",
+	proto=proto,
+	port=port
+) }}}
+{{%- endmacro %}}
+
+
+{{% macro iptables_desc(traffic_action, how_to_do, proto, port) -%}}
```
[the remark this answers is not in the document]
mpreisler: it's fine to do it in another PR as well, I am just saying

Issue #2836, opened 2018-05-15
Abrupt end of remediation.


matejak:
The bash remediation script for CentOS 7 , hard stops at fix 137/243 in ssg-centos7-role-stig-rhel7-disa.sh.  It does not complete the remainder of the script.

PR #2837, opened 2018-05-15
Fixed unintended behavior of remediations.

matejak:
Fixes: #2836

- Actual remediations were moved into functions.
- The 'exit' statement has been substituted by return, so series of
  remediations doesn't end abruptly.
- Minor bug has been fixed in shared functionality.


jan-cerny:
This patch looks good to me.

PR #2843, opened 2018-05-17
Readme improvements

mpreisler:
Make it clearer that this project generates more than just SCAP files. Update some of the examples.

More improvements are coming.


konstruktoid:
Markdown lint:
```
$ mdl README.md | sed 's/.*://g' | sort | uniq
 MD009 Trailing spaces
 MD010 Hard tabs
 MD012 Multiple consecutive blank lines
 MD013 Line length
 MD033 Inline HTML
 MD034 Bare URL used
```

README.md
```diff
@@ -1,46 +1,111 @@
 ## Welcome!
-The purpose of this project is to create *SCAP content* for various
-platforms -- Red Hat Enterprise Linux 6, Red Hat Enterprise Linux 7,
-Fedora, and others.  *"SCAP content"* refers to documents  in the *XCCDF*,
-*OVAL* and *Source DataStream* formats.  These documents can be presented
+
+[![Nightly ZIP Status](https://jenkins.open-scap.org/job/scap-security-guide-nightly-zip/badge/icon)](https://jenkins.open-scap.org/job/scap-security-guide-nightly-zip/) [![Nightly 5.10 ZIP Status](https://jenkins.open-scap.org/job/scap-security-guide-nightly-oval510-zip/badge/icon)](https://jenkins.open-scap.org/job/scap-security-guide-nightly-oval510-zip/) [![Link-checker Status](https://jenkins.open-scap.org/job/scap-security-guide-linkcheck/badge/icon)](https://jenkins.open-scap.org/job/scap-security-guide-linkcheck/) [![CentOS CI Status](https://ci.centos.org/job/openscap-scap-security-guide/badge/icon)](https://ci.centos.org/job/openscap-scap-security-guide/) [![landscape.io Code Health](https://landscape.io/github/OpenSCAP/scap-security-guide/master/landscape.png)](https://landscape.io/github/OpenSCAP/scap-security-guide/master)
```
matejak: I suggest breaking badge entries one per line, it should still display them side-by-side.
And concerning the Landscape badge, I think that we should take it out as checks are not running and the displayed value is stale. We can re-add it in another PR.

README.md
```diff
@@ -1,46 +1,111 @@
...
+appear in all of our output formats but are all sourced from the YAML rule files.
 
-The goal of this project to enable the creation of multiple security
-baselines from a single set of high-quality SCAP content.
+We understand that depending on your organization's needs you may need
+to use a specific security content format. We let you choose.
 
-The SSG homepage is https://www.open-scap.org/security-policies/scap-security-guide/
+---
+Our OpenControl-inspired YAML rule format. Write once to build security content
```
matejak: * The "Our OpenControl-inspired YAML rule format..." appears out of nowhere - is it one of the choices? Probably not, as there are no other choices following. If the main purpose is to show the input format, it should be clarified.
* What about linking to the OpenControl project?

README.md
mpreisler: I experimented with one badge per line and IMO it didn't look good. I kept the landscape.io because it was already there, I can remove it or we can figure out wtf is going on with it in another PR. Either way, landscape was there already.

matejak: I am OK with it, at least there is one more reason to re-add it :-)

matejak:
Great contribution that keeps the shape of the project's outer shell reflect improving shape of project's internals.

PR #2844, opened 2018-05-17
Add warnings into YAML

redhatrises:
#### Description:

- Add Warnings to YAML

#### Rationale:

- Missed in the YAML conversion

### DIFF
--- a/shared/utils/yaml-to-shorthand.py
+++ b/shared/utils/yaml-to-shorthand.py
@@ -21,11 +21,10 @@ sys.path.insert(0, os.path.join(
 from ssgcommon import open_yaml, required_yaml_key
 
 
-def add_warning(element, warnings):
-    if warnings:
-        for i, warning_list in enumerate(warning for warning in warnings):
-            warning = add_sub_element(element, "warning", warning_list.values()[0])
-            warning.set("category", warning_list.keys()[0])
+def add_warnings(element, warnings):
+    for warning_list in warnings:
+        warning = add_sub_element(element, "warning", warning_list.values()[0])
+        warning.set("category", warning_list.keys()[0])
 
     return element
 
@@ -118,7 +117,7 @@ class Value(object):
         value.description = required_yaml_key(yaml_contents, "description")
         value.type = required_yaml_key(yaml_contents, "type")
         value.options = required_yaml_key(yaml_contents, "options")
-        value.warning = yaml_contents.get("warnings", [])
+        value.warnings = yaml_contents.get("warnings", [])
 
         return value
 
@@ -130,8 +129,8 @@ class Value(object):
         title.text = self.title
         add_sub_element(value, 'description', self.description)
 
-        if self.id_ is not "conditional_clause":
-            value = add_warning(value, self.warning)
+        if self.id_ != "conditional_clause":
+            add_warnings(value, self.warnings)
 
         for selector, option in self.options.items():
             # do not confuse Value with big V with value with small v
@@ -310,7 +309,7 @@ class Group(object):
         group.prodtype = yaml_contents.get("prodtype", "all")
         group.title = required_yaml_key(yaml_contents, "title")
         group.description = required_yaml_key(yaml_contents, "description")
-        group.warning = yaml_contents.get("warnings", [])
+        group.warnings = yaml_contents.get("warnings", [])
 
         return group
 
@@ -323,8 +322,8 @@ class Group(object):
         title.text = self.title
         add_sub_element(group, 'description', self.description)
 
-        if self.id_ is not "conditional_clause":
-            group = add_warning(group, self.warning)
+        if self.id_ != "conditional_clause":
+            add_warnings(group, self.warnings)
 
         for v in self.values.values():
             group.append(v.to_xml_element())
@@ -383,7 +382,7 @@ class Rule(object):
         rule.ocil_clause = yaml_contents.get("ocil_clause")
         rule.ocil = yaml_contents.get("ocil")
         rule.external_oval = yaml_contents.get("oval_external_content")
-        rule.warning = yaml_contents.get("warnings", [])
+        rule.warnings = yaml_contents.get("warnings", [])
 
         return rule
 
@@ -448,7 +447,7 @@ class Rule(object):
             if self.ocil_clause:
                 ocil.set("clause", self.ocil_clause)
 
-        rule = add_warning(rule, self.warning)
+        add_warnings(rule, self.warnings)
 
         return rule
 
--- a/shared/utils/yaml-to-shorthand.py
+++ b/shared/utils/yaml-to-shorthand.py
@@ -38,6 +38,17 @@ def add_sub_element(parent, tag, data):
 
 
 def add_warning_elements(element, warnings):
+    # The use of [{dict}, {dict}] in warnings is to handle the following
+    # scenario where multiple warnings have the same category which is
+    # valid in SCAP and our content:
+    #
+    # warnings:
+    #     - general: Some general warning
+    #     - general: Some other general warning
+    #     - general: |-
+    #         Some really long multiline general warning
+    #
+    # Each of the {dict} should have only one key/value pair.
     for warning_dict in warnings:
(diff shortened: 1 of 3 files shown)


mpreisler:
> @mpreisler is it easier to understand now?

I still don't get the decision to use a list of dictionaries, if you assume one warning of each category make the whole thing a dict, that's fine and will work. the code will be simpler. the code actually assumes that ( but without an assert)

shared/utils/yaml-to-shorthand.py
```diff
@@ -21,6 +21,14 @@
 from ssgcommon import open_yaml, required_yaml_key
 
 
+def add_warnings(element, warnings):
+    for warning_dict in warnings:
+        warning = add_sub_element(element, "warning", warning_dict.values()[0])
+        warning.set("category", warning_dict.keys()[0])
```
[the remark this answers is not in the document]
redhatrises: correct. The YAML creates a list of dictionaries. We need to be able to handle multiple instances of the same warning type e.g. general, functionality, etc.

shared/utils/yaml-to-shorthand.py
```diff
@@ -21,6 +21,14 @@
...
+
+    return element
```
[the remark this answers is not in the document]
redhatrises: Don't you need to return the Etree element back to the function calling it?

shared/utils/yaml-to-shorthand.py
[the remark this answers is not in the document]
mpreisler: why not just a dict?

shared/utils/yaml-to-shorthand.py
```diff
@@ -130,8 +129,8 @@ def to_xml_element(self):
         title.text = self.title
         add_sub_element(value, 'description', self.description)
 
-        if self.id_ is not "conditional_clause":
-            value = add_warning(value, self.warning)
+        if self.id_ != "conditional_clause":
+            add_warnings(value, self.warnings)
```
mpreisler: why is this condition even here? will it pick up any warnings for conditional clause?

shared/utils/yaml-to-shorthand.py
```diff
@@ -323,8 +322,8 @@ def to_xml_element(self):
         title.text = self.title
         add_sub_element(group, 'description', self.description)
 
-        if self.id_ is not "conditional_clause":
-            group = add_warning(group, self.warning)
+        if self.id_ != "conditional_clause":
+            add_warnings(group, self.warnings)
```
mpreisler: conditional_clause Group?!

shared/utils/yaml-to-shorthand.py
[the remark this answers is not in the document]
mpreisler: no?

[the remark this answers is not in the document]
mpreisler: (in practice this is going to be pass by reference)

shared/utils/yaml-to-shorthand.py
```diff
@@ -21,6 +21,14 @@
 from ssgcommon import open_yaml, required_yaml_key
 
 
+def add_warnings(element, warnings):
```
mpreisler: IMO rename to add_warning_elements and move after add_sub_element. You are using add_sub_element in the function so it helps when it's defined after you define add_sub_element

shared/utils/yaml-to-shorthand.py
redhatrises: @mpreisler yes.... conditional_clause kept showing up and throwing errors.

shared/utils/yaml-to-shorthand.py
redhatrises: @mpreisler yes.... conditional_clause kept showing up and throwing errors.

shared/utils/yaml-to-shorthand.py
mpreisler: but its warnings are going to be empty list, right? the issue is elsewhere, this just hides it.

the issue does not originate in this PR, it looks like we do not init all membes in all the ctors. once you fix that it will work.

redhatrises:
> I still don't get the decision to use a list of dictionaries, if you assume one warning of each category make the whole thing a dict, that's fine and will work. the code will be simpler. the code actually assumes that ( but without an assert)

@mpreisler we need to support multiple warnings of the same category which existed in the old content. Is there a way that we can do the following in YAML?

```
warning:
    general: |-
        blah blah blah
    general: |-
        blah blah blah
```

shared/utils/yaml-to-shorthand.py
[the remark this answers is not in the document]
redhatrises: @mpreisler correct me if I am wrong, but a dict cannot handle the following:

```
warning:
    general: |-
        blah blah blah
    general: |-
        blah blah blah
```

mpreisler:
> @mpreisler we need to support multiple warnings of the same category which existed in the old content. Is there a way that we can do the following in YAML?

this was the reason for my comment, you had all this additional logic but then only used the first key and value.

what you are looking for is a multimap, I am not sure if yaml supports that

shared/utils/yaml-to-shorthand.py
mpreisler: as I said, the problem is that the constructors (__init__ methods) don't initialize all members. fix that and the issue will go away - no need for the self.id_ checks above.

mpreisler: conditional_clause is there for OCIL, ocil won't work without it AFAIK

shared/utils/yaml-to-shorthand.py
```diff
@@ -45,6 +37,14 @@ def add_sub_element(parent, tag, data):
     return element
 
 
+def add_warning_elements(element, warnings):
+    for warning_dict in warnings:
+        warning = add_sub_element(element, "warning", warning_dict.values()[0])
+        warning.set("category", warning_dict.keys()[0])
```
mpreisler: I think I get the syntax now but this *needs* to be explained here. the code looks like a total WTF

please also assert that there is exactly one key and exactly one value

shared/utils/yaml-to-shorthand.py
```diff
@@ -45,6 +37,14 @@ def add_sub_element(parent, tag, data):
...
+
+    return element
```
mpreisler: no need to return

shared/utils/yaml-to-shorthand.py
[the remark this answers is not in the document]
redhatrises: Looks like the return is needed; otherwise, it is `None`.

shared/utils/yaml-to-shorthand.py
redhatrises: I get an error as it "returns" `None` if the return doesn't exist.

shared/utils/yaml-to-shorthand.py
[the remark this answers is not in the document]
mpreisler: no, it is definitely not necessary, don't use the return value of this function, don't even assign it. the function will affect what you pass into it because python is pass by object or pass by ref or whatever you call it.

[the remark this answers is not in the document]
mpreisler: `new_element = add_warning_elements(old_element, warnings)` will affect old_element without copying it! id(new_element) will be exactly the same as id(old_element).

redhatrises:
@mpreisler ready for your re-review.

shared/utils/yaml-to-shorthand.py
```diff
@@ -37,6 +37,23 @@ def add_sub_element(parent, tag, data):
...
+    # scenario where multiple warnings have the same category which is
+    # valid in SCAP and our content:
+    #
+    # warnings:
+    #     - general: Some general warning
+    #     - general: Some other general warning
+    #     - general: |-
+    #         Some really long multiline general warning
+    #
+    # Each of the {dict} should have only one key/value pair.
```
mpreisler: +1 for the comment!

mpreisler:
looks good now, thx!

ACK

PR #2845, opened 2018-05-18
Fix broken Python 3 build

jan-cerny:
In Python 2, dict methods items(), keys() and values() return
a list. But in Python 3, they return view objects. The returned
value needs to be converted to a list explicitely if we want to
access the members using an index.
Addressing:
```
  File "/home/jcerny/scap-security-guide/shared/utils/yaml-to-shorthand.py", line 53, in add_warning_elements
    warning = add_sub_element(element, "warning", warning_dict.values()[0])
TypeError: 'dict_values' object does not support indexing
```


mpreisler:
Crap, missed that in the review.

Good catch, ACK!

redhatrises:
Thanks @jan-cerny for the catch!

PR #2846, opened 2018-05-18
Do not depend on Python 2

jan-cerny:
We can build SSG using both Python 2 and Python 3. However, on systems
where Python 2 is not installed, CMake will fail, because it tries to
find Python 2 interpreter. We do not have to require any version.


mpreisler:
Hmm, isn't the proper solution to look for python2 and python3, mark both of them optional and then check that at least one is found?

mpreisler:
Looked some more into this and it gets crazy complex, this is a good pragmatic solution, nobody has python1 anyway...

PR #2848, opened 2018-05-18
Install PyYAML and python-jinja2 for centos CI

mpreisler:
Otherwise centos CI fails. See https://ci.centos.org/job/openscap-scap-security-guide/262/console


redhatrises:
Ack.

PR #2849, opened 2018-05-18
Add remaining missing warning content

redhatrises:
#### Description:

- Add remaining missing warning content
- Remove warning from dracut rule as it was placed in the wrong file accidentally


mpreisler:
ACK

PR #2850, opened 2018-05-18
Unify service templates

mpreisler:
Almost lost my mind doing this...

Unified all the service templates, both enabled and disabled, removed extra BASH service disabled templates, they are essentially the same as the shared ones.


rhel6/templates/template_OVAL_service_disabled
```diff
@@ -1,98 +0,0 @@
...
-      <title>Service %SERVICENAME% Disabled</title>
-      <affected family="unix">
-        <platform>Red Hat Enterprise Linux 6</platform>
-      </affected>
-      <description>The %SERVICENAME% service should be disabled if possible.</description>
-    </metadata>
-   <criteria comment="package %PACKAGENAME% removed or service %SERVICENAME% is not configured to start" operator="OR">
-    <extend_definition comment="%PACKAGENAME% removed" definition_ref="package_%PACKAGENAME%_removed" />
-    <criteria operator="AND" comment="service %SERVICENAME% is not configured to start">
-      <criterion comment="%SERVICENAME% runlevel 0" test_ref="test_runlevel0_%SERVICENAME%" />
```
matejak: I see a discrepancy between this line of the RHEL6 template and [the new template template](https://github.com/OpenSCAP/scap-security-guide/pull/2850/files?utf8=%E2%9C%93&diff=split#diff-c71262d467c8b535cfbfd7b53f8eee36R51) (the `_off` suffix is there, but not here).

mpreisler: yeah, that was necessary because of conflicts with some hardcoded fixes. probably should have mentioned it here.

mpreisler: OVAL has this idea that objects that are the same can be reused. In practice this has backfired many many times for us. We even created warnings and triggers in the build system for it. I think we are best off if the object state and test are all defined in one file and only used in that one file. It can be a bit wasteful in the end but it prevents hard to find bugs.

matejak: OK, good to know.

matejak:
I absolutely love and appreciate this PR, which will enable us to get templates under control and reduce code duplication. However, it can't be merged yet:
I executed the SSG test suite, trying the `abled` rules, and this PR breaks check for `content_rule_service_firewalld_enabled` on RHEL7 content. By examining diffs between unlinked OVALs, it looks like that some sysvinit stuff got where systemd should be. The diff between the rhel7 oval before and after this PR is huge, part of it may be caused by different line order, but something is certainly wrong.

mpreisler:
> OVALs, it looks like that some sysvinit stuff got where systemd should be. The diff between the rhel7 oval before and after this PR is huge, part of it may be caused by different line order, but something is certainly wrong.

Fixed. I forgot to commit & push a hunk in `template_common.py`, good catch!

matejak:
OK, looks much better now. Thank you for this difficult and highly beneficial work!

PR #2855, opened 2018-05-22
Unified all the sysctl templates [DO NOT MERGE]

mpreisler:
There was no difference in them... Just copy pasted code...


mpreisler:
Ok, my diffing bash script was totally broken and wasn't giving me full results. I need to revisit this because there are some subtle differences.

mpreisler:
This needs to be done much more carefully, I will start over.

PR #2856, opened 2018-05-22
Unify kernel_module_disabled templates

redhatrises:
Only


shared/templates/template_OVAL_kernel_module_disabled
```diff
@@ -80,4 +104,23 @@
     <ind:pattern operation="pattern match">^\s*install\s+%KERNMODULE%\s+(/bin/false|/bin/true)$</ind:pattern>
     <ind:instance datatype="int">1</ind:instance>
   </ind:textfilecontent54_object>
+
+    <ind:textfilecontent54_object id="obj_kernmod_%KERNMODULE%_runmodprobed"
```
mpreisler: indents borked

mpreisler:
Is the rhel6 different because the code there is inferior to the newer code or does this really work differently on rhel6?

redhatrises:
> Is the rhel6 different because the code there is inferior to the newer code or does this really work differently on rhel6?

@mpreisler it is because only `/etc/modprobe.conf` and `/etc/modprobe.d` are valid in rhel6. In rhel7 and newer, it is all over the place.

PR #2861, opened 2018-05-24
Continue converting XSLT macros to jinja.

matejak:
See #2835 for past PR with description.

This PR is linear, XSLT macros are just replaced by jinja ones, sometimes macros are split.

However, there is an exception with parts where generic macros (e.g. `package_install`) were introduced:
- `product.yml` defined `pkg_system`, which is the package format. It is very useful, so it should be kept, but we also need package manager, which determines the package format. ~Therefore, the `product.yml` files are now also template-processed and `pkg_system` and `pkg_manager` are determined from the user-input of package manager.~ The `pkg_system` key has disappeared from the yaml, but it is added to the parsed yaml dict by the build system.
- ~The current setup is not totally elegant, because authors of rules have to know that they have to pass the `pkg_*` to those generic macros. I was not able to get the code working the elegant way, when the macro "knows" what is the current backend.~
- The generic macros (s.a. `package_install`) can't be handled easily due to the way how jinja was designed. Jinja uses "dumb" caching, so I went for addition of generic macros by aliasing existing specialized macros during the build system processing.


rhel6/guide/system/network/network-uncommon/kernel_module_rds_disabled.rule
```diff
@@ -23,4 +23,5 @@ references:
     srg: SRG-OS-000096
     stigid: RHEL-06-000126
 
-ocil: '<module-disable-check-macro module="rds" />'
+ocil: |
```
redhatrises: @matejak why `|` instead of a single line?

shared/modules/macros.jinja
```diff
@@ -88,3 +88,80 @@ ocil: |
...
+{{%- endmacro %}}
+
+
+{{%- macro firewalld_prevent_access(port, proto, service) %}}
+{{{ _firewalld_check("prevent access", port, proto, service) }}}
+    If <code>firewalld</code> is configured to prevent access, no output will be returned.
+{{%- endmacro %}}
+
+
+{{%- macro firewalld_prevent_desc(port, proto, service) %}}
```
mpreisler: should probably be two macros, one for port/proto and one for service

shared/modules/macros.jinja
```diff
@@ -88,3 +88,80 @@ ocil: |
...
+    {{% endif %}}
+{{%- endmacro %}}
+
+
+{{%- macro firewalld_allow_access(port, proto, service) %}}
+{{{ _firewalld_check("allow access", port, proto, service) }}}
+    If <code>firewalld</code> is configured to allow access through the firewall, something similar to the following will be output:
+    {{% if service %}}
+    If it is a service:
+    <code>&gt;{{{ service }}}</code>
```
jan-cerny: Why is this > here? I know it was there before, but maybe it should not be there.

matejak: I got the impression that it symbolizes beginning of output of the command. I will be happy to take it out though.

shared/modules/macros.jinja
matejak: That would make four macros instead of two, plus the current code allows and meaningfully handles the case when all arguments are supplied. Do we really want another split?

matejak:
@openscap-ci jenkins test this please

rhel6/guide/system/network/network-uncommon/kernel_module_rds_disabled.rule
redhatrises: Go ahead and make a single line.

redhatrises: Go ahead and make a single line.

shared/modules/macros.jinja
mpreisler: maybe not, I didn't read it properly

shared/modules/macros.jinja
```diff
@@ -88,3 +88,229 @@ ocil: |
...
+{{%- macro socket_and_service_disabled_check_with_systemd_complete_ocil_entry(name) %}}
+ocil: |-
+    {{{ xinetd_disabled_check_with_systemd(name) }}}
+    {{{ socket_disabled_check_with_systemd(name) }}}
+
+ocil_clause: "service and/or socket are running"
+{{%- endmacro %}}
+
+
+{{%- macro rpm_package_check(package) %}}
```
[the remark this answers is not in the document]
mpreisler: +100

that would be great we could also remove a lot of description boilerplate that way

shared/modules/macros.jinja
```diff
@@ -88,3 +88,229 @@ ocil: |
...
+    <code>{{{ service }}}</code> service:
+    <pre>$ sudo service {{{ service }}} status</pre>
+    If the service is enabled, it should return the following: <pre>{{{ service }}} is running...</pre>
+{{%- endmacro %}}
+
+
+{{%- macro systemd_socket_disabled(socket) %}}
+    The <code>{{{ socket }}}</code> socket can be disabled with the following command:
+    <pre>$ sudo systemctl disable {{{ socket }}}.socket</pre>
+{{%- endmacro %}}
```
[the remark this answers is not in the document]
matejak: Sure, let's just see the acceptance of the `package_install` and other macros.

debian8/product.yml
```diff
@@ -4,6 +4,6 @@ benchmark_root: "./guide"
 
 profiles_root: "./profiles"
 
-pkg_system: "dpkg"
+{{{ declare_pkg_system_and_manager(pkg_manager="apt-get") }}}
```
redhatrises: Why was this change made? We need to keep a standard schema of `key: jinja_template`.

rhel6/guide/services/dhcp/disabling_dhcp_server/package_dhcp_removed.rule
```diff
@@ -20,4 +20,4 @@ references:
     disa: 366
     nist: CM-7
 
-ocil: '<package-check-macro package="dhcp" />'
+{{{ package_check_complete_ocil_entry(package="dhcp", pkg_system=pkg_system) }}}
```
redhatrises: All `ocil` keys need to remain as `ocil: jinja_template`

debian8/guide/system/logging/package_rsyslog_installed.rule
```diff
@@ -18,4 +18,4 @@ references:
 
 ocil_clause: 'the package is not installed'
 
-ocil: '{{{ rpm_package_check(package="rsyslog") }}}'
+ocil: '{{{ package_check(package="rsyslog", pkg_system=pkg_system) }}}'
```
mpreisler: could we please avoid passing the pkg_system? you should be able to get it from the environment

debian8/product.yml
mpreisler: +1

mpreisler: IMO it's cleaner to just have pkg_system and pkg_manager as separate keys instead of this super long macro name and logic. It's way more self documenting to just have 2 keys

debian8/guide/system/logging/package_rsyslog_installed.rule
redhatrises: +1

rhel6/guide/services/dhcp/disabling_dhcp_server/package_dhcp_removed.rule
mpreisler: @redhatrises as @matejak pointed out earlier we need the ocil_clause and ocil itself, that can't be achieved with the syntax you suggest. You could workaround it but it would be even more confusing and error prone.

redhatrises: Understood. However, `ocil` should be a required key regardless. `ocil_clause` can be an optional key. This brings the content in line with ansible and opencontrol. In addition with the template written as is, there is no way to customize `ocil_clause` which there should be a way to customize it in the template.

debian8/product.yml
matejak: You have missed the reasoning in the updated PR description. Technically, it is OK to violate the DRY principle by saying that if a system uses `yum`, it is RPM-based though - each approach has a weakness.

rhel6/guide/services/dhcp/disabling_dhcp_server/package_dhcp_removed.rule
matejak: There are cases when `ocil_clause` is known, so listing the two here explicitly would violate the DRY principle on a massive scale. Also note that the current template macro hierarchy allows you to keep `ocil` and `ocil_clause` defined explicit by using the `ocil: {{{ package_check(package="dhcp", pkg_system=pkg_system) }}}` syntax.

debian8/guide/system/logging/package_rsyslog_installed.rule
matejak: I would love to. But as I have described in the PR description, I have not managed to get it working. I will give it a shot again, when I think of it, it should work without that argument.

shared/modules/macros.jinja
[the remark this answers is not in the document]
matejak: BTW done, enjoy the `package_check` macro.

matejak:
I consider all suggestions to be addressed, and the PR is therefore ready for the merge review - no other stuff will be added.
I have tried to keep the commit history as clean as possible, so I have used rebases. As a result, some links in the conversation tab may be broken. Please let me know whether this is OK, or you would like me to take another approach next time.

mpreisler:
> Please let me know whether this is OK, or you would like me to take another approach next time.

I think that's OK, it's a trade-off.

debian8/guide/system/permissions/files/permissions_important_account_files/file_permissions_etc_gshadow.rule
```diff
@@ -3,9 +3,9 @@ documentation_complete: true
 title: 'Verify Permissions and ownership on <tt>gshadow</tt> File'
 
 description: |-
-    <fileperms-desc-macro file="/etc/gshadow" perms="0640" />
-      <fileowner-desc-macro file="/etc/gshadow" owner="root" />
-      <filegroupowner-desc-macro file="/etc/gshadow" group="shadow" />
+    {{{ describe_file_permissions(file="/etc/gshadow", perms="0640") }}}
+      {{{ describe_file_owner(file="/etc/gshadow", owner="root") }}}
+      {{{ describe_file_group_owner(file="/etc/gshadow", group="shadow") }}}
```
mpreisler: I love the naming of these functions, it says "describe_*". Could we please follow this pattern with the other macros that are to be used in descriptions? I think it makes it very self documenting and self assuring that you are using the right macro.

mpreisler:
This is a great step forward in SSG upstream, thank you for this PR. I like that you took the opportunity to change how some macros work and what they are called.

Here is a suggestion (it can also be done in a separate PR if you want):
- macros used in descriptions: `describe_*`
- macros used in OCILs: `ocil_*`

What do you think? I think this naming will make it more self documenting and contributors will get the assurance that they are using the right macro.

matejak:
I think that it is OK to rename macros used in `ocil` statements so they are prefixed by `ocil_`. Although they typically describe a setting, coming up with a short, but appropriate name isn't easy, and in SSG, at least almost any description of setting is going to be used in `ocil`.
I can do this rename and documentation in this PR.

sle12/product.yml
```diff
@@ -6,4 +6,4 @@ profiles_root: "./profiles"
 
 init_system: "systemd"
 
-{{{ declare_pkg_system_and_manager(pkg_manager="yum") }}}
+pkg_manager : "yum"
```
mpreisler: space before `:`

also, what is the commit message about?

mpreisler:
> I can do this rename and documentation in this PR.

has this been done or not? github gets a little weird about rebases and doesn't notify me, at least not always

matejak:
@mpreisler It is lots of changes. I will explicitly post when it can be considered done.

matejak:
Macros have been renamed, and the developer documentation has been updated to reflect introduction of the new jinja macro system.

mpreisler:
Great! Big kudos for documentation changes!

PR #2863, opened 2018-05-25
Remove attribute var_check from var_ref element

yuumasato:
#### Description:

- Remove `var_check` attribute from `var_ref` element.

#### Rationale:

- `var_ref` elements cannot have `var_check` attribute

- Fixes SCAP Val validation check


mpreisler:
@openscap-ci test this please

PR #2864, opened 2018-05-25
Sysctl unification2

mpreisler:
static variants of sysctl templates are merged. Had to add a conditional for rhel6 because it's slightly different.

Only IPv6 remains and I will leave that for another PR because there are differences but I am not entirely sure why. It either checks for kernel modules or sysctl settings and I think we should be looking at both - either no kernel module for the ipv6 stack or it's disabled at sysctl level.


mpreisler:
@openscap-ci test this please

yuumasato:
Content generated is the same, only change is `platform` element in metadata, but that is expected.
Ack.

PR #2866, opened 2018-05-28
Unshadow packages_removed WRT RHEL7

yuumasato:
#### Description:

- Removed packages from shared packages_removed csv that already were in RHEL7 csv
- Re-added the packages in products that were relying on shared csv.

#### Rationale:

- This will un-shadow lots of OVAL and fixes for RHEL7 regarding package removed rules.


mpreisler:
ACK

This is awesome, thx!

PR #2873, opened 2018-05-29
Unshadow packages removed WRT RHEL6

yuumasato:
#### Description:

- Removed packages from shared packages_removed csv that already were in RHEL6 csv
- Re-added the packages in products that were relying on shared csv.

#### Rationale:

- This will un-shadow some OVAL and fixes for RHEL6 regarding package_removed rules.

### DIFF
--- a/fedora/templates/csv/packages_removed.csv
+++ b/fedora/templates/csv/packages_removed.csv
@@ -1,3 +1,6 @@
+mcstrans
 net-snmp
 openssh-server
 prelink
+setroubleshoot
+xorg-x11-server-common
--- a/firefox/templates/csv/packages_removed.csv
+++ b/firefox/templates/csv/packages_removed.csv
@@ -1,0 +1,2 @@
+setroubleshoot
+xorg-x11-server-common
--- a/rhel7/templates/csv/packages_removed.csv
+++ b/rhel7/templates/csv/packages_removed.csv
@@ -9,6 +9,7 @@ mcstrans
 net-snmp
 ntpdate
 openldap-servers
+openssh-server
 prelink
 rsh
 rsh-server
--- a/rhel7/templates/csv/packages_removed.csv
+++ b/rhel7/templates/csv/packages_removed.csv
@@ -5,13 +5,18 @@ dhcp
 dovecot
 httpd
 iputils
+mcstrans
 net-snmp
 ntpdate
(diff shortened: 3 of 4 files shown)


mpreisler:
Hmmm, good point @redhatrises 

IMO if the check is for a rule that is defined in shared/guide it should be in shared in some way. Unfortunately right now if it's in shared it's gonna be in every product we build.

yuumasato:
> IMO if the check is for a rule that is defined in shared/guide it should be in shared in some way. 

Makes sense, we should transfer control of what is generated from templates from the CSV files to the rule itself.
The rules in shared already define applicability through `prodtype`, we duplicate this info to the CSV files.

yuumasato:
> Makes sense, we should transfer control of what is generated from templates from the CSV files to the rule itself.
 The rules in shared already define applicability through prodtype, we duplicate this info to the CSV files.

This may be trickier, I just noticed definition for `package_openssh-server_removed`, which we don't have a rule for, but lots of `sshd` checks depend on it.

mpreisler:
@redhatrises brings up a good point, this introduces another level of indirection. IMO if the check is for a rule defined in shared it has to be in shared. Otherwise it will make the final goal - move checks, fixes and rule metdata close together - more difficult. I think having bloated OVALs as a temporary side-effect is OK.

mpreisler:
> The unshadowing should IMO move the other way. product -> shared instead of shared -> multiple products

@yuumasato what do you think? this is waiting for your feedback

mpreisler:
I thought about it some more and I think this will make the final untangle easier. I can easily compare lists between everything that is using the shared linux_os guide. Let's merge this and continue with this, any unshadow helps really.

yuumasato:
@mpreisler For future un-shadowings which approach should we take?
Will it be easier to untangle and unify rules if CSV entries are moved to each product, or to shared? Or it doesn't matter? As long as there is no shadowings?

mpreisler:
> @mpreisler For future un-shadowings which approach should we take?
> Will it be easier to untangle and unify rules if CSV entries are moved to each product, or to shared? Or it doesn't matter? As long as there is no shadowings?

Both have advantages and disadvantages. I think moving CSV entries to products will make it easy to unify in the end. We will have to guarantee that all CSV files of products using the linux_os guide are the same.

PR #2874, opened 2018-05-29
Unified the sysctl_ipv6 templates

mpreisler:
The last of the rhel7 templates unified. I used jinja2 conditional to deal with the 2 cases in the project. I am not sure why we don't check for both kernel module and sysctl but this is not the PR to fix that.


yuumasato:
Looks good, changes in content generated are only on `platform` element of OVAL metadata. Ack.

PR #2875, opened 2018-05-29
Prepare ocp3 build

mpreisler:
We are not building the ocp3 product because it's empty. This doesn't change that fact but it moves the ocp3 source code closer to current state of SSG master. Because it was not being built we didn't do some of the refactoring and we never created a `product.yml` for it.


yuumasato:
Ack.

PR #2877, opened 2018-05-31
Added codacy config

matejak:
Configured the [codacy](https://app.codacy.com/app/projectList/OpenSCAP/scap-security-guide) service in a way that it analyses only valid Python or shell code.

As some of our code is not ready for analysis (e.g. bash snippets that are processed by the build system) and the XML content that is not real code, huge part of the project has to be blacklisted from the checker.


mpreisler:
What is it for? What is the Scrutinizer for?

Could we see some examples of what these two tools find?

matejak:
Sure, did you try to click the Details link in check results?
Those tools are mainly Python analysers (landscape.io seems to be out of order). Their benefit is in identifying weak spots in Python code (bad practices, duplicated code, ...) that can be addressed when one feels surplus and wants to contribute to SSG tech debt reduction.
Scurtinizer doesn't seem to require more than logging in with GitHub, whereas I have created an OpenSCAP organization in Codeacy and adding users to it is based on e-mail addresses. Both checkers have their strong and weak spots, so why not to go for both.

mpreisler:
> Sure, did you try to click the Details link in check results?

Yup, I did but it said no issues found.

matejak:
If you click the details, you will see what the service is about.
As I have written in the description, errors in our code drown in the sea of XML files in SSG, the codacy check needs this blacklist in order to become more focused and recognize that number of issues that SSG has is not negligible given the project size.

mpreisler:
I tried looking more into this but it wants read/write access to all my repos. Not sure if I can work around that. Let's merge this and see what codacy is about, we can remove it later if it isn't useful.

PR #2879, opened 2018-05-31
Ansible service result deprecation warning

mpreisler:
Avoid deprecation warnings such as:
```
[DEPRECATION WARNING]: Using tests as filters is deprecated. Instead of using `result|version_compare` instead use `result is version_compare`. This feature will be removed in version 2.9. 
Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
```


redhatrises:
Ack.

PR #2884, opened 2018-06-01
Add new rule file_groupowner_sapmnt_SID_exe to profile sap on ol7

xiruiyang:
#### Description:

- Add new rule _file_groupowner_sapmnt_SID_exe_ to profile sap on ol7
- Add new OVAL template file _template_OVAL_file_groupowner_ in shared folder
- there is a difference in the .csv file format.  The third column was _alt_name_, but it was not really used, so I think it's safe to redefine it to _pathregex_ at this moment.
- The _file_groupowner.csv_ file uses regex, should it be renamed to _file_groupowner_regex.csv_?
- The individual bash remediation  _file_groupowner_sapmnt_SID_exe.sh_ does not occur in build. There is only the bash file generated from template. Could there be an Enhancement Request that if individual bash file exists, it should be used instead of the one generated from template?

### DIFF
--- a/ol7/profiles/sap.profile
+++ b/ol7/profiles/sap.profile
@@ -17,3 +17,4 @@ selections:
     - package_ypserv_removed
     - var_accounts_authorized_local_users_regex=ol7forsap
     - accounts_authorized_local_users
+    - file_groupowner_sapmnt_SID_exe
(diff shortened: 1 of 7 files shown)


xiruiyang:
Sorry, just noticed I did broken the bash and ansible remediations of _file_groupowner_cron_allow_ (I only searched _file_groupowner_etc_cron_allow_ before hence didn't find out the difference).
Would you like to modify the rule to _file_groupowner_etc_cron_allow_ or shall I rename my code to _regex ?

xiruiyang:
I close this PR and will open a new one after solve the conflict of the third column.

PR #2886, opened 2018-06-01
Unshadowed the disable_prelink remediation.

matejak:
Removed duplicated fix for prelink, went the `bash_remediation_functions` way because it was also part of the FIPS fix.


shared/bash_remediation_functions/disable_prelink.sh
```diff
@@ -0,0 +1,16 @@
+function disable_prelink {
+	# Disable prelinking and don't even check
+	# whether it is installed.
+	if grep -q ^PRELINKING /etc/sysconfig/prelink
+	then
+		sed -i 's/PRELINKING.*/PRELINKING=no/g' /etc/sysconfig/prelink
```
mpreisler: won't this replace comments as well? if I am reading this correctly if you run it twice it will replace the `# Set PRELINKING=no per...` comment iwth just `# PRELINKING=no`.

mpreisler: why is this different form the rhel7 regex? shouldn't we add the comment as well?

shared/bash_remediation_functions/disable_prelink.sh
```diff
@@ -0,0 +1,16 @@
...
+	else
+		printf '\n%s\n' '# Set PRELINKING=no per security requirements' >> /etc/sysconfig/prelink
+		echo 'PRELINKING=no' >> /etc/sysconfig/prelink
+	fi
+
+	# Undo previous prelink changes to binaries if prelink is available.
+	if test -x /usr/sbin/prelink; then
```
mpreisler: I am wonder why you are using the `test` syntax and not the `[` syntax. I know they are the same but the later is more well known to bash devs AFAIK.

shared/bash_remediation_functions/disable_prelink.sh
```diff
@@ -0,0 +1,16 @@
+function disable_prelink {
+	# Disable prelinking and don't even check
+	# whether it is installed.
+	if grep -q ^PRELINKING /etc/sysconfig/prelink
+	then
+		sed -i 's/PRELINKING.*/PRELINKING=no/g' /etc/sysconfig/prelink
+	else
+		printf '\n%s\n' '# Set PRELINKING=no per security requirements' >> /etc/sysconfig/prelink
```
mpreisler: why not add both in one statement? why printf then echo?

mpreisler: also, maybe I am missing something but I don't get the point of the `printf`, it's just putting a string in a string

shared/bash_remediation_functions/disable_prelink.sh
mpreisler: you might want to remove the `/g` too, this regex should replace the last occurrence and only once if I am understanding it correctly

shared/bash_remediation_functions/disable_prelink.sh
matejak: There used to be `echo -e`, which sucks. Printf is safer, but I have rewritten it, so it provokes less.

shared/bash_remediation_functions/disable_prelink.sh
matejak: Probably because `man [` typically won't tell you anything useful, `man test` will go straight to the point.

shared/bash_remediation_functions/disable_prelink.sh
matejak: Good catch, I see that unshadowing requires way more brain activity than one would say at the first glace.

shared/bash_remediation_functions/disable_prelink.sh
```diff
@@ -3,10 +3,10 @@ function disable_prelink {
 	# whether it is installed.
 	if grep -q ^PRELINKING /etc/sysconfig/prelink
 	then
-		sed -i 's/PRELINKING.*/PRELINKING=no/g' /etc/sysconfig/prelink
+		sed -i 's/^PRELINKING[:blank:]*=[:blank:]*[:alpha:]*/PRELINKING=no/' /etc/sysconfig/prelink
 	else
-		printf '\n%s\n' '# Set PRELINKING=no per security requirements' >> /etc/sysconfig/prelink
-		echo 'PRELINKING=no' >> /etc/sysconfig/prelink
+		printf '\n'
```
mpreisler: did you mean printf into the file? this is not redirected to it

matejak: Sure, I have fixed the missing redirect in rebase.

mpreisler: Have you pushed the rebase? It still shows up in the hunk.

matejak: Yes, finally.

shared/bash_remediation_functions/disable_prelink.sh
```diff
@@ -0,0 +1,16 @@
+function disable_prelink {
+	# Disable prelinking and don't even check
+	# whether it is installed.
+	if grep -q ^PRELINKING /etc/sysconfig/prelink
+	then
+		sed -i 's/^PRELINKING[:blank:]*=[:blank:]*[:alpha:]*/PRELINKING=no/' /etc/sysconfig/prelink
+	else
+		printf '\n' >> /etc/sysconfig/prelink
```
mpreisler: I wish this wasn't a separate line but I won't hold up the merge because of it...

mpreisler:
ACK

PR #2891, opened 2018-06-01
Set SLE12 init system to systemd

mpreisler:
SLE 12 comes with systemd [1]

[1] https://www.suse.com/media/white-paper/systemd_in_suse_linux_enterprise_12_white_paper.pdf

init_systemd not being set was causing all sorts of problems with descriptions in the content.


redhatrises:
ack

PR #2893, opened 2018-06-02
Misc OpenStack updates

shawndwells:
Refreshing for OSP12


shared/guide/services/openstack/keystone/keystone_lockout_failure_attempts.var
```diff
@@ -0,0 +1,18 @@
...
+
+title: 'Maximum Number of Failed Authentication Attempts'
+
+description: |-
+    Specifies the maximum number of failed authentication attempts allowed
+    prior to disabling a Keystone user.
+
+type: string
+
+operator: less than or equal
```
shawndwells: Where can I find docs on what valid variable operators are?

PR #2899, opened 2018-06-04
Scenario scripts are now uploaded in bulk.

dahaic:
#### Description:

- Now whole data directory is uploaded as is, so writer of the scenarios
can also create library functions, and import from them to make scripts
more readable and less prone to copy-paste errors.

#### Rationale:

- Previous design was overcomplicating it.

### DIFF
--- a/tests/data/utils.py
+++ b/tests/data/utils.py
@@ -1,6 +1,9 @@
+import os
 import os.path
+import tarfile
 
 _DIR = os.path.dirname(__file__)
+_TAR_BASENAME = 'data.tar'
 _SSG_PREFIX = 'xccdf_org.ssgproject.content_'
 
 
@@ -14,5 +17,23 @@ def iterate_over_rules():
     """
     for dir_name, directories, files in os.walk(_DIR):
         leaf_dir = os.path.basename(dir_name)
+        rel_dir = '.' + dir_name[len(_DIR):]
         if leaf_dir.startswith('rule_'):
-            yield dir_name, _SSG_PREFIX + leaf_dir, files
+            yield rel_dir, _SSG_PREFIX + leaf_dir, files
+
+
+def _exclude_utils(file_name):
+    if file_name in ['./__init__.py', './utils.py']:
+        return True
+    if file_name.endswith('pyc'):
+        return True
+    if file_name.endswith('swp'):
+        return True
+    False
+
+
+def create_tarball(tar_dir):
+    archive_path = os.path.join(tar_dir, _TAR_BASENAME)
+    with tarfile.TarFile.open(archive_path, 'w') as tarball:
+        tarball.add(_DIR, arcname='.', exclude=_exclude_utils)
+    return archive_path
--- a/tests/data/utils.py
+++ b/tests/data/utils.py
@@ -29,7 +29,7 @@ def _exclude_utils(file_name):
         return True
     if file_name.endswith('swp'):
         return True
-    False
+    return False
 
 
 def create_tarball(tar_dir):
--- a/tests/ssg_test_suite/rule.py
+++ b/tests/ssg_test_suite/rule.py
@@ -75,17 +75,15 @@ def _send_scripts(domain_ip):
                                   stdout=log_file,
                                   stderr=subprocess.STDOUT)
         except subprocess.CalledProcessError as e:
-            logging.error("Cannot create directoru {0}.".format(remote_dir))
+            logging.error("Cannot create directory {0}.".format(remote_dir))
             return False
 
-    command = "scp {0} {1}:{2}".format(archive_file, machine, remote_dir)
-    with open(log_file_name, 'a') as log_file:
+        command = "scp {0} {1}:{2}".format(archive_file, machine, remote_dir)
         subprocess.check_call(shlex.split(command),
                               stdout=log_file,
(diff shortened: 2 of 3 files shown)


tests/data/utils.py
```diff
@@ -14,5 +17,25 @@ def iterate_over_rules():
...
+    if file_name.endswith('pyc'):
+        return True
+    if file_name.endswith('swp'):
+        return True
+    False
+
+def create_tarball(tar_dir):
+    archive_path = os.path.join(tar_dir, _TAR_BASENAME)
+    tarball = tarfile.TarFile.open(archive_path, 'w')
+    cwd = os.getcwd()
```
mpreisler: `try finally` please

tests/data/utils.py
```diff
@@ -14,5 +17,25 @@ def iterate_over_rules():
...
+    os.chdir(_DIR)
+    tarball.add('.', exclude=_exclude_utils)
```
mpreisler: why not just pass `_DIR`? or am I missing something? you can use arcname="." for the dir to get the same behavior but it will be reentrant and much easier code

tests/data/utils.py
```diff
@@ -14,5 +17,25 @@ def iterate_over_rules():
...
+        return True
+    if file_name.endswith('pyc'):
+        return True
+    if file_name.endswith('swp'):
+        return True
+    False
+
+def create_tarball(tar_dir):
+    archive_path = os.path.join(tar_dir, _TAR_BASENAME)
+    tarball = tarfile.TarFile.open(archive_path, 'w')
```
mpreisler: use with tarfile.TarFile.open(...) as tarball, it's a nicer and safer pattern because any exception thrown will close the archive and cleanup

dahaic: *facepalm* yeah, I haven't realized I can use it. Thanks.

tests/data/utils.py
dahaic: Good catch, it looks much nicer now.

tests/data/utils.py
```diff
@@ -14,5 +17,23 @@ def iterate_over_rules():
...
+
+
+def _exclude_utils(file_name):
+    if file_name in ['./__init__.py', './utils.py']:
+        return True
+    if file_name.endswith('pyc'):
+        return True
+    if file_name.endswith('swp'):
+        return True
+    False
```
matejak: You probably meant `return False`

tests/ssg_test_suite/rule.py
```diff
@@ -60,27 +60,44 @@ def get_viable_profiles(selected_profiles, datastream, benchmark):
...
+    log_file_name = os.path.join(LogHelper.LOG_DIR, "data.upload.log")
+
+    command = "ssh {0} mkdir -p {1}".format(machine, remote_dir)
+    with open(log_file_name, 'a') as log_file:
+        try:
+            subprocess.check_call(shlex.split(command),
+                                  stdout=log_file,
+                                  stderr=subprocess.STDOUT)
+        except subprocess.CalledProcessError as e:
+            logging.error("Cannot create directoru {0}.".format(remote_dir))
```
matejak: Cannot create **directory**

tests/ssg_test_suite/rule.py
```diff
@@ -60,27 +60,44 @@ def get_viable_profiles(selected_profiles, datastream, benchmark):
...
-    command = "scp {0} {1}:{2}".format(' '.join(scripts_list),
-                                       machine,
-                                       remote_dir)
+    command = "scp {0} {1}:{2}".format(archive_file, machine, remote_dir)
     with open(log_file_name, 'a') as log_file:
         subprocess.check_call(shlex.split(command),
                               stdout=log_file,
                               stderr=subprocess.STDOUT)
 
+    command = "ssh {0} tar xf {1} -C {2}".format(machine, remote_archive_file, remote_dir)
```
matejak: You can continue using that `log_file` object as long as you want, no need to get it again in lines below.

tests/ssg_test_suite/rule.py
```diff
@@ -90,7 +107,7 @@ def _apply_script(rule_dir, domain_ip, script):
     with open(log_file_name, 'a') as log_file:
         log_file.write('##### {0} / {1} #####\n'.format(rule_name, script))
 
-    command = "ssh {0} bash -x {1}".format(machine, script_remote_path)
+    command = "ssh {0} cd {1}; bash -x {2}".format(machine, rule_dir, script)
     with open(log_file_name, 'a') as log_file:
```
matejak: Ditto, just remove this line, realign or relocate the `command` assignment.

tests/ssg_test_suite/rule.py
```diff
@@ -134,7 +151,12 @@ def perform_rule_check(options):
     ssg_test_suite.virt.start_domain(dom)
     domain_ip = ssg_test_suite.virt.determine_ip(dom)
     scanned_something = False
-    for rule_dir, rule, scripts in iterate_over_rules():
+
```
matejak: This function is becoming really complex, I would appreciate extracting some functionality outside of it, but it is not a merge blocker.

dahaic:
@matejak take a look please, I think I fixed all the problems :)

matejak:
Nice one, can't wait to use this new feature!

PR #2900, opened 2018-06-04
Make sure we parse all yaml

mpreisler:
#### Description:

- Tightens up the YAML parser, when there is any unparsed/unconsumed data it will fail the build
- Adds support for the interactive key in Values
- Adds support for the operator key in Values

#### Rationale:

- Makes it possible to notice typos, errors in yaml, unused stuff that people assume is used (operator, interactive, ..)
- Fix regression regarding interactive and operator attributes for XCCDF Values


mpreisler:
Do not merge yet, I will fix interactive and oeprator here as well.

redhatrises:
Ack.

PR #2901, opened 2018-06-04
Yaml constructors

mpreisler:
This doesn't change behavior, it just makes things a bit more predictable and documented. It's good practice to "announce" all member vars in the constructor.

I also avoided the `type` python keyword shadowing.


dahaic:
ACK, thank you.

PR #2902, opened 2018-06-05
Add new tests for rule_auditd_data_retention_num_logs

mildass:
#### Description:

Add new tests for `rule_auditd_data_retention_num_logs`, that check what happens if the value is missing or if the value is above/below limit.


dahaic:
@mildass I am going to do exception and merge this, because we don't want to kill jenkins again by repushing everything :) But please note that in latest build, there is no `common` profile. Tested it with pci-dss and it passes.

And as we discussed, installation of packages is not good practice in functions called "set_parameters_value" or "delete_parameter" :) (No need to change it for every PR, we'll fix it after merging all of them)

dahaic:
ACK, and thank you for the patch!

PR #2903, opened 2018-06-05
Add tests for rule_auditd_data_retention_max_log_file

mildass:
#### Description:

Add new tests for `rule_auditd_data_retention_max_log_file`, that check what happens if the value is missing or if the value is above/below limit.


dahaic:
Same comments as #2902 . Good work, and ACK. :)

PR #2904, opened 2018-06-05
Add tests for rule_auditd_data_retention_max_log_file_action

mildass:
#### Description:

Add tests for `rule_auditd_data_retention_max_log_file_action`, that check what happens when parameter is set to possible values or if the parameter is missing.


mildass:
Updated profile and it should be ok now (only rotate option pass, others fail).

dahaic:
Thank you @mildass ! ACK

PR #2906, opened 2018-06-05
Add tests for rule_auditd_data_retention_space_left_action

mildass:
#### Description:

Add tests for `rule_auditd_data_retention_space_left_action`, that check what happens when parameter is set to possible values or if the parameter is missing.


mildass:
Updated profile and there should be no error in test run now, but not sure if `xccdf_org.ssgproject.content_profile_nist-800-171-cui` profile is ok.

dahaic:
Thanks @mildass , works perfectly :) The profile used is perfectly valid.

PR #2908, opened 2018-06-05
Add tests for rule_auditd_data_retention_admin_space_left_action

mildass:
#### Description:

Add tests for `rule_auditd_data_retention_admin_space_left_action`, that check what happens when parameter is set to possible values or if the parameter is missing.


dahaic:
Thank you @mildass . Works like a charm :)

PR #2909, opened 2018-06-05
Add tests for rule_auditd_data_retention_space_left

mildass:
#### Description:
Add tests and remove/change old tests for `rule_auditd_data_retention_space_left`, that check what happens if the value is missing or if the value is above/below limit.


yuumasato:
LGTM, with same comments as in https://github.com/OpenSCAP/scap-security-guide/pull/2902#issuecomment-395491495. Ack.

PR #2913, opened 2018-06-05
Strict build validation for rule identifiers/CCEs

cipherboy:
#### Description:

- Implements stricter validation of identifiers, refusing to build if validations fail:
  - Refuses to build if CCE is empty
  - Refuses to build if CCE is invalid
- Fixes existing rules to conform to these new standards.

### DIFF
--- a/shared/utils/yaml-to-shorthand.py
+++ b/shared/utils/yaml-to-shorthand.py
@@ -470,7 +470,7 @@ class Rule(object):
         self.rationale = ""
         self.severity = "unknown"
         self.references = {}
-        self.identifiers = []
+        self.identifiers = {}
         self.ocil_clause = None
         self.ocil = None
         self.external_oval = None
@@ -494,7 +494,7 @@ class Rule(object):
         rule.severity = required_yaml_key(yaml_contents, "severity")
         del yaml_contents["severity"]
         rule.references = yaml_contents.pop("references", {})
-        rule.identifiers = yaml_contents.pop("identifiers", [])
+        rule.identifiers = yaml_contents.pop("identifiers", {})
         rule.ocil_clause = yaml_contents.pop("ocil_clause", None)
         rule.ocil = yaml_contents.pop("ocil", None)
         rule.external_oval = yaml_contents.pop("oval_external_content", None)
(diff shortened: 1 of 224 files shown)


shared/utils/yaml-to-shorthand.py
```diff
@@ -494,7 +494,7 @@ def from_yaml(yaml_file, product_yaml=None):
         rule.severity = required_yaml_key(yaml_contents, "severity")
         del yaml_contents["severity"]
         rule.references = yaml_contents.pop("references", [])
-        rule.identifiers = yaml_contents.pop("identifiers", [])
```
mpreisler: it's not clear to me why this change was needed, it seems like defaulting to [] and having the assurance that rule.identifiers is always a list() would be OK and would work with the rest of the code

shared/utils/yaml-to-shorthand.py
```diff
@@ -508,8 +508,27 @@ def from_yaml(yaml_file, product_yaml=None):
             raise RuntimeError("Unparsed YAML data in '%s'.\n\n%s"
                                % (yaml_file, yaml_contents))
 
+        if not rule.validate_identifiers(yaml_file):
```
mpreisler: this will never return False :-) Just running it is fine.

shared/utils/fix-cces.py
```diff
@@ -0,0 +1,168 @@
...
+    elif sys.argv[1] == 'invalid':
+        find_invalid_cce(sys.argv[2])
+    else:
+        print("Usage: %s mode /full/path/to/src/directory" % sys.argv[0])
+        print("Modes:")
+        print("\tempty - check and fix rules with empty identifiers")
+        print("\tinvalid - find rules with invalid identifiers")
+
+if __name__ == "__main__":
+    __main__()
```
mpreisler: hmm, this is a cool script and might help us write other scripts but on the other hand it is a one-off, I don't expect anyone to ever run it again because build system now forbids invalid identifiers so there is no way to introduce them

shared/utils/yaml-to-shorthand.py
mpreisler: discussed this with @cipherboy , turns out it needs to be a dict()

cipherboy:
Shorter PEP8 violation list now. :)

shared/utils/yaml-to-shorthand.py
```diff
@@ -508,8 +508,25 @@ def from_yaml(yaml_file, product_yaml=None):
             raise RuntimeError("Unparsed YAML data in '%s'.\n\n%s"
                                % (yaml_file, yaml_contents))
 
+        rule.validate_identifiers(yaml_file)
         return rule
 
+    def validate_identifiers(self, yaml_file):
+        if self.identifiers is None:
```
mpreisler: this will never happen

cipherboy: @mpreisler Updated

mpreisler:
Crap, I merged the other PR and it created a conflict. This needs a rebase now.

cipherboy:
@mpreisler Rebased.

mpreisler:
ACK

PR #2914, opened 2018-06-05
Make Rule.references a dict

cipherboy:
In the build system, Rules's references element is initialized to a list despite being a dictionary in the YAML. This fixes it to be a dictionary.


mpreisler:
ACK

PR #2915, opened 2018-06-05
Replace landscape.io badge with scrutinizer

mpreisler:
landscape.io stopped working a while back and is showing outdated
scores.


matejak:
ACK.

PR #2917, opened 2018-06-05
Strict handling of identifiers and references

cipherboy:
#### Description:

- Enforces `identifiers` and `references` sections to be non-empty E.g.: the following is now invalid:
```
references:

something_else:
    - value
```
- Enforces `identifiers` and `references` values to be strings. E.g.: the following:
```
references:
   - disa: 1234
```
Becomes:
```
references:
   - disa: "1234"
```
- All existing violating rules have been rewritten.

#### Rationale:
- This simplifies the build system in several places, allows us to enforce types on `Rules.references` and `Rules.identifiers` and their contained objects.


cipherboy:
Per @pep8speaks:
> Line 112:13: E722 do not use bare except'

This was intentional as my pseudo-build parser can't handle the new macros like `'{{{ ocil_service_enabled(service="firewalld") }}}'` (e.g., this [file](https://github.com/OpenSCAP/scap-security-guide/blob/master/linux_os/guide/system/network/network-firewalld/firewalld_activation/service_firewalld_enabled.rule)) for some reason. Since there were only a few like that, I opted to ignore it and let the build fail it and manually edit those. I'm likely missing some logic around `product.yml` handling, but that is in the `misc/fix-rules.py` script, not the build system so it is fine.

mpreisler:
@openscap-ci test this please

mpreisler:
ACK

PR #2920, opened 2018-06-06
Add tests for rule_auditd_audispd_encrypt_sent_records

mildass:
#### Description:
Add tests for `rule_auditd_audispd_encrypt_sent_records`, which test if the encrypt is (or isn't) activated and if parameter isn't defined.


dahaic:
Same comment as in #2902 . Plus one more point to fix - this rule has no check, nor remediations written, thus specifying `remediation =  bash` does not make much sense. (It will fail also fail, as result "notchecked" is not what is expected, but I guess we should leave that as deliberate)

PR #2921, opened 2018-06-06
Add tests for rule_auditd_audispd_configure_remote_server

mildass:
#### Description:
Add tests for `rule_auditd_audispd_configure_remote_server`, which test what happend when the parametr isn't defined or when is defined as numeric adress or hostname.


matejak:
ACK. There is that issue with installing package inside the function. Note for reference: The `set_parameters_value` function is not duplicated across similar PRs, as it is the same file with the same contents.

PR #2922, opened 2018-06-06
Add new rule file_owner_sapmnt_SID_exe to profile sap on ol7 

xiruiyang:
#### Description:
- Add new rule file_owner_sapmnt_SID_exe to profile sap on ol7 
- Add new OVAL template for verifying directory/file ownership
- Add function to generate OVAL from template for verifying directory/file ownership
- The existing csv format "PATH, USER, [ALT_ANME]" is extended with "[directory|file], [PATHREGEX]"
- Limitation: it only works when each OS/VM only has one SAP System ID (SID) and hence one sidadm. I may take uid of sidadm users from /etc/passwd into a variable, but it looks the file_test will compare user_id with each item in the variable and hope it equals to each of them. The result if false if there are two uids from /etc/passwd that get into the variable.  Any advise on how to deal with this situation?

#### Rationale:

- I was planning to merge permissions, owner and groupowner to one rule access_privileges but am hesitated because  the limitation on file_owner. 

The file_permissions and file_groupowner does not have the above issue. Both are ready to create pull request if they do not need to get into one rule access_privileges. If you want to have a look:
https://github.com/xiruiyang/scap-security-guide/tree/permissions
https://github.com/xiruiyang/scap-security-guide/tree/groupowner
or I may create separate pull request for them?

### DIFF
--- a/linux_os/guide/system/software/sap/file_owner_sapmnt_SID_exe.rule
+++ b/linux_os/guide/system/software/sap/file_owner_sapmnt_SID_exe.rule
@@ -1,0 +1,30 @@
+documentation_complete: true
+
+prodtype: ol7
+
+title: 'Owner of Directory /sapmnt/SID/exe is sidadm'
+
+description: |-
+    The directory <tt>/sapmnt/SID/exe</tt> should be owned by <tt>sidadm</tt>.
+    <tt>SID</tt> is SAP System ID which is always three alphanumeric characters
+    beginning with an alphabetic character. <tt>SID</tt> is in uppercase while the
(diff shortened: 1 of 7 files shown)


mpreisler:
Please rebase this against upstream master to get rid of the merge commits.

mpreisler:
@xiruiyang for reference, you don't have to close and reopen, you can do something like this:
```
git checkout master
git pull upstream master --ff-only
git checkout $feature_branch
git rebase -i master
git push --force origin $feature_branch
```

This will update github and jenkins and everything.

xiruiyang:
Will open a new PR later. Thanks.

PR #2923, opened 2018-06-06
Unshadow all the CPE OVAL checks

mpreisler:
#### Description:

- Deduplicates OVAL code related to CPE checks, all of this is universal, you are checking whether the platform is something, so the check itself applies to everything, not just the target platform

#### Rationale:

- Less code is better
- Brings bugfixes to shared CPE OVALs to the products that copied the code and shadowed it, this fixes CPE OVAL on RHEL-OSP7, Debian and others.


matejak:
@openscap-ci test this please

matejak:
@mpreisler Jenkins test fails for legitimate reasons, please take a look at it.

mpreisler:
rhel-osp7 is a problematic name, there must be parts of the build system that can't handle it

shawndwells:
Iirc isn’t it just called “osp” now? The RHEL preface is old

mpreisler:
Hmm, so the problem is that once we pull all the multi platform all checks into chromium and firefox it uncovers hidden dependencies and these 2 projects now fail. There are 2 options, either I can fix all the multi_platform_all stuff or add Chromium and Firefox to the CPE OVALs :-/

mpreisler:
the debian8 CPE OVAL was unshadowed in #2966 
removed that commit from this PR

mpreisler:
I have applied workarounds and this now builds on my machine and the test suite passes.

yuumasato:
> I can fix all the multi_platform_all stuff 

@mpreisler by this you meant to expand multi_platform_all to Chromium and Firefox, or something else?

Nice workaround, this way Chromium and Firefox are not added to all multiplatform checks.
ACK from me.

Will wait for some more feedback before merging.

mpreisler:
> @mpreisler by this you meant to expand multi_platform_all to Chromium and Firefox, or something else?

Yes but that uncovers other issues. Probably should be done in another PR.

redhatrises:
Hmmm..... Chromium and Firefox aren't necessarily part of the unix family. Do they need to have that?

mpreisler:
> Hmmm..... Chromium and Firefox aren't necessarily part of the unix family. Do they need to have that?

that is true but some of the checks extend it and the criteria differs depending on whether you are on unix family or not.

yuumasato:
> Hmmm..... Chromium and Firefox aren't necessarily part of the unix family. Do they need to have that?

Maybe `installed_app_is_firefox` should be called `installed_app_is_unix_firefox`?

Also, CPE checks kinda need to be run everywhere. A CPE check for rhel7 should be able to run in an debian and return false.

mpreisler:
> Also, CPE checks kinda need to be run everywhere. A CPE check for rhel7 should be able to run in an debian and return false.

yes, 100%

yuumasato:
@redhatrises Do you have objections to the workaround?

redhatrises:
Changing `installed_app_is_firefox` to `installed_app_is_unix_firefox`? If that is what you are referring to then, yes, that makes no sense to me since we have an extended_definition that checks if it is part of the unix family.

yuumasato:
@redhatrises Sorry I was not clear. I was refering to commit b8bdebd.
I think renaming of the OVAL should be done in another PR.

redhatrises:
@yuumasato no that is fine.

PR #2916, opened 2018-06-05
Sensible support for oval 5.11

mpreisler:
#### Description:

- Much simplified the system we need to support both OVAL 5.10 and 5.11
- Some products will have extraneous OVAL defs when built for OVAL 5.10 if those defs are only necessary for OVAL 5.11 to work
- The build_config.yml mechanism is universal and we can use it to pass anything we know at configure time (from cmake or user) to jinja2 templates or even other things. This can enable simpler handling of other arguments, we don't have to keep adding command line arguments to the build scripts.

#### Rationale:

- This was a huge mess and very hard to explain to other people
- Extraneous OVAL defs are not a big deal

PR #2924, opened 2018-06-06
[WIP] Unified ubuntu debian guides

mpreisler:
#### Description:

- Remove duplicate code, make things simpler
- This will not build, it relies on #2916 and that has not been merged yet
- This is not the final solution. We want ubuntus and debian to use linux_os/guide. This is just a step in that direction. Now we have to merge debian8 guide into linux_os, previously we had to merge 3 different guides into it.


mpreisler:
Look at that diffstat ;-)

matejak:
Needs a rebase.

PR #2930, opened 2018-06-07
Use the C implementation of SafeLoader instead of pure python

mpreisler:
It's way faster...

```
times for a full SSG build using ninja

python SafeLoader
real    1m45.529s

CSafeLoader
real    1m29.529s
```

### DIFF
--- a/shared/modules/ssgcommon.py
+++ b/shared/modules/ssgcommon.py
@@ -2,10 +2,15 @@ import datetime
 import platform
 import subprocess
 import re
-import yaml
 import codecs
 import jinja2
 import os.path
+import yaml
+
+try:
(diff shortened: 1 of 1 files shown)


cipherboy:
LGTM; tests pass locally and I see a similar-ish speedup under make too.

mpreisler:
I have another optimization - jinja2 bytecode cache. I have written the patch but can't submit yet because it's waiting for the sensible OVAL 5.11 PR. Without build_config.yml I can't determine where the jinja2 bytecode cache should be.

mpreisler:
The errors are pretty much false positives and again, this is blocking other PRs... Plus it will make the tests quicker...

PR #2905, opened 2018-06-05
Continue converting XSLT macros to jinja (part 2)

matusmarhefka:
See the #2835 PR for rationale and description.

PR #2931, opened 2018-06-07
Optionally cache jinja2 bytecode into files in the user provided folder

mpreisler:
Defaults to OFF, consider this experimental, this saves ~12 seconds on
my machine for a full build with ninja.

The bytecode takes roughly 80 MB for a full SSG build which is quite a bit... That's why I am reluctant to just enable this by default for everyone.


CMakeLists.txt
```diff
@@ -97,6 +98,13 @@ if (SSG_SVG_IN_XCCDF_ENABLED AND NOT OSCAP_SVG_SUPPORT_RESULT EQUAL 0)
     message(WARNING "Your version of OpenSCAP does not support having the SVG logo in the XCCDF, disabling SVG logo.")
 endif()
 
+if (SSG_JINJA2_CACHE_ENABLED)
+    file(MAKE_DIRECTORY "${CMAKE_BINARY_DIR}/jinja2_cache")
+    set(SSG_JINJA2_CACHE_DIR "${CMAKE_BINARY_DIR}/jinja2_cache")
```
cipherboy: I'd make this configurable and default it to the build subdirectory if it is not set when `SSG_JINJA2_CACHE_ENABLED=true`. If I understand how they're implementing the [cache](http://jinja.pocoo.org/docs/2.10/api/#bytecode-cache) (and the [bucket](http://jinja.pocoo.org/docs/2.10/api/#jinja2.bccache.Bucket)), it is safe to use a global cache. That way, on the jenkins workers, we can use a shared global cache across all PRs, improving all build times.

(I'll mock these assumptions tomorrow to validate they hold).

mpreisler: That's a pretty cool idea. Keep in mind though that jinja2 will never clean any bytecode, we would have to set-up some periodic cache purging to avoid it taking all disk space.

mpreisler: I have changed the option to a PATH cmake variable. Now you can set it to any directory.

cipherboy:
@mpreisler I think this saves ~50 seconds when you consider building it without a cache vs. with a primed cache. Here's my times for building:

Jinja cache (primed):
real	1m23.303s
user	4m11.533s
sys	0m39.044s

Jinja cache (not primed):
real	1m36.642s
user	4m15.568s
sys	0m40.121s

No Jinja cache:
real	2m10.830s
user	5m42.066s
sys	0m45.276s

I performed the following additional testing to emulate CI using it:
- Cloned this branch locally, built and tested. 
- Cloned "remove TBDs" and applied it on top of your branch, built and tested. 
- Cloned #2905, applied it on top of this merged branch, built and tested it.

In this case, cache seems to be working as expected. My cache is currently 155MB, but I've built from two different build directories (~/GitHub/mpreisler/ssg, ~/GitHub/cipherboy/ssg).

CMakeLists.txt
```diff
@@ -29,6 +29,8 @@ option(SSG_OVAL_SCHEMATRON_VALIDATION_ENABLED "If enabled, schematron validation
 option(SSG_SHELLCHECK_BASH_FIXES_VALIDATION_ENABLED "If enabled, shellcheck validation of bash fixes will be performed as part of the ctest tests. Shellcheck tests don't pass right now, this option is discouraged until that's fixed." FALSE)
 option(SSG_LINKCHECKER_VALIDATION_ENABLED "If enabled, linkchecker will be used to validate URLs in all the HTML guides and tables." TRUE)
 option(SSG_SVG_IN_XCCDF_ENABLED "If enabled, the built XCCDFs will include the SVG SCAP Security Guide logo." TRUE)
+option(SSG_JINJA2_CACHE_ENABLED "If enabled, the jinja2 templating files will be cached into bytecode. Also see SSG_JINJA2_CACHE_DIR." TRUE)
+set(SSG_JINJA2_CACHE_DIR "${CMAKE_BINARY_DIR}/jinja2_cache" CACHE PATH "Where should the jinja2 cached bytecode be stored. This speeds up builds at the expense of disk space. You can use one location for multiple SSG builds if necessary.")
```
cipherboy: `s/if necessary/for performance improvements/`
`s/Where should the jinja2 cached bytecode be stored/Where the jinja2 cached bytecode should be stored/`

mpreisler: Fixed, thx!

cipherboy:
ACK

PR #2932, opened 2018-06-08
Remove tbds

cipherboy:
- Updates build system to prevent invalid product CCEs as well as primary CCEs
- Removes prefixes from CCEs which have an internal 'cce' prefix.
- Removes CCEs with 'TDB'


cipherboy:
:) Better pep8 while I'm at it.

mpreisler:
ACK. Great!

PR #2934, opened 2018-06-08
Purge multi platform oval

mpreisler:
For OVAL checks with `multi_platform_all`:

We broke the multi platform oval support so now for all defs it shows just this:
```
        <ns0:affected family="unix">
        </ns0:affected>
```

With this change it will show this:
```
        <ns0:affected family="unix">
          <ns0:platform>Oracle Linux 7</ns0:platform>
          <ns0:platform>Wind River Linux 8</ns0:platform>
          <ns0:platform>Ubuntu 1404</ns0:platform>
          <ns0:platform>Ubuntu 1604</ns0:platform>
          <ns0:platform>Debian 8</ns0:platform>
          <ns0:platform>Red Hat Enterprise Linux 7</ns0:platform>
          <ns0:platform>Red Hat Enterprise Linux 6</ns0:platform>
          <ns0:platform>Fedora 26</ns0:platform>
          <ns0:platform>Fedora 27</ns0:platform>
          <ns0:platform>Fedora 28</ns0:platform>
        </ns0:affected>
```

### DIFF
--- a/oval.config.in
+++ b/oval.config.in
@@ -21,7 +21,7 @@
 # Note: this file uses .ini style formatting
 #
 [Platform]
-multi_platform_all = multi_platform_fedora, multi_platform_rhel, multi_platform_debian, multi_platform_ubuntu, multi_platform_wrlinux, multi_platform_ol
+multi_platform_all = multi_platform_fedora,multi_platform_rhel,multi_platform_debian,multi_platform_ubuntu,multi_platform_wrlinux,multi_platform_ol
 multi_platform_fedora = 28,27,26
 multi_platform_debian = 8
 multi_platform_ubuntu = 1604,1404
--- a/oval.config.in
+++ b/oval.config.in
@@ -21,7 +21,7 @@
 # Note: this file uses .ini style formatting
 #
 [Platform]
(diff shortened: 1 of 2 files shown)


redhatrises:
@mpreisler agreed.... of course, it would be nice to also add that to this PR.

mpreisler:
@redhatrises is there any reason to even have it in the final OVAL? AFAIK it's not used and nobody complained about empty `<affected>` elements so far.

mpreisler:
I am thinking maybe we can just remove <affected> from the final OVAL altogether.

redhatrises:
I would keep them as we don't really know who uses them and who doesn't. It would also be handy for future tools that might need to determine what platform is supported.

yuumasato:
> I am thinking maybe we can just remove <affected> from the final OVAL altogether.

Removing affected element may be a better approach. When no `affected` elements exist, everything is applicable. From OVAL spec:
> 4.3.5 AffectedType
The AffectedType is a container type for the list of affected platforms and products. Note that the
absence of a platform or product implies that the OVAL Definition applies to all platforms or products.

yuumasato:
> Removing affected element may be a better approach.

I take it back.

> When no affected elements exist, everything is applicable. 

Actually, `platform` or `products`.

`metadata` is allowed to have zero `affected` elements, but it is not explicit what it is implied from that.

redhatrises:
Ack. But IMO for supportability, we should keep `affected` elements.

PR #2935, opened 2018-06-08
Silence warnings about unknown target from templates when the target …

mpreisler:
…is actually supported

Don't spit out unknown target warnings when the template just doesn't support it.

### DIFF
--- a/shared/templates/template_common.py
+++ b/shared/templates/template_common.py
@@ -16,10 +16,11 @@ class ActionType:
 
 
 class UnknownTargetError(ValueError):
-    def __init__(self, msg):
+    def __init__(self, lang):
         super(UnknownTargetError, self).__init__(
-            "Unknown target language: \"{0}\"".format(msg)
+            "Unknown target language: \"{0}\"".format(lang)
         )
+        self.lang = lang
 
 
 class TemplateNotFoundError(RuntimeError):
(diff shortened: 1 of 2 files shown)


shared/templates/template_common.py
```diff
@@ -206,7 +207,12 @@ def csv_map(self, filename, language):
 
             except UnknownTargetError as e:
                 if self.action == ActionType.BUILD:
-                    sys.stderr.write(str(e) + "\n")
+                    # If lang is one of these it just means that the template
+                    # doesn't implement that language. It doesn't mean that the
+                    # target is invalid.
+                    if e.lang not in ["oval", "ansible", "bash", "puppet",
```
cipherboy: Are we likely to add new build targets sometime soon? It looks like these are hard coded in a bunch of places; you might want to document somewhere that this is another location that'll need to be updated if a new target is added.

mpreisler: yeah, should probably be somewhere in ssgcommon.py. I will look into that.

mpreisler: It's used in one other spot. I have put the list into a variable in template_common, it seems to fit well there. It's reused in generate-from-templates.py.

dahaic:
Build logs are much nicer now, thanks @mpreisler !

PR #2936, opened 2018-06-08
Output just one line with all affected profiles when unselecting empt…

mpreisler:
…y groups

Makes the output shorter, easier to spot issues.


mpreisler:
Example of output:
```
Unselected empty groups in anssi_np_nt28_restrictive, anssi_np_nt28_minimal, standard, anssi_np_nt28_average, anssi_np_nt28_high.
```

redhatrises:
ack

PR #2938, opened 2018-06-11
Add new rule accounts_authorized_local_users_sidadm_orasid to profile sap on ol7

xiruiyang:
#### Description:

- Only user accounts _sidadm_, _orasid_, _sapadm_, _oracle_ and authorized users in an external list are allowed on the OS/VM
- Individual test of existence of SAP users _sidadm_, _orasid_, _sapadm_ and Oracle user _oracle_ according to SAP system ID and installed SAP/Oracle software components
- Works when there is exactly one SAP system on each OS/VM

### DIFF
--- a/linux_os/guide/system/software/sap/accounts_authorized_local_users_sidadm_orasid.rule
+++ b/linux_os/guide/system/software/sap/accounts_authorized_local_users_sidadm_orasid.rule
@@ -1,0 +1,49 @@
+documentation_complete: true
+
+prodtype: ol7
+
+title: 'Only sidadm and orasid/oracle User Accounts Exist on Operating System'
+
+description: |-
+    SAP tends to use the server or virtual machine exclusively. There should be only
+    SAP system users <tt>sidadm</tt> and <tt>orasid</tt> that exist on the operating
+    system (or virtual machine). If SAP Host Agent is installed, the user <tt>sapadm</tt>
+    must exist too. With Oracle Database using <tt>oracle</tt> user, the user <tt>oracle</tt>
+    should exist as well. 
+    While <tt>SID</tt> is the SAP System ID, which is always three alphanumeric characters
+    in upper case, beginning with an alphabetic character, the user names <tt>sidadm</tt> 
+    and <tt>orasid</tt> are in lower case.
+    Besides the above SAP users that are automatically detected, other operating system
+    users can be customized in the refine value variable
+    <tt>var_accounts_authorized_local_users_regex</tt>.
+    OVAL regular expression is used for the user list.
+    <tt>Limitation: Currently this OVAL test only works when there is exactly one SAP
+    system on each operating system (or virtual machine). </tt> The SAP system users
+    <tt>sidadm</tt>, <tt>orasid</tt>, <tt>sapadm</tt> and <tt>oracle</tt> can be automatically
+    detected. If there is no or multiple SAP systems installed on the operating system
+    (or virtual machine), the test will fail. In this case, please use the general purpose
+    test <tt>accounts_authorized_local_users</tt> and customize the refine value variable
+    <tt>var_accounts_authorized_local_users_regex</tt> by adding all the authorized
+    user names to the list.
+    Test result of both <tt>false</tt> or <tt>error</tt> means mismatch of user names and
+    SAP system. Nevertheless, the bash remediation commands can always be used to delete
+    unexpected users on the operating system. It works in all cases zero, one and multiple
+    SAP systems on the operating system (or virtual machine).
+
+rationale: |-
+    Accounts providing no operational purpose provide additional opportunities for
+    system compromise. Unnecessary accounts include user accounts for individuals not
+    requiring access to the system and application accounts for applications not installed
+    on the system.
+
+severity: medium
+
+ocil_clause: 'there are unauthorized local user accounts on the system'
+
+ocil: |-
+    To verify that there are no unauthorized local user accounts, run the following command:
+    <pre>$ less /etc/passwd </pre>
+    Inspect the results, and if unauthorized local user accounts exist, remove them by
+    running the following command:
+    <pre>$ sudo userdel <i>unauthorized_user</i></pre>
+
--- a/linux_os/guide/system/software/sap/accounts_authorized_local_users_sidadm_orasid.rule
+++ b/linux_os/guide/system/software/sap/accounts_authorized_local_users_sidadm_orasid.rule
@@ -9,23 +9,16 @@ description: |-
     SAP system users <tt>sidadm</tt> and <tt>orasid</tt> that exist on the operating
     system (or virtual machine). If SAP Host Agent is installed, the user <tt>sapadm</tt>
     must exist too. With Oracle Database using <tt>oracle</tt> user, the user <tt>oracle</tt>
-    should exist as well. 
-    While <tt>SID</tt> is the SAP System ID, which is always three alphanumeric characters
-    in upper case, beginning with an alphabetic character, the user names <tt>sidadm</tt> 
-    and <tt>orasid</tt> are in lower case.
+    should exist as well. While <tt>SID</tt> is the SAP System ID, which is always
+    three alphanumeric characters in upper case, beginning with an alphabetic character,
+    the user names <tt>sidadm</tt> and <tt>orasid</tt> are in lower case.
+    <br /> <br />
     Besides the above SAP users that are automatically detected, other operating system
     users can be customized in the refine value variable
     <tt>var_accounts_authorized_local_users_regex</tt>.
     OVAL regular expression is used for the user list.
-    <tt>Limitation: Currently this OVAL test only works when there is exactly one SAP
-    system on each operating system (or virtual machine). </tt> The SAP system users
-    <tt>sidadm</tt>, <tt>orasid</tt>, <tt>sapadm</tt> and <tt>oracle</tt> can be automatically
-    detected. If there is no or multiple SAP systems installed on the operating system
-    (or virtual machine), the test will fail. In this case, please use the general purpose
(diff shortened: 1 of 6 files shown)


dahaic:
Hello @xiruiyang  thank you for the PR. Have you tried to use SSG Test Suite? Right now, as I am not well versed in the things SAP, I can only review it from perspective of SSG build system. I.e. if OVAL is well made, there are no obvious flaws, etc. By providing test scenarios, you can more explicitly describe situations this OVAL (and remediations) should handle. Those would probably be easier to understand, then just OVAL.

But it's not something we have to do for this PR, if you are ok that the content itself will be only sanity-checked by me. :)

shared/fixes/bash/accounts_authorized_local_users_sidadm_orasid.sh
```diff
@@ -0,0 +1,56 @@
...
+oracle=false
+SIDlist=$(find /sapmnt -regex '^/sapmnt/[A-Z][A-Z0-9][A-Z0-9]/exe/brspace$' \
+	-o -regex '^/sapmnt/[A-Z][A-Z0-9][A-Z0-9]/exe/uc/[a-z0-9_]+/brspace$' \
+	-o -regex '^/sapmnt/[A-Z][A-Z0-9][A-Z0-9]/exe/nuc/[a-z0-9_]+/brspace$')
+for i in $SIDlist ; do
+	if [ $(ls -ld $i | awk '{print $3}') == "oracle" ]; then
+		oracle=true
+	fi 	
+done
+if $oracle == true ; then
```
matejak: Please use 
`if test "$oracle" = true`
i.e. use `test` or square brackets, double-quote `$oracle`, and use just one equals for greater portability.

xiruiyang:
@dahaic thank you for checking the PR and for your advice about SSG Test Suite.  I will try it for the OVAL tests in the future.  :+1:

shared/fixes/bash/accounts_authorized_local_users_sidadm_orasid.sh
```diff
@@ -0,0 +1,56 @@
...
+
+# add users sidamd and orasid
+userlist="root"
+
+# if /sapmnt/SID exists, add sidadm to the list
+SIDlist=$(find /sapmnt -regex '^/sapmnt/[A-Z][A-Z0-9][A-Z0-9]$')
+for i in $SIDlist ; do
+	SID=${i:8:3}
+	userlist="$userlist|$(echo "$SID" | sed -e 's/\(.*\)/\L\1/')adm"
+done
```
matejak: Your remediations are isolated tasks, I suggest that you make those tasks functions that you just invoke later.
One additional advantage is that name of the function may even make some comments redundant.

shared/fixes/bash/accounts_authorized_local_users_sidadm_orasid.sh
```diff
@@ -0,0 +1,56 @@
...
+
+# if /sapmnt/SID/exe/brspace or /sapmnt/SID/exe/<type>/<platform>/brspace exists,
+# add orasid to the list
+SIDlist=$(find /sapmnt -regex '^/sapmnt/[A-Z][A-Z0-9][A-Z0-9]/exe/brspace$' \
+	-o -regex '^/sapmnt/[A-Z][A-Z0-9][A-Z0-9]/exe/uc/[a-z0-9_]+/brspace$' \
+	-o -regex '^/sapmnt/[A-Z][A-Z0-9][A-Z0-9]/exe/nuc/[a-z0-9_]+/brspace$')
+for i in $SIDlist ; do
+        SID=${i:8:3}
```
matejak: Please try to change naming here - if I have a SIDlist, then I expect one element of it to be SID. However, SID is just part of element of SIDlist, which is somehow confusing.

shared/fixes/bash/accounts_authorized_local_users_sidadm_orasid.sh
```diff
@@ -0,0 +1,56 @@
...
+
+# if /usr/sap/hostctrl exists, add sapadm to the list
+if [ -d /usr/sap/hostctrl ] ; then
+	userlist="$userlist|sapadm"
+fi
+
+# if owner of any /sapmnt/SID/exe/brspace or /sapmnt/SID/exe/<type>/<platform>/brspace 
+# file is oracle, add oracle to the list
+oracle=false
+SIDlist=$(find /sapmnt -regex '^/sapmnt/[A-Z][A-Z0-9][A-Z0-9]/exe/brspace$' \
```
matejak: We have a duplication with line 20, could you extract this `find` into a bash function?

shared/fixes/bash/accounts_authorized_local_users_sidadm_orasid.sh
```diff
@@ -0,0 +1,56 @@
...
+	-o -regex '^/sapmnt/[A-Z][A-Z0-9][A-Z0-9]/exe/uc/[a-z0-9_]+/brspace$' \
+	-o -regex '^/sapmnt/[A-Z][A-Z0-9][A-Z0-9]/exe/nuc/[a-z0-9_]+/brspace$')
+for i in $SIDlist ; do
+	if [ $(ls -ld $i | awk '{print $3}') == "oracle" ]; then
```
matejak: Please prefer `=` over `==`, both work, the former is just more portable.

shared/fixes/bash/accounts_authorized_local_users_sidadm_orasid.sh
```diff
@@ -0,0 +1,56 @@
...
+if $oracle == true ; then
+	userlist="$userlist|oracle"
+fi 
+
+# delete users that is in /etc/passwd but neither in the userlist
+# nor in default_os_user nor in the var_accounts_authorized_local_users_regex
+default_os_user=^$default_os_user$
+userlist=^$userlist$
+for username in $( sed 's/:.*//' /etc/passwd ) ; do
+	if [[ ! "$username" =~ ($default_os_user|$userlist|$var_accounts_authorized_local_users_regex) ]]; then
```
matejak: This is slightly mind bending, could you please consider using `grep`? Something like
`printf "%s" "$username" | grep -q -e "$default_os_user" -e "$userlist" -e "$var_accounts_authorized_local_users_regex"`. People use `grep` all the time, but exceptions apply to the syntax in `[[` statements, which makes people uncertain.

redhatrises:
@xiruiyang what is the difference between this rule and `accounts_authorized_local_users.rule`? Seems like they should be integrated.

xiruiyang:
@matejak Improvements were made according to your recommendations. Thanks a lot. :+1:  The duplication _find_ command was deleted. The last one was not modified since the variable var_..._regex may include regex, it does not work with grep.

xiruiyang:
@redhatrises This rule is only for OS/VM that has exactly one SAP system on it.  The other rule _account_authorized_local_users.rule_ is for general purpose, not only for SAP. If any system (with or without any specific software) has the requirement that only authorized users may exist in /etc/passwd, the other rule may be used.

linux_os/guide/system/software/sap/accounts_authorized_local_users_sidadm_orasid.rule
```diff
@@ -0,0 +1,49 @@
...
+    must exist too. With Oracle Database using <tt>oracle</tt> user, the user <tt>oracle</tt>
+    should exist as well. 
+    While <tt>SID</tt> is the SAP System ID, which is always three alphanumeric characters
+    in upper case, beginning with an alphabetic character, the user names <tt>sidadm</tt> 
+    and <tt>orasid</tt> are in lower case.
+    Besides the above SAP users that are automatically detected, other operating system
+    users can be customized in the refine value variable
+    <tt>var_accounts_authorized_local_users_regex</tt>.
+    OVAL regular expression is used for the user list.
+    <tt>Limitation: Currently this OVAL test only works when there is exactly one SAP
```
dahaic: I feel it would make the prose more readable, if you put the Limitations part into `warning` style. See how to use it https://github.com/OpenSCAP/scap-security-guide/blob/master/linux_os/guide/services/ssh/ssh_server/sshd_disable_rhosts_rsa.rule

linux_os/guide/system/software/sap/accounts_authorized_local_users_sidadm_orasid.rule
```diff
@@ -0,0 +1,49 @@
...
+    system on each operating system (or virtual machine). </tt> The SAP system users
+    <tt>sidadm</tt>, <tt>orasid</tt>, <tt>sapadm</tt> and <tt>oracle</tt> can be automatically
+    detected. If there is no or multiple SAP systems installed on the operating system
+    (or virtual machine), the test will fail. In this case, please use the general purpose
+    test <tt>accounts_authorized_local_users</tt> and customize the refine value variable
+    <tt>var_accounts_authorized_local_users_regex</tt> by adding all the authorized
+    user names to the list.
+    Test result of both <tt>false</tt> or <tt>error</tt> means mismatch of user names and
```
dahaic: Results is `fail` not `false`

matejak:
@xiruiyang Thank you for your modifications! I am OK with the `[[` section in the remediation, but I have noticed that a regexp that could live in a variable with a nice name is used repeatedly as a literal, and I still see some room of improvement with the naming, so please give it another round.

xiruiyang:
@dahaic @matejak improvement were made to the rule and the bash file:
- moved limitation to warnings
- improved the usage of find regex
- made it possible for bash to fix when there is no SAP system or /sapmnt is a symlink

dahaic:
Thank you for the changes, and for the contribution @xiruiyang ! ACK :)

xiruiyang:
Thank you @dahaic and @matejak :-)

PR #2939, opened 2018-06-11
Moved executables not used by the build system to misc, deleted unuse…

mpreisler:
…d code

Pretty minor stuff.


redhatrises:
Nack on moving to misc. Let's keep this in utils. I want to remove `shared/misc` anyway.

mpreisler:
@redhatrises utils has build system stuff, misc has other stuff. We can change the names but we should clearly separate them.

redhatrises:
@mpreisler are any of the utils useful that have been deleted?

redhatrises:
I think then that this PR should be closed and a new one opened that moves tools in their respective folders per irc discussion.

mpreisler:
> @mpreisler are any of the utils useful that have been deleted?

The verify CCEs are not useful, they have been replaced by @cipherboy 's work and we now validate CCEs when building. No need to have an extra tool for it. Plus the verify-cce script is broken and using lxml.

cipherboy:
@mpreisler / @redhatrises 

I agree with the executable move change.

With respect to removing `verify-cce.py`, if it were working, I'd suggest keeping it as it has different functionality than what is in the build system. However, it is broken, requires several changes to make it work, and it is not used as part of the build system. Therefore I'm inclined to remove it with a chance at replacing it with a working version as part of a later PR.

redhatrises:
That's fine. I still think that we should close this PR and open a new one that refactors all the scripts in `./build-scripts` and either `shared/utils` or `./utils`. Get it all done at once.

PR #2940, opened 2018-06-11
Remove testoval

mpreisler:
It's broken and not used by anything / anyone. It's not documented.

This may be controversial, please leave this open and let's get opinions of at least a few people before we decide what to do with this.


xiruiyang:
@mpreisler I'm using this when developing new OVAL test. Is there a replacement for it? Thanks.

redhatrises:
Nack. This is being used.

redhatrises:
Also, it is documented at https://github.com/OpenSCAP/scap-security-guide/blob/master/docs/manual/developer_guide.adoc#71-testing-oval-content

PR #2942, opened 2018-06-12
Remove unused scripts.

matejak:
* `shared/misc/cce_extract.py` script has misleading description in comments, uses `lxml`, isn't Python3 compatible and isn't mentioned anywhere in the project. Although it could be refactored, it may make more sense to just remove it.
* `shared/misc/find_orphans.py` looks fine, but it doesn't work for some time as the directory structure changed since it has been written. As nobody noticed, it makes sense to drop it.


mpreisler:
ACK.

The CCE script is obsolete now. The find orphans looks useful but I dont' like the approach. I think we should have a script that does this check on the final XCCDF and OVAL files and finds unused OVALs. Doing that requires an almost complete rewrite of the script so I am OK with removing it for now.

PR #2943, opened 2018-06-12
Get rid of XSLT platform-cpes-macro.

matejak:
- Moved data used by the macro to `product.yml` files.
- Moved the macro functionality to `yaml-to-shorthand.py`

This PR modifies the implementation, behavior should stay the same.
Whenever the `yaml-to-shorthand.py` detected that the macro was specified in the yaml under the `cpes` key, it disregarded all other members of `cpes` list, and passed the macro through as an element to the shorthand XML. It would then be picked up by the XSLT transform.
Now, if the `cpes` list from the benchmark's yaml is empty, `yaml-to-shorthand.py` uses contents of `product.yml` and performs the substitution, so shorthand already contains platform CPEs.


mpreisler:
I suggest an alternative (and IMO easier) approach. Use jinja2 inside benchmark.guide to add the CPEs.

shared/utils/yaml-to-shorthand.py
```diff
@@ -226,6 +226,7 @@ def __init__(self, id_):
         self.front_matter = ""
         self.rear_matter = ""
         self.cpes = []
+        self.platform_cpes = []
```
mpreisler: CPEs and platform CPEs? IMO we should take this opportunity to unify it. There are only one CPEs. Calling it platform may also be misleading because we use them for JRE, Firefox, etc... You can think of apps as platforms but IMO it's a bit misleading.

matejak:
I thought of using jinja, but I don't see how it would be done, unless I disable jinja cache completely.
Jinja is able to read environment at macro-define-time, and it doesn't have mechanism to detect that a macro depends on an environment symbol that has changed, so its cache should be purged and just that macro should be redefined.

shared/utils/yaml-to-shorthand.py
matejak: OK, what does this mean in practice? I guess that we are still going to distinguish between them in the code (as one set of CPEs comes from benchmark config and the other fone from the product config), but they should both be `cpes:` in yamls?

mpreisler: What's the difference between platform_cpes and cpes? Do we need both? Probably not because the confusion is not worth the benefit if there even is any benefit..

PR #2944, opened 2018-06-12
Remove subprocess check output

cipherboy:
While looking at refactoring `ssgcommon.py`, I noticed that this shim for Python 2.6 was still in our build tree; however, [some](https://github.com/OpenSCAP/scap-security-guide/blob/master/tests/ssg_test_suite/oscap.py#L61) [other](https://github.com/OpenSCAP/scap-security-guide/blob/master/shared/transforms/pcidss/generate_pcidss_json.py#L163) [instances](https://github.com/OpenSCAP/scap-security-guide/blob/master/shared/misc/generate_contributors.py#L107) don't use the shim.

These commits remove the shim's usages and the shim itself.

### DIFF
--- a/shared/modules/ssgcommon.py
+++ b/shared/modules/ssgcommon.py
@@ -1,6 +1,5 @@
 import datetime
 import platform
-import subprocess
 import re
 import codecs
 import jinja2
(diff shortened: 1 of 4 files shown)


redhatrises:
Hmmm..... I am pretty sure that the shim is needed for python2.6 which is used on rhel6 which is important for downstream users, no?

mpreisler:
> Hmmm..... I am pretty sure that the shim is needed for python2.6 which is used on rhel6 which is important for downstream users, no?

We are slowly but surely dropping support for python2.6 in SSG upstream. Content suitable for RHEL6 can still be built, it's just that you need python 2.7 or higher to do it.

redhatrises:
> We are slowly but surely dropping support for python2.6 in SSG upstream. Content suitable for RHEL6 can still be built, it's just that you need python 2.7 or higher to do it.

Can this be done on RHEL6 for users who consistently use upstream and downstream users who are not RHT internal?

cipherboy:
Based on feedback, I'm closing this.

PR #2945, opened 2018-06-12
Better paths for build scripts

mpreisler:
Easier to understand, clear separation between build scripts and utilities.


mpreisler:
> Does it make sense to continue to have the utils folder under shared or no?

IMO it does not. But I don't feel strongly either way. I moved it because that's what we discussed yesterday.

cipherboy:
I'm fine with a flatter structure. Eventually I'd like `ssg` to be a base directory python module, and holding all of the `ssg`-related logic in there. The goal would also be to clean up imports, documentation, and PEP8 to hopefully make it more maintainable. It'd be really great if we could get it down to "import ssg" but I'm not sure if that's possible.

So perhaps something like:
 - `/build-scripts` for front ends used during build,
 - `/utils` for misc. executable utilities (with some being module calls into `ssg`)
 - `/ssg` for holding a single python module for build+utils
 - `/shared` for not-executed code and non-code assets like macros, XLSTs, templates, images, etc.

mpreisler:
> @mpreisler Both fix-rules.py and upload_ansible_roles_to_galaxy.py need their ssgcommon import to be updated.

Good catch, fixed.

redhatrises:
`docs/manual` should be updated as well.

mpreisler:
> docs/manual should be updated as well.

Done, a lot of outdated paths there. Hopefully it's all fixed.

mpreisler:
@openscap-ci test this please

cipherboy:
Let's get this done with as it blocks #2949. ACK.

cipherboy:
Thanks @mpreisler!

PR #2948, opened 2018-06-13
Changed rpm_verify_permissions rule remediation to print only errors

matusmarhefka:
#### Rationale:
The rpm_verify_permissions rule remediation uses 'rpm --setperms' which
produces thousands of 'cannot acces' messages in case of missing files
(this mainly applies for containers which are stripped off certain files),
for example:

`chmod: cannot access '/usr/share/locale/aa': No such file or directory`

This commit adds '--quiet' option to the rpm command to only display
error messages.

### DIFF
--- a/rhel6/guide/system/software/integrity/rpm_verification/rpm_verify_permissions.rule
+++ b/rhel6/guide/system/software/integrity/rpm_verification/rpm_verify_permissions.rule
@@ -2,7 +2,7 @@ documentation_complete: true
 
 title: 'Verify and Correct File Permissions with RPM'
 
-description: "The RPM package management system can check file access\npermissions of installed software packages, including many that are\nimportant to system security. After locating a file with incorrect\npermissions which can be found with <pre>$ rpm -Va | grep '^.M'</pre>,\nrun the following command to determine which package owns it:\n<pre>$ rpm -qf <i>FILENAME</i></pre>\nNext, run the following command to reset its permissions to \nthe correct values:\n<pre>$ sudo rpm --setperms <i>PACKAGENAME</i></pre>"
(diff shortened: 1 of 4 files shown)


yuumasato:
Looks good, but there is no need to change the official disa stig references.

mpreisler:
> Looks good, but there is no need to change the official disa stig references.

+1 please revert that change, the references are official and we can't change them. Otherwise this looks good.

rhel6/guide/system/software/integrity/rpm_verification/rpm_verify_permissions.rule
```diff
@@ -2,7 +2,7 @@ documentation_complete: true
 
 title: 'Verify and Correct File Permissions with RPM'
 
-description: "The RPM package management system can check file access\npermissions of installed software packages, including many that are\nimportant to system security. After locating a file with incorrect\npermissions which can be found with <pre>$ rpm -Va | grep '^.M'</pre>,\nrun the following command to determine which package owns it:\n<pre>$ rpm -qf <i>FILENAME</i></pre>\nNext, run the following command to reset its permissions to \nthe correct values:\n<pre>$ sudo rpm --setperms <i>PACKAGENAME</i></pre>"
+description: "The RPM package management system can check file access\npermissions of installed software packages, including many that are\nimportant to system security. After locating a file with incorrect\npermissions which can be found with <pre>$ rpm -Va | grep '^.M'</pre>,\nrun the following command to determine which package owns it:\n<pre>$ rpm -qf <i>FILENAME</i></pre>\nNext, run the following command to reset its permissions to \nthe correct values:\n<pre>$ sudo rpm --quiet --setperms <i>PACKAGENAME</i></pre>"
```
mpreisler: I won't hold this PR because of this but why not reflow this so it's readable :-/

matusmarhefka: Agree.

mpreisler:
ACK

PR #2949, opened 2018-06-13
Refactor build system, create ssg module, and improve maintainability

cipherboy:
~For initial feedback only. I have made _everything_ a separate commit to make it easy to see what has changed/review in each step. If enough ACKs are acquired, I will squash adjacent commits such that each one builds on its own. Further, the identified dependencies will require rebase + additional work if/after they are merged.~

This has been squashed into fewer commits and dependent PRs have been merged. 

Dependencies: ~#2945~
Tentative dependencies: ~#2943~, ~#2950~

#### Description & Scope

This fairly large series of commits refactors a significant portion of the build system to use a new module, `ssg`, and places all shared code into this module. 

- Adds new '/ssg' module based off of `ssgcommon.py`
- Adds contents of `/shared/modules` to `/ssg`
    - **There is one remaining module, testoval.py that I will move in a second PR.**
- Refactors `/ssg` for smaller, logically scoped files.
- Adds `/ssg` to CMake, removing file-based PYTHONPATH hacking.
- Uses `/ssg` module everywhere.
- Other misc. cleanups including:
    - Consistency of shim usage -> More in later PRs.
    - Fixing broken scripts -> `xccdf2csv_stig.py` mostly, others in later PR.
    - Ensuring python2 / python3 compatibility. --> `ssg/_stig.py` only updated reference, unit testing will help with this more.

#### Rationale & Plans

This is the first in a series of PRs looking to improve the long-term maintainability of SSG. Later PRs will improve developer documentation and in-source documentation.

I'd like this PR to merge before I look at building a library editor for the subset of YAML/Jinja that rules/groups/vars/profiles/etc. use, and before trying to add easy file viewing/editing buttons in the generated reports.


cmake/SSGCommon.cmake
```diff
@@ -65,7 +65,7 @@ macro(ssg_build_bash_remediation_functions)
 
     add_custom_command(
         OUTPUT "${CMAKE_BINARY_DIR}/bash-remediation-functions.xml"
-        COMMAND "${PYTHON_EXECUTABLE}" "${SSG_SHARED_UTILS}/generate-bash-remediation-functions.py" --input "${SSG_SHARED}/bash_remediation_functions" --output "${CMAKE_BINARY_DIR}/bash-remediation-functions.xml"
+        COMMAND ${CMAKE_COMMAND} -E env "PYTHONPATH=$ENV{PYTHONPATH}" "${PYTHON_EXECUTABLE}" "${SSG_SHARED_UTILS}/generate-bash-remediation-functions.py" --input "${SSG_SHARED}/bash_remediation_functions" --output "${CMAKE_BINARY_DIR}/bash-remediation-functions.xml"
```
cipherboy: According to [here](https://cmake.org/Bug/view.php?id=5145), build failure is because `cmake -E env` isn't available before CMake 3.1. I don't think we're willing to up the CMake minimum version at this time, so I'll replaced it with just `env`, though this might not play nicely on Windows. Do we support building on Windows?

ssg/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+import ssg._map_product as map_product
+import ssg._parse_oval as parse_oval
+import ssg._stig as stig
+import ssg._utils as utils
+import ssg._xccdf as xccdf
+import ssg._xml as xml
+import ssg._yaml as yaml
+
+# Compatibility shims
+import ssg._shims as shims
```
redhatrises: Having an underscore `_` seems weird to me.... I would expect to not have an `underscore` at all.

ssg/pep8.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+python3-pep8 *.py
```
redhatrises: I don't think that a standalone script that runs pep8 should be added here. Seems like it should be added to CMake.

mpreisler: yeah, IMO this should be a ctest check

ssg/__init__.py
mpreisler: I am not 100% sure but I think this is to prevent name clashes. For example you need `_yaml` to avoid `import yaml` importing your own module inside your own module...

This `import _abc as abc` just renames it in the ssg package so that we have nice package names when we are using the module.

cmake/SSGCommon.cmake
redhatrises: Yes, we should not use `env` for multi-platform support.

ssg/pep8.sh
cipherboy: Yeah, sorry, this was a personal artifact that leaked in when I bulk added `ssg`. I want to add this as a `ctest` eventually but haven't gotten around to it yet.

ssg/__init__.py
cipherboy: Yeah, @mpreisler is correct here. There's some workaround with absolute vs. relative paths that you can do, but both `xml` and `yaml` conflict, and naming `_xml` instead of `xml` ends up being shorter. I opted to make everything consistent after that, so that's why, e.g., `_shims` begins with an underscore.

When loading from outside, `import ssg` will import everything with expected names (so you can use `ssg.xml.parse_xml_file(...)`), but if you want to import specific functions, you have to use the underscore (e.g., `from ssg._xml import parse_xml_file` or `import ssg._stig as stig`).

matejak: Python3 has relative imports that should prevent shadowing of imports. I am not against underscores, though.
What should be changed are scripts that define functions and that have dash in their names, s.a. `combine-ovals.py`. Those files can't be imported normally, and therefore can't be unit-tested.

cmake/SSGCommon.cmake
mpreisler: We do not support buildling on Windows but we really really want to. The rest of cmake port was done with as much care to make sure it can be used on Windows.

ssg/__init__.py
mpreisler: @matejak they can be imported normally. check out `__import__`

cipherboy:
@openscap-ci tests this please

cipherboy:
@openscap-ci test this please

ssg/__init__.py
matejak: @mpreisler I know that it can be done, by "normal imports" I meant import statements. It is not accurate that those files couldn't be unit tested. However, I think that files that define functions that are testable, should have import-friendly names.

cipherboy: @matejak I hope to address this in a second PR after this lands. I want to take everything in `build-scripts` and `utils` and move functionality into the `ssg` module such that it can be better unit tested... That way `build-scripts` and `utils` are just wrappers for calling (main methods). When I move them into `build-scripts`, I'll clean up their names.

matejak: @cipherboy I like the idea not to do everything in this PR.
Just to name an example, the `generate_contributors.py` has a `main` function. That one could be split into multiple functions, but it doesn't make economical sense to unit-test any of them - they just generate text. The rest of the functionality could be imported from a separate package and separate module with underscores in their names.
The script itself can go back to being `generate-contributors.py`.

.pyenv.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+script_dir="$(dirname "${BASH_SOURCE[@]}")"
+current_dir="$(pwd)"
+pathdir="$(python -c "import os.path; print(os.path.abspath(os.path.join('$current_dir', '$script_dir')))")"
+echo "Setting PYTHONPATH=$pathdir"
+export PYTHONPATH="$pathdir"
```
matejak: This is interesting, what is the use case of this script?

CMakeLists.txt
```diff
@@ -7,6 +7,12 @@ set(CMAKE_INSTALL_LIBDIR "/nowhere")
 include(GNUInstallDirs)
 include(FindPythonModule)
 
+if ( "$ENV{PYTHONPATH}" STREQUAL "" )
+    set(ENV{PYTHONPATH} "${PROJECT_SOURCE_DIR}:${PROJECT_SOURCE_DIR}/shared:${PROJECT_SOURCE_DIR}/shared/modules")
```
matejak: Why to set `PYTHONPATH` to anything else than project source dir? There is that testoval module, but why does that one deserve its own directory?

.pyenv.sh
cipherboy: Since we don't envision installing this python module system-wide, every script which uses `ssg` needs to know its location via `PYTHONPATH`. Scripts in `build-scripts` are passed it via `cmake`, but scripts in `utils` have no way of knowing `ssg`'s location. My aim was to try and pull out as many of the `sys.path` modifications as I could. This way, the location of `ssg` could be changed at a future date without impacting every single script.

Usage would be: `source .pyenv.sh`; then `PYTHONPATH` gets set to a location with `ssg` and any script depending on it can be run after that. It could also potentially be used as part of `cmake` behind a platform flag if I remove `Setting ` from the echo (then usage could be `.pyenv.sh` python /script.py`).

Also with `cmake`,  and an extension to the script, we could do (in pseudo-cmake):

```
if unix:
 set pyenv ${basedir}/.pyenv.sh
else if windows:
 set pyenv ${basedir}/.pyenv.bat
```
...
```
 ${pyenv} ${python executable} args 
```

and the script could be modified run the specified command with the correct PYTHONPATH set. This would help with Windows support eventually, while still allowing us to support old platforms (read: RHEL6/RHEL7) which don't have CMake 3.1 / `cmake -E env` support.

All of this seems like work for future PRs. Thoughts?

CMakeLists.txt
cipherboy: @matejak `testoval` module is being moved in the next PR. I can move it to this PR if you want though, and then we can clean up CMake early. I agree that this PR is large though :) Preference? 

Perhaps remove `pep8` fixes but add moving `testoval`?

cipherboy:
> Underscores in module names may cause some controversy, but as they are hidden for users of the ssg package, they are just an implementation detail. I hope that we will move away from leading underscores when we drop Python2 support, but we will be able to do it at any time without breaking backwards compatibility.

Agreed, and by removing as many usages of `from <somewhere> import x, y, z` by consumers of `ssg`, this should make the move easier. 

> The current way assumes that the root of SSG will be added to PYTHONPATH, which is somehow messy. I would prefer to move the ssg directory one directory level below, e.g. python, source, engine or whatever else.

As mentioned deeper above, (I think) PYTHONPATH can be changed in two places (CMake + `.pyenv.sh`) and this can change. If an agreed-upon location is decided, I can move it. However, I was hoping to make the directory structure a little more flat. :)

cipherboy:
Ah, oops. When I squashed all commits, I forgot to push to a separate branch to save the original commits as well. Removing PEP8 fixes will be difficult.... :/

.pyenv.sh
redhatrises: I personally would prefer something a little more robust than the use of cmd and shell scripts, and I do see a future possibility to installing `ssg` system-wide to further make content writing more accessible to users.

ssg/__init__.py
matejak: Sure, `generate-contributors` was just an example.

.pyenv.sh
matejak: @cipherboy please add the condensed intention of the script to its comment.
@redhatrises you have a good point, we could move towards making this an installable package with library part (the `ssg` in this directory) and maybe executable part (contents of `utils`, `build-scripts`).

matejak:
@cipherboy Regarding the directory structure, I think that I understand your point of view, I also don't like deep directory structure even for small projects. However, root directory may contain Python scripts that would become importable. Having `PYTHONPATH` in a subdirectory would provide greater control over what can and can't be imported.

cipherboy:
> @cipherboy Regarding the directory structure, I think that I understand your point of view, I also don't like deep directory structure even for small projects. However, root directory may contain Python scripts that would become importable. Having PYTHONPATH in a subdirectory would provide greater control over what can and can't be imported.

How about we fix this in a future PR? The technical reason I'm against fixing it this issue in this PR is that any other changes to the build system which want to happen have a large chance of getting dropped silently, since e.g., `ssgcommon.py` was marked as deleted vs. recognizing that parts of it were split into other files. If you or anyone else has suggestions of how to improve this and make it a noisy merge conflict or otherwise do the right thing, I'd appreciate it.

ssg/__init__.py
cipherboy: @matejak Is https://github.com/OpenSCAP/scap-security-guide/pull/2967/commits/1c0d49006f68f4b92831335824a78901efba8501 from #2967 what you're thinking?

cipherboy: (Note that I'm going to turn all the _build_*.py scripts into a proper submodule after this, but it is an example of an issue from my other comment [here](https://github.com/OpenSCAP/scap-security-guide/pull/2949#issuecomment-397594152).

CMakeLists.txt
cipherboy: @matejak Moved into this PR, updated CMakeLists.txt in the last commit.

.pyenv.sh
cipherboy: @matejak Is the new comment sufficient? https://github.com/cipherboy/scap-security-guide/blob/a8986890b3d72f717fb59b61271226ff3fcc0b3c/.pyenv.sh

matejak:
I see that `testoval` was moved, great! Moving `ssg` somewhere else is not even decided, and definitely isn't part of this PR. I see this as very close to being merged, I will perform another regression review, and likely merge it later today.

cipherboy:
@matejak Thanks!

cipherboy:
(Rebased this on top of the latest master to make the other PR easier for me).

matejak:
No regressions were found and nobody fundamentally opposes the nature of this PR, so it is my pleasure to merge it. Great work!

PR #2950, opened 2018-06-13
Remove xccdf2csv-stig module and transforms

cipherboy:
From what I can tell, this code is not used as part of the build system or the test suite. Further, the code uses `lxml.etree` whereas everywhere else uses `xml.etree.ElementTree`.


cipherboy:
If anyone is using this, I'll close the pull request and replace it with one commenting the use case and merging the transforms into one in `/shared` (they're all identical).

redhatrises:
This is actually a valuable tool for those who use Excel (for better or for worse). Not to be done in this PR, but it would be great to update.

cipherboy:
@redhatrises Okay, I'll keep it and update in a separate PR. Why is it under "transforms" though -- should it be a "utils" instead? Perhaps moving one copy to "utils" should be part of @mpreisler's PR #2945.

mpreisler:
> @redhatrises Okay, I'll keep it and update in a separate PR. Why is it under "transforms" though -- should it be a "utils" instead? Perhaps moving one copy to "utils" should be part of @mpreisler's PR #2945.

Yeah, I agree, it shouldn't be in transforms

redhatrises:
@cipherboy I think it was placed there by accident.

cipherboy:
Okay, based on feedback, I'm closing this.

PR #2951, opened 2018-06-13
Unshadow shared sysctl

yuumasato:
#### Description:

- Drop checks in shared that can be generated from templates
- Add dropped checks to CSV of products that need them

#### Rationale:

- Un-shadows `kernel_randomize_va_space` checks
- Un-shadows `net_ipv6_conf_all_disable_ipv6` check

#### Notes
- I actually don't like that I'm creating new CSV files


mpreisler:
ACK

PR #2952, opened 2018-06-13
Get rid of more XSLT macros.

matejak:
* Converted the product-name-macro.

### DIFF
--- a/chromium/product.yml
+++ b/chromium/product.yml
@@ -6,3 +6,5 @@ profiles_root: "./profiles"
 
 cpes:
   - "cpe:/a:google:chromium-browser"
+
+full_name: Chromium
--- a/ubuntu1404/product.yml
+++ b/ubuntu1404/product.yml
@@ -10,3 +10,5 @@ init_system: "systemd"
 
 cpes:
   - "cpe:/o:canonical:ubuntu_linux:14.04"
+
+full_name: "Ubuntu 14.04"
--- a/ubuntu1604/product.yml
+++ b/ubuntu1604/product.yml
@@ -10,3 +10,5 @@ init_system: "systemd"
 
 cpes:
   - "cpe:/o:canonical:ubuntu_linux:16.04"
+
+full_name: "Ubuntu 16.04"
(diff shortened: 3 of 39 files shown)


chromium/product.yml
```diff
@@ -6,3 +6,5 @@ profiles_root: "./profiles"
 
 cpes:
   - "cpe:/a:google:chromium-browser"
+
+full_name: Google Chromium
```
mpreisler: just FYI, Google Chromium doesn't exist. It's either Chromium or Google Chrome.

mpreisler:
Looks good. I like that you are keeping full product names in yaml. That should prove much better than the constantly out of date map_product_module.py

matejak:
That's it for the time being, further work is complicated by caching - as all macros are cached at read-time, they lose ability to reflect changes made to global variables, which makes e.g. `os-type-macro` impossible to replace in non-hackish way.

ubuntu1404/product.yml
```diff
@@ -10,3 +10,5 @@ init_system: "systemd"
 
 cpes:
   - "cpe:/o:canonical:ubuntu_linux:14.04"
+
+full_name: Ubuntu 1404
```
mpreisler: this should be 14.04

ubuntu1604/product.yml
```diff
@@ -10,3 +10,5 @@ init_system: "systemd"
 
 cpes:
   - "cpe:/o:canonical:ubuntu_linux:16.04"
+
+full_name: Ubuntu 1604
```
mpreisler: 16.04

I am also wondering if we should include the nickname for the release. I know a lot of ubuntu users know the nicknames better than release numbers.

mpreisler:
I suggested 2 more super minor changes that I didn't notice before, then it can be merged.

matejak:
Fixed by rebase. For the reference, those values have been exported from respective `transforms/constants.xslt`, but the full name constant is still used elsewhere, so it has to stay there.

PR #2955, opened 2018-06-13
element_child_cache needs to be a global variable, also use defaultdict

mpreisler:
Fixes scrutinizer issue, makes code slightly easier to read.

Mostly just testing scrutinizer.

I have to say I hate how the element_child_cache is done and the fact that it's global... I have only myself to blame :-D


cipherboy:
LGTM.

cipherboy:
@mpreisler Thanks!

PR #2957, opened 2018-06-13
Fix pytest unit tests

mpreisler:
I have accidentally broken python unit tests in #2945 and jenkins somehow missed it. This fixes it.

I also took the liberty to improve how they are defined and refactored the cmake code, indents were off and it was all defined in root CMakeLists and SSGCommon which is not the right place for it.


cipherboy:
@openscap-ci test this please

cipherboy:
Another failure from #2945 against me. :P 

This runs and works fine on my system (now with pytest + pycov installed!). ACK. I'll re-run failing tests.

PR #2958, opened 2018-06-13
Ignore the ropeproject folders

mpreisler:
ROPE is a python vi plugin for code completion and stuff like that, it
generates these files for caching purposes.

This should avoid accidentally adding this cache into git.


matejak:
Yes, [python-mode](https://github.com/python-mode/python-mode) uses it. ACK.

PR #2960, opened 2018-06-13
The shared overridden checks were only used in RHEL7

mpreisler:
Despite the comment inside them about rhel6 they are not used for rhel6.
As far as I can tell the template generated dmesg restrict check does
the same thing as the overridden one so there is no reason to keep both.


yuumasato:
The templates have the same comments as the removed checks.
And only RHEL7 selects `sysctl_kernel_dmesg_restrict` rule.
Ack.

PR #2961, opened 2018-06-13
All the sebool checks and fixes are only used on RHEL7

mpreisler:
For some reason we build selinux boolean checks and fixes for every single product, despite the yaml metadata rules all having `prodtype: rhel7`

This makes resulting OVAL files smaller.


yuumasato:
Also, only RHEL7 selects `sebool` rules.
Ack.

PR #2966, opened 2018-06-14
There is no need to duplicate the debian8 CPE OVAL def

mpreisler:
This is a less ambitious version of #2923 that I need for debian and ubuntu unshadowing. The other CPE OVALs will have to wait because they uncover serious issues.


mpreisler:
> @mpreisler Is the file in shared/ good enough? The version number is different, there's a few minor ref_id differences, and the debian8 one first checks that /etc/debian_version exists.

Yeah, you are right. The CPE OVAL def in debian8 is slightly better and seems more robust. I decided to keep that instead of the other one.

cipherboy:
Much better, LGTM :)

PR #2969, opened 2018-06-14
Fail early in build-scripts/verify-references.py on error

cipherboy:
Fail earlier in `verify-references.py` on an error. It already errors eventually; this just fails earlier.


mpreisler:
Hmm, isn't the point of this to write all the errors at once? Instead of just telling you one error, you fix it, rerun, fix next error?

cipherboy:
Perhaps, but outside of this particular script, every other build script seems to exit upon encountering an error, either via raising or via a direct `sys.exit(...)` call. I like consistency, but that's me :)

yuumasato:
I prefer the approach in which most errors are printed until it is not possible to continue.

cipherboy:
Duly noted; I'll close this PR then.

PR #2970, opened 2018-06-14
Remove find_untemplated.py

mpreisler:
The script is outdated and doesn't do anything useful anymore. Check out
utils/find_shadowed_files.py for a similar tool.


cipherboy:
Looks like it was added in 2012 and updated throughout the years to change a couple of shebangs and that's it. ACK from me.

cipherboy:
@mpreisler Thanks for the PR!

PR #2973, opened 2018-06-15
put string /sapmnt/<SID> into a variable and renamed two other variab…

xiruiyang:
#### Description:
- put string /sapmnt/<SID> into a variable and renamed two other varialbes for more descriptive as suggested by matejak

### DIFF
--- a/linux_os/guide/system/software/sap/accounts_authorized_local_users_sidadm_orasid.rule
+++ b/linux_os/guide/system/software/sap/accounts_authorized_local_users_sidadm_orasid.rule
@@ -19,7 +19,7 @@ description: |-
     OVAL regular expression is used for the user list.
     <br /> <br />
     Test result of both <tt>fail</tt> or <tt>error</tt> means mismatch of user names and
-    SAP system.<tt>*</tt> The bash remediation commands can be used to delete unexpected users on
+    SAP system. The bash remediation commands can be used to delete unexpected users on
     the operating system. 
 
 rationale: |-
@@ -42,14 +42,14 @@ ocil: |-
 warnings:
     - general: |-
         Currently this rule only works with following limitations:
-        <br /> <br />
+        <br />
         1. <tt>SAP system mount directory</tt> is <tt>/sapmnt</tt> (mounted or local file system
         or a symbolic link to the target directory);
-        <br /> <br />
+        <br />
         2. there is maximum one SAP System on each operating system or virtual machine (maximum
         one SID in /sapmnt and /usr/sap).
-        <br /> <br />
-        In this case, the SAP system users <tt>sidadm</tt>, <tt>orasid</tt>, <tt>sapadm</tt>
+        <br />
+        With the above limitations, the SAP system users <tt>sidadm</tt>, <tt>orasid</tt>, <tt>sapadm</tt>
         and <tt>oracle</tt> can be automatically detected. 
         <br /> <br />
         For other cases, please use the general purpose rule <tt>accounts_authorized_local_users</tt>
--- a/linux_os/guide/system/software/sap/accounts_authorized_local_users_sidadm_orasid.rule
+++ b/linux_os/guide/system/software/sap/accounts_authorized_local_users_sidadm_orasid.rule
@@ -19,9 +19,8 @@ description: |-
     OVAL regular expression is used for the user list.
     <br /> <br />
     Test result of both <tt>fail</tt> or <tt>error</tt> means mismatch of user names and
-    SAP system. Nevertheless, the bash remediation commands can always be used to delete
-    unexpected users on the operating system. It works in all cases zero, one and multiple
-    SAP systems on the operating system (or virtual machine).
+    SAP system.<tt>*</tt> The bash remediation commands can be used to delete unexpected users on
+    the operating system. 
 
 rationale: |-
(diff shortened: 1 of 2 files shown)


xiruiyang:
@dahaic and @matejak , I made small changes to the string and variables in the bash remediation for PR#2938. Could you please take a quick look?  Thanks and Best Regards, Xirui  :-)

linux_os/guide/system/software/sap/accounts_authorized_local_users_sidadm_orasid.rule
```diff
@@ -19,9 +19,8 @@ description: |-
     OVAL regular expression is used for the user list.
     <br /> <br />
     Test result of both <tt>fail</tt> or <tt>error</tt> means mismatch of user names and
-    SAP system. Nevertheless, the bash remediation commands can always be used to delete
-    unexpected users on the operating system. It works in all cases zero, one and multiple
-    SAP systems on the operating system (or virtual machine).
+    SAP system.<tt>*</tt> The bash remediation commands can be used to delete unexpected users on
```
dahaic: What do you mean by the <tt>*</tt> ? It feels out of place there. Was it supposed to be some pointer to the warning section?

linux_os/guide/system/software/sap/accounts_authorized_local_users_sidadm_orasid.rule
```diff
@@ -42,13 +41,20 @@ ocil: |-
...
-        authorized to the system. 
+        Currently this rule only works with following limitations:
         <br /> <br />
-        In this case, please use the general purpose test <tt>accounts_authorized_local_users</tt>
+        1. <tt>SAP system mount directory</tt> is <tt>/sapmnt</tt> (mounted or local file system
+        or a symbolic link to the target directory);
+        <br /> <br />
+        2. there is maximum one SAP System on each operating system or virtual machine (maximum
+        one SID in /sapmnt and /usr/sap).
+        <br /> <br />
```
dahaic: I like how you split it, I would maybe get rid of this newlines - the explanation below is directly connected to the point 2, so it make sense to have it in the same paragraph. Rest is ok.

shared/fixes/bash/accounts_authorized_local_users_sidadm_orasid.sh
```diff
@@ -7,31 +7,31 @@ default_os_user="root"
 
 # add users sidamd, orasid, sapadm and oracle if needed
 userlist="root"
+sapmnt_SID_stem="/sapmnt/[A-Z][A-Z0-9][A-Z0-9]"
```
dahaic: The regex is not identical to the one you carved out of the script - is that deliberate, to not have explicit beginning and ending there?

linux_os/guide/system/software/sap/accounts_authorized_local_users_sidadm_orasid.rule
xiruiyang: @dahaic yes, it's supposed to be notation that there is a warning section for  this sentence. Any suggestion?

linux_os/guide/system/software/sap/accounts_authorized_local_users_sidadm_orasid.rule
xiruiyang: OK. Thanks @dahaic

shared/fixes/bash/accounts_authorized_local_users_sidadm_orasid.sh
xiruiyang: @dahaic It shows the same result with and without the beginning and ending characters '^' and '$' in bash.  But you are correct, it's better to add these characters to keep consistent with OVAL pattern match convention.

shared/fixes/bash/accounts_authorized_local_users_sidadm_orasid.sh
```diff
@@ -7,31 +7,31 @@ default_os_user="root"
...
 
 # if /sapmnt is a directory or a symbolic link to a directory,
 # then try to add SAP system users to the userlist
 if [ -d "/sapmnt" ] ; then 
 	# if /sapmnt/SID exists, add sidadm to the userlist
-	sapmntSIDlist=$(find /sapmnt/ -regex '^/sapmnt/[A-Z][A-Z0-9][A-Z0-9]$')
-	for i in $sapmntSIDlist ; do
+	path_sapmnt_SID_list=$(find /sapmnt/ -regex "^$sapmnt_SID_stem$")
+	for i in $path_sapmnt_SID_list ; do
```
matejak: I would suggest renaming `i` to something like `path_including_SID`, so it is also more obvious what is the intent behind `${...:8:3}`.

xiruiyang:
@matejak renamed the variables, thanks.

shared/fixes/bash/accounts_authorized_local_users_sidadm_orasid.sh
dahaic: You are right, it's implicit in the case of `-regex` it seems, sorry, I could have checked it myself before commenting :) Anyway I think it's slightly more readable now, for those not familiar with the argument.

dahaic:
Thank you for the fixes, @xiruiyang , it looks great now. ACK

PR #2974, opened 2018-06-15
Update cpe-generate.py help text

cipherboy:
The help text for `cpe-generate.py` was out of date and did not reflect the current set of nor count of arguments necessary to run the script. This should update that. 

If desired, I could convert this to using `argparse`. Preferences from anyone?

### DIFF
--- a/build-scripts/cpe-generate.py
+++ b/build-scripts/cpe-generate.py
@@ -212,5 +212,6 @@ def main():
 
     sys.exit(0)
 
+
 if __name__ == "__main__":
     main()
--- a/build-scripts/cpe-generate.py
+++ b/build-scripts/cpe-generate.py
@@ -6,6 +6,7 @@ import fnmatch
 import sys
 import os
 import ssg
+import argparse
 
 ElementTree = ssg.xml.ElementTree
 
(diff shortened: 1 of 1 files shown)


mpreisler:
I really don't want to change the scope of this but could we please get argparse support here? :-)

build-scripts/cpe-generate.py
```diff
@@ -59,18 +60,25 @@ def collect_nodes(tree, reflist):
...
+        "along with a synchronized CPE dictionary file.")
+    p.add_argument("product", help="Name of the product")
+    p.add_argument("idname", help="Identifier prefix")
+    p.add_argument("cpeoutdir", help="Artifact output directory")
+    p.add_argument("ovalfile", help="OVAL file to process")
+    p.add_argument("cpedictfile", help="CPE dictionary file to process")
+
+    args = p.parse_args()
+
+    return args.product, args.idname, args.cpeoutdir, args.ovalfile, args.cpedictfile
```
mpreisler: how about returning args here and then using the args dict in main()? I think that would make it more readable and you'd avoid having to list all the args twice.

cipherboy:
@mpreisler Updated.

PR #2967, opened 2018-06-14
Refactor and modularize build-scripts 

cipherboy:
Continuation of #2949.

~Dependencies: #2949. Will rebase after merge. Note that this currently has commits from #2949 as well.~

#### Description & Scope

The aim of this PR is to share code between various parts of the build system by turning them into modules. This fixes internal PEP8 issues and refactors the existing code.

 - ~`testoval_module.py` is moved to `ssg/_oval.py`~ -- Part of #2949. 
 - `add_stig_references.py` has code moved to `ssg/_stig.py`
 - `build-all-guides.py` has code moved to `ssg/_build_guides.py`
 - `build-all-remediation-roles.py` has code moved to `ssg/_build_roles.py`
 - `enable-derivatives.py` has code moved to `ssg/_build_derivatives`
 - `relabel-ids.py` has code moved to `ssg/_build_renumber.py`
 - `yaml-to-shorthand.py` has code moved to `ssg/_build_yaml.py` -- this is in keeping with `_build_roles.py`, where the name is based off of the source files, not the target. 
 - ... and more! See commit log for explicit list.

The eventual goal is to move all build-related modules into a submodule called `build`. (E.g., all files with `_build` as a prefix). 

#### Rationale

Each of these build scripts has significant internal code, including some that can be easily shared between multiple scripts. By converting these into modules and refactoring, eventually these parts can be unit tested and better documented.

#### Future Plans

Three more refactoring PRs will likely follow:
 - One for `/utils`, in a similar vein as this one
 - One for `/ssg`, to better organize the final module
 - One for the entire system, replacing `-` in paths with `_` for global consistency. 

After the refactoring is complete, two additional PRs will likely follow:
 - Documentation (inline + external) of the new system
 - Unit tests of critical components

PR #2979, opened 2018-06-15
We import the openscap python bindings just to guess the default OVAL…

mpreisler:
… version?!?

~~The build system always knows the version so let's just make it required.~~

### DIFF
--- a/ssg/_oval.py
+++ b/ssg/_oval.py
@@ -132,14 +132,30 @@ def read_ovaldefgroup_file(testfile):
     return body
 
 
+def get_openscap_supported_oval_version():
+    try:
+        from openscap import oscap_get_version
+        if [int(x) for x in str(oscap_get_version()).split(".")] >= [1, 2, 0]:
+            return "5.11"
+    except ImportError:
+        pass
+
+    return "5.10"
+
+
 def parse_options():
     usage = "usage: %(prog)s [options] definition_file.xml"
     parser = argparse.ArgumentParser(usage=usage)
     # only some options are on by default
 
-    parser.add_argument("--oval_version", required=True,
+    oscap_oval_version = get_openscap_supported_oval_version()
+
+    parser.add_argument("--oval_version",
+                        default=oscap_oval_version,
                         dest="oval_version", action="store",
-                        help="OVAL version to use. Example: 5.11, 5.10, ...")
+                        help="OVAL version to use. Example: 5.11, 5.10, ... "
+                             "If not supplied the highest version supported by "
+                             "openscap will be used: %s" % (oscap_oval_version))
     parser.add_argument("-q", "--quiet", "--silent", default=False,
                         action="store_true", dest="silent_mode",
                         help="Don't show any output when testing OVAL files")
--- a/ssg/_oval.py
+++ b/ssg/_oval.py
@@ -9,7 +9,6 @@ import subprocess
 
 from ssg._constants import oval_footer as footer
 from ssg._constants import oval_namespace as ovalns
-from ssg._constants import OSCAP_OVAL_VERSION
 from ssg._xml import ElementTree as ET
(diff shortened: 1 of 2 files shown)


cipherboy:
:) Should've thought about that one a little more before I just moved it around. ACK pending jenkins.

redhatrises:
Can we at least set a default in the arg?

mpreisler:
> Can we at least set a default in the arg?

We could I guess, I can't think of any use-case for it though. This is only called as part of the build system and the build system will always pass the version.

redhatrises:
Sorry..... why is this a part of the build system? If I am creating OVALs and want to test them, I am going to run the script directly.

mpreisler:
Ah OK, I got confused. This is for testoval.py and not combine-ovals.py

OK, what should it default to?

cipherboy:
@mpreisler `combine-ovals.py` will have functionality moved into `_build_ovals.py` in #2967.

cipherboy:
@mpreisler Probably default to the older version? I don't think we need to call out to the external python module and do all that work for every module load of `ssg`.

redhatrises:
Should at least be 5.10... Although having it autodetect what version was supported was really nice feature IMO.

mpreisler:
we can keep the autodetection but the way it's written it will do a lot of extra work every time ssg is imported... If we want autodetection it should default to None and then there is extra logic that imports the openscap bindigns if it is None...

cipherboy:
We could stick it in a function and return the result and take the hit on the PEP8/Scrutinizer for now.

mpreisler:
Ok, the functionality is back and now it even works!

```
$ PYTHONPATH=. ./utils/testoval.py  -h
usage: usage: testoval.py [options] definition_file.xml

positional arguments:
  XMLFILE               OVAL XML file

optional arguments:
  -h, --help            show this help message and exit
  --oval_version OVAL_VERSION
                        OVAL version to use. Example: 5.11, 5.10, ... If not
                        supplied the highest version supported by openscap
                        will be used: 5.11
  -q, --quiet, --silent
                        Don't show any output when testing OVAL files
```

The original implementation had this annoying problem:
```
>>> openscap.oscap_get_version() > 666.666
True
```

Turns out you can't compare bools and floats like this :-)

ssg/_oval.py
```diff
@@ -133,15 +132,30 @@ def read_ovaldefgroup_file(testfile):
     return body
 
 
+def get_openscap_supported_oval_version():
+    try:
+        from openscap import oscap_get_version
+        if str(oscap_get_version()).split(".") >= ["1", "2", "0"]:
```
cipherboy: @mpreisler This doesn't quite do what we want it to do, but it is a nice hack otherwise:

```python
>>> '1.2.16'.split(".") >= ['1', '2', '1']
True
>>> '1.2.16'.split(".") >= ['1', '2', '2']
False
```

I'm almost tempted to leave it, but add a comment about how this doesn't compare versions how we'd like it.

matejak: The builtin (but undocumented) `distutils.version.LooseVersion` should behave correctly and should be easy to use.

mpreisler: lol, what a fail... I will fix this.

mpreisler: fixed

mpreisler: > The builtin (but undocumented) distutils.version.LooseVersion should behave correctly and should be easy to use.

I went with list comprehension because python docs says that you don't want to use distutils directly. I suppose it doesn't make much difference either way.

cipherboy:
This is better. I like the cleaner interface suggested by @matejak but agree with @mpreisler that we shouldn't rely on `distutils`. ACK.

cipherboy:
@mpreisler Thanks!

PR #2980, opened 2018-06-15
Generate the mount_option_var_tmp_bind.xml OVAL check from template

mpreisler:
No need to write it by hand.

I am not 100% sure about the debian and ubuntu paths so I added a fixme.


sle12/templates/csv/mount_options.csv
```diff
@@ -6,14 +6,4 @@
 #     '$' to reference a variable, e.g. var_removable_partition,nodev)
 #  If the remediation can create (i.e. not just modify) an /etc/fstab line,
 #  add the 'create_fstab_entry_if_needed' literal string as the third argument.
-/dev/shm,nodev
```
cipherboy: It is not obvious to me why these are going away in SLE12. Perhaps could split the commit or add a comment on why these are getting removed?

mpreisler: I am 99% sure this was an accident, I have updated the PR.

cipherboy:
Thanks @mpreisler; this looks better. I'm not sure I understand the `maybe this should be /tmp` portion (iirc no system has `/tmp` bind-mounted as `tmpfs` is the filesystem backing `/tmp` on modern distros); thoughts?

mpreisler:
> Thanks @mpreisler; this looks better. I'm not sure I understand the maybe this should be /tmp portion (iirc no system has /tmp bind-mounted as tmpfs is the filesystem backing /tmp on modern distros); thoughts?

I think you are correct. I had the comment there because the ubuntu checks were checking for /tmp being bind mounted which I thought was wrong. I have removed that comment because it probably does more harm than good.

cipherboy:
This looks better. Thanks @mpreisler! ACK.

PR #2981, opened 2018-06-15
Ansible playbook check syntax

mpreisler:
I noticed that one of the ansible roles had a syntax error in it. This PR fixes that issue and adds an automated test that checks all generated ansible roles for syntax issues.


redhatrises:
Ack

Issue #2983, opened 2018-06-18
Templating for sockets disabled seems to be unused


yuumasato:
#### Description of problem:
Some hand written remediations disable the socket associated with a service, as part of remediation for rules that disable services.
Our fix templates for services disabled don't disable sockets.

I even noticed that we have an old and unused [generator](https://github.com/OpenSCAP/scap-security-guide/blob/master/shared/templates/create_sockets_disabled.py) for it.

#### SCAP Security Guide Version:
0.1.39

#### Steps to Reproduce:

1.
2.
3.
4.

#### Actual Results:
The fixes to disable the sockets are not generated by templates

#### Expected Results:
Services that provide sockets should have templated remediation to disable them.

yuumasato:
The unused template is fixed by #3057.
And the socket disabled functionality is added by: #3029, #3025, and #3055

PR #2984, opened 2018-06-18
Unshadow shared check service_dovecot_disabled

yuumasato:
#### Description:

- Drop service_dovecot_disabled check from shared 
- ~~Drop Ansible fix service_avahi-daemon_disabled form shared~~ commit dropped
- ~~Drop templated fixes for account_password_pam_~~ commit dropped
  - ~~maxclassrepeat~~
  - ~~maxrepeat~~
  - ~~minlen~~
- ~~Improve regex in template for accounts_password_pam to match commented and or indented line~~ commit dropped

#### Rationale:

- This will unshadow for other products as well.

### Note 
- Ansible fix for `service_avahi-daemon_disabled` from shared also included task to disable the socket associated with the service, but the templated fix doesn't disable sockets. I have filed #2983 to investigate usage and templating for disabling sockets.


mpreisler:
> Ansible fix for service_avahi-daemon_disabled from shared also included task to disable the socket associated with the service, but the templated fix doesn't disable sockets. I have filed #2983 to investigate usage and templating for disabling sockets.

AFAIK that's why it's shadowed right now, it also disables a socket. We can't unshadow it as you suggest in this PR.

yuumasato:
Right, we cannot generate the Ansible fix for `service_avahi-daemon_disabled` exactly as in shared right now, it makes sense to keep it.
Its unshadowing can be done in different PR, I have dropped the commit.

mpreisler:
You dropped the shared ansible fixes which means they will be missing for RHEL6 if this gets merged. Was that intentional?

yuumasato:
I was careless when dropping them, I will look into templating them for RHEL6. But I think in another PR. I will drop these commits soon.

yuumasato:
I have dropped the commits.

mpreisler:
ACK

PR #2985, opened 2018-06-18
Fix auditd rules tests

mildass:
#### Description:

Change profiles in tests, because of no `common` profile in latest build.
Remove `yum install` from functions and add it to tests.
Add/remove `# remediation = ` comments in tests corresponding to existing remediation scripts.


dahaic:
Hi! Great stuff. I have one more plea - can we have a bit different structure for helper functions?

`set_parameters_value` and `delete_parameter` can be sourced out of something like `utils.sh` file placed directly in `data/` directory (to imply it's usable for any scenario in the scenario tree)

`yum install audit audispd-plugins` can be installed as part of some `prepare_audit_environment` function, sourced from `group_auditing/` or `group_auditd_data_retention/` directory. And ideally, `audit_lib.sh` or whatever will be called the file containing the preparation function, will source the `utils.sh`. So at the end, scenario will look something like:

```
#!/bin/bash
# profiles = xccdf_org.ssgproject.content_profile_nist-800-171-cui
# remediation = bash

. ../../audit_lib.sh # this will source all functions
prepare_audit_test_environment # this will install needed packages
set_parameters_value /etc/audisp/plugins.d/syslog.conf "active" "yes"
```

mildass:
That sounds good.
I made changes, that u suggested, so take a look at it, if it's ok. Thanks

dahaic:
Thank you for the fixes! Looks great now :)

PR #2986, opened 2018-06-18
Moved the shared ansible accounts_password_pam_* fixes to rhel6 fixes

mpreisler:
They are only used by RHEL6, RHEL7 ones are generated from templates.


yuumasato:
Why not generate them for RHEL6 from templates too?

mpreisler:
> Why not generate them for RHEL6 from templates too?

We could I guess. I was not sure about that and my goal was to unshadow rhel7 so I went this route.

yuumasato:
I just noticed that this can't be simply moved. RHEL6 doesn't have `/etc/security/pwquality.conf`. The config is done via `/etc/pam.d/system-auth`.

mpreisler:
#2993 might have obsoleted this. I will verify later.

yuumasato:
The files moved to RHEL6 were actually dropped in 1abab14ec29984873d2af147427a7e5119a455fa.
Why aren't the files shown with conflicts?
I'm closing this.

Issue #2982, opened 2018-06-16
Couple of the file paths are problematic on Windows


mpreisler:
#### Description of problem:
Windows has a path limit of 260 chars for the whole file path. It can be circumvented if you use the new API but a lot of tools still use the old API so you do run into the problem from time to time. Examples:

```
Scanning for dependencies of wrlinux fixes (bash, ansible, puppet and anaconda)...
Scanning for dependencies of wrlinux checks (OVAL)...
Traceback (most recent call last):
  File "C:/Users/Martin/d/scap-security-guide/shared/utils/yaml-to-shorthand.py", line 724, in <module>
    main()
  File "C:/Users/Martin/d/scap-security-guide/shared/utils/yaml-to-shorthand.py", line 720, in main
    args.bash_remediation_fns, args.output, product_yaml)
  File "C:/Users/Martin/d/scap-security-guide/shared/utils/yaml-to-shorthand.py", line 662, in add_from_directory
    product_yaml)
  File "C:/Users/Martin/d/scap-security-guide/shared/utils/yaml-to-shorthand.py", line 662, in add_from_directory
    product_yaml)
  File "C:/Users/Martin/d/scap-security-guide/shared/utils/yaml-to-shorthand.py", line 662, in add_from_directory
    product_yaml)
  [Previous line repeated 2 more times]
  File "C:/Users/Martin/d/scap-security-guide/shared/utils/yaml-to-shorthand.py", line 647, in add_from_directory
    group = Group.from_yaml(group_file, product_yaml)
  File "C:/Users/Martin/d/scap-security-guide/shared/utils/yaml-to-shorthand.py", line 396, in fro
m_yaml
    yaml_contents = ssgcommon.open_and_macro_expand_yaml(yaml_file, product_yaml)
  File "C:\Users\Martin\d\scap-security-guide\shared\modules\ssgcommon.py", line 362, in open_and_macro_expand_yaml
    return open_and_expand_yaml(yaml_file, substitutions_dict)
  File "C:\Users\Martin\d\scap-security-guide\shared\modules\ssgcommon.py", line 282, in open_and_expand_yaml
    expanded_template = process_file_with_jinja(yaml_file, substitutions_dict)
  File "C:\Users\Martin\d\scap-security-guide\shared\modules\ssgcommon.py", line 255, in process_file_with_jinja
    template = get_jinja_environment().get_template(filepath)
[...]

PR #2987, opened 2018-06-18
Simplified some of the postfix related group IDs

mpreisler:
Some of the postfix group IDs are still ridiculous but this at least gets us below the Windows MAX_PATH of 260.

Before:
```
$ find . | wc -L
227
```

After:
```
$ find . | wc -L
192
```

This gives us about 60 chars margin which is still not great but it's an improvement. The path to the SSG git root can be as long as 68 characters.

fixes: #2982


mpreisler:
Also added an automated test to check max path len. This way we are sure not to reintroduce super long paths again.

tests/ensure_paths_are_short.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python2
```
matejak: Is this because of RHEL6 builds? I think that this is a script that can run both Python 2 and Python 3, but do we want to be backwards-compatible?

matejak:
I see that we repeat `postfix` multiple times in the full filename, couldn't it be avoided somehow? I imagine that the rule filename has to stay, but what about `postfix_server_configuration`? It is already under `postfix` group.

tests/ensure_paths_are_short.py
mpreisler: The script can run with both python2 and python3. What do you suggest the shebang should be?

mpreisler:
> I see that we repeat postfix multiple times in the full filename, couldn't it be avoided somehow? I imagine that the rule filename has to stay, but what about postfix_server_configuration? It is already under postfix group.

Yeah, that is for another pull request though...

matejak:
Sure, ACK.

PR #2971, opened 2018-06-14
Oval510 service disabled

mpreisler:
Unshadowed service_sshd_disabled from all products, offered a better fallback for service disabled in case we don't have OVAL 5.11 features available.

PR #2988, opened 2018-06-19
Fill in the proper definition ID for OVAL 5.10 service disabled

mpreisler:
Fixing a regression that I caused in #2971

Oops..


yuumasato:
What great you review you had.
Ack.

PR #2990, opened 2018-06-19
Remove regex replace in shared/templates/template_common.py

cipherboy:
When @mpreisler's #2971 was merged, it introduced different whitespace formatting in the shared templated OVAL content. The regexes in `services_enabled` and `services_disabled` no longer cleanly applied, breaking the generated OVAL. This removes the regexes, and when packagename is empty (in the CSV), it instead defaults to the servicename. This is the expected behavior per the CSV template file: https://raw.githubusercontent.com/OpenSCAP/scap-security-guide/master/rhel7/templates/csv/services_disabled.csv

This fixes the nightly build. Can test via the following from the root of ssg.

```bash
rm -rf build && mkdir build && cd build && cmake -G "Unix Makefiles" -D CMAKE_BUILD_TYPE=Release -DSSG_TARGET_OVAL_MINOR_VERSION:STRING=10 -DSSG_PRODUCT_CHROMIUM:BOOL=OFF -DSSG_PRODUCT_DEBIAN8:BOOL=OFF -DSSG_PRODUCT_FEDORA:BOOL=OFF -DSSG_PRODUCT_FIREFOX:BOOL=OFF -DSSG_PRODUCT_JBOSS_FUSE6:BOOL=OFF -DSSG_PRODUCT_JRE:BOOL=OFF -DSSG_PRODUCT_OPENSUSE:BOOL=OFF -DSSG_PRODUCT_OSP7:BOOL=OFF -DSSG_PRODUCT_SUSE11:BOOL=OFF -DSSG_PRODUCT_SUSE12:BOOL=OFF -DSSG_PRODUCT_UBUNTU14:BOOL=OFF -DSSG_PRODUCT_UBUNTU16:BOOL=OFF -DSSG_PRODUCT_WRLINUX:BOOL=OFF .. && make rhel7
```


cipherboy:
This works now; some products had ssh packages that didn't use the service name and were hidden as a result.

dahaic:
Thank you Alex for the fix, works great! ACK

PR #2991, opened 2018-06-19
Group files renamed

mpreisler:
A pretty simple refactor, we no longer require group filenames to be $group_id.group, they can be just group.yml now. This is enough because the parent folder's name will match the Group ID.

tl;dr: `group_id/group_id.group` renamed to `group_id/group.yml`

Also renamed `guide.benchmark` to `benchmark.yml` for consistency.


mpreisler:
@openscap-ci test this please

yuumasato:
Nice simplification of the file naming, let's not repeat the ID's every time.

dahaic:
Functionality is preserved, I agree it's nice simplification, ACK and thanks!

PR #2992, opened 2018-06-20
[WIP] Add rule_sssd_ldap_start_tls

mildass:
#### Description:

Add OVAL and remediation script for Configure SSSD LDAP Backend to Use TLS For All Transactions rule


mildass:
There will be a need for a discussion about remediation script, because `ldap_id_use_start_tls` must be in `[domain/...]` group, and these two are separated lines, which can't be matched with grep (it looks that grep matches only lines, not multilines). If someone knows better solution than this pipeing (cat file | sed newlines to '\n' | grep), please suggest it. Otherwise I will make some function for this pipe solution, because we will need it in other SSSD rules.

dahaic:
@mildass definitely rebase this PR, please :)

mildass:
@cipherboy Thanks! I tried`grep -z` and `grep -P` before, but it didn't work for me. I didn't tried combination you mentioned and it works great. :thumbsup: And it's better solution in my opinion.

OVAL and remediation works for me, so it's ready for review

dahaic:
@mildass please, provide also test scenarios, to have full package :) (And it would be great to have Ansible remediation as well)

dahaic:
@mildass I am not sure about using same code you want to test within the test itself. You want to have scenarios build in different way and you want tests to be clear and simple. What about replacing tested configuration files with premade failing/passing ones?

mpreisler:
@openscap-ci test this please

mildass:
@dahaic Ok. I've changed it, that there are passing configs. Tests use them and make just small changes to make it clear what they test.

Issue #2962, opened 2018-06-14
xccdf_org.ssgproject.content_rule_accounts_password_pam_maxrepeat selector="2". DISA STIG RHEL-07-010180 = 3 


amitkumar50:
#### Description of problem:
Rule ID xccdf_org.ssgproject.content_rule_accounts_password_pam_maxrepeat for DISA STIG RHEL-07-010180 has an incorrect value. The description, oval check, and remediation all have maxrepeat = 2, but the DISA STIG says that this should be maxrepeat = 3

#### SCAP Security Guide Version:
scap-security-guide-0.1.36-7.el7.noarch

#### Operating System Version:
rhel-7

#### Steps to Reproduce:
From http://iase.disa.mil/stigs/os/unix-linux/Pages/index.aspx/Red Hat Enterprise Linux 7 STIG - Ver 1, Rel 4/U_Red_Hat_Enterprise_Linux_7_STIG_V1R4_Manual-xccdf.xml
Add the following line to "/etc/security/pwquality.conf conf" (or modify the line to have the required value):^M
^M
maxrepeat = 3The "maxrepeat" option sets the maximum number of allowed same consecutive characters in a new password.

From /usr/share/xml/scap/ssg/content/ssg-rhel7-ds.xml
Profile id="xccdf_org.ssgproject.content_profile_common"
Profile id="xccdf_org.ssgproject.content_profile_stig-rhevh-upstream"
Profile id="xccdf_org.ssgproject.content_profile_ospp-rhel7"
Profile id="xccdf_org.ssgproject.content_profile_nist-800-171-cui"
refine-value idref="xccdf_org.ssgproject.content_value_var_password_pam_maxrepeat" selector="2"/

cipherboy:
This one is an easy fix if you want to open a PR:

- [rhel7/profiles/ospp.profile line 368](https://github.com/OpenSCAP/scap-security-guide/blob/master/rhel7/profiles/ospp.profile#L368)
- [linux_os/guide/system/accounts/accounts-pam/password_quality/password_quality_pwquality/accounts_password_pam_maxrepeat.rule line 28](https://github.com/OpenSCAP/scap-security-guide/blob/master/linux_os/guide/system/accounts/accounts-pam/password_quality/password_quality_pwquality/accounts_password_pam_maxrepeat.rule#L28)

Thanks!

PR #2996, opened 2018-06-20
Fix maxrepeat=2->maxrepeat=3 for DISA/STIG guidelines

cipherboy:
Fixes: #2962

Sets `maxrepeat=3` per STIG guidelines. See #2962 for more information.

### DIFF
--- a/rhel7/profiles/ospp.profile
+++ b/rhel7/profiles/ospp.profile
@@ -365,7 +365,6 @@ selections:
     - var_accounts_tmout=10_min
     - var_password_pam_difok=8
     - var_password_pam_maxclassrepeat=4
-    - var_password_pam_maxrepeat=2
     - var_password_pam_minclass=4
     - var_password_pam_unix_remember=5
     - account_disable_post_pw_expiration
@@ -374,7 +373,6 @@ selections:
     - accounts_minimum_age_login_defs
     - accounts_password_pam_difok
     - accounts_password_pam_maxclassrepeat
-    - accounts_password_pam_maxrepeat
     - accounts_password_pam_minclass
     - accounts_password_pam_unix_remember
     - accounts_password_warn_age_login_defs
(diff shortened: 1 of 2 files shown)


rhel7/profiles/ospp.profile
```diff
@@ -365,7 +365,7 @@ selections:
     - var_accounts_tmout=10_min
     - var_password_pam_difok=8
     - var_password_pam_maxclassrepeat=4
-    - var_password_pam_maxrepeat=2
+    - var_password_pam_maxrepeat=3
```
yuumasato: DISA STIG profile requires `maxrepeat=3`, but I'm not sure about OSPP Policy.
Also, default selector for `var_password_pam_maxrepeat` is 3 already. If OSPP requires it to be 3, this line can be removed.

shawndwells:
On 6/20/18 12:32 PM, Watson Yuuma Sato wrote:
> DISA STIG profile requires |maxrepeat=3|, but I'm not sure about OSPP 
> Policy.
> Also, default selector for |var_password_pam_maxrepeat| is 3 already. 
> If OSPP requires it to be 3, this line can be removed.

``maxrepeat`` is no longer a US Gov requirement. The finalized controls 
for OSPP are listed here:
https://www.niap-ccevs.org/MMO/PP/424.CANX/

Should actually be removed from OSPP profile.

Layered profiles that are agency-specific like the DoD STIG might still 
want the check.

cipherboy:
Updated based on feedback; I removed `accounts_password_pam_maxrepeat` and `var_password_pam_maxrepeat=2` from `rhel7/ospp.profile`

mpreisler:
Hmm, shouldn't it also be in `./stig-rhel7-disa.profile`?

cipherboy:
@mpreisler I'm not sure I follow. `var_password_pam_maxrepeat=3` and `accounts_password_pam_maxrepeat` are both set in that profile.

mpreisler:
ACK

mpreisler:
> var_password_pam_maxrepeat=3 and accounts_password_pam_maxrepeat are both set in that profile.

if you grep for the right thing it's there, if you make typos it's not :P

shawndwells:
With this being removed from OSPP, was it re-added into profiles that extend ospp, such as the 800-171 profile?

linuxdan:
@shawndwells 
I am still getting this from scap-security-guide-0.1.36-10.el7_5.noarch

What version should contain the fix ?

cipherboy:
@linuxdan - On the right, the milestone is `0.1.40` -- so that version will have the fix when it makes its way downstream. :)

linuxdan:
Thanks 
I am still learning all the bells and whistles of this.

PR #2997, opened 2018-06-20
[RFC] Rename utils in build-scripts to remove dash

cipherboy:
Per comment by @matejak on #2949 [here](https://github.com/OpenSCAP/scap-security-guide/pull/2949#discussion_r195127535), it seems like there's some preference for underscores versus dashes in the file name. At any rate, the project is not consistent in the naming of utilities and scripts, but IMO, should be.

This renames the scripts to use underscores and updates the build system correspondingly. 

Dependency: #2967 -- merged
Part of: #2949


redhatrises:
We were actually moving to use dash and not underscores in general. Not sure why we are moving back to underscores?

mpreisler:
> We were actually moving to use dash and not underscores in general. Not sure why we are moving back to underscores?

We were but we didn't realize that you can't *easily* (without the __import__ trick) import python files with dashes in them. This is a pragmatic move to make everything testable by unit tests. Plus, personally, I don't care which convention is used as long as it's consistent.

cipherboy:
Back on 4459627a033ce091ec85ebc28b1c7b49123d52c0 (pre-internship starting), there were 40 python files with `_` in it and 19 with `-`; most of the 40 are templates or modules with fewer are utils or misc (and all of the ones with `-` are scripts), but as a whole, I think it makes sense to unify _everything_ to a single naming convention.

mpreisler:
ACK

discussed this with @redhatrises off github and he's OK with it

PR #2993, opened 2018-06-20
Unshadow accounts password pam fixes

yuumasato:
#### Description:

- Drop accounts_password_pam fixes from shared
- Add CSV file for RHEL6
- Unify Ansible and Bash templates for all RHEL and Fedora using jinja
- Minor regex improvement in Ansible template
- Drop unnecessary CSV for SLE12

#### Rationale:

- Un-shadows accounts_password_pam for RHEL6, and RHEL7.

PR #2998, opened 2018-06-20
Add metadata to accounts_password fixes templates

yuumasato:
#### Description:

- Add metadata for remediations
- Inspired on remediations dropped as part of #2993

#### Rationale:

- Metadata on fixes can help filter them, specially on Ansible fixes.


mpreisler:
ACK

PR #3000, opened 2018-06-20
Made OVAL parse error more user friendly.

dahaic:
#### Description:

- When OVAL was malformed, pretty lame error (traceback) has been reported.

#### Rationale:

- By making proper output, user is faster to realize the problem.

### DIFF
--- a/ssg/_build_ovals.py
+++ b/ssg/_build_ovals.py
@@ -14,7 +14,7 @@ except ImportError:
     from ConfigParser import SafeConfigParser
 
 from ssg._constants import oval_namespace as oval_ns
-from ssg._constants import oval_footer as footer
+from ssg._constants import oval_footer
 from ssg._constants import oval_header
 from ssg._products import parse_name, multi_list, map_name
 from ssg._jinja import process_file
@@ -242,7 +242,18 @@ def _check_is_loaded(loaded_dict, filename, version):
 
 
 def _check_oval_version_from_oval(xml_content, oval_version):
-    oval_file_tree = ElementTree.fromstring(oval_header + xml_content + footer)
+    try:
+        argument = oval_header + xml_content + oval_footer
+        oval_file_tree = ElementTree.fromstring(argument)
+    except ElementTree.ParseError, p:
+        line, column = p.position
+        lines = argument.splitlines()
+        before = '\n'.join(lines[:line])
+        column_pointer = ' ' * (column - 1) + '^'
+        sys.stderr.write(
+            "%s\n%s\nError when parsing OVAL file.\n" %
+            (before, column_pointer))
+        sys.exit(1)
     for defgroup in oval_file_tree.findall("./{%s}def-group" % oval_ns):
         file_oval_version = defgroup.get("oval_version")
 


dahaic:
New output (snipped just for this comment):
```
[ 10%] Built target generate-internal-rhel7-xccdf-unlinked.xml

<oval_definitions
    xmlns="http://oval.mitre.org/XMLSchema/oval-definitions-5"
    xmlns:oval="http://oval.mitre.org/XMLSchema/oval-common-5"
    xmlns:ind="http://oval.mitre.org/XMLSchema/oval-definitions-5#independent"
    xmlns:unix="http://oval.mitre.org/XMLSchema/oval-definitions-5#unix"
    xmlns:linux="http://oval.mitre.org/XMLSchema/oval-definitions-5#linux"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
<SNIP>
  <ind:textfilecontent54_object id="object_use_starttls_sssd_conf" />
    <ind:pattern operation="pattern match">^[\s]*\[domain\/[^]]*]([^\n\[\]]*\n+)+?[\s]*ldap_id_use_start_tls[ ]*=[ ]*((?i)true)[ ]*$</ind:pattern>
   ^
Error when parsing OVAL file.
make[3]: *** [rhel7/oval-unlinked.xml] Error 1
...
```

cipherboy:
@dahaic I like this. Could we perhaps also print the called command line? When debugging the nightly build, having the command that failed would've been nice: I had already built the project, I just need to rerun the failing command. Only issue with that is escaping... Perhaps an addition to `ssg.utils`

ssg/_build_ovals.py
```diff
@@ -242,7 +242,18 @@ def _check_is_loaded(loaded_dict, filename, version):
...
 def _check_oval_version_from_oval(xml_content, oval_version):
-    oval_file_tree = ElementTree.fromstring(oval_header + xml_content + footer)
+    try:
+        argument = oval_header + xml_content + oval_footer
+        oval_file_tree = ElementTree.fromstring(argument)
+    except ElementTree.ParseError, p:
+        line, column = p.position
+        lines = argument.splitlines()
+        before = '\n'.join(lines[:line])
+        column_pointer = ' ' * (column - 1) + '^'
```
mpreisler: fancy

ssg/_build_ovals.py
```diff
@@ -242,7 +242,18 @@ def _check_is_loaded(loaded_dict, filename, version):
...
+        sys.stderr.write(
```
mpreisler: I would personally use print("blabla", file=sys.stderr) because it will do the final newline for you but it's not a big deal.

redhatrises:
I like this as well... Do you think that we could make the output more obvious so that it is more eye catching?

mpreisler:
> I just need to rerun the failing command. Only issue with that is escaping... Perhaps an addition to ssg.utils

interesting idea. we could create our own exception, something like SSGInputException and it could just output `" ".join(sys.argv)` as part of the exception message. That way you can just fix the input and rerun that command it printed to verify that you fixed it.

dahaic:
@cipherboy even though I see benefit in knowing the command line, I have been debugging it by simply reissuing `make` command, which was no hassle in my case (as it took just few seconds before failing again). I am not sure, if average content creator needs to be exposed to the internals of SSG build system :)

ssg/_build_ovals.py
dahaic: Yeah, it makes sense, it's just that `sys.stderr.write` is used across the file. I am fine with changing it everywhere if I change it here :) Should I?

cipherboy:
> interesting idea. we could create our own exception, something like SSGInputException and it could just output " ".join(sys.argv) as part of the exception message. That way you can just fix the input and rerun that command it printed to verify that you fixed it.

I was explicitly avoiding mentioning `" ".join(sys.argv)` for the reason that inputs can be quoted on the command line; these quotes are not persisted into `sys.argv` and thus `" ".join(sys.argv)` will not work.

E.g., calling `./some-script.py "/path/with spaces"` would output `./some-script.py /path/with spaces`; copying and pasting that will not work. I suppose that, since #2949, calling the script is unlikely to work (as we'd need to detect environment variables like `PYTHONPATH` and output those too), I'll agree with @dahaic.

cipherboy:
I'm ok with this PR as is. I'll be changing everything to be more consistent anyways, so IMO, I'd just leave the `sys.stderr.write` and I'll get it as a separate PR / as part of cleanup.

dahaic:
@redhatrises I have updated the output to contain all the follow-up lines. I think it's more than enough for curious contributor to see were is the problem, as all the lines that follows are make exits :)

mpreisler:
ACK

PR #3001, opened 2018-06-20
Use the generated sysctl_kernel_randomize_va_space instead of an over…

mpreisler:
…ride

The override isn't any different.

More unshadowing. Most of the easy pickins are gone :-(

This is what the generated `sysctl_kernel_randomize_va_space.sh` looks like:
```
# platform = multi_platform_rhel
# reboot = true
# strategy = disable
# complexity = low
# disruption = medium

#
# Set runtime for kernel.randomize_va_space
#
/sbin/sysctl -q -n -w kernel.randomize_va_space=2

#
# If kernel.randomize_va_space present in /etc/sysctl.conf, change value to "2"
#	else, add "kernel.randomize_va_space = 2" to /etc/sysctl.conf
#
replace_or_append '/etc/sysctl.conf' '^kernel.randomize_va_space' "2" '@CCENUM@'
```


redhatrises:
Ack.

PR #3002, opened 2018-06-20
Fix python3 regression from #3000

cipherboy:
This is a regression introduced in #3000. Comma for naming the resulting exception is valid syntax in python2, but does not work in python3.


mpreisler:
ACK

PR #3004, opened 2018-06-20
There is no need to set the sshd_listening_port to "default"

mpreisler:
Because it will be "default" when not set to anything...

This avoids errors such as:
```
OpenSCAP Error: Invalid selector 'default' for xccdf:value/@id='sshd_listening_port'. Using null value instead. [xccdf_policy.c:2103]
Could not resolve xccdf:sub/@idref='sshd_listening_port'! [xccdf_policy_substitute.c:107]
A fix for Rule/@id="firewalld_sshd_port_enabled" was skipped: Text substitution failed. [xccdf_policy_remediate.c:552]
```

This is a regression caused by #2933

### DIFF
--- a/eap6/profiles/stig-eap6-disa.profile
+++ b/eap6/profiles/stig-eap6-disa.profile
@@ -6,7 +6,6 @@ description: 'This is a *draft* profile for STIG. This profile is being develope
     a STIG in coordination with DISA FSO.'
 
 selections:
-    - var_jboss_profile=default
     - jboss_eap_configure_secure_management_access
     - jboss_eap_configure_https
     - jboss_eap_configure_host_access_restrictions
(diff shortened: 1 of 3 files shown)


cipherboy:
@mpreisler Is this a universal thing? Do  the other occurrences of `default` need to be removed?
 - `firefox/profiles/stig-firefox-upstream.profile:21` 
 - `eap6/profiles/stig-eap6-disa.profile:9`

mpreisler:
> Is this a universal thing? Do the other occurrences of default need to be removed?
> 
> firefox/profiles/stig-firefox-upstream.profile:21
> eap6/profiles/stig-eap6-disa.profile:9

good catch, I will also fix those

redhatrises:
Ack

PR #3006, opened 2018-06-20
Remove all affected elements from final OVALs

mpreisler:
In the SSG use-case we don't expect users to use the definitions
directly, therefore they will use the XCCDF's CPE dict capability
instead of OVAL's affected element. The affected element is thus not
useful and can be removed, shrinking the OVAL file sizes quite a bit.

Size before change:
```
2.6M    ssg-rhel7-oval.xml
```

Size after change:
```
2.2M    ssg-rhel7-oval.xml
```

This may be controversial, so let's get firm consensus here before we merge. We have heard that some tools may be using the affected element. If they are there will be no regression because no affected element means the definition applies to all platforms so it will still all work. I doubt it's used for applicability though because the affected platform names are not CPE IDs. They are human readable long string names, not suitable for machine processing.


redhatrises:
I personally say nack to this. The tooling can be updated to make this less painful IMO. CIS and others use platform to sort and validate content. Also if running an OVAL scan, are XCCDF CPEs taken into account?

mpreisler:
> Also if running an OVAL scan,

You can't run an OVAL scan with SSG content, you need variables exported from XCCDF. You export them to a file and then use that file for an OVAL scan but personally I doubt people are actually using this. Plus as I said, it will still all work, according to spec no affected elements means it's all applicable everywhere.

Could you paste a link about what CIS and others are doing? Maybe it would help me understand the use-case.

mpreisler:
Here is an example of what happens right now, in ssg-rhel7-oval.xml:
```
    <ns0:definition class="compliance" id="oval:ssg-package_uuidd_installed:def:1" version="1">
      <ns0:metadata>
        <ns0:title>Package uuidd Installed</ns0:title>
        <ns0:affected family="unix">
          <ns0:platform>Oracle Linux 7</ns0:platform>
          <ns0:platform>Wind River Linux 8</ns0:platform>
          <ns0:platform>Ubuntu 1404</ns0:platform>
          <ns0:platform>Ubuntu 1604</ns0:platform>
          <ns0:platform>Debian 8</ns0:platform>
          <ns0:platform>Red Hat Enterprise Linux 7</ns0:platform>
          <ns0:platform>Red Hat Enterprise Linux 6</ns0:platform>
          <ns0:platform>Fedora 26</ns0:platform>
          <ns0:platform>Fedora 27</ns0:platform>
          <ns0:platform>Fedora 28</ns0:platform>
        </ns0:affected>
        <ns0:description>The RPM package uuidd should be installed.</ns0:description>
        <reference ref_id="package_uuidd_installed" source="ssg"/>
      </ns0:metadata>
      <ns0:criteria>
        <ns0:criterion comment="package uuidd is installed" test_ref="oval:ssg-test_package_uuidd_installed:tst:1"/>
      </ns0:criteria>
    </ns0:definition>
```

I doubt anybody can use that for any sorting...

dahaic:
Functionality like container scanning is strongly dependent on using datastreams (or to be precise XCCDF with SSG CPE dictionary). We have broken assumption of separately consumable OVALs long time ago, AFAIK. (Providing SCAP content is not easy, so making our lives even harder by supporting OVAL is not desirable, IMO)

So I am all for removing the `affected` element. :)

mildass:
I have this OVAL
```
<def-group>
  <definition class="compliance" id="sssd_ldap_start_tls" version="1">
    <metadata>
      <title>Configure SSSD LDAP Backend to Use TLS For All Transactions</title>
      <affected family="unix">
      </affected>
      <description>LDAP should be used for authentication and use STARTTLS</description>
    </metadata>
    <criteria operator="AND">
      <criterion comment="Using LDAP for authentication set within /etc/sysconfig/authconfig" test_ref="test_use_ldap_authentication" />
      <criterion comment="LDAP uses STARTTLS set within /etc/sssd/sssd.conf" test_ref="test_use_starttls" />
    </criteria>
  </definition>

  <ind:textfilecontent54_test check="all" check_existence="all_exist"
  comment="Ensures that LDAP is being used for authentication"
  id="test_use_ldap_authentication" version="1">
    <ind:object object_ref="object_use_ldap_authentication_authconfig" />
  </ind:textfilecontent54_test>

  <ind:textfilecontent54_test check="all" check_existence="all_exist"
  comment="Ensures that LDAP uses STARTTLS"
  id="test_use_starttls" version="1">
    <ind:object object_ref="object_use_starttls_sssd_conf" />
  </ind:textfilecontent54_test>

  <ind:textfilecontent54_object id="object_use_ldap_authentication_authconfig" version="1">
    <ind:filepath>/etc/sysconfig/authconfig</ind:filepath>
    <ind:pattern operation="pattern match">^USELDAPAUTH=((?i)yes)[ ]*$</ind:pattern>

  <ind:textfilecontent54_object id="object_use_starttls_sssd_conf" version="1">
    <ind:pattern operation="pattern match">^[\s]*\[domain\/[^]]*]([^\n\[\]]*\n+)+?[\s]*ldap_id_use_start_tls[ ]*=[ ]*((?i)true)[ ]*$</ind:pattern>
</def-group>
```
and oscap doesn't check it when there aren't `affected` and specified `platform` tags. What am I missing? Because with affected it works.

mpreisler:
> and oscap doesn't check it when there aren't affected and specified platform tags. What am I missing? Because with affected it works.

This is not the right place to ask about this IMO, please consider using the mailing list in the future for support questions. However, as far as I know, openscap doesn't do anything with the affected element. It loads it and stores it but the contents are never processed.

yuumasato:
I think we mostly use it to filter which definitions from shared folder should be fetched into the oval files. And IMO it is awkward to have lots of unrelated platforms in an oval file.

My opinion is that we can keep the affected element empty. (It was empty for some time due to a bug, and it implied applicability to all platforms)
Or keep only the affected element for the platform we are building the OVAL file for.

redhatrises:
> Or keep only the affected element for the platform we are building the OVAL file for.

Yes. Keep only the affected element for the platform we are building the OVAL file for. That way we support any and all downstream uses... and don't get a black eye like we did last time for removing the ability to create packages.

> (Providing SCAP content is not easy, so making our lives even harder by supporting OVAL is not desirable, IMO)

Going to have to support it for a long time, and more is coming down the pipe.

jan-cerny:
I like @mildass point that it is not processed by OpenSCAP at all. I also agree that plain OVAL is useless in SSG. 

@redhatrises I would like to see how this element is used  "to sort and validate content".

mpreisler:
Here is a suggestion for a compromise. Instead of removing affected altogether I could change the code to remove all child platforms and then add one child platform element of the platform we are building for. Would that be acceptable, @redhatrises ?

mpreisler:
> Because I created OVAL check for rule and oscap doesn't check it when there isn't affected tag:

Please triple-check that it's not something else. Attach oscap to a debugger a find why this is happening. If what you are saying is true then we have a serious regression in oscap. This worked when we were certifying for SCAP 1.2.

mpreisler:
> Please! 1000%

This is now done, ready for your re-review

redhatrises:
ACK. Really appreciate your work and compromise on this @mpreisler!

PR #3007, opened 2018-06-20
CCE not being set in an XCCDF does not warrant a warning

mpreisler:
We can always get CCE stats using `make rhel7-stats`. If we need a list
of rules for allocating CCEs we should implement better tooling that
tells you in which .rule file the CCE is missing. This is too late in
the build to be checking for missing CCEs.


redhatrises:
Ack

PR #3008, opened 2018-06-21
[DO NOT MERGE] Unshadow sle12

matejak:
* Removed avahi fix from sle12 template.
* Removed unused create_sockets_disabled template generator.


sle12/templates/csv/services_disabled.csv
```diff
@@ -3,7 +3,6 @@ abrtd,abrt,
 acpid,,
 atd,at,
 autofs,,
-avahi-daemon,avahi,
```
yuumasato: This will also stop generation of templated OVAL and Bash for service_avahi-daemon_disabled.

yuumasato: Workaround I see are, also moving OVAL and Bash to shared folder, or add another column limiting generation via templates for target languages.

matejak: I see, another possibilities are:
* Make the avahi remediation to use the template. It would lose the "disable socket" functionality that way.
* Make service disable remediations more generic, so they can optionally contain socket disable Ansible remediation.

mpreisler:
NACK

we need to keep the socket service deactivation. I think we should add it to the service disabled templates.

mpreisler:
Based on discussing with @matejak and the rest of the team I believe this should be closed.

matejak:
This PR is legitimate, but only after #3029, #3025 and successor of #3030 are merged.

yuumasato:
But shouldn't we move the other direction?
Drop the fix from shared, and add to CSVs?

matejak:
@yuumasato You are right, it should be the other way around. So I close this for good :-)

PR #3009, opened 2018-06-21
Unshadow RHEL6 audit_rules_dac_modification

yuumasato:
#### Description:

- Unify templates OVAL and Bash templates for audit_rules_dac_modification

#### Rationale:

- Main reason is to unshadow these for RHEL6

#### Notes
- OVAL for RHEL6 remains unchanged
- The Bash fixes were slight changed by templating, as each was unique.
  - The `GROUP` or for RHEL6 fixes varied between `chmod`, `chown` and `xattr`, depending on rule, now it is all the same `perm_mod`.
  - RHEL6 had identical remediations for a group, for example, the following fixes for `xattr` were identical:
```
rhel6/fixes/bash/audit_rules_dac_modification_removexattr.sh
rhel6/fixes/bash/audit_rules_dac_modification_lremovexattr.sh
rhel6/fixes/bash/audit_rules_dac_modification_fremovexattr.sh
rhel6/fixes/bash/audit_rules_dac_modification_fsetxattr.sh
rhel6/fixes/bash/audit_rules_dac_modification_setxattr.sh
rhel6/fixes/bash/audit_rules_dac_modification_lsetxattr.sh
```
  Now they are individual, each fixing a specific audit rule.


yuumasato:
I'm not sure if changes in Bash remediation are ok.I would appreciate feedback on the notes in PR description.

shared/templates/template_BASH_audit_rules_dac_modification
```diff
@@ -11,8 +11,15 @@ for ARCH in "${RULE_ARCHS[@]}"
 do
 	PATTERN="-a always,exit -F arch=$ARCH -S %ATTR%.*"
 	GROUP="perm_mod"
+{{% if product != "rhel6" %}}
 	FULL_RULE="-a always,exit -F arch=$ARCH -S %ATTR% -F auid>=1000 -F auid!=4294967295 -F key=perm_mod"
+{{% else %}}
+	FULL_RULE="-a always,exit -F arch=$ARCH -S %ATTR% -F auid>=500 -F auid!=4294967295 -k perm_mod"
```
mpreisler: aren't `-F key=value` and -k value` equivalent? or is it a legit difference between rhel6 and 7?

Assuming they are the same, could you please instead set a variable for the auid and just do a condition around that?

shared/templates/template_BASH_audit_rules_dac_modification
```diff
@@ -11,8 +11,15 @@ for ARCH in "${RULE_ARCHS[@]}"
...
+{{% endif %}}
+
+{{% if product != "rhel6" %}}
 	# Perform the remediation for both possible tools: 'auditctl' and 'augenrules'
-	fix_audit_syscall_rule "auditctl" "$PATTERN" "$GROUP" "$ARCH" "$FULL_RULE"
 	fix_audit_syscall_rule "augenrules" "$PATTERN" "$GROUP" "$ARCH" "$FULL_RULE"
```
mpreisler: same here, one little part of it is different, we shouldn't repeat the whole line because of it.

redhatrises:
Both augenrules and `-F key=` are supported on RHEL6. A new PR should be open to handle augenrules in rhel6.

redhatrises:
@yuumasato do you want to submit another PR to fix these rules for rhel6 as well?

yuumasato:
@redhatrises done in https://github.com/OpenSCAP/scap-security-guide/pull/3023

PR #3010, opened 2018-06-21
Improve CMake time by building all remediations at once

cipherboy:
This restructures `generate_from_templates` to accept both a list of inputs/outputs and a list of languages; for each language and each input/output pair, it performs its original behavior. This allows for fewer calls to `generate_from_templates.py`, improving build time.

Before fix:
```
cmake: 34s / 31s (real/user)
ninja: 1m35.891s / 4m37.459s (real/user)
```

After fix:
```
cmake: 17s / 13s (real/user)
ninja: 1m29.117s / 4m15.623s (real/user)
```

### DIFF
--- a/build-scripts/generate_from_templates.py
+++ b/build-scripts/generate_from_templates.py
@@ -22,29 +22,34 @@ def parse_args():
     output_sp = sp.add_parser('list-outputs', help="Generate output list")
     output_sp.set_defaults(cmd="list_outputs")
 
-    p.add_argument('--language', metavar="LANG", default=None,
-                   help="Scripts of which language should we generate? "
-                   "Default: all.")
-    p.add_argument("-i", "--input", action="store", required=True,
-                   help="input directory")
-    p.add_argument("-o", "--output", action="store", required=True,
-                   help="output directory")
-    p.add_argument("-s", "--shared", metavar="PATH", required=True,
+    p.add_argument('--languages', metavar="LANGS", required=True, type=str,
+                   nargs='+', help="List of languages to generate")
+    p.add_argument("--input", required=True, type=str, nargs='+',
+                   help="List of input directories")
+    p.add_argument("--output", required=True, type=str, nargs='+',
+                   help="List of output directories")
+    p.add_argument("--shared", metavar="PATH", required=True,
                    help="Full absolute path to SSG shared directory")
 
-    return p.parse_args()
+    args = p.parse_args()
+    if len(args.input) != len(args.output):
+        print("Error: number of inputs and number of outputs must match!",
+              file=sys.stderr)
+        sys.exit(1)
+
+    return args
 
 
 if __name__ == "__main__":
     args = parse_args()
 
-    builder = ssg.build_templates.Builder()
-    if args.language is not None:
-        builder.set_langs([args.language])
+    for index in range(0, len(args.input)):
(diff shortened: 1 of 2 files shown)


cipherboy:
Would appreciate a check to make sure this doesn't remove any functionality.

build-scripts/generate_from_templates.py
```diff
@@ -22,29 +22,36 @@ def parse_args():
     output_sp = sp.add_parser('list-outputs', help="Generate output list")
     output_sp.set_defaults(cmd="list_outputs")
 
-    p.add_argument('--language', metavar="LANG", default=None,
-                   help="Scripts of which language should we generate? "
+    p.add_argument('--languages', metavar="LANGS", required=True, type=str, nargs='+',
+                   help="List of languages to generate"
                    "Default: all.")
```
mpreisler: it says required=True, so there is no default

cipherboy:
@mpreisler Fair catch. Updated the PR. :)

cipherboy:
@openscap-ci test this please

cmake/SSGCommon.cmake
```diff
@@ -153,57 +153,50 @@ macro(ssg_build_ocil_unlinked PRODUCT)
...
+          DEPENDS "${SSG_BUILD_SCRIPTS}/combine_remediations.py"
+          COMMENT "[${PRODUCT}-content] generating ${LANGUAGE}-fixes.xml"
+      )
+      add_custom_target(
+          generate-internal-${PRODUCT}-${LANGUAGE}-fixes.xml
+          DEPENDS "${CMAKE_CURRENT_BINARY_DIR}/${LANGUAGE}-fixes.xml"
+      )
+    endforeach()
 
 if (SSG_SHELLCHECK_BASH_FIXES_VALIDATION_ENABLED AND SHELLCHECK_EXECUTABLE AND "${LANGUAGE}" STREQUAL "bash")
```
mpreisler: this doesn't work anymore, right? there is no LANGUAGE in this macro

mpreisler: maybe we can remove the shellcheck test completely because it never really did anything useful. It would be much more useful to shellcheck the final remediation roles.

cipherboy: I'll remove them,  good catch. Otherwise, I could move it up as part of the foreach loop.

cipherboy:
@mpreisler Updated!

build-scripts/generate_from_templates.py
```diff
@@ -22,29 +22,35 @@ def parse_args():
...
 if __name__ == "__main__":
     args = parse_args()
 
-    builder = ssg.build_templates.Builder()
-    if args.language is not None:
-        builder.set_langs([args.language])
+    for language in args.languages:
+        for index in range(0, len(args.input)):
+            builder = ssg.build_templates.Builder()
+            builder.set_langs([language])
```
mpreisler: I am wondering why not just use one builder and set_langs(languages)? Is it broken?

cipherboy: @mpreisler Ah, weird. I thought that feature was broken when we were debugging the CMake issues yesterday, but I guess it worked locally today. Weird? Updated. :)

mpreisler:
ACK

PR #3011, opened 2018-06-21
Fix worker join

cipherboy:
When using `queue.join()`, often the python interpreter will start exiting before the workers themselves exit. This fixes that by waiting for all workers to join. This prevents the warning about `NoneType` not having the exception type that we're catching in the workers.

This PR also removes the informational messages during the building of roles and guides so real warnings can be seen.

### DIFF
--- a/build-scripts/build_all_guides.py
+++ b/build-scripts/build_all_guides.py
@@ -90,7 +90,8 @@ def main():
         worker.daemon = True
         worker.start()
 
-    queue.join()
+    for worker in workers:
+        worker.join()
 
     index_source = ssg.build_guides.build_index(benchmarks, input_basename,
                                                 index_links, index_options,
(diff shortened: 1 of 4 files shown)


build-scripts/build_all_guides.py
```diff
@@ -90,7 +91,8 @@ def main():
         worker.daemon = True
         worker.start()
 
-    queue.join()
+    for worker in workers:
+        worker.join()
```
mpreisler: shouldn't we also join the queue?

cipherboy: If all workers exit and the queue has not joined, then we risk running forever. However, you are correct that the logic in the workers called `queue.task_done()`; this wouldn't work under this system --and would leave the workers running--so I changed it to `break`.

cipherboy: This might behave a little more optimally now. I also cleared the queue and left the `queue.task_done()` and the `break`.

PR #3012, opened 2018-06-21
add one more condition for user oracle in bash to keep consistent wit…

xiruiyang:
…h OVAL test accounts_authorized_local_users_sidadm_orasid

	modified:   shared/fixes/bash/accounts_authorized_local_users_sidadm_orasid.sh

#### Description:

- added one more condition "if /oracle/SID exist then add oracle to userlist". This is not necessary if the SAP system is set properly. But it could happen that the owner of brspace was improperly set to other user when it should be owned by oracle. Especially after a SAP kernel update either manually or via an old script, such kind of mistake could happen. This condition is to reduce the chance of one mistake causing another mistake.

### DIFF
--- a/shared/fixes/bash/accounts_authorized_local_users_sidadm_orasid.sh
+++ b/shared/fixes/bash/accounts_authorized_local_users_sidadm_orasid.sh
@@ -8,6 +8,21 @@ default_os_user="root"
 # add users sidamd, orasid, sapadm and oracle if needed
 userlist="root"
 sapmnt_SID_stem="/sapmnt/[A-Z][A-Z0-9][A-Z0-9]"
+oracle_SID_stem="/oracle/[A-Z][A-Z0-9][A-Z0-9]"
+
+# if owner of any directory or file in the given list is the user oracle, return 1.
+# otherwise return 0. 
+# Usage: is_owner_oracle "${list[@]}"
+function is_owner_oracle {
+	local path_list=("$@")
+	local is_oracle=0
+	for path in $path_list ; do
+		if [ $(ls -ld $path | awk '{print $3}') = "oracle" ]; then
+			is_oracle=1
+		fi
+	done
+	echo "$is_oracle"
+} 
 
 # if /sapmnt is a directory or a symbolic link to a directory,
 # then try to add SAP system users to the userlist
@@ -25,23 +40,22 @@ if [ -d "/sapmnt" ] ; then
 
 	# if brspace exist in any of the above directory of a SID, add orasid to the userlist 
 	for path_to_brspace in $path_to_brspace_list ; do
-        	SID=${path_to_brspace:8:3}
-        	userlist="$userlist|ora$(echo "$SID" | sed -e 's/\(.*\)/\L\1/')"
+		SID=${path_to_brspace:8:3}
+		userlist="$userlist|ora$(echo "$SID" | sed -e 's/\(.*\)/\L\1/')"
 	done
 
 	# if owner of any brspace file is oracle, add oracle to the userlist
-	oracle=false
-	for path_to_brspace in $path_to_brspace_list ; do
-        	if [ $(ls -ld $path_to_brspace | awk '{print $3}') = "oracle" ]; then
-                	oracle=true
-        	fi
-	done
-	if test "$oracle" = true ; then
-        	userlist="$userlist|oracle"
-	fi
+	if test "$(is_owner_oracle "${path_to_brspace_list[@]}")" = 1 ; then userlist="$userlist|oracle" ; fi
 fi
 
-# if /usr/sap/hostctrl is a directory or a symbolic linkd to a directory, add sapadm to the list
+# if owner of any /oracle/SID directory is oracle, add oracle to the userlist
+# the user oracle could be added twice in the userlist, but it is harmlos to the final result
+if [ -d "/oracle" ] ; then
+	path_oracle_SID_list=$(find /oracle/ -regex "^$oracle_SID_stem$")
+	if test "$(is_owner_oracle "${path_oracle_SID_list[@]}")" = 1 ; then userlist="$userlist|oracle" ; fi
+fi
+
+# if /usr/sap/hostctrl is a directory or a symbolic link to a directory, add sapadm to the list
 if [ -d /usr/sap/hostctrl ] ; then
 	userlist="$userlist|sapadm"
 fi
--- a/shared/fixes/bash/accounts_authorized_local_users_sidadm_orasid.sh
+++ b/shared/fixes/bash/accounts_authorized_local_users_sidadm_orasid.sh
@@ -12,10 +12,10 @@ oracle_SID_stem="/oracle/[A-Z][A-Z0-9][A-Z0-9]"
 
 # if owner of any directory or file in the given list is the user oracle,
 # add the user oracle to the variable userlist. 
-# Usage: verify_oracle_user_to_userlist "${array[@]}"
+# Usage: verify_oracle_user_to_userlist "$path_list"
 # Note: this function might modify the value of the global variable userlist
 function verify_oracle_user_to_userlist {
-	local path_list=("$@")
+	local path_list="$1"
 	local is_oracle=no
 	for path in $path_list ; do
 		if [ $(ls -ld "$path" | awk '{print $3}') = "oracle" ]; then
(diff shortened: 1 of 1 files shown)


xiruiyang:
@dahaic @matejak another small change was made for the bash remediation of accounts_authorized_local_users_sidadm_orasid, could you please take a look again? Thanks.

shared/fixes/bash/accounts_authorized_local_users_sidadm_orasid.sh
```diff
@@ -8,6 +8,21 @@ default_os_user="root"
...
 sapmnt_SID_stem="/sapmnt/[A-Z][A-Z0-9][A-Z0-9]"
+oracle_SID_stem="/oracle/[A-Z][A-Z0-9][A-Z0-9]"
+
+# if owner of any directory or file in the given list is the user oracle, return 1.
+# otherwise return 0. 
+# Usage: is_owner_oracle "${list[@]}"
+function is_owner_oracle {
+	local path_list=("$@")
+	local is_oracle=0
+	for path in $path_list ; do
```
matejak: I appreciate usage of local variables. However, you probably meant `"${path_list[@]}"` here.

shared/fixes/bash/accounts_authorized_local_users_sidadm_orasid.sh
```diff
@@ -8,6 +8,21 @@ default_os_user="root"
...
 userlist="root"
 sapmnt_SID_stem="/sapmnt/[A-Z][A-Z0-9][A-Z0-9]"
+oracle_SID_stem="/oracle/[A-Z][A-Z0-9][A-Z0-9]"
+
+# if owner of any directory or file in the given list is the user oracle, return 1.
+# otherwise return 0. 
+# Usage: is_owner_oracle "${list[@]}"
+function is_owner_oracle {
+	local path_list=("$@")
+	local is_oracle=0
```
matejak: You use 0 and 1, which may be slightly confusing (e.g. commands `true` returns 0 and `false` returns 1). I incline to use "yes" and "no" instead of numbers.

shared/fixes/bash/accounts_authorized_local_users_sidadm_orasid.sh
```diff
@@ -25,23 +40,22 @@ if [ -d "/sapmnt" ] ; then
...
-        	userlist="$userlist|oracle"
-	fi
+	if test "$(is_owner_oracle "${path_to_brspace_list[@]}")" = 1 ; then userlist="$userlist|oracle" ; fi
+fi
+
+# if owner of any /oracle/SID directory is oracle, add oracle to the userlist
+# the user oracle could be added twice in the userlist, but it is harmlos to the final result
+if [ -d "/oracle" ] ; then
+	path_oracle_SID_list=$(find /oracle/ -regex "^$oracle_SID_stem$")
+	if test "$(is_owner_oracle "${path_oracle_SID_list[@]}")" = 1 ; then userlist="$userlist|oracle" ; fi
```
matejak: I see that we have a duplicated line here, what about extracting it to a separate function that would accept array as a parameter?

shared/fixes/bash/accounts_authorized_local_users_sidadm_orasid.sh
xiruiyang: Good suggestion, I will modify 0 and 1 to no and yes. Thanks.

shared/fixes/bash/accounts_authorized_local_users_sidadm_orasid.sh
xiruiyang: I was trying to avoid modify the value of global variable inside a function, which looks more dangerous than repeated code lines to me. Anyway, I moved the repeated lines to the function above, you may choose which one you prefer :-)  Thanks.

shared/fixes/bash/accounts_authorized_local_users_sidadm_orasid.sh
matejak: I have reviewed your code again and I see the issue - `path_to_brspace_list` is a space-delimited string, not an array. It is OK to keep it as such if you are 100% sure that individual filenames don't contain spaces.
I assume that they don't, so you just need to convert array handling to string handling, I have posted new comments where applicable.

shared/fixes/bash/accounts_authorized_local_users_sidadm_orasid.sh
```diff
@@ -10,18 +10,21 @@ userlist="root"
...
-# if owner of any directory or file in the given list is the user oracle, return 1.
-# otherwise return 0. 
-# Usage: is_owner_oracle "${list[@]}"
-function is_owner_oracle {
+# if owner of any directory or file in the given list is the user oracle,
+# add the user oracle to the variable userlist. 
+# Usage: verify_oracle_user_to_userlist "${array[@]}"
+# Note: this function might modify the value of the global variable userlist
+function verify_oracle_user_to_userlist {
 	local path_list=("$@")
```
matejak: You get a string on input, so write `path_list="$1"`

shared/fixes/bash/accounts_authorized_local_users_sidadm_orasid.sh
```diff
@@ -45,14 +48,14 @@ if [ -d "/sapmnt" ] ; then
 	done
 
 	# if owner of any brspace file is oracle, add oracle to the userlist
-	if test "$(is_owner_oracle "${path_to_brspace_list[@]}")" = 1 ; then userlist="$userlist|oracle" ; fi
```
matejak: `path_oracle_SID_list` is space-separated string, so pass just "$path_to_brspace_list".

shared/fixes/bash/accounts_authorized_local_users_sidadm_orasid.sh
```diff
@@ -45,14 +48,14 @@ if [ -d "/sapmnt" ] ; then
...
+	verify_oracle_user_to_userlist "${path_to_brspace_list[@]}"
 fi
 
 # if owner of any /oracle/SID directory is oracle, add oracle to the userlist
 # the user oracle could be added twice in the userlist, but it is harmlos to the final result
 if [ -d "/oracle" ] ; then
 	path_oracle_SID_list=$(find /oracle/ -regex "^$oracle_SID_stem$")
-	if test "$(is_owner_oracle "${path_oracle_SID_list[@]}")" = 1 ; then userlist="$userlist|oracle" ; fi
```
matejak: ditto.

shared/fixes/bash/accounts_authorized_local_users_sidadm_orasid.sh
```diff
@@ -10,18 +10,21 @@ userlist="root"
...
+# if owner of any directory or file in the given list is the user oracle,
+# add the user oracle to the variable userlist. 
+# Usage: verify_oracle_user_to_userlist "${array[@]}"
+# Note: this function might modify the value of the global variable userlist
+function verify_oracle_user_to_userlist {
 	local path_list=("$@")
-	local is_oracle=0
+	local is_oracle=no
 	for path in $path_list ; do
 		if [ $(ls -ld $path | awk '{print $3}') = "oracle" ]; then
```
matejak: To be somehow safer, use `"$path"`, as it is not supposed to contain spaces.

xiruiyang:
@matejak   Did you see my comment above? It looks hidden after your comment (shown as pending?) https://github.com/OpenSCAP/scap-security-guide/pull/3012/files/cb816c1610da35890a27399070a024e313501f81

In case there is no or only one SID, the value of the variable is a string. But if there are multiple SIDs, it is an array. For example, there are two SIDs (LIN and QO1), the path_to_brspace_list is:
/sapmnt/LIN/exe/brspace
/sapmnt/QO1/exe/uc/linuxx86_64/brspace

and the path_oracle_SID_list is:
/oracle/LIN
/oracle/QO1

The OVAL test only works when there is no or only one SID, but the bash remediation works when there is multiple SIDs as well. At least, no user should be accidentally delete when it has a SID with it regardless what is the OVAL test result.

matejak:
@xiruiyang I am sorry, but I don't see any new comments from you. It may be a github bug, we have seen some strange comment behavior in other PRs.
Anyway, I see it as follows:
The variable `path_oracle_SID_list` defined by `path_oracle_SID_list=$(find /oracle/ -regex "^$oracle_SID_stem$")` is not an array, but a string. You can use it in a `for` loop, and it will iterate through it, because `for` uses whitespaces as separator of string elements.
You can try for yourself - if you put a helper print statement, e.g. `echo "${path_oracle_SID_list[1]}"`, (i.e. print second element of the array) I am pretty sure, that the result will be blank. On the other hand, I bet that `echo "${path_oracle_SID_list[0]}"` will show all filepaths, space or newline-separated.
The `for path in $path_list` is the same as `for path in ${path_list[0]}`, if you don't index the array for a particular element, bash just assumes that you want the first element of the array.
My point is that the usage of arrays is redundant, as you always use only the first element. Therefore, I suggest removing usage of arrays, as it may be misleading for others.
Finally, I share a link about saving command results into an array. I just mention it for reference, because trying to implement the correct solution could result in too hard to maintain code to be worth it: https://stackoverflow.com/questions/23356779/how-can-i-store-find-command-result-as-arrays-in-bash

xiruiyang:
@matejak you are correct, _echo "${path_oracle_SID_list[0]}"_ shows all filepaths and _echo "${path_oracle_SID_list[1]}"_ is blank. :+1:  I made the modifications according. Thanks a lot.

matejak:
Nice, I am glad it worked for you too. Thank you for your contribution!

PR #3013, opened 2018-06-21
Integrate audit_rules_sysadmin_actions rhel6 check into shared

yuumasato:
#### Description:

- Use jinja in shared check to customize it for RHEL6
- Drop handwritten RHEL6 check

#### Rationale:

- Unshadows RHEL6 `audit_rules_sysadmin_actions`

#### Notes
- Difference between checks is that there is not `augenrules` for RHEL6, so only `/etc/audit/audit.rules` is verified in RHEL6 (auditctl).
- I wanted to avoid nearly identical tests for auditclt, so I have reused the test `test_audit_rules_sysadmin_actions_sudoers_auditctl` from shared for RHEL6 check.
  - The regex that existed in shared was slightly better as it allowed key to be specified as `-k` or `-F key=`
  - The ID for the test is also slightly different. It is an "internal" id, so I guess there shouldn't be much impact to end users.
  - Any thoughts on re-utilization of test?


shared/checks/oval/audit_rules_sysadmin_actions.xml
```diff
@@ -3,11 +3,20 @@
     <metadata>
       <title>Audit System Administrator Actions</title>
       <affected family="unix">
+{{% if product == "rhel6" %}}
+        <platform>Red Hat Enterprise Linux 6</platform>
+{{% else %}}
         <platform>Red Hat Enterprise Linux 7</platform>
         <platform>multi_platform_fedora</platform>
+{{% endif %}}
```
mpreisler: these conditionals are not necessary. you can just say that it's applicable to all three

mpreisler:
@openscap-ci test this please (jenkins was out of whack)

shared/checks/oval/audit_rules_sysadmin_actions.xml
yuumasato: My concern is that we will end up generating definitions applicable for RHEL6, with checks for `augenrules` when it is not supported. These definitions will be in OVAL file for RHEL7 and Fedora.

Do you think it can cause problems or confusion?
Like why a check that is applicable for RHEL6 differ in OVAL files for different systems?

mpreisler: Wait, that's not how it works though, is it? It will generate the rhel6 definitions with rhel6 contents and the rhel7 definitions for rhel7. You don't have to put conditionals around affected platform. The definitions inside rhel7 oval will be for rhel7...

yuumasato: > It will generate the rhel6 definitions with rhel6 contents and the rhel7 definitions for rhel7.

Yes, but the definition in rhel7 oval will say in its affected/platform that it can be applied to rhel6, and this definition will check for augenrules, which rhel6 does not support.
While in rhel6 oval it will apply to rhel7, and not check rules defined for augenrules.

The scenario that someone will actually compare these definitions between oval files can be a very corner case though.

mpreisler: Yes but that is inevitable and is even worse with multi_platform_all which we use a lot. I want to fix it in #3006 and would appreciate your input there.

redhatrises:
RHEL6 actually does allow use to use augenrules.

yuumasato:
Jenkins fail is strange, and looks unrelated, anyway.

> RHEL6 actually does allow use to use augenrules,

Indeed, it was not there when launched, but now it supports.
I'm closing this as most of changes here don't make sense anymore.

PR #3014, opened 2018-06-21
[WIP] Refactor and modularize utils

cipherboy:
Continuation of #2949. 

#### Description & Scope

The aim of this PR is to refactor the code in the `utils/` directory and place common code in the `ssg` module. This is the last of the external refactors, before we can refactor the `ssg` module itself.

- `count_oval_objects.py` has code moved to `ssg/_utils_count.py`
- `create-stig-overlay.py` has code moved to `ssg/_utils_overlay.py`
- `find_duplicates.py` has code moved to `ssg/_utils_duplicates.py`
- `find_shadowed_files.py` has code moved to `ssg/_utils_shadowed.py`
- `fix-rules.py` has code moved to `ssg/_utils_fixes.py`
- .. and more!

Note that in the case of `count_oval_objects.py`, functionality was fixed; in the case of `fix-rules.py`, functionality was restored. The latter included changing `ssg/_jinja.py` to make `jinja2_cache_enabled` not a required key, and handling its missing gracefully. This seemed to be the original intent of the code. Note that, in the case of `create-stig-overlay.py`, functionality could not fully be tested by the author and help testing would be appreciated.

#### Rationale

Each of these utilities has significant internal code, including some that can be easily shared between multiple scripts. By converting these into modules and refactoring, eventually these parts can be unit tested and better documented.

#### Future Plans

Two more refactoring PRs will likely follow:

- One for `ssg`, to better organize the final module
- One for the `utils` folder, replacing - in paths with _ for global consistency.

After the refactoring is complete, two additional PRs will likely follow:

- Documentation (inline + external) of the new system
- Unit tests of critical components


mpreisler:
Does it make sense to move one time utility code into the ssg module? What is the reason for it to be there if it's only used by the CLI utility?

cipherboy:
>
> Does it make sense to move one time utility code into the ssg module? What
> is the reason for it to be there if it's only used by the CLI utility?
>

Are you talking about fix-rules.py? The goal would be to extend the
functionality in future work and build a jinja2 aware YAML parser for
applying transformations and computing differences among input files. The
core parsing logic would be reusable and eventually the editing logic can
be updated to use the new logic.

Also, much of this applies to profiles and groups and vars as well as
rules, so using this elsewhere would definitely be possible.

>

mpreisler:
> Are you talking about fix-rules.py?

I am talking in general about the utils. I suppose it's all for the best becaue this will enable us to remove duplicate code but I think we must avoid the other extreme. We must avoid having too much in the shared ssg module. If it's only related to the util/build-script and there is little chance of it being reused, let's have it in that file.

cipherboy:
(This is blocked until Debian8 and wrlinux guides are merged as common tooling that exists in this PR and those PRs need to be synced).

matejak:
Postponed to 0.1.41, as this PR is still quite in progress.

jan-cerny:
Blocked by #3156

cipherboy:
Closing as this might be perpetually WIP. :crying_cat_face:

PR #3015, opened 2018-06-21
Unshadow RHEL6 audit_rules_sysadmin_actions

yuumasato:
#### Description:

- Drop audit_rules_sysadmin_actions from RHEL6 product.
- Enable RHEL6 in:
  - audit_rules_augenrules

Follow up of https://github.com/OpenSCAP/scap-security-guide/pull/3013


redhatrises:
@openscap-ci test this please

mpreisler:
needs a rebase on top of #3016 

otherwise jenkins will keep failing

yuumasato:
Rebased.

yuumasato:
The OVAL definition `audit_rules_augenrules` went missing for `rhel-osp7`, so I added it explicitly in the definition.

This happens because map_name() is happy to match with begining of product string.
"rhel-osp" product begins with "rhel", so "Red Hat Enterprise Linux 7" is fine.
But check for "multi_platform_" is exact, so obviously multi_platform_rhel, does not match with multi_platform_rhel-osp.

Not sure, but I believe this is by design, so that variations of a product can easily match the base platform. For the future, we may want to have similar capability for `multi_platform_`, so that we don't need to go all over the definitions adding `multi_platform_rhel-osp` if needed.

shared/checks/oval/audit_rules_augenrules.xml
```diff
@@ -3,7 +3,8 @@
     <metadata>
       <title>Record Any Attempts to Run semanage</title>
       <affected family="unix">
-        <platform>Red Hat Enterprise Linux 7</platform>
+        <platform>multi_platform_rhel</platform>
+        <platform>multi_platform_rhel-osp</platform>
```
mpreisler: I don't think multi_platform_rhel-osp even works. Use the full name because there is only one in SSG.

yuumasato: It works¸ the definition is in rhel-osp7 content. The build went from fail to success.
But yeah, there is only one OSP product.

yuumasato: Updated.

PR #3016, opened 2018-06-21
Avoid the ansible playbook syntax checks with old versions of OpenSCAP

mpreisler:
See https://github.com/OpenSCAP/openscap/pull/977

Make jenkins pass again!


redhatrises:
LGTM. Ack.

PR #3020, opened 2018-06-21
Use shared bash fix for RHEL7's auditd_data_retention_space_left_action

mpreisler:
The differences between shared and overriden are very small and mostly
stylistic, let's just use the shared.


yuumasato:
Ack.

PR #3021, opened 2018-06-21
ol7 is the only OS using uuidd package installed, no need to have it …

mpreisler:
…in shared CSV


yuumasato:
And it is already in ol7's CSV. Ack.

PR #3024, opened 2018-06-22
Make cmake to favorize Python3.

matejak:
This PR changes the cmake setting, so it will pick up Python3 interpreter if available, and will fall back to Python2.
The Python3 build is ordered differently (maybe because there was an implementation change to the 3.6 dicts), but datastreams that I have checked contain the same lines.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,7 +81,10 @@ if ("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_BINARY_DIR}")
     )
 endif()
 
+
+set(Python_ADDITIONAL_VERSIONS 3 2)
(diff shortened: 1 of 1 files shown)


cipherboy:
LGTM, though I will note that compile times are slower on Python3... Hopefully #3010 will help with that some:

Python 2:
```
CMake:
real	0m40.361s
user	0m32.271s
sys	0m7.562s

Build:
real	1m38.852s
user	5m1.364s
sys	0m47.545s

```

Python 3:
```
CMake:
real	1m16.732s
user	1m7.586s
sys	0m8.082s

Build:
real	1m50.306s
user	6m0.905s
sys	0m42.927s
```

(Python2 tested on `HEAD~` of the PR).

PR #3025, opened 2018-06-22
Updated the ansible service_disabled template to account for sockets

mpreisler:
Ansible remediations will now also try to disable the socket of a service they are trying to disable.

This makes it possible to streamline the fixes even more and avoid overriding them.


yuumasato:
Looks good, Ack.

PR #3026, opened 2018-06-22
Added dod_banners selector for RHEL6

mpreisler:
This prevents selector errors such as:
OpenSCAP Error: Invalid selector 'dod_banners' for xccdf:value/@id='login_banner_text'. Using null value instead. [xccdf_policy.c:2103]
Could not resolve xccdf:sub/@idref='login_banner_text'! [xccdf_policy_substitute.c:107]
A fix for Rule/@id="gconf_gdm_set_login_banner_text" was skipped: Text substitution failed. [xccdf_policy_remediate.c:552]
Invalid selector 'dod_banners' for xccdf:value/@id='login_banner_text'. Using null value instead. [xccdf_policy.c:2103]


cipherboy:
@mpreisler Is `dod_banners` the exact same as `dod_default`? Can we not change occurrences of `dod_banners` to `dod_default`?

redhatrises:
They are not the same. `dod_banners` has 2 banners in it with an `|` or in the regex.

cipherboy:
@redhatrises Ah, good catch, I missed that one.

ACK from me then.

PR #3027, opened 2018-06-22
Refactor ssg module

cipherboy:
After #3024, build times significantly increased due to automatic use of Python 3. This attempts to do something about it by restructuring the `ssg` module. In essence, these changes leave the `ssg` module intact, but require external scripts to do explicit imports of desired components.

We also rename components back to their original names (i.e., without the prefixed `_`). This has a few other misc. changes to prefer Python 3 and reduce our use of try/except for handling preventable errors.

(On Python 3):

Before changes (965c8116677bc8841b90313c8775380a6aca8101):

```
cmake:
real    0m30.620s
user    0m26.861s
sys     0m3.401s

build:
real    2m2.510s
user    5m38.069s
sys     0m46.492s
```

After changes (1120567d6d45c189ee909e0ccffe3694c282ac26):
```
cmake:
real    0m19.405s
user    0m16.478s
sys     0m2.673s

build:
real    1m41.005s
user    4m44.453ss
sys     0m40.625s
```


build-scripts/build_all_guides.py
```diff
@@ -15,8 +15,11 @@
 import threading
 import sys
 
-import ssg
-ElementTree = ssg.xml.ElementTree
+import ssg.build_guides
+import ssg.xccdf
+import ssg.utils
+import ssg.xml
+from ssg.xml import ElementTree
```
mpreisler: how about using ssg.xml.ElementTree in the code? it's slightly more chatty but it helps new devs figure out where to get it. maybe it will prevent people from importing the python ElementTree

cipherboy: @mpreisler Okay, to my knowledge all external uses of `ElementTree` in `build-scripts` have been replaced by `ssg.xml.ElementTree`.

cipherboy:
`3 min 29 sec`, down from `4 min 4 sec` for the last PR -- I think we have a winner :)

mpreisler:
Great work!

cipherboy:
(This was part of #2949 efforts).

PR #3028, opened 2018-06-22
Fix python2 build after refactoring ssg module

cipherboy:
Regression in #3027

```
CMake:
real	0m15.610s
user	0m12.601s
sys	0m2.787s
```

### DIFF
--- a/ssg/build_yaml.py
+++ b/ssg/build_yaml.py
@@ -21,7 +21,7 @@ def add_sub_element(parent, tag, data):
     # TODO: Remove this function after we move to Markdown everywhere in SSG
     try:
         ustr = str("<{0}>{1}</{0}>").format(tag, data)
-    except NameError:
+    except UnicodeEncodeError:
         ustr = unicode("<{0}>{1}</{0}>").format(tag, data)
 
     try:


ssg/build_yaml.py
```diff
@@ -21,7 +21,7 @@ def add_sub_element(parent, tag, data):
     # TODO: Remove this function after we move to Markdown everywhere in SSG
     try:
         ustr = str("<{0}>{1}</{0}>").format(tag, data)
-    except NameError:
+    except UnicodeEncodeError:
```
mpreisler: we discussed this and we need a different solution here. consider this:
we are using python3, the unicode encode causes an error because the utf-8 is malformed. this will get us to the line under this comment where it will raise a NameError because unicode() is not defined on python3. This hides the exception and doesn't tell you where the error is.

dahaic:
Jenkins, test this please.

cipherboy:
@dahaic This shouldn't be merged.

jan-cerny:
@cipherboy Isn't the problem that the 2 lines are swapped?
I think it should be:

```
try:
  ustr = unicode(...)
except NameError:
  ustr = str(...)
```

because `unicode()` is not defined in Python 3, but `str()` on Python 2 will fail on non-ASCII characters, so we should try `unicode()` first.

jan-cerny:
It used to be like that before: https://github.com/OpenSCAP/scap-security-guide/blob/b1dfc6f2303a01160ad788feca55717a4e1dbf76/ssg/_build_yaml.py#L21

PR #3030, opened 2018-06-25
Refactored bash service disable remediation + support for disable sockets

matejak:
The remediation got split into multiple functions.
The systemd service disable function now disables service and checks for availability of socket unit file. If it exists, it also disables socket.
I have checked usage of the shared remediation function and ensured that the refactoring doesn't break current use cases.

### DIFF
--- a/shared/bash_remediation_functions/service_command.sh
+++ b/shared/bash_remediation_functions/service_command.sh
@@ -3,72 +3,81 @@
 # Example Call(s):
 #
 #     service_command enable bluetooth
-#     service_command disable bluetooth.service
 #
 #     Using xinetd:
-#     service_command disable rsh.socket xinetd=rsh
+#     service_command disable rsh xinetd=rsh
 #
+SYSTEMCTL_EXEC='/usr/bin/systemctl'
+SERVICE_EXEC='/sbin/service'
+CHKCONFIG_EXEC='/sbin/chkconfig'
+
+
+function xinetd_disable {
+	sed -i "s/disable.*/disable         = no/gI" "/etc/xinetd.d/$1"
+}
+
+
+function xinetd_enable {
+	sed -i "s/disable.*/disable         = yes/gI" "/etc/xinetd.d/$1"
+}
+
+
+function systemd_disable {
+	local service_stem="$1"
+	"$SYSTEMCTL_EXEC" stop "$service_stem".service
+	"$SYSTEMCTL_EXEC" disable "$service_stem".service
+	# Disable socket activation if we have a unit file for it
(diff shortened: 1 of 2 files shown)


mpreisler:
Maybe it's just me but could you please declare functions before you use them? It makes it MUCH more readable to people unfamiliar with your code / coding style. From multiple reviews of your code I suspect you like writing a high-level function and then filling in the functions doing the heavy lifting. That is fine but please reorganize it so that people reading your code can read it top to bottom.

I also think we should just use jinja2 to do the heavy-lifting here. However that's not in the scope of this PR so I won't insist on it.

matejak:
For reference, the ordering here [has a reason](https://softwareengineering.stackexchange.com/a/138325).
However, we have agreed that the ordering of functions in the project is already mainly the other way (low-level first, highest level last).

redhatrises:
So.... the whole point of the `service_command` was so that we wouldn't have to go to this level in our bash remediation scripts, but I guess that having just `service_command` was too confusing?

mpreisler:
> So.... the whole point of the service_command was so that we wouldn't have to go to this level in our bash remediation scripts, but I guess that having just service_command was too confusing?

I agree that this PR goes a bit too far, especially considering this is going to get pasted all over the place in the XCCDF. Great bash style has its place but for remediations we probably want as compact as possible. In the future we should also just use jinja2 and avoid as much runtime detection as we can.

matejak:
Well, if jinja is used on top of this PR, then the actual remediation code will be short and in good style. Moreover, pasting of helper functions all over again is a build system quirk that can be fixed at the source.
I vote very strongly for keeping the functionality extracted, as it will be easy to convert into macros, that will produce short code (good for users) from more verbose source (good for us devs).

shared/bash_remediation_functions/service_command.sh
```diff
@@ -3,72 +3,81 @@
...
-
-  if [ "$service_operation" = 'disable' ] ; then
-    sed -i "s/disable.*/disable         = no/gI" /etc/xinetd.d/$xinetd
-  else
-    sed -i "s/disable.*/disable         = yes/gI" /etc/xinetd.d/$xinetd
-  fi
-fi
+	# If systemctl is installed, use systemctl command; otherwise, use the service/chkconfig commands
+	# foo_$service_state => foo_disabled / foo_enabled
+	if test -f "/usr/bin/systemctl"; then
```
mpreisler: you declared all the fancy variables and then use a hardcoded string here :-) I also think `which` would be more prudent.

mpreisler:
I am sorry but I think this goes way too far. This is trivial code spanning a zillion lines and being extracted to multiple functions. IMO that's ridiculous...

matejak:
Succeeded by #3055

PR #3033, opened 2018-06-25
Modernized the Dockerfile to use python3, Fedora 28 and ninja build

mpreisler:


cipherboy:
@mpreisler LGTM. Question about expected usage. If you type `docker build .`, it will copy in the current contents of `scap-security-guide` into the container. To actually perform the SSG build, you have to type `docker run <id>`... To make this easier, should we change the last `CMD` into a `RUN`, such that you can type `docker build .`, and have it update contents and make rebuilds more obvious? You'll end up running `docker build .` again anyways if you changed the contents of the directory, and we don't have any long-running contents after this.

Perhaps add the test suite as the `RUN` command?

mpreisler:
> Perhaps add the test suite as the RUN command?

good points, I will incorporate that into this PR

cipherboy:
@mpreisler Thanks for the changes! :)

PR #3034, opened 2018-06-25
Make the shadow detector use python3 pathlib

matejak:
The pathlib module is part of Python3 std lib, and is much easier to read.

### DIFF
--- a/utils/find_shadowed_files.py
+++ b/utils/find_shadowed_files.py
@@ -2,10 +2,11 @@
 
 import os
 import argparse
-import sys
+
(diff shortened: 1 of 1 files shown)


mpreisler:
NACK.

python2 doesn't have pathlib, feels wrong to break python2 support in this script just because it might be easier.

PS: This is a script that will be removed from SSG repo in a couple months, if it works it works and nobody cares about refactoring it :-)

matejak:
OK, I am not so surprised.

PR #3035, opened 2018-06-25
Small style fixes

cipherboy:
To make parsing easier when merging `rhel6` into `linux_os` guides, remove extraneous double space. The change in the shared Ansible folder is unnecessary but as it is the only one, might as well keep it for consistency.


redhatrises:
ACK

PR #3037, opened 2018-06-25
Deduplication and merger of RHEL6 -> Linux OS guide

cipherboy:
This starts the merge of the RHEL6 guide into the Linux OS guide to prevent duplication of efforts. These changes should be fairly trivial, but might conflict with many other pull requests. As these changes are automated, they can be easily updated if necessary.

For all files in `rhel6/guide` for which there is an exactly corresponding `linux_os/guide` file, change the file in `linux_os/guide` to:

 - Add RHEL6 to the `prodtype` field, where there is already a `prodtype` field.
 - Add RHEL6 identifiers where they differ or are missing.
 - Add RHEL6 references where they differ or are missing.

PR #3041, opened 2018-06-25
Merge security_patches_up_to_date.rule from rhel6->linux_os

cipherboy:
This file had to be manually updated due to duplicated sections (`ocil_external_content`). I believe this file to be good to go with these updates in Linux OS. I would appreciate a comprehensive comparison of this file to its RHEL 6 equivalent: 

https://github.com/OpenSCAP/scap-security-guide/blob/master/rhel6/guide/system/software/updating/security_patches_up_to

https://github.com/OpenSCAP/scap-security-guide/blob/master/linux_os/guide/system/software/updating/security_patches_up_to

Part of #3037 efforts.

### DIFF
--- a/linux_os/guide/system/software/updating/security_patches_up_to_date.rule
+++ b/linux_os/guide/system/software/updating/security_patches_up_to_date.rule
@@ -32,9 +32,13 @@ rationale: |-
 severity: high
 
 identifiers:
+    cce@rhel6: 27635-2
     cce@rhel7: 26895-3
 
 references:
+    stigid@rhel6: RHEL-06-000011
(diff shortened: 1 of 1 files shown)


redhatrises:
stigviewer.com is tempting to use for its ease and convenience, but it is heavily outdated. Please don't use it.

redhatrises:
Per https://github.com/OpenSCAP/scap-security-guide/blob/master/shared/references/disa-stig-rhel6-v1r18-xccdf-manual.xml#L486, this is a Medium.

redhatrises:
Ack as this PR is only for identifiers, references, and cce.

PR #3043, opened 2018-06-25
Missing references in Linux OS guides

cipherboy:
Several rules in the `linux_os` guide tree were missing a references section completely, so the work in #3037 does not update them. This adds the missing references from the RHEL 6 guides into the Linux OS shared guide.

Part of #3037 efforts.


cipherboy:
@redhatrises Updated!

redhatrises:
Ack.

PR #3046, opened 2018-06-26
Fix login.defs umask configuration.

### LINKED ISSUE #2989: RHEL7 Ansible remediation: UMASK in login.defs incorrectly set
#### Description of problem:
When running ansible remediation against RHEL7, the check tagged 'accounts_umask_etc_login_defs', which sets the default UMASK entry in /etc/login.defs, incorrectly sets this value.

The issue appears to be that the umask is being treated by Ansible/Jinja2/Python as an octal (which technically it is), and being written as the decimal equivalent. This is from the variable _var_accounts_user_umask_.

Tested with 0.1.36 that ships with RHEL, with v0.1.39 from git tag and master branch (HEAD = 37d87207bf211e17c4611d69e65d0428cae06545).

CentOS Linux release 7.5.1804 (Core)
I've also been testing on RHEL7 workstation.

#### Steps to Reproduce:

1. Build from source as per https://github.com/OpenSCAP/scap-security-guide/blob/master/BUILD.md#building
2. Generate ansible remediation. In my case, I'm building for DISA STIG. _oscap xccdf generate fix --fix-type ansible --profile xccdf_org.ssgproject.content_profile_stig-rhel7-disa --output stig-rhel7-role.yml ./ssg-rhel7-ds.xml_
3. Apply the offending control. _ansible-playbook -b -c local -i localhost, --tags accounts_umask_etc_login_defs stig-rhel7-role.yml_

#### Actual Results:
```
$ grep UMASK /etc/login.defs
UMASK 63
## NOTE
$ printf "%d\n" 077
63
```

#### Expected Results:
```
$ grep UMASK /etc/login.defs
UMASK 077
```

Note above in the actual results where I've shown that "077" octal is "63" decimal, which is the result I get in /etc/login.defs.

If I edit the generated playbook and make the _var_accounts_user_umask_ variable a string instead of a number, I get the desired result (*'077'* instead of *077*)

matejak:
This PR fixes two issues:
Major: The Ansible variable expansion happens inside double quotes. This ensures that octal numbers, booleans and other values that could be misinterpreted by the YML format are passed through accurately. Plus double quotes are escapable. Fixes #2989.
Minor: The OVAL check was cas-insensitive, so `UMASK` was as good as `umask`. This doesn't comply with XCCDF, nor with man pages.
The test suite now pass for the `login_defs` rule stem for both bash and ansible remediation engines.

### DIFF
--- a/shared/checks/oval/accounts_umask_etc_login_defs.xml
+++ b/shared/checks/oval/accounts_umask_etc_login_defs.xml
@@ -18,7 +18,7 @@
   <ind:textfilecontent54_object id="obj_umask_from_etc_login_defs"
   comment="Umask value from /etc/login.defs" version="1">
     <ind:filepath>/etc/login.defs</ind:filepath>
-    <ind:pattern operation="pattern match">^[\s]*(?i)UMASK(?-i)[\s]+([^#\s]*)</ind:pattern>
+    <ind:pattern operation="pattern match">^[\s]*UMASK[\s]+([^#\s]*)</ind:pattern>
     <ind:instance datatype="int" operation="greater than or equal">1</ind:instance>
   </ind:textfilecontent54_object>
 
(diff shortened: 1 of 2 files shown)


redhatrises:
Ack. Thanks @matejak

PR #3049, opened 2018-06-26
Unshadow audit_rules_unsuccessful_file_modification

yuumasato:
#### Description:

- Drop the check from rhel6
- Update the check in shared to account for different AUID
- Update extended-definition's template to apply to also rhel6

#### Rationale:

- Both shared and rhel6 are checking the same thing but from different points of view.
  - share check iterates over each rule syscal, and
  - rhel6 check iterates over each error (EACCES/EPERM)
- I've favored the shared check, as its regex'es are better in general, supports augenrules, and is more readable.


shared/checks/oval/audit_rules_auditctl.xml
```diff
@@ -3,7 +3,7 @@
     <metadata>
       <title>Record Any Attempts to Run semanage</title>
       <affected family="unix">
-        <platform>Red Hat Enterprise Linux 7</platform>
+        <platform>multi_platform_rhel</platform>
```
mpreisler: how could this OVAL possibly work on RHEL6? It relies on systemd!!

cipherboy:

shared/checks/oval/audit_rules_auditctl.xml
yuumasato: Well, that was a huge oversight, :D

yuumasato: PR #3056 should address this issue.

mpreisler: Rebase this please

yuumasato: Rebased.

PR #3050, opened 2018-06-26
To be on the safe side, force ansible XCCDF values to be interpreted as str…

mpreisler:
…ings

Avoid quotes though because that enables all sorts of escaping rules
that we would have to work around.

This fixes a regression which causes invalid Ansible syntax and breaks some of our roles - especially rhel6 and 7. Regression happened in b0eb3b7f7baa1a57dac3e373209d20bd55b3f215 and was reviewed by me (oops :-D)

### DIFF
--- a/ssg/build_remediations.py
+++ b/ssg/build_remediations.py
@@ -190,11 +190,15 @@ def expand_xccdf_subs(fix, remediation_type, remediation_functions):
                 "substituting directly."
             )
 
+        # we use the horrid "!!str |-" syntax to force strings without using
+        # quotes. quotes enable yaml escaping rules so we'd have to escape all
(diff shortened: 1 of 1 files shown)


cipherboy:
LGTM, builds locally and tests pass: 

```
(snip)
[420/420] [sl7-guides] generating HTML guides for all profiles in ssg-sl7-ds.xml
real 111.78
user 303.19
sys 41.43
```

```
(snip)
157/170 Test #104: ansible-role-syntax-check-rhel7 ..........................................   Passed   78.08 sec
(snip)
100% tests passed, 0 tests failed out of 170

Total Test time (real) = 289.08 sec
```

cipherboy:
@mpreisler Thanks for the fix! :)

PR #3052, opened 2018-06-26
ssg module - even more linting fixes

cipherboy:
Even more module linting fixes! :D

Part of #2949 efforts.


mpreisler:
Will there be even more later? :D

cipherboy:
@mpreisler This is updated and builds correctly now. :)

PR #3054, opened 2018-06-27
Enable Travis CI with containerized build

jan-cerny:
#### Description:

This will build and test SSG on Travis CI using a Docker container.
Also adds a build status badge to README.

#### Rationale:

Our code will be tested on other CI infrastructure

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+sudo: required
+
+services:
+    - docker
+
+language: python
+
+before_install:
+    - docker build --no-cache --tag ssg:latest .
+
+script:
+    - docker run ssg:latest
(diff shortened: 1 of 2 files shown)


jan-cerny:
The problem is that Travis CI uses Ubuntu 14.04 Trusty Tahr, which has OpenSCAP 1.0.2, which cannot be used to build SSG. See https://packages.ubuntu.com/trusty/libs/libopenscap8

I suggest that we provide a PPA that will provide DEB packages with latest version of OpenSCAP for Ubuntu 14.04.

Other solution could be  that the SSG build in Travis would build OpenSCAP from GIT as well. But that would be slower. I would prefer the PPA.

jan-cerny:
@cipherboy suggested using containerized build. I'm going to look into this possibility.

jan-cerny:
The container build was a great idea. It looks like the Travis CI has no problem to build SSG inside an container with our Dockerfile. Thanks @cipherboy !

cipherboy:
@jan-cerny \o/ Yay! 

One thought for you/the rest of the team: our `Dockerfile` uses `fedora:28`, so that is what travis/docker is using: https://travis-ci.org/OpenSCAP/scap-security-guide/builds/399109009#L446

Do we want to add other platforms for the travis build via. multiple dockerfile? iirc, the original point of this PR was to add Ubuntu support, but by using our `Dockerfile`, we just add another Fedora 28 build.

jan-cerny:
Yeah, we could add other Dockerfiles, e.g. Ubuntu 18.04 has OpenSCAP 1.2.15, which would hopefully work (I haven't tested that yet).

mpreisler:
We could add more Dockerfiles but I think even just this is a good step forward.

PR #3055, opened 2018-06-27
Rewrite bash service disable/enable templates

matejak:
This PR rewrites the SSG template for service disable/enable.
As a prerequisite, support for jinja2 in templates has been introduced. I have separated expansion of yaml content and expansion of generic content.
As jinja can't easily raise exception, so if something goes wrong, the `JINJA TEMPLATE ERROR` will be present in the output. I think that raising exceptions from jinja is possible, but IMO out of scope of this PR.

### DIFF
--- a/shared/templates/template_BASH_service_disabled
+++ b/shared/templates/template_BASH_service_disabled
@@ -3,7 +3,22 @@
 # strategy = enable
 # complexity = low
 # disruption = low
-# Include source function library.
-. /usr/share/scap-security-guide/remediation_functions
+{{%- if init_system == "systemd" %}}
 
-service_command disable %DAEMONNAME%
+SYSTEMCTL_EXEC='/usr/bin/systemctl'
+"$SYSTEMCTL_EXEC" stop '%DAEMONNAME%.service'
+"$SYSTEMCTL_EXEC" disable '%DAEMONNAME%.service'
+# Disable socket activation if we have a unit file for it
+"$SYSTEMCTL_EXEC" list-unit-files | grep -q '^%DAEMONNAME%.socket\>' && "$SYSTEMCTL_EXEC" disable '%DAEMONNAME%.socket'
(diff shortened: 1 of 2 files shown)


mpreisler:
I am 99% sure the changes to generate-from-templates are not necessary. Jinja will be used down the line when remediations are combined. Why did you add support for jinja2 to templates? Please undo that.

matejak:
The downside of this that built files in `rhel7/fixes/bash/` are not valid scripts. If it is not a problem, the last (and possibly also the first) commits are not needed.

mpreisler:
> If it is not a problem, the last (and possibly also the first) commits are not needed.

yeah, that's not a big problem IMO. Even if it is a problem, this is not the right time to fix it as a bunch of sneaky commits in an unrelated PR :-)

Please include 6b2b3947427828b1dbbead04aa0ebc47ed171d83 and drop the other commits.

matejak:
I have rebased to strip the two commits.

shared/templates/template_BASH_service_disabled
```diff
@@ -3,7 +3,22 @@
 # strategy = enable
 # complexity = low
 # disruption = low
-# Include source function library.
-. /usr/share/scap-security-guide/remediation_functions
+{{%- if init_system == "systemd" %}}
```
mpreisler: what's the reason for the dash?

cipherboy: @mpreisler / @matejak Also in #3058, you add a dash on both sides on all of the macros.

matejak: It strips whitespace: http://jinja.pocoo.org/docs/2.10/templates/#whitespace-control
The official documentation is not so great, but generally you want to have whitespace in jinja code, because it helps you to understand what's going on, but you don't want to have the result full of line breaks.

PR #3058, opened 2018-06-27
Make all jinja macros defined in jinja files.

matejak:
Previously, some macros were added to the list of available jinja macros by Pyhon code.
This could confuse people looking for those definitions, and it also required editing jinja and Python files in case of macro renames.
This PR makes sure that every jinja macro is defined in a jinja file, rather than automagically generated by the Python build system.
Remark: The dashes in jinja control sequences are arranged in such way that the built datastreams before and after are almost the same.
There was a somehow unrelated jinja bug discovered during the process, and fix for it is part of this PR.

### DIFF
--- a/docs/manual/developer_guide.adoc
+++ b/docs/manual/developer_guide.adoc
@@ -596,17 +596,19 @@ Available CCEs that can be assigned to new rules are listed in the `shared/refer
 Some of existing rule definitions contain elements that use macros.
 There are two implementations of macros:
 
-* link:http://jinja.pocoo.org/docs/2.10/[Jinja macros], that are defined in `shared/modules/macros.jinja`, and
-* legacy XSLT macros, which are defined in `shared/transforms/*.xslt`.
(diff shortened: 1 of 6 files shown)


cipherboy:
(Not a review)

It is not obvious to me why Jenkins took so long building this PR.

On my system (9c0a1f129f5ffbf088c831d68edcf6dda8e71950 / head of PR):
```
Cmake:
real 14.84
user 12.71
sys 2.05

Build:
real 74.59
user 239.07
sys 28.03
```

On master / PR branching point (4aff3d6abdbfc2df59cfcb457d4e422717508d42):
```
Cmake:
real 14.25
user 12.27
sys 1.91

Build:
real 68.20
user 229.21
sys 27.06
```

So while it is slower, it is not noticeably so.

matejak:
It may be caused by running out of CPU credits - there have been a huge PR traffic lately.

cipherboy:
LGTM. Thanks for updating the documentation too, @matejak! :)

PR #3059, opened 2018-06-27
ssg module - even more than even more linting fixes

cipherboy:
Dependency: ~#3052~
The last of the easy module linting fixes left over from #2949. :)

This also fixes a regression introduced in the refactor where calling `profile_stats.py` with a profile other than all would error.

#### Plan

After this, I hope to:
 - Reconsider creating a build sub-module
 - Look at refactoring and moving around larger sections of code
 - Renaming all constants and decreasing use of global variables
 - Inline documentation
 - External documentation
 - Build speed

### DIFF
--- a/ssg/build_profile.py
+++ b/ssg/build_profile.py
@@ -51,7 +51,9 @@ class XCCDFBenchmark(object):
             if rule_id is None:
                 raise RuntimeError("Can't index a rule with no id attribute!")
 
-            assert(rule_id not in self.indexed_rules)
+            if rule_id in self.indexed_rules:
+                raise RuntimeError("Multiple rules exist with same id attribute: %s!" % rule_id)
+
             self.indexed_rules[rule_id] = rule
 
     def get_profile_stats(self, profile):
@@ -59,7 +61,7 @@ class XCCDFBenchmark(object):
 
         # Holds the intermediary statistics for profile
         profile_stats = {
-            'profile_id': None,
+            'profile_id': "",
             'ssg_version': 0,
             'rules_count': 0,
             'implemented_ovals': [],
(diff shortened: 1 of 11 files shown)


ssg/build_profile.py
```diff
@@ -51,15 +51,17 @@ def __init__(self, filepath):
             if rule_id is None:
                 raise RuntimeError("Can't index a rule with no id attribute!")
 
-            assert(rule_id not in self.indexed_rules)
+            if rule_id in self.indexed_rules:
+                raise RuntimeError("Rule exists with multiple id attributes: %s!" % rule_id)
```
mpreisler: shouldn't this be "multiple rules with the same id"?

ssg/build_profile.py
```diff
@@ -100,8 +102,8 @@ def get_profile_stats(self, profile):
                       % profile)
                 print("* Available profiles:")
                 profiles_avail = self.tree.findall("./{%s}Profile" % (xccdf_ns))
-                for profile in profiles_avail:
-                    print("** %s" % profile.get('id'))
+                for _profile in profiles_avail:
+                    print("** %s" % _profile.get('id'))
```
mpreisler: why?

cipherboy: `profile` is shadowed from the outer scope, and `pylint` doesn't like it. This is technically a false-positive since the `sys.exit` call immediately following this loop ensures that the changes to `profile` won't get seen, but that's a different matter. See issue on line 105 on [Scrutinizer](https://scrutinizer-ci.com/g/OpenSCAP/scap-security-guide/inspections/a6da40b5-2622-4920-8a8a-d684fb475803/issues/files/ssg/build_profile.py?status=all&orderField=path&order=asc&honorSelectedPaths=0#inspectioncomment-106230593).

ssg/build_profile.py
cipherboy: @mpreisler Updated; good catch

ssg/build_profile.py
cipherboy: (I can drop that change if you want).

cipherboy:
@openscap-ci test this please

cipherboy:
https://scrutinizer-ci.com/g/OpenSCAP/scap-security-guide/inspections/9995fde4-47e0-44a3-9a0c-2a150f003704

PR #3060, opened 2018-06-27
Refactor usage of argparse in test_suite

cipherboy:
All that this changes is wrapping argparse in a `parse_args` function and the bulk of the script in a `main` function.

### DIFF
--- a/tests/test_suite.py
+++ b/tests/test_suite.py
@@ -16,134 +16,143 @@ import ssg_test_suite.rule
 from ssg_test_suite import xml_operations
 
 
-parser = argparse.ArgumentParser()
+def parse_args():
+    parser = argparse.ArgumentParser()
 
-common_parser = argparse.ArgumentParser(add_help=False)
-common_parser.add_argument("--hypervisor",
-                           dest="hypervisor",
-                           metavar="HYPERVISOR",
-                           default="qemu:///session",
-                           help="libvirt hypervisor")
-common_parser.add_argument("--domain",
-                           dest="domain_name",
-                           metavar="DOMAIN",
-                           required=True,
-                           help=("Specify libvirt domain to be used as a test "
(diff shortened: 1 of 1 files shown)


cipherboy:
Hello @scrutinizer-notifier! Nice to see you again :)

cipherboy:
I'm marking this as WIP as I realized the test suite doesn't work on Python 3.

cipherboy:
I'll update python3 in a separate PR.

dahaic:
@cipherboy does it make sense to also put that default handling logic into function? (Separate, probably)

cipherboy:
@dahaic Let me see if I understood what you said... (I think you're talking about defaults w.r.t. arguments, and the logic around setting them). If so:

Regarding lines [119 to 126](https://github.com/cipherboy/scap-security-guide/blob/1b47a54f3296a8a144ca7877dee992747071c2ea/tests/test_suite.py#L119-L126), I'm inclined to leave them in as-is, because it is updating an existing option based on other, parsed options. (`bench_id` depends on `options.{datastream, xccdf_id}` and you're calling out to an external function (`xml_operations.infer_benchmark_id_from_component_ref_id`)).

Regarding lines [132 to 133](https://github.com/cipherboy/scap-security-guide/blob/1b47a54f3296a8a144ca7877dee992747071c2ea/tests/test_suite.py#L132-L133), I'm also inclined to leave them there, because moving them up to `parse_args` would introduce more lines than it is worth since it only minimally impacts readability. I'd be more inclined to move the comment above it down to after the if, and add a new comment detailing that this is cleaning up the options.

dahaic:
@cipherboy yeah, I was talking about those. Thanks for clarification, I don't see the need for any changes. ACK

cipherboy:
@dahaic Thanks! :)

PR #3062, opened 2018-06-28
Move example profile ("CS2") from RHEL6 to docs

cipherboy:
Perhaps the branch name is misleading, but we remove the CS2 (**NOT C2S**) example profile from `rhel6` and add it to the developer documentation. 

This prevents it from being built every time we build SSG or the `rhel6` product.


redhatrises:
Are we still using guide.xslt?

shawndwells:
CS2 is a baseline for a very specific set of customers. Didn't realize the profile description was so vague.

Hey @jeffblank, can we be more... verbose... about the profile description? Looks like it was never changed to reflect attribution.

redhatrises:
@shawndwells are you sure that it is CS2 and not C2S?

shawndwells:
On 6/28/18 12:58 PM, redhatrises wrote:
> @shawndwells <https://github.com/shawndwells> are you sure that it is 
> CS2 and not C2S?

Yah. C2S is the CIA's public cloud image, CS2 is... some other agency... 
that apparently was never publicly attributed but Jeff and I spoke 
onstage at Summit about several times.

cipherboy:
I'm marking this as closed to prevent merging. Pending feedback from @jeffblank, we can add a more suitable description and title.

Perhaps we should also consider porting this profile to RHEL7.

PR #3064, opened 2018-06-28
Small changes to build output

cipherboy:
I'd like to get feedback on this. Removed notifications are currently commented out because they don't appear to be errors or warnings.

Based on feedback, I'll either remove the lines completely, or consider adding a VERBOSE option to the build that extends output, but which is default disabled. 

This'll help make builds more informative by reducing the quantity of extraneous information; limiting it to Errors and warnings. 

Thoughts?

### DIFF
--- a/build-scripts/sds_move_ocil_to_checks.py
+++ b/build-scripts/sds_move_ocil_to_checks.py
@@ -191,9 +191,6 @@ def main():
             # Also update the ID replacing 'ecomp' with 'comp'
             move_ocil_content_from_ds_extended_component_to_ds_component(datastreamtree, oldocilhref)
 
-    else:
-        print("No extended-components, nothing to do...")
-
     # if the in and out files are the same and we didn't do any changes we can
     # skip the serialization
     if extendedcomps is not None or outdatastreamfile != indatastreamfile:
--- a/build-scripts/unselect_empty_xccdf_groups.py
+++ b/build-scripts/unselect_empty_xccdf_groups.py
@@ -140,8 +140,6 @@ def main():
 
         affected_profiles.append(profile_element.get("id"))
 
-    print("Unselected empty groups in %s." % (", ".join(affected_profiles)))
-
     input_tree.write(options.output)
 
 
--- a/cmake/SSGCommon.cmake
+++ b/cmake/SSGCommon.cmake
@@ -99,16 +99,11 @@ macro(ssg_build_shorthand_xml PRODUCT)
     )
 endmacro()
 
-macro(ssg_build_xccdf_unlinked PRODUCT)
-    file(GLOB STIG_REFERENCE_FILE_LIST "${SSG_SHARED_REFS}/disa-stig-${PRODUCT}-*-xccdf-manual.xml")
-    list(APPEND STIG_REFERENCE_FILE_LIST "not-found")
-    list(GET STIG_REFERENCE_FILE_LIST 0 STIG_REFERENCE_FILE)
-
+macro(ssg_build_xccdf_unlinked_no_stig PRODUCT)
     add_custom_command(
         OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/xccdf-unlinked-resolved.xml"
         COMMAND "${XSLTPROC_EXECUTABLE}" --stringparam ssg_version "${SSG_VERSION}" --output "${CMAKE_CURRENT_BINARY_DIR}/xccdf-unlinked-resolved.xml" "${CMAKE_CURRENT_SOURCE_DIR}/transforms/shorthand2xccdf.xslt" "${CMAKE_CURRENT_BINARY_DIR}/shorthand.xml"
         COMMAND "${OPENSCAP_OSCAP_EXECUTABLE}" xccdf resolve -o "${CMAKE_CURRENT_BINARY_DIR}/xccdf-unlinked-resolved.xml" "${CMAKE_CURRENT_BINARY_DIR}/xccdf-unlinked-resolved.xml"
-        COMMAND env "PYTHONPATH=$ENV{PYTHONPATH}" "${PYTHON_EXECUTABLE}" "${SSG_BUILD_SCRIPTS}/add_stig_references.py" --disa-stig "${STIG_REFERENCE_FILE}" --unlinked-xccdf "${CMAKE_CURRENT_BINARY_DIR}/xccdf-unlinked-resolved.xml"
         DEPENDS generate-internal-${PRODUCT}-shorthand.xml
         DEPENDS "${CMAKE_CURRENT_BINARY_DIR}/shorthand.xml"
         DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/transforms/shorthand2xccdf.xslt"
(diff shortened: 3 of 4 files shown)


redhatrises:
Personally like the idea. I think adding a VERBOSE option would be way better than commenting out or removing the capability to get the output.

mpreisler:
@cipherboy and I also discussed having a tool that will tell you about missing coverage, identifiers, etc... instead of it ever showing up in build output. We could probably re-purpose profile-stats.py for this.

jan-cerny:
I would also prefer verbose option.

cmake/SSGCommon.cmake
```diff
@@ -104,18 +104,32 @@ macro(ssg_build_xccdf_unlinked PRODUCT)
...
+            COMMAND "${OPENSCAP_OSCAP_EXECUTABLE}" xccdf resolve -o "${CMAKE_CURRENT_BINARY_DIR}/xccdf-unlinked-resolved.xml" "${CMAKE_CURRENT_BINARY_DIR}/xccdf-unlinked-resolved.xml"
+            COMMAND env "PYTHONPATH=$ENV{PYTHONPATH}" "${PYTHON_EXECUTABLE}" "${SSG_BUILD_SCRIPTS}/add_stig_references.py" --disa-stig "${STIG_REFERENCE_FILE}" --unlinked-xccdf "${CMAKE_CURRENT_BINARY_DIR}/xccdf-unlinked-resolved.xml"
+            DEPENDS generate-internal-${PRODUCT}-shorthand.xml
+            DEPENDS "${CMAKE_CURRENT_BINARY_DIR}/shorthand.xml"
+            DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/transforms/shorthand2xccdf.xslt"
+            DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/transforms/constants.xslt"
+            DEPENDS "${SSG_SHARED_TRANSFORMS}/shared_constants.xslt"
+            COMMENT "[${PRODUCT}-content] generating xccdf-unlinked-resolved.xml"
+        )
+    endif()
```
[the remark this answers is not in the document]
jan-cerny: Would a macro be a way?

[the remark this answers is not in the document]
cipherboy: @yuumasato Good catch that the current version (in [master](https://github.com/OpenSCAP/scap-security-guide/blob/master/cmake/SSGCommon.cmake#L107-L118)) doesn't depend on `STIG_REFERENCE_FILE`. I'll fix that as part of updating this PR (though it'll definitely need an update to add a `verbose` option, which I might not get to today).

mpreisler:
I am not in favor of a verbose option, I'd much rather have a tool that tells me coverage details about all products and profiles. I do not want to see that in build output because most of the time it's not my fault that it's there. I only want to see legit warnings and errors and if there are any it should mean I **need** to fix something.

build-scripts/combine_remediations.py
```diff
@@ -144,11 +144,11 @@ def main():
...
                         )
                 else:
-                    sys.stderr.write("Skipping '%s' remediation script. "
+                    sys.stderr.write("Warning: Skipping '%s' remediation script. "
                                      "The platform identifier in the "
                                      "script is missing!\n" % (filename))
-    sys.stderr.write("Merged %d %s remediations.\n"
-                     % (included_fixes_count, args.remediation_type))
+    #sys.stderr.write("Merged %d %s remediations.\n"
+    #                 % (included_fixes_count, args.remediation_type))
```
mpreisler: I think we should keep this, it confirms that stuff is working and the output is not overwhelming.

ssg/build_ovals.py
```diff
@@ -272,6 +272,6 @@ def checks(env_yaml, oval_version, oval_dirs):
                     body.append(xml_content)
                     included_checks_count += 1
                     already_loaded[filename] = oval_version
-    sys.stderr.write("Merged %d OVAL checks.\n" % (included_checks_count))
+    #sys.stderr.write("Merged %d OVAL checks.\n" % (included_checks_count))
```
mpreisler: same here, I think this confirms that things are working and it's just one line

build-scripts/sds_move_ocil_to_checks.py
```diff
@@ -191,8 +191,8 @@ def main():
             # Also update the ID replacing 'ecomp' with 'comp'
             move_ocil_content_from_ds_extended_component_to_ds_component(datastreamtree, oldocilhref)
 
-    else:
-        print("No extended-components, nothing to do...")
+    #else:
+    #    print("No extended-components, nothing to do...")
```
mpreisler: +100

build-scripts/unselect_empty_xccdf_groups.py
```diff
@@ -140,7 +140,7 @@ def main():
 
         affected_profiles.append(profile_element.get("id"))
 
-    print("Unselected empty groups in %s." % (", ".join(affected_profiles)))
+    # print("Unselected empty groups in %s." % (", ".join(affected_profiles)))
```
mpreisler: probably OK not to print this.

mpreisler:
I left some comments, IMO if they are addressed this can be merged. This still keeps the:
```
WARNING: OVAL check 'file_permissions_sshd_private_key' was not found, removing <check-content> element from the XCCDF rule.
```

warnings which I think we should get rid of but they need to be replaced by an easy way to list missing OVALs using `profile-stats.py`

cmake/SSGCommon.cmake
```diff
@@ -99,11 +99,25 @@ macro(ssg_build_shorthand_xml PRODUCT)
...
+        COMMAND "${XSLTPROC_EXECUTABLE}" --stringparam ssg_version "${SSG_VERSION}" --output "${CMAKE_CURRENT_BINARY_DIR}/xccdf-unlinked-resolved.xml" "${CMAKE_CURRENT_SOURCE_DIR}/transforms/shorthand2xccdf.xslt" "${CMAKE_CURRENT_BINARY_DIR}/shorthand.xml"
+        COMMAND "${OPENSCAP_OSCAP_EXECUTABLE}" xccdf resolve -o "${CMAKE_CURRENT_BINARY_DIR}/xccdf-unlinked-resolved.xml" "${CMAKE_CURRENT_BINARY_DIR}/xccdf-unlinked-resolved.xml"
+        DEPENDS generate-internal-${PRODUCT}-shorthand.xml
+        DEPENDS "${CMAKE_CURRENT_BINARY_DIR}/shorthand.xml"
+        DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/transforms/shorthand2xccdf.xslt"
+        DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/transforms/constants.xslt"
+        DEPENDS "${SSG_SHARED_TRANSFORMS}/shared_constants.xslt"
+        COMMENT "[${PRODUCT}-content] generating xccdf-unlinked-resolved.xml"
+    )
+add_custom_target(
```
mpreisler: indents

cipherboy: Oops, that one snuck by me. Updated!

cipherboy:
@mpreisler said:
> warnings which I think we should get rid of but they need to be replaced by an easy way to list missing OVALs using profile-stats.py

I have personally found those warnings to be helpful when merging rhel6->linux_os, so I'm preferential to keeping them.

mpreisler:
ACK

PR #3065, opened 2018-06-28
linux_os grub2 refactor

cipherboy:
In support of merging RHEL6 into Linux OS guide (#3037), we need to refactor the bootloader-specific stuff to make merging easier, and to be able to handle two different bootloaders (well, `grub-legacy` and `grub2` might as well be two different bootloaders due to the number of differences between configuring them and their output artifacts).

#### Description of changes 

 - Anything with `bootloader` in the rule name is moved to `grub2_*`. 
 - Modified description in `install_PAE_kernel...` to remove excessively-grub2 specific language.
 - Moved `system/accounts/accounts-physical/bootloader` to `system/bootloader-grub2` in Linux OS. 

#### Plans

If these changes are accepted, we can then create a `bootloader-grub-legacy` mirror of everything that is in `bootloader-grub2`. Alternatively, for items not in `system/bootloader-grub2`, we can later introduce Jinja macros for editing descriptions and OVALs.


cipherboy:
@mpreisler / @redhatrises -- an alternative approach would be not renaming anything and purely using jinja2 macros to accomplish all of this. Thoughts?

linux_os/guide/system/permissions/restrictions/enable_nx/install_PAE_kernel_on_x86-32.rule
```diff
@@ -13,8 +13,8 @@ description: |-
     be installed to enable XD or NX support:
     <pre>$ sudo yum install kernel-PAE</pre>
     The installation process should also have configured the
-    bootloader to load the new kernel at boot. Verify this at reboot
-    and modify <tt>/etc/default/grub</tt> if necessary.
```
redhatrises: @cipherboy this should actually stay.

cipherboy: @redhatrises :) Okie, updated with a jinja check.

redhatrises:
Ack.

PR #3066, opened 2018-06-29
Replaced the os-type-macro XSLT by jinja2 macros.

matejak:
The last user-facing XSLT macro has been subsituted by `if_product_name_is` and `if_product_pkg_system_is`.

### DIFF
--- a/shared/macros.jinja
+++ b/shared/macros.jinja
@@ -539,4 +539,3 @@ ocil_clause: "the correct value is not returned"
         <a xmlns='http://www.w3.org/1999/xhtml' href='{{{ link }}}'>{{{ link }}}</a>
     {{%- endif %}}
 {{%- endmacro %}}
-
--- a/linux_os/guide/intro/general-principles/principle-minimize-software/group.yml
+++ b/linux_os/guide/intro/general-principles/principle-minimize-software/group.yml
@@ -5,10 +5,11 @@ title: 'Minimize Software to Minimize Vulnerability'
 description: |-
     The simplest way to avoid vulnerabilities in software is to avoid
     installing that software. On {{{ full_name }}},
-    <os-type-macro type="rhel7,rhel6,fedora,opensuse,sle11,sle12,osp7,eap6,fuse6,wrlinux,ol7">the RPM Package Manager (originally Red Hat
-    Package Manager, abbreviated RPM)</os-type-macro><os-type-macro type="debian">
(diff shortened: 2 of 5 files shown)


shared/macros.jinja
```diff
@@ -540,3 +540,17 @@ ocil_clause: "the correct value is not returned"
     {{%- endif %}}
 {{%- endmacro %}}
 
+
+{{# Call e.g. {{{ if_product_name_is(rhel6="Old RHEL", rhel5="Very old RHEL") }}} #}}
+{{% macro if_product_name_is() -%}}
```
mpreisler: why invent a new code flow mechanism when jinja2 already has if? This does't bring any benefits over doing a simple `if`...

shared/macros.jinja
```diff
@@ -540,3 +540,17 @@ ocil_clause: "the correct value is not returned"
...
+    {{% if product in kwargs -%}}
+	{{{ kwargs[product] }}}
+    {{%- endif %}}
+{{%- endmacro %}}
+
+
+{{% macro if_product_pkg_system_is() -%}}
```
mpreisler: same here...

shared/macros.jinja
matejak: This design is simple, it does not abuse jinja in any way, and it allows the macro to remain the same if we add/rename a product.

shared/macros.jinja
matejak: Converting this to a `if/elif` statement is much more feasible, but still, this is code where swich/case statement would be the best, and in Python => jinja, using dictionaries instead is a [widely accepted practice](https://stackoverflow.com/questions/60208/replacements-for-switch-statement-in-python). `kwargs` is such dictionary.

shared/macros.jinja
mpreisler: my comment is not about the actual implementation of the macro, that's fine. my comment is about why you would even define a macro for this when you can use language features that are much easier to read for everyone

matejak: OK, I see your point now. The macro is less verbose, but still readable. However, as the functionality is not used in many files, I agree with you as I give it a second thought.

matejak:
PR design has been changed in the rebase, so no macros, but only directives are used.

mpreisler:
Removed the usability label because the PR has nothing to do with usability.

ACK

PR #3069, opened 2018-07-02
Deduplicated partition tests in the test suite.

matejak:
Took advantage of the possibility to share common functionality between
different tests.


mildass:
ACK.
Test `dvd_bad_opts.fail.sh` is failing, but it's not because of this PR. It was failing before so PR is ok and merging.

PR #3071, opened 2018-07-02
Test suite python3

cipherboy:
Depends on #3060.

In this PR, I fix python3 support in the `test_suite` (e.g., the rule virtual machine test suite). 

The only things that are really wrong are that `data/__init__.py` doesn't use absolute imports and that `parse_args` uses two parsers (`parser` and `common_parser`)... I removed `parser` and rename `common_parser` -> `parser`.

Now it should run on python3 as well! :)

### DIFF
--- a/tests/data/__init__.py
+++ b/tests/data/__init__.py
@@ -1,3 +1,3 @@
-from utils import iterate_over_rules
-from utils import create_tarball
-from utils import _DIR as DATA_DIR
+from data.utils import iterate_over_rules
+from data.utils import create_tarball
+from data.utils import _DIR as DATA_DIR
--- a/tests/test_suite.py
+++ b/tests/test_suite.py
@@ -17,48 +17,46 @@ from ssg_test_suite import xml_operations
 
 
 def parse_args():
-    parser = argparse.ArgumentParser()
+    parser = argparse.ArgumentParser(add_help=False)
+    parser.add_argument("--hypervisor",
+                        dest="hypervisor",
+                        metavar="HYPERVISOR",
+                        default="qemu:///session",
+                        help="libvirt hypervisor")
+    parser.add_argument("--domain",
+                        dest="domain_name",
+                        metavar="DOMAIN",
+                        required=True,
+                        help=("Specify libvirt domain to be used as a test "
+                              "bed. This domain will get remediations "
+                              "applied in it, possibly making system "
+                              "unusable for a moment. Snapshot will be "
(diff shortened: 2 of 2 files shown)


dahaic:
@cipherboy @mpreisler have you tested it? I am pretty sure this broke the interface. For example I am not able to run it, and this is also regression:
```
[dahaic@dhcp-24-168 tests]$ ./test_suite.py --help
usage: test_suite.py [--hypervisor HYPERVISOR] --domain DOMAIN --datastream
                     DATASTREAM --xccdf-id REF-ID [--loglevel LOGLEVEL]
                     [--logdir LOGDIR]
                     [--remediate-using {oscap,ansible,bash}]
                     {profile,rule} ...
test_suite.py: error: too few arguments
```

cipherboy:
To expand on @dahaic's comment...

If you revert the changes to `test_suite.py` introduced in the second commit, you can perform what this stackoverflow link suggests: https://stackoverflow.com/questions/22990977/why-does-this-argparse-code-behave-differently-between-python-2-and-3

In particular, it looks like it was an explicit difference between 2 and 3.

My change works if you're willing to move the location of 'rule' in the parser, e.g., to just before the rules:

```
python2 ./test_suite.py --hypervisor 'qemu+ssh://root@remote.cipherboy.com/system' --domain ographics --datastream ../build/ssg-centos7-ds.xml --xccdf-id scap_org.open-scap_cref_ssg-rhel7-xccdf-1.2.xml rule ipv4_conf_default_secure_redirects
```

dahaic:
No, it does not work for me, I have tried. :)
```
[dahaic@dhcp-24-168 tests]$ sudo ./test_suite.py --hypervisor qemu:///system --domain ssg-test-suite-rhel7 --datastream ../build/ssg-rhel7-ds.xml --xccdf-id scap_org.open-scap_cref_ssg-rhel7-xccdf-1.2.xml profile stig-rhel7-disa
usage: test_suite.py profile [-h] [--hypervisor HYPERVISOR] --domain DOMAIN
                             --datastream DATASTREAM --xccdf-id REF-ID
                             [--loglevel LOGLEVEL] [--logdir LOGDIR]
                             [--remediate-using {oscap,ansible,bash}]
                             {profile,rule} ... DSPROFILE [DSPROFILE ...]
test_suite.py profile: error: argument subparser_name: invalid choice: 'stig-rhel7-disa' (choose from 'profile', 'rule')
```
(Rule variant fails same way)

PR #3073, opened 2018-07-03
Add Ubuntu to Travis CI

jan-cerny:
#### Description:

Adds a simple Dockerfile to build SSG in a container based
on Ubuntu 18.04. Then it adds a Travis CI job that builds
SSG and runs tests in that container. The Travis configuration
had to be changed to define a build matrix.

#### Rationale:
Adds another OS - Ubuntu - that we don't test the build on yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,15 @@ services:
 
 language: python
 
-before_install:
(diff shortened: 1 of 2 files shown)


dahaic:
Thank you @jan-cerny i'll be great to have this available.

PR #3075, opened 2018-07-03
Update unit tests to work with new SSG layout and Python3

matejak:


tests/CMakeLists.txt
```diff
@@ -15,6 +15,11 @@ endmacro()
 if (PY_PYTEST)
     ssg_python_unit_tests("build-scripts")
     ssg_python_unit_tests("utils")
+    # The SSG unit test is special - the ssg subdir can't be in the PYTHONPATH.
+    add_test(
+        NAME "python-unit-ssg"
+        COMMAND "${PYTHON_EXECUTABLE}" -m pytest ${PYTEST_COVERAGE_OPTIONS} "${CMAKE_SOURCE_DIR}/tests/unit/ssg"
```
mpreisler: could we rename it to ssg_module? sounds easier and will bring way less "wtf" moments in the future.

matejak: Good suggestion, fixed in the rebase.

matejak:
Scrutinizer runs the tests without running into problems. I have added a small test for oval.py, and fixed a bug in the process.

ssg/oval.py
```diff
@@ -14,7 +14,7 @@
 from .xml import oval_generated_header
 from .id_translate import IDTranslator
 
-SHARED_OVAL = re.sub(r'shared.*', 'shared', __file__) + '/checks/oval/'
+SHARED_OVAL = re.sub(r'ssg/.*', 'shared', __file__) + '/checks/oval/'
```
yuumasato: I'm intrigued by why `SHARED_OVAL` is defined this way, but I see that totally changing it is not part of this PR.

ssg/oval.py
```diff
@@ -106,24 +106,27 @@ def replace_external_vars(tree):
...
-              % testfile, file=sys.stderr)
-        sys.exit(1)
+            searchfile = os.path.join(root, testfile).strip()
+            if os.path.isfile(searchfile):
+                return searchfile
 
-    return testfile
+    # If we have not returned, we haven't found anything
+    msg = "%s does not exist!" % testfile
+    raise RuntimeError(msg)
```
yuumasato: This probably should be an `IOError`.

tests/CMakeLists.txt
```diff
@@ -15,6 +15,14 @@ endmacro()
 if (PY_PYTEST)
     ssg_python_unit_tests("build-scripts")
     ssg_python_unit_tests("utils")
+    # The SSG unit test is special - the ssg subdir can't be in the PYTHONPATH.
```
yuumasato: Can you clarify why it is special?

matejak: I have rebased to expand the comment.

ssg/oval.py
matejak: I took a look at [Python3 docs](https://docs.python.org/3/library/exceptions.html#IOError), and `IOError` seems to be appropriate when the OS can't read from a file, as it is an alias to `OSError`. However, I gave it another thought and written the code in a way that it doesn't use exceptions.

yuumasato:
LGTM, thanks.

tests/CMakeLists.txt
mpreisler: Was this done or not? The hunk doesn't seem to have it.

mpreisler:
Wait, I thought we were gonna rename the ssg unit tests to something else to avoid the while PYTHONPATH problem... It was one of the first comments in this PR.

yuumasato:
I get it now, I only considered the name of the test, not the name of the module itself.

mpreisler:
I don't understand your comment at all though

PR #3076, opened 2018-07-03
Unshadow the xwindows_runlevel OVAL

mpreisler:


redhatrises:
This will most likely be changed later, but ack for now

PR #3085, opened 2018-07-04
Unshadow rhel6 audit_rules_time_

yuumasato:
#### Description:

- Basically dropping rhel6 OVAL definitions in favor of the shared definitions.

#### Rationale:
- The regexes are more flexible with regards to placement of syscal in the rule, and they allow key to be specified as `-k` or `-F key=`.
- Also enables check of rules for augenrules in rhel6.


dahaic:
I did profile check with SSG Test Suite and behaviour hasn't changed, looks good. ACK

PR #3086, opened 2018-07-04
No OVAL for xwindows_runlevel_setting when product uses systemd and targets OVAL 5.10

yuumasato:
#### Description:

- OVAL definition for `xwindows_runlevel_setting` will be empty when product `init_system == systemd` and target OVAL is 5.10.

#### Rationale:
- This definition requires that we check which file the symbolic link is referencing, and AFAIK there is no way to it with OVAL 5.10.
- If `oval_version` is defined as [previously](https://github.com/OpenSCAP/scap-security-guide/pull/3076/files#diff-063702df45814951fc68f86d648e1113L1), the definition will not be available at all when target OVAL is 5.10.
- This way, the OVAL definition will still be available for RHEL6 when target OVAL is 5.10
- Fixes validation of OVAL 5.10 content


yuumasato:
This looks hackish, but I don't know how to improve it, :-)

dahaic:
Nah, it's not hackish. Looks good to me. ACK

PR #3087, opened 2018-07-04
Removed obsoleted check macro xsl tests.

dahaic:
#### Description:

- Removal of tests within xslt that bears no value anymore. 

#### Rationale:

- With replacement of macros with jinja (see #2835 )this test no longer serves it's purpose and can be removed.

### DIFF
--- a/shared/transforms/shared_shorthand2xccdf.xslt
+++ b/shared/transforms/shared_shorthand2xccdf.xslt
@@ -390,14 +390,7 @@
           <xsl:attribute name="system">ocil-transitional</xsl:attribute>
           <check-export>
 
-            <xsl:attribute name="export-name">
-              <!-- add clauses if specific macros are found within -->
-              <xsl:if test="fileperms-check-macro or fileowner-check-macro or filegroupowner-check-macro">it does not</xsl:if>
-              <xsl:if test="service-disable-check-macro">the service is running</xsl:if>
-              <xsl:if test="socket-disable-check-macro">the socket is running</xsl:if>
-              <xsl:if test="service-enable-check-macro">the service is not running</xsl:if>
-              <xsl:if test="module-disable-check-macro">no line is returned</xsl:if>
-            </xsl:attribute>
+            <xsl:attribute name="export-name"/>
 
             <!-- add clause if explicitly specified (and also override any above) -->
             <xsl:if test="@clause">
--- a/shared/transforms/shared_shorthand2xccdf.xslt
+++ b/shared/transforms/shared_shorthand2xccdf.xslt
@@ -389,14 +389,7 @@
         <check>
           <xsl:attribute name="system">ocil-transitional</xsl:attribute>
           <check-export>
-
-            <xsl:attribute name="export-name"/>
-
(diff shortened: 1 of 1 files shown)


shared/transforms/shared_shorthand2xccdf.xslt
```diff
@@ -390,14 +390,7 @@
...
 
-            <xsl:attribute name="export-name">
-              <!-- add clauses if specific macros are found within -->
-              <xsl:if test="fileperms-check-macro or fileowner-check-macro or filegroupowner-check-macro">it does not</xsl:if>
-              <xsl:if test="service-disable-check-macro">the service is running</xsl:if>
-              <xsl:if test="socket-disable-check-macro">the socket is running</xsl:if>
-              <xsl:if test="service-enable-check-macro">the service is not running</xsl:if>
-              <xsl:if test="module-disable-check-macro">no line is returned</xsl:if>
-            </xsl:attribute>
+            <xsl:attribute name="export-name"/>
```
yuumasato: Is this attribute without value needed?
I think this can be removed.

dahaic: Don't ask me why, but it looks like it is :D When I removed it, it failed to build...

mpreisler: NACK, this needs to be fixed for real

dahaic: Sorry, I don't understand what does it mean "fixed for real". This PR is meant to remove macro artefacts no longer in use. It's not fixing anything.

dahaic: AFAICT the macros were replaced correctly using Jinja, so there is nothing to fix.

mpreisler: I mean removing the attribute completely, don't leave it there because it doesn't build. We need to understand why it fails and remove all traces of it.

dahaic: Ah, ok. Makes sense

dahaic: OCIL questions end with sentence
`Is it the case that` which continues with the contents of "export-name". So the attribute is still needed, unless we'll remove it altogether and update OCIL question directly. So I cannot remove the attribute in this "cleanup" PR. I suggest to investigate streamlining OCIL generation is separate effort.

mpreisler: well OK but then you broke that with this PR so we can't merge it

dahaic: I am not aware of any breakage - part I remove in the PR is replacement of macros with texts. With macros rewrite to jinja, there are no more macros to replace in this phase.
`xccdf-unlinked-resolved.xml` does not differ between master and my PR.

dahaic: @mpreisler thanks for tip how to make the code slightly nicer. Updated.

PR #3088, opened 2018-07-04
Unshadowed some stray osp7 ovals.

dahaic:
#### Description:

- OVALs in rhel-osp7 product that were shared with shared were removed

#### Rationale:

- Part of unshadowing initiative (i.e. only single source of rule exists)


yuumasato:
Thanks, Ack.

PR #3089, opened 2018-07-05
exec-shield requires special handling, don't use the template for it …

mpreisler:
Unshadows the check on RHEL6


cipherboy:
Why leave the line in the csv file?

mpreisler:
> Why leave the line in the csv file?

I did that mostly because that's the way it is in RHEL7, it also reassures me that it's intentionally not enabled in the CSV file.

cipherboy:
Fair enough. Looks like that's what @ybznek did for `rhel7`. ACK.

PR #3090, opened 2018-07-05
Unshadow bash fix display_login_attempts on RHEL6

mpreisler:


cipherboy:
Does RHEL6 not have `postlogin` in `/etc/pam.d`?

mpreisler:
> Does RHEL6 not have postlogin in /etc/pam.d?

not sure. Either way, let's not change how it works on RHEL6 in this PR.

cipherboy:
Fair enough. ACK.

PR #3091, opened 2018-07-05
Unshadow sshd set keepalive and sshd set idle timeout

mpreisler:


cipherboy:
LGTM.

cipherboy:
Thanks, @mpreisler!

PR #3092, opened 2018-07-05
Unshadow network_ipv6_disable_rpc on RHEL6

mpreisler:
We are actually using the RHEL6 bash fix even for RHEL7, it was better
quality.


cipherboy:
ACK. Thanks @mpreisler!

PR #3093, opened 2018-07-05
Unshadow disable ctrlaltdel

mpreisler:
Uses the trick by @yuumasato to avoid invalid OVAL 5.10 on systemd systems.

See #3086


cipherboy:
LGTM. Thanks @mpreisler!

PR #3070, opened 2018-07-02
[FEEDBACK REQUESTED] Merge system in rhel6, linux_os

cipherboy:
Continuation of #3037. Independent of #3068 (should not merge conflict).

#### Description

This continues the merger of `rhel6/guide` into `linux_os/guide`. After this, everything in `rhel6/guide/system` will be in `linux_os/guide/system. `prodtype` shouldn't need to be adjusted as a previous PR (namely, #3037) added `rhel6` to all 1-1 duplicated files.

This PR:
 - Adds file that are in `rhel6` but not in `linux_os`
 - Updates descriptions and syntax changes
 - Fixes rhel6/rhel7 differences.

This was performed by walking the rhel6 tree, diffing the files it finds compared to the equivalent linux_os folder and manually comparing the output and deciding what happens. Please review this carefully as this was a lot of manual work and in some cases I might have chosen incorrectly. :)

To review, I'd probably first:
 - Look at the diff and make sure the diff seem reasonable.
 - Look at the parallels of the files changed (the equivalent file in `rhel6` that was changed in `linux_os`).
 - Perform a file-name based diff of the contents of `rhel6/guide/system` and `linux_os/guide/system` and see what differs.

But to each their own. :)

#### Plans

After this, I'll work on doing a second review of `rhel6` and `linux_os` trees, and work on converting the profiles in `rhel6` to point to the ones in `linux_os` and diff the outputted build artifacts. Then we can flip the switch to using `linux_os` in `rhel6`.

PR #3095, opened 2018-07-05
Unshadow enable_selinux_bootloader bash fix on RHEL6

mpreisler:


cipherboy:
In #3070, I'm moving a lot of these bootloader ones into separate grub2 / grub-legacy checks. Do we want to keep them separate in the OVAL as well? I guess a lot of this code is duplicated, so perhaps we should just write jinja macros to substitute in the correct config path.

(@redhatrises ACKed #3065, but that didn't update the OVAL checks to rename them to grub2 and not much feedback was given on it).


Thoughts from either @mpreisler or @redhatrises? These thoughts apply to #3095 and #3094.

redhatrises:
@cipherboy yes let's update to separate OVAL, but we will do that after all the XCCDF (guide) content as been merged.

PR #3098, opened 2018-07-06
Unshadow dir_perms_world_writable_system_owned on RHEL6

mpreisler:
Same logic @yuumasato used in #3009 except I flipped it to make it easier to read.


cipherboy:
@mpreisler Do you want to add Fedora support too?

mpreisler:
> @mpreisler Do you want to add Fedora support too?
DONE

redhatrises:
Ack.

PR #3097, opened 2018-07-06
RHEL6, RHEL7, Fedora -> shared ovals

cipherboy:
As part of #3037 efforts...

Move ovals shared by `rhel6` and `rhel7` into the shared folder and mark them as `multi_platform_{rhel,fedora}` as in many cases these checks are agnostic of providers.

Also updates the names of some `rhel6` checks to match their new names under #3068 / #3070 titles.

This is required for #3072 to work.

PR #3099, opened 2018-07-06
Unshadow accounts_password_pam_retry

mpreisler:
We already have a fairly universal check in shared for this.


cipherboy:
@mpreisler See #3097.

cipherboy:
NACK per discussion with @mpreisler.

PR #3100, opened 2018-07-06
Unshadow configure ldap ca

mpreisler:


jan-cerny:
The generated OVAL content for both RHEL 6 and 7 looks good. Thank you.

PR #3102, opened 2018-07-06
Unshadow require single user

mpreisler:


jan-cerny:
Hi, do you plan to remove `rhel6/fixes/bash/require_singleuser_auth.sh`?

yuumasato:
LGTM, ACK.

shared/fixes/bash/require_singleuser_auth.sh
```diff
@@ -1,5 +1,13 @@
...
+# platform = multi_platform_rhel
+{{%- if init_system == "systemd" -%}}
 grep -q "^ExecStart=\-.*/sbin/sulogin" /usr/lib/systemd/system/rescue.service
 if ! [ $? -eq 0 ]; then
     sed -i "s/ExecStart=-.*-c \"/&\/sbin\/sulogin; /g" /usr/lib/systemd/system/rescue.service
 fi
+{{%- else -%}}
+grep -q ^SINGLE /etc/sysconfig/init && \
+  sed -i "s/SINGLE.*/SINGLE=\/sbin\/sulogin/g" /etc/sysconfig/init
+if ! [ $? -eq 0 ]; then
```
cipherboy: Since this is in the original, it shouldn't be fixed in this PR, but why do we use the ` ! [ -eq ]` pattern instead of `[ -ne ]` pattern?

cipherboy:
LGTM.

PR #3107, opened 2018-07-09
Unshadow audit_rules_file_deletion_events

yuumasato:
#### Description:
- Enabled RHEL6 in shared `audit_rules_file_deletion_events` and template of extended definitions.

#### Rationale:

- Favored shared OVAL definition of `audit_rules_file_deletion_events` as they have better regular expression and also check for configuration done via augenrules


shawndwells:
This particular rule was created to align with DISA.

The DISA content has one rule that attests multiple things (``rename``, ``renameat``, ``unlinkat``....). In SSG we break them into independent rules.

New profiles shouldn't be selecting this rule. If this moves to ``shared/``, perhaps a comment could be added to use the ``audit_rules_file_deletion_*`` rules instead?

yuumasato:
@shawndwells Nothing is changed regarding behavior and presentation of rule in this PR, `audit_rules_file_deletion_events` will still check all syscalls.

I believe this warning should be part of the rule description, and done in different PR, can you please check #3113?

yuumasato:
Anything else regarding this PR? 
From my point of view this is not blocked, the issue raised by @shawndwells will be addressed in #3113.

jan-cerny:
I have reviewed the PR and checked the RHEL6 build artifacts. The changes look OK to me. Thanks for your contribution.

PR #3108, opened 2018-07-09
Unshadow audit_rules_media_export

yuumasato:
#### Description:

- Dropping rhel6 definition in favor of the shared definition.

#### Rationale:

- The regexes are more flexible with regards to placement of syscal in the rule, and they allow key to be specified as -k or -F key=.
- Also enables check of rules for augenrules in rhel6.


shared/checks/oval/audit_rules_media_export.xml
```diff
@@ -31,13 +31,17 @@
       </criteria>
     </criteria>
   </definition>
-
+  {{% if product == "rhel6" %}}
+    {{% set AUID = 500 %}}
+  {{% else %}}
+    {{% set AUID = 1000 %}}
+  {{% endif %}}
```
mpreisler: this is repeated so much that I think we should somehow share the code. We could have `auid: 1000` in product.yml or create a jinja2 macro.

yuumasato: That is a good idea, I have proposed a solution in #3114. Once merged I'll updated this PR.

yuumasato: This has been updated to use the new `auid` value.

PR #3109, opened 2018-07-09
Unshadow RHEL6 OVAL and BASH audit_rules_immutable

yuumasato:
#### Description:

- Drop rhel6 exclusive OVAL and bash fix for rule `audit_rules_immutable`

#### Rationale:

- Both the definition and remediation from shared check and fix rules generated by augenrules.


jan-cerny:
I have checked it and the resulting OVAL and XCCDF for both RHEL 6 and 7 looks OK wrt this rule. ACK.

PR #3111, opened 2018-07-09
Add rule_sssd_ldap_configure_tls_ca_dir

mildass:
#### Description:

Adds OVAL, bash remediation, ansible remediation and tests for sssd_ldap_configure_tls_ca_dir rule.

Changes library function that is used also in sssd_ldap_start_tls.


dahaic:
@mildass first of all, thanks for the PR, and for the content within. It a good stuff.

Amount of code that I had to update is a bit higher then I would expect, though :) I would strongly recommend to focus more on the test side of the delivery next time. Tests were not enabled for ansible remediations, (and ansible remediation had issue that was discoverable by them). There has been positive scenario missing, which would point out to the issue with test basis (configuration was not correct, directory value differed to what we expect in the profile). And one of the scenarios delivered was copy-paste error and did nothing (which was not discovered because test basis was mistakenly broken ;) )

So overall it's ok, it was not hard to fix those, but please be more thorough next time :)

dahaic:
Jenkins test this please.

jan-cerny:
It works OK with my Centos VM. Thank you for your contribution.

PR #3112, opened 2018-07-10
[WIP] Restructure Dockerfiles to improve caching

cipherboy:
Our Dockerfiles don't support caching well; we install updates (which change frequently) before installing our base dependencies; these dependencies thus cannot be cached. 

Since updating, installing dependencies, removing unnecessary files, and creating the home directory was all one step, we could not cache any of these steps independently of the others. This switches the order so that the most-static steps are executed first, followed by the most-dynamic steps.

Note that I kept the `rm -rf`s and added an `apt-get -qq clean all` to Ubuntu; these reduce the size of the generated images, but likely not by a significant amount since they exist in the base image, which must also be cached.

### DIFF
--- a/Dockerfiles/fedora_python2
+++ b/Dockerfiles/fedora_python2
@@ -4,10 +4,9 @@ ENV OSCAP_USERNAME oscap
 ENV OSCAP_DIR scap-security-guide
 ENV BUILD_JOBS 4
 
-RUN dnf -y upgrade && \
-    dnf -y install cmake ninja-build openscap-utils python2-jinja2 python2-pyyaml && \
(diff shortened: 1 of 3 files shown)


cipherboy:
This is ready for review, but leaving as WIP because I'd like to get CentOS 6 merged first and then update this one for CentOS 6 and then consider merging this one.

matejak:
NACK, there is a reason why everything is on one line - it reduces the image size.
On a slightly related note, I encourage to follow the [Dockerfile best practices guide](https://github.com/projectatomic/container-best-practices/blob/master/creating/creating_index.adoc#chaining-commands) and use the `&& true` trick. The guide also contains explanation why to chain those commands.

PR #3113, opened 2018-07-10
Add disa stig warnings to aggregated audit rules

yuumasato:
#### Description:

- Add a warning that recommends users to select individual rules regarding auditing of file deletion events.

#### Rationale:

- Some audit rules are responsible for checking many syscalls, these rules were written with DISA use case in mind and are not recommended to be used in other profiles.

### DIFF
--- a/linux_os/guide/system/auditing/auditd_configure_rules/audit_file_deletion_events/audit_rules_file_deletion_events.rule
+++ b/linux_os/guide/system/auditing/auditd_configure_rules/audit_file_deletion_events/audit_rules_file_deletion_events.rule
@@ -54,3 +54,14 @@ ocil: |-
     {{{ ocil_audit_syscall(syscall="renameat") }}}
 
 {{{ ocil_clause_entry_audit_syscall() }}}
+
+warnings:
+    - general: |-
+        This rule checks for multiple syscalls related to file deletion;
+        it was written with DISA STIG in mind. Other policies should use a
+        separate rule for each syscall that needs to be checked. For example:
+        <ul>
+        <li><tt>audit_rules_file_deletion_events_rmdir</tt></li>
+        <li><tt>audit_rules_file_deletion_events_unlink</tt></li>
+        <li><tt>audit_rules_file_deletion_events_unlinkat</tt></li>
+        </ul>
--- a/linux_os/guide/system/auditing/auditd_configure_rules/audit_kernel_module_loading/audit_rules_kernel_module_loading.rule
+++ b/linux_os/guide/system/auditing/auditd_configure_rules/audit_kernel_module_loading/audit_rules_kernel_module_loading.rule
@@ -53,3 +53,14 @@ ocil: |-
     {{{ ocil_audit_syscall(syscall="delete_module") }}}
 
 {{{ ocil_clause_entry_audit_syscall() }}}
+
+warnings:
+    - general: |-
+        This rule checks for multiple syscalls related to kernel module loading and unloading;
+        it was written with DISA STIG in mind. Other policies should use a
+        separate rule for each syscall that needs to be checked. For example:
+        <ul>
+        <li><tt>audit_rules_kernel_module_loading_insmod</tt></li>
+        <li><tt>audit_rules_kernel_module_loading_rmmod</tt></li>
+        <li><tt>audit_rules_kernel_module_loading_modprobe</tt></li>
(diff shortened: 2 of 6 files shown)


yuumasato:
I believe other rules should have this warnings too, I'll look for them later.

linux_os/guide/system/auditing/auditd_configure_rules/audit_file_deletion_events/audit_rules_file_deletion_events.rule
```diff
@@ -54,3 +54,14 @@ ocil: |-
     {{{ ocil_audit_syscall(syscall="renameat") }}}
 
 {{{ ocil_clause_entry_audit_syscall() }}}
+
+warnings:
+    - general: |-
+        This rule aims to attest collection of events from multiple syscalls
+        related to file deletion. It is written with DISA STIG in mind, other
+        policies should use separated rules for each syscall, namely:
```
redhatrises: @yuumasato I don't think that we need to mention the rules below as it becomes another maintenance burden if this list changes. Let's either remove the mentioned audit rules, or change `, namely:` to `. For example:`

yuumasato: @shawndwells We should also consider scenario of someone tailoring a Profile and adding these "combined" audit rules, they won't see the developer warning.

yuumasato:
I have updated the warning, and extended it to other similar rules. Let me know If i missed any.

The warning looks like this in the guide.
rhel7-pci-dss-guide


> We should also consider scenario of someone tailoring a Profile and adding these "combined" audit rules, they won't see the developer warning.

But, SCAP Workbench doesn't show warnings of rules in its interface....

yuumasato:
@shawndwells Ironically DISA STIG profile doesn't select `audit_rules_file_deletion_events`. Is this because the DISA content changed and they are also moving away from these aggregated rules?

linux_os/guide/system/auditing/auditd_configure_rules/audit_file_deletion_events/audit_rules_file_deletion_events.rule
```diff
@@ -54,3 +54,14 @@ ocil: |-
     {{{ ocil_audit_syscall(syscall="renameat") }}}
 
 {{{ ocil_clause_entry_audit_syscall() }}}
+
+warnings:
+    - general: |-
+        This rule aims to attest collection of events from multiple syscalls
```
cipherboy: I'm not sure this sentence makes sense. I think you mean "This rule aims to test a collection"... or "attest to a collection of events"... But even then I'm not entirely sure it makes sense. Do you mind expanding on what you mean here?

linux_os/guide/system/auditing/auditd_configure_rules/audit_file_deletion_events/audit_rules_file_deletion_events.rule
```diff
@@ -54,3 +54,14 @@ ocil: |-
...
+        related to file deletion. It was written with DISA STIG in mind, other
```
cipherboy: I'd use a semicolon instead of a comma because these are two related but otherwise independent sentences.

linux_os/guide/system/auditing/auditd_configure_rules/audit_file_deletion_events/audit_rules_file_deletion_events.rule
```diff
@@ -54,3 +54,14 @@ ocil: |-
...
+        policies should use separated rules for each syscall. For example:
```
cipherboy: "should use rules separated"
or
"should use separate rules"

linux_os/guide/system/auditing/auditd_configure_rules/audit_file_deletion_events/audit_rules_file_deletion_events.rule
yuumasato: I've rephrased, It is actually shorter now, hope it is clearer.

yuumasato:
Based on feedback I've improved the warning.

cipherboy:
Thansk @yuumasato! I like the new changes :) ACK from me.

mpreisler:
ACk from me too

PR #3114, opened 2018-07-10
Add uid_min and implied auid product.yml

yuumasato:
Avoid setting value of `auid` in every definition or fix. Lets set them product wise.

#### Description:

- Adds `uid_min` to product.yml.
  - If not defined, its default value is 1000
- `auid` is implied and set to `uid_min`.
- I have made them two separate values, because there are rules based on value of `uid_min` and there are the auditd rules that use `auid`, and it may be confusing to use them interchangeably.

#### Rationale:

- The `auid` parameter used in `auditd` rules should be set according to `uid_min` of the platform.

### DIFF
--- a/ssg/yaml.py
+++ b/ssg/yaml.py
@@ -8,7 +8,7 @@ from .jinja import _rename_items
 from .jinja import process_file
 from .constants import (PKG_MANAGER_TO_SYSTEM,
                         JINJA_MACROS_BASE_DEFINITIONS, JINJA_MACROS_HIGHLEVEL_DEFINITIONS)
-from .constants import default_uid_min
+from .constants import DEFAULT_UID_MIN
 
 try:
     from yaml import CSafeLoader as yaml_SafeLoader
@@ -48,8 +48,8 @@ def _get_implied_properties(existing_properties):
         pkg_manager = existing_properties["pkg_manager"]
         result["pkg_system"] = PKG_MANAGER_TO_SYSTEM[pkg_manager]
 
-    if ("uid_min" not in existing_properties):
-        result["uid_min"] = result["auid"] = default_uid_min
+    if "uid_min" not in existing_properties:
+        result["uid_min"] = result["auid"] = DEFAULT_UID_MIN
     else:
         result["auid"] = existing_properties["uid_min"]
 
--- a/ssg/yaml.py
+++ b/ssg/yaml.py
@@ -49,9 +49,13 @@ def _get_implied_properties(existing_properties):
         result["pkg_system"] = PKG_MANAGER_TO_SYSTEM[pkg_manager]
 
     if "uid_min" not in existing_properties:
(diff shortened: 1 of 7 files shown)


ssg/yaml.py
```diff
@@ -46,6 +47,12 @@ def _get_implied_properties(existing_properties):
             "pkg_system" not in existing_properties):
         pkg_manager = existing_properties["pkg_manager"]
         result["pkg_system"] = PKG_MANAGER_TO_SYSTEM[pkg_manager]
+
+    if "uid_min" not in existing_properties:
+        result["uid_min"] = result["auid"] = DEFAULT_UID_MIN
+    Else:
```
mpreisler: `Else`?

yuumasato: /o\

ssg/yaml.py
```diff
@@ -46,6 +47,12 @@ def _get_implied_properties(existing_properties):
             "pkg_system" not in existing_properties):
         pkg_manager = existing_properties["pkg_manager"]
         result["pkg_system"] = PKG_MANAGER_TO_SYSTEM[pkg_manager]
+
+    if "uid_min" not in existing_properties:
+        result["uid_min"] = result["auid"] = DEFAULT_UID_MIN
+    else:
+        result["auid"] = existing_properties["uid_min"]
```
mpreisler: This is doing two things, setting uid_min if it's not set in product and then setting auid according to uid_min. It doesn't respect any auid override so it's not clear to me why there even are 2 variables. Regardless, the result["auid"] logic should be a separate line. Handle uid_min, then handle auid, ...

yuumasato: I have updated the logic to not override `auid` if it is already set.

jan-cerny:
I have checked the output OVALs for RHEL 6, 7 and Fedora.

IMO @mpreisler's comments have been adressed.

I think this can be merged safely.

Thank you for your contribution.

ssg/yaml.py
```diff
@@ -46,6 +47,16 @@ def _get_implied_properties(existing_properties):
             "pkg_system" not in existing_properties):
         pkg_manager = existing_properties["pkg_manager"]
         result["pkg_system"] = PKG_MANAGER_TO_SYSTEM[pkg_manager]
+
+    if "uid_min" not in existing_properties:
+        result["uid_min"] = DEFAULT_UID_MIN
+
+    if "auid" not in existing_properties:
+        if "uid_min" in existing_properties:
+            result["auid"] = existing_properties["uid_min"]
```
mpreisler: there is a much nicer way to do this in Python, existing_properties.get("uid_min", DEFAULT_UID_MIN). It's much simpler and shorter

mpreisler:
I wish the code was nicer but it's not a big deal. I added a comment anyway for future reference...

PR #3115, opened 2018-07-10
ol7 pci-dss profile init

iokomin:
#### Description:
Suggested initial pci-dss profile draft for ol7 product. In general it is just porting shared rhel7 rules to ol7.

Affected changes:
- Created pci-dss profile with related existing ol7 rules included
- Number of pci-dss file permissions/ownership rules adopted for ol7
- libreswan package rule adopted for ol7
- Made ipsec and grub2 guide groups content minor updates to be generic and support ol7 product

#### Rationale:
Add pci-dss profile to ol7 product

#### Testing (Oracle Linux 7):
- Checked rules and remediation script content
- Checked generated guide content
- Checked rules evaluation and remediation scripts results


ol7/profiles/pci-dss.profile
```diff
@@ -0,0 +1,111 @@
+documentation_complete: true
+
+title: 'PCI-DSS v3 Control Baseline Draft for Oracle Linux 7'
+
+description: 'Ensures PCI-DSS v3 related security configuration settings are applied.'
+
+selections:
```
matejak: I don't see the benefit of having those commented lines here.
I guess that you want to enable those rules later, but having those lines doesn't have any benefit. Anybody can use tools based on `diff` to see the difference between the `ol7` and e.g. `rhel7` PCI-DSS profile to see what rules are missing. The git-centric point of view when you add a commented line, which you later substitute for line without comment doesn't make sense either.

matejak:
Thank you for this pull request! I have checked it, and I confirm that it works as you indicate in your PR comment - rules and the guide are correct.
I have a remark about comments, I believe that code to-be-uncommented shouldn't belong to this PR.
Aside from that, I am in favor of merging it.

ol7/profiles/pci-dss.profile
shawndwells: Far easier to comment out rules that need to be enabled later than create 100+ github tickets to track things. I'm guilty of using comments as placeholders too. 

Yes, tools like ``diff`` could be used to see differences vs github tickets. Each developer has their own workflow -- for myself, having to use ``diff`` and figure out how to parse the results would be incredibly annoying. Especially when trying to update 100+ rules.... running diff 100+ times? No way.

In the pursuit of project usability/accessibility/ease of contributing vs elegant git logs, ease of contributing should be favored. Suggest we nudge PRs to use "clean" commits as @matejak suggested but not block PRs over the issue.

iokomin: Guys, thanks for your comments.
I just considered having entire list of rules would be easier to track and maintain. However no problem with removing commented out rules as I definitely can keep this list outside of source project.
Could you please have a look at the updated request?
Note: Due to number of origins on my system incorrect commit had been pushed by mistake first. It is updated, rebased and hope fine now.

cipherboy:
@iokomin: Sorry! Looks like I caused a merge conflict in your PR. Thanks for updating it!

I lack an OL7 system so I can't comment on whether or not the fixes are applicable to OL7 but they seem sane. I think the PCI-DSS profile is a bit light but that can be improved in future PRs. 

So, ACK from me as well.

shawndwells:
Merging since comments were removed. 

Did basic sanity check - but like @cipherboy, no local access to OL7.

PR #3103, opened 2018-07-06
Travis - CentOS 6

cipherboy:
Dependencies:
 - ~#3101~ -- merged, thanks @mpreisler!
 - ~#3116~ -- merged, thanks @jan-cerny 

Adds CentOS 6 docker build to Travis CI. This helps prevent make-only build failures, allows us to test older Python versions (2.6). 

~I'll rebase this PR after #3116 merges.~ Rebased.

PR #3116, opened 2018-07-10
Fix parallel make regression

cipherboy:
- Fixes dependencies for `generate_from_template.py` (needs to depend on all `shared/fixes/{lang}` and `{product}/fixes/{lang}` files).
 - Uses add_custom_target to prevent multiple parallel runs of `generate_from_template.py` (otherwise, the second `add_custom_command` may trigger multiple `add_custom_command` (and `generate_from_template.py`) calls, which causes problems due to both calling `cmake -E remove_directory`.)

Regression in: #3010 

This should help the RHEL6/CENTOS6 build in #3103.


jan-cerny:
With this PR, I haven't experienced any problems when building SSG inside a CentOS 6 container and the test passed.
Also, I haven't experienced any issue when building on F27 with both GNU Makefiles and Ninja.

Thank you very much for this great fix!

PR #3117, opened 2018-07-11
Unshadow rhel6 audit rules kernel module loading

yuumasato:
#### Description:

- Dropped rhel6 definition for `audit_rules_kernel_module_loading`
- Update shared definition to mimic rhel6 behavior 
- Parameterized path to binaries as they are different for rhel6.

#### Rationale:

- I wasn't sure if rhel6 should check for these too, so I just kept the same behavior.
- Also, rule description for `audit_rules_kernel_module_loading` doesn't mention `finit_module` nor `create_module`.


jan-cerny:
I have built SSG with your patches and I think the OVAL and XCCDFs are fine. Thank you for your contribution.

PR #3118, opened 2018-07-11
Unshadow rhel6 audit_rules_login_events

yuumasato:
#### Description:

- Drops rhel6 check `audit_rules_login_events` in favor of the shared one
- Regular expression on shared definitions are more flexible, they allow key to be specified as `-k` `-F key=`


yuumasato:
Thank you for the review.

PR #3121, opened 2018-07-11
Drop rhel6 audit_rules_networkconfig_modification

yuumasato:
#### Description:
- Dropped rhel6 check audit_rules_networkconfig_modification in favor of the shared one.
- Updated applicability of definition and extend definitions to include rhel6.

#### Rationale:
- Regular expression from shared allow key to be specified in two ways (`-k` and `-F key=`).
- And definition from shared also check rules defined using `augenrules`.


mpreisler:
The names of the files are different, so this is not a drop-in replacement. Hvae the rule metadata been updated in another PR?

cipherboy:
@mpreisler To my knowledge, I didn't touch any of those files.

yuumasato:
@mpreisler What files with different names are you talking about?

One thing I forgot to mention in description of PR is that shared definition `audit_rules_networkconfig_modification` extends `audit_rules_networkconfig_modification_domainname` and `audit_rules_networkconfig_modification_hostname`, while the rhel6 definition had it all within itself. 
So I have changed affected platforms of extended defs to `multi_platform_rhel`.

mpreisler:
@yuumasato Ah, ok, that was not clear to me.

PR #3122, opened 2018-07-11
Unshadow rhel6 audit_rules_session_events

yuumasato:
#### Description:
- Dropped rhel6 check `audit_rules_session_events` in favor of the shared one.
- Updated applicability of definition and extend definitions to include rhel6.

#### Rationale:
- Regular expression from shared allow key to be specified in two ways (`-k` and `-F key=`).
- And definition from shared also check rules defined using `augenrules`.


shawndwells:
Thanks for cleaning all this up @yuumasato! Ack once travis-ci passes.

PR #3126, opened 2018-07-12
Enable ol7 pci-dss shadow rsyslog rules

iokomin:
#### Description:
shadow and rsyslog files permissions and ownership rules added to ol7 pci-dss profile
- Added rsyslog_files_permissions, rsyslog_files_ownership, rsyslog_files_groupownership support on ol7
- Created shared file_owner_etc_shadow and file_groupowner_etc_shadow rules
Note: It looks reasonable for me to replace  userowner_shadow_file and groupowner_shadow_file with these taking for free generated remediation (and guess OVAL in the future) content from template. Unfortunately I don't have RHEL on hand to verify it works on this system as well.
- Fixed rsyslog_files_groupownership oval as expected group id mistakenly mixed up between rhel and debian/ubuntu systems
- Fixed rsyslog_files_permissions.sh which reset env PATH variable and as a result lost all env binaries available.

#### Rationale:

Add more ol7 pci-dss rules

#### Testing (Oracle Linux 7):
- Checked OVAL and remediation scripts content
- Checked guide content
- Checked rules evaluation an remediation scripts results


linux_os/guide/system/logging/ensure_rsyslog_log_file_configuration/group.yml
```diff
@@ -2,5 +2,11 @@ documentation_complete: true
...
+description: "The file <tt>/etc/rsyslog.conf</tt> controls where log message are written.\nThese are controlled by lines called <i>rules</i>, which consist of a\n<i>selector</i> and an <i>action</i>.\nThese rules are often customized depending on the role of the system, the\nrequirements of the environment, and whatever may enable\nthe administrator to most effectively make use of log data.
```
cipherboy: I'm looking at converting this from a single string (`description: "<some_text>"`) to a block quote-style string:

```
description: |-
  <some longer text with "\n" evaluated>
```

So the edited original would look like:

```
description: |-
    The file <tt>/etc/rsyslog.conf</tt> controls where log message are written.
    These are controlled by lines called <i>rules</i>, which consist of a
    <i>selector</i> and an <i>action</i>.
    These rules are often customized depending on the role of the system, the
    requirements of the environment, and whatever may enable
    the administrator to most effectively make use of log data.
    The default rules in Red Hat Enterprise Linux 7 are:
    <pre>*.info;mail.none;authpriv.none;cron.none                /var/log/messages
    authpriv.*                                              /var/log/secure
    mail.*                                                  -/var/log/maillog
    cron.*                                                  /var/log/cron
    *.emerg                                                 *
    uucp,news.crit                                          /var/log/spooler
    local7.*                                                /var/log/boot.log</pre>
    See the man page <tt>rsyslog.conf(5)</tt>    \ for more information.
    <i>Note that the <tt>rsyslog</tt> daemon can be configured to use a timestamp format that
    some log processing programs may not understand. If this occurs, 
    edit the file <tt>/etc/rsyslog.conf</tt> and add or edit the following line:</i>
    <pre>$ ActionFileDefaultTemplate RSYSLOG_TraditionalFileFormat</pre>
```

This would allow us to edit only the center text. Further, I'd propose using our jinja macro `{{{ full_name }}}`  instead of doing the conditional around product. This'll help me in the `debian8` as well; that way, we can condition on the product only for the contents of the defaults.

What do you think?

linux_os/guide/system/logging/ensure_rsyslog_log_file_configuration/rsyslog_files_groupownership.rule
```diff
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: rhel6,rhel7,fedora
+prodtype: rhel6,rhel7,fedora,ol7
```
cipherboy: I'm fine with dropping the prodtype here, since after the debian8 merger, it'd look like

`prodtype: debian8,ubuntu1404,ubuntu1604rhel6,rhel7,fedora,ol7`

and become way too messy.

linux_os/guide/system/logging/ensure_rsyslog_log_file_configuration/rsyslog_files_ownership.rule
```diff
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: rhel6,rhel7,fedora
+prodtype: rhel6,rhel7,fedora,ol7
```
cipherboy: (same as above)

linux_os/guide/system/logging/ensure_rsyslog_log_file_configuration/rsyslog_files_permissions.rule
```diff
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: rhel6,rhel7,fedora
+prodtype: rhel6,rhel7,fedora,ol7
```
cipherboy: (and here :)

linux_os/guide/system/logging/package_rsyslog_installed.rule
```diff
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: rhel6,rhel7,fedora
+prodtype: rhel6,rhel7,fedora,ol7
```
cipherboy: Yep.

linux_os/guide/system/permissions/files/permissions_important_account_files/groupowner_shadow_file.rule
```diff
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: rhel6,rhel7,fedora
+prodtype: rhel6,rhel7,fedora,ol7
```
cipherboy: I'd be fine dropping this prodtype.

linux_os/guide/system/permissions/files/permissions_important_account_files/userowner_shadow_file.rule
```diff
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: rhel6,rhel7,fedora
+prodtype: rhel6,rhel7,fedora,ol7
```
cipherboy: I'd be fine dropping this prodtype.

linux_os/guide/system/permissions/files/permissions_important_account_files/file_groupowner_etc_shadow.rule
```diff
@@ -0,0 +1,28 @@
+documentation_complete: true
```
cipherboy: @iokomin I'm not sure why you duplicated this file. I'd prefer if you instead rename `groupowner_shadow_file.rule` -> `file_groupowner_etc_shadow.rule` and replaced instances of `groupowner_shadow_file` with `file_groupowner_etc_shadow` and updated the corresponding oval checks to add `multi_platform_{rhel,fedora,ol,debian,ubuntu}` and drop the `prodtype` tag from the original.

linux_os/guide/system/permissions/files/permissions_important_account_files/file_owner_etc_shadow.rule
```diff
@@ -0,0 +1,31 @@
+documentation_complete: true
```
cipherboy: The same comment in the other file applies here, except renaming this one from `userowner_shadow_file.rule` -> `file_owner_etc_shadow.rule` and updating the corresponding fixes/ovals as above.

linux_os/guide/system/permissions/files/permissions_important_account_files/file_groupowner_etc_shadow.rule
cipherboy: Also make sure to update the corresponding fixes as well. :)

shared/checks/oval/file_groupowner_etc_shadow.xml
```diff
@@ -0,0 +1,28 @@
+<def-group>
```
cipherboy: Same as above re: rename.

shared/checks/oval/file_owner_etc_shadow.xml
```diff
@@ -0,0 +1,27 @@
+<def-group>
```
cipherboy: Same as above re: rename.

shared/fixes/bash/rsyslog_files_permissions.sh
```diff
@@ -1,4 +1,4 @@
-# platform = multi_platform_rhel
+# platform = multi_platform_rhel,Oracle Linux 7
```
cipherboy: You're kinda inconsistent on `multi_platform_ol` versus `Oracle Linux 7`; is there a particular reason why this needs to be OL7-only?

shared/fixes/bash/rsyslog_files_permissions.sh
```diff
@@ -35,18 +35,18 @@ do
 	fi
 done
 
-for PATH in "${LOG_FILE_PATHS[@]}"
+for LOG_FILE_PATH in "${LOG_FILE_PATHS[@]}"
```
cipherboy: (not an issue) Thank you very much! I like this fix. :)

linux_os/guide/system/logging/ensure_rsyslog_log_file_configuration/group.yml
iokomin: I totally agree with you, multi-line description is much easier to maintain including jinja macro usage. 
The reason I didn't use full_name here is that this is a group file which could be included in generated guide for other products and just replacing product name might confuse readers: assume reader noticing target platform in description but following set of RHEL7 related files is different (it should be updated as well). Thus I left existing RHEL7 as a default example. Let me know if I was wrong and just can use full_name instead :)

linux_os/guide/system/logging/ensure_rsyslog_log_file_configuration/rsyslog_files_groupownership.rule
iokomin: No problem with dropping this off if it is expected to work on all affected prodtypes.

linux_os/guide/system/permissions/files/permissions_important_account_files/file_owner_etc_shadow.rule
iokomin: Rules content are same in fact, I just copied userowner_shadow_file content as it works.

shared/fixes/bash/rsyslog_files_permissions.sh
iokomin: At the moment only OL7 support added. While some profiles in the future might be OL6 specific and due to some differences between releases like grub/grub2 multi_platform won't work here.
But in this case there is no reason to use specific platform name. I'll try to be more consistent in the future :)

linux_os/guide/system/permissions/files/permissions_important_account_files/file_owner_etc_shadow.rule
cipherboy: @iokomin Yes, that's my point... the content is the same, so I'd suggest using the same file :) No point duplicating the same content in multiple places; that's why we're merging everything back into a single guide. Since most everything (the other rules in that folder for instance, but also the stuff in the `debian8` guide as well) uses the `file_` prefix, I'd prefer if the original source that you copied from (`userowner_shadow_file.rule`) was renamed to the name you chose here. Then we wouldn't have two, and you get to keep the more consistent name. :)

linux_os/guide/system/permissions/files/permissions_important_account_files/file_groupowner_etc_shadow.rule
iokomin: Renaming is definitely preferable way. I duplicated these rules this time as I don't have other platforms on hand to verify these changes. This change also require touching profiles in other products. I decided to consult with you guys first, is it OK for me do changes like this or better SSG gurus will bring up everything in sync :)

shared/fixes/bash/rsyslog_files_permissions.sh
cipherboy: Ahh okie, then either way is fine by me. :)

linux_os/guide/system/permissions/files/permissions_important_account_files/file_groupowner_etc_shadow.rule
cipherboy: It's fine if you feel comfortable doing this, or else I can create a quick PR to do this and you can rebase on top of it when it merges. :)

If you have questions, we do have an IRC channel ([#openscap](https://www.open-scap.org/resources/support/#irc-freenode)) where a bunch of us lurk.

cipherboy: @iokomin Sure! I'm going to propose a WIP `debian8` one first to start getting feedback from @mpreisler and @redhatrises, but then I'll do that.

iokomin: @cipherboy I've moved this PR to WIP. Will be waiting for updates from you to rebase it and fix things to be updated.

cipherboy: @iokomin Thanks! I believe #3130 should do what we need :) Let me know if you have any thoughts on it.

iokomin:
Blocked by #3130

jan-cerny:
This is blocked by #3156

cipherboy:
@iokomin \o #3156 finally merged! Any chance you can update this and rebase? :)

iokomin:
@cipherboy, thanks for notification, going to complete this PR early next week.

cipherboy:
@iokomin Thanks! :)

shared/checks/oval/rsyslog_files_groupownership.xml
```diff
@@ -95,9 +96,9 @@
   <unix:file_state id="state_rsyslog_files_groupownership" version="1">
     <unix:type operation="equals">regular</unix:type>
     {{% if product in ["debian8", "ubuntu1404", "ubuntu1604"] %}}
-    <unix:group_id datatype="int">0</unix:group_id>
-    {{% else %}}
     <unix:group_id datatype="int">4</unix:group_id>
+    {{% else %}}
```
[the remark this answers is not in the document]
iokomin: Fix inspired by existing data failure on OL (expect same for RHEL):  group_id 4 checked while 0 expected. Code change analysis shown file state error introduced by 64ea161 commit, please have a look.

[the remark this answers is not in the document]
cipherboy: I verified that the owner is `adm` (id 4) on an Ubuntu machine. Good catch!

cipherboy:
I like these changes. ACK from me.

I lack an OL7 machine so I'll trust that these changes work for OL7 because they are proposed by @iokomin.

Thanks for your contributions! :)

PR #3127, opened 2018-07-12
Improve logic around uid_min and auid

yuumasato:
#### Description:

- Improve logic around `uid_min` and `auid`


cipherboy:
LGTM. :)

PR #3129, opened 2018-07-12
debian8 -> linux_os merge

cipherboy:
#### Ramblings

This is a merger of the `debian8` guide into the shared `linux_os` guide. After this, all Linux operating systems will use a single, shared guide. I haven't significantly tested this outside of build+tests. 

I wish there was a nicer way to handle packages/services having different names (other than duplicating `service_ntpd_enabled` -> `service_ntp_enabled`, etc.) but perhaps that's a minor thing in the grand scheme of things.

@redhatrises If you want this split into separate PRs for review purposes, I can do that. However, this is ~10x smaller than the `rhel6` merge, so it should be nicer to review anyways. :)

#### Rationale

This helps with maintenance costs but is a breaking change for anyone who carries patches upstream. By having a single `linux_os/guides` folder, we can now reorganize all of the rules, checks, and fixes into a single folder, greatly simplifying the project structure, and increasing ease of contributing. :)

Eventually `debian8/guides` will be removed in a separate PR.


cipherboy:
This build will intentionally fail until #3130 is merged, as the Linux OS guide is missing renames in this branch.

cipherboy:
(Removed @pep8speaks comment because .py file was unintentionally added to this PR and was later removed, comment was taking up too much space :)

cipherboy:
(Pushed updates that include both #3146 and #3149 so this will pass CI).

matejak:
Postponed to 0.1.41, as this PR is not essential for 0.1.40.

jan-cerny:
blocked by #3156

cipherboy:
(rebased now that #3156 has merged. Thanks @redhatrises !! )

ubuntu1404/profiles/standard.profile
```diff
@@ -1,10 +1,10 @@
 documentation_complete: true
 
-title: 'Standard System Security Profile for Ubuntu 14'
+title: 'Standard System Security Profile for Ubuntu 14.04'
```
redhatrises: @cipherboy since we are fixing this content now... remove the `.04`. We don't keep track of minor versions.... and neither do we want to.

ubuntu1604/profiles/standard.profile
```diff
@@ -1,10 +1,10 @@
 documentation_complete: true
 
-title: 'Standard System Security Profile for Ubuntu 16'
+title: 'Standard System Security Profile for Ubuntu 16.04'
```
redhatrises: @cipherboy since we are fixing this content now... remove the `.04`. We don't keep track of minor versions.... and neither do we want to.

ubuntu1404/profiles/standard.profile
redhatrises: Crap.... nm... fix it later

redhatrises:
Ack.

cipherboy:
Thanks for the merge, @redhatrises!

PR #3130, opened 2018-07-12
[WIP] Rename shadow file rules

cipherboy:
Per commentary on #3126, and for helping #3129. 

These two rules do not match the naming conventions in `linux_os`:
 - `groupowner_shadow_file.rule` -> `file_groupowner_etc_shadow.rule`
 - `userowner_shadow_file.rule` -> `file_owner_etc_shadow.rule`

This brings all of the linux-es into compliance except `wrlinux`, which still needs to be merged; there is no point keeping these names in `wrlinux` and they can be easily dropped and renamed when that guide is merged.

Since the corresponding fixes and ovals are clearly applicable across every platform, I have enabled them as such. Most of the changes should be fairly trivial.

This is required to merge both #3126 and #3129.


shared/fixes/bash/file_owner_etc_shadow.sh
```diff
@@ -0,0 +1,2 @@
+# platform = multi_platform_all
+chown root /etc/shadow
```
iokomin: Not required, shared/templates/create_file_owner.py should generate this fix for you from template

shared/fixes/bash/file_groupowner_etc_shadow.sh
```diff
@@ -0,0 +1,2 @@
+# platform = multi_platform_all
+chgrp root /etc/shadow
```
iokomin: Not required, shared/templates/create_file_groupowner.py should generate this fix for you from template

cipherboy:
@iokomin Thanks! Good catch; updated :)

jan-cerny:
@cipherboy I have checked the build output. In RHEL 6 and RHEL 7 XCCDFs, rules file_groupowner_etc_shadow and  file_owner_etc_shadow do not have OVAL, they have only OCIL. I think that's not expected.

cipherboy:
@jan-cerny Thanks for the comment. I'll fix that in a separate PR and then we can merge this one. Marking WIP...

jan-cerny:
Blocked by #3135

jan-cerny:
So I understand it well this is still blocked by allowing templated files to use Jinja?

PR #3132, opened 2018-07-13
Fix utilities for compraison, testoval

cipherboy:
These are broken since the refactor of the `ssg` module and since the rearrangement of utilities. Some of these changes are likely duplicated in #3014 / unnecessary with #3014, but that is fine as that is stalled for the time being.


cipherboy:

mpreisler:
Jenkins is dead but this passed on travis and locally on my machine. ACK

PR #3133, opened 2018-07-13
Reorganize ovals from debian8

cipherboy:
As part of moving `debian8`->`linux_os`, I wished to reclaim some of the debian8-specific ovals that are really cross-platform. 

This was started by noticing `sudo_require_authentication` was used by the Ubuntu guides, but they where not included. As this is cross-platform, `rhel`, `fedora`, `debian`, and `ubuntu` were added to the list.

Then I did a more thorough listing of ovals in debian8, and noticed that `systemmap` was really cross platform (since every Linux needs a `System.map*` file to boot and these permissions are sane), so this has been moved to shared and updated accordingly. Note that in the future, this should be updated to split the root checks from the file permissions check. 

Other platforms affected are `grub2_enable_iommu_force` multi_platform_all (since this should work on any GRUB2-platform (and perhaps in the future we can add an explicit guard to RHEL6 to prevent GRUB2 checks from appearing). This required an update in `rhel7/checks/oval/grub2_enable_fips_mode.xml` to remove the grub2 check. Perhaps a better structure for shared checks like `grub2_default_exists` should be done in the future.

`file_permissions_unauthorized_sgid.xml` was largely the same across `rhel6`, `rhel7`, and `OL7`, so I consolidated that one. `wrlinux` has its own still as that one only has 3 items. To check this, diff `rhel7` and `ol7` and ensure all of the items are the same. This forms the base of the `shared` OVAL. Then diff `rhel6` and `rhel7 ` and note that `jinja` guards in `shared` make the list the same for `rhel6` and `rhel7` depending on product. 

Then I removed `syslogng_files_permissions.xml` as it was unused in `debian8`.


rhel7/checks/oval/grub2_enable_fips_mode.xml
```diff
@@ -14,7 +14,6 @@
       <criteria operator="OR">
         <criterion test_ref="test_grub2_enable_fips_mode" comment="check for fips=1 in /etc/default/grub via GRUB_CMDLINE_LINUX" />
         <criteria operator="AND">
-          <criterion test_ref="test_grub2_default_exists" comment="check for GRUB_CMDLINE_LINUX_DEFAULT exists in /etc/default/grub" />
```
redhatrises: @cipherboy why was this removed?

cipherboy: Ah sorry. Meant to remove the or. It should only be an `AND`.

cipherboy:
@redhatrises This is updated. :) Should look better.

rhel7/checks/oval/grub2_enable_fips_mode.xml
```diff
@@ -12,9 +12,12 @@
       <extend_definition comment="prelink disabled" definition_ref="disable_prelink" />
       <extend_definition comment="package dracut-fips installed" definition_ref="package_dracut-fips_installed" />
       <criteria operator="OR">
-        <criterion test_ref="test_grub2_enable_fips_mode" comment="check for fips=1 in /etc/default/grub via GRUB_CMDLINE_LINUX" />
         <criteria operator="AND">
-          <criterion test_ref="test_grub2_default_exists" comment="check for GRUB_CMDLINE_LINUX_DEFAULT exists in /etc/default/grub" />
+          <extend_definition definition_ref="test_grub2_default_exists" comment="check for GRUB_CMDLINE_LINUX_DEFAULT exists in /etc/default/grub" />
+          <criterion test_ref="test_grub2_enable_fips_mode" comment="check for fips=1 in /etc/default/grub via GRUB_CMDLINE_LINUX" />
+        </criteria>
```
redhatrises: @cipherboy please remove the changes in lines 16-18.

mpreisler: why?

cipherboy: @redhatrises Sorry, misread that test. I thought it was `/etc/default/grub` exists. My bad!

cipherboy:
The `centos6` issues don't happen locally any more so I'm calling it a travis-ci glitch. Retrying...

cipherboy:
\o/ It passes!

rhel7/checks/oval/grub2_enable_fips_mode.xml
jan-cerny: @cipherboy I'm sorry, I don't understand. What should be removed? What shouldn't be removed? Why is this check removed? Are you sure you don't want to check existence of `GRUB_CMDLINE_LINUX_DEFAULT` in `/etc/default/grub`?

shared/checks/oval/file_permissions_unauthorized_sgid.xml
```diff
@@ -48,14 +55,16 @@
     <value>/usr/bin/wall</value>
     <value>/usr/bin/write</value>
     <value>/usr/lib/vte/gnome-pty-helper</value>
+    {{% if product != "rhel6" %}}
```
jan-cerny: I think it would be better to move these values inside `else` on line 42, and remove this `if`. That would reduce number of conditions and improve readability.

cipherboy:
@jan-cerny Thanks for the feedback! :) I had originally done that to improve review (that the file list didn't change with these updates), but I've condensed the logic since you're otherwise happy with it.

Updated.

jan-cerny:
@cipherboy And what about my comment about "check for GRUB_CMDLINE_LINUX_DEFAULT exists in /etc/default/grub" ?

rhel7/checks/oval/grub2_enable_fips_mode.xml
```diff
@@ -14,7 +14,7 @@
       <criteria operator="OR">
         <criterion test_ref="test_grub2_enable_fips_mode" comment="check for fips=1 in /etc/default/grub via GRUB_CMDLINE_LINUX" />
         <criteria operator="AND">
-          <criterion test_ref="test_grub2_default_exists" comment="check for GRUB_CMDLINE_LINUX_DEFAULT exists in /etc/default/grub" />
+          <extend_definition definition_ref="test_grub2_default_exists" comment="check for GRUB_CMDLINE_LINUX_DEFAULT exists in /etc/default/grub" />
```
jan-cerny: I can't find any definition with ID "test_grub2_default_exists" in ssg-rhel7-oval.xml.

cipherboy:
CI Failure looks like Travis is having problems again... https://travis-ci.org/OpenSCAP/scap-security-guide/jobs/406229475#L1224

cipherboy:
@jan-cerny Updated!

(again, `centos6` issue seems to be travis-related, and I cannot reproduce locally....)

shared/checks/oval/grub2_default_exists.xml
```diff
@@ -0,0 +1,27 @@
+<def-group>
+  <definition class="compliance" id="grub2_default_exists" version="1">
+    <metadata>
+      <title>Force IOMMU usage in GRUB2</title>
+      <affected family="unix">
+        <platform>multi_platform_all</platform>
+      </affected>
+      <description>Look for argument iommu=force in the kernel line in /etc/default/grub.</description>
+    </metadata>
```
jan-cerny: This OVAL definition does a different thing than described in title and description.

cipherboy:
@jan-cerny Ahh my bad. This is updated :)

jan-cerny:
@cipherboy Thank you very much. Kudos!

cipherboy:
Thanks for the merge @jan-cerny! :)

PR #3138, opened 2018-07-18
Avoid jinja eating the whitespace after the platform identifier in fixes

mpreisler:
I have caused 2 regressions while unshadowing the bash fixes. The platform line was being joined with the next line, the resulting bash fix was therefore not valid and even the platform comment couldn't be parsed. This fixes that. I also made the build script fatal error when the platform identifier is missing. This is the only thing that works because we can't always rely on a thorough review.


redhatrises:
Ack

PR #3140, opened 2018-07-19
Added example profile with tested rule that uses filehash58

matejak:
We need to test a rule that uses `filehash58` probe in the SSG suite.

I have added a rule that checks that RHEL7 `root`'s `.bashrc` hash matches a vanilla `.bashrc`. The rule has OVAL check and remediation, both can be tested by the SSG test suite.

### DIFF
--- a/rhel7/fixes/bash/root_bashrc_not_modified.sh
+++ b/rhel7/fixes/bash/root_bashrc_not_modified.sh
@@ -1,0 +1,4 @@
+# platform = Red Hat Enterprise Linux 7
+
+rhel7_bashrc_base64_encoded="IyAuYmFzaHJjCgojIFVzZXIgc3BlY2lmaWMgYWxpYXNlcyBhbmQgZnVuY3Rpb25zCgphbGlhcyBybT0ncm0gLWknCmFsaWFzIGNwPSdjcCAtaScKYWxpYXMgbXY9J212IC1pJwoKIyBTb3VyY2UgZ2xvYmFsIGRlZmluaXRpb25zCmlmIFsgLWYgL2V0Yy9iYXNocmMgXTsgdGhlbgoJLiAvZXRjL2Jhc2hyYwpmaQo="
+printf "%s" "$rhel7_bashrc_base64_encoded" | base64 -d > /root/.bashrc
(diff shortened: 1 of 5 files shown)


rhel7/fixes/bash/root_bashrc_not_modified.sh
```diff
@@ -0,0 +1,4 @@
+# platform = Red Hat Enterprise Linux 7
+
+rhel7_bashrc_base64_encoded="IyAuYmFzaHJjCgojIFVzZXIgc3BlY2lmaWMgYWxpYXNlcyBhbmQgZnVuY3Rpb25zCgphbGlhcyBybT0ncm0gLWknCmFsaWFzIGNwPSdjcCAtaScKYWxpYXMgbXY9J212IC1pJwoKIyBTb3VyY2UgZ2xvYmFsIGRlZmluaXRpb25zCmlmIFsgLWYgL2V0Yy9iYXNocmMgXTsgdGhlbgoJLiAvZXRjL2Jhc2hyYwpmaQo="
```
cipherboy: @matejak In general, this is fine and I understand the use of `base64` when wishing to handle binary data, but `/root/.bashrc` shouldn't contain arbitrary binary data, and more importantly, this one from `rhel7` doesn't. 

I'd personally prefer the form of:

```bash
cat << '_EOF' > /root/.bashrc
# .bashrc

# User specific aliases and functions

alias rm='rm -i'
alias cp='cp -i'
alias mv='mv -i'

# Source global definitions
if [ -f /etc/bashrc ]; then
	. /etc/bashrc
fi
_EOF
```

This way, it is clear to users what the blob is that they're writing to root's `~root/.bashrc`. :)

mpreisler:
NACK from me, let's not have strictly testing rules in SSG. This should be tested in OpenSCAP test suite.

redhatrises:
Agree with @mpreisler here. Why are we testing a probe here and not in the OpenSCAP test suite?

matejak:
There are some advantages of the SSG test suite (test data don't make mess, we test in an isolated environment where we can't break anything, SSG as a test framework is already set up), but as the primary goal is to test implementation of actual probes, it makes more sense to put those closer to the scanner. Therefore I close this PR and I will reuse its contents when extending the oscap's test suite.
@cipherboy You have good points, thanks for taking a look.

PR #3143, opened 2018-07-19
Remove FIPS remediations from CentOS as well as RHEL CCEs

redhatrises:
#### Description:

- Remove FIPS remediations from CentOS as well as RHEL CCEs

#### Rationale:

- CentOS is not FIPS certified, so it really shouldn't have remediation fixes
- RHEL CCEs do not translate to CentOS. CentOS should request their own CCEs

### DIFF
--- a/ssg/build_derivatives.py
+++ b/ssg/build_derivatives.py
@@ -94,3 +94,14 @@ def remove_idents(tree_root, namespace, prod="RHEL"):
             if ref.text is not None:
                 if re.search(ref_exp, ref.text):
                     rule.remove(ref)
+
+        for fix in rule.findall(".//{%s}fix" % (namespace)):
+            if "fips" in fix.get("id"):
+                rule.remove(fix)
+            sub_elems = fix.findall(".//{%s}sub" % (namespace))
+            for sub_elem in sub_elems:
+                sub_elem.tail = re.sub(r"[\s]+- CCE-.*", "", sub_elem.tail)
+                sub_elem.tail = re.sub(r"CCE-[0-9]*-[0-9]*", "", sub_elem.tail)
+            if fix.text is not None:
+                fix.text = re.sub(r"[\s]+- CCE-.*", "", fix.text)
+                fix.text = re.sub(r"CCE-[0-9]*-[0-9]*", "", fix.text)


shawndwells:
Thanks for doing this. Balancing quality content for the community vs accurately presenting security status of CentOS is hard.

There are deployments using CentOS with fips=1 and incorrectly assuming FIPS requirements are met (e.g. for FedRAMP, FISMA, government, or other regulated industries). That belief is reenforced when SSG content passes FIPS checks on CentOS, remediates findings, links to RHEL CCEs. 

One of the tradeoffs made for CentOS' use of FIPS-validated SSH ciphers was to create a "Use Strong Crypto for SSH" rule. CentOS content reflects SSH system cannot be configured to use FIPS ciphers (a rule which will always fail) however strong ciphers were implemented. Should something similar be created for CentOS enablement of FIPS?

In my opinion this would actually be inappropriate because a system is either using FIPS validated crypto or it isn't. But wanted to have the public discussion because this patch will make the life of CentOS system owners more challenging.

cipherboy:
@shawndwells What do you think of making `centos6`, etc.. full products? (This was suggested by @mpreisler). With the `rhel6` -> `linux_os` guide merge, the content is nearly free and no additional maintenance work, all of the product-specific references are fixed, and if the community wishes to maintain CCEs or get FIPS certification for CentOS, these can be added into the project and correctly scoped. Further, we're not tying these to be purely build-time derivatives of existing RHEL products and thus have more granularity and control over what content they expose.

From a personal perspective, I think CCEs / FIPS certification / etc. should be questions the community (of CentOS users) asks NIST/etc. about if they care about it, and it isn't up to our project to decide if CentOS is or is not FIPS certified, and whether or not they can use RHEL's CCEs. I'd also personally take a mixed stance as CentOS (and Fedora and perhaps other Linux distributions) share a lot of the same configuration, so CCEs scoped to Distros doesn't make sense (as opposed to CCEs scoped to projects which I personally feel is the correct scoping in the majority of cases as the file path may differ but rarely does the configuration values differ significantly)... ...but FIPS certification has rigorous testing requirements and thus we shouldn't enable FIPS-based rules for CentOS by default and unless they show an interest in becoming certified.

But that's my two cents.

redhatrises:
Since we are on this topic. CentOS does:

"CentOS Linux does not contain Red Hat Enterprise Linux or Fedora Linux; nor does it have any of their certifications, although it is built from the same source code as the Red Hat Enterprise Linux."
https://wiki.centos.org/FAQ/General#head-4b2dd1ea6dcc1243d6e3886dc3e5d1ebb252c194

"CentOS Linux is NOT Red Hat® Linux, it is NOT Fedora™ Linux. It is NOT Red Hat® Enterprise Linux. It is NOT RHEL. CentOS Linux does NOT contain Red Hat® Linux, Fedora™, or Red Hat® Enterprise Linux. CentOS Linux is NOT a clone of Red Hat® Enterprise Linux. CentOS Linux is built from publicly available source code provided by Red Hat®, Inc for Red Hat® Enterprise Linux in a completely different (CentOS Project maintained) build system."
https://wiki.centos.org/FAQ/General#head-c4a240cfc25e7fddd009d8c2ba18abfb87042feb

shawndwells:
On 7/19/18 1:48 PM, Alexander Scheel wrote:
>
> @shawndwells <https://github.com/shawndwells> What do you think of 
> making |centos6|, etc.. full products? (This was suggested by 
> @mpreisler <https://github.com/mpreisler>). With the |rhel6| -> 
> |linux_os| guide merge, the content is nearly free and no additional 
> maintenance work, all of the product-specific references are fixed, 
> and if the community wishes to maintain CCEs or get FIPS certification 
> for CentOS, these can be added into the project and correctly scoped. 
> Further, we're not tying these to be purely build-time derivatives of 
> existing RHEL products and thus have more granularity and control over 
> what content they expose.
>

I think that makes sense. Would be an opportunity to recognize CentOS as 
a first class citizen. Solves the problem of rule selections and profile 
names like you pointed out (e.g. how there is no CentOS STIG, so there 
shouldn't be a STIG profile in CentOS datastream).

> From a personal perspective, I think CCEs / FIPS certification / etc. 
> should be questions the community (of CentOS users) asks NIST/etc. 
> about if they care about it, and it isn't up to our project to decide 
> if CentOS is or is not FIPS certified, and whether or not they can use 
> RHEL's CCEs. I'd also personally take a mixed stance as CentOS (and 
> Fedora and perhaps other Linux distributions) share a lot of the same 
> configuration, so CCEs scoped to Distros doesn't make sense (as 
> opposed to CCEs scoped to projects which I personally feel is the 
> correct scoping in the majority of cases as the file path may differ 
> but rarely does the configuration values differ significantly)... 
> ...but FIPS certification has rigorous testing requirements and thus 
> we shouldn't enable FIPS-based rules for CentOS by default and unless 
> they show an interest in becoming certified.
>
> But that's my two cents.
>

Lack of CCE reuse is governed by NIST directly. SSG currently has the 
scaffolding to support per-product CCEs (e.g. ``rhel6@cce``, 
``rhel7@cce``) which gives CentOS community an easy process to start 
adding them (if they chose).

So, sounds like the general consensus is:
- Merge this patch
- Moving forward, make CentOS a full product
- Moving forward, revisit profile names/rule selections once CentOS is a 
full product

@cipherboy if that's correct, suggest you merge the PR as agreement. 
Otherwise we can continue the conversation.

redhatrises:
Need to figure out how to handle mixed elements before it can be merged.

shawndwells:
On 7/19/18 2:13 PM, redhatrises wrote:
>
> Need to figure out how to handle mixed elements before it can be merged.
>

Either updating this PR with mixed elements, or opening a ticket to 
track mixed elements and merging current PR, seems to makes sense. Good 
with either.

redhatrises:
In the meantime, this is ready. Personally, I say that the making centos and scientific linux as separate products are not a priority right now with all the other stuff that needs to be completed.

shawndwells:
LGTM. 

Agree on priorities too.

shawndwells:
On 7/19/18 5:37 PM, Martin Preisler wrote:
> Assigned #3143 
> <https://github.com/OpenSCAP/scap-security-guide/pull/3143> to 
> @shawndwells <https://github.com/shawndwells>.

There has to be a bot that could do cleanup like this....

PR #3146, opened 2018-07-20
Unify handling of file permissions checks

cipherboy:
#### Description

This combines the four separate file permissions templates into a single template. In particular, we had four templates:

- `shared/templates/create_file_groupowner.py` -- template for handling group ownership of files
- `shared/templates/create_file_owner.py` -- template for handling user ownership of files
- `shared/templates/create_file_permissions.py` -- template for handling unix mode permissions of files
- `shared/templates/create_permission.py` -- template for handling all (groupowner, owner, permissions) in one check

This renames `shared/templates/create_permission.py` -> `shared/templates/create_permissions.py` and updates it to create all four checks at once. This then removes a bunch of shadowed checks, renames inconsistencies in rules, and updates the corresponding profiles and overlays.

This updates all of the `<PRODUCT>/templates/csv/*.csv` to move content into `file_dir_permissions.csv`; this name is slightly misleading but was kept for historical reasons as it already contains the necessary structure. This also adds a preferred name column, and makes uid/gid/mode optional. 

#### Rationale

This blocks #3130, which in turn blocks #3126 and #3129. In particular, OVAL content for `file_owner_etc_shadow` and `file_groupowner_etc_shadow` was not getting generated. This fixes that, and likely obsoletes #3130, but that will need to be assessed after this is merged. 

(Rewrote commit message)


mpreisler:
This is a really nice change, thank you!

"banach-space" - I think I will start naming branches similarly ;-)

jan-cerny:
@yuumasato Which milestone do you recommend for this PR?

shared/templates/template_OVAL_permissions
```diff
@@ -1,33 +1,22 @@
 <def-group>
-  <definition class="compliance" id="file_permissions{{{ FILEID }}}" version="1">
+  <definition class="compliance" id="permissions{{{ FILEID }}}" version="1">
     <metadata>
-      <title>Verify {{{ FILEPATH }}} Permissions</title>
+      <title>Ensure Correct Mode, Owner, Gruop Owner for %FILEPATH</title>
```
jan-cerny: I think FILEPATH should be in triple curly brackets as well.

opensuse/templates/csv/file_dir_permissions.csv
```diff
@@ -0,0 +1 @@
+/etc,cron.allow,,,0644,cron_allow
```
jan-cerny: If I understand it well, in OpenSUSE benchmark, there used to be a check for `root` owner and `root` group, but you removed that and replaced that by checking `0644` mode? Why?

rhel6/checks/oval/file_group_owner_grub_conf.xml
```diff
@@ -1,37 +0,0 @@
-<def-group>
-  <definition class="compliance" id="file_group_owner_grub_conf" version="1">
-    <metadata>
-      <title>File grub.conf Owned By root Group </title>
-      <affected family="unix">
-        <platform>Red Hat Enterprise Linux 6</platform>
-      </affected>
-      <description>The grub.conf file should be owned by the root group. By default, this file is located at /boot/grub/grub.conf or, for EFI systems, at /boot/efi/EFI/redhat/grub.conf</description>
```
jan-cerny: I can't grep any mention of EFI in `ssg-rhel6-oval.xml`. Why was the EFI alternative removed?

rhel7/templates/csv/file_dir_permissions.csv
```diff
@@ -1,5 +1,11 @@
-/etc,gshadow,0,0,0000
 /etc,shadow,0,0,0000
-/etc,passwd,0,0,0000
-/etc,group,0,0,0000
-#/boot/grub,grub.conf,0,0,600 # different filename
+/etc,gshadow,0,0,0000
+/etc,passwd,0,0,0644
+/etc,group,0,0,0644
+/etc,cron.allow,,,0644,cron_allow
```
jan-cerny: Is it intended that it doesn't check for ownersip of `cron.allow` anymore?

shared/templates/template_OVAL_permissions
jan-cerny: Gruop

shared/templates/template_OVAL_file_permissions
```diff
@@ -0,0 +1,27 @@
...
+  <unix:file_test check="all" check_existence="all_exist" comment="Testing mode of {{{ FILEPATH }}}" id="test_file_permissions{{{ FILEID }}}" version="1">
+    <unix:object object_ref="object_file_permissions{{{ FILEID }}}" />
+    <unix:state state_ref="state_file_permissions{{{ FILEID }}}_mode_{{{ FILEMODE }}}" />
+  </unix:file_test>
+  <unix:file_state id="state_file_permissions{{{ FILEID }}}_mode_{{{ FILEMODE }}}" version="1">
+    {{{ STATEMODE }}}
+  </unix:file_state>
+  <unix:file_object comment="{{{ FILEPATH }}}" id="object_file_permissions{{{ FILEID }}}" version="1">
+    <unix:path>{{{ FILEDIR }}}</unix:path>
+    {{{ UNIX_FILENAME }}}
```
jan-cerny: I know it was already there before, but I find it confusing that the tag is not visible here, but it is included in variable. Also it's confusing that it is not done the same way for the FILEDIR variable. People might think there is a missing tag in this OVAL. Why not to do the same thing for `FILEDIR` as well?

shared/templates/create_permissions.py
```diff
@@ -0,0 +1,217 @@
...
+                "./oval/permissions{0}.xml", path_id
+            )
+
+        else:
+            raise UnknownTargetError(target)
+
+    def csv_format(self):
+        return("CSV should contains lines of the format: "
+               "directory path,file name,owner uid (numeric),group "
+               "owner gid (numeric),mode")
```
jan-cerny: Could we also document the 6th field?

jan-cerny:
@cipherboy 
From ssg-rhel6-ocil.xml:
```
6766     <ns0:boolean_question id="ocil:ssg-file_groupowner_etc_shadow_question:question:1">
6767       <ns0:question_text> To check the group ownership of /etc/shadow, run the command: $ ls -lL /etc/shadow If properly configured, the output should indicate the following group-owner. root
6768             Is it the case that ?
6769             </ns0:question_text>
6770     </ns0:boolean_question>
```
The line 6768 is weird. And it's there multiple times.

yuumasato:
> @yuumasato Which milestone do you recommend for this PR?

As it is renaming rules it would be great and easier to have it in 0.1.40 release.

shared/templates/template_OVAL_file_owner
```diff
@@ -0,0 +1,25 @@
+<def-group>
+  <definition class="compliance" id="file_owner{{{ FILEID }}}" version="1">
+    <metadata>
+      <title>Verify {{{ FILEPATH }}} Owner Permissions</title>
```
yuumasato: Drop Permissions from tittle, this OVAL is checking the file owner, not the file owner permissions.

shared/templates/create_permissions.py
```diff
@@ -0,0 +1,217 @@
...
+                    )
+                self.file_from_template(
+                    "./template_BASH_permissions",
+                    {
+                        "FILEPATH":      full_path,
+                        "FILEMODE":      mode,
+                        "FILEUID":       uid,
+                        "FILEGID":       gid,
+                    },
+                    "./bash/permissions{0}.sh", path_id
```
yuumasato: I believe this bash remediation should only be generated if `mode`, `uid` and `gid` are defined.

cipherboy:
@jan-cerny said:
> ...
> The line 6768 is weird. And it's there multiple times.

So I'm not entirely sure how this gets created.

In [`shared/macros.jinja`](https://github.com/openscap/scap-security-guide/blob/master/shared/macros.jinja#L480-L485):
```xml
{{% macro ocil_sysctl_option_value(sysctl, value) -%}}
    The status of the <code>{{{ sysctl }}}</code> kernel parameter can be queried
    by running the following command:
    <pre>$ sysctl {{{ sysctl }}}</pre>
    The output of the command should indicate a value of <code>{{{ value }}}</code>.
    If this value is not the default value, investigate how it could have been
    adjusted at runtime, and verify it is not set improperly in <tt>/etc/sysctl.conf</tt>.
{{%- endmacro %}}
```

But in [`shared/transforms/xccdf-create-ocil.xslt`](https://github.com/openscap/scap-security-guide/blob/master/shared/transforms/xccdf-create-ocil.xslt#L51-L62): 
```xslt
<boolean_question id="{@id}_question">
    <question_text>
    <xsl:apply-templates select="cdf:check[@system='ocil-transitional']/cdf:check-content"/>
    Is it the case that <xsl:value-of select="cdf:check[@system='ocil-transitional']/cdf:check-export/@export-name"/>?
    </question_text>
</boolean_question>
```

I'm not quite sure how these parts fit together, but to me it looks like the check isn't setting an `export-name`? I'm not sure how these get set, as they only seem to be referenced in `ssg/build_renumber.py` (on lines 85/86/201) and in various XSLTs in `shared/transforms`... Thoughts?

shared/templates/create_permissions.py
```diff
@@ -0,0 +1,217 @@
...
+
+            self.file_from_template(
+                "./template_ANSIBLE_permissions",
+                {
+                    "FILEPATH":      full_path,
+                    "FILEMODE":      mode,
+                    "FILEUID":       uid,
+                    "FILEGID":       gid,
+                },
+                "./ansible/permissions{0}.yml", path_id
```
yuumasato: Same as with `template_BASH_permission`, generate only if `mode`, uid,` and `gid` are defined.

shared/templates/create_permissions.py
```diff
@@ -0,0 +1,220 @@
...
+                    "FILEID":        path_id,
+                    "FILEPATH":      full_path,
+                    "FILEDIR":       dir_path,
+                    "FILEMODE":      mode,
+                    "FILEUID":       uid,
+                    "FILEGID":       gid,
+                    "UNIX_DIR":      unix_dir,
+                    "UNIX_FILENAME": unix_filename
+                },
+                "./oval/permissions{0}.xml", path_id
```
yuumasato: I couldn't find any rule named with prefix `permissions_`... Will this be picked up by build system?

shared/templates/create_permissions.py
cipherboy: I updated this by adding a line at the top of [`create_permissions.py`](https://github.com/cipherboy/scap-security-guide/blob/banach-space/shared/templates/create_permissions.py#L17) and adding gated checks in the jinja templates.

shared/templates/create_permissions.py
cipherboy: So... I'm not sure here either, but this was the original behavior. If we want rules that begin with `permissions_`, we should have the corresponding rules in the `linux_os` tree and they should be referenced as such in the correct profiles. As it stands, it looks like there are no such rules.

shared/templates/create_permissions.py
cipherboy: (same as above: see https://github.com/cipherboy/scap-security-guide/blob/banach-space/shared/templates/template_ANSIBLE_permissions)

cipherboy:
@jan-cerny said:
> @cipherboy some .rule files have `ocil_clause` value that gets propagated to the sentence `Is it the case that ...`.

In this case, `file_groupowner_etc_shadow` lacks such an `ocil_clause`, so perhaps it is best to add those in a separate PR. When `ocil_clause` is missing, perhaps too our XSLT macro should detect that.

cipherboy:
This warrants a little longer discussion I think while we are revisiting this. 

#### Current State

In current `master` (c803775fd3ae018eb0212353bba92d3f30894807), we have four individual templates. These are:

 - [`shared/templates/create_file_groupowner.py`](https://github.com/OpenSCAP/scap-security-guide/blob/c803775fd3ae018eb0212353bba92d3f30894807/shared/templates/create_file_groupowner.py) -- templates the check/remediations for the group owner of a file.
 - [`shared/templates/create_file_owner.py`](https://github.com/OpenSCAP/scap-security-guide/blob/c803775fd3ae018eb0212353bba92d3f30894807/shared/templates/create_file_owner.py) -- templates the check/remediations for the user owner of a file.
 - [`shared/templates/create_file_permissions.py`](https://github.com/OpenSCAP/scap-security-guide/blob/c803775fd3ae018eb0212353bba92d3f30894807/shared/templates/create_file_permissions.py) -- templates the check/remediations for unix mode permissions a file.

AND

 - [`shared/templates/create_permission.py`](https://github.com/OpenSCAP/scap-security-guide/blob/c803775fd3ae018eb0212353bba92d3f30894807/shared/templates/create_permission.py) -- templates the check/remediations for the group owner, user owner, and unix mode permissions a file.

Functionally, the first three do the same as the last one, except create three separate files as output artifacts (`file_permissions_*`, `file_owner_*`, and `file_groupowner_*`). HOWEVER, it is important to know that the last one (`create_permission.py`) creates the same output name (`file_permissions_*`) as `create_file_permissions.py`.... This means that (according to my tests), `create_permission.py` wins on the OVAL generation, in that it enforces uid/gid/mode. See `build/rhel7/checks/oval/file_permissions_etc_group.xml` for example.

Especially when `file_owner_*` and `file_groupowner_*` exist and are also templated, I believe this is the wrong choice because it is 1) misleading (the rules mention only unix mode but the oval checks validate uid/gid too), and 2) hard to control.

#### Proposed State

This PR allows for finer granularity in the structure. What it proposes is the following:

 - Remove the three separate templates
 - Update the functionality of the one template to generate four files: one for each check+remediation individually (uid/gid/mode) and one combined check+remediation. This then creates multiple files:

 - `file_permissions_*` - mode check+remediation
 - `file_owner_*` - uid check+remediation
 - `file_groupowner_*` - gid check+remediation
- `permissions_*` -- mode+uid+gid check, as applicable (so if one or two of those aren't specified in the template, they won't be checked).

Since this is one file doing all of this work (+jinja!!) it should be faster now but I don't have specific numbers there.

#### Discussion Points

The points of discussions are as follows:

- Are the `permissions_*` checks (that do everything) needed by anything?
    - As far as I can tell, these were unused due to the quirks of the build system: [`rhel7/..../file_dir_permissions.csv`](https://github.com/OpenSCAP/scap-security-guide/blob/master/rhel7/templates/csv/file_dir_permissions.csv) and [`rhel7/..../file_permissions.csv`](https://github.com/OpenSCAP/scap-security-guide/blob/master/rhel7/templates/csv/file_permissions.csv)
- If they are used, does this PR need to add them back in under their new name?
    - Further, because they had the same rule name, it is hard to know which profiles strictly require these changes, and thus creating new rule files with the new name is hard without domain knowledge.
- Do we need documentation inside the rules/template code to mention that they overlap?


I'm leaning towards adding the capabilities here, but not yet doing anything with them. Using them in the relevant profiles can be a separate PR, by someone more familiar with the specifics of the profiles.

jan-cerny:
@cipherboy The bug with incomplete "Is it the case that" in OCIL is reproducible in current master (as of head c803775fd3ae018eb0212353bba92d3f30894807). That means it is not related to this PR. We need to fix that, though. I have opened an issue where we can track that. See https://github.com/OpenSCAP/scap-security-guide/issues/3150 . Let's don't do it here.

yuumasato:
> Are the permissions_* checks (that do everything) needed by anything?
> 
>     As far as I can tell, these were unused due to the quirks of the build system: rhel7/..../file_dir_permissions.csv and rhel7/..../file_permissions.csv

@cipherboy I don't understand, there were no generated `permissions_` checks before, that checked for mode, uid and gid.

`file_dir_permissions.csv` was used to spawn `ansible`, `bash` and `oval` under prefix of `file_permissions`.
`file_permissions.csv` was used to spawn `ansible` and `bash` under prefix of `file_permissions`.
So, the `ansible` and `bash` generated were overwriting each other, but they were pretty much the same thing. Their goal was to check for permissions of a file, directory, or contents of a directory.
But the OVAL generated checked for mode, and for `uig` and `gid`. 
I have the impression that it was a workaround or some fix, because the templates don't generate OVALs for `file_owner` and `file_groupowner`.
And I think the confusion comes from this, because there is no rule that checs for all three,

shared/templates/template_ANSIBLE_file_permissions
```diff
@@ -3,21 +3,12 @@
...
-  tags:
-    @ANSIBLE_TAGS@
-
-- name: Set permissions
+- name: Ensure permission {{{ FILEMODE }}} on {{{ FILEPATH }}}
   file:
-    path: "{{ item.path }}"
-    mode: {{{ MODE }}}
+    path="{{item}}"
+    mode={{{ FILEMODE }}}
```
yuumasato: Does this syntax work? I'm used to colons.
Also, there are no references to equals here: https://docs.ansible.com/ansible/2.5/modules/file_module.html#examples

yuumasato:
> I'm leaning towards adding the capabilities here, but not yet doing anything with them. Using them in the relevant profiles can be a separate PR, by someone more familiar with the specifics of the profiles.

With the trend to use separate rules, instead of rules that check multiple things, there may be no interest in writing rules that check for `mode`, `uid` and `gid` together.

cipherboy:
@yuumasato said:
> @cipherboy I don't understand, there were no generated permissions_ checks before, that checked for mode, uid and gid.

I've updated my comment now with feedback. Perhaps it'll be a bit more clear now. @redhatrises seemed to be coming from the position that these were required for DISA STIG compliance, but I'm not sure they're actually used. I'm in favor of removing the new "do everything" (`permissions_*`) checks/remediations that I generate and updating the PR accordingly.

cipherboy:
Rebased after #3152 was merged.

yuumasato:
I am leaning towards merging this, and have the renamed rules in release 0.1.40.
And continue discussion and make needed changes later.

yuumasato:
@cipherboy I've found a bug, I'm sorry I didn't catch it on previous reviews.

The definitions checking for the following files are generated with wrong id (and filename):
- `cron_allow`
- `grub_conf`
- `httpd_server_conf_d_files`
- `shd_private_key`
- `shd_pub_key`

They are generated as `file_groupownercron_allow`, while the rule id is `file_groupowner_cron_allow`.
It seems to happens for all generators.

shared/templates/create_permissions.py
```diff
@@ -0,0 +1,224 @@
...
+        if uid == "" and gid == "" and mode == "":
+            # If UID, GID, and mode are not defined, skip processing this file
+            return
+
+        # build a string out of the path that is suitable for use in id tags
+        # example:	/etc/resolv.conf --> _etc_resolv_conf
+        # path_id maps to FILEID in the template
+
+        if len(path_info) == 6:
+            path_id = "_" + path_info[5]
```
cipherboy: @yuumasato Since all of the other IDs are path-based, they begin with a starting `/`; hence they'll be converted to a leading `_`, making the IDs look good and function as expected.

In user-supplied IDs, we didn't prefix them with an underscore, so now we do. :) Good catch, thanks!

yuumasato:
@cipherboy Thanks for updating quickly!
ACK.

PR #3147, opened 2018-07-23
Py2py3 sanity

mpreisler:
Build SSG with python2, then python3, in different build folders. Then compare results. This is using the compare_generated.sh script originally by @ybznek that I modified a bit.

This should help us catch any unexpected differences.

Since this is an integration test I haven't hooked it into cmake, I don't think that would make sense because we would be testing 2 different build folders with a context of yet another build folder.

### DIFF
--- a/tests/python2_vs_python3_sanity.sh
+++ b/tests/python2_vs_python3_sanity.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+set -euo pipefail
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+pushd "$DIR/.." > /dev/null
+
+rm -rf build-py2 build-py3
+
+mkdir -p build-py2
+pushd build-py2 > /dev/null
+rm -rf *
+cmake -G Ninja -DPYTHON_EXECUTABLE=/usr/bin/python2 ../
+ninja
+popd > /dev/null
+
+mkdir -p build-py3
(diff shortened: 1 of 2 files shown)


cipherboy:
These changes look good enough to me. I ran the new tool and nothing appears to have happened to make them different.

tests/python2_vs_python3_sanity.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+set -euo pipefail
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
```
cipherboy: I suppose the (most obvious) alternative to this lovely hack is calling out to a python interpreter, so I guess it can be excused. :)

utils/compare_generated.sh
```diff
@@ -1,17 +1,19 @@
 #!/bin/bash
-if [ "$#" -lt 2 ];
+
+if [ "$#" -lt 3 ];
```
cipherboy: I wouldn't mind a check for `3 <= $# <= 4`, but I'll let this slide. The more common case is "I don't know how to run it so here let me pass 0 arguments or a `-h` / `--help` to figure out how and this catches that, so I'm inclined to leave it. More defensive coding would validate that args 1 and 2 are directories that exist, args 3 and 4 are the correct strings if they exist, but I guess that's unnecessary since it is a manually run script that doesn't generate any build/etc. artifacts.

PR #3149, opened 2018-07-23
Remove prodtype from file permission and ownership checks

cipherboy:
Independent of #3146, though prefer merged after. 

This PR finishes the work of #3130 by removing `prodtype` from a bunch of rules which don't need them (`file_*.rule`), letting #3126 and #3129 continue. Since these are globally templated files, there is no `prodtype` that should be applied besides `all`. Thus, remove them.

### DIFF
--- a/utils/fix_file_ocilclause.py
+++ b/utils/fix_file_ocilclause.py
@@ -1,0 +1,258 @@
+import sys
+import os
+import argparse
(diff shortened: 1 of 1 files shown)


cipherboy:
Closing this as it is a duplicate of #3156 and is likely completely covered by it.

PR #3151, opened 2018-07-24
rm -rf debian8/guide

cipherboy:
~Blocked on #3129.~

After #3129 is merged, do a comparison of the guides pre-/post-merge and validate that the changes are sane. 

Then, remove the entire `debian8` guide because #3129 switches the guide over to `linux_os`. :)

~This includes a lot of duplicated commits and SHOULD NOT BE MERGED directly yet.~


cipherboy:
(reword commit message, dropped duplicated commits)

cipherboy:
Looks like the ovals are good and that there's a bunch of new fixes due to templating working in the new stuff. Thanks @redhatrises! :)

Here's the oval output: https://paste.fedoraproject.org/paste/3qBkYMmZ~viDwFbVyZy2OA
Here's the fixes output: https://paste.fedoraproject.org/paste/UrF7R3iBnXb7a6AiY9fptg

The interesting thing about the fixes is it points out that `content_rule_security_patches_up_to_date` has different severity...? 

https://github.com/OpenSCAP/scap-security-guide/blob/master/linux_os/guide/system/software/updating/security_patches_up_to_date.rule


... but I don't actually see that in the debian8 tree so I'm not entirely sure why that is. Other than that, it looks fine to my eye. 

Other than that, nothing is removed, so that is good :)

yuumasato:
With the move of `debian8` to shared guide,  the generated content changed a lot, not only for `debian8`.
A good change is that the number of rules and groups available for `debian` increased, but the`id`s of some groups and rules changed, and this will break existing customizations.
About other products, I noticed that `apt` rules are not limited to products based on `apt` packages, this should be changed, but not as part of this PR.

Now, regarding changes of this PR, all rules dropped from this guide have been moved to `linux_os` guide.

Ack, @cipherboy thank you for working on merging the guides. ;)

cipherboy:
Thanks @yuumasato! I'll look at introducing better prodtypes for package-manager related rules. :)

PR #3153, opened 2018-07-24
wrlinux -> linux_os guide merger

cipherboy:
#### Description

Most of the rules in the `wrlinux` guide are unused, and are at best copies of existing rules in `linux_os` or otherwise poorly modified. This is completely independent of the `debian8` merger (#3129), but I've not tested for cross-conflicts.

In essence, remove a bunch of prodtypes and then update the `product.yml` to point to `linux_os`. It should be simple to verify that all the existing rules are in `linux_os`, that the contents of differ only a little.

Easiest one yet!

#### Rationale

This reduces maintenance costs and unifies much product structure. This is the last remaining Linux product with a guide:

```
./wrlinux/guide
./firefox/guide
./linux_os/guide
./debian8/guide
./eap6/guide
./chromium/guide
./jre/guide
./fuse6/guide
```

So, since `debian8` is gone in #3151 and this one can be removed after a little more testing, we can now reorganize SSG content structure more easily and without having to do lots of duplicated actions.

(this one is much less epic as a PR, but is still good to have done... hence I'm not adding the EPIC label :)

### DIFF
--- a/linux_os/guide/system/accounts/accounts-restrictions/password_storage/accounts_password_all_shadowed.rule
+++ b/linux_os/guide/system/accounts/accounts-restrictions/password_storage/accounts_password_all_shadowed.rule
@@ -1,7 +1,5 @@
 documentation_complete: true
 
-prodtype: rhel6,rhel7,fedora,ol7
-
 title: 'Verify All Account Password Hashes are Shadowed'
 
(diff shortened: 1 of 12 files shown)


redhatrises:
Ack

PR #3154, opened 2018-07-24
rm -rf wrlinux/guide

cipherboy:
This is an intentional build failure. I want this to be merged after #3153 so that we have more time to validate and test the changes before removing everything if we desire.

Also, this would allow us to validate that we don't want to keep any of these files. I believe the answer is "no" though. 

Remove the unused `wrlinux/guide` folder after `wrlinux` is switched to the `linux_os` guide. :)


cipherboy:
(reword commit message)

mpreisler:
#3153 was merged so let's make this a blocker to avoid having clearly duplicated code in a release.

yuumasato:
I have checked the generated content for `wrlinux` and it looks pretty much the same.
There are some changes in rule descriptions, in general improving them, and added references to rules.

Generally `wrlinux` benchmark improved by moving to shared `linux_os` guide. Also, there are more rules included in the benchmark, improving options for customization, but it means that the benchmark file got bigger.

ACK, thanks for working on merging the guides.

PR #3157, opened 2018-07-24
Move signal logic from template to generator

### LINKED ISSUE #3155: RHEL 7 pwquality credits regex is missing the "dash"
#### Description of problem:
RHEL/CentOS 7  pwquality negative credits (ucredit,dcredit,ocredit) are always failing because the regex is not catching the "dash"
`^dcredit[\s]*=[\s]*(\d+)(?:[\s]|$)`      should be      `^dcredit[\s]*=[\s]*(-?\d+)(?:[\s]|$)`

Latest pull 7/24/2018, 
CentOS Linux release 7.5.1804 (Core)

#### Steps to Reproduce:

1. oscap xccdf eval --profile xccdf_org.ssgproject.content_profile_stig-rhel7-disa  --rule xccdf_org.ssgproject.content_rule_accounts_password_pam_dcredit --results results-dcredit.xml  ssg-centos7-ds.xml
2.  Replace ^dcredit[\s]*=[\s]*(\d+)(?:[\s]|$)   with   ^dcredit[\s]*=[\s]*(-?\d+)(?:[\s]|$) in centos7-ds.xml
3.oscap xccdf eval --profile xccdf_org.ssgproject.content_profile_stig-rhel7-disa  --rule xccdf_org.ssgproject.content_rule_accounts_password_pam_dcredit --results results-dcredit.xml  ssg-centos7-ds.xml

#### Actual Results:
cat /etc/security/pwquality.conf |grep dcredit
# Per CCE-CCE-27214-6: Set dcredit = -1 in /etc/security/pwquality.conf
dcredit = -1

Title   Set Password Strength Minimum Digit Characters
Rule    xccdf_org.ssgproject.content_rule_accounts_password_pam_dcredit
Result  fail

#### Expected Results:
Title   Set Password Strength Minimum Digit Characters
Rule    xccdf_org.ssgproject.content_rule_accounts_password_pam_dcredit
Result  pass

yuumasato:
#### Description:

- Moves logic of credit sign from template to generator
- Also fix a wrong template substitution

#### Rationale:

- Logic stopped working after we stopped using [custom string-replacement in templates ](https://github.com/OpenSCAP/scap-security-guide/pull/3135/files#diff-24dbdafa7989c5054be38a4c4bbf50f9L24)

- Fixes #3155 (thanks @mako7c4 for filing the issue)

### DIFF
--- a/shared/templates/create_accounts_password.py
+++ b/shared/templates/create_accounts_password.py
@@ -28,9 +28,12 @@ class AccountsPasswordGenerator(FilesGenerator):
                 "./ansible/accounts_password_pam_{0}.yml", VARIABLE
             )
         elif target == "oval":
+            # Only credit related variables allow negative values
+            SIGN = "-?" if VARIABLE.endswith("credit") else ""
             self.file_from_template(
                 "./template_OVAL_accounts_password",
                 {
+                    "SIGN": SIGN,
                     "VARIABLE": VARIABLE,
                     "OPERATION": OPERATION
                 },
--- a/shared/templates/template_OVAL_accounts_password
+++ b/shared/templates/template_OVAL_accounts_password
@@ -20,13 +20,6 @@
 {{% endif %}}
   </definition>
 
-{{# Only credit related variables allow negative values #}}
-{{% if "credit" in "{{{ VARIABLE }}}" %}}
-  {{% set sign = "-?" %}}
-{{% else %}}
-  {{% set sign = "" %}}
-{{% endif %}}
-
 {{% if product == "rhel6" %}}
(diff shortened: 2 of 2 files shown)


cipherboy:
Ahh missed that. Good catch @yuumasato! An alternative would be `{{% if "credit" in VARIABLE %}}` -- that wouldn't require pushing the logic into the generator if you don't want to.

cipherboy:
Ack pending CI from me. Thanks for the fix @yuumasato!

cipherboy:
Tested locally and verified that it does as advertised on the tin. Thanks!

yuumasato:
@cipherboy I tried your suggestion, but it wasn't working because of the other bug `negative_symbol`, which I only noticed after I moved the logic to the generator.
In the end I think it makes more sense to have it in the generator, as `SIGN` doesn't depend on any "environmental jinja variables".

Thanks for review and merging. :)

PR #3158, opened 2018-07-24
Fix template_OVAL_package_installed

cipherboy:
In #3135, a regression was introduced regarding variables inside
jinja macros. This fixes the other regression not fixed in #3157.


yuumasato:
Ack. This changes prevent OVAL for package installed from asserting for `evr_string` when `EVR` is empty.

PR #3159, opened 2018-07-24
Reduce build warnings by updating templates

cipherboy:
There are several warnings in the build due to templates being incorrect. This fixes the majority of those warnings. 

The warnings are of the form:

```
WARNING: OVAL definition 'service_nails_enabled' extends non-existing 'package_nails_installed', removing it from OVAL definitions.
WARNING: OVAL definition 'service_netconsole_disabled' extends non-existing 'package_netconsole_removed', removing it from OVAL definitions.
WARNING: OVAL definition 'service_ntpd_disabled' extends non-existing 'package_ntp_removed', removing it from OVAL definitions.
WARNING: OVAL definition 'service_sshd_enabled' extends non-existing 'package_openssh-server_installed', removing it from OVAL definitions.
```

(anything that begins with `WARNING: OVAL definition` is in scope for this PR).

This reduces warnings to about 1/5th of their previous quantity (~55 to < 10). 

It'd be great to get this into 0.1.40, to improve the release but not necessarily a requirement.


cipherboy:
This should be updated. Thanks for the feedback, @jan-cerny! :)

jan-cerny:
@cipherboy thanks

PR #3161, opened 2018-07-25
Add missing OCIL Clause in file permission rules

### LINKED ISSUE #3150: OCIL contains weird incomplete questions
#### Description of problem:

In OCIL, the sentence "Is it the case that" should continue by something after "that". It's cut instead.

See this example from ssg-rhel6-ocil.xml:
```
    <ns0:boolean_question id="ocil:ssg-file_permissions_etc_shadow_question:question:1">
      <ns0:question_text> To check the permissions of /etc/shadow, run the command: $ ls -l /etc/shadow If properly configured, the output should indicate the following permissions: ----------
              Is it the case that ?
              </ns0:question_text>
    </ns0:boolean_question>
```

It's there multiple times.

 current master (as of head c803775fd3ae018eb0212353bba92d3f30894807)

F27

#### Steps to Reproduce:

1. have a look to ssg-rhel6-ocil.xml
2.
3.
4.

#### Actual Results:
OCIL contains incomplete questions.

#### Expected Results:
OCIL doesn not contain incomplete questions.

some `.rule` files have `ocil_clause` value that gets propagated to the sentence `Is it the case that ....`

cipherboy:
Per report from @jan-cerny, a lot of the file permission rules lack OCIL clauses. This implements a new `jinja` macros to template the clause and adds them to a lot of missing rules in `linux_os/guide`.

This adds the python utility I wrote to automatically add these and then fixes #3150. Since this utility is based off a lot of common code and is not part of the build system, I'd prefer to just add it and review it later when I finish #3014.

### DIFF
--- a/utils/fix_file_ocilclause.py
+++ b/utils/fix_file_ocilclause.py
@@ -1,0 +1,248 @@
+import sys
+import os
+import argparse
+import subprocess
+import jinja2
+import yaml
+
+import ssg
+
+
+def _create_profile_cache(ssg_root):
+    profile_cache = {}
+
+    product_list = ['debian8', 'fedora', 'ol7', 'opensuse', 'rhel6', 'rhel7',
+                    'sle11', 'sle12', 'ubuntu1404', 'ubuntu1604', 'wrlinux']
+
+    for product in product_list:
+        found_obj_name = False
+        prod_profiles_dir = os.path.join(ssg_root, product, "profiles")
+        for _, _, files in os.walk(prod_profiles_dir):
+            for filename in files:
+                profile_path = os.path.join(prod_profiles_dir, filename)
+                parsed_profile = yaml.load(open(profile_path, 'r'))
+                for _obj in parsed_profile['selections']:
+                    obj = _obj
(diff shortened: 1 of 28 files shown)


utils/fix_file_ocilclause.py
```diff
@@ -0,0 +1,248 @@
...
+
+    for line_num in range(0, len(new_contents)):
+        line = new_contents[line_num]
+        if line == old_line:
+            new_contents[line_num] = key + ": " + new_value
+            updated = True
+            break
+
+    if not updated:
+        assert(False)
```
jan-cerny: It would be better to use `assert updated` instead.

cipherboy:
Agreed about assert usage. That code is unused in that script and is
needlessly duplicated between several utility scripts. I'll replace all
asserts with exceptions in the final version. All of the scripts are
re-executable in that they don't modify a file unless they detect they need
to, and then they fix it in a way that they won't edit it twice, so an
assert was a quick way to flag a file as needing manual review.

Thanks for the review!

PR #3163, opened 2018-07-26
Add bash remediation for rpm_verify_hashes

mildass:
#### Description:
Add bash remediation for Verify File Hashes with RPM rule.

### DIFF
--- a/shared/fixes/bash/rpm_verify_hashes.sh
+++ b/shared/fixes/bash/rpm_verify_hashes.sh
@@ -1,0 +1,8 @@
+# platform = multi_platform_rhel, multi_platform_fedora
+
+# Find which files have incorrect hash (not in /etc, because there are all system related config. files) and then get files names
+files_with_incorrect_hash="$(rpm -Va | grep -E '^..5.* /(bin|sbin|lib|lib64|usr)/' | sed -r 's;^.*\s(.+)$;\1;g' )"
+# From files names get package names and change newline to space, because rpm writes each package to new line
+packages_to_reinstall="$(rpm -qf $files_with_incorrect_hash | tr '\n' ' ')"
+
+{{{ pkg_manager }}} reinstall -y $packages_to_reinstall
--- a/tests/data/group_system/group_software/group_integrity/group_software-integrity/group_rpm_verification/rule_rpm_verify_hashes/fresh_system.pass.sh
+++ b/tests/data/group_system/group_software/group_integrity/group_software-integrity/group_rpm_verification/rule_rpm_verify_hashes/fresh_system.pass.sh
@@ -4,6 +4,6 @@
 
 # verify (-V) all installed (-a) packages digests,
 # if digest differs from package metadata, then attribute '5' is printed
-result=$(rpm -Va | grep '^..5')
+result=$(rpm -Va | grep -E '^..5.* /(bin|sbin|lib|lib64|usr)/')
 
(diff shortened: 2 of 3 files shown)


shared/fixes/bash/rpm_verify_hashes.sh
```diff
@@ -0,0 +1,12 @@
+# platform = multi_platform_rhel, multi_platform_fedora
+
+# Find which files have incorrect hash (not in /etc, because there are all system related config. files) and then get files names
+files_with_incorrect_hash="$(rpm -Va | grep -E '^..5.* /(bin|sbin|lib|lib64|usr)/' | sed -r 's;^.*\s(.+)$;\1;g' )"
```
redhatrises: This script should probably be similar to `shared/fixes/bash/rpm_verify_permissions.sh`

Also, we shouldn't limit to specific directories.

yuumasato: Would it make sense to update OVAL definition too?
https://github.com/OpenSCAP/scap-security-guide/blob/master/shared/checks/oval/rpm_verify_hashes.xml#L33

dahaic: Note that it was deliberate change, to limit the directories in the oval. Are we running in circles?

redhatrises: @dahaic no you are right. I am the one running in circles.

shared/fixes/bash/rpm_verify_hashes.sh
```diff
@@ -0,0 +1,12 @@
...
+# From files names get package names and change newline to space, because rpm writes each package to new line
+packages_to_reinstall="$(rpm -qf $files_with_incorrect_hash | tr '\n' ' ')"
+
+if which dnf ; then
```
yuumasato: We can use jinja macros and check for `pkg_manager` and avoid this `if` in final remediation snippet.

mildass:
@yuumasato Updated.
And I found out that one of tests checks hashes in all folders (but OVAL/remediation doesn't check/fix files in e.g. `/etc` folder). I changed that test to check only folders which are being checked/fixes and added as new commit to this PR.

tests/data/group_system/group_software/group_integrity/group_software-integrity/group_rpm_verification/rule_rpm_verify_hashes/fresh_system.pass.sh
```diff
@@ -4,6 +4,6 @@
 
 # verify (-V) all installed (-a) packages digests,
 # if digest differs from package metadata, then attribute '5' is printed
-result=$(rpm -Va | grep '^..5')
+result=$(rpm -Va | grep '^..5.* /(bin|sbin|lib|lib64|usr)/')
```
yuumasato: @mildass For me, this needs `-E` for grep to list the filenames.

mildass: @yuumasato Yes, sorry. I forgot it and I didn't think that's wrong, because test passed (but it passed because of empty grep output).

yuumasato:
Fail in Travis CI is due to linkchecker, unrelated to this PR.

PR #3164, opened 2018-07-26
Add bash remediation for set_password_hashing_algorithm_libuserconf rule.

mildass:
#### Description:

Add bash remediation script for Set Password Hashing Algorithm in /etc/libuser.conf rule.


mildass:
@dahaic `replace_or_append` appends line to the end of the file if it's missing. In this test we need to check line in section (`[defaults]`) and if it's not there then add it to section, not to the end of file. That's why I didn't use it.

mildass:
@dahaic Ok, I added them as a new commit.

dahaic:
Good stuff @mildass. Thanks!

PR #3165, opened 2018-07-26
Fix ansible for set_password_hashing_algorithm_libuserconf

mildass:
#### Description:
Fix regex in `insertafter`. If ansible doesn't find `crypt_style` line in file, then it inserts it after `.default]` (old regex) line, but according to `man libuser.conf`, that section is called `[defaults]` and there can be leading and trailing white spaces, so the old regex won't match correct cases (`[defaults]`, `  [defaults] ` etc.) and adds `crypt_style = sha512` to the end of file.

### DIFF
--- a/shared/fixes/ansible/set_password_hashing_algorithm_libuserconf.yml
+++ b/shared/fixes/ansible/set_password_hashing_algorithm_libuserconf.yml
@@ -6,7 +6,7 @@
 - name: Set Password Hashing Algorithm in /etc/libuser.conf
   lineinfile:
     dest: /etc/libuser.conf
-    insertafter: "^.default]"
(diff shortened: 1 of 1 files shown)


jan-cerny:
This is a nice catch @mildass . Thank you. Have you verified that the `]` doesn't have to be escaped?

mildass:
@jan-cerny Yes, I tried it with `ansible-playbook` and it worked. So it should be ok.

jan-cerny:
@mildass That's awesome.

dahaic:
@jan-cerny @mildass btw, we have this beautiful test suite, which would, having the test scenarios added, satisfy all the curiosity I see in the comments.

PR #3166, opened 2018-07-26
Fix linkchecker - disable ssl verification

cipherboy:
Add linkcheckerrc to set sslverify=0

This prevents us from validating the certificates of the links we check.
Because https://pcisecuritystandards.org does not serve the intermediate
certificate chain, we won't pass linkchecker tests with sslverify=1 (the
default). On the other hand, most browsers will cache intermediate CA
certs, letting the user access this website. Maintaining our own CA chain
just to add the GoDaddy intermediate cert we need is both fragile and
too much work for the little value it provides, so this is a better
compromise.

Also adds a docker image to run just the linkchecker test and adds it to travis-ci. Since we're waiting on other container builds anyways, we might as well run it here too. An alternative would be to add linkchecker to an existing run, but this way, hopefully it'll be done in parallel versus in series. :)

Also adds `--output-on-failure` to all Docker images to give useful output on test failures.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,3 +27,8 @@ matrix:
             - docker build --no-cache --tag ssg_$BASE_IMAGE:latest -f Dockerfiles/$BASE_IMAGE .
           script:
             - docker run ssg_$BASE_IMAGE:latest
+        - env: BASE_IMAGE="fedora_linkchecker"
+          before_install:
+            - docker build --no-cache --tag ssg_$BASE_IMAGE:latest -f Dockerfiles/$BASE_IMAGE .
+          script:
+            - docker run ssg_$BASE_IMAGE:latest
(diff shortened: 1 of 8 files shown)


jan-cerny:
@cipherboy The update is in stable https://bodhi.fedoraproject.org/updates/FEDORA-2018-9c3620944d

cipherboy:
@jan-cerny Thanks! :) This is updated and the CentOS CI is rerunning because it failed in the update process again. >:(

cipherboy:
@jan-cerny  \o/ CentOS passed this time :)

matejak:
Thank you for this PR, it looks great. I would like the Dockerfiles RUN sections to be formatted according to [best practices](https://github.com/projectatomic/container-best-practices/blob/master/creating/creating_index.adoc#chaining-commands), but that would be scope creep.

PR #3169, opened 2018-07-27
Refactored the compare_generated.sh script.

matejak:
* Added CLI using Argbash.
* Made it possible to operate on directories as well as files.
* Made it independent from directory it is executed in.
* Added some error handling.


mpreisler:
I wish the script was in python instead so that it'd run on Windows but this is a huge step forward regardless.

PR #3170, opened 2018-07-27
Jinja cleanup + tests

matejak:
* Removed no longer used _rename_items
* Made `_extract_substitutions_dict_from_template` a public function with docstring.
* Added tests for using external symbols in jinja macros.

### DIFF
--- a/ssg/jinja.py
+++ b/ssg/jinja.py
@@ -68,7 +68,12 @@ def _get_jinja_environment(substitutions_dict):
 _get_jinja_environment.env = None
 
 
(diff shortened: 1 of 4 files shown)


mpreisler:
scrutinizer fails to run the tests, it could be a conflict between ssg and tests.ssg, we should really choose a different name for tests.ssg (which I thought we agreed upon in another review here on github), it will just continue being a problem in the future.

matejak:
It was just a missing `jinja2` package on the scrutinizer system. I have restarted the inspection and it passed, but it is not reflected here on github.

jan-cerny:
@matejak Thanks for clarification.

PR #3171, opened 2018-07-31
Reformatted chained commands in Dockerfiles.

matejak:
Reformatted RUN directivesso they comply with best practices:
https://github.com/projectatomic/container-best-practices/blob/master/creating/creating_index.adoc#chaining-commands


Dockerfiles/centos6
```diff
@@ -20,4 +22,7 @@ RUN rm -rf $OSCAP_DIR/build/*
 
 WORKDIR /home/$OSCAP_USERNAME/$OSCAP_DIR/build
 
-CMD cmake -DPYTHON_EXECUTABLE=/usr/bin/python2 .. && make -j $BUILD_JOBS && ctest --output-on-failure -j $BUILD_JOBS
+CMD true \
+	&& cmake -DPYTHON_EXECUTABLE=/usr/bin/python2 .. \
+	&& make -j $BUILD_JOBS \
+	&& ctest --output-on-failure -j $BUILD_JOBS
```
cipherboy: The style guide isn't explicit about whether or not this should exactly mirror the `RUN` command. I'd be inclined to add the `&& true` here as well, because if any previous command exists with non-zero, `&& true` won't clobber the return code (and won't even get executed), and if everything includes zero return codes, then `&& true` will as well. Might as well keep it consistent everywhere. Same thing in the later Dockerfiles.


Thoughts?

matejak: Sure, I like your view. I have done that in the rebase. It is not common to have chained commands in `CMD`, but formatting it like this makes it very easy to see what's going on, and diffs also become more useful.

PR #3172, opened 2018-07-31
SSG test suite cleanup

matejak:
This PR introduces SSG TS cleanups that I have performed in course of preparations to extend the its functionality.

`oscap.py`:
* Removed unused import.
* There are now `run_cmd_local`, `run_cmd_remote`, and `_run_cmd` that is used by both.

`rule.py`
* Removed unused imports and usage of obsolete module `string`.
* Stopped specifying commands as strings and splitting them later, now they are specified as tuples from the get go.

`test_suite.py`
* Extracted argument normalization (or resolution if you will) and determination of the logging directory into functions.

### DIFF
--- a/tests/ssg_test_suite/rule.py
+++ b/tests/ssg_test_suite/rule.py
@@ -6,8 +6,6 @@ import logging
 import os
 import os.path
 import re
-import shlex
-import string
 import subprocess
 import sys
 
@@ -34,7 +32,7 @@ def _parse_parameters(script):
                               re.MULTILINE)
             if found is None:
                 continue
-            params[parameter] = string.split(found.group(1), ', ')
+            params[parameter] = found.group(1).split(', ')
     return params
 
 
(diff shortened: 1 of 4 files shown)


tests/ssg_test_suite/rule.py
```diff
@@ -68,24 +66,24 @@ def _send_scripts(domain_ip):
     logging.debug("Uploading scripts.")
     log_file_name = os.path.join(LogHelper.LOG_DIR, "data.upload.log")
 
-    command = "ssh {0} mkdir -p {1}".format(machine, remote_dir)
+    command = ("ssh", machine, "mkdir", "-p", remote_dir)
```
jan-cerny: Could the `run_cmd_remote` be used here instead?

tests/ssg_test_suite/rule.py
```diff
@@ -68,24 +66,24 @@ def _send_scripts(domain_ip):
...
                                   stdout=log_file,
                                   stderr=subprocess.STDOUT)
         except subprocess.CalledProcessError as e:
             logging.error("Cannot create directory {0}.".format(remote_dir))
             return False
 
-        command = "scp {0} {1}:{2}".format(archive_file, machine, remote_dir)
-        subprocess.check_call(shlex.split(command),
+        command = ("scp", archive_file, "{0}:{1}".format(machine, remote_dir))
+        subprocess.check_call(command,
```
jan-cerny: Could this be inside a `try .., except`?

tests/ssg_test_suite/rule.py
matejak: Unfortunatelly, `run_cmd_remote` has hardwired logging configuration, so it is not a drop-in replacement for this construct.

tests/ssg_test_suite/rule.py
matejak: Sure, and as those calls are very similar, I have made a small function that ensures that they are performed in the same way.

jan-cerny:
Thank you.

PR #3176, opened 2018-08-02
Fix grub2_nousb_argument: .rule.rule extension

cipherboy:
Well, apparently I did a really silly thing and named the rule `grub2_nousb_argument.rule.rule`. Needless to say this causes a few problems when moving rules...

As far as I can tell, these two are the only references of the `grub2_nousb_argument` (filename, and the fedora profile) identifier. 

:+1: Good job, me!


redhatrises:
hahahaha.... happens to everyone at some point. Ack.

PR #3177, opened 2018-08-02
Use a defaultdict for config

cipherboy:
Signed-off-by: Alexander Scheel <ascheel@redhat.com>

We duplicate the functionality of a `defaultdict` and make it harder to add a configuration parameter in the future. Simplifying this will also help with a rule reorganization. :)

PR #3156, opened 2018-07-24
Simplify prodtype again

cipherboy:
Simplify the `prodtype`s again!

Create a utility to track `prodtype` usage vs. actual value and update the rules/vars accordingly. Like #3161, I'd prefer to review the utility later as part of #3014.

PR #3178, opened 2018-08-02
Reorganize rules

cipherboy:
# Feedback is actively requested

#### State of the PR

~This PR is currently WIP as we're missing a few things:~

This PR is up to date and ready to go! :D

 - ~PR #3156~ and ~#3129~ and ~#3126~ -- because this PR will change directory structure significantly enough to not bother merging this before those and this will cause lots of conflicts in PRs. -- Merged!
- ~PR #3172~ -- because I need to add support for the new structure in the SSG Test Suite. Merged, thanks @jan-cerny and @matejak for the changes! :)
- ~PR #3179~ -- because these changes conflict and I'd like to test as much on Docker as possible. -- per comments from @matejak, my changes should be independent.
- ~PR #3177~ -- because I refactored that file and opened a PR and I want to use those changes here when I refactor that file even more.  - Merged, thanks @jan-cerny! 
- ~PR #3184~ -- because this requires tests to be in their correct places before they'll be automatically moved. Currently I'm carrying those commits in this PR as well.  - Merged, thanks @jan-cerny!
- ~PR #3188~ -- because that includes the build system changes to move rules around. Those changes are duplicated here as well. -- Updated recently and now merged, thanks for the feedback and the merge @matejak! 
- ~PR #3198~ -- because that includes a useful function that this code depends on. Those changes are duplicated here as well.  -- Merged, thanks for the feedback @matejak!

What is done:
 - There exists a utility to move stuff around
 - Everything (~but `debian8`...~) is moved.
 - It builds!

I need to:
 - ~Update the unit tests and utilities as the `ssg.oval` unit test fails because it doesn't understand the new structure~ -- Done!
 - ~Update the SSG Test Suite to understand the new location for rules.~ -- I have decided to postpone moving the contents of `tests/data` to a separate PR so this one can be reviewed as those changes are technically independent of these.

#### Description:

This is the start of reorganizing all of the rules. This changes rules to be a directory rather than a file. An example would look like [this](https://github.com/cipherboy/scap-security-guide/tree/reorganize-rules/linux_os/guide/system/accounts/accounts-banners/banner_etc_issue):

```
./banner_etc_issue/rule.yml
./banner_etc_issue/bash/rhel6.sh
./banner_etc_issue/bash/rhel7.sh
./banner_etc_issue/oval/shared.xml
./banner_etc_issue/tests/banner_etc_issue_disa_dod_short.pass.sh
./banner_etc_issue/tests/banner_etc_issue_disa_double_banner.fail.sh
./banner_etc_issue/tests/banner_etc_issue_disa_usgcb_banner.fail.sh
./banner_etc_issue/tests/banner_etc_issue_disa_dod_default_banner.pass.sh
./banner_etc_issue/tests/banner_etc_issue_ospp_usbcg_banner.pass.sh
```

Each rule directory:
1. The directory name **must** be the rule id
2. The directory **must** contain a file named `rule.yml`
3. The directory **may** contain folders in the set `{oval, bash, ansible, anaconda, puppet}`
3.1. Inside each of these, there **may** be files named `shared.ext` and `{{{ product }}}.ext` -- with `{{{ product }}}.ext` being preferred over `shared.ext` when it exists. (`.ext` is not a literal "`.ext`" but is the correct extension for that file format -- `.xml`, `.sh`, `.yml`, `.anaconda`, `.pp` respectively). 
4. The directory **may** contain a folder named `tests`, which contains SSG Test Suite data according to earlier standards. 

Note that this PR _adds_ support for rule directories and does not remove support for rule files. If, after this PR is accepted, we wish to remove support for rule files, we can.

#### Commits: 

1. ~7010d9a -- pulled from #3177~ - rebased and dropped. 
2. Add an automated utility for moving rules around.
3. Move everything (separate commit to make rebasing easier).
4. Fix the build system so that everything builds. This includes a refactor of `build-scripts/combine_remediations.py`.

#### Rationale:

This has a many benefits: 

0. A directory means we can group everything related to the rule in one filesystem location.
1. `rule.yml` -- because it ends in `.yml` now instead of `.rule`, it is properly syntax highlighted. Otherwise, unchanged between the current state of a rule
2. Folders named `{oval, bash, ansible, anaconda, puppet}` give us better visibility into shadowing across produts and other related issues. 
3. [WIP] Folder named `tests` to contain all of the SSG Test Suite testing data related to that rule.
    - Postponed to a later PR.
4. A better view of what is duplicated across products without having a shared reference. In the example, we can clearly see that the `bash` fixes are duplicated across `rhel6` and `rhel7`, and a diff shows that only a few differences occur, and should probably be merged at some point in the future: 

```
< # platform = Red Hat Enterprise Linux 6
---
> # platform = Red Hat Enterprise Linux 7
6c6,7
< expanded=$(echo "$login_banner_text" | sed 's/\[\\s\\n\][+*]/ /g;s/\\//g;s/[^-]- /\n\n-/g')
---
> # When there are multiple banners in login_banner_text, the first banner should be the one for RHEL7
> expanded=$(echo "$login_banner_text" | sed 's/(\\\\\x27)\*/\\\x27/g;s/(\\\x27)\*//g;s/\^(\(.*\)|.*$/\1/g;s/\[\\s\\n\][+*]/ /g;s/\\//g;s/[^-]- /\n\n-/g;s/(n)\**//g')
```
5. Contributing new content or improving existing content becomes **significantly** easier as everything is in one location.
6. It is easier to tell what rules have checks/remediations and which rules do not. 
7. Eventually we might want to completely remove product-specific files and remove the directories, however, that is a later step involving lots of manual work merging different files with jinja and we can have a single, flat directory for each rule, with one subdirectory for tests.
8. New utilities can be built for understanding the content better. See #3193 for my current progress.

The downside is that all external patches to this will break, horribly (if they haven't already with the guide merges).

-----

Documentation for these changes are carried in #3201.


cipherboy:
(Updated due to merge conflicts from #3156).

matejak:
I like the idea and the initial approach.
I see one issue that may not be entirely trivial - some rule tests use shared functionality, so they source a test micro-library from directories somewhere above, and use sourced functions in an easy way. For example, the mount checks in the `group_partitions` directory work that way (but there are many more). The reorg will inevitably break their current property that tests in the source tree are the same as tests on the tested setup, so e.g. shellchecking them won't be possible and writing them will become slightly less straightforward.
However, as long as test micro-libraries remain supported, I don't see that limitation as significant.

cipherboy:
Thanks for the feedback @matejak!

I think what I'll introduce is then a "tests" folder in a non-rule-directory (i.e., group) as well to add the library functionality. Then all parent directory's "tests" directories will be included for a rule.

My build script already ignores non-rule directories (well, it recurses into them, but explicitly conditions on there being a `rule.yml` file before acting on a directory being a rule. 

This would have the added benefit of making rule directories and group directories the same: a group directory is a folder with `group.yml` and can contain a `tests` library directory, and a rule directory is a folder with `rule.yml` and can contain a  `tests` directory that tests the functionality of the rules.

@dahaic Are you okay with pulling the tests into the content directory so closely?

dahaic:
Yeah, it makes a lot of sense. What I have been trying to achieve with the way tests are grouped in the data directory is ability to run them on the development machine right away. Do we want to retain that? In that case, I suggest to put only `./tests` subdirectories into the tarball going to tested machine. It would be slightly annoying to have to source with minimum of `../../tests/<somelib>`. But the usecase would work.

Another option is to prepare tarballs in such a way there would be no `tests` subdirectories, and it would look the very same way it looks now. That would make it not runnable directly, but would make sourcing slightly nicer. (And the tarball creation would be a tad bit more complex).

Preferences @cipherboy @yuumasato @matusmarhefka ?

cipherboy:
I like the former (keeping `tests` in the tarball) and actually I'm thinking I'm partial to shipping all of the guide content. For one, it'd make things easier to debug (I'm failing on this test, I can quickly (in the VM) open the OVAL and figure out why it is failing). For two, the total contents of `linux_os/guide` on my branch are 18MB. That's a rounding error in today's disk sizes, even for a VM. You're going to need 2-3 orders of magnitude more disk space for OS install than you are for our content. 

Then you get the benefits of both not having to make a complicated tarball and being able to run scripts directly. 

Thoughts?

yuumasato:
> It would be slightly annoying to have to source with minimum of ../../tests/<somelib>.

It would be annoying but it is explicit. Writing a test and sourcing a file as if it existed in local directory while it actually is some directories above is confusing.

> I'm partial to shipping all of the guide content.

Do we need whole guide content to debug a rule?

> For one, it'd make things easier to debug (I'm failing on this test, I can quickly (in the VM) open the OVAL and figure out why it is failing).

I like the idea of being able to debug using SSGTestSuite. There could be a flag to enable debugging, if result of test scenario is not as expected, pause execution and allow tester to ssh into VM and check things.
In this case, it may be useful to also send the oval snippet of the rule along with `testoval.py`.

cipherboy:
I'm removing the `[WIP]` flag because this is ready to be merged pending the dependent PRs getting merged. Please comment on those first and then I'll update this one. :)

cipherboy:
I have updated these a bit today. In particular:

- I have fixed a bug in the reorganization where `puppet` and `anaconda` output was getting swapped and called the other thing. This has been fixed.
- I have fixed a bug where certain products which used `linux_os` guide weren't getting moved there. This is fixed and updated and is why #3198 is now included here.

Both of these two issues were found with the tooling in #3193. :)

cipherboy:
@matejak (I seem to recall you've worked on the Scrutinizer stuff / have access to the VM):

This failed with the following error

```python
ssg/rules.py:7: in <module>
    from .build_remediations import REMEDIATION_MAP, is_applicable_for_product
ssg/build_remediations.py:13: in <module>
    from .products import parse_name, multi_list, map_name
ssg/products.py:6: in <module>
    from .yaml import open_raw
ssg/yaml.py:5: in <module>
    import yaml
E   ModuleNotFoundError: No module named 'yaml'

----------- coverage: platform linux, python 3.6.3-final-0 -----------
Coverage XML written to file coverage.xml

!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!
=========================== 1 error in 0.76 seconds ============================
```

I'm guessing you need to add `PyYAML` as a dependency somehow? ...

matejak:
@cipherboy I have fixed it, but it will flip to green when you push again. BTW you are also an admin of SSG on Scrutinizer.

cipherboy:
@matejak Ahhh I see the page now. Thanks! I'll keep that in mind for future reference. :)

cipherboy:
@matejak I've updated this PR with the commits from #3188. :)

cipherboy:
#3188 was merged (thanks @matejak) so this is ready for review! :)

PR #3180, opened 2018-08-08
Add remediation and tests for mount_option_krb_sec_emote_filesystems rule

mildass:
#### Description:
Add bash and ansible remediation with tests for Mount Remote Filesystems with Kerberos Security rule.

### DIFF
--- a/shared/fixes/ansible/mount_option_krb_sec_remote_filesystems.yml
+++ b/shared/fixes/ansible/mount_option_krb_sec_remote_filesystems.yml
@@ -1,0 +1,19 @@
+# platform = multi_platform_all
+# reboot = false
+# strategy = configure
+# complexity = low
+# disruption = medium
+
+- name: "Get nfs and nfs4 mount points, that don't have Kerberos security option"
+  shell: grep -E "[[:space:]]nfs[4]?[[:space:]]" /etc/fstab | grep -v "sec=krb5:krb5i:krb5p" | awk '{print $2}'
+  register: points_register
+  check_mode: no
+  tags:
+    @ANSIBLE_TAGS@
+
+- name: "Add Kerberos security to mount points"
+  shell: awk '$2=="{{ item }}"{$4=$4",sec=krb5:krb5i:krb5p"}1' /etc/fstab > fstab.tmp && mv fstab.tmp /etc/fstab
+  with_items:
+    - "{{ points_register.stdout_lines }}"
+  tags:
+    @ANSIBLE_TAGS@
--- a/tests/data/group_services/group_nfs_and_rpc/group_nfs_configuring_clients/group_mounting_remote_filesystems/fstab
+++ b/tests/data/group_services/group_nfs_and_rpc/group_nfs_configuring_clients/group_mounting_remote_filesystems/fstab
@@ -1,0 +1,8 @@
+# /etc/fstab: static file system information
+# Example fstab file with mounted NFS file systems
+# <file system>                   <mount point>   <type>  <options>                             <dump>  <pass>
+
+/tmp/folder0                        /folder0       ext4   defaults,sec=krb5:krb5:krb5i:krb5p              0  0
+some.server:/remote/folder1         /folder1       nfs    defaults,noexec,nosuid,sec=krb5:krb5i:krb5p     0  0
+/home/folder2                       /folder2       auto   defaults                                        0  0
+another.server:/remote/folder3      /folder3       nfs4   defaults,noexec,nosuid,sec=krb5:krb5i:krb5p     0  0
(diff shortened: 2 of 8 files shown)


shared/fixes/ansible/mount_option_krb_sec_remote_filesystems.yml
```diff
@@ -0,0 +1,20 @@
+# platform = multi_platform_rhel, multi_platform_fedora
```
yuumasato: It would be great if this module could do the changes via Ansible module `mount`. This [template](https://github.com/OpenSCAP/scap-security-guide/blob/master/shared/templates/template_ANSIBLE_mount_option) for mount options can serve as inspiration

tests/data/group_services/group_nfs_and_rpc/group_nfs_configuring_clients/group_mounting_remote_filesystems/fstab
```diff
@@ -0,0 +1,6 @@
+# /etc/fstab: static file system information
+# Example fstab file with mounted NFS file systems
+# <file system>                   <mount point>   <type>  <options>                             <dump>  <pass>
+
+some.server:/remote/folder1         /folder1       nfs    defaults,noexec,nosuid,sec=krb5:krb5i:krb5p     0  0
+another.server:/remote/folder2      /folder2       nfs4   defaults,noexec,nosuid,sec=krb5:krb5i:krb5p     0  0
```
yuumasato: There could be other unrelated mounts here besides the `nfs` ones. Or it could another test scenario, not sure if it is needed to have two scenarios for pass scenario in which all mounts of `nfs` type have the kerberos options.

yuumasato:
Discussing with @mildass, it looks sensible to have remediation that will not mount the remote filesystem, only configure `/etc/fstab`.

To verify that remediated `fstab` can mount its remote filesystem, the test scenario would have to setup a server for NFS, and I consider this out of scope.

@redhatrises @dahaic Do you consider necessary to have a note in Rule description saying that remediation  will not mount the remote filesystem?

mildass:
We discussed with @yuumasato the solution I made at ansible remediation - the `shell: grep ...` and `shell: awk ...`, that it would be better though ansible module `mount`. Yes, it would but:
- module `mount` usually uses `mount` command to get mounted devices and then it changes their configuration in `/etc/fstab`. It doesn't see devices, which are configured in `/etc/fstab` but aren't mounted in the moment of remediation
- with module `mount` and its `state` parameter set to `present` (specifies that the device is to be configured in fstab and does not trigger or require a mount) it is possible to achieve what we want. But it needs to know device source, path, fstype, options, dump and passno and it creates new record in `/etc/fstab` (= for this we would need 9 ansible tasks - grep nfs from fstab, 6x cut column, delete old nfs records and then `mount` task with `present` state)

Solution with `mount` won't be that clear and readable in my opinion. That's why I used just grep/awk in ansible.

shared/fixes/ansible/mount_option_krb_sec_remote_filesystems.yml
```diff
@@ -0,0 +1,20 @@
...
+# reboot = false
+# strategy = unknown
```
yuumasato: I think strategy = configure is more appropriate

shared/fixes/ansible/mount_option_krb_sec_remote_filesystems.yml
```diff
@@ -0,0 +1,20 @@
...
+# complexity = low
+# disruption = medium
+
+- name: "Get nfs and nfs4 mount points, that don't have Kerberos security option"
+  shell: grep "[[:space:]]nfs[4]*[[:space:]]" /etc/fstab | grep -v "sec=krb5:krb5i:krb5p" | awk '{print $2}'
```
yuumasato: Just like in bash regex, use `?` instead of `*` to specify quantifier for `[4]`.

tests/data/group_services/group_nfs_and_rpc/group_nfs_configuring_clients/group_mounting_remote_filesystems/rule_mount_option_krb_sec_remote_filesystems/missing_krb_sec.fail.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+# profiles = xccdf_org.ssgproject.content_profile_stig-rhel7-disa
+
+cp ../fstab /etc/
+# delete sec=.. only from nfs4
+sed -i 's|\(.*nfs[4].*\),sec=krb5:krb5i:krb5p\(.*\)|\1\2|' /etc/fstab
```
yuumasato: there is no need to put `4` within square brackets.

shared/fixes/ansible/mount_option_krb_sec_remote_filesystems.yml
yuumasato: Is there a reason for this to apply only for rhel and fedora? The bash remediation applies to all platforms.

mildass:
@yuumasato Updated. Hope it's ok now

PR #3183, opened 2018-08-09
Ignore .pytest_cache

cipherboy:

On pytest failure, this directory is left around. Ignore it.


mpreisler:
The travis failure is the infamous "bytecode marshal error", nothing to do with this PR.

PR #3184, opened 2018-08-10
Fix ssg test suite rules

cipherboy:
While working on the SSG test suite changes for #3178, I noticed that a few rules are in the wrong location, due to previous guide mergers and other changes (I was not able to find the source for the `auditd_data_retention` -> `configure_auditd_data_retention` move, but it exists in 4dc1e1fbc3babe47456e76b726637b1067d0f215).

I also validated that all of the rules in the `auditd_data_retention` group are actually in the `linux_os/guide` path for `configure_auditd_data_retention`, so that is good. 

This fixes their locations in the current SSG test suite.


cipherboy:
I restarted the Fedora Python2 worker because $TRAVIS (failed to connect to Docker to pull images).

PR #3185, opened 2018-08-10
Fix testoval.py: namespaces, type errors, variable shadowing

cipherboy:
If you want, I can minimize the diff, but I thought I'd try making the script more readable. 

With the `ssg` module refactor + later PRs, I broke this script. However, I believe it to be broken before too: the `definitions`/etc. elements were not namespaced, and hence were not OVAL XML elements. This fixes that such that they're namespaced. Also, with python2/python3 compatibility, we need to convert the output the command (`cmd_out`) from a `bytes`-like object in python3. Lastly, `find_testfile_or_exit` shadowed its input variable, so the file path printed in the error was always "None".

(The scope of this PR is just a fix + minor style changes. I want to reorganize this module and refactor it later).

### DIFF
--- a/ssg/oval.py
+++ b/ssg/oval.py
@@ -18,11 +18,12 @@ SHARED_OVAL = re.sub(r'ssg/.*', 'shared', __file__) + '/checks/oval/'
 
 
 # globals, to make recursion easier in case we encounter extend_definition
-definitions = ET.Element("definitions")
-tests = ET.Element("tests")
-objects = ET.Element("objects")
-states = ET.Element("states")
-variables = ET.Element("variables")
+ET.register_namespace("oval", ovalns)
+definitions = ET.Element("oval:definitions")
+tests = ET.Element("oval:tests")
+objects = ET.Element("oval:objects")
+states = ET.Element("oval:states")
+variables = ET.Element("oval:variables")
 silent_mode = False
 
 
@@ -108,14 +109,13 @@ def replace_external_vars(tree):
 
 def find_testfile_or_exit(testfile):
     """Find OVAL files in CWD or shared/oval and calls sys.exit if the file is not found"""
-    testfile = find_testfile(testfile)
-    if testfile is None:
-        print(
-            "ERROR: {0} does not exist! Please specify a valid OVAL file."
-            .format(testfile), file=sys.stderr)
+    _testfile = find_testfile(testfile)
+    if _testfile is None:
+        print("ERROR: %s does not exist! Please specify a valid OVAL file." % testfile,
+              file=sys.stderr)
         sys.exit(1)
     else:
-        return testfile
+        return _testfile
 
 
 def find_testfile(testfile):
@@ -187,6 +187,7 @@ def main():
     header = oval_generated_header("testoval.py", oval_version, "0.0.1")
     testfile = find_testfile_or_exit(testfile)
     body = read_ovaldefgroup_file(testfile)
+
     defname = _add_elements(body, header)
     if defname is None:
         print("Error while evaluating oval: defname not set; missing "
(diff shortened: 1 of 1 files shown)


ssg/oval.py
```diff
@@ -18,11 +18,11 @@
 
 
 # globals, to make recursion easier in case we encounter extend_definition
-definitions = ET.Element("definitions")
-tests = ET.Element("tests")
-objects = ET.Element("objects")
-states = ET.Element("states")
-variables = ET.Element("variables")
+definitions = ET.Element("ns0:definitions")
```
cipherboy: I think the more correct thing to do is register "oval" as a namespace and use it that way:

https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.register_namespace

Thoughts as to if this is preferred?

cipherboy:
Ah yeah. This is run by first sourcing `/.pyenv.sh` and then the usual way. 

So:

```bash
source ../some/arbitrary/relative/or/absolute/path/to/scap-security-guide/.pyenv.sh
[python{2,3}] /some/arbitrary/relative/or/absolute/path/to/scap-security-guide/utils/testoval.py <args>
```

That is, `.pyenv.sh` is "smart" in that it can be called from anywhere and will set the PYTHONPATH appropriately. Then you call `testoval.py` in the regular way.

redhatrises:
This maybe out of scope of this PR, but https://github.com/OpenSCAP/scap-security-guide/pull/3185#issuecomment-412167307 should be added to `developer_guide`.

ssg/oval.py
jan-cerny: Yes, I think that the current solution could crash if at some other place some other namespace got assigned the same prefix. I'm not an expert, but from what I read in the docs, using `register_namespace` should be correct.

ssg/oval.py
```diff
@@ -197,38 +196,49 @@ def main():
...
+
     # re-map all the element ids from meaningful names to meaningless
     # numbers
     testtranslator = IDTranslator("scap-security-guide.testing")
     ovaltree = testtranslator.translate(ovaltree)
     (ovalfile, fname) = tempfile.mkstemp(prefix=defname, suffix=".xml")
     os.write(ovalfile, ET.tostring(ovaltree))
     os.close(ovalfile)
+
+    cmd = "oscap oval eval --results " + fname + "-results " + fname
```
jan-cerny: Shouldn't it be a list instead? See https://docs.python.org/3/library/subprocess.html#subprocess.Popen

jan-cerny: I think if it was a list, we could get rid of 'shell=True' below, which has "Security considerations".  https://docs.python.org/3/library/subprocess.html#security-considerations

cipherboy: Sure, I'll add this in this PR. It is a `str` because that's what it was [currently in master](https://github.com/OpenSCAP/scap-security-guide/blob/master/ssg/oval.py#L213) (I moved it up to add the non-error `cmd` statement). I think I'll also simplify the `print(..)` on new-line number 228.

cipherboy:
Okay, this should be a little cleaner. :) Thoughts @jan-cerny?

cipherboy:
@redhatrises said:
> This maybe out of scope of this PR, but #3185 [(comment)](https://github.com/OpenSCAP/scap-security-guide/pull/3185#issuecomment-412167307) should be added to developer_guide.

This was added in #3186 which was merged by @mpreisler.

cipherboy:
`hostname: 3072aed0-051f-4172-beaf-42a69c84f452@1.production-4-worker-org-a-1-gce` -- Travis failure unrelated. Restarting job...

mpreisler:
ACK

PR #3179, opened 2018-08-03
SSG test suite support for Docker.

matejak:
This PR adds support for scanning Docker containers (as opposed to VM instances), together with some refactoring.

* Added Python3 compatibility.

* Got rid of `run_profile` and `run_rule` boilerplate functions. Introduced the `Checker` class that allows some code sharing between profile and rule checks.
* Relocated test environments-related code to the `test_env` submodule. This allows to use the snapshot-like functionality in form of context managers and to clearly separate libvirt-related code from the docker-related code.
* Implemented Docker test environment support, added support for it to the CLI and expanded the README accordingly.
* Introduced one test suite launcher for each backend - now there are `test_suite-docker.py` and `test_suite-vm.py`.

Known issues that are intentionally not part of this PR:
- The snapshot handling got a bit stupid - it can't do reverts without deleting the snapshot. This is a performance regression and can be fixed later.
- Terminating the program during snapshot creation/deletion may not delete them. It used to work somewhat better (but not perfectly) by using `atexit`, but that would be too hackish for both VMs and Docker images/containers at the same time. It can be fixed later.
- Plenty of tests don't work in containers (mounts, dbus, ...).

PR #3187, opened 2018-08-10
SSG Module: add setup.py for installation

cipherboy:
Per @redhatrises's [comment](https://github.com/OpenSCAP/scap-security-guide/pull/2949#discussion_r195463466) on #2949, this creates the `ssg` Python module by adding a `setup.py` file.

The suggested usage is to avoid the necessity of `.pyenv.sh` usage, and allow `ssg` to be found more easily on non-*nix systems. Further, we add a call to contributions for others to contribute useful features.

Questions:
 - Can anyone verify that `license='BSD'` does refer to the BSD-3 clause? I was not able to verify this immediately.
 - Are the `author` and `author_email` ok?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,6 @@ CMakeLists.txt.user
 
 # Ignore pytest cache
 .pytest_cache/
+
+# Ignore Python egg for ssg when installing with pip
+ssg.egg-info/
--- a/docs/manual/developer_guide.adoc
+++ b/docs/manual/developer_guide.adoc
@@ -519,6 +519,25 @@ An alternative is adding the directory where `testoval.py` resides to your PATH.
 $ export PATH=$PATH:/home/_username_/scap-security-guide/utils/
 ----
 
+== The SSG Python Module
+
+Locaated in the `ssg` directory are several useful Python modules for working with
+the content in SSG. Some of these utilities are of broader interest to the broader
+community working with OVAL content as well. However, the modules with prefix `build_`
+are only useful for building SSG. Contributions to this module are actively welcomed.
+
+By installing this in the user context from an active repository, as shown below, you
+can use the Python utilities in the `utils` directory without setting `PYTHONPATH`.
+
+To install the SSG Python module in a user scope:
+----
+$ pip install --user -e ./
+----
+
+Note that this module is not published on PyPI and is not obtainable from there.
+Further, we don't recommend installing this module in the system scope as it includes
+many utilities meant for development. Use of a virtualenv would be suggested instead.
+
 == Contributing with XCCDFs, OVALs and remediations
 
 There are three main types of content in SSG, they are rules, defined using the XCCDF standard, checks, usually written in link:https://oval.mitre.org/language/about/[OVAL] format, and remediations, that can be executed on ansible, bash, anaconda installer and puppet.
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,12 @@
+from setuptools import setup
+
+setup(name='ssg',
+      version='0.1',
+      description='SCAP Security Guide build and utilities module for generating compliance content',
+      url='http://github.com/OpenSCAP/scap-security-guide',
+      author='The OpenSCAP Team',
+      author_email='open-scap-list@redhat.com',
+      license='BSD',
+      packages=['ssg'],
+      install_requires=['jinja2', 'PyYAML'],
+      zip_safe=False)


mpreisler:
Are there real use cases of this?

I am not sold on this - which means I am **against** it.

cipherboy:
TL;DR: Now? Perhaps not. Later? We should definitely think about it.


Perhaps it is premature.
                  
What I'd like to see is the `ssg/` module include more of utils and expose them via a module calling interface (`__main__.py`). The scripts in `utils` can then largely go away and developer workflow is as follows:

 - Source `.pyenv.sh` or otherwise (via `setup.py`) put `ssg` in the module scope.
    - We still need a `.pyenv.bat` for Windows.
    - Perhaps these should live under `utils` as well.
 - Expose only thin wrappers around this in `utils/` as bash scripts / python modules compatibility reasons.
 - Have the preferred / documented way be to call these utilities from their module locations.
     - Some python modules ship separate wrappers in addition to their `module.submodule` interface, which I'd like to do here.

I'd like to see a better story around contributing to SSG. What I'm missing is: 
 - How can I fix existing content? -- This is largely being addressed by #3178 because it'll be a lot easier to find stuff, and allow those who interact with our content to specifically comment on what needs to be fixed.
 - How do I report an issue? Part of the previous question, but also, needs to add support for our generated reports and tying together reports <-> new rule locations.
 - How can I write content? -- This can partially be addressed by #3178 and partly is already addressed in the developer documentation. Perhaps also "model" rules which have _everything_ in them (tests, remediations, ovals, etc.)
 - How can I be sure my changes are correct? There are two parts to this:
    - How can I be sure my generated content is correct? We're lacking tools here; some of the utilities are fragile / don't show much in the way of interesting things (which I'd like to fix before I leave if I have time). But as far as introspecting the `build/ssg-*.xml` files _without_ already knowing being a subject matter expert on OVAL/XCCDF/.... etc, that is hard and _should_ be addressed.
    - How can I be sure my checks and remediations are correct? That's what the SSG Test Suite does and #3179 makes easier. But here too, having quicker ways to test this locally like `testoval.py` working again and perhaps other tooling would be nice. 
        - One concrete example would be a tool for generating the "build output artifacts" for a collection of rules. Generate a testing profile, generate all of the remediations and data streams and XCCDFs and etc., but only for those rules, for a specific product. 
        - Another concrete example of how this could be improved is a local testing option, and better support for non-CentOS/RHEL-like OS in the test suite directly. I think that'd be largely driven by community support. Perhaps our SUSE/Oracle contributors would like to add support for testing their platforms. Further, how can we test our non-Linux platforms? We have Chromium, JRE, and other products in our repository and outside of Linux platforms.
 - How can I extend SSG to be useful to `<my use case>`? This is largely what the SSG module refactors (#2949) were trying to achieve, and I'm not yet entirely sure what else I'd want to see here in terms of tooling. But I'd argue this is a case especially where adding tools to interface with SSG without requiring all of the strictly-standards driven stuff (OpenSCAP in particular, OVAL/XCCDF/DS generated XML content) and allowing people to interact with the input files (ymls/etc.) in a post- #3178 SSG would be very helpful for driving community involvement and newer use cases. Especially in a Compliance As Code mentality.
 - How can I view coverage metrics of SSG so we know where to contribute to next? This is something that gets vastly easier with #3178 that I'd like to improve as well before leaving if I have time. 
 - Do we care about building templates on each build, or can we check-in our templated artifacts? This is a change I'd like to experiment with after #3178 because it'd save on build times: rather than having to dynamically generate dependencies, we can check in artifacts to our build tree. When we change the templates, yeah, everything would change as a result, but it'd be a lot easier to validate these changes as part of the PR review process. Also, it'd enable non-OSCAP use cases of these templated artifacts as well and reduce the number of "special cases" we'd have to maintain.

And for other things to think about from a community perspective:
 - Is there an alternative to OVAL as the source language? Yeah, it is a useful, standard-mandated output language that we've built stuff around, but it is also a barrier to entry for a lot of people. Writing more templates, generators, anything (like we did for groups and rules and such by converting them to YAML) would be helpful for driving adoption here, even if it isn't necessarily the default across all of the build system.
- Can we add better safety mechanisms into our remediations. Recoverability. Assertions that we need to fix something before assuming that we have to. Testing of different edge cases as part of SSG perhaps. These are important in a outside-of-SSG / non-OSCAP use case that we could encourage more.

A lot of these non-OSCAP use cases are important because by enabling them, if we get meaningful users, they'd also validate, test, and hopefully improve the quality of our with-OSCAP use cases.

And it is those non-OSCAP use cases that'd benefit from having an SSG module to interact with our content, even if it was driven by specific rule ids.

Thoughts?

cipherboy:
I agree with concerns 1 and 2, but I had hoped to propose ways around that. I'm operating under these two assumptions:

1. The preferred way to install Python modules, from a Python point of view, in nearly any capacity is with `pip`, with a fallback via `$SYSTEM_PKG_MANAGER`. 
2. To use `pip` to install a module, you need a `setup.py` script.

If we can do 1 without 2, I'd be really happy. Perhaps we can parse the command line or something and see if `setup.py` is being run directly, but that gets really complicated and hacky.

To solve concerns 1 and 2, my proposed docs change says this:

```adoc
By installing this in the user context from an active repository, as shown below, you
can use the Python utilities in the `utils` directory without setting `PYTHONPATH`.
To install the SSG Python module in a user scope:

----
$ pip install --user -e ./
----

Note that this module is not published on PyPI and is not obtainable from there.
Further, we don't recommend installing this module in the system scope as it includes
many utilities meant for development.
```

What `pip install --user -e ./` accomplishes is syncing `module <-> repo` state: `--user` installs in the user scope, but `-e` sets up an editable/development [installation](https://pip.pypa.io/en/stable/reference/pip_install/#cmdoption-e) that is roughly equivalent to `setup.py develop`. In particular, the installation looks like this:

```
cipherboy:~$ python2
Python 2.7.15 (default, May 16 2018, 17:50:09) 
[GCC 8.1.1 20180502 (Red Hat 8.1.1-1)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import ssg
>>> ssg
<module 'ssg' from '/home/cipherboy/GitHub/c2/scap-security-guide/ssg/__init__.pyc'>
>>> 
```

and the only "installed" file is `/home/cipherboy/.local/lib/python2.7/site-packages/ssg.egg-link`:
```
/home/cipherboy/GitHub/c2/scap-security-guide
.
```


If you then update `ssg/__init__.py` to say, print something:
```
python2
Python 2.7.15 (default, May 16 2018, 17:50:09) 
[GCC 8.1.1 20180502 (Red Hat 8.1.1-1)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import ssg
DO SOMETHING
>>> 
```
It reflects changes to the module as they occur, just like setting `PYTHONPATH` would. 


(this is mostly about point 1).

Point 2, I think we just need to dissuade usage of `setup.py` directly and instead suggest using `pip` with the `-e` flag or suggest against `setup.py install` and do `setup.py develop`, but apparently (in 2014) the suggested was `pip` over `setup.py`: https://mail.python.org/pipermail/distutils-sig/2014-April/024112.html (I think `pip` also handles dependencies better than `setup.py`).

Point 3 I'll need to think about more.

dahaic:
@matejak can you take a look at this? :)

matejak:
test this please

matejak:
I am also more in favor of closing this - we can manage fairly well without `setup.py` (as @mpreisler points out) and we don't want to confuse people who could try to install the package system-wide.
I agree that time for `setup.py` may come in the future, so let's see about that.

Issue #1161, opened 2016-03-31
need new rule: GDM TimedLoginEnable=false


shawndwells:
![image](https://cloud.githubusercontent.com/assets/5713754/14188982/109a15f8-f759-11e5-851f-f4bd4ce59ba7.png)

bcon:
Could you sneak in a fix to this? The SPAWAR SCC, DISA STIG, and Tenable Nessus look for the false to be lowercase. The implementation of this in the SCAP content was "False" which makes a false positive result in Tenable Nessus and SCC scans.

Looking at GDM docs, the lowercase is probably technically correct (e.g. https://wiki.gnome.org/Projects/GDM/2.22/Configuration)

PR #3188, opened 2018-08-10
Add rule directory functionality to the build system

cipherboy:
A rule directory is a folder in a guide with the following properties:

 - The directory name is the rule id
 - The directory has the following structure:
	`rule.yml` --- metadata previously known as `{{{ rule_id }}}`.rule
	`ovals/` --- directory containing OVAL checks
	`anaconda/` --- directory containing anaconda remediations
	`ansible/` --- directory containing ansible remediations
	`bash/` --- directory containing bash remediations
	`puppet/` --- directory containing puppet remediations
 - In each of the `{ovals/, anaconda/, ansible/, bash/, puppet/}` directories,
   the structure is as follows, for the correct extension, `ext`, of content of
   that type:
	`shared.ext` --- content applicable to all products
	`{{{ product }}}.ext` --- content applicable only to `{{{ product }}}`

This adds support to the build system for rule directories, while
preserving backwards compatibility with rule files. Note that this
commit does not implement test suite support for test data under
the guide directories.

----

Documentation will be carried as part of #3178. Since the feedback on that PR has been largely positive (with the discussion largely focusing on the test suite changes), I've decided to propose this PR in parallel to speed up adoption and to unblock part of the merger from the SSG test suite related changes. These changes are ready to be reviewed as-is. To review the changes with the rules moving, I'd suggest pulling #3178 and testing that.

### DIFF
--- a/ssg/build_remediations.py
+++ b/ssg/build_remediations.py
@@ -6,15 +6,34 @@ import os
 import os.path
 import re
 import codecs
+from collections import defaultdict, namedtuple
 
+from .jinja import process_file as jinja_process_file
 from .xml import ElementTree
 from .products import parse_name, multi_list, map_name
 
+REMEDIATION_TO_EXT_MAP = {
+    'anaconda': '.anaconda',
+    'ansible': '.yml',
+    'bash': '.sh',
+    'puppet': '.pp'
+}
 
-def fix_is_applicable_for_product(platform, product):
+FILE_GENERATED_HASH_COMMENT = '# THIS FILE IS GENERATED'
+
+REMEDIATION_CONFIG_KEYS = ['complexity', 'disruption', 'platform', 'reboot',
+                           'strategy']
+REMEDIATION_ELM_KEYS = ['complexity', 'disruption', 'reboot', 'strategy']
+
+
+def is_applicable_for_product(platform, product):
     """Based on the platform dict specifier of the remediation script to
     determine if this remediation script is applicable for this product.
     Return 'True' if so, 'False' otherwise"""
+
+    # If the platform is None, platform must not exist in the config, so exit with False.
+    if not platform:
+        return False
 
     product, product_version = parse_name(product)
 
@@ -24,10 +43,9 @@ def fix_is_applicable_for_product(platform, product):
 
     # First test if platform isn't for 'multi_platform_all' or
     # 'multi_platform_' + product
-    result = False
     for _platform in multi_platforms:
         if _platform in platform and product in multi_list:
-            result = True
+            return True
 
     product_name = ""
     # Get official name for product
@@ -39,11 +57,11 @@ def fix_is_applicable_for_product(platform, product):
     # Test if this is for the concrete product version
     for _name_part in platform.split(','):
         if product_name == _name_part.strip():
-            result = True
+            return True
 
     # Remediation script isn't neither a multi platform one, nor isn't
     # applicable for this product => return False to indicate that
-    return product_name, result
+    return False
 
 
 def get_available_functions(build_dir):
@@ -111,17 +129,13 @@ def get_fixgroup_for_type(fixcontent, remediation_type):
 
 
 def is_supported_filename(remediation_type, filename):
-    if remediation_type == 'anaconda':
-        return filename.endswith('.anaconda')
+    """
+    Checks if filename has a supported extension for remediation_type.
 
-    elif remediation_type == 'ansible':
-        return filename.endswith('.yml')
-
-    elif remediation_type == 'bash':
-        return filename.endswith('.sh')
-
-    elif remediation_type == 'puppet':
-        return filename.endswith('.pp')
+    Exits when remediation_type is of an unknown type.
+    """
+    if remediation_type in REMEDIATION_TO_EXT_MAP:
+        return filename.endswith(REMEDIATION_TO_EXT_MAP[remediation_type])
 
     sys.stderr.write("ERROR: Unknown remediation type '%s'!\n"
                      % (remediation_type))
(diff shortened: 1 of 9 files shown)


cipherboy:
Thanks for assigning this, @matejak!

I've updated #3178 to fix a few bugs in the reorganization, so hopefully this'll look better now. They were discovered when analyzing the output of the new tool `rule_dir_stats.py` in #3193. Feel free to pull either of those and look at the output if you're interested in analyzing the build structure.

ssg/build_remediations.py
```diff
@@ -6,16 +6,31 @@
 import os.path
 import re
 import codecs
+from collections import defaultdict
 
+from .jinja import process_file as jinja_process_file
 from .xml import ElementTree
 from .products import parse_name, multi_list, map_name
 
+REMEDIATION_MAP = {
```
matejak: I would go for more descriptive `REMEDIATION_TO_EXTENSION_MAP`, so it is clear what is mapped to remediations.

ssg/build_remediations.py
```diff
@@ -6,16 +6,31 @@
...
+    'anaconda': '.anaconda',
+    'ansible': '.yml',
+    'bash': '.sh',
+    'puppet': '.pp'
+}
 
-def fix_is_applicable_for_product(platform, product):
+FILE_GENERATED = '# THIS FILE IS GENERATED'
```
matejak: I know that this is relocated, but what about `FILE_GENERATED_HASH_COMMENT`?

ssg/build_remediations.py
```diff
@@ -146,6 +151,81 @@ def get_populate_replacement(remediation_type, text):
     sys.exit(1)
 
 
+def from_file(file_path, env_yaml):
```
matejak: This name is really not descriptive.

ssg/build_remediations.py
```diff
@@ -146,6 +151,81 @@ def get_populate_replacement(remediation_type, text):
...
+
+    fix_file_lines = jinja_process_file(file_path, env_yaml).splitlines()
+
+    # Assignment automatically escapes shell characters for XML
+    for line in fix_file_lines:
+        line += "\n"
+        if line.startswith('#'):
+            try:
+                (key, value) = line.strip('#').split('=')
+                if key.strip() in ['complexity', 'disruption',
```
matejak: The collection of attributes could be defined in a nice constant bearing a descriptive name.

ssg/build_remediations.py
```diff
@@ -146,6 +151,81 @@ def get_populate_replacement(remediation_type, text):
...
+            except ValueError:
+                if not line.startswith(FILE_GENERATED):
+                    mod_file.append(line)
+        else:
+            mod_file.append(line)
+
+    return mod_file, config
+
+
+def process_fix(fixes, args, env_yaml, product, file_path, fix_name):
```
matejak: Scrutinizer is right about arguments, but it is difficult to do something sensible with it without heavy refactoring. At least passing `args` is confusing from the caller perspective, as it is not obvious that the function takes a look only at remediation type.

ssg/build_remediations.py
```diff
@@ -146,6 +151,81 @@ def get_populate_replacement(remediation_type, text):
...
+    if not is_supported_filename(args.remediation_type, file_path):
+        return
+
+    fix_contents, config = from_file(file_path, env_yaml)
+
+    if is_applicable_for_product(config['platform'], product):
+        fixes[fix_name] = (fix_contents, config)
+    elif not config['platform']:
+        raise RuntimeError(
```
matejak: This exception could have been raised by some higher-level wrapper of `from_file`.

ssg/build_remediations.py
```diff
@@ -146,6 +151,81 @@ def get_populate_replacement(remediation_type, text):
     sys.exit(1)
 
 
+def from_file(file_path, env_yaml):
+    mod_file = []
+    config = defaultdict(lambda: None)
```
matejak: Here, we make sure that we will end up with dictionary containing all keys, but some of them may be `None`. Later in the code, you have to check whether the associated value is not `None`, so you have a sequence of `if`s instead of one loop (in `write_fixes`). Check it out!

ssg/build_remediations.py
```diff
@@ -146,6 +151,81 @@ def get_populate_replacement(remediation_type, text):
...
+        if config['complexity']:
+            fix_elm.set("complexity", config['complexity'])
+        if config['disruption']:
+            fix_elm.set("disruption", config['disruption'])
+        if config['reboot']:
+            fix_elm.set("reboot", config['reboot'])
+        if config['strategy']:
+            fix_elm.set("strategy", config['strategy'])
+
+        fix_elm.text = "".join(fix_contents)
```
matejak: What about joining over a newline in exchange of not appending newline to those strings in `from_file`?

ssg/rules.py
```diff
@@ -0,0 +1,130 @@
...
+    results = []
+    for oval_file in ["%s.xml" % product, "shared.xml"]:
+        oval_path = os.path.join(oval_dir, oval_file)
+        if os.path.isfile(oval_path):
+            results.append(oval_path)
+
+    return results
+
+
+def get_rule_dir_remediations(dir_path, remediation_type, product=None):
```
matejak: Ditto, the best thing would be to somehow reuse the code pattern (but such requirement is out of scope of this PR).

ssg/rules.py
```diff
@@ -0,0 +1,130 @@
...
+    """
+    rule_yaml = get_rule_dir_yaml(dir_path)
+
+    is_dir = os.path.isdir(dir_path)
+    has_rule_yaml = os.path.exists(rule_yaml)
+
+    return is_dir and has_rule_yaml
+
+
+def get_rule_dir_ovals(dir_path, product=None):
```
matejak: This function does two quite different things depending on whether product is None. What about having a private function with common part and one function without the product argument and another one that accepts it? Do we have a use case for a call with a product that is None?

ssg/build_remediations.py
```diff
@@ -39,11 +53,11 @@ def fix_is_applicable_for_product(platform, product):
     # Test if this is for the concrete product version
     for _name_part in platform.split(','):
         if product_name == _name_part.strip():
-            result = True
+            return True
 
     # Remediation script isn't neither a multi platform one, nor isn't
     # applicable for this product => return False to indicate that
-    return product_name, result
+    return False
```
matejak: I appreciate the return value change very much.

ssg/rules.py
cipherboy: @matejak Yes, the use case is here:

https://github.com/cipherboy/scap-security-guide/blob/rule-dirs-tools/utils/rule_dir_json.py#L94

I would argue that it isn't that it does two _different_ things based on the value of `product`, but more that `product` is a filter which limits the output of `get_rule_dir_ovals` to only things which _could_ apply to that value of `product`, where `None` represents `all`.

The code could be improved by forward porting the [`is_applicable` function](https://github.com/cipherboy/scap-security-guide/blob/rule-dirs-tools/ssg/rules.py#L10-L20) (+ dependencies to parse the `<platform>` and configuration in remediations) introduced in that same `rule-dirs-tools` branch (for #3193) and having `is_applicable` return True `if product == None`. While that wouldn't _technically_ affect the build system (as the pruning of OVALs / remediations happen at a later time), it'd view it as a semantic change and in the case of `remediations`, requires the introduction of parsing `jinja`, which is best done with the entire build config (so we can use the jinja bytecode cache).

However, that would break the later use case in `rule-dirs-tools` and complicate the function signature. At that point, you might as well return the parsed OVAL/fix and that'd also be discarded for the tools. 

Thoughts?

ssg/build_remediations.py
cipherboy: Would adding a descriptive doc string here be acceptable instead of renaming the function?

I'm partial to functions which aren't repetitive. The fully qualified name to this function is `ssg.build_remediations.from_file(...)`, which I believe makes sense. Most of the time, we call `import ssg.build_remediations as remediation` (`build-scripts/combine_remediations.py#L11`), so the qualified name is `remediation.from_file(...)` which also makes sense.

cipherboy: I've also converted the return type to a `namedtuple` so that should be more obvious as well.

ssg/build_remediations.py
cipherboy: Sure. One quirk is that `fix_elm.text` needs to end with a new line due to `expand_xccdf_subs`. Don't ask me why but yeah:

```
Error performing XCCDF <sub> substitution for function package_remove in package_nss-pam-ldapd_removed fix. Exiting...
```

ssg/build_remediations.py
cipherboy: I think this is the higher-level wrapper of `from_file`? I can't move it to `from_file` because I use `from_file` in the `rule_dirs_json` to parse the config element. And creating a wrapper for `from_file` just to check if `platform` exists in the config for use in this function doesn't make sense to me.

I moved this up to before the if statement that sets fixes so hopefully this is better.

ssg/rules.py
cipherboy: Yep, that's my current implementation. I'll push changes soon. :)

ssg/build_remediations.py
matejak: I see your point. I think that e.g. `parse_from_file` won't ruin the fully qualified name, while it hints that some processing of the text and maybe even metadata extraction may take place.

On an unrelated note, I prefer function names over docstrings because docstrings may become obsolete and consequently misleading before you know it.

ssg/rules.py
cipherboy: Yep, this is updated. :)

matejak:
Great job, I have checked your changes and I like them very much and I really appreciate those unit tests!
I have verified that the functionality works, so there is one last thing to do - you can basically rip parts of your PR description and incorporate it into developer doc, so the change of how rules can be defined is properly documented.

cipherboy:
@matejak yep! That's on my docket of things to do. Since I have two other PRs which depend on this one (and changes to this one mean I have to rebase both of those), I've been wanting to stick that in a separate PR. Mind if I do that after openscap #1161?

cipherboy:
`hostname: ca54afb3-5f29-448f-ac5d-7b784313ada3@1.production-1-worker-org-a-5-gce` -- CentOS failure on Travis, restarting.

cipherboy:
@matejak This is done in #3201. :)

matejak:
That's OK then, as this PR keeps the backward compatibility, we don't have to have those docs here at all costs, and they can be handled separately.
Great work, thank you very much!

PR #3191, opened 2018-08-13
Fix required_key usage on build-only key

cipherboy:
Not sure why we're using `required_key` on this... Thought I fixed it earlier. 

`jinja2_cache_dir` depends on `jinja2_cache_enabled` being true, so that makes sense to use a `required_key` there, but we should be perfectly fine if we don't have a `jinja2_cache_enabled` key--default to `False`!

### DIFF
--- a/ssg/jinja.py
+++ b/ssg/jinja.py
@@ -45,7 +45,7 @@ class AbsolutePathFileSystemLoader(jinja2.BaseLoader):
 def _get_jinja_environment(substitutions_dict):
     if _get_jinja_environment.env is None:
         bytecode_cache = None
-        if required_key(substitutions_dict, "jinja2_cache_enabled") == "true":
+        if substitutions_dict.get("jinja2_cache_enabled") == "true":
             bytecode_cache = jinja2.FileSystemBytecodeCache(
                 required_key(substitutions_dict, "jinja2_cache_dir")
             )


ssg/jinja.py
```diff
@@ -45,7 +45,8 @@ def uptodate():
 def _get_jinja_environment(substitutions_dict):
     if _get_jinja_environment.env is None:
         bytecode_cache = None
-        if required_key(substitutions_dict, "jinja2_cache_enabled") == "true":
+        have_key = 'jinja2_cache_enabled' in substitutions_dict
+        if have_key and substitutions_dict["jinja2_cache_enabled"] == "true":
```
mpreisler: nicer way to do this is `substitutions_dict.get("jinja2_cache_enabled") == "true"`

cipherboy: @mpreisler Good point, forgot about `get`. :) Updated.

mpreisler:
Can you describe the case when the jinja2_cache_enabled key won't be in the dict? We always have it in build_config.yml so I am thinking it not being there might be a bug somewhere else that we would hide with this change.

cipherboy:
Well, for instance, in my branch [`rule-dir-tools`](https://github.com/cipherboy/scap-security-guide/blob/rule-dirs-tools/utils/rule_dir_json.py#L135), I need _a_ `product.yml` to parse the remediations config (because remediations use `jinja`), but I don't care which one that is because in practice, no remediation uses `jinja` in the config. I could write a separate parser for remediations, but there's no point in duplicating that code when I use similar hacks for parsing the [`rule.yml`](https://github.com/cipherboy/scap-security-guide/blob/rule-dirs-tools/utils/rule_dir_json.py#L70-L73) -- for `prodtype` and `title`, neither of which change. Technically I need it for parsing OVAL's `platform` as well, but in practice this hasn't been an issue because OVAL just puts it all in the `text` element.

cipherboy:
This is mostly important when doing stuff with various utilities to interact with SSG content, e.g.:

- https://github.com/OpenSCAP/scap-security-guide/blob/master/utils/fix-rules.py#L75-L128

PR #3192, opened 2018-08-13
CMAKE_BUILD_TYPE support

mpreisler:
The only thing that's different when using the Debug build type is the addition of documentation_complete=false content. That let's developers test work in progress content conveniently without flipping switches back and forth.

image

### DIFF
--- a/ssg/yaml.py
+++ b/ssg/yaml.py
@@ -28,7 +28,7 @@ def _save_rename(result, stem, prefix):
     result["{0}_{1}".format(prefix, stem)] = stem
 
 
-def _open_yaml(stream, original_file=None):
+def _open_yaml(stream, original_file=None, substitutions_dict={}):
     """
     Open given file-like object and parse it as YAML.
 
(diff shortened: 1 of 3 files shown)


ssg/yaml.py
```diff
@@ -28,7 +28,7 @@ def _save_rename(result, stem, prefix):
     result["{0}_{1}".format(prefix, stem)] = stem
 
 
-def _open_yaml(stream, original_file=None):
+def _open_yaml(stream, original_file=None, substitutions_dict={}):
```
cipherboy: I think the one new scrutinizer issue here is a non-issue.

mpreisler: In this case it's not an issue but mutable default args are a huge issue in Python. It's super confusing and hard to debug and not even reentrant. I can change it to substitutions_dict=None and then instantiate a dict if it's None inside the function.

Again, in this specific case it's not an issue because we never modify it but if anyone adds a code that adds something to that dict we will be surprised about the results :-D

PR #3189, opened 2018-08-13
OL7 pci-dss aide and password hashing rules

iokomin:
#### Description:

Simple PR to add OL7 support for existing aide and password hashing pci-dss rules

#### Rationale:

Add more ol7 pci-dss rules

#### Testing (Oracle Linux 7):
- Checked OVAL and remediation scripts content
- Checked guide content
- Checked  rules evaluation an remediation scripts results

PR #3194, opened 2018-08-13
Add multi_platform_fedora to AIDE and password hasing checks and fixes

cipherboy:
While reviewing #3189, I discovered that several of the updated checks and remediations should apply to Fedora as well. This updates them accordingly. 

Since they apply to rhel6, rhel7 and ol7, I'm fairly confident these will apply on Fedora. Further, the `aide` package exists on Fedora so this is valid.


cipherboy:
`hostname: b9986aca-4fed-46a6-9472-3c625e3b382c@1.production-4-worker-org-f-1-gce` -- Travis failure unrelated. CentOS jinja2 failure. Restarting.

PR #3195, opened 2018-08-14
Add remediation and tests for mount_option_noexec_remote_filesystems rule

mildass:
#### Description:
Add bash and ansible remediation with tests for Mount Remote Filesystems with noexec rule.

Similar to #3180 PR. The only difference is that this checks `noexec` option.


mildass:
Closing this one because I'll template all `mount_option` rules

PR #3199, opened 2018-08-15
Install dracut-fips when fips mode is enabled in the profile.

dahaic:
#### Description:

- Include installation of dracut_fips in profiles that enables fips 

#### Rationale:

- When we enable fips in grup config, we should also install dracut_fips package. Otherwise, machine might not work properly.

- Fixes RHBZ#1587911 (cannot boot machine installed with disa stig, when root is encrypted)

### DIFF
--- a/linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/rule.yml
+++ b/linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/rule.yml
@@ -5,8 +5,9 @@ prodtype: rhel7
 title: 'Enable FIPS Mode in GRUB2'
 
 description: |-
-    To ensure FIPS mode is enabled, rebuild <tt>initramfs</tt> by running the following command:
-    <pre>dracut -f</pre>
+    To ensure FIPS mode is enabled, install package <tt>dracut-fips</tt>, and rebuild <tt>initramfs</tt> by running the following commands:
+    <pre>{{{ package_install("dracut-fips") }}}
+    dracut -f</pre>
     After the <tt>dracut</tt> command has been run, add the argument <tt>fips=1</tt> to the default
     GRUB 2 command line for the Linux operating system in
     <tt>/etc/default/grub</tt>, in the manner below:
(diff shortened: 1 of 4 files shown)


redhatrises:
We probably shouldn't add a non-DISA rule to a DISA profile. `grub2_enable_fips_mode` could get an `.anaconda` rule to handle the dracut-fips installation.

yuumasato:
> grub2_enable_fips_mode could get an .anaconda rule to handle the dracut-fips installation.

This could be the way to go, very similar to approach taken for [rule `smartcard_auth`](https://github.com/OpenSCAP/scap-security-guide/search?q=filename%3Asmartcard_auth&unscoped_q=filename%3Asmartcard_auth). Bash remediation for `grub2_enable_fips_mode` already installs `dracut-fips` package.

yuumasato:
> We probably shouldn't add a non-DISA rule to a DISA profile.

This DISA non-DISA rule thing can get annoying, at times.

ykorkmaz:
dracut-fips-aesni should be installed as well when fips is enabled.

dahaic:
@ykorkmaz it's a good point with dracut-fips-aesni. I am not sure at this moment how this package behaves when cpu does not support aesni. So I propose to leave this PR without it, and I will create new PR when I get the information. (dracut-fips-aesni, as far as I understand the purpose, won't make computer safer, just faster for some operations, thus it's optional)

linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/rule.yml
```diff
@@ -5,8 +5,9 @@ prodtype: rhel7
 title: 'Enable FIPS Mode in GRUB2'
 
 description: |-
-    To ensure FIPS mode is enabled, rebuild <tt>initramfs</tt> by running the following command:
-    <pre>dracut -f</pre>
+    To ensure FIPS mode is enabled, install package <tt>dracut-fips</tt>, and rebuild <tt>initramfs</tt> by running the following commands:
+    <pre>yum install -y dracut-fips
```
yuumasato: Although this rule currently only applies to `rhel7`, I think it makes sense to use `{{{ pkg_manager }}}` instead of `yum` command.

dahaic:
@ykorkmaz I have created issue #3221 as it's more complex problem then could be solved in this PR, and will probably take some time.

@yuumasato updated description to use macro, and updated macro to support this exact usecase, as the original macro put ugly descriptive wrapped around the command :)

yuumasato:
Nice, thanks for the patch!
I have also checked macro `yum_describe_package_install`, and related, and they are not used anywhere else.
ACK.

Issue #2876, opened 2018-05-30
Provide fast and simple example Profile


yuumasato:
#### Description of problem:

Often we would like to have easy profile:
- to just perform a quick evaluation
- that is easy to understand and simple to use
- to use as sample for tutorial or presentation

#### Actual Results:
There is no dedicated Profile for quick and easy scan, such Profile needs to be created from scratch or customized from existing Profile.

#### Expected Results:
There is an example Profile with simple rules selected, with focus on rules that are easy to understand and fast to scan.

#### Addition Information/Debugging Steps:
Inspired by https://github.com/OpenSCAP/openscap/issues/198

PR #3202, opened 2018-08-16
Add a sample product for demonstration purposes

cipherboy:
Closes: #2876

I chose these because they're generally applicable to both Fedora and RHEL, but also can be run on other platforms. 

```
$ time sudo oscap xccdf eval --profile sample ./ssg-fedora-ds.xml 
Title   Ensure gpgcheck Enabled For All Yum Package Repositories
Rule    xccdf_org.ssgproject.content_rule_ensure_gpgcheck_never_disabled
Result  pass

Title   Ensure gpgcheck Enabled In Main Yum Configuration
Rule    xccdf_org.ssgproject.content_rule_ensure_gpgcheck_globally_activated
Result  pass

Title   Verify Permissions on passwd File
Rule    xccdf_org.ssgproject.content_rule_file_permissions_etc_passwd
Result  pass

Title   Verify User Who Owns shadow File
Rule    xccdf_org.ssgproject.content_rule_file_owner_etc_shadow
Result  pass

Title   Verify Group Who Owns shadow File
Rule    xccdf_org.ssgproject.content_rule_file_groupowner_etc_shadow
Result  pass

Title   Verify Group Who Owns passwd File
Rule    xccdf_org.ssgproject.content_rule_file_groupowner_etc_passwd
Result  pass

Title   Verify Permissions on shadow File
Rule    xccdf_org.ssgproject.content_rule_file_permissions_etc_shadow
Result  pass

Title   Verify Permissions on group File
Rule    xccdf_org.ssgproject.content_rule_file_permissions_etc_group
Result  pass

Title   Verify Group Who Owns gshadow File
Rule    xccdf_org.ssgproject.content_rule_file_groupowner_etc_gshadow
Result  pass

Title   Verify User Who Owns gshadow File
Rule    xccdf_org.ssgproject.content_rule_file_owner_etc_gshadow
Result  pass

Title   Verify User Who Owns passwd File
Rule    xccdf_org.ssgproject.content_rule_file_owner_etc_passwd
Result  pass

Title   Verify Group Who Owns group File
Rule    xccdf_org.ssgproject.content_rule_file_groupowner_etc_group
Result  pass

Title   Verify Permissions on gshadow File
Rule    xccdf_org.ssgproject.content_rule_file_permissions_etc_gshadow
Result  pass

Title   Verify User Who Owns group File
Rule    xccdf_org.ssgproject.content_rule_file_owner_etc_group
Result  pass

Title   Prevent Log In to Accounts With Empty Password
Rule    xccdf_org.ssgproject.content_rule_no_empty_passwords
Result  fail

Title   Disable SSH Root Login
Rule    xccdf_org.ssgproject.content_rule_sshd_disable_root_login
Result  fail

Title   Set SSH Idle Timeout Interval
Rule    xccdf_org.ssgproject.content_rule_sshd_set_idle_timeout
Result  fail

Title   Disable SSH Access via Empty Passwords
Rule    xccdf_org.ssgproject.content_rule_sshd_disable_empty_passwords
Result  fail


real	0m1.675s
user	0m1.004s
sys	0m0.224s
```

Fixing something then moves the result to true:

```
Title   Ensure gpgcheck Enabled For All Yum Package Repositories
Rule    xccdf_org.ssgproject.content_rule_ensure_gpgcheck_never_disabled
Result  pass

Title   Ensure gpgcheck Enabled In Main Yum Configuration
Rule    xccdf_org.ssgproject.content_rule_ensure_gpgcheck_globally_activated
Result  pass

Title   Verify Permissions on passwd File
Rule    xccdf_org.ssgproject.content_rule_file_permissions_etc_passwd
Result  pass

Title   Verify User Who Owns shadow File
Rule    xccdf_org.ssgproject.content_rule_file_owner_etc_shadow
Result  pass

Title   Verify Group Who Owns shadow File
Rule    xccdf_org.ssgproject.content_rule_file_groupowner_etc_shadow
Result  pass

Title   Verify Group Who Owns passwd File
Rule    xccdf_org.ssgproject.content_rule_file_groupowner_etc_passwd
Result  pass

Title   Verify Permissions on shadow File
Rule    xccdf_org.ssgproject.content_rule_file_permissions_etc_shadow
Result  pass

Title   Verify Permissions on group File
Rule    xccdf_org.ssgproject.content_rule_file_permissions_etc_group
Result  pass

Title   Verify Group Who Owns gshadow File
Rule    xccdf_org.ssgproject.content_rule_file_groupowner_etc_gshadow
Result  pass

Title   Verify User Who Owns gshadow File
Rule    xccdf_org.ssgproject.content_rule_file_owner_etc_gshadow
Result  pass

Title   Verify User Who Owns passwd File
Rule    xccdf_org.ssgproject.content_rule_file_owner_etc_passwd
Result  pass

Title   Verify Group Who Owns group File
Rule    xccdf_org.ssgproject.content_rule_file_groupowner_etc_group
Result  pass

Title   Verify Permissions on gshadow File
Rule    xccdf_org.ssgproject.content_rule_file_permissions_etc_gshadow
Result  pass

Title   Verify User Who Owns group File
Rule    xccdf_org.ssgproject.content_rule_file_owner_etc_group
Result  pass

Title   Prevent Log In to Accounts With Empty Password
Rule    xccdf_org.ssgproject.content_rule_no_empty_passwords
Result  fail

Title   Disable SSH Root Login
Rule    xccdf_org.ssgproject.content_rule_sshd_disable_root_login
Result  pass

Title   Set SSH Idle Timeout Interval
Rule    xccdf_org.ssgproject.content_rule_sshd_set_idle_timeout
Result  fail

Title   Disable SSH Access via Empty Passwords
Rule    xccdf_org.ssgproject.content_rule_sshd_disable_empty_passwords
Result  pass


real	0m1.562s
user	0m0.966s
sys	0m0.208s
```

Note that these checks require admin still -- do we also want to ensure you don't need `sudo` permission to run them? If so, I can remove the `ssh` checks.


yuumasato:
@cipherboy I don't think need of root privileges is a problem, it can serve the of setting expectations of what are the requirements to scan and remediate with openscap.

I'm thinking that maybe this example profile could be on its own product, applicable to all things, or maybe all linuxes, for now. These sample profiles, marked `documentation_complete: true`, among other profiles, and spread across products can be obnoxious and create extra maintenance for distribution/shipping.

On a quick glance the selection of Rules look good.

redhatrises:
Agree with what @yuumasato said. In fact, the `standard` profile would better serve as the example profile rather than creating a new one.

cipherboy:
@redhatrises The issue with the standard profile (which I'm not sure what "standard" it is conforming too -- perhaps if its not you're suggesting we should prune it down to a smaller size) is that it is slow (e.g., `rpm-verify-*`, `file_permissions_{library, binary}_dirs`, etc). 

If we can change the `standard` profile, then yeah, I'm in favor of moving `example` -> `standard` and just pruning `standard`, but it'd be interesting to see if anyone is relying on the contents of `standard`.

jan-cerny:
I agree with @cipherboy that the standard profile is bad for this purpose. Sometimes, you need to check if something works with SCAP Workbench or `oscap` on various systems and you have to either apply a tailoring or wait a terribly long time. Therefore I like this example profile very much because it solves all my use cases.

redhatrises:
We should only have one example profile.... what even be better is to build an example profile that works regardless of os.

cipherboy:
This product will only work on RHEL-like as it might depend on `dnf`, but is otherwise its own profile. Thoughts?

redhatrises:
I like these changes.

CMakeLists.txt
```diff
@@ -217,6 +218,7 @@ message(STATUS "Ubuntu 14: ${SSG_PRODUCT_UBUNTU14}")
 message(STATUS "Ubuntu 16: ${SSG_PRODUCT_UBUNTU16}")
 message(STATUS "WRLinux: ${SSG_PRODUCT_WRLINUX}")
 message(STATUS "Oracle Linux 7: ${SSG_PRODUCT_OL7}")
+message(STATUS "Example: ${SSG_PRODUCT_EXAMPLE}")
```
yuumasato: This cosmetic change, but it can be more visible if on top of the list.

example/transforms/xccdf2table-profilecisrefs.xslt
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8" standalone="yes"?>
```
yuumasato: Is think  this file is not used because no tables are built.
This should be removed or a CIS reference table added.

example/profiles/default.profile
```diff
@@ -0,0 +1,26 @@
+documentation_complete: true
```
yuumasato: I think `example` is a better ID than `default`.
The `default` profile is implicit to be the Benchmark "as is", with the rule selection not being "altered" by a profile.
Also, it can be less misleading to us when somebody says they tried the `example` profile, rather then the `default` profile.

cipherboy: Okay, I changed this to `example` profile in `example` product.

CMakeLists.txt
cipherboy: I've moved everything about this product to be at the same position (top) of their respective sections.

CMakeLists.txt
```diff
@@ -66,6 +66,7 @@ option(SSG_PRODUCT_UBUNTU14 "If enabled, the Ubuntu14 SCAP content will be built
 option(SSG_PRODUCT_UBUNTU16 "If enabled, the Ubuntu16 SCAP content will be built" TRUE)
 option(SSG_PRODUCT_WRLINUX "If enabled, the WRLinux SCAP content will be built" TRUE)
 option(SSG_PRODUCT_OL7 "If enabled, the Oracle Linux 7 SCAP content will be built" TRUE)
+option(SSG_PRODUCT_EXAMPLE "If enabled, the Example SCAP content will be built" TRUE)
```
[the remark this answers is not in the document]
cipherboy: Fair point. Can we build it for release on at least Fedora and perhaps RHEL/CentOS?

cipherboy:
@yuumasato Updated! A few comments inline. :)

CMakeLists.txt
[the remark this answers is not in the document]
redhatrises: Agree with @yuumasato. Shouldn't add the example into a release. I am not seeing the benefit considering it's greatest use would be for development.

cipherboy:
Okay. This is disabled by default in this PR right now btw.

CMakeLists.txt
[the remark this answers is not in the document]
yuumasato: > Can we build it for release on at least Fedora and perhaps RHEL/CentOS?

@cipherboy I guess this is up to the packagers, no?

yuumasato:
The example product is there and not built by default, but I wonder how discoverable it is and how people can build it.

I think there should be documentation about this example product, something in `cmake` output or a `README` file in `example` directory. But I won't block this.

CMakeLists.txt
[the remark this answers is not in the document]
cipherboy: @yuumasato Yes, agreed that it is. For Fedora at least, many of the reviewers of OpenSCAP upstream are also packagers / can influence packaging: https://src.fedoraproject.org/rpms/scap-security-guide 

Perhaps this was the wrong place to suggest including this change though.

PR #3201, opened 2018-08-16
Add developer documentation on rule directories

cipherboy:
## Feedback actively requested.

This documents the structure of rule directories and the existence
of the `rule_dir_*.py` utilities.

Signed-off-by: Alexander Scheel <ascheel@redhat.com>

This PR describes the changes made in #3178, #3188, and #3193 to support rule directories. 

I'd like #3204 to merge first, to give us a description for `rule.yml`.

PR #3204, opened 2018-08-16
Update developer manual to describe YAML rules

cipherboy:
YAML formatting for rules was introduced in #2592 by @mpreisler,
but weren't documented. This adds documentation about fields
in a rule and updates the referenced locations to be current.

Very useful when documenting what a rule directory is, to have something to reference about the `rule.yml` content (#3201). :)

### DIFF
--- a/docs/manual/developer_guide.adoc
+++ b/docs/manual/developer_guide.adoc
@@ -570,22 +570,31 @@ Contributions can be made for rules, checks, remediations or even utilities. The
 
 ==== Rules
 
-Rules are input in a simplified XCCDF format, which is basically an XML container.
-Rules are defined as members of a `Group` in a XML file.
-In the likely case if the rule can be reused in multiple platforms, the file containing the definition will be placed under the `shared/xccdf` directory.
-If the rule is platform-specific, place it under the `<platform>/xccdf` directory.
-The exact location depends on a rule category.
-For an example of rule group, see `shared/xccdf/system/software/disk_partitioning.xml`.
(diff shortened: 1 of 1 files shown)


docs/manual/developer_guide.adoc
```diff
@@ -594,16 +600,17 @@ The <tt>/tmp</tt> partition is used as temporary storage by many programs. Placi
 * `description`: Human-readable HTML description, which provides broader context for non-experts than the rationale. For example, description of the `partition_for_tmp` rule states that:
 +
 The <tt>/var/tmp</tt> directory is a world-writable directory used for temporary file storage. Ensure it has its own partition or logical volume at installation time, or migrate it using LVM.
-* `ocil`: Defines assert statements. The `clause` attribute contains the statement, and the element text describes how to determine whether the statement is true or false. Check out rule `encrypt_partitions` in `shared/xccdf/system/software/disk_partitioning.xml`, that contains the `partitions do not have a type of crypto_LUKS` clause.
+* `ocil`: Defines asserting statements to check whether or not the rule is valid.
+* `ocil_clause` attribute contains the statement, and the element text describes how to determine whether the statement is true or false. Check out `encrypt_partitions.rule` in `linux_os/guide/system/software/disk_partitioning/`: this contains a `partitions do not have a type of crypto_LUKS` `ocil_clause`.
 
 A rule may contain those reference-type elements:
```
jan-cerny: I think it mixes elements, attributes, XML and YAML. I'd suggest explicitly defining that this section describes YAML structure and then adjust the rest of text accordingly.

PR #3208, opened 2018-08-20
Introduce Rawhide product

yuumasato:
#### Description:

Introducing Fedora Rawhide product.
Rawhide encompasses content from `fedora` and `rhel7`.


yuumasato:
Some checks and remediations are not included in datastream because there are only shared versions of them, e.g. [audit_rules_file_deletion_events](https://github.com/OpenSCAP/scap-security-guide/blob/master/linux_os/guide/system/auditing/auditd_configure_rules/audit_file_deletion_events/audit_rules_file_deletion_events/bash/rhel7.sh).
I'm thinking about renaming them to `shared.xml`, and add `rawhide` platform. Thoughts?

cipherboy:
@yuumasato said:
> Some checks and remediations are not included in datastream because there are no shared versions of them, e.g. audit_rules_file_deletion_events.
> I'm thinking about renaming them to shared.xml, and add rawhide platform. Thoughts?

I'd agree with that; I think they should also probably have non-rawhide Fedora added (`multi_platform_fedora`) as they'll largely be applicable there too.

redhatrises:
Nack. This should all be under Fedora.

redhatrises:
Realized that the nack may have sounded harsh. What I really am trying to say is how about we use Fedora for all Future content development so that we can keep doing bleeding edge work on Fedora. That way we minimize our future workload.

yuumasato:
> What assurance exists that checks created for RHEL work on Fedora Rawhide the exact same way?

@shawndwells None, rawhide being rawhide, it would be implied that somethings could be broken.

 > use Fedora for all Future content development so that we can keep doing bleeding edge work on Fedora. That way we minimize our future workload.

@redhatrises Well, this makes sense, Fedora is in need of some love.
Also, doing bleeding edge work on Fedora shall raise levels of testing needed for it, the product is always spanning two releases plus rawhide, great chances of incompatibilities popping out.

> With Rawhide being Fedora's daily spin, still questioning how it'll be possible to keep content updated and functional.

@shawndwells The same question currently applies to Fedora, it has been rolling versions since beginnings of time, what assurance there is that some rule developed at times of an older release still work on the latest greatest Fedora?

It seems to me that developing these Profiles on Fedora is the way, along with tests across current versions of Fedora.

PR #3209, opened 2018-08-20
SSG - Module Documentation

cipherboy:
This PR adds a bunch of docstrings to the build "sub module" of SSG.


cipherboy:
The four new issues found are non-issues / issues before this PR. I think it thinks they're new because the lines changed? Not entirely sure.

mpreisler:
Great stuff!

PR #3205, opened 2018-08-16
SSG unit tests

cipherboy:
Random miscellaneous unit tests and small improvements to the SSG module.

PR #3193, opened 2018-08-13
Rule directory tools

cipherboy:
This is a WIP PR dependent upon several others, but is opened to provide visibility.

#### Dependencies:

Currently several commits from other PRs are carried here. This PR will be rebased as necessary.

 - ~#3188 - To add build support~ -- merged, thanks @matejak!
 - ~#3178 - To reorganize rules so this has something to use~ -- merged, thanks @mpreisler! 
 - ~#3191~ - Because I'm not parsing rules as part of the build system -- merged, thanks @mpreisler!
 - ~#3198~ - Because this is a useful function -- merged, thanks @matejak!

#### Description:

This PR adds tools for inspecting the source SSG content from the perspective of rule directories. This adds three main utilities:

 - `rule_dir_json` -- builds a JSON object that represents the state of the SSG content. This allows the FS to be walked once for all of SSG, parses `product_yaml` files, understands `prodtype`s, and creates one large object which can be walked. While the order of the keys are not necessarily stable, all lists are sorted. providing relatively stable output.
 - `rule_dir_stats` -- interacts with the JSON object from `rule_dir_json`  to display useful information. This'll include filtering by product, listing rules without OVALs/remediations, listing rules with too many OVALs/remediations. faulty prodtypes, and other sanity checks. ~**WIP**~ Done! See output below. 
- `rule_dir_diff` -- interacts with two JSON objects and diffs the objects, but otherwise provides a similar functionality as `rule_dir_stats`. ~**WIP**~ -- Done! See output below.

Hopefully these will prove useful long term in addition to me during the migration. My thought is that `rule_dir_stats` might be useful to find places to contribute, and `rule_dir_diff` might be useful for reviewing large PRs. 

**To be clear:** these operate on the source content artifacts (of `linux_os/guide`, etc.), not on the built content. Hence they are different than `profile-stats` or `stats` in the build system. 

#### rule_dir_json

Help Text:
```
usage: rule_dir_json.py [-h] [-r ROOT] [-o OUTPUT]

optional arguments:
  -h, --help            show this help message and exit
  -r ROOT, --root ROOT  Path to SSG root directory (defaults to pwd)
  -o OUTPUT, --output OUTPUT
                        File to write json output to (defaults to output.json)
```

Description (from commit d1536884f101f855dd09848ed3f8731f66e016fd):
```
Add tool to walk rule directories and output json

The new utils/rule_dir_json.py walks the SSG root and
locates all rule directories across all products. These
rule directories are parsed for check and fix content,
[...]

PR #3214, opened 2018-08-21
SSG - More Unit Tests

cipherboy:
After the popular success that is #3205, I've decided to add another couple!

(Also, paying penance for the untestable mess that is #3193).

### DIFF
--- a/ssg/ansible.py
+++ b/ssg/ansible.py
@@ -1,5 +1,5 @@
 """
-Common ansible functions for SSG
+Common functions for processing Ansible in SSG
 """
 
 from __future__ import absolute_import
@@ -7,6 +7,7 @@ from __future__ import print_function
 
 from .constants import ansible_version_requirement_pre_task_name
 from .constants import min_ansible_version
+
 
 def add_minimum_version(ansible_src):
     """
@@ -21,5 +22,15 @@ def add_minimum_version(ansible_src):
            "You must update Ansible to at least version %s to use this role."
           """ % (ansible_version_requirement_pre_task_name,
                  min_ansible_version, min_ansible_version))
+
+    if ' - hosts: all' not in ansible_src:
+        return ansible_src
+
+    if 'pre_task' in ansible_src:
+        if 'ansible_version.full is version_compare' in ansible_src:
+            return ansible_src
+
+        raise ValueError("A pre_task already exists in ansible_src; failing to process: %s" %
+                         ansible_src)
 
     return ansible_src.replace(" - hosts: all", pre_task, 1)
--- a/tests/unit/ssg/test_ansible.py
+++ b/tests/unit/ssg/test_ansible.py
@@ -2,50 +2,57 @@ import pytest
 
 import os
 import re
+import textwrap
 import ssg.ansible
 
 
-def no_whitespace_compare(left, right):
+def strings_equal_except_whitespaces(left, right):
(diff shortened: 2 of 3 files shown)


tests/unit/ssg/test_ansible.py
```diff
@@ -0,0 +1,56 @@
...
+def no_whitespace_compare(left, right):
+    return re.sub(r'\s', '', left) == re.sub(r'\s', '', right)
+
+
+def test_add_minimum_version():
+    good_snippet = """
+ - hosts: all
+   vars:
+   tasks:
+"""
```
cipherboy: I'm not sure if we want this indented. I originally de-indented these because of whitespace differences between that and the output, but then I added the `no_whitespace_compare` function and decided to just ignore whitespace altogether. 

However, I find that four space indent on the triple quotes look funny (because the rest have to be 3 spaces because YAML....). Thoughts?

ssg/ansible.py
```diff
@@ -22,4 +23,14 @@ def add_minimum_version(ansible_src):
           """ % (ansible_version_requirement_pre_task_name,
                  min_ansible_version, min_ansible_version))
 
+    if ' - hosts: all' not in ansible_src:
```
cipherboy: So... We apparently build on older `oscap`s than 1.2.16. 

This appears related to this PR: https://github.com/OpenSCAP/openscap/pull/895

When doing so, we do not include the Ansible snippet, but otherwise do not update the content accordingly.

The `min_ansible_version` is 2.3, which was released ["Mid 2017"](https://docs.ansible.com/ansible/2.5/roadmap/ROADMAP_2_3.html) and tagged pApril 2017](https://github.com/ansible/ansible/releases/tag/v2.3.0.0-1). 

OpenSCAP 1.2.16 was released in [November 2017](https://github.com/OpenSCAP/openscap/releases/tag/1.2.16). 

I'd have thought it to be somewhat unlikely that OpenSCAP 1.2.16 is not included but Ansible 2.3 is... However, Ubuntu proves me wrong as always:

- [Ansible v2.5](https://packages.ubuntu.com/search?suite=bionic&section=all&arch=any&keywords=Ansible&searchon=names)
- [OpenSCAP v1.2.15](https://packages.ubuntu.com/search?keywords=OpenSCAP&searchon=names&suite=bionic&section=all)

So we might want to:

1. Poke the Ubuntu/Debian maintainers for OpenSCAP projects and get them to bump the versions.
2. Figure out what we want to do about minimum Ansible versions including in this.

Until then, I'm changing the `ValueError` to a `return ansible_src`.

tests/unit/ssg/test_ansible.py
```diff
@@ -2,50 +2,61 @@
...
+       tasks:
+    """
+    processed_snippet = textwrap.dedent(processed_snippet)
 
     output = ssg.ansible.add_minimum_version(good_snippet)
-    assert no_whitespace_compare(output, processed_snippet)
+    import sys
+    print(output, file=sys.stderr)
+    print(output, file=sys.stderr)
+    print(processed_snippet, file=sys.stderr)
```
matejak: Looks like leftover debug code.

cipherboy: Yep, good catch. :)

matejak:
Thanks for your updates, it is a pleasure to merge this PR.

PR #3215, opened 2018-08-21
Document existence of example product

cipherboy:
This adds:

 - A `README.md` to the Example product.
 - A message in CMake that the Example product is disabled, when it is disabled.
 - A line in the Developer documentation stating that these are examples for adding a new product + build system changes. 

This should resolve @yuumasato's comment on #3202.


dahaic:
@cipherboy did you forgot this PR exists? It's lonely :(

cipherboy:
@dahaic I think this should be suitably updated.

dahaic:
Dekuji @cipherboy , ACK.

PR #3196, opened 2018-08-14
Fix grub2 conf rules

iokomin:
#### Description:
Fixing grub2 conf permissions/ownership build warnings introduced in latest templates unification.
grub2 grub.cfg file path updated in corresponding rhel7, fedora and ol7 templates.

Corresponding build warnings:
`
WARNING: OVAL check 'file_owner_efi_grub2_cfg' was not found, removing <check-content> element from the XCCDF rule.
WARNING: OVAL check 'file_groupowner_efi_grub2_cfg' was not found, removing <check-content> element from the XCCDF rule.
WARNING: OVAL check 'file_groupowner_grub2_cfg' was not found, removing <check-content> element from the XCCDF rule.
WARNING: OVAL check 'file_owner_grub2_cfg' was not found, removing <check-content> element from the XCCDF rule.
WARNING: OVAL check 'file_permissions_grub2_cfg' was not found, removing <check-content> element from the XCCDF rule.
`
#### Testing:
- Checked generated oval definitions, bash and ansible remediation scripts content
- Checked non-efi grub2 rules to pass on OL7, checked bash remediation scripts to work
- Checked mentioned OVAL check warnings fixed in rhel7, ol7 and fedora builds

PR #3216, opened 2018-08-23
Add prodtype: rhel6 to more grub_legacy rules

cipherboy:

Per comments by @iokomin on #3196, there's a few `grub_legacy` rules which apply to all products. This fixes these and removes them from all other products but `rhel6`, which should be the only product we support which doesn't use grub2.


yuumasato:
Looks good, no other products select `grub_legacy` rules.

PR #3219, opened 2018-08-23
Rule dir tools tests

cipherboy:
A few tests against the functions in `ssg/rule_dir_stats.py`. Not against everything, but something to get started with. :)

### DIFF
--- a/tests/unit/ssg/test_rule_dir_stats.py
+++ b/tests/unit/ssg/test_rule_dir_stats.py
@@ -40,3 +40,149 @@ def test_missing_remediation():
     assert not rds.missing_remediation(good_rule, 'bash')
     assert rds.missing_remediation(bad_rule, 'bash')
     assert rds.missing_remediation(bad_rule, 'anaconda')
+
+
+def test_two_plus_oval():
+    three_rule = {
+        "id": "three_rule",
+        "ovals": {
+            "rhel6.xml": {},
+            "rhel7.xml": {},
+            "fedora.xml": {}
+        }
+    }
+    two_rule = {
+        "id": "two_rule",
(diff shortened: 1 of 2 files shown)


tests/unit/ssg/test_rule_dir_stats.py
```diff
@@ -0,0 +1,188 @@
...
+    one_rule = {
+        "id": "one_rule",
+        "ovals": {
+            "rhel6.xml": {},
+        }
+    }
+    empty_rule = {
+        "id": "bad_rule",
+        "ovals": {}
+    }
```
yuumasato: I have a feeling that it makes sense to start testing from `rule_dir_json.py`, and have all these rules in `data` dir.

cipherboy: Sure, I think this'll require a code restructuring though, as `rule_dir_json` doesn't currently use code in the `ssg` folder... So we're not guaranteeing order of execution of the unit tests for `utils` vs `ssg`... Which means we can't depend on `rule_dir_json.py` executing (and output existing) before these checks unless we call it ourselves. 

Thoughts?

yuumasato: @cipherboy I'm probably mixing things up, what I had in mind was more like integration tests. To load the rules from `data`dir and test the `stats` script.
As these are unit tests, there test rules are fine like this.

cipherboy: Ahh yes, integration tests would be good to have, I agree. I'll add them in a separate PR. :)

yuumasato:
Looks good, sorry for delay.

PR #3223, opened 2018-08-27
SSG Test Suite results analysis

matejak:
It features new SSG Test Suite ability - analysis of rule test results.

During a test run, each rule scenario test result is tracked. When the test run finishes, results are saved in a `results.json` file to the log directory.
A newly introduced script reads those files from various test runs and reports
* basic success/consistency statistics, and
* detailed reports of cases when a scenario ended up differently in different test runs.

Finally, the functionality got some unit tests, and the unit test system setup got streamlined.


matejak:
jenkins test this please

tests/CMakeLists.txt
```diff
@@ -2,29 +2,21 @@ if (ENABLE_PYTHON_COVERAGE)
     set(PYTEST_COVERAGE_OPTIONS --cov-append --cov-report=xml --cov "${CMAKE_SOURCE_DIR}/")
 endif()
 
-macro(ssg_python_unit_tests PYTHON_COMPONENT)
+macro(ssg_python_unit_tests PYTHON_COMPONENT_ID RELATIVE_PYTHONPATH)
     add_test(
-        NAME "python-unit-${PYTHON_COMPONENT}"
-        COMMAND "${PYTHON_EXECUTABLE}" -m pytest ${PYTEST_COVERAGE_OPTIONS} "${CMAKE_SOURCE_DIR}/tests/unit/${PYTHON_COMPONENT}"
+	    NAME "python-unit-${PYTHON_COMPONENT_ID}"
+	    COMMAND "${PYTHON_EXECUTABLE}" -m pytest ${PYTEST_COVERAGE_OPTIONS} "${CMAKE_SOURCE_DIR}/tests/unit/${PYTHON_COMPONENT_ID}"
```
mpreisler: indents are borked

matejak:
@mpreisler I have fixed the indent.

tests/README.md
```diff
@@ -205,3 +205,17 @@ You can do that by putting following lines in your `$HOME/.ssh/config` file:
...
 UserKnownHostsFile /dev/null
 ```
+
+## Analysis of results
+
+The rule tests results are saved as `results.json` into the corresponding log directory.
+You can then analyze those results by running e.g.
+
+```
+python analyze_results.py $(find . -name results.json)
```
mpreisler: `python3`?

tests/analyze_results.py
```diff
@@ -0,0 +1,129 @@
+import json
```
mpreisler: shebang please

tests/analyze_results.py
```diff
@@ -0,0 +1,129 @@
...
+import collections
+import functools
+import re
+import argparse
+
+
+@functools.total_ordering
+class RuleResult(object):
```
mpreisler: if we are using this for loading the json I'd suggest to also use it for creating it. Otherwise it will be difficult to keep both in sync.

tests/analyze_results.py
```diff
@@ -0,0 +1,129 @@
...
+
+def aggregate_results_by_scenarios(rule_results):
+    aggregated = collections.defaultdict(list)
+    for result in rule_results:
+        aggregated[result.scenario].append(result)
+    return aggregated
+
+
+def analyze_differences(rules):
+    rules = sorted(rules)
```
mpreisler: this is a bit cryptic and unexpected, we sort a list of rules and then only compare the lowest and highest. can you explain why? is rules expected to have more than 2 items?

tests/analyze_results.py
```diff
@@ -0,0 +1,129 @@
...
+    return parser.parse_args()
+
+
+def print_result_differences(json_results):
+    results = [json.load(open(fname, "r")) for fname in json_results]
+    rules = [RuleResult(r) for r in sum(results, [])]
+    ar = aggregate_results_by_scenarios(rules)
+
+    rules_that_ended_same = 0
+    rules_that_ended_by_success = 0
```
mpreisler: I think it would be great if you explained these two variables in a comment. It's not trivial to infer it from the code.

tests/analyze_results.py
```diff
@@ -0,0 +1,129 @@
...
+    differences = []
+    for rule_id, results in ar.items():
+        if len(results) < 2:
+            if results[0].success:
+                rules_that_ended_by_success += 1
+            # There won't be differences to analyze
+            continue
+        difference = analyze_differences(results)
+        if difference:
+            rule_stem = re.sub(".*content_rule_", "", rule_id[0])
```
mpreisler: shouldn't it be `^xccdf_.+` ...? also want to assert that it replaces at least something.

tests/README.md
matejak: Both python 2 and 3 are supported, so I would leave it like this.

matejak:
I have altered the PR according to the feedback. The `RuleResult` class has been moved to the `common` module, and there were other various minor improvements.

PR #3225, opened 2018-08-27
Drop anaconda fix for /dev/cdrom mount points

yuumasato:
#### Description:

- Stop generation Anaconda remediation for 
  - `mount_option_nodev_removable_partitions`
  - `mount_option_nosuid_removable_partitions`
  - `mount_option_noexec_removable_partitions`

#### Rationale:

- The existance of the fix itself, is causing `oscap-anaconda-addon` to require the
partition `/dev/cdrom` to exist. This should not the use case for this rule.

- [rhbz#1618840](https://bugzilla.redhat.com/show_bug.cgi?id=1618840)

### DIFF
--- a/shared/templates/csv/mount_options.csv
+++ b/shared/templates/csv/mount_options.csv
@@ -7,9 +7,9 @@
 #  If the remediation can create (i.e. not just modify) an /etc/fstab line,
 #  add the 'create_fstab_entry_if_needed' literal string as the third argument.
 
-var_removable_partition,nodev,create_fstab_entry_if_needed
-var_removable_partition,nosuid,create_fstab_entry_if_needed
-var_removable_partition,noexec,create_fstab_entry_if_needed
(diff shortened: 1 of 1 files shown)


shawndwells:
Would it be appropriate to update the remediation to detect if /dev/cdrom is present, then take action? Would that solve the anaconda issues?

yuumasato:
@shawndwells This detection capability is something that oscap-anaconda-addon should support, as the generated Anaconda snippet for this is 
```
part <sub idref="xccdf_org.ssgproject.content_value_var_removable_partition" use="legacy"/> --mountoptions="nodev"
```

redhatrises:
@yuumasato this is more of a problem during anaconda install correct, but not post install, correct?

yuumasato:
@redhatrises That is correct. The Bash remediation is generated and if there exists the `/dev/cdrom` mount, it will be remediated.

dahaic:
Thanks @yuumasato, there are no more strange part anaconda remediations in the datastream.

PR #3227, opened 2018-08-28
Remove minor versions from ubuntu content

redhatrises:
This removes minor versions from Ubuntu content and directory structure. Ubuntu has been out-of-line with the rest of the project for a while. Since we have gone through a bunch of content restructuring, this should be fixed now.


mpreisler:
I agree with @cipherboy , was about to write the same.

mpreisler:
I disagree, the Ubuntu versioning is strange but we are not going to change their versioning in SSG. So let's use it as it is and let's be precise.

Your suggestion is equivalent to having just one "rhel" content for all rhels AND fedora.

cipherboy:
Further, Ubuntu releases are much more predicable: there are two a year, in
04 and 10 (April and October). Even number years and April releases are
always LTSes. 14.04, 16.04, 18.04... And yes, as Martin said, most
enterprises run LTSes on servers and so will be scanning those. So we won't
have to support 14.06 or 14.08, always just two releases a year (which
minimally changes), or every two years, a new LTS.

If anything, I'd suggest that, rather than dropping the months from the
releases, I'd suggest dropping the entire version number and calling it
"Ubuntu" and using feature detections to support different content for
different versions. We can maybe see if we can get LTS content to apply up
through the next LTS (so 16.04 content applies up through 17.10 perhaps),
though we'd want feedback from the community on that. Perhaps we'd want to
go the other way (14.10->16.04)... I'd have to think about that.

That's my 2c though.

mpreisler:
The consensus here seems to be to not change the ubuntu versions. Therefore I am closing this to clean up the PR queue. Please reopen if you want to revisit this.

PR #3231, opened 2018-08-29
Initial OSPP42 Profile for Fedora

yuumasato:
#### Description:

- Ad OSPP42 Profile for Fedora (This mirrors RHEL7 selection of Rules)
- Necessary Rules are enabled for Fedora (prodtype += fedora)

#### Rationale:

- Aim is to establish initial Profile for collaboration on OVAL and Bash coverage.


redhatrises:
@yuumasato should we just make this profile called `ospp` since OSPP4.2 is the newest, and I *think* is now outside of draft form.

shawndwells:
Suggest not naming the profile OSPP as Fedora will not be going through Common Criteria. Don't want to create impression it holds the same security reviews and baselines of RHEL.

shawndwells:
@redhatrises yes, the draft was finalized... direct link @ https://www.niap-ccevs.org/MMO/PP/424.CANX/

yuumasato:
> Suggest not naming the profile OSPP as Fedora will not be going through Common Criteria.

Fedora itself might not, but can someone take Fedora and go through certification?
Is the product certified or an instance of the product certified? 

> Don't want to create impression it holds the same security reviews and baselines of RHEL.

Is it the case that a Profile for a distro named after a policy imply that the distro is certified?

Also, do you have some suggestion?

redhatrises:
@shawndwells wouldn't keeping `documentation_complete: false` satisfy your concerns as the profile wouldn't be included on builds?

dahaic:
Agreed about the filename / profile ID. `ospp` is sufficient, we should strive for not changing IDs too much. OSPP is general guidance, and as such does not imply OS is certified. For that reason, I would keep OSPP there, but change description appropriately. To something like:

"This profile reflects mandatory configuration controls identified in the NIAP Configuration Annex to the Protection Profile for General Purpose Operating Systems (Protection Profile Version 4.2).

As Fedora OS is moving target, this profile does not guarantee to provide security levels required from US National Security Systems. Main goal of the profile is to provide Fedora developers with hardened environment similar to the one mandated by US National Security Systems."

After this change, I would be for switching `documentation_complete: true` as we honestly want to provide this option to wider audience of Fedora users and developers.

yuumasato:
I have removed `draft` mention in Profile.
Added note about Fedora and provided security levels.
And changed ID to just `ospp`.

dahaic:
Thanks @yuumasato. I believe this won't mislead anybody, anymore.

dahaic:
@redhatrises sorry, I haven't noticed you assigned it to yourself. But it's not set in stone and we can update it if something is still wrong from your point of view.

PR #3232, opened 2018-08-30
Remove slash from audit login_events description

yuumasato:
#### Description:

- Remove slash from example faillock audit rule
- Update scenario for SSGTestSuite

#### Rationale:

- From audit PoV, it doesn't matter if the trailing slash is there or not. But our check is written so that the slash should no be there.
- Based on discussion in https://github.com/OpenSCAP/scap-security-guide/pull/2710 and confusion regarding recommended manual fix.


dahaic:
@yuumasato when you are touching this, can you rename `default.fail.sh` scenario to more meaningful `empty.fail.sh`? Default is bit ambiguous.

dahaic:
Thanks @yuumasato . Ack

PR #3233, opened 2018-08-30
Add password pam accounts rules to fedora

yuumasato:
#### Description:

- Fill up csv file for `accounts_password_pam_` rules
- Add test for `accounts_password_pam_ocredit`

#### Rationale:

- These rules apply for Fedora, lets just generate them all.

### DIFF
--- a/fedora/templates/csv/accounts_password.csv
+++ b/fedora/templates/csv/accounts_password.csv
@@ -4,7 +4,11 @@
 # - operation is the comparison operation to perform in OVAL check
 
 dcredit,less than or equal
+difok,greater than or equal
 lcredit,less than or equal
+maxclassrepeat,less than or equal
+maxrepeat,less than or equal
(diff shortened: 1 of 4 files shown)


yuumasato:
I plan to add tests for other rules, or work something out for the templated rules.
Some smart testing for templated rules...

dahaic:
@yuumasato so are we waiting for #3244 ? Or will you just make preliminary trivial test coverage, and proper one will be added later?

yuumasato:
> @yuumasato so are we waiting for #3244 ?

@dahaic I was waiting for it, but the author seems to be slacking off.

> Or will you just make preliminary trivial test coverage, and proper one will be added later?

We can reduce scope of this to just test `accounts_password_pam_ocredit`, and expand it later...

yuumasato:
I'll extend the test for other rules when template testing is introduced.

PR #3234, opened 2018-08-31
Add OVAL and remediation for auditd_audispd_configure_remote_server

jan-cerny:
#### Description:
This adds:
- OVAL
- bash remediation
- fixes of SSG TS tests

 for Rule auditd_audispd_configure_remote_server

The referenced variable is already defined in `linux_os/guide/system/auditing/configure_auditd_data_retention/var_audispd_remote_server.var`

#### Rationale:
This rule is part of Fedora OSPP profile.

### DIFF
--- a/linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_audispd_configure_remote_server/oval/shared.xml
+++ b/linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_audispd_configure_remote_server/oval/shared.xml
@@ -1,0 +1,34 @@
+<def-group>
+  <definition class="compliance" id="auditd_audispd_configure_remote_server" version="1">
+    <metadata>
+      <title>Configure audispd Plugin Remote Server IP address or Hostname</title>
+      <affected family="unix">
+        <platform>multi_platform_all</platform>
+      </affected>
+      <description>remote_server setting in /etc/audisp/audisp-remote.conf is set to a certain IP address or hostname</description>
+    </metadata>
+    <criteria>
+        <criterion comment="remote_server setting in /etc/audisp/audisp-remote.conf" test_ref="test_auditd_audispd_configure_remote_server" />
+    </criteria>
+  </definition>
+
+  <ind:textfilecontent54_test check="all" comment="remote server to send audit records" id="test_auditd_audispd_configure_remote_server" version="1">
+    <ind:object object_ref="object_auditd_audispd_configure_remote_server" />
+    <ind:state state_ref="state_auditd_audispd_configure_remote_server" />
+  </ind:textfilecontent54_test>
+
+  <ind:textfilecontent54_object id="object_auditd_audispd_configure_remote_server" version="1">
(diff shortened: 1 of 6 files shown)


linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_audispd_configure_remote_server/oval/shared.xml
```diff
@@ -0,0 +1,34 @@
...
+  </definition>
+
+  <ind:textfilecontent54_test check="all" comment="remote server to send audit records" id="test_auditd_audispd_configure_remote_server" version="1">
+    <ind:object object_ref="object_auditd_audispd_configure_remote_server" />
+    <ind:state state_ref="state_auditd_audispd_configure_remote_server" />
+  </ind:textfilecontent54_test>
+
+  <ind:textfilecontent54_object id="object_auditd_audispd_configure_remote_server" version="1">
+    <ind:filepath>/etc/audisp/audisp-remote.conf</ind:filepath>
+    <!-- Allow only space (exactly) as delimiter: https://fedorahosted.org/audit/browser/trunk/src/auditd-config.c#L426 -->
```
yuumasato: This link goes to fedorahosted retired page.

jan-cerny: Nice catch! It's there 12 times in multiple rules

jan-cerny: I would prefer to remove the link because the place in code can easily change over time and we don't have capacity to watch the changes in auditd sources.

jan-cerny: @yuumasato Do you like the solution that I removed the link? If yes, could I open a separate PR that removes the remaining occurrences from other OVALs?

tests/data/group_system/group_auditing/group_configure_auditd_data_retention/rule_auditd_audispd_configure_remote_server/remote_server_numeric_address.pass.sh
```diff
@@ -1,6 +0,0 @@
-#!/bin/bash
```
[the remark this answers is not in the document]
jan-cerny: Yes, a numeric address is valid in the config file, but the OVAL checks that `remote_server` equals to value specified in the XCCDF value. The OVAL check will not be satisfied by setting `remote_server`  to numeric address, therefore this test would fail.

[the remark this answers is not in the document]
yuumasato: Aha! I see, there is no value setting a numeric address, more over, no Profile selecting it.
Ok, we can live without this test. There is not much difference between dot separated strings and dot separated numbers...

PR #3235, opened 2018-08-31
Cleanup kickstart for SSGTestSuite domain

yuumasato:
#### Description:

- Remove build of openscap from repository
- Drop `Base` package group that doesn't exist in Fedora
- Document how to install domain for Fedora

#### Rationale:

- There is no need to build openscap from source, openscap-1.2.15 is packaged in supported distributions.
- Make it easier to install Fedora testing domain to test content developed for Fedora.


jan-cerny:
LGTM. Thank you.

PR #3236, opened 2018-08-31
Filter out swap files

jan-cerny:
This prevents SSG test suite from trying to run vim swap files.

### DIFF
--- a/tests/data/utils.py
+++ b/tests/data/utils.py
@@ -24,7 +24,9 @@ def iterate_over_rules():
         leaf_dir = os.path.basename(dir_name)
         rel_dir = '.' + dir_name[len(_DIR):]
         if leaf_dir.startswith('rule_'):
-            result = Rule(rel_dir, _SSG_PREFIX + leaf_dir, files)
+            # use only .sh files
+            scripts = filter(lambda x: x.endswith(".sh"), files)
+            result = Rule(rel_dir, _SSG_PREFIX + leaf_dir, scripts)
             yield result
 
 
--- a/tests/data/utils.py
+++ b/tests/data/utils.py
@@ -19,13 +19,13 @@ def iterate_over_rules():
             directory -- absolute path to the rule directory
             id -- full rule id as it is present in datastream
             files -- list of executable .sh files in the
-            directory containing the test cases in Bash
+            directory containing the test scenarios in Bash
     """
     for dir_name, directories, files in os.walk(_DIR):
         leaf_dir = os.path.basename(dir_name)
         rel_dir = '.' + dir_name[len(_DIR):]
         if leaf_dir.startswith('rule_'):
-            # Filter out everything except the shell test cases.
+            # Filter out everything except the shell test scenarios.
             # Other files in rule directories are editor swap files
             # or other content than a test case.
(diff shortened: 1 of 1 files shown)


tests/data/utils.py
```diff
@@ -24,7 +24,9 @@ def iterate_over_rules():
         leaf_dir = os.path.basename(dir_name)
         rel_dir = '.' + dir_name[len(_DIR):]
         if leaf_dir.startswith('rule_'):
-            result = Rule(rel_dir, _SSG_PREFIX + leaf_dir, files)
+            # filter out swap files
+            scripts = filter(lambda x: not x.endswith("swp"), files)
```
yuumasato: I would like ask to exclude vim backup files as well, their names ends with `~`.

cipherboy: Do we want to just whitelist `.sh` scripts? I think that'd make more sense to me personally, unless the few files not ending in `.sh` (like `group_system/group_auditing/group_auditd_configure_rules/group_audit_privileged_commands/rule_audit_rules_privileged_commands/rhel7_privileged.rules`) need to be executed, but still, I'd rather have a whitelist rather than a blacklist; seems less fragile.

jan-cerny: aha, ok

mpreisler: I also strongly prefer whitelist

jan-cerny: aha

tests/data/utils.py
```diff
@@ -24,7 +24,9 @@ def iterate_over_rules():
         leaf_dir = os.path.basename(dir_name)
         rel_dir = '.' + dir_name[len(_DIR):]
         if leaf_dir.startswith('rule_'):
-            result = Rule(rel_dir, _SSG_PREFIX + leaf_dir, files)
+            # use only .sh files
```
mpreisler: This comment doesn't add much to the line that comes after it. I would suggest to explain why we only care about .sh files and nothing else. What are we skipping? What do the .sh files contain?

tests/data/utils.py
```diff
@@ -24,7 +24,14 @@ def iterate_over_rules():
         leaf_dir = os.path.basename(dir_name)
         rel_dir = '.' + dir_name[len(_DIR):]
         if leaf_dir.startswith('rule_'):
-            result = Rule(rel_dir, _SSG_PREFIX + leaf_dir, files)
+            # Use only .sh files
+            # We care only about '.sh' files because only these files
+            # are supposed to contain bash scripts with test cases
+            # and we can expect them to be executable.
+            # Other files in rule directories are editor swap files
+            # or other content than a test case.
```
mpreisler: I don't want to be nitpicky but if it was me I would change the "files" description in the doctext to say that it's a list of executable test-cases, maybe even rename it from files to testcases.

The comment itself is way too verbose, it could just say "Filter out everything except the shell test-cases". All the extra info you have in the comment should be in the doctext.

dahaic:
@jan-cerny sorry for continued nitpick - I tried to establish nomenclature of calling the test-triggering files "scenarios". Would it be possible to update it once more to use this? Otherwise looks good to me.

jan-cerny:
@dahaic If it's you it's always possible.

PR #3237, opened 2018-08-31
Fix running of SSG TS

jan-cerny:
#### Description:

Install tar package on SSG TS VM

Do not ignore failures of external commands

#### Rationale:

On Fedora 28 Server the 'tar' package is not installed by default
which means that SSG test suite is not able to extract the archive
with test cases.

The CalledProcessError exception was caught and the return value of
this function was ignored. That meant if some command went wrong
(eg. uploading the tarball to the VM was not successful or extraction
of the tarball was not successful) it was ignored and the test suite
continued happily, without letting the user know. We need to terminate
if some of the test harness setup tasks went wrong.


dahaic:
Looks good, thanks @jan-cerny

PR #3238, opened 2018-08-31
SSG Test Suite manual debug

dahaic:
#### Description:

- Added option to pause execution of the test suite when unexpected result is encountered. This happens before test environment is reverted, so tester can take a look at the problem manually.

#### Rationale:

- Quite often the development and debugging of test scenarios is strange - something fails, but because environment is created and then destroyed, it's not 100 % clear whether the problem is in the scenario or where. With this, you can connect to the VM / Container and compare your expectations with reality. And then continue with evaluation of another scenario.


tests/ssg_test_suite/oscap.py
```diff
@@ -273,6 +274,11 @@ def _generate_report_file(self):
         ])
         self._filenames_to_clean_afterwards.add(self.report_path)
 
+    def _wait_for_continue(self):
+        """ In case user requests to leave machine in failed state for hands
+        on debugging, ask for keypress to continue."""
+        raw_input("Paused for manual debugging. Continue by pressing return.")
```
mpreisler: NACK, `NameError: name 'raw_input' is not defined`

tests/test_suite.py
```diff
@@ -109,6 +108,12 @@ def parse_args():
                                    "ask for subset of all rules this way. (If you "
                                    "type ipv6 as a target, all rules containing "
                                    "ipv6 within id will be performed."))
+    parser_rule.add_argument("--manualdebug",
```
mpreisler: is there any other debug than "manual" debug? If not, why have the extra adjective there?

tests/test_suite.py
```diff
@@ -109,6 +108,12 @@ def parse_args():
...
+                             dest="manual_debug",
+                             action="store_true",
+                             help=("Stop execution when error is encountered, so "
+                                   "developer can connect to VM / container with "
+                                   "scenario set up"))
```
mpreisler: you use 2 different quote styles in 2 hunks in a single patch

mpreisler: alternative help:
```
If an error is encountered, all execution on the VM / container will pause to allow debugging.
```

tests/ssg_test_suite/oscap.py
cipherboy: In `ssg` module, we have a wrapper for this, `ssg.shims.input_func`: https://github.com/OpenSCAP/scap-security-guide/blob/master/ssg/shims.py#L38-L42

tests/test_suite.py
dahaic: You are right, it's not necessary to be public facing. I have left manual_debug variable in the code though. I feel it's more clear than having only `debug`.

tests/test_suite.py
dahaic: Good notes.

tests/ssg_test_suite/oscap.py
dahaic: Thanks @cipherboy

dahaic:
@jan-cerny @yuumasato review/merge pls :)

jan-cerny:
It works for me. I have tested it on tests for rule_accounts_tmout on Fedora OSPP profile which do not have OVAL yet.

Issue #612, opened 2015-07-17
Check the fedora crypto policies


nmav:
Fedora provides a [crypto policy](https://fedoraproject.org/wiki/Changes/CryptoPolicy) which is respected distribution-wide. It would be good to enhance the scanner to check and verify that the crypto policies set are applied, reasonable and known to be secure.

shawndwells:
IIRC, the check would review `````/etc/crypto-policies/config````` for the `````{LEGACY DEFAULT FUTURE}````` values. The profile would determine what values are acceptable (e.g. DoD STIG may need FUTURE). 

Do I understand correctly?

nmav:
I believe you are correct. I think there two Some checks to perform are:
* Whether the set policy has been applied (e.g, whether the generated files have newer creation date than the policy - to make this check reliable I'll have to make update-crypto-policies transactional)
* Whether the policy set is reasonable (I guess the definition of reasonable would have to be policy-specific)

PR #3240, opened 2018-08-31
Add initial crypto content to Fedora

redhatrises:
#### Description:

- Adds initial crypto content to Fedora. There is still work to do, but this handles non-custom configurations.
- Part of #612


linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/oval/fedora.xml
```diff
@@ -0,0 +1,37 @@
+<def-group>
+  <definition class="compliance" id="configure_crypto_policy" version="1">
+    <metadata>
+      <title>Configure System Cryptographic Policies</title>
+      <affected family="unix">
+        <platform>multi_platform_fedora</platform>
+      </affected>
+      <description>Look for cipher policy correctly configured in /etc/crypto-policy/config.</description>
```
yuumasato: I would say `crypto policy`, instead of `cipher policy`, as it defines more than just ciphers.
There are other mentions of `cipher policy` below too.

linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/oval/fedora.xml
```diff
@@ -0,0 +1,37 @@
...
+    </metadata>
+    <criteria>
+      <criterion comment="check for cipher policy correctly configured in /etc/crypto-policy/config"
+      test_ref="test_configure_crypto_policy" />
+    </criteria>
+  </definition>
+
+  <ind:textfilecontent54_test id="test_configure_crypto_policy"
+  comment="check for cipher policy correctly configured in /etc/crypto-policy/config"
+  check="all" check_existence="all_exist" version="1">
```
yuumasato: Maybe `check_existence` should be `only_one_exists`?

linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/oval/fedora.xml
```diff
@@ -0,0 +1,37 @@
...
+    <metadata>
+      <title>Configure System Cryptographic Policies</title>
+      <affected family="unix">
+        <platform>multi_platform_fedora</platform>
+      </affected>
+      <description>Look for cipher policy correctly configured in /etc/crypto-policy/config.</description>
+    </metadata>
+    <criteria>
+      <criterion comment="check for cipher policy correctly configured in /etc/crypto-policy/config"
+      test_ref="test_configure_crypto_policy" />
```
yuumasato: I think that time stamp of the config file needs to be checked as well. Editing the file manually without running `update-crypto-policies` causes a false-negative.

From comments in `/etc/crypto-policies/config`:
```
# After modifying this file, you need to run update-crypto-policies
# for the changes to propagate.
```

To check whether crypto-policy is applied, verify time stamp of files /etc/crypto-policies/state/current and /etc/crypto-policies/config. If timestamp of the former is newer than the latter, the policy is applied.

linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/rule.yml
```diff
@@ -0,0 +1,25 @@
...
+
+title: 'Configure System Cryptography Policy'
+
+description: |-
+    To configure the system cyptography policy to use ciphers only from the <tt><sub idref="var_system_crypto_policy" /></tt>
+    policy, run the following command:
+    <pre>$ sudo update-crypto-policies --set <sub idref="var_system_crypto_policy" /></pre>
+
+rationale: |-
+    Centralized cryptographic polices simply applying secure ciphers across an operating system and
```
Jakuje: s/simply/simplifies/

linux_os/guide/system/software/integrity/crypto/var_system_crypto_policy.var
```diff
@@ -0,0 +1,17 @@
...
+type: string
+
+operator: equals
+
+interactive: false
+
+options:
+    default: DEFAULT
+    legacy: LEGACY
+    future: FUTURE
```
Jakuje: There is FIPS policy as well. Not sure if it is already released, but it will be there sooner or later:
https://gitlab.com/redhat-crypto/fedora-crypto-policies/blob/master/back-ends/profiles/FIPS.pl

t8m: It is there.

linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/rule.yml
t8m: Centralized cryptography policy simplifies ...

linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/oval/fedora.xml
redhatrises: @yuumasato yes... this PR does not fully implement everything. This will be done in a follow up PR.

linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/rule.yml
redhatrises: actually should be `policies simplify`

linux_os/guide/system/software/integrity/crypto/var_system_crypto_policy.var
redhatrises: Added `FIPS` and `NEXT`

linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/oval/fedora.xml
yuumasato: Fine.

PR #3241, opened 2018-09-03
Add OVAL and Bash for auditd_audispd_encrypt_sent_records

jan-cerny:
#### Description:
This PR:
* adds OVAL
* adds Bash remediation
* adds OSPP profile to the test cases

#### Rationale:


shawndwells:
@mildass has to accept the invite into the new org. Should show up as being able to own tickets once accepted.

PR #3242, opened 2018-09-03
Fix remediation testing

dahaic:
#### Description:

- Fixed testing of remediation and final phases, and removed final phase in case of oscap remediation, as it it superfluous


yuumasato:
Looks good, I see remediation scan run in logs.
Fail in Travis is unrelated, marshal error in CentOS6.

PR #3250, opened 2018-09-03
Enable rsyslog_remote_loghost OVAL on Fedora

mpreisler:
Bash and Ansible already being there this seems like it's just an oversight that it wasn't already enabled.


jan-cerny:
I like the way how it's done very much.

Issue #1727, opened 2017-02-27
gpgcheck Enabled In Main Dnf Configuration


ybznek:
On my Fedora system `/etc/yum.conf` and `/etc/dnf/dnf.conf` exist.

If at least one of that config files contains `gpgcheck=1`, SSG rule will pass.

expected state:
rule will fail, if at least on of that files doesn't contain `gpgcheck=1`

matejak:
Update: Both files exist on Fedora 29, and they may contain repository definitions. Those repositories may enable or disable `gpgcheck`. The check should also go through other files where repositories are defined, and check for gpgcheck status there as well.

JaskaranNarula:
Question: @yuumasato @ybznek  does this stuff needs to be written in bash or xml ? 
Also, if bash then do I have to consider epel as well as in the rule ?

yuumasato:
@JaskaranNarula This issue was about updating the OVAL (xml) of [ensure_gpgcheck_globally_activated](https://github.com/ComplianceAsCode/content/tree/master/linux_os/guide/system/software/updating/ensure_gpgcheck_globally_activated) to check both `/etc/yum.conf` and '/etc/dnf/dnf.conf`, and  Bash remediation should follow as well.

But on F31 the original issue doesn't seem to manifest anymore, it contains only `/etc/dnf/dnf.conf` file, so I think this can be closed.

marcusburghardt:
No longer an issue.

PR #3251, opened 2018-09-04
Use implied properties to parametrize package manager configuration settings rules

jan-cerny:
#### Description:

Some rules describe package manager configuration. However, on RHEL we use `yum` and we set configuration settings in `/etc/yum.conf`, but on Fedora we use `dnf`and we set configuration in `/etc/dnf/dnf.conf`.

We can use implied properties to customize the rules and remediations. These variables will be substituted in final output, so there will be "dnf" mentioned in Fedora content and "yum" mentioned in RHEL content. Also, the rules and remediations will contain correct path to the configuration file depending on the package manager used on target platform.

Since this is a radical change, I have used the new approach only in rule `ensure_gpgcheck_globally_activated`. If the feedback on this change will be positive, then I can do a similar changes in all other rules from this group.

#### Rationale:

This will prevent us from copying code.

### DIFF
--- a/docs/manual/developer_guide.adoc
+++ b/docs/manual/developer_guide.adoc
@@ -632,6 +632,12 @@ Sometimes, a rule requires `ocil` and `ocil_clause` to be specified, and they de
 Macros that begin with `complete_ocil_entry_` were designed for exactly this purpose, as they make sure that OCIL and OCIL clauses are defined and consistent.
 Macros that begin with underscores are not meant to be used in descriptions.
 
+To parametrize rules and remediations as well as Jinja macros, you can use product-specific variables defined in `product.yml` in product root directory.
+Moreover, you can define *implied properties* which are variables inferred from them.
+For example, you can define a condition that checks if the system uses `yum` or `dnf` as a package manager and based on that populate a variable containing correct path to the configuration file.
+The inferring logic is implemented in `_get_implied_properties` in `ssg/yaml.py`.
+Constants and mappings used in implied properties should be defined in `ssg/constants.py`.
+
 Rules are unselected by default - even if the scanner reads rule definitions, they are effectively ignored during the scan or remediation.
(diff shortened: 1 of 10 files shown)


yuumasato:
How about setting `pkg_manager_config_file` as an [implied property](https://github.com/OpenSCAP/scap-security-guide/blob/master/ssg/yaml.py#L56)?

I think these implied properties will need more visibility as we put more stuff on it.

jan-cerny:
It's great that we already a mechanism for that. Thanks for showing it.

jan-cerny:
@pep8speaks I don't change this line.

yuumasato:
@jan-cerny looks good to me.
But I wonder if list of implied properties should be listed somewhere in documentation, this can be done later though.

jan-cerny:
Do you think it should be done also for rules that are not included in any Fedora profile?

redhatrises:
Fixes #1727

yuumasato:
@jan-cerny I would limit the scope of this Pr just for Fedora, for now.
Is there anything else keeping this as WIP?

jan-cerny:
I haven't discovered other occurrences of `/etc/yum.conf` in Fedora rules.

There are some rules that suggest in their description to run `yum install` or `yum erase`. Those things should be replaced by Jinja macros. We have already defined them in our .jinja files.  I think that should be done in a separate PR.

yuumasato:
> I haven't discovered other occurrences of /etc/yum.conf in Fedora rules.

Thanks for checking. Ack.

> think that should be done in a separate PR.

Agreed.

mpreisler:
> @jan-cerny I would limit the scope of this Pr just for Fedora, for now.
> Is there anything else keeping this as WIP?

If it was me I would do it for all rules. The reason is simple, once I know the issue and I know how to fix it I can very efficiently fix it for all and save future cost. The cost of figuring out what to fix and picking it up is very high, the cost of running a `sed` command for all rules is very low.

Just my 2c.

PR #3253, opened 2018-09-04
Rule sshd_enable_warning_banner for Fedora

jan-cerny:
#### Description:

1. Extend OVAL and Bash for sshd_enable_warning_banner to all platforms
Unfortunately, I have discovered that we can't use replace_or_append functions if the value of key cotains a forward slash. If we don't escape the slash, the "replace" mode is broken, if we escape the slash, the "append" mode is broken.

2. Add tests for sshd_enable_warning_banner

#### Rationale:
This rule is a part of Fedora OSPP profile


dahaic:
Jenkins test this, please.

dahaic:
Ack, thanks @jan-cerny !

PR #3254, opened 2018-09-04
Removed obsolete and duplicated README files

mpreisler:
We no longer keep those per product, instead we have developer docs.


redhatrises:
Ack

PR #3255, opened 2018-09-04
These files are not used and can be confusing to contributors

mpreisler:
They are a remnant of the old XML+XSLT system of building security
content.


redhatrises:
YES!!!!! ACK

PR #3256, opened 2018-09-05
Rule  service_debug-shell_disabled for Fedora

jan-cerny:
#### Description:

1. Add OVAL for service_debug-shell_disabled on Fedora

This rule is templated. We have to add `systemd` to a list of
removed packages because debug-shell is provided by `systemd`
package. If we forget about it, the respective OVAL definition
will be removed from the final datastream by our build system.

2.  Add tests for rule service_debug-shell_disabled 

#### Rationale:

### DIFF
--- a/fedora/templates/csv/packages_removed.csv
+++ b/fedora/templates/csv/packages_removed.csv
@@ -5,3 +5,4 @@ prelink
 setroubleshoot
 xorg-x11-server-common
 sssd
+systemd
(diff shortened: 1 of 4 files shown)


fedora/templates/csv/packages_removed.csv
```diff
@@ -5,3 +5,4 @@ prelink
 setroubleshoot
 xorg-x11-server-common
 sssd
+systemd
```
dahaic: @jan-cerny are you sure you want to remove `systemd` from the system?

jan-cerny: @dahaic Yes, I am. :smiley_cat: 

It's consistent with what we do in RHEL 7.  have to add systemd to a list of
removed packages because debug-shell is provided by systemd package. If we forget about it, the respective OVAL definition will be removed from the final datastream by our build system. Check how does the OVAL template for service_.*disabled_rules look like.

dahaic: Ah, makes sense! :)

PR #3258, opened 2018-09-05
Add Fedora to rule accounts_tmout

jan-cerny:
#### Description:

This adds platform element for Fedora on OVAL and remediation for rule accounts_tmout

The refine-value for var_accounts_tmout in Fedora OSPP profile is
set consistently with RHEL 7 OSPP profile.

#### Rationale:


redhatrises:
Ack

PR #3257, opened 2018-09-05
Add selinux_confinement_of_daemons for Fedora

jan-cerny:
#### Description:

Adds `multi_platform_fedora` to OVAL for rule selinux_confinement_of_daemons.

#### Rationale:
This rule is a part of Fedora OSPP profile.

PR #3260, opened 2018-09-05
Enabled selinux_all_devicefiles_labeled OVAL check for Fedora

mpreisler:
Pretty much the same as #3257


redhatrises:
Ack when jenkins finishes.

PR #3261, opened 2018-09-05
Dconf gnome fixes fedora

mpreisler:
For the most part this is just mechanical to enable it on Fedora. Ansible fixes are not all there but I wasn't focusing on that for this PR because right now that's out of scope of what I want to achieve.


redhatrises:
ACk

PR #3264, opened 2018-09-06
Audit auid!=unset and auid!=4294967295 are equivalent

yuumasato:
#### Description:

- Update OVAL checks for audit rules to accept `unset` as equivalent of `4294967295`.

#### Rationale:

- Audit `auid!=unset` and `auid!=4294967295` are equivalent.

### DIFF
--- a/linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands/oval/shared.xml
+++ b/linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands/oval/shared.xml
@@ -75,7 +75,7 @@
   </ind:textfilecontent54_test>
   <ind:textfilecontent54_object id="object_arpc_suid_sgid_augenrules" version="1">
     <ind:filepath operation="pattern match">^/etc/audit/rules\.d/.*\.rules</ind:filepath>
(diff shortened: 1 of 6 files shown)


yuumasato:
> Conversely the upstream audit.rules still has auid!=unset... perhaps because they can't ensure audit=1?

I guess then, that all audit rules would have to be associated with a check for kernel argument `audit=1`

> Also, should the XCCDF be updated to reflect unset vs 4294967295?

@shawndwells I can do that in following PRs, what do you think?

shawndwells:
On 9/6/18 8:40 AM, Watson Yuuma Sato wrote:
>
>     Conversely the upstream audit.rules still has auid!=unset...
>     perhaps because they can't ensure audit=1?
>
> I guess then, that all audit rules would have to be associated with a 
> check for kernel argument |audit=1|
>

Thinking through this more, while it could be more efficient in *code* 
to drop auid, may have negative impacts with downstream content. For 
example what are the odds DISA would go into the OVAL and notice the 
dependency on audit=1? Wouldn't be immediately apparent and could be missed.


>     Also, should the XCCDF be updated to reflect unset vs 4294967295?
>
> @shawndwells <https://github.com/shawndwells> I can do that in 
> following PRs, what do you think?
>
That'd be great!

PR #3271, opened 2018-09-07
Fix umask_for_daemons by specifying profile set umask

### LINKED ISSUE #2751: acceptable umask showing as "fail"
#### Description of problem:
When scanning RHEL 6 against the stig-rhel6-disa profile, I get a failure for a compliant option:

Var ref | Value
-- | --
oval:ssg-var_etc_init_d_functions_umask_as_number:var:1 | 18

The xccdf says: `To set the default umask for daemons, edit the following line, inserting 022 or 027 for umask appropriately`

0.1.38

RHEL 6.9

#### Steps to Reproduce:

1. set umask in /etc/init.d/functions to 022
2. run scan
3. see failure

#### Actual Results:
failure

#### Expected Results:
pass

Seems that the oval doesn't support multiple values.  Is see:
```
rhel6/profiles/C2S.profile:    - var_umask_for_daemons=027
```

redhatrises:
- Fixes #2751


linux_os/guide/system/permissions/restrictions/daemon_umask/umask_for_daemons/rule.yml
```diff
@@ -8,8 +8,7 @@ description: |-
     The file <tt>/etc/init.d/functions</tt> includes initialization
     parameters for most or all daemons started at boot time.  The default umask of
     022 prevents creation of group- or world-writable files.  To set the default
```
yuumasato: If `var_umask_for_daemons` is tailored to `027`, this snippet would look out of place `The default umask of 022 prevents creation of group- or world-writable files. `

redhatrises: @yuumasato I think it is important to keep it as many people get confused by `umask`, so I refactored the script to hopefully make more sense now as well as highlight the affects of changing the umask.

PR #3272, opened 2018-09-07
Update Developer Guide

jan-cerny:
#### Description:

* Add a Legacy notice that explains why "ssg" can be seen in the repository.
* Add a short introduction
* Update mentions of SSG

#### Rationale:
Since this project has moved to ComplianceAsCode, some things have been renamed, but some of them still keep the SSG in their name or contents.

### DIFF
--- a/docs/manual/developer_guide.adoc
+++ b/docs/manual/developer_guide.adoc
@@ -10,7 +10,7 @@ toc::[]
 
 == Introduction
 
-This document tries provides information useful for ComplianceAsCode/content project developers.
+This document tries provides information useful for ComplianceAsCode/content project contributors.
 We will guide you through the structure of the project.
 We will explain the directory layout, used formats and the build system.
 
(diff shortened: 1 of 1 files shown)


docs/manual/developer_guide.adoc
```diff
@@ -1,4 +1,4 @@
-= SCAP Security Guide Developer Guide
+= ComplianceAsCode Developer Guide
```
mpreisler: wait, ComplianceAsCode is the org, not the project, right?

jan-cerny: I think we can put it here as well.

jan-cerny:
I will remove the reanming changes from this PR. I will keep here only the legacy notice, which was my action item.

jan-cerny:
I have updated this PR by removing the renaming. The renaming will be done in separate PR after the final name will be clear.

docs/manual/developer_guide.adoc
```diff
@@ -10,6 +10,10 @@ toc::[]
 
 == Introduction
 
+This document tries provides information useful for ComplianceAsCode/content project developers.
```
dahaic: Slight reword: This document provides information useful for `ComplianceAsCode/content` project contributors.

docs/manual/developer_guide.adoc
```diff
@@ -935,3 +939,26 @@ The SSG build and templating system is mostly written in Python.
 * The common pattern is to dynamically add the `shared/modules` to the import path. The `ssgcommon` module has many useful utility functions and predefined constants. See scripts at `./build-scripts` as an example of this practice.
 * Follow the link:https://www.python.org/dev/peps/pep-0008/[PEP8 standard].
 * Try to keep most of your lines length under 80 characters. Although the 99 character limit is within link:https://www.python.org/dev/peps/pep-0008/#maximum-line-length[PEP8 requirements], there is no reason for most lines to be that long.
+
+== Legacy Notice
+
+This project has been created by renaming SCAP Security Guide Project (SSG).
```
dahaic: I would condense this explanation paragraph. Project outgrown former name SCAP Security Guide, and changes name to imply broader scope than just SCAP.

Rest is filler that would be better removed, imo.

docs/manual/developer_guide.adoc
```diff
@@ -935,3 +939,26 @@ The SSG build and templating system is mostly written in Python.
...
+security content.
+The word SCAP in the project name started to limit our expansion, because people were afraid
+that SCAP is overkill for their needs.
+Therefore, the SCAP Security Guide has been transformed into ComplianceAsCode/content, which better
+described the aim of the project.
+
+This git repository was created by simply renaming and moving the SCAP Security Guide (SSG)
+repository to a different GitHub organization.
+
+However, due to this history, the repository contains many mentions of SCAP Security Guide.
```
dahaic: Not however - just Due to this...

docs/manual/developer_guide.adoc
```diff
@@ -935,3 +939,26 @@ The SSG build and templating system is mostly written in Python.
...
+These legacy mentions are continuously cleaned up as the project develops.
+Some of them are kept due to backwards compatibility.
+
+For example, the output files produced by our build system still start by `ssg-` prefix.
+The reason for this is that the various Linux distributions package ship these files
```
dahaic: Remove "the reason for this is that the".

Also - do we actually plan to keep it like that for backwards compatibility? I believe that's problem of downstream packagers, and we don't have to care too much (when it's one time renaming effort, I mean). File names needs to make sense, that's whole requirement I see.

dahaic:
Thank you @jan-cerny. ACK

PR #3275, opened 2018-09-07
Replace occurences of yum in rules

jan-cerny:
#### Description:
Replace occurrences of yum in rules by Jinja macros

#### Rationale:
We should use already defined Jinja macros to add package-related
commands instead of hard coding them. This will enable to easily
include these rules into Fedora and other products
that use different package managers.


jan-cerny:
We will need to do a similar thing for remediations and for OVALs.

PR #3276, opened 2018-09-07
Update of Test Suite

dahaic:
#### Description:

- Fixup of failing profile mode, which was blocking functionality tests of created VM.
- VM generating script.
- Update of generated VM to have also possibly expected partitions.

#### Rationale:

- Having commands in README are fine, but script which will do it is better :)

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -28,25 +28,17 @@ ssh on the libvirt domain and have Ansible installed on the host machine.
 ### Domain preparation
 
 
-1. Install domain, using `kickstarts/rhel_centos_7.cfg`
+1. Install domain, using provided kickstart
 
    To install the domain via graphical install, supply boot option:  
-   `inst.ks=https://raw.githubusercontent.com/OpenSCAP/scap-security-guide/master/tests/kickstarts/rhel_centos_7.cfg`.  
+   `inst.ks=https://raw.githubusercontent.com/ComplianceAsCode/content/master/tests/kickstarts/rhel_centos_7.cfg`.
    And the installation will be guided by the kickstart.
 
-   To install the domain via command line, see example command below:  
-   1. To install a CentOS VM:
-    ```bash
-    virt-install -n ssg-test-suite-centos -r 2048 --vcpus=2 \
-    --os-variant=rhel7 --accelerate \
-    --disk path=/var/lib/libvirt/images/ssg-test-suite-centos.qcow,size=12 \
-    -x "inst.ks=https://raw.githubusercontent.com/OpenSCAP/scap-security-guide/master/tests/kickstarts/rhel_centos_7.cfg" \
-    --location http://mirror.centos.org/centos/7/os/x86_64
-    ```
-   2. To setup a Fedora machine, change the following parameters:  
-     `--os-variant=fedora28`  
-     `--location https://dl.fedoraproject.org/pub/fedora/linux/releases/28/Everything/x86_64/os`  
-     NOTE: The Everything compose is necessary to be able to install `openscap-scanner` during kickstart installation.
+   To install the domain via command line, you can use script `./install_vm.py`. Common usage will look like:
+   ```
+   ./install_vm.py --domain test-suite-fedora --distro fedora
+   ```
+
 1. Import ssh key to the machine and make sure that you can use them to log in
    as the `root` superuser.
 1. In case of installing a RHEL domain, setup subscription so that the machine can install and uninstall packages.
--- a/tests/README.md
+++ b/tests/README.md
@@ -37,9 +37,9 @@ ssh on the libvirt domain and have Ansible installed on the host machine.
    To install the domain via command line, see example command below:  
    1. To install a CentOS VM:
     ```bash
-    virt-install -n ssg-test-suite-centos -r 4096 --vcpus=4 \
+    virt-install -n ssg-test-suite-centos -r 2048 --vcpus=2 \
     --os-variant=rhel7 --accelerate \
-    --disk path=/var/lib/libvirt/images/ssg-test-suite-centos.qcow,size=50 \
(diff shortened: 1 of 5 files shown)


tests/install_vm.py
```diff
@@ -0,0 +1,64 @@
...
+    data['url'] = "https://download.fedoraproject.org/pub/fedora/linux/releases/28/Everything/x86_64/os"
+elif parser.distro == "centos7":
+    data['variant'] = "centos7"
+    data['url'] = "http://mirror.centos.org/centos/7/os/x86_64"
+
+command = 'virt-install -n {domain} -r {ram} --vcpus={cpu} --os-variant={variant} --accelerate --disk path={disk_path},size=12 -x "inst.ks={kickstart}" --location {url}'.format(**data)
+if parser.dry:
+    print(command)
+else:
+    os.system(command)
```
mpreisler: I suggest wrapping main in `def main():`, it's a nice paradigm that let's you import the python file without executing it. We use it everywhere else.

tests/install_vm.py
```diff
@@ -0,0 +1,64 @@
...
+    help="Number of CPU cores configured for the VM."
+)
+parser.add_argument(
+    "--dry",
+    dest="dry",
+    action="store_true",
+    help="Print command line instead of triggering command."
+)
+
+parser = parser.parse_args()
```
mpreisler: this makes me weep

mpreisler: please do `args = parser.parse_args()`, avoid `parser.__dict__`, that's a nasty workaround. Make this legible please.

tests/install_vm.py
dahaic: I agree

tests/install_vm.py
dahaic: Eh, yeah it's actually stupid right. Will fix. Unfortunately the `__dict__` is not so easily avoidable. It won't be accepted by the format function, afaict. But it might be more readable to add rest into `args` and call `__dict__` at the very end. What do you think?

dahaic: @mpreisler changed it - it's still not perfect, but I want to avoid need to be explicit when formatting the string. It wouldn't help readability.

tests/README.md
```diff
@@ -28,25 +28,17 @@ ssh on the libvirt domain and have Ansible installed on the host machine.
 ### Domain preparation
 
 
-1. Install domain, using `kickstarts/rhel_centos_7.cfg`
+1. Install domain, using provided kickstart
 
    To install the domain via graphical install, supply boot option:  
    `inst.ks=https://raw.githubusercontent.com/OpenSCAP/scap-security-guide/master/tests/kickstarts/rhel_centos_7.cfg`.
```
jan-cerny: This link probably should be updated as well.

tests/README.md
```diff
@@ -28,25 +28,17 @@ ssh on the libvirt domain and have Ansible installed on the host machine.
...
-    -x "inst.ks=https://raw.githubusercontent.com/OpenSCAP/scap-security-guide/master/tests/kickstarts/rhel_centos_7.cfg" \
-    --location http://mirror.centos.org/centos/7/os/x86_64
-    ```
-   2. To setup a Fedora machine, change the following parameters:  
-     `--os-variant=fedora28`  
-     `--location https://dl.fedoraproject.org/pub/fedora/linux/releases/28/Everything/x86_64/os`  
-     NOTE: The Everything compose is necessary to be able to install `openscap-scanner` during kickstart installation.
+   To install the domain via command line, you can use script provided:
+   ```
+   ./install_vm.py --help
```
jan-cerny: If I run this command, it only shows a help and does not install anything.

tests/install_vm.py
```diff
@@ -0,0 +1,69 @@
...
+        dest="dry",
+        action="store_true",
+        help="Print command line instead of triggering command."
+    )
+
+    return parser.parse_args()
+
+def main():
+    data = parse_args()
+    data.kickstart = "https://raw.githubusercontent.com/OpenSCAP/scap-security-guide/master/tests/kickstarts/rhel_centos_7.cfg"
```
jan-cerny: We can use the new repository URL.

tests/install_vm.py
```diff
@@ -0,0 +1,69 @@
...
+    data.disk_path = os.path.join(data.disk_dir, data.domain) + ".qcow"
+
+    if data.distro == "fedora":
+        data.variant = "fedora27" # this is for support in RHEL7, where fedora28 is not known yet
```
jan-cerny: This is a great catch! :1st_place_medal:

tests/ssg_test_suite/test_env.py
```diff
@@ -165,11 +161,7 @@ def _delete_saved_state(self, snapshot):
         self.snapshot_stack.revert()
 
     def _local_oscap_check_base_arguments(self):
-        command_base = []
-        command_base.extend(
-            ['oscap-vm', "domain", self.domain_name,
-             'xccdf', 'eval'])
-        return command_base
+        return ['oscap-vm', "domain", self.domain_name, 'xccdf', 'eval']
```
jan-cerny: What if this would be just a variable instead of a function? It anyway just returns some data,

tests/README.md
dahaic: good point

tests/ssg_test_suite/test_env.py
dahaic: There are multiple `_base_arguments` methods. I agree these could be simplified. As the fix is piggybacked just because it was blocking, I would prefer not to rework it more than necessary. Can be adjusted in different PR.

dahaic:
@jan-cerny ok, fixed what I feel should be fixed :)

jan-cerny:

PR #3277, opened 2018-09-07
Fix testoval.py

### LINKED ISSUE #3239: testoval.py is producing incorrect results in addition reading the wrong OVAL file.
#### Description of problem:
PYTHONPATH=`~/development/scap-security-guide/.pyenv.sh` ~/development/scap-security-guide/utils/testoval.py shared.xml 
Evaluating with OVAL tempfile: /tmp/apt_sources_list_official8wMvT2.xml
OVAL Schema Version: 5.11
Writing results to: /tmp/apt_sources_list_official8wMvT2.xml-results
Running command: oscap oval eval --results /tmp/apt_sources_list_official8wMvT2.xml-results /tmp/apt_sources_list_official8wMvT2.xml

Definition oval:scap-security-guide.testing-apt_sources_list_official:def:1: false
Evaluation done.

master

#### Steps to Reproduce:

1.
2.
3.
4.

redhatrises:
- Fixes issue where testoval.py wouldn't work with duplicate OVAL file names
- Add missing namespace of literal_component
- Fixes #3239

### DIFF
--- a/ssg/oval.py
+++ b/ssg/oval.py
@@ -119,7 +119,7 @@ def replace_external_vars(tree):
             sys.exit(2)
         # replace tag name: external -> local
         node.tag = "{%s}local_variable" % ovalns
-        literal = ET.Element("literal_component")
+        literal = ET.Element("oval:literal_component")
         literal.text = os.environ[extvar_id]
         node.append(literal)
         # TODO: assignment of external_variable via environment vars, for
@@ -145,6 +145,9 @@ def find_testfile(oval_id):
 
     Returns path to OVAL file or None if not found.
     """
+    if os.path.isfile(os.path.abspath(oval_id)):
+        return os.path.abspath(oval_id)
+
     if oval_id.endswith(".xml"):
         oval_id, _ = os.path.splitext(oval_id)
         oval_id = os.path.basename(oval_id)


ssg/oval.py
```diff
@@ -119,7 +119,7 @@ def replace_external_vars(tree):
             sys.exit(2)
         # replace tag name: external -> local
         node.tag = "{%s}local_variable" % ovalns
-        literal = ET.Element("literal_component")
+        literal = ET.Element("oval:literal_component")
```
mpreisler: Are you sure this works? Is the oval ns registered in this context? If this works I think it's undefined behavior and we are just lucky. We should do this with the proper API.

redhatrises: @mpreisler then this issue is much larger and was created with https://github.com/ComplianceAsCode/content/pull/3185 which you acked. ;)

mpreisler: nvm, there is a `ET.register_namespace("oval", ovalns)` call there.

mpreisler: > @mpreisler then this issue is much larger and was created with #3185 which you acked. ;)

as you surely know, I never make mistakes :-DDD

redhatrises: 2014 mpreisler would.... but not 2018 mpreisler :D

cipherboy:
ACK from me. Thanks @redhatrises!

I wonder if we could simplify the logic of find oval...

dahaic:
Thanks, works for me. The issue linked is terrible, by the way.

PR #3278, opened 2018-09-10
Simplify OVAL in ensure_gpgcheck_globally_activated

jan-cerny:
#### Description:

This OVAL is much simpler thanks to Jinja macros. Platform check
is not needed anymore.

#### Rationale:


mildass:
Looks good. Thanks

PR #3281, opened 2018-09-10
Tests and bash for gnome_gdm_disable_guest_login

jan-cerny:
#### Description:

Adds platform string into the bash remediation and renames the
script to be used in Fedora profile.
Adds tests to test suite.

#### Rationale:


redhatrises:
Ack

PR #3265, opened 2018-09-06
Audit rules unsuccesful file deletion

yuumasato:
#### Description:

- Add rules to check audit rules to log for unsuccessful file deletion attempts. 
  - Rule prose was derived from existing unsuccessful file modification rule.
- OVALs are the same as generated by template [template_OVAL_audit_rules_unsuccessful_file_modification](https://github.com/ComplianceAsCode/content/blob/master/shared/templates/template_OVAL_audit_rules_unsuccessful_file_modification).
  - I'm basically abusing file modification to check for file deletion, which is kind of a file modification.
- Added basic test for `audit_rules_unsuccessful_file_modification_unlink`.
  - After #3244 they should be tested by template test scenarios.

#### Rationale:

- OSPP4.2 likes to audit unsuccessful events

PR #3282, opened 2018-09-10
Update to audit rules based on feedback from jcerny

dahaic:
#### Description:

- Cleans up tests slightly, removing sanitization of exit codes when it is not needed.
- Adjusts wording of disclaimers in new audit rules.

#### Rationale:

- Some good points by Jan has been skipped when I merged PR #3265


jan-cerny:
Works OK for me.

PR #3284, opened 2018-09-10
Ansible remediation fails on ensure group owner if cron.allow doesn't exist.

rchayes:
… exist.

#### Description:

- Ansible remediation fails on trying to set group owner of /etc/cron.allow if cron.allow doesn't exist but existence of file is not required.  

#### Rationale:

- Added to ansible template to only attempt to set group owner of file if the file exists.  Will duplicate changes for setting owner remediation and create another pull request if this looks ok.

### DIFF
--- a/shared/templates/template_ANSIBLE_file_groupowner
+++ b/shared/templates/template_ANSIBLE_file_groupowner
@@ -3,11 +3,17 @@
 # strategy = configure
 # complexity = low
 # disruption = low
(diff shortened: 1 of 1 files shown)


jan-cerny:
The playbook works great for me with your patch. Thanks!

redhatrises:
@rchayes will you do the same for file owner as well?

PR #3287, opened 2018-09-11
Add bash and tests for rule grub2_audit_argument

jan-cerny:
#### Description:
Add a bash remediation and basic test scenarios for rule grub2_audit_argument

#### Rationale:


jan-cerny:
@matusmarhefka and me have noticed that there is already an OVAL in `rhel6` directory and Ansible and Bash in `shared` directory for "bootloader_audit_argument" in `shared`, which is very similar to this. However, no rule "bootloader_audit_argument" exists anywhere and is not included in any profile.

My opinion is that it is a leftover from some unfinished unshadowing activity.

I will consolidate those files into `linux_os`.

jan-cerny:
I have changed some comments.

PR #3288, opened 2018-09-11
Move bootloader_audit_argument content to a correct place

jan-cerny:
#### Description:

I have found OVAL, Bash and Ansible for bootloader_audit_argument.
But no such rule exists. But there is rule grub_legacy_audit_argument
which has no OVAL, no Bash and no Ansible. That looks like
a wrong rename during unshadowing and should be consolidated.
Moreover, there exist a very similar rule grub2_audit_argument,
which is targeted at RHEL7 and Fedora. Therefore, we will
mark grub_legacy_audit_argument as RHEL 6 only.

#### Rationale:

The  grub_legacy_audit_argument was missing OVAL, Bash and Ansible and there was content that belongs to no rule.


shared/fixes/bash/bootloader_audit_argument.sh
```diff
@@ -1,14 +0,0 @@
-# platform = multi_platform_rhel
-
-{{% if product == "rhel7" %}}
-# Correct the form of default kernel command line in /etc/default/grub
-grep -q ^GRUB_CMDLINE_LINUX=\".*audit=0.*\" /etc/default/grub && \
-  sed -i "s/audit=[^[:space:]\+]/audit=1/g" /etc/default/grub
-if ! [ $? -eq 0 ]; then
-  sed -i "s/\(GRUB_CMDLINE_LINUX=\)\"\(.*\)\"/\1\"\2 audit=1\"/" /etc/default/grub
-fi
-{{% endif %}}
```
yuumasato: How about moving `rhel7` section of the fix be moved to the `grub2` [rule](https://github.com/ComplianceAsCode/content/tree/d7df03b5ffb8d76973eab64188bcd41f7835b70f/linux_os/guide/system/auditing/grub2_audit_argument)?

yuumasato: Nevermind, I see the other PR now.

jan-cerny: Yes. However I wonder that the RHEL 6 version does not edit the config file at all.

redhatrises:
ack

PR #3290, opened 2018-09-11
Add Fedora to Bash for sshd_disable_kerb_auth

jan-cerny:
#### Description:
Add Fedora to Bash for sshd_disable_kerb_auth

#### Rationale:

Note: Tests are already there.


redhatrises:
LGTM

PR #3106, opened 2018-07-09
[WIP] Inconsistent memcache_timeout, ssh_known_hosts_timeout wrt man sssd.conf

amitkumar50:
ssg-rhel7-ds.xml reports inconsistent values wrt sssd.conf
ssg-rhel7-ds.xml recommends:

```
memcache_timeout = 86400
ssh_known_hosts_timeout = 86400
```

sssd.conf default

```
memcache_timeout = 300
ssh_known_hosts_timeout = 180
```

PR changes memcache_timeout,ssh_known_hosts_timeout values as defined in sssd.conf

PLEASE VERIFY THIS WITH SSSD DEVELOPERS PRIOR TO MERGE.

Signed-off-by: Amit Kumar <amitkuma@redhat.com>

PR #3292, opened 2018-09-11
Finish rule sssd_memcache_timeout for Fedora

jan-cerny:
#### Description:
Adds a Bash remediation and tests for rule sssd_memcache_timeout.
Remediation should be also applicable to RHEL systems.
Makes use of XCCDF value `var_sssd_memcahe_timeout` and fixes the broken regex in OVAL.

#### Rationale:

### DIFF
--- a/linux_os/guide/services/sssd/sssd_memcache_timeout/ansible/shared.yml
+++ b/linux_os/guide/services/sssd/sssd_memcache_timeout/ansible/shared.yml
@@ -10,7 +10,7 @@
     dest: /etc/sssd/sssd.conf
     section: nss
     option: memcache_timeout
-    value: {{ var_sssd_memcache_timeout }}
+    value: "{{ var_sssd_memcache_timeout }}"
     create: yes
   tags:
     @ANSIBLE_TAGS@


redhatrises:
Blocked by #3106.

I am going to finish up #3106 and then this will need a rebase.

linux_os/guide/services/sssd/sssd_memcache_timeout/bash/shared.sh
```diff
@@ -0,0 +1,21 @@
+# platform = multi_platform_rhel, multi_platform_fedora
+
+# Include source function library.
+. /usr/share/scap-security-guide/remediation_functions
+
+SSSD_CONF="/etc/sssd/sssd.conf"
+MEMCACHE_TIMEOUT_REGEX="[[:space:]]*\[nss]([^(\n)]*(\n)+)+?[[:space:]]*memcache_timeout"
+NSS_REGEX="[[:space:]]*\[nss]"
+TIMEOUT="86400"
```
redhatrises: Replace `TIMEOUT` with `var_sssd_memcache_timeout` as it is now configurable in `var_sssd_memcache_timeout.var`

jan-cerny:
I'm sorry I completely forgot about the other PR that touches this rule. I will rebase this and update accordingly. Thanks @redhatrises for pointing this out.

jan-cerny:
The rule `sssd_memcache_timeout` is now broken in current master (as of 3c171f018f4). 

It is probably caused by #3106.


```
[jcerny@thinkpad build{master}]$ sudo cat  /etc/sssd/sssd.conf
[nss]
memcache_timeout = 86400
[jcerny@thinkpad build{master}]$ sudo oscap xccdf eval --profile ospp --rule xccdf_org.ssgproject.content_rule_sssd_memcache_timeout ssg-fedora-ds.xml                                                            
Title   Configure SSSD's Memory Cache to Expire
Rule    xccdf_org.ssgproject.content_rule_sssd_memcache_timeout
E: oscap:     Error occured when comparing a variable 'oval:ssg-var_sssd_memcache_timeout:var:1' value '300' with collected item entity = '                                                                       
'
Result  fail

OpenSCAP Error: Conversion of the string "
" to an integer (64 bits) failed: Invalid argument [/home/jcerny/openscap/src/OVAL/results/oval_cmp.c:116]                                                                                                        
[jcerny@thinkpad build{master}]$ 
```

I need to figure out what's wrong with the rule before I will be able to update this PR.

jan-cerny:
I think I have found the problem:

The regular expression use in OVAL `^\[nss]([^\n]*\n+)+?memcache_timeout[\s]+=[\s]+(\d+)$` is wrong.
For example, let's have the following file:


```
[nss]
something = bogus
memcache_timeout = 86400
```
On this example, the first group `([^\n]*\n+)+?` matches `something = bogus\n` and  second group matches `86400`.

Then `oscap` tries to convert `something = bogus\n` to integer in order to compare it with the ext. variable, which of course fails. 

I think the first group should be non-capturing.

jan-cerny:
I have fixed the regular expression, I have set sssd memcache timeout to 1 day in Fedora OSPP profile, I have set XCCDF value in bash and Ansible remediation.

redhatrises:
ACK.

PR #3293, opened 2018-09-11
Ansible remediation fails on setting owner of file if cron.allow doesn't exist

rchayes:
#### Description:

- Ansible remediation fails on set owner if cron.allow doesn't exist.

#### Rationale:

- Checks for existence of file before setting owner.


redhatrises:
Ack.

PR #3294, opened 2018-09-11
Remove use of loops for single items in Ansible templates

redhatrises:
- with_items loop was used for a single item list in many of the Ansible templates.
  This removes use of loops for single valued lists. In addition, with_items will
  be replaced by loop and flatten in the future.


mpreisler:
Thx, this will make it much simpler to understand.

mpreisler:
ACK

PR #3296, opened 2018-09-11
Run crypto command in Ansible script

redhatrises:
Missed this in #3240

### DIFF
--- a/linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/ansible/fedora.yml
+++ b/linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/ansible/fedora.yml
@@ -13,3 +13,6 @@
     create: yes
   tags:
     @ANSIBLE_TAGS@
+
+- name: Verify that Crypto Policy is Set (runtime)
+  shell: /usr/bin/update-crypto-policies set {{ var_system_crypto_policy }}


linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/ansible/fedora.yml
```diff
@@ -13,3 +13,15 @@
     create: yes
   tags:
     @ANSIBLE_TAGS@
+
+- name: Verify that Crypto Policy is Set for the Current State
```
dahaic: I believe you do not want to actually create the file? Do you want, instead to run `update-crypto-policies` based on the result?

```
- name: Verify that Crypto Policy is Set for the Current State
  lineinfile:
    path: /etc/crypto-policies/state/current
    regexp: '^(?!#)(\S+)$'
    line: "{{ var_system_crypto_policy }}"
    create: yes
  check_mode: yes
  register: crypto_state_current
  tags:
    @ANSIBLE_TAGS@

- name: Verify that Crypto Policy is Set (runtime)
  shell: /usr/bin/update-crypto-policies set {{ var_system_crypto_policy }}
  when: crypto_state_current is changed
```

redhatrises: @dahaic I was actually thinking both since we do this same logic in other places in our content.

dahaic: Isn't `state/current` just information, meant to be changed internally? That's something we should leave for the script to change - the same way as we let systemd to set symlinks as it sees fit, instead of handling it ourselves (when service is being enabled/disabled)

redhatrises: I am fine with that logic. Removed `state/current` ansible fix script.

dahaic:
ACK, thanks @redhatrises !

PR #3298, opened 2018-09-11
Enable Ansible audit_rules_file_deletion_events template

redhatrises:


mpreisler:
That is really weird. On the off chance that this was done intentionally, do the templates generate good ansible code? Do the fixes work well?

redhatrises:
> That is really weird. On the off chance that this was done intentionally, do the templates generate good ansible code? Do the fixes work well?

Okay... so if I am following the history right, it looks like the `audit_rules_file_deletion_events` python script was added with the commented out ansible lines and no Ansible template. Then, @shawndwells added the template and looks like he enabled it, but had a very screwed up PR that needed a huge rebase which ended up commenting it out again.... assuming I am following the history correctly.

redhatrises:
@mpreisler code is every similar to `template_ANSIBLE_audit_rules_dac_modification`. I just removed what appears to be a duplicate rule insert for x86.

PR #3302, opened 2018-09-11
Ansible remediation fails setting file permissions if wildcard is used.

rchayes:
#### Description:

- Ansible remediation fails setting permissions if file path is provided with wildcard (i.e. /etc/ssh/*.pub).

#### Rationale:

- Altered variables to split file directory and file name in create_permission.py.  Ansible template now utilizes find and iterates through matched files to set file permissions.  Seems to work fine with individual files or path with wildcard.

### DIFF
--- a/shared/templates/create_permissions.py
+++ b/shared/templates/create_permissions.py
@@ -101,7 +101,8 @@ class PermissionGenerator(FilesGenerator):
                 self.file_from_template(
                     "./template_ANSIBLE_file_permissions",
                     {
-                        "FILEPATH":      full_path,
+                        "FILEPATH":      dir_path,
+                        "FILENAME":      file_name,
                         "FILEMODE":      mode,
                     },
                     "./ansible/file_permissions{0}.yml", path_id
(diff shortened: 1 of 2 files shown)


redhatrises:
Rather than do this for all files, we should probably only do this for those using regex.... could even create a regex template.

rchayes:
I'm not sure I see the advantage to breaking them out.  The find handles a "search" for a single file as well as a wildcard.  Additionally, if they stay separate there needs to be a check for individual files to ensure they exist or setting permissions will break the playbook (find handles this as is).  That being said I don't mind breaking it into separate files, separate template file and separate csv for regex paths I assume?

redhatrises:
> The find handles a "search" for a single file as well as a wildcard. 

Don't need to be doing a "search" for a single defined file or directory.

>  separate template file and separate csv for regex paths I assume?

If that is easiest, yes.

rchayes:
Ok I'll see if I can knock it out this afternoon.  Thanks.

matejak:
OK to test

matejak:
add to whitelist

yuumasato:
@rchayes I'll be closing this as this seems to be superseded by #3616.

PR #3303, opened 2018-09-11
Add missing OSPP FIPS remediations

redhatrises:

### DIFF
--- a/linux_os/guide/system/software/integrity/fips/package_dracut-fips_installed/ansible/rhel7.yml
+++ b/linux_os/guide/system/software/integrity/fips/package_dracut-fips_installed/ansible/rhel7.yml
@@ -1,0 +1,13 @@
+# platform = Red Hat Enterprise Linux 7
+# reboot = false
+# strategy = enable
+# complexity = low
+# disruption = low
+- name: Ensure dracut-fips is installed
+  package:
+    name: dracut-fips
+    state: present
+    when: ansible_distribution == 'Red Hat Enterprise Linux'
+  tags:
+    @ANSIBLE_TAGS@
+
--- a/linux_os/guide/system/software/integrity/fips/package_dracut-fips_installed/bash/rhel7.sh
+++ b/linux_os/guide/system/software/integrity/fips/package_dracut-fips_installed/bash/rhel7.sh
@@ -1,0 +1,8 @@
+# platform = Red Hat Enterprise Linux 7
+
+# include remediation functions library
+. /usr/share/scap-security-guide/remediation_functions
+
+if [ $(cat /etc/system-release-cpe | cut -d":" -f3) == "redhat" ] ; then
+	package_install dracut-fips
+fi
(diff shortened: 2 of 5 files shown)


linux_os/guide/system/software/integrity/fips/package_dracut-fips_installed/ansible/rhel7.yml
```diff
@@ -0,0 +1,12 @@
+# platform = Red Hat Enterprise Linux 7
+# reboot = false
+# strategy = enable
+# complexity = low
+# disruption = low
+- name: Ensure dracut-fips is installed
```
jan-cerny: I think we should use templates instead.

dahaic:
@redhatrises by that you mean you want this to be available only on platforms that are certifiable? I still do not get it. I agree we need specific OVAL, as it has `installed_OS_is_certified`. Remediation though is exactly the same as any other rule.

By not templating those, you just increase the probability of failure.

mpreisler:
I agree with @dahaic and @jan-cerny , we should use custom OVAL  because of installed_OS_is_certified but the fixes can be templated. That way we lower the probability of bugs and get better coverage across fix languages.

redhatrises:
> @redhatrises by that you mean you want this to be available only on platforms that are certifiable? I still do not get it. I agree we need specific OVAL, as it has `installed_OS_is_certified`. Remediation though is exactly the same as any other rule.
> 
> By not templating those, you just increase the probability of failure.

Actually they are no longer the same. As the OVAL checks for a certified OS, so now do the remediations. FIPS content, due to certification, needs to be not be blindly implemented like it has been through templates.

jan-cerny:
Does it try to solve the problem that CentOS and similar derivatives are not FIPS certified? Why should a remediation check for CPE?

linux_os/guide/system/software/integrity/fips/package_dracut-fips_installed/ansible/rhel7.yml
```diff
@@ -0,0 +1,13 @@
+# platform = Red Hat Enterprise Linux 7
```
mpreisler: use multi_platform_rhel instead please

linux_os/guide/system/software/integrity/fips/package_dracut-fips_installed/bash/rhel7.sh
```diff
@@ -0,0 +1,8 @@
+# platform = Red Hat Enterprise Linux 7
```
mpreisler: multi_platform_rhel

mpreisler:
Ah, OK. Could you please unify the fix flags (severity, disruption, etc...) and use multi_platform_rhel for all of them? Then I think this can be merged.

redhatrises:
> Ah, OK. Could you please unify the fix flags (severity, disruption, etc...) and use multi_platform_rhel for all of them? Then I think this can be merged.

done

shawndwells:
Requested changes were made. Merging.

PR #3306, opened 2018-09-12
Fedora accounts_passwords_pam_faillock_deny_root rule and remediation fix

mildass:
#### Description:

- Add accounts_passwords_pam_faillock_deny_root rule for Fedora
- Fix OVAL (line with `pam_unix.so` can end right after `pam_unix.so`. Changing quantifier `+` to `*` fixes it)
- Fix bash remediation (it was fixing only cases when both tested lines were missing OR were here but wrong. It didn't work when one line was here but incorrect and other line was missing)
- Add tests


jan-cerny:
Thank you very much. The result of datastream looks OK for me.

PR #3308, opened 2018-09-12
Rule require_singleuser_auth for Fedora

jan-cerny:
#### Description:
Adjusts rule require_singleuser_auth for Fedora.

On Fedora, "/sbin/sulogin" has been superseded by "/usr/lib/systemd/systemd-sulogin-shell". We need to adjust the OVAL and Bash remediation accordingly.

Also adds 2 sceanrios to the test suite.

#### Rationale:

### DIFF
--- a/linux_os/guide/system/accounts/accounts-physical/require_singleuser_auth/bash/shared.sh
+++ b/linux_os/guide/system/accounts/accounts-physical/require_singleuser_auth/bash/shared.sh
@@ -2,20 +2,19 @@
 
 {{% if init_system == "systemd" -%}}
 
-{{% if product == "fedora" -%}}
 service_file="/usr/lib/systemd/system/rescue.service"
+
+{{% if product == "fedora" -%}}
 sulogin="/usr/lib/systemd/systemd-sulogin-shell"
+{{%- else -%}}
+sulogin="/sbin/sulogin"
+{{%- endif %}}
+
 if grep "^ExecStart=.*" "$service_file" ; then
     sed -i "s%^ExecStart=.*%ExecStart=-$sulogin rescue%" "$service_file"
 else
     echo "ExecStart=-$sulogin rescue" >> "$service_file"
 fi
-{{%- else -%}}
-grep -q "^ExecStart=\-.*/sbin/sulogin" /usr/lib/systemd/system/rescue.service
-if ! [ $? -eq 0 ]; then
-    sed -i "s/ExecStart=-.*-c \"/&\/sbin\/sulogin; /g" /usr/lib/systemd/system/rescue.service
-fi
-{{%- endif -%}}
 
 {{%- else -%}}
 grep -q ^SINGLE /etc/sysconfig/init && \
--- a/linux_os/guide/system/accounts/accounts-physical/require_singleuser_auth/bash/shared.sh
+++ b/linux_os/guide/system/accounts/accounts-physical/require_singleuser_auth/bash/shared.sh
@@ -1,10 +1,22 @@
-# platform = multi_platform_rhel
+# platform = multi_platform_rhel, multi_platform_fedora
 
 {{% if init_system == "systemd" -%}}
+
+{{% if product == "fedora" -%}}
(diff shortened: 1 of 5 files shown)


redhatrises:
Just glancing at this, but I don't suppose that we can override this in `/etc/systemd/`? If we can't that is fine, but it would be better for handling updates and such without having to modify `/usr/lib/systemd/

jan-cerny:
@redhatrises Good point.

Actually this is a default value on Fedora.

Do you think that the OVAL should look at multiple places? Eg. `.d` directories, `/etc/systemd`? And what about reworking the OVAL with systemdunitproperty test?

jan-cerny:
I am unable to get the ExecStart value using systemdunitproperty test in OVAL, so it seems it has been done using textfilecontent54 test by purpose.

redhatrises:
> Do you think that the OVAL should look at multiple places?

Probably should since I am sure many people won't make the change in `/etc` even though from a maintenance standpoint it's nicer.

mpreisler:
> I am unable to get the ExecStart value using systemdunitproperty test in OVAL, so it seems it has been done using textfilecontent54 test by purpose.

I think this should be reported as a bug in openscap

yuumasato:
> If we can't that is fine, but it would be better for handling updates and such without having to modify `/usr/lib/systemd/`

Just as fact, we check whether `auditd` is using `auditctl` or `augenrules` by looking at `/usr/lib/systemd/`:

- https://github.com/ComplianceAsCode/content/blob/master/shared/checks/oval/audit_rules_auditctl.xml
- https://github.com/ComplianceAsCode/content/blob/master/shared/checks/oval/audit_rules_augenrules.xml

jan-cerny:
@yuumasato Do you think that we should read only `/usr/lib/systemd` and not `/etc/systemd/user`? Or do you think the check for auditctl and augen rules should be also extended to read `/etc/systemd/user` ?

jan-cerny:
@yuumasato My opinion is that we should be consistent. What happens if we don't check `/etc/systemd/user` ?

jan-cerny:
The problem with /etc/systemd will be solved in a separate PR because it affects multiple rules, not just this. I will strip off the commit here to be consistent with what we have in other rules.

jan-cerny:
I have removed 23bb27b and rebased on top of master's head.

jan-cerny:
This problem is tracked in https://github.com/ComplianceAsCode/content/issues/3359

mpreisler:
@openscap-ci test this please

linux_os/guide/system/accounts/accounts-physical/require_singleuser_auth/bash/shared.sh
```diff
@@ -1,10 +1,22 @@
...
+{{% if product == "fedora" -%}}
+service_file="/usr/lib/systemd/system/rescue.service"
+sulogin="/usr/lib/systemd/systemd-sulogin-shell"
+if grep "^ExecStart=.*" "$service_file" ; then
+    sed -i "s%^ExecStart=.*%ExecStart=-$sulogin rescue%" "$service_file"
+else
+    echo "ExecStart=-$sulogin rescue" >> "$service_file"
+fi
+{{%- else -%}}
 grep -q "^ExecStart=\-.*/sbin/sulogin" /usr/lib/systemd/system/rescue.service
```
dahaic: can you reuse `service_file` variable? (Possibly change that one and `sulogin` to CONSTANTS) The logic difference is now looks bigger than it really is.

jan-cerny:
I have reused the service file in the non-Fedora systemd remediation.

matejak:
jenkins test this please

yuumasato:
@jan-cerny Nevermind my comment! ;D

jan-cerny:
@yuumasato This rule has a section for RHEL6, which is selected by a Jinja macro depending on product type. This section refers to /etc/sysconfig/init.

jan-cerny:
@openscap-ci test this please

jan-cerny:
aha

yuumasato:
These changes look good to me.

PR #3309, opened 2018-09-12
Fixed grep usage.

matejak:
Made improvements to `grep` invocations that fix a class of Shellcheck warnings, improves performance and fixes possible bugs caused by `grep` ranges being influenced by locale.

* Removed redundant backticks in conditions.
* Added LC_ALL=C as we are interested in ASCII content and `[A-Z]` can be interpreted differently depending on locale setting.
* Added -m 1 to stop searching after one match is found if applicable.


mpreisler:
I wonder whether we should have some automated testing that will fail PRs with "wrong" usage of `grep` in the bash code.

Ideally we'd have a shared bash function that would do this but that is out of scope for this PR. Plus shared functions are kind of nasty in our codebase.

PR #3311, opened 2018-09-12
Finish rule sssd_offline_cred_expiration

matusmarhefka:
#### Description:

Added a bash remediation and test coverage for the rule sssd_offline_cred_expiration.

#### Rationale:

### DIFF
--- a/linux_os/guide/services/sssd/sssd_offline_cred_expiration/oval/shared.xml
+++ b/linux_os/guide/services/sssd/sssd_offline_cred_expiration/oval/shared.xml
@@ -25,7 +25,7 @@
   </ind:textfilecontent54_test>
   <ind:textfilecontent54_object id="obj_sssd_offline_cred_expiration" version="1">
     <ind:filepath>/etc/sssd/sssd.conf</ind:filepath>
-    <ind:pattern operation="pattern match">^\[pam]([^\n]*\n+)+?offline_credentials_expiration[\s]+=[\s]+1$</ind:pattern>
+    <ind:pattern operation="pattern match">^[\s]*\[pam](?:[^\n\[]*\n+)+?[\s]*offline_credentials_expiration[\s]*=[\s]*1$</ind:pattern>
     <ind:instance datatype="int">1</ind:instance>
   </ind:textfilecontent54_object>
 </def-group>
--- a/tests/data/group_services/group_sssd/rule_sssd_offline_cred_expiration/comment.fail.sh
+++ b/tests/data/group_services/group_sssd/rule_sssd_offline_cred_expiration/comment.fail.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+# profiles = xccdf_org.ssgproject.content_profile_ospp
+
+source common.sh
+
+echo -e "[pam]\n#offline_credentials_expiration = 1" >> $SSSD_CONF
--- a/tests/data/group_services/group_sssd/rule_sssd_offline_cred_expiration/correct_value.pass.sh
+++ b/tests/data/group_services/group_sssd/rule_sssd_offline_cred_expiration/correct_value.pass.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+# profiles = xccdf_org.ssgproject.content_profile_ospp
+
+source common.sh
+
+echo -e "[pam]\noffline_credentials_expiration = 1" >> $SSSD_CONF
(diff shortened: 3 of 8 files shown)


linux_os/guide/services/sssd/sssd_offline_cred_expiration/oval/shared.xml
```diff
@@ -25,7 +25,7 @@
   </ind:textfilecontent54_test>
   <ind:textfilecontent54_object id="obj_sssd_offline_cred_expiration" version="1">
     <ind:filepath>/etc/sssd/sssd.conf</ind:filepath>
-    <ind:pattern operation="pattern match">^\[pam]([^\n]*\n+)+?offline_credentials_expiration[\s]+=[\s]+1$</ind:pattern>
+    <ind:pattern operation="pattern match">^\[pam](?:[^\n]*\n+)+?offline_credentials_expiration[\s]+=[\s]+1$</ind:pattern>
```
redhatrises: @matusmarhefka why this change?

matusmarhefka: @redhatrises So why would you want to capture this group? It just doesn't make sense to me and that is why I rather set it as a group without capture.

jan-cerny: It's not needed to do that, the rule works without that, but it's better.

tests/data/group_services/group_sssd/rule_sssd_offline_cred_expiration/comment.fail.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+# profiles = xccdf_org.ssgproject.content_profile_ospp
+
+SSSD_CONF="/etc/sssd/sssd.conf"
```
dahaic: Please move the setup boilerplate to function that you can source and call? It's cleaner that way.

tests/data/group_services/group_sssd/rule_sssd_offline_cred_expiration/comment.fail.sh
```diff
@@ -0,0 +1,11 @@
...
+
+dnf -y install sssd
```
dahaic: We don't want the test to work only on fedora, please change the logic to install package by appropriate packaging mechanism. You can get inspired by
`data/group_system/group_software/group_updating/group_updating_utils.sh`

linux_os/guide/services/sssd/sssd_offline_cred_expiration/oval/shared.xml
dahaic: I agree with Matus - explicit is better.

tests/data/group_services/group_sssd/rule_sssd_offline_cred_expiration/wrong_value.fail.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+# profiles = xccdf_org.ssgproject.content_profile_ospp
+
+SSSD_CONF="/etc/sssd/sssd.conf"
+dnf -y install sssd
+systemctl enable sssd
```
dahaic: ditto

tests/data/group_services/group_sssd/rule_sssd_offline_cred_expiration/correct_value.pass.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+# profiles = xccdf_org.ssgproject.content_profile_ospp
+
+SSSD_CONF="/etc/sssd/sssd.conf"
+
+dnf -y install sssd
+systemctl enable sssd
```
dahaic: ditto

tests/data/group_services/group_sssd/rule_sssd_offline_cred_expiration/comment.fail.sh
matusmarhefka: I will use `yum` as it should work on both, Fedora and RHEL. IMO this should be solved by the test suite which should export some variable for installing and removing packages based on the tested VM.

matusmarhefka:
Thanks for review @dahaic. It turned out that the original regex by @redhatrises used in OVAL check was not correct as it would match also options from subsequent sections after the `[pam]` section. I have corrected the regex and now all tests pass:
```
INFO - xccdf_org.ssgproject.content_rule_sssd_offline_cred_expiration
INFO - Script wrong_section.fail.sh using profile xccdf_org.ssgproject.content_profile_ospp OK
```

dahaic:
Good stuff. Thank you @matusmarhefka

PR #3312, opened 2018-09-12
tests/ssg_test_suite: Fixed cases where IGNORE_KNOWN_HOSTS_OPTIONS must be a list

matusmarhefka:
Test suite failed in case `--remediate-using ansible` was used because `IGNORE_KNOWN_HOSTS_OPTIONS` variable was expected to be a list - this PR converts this variable to list in such cases.


matejak:
Good catch, PR is highly appreciated!

PR #3313, opened 2018-09-12
Remove the extend_definition for dconf_gnome_screensaver_lock_enabled

jan-cerny:
GNOME screen saver lock delay is tested by another rule. The
delay is not mentioned in the rule description and causes false
positives if the dconf_gnome_screensaver_lock_delay is unselected
by tailoring.


mpreisler:
Good fix, makes sense to me.

ACK

PR #3316, opened 2018-09-12
Fix regex for unauthorized syscall rule with one syscall and filter

yuumasato:
#### Description:

- Regular expression to rule that records unsuccessful syscall events without filter matched audit rules with one syscall and filter.

#### Rationale:

- Test `one_sys_with_filter.fail.sh` doesn't pass without this patch.


jan-cerny:
I have run the tests that you provided and they pass for me. Thank you for your contributions.

PR #3317, opened 2018-09-12
Unknown severity shouldn't be used when a different severity is specified

redhatrises:


jan-cerny:
This was introduced by https://github.com/ComplianceAsCode/content/commit/88925baeeee85088092a33a6eb237665bcdf3cd2 . I think the only reason why it was introduced is to keep the severity of RHEL6 during unshadowing activities. But it doesn't make sense to keep "unknown" severity.

Thank you very much for this contribution.

PR #3319, opened 2018-09-14
Update rule accounts_password_pam_retry

jan-cerny:
#### Description:

This PR fixes OVAL, enables Bash remediation fro Fedora and adds 3 basic test scenarios for rule `accounts_password_pam_retry`.

This rule checks PAM modules arguments in `/etc/pam.d/system-auth`. We have discussed with @yuumasato if it would be better to check configuration in ` /etc/security/pwquality.conf` instead. Other similar rules from this directory already check ` /etc/security/pwquality.conf` instead of   `/etc/pam.d/system-auth`. All other rules from this group are templated.

But I looked into that again and I have realized that the rule description in 
https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/accounts/accounts-pam/password_quality/password_quality_pwquality/accounts_password_pam_retry/rule.yml
speaks about "permitted per-session". The other rules doesn't mention that "per-session":
Instead, `/etc/security/pwquality.conf` specifies default password quality requirements for the system passwords.

Moreover, `man 5 pwquality.conf` doesn't mention `retry` option at all, whereas `man 8 pam_pwquality` mentions `retry`. From that it seems it isn't possible to set `retry` option in ` /etc/security/pwquality.conf`.

Hopefully the rule isn't templated on purpose.

#### Rationale:

### DIFF
--- a/linux_os/guide/system/accounts/accounts-pam/password_quality/password_quality_pwquality/accounts_password_pam_retry/bash/shared.sh
+++ b/linux_os/guide/system/accounts/accounts-pam/password_quality/password_quality_pwquality/accounts_password_pam_retry/bash/shared.sh
@@ -1,8 +1,8 @@
-# platform = Red Hat Enterprise Linux 7
+# platform = Red Hat Enterprise Linux 7, multi_platform_fedora
 . /usr/share/scap-security-guide/remediation_functions
 populate var_password_pam_retry
 
-if grep -q "retry=" /etc/pam.d/system-auth; then   
+if grep -q "retry=" /etc/pam.d/system-auth ; then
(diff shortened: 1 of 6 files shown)


yuumasato:
> Hopefully the rule isn't templated on purpose.

As of now it is not possible to edit `retry` option in `/etc/security/pwquality.conf`, so this check will remain not templated. 

But thanks to @redhatrises , it will soon be possible to set `retry` option in `pwquality.conf` file.
Check https://github.com/libpwquality/libpwquality/pull/10.

yuumasato:
Ack, thanks for @jan-cerny .

PR #3321, opened 2018-09-14
Consistent profile IDs

mpreisler:
Let's start maintaining profile ID compatibility across versions. Before we do that let's change all IDs to short, sensible, descriptive IDs.

I will create a separate PR that changes titles of profiles to improve UX in SCAP Workbench, Satellite 6, HTML guides and other user facing apps.


redhatrises:
IIRC the profiles with `disa` in the title were called that for a reason. The Fedora `ospp` profile should stay at `ospp`, and why rename all the `rhel7` and `rhel6` profiles? Why not just keep them as is and phase them out when the products become end-of-life?

mpreisler:
@redhatrises What's the reason then? STIG is a DISA profile, DISA is implied. So is the product for all profiles.

The idea is to have simpler profile IDs so that it's simpler to write and remember them. Whenever people ask me about command lines and profiles I never remember if it's stig-rhel7-disa or disa-stig-rhel7 or whatnot.

redhatrises:
> @redhatrises What's the reason then? STIG is a DISA profile, DISA is implied. So is the product for all profiles.

It's to denote an official STIG vs a Vendor STIG because you can have a vendor STIG. In fact, we have  Vendor STIGs right now for satellite, ipa, ansible, amq, etc. that are not official DISA STIGs. Also, we have already created consternation once in the rhel7 and rhel6 ecosystems by renaming the profiles, can we please sunset the profile naming with rhel<=7 and do the new naming with Fedora and rhel>=7?

shawndwells:
On 9/14/18 3:54 PM, redhatrises wrote:
>
>     @redhatrises <https://github.com/redhatrises> What's the reason
>     then? STIG is a DISA profile, DISA is implied. So is the product
>     for all profiles.
>
> It's to denote an official STIG vs a Vendor STIG because you can have 
> a vendor STIG. In fact, we have Vendor STIGs right now for satellite, 
> ipa, ansible, amq, etc. that are not official DISA STIGs. Also, we 
> have already created consternation once in the rhel7 and rhel6 
> ecosystems by renaming the profiles, can we please sunset the profile 
> naming with rhel<=7 and do the new naming with Fedora and rhel>=7?
>
+1

Eventually the "Vendor STIG" vs "DISA STIG" silliness will cease as 
everything is transitioned to DoD Configuration Annexes..... but that 
will be a slow transition.

This has already happened in the operating system space though. 
OSPP+Config Annex replaces what was known as the "RHEL7 Vendor STIG." 
Other tech like Satellite/IPA/Ansible/RHV still follow the old system.

jan-cerny:
I like this change because I think it makes the use of ComplianceAsCode content easier. Users will not have to remember differences in IDs. They will have to type less. The IDs should be same on all platforms. I also find the product prefixes and suffixes superfluous, because they are often implied from the file name.

yuumasato:
> It's to denote an official STIG vs a Vendor STIG because you can have a vendor STIG.

 @redhatrises  So is it the case that "SSG" STIG Profile is mistaken as the official STIG even when it is part of files named with prefix `ssg-`?
Can this confusion be clarified in Profile Description? Or title?
Something like was done for C2S Profile:
```
 This baseline was inspired by the Center for Internet Security (CIS) Red Hat Enterprise Linux 7 Benchmark, v2.1.1 - 01-31-2017. 
```

And also, can you obtain official STIGs somewhere else other than https://iase.disa.mil/stigs/Pages/index.aspx?

mpreisler:
It would be trivial to fix the jenkins issue but until we get consensus here I don't want to invest more time. Please provide feedback!

redhatrises:
Just like the OpenSCAP API which shouldn't change, this also should change for existing downstream releases. It is viewed as frustrating and disruptive to many downstream users. In addition, many of them have to go through change processes and other hoops to be able to use the new profile names because it is a security configuration change. Having the changes for after rhel7 makes more sense and allows a nicer implementation for downstream users.

@yuumasato it actually means that we can have the 2 STIG profiles. One vendor and one not e.g.

```
stig-rhel7-disa      <- non-vendor
stig-rhel7              <- vendor
```

redhatrises:
Closing this PR as this has been done for new versions and products.

PR #3322, opened 2018-09-14
Fail the build if an invalid PCI DSS ref is encountered

mpreisler:
Since we recently fixed the last invalid PCI-DSS ref I thought it would
be nice if we forbid adding any new invalid PCI-DSS refs.


mpreisler:
Side note: I think it's weird that this script is in `shared/transforms` but that is for another PR.

PR #3324, opened 2018-09-14
Add OpenSC content

### LINKED ISSUE #2324: Update RHEL7 smartcard settings to use OpenSC
OpenSC is the new and correct way to configure smartcards starting in RHEL 7.4. We need to update the smartcard rules to reflect this.

redhatrises:
- Fixes #2324


redhatrises:
Not necessarily sold on the location of the content, so if anyone has a better idea or cares for a different location. Need to add OVAL and some more remediation scripts before I would consider this no longer a WIP.

redhatrises:
Should be good to review now.

fedora/templates/csv/services_enabled.csv
```diff
@@ -3,5 +3,6 @@ chronyd,chrony,
 crond,cronie,
 firewalld,firewalld,
 ntpd,ntp,
+pcscd,pcsc-lite,
```
jan-cerny: This is related to rule `linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/smartcard_auth/`. Your PR adds this service to the list of enabled services for Fedora, but the OVAL for that rule is not enabled on Fedora, and the remediation scripts are not enabled on Fedora as well. I think you need to add multi_platform_fedora there.

linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/configure_opensc_card_drivers/bash/shared.sh
```diff
@@ -0,0 +1,14 @@
...
+# disruption = low
+
+. /usr/share/scap-security-guide/remediation_functions
+populate var_smartcard_drivers
+
+grep -qs "card_drivers =" /etc/opensc*.conf && \
+	sed -i "s/card_drivers =.*/card_drivers = $var_smartcard_drivers;/g" /etc/opensc*.conf
+if ! [ $? -eq 0 ]; then
+	sed -i "s/.*card_drivers =.*/        card_drivers = $var_smartcard_drivers;/g" /etc/opensc*.conf
+fi
```
jan-cerny: Will `replace_or_append` function work here?

linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/var_smartcard_drivers.var
```diff
@@ -0,0 +1,57 @@
...
+    itacns: itacns
+    jpki: jpki
+    MaskTech: MaskTech
+    mcrd: mcrd
+    muscle: muscle
+    myeid: myeid
+    npa: npa
+    oberthur: oberthur
+    openpgp: openpgp
+    other:
```
jan-cerny: Is this empty value intentional? It will get translated to "None" by the build system.

linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/configure_opensc_card_drivers/rule.yml
```diff
@@ -0,0 +1,45 @@
+documentation_complete: true
+
+prodtype: rhel7,fedora
+
+title: 'Configure opensc Smart Card Drivers'
+
```
jan-cerny: Will these rules be included in a profile?

linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/var_smartcard_drivers.var
redhatrises: Yes, it is, so that it is changeable.

linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/configure_opensc_card_drivers/rule.yml
redhatrises: Yes. I believe there is an SSSD rule that needs to be added as well before adding to the profile(s). Creating a separate PR for profile changes.

linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/configure_opensc_card_drivers/bash/shared.sh
redhatrises: No because of the uniqueness of `opensc.conf` naming and how the `.conf` is structured and formatted as well.

jan-cerny:
@redhatrises Thank you.

PR #3329, opened 2018-09-18
Untemplate rule mount_option_var tmp_bind

yuumasato:
#### Description:

- Remove entries of `/var/tmp,bind` from CSV files
- Re-add previous OVAL check
  - Extend it to also check fstab configuration
- Re-add bash fix,
  - Extend it to also mount binded `/var/tmp`
- Add tests

#### Rationale:

- The rule is not only about the correct mount option, it is also about binding the correct paths.

#### Note for reviewer: 
- The rule is not selected in any Profile, you will have to select it manually.


yuumasato:
I have added both, to test scenarios, and remediation function, clean up commands which umount any existing`/var/tmp` partition and remove any `/var/tmp` entry in `/etc/fstab`.

The last [commit](3af4ff2aab114fea7a05b340793d8f621b5e44a4) makes auxiliary function append the new fstab line, instead of overwriting whole file. 
This makes test scenario for this rule more real, with other stuff in `/etc/fstab`.
But this also changes behavior for other test scenarios which used it, I have briefly checked them, and I think this change shouldn't affect them.

linux_os/guide/system/permissions/partitions/mount_option_var_tmp_bind/oval/shared.xml
```diff
@@ -0,0 +1,85 @@
...
+      order to consolidate temporary storage into one location protected by the
+      same techniques as /tmp.</description>
+    </metadata>
+    <criteria operator="AND">
+      <criterion comment="Ensure /var/tmp is configured to bind mount to /tmp"
+      test_ref="test_configure_mount_option_var_tmp_bind_tmp" />
+      <criterion comment="Ensure /var/tmp is mounted"
+      test_ref="test_mount_option_var_tmp" />
+      <criteria operator="OR">
+        <criterion comment="Ensure /tmp is bind mounted"
```
matejak: The comment is misleading.

linux_os/guide/system/permissions/partitions/mount_option_var_tmp_bind/bash/shared.sh
```diff
@@ -0,0 +1,17 @@
...
+if grep -q -P '.*\/var\/tmp.*' /etc/fstab
+then
+  sed -i '/.*\/var\/tmp.*/d' /etc/fstab
+fi
+umount /var/tmp
+
+# Bind-mount /var/tmp to /tmp via /etc/fstab (preserving the /etc/fstab form)
+printf "%-24s%-24s%-8s%-32s%-3s\n" "/tmp" "/var/tmp" "none" "rw,nodev,noexec,nosuid,bind" "0 0" >> /etc/fstab
+
+mkdir /var/tmp
```
matejak: I would put `mkdir -p` here so it is more robust.

matejak:
Great work, I am happy to merge this PR.

PR #3331, opened 2018-09-18
Finish rule accounts_passwords_pam_faillock_interval for Fedora

jan-cerny:
#### Description:
This PR:
* Adds more criteria to the OVAL definition to align the OVAL definition with the description in XCCDF rule
* Fixes formatting of XCCDF rule description and OCIL description
* Adds multi_platform_fedora to both OVAL and Bash remediation
* Adds basic test scenarios for this rule
* Fixes bugs found in Bash remediation by the new tests

#### Rationale:

### DIFF
--- a/linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_passwords_pam_faillock_deny/bash/shared.sh
+++ b/linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_passwords_pam_faillock_deny/bash/shared.sh
@@ -2,39 +2,12 @@
 . /usr/share/scap-security-guide/remediation_functions
 populate var_accounts_passwords_pam_faillock_deny
 
+include_set_faillock_option
+
 AUTH_FILES[0]="/etc/pam.d/system-auth"
 AUTH_FILES[1]="/etc/pam.d/password-auth"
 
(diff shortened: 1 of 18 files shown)


mildass:
Thank @jan-cerny Works great, but wouldn't it be better to use `set_faillock_option_to_value_in_pam_file` from shared functions in that bash remediation?

jan-cerny:
I have added a baseline configuration for test scenarios and a new test.

mildass:
@jan-cerny If that remediation doesn't work properly on this rule, then I think it doesn't work on others (e.g. `accounts_passwords_pam_faillock_unlock_time` - same as this one, but works with unlock_time parameter).

Maybe it would be good if you make your remediation function configurable (to change/add value for arbitrary option, not only `fail_interval`) and add it to `shared/bash_remediation_functions` so we can use it for other `accounts_passwords_pam` rules.

jan-cerny:
@mildass That is a great observation. I'll try to look into where it is still used.

jan-cerny:
@openscap-ci test this please

jan-cerny:
@openscap-ci test this please

jan-cerny:
@openscap-ci test this please

dahaic:
Looks great, thank you @jan-cerny . ACK

PR #3332, opened 2018-09-18
Fix regular expression in sssd_memcache_timeout

jan-cerny:
#### Description:

It turned out that the original regex used in OVAL check
was not correct as it would match also options from subsequent sections
after the [nss] section. Also adds test to the SSG TS.

#### Rationale:
See https://github.com/ComplianceAsCode/content/pull/3311 for more details.

### DIFF
--- a/tests/data/group_services/group_sssd/rule_sssd_memcache_timeout/wrong_section.fail.sh
+++ b/tests/data/group_services/group_sssd/rule_sssd_memcache_timeout/wrong_section.fail.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+# profiles = xccdf_org.ssgproject.content_profile_ospp
+
+SSSD_CONF="/etc/sssd/sssd.conf"
+TIMEOUT="86400"
+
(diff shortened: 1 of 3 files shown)


mpreisler:
I really dislike how the regex is written but I don't know how we can do better...

ACK

tests/data/group_services/group_sssd/rule_sssd_memcache_timeout/wrong_section.fail.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# profiles = xccdf_org.ssgproject.content_profile_ospp
+
+SSSD_CONF="/etc/sssd/sssd.conf"
+TIMEOUT="86400"
+
+dnf -y install sssd
```
dahaic: I don't like explicit use of `dnf`.

mpreisler: haven't noticed this :-/

PR #3340, opened 2018-09-19
Ubuntu1804

xhudik:
#### Description:

- ubuntu1804 profile created


mpreisler:
@openscap-ci add to whitelist

mpreisler:
This is awesome, kudos!

redhatrises:
@xhudik this is great work! Will you please remove `ubuntu1804/overlays/stig_overlay.xml` from this PR? The `stig_overlay.xml` is for `rhel6` and is invalid for `ubuntu`.

jan-cerny:
@redhatrises Are those `stig_overlay.xml` files actually used anywhere?

redhatrises:
> @redhatrises Are those `stig_overlay.xml` files actually used anywhere?

Yes. For those who do STIG mapping. Also at some point, we need to bring back `make stig` which is also a user of this overlay.

jan-cerny:
@xhudik  Hi Tomáš, This is a great job.

I think next step is to go trough the rules included in this profile and enable OVAL checks, bash scripts, etc. for Ubuntu for these rules if they're not enabled yet.

You can get inspired eg. by https://github.com/ComplianceAsCode/content/pull/3289 , https://github.com/ComplianceAsCode/content/pull/3243 , https://github.com/ComplianceAsCode/content/pull/3336   https://github.com/ComplianceAsCode/content/pull/3258 where we have enabled some rules on Fedora which were only in RHEL7 before.

PR #3341, opened 2018-09-19
Update shell command in shared.yml

Mixer9:
#### Description:

Good idea by @jan-cerny to add the ```--nofiledigest``` to this files rpm command as well.  The file digest is not needed nor useful in this command for the purpose of getting the list of files.

#### Rationale:

The file digest is not needed nor useful in this command for the purpose of getting the list of files.

- Fixes # 3328


jan-cerny:
@Mixer9 Thank you!

PR #3345, opened 2018-09-19
update shared.yml

Mixer9:
#### Description:

Return last field no matter the number of fields

#### Rationale:

This change has the same effect as #3328 and potentially could return empty values based on number of field differences, using awk and printing the last field no matter how many fixes this issue.


jan-cerny:
@openscap-ci ok to test

jan-cerny:
The new command works great. Thank you.

PR #3346, opened 2018-09-19
update shared.sh

Mixer9:
#### Description:

Return last field no matter the number of fields

#### Rationale:

This change has the same effect as #3328 and potentially could return empty values based on number of field differences, using awk and printing the last field no matter how many fixes this issue.


cipherboy:
@openscap-ci add to whitelist.

PR #3347, opened 2018-09-19
Created Ansible shared.yml

Mixer9:
Created ansible playbook to correct ownership of files with rpm.

#### Description:

Correct file ownership using rpm

#### Rationale:

Use Ansible to correct file ownership using rpm. The vendor should maintain ownership settings,  files that the ownership has changed should be investigated.

### DIFF
--- a/linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_ownership/ansible/shared.yml
+++ b/linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_ownership/ansible/shared.yml
@@ -4,7 +4,7 @@
 # complexity = high
 # disruption = medium
 - name: "Read list of files with incorrect ownership"
-  shell: "rpm -Va --nofiledigest | grep "^.....\(U\|.G\)" | awk '{print $NF}'"
+  shell: rpm -Va --nofiledigest | grep "^.....\(U\|.G\)" | awk '{print $NF}'
   register: files_with_incorrect_ownership
   failed_when: False
   changed_when: False
--- a/linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_ownership/ansible/shared.yml
+++ b/linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_ownership/ansible/shared.yml
@@ -1,0 +1,27 @@
+# platform = multi_platform_rhel,multi_platform_fedora
+# reboot = false
+# strategy = restrict
+# complexity = high
+# disruption = medium
+- name: "Read list of files with incorrect ownership"
+  shell: "rpm -Va --nofiledigest | grep "^.....\(U\|.G\)" | awk '{print $NF}'"
+  register: files_with_incorrect_ownership
+  failed_when: False
+  changed_when: False
+  check_mode: no
+  tags:
+    @ANSIBLE_TAGS@
+    
+- name: Create list of uniq packages
+  shell: "rpm -qf {{ files_with_incorrect_ownership.stdout_lines }}|sort |uniq"
+  register: uniq_list_of_packages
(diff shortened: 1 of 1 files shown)


linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_ownership/ansible/shared.yml
```diff
@@ -0,0 +1,27 @@
...
+- name: "Read list of files with incorrect ownership"
+  shell: "rpm -Va --nofiledigest | grep "^.....\(U\|.G\)" | awk '{print $NF}'"
+  register: files_with_incorrect_ownership
+  failed_when: False
+  changed_when: False
+  check_mode: no
+  tags:
+    @ANSIBLE_TAGS@
+    
+- name: Create list of uniq packages
```
redhatrises: I don't think that we need to create a unique list of packages. Should be all.

Mixer9:
The reason you need it to be uniq is because the list of files potentially
could have multiple files ( hundreds ) that come from the same package, if
you don't uniq it the setperms command will be run multiple times against
the same package. No need to do that.

Mixer9

linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_ownership/ansible/shared.yml
Mixer9: The reason you need it to be uniq is because the list of files potentially
could have multiple files ( hundreds ) that come from the same package, if
you don't uniq it the setperms command will be run multiple times against
the same package. No need to do that.

mpreisler:
@openscap-ci add to whitelist

linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_ownership/ansible/shared.yml
redhatrises: Actually, there is a reason to do it. While they are called the same name e.g. `zlib`, they often install different files when talking about different architecture installs e.g. `i686`, `x86_64`, `noarch`, etc. Also, the `uniq` doesn't work because `rpm -qf` correctly returns the file name. For example:

```
# rpm -qa | grep zlib-1
zlib-1.2.11-8.fc28.x86_64
zlib-1.2.11-8.fc28.i686

# rpm -Va --nodigest | grep libz
.....U...    /usr/lib64/libz.so.1.2.11
.....U...    /usr/lib/libz.so.1.2.11

# rpm -qf /usr/lib/libz.so.1.2.11 /usr/lib64/libz.so.1.2.11
zlib-1.2.11-8.fc28.i686
zlib-1.2.11-8.fc28.x86_64
```

jan-cerny:
Hi, the CI check reports a syntax error. Please fix quotes in your patch.
```
ERROR! Syntax Error while loading YAML.
  did not find expected key

The error appears to have been in '/home/jenkins/workspace/scap-security-guide-pull-requests/build/roles/ssg-rhel7-role-stig-rhel7-disa.yml': line 8219, column 46, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

    - name: "Read list of files with incorrect ownership"
      shell: "rpm -Va --nofiledigest | grep "^.....\(U\|.G\)" | awk '{print $NF}'"
                                             ^ here
We could be wrong, but this one looks like it might be an issue with
unbalanced quotes.  If starting a value with a quote, make sure the
line ends with the same set of quotes.  For instance this arbitrary
example:

    foo: "bad" "wolf"

Could be written as:

    foo: '"bad" "wolf"'
```
```

jan-cerny:
@redhatrises I think @Mixer9 is correct here, `rpm -qf` returns the full package name including the architecture.

jan-cerny:
@openscap-ci test this please

Issue #3344, opened 2018-09-19
STIG AMBIGUITY - RHEL-07-040720 - If the Trivial File Transfer Protocol (TFTP) server is required, the TFTP daemon must be configured for secure mode


linuxdan:
#### Description of problem:
STIG says:
Check Text: Verify the TFTP daemon is configured to operate in secure mode.
Check to see if a TFTP server has been installed with the following commands:

yum list installed | grep tftp-server

tftp-server.x86_64 x.x-x.el7 rhel-7-server-rpms

If a TFTP server is not installed, this is Not Applicable.
If a TFTP server is installed, check for the server arguments with the following command: 

grep server_arge /etc/xinetd.d/tftp  

server_args = -s /var/lib/tftpboot

If the "server_args" line does not have a "-s" option and a subdirectory is not assigned, this is a finding.
Fix Text: Configure the TFTP daemon to operate in secure mode by adding the following line to "/etc/xinetd.d/tftp" (or modify the line to have the required value):
server_args = -s /var/lib/tftpboot

For Red Hat Satellite, that line in that file is 
server_args     = -v -s /var/lib/tftpboot -m /etc/tftpd.map

Tenable, in their infinite brainlessness, fail that because it is not an exact match

The file itself cannot be permanently changed because it is hard-coded thru Puppet and Forman and goodness-knows-what-else.

Other than fixing the typo in the grep line, the wording needs to change to permit variations.
Maybe saying something like "server_args = [other options] -s /var/lib/tftpboot [more options]"
I am shooting from the hip/in the dark on this.

#### STIG Version:
Red Hat Enterprise Linux 7 Security Technical Implementation Guide :: Release: 4 Benchmark Date: 26 Jan 2018

trevor-vaughan:
@linuxdan I think that the text is perfectly clear.

If using `tftpd`, you must have a `-s <directory>` entry in the options string.

If Tenable is triggering this as a finding, then that is a bug in their software, not the prose text.

linuxdan:
They say that they are doing exactly what the STIG tells them to do. 
Which, to them, means exactly matching what the STIG says should be returned from the Greg command. 
Tenable won’t budge, so I am approaching the issue from the other end.

shawndwells:
I think there's a legitimate bug in the SSG content over this.

While the XCCDF says ``-s`` should be an option, the OVAL is looking for an exact match:
[...]

PR #3348, opened 2018-09-19
Fix for issue 3344

linuxdan:
Rather than only matching this line, exactly,
`server_args = -s /var/lib/tftpboot`
allow for additional arguments in the front like I found in a Red Hat Satellite server
`server_args = -v -s /var/lib/tftpboot -m /etc/tftpd.map`

#### Description:
This change is caused by how Tenable has coded Nessus to scan for this.
They are looking for an exact string rather than specific arguments in a configuration line.

#### Rationale:
Change the regex to allow for arguments between the equal sign and the "-s"

- Fixes # 3344


shawndwells:
ok to test

shawndwells:
(links to https://github.com/ComplianceAsCode/content/issues/3344)

shawndwells:
ack pending jenkins

shawndwells:
Congrats on your first commit, @linuxdan! This will be part of our next upstream release. Downstreams (e.g. RHEL) will pick it up during their next rebase.

PR #3351, opened 2018-09-20
Tests and fixes for gnome screensaver lock delay

yuumasato:
#### Description:

- Tests for `dconf_gnome_screensaver_lock_enabled`
- Fix `dconf_settings` function not to accept commented lines
- Reload dconf settings every time we change it


tests/data/group_system/group_software/group_gnome/group_gnome_screen_locking/rule_dconf_gnome_screensaver_lock_enabled/comment.fail.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+# profiles = xccdf_org.ssgproject.content_profile_ospp
+
+yum -y install dconf
+
+# It is ok if string is not found in any file
+file=$(grep -R "lock-enabled" /etc/dconf/db/local.d) || true
```
mpreisler: nack, we need to test using acutal dconf, not checking files. let's do real integration tests here. Look up dconf read

dahaic:
Works great @yuumasato . Both tests and fix, thanks!

PR #3352, opened 2018-09-20
Add SSSD rule to enable smart cards and add smart card rules to RHEL7 OSPP

redhatrises:
This PR replaces existing smart card rules (mainly `smartcard_auth` since it is now deprecated) in the RHEL7 OSPP profile with the new OpenSC/SSSD integration.

### DIFF
--- a/linux_os/guide/services/sssd/sssd_enable_smartcards/oval/shared.xml
+++ b/linux_os/guide/services/sssd/sssd_enable_smartcards/oval/shared.xml
@@ -1,0 +1,32 @@
+<def-group oval_version="5.11">
+  <definition class="compliance" id="sssd_enable_smartcards" version="1">
+    <metadata>
+      <title>Enable Smartcards in SSSD</title>
+      <affected family="unix">
+        <platform>Red Hat Enterprise Linux 7</platform>
+        <platform>multi_platform_fedora</platform>
+      </affected>
+      <description>SSSD should be configured to authenticate access to the system
+    using smart cards.</description>
+    </metadata>
+    <criteria operator="OR">
+      <criteria operator="OR">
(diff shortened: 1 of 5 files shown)


dahaic:
@redhatrises is there any OSPP reference to be added to the rules you add in the profile? What is the justification? I would prefer to have it in the profile in the form of comment, because when anybody asks "why is this in the profile" there would be more than just `git blame` -> PR without description :)

redhatrises:
@dahaic there are no OSPP references for the `ospp.profile`. There are for `ospp42`. All these rules are doing is replacing the `smartcard_auth` rule as it is deprecated.

dahaic:
Thank you for the explanation @redhatrises I'll take a look tomorrow.

linux_os/guide/services/sssd/sssd_enable_smartcards/oval/shared.xml
```diff
@@ -0,0 +1,32 @@
+<def-group oval_version="5.11">
+  <definition class="compliance" id="sssd_enable_smartcards" version="1">
+    <metadata>
+      <title>Enable Smartcards in SSSD</title>
+      <affected family="unix">
+        <platform>Red Hat Enterprise Linux 7</platform>
```
mpreisler: the fixes use multi_platform_rhel and this uses RHEL7, which is the correct one?

redhatrises: Thanks @mpreisler. Should be `RHEL7` and not `multi_platform_rhel`. I wasn't thinking here.

redhatrises:
@openscap-ci test this please

dahaic:
@openscap-ci test this please

dahaic:
Thank you @redhatrises, ACK

PR #3353, opened 2018-09-20
Select missing audit privileged commands rules for OSPP42

yuumasato:
#### Description:

Select audit rules privileged commands for binaries below:
- crontab
- umount

#### Rationale:

- OSPP4.2 would like you to have execution of these binaries


dahaic:
ACK, thanks @yuumasato

PR #3355, opened 2018-09-21
Add a warning of missing remediation

jan-cerny:
#### Description:
Add warnings that remediation is not available.
This will be visible in HTML report therefore it will help users to identify why those rules haven't been fixed automatically.

#### Rationale:
These rules are a part of Fedora OSPP profile.


mildass:
Thanks

PR #3356, opened 2018-09-21
Enable OVAL for rule grub2_enable_fips_mode on Fedora

jan-cerny:
#### Description:
I change the condition that the installed OS must be RHEL 7
to a check that installed OS is certified, because there
can be multiple operating systems that are FIPS certified,
not only RHEL 7. However, the rule will always fail on
Fedora.  I also add Fedora to the disclaimer.  I had to to
add `dracut-fips` to the list of installed packages,
otherwise the OVAL definition package_dracut-fips_installed
that extends this OVAL definition would be missing and that
would lead to the fact that the whole OVAL definition will
be stripped by our build system.

#### Rationale:

### DIFF
--- a/linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/anaconda/shared.anaconda
+++ b/linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/anaconda/shared.anaconda
@@ -1,3 +1,3 @@
-# platform = multi_platform_rhel
+# platform = Red Hat Enterprise Linux 7, multi_platform_fedora
 
 package --add=dracut-fips
--- a/linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/bash/shared.sh
+++ b/linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/bash/shared.sh
@@ -1,4 +1,4 @@
-# platform = Red Hat Enterprise Linux 7
+# platform = Red Hat Enterprise Linux 7, multi_platform_fedora
 
 # include remediation functions library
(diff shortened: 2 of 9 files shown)


fedora/templates/csv/packages_installed.csv
```diff
@@ -3,6 +3,7 @@ audit
 chrony
 cronie
 dconf
+dracut-fips
```
redhatrises: This should not be enabled here. Please remove this. All changes for dracut-fips should be made here. `linux_os/guide/system/software/integrity/fips/package_dracut-fips_installed`

shawndwells:
This is a really hard call. The check is about enabling FIPS mode, which is a legitimate configuration check , but combined with the ``ospp`` profile gives the sense that Fedora is FIPS certified.

How can we separate conceptions that Fedora is FIPS certified?

dahaic:
@shawndwells we can try to be even more explicit with the description of the OSPP profile than we are now.

jan-cerny:
> In regards to updating the platform tags, Fedora does not have FIPS evaluations. This check should return failed, or at least notapplicable, when running on Fedora.

This will surely happen. The check always returns `fail` on Fedora, because it depends on `installed_OS_is_certified`, which always fails on Fedora. `installed_OS_is_certified` passes only on RHEL6 and RHEL7.

fedora/templates/csv/packages_installed.csv
jan-cerny: @redhatrises fixed

jan-cerny:
I have changed it to not use the template for package_dracut-fips_installed.

dahaic:
Thank you @jan-cerny, ACK

PR #3357, opened 2018-09-24
Add checks for removals of abrt and sendmail to OSPP

jan-cerny:
#### Description:

Adds rules that check that packages abrt and sendmail are removed to Fedora OSPP and RHEL7 OSPP 4.2 profiles.

#### Rationale:
These rules should be part of OSPP.

### DIFF
--- a/linux_os/guide/services/base/group.yml
+++ b/linux_os/guide/services/base/group.yml
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: rhel6, rhel7, fedora
+prodtype: rhel6,rhel7,fedora
 
 title: 'Base Services'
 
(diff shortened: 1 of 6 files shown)


linux_os/guide/services/base/group.yml
```diff
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: rhel6, rhel7
+prodtype: rhel6, rhel7, fedora
```
redhatrises: @jan-cerny can you remove spaces here? For some reason, they have started to appear in the codebase.

jan-cerny: ok

dahaic:
Thank you @jan-cerny

PR #3360, opened 2018-09-25
Add mount options /dev/shm to OSPP profiles

jan-cerny:
#### Description:
This commit adds rules of /dev/shm mount options to Fedora OSPP
profile and RHEL 7 OSPP 4.2 profile.
The template for Fedora will not generate Anaconda remediations
because the same limitation is set in RHEL 7 rules.

#### Rationale:
These rules should be included in OSPP.


dahaic:
Thank you @jan-cerny . ACK

PR #3362, opened 2018-09-25
Fix file permissions for rule_file_permissions_sshd_private_key

matusmarhefka:
#### Description:

RHEL, CentOS and Fedora ssh private key files should be `0640`. This is because `ssh_keys` group owns ssh private key files in `/etc/ssh` and also `ssh-keysign` binary:
```
$ ls -l /usr/libexec/openssh/ssh-keysign 
-r-xr-sr-x. 1 root ssh_keys 461672 Aug 25 14:31 /usr/libexec/openssh/ssh-keysign
$ ls -l /etc/ssh/ | grep rsa
-rw-r-----. 1 root ssh_keys   1675 Jun  5 19:30 ssh_host_rsa_key
-rw-r--r--. 1 root root        382 Jun  5 19:30 ssh_host_rsa_key.pub
```
Other products should still use `0600` for ssh private key files


redhatrises:
Ack.

PR #3365, opened 2018-09-25
OSPP profile title change + description reformatting

matejak:
This PR renames OSPP 4.2-related profiles, so they feature OSPP in their title.

A profile's description has been reformatted to use less special characters and achieve the same result.


dahaic:
Thank you @matejak

PR #3366, opened 2018-09-25
Fix an underscore

jan-cerny:
#### Description:
The correct name of the variable is pam_file.

#### Rationale:
Referencing non-existing variable is a bug.


jan-cerny:
Yes, that is probably why we haven't cached the problem before.

jan-cerny:
I have added a new test scenario. I hope it is possible to test it this way.

yuumasato:
Thanks, the test works fine, it fails for datastream without this fix.
I noticed that it tests both `preauth` and `authfail` functions in one scenario, but I don't this is a big deal.
Ack!

PR #3367, opened 2018-09-26
Initial implementation of the stable profile ID test

mpreisler:
This will prevent people from renaming or removing profiles we deem
important ("stable"). I am hoping this will help our users, especially
downstream users that run continuous compliance deployments and want to
avoid changing their configs every time we ship an update.

Please don't try to make me add all your beloved profiles in this PR :-D. If we agree on the general idea here we can promote profiles to "stable" in later PRs.

### DIFF
--- a/tests/stable_profile_ids.py
+++ b/tests/stable_profile_ids.py
@@ -1,0 +1,95 @@
+#!/usr/bin/env python3
+
+from __future__ import print_function
+
+import os
+import ssg.xml
+import ssg.xccdf
+import ssg.build_guides
+import argparse
+import glob
+from collections import defaultdict
+
+
+# Maps shortened benchmark IDs to list of profiles that need to be contained
+# in said benchmarks. Delete the predictable prefix of each benchmark ID to get
+# the shortened ID. For example xccdf_org.ssgproject.content_benchmark_RHEL-7
+# becomes just RHEL-7. Apply the same to profile IDs:
+# xccdf_org.ssgproject.content_profile_ospp42 becomes ospp42
+
+STABLE_PROFILE_IDS = {
+    "RHEL-6": ["pci-dss"],
+    "RHEL-7": ["ospp", "ospp42", "pci-dss"],
+}
+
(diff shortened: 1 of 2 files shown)


mpreisler:
Also see #3321 for more context.

tests/stable_profile_ids.py
```diff
@@ -0,0 +1,95 @@
...
+PROFILE_ID_PREFIX = "xccdf_org.ssgproject.content_profile_"
+
+
+def parse_args():
+    p = argparse.ArgumentParser()
+
+    p.add_argument("build_dir", type=str,
+                   help="Directory with the datastreams that will be checked "
+                        "for stable profile IDs. All files matching "
+                        "$BUILD_DIR/ssg-*-ds.xml checked.")
```
jan-cerny: We can check only the products that have the stable profile IDs specified.

mpreisler: not sure what you are getting at. could you please outline the change you are expecting?

jan-cerny: For example we can get a list of keys from STABLE_PROFILE_IDS dictionary and transform the list to the list of datastream file names.

mpreisler: I don't see a tangible benefit. It's a different approach to the same thing...

jan-cerny: The tangible benefit is that you will open only RHEL 6 and RHEL 7 datastreams.

mpreisler: So you want to avoid the XML parse of all the datastreams? Isn't it our expectation that we will have stable profiles in all the datastreams we ship in the future?

mpreisler:
```
mpreisle@t450s ~/d/scap-security-guide $ time python3 ./tests/stable_profile_ids.py ./build

real    0m2.889s
user    0m2.701s
sys     0m0.172s
```

mpreisler: You are of course free to optimize this later ;-)

jan-cerny:
The test works if we break a profile.

PR #3390, opened 2018-09-26
Update platform in OVAL for derivatives

redhatrises:
#### Description:

- Downstream does a sed on the content that creates numerous content bugs. This should solve the
  problem for OVAL content.


shawndwells:
What a nice community steward you are @redhatrises

PR #3393, opened 2018-09-27
Added more stabilized profile IDs

dahaic:
#### Description:

- Add more profile IDs that should not be changed anymore

#### Rationale:

- Changing profiles you use can be pain, if it's hardcoded into automation. This way, we will ensure we don't break this backwards compatibility by accident.

### DIFF
--- a/tests/stable_profile_ids.py
+++ b/tests/stable_profile_ids.py
@@ -18,8 +18,11 @@ from collections import defaultdict
 # xccdf_org.ssgproject.content_profile_ospp42 becomes ospp42
 
 STABLE_PROFILE_IDS = {
-    "RHEL-6": ["pci-dss"],
-    "RHEL-7": ["ospp", "ospp42", "pci-dss"],
+    "FEDORA": ["standard"],
+    "RHEL-6": ["C2S", "CS2", "CSCF-RHEL6-MLS", "fisma-medium-rhel6-server",
+               "pci-dss", "rht-ccp", "stig-rhel6-disa", "usgcb-rhel6-server"],
+    "RHEL-7": ["C2S", "cjis", "hipaa", "nist-800-171-cui", "rht-ccp",
+               "ospp", "ospp42", "pci-dss", "stig-rhel7-disa"],
 }
 
 BENCHMARK_ID_PREFIX = "xccdf_org.ssgproject.content_benchmark_"


shawndwells:
Also, why are there two OSPP profiles in rhel7? ospp vs ospp42

dahaic:
Thank you for the note regarding `nist-CL-IL-AL`. For the rhel7 ospp - original one is USGCB compile of several guidelines, mainly OSPP v4.0. ospp42 is profile following OSPP v4.2 specification.
I believe v4.0 is still followed by some organizations, thus we do not replace, but have two profiles side by side.

redhatrises:
> Thank you for the note regarding `nist-CL-IL-AL`. For the rhel7 ospp - original one is USGCB compile of several guidelines, mainly OSPP v4.0. ospp42 is profile following OSPP v4.2 specification.
> I believe v4.0 is still followed by some organizations, thus we do not replace, but have two profiles side by side.

This `ospp4.2` profile should become the `ospp` profile, but that is out of scope of this PR.

redhatrises:
ACK.

shawndwells:
On 9/27/18 2:20 PM, Marek Haičman wrote:
> Thank you for the note regarding |nist-CL-IL-AL|. For the rhel7 ospp - 
> original one is USGCB compile of several guidelines, mainly OSPP v4.0. 
> ospp42 is profile following OSPP v4.2 specification.
> I believe v4.0 is still followed by some organizations, thus we do not 
> replace, but have two profiles side by side.

There really should be just one OSPP profile, which gets updated to 
align to new OSPP releases from NIAP. Definitely don't want to ship 
multiple ones -- way to confusing.

jan-cerny:
@dahaic @shawndwells @redhatrises 

I am confused by this. Can you clarify if we plan to have both `ospp42` and `ospp` as a stable profile IDs? If one of them is planned to be removed, then it should be removed from the test case.

mpreisler:
I have the same question as @jan-cerny 

Furthermore:

> This ospp4.2 profile should become the ospp profile, but that is out of scope of this PR.

This is exactly in scope of this PR :) The idea is that once the profile becomes stable we *never* rename/remove it.

dahaic:
AFAIK, not every user of OSPP follows "latest greatest". I.e. some has to stay with OSPP 4.0 for the certification to work. Thus there probably won't be rename of `ospp42` to `ospp`.

PR #3399, opened 2018-09-28
Fix permission templates to allow non-existing filename

redhatrises:


shawndwells:
not to self: ``if not file_name`` couldn't be used here because if it's not null or empty the script will think that it is a file and generate an OVAL to check a file rather than a filepath (thanks to @redhatrises for the schooling)

PR #3414, opened 2018-09-28
Rename scheduler ocp rules

redhatrises:
#### Description:

- Renames OCP scheduler rules to reduce current and future rule naming conflicts


shawndwells:
thanks for cleaning this up!

merging this since the build is broken for other reasons
(will take a look later tonight!)

PR #3492, opened 2018-10-01
SSG test suite trivial fix.

matejak:
Fix `map_on_top` for the case when there is nothing to map.

### DIFF
--- a/tests/ssg_test_suite/test_env.py
+++ b/tests/ssg_test_suite/test_env.py
@@ -19,6 +19,8 @@ class SavedState(object):
         self.initial_running_state = True
 
     def map_on_top(self, function, args_list):
+        if not args_list:
+            return
         current_running_state = self.initial_running_state
         function(* args_list[0])
         for idx, args in enumerate(args_list[1:], 1):


matejak:
It happens when the function argument `args_list` is an empty list.

jan-cerny:
@matejak And in which situation args_list is an empty list?

matejak:
It already happened, but it is quite rare to trigger this condition with the current code/data. But, more importantly, Python3.6's `map(lambda x: print(f"hello {x}"), [])` doesn't crash, it just won't print anything. Therefore, our `map_on_top` probably shouldn't crash either.

mpreisler:
I agree with @jan-cerny , IMO we want to figure out when/why this even happens. It's obvious that the error happens when args_list is an empty list, the question is why that ever happens. We should fix the root cause, not quickfix the symptom.

matejak:
@jan-cerny Yes, this is the most likely cause.
@mpreisler There may not be any issue that would have that root cause - in the future, we may e.g. filter scenarios that are applicable for containers. In that case, it may happen that there are no scenarios left, and it is totally legitimate. A check for argument count is therefore needed somewhere. I find the body of `map_on_top` to be the best place for that:
- let's be consistent with Python's `map`, and 
- let's check on one place instead of checking before every `map_on_top` invocation.

jan-cerny:
@matejak But the directory has to exist to trigger it?

jan-cerny:
But if the directory exists it fixes it perfectly for me.

PR #3513, opened 2018-10-04
Fix `utils/rule_dir_json.py`

cipherboy:
This fixes the parsing of OVALs by `rule_dir_json` & friends; in particular, we now parse JINJA macros before loading the contents of the OVAL.

This also adds it to the cmake tests so it hopefully won't get broken in the future... :)

### DIFF
--- a/ssg/oval.py
+++ b/ssg/oval.py
@@ -13,6 +13,7 @@ from .constants import oval_namespace as ovalns
 from .rules import get_rule_dir_id, get_rule_dir_ovals, find_rule_dirs
 from .xml import ElementTree as ET
 from .xml import oval_generated_header
+from .yaml import process_file
 from .id_translate import IDTranslator
 
 SHARED_OVAL = re.sub(r'ssg/.*', 'shared', __file__) + '/checks/oval/'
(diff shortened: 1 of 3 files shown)


cipherboy:
@mpreisler CI times might go up, but I'm not sure if this is because I'm slamming my disks/cpu when I test:

```
        Start  39: test-rule-dir-json
118/221 Test  #39: test-rule-dir-json .......................................................   Passed   33.26 sec
```

ssg/oval.py
```diff
@@ -97,7 +98,7 @@ def applicable_platforms(oval_file):
 
     platforms = []
     header = oval_generated_header("applicable_platforms", "5.11", "0.0.1")
-    body = read_ovaldefgroup_file(oval_file)
+    body = process_file(oval_file, {})
```
cipherboy: To clarify why times to run this script have increased...

We're now processing every single OVAL file with JINJA macros (no cache) before trying to parse the `affected/platforms` tag. It's a big hit relative to the script (iirc 5s -> 30s), but it is still much quicker than a full build of all the content. At some point in the future, this could be dropped if someone wants to write a dirty line-based XML parser, but that's a bit more work.

mpreisler: It's necessary in the general case because you could theoretically use jinja code to emit platform elements.

cipherboy: But in that case, you'd need to process the OVAL with each of the platform YAML files and detect if the applicable platform changes with each call... Because you could use that same JINJA and condition on the product variable in the `product.yml` file... So it'd get really messy and I'd recommend not doing that.

mpreisler: Yeah, I see your point. Probably best to say that it's not supported because it's incredibly difficult to support.

cipherboy: And we'd need to support it in the fix content and the YAML files as well... Which gets awfully messy awfully quickly.

PR #3519, opened 2018-10-09
Improve robustness of ensure_gpgcheck_never_disabled Bash fix

yuumasato:
#### Description:

- Allow Bash fix to match spaces between "gpgcheck" and "="

#### Rationale:

- The `repo` file can be configure with spaces between key and value.


shawndwells:
thanks, @yuumasato !

PR #3520, opened 2018-10-09
Create /dev/shm entry in /etc/fstab if needed

yuumasato:
#### Description:

- Generate remediation with capability to add  `/etc/fstab` entry if needed.

#### Rationale:

- `/dev/shm` may not be in `/etc/fstab`, add entry if needed.
- One of the mount options rules `noexec`, `nosuid`, `nodev` for `/dev/shm` always failed, while the other two succeeded.


jan-cerny:
looks good to me

PR #3529, opened 2018-10-12
Add profiles to man page automatically

jan-cerny:
#### Description:
Adds a script that will generate the man page during build from the data in datastreams.

#### Rationale:

The list of profiles in man page gets outdated, because people sometimes
don't add new profiles or products to man page. Also they don't update
their descriptions in the man page. We would like to avoid outdated
information and we would like to avoid manual work. I propose to
generate the list of profiles and their descriptions by an automated
script which will be run as a part of build process. This will also
solve the problem that we have on RHEL where the man page describes
Fedora profiles which are not shipped. Only the products that are
built will be added to man page.


build-scripts/generate_man_page.py
```diff
@@ -0,0 +1,208 @@
...
+import ssg.build_profile
+import ssg.constants
+import ssg.xml
+import os
+
+import xml.etree.ElementTree as ET
+
+prolog = """
+.TH scap-security-guide 8 "26 Jan 2013" "version 1"
+
```
jan-cerny: We will need to update also this text.

redhatrises:
@jan-cerny this is awesome work! There is a `ValueError` that gets raised:
```
  File "/home/jenkins/workspace/scap-security-guide-pull-requests/label/rhel6/build-scripts/generate_man_page.py", line 208, in <module>
    main()
  File "/home/jenkins/workspace/scap-security-guide-pull-requests/label/rhel6/build-scripts/generate_man_page.py", line 161, in main
    print_all_profiles(input_dir, output_file)
  File "/home/jenkins/workspace/scap-security-guide-pull-requests/label/rhel6/build-scripts/generate_man_page.py", line 168, in print_all_profiles
    print_profiles_in_ds(os.path.abspath(item), output_file)
  File "/home/jenkins/workspace/scap-security-guide-pull-requests/label/rhel6/build-scripts/generate_man_page.py", line 177, in print_profiles_in_ds
    output_file.write(".SH\n{}\n\n".format(benchmark_title))
ValueError: zero length field name in format
```

matejak:
This is a Python2.6. You can e.g. use `format` with kwargs.

build-scripts/generate_man_page.py
```diff
@@ -0,0 +1,208 @@
+#!/usr/bin/env python2
```
mpreisler: why python2?

build-scripts/generate_man_page.py
```diff
@@ -0,0 +1,208 @@
...
+from __future__ import print_function
+
+import argparse
+import sys
+import ssg.build_profile
+import ssg.constants
+import ssg.xml
+import os
+
+import xml.etree.ElementTree as ET
```
mpreisler: we have an elementtree wrapper in the ssg module, don't import it like this, there are multiple of them and the wrapper will make sure you are using the right one

build-scripts/generate_man_page.py
mpreisler: IMO it would be much better to have a jinja2 template for this outside of the script, it would make it much easier to edit and much easier to add more logic

mpreisler:
We just discussed this with @jan-cerny , this is a great use-case for jinja2 templating. Would make it much easier to edit the template and add more logic.

build-scripts/generate_man_page.py
mpreisler: image

build-scripts/generate_man_page.py
```diff
@@ -10,7 +10,10 @@
 import os
 import io
 
-import xml.etree.ElementTree as ET
+try:
+    from xml.etree import cElementTree as ElementTree
+except ImportError:
+    import cElementTree as ElementTree
```
mpreisler: ?!?

Just import ssg.xml and get the ElementTree from there. Definitely don't copy paste code!

https://github.com/ComplianceAsCode/content/blob/master/ssg/xml.py

build-scripts/generate_man_page.py
```diff
@@ -9,13 +9,14 @@
 import ssg.xml
 import os
 import io
+from jinja2 import Template
```
mpreisler: we have wrappers and code for this, I suggest you use the wrappers instead of rolling your own. The advantage is that you will be able to pass all of build_config and other SSG specific things.

build-scripts/generate_man_page.py
jan-cerny: Aha, I'll do that. I'll provide a comment to prevent readability problems.

build-scripts/generate_man_page.py
jan-cerny: That sounds awesome!

docs/man_page_template.jinja
```diff
@@ -0,0 +1,134 @@
...
+high-level and widely-accepted policies such as NIST 800-53 provides prose
+stating that System Administrators must audit "privileged user actions," but do
+not define what "privileged actions" are. The SSG bridges the gap between
+generalized policy requirements and specific implementation guidance, in SCAP
+formats to support automation whenever possible.
+
+The projects homepage is located at:
+https://www.open-scap.org/security-policies/scap-security-guide
+
+{{{ profiles }}}
```
mpreisler: Instead of just concatenating profiles here consider using jinja2's `for`. See http://jinja.pocoo.org/docs/2.10/templates/

That way you can keep the formatting in the actual template.

build-scripts/generate_man_page.py
```diff
@@ -0,0 +1,75 @@
...
+        "{%s}title" % (ssg.constants.XCCDF12_NS)).text
+    profiles_in_ds += u".SH\nProfiles in %s\n\n" % (benchmark_title)
+    ds_filename = os.path.basename(ds_filepath)
+    profiles_in_ds += u"Source Datastream: \\fI %s\\fR\n\n" % (ds_filename)
+    profiles_in_ds += u"The %s is broken into 'profiles', groupings of security settings " \
+        "that correlate to a known policy. Available profiles are:\n\n" % (
+            benchmark_title)
+    profiles = get_profiles_info(benchmark)
+    for profile_id, title, description in profiles:
+        profiles_in_ds += u".B %s\n\n.RS\nProfile ID: \\fI%s\\fR\n\n%s\n.RE\n\n\n" % (
```
mpreisler: this text formatting stuff should be in the jinja2 template. You can just pass in a structure and let jinja2 run the for and output this for every line.

docs/man_page_template.jinja
jan-cerny: That sounds great, I'll look into this possibility.

jan-cerny: That sounds great, I'll look into this possibility.

jan-cerny: That sounds great, I'll look into this possibility.

jan-cerny: That sounds great, I'll look into this possibility.

jan-cerny: That sounds great, I'll look into this possibility.

jan-cerny:
I have changed the logic to define formatting in the template itself, not in the Python script.

build-scripts/generate_man_page.py
```diff
@@ -0,0 +1,75 @@
...
+
+def get_all_profiles(input_dir):
+    all_profiles = ""
+    for item in sorted(os.listdir(input_dir)):
+        if item.endswith("-ds.xml"):
+            all_profiles += get_profiles_in_ds(os.path.join(input_dir, item))
+    return all_profiles
+
+
+def get_profiles_in_ds(ds_filepath):
```
mpreisler: This function is unused, why is it still there?

jan-cerny:
I have removed the commits that improved the text and I have removed the commits that removed the outdated information from this PR.

PR #3532, opened 2018-10-15
Rename controller and api-server rules

redhatrises:
- Standardizes on using an underscore rather than mixed dash and underscore
- Breaks a part certificate rules into separate rules for CCE configuration requirements


shawndwells:
Sweet! Thanks for renaming these and updating the profile.

PR #3536, opened 2018-10-16
Add newline in between tasks to increase readability

redhatrises:
#### Description:

- Adding a newline in between tasks in playbooks not only increases readability but is encouraged 
  by Ansible.


mpreisler:
The comment is actually correct. However the code is fairly fragile, I am wondering if there is a better way to do this.

redhatrises:
> The comment is actually correct. However the code is fairly fragile, I am wondering if there is a better way to do this.

@mpreisler have ideas?

shawndwells:
Functional code is better than none. Can revisit as needed.

Thanks @redhatrises

PR #3541, opened 2018-10-22
[wip] initial OVAL for OCP Controller

shawndwells:


applications/openshift/controller/controller_address/oval/shared.xml
```diff
@@ -0,0 +1,25 @@
+<def-group>
+  <definition class="compliance" id="controller_address" version="1">
```
redhatrises: This OVAL should also check the configuration in `/etc/origin/master/master-config.yaml`. I would follow what #3542 does. That why you check both runtime and file configuration like we do with the kernel or sysctl checks.

shawndwells:
incorporated this into another branch, closing.

PR #3543, opened 2018-10-22
Fix ocp_service_runtime_config id name to match template

redhatrises:
#### Description:

- `id` was incorrect and didn't match template naming.

### DIFF
--- a/shared/templates/create_ocp_service_runtime_config.py
+++ b/shared/templates/create_ocp_service_runtime_config.py
@@ -10,10 +10,15 @@ from template_common import FilesGenerator, UnknownTargetError
 
 class OCPServiceRuntimeConfigGenerator(FilesGenerator):
     def generate(self, target, ocp_process_info):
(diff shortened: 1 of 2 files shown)


shawndwells:
Certain checks will result in duplicate OVAL, e.g. those relating to --address.

Was originally thinking the OVAL rules could have the name of the PID, however this would result in very long names. Perhaps the template could be updated to manually assign the OVAL name?

shawndwells:
e.g. ``controller_address` and ``scheduler_address_argument`` would result in same OVAL (``ocp_process_runtime_address``)

redhatrises:
> e.g. `controller_address` and `scheduler_address_argument` would result in same OVAL (`ocp_process_runtime_address`)

You could also just easily add an `extend_definition` to the existing OVAL and be done with it;
however, you need to also check the `.yaml` file for proper configuration. In other words, it is not enough to just check the process runtime.

Issue #3544, opened 2018-10-23
Request for clarification - "interactive users"


linuxdan:
#### Description of problem:
In the section of the SSG labelled Account and Access Control, sub-section Secure Session Configuration Files for Login Accounts (and elsewhere) there are a number of rules that refer to "(local) interactive users".   Would it be safe to say that a login is "interactive" if the shell field from /etc/passwd exists in /etc/shells and is not "nologin" ?

#### SCAP Security Guide Version:
DISA STIG for Red Hat Enterprise Linux 7

#### Operating System Version:
Red Hat Enterprise Linux 7

shawndwells:
On 10/23/18 7:53 AM, Daniel E. White wrote:
>
>
>         Description of problem:
>
> In the section of the SSG labelled Account and Access Control, 
> sub-section Secure Session Configuration Files for Login Accounts (and 
> elsewhere) there are a number of rules that refer to "(local) 
> interactive users". Would it be safe to say that a login is 
> "interactive" if the shell field from /etc/passwd exists in 
> /etc/shells and is not "nologin" ?
>

"Interactive users" reflects accounts which can authenticate into an 
interactive shell, such as bash, upon logon. Generally equivalent to 
saying "these accounts are for humans who logon to the system."

Setting the shell to /bin/false, /sbin/nologin, or even /bin/true, is 
indicative that a given account is *not* for an interactive user. Our 
OVAL will also accept /bin/sync and /sbin/shutdown and /sbin/halt as 
indicative an account is non-interactive.

OVAL code that evaluates this:
https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/accounts/accounts-restrictions/root_logins/no_shelllogin_for_systemaccounts/oval/shared.xml#L77#L82

In being thorough, human/interactive accounts should have a UID greater 
than UID_MIN and less than UID_MAX as defined in /etc/login.defs. The 
OVAL checks UIDs outside this range and ensures they have a 
non-interactive shell set.

linuxdan:
Thanks for a complete and thorough response

linuxdan:
How about a dash of Python-Fu ?
I did it like this because, on RHEL 7 anyway, `etc/shells` contains
both `/sbin/nologin` and `/usr/sbin/nologin`
```python
#!/usr/bin/env python

import json

nono_shells = [ 'nologin', 'sync', 'shutdown', 'halt' ]

with open('/etc/shells') as shell_in:
[...]

PR #3547, opened 2018-10-23
add /usr/sbin/nologin to no_shelllogin_for_systemaccounts

shawndwells:
@linuxdan mentioned in https://github.com/ComplianceAsCode/content/issues/3544#issuecomment-432381417 that RHEL7 has both /sbin/nologin and /usr/sbin/nologin. He's right. Updating OVAL to accept either one.

### DIFF
--- a/linux_os/guide/system/accounts/accounts-restrictions/root_logins/no_shelllogin_for_systemaccounts/oval/shared.xml
+++ b/linux_os/guide/system/accounts/accounts-restrictions/root_logins/no_shelllogin_for_systemaccounts/oval/shared.xml
@@ -77,7 +77,7 @@
   <!-- Get all /etc/passwd entries having shell defined as OVAL object -->
   <ind:textfilecontent54_object id="object_etc_passwd_entries" version="1">
     <ind:filepath>/etc/passwd</ind:filepath>
-    <ind:pattern operation="pattern match">^(?!root).*:x:([\d]+):[\d]+:[^:]*:[^:]*:(?!\/sbin\/nologin|\/bin\/sync|\/sbin\/shutdown|\/sbin\/halt).*$</ind:pattern>
+    <ind:pattern operation="pattern match">^(?!root).*:x:([\d]+):[\d]+:[^:]*:[^:]*:(?!\/sbin\/nologin|\/usr\/sbin\/nologin|\/bin\/sync|\/sbin\/shutdown|\/sbin\/halt).*$</ind:pattern>
     <ind:instance datatype="int" operation="greater than or equal">1</ind:instance>
   </ind:textfilecontent54_object>
 


redhatrises:
> @linuxdan mentioned in [#3544 (comment)](https://github.com/ComplianceAsCode/content/issues/3544#issuecomment-432381417) that RHEL7 has both /sbin/nologin and /usr/sbin/nologin. He's right. Updating OVAL to accept either one.

There's actually really no difference between them as `/sbin` is now a softlink to `/usr/sbin` which is also true for Fedora.

shawndwells:
On 10/23/18 5:01 PM, redhatrises wrote:
>
> There's actually really no difference between them as |/sbin| is now a 
> softlink to |/usr/sbin| which is also true for Fedora.
>

That's what I thought to, especially given same filesize. But they look 
to be independent files (at least on RHEL 7):


$ ll /usr/sbin/nologin /sbin/nologin
-rwxr-xr-x. 1 root root 7248 Jul 12 05:51 /sbin/nologin
-rwxr-xr-x. 1 root root 7248 Jul 12 05:51 /usr/sbin/nologin

redhatrises:
> On 10/23/18 5:01 PM, redhatrises wrote: There's actually really no difference between them as |/sbin| is now a softlink to |/usr/sbin| which is also true for Fedora.
> That's what I thought to, especially given same filesize. But they look to be independent files (at least on RHEL 7): $ ll /usr/sbin/nologin /sbin/nologin -rwxr-xr-x. 1 root root 7248 Jul 12 05:51 /sbin/nologin -rwxr-xr-x. 1 root root 7248 Jul 12 05:51 /usr/sbin/nologin

This makes sense, and bet you they are the same. ;-) The symlink is over a directory, so the files will be regular files.

shawndwells:
On 10/23/18 5:30 PM, redhatrises wrote:
>
> This makes sense, and bet you they are the same. ;-) The symlink is 
> over a directory, so the files will be regular files.
>

ugghhh you're right about the directories.

For values within /etc/passwd though, is there a reason to not accept 
both /sbin/nologin and /usr/sbin/nologin?

redhatrises:
> On 10/23/18 5:30 PM, redhatrises wrote: This makes sense, and bet you they are the same. ;-) The symlink is over a directory, so the files will be regular files.
> ugghhh you're right about the directories. For values within /etc/passwd though, is there a reason to not accept both /sbin/nologin and /usr/sbin/nologin?

Good question. Generally `/sbin/nologin` is the expected and accepted convention across multiple derivatives. If you update this check to reference `/usr/sbin`, you should make sure to do the same for all.

linux_os/guide/system/accounts/accounts-restrictions/root_logins/no_shelllogin_for_systemaccounts/oval/shared.xml
```diff
@@ -77,7 +77,7 @@
   <!-- Get all /etc/passwd entries having shell defined as OVAL object -->
   <ind:textfilecontent54_object id="object_etc_passwd_entries" version="1">
     <ind:filepath>/etc/passwd</ind:filepath>
-    <ind:pattern operation="pattern match">^(?!root).*:x:([\d]+):[\d]+:[^:]*:[^:]*:(?!\/sbin\/nologin|\/bin\/sync|\/sbin\/shutdown|\/sbin\/halt).*$</ind:pattern>
+    <ind:pattern operation="pattern match">^(?!root).*:x:([\d]+):[\d]+:[^:]*:[^:]*:(?!\/sbin\/nologin|\/usr\/sbin\/nologin|\/bin\/sync|\/sbin\/shutdown|\/sbin\/halt).*$</ind:pattern>
```
redhatrises: This regex should be better updated to handle different paths (`/usr/sbin` and `/sbin`). Without checking if this works or not, but maybe something like:
(\/usr\/sbin\/|\/sbin\/)(nologin|sync|shutdown|halt)

shawndwells:
i don't care enough about this.

Issue #3454, opened 2018-09-30
RHEL 7 STIG Update - RHEL-07-030700 - Rule Update


mrabe142:
As part of the latest RHEL 7 STIG update referenced in https://github.com/ComplianceAsCode/content/issues/3370, a rule has been updated in the current version.  Information about the rule change is outlined below:

**Rule Metadata**
> Red Hat Enterprise Linux 7 Security Technical Implementation Guide :: Release: 1 Benchmark Date: 27 Jul 2018
> Vuln ID: V-72163	   	Rule ID: SV-86787r5_rule	   	STIG ID: RHEL-07-030700	   
> Severity: CAT II	   	Check Reference: M	   	Classification: Unclass

**New Rule Title**
> **Rule Title:** The Red Hat Enterprise Linux operating system must audit all uses of the sudoers file and all files in the /etc/sudoers.d/ directory.

**Previous Check + Fix Content**
> **Check Text:** Verify the operating system generates audit records when successful/unsuccessful attempts to use the "sudoer" command occur.
> 
> Check for modification of the following files being audited by performing the following commands to check the file system rules in "/etc/audit/audit.rules": 
> 
> <span>#</span> grep -i "/etc/sudoers" /etc/audit/audit.rules
> 
> -w /etc/sudoers -p wa -k privileged-actions
> 
> <span>#</span> grep -i "/etc/sudoers.d/" /etc/audit/audit.rules
> 
> -w /etc/sudoers.d/ -p wa -k privileged-actions
> 
> If the commands do not return output that does not match the examples, this is a finding.
> 
> **Fix Text:** Configure the operating system to generate audit records when successful/unsuccessful attempts to use the "sudoer" command occur.
> 
> Add or update the following rule in "/etc/audit/rules.d/audit.rules":
> 
> -w /etc/sudoers -p wa -k privileged-actions
> 
> -w /etc/sudoers.d/ -p wa -k privileged-actions
> 
> The audit daemon must be restarted for the changes to take effect.

**New Check + Fix Content**
> **Check Text:** Verify the operating system generates audit records when successful/unsuccessful attempts to access the "/etc/sudoers" file and files in the "/etc/sudoers.d/" directory. 
> 
> Check for modification of the following files being audited by performing the following commands to check the file system rules in "/etc/audit/audit.rules": 
> 
> <span>#</span> grep -i "/etc/sudoers" /etc/audit/audit.rules
> 
> -w /etc/sudoers -p wa -k privileged-actions
[...]

PR #3550, opened 2018-10-24
OL7 audit media_export, file_deletion_events and sysadmin_actions rules

iokomin:
#### Description:

Added OL7 support for audit media_export, file_deletion_events and sysadmin_actions rules

- Added ol7 platform in mentioned rules and corresponding content
- Updated content to use auid macro
- Merged bash fix to shared for audit_rules_file_deletion_events rule
- Updated audit_rules_sysadmin_actions: 
  - ocil to match auditctl output
  - bash fix to handle sudoers.d (I didn't touch OVAL as it is also applicable to rhel6 where AFAIK sudoers.d is not a requirement, expect it will be properly addressed as part of #3454 fix)
- Enabled rules:
  - audit_rules_media_export
  - audit_rules_file_deletion_events

Testing successful on Oracle Linux 7
Verified auid macro produces expected values for rhel6 and rhel7 builds.


shawndwells:
Thanks for the rule updates! Looking forward to seeing the OL7 support extended.

iokomin:
@shawndwells thanks for reviewing my PRs so fast! Will do my best)

PR #3548, opened 2018-10-23
Finish ocp etcd content

redhatrises:
- Remove variables for certificate configuration as the variable settings can encourage bad security practices
- Add remaining etcd OVAL checks

PR #3552, opened 2018-10-25
Add etcd content to profile

redhatrises:
#### Description:

- Missed adding these checks in #3548


jan-cerny:
I have checked that the rules get included into the profile in output DS. Thanks.

PR #3542, opened 2018-10-22
Add kubelet OVAL checks

redhatrises:

PR #3554, opened 2018-10-25
Add OSP Scheduler OVAL checks

redhatrises:


applications/openshift/scheduler/scheduler_address_argument/oval/shared.xml
```diff
@@ -0,0 +1,30 @@
...
+  </definition>
+
+  <ind:textfilecontent54_test check="all" check_existence="all_exist" comment="scheduler network address is configured" id="test_scheduler_address_argument" version="1">
+    <ind:object object_ref="object_scheduler_address_argument" />
+    <ind:state state_ref="state_scheduler_address_argument" />
+  </ind:textfilecontent54_test>
+
+  <ind:textfilecontent54_object id="object_scheduler_address_argument" version="1">
+    <ind:filepath>/etc/origin/master/master-config.yaml</ind:filepath>
+    <ind:pattern operation="pattern match">^[\s]*kubernetesMasterConfig\:(?:[^\n]*\n+)+?[\s]*schedulerArguments\:(?:[^\n]*\n+)+?[\s]*address\:[\s]*[\n]+[\s]*-[\s]+(\S+)[\s]*$</ind:pattern>
```
shawndwells: Have seen the configs expressed multiple ways:

`````
kubernetesMasterConfig:
  schedulerArguments:
    address: 
      - 'value here'
`````

and
`````
kubernetesMasterConfig:
  schedulerArguments:
    address: 
      - noticks
`````

and
`````
kubernetesMasterConfig:
  schedulerArguments:
    address: theValue
`````

and
`````
kubernetesMasterConfig:
  schedulerArguments:
    address: 'the value'
`````

The regex doesn't seem to account for all of them. For example when using 
`````
kubernetesMasterConfig:
  schedulerArguments:
    address: 
      - 'value'
`````

the capture group ``(\S+)`` will include the ticks ``'``

File: applications/openshift/scheduler/scheduler_address_argument/rule.yml [hunk 2]
```diff
@@ -18,13 +20,12 @@ rationale: |-
...
-ocil_clause: '<tt>--address</tt> argument is not set to the IP address of the loopback device'
+ocil_clause: 'scheduler network address is not set to the IP address of the loopback device'
 
 ocil: |-
     Run the following command on the master node(s):
-     <pre>$ ps -ef | grep kube-apiserver</pre>
-
-    Verify that the <tt>--address</tt> argument is set to <tt>127.0.0.1</tt>.
+    <pre>$ sudo grep -A2 address /etc/origin/master/master-config.yaml</pre>
+    The output should return <tt>false</tt>.
```
shawndwells: The default is loopback/127.0.0.1, but some admins may manually specify it anyway. Don't want that to be a false finding.

Could change the OCIL clause to "If there is output, the IP address of the loopback device should be returned. No output indicates OpenShift is configured to the default setting of 127.0.0.1, which satisfies this requirement"... probably can word better than I.

redhatrises:
@shawndwells removed the scheduler address option as it is now deprecated in OpenShift. Also, the ticks do matter on certain checks and are not expected on other checks. The atomic-openshift-node service will fail on restart if the ticks are used incorrectly hence the checks only allowing one method.

jan-cerny:
In `make` output there is this:
```
WARNING: OVAL definition 'scheduler_profiling_argument' extends non-existing 
'ocp_service_runtime_config_scheduler_profiling_argument', removing it from OVAL definitions.```

jan-cerny:
Is this blocked by #3542 ?

redhatrises:
> Is this blocked by #3542 ?

@jan-cerny yeah... kinda forgot to mention that.

shawndwells:
> In `make` output there is this:
> 
> ```
> WARNING: OVAL definition 'scheduler_profiling_argument' extends non-existing 
> 'ocp_service_runtime_config_scheduler_profiling_argument', removing it from OVAL definitions.
> ```
> Most likely you'll need to add something to the CSV for this template, see `ocp3/templates/csv/ocp_service_runtime_config.csv`

Realized I was to quick on the merge. This wasn't done.

@redhatrises could you create a PR to add this?

redhatrises:
@shawndwells removed the service runtime check since a `ps -ef` doesn't show anything.

PR #3559, opened 2018-10-26
Add OL7 account rules

iokomin:
#### Description:

Added ol7 support to account rules including set of generated from template.

Despite proposed a lot of changes, in general it is mostly text formatting and adding ol7 to the list of supported platforms.

Checked rules content, verified evaluation and fixes to work on Oracle Linux 7.


jan-cerny:
Thank you very much for your contribution.

iokomin:
@jan-cerny thanks a lot for your review!

PR #3561, opened 2018-10-29
Ol7 dconf gnome screensaver

iokomin:
#### Description:

Added ol7 support to dconf_gnome_screensaver rules including dependencies.
In addition:
- Updated text formatting to use multi line
- Updated 'idle-delay' value in dconf_gnome_screensaver_idle_delay description
- Updated 'idle-activation-enabled' key in dconf_gnome_screensaver_idle_activation_enabled description
- To be in sync with the dconf_gnome_screensaver_mode_blank bash fix updated pattern in OVAL to match values using type.

Checked gnome options to be applicable on OL7
Verified evaluation and fixes to work on OL7.

### DIFF
--- a/linux_os/guide/system/software/gnome/gnome_screen_locking/dconf_gnome_screensaver_idle_delay/rule.yml
+++ b/linux_os/guide/system/software/gnome/gnome_screen_locking/dconf_gnome_screensaver_idle_delay/rule.yml
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: rhel7,fedora
+prodtype: rhel7,fedora,ol7
 
 title: 'Set GNOME3 Screensaver Inactivity Timeout'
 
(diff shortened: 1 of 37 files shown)


linux_os/guide/system/software/gnome/gnome_screen_locking/dconf_gnome_screensaver_idle_delay/rule.yml
```diff
@@ -12,7 +12,7 @@ description: |-
     For example, to configure the system for a 15 minute delay, add the following to
     <tt>/etc/dconf/db/local.d/00-security-settings</tt>:
     <pre>[org/gnome/desktop/session]
-    idle-delay='uint32 900'</pre>
+    idle-delay=uint32 900</pre>
```
matejak: Good catch!

iokomin:
@matejak, thanks for taking your time on review!

PR #3564, opened 2018-10-30
OL7 more pci dss rules

iokomin:
#### Description:

Enabled more ol7 pci-dss profile rules

- Enabled rules
    - ensure_logrotate_activated
    - sshd_set_idle_timeout
    - disable_prelink
    - display_login_attempts
    - gid_passwd_group_same
    - grub2_audit_argument
    - smartcard_auth
    - chronyd_or_ntpd_specify_multiple_servers
    - chronyd_or_ntpd_specify_remote_server
- renamed rhel7_ensure_there_are_servers_in_ntp_compatible_config_file to something more generic
- var_multiple_time_servers.var extended with ol7 specific servers
- smartcard_auth rhel7 bash fix merged to shared, fixed SMARTCARD_AUTH_SECTION to match OVAL expectation (hope last one is correct)
- in smartcard_auth OVAL fixed regex issue introduced in  1687648 commit

Verified rules and remediation to work on Oracle Linux 7
Checked merged content to present in rhel7 xccdf.

### DIFF
--- a/linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/smartcard_auth/bash/shared.sh
+++ b/linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/smartcard_auth/bash/shared.sh
@@ -1,4 +1,4 @@
-# platform = Red Hat Enterprise Linux 7
+# platform = Red Hat Enterprise Linux 7, multi_platform_ol
 . /usr/share/scap-security-guide/remediation_functions
 
 # Install required packages
@@ -39,7 +39,7 @@ pam_pkcs11.so nodebug"
 SMARTCARD_AUTH_CONF="/etc/pam.d/smartcard-auth"
 # Define 'pam_pkcs11.so' auth section to be appended past $PAM_ENV_SO into $SMARTCARD_AUTH_CONF
 SMARTCARD_AUTH_SECTION="\
-auth        [success=done ignore=ignore default=die] pam_pkcs11.so wait_for_card card_only"
+auth        [success=done ignore=ignore default=die] pam_pkcs11.so nodebug wait_for_card"
 # Define expected 'pam_permit.so' row in $SMARTCARD_AUTH_CONF
 PAM_PERMIT_SO="account.*required.*pam_permit.so"
 # Define 'pam_pkcs11.so' password section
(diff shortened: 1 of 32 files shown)


linux_os/guide/system/auditing/grub2_audit_argument/rule.yml
```diff
@@ -57,6 +57,8 @@ warnings:
         <li>On UEFI-based machines, issue the following command as <tt>root</tt>:
 {{% if product == "rhel7" %}}
         <pre>~]# grub2-mkconfig -o /boot/efi/EFI/redhat/grub.cfg</pre></li>
+{{% elif product == "ol7" %}}
```
redhatrises: since `ol7` and `rhel7` are the same configuration, it would make sense to use an `or` instead of an `elif`

iokomin: @redhatrises you are right. I've decided to make it under separate condition as cfg path could be changed to ol specific. However while they are same it makes sense to keep just one instance, I'll update it.

linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/smartcard_auth/bash/shared.sh
```diff
@@ -39,7 +39,7 @@ pam_pkcs11.so nodebug"
 SMARTCARD_AUTH_CONF="/etc/pam.d/smartcard-auth"
 # Define 'pam_pkcs11.so' auth section to be appended past $PAM_ENV_SO into $SMARTCARD_AUTH_CONF
 SMARTCARD_AUTH_SECTION="\
-auth        [success=done ignore=ignore default=die] pam_pkcs11.so wait_for_card card_only"
+auth        [success=done ignore=ignore default=die] pam_pkcs11.so nodebug wait_for_card"
```
shawndwells: Why add ``nodebug`` here?

shawndwells: Scratch that. Just looked at the OVAL.

iokomin:
Guys, thanks for your review!

PR #3566, opened 2018-10-30
Update api-server rules

redhatrises:


applications/openshift/api-server/api_server_admission_control_plugin_AlwaysAdmit/rule.yml
```diff
@@ -1,28 +1,25 @@
 documentation_complete: true
 
-title: 'Ensure that the admission control plugin AlwaysAdmit is not set'
+title: 'Ensure that the admission control plugin AlwaysAdmit is not Configured'
```
shawndwells:
```suggestion
title: 'Ensure that the admission control plugin AlwaysAdmit is not configured'
```

redhatrises: This PR is a WIP, but I am actually thinking that some of the titling will change to match what we do with the rest of the content.

redhatrises:
@openscap-ci test this please

applications/openshift/api-server/api_server_admission_control_plugin_AlwaysPullImages/oval/shared.xml
```diff
@@ -0,0 +1,29 @@
...
+  </definition>
+
+  <ind:textfilecontent54_test check="all" check_existence="all_exist" comment="AlwaysPullImages is configured" id="test_api_server_admission_control_plugin_AlwaysPullImages" version="1">
+    <ind:object object_ref="object_api_server_admission_control_plugin_AlwaysPullImages" />
+    <ind:state state_ref="state_api_server_admission_control_plugin_AlwaysPullImages" />
+  </ind:textfilecontent54_test>
+
+  <ind:textfilecontent54_object id="object_api_server_admission_control_plugin_AlwaysPullImages" version="1">
+    <ind:filepath>/etc/origin/master/master-config.yaml</ind:filepath>
+    <ind:pattern operation="pattern match">^[\s]*admissionConfig\:(?:[^\n]*\n+)+?[\s]*pluginConfig\:(?:[^\n]*\n+)+?[\s]*AlwaysPullImages\:[\s]*[\n]+[\s]*configuration\:[\s]*[\n]+[\s]*kind\:[\s]+DefaultAdmissionConfig[\s]*[\n]+[\s]*apiVersion\:[\s]+v1[\s]*[\n]+[\s]*disable\:[\s]+(\S+)[\s]*$</ind:pattern>
```
shawndwells: Current regex evaluates more than ``disable: false``, matching on things like ``apiVersion`` and ``DefaultAdmissionConfig``. Are those relevant to the configuration check?

applications/openshift/api-server/api_server_token_auth/oval/shared.xml
```diff
@@ -0,0 +1,25 @@
...
+    </criteria>
+  </definition>
+
+  <ind:textfilecontent54_test check="all" check_existence="none_exist" comment="token-auth-file does not exist" id="test_api_server_token_auth" version="1">
+    <ind:object object_ref="object_api_server_token_auth" />
+  </ind:textfilecontent54_test>
+
+  <ind:textfilecontent54_object id="object_api_server_token_auth" version="1">
+    <ind:filepath>/etc/origin/master/master-config.yaml</ind:filepath>
+    <ind:pattern operation="pattern match">^[\s]*token-auth-file\:[\n]+[\s*]-[\s]*(\S+)[\s]*$</ind:pattern>
```
shawndwells: Should this have ``servingInfo``, as the other API Server OVALs do?

applications/openshift/api-server/api_server_secure_port/rule.yml
```diff
@@ -1,29 +1,30 @@
...
-    <pre>$ ps -ef | grep kube-apiserver</pre>
-
-    Verify that the <tt>--secure-port</tt> argument is either not set
-    (reflecting default value of <tt>6443</tt>) or is set to an integer value
-    between 1 and 65535.
-
-references:
-    cis: 1.1.7
+    <pre>$ sudo grep -A2 secure-port /etc/origin/master/master-config.yaml</pre>
+    The output should not return <pre>0</pre>.
```
shawndwells:
```suggestion
    The output should not return <tt>0</tt>.
```

applications/openshift/api-server/api_server_token_auth/oval/shared.xml
redhatrises: No because the evaluation is that `token-auth-file` does not exist at all.

applications/openshift/api-server/api_server_secure_port/rule.yml
redhatrises: Why `<tt>`? `<pre/>` is used for a lot of these.

applications/openshift/api-server/api_server_admission_control_plugin_AlwaysPullImages/oval/shared.xml
redhatrises: I'd say so for a valid configuration check.

applications/openshift/api-server/api_server_secure_port/rule.yml
shawndwells: pre for code stanzas, tt for teletype (e.g. in-line, ``like this``)

applications/openshift/api-server/api_server_token_auth/oval/shared.xml
shawndwells: fair enough :)

applications/openshift/api-server/api_server_admission_control_plugin_AlwaysPullImages/oval/shared.xml
shawndwells: should they be independent configuration checks then?

shawndwells: in the same way PAM options are broken out in RHEL

applications/openshift/api-server/api_server_secure_port/rule.yml
shawndwells: example in https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/dhcp/disabling_dhcp_client/sysconfig_networking_bootproto_ifcfg/rule.yml#L9

applications/openshift/api-server/api_server_admission_control_plugin_AlwaysPullImages/oval/shared.xml
redhatrises: @shawndwells preferably not. These checks are already complicated as is. Let's not add way more complication to it.

shawndwells: the xccdf is worded vaguely enough to scope the subconfiguration elements. Might be something to review in future editions of the OCP content.

applications/openshift/api-server/api_server_secure_port/rule.yml
redhatrises: @shawndwells ummmm... okay.... we have a lot of OCIL checks that use `pre` rather than `tt` for everything. Interestingly enough, `<tt>` and `<pre>` are not converted into XML elements in the OCIL files, so I wonder if those elements should be removed completely e.g `Is it the case that &lt;tt&gt;--insecure-port&lt;/tt&gt; not set to &lt;tt&gt;0&lt;/tt&gt;?`?

shawndwells: ``pre`` and ``tt`` should be kept. Here's how they look in the OpenSCAP reports:

image


In some places it looks like we're using ``<code>`` and ``<tt>`` interchangeably, such as the jinja macros.... might need to address that.

https://github.com/ComplianceAsCode/content/blob/master/shared/macros.jinja#L450#L453
`````
{{%- macro describe_file_group_owner(file, group) %}}
    To properly set the group owner of <code>{{{ file }}}</code>, run the command:
    <pre>$ sudo chgrp {{{ group }}} {{{ file }}}</pre>
{{%- endmacro %}}

`````

shawndwells: Example of both used properly in an OCIL:
https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/logging/log_rotation/ensure_logrotate_activated/rule.yml#L33#L37

shawndwells: wrt to not rendering in the OCIL files.... leaning towards that being a bug

redhatrises: @shawndwells probably should be fixed at some point. Talking about removing `<tt>` from OCIL only since it's not generated correctly. `<pre>` does not even exist in generated OCIL either. 

OpenSCAP reports aren't showing OCIL, so `<pre>` and `<tt>` would stay in the other elements that convert it correctly.

shawndwells: Ah, believe I understand now. The ``question_text`` element of OCIL only takes plaintext, so our enriched tags are being converted so something like this:

`````
Is it the case that &lt;tt&gt;--feature-gates&lt;/tt&gt; argument is set to a value that includes &lt;tt&gt;AdvancedAuditing=false&lt;/tt&gt;?
`````

shawndwells: in which case yes, html tags should likely be stripped for the ssg-product-ocil.xml files

but until then... pre and tt should be used.

also opened https://github.com/ComplianceAsCode/content/issues/3583 to track tt vs code

redhatrises: > Ah, believe I understand now. The `question_text` element of OCIL only takes plaintext, so our enriched tags are being converted so something like this:
> 
> ```
> Is it the case that &lt;tt&gt;--feature-gates&lt;/tt&gt; argument is set to a value that includes &lt;tt&gt;AdvancedAuditing=false&lt;/tt&gt;?
> ```

@shawndwells correct. However, pre is generated corrected to text, so let's not use `tt` for OCIL.

shawndwells: @redhatrises kindly pointed out that OCIL isn't rendered into the HTML guides anyway. We can brush this under the rug until something more permanent is thought of.

PR #3567, opened 2018-10-31
Updated BUILD.md to make clear that CTest does not run SSG Test Suite

matusmarhefka:


jan-cerny:
@openscap-ci add to whitelist

PR #3572, opened 2018-10-31
controller_terminated_pod_gc_threshhold

shawndwells:
The OpenShift Container Platform node performs two types of garbage collection: container garbage collection and image garbage collection. Container garbage collection is enabled by default and happens automatically in response to eviction thresholds being reached  However, the preferred approach of using a controller (e.g. DeploymentConfig, ReplicationSet) to manage Pods will not leave Pods in terminated states, which are created when using Jobs or standalone Pods.
OpenShift does not set the terminated-pod-gc-threshold value differently from the Kubernetes default of 12500.  Setting this value explicitly to can be achieved via kubernetesMasterConfig.controllerArguments in the master-config.yaml.
https://docs.openshift.com/container-platform/3.10/admin_guide/garbage_collection.html)

Given mitigations of this is being dropped.

### DIFF
--- a/applications/openshift/controller/controller_terminated_pod_gc_threshhold/oval/shared.xml
+++ b/applications/openshift/controller/controller_terminated_pod_gc_threshhold/oval/shared.xml
@@ -1,0 +1,29 @@
+<def-group>
+  <definition class="compliance" id="controller_terminated_pod_gc_threshhold" version="1">
+    <metadata>
+      <title>Ensure that terminated-pod-gc-threshold is enabled.</title>
+      <affected family="unix">
+        <platform>multi_platform_ocp</platform>
+      </affected>
+      <description>Ensure terminated-pod-gc-threshold is enabled.</description>
+    </metadata>
+    <criteria operator="AND">
+      <criterion comment="terminated-pod-gc-threshold is configured" test_ref="test_controller_terminated_pod_gc_threshhold" negate="true" />
+    </criteria>
+  </definition>
+
+  <ind:textfilecontent54_test check="all" check_existence="all_exist" comment="terminated-pod-gc-threshold is configured" id="test_controller_terminated_pod_gc_threshhold" version="1">
+    <ind:object object_ref="object_controller_terminated_pod_gc_threshhold" />
+    <ind:state state_ref="state_controller_terminated_pod_gc_threshhold" />
+  </ind:textfilecontent54_test>
+
+  <ind:textfilecontent54_object id="object_controller_terminated_pod_gc_threshhold" version="1">
+    <ind:filepath>/etc/origin/master/master-config.yaml</ind:filepath>
+    <ind:pattern operation="pattern match">^[\s]*kubernetesMasterConfig\:(?:[^\n]*\n+)+?[\s]*controllerArguments\:[\s]*[\n]+[\s]*terminated-pod-gc-threshold\:[\s]*[\n]+[\s]*-[\s]+(\S+).*$</ind:pattern>
+    <ind:instance datatype="int">1</ind:instance>
+  </ind:textfilecontent54_object>
+
+  <ind:textfilecontent54_state id="state_controller_terminated_pod_gc_threshhold" version="1">
+    <ind:subexpression datatype="string" operation="pattern match">^'false'$</ind:subexpression>
+  </ind:textfilecontent54_state>
+</def-group>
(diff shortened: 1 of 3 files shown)


redhatrises:
@shawndwells shouldn't we keep this rule and just check that the value is `> 0`? Seems like if someone disables it, availability will be affected.

shawndwells:
yeah, that argument is sound. marking as WIP to update.

shawndwells:
@redhatrises updated PR to ensure this is enabled

applications/openshift/controller/controller_terminated_pod_gc_threshhold/rule.yml
```diff
@@ -22,14 +24,18 @@ rationale: |-
...
+    <pre>$ sudo grep terminated-pod-gc-threshold /etc/origin/master/master-config.yaml</pre>
+
+    Verify that the <tt>terminated-pod-gc-threshold</tt> argument is not set
+    to <tt>false</tt>.
 
-    Verify that the <tt>--terminated-pod-gc-threshold</tt> argument is set
-    as appropriate.
+    If the configuration file does not explicitly set
+    <tt>terminated-pod-gc-threshold</tt>, and the grep command returns no
+    text, the default value of <tt>true</tt> is being used..
```
redhatrises:
```suggestion
    text, the default value of <tt>true</tt> is being used.
```

applications/openshift/controller/controller_terminated_pod_gc_threshhold/rule.yml
```diff
@@ -1,15 +1,17 @@
 documentation_complete: true
 
-title: 'Ensure that the --terminated-pod-gc-threshold argument is set'
+title: 'Ensure terminated-pod-gc-threshold argument is enabled'
```
redhatrises:
```suggestion
title: 'Enable terminated-pod-gc-threshold for the Controller Manager'
```

applications/openshift/controller/controller_terminated_pod_gc_threshhold/oval/shared.xml
```diff
@@ -0,0 +1,29 @@
+<def-group>
+  <definition class="compliance" id="controller_terminated_pod_gc_threshhold" version="1">
+    <metadata>
+      <title>Ensure that terminated-pod-gc-threshold is enabled.</title>
```
redhatrises:
```suggestion
      <title>Enable terminated-pod-gc-threshold for the Controller Manager</title>
```

applications/openshift/controller/controller_terminated_pod_gc_threshhold/oval/shared.xml
```diff
@@ -0,0 +1,29 @@
...
+      <affected family="unix">
+        <platform>multi_platform_ocp</platform>
+      </affected>
+      <description>Ensure terminated-pod-gc-threshold is enabled.</description>
+    </metadata>
+    <criteria operator="AND">
+      <criterion comment="terminated-pod-gc-threshold is configured" test_ref="test_controller_terminated_pod_gc_threshhold" negate="true" />
```
redhatrises: Rather than `negate="true"`, I wonder if it makes more sense when reading this OVAL to just check if it is set to true?

shawndwells: Not sure. Initial reasoning is that it's true by default, which means no special configuration required. To ensure operators don't have to go around manually specifying default settings, swapped to using negate and failing if ``false`` is present.

redhatrises: It would be the same result anyway.

redhatrises:
@openscap-ci test this please

applications/openshift/controller/controller_terminated_pod_gc_threshhold/oval/shared.xml
shawndwells: Not following. If the OVAL is set to check for ``true``, operators would be forced to go specify that configuration manually (even though it's the default). 

Current OVAL will fail if false is present, vs forcing operators to manually configure default settings in their conf files

redhatrises: good point

redhatrises:
ACK

Issue #3569, opened 2018-10-31
Regression: Rules that aren't applicable to containers aren't marked as "machine only" anymore


jan-cerny:
#### Description of problem:
Certain rules doesn't make sense in containers, they shouldn't be evaluated if the scanned target is a container, they should be evaluated only if scan target is bare-metal or VM.

This problem is supposed to be solved by marking those rules by special CPEs, eg. `cpe:/a:machine`
or `cpe:/a:container`.  See eg. https://github.com/ComplianceAsCode/content/pull/2337 where we added `<platform idref="cpe:/a:machine" />` to multiple rules or groups.

However, then we converted the rules and groups to new YAML format. Unfortunately, the CPEs were not converted. That means all the rules are now applicable to containers.

#### SCAP Security Guide Version:
0.1.41

#### Operating System Version:
all

#### Steps to Reproduce:

1.
2.
3.
4.

#### Actual Results:
services, mcafee, etc. are evaluated when scanning a container

#### Expected Results:
services, mcafee, etc. are not evaluated when scanning a container

#### Addition Information/Debugging Steps:
Check fcfb389319c578d91d3147f368bf53161f032108 to see which rules used to have the machine or container CPEs.

It would be useful to have a test for this to prevent repeating this regression in future.

jan-cerny:
CCDF Groups and Rules affected by fcfb389:
Groups: https://paste.fedoraproject.org/paste/BOef4M9XqsK8Iz~JjNmgoQ
Rules: https://paste.fedoraproject.org/paste/d5g7Y5ARXnrv-TU-zLCKwg

jan-cerny:
Closed in #3576

PR #2337, opened 2017-09-19
Services are machine only

yuumasato:
Let's not apply Rules related to enabling or configuring services on containers.

Checking these Rules will result in false positives most of the time.
A services configured in a container will not be started unless
an init system is the entry point and starts them or they
are the entry points them selves.

PR #3576, opened 2018-11-01
Container-only rules

jan-cerny:
#### Description:
Fixes  regression #3569 .

Adds a new field into rule and group YAML and a support for this new field into our scripts. It allows us to mark rules as bare-metal machine only or container only by adding "platform: machine" or "platform: container" to rule.yml or group.yml. This will be populated into XCCDF as platform element. The platform element is a child element of Rule or Group element. For convenience the CPE in idref attribute value is defined as a constant (under ssg/constants.py).

Then, we mark rules as machine only - we mark those rules that were marked as machine-only before we converted to the new YAML format.

#### Rationale
Certain rules doesn't make sense in containers, they shouldn't be evaluated if the scanned target is a container, they should be evaluated only if scan target is bare-metal or VM.

This problem is supposed to be solved by marking those rules by special CPEs, eg. cpe:/a:machine
or cpe:/a:container. See eg. #2337 where we added <platform idref="cpe:/a:machine" /> to multiple rules or groups.

However, -then- after we added the CPEs we converted the rules and groups to new YAML format. Unfortunately, the CPEs were not converted. That means all the rules are now applicable to containers.

### DIFF
--- a/ssg/build_yaml.py
+++ b/ssg/build_yaml.py
@@ -436,7 +436,10 @@ class Group(object):
 
         if self.platform:
             platform_el = ET.SubElement(group, "platform")
-            platform_cpe = XCCDF_PLATFORM_TO_CPE[self.platform]
+            try:
+                platform_cpe = XCCDF_PLATFORM_TO_CPE[self.platform]
+            except KeyError:
+                raise ValueError("Unsupported platform '%s' in rule '%s'." % (self.platform, self.id_))
             platform_el.set("idref", platform_cpe)
 
         for _value in self.values.values():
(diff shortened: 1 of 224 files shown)


jan-cerny:
Automatic strip that produced this: https://paste.fedoraproject.org/paste/dkkFzpSzZPoYqdrDTdHOJw

ssg/build_yaml.py
```diff
@@ -610,6 +620,11 @@ def to_xml_element(self):
 
         add_warning_elements(rule, self.warnings)
 
+        if self.platform:
+            platform_el = ET.SubElement(rule, "platform")
+            platform_cpe = XCCDF_PLATFORM_TO_CPE[self.platform]
```
matejak: It would be nice to have typos s.a. `platform: mchine` handled gracefully.

matejak:
Great job, we have code that works, and a test. The regression has been fix in the content as well. Thank you very much for your pull request!

jan-cerny:
@matejak Are there more changes needed? Do you need changes to support #3578 ? I think it could be merged.

matejak:
Sorry, I have thought that I have merged it already.

PR #3577, opened 2018-11-01
Added ansible-lint and yamllint checks of ansible playbooks into CTest

matusmarhefka:
#### Description:

CMake is extended with option ANSIBLE_CHECKS to also enable `ansible-lint`
and `yamllint` checks of generated playbooks (thse checks are not enabled
by default in CTest). To perform these checks:
```
$ cd build/
$ cmake -DANSIBLE_CHECKS=ON ../ && make
$ ctest -R ansible
```

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -33,10 +33,17 @@ source code:
 yum install git
 ```
 
-(optional) Install the ShellCheck package to perform fix script static analysis:
+(optional) Install the `ShellCheck` package to perform fix script static analysis:
 
 ```bash
 yum install ShellCheck
+```
+
+(optional) Install `yamllint` and `ansible-lint` packages to perform Ansible
+playbooks checks. These checks are not enabled by default in CTest, to enable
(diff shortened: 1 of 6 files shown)


BUILD.md
```diff
@@ -171,7 +171,8 @@ tables/table-rhel7-stig.html
 ### Testing
 
 To ensure validity of built artifacts prior to installation, we recommend
-running our test suite against the build output. This is done with CTest:
+running our test suite against the build output. This is done with CTest
+and you need to also install `yamllint` and `ansible-lint` packages:
```
jan-cerny: I have noticed that there is an optional test dependency ShellChekc. Is it possible to describe installing both packages at the same place in the docs?

matusmarhefka: Done. I moved the part about packages to the proper place.

jan-cerny:
I'm able to get a huge number of errors. That's great!

PR #3578, opened 2018-11-01
Make our Ansible content to skip tasks on platforms that are not applicable.

matejak:
Some rules can define platforms that they are applicable for using the `<platform idref="cpe:/a:platform-id"/>` tag - see #3576.
This PR adds support for `@ANSIBLE_ENSURE_PLATFORM@` macro that expands to a `when: ...` Ansible clause, if the respective rule is bound to a platform.

- If a known platform is encountered, respective `when` entry is created.
- If an unknown platform is encountered, a descriptive error comment is created.
- If no platform is specified, an informative comment is created.

Like this, there won't be any annoying failures when remediating a container with rules that are machine-only.

How to test this:
#. Build SSG.
#. Choose a product and open it's `shorthand.xml`
#. Find a rule that has an Ansible remediation and add e.g. `<platform idref="cpe:/a:machine"/>` as a sub-element of the respective `Rule` element.
#. Re-run `make` - the shorthand won't be regenerated, but changes will be picked up.
#. Check out the HTML guide and expand the Ansible remediation of the concerned rule.

Remarks:
- Only one platform per rule is supported at the moment.
- The 'machine' platform is supported.
- The Ansible detection of whether or not are we running on a container is based on running `ansible -m setup localhost` in a Docker container.


jan-cerny:
@matejak 

1. Would it be possible that the `@ANSIBLE_ENSURE_PLATFORM@` (or its contents) will be added automagically to every Ansible remediation so that the people will not have to add it manually when they add a new Ansible remediation?

2. In #3576 I allowed whole groups to have `<platform idref="cpe:/a:machine">` . I'm afraid that it will not work with your code and would be more difficult to handle. Would it be better to change #3576 to abandon the option to have `platform` element in `Group` elements and have this element ONLY in `Rule` elements?

matejak:
@jan-cerny 
1. We already have the `tags: @ANSIBLE_TAGS@` in every remediation, so I also thought of a boilerplate autogeneration. So I think that it is a good idea, but let's do it in a separate pull request.
2. I don't know - I am slightly in favor of having everything in the rule - the XSLT is annoying, and propagating group properties to rules is much easier to do in the Python code, am I right?

jan-cerny:
@matejak What do you mean by "propagating group properties to rules" in Python code?

redhatrises:
I am completely in favor of having the XSLT replaced by python code..... just in case someone wants to do that.

shared/transforms/xccdf-addremediations.xslt
```diff
@@ -109,7 +116,35 @@
...
+  <xsl:variable name='newline'><xsl:text>
+</xsl:text></xsl:variable>
+
+  <xsl:variable name="rep5">
+    <xsl:variable name="platform_condition">
+      <xsl:choose>
+        <xsl:when test="$rule/xccdf:platform">
+          <xsl:choose>
+            <xsl:when test="$platform_cpe = 'machine'">
+              <xsl:value-of select="concat('when:  # Bare-metal-only rule, not applicable for containers', $newline, '    - ansible_virtualization_role == &quot;guest&quot;', $newline, '    - ansible_virtualization_type == &quot;docker&quot;')"/>
```
jan-cerny: I would improve this, because it is not meant to be "bare-metal only", it should be applicable in virtual machines as well.

shared/transforms/xccdf-addremediations.xslt
```diff
@@ -109,7 +116,35 @@
...
+            </xsl:when>
+            <xsl:otherwise>
+              <xsl:value-of select="concat('# Error ensuring that the platform is applicable - unknown platform CPE spec encountered: &quot;', $platform_cpe, '&quot;')"/>
+            </xsl:otherwise>
+	  </xsl:choose>
+        </xsl:when>
+        <xsl:otherwise>
+          <xsl:value-of select="'# This rule is applicable for all platforms'"/>
```
jan-cerny: I think this is not needed to be emphasized, it is kind of implicit.

matejak: OK

shared/transforms/xccdf-addremediations.xslt
matejak: I have expanded the description.

shared/transforms/xccdf-addremediations.xslt
```diff
@@ -109,7 +116,57 @@
...
+  <xsl:variable name='newline'><xsl:text>
+</xsl:text></xsl:variable>
+
+  <xsl:variable name="rep5">
+    <xsl:variable name="platform_condition">
+      <xsl:choose>
+        <xsl:when test="$rule/xccdf:platform">
+          <xsl:choose>
+            <xsl:when test="$platform_cpe = 'machine'">
+              <xsl:value-of select="concat('when:  # Bare-metal-only rule, not applicable for containers', $newline, '    - ansible_virtualization_role != &quot;guest&quot;', $newline, '    - ansible_virtualization_type != &quot;docker&quot;')"/>
```
jan-cerny: I think that the rules are not bare-metal only. They're also applicable to virtual machines.

shared/transforms/xccdf-addremediations.xslt
```diff
@@ -109,7 +116,57 @@
...
+            </xsl:when>
+            <xsl:otherwise>
+              <xsl:value-of select="concat('# Error ensuring that the platform is applicable - unknown platform CPE spec encountered: &quot;', $platform_cpe, '&quot;')"/>
+            </xsl:otherwise>
+	  </xsl:choose>
+        </xsl:when>
+        <xsl:otherwise>
+          <xsl:value-of select="'# This rule is applicable for all platforms'"/>
```
jan-cerny: I think the fact the task is applicable to all platforms is kind of implicit so it doesn't needed to be emphasized. Moreover, it is not a rule, it is an Ansible task.

jan-cerny:
When I generate a playbook using me
```
oscap xccdf generate fix --fix-type ansible --profile ospp ssg-fedora-ds.xml > playbook.yml
```
and I run the syntax check
```
ansible-playbook --syntax-check playbook.yml
```
I get plenty of warnings like this.

```
[WARNING]: While constructing a mapping from /home/jcerny/scap-security-guide/build/playbook.yml, line
8457, column 7, found a duplicate dict key (when). Using last defined value only.
```
Could you please take a look at this problem?

Morevoer, I have the following problem when I run 
```
ansible-playbook -e ansible_python_interpreter=/usr/bin/python3 -i "2e34ac53a67a354f8acc07649fed273c0a6da1159922647c009f2bb1f9059c2e," -c docker ~/scap-security-guide/build/playbook.ym
```
The aforementioned container  is created by `docker run -d fedora:29 sleep 600`

```
< TASK [Ensure GPG check Enabled for Local Packages (Yum)] >
 ----------------------------------------------------------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||

fatal: [2e34ac53a67a354f8acc07649fed273c0a6da1159922647c009f2bb1f9059c2e]: FAILED! => {"msg": "The conditional check '(ansible_distribution == \"RedHat\" or ansible_distribution == \"CentOS\" or yum_config_file.
stat.exists) and' failed. The error was: template error while templating string: unexpected 'end of statement block'. String: {% if (ansible_distribution == \"RedHat\" or ansible_distribution == \"CentOS\" or yu
m_config_file.stat.exists) and %} True {% else %} False {% endif %}\n\nThe error appears to have been in '/home/jcerny/scap-security-guide/build/playbook.yml': line 1033, column 7, but may\nbe elsewhere in the f
ile depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n    - name: Ensure GPG check Enabled for Local Packages (Yum)\n      ^ here\n"}                                              
        to retry, use: --limit @/home/jcerny/scap-security-guide/build/playbook.retry
```

That's because there are some trailing `and`s in the playbook, eg.
```
    when: ansible_distribution == "Fedora" and 
```

Can you take a look at this problem as well?

jan-cerny:
Better use this as inventory:
```
test_container ansible_connection=docker ansible_python_interpreter=/usr/bin/python3
```
and run your container:
```
docker run --name test_container -d fedora:29 sleep 600
```
and then run playbook with that inventory:
```
ansible-playbook -i inventory ~/scap-security-guide/build/playbook.yml
```

matejak:
You should also be able to run
```
docker run -ia --rm --name test_container -d fedora:29 /bin/bash
```
in a terminal, perform the Ansible run in another terminal, and just exit the shell in the first terminal after Ansible finishes.

jan-cerny:
@matejak I have found that the conditions are not inserted correctly which leads to errors.
For example, when I apply a Fedora OSPP playbook to a Fedora 29 container I get:
```
< TASK [get back device previous mount option] >
 ----------------------------------------------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||

fatal: [test_container]: FAILED! => {"msg": "The conditional check '(device_name.stdout | length > 0) and ansible_virtualization_role != \"guest\" and ansible_virtualization_type != \"docker\")' failed. The error was: template error while templating string: unexpected ')'. String: {% if (device_name.stdout | length > 0) and ansible_virtualization_role != \"guest\" and ansible_virtualization_type != \"docker\") %} True {% else %} False {% endif %}\n\nThe error appears to have been in '/home/jcerny/scap-security-guide/build/roles/ssg-fedora-role-ospp.yml': line 1639, column 9, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n    - block:\n      - name: get back device previous mount option\n        ^ here\n"}                                                               
        to retry, use: --limit @/home/jcerny/scap-security-guide/build/roles/ssg-fedora-role-ospp.retry
```
The problem seems to be here:
```
when: (device_name.stdout | length > 0) and ansible_virtualization_role != "guest" and ansible_virtualization_type != "docker")
```
notice the `)` at the end which has no partner.

Could you please take a look at this problem and find out if there aren't other similar occurrences?

matejak:
It is fixed, but just for reference, the `ansible-playbook --syntax-check` doesn't care about unmatched parentheses in `when:` statements.

matejak:
@jan-cerny Thank you for your thorough review! I owe you a cake on this one.

PR #3579, opened 2018-11-02
Added Crypto Policy rule for the BIND service.

matejak:
The rule ensures that the service adheres to the system-wide crypto policy.

Remarks:
* There is a new Crypto Policy group for rules like this.
* Some corner cases may be missed by the OVAL check - ensuring that the `include` statement is the last line of the `options` section and is not commented out by `/* ... */` comments is quite difficult.
* The rule is part of the Fedora `standard` profile.
* The SSG tests pass, I recommend using the Docker backend.
* Ansible check/remediation is IMO not possible - the BIND config file format is not understood by common Ansible modules.

PR #3580, opened 2018-11-05
Added Crypto Policy rule for JRE.

matejak:
The rule ensures that the service adheres to the system-wide crypto policy.

Remarks:
* The rule is part of the Fedora `standard` profile.
* The SSG tests pass, I recommend using the Docker backend.
* The Ansible fix passes the SSG Test Suite if you make sure that Python is installed in the backend.

### DIFF
--- a/jre/guide/java/java_jre_configure_crypto_policy/ansible/shared.yml
+++ b/jre/guide/java/java_jre_configure_crypto_policy/ansible/shared.yml
@@ -1,4 +1,4 @@
-# platform = multi_platform_all
+# platform = multi_platform_all,Java Runtime Environment
 # reboot = true
 # strategy = disable
 # complexity = low
--- a/jre/guide/java/java_jre_configure_crypto_policy/ansible/shared.yml
+++ b/jre/guide/java/java_jre_configure_crypto_policy/ansible/shared.yml
@@ -1,4 +1,4 @@
-# platform = multi_platform_all,Java Runtime Environment
+# platform = Java Runtime Environment
 # reboot = true
 # strategy = disable
 # complexity = low
--- a/jre/guide/java/java_jre_configure_crypto_policy/oval/shared.xml
+++ b/jre/guide/java/java_jre_configure_crypto_policy/oval/shared.xml
@@ -1,25 +1,26 @@
 <def-group>
-  <definition class="compliance" id="configure_jre_crypto_policy" version="1">
+  <definition class="compliance" id="java_jre_configure_crypto_policy" version="1">
     <metadata>
       <title>Configure JRE to use System Crypto Policy.</title>
       <affected family="unix">
         <platform>multi_platform_fedora</platform>
+        <platform>Java Runtime Environment</platform>
       </affected>
       <description>JRE should be configured to use the system-wide crypto policy setting.</description>
     </metadata>
     <criteria>
-      <criterion test_ref="test_configure_jre_crypto_policy"
+      <criterion test_ref="test_java_jre_configure_crypto_policy"
       comment="Check that the configuration mandates usage of system-wide crypto policies." />
     </criteria>
   </definition>
 
-  <ind:textfilecontent54_test id="test_configure_jre_crypto_policy"
+  <ind:textfilecontent54_test id="test_java_jre_configure_crypto_policy"
   comment="Check that the configuration mandates usage of system-wide crypto policies."
   check="all" check_existence="all_exist" version="1">
-    <ind:object object_ref="object_configure_jre_crypto_policy" />
+    <ind:object object_ref="object_java_jre_configure_crypto_policy" />
   </ind:textfilecontent54_test>
 
-  <ind:textfilecontent54_object id="object_configure_jre_crypto_policy"
+  <ind:textfilecontent54_object id="object_java_jre_configure_crypto_policy"
   version="1">
     <ind:filepath>/usr/lib/jvm/jre/lib/security/java.security</ind:filepath>
     <ind:pattern operation="pattern match">^\s*security.useSystemPropertiesFile=true\s*$</ind:pattern>
--- a/jre/guide/java/java_jre_configure_crypto_policy/oval/shared.xml
+++ b/jre/guide/java/java_jre_configure_crypto_policy/oval/shared.xml
@@ -3,7 +3,6 @@
     <metadata>
       <title>Configure JRE to use System Crypto Policy.</title>
       <affected family="unix">
-        <platform>multi_platform_fedora</platform>
         <platform>Java Runtime Environment</platform>
       </affected>
       <description>JRE should be configured to use the system-wide crypto policy setting.</description>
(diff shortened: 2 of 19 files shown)


shawndwells:
Should this be added to the JRE content, vs Fedora?

https://github.com/ComplianceAsCode/content/tree/master/jre

redhatrises:
Is this not configurable in `/etc/.java`?

matejak:
@redhatrises I have found no indication that the `java.security` file could be located elsewhere than the `lib\security` subdirectory of the respective JRE.

@shawndwells This is a tough question - there will be more rules that relate to crypto policies, so I decided to go this way in #3579. You typically aren't interested in SSH server to satisfy all requirements, ignoring other services that use encryption. This group aggregates rules that are all about encryption, and they are likely to fail for the same reason. So instead of getting five FAILs across multiple groups, you will get them in one group, which points you to the root cause.

redhatrises:
> @redhatrises I have found no indication that the `java.security` file could be located elsewhere than the `lib\security` subdirectory of the respective JRE.
> 
> @shawndwells This is a tough question - there will be more rules that relate to crypto policies, so I decided to go this way in #3579. You typically aren't interested in SSH server to satisfy all requirements, ignoring other services that use encryption. This group aggregates rules that are all about encryption, and they are likely to fail for the same reason. So instead of getting five FAILs across multiple groups, you will get them in one group, which points you to the root cause.

Agreed that this is a tough question. However after mulling it over, the rule and associated content should be in the JRE content and not the OS content. A couple of reasons for this:

1. JRE is an application that does not have to be installed on Fedora. So, you would get a fail if JRE is not installed (which there are scenarios where this is the case) which also means it should not be enabled by default in the `standard` profile.
2. If JRE is installed, setting the crypto policy to use the system crypto policy is not enough for securing Java. More has to be done which is what the JRE content is for.

matejak:
@shawndwells , @redhatrises I have moved the rule to the jre section and renamed it, so it conforms to other Java rules. The downside of this is that it can't be easily tested by the test suite, but tests are passing for the earlier commit, so it may not be an issue.

jre/guide/java/java_jre_configure_crypto_policy/ansible/shared.yml
```diff
@@ -1,4 +1,4 @@
-# platform = multi_platform_all
+# platform = multi_platform_all,Java Runtime Environment
```
yuumasato: Shouldn't the platform be only `Java Runtime Environment`? 
Do you keep it for testing reasons?
I have same comment for Bash and OVAL content.

yuumasato:
@redhatrises What is the practice in the field regarding JRE content?
When people are concerned about Java in their systems, do they scan the first the OS, for example Fedora or RHEL content, and then scan the Java configurations, using JRE content?

tests/data/group_system/group_crypto_policies/rule_configure_jre_crypto_policy/commented_out.fail.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+# profiles = xccdf_org.ssgproject.content_profile_standard
```
yuumasato: There is no `standard` profile for JRE. Should there be one?
Because the only profile for JRE is one derived from STIG.

jre/guide/java/java_jre_configure_crypto_policy/oval/shared.xml
```diff
@@ -0,0 +1,30 @@
+<def-group>
+  <definition class="compliance" id="java_jre_configure_crypto_policy" version="1">
+    <metadata>
+      <title>Configure JRE to use System Crypto Policy.</title>
+      <affected family="unix">
+        <platform>multi_platform_fedora</platform>
```
yuumasato: I think there is no need to keep platform Fedora.
Platform `Java Runtime Environment` is applicable as long as java is installed.

jre/guide/java/java_jre_configure_crypto_policy/ansible/shared.yml
matejak: As the remediation content won't get anywhere without the respective rule, I don't think that it makes sense to restrict it in that way. The Ansible playbook will do its job on Fedora, and maybe on some other Linux distribution, so let's not restrict Ansible for the Java "platform".

tests/data/group_system/group_crypto_policies/rule_configure_jre_crypto_policy/commented_out.fail.sh
yuumasato: I'm okay to leave the rule out of any profile for now. Adding a new profile is totally out of scope for this PR.

matejak:
After discussing the meaning of platform, I have removed other than Java-related platforms from the content.

tests/data/group_system/group_crypto_policies/rule_configure_jre_crypto_policy/commented_out.fail.sh
matejak: I have marked this for the `(all)` profile, that was introduced in the recent OpenSCAP.

jan-cerny:
I cannot run the tests.
```
[jcerny@thinkpad tests{crypto_jre}]$ python3 test_suite.py rule --debug --libvirt qemu:///system ssg-test-suite-fedora --datastream ../build/ssg-jre-ds.xml --xccdf-id scap_org.open-scap_cref_ssg-jre-xccdf-1.2.xml java_jre_configure_crypto_policy
INFO - Logging into /home/jcerny/scap-security-guide/tests/logs/rule-custom-2018-11-12-1410/test_suite.log                                                                                                        
ERROR - No matching rule ID found for '['java_jre_configure_crypto_policy']'
ERROR - Nothing has been tested!
```

matejak:
The problem boils down to the fact that the rule is not part of any profile. Therefore, it has to be used with the `(all)` profile, which the unpatched testsuite can't handle, and even `oscap-ssh` can't handle it. The problem is that `(all)` causes shell syntax errors.

linux_os/guide/system/crypto_policies/group.yml
```diff
@@ -0,0 +1,5 @@
+documentation_complete: true
```
redhatrises: @matejak this file needs to be removed from this PR, but after that, I think that it is good to merge.

jre/guide/java/java_jre_configure_crypto_policy/rule.yml
```diff
@@ -0,0 +1,27 @@
...
+    Overriding the system crypto policy makes the behavior of the Java runtime violates expectations,
+    and makes system configuration more fragmented.
+
+severity: unknown
+
+ocil_clause: |-
+    the <pre>java.security</pre> doesn't contain the <pre>security.useSystemPropertiesFile=true</pre> line
+
+ocil: |-
+    To verify that Java uses the system crypro policy, check out that the respective JRE config file
```
jan-cerny: @matejak crypro

jan-cerny:
@matejak please remove documentation_complete and change "crypro", and then it can be merged, it works great on my machine

matejak:
Thank you all for looking into this PR, there is no crypro and no stray group YAML any more.

redhatrises:
LGTM

PR #3581, opened 2018-11-05
Update RHOSP content

redhatrises:
- Update content to OSP13
- Move OpenStack content to applications directory
- Change name from "rhel-osp" to "rhosp" for accuracy
- Remove RHEL templates and use OSP templates


redhatrises:
Do we really need to verify if there is an OVAL check for XCCDF rules? Really XCCDF content can be created without OVAL, so not sure why we have to actually error if no OVAL corresponds to an XCCDF rule.

shawndwells:
On 11/5/18 7:59 PM, redhatrises wrote:
> Do we really need to verify if there is an OVAL check for XCCDF rules? 
> Really XCCDF content can be created without OVAL, so not sure why we 
> have to actually error if no OVAL corresponds to an XCCDF rule.

There are some cases where OVAL is specifically not included, e.g. the 
drive encryption requirements in RHEL.

Didn't know the build system was enforcing XCCDF+OVAL

tests/unit/ssg-module/test_rules.py
```diff
@@ -14,13 +14,13 @@ def test_is_applicable():
     assert ssg.rules.is_applicable('multi_platform_rhel', 'rhel7')
     assert ssg.rules.is_applicable('Red Hat Enterprise Linux 7', 'rhel7')
 
-    assert ssg.rules.is_applicable('all', 'rhel-osp7')
-    assert ssg.rules.is_applicable('multi_platform_rhel-osp', 'rhel-osp7')
-    assert ssg.rules.is_applicable('rhel-osp7', 'rhel-osp7')
-    assert ssg.rules.is_applicable('Red Hat OpenStack Platform 7', 'rhel-osp7')
-    assert not ssg.rules.is_applicable('rhel7', 'rhel-osp7')
+    assert ssg.rules.is_applicable('all', 'rhosp13')
+    assert ssg.rules.is_applicable('multi_platform_rhosp', 'rhosp13')
```
shawndwells: thanks for adding this! will make adding future versions easier.

PR #3582, opened 2018-11-06
Add prodtype to OCP to prevent polluting non-ocp builds

redhatrises:
Prevents OpenShift content from being added to non-OpenShift builds.


applications/openstack/cinder/group.yml
```diff
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: osp13
+prodtype: rhosp13
```
shawndwells: thanks for fixing these, too

PR #3584, opened 2018-11-06
Add Crypto Policy for Kerberos

yuumasato:
#### Description:

- The rule ensures that kerberos adheres to the system-wide crypto policy.

#### Rationale:

- Kerberos could be configured to opt-out of the system-wide crypto policies.

### DIFF
--- a/linux_os/guide/system/software/integrity/crypto/configure_kerberos_crypto_policy/ansible/fedora.yml
+++ b/linux_os/guide/system/software/integrity/crypto/configure_kerberos_crypto_policy/ansible/fedora.yml
@@ -1,6 +1,6 @@
 # platform = multi_platform_fedora
-# reboot = false
-# strategy = restrict
+# reboot = true
+# strategy = configure
 # complexity = low
 # disruption = low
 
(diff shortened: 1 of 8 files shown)


redhatrises:
@yuumasato the content should reside under `linux_os/guide/system/software/integrity/crypto`.

yuumasato:
I have relocated the rule.

redhatrises:
@yuumasato one more thing, please remove the `<br/>` from OCIL.

yuumasato:
@redhatrises fixed, thanks for review.

linux_os/guide/system/software/integrity/crypto/configure_kerberos_crypto_policy/ansible/fedora.yml
```diff
@@ -0,0 +1,13 @@
+# platform = multi_platform_fedora
+# reboot = false
```
matusmarhefka: Shouldn't this be rather `true`? If the configuration is already loaded linking the file won't reload it right?

yuumasato: The service can be either restarted, or the system rebooted.
I have updated the remediation metadata in ansible fix, and added metadata to bash fix.

matusmarhefka:
LGTM

PR #3585, opened 2018-11-06
Add ocp-node checks to master

redhatrises:
I accidentally removed this when I shouldn't have.

### DIFF
--- a/ocp3/profiles/opencis-ocp-master.profile
+++ b/ocp3/profiles/opencis-ocp-master.profile
@@ -12,6 +12,8 @@ description: |-
     ensure a system is in compliance or consistency with the CIS
     baseline.
 
+extends: opencis-ocp-node
+
 selections:
     - file_groupowner_etc_origin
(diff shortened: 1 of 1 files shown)


shawndwells:
Should there be two OCP profiles? Would there be a simplified user experience if there was a single profile (e.g. STIG) that contained logic to evaluate node/master as appropriate?

shawndwells:
vs having ``--profile ocp-stig-node`` ``--profile ocp-stig-master``

redhatrises:
> vs having `--profile ocp-stig-node` `--profile ocp-stig-master`

How would you envision doing that?

jan-cerny:
We could have CPEs to check if the target is OCP node or OCP master and set the `platform:` in each rule YAML. It would require to improve handling of this key in our build system, for example we will need to support a list instead of a single value there.

jan-cerny:
But let's do that in separate PR.

PR #3586, opened 2018-11-06
Removed controller_service_account_private_key

shawndwells:
openshift/api-server/api_server_service_account_key/rule.yml  evaluates the same config setting.

Upstream kubernetes breaks the private keys into their own config options (e.g. API, controller...) whereas
openshift creates this private key and maps the configuration without any intervention requred


redhatrises:
ACK

PR #3592, opened 2018-11-07
Check if Libreswan is configured to use crypto-policies

jan-cerny:
#### Description:
Adds a rule that check if libreswan configuration adheres to system-wide crypto policy.Also adds tests for SSG Test Suite for this rule.

#### Rationale:
Libreswan shouldn't be configured to opt out from system-wide crypto policies.


redhatrises:
@jan-cerny the content should reside under `linux_os/guide/system/software/integrity/crypto`.

jan-cerny:
@redhatrises moved

redhatrises:
LGTM.

PR #3593, opened 2018-11-07
Added Crypto Policy rule for SSH

matusmarhefka:
#### Description:

- The rule `configure_ssh_crypto_policy` was added including OVAL check,
bash and ansible remediations.
- Also added a test for the rule (into SSG Test Suite).

#### Rationale:

- SSH could be configured to opt-out of the system-wide crypto policy, this rule should prevent that situation.

### DIFF
--- a/fedora/profiles/standard.profile
+++ b/fedora/profiles/standard.profile
@@ -87,3 +87,4 @@ selections:
     - sshd_idle_timeout_value=5_minutes
     - sshd_set_idle_timeout
     - sshd_set_keepalive
+    - configure_ssh_crypto_policy
--- a/linux_os/guide/system/software/integrity/crypto/configure_ssh_crypto_policy/ansible/shared.yml
+++ b/linux_os/guide/system/software/integrity/crypto/configure_ssh_crypto_policy/ansible/shared.yml
@@ -1,0 +1,13 @@
+# platform = multi_platform_fedora
+# reboot = true
+# strategy = disable
+# complexity = low
+# disruption = medium
+
+- name: "@RULE_TITLE@"
+  lineinfile:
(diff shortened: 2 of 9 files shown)


linux_os/guide/system/crypto_policies/configure_ssh_crypto_policy/ansible/shared.yml
```diff
@@ -0,0 +1,13 @@
+# platform = multi_platform_fedora
+# reboot = true
+# strategy = disable
+# complexity = low
+# disruption = medium
+
+- name: "@RULE_TITLE@"
+  lineinfile:
+    dest: /etc/sysconfig/sshd
+    state: absent
```
redhatrises: @matusmarhefka I believe that this should be `state: present`

redhatrises:
@matusmarhefka the content should reside under `linux_os/guide/system/software/integrity/crypto`.

linux_os/guide/system/crypto_policies/configure_ssh_crypto_policy/ansible/shared.yml
matusmarhefka: @redhatrises Definitely not, if this regexp would be present in the `/etc/sysconfig/sshd` it would mean that sshd would opt out from the system-wide crypto policy which is exactly what this rule should prevent.

matusmarhefka:
@redhatrises Thanks, I have moved the rule under `linux_os/guide/system/software/integrity/crypto`.

redhatrises:
> @redhatrises Thanks, I have moved the rule under `linux_os/guide/system/software/integrity/crypto`.

@matusmarhefka you are right... I think I looked at the OVAL wrong.

LGTM

PR #3594, opened 2018-11-07
Added Crypto Policy rule for OpenSSL.

matejak:
The rule ensures that the library adheres to the system-wide crypto policy by default.

Remarks:

* The rule is part of the Fedora standard profile.
* The SSG tests pass, I recommend using the Docker backend.
* There is no Ansible fix for the rule, as the OpenSSL config file is something between an ini file and normal config file. The fix is probably doable, but it is far from being trivial.

### DIFF
--- a/linux_os/guide/system/software/integrity/crypto/configure_openssl_crypto_policy/oval/shared.xml
+++ b/linux_os/guide/system/software/integrity/crypto/configure_openssl_crypto_policy/oval/shared.xml
@@ -1,0 +1,29 @@
+<def-group>
+  <definition class="compliance" id="configure_openssl_crypto_policy" version="1">
+    <metadata>
+      <title>Configure OpenSSL to use System Crypto Policy</title>
+      <affected family="unix">
+        <platform>multi_platform_fedora</platform>
+      </affected>
+      <description>OpenSSL should be configured to use the system-wide crypto policy setting.</description>
+    </metadata>
+    <criteria>
+      <criterion test_ref="test_configure_openssl_crypto_policy"
+      comment="Check that the configuration mandates usage of system-wide crypto policies." />
+    </criteria>
+  </definition>
+
+  <ind:textfilecontent54_test id="test_configure_openssl_crypto_policy"
+  comment="Check that the configuration mandates usage of system-wide crypto policies."
+  check="all" check_existence="all_exist" version="1">
+    <ind:object object_ref="object_configure_openssl_crypto_policy" />
+  </ind:textfilecontent54_test>
+
+  <ind:textfilecontent54_object id="object_configure_openssl_crypto_policy"
+  version="1">
+    <ind:filepath>/etc/pki/tls/openssl.cnf</ind:filepath>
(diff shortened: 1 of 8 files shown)


matejak:
@redhatrises Moved in a rebase.

linux_os/guide/system/software/integrity/crypto/configure_openssl_crypto_policy/oval/shared.xml
```diff
@@ -0,0 +1,29 @@
+<def-group>
+  <definition class="compliance" id="configure_openssl_crypto_policy" version="1">
+    <metadata>
+      <title>Configure OpenSSL to use System Crypto Policy.</title>
```
redhatrises: @matejak please remove `.` from the titlle

linux_os/guide/system/software/integrity/crypto/configure_openssl_crypto_policy/rule.yml
```diff
@@ -0,0 +1,31 @@
...
+
+ocil_clause: |-
+    the OpenSSL config file doesn't contain the whole section,
+    or that the section doesn't have the <pre>.include /etc/crypto-policies/back-ends/openssl.config</pre> directive
+
+ocil: |-
+    To verify that OpenSSL uses the system crypro policy, check out that the OpenSSL config file
+    <pre>/etc/pki/tls/openssl.cnf</pre> contains the <pre>[ crypto_policy ]</pre> section with the
+    <pre>.include /etc/crypto-policies/back-ends/openssl.config</pre> directive:
+    <pre>grep -A 3 '\[\s*crypto_policy\s*\]\s*$' /etc/pki/tls/openssl.cnf | grep '^\s*\.include \s*/etc/crypto-policies/back-ends/openssl.config\s*$' </pre>.
```
redhatrises: @matejak is there a way to make this grep easier? It's really cumbersome which makes it harder for auditors.

linux_os/guide/system/software/integrity/crypto/configure_openssl_crypto_policy/oval/shared.xml
matejak: done.

linux_os/guide/system/software/integrity/crypto/configure_openssl_crypto_policy/rule.yml
matejak: I can only make it more complicated :-( It is not technically possible to ensure that an `.ini` file has correct syntax just with grep.

redhatrises: @matejak let's make the grep the following:

`grep -A3 crypto_policy /etc/pki/tls/openssl.cnf`

Since OCIL is used for audits, we don't really need to worry about handling the syntax, and it's much easier to deal with.

matejak: OK, I went for a simplification that checks that the policies config file is at least mentioned in the right context, but it doesn't guarantee at all that it is present in the right section :-(

redhatrises:
@matejak thank you for your work on this stuff and for the changes! Do you mind rebasing?

matejak:
@redhatrises  Rebased, thanks for the review!

redhatrises:
ACK. Thanks @matejak

PR #3596, opened 2018-11-09
Chage RPM permission verify to use awk

tedbrunell:
#### Description:

Changed the grep command to awk to more accurately determine flags returned from rpm -Va during file verification.

#### Rationale:

The grep command that was used searched for lines starting with '.M'.  If the file size is also different, the line will start with 'SM' and cause the test to fail.  This change makes the check only look at the second letter, representing a permission change.


redhatrises:
@openscap-ci add to whitelist

jan-cerny:
@tedbrunell  Thank you very much for this pull request. Nice catch!

Please remove the merge commit from the PR. You can use interactive git rebase to do that. Also if you don't want to destroy your remotes, you can create a new branch for it.

tbrunell:
Thanks, Jan.  Can you delete the PR?

I am going to more closely follow the developer guide and will resubmit it
with additional patches for the rpm_verify_ownership as well.

Thanks,
Ted

PR #3598, opened 2018-11-12
Replace grep with awk for rpm_verification tests as it more acurately…

tedbrunell:
… detects flags that are returned by rpm -Va.

#### Description:

Replaces grep with awk to better test flags that are returned by rpm -vA during rpm_verification

#### Rationale:

When testing rpm_verification, the search string that grep was using would only match if the flag was prefaced with a "." creating false returns if something else had changed with the file such as the size or if owner and group membership had both been modified since the file was installed by RPM.  These patches fix that in the check content as well as bash and Ansible by using awk to look at the single flag for permission modified, owner modification or a change to group membership.


jan-cerny:
@tedbrunell Looks good to me. Thank you!

PR #3600, opened 2018-11-14
Fix product naming in OCP content

redhatrises:
- Remove RHEL and Enterprise wording from OCP content


shawndwells:
thanks @redhatrises

PR #3602, opened 2018-11-15
Update configure_crypto_policy rule and add tests for it

mildass:
Update Configure crypto policy rule so it checks symlinks for all crypto libraries (`/etc/crypto-policies/back-ends/*.config -> /usr/share/crypto-policies/POLICY/*.txt`) and add tests for this rule.

### DIFF
--- a/linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/oval/fedora.xml
+++ b/linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/oval/fedora.xml
@@ -1,3 +1,34 @@
+{{%- macro crypto_policy_symlink_criterion(library) %}}
+      <criterion comment="Check if /etc/crypto-policies/back-ends/{{{ library }}}.config is linked to /usr/share/crypto-policies/POLICY/{{{ library }}}.txt" test_ref="test_crypto_policy_{{{ library }}}_symlink" />
+{{%- endmacro %}}
+
+{{%- macro crypto_policy_symlink_check(library) %}}
+  <local_variable id="var_crypto_policy_{{{ library }}}_path" datatype="string" comment="Regex variable for canonical path to targeted {{{ library }}} policy" version="1">
+    <concat>
+      <literal_component>^/usr/share/crypto-policies/</literal_component>
+      <variable_component var_ref="var_system_crypto_policy"/>
+      <literal_component>/{{{ library }}}.txt$</literal_component>
+    </concat>
+  </local_variable>
+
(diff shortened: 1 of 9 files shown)


linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/oval/fedora.xml
```diff
@@ -1,3 +1,34 @@
+{{%- macro crypto_policy_symlink_criterion(library) %}}
+      <criterion comment="Check if /etc/crypto-policies/back-ends/{{{ library }}}.config is linked to /usr/share/crypto-policies/POLICY/{{{ library }}}.txt" test_ref="test_crypto_policy_{{{ library }}}_symlink" />
+{{%- endmacro %}}
+
+{{%- macro crypto_policy_symlink_check(library) %}}
+  <local_variable id="var_crypto_policy_{{{ library }}}_path" datatype="string" comment="Regex variable for canonical path to itargeted {{{ library }}} policy" version="1">
```
yuumasato: typo: `itargeted`

linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/oval/fedora.xml
```diff
@@ -1,3 +1,34 @@
...
+    <concat>
+      <literal_component>^/usr/share/crypto-policies/</literal_component>
+      <variable_component var_ref="var_system_crypto_policy"/>
+      <literal_component>/{{{ library }}}.txt$</literal_component>
+    </concat>
+  </local_variable>
+
+  <unix:symlink_test id="test_crypto_policy_{{{ library }}}_symlink"
+  comment="check if /etc/crypto-policies/back-ends/{{{ library }}}.config is linked to crypto-policies"
```
yuumasato: Suggestion for comment: `check if /etc/crypto-policies/back-ends/{{{ library }}}.config is linked to correct target library config`

mildass:
@yuumasato About macro  - for that it would be needed to add macros expanding from `macros.jinja` for OVALs to build scripts. I was looking to these build scripts and I'm not sure what I need to do to achieve that. I will try to find out and if it would be some minor change, then it's absolutely worthy. If it needs bigger change in build, then I'm not sure, if it's worthy because of this one rule or if it would be useful for more rules.
About other things - updated and added test scenario (hope that you meant test scenario like this, that makes symlink which points to file from different policy)

yuumasato:
@mildass The added test looks good.

>  I will try to find out and if it would be some minor change, then it's absolutely worthy.

Thanks for looking into it, and actually it can be done in following PRs. There is no reason to block this PR.

PR #3607, opened 2018-11-19
Fix Ansible remediations to be parseable by oscap

### LINKED ISSUE #3597: Regression: Ansible Playbooks have "vars:" empty and contain "set_fact"
#### Description of problem:
SCAP content uses "XCCDF values" to parametrize rules. We also use them to customize Ansible tasks.

In our XCCDFs or datastreams we have Ansible snippets in `<fix>` element. If the Ansible snippet is parametrized by XCCDF value  it contains `set_fact:` that defines a variable and the value is substituted by `<sub>` element. In our RHEL7 datastream it looks like this:

```
<fix id="accounts_password_pam_minlen" system="urn:xccdf:fix:script:ansible" complexity="low" disruption="low" strategy="restrict">- name: XCCDF Value var_password_pam_minlen # promote to      variable
  set_fact:
    var_password_pam_minlen: "<sub idref="xccdf_org.ssgproject.content_value_var_password_pam_minlen" use="legacy"/>"
  tags:  
    - always 
```

When `oscap` generates Ansible Playbook from an XCCDF or Datastream using `oscap xccdf generate fix` it looks for these sections. It extracts the variables and their values and drops these sections. Finally, it puts the list of variables at the beginning of the playbook in `vars:`.

The problem is that this behavior is broken. With current ComplianceAsCode it doesn't work  anymore. The Ansible Playbooks that we build now contain many `XCCDF Value ...` tasks with `set_fact`  and the list of `vars` is empty.

I have found using `git bisect` that the regression is caused by 95e9d5130f7b20677af0fd8b23b8fb2ad0900d5b . This commit changes the format of `set_fact` clause in ComplianceAsCode. Therefore `oscap` cannot match it.

The OpenSCAP code that handles the variables in Ansible Playbooks is here: https://github.com/OpenSCAP/openscap/blob/maint-1.2/src/XCCDF_POLICY/xccdf_policy_remediate.c#L586
Also see https://github.com/OpenSCAP/openscap/pull/896 for context.

current upstream head as of 2018-11-12

all

#### Steps to Reproduce:

1. cd build
2. cmake ..
3. make rhel7
4. vim roles/ssg-rhel7-role-ospp.yml (just an example, all playbooks are affected)

#### Actual Results:
`vars:` is empty
```
 - hosts: all
   pre_tasks:
     - name: Verify Ansible meets SCAP-Security-Guide version requirements.
       assert:
         that: "ansible_version.full is version_compare('2.3', '>=')"
         msg: >
           "You must update Ansible to at least version 2.3 to use this role."
          
   vars:
   tasks:
    - name: Disable service zebra
      service:
```

#### Expected Results:
something like this - see SSG 0.1.39 "roles"
```
 - hosts: all
   pre_tasks:
     - name: Verify Ansible meets SCAP-Security-Guide version requirements.
       assert:
         that: "ansible_version.full is version_compare('2.3', '>=')"
         msg: >
           "You must update Ansible to at least version 2.3 to use this role."
          
   vars:
      sshd_listening_port: "22"
      sshd_approved_macs: "hmac-sha2-512,hmac-sha2-256,hmac-sha1,hmac-sha1-etm@openssh.com,hmac-sha2-256-etm@openssh.com,hmac-sha2-512-etm@openssh.com"
      sshd_idle_timeout_value: "600"
      var_sssd_memcache_timeout: "300"
      var_accounts_passwords_pam_faillock_deny: "3"
      var_accounts_passwords_pam_faillock_unlock_time: "never"
      var_accounts_passwords_pam_faillock_fail_interval: "900"
..... [snip] .....
      var_auditd_max_log_file: "6"
      var_auditd_max_log_file_action: "rotate"
      var_auditd_flush: "data"
      var_auditd_admin_space_left_action: "single"
      var_auditd_action_mail_acct: "root"
      var_auditd_space_left_action: "email"
      var_auditd_num_logs: "5"
   tasks:
    - name: Disable service zebra
      service:
```

This regression breaks generated Ansible Roles published on RedHatOfficial.

A regression test strongly needed in CI.

As pointed out in https://github.com/ComplianceAsCode/content/issues/3370 `set_fact` effectively makes it impossible for users to customize the Playbooks in their Ansible Inventories using host/groups variables because `set_fact` has top priority.

jan-cerny:
#### Description:

This changes the way the XCCDF Values are inserted into Ansible snippets in XCCDF `fix` elements.
It makes it again working with code in OpenSCAP that extracts them and puts them to `vars:` section at the beginning of generated playbook.

#### Rationale:
Fixes #3597.

### DIFF
--- a/ssg/build_remediations.py
+++ b/ssg/build_remediations.py
@@ -301,15 +301,20 @@ def expand_xccdf_subs(fix, remediation_type, remediation_functions):
                 "substituting directly."
             )
 
-        # we use the horrid "!!str |-" syntax to force strings without using
-        # quotes. quotes enable yaml escaping rules so we'd have to escape all
-        # the backslashes and at this point we don't know if there are any.
+        # If you change this string make sure it still matches the pattern
+        # defined in OpenSCAP. Otherwise you break variable handling in
+        # 'oscap xccdf generate fix' and the variables won't be customizable!
+        # https://github.com/OpenSCAP/openscap/blob/1.2.17/src/XCCDF_POLICY/xccdf_policy_remediate.c#L588
+        #   const char *pattern =
+        #     "- name: XCCDF Value [^ ]+ # promote to variable\n  set_fact:\n"
+        #     "    ([^:]+): (.+)\n  tags:\n    - always\n";
+        # We use !!str typecast to prevent treating values as different types
+        # eg. yes as a bool or 077 as an octal number
         fix_text = re.sub(
             r"- \(xccdf-var\s+(\S+)\)",
             r"- name: XCCDF Value \1 # promote to variable\n"
             r"  set_fact:\n"
-            r"    \1: !!str |-\n"
-            r"        (ansible-populate \1)\n"
+            r"    \1: !!str (ansible-populate \1)\n"
             r"  tags:\n"
             r"    - always",
             fix_text


jan-cerny:
@yuumasato That needs to be fixed in OpenSCAP. OpenSCAP generates the output in a stupid way - it iterates over all `fix` elements and it puts everything to output immediately, it doesn't check if the variable has already been defined in a different rule. I'm going to provide a patch for OpenSCAP.

jan-cerny:
@yuumasato https://github.com/OpenSCAP/openscap/pull/1274 should help with this problem.

matejak:
test this please

jan-cerny:
This PR doesn't have to wait for https://github.com/OpenSCAP/openscap/pull/1274 because https://github.com/OpenSCAP/openscap/pull/1274 is a PR for OpenSCAP and you'll still build ComplianceAsCode using old version of OpenSCAP until the new version of OpenSCAP will be released and delivered to you by Fedora updates. This PR should work with older versions of OpenSCAP. The problem with duplicated variables was also present in our playbooks before regression #3597 happened.

yuumasato:
@jan-cerny Thanks for the not, I was not aware that the duplicated variables already occurred before this regression.
Moreover, the last definition of the variable will be used, which is the same.
Ack.

PR #3608, opened 2018-11-19
Fix ansible-lint mount option errors

yuumasato:
#### Description:

- Tasks that gather information never make changes to the system.
Lets make them return result as not changed.
- Skip ANSIBLE0006 warnings in tasks that use shell module and mount command to fetch data about mount points.

#### Rationale:

- Ansible mount module cannot be used to fetch data.

- Fixes following ansible-lint messages:
```
Task: get back device associated to mountpoint
- [EANSIBLE0006] mount used in place of mount module
- [EANSIBLE0012] Commands should not change things if nothing needs doing
```


matejak:
LGTM, this is a legitimate tweak of usage of the `shell` Ansible module.

PR #3610, opened 2018-11-19
Galaxy Upload script refactoring.

matejak:
This PR features refactoring of the script that is not supposed to alter it's behavior - if called in the same way as the pre-refactoring script, it should produce exactly the same output (except e.g. random ordering issues).

Notable changes:
* Split the large `update_repository` function into smaller methods of the newly-introduced `Repository` class.
* Backwards-compatible CLI changes: Allowed specification of alternative organization, allowed restriction of processed profiles.

### DIFF
--- a/utils/upload_ansible_roles_to_galaxy.py
+++ b/utils/upload_ansible_roles_to_galaxy.py
@@ -43,6 +43,17 @@ yaml.add_representer(collections.OrderedDict, dict_representer)
 yaml.add_constructor(_mapping_tag, dict_constructor)
 # End arcaduf gist
 
+PROFILE_WHITELIST = set([
+    "C2S",
+    "hipaa",
+    "nist-800-171-cui",
+    "ospp",
+    "pci-dss",
+    "rht-ccp",
+    "stig-rhel7-disa",
+])
+
+
 ORGANIZATION_NAME = "RedHatOfficial"
 GIT_COMMIT_AUTHOR_NAME = "Red Hat Security Automation development team"
 GIT_COMMIT_AUTHOR_EMAIL = "scap-security-guide@lists.fedorahosted.org"
@@ -271,7 +282,8 @@ def parse_args():
         help="Name of the Github organization")
     parser.add_argument(
         "--profile", "-p", default=[], nargs="*", action="append",
-        help="What profiles to upload, if not specified, upload them all.")
+        metavar="PROFILE", choices=PROFILE_WHITELIST,
+        help="What profiles to upload, if not specified, upload all that are applicable.")
     return parser.parse_args()
 
 
@@ -290,15 +302,7 @@ def locally_clone_and_init_repositories(organization, repo_list):
 def main():
     args = parse_args()
 
-    role_whitelist = set([
-        "rhel7-role-C2S",
-        "rhel7-role-hipaa",
-        "rhel7-role-nist-800-171-cui",
-        "rhel7-role-ospp",
-        "rhel7-role-pci-dss",
-        "rhel7-role-rht-ccp",
-        "rhel7-role-stig-rhel7-disa"
-    ])
+    role_whitelist = {"rhel7-role-%s" % p for p in PROFILE_WHITELIST}
     if args.profile:
         selected_roles = {"rhel7-role-%s" % p for p in args.profile}
         role_whitelist.intersection_update(selected_roles)
--- a/utils/upload_ansible_roles_to_galaxy.py
+++ b/utils/upload_ansible_roles_to_galaxy.py
@@ -38,6 +38,7 @@ def dict_representer(dumper, data):
 def dict_constructor(loader, node):
     return collections.OrderedDict(loader.construct_pairs(node))
 
+
 yaml.add_representer(collections.OrderedDict, dict_representer)
 yaml.add_constructor(_mapping_tag, dict_constructor)
 # End arcaduf gist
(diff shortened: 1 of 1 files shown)


utils/upload_ansible_roles_to_galaxy.py
```diff
@@ -82,132 +83,178 @@ def clone_and_init_repository(parent_dir, repo):
...
-    if vars_local_content != vars_remote_content.decoded_content:
-        repository.update_file(
-            "/vars/main.yml",
-            "Updates vars/main.yml",
-            vars_local_content,
-            vars_remote_content.sha,
-            author=InputGitAuthor(
-                GIT_COMMIT_AUTHOR_NAME, GIT_COMMIT_AUTHOR_EMAIL)
-        )
+class Repository(object):
```
yuumasato: I suggest to call it `Role`.

utils/upload_ansible_roles_to_galaxy.py
```diff
@@ -82,132 +83,178 @@ def clone_and_init_repository(parent_dir, repo):
...
+    def __init__(self, repo, local_filename):
+        self.repository = repo
```
yuumasato: And this would be `remote_role` or `tracked_role`.

yuumasato:
@matejak Also, how have you tested that generation of the roles? Did you run and update the Roles in the repository?

utils/upload_ansible_roles_to_galaxy.py
matejak: I went for `remote_repo` - it is true that it is remote, but the type and usage of the variable are determined by the fact that it is a wrapper of a Github repository.

matejak:
@yuumasato Yes, I have tested #3611 [with the HIPAA profile](https://github.com/Ansible-Security-Compliance/rhel7-role-hipaa/commits/master). It looks like that the profile has changed but there was no obvious breaking behavior observed.

utils/upload_ansible_roles_to_galaxy.py
```diff
@@ -219,9 +266,27 @@ def parse_args():
         help="Path to directory containing the generated roles. Most "
         "likely this is going to be ./build/roles",
         dest="build_roles_dir")
+    parser.add_argument(
+        "--organization", "-o", default=ORGANIZATION_NAME,
+        help="Name of the Github organization")
+    parser.add_argument(
+        "--profile", "-p", default=[], nargs="*", action="append",
+        help="What profiles to upload, if not specified, upload them all.")
```
yuumasato: This may give the impression that it will upload all profiles, while only the ones in allowed list will be uploaded.

yuumasato:
@matejak I was looking into the RedHatOfficial for updates, ;P

jan-cerny:
Is there a non-destructive way of testing it?

utils/upload_ansible_roles_to_galaxy.py
```diff
@@ -82,132 +83,178 @@ def clone_and_init_repository(parent_dir, repo):
...
+        self.gather_data()
+        self._reformat_local_content()
+        self.title = re.search(
+            r'Profile Title:\s+(.+)$', self.description, re.MULTILINE).group(1)
+        # Why so?
+        self.description = self.description.replace('\n', '  \n')
+
+        self._update_tasks_content_if_needed()
+        self._update_vars_content_if_needed()
+        self._update_readme_content_if_needed()
```
redhatrises: Updating the readme should only be done when the repository is first initialized with the content. Reason being is that we can add build tags which is a nice add in Galaxy. Also the role name can change to comply with latest ansible naming requirements as well as make it easier to use the role.

utils/upload_ansible_roles_to_galaxy.py
```diff
@@ -82,132 +83,178 @@ def clone_and_init_repository(parent_dir, repo):
...
+        self._update_meta_content_if_needed()
```
redhatrises: Updating the meta content should only be done when the repository is first initialized with the content. Reason being is the meta tags are completely wrong and will deviate by profile to match accurately what the profile is and does.

utils/upload_ansible_roles_to_galaxy.py
```diff
@@ -82,132 +83,178 @@ def clone_and_init_repository(parent_dir, repo):
...
+
+        repo_description = (
+            "{title} - Ansible role generated from the SCAP Security Guide project"
```
redhatrises: s/SCAP Security Guide/ComplianceAsCode/

matejak:
@jan-cerny The way how to test it involves using the "old" Galaxy. Then, you can clone the repository, throw away some commits, and force push to undo those changes.

utils/upload_ansible_roles_to_galaxy.py
matejak: done.

matejak:
@redhatrises This PR is a refactoring - it doesn't aim to change the script's behavior. Those readme and meta changes that you are suggesting are actual changes. As this PR is already big as it is, I suggest merging it and opening another one to change the script's logic.

yuumasato:
I'm for merging this and changing behavior in subsequent PRs.

utils/upload_ansible_roles_to_galaxy.py
yuumasato: @redhatrises Can you point us to some documentation regarding this build tags?

PR #3611, opened 2018-11-19
 Added support for default role variables

matejak:
The script now generates `when:` clauses to each task using newly introduced variables.
Those variables are extracted from existing tags. For every variable added, a corresponding entry in `defaults/main.yml` is created, where the variable is set to true, so the playbook behaves in the same way as before this change.

~For reference, this PR basically contains #3610.~

### DIFF
--- a/utils/upload_ansible_roles_to_galaxy.py
+++ b/utils/upload_ansible_roles_to_galaxy.py
@@ -103,6 +103,7 @@ class Role(object):
         self.vars_data = []
         self.tasks_data = []
         self.pre_tasks_data = []
+        self.added_variables = set()
 
         self.tasks_local_content = None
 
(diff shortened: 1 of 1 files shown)


jan-cerny:
Blocked by #3610

jan-cerny:
Hi @matejak Could you share with us why it isn't done the same way at the same time `tags:` is created?

matejak:
> Hi @matejak Could you share with us why it isn't done the same way at the same time `tags:` is created?
This can only work with ansible roles. We don't build roles in SSG, we build only self-contained playbooks. However, the upload script takes those playbooks one by one and generates role out of each of them. It would be more elegant to generate roles directly from the content, and basically just copying those to the remote repository, and this is a significant shortcut.

> Have you also considered tasks that already contain `when:`?
Yes, check out the function `add_variables_to_tasks`, it handles all possible cases related to `when:`.

utils/upload_ansible_roles_to_galaxy.py
```diff
@@ -135,12 +136,55 @@ def gather_data(self):
...
-            self.tasks_data, width=120, indent=4, default_flow_style=False)
-
         description = self.get_description_from_filedata(filedata)
         self.description = description
 
+    def add_variables_to_tasks(self):
+        for task in self.tasks_data:
+            if "when" not in task:
+                task["when"] = []
+            elif isinstance(task["when"], str):
```
jan-cerny: What happens if the task `when` is a list? Eg.
```
when:
  - aaa
  - bbb
```

matejak: Then, the `task["when"]` is a list, and `add_variables_to_task` will simply append another items to the list.

utils/upload_ansible_roles_to_galaxy.py
```diff
@@ -135,12 +136,57 @@ def gather_data(self):
...
+        variables_to_add = {tag for tag in task["tags"] if self.tag_is_valid_variable(tag)}
+        task["when"] += ["{{{{ {varname} | bool }}}}".format(varname=v) for v in variables_to_add]
+        self.added_variables.update(variables_to_add)
+
+    def add_task_variables_to_default_variables_if_needed(self):
+        default_vars_to_add = sorted(self.added_variables)
+        lines = [
+            "---", "# defaults file for {role_name}\n".format(role_name=self.remote_repo.name),
+        ]
+        lines += ["{var_name}: true".format(var_name=var_name) for var_name in default_vars_to_add]
```
jan-cerny: It creates a variable in single quotes`'`. Shouldn't it be in double quotes `"`?

matejak: I have found out that it is the lower level of the code that applies single quotes, as it detects that an item is a string, so it is quoted. In this case, the difference between single and double quotes is not important, as it has something to do with escaping non-ASCII chars.
Therefore, I propose to ignore this matter.

matejak:
Concerning variables, the recent build + upload script export variables from tasks to `vars/main.yml`, i.e. to [role variables](https://docs.ansible.com/ansible/2.7/user_guide/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable), see also `update_vars_content_if_needed`. This looks like wanted behavior, but it is true that those variables can't be overriden in inventory files.
The solution would be to abandon role variables, and start merging them to defaults, but this would mean slight backward-incompatibility and removal of some code from the script. I feel that this should be discussed in another PR.

utils/upload_ansible_roles_to_galaxy.py
jan-cerny: @matejak Aha OK. But a different way of question: Does Ansible expand the Jinja 2 templates if the templates are inside a single-quoted string?

jan-cerny:
@matejak I'm fine with doing that in another PR. The question is if we want to eg. enable to customize the minimum password length in inventory files or not.

utils/upload_ansible_roles_to_galaxy.py
matejak: I think so. In the [Ansible documentation](https://docs.ansible.com/ansible/2.5/user_guide/playbooks_filters.html#combination-filters), there is pointed out that jinja2 is expanded behind double quotes. And there is no indication anywhere that single quotes are treated differently. And it also is the expected behavior - jinja2 looks for curly braces, not being disturbed by any other chars, and YAML vs quotes has been touched in my previous comment.

matejak:
@jan-cerny About those tags, I thought of "why not" - the current implementation doesn't switch anything off by default, but I can imagine that someone wants to disable high disruption remediations. I think that it is OK to have it like that, introducing whitelist or blacklist of tags and default variables can be done later.

jan-cerny:
@matejak Great job!

PR #3612, opened 2018-11-19
Ignore warnings on tasks using rpm shell command

yuumasato:
#### Description:

- Do not warn about ANSIBLE0006 on these shell `rpm` tasks 

#### Rationale:

- The tasks cannot be executed by rpm module

- `ansible-lint` will ignore messages like the following:
```Task: Read list of files with incorrect ownership
- [EANSIBLE0006] rpm used in place of yum or rpm_key module

Task: Create list of uniq packages
- [EANSIBLE0006] rpm used in place of yum or rpm_key module

Task: Correct file ownership with RPM
- [EANSIBLE0006] rpm used in place of yum or rpm_key module
- [EANSIBLE0013] Use shell only when shell functionality is required
```


redhatrises:
LGTM

PR #3615, opened 2018-11-20
Add aarch64 64-bit OVAL check and add to 64-bit system checks

redhatrises:
Adds capability to verify 64-bit system checks to the 64-bit arm platform.


shawndwells:
Since we don't have ARM servers to test on, spot checked and looks ok. Thanks @redhatrises!

PR #3618, opened 2018-11-21
Replace screen with tmux

### LINKED ISSUE #3605: Packages missing during RHEL8 OSPP install
image

redhatrises:
- Fixes #3605


redhatrises:
One thing I notice is that `ctrl+b :lock-session` or `tmux lock-session` don't work. Wondering if we are now going to need an additional rule to check that tmux is configured so that a session lock does work.

jan-cerny:
@redhatrises There needs to be
```
set -g lock-command 'vlock -c'
```
in `/etc/tmux.conf`

and `vlock` command has to be present on the system to make the tmux lock working.

Would it make sense to have a rule that checks it?

jan-cerny:
@redhatrises Could you also change `linux_os/guide/system/accounts/accounts-physical/screen_locking/console_screen_locking/group.yml` to mention `tmux`?

redhatrises:
> @redhatrises There needs to be
> 
> ```
> set -g lock-command 'vlock -c'
> ```
> 
> in `/etc/tmux.conf`
> 
> and `vlock` command has to be present on the system to make the tmux lock working.
> 
> Would it make sense to have a rule that checks it?

@jan-cerny yes a separate rule will be needed which I will open a new PR for that new tmux rule.

PR #3619, opened 2018-11-22
Fix auditd_audispd_encrypt_sent_records on Fedora and RHEL8

jan-cerny:
#### Description:
In Audit 3.0 which is present in Fedora >= 29 and RHEL8, the
`audisp-remote.conf` moved to `/etc/audit` and the `enable_krb5`
option has been superseded by `transport` option.
See `man 5 audisp-remote.conf`.

#### Rationale:

### DIFF
--- a/linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_audispd_encrypt_sent_records/bash/shared.sh
+++ b/linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_audispd_encrypt_sent_records/bash/shared.sh
@@ -1,8 +1,14 @@
 # platform = multi_platform_all
 . /usr/share/scap-security-guide/remediation_functions
 
-var_enable_krb5="yes"
-
+{{% if product in ["rhel8", "fedora"] %}}
+AUDISP_REMOTE_CONFIG="/etc/audit/audisp-remote.conf"
+option="^transport"
+value="KRB5"
+{{% else %}}
 AUDISP_REMOTE_CONFIG="/etc/audisp/audisp-remote.conf"
+option="^enable_krb5"
+value="yes"
+{{% endif %}}
 
-replace_or_append $AUDISP_REMOTE_CONFIG '^enable_krb5' "$var_enable_krb5" "@CCENUM@"
+replace_or_append $AUDISP_REMOTE_CONFIG "$option" "$value" "@CCENUM@"
--- a/linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_audispd_encrypt_sent_records/oval/shared.xml
+++ b/linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_audispd_encrypt_sent_records/oval/shared.xml
@@ -5,24 +5,36 @@
       <affected family="unix">
         <platform>multi_platform_all</platform>
       </affected>
+{{% if product in ["rhel8", "fedora"] %}}
+      <description>transport setting in /etc/audit/audisp-remote.conf is set to 'KRB5'</description>
+{{% else %}}
       <description>enable_krb5 setting in /etc/audisp/audisp-remote.conf is set to 'yes'</description>
+{{% endif %}}
     </metadata>
 
     <criteria>
-      <criterion comment="enable_krb5 setting in audisp-remote.conf" test_ref="test_auditd_audispd_encrypt_sent_records" />
+      <criterion comment="setting in audisp-remote.conf" test_ref="test_auditd_audispd_encrypt_sent_records" />
     </criteria>
 
   </definition>
 
-  <ind:textfilecontent54_test check="all" comment="enable_krb5 setting in audisp-remote.conf" id="test_auditd_audispd_encrypt_sent_records" version="1">
+  <ind:textfilecontent54_test check="all" comment="setting in audisp-remote.conf" id="test_auditd_audispd_encrypt_sent_records" version="1">
     <ind:object object_ref="object_auditd_audispd_encrypt_sent_records" />
   </ind:textfilecontent54_test>
 
   <ind:textfilecontent54_object id="object_auditd_audispd_encrypt_sent_records" version="1">
+{{% if product in ["rhel8", "fedora"] %}}
+    <ind:filepath>/etc/audit/audisp-remote.conf</ind:filepath>
+{{% else %}}
     <ind:filepath>/etc/audisp/audisp-remote.conf</ind:filepath>
+{{% endif %}}
     <!-- Allow only space (exactly) as delimiter -->
     <!-- Require at least one space before and after the equal sign -->
+{{% if product in ["rhel8", "fedora"] %}}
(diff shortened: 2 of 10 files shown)


mildass:
@jan-cerny Sorry, it was my fault. I didn't delete build folder before building and it caused that problem.

linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_audispd_encrypt_sent_records/bash/shared.sh
```diff
@@ -1,8 +1,14 @@
 # platform = multi_platform_all
 . /usr/share/scap-security-guide/remediation_functions
 
-var_enable_krb5="yes"
-
+{{% if product in ["rhel8", "fedora"] %}}
+AUDISP_REMOTE_CONFIG="/etc/audit/audisp-remote.conf"
```
yuumasato: @jan-cerny According to manual, the transport option is set in `/etc/audit/auditd.conf` file.

linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_audispd_encrypt_sent_records/oval/shared.xml
```diff
@@ -5,24 +5,36 @@
       <affected family="unix">
         <platform>multi_platform_all</platform>
       </affected>
+{{% if product in ["rhel8", "fedora"] %}}
+      <description>transport setting in /etc/audit/audisp-remote.conf is set to 'KRB5'</description>
```
yuumasato: Please, change all instances of the config file for Fedora and RHEL8.

linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_audispd_encrypt_sent_records/rule.yml
```diff
@@ -31,8 +37,14 @@ ocil_clause: 'audispd is not encrypting audit records when sent over the network
 ocil: |-
     To verify the audispd plugin encrypts audit records off-loaded onto a different
```
yuumasato: In latest Fedora and RHEL8 audispd plugin is not used to off-load the records. This paragraph should be within the `else` case, and a paragraph for Fedora and RHEL8 added.

linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_audispd_encrypt_sent_records/oval/shared.xml
```diff
@@ -5,24 +5,36 @@
...
+{{% else %}}
       <description>enable_krb5 setting in /etc/audisp/audisp-remote.conf is set to 'yes'</description>
+{{% endif %}}
     </metadata>
 
     <criteria>
-      <criterion comment="enable_krb5 setting in audisp-remote.conf" test_ref="test_auditd_audispd_encrypt_sent_records" />
+      <criterion comment="setting in audisp-remote.conf" test_ref="test_auditd_audispd_encrypt_sent_records" />
```
yuumasato: Since `audisp-remote` is not a common element for all products, the comment should be about enabling encryption for transport of audit records.

linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_audispd_encrypt_sent_records/rule.yml
```diff
@@ -7,9 +7,15 @@ title: 'Encrypt Audit Records Sent With audispd Plugin'
 description: |-
     Configure the operating system to encrypt the transfer of off-loaded audit
     records onto a different system or media from the system being audited.
+{{% if product in ["rhel8", "fedora"] %}}
+    Uncomment the <tt>transport</tt> option in <pre>/etc/audit/audisp-remote.conf</pre>,
```
yuumasato: In file `/etc/audit/auditd.conf` `transport` option is already set as `TCP`.

linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_audispd_encrypt_sent_records/bash/shared.sh
jan-cerny: Thank you. I had look into this again and I have found it's in both files.
Try `man auditd.conf` and `man audisp-remote.conf` and it's documented  in both of them.
Now I don't know. Should we check both files? Does one of them override the other?

linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_audispd_encrypt_sent_records/rule.yml
jan-cerny: You're right it's not commented.

linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_audispd_encrypt_sent_records/bash/shared.sh
jan-cerny: @yuumasato 

Based on discussion with Steve Grubb on the mailing list https://lists.fedorahosted.org/archives/list/scap-security-guide@lists.fedorahosted.org/thread/5PZF5UTWX3K3OGRH353SDNT5U5CJ4FB7/ , configuration in `/etc/audit/auditd.conf` would be for  the aggregating server rather than the client that is sending the data.

I think that this specific rule is about the client that off-loads the data to a server. Therefore `/etc/audit/audisp-remote.conf` should be a correct file in this rule.

linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_audispd_encrypt_sent_records/rule.yml
yuumasato: In `/etc/audit/audisp-remote.conf `, `transport` option is not commented also.

linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_audispd_encrypt_sent_records/rule.yml
yuumasato: This is ok, actually, audispd plugins are used.

yuumasato:
@jan-cerny See example of dirty workaround on how to write test scenarios for multiple products: https://github.com/ComplianceAsCode/content/tree/master/tests/data/group_system/group_auditing/group_auditd_configure_rules/group_audit_privileged_commands/rule_audit_rules_privileged_commands

jan-cerny:
@yuumasato I don't prefer workarounds. Instead I propose to handle multiple products by specifying platform in the scenarios. https://github.com/ComplianceAsCode/content/pull/3627

jan-cerny:
I have added test scenarios for this.
However, if you run them without #3627 the scenarios that don't apply to the tested system will error.

jan-cerny:
Rebased on master.

yuumasato:
The test scenarios run fine with the platform dependent test scenarios.
@jan-cerny Thank you for these fixes.

PR #3620, opened 2018-11-22
Fix auditd_audispd_syslog_plugin_activated on Fedora and RHEL8

jan-cerny:
#### Description:
In Audit 3.0 which is present in Fedora >= 29 and RHEL8, the
`syslog.conf` moved to `/etc/audit/plugins.d/`.

#### Rationale:


jan-cerny:
I have added test scenarios.

jan-cerny:
Rebased on master.

PR #3622, opened 2018-11-22
Update ensure_redhat_gpgkey_installed for RHEL8

jan-cerny:
#### Description:
Fedora rule ensure_fedora_gpgkey_installed shouldn't be applicable to RHEL8.
Option gpg --with-fingeprint no longer works in RHEL8.
The key fingerprint can be obtained by gpg --show-key instead.
To parse the results using --with-colons is recommended.
We have to change the filters to parse the key fingerprints
and therefore it is more convenient for us the have the hard-coded
fingerprints without spaces.
There was a bug in bash remediation GPG_OUT is defined as an array
but then it was handled as a variable, which means only the first
item in the array was checked.

#### Rationale:


redhatrises:
ACK

PR #3623, opened 2018-11-26
New rule to enable FIPS mode on RHEL8 and Fedora

jan-cerny:
#### Description:
On RHEL8, a different method is used to enable FIPS mode than on
RHEL6/7.  Package `dracut-fips` doesn't exist anymore, GRUB config file
doesn't need to be edited. Instead, a new utility `fips-mode-setup`
should be used to setup and configure FIPS mode. To verify that FIPS
mode is enabled, the following 2 conditions have to be fulfilled:
  1. /etc/system-fips exists
  2. /proc/sys/crypto/fips_enabled is set to 1

The same facts apply also for Fedora 29.
This commit adds a new rule `enable_fips_mode` for RHEL8 and Fedora,
which replaces the old rule `grub2_enable_fips_mode` in RHEL8 and Fedora
OSPP profiles. The platform in `grub2_enable_fips_mode` and dependent
rule `package_dracut-fips_installed` in changed to exclude RHEL8 and
Fedora. Currently it fails on both RHEL8 and Fedora because it depends
on rule `installed_OS_is_certified` which allows only RHEL6 and RHEL7.

#### Rationale:
The current way of checking FIPS mode doesn't apply to RHEL8 and Fedora.

### DIFF
--- a/linux_os/guide/system/software/integrity/fips/enable_fips_mode/oval/shared.xml
+++ b/linux_os/guide/system/software/integrity/fips/enable_fips_mode/oval/shared.xml
@@ -1,0 +1,19 @@
+<def-group oval_version="5.10">
+  <definition class="compliance" id="enable_fips_mode" version="1">
+    <metadata>
+      <title>Enable FIPS Mode</title>
+      <affected family="unix">
+        <platform>Red Hat Enterprise Linux 8</platform>
+        <platform>multi_platform_fedora</platform>
+      </affected>
+      <description>Check if FIPS mode is enabled on the system</description>
+    </metadata>
+    <criteria operator="AND">
+      <extend_definition comment="check /etc/system-fips exists" definition_ref="etc_system_fips_exists" />
+      <extend_definition comment="check sysctl crypto.fips_enabled = 1" definition_ref="sysctl_crypto_fips_enabled" />
+      <extend_definition comment="Dracut FIPS module is enabled" definition_ref="enable_dracut_fips_module" />
+      <extend_definition comment="system cryptography policy is configured" definition_ref="configure_crypto_policy" />
+      <extend_definition comment="Installed OS is certified" definition_ref="installed_OS_is_certified" />
+    </criteria>
+  </definition>
+</def-group>
--- a/linux_os/guide/system/software/integrity/fips/enable_fips_mode/oval/shared.xml
+++ b/linux_os/guide/system/software/integrity/fips/enable_fips_mode/oval/shared.xml
@@ -14,6 +14,17 @@
       <extend_definition comment="Dracut FIPS module is enabled" definition_ref="enable_dracut_fips_module" />
       <extend_definition comment="system cryptography policy is configured" definition_ref="configure_crypto_policy" />
       <extend_definition comment="Installed OS is certified" definition_ref="installed_OS_is_certified" />
+      <criterion comment="check if system crypto policy selection in var_system_crypto_policy in the profile is set to FIPS" test_ref="test_system_crypto_policy_value" />
     </criteria>
   </definition>
+  <ind:variable_test check="at least one" comment="tests if var_system_crypto_policy is set to FIPS" id="test_system_crypto_policy_value" version="1">
+    <ind:object object_ref="obj_system_crypto_policy_value" />
+    <ind:state state_ref="ste_system_crypto_policy_value" />
+  </ind:variable_test>
+  <ind:variable_object id="obj_system_crypto_policy_value" version="1">
+    <ind:var_ref>var_system_crypto_policy</ind:var_ref>
+  </ind:variable_object>
+  <ind:variable_state comment="variable value is set to 'FIPS'" id="ste_system_crypto_policy_value" version="1">
+    <ind:value operation="equals" datatype="string">FIPS</ind:value>
+  </ind:variable_state>
 </def-group>
--- a/linux_os/guide/system/software/integrity/fips/enable_fips_mode/oval/shared.xml
+++ b/linux_os/guide/system/software/integrity/fips/enable_fips_mode/oval/shared.xml
@@ -27,4 +27,5 @@
   <ind:variable_state comment="variable value is set to 'FIPS'" id="ste_system_crypto_policy_value" version="1">
     <ind:value operation="equals" datatype="string">FIPS</ind:value>
   </ind:variable_state>
+  <external_variable comment="defined crypto policy" datatype="string" id="var_system_crypto_policy" version="1" />
 </def-group>
--- a/linux_os/guide/system/software/integrity/fips/enable_fips_mode/rule.yml
+++ b/linux_os/guide/system/software/integrity/fips/enable_fips_mode/rule.yml
@@ -8,13 +8,6 @@ description: |-
     To enable FIPS mode, run the following command:
     <pre>fips-mode-setup --enable</pre>
     <br />
-    The <tt>fips-mode-setup</tt> command completes the installation of FIPS
-    modules by calling <tt>fips-finish-install</tt>. Then, it changes the
-    system crypto policy to FIPS by calling <tt>update-crypto-policies</tt>
-    tool. Also, the command modifies the boot loader configuration to add
-    <tt>fips=1</tt> and <tt>boot=&lt;boot-device&gt;</tt> options to the kernel
-    command line.
-    <br />
     On a system running in FIPS mode, the kernel FIPS mode flag
     (<tt>/proc/sys/crypto/fips_enabled</tt>) should be set to <tt>1</tt> and
     the <tt>/etc/system-fips</tt> should exist. The system crypto policy should
--- a/linux_os/guide/system/software/integrity/fips/enable_fips_mode/rule.yml
+++ b/linux_os/guide/system/software/integrity/fips/enable_fips_mode/rule.yml
@@ -1,0 +1,39 @@
+documentation_complete: true
+
+prodtype: rhel8,fedora
+
+title: Enable FIPS Mode
+
+description: |-
+    To enable FIPS mode, run the following command:
+    <pre>fips-mode-setup --enable</pre>
+
+rationale: |-
+    Use of weak or untested encryption algorithms undermines the purposes of utilizing encryption to
+    protect data. The operating system must implement cryptographic modules adhering to the higher
+    standards approved by the federal government since this provides assurance they have been tested
+    and validated.
+
+severity: high
+
+ocil_clause: 'FIPS mode is not enabled'
+
+ocil: |-
+    To verify that FIPS is enabled properly, run the following command:
+    <pre>fips-mode-setup --check</pre>
+    The output should contain the following:
+    <pre>FIPS mode is enabled.</pre>
+
+warnings:
+    - general: |-
+        The system needs to be rebooted for these changes to take effect.
+    - regulatory: |-
+        The ability to enable FIPS does not denote FIPS compliancy or certification.
+        Red Hat, Inc. and Red Hat Enterprise Linux are respectively FIPS certified and compliant. Community
+        projects such as CentOS, Scientific Linux, Fedora, etc. do not necessarily meet FIPS certification and compliancy.
+        Therefore, non-certified vendors and/or projects do not meet this requirement even if technically feasible.
+        <br /><br />
+        See <b>{{{ weblink(link="http://csrc.nist.gov/groups/STM/cmvp/documents/140-1/1401vend.htm") }}}</b>
+        for a list of FIPS certified vendors.
+
+platform: machine
--- a/linux_os/guide/system/software/integrity/fips/enable_fips_mode/rule.yml
+++ b/linux_os/guide/system/software/integrity/fips/enable_fips_mode/rule.yml
@@ -7,6 +7,20 @@ title: Enable FIPS Mode
 description: |-
     To enable FIPS mode, run the following command:
     <pre>fips-mode-setup --enable</pre>
+    <br />
+    The <tt>fips-mode-setup</tt> command completes the installation of FIPS
+    modules by calling <tt>fips-finish-install</tt>. Then, it changes the
+    system crypto policy to FIPS by calling <tt>update-crypto-policies</tt>
+    tool. Also, the command modifies the boot loader configuration to add
+    <tt>fips=1</tt> and <tt>boot=&lt;boot-device&gt;</tt> options to the kernel
+    command line.
+    <br />
+    On a system running in FIPS mode, the kernel FIPS mode flag
+    (<tt>/proc/sys/crypto/fips_enabled</tt>) should be set to <tt>1</tt> and
+    the <tt>/etc/system-fips</tt> should exist. The system crypto policy should
+    be set to FIPS in <tt>/etc/crypto-policies/config</tt>. Also, the Dracut
+    <tt>fips</tt> module should be loaded. Furthermore, the system running in
+    FIPS mode should be FIPS certified.
 
 rationale: |-
     Use of weak or untested encryption algorithms undermines the purposes of utilizing encryption to
--- a/linux_os/guide/system/software/integrity/fips/enable_fips_mode/rule.yml
+++ b/linux_os/guide/system/software/integrity/fips/enable_fips_mode/rule.yml
@@ -8,12 +8,15 @@ description: |-
     To enable FIPS mode, run the following command:
     <pre>fips-mode-setup --enable</pre>
     <br />
-    On a system running in FIPS mode, the kernel FIPS mode flag
-    (<tt>/proc/sys/crypto/fips_enabled</tt>) should be set to <tt>1</tt> and
-    the <tt>/etc/system-fips</tt> should exist. The system crypto policy should
-    be set to FIPS in <tt>/etc/crypto-policies/config</tt>. Also, the Dracut
-    <tt>fips</tt> module should be loaded. Furthermore, the system running in
-    FIPS mode should be FIPS certified.
+    The <tt>fips-mode-setup</tt> command will configure the system in
+    FIPS mode by automatically configuring the following:
+    <ul>
+    <li>Setting the kernel FIPS mode flag (<tt>/proc/sys/crypto/fips_enabled</tt>) to <tt>1</tt></li>
+    <li>Creating <tt>/etc/system-fips</tt></li>
+    <li>Setting the system crypto policy in <tt>/etc/crypto-policies/config</tt> to <tt>FIPS</tt></li>
+    <li>Loading the Dracut <tt>fips</tt> module</li>
+    </ul>
+    Furthermore, the system running in FIPS mode should be FIPS certified by NIST.
 
 rationale: |-
     Use of weak or untested encryption algorithms undermines the purposes of utilizing encryption to
--- a/linux_os/guide/system/software/integrity/fips/etc_system_fips_exists/rule.yml
+++ b/linux_os/guide/system/software/integrity/fips/etc_system_fips_exists/rule.yml
@@ -25,4 +25,16 @@ ocil: |-
     The output should be similar to the the following:
     <pre>-rw-r--r--. 1 root root 36 Nov 26 11:31 /etc/system-fips</pre>
 
+warnings:
+    - general: |-
+        The system needs to be rebooted for these changes to take effect.
+    - regulatory: |-
+        The ability to enable FIPS does not denote FIPS compliancy or certification.
+        Red Hat, Inc. and Red Hat Enterprise Linux are respectively FIPS certified and compliant. Community
+        projects such as CentOS, Scientific Linux, Fedora, etc. do not necessarily meet FIPS certification and compliancy.
+        Therefore, non-certified vendors and/or projects do not meet this requirement even if technically feasible.
+        <br /><br />
+        See <b>{{{ weblink(link="http://csrc.nist.gov/groups/STM/cmvp/documents/140-1/1401vend.htm") }}}</b>
+        for a list of FIPS certified vendors.
+
 platform: machine
--- a/linux_os/guide/system/software/integrity/fips/etc_system_fips_exists/rule.yml
+++ b/linux_os/guide/system/software/integrity/fips/etc_system_fips_exists/rule.yml
@@ -1,0 +1,28 @@
+documentation_complete: true
+
+prodtype: rhel8,fedora
+
+title: Ensure '/etc/system-fips' exists
+
+description: |-
+    On a system where FIPS mode is enabled, <tt>/etc/system-fips</tt> must exist.
+    To enable FIPS mode, run the following command:
+    <pre>fips-mode-setup --enable</pre>
+
+rationale: |-
+    Use of weak or untested encryption algorithms undermines the purposes of utilizing encryption to
+    protect data. The operating system must implement cryptographic modules adhering to the higher
+    standards approved by the federal government since this provides assurance they have been tested
+    and validated.
+
+severity: high
+
+ocil_clause: /etc/system-fips does not exist
+
+ocil: |-
+    To verify <tt>/etc/system-fips</tt> exists, run the following command:
+    <pre>ls -l /etc/system-fips</pre>
+    The output should be similar to the the following:
+    <pre>-rw-r--r--. 1 root root 36 Nov 26 11:31 /etc/system-fips</pre>
+
+platform: machine
--- a/linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/anaconda/shared.anaconda
+++ b/linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/anaconda/shared.anaconda
@@ -1,3 +1,3 @@
-# platform = Red Hat Enterprise Linux 7,Red Hat Enterprise Linux 8,multi_platform_fedora
+# platform = Red Hat Enterprise Linux 7
 
 package --add=dracut-fips
--- a/linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/bash/shared.sh
+++ b/linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/bash/shared.sh
@@ -1,4 +1,4 @@
-# platform = Red Hat Enterprise Linux 7,Red Hat Enterprise Linux 8,multi_platform_fedora
+# platform = Red Hat Enterprise Linux 7
 
 # include remediation functions library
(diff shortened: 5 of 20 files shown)


linux_os/guide/system/software/integrity/fips/package_dracut-fips_installed/rule.yml
```diff
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: rhel6,rhel7,rhel8,fedora
+prodtype: rhel6,rhel7
```
redhatrises: This rule is still applicable to rhel8. Please undo the changes.

linux_os/guide/system/software/integrity/fips/package_dracut-fips_installed/oval/shared.xml
```diff
@@ -8,8 +8,8 @@
     <metadata>
       <title>Package dracut-fips Installed</title>
       <affected family="unix">
-        <platform>multi_platform_rhel</platform>
-        <platform>multi_platform_fedora</platform>
+        <platform>Red Hat Enterprise Linux 6</platform>
+        <platform>Red Hat Enterprise Linux 7</platform>
```
redhatrises: Please undo the changes. OVAL is still applicable.

linux_os/guide/system/software/integrity/fips/package_dracut-fips_installed/bash/shared.sh
```diff
@@ -1,4 +1,4 @@
-# platform = multi_platform_rhel, multi_platform_fedora
+# platform = Red Hat Enterprise Linux 6, Red Hat Enterprise Linux 7
```
redhatrises: This remediation is still applicable to rhel8. Please undo the changes.

linux_os/guide/system/software/integrity/fips/package_dracut-fips_installed/ansible/shared.yml
```diff
@@ -1,4 +1,4 @@
-# platform = multi_platform_rhel, multi_platform_fedora
+# platform = Red Hat Enterprise Linux 6, Red Hat Enterprise Linux 7
```
redhatrises: Still applicable to rhel8. Please undo the changes.

linux_os/guide/system/software/integrity/fips/package_dracut-fips_installed/anaconda/shared.anaconda
```diff
@@ -1,3 +1,3 @@
-# platform = multi_platform_rhel, multi_platform_fedora
+# platform = Red Hat Enterprise Linux 6, Red Hat Enterprise Linux 7
```
redhatrises: Still applicable to rhel8. Please undo the changes.

linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/rule.yml
```diff
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: rhel7,rhel8,fedora
+prodtype: rhel7
```
redhatrises: Still applicable to rhel8. Please undo the changes.

linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/oval/shared.xml
```diff
@@ -4,8 +4,6 @@
       <title>Enable FIPS Mode in GRUB2</title>
       <affected family="unix">
         <platform>Red Hat Enterprise Linux 7</platform>
```
redhatrises: Still applicable to rhel8. Please undo the changes.

linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/bash/shared.sh
```diff
@@ -1,4 +1,4 @@
-# platform = Red Hat Enterprise Linux 7,Red Hat Enterprise Linux 8,Red Hat Enterprise Linux 8,multi_platform_fedora
+# platform = Red Hat Enterprise Linux 7
```
redhatrises: Still applicable to rhel8. Please undo the changes.

linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/anaconda/shared.anaconda
```diff
@@ -1,3 +1,3 @@
-# platform = Red Hat Enterprise Linux 7,Red Hat Enterprise Linux 8,multi_platform_fedora
+# platform = Red Hat Enterprise Linux 7
```
redhatrises: Still applicable to rhel8. Please undo the changes.

fedora/templates/csv/sysctl_values.csv
```diff
@@ -6,3 +6,4 @@ kernel.kptr_restrict,1
 kernel.randomize_va_space,2
 kernel.yama.ptrace_scope,1
 net.ipv6.conf.all.disable_ipv6,1
+crypto.fips_enabled,1
```
redhatrises: FIPS checks should never be templated as FIPS enablement should be heavily scrutinized.

linux_os/guide/system/software/integrity/fips/enable_fips_mode/oval/shared.xml
```diff
@@ -0,0 +1,25 @@
...
+    <metadata>
+      <title>Enable FIPS Mode</title>
+      <affected family="unix">
+        <platform>Red Hat Enterprise Linux 8</platform>
+        <platform>multi_platform_fedora</platform>
+      </affected>
+      <description>Check existence of /etc/system-fips and verify if FIPS is enabled in Linux kernel.</description>
+    </metadata>
+    <criteria operator="AND">
+      <criterion test_ref="test_etc_system_fips" comment="check existence of /etc/system-fips" />
```
redhatrises: I expect this to be an `<extend_definition />` as this is also applicable to RHEL7

linux_os/guide/system/software/integrity/fips/enable_fips_mode/oval/shared.xml
```diff
@@ -0,0 +1,25 @@
...
+      <extend_definition definition_ref="sysctl_runtime_crypto_fips_enabled" comment="check sysctl crypto.fips_enabled = 1" />
+      <extend_definition comment="Installed OS is certified" definition_ref="installed_OS_is_certified" />
+    </criteria>
+  </definition>
+
+  <unix:file_test version="1" id="test_etc_system_fips" check="all" check_existence="all_exist" comment="/etc/system-fips exists">
```
redhatrises: Please move this to a separate OVAL file.

linux_os/guide/system/software/integrity/fips/package_dracut-fips_installed/rule.yml
jan-cerny: @redhatrises Why do you think this rule is applicable to RHEL 8? Package `dracut-fips` doesn't exist in RHEL 8.

rhel8/templates/csv/sysctl_values.csv
```diff
@@ -6,3 +6,4 @@ kernel.kptr_restrict,1
 kernel.randomize_va_space,2
 kernel.yama.ptrace_scope,1
 net.ipv6.conf.all.disable_ipv6,1
+crypto.fips_enabled,1
```
redhatrises: FIPS checks should never be templated as FIPS enablement should be heavily scrutinized.

linux_os/guide/system/software/integrity/fips/enable_fips_mode/oval/shared.xml
```diff
@@ -0,0 +1,25 @@
...
+      <unix:object object_ref="object_etc_system_fips" />
+  </unix:file_test>
+
+  <unix:file_object version="1" id="object_etc_system_fips">
+      <unix:filepath>/etc/system-fips</unix:filepath>
+  </unix:file_object>
+</def-group>
```
redhatrises: I expect to see all FIPS checks in this OVAL file. So, it needs to also check that the fips package is installed and that fips is enabled in grub as well.

linux_os/guide/system/software/integrity/fips/package_dracut-fips_installed/rule.yml
redhatrises: @jan-cerny ooppss... you are right... I checked the wrong RHEL8 build. Nevermind on this rule.

linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/anaconda/shared.anaconda
jan-cerny: @redhatrises  I took a look into this again and this Anaconda remediation only adds "dracut-fips" package which again doesn't exist in RHEL8.

linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/bash/shared.sh
jan-cerny: @redhatrises I don't know if we should check `/etc/default/grub` on RHEL8 because the utility `fips-mode-setup` which should be used for enabling FIPS mode on RHEL8 doesn't modify that `/etc/default/grub`. On the other hand, we could say that running `fips-mode-setup`  is not enough and users are required to modify `/etc/default/grub` as well.

Does anybody have any idea if modifying  `/etc/default/grub` should still be required in order to enable FIPS mode in RHEL8?

jan-cerny:
@redhatrises I looked into the FIPS mode on RHEL8 again today. I also discussed this with @t8m who designed the FIPS mode enablement on RHEL8.

To summarize everything that I already know:
* FIPS mode on RHEL8 is pretty different from RHEL7.
* the way to enable FIPS mode on RHEL8 is using `fips-mode-setup --enable`
* If FIPS mode is enabled, `/etc/system-fips` must exist 
* If FIPS mode is enabled, `crypto.fips_enabled` in `sysctl` should be `1`. It is not important how this is achieved. I think that means we won't check `/etc/default/grub` in RHEL8, as `fips-mode-setup` doesn't modify this file.
* `dracut-fips` doesn't exist in RHEL8, therefore the rule checking if `dracut-fips` is installed is not applicable on RHEL8 anymore.
* To be more sure, we can check  if `/etc/dracut.conf.d/40-fips.conf` contains `add_dracutmodules+=" fips "`. This line is generated by `fips-mode-setup` when you run `fips-mode-setup --enable`.
* also we can check the system-wide crypto policy setting. In FIPS mode on RHEL8, `/etc/crypto-policies/config` should contain `FIPS`.
* relevant docs are man pages for fips-mode-setup and fips-finish-install

Based on these information I plan to:
* Add a check that checks FIPS in crypto-policies configuration.
* Add a check that checks if `/etc/dracut.conf.d/40-fips.conf` contains `add_dracutmodules+=" fips "`.
* Rebase this PR on top of master to resolve merge-conflicts that have appeared in meanwhile.

linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/bash/shared.sh
t8m: Running fips-mode-setup is enough.

linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/oval/shared.xml
t8m: No, it is not.

linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/bash/shared.sh
redhatrises: @jan-cerny @t8m what happens on a kernel upgrade? Also what happens if someone sets `fips=0` in `/etc/default/grub`

jan-cerny:
I have updated the commit.

jan-cerny:
Rebased on master.

linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/bash/shared.sh
t8m: Kernel upgrade should keep the kernel options as edited by grubby. Unfortunately there is a regression in this with the new grubby wrapper for BLS which needs to be fixed. I am not sure what happens if users set conflicting options via grubby and via /etc/default/grub however please note that for example there is no grub on s390x so I do not think it makes much sense to look at /etc/default/grub as it is not a canonical source of the information.

linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/anaconda/shared.anaconda
yuumasato: I consider this solved, there is not package `dracut-fips` in RHEL8.

linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/oval/shared.xml
yuumasato: Process for enabling FIPS mode in RHEL8 and Fedora have changed.

linux_os/guide/system/software/integrity/fips/enable_fips_mode/oval/shared.xml
```diff
@@ -0,0 +1,19 @@
...
+        <platform>Red Hat Enterprise Linux 8</platform>
+        <platform>multi_platform_fedora</platform>
+      </affected>
+      <description>Check if FIPS mode is enabled on the system</description>
+    </metadata>
+    <criteria operator="AND">
+      <extend_definition comment="check /etc/system-fips exists" definition_ref="etc_system_fips_exists" />
+      <extend_definition comment="check sysctl crypto.fips_enabled = 1" definition_ref="sysctl_crypto_fips_enabled" />
+      <extend_definition comment="Dracut FIPS module is enabled" definition_ref="enable_dracut_fips_module" />
+      <extend_definition comment="system cryptography policy is configured" definition_ref="configure_crypto_policy" />
```
yuumasato: I was wondering if `enable_fips_mode`  check should really look into the selected crypto-policy, or we should select the `configure_crypto_policy` rule in the profile instead.
But I did not know that the remediation command would also be setting FIPS crypto-policy.

I think it should be more clear in rule description.

linux_os/guide/system/software/integrity/fips/enable_fips_mode/rule.yml
```diff
@@ -0,0 +1,39 @@
+documentation_complete: true
+
+prodtype: rhel8,fedora
+
+title: Enable FIPS Mode
+
+description: |-
+    To enable FIPS mode, run the following command:
+    <pre>fips-mode-setup --enable</pre>
```
yuumasato: I think here it should also be described the state characterized as FIPS enabled. Basically what the OVAL check is veriffyin, that is crypto-policy set to FIPS, kernel parameter, dracut module enabled and `/etc/system-fips` file.
At first look, it was not clear that `fips-mode-setup` would do all these changes.

linux_os/guide/system/software/integrity/fips/enable_fips_mode/oval/shared.xml
jan-cerny: > I was wondering if enable_fips_mode check should really look into the selected crypto-policy, or we should select the configure_crypto_policy rule in the profile instead.

I think we are going to add rule configure_crypto_policy anyway because we are going to add rules about crypto policies for RHEL8 profiles.

I thought we should not duplicate the OVAL from rule configure_crypto_policy, so I preferred to add it as an extend_definition.

My concern was: Somebody will have a tailoring in which he removes the rule configure_crypto_policy from the profile, then he messes up the crypto policy settings, then he scans his machine, the rule enable_fips_mode will pass, but he will be in fact in inconsistent state.

> But I did not know that the remediation command would also be setting FIPS crypto-policy.

I look into that and I found the `fips-mode-setup` tool calls `update-crypto-policies --set FIPS`.

I agree it should be more clear in the rule description.

linux_os/guide/system/software/integrity/fips/enable_fips_mode/rule.yml
jan-cerny: Yes, I agree.

linux_os/guide/system/software/integrity/fips/enable_fips_mode/oval/shared.xml
yuumasato: > I look into that and I found the `fips-mode-setup` tool calls `update-crypto-policies --set FIPS`.

@jan-cerny Exactly! It sets the crypto-policies to `FIPS`.

Now I believe that OVAL definition for `enable_fips_mode` should actually look for `FIPS` policy, not what is defined in `var_system_crypto_policy`

jan-cerny: @yuumasato Shouldn't it be that `var_system_crypto_policy` should be set to `fips`?

t8m: The check should be that for system FIPS mode the crypto policy must be set to FIPS.

jan-cerny: @t8m Yes, I agree, that's clear that the crypto policy must be set to FIPS. The discussion is now rather about the way we implement the check. We already have a rule that checks the crypto policy setting. In my patch I just reuse this rule to check if crypto policy is set to FIPS. @yuumasato is proposing a different way to achieve the same thing, but his way is not clear to me.

yuumasato: @jan-cerny the purpose of `var_system_crypto_policy` is for you to be able to customize what crypto policy to check for. But in this case want it fixed to `FIPS`, the rule for `enable_fips_mode` should always check for `FIPS` mode, and not what is set in the variable.

yuumasato: @jan-cerny Not sure how to implement this though.

One option is to declare `var_system_crypto_policy` within definition `enable_fips_mode`. Not sure if `var_system_crypto_policy` would pick this value, though. If this approach works it could potentially override configuration of `var_system_crypto_policy` in the Profile, which can be confusing to users.

Another option, simpler, is to add a criteria that checks if `var_system_crypto_policy` is set to `FIPS`.

linux_os/guide/system/software/integrity/fips/enable_fips_mode/rule.yml
yuumasato: Thanks for updating the description. 
The paragraph below, right after the command seems too technical and not necessary in rule description. 
The description of FIPS state in last paragraph should be enough.

redhatrises: Agree. The paragraph's below are too much.

linux_os/guide/system/software/integrity/fips/enable_fips_mode/oval/shared.xml
redhatrises: @jan-cerny we should also add a check to verify that fips is not disabled in `/etc/default/grub`

linux_os/guide/system/software/integrity/fips/enable_fips_mode/rule.yml
jan-cerny: @yuumasato OK

linux_os/guide/system/software/integrity/fips/enable_fips_mode/oval/shared.xml
jan-cerny: @yuumasato Thaks for your suggestions. I'll try if they work.

jan-cerny:
It looks like an after-merge conflict. I think it will be better if I rebase.

linux_os/guide/system/software/integrity/fips/enable_fips_mode/rule.yml
```diff
@@ -0,0 +1,46 @@
...
+    <br />
+    On a system running in FIPS mode, the kernel FIPS mode flag
+    (<tt>/proc/sys/crypto/fips_enabled</tt>) should be set to <tt>1</tt> and
+    the <tt>/etc/system-fips</tt> should exist. The system crypto policy should
+    be set to FIPS in <tt>/etc/crypto-policies/config</tt>. Also, the Dracut
+    <tt>fips</tt> module should be loaded. Furthermore, the system running in
+    FIPS mode should be FIPS certified.
```
redhatrises: This paragraph should be removed because it reads as if more configurations need to take place rather than just running `fips-mode-setup --enable`

linux_os/guide/system/software/integrity/fips/etc_system_fips_exists/rule.yml
```diff
@@ -0,0 +1,28 @@
...
+severity: high
+
+ocil_clause: /etc/system-fips does not exist
+
+ocil: |-
+    To verify <tt>/etc/system-fips</tt> exists, run the following command:
+    <pre>ls -l /etc/system-fips</pre>
+    The output should be similar to the the following:
+    <pre>-rw-r--r--. 1 root root 36 Nov 26 11:31 /etc/system-fips</pre>
+
```
redhatrises: Please add the FIPS warnings to this rule.

linux_os/guide/system/software/integrity/fips/sysctl_crypto_fips_enabled/rule.yml
```diff
@@ -0,0 +1,30 @@
...
+severity: high
+
+ocil_clause: 'crypto.fips_enabled is not 1'
+
+ocil: |-
+    To verify that kernel parameter 'crypto.fips_enabled' is set properly, run the following command:
+    <pre>sysctl crypto.fips_enabled</pre>
+    The output should contain the following:
+    <pre>crypto.fips_enabled =  1</pre>
+
```
redhatrises: Please add the FIPS warnings to this rule.

linux_os/guide/system/software/integrity/fips/enable_fips_mode/rule.yml
yuumasato: it would be nice to have described what the command will do, like what settings are necessary to be in FIPS mode.

redhatrises: Most people don't care about what the command will do. Just that it does as advertised. We don't describe all the steps of running an rpm installation or update or the grub2-mkconfig command. 

Still, the way it is written isn't in the form of describing what the command will do but rather additional tasks.

yuumasato: > We don't describe all the steps of running an rpm installation or update or the grub2-mkconfig command.

Well, there is no need for a pseudo-algorithm of the command.

@redhatrises My rational for advocating for description of the expected FIPS state, is that it can help when the rule fails. Sometimes it can be hard to know what may be going wrong.

In this case, the remediation is simple as running a command, there is not much room for mistakes here. But the OVAL definition checks a lot of stuff, and with these described it is easier to check and debug, or for non-experts in OVAL to understand what the rule expects.

There are cases where check and remediation are complex, like in audit rules, for example. When rule fails scan, with description in rule, people can go and look up into their audit rules and compare what is expected to be the there, and what is actually there.

jan-cerny:
I have copied the warnings to the other rules.

linux_os/guide/system/software/integrity/fips/enable_fips_mode/rule.yml
redhatrises: > > We don't describe all the steps of running an rpm installation or update or the grub2-mkconfig command.
> 
> Well, there is no need for a pseudo-algorithm of the command.
> 
> @redhatrises My rational for advocating for description of the expected FIPS state, is that it can help when the rule fails. Sometimes it can be hard to know what may be going wrong.

That's what `--oval-results` is for.

> In this case, the remediation is simple as running a command, there is not much room for mistakes here. But the OVAL definition checks a lot of stuff, and with these described it is easier to check and debug, or for non-experts in OVAL to understand what the rule expects.
> 
> There are cases where check and remediation are complex, like in audit rules, for example. When rule fails scan, with description in rule, people can go and look up into their audit rules and compare what is expected to be the there, and what is actually there.

Anyway the wording needs to be rewritten in such a way so that it doesn't read as there are addition steps to follow.
For example:

```
    The <tt>fips-mode-setup</tt> command will configure the system in 
    FIPS mode by automatically configuring the following:
    <ul>
    <li>Setting <tt>/proc/sys/crypto/fips_enabled</tt> to <tt>1</tt></li>
    <li>Creating <tt>/etc/system-fips</tt></li>
    <li>Enabling FIPS in the system crypto policy in <tt>/etc/crypto-policies/config</tt></li>
    <li>Loading the Dracut <tt>fips</tt> module</li> 
    </ul>
    Furthermore, the system running in FIPS mode should be FIPS
    certified by NIST.
```

linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/bash/shared.sh
redhatrises: While that may be true, we will need to validate that fips hasn't been disabled elsewhere. So using /etc/default/grub as an example, if an administrator can set `fips=0` in /etc/default/grub which overrides grubby, that's a compliance issue that we need to flag with an OVAL check.

linux_os/guide/system/software/integrity/fips/enable_fips_mode/oval/shared.xml
jan-cerny: @yuumasato I forgot to update you that they work.

redhatrises:
Ack. Thanks for reviews and comments everyone.

linux_os/guide/system/software/integrity/fips/enable_fips_mode/rule.yml
t8m: `Furthermore, the system running in FIPS mode should be FIPS certified by NIST.`
This is a misleading sentence. I'd suggest removing it or replacing with something like:
`Furthermore for the system to be FIPS compliant the components implementing cryptography have to be FIPS certified by NIST.`

PR #3625, opened 2018-11-26
Rules for finit and create kernel modules should not aply to RHEL6.

yuumasato:
#### Description:

- Rule `audit_rules_kernel_module_loading_finit` should not apply to RHEL6.
- Rule `audit_rules_kernel_module_loading_create` should not apply to RHEL6.

#### Rationale:

- The rule for RHEL6 should be `audit_rules_kernel_module_loading`


redhatrises:
ACK

PR #3627, opened 2018-11-27
Introduce platform dependent test scenarios

jan-cerny:
We face a problem that for a single rule we need to have different sets
of test scenarios depending on target platform. For example, the rule
'auditd_audispd_encrypt_sent_records' exists on both RHEL 7 and RHEL 8,
but on each of them a different file path and a different string
is checked. We need to be able to define test scenarios that are
applicable only to selected platforms and should be skipped when testing
other platforms.

This will be achieved by using platform option in scenario script headers.
It works similar way we specify platforms in Bash remediations. We can
use `multi_platform_rhel`, `multi_platform_fedora`, etc.


tests/ssg_test_suite/common.py
```diff
@@ -18,6 +18,38 @@
...
+        "cpe:/o:centos:centos:7"
+    ],
+    "Red Hat Enterprise Linux 8": [
+        "cpe:/o:redhat:enterprise_linux:8",
+        "cpe:/o:centos:centos:8"
+    ],
+    "Fedora": [
+        "cpe:/o:fedoraproject:fedora:28",
+        "cpe:/o:fedoraproject:fedora:29",
+        "cpe:/o:fedoraproject:fedora:30"
```
yuumasato: Would it be possible to source these cpes from a place where we already track it?
For example the `product.yml`?

tests/ssg_test_suite/common.py
```diff
@@ -18,6 +18,38 @@
...
+    ],
+}
+
+# Keys are strings used in test scenario header in 'platform' list
+TEST_MULTI_PLATFORMS = {
```
yuumasato: In `ssg/products.py` we already maintain a mapping of platforms to names.
I'd rather avoid maintaining a separate mapping, and extend that functionality to map from platforms to cpes.

tests/ssg_test_suite/rule.py
```diff
@@ -150,7 +152,24 @@ def _matches_target(rule_dir, targets):
...
+    scenario_cpes = set()
+    for platform in scenario_platforms:
+        try:
+            if platform.startswith("multi_platform_"):
+                platform_cpes = set(common.TEST_MULTI_PLATFORMS[platform])
+            else:
+                platform_cpes = set(common.TEST_PLATFORMS[platform])
+        except KeyError:
+            raise ValueError(
+                "Platform '%s' is not supported by the test suite." % platform)
```
yuumasato: What is the reason a platform is not supported for testing?
Is it because there wouldn't be a mapping from the product name to its cpe?

tests/ssg_test_suite/common.py
yuumasato: This mapping of product name to cpe could be outsourced to `ssg/products.py`

tests/README.md
```diff
@@ -141,6 +141,28 @@ expects success.
 Support files, that are available to the scenarios during their runtime. Can
 be used as common libraries.
 
+### Scenarios format
+
+Scenarios are simple bash scripts. The scenarios start with a header which provides metadata.
+The header consists of comments (starting by `#`). Possible values are:
+- `platform` is a comma-separated list of platforms where the test scenario can be run. This is similar to `platform` used in our remediations. Examples of values: `multi_platform_rhel`, `Red Hat Enterprise Linux 7`, `multi_platform_all`. If `platform` is not specified in the header, `multi_platform_all` is assumed.
+- `profiles` is a comma-separated list of profiles which are using this scenario.
+- `remediation` is a comma-separated list of allowed remediation types (eg. `bash`, `ansible`).
```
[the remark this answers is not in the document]
jan-cerny: Great catch! I'll update that.

tests/ssg_test_suite/rule.py
jan-cerny: Yes, it's exactly that. I agree it's kind of stupid. It is related to previous comment. Basically the problem is that it isn't easy to get such a mapping now and we'll need to change the way we define product names and CPEs to prevent maintaining a separate list of mappings for the test suite.

tests/ssg_test_suite/common.py
jan-cerny: @yuumasato  Yes, I like the idea to have it defined at single place.

There are multiple problems with that:
* The CPEs for each product are defined in the product-specific `product.yml`. We probably shouldn't parse these files in the test suite. It would be better to change the content build system to source it from a single place as well.
* The CPEs for derivatives (CentOS, Scientific Linux) are not defined in `product.yml` but `ssg/constants.py`.
* The code in `ssg/products.py` maps shortened names of products (eg. `rhel7`) to official names, eg. when you build `rhel7` product you need to know that you  should also include remediations marked as one of  `['multi_platform_rhel','multi_platform_all', 'Red Hat Enterprise Linux 7']`.  But we need kind of opposite mapping - eg. when we specify `multi_platform_rhel` we would like to expand it to `['rhel6','rhel7']` and then convert it a list of CPEs.

I would like to have all the mappings defined at a single point. But I'm not sure if it's not a scope creep. Should it be done in a separate PR?

matejak:
@jan-cerny If you have surplus, you may consider testing some of the newly introduced functionality in the SSG unit-testing framework. In that case, check out the `content/tests/unit/ssg_test_suite` folder.
As the implementation of platform identification may change in the future so it is in one place both for the build system and the test suite, having some easy tests for the `_matches_platform` function looks like as a good stretch goal.

jan-cerny:
@matejak I have added a basic unit test for that.

tests/ssg_test_suite/common.py
yuumasato: > * The CPEs for each product are defined in the product-specific `product.yml`. We probably shouldn't parse these files in the test suite. It would be better to change the content build system to source it from a single place as well.

I'm not sure it would be better, for someone writing content, or adding a new product, it wouldn't be straight forward to go into `ssg/constants.py` and add the mappings there.

I'm okay to go take this path for now, but in the end, I think all data specific for the product should be in its `product.yml`, for example, the product tag name, full name, and cpes.
And the build system would know what products are part of which `multi_platform_` label, and it gathers the data and builds the mappings.

This mapping of product could even be extended to something else, something not tied to the product name, but feature based.  This could be a "label" defined in the `product.yml`, that would aggregate rules, to products that provide the feature.
Some labels I can imagine:`crypto_policy`, `fips`, `grub_legacy`, `grub2`.

jan-cerny:
Unfortunately, I have discovered that in build_remediations.py we have a different logic that reuses the fact that build_remediations.py knows which product directory (eg. `rhel6/`, `rhel7/`) it works with. The test suite doesn't (couldn't) have this information, so we need an explicit mapping somewhere.

I have noticed that we could reuse my new code in the test suite and ssg.constants to refactor  build_remediations.py and similar scripts. But I vote against doing this refactoring in this PR because that would be a scope creep.

tests/ssg_test_suite/common.py
```diff
@@ -132,3 +134,39 @@ def _run_cmd(command_list, verbose_path, env=None):
...
+        return platform_cpes
+
+
+def matches_platform(scenario_platforms, benchmark_cpes):
+    if "multi_platform_all" in scenario_platforms:
+        return True
+    scenario_cpes = set()
+    for p in scenario_platforms:
+        scenario_cpes |= _get_platform_cpes(p)
+    print(scenario_cpes)
```
jan-cerny: Please remove this!

tests/ssg_test_suite/common.py
yuumasato: @jan-cerny yes. In the future we could.

tests/unit/ssg_test_suite/test_matches_platform.py
```diff
@@ -0,0 +1,146 @@
...
+
+def test_typo():
+    scenario_platforms = ["Rrd Hat Enterprise Linux 7"]
+    benchmark_cpes = {"cpe:/o:redhat:enterprise_linux:7"}
+    with pytest.raises(ValueError):
+        common.matches_platform(scenario_platforms, benchmark_cpes)
+
+
+def test_wrong_multi_platform():
+    scenario_platforms = ["multi_platform_fidorka"]
```
yuumasato: Fidorka! :)

PR #3632, opened 2018-11-30
Drop deprecated or removed packages in RHEL8

yuumasato:
#### Description:

- Drop deprecated or removed packages from RHEL8 CSV's.

#### Rationale:

- According to [Changes to packages](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8-beta/html/8.0_beta_release_notes/changes_to_packages) page in Beta  release notes, these packages won't be part of RHEL8.


rhel8/templates/csv/packages_removed.csv
```diff
@@ -15,24 +15,19 @@ httpd
 iputils
 kernel-tools
 kexec-tools
-libcgroup
-libcgroup-tools
```
redhatrises: @yuumasato why were these dropped? Packages still exist.

redhatrises:
@yuumasato let's just remove these from the profiles for now until we are sure that the rpms don't end up in epel repo or some other repo.

yuumasato:
> @yuumasato let's just remove these from the profiles for now until we are sure that the rpms don't end up in epel repo or some other repo.

I was wondering about EPEL repos, if they are available there we should still track these rpms.

rhel8/templates/csv/packages_removed.csv
jan-cerny: They're mentioned that they are removed in the docs https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8-beta/html/8.0_beta_release_notes/changes_to_packages but they exist on my RHEL8 beta, so that means the docs are wrong.

yuumasato: Since these packages are still shipped, I'll put them back into the list.

yuumasato:
> @yuumasato let's just remove these from the profiles for now

@redhatrises Maybe I don't understand what you mean, you suggest to just remove rules from the profile but leave them on CSV files?

rhel8/templates/csv/packages_installed.csv
```diff
@@ -9,6 +9,7 @@ libreswan
 ntp
```
yuumasato: I have kept `ntp` here because rule `service_chronyd_or_ntpd_enabled` relies on check for `package_ntp_installed`.
`chronyd_or_ntpd` rules need a `chronyd` only rule version befor ntp can be removed from this CSV.

jan-cerny: That could be achieved by JInja macros in the rule, OVALs and remediations.

yuumasato: I could filter out checks for `ntp` and make it work, but the rule would still be called `chronyd_or_ntpd`, which would be confusing...

jan-cerny: We can rename the rule to `clock_synchronization_enabled`. But I agree with keeping `ntp` here, it's easier.

redhatrises: There is already a ticket to break out these rules which should be separate rules and not rules joined by jinja.

redhatrises:
> > @yuumasato let's just remove these from the profiles for now
> 
> @redhatrises Maybe I don't understand what you mean, you suggest to just remove rules from the profile but leave them on CSV files?

Yes to remove them from the profile, but keep them enabled. Of course, we can hold off on this PR for now.

jan-cerny:
Hi @yuumasato @redhatrises, what is the status of this?

yuumasato:
Closing in favor of #3742

PR #3635, opened 2018-11-30
Removed a stray print of a profile id.

matejak:
The print appears when building content for derived products.


redhatrises:
Mmmm.... thought I had removed that.... thanks @matejak

PR #3637, opened 2018-12-01
document required build packages

shawndwells:
Document what packages are required to build ComplianceAsCode


jan-cerny:
Build dependencies are already covered by https://github.com/ComplianceAsCode/content/blob/master/BUILD.md . It doesn't mention Python 3 packages, though. What about merging the 2 documents? It would be better to have all this developer information at a single place.

shawndwells:
@jan-cerny sure - feel free to merge the two. Documentation is extremely poor for ComplianceAsCode -- content is scattered across multiple locations with no clear indication of where to find things.

PR #3638, opened 2018-12-03
initial rhel8 ospp template

shawndwells:
Initial RHEL8 OSPP profile. Does not add new rules, rather adds verbose comments to show where rules should/should not go. Meant to orientate the conversation of what goes into OSPP.

Fully expected sections will add/remove/change as dialog continues.

### DIFF
--- a/rhel8/profiles/ospp.profile
+++ b/rhel8/profiles/ospp.profile
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-title: 'OSPP - Protection Profile for General Purpose Operating Systems'
+title: 'Protection Profile for General Purpose Operating Systems'
 
 description: |-
     This profile reflects mandatory configuration controls identified in the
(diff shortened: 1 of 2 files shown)


rhel8/profiles/ospp.profile
```diff
@@ -106,94 +304,532 @@ selections:
...
+
+
+
+
+    #######################################################
+    ## Rules that need a home?
+
+
+#    - sysctl_kernel_yama_ptrace_scope
+#    - sysctl_kernel_kptr_restrict
```
jan-cerny: Will we keep the rules that don't have a home in the profile? If there isn't appropriate category for them in OSPP, should they be removed?

shawndwells: Don't know. For now, commented out rules that I didn't know what to do with.... IIRC there are some "catch all" categories we can map rules into at our discretion.

shawndwells:
rebased & ready for merge

redhatrises:
lgtm.

PR #3640, opened 2018-12-03
Improve audit unset and fix uid values

yuumasato:
#### Description:

- Use keyword `unset` instead of value `4294967295` for unknown user ids in audit rules.
- Use `auid` value defined by build system instead of hardcoded `1000`.
  - Default `auid` value defined by build system is `1000`, unless `uid_min` is defined in `product.yml.

#### Rationale:

- `unset` is much more readable.
- RHEL6 minimal `UID` starts at [500](https://github.com/ComplianceAsCode/content/blob/bd38c1427b7922800a606946683dfa5acde4881e/rhel6/product.yml#L13).

### DIFF
--- a/shared/templates/template_ANSIBLE_audit_rules_dac_modification
+++ b/shared/templates/template_ANSIBLE_audit_rules_dac_modification
@@ -37,7 +37,7 @@
 - name: Inserts/replaces the {{{ ATTR }}} rule in rules.d when on x86
   lineinfile:
     path: "{{ all_files[0] }}"
-    line: "-a always,exit -F arch=b32 -S {{{ ATTR }}} -F auid>={{{ auid }}} -F auid!=4294967295 -F key=perm_mod"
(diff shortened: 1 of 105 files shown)


shared/templates/template_ANSIBLE_audit_rules_dac_modification
```diff
@@ -37,7 +37,7 @@
 - name: Inserts/replaces the {{{ ATTR }}} rule in rules.d when on x86
   lineinfile:
     path: "{{ all_files[0] }}"
-    line: "-a always,exit -F arch=b32 -S {{{ ATTR }}} -F auid>=1000 -F auid!=4294967295 -F key=perm_mod"
```
jan-cerny: Can the `auid` in Ansible tasks be changed to `unset` as well?

yuumasato: Nice catch, fixed!
Also fixed some other places I forgot about.

jan-cerny:
Great!

PR #3642, opened 2018-12-03
Improve parameters for ensure_fedora_gpgkey_installed

yuumasato:
#### Description:

- The Rule, OVAL and Bash fix will source the parameters for the gpg-pubkey from `fedora/product.yml`.
- Add back the checks for the two keys of the supported releases.

#### Rationale:

- Parametrized`gpg-pubkey` values in OVAL check will make it easy to move forward when next releases come.

### DIFF
--- a/fedora/product.yml
+++ b/fedora/product.yml
@@ -12,11 +12,11 @@ init_system: "systemd"
 
 # The fingerprints below are retrieved from https://getfedora.org/keys/
 latest_version: 29
-latest_release_fingerprint: "5A03 B4DD 8254 ECA0 2FDA  1637 A20A A56B 4294 76B4"
+latest_release_fingerprint: "5A03B4DD8254ECA02FDA1637A20AA56B429476B4"
 latest_pkg_release: 5a886537
 latest_pkg_version: 429476b4
 
 previous_version: 28
-previous_release_fingerprint: "128C F232 A937 1991 C8A6  5695 E08E 7E62 9DB6 2FB1"
+previous_release_fingerprint: "128CF232A9371991C8A65695E08E7E629DB62FB1"
 previous_pkg_release: 59920156
 previous_pkg_version: 9db62fb1
--- a/fedora/product.yml
+++ b/fedora/product.yml
@@ -13,10 +13,10 @@ init_system: "systemd"
 # The fingerprints below are retrieved from https://getfedora.org/keys/
 latest_version: 29
 latest_release_fingerprint: "5A03B4DD8254ECA02FDA1637A20AA56B429476B4"
-latest_pkg_release: 5a886537
-latest_pkg_version: 429476b4
+latest_pkg_release: "5a886537"
+latest_pkg_version: "429476b4"
 
 previous_version: 28
 previous_release_fingerprint: "128CF232A9371991C8A65695E08E7E629DB62FB1"
-previous_pkg_release: 59920156
-previous_pkg_version: 9db62fb1
+previous_pkg_release: "59920156"
+previous_pkg_version: "9db62fb1"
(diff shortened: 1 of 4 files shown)


jan-cerny:
Would changes done in https://github.com/ComplianceAsCode/content/pull/3622  be needed on Fedora as well?

yuumasato:
> Would changes done in #3622 be needed on Fedora as well?

I can check on Fedora later..

yuumasato:
@jan-cerny I have changed the fix to use `--with-colon`, as it is more suitable for scripted calls to `gpg`.
The bug with `GPG_OUT` doesn't manifest in Fedora because it only checks for one key.

linux_os/guide/system/software/updating/ensure_fedora_gpgkey_installed/bash/shared.sh
```diff
@@ -25,14 +25,14 @@ if [ "${RPM_GPG_DIR_PERMS}" -le "755" ]
 then
   # If they are safe, try to obtain fingerprints from the key file
   # (to ensure there won't be e.g. CRC error).
-  IFS=$'\n' GPG_OUT=($(gpg --with-fingerprint "${REDHAT_RELEASE_KEY}" | grep 'Key fingerprint ='))
+  IFS=$'\n' GPG_OUT=($(gpg --with-fingerprint --with-colons "${REDHAT_RELEASE_KEY}" | grep '"^fpr' | cut -d ":" -f 10))
```
matejak: Unfortunately, this is "abuse" of `IFS` - the correct way is to save `IFS` and then restore it. Setting `IFS` on the same line may create a false impression that it applies to that one line only, but in reality, it is overwritten from that place on (see [the bash video](https://redhat.bluejeans.com/playback/guid/NDg0NjkwNzgyMDozMTczMTctNjk4MDQ5OTYtNmM5Yi00ZTBmLWFhNDctNWJlZGJlMzJkZjFj) around 17. minute). Unsetting IFS may restore it to the original value, but it is [reported](https://stackoverflow.com/questions/39545837/unset-ifs-unexpected-behaviour) to be unsafe. So I suggest to have there an auxiliary `SAVE_IFS` variable.

fedora/product.yml
```diff
@@ -11,7 +11,12 @@ pkg_manager: "dnf"
 init_system: "systemd"
 
 # The fingerprints below are retrieved from https://getfedora.org/keys/
-latest_version: 28
-latest_release_fingerprint: "128C F232 A937 1991 C8A6  5695 E08E 7E62 9DB6 2FB1"
-previous_version: 27
-previous_release_fingerprint: "860E 19B0 AFA8 00A1 7518  81A6 F55E 7430 F528 2EE4"
+latest_version: 29
+latest_release_fingerprint: "5A03B4DD8254ECA02FDA1637A20AA56B429476B4"
+latest_pkg_release: 5a886537
```
matejak: I would add quotes here, so the version string is not confused with a number by the YAML parser.

matejak:
Thank you for your pull request, it presents a great improvement of handling Fedora keys. I have some remarks that could cause problems if not fixed, and a completely optional remark about using Jinja2 macros in the OVAL.

linux_os/guide/system/software/updating/ensure_fedora_gpgkey_installed/bash/shared.sh
yuumasato: Thanks for the references, unfortunately I cannot see the video.

fedora/product.yml
yuumasato: Nice catch, made them strings.

yuumasato:
@matejak thank you for the review.
I have made the improvements. The macro made the check look slim and beautiful.

matejak:
Great work, thank you very much for reducing the OVAL boilerplate as well!

PR #3643, opened 2018-12-05
Split rule installed_OS_is certified

jan-cerny:
#### Description:
Splits rule `installed_OS_is certified` to 2 rules:
 * installed OS is vendor supported (is RHEL)
 * installed OS has received FIPS certification

The rules that require FIPS certification (eg. `sshd_use_approved_ciphers`) will now depend on `installed_OS_is_FIPS_certified`. The profiles will contain `installed_OS_is_vendor_supported`

#### Rationale:

The original intention of the rule installed_OS_is_certified was to serve as dependency for FIPS-related checks such as` grub2_enable_FIPS_mode`. Over the time new requirements have been added to ensure Red Hat Enterprise Linux is evaluated (and not CentOS).

I am not sure about the wording. Also, should the rule mention also different certifications, like CC? I understand CC certification as independent on FIPS 140-2 ceritfication. Do you think some the profiles need to contain `installed_OS_is_FIPS_certified`? Or do we keep  `installed_OS_is_FIPS_certified` just as a dependency? What about OSPP which already contains `grub2_enable_FIPS_mode`? Would you suggest a different group than "integrity"?


linux_os/guide/system/software/integrity/certified-vendor/installed_OS_is_FIPS_certified/oval/shared.xml
```diff
@@ -1,16 +1,15 @@
...
-  <definition class="compliance"
-  id="installed_OS_is_certified" version="1">
+  <definition class="compliance" id="installed_OS_is_FIPS_certified" version="1">
     <metadata>
-      <title>Vendor Certified Operating System</title>
+      <title>FIPS 140-2 Certified Operating System</title>
       <affected family="unix">
         <platform>multi_platform_rhel</platform>
         <platform>multi_platform_rhosp</platform>
         <platform>multi_platform_fedora</platform>
```
yuumasato: I think Fedora should be removed, not sure about OpenStack though.

linux_os/guide/system/software/integrity/certified-vendor/installed_OS_is_vendor_supported/oval/shared.xml
```diff
@@ -0,0 +1,21 @@
+<def-group>
+  <definition class="compliance" id="installed_OS_is_vendor_supported" version="1">
+    <metadata>
+      <title>Vendor Supported Operating System</title>
+      <affected family="unix">
+        <platform>multi_platform_rhel</platform>
+        <platform>multi_platform_rhosp</platform>
+        <platform>multi_platform_fedora</platform>
```
yuumasato: I think Fedora should not be considred a supported system.

linux_os/guide/system/software/integrity/certified-vendor/installed_OS_is_FIPS_certified/rule.yml
```diff
@@ -0,0 +1,44 @@
+documentation_complete: true
+
+prodtype: rhel6,rhel7,rhel8,fedora,ol7
```
yuumasato: Fedora should removed.

rhel7/profiles/stig-rhel7-disa.profile
```diff
@@ -119,7 +119,7 @@ selections:
     - selinux_policytype
     - disable_ctrlaltdel_reboot
     - accounts_umask_etc_login_defs
-    - installed_OS_is_certified
+    - installed_OS_is_vendor_supported
```
yuumasato: Maybe `stig-rhel7-disa` should select both rule,` vendor_supported` and `FIPS_certified`.
@shawndwells would appreciate your input here

linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/oval/shared.xml
```diff
@@ -10,7 +10,7 @@
       <description>Look for argument fips=1 in the kernel line in /etc/default/grub.</description>
     </metadata>
     <criteria operator="AND">
-      <extend_definition comment="Installed OS is certified" definition_ref="installed_OS_is_certified" />
+      <extend_definition comment="Installed OS is FIPS certified" definition_ref="installed_OS_is_FIPS_certified" />
```
yuumasato: I'm a bit confused about what checks we should be evaluating for this rule (and others witch extend definition `installed_OS_is_certified`.

In the rule, description only talks about how to enable FIPS mode, and there is a regulatory warning saying that `...ability to enable FIPS mode does not denote FIPS compliancy or certification.`
So there can be situation 1. **a FIPS certified OS with FIPS mode enabled** and situation 2. **a non-FIPS certified OS with FIPS mode enabled**. 

From rule warning I understand that both situation 1 and 2 should evaluate to pass, because FIPS is enabled.
So this check should actually have `installed_OS_is_certified` removed and not replaced by any other check?

So, as I understand, the verification for FIPS certified system should be done by selecting the rule `intalled_OS_is_FIPS_certified` in relevant profiles.

linux_os/guide/system/software/integrity/certified-vendor/installed_OS_is_FIPS_certified/oval/shared.xml
jan-cerny: @yuumasato It doesn't mean that the definition passes on Fedora, it means that the definition is applicable to Fedora. The definition fails on Fedora, it passes only on RHEL6 and RHEL7.

linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/oval/shared.xml
jan-cerny: Great question!

I was also confused. I understand the purpose of the `extend_definiton`s to `installed_OS_is_certified` in this kind of rules as a check if installed OS has received FIPS certification. I think that the expectation is that all these rules (eg. fail on a system that isn't FIPS certified.  I don't know if that's correct or not, but it preserves the current behavior.

redhatrises: > I think that the expectation is that all these rules (eg. fail on a system that isn't FIPS certified. I don't know if that's correct or not, but it preserves the current behavior.

@jan-cerny you are very correct.

jan-cerny: So that means the extend_definition is used here correctly and shouldn't be removed. @redhatrises Apart from that, what about @yuumasato 's suggestion to include rule `installed_OS_is_FIPS_certified` in profiles?

linux_os/guide/system/software/integrity/certified-vendor/installed_OS_is_FIPS_certified/oval/shared.xml
yuumasato: This could be `multi_platform_all` then.

redhatrises: Actually no, the applicability that is currently set should stay.

linux_os/guide/system/software/integrity/certified-vendor/installed_OS_is_FIPS_certified/rule.yml
redhatrises: We should remove fedora from anything related to FIPS.

linux_os/guide/system/software/integrity/certified-vendor/installed_OS_is_vendor_supported/oval/shared.xml
yuumasato: Making this check applicable to Fedora doesn't make it a Vendor Supported OS.

linux_os/guide/system/software/integrity/certified-vendor/installed_OS_is_FIPS_certified/rule.yml
yuumasato: > I think if we remove Fedora from here now, the rules will be missing OVAL, because the dependent definition won't be there.

I think that as long as OVAL definition for `installed_OS_is_FIPS_certified` is marked as applicable for Fedora, it should be fine. This needs to be tested though.

> We should remove fedora from anything related to FIPS.

@redhatrises Do you consider it crucial that FIPS related rules are not listed in Fedora benchmarks?

linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/oval/shared.xml
yuumasato: I gave it some thought, and I consider adding `installed_OS_is_FIPS_certified` rule into profiles could be done if required formally or explicitly. Anyhow, FIPS certification is be checked indirectly by the `_fips_` rules that extend definition `installed_OS_is_FIPS_certified`.

linux_os/guide/system/software/integrity/certified-vendor/installed_OS_is_FIPS_certified/oval/shared.xml
yuumasato: I don't have a strong opinion on this. I'm okay with the applicability as is now,  but I don't see why this can't apply for other systems, they can also be FIPS certified.

PR #3646, opened 2018-12-07
Select crypto policy rules in rhel8

yuumasato:
#### Description:

- Selects crypto-policy related rules in RHEL8
- Rename `kerberos_crypto_policy` fixes to be part of shared, so that it can be used by Fedora and RHEl8.

#### Rationale:

- Configuring system-wide crypto policy should be the way to configure cryptographic algorithms to use in the system.
- Selecting the appropriate crypto policy for you reduces the risk of using inconsistent and outdated algorithms.


rhel8/profiles/ospp.profile
```diff
@@ -198,3 +198,10 @@ selections:
     - mount_option_dev_shm_nodev
     - mount_option_dev_shm_noexec
     - mount_option_dev_shm_nosuid
+    - var_system_crypto_policy=fips
```
matusmarhefka: What is the reason of having FIPS instead of DEFAULT for ospp?

matejak: It's because OSPP requires FIPS mode.

yuumasato: To be consistent with rule `grub2_enable_fips_mode`, selected in `ospp` profile.

PR #3648, opened 2018-12-10
Fix the machine/container filter for Ansible remediations.

matejak:
The previous behaviour could result in rule being skipped if the target was a guest (even a VM one) OR it had something to do with docker.
Now, it is skipped iff both of those statements are true.


yuumasato:
Thank you. The PR description needs to be updated WRT new changes.

The rule is skipped when the target system is a "docker" guest.
In other words, the Rule is run if the target is a not a `guest`, or, in case it is a `guest`, not a `docker` type guest.

yuumasato:
LGTM, fixes the case when a rule marked as machine only cannot be tested on a VM via TestSuite.

PR #3649, opened 2018-12-10
Extend ol7 CPE dict with container and machine platforms

iokomin:
#### Description:

Extend OL7 CPE dict with "container" and "machine"

#### Rationale:

Set of shared rules updated with "cpe:/a:machine" platform limitation leading to "notapplicable" evaluation result on OL7.

Note: I might think it could be reasonable to update openscap package default cpe dictionary to include these items otherwise "--cpe" option will be always required for evaluation scans.

#### Testing:
- Checked to work on OL7
- Checked "make all" build to pass.


redhatrises:
ACK

PR #3658, opened 2018-12-11
adding ansible remediation for kerberos encryption via audispd in rhel7

aghassemlouei:
#### Description:

- Adding ansible remediation/fix content as per DISA STIG RHEL-07-030310

#### Rationale:

- No previous Ansible remediation/fix content existed


matejak:
add to whitelist

jan-cerny:
Hi, I tried to generate a playbook from the built RHEL7 datastream and then apply the playbook on my RHEL7 machine. It works OK, the configuration entry gets there. Thank you very much for your contribution.

Just a note than in future we will need a different Ansible task for RHEL8 and Fedora because the configuration path and the option has changed in audit 3.0.

PR #3662, opened 2018-12-12
Make the man page depend of derivate contents

yuumasato:
#### Description:

- The man page should also depend on derivative content

#### Rationale:

- Fixes a race condition scenario where the man page build starts before the derivative datastream is finished.
- See build [logs](https://kojipkgs.fedoraproject.org//work/tasks/4965/31424965/build.log).


jan-cerny:
You're a genius!

PR #3665, opened 2018-12-17
Add OL8 product support

iokomin:
#### Description:

Add OL8 product support:
- Created OL8 product infrastructure
- Migrated existing OL7 rules: guide content, remediation, templates
- Minor fix - added default to file_groupowner_logfiles_value.var, otherwise first in the list 'adm' is default on rhel and ol.
- Added initial versions of OL8 standard and pci-dss profiles

Note: I realize this PR is quite huge however it mostly includes OL8 product infra and extension in existing OL7 rules. Just let me know if splitting into smaller PRs would be more convenient here for review.

#### Rationale:

OL8 will be released soon, new product version support is required.

#### Testing:
Verified OL8 and 'all' build targets to pass
Checked OL8 pci-dss and standard profiles generated content
Checked migrated rules evaluation and remediation scripts to work on target OS


linux_os/guide/system/auditing/auditd_configure_rules/audit_kernel_module_loading/audit_rules_kernel_module_loading_finit/ansible/shared.yml
```diff
@@ -1,4 +1,4 @@
-# platform = multi_platform_rhel
+# platform = multi_platform_rhel, multi_platform_ol
```
shawndwells: small nitpick:  this line adds a space whereas others remove it (e.g. ``audit_rules_kernel_module_loading_finit``).

iokomin: @shawndwells, thanks for your review..I agree, looks inconsistent. I noticed spaces in remediation platforms lists handled in different ways (even for RHEL and fedora types). Thus in some cases I was trying to add OL keeping existing approach, while in general preferred to trim values and leave just commas. 
To use spaces or not still confusing stuff for me. Is there any guidance/preference for such lists style to follow?

PR #3666, opened 2018-12-17
Improve description of Rule enable_fips_mode 

yuumasato:
#### Description:

- Improve description of `enable_fips_mode`, per suggestion in https://github.com/ComplianceAsCode/content/pull/3623#discussion_r241686005

#### Rationale:

- Previous description could imply that a system needs to be certified by NIST to be FIPS compliant, while actually the components implementing cryptography need to be FIPS certified by NIST.


redhatrises:
In the compliance realm, the existing statement is viewed more accurately than this change because the view is looked at more at a product level than a module level. This change is also not accurate, because it can be misinterpreted that once a component e.g. openssl is certified by NIST, it doesn't matter what product it runs on (which it does).

yuumasato:
Oh my, I don't know what is what anymore.

yuumasato:
So, from Compliance point of view, original descriptions seems good, but from technical component level point of view it seems inaccurate. 

From a high level point of view, I see the rules and its descriptions as targeted to auditors and administrators going through an audit. So I lean towards keeping the description as it is.

PR #3667, opened 2018-12-20
Simply some product profile names

redhatrises:
- Simplifies chromium, firefox, and jre profile names going forward
- Simplifies ocp and osp profile names going forward

This doesn't change existing in production (rhel6, rhel7, etc.) profiles, so the change is not that invasive. The applications chromium, firefox, and jre content needs to be updated for the latest releases. OCP and OSP content is considered Beta.


shawndwells:
lgtm , thanks @redhatrises

PR #3669, opened 2019-01-02
Fix extended definition reference in enable_fips_mode OVAL

jan-cerny:
There is a problem that the rule `enable_fips_mode` introduced in https://github.com/ComplianceAsCode/content/pull/3623 depends on `installed_OS_is_certified`. However, that rule which was removed by the https://github.com/ComplianceAsCode/content/pull/3643, which was unfortunately merged before updating the rule `enable_fips_mode` accordingly. As a result, the rule `enable_fips_mode` in the built datastream doesn't contain OVAL.


matejak:
I can confirm that this PR fixes the regression that has been introduced earlier. Thank you for this PR!

PR #3670, opened 2019-01-04
Remove a stray greater than in auditd userhelper rule

sikevux:
#### Description:

Currently the autid_rules_privileged_commands_userhelper is using a greater than after "auditd" unlike all other auditd rules.

#### Rationale:

To make sure that all auditd rules looks the same in the title field


yuumasato:
@openscap-ci ok to test

yuumasato:
Thanks for the fix.
Ack once Jenkin finishes.

PR #3673, opened 2019-01-07
Update applicability of crypto policy tests

yuumasato:
#### Description:

- Restrict tests for system-wide crypto policy to platforms Fedora and RHEL8 using profiles OSPP and standard.
- Select crypto policy rules in Fedora OSPP profile.

#### Rationale:

- Makes it easier to test crypto policy rules.


redhatrises:
LGTM

PR #2480, opened 2017-12-08
Deprecate RhostsRSAAuthentication as it have been deprecated in 7.4

redhatrises:
#### Description:

- Deprecate RhostsRSAAuthentication as it have been deprecated in 7.4 due to the removal of Protocol 1 in sshd_config

#### Rationale:

- RhostsRSAAuthentication has been deprecated
- Fixes #2478

PR #3675, opened 2019-01-08
Remove the sshd_disable_rhosts_rsa rule from RHEL8 and Fedora profiles.

matejak:
The rule references a sshd configuration option that is not relevant for up-to-date RHEL>=7 systems.

There is no mention of `RhostsRSAAuthentication` in the manpage of `sshd_config` on RHEL 7.6, RHEL 8 and Fedora. On RHEL 6.10, it says that the `RhostsRSAAuthentication` is off by default, and that it applies to SSH v1 only.
As I don't know how the situation is e.g. on RHEL 7.1, I kept it in RHEL 7 profiles, but I am pretty sure that it is redundant in RHEL 8 and Fedora.

### DIFF
--- a/fedora/profiles/ospp.profile
+++ b/fedora/profiles/ospp.profile
@@ -72,7 +72,6 @@ selections:
     - disable_host_auth
     - sshd_disable_gssapi_auth
     - sshd_disable_kerb_auth
-    - sshd_disable_rhosts_rsa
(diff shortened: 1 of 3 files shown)


shawndwells:
some of this was touched on in https://github.com/ComplianceAsCode/content/pull/2480

PR #3677, opened 2019-01-08
Remove rules that are shadowed by crypto policies rules.

matejak:
`sshd_use_approved_ciphers` and `sshd_use_approved_macs` mandated usage of FIPS-enabled algorithms, I have replaced them with FIPS crypto policy setup rules.

Since introduction of crypto policies, those old rules are completely irrelevant - all modifications to the sshd config is overriden by current crypto policy setup in RHEL8 and Fedora.
Those rules were not present in Fedora profiles.

### DIFF
--- a/rhel8/profiles/cjis.profile
+++ b/rhel8/profiles/cjis.profile
@@ -99,7 +99,9 @@ selections:
     - sshd_disable_empty_passwords
     - sshd_enable_warning_banner
     - sshd_do_not_permit_user_env
-    - sshd_use_approved_ciphers
+    - var_system_crypto_policy=fips
+    - configure_crypto_policy
+    - configure_ssh_crypto_policy
     - kernel_module_dccp_disabled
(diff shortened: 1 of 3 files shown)


redhatrises:
This PR makes sense to me. @shawndwells agree or foresee any issues?

shawndwells:
How do we make sure the available ciphers are only those that are FIPS validated, vs the machine being in FIPS mode?

yuumasato:
> On the other hand, when the system is switched into the FIPS mode, I am pretty sure that flipping the current crypto policy to FIPS is part of the process.

Yes, command `fips-mode-setup --enable` will select crypto-policy FIPS. Additionaly, rule `enable_fips_mode` checks that selected crypto-policy is FIPS.

redhatrises:
@shawndwells do the statements above satisfy your questions?

shawndwells:
@redhatrises yeah, seems good!

PR #3678, opened 2019-01-08
Save and restore value of IFS

yuumasato:
#### Description:

- Save value of IFS before overwriting it, and restore it after use.

#### Rationale:

- Some bash fix scripts do not set back the orignal value of IFS.
  Correct approach is to save and restore IFS value.

Thanks @matejak for pointing this out.

### DIFF
--- a/linux_os/guide/system/permissions/partitions/mount_option_nodev_nonroot_local_partitions/bash/rhel6.sh
+++ b/linux_os/guide/system/permissions/partitions/mount_option_nodev_nonroot_local_partitions/bash/rhel6.sh
@@ -12,7 +12,8 @@ IFS_BKP="$IFS"
 
 # Load /etc/fstab's content with LABEL= and UUID= tags expanded to real
 # device names into FSTAB_REAL_DEVICES array splitting items by newline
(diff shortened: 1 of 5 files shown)


linux_os/guide/system/permissions/partitions/mount_option_nodev_nonroot_local_partitions/bash/rhel6.sh
```diff
@@ -7,17 +7,22 @@
...
+# Backup IFS value
+IFS_BKP="$IFS"
+
 # Load /etc/fstab's content with LABEL= and UUID= tags expanded to real
 # device names into FSTAB_REAL_DEVICES array splitting items by newline
-IFS=$'\n' FSTAB_REAL_DEVICES=($(findmnt --fstab --evaluate --noheadings))
+IFS=$'\n'
+FSTAB_REAL_DEVICES=($(findmnt --fstab --evaluate --noheadings))
 
 for line in ${FSTAB_REAL_DEVICES[@]}
```
matejak: It is great that arrays are used here, as the IFS is undefined at this point.

matejak:
Great job improving the IFS handling! We really use this tweak a lot in our remediations.

PR #3679, opened 2019-01-09
Do not add dconf value type for login banner text

yuumasato:
#### Description:

- ~Make regular expression for dconf_gnome_login_banner_text stricter~
- Do not add `string` during login banner remediations

#### Rationale:

- ~Lets not match `string` for comparison with value `login_banner_text`.~
- According to documentation `string` is not necessary: https://help.gnome.org/admin/system-admin-guide/stable/login-banner.html.en


redhatrises:
@yuumasato does this change also need to happen in remediation scripts?

yuumasato:
@redhatrises I had checked the bash remediation, and it looks good.
But the Ansible remediation needs an update. It needs to filter out regular expression elements, just like it is done in Bash.
Will look into it soon.

yuumasato:
The remediations wont add `string` anymore.
The Ansible remediation still needs to be addressed.

dahaic:
@yuumasato do you plan to continue work on this one? Is it possible to merge just part that is ready?

yuumasato:
@dahaic Sure, it has been lingering for long.

matejak:
@yuumasato Could you please rebase? It looks like I don't have rights. The rebase should be easy - only one file has conflicts and they all follow the same pattern - the master branch is wrong, whereas the commit is right.

ggbecker:
305a2b9864071f6275737df305a38168242263e9 Resolves #4546

yuumasato:
I have rebased and solved some conflicts.
@matejak please, check if I didn't break anything.

matejak:
I find it OK, moreover the test suite tests pass as well.

ggbecker:
PR seems good. I'll complement with a couple of test scenarios which covers RHEL7 stig profile (dod_banner).

Found issues:
The banner: https://github.com/ComplianceAsCode/content/blob/0f3926162f8aaaa2215479e39d5e9c43eae63668/linux_os/guide/system/accounts/accounts-banners/login_banner_text.var#L18

Still fails to check after remediating. This banner is not used by any profile so it is not a blocker for this PR but test coverage should be increased to include all possible values for this variable.

ggbecker:
I've created a new issue to handle the ansible playbook issue (filter out regular expression elements) so the PR doesn't get blocked. #4574

Otherwise LGTM. Merging it.

PR #3680, opened 2019-01-10
Add OL8 crypto policy rules

iokomin:
#### Description:

Add OL8 support to crypto policy rules

#### Testing:
- checked ol8 and all build targets to pass
- checked rules on target system


shawndwells:
lgtm. Thanks @iokomin !

PR #3682, opened 2019-01-11
Added crypto policies info to related rules.

matejak:
This PR aims to expand description of rules on systems that feature system-wide crypto policies. We have rules that are either shadowed by it, or they are not, although one could fear so:

* The `sshd_use_...` rules are shadowed by the system crypto policy.
* The `set_password_hashing_algorithm_...` rules have no relation to the crypto policy, although they may look so.

### DIFF
--- a/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers/rule.yml
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers/rule.yml
@@ -41,6 +41,14 @@ rationale: |-
     utilize authentication that meets industry and government requirements. For government systems, this allows
     Security Levels 1, 2, 3, or 4 for use on {{{ full_name }}}.
 
+{{% if product in ["fedora", "rhel8","ol8"] %}}
+warnings:
+    - general: |-
+        This rule seems to be included in a benchmark of a platform that supports crypto policies.
+        Crypto policies centralise crypto algorithms configuration across the whole system, and unless overriden in a very specific way, they take priority over legacy configuration, such as one described by this rule.
+        Therefore, this rule is irrelevant and shouldn't even be part of the {{{ full_name }}} benchmark at all.
+{{% endif %}}
+
 severity: medium
 
 identifiers:
--- a/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_macs/rule.yml
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_macs/rule.yml
@@ -31,6 +31,14 @@ rationale: |-
     DoD Information Systems are required to use FIPS-approved cryptographic hash
     functions. The only SSHv2 hash algorithms meeting this requirement is SHA2.
 
+{{% if product in ["fedora", "rhel8","ol8"] %}}
+warnings:
+    - general: |-
+        This rule seems to be included in a benchmark of a platform that supports crypto policies.
+        Crypto policies centralise crypto algorithms configuration across the whole system, and unless overriden in a very specific way, they take priority over legacy configuration, such as one described by this rule.
+        Therefore, this rule is irrelevant and shouldn't even be part of the {{{ full_name }}} benchmark at all.
+{{% endif %}}
+
 severity: medium
 
 identifiers:
--- a/linux_os/guide/system/accounts/accounts-pam/set_password_hashing_algorithm/set_password_hashing_algorithm_logindefs/rule.yml
+++ b/linux_os/guide/system/accounts/accounts-pam/set_password_hashing_algorithm/set_password_hashing_algorithm_logindefs/rule.yml
@@ -9,6 +9,10 @@ description: |-
     the system will use SHA-512 as the hashing algorithm:
     <pre>ENCRYPT_METHOD SHA512</pre>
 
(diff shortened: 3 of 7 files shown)


linux_os/guide/system/accounts/accounts-pam/set_password_hashing_algorithm/set_password_hashing_algorithm_logindefs/rule.yml
```diff
@@ -9,6 +9,10 @@ description: |-
     the system will use SHA-512 as the hashing algorithm:
     <pre>ENCRYPT_METHOD SHA512</pre>
 
+    {{% if product in ["fedora", "rhel8","ol8"] %}}
+    Although this rule involves settings related to crypto algorithms, it does not overlap with system-wide Crypto Policies and can be safely used on systems where system-wide Crypto Policies are enabled s.a. {{{ full_name }}}.
```
jan-cerny: crypto policies probably don't use capital letters

linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers/rule.yml
```diff
@@ -41,6 +41,14 @@ rationale: |-
     utilize authentication that meets industry and government requirements. For government systems, this allows
     Security Levels 1, 2, 3, or 4 for use on {{{ full_name }}}.
 
+{{% if product in ["fedora", "rhel8","ol8"] %}}
+warnings:
+    - general: |-
+        This rule seems to be included in a benchmark of a platform that supports crypto policies.
+        Crypto policies centralise crypto algorithms configuration across the whole system, and unless overriden in a very specific way, they take priority over legacy configuration, such as one described by this rule.
```
jan-cerny: should be overridden

linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers/rule.yml
```diff
@@ -41,6 +41,14 @@ rationale: |-
...
+        Therefore, this rule is irrelevant and shouldn't even be part of the {{{ full_name }}} benchmark at all.
+{{% endif %}}
```
jan-cerny: Instead, this rule should specify `prodtype` followed by a list of product that do not have system-wide crypto policies. Then, the rule would never get included into benchmarks in which it doesn't make sense, therefore you won't need to add this warning there.

matejak:
@jan-cerny I will address @redhatrises comments first, as they are about the intention of this PR.

@redhatrises so let me to address your points:
1. What is documented elsewhere? There is no resource to tell you what rules are subject to being shadowed by crypto policies. The sshd rules are easy to guess, but the other ones can really raise doubts.
2. You can't disable crypto policy for a service/library that easily. Moreover it is like bypassing how crypto settings are supposed to work and how they are documented in RHEL8/Fedora. Last but not least, you would be in conflict with other crypto policies rules in SSG.
3. The sshd rules won't show any warning if you don't include them in a wrong product. So considering that, you can consider this PR to be like an emergency handbrake. As @jan-cerny commented, those rules could have prodtypes associated, but then it is a pain maintaining all those lists. As we don't have semantic prodtypes (e.g. product class `products_with_crypto_policies`) and a blacklisting capabilities (prodtype algebra), it is questionable what is worse - introduce a warning when a bad prodtype is selected, or enumerate all applicable prodtypes?

jan-cerny:
1. Most of the stuff mentioned in our rules is documented elsewhere. That doesn't mean we can't write something explicitly if we feel it's important.
2. We have other rules that check if you opt in or opt out from system-wide crypto policy.
3. The recommended way of configuring crypto settings of apps/libraries on RHEL8/Fedora is to use system-wide crypto policies. Other ways aren't recommended. My opinion is that our benchmarks shouldn't recommend something that's not recommended, so I would rather remove these rules from RHEL8/Fedora benchmarks completely.

jan-cerny:
Hi @matejak @redhatrises. What is the status of this?

jan-cerny:
@matejak Please resolve conflicts.

matejak:
This is a non-critical PR. I agree that it should be solved by having more flexible platform system, but we don't have that yet.
Rebasing to resolve conflicts, let's see what comes next.

linux_os/guide/services/ssh/ssh_server/sshd_use_approved_macs/rule.yml
```diff
@@ -31,6 +31,14 @@ rationale: |-
     DoD Information Systems are required to use FIPS-approved cryptographic hash
     functions. The only SSHv2 hash algorithms meeting this requirement is SHA2.
 
+{{% if product in ["fedora", "rhel8","ol8"] %}}
+warnings:
+    - general: |-
+        This rule seems to be included in a benchmark of a platform that supports crypto policies.
+        Crypto policies centralise crypto algorithms configuration across the whole system, and unless overriden in a very specific way, they take priority over legacy configuration, such as one described by this rule.
```
jan-cerny: here again it should be "overridden"

jan-cerny:
@matejak Do you have any updates?

jan-cerny:
@matejak Do you have any updates?

PR #3689, opened 2019-01-16
Fix no_files_unowned_by_user OVAL check to use password_object

### LINKED ISSUE #2160: Rule: xccdf_org.ssgproject.content_rule_no_files_unowned_by_user - Incorrectly detects ldap-user owned files
Rule                                                       |OS           | scap-security-guide 
---------------------------------------------------------- |------------ | -------------------  
xccdf_org.ssgproject.content_rule_no_files_unowned_by_user |RHEL 7.1     | 0.1.34              

The check does not properly detect sssd ldap users and marks files owned by them as unowned, this is also applicable to the unowned_by_group check.

redhatrises:
- password_object uses getent which would allow checking file ownership for centralized
  users if enumeration or getent is setup properly. If enumeration or getent is not setup
  properly, there is no sane fix to this.
- Fixes #2160

### DIFF
--- a/linux_os/guide/system/permissions/files/no_files_unowned_by_user/oval/shared.xml
+++ b/linux_os/guide/system/permissions/files/no_files_unowned_by_user/oval/shared.xml
@@ -17,14 +17,12 @@
   </unix:file_state>
 
   <local_variable id="file_permissions_unowned_userid_list" comment="List of valid user ids" datatype="int" version="1">
-    <object_component item_field="subexpression" object_ref="file_permissions_unowned_userid_list_object" />
+    <object_component item_field="user_id" object_ref="file_permissions_unowned_userid_list_object" />
   </local_variable>
 
-  <ind:textfilecontent54_object id="file_permissions_unowned_userid_list_object" version="1">
-    <ind:filepath>/etc/passwd</ind:filepath>
-    <ind:pattern operation="pattern match">^[^:]+:[^:]+:([\d]+):[\d]+:[^:]*:[^:]+:[^:]*$</ind:pattern>
-    <ind:instance operation="greater than or equal" datatype="int">1</ind:instance>
-  </ind:textfilecontent54_object>
+  <unix:password_object id="file_permissions_unowned_userid_list_object" version="1">
+    <unix:username datatype="string" operation="pattern match">.*</unix:username>
+  </unix:password_object>
 
   <unix:file_object comment="all local files" id="file_permissions_unowned_object" version="1">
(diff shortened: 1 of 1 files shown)


linux_os/guide/system/permissions/files/no_files_unowned_by_user/oval/shared.xml
```diff
@@ -17,14 +17,12 @@
...
-    <object_component item_field="subexpression" object_ref="file_permissions_unowned_userid_list_object" />
+    <object_component item_field="user_id" object_ref="file_permissions_unowned_userid_list_object" />
   </local_variable>
 
-  <ind:textfilecontent54_object id="file_permissions_unowned_userid_list_object" version="1">
-    <ind:filepath>/etc/passwd</ind:filepath>
-    <ind:pattern operation="pattern match">^[^:]+:[^:]+:([\d]+):[\d]+:[^:]*:[^:]+:[^:]*$</ind:pattern>
-    <ind:instance operation="greater than or equal" datatype="int">1</ind:instance>
-  </ind:textfilecontent54_object>
+  <unix:password_object id="file_permissions_unowned_userid_list_object" version="1">
```
shawndwells: ``password_object`` appears to only accept a single user name, not a list. Will the rule iterate through all elements of ``file_permissions_unowned_userid_list_object ``?

Ref https://oval.mitre.org/language/version5.10/ovaldefinition/documentation/unix-definitions-schema.html
``
A password object consists of a single username entity that identifies the user(s) whose password is to be evaluated.
``

redhatrises: > `password_object` appears to only accept a single user name, not a list. Will the rule iterate through all elements of `file_permissions_unowned_userid_list_object `?

Each user id is stored in a local_variable just like it is with textfilecontent54_object.

> Ref https://oval.mitre.org/language/version5.10/ovaldefinition/documentation/unix-definitions-schema.html
> `A password object consists of a single username entity that identifies the user(s) whose password is to be evaluated.`

The language is confusing, but it means that there is no other element other than "username" for the "password_object".

You can run the following before and after the change to double check me:
```PYTHONPATH=`./.pyenv.sh` utils/testoval.py linux_os/guide/system/permissions/files/no_files_unowned_by_user/oval/shared.xml```
I noticed no difference in the results.

shawndwells: Thanks!

PR #3690, opened 2019-01-17
New Test Suite backend that uses Podman

jan-cerny:
#### Description:
Introduces Podman backend to SSG Test Suite harness.

The Podman backend works similar way as the Docker backend. However, it's implemented by invoking podman CLI calls (subprocess) instead of using Python 3 bindings, because as of python3-podman version 0.12.1.2 it doesn't allow to run, inspect and commit containers. Users will use
this backend using `--container` cmdline option.

The Docker backend and Podman backend are very similar because both of them are a specific implementation of a generic container backend.  We introduce a new parent class ContainerTestEnv that contains common code. Then, DockerTestEnv and PodmanTestEnv will inherit from ContainerTestEnv.

The documentation in `tests/README.md` is updated accordingly.

The Dockerfiles fot SSG test suite are updated this way:
- It's more appopriate to describe CentOS container as CentOS-based than RHEL based.
- The utility that generates SSH keys is ssh-keygen, not sshd-keygen which is a legacy service.
- We remove pam_loginuid.so from SSHD PAM configuration in the container as it prevents users to log in via ssh to the container. As our test suite doesn't check if oscap scans produce audit messages, we can simply remove that.

#### Rationale:
On RHEL8, Docker isn't available and it's replaced by Podman. Therefore, we need to implement a backend for Podman to run container-based tests in SSG Test suite.

For more details, please see the respective commit messages of each commit.

### DIFF
--- a/tests/ssg_test_suite/test_env.py
+++ b/tests/ssg_test_suite/test_env.py
@@ -4,6 +4,7 @@ import contextlib
 import sys
 import os
 import time
+import subprocess
 
 import docker
 
@@ -172,7 +173,7 @@ class VMTestEnv(TestEnv):
 
 
 class DockerTestEnv(TestEnv):
-    name = "container-based"
+    name = "docker-based"
 
     def __init__(self, mode, image_name):
         super(DockerTestEnv, self).__init__(mode)
@@ -249,6 +250,7 @@ class DockerTestEnv(TestEnv):
         return new_container
 
     def _find_image_by_name(self, image_name):
+        # only in DockerTestEnv
         return self.client.images.get(image_name)
 
     def _new_container_from_image(self, image_name, container_name):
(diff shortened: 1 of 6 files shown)


tests/ssg_test_suite/test_env.py
```diff
@@ -248,40 +227,152 @@ def reset_state_to(self, state_name, new_running_state_name):
...
+        assert associated_image == image
+        self._remove_image(associated_image)
+
+    def offline_scan(self, args, verbose_path):
+        command_list = self._local_oscap_check_base_arguments() + args
+
+        return common.run_cmd_local(command_list, verbose_path)
+
+    def _commit(self, container, image):
+        pass
```
matejak: I would prefer to have `NotImplementedError` here, as this is basically an Abstract Base Class.

tests/ssg_test_suite/test_env.py
```diff
@@ -248,40 +227,152 @@ def reset_state_to(self, state_name, new_running_state_name):
...
+            subprocess.check_output(podman_cmd, stderr=subprocess.STDOUT)
+        except subprocess.CalledProcessError as e:
+            msg = "Command '{0}' returned {1}:\n{2}".format(" ".join(e.cmd), e.returncode, e.output.decode("utf-8"))
+            raise RuntimeError(msg)
+
+    def _new_container_from_image(self, image_name, container_name):
+        long_name = "{0}_{1}".format(self._name_stem, container_name)
+        podman_cmd = ["podman", "run", "--name", long_name,
+                      "--publish", "22", "--detach", image_name,
+                      "/usr/sbin/sshd", "-D"]
```
matejak: This is rather a question, but do we really need to override the `CMD` from the `Dockerfile`? We don't do that in the Docker case.

tests/ssg_test_suite/test_env.py
jan-cerny: OK, I'll change that.

tests/ssg_test_suite/test_env.py
jan-cerny: We do that in Docker case:
https://github.com/ComplianceAsCode/content/blob/c3178080fc69045a1ea5dd4cd35be79d0bf33023/tests/ssg_test_suite/test_env.py#L256-L259

Moreover, our Dockerfiles do not specify `CMD`, therefore if we use our Dockerfiles, we aren't overriding anything.

tests/ssg_test_suite/test_env.py
jan-cerny: Changed.

tests/ssg_test_suite/test_env.py
matejak: Damn, you are right!

matejak:
Good job, thank you!

PR #3692, opened 2019-01-18
Fix handling of key in remediation of audit rules privileged commands

yuumasato:
#### Description:

- Drop redundant `remove_all_rules.fail` test scenario
- Update platform of test scenarios
- Add new test scenarios to handle audit rules whose keys are specified using both `-k mykey` and `-F key=mykey`. Both settings should be accepted.
  - two_rules_mixed_keys: The remediation script should be able to handle mixed keys
  - rules_configured_mixed_keys: The OVAL should be able to recognize mixed keys

#### Rationale:

- Bash fix did not recognize correct audit rules with keys specified using `-F key=` syntax.
- Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1667108.


jan-cerny:
Thanks

PR #3697, opened 2019-01-23
Add test scenario for accounts_password_minlen_login_defs

yuumasato:
#### Description:

- Add test scenario for `accounts_password_minlen_login_defs`


jan-cerny:
We can have also scenarios checking that:
* PASS_MIN_LEN is not present at all
* PASS_MIN_LEN is commented out

yuumasato:
@jan-cerny Well noted, they have been added.

Issue #3705, opened 2019-01-28
assign severity to accounts_logon_fail_delay


shawndwells:
RHEL-07-010430 (Ensure the Logon Failure Delay is Set Correctly in login.defs)

SSG: Unknown
STIG: Medium

DISA believes this should be medium, however by their own definition only meets low.

@tedbrunell can you triage this with DISA?

shawndwells:
CAT I (high)
Any vulnerability, the exploitation of which will directly and immediately result in loss of Confidentiality, Availability, or Integrity.

CAT II (medium)
Any vulnerability, the exploitation of which has a potential to result in loss of Confidentiality, Availability, or Integrity.

CAT III (low)
Any vulnerability, the existence of which degrades measures to protect against loss of Confidentiality, Availability, or Integrity.

PR #3707, opened 2019-01-28
assign low severity to accounts_logon_fail_delay

shawndwells:
Rule currently does not have a severity code. https://github.com/ComplianceAsCode/content/issues/3705 requests alignment to DISA rating of "medium" however this appears to only be low.

Other rules limit false logon attempts, whereas this only effects delay between the logon prompts. Failure to implement this rule could "degrade" protections against DDoS but not leave the system exposed to them (since # inaccurate logins will still lock the account).

While https://github.com/ComplianceAsCode/content/issues/3705 gets figured out, this will at least assign something to severity instead of "unknown"


redhatrises:
Ack

PR #3721, opened 2019-01-28
Show MixedIndentationError message when you mix tabs and spaces in yaml

### LINKED ISSUE #3698: Show an error message in CMake when you mix indentation in YAML files
#### Description of problem:
YAML requires to use spaces for indents. However, it is easy to mix spaces and tabs. If you break indentations, the build fails. The error message is extremely unhelpful and it's hard to find out the root cause of the problem.

current upstream as of 2018-01-25

all

#### Steps to Reproduce:

1. use wrong indentation in some of the *.profile files
2. cmake ..
3. make
4.

#### Actual Results:
weird error message

#### Expected Results:
useful error message is displayed that points the user directly to the problem

redhatrises:
- Fixes #3698

### DIFF
--- a/ssg/yaml.py
+++ b/ssg/yaml.py
@@ -4,6 +4,7 @@ from __future__ import print_function
 import codecs
 import yaml
 import sys
+import re
 
 from .jinja import extract_substitutions_dict_from_template, process_file
 from .constants import (PKG_MANAGER_TO_SYSTEM,
@@ -48,9 +49,20 @@ def _open_yaml(stream, original_file=None, substitutions_dict={}):
 
         return yaml_contents
     except Exception as e:
+        count = 0
         _file = original_file
         if not _file:
             _file = stream
+        with open(_file, "r") as e_file:
+            lines = e_file.readlines()
+            for line in lines:
+                count = count + 1
+                if re.match(r"^\s*\t+\s*", line):
+                    print("Exception while handling file: %s" % _file, file=sys.stderr)
+                    print("TabIndentationError: Line %s contains tabs instead of spaces:" % (count), file=sys.stderr)
+                    print("%s\n\n" % repr(line.strip("\n")), file=sys.stderr)
+                    sys.exit(1)
+
         print("Exception while handling file: %s" % _file, file=sys.stderr)
         raise e
 


jan-cerny:
Hi @redhatrises ,

Yesterday I took a look into the same thing, but I didn't come to this creative solution.

I have found that yaml.load() method raises YAMLError exception, which contains the exact location of the problem. However, I have tried that and I found that it works only if we don't use CSafeLoader parser, which we want to use, because it's way faster.

ssg/yaml.py
```diff
@@ -48,9 +49,20 @@ def _open_yaml(stream, original_file=None, substitutions_dict={}):
...
     except Exception as e:
+        count = 0
         _file = original_file
         if not _file:
             _file = stream
+        with open(_file, "r") as e_file:
+            lines = e_file.readlines()
+            for line in lines:
+                count = count + 1
+                if re.match(r"^\t+\s+", line) or re.match(r"^\s+\t+", line):
```
jan-cerny: If there is just a single tab, this will not match. I have found that the YAML spec doesn't allow using tabs for indentation at all.  See https://yaml.org/spec/1.1/#indentation%20space/

ssg/yaml.py
```diff
@@ -48,9 +49,20 @@ def _open_yaml(stream, original_file=None, substitutions_dict={}):
...
+                    print("Exception while handling file: %s" % _file, file=sys.stderr)
+                    print("TabIndentationError: Line %s contains tabs instead of spaces:" % count)
+                    print("%s\n\n" % repr(line.strip("\n")))
```
jan-cerny: Please add `, file=sys.stderr` to all `print` statements, otherwise the error message won't be visible for the user.

redhatrises: @jan-cerny thanks. Fixed it!

ssg/yaml.py
redhatrises: @jan-cerny thanks. Fixed like how we talked.

ssg/yaml.py
```diff
@@ -48,9 +49,20 @@ def _open_yaml(stream, original_file=None, substitutions_dict={}):
...
         _file = original_file
         if not _file:
             _file = stream
+        with open(_file, "r") as e_file:
+            lines = e_file.readlines()
+            for line in lines:
+                count = count + 1
+                if re.match(r"^\s*\t+\s*", line):
+                    print("Exception while handling file: %s" % _file, file=sys.stderr)
+                    print("TabIndentationError: Line %s contains tabs instead of spaces:" % count)
```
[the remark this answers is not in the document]
redhatrises: Oops... Thanks. @jan-cerny fixed

jan-cerny:
Thanks!

PR #3723, opened 2019-02-01
Fix Ansible remediation for PAM

jan-cerny:
#### Description:
1.  Replace ':' by '=' in PAM module_arguments
This ':' was propagated to PAM configuration file. But ':' is not a valid character in PAM configuration in module arguments. Instead, '=' should be used.

2. Apply Ansible remediations for rules `accounts_passwords_pam_faillock_deny_root` and `accounts_passwords_pam_faillock_deny` to both `/etc/pam.d/system-auth` and `/etc/pam.d/password-auth`. The rule descriptions in these rules say that both `system-auth` and `password-auth` should be modified, but Ansible remediations modified only `system-auth`.

3. Aling `pam_faillock.so` settings with OVAL and rule description for rules `accounts_passwords_pam_faillock_deny_root` and `accounts_passwords_pam_faillock_deny`.

4.  Run test scenario `missing-system-auth.fail.sh` only with bash. The Ansible remediation fails if PAM config files aren't present at all.

5. Each Ansible remediation will fix only the argument relevant to its rule.

#### Rationale:
RHBZ https://bugzilla.redhat.com/show_bug.cgi?id=1636392 complains about broken PAM configuration after applying Ansible playbooks.

### DIFF
--- a/tests/data/group_system/group_accounts/group_accounts-pam/group_locking_out_password_attempts/rule_accounts_passwords_pam_faillock_deny_root/missing-system-auth.fail.sh
+++ b/tests/data/group_system/group_accounts/group_accounts-pam/group_locking_out_password_attempts/rule_accounts_passwords_pam_faillock_deny_root/missing-system-auth.fail.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 # profiles = xccdf_org.ssgproject.content_profile_nist-800-171-cui, xccdf_org.ssgproject.content_profile_ospp
+# remediation = bash
+
 . set-up-pamd.sh
 
 set-up-pamd
--- a/tests/data/group_system/group_accounts/group_accounts-pam/group_locking_out_password_attempts/rule_accounts_passwords_pam_faillock_deny_root/missing-system-auth.fail.sh
+++ b/tests/data/group_system/group_accounts/group_accounts-pam/group_locking_out_password_attempts/rule_accounts_passwords_pam_faillock_deny_root/missing-system-auth.fail.sh
@@ -1,8 +1,0 @@
-#!/bin/bash
-# profiles = xccdf_org.ssgproject.content_profile_nist-800-171-cui, xccdf_org.ssgproject.content_profile_ospp
-# remediation = bash
-
-. set-up-pamd.sh
-
-set-up-pamd
-rm -f /etc/pam.d/system-auth
--- a/linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_passwords_pam_faillock_deny/ansible/shared.yml
+++ b/linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_passwords_pam_faillock_deny/ansible/shared.yml
@@ -18,7 +18,7 @@
     new_module_path: pam_faillock.so
     module_arguments: 'preauth
         silent
-        deny: {{ var_accounts_passwords_pam_faillock_deny }}
+        deny={{ var_accounts_passwords_pam_faillock_deny }}
         unlock_time={{ var_accounts_passwords_pam_faillock_unlock_time }}
         fail_interval={{ var_accounts_passwords_pam_faillock_fail_interval }}'
     state: before
@@ -37,7 +37,7 @@
     new_module_path: pam_faillock.so
     module_arguments: 'preauth
         silent
-        deny: {{ var_accounts_passwords_pam_faillock_deny }}
+        deny={{ var_accounts_passwords_pam_faillock_deny }}
         unlock_time={{ var_accounts_passwords_pam_faillock_unlock_time }}
         fail_interval={{ var_accounts_passwords_pam_faillock_fail_interval }}'
     state: after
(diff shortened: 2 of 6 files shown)


jan-cerny:
@openscap-ci test this please

yuumasato:
I don't know what is missing in this PR, all test scenarios pass:
```
[coruscant tests{pam_quotes}]$ ./test_suite.py rule --libvirt qemu:///system rhel7.6-kicked --datastream ../build/ssg-rhel7-ds.xml --xccdf-id scap_org.open-scap_cref_ssg-rhel7-xccdf-1.2.xml --dontclean --debug --remediate-using=ansible accounts_passwords_pam_faillock_deny
INFO - Logging into /home/wsato/git/scap-security-guide/tests/logs/rule-custom-2019-02-12-1316/test_suite.log
INFO - xccdf_org.ssgproject.content_rule_accounts_passwords_pam_faillock_deny
INFO - Script skip_longer.fail.sh using profile xccdf_org.ssgproject.content_profile_C2S OK
ERROR - Script skip_longer_comment.fail.sh using profile xccdf_org.ssgproject.content_profile_C2S found issue:
ERROR - Scan has exited with return code 0, instead of expected 2 during stage initial
Paused for manual debugging. Continue by pressing return.
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_accounts_passwords_pam_faillock_deny'.
INFO - Script default_die_pam_unix.fail.sh using profile xccdf_org.ssgproject.content_profile_C2S OK
INFO - Script skip_correctly.pass.sh using profile xccdf_org.ssgproject.content_profile_C2S OK
INFO - Script config_without_skip.pass.sh using profile xccdf_org.ssgproject.content_profile_C2S OK
INFO - Script remediable_sssd_authconfig.fail.sh using profile xccdf_org.ssgproject.content_profile_C2S OK
INFO - Script skip_correctly_short.pass.sh using profile xccdf_org.ssgproject.content_profile_C2S OK
INFO - Script disablefaillock_authconfig.fail.sh using profile xccdf_org.ssgproject.content_profile_C2S OK
INFO - xccdf_org.ssgproject.content_rule_accounts_passwords_pam_faillock_deny_root
INFO - Script missing-required-pam_faillock.fail.sh using profile xccdf_org.ssgproject.content_profile_nist-800-171-cui OK
INFO - Script missing-required-pam_faillock.fail.sh using profile xccdf_org.ssgproject.content_profile_ospp OK
INFO - Script missing-default-die-pam_faillock.fail.sh using profile xccdf_org.ssgproject.content_profile_nist-800-171-cui OK
INFO - Script missing-default-die-pam_faillock.fail.sh using profile xccdf_org.ssgproject.content_profile_ospp OK
INFO - Script missing-system-auth.fail.sh using profile xccdf_org.ssgproject.content_profile_nist-800-171-cui OK
INFO - Script missing-system-auth.fail.sh using profile xccdf_org.ssgproject.content_profile_ospp OK
INFO - Script both-correct.pass.sh using profile xccdf_org.ssgproject.content_profile_nist-800-171-cui OK
INFO - Script both-correct.pass.sh using profile xccdf_org.ssgproject.content_profile_ospp OK
```

Test scenario `skip_longer_comment.fail.sh` errors due to bug in OVAL.

One remark is that this Ansible remediation concerns rules:
- `accounts_passwords_pam_faillock_deny`
- `accounts_passwords_pam_faillock_unlock_time`
- `accounts_passwords_pam_faillock_interval`

I believe that individual remediations would not work, as they would be fighting each other regarding witch module arguments should be there, and the last one executed would win.

So, should all those three rules have this very same remediation?
At the same time all three rules tightly coupled, along with 3 variables used to define their parameters.

dahaic:
@yuumasato when I read `pamd` page, I see this behaviour described:
```
When state is 'args_present' any args listed in module_arguments are added if missing from the existing rule. Furthermore, if the module argument takes a value denoted by '=', the value will be changed to that specified in module_arguments.
```
Isn't this what we need?

yuumasato:
> Isn't this what we need?

So it will no remove any not argument not present?

yuumasato:
When `state` is `after` or `before`, it will add the new pam rule with arguments if it was not there previously, but if it was there, it won't add the arguments.
When `state` is `args_present`, it will add the arguments only when the exact module is found. 

So I have split each`preauth` and `authfail` tasks into two:
- first will add the `pam_faillock` rule with correct arguments if it was not there yet;
- second will make sure the correct argument is there, in case there was a `pam_faillock` rule already.

tests/data/group_system/group_accounts/group_accounts-pam/group_locking_out_password_attempts/rule_accounts_passwords_pam_faillock_deny_root/missing-system-auth.fail.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 # profiles = xccdf_org.ssgproject.content_profile_nist-800-171-cui, xccdf_org.ssgproject.content_profile_ospp
+# remediation = bash
```
dahaic: After some thinking about this, I believe this scenario is completely invalid. We have some rules, where missing config file is finding that is fixable. Usually it's to override some default values.

PAM configurations are not for overriding defaults, but to define behaviour. So in case system-auth is missing, either machine is completely broken and in very unsupported configuration, or not using PAM. We shouldn't try to fix it in such case.

jan-cerny:
@dahaic I have removed test scenario `missing-system-auth.fail.sh`.

jan-cerny:
@dahaic I have updated the bash remediation for rule `accounts_passwords_pam_faillock_deny_root`

dahaic:
Thank you @jan-cerny !! ACK, Merge

PR #3724, opened 2019-02-01
Made the test suite easier to use:

matejak:
- Expanded the Readme on ssh key creation.
- Modified the test suite to be oscap 1.2-1.3 compatible by fixing the --verbose option.
- Removed usage of `Enum`.
- Made the test suite runnable when the `libvirt` module is not installed.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -224,21 +224,38 @@ Following services need to be supported:
 - `oscap` (`openscap-scanner` - the container has to be able to scan itself)
 - You may want to include another packages, as base images tend to be bare-bone and tests may require more packages to be present.
 
-Using Podman:
+#### Using Podman
 
 NOTE: With Podman, you have to run all the operations as root. Podman supports rootless containers, but the test suite internally uses a container exposing a TCP port. As of Podman version 0.12.1.2, port bindings are not yet supported by rootless containers.
 
-```
-# public_key="ssh-rsa AAAAB3NzaC1y...rJSs4BL root@localhost"
-# podman build --build-arg CLIENT_PUBLIC_KEY="$public_key" -t ssg_test_suite -f test_suite-rhel .
-```
-
-Using Docker:
+The root user therefore needs an key without passphrase, so it can log in to the container without any additional interaction.
+Root user typically doesn't have a SSH key, or it has one without a passphrase.
+Use this test to find out whether it has a key:
 
 ```
-public_key="ssh-rsa AAAAB3NzaC1y...rJSs4BL me@localhost"
-docker build --build-arg CLIENT_PUBLIC_KEY="$public_key" -t ssg_test_suite -f test_suite-rhel .
+sudo test -f /root/.ssh/id_rsa && echo "Root user already has an id_rsa key" || echo "Root user has no id_rsa key"
 ```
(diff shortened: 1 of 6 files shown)


tests/README.md
```diff
@@ -224,22 +224,39 @@ Following services need to be supported:
...
+Go into the `Dockerfiles` directory of the project, and execute the following:
+
+```
+public_key="$(sudo cat /root/.ssh/id_rsa.pub)"
+podman build --build-arg CLIENT_PUBLIC_KEY="$public_key" -t ssg_test_suite -f test_suite-rhel .
+```
+
+#### Using Docker
+
+The procedure is same as using Podman, you just swap the `podman` call with `sudo docker`.
```
jan-cerny: Please make it clear that the commands have to be run as a root. Podman has to be run as root. The convention is to put `#` to beginning of the line to symbolize the root shell prompt.

tests/ssg_test_suite/test_env.py
```diff
@@ -120,6 +119,12 @@ class VMTestEnv(TestEnv):
 
     def __init__(self, mode, hypervisor, domain_name):
         super(VMTestEnv, self).__init__(mode)
+
+        try:
+            import libvirt
+        except ImportError:
+            raise RuntimeError("Can't import libvirt module, libvirt backend will therefore not work.")
```
jan-cerny: It would be nice to specify in README that libvirt is now only an optional dependency

tests/README.md
```diff
@@ -224,22 +224,39 @@ Following services need to be supported:
 - `oscap` (`openscap-scanner` - the container has to be able to scan itself)
 - You may want to include another packages, as base images tend to be bare-bone and tests may require more packages to be present.
 
-Using Podman:
+#### Using Podman
 
 NOTE: With Podman, you have to run all the operations as root. Podman supports rootless containers, but the test suite internally uses a container exposing a TCP port. As of Podman version 0.12.1.2, port bindings are not yet supported by rootless containers.
 
+The root user therefore needs an key without passphrase, so it can log in to the container without any additional interaction.
```
yuumasato: Is the SSH key a hard requirement or usability improvement?

matejak: It is a requirement, because one usually doesn't know the root password. Moreover, one would have to enter it 3 times per scenario, which is IMO totally beyond usability.

tests/ssg_test_suite/test_env.py
yuumasato: I have some additions regarding required packages that I can provide soon.

matejak: This is out of scope of this PR. We don't have the `libvirt` dependency documented here at all.

tests/README.md
matejak: Made that change. The `# command` convention is usually coupled with the `$ command` convention for no-root, which we don't do ATM. For that reason, I have found the hash to be confusing, so I have removed it.

tests/README.md
yuumasato: So it is a requirement, but also with usability effects.
What is confusing to me is that the sentence justifies the requirement with usability argument.

PR #3725, opened 2019-02-01
Implement the rule for dconf to use only the textfile data backend.

### LINKED ISSUE #3736: Extend dconf rule set to enforce usage of the textfile data backend.
OVAL can't check the default dconf binary DB backend, so we have to make sure that we force dconf to use textfile backend. For more info, check out the [Red Hat Bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1631378).

matejak:
The rule contains OVAL, bash and Ansible fix.
There are some test scenarios, the Ansible fix for a config file where the `service-db:keyfile/user` line is present lower in the line is failing. It looks like that the `lineinfile` module is not able to do it's job to a full extent.
Usage of this OVAL check in checks for other rules is not in scope of this PR.
Fixes #3736

### DIFF
--- a/linux_os/guide/system/software/gnome/dconf_use_text_backend/ansible/shared.yml
+++ b/linux_os/guide/system/software/gnome/dconf_use_text_backend/ansible/shared.yml
@@ -1,0 +1,16 @@
+# platform = multi_platform_rhel,multi_platform_fedora,multi_platform_ol
+# reboot = false
+# strategy = unknown
+# complexity = low
+# disruption = medium
+- name: "Force dconf to use the textfiles instead of a binary DB"
+  lineinfile:
+    path: '/etc/dconf/profile/user'
+    regexp: '^service-db:keyfile/user$'
+    line: 'service-db:keyfile/user'
+    insertbefore: 'BOF'
+    create: yes
+  tags:
+    @ANSIBLE_TAGS@
+  @ANSIBLE_ENSURE_PLATFORM@
+
--- a/linux_os/guide/system/software/gnome/dconf_use_text_backend/ansible/shared.yml
+++ b/linux_os/guide/system/software/gnome/dconf_use_text_backend/ansible/shared.yml
@@ -3,13 +3,24 @@
 # strategy = unknown
 # complexity = low
 # disruption = medium
-- name: "Force dconf to use the textfiles instead of a binary DB"
+- name: "Remove the existing \"use textfile backend\" directive from the config - it may not be at the file's very top"
+  lineinfile:
+    path: '/etc/dconf/profile/user'
+    regexp: '^service-db:keyfile/user.*'
+    state: 'absent'
+  check_mode: no
+  tags:
+    @ANSIBLE_TAGS@
+  @ANSIBLE_ENSURE_PLATFORM@
+
+- name: "Insert the \" use textfiles backend\" directive at the top of the config file"
   lineinfile:
     path: '/etc/dconf/profile/user'
     regexp: '^service-db:keyfile/user$'
     line: 'service-db:keyfile/user'
     insertbefore: 'BOF'
     create: yes
+  check_mode: no
   tags:
     @ANSIBLE_TAGS@
   @ANSIBLE_ENSURE_PLATFORM@
--- a/linux_os/guide/system/software/gnome/dconf_use_text_backend/oval/shared.xml
+++ b/linux_os/guide/system/software/gnome/dconf_use_text_backend/oval/shared.xml
@@ -1,0 +1,37 @@
+<def-group>
+  <definition class="compliance" id="dconf_use_text_backend" version="1">
+    <metadata>
+      <title>{{{ full_name }}}</title>
+      <affected family="unix">
+        <platform>Red Hat Enterprise Linux 7</platform>
+        <platform>Red Hat Enterprise Linux 8</platform>
+        <platform>multi_platform_fedora</platform>
+        <platform>multi_platform_ol</platform>
+      </affected>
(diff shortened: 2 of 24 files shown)


jan-cerny:
What about Ansible playbooks? Will they check if the textfile data backend is used?

Could you also take a look at https://docs.ansible.com/ansible/latest/modules/dconf_module.html ?

jan-cerny:
And what about other profiles that contain dconf rules? Do you want to include your new rule to them?

matejak:
Thank you @jan-cerny for checking, I have updated the description, I hope it helps.

jan-cerny:
@matejak Thank you for updating the PR. What about answering my questions?

matejak:
Sorry, some of your questions were indeed not answered:
* The Ansible `dconf` module uses the `dconf` functionality, so it works for any backend.
* I have added the rule to other profiles that use dconf, thanks for noticing.
  Just for reference, I have used this snippet to find applicable profile files:
  ```
  vim -p $(for f in $(grep -Rl --include '*.profile' dconf *); do grep -q dconf_use_tex $f || echo $f; done)
  ```

jan-cerny:
>   The Ansible `dconf` module uses the `dconf` functionality, so it works for any backend.

Do I understand it well that this means that we can't use the `dconf` Ansible module because we require to use only plaintext backend?

matejak:
I think that this is OK:
* It looks like that the Ansible remediation is able to bring the system into a compliant state under any dconf backend. However, it is not documented whether it touches text config files - this is an implementation detail of the module itself, which uses the `dconf` tool / DBUS at the moment.
  Ansible makes sure that runtime values are OK, and OVAL checks for text files, so we are safe here.
* OVAL is able to detect that the system is compliant iff dconf uses the text backend.

So if we force the textfile backend, OVAL checks will reflect the true state of the system, while the Ansible will presumably continue to do the business as usual, as it already keeps text files up-to-date.

linux_os/guide/system/software/gnome/dconf_use_text_backend/oval/shared.xml
```diff
@@ -0,0 +1,37 @@
+<def-group>
+  <definition class="compliance" id="dconf_use_text_backend" version="1">
+    <metadata>
+      <title>{{{ full_name }}}</title>
```
jan-cerny: `full_name` translates to the full name of the product, eg. `Red Hat Enterprise Linux 7". It doesn't look like a right definition title.

matejak: Good catch.

matejak:
I have extended the Ansible to be more robust, but it will not be able to "scan" using the `--check` option.

linux_os/guide/system/software/gnome/dconf_use_text_backend/ansible/shared.yml
```diff
@@ -0,0 +1,26 @@
+# platform = multi_platform_rhel,multi_platform_fedora,multi_platform_ol
+# reboot = false
+# strategy = unknown
+# complexity = low
+# disruption = medium
+- name: "Force dconf to use the textfiles instead of a binary DB"
+  lineinfile:
+    path: '/etc/dconf/profile/user'
+    regexp: '^service-db:keyfile/user.*'
+    create: yes
```
yuumasato: `create` should not be here.

linux_os/guide/system/software/gnome/dconf_use_text_backend/ansible/shared.yml
```diff
@@ -0,0 +1,26 @@
+# platform = multi_platform_rhel,multi_platform_fedora,multi_platform_ol
+# reboot = false
+# strategy = unknown
+# complexity = low
+# disruption = medium
+- name: "Force dconf to use the textfiles instead of a binary DB"
```
yuumasato: Please make the name more meaningful, two tasks with same name, doing different things is very confusing.

linux_os/guide/system/software/gnome/dconf_use_text_backend/ansible/shared.yml
shawndwells: Also, should check if file exists before editing else Ansible will throw an error. Here's code used in the templates to prevent that:

https://github.com/ComplianceAsCode/content/pull/3737/files

matejak: @shawndwells Thank you for your remark, I have double-checked, and discovered that we have have this covered - we have a [test scenario](tests/data/group_system/group_software/group_gnome/rule_dconf_use_text_backend/no_config.fail.sh) that contains a test scenario when the file is absent. I have tested it through and through, and if `create: yes` is missing for the *insert* task, the scenario fails, but when it is missing for the *remove* task, it passes. Therefore, I have modified the PR according to what @yuumasato suggested.

yuumasato:
> but it will not be able to "scan" using the `--check` option.

This can be a confusing issue.
Would it be better to mark those two tasks as check_mode "no"?
Otherwise a playbook with this rule will never pass a check.

linux_os/guide/system/software/gnome/dconf_use_text_backend/ansible/shared.yml
matejak: Good point, renamed.

matejak:
You are correct about the check mode, I have specified the `check_mode: no`.

yuumasato:
@matejak Great, thank you!

PR #3726, opened 2019-02-01
Update testsuite docs to show docker needs sudo

shawndwells:


jan-cerny:
No, Docker backend doesn't require sudo, but Podman backend requires sudo.

yuumasato:
if your user is part of `docker` group, you don't require sudo.

yuumasato:
@shawndwells I'll close this, and note in #3732 that you either use `sudo` or add the user to `docker` group.

And the online editing tool didn't create a branch in your fork, we should avoid using it.

shawndwells:
@jan-cerny wrt docker needing sudo.... it did on RHEL7 when @matejak was sitting next to me :)

@yuumasato works for me!

Issue #3722, opened 2019-01-28
create-stig-overlay error: TypeError: write() got an unexpected keyword argument 'pretty_print'


shawndwells:
`````
$ PYTHONPATH=`./.pyenv.sh` utils/create-stig-overlay.py --disa-xccdf=shared/references/disa-stig-rhel7-v2r2-xccdf-manual.xml --ssg-xccdf=build/ssg-rhel7-xccdf.xml -o rhel7/overlays/stig_overlay.xml 
Traceback (most recent call last):
  File "utils/create-stig-overlay.py", line 172, in <module>
    main()
  File "utils/create-stig-overlay.py", line 168, in main
    new_stig_overlay(disa_xccdftree, ssg_xccdftree, args.output_file)
  File "utils/create-stig-overlay.py", line 125, in new_stig_overlay
    xml_declaration=True)
TypeError: write() got an unexpected keyword argument 'pretty_print'
Merge state

`````

jamescassell:
this makes it really hard to diff the overlay: https://github.com/ComplianceAsCode/content/issues/2761

PR #3727, opened 2019-02-01
update stig overlay, remove pretty_print=True

shawndwells:
@redhatrises I forgot why we made this change.... do you remember? capturing via PR while our brains become unfried.


jan-cerny:
That's because `xml.etree.ElementTree.write()` method doesn't have any `pretty_print` attribute. See discussion in #3722

redhatrises:
Fixes #3722

redhatrises:
ack

PR #3729, opened 2019-02-01
Update package lists for Fedora in BUILD.md

yuumasato:
#### Description:

- Update install package command for Fedora
- Added default build command `make` to list of packages to install
- Fixed some mentions of `scap-security-guide`


jan-cerny:
Thanks!

PR #3746, opened 2019-02-08
Enable OL selinux rules

iokomin:
#### Description:

Added OL support to selinux rules set

Changes:
- Added ol7, ol8 prodtype to applicable rules
- Added OL support to selinux templates
- Fixed var_selinuxuser_execstack.var default value

#### Testing:
- checked ol8, ol7 and all build targets to pass
- checked rules content and evaluation on target system


redhatrises:
LGTM

PR #3748, opened 2019-02-11
Update crypto-policy check for nss.config

yuumasato:
#### Description:

- Change rule `configure_crypto_policy` to only check for existence of `/etc/crypto-policies/back-ends/nss.config`.
- The rule will pass if `/etc/crypto-policies/back-ends/nss.config` exists, be it a symlink or a file.

#### Rationale:
- Crypto-policy configuration for `NSS` doesn't use symlinks anymore, it copies the back-end policy file.

### DIFF
--- a/linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/oval/shared.xml
+++ b/linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/oval/shared.xml
@@ -55,11 +55,11 @@
       {{{ crypto_policy_symlink_criterion(library="java") }}}
       {{{ crypto_policy_symlink_criterion(library="krb5") }}}
       {{{ crypto_policy_symlink_criterion(library="libreswan") }}}
-      {{{ crypto_policy_symlink_criterion(library="nss") }}}
       {{{ crypto_policy_symlink_criterion(library="openssh") }}}
       {{{ crypto_policy_symlink_criterion(library="opensshserver") }}}
       {{{ crypto_policy_symlink_criterion(library="openssl") }}}
   {{% endif %}}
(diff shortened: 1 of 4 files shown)


matusmarhefka:
We should add also a negative test scenario where nss crypto policy is missing.

matusmarhefka:
I have reviewed and tested the new changes and the rule is finally passing all scenarios. Thank you @yuumasato

redhatrises:
@yuumasato question: why are we checking if `nss.config` exists in the crypto policy? Why do we care?

yuumasato:
> why are we checking if `nss.config` exists in the crypto policy? Why do we care?

@redhatrises Idea is to check whether configuration for the back-ends correspond to actual selected crypto-policy.
And this is done by checking canonical path of the symlink against expected crypto-policy.

But when there are extra configs in `/etc/crypto-policies/local.d/`, the back-end are configured with files.

Maybe the check is going too far and pushing the limits of what should be verified with regards to configuration of crypto-policies.

redhatrises:
> > why are we checking if `nss.config` exists in the crypto policy? Why do we care?
> 
> @redhatrises Idea is to check whether configuration for the back-ends correspond to actual selected crypto-policy.
> And this is done by checking canonical path of the symlink against expected crypto-policy.

@yuumasato thanks! I couldn't remember why we were checking them. These should probably be broken out into separate rules, but that is a future task.

> But when there are extra configs in `/etc/crypto-policies/local.d/`, the back-end are configured with files.
> 
> Maybe the check is going too far and pushing the limits of what should be verified with regards to configuration of crypto-policies. 

Or.... maybe it need to go further by actually checking the configurations themselves.... but that is a future endeavor as well.

redhatrises:
Since some of this discussion is about future changes and will be decided later, ACK'ing.

PR #3751, opened 2019-02-12
Added the build_product script.

matejak:
The script wipes out the build dir and re-configures and builds a single product.


matusmarhefka:
Very helpful script! Thanks for implementing it @matejak

PR #3753, opened 2019-02-12
Check for software updates in RHEL8 profiles

yuumasato:
#### Description:

- Select rule `security_patches_up_to_date` in PCI-DSS and OSPP profiles

#### Rationale:

- It would be good to have the OS updated.


redhatrises:
Ack.

PR #3755, opened 2019-02-12
Add warning about dependent rule to dconf rules.

matejak:
At the moment, dconf rules are about text config files, while dconf uses binary database as the backend by default. Extending the OVAL of other rules with the backend check is not a good solution, as it complicates tests, remediations and debugging of what went wrong.
Therefore, we thought of a compromise - let's put a warning to the rule description, which this PR does.

### DIFF
--- a/shared/macros.jinja
+++ b/shared/macros.jinja
@@ -546,3 +546,13 @@ ocil_clause: "the correct value is not returned"
         <a xmlns='http://www.w3.org/1999/xhtml' href='{{{ link }}}'>{{{ link }}}</a>
     {{%- endif %}}
 {{%- endmacro %}}
+
+
+{{% macro body_of_warning_about_dependent_rule(rule_id, why) -%}}
+        When selecting this rule in a profile, 
+        {{%- if why %}}
+            make sure that rule with ID <code>{{{ rule_id }}}</code> is selected as well: {{{ why }}}
+        {{%- else %}}
+            rule <code>{{{ rule_id }}}</code> has to be selected as well.
+        {{%- endif %}}
+{{% endmacro %}}
--- a/linux_os/guide/system/software/gnome/gnome_login_screen/dconf_gnome_disable_restart_shutdown/rule.yml
+++ b/linux_os/guide/system/software/gnome/gnome_login_screen/dconf_gnome_disable_restart_shutdown/rule.yml
@@ -28,6 +28,10 @@ rationale: |-
 
 severity: high
 
+warnings:
+    - dependency: |-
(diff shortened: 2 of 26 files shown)


shawndwells:
Seems reasonable workaround for now. Where can we follow the work to create the dconf DB OVAL probe?

redhatrises:
The only comment I have is to please, _please_ do not use jinja templates to encapsulate yaml keys like it is done here... it not only makes it extremely difficult to merge this content using tools that don't use jinja including future GUI work, but it also means that there is rework to fix this later.

matejak:
@redhatrises Thank you for your feedback, I have changed the implementation, so it now substitutes only the text of the warning.

shared/macros.jinja
```diff
@@ -546,3 +546,13 @@ ocil_clause: "the correct value is not returned"
         <a xmlns='http://www.w3.org/1999/xhtml' href='{{{ link }}}'>{{{ link }}}</a>
     {{%- endif %}}
 {{%- endmacro %}}
+
+
+{{% macro body_of_warning_about_dependent_rule(rule_id, why) -%}}
+        When selecting this rule in a profile, 
+        {{%- if why %}}
+            make sure that rule with ID {{{ weblink(link="#xccdf_org.ssgproject.content_rule_" + rule_id, text=rule_id) }}} is selected as well: {{{ why }}}
```
redhatrises: @matejak does `{{{ weblink(link="#xccdf_org.ssgproject.content_rule_" + rule_id, text=rule_id) }}}` work for both Datastream and XCCDF?

matejak: I suppose so, what would be the catch? It works for datastream-powered report for sure.

yuumasato: I think the catch is that in a pure XCCDF Benchmark, the rules don't have the prefix `xccdf_org.ssgproject.content_rule_`, they are just `rule_id`.

matejak: So what is better? 
- Alink that does nothing, but that is correct in the recommended use case, or 
- no link at all, only text short rule ID?

redhatrises: I would probably say a short rule ID because you would have a bug in the XCCDF-only content if you use a link.

yuumasato: Yeah, in the past we have gone with short rule ID, see https://github.com/ComplianceAsCode/content/pull/3113.

matejak: OK then - fixed by rebase.

redhatrises:
LGTM. Thanks for your PR @matejak!

PR #3757, opened 2019-02-13
Remove SSH option 'GSSAPIAuthentication=no' from the test suite

matusmarhefka:
#### Description:

- The `GSSAPIAuthentication=no` ssh option set by the test suite is causing that in some cases it is needed to enter a password when accessing a VM.


yuumasato:
LGTM, Ack.

PR #3759, opened 2019-02-13
ensure yum repos have valid section header

jgwl:
#### Description:
- This adds a rule to check for a [header] in the file before requiring that gpgcheck=1 is in existence.

#### Rationale:

- Files under yum.repos.d may not have any content (e.g., redhat.repo can be empty), so adding an orphaned gpgcheck=1 will break yum and subsequent tasks.


shawndwells:
@openscap-ci ok to test

redhatrises:
Ack! Thanks for the contribution @jgwl!

PR #3760, opened 2019-02-13
Fix dracut-fips check indentation

jgwl:
#### Description:

- Removes incorrect indention on when conditional.

#### Rationale:

- Indention for the when conditional was incorrect, causing this task to fail.


redhatrises:
@openscap-ci add to whitelist

redhatrises:
Ack when jenkins finishes

PR #3762, opened 2019-02-13
Check existence of pkcs11-switch

jgwl:
#### Description:

- Check existence of pkcs11-switch

#### Rationale:

- If system does not have pkcs11-switch, it cannot run the preceding tasks which rely on that file. This checks for the existence first.


yuumasato:
Thanks!

PR #3767, opened 2019-02-15
Extend list of noshells - add /bin/false

yuumasato:
#### Description:

- Consider users with shell /bin/false to be users without shell

#### Rationale:

- Considered an old practice, but the check should accept `/bin/false` as non-shell.
- Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1676661

Fixes: #3768


matejak:
Thank you for the PR, this is a no-brainer merge.

shawndwells:
Believe this needs to be reverted. ``/bin/false`` is an unacceptable shell from a security perspective and was purposefully not included. Good writeup on reasoning:

https://www.pixelstech.net/article/1328532389-SSH-Security-and-You---bin-false-is-%2Anot%2A-security

PR #3770, opened 2019-02-15
Revert PR #3767

redhatrises:
- Using outdated methods is a bad security practice. In addition, using
  /bin/false is considered unacceptable for C2S and DISA STIG. There is also
  a logging issue with using /bin/false.


yuumasato:
@redhatrises What to you think of adding justification to rule?

redhatrises:
@yuumasato it is already not mentioned as an acceptable option in the STIG and CIS benchmarks. Also, we need to add an article in the downstream JIRA that when an OpenSCAP scan flags an issue with another product or rpm that it is usually an issue with that product/RPM and not with SSG.

Issue #3735, opened 2019-02-05
cmake error with make zipfile


afcollins:
#### Description of problem:
When I try to build the content using the zipfile target, there is an invalid cmake command being generated.

#### SCAP Security Guide Version:
Building from source:
```
[root@dns content]# git log -1
commit b7b54d27e4c5797ad741c6a2ef73612159c566c9
Merge: bec53b9 f8fe170
Author: Jan Černý <jcerny@redhat.com>
Date:   Mon Feb 4 13:21:29 2019 +0100

    Merge pull request #3731 from matejak/ssgts_libvirt_fix

    Test suite libvirt fix and Dockerfiles refresh.
```
#### Operating System Version:
```
# uname -a
rhel7
Linux dns 3.10.0-862.11.6.el7.x86_64 #1 SMP Fri Aug 10 16:55:11 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

#### Steps to Reproduce:
1. make zipfile

#### Actual Results:
```
[root@dns build]# make zipfile/fast
[100%] Building zipfile at /root/content/build/zipfile/scap-security-guide-0.1.43.zip
remove_dir
make_dir
copy license
mkdir
CMake Error: cmake version 2.8.12.2
Usage: /usr/bin/cmake -E [command] [arguments ...]
Available commands: 
```

#### Expected Results:
To complete successfully.

#### Addition Information/Debugging Steps:
Added echo statements to the generated CMakeFiles/zipfile.dir/build.make (between lines 54 and 66), and the offending cmake statement is:
` 62         /usr/bin/cmake -E copy /root/content/rhel{6,7}/kickstart/*-ks.cfg zipfile/scap-security-guide    -0.1.43/kickstart
`

afcollins:
I found the offending line in CMakeFiles/zipfile.dir/build.make, but do not immediately know where the source is to generate this line.  Looks like the problem is rhel{6,7} resolving to multiple source arguments, and cmake only accepts one?

redhatrises:
@afcollins any reason as to why you are not getting the built release zip file at https://github.com/ComplianceAsCode/content/releases/download/v0.1.42/scap-security-guide-0.1.42.zip?

redhatrises:
Looks like `cmake -E copy` doesn't support wildcards in 2.8.12, so `cmake/SSGCommon.cmake` needs to be fixed.

Offending lines:
```
[...]

PR #3771, opened 2019-02-15
Possible fix for #3735

redhatrises:
Possible fix for #3735. Really open to suggestions here as I can reproduce #3735 on RHEL7, but I can't consistently if at all reproduce this on Fedora.


yuumasato:
>  but I can't consistently if at all reproduce this on Fedora.

This doesn't seem to be an issue in Fedora, I have cmake 3.12.1 installed and it works.
Although, documentation says wildcards are not supported: https://cmake.org/cmake/help/v3.12/manual/cmake.1.html#command-line-tool-mode

redhatrises:
bump to anyone... need to know what we should do here.

yuumasato:
After playing with cmake, I'm convinced that the problem is that cmake 2.8.12 doesn't support copying multiple files at once.
Globbing is not the problem, when a glob results in a single file, the copy command works.

I believe this PR is the best way to move forward and support cmake 2.8.12.

cmake/SSGCommon.cmake
```diff
@@ -1107,7 +1107,7 @@ macro(ssg_build_zipfile ZIPNAME)
         COMMAND ${CMAKE_COMMAND} -E copy "${CMAKE_SOURCE_DIR}/Contributors.md" "zipfile/${ZIPNAME}"
         COMMAND ${CMAKE_COMMAND} -E copy "${CMAKE_SOURCE_DIR}/LICENSE" "zipfile/${ZIPNAME}"
         COMMAND ${CMAKE_COMMAND} -E make_directory "zipfile/${ZIPNAME}/kickstart"
-        COMMAND ${CMAKE_COMMAND} -E copy "${CMAKE_SOURCE_DIR}/rhel{6,7}/kickstart/*-ks.cfg" "zipfile/${ZIPNAME}/kickstart"
+        COMMAND ${CMAKE_COMMAND} -DFILES="${CMAKE_SOURCE_DIR}/rhel*/kickstart/*-ks.cfg" -DPATH="zipfile/${ZIPNAME}/kickstart" -P "${CMAKE_SOURCE_DIR}/cmake/CopyFiles.cmake"
```
yuumasato: I suggest to follow same convention as `cp` command.
So rename `FILES` to `SOURCE`, and `PATH` to `DEST`.

yuumasato:
I wish we could get rid of cmake 2.8.12....

yuumasato:
@redhatrises I will take the liberty of pushing changes to this branch, :)

yuumasato:
Rebase on latest master. Also:
- Extended use of `CopyFiles` cmake script
- Fixed zipfile target dependency on guides and tables
- Updated `nist-zipfile` target

redhatrises:
> @redhatrises I will take the liberty of pushing changes to this branch, :)

Thanks @yuumasato! Didn't expect you to do that! much appreciated.

PR #3794, opened 2019-02-18
Update contributors list for v0.1.43

yuumasato:
#### Description:

- Update contributors list

PR #3798, opened 2019-02-18
Map lkinser to Lee Kinser in ssg/contributors.py

redhatrises:
- Part of #3794


shawndwells:
Well yah, i suppose this would make that change persistent. Thanks!

PR #3799, opened 2019-02-18
Enable RHV4 product building

redhatrises:


redhatrises:
> Should the CPE be `installed_OS_is_rhv4` or `rhvh4`?

@shawndwells should be `installed_OS_is_rhv4`

PR #3800, opened 2019-02-19
Linearize rules

matejak:
This PR extends capabilities of the `yaml_to_shorthand.py` script to save rules into product directories after template expansion (e.g. into `build/rhel7/rules/...`).

Why:
* To move the intermediate representation away from XML-based,
* To be able to work with rule definitions during the build process easily.
* To debug template/yaml format-related issues with content input - the yaml in the build tree is result of all template expansions yaml interpretations.

As we use templating in yaml source, it may be that individual definitions may vary between products, so this has to be done for each product separately.

As a side-task, the build functionality has been refactored in order to reduce the complexity of the code - adding more `if`s would make it a huge mess.

### DIFF
--- a/cmake/SSGCommon.cmake
+++ b/cmake/SSGCommon.cmake
@@ -97,8 +97,8 @@ macro(ssg_build_shorthand_xml PRODUCT)
     string(REPLACE "\n" ";" SHORTHAND_INPUTS "${SHORTHAND_INPUTS_STR}")
 
     add_custom_command(
-        OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/shorthand.xml"
-        COMMAND env "PYTHONPATH=$ENV{PYTHONPATH}" "${PYTHON_EXECUTABLE}" "${SSG_BUILD_SCRIPTS}/yaml_to_shorthand.py" --build-config-yaml "${CMAKE_BINARY_DIR}/build_config.yml" --product-yaml "${CMAKE_CURRENT_SOURCE_DIR}/product.yml" --bash_remediation_fns "${CMAKE_BINARY_DIR}/bash-remediation-functions.xml" --output "${CMAKE_CURRENT_BINARY_DIR}/shorthand.xml" build
+        OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/shorthand.xml" "${CMAKE_CURRENT_BINARY_DIR}/rules"
+        COMMAND env "PYTHONPATH=$ENV{PYTHONPATH}" "${PYTHON_EXECUTABLE}" "${SSG_BUILD_SCRIPTS}/yaml_to_shorthand.py" --resolve-rules-into "${CMAKE_CURRENT_BINARY_DIR}/rules" --build-config-yaml "${CMAKE_BINARY_DIR}/build_config.yml" --product-yaml "${CMAKE_CURRENT_SOURCE_DIR}/product.yml" --bash_remediation_fns "${CMAKE_BINARY_DIR}/bash-remediation-functions.xml" --output "${CMAKE_CURRENT_BINARY_DIR}/shorthand.xml" build
         COMMAND "${XMLLINT_EXECUTABLE}" --format --output "${CMAKE_CURRENT_BINARY_DIR}/shorthand.xml" "${CMAKE_CURRENT_BINARY_DIR}/shorthand.xml"
         DEPENDS ${SHORTHAND_INPUTS}
         DEPENDS generate-internal-bash-remediation-functions.xml
--- a/ssg/build_yaml.py
+++ b/ssg/build_yaml.py
@@ -486,6 +486,21 @@ class Group(object):
 class Rule(object):
     """Represents XCCDF Rule
     """
+    YAML_KEYS_DEFAULTS = {
+        "prodtype": lambda: "all",
+        "title": lambda: RuntimeError("Missing key 'title'"),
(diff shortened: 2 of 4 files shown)


ssg/build_yaml.py
```diff
@@ -539,6 +541,25 @@ def from_yaml(yaml_file, env_yaml=None):
         rule.validate_references(yaml_file)
         return rule
 
+    def to_contents_dict(self):
+        """
+        Returns a dictionary that is the same schema as the dict obtained when loading rule YAML.
+        """
+        interesting_keys = [
```
redhatrises: @matejak can this be automated rather than being explicit?

redhatrises:
> Why? As we don't like our intermediate representation to be XML-based, we want to have an easy way to look at rule definition during the build process. As we use templating in yaml source, it may be that individual definitions may vary between products, so this has to be done for each product.

@matejak would you also make the argument that it will help us identify issues with rules and rules templates during the build process?

matejak:
@redhatrises I have updated the PR description, you are right about that aspect.

ssg/build_yaml.py
matejak: I have moved to semi-automated, I am somehow not in favor of automated, as they say that [explicit is better than implicit](https://www.python.org/dev/peps/pep-0020/).

ssg/build_yaml.py
```diff
@@ -539,6 +547,30 @@ def from_yaml(yaml_file, env_yaml=None):
...
+    def to_contents_dict(self):
+        """
+        Returns a dictionary that is the same schema as the dict obtained when loading rule YAML.
+        """
+
+        yaml_contents = dict()
+        for key in Rule.YAML_KEYS_DEFAULTS:
+            yaml_contents[key] = getattr(self, key)
+
+        yaml_contents["documentation_complete"] = True
```
yuumasato: Will this value be used later?
I think at this point, all rules are supposed to be picked up for build.

ssg/build_yaml.py
```diff
@@ -645,107 +677,166 @@ def to_file(self, file_name):
...
+        self.values = []
+        self.subdirectories = []
+
+        self.profiles_dir = profiles_dir
+        self.bash_remediation_fns = bash_remediation_fns
+        self.env_yaml = env_yaml
+
+        self.parent_group = None
+
+    def _examine_directory(self, guide_directory):
```
yuumasato: What is the directory being examined for?
I would say pre-loading.

ssg/build_yaml.py
matejak: I was not sure about that. I guess that it can be safely removed.

ssg/build_yaml.py
matejak: preloading is also abstract and, in a way, fuzzy - but what about collecting?

matejak:
@yuumasato Rebased. Linearizing is not a bad term - it is putting into the line, i.e. next to each other vs keeping it in tree-like structure.

ssg/build_yaml.py
yuumasato: Collect seems good.

cmake/SSGCommon.cmake
```diff
@@ -97,8 +97,8 @@ macro(ssg_build_shorthand_xml PRODUCT)
     string(REPLACE "\n" ";" SHORTHAND_INPUTS "${SHORTHAND_INPUTS_STR}")
 
     add_custom_command(
-        OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/shorthand.xml"
-        COMMAND env "PYTHONPATH=$ENV{PYTHONPATH}" "${PYTHON_EXECUTABLE}" "${SSG_BUILD_SCRIPTS}/yaml_to_shorthand.py" --build-config-yaml "${CMAKE_BINARY_DIR}/build_config.yml" --product-yaml "${CMAKE_CURRENT_SOURCE_DIR}/product.yml" --bash_remediation_fns "${CMAKE_BINARY_DIR}/bash-remediation-functions.xml" --output "${CMAKE_CURRENT_BINARY_DIR}/shorthand.xml" build
+        OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/shorthand.xml" "${CMAKE_CURRENT_BINARY_DIR}/rules"
+	COMMAND env "PYTHONPATH=$ENV{PYTHONPATH}" "${PYTHON_EXECUTABLE}" "${SSG_BUILD_SCRIPTS}/yaml_to_shorthand.py" --linearize-rules-into "${CMAKE_CURRENT_BINARY_DIR}/rules" --build-config-yaml "${CMAKE_BINARY_DIR}/build_config.yml" --product-yaml "${CMAKE_CURRENT_SOURCE_DIR}/product.yml" --bash_remediation_fns "${CMAKE_BINARY_DIR}/bash-remediation-functions.xml" --output "${CMAKE_CURRENT_BINARY_DIR}/shorthand.xml" build
```
jan-cerny: indents

yuumasato:
> next to each other vs keeping it in tree-like structure

It make sense to me now. But besides that you are also changing the contents of the rule.
As I read from description of PR, important part is to fill in the "gaps" in rule description for each product, with goal of establishing new intermediate format, and debugging purposes.

matejak:
OK, I am changing that to `resolve`, like "resolve into a directory" and I have also fixed indents by a rebase.

yuumasato:
This is a good step forward, thank you @matejak.

PR #3801, opened 2019-02-19
Splits combine_remediations.py into 2 scripts

jan-cerny:
#### Description:
Splits combine_remediations.py into 2 scripts.

First script (combine_remediations.py) gathers all remediation snippets, resolves Jinja macros and outputs all of the snippets to a single directory, currently set to CMake as `build/<product>/all_fixes/<language>`.

Second script (generate_fixes_xml.py) merges all the remediation snippets from this directory into a single XML file `<language>-fixes.xml`.  The scripts are called by CMake during the build and generate_fixes_xml.py depends on outputs of combine_remediations.py.

#### Rationale:
We will use this when we will generate Ansible playbook for each rule for each profile.

### DIFF
--- a/cmake/SSGCommon.cmake
+++ b/cmake/SSGCommon.cmake
@@ -241,7 +241,7 @@ macro(_ssg_build_remediations_for_language PRODUCT LANGUAGES)
 
       add_custom_command(
           OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/${LANGUAGE}-fixes.xml"
-          COMMAND env "PYTHONPATH=$ENV{PYTHONPATH}" "${PYTHON_EXECUTABLE}" "${SSG_BUILD_SCRIPTS}/combine_remediations.py" --build-config-yaml "${CMAKE_BINARY_DIR}/build_config.yml" --product-yaml "${CMAKE_CURRENT_SOURCE_DIR}/product.yml" --remediation_type "${LANGUAGE}" --build_dir "${CMAKE_BINARY_DIR}" --output "${CMAKE_CURRENT_BINARY_DIR}/${LANGUAGE}-fixes.xml" "${BUILD_REMEDIATIONS_DIR}/shared/${LANGUAGE}" "${SSG_SHARED}/fixes/${LANGUAGE}" "${BUILD_REMEDIATIONS_DIR}/${LANGUAGE}" "${CMAKE_CURRENT_SOURCE_DIR}/fixes/${LANGUAGE}"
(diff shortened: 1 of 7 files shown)


cmake/SSGCommon.cmake
```diff
@@ -238,16 +238,29 @@ macro(_ssg_build_remediations_for_language PRODUCT LANGUAGES)
     foreach(LANGUAGE ${LANGUAGES})
       file(GLOB EXTRA_LANGUAGE_DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/fixes/${LANGUAGE}/*")
       file(GLOB EXTRA_SHARED_LANGUAGE_DEPENDS "${SSG_SHARED}/fixes/${LANGUAGE}/*")
+      set(ALL_FIXES_DIR "${CMAKE_CURRENT_BINARY_DIR}/all_fixes/${LANGUAGE}")
```
redhatrises: Can we not just put all remediation scripts in the `fixes` directory rather than just having `all_fixes` and `fixes`?

jan-cerny: Yes, I agree, I also don't like it.

jan-cerny:
@redhatrises 

> Can we not just put all remediation scripts in the `fixes` directory rather than just having `all_fixes` and `fixes`?

It's more complicated, we need a cleanup. See https://github.com/ComplianceAsCode/content/issues/3816 for more details. 

I have created a temporary solution. First, the `generate_from_templates.py` will create a directory `fixes_from_templates`. Then, `combine_remediations.py` will create `fixes` directory, which will contain all the fixes (both explicit and generated).

ssg/build_remediations.py
```diff
@@ -166,22 +166,10 @@ def get_populate_replacement(remediation_type, text):
...
-    we need a env_yaml context to process these. In practice, no remediations
-    use jinja in the configuration, so for extracting only the configuration,
-    env_yaml can be an abritrary product.yml dictionary.
-
-    If the logic of configuration parsing changes significantly, please also
-    update ssg.fixes.parse_platform(...).
-    """
-
+def _parse_remediation(fix_file_lines):
     mod_file = []
```
matejak: `mod_file` is a bad name. What comes to my first is that `remediation_content` is better.

ssg/build_remediations.py
```diff
@@ -166,22 +166,10 @@ def get_populate_replacement(remediation_type, text):
...
-    Parses a remediation from a file. As remediations contain jinja macros,
-    we need a env_yaml context to process these. In practice, no remediations
-    use jinja in the configuration, so for extracting only the configuration,
-    env_yaml can be an abritrary product.yml dictionary.
-
-    If the logic of configuration parsing changes significantly, please also
-    update ssg.fixes.parse_platform(...).
-    """
-
+def _parse_remediation(fix_file_lines):
```
matejak: This function splits remediation content and metadata, which are parsed into a dictionary. Therefore, I would suggest to be more concrete and consider naming e.g. `split_remediation_content_and_metadata` or something like that.

ssg/build_remediations.py
```diff
@@ -204,6 +192,32 @@ def parse_from_file(file_path, env_yaml):
     return remediation(mod_file, config)
 
 
+def parse_from_file(file_path, env_yaml):
```
matejak: Obviously, the naming suggestion apply here as well.

A Remediation class would solve the situation and avoid too long names, as it would read files either with or without jinja expansion, and it would have methods for metadata extraction.

ssg/build_remediations.py
```diff
@@ -259,6 +273,26 @@ def write_fixes(remediation_type, build_dir, output_path, fixes):
...
+    Writes fixes as files to output_dir, each fix as a separate file
+    """
+    try:
+        extension = REMEDIATION_TO_EXT_MAP[remediation_type]
+    except KeyError:
+        raise ValueError("Unknown remediation type %s." % remediation_type)
+
+    if not os.path.exists(output_dir):
+        os.makedirs(output_dir)
+    for fix_name in fixes:
```
matejak: You can iterate over `.items()` and have `for fix_name, fix in fixes.items(): ...`, allowing you to avoid using `fix_contents` and `config`.

jan-cerny:
@matejak I have renamed variable `mod_file` to `remediation_contents`, renamed function `_parse_remediation` to `_split_remediation_content_and_metadata`, renamed function `parse_from_file` to `parse_from_file_with_jinja` and I have iterated dictionary intelligently.

matejak:
Oh, I thought that I have merged it already. Great job anyway!

PR #3802, opened 2019-02-19
Move unit tests

matejak:
Some unit-tests were ignored during the test phase, because they were in a wrong directory.

This PR moves them, and fixes one failing test.


tests/unit/ssg-module/test_rule_yaml.py
```diff
@@ -1,6 +1,7 @@
 import pytest
 
 import os
+import tempfile
```
redhatrises: @matejak did you mean to add this import?

matejak: Not really, good catch!

redhatrises:
LGTM. Ack.

Issue #3245, opened 2018-09-03
Remove leftovers from shared/checks/oval


jan-cerny:
#### Description of problem:

We have moved everything under linux_os directory. However, there is still a bunch of files in shared/checks/oval directory.

I think it will be really confusing for a newcomer if  they will see this directory. Then it will not be clear for them where they should contribute their OVALs.

These files should be removed or reduplicated.

#### SCAP Security Guide Version:
current master, as of 3rd September 2018

#### Operating System Version:
all

#### Steps to Reproduce:

1. cd shared/checks/oval
2. ls
3.
4.

#### Actual Results:
Many files are in shared/checks/oval

#### Expected Results:
No files are in shared/checks/oval or only meaningful files are in shared/checks/oval

cipherboy:
These should also be audited in the product checks directory `<product>/checks/oval`, as I do recall a few not being moved.

PR #3812, opened 2019-02-20
Moved contents of shared/fixes to their corresponding rules.

matejak:
This PR moves bash and ansible remediations to relevant rules.

This is related to #3245, as it empties the `shared/fixes` directory.


redhatrises:
LGTM

Issue #3816, opened 2019-02-21
Clean up remediation directories


jan-cerny:
#### Description of problem:
There are multiple sources from which the remediations are merged into the final content:
1. templates
2. shared templates
3. rule directories
4. `/<product>/fixes` directory
5. `/shared/fixes` directory

The fact that there are 5 possible sources makes the build process unnecessary complex. It is also confusing for contributors.

I believe that `/<product>/fixes` directories should be removed and they should not be used at all. We can move the files from `/<product>/fixes` to directories for the given rules. Most of the products (eg. fedora, rhel7) already don't have `/<product>/fixes`. Other products contain only a small amount of remediations there.

I believe that  `/shared/fixes` directory should also be removed and the remediations should be moved to rule directories in `linux_os/guide`.   The `/shared/fixes` contains only 1 Ansible remediation and 4 Bash intermediations.

After `/<product>/fixes` and `/shared/fixes` will be removed, there will be only 3 sources left, which IMO makes the system less confusing.

#### SCAP Security Guide Version:
current upstream master as of 21st February 2018

#### Operating System Version:
all

#### Steps to Reproduce:

1. check code of `_ssg_build_remediations_for_language` in  `cmake/SSGCommon.cmake`
2. examine which directories and paths are used when calling `generate_from_templates.py` and `combine_remediations.py`
3. check also the dependencies of the targets and command defined in the aforementioned macro

#### Actual Results:
the build system uses `/<product>/fixes` and `/shared/fixes`

#### Expected Results:
1. The build system doesn't use `./<product>/fixes` and `./shared/fixes`.
2. The `_ssg_build_remediations_for_language` macro in `cmake/SSGCommon.cmake` is simple

jan-cerny:
Second part is addressed in https://github.com/ComplianceAsCode/content/pull/3812

jan-cerny:
I think only thing that remains there is to meet
> The _ssg_build_remediations_for_language macro in cmake/SSGCommon.cmake is simple

PR #3817, opened 2019-02-21
Get rid of  /<product>/fixes directories

jan-cerny:
#### Description:
Removes  `/<product>/fixes` directory. They were mostly empty, apart from one remediation in RHOSP13, but that doesn't belong to any rule, so we remove it as well.

#### Rationale:
It's confusing to have multiple sources for remediations. See #3816 for more context and explanation.


yuumasato:
LGTM, the mentioned remediation is added in https://github.com/ComplianceAsCode/content/commit/18cae48e4291a2b2ff0c3e4e4f8fe561125079cc but later seems to be dropped.

Issue #3044, opened 2018-06-25
is_valid_cce should check check-digit


cipherboy:
Per requirement A17 [here](http://people.redhat.com/swells/nist-scap-validation/scap-val-requirements-1.2.html), the last digit in CCE identifiers is a "check-digit". However, `is_valid_cce` [does not](https://github.com/OpenSCAP/scap-security-guide/blob/master/ssg/checks.py#L44-L53) validate these digits. 

It isn't immediately clear how to validate these check digits. Feedback is necessary before continuing.

cipherboy:
I agree with your description of what it is currently doing.

This bug was about extending the validations to also verify the check
digit.

Just like is_valid_cce currently prevents labels like "TBD", we could have
it prevent off-by-one typos like 1234-5 vs 1234-6. See the link about
requirement A17:

 A17 N/A This is an additional, common-sense check. CCE number is in an
invalid format or the check-digit does not match. It should be of format
CCE-XXXX-X or CCE-XXXXX-X where each X is a digit, and the final X is a
check-digit.

This is something that we might need to ask NIST about. At the time I had
not seen the correct document describing the computation of the check
digit.


If this isn't desired, we can close the issue.

mpreisler:
IMO it is desired, this would be a nice feature. I am all for keeping it open.

redhatrises:
funny enough that is what they test for.
```
test="if(@system eq 'http://cce.mitre.org' or @system eq 'CCE') then matches(., '^CCE-\d+-\d$')
```
But maybe this is the code that you are looking for to validate the checksum?
```
test="if((@system eq 'http://cce.mitre.org' or @system eq 'CCE') and matches(., '^CCE-\d+-\d$')) 
then (sum(for $j in (for $i in reverse(string-to-codepoints(concat(substring(.,5,string-length(.)-6),substring(.,string-length(.),1))))[position() mod 2 = 0] return ($i - 48) * 2, 
for $i in reverse(string-to-codepoints(concat(substring(.,5,string-length(.)-6),substring(.,string-length(.),1))))[position() mod 2 = 1] return ($i - 48)) return ($j mod 10, $j idiv 10)) mod 10) eq 0 else true()"
```

cipherboy:
Making @redhatrises's comment a little more readable for the future....

```
test="
if (
  (@system eq 'http://cce.mitre.org' or @system eq 'CCE')
  and
  matches(., '^CCE-\d+-\d$')
) then (
  sum (
[...]

PR #3819, opened 2019-02-21
Add extended CCE validation per XPath definition

cipherboy:
#### Description:

Per the [specification](https://github.com/ComplianceAsCode/content/issues/3044#issuecomment-419526562) given by @redhatrises, adds extended validation to our CCE identifiers. 

As a result of this however, some CCEs in our files are not valid, either in this implementation or in the XPath canonical implementation. This cannot be merged until these are fixed.

In particular, I found this one:

```
ValueError: CCE Identifiers must be valid: value 80335-1 for cce cce@rhel7 in file /home/cipherboy/GitHub/cipherboy/scap-security-guide/fedora/../linux_os/guide/system/network/network-ipv6/configuring_ipv6/disabling_ipv6_autoconfig/sysctl_net_ipv6_conf_default_accept_source_route/rule.yml
```

Testing it [here](http://xpather.com/) with the following code gives a return value of 8:
```XPath
  sum (
    for $j in (
      for $i in reverse(
        string-to-codepoints(
          concat(
            substring("CCE-80335-1", 5, string-length("CCE-80335-1")-6),
            substring("CCE-80335-1", string-length("CCE-80335-1"), 1)
          )
        )
      )[position() mod 2 = 0] return ($i - 48) * 2,
      for $i in reverse(
        string-to-codepoints(
          concat(
            substring("CCE-80335-1", 5, string-length("CCE-80335-1")-6),
            substring("CCE-80335-1", string-length("CCE-80335-1"), 1)
          )
        )
      )[position() mod 2 = 1] return ($i - 48)
    ) return ($j mod 10, $j idiv 10)
  ) mod 10
```

#### Rationale:

- Better ensures integrity and correctness of our CCE identifiers. 

Resolves: #3044

### DIFF
--- a/ssg/build_renumber.py
+++ b/ssg/build_renumber.py
@@ -425,7 +425,7 @@ def verify_correct_form_of_referenced_cce_identifiers(xccdftree):
         identcce = _find_identcce(rule)
         if identcce is not None:
             cceid = identcce.text
-            if not is_cce_valid(cceid):
+            if not is_cce_format_valid(cceid):
                 print("Warning: CCE '{0}' is invalid for rule '{1}'. Removing CCE..."
                       .format(cceid, rule.get("id"), file=sys.stderr))
                 rule.remove(identcce)
--- a/ssg/build_renumber.py
+++ b/ssg/build_renumber.py
@@ -9,7 +9,7 @@ from .parse_oval import resolve_definition, find_extending_defs, get_container_g
 from .xml import parse_file, map_elements_to_their_ids
 
 
-from .checks import get_content_ref_if_exists_and_not_remote, is_cce_valid
+from .checks import get_content_ref_if_exists_and_not_remote, is_cce_value_valid, is_cce_format_valid
 from .utils import SSGError
 from .xml import ElementTree as ET
 oval_ns = oval_namespace
@@ -177,7 +177,7 @@ class OVALFileLinker(FileLinker):
                 "OVAL rule '{0}' doesn't have a description, which is mandatory".format(ovalid)
 
             xccdfcceid = idmappingdict[ovalid]
-            if is_cce_valid(xccdfcceid):
+            if is_cce_format_valid(xccdfcceid) and is_cce_value_valid(xccdfcceid):
                 # Then append the <reference source="CCE" ref_id="CCE-ID" /> element right
                 # after <description> element of specific OVAL check
                 ccerefelem = ET.Element('reference', ref_id=xccdfcceid,
--- a/ssg/build_yaml.py
+++ b/ssg/build_yaml.py
@@ -586,7 +586,7 @@ class Rule(object):
                     raise ValueError("CCE Identifier format must be valid: invalid format '%s' for CEE '%s'"
                                      " in file '%s'" % (ident_val, ident_type, yaml_file))
                 if not is_cce_value_valid("CCE-" + ident_val):
-                    raise ValueError("CCE Identifier value must be valid: invalid value '%s' for CEE '%s'"
+                    raise ValueError("CCE Identifier value is not a valid checksum: invalid value '%s' for CEE '%s'"
                                      " in file '%s'" % (ident_val, ident_type, yaml_file))
 
     def validate_references(self, yaml_file):
--- a/ssg/build_yaml.py
+++ b/ssg/build_yaml.py
@@ -12,7 +12,7 @@ from .constants import XCCDF_PLATFORM_TO_CPE
 from .constants import PRODUCT_TO_CPE_MAPPING
 from .rules import get_rule_dir_id, get_rule_dir_yaml, is_rule_dir
 
-from .checks import is_cce_valid
+from .checks import is_cce_format_valid, is_cce_value_valid
 from .yaml import open_and_expand, open_and_macro_expand
 from .utils import required_key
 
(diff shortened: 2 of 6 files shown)


ssg/checks.py
```diff
@@ -54,7 +54,33 @@ def is_cce_valid(cceid):
     http://people.redhat.com/swells/nist-scap-validation/scap-val-requirements-1.2.html
     """
     match = re.match(r'^CCE-\d{4,5}-\d$', cceid)
-    return match is not None
+    if match is None:
```
redhatrises: @cipherboy should we have this check in a separate function so that we can have better error messages as to why the cce check failed?

cipherboy: The error message is raised at a higher level than this function, e.g., [here](https://github.com/openscap/scap-security-guide/blob/master/ssg/build_yaml.py#L555). We'd have to turn it into two function calls and update it in every location (`is_cce_valid_format` + `is_cce_valid_contents` or some such), with different error messages for each. 

It is only used in two places, one of which errors (`build_yaml.py` line 555), the other just prints a warning (`build_renumber.py` line 428).

redhatrises: I think we should update the error messages; otherwise, it is going to be cryptic as to why it failed. Another idea is maybe return the errors or raise the errors in the function itself?

redhatrises:
@cipherboy #3820 should fix some of the identified wrong CCEs.

redhatrises:
Taking this PR out of WIP. #3820 needs to be merged first.

ssg/build_yaml.py
```diff
@@ -552,9 +552,12 @@ def validate_identifiers(self, yaml_file):
...
                                  % (ident_type, yaml_file))
             if ident_type[0:3] == 'cce':
-                if not is_cce_valid("CCE-" + ident_val):
-                    raise ValueError("CCE Identifiers must be valid: value %s for cce %s"
-                                     " in file %s" % (ident_val, ident_type, yaml_file))
+                if not is_cce_format_valid("CCE-" + ident_val):
+                    raise ValueError("CCE Identifier format must be valid: invalid format '%s' for CEE '%s'"
+                                     " in file '%s'" % (ident_val, ident_type, yaml_file))
+                if not is_cce_value_valid("CCE-" + ident_val):
+                    raise ValueError("CCE Identifier value must be valid: invalid value '%s' for CEE '%s'"
```
cipherboy: If we want to, could say "CCE doesn't pass checksum validations" or something. Not sure if its worth changing though.

ssg/build_renumber.py
```diff
@@ -177,7 +177,7 @@ def _add_cce_id_refs_to_oval_checks(self, idmappingdict):
                 "OVAL rule '{0}' doesn't have a description, which is mandatory".format(ovalid)
 
             xccdfcceid = idmappingdict[ovalid]
-            if is_cce_valid(xccdfcceid):
+            if is_cce_format_valid(xccdfcceid) and is_cce_value_valid(xccdfcceid):
```
cipherboy: Did we want to provide a `is_cce_valid` wrapper for this along with the test cases?

Else, `tests/unit/ssg-module/test_checks.py` needs to be updated too.

cipherboy:
I validated this on the list of available CCEs and it looks good:

```python
import ssg.checks
cces = open("./shared/references/cce-rhel-avail.txt", 'r').read().split("\n")[:-1]

for cce in cces:
    if not ssg.checks.is_cce_valid(cce):
        print(cce)
```

ssg/build_yaml.py
redhatrises: @cipherboy updated the error message.

ssg/build_renumber.py
redhatrises: Updated the tests.

tests/unit/ssg-module/test_checks.py
```diff
@@ -32,14 +32,23 @@ def test_get_oval_path():
...
-    assert not icv("12345-6")
     assert not icv("TBD")
     assert not icv("CCE-TBD")
     assert not icv("CCE-abcde-f")
+
+
+def test_is_cce_value_valid():
+    icv = ssg.checks.is_cce_value_valid
+    assert icv("CCE-27191-6")
+    assert icv("CCE-7223-7")
```
cipherboy: This needed to be `CCE-27223-7` like in #3820, so I pushed an update.

redhatrises: Thanks @cipherboy

redhatrises:
@cipherboy can you rebase against the master?

redhatrises:
@cipherboy nevermind.... I rebased. Sorry for the noise.

redhatrises:
It's passing Jenkins now!!!!!! HHHHoooooorrrraaayyyyyyy

ssg/checks.py
```diff
@@ -57,6 +57,33 @@ def is_cce_valid(cceid):
...
+    odds = digits[0::2]
+
+    # XPath flattens the for loop's return into a single list; do this in
+    # via a double sum.
+    tuples = map(lambda j: [j % 10, j // 10], evens + odds)
+    value = sum(sum(tuples, [])) % 10
+
+    # Valid CCE <=> value == 0
+    return value == 0
+
```
yuumasato: This looks more complex than necessary. 
This is the check algorithm, https://en.wikipedia.org/wiki/Luhn_algorithm.
Is there a special reason to do it exactly like it is done in XPath script?

redhatrises: Yes, because it is how NIST does it in the validation program, which in this case, is the official documentation of how to calculate and validate CCE. 

Have you been able to find where the NIST defines how to calculate CCE such as explicit use of the Luhn algorithm?

redhatrises: OMG.... I actually found the reference page for CCE that does specify Luhn.... @yuumasato you are amazing to find that! Super kudos to you.

@cipherboy can you update the PR to use https://en.wikipedia.org/wiki/Luhn_algorithm ?

yuumasato: Had fond this: https://cce.mitre.org/lists/entries.html
But looking under NIST, found this: https://csrc.nist.gov/Projects/Security-Content-Automation-Protocol/faqs 
Look in " What is the format of the CCE Identifier (CCE-ID) number?"

redhatrises: @yuumasato yeah... I had no idea why I couldn't find it. Great job!

yuumasato: It was not easy to find it, :P

yuumasato:
Awesome!
Thank you @cipherboy and @redhatrises for looking into this.

PR #3820, opened 2019-02-21
Fix invalid CCEs

redhatrises:
- Found a couple of CCEs that were typo'd. All thanks goes to @cipherboy for his work!


shawndwells:
For the IPv6 one, @redhatrises went through the git history and found the typo in the git logs. Crazy.

PR #3824, opened 2019-02-25
Build Ansible Playbooks for each rule in every profile

jan-cerny:
#### Description:
For each profile it creates Ansible Playbooks for each rule that has an Ansible remediation available.  The build is a part of product target for all products.  The Playbook contains values of XCCDF Values selected in the profile. The output is generated to `/build/<product>/playbooks/<profile>`. Each Playbook is in a separate file `<rule_id>.yml`.

#### Rationale:
This format is required by components who need to use Playbooks that contain only fixes for single rule.


build-scripts/build_rule_playbooks.py
```diff
@@ -0,0 +1,272 @@
...
+        playbook = [play_ordered_dict]
+        playbook_path = os.path.join(output_dir, rule_id + ".yml")
+        with open(playbook_path, "w") as playbook_file:
+            ssg.yaml.ordered_dump(
+                playbook, playbook_file, default_flow_style=False
+            )
+
+    def create_playbooks_for_profile(self, profile_path, variables):
+        """
+        Creates playbooks for profile from tasks from snippets.
```
yuumasato: This is confusing.
Maybe you want to say `Creates a playbook for each rule selected in a profile`.

build-scripts/build_rule_playbooks.py
```diff
@@ -0,0 +1,272 @@
...
+        relative_guide_dir = ssg.utils.required_key(self.env_yaml,
+                                                    "benchmark_root")
+        self.guide_dir = os.path.abspath(os.path.join(product_dir,
+                                                      relative_guide_dir))
+        relative_profiles_dir = ssg.utils.required_key(self.env_yaml,
+                                                       "profiles_root")
+        self.profiles_dir = os.path.abspath(
+            os.path.join(product_dir, relative_profiles_dir))
+
+    def get_variable_selectors_from_profile(self, profile):
```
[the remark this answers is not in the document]
jan-cerny: Thanks, but I have found this needs to be reworked because `get_rules_from_profile` now ignores extended profiles, which  is wrong.

redhatrises:
@jan-cerny can you please be more descriptive for the PR's rationale.

build-scripts/build_rule_playbooks.py
```diff
@@ -0,0 +1,335 @@
...
+        """
+        profile_rules, profile_refines = self.get_profile_selections(profile)
+        profile_playbooks_dir = os.path.join(self.output_dir, profile.id_)
+        os.makedirs(profile_playbooks_dir)
+        snippet_path = os.path.join(self.input_dir, rule_id + ".yml")
+        if rule_id in profile_rules:
+            self.create_playbook(
+                snippet_path, rule_id, variables,
+                profile_refines, profile_playbooks_dir
+            )
```
yuumasato: Please, add an error message if specified `rule_id` is not part of the profile.

build-scripts/build_rule_playbooks.py
```diff
@@ -0,0 +1,335 @@
...
+        extracted from snippets. Created Playbooks are parametrized by
+        variables according to profile selection. Playbooks are written into
+        a new subdirectory in output_dir.
+        """
+        profile_rules, profile_refines = self.get_profile_selections(profile)
+
+        profile_playbooks_dir = os.path.join(self.output_dir, profile.id_)
+        os.makedirs(profile_playbooks_dir)
+
+        for snippet in os.listdir(self.input_dir):
```
yuumasato: It should be faster to cycle through `profile_rules` instead of all snippets found in directory.

ssg/build_yaml.py
```diff
@@ -131,6 +131,16 @@ def to_xml_element(self):
 
         return element
 
+    def get_rules(self):
```
yuumasato: `get_rules` can give impression that it returns all rules from profile, while it only returns explicitly selected rules.
The rules selected by "extension" are not returned.
My suggestion: `get_rule_selectors`

build-scripts/build_rule_playbooks.py
```diff
@@ -0,0 +1,335 @@
...
+        if profile.extends:
+            extended_profile_path = os.path.join(
+                self.profiles_dir, profile.extends + ".profile")
+            extended_profile = ssg.build_yaml.Profile.from_yaml(
+                extended_profile_path, self.env_yaml)
+            if not profile:
+                sys.stderr.write(
+                    "Could not parse profile %s.\n" % extended_profile_path)
+                return None
+            rules, variables = self.get_profile_selections(extended_profile)
```
yuumasato: I have a feeling that processing of "extened rules" should also be part of "Profile" class, but not sure if it will be useful.
I'm fine if it remains here though.

build-scripts/build_rule_playbooks.py
jan-cerny: Done.

build-scripts/build_rule_playbooks.py
jan-cerny: Fixed

ssg/build_yaml.py
jan-cerny: Fixed

build-scripts/build_rule_playbooks.py
jan-cerny: The profile class would need to have context of other profiles. I think it isn't necessary for now.

ssg/build_yaml.py
```diff
@@ -131,6 +131,16 @@ def to_xml_element(self):
 
         return element
 
+    def get_rule_selectors(self):
+        return list(filter(lambda x: "=" not in x, self.selections))
+
+    def get_variable_selectors(self):
+        variables = dict()
+        for var_selection in filter(lambda x: "=" in x, self.selections):
```
jan-cerny: If there will be 2 different lists in profile YAML files (one for rules and second for values), then we won't need to create weird methods.

yuumasato: I understand, but separating the variable selectors from rule selectors makes experience of crafting profiles a bit worse.
It is very convenient to have the variable selector close to the rule selector in profile yaml.

ssg/playbook_builder.py
```diff
@@ -158,6 +158,10 @@ def create_playbook(self, snippet_path, rule_id, variables,
                 (rule_id, snippet_path)
             )
 
+        tags = set()
+        for task in play_tasks:
+            tags |= set(task.pop("tags", []))
+
```
yuumasato: It would be interesting to have a test for this for a rule with multiple tasks. For example Ansible fix for audit rule privileged commands.

build-scripts/build_rule_playbooks.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def parse_args():
+    p = argparse.ArgumentParser()
+    p.add_argument(
+        "--input-dir", required=True, dest="input_dir",
+        help="Input directory that contains all Ansible remediations"
+        "snippets for the product we are building. "
+        "e.g. ~/scap-security-guide/build/fedora/fixes/ansible"
+    )
```
matejak: If we know the product YAML, we think that we also know where to look for the input directory - it is `build/<product-that-we-know>/fixes/ansible`. This means that we don't have to require this, because we know the default.

build-scripts/build_rule_playbooks.py
```diff
@@ -0,0 +1,47 @@
...
+    p.add_argument(
+        "--output-dir", required=True, dest="output_dir",
+        help="Output directory to which the rule-based Ansible playbooks "
+        "will be generated. "
+        "e.g. ~/scap-security-guide/build/fedora/playbooks"
+    )
```
matejak: Ditto, we know a sensible default for the output directory in a similar way as with the input dir.

build-scripts/build_rule_playbooks.py
```diff
@@ -0,0 +1,47 @@
...
+    p.add_argument(
+        "--product-yaml", required=True, dest="product_yaml",
+        help="YAML file with information about the product we are building. "
+        "e.g.: ~/scap-security-guide/rhel7/product.yml"
```
matejak: We could transfer this to a positional argument, as it is the only argument that is essentially required.

build-scripts/build_rule_playbooks.py
```diff
@@ -0,0 +1,47 @@
...
+    )
+    p.add_argument(
+        "--profile",
+        help="Generate Playbooks only for given Profile ID"
```
matejak: I would mention that we accept short profile IDs s.a. `ospp` and I would make it more explicit that this argument can only restrict the selection of profiles, i.e. that all are on by default.

ssg/playbook_builder.py
```diff
@@ -0,0 +1,275 @@
...
+                    value = self.choose_variable_value(var_id, variables,
+                                                       refinements)
+                    values[var_id] = value
+                else:
+                    raise ValueError("Found unknown item '%s'" % item)
+            else:
+                tasks.append(item)
+        return tasks, values
+
+    def get_tags_from_comments(self, snippet_path):
```
matejak: Are we sure that we need this? Enhanced snippets should already contain those.

ssg/playbook_builder.py
```diff
@@ -0,0 +1,275 @@
...
+                "Ansible remediation for rule '%s' in '%s' "
+                "doesn't contain any task." %
+                (rule_id, snippet_path)
+            )
+
+        tags = set()
+        for task in play_tasks:
+            tags |= set(task.pop("tags", []))
+
+        # play should be a dictionary, but to keep a logical order of elements
```
matejak: If keys are added to `OrderedDict` one-by-one, there is no need for the sequence, or for the comment. Although your approach is correct, I would favour explicit assignment.

ssg/playbook_builder.py
jan-cerny: Ah, I forgot to remove it, it's not used anywhere.

jan-cerny: I have removed unused method get_tags_from_comments

ssg/playbook_builder.py
jan-cerny: But from this code it's obviously visible that that we want to order it.

matejak: It is obvious from the point when `OrderedDict` appears.

build-scripts/build_rule_playbooks.py
jan-cerny: OK, I'm not sure if we will need them when the script is used standalone, so I created an issue as you suggested. See https://github.com/ComplianceAsCode/content/issues/4034

build-scripts/build_rule_playbooks.py
jan-cerny: OK, I'm not sure if we will need them when the script is used standalone, so I created an issue as you suggested. See #4034

jan-cerny:
@matejak I have removed unused method `get_tags_from_comments`. I have improved help. I have changed the use of the OrderedDict. I have created https://github.com/ComplianceAsCode/content/issues/4034  to track that required options do not need be required.

build-scripts/build_rule_playbooks.py
jan-cerny: Done

ssg/playbook_builder.py
jan-cerny: Done

build-scripts/build_rule_playbooks.py
jan-cerny: Depends on resolution of #4034

ssg/playbook_builder.py
yuumasato: Filed https://github.com/ComplianceAsCode/content/issues/4037 regarding this.

yuumasato:
Great patch @jan-cerny!

PR #3825, opened 2019-02-25
Propagate group properties to resolved rules.

matejak:
This PR improves the "resolve rule" into `build/<product>/rules` functionality.

Some groups (sshd, gnome, ...) have `platform` set to `machine`, so it doesn't have to be specified in all rules in underlying groups.
This enables properties inheritance (starting with only the `platform` property) that overrides child (group, rule) property if it is null.


yuumasato:
LGTM!

PR #3827, opened 2019-02-25
Produce resolved Ansible snippets early in the build process.

matejak:
This PR changes the Ansible remediations generation process. It aims to make build more transparent and reusable by exposing almost-ready remediations as individual files in the build directory. Next, it gets rid of `@ANSIBLE_TAGS@` and other placeholders and slightly reduces the amount of XSLT transforms performed during the build.

Past state:
* Ansible remediations were aggregated to the huge `fixes-ansible.xml` file as templates - they have contained those `@ANSIBLE_TAGS@` and other placeholders, they weren't valid YAML snippets.
* Later in the build process, an XSLT transform expanded those snippets to valid Ansible/YAML.

New state:
* The `@ANSIBLE_TAGS@` and some (not all) other placeholders were removed, as they are not meant to be used.
* Individual snippets are exported to `build/<product>/fixes/ansible/<rule-id>.yml` in form of valid YAML files, with `tags` and `when` populated correctly.

Regressions:
At this stage, no regressions are expected. As `pyyaml` is used in part of the processing, the indentation, quotation and interpretation of boolean values changed, so it is not trivial to compare new and old remediations. I have used profile roles with leading whitespace trimmed, which allows for quite nice comparison.

How it has been done:
* The remediation build has been slightly refactored by introducing classes - now e.g. Bash and Ansible remediations are quite different, and method override is a convenient way how to introduce additional steps to remediation build.
* The `AnsibleRemediation` class in `ssg.ansible` reads the remediation stub, looks up the resolved rule in the build directory, and adds metadata to the remediation.
* This happens during the remediation collection phase, i.e. before embedding remediations in huge XML files.
* Some XSLT is still performed later in the build, but the scope of it is now very limited.

The ugly:
There is now an `AnsibleRemediation` class in `ssg.ansible` and `ssg.build_remediations`. Although there is no code duplication, those two classes should be merged.
See also #3830.

### DIFF
--- a/build-scripts/combine_remediations.py
+++ b/build-scripts/combine_remediations.py
@@ -28,6 +28,10 @@ def parse_args():
         help="YAML file with information about the product we are building. "
         "e.g.: ~/scap-security-guide/rhel7/product.yml"
     )
+    p.add_argument(
+        "--resolved-rules", required=True,
+        help="Directory with <rule-id>.yml resolved rule YAMLs"
+    )
     p.add_argument("--remediation_type", required=True,
                    help="language or type of the remediations we are combining."
                    "example: ansible")
@@ -56,6 +60,8 @@ def main():
 
     # As fixes is continually updated, the last seen fix that is applicable for a
     # given fix_name is chosen to replace newer fix_names
+    remediation_cls = remediation.REMEDIATION_TO_CLASS[args.remediation_type]
+
     fixes = dict()
     for fixdir in args.fixdirs:
         if os.path.isdir(fixdir):
@@ -63,10 +69,10 @@ def main():
                 file_path = os.path.join(fixdir, filename)
                 fix_name, _ = os.path.splitext(filename)
 
+                remediation_obj = remediation_cls(
+                    env_yaml, args.resolved_rules, product, file_path, fix_name)
                 # Fixes gets updated with the contents of the fix, if it is applicable
-                remediation.process_fix(fixes, args.remediation_type,
-                                        env_yaml, product, file_path,
-                                        fix_name)
+                remediation_obj.process(fixes)
 
     # Walk the guide last, looking for rule folders as they have the highest priority
     for _dir_path in ssg.rules.find_rule_dirs(guide_dir):
@@ -76,8 +82,11 @@ def main():
         for _path in reversed(contents):
             # To be compatible with the later checks, use the rule_id
             # (i.e., the value of _dir) to create the fix_name
-            remediation.process_fix(fixes, args.remediation_type, env_yaml,
-                                    product, _path, rule_id)
+
+            remediation_obj = remediation_cls(
+                env_yaml, args.resolved_rules, product, _path, rule_id)
+            # Fixes gets updated with the contents of the fix, if it is applicable
+            remediation_obj.process(fixes)
 
     remediation.write_fixes_to_dir(fixes, args.remediation_type,
                                    args.output_dir)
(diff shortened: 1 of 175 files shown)


ssg/build_remediations.py
```diff
@@ -218,26 +220,94 @@ def parse_from_file_without_jinja(file_path):
...
+            self.resolved_rules, self.fix_name + ".yml")
+        if not os.path.isfile(rule_path):
+            msg = ("Ansible snippet for rule {rule_id} "
+                   "doesn't have respective rule YML at {rule_fname}"
+                   .format(rule_id=os.path.splitext(self.fix_name)[0], rule_fname=rule_path))
+            print(msg, file=sys.stderr)
+            return result
+
+        import ssg.ansible
+        from ssg import build_yaml
```
jan-cerny: Can the imports be at the beginning of file?

ssg/ansible.py
```diff
@@ -50,3 +56,137 @@ def remove_trailing_whitespace(ansible_src):
...
+        result += _strings_to_list(new_contents)
+    if not result:
+        result = ""
+    if len(result) == 1:
+        result = result[0]
+    return result
+
+
+class AnsibleRemediation(object):
+
```
jan-cerny: Do you think that this script cans also replace the `@RULE_TITLE@` macro?

tests/unit/ssg-module/data/file_owner_grub2_cfg.yml
```diff
@@ -0,0 +1,30 @@
+# File processed by the build system, i.e. retrieved from the build/rhel7/rules/file_owner_grub2_cfg.yml
+#
+description: 'The file <tt>/boot/grub2/grub.cfg</tt> should
+
+  be owned by the <tt>root</tt> user to prevent destruction
+
```
jan-cerny: Why are these empty lines there?

matejak: It has been taken straight from the build, so this corresponds to the produced content.

ssg/ansible.py
matejak: The proper solution to this is replacing those with jinja macros. As it is easy, but not trivial, I propose to consider it out of scope of this PR.

ssg/build_remediations.py
matejak: Unfortunately not, there is an import loop, and late imports are the only way how to make it work. The proper solution would be to unify the two remediation classes and straightening the dependency graph as a consequence.

matejak:
retest this please

matejak:
test this please

matejak:
test this please

matejak:
A reference note - extra care is needed when dealing with build steps that produce more than one outputs that is used by multiple build targets - it is advisable to mention only the latest output that is produced, and [according to cmake docs](https://cmake.org/cmake/help/v3.0/command/add_custom_command.html#add-custom-command) specify the "internal" target as dependency (as opposed to the actual produced file).

jan-cerny:
@matejak When I left the office yesterday I had the same idea near the fake lake :smiley: :smiley:

jan-cerny:
@matejak Great! I like your PR very much, it's very helpful. The generated content in `fixes` directories looks fine.

build-scripts/combine_remediations.py
```diff
@@ -28,6 +28,10 @@ def parse_args():
         help="YAML file with information about the product we are building. "
         "e.g.: ~/scap-security-guide/rhel7/product.yml"
     )
+    p.add_argument(
+        "--resolved-rules", required=True,
```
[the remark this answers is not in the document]
matejak: Tracked in #4027.

PR #4029, opened 2019-02-27
Unify command line arguments terminology

### LINKED ISSUE #4027: Unify terminology concerning rule YAMLS resolution at build-time
In  #3827, the directory with resolved rules is referred to as `--resolved-rules`, which differs from the `yaml_to_shorthand` script, which uses more accurate `--resolved-rules-dir`.

Make sure that this concept is used consistently in the project.

jan-cerny:
#### Description:
Changes `--resolved-rules` in `combine_remediations.py` and `--resolve-rules-into` in `yaml_to_shorthand.py` both to `--resolved-rules-dir`, because they are both describing the same directory. 

Also changes `--output_dir` in `combine_remediations.py` to `--output-dir` to use dashes consistently in the CLI interface.

Also updates the invocations of these scripts in CMake accordingly.

#### Rationale:
Resolves inconsistency created by  #3827 in whichthe directory with resolved rules is referred to as `--resolved-rules`, which differs from the `yaml_to_shorthand` script, which uses more accurate `--resolved-rules-dir`.
Fixes #4027

### DIFF
--- a/build-scripts/combine_remediations.py
+++ b/build-scripts/combine_remediations.py
@@ -19,27 +19,27 @@ import ssg.xml
 def parse_args():
     p = argparse.ArgumentParser()
     p.add_argument(
-        "--build-config-yaml", required=True, dest="build_config_yaml",
+        "--build-config-yaml", required=True,
         help="YAML file with information about the build configuration. "
         "e.g.: ~/scap-security-guide/build/build_config.yml"
     )
     p.add_argument(
-        "--product-yaml", required=True, dest="product_yaml",
+        "--product-yaml", required=True,
         help="YAML file with information about the product we are building. "
         "e.g.: ~/scap-security-guide/rhel7/product.yml"
     )
     p.add_argument(
-        "--resolved-rules", required=True,
+        "--resolved-rules-dir", required=True,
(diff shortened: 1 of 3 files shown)


build-scripts/combine_remediations.py
```diff
@@ -29,14 +29,14 @@ def parse_args():
...
     )
     p.add_argument("--remediation_type", required=True,
                    help="language or type of the remediations we are combining."
                    "example: ansible")
     p.add_argument(
-        "--output_dir", required=True,
+        "--output-dir", required=True, dest="output_dir",
         help="output directory where all remediations will be saved"
     )
     p.add_argument("fixdirs", metavar="FIX_DIR", nargs="+",
```
yuumasato: Please, change this to `fix_dirs` as well.

matejak:
This PR needs rebase, I would also remove the `dest=...`, because those are redundant.

jan-cerny:
I have resolved the conflicts, rebased on top of master, removed redundant `dest` arguments and renamed `fixdirs` variable to `fix_dirs`. I have also replaced underscores by dashes in all options in both scripts.

yuumasato:
Great, thank you!

PR #4035, opened 2019-02-28
Detect podman as container environment 

yuumasato:
#### Description:

- Extend `installed_env_is_a_container` to detect podman environment as a container.
- `installed_env_is_a_machine` extends and negates `installed_env_is_a_container`.

#### Rationale:

- Contents that are not for containers should not run on podman container
- An environment which is not a container is a machine environment.


jan-cerny:
I have tested it on `ssg-fedora-cpe-oval.xml` with OpenSCAP maint-1.2. It correctly detects host (machine), Docker container and Podman container. Great job!

PR #4038, opened 2019-03-01
Update Ansible section in Developer Guide

jan-cerny:
#### Description:
After recent changes, tags section are no longer required.
Also adds explanation where the Ansible remediations are used in the build process.

#### Rationale:
#3824 has enabled to generate separate playbooks for each rule and #3827 has removed the need to put tags and other placeholders into the Ansible remediation. We need to reflect those changes in documentation.


mildass:
Looks good to me

PR #4040, opened 2019-03-01
Make sure that Ansible tags don't contain spaces.

matejak:
We treat references as CSV string, so spaces in the description should be stripped.

For example, notice spaces in:
```
linux_os/guide/system/software/integrity/disable_prelink/rule.yml:
nist: 'CM-6(d),CM-6(3),SC-28, SI-7'

linux_os/guide/system/selinux/selinux_all_devicefiles_labeled/rule.yml:
cui: '3.1.2, 3.1.5,3.7.2'
```

This results in Ansible reference-based tags to contain spaces, e.g.

```
    tags:
    ...
    - NIST-800-171-3.1.1
    - NIST-800-171- 3.1.5
```


ssg/build_yaml.py
```diff
@@ -612,6 +613,11 @@ def validate_references(self, yaml_file):
                 raise ValueError("References must not be empty: %s in file %s"
                                  % (ref_type, yaml_file))
 
+    def normalize_references(self):
+        for ref_type, ref_val in self.references.items():
+            csv_elements = [ref.strip() for ref in ref_val.split(",")]
+            self.references[ref_type] = ",".join(csv_elements)
```
yuumasato: Should we issue an error and fail the build instead of fixing it silently?

matejak: I have also thought of that. I am slightly in favor of being more strict, but I have anticipated that the community may be against errors like that.

matejak:
Check out the updated description for an example.

ssg/build_yaml.py
redhatrises: @matejak I am completely in favor of being more strict in all YAML keys that contain comma-separated lists with a failure. While not the scope of this PR, it would also be nice to fail on trailing whitespace.

yuumasato: These errors will be easy to find and fix.

matejak:
I have changed the implementation to more strict validation, and I have fixed errors.

redhatrises:
LGTM.

PR #4043, opened 2019-03-01
Add change_when false to tasks that won't ever make changes

yuumasato:
#### Description:

- Add `changed_when: false` to shell tasks that won't ever make changes

#### Rationale:

- The `find` and `grep` commands won't make changes in the system.
  Their goal is to collect info.


jan-cerny:
Nice catch! Thank you for your patch.

PR #4044, opened 2019-03-03
Remove duplicate RHEL7 OSPP profiles

shawndwells:
Removing the duplicate -- why was this created in the first place?


yuumasato:
One profile conforms to OSPP4.2 and the other to OSPP4.0.

redhatrises:
> One profile conforms to OSPP4.2 and the other to OSPP4.0.

Once a new version of OSPP is released, the old version is invalid, and no one can certify or use it.

redhatrises:
@shawndwells there are some major differences between the OSPP and OSPP4.2 profile, can you resolve them?

jan-cerny:
NACK. Profile `ospp42` is required to be in the RHEL7 benchmark. It's a stable profile ID. We should not be rename or remove stable profiles. We have a test for that, which obviously fails on this PR.

shawndwells:
On 3/12/19 7:05 AM, Jan Černý wrote:
> NACK. Profile |ospp42| is required to be in the RHEL7 benchmark. It's 
> a stable profile ID. We should not be rename or remove stable 
> profiles. We have a test for that, which obviously fails on this PR.


Nack to your nack! :)

The stable ID is `ospp`.

Akin to how DoD STIG content is updated as DISA creates new versions, so 
should the OSPP profile be. We don't create a "rhel-stig-disa-v1r1"... 
"rhel-stig-disa-v1r2".... "rhel-stig-disa-v1r3"... etc.

Want to ensure the ospp42 profile never ships downstream in RHEL. Will 
be to confusing.

jan-cerny:
> Want to ensure the ospp42 profile never ships downstream in RHEL. Will be to confusing.

It happened. RHEL 7.6 released on 2018-10-30 contains both `ospp` and `ospp42` profiles.

shawndwells:
On 3/15/19 3:43 AM, Jan Černý wrote:
>
> It happened. RHEL 7.6 released on 2018-10-30 contains both |ospp| and 
> |ospp42| profiles.
>


Yikes. Not sure who approved that -- certainly not the government group. 
Please remove it.

yuumasato:
@shawndwells Is it a deal breaker if there are two "ospp" profiles?

From what you say, Public Sector will be using OSPP 4.2, but from https://github.com/ComplianceAsCode/content/pull/3393#issuecomment-427584687 other users would still use OSPP 4.0. Not sure who, though.

jan-cerny:
If we want to remove the OSPP 4.0 profile, we remove the `ospp.profile` file, not the `ospp42.profile` file. Then, we could rename `ospp42.profile` to `ospp.profile`.

rhel7/profiles/ospp42.profile
```diff
@@ -1,204 +0,0 @@
-documentation_complete: true
```
jan-cerny: The `ospp.profile` file should be removed instead. content of `ospp42.profile` file should be kept, because this one is more up2date, but  the file should be renamed to `ospp42.profile`.

jan-cerny: Please remove also the ospp42 from the stable profile IDs test.

shawndwells:
Multiple PRs floating around to refactor the OSPP profile. Will fold this into one of them.

PR #2328, opened 2017-09-14
Drop firewalld default zone and sshd port fixes

yuumasato:
- Fixes #2202 

The proper fix for #2202 is to have a remediation for `firewalld_sshd_port_enabled` which set ups a firewalld zone with SSH and an interface assigned to it.

But providing a good fix for `firewalld_sshd_port_enabled` can be very complicated
and will very likely not fit to everyone's use case. And because of that
we will drop remediation for `set_firewalld_sshd_port`, which is causing the
remediated machine to lock down and refuse all connections.

Existent test cases for 'firewalld_sshd_port_enabled' are kept because they are still useful to test the OVAL definition.

PR #2285, opened 2017-08-28
Firewalld open sshd port

yuumasato:
Select Rule `firewalld_sshd_port_enabled` in Profiles that
- set default zone drop for firewalld and
- allows sshd to be enabled
Except for `stig-rhel7-disa` Profile.

Add remediation for `configure_firewalld_ports`, used only by `stig-rhel7-disa` Profile.

Add OVAL definition to check if current default zone allows ssh. The check fetches current set default zone from `/etc/firewalld/firewalld.conf` and verifies if it allows ssh service.

**Issue:** Above check will lead us to a situation where we will need two runs of scans and remediation.
During first scan, all zones, and current zone will allow service ssh. But then remediation will change default zone to `drop`. And default zone `drop` is not customized to allow `ssh`, and a second run of remediation will be needed.
**Question:** Can we compromise on this check and always check for `/etc/firewalld/zones/drop.xml`?

PR #4045, opened 2019-03-03
Bash remediation for firewalld_sshd_port_enabled

shawndwells:


yuumasato:
We had a remediation for this, and it caused problems when put together with `set_firewalld_default_zone`. See, https://github.com/ComplianceAsCode/content/pull/2285 and  https://github.com/ComplianceAsCode/content/pull/2328.

In summary, the issue was that:
- an interface not explicitly assigned to a zone is assigned to default zone, and `set_firewalld_default_zone` will move all such interfaces to drop zone.
- remediation for `firewalld_sshd_port_enabled` enables  `ssh` for default zone, which can be `drop, if run after `set_firewalld_default_zone`. And strange situation where we enable `ssh` service in `drop` zone.

The attempted solution in #2285 was to:
- Introduce a Value for target SSH zone,
- enable `ssh` service in that zone.
- if there was no interface was associated with that zone,
  - Add first interface found to the zone.

But adding the first interface found to a zone is not ideal. But a value to select interface could handle this, 
see comment in https://github.com/ComplianceAsCode/content/pull/2285#issuecomment-329176308.

PR #4046, opened 2019-03-03
Remediations for set_firewalld_default_zone

shawndwells:


jan-cerny:
This will prevent users from logging in by SSH and could conflict with existing firewall configuration.
I believe we haven't added the remediation intentionally.  See https://github.com/ComplianceAsCode/content/blob/08628341e72eaa0e94bc898774c575011ace05c1/linux_os/guide/system/network/network-firewalld/ruleset_modifications/set_firewalld_default_zone/rule.yml#L50-L55

Also, your Ansible task has wrong syntax.

PR #4047, opened 2019-03-03
Remove legacy ensure_gpgcheck_repo_metadata rule from profiles

shawndwells:
This was removed from other profiles -- didn't realize it was still lingering


yuumasato:
@shawndwells these rules were removed from STIG profile in #3706 as policy for them changed.
Should they be dropped for these profiles as well?

yuumasato:
Closing as I see #4206 superseding this.

PR #4049, opened 2019-03-04
Add OVAL namespace to oval/metadata/reference

yuumasato:
#### Description:

- Add `reference` element in OVAL namespace.

#### Rationale:

- The oval definitions should have `metadata/reference` in OVAL namespace.

- Content diff

From:
```xml
          <oval:metadata>
            <oval:title>Send Logs to a Remote Loghost</oval:title>
            <oval:affected family="unix">
              <oval:platform>Red Hat Enterprise Linux 7</oval:platform>
            </oval:affected>
            <oval:description>Syslog logs should be sent to a remote loghost</oval:description>
            <oval:reference ref_id="CCE-27343-3" source="CCE"/>
            <oval:reference ref_id="rsyslog_remote_loghost" source="ssg"/>
          </oval:metadata>
```
to:
```xml
          <oval:metadata>
            <oval:title>Send Logs to a Remote Loghost</oval:title>
            <oval:affected family="unix">
              <oval:platform>Red Hat Enterprise Linux 7</oval:platform>
            </oval:affected>
            <oval:description>Syslog logs should be sent to a remote loghost</oval:description>
            <reference ref_id="CCE-27343-3" source="CCE"/>
            <reference ref_id="rsyslog_remote_loghost" source="ssg"/>
          </oval:metadata>
```

### DIFF
--- a/ssg/build_renumber.py
+++ b/ssg/build_renumber.py
@@ -180,7 +180,7 @@ class OVALFileLinker(FileLinker):
             if is_cce_format_valid(xccdfcceid) and is_cce_value_valid(xccdfcceid):
                 # Then append the <reference source="CCE" ref_id="CCE-ID" /> element right
                 # after <description> element of specific OVAL check
-                ccerefelem = ET.Element('reference', ref_id=xccdfcceid,
+                ccerefelem = ET.Element('{%s}reference' % self.CHECK_NAMESPACE, ref_id=xccdfcceid,
                                         source="CCE")
                 metadata = rule.find(".//{%s}metadata" % self.CHECK_NAMESPACE)
                 metadata.append(ccerefelem)
--- a/ssg/id_translate.py
+++ b/ssg/id_translate.py
@@ -74,7 +74,7 @@ class IDTranslator(object):
                     if metadata is None:
                         metadata = ElementTree.SubElement(element, "metadata")
                     defnam = ElementTree.Element(
-                        "reference", ref_id=idname, source=self.content_id)
+                        "{%s}reference" % oval_ns, ref_id=idname, source=self.content_id)
                     metadata.append(defnam)
 
                 # set the element to the new identifier


jan-cerny:
@yuumasato Nice catch. The patch looks good to me. You probably swapped "from" and "to" in the description. It's unfortunate that OVAL specification allows any XML element in `metadata` so the schema or schematron validation haven't caught this. Could we have a test for it?

yuumasato:
> You probably swapped "from" and "to" in the description.

Ops, yeah, they are swapped.

> It's unfortunate that OVAL specification allows any XML element in `metadata` so the schema or schematron validation haven't caught this. Could we have a test for it?

I'm not sure it would be of benefit. These were actually caught by `scapval`:
```
      <derived-requirement id="SRC-207-1">
        <status>WARNING</status>
        <test test-id-ref="scap-general-oval-def-rule-compliance-cce">
          <status>FAIL</status>
          <message>'oval-def:definition oval:ssg-rsyslog_remote_loghost:def:1'</message>
          <location line="45" column="100" .../>
        </test>
```
And the requirement reads as:
```
    <requirement id="SRC-207">
        <summary/>
        <reference ref="NIST800-126r3" section="3.3"/>
        <statement>Required values for the @class attribute of an OVAL Definition are as follows:~"compliance" if it represents a check for the system's
            configuration complying with policy requirements (for example, having the required value for a specific configuration setting).~"vulnerability" if
            it represents a check for the presence of a particular software flaw vulnerability on a system.~"patch" if it represents a check for whether a
            discrete patch needs to be installed on the system.~"inventory" if it represents a check for the presence of a product of interest on the
            system.~The following requirements apply to particular classes of OVAL Definitions:~~For compliance class definitions:~If an OVAL compliance class
            definition maps to one or more CCE identifiers, the definition SHOULD include &lt;oval-def:reference&gt; elements that reference those identifiers
            using the following format: ~&lt;oval-def:reference source="http://cce.mitre.org" ref_id="CCE_identifier"/&gt;_x000B__x000B_The source attribute
            SHALL be defined using either "http://cce.mitre.org" (preferred method) or "CCE".
        </statement>
        <derived-requirements>
            <derived-requirement id="SRC-207-1" type="SHOULD" scapval-ext:check-type="SCHEMATRON">
                <statement>OVAL definitions of class 'compliance' should include a reference to a CCE, where applicable.</statement>
                <message>''{0}''</message>
            </derived-requirement>
        </derived-requirements>
    </requirement>
```

jan-cerny:
OK

PR #4050, opened 2019-03-04
[WIP][RFC] Add grub2 ownership rules to check UEFI and non-UEFI systems

### LINKED ISSUE #2872: Rules for non-UEFI grub.cfg cannot remediate UEFI systems
#### Description of problem:
This may sound obvious, but for example:
Rule description of Rule `file_permissions_grub_cfg` says `grub.cfg` file should have permissions 600 or stronger. Mentions location of both files, `/boot/grub2/grub.cfg` and `/boot/efi/EFI/redhat/grub.cfg`, and will check for permissions in both places. (for `/boot/efi/EFI/redhat/grub.cfg` it will actually check for 700).
But [remediation](https://github.com/OpenSCAP/scap-security-guide/blob/master/shared/fixes/bash/file_permissions_grub2_cfg.sh) associated with it only fixes `/boot/grub2/grub.cfg`, for non-UEFI systems.

So rule can evaluate whether a system has its `grub.cfg` file with correct permissions on both systems, but can only fix permission on one type.

This also happens for Rules:
- `file_user_owner_grub2_cfg`
- `file_group_owner_grub2_cfg`

scap-security-guide-0.1.39

#### Steps to Reproduce:

1. Install a system with UEFI boot
1. Break the rule (`chmod 644 /boot/efi/EFI/redhat/grub.cfg`)
1. Scan a system with UEFI boot with a profile with Rule `file_group_owner_grub2_cfg` selected , for example, rht-ccp profile
1. Rule will fail, try to remediate
1. Rule will try to remediate `/boot/grub2/grub.cfg`

#### Actual Results:
Rule tries to remediate `/boot/grub2/grub.cfg`

#### Expected Results:
Permissions on file `/boot/efi/EFI/redhat/grub.cfg` are fixed.

Possible solutions are either:
- Improve remediation for `file_.*_grub2_cfg` rules to be capable of fixing UEFI systems, or
- Separate completely checks and description.

iokomin:
#### Description:
Existing rules file_permissions_grub_cfg, file_groupowner_grub2_cfg and file_owner_grub2_cfg
don't work on UEFI systems, see #2872 for details. 

This PR addresses ownership rules - suggested to add conditional rules which extend existing grub2 cfg file template definitions. 

Please leave your feedback if proposed fix works for grub2 rules on UEFI and non-UEFI systems.

#### Rationale:
- Fixes #2872

### DIFF
--- a/linux_os/guide/system/bootloader-grub2/file_groupowner_grub2_config/bash/shared.sh
+++ b/linux_os/guide/system/bootloader-grub2/file_groupowner_grub2_config/bash/shared.sh
@@ -1,0 +1,14 @@
+# platform = multi_platform_all
+# reboot = false
+# strategy = configure
+# complexity = low
+# disruption = low
+
+GRUB2_CONF_NON_EFI=/boot/grub2/grub.cfg
+GRUB2_CONF_EFI=/boot/efi/EFI/redhat/grub.cfg
+
+if [ -f $GRUB2_CONF_EFI ]; then
+    chgrp 0 $GRUB2_CONF_EFI
+elif [ -f $GRUB2_CONF_NON_EFI ]; then
+    chgrp 0 $GRUB2_CONF_NON_EFI
+fi
--- a/linux_os/guide/system/bootloader-grub2/file_groupowner_grub2_config/oval/shared.xml
+++ b/linux_os/guide/system/bootloader-grub2/file_groupowner_grub2_config/oval/shared.xml
@@ -1,0 +1,20 @@
+<def-group>
+  <definition class="compliance" id="file_groupowner_grub2_config" version="1">
+    <metadata>
+      <title>Grub2 configuration file Owned By root User</title>
+      <affected family="unix">
+        <platform>Red Hat Enterprise Linux 7</platform>
+        <platform>Red Hat Enterprise Linux 8</platform>
+        <platform>multi_platform_fedora</platform>
+        <platform>multi_platform_ol</platform>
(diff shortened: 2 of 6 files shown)


redhatrises:
Kind NACK. These rules add more than one configuration item to them as well as duplicate the existing rules. Per CCE, should be 1-to-1, or one rule to one component. As both rules already exist, the EFI rules should just be added to the profiles themselves.

iokomin:
@redhatrises, I got your point and follow 1-to-1 rule per component does make sense for me. 
Note: My assumption in favor of proposed solution was that we are reusing existing per-cfg definitions and rule addresses GRUB2 cfg file.

I've reviewed existing solutions and seems "test_bootloader_uefi_grub_cfg" from https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/bootloader-grub2/grub2_uefi_password/oval/shared.xml#L16 should work for these rules then. I would better return "notapplicable" in case UEFI rule tried on non-UEFI system, but if returning 'pass' acceptable then we could go this way.

Please leave your feedback on using "test_bootloader_uefi_grub_cfg" as a solution, I'll prepare corresponding changes.

redhatrises:
Basically, there should be separate rules for EFI grub checks and non-EFI grub checks. Then, non-EFI checks should pass on EFI systems and EFI checks should pass on non-EFI systems.

iokomin:
Closing RFC, #2872 will be addressed in another PR.

PR #4051, opened 2019-03-04
Use separate rule for each audit syscall in pci-dss

iokomin:
#### Description:
Use separate rule for each audit syscall in pci-dss

#### Rationale:
Follow combined rules warnings to use separate rules. As a result it makes syscall rules OVAL and remediation content easier to maintain.


jan-cerny:
LGTM. Thank you!

PR #4054, opened 2019-03-05
Add missing rules for docker and psacct packages installed

jan-cerny:
#### Description:
*  Create rule `package_docker_installed` and add it to the profile where `service_docker_enabled` is included.
*  Create rule `package_psacct_installed` and add it to all profiles where `service_psacct_enabled` is included.
* Remove docker rules form RHEL8 benchmark because Docker isn't available on RHEL8.

#### Rationale:
* Ansible remediation for rule `service_docker_enabled` fails because the service `docker` was not found because package `docker` was not installed.
* Ansible remediation for rule `service_psacct_enabled` fails because the service `psacct` was not found because package `psacct` was not installed.


yuumasato:
LGTM.

PR #4055, opened 2019-03-05
Update ansible template, readme, and script to bring in line with Ansible Galaxy

redhatrises:
- Align role name to `ansible_role_<profile_name>`
- Add `role_name` to metadata to allow simpler short role naming
- Add new capability to tag releases 
- Remove XCCDF references in readme. This doesn't make sense for an ansible role
- Remove references to "Playbook" unless used in the context of calling an ansible role
- Add product whitelist to remove hardcoded `rhel7`
- Add capability to pass token with arg option
- Update Readme and metadata only with non-role specific (e.g. role tags) changes
- Map ssg ansible role yaml files to `ansible_role_<profile_name>` convention

### DIFF
--- a/utils/ansible_galaxy_readme_template.md
+++ b/utils/ansible_galaxy_readme_template.md
@@ -4,7 +4,10 @@
 @DESCRIPTION@
 
 The tasks that are used in this role are generated using OpenSCAP.
-See the OpenSCAP project for more details at [https://github.com/OpenSCAP/openscap](https://github.com/OpenSCAP/openscap)
+See the OpenSCAP project for more details on Ansible playbook generation at [https://github.com/OpenSCAP/openscap](https://github.com/OpenSCAP/openscap)
+
+To submit a fix or enhancement for an Ansible task that is failing or missing,
+see the ComplianceAsCode project at [https://github.com/ComplianceAsCode/content](https://github.com/ComplianceAsCode/content)
 
 Requirements
 ------------
@@ -31,7 +34,7 @@ download and install the role. Then, you can use the following playbook snippet 
       roles:
          - { role: RedHatOfficial.@ROLE_NAME@ }
 
-Then, check the playbook using (on the localhost) the following example:
+Next, check the playbook using (on the localhost) the following example:
 
     ansible-playbook -i "localhost," -c local --check playbook.yml
 
--- a/utils/ansible_galaxy_readme_template.md
+++ b/utils/ansible_galaxy_readme_template.md
@@ -3,6 +3,9 @@
 
 @DESCRIPTION@
 
+The tasks that are used in this role are generated using OpenSCAP.
+See the OpenSCAP project for more details at [https://github.com/OpenSCAP/openscap](https://github.com/OpenSCAP/openscap)
+
 Requirements
 ------------
 
@@ -18,17 +21,17 @@ Dependencies
 
 N/A
 
-Example Playbook
+Example Role Usage
 ----------------
 
 Run `ansible-galaxy install RedHatOfficial.@ROLE_NAME@` to
-download and install the role. Then you can use the following playbook snippet.
+download and install the role. Then, you can use the following playbook snippet to run the Ansible role:
 
     - hosts: all
       roles:
          - { role: RedHatOfficial.@ROLE_NAME@ }
 
-Then first check the playbook using (on the localhost):
+Then, check the playbook using (on the localhost) the following example:
 
     ansible-playbook -i "localhost," -c local --check playbook.yml
 
--- a/utils/upload_ansible_roles_to_galaxy.py
+++ b/utils/upload_ansible_roles_to_galaxy.py
@@ -24,6 +24,7 @@ except ImportError:
 
 
 import ssg.ansible
+import ssg.yaml
 
 # The following code preserves ansible yaml order
 # code from arcaduf's gist
@@ -139,7 +140,7 @@ class Role(object):
         with io.open(self.local_playbook_filename, 'r', encoding="utf-8") as f:
             filedata = f.read()
 
-        self.role_data = yaml.load(filedata)
+        self.role_data = ssg.yaml.ordered_load(filedata)
         if "vars" in self.role_data[0]:
             self.vars_data = self.role_data[0]["vars"]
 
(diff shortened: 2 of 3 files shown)


utils/ansible_galaxy_readme_template.md
```diff
@@ -3,6 +3,9 @@
 
 @DESCRIPTION@
 
+The tasks that are used in this role are generated using OpenSCAP.
+See the OpenSCAP project for more details at [https://github.com/OpenSCAP/openscap](https://github.com/OpenSCAP/openscap)
```
shawndwells: Should ComplianceAsCode be mentioned here instead?

e.g. ``Tasks used in this role are generated from the ComplianceAsCode Project. For more details, see https://github.com/ComplianceAsCode/content``

utils/ansible_galaxy_readme_template.md
```diff
@@ -18,17 +21,17 @@ Dependencies
 
 N/A
 
-Example Playbook
+Example Role Usage
```
shawndwells: Thanks for doing this. Know it was annoying some people :)

utils/ansible_galaxy_readme_template.md
```diff
@@ -18,17 +21,17 @@ Dependencies
...
 Run `ansible-galaxy install RedHatOfficial.@ROLE_NAME@` to
-download and install the role. Then you can use the following playbook snippet.
+download and install the role. Then, you can use the following playbook snippet to run the Ansible role:
 
     - hosts: all
       roles:
          - { role: RedHatOfficial.@ROLE_NAME@ }
 
-Then first check the playbook using (on the localhost):
+Then, check the playbook using (on the localhost) the following example:
```
shawndwells:
```suggestion
Next, check the playbook using (on the localhost) the following example:
```

shawndwells: (because ``Then,...`` was used in the last paragraph.

redhatrises: @shawndwells fixed

utils/ansible_galaxy_readme_template.md
redhatrises: @shawndwells I added both. People need to understand that the tasks are generated using OpenSCAP, but that ComplianceAsCode is where to submit fixes.

utils/upload_ansible_roles_to_galaxy.py
```diff
@@ -362,10 +445,20 @@ def locally_clone_and_init_repositories(organization, repo_list):
 def main():
     args = parse_args()
 
-    all_role_whitelist = {"rhel7-role-%s" % p for p in PROFILE_WHITELIST}
+    all_product_whitelist = {"%s" % p for p in PRODUCT_WHITELIST}
```
jan-cerny: `all_product_whitelist` is now a set that contains same elements as `PRODUCT_WHITELIST`. Please remove this.

utils/upload_ansible_roles_to_galaxy.py
```diff
@@ -362,10 +445,20 @@ def locally_clone_and_init_repositories(organization, repo_list):
...
+    product_whitelist = set(all_product_whitelist)
+    if args.product:
+        selected_products = {"%s" % p for p in args.product}
```
jan-cerny: `selected_products = set(args.product)` will do the same job

utils/upload_ansible_roles_to_galaxy.py
```diff
@@ -362,10 +445,20 @@ def locally_clone_and_init_repositories(organization, repo_list):
...
+
+    all_role_whitelist = []
+    for product in product_whitelist:
+        for p in PROFILE_WHITELIST:
+            all_role_whitelist.append("%s-role-%s" % (product, p))
+
     role_whitelist = set(all_role_whitelist)
     if args.profile:
-        selected_roles = {"rhel7-role-%s" % p for p in args.profile}
+        selected_roles = {"%s" % p for p in args.profile}
```
jan-cerny: `selected_roles = set(args.profile)` should do the same job, but it's more understandable.

jan-cerny: Let's assume I run the script with `--product rhel8 --profile ospp`. At this point, the variables contain:
```
role_whitelist = {'rhel8-role-nist-800-171-cui', 'rhel8-role-pci-dss', 'rhel8-role-ospp', 'rhel8-role-stig-rhel7-disa', 'rhel8-role-hipaa', 'rhel8-role-C2S', 'rhel8-role-cjis', 'rhel8-role-rht-ccp'}
selected_roles = {'ospp'} 
```
The intersection of those 2 sets is obviously an empty set, so after executing the line below the `role_whitelist` is an empty set. That means no role will be updated if user specifies `--profile`.

utils/upload_ansible_roles_to_galaxy.py
redhatrises: @jan-cerny fixed

utils/upload_ansible_roles_to_galaxy.py
redhatrises: @jan-cerny fixed

utils/upload_ansible_roles_to_galaxy.py
redhatrises: @jan-cerny this should be working now.

jan-cerny:
It works OK for me.

I have one more suggestion: People who use SSH key authentication would prefer if the repository is cloned using SSH instead of HTTPs here: https://github.com/ComplianceAsCode/content/blob/0cbce045d17ec1adf4d095b37893b1cf1ecc7654/utils/upload_ansible_roles_to_galaxy.py#L86
i.e. `git clone git@github.com:%s/%s`.
If HTTPs is used, it asks for username and passphrase which is annoying, and the authentication fails when people use 2FA. If the SSH is used it would work seamlessly.

redhatrises:
> It works OK for me.
> 
> I have one more suggestion: People who use SSH key authentication would prefer if the repository is cloned using SSH instead of HTTPs here:
> 
> [content/utils/upload_ansible_roles_to_galaxy.py](https://github.com/ComplianceAsCode/content/blob/0cbce045d17ec1adf4d095b37893b1cf1ecc7654/utils/upload_ansible_roles_to_galaxy.py#L86)
> 
> Line 86 in [0cbce04](/ComplianceAsCode/content/commit/0cbce045d17ec1adf4d095b37893b1cf1ecc7654)
>  "git clone https://github.com/%s/%s" % (organization, repo)) 
> 
> i.e. `git clone git@github.com:%s/%s`.
> If HTTPs is used, it asks for username and passphrase which is annoying, and the authentication fails when people use 2FA. If the SSH is used it would work seamlessly.

Github recommend `https` over ssh. Also, there was numerous issues with script using the git version. For handling 2FA, using a github personal access token makes everything all easy and simple.

jan-cerny:
@redhatrises Aha OK

PR #4059, opened 2019-03-06
rsyslog_remote_loghost rule update for ANSSI NT28 profile

matusmarhefka:
#### Description:

* Fixed rsyslog_remote_loghost rule recommendation ID from NT28(R5)
  to the NT28(R7) and moved it to the intermediary level profile.
* For details see:
  https://www.ssi.gouv.fr/uploads/2015/10/NP_Linux_Configuration.pdf

### DIFF
--- a/rhel7/profiles/anssi_nt28_enhanced.profile
+++ b/rhel7/profiles/anssi_nt28_enhanced.profile
@@ -8,4 +8,5 @@ description: 'Draft profile for ANSSI compliance at the enhanced level. ANSSI st
 
 extends: anssi_nt28_intermediary
 
-selections: []
+selections:
+    - rsyslog_remote_loghost
--- a/linux_os/guide/system/logging/rsyslog_sending_messages/rsyslog_remote_loghost/rule.yml
+++ b/linux_os/guide/system/logging/rsyslog_sending_messages/rsyslog_remote_loghost/rule.yml
@@ -42,7 +42,7 @@ identifiers:
 
 references:
     cis@debian8: 5.1.5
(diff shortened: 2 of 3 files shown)


rhel7/profiles/anssi_nt28_enhanced.profile
```diff
@@ -8,4 +8,5 @@ description: 'Draft profile for ANSSI compliance at the enhanced level. ANSSI st
 
 extends: anssi_nt28_intermediary
 
-selections: []
+selections:
+    - rsyslog_remote_loghost
```
redhatrises: @matusmarhefka should this go into the intermediary profile instead of the enchanced since the intermediary extends the enhanced profile?

jan-cerny: @redhatrises I disagree. The rule belongs to ANSSI requirement R7. According to https://www.ssi.gouv.fr/uploads/2015/10/NP_Linux_Configuration.pdf page 12 the requirement R7 belongs only to renforcé (which we translate as enhanced) and élevé (which we translate as high).

Also, enhanced extends intermediary, intermediary doesn't extend enhanced.
Check page 6 in  https://www.ssi.gouv.fr/uploads/2015/10/NP_Linux_Configuration.pdf where is a nice picture of the profile hierarchy.

redhatrises: Now that I know that we have an unofficial english version. I rescind my comment. :)

redhatrises:
LGTM.

PR #4060, opened 2019-03-06
Dont generate templated content for nails

yuumasato:
#### Description:

- Don't generate templated contents for nails

#### Rationale:

- Ansible task for service nails enabled fails with service not found.
```
TASK [Enable service nails] ***************************************************************************************************************************************************************************************
fatal: [rhel7.6-kicked]: FAILED! =>
{"changed": false, "msg": "Could not find the requested service nails: host"} 
```

- Decided to drop generation of content, as:
  - It is not selected in any profile
  - The package is no provided by RHEL


redhatrises:
Kind Nack... this useful as it is mandated in some environments. Plus, is it added to a role where the profile is not selected?

yuumasato:
> Plus, is it added to a role where the profile is not selected?

No, it isn't added to a role unless selected in the profile.

I don't have a strong opinion to remove this.

PR #4063, opened 2019-03-07
Preprocess references and identifiers during the build time.

matejak:
When building for product X, if there is a qualified identifier `cce@X`, create internally an unqualified identifier `cce` with the same value as `cce@X`, and `cce@X` is removed, as all other irrelevant `cce@...`.
This brings the non-XSLT content generation one step closer, as now, there will be a `cce` property, which means "CCE for the product which is currently being built".

If there already is a unqualified identifier `cce` with different value than `cce@X`, raise an exception.

If references are involved, if e.g. product X is being built, and there are `stigid@X` equal to `"qualified"` and there is `stigid` equal to `"default"`, the reference `stigid` is modified, and it has the value of the `stigid@X`, i.e. `"qualified"` without raising an exception. Like that, the unqualified reference is treated as a default that can be overriden.


shawndwells:
On 3/7/19 10:52 AM, Matěj Týč wrote:
>
> When building for product X, if there is a qualified identifier 
> |cce@X|, create internally an unqualified identifier |cce| with the 
> same value as |cce@X|. This can be thought of as an alias, and it is 
> present in processed rules in the |build/X/rules/| directory.
> This brings the non-XSLT content generation one step closer, as the 
> rule pre-selects all relevant attributes, and all consumers of this 
> data don't have to figure out what the actual rule's CCE for the 
> particular product is.
>
> If there already is a unqualified identifier |cce| with different 
> value than |cce@X|, raise an exception.
>
> The content currently violates this restriction, e.g. 
> |rsyslog_remote_loghost| rule contains |cis@debian8: 5.1.5| and |cis: 
> 4.2.1.4| references, so some rules may need clarification.
>


Since CCEs are vendor specific, should ``cce@rhel7`` be aliased to 
``cce`` for a given rule? Don't want Debian CCEs showing up in RHEL 
content, or RHEL CCEs in CentOS for example.

matejak:
The description may be confusing, so let me add an additional explanation.

What **doesn't change**:

- The current build result - shorthand XCCDF is the same, and everything else stays the same, too.

What **changes**:

- The processed rules that get re-saved to yaml files to `build/<product>/rules/` are changed - some sort of alias has been added, as noted in the PR description.
- Rules that feature a mix of qualified and unqualified references/identifiers now cause build exception.

Why:

- Processed rules may be used by other parts of the build system - for example, they are used to complete the Ansible content at the moment. The introduced step will make it even easier - so far, a rule's CCE had to be deduced from the processed rule and the product. If this is merged, a rule's CCE will be simply available as a `cce` dictionary key.

redhatrises:
@matejak I think that this is great for identifiers as they have some strict checking. However for references, there shouldn't be strict checking except maybe if they are the same value. The reason being is that older product versions might have different values and mappings than newer products and versions (e.g. RHEL6 SRG mappings vs RHEL7 SRG mappings)

jan-cerny:
@matejak Would it make sense to do it just for CCEs at this moment?

matejak:
Great idea, of @jan-cerny and @redhatrises, this PR now does exactly that.
However, it needs double-checking that it doesn't change the output.

jan-cerny:
@matejak Why is `build/.gitkeep` removed?

ssg/build_yaml.py
```diff
@@ -570,6 +570,48 @@ def from_yaml(yaml_file, env_yaml=None):
         rule.validate_references(yaml_file)
         return rule
 
+    def add_product_specific_items(self, product):
```
jan-cerny: I think this method should have a different name, because it doesn't add anything, it does the opposite - it removes items that are not relevant to the product.

matejak:
Good point, I went for a more fitting name.

ssg/ansible.py
```diff
@@ -120,31 +120,29 @@ def update_tags_from_rule(self, platform, to_update):
         to_update["tags"] = tags
 
     def _get_cce(self, platform):
```
jan-cerny: Remove `platform` because it's an unused parameter.

ssg/build_yaml.py
```diff
@@ -570,6 +570,52 @@ def from_yaml(yaml_file, env_yaml=None):
...
+            except ValueError as exc:
+                msg = (
+                    "Error processing {what} for rule '{rid}': {msg}"
+                    .format(what=name, rid=self.id_, msg=str(exc))
+                )
+                raise ValueError(msg)
+            dic.clear()
+            dic.update(new_items)
+
+    def _make_items_product_specific(self, items_dict, product_suffix, allow_overwrites=False):
```
jan-cerny: This is a nice rename, However, there already is `make_items_product_specific`. Please rename one of them

ssg/build_yaml.py
```diff
@@ -570,6 +570,52 @@ def from_yaml(yaml_file, env_yaml=None):
         rule.validate_references(yaml_file)
         return rule
 
+    def normalize(self, product):
+        self.make_items_product_specific(product)
```
jan-cerny: The method `normalize` doesn't do anything additional on top of `make_items_product_specific`. Please remove one of them.

jan-cerny:
@matejak Please resolve the CI failure, looks like the code in `make_refs_and_identifiers_product_specific` doesn't handle the `None` type properly.

jan-cerny:
```
=================================== FAILURES ===================================
_______________________ test_make_items_product_specific _______________________

    def test_make_items_product_specific():
        rule = ssg.build_yaml.Rule("something")
    
        rule.identifiers = {
            "cce@rhel6": "27100-7",
            "cce@rhel7": "27445-6",
            "cce@rhel8": "80901-2",
        }
>       rule.make_items_product_specific("rhel7")
E       AttributeError: 'Rule' object has no attribute 'make_items_product_specific'

../../tests/unit/ssg-module/test_build_yaml.py:36: AttributeError

```

jan-cerny:
@openscap-ci test this please

matejak:
@redhatrises , @shawndwells How is it with STIG IDs? Are they more like CCEs, when one rule has multiple CCEs as there is one CCE for rule-product combination? Or is there one Stig ID per rule for any product?

shawndwells:
Depends which one you’re referring to.

One config check may have multiple mappings to an OS SRG or CCI.

There should be a one-to-one relationship with the RHEL-07-###### IDs though. Disa uses those instead of CCEs.

In practice many disa rules are not granular enough and check multiple settings. So sometimes there may be duplicate RHEL-07-###### mappings.

matejak:
Thank you for the answer, I understood that multiple rules may share a STIG ID. However, what about a slightly different case, when a rule is applicable for different OSs?
Consider this: Rule

* `file_permissions_home_directories` has  `stigid: "020630"` specified in `rule.yml`, while rule
* `accounts_no_uid_except_zero` has `stigid@rhel7: "020310"` specified.

As a result, the [respective guide](https://static.open-scap.org/ssg-guides/ssg-rhel7-guide-stig.html#xccdf_org.ssgproject.content_rule_file_permissions_home_directories) shows that the former rule got the `RHEL-07-020640` STIG ID, as it was built for the RHEL7 product. If it was part of OL7 profile, it would get the `OL-07-020640` ID automatically. Is this intended, or is it a mistake in the content and we want all `stigid` to be product-qualified, i.e. `stigid@rhel7: "020630"` is OK, but `stigid: "020630"` is wrong?

shawndwells:
On 6/11/19 11:32 AM, Matěj Týč wrote:
>
> Thank you for the answer, I understood that multiple rules may share a 
> STIG ID. However, what about a slightly different case, when a rule is 
> applicable for different OSs?
> Consider this: Rule
>
>   * |file_permissions_home_directories| has |stigid: "020630"|
>     specified in |rule.yml|, while rule
>   * |accounts_no_uid_except_zero| has |stigid@rhel7: "020310"| specified.
>
> As a result, the respective guide 
> <https://static.open-scap.org/ssg-guides/ssg-rhel7-guide-stig.html#xccdf_org.ssgproject.content_rule_file_permissions_home_directories> 
> shows that the former rule got the |RHEL-07-020640| STIG ID, as it was 
> built for the RHEL7 product. If it was part of OL7 profile, it would 
> get the |OL-07-020640| ID automatically. Is this intended, or is it a 
> mistake in the content and we want all |stigid| to be 
> product-qualified, i.e. |stigid@rhel7: "020630"| is OK, but |stigid: 
> "020630"| is wrong?
>

Every operating system (RHEL 6, RHEL 7, OEL6, OEL7) will have their own, 
unique, identifiers.

For config checks with just "stigid: value" we should check which 
operating system (RHEL, OEL, Ubuntu....) that is actually for and update 
to syntax similar to "stigid@rhel7: value"

linux_os/guide/services/http/securing_httpd/httpd_configure_log_format/rule.yml
```diff
@@ -24,7 +24,7 @@ identifiers:
     cce@rhel7: "80548-1"
 
 references:
-    stigid: WA00612
```
jan-cerny: @matejak Why is it twice in `ssg-rhel7-ds.xml`?

See:
```
<reference href="https://public.cyber.mil/stigs/downloads/?_dl_facet_stigs=operating-systems%2Cunix-linux">WA00612</reference>
<reference href="https://public.cyber.mil/stigs/downloads/?_dl_facet_stigs=operating-systems%2Cunix-linux">RHEL-07-WA00612</reference>
```

jan-cerny: @matejak Actually, a lot if items are duplicated, once with the RHEL-07- prefix, and once without that prefix. 

Try this:
```
grep "https://public.cyber.mil/stigs/downloads/?_dl_facet_stigs=operating-systems%2Cunix-linux" build/ssg-rhel7-ds.xml
```

Is it intentional?

jan-cerny:
@matejak Do you have any answers? Could you also resolve the conflicting files, please?

linux_os/guide/services/http/securing_httpd/httpd_configure_log_format/rule.yml
matejak: Thanks for noticing, the duplication is a bug - there should be only IDs with prefixes.

jan-cerny:
@matejak I find confusing that RHEL6 STIG IDs are defined as eg. `stigid@rhel6: RHEL-06-000247` but the RHEL7 STIG IDs are defined as eg. ` stigid@rhel7: "021300"` and the RHEL-07- prefix is added automatically. Why do we have this difference? Can it be unified?

redhatrises:
> @matejak I find confusing that RHEL6 STIG IDs are defined as eg. `stigid@rhel6: RHEL-06-000247` but the RHEL7 STIG IDs are defined as eg. ` stigid@rhel7: "021300"` and the RHEL-07- prefix is added automatically. Why do we have this difference? Can it be unified?

This is actually done in the local XSLT transforms file.

matejak:
It has just been unified in the latest commit.
@redhatrises This PR aims to take over some of the build process part that has been carried out by XSLT, so now a `rule.yml` has to have a qualified STIG ID, and a compiled rule file (that can be found in `build/<product>/rules/<rule-id>.yml`, e.g. `build/rhel6/rules/sysctl_net_ipv4_conf_all_log_martians.yml`) contains `stigid` or `CCE` that is unqualified, but specific to the product which is being compiled.

jan-cerny:
@matejak LGTM

PR #4067, opened 2019-03-08
[WIP] Added remediations for rhel8 audit rules which need ordering

matusmarhefka:
#### Description:

* Remediations added by this commit need to configure audit
  rules in the `/etc/audit/rules.d/` and preserve their order.
* The remediations are cherrypicking lines from the
  `/usr/share/doc/audit/rules/30-ospp-v42.rules` master file which
  belongs to the `audit` package and add them into a file in the
  `/etc/audit/rules.d/` folder preserving the specific order from
  the master file.


redhatrises:
IMO, This should be shared across fedora, OL, RHEL7, etc. Also, we should be not depending on and grabbing the items from a file that is in a different package.

matusmarhefka:
Closing this, we have decided to solve ordering differently.

PR #4105, opened 2019-03-09
rpm_verify_hashes - changes in config files can be expected

yuumasato:
#### Description:

- Ignore changes in RPM's config files

#### Rationale:

- This should help reduce false positives.

Still, some files make the rule fail.
For example, remediation for `require_singleuser_auth` edits `/usr/lib/systemd/system/rescue.service` and breaks its "integrity".
Not sure if `/usr/lib/systemd/system/rescue.service` can be considered a config file.
After remediation from fresh install of RHEL8.
```
[root@localhost ~]# rpm -Va | grep '..5'
S.5....T.  c /var/lib/unbound/root.key
S.5....T.  c /etc/profile
S.5....T.  c /etc/audit/audit-stop.rules
S.5....T.  c /etc/dnf/dnf.conf
S.5....T.  c /etc/security/pwquality.conf
S.5....T.  c /etc/pki/tls/openssl.cnf
S.5....T.  c /etc/issue
S.5....T.  c /etc/rsyslog.conf
S.5....T.  c /etc/ssh/sshd_config
S.5....T.  c /etc/login.defs
S.5....T.  c /etc/pam.d/password-auth
S.5....T.  c /etc/pam.d/system-auth
S.5....T.  c /etc/sysctl.conf
S.5....T.    /usr/lib/systemd/system/rescue.service
S.5....T.  c /etc/crypto-policies/config
```

### DIFF
--- a/linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_hashes/oval/shared.xml
+++ b/linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_hashes/oval/shared.xml
@@ -36,7 +36,7 @@
   </linux:rpmverifyfile_object>
   <linux:rpmverifyfile_state id="state_files_fail_md5_hash" version="1" operator="AND">
     <linux:md5_differs>fail</linux:md5_differs>
-    <!-- <linux:configuration_file datatype="boolean">false</linux:configuration_file> -->
+    <linux:configuration_file datatype="boolean">false</linux:configuration_file>
     <!-- <linux:documentation_file datatype="boolean">false</linux:documentation_file> -->
     <!-- <linux:ghost_file datatype="boolean">false</linux:ghost_file> -->
     <!-- <linux:license_file datatype="boolean">false</linux:license_file> -->
--- a/linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_hashes/oval/shared.xml
+++ b/linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_hashes/oval/shared.xml
@@ -38,7 +38,7 @@
     <linux:md5_differs>fail</linux:md5_differs>
     <linux:configuration_file datatype="boolean">false</linux:configuration_file>
     <!-- <linux:documentation_file datatype="boolean">false</linux:documentation_file> -->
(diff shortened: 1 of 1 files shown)


angystardust:
Hi @yuumasato We should definitely also exclude the "ghost" files from the rpm -V output to minimize false positive results.
For reference: 

> The --noghost option has been added to the rpm -V command.  If used with this option, rpm -V verifies only the non-ghost files that were altered, which helps diagnose system problems. (BZ#1395818) 
> The parameter behavior has been fixed in RHEL 7.6 GA:
> https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/7.6_release_notes/new_features_installation_and_booting
> https://access.redhat.com/errata/RHEA-2018:3250

jan-cerny:
@angystardust Probably yes, nice catch.

However it still will not solve all the issues. The aforementioned example `/usr/lib/systemd/system/rescue.service` is neither config nor ghost file.

angystardust:
Hi @jan-cerny in my testing env based on a RHEL7.5, the following command produces no output:
`rpm --noghost --noconfig -Va | grep '..5'`

redhatrises:
> Hi @yuumasato We should definitely also exclude the "ghost" files from the rpm -V output to minimize false positive results.
> For reference:
> 
> > The --noghost option has been added to the rpm -V command.  If used with this option, rpm -V verifies only the non-ghost files that were altered, which helps diagnose system problems. (BZ#1395818)
> > The parameter behavior has been fixed in RHEL 7.6 GA:
> > https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/7.6_release_notes/new_features_installation_and_booting
> > https://access.redhat.com/errata/RHEA-2018:3250

If the ghost files are non-configuration files, this is a problem, and they should not be excluded at all.

yuumasato:
I wonder what check is done to determine a change in a `ghost` file.
Just its existence can be considered a change?

yuumasato:
> If the ghost files are non-configuration files, this is a problem, and they should not be excluded at all.

Good point.

angystardust:
@yuumasato According to my understanding, ghost files are files whose contents are not included in the package payload so there's no way for rpm to verify their integrity

yuumasato:
I am confused by the `config` and `ghost` and `md5_differs`, I'll look into tests for them.

PR #4108, opened 2019-03-11
Assign RHEL8 CCE to OSPP rules which were missed during last CCE batc…

ggbecker:
…h assignment.

#### Description:

- Assign CCE to OSPP rules which don't have it assigned for RHEL8

#### Rationale:

- Make rules from OSPP profile compliant with NIST requirements regarding CCE identifiers.


yuumasato:
LGTM, ack when Jenkins finishes.

Issue #4001, opened 2019-02-27
SI-6 a | CCI-002696 | SRG-OS-000445-VMM-001780 |   | The VMM must verify correct operation of all security functions. | 


shawndwells:
SI-6 a | CCI-002696 | SRG-OS-000445-VMM-001780 |   | The VMM must verify correct operation of all security functions. | 

Without verification of the security functions, security functions may not operate correctly and the failure may go unnoticed. Security function is defined as the hardware, software, and/or firmware of the VMM responsible for enforcing the system security policy and supporting the isolation of code and data on which the protection is based. Security functionality includes, but is not limited to, establishing system accounts, configuring access authorizations (i.e., permissions, privileges), setting events to be audited, and setting intrusion detection parameters. |   | Verify the VMM verifies correct operation of all security functions.If it does not, this is a finding. | Configure the VMM to verify correct operation of all security functions. | CAT II
-- | -- | -- | -- | -- | -- | -- | -- | -- | --

PR #4197, opened 2019-03-13
Virt mapping

redhatrises:
More VMM SRG mappings.


redhatrises:
> `SRG-OS-000445-VMM-001780` (#4001) is a weird requirement about verifying the _functionality_ of various security features. Not sure this maps cleanly to SELinux enablement since the control is about functional testing (akin to what's done in common criteria).
> 
> Noticed VMWare doesn't have this control selected in their baselines.
> 
> What do you think about mapping this to installed_os_is_rhel/rhv since common criteria validates functional testing?

@shawndwells was mapping this one to some degree based on the NIST and DISA identifiers. Maybe those are wrong?

shawndwells:
On 3/14/19 3:40 PM, Gabe Alford wrote:
> @shawndwells <https://github.com/shawndwells> was mapping this one to 
> some degree based on the NIST and DISA identifiers. Maybe those are wrong?

Could be. Liberties are often taken when mapping NIST controls, 
creatively mapping them to support/justify configuration rules.

Do you think it makes sense to remove for now, and ask Ted to ping DISA 
about how to interpret this control?

redhatrises:
> On 3/14/19 3:40 PM, Gabe Alford wrote: @shawndwells <https://github.com/shawndwells> was mapping this one to some degree based on the NIST and DISA identifiers. Maybe those are wrong?
> Could be. Liberties are often taken when mapping NIST controls, creatively mapping them to support/justify configuration rules. Do you think it makes sense to remove for now, and ask Ted to ping DISA about how to interpret this control?

@shawndwells removed `SRG-OS-000445-VMM-001780` so that we can figure it out later.

shawndwells:
@redhatrises works for me!

Issue #3823, opened 2019-02-22
table to visualize maturity of rules


shawndwells:
Something to visualize maturity/completeness of a given rule. For example:

| Rule | OVAL | Bash | Ansible | Test Suite |
| --- | --- | --- | --- | --- |
| rule1 | X | - | X | - |
| rule 2 | X | X | X | X |
| rule 3 | - | - | X | X |
| rule 4 | X | - | - | - |

Would allow easy visualization of what rules need what kinds of improvements.

Something like...
MVP: HTML table generated during build
v1: Build auto publishes to a GitHub page after every PR, ensuring content dynamic/auto updated

Random features:
- clicking on a rule would bring you to GitHub URL of where rule lives in the code, so you don't have to hunt for it

jan-cerny:
I think `profile_stats.py` could be reused for this.

shawndwells:
What is profile_stats? Looks like it's used in build scripts, but no docs mention it:

`````
$ grep -rin "profile_stats" *
build-scripts/profile_stats.py:95:        ret.append(benchmark.show_profile_stats(args.profile, args))
build-scripts/profile_stats.py:102:                ret.append(benchmark.show_profile_stats(profile, args))
cmake/SSGCommon.cmake:612:        COMMAND env "PYTHONPATH=$ENV{PYTHONPATH}" "${PYTHON_EXECUTABLE}" "${SSG_BUILD_SCRIPTS}/profile_stats.py" --benchmark "${CMAKE_BINARY_DIR}/ssg-${PRODUCT}-xccdf.xml" --profile all
cmake/SSGCommon.cmake:619:        COMMAND env "PYTHONPATH=$ENV{PYTHONPATH}" "${PYTHON_EXECUTABLE}" "${SSG_BUILD_SCRIPTS}/profile_stats.py" --benchmark "${CMAKE_BINARY_DIR}/ssg-${PRODUCT}-xccdf.xml"
`````

shawndwells:
Tried running, no dice:

$ ./profile_stats.py
env: python2: No such file or directory

$ ./profile_stats.py --help
env: python2: No such file or directory

$ python --version
Python 2.7.10

$ /usr/bin/env python
Python 2.7.10 (default, Aug 17 2018, 19:45:58)
[GCC 4.2.1 Compatible Apple LLVM 10.0.0 (clang-1000.0.42)] on darwin
Type "help", "copyright", "credits" or "license" for more information.

[...]

PR #4199, opened 2019-03-15
Mention profile_stats.py in Developer Guide

jan-cerny:
#### Description:

Mention profile_stats.py in Developer Guide

#### Rationale:

In https://github.com/ComplianceAsCode/content/issues/3823 we have noticed that profile_stats.py isn't mentioned anywhere even if it's a very useful script.


yuumasato:
Thank you, LGTM.

shawndwells:
Thanks for adding this into the docs!

PR #4200, opened 2019-03-15
Avoid duplication of audit rules

### LINKED ISSUE #4104: Applying `stig-rhel7-disa profile on RHEL7 creates duplicate rules
#### Description of problem:
After running oscap with the  `stig-rhel7-disa` profile on RHEL7, duplicate rules are created in `/etc/audit/rules.d/privileges.rules`.  Repeated execution of oscap causes additional duplicate rules to be appended to the file.

v0.1.43

Red Hat Enterprise Linux Server 7.6

#### Steps to Reproduce:

1.  Download SCAP Security Guide content v0.1.43

2.  Run command to re-mediate system with `stig-rhel7-disa` profile:

`oscap xccdf eval --remediate --profile stig-rhel7-disa ssg-rhel7-ds.xml`

3.  View created rules file `/etc/audit/rules.d/privileges.rules`

4.  Repeat command in Step 2

5.  Repeat Step 3

#### Actual Results:
The `crontab` rule was extracted as an example.  See below:

After initial oscap execution -
```
-a always,exit -F path=/usr/bin/crontab -F perm=x -F auid>=1000 -F auid!=unset -F key=privileged
-a always,exit -F path=/usr/bin/crontab -F perm=x -F auid>=1000 -F auid!=unset -k privileged
```
After second oscap execution -
```
-a always,exit -F path=/usr/bin/crontab -F perm=x -F auid>=1000 -F auid!=unset -F key=privileged
-a always,exit -F path=/usr/bin/crontab -F perm=x -F auid>=1000 -F auid!=unset -k privileged
-a always,exit -F path=/usr/bin/crontab -F perm=x -F auid>=1000 -F auid!=unset -k privileged
```

#### Expected Results:
Only one rule should be created per STIG requirement and additional execution of oscap should not create duplicate rules.

The duplicate rules also causes `augenrules --load` to fail with the following error:
```
Error sending add rule data request (Rule exists)
There was an error in line 106 of /etc/audit/audit.rules
```

yuumasato:
#### Description:

- Update `template_OVAL_audit_rules_privileged_commands` to fail if a duplicate rule is found.
- Fix bug in `template_BASH_audit_rules_privileged_commands`
  -  the value of `ARCH` set in previous remediation was being used
- Fix bug in `perform_audit_rules_privileged_commands_remediation.sh`
  - the function wasn't identifying correctly if an audit rule was already present in a previously checked file
- Adds tests for the scenarios 

#### Rationale:

- Fixes #4104

### DIFF
--- a/shared/bash_remediation_functions/perform_audit_rules_privileged_commands_remediation.sh
+++ b/shared/bash_remediation_functions/perform_audit_rules_privileged_commands_remediation.sh
@@ -71,7 +71,7 @@ declare -a sbinaries_to_skip=()
 for sbinary in "${privileged_binaries[@]}"
 do
 
-	# Check if this sbinary wasn't already handled in some of the previous iterations
+	# Check if this sbinary wasn't already handled in some of the previous sbinary iterations
 	# Return match only if whole sbinary definition matched (not in the case just prefix matched!!!)
 	if [[ $(sed -ne "\|${sbinary}|p" <<< "${sbinaries_to_skip[*]}") ]]
 	then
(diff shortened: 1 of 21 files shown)


jan-cerny:
I have cheked that with you patch the augenrules doesn't fail to load anymore and the privileged commands are there only once. That's awesome!

PR #4201, opened 2019-03-18
Removed pattern matching when searching for .shosts file.

matejak:
This ensures consistency with rule description.


yuumasato:
LGTM, the rule is looking for a specific file.

PR #4204, opened 2019-03-19
File regex permissions find use regex

yuumasato:
#### Description:

- Specify in `template_ANSIBLE_file_regex_permissions` that `patterns` is a regular expression.
- Add a test scenario for multiple ssh keys

#### Rationale:

- The pattern in defined in CSV file is a regular expression
https://github.com/ComplianceAsCode/content/blob/95a815314a1d905bef10142445fee3aaa3f2cbc8/rhel7/templates/csv/file_dir_permissions.csv#L9

### DIFF
--- a/shared/templates/template_ANSIBLE_file_regex_permissions
+++ b/shared/templates/template_ANSIBLE_file_regex_permissions
@@ -7,6 +7,7 @@
   find:
     paths: "{{{ FILEPATH }}}"
     patterns: "{{{ FILENAME }}}"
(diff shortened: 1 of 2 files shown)


jan-cerny:
Nice catch! I have checked that in the built RHEL7 datastream the Ansible remediations for file_permissions_sshd_private_key and file_permissions_sshd_pub_key look OK. Note that if `use_regex` is not set to yes, it understands the pattern as a shell glob :slightly_smiling_face:

PR #2189, opened 2017-07-27
Disable remediation for "repo_gpgcheck=1"

jan-cerny:
This makes the remediation ineffective, so that it won't break
yum. A warning and a comment is added to explain the reasons.

PR #4206, opened 2019-03-20
Remove ensure_gpgcheck_repo_metadata check from rhel profiles

shawndwells:
This was removed from RHEL 8 already, appears to be a bug in RHEL 7 content.

@dahaic we'll want to ensure this makes downstream rhel 7.7 release.

### DIFF
--- a/rhel7/profiles/hipaa.profile
+++ b/rhel7/profiles/hipaa.profile
@@ -88,6 +88,7 @@ selections:
     - ensure_redhat_gpgkey_installed
     - ensure_gpgcheck_globally_activated
     - ensure_gpgcheck_never_disabled
+    - ensure_gpgcheck_repo_metadata
     - ensure_gpgcheck_local_packages
     - grub2_audit_argument
     - service_auditd_enabled
--- a/rhel7/profiles/hipaa.profile
+++ b/rhel7/profiles/hipaa.profile
@@ -88,7 +88,6 @@ selections:
     - ensure_redhat_gpgkey_installed
     - ensure_gpgcheck_globally_activated
     - ensure_gpgcheck_never_disabled
-    - ensure_gpgcheck_repo_metadata
     - ensure_gpgcheck_local_packages
     - grub2_audit_argument
     - service_auditd_enabled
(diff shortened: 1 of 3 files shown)


yuumasato:
> This was removed from RHEL 8 already, appears to be a bug in RHEL 7 content.

Hello @shawndwells, can you be more clear why do you consider this a bug?

shawndwells:
On 3/26/19 7:58 AM, Watson Yuuma Sato wrote:
> Hello @shawndwells <https://github.com/shawndwells>, can you be more 
> clear why do you consider this a bug?


Rule was removed from US government baselines ages ago (e.g. DISA STIG), 
but looks like that change wasn't updated across all the profiles.

redhatrises:
> On 3/26/19 7:58 AM, Watson Yuuma Sato wrote: Hello @shawndwells <https://github.com/shawndwells>, can you be more clear why do you consider this a bug?
> Rule was removed from US government baselines ages ago (e.g. DISA STIG), but looks like that change wasn't updated across all the profiles.

It was removed from the STIG because it was dropped from the STIG but not other profiles. Ultimately, the question should be if this rule should be added and enforced.

yuumasato:
> Ultimately, the question should be if this rule should be added and enforced.

And I believe the answer is the policy to which the profile is aligning, and I understand there can be exceptions due to practical reasons.

If the policy requires such rule, will it be easier to deal with the finding when rule `ensure_gpgcheck_repo_metadata` is removed?

I remember there was a warning in the rule, added in #2189.
It seem that it fell off the cart during conversion to yml rules.
See, https://github.com/ComplianceAsCode/content/pull/2592/commits/fcfb389319c578d91d3147f368bf53161f032108#diff-80a4dbecd637e895b68998ee902d6a7fL295 and https://github.com/ComplianceAsCode/content/commit/9220c943ab010ab131664d4d658f62c3fb97b6e3

Does re-adding the warning makes things easier?

shawndwells:
These profiles inherited it only because it was mandated by DISA back in the day. It's no longer required, suggest dropping from all profiles.

For the US Gov `ospp` profile we have domain to make that determination ourselves. 

Atif Chaugtai leads the Healthcare vertical. I'll ask him about if he wants HIPAA baseline to continue having this rule.

In the mean time, updated this PR to only modify the OSPP profile (as to get this merged now).

This is one of the 'must have' patches -- need to ensure RHEL 7.7 doesn't end up continuing to be broken over this.

jan-cerny:
@shawndwells OK, thanks for clarification!

Issue #4195, opened 2019-03-13
Rule ID used as task name in generated per-rule playbooks


matejak:
The playbook name uses rule ID, instead of more meaningful title. Consider the `sshd_disable_root_login` remediation:

```
- name: sshd_disable_root_login
  hosts: '@@HOSTS@@'
  become: true
  tags:
  - NIST-800-53-IA-2
  ...
  - low_complexity
  tasks:
  - name: Disable SSH Root Login
    lineinfile:
      create: true
      ...
```

PR #4207, opened 2019-03-21
Use rule title instead of rule ID as a playbook name

jan-cerny:
#### Description:
When rule.yml is parsed it will add the rule title as a comment into snippet in `/build/<product>/fixes/ansible/`. This comment will be parsed while playbooks are created and used as the playbook title.

#### Rationale:
Fixes: #4195

### DIFF
--- a/ssg/build_remediations.py
+++ b/ssg/build_remediations.py
@@ -25,7 +25,7 @@ REMEDIATION_TO_EXT_MAP = {
 FILE_GENERATED_HASH_COMMENT = '# THIS FILE IS GENERATED'
 
 REMEDIATION_CONFIG_KEYS = ['complexity', 'disruption', 'platform', 'reboot',
-                           'strategy']
+                           'strategy', 'title']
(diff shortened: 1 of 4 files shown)


ssg/build_remediations.py
```diff
@@ -25,7 +25,7 @@
 FILE_GENERATED_HASH_COMMENT = '# THIS FILE IS GENERATED'
 
 REMEDIATION_CONFIG_KEYS = ['complexity', 'disruption', 'platform', 'reboot',
```
matejak: I don't like this so much - this PR is mixing remediation metadata and rule data. I would prefer re-reading the rule YAML from the processed rules directory, or to introduce the rule title as a new metadata attribute.

jan-cerny: Aha cool

jan-cerny:
Replaced by https://github.com/ComplianceAsCode/content/pull/4211

Issue #4192, opened 2019-03-13
non-deterministic playbook tag order


jharting:
Running e.g. `./build_product rhel7` repeatedly I observe play tags ordered differently each time.

```
   hosts: '@@HOSTS@@'
   become: true
   tags:
-  - NIST-800-53-AC-6
+  - restrict_strategy
+  - sshd_use_priv_separation
   - DISA-STIG-RHEL-07-040460
   - low_disruption
-  - NIST-800-171-3.1.12
-  - medium_severity
-  - CCE-80223-1
   - NIST-800-53-AC-17(b)
-  - sshd_use_priv_separation
   - low_complexity
-  - restrict_strategy
+  - NIST-800-53-AC-6
+  - CCE-80223-1
+  - NIST-800-171-3.1.12
+  - medium_severity
   tasks:
   - name: Enable use of Privilege Separation
     lineinfile:
```

Observed with 976dbd1eb47b84fe31536cc407d63a809a87bbaa

matejak:
This is interesting, is it python2 or python3-based build?

jharting:
@matejak Python 2.7.15

PR #4208, opened 2019-03-21
Order tags in playbooks

jan-cerny:
#### Description:
If we sort the list, the order of tags won't be random.
Also updates the test and fixes a bug in the test.

#### Rationale:
Fixes: #4192


matejak:
Nice, I was thinking of forcing `OrderedDict`, but this approach is much more robust!

jan-cerny:
@matejak No, OrderedDict isn't involved here. The list of playbook tags is created as a merge of all tags from all tasks in the given playbooks. To prevent duplicated tags it's implemented as an union of sets. Sets are unordered, so their conversion to a list might be random.

PR #4209, opened 2019-03-21
Simplify CMake macro _ssg_build_remediations_for_language

jan-cerny:
#### Description:
The list of dependencies doesn't have to be generated at runtime because they're located all in the single directory in the build tree. Moreover, it slows down cmake.

#### Rationale:

Fixes: #3816

### DIFF
--- a/cmake/SSGCommon.cmake
+++ b/cmake/SSGCommon.cmake
@@ -197,35 +197,19 @@ endmacro()
 
 macro(_ssg_build_remediations_for_language PRODUCT LANGUAGES)
     set(BUILD_REMEDIATIONS_DIR "${CMAKE_CURRENT_BINARY_DIR}/fixes_from_templates")
-
-    execute_process(
-        COMMAND env "PYTHONPATH=$ENV{PYTHONPATH}" "${PYTHON_EXECUTABLE}" "${SSG_BUILD_SCRIPTS}/generate_from_templates.py" --languages ${LANGUAGES} --input "${CMAKE_CURRENT_SOURCE_DIR}/templates" "${SSG_SHARED}/templates" --output "${BUILD_REMEDIATIONS_DIR}" "${BUILD_REMEDIATIONS_DIR}/shared" --shared "${SSG_SHARED}" --build-config-yaml "${CMAKE_BINARY_DIR}/build_config.yml" --product-yaml "${CMAKE_CURRENT_SOURCE_DIR}/product.yml" list-inputs
-        OUTPUT_VARIABLE LANGUAGE_REMEDIATIONS_DEPENDS_STR
-    )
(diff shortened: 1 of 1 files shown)


cmake/SSGCommon.cmake
```diff
@@ -197,44 +197,27 @@ endmacro()
...
     )
     foreach(LANGUAGE ${LANGUAGES})
       set(ALL_FIXES_DIR "${CMAKE_CURRENT_BINARY_DIR}/fixes/${LANGUAGE}")
 
       add_custom_command(
           OUTPUT "${ALL_FIXES_DIR}"
           COMMAND env "PYTHONPATH=$ENV{PYTHONPATH}" "${PYTHON_EXECUTABLE}" "${SSG_BUILD_SCRIPTS}/combine_remediations.py" --resolved-rules-dir "${CMAKE_CURRENT_BINARY_DIR}/rules" --build-config-yaml "${CMAKE_BINARY_DIR}/build_config.yml" --product-yaml "${CMAKE_CURRENT_SOURCE_DIR}/product.yml" --remediation-type "${LANGUAGE}" --output-dir "${ALL_FIXES_DIR}" "${BUILD_REMEDIATIONS_DIR}/shared/${LANGUAGE}" "${BUILD_REMEDIATIONS_DIR}/${LANGUAGE}"
-          DEPENDS ${LANGUAGE_REMEDIATIONS_DEPENDS}
-          DEPENDS ${LANGUAGE_REMEDIATIONS_OUTPUTS}
+          DEPENDS "${BUILD_REMEDIATIONS_DIR}"
```
yuumasato: Before, it depended on a list of files, now it depends on a single directory.
Can't this cause concurrence problems? As soon as the directory is created, it has all dependencies to start combining remediations.
For each languagne, the`combine_remediations.py` should only start after its corresponding `generate_from_templates.py` finishes.

jan-cerny: I need to check it.

On the other hand, generate_from_templates.py is called only once and generates all languages at single run.

jan-cerny: I have build the content using `cmake -G Ninja .. && ninja -j4` both with and without my patch and I compared RHEL6,7,8 and Fedora datastreams using meld. The only differences are:
* timestamps (obviously)
* different order of check-export child elements  within a single check element. I think that's OK because check-export can have any order.

yuumasato: OK, probably there won't be concurrency problems, this command depends on `generate-internal-language-remedations-${PRODUCT}`, which is the command that generates from templates.

Looking again,`DEPENDS "${BUILD_REMEDIATIONS_DIR}"` was only tracking changes in input files. I think this line can be removed.

jan-cerny: @yuumasato OK

yuumasato:
Looks good, this is also working fine with Ninja.

PR #4210, opened 2019-03-21
Simplify build_rule_playbooks.py CLI

### LINKED ISSUE #4034: Some arguments in build_rule_playbooks.py don't have to be required
#### Description of problem:
As pointed out in https://github.com/ComplianceAsCode/content/pull/3824,
If we know the product YAML, we think that we also know where to look for the input directory - it is `build/<product-that-we-know>/fixes/ansible`. This means that we don't have to require `--input-dir`, because we know the default.
The same for `--output-dir`

upstream head as of 28th February 2019

all

#### Steps to Reproduce:

1. build_rule_playbooks.py --help
2.
3.
4.

#### Actual Results:
 `--input-dir` and `--output-dir` are required options

#### Expected Results:
 `--input-dir` and `--output-dir` aren't required options and their values are inferred from product name if they're not provided

Please investigate if the aforementioned options are needed if the script is used stand-alone.

jan-cerny:
#### Description:
It isn't recommended to require `--input-dir` and `--output-dir` and `--resolved-rules-dir`, because in their typical usage in ComplianceAsCode build system in CMake, their value can be computed from the CMAKE_SOURCE_DIR and the ID of the product we want to build Playbooks for.
We will keep `--input-dir` and `--output-dir`  and `--resolved-rules-dir` for users who want to use the script stand-alone, eg. for development purposes.

#### Rationale:
Fixes #4034.

### DIFF
--- a/build-scripts/build_rule_playbooks.py
+++ b/build-scripts/build_rule_playbooks.py
@@ -2,31 +2,41 @@
 
 import argparse
 import ssg.playbook_builder
+import os
 
 
 def parse_args():
     p = argparse.ArgumentParser()
     p.add_argument(
-        "--input-dir", required=True, dest="input_dir",
+        "--input-dir",
         help="Input directory that contains all Ansible remediations "
         "snippets for the product we are building. "
-        "e.g. ~/scap-security-guide/build/fedora/fixes/ansible"
+        "e.g. ~/scap-security-guide/build/fedora/fixes/ansible. "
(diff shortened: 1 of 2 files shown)


jan-cerny:
I have resolved the conflicts and rebased me on top of master's head and updated the PR description accordingly.

yuumasato:
Looks good, thank you! :)

PR #4211, opened 2019-03-22
Use rule title instead of rule ID as a playbook name

jan-cerny:
#### Description:
The rule title will be obtained from preprocessed rule YAMLs located at `/build/<product>/rules`.

Replaces #4207

#### Rationale:
Fixes: #4195


matejak:
This PR indeed fixes the issue. Good job!

Issue #4194, opened 2019-03-13
No metadata in generated per-rule playbooks


matejak:
The metadata in form of comments s.a.
```
# reboot = false
...
```
are not present in generated per-rule playbooks.

PR #4212, opened 2019-03-22
Add remediation metadata to the generated playbooks

jan-cerny:
#### Description:
This will add the remedaition metadata in form of comments s.a.
```
# reboot = false
...
```
because are not present in generated per-rule playbooks.

Also adds reboot tag to generated snippets.

#### Rationale:
Fixes: #4194

### DIFF
--- a/ssg/ansible.py
+++ b/ssg/ansible.py
@@ -101,7 +101,11 @@ class AnsibleRemediation(object):
         if "disruption" in self.config:
             tags.append("{0}_disruption".format(self.config["disruption"]))
         if "reboot" in self.config:
-            tags.append("{0}_reboot".format(self.config["reboot"]))
+            if self.config["reboot"] == "true":
+                reboot_tag = "reboot_required"
+            else:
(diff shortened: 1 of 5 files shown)


ssg/ansible.py
```diff
@@ -100,6 +100,8 @@ def update_tags_from_config(self, to_update):
             tags.append("{0}_complexity".format(self.config["complexity"]))
         if "disruption" in self.config:
             tags.append("{0}_disruption".format(self.config["disruption"]))
+        if "reboot" in self.config:
+            tags.append("{0}_reboot".format(self.config["reboot"]))
```
matejak: `false_reboot` / `true_reboot` are somehow confusing, what about `reboot_required` / `no_reboot_needed`?

jan-cerny: yes, thank you for the suggestion

jan-cerny: @matejak changed

matejak:
Works great, merging!

PR #4213, opened 2019-03-22
Test suite updates to fix ssh pubkey access to the tested VM

matusmarhefka:
#### Description:

- Fixed the code exporting `SSH_ADDITIONAL_OPTIONS` which was not working correctly and exported only one ssh option omitting others.
- `unselect_rules_list` file updated with rules which might break the ssh access to VM.


tests/unselect_rules_list
```diff
@@ -3,23 +3,11 @@ xccdf_org.ssgproject.content_rule_ensure_gpgcheck_local_packages
...
-xccdf_org.ssgproject.content_rule_grub2_uefi_password
-xccdf_org.ssgproject.content_rule_installed_OS_is_certified
-xccdf_org.ssgproject.content_rule_install_antivirus
-xccdf_org.ssgproject.content_rule_install_hids
-xccdf_org.ssgproject.content_rule_install_mcafee_antivirus
-xccdf_org.ssgproject.content_rule_install_mcafee_cma_rt
-xccdf_org.ssgproject.content_rule_install_mcafee_hbss_accm
-xccdf_org.ssgproject.content_rule_install_mcafee_hbss_hips
-xccdf_org.ssgproject.content_rule_install_mcafee_hbss_pa
-xccdf_org.ssgproject.content_rule_mcafee_antivirus_definitions_updated
```
yuumasato: @matusmarhefka Was the removal os this rules expected as well? They are not mentined in PR description.

matusmarhefka: @yuumasato Yes it was, these rules have no remediation, therefore they can remain selected.

yuumasato: @matusmarhefka The rules at the top of the chunk have remediation, like the ones related to `rule_rpm_verify_hashes`, permissions and ownership. Should they be removed as well?

matusmarhefka: Yes, there should be only rules which might break the access to the machine for the test suite or destructive rules (which influence other rules remediations). We want to test rpm_verify family of rules. Of course, we can update this list in the future, so if you think that some rules should be added/removed from the list, please open a PR and we can discuss there.

yuumasato: Just wanted to be sure. :)

Issue #3830, opened 2019-02-26
Introduce proper unified remediation classes


matejak:
Unify the `AnsibleRemediation` class in `ssg.ansible` and the one in `ssg.build_remediations`. The existence of those two classes presents
* additional complexity in the build system, and
* module-level dependency cycle.

There should be one class that will do everything that can be requested from an Ansible remediation, while sharing code with other remediation types as possible.

This issue is tackled, but not solved by #3827.

Issue #3831, opened 2019-02-26
Resolve @...@ macros in remediations using jinja.


matejak:
The legacy XSLT operations are used to resolve:
* `@CCENUM@`
* `@RULE_ID@`
* `@RULE_TITLE@`

macros. The source snippets should be rewritten to use Jinja, and the code should be updated, so it can expand those macros.

This is made possible by #3827, as it brings information from rule YAMLs together with remediation snippets during their processing stage.

PR #4216, opened 2019-03-25
Replace legacy @RULE_ID@ AND @RULE_TITLE@ substitutions with jinja

matejak:
This PR unifies the Ansible functionality used to

* generate general remediations output, and
* add rule-specific metadata to the Ansible output.

Moreover, it adds functionality for expanding templates the jinja way.

In order to accomplish this, the existing code (and it's respective tests) had to be moved around, so there are no circular dependencies.

Fixes: #3830, #3831

### DIFF
--- a/build-scripts/combine_remediations.py
+++ b/build-scripts/combine_remediations.py
@@ -62,31 +62,16 @@ def main():
     # given fix_name is chosen to replace newer fix_names
     remediation_cls = remediation.REMEDIATION_TO_CLASS[args.remediation_type]
 
+    rule_id_to_remediation_map = collect_fixes(
+        product, guide_dir, args.fix_dirs, args.remediation_type)
+
     fixes = dict()
-    for fixdir in args.fix_dirs:
-        if os.path.isdir(fixdir):
-            for filename in os.listdir(fixdir):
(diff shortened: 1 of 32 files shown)


jan-cerny:
I have noticed that some rules, eg.` sysctl_kernel_randomize_va_space` now have a new tag in their Ansible tasks, eg. `DISA-STIG-RHEL-08-040201`. Is that expected?

matejak:
The problem is that the STIG ID is not always product-qualified, we have `grep -R --include rule.yml stigid: linux_os/ | wc -l` == 67 rules that have unqualified STIG IDs in the `linux_os` tree. At the same time, we have 445 rules that have qualified STIG IDs (some of them have both).
However, this change is not expected, so I have pushed a commit fixing that.

jan-cerny:
Great! However, I tried to build the RHEL 6 datastream with the latest update but I think this is also not expected: Before: `DISA-STIG-RHEL-06-RHEL-06-000256`, After: `DISA-STIG--RHEL-06-000256` Notice the double dash in the latter tag.

Regarding the substitutions in Ansible Playbooks, they work for me.

jan-cerny:
@matejak please resolve the conflicts

PR #4218, opened 2019-03-26
Use product IDs correctly in SSG_PRODUCT macros

jan-cerny:
#### Description:
Use product IDs correctly in SSG_PRODUCT macros

#### Rationale:
The product IDs are specified by the product directory name and the product entry in `product.yml`s. The SSG_PRODUCT macros should be SSG_PRODUCT_<PRODUCT_ID>, but this was not always true.  For example, the `rhosp13` product was controlled by SSG_PRODUCT_OSP13 (notice the missing RH in the macro name). 

This commit makes all macros consistently named using the product IDs.

The effect of this change is that users can now use `build_product` script to easily build also `rhosp13`, `eap6`, `fuse6`, SLE and Ubuntu products.


shawndwells:
Thanks for cleaning this up

PR #4224, opened 2019-03-27
Fix audit_rules_privileged_commands ansible remediation

redhatrises:
- Playbooks and roles where failing due to an undefined value when the audit_rules_privileged_commands ansible remediation was run in a container


shawndwells:
logic makes sense. thanks!

PR #4225, opened 2019-03-28
Add YUM installation checks to various rules

shawndwells:
Update checks to evaluate if YUM is installed before failing as some systems are readonly and may not have a package manager.


redhatrises:
@shawndwells can you provide some more context around this change and why it is needed?

shawndwells:
On 3/27/19 9:19 PM, Gabe Alford wrote:
>
> @shawndwells <https://github.com/shawndwells> can you provide some 
> more context around this change and why it is needed?
>

Some read-only systems do not have a package manager installed. For 
these distro spins, all the YUM rules come up as false findings (and 
broken remediations)

yuumasato:
Closing, as this should be solved #4248.

PR #4226, opened 2019-03-28
Make sure that Ansible auditing templates have defined attributes

redhatrises:
This is needed to not fail ansible rules in containers.


yuumasato:
Looks good and makes sense.

PR #4229, opened 2019-03-28
Update the RHEL8 profile

redhatrises:
- Update description as OSPP will meet more baseline requirements than just OSPP
- Add and enable some missing rules

### DIFF
--- a/rhel8/profiles/ospp.profile
+++ b/rhel8/profiles/ospp.profile
@@ -3,9 +3,34 @@ documentation_complete: true
 title: 'Protection Profile for General Purpose Operating Systems'
 
 description: |-
-    This profile reflects mandatory configuration controls identified in the
-    NIAP Configuration Annex to the Protection Profile for General Purpose
-    Operating Systems (Protection Profile Version 4.2).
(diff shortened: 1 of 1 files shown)


matejak:
Do we have a consensus over what the OSPP profile should be about? The last thing that I recall is that we want it to be the minimal profile that all other profiles will use as base. Has this changed?
If not, then adding rules to the profile requires more concrete justification.

shawndwells:
On 6/12/19 7:13 AM, Matěj Týč wrote:
> Do we have a consensus over what the OSPP profile should be about? The 
> last thing that I recall is that we want it to be the minimal profile 
> that all other profiles will use as base. Has this changed?
> If not, then adding rules to the profile requires more concrete 
> justification.



OSPP reflects settings required from the Operating System Protection 
Profile with the NIAP Configuration Annex.


AKA, it is the configuration baseline Red Hat will be putting through 
Common Criteria.

shawndwells:
Changes incorporated into https://github.com/ComplianceAsCode/content/pull/4446. Closing this PR.

PR #4230, opened 2019-03-28
Set pipefail for pipe usage in Ansible shell tasks

redhatrises:
- Ansible lint: [306] Shells that use pipes should set the pipefail option


yuumasato:
Looks good!

PR #4234, opened 2019-03-30
additional UBI prep

shawndwells:
Updates multiple rules to evaluate if associated RPM is installed before evaluating.

Needed because some distros/spins may not have shadow-utils.


yuumasato:
I think this will be easier if done through `xccdf:platform` and a CPE-OVAL for `shadow-utils`, the same we have done for `machine` `container` detection.

shawndwells:
Would definitely be cleaner!

The platform in question uses the RHEL server CPEs, and Redhat-release output is also RHEL server. Do you have suggestions on how this could be accomplished?

yuumasato:
> Do you have suggestions on how this could be accomplished?

It would be very similar to how this PR's approach, but on the XCCDF level, the "platform" would be package `shadow-utils` installed.

In this case, the rule would return `notapplicable`, instead of `pass`.

yuumasato:
I'll look into this CPE approach.

yuumasato:
Implemented it for `shadow-util`: https://github.com/ComplianceAsCode/content/pull/4241.
After merged it can be expanded to other packages...

shawndwells:
rolled into https://github.com/ComplianceAsCode/content/pull/4243

PR #4235, opened 2019-04-01
remove ensure_gpgcheck_repo_metadata from HIPAA profiles

shawndwells:
Double checked with Atif. ``ensure_gpgcheck_repo_metadata`` isn't needed for HIPAA profile either.


redhatrises:
ACK

PR #4239, opened 2019-04-03
Add profile diff tool

ggbecker:
#### Description:

- This PR introduces a new utility tool to compare rules from given two YAML profiles. As result it creates a new profile with rules that are contained exclusively in profile1. profile2 can also be a list of rules separated by new line.

#### Rationale:

- It can be used to create a new YAML profile containing only exclusive rules from profile1. This is useful when is needed to know what else a profile does beside the expected set of rules.


shawndwells:
Could be a neat tool! Could documentation also be added somewhere, perhaps the developer docs?

redhatrises:
Great job! Will be helpful in the project itself. Please merge this functionality into the profile_stats scripts. We are trying not to have so many different scripts to run to get info from profiles.

matejak:
We have some profile-parsing functionality in `ssg/build_yaml.py`, may be worth checking out too.

jan-cerny:
> Please merge this functionality into the profile_stats scripts. We are trying not to have so many different scripts to run to get info from profiles.

Would be cool, but this one compares 2 profiles and `profile_stats.py` works with 1 profile.

@ggbecker Would the merge to `profile_stats.py` be feasible?

ggbecker:
The use case is a bit different, profile_stats acts upon xccdf files. The proposed tool loads the profile yaml.

@jan-cerny It can be merged but I don't think that something can be reused from the profile_stats source code.

jan-cerny:
@ggbecker I see, then let's not merge them.

@ggbecker Please check if you can use `Profile` class from `ssg/build_yaml.py`. I think that the `Profile.from_yaml` static method can be used for parsing the YAML. Please also add a description to the developer guide because otherwise you script will be ignored by contributors.

ggbecker:
I managed to create new functions in build_yaml.py file so it makes sense to have this tool incorporated to the profile_stats.py. I will propose to change the name from profile_stats to profile_tool as it does more than just statistics, but for now I'll leave as it is for the sake of reviewing.

docs/manual/developer_guide.adoc
```diff
@@ -594,9 +594,18 @@ For example, to check which rules in RHEL8 OSPP profile are missing remediations
...
 ----
 
+The tool also can create diff between YAML profiles and generate a new one.
+
+For example, to generate a profile diff between two profiles, run this command:
+----
+$ ./build-scripts/profile_stats.py diff --profile1 rhel7/profiles/ospp.profile --profile2 rhel7/profiles/pci-dss.profile 
+----
+
+A new profile will be generated in the current directory with the name, in this case, ospp-compared_to-pci-dss.profile
```
jan-cerny: I would rephrase this, because I would expect an output similar to `diff` output, but in fact it generates a new profile containing only rules from profile1 that are not contained in profile2.

build-scripts/profile_stats.py
```diff
@@ -19,76 +20,124 @@ def parse_args():
...
                         help="Show all available statistics.")
-    parser.add_argument("--format", default="plain",
+    parser_stats.add_argument("--format", default="plain",
                         choices=["plain", "json", "csv"],
                         help="Which format to use for output.")
 
+    profile_diff_desc = \
+        "Compare selected/unselected rules from given profiles and " + \
+        "produce a new profile containing only rules from profile1 that " + \
+        "is not contained in profile2. Variables are not compared and " + \
```
jan-cerny: are not

docs/manual/developer_guide.adoc
ggbecker: I changed the approach of the feature to subtraction instead of diff.

jan-cerny:
thanks!

PR #4243, opened 2019-04-03
Create additional package CPEs

shawndwells:
- Creates package-based CPEs based on suggestion from @yuumasato ;
- Refactors SSH server checks


linux_os/guide/services/ssh/ssh_server/disable_host_auth/oval/shared.xml
```diff
@@ -7,32 +8,21 @@
...
-        definition_ref="package_openssh-server_removed" />
-      </criteria>
-      <criteria comment="sshd is installed and configured" operator="AND">
-        <extend_definition comment="sshd is required or requirement is unset"
-        definition_ref="sshd_required_or_unset" />
-        <extend_definition comment="rpm package openssh-server installed"
-        definition_ref="package_openssh-server_installed" />
-        <criterion comment="Check HostbasedAuthentication in /etc/ssh/sshd_config"
-        test_ref="test_sshd_hostbasedauthentication" />
-      </criteria>
```
yuumasato: @shawndwells Are you ok if a profile where `sshd` is "required" returns `notapplicable` instead of fail?

Read required as in https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/ssh/sshd_required.var.

Current profiles that **require** `sshd` to be installed are:
```console
$ grep -R "sshd_required" */profiles/*
ol7/profiles/stig-ol7-disa.profile:    - sshd_required=yes
rhel7/profiles/ospp.profile:    - sshd_required=yes
rhel7/profiles/rhelh-stig.profile:    - sshd_required=yes
rhel7/profiles/stig-rhel7-disa.profile:    - sshd_required=yes
rhv4/profiles/rhvh-stig.profile:    - sshd_required=yes
```

I'd love to get rid of `sshd_required_or_unset` and `sshd_not_required_or_unset`.

shawndwells: If ssh server isn't installed, the checks should be ``not applicable``.

IMHO our role isn't to determine if SSH is required. More that if SSH server is installed, it should be configured a certain way.

Updated profiles to remove sshd_required variable.

linux_os/guide/system/software/integrity/fips/sysctl_crypto_fips_enabled/rule.yml
```diff
@@ -52,4 +52,4 @@ warnings:
         See <b>{{{ weblink(link="http://csrc.nist.gov/groups/STM/cmvp/documents/140-1/1401vend.htm") }}}</b>
         for a list of FIPS certified vendors.
 
-platform: machine
```
yuumasato: @shawndwells is the intention of this change to substitute the way we detect a machine environment?
Currently we say that if it is not a container, it is a machine.

If a machine compiles its own kernel, would there be a kernel package?

linux_os/guide/system/software/integrity/fips/enable_fips_mode/rule.yml
```diff
@@ -59,4 +59,4 @@ warnings:
         See <b>{{{ weblink(link="http://csrc.nist.gov/groups/STM/cmvp/documents/140-1/1401vend.htm") }}}</b>
         for a list of FIPS certified vendors.
 
-platform: machine
+platform: kernel
```
jan-cerny: You can install the kernel package in a container. This rule probably still shouldn't be applicable in that case, because the kernel on the host should handle FIPS and the kernel in the container won't run.

shared/checks/oval/installed_env_has_openssh-server_package.xml
```diff
@@ -0,0 +1,37 @@
...
+  </linux:rpminfo_object>
+
+  <!-- everything else -->
+  <linux:rpminfo_test check="all" check_existence="at_least_one_exists" comment="system has package openssh-server installed" id="test_env_has_openssh-server_installed" version="1">
+    <linux:object object_ref="obj_env_has_openssh-server_installed" />
+  </linux:rpminfo_test>
+  <linux:rpminfo_object id="obj_env_has_openssh-server_installed" version="1">
+    <linux:name>openssh-server</linux:name>
+  </linux:rpminfo_object>
+
```
jan-cerny: I think you're disabling the Debian and Ubuntu profiles by this. The  Debian and Ubuntu profiles also contain rules related to SSH but your platform CPE test depends solely on RPM which isn't used on these distributions.

shared/checks/oval/installed_env_has_systemd_package.xml
```diff
@@ -0,0 +1,24 @@
...
+      <criterion comment="Package systemd is installed" test_ref="test_env_has_systemd_installed" />
+    </criteria>
+  </definition>
+
+  <linux:rpminfo_test check="all" check_existence="at_least_one_exists" comment="system has package systemd installed" id="test_env_has_systemd_installed" version="1">
+    <linux:object object_ref="obj_env_has_systemd_installed" />
+  </linux:rpminfo_test>
+  <linux:rpminfo_object id="obj_env_has_systemd_installed" version="1">
+    <linux:name>systemd</linux:name>
+  </linux:rpminfo_object>
```
jan-cerny: This also can make some rules `notapllicable` on non-RedHat products

redhatrises:
To some degree, I am concerned a little to the complication as well as difficulty in scaling this PR across multiple products brings to the content. Not sure if we should just do pass/fail in the OVAL instead... but maybe this is the right way...?

linux_os/guide/system/software/integrity/fips/sysctl_crypto_fips_enabled/rule.yml
yuumasato: @shawndwells I suggest to stick to current detection of `machine` for now, and move and discuss addition of `kernel` platform to another PR.

shared/checks/oval/installed_env_has_openssh-server_package.xml
yuumasato: Yes, this should also support DEB packages.

yuumasato:
> To some degree, I am concerned a little to the complication as well as difficulty in scaling this PR across multiple products brings to the content. Not sure if we should just do pass/fail in the OVAL instead... but maybe this is the right way...?

@redhatrises Sure, this should be made easier and automated.

I'm also thinking that using CPE Applicability Language may be easier than using CPE Names.
There wouldn't be need for CPE Names and no need to add entries in CPE Dictionary.

shared/checks/oval/installed_env_has_systemd_package.xml
shawndwells: If they don't have systemd, is that a problem?

shawndwells:
@yuumasato pinged me offline. He wants to refactor this PR and will submit a new one later today.

PR #4240, opened 2019-04-03
Mark platform as "machine only" for rules which are missing and are required to have

ggbecker:
#### Description:

- Mark rules required to have platform as machine only.

#### Rationale:

- Some rules are not applicable in container scanning, therefore they need to be marked as machine only. Some rules appear to have it as unmarked, but its group was marked as machine instead so it will inherit this parameter.

PR #4246, opened 2019-04-05
Remove machine only from rules that are applicable to containers

ggbecker:
#### Description:

- Remove platform machine only from rules that are applicable to container.

#### Rationale:

- These rules are applicable to containers therefore can't be marked as machine only.

- Address raised issues on this PR: https://github.com/ComplianceAsCode/content/pull/4240

### DIFF
--- a/linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/group.yml
+++ b/linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/group.yml
@@ -11,5 +11,3 @@ description: |-
     In Red Hat Enterprise Linux servers and workstations, hardware token login
     {{% endif %}}
     is not enabled by default and must be enabled in the system settings.
-
-platform: machine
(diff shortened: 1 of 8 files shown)


linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/group.yml
```diff
@@ -11,5 +11,3 @@ description: |-
     In Red Hat Enterprise Linux servers and workstations, hardware token login
     {{% endif %}}
     is not enabled by default and must be enabled in the system settings.
-
-platform: machine
```
jan-cerny: The group smart_card_login conatins rule service_pcscd_enabled, which uses systemd check but OpenSCAP cannot evaluate systemd checks in containers.

ggbecker: Agreed. Rule reverted to as machine only.

redhatrises: > The group smart_card_login conatins rule service_pcscd_enabled, which uses systemd check but OpenSCAP cannot evaluate systemd checks in containers.

This is something that needs to be fixed, btw.

matusmarhefka:
ACK

PR #4250, opened 2019-04-08
Update STIG and NIST CUI naming

redhatrises:
- This is needed to reduce complexity of Ansible uploading script. This way we don't have to handle fixing names like `rhel7_stig_rhel7_disa` for the Ansible role name in the `meta/main.yml` file.


shawndwells:
Thanks for bringing the names between RHEL8/RHEL7 into alignment. This will be something downstreams may want to ensure gets documented in their releases notes, e.g. for RHEL and OEL 7.7 (@dahaic @iokomin).

PR #4276, opened 2019-04-12
Fix remediation terminology

jan-cerny:
Currently, we build Ansible Playbooks for profiles into `roles` directory and they're called `ssg-product-role-profile.yml`. That can make a false impression that those files are Ansible roles, but they are
not Ansible roles, they are Ansible Playbooks.  Similarly, we now build "Bash roles". But the term "Bash role" doesn't exist, the Bash files that we build should be called "Bash remediation scripts" instead. This
patch fixes this confusion of terms.  The Ansible Playbooks will be built into `ansible` directory and they will have `playbook` instead of `role` in their name. The Bash remediation scripts will be built into
`bash` directory and they will have `script` instead of `role` in their name. Hopefully it will be less confusing for users. The playbook and script headers are also confusing but that has already been fixed in OpenSCAP in https://github.com/OpenSCAP/openscap/pull/1261.

### DIFF
--- a/build-scripts/build_all_remediation_roles.py
+++ b/build-scripts/build_all_remediation_roles.py
@@ -4,7 +4,7 @@ from __future__ import print_function
 
 """
 Takes given XCCDF or DataStream and for every profile in it it generates one
-ansible and bash remediation roles.
+Ansible Playbook and Bash script.
 
 Author: Martin Preisler <mpreisle@redhat.com>
 """
@@ -77,7 +77,8 @@ def main():
 
     if args.cmd == "list_outputs":
         role_paths = ssg.build_roles.get_output_paths(benchmarks, benchmark_profile_pairs,
-                                                      path_base, extension, output_dir)
+                                                      path_base, extension, output_dir,
+                                                      template)
 
         for role_path in role_paths:
             print(role_path)
--- a/build-scripts/build_all_remediation_roles.py
+++ b/build-scripts/build_all_remediation_roles.py
@@ -59,6 +59,9 @@ def main():
     extension = args.extension
     template = args.template
 
+    if path_base.startswith("ssg-"):
+        path_base = path_base[len("ssg-"):]
+
     if args.cmd == "list_inputs":
         print(input_path)
         sys.exit(0)
--- a/utils/upload_ansible_roles_to_galaxy.py
+++ b/utils/upload_ansible_roles_to_galaxy.py
@@ -409,7 +409,7 @@ def parse_args():
     parser.add_argument(
         "--build-playbooks-dir", required=True,
         help="Path to directory containing the generated Ansible Playbooks. "
-        "Most likely this is going to be ./build/playbooks",
+        "Most likely this is going to be ./build/ansible",
         dest="build_playbooks_dir")
     parser.add_argument(
         "--organization", "-o", default=ORGANIZATION_NAME,
--- a/utils/upload_ansible_roles_to_galaxy.py
+++ b/utils/upload_ansible_roles_to_galaxy.py
@@ -449,8 +449,8 @@ def select_roles_to_upload(product_whitelist, profile_whitelist,
     for filename in os.listdir(build_playbooks_dir):
         root, ext = os.path.splitext(filename)
         if ext == ".yml":
-            # the format is ssg-product-playbook-profile.yml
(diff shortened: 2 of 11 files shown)


utils/upload_ansible_roles_to_galaxy.py
```diff
@@ -467,7 +467,7 @@ def main():
     # the last 4 cut chars are for ".yml"
     available_roles = set(
         [f[4:-4]
-         for f in os.listdir(args.build_roles_dir) if f.endswith(".yml")]
+         for f in os.listdir(args.build_playbook_dir) if f.endswith(".yml")]
```
jan-cerny: missing s

redhatrises:
@jan-cerny make sure to also fix the `upload_ansible_roles_to_galaxy.py`. Also, I wonder if we should get rid of `ssg-` in front of the files? Might be nicer.

jan-cerny:
@redhatrises 

Yes, removing `ssg-` prefix from the playbook and bash remediation files would be nice.

What do you think about proposed names of the directories - `playbooks` and `bash-scripts`? Would `ansible` and `bash` be better?

matusmarhefka:
What about renaming `ssg/build_roles.py` and ` build-scripts/build_all_remediation_roles.py` too? Also there are mentions of roles in `README.md` which should be corrected.

build-scripts/build_all_remediation_roles.py
```diff
@@ -59,6 +59,9 @@ def main():
     extension = args.extension
     template = args.template
 
+    if path_base.startswith("ssg-"):
+        path_base = path_base[len("ssg-"):]
```
matejak: I would say either save the `ssg-` prefix to a variable, or hardcode trim `path_base[3:]`.

jan-cerny:
@matusmarhefka  I have renamed them.

build-scripts/build_all_remediation_roles.py
jan-cerny: I have saved the prefix in a new variable.

jan-cerny:
@matusmarhefka 
> Also there are mentions of roles in README.md which should be corrected.

I have a problem with that. These mentions are about content shipped in RPM package and installed in `/usr/share/scap-security-guide/ansible/` and this pull request doesn't change the spec files. The text in `README.md` doesn't specify which Linux distribution it is related to. Is it Debian? RHEL8? Different Linux distributions will package the content in a different way. If you think this should be reworked or updated, please submit a separate PR.

matusmarhefka:
@jan-cerny But the next time you will build the package on whatever Linux distro there will be Ansible playbooks with `playbook` instead of `role` in their name right? So even though specfiles are not changed the file names of playbooks will be.

jan-cerny:
@matusmarhefka If I change it in README.md, then people might object "I use RHEL 7.6 and it's not there like that, it's different!" But if I don't change it in README.md, the same situation can happen.

I suggest changing the sentence `We assume you have installed SCAP Security Guide system-wide into a standard location as instructed in the previous section.` to `We assume you have installed SCAP Security Guide system-wide into a standard location  from current upstream sources as instructed in the previous section.` and then we can change `playbook` to `role` there. What do you think about that?

matusmarhefka:
@jan-cerny Makes sense, I agree.

jan-cerny:
@matusmarhefka I have updated the REDME and I have also rebased this on the top of master.

jan-cerny:
The question about the names of directories remains unanswered, but I think we should agree on that before we merge it. Asking again: What are your opinions?

matusmarhefka:
I am fine with both, but if I need to pick one I would go for the plain `bash`/`ansible` directory names.

jan-cerny:
> I am fine with both, but if I need to pick one I would go for the plain bash/ansible directory names.

I like it, this will have a benefit that it's consistent with dowstream packaging. For example in Fedora we have the playbooks in `/usr/share/scap-security-guide/ansible/` so it would be nice to have them in `/build/ansible` in upstream as well.

jan-cerny:
@matusmarhefka Are any further changes needed in this PR?

matusmarhefka:
@jan-cerny No and as nobody else expressed their opinion about directory names and the PR LGTM as it is now, I am merging it. Thanks for fixing the terminology!

PR #4279, opened 2019-04-15
Fix reference to doc/commit which shows how to create new product.

ggbecker:
#### Description:

- Fix broken link to documentation on how to create new product.

#### Rationale:

- Current version tries to point to a file in master branch which is not desired. Changed to correct commit link.


yuumasato:
@ggbecker Nice catch again!

PR #4280, opened 2019-04-16
All SELinux related rules marked as not applicable to containers

matusmarhefka:
#### Description:

* The rule `docker_selinux_enabled` moved from `system/selinux` to `services/docker`.
* SELinux is not namespaced which means that containers do not have their own
  separate SELinux policies. SELinux will always appear to be disabled when
  inside a container (https://danwalsh.livejournal.com/73099.html). Therefore,
  all the rules from the `system/selinux` were marked with `platform: machine`
  which will make them not applicable when scanning container filesystems.


yuumasato:
The relocation of rule `docker_selinux_enabled` makes sense to me.

PR #4290, opened 2019-04-17
Mark service rules as machine only.

ggbecker:
#### Description:

- These rules pass when scanning container but we want them as notapplicable.

#### Rationale:

- Rules who have ocil as ocil_service_disabled or complete_ocil_entry_socket_and_service_disabled were marked as machine only.


redhatrises:
@ggbecker @yuumasato question: what if I am running one of these services in a container?

yuumasato:
@redhatrises that is a perfectly valid scenario, but checking state of services can be tricky.

The checks for service either use `systemd` probe, or they check for systemd target dependency symlinks.

As I understand, a service in a container would be running via entry point, or via init (so systemd).
In case of service started by entry point, the actual checks wouldn't be much effective.
And in case of service started via init, the `systemd` probe (and some others) is/are not supported in offline mode.
Another alternative is to install and run openscap in the container,  but that is not much useful.

Issue #3136, opened 2018-07-18
DISA STIG RHEL-07: modprobe: FATAL: Module sha1 not found and FIPS integrity test failed


ykorkmaz:
#### Description of problem:
After the CENTOS 7 installation on a physical server where DISA STIG profile is selected, I got the following error message and can not boot into OS: "dracut-pre-trigger[646]: modprobe: FATAL: Module sha1 not found" and "FIPS integrity test failed".
Actually, I repeated the installation already a couple of times and got the same error message and think it is because of the security profile selected which is DISA STIG RHEL7.

#### SCAP Security Guide Version:
Built-in CentOS

#### Operating System Version:
CentOS 7

#### Steps to Reproduce:

1. Start OS installation on a server with UEFI boot
2. Select DISA STIG RHEL7 profile
3. Reboot the server
4.

#### Actual Results:
FIPS verification failed

#### Expected Results:
FIPS verification successful

ykorkmaz:
It turned out that UUID of the boot partition was not specified in the in the GRUB_CMDLINE_LINUX key in /etc/default/grub file. After adding it manually and rebuilding the grub.conf, the problem has been resolved.

However, I have selected the DISA STIG RHEL7 profile during installation and UUID should have been already added to the boot loader configuration to enable FIPS as described in the following documentation: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/security_guide/chap-federal_standards_and_regulations

Somehow the selected profile only adds the fips=1 parameter but not UUID of the boot partition which causes the problem after an update or so.

matejak:
For reference, we have a related BZ issue:

https://bugzilla.redhat.com/show_bug.cgi?id=1672737

It is still valid, the remediation has to reflect that.

Mab879:
Thanks for opening this issue.

RHEL 7 has been removed from the project. See #12093 for more details.

I'm closing this issue since RHEL7 has been removed from the project and there are signs that issue might already be fixed.

PR #4291, opened 2019-04-18
Updated bash remediation for the rule grub2_enable_fips_mode

cpellitt:
Updated bash remediation for the rule grub2_enable_fips_mode

- Install dracut-fips-aesni on supported hardware

- Disable ed25519 key in SSH configuration
ed25519 is not supported in FIPS mode. If sshd attempts to load a non-existent key, it displays a warning every time it starts. If the key was previously generated while the system was in non-FIPS mode, it can cause sshd to fail.

- Correct sed quoting syntax
This line has a shellcheck error:
sed -i 's/\(^GRUB_CMDLINE_LINUX=".*\)"/\1 boot=UUID='${BOOT_UUID}'"/'  /etc/default/grub
SC2086: Double quote to prevent globbing and word splitting.

The syntax is unusual, with the sed command made up of two single quoted strings and an unquoted variable in the middle. I put double quotes around the whole thing so that the variable expands and escaped the double quotes inside.

### DIFF
--- a/linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/bash/shared.sh
+++ b/linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/bash/shared.sh
@@ -7,6 +7,11 @@ disable_prelink
 
 package_install dracut-fips
 
+# Enable AESNI if supported
+if grep -q -m1 -o aes /proc/cpuinfo; then
+	package_install dracut-fips-aesni
+fi
+
 dracut -f
 
 # Correct the form of default kernel command line in  grub
@@ -26,8 +31,13 @@ if grep -q '^GRUB_CMDLINE_LINUX=".*boot=.*"'  /etc/default/grub; then
 	sed -i 's/\(^GRUB_CMDLINE_LINUX=".*\)boot=[^[:space:]]*\(.*"\)/\1 boot=UUID='"${BOOT_UUID} \2/" /etc/default/ grub
 else
 	# no existing boot=arg is present, append it
-	sed -i 's/\(^GRUB_CMDLINE_LINUX=".*\)"/\1 boot=UUID='${BOOT_UUID}'"/'  /etc/default/grub
+	sed -i "s/\(^GRUB_CMDLINE_LINUX=\".*\)\"/\1 boot=UUID=${BOOT_UUID}\"/" /etc/default/grub
 fi
 
 # Correct the form of kernel command line for each installed kernel in the bootloader
 /sbin/grubby --update-kernel=ALL --args="fips=1 boot=UUID=${BOOT_UUID}"
+
+# Disable ed25519 key in SSH configuration (does not work in FIPS mode)
+if grep -q '^HostKey /etc/ssh/ssh_host_ed25519_key' /etc/ssh/sshd_config; then
+	sed -i 's|^HostKey /etc/ssh/ssh_host_ed25519_key|#HostKey /etc/ssh/ssh_host_ed25519_key|' /etc/ssh/sshd_config
+fi


linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/bash/shared.sh
```diff
@@ -7,6 +7,11 @@ disable_prelink
 
 package_install dracut-fips
 
+# Enable AESNI if supported
+if grep -q -m1 -o aes /proc/cpuinfo; then
```
redhatrises: Preferentially, this would be in an OVAL file and added to the OVAL grub FIPS check using the OVAL and bash templates.

linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/bash/shared.sh
```diff
@@ -26,8 +31,13 @@ if grep -q '^GRUB_CMDLINE_LINUX=".*boot=.*"'  /etc/default/grub; then
...
+	sed -i "s/\(^GRUB_CMDLINE_LINUX=\".*\)\"/\1 boot=UUID=${BOOT_UUID}\"/" /etc/default/grub
 fi
 
 # Correct the form of kernel command line for each installed kernel in the bootloader
 /sbin/grubby --update-kernel=ALL --args="fips=1 boot=UUID=${BOOT_UUID}"
+
+# Disable ed25519 key in SSH configuration (does not work in FIPS mode)
+if grep -q '^HostKey /etc/ssh/ssh_host_ed25519_key' /etc/ssh/sshd_config; then
+	sed -i 's|^HostKey /etc/ssh/ssh_host_ed25519_key|#HostKey /etc/ssh/ssh_host_ed25519_key|' /etc/ssh/sshd_config
+fi
```
redhatrises: Hmmmm..... not sure that this makes sense to be in the FIPS grub rememdiation.....

linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/bash/shared.sh
cpellitt: This is a preferred configuration, but not a security requirement. I would think that the OVAL should only check that the security requirement is met.

linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/bash/shared.sh
cpellitt: I am not sure where else you would put this. The need for this change is directly tied to enabling FIPS mode.

redhatrises: This is a separate rule with separate OVAL and remediations then as the SSH HostKey's now need to be verified against FIPS.

linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/bash/shared.sh
redhatrises: OVAL is more than about meeting security requirements. Please add an `extended_definition` check to the grub2_enabled_fips OVAL for AES as there is a configuration change here. So, there needs to be:

1. An extended_definition check added to grub2_enable_fips_mode OVAL
2. A dracut-fips-aesni OVAL check in shared/checks that checks for AESNI enabled processors and that the `dracut-fips-aesni` package is installed.
3. The `dracut-fips-aesni` package needs to be added to the correct OS templates.

cpellitt:
I am not familiar with writing XCCDF and OVAL, so I will leave this for someone else to implement.

linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/bash/shared.sh
```diff
@@ -26,8 +31,13 @@ if grep -q '^GRUB_CMDLINE_LINUX=".*boot=.*"'  /etc/default/grub; then
 	sed -i 's/\(^GRUB_CMDLINE_LINUX=".*\)boot=[^[:space:]]*\(.*"\)/\1 boot=UUID='"${BOOT_UUID} \2/" /etc/default/ grub
 else
 	# no existing boot=arg is present, append it
-	sed -i 's/\(^GRUB_CMDLINE_LINUX=".*\)"/\1 boot=UUID='${BOOT_UUID}'"/'  /etc/default/grub
+	sed -i "s/\(^GRUB_CMDLINE_LINUX=\".*\)\"/\1 boot=UUID=${BOOT_UUID}\"/" /etc/default/grub
```
yuumasato: This line could be the fix for #3136!

cpellitt: I tested this line that currently exists in the code, and it did seem to work properly even if it was imperfect.

Your referenced issue looks like it was fixed with this commit:
https://github.com/ComplianceAsCode/content/commit/2a044d57f9200a281c105453b263b3efdee2a500

Anyone with that issue was most likely using an older version that did not contain that change.

PR #4293, opened 2019-04-24
Finish renaming STIG and CUI profiles

jan-cerny:
In #4250 multiple STIG and CUI profiles have been renamed, but
there are still a lot of occurrences of the old names, mainly in the
test suite scenarios. This patch finishes the renaming everywhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,7 +208,7 @@ ssg-rhel7-role-C2S.yml
 ssg-rhel7-role-cjis-rhel7-server.yml
 ssg-rhel7-role-common.yml
 ssg-rhel7-role-docker-host.yml
-ssg-rhel7-role-nist-800-171-cui.yml
+ssg-rhel7-role-cui.yml
 ...
 ```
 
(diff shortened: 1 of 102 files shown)


redhatrises:
@jan-cerny good call! This also highlights how we should be using testing profiles to evaluate the rules rather than the profiles themselves.

jan-cerny:
@redhatrises 
> This also highlights how we should be using testing profiles to evaluate the rules rather than the profiles themselves.

Do you mean that we should have a separate "testing"  profile?

redhatrises:
> @redhatrises
> 
> > This also highlights how we should be using testing profiles to evaluate the rules rather than the profiles themselves.
> 
> Do you mean that we should have a separate "testing" profile?

Yes very much so. 2 actually. 1 to test rules, and the other to test changes to variables.

ACK.

jan-cerny:
> Yes very much so. 2 actually. 1 to test rules, and the other to test changes to variables

They could use the virtual "all" profile which contains all the rules in the benchmark.

PR #4295, opened 2019-04-25
Build playbooks also for virtual '(all)' profile

jan-cerny:
#### Description:
Generates playbooks for all product rules without profile context to a directory.

The playbooks will be generated into `/build/${product}/playbooks/all` directory.

To do the build easily, we fixed a problem that `/build/${product}/rules` directory contained also rules that weren't applicable to given product, now the directory will contain only relevant rules.

The PR also adds a small test that compares the built datastream and `/build/${product}/playbooks/all` to check if Ansible Playbooks were generated for all rules that have an Ansible remediation available in datastream.

#### Rationale:

A product typically contains lots of rules, not all of them are part of any profile. As it may happen that a rule falls out of a profile, it is a good idea to be able to generate playbooks for all rules that are not part of any profile using the build_rule_playbooks.py script.

### DIFF
--- a/cmake/SSGCommon.cmake
+++ b/cmake/SSGCommon.cmake
@@ -257,6 +257,10 @@ macro(ssg_build_ansible_playbooks PRODUCT)
         generate-${PRODUCT}-ansible-playbooks
         DEPENDS "${ANSIBLE_PLAYBOOKS_DIR}"
     )
+    add_test(
+        NAME "${PRODUCT}-ansible-playbooks-generated-for-all-rules"
+        COMMAND env "PYTHONPATH=$ENV{PYTHONPATH}" "${PYTHON_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/tests/ansible_playbooks_generated_for_all_rules.py" --build-dir "${CMAKE_BINARY_DIR}" --product "${PRODUCT}"
+    )
 endmacro()
 
 macro(ssg_build_remediations PRODUCT)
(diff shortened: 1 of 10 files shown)


jan-cerny:
@ggbecker Great idea! I'll add it to the docs.

jan-cerny:
@ggbecker I have added a short pargraph about this output there

ggbecker:
Lets be a little extreme here, if I have a profile named 'all', it won't build the content. Should we use some kind of keyword for this 'all' generated playbooks?

```
  File "/home/user/workspace/github/content/build-scripts/build_rule_playbooks.py", line 84, in <module>
    main()
  File "/home/user/workspace/github/content/build-scripts/build_rule_playbooks.py", line 80, in main
    playbook_builder.build(args.profile, args.rule)
  File "/home/user/workspace/github/content/ssg/playbook_builder.py", line 288, in build
    self.create_playbooks_for_all_rules(variables)
  File "/home/user/workspace/github/content/ssg/playbook_builder.py", line 242, in create_playbooks_for_all_rules
    os.makedirs(profile_playbooks_dir)
  File "/usr/lib64/python3.7/os.py", line 221, in makedirs
    mkdir(name, mode)
FileExistsError: [Errno 17] File exists: '/home/user/workspace/github/content/build/rhel8/playbooks/all'
```

jan-cerny:
@ggbecker  we use `(all)` in OpenSCAP instead, but people have to  quote that in terminal, which is annoying, but otherwise they would get `-bash: syntax error near unexpected token 'all'` because parenthesis have special meaning.

Question is if it's worth the hassle because I doubt people will ever create profile `all` in SSG.

ggbecker:
> @ggbecker we use `(all)` in OpenSCAP instead, but people have to quote that in terminal, which is annoying, but otherwise they would get `-bash: syntax error near unexpected token 'all'` because parenthesis have special meaning.
> 
> Question is if it's worth the hassle because I doubt people will ever create profile `all` in SSG.

Okay, I agree we don't need to overengineer this.

PR #4300, opened 2019-04-25
Add content to handle AES-NI

### LINKED ISSUE #3221: Introduce support for dracut-fips-aesni package in Anaconda
Package `dracut-fips-aesni` enables usage of aes instruction set of x86 CPUs. This has significant impact for some crypto related workflows. It does not have (AFAIK) functionality implications.

We cannot install it indiscriminately, as supported configuration for Red Hat Enterprise Linux 7 is this package installed only in case CPU supports given instruction set. So if we want to install package, we need to first check the support.

That means we cannot (at least without changes to anaconda addon) install it during installation whatsoever. So there would be discrepancy between installation hardening and normal bash runtime hardening.

It is not clear if we should put it to the OVAL.

redhatrises:
- Fixes #3221

### DIFF
--- a/linux_os/guide/system/software/integrity/fips/package_dracut-fips-aesni_installed/oval/shared.xml
+++ b/linux_os/guide/system/software/integrity/fips/package_dracut-fips-aesni_installed/oval/shared.xml
@@ -1,0 +1,44 @@
+<!-- DO NOT TEMPLATE this check. dracut-fips-aesni should only be installed on 
+ certified systems, and changes to this file should be scrutinized through the
+ review process.
+-->
+<def-group>
+  <definition class="compliance" id="package_dracut-fips-aesni_installed"
+  version="1">
+    <metadata>
+      <title>Package dracut-fips-aesni Installed</title>
+      <affected family="unix">
+        <platform>Red Hat Enterprise Linux 6</platform>
+        <platform>Red Hat Enterprise Linux 7</platform>
+        <platform>Oracle Linux 7</platform>
+      </affected>
+      <description>The RPM package dracut-fips-aesni should be installed.</description>
+    </metadata>
+    <criteria operator="OR">
+      <criterion comment="System does not support AES instruction set" test_ref="test_processor_aes_instruction" />
+      <criteria operator="AND">
+        <extend_definition comment="Installed OS is FIPS certified" definition_ref="installed_OS_is_FIPS_certified" />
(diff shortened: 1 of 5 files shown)


linux_os/guide/system/software/integrity/fips/package_dracut-fips-aesni_installed/anaconda/shared.anaconda
```diff
@@ -0,0 +1,3 @@
+# platform = Red Hat Enterprise Linux 6,Red Hat Enterprise Linux 7,Oracle Linux 7
+
+package --add=dracut-fips-aesni
```
jan-cerny: The reporter in the issues #3221 says that:

> We cannot install it indiscriminately, as supported configuration for Red Hat Enterprise Linux 7 is this package installed only in case CPU supports given instruction set. So if we want to install package, we need to first check the support.

> That means we cannot (at least without changes to anaconda addon) install it during installation whatsoever. 

From what I get Anaconda still doesn't have any changes like that.

My understanding is that this remediation will break installation on other architectures. I haven't tested it, though.

redhatrises: @jan-cerny went ahead and removed the anaconda remediation for now.

linux_os/guide/system/software/integrity/fips/package_dracut-fips-aesni_installed/oval/shared.xml
```diff
@@ -0,0 +1,44 @@
...
+        test_ref="test_package_dracut-fips-aesni_installed" />
+      </criteria>
+    </criteria>
+  </definition>
+
+  <ind:textfilecontent54_test check="all" check_existence="none_exist" comment="query /proc/cpuinfo" id="test_processor_aes_instruction" version="1">
+    <ind:object object_ref="object_processor_aes_instruction" />
+  </ind:textfilecontent54_test>
+  <ind:textfilecontent54_object id="object_processor_aes_instruction" version="1">
+    <ind:filepath>/proc/meminfo</ind:filepath>
```
yuumasato: This should be `cpuinfo`

redhatrises:
@yuumasato done and build passes

yuumasato:
Nice touch adding the FIPS regulatory warning

PR #4297, opened 2019-04-25
Remove building of ScientificLinux 8

redhatrises:
- Sadly, as of now, the distrobution is ending and won't be continuing past version 7

PR #4301, opened 2019-04-26
Remove remaining references to ScientificLinux8.

ggbecker:
#### Description:

- Remove remaining references to ScientificLinux8 as a follow up to https://github.com/ComplianceAsCode/content/pull/4297

#### Rationale:

- ScientificLinux8 won't be develop so it doesn't make sense to have this derivative as per: https://listserv.fnal.gov/scripts/wa.exe?A2=SCIENTIFIC-LINUX-ANNOUNCE;11d6001.1904


redhatrises:
Thanks @ggbecker! I knew I was missing something but was blanking on what.

PR #4306, opened 2019-04-26
create general OCP security prose

shawndwells:
Updates OCP content to reflect generic/general security guidance (e.g. rules not specific to a subsystem like etcd).


redhatrises:
@shawndwells what are your thoughts to coming up with something better than `general` settings?

applications/openshift/general/general_configure_imagepolicywebhook/oval/shared.xml
```diff
@@ -0,0 +1,31 @@
...
+  </definition>
+
+  <ind:textfilecontent54_test check="all" check_existence="all_exist" comment="admissionConfig is configured" id="test_general_configure_imagepolicywebhook" version="1">
+    <ind:object object_ref="object_general_configure_imagepolicywebhook" />
+    <ind:state state_ref="state_general_configure_imagepolicywebhook" />
+  </ind:textfilecontent54_test>
+
+  <ind:textfilecontent54_object id="object_general_configure_imagepolicywebhook" version="1">
+    <ind:filepath>/etc/origin/master/master-config.yaml</ind:filepath>
+    <ind:pattern operation="pattern match">^[\s]*admissionConfig\:\n[\s]*pluginConfig\:\n[\s]*openshift.io\/ImagePolicy:$</ind:pattern>
```
shawndwells: I'm not sure if I did this regex correctly. The sample config I used was:

`````
admissionConfig:
  pluginConfig:
    openshift.io/ImagePolicy:
junk that shouldn't match
`````

Was able to match the appropriate lines and tested on regex101.com.... but not sure if there need to be OVAL-specific things in the regex.

**Comment by shawndwells** [comment]
> @shawndwells what are your thoughts to coming up with something better than `general` settings?

Using the language from the marketing OpenShift Configuration Guide. Totally open to new ideas! Rules in this section are generic to all of openshift vs a specific component (e.g. etcd or kube).

**Comment by jan-cerny** [comment]
@shawndwells "General" looks fine because they are general.

File: applications/openshift/general/general_configure_imagepolicywebhook/oval/shared.xml (hunk 1 above)
Comment (jan-cerny) [reply applications/openshift/general/general_configure_imagepolicywebhook/oval/shared.xml]: If you want to match exactly this, then it should be fine, except `:` doesn't have to be escaped and `.` in `openshift.io` should be escaped.

The problem is that you don't want to match only exactly this, because there can be multiple different entries in `pluginConfig`. And  `openshift.io/ImagePolicy` isn't necessarily the first one right after pluginConfigLine. 

There needs to be a regex that matches also lines in between.

This should be correct:
```
^[\s]*admissionConfig:(?:[^\n]*\n+)+?[\s]*pluginConfig:(?:[^\n]*\n+)+?[\s]*openshift\.io\/ImagePolicy:$
```


Take a look at eg. https://github.com/ComplianceAsCode/content/blob/master/applications/openshift/api-server/api_server_admission_control_plugin_DenyEscalatingExec/oval/shared.xml

shawndwells: Thanks @jan-cerny ! Updated the regex.

PR #4307, opened 2019-04-27
[issue 4264] disable controller profiling

shawndwells:
Resolves https://github.com/ComplianceAsCode/content/issues/4264

- Adds new rule for OPENSHIFT_PROFILE != web
- When assigning CCE noticed there were CTRL+M characters in the CCE list. Removed them.


jan-cerny:
Nice catch with the EOLs! Thanks. It would be better to do things like that in a separate PR next time.

PR #4308, opened 2019-04-29
Rename profile_stats to profile_tool and update usage by CMake.

ggbecker:
- Rename profile_stats to profile_tool.
- Fix usage of tool by cmake according new parameters format.

I also make some code improvements regarding meaning of profile subtraction and removed unused parameter from build_yaml.Profile class which was introduced in previous profile subtraction PR.

#### Rationale:

- As a follow up to this https://github.com/ComplianceAsCode/content/pull/4239, the tool has been renamed.


jan-cerny:
I have built RHEL8 content and I have run `make rhel8-profile-stats` and the output was expected. Thanks.

PR #4309, opened 2019-04-29
update ./etcd/etcd_peer_key_file/ to use = instead of space

shawndwells:
the separator should be ``=`` on openshift


redhatrises:
LGTM

PR #4313, opened 2019-04-29
[wip] remove extended def from ./kubelet/kubelet_configure_client_ca/oval/s…

shawndwells:
the extended check doesn't exist, causing this rule to always fail


shawndwells:
merged into a different PR, closing

PR #4302, opened 2019-04-26
Add script to update generated SDS to SCAP 1.3

yuumasato:
#### Description:

- Add script that updates generated DataStream to be compliant with SCAP 1.3.

#### How to test

- Use [scapval-1.3.2](https://csrc.nist.gov/Projects/Security-Content-Automation-Protocol/SCAP-Releases/SCAP-1-3)
  - Use `openjdk-1.8.0` on a RHEL environment. (for some reason it doesn't work on Fedora)
  - Increase allocated memory: `-Xmx1024m`
- Workaround bugs in scap-schematron-rules-1.3.2 (just to test that scapval passes):
  - `group_id` and `user_id` cannot be zero, tracked here https://github.com/OVAL-Community/OVAL/issues/23
    - workaround it : 
      - `sed -i -E "s/(group_id .*\">)/\1-1/" ./ssg-rhel7-ds.xml`
      - `sed -i -E "s/(user_id .*\">)/\1-1/" ./ssg-rhel7-ds.xml`
  - OCIL checks for `security_patches_up_to_date` is causing fail of SRC-377, when requirement is about OVAL checks. I have e-mailed NIST about this to try to sort out what is the issue.
    - workaround it:
      - `awk '/check-content-ref.*up_to_date_ocil/{for(x=NR-1;x<=NR+1;x++)d[x];}{a[NR]=$0}END{for(i=1;i<=NR;i++)if(!(i in d))print a[i]}' ./ssg-rhel7-ds.xml > ./ssg-rhel7-ds-1.3.xml`
- Validate it: `java -Xmx1024m -jar scapval-1.3.2.jar -file ./ssg-rhel7-ds-1.3.xml -scapversion 1.3 > scap-1.3.log`

- Report of scapval with workarounds: https://fedorapeople.org/~wsato/scap-1.3/validation-report.html

#### Rationale:

- SCAP 1.3 should supersede SCAP 1.2 
- In the near future OpenSCAP should be update to handle generation of SCAP 1.3 DS.
  - More specifically the command `oscap ds sds-compose`

PR #4319, opened 2019-04-30
Build multiple scap versions

yuumasato:
#### Description:

- ~~Generate SCAP 1.3 content by default.~~
- ~~To generate SCAP 1.2 content:
  `cmake -DSSG_TARGET_SCAP_MINOR_VERSION:STRING=2 ../ `~~
- Generate both SCAP 1.2 and 1.3

#### Rationale:
- OpenSCAP cannot scan when remote datastream component source is missing. (Without `--fetch-remote-resources`).
  See, https://github.com/ComplianceAsCode/content/pull/4302#issuecomment-487639000
- ~~Until this is fixed, default to generating SCAP 1.2 Source Data Streams.~~

### DIFF
--- a/build-scripts/update_sds_to_scap_1_3.py
+++ b/build-scripts/update_sds_to_scap_1_3.py
@@ -1,6 +1,7 @@
 
 from __future__ import print_function
 
+import argparse
 import sys
 try:
     from urllib.parse import urlparse
@@ -107,18 +108,18 @@ def move_patches_up_to_date_to_source_data_stream_component(datastreamtree):
             component_ref_feed.set('{%s}href' % xlink_namespace, href_url)
             ds_checks.append(component_ref_feed)
 
+def parse_args():
+    p = argparse.ArgumentParser()
+    p.add_argument("--input", required=True)
+    p.add_argument("--output", required=True)
+
+    return p.parse_args()
 
 def main():
-    if len(sys.argv) < 3:
-        print("This script updates SCAP 1.2 Source DataStream to SCAP 1.3")
-        sys.exit(1)
+    args = parse_args()
 
-    # Input datastream file
-    indatastreamfile = sys.argv[1]
-    # Output datastream file
-    outdatastreamfile = sys.argv[2]
     # Datastream element tree
-    datastreamtree = ssg.xml.ElementTree.parse(indatastreamfile).getroot()
+    datastreamtree = ssg.xml.ElementTree.parse(args.input).getroot()
 
     # Set SCAP version to 1.3
     datastreamtree.set('schematron-version', '1.3')
@@ -127,7 +128,7 @@ def main():
     # Move reference to remote OVAL content to a source data stream component
     move_patches_up_to_date_to_source_data_stream_component(datastreamtree)
 
-    ssg.xml.ElementTree.ElementTree(datastreamtree).write(outdatastreamfile)
+    ssg.xml.ElementTree.ElementTree(datastreamtree).write(args.output)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


redhatrises:
@yuumasato if we start generating different scap versions, I am wondering if we should rename the script as the script name no longer makes sense.

CMakeLists.txt
```diff
@@ -40,6 +40,10 @@ set(SSG_TARGET_OVAL_MAJOR_VERSION "5" CACHE STRING "Which major version of OVAL
 set(SSG_TARGET_OVAL_MINOR_VERSION "11" CACHE STRING "Which minor version of OVAL are we targetting. Possible choices are 10 or 11.")
 set(SSG_TARGET_OVAL_VERSION "${SSG_TARGET_OVAL_MAJOR_VERSION}.${SSG_TARGET_OVAL_MINOR_VERSION}")
 
+set(SSG_TARGET_SCAP_MAJOR_VERSION "1" CACHE STRING "Which major version of SCAP are we targetting.")
+set(SSG_TARGET_SCAP_MINOR_VERSION "2" CACHE STRING "Which minor version of SCAP are we targetting. Posiblle choices are 2 or 3")
```
shawndwells:
```suggestion
set(SSG_TARGET_SCAP_MINOR_VERSION "2" CACHE STRING "Which minor version of SCAP are we targetting. Possible choices are 2 or 3")
```

shawndwells:
With SCAP 1.3 being the latest official standard, are there cases where SCAP 1.2 should still be generated? Why not have 1.3 be default?

yuumasato:
> With SCAP 1.3 being the latest official standard, are there cases where SCAP 1.2 should still be generated? Why not have 1.3 be default?

@shawndwells What I'm concerned about is that SCAP 1.3 content will **only** work in with OpenSCAP when `--fetch-remote-resouces` is provided.
And I think that forcing the download of the feed every time is not an acceptable user experience, and I believe this will break offline and airgaped use cases.

See, https://github.com/ComplianceAsCode/content/pull/4302#issuecomment-487639000.
When patch https://github.com/OpenSCAP/openscap/pull/1324 is shipped, I believe we can switch to generating SCAP 1.3 by default.

jan-cerny:
What about generating both versions at once?

yuumasato:
> I am wondering if we should rename the script as the script name no longer makes sense.

@redhatrises  Do you have a suggestion?

jan-cerny:
I suggest:
* by default build SCAP-1.3 datastreams containing OVAL 5.11
* keep cmake option to set OVAL version (5.10/5.11)
* add cmake option to set DS version (1.2/1.3)
* since they are 2 separate cmake options it will enable users to build SCAP 1.2 with OVAL 5.10, and also they will be able to build weird hybrids if they want

yuumasato:
I'm fine with SCAP 1.3 being the default build.

shawndwells:
On 5/2/19 5:30 AM, Watson Yuuma Sato wrote:
>
>     With SCAP 1.3 being the latest official standard, are there cases
>     where SCAP 1.2 should still be generated? Why not have 1.3 be default?
>
> @shawndwells <https://github.com/shawndwells> What I'm concerned about 
> is that SCAP 1.3 content will *only* work in with OpenSCAP when 
> |--fetch-remote-resouces| is provided.
> And I think that forcing the download of the feed every time is not an 
> acceptable user experience, and I believe this will break offline and 
> airgaped use cases.
>
> See, #4302 (comment) 
> <https://github.com/ComplianceAsCode/content/pull/4302#issuecomment-487639000>.
> When patch OpenSCAP/openscap#1324 
> <https://github.com/OpenSCAP/openscap/pull/1324> is shipped, I believe 
> we can switch to generating SCAP 1.3 by default.
>

You're totally right. Good call.

shawndwells:
> I suggest:
> 
> * by default build SCAP-1.3 datastreams containing OVAL 5.11
> * keep cmake option to set OVAL version (5.10/5.11)
> * add cmake option to set DS version (1.2/1.3)
> * since they are 2 separate cmake options it will enable users to build SCAP 1.2 with OVAL 5.10, and also they will be able to build weird hybrids if they want

+1

image

yuumasato:
I'm fine generating both contents in one sweep.

jan-cerny:
@yuumasato please resolve the conflicts
Otherwise looks great!

yuumasato:
image

PR #4323, opened 2019-05-01
Make sure that registered variables are defined for SSSD Ansible content

redhatrises:


shawndwells:
Cleaner. Ack pending jenkins.

PR #4324, opened 2019-05-02
Profile/Product statistics in HTML - improvement for profile_stats.py and jenkins integration

ggbecker:
#### Description:

- Introduces new HTML output format to profile_tool.py statistics tool. This is used by a newly created jenkins job to display information about products and profiles: https://jenkins.complianceascode.io/job/scap-security-guide-stats/Statistics/

#### Rationale:

- This will provide a better way to access the information about product/profile completeness regarding their selected rules. Completeness in a sense of which OVAL/bash/ansible are implemented, missing CCEs, etc. There is room for improvement on what data should be displayed and in which way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@
 [![CentOS CI Status](https://ci.centos.org/job/openscap-scap-security-guide/badge/icon)](https://ci.centos.org/job/openscap-scap-security-guide/)
 [![Travis CI Build Status](https://travis-ci.org/OpenSCAP/scap-security-guide.svg?branch=master)](https://travis-ci.org/OpenSCAP/scap-security-guide)
 [![Scrutinizer Code Quality](https://scrutinizer-ci.com/g/ComplianceAsCode/content/badges/quality-score.png?b=master)](https://scrutinizer-ci.com/g/ComplianceAsCode/content/?branch=master)
+[![Profile Statistics](https://jenkins.complianceascode.io/job/scap-security-guide-stats/badge/icon)](https://jenkins.complianceascode.io/job/scap-security-guide-stats/)
 
 <a href="docs/readme_images/report_sample.png"><img align="right" width="250" src="docs/readme_images/report_sample.png" alt="Evaluation report sample"></a>
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -250,6 +250,9 @@ add_subdirectory("tests")
 add_custom_target(stats)
 add_custom_target(profile-stats)
 
+add_custom_target(html-stats)
+add_custom_target(html-profile-stats)
+
 ssg_build_bash_remediation_functions()
 
 ssg_build_man_page()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -104,6 +104,7 @@ find_python_module(yaml REQUIRED)
 find_python_module(jinja2 REQUIRED)
 find_python_module(pytest)
 find_python_module(pytest_cov)
+find_python_module(json2html)
 
 include(CMakeDependentOption)
 cmake_dependent_option(ENABLE_PYTHON_COVERAGE "Enable Python tests with coverage support" ON "PY_PYTEST_COV" OFF)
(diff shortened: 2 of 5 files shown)


redhatrises:
Sweet! Is the json2html package in any of the RHEL7 repos?

shawndwells:
Fantastic!! Really will help identify what profiles need work.

Would be great to see details about the missing rules. Should that 
functionality be added in a future PR, and merge this one as is?

ggbecker:
> Sweet! Is the json2html package in any of the RHEL7 repos?

I'm not aware of any rhel repo containing this package. For the moment you can install through pip.

ggbecker:
> Fantastic!! Really will help identify what profiles need work. Would be great to see details about the missing rules. Should that functionality be added in a future PR, and merge this one as is?

The change of what is being showed should be simple. We just would need to change which arguments are being used in the profile_tool.py (previously profile_stats.py) script. We already have the information there but the HTML is kinda ugly, it needs polishing. I'm okay having it merged now and then improve the data and how it should be displayed.

redhatrises:
> > Sweet! Is the json2html package in any of the RHEL7 repos?
> 
> I'm not aware of any rhel repo containing this package. For the moment you can install through pip.

Slight concern about that. If people need to use profile_stats.py and are on rhel7, is profile_stats.py now broken because of the json2html package?

redhatrises:
> > > Sweet! Is the json2html package in any of the RHEL7 repos?
> > 
> > 
> > I'm not aware of any rhel repo containing this package. For the moment you can install through pip.
> 
> Slight concern about that. If people need to use profile_stats.py and are on rhel7, is profile_stats.py now broken because of the json2html package?

There is also the case where people do not have access or cannot use pip.

ggbecker:
> > > > Sweet! Is the json2html package in any of the RHEL7 repos?
> > > 
> > > 
> > > I'm not aware of any rhel repo containing this package. For the moment you can install through pip.
> > 
> > 
> > Slight concern about that. If people need to use profile_stats.py and are on rhel7, is profile_stats.py now broken because of the json2html package?
> 
> There is also the case where people do not have access or cannot use pip.

profile_stats.py will only require json2html if you request to generate the output data in HTML format.

But I agree this is not the best solution, json2html was the best solution for prototyping. One option could be remove this html transformation and let jenkins handle it and keep profile_stats.py producing only the usual output data in plain/JSON/csv formats.

redhatrises:
> > > > > Sweet! Is the json2html package in any of the RHEL7 repos?
> > > > 
> > > > 
> > > > I'm not aware of any rhel repo containing this package. For the moment you can install through pip.
> > > 
> > > 
> > > Slight concern about that. If people need to use profile_stats.py and are on rhel7, is profile_stats.py now broken because of the json2html package?
> > 
> > 
> > There is also the case where people do not have access or cannot use pip.
> 
> profile_stats.py will only require json2html if you request to generate the output data in HTML format.
> 
> But I agree this is not the best solution, json2html was the best solution for prototyping. One option could be remove this html transformation and let jenkins handle it and keep profile_stats.py producing only the usual output data in plain/JSON/csv forma

I like the addition. My only concern is that we don't break functionality for downstream users, so whatever prevents that.

README.md
```diff
@@ -7,6 +7,7 @@
 [![CentOS CI Status](https://ci.centos.org/job/openscap-scap-security-guide/badge/icon)](https://ci.centos.org/job/openscap-scap-security-guide/)
 [![Travis CI Build Status](https://travis-ci.org/OpenSCAP/scap-security-guide.svg?branch=master)](https://travis-ci.org/OpenSCAP/scap-security-guide)
 [![Scrutinizer Code Quality](https://scrutinizer-ci.com/g/ComplianceAsCode/content/badges/quality-score.png?b=master)](https://scrutinizer-ci.com/g/ComplianceAsCode/content/?branch=master)
+[![Profile Statistics](https://jenkins.complianceascode.io/job/scap-security-guide-stats/badge/icon)](https://jenkins.complianceascode.io/job/scap-security-guide-stats/)
```
jan-cerny: :heart_eyes: :heart_eyes:

jan-cerny:
It will only fail if you add `--format html` and you don't have the package installed, otherwise it works fine.

PR #4326, opened 2019-05-03
Do not overwrite 1.2 DS with 1.3 DS during xmllint

yuumasato:
#### Description:

- Do not overwrite SCAP 1.2 DS with SCAP 1.3 DS during `xmllint`


yuumasato:
@ggbecker Thanks!

PR #4327, opened 2019-05-03
Add test wrapper around SCAPVal tool

jan-cerny:
Runs SCAPVal tool on all built datastreams and then parses the results.

This wrapper will run in a special job in our Jenkins, which will help us to preserve SCAP 1.3 conformance.

Example usage:
```
python run_scapval.py --scap-version 1.3 \
--scapval-path /opt/scapval/scapval-1.3.2.jar \
--build-dir /home/jcerny/content/build
```

### DIFF
--- a/tests/run_scapval.py
+++ b/tests/run_scapval.py
@@ -72,6 +72,7 @@ def workaround_datastream(datastream_path):
 
 def test_datastream(datastream_path,  scapval_path, scap_version):
     result_path = datastream_path + ".result.xml"
+    report_path = datastream_path + ".report.xml"
     if scap_version == "1.3":
         datastream_path = workaround_datastream(datastream_path)
     scapval_command = [
@@ -80,7 +81,8 @@ def test_datastream(datastream_path,  scapval_path, scap_version):
             "-jar", scapval_path,
             "-scapversion", scap_version,
             "-file", datastream_path,
-            "-valresultfile", result_path
+            "-valresultfile", result_path,
+            "-valreportfile", report_path
             ]
     subprocess.check_output(scapval_command, stderr=subprocess.STDOUT)
     return process_results(result_path)
--- a/tests/run_scapval.py
+++ b/tests/run_scapval.py
@@ -8,7 +8,7 @@ import os
 import xml.etree.ElementTree as ET
 import sys
 
-NS = "http://csrc.nist.gov/ns/decima/results/1.0"
+SCAPVAL_RESULTS_NS = "http://csrc.nist.gov/ns/decima/results/1.0"
 oval_unix_ns = "http://oval.mitre.org/XMLSchema/oval-definitions-5#unix"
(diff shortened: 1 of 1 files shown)


jan-cerny:
SCAP 1.3 conformance results (as of HEAD = 8cb2d0f351faff5440742258782281164953b0a6):
* ssg-fuse6-ds-1.3.xml fails on SRC-15
* ssg-debian8-ds-1.3.xml fails on SRC-15
* ssg-ol7-ds-1.3.xml fails on SRC-330
* other *-ds-1.3.xml pass

See attached reports 
[reports.zip](https://github.com/ComplianceAsCode/content/files/3147645/reports.zip)

jan-cerny:
* ssg-fuse6-ds-1.3.xml fails on SRC-15
  * there are 3 CPEs used in the benchamrk, but two of them are not defined in the CPE dictionary (they're commented out) and they miss OVALs
* ssg-debian8-ds-1.3.xml fails on SRC-15
  * the Debian 8 CPE dictionary contains only Debian 8 CPE, but doesn't contain machine CPE or recently introduced package CPEs
* ssg-ol7-ds-1.3.xml fails on SRC-330
   * OL7 profiles unselect "sap" profile instead of "sap" group. Looks like a bug in OpenSCAP in `xsl/xccdf_1.1_to_1.2.xsl`. Try to run this command that converts XCCDF 1.1 to XCCDF 1.2: `xsltproc --stringparam reverse_DNS "org.ssgproject.content" --output  /tmp/xccdf12.xml ~/openscap/xsl/xccdf_1.1_to_1.2.xsl build/ssg-ol7-xccdf.xml `. You will see that `<select idref="sap" selected="false"/>` is incorrectly translated to `<select idref="xccdf_org.ssgproject.content_profile_sap" selected="false"/>`. A proper fix is to fix the XSLT templates in OpenSCAP. A workaround is to rename either the `sap` Group or rename `sap` OL7 profile so that the 2 names don't conflict.

jan-cerny:
Fuse6 and Debian8 problems are addressed in  PR #4329 .
OL7 problems are addressed in PR #4332 .

tests/run_scapval.py
```diff
@@ -0,0 +1,116 @@
...
+    scapval_command = [
+            "java",
+            "-Xmx1024m",
+            "-jar", scapval_path,
+            "-scapversion", scap_version,
+            "-file", datastream_path,
+            "-valresultfile", result_path,
+            "-valreportfile", report_path
+            ]
+    subprocess.check_output(scapval_command, stderr=subprocess.STDOUT)
```
ggbecker: Please, print the stdout and/or stderr if an exception is raised by this function. Which means that scapval command returned non-zero exit code.

ggbecker:
Please rebase with master.

jan-cerny:
@ggbecker I have rebased and caught exception.

ggbecker:
LGTM

PR #4328, opened 2019-05-03
Fix-ups and remote host support for tests/install_vm.py

comps:
See the commit messages for each separate commit for more details.

The `install_vm.py` tool lacked support for remote libvirt hosts and also had a bunch of questionable options and hardcoded defaults - I've tried to clean it up to the best of my knowledge and it should now work for both local and remote installs.

If you use this tool in some non-default / non-standard way, please test these changes and let me know if something breaks!

### DIFF
--- a/tests/install_vm.py
+++ b/tests/install_vm.py
@@ -94,13 +94,13 @@ def main():
     print("Using SSH public key from file: {0}".format(data.ssh_pubkey))
     print("Using hypervisor: {0}".format(data.libvirt))
 
-    if not data.disk_dir:
-        if data.libvirt == "qemu:///system":
-            data.disk_dir = "/var/lib/libvirt/images/"
-        else:
-            data.disk_dir = home_dir + "/.local/share/libvirt/images/"
-    data.disk_path = os.path.join(data.disk_dir, data.domain) + ".qcow2"
-    print("Location of VM disk: {0}".format(data.disk_path))
+    if data.disk_dir:
+        disk_path = os.path.join(data.disk_dir, data.domain) + ".qcow2"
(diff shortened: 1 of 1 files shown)


dahaic:
@openscap-ci add to whitelist

dahaic:
@openscap-ci test this please

tests/install_vm.py
```diff
@@ -19,13 +19,6 @@ def parse_args():
         default="kickstarts/test_suite.cfg",
         help="Path to a kickstart file for installation of a VM."
     )
-    parser.add_argument(
-        "--distro",
```
jan-cerny: I prefer keeping this option, because it's an user friendly feature, the purpose of it is that users don't have to search for URLs of Fedora or CentOS the internet. That's convenient for them. You can use `--url` anytime to override the default hard coded URL if you want.

I suggest removing only `choices` keyword to not limit the choices of distributions and removing the `data.variant` variable, but keeping the `--distro`. Possibly, you can extend or update the hard coded URls.

comps: Ok, my problem with it was that it was directly related to `--variant` for `virt-install`; having it solely as an URL convenience (without the `virt-install` snippet) makes sense. Thanks!

comps:
Rebased, changes since last push:
- f6a8fb1: added format=qcow2 to default disk_spec; while installation would work fine, the current test suite needs internal snapshots, so it would just fail later on
- 1db3559: added back --distro as an URL alias only (fedora/centos7)
- e89cb5b: added --disk for more advanced storage specification (pools)

jan-cerny:
@comps Thank you for your patches!

PR #4329, opened 2019-05-06
Fix Debian8 and Fuse6 CPE dictionaries

yuumasato:
#### Description:

- Debian8 CPE dictionaries are missing CPE entries for `machine`, `container` and other Package CPEs
- Fuse6 Benchmark includes CPEs that are not listed in CPE Dictionary
  - The checks for the CPE aren't implemented too


jan-cerny:
I have run SCAPVal 1.3.2 on Fuse 6 and Debian 8 SCAP 1.3 Datastreams. Please see the attached HTML reports. This patch fixes the fails in both DSs.
[reports.zip](https://github.com/ComplianceAsCode/content/files/3151055/reports.zip)

PR #4330, opened 2019-05-06
Don't create blank prelink config if not installed

comps:
RHEL-7 and later doesn't use prelink and the

   printf '\n' >> /etc/sysconfig/prelink

was creating an empty file.

Even though an argument could be made to disable it "just in case somebody installs it", keeping zombie files of a no-longer-used software on the system is a bad practice.


yuumasato:
LGTM.
The OVAL check will pass if `prelink` is not installed.
And this way, the bash fix will be aligned with Ansible fix.

PR #4331, opened 2019-05-06
Remove the 'build' directory by default

comps:
Most OSS projects that use cmake let the user or a script create
the directory instead of it being pre-created.

While it might seems just like a convenience here, if you actually
remove the directory (to make a new build) and later commit with
`git commit -a`, the removal of `build/.gitkeep` will be added to
the commit.

.gitignore does not help here as it only prevents untracked files
from being added - changes in existing tracked files are still
registered.

While there are other local-only workarounds like

  git update-index --skip-worktree build/.gitkeep

a cleaner solution is removing the default build dir altogether.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,6 +90,7 @@ if ("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_BINARY_DIR}")
     message(FATAL_ERROR "In-source builds are not supported! Please use out of source builds:\n"
        "$ cd scap-security-guide\n"
        "$ rm CMakeCache.txt\n"
+       "$ mkdir build\n"
        "$ cd build\n"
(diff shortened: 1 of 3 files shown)


redhatrises:
Kind Nack. This is provided for content authors and not developers who are familiar with Cmake.

comps:
The main problem I wanted to address was the removed `.gitkeep` getting in the way of `git add -a` rather than cmake ecosystem conformance. If you don't have problems with it polluting pull requests, feel free to close this PR - the change was really meant to help with project maintenance and if it doesn't help, it's useless.

I can re-submit the `build_project` multi-job tweak separately.

redhatrises:
> I can re-submit the `build_project` multi-job tweak separately.

@comps please do submit the `build_project` multi-job tweak separately.

PR #4332, opened 2019-05-07
Rename group sap to sap_host

jan-cerny:
#### Description:
Rename group sap to sap_host

#### Rationale:
Workaround for OpenSCAP bug described in https://github.com/OpenSCAP/openscap/issues/1325
The template cannot handle situation when a profile has a same name as some group.  'sap' group conflicts with 'sap' profile in OL7 content. These settings are not about configuring SAP itself but about
configuring the underlying OS host, let's rename the group to sap_host. Another approach could be to rename the OL7 sap profile, but that could break some things for the users because they will have to start using the new ID and I have no idea about the way this profile is used by users.


yuumasato:
LGTM.
Ping @xiruiyang, this will affect existing tailorings that select or unselect group `sap`.

jan-cerny:
So would be renaming the profile instead be better? I assume not because that would also break the tailoring

yuumasato:
I believe that changing group ID has a  smaller impact than changing profile ID.

Ack.

PR #4335, opened 2019-05-09
Include existing runtime mount options in fstab

comps:
When the function added the fstab line, it completely ignored any
runtime mount options. This was specifically a problem for virtual
mounts like /dev/shm that are mounted by the kernel or by init with
ie. `rw,nosuid,nodev,seclabel'.

The fstab line enforced `rw,nodev,seclabel', lowering security.

Therefore include existing runtime mount options on the newly
created fstab line to prevent this.


yuumasato:
Looks great, thank you for this enhancement!

PR #4336, opened 2019-05-09
Use multiple jobs in build_product

comps:
Simple speedup of the build. The nproc(1) utility is from coreutils, so it should be present on the system, but use fallback if it isn't.


matusmarhefka:
@comps Nice tweak, thanks for the PR.

PR #4337, opened 2019-05-09
Use separate partition for /var/tmp in tests/kickstart

comps:
/tmp is mounted as tmpfs by systemd (and will be explicitly remediated
anyway), /var/tmp is supposed to be persistent according to FHS and is
needed for mount_option_var_tmp_nodev.

There is another rule, mount_option_var_tmp_bind, which is in conflict
with this, but its remediation script already accounts with /var/tmp
being on a separate partition, taking appropriate steps.


jan-cerny:
Thank you for your patch.

PR #4338, opened 2019-05-09
Use VM install repo URL on the installed system

comps:
Fedora doesn't need this (installs mirrors implicitly) and the list can be extended if this logic is checked to work with ie. SLES or openSUSE.

I don't like the `gpgcheck=0`, but the VM is supposed to be transient and this works with internal or otherwise temporary unsigned repositories.

### DIFF
--- a/tests/install_vm.py
+++ b/tests/install_vm.py
@@ -3,6 +3,8 @@
 import argparse
 import os
 import sys
+import subprocess
+
 
 def parse_args():
     parser = argparse.ArgumentParser()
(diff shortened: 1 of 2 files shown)


comps:
Fixed the PEP8 issue.

tests/install_vm.py
```diff
@@ -83,6 +85,25 @@ def parse_args():
     return parser.parse_args()
 
 
+def detect_url_os(url):
+    """Given a repository URL, guess an OS type and version."""
+    cmd = ["osinfo-detect", "-f", "env", "-t", "tree", url]
```
matusmarhefka: Please also add `libosinfo` package into the "Libvirt backend" section of `tests/README.md`.

comps: @matusmarhefka `libosinfo` is already used and required by `virt-install`, so if you have `virt-install` you already have `libosinfo`. In addition, the python exception takes care of cases where it's not present and the repository is simply not set up (still allowing some level of testing or manual repo setup).

Are you sure you want me to put it in `README.md`? .. Maybe as "optional"?

matusmarhefka: You are right. In that case there is no need to put the package there.

matusmarhefka:
@comps Good stuff, thanks for the extension.

PR #4340, opened 2019-05-10
Update reference links in docs

shawndwells:
- Update developer guide with links to NIST mailing lists
- Update README to remove SSG references


shawndwells:
Self nack. Looks like the docs use some format (adoc?!) instead of markdown, that does not support lists or tables in standard ways. To much effort to figure out this weird language.

PR #4342, opened 2019-05-13
Remove a message from stderr

jan-cerny:
When a remediation can't find the respective rule YML, it produces
a message on stderr, which pollutes the build output. In fact, this
is a normal situation when the remediation exists but the rule
isn't a part of the given product.

### DIFF
--- a/ssg/build_remediations.py
+++ b/ssg/build_remediations.py
@@ -195,9 +195,8 @@ class Remediation(object):
         self.associated_rule = None
 
     def load_rule_from(self, rule_path):
-        if os.path.isfile(rule_path):
-            self.associated_rule = build_yaml.Rule.from_yaml(rule_path)
-            self.expand_env_yaml_from_rule()
+        self.associated_rule = build_yaml.Rule.from_yaml(rule_path)
+        self.expand_env_yaml_from_rule()
 
     def expand_env_yaml_from_rule(self):
         if not self.associated_rule:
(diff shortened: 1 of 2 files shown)


jan-cerny:
@openscap-ci test this please

matusmarhefka:
@jan-cerny Can we rather disable the error only in case that the rule isn't a part of the given product and leave the error in other cases?

jan-cerny:
@matusmarhefka I have changed it to check the existence in the caller instead. If user uses the method in its own code with a nonexistent path it will hopefully raise him an exception. I also removed the unused code that I noticed.

jan-cerny:
@openscap-ci test this please

ssg/build_remediations.py
```diff
@@ -258,10 +246,6 @@ class BashRemediation(Remediation):
     def __init__(self, file_path):
         super(BashRemediation, self).__init__(file_path, "bash")
 
-    def load_associated_rule(self, resolved_rules_dir, rule_id):
-        # No point in loading rule for this remediation type as of now
-        pass
-
 
 class AnsibleRemediation(Remediation):
```
matusmarhefka: Add `isfile` check also into the `from_snippet_and_rule` method of this class.

jan-cerny: OK, I will do that.

jan-cerny: done

matusmarhefka:
@jan-cerny LGTM, thanks.

PR #4343, opened 2019-05-13
tests/install_vm.py: Do not abort if ostype detection fails

matusmarhefka:
#### Description:

- Fixes `tests/install_vm.py` abort in case `osinfo-detect` doesn't magage to detect OS type.


matusmarhefka:
@openscap-ci test this please

PR #4345, opened 2019-05-15
ANSSI network sysctl

shawndwells:
Add ANSSI network sysctl rules


dahaic:
Hey @shawndwells , do you plan to put more rules there, or should we merge it as it is? We can build on top of that later, if merged.

shawndwells:
seems mergable now

dahaic:
Great, thank you!

PR #4347, opened 2019-05-16
Improve error message

jan-cerny:
#### Description:
Improve error message

#### Rationale:
profile_id is always `None` here, so it never prints anything useful.


matusmarhefka:
ACK.

PR #4349, opened 2019-05-16
User guide build link to develper guide

yuumasato:
#### Description:

- Remove steps to build content from User Guide
- Refer to Developer Guide on how to build the content
- Also suggest possibility to use nightly zip from upstream

#### Rationale:

- Lets keep the build instructions in the Developer Guide, and reference it when needed.


matusmarhefka:
ACK.

PR #4353, opened 2019-05-16
Scap 1.3 content adjustments

yuumasato:
#### Description:

- Link to uncompressed external OVAL content
- Drop OCIL check for rule `security_patches_up_to_date`

#### Rationale:

- From Table 8 of SCAP 1.3 standard [1]:
  `A URI to the target component ... When referencing external content, the URI
SHALL be in the form of
scheme: ... and SHALL dereference to an XML stream that includes the SCAP
source data stream collection and the target component ....`
- From section 3.2.4.3 of [1]:
`An XCCDF benchmark MAY include a patches up-to-date rule that SHALL reference an OVAL source data stream component.`
  This should be read as only an OVAL check is allowed. Although there is no mention of OCIL check here.

[1] NIST.SP.800-126 revision 3

#### Notes:
Dropping whole OCIL check for this rule may be controversial, so I'm initially proposing this as Draft PR.

### DIFF
--- a/linux_os/guide/system/software/updating/security_patches_up_to_date/rule.yml
+++ b/linux_os/guide/system/software/updating/security_patches_up_to_date/rule.yml
@@ -74,30 +74,3 @@ oval_external_content: "https://linux.oracle.com/security/oval/com.oracle.elsa-a
 {{# The rule will be "notchecked" #}}
 {{% endif %}}
 
-ocil_clause: 'updates are not installed'
-
-{{# TODO: What about non-rpm systems? #}}
-ocil: |-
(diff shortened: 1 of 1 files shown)


yuumasato:
> Why remove the OCIL?

The presence of an OCIL check in a `security_patches_up_to_date` rule is seen as bug in the content, with regards to validation.

>     * From section 3.2.4.3 of [1]:
>       `An XCCDF benchmark MAY include a patches up-to-date rule that SHALL reference an OVAL source data stream component.`
>       This should be read as only an OVAL check is allowed. Although there is no mention of OCIL check here.

From the quote alone I don't get this interpretation, but NIST is reading the quote above as OVAL exclusive, as no other check systems allowed.

PR #4357, opened 2019-05-20
Make hardened containers smaller

dahaic:
#### Description:

- Changes to make DISA STIG hardened containers smaller.

#### Rationale:

- When hardening containers to DISA STIG compliance, there is lot of packages installed. This is unnecessary, and not desired.

- Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1711893


yuumasato:
LGTM, I don't think `aide` is need in a container.
@shawndwells what do you think? I remember you requested for it to be available for containers.

ggbecker:
@dahaic I have seen that the STIG v2r3 still references the authconfig binary in some sections even after explicitly removing the package (authconfig-gtk) installation requirement. Is it safe to assume that we can drop the authconfig-gtk? Maybe we could change to the command line version, which is `authconfig`.

shawndwells:
> LGTM, I don't think `aide` is need in a container.
> @shawndwells what do you think? I remember you requested for it to be available for containers.

Without AIDE, how will integrity checking be performed for a running container?

In the Red Hat/OpenShift realm, the integrity checking could be dismissed as a property of the container platform itself. Perhaps something involving the settings of container registries, Clair/Quay, immutable images.

But what about for other scenarios where customers are not using registries, have long running containers, and allow them to be mutable?

dahaic:
@shawndwells In other aspects of our content (remediations, check), we do not cater for convoluted or extreme customer configuration. I believe what you describe is corner case, and there's no need for that in the content. Not to mention that the use of AIDE would be something like:
* aide is installed
* aide database is created when container starts
* aide is executed periodically and logs are send to the container host

I don't think this is a case we want to solve from inside of the container :)

shawndwells:
> @shawndwells In other aspects of our content (remediations, check), we do not cater for convoluted or extreme customer configuration. I believe what you describe is corner case, and there's no need for that in the content. Not to mention that the use of AIDE would be something like:
> 
> * aide is installed
> * aide database is created when container starts
> * aide is executed periodically and logs are send to the container host
> 
> I don't think this is a case we want to solve from within the container :)

How would this be different than service-level configuration checks? e.g. those that say "if sshd is installed, then evaluate it"?

"if PAM installed, then ensure smart card auth enabled"

shawndwells:
> @shawndwells In other aspects of our content (remediations, check), we do not cater for convoluted or extreme customer configuration. I believe what you describe is corner case, and there's no need for that in the content. Not to mention that the use of AIDE would be something like:
> 
> * aide is installed
> * aide database is created when container starts
> * aide is executed periodically and logs are send to the container host
> 
> I don't think this is a case we want to solve from within the container :)

So then, sounds like AIDE has to stay.

dahaic:
@shawndwells no, that means AIDE in the container itself is useless.

shawndwells:
> @shawndwells no, that means AIDE in the container itself is useless.

Huh? Can you answer more directly then:  Without AIDE, how will integrity checking of running containers be performed? This is a requirement for all U.S. Government systems, all HIPAA systems, all PCI-DSS, etc.

dahaic:
I have a feeling that using AIDE inside of the container is not the proper way how to perform integrity checking. That is should be solved on the container host level.

As I am not expert on containers to propose the solution right away, and AIDE is better than nothing, I will revert AIDE part of this PR :)

matusmarhefka:
@shawndwells @dahaic So for AIDE you should create a baseline database of files on initial run. Then it is recommended to copy the database to some read-only media and perform checks against this copy.

IMO, it does not make sense in containers. Imagine your container gets hacked and an intruder can modify all files in the hacked container. If you would store the AIDE database in such a container it can be modified as well and you would not detect any intrusion at all.

I think you should run AIDE on the host. Creating a baseline database for containers can be done by mounting them to some well-known location on the host so the host database contains entries for these containers too. Then, for example, backup the created database to read-only mount and use it to perform checks (so the database cannot be altered). When performing checks, you would of course first mount the containers filesystems to the well-know locations as you did first time when creating the baseline database. Then perform the integrity checking.

The other option you have which is IMO better for containers is to use `docker diff`/`podman diff` which should list all modified files in a container compared to the container image from which the container has been started. For example removing `ls` command in a container would produce this:
```
podman diff 7b4d329b8fc0
C /usr
C /usr/bin
D /usr/bin/ls
```

dahaic:
Thank you @matusmarhefka . `podman diff` command seems to be much better approach.

matusmarhefka:
@dahaic Btw add `platform: machine` rather into the `linux_os/guide/system/software/integrity/software-integrity/aide/group.yml` and remove it from all the `rule.yml` files inside that group.

redhatrises:
@dahaic can you add `platform: machine` for the following package installs:

1. opensc
2. esc
3. dracut-fips
4. pam_pkcs11

mbach04:
Is it possible to get squashing built into the remediation as well? The resultant containers have a lot of history in them.

jnach:
Solutions such as Sysdig and TwistLock have risen to address container integrity from a platform level, combined with clair/quay and other variants of integrity checking, the preferred way to do this is from the platform, not the image, which should be immutable by best practice anyway.

I think squashing would be good in remediation too. There are three ways to perform this that I'm aware of:
- `docker export <container_id> | docker import - <squashed name>`
- docker-squash utility, via `docker save <image id> | sudo docker-squash -t newtag | docker load`. This requires a newer build of docker, or experimental mode on older versions. (OpenShift docker is locked at 1.13 at the time of this writing, 3.11, and not out of the box compatible here)
- As a flag during the build option, which may not apply to remediation via `docker build --squash <imagename>`

dahaic:
Now, the possible squashing won't be solved in this pull request :) But I have a question - my understanding is that hardening of the container will be, for the most of the time, just one of the steps in CI/CD pipeline, with cleanup and squashing being separate step, done by specialized tool.

Is that wrong assumption?

dahaic:
Just a note: with this PR, remediation of UBI7 to DISA STIG profile does not install any new packages. (207MB in comparison to 204MB of UBI7 image)

mbach04:
> Now, the possible squashing won't be solved in this pull request :) But I have a question - my understanding is that hardening of the container will be, for the most of the time, just one of the steps in CI/CD pipeline, with cleanup and squashing being separate step, done by specialized tool.
> 
> Is that wrong assumption?

No, you're correct. Just raising the issue here. We're certainly doing this on our end, but we should at least (as a documentation effort) notify users of the --remediate flag that their container history left unchecked will leave some unwanted bloat.

matusmarhefka:
I think all the issues have been addressed. Merging, thanks @dahaic

PR #4361, opened 2019-05-21
Add rhel7 ccc (Common Criteria Certification) profile

comps:
This profile is a draft for a future CC certification of RHEL, based on the OSPP (NIAP) requirements. It includes functionality (rules) beyond the OSPP scope itself.

### DIFF
--- a/rhel7/CMakeLists.txt
+++ b/rhel7/CMakeLists.txt
@@ -13,6 +13,7 @@ ssg_build_html_table_by_ref(${PRODUCT} "cis")
 ssg_build_html_table_by_ref(${PRODUCT} "cui")
 ssg_build_html_table_by_ref(${PRODUCT} "pcidss")
 ssg_build_html_table_by_ref(${PRODUCT} "anssi")
+ssg_build_html_table_by_ref(${PRODUCT} "ospp")
 
 ssg_build_html_nistrefs_table(${PRODUCT} "standard")
 ssg_build_html_nistrefs_table(${PRODUCT} "ospp")
--- a/rhel7/profiles/ccc.profile
+++ b/rhel7/profiles/ccc.profile
@@ -1,0 +1,271 @@
+documentation_complete: true
+
+title: 'DRAFT - Common Criteria Certification profile using OSPP'
+
+description: |-
+    This profile is a draft for a future CC certification of RHEL, based on the
+    OSPP (NIAP) requirements. It includes functionality (rules) beyond the OSPP
+    scope itself.
+
+selections:
+    ### Boot
+    - grub2_uefi_password
+
+    # systemd
+    - require_singleuser_auth
+    - grub2_disable_interactive_boot
+    - disable_ctrlaltdel_reboot
+    - disable_ctrlaltdel_burstaction
+    - service_debug-shell_disabled
+    - service_kdump_disabled
+    - service_autofs_disabled
+
+    # Partitioning
+    - mount_option_home_nodev
+    - mount_option_home_nosuid
+    - mount_option_tmp_nodev
+    - mount_option_tmp_noexec
+    - mount_option_tmp_nosuid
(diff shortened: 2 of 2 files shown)


rhel7/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ ssg_build_html_nistrefs_table(${PRODUCT} "standard")
 ssg_build_html_nistrefs_table(${PRODUCT} "ospp")
 ssg_build_html_nistrefs_table(${PRODUCT} "C2S")
 ssg_build_html_nistrefs_table(${PRODUCT} "stig")
+ssg_build_html_nistrefs_table(${PRODUCT} "ccc")
```
yuumasato: Probably `ospp` references would be more useful.
Right now, there is no version of `ssg_build_html_nistrefs_table` for `ospp` refs.

An alternative is to generate a product table for `ospp` references:
`ssg_build_html_table_by_ref(${PRODUCT} "ospp")`
 Although the table is not per profile, but for whole datastream, I think it will be more useful.

rhel7/profiles/ccc.profile
```diff
@@ -0,0 +1,271 @@
...
+    - dconf_gnome_login_banner_text
+    - disable_users_coredumps
+    - var_accounts_max_concurrent_login_sessions=10
+    - accounts_max_concurrent_login_sessions
+    - securetty_root_login_console_only
+    - var_password_pam_unix_remember=5
+    - accounts_password_pam_unix_remember
+
+    # Password
+#    - no_empty_passwords   # duplicate in Login
```
yuumasato: I think you can drop this comment.
Unless you would like to keep track of this for some reason.

rhel7/profiles/ccc.profile
```diff
@@ -0,0 +1,271 @@
...
+    - sshd_disable_root_login
+    - sshd_enable_strictmodes
+    - disable_host_auth
+    - sshd_disable_user_known_hosts
+    - sshd_disable_rhosts
+    - sshd_disable_empty_passwords
+    - sshd_disable_kerb_auth
+    - sshd_disable_gssapi_auth
+    - sshd_idle_timeout_value=10_minutes
+#    - sshd_set_idle_timeout  # duplicate, see below
```
yuumasato: I think you can drop this comment.
Unless you would like to keep track of this for some reason.

rhel7/CMakeLists.txt
comps: There is already the same line few lines above my addition. Should I therefore just remove this addition?

yuumasato: What I suggested is different actually: 
`ssg_build_html_nistrefs_table` vs
`ssg_build_html_table_by_ref`

comps:
Comments kept per discussion, added a second commit for `ssg_build_html_table_by_ref`.

yuumasato:
Ack.

shawndwells:
Why is there a separate ``ccc`` profile vs OSPP?

Also - this does not cover the requirements for Common Criteria. Needs several updates to be inline with OSPP and the configuration annexes.

yuumasato:
> Why is there a separate `ccc` profile vs OSPP?

As far as I understand and observe, the OSPP Profile and the CCC Profile have a different take on what is defined in `pp_os_v4.2.1`.

They cannot be the same, the current set of rules in both profile are significantly different.
Check [ccc_sub_ospp.txt](https://github.com/ComplianceAsCode/content/files/3301482/ccc_sub_ospp.txt) for rules that re present only in RHEL7 `ccc` Profile.
Check [ospp_sub_ccc.txt](https://github.com/ComplianceAsCode/content/files/3301483/ospp_sub_ccc.txt) for rules that are only present in RHEL7 `ospp` Profile.

*diff generated with `build-scripts/profile_tool.py`.

PR #4366, opened 2019-05-22
[WIP] Fix OVAL checks for gnome rules and add dconf gdm profile rule.

ggbecker:
#### Description:

- Address OVAL checks for gnome rules. This PR is in a very initial state.

#### Rationale:

- Previous OVAL checks generate false positives because the file which is checked doesn't reflect the actual state of dconf's database. This is tightly connected to the rule which sets the dconf to use text backend which allows us to probe static files for dconf's configuration.


linux_os/guide/system/accounts/accounts-banners/gui_login_banner/dconf_gnome_banner_enabled/ansible/shared.yml
```diff
@@ -5,15 +5,15 @@
...
+    dest: "/etc/dconf/db/gdm.d/00-security-settings"
     section: "org/gnome/login-screen"
     option: banner-message-enable
     value: "true"
     create: yes
 
 - name: "Prevent user modification of GNOME banner-message-enabled"
   lineinfile:
-    path: /etc/dconf/db/local.d/locks/00-security-settings-lock
+    path: /etc/dconf/db/gdm.d/locks/00-security-settings-lock
```
ggbecker: This lock mechanism doesn't work when the dconf is set to use text backend

linux_os/guide/system/accounts/accounts-banners/gui_login_banner/dconf_gnome_banner_enabled/oval/shared.xml
```diff
@@ -27,8 +27,8 @@
   </ind:textfilecontent54_test>
   <ind:textfilecontent54_object id="obj_banner_gui_enabled"
   version="1">
-    <ind:path>/etc/dconf/db/gdm.d/</ind:path>
-    <ind:filename operation="pattern match">^.*$</ind:filename>
+    <ind:path>/var/lib/gdm/.config/dconf</ind:path>
```
redhatrises: @ggbecker why this change? If I can configure `/etc` is the top of the tree, that should be checked over `/var`.

ggbecker: I'm not sure if I understood what you are trying to say. But files under `/etc/dconf/db/gdm.d/` serve only the purpose of input configuration for `dconf`. If someone changes the database by other means, this configuration won't reflect to files under `/etc/dconf/db/gdm.d/`. That's why we are changing dconf to use text backend which synchronizes the binary database to the file `/var/lib/gdm/.config/dconf/user.txt` and that's why we are changing the check to this file.

linux_os/guide/system/accounts/accounts-banners/gui_login_banner/dconf_gnome_banner_enabled/ansible/shared.yml
matejak: As of 23rd of May, we are not sure about this. We are in contact with dconf experts, who also have to look things up before they tell us how we are.

linux_os/guide/system/accounts/accounts-banners/gui_login_banner/dconf_gnome_banner_enabled/oval/shared.xml
matejak: @ggbecker is correct, I will rephrase a bit: If (and only if) the configuration under `/etc/dconf/db/...` specifies BOTH the respective key values AND those values are configured to be locked, subsequent execution of `dconf` will make sure that this configuration is applied, and local user settings are ignored.
However, OVAL can't check in detail what `dconf` did, as it saves it's database to a binary file. We are able to check the DB if it is switched to a text format, but it may be that in that case, we lose the possibility of locking settings - see the conversation above.
In any way, the GDM settings are not user settings, so the respective database is  located in `/var/lib/gdm/.config/dconf/`, and if the backend is set to text keyfiles, it is the `/var/lib/gdm/.config/dconf/user.txt` file. We have verified this on a VM, and it indeed works like this.

matejak:
We should not considering merge before we are sure that setting dconf backend to text keyfiles doesn't break other dconf rules, as we suspect that it may interfere with locking.

linux_os/guide/system/accounts/accounts-banners/gui_login_banner/dconf_gnome_banner_enabled/oval/shared.xml
redhatrises: A couple of observations here:
1. The OVAL checks need to check the `/etc` configuration (static configuration) for the system administrator configuration in addition to the `/var/lib` (runtime configuration) checks. 
2. Having a Dconf probe is becoming more necessary.

linux_os/guide/system/software/gnome/dconf_use_text_backend_gdm_profile/rule.yml
```diff
@@ -0,0 +1,37 @@
...
+    By default, DConf uses a binary database as a data backend.
+    The database is compiled from config files by the <pre>dconf update</pre> command.
+
+    dconf can be configured to look into those text files directly by inserting the
+    <pre>service-db:keyfile/user</pre>
+    directive at the beginning of the <pre>/etc/dconf/profile/user</pre> and
+    <pre>/usr/share/dconf/profile/gdm</pre> files.
+
+rationale: |-
+    Unlike text config files, the binary database is impossible to check by OVAL.
```
redhatrises: `Unlike text config files, the binary database is impossible to check by OVAL.` should be `Because a dconf OVAL probe does not exist yet, the binary database cannot be evaluated by OVAL.`

ggbecker:
Closing as we are taking different approach to address false positives in some dconf rules.

PR #4367, opened 2019-05-23
User Guide: Update table of old content

yuumasato:
#### Description:

- Update list of products to which content is not developed

#### Rationale:

- Webmin is not deprecated, but its content is not developed by community
anymore.


redhatrises:
LGTM. Thanks @yuumasato!

PR #4370, opened 2019-05-23
Update RHEL to Red Hat Enterprise Linux in DISA STIG profile and add language for containers

redhatrises:


matusmarhefka:
LGTM.

PR #4372, opened 2019-05-27
Update Fedora CPE check applicability also for Fedora 30

matusmarhefka:
#### Description:

* Since Fedora 30 there is no more a single `fedora-release` RPM
  package. There is `fedora-release-common` and then the specific
  package for the type of realease, for example
  `fedora-release-workstation`.


jan-cerny:
It works for me on both Fedora 29 and Fedora 30 with both SCAP 1.2 and SCAP 1.3 datastreams.

PR #4375, opened 2019-05-27
Remove dconf_use_text_backend rule from profiles.

ggbecker:
#### Description:

Our OVAL probe for checking dconf settings does not work properly thus rule dconf_use_text_backend is not applicable. Removing rule related content until we came up with proper solution (either dconf probe in OpenSCAP or checking right files when setting dconf to use text backend [draft here: #4366]).


matusmarhefka:
LGTM. Thanks @ggbecker

PR #4376, opened 2019-05-27
Remove dconf oval checks

ggbecker:
#### Description:

- This PR aims to remove false positives generated by dconf OVAL checks. The OVAL implementation checks static configuration files (/etc/dconf/db/<database>.d/*) which don't reflect the actual database state. To proper solve the issue it is needed to have either the dconf probe in OpenSCAP or check the right files when the dconf is set to use text backend (initial draft: #4366). This PR is strongly related to #4375 but it makes sense to keep it separated so it is easier to review.

OCIL checks were revisited as well because some of them instructed the user to perform the same as the OVAL check.

#### Rationale:

Avoid false positives from dconf rules.


redhatrises:
Kinda NACK. Static configuration checks are required.

ggbecker:
Closing as we are taking different approach to detect when the dconf's binary database is changed.

PR #4377, opened 2019-05-27
Smartcards auth in OL8 should be done via sssd 

iokomin:
#### Description:

Smartcards auth in OL8 should be done via sssd 
- Enable smartcard services and packages on OL
- Added OL support for force_opensc_card_drivers rule
- Add OL support for configure_opensc_card_drivers rule
- Fixed configure_opensc_card_drivers and force_opensc_card_drivers remediation to use opensc-tool
- Add OL support for sssd_enable_smartcards rule
- Add OL7 support for configure_opensc_nss_db rule
- Remove OL8 support from smartcard_auth rule
- Update OL8 profiles to do smartcards auth via sssd 

#### Testing:
- Checked ol7,ol8 builds to pass
- Checked rules and remediation to pass on OL8

### DIFF
--- a/linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/configure_opensc_card_drivers/bash/shared.sh
+++ b/linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/configure_opensc_card_drivers/bash/shared.sh
@@ -1,4 +1,4 @@
-# platform = Red Hat Enterprise Linux 7,Red Hat Enterprise Linux 8,multi_platform_fedora,multi_platform_rhv
+# platform = Red Hat Enterprise Linux 7,Red Hat Enterprise Linux 8,multi_platform_fedora,multi_platform_rhv,multi_platform_ol
 # reboot = false
 # strategy = configure
 # complexity = low
@@ -7,8 +7,8 @@
 . /usr/share/scap-security-guide/remediation_functions
 populate var_smartcard_drivers
 
-grep -qs "card_drivers =" /etc/opensc*.conf && \
-	sed -i "s/card_drivers =.*/card_drivers = $var_smartcard_drivers;/g" /etc/opensc*.conf
-if ! [ $? -eq 0 ]; then
-	sed -i "s/.*card_drivers =.*/        card_drivers = $var_smartcard_drivers;/g" /etc/opensc*.conf
+OPENSC_TOOL="/usr/bin/opensc-tool"
+
+if [ -f "${OPENSC_TOOL}" ]; then
+    ${OPENSC_TOOL} -S app:default:card_drivers:$var_smartcard_drivers
 fi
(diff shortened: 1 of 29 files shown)


linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/force_opensc_card_drivers/bash/shared.sh
```diff
@@ -7,8 +7,8 @@
...
 populate var_smartcard_drivers
 
-grep -qs "force_card_driver =" /etc/opensc*.conf && \
-	sed -i "s/force_card_driver =.*/force_card_driver = $var_smartcard_drivers;/g" /etc/opensc*.conf
-if ! [ $? -eq 0 ]; then
-	sed -i "s/.*force_card_driver =.*/        force_card_driver = $var_smartcard_drivers;/g" /etc/opensc*.conf
+OPENSC_TOOL="/usr/bin/opensc-tool"
+
+if [ -f "${OPENSC_TOOL}" ]; then
+    ${OPENSC_TOOL} -S app:default:force_card_driver:$var_smartcard_drivers
```
Jakuje: This configuration option does not have any effect in RHEL8 so I do not think it makes sense to bother with it. By specifying the `card_drivers` to `cac`, setting the `force_card_driver` to the same value does not have any effect in RHEL7 either so I propose to drop this step completely (probably as part of different housekeeping PR).

linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/configure_opensc_card_drivers/bash/shared.sh
```diff
@@ -7,8 +7,8 @@
...
 populate var_smartcard_drivers
 
-grep -qs "card_drivers =" /etc/opensc*.conf && \
-	sed -i "s/card_drivers =.*/card_drivers = $var_smartcard_drivers;/g" /etc/opensc*.conf
-if ! [ $? -eq 0 ]; then
-	sed -i "s/.*card_drivers =.*/        card_drivers = $var_smartcard_drivers;/g" /etc/opensc*.conf
+OPENSC_TOOL="/usr/bin/opensc-tool"
+
+if [ -f "${OPENSC_TOOL}" ]; then
+    ${OPENSC_TOOL} -S app:default:card_drivers:$var_smartcard_drivers
```
Jakuje: Looks good to me. This feature is present in OpenSC for 10+ years so I think we can depend on that in the scripts. Indeed, it looks nicer than the seds.

ol8/profiles/pci-dss.profile
```diff
@@ -110,7 +110,13 @@ selections:
     - display_login_attempts
     - gid_passwd_group_same
     - grub2_audit_argument
-    - smartcard_auth
+    - package_opensc_installed
+    - var_smartcard_drivers=cac
```
Jakuje: One more note, in RHEL8, there is a new driver for legacy CAC version 1 cards, which might need to be enabled too if these cards are in use.

iokomin: @Jakuje, thanks a lot for your CAC1 note, I'll add this to OL smartcards support research tasks. Suggested set of changes in general aimed to sync existing RHEL smartcards rules with supported features on OL.

iokomin:
@jan-cerny, thanks for merging this PR!

PR #4379, opened 2019-05-28
Fix deprecated IncludeConfig directive in rsyslog_files OVALs

### LINKED ISSUE #4363: Fix rsyslog_files rules for rsyslog v8.33.0
#### Description of problem:
Existing rsyslog_files rules and bash remediation scripts rely on $IncludeConfig directive which is deprecated starting from v8.33.0, include() object should be used instead:
https://www.rsyslog.com/doc/v8-stable/rainerscript/include.html

Affected tests:
- rsyslog_files_groupownership
- rsyslog_files_ownership
- rsyslog_files_permissions

v0.1.43

Oracle Linux 8
Assume RHEL8 and latest Fedora should be affected as well

#### Steps to Reproduce:
1. run rsyslog_files_permissions evaluation
2. 
3.

#### Actual Results:
OVAL definition marked as failed with "Referenced variable has no values (oval:ssg-var_rfp_log_files_paths:var:1)."

#### Expected Results:
Evaluation passed

Bash remediation script for rsyslog_files_permissions should be also revised to take into account  include() object

iokomin:
#### Description:
Fixes #4363 - $IncludeConfig directive is deprecated starting from rsyslog v8.33.0,
include() object should be used instead on RHEL8 and OL8.

Signed-off-by: Ilya Okomin <ilya.okomin@oracle.com>

#### Testing:
-Checked builds for ol7,ol8,rhel8
-Checked rules evaluation on ol8

### DIFF
--- a/linux_os/guide/system/logging/ensure_rsyslog_log_file_configuration/rsyslog_files_groupownership/oval/shared.xml
+++ b/linux_os/guide/system/logging/ensure_rsyslog_log_file_configuration/rsyslog_files_groupownership/oval/shared.xml
@@ -22,36 +22,57 @@
 
   </definition>
 
-  <!-- First obtain rsyslog's $IncludeConfig directive and include() object (introduced in rsyslog v8.33.0) values.
-       To workaround empty include objects case, when FunctionGroup operations return "does not exist" result, added empty string match -->
+  <!-- First obtain rsyslog's $IncludeConfig directive and include() object (introduced in rsyslog v8.33.0) values.  -->
   <ind:textfilecontent54_object id="object_rfg_rsyslog_include_config_value" comment="rsyslog's $IncludeConfig directive and include() object values" version="1">
     <ind:filepath>/etc/rsyslog.conf</ind:filepath>
-    <ind:pattern operation="pattern match">^(?:include\([\n\s]*file="([^\s;]+)".*|\$IncludeConfig[\s]+([^\s;]+)|())$</ind:pattern>
+    <ind:pattern operation="pattern match">^(?:include\([\n\s]*file="([^\s;]+)".*|\$IncludeConfig[\s]+([^\s;]+))$</ind:pattern>
     <ind:instance datatype="int" operation="greater than or equal">1</ind:instance>
   </ind:textfilecontent54_object>
 
   <!-- Turn that glob value into Perl's regex so it can be used as filepath pattern below -->
   <local_variable id="var_rfg_include_config_regex" datatype="string" version="1" comment="$IncludeConfig value converted to regex">
-    <glob_to_regex>
-      <object_component item_field="subexpression" object_ref="object_rfg_rsyslog_include_config_value" />
-    </glob_to_regex>
+    <unique>
+      <glob_to_regex>
+        <object_component item_field="subexpression" object_ref="object_rfg_rsyslog_include_config_value" />
+      </glob_to_regex>
+    </unique>
   </local_variable>
 
-  <!-- Create collection of unique include_config regex values together with '/etc/rsyslog.conf'
-       path literal to obtain final list of all rsyslog's configuration files
-  -->
-  <local_variable id="var_rfg_all_log_files_as_collection" datatype="string" version="1" comment="Locations of all rsyslog configuration files as collection">
-    <unique>
-      <variable_component var_ref="var_rfg_include_config_regex" />
-      <literal_component datatype="string">^/etc/rsyslog.conf$</literal_component>
-    </unique>
+  <!-- Create a variable_object from the regex variable
+       If the variable has no values, there won't be any objects -->
+  <ind:variable_object id="object_var_rfg_include_config_regex" comment="Make variable object from regex variable" version="1">
+    <ind:var_ref>var_rfg_include_config_regex</ind:var_ref>
+  </ind:variable_object>
+
+  <local_variable id="var_rfg_syslog_config" datatype="string" version="1" comment="Locations of all rsyslog configuration files as collection">
+    <literal_component datatype="string">^/etc/rsyslog.conf$</literal_component>
+  </local_variable>
+
+  <ind:variable_object id="object_var_rfg_syslog_config" comment="Make variable object for use" version="1">
+    <ind:var_ref>var_rfg_syslog_config</ind:var_ref>
+  </ind:variable_object>
+
+  <!-- Combine the two variable_objects into one variable_object
+       We do it this way to avoid referencing an empty variable in a state comparison, which
+       will cause a test to evaluate to fail. Combining an empty set of objects is fine though -->
+  <ind:variable_object id="object_var_rfg_all_log_files" comment="Filter out empty string" version="1">
+    <set>
+      <object_reference>object_var_rfg_include_config_regex</object_reference>
+      <object_reference>object_var_rfg_syslog_config</object_reference>
+    </set>
+  </ind:variable_object>
+
+  <!-- In element filepath of object_rfg_log_files_paths we need to pass a list of values,
+       a list of objects won't do. So we make a local_variable from the variable_objects. -->
+  <local_variable id="var_rfg_all_log_files" datatype="string" version="1" comment="Locations of all rsyslog configuration files as collection">
+    <object_component object_ref="object_var_rfg_all_log_files" item_field="value"/>
   </local_variable>
 
   <!-- For each item from that collection (particular rsyslog's configuration file path) search
        that rsyslog's configuration file to select file paths for log files directives
   -->
   <ind:textfilecontent54_object id="object_rfg_log_files_paths" comment="All rsyslog configuration files" version="1">
-    <ind:filepath operation="pattern match" var_ref="var_rfg_all_log_files_as_collection" var_check="at least one" />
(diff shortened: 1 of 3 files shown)


jan-cerny:
What if the user still uses the deprecated `$IncludeConfig` keyword? My understanding is that it's still possible to use it, it's only deprecated. We probably should check both `$IncludeConfig` and `include()`.

iokomin:
@jan-cerny , thanks for your point, I think you are right and usage of both options might have place according to docs. I'll read more on how these two guys can be used together in rsyslog. 
If anyone already has a clear view on result rsyslog behavior - suggestions would be much appreciated.

jvymazal:
My recommendation would be to definitely check both, while the new syntax is definitely superior, it is still safe to use both and rsyslog users are generally quite hesitant to make config changes and updates so currently you probably still encounter more of the old syntax.

iokomin:
@jvymazal thanks for your input, I'll take it into account.

iokomin:
@jan-cerny please find alternative implementation to support both rsyslog IncludeConfig directive and include() object.
Changes:
- Updated pattern to match both $IncludeConfig and include() entries
- Empty string added to this pattern to workaround FunctionGroup operations failure when no match found
- concatenation+split replaced by unique operation

Testing on OL8:
- checked /etc/rsyslog.conf with commented IncludeConfig and include lines (no match)
- checked /etc/rsyslog.conf with $IncludeConfig, multiline include() and single line include() entries

Note: 
I might think of rsyslog.conf file without empty lines and include entries where empty string workaround might not work. However, I suppose it would be a non-common case and it can be addressed in separate PR if needed as it requires more text processing. Thoughts?

matejak:
The problem with this OVAL is its complexity and the relative inability to test it. Do you think that you could check out [the test suite](https://github.com/ComplianceAsCode/content/tree/master/tests) that is capable to test the OVAL in practice? If test scenarios are supplied, we will know what use cases the OVAL is able to handle and what not.

iokomin:
@matejak thanks for your review. Honestly, suggested solution complexity is the same as the original one - only added alternative regex, also fixed existing issues: only one entry taken into account and failure when no entries matched. Anyway I'll take a look what could be added to the test suites list, definitely makes sense.

iokomin:
@matejak please find test scenarios provided, see #4743
Note: Checked to work on OL backends, but should also be applicable on RHEL and Fedora.

mildas:
@matejak This PR is ready to review. I merged test scenarios, so you can test it easily. Only ".fail.sh" test scenarios give ERROR during remediation part, because we don't have remediations for these rules.

@iokomin Thank you for creating tests! Wouldn't you like to also create remediations for them?

iokomin:
@mildas thanks a lot for your review and checks. As for remediation scripts I'll give it a try.

NicolasMarcotte:
@iokomin Concerning RHEL8,  I applied the regex directly in /usr/share/xml/scap/ssg/content/ssg-rhel8-ds.xml , I wanted to see if it fixed that problem on fresh RHEL8 install and it did.

But it now produce the following warning: `E: oscap:     Failed to validate the pattern: pcre_exec() returned PCRE_ERROR_NOMATCH for pattern '^$' and a test path '/'. This indicates the pattern doesn't match a leading '/'.` when I run ` oscap xccdf eval    --rule xccdf_org.ssgproject.content_rule_rsyslog_files_groupownership  --oval-results --profile xccdf_org.ssgproject.content_profile_pci-dss /usr/share/xml/scap/ssg/content/ssg-rhel8-ds.xml` as root.

On my system that warning can be suppressed without side-effects by adding a / inside the empty () to produce the following regex : `^(?:include\([\n\s]*file="([^\s;]+)".*|\$IncludeConfig[\s]+([^\s;]+)|(/))$`

yuumasato:
@NicolasMarcotte I don't quite understand how the `/` fixed the openscap error.

Maybe taking a step back can help me. Is the regex trying to match the following?
- `include(file="some/path")`
- `$IncludeConfig (some/path)`
- `$IncludeConfig ()`

iokomin:
@NicolasMarcotte thanks for pointing to this warning. 
Empty line match was introduced to workaround /etc/rsyslog.conf case when no include|$IncludeConfig match found - OVAL concat|unique operation returns "No values" if just one of it's arguments has no matched values (thus with empty line we'll have at least one for 99% sure).
Proposed regex change unfortunately won't work for the case when /etc/rsyslog.conf doesn't include any of include|$IncludeConfig lines. I'll take a look if this workaround could be workarounded to avoid warnings :)

iokomin:
@yuumasato, I've checked DEVEL log and appears openscap ./src/OVAL/probes/oval_fts.c -> process_pattern_match() has sanity check limitation which requires file path pattern to match leading "/". Is there any elegant OVAL logic could be suggested to workaround rsyslog.conf case with empty include|$IncludeConfig set?

yuumasato:
> @yuumasato, I've checked DEVEL log and appears openscap ./src/OVAL/probes/oval_fts.c -> process_pattern_match() has sanity check limitation which requires file path pattern to match leading "/".

@iokomin Interesting, so the problem is that the empty string is fed into `textfilecontent54_object` `filepath` `pattern_match` operation in [this line](https://github.com/ComplianceAsCode/content/pull/4379/files#diff-086f68e54c6fb45f0799567beb932498R54).

One thing that might work is to take `var_rfg_all_log_files_as_collection`, and filter out the empty string before using it in `object_rfg_log_files_paths`. I think this will require creation of an intermediate object and use of filters.
As `var_rfg_all_log_files_as_collection` concatenates a `literal_component`, we know that removing the empty string will not create an empty variable.

iokomin:
@yuumasato  thanks for looking into it and your idea is absolutely clear to me. However could you please give a hint, how proposed filtering can be implemented? If you had working example or reference on existing piece of SSG OVAL code that would be really helpful to close this PR. Unfortunately I'm running out of ideas how to do this while these rules would be better to fix for a number of platforms and critical profiles.

shawndwells:
bump to @yuumasato

shawndwells:
and +cc to @matejak as owner of this PR

yuumasato:
> @yuumasato thanks for looking into it and your idea is absolutely clear to me. However could you please give a hint, how proposed filtering can be implemented? If you had working example or reference on existing piece of SSG OVAL code that would be really helpful to close this PR. Unfortunately I'm running out of ideas how to do this while these rules would be better to fix for a number of platforms and critical profiles.

@iokomin Sorry, I don't have any working example. I'm not even sure my hackish idea works.
I may have time to tinker with it later this week.

yuumasato:
@iokomin Hello, my original idea didn't work because `var_rfg_include_config_regex` returns one variable with multiple values. Because of the they way filters work (they remove the whole variable) the approach I mentioned would only work if there were multiple values with one value each.

But I came up with a solution that worked for me.
I don't see the issue with pcre anymore, but now it causes the following, :smiley: :
```
W: oscap:     Obtrusive data from probe!
W: oscap:     Obtrusive data from probe!
```
@jan-cerny Do you remember what these could be?

I expect build of this to fail without #5193

yuumasato:
@iokomin Hope you don't mind that I pushed to your PR.
After #5193 is merged, could you rebase into latest master and check if it works for you?
Rebasing would also give easy access to the test scenarios.

iokomin:
@yuumasato thanks for your time and sure your pushes are just welcome. I will definitely rebase and check your solution. The only question you can probably answer right away (that case which required workaround and I failed to resolve):
In your push if "object_var_rfg_include_config_regex" is empty, will "object_var_rfg_all_log_files" set be non empty containing value from "object_var_rfg_syslog_config"?

yuumasato:
> In your push if "object_var_rfg_include_config_regex" is empty, will "object_var_rfg_all_log_files" set be non empty containing value from "object_var_rfg_syslog_config"?

Yes, the `<set>` will merge the two collection of objects, and it doesn't care if one of them is an empty set.. By default the operator is  `union`.

yuumasato:
> now it causes the following,:
> 
> ```
> W: oscap:     Obtrusive data from probe!
> W: oscap:     Obtrusive data from probe!
> ```

There is https://bugzilla.redhat.com/show_bug.cgi?id=1722003 for this.

yuumasato:
@iokomin #5193 was merged. Could you rebase onto latest master? Thanks

iokomin:
@yuumasato awesome, checked your <set> solution to work as expected! dev branch rebased onto latest master.
Thanks a lot for your help with this PR.

yuumasato:
> @yuumasato awesome, checked your solution to work as expected! dev branch rebased onto latest master.

Great!

@iokomin Do you still plan to add remediations to this PR?
If not, or if it would take some time, I would merge this.
Let's not hold this much longer.

iokomin:
@yuumasato let's merge this OVAL fix while remediation part could be fixed as part of separate issue/PR. I have remediation work in plans however at the moment I have bunch of other content tasks w.r.t. OL8 release. Will do my best to look at remediation part as soon as rush is over.
Thanks again for your review and suggested fix.

jan-cerny:
@yuumasato It's a warning message which meaning is unclear. It appears often when you use sets or variables. My understanding is that if `oscap` produces expected OVAL results the message could be removed.

PR #4382, opened 2019-05-30
Introduced the "DConf System DBs are in sync with keyfiles" rule.

matejak:
The new rule is the missing piece of the puzzle when we talk about verifying dconf settings.
Our OVAL tests can check keyfiles, and this rule makes sure that the respective DB is up-to-date, e.g. that `/etc/dconf/db/local` is newer than files in the `/etc/dconf/db/local.d` directory.

More info about dconf and system settings: https://help.gnome.org/admin/system-admin-guide/stable/dconf-keyfiles.html.en

The rule

- [x] is included in all profiles applicable,
- [x] is parametrized wrt RHEL/Fedora differences, and
- [x] handles empty keyfile directories gracefully (i.e. `/etc/dconf/db/local.d` with no files whatsoever means that the `/etc/dconf/db/local` part of the rule should pass).


ggbecker:
> The new rule is the missing piece of the puzzle when we talk about verifying dconf settings.
> Our OVAL tests can check keyfiles, and this rule makes sure that the respective DB is up-to-date, e.g. that `/etc/dconf/db/local` is newer than files in the `/etc/dconf/db/local.d` directory.
> 
> More info about dconf and system settings: https://help.gnome.org/admin/system-admin-guide/stable/dconf-keyfiles.html.en
> 
> TODO: The rule
> 
>     * [ ]  is included in all profiles applicable,
> 
>     * [ ]  is parametrized wrt RHEL/Fedora differences, and
> 
Fedora does not contain the directory `/etc/dconf/db/gdm.d/` and the database by default (which is provided by `gdm` package in RHEL7). This would cause the OVAL check to fail because it checks for both `gdm.d` and `local.d` and its respective binary databases (`gdm` and `local`).

I have tested configuring the gnome login banner using `local.d` and it works in both fedora29 and RHEL7.7. This means that we are able to change all of our rules (approx. 4-5) which configures `dconf` using `gdm.d`. This would be one way to simplify the rules and use only `local.d` to configure `dconf`. Thus reducing the complexity of the proposed rule. What do you think about?
>     * [ ]  handles empty keyfile directories gracefully (i.e. `/etc/dconf/db/local.d` with no files whatsoever means that the `/etc/dconf/db/local` part of the rule should pass).

matejak:
I have added a jinja-based parametrization that allows Fedora and non-Fedora products to be treated differently. As the [official gdm banner guidance](https://help.gnome.org/admin/system-admin-guide/stable/login-banner.html.en) mentions the `gdm.d` directory, I would suggest sticking to it for our purposes.

redhatrises:
> I have added a jinja-based parametrization that allows Fedora and non-Fedora products to be treated differently. As the [official gdm banner guidance](https://help.gnome.org/admin/system-admin-guide/stable/login-banner.html.en) mentions the `gdm.d` directory, I would suggest sticking to it for our purposes.

Does `RHEL8` get rid of the `gdm.d` directory as well? Or is that starting with Fedora 30?

matejak:
@redhatrises We can introduce a rule for the dconf gdm profile (additionally to the user profile), which will basically mandate this. As the dconf documentation mentions gdm.d, it seems the right path to go.

ggbecker:
> > I have added a jinja-based parametrization that allows Fedora and non-Fedora products to be treated differently. As the [official gdm banner guidance](https://help.gnome.org/admin/system-admin-guide/stable/login-banner.html.en) mentions the `gdm.d` directory, I would suggest sticking to it for our purposes.
> 
> Does `RHEL8` get rid of the `gdm.d` directory as well? Or is that starting with Fedora 30?

I've tested in RHEL8 (RHEL-8.0.0-20190404.2), it does not contain `gdm.d` directory by default and neither contains `gdm` in the `/usr/share/dconf/profile/gdm` profile:

$rpm -qa dconf
dconf-0.28.0-3.el8.x86_64

$rpm -qa gdm
gdm-3.28.3-20.el8.x86_64

$ls /etc/dconf/db/
distro  distro.d  ibus  ibus.d  local  local.d  site  site.d

$cat /usr/share/dconf/profile/gdm 
user-db:user
system-db:local
system-db:site
system-db:distro
file-db:/usr/share/gdm/greeter-dconf-defaults

ggbecker:
It seems that not having these directories in place is intentional. From this [comment](https://bugzilla.gnome.org/show_bug.cgi?id=722241#c1): 
```
If users want to make additional customisations to the gdm login session
then they should create their own dconf database in /etc/dconf and then
replace the 'gdm' profile with one that references their new db (by way
of /etc/dconf/profile/gdm, which is searched first).
```
This is probably why these directories don't exist anymore in fedora/RHEL8. In RHEL7.7 it still exists as an outcome from this: https://access.redhat.com/solutions/3599341

Nevertheless, I think it is not a problem if they are not there. It is just a matter of ensuring they will be created by the remediation and that the `gdm` profile references this newly created dconf database.

matejak:
FYI, there is a gdm.d-related BZ for RHEL 7: https://bugzilla.redhat.com/show_bug.cgi?id=1664284

matejak:
And BTW, the difference between RHEL 7 and RHEL 8 + Fedora is a bug, let's hope it will get fixed soon!
https://bugzilla.redhat.com/show_bug.cgi?id=1715905
https://bugzilla.redhat.com/show_bug.cgi?id=1715908

jan-cerny:
Thank you very much!

trevor-vaughan:
Sorry that I'm late to the party on this one but I think that there are some issues that need to be addressed.

1. The configuration files in `/etc/dconf/db/*.d` are completely arbitrary and governed by the inclusion in one of the files in `/etc/dconf/profile`. To properly evaluate the loaded profiles, you'll need to process `/etc/dconf/profile/*` and/or use the command line utilities *as different users under restriction* since overrides can have different behavior based on inclusion overrides.
2. I don't see anything checking for the appropriate lock files in the `locks` directory. I may be missing something though since there were a couple of commands being run that I don't recognize from any of my EL7 systems.
3. All validations must be run as a regular user since dconf works on an inheritance model meaning that I think that users can completely fool the SCAP checks as written by simply not including the `local` profile in my included profile list but keeping the contents in place. Alternatively, I can put a different rule set above the `local` profile which will provide different default options and ignore the `local` settings as expected by these checks.

matejak:
1. Could you please provide a concrete example of how one could manage to go around our remediations?
2. The OVAL checks for all files under the `/etc/dconf/db/local.d` (or `gdm.d`) directories address the whole file tree above that directory, so contents of the `locks` directory are processed as well.
3. I don't think that the regular user is much involved here. If a setting is marked as locked, the direction of inheritance is flipped, so user's settings don't take precedence. True, a bad actor could mess with the profile file, but such user would have to have admin rights. Again, could you provide us with some examples?

PR #4383, opened 2019-06-02
Update CSRF_COOKIE_SECURE checks

shawndwells:
- Update CSRF_COOKIE_SECURE checks to use standard rule.yml template
- Assign CCEs
- Create NIST mappings. Closest I could find is https://nvd.nist.gov/800-53/Rev4/control/SC-5 about "denial of service" attacks
- Create Ansible snippets


applications/openstack/horizon/horizon_csrf_cookie_secure/rule.yml
```diff
@@ -2,18 +2,40 @@ documentation_complete: true
...
+    with an HTTPS connection.
+
+rationale: |-
+    CSRF (Cross-site request forgery) is an attack which forces an end user
+    to execute unauthorized commands on a web application in which he/she is
+    currently authenticated. A successful CSRF exploit can compromise end
+    user data and operations in case of normal user. If the targeted end
+    user has admin privileges, this can compromise the entire web application.
+
+severity: high
```
shawndwells: For conversation during review, marked this as ``high`` because a cross site scripting attack would immediately result in loss of C/I/A. Marking something as ``high`` severity does have ripple effects and would appreciate dialog on if this should be something else (e.g. medium).

DISA definitions:
- CAT I Any vulnerability, the exploitation of which will directly and
immediately result in loss of Confidentiality, Availability, or Integrity.
- CAT II Any vulnerability, the exploitation of which has a potential to result in
loss of Confidentiality, Availability, or Integrity.
- CAT III Any vulnerability, the existence of which degrades measures to
protect against loss of Confidentiality, Availability, or Integrity.

shawndwells:
@keithkjackson Here's an updated RHOSP rule to show how things should be templated out. Must still pass PR review, but gives an idea you could start from.

shawndwells:
bump for merge review

redhatrises:
@shawndwells please use the `command` module and not the `shell` module for your ansible remediations. Shell should really only be used if you need variables like $HOME and operations like "<", ">", "|", ";" and "&"

matejak:
Replacing the module [is a good idea](https://blog.confirm.ch/ansible-modules-shell-vs-command/). Aside from that, there is a CCE conflict. I can't push those fixes myself as the target repo probably doesn't exist any more, could you please look into this?

redhatrises:
Closing in favor of #4761

PR #4385, opened 2019-06-05
Improve profile statistics for jenkins job

ggbecker:
#### Description:

- Improve which data is presented in jenkins statistics job: [Statistics](https://jenkins.complianceascode.io/view/SCAP%20Security%20Guide/job/scap-security-guide-stats/Statistics/)
- For each statistics, if they are higher than zero, list the rules in a separate file so it is easy to identify them.

#### Rationale:

- Show more useful data.

### DIFF
--- a/build-scripts/profile_tool.py
+++ b/build-scripts/profile_tool.py
@@ -69,6 +69,8 @@ def parse_args():
     parser_stats.add_argument("--format", default="plain",
                         choices=["plain", "json", "csv", "html"],
                         help="Which format to use for output.")
+    parser_stats.add_argument("--output",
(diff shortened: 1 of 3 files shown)


jan-cerny:
@ggbecker Cool!

Please fix the PEP8 problems.

ggbecker:
@jan-cerny I have submitted a new commit that fix some of the pep issues. The remaining issues I would prefer to not fix as it would deviate from the others parts of the code and regarding `line too long` I would avoid changing as it would make it less legible and they exceed only a few characters from the limit. What do you think?

jan-cerny:
@ggbecker That's fine.

PR #4391, opened 2019-06-12
Match the comment to the actual "less than or equal" operation.

adelton:
#### Description:

- Match the comment to the actual "less than or equal" operation.

#### Rationale:

- Consistency.


matejak:
add to whitelist

matejak:
LGTM, thank you for the PR.

PR #4392, opened 2019-06-12
https://iase.disa.mil/ reports "This website is no longer being updated."

adelton:
And "Please click below to visit the DoD Cyber Exchange public" with link
to https://public.cyber.mil/.

#### Description:

- Clicking on links like "CCI-000368" in output HTML report generates links like http://iase.disa.mil/stigs/app-security/app-servers/Pages/general.aspx and shows message  "This website is no longer being updated." and redirect to https://public.cyber.mil/ without preserving the exact target under it.
- Equivalent of http://iase.disa.mil/stigs/os/unix-linux/Pages/red-hat.aspx does not seem to exist in the new search UI, so I changed the text in docs/man_page_template.jinja to point to the unix-linux.
- I don't know what the target eqivalent of http://iase.disa.mil/stigs/app-security/app-servers/Pages/general.aspx should be. It is used as value of disa-appsrguri in shared/transforms/shared_shorthand2xccdf.xslt to show links for "SRG-APP-"? Where should it point?

#### Rationale:

- The links in HTML outputs and overall should point to live targets.


adelton:
Hello @pep8speaks, what am I supposed to do with the long lines when the URL are long?

jan-cerny:
Hi @adelton , I'm afraid that @pep8speaks won't reply, so I will try to reply instead of them. In this specific case it's OK to keep the long lines, don't try to shorten the URLs.

matejak:
test this please

yuumasato:
`disa-appsrguri` should probably point to `https://public.cyber.mil/stigs/downloads/?_dl_facet_stigs=application-servers`, as there you can find document `Application Server SRG - Ver 2, Rel 5`.

adelton:
> `disa-appsrguri` should probably point to `https://public.cyber.mil/stigs/downloads/?_dl_facet_stigs=application-servers`, as there you can find document `Application Server SRG - Ver 2, Rel 5`.

Thanks. Updated, rebased on master -> 5a7c92a1c81ecf28746575180389e12945663154.

adelton:
However, it seems that https://public.cyber.mil/stigs/downloads/?_dl_facet_stigs=app-security%2Capplication-servers and https://public.cyber.mil/stigs/downloads/?_dl_facet_stigs=application-servers give the same content. And we've used the https://public.cyber.mil/stigs/downloads/?_dl_facet_stigs=app-security%2Capplication-servers for `disa-stigs-apps-appserver-uri`. Are `disa-stigs-apps-appserver-uri` and `disa-appsrguri` supposed to be the same?

yuumasato:
It seems that intention of 
- `disa-stigs-apps-appserver-uri` was to list all application server STIGS, and
- `disa-appsrguri` was to list general SRGS

See, https://github.com/ComplianceAsCode/content/pull/2394

In the old webpage, they were listed separately, but now they are listed together, and there doesn't seem to be a way to list them separately.

IMO it should be fine to have them the same.

PR #4393, opened 2019-06-12
Call dconf update in all dconf-related shared remediation functions.

matejak:
#### Description:

Remediations have to make sure that they not only generate correct keyfiles, but that they also apply them to the system database.


ggbecker:
LGTM.

PR #4394, opened 2019-06-13
Remove ed25519 key generation from Dockerfiles as it is not supported when FIPS is enabled.

ggbecker:
#### Description:

- Latest RHEL7 container image (FIPS enabled) doesn't build if it tries to generate this ssh ed25519 key. Remove also from other distributions (Fedora, CentOS)

https://access.redhat.com/solutions/3643252

Ansible playbooks from [Security Labs](https://github.com/RedHatDemos/SecurityDemos/blob/master/2019Labs/CustomSecurityContent/setup/) throw error when trying to build the RHEL container image.


matejak:
Great fix, thanks!

PR #4396, opened 2019-06-14
Install just things needed for the sssd service to run.

adelton:
#### Description:

- Install just things needed for the sssd service to run.

#### Rationale:

- No need to pull in the whole sssd package and its dependencies.
- Namely, sssd-common (and its dependencies) are just fine.

### DIFF
--- a/tests/data/group_services/group_sssd/rule_sssd_memcache_timeout/comment.fail.sh
+++ b/tests/data/group_services/group_sssd/rule_sssd_memcache_timeout/comment.fail.sh
@@ -5,7 +5,7 @@
 SSSD_CONF="/etc/sssd/sssd.conf"
 TIMEOUT="180"
 
-yum -y install sssd
+yum -y install /usr/lib/systemd/system/sssd.service
 systemctl enable sssd
 mkdir -p /etc/sssd
 touch $SSSD_CONF
--- a/tests/data/group_services/group_sssd/rule_sssd_memcache_timeout/correct_value.pass.sh
+++ b/tests/data/group_services/group_sssd/rule_sssd_memcache_timeout/correct_value.pass.sh
@@ -7,7 +7,7 @@ SSSD_CONF="/etc/sssd/sssd.conf"
 # this should pass for every product which contains ospp profile
 TIMEOUT="180"
 
-yum -y install sssd
+yum -y install /usr/lib/systemd/system/sssd.service
 systemctl enable sssd
 mkdir -p /etc/sssd
 touch $SSSD_CONF
--- a/tests/data/group_services/group_sssd/rule_sssd_memcache_timeout/wrong_section.fail.sh
+++ b/tests/data/group_services/group_sssd/rule_sssd_memcache_timeout/wrong_section.fail.sh
@@ -5,7 +5,7 @@
 SSSD_CONF="/etc/sssd/sssd.conf"
 TIMEOUT="180"
 
-yum -y install sssd
+yum -y install /usr/lib/systemd/system/sssd.service
 systemctl enable sssd
 mkdir -p /etc/sssd
 touch $SSSD_CONF
--- a/tests/data/group_services/group_sssd/rule_sssd_memcache_timeout/wrong_value.fail.sh
+++ b/tests/data/group_services/group_sssd/rule_sssd_memcache_timeout/wrong_value.fail.sh
@@ -8,7 +8,7 @@ SSSD_CONF="/etc/sssd/sssd.conf"
 # Let's put there a higher value to fail
 TIMEOUT="99999"
 
-yum -y install sssd
+yum -y install /usr/lib/systemd/system/sssd.service
 systemctl enable sssd
 mkdir -p /etc/sssd
 touch $SSSD_CONF
--- a/tests/data/group_services/group_sssd/rule_sssd_offline_cred_expiration/common.sh
+++ b/tests/data/group_services/group_sssd/rule_sssd_offline_cred_expiration/common.sh
@@ -2,7 +2,7 @@
 
 SSSD_CONF="/etc/sssd/sssd.conf"
 
-yum -y install sssd
+yum -y install /usr/lib/systemd/system/sssd.service
 systemctl enable sssd
 mkdir -p /etc/sssd
 touch $SSSD_CONF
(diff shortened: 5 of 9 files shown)


jan-cerny:
@adelton I have run test scenarios on rule `sssd_memcache_timeout` on a Fedora 30 virtual machine target, and I have got multiple errors. 

For example, the test scenario `comment.fail.sh` has failed because the rule  `sssd_memcache_timeout`  has been evaluated  as `pass` instead of expected result `fail` (the expected result is always inferred form the test scenario file name).

I have run the test suite in debug mode and I have seen that  `sssd` service is enabled and `memcache_timeout` is commented out in `/etc/sssd.conf`. Therefore, it seems that the scenario has set up the environment successfully. 

So, I had to take a look at the corresponding OVAL in `/linux_os/guide/services/sssd/sssd_memcache_timeout/oval/shared.xml`. From the OVAL you can read that the rule will pass if: `sssd` service is disabled OR `sssd.conf` doesn't exist OR `memcache_timeout` is configure properly. At the first sight, it seems that all three conditions are not fulfilled, so the rule should be evaluated as fail.

However, if we take a closer look at the definition that check for "disabled service"  which is implemented in `/shared/templates/template_OVAL_service_disabled` we will see the definition can pass if the respective package *isn't* installed. In our case it passes because `sssd` package isn't present, which is because we haven't installed `sssd` package, but we have installed `sssd-common` package because the real package that provides `/usr/lib/systemd/system/sssd.service` on Fedora 30 is `sssd-common`, not `sssd`. 

I think that we should check presence of the `sssd-common` package instead. It's possible to change  the package used in this definition easily by modifying `/fedora/templates/csv/services_disabled.csv` to `sssd,sssd-common,`  and adding `sssd-common` to `/fedora/templates/csv/packages_removed.csv` (it's a templated rule). 

I have tried this change and it fixes the failing test case for me on Fedora. 

However, it needs to be checked if this change should be done also on other operating systems, because the sssd-related rules are also part of RHEL7, RHEL8, RHV4 and OL8 profiles.

Have you experienced these errors as well? Could you please check the problem with `sssd` vs `sssd-common`?

yuumasato:
One important change to take into account is that in Fedora 30 and RHEL8 `sssd.service` is installed and running by default.
The test scenarios were written fro RHEL7, and there the service is not installed by default.

adelton:
@jan-cerny, yes, this commit is meant to bring to surface the fact that the current `sssd_memcache_timeout` does not work at all (== gives false positive results) on minimal installations where the `sssd` package is not present yes `sssd.service` is happily loaded and configured with any `memcache_timeout` value.

I believe checking for the package presence is not needed at all and should be removed.

adelton:
Furthermore, I believe @dahaic has hit some issues with tests for this rule on RHEL 7 even without this commit applied.

jan-cerny:
@dahaic Could you share your issues with us, please? I run the tests for the rule `sssd_memcache_timeout` on a RHEL 7.6 virtual machine target from current upstream master as of HEAD 80988ea944d8af120731f5935d548c68cbc7c110 and I don't experience any error.

jan-cerny:
@adelton I think the original intention was that the rule will pass if sssd isn't running.

adelton:
> the rule will pass if sssd isn't running

Certainly. But is looking at installed packages the way to determine if the service is running?

If you install minimal RHEL 8 (for example in beaker), you will *not* have `sssd` package installed. You will have `sssd-common` installed. And then you set the value of `memcache_timeout` to 99999, and test that rule, and it will pass. In spite of the fact that `sssd.service` is up and running and value 99999 > 86400 is used.

Because the rule will see `sssd` package not present and not bother looking at the value at all.

In other words, this pull request is supposed to bring to surface bug in the OVAL. You should yell "this breaks the tests". And you'd be right, the OVAL needs to be fixed as well. But as mentioned, the tests for `sssd_memcache_timeout` have been failing for @dahaic even without this change. Which is the second thing that this pull request is meant to point to because you will surely run the test on vanilla content (without which change) first. ;-)

jan-cerny:
@adelton Does changing the CSVs fix it?

dahaic:
I have tried RHEL8 VM, and it's passing as well, on master 7f9ac3da86f9c391ce78ce63bda60ea9f2a5fab7

```
[dahaic@psyduck tests]$ ./test_suite.py rule --libvirt qemu:///session ssg-test-suite-rhel8 --datastream ../build/ssg-rhel8-ds.xml memcache
INFO - Logging into /home/dahaic/RH/git/upstream/dahaic/scap-security-guide/tests/logs/rule-custom-2019-06-20-1306/test_suite.log
INFO - xccdf_org.ssgproject.content_rule_sssd_memcache_timeout
INFO - Script wrong_section.fail.sh using profile xccdf_org.ssgproject.content_profile_ospp OK
```

jan-cerny:
@dahaic Thank you for confirming that it passes not only for me.

adelton:
> I have tried RHEL8 VM, and it's passing as well, on master [7f9ac3d](https://github.com/ComplianceAsCode/content/commit/7f9ac3da86f9c391ce78ce63bda60ea9f2a5fab7)

Great. Can you now try the test with this pull request's changes applied?

jan-cerny:
@dahaic 

> This should be easily fixable by altering templates/csv/services_disabled.csv

See comment 3 for a more detailed description of needed changes.

jan-cerny:
However, I suggest considering if we can fix it using "applicability check". Using it has a benefit that the rule is less complex because it only checks the config file option and doesn't check any packages or services. One option is the rule would be applicable only if the sssd service is configured to be enabled. Another option is that the rule would be applicable if sssd-common is installed.

dahaic:
I can confirm that with changes described in comment 3, both test runs are passing. I'm slightly opposed to asking for change to applicability language here though - that can be done as the follow-up PR for the whole sssd group without creeping this PR too much.

jan-cerny:
@dahaic OK, please open the new PR

adelton:
Uff uff. I thought that you'd complain loudly that this PR breaks the tests (unexpected passes or something) and we will iterate on fixing the OVALs here ...

adelton:
BTW, I'm proposing `platform: sssd` applicability check as part of https://github.com/ComplianceAsCode/content/pull/4430.

PR #4398, opened 2019-06-14
Add ansible remediation for require_singleuser_auth

justin-stephenson:
#### Description:

- Add ansible remediation for require_singleuser_auth

### DIFF
--- a/linux_os/guide/system/accounts/accounts-physical/require_singleuser_auth/ansible/shared.yml
+++ b/linux_os/guide/system/accounts/accounts-physical/require_singleuser_auth/ansible/shared.yml
@@ -1,0 +1,25 @@
+# platform = multi_platform_all
+# reboot = false
+# strategy = restrict
+# complexity = low
+# disruption = low
+
+{{% if init_system == "systemd" -%}}
+- name: require single user mode password
+  lineinfile:
+      create: yes
+      dest: /usr/lib/systemd/system/rescue.service
+      regexp: "^#?ExecStart="
+      {{% if product in ["fedora", "rhel8"] -%}}
+      line: "ExecStart=-/usr/lib/systemd/systemd-sulogin-shell rescue"
+      {{%- else -%}}
+      line: "ExecStart=-/sbin/sulogin rescue"
+      {{%- endif %}}
+{{%- else -%}}
+- name: require single user mode password
+  lineinfile:
+      create: yes
+      dest: /etc/sysconfig/init
+      regexp: "^#?SINGLE="
+      line: "SINGLE=/sbin/sulogin"
+{{%- endif %}}


jan-cerny:
@openscap-ci add to whitelist

linux_os/guide/system/accounts/accounts-physical/require_singleuser_auth/ansible/shared.yml
```diff
@@ -0,0 +1,25 @@
...
+# disruption = low
+
+{{% if init_system == "systemd" -%}}
+- name: require single user mode password
+  lineinfile:
+      create: yes
+      dest: /usr/lib/systemd/system/rescue.service
+      regexp: "^#?ExecStart="
+      {{% if product == "fedora" -%}}
+      line: "ExecStart=-/usr/lib/systemd/systemd-sulogin-shell"
```
jan-cerny: The corresponding Bash remediation (`in linux_os/guide/system/accounts/accounts-physical/require_singleuser_auth/bash/shared.sh`) adds also `rescue` on this line.

linux_os/guide/system/accounts/accounts-physical/require_singleuser_auth/ansible/shared.yml
```diff
@@ -0,0 +1,25 @@
...
+# complexity = low
+# disruption = low
+
+{{% if init_system == "systemd" -%}}
+- name: require single user mode password
+  lineinfile:
+      create: yes
+      dest: /usr/lib/systemd/system/rescue.service
+      regexp: "^#?ExecStart="
+      {{% if product == "fedora" -%}}
```
jan-cerny: On my RHEL 8.0.0 machine, `/usr/lib/systemd/system/rescue.service` contains `ExecStart=-/usr/lib/systemd/systemd-sulogin-shell rescue`. So I think it isn't applicable only to Fedora but the condition should be extended also to RHEL8.

jan-cerny: I have realized that if we change this condition we need to change the condition also in Bash, OVAL and rule.yml, because everything that is related to the rule should consistently check or modify the same values.

justin-stephenson:
Thank you for the review @jan-cerny, changes made as requested.

linux_os/guide/system/accounts/accounts-physical/require_singleuser_auth/ansible/shared.yml
yuumasato: I have filed https://github.com/ComplianceAsCode/content/pull/4407 to address this issue.

yuumasato:
I have tested the remediation with RHEL7 and RHEL8 and they work fine.

PR #4400, opened 2019-06-17
Options memcache_timeout and offline_credentials_expiration are performance-related, not security-related.

adelton:
#### Description:

- Stop setting SSSD options `memcache_timeout` and `offline_credentials_expiration` in OSPP profiles.

#### Rationale:

- They are performance related, not security related.

### DIFF
--- a/fedora/profiles/ospp.profile
+++ b/fedora/profiles/ospp.profile
@@ -68,9 +68,6 @@ selections:
     - gnome_gdm_disable_automatic_login
     - gnome_gdm_disable_guest_login
     - sssd_run_as_sssd_user
-    - sssd_offline_cred_expiration
-    - sssd_memcache_timeout
-    - var_sssd_memcache_timeout=1_day
     - disable_host_auth
     - sshd_disable_gssapi_auth
     - sshd_disable_kerb_auth
--- a/linux_os/guide/services/sssd/sssd_memcache_timeout/rule.yml
+++ b/linux_os/guide/services/sssd/sssd_memcache_timeout/rule.yml
@@ -30,7 +30,6 @@ references:
     disa: "2007"
     nist: 'IA-5(10),IA-5(13)'
     nist-csf: PR.AC-1,PR.AC-6,PR.AC-7
-    ospp: FIA_AFL.1
     srg: SRG-OS-000383-GPOS-00166
     vmmsrg: SRG-OS-000383-VMM-001570
     isa-62443-2013: 'SR 1.1,SR 1.10,SR 1.2,SR 1.3,SR 1.4,SR 1.5,SR 1.7,SR 1.8,SR 1.9,SR 2.1'
--- a/linux_os/guide/services/sssd/sssd_offline_cred_expiration/rule.yml
+++ b/linux_os/guide/services/sssd/sssd_offline_cred_expiration/rule.yml
@@ -27,7 +27,6 @@ references:
     disa: "2007"
     nist: IA-5(13)
     nist-csf: PR.AC-1,PR.AC-6,PR.AC-7
-    ospp: FIA_AFL.1
     srg: SRG-OS-000383-GPOS-00166
     vmmsrg: SRG-OS-000383-VMM-001570
     isa-62443-2013: 'SR 1.1,SR 1.10,SR 1.2,SR 1.3,SR 1.4,SR 1.5,SR 1.7,SR 1.8,SR 1.9,SR 2.1'
--- a/ol8/profiles/ospp.profile
+++ b/ol8/profiles/ospp.profile
@@ -67,9 +67,6 @@ selections:
     - gnome_gdm_disable_automatic_login
     - gnome_gdm_disable_guest_login
     - sssd_run_as_sssd_user
-    - sssd_offline_cred_expiration
-    - sssd_memcache_timeout
-    - var_sssd_memcache_timeout=1_day
     - disable_host_auth
     - sshd_disable_gssapi_auth
     - sshd_disable_kerb_auth
--- a/rhel7/profiles/ospp.profile
+++ b/rhel7/profiles/ospp.profile
@@ -390,8 +390,6 @@ selections:
     - package_pcsc-lite_installed
     - service_pcscd_enabled
     - sssd_enable_smartcards
-    - sssd_memcache_timeout
-    - sssd_offline_cred_expiration
     - sssd_ssh_known_hosts_timeout
     - encrypt_partitions
     - ensure_redhat_gpgkey_installed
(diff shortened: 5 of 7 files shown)


redhatrises:
`memcache_timeout` and `offline_credentials_expiration` are actually valid security related checks. User caches in secure environments should never persist for long. These checks map to NIST 800-53 IA-5(13), so the question is does OSPP and the configuration annex call out IA-5.

shawndwells:
> `memcache_timeout` and `offline_credentials_expiration` are actually valid security related checks. User caches in secure environments should never persist for long. These checks map to NIST 800-53 IA-5(13), so the question is does OSPP and the configuration annex call out IA-5.

Yes, great job citing sources. Specifically IA-5(f) and IA-5(g:
`````
f. Establishing minimum and maximum lifetime restrictions and reuse conditions for authenticators;

g. Changing/refreshing authenticators [Assignment: organization-defined time period by authenticator type];
`````

Which can be found at https://nvd.nist.gov/800-53/Rev4/control/IA-5

adelton:
As for caches, SSSD already has reasonable default for `memcache_timeout` of 300 s which is much more strict than the current `var_sssd_memcache_timeout=1_day`.
As for `offline_credentials_expiration`, if the auth provider is offline for a day, is the right course of action disruption of service (because with offline provider, there is not way to reauthenticate anyway)?
@stevegrubb, @jhrozek, @mzidek-rh, what is your opinion about these two options?

mzidek-rh:
Hi,
the nist IA-5(13) says:

"The information system prohibits the use of cached authenticators after [Assignment: organization-defined time period"

So, reading this, and the other citations mentioned in this thread, I think we should keep offline_credentials_expiration in the list, because someone might have a policy that people "must re-authenticate" within certain period of time (I think @jhrozek mentioned something similar on IRC yesterday). The fact that this is "offline" makes it harder to judge, but maybe denying access/disrupting service could be less problematic in some cases then missing re-authentication. Idk.

But memcache_timeout is not related because it only caches results of (some) NSS requests and does not deal with authentication. It really is just performance option to avoid contacting sssd_nss when the same information is requested again in a short a period of time. If someone made the timeout crazy big, it would still not allow anyone to authenticate for longer period (it would just give outdated info for some NSS requests).

adelton:
> because someone might have a policy that people "must re-authenticate" within certain period of time (I think @jhrozek mentioned something similar on IRC yesterday).

That is certainly valid concern. However, then that value of "certain period of time" would need to be defined as a parameter and used consistently everywhere, for example as `offline_credentials_expiration` value, not merely hardcode that `offline_credentials_expiration = 1` which the content is doing now.

By the way, are sub-integer values supported for `offline_credentials_expiration`, to allow expiration shorter than one day?

mzidek-rh:
> > because someone might have a policy that people "must re-authenticate" within certain period of time (I think @jhrozek mentioned something similar on IRC yesterday).
> 
> That is certainly valid concern. However, then that value of "certain period of time" would need to be defined as a parameter and used consistently everywhere, for example as `offline_credentials_expiration` value, not merely hardcode that `offline_credentials_expiration = 1` which the content is doing now.
> 
> By the way, are sub-integer values supported for `offline_credentials_expiration`, to allow expiration shorter than one day?

No.

jhrozek:
> On 18 Jun 2019, at 09:22, Jan Pazdziora <notifications@github.com> wrote:
> 
> As for caches, SSSD already has reasonable default for memcache_timeout of 300 s which is much more strict than the current var_sssd_memcache_timeout=1_day.
> As for offline_credentials_expiration, if the auth provider is offline for a day, is the right course of action disruption of service (because with offline provider, there is not way to reauthenticate anyway)?
> @stevegrubb, @jhrozek, @mzidek-rh, what is your opinion about these two options?
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 

I’m sorry about the delay.

I wonder if the current test for the memcache does what you intend to do? Because sssd has several layers of caches, the memory cache is the top-level fastest one which the application that performs NSS calls loads by virtue of loading the SSSD NSS module. Setting the memcache expiration to a low value[1] just expires this cache layer, but does nothing about the on-disk cache. And what I don’t understand from the NIST snippet is whether you care about the cache being just expired, but still reachable, or whether you need to ensure that there is no cache.

The former might be doable by setting all cache expiration timeouts, the latter is really not implemented in sssd[2]. If there are cache entries present but expired, they would only be used if there is no way to refresh the cache from the server.

Finally, we used to have a list of security-sensitive sssd options on the old wiki. I don’t think we migrated that list but it wouldn’t be that hard to do. OTOH that list was not written with any specific policy in mind, just using our best judgement.

[1] 1_day is a really high value. The default, as adelton said is 300 seconds.
[2]. Sort of. SSSD can enable a periodical taks that reaps stale cache entries, but since this task needs to traverse the whole cache, it is quite expensive and disabled by default. And since it’s disabled by default, it’s not the most stable piece of SSSD.

adelton:
Hello, is anything else needed to get this merged?

My reading of the explanation that the friendly SSSD folks provided is that in case of `memcache_timeout` we are actually fixing the existing unfortunate setting by removing `sssd_memcache_timeout` and `var_sssd_memcache_timeout=1_day` from the OSPP profile.

As for `offline_credentials_expiration`, it likely does not do what is intended here anyway.

adelton:
I can of course also rebase on master to allow for nicer merge commit. Or just do rebase merge.

yuumasato:
> These checks map to NIST 800-53 IA-5(13), so the question is does OSPP and the configuration annex call out IA-5.

I could not find IA-5(13) in the configuration annex, nor in these mappings by [NIAP](https://www.niap-ccevs.org/MMO/PP/pp_os_v4.1-map.pdf).

It makes sense to remove these from `ospp` profiles.

@adelton Please, rebase on top of latest master.

adelton:
Rebased -> 2ece4662491653d990fa3126a5e88158ee88d002.

yuumasato:
@adelton One more thing, can you remove the `ospp` reference in the rules?
I'm sorry I didn't ask for this earlier.

adelton:
Is 87ede7d8285a45f8d3a04ec0f33ad05bcd964f23 what you had in mind?

yuumasato:
@adelton Almost, just remove the `ospp` references, the `nist` ones can be kept.

adelton:
So what did you mean by that

> I could not find IA-5(13) in the configuration annex

comment?

yuumasato:
> I could not find IA-5(13) in the configuration annex

The [configuration annex](https://www.niap-ccevs.org/MMO/PP/-442ConfigAnnex-/) is also related to OSPP, and the mapping I pointed out also doesn't map IA-5(13) to any OSPP SFR.

But the rule is still relevant from point of view of NIST controls.

adelton:
Ah, OK. I've misunderstood the intent. Bringing the `nist` values back -> 8ba99e8d15e948bd7aed250f255490479b3f97d6.

PR #4401, opened 2019-06-17
Add Dockerfile for RHEL 8.

adelton:
#### Description:

- To test in RHEL 8 containers, we need Dockerfile for RHEL 8.

#### Rationale:

- Testing on RHEL 8 container should be as simple as on RHEL 7.


jan-cerny:
I have built a Docker image using your Dockerfile and I have used it  to run test scenarios for rule `accounts_tmout` from `ssg-rhel8-ds.xml`. Evertyhing went smoothly. I use `docker-1.13.1-96.gitb2f74b2.el7.x86_64`. Thank you!

PR #4402, opened 2019-06-17
The user running the tests does not need to be root.

adelton:
#### Description:

- It shouldn't be necessary to create the SSH key pair as root, just run it under current user (which might happen to be root).

#### Rationale:

- Avoid confusing people who do not want to mess up with root's content on their machines.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -90,29 +90,18 @@ To use Podman backend, you need to have:
 *NOTE*: With Podman, you have to run all the operations as root. Podman supports rootless containers, but the test suite internally uses a container exposing a TCP port. As of Podman version 0.12.1.2, port bindings are not yet supported by rootless containers.
 
 ##### Building podman base image
-The root user needs to log in to the container via SSH, so lets setup a key without passphrase, so it can happen without any additional interaction.
-Root user typically doesn't have a SSH key, or it has one without a passphrase.
-Use this test to find out whether it has a key:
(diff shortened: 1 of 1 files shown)


adelton:
What parts exactly do you need to run as root? Why doesn't `sudo podman` do the trick for you?

dahaic:
Unfortunately the `./test_suite.py` command. I'll try to go through the error messages tomorrow, to understand which action of podman is triggering it.

matejak:
Previous versions of podman couldn't publish ports when run by non-privileged user. Therefore, the whole test suite had to be executed as root (to run podman as root), and the corresponding container image had to be built with root SSH keys.
However, recent podman can publish a non-privileged port when executed by a non-privileged user. As our test suite code assumes that the container listens on port 22 (which is privileged), we have to

1. Execute podman so it exports it's port 22 to let's say 2200, and
2. Connect to the container on the non-default 2200 port.

dahaic:
Sorry folks, I am still struggling to setup by laptop to allow rootless podman. So I don't want to block this PR, and I am unassigning myself.

matejak:
We need a PR that adds support for custom SSH ports first. When podman is executed under normal user, the container won't get an IP address, and we will need to connect to localhost and a custom SSH port. Our test suite doesn't count with this.

adelton:
Withdrawing.

PR #4403, opened 2019-06-17
Various bash fixes.

matejak:
Fixed quoting issues, removed IFS manipulation in favor of the readarray bash builtin.

See e.g. https://unix.stackexchange.com/questions/263883/bash-pipe-find-output-into-readarray for example usage of `readarray`.

### DIFF
--- a/linux_os/guide/system/permissions/partitions/mount_option_nodev_nonroot_local_partitions/bash/rhel6.sh
+++ b/linux_os/guide/system/permissions/partitions/mount_option_nodev_nonroot_local_partitions/bash/rhel6.sh
@@ -12,17 +12,15 @@ IFS_BKP="$IFS"
 
 # Load /etc/fstab's content with LABEL= and UUID= tags expanded to real
 # device names into FSTAB_REAL_DEVICES array splitting items by newline
-IFS=$'\n'
-FSTAB_REAL_DEVICES=($(findmnt --fstab --evaluate --noheadings))
+readarray FSTAB_REAL_DEVICES < <(findmnt --fstab --evaluate --noheadings)
 
-for line in ${FSTAB_REAL_DEVICES[@]}
+for line in "${FSTAB_REAL_DEVICES[@]}"
 do
     # For each line:
     # * squeeze multiple space characters into one,
     # * split line content info four columns (target, source, fstype, and
     #   mount options) by space delimiter
-    IFS=$' '
-    read TARGET SOURCE FSTYPE MOUNT_OPTIONS <<< "$(echo $line | tr -s ' ')"
+    read TARGET SOURCE FSTYPE MOUNT_OPTIONS <<< "$(echo "$line" | tr -s ' ')"
 
     # Filter the targets according to the following criteria:
     # * don't include record for root partition,
@@ -65,10 +63,8 @@ do
             # Replace old mount options for particular /etc/fstab's row (for this target
             # and fstype) with new mount options
             sed -i "s#${TARGET_HEAD}\(.*\)${TARGET_TAIL}#${TARGET_HEAD}${MOUNT_OPTIONS}${TARGET_TAIL}#" /etc/fstab
-
+            # Reset IFS back to default
+            IFS="$IFS_BKP"
         fi
     fi
 done
-
-# Reset IFS back to default
-IFS="$IFS_BKP"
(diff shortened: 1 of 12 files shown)


linux_os/guide/system/software/updating/ensure_oracle_gpgkey_installed/bash/shared.sh
```diff
@@ -14,7 +14,7 @@ if [ "${RPM_GPG_DIR_PERMS}" -le "755" ]
 then
   # If they are safe, try to obtain fingerprints from the key file
   # (to ensure there won't be e.g. CRC error)
-  IFS=$'\n' GPG_OUT=($(gpg --with-fingerprint --with-colons "$OL_RELEASE_KEY" | grep "^fpr" | cut -d ":" -f 10))
```
matejak: IFS trick doesn't work at all.

shared/bash_remediation_functions/fix_audit_watch_rule.sh
```diff
@@ -53,40 +54,35 @@ then
...
-		if [ ! -e "$files_to_inspect" ]
+		if [ ! -e "$key_rule_file" ]
 		then
-			touch "$files_to_inspect"
-			chmod 0640 "$files_to_inspect"
+			touch "$key_rule_file"
+			chmod 0640 "$key_rule_file"
 		fi
+
+		files_to_inspect+=("$key_rule_file")
```
matejak: The previous behavior violated comment `# Append '/etc/audit/rules.d/$key.rules' into list of files ...`

linux_os/guide/system/permissions/partitions/mount_option_nodev_nonroot_local_partitions/bash/rhel6.sh
```diff
@@ -65,10 +63,8 @@ do
             # Replace old mount options for particular /etc/fstab's row (for this target
             # and fstype) with new mount options
             sed -i "s#${TARGET_HEAD}\(.*\)${TARGET_TAIL}#${TARGET_HEAD}${MOUNT_OPTIONS}${TARGET_TAIL}#" /etc/fstab
-
+            # Reset IFS back to default
+            IFS="$IFS_BKP"
```
mildas: Why reset `IFS` every iteration?

matejak: Basically this minimizes the time when `IFS` is different from what is the common expectation.

linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_ownership/bash/shared.sh
```diff
@@ -4,29 +4,23 @@
 # complexity = high
 # disruption = medium
 
-# Declare array to hold list of RPM packages we need to correct permissions for
-SETPERMS_RPM_LIST=()
+# Declare array to hold set of RPM packages we need to correct permissions for
+declare -A SETPERMS_RPM_DICT
```
matejak: Changes here are very similar to changes that were done to the tests/data/group_system/group_software/group_integrity/group_software-integrity/group_rpm_verification/rule_rpm_verify_permissions/all_permissions_ok.pass.sh test scenario earier.

linux_os/guide/system/software/updating/ensure_fedora_gpgkey_installed/bash/shared.sh
```diff
@@ -25,13 +25,8 @@ if [ "${RPM_GPG_DIR_PERMS}" -le "755" ]
 then
   # If they are safe, try to obtain fingerprints from the key file
   # (to ensure there won't be e.g. CRC error).
-  # Backup IFS value
-  IFS_BKP="$IFS"
-  IFS=$'\n'
-  GPG_OUT=($(gpg --with-fingerprint --with-colons "${REDHAT_RELEASE_KEY}" | grep '"^fpr' | cut -d ":" -f 10))
+  readarray GPG_OUT < <(gpg --with-fingerprint --with-colons "${REDHAT_RELEASE_KEY}" | grep '"^fpr' | cut -d ":" -f 10)
```
jan-cerny: I know it was there before, but I have tried this command on my machine and I don't get any line that would start with `"`.  The grep regular expression should be just `^fpr`, without `"` I think. See the similar remediations for RHEL and OL.

matejak: You are right, other remediations don't have the double quote either.

jan-cerny:
Great! Thanks!

PR #4405, opened 2019-06-17
RHEL8 CCC Profile

yuumasato:
#### Description:

- Add a Draft profile for Common Criteria Certification of RHEL8.
- Based on RHEL7, and adjusted for RHEL8
- Note that this is still a profile that is still in development.

### DIFF
--- a/rhel8/profiles/ccc.profile
+++ b/rhel8/profiles/ccc.profile
@@ -1,0 +1,270 @@
+documentation_complete: true
+
+title: 'DRAFT - Common Criteria Certification profile using OSPP'
+
+description: |-
+    This profile is a draft for a future CC certification of RHEL, based on the
+    OSPP (NIAP) requirements. It includes functionality (rules) beyond the OSPP
+    scope itself.
+
+selections:
+    ### Boot
+    - grub2_uefi_password
+
+    # systemd
+    - require_singleuser_auth
+    - grub2_disable_interactive_boot
+    - disable_ctrlaltdel_reboot
+    - disable_ctrlaltdel_burstaction
+    - service_debug-shell_disabled
+    - service_kdump_disabled
+    - service_autofs_disabled
+
+    # Partitioning
+    - mount_option_home_nodev
+    - mount_option_home_nosuid
+    - mount_option_tmp_nodev
+    - mount_option_tmp_noexec
+    - mount_option_tmp_nosuid
+    - mount_option_var_tmp_nodev
+    - mount_option_var_tmp_noexec
+    - mount_option_var_tmp_nosuid
+    - mount_option_dev_shm_nodev
+    - mount_option_dev_shm_noexec
+    - mount_option_dev_shm_nosuid
+
+    ### Kernel Config
+
+    # Boot prompt
+    - package_dracut-fips_installed
+    - grub2_enable_fips_mode
+    - grub2_audit_argument
+    - grub2_audit_backlog_limit_argument
+    - grub2_slub_debug_argument
+    - grub2_page_poison_argument
+    - grub2_vsyscall_argument
+
+    # Sysctls
+
+    # Security Settings
+    - sysctl_kernel_kptr_restrict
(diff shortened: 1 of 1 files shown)


shawndwells:
This is what the ospp profile is for. NACK.

yuumasato:
> This is what the ospp profile is for. NACK.

@shawndwells Currently the `ccc` profile is not a subset of `ospp` profile, the rule set in both profiles are significantly different. 

These are the rules present only in `ccc` profile: [ccc_sub_ospp.txt](https://github.com/ComplianceAsCode/content/files/3302492/ccc_sub_ospp.txt)
These are the rules present only in `ospp` profile: [ospp_sub_ccc.txt](https://github.com/ComplianceAsCode/content/files/3302493/ospp_sub_ccc.txt)

From what I understand this is because, although, the policies come from the same set of requirements (policy), they slightly differ on selections and extended packages.

Another approach would be to have a `core` profile with rules common to both profile, but even then as `ccc` is not a perfect subset of `ospp`, there would still exist `ccc` and `ospp` profiles that extend `core`.

shawndwells:
> > This is what the ospp profile is for. NACK.
> 
> @shawndwells Currently the `ccc` profile is not a subset of `ospp` profile, the rule set in both profiles are significantly different.
> 
> These are the rules present only in `ccc` profile: [ccc_sub_ospp.txt](https://github.com/ComplianceAsCode/content/files/3302492/ccc_sub_ospp.txt)
> These are the rules present only in `ospp` profile: [ospp_sub_ccc.txt](https://github.com/ComplianceAsCode/content/files/3302493/ospp_sub_ccc.txt)
> 
> From what I understand this is because, although, the policies come from the same set of requirements (policy), they slightly differ on selections and extended packages.
> 
> Another approach would be to have a `core` profile with rules common to both profile, but even then as `ccc` is not a perfect subset of `ospp`, there would still exist `ccc` and `ospp` profiles that extend `core`.

Sounds like the OSPP profile should be patched with the missing rules.

OSPP should be the OSPP 4.2 requirements, plus the Config Annex requirements.

Other profiles (e.g. C2S, HIPAA....) will extend OSPP as needed.

yuumasato:
> Sounds like the OSPP profile should be patched with the missing rules.

From the `ospp` Profile point of view, this may make sense. But I'm not sure if it makes sense from `ccc` Profile point of view.

rhel8/profiles/ccc.profile
```diff
@@ -0,0 +1,261 @@
...
+    - accounts_password_pam_lcredit
+    - var_password_pam_difok=4
+    - accounts_password_pam_difok
+    - var_password_pam_maxrepeat=3
+    - accounts_password_pam_maxrepeat
+    - var_password_pam_maxclassrepeat=4
+    - accounts_password_pam_maxclassrepeat
+
+    # umask
+    - accounts_umask_etc_profile
```
jan-cerny: Please set value variable `var_accounts_user_umask` because it is used in rule `accounts_umask_etc_profile`.

yuumasato: The default value is "027" and is what the profile expects. But there should be no harm in selecting it explicitly.

jan-cerny: Great!

My opinion is to set them always explicitly, because:

* the default might change which would change the profiles that rely on default, which might happen without noticing
* it is harder to find the actual value if the variable is not set in profile yaml, because you have to first realize that the rule is parametrized, then find the name of the variable, then find then variable yaml file, then open it, ...

I would even say that in future we should have a test or a build-time check that enforces explicit settings.

jan-cerny:
@yuumasato Will this be closed?

yuumasato:
@jan-cerny This PR already contains most of the necessary rules. I advocate to rename this profile as the `ospp`. 

I can get to this if it seams reasonable, and try to keep the nice requirement comments.

shawndwells:
I'm working on the OSPP profile today. I'll merge this now, and integrate into the OSPP profile and remove once done.

PR #4406, opened 2019-06-17
Fix example in csv_map

cipherboy:
#### Description:

 - Just a small comment fix to reflect reality. `except-for` instead of `only-for`. 


dahaic:
Thanks!

PR #4408, opened 2019-06-18
Profile subtraction: Sort selection of rules after subtraction

yuumasato:
#### Description:

- Subtraction of profiles is used by `profile_tool.py` with `sub` command.

#### Rationale:

- If rule selection is sorted, it is easier to compare profiles.


ggbecker:
LGTM

PR #4409, opened 2019-06-18
profile_tool.py: Handle exception when profile is not found

yuumasato:
#### Description:

- Catch exception `TemplateNotFound` thrown when trying to open non-existent profile.

#### Rationale:

- Let's not traceback, but print error message and exit.

### DIFF
--- a/build-scripts/profile_tool.py
+++ b/build-scripts/profile_tool.py
@@ -4,6 +4,7 @@ from __future__ import print_function
 
 import json
 import argparse
+import jinja2
 import os
 import os.path
 import sys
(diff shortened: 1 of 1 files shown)


jan-cerny:
Works great!
```
[jcerny@thinkpad scap-security-guide{profile_tool_handle_profile_not_found}]$  ./build-scripts/profile_tool.py sub --profile1 rhel7/profiles/ospp.profile  --profile2 rhel7/profiles/abcd
Error: Profile /home/jcerny/scap-security-guide/rhel7/profiles/abcd could not be found.
```

ggbecker:
Thanks for the patch.

PR #4410, opened 2019-06-18
Fix installation of dracut fips aesni

yuumasato:
#### Description:

- Fix condition when to install dracut-fips-aesni
- Update grub2_enable_fips_mode bash remediation


matejak:
Looks great (at least according to https://docs.ansible.com/ansible/latest/user_guide/playbooks_error_handling.html#controlling-what-defines-failure )
Thank you for the PR!

PR #4411, opened 2019-06-18
Add ansible remediation for auditd_audispd_syslog_plugin_activated

justin-stephenson:
#### Description:

- Add ansible remediation for `auditd_audispd_syslog_plugin_activated`

### DIFF
--- a/linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_audispd_syslog_plugin_activated/ansible/shared.yml
+++ b/linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_audispd_syslog_plugin_activated/ansible/shared.yml
@@ -1,0 +1,15 @@
+# platform = multi_platform_all
+# reboot = false
+# strategy = configure
+# complexity = low
+# disruption = low
+
+- name: enable syslog plugin
+  lineinfile:
+    {{% if product in ["rhel8", "fedora", "ol8"] -%}}
+    dest: /etc/audit/plugins.d/syslog.conf
+    {{%- else -%}}
+    dest: /etc/audisp/plugins.d/syslog.conf
+    {{%- endif %}}
+    regexp: "^active"
+    line: active = yes


linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_audispd_syslog_plugin_activated/ansible/shared.yml
```diff
@@ -0,0 +1,20 @@
+# platform = multi_platform_all
+# reboot = false
+# strategy = configure
+# complexity = low
+# disruption = medium
+
+- name: install audispd-plugins
+  package:
+    name: audispd-plugins
```
jan-cerny: Nice catch! However, the corresponding rule.yml doesn't mention installing this package, the Bash script also doesn't install the package, and the OVAL definition doesn't check for presence of this package. I think that we should be consistent and either remove this task from this playbook or add it to Bash, rule and OVAL. If we don't have it, the question is how we ensure the package is present and the rule is working. I have checked the sources and we don't have any rule that ensures the presence of the package, but when the package isn't installed the syslog plugin obviously can't be activated. @all Any opinions?

cipherboy: This might be slightly unrelated. My 2c. are that installing this package should be a separate rule. That way if multiple rules require a package, we don't have to duplicate package installation in both. This could be applied more generally to other rules with remediation portions. 

OTOH, I don't recall us having a generic rule dependency step... So these would have to be managed manually currently, yuck.

jan-cerny: It seems to me that it's a generic problem of our project which affects multiple rules, not only auditd rules. We will not solve it in this pull request. For now, I suggest making the Ansible task consistent with the rule.yml, Bash and OVAL, which don't install the package and don't test for its presence. @justin-stephenson Could you please remove the "package" part from your patch?

justin-stephenson: @jan-cerny PR updated, thanks.

jan-cerny:
@justin-stephenson thanks!

PR #4413, opened 2019-06-18
Move jinja macro loading out of ssg.yaml

cipherboy:
#### Description:

In order to facilitate use of Jinja macros in non-YAML code (such as
OVAL and Ansible content), we should make the loading of macros
independent of parsing YAML. This moves the processing code from
`ssg.yaml` to `ssg.jinja` and adds a `process_files_with_macros(...)` call
to `ssg.jinja`.

#### Rationale:

This'll facilitates the use of Jinja macros in other languages like OVAL checks and Ansible remediations. That will be added in a separate PR. 

### DIFF
--- a/ssg/jinja.py
+++ b/ssg/jinja.py
@@ -4,6 +4,8 @@ from __future__ import print_function
 import os.path
 import jinja2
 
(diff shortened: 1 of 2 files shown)


jan-cerny:
@cipherboy thanks!

PR #4414, opened 2019-06-18
Add ansible remediation for accounts_umask_etc_profile

justin-stephenson:
#### Description:

- Add ansible remediation for `accounts_umask_etc_profile`

### DIFF
--- a/linux_os/guide/system/accounts/accounts-session/user_umask/accounts_umask_etc_profile/ansible/shared.yml
+++ b/linux_os/guide/system/accounts/accounts-session/user_umask/accounts_umask_etc_profile/ansible/shared.yml
@@ -1,0 +1,12 @@
+# platform = multi_platform_all
+# reboot = false
+# strategy = restrict
+# complexity = low
+# disruption = low
+- (xccdf-var var_accounts_user_umask)
+
+- name: Set user umask in /etc/profile
+  replace:
+      path: /etc/profile
+      regexp: "umask.*"
+      replace: "umask {{ var_accounts_user_umask }}"
--- a/linux_os/guide/system/accounts/accounts-session/user_umask/accounts_umask_etc_profile/oval/shared.xml
+++ b/linux_os/guide/system/accounts/accounts-session/user_umask/accounts_umask_etc_profile/oval/shared.xml
@@ -18,7 +18,7 @@
   <ind:textfilecontent54_object id="obj_umask_from_etc_profile"
   comment="Umask value from /etc/profile" version="1">
     <ind:filepath>/etc/profile</ind:filepath>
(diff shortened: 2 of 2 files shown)


jan-cerny:
Hi @justin-stephenson , thanks for your great patch. However, I have find out multiple problems with the rule `accounts_umask_etc_profile`:
1. umask should be set to 027 for accounts with uid >= 1000, but our rule doesn't check for uid
2. the corresponding OVAL  allows UMASK written in capitals, which I think won't work because bash is case sensitive, so we need to make the regex to match only `umask`.
3. /etc/profile can source  other scripts which can call again umask with a different value
I think if we will fix the rule we probably will need to modify the Ansible Playbook as well.

redhatrises:
> Hi @justin-stephenson , thanks for your great patch. However, I have find out multiple problems with the rule `accounts_umask_etc_profile`:
> 
>     1. umask should be set to 027 for accounts with uid >= 1000, but our rule doesn't check for uid

@jan-cerny umask should be set for all accounts. Nothing in our rule says `uid >= 1000`

justin-stephenson:
Thank you @jan-cerny for the review.

> Hi @justin-stephenson , thanks for your great patch. However, I have find out multiple problems with the rule `accounts_umask_etc_profile`:
> 
>     1. umask should be set to 027 for accounts with uid >= 1000, but our rule doesn't check for uid
> 
>     2. the corresponding OVAL  allows UMASK written in capitals, which I think won't work because bash is case sensitive, so we need to make the regex to match only `umask`.

Is this just a matter of changing `UMASK` to `umask` in `oval/shared.yml` for this rule?
> 
>     3. /etc/profile can source  other scripts which can call again umask with a different value
> 
To address this does it makes sense to create a drop directory file in `/etc/profile.d/` instead of replacing `/etc/profile` umask lines? The filename can use a high-number prefix to override other included files.

redhatrises:
> Thank you @jan-cerny for the review.
> 
> > Hi @justin-stephenson , thanks for your great patch. However, I have find out multiple problems with the rule `accounts_umask_etc_profile`:
> > ```
> > 1. umask should be set to 027 for accounts with uid >= 1000, but our rule doesn't check for uid
> > 
> > 2. the corresponding OVAL  allows UMASK written in capitals, which I think won't work because bash is case sensitive, so we need to make the regex to match only `umask`.
> > ```
> 
> Is this just a matter of changing `UMASK` to `umask` in `oval/shared.yml` for this rule?

That should be fine.

> 
> > ```
> > 3. /etc/profile can source  other scripts which can call again umask with a different value
> > ```
> 
> To address this does it makes sense to create a drop directory file in `/etc/profile.d/` instead of replacing `/etc/profile` umask lines? The filename can use a high-number prefix to override other included files.

It should address both `/etc/profile` and `/etc/profile.d/`.

PR #4415, opened 2019-06-18
Macros everywhere

cipherboy:
#### Description

Depends on #4413; will be rebased once that is merged. 
    
Templates, remediations, and OVAL checks have been enabled for Jinja
macro use. This should let us cut down reuse of common code snippets by
moving them to Jinja macros. We can also augment our templating system
by using Jinja macros.

#### Rationale

I opted to split out these changes from #4413 because they might be more controversial. I recall discussions about them in the past but don't remember specifics or their outcomes. My goal with these are to introduce new macros that we can use in the `lineinfile`-type templating. 

This gives us a few options for the future:

 - Use macros to augment templating. This is probably the least work option and is what I'm working towards now.
 - Replace templating entirely with macros. This means the current templating logic could write out a remediation/check/etc. which uses a Jinja macro and check that into the source tree. All changes to the content would go through the Jinja macro. Users/developers could see that the rule has a check/remediation and also be able to edit it only one place (i.e., the Jinja macro definition file).

### DIFF
--- a/ssg/oval.py
+++ b/ssg/oval.py
@@ -13,7 +13,7 @@ from .constants import oval_namespace as ovalns
 from .rules import get_rule_dir_id, get_rule_dir_ovals, find_rule_dirs
 from .xml import ElementTree as ET
 from .xml import oval_generated_header
-from .yaml import process_file
+from .jinja import process_file_with_macros
 from .id_translate import IDTranslator
 
 SHARED_OVAL = re.sub(r'ssg/.*', 'shared', __file__) + '/checks/oval/'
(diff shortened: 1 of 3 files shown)


ssg/oval.py
```diff
@@ -13,7 +13,7 @@
 from .rules import get_rule_dir_id, get_rule_dir_ovals, find_rule_dirs
 from .xml import ElementTree as ET
 from .xml import oval_generated_header
-from .yaml import process_file
+from .jinja import process_file_with_macros
```
cipherboy: This is kinda funny. The import here was technically wrong: `ssg.yaml` imports `process_file` from `ssg.jinja`... 

Note that we don't have to import `process_file_with_macros` in `ssg.yaml` as sometimes we have two sets of logic; one with Jinja macros and one without (and use `process_file` in both).

jan-cerny:
@cipherboy https://github.com/ComplianceAsCode/content/pull/4413 has been merged, please rebase

jan-cerny:
I have compared the generated RHEL8 datastreams with and without b747027257d06c45cec6b1ae2c943f064e2365eb and they're identical except time stamps. I have also checked the playbooks in `/build/rhel8/playbooks` and they're identical as well. I think it can be merged after you rebase.

cipherboy:
@jan-cerny Rebased, thanks!

PR #4416, opened 2019-06-19
Use Ninja in build_product by default

jan-cerny:
#### Description:
Attempts to find Ninja, if Ninja is available it will use it otherwise it will use make.

#### Rationale:
Ninja builds faster than make.


matejak:
LGTM, thanks!

PR #4417, opened 2019-06-19
tests/install_vm.py: fix python3 traceback in detect_url_os function

matusmarhefka:
#### Description:

In python3 the `subprocess.Popen` returns byte string instead of normal
string, therefore it needs to be decoded.


matejak:
ACK

PR #4418, opened 2019-06-19
Build multiple products

cipherboy:
#### Description:

After @jan-cerny's #4416 I felt this was the next obvious improvement (also, I wanted to build multiple profiles at once for some of my testing but I didn't want to maintain a separate list of products in my own [`build`](https://github.com/cipherboy/dotfiles/blob/master/bashrc.d/build.sh) scripts).

This adds two things:

 - `SSG_PRODUCT_DEFAULT` -- a flag which controls the behavior of all default products. This defaults to `TRUE`, and is substituted into all products which default to `TRUE`, namely all but the example one. When you disable `SSG_PRODUCT_DEFAULT`, the default value of all otherwise-default-enable products becomes `FALSE`; however you can enable any profile explicitly and it will behave as expected.
 - Updates `build_product` to support multiple products on the command line. I also simplified some of the casing logic / etc.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,29 +48,39 @@ option(SSG_SEPARATE_SCAP_FILES_ENABLED "If enabled, separate SCAP files (OVAL, X
 option(SSG_JINJA2_CACHE_ENABLED "If enabled, the jinja2 templating files will be cached into bytecode. Also see SSG_JINJA2_CACHE_DIR." TRUE)
 set(SSG_JINJA2_CACHE_DIR "${CMAKE_BINARY_DIR}/jinja2_cache" CACHE PATH "Where the jinja2 cached bytecode should be stored. This speeds up builds at the expense of disk space. You can use one location for multiple SSG builds for performance improvements.")
 
-option(SSG_PRODUCT_CHROMIUM "If enabled, the Chromium SCAP content will be built" TRUE)
-option(SSG_PRODUCT_DEBIAN8 "If enabled, the Debian 8 SCAP content will be built" TRUE)
-option(SSG_PRODUCT_EAP6 "If enabled, the JBoss EAP6 SCAP content will be built" TRUE)
+# SSG_PRODUCT_DEFAULT modifies the behavior of all other options. Products
+# which should be built by default should use the value ${SSG_PRODUCT_DEFAULT}
+# instead of the boolean True. This allows us disable all default products by
+# passing `-DSSG_PRODUCT_DEFAULT=OFF` and then manually specifying a list of
+# products to build.
+option(SSG_PRODUCT_DEFAULT "If enabled, all default release products will be built; otherwise only explicitly enabled products will be" TRUE)
+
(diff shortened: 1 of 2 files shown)


build_product
```diff
@@ -1,16 +1,27 @@
 #!/bin/bash
 
 to_uppercase() {
-	printf '%s' "$1" | tr 'a-z' 'A-Z'
+	echo "$1" | tr '[:lower:]' '[:upper:]'
 }
 
 to_lowercase() {
-	printf '%s' "$1" | tr 'A-Z' 'a-z'
+	echo "$1" | tr '[:upper:]' '[:lower:]'
```
matejak: Why to swap printf with echo? printf is a bit better, as it doesn't "swallow" (invalid) product names s.a. `-e`, `-n` etc.

CMakeLists.txt
```diff
@@ -48,29 +48,30 @@ option(SSG_SEPARATE_SCAP_FILES_ENABLED "If enabled, separate SCAP files (OVAL, X
 option(SSG_JINJA2_CACHE_ENABLED "If enabled, the jinja2 templating files will be cached into bytecode. Also see SSG_JINJA2_CACHE_DIR." TRUE)
 set(SSG_JINJA2_CACHE_DIR "${CMAKE_BINARY_DIR}/jinja2_cache" CACHE PATH "Where the jinja2 cached bytecode should be stored. This speeds up builds at the expense of disk space. You can use one location for multiple SSG builds for performance improvements.")
 
-option(SSG_PRODUCT_CHROMIUM "If enabled, the Chromium SCAP content will be built" TRUE)
-option(SSG_PRODUCT_DEBIAN8 "If enabled, the Debian 8 SCAP content will be built" TRUE)
-option(SSG_PRODUCT_EAP6 "If enabled, the JBoss EAP6 SCAP content will be built" TRUE)
+option(SSG_PRODUCT_DEFAULT "If enabled, all default release products will be built; otherwise only explicitly enabled products will be" TRUE)
```
matejak: Please add a newline or other visual separator, so the default value stands out from actual products.

build_product
cipherboy: I will change back. I had changed to avoid the string format qualifier (as `printf "%s" <var>` is equivalent to `echo` sans the edge cases you describe and seems clearer for this use case) . Under mine, invalid product names still get detected and reported correctly:

```
The argument '-n' is not supported.
```

More concisely,

```bash
tr '[:upper:]' '[:lower:]' <<< "$1"
```

would work, but I'm guessing most people aren't familiar with the `<<<` operator.

cipherboy:
@matejak Thanks for the review! I've reverted the `printf->echo` changes, added a new line and comments to the `SSG_PRODUCT_DEFAULT`and updated the help message.

build_product
matejak: Yes, did not know what `<<<` does either. I would say that in our case, printf is just about 1% better than echo :-)

matejak:
Thanks you for the PR, merging!

PR #4419, opened 2019-06-19
Test Suite: new test mode: Combined. Test a profile using test scenarios.

yuumasato:
#### Description:

- Enable testing of rules selected in a profile using ~~rule mode~~ test scenarios.
  ~~- Introduce a new argument `--profile` that will use last argument as a profile `ID`.~~
  - Adds scraper functions that return rules selected in a profile.
  - Adds a new checker `CombinedChecker`
    - The checker outputs message for every selected rule that was not tested
    - The checker overrides `# profiles` metadata in test scenarios and sets it as the target profile

#### Rationale:

- Allows batch testing with test scenarios for rules that are selected in a profile.
- Facilitate test scenario testing by overriding `# profiles` metadata, but with increased risk of test failures due to differences in variable selection and setup.

### DIFF
--- a/tests/ssg_test_suite/common.py
+++ b/tests/ssg_test_suite/common.py
@@ -6,6 +6,8 @@ import functools
 from ssg.constants import MULTI_PLATFORM_MAPPING
 from ssg.constants import PRODUCT_TO_CPE_MAPPING
 from ssg.constants import FULL_NAME_TO_PRODUCT_MAPPING
+from ssg_test_suite.log import LogHelper
+import data
 
 Scenario_run = namedtuple(
     "Scenario_run",
@@ -175,3 +177,48 @@ def matches_platform(scenario_platforms, benchmark_cpes):
     for p in scenario_platforms:
         scenario_cpes |= _get_platform_cpes(p)
     return len(scenario_cpes & benchmark_cpes) > 0
+
+
+def run_with_stdout_logging(command, args, log_file):
+    log_file.write("{0} {1}\n".format(command, " ".join(args)))
+    subprocess.check_call(
+        (command,) + args, stdout=log_file, stderr=subprocess.STDOUT)
+
+
+def send_scripts(domain_ip):
+    remote_dir = './ssgts'
+    archive_file = data.create_tarball('.')
+    remote_archive_file = os.path.join(remote_dir, archive_file)
+    machine = "root@{0}".format(domain_ip)
+    logging.debug("Uploading scripts.")
+    log_file_name = os.path.join(LogHelper.LOG_DIR, "data.upload.log")
+
+    with open(log_file_name, 'a') as log_file:
+        args = SSH_ADDITIONAL_OPTS + (machine, "mkdir", "-p", remote_dir)
+        try:
+            run_with_stdout_logging("ssh", args, log_file)
+        except Exception:
+            msg = "Cannot create directory {0}.".format(remote_dir)
+            logging.error(msg)
+            raise RuntimeError(msg)
+
+        args = (SSH_ADDITIONAL_OPTS
+                + (archive_file, "{0}:{1}".format(machine, remote_dir)))
+        try:
+            run_with_stdout_logging("scp", args, log_file)
+        except Exception:
+            msg = ("Cannot copy archive {0} to the target machine's directory {1}."
+                   .format(archive_file, remote_dir))
+            logging.error(msg)
+            raise RuntimeError(msg)
+
+        args = (SSH_ADDITIONAL_OPTS
+                + (machine, "tar xf {0} -C {1}".format(remote_archive_file, remote_dir)))
+        try:
+            run_with_stdout_logging("ssh", args, log_file)
+        except Exception:
+            msg = "Cannot extract data tarball {0}.".format(remote_archive_file)
+            logging.error(msg)
+            raise RuntimeError(msg)
+
+    return remote_dir
--- a/tests/ssg_test_suite/rule.py
+++ b/tests/ssg_test_suite/rule.py
@@ -14,6 +14,7 @@ from ssg_test_suite import xml_operations
 from ssg_test_suite import test_env
 from ssg_test_suite import common
 from ssg_test_suite.log import LogHelper
+from ssg.constants import OSCAP_PROFILE
 import data
 
 logging.getLogger(__name__).addHandler(logging.NullHandler())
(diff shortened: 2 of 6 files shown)


tests/test_suite.py
```diff
@@ -121,7 +121,9 @@ def parse_args():
                                    "target is handled as a substring - so you can "
                                    "ask for subset of all rules this way. (If you "
                                    "type ipv6 as a target, all rules containing "
-                                   "ipv6 within id will be performed."))
+                                   "ipv6 within id will be performed."
+                                   "If --profile is specified, this argument "
+                                   "specifies the profile to be tested."))
```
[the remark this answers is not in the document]
yuumasato: To have a mutually exclusive using `argparse`, the arguments must be all optional.
That would look like this:
```
usage: test_suite.py rule [-h]
                          (--docker BASE_IMAGE | --container BASE_IMAGE | --libvirt HYPERVISOR DOMAIN HYPERVISOR DOMAIN)
                          --datastream DATASTREAM
                          [--xccdf-id REF-ID | --xccdf-id-number REF-ID-SELECT]
                          [--loglevel LOGLEVEL] [--logdir LOGDIR]
                          [--mode {online,offline}]
                          [--remediate-using {oscap,bash,ansible}] [--debug]
                          [--dontclean] [--scenarios SCENARIOS_REGEX]
                          [--profile_target PROFILE_TARGET | --rule_target TARGET [TARGET ...]]
```
I'm not a fan of this approach as you'd have to specify `--rule-target` for the default behavior.
@jan-cerny Do you have a suggestion how to improve this?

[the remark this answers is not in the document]
yuumasato: I have tried to improve the argument help message, hopefully they are clearer.

tests/ssg_test_suite/common.py
```diff
@@ -26,6 +26,9 @@
     "-o", "UserKnownHostsFile=/dev/null",
 ) + SSH_ADDITIONAL_OPTS
 
+CONTENT_BASE_PREFIX = "xccdf_org.ssgproject."
+CONTENT_PROFILE_PREFIX = CONTENT_BASE_PREFIX + "content_profile_"
+CONTENT_RULE_PREFIX = CONTENT_BASE_PREFIX + "content_rule_"
```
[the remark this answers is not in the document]
yuumasato: Moved them to `constants.py`

tests/test_suite.py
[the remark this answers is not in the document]
jan-cerny: @yuumasato I don't, I haven't realized that.

tests/test_suite.py
```diff
@@ -136,6 +136,14 @@ def parse_args():
                              dest="scenarios_regex",
                              default=None,
                              help="Regular expression matching test scenarios to run")
+    parser_rule.add_argument("--source", choices=['rule', 'profile'],
+                              default='rule',
+                              dest="target_source",
+                              help=("Specify wether target is a list of rules, or a profile. "
```
jan-cerny: @yuumasato Please change it to "whether".

tests/ssg_test_suite/rule.py
```diff
@@ -340,4 +341,25 @@ def perform_rule_check(options):
...
+        if not profile.startswith(OSCAP_PROFILE):
+            profile = OSCAP_PROFILE+profile
+        logging.info("Performing rule test using profile: {0}".format(profile))
+
+        # Fetch target list from rules selected in profile
+        target_rules = xml_operations.get_all_rule_ids_in_profile(
+                options.datastream, options.benchmark_id,
+                profile, logging)
+        logging.debug("Profile {0} expanded to following list of "
+                      "rules: {1}".format(profile, target_rules))
```
jan-cerny: @yuumasato I have a question: What happens if some of the rules don't have any test scenarios?

jan-cerny:
There is a confusing thing: You provide profile ID as an argument to ./test_suite.py, but it runs test scenarios for rules according to `# profile` directive in each scenario, and that can be a different profile. It gives a false impression that the rule is tested in a context of the profile.

```
./test_suite.py rule --libvirt qemu:///system ssgts_rhel8 \
--datastream ../build/ssg-rhel8-ds.xml \
--source profile cjis

....

INFO - Script line_not_there.pass.sh using profile xccdf_org.ssgproject.content_profile_ospp OK
INFO - xccdf_org.ssgproject.content_rule_sshd_disable_rhosts
```

yuumasato:
> There is a confusing thing: You provide profile ID as an argument to ./test_suite.py, but it runs test scenarios for rules according to `# profile` directive in each scenario, and that can be a different profile. It gives a false impression that the rule is tested in a context of the profile.
> 
> ```
> ./test_suite.py rule --libvirt qemu:///system ssgts_rhel8 \
> --datastream ../build/ssg-rhel8-ds.xml \
> --source profile cjis
> 
> ....
> 
> INFO - Script wrong_value.fail.sh using profile xccdf_org.ssgproject.content_profile_ospp OK
> INFO - Script correct_value.pass.sh using profile xccdf_org.ssgproject.content_profile_ospp OK
> INFO - Script comment.pass.sh using profile xccdf_org.ssgproject.content_profile_ospp OK
> INFO - Script line_not_there.pass.sh using profile xccdf_org.ssgproject.content_profile_ospp OK
> INFO - xccdf_org.ssgproject.content_rule_sshd_disable_rhosts
> ```

@jan-cerny But these are the normal outputs of "rule mode". Each test scenario targets a profile, thus, the message with  `using profile`.

tests/ssg_test_suite/rule.py
yuumasato: The rule won't be tested.

yuumasato:
> It gives a false impression that the rule is tested in a context of the profile.

would it make sense then for the Test Suite to override the `#profile` metadata, when running "rule mode" on rules selected on a profile? 
As the rule is already selected in the argument profile, this shouldn't be a problem.
But there would be an issue with any variable specifics that the test scenario sets up.

tests/ssg_test_suite/rule.py
yuumasato: I will look into writing a message when a rule is not tested.

tests/test_suite.py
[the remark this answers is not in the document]
matejak: What about making this a new subcommand? The rule mode does what it says - you just specify the rule regex, and it tests it. This improvement adds complexity to the command-line invocation, and even more complexity is coming to the rule mode itself (We are for example thinking of narrowing down scenarios, so if you are fighting with a failing one, you get faster test suite feedback on your changes by skipping non-relevant scenarios.)
I believe that a new subcommand can be introduced with quite low overhead.

matejak:
> would it make sense then for the Test Suite to override the `#profile` metadata, when running "rule mode" on rules selected on a profile?

This makes sense, and it is one more reason why this great improvement should be separated from the rule subcommand.
Regarding variables, some may be worked around (e.g. timeouts - you write tests for 1 seconds vs 1 year, and they will pass/fail in the same fashion in all profiles), and hopefully those problematic cases won't be many, so things may be kept under control.

On a slightly related note, we should be able to use the `(all)` profile now as soon as it gets fixed in the test suite / oscap-ssh.

jan-cerny:
> why this great improvement should be separated from the rule subcommand.

Do you mean introducing a third mode next to to rule and profile? I like that, it is friendlier for the users than "profile target of rule mode".

jan-cerny:
> would it make sense then for the Test Suite to override the #profile metadata, when running "rule mode" on rules selected on a profile? 

This override makes sense also when you test a single rule, but you want to do it in a context of different profile than specified in the test scenario.

yuumasato:
> Do you mean introducing a third mode next to to rule and profile? I like that, it is friendlier for the users than "profile target of rule mode".

What if I call this mode "profile target of rule mode"?
Jokes apart, suggestion on how to call this mode?

jan-cerny:
We will have 3 modes:
1. runs test scenarios only for selected rules - "single-rule"
2. runs test scenarios for all rules in the given profile - "all-rules"
3. runs whole profile, doesn't run test scenarios - "one-pass"

tests/test_suite.py
yuumasato: This doesn't apply anymore.

tests/ssg_test_suite/rule.py
yuumasato: The `CombinedChecker` will output a list of rules not tested at the end of its run.

yuumasato:
@jan-cerny @matejak Thank you for your feedback.
I have reworked the whole thing, and updated the PR description.

jan-cerny:
@yuumasato Thanks, I'll take a look.

jan-cerny:
Works great for me!

PR #4421, opened 2019-06-20
Add option for extra repository in install_vm.py script.

ggbecker:
#### Description:

- This allows to install RHEL8 machines providing the AppStream as
extra repository.

#### Rationale:

- Provide support for adding extra repository during OS installation, enabling RHEL8 installation.


dahaic:
Works like a charm, thank you!!

PR #4422, opened 2019-06-20
Remove duplicated CCE suffix from replace_or_append bash remediation.

ggbecker:
#### Description:

- Remove duplicated `CCE-` suffix from documentation added by bash remediation.

#### Rationale:

- @CCENUM@ when expanded already contains the suffix `CCE-`.


dahaic:
Thanks!

PR #4289, opened 2019-04-17
SSH service rules for SLE12

maltek:
This one is a bit bigger.

It adds the metadata changes for a bunch of SSH rules to work with SLE12. Since the RPM package is named differently in SLE, it also needs a lot of conditionals to OVAL rules. Then there are a few rules where our STIG demands pretty much the same thing, but with a slight difference: e.g. it demands that `StrictModes` is explicitly set to `yes`, not just unset.

Then there's one thing I already know to be controversial, but would like to discuss further: `sshd_enable_warning_banner` gets two STIG IDs for SLE12 in 5f3f2bd, and 7980b4d enables that to work better. There's really no way to separate these two STIGs into different rules, because they are both about setting `Banner /etc/issue/`. We'd like to get rid of one of those STIGs in the future since there's no point in having two about the same thing, but for now adding both IDs to the same rule seems most appropriate to me.

PR #4424, opened 2019-06-20
Fix sshd OVAL checks and update/add test scenarios.

ggbecker:
#### Description:

- Fixed OVAL checks according STIG recommendations see links below for example. Test scenarios (pass/fail condition) from rules which were changed have been updated.

- Added test scenarios for couple of rules and added bash remediation for sshd_enable_x11_forwarding.

#### Rationale:

- From [STIG StrictModes](https://www.stigviewer.com/stig/red_hat_enterprise_linux_7/2017-12-14/finding/V-72263) `If "StrictModes" is set to "no", is missing, or the returned line is commented out, this is a finding. `, we can infer that some of our OVAL checks were wrong because when the line is commented out or if it is not there the OVAL check didn't interpret as fail. Now it ensures that the desired line is present and has the correct value.

- Switched some occurrences of `negate="true"`+`none_exist` to `all_exist` to avoid confusion.

- This PR does not handle multiple occurrences of the same desired line in file. e.g. if there are `StrictModes yes` and then `StrictModes no` within the same configuration file, the OVAL check will pass.

Update1: Related issue: https://bugzilla.redhat.com/show_bug.cgi?id=1714204


yuumasato:
>     * This PR does not handle multiple occurrences of the same desired line in file. e.g. if there are `StrictModes yes` and then `StrictModes no` within the same configuration file, the OVAL check will pass.

Wouldn't using `check_existence="only_one_exists " resolve this?
This attribute with this value that only one object defined by the description exists on the system.
https://github.com/OVALProject/Language/blob/master/docs/oval-common-schema.md#---existenceenumeration---

I wonder if it is worth adding a test with duplicated "desired" line, just for it fail expections. (`pass` when expected is `fail`)

linux_os/guide/services/ssh/ssh_server/sshd_enable_strictmodes/oval/shared.xml
```diff
@@ -29,14 +29,14 @@ and configurations.</description>
...
-  <ind:textfilecontent54_test check="all" check_existence="none_exist"
+  <ind:textfilecontent54_test check="all" check_existence="all_exist"
   comment="tests the value of StrictModes setting in the /etc/ssh/sshd_config file"
   id="test_sshd_enable_strictmodes" version="1">
     <ind:object object_ref="obj_sshd_enable_strictmodes" />
   </ind:textfilecontent54_test>
   <ind:textfilecontent54_object id="obj_sshd_enable_strictmodes" version="1">
     <ind:filepath>/etc/ssh/sshd_config</ind:filepath>
-    <ind:pattern operation="pattern match">^[\s]*(?i)StrictModes(?-i)[\s]+no[\s]*(?:|(?:#.*))?$</ind:pattern>
+    <ind:pattern operation="pattern match">^[\s]*(?i)StrictModes(?-i)[\s]+yes[\s]*(?:|(?:#.*))?$</ind:pattern>
```
shawndwells: The default for ``strictmodes`` is yes, so the fail condition should only be if someone disables and sets to ``no``.

shawndwells: ref: https://www.freebsd.org/cgi/man.cgi?sshd_config(5)

linux_os/guide/services/ssh/ssh_server/disable_host_auth/oval/shared.xml
```diff
@@ -25,14 +25,14 @@
...
-  <ind:textfilecontent54_test check="all" check_existence="none_exist"
+  <ind:textfilecontent54_test check="all" check_existence="all_exist"
   comment="sshd HostbasedAuthentication" id="test_sshd_hostbasedauthentication"
   version="1">
     <ind:object object_ref="object_sshd_hostbasedauthentication" />
   </ind:textfilecontent54_test>
   <ind:textfilecontent54_object id="object_sshd_hostbasedauthentication" version="2">
     <ind:filepath>/etc/ssh/sshd_config</ind:filepath>
-    <ind:pattern operation="pattern match">^[\s]*(?i)HostbasedAuthentication(?-i)[\s]+yes[\s]*(?:|(?:#.*))?$</ind:pattern>
+    <ind:pattern operation="pattern match">^[\s]*(?i)HostbasedAuthentication(?-i)[\s]+no[\s]*(?:|(?:#.*))?$</ind:pattern>
```
shawndwells: The default is ``no``, so the failure condition should be if someone specifically sets this to ``yes``

linux_os/guide/services/ssh/ssh_server/sshd_disable_kerb_auth/oval/shared.xml
```diff
@@ -29,14 +29,14 @@ the SSH Server.</description>
...
-  <ind:textfilecontent54_test check="all" check_existence="none_exist"
+  <ind:textfilecontent54_test check="all" check_existence="all_exist"
   comment="tests the value of KerberosAuthentication setting in the /etc/ssh/sshd_config file"
   id="test_sshd_disable_kerb_auth" version="1">
     <ind:object object_ref="obj_sshd_disable_kerb_auth" />
   </ind:textfilecontent54_test>
   <ind:textfilecontent54_object id="obj_sshd_disable_kerb_auth" version="1">
     <ind:filepath>/etc/ssh/sshd_config</ind:filepath>
-    <ind:pattern operation="pattern match">^[\s]*(?i)KerberosAuthentication(?-i)[\s]+yes[\s]*(?:|(?:#.*))?$</ind:pattern>
+    <ind:pattern operation="pattern match">^[\s]*(?i)KerberosAuthentication(?-i)[\s]+no[\s]*(?:|(?:#.*))?$</ind:pattern>
```
shawndwells: The default is ``no``, so the failure condition should be if ``yes`` is present.

linux_os/guide/services/ssh/ssh_server/sshd_disable_rhosts/oval/shared.xml
```diff
@@ -29,15 +29,15 @@
...
+  <ind:textfilecontent54_test check="all" check_existence="all_exist"
   comment="Tests the value of the IgnoreRhosts[\s]*(&lt;:nocomment:&gt;*) setting in the /etc/ssh/sshd_config file"
   id="test_sshd_rsh_emulation_disabled" version="1">
     <ind:object object_ref="obj_sshd_rsh_emulation_disabled" />
   </ind:textfilecontent54_test>
   <ind:textfilecontent54_object id="obj_sshd_rsh_emulation_disabled"
   version="2">
     <ind:filepath>/etc/ssh/sshd_config</ind:filepath>
-    <ind:pattern operation="pattern match">^[\s]*(?i)IgnoreRhosts(?-i)[\s]+no[\s]*(?:|(?:#.*))?$</ind:pattern>
+    <ind:pattern operation="pattern match">^[\s]*(?i)IgnoreRhosts(?-i)[\s]+yes[\s]*(?:|(?:#.*))?$</ind:pattern>
```
shawndwells: The default is ``yes``, so the failure condition should be if someone explicitly changes to ``yes``

yuumasato:
> Instead, the checks should be if someone explicitly changes away from the secure defaults.

That is the way they were.

For rule `sshd_enable_strictmodes`, the [STIG guidance](https://www.stigviewer.com/stig/red_hat_enterprise_linux_7/2017-12-14/finding/V-72263) says it is a finding if the line is setting the default value is missing.

@shawndwells How to reconcile these? Would a warning in the rule description help waive the finding?

shawndwells:
> > Instead, the checks should be if someone explicitly changes away from the secure defaults.
> 
> That is the way they were.
> 
> For rule `sshd_enable_strictmodes`, the [STIG guidance](https://www.stigviewer.com/stig/red_hat_enterprise_linux_7/2017-12-14/finding/V-72263) says it is a finding if the line is setting the default value is missing.
> 
> @shawndwells How to reconcile these? Would a warning in the rule description help waive the finding?

Sounds like a bug in the DISA content. @tedbrunell can help get DISA to fix that :)

yuumasato:
@shawndwells Thank you.

> the checks should be if someone explicitly changes away from the secure defaults.

I can agree with this, lets keep checking if deviate from secure defaults.
@ggbecker What are your thoughts on this?

ggbecker:
> @shawndwells Thank you.
> 
> > the checks should be if someone explicitly changes away from the secure defaults.
> 
> I can agree with this, lets keep checking if deviate from secure defaults.
> @ggbecker What are your thoughts on this?

It definitely makes the checks easier. I'll update the PR soon and sync with @maltek so #4289 can be updated based on this information as it also touches these rules.

Basically we need to update and recheck if the OVAL checks are in the form `none_exist` - `config nondefault`.

ggbecker:
Closing this in favor of newly added OVAL jinja macros for line in file. See: https://github.com/ComplianceAsCode/content/pull/4593 and https://github.com/ComplianceAsCode/content/pull/4624.

Rules are being ported to use these macros which aim to reduce costs of maintenance and to provide reliable OVAL checks.

PR #4426, opened 2019-06-20
Fix broken test scenario name

jan-cerny:
Renames o_trunc_write.fails.sh to o_trunc_write.fail.sh because
"fails" is not a valid XCCDF result.


dahaic:
Good catch!

PR #4427, opened 2019-06-20
Fix count_oval_object get wrong count

Sep0lkit:
#### Description:

- oval_root no point to the right oval document root, so get the wrong count._

#### Rationale:

- xccdf documents have check-content-ref: oval and ocil
- oval_root variable point to ocil and the rule get no test and no objects

### DIFF
--- a/utils/count_oval_objects.py
+++ b/utils/count_oval_objects.py
@@ -37,7 +37,7 @@ def load_xml(file_name):
         exit(-1)
 
 
-def get_ext_def_tests(oval_root,def_refs):
+def get_ext_def_tests(oval_root, def_refs):
     t = []
     for d in oval_root.findall(".//definition"):
         if d.attrib.get('id') == def_refs:
@@ -49,9 +49,9 @@ def get_ext_def_tests(oval_root,def_refs):
             t.append(test_ref)
         for extend_def in definition.findall(".//extend_definition"):
             extend_ref = extend_def.attrib["definition_ref"]
-            t = t + (get_ext_def_tests(oval_root,extend_ref))
+            t = t + (get_ext_def_tests(oval_root, extend_ref))
     return t
-        
+
 
 def find_oval_objects(oval_refs):
     ''' Finds OVAL objects according to definitions ID '''
@@ -77,9 +77,8 @@ def find_oval_objects(oval_refs):
                 tests.append(test_ref)
             for extend_def in definition.findall(".//extend_definition"):
                 extend_ref = extend_def.attrib["definition_ref"]
-                #extend_refs.append(extend_ref)
-                t = get_ext_def_tests(oval_root,extend_ref)
-                tests += t 
+                t = get_ext_def_tests(oval_root, extend_ref)
+                tests += t
 
     # find references to objects in tests
     for key in oval_files:
--- a/utils/count_oval_objects.py
+++ b/utils/count_oval_objects.py
@@ -37,6 +37,22 @@ def load_xml(file_name):
         exit(-1)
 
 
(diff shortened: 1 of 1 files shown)


yuumasato:
@open ok to test

yuumasato:
@openscap-ci ok to test

yuumasato:
@Sep0lkit Nice catch! Thanks for the fix.
 
The objects defined in definitions which are referenced only via `extend_definition` are not being counted, although the definition itself is linked to a Rule. So I think they should _probably_ be counted? 
What do you think?

See for example rule `service_chronyd_or_ntpd_enabled`, it extends `oval:ssg-service_chronyd_enabled:def:1`, but its object `oval:ssg-obj_service_running_chronyd:obj:1` is not counted.

Would you be interested in fixing this issue too?

Sep0lkit:
@yuumasato understand your view, I will try to fix this issue.

yuumasato:
ITOH, if a definition is extended by multiple definitions, the OVAL count will be bloated up.

utils/count_oval_objects.py
```diff
@@ -54,29 +71,38 @@ def find_oval_objects(oval_refs):
...
                 break
         if definition is not None:
+            extend_refs = []
             for criterion in definition.findall(".//criterion"):
                 test_ref = criterion.attrib["test_ref"]
                 tests.append(test_ref)
+            for extend_def in definition.findall(".//extend_definition"):
+                extend_ref = extend_def.attrib["definition_ref"]
+                t = get_ext_def_tests(oval_root, extend_ref)
+                tests += t
```
yuumasato: As far as I know, there shouldn't be difference under the hood, but lets be consistent and use `tests.extend()`

utils/count_oval_objects.py
```diff
@@ -28,14 +28,31 @@ def load_xml(file_name):
...
+                el.tag = el.tag.split('}', 1)[1]  # strip all namespaces
         root = it.root
         return root
     except:
         sys.stderr.write("Error while loading file " + file_name + ".\n")
         exit(-1)
 
 
+def get_ext_def_tests(oval_root, def_refs):
+    t = []
```
yuumasato: Please name the variable clearly, for example `extended_tests`.

utils/count_oval_objects.py
```diff
@@ -54,29 +71,38 @@ def find_oval_objects(oval_refs):
                 definition = d
                 break
         if definition is not None:
+            extend_refs = []
             for criterion in definition.findall(".//criterion"):
                 test_ref = criterion.attrib["test_ref"]
                 tests.append(test_ref)
+            for extend_def in definition.findall(".//extend_definition"):
+                extend_ref = extend_def.attrib["definition_ref"]
+                t = get_ext_def_tests(oval_root, extend_ref)
```
yuumasato: Here as well, name the variable more significantly.

yuumasato:
@Sep0lkit Thanks for updating the PR, and sorry for delay in review.

The script will now list the all OVAL object types used in each rule correctly.

But I have noticed that the numbers listed at the end are not actually the number of times an object is used/instantiated, but the number of times a rule used an object of that type.

The summary title of `Count of used OVAL objects` is misleading, could you updated it to `Count of rules using the object`.

matejak:
Hello @Sep0lkit, there are some small suggestions to your PR, so we would love to see more commits from you. Do you still plan to pursue the merge?

shawndwells:
Closing due to approximately 5 months of inactivity. 

@Sep0lkit or @matejak , if you'd like to reopen and continue this work, please do!

Issue #3645, opened 2018-12-05
Ansible remediation for 'Disable SSH Support for User Known Hosts' breaks sshd when there are Match groups


eclausell:
#### Description of problem:
Encountered on RHEL 7.5. If /etc/sshd_config contains a Match block, the current remediation

#### SCAP Security Guide Version:
0.1.38 

#### Operating System Version:
RHEL 7.5

#### Steps to Reproduce:

1. On system with ansible (2.4.2), openssh, openscap-scanner and  scap-security-guide installed, (and /etc/ansible/hosts configured for the local), modify /etc/ssh/sshd_config so that it:
  Does not have a ignoreUserKnownHosts configuration
  Has a Match group at the end of the file, e.g. 
```
Subsystem sftp internal-sftp
Match Group users
    ChrootDirectory /home
    AllowTCPForwarding no
    X11Forwarding no
    ForceCommand internal-sftp
<EOF>
```
2. Run the RHEL 7.5 Stig playbook, for the sshd_disable_user_known_hosts rule.
`ansible-playbook --tags=sshd_disable_user_known_hosts  /usr/share/scap-security-guide/ansible/ssg-rhel7-role-stig-rhel7-disa.yml`

#### Actual Results:
![image](https://user-images.githubusercontent.com/20073280/49538038-1b316080-f890-11e8-84fb-09f37f99da6e.png)

Fails because the IgnoreUserKnownHosts entry is added at the end of the file and is considered part of the Match group.
#### Expected Results:
![image](https://user-images.githubusercontent.com/20073280/49537915-d60d2e80-f88f-11e8-8ece-4e1509e347b4.png)

With IgnoreUserKnownHosts added before first Match group .

#### Addition Information/Debugging Steps:
There are likely similar problems with other sshd_config remediation tasks and bash remediations that append to sshd_config.

yuumasato:
@eclausell Would adding to beginning of file be a better approach? Could there be similar problems?

eclausell:
I believe that would work.

ggbecker:
Ensure that bash remediation also has the same behavior.

PR #4428, opened 2019-06-20
Add Ansible macros for lineinfile remediations

cipherboy:
#### Description:

Initial work-in-progress PR for `lineinfile`-style Ansible remediations. I'd appreciate feedback regarding initial design. I'll be updating this PR with additional Ansible macros to handle configuration directories (see the example [`in my testbed`](https://github.com/cipherboy/testbed/blob/master/ansible-lineinfile/roles/check-config-dir-duplicates/tasks/main.yml)). 

This PR adds `lineinfile` style ansible remediations as Jinja macros. In addition to the "simple" `lineinfile` remediation (a direct passthrough to turn a macro call into Ansible script), there are `one_line` and `config` based remediations. The former ensures there's only one of the specified line in the file, the latter handles building a regex from the parameter (with customizable prefix and separators regex parts). If absent, it will add the line `parameter + separator + value` to the file. 

One example is the rule `disable_host_auth` to set `HostbasedAuthentication no`  in `/etc/ssh/sshd_config`.

#### Goals:

- Add Ansible macros in this PR
- Add Bash macros to accomplish the same in a later PR.
- Add OVAL checks to check for these in a later PR.
- Extend the templating system to use these macros.

#### Rationale:

Hopefully this PR makes sense. By extending the use of macros everywhere (in #4415), we can hopefully make templating easier. This would also set us up to simplify templating later if we wanted to.

I think there's three common forms:
 - A basic config file,
 - A config file + a config directory,
 - A structured config file (ini, xml, etc).

Of these, this PR will hope to address 1 and 2 most thoroughly, though I'd like to add support for `ini` files in Ansible as well. Supporting `ini` and `xml` will be much harder in `bash` remediations. I haven't yet looked at OVAL support.

### DIFF
--- a/linux_os/guide/services/ssh/ssh_server/disable_host_auth/ansible/shared.yml
+++ b/linux_os/guide/services/ssh/ssh_server/disable_host_auth/ansible/shared.yml
@@ -3,9 +3,4 @@
 # strategy = restrict
 # complexity = low
 # disruption = low
-- name: Disable Host-Based Authentication
-  lineinfile:
-    create: yes
-    dest: /etc/ssh/sshd_config
-    regexp: ^HostbasedAuthentication
-    line: HostbasedAuthentication no
+{{{ ansible_sshd_set(parameter="HostbasedAuthentication", value="no") }}}
--- a/linux_os/guide/services/ssh/ssh_server/sshd_disable_user_known_hosts/ansible/shared.yml
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_disable_user_known_hosts/ansible/shared.yml
@@ -3,13 +3,4 @@
 # strategy = restrict
 # complexity = low
 # disruption = low
-- name: "Disable SSH Support for User Known Hosts"
-  lineinfile:
-    create: yes
-    dest: /etc/ssh/sshd_config
-    regexp: ^IgnoreUserKnownHosts
-    line: IgnoreUserKnownHosts yes
-    insertbefore: ^Match
-    firstmatch: yes 
-    validate: /usr/sbin/sshd -t -f %s
-  #notify: restart sshd
+{{{ ansible_sshd_set(parameter="IgnoreUserKnownHosts", value="yes") }}}
--- a/linux_os/guide/services/ssh/ssh_server/sshd_print_last_log/ansible/shared.yml
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_print_last_log/ansible/shared.yml
@@ -1,9 +1,6 @@
 # platform = multi_platform_rhel,multi_platform_ol,multi_platform_rhv
-- name: Print last log
-  lineinfile:
-    create: yes
-    dest: /etc/ssh/sshd_config
-    regexp: ^PrintLastLog
-    line: PrintLastLog yes
-    validate: /usr/sbin/sshd -t -f %s
-  #notify: restart sshd
+# reboot = false
+# strategy = restrict
+# complexity = low
+# disruption = low
+{{{ ansible_sshd_set(parameter="PrintLastLog", value="yes") }}}
--- a/shared/macros-ansible.jinja
+++ b/shared/macros-ansible.jinja
@@ -1,0 +1,152 @@
+{{#
+  A wrapper over the Ansible lineinfile module. This handles the most common
+  options for us. regex is optional and when blank, it won't be included in
+  the Ansible script; this allows arbitrary additions to files. new_line will
+  only be passed when state is present. with_items will be specified only if
+  non-empty, allowing for iterating through a variable of content (with the
+  appropriate macro-based path). register will be specified only if non-empty,
+  allowing for saving the output of this lineinfile module.
+
+  Note that all string-like parameters are single quoted in the YAML.
+#}}
+{{%- macro ansible_lineinfile(msg='', path='', regex='', new_line='', create='no', state='present', with_items='', register='', when='', validate='', insert_after='', insert_before='') -%}}
+- name: "{{{ msg or rule_title }}}"
+  lineinfile:
+    path: '{{{ path }}}'
+    create: {{{ create }}}
+    {{%- if regex %}}
+    regexp: '{{{ regex }}}'
+    {{%- endif %}}
+    {{%- if state == 'present' %}}
+    line: '{{{ new_line }}}'
+    state: present
+    {{%- if insert_after %}}
+    insertafter: '{{{ insert_after }}}'
+    {{%- elif insert_before %}}
+    insertbefore: '{{{ insert_before }}}'
+    {{%- endif %}}
+    {{% else %}}
+    state: '{{{ state }}}'
+    {{%- endif %}}
+    {{%- if validate %}}
+    validate: '{{{ validate }}}'
+    {{%- endif %}}
+  {{%- if with_items %}}
+  with_items: '{{{ with_items }}}'
+  {{%- endif %}}
+  {{%- if register %}}
+  register: '{{{ register }}}'
+  {{%- endif %}}
+  {{%- if when %}}
+  when: '{{{ when }}}'
+  {{%- endif %}}
+{{%- endmacro %}}
+
+{{#
+  Check the file system status of an object. Not a full implementation.
+#}}
+{{%- macro ansible_stat(msg='', path='', register='') %}}
+- name: '{{{ msg or rule_title }}}'
+  stat:
+    path: '{{{ path }}}'
+  {{%- if register %}}
+  register: '{{{ register }}}'
+  {{%- endif %}}
+{{%- endmacro %}}
+
+{{#
+  Find files matching a particular value. Not a full implementation.
+#}}
+{{%- macro ansible_find(msg='', paths='', recurse='yes', follow='no', contains='', register='', when='') %}}
+- name: '{{{ msg or rule_title }}}'
+  find:
+    paths: '{{{ paths }}}'
+    recurse: '{{{ recurse }}}'
+    follow: '{{{ follow }}}'
+    {{%- if contains %}}
+    contains: '{{{ contains }}}'
+    {{%- endif %}}
+  {{%- if register %}}
+  register: '{{{ register }}}'
+  {{%- endif %}}
+  {{%- if when %}}
+  when: '{{{ when }}}'
+  {{%- endif %}}
+{{%- endmacro %}}
+
+{{#
+  A wrapper for adding one, unique line to a file. A regex must be specified
+  to tell if the line is unique. This is helpful in configuration files where
+  a single configuration parameter might have multiple values, but only one
+  value is approved. All lines matching the regex are first removed and then
+  the new line is appended to the file.
+#}}
+{{%- macro ansible_only_lineinfile(msg, path, line_regex, new_line, create='no', block=False, validate='', insert_after='', insert_before='') -%}}
+{{%- if block %}}
+- name: "{{{ msg or rule_title }}}"
+  block:
+    {{{ ansible_lineinfile("Deduplicate values from " + path, path, regex=line_regex, new_line=None, create='no', state='absent')|indent }}}
+    {{{ ansible_lineinfile("Insert correct line to " + path, path, regex=None, new_line=new_line, create=create, state='present', validate=validate, insert_after=insert_after, insert_before=insert_before)|indent }}}
+{{%- else %}}
+{{{ ansible_lineinfile("Deduplicate values from " + path, path, regex=line_regex, new_line=None, create='no', state='absent') }}}
+{{{ ansible_lineinfile("Insert correct line into " + path, path, regex=None, new_line=new_line, create=create, state='present', validate=validate, insert_after=insert_after, insert_before=insert_before) }}}
+{{%- endif %}}
+{{%- endmacro %}}
+
+{{#
+  Ensure the configuration is set in a file. Note this handles generic
+  key-seperator-value files with no sense of structure. In particular,
+  ini configuration files are best served with the ini Ansible module
+  instead of lineinfile-based solutions.
+#}}
+{{%- macro ansible_set_config_file(msg, file, parameter, separator=' ', separator_regex='\s+', value='', prefix_regex='^\s*', create='no', validate='', insert_after='', insert_before='') %}}
+{{{ ansible_only_lineinfile(msg, file, prefix_regex + parameter + separator_regex, parameter + separator + value, create=create, block=True, validate=validate, insert_after=insert_after, insert_before=insert_before) }}}
+{{%- endmacro %}}
+
+{{#
+  Ensure the configuration is set in a file and not conflicted by a
+  configuration in a directory. Note this handles generic key-separator-value
+  files with no sense of structure. In particular, ini configuration files are
+  best served with the ini Ansible module instead of lineinfile-based
+  solutions.
+#}}
+{{%- macro ansible_set_config_file_dir(msg, config_file, config_dir, set_file, parameter, separator=' ', separator_regex='\s+', value='', prefix_regex='^\s*', create='no', validate='', insert_after='', insert_before='') %}}
+{{%- set var_dir = config_dir | replace("/", "_") | replace("-", "_") | replace(".", "_") -%}}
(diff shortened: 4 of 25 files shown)


shared/macros-ansible.jinja
```diff
@@ -0,0 +1,54 @@
...
+
+{{#
+  A wrapper for adding one, unique line to a file. A regex must be specified
+  to tell if the line is unique. This is helpful in configuration files where
+  a single configuration parameter might have multiple values, but only one
+  value is approved. All lines matching the regex are first removed and then
+  the new line is appended to the file.
+#}}
+{{% macro ansible_one_line_in_file(msg, path, line_regex, new_line, create='no') -%}}
+{{{ ansible_simple_line_in_file("[DEDUPLICATE] " + msg, path, regex=line_regex, new_line=None, create='no', state='absent') }}}
```
cipherboy: Technically we could add a third one here which only checks for the if the line already exists in the file. If it does, skip these two steps. The problem is that the current Ansible roles append:

```yaml
when: ansible_virtualization_role != "guest" or ansible_virtualization_type != "docker"
```

(and YAML/Ansible doesn't support joining multiple `when` clauses with an `AND`...) so we can't easily tie the three together like that.

yuumasato: I'm not sure I follow here, when the conditions are in a list won't Ansible `AND`' them?

cipherboy: Ahhh I understand now. I wasn't aware that we'd automatically convert this to a list. My bad -- still getting familiar with the build system. :)

shared/macros-ansible.jinja
```diff
@@ -0,0 +1,54 @@
...
+  the Ansible script; this allows arbitrary additions to files. new_line will
+  only be passed when state is present. with_items will be specified only if
+  non-empty, allowing for iterating through a variable of content (with the
+  appropriate macro-based path). register will be specified only if non-empty,
+  allowing for saving the output of this lineinfile module.
+
+  Note that all string-like parameters are single quoted in the YAML.
+#}}
+{{% macro ansible_simple_line_in_file(msg='', path='', regex='', new_line='', create='no', state='present', with_items='', register='') -%}}
+- name: "{{{ msg }}}"
```
jan-cerny: A small idea for improvement: In some of the rules, we use `- name: "{{{ rule_title }}}"` which means the Ansible task has the same name as the title of the rule it belongs to. We can make the `msg` parameter optional and if `msg` isn't present we can put the rule title to `name` field.

shared/macros-ansible.jinja
```diff
@@ -0,0 +1,54 @@
...
+{{#
+  A wrapper for adding one, unique line to a file. A regex must be specified
+  to tell if the line is unique. This is helpful in configuration files where
+  a single configuration parameter might have multiple values, but only one
+  value is approved. All lines matching the regex are first removed and then
+  the new line is appended to the file.
+#}}
+{{% macro ansible_one_line_in_file(msg, path, line_regex, new_line, create='no') -%}}
+{{{ ansible_simple_line_in_file("[DEDUPLICATE] " + msg, path, regex=line_regex, new_line=None, create='no', state='absent') }}}
+{{{ ansible_simple_line_in_file(msg, path, regex=None, new_line=new_line, create=create, state='present') }}}
```
jan-cerny: Another idea: This creates 2 independent Ansible tasks. Our build system adds tags to each task, so the output becomes little bloated (`build/rhel8/fixes/ansible/disable_host_auth.yml`) Do you think it would be feasible to use blocks (https://docs.ansible.com/ansible/latest/user_guide/playbooks_blocks.html) to merge the 2 tasks together?  I don't know if it's possible due to the indentation, but it might be worth investigating.

cipherboy: Ah that'd be an interesting idea, I'll go and try it. :) Thanks @jan-cerny!

shared/macros-ansible.jinja
cipherboy: @jan-cerny Is there a oneline for this in Jinja or do I need an `if`/`else` statement?

shared/macros-ansible.jinja
yuumasato: No need to apologize. :)

shared/macros-ansible.jinja
jan-cerny: @cipherboy There is a built-in `default` filter (http://jinja.pocoo.org/docs/2.10/templates/#list-of-builtin-filters), but the if/else statement is well readable.

shared/macros-ansible.jinja
```diff
@@ -0,0 +1,54 @@
...
+{{%- endmacro %}}
+
+{{#
+  Ensure the configuration is set in a file. Note this handles generic
```
jan-cerny: Another idea: It would be awesome if the macro is mentioned also in Developer's guide.

shared/macros-ansible.jinja
cipherboy: I had gone with `msg or rule_title` which appears to work as well.

shared/macros-ansible.jinja
cipherboy: Yep, I'm thinking of documenting "how to add new macros" as well.

shared/macros-ansible.jinja
cipherboy: @jan-cerny Jinja appears to provide a "indent" helper function so this is doable. :)

cipherboy:
Sample generated Ansible content:

```yaml
<snip>

- name: Set SSH Client Alive Max Count
  block:

    - name: Deduplicate values from /etc/ssh/sshd_config
      lineinfile:
        path: /etc/ssh/sshd_config
        create: false
        regexp: (?i)^\s*ClientAliveCountMax\s+
        state: absent

    - name: Insert correct line to /etc/ssh/sshd_config
      lineinfile:
        path: /etc/ssh/sshd_config
        create: true
        line: ClientAliveCountMax {{ var_sshd_set_keepalive }}
        state: present
        validate: /usr/sbin/sshd -t -f %s
  when: ansible_virtualization_role != "guest" or ansible_virtualization_type != "docker"
  tags:

<snip>
```

linux_os/guide/services/ssh/ssh_server/sshd_print_last_log/ansible/shared.yml
```diff
@@ -1,9 +1,6 @@
...
-    create: yes
-    dest: /etc/ssh/sshd_config
-    regexp: ^PrintLastLog
-    line: PrintLastLog yes
-    validate: /usr/sbin/sshd -t -f %s
-  #notify: restart sshd
+# reboot = false
+# strategy = restrict
+# complexity = low
+# disruption = low
```
cipherboy: These were missing from this file so I added them.

shared/macros-ansible.jinja
```diff
@@ -0,0 +1,128 @@
...
+  block:
+    {{{ ansible_lineinfile("Deduplicate values from " + config_file, config_file, regex=line_regex, create='no', state='absent')|indent }}}
+    {{{ ansible_stat("Check if " + config_dir + " exists", path=config_dir, register=dir_exists)|indent }}}
+    {{{ ansible_find("Check if the parameter " + parameter + " is present in " + config_dir, paths=config_dir, contains=line_regex, register=dir_parameter, when=find_when)|indent }}}
+    {{{ ansible_lineinfile("Remove parameter from files in " + config_dir, path="{{ item.path }}", regex=line_regex, state="absent", with_items=lineinfile_items, when=lineinfile_when)|indent }}}
+    {{{ ansible_lineinfile("Insert correct line to " + set_file, set_file, new_line=new_line, create=create, state='present', validate=validate)|indent }}}
+{{%- endmacro %}}
+
+{{%- macro ansible_sshd_set(msg, parameter, value) %}}
+{{{ ansible_set_config_file(msg, "/etc/ssh/sshd_config", parameter, value=value, create="yes", prefix_regex='(?i)^\s*', validate="/usr/sbin/sshd -t -f %s") }}}
```
cipherboy: `sshd` configuration (unlike `ssh` client configuration) doesn't have a directory. Parameters are case sensitive, but values aren't. We also have a utility to validate the new configuration.

cipherboy:
Sample generated `/etc/profile` content to test directories:

```yaml
- name: Set Interactive Session Timeout
  block:

    - name: Deduplicate values from /etc/profile
      lineinfile:
        path: /etc/profile
        create: false
        regexp: ^\s*TMOUT=
        state: absent

    - name: Check if /etc/profile.d exists
      stat:
        path: /etc/profile.d
      register: _etc_profile_d_exists

    - name: Check if the parameter TMOUT is present in /etc/profile.d
      find:
        paths: /etc/profile.d
        recurse: 'yes'
        follow: 'no'
        contains: ^\s*TMOUT=
      register: _etc_profile_d_has_parameter
      when: _etc_profile_d_exists.stat.isdir is defined and _etc_profile_d_exists.stat.isdir

    - name: Remove parameter from files in /etc/profile.d
      lineinfile:
        path: '{{ item.path }}'
        create: false
        regexp: ^\s*TMOUT=
        state: absent
      with_items: '{{ _etc_profile_d_has_parameter.files }}'
      when: _etc_profile_d_has_parameter.matched

    - name: Insert correct line to /etc/profile
      lineinfile:
        path: /etc/profile
        create: true
        line: TMOUT={{ var_accounts_tmout }}
        state: present
        validate: bash -n %s
  tags:
    - accounts_tmout
```

shared/macros-ansible.jinja
jan-cerny: @cipherboy wow, it does!

linux_os/guide/services/ssh/ssh_server/disable_host_auth/ansible/shared.yml
```diff
@@ -3,9 +3,4 @@
 # strategy = restrict
 # complexity = low
 # disruption = low
-- name: Disable Host-Based Authentication
-  lineinfile:
-    create: yes
-    dest: /etc/ssh/sshd_config
-    regexp: ^HostbasedAuthentication
-    line: HostbasedAuthentication no
+{{{ ansible_sshd_set('', "HostbasedAuthentication", "no") }}}
```
jan-cerny: It would be better to make  `msg` an optional argument because you always pass an empty string there anyway.

cipherboy: Yeah that's a good change. That and documentation are the two things I want to do still in this patch. 

Other than the two sets of rules I've changed, I don't think I'll track down everywhere these new macros can be used.

cipherboy:
@matejak, @jan-cerny, and @yuumasato: I think this is a good scope for this PR and would appreciate review. 

We've added some remediations using the new macros so we test that they work. We've added the macros themselves and documented them.

I'll move onto Bash macros next.

jan-cerny:
@cipherboy I like it very much, it's great that it's very easy to use. I have checked the generated playbooks and they look fine.

linux_os/guide/services/ssh/ssh_server/sshd_disable_user_known_hosts/ansible/shared.yml
```diff
@@ -3,13 +3,4 @@
 # strategy = restrict
 # complexity = low
 # disruption = low
-- name: "Disable SSH Support for User Known Hosts"
-  lineinfile:
-    create: yes
-    dest: /etc/ssh/sshd_config
-    regexp: ^IgnoreUserKnownHosts
-    line: IgnoreUserKnownHosts yes
-    insertbefore: ^Match
```
cipherboy: This is the only thing I didn't replicate. However, I can't seem to figure out *why* this was added. This configuration value isn't affected by `Match` groups...

cipherboy: Note that the bash remediation doesn't care [about this](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/ssh/ssh_server/sshd_disable_user_known_hosts/bash/shared.sh).

cipherboy: It appears this is the relevant bug report: https://github.com/ComplianceAsCode/content/issues/3645 

I'm guessing most of our ssh configuration parameters have this or similar problems. Should we just add this to the top of the configuration file?

cipherboy: Ok, this is fixed; all SSH-based thingies are added before `Match` sections.

cipherboy:
Most correctly, the behavior I think we want for SSH config is:

 - Remove all conflicting options
 - If there's a match group, add just before the first match group. 
 - Else add to the end of the file.

But that could be extended in a later PR.

shared/macros-ansible.jinja
```diff
@@ -0,0 +1,152 @@
...
+#}}
+{{%- macro ansible_only_lineinfile(msg, path, line_regex, new_line, create='no', block=False, validate='', insert_after='', insert_before='') -%}}
+{{%- if block %}}
+- name: "{{{ msg or rule_title }}}"
+  block:
+    {{{ ansible_lineinfile("Deduplicate values from " + path, path, regex=line_regex, new_line=None, create='no', state='absent')|indent }}}
+    {{{ ansible_lineinfile("Insert correct line to " + path, path, regex=None, new_line=new_line, create=create, state='present', validate=validate, insert_after=insert_after, insert_before=insert_before)|indent }}}
+{{%- else %}}
+{{{ ansible_lineinfile("Deduplicate values from " + path, path, regex=line_regex, new_line=None, create='no', state='absent') }}}
+{{{ ansible_lineinfile("Insert correct line into " + path, path, regex=None, new_line=new_line, create=create, state='present', validate=validate) }}}
```
yuumasato: Is there any reason for this second call of `ansible_lineinfile` not setting `insert_after`?

cipherboy:
@yuumasato Ah sorry. Fixed by update. :)

jan-cerny:
@cipherboy Looks great!
I have tried to run a couple of test scenarios with `--remediate-using ansible` and they pass.

```
[jcerny@thinkpad tests{cipherboy-jinja-ansible-lineinfile}]$ ./test_suite.py rule --libvirt qemu:///system ssgts_rhel8 --datastream ../build/ssg-rhel8-ds.xml --remediate-using ansible rule_sshd_disable_gssapi_a
uth
INFO - Logging into /home/jcerny/scap-security-guide/tests/logs/rule-custom-2019-06-25-1247/test_suite.log
INFO - xccdf_org.ssgproject.content_rule_sshd_disable_gssapi_auth
INFO - Script line_not_there.fail.sh using profile xccdf_org.ssgproject.content_profile_ospp OK
[jcerny@thinkpad tests{cipherboy-jinja-ansible-lineinfile}]$ ./test_suite.py rule --libvirt qemu:///system ssgts_rhel8 --datastream ../build/ssg-rhel8-ds.xml --remediate-using ansible accounts_tmout
INFO - Logging into /home/jcerny/scap-security-guide/tests/logs/rule-custom-2019-06-25-1251/test_suite.log
INFO - xccdf_org.ssgproject.content_rule_accounts_tmout
INFO - Script supercompliance.pass.sh using profile xccdf_org.ssgproject.content_profile_ospp OK
INFO - Script line_not_there.fail.sh using profile xccdf_org.ssgproject.content_profile_ospp OK
[jcerny@thinkpad tests{cipherboy-jinja-ansible-lineinfile}]$ ./test_suite.py rule --libvirt qemu:///system ssgts_rhel8 --datastream ../build/ssg-rhel8-ds.xml --remediate-using ansible rule_sshd_disable_kerb_auth
INFO - Logging into /home/jcerny/scap-security-guide/tests/logs/rule-custom-2019-06-25-1257/test_suite.log
INFO - xccdf_org.ssgproject.content_rule_sshd_disable_kerb_auth
INFO - Script line_not_there.pass.sh using profile xccdf_org.ssgproject.content_profile_ospp OK
```

@yuumasato Do you agree it can be merged?

yuumasato:
> @yuumasato Do you agree it can be merged?

Yes, this looks great. 
@cipherboy Thank you.

cipherboy:
Cool, thanks folks!

PR #4429, opened 2019-06-20
Quietly handle documentation not complete errors

cipherboy:
When the documentation isn't complete, `ssg.yaml` would silently return
None. This means that code would just detect if the result was `None`, not
knowing why (it could be an empty profile).

Instead, we raise a `DocumentationNotComplete` error: this will tell
callers why the result was `None` and what to do to fix it (run in debug
mode). In the most common output, this now shows:

    Skipping incomplete profile rhel7/profiles/anssi_nt28_minimal.profile. To include incomplete profiles, build in debug mode.

Instead of the former message:

    Could not parse profile `rhel7/profiles/anssi_nt28_intermediary.profile`.

Hopefully this message will be clearer for your forgetful former intern... :)


dahaic:
Thanks!

PR #4241, opened 2019-04-03
Test shadow-utils presence with CPE applicability

yuumasato:
#### Description:

- Introduce shadow-utils CPE
- Set relevant rules with `platform: shadow-utils`

#### Rationale:

- Some environments may not have package `shadow-utils` installed

#### Notes:
- The way we handle the CPE dictionaries is not great, and will need improvements to scale.

PR #4430, opened 2019-06-20
Add rule sssd_run_as_sssd_user for SSSD to drop root privileges.

adelton:
#### Description:

- Add rule sssd_run_as_sssd_user for SSSD to drop root privileges.

#### Rationale:

- Running SSSD as non-root decreases the chance of bad things happening.

This is an initial stab at the rule. Missing are the `identifiers` and `references` in `rule.yml` and check for the SSSD service being enabled. I need some guidance on those.


linux_os/guide/services/sssd/sssd_run_as_sssd_user/rule.yml
```diff
@@ -0,0 +1,28 @@
...
+    To minimize privileges of SSSD processes, they are configured to
+    run as non-root user.
+
+severity: medium
+
+identifiers:
+    cce@rhel8: 12345-5
+
+references:
+    ospp: FIA_AFL.1-FIXME-and-identifiers
```
yuumasato: Regarding OSPP, I could not identify to which requirement this rule is related.
Closest thing was an Objective for the Operational Environment for the user,([OE.PROPER_USER](https://www.niap-ccevs.org/MMO/PP/-442-/#SecurityObjectivesTOEorEnvironment)), but it is not related to service users, though.

But I found requirements that support this rule, like:

- [FPT_ACF_EXT.1.1](https://www.niap-ccevs.org/MMO/PP/-442-/#FPT_ACF_EXT.1.1)
- [FPT_ACF_EXT.1.1](https://www.niap-ccevs.org/MMO/PP/-442-/#FDP_ACF_EXT.1.1)

@stevegrubb Can you give us a hint?

adelton:
> that it should be evaluated when `sssd.service` is installed in the system

Well, better applicability check would be to test if the `sssd.service` is enable, wouldn't it? Do we have a way to do that easily?

yuumasato:
> > that it should be evaluated when `sssd.service` is installed in the system
> 
> Well, better applicability check would be to test if the `sssd.service` is enable, wouldn't it? 

One could argue that if the service is installed, there is possibility to configure it. Independent of its active or enabled state.

>Do we have a way to do that easily?

Yes, instead of checking for package installed with `rpminfo`, checking for the service status with `systemdunitproperty`.

adelton:
Updated `identifiers` and `references`, rebased on master -> a43134a0cf91850b24ea0e8b78b25dda29fdef3f.

adelton:
> As we briefly discussed, only issue with this rule is WRT to applicability of the rule, that it should be evaluated when `sssd.service` is installed in the system. To address this, something like `platform: sssd` in the `rule.yml` could be introduced for SSSD rules.
> 
> But I don't consider this issue as blocking this PR.
> See, #4241 to have an idea how it is like.

I'm now adding it to this pull request -> a7a6ab3c5295f5042f7afe5bf622e80b7fb79509.

Ideally we would check for the service to be also enabled but let's leave this for the future. Subsequent change (likely after this PR is merged) would then get rid of the `service_sssd_disabled` criterions in other SSSD OVALs.

adelton:
I've now updated the first commit to remove the proposed CEE from shared/references/cce-redhat-avail.txt, per recommendation in https://github.com/ComplianceAsCode/content/pull/4445#issuecomment-505304858.

linux_os/guide/services/sssd/sssd_run_as_sssd_user/rule.yml
```diff
@@ -0,0 +1,29 @@
...
+    run as non-root user.
+
+severity: medium
+
+identifiers:
+    cce@rhel8: 81062-2
+
+references:
+    ospp: FPT_ACF_EXT.1.1
+    ospp: FDP_ACF_EXT.1.1
```
yuumasato: So it seems that the correct reference should be `FMT_SMF_EXT.1`.

ssg/constants.py
```diff
@@ -383,6 +383,7 @@
     "shadow-utils": "cpe:/a:shadow-utils",
     "systemd": "cpe:/a:systemd",
     "yum": "cpe:/a:yum",
+    "sssd": "cpe:/a:sssd",
```
yuumasato: nitpick: Please keep the CPEs ordered alphabetically after `machine` and `container` platform CPEs.

linux_os/guide/services/sssd/sssd_run_as_sssd_user/rule.yml
adelton: Fixed in 57abcf27fdf6d46c0251e0fc4a916a5d163a0662.

ssg/constants.py
adelton: Fixed in 8fb9b8c2cfeb7659952fab86a69739ff9f05edd6.

yuumasato:
@openscap-ci test this please

yuumasato:
Something strange is going on with Jenkins...

yuumasato:
Thank you for the rule.

PR #4431, opened 2019-06-20
Add verbose output to the verbose logs

jan-cerny:
The verbose logs for "remediation" and "final" steps
didn't contain verbose output.


matejak:
Great improvement, this is how the verbose option should be inherited.

PR #4433, opened 2019-06-21
Add partition rules to mount_options.csv file for RHEL8 and update test scenarios.

ggbecker:
Update mount_options.csv file in RHEL8 with required partitions as part of ccc rules.
Update test scenarios to unmount tested partition in rule_mount_option_var_tmp_noexec (otherwise it can be already mounted and the test scenario fails to setup).


matejak:
LGTM. Tests/RHEL8 profiles need to be tweaked a bit in order to test this PR, as the `mount_option_var_tmp_noexec` rule is not part of any profile.
Thank you for the PR!

PR #4420, opened 2019-06-19
Add ansible remediation for audit_rules_unsuccessful_file_modification*

justin-stephenson:
#### Description:

- Add ansible remediation for rules below, each ansible playbook is the same for each

  - audit_rules_unsuccessful_file_modification_openat_o_creat
  - audit_rules_unsuccessful_file_modification_open_by_handle_at_o_creat
  - audit_rules_unsuccessful_file_modification_open_o_creat
  - audit_rules_unsuccessful_file_modification_creat
  - audit_rules_unsuccessful_file_modification_openat_o_trunc_write
  - audit_rules_unsuccessful_file_modification_open_by_handle_at_o_trunc_write
  - audit_rules_unsuccessful_file_modification_open_o_trunc_write**

PR #4058, opened 2019-03-06
Add audit_rules_unsuccessful_file_modification_detailed remediation scripts

redhatrises:

PR #4440, opened 2019-06-21
Revert "Add ansible remediation for audit_rules_unsuccessful_file_modification*"

shawndwells:
Reverts ComplianceAsCode/content#4420


jan-cerny:
From #4420:

> @jan-cerny This should be reverted. The ansible remediations fix more than what the XCCDF calls for.

Yes, but the Bash remediations do it as well.

yuumasato:
> This should be reverted. The ansible remediations fix more than what the XCCDF calls for.

We are aware of the issue that it fixes more than one configuration item. This is an interim fix.
The proper fix for this requires checking order of audit rules and inserting audit rules in order if necessary. See also discussion in https://github.com/ComplianceAsCode/content/pull/4058#issuecomment-485770021

shawndwells:
Bump - need to get these changes reverted, especially before next release. Results in the audit rules being added multiple times.

jan-cerny:
@shawndwells 

> Results in the audit rules being added multiple times.

Ansible should ensure it's only once there. I can't reproduce them being added multiple times. What did you run that it resulted in rules added multiple times? Perhaps some other rule conflicts with these playbooks?

jan-cerny:
I have applied again the Ansible playbook for RHEL 8  OSPP profile (`build/ansible/rhel8-playbook-ospp.yml) built from current master (as of HEAD cf9862d2ad56f4d061713943a99c4c0af450c036) and I can see every rule only once in `/etc/audit/rules.d`. Even after I executed the same playbook again it didn't duplicate.

shawndwells:
Works for me. Likely screwed something up on my end. Lets close this.

PR #4443, opened 2019-06-24
Enable configure_crypto_policy set DEFAULT test scenario for RHEL8.

ggbecker:
Enable set DEFAULT configure_crypto_policy test scenario for RHEL8.

### DIFF
--- a/tests/data/group_system/group_software/group_integrity/group_crypto/rule_configure_crypto_policy/policy_default_set.pass.sh
+++ b/tests/data/group_system/group_software/group_integrity/group_crypto/rule_configure_crypto_policy/policy_default_set.pass.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# platform = multi_platform_fedora
-# profiles = xccdf_org.ssgproject.content_profile_standard
+# platform = multi_platform_fedora,Red Hat Enterprise Linux 8
+# profiles = xccdf_org.ssgproject.content_profile_ospp,xccdf_org.ssgproject.content_profile_standard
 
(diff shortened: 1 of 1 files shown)


mildas:
What about other test scenarios for `configure_crypto_policy`?

ggbecker:
> What about other test scenarios for `configure_crypto_policy`?

The other test scenarios are already configured to run for RHEL8, e.g.:
 https://github.com/ComplianceAsCode/content/blob/4aec5060d42009b5f3e9cbbd7ac24f6fa42f16e2/tests/data/group_system/group_software/group_integrity/group_crypto/rule_configure_crypto_policy/symlink_to_wrong_policy.fail.sh#L2

mildas:
@ggbecker Ah, okay.
I have one question - does this test make sense? I mean the test sets crypto policies to `DEFAULT`, but OVAL expects `FIPS` value, so the test fails every time although it should pass (`*.pass.sh`). It would make sense for profile, that wants `DEFAULT`, but I didn't find any and it fails at fedora too.

ggbecker:
> @ggbecker Ah, okay.
> I have one question - does this test make sense? I mean the test sets crypto policies to `DEFAULT`, but OVAL expects `FIPS` value, so the test fails every time although it should pass (`*.pass.sh`). It would make sense for profile, that wants `DEFAULT`, but I didn't find any and it fails at fedora too.

It makes sense only for the `standard` profile which uses default variable for this rule (`DEFAULT`). So we can just remove the OSPP profile from `profiles` header and it'll be ok.

mildas:
Looks good, I will merge it after build passing.

PR #4450, opened 2019-06-25
Fix minlen test scenarios and include RHEL8 platform

ggbecker:
Note: Git refuses to keep the history of renaming files so it is quite inconvenient that it thinks that files got deleted. I'll try to point out the relevant changes by making comments in the PR.


tests/data/group_system/group_accounts/group_accounts-restrictions/group_password_expiration/rule_accounts_password_minlen_login_defs/password_minlen_15.pass.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+#
+# profiles = xccdf_org.ssgproject.content_profile_ospp
+# platform = multi_platform_fedora,Red Hat Enterprise Linux 8
+
+if grep -q "^PASS_MIN_LEN" /etc/login.defs; then
+	sed -i "s/^PASS_MIN_LEN.*/PASS_MIN_LEN 15/" /etc/login.defs
```
ggbecker: here the value changed from 12 -> 15. since the ospp profile is not setting the variable for this rule, it uses the default (15)

jan-cerny: It should be 12 for RHEL8 because RHEL8 OSPP will set 12. See that we set in #4405 var_accounts_password_minlen_login_defs=12

tests/data/group_system/group_accounts/group_accounts-restrictions/group_password_expiration/rule_accounts_password_minlen_login_defs/password_minlen_commented.fail.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+#
+# profiles = xccdf_org.ssgproject.content_profile_ospp
+# platform = multi_platform_fedora,Red Hat Enterprise Linux 8
+
+sed -i "s/.*PASS_MIN_LEN.*/#PASS_MIN_LEN 12/" /etc/login.defs
```
mildas: Just a minor thing - change `12` to `15` here (or more). Because it can happen that the check is not testing comments, but only number after `PASS_MIN_LEN` and in this test scenario it will give ok result (expected fail), but because of the number under limit and not because of `#`.

tests/data/group_system/group_accounts/group_accounts-restrictions/group_password_expiration/rule_accounts_password_minlen_login_defs/password_minlen_15.pass.sh
ggbecker: that means that filename `password_minlen_12.fail.sh` should be reverted to `password_minlen_12.pass.sh`. But test will fail until new RHEL8/OSPP profile gets in place

tests/data/group_system/group_accounts/group_accounts-restrictions/group_password_expiration/rule_accounts_password_minlen_login_defs/password_minlen_commented.fail.sh
ggbecker: well observed, but we can keep it like that because of value `12` in new OSPP profile.

tests/data/group_system/group_accounts/group_accounts-restrictions/group_password_expiration/rule_accounts_password_minlen_login_defs/password_minlen_15.pass.sh
mildas: what about keeping `password_minlen_12.pass.sh` test only for fedora standard profile until OSPP gets in place?

ggbecker: The problem here is that even OSPP (which is the one selected in the `profiles` test scenario header) from fedora doesn't select the variable value explicit, so it is using the default value which is `15` and the test will fail aswell. Not sure what to do here.

mildas: I meant `standard` profile for Fedora, but yea it's hard to say what to do with this one in OSPP. I don't know if failing test scenario is ok for few days.
@jan-cerny do you have any suggestion?

jan-cerny: In general, we can have different test scenarios for different profiles and different platforms. Eg. first scenario for Fedora standard profile that uses value 15 and second scenario for RHEL8 OSPP profile that uses value 12.

Regarding the fact that the tests are currently failing. The most correct solution would be to hold this PR until the RHEL 8 OSPP profile is aligned with the CC requirements and then rebase on top of the master.
However, the tests are not a part of PR gating, so we can say that we have prepared the test scenarios for future and they will start to pass once the OSPP profile is fixed (seems almost like test driven development).

mildas:
After the discussion and comment from Honza, we can say that these test scenarios are for future when OSPP profile is done. So change `password_minlen_12.pass.sh` only for RHEL8 OSPP, because I think that Fedora OSPP profile will remain the same (15 and more -> pass).
@ggbecker Is it okay?

mildas:
Thank you @ggbecker.

For others: `password_minlen_12.pass.sh` test fails until rhel8 ospp will be finished.

PR #4452, opened 2019-06-25
Add nosuid Option to /boot.

adelton:
#### Description:

- Add nosuid Option to /boot.

#### Rationale:

- Users should not be able to execute SUID or SGID binaries from boot partitions.


linux_os/guide/system/permissions/partitions/mount_option_boot_nosuid/rule.yml
```diff
@@ -0,0 +1,24 @@
...
+    The <tt>nosuid</tt> mount option can be used to prevent
+    execution of setuid programs in <tt>/boot</tt>. The SUID and SGID permissions
+    should not be required in these user data directories.
+    {{{ describe_mount(option="nosuid", part="/boot") }}}
+
+rationale: |-
+    The presence of SUID and SGID executables should be tightly controlled. Users
+    should not be able to execute SUID or SGID binaries from boot partitions.
+
+severity: unknown
```
shawndwells: A severity should be set.

linux_os/guide/system/permissions/partitions/mount_option_boot_nosuid/rule.yml
```diff
@@ -0,0 +1,24 @@
...
+
+identifiers:
+    cce@rhel8: 81033-3
```
shawndwells: Should a CCE be assigned for RHEL 6 and 7?

linux_os/guide/system/permissions/partitions/mount_option_boot_nosuid/rule.yml
```diff
@@ -0,0 +1,24 @@
...
+
+references:
```
shawndwells: Can likely nab the NIST references from other partition option checks.

linux_os/guide/system/permissions/partitions/mount_option_boot_nosuid/rule.yml
shawndwells: Also, make sure to commit an updated cce-available file.

linux_os/guide/system/permissions/partitions/mount_option_boot_nosuid/rule.yml
```diff
@@ -0,0 +1,24 @@
+documentation_complete: true
+
+prodtype: rhel8,fedora,ol8
```
shawndwells: What about RHEL 7, maybe RHEL 6 too?

linux_os/guide/system/permissions/partitions/mount_option_boot_nosuid/rule.yml
adelton: @yuumasato, @stevegrubb, what should be the severity for these `mount_option_*` rules? We currently have `unknown` in most of them, with just `mount_option_dev_shm_*` having `medium`.

linux_os/guide/system/permissions/partitions/mount_option_boot_nosuid/rule.yml
adelton: So `nist: CM-7,MP-2` and `nist-csf: PR.IP-1,PR.PT-2,PR.PT-3`?

linux_os/guide/system/permissions/partitions/mount_option_boot_nosuid/rule.yml
adelton: I'm working based on the RHEL 8 kickstart by @stevegrubb. Not sure to what extend this is applicable to older versions. We can always add them later en masse.

linux_os/guide/system/permissions/partitions/mount_option_boot_nosuid/rule.yml
adelton: Mea culpa, fixed in 8bd439d5c5c558db5b5ae8a6d418fb8b02bec0d9 (also rebased on master).

adelton:
> Should this also be enabled in profiles?

It should. However, ol8/profiles/ospp.profile currently only has `mount_option_dev_shm_*`, so only small subset of what we are tracking as things that are already done.

@yuumasato, do you want me to add all the current `mount_option_*` rules there now and then keep adding the new ones one by one, or do you want just the rules created and then added to the profile(s) in one final commit?

linux_os/guide/system/permissions/partitions/mount_option_boot_nosuid/rule.yml
yuumasato: Honestly, I don't know how the `severity` is set.

adelton:
I've now added `nist: CM-7,MP-2` and `nist-csf: PR.IP-1,PR.PT-2,PR.PT-3` references to make the tests pass, and I've added entry to rhel8/templates/csv/mount_options.csv to get OVAL generated. And rebased on master -> 066db6d90e1b8db4a46d73d787524971955bd8f9.

If the references are OK and the rest looks sane, the only thing missing is the severity value.

linux_os/guide/system/permissions/partitions/mount_option_boot_nosuid/rule.yml
```diff
@@ -0,0 +1,26 @@
...
+
+title: 'Add nosuid Option to /boot'
+
+description: |-
+    The <tt>nosuid</tt> mount option can be used to prevent
+    execution of setuid programs in <tt>/boot</tt>. The SUID and SGID permissions
+    should not be required in these user data directories.
```
yuumasato: boot partition shouldn't contain user data.

yuumasato:
I'm not sure about `ospp` reference though.

linux_os/guide/system/permissions/partitions/mount_option_boot_nosuid/rule.yml
adelton: Fixed in 22122d07991cf6d5f74e963c91d12913b9c425f9.

adelton:
> `medium` seems reasonable.

Applied in 22122d07991cf6d5f74e963c91d12913b9c425f9.

adelton:
The `ospp` reference is not for example in linux_os/guide/system/permissions/partitions/mount_option_var_tmp_nosuid/rule.yml either.

We can likely add them en masse when we are adding the rules to the `ospp.profile`s.

yuumasato:
> We can likely add them en masse when we are adding the rules to the `ospp.profile`s.

I agree.

Issue #4448, opened 2019-06-25
SSGTS should error if the tested rule isn't present in the DS


jan-cerny:
#### Description of problem:
When we run test scenarios for a rule we don't verify if the rule is present in the given datastream. We can see error messages in the test suite output, but the messages aren't helpful, because they don't show the root cause of the problem.

#### SCAP Security Guide Version:
current upstream as of HEAD 40597c4be76a6e6d4c263828c8576fca29ac5228

#### Operating System Version:
Fedora 30

#### Steps to Reproduce:

1.  `grep java_jre_configure_crypto_policy build/ssg-rhel8-ds.xml`
2. `./test_suite.py rule --libvirt qemu:///system ssgts_rhel8 --datastream ../build/ssg-rhel8-ds.xml java_jre_configure_crypto_policy`
3.
4.

#### Actual Results:
```
Setting console output to log level INFO
INFO - The base image option has not been specified, choosing libvirt-based test environment.
INFO - Logging into /home/jcerny/scap-security-guide/tests/logs/rule-custom-2019-06-25-1454/test_suite.log

INFO - xccdf_org.ssgproject.content_rule_java_jre_configure_crypto_policy
ERROR - Failed to cleanup file '/home/jcerny/scap-security-guide/tests/logs/rule-custom-2019-06-25-1454/xccdf_org.ssgproject.content_rule_java_jre_configure_crypto_policy-empty_file.fail.sh-initial.html'
ERROR - Script empty_file.fail.sh using profile (all) found issue:
ERROR - Rule xccdf_org.ssgproject.content_rule_java_jre_configure_crypto_policy has not been evaluated! Wrong profile selected in test scenario?
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_java_jre_configure_crypto_policy'.
ERROR - Failed to cleanup file '/home/jcerny/scap-security-guide/tests/logs/rule-custom-2019-06-25-1454/xccdf_org.ssgproject.content_rule_java_jre_configure_crypto_policy-false.fail.sh-initial.html'
ERROR - Script false.fail.sh using profile (all) found issue:
ERROR - Rule xccdf_org.ssgproject.content_rule_java_jre_configure_crypto_policy has not been evaluated! Wrong profile selected in test scenario?
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_java_jre_configure_crypto_policy'.
[...]

PR #4454, opened 2019-06-25
Print an error message when rule isn't found

jan-cerny:
#### Description:
Displays an error message when you try to test a rule that has
a test scenario but isn't part of the supplied SCAP datastream.

#### Rationale:
From the existing output it was hard to figure out what exactly happened.

Fixes: #4448


matejak:
This works great, good job!

PR #4453, opened 2019-06-25
Add nodev Option to /boot.

adelton:
#### Description:

- Add nodev Option to /boot.

#### Rationale:

- The only legitimate location for device files is the /dev directory located on the root partition.

PR #4542, opened 2019-06-26
Add nodev Option to /var.

adelton:
#### Description:

- Add nodev Option to /var.

#### Rationale:

- The only legitimate location for device files is the /dev directory located on the root partition.


adelton:
Similar to https://github.com/ComplianceAsCode/content/pull/4452 or https://github.com/ComplianceAsCode/content/pull/4453, we'd need to know the severity, the references, whether to list older RHEL versions, and if to add this to the `ospp.profile` files right away.

yuumasato:
@adelton References and severity can be set similar to #4452.
As `ospp` profile is being reworked, I suggest to add the rule there later.
And I would say to enlist older RHEL versions and their profiles later, to avoid scope creeping this PR.

shawndwells:
CM-7 for the NIST references.

adelton:
Updated, rebased on master -> 6e11fdb1762fb7e730c6e509f5ec0ee76fca0005.

PR #4544, opened 2019-06-26
[WIP] Enable test suit support for podman executed by non-privileged user

matejak:
When `podman run` is executed by root, the newly created container gets it's IP address, and it listens on port 22. However, when executed by normal user, there is no IP assigned, and the containerized process listens on localhost. So it is possible to use the test suite container running sshd, but it has to listen on unprivileged port, as normal users can't listen on port 22.

This PR aims to extend the test suite, so it is possible to operate on entities that expose SSH service on unprivileged, and therefore non-standard ports.

### DIFF
--- a/tests/ssg_test_suite/test_env.py
+++ b/tests/ssg_test_suite/test_env.py
@@ -54,6 +54,8 @@ class SavedState(object):
 
 
 class TestEnv(object):
+    SSH_PORT = "22"
+
     def __init__(self, scanning_mode):
         self.running_state_base = None
         self.running_state = None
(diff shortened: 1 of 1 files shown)


adelton:
Is there anything else needed in this commit?

matejak:
There is way too many things that are needed.
The whole test suite assumes that it will be able to connect to the target on the port 22, so the code has to be parametrized to allow for different port numbers. There are functions and methods that use `ssh` or `scp`, and they need to get reorganized.

matejak:
Superseded by #4789

Issue #4541, opened 2019-06-26
Adding a non-existent rule to a profile doesn't stop build but results in invalid content


jan-cerny:
#### Description of problem:
If you add a rule which doesn't exist to a profile, the build system builds the content, but the result is invalid.

#### SCAP Security Guide Version:
current upstream master as of head 4edc372ddeaffd1c5667037d0b5b7a05933664ee

#### Operating System Version:
Fedora 30

#### Steps to Reproduce:

1. `echo "    - nonexistent_rule" >> rhel8/profiles/ospp.profile`
2. `./build_product rhel8`
3. `oscap ds sds-validate build/ssg-rhel8-ds.xml`
4. `grep nonexistent_rule build/ssg-rhel8-ds.xml`

#### Actual Results:
Cotent builds but contains bogus elements which cause that validation against XML schemas fail.

```
[jcerny@thinkpad scap-security-guide{master}]$ oscap ds sds-validate build/ssg-rhel8-ds.xml 
File 'build/ssg-rhel8-ds.xml' line 50601: Element '{http://checklists.nist.gov/xccdf/1.2}select', attribute 'idref': 'dangling reference to nonexistent_rule!' is not a valid value of the atomic type 'xs:NCName'.
File 'build/ssg-rhel8-ds.xml' line 50601: Element '{http://checklists.nist.gov/xccdf/1.2}select', attribute 'idref': Warning: No precomputed value available, the value was either invalid or something strange hap
pend.
File 'build/ssg-rhel8-ds.xml' line 51580: Element '{http://checklists.nist.gov/xccdf/1.2}select', attribute 'idref': 'dangling reference to nonexistent_rule!' is not a valid value of the atomic type 'xs:NCName'.
File 'build/ssg-rhel8-ds.xml' line 51580: Element '{http://checklists.nist.gov/xccdf/1.2}select', attribute 'idref': Warning: No precomputed value available, the value was either invalid or something strange hap
pend.
OpenSCAP Error: Invalid SCAP Source Datastream (1.2) content in build/ssg-rhel8-ds.xml. [/builddir/build/BUILD/openscap-1.3.1/src/source/oscap_source.c:346]
```

#### Expected Results:
* Error message during the build: "Rule nonexistent_rule doesn't exist".
* Build terminates.

#### Addition Information/Debugging Steps:
It can happen when you accidentally make a typo in rule IDs when creating a profile.
[...]

PR #4545, opened 2019-06-26
Check if all rules in profile exist

jan-cerny:
#### Description:
Prints an error message and aborts the build if profile selects a rule that doesn't exist or isn't present in the built product. The check is done early in the build process which means users won't have to validate the built content to find mistakes such as typo in rule ID in the profile file.

To enable this I had to remove groups selected in profile. This is OK because selecting a group in a profile doesn't have any effect. All rules are unselected by default and if you don't select the rules explicitly in the profile they aren't evaluated regardless the parent group. In other words, selecting groups doesn't affect selecting rules.

#### Rationale:
We want to catch a wrong rule in profile during the build.

Fixes: #4541

### DIFF
--- a/build-scripts/yaml_to_shorthand.py
+++ b/build-scripts/yaml_to_shorthand.py
@@ -71,7 +71,7 @@ def main():
         profiles = loader.loaded_group.profiles
         for p in profiles:
             p.validate_variables(loader.all_values)
-            p.validate_rules(loader.all_rules)
+            p.validate_rules(loader.all_rules, loader.all_groups)
 
         loader.export_group_to_file(args.output)
     elif args.action == "list-inputs":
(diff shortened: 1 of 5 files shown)


mildas:
Wouldn't it be better to load also groups and check for them (if rule doesn't exist and if group doesn't exist then throw error) than deleting groups from RHEL6 profiles?

jan-cerny:
@mildas But why would anybody select groups in profiles?

matejak:
Profiles should select only rules.
Selecting groups is just noise, as it doesn't have any effect. As groups can grow or shrink uncontrollably, the feature of selecting groups would be potentially quite dangerous.

ggbecker:
> Wouldn't it be better to load also groups and check for them (if rule doesn't exist and if group doesn't exist then throw error) than deleting groups from RHEL6 profiles?

I agree, but with a slightly different approach. Basically add an extra check if the selection matches a group and if so, it throws a different message for example: "You have selected a group instead of a rule. Groups have no effect in the profile and are not allowed to be selected. Please remove before proceeding". I couldn't find any documentation related to this and it might confuse the user.

ggbecker:
LGTM. Tested with groups and non existent rules and it works great. Thanks for the patch. Waiting on the build to merge.

PR #4547, opened 2019-06-26
Disallow kernel profiling by unprivileged users.

adelton:
#### Description:

- Disallow kernel profiling by unprivileged users.

#### Rationale:

- Kernel profiling can reveal sensitive information about kernel behaviour.


yuumasato:
It seems that the value of `2` is the most strict, do the other possible values for kernel.perf_event_paranoid` make sense? Would a policy select these other values? 

If yes, we should make a value out of it.

shawndwells:
> It seems that the value of `2` is the most strict, do the other possible values for kernel.perf_event_paranoid` make sense? Would a policy select these other values?
> 
> If yes, we should make a value out of it.

so others don't have to dig up the link:
https://www.kernel.org/doc/Documentation/sysctl/kernel.txt

`````

==============================================================

perf_event_paranoid:

Controls use of the performance events system by unprivileged
users (without CAP_SYS_ADMIN).  The default value is 2.

 -1: Allow use of (almost) all events by all users
     Ignore mlock limit after perf_event_mlock_kb without CAP_IPC_LOCK
>=0: Disallow ftrace function tracepoint by users without CAP_SYS_ADMIN
     Disallow raw tracepoint access by users without CAP_SYS_ADMIN
>=1: Disallow CPU event access by users without CAP_SYS_ADMIN
>=2: Disallow kernel profiling by users without CAP_SYS_ADMIN

==============================================================

`````

**Comment by shawndwells** [comment]
> It seems that the value of `2` is the most strict, do the other possible values for kernel.perf_event_paranoid` make sense? Would a policy select these other values?
> 
> If yes, we should make a value out of it.

As with the common criteria profiles, even the DoD STIG, the underlying assumption is these baselines are for production machines. Extending that assumption, there should not be a need to perform profiling on production.

Certainly, there will be situations where system administrators will need to put the system into a kind of "debug" mode and backout of these settings. But that should be a specific action vs default state.

File: linux_os/guide/system/permissions/restrictions/sysctl_kernel_perf_event_paranoid/rule.yml [hunk 1]
```diff
@@ -0,0 +1,23 @@
...
+identifiers:
+    cce@rhel8: 81054-9
+
+references:
+    anssi: NT28(R25)
+    ospp: FMT_SMF_EXT.1
+
+{{{ complete_ocil_entry_sysctl_option_value(sysctl="kernel.perf_event_paranoid", value="2") }}}
+
+platform: machine
```
shawndwells: Is this truly only applicable to bare metal and virtual machines?

adelton: What other types of machines are there, for the `platform` value?

adelton:
> so others don't have to dig up the link:
> https://www.kernel.org/doc/Documentation/sysctl/kernel.txt

What is our stance towards referencing the documentation like this from the rules?

linux_os/guide/system/permissions/restrictions/sysctl_kernel_perf_event_paranoid/rule.yml
ggbecker: I think @shawndwells meant applicability to containers. Because even though you are in a container environment, you may have the kernel package installed even though it is not going to be run. Our rules which touch these kernel configuration are already restricting the applicability to `machine`, e.g. https://github.com/ComplianceAsCode/content/blob/af36c2a29e5747f287b44922825c96a2a3992e17/linux_os/guide/system/permissions/restrictions/sysctl_kernel_kexec_load_disabled/rule.yml#L20

If we decide to change the applicability we need to consider other rules as well.

linux_os/guide/system/permissions/restrictions/sysctl_kernel_perf_event_paranoid/rule.yml
```diff
@@ -0,0 +1,23 @@
...
+rationale: |-
+    Kernel profiling can reveal sensitive information about kernel behaviour.
+
+severity: medium
+
+identifiers:
+    cce@rhel8: 81054-9
+
+references:
+    anssi: NT28(R25)
```
yuumasato: This rule actually fulfills `NT28(R23)`.
And since this reference is here, please enable it for RHEL7, including the CSV file, and select it in `rhel7/profiles/anssi_nt28_intermediary.profile`

linux_os/guide/system/permissions/restrictions/sysctl_kernel_perf_event_paranoid/rule.yml
yuumasato: I think this should be kept as `machine` only.
I could not find mention that `kernel.perf_event_paranoid` is safe or namespaced.

adelton:
> Do you plan to update rule description?

Hmm, it sounds like something the templating for `sysctl` could do automatically ... so I don't think we should be putting this to this pull request.

linux_os/guide/system/permissions/restrictions/sysctl_kernel_perf_event_paranoid/rule.yml
adelton: `NT28(R23)` instead of `NT28(R25)`, or `NT28(R23)` in addition to `NT28(R25)`?

yuumasato: `NT28(R23)` instead of `NT28(R25)`.
Sorry I wasn't clear.

adelton:
Fixed `NT28(R23)`, added RHEL 7, rebased on master -> 5b06f7c6cbeee5969ae10d6113dd84bc875a5330.

linux_os/guide/system/permissions/restrictions/sysctl_kernel_perf_event_paranoid/rule.yml
shawndwells: To elaborate, yes, @ggbecker was right. I was asking if this was applicable inside containers. Checking the kernel.org and red hat documentation there seems to be very limited docs on what sysctls are applicable inside containers and which ones are not.

shawndwells:
> > there should not be a need to perform profiling on production.
> 
> This makes sense, so it should be fine to have fixed value of `2`.
> 
> > What is our stance towards referencing the documentation like this from the rules?
> 
> It is OK to reference external documentation. Use the `weblink` macro, it will add an HTML anchor and set the correct namespace for you. Usage example:
> 
> ```
> {{{ weblink(link="https://www.kernel.org/doc/Documentation/sysctl/kernel.txt") }}}
> ```
> 
> @adelton Do you plan to update rule description?

Do the macros also support conditional checking? e.g. ``if rhel use https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/7.6_release_notes/chap-red_hat_enterprise_linux-7.6_release_notes-kernel_parameters_changes, else use kernel.org link`` ?

Issue #4493, opened 2019-06-25
need new rule: unprivileged_bpf_disabled


shawndwells:
#echo "kernel.unprivileged_bpf_disabled = 1" >> $CONFIG

PR #4563, opened 2019-06-27
Disable access to network bpf() syscall from unprivileged processes.

adelton:
#### Description:

- Disable access to network bpf() syscall from unprivileged processes.

#### Rationale:

- Loading and accessing the packet filters programs and maps using the bpf() syscall has the potential of revealing sensitive information about the kernel state.


linux_os/guide/system/permissions/restrictions/sysctl_kernel_unprivileged_bpf_disabled/rule.yml
```diff
@@ -0,0 +1,24 @@
...
+    Loading and accessing the packet filters programs and maps using the bpf()
+    syscall has the potential of revealing sensitive information about the kernel state.
+
+severity: medium
+
+identifiers:
+    cce@rhel8: 82974-7
+
+references:
+    anssi: NT28(R25)
```
jan-cerny: ANSSI NT28 R25 says: 

> It is recommended to load the Yama security module at startup (by example passing the security = yama argument to the kernel) and configure the sysctl kernel.yama.ptrace_scope to a value of at least 1.

I think this rule isn't related to YAMA so this reference should be removed. I haven't found any mention of bpf in ANSSI NT28.

jan-cerny:
Fixes #4493

adelton:
Removed that `anssi: NT28(R25)` reference, rebased on master -> 2f9039139360627995ac7bd01f5294a993fde38c.

adelton:
Rebased on master -> 6f4281133b6d9c504b99a2b8c94d9757a94a0144.

jan-cerny:
@adelton thank you

PR #4565, opened 2019-06-27
readarray fixes

matejak:
* Remove `IFS`-based decomposition in favor of plain `cut`.
* Added -t to `readarray` call: The -t option removes separators from individual array items. Without `-t`, individual array items contain the trailing newline.


jan-cerny:
@openscap-ci test this please

ggbecker:
LGTM. I have tested manually commands (when possible) from before and now and they produce the same result.

Thanks for the patch.

PR #4567, opened 2019-06-27
Added a script for checking schema of generated Ansible playbooks.

matejak:


tests/assert_ansible_schema.py
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env python
+
+import argparse as ap
```
jan-cerny: I think you don't have to rename it.

tests/assert_ansible_schema.py
```diff
@@ -0,0 +1,107 @@
...
+import re
+import sys
+
+import yaml
+
+
+COMMENT_KEYS = set((
```
jan-cerny: superfluous parentheses

tests/assert_ansible_schema.py
```diff
@@ -0,0 +1,107 @@
...
+    "platform",
+    "reboot",
+    "strategy",
+    "complexity",
+    "disruption",
+))
+
+TAGS = set((
```
jan-cerny: superfluous parentheses

tests/assert_ansible_schema.py
```diff
@@ -0,0 +1,107 @@
...
+            .format(fname=fname, err=err)
+        )
+        print(msg, file=sys.stderr)
+        return 1
+    return 0
+
+
+if __name__ == "__main__":
+    parser = make_parser()
+    args = parser.parse_args()
```
jan-cerny: I think we don't need 3 functions for parsing 1 argument.

tests/assert_ansible_schema.py
```diff
@@ -0,0 +1,107 @@
...
+            "A playbook tag {tag} contains @, which is unexpected.".format(tag=tag)
+
+        found = re.search(tag_regex, tag)
+        if found:
+            caught_tags.add(found.group(1))
+
+        if re.search(cce_regex, tag):
+            caught_tags.add("CCE")
+
+    assert "CCE" in caught_tags, "There is no CCE tag in the playbook"
```
jan-cerny: Does every product have CCEs? Isn't it RHEL only?

tests/assert_ansible_schema.py
matejak: Not really, `set` accepts iterable as it's only argument. So it is a set defined using a tuple.

tests/assert_ansible_schema.py
matejak: OK.

tests/assert_ansible_schema.py
matejak: Now there are 3 arguments :-)

tests/assert_ansible_schema.py
matejak: You are right, moreover not all RHEL rules contain them.
As a result, this has been marked as optional.

cmake/SSGCommon.cmake
```diff
@@ -261,6 +261,12 @@ macro(ssg_build_ansible_playbooks PRODUCT)
         NAME "${PRODUCT}-ansible-playbooks-generated-for-all-rules"
         COMMAND env "PYTHONPATH=$ENV{PYTHONPATH}" "${PYTHON_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/tests/ansible_playbooks_generated_for_all_rules.py" --build-dir "${CMAKE_BINARY_DIR}" --product "${PRODUCT}"
     )
+    if("${PRODUCT}" MATCHES "rhel")
+        add_test(
+            NAME "${PRODUCT}-ansible-assert-playbooks-schema"
+            COMMAND sh env "${PYTHON_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/tests/assert_ansible_schema.py" "${CMAKE_BINARY_DIR}/${PRODUCT}/playbooks/all/*"
```
jan-cerny: What is `sh env`?

tests/assert_ansible_schema.py
jan-cerny: I'm sorry, I meant that you should define the set normally.

cmake/SSGCommon.cmake
matejak: If one wants to use wildcards in commands, one needs the shell to expand wildcards.

tests/assert_ansible_schema.py
matejak: Clarification: Writing `my_set = {1, 2, 3}` is the best way to define a set, but it is not available on python<2.7, and we still have python2.6 in our CI.

jan-cerny:
@matejak Thanks, but please next time add a description.

matejak:
Sorry about that, I completely forgot.

PR #4568, opened 2019-06-28
install_vm.py: fix for osinfo-detect not working under sudo/su

matusmarhefka:
#### Description:

The issue in osinfo-detect https://gitlab.com/libosinfo/libosinfo/issues/30
was preventing the script to install RHEL distros under sudo/su. This commit
fixes it and implements simpler logic for adding repositories into the
kickstart.

### DIFF
--- a/tests/install_vm.py
+++ b/tests/install_vm.py
@@ -24,8 +24,8 @@ def parse_args():
     parser.add_argument(
         "--distro",
         dest="distro",
-        default="fedora",
-        help="What distribution to install (URL alias for fedora/centos7)"
+        required=True,
+        help="What distribution to install (URL alias for fedora/rhel7/centos7/rhel8)"
     )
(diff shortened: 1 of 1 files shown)


tests/install_vm.py
```diff
@@ -24,8 +24,8 @@ def parse_args():
     parser.add_argument(
         "--distro",
         dest="distro",
-        default="fedora",
-        help="What distribution to install (URL alias for fedora/centos7)"
+        required=True,
```
ggbecker: please use choices here: https://docs.python.org/3/library/argparse.html#choices

matusmarhefka: Good suggestion, thanks. I have added commit with choices for the `--distro`.

ggbecker:
LGTM. Thanks for the patch.

PR #4570, opened 2019-06-28
Improve template sysctl to support other types of values

yuumasato:
#### Description:

- Extend sysctl templates to support other data types
- Update docs in sysctl_value CSV files 

#### Rationale:
- Most of the sysctl values are integers, but some can be strings.

### DIFF
--- a/shared/templates/create_sysctl.py
+++ b/shared/templates/create_sysctl.py
@@ -1,6 +1,6 @@
 import re
 
-from template_common import FilesGenerator, UnknownTargetError
+from template_common import FilesGenerator, UnknownTargetError, CSVLineError
 
 
 class SysctlGenerator(FilesGenerator):
@@ -44,6 +44,8 @@ class SysctlGenerator(FilesGenerator):
             data_type = "int"
         elif len(serviceinfo) == 3:
             sysctl_var, sysctl_val, data_type = serviceinfo
+        else:
+            raise CSVLineError()
 
         # convert variable name to a format suitable for 'id' tags
         sysctl_var_id = re.sub('[-\.]', '_', sysctl_var)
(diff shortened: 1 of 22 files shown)


shared/templates/create_sysctl.py
```diff
@@ -33,8 +33,19 @@ def get_files_for_id(self, var_id):
...
-        sysctl_var, sysctl_val = serviceinfo
+        # items can be in format
+        # <sysctl_parameter_name, value> or
+        # <sysctl_parameter_name, value, type>
+        try:
+            sysctl_var, sysctl_val = serviceinfo
+            # Default data type for sysctl is int
+            data_type = "int"
+        except ValueError as e:
+            if str(e) == "too many values to unpack (expected 2)":
```
yuumasato: Maybe I'm being overly precautious here.

matejak: I would go for `if len(serviceinfo) == 3`, the examination of the exception error message may change, and four values would have the same message as three.

shared/templates/create_sysctl.py
```diff
@@ -117,4 +130,4 @@ def generate(self, target, serviceinfo):
 
     def csv_format(self):
         return("CSV should contains lines of the format: " +
-               "sysctlvariable,sysctlvalue")
+               "sysctlvariable,sysctlvalue,<datatype>")
```
matejak: Let's use `[,<datetype>]` in the help message - we already have this marker of optional options elsewhere.

shared/templates/create_sysctl.py
yuumasato: Fair enough, updated per suggestion.

yuumasato: Also dealt with exception case, when CSV line is out of the norm

shared/templates/create_sysctl.py
yuumasato: Updated.

matejak:
I am OK with the PR now, but the CI seems to start finding issues.

yuumasato:
A test CSV line was pushed by mistake, it should be fixed now.

matejak:
Thank you for the PR, merging!

Issue #4457, opened 2019-06-25
disable "can" kernel module


shawndwells:
disable "can" kernel module

PR #4572, opened 2019-06-28
Disable CAN Support.

adelton:
#### Description:

- Disable CAN Support.

#### Rationale:

- Disabling CAN protects the system against exploitation of any flaws in its implementation.


adelton:
Not sure if linux_os/guide/system/network/network-uncommon is the correct location, CAN being a protocol for serial lines ...

shawndwells:
Resolves https://github.com/ComplianceAsCode/content/issues/4457

Issue #4455, opened 2019-06-25
blacklist firewire-core kernel module


shawndwells:
need rule to blacklist firewire-core

PR #4573, opened 2019-06-28
Disable IEEE 1394 (FireWire) Support.

adelton:
#### Description:

- Disable IEEE 1394 (FireWire) Support.

#### Rationale:

- Disabling FireWire protects the system against exploitation of any flaws in its implementation.


adelton:
Similar to https://github.com/ComplianceAsCode/content/pull/4572, I'm not sure if linux_os/guide/system/network/network-uncommon is the correct location or if we should be looking for different place for these serial protocols to live.

shawndwells:
> Similar to #4572, I'm not sure if linux_os/guide/system/network/network-uncommon is the correct location or if we should be looking for different place for these serial protocols to live.

FireWire is pretty uncommon these days, so seems like a good place to live. The name ``uncommon`` was specifically chosen over something like ``depricated`` to indicate the protocol/technology may still be supported or maintained, but .... well.... uncommon :)

shawndwells:
Resolves https://github.com/ComplianceAsCode/content/issues/4455

PR #4576, opened 2019-07-01
Remove ansible_playbook_set_hosts function from test suite

matusmarhefka:
The `ansible_playbook_set_hosts` function is no longer needed as
openscap puts `hosts` section into the generated playbooks.


mildas:
LGTM, thank you

PR #4578, opened 2019-07-01
Add profile metadata override in rule mode

yuumasato:
#### Description:

- Reorganize where the changes to test scenario parameters should happen
  - It is unnecessary to have whole `_parse_parameters()` in `class CombinedChecker`

- Add  `--profile` option to Test Suite's `rule mode`
  - Allows tester to override `#profiles` metadata specified in the test scenario
  - Easier to test new profiles using existing test scenarios
  - When debugging a test scenario WRT to a profile, it is easier to just test it for this one profile


jan-cerny:
Thanks for this great improvement.

PR #4583, opened 2019-07-02
Use block style when a string contains a newline

jan-cerny:
By default, the yaml.Dumper dumps multiline strings surrounded by single quotes. If a newline character is a part of the string it is represented by an empty line in a single quoted string. The line breaks in a single quoted string are replaced by space. This assumes the user needs to know YAML specification to understand the behavior of the single quoted string. Also it looks ugly. Instead, we can use a literal
block style, represented by '|' character. In literal blocks, the line breaks are preserved and no extra empty lines are needed.


mildas:
Tried it and it works. Not sure with the `tag:yaml.org,2002:str` tag, but documentation examples uses same tag and doesn't explain what it should be, so looks good to me. Thank you

PR #4587, opened 2019-07-03
Correct SSH StrictModes

jan-cerny:
#### Description:
Fixes a missing 's' in StrictModes option from /etc/ssh/sshd_config.

#### Rationale:
See man sshd_config that it's StrictModes.


mildas:
Thank you

PR #4592, opened 2019-07-04
Fix macro expansion in OVAL by using available process_file_with_macros.

ggbecker:
#### Description:

- Use process_file_with_macros in OVAL expansion so jinja macros can work with OVAL.

#### Rationale:

- Jinja macros can be expanded within OVAL checks.


yuumasato:
Makes sense.

PR #4593, opened 2019-07-04
[WIP] Create jinja macro for lineinfile OVAL check.

ggbecker:
#### Description:

- Introduces jinja macro for OVAL check for line in file.

#### Rationale:

- Promote reuse of OVAL checks by using macros.

--
TODO:

- [x] Change tests which touch OVAL (ctest still fail) - preprocess macros before testing macroed OVAL xml files.
- [x] Include rule title into OVAL description - process the yaml file does not work during OVAL expansion
- [x] Documentation in line
- [x] Handle missing configuration file
- [x] Macro test added or expanded existent test (expanded test_parse_affected test case to expand all OVAL macros)
- [ ] Documentation is in developer guide
- [ ] Incorporate platform high level macro
- [ ] ~~Revert OVAL checks tested (sshd and grub2) because the macro doesn't aim for checking package presence and it can be consider a regression (e.g.: check if sshd is installed). They will be ported later in another PR which already will make use of CPEs.~~
- [ ] ~~change applicability of rules (CPE) because the OVAL macro doesn't handle required packages~~


shared/macros-oval.jinja
```diff
@@ -0,0 +1,64 @@
+{{%- macro oval_lineinfile(msg='', path='', parameter='', value='', default='', application='', rpm_package='') -%}}
```
yuumasato: I believe that the macro `oval_lineinfile` should be more basic, just expand into a `textfilecontent54_test`, `textfilecontent54_object`, and optionally a `textfilecontent54_state`.

yuumasato: The `rpm_package` parameter seems out of place in the macro.

ggbecker:
Observation: data such as: `{{{ rule_id }}}` and `{{{ rule_title }}}` are not available at the moment the macro is expanded. The `rule.yml` file must be processed before the macro expansion occurs.

shared/macros-oval.jinja
```diff
@@ -0,0 +1,64 @@
...
+      <affected family="unix">
+        <platform>multi_platform_all</platform>
+      </affected>
+      <description>{{{ rule_title }}}</description>
+    </metadata>
+    <criteria comment="{{{ application }}} is configured correctly or is not installed"
+    operator="OR">
+      <criteria comment="{{{ application }}} is not installed" operator="AND">
+        <extend_definition comment="{{{ application }}} is not required or requirement is unset"
+        definition_ref="{{{ application }}}_not_required_or_unset" />
```
yuumasato: These `not_required_or_unset` and its friends are very specific for SSH, and should be part of `oval_sshd_config`.
We should not replicate this mechanism to other sets of rules, they are hard to understand and maintain.

This construct is there to allow a requirement which I'm not sure makes sense.
That the policy gives freedom to have `sshd` installed or not, both situations should be pass.
But if `sshd` is installed, it shall be configured securely.

shared/macros-oval.jinja
```diff
@@ -0,0 +1,64 @@
...
+        definition_ref="{{{ application }}}_required_or_unset" />
+        <extend_definition comment="rpm package {{{ rpm_package }}} installed"
+        definition_ref="package_{{{ rpm_package }}}_installed" />
+        {{%- if default %}}
+        <criteria comment="{{{ application }}} is configured or has default value" operator="OR">
+        {{%- endif %}}
+          <criterion comment="Check {{{ parameter }}} in {{{ path }}}"
+          test_ref="test_{{{ rule_id }}}" />
+        {{%- if default %}}
+          <criterion comment="Check for default value of {{{ parameter }}} in {{{ path }}}"
```
yuumasato: The comment is misleading to me, the test is checking that the config file is not overriding the default value.
Not that the default value is set in the config file.

shared/macros-oval.jinja
cipherboy: I largely agree with @yuumasato here.

More generally, I think the OVAL should condition on the status of the file (Does `/etc/ssh/sshd_config` exist at all? -- and conditionally fail based on a jinja-parameter like `missing_fail`?). We'd then make `missing_fail` condition on the status of the RPM package: if `sshd` isn't installed, a missing `/etc/ssh/sshd_config` is fine. If `sshd` is installed, a missing `/etc/ssh/sshd_config` is bad, because that means we don't know where the config for it resides. That seems to be the more robust solution. 

The other problem we have with OVAL is that we have to put everything inside a single `<def-group />` element, but some portions need to be inside the `criteria` sub-element while others can be appended to the end. This makes it hard to define a single `lineinfile`-type macro for it. I think we should have a way of expanding different portions (`criteria`, `test`, `object`, and `state`), and then build higher-level macros over these as we need to.

shared/macros-oval.jinja
cipherboy: I think in an ideal world, `sshd` installation would be a separate rule which would be co-depended on by other rules. If `sshd` is installed, the profile can select the set of required configuration (no root login, GSSAPI enabled, etc). If it isn't installed, the entire set of rules will pass regardless of configuration state.

IMO, this applies to a lot of our rules for service configuration, since not everyone will run all services. Only a small core of system configuration is truly package-availability independent (`/etc/profile` changes which, while not everyone will want, will always exist). It'd be nice to have a macro-based framework for handling this but I think its tough. You end up needing to pass arrays of identifiers I think...

Anyhow, something to think about.

shared/macros-oval.jinja
```diff
@@ -0,0 +1,64 @@
+{{%- macro oval_lineinfile(msg='', path='', parameter='', value='', default='', application='', rpm_package='') -%}}
+<def-group>
+  <definition class="compliance" id="{{{ rule_id }}}" version="1">
+    <metadata>
+      <title>{{{ msg or rule_title }}}</title>
+      <affected family="unix">
+        <platform>multi_platform_all</platform>
```
cipherboy: This gets into the whole "what do we want to macro-ize" issue. IMO, it is wrong to claim `multi_platform_all` for every single OVAL we could possibly generate with this. One example would be Debian-like vs. RHEL-like naming on Apache's httpd server: on some systems this is `/etc/apache2`, on others it is `/etc/httpd`.

IMO, this should be configurable. Perhaps we need a macro just for platform expansion (`all` -> `multi_platform_all`, etc.).

shared/macros-oval.jinja
jan-cerny: Hi, interesting discussion. This construction has bitten us multiple times, last time in https://github.com/ComplianceAsCode/content/pull/4396. My opinion is that the rules like this should only check the configuration option in the configuration file(s) and they shouldn't check if service is installed or package is installed. Instead, we can do it using a CPE check, which will result to "not applicable" if SSH server isn't available.
The current approach has multiple drawbacks:
* gives a false impression that sshd is configured properly when actually sshd isn't present at all
* the HTML report in OpenSCAP older than 1.3.1 can't display the results when `extend_definition` is used, therefore it's hard to debug why the rules passed or failed
* The checks are not in line with the rule description - the prose doesn't mention installing a package or enabling any service.
*  The rule doesn't check a single isolated configuration item but actually 3 "independent" items, even if it has a single CCE assigned.

I suggest that the new macros generate only the textfilecontent test, without any extend_definition elements  and introduce the CPE applicability check in a separate PR.

shared/macros-oval.jinja
mildas: Agree. The best it would be to have macros for basic things - `test`, `object` and `state`. These basic macros could be used in macro for one check and the "one check" macros in whole rule check (e.g. `sshd`).

mildas:
Maybe we don't need to have so robust macro, that it would contain whole OVAL so the result will be one-line OVAL files only with macro. I'd say `def-group` and `metadata` can be outside of macro and macro will include the rest.

ggbecker:
I tried to make some improvements based on the discussion. I also looked into `macros-ansible.jinja` and could use some of the concepts which are already there. I have tested with three rules so far: `sshd_enable_strictmodes`, `sshd_disable_compression`, `grub2_nousb_argument`. It looks promising, but it still misses documentation, fix the ctests, improve reuse of rule text into the OVAL xml file, etc.

Also, I removed the package presence handling for now as @jan-cerny suggested and we can check for the package using applicability (e.g.: CPE dictionaries)

shared/macros-oval.jinja
```diff
@@ -0,0 +1,88 @@
...
+    {{%- if none_exist %}}
+    <ind:pattern operation="pattern match">{{{ prefix_regex }}}{{{ parameter }}}{{{ separator_regex }}}</ind:pattern>
+    {{%- else %}}
+    <ind:pattern operation="pattern match">{{{ prefix_regex }}}{{{ parameter }}}{{{ separator_regex }}}(.*)[\s]*(?:|(?:#.*))?$</ind:pattern>
+    {{%- endif %}}
+    <ind:instance datatype="int">1</ind:instance>
+  </ind:textfilecontent54_object>
+{{%- endmacro %}}
+
+{{%- macro oval_line_in_file_state(value='', exact_match='') -%}}
```
mildas: It would be better to not complicate it and have this macro only with one parameter, e.g. `line_regex`.

shared/macros-oval.jinja
```diff
@@ -0,0 +1,88 @@
...
+  </ind:textfilecontent54_test>
+{{%- endmacro %}}
+
+{{%- macro oval_line_in_file_object(path='', prefix_regex='^[\s]*', parameter='', separator_regex='[\s]+', value='', none_exist='') -%}}
+  <ind:textfilecontent54_object id="obj_{{{ rule_id }}}{{{ none_exist }}}" version="1">
+    <ind:filepath>{{{ path }}}</ind:filepath>
+    {{%- if none_exist %}}
+    <ind:pattern operation="pattern match">{{{ prefix_regex }}}{{{ parameter }}}{{{ separator_regex }}}</ind:pattern>
+    {{%- else %}}
+    <ind:pattern operation="pattern match">{{{ prefix_regex }}}{{{ parameter }}}{{{ separator_regex }}}(.*)[\s]*(?:|(?:#.*))?$</ind:pattern>
```
mildas: Most of the time around separator can be spaces, so add next to `{{{ separator_regex }}}` `\s*` or maybe parameter, so user can choose if he wants them (and default `\s*`).

shared/macros-oval.jinja
ggbecker: I agree, I will incorporate this logic when #4596 gets merged.

shared/macros-oval.jinja
ggbecker: I'm not sure if I get it. But the user can specify the `separator_regex` which means he can do something like `separator_regex='\s*=\s*'`. Wouldn't this be enough to cover this case?

shared/macros-oval.jinja
ggbecker: In this case I would like to avoid having to construct the regex before, maybe the `exact_match` keyword was not the best option here. I'll make a new commit soon to change the approach and we can continue the discussion.

shared/macros-oval.jinja
mildas: Yeah, right. When I saw it for the first time it came to my mind that I'd input `=` as separator and it wouldn't work. But because it's regex there's no problem with it.

It would be bit more convenient, if the user could input only separator and doesn't think about whitespace characters around, but it could cause problem with files, where is forbidden to have whitespaces.

shared/macros-oval.jinja
```diff
@@ -0,0 +1,90 @@
...
+  {{%- endif %}}
+  id="test_{{{ rule_id }}}{{{ none_exist }}}" version="1">
+  <ind:object object_ref="obj_{{{ rule_id }}}{{{ none_exist }}}" />
+  {{%- if not none_exist %}}
+  <ind:state state_ref="state_{{{ rule_id }}}" />
+  {{%- endif %}}
+  </ind:textfilecontent54_test>
+{{%- endmacro %}}
+
+{{%- macro oval_line_in_file_object(path='', prefix_regex='^[\s]*', parameter='', separator_regex='[\s]+', value='', none_exist='') -%}}
```
mildas: `value` isn't used in this macro.

shared/macros-oval.jinja
yuumasato: I think we still need the second parameters.

I see where you are going, the intention is to check whether the subexpression in the matched object should be the only value, or the parameter can have other multiple values.
And `value` can also be a regex, so the goal is to remove the burden from content writer to also have to provide `value` with regex accounting for other extra things.

Having an argument `exact_match` can be confusing. Why would I want to mach exactly with a regex?
I suggest to rename it to `multi_value`, or something similar.

Note: The `<subexpression>`'s operation is `pattern match`, so we are doing pattern matching all the time, :stuck_out_tongue: .

shared/macros-oval.jinja
ggbecker: fixed

ssg/utils.py
```diff
@@ -183,6 +183,16 @@ def read_file_list(path):
...
     return file_contents
 
+def split_string_content(content):
+    """
+    Split the string content and returns as a list.
+    """
+
+    file_contents = content.split("\n")
+    if file_contents[-1] == '':
+        file_contents = file_contents[:-1]
```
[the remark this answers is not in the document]
ggbecker: Will do. Thanks

shared/macros-oval.jinja
ggbecker: Updated to `Check absence of parameter`.

shared/macros-oval.jinja
ggbecker: I changed to multi_value. I didn't understand exactly what the `<subexpression>` means here, maybe I have to dig up in the OVAL standard a bit.

shared/macros-oval.jinja
```diff
@@ -1,4 +1,16 @@
-{{%- macro oval_check_config_file(msg='', path='', prefix_regex='^\s*', parameter='', separator_regex='(?-i)[\s]+', value='', default_pass='', application='', match_whole_word='') -%}}
+{{#
+  High level macro which checks for a particular set of parameter plus value in a configuration file.
+  This function can take eight values:
+    - path (String): Path to the configuration file to be checked.
+    - prefix_regex (String): Regular expression to be used in the beginning of the OVAL text file content check.
+    - parameter (String): The parameter to be checked in the configuration file.
+    - value (String): The value to be checked. This can also be a regular expression (e.g: value1|value2 can match both values).
+    - separator_regex (String): Regular expression to be used as the separator of parameter and value in a configuration file.
```
yuumasato: Should it mention that if white spaces are allowed, the regex should account for it?

shared/macros-oval.jinja
```diff
@@ -1,4 +1,16 @@
...
+    - default_pass (Any String): If set, it allows the check to pass when the default configuration value is acceptable.
```
yuumasato: I suggest: `Implies that the value for the parameter is the default value. If set, the check will pass if the parameter is not set in the configuration file.

yuumasato: Also, `Any String` is strange. I would just say `boolean`.

shared/macros-oval.jinja
```diff
@@ -69,21 +104,46 @@
   </ind:textfilecontent54_object>
 {{%- endmacro %}}
 
-{{%- macro oval_line_in_file_state(value='', match_whole_word='') -%}}
-{{%- if match_whole_word %}}
-{{%- set regex = "^("+value+")[\s]*$" -%}}
-{{%- else %}}
+{{#
+  Macro to check if a expected value can be found in the extracted information of an OVAL object (State definition).
+  This function can take twi values:
```
yuumasato: `two arguments`?

shared/macros-oval.jinja
```diff
@@ -23,14 +35,21 @@
...
   {{%- if none_exist %}}
   {{{ oval_line_in_file_test(path, parameter, none_exist) }}}
   {{{ oval_line_in_file_object(path, prefix_regex, parameter, separator_regex, none_exist) }}}
   {{%- endif %}}
 </def-group>
 {{%- endmacro %}}
 
+{{#
+  Macro to define the criteria of the OVAL check (Criteria definition).
+  This function can take three values:
```
yuumasato: I think they are parameter, not values.

shared/macros-oval.jinja
ggbecker: The problem here is that if I put `boolean` here it might confuse user because they would think that false would disable the option, but it will in fact enable it, because any string would activate the option. I will investigate boolean type in jinja macros

shared/macros-oval.jinja
ggbecker: Ok. Will change.

shared/macros-oval.jinja
ggbecker: I will fix it.

shared/macros-oval.jinja
```diff
@@ -0,0 +1,149 @@
+{{#
+  High level macro which checks for a particular set of parameter plus value in a configuration file.
```
mildas: set of parameter**s**

shared/macros-oval.jinja
ggbecker: I think so. I'll fix it.

shared/macros-oval.jinja
```diff
@@ -0,0 +1,149 @@
...
+    {{%- endif %}}
+    <ind:instance datatype="int">1</ind:instance>
+  </ind:textfilecontent54_object>
+{{%- endmacro %}}
+
+{{#
+  Macro to check if a expected value can be found in the extracted information of an OVAL object (State definition).
+  This function can take twi values:
+    - value (String): The value to be checked. This can also be a regular expression (e.g: value1|value2 can match both values).
+    - multi_value (Any String): If set, multiple values can be present for the parameter (e.g: parameter="value1 value2 value3").
```
mildas: This description is a bit confusing. Meaning is if you set `multi_value` then you can use multiple values in `value` parameter, right? 
e.g. `multi_value (Any String): If set, multiple values can be present for parameter "value" (e.g: value="value1 value2 value3").` would be more accurate.

shared/macros-oval.jinja
```diff
@@ -0,0 +1,149 @@
...
+#}}
+{{%- macro oval_line_in_file_state(value='', multi_value='') -%}}
+{{%- if multi_value %}}
+{{%- set regex = "^.*"+value+".*$" -%}}
+{{%- else %}}
+{{%- set regex = "^("+value+")[\s]*$" -%}}
```
mildas: Are these brackets correct?
Most of the time state is some string without brackets, which was captured in object regex.

shared/macros-oval.jinja
ggbecker: actually I'll change to `checks if a particular combination of parameter and value in a configuration file is set`

shared/macros-oval.jinja
ggbecker: Maybe the concept (`multi_value`) is still confusing... I'll try to come up with something more clear.

shared/macros-oval.jinja
ggbecker: Brackets by definition are either `[` or `]`. Do you mean parenthesis? `(` or `)`

If you mean parenthesis, then it actually doesn't matter because this serve for group expression IIRC, it doesn't mean that it will try to match `(value)`

I can remove if there is some other effect I'm not seeing.

mildas: Yea, I meant paranthesis. If it's intentional then it's ok.

shared/macros-oval.jinja
mildas: Isn't plural still correct phrase for this , `combination of parameters`?

shared/macros-oval.jinja
ggbecker: I changed to actual boolean type and it works, now it is possible to specify `default_pass=true` without double quotes.

shared/macros-oval.jinja
ggbecker: But it is referring to combination which is single so `a combination of parameters is...`. IMO

shared/macros-oval.jinja
yuumasato: I believe that parenthesis in a state don't do anything. If they have no purpose they can as well be removed.

The parenthesis are more important in the object, where what is captured in the groups is provided as `subexpression` for the state.

ggbecker: Ok, I'll remove it. Thanks for the information.

shared/macros-oval.jinja
```diff
@@ -0,0 +1,199 @@
...
+  Macro to check if a expected value can be found in the extracted information of an OVAL object (State definition).
+  This function can take two parameters:
+    - value (String): The value to be checked. This can also be a regular expression (e.g: value1|value2 can match both values).
+    - multi_value (boolean): If set, it means that the parameter can accept multiple values and the expected value must be present in the current list of values.
+#}}
+{{%- macro oval_line_in_file_state(value='', multi_value='') -%}}
+{{%- if multi_value %}}
+{{%- set regex = "^.*"+value+".*$" -%}}
+{{%- else %}}
+{{%- set regex = "^"+value+"[\s]*$" -%}}
```
yuumasato: Sorry for nitpicking this regex again.
Is `[\s]*` needed after the value?
As far as I understand, in line 115:
``` {{%- set regex = prefix_regex+parameter+separator_regex+"(.*)[\s]*(?:|(?:#.*))?$" -%}} ```
there is `(.*)[\s]*`, thus in the subexpression there shouldn't be any trailing white spaces.

Comment (ggbecker) [reply shared/macros-oval.jinja]: It makes sense. I'll remove it. Please don't be shy on pointing out nitpicks in those regular expressions, I am no expert in regex and they will probably need more tweaks

File: shared/macros-oval.jinja [hunk 17]
```diff
@@ -0,0 +1,199 @@
...
+    - path (String): Path to the configuration file to be checked.
+    - prefix_regex (String): Regular expression to be used in the beginning of the OVAL text file content check.
+    - parameter (String): The parameter to be checked in the configuration file.
+    - separator_regex (String): Regular expression to be used as the separator of parameter and value in a configuration file. If spaces are allowed, this should be included in the regular expression.
+    - default_pass (boolean): Implies that the value for the parameter is the default value. If set, the check will pass even if the parameter is not set in the configuration file.
+#}}
+{{%- macro oval_line_in_file_object(path='', prefix_regex='^[\s]*', parameter='', separator_regex='[\s]+', default_pass='') -%}}
+{{%- if default_pass %}}
+{{%- set suffix_id = "_default_pass" -%}}
+{{%- set regex = prefix_regex+parameter+separator_regex -%}}
```
yuumasato: I would add a comment here to explain that there is no need for a capture group when `default_pass=true`, because when `default_pass=true` the test doesn't have a state and it checks for the absence of the `parameter` in config file.

shared/macros-oval.jinja
```diff
@@ -0,0 +1,204 @@
...
+    - parameter (String): The parameter to be checked in the configuration file.
+    - separator_regex (String): Regular expression to be used as the separator of parameter and value in a configuration file. If spaces are allowed, this should be included in the regular expression.
+    - default_pass (boolean): Implies that the value for the parameter is the default value. If set, the check will pass even if the parameter is not set in the configuration file.
+#}}
+{{%- macro oval_line_in_file_object(path='', prefix_regex='^[\s]*', parameter='', separator_regex='[\s]+', default_pass='') -%}}
+{{%- if default_pass %}}
+{{%- set suffix_id = "_default_pass" -%}}
+{{#
+  There is no need for having a regular expression with a capture
+  group "(.*)" when checking if the line in file is not present in
```
yuumasato: Suggestion to rephrase: `when checking if the parameter is not present in`

shared/macros-oval.jinja
```diff
@@ -10,23 +10,36 @@
...
     </metadata>
     {{%- if missing_config_file_fail %}}
     <criteria comment="{{{ application }}} is configured correctly and configuration file exists"
     operator="AND">
     {{%- endif %}}
+    {{%- if application == "sshd" %}}
+    <criteria comment="{{{ application }}} is configured correctly or is not installed"
+    operator="OR">
+        {{{- application_not_required_or_requirement_unset() }}}
+    {{%- else %}}
```
yuumasato: If I would be using macro `oval_check_config_file`, how would I make it check if a package is installed?
Should there be an `extend_definition` in the else clause to check if the package is installed?

shared/macros-oval.jinja
mildas: It's much better. Only one thing - maybe add (to brackets or after `the parameter` word), that with phrase `the parameter` you mean parameter `value` (not parameter `multi_value`). Because user may get from the documentation, that if he wants multiple values, then he should input them in `multi_value`.

ggbecker: This version is outdated. I rephrased this sentence to:

`multi_value (boolean): If set, it means that the parameter can accept multiple values and the expected value must be present in the current list of values.`

shared/macros-oval.jinja
ggbecker: Done.

shared/macros-oval.jinja
ggbecker: The idea was to avoid the OVAL check to test if the package is installed or not and let this be handled by rule applicability (e.g.: CPE dictionaries). Is it really needed?

ggbecker:
PR is in a state that can be merged. Some work still needs to be done when other PRs get merged. The extended definition for sshd configuration rules were kept to not regress, but thei are meant to be removed when it is resolved by rule applicability (e.g.: CPE). Still need to resolve conflicts due to recent changes in master.

shared/macros-oval.jinja
yuumasato: I see, that makes sense.

shared/macros-oval.jinja
```diff
@@ -0,0 +1,236 @@
...
+      <affected family="unix">
+{{#-
+  Use macro oval_affected(products) when it is available
+#}}
+        <platform>multi_platform_rhel</platform>
+        <platform>multi_platform_fedora</platform>
+        <platform>multi_platform_rhv</platform>
+        <platform>multi_platform_ol</platform>
+      </affected>
+      <description>{{{ rule_description }}}</description>
```
yuumasato: After checking the built DataStream, I'm not fond of having rule description here.
It can be too long, and doesn't add much value.

As the OVAL description, shomething like what was in the definition for `grub2_nousb_argument` is more useful:
`Look for 'nousb' argument in the kernel line in /etc/default/grub`

shared/macros-oval.jinja
```diff
@@ -0,0 +1,236 @@
...
+    - prefix_regex (String): Regular expression to be used in the beginning of the OVAL text file content check.
+    - parameter (String): The parameter to be checked in the configuration file.
+    - value (String): The value to be checked. This can also be a regular expression (e.g: value1|value2 can match both values).
+    - separator_regex (String): Regular expression to be used as the separator of parameter and value in a configuration file. If spaces are allowed, this should be included in the regular expression.
+    - default_pass (boolean): Implies that the value for the parameter is the default value. If set, the check will pass even if the parameter is not set in the configuration file.
+    - application (String): The application which the configuration file is being checked. Can be any value and does not affect the OVAL check.
+    - multi_value (boolean): If set, it means that the parameter can accept multiple values and the expected value must be present in the current list of values.
+    - missing_config_file_fail (boolean): If set, the check will fail if the configuration is not existent in the system.
```
yuumasato: Description: `if the configuration file does not exist in the system`.

ssg/build_ovals.py
```diff
@@ -267,13 +271,20 @@ def checks(env_yaml, yaml_path, oval_version, oval_dirs):
     for _dir_path in find_rule_dirs(guide_dir):
         rule_id = get_rule_dir_id(_dir_path)
 
+        rule_path = os.path.join(_dir_path, "rule.yml")
+        rule = Rule.from_yaml(rule_path, env_yaml)
+
+        local_env_yaml['rule_id'] = rule.id_
+        local_env_yaml['rule_title'] = rule.title
+        local_env_yaml['rule_description'] = re.sub('<.*?>', '"', rule.description)
```
yuumasato: Don't forget to remove `rule_description` here.

shared/macros-oval.jinja
```diff
@@ -0,0 +1,236 @@
...
+  This function can take three parameters:
+    - path (String): Path to the configuration file to be checked.
+    - parameter (String): The parameter to be checked in the configuration file.
+    - default_pass (boolean): Implies that the value for the parameter is the default value. If set, the check will pass even if the parameter is not set in the configuration file.
+#}}
+{{%- macro oval_line_in_file_criterion(path='', parameter='', default_pass=false) -%}}
+{{%- set prefix_text = "Check the" -%}}
+{{%- set suffix_id = "" -%}}
+{{%- if default_pass %}}
+{{%- set suffix_id = "_default_pass" -%}}
```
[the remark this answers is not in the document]
ggbecker: Changed to `_default_not_overridden` and tried to deduplicate code by using a constant variable set in the beginning of the macro.

shared/macros-oval.jinja
```diff
@@ -0,0 +1,236 @@
+{{#
+  High level macro which checks if a particular combination of parameter and value in a configuration file is set.
+  This function can take nine parameters:
+    - path (String): Path to the configuration file to be checked.
+    - prefix_regex (String): Regular expression to be used in the beginning of the OVAL text file content check.
+    - parameter (String): The parameter to be checked in the configuration file.
+    - value (String): The value to be checked. This can also be a regular expression (e.g: value1|value2 can match both values).
+    - separator_regex (String): Regular expression to be used as the separator of parameter and value in a configuration file. If spaces are allowed, this should be included in the regular expression.
+    - default_pass (boolean): Implies that the value for the parameter is the default value. If set, the check will pass even if the parameter is not set in the configuration file.
```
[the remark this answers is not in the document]
ggbecker: I changed it to `parameter_missing_pass` because it is literally what it will do, it can be that default values doesn't exist for a particular application, and simply by not setting the configuration is enough. But also, this serve if the application is configured correctly only using default values which means not overriding them in the configuration file. IMO `parameter_missing_pass` fits better.

shared/macros-oval.jinja
ggbecker: I changed to macro: `Look for parameter '{{{ parameter }}}' with value '{{{ value }}}' in '{{{ path }}}'`

shared/macros-oval.jinja
ggbecker: Fixed.

ssg/build_ovals.py
ggbecker: Removed.

ggbecker:
I have rebased with master (ce56dd5c3aa848b06b15cacc37f795ff26e670a0)

linux_os/guide/system/permissions/mounting/grub2_nousb_argument/oval/shared.xml
```diff
@@ -1,34 +1,26 @@
...
     <metadata>
-      <title>Disable Kernel Support for USB via Bootloader Configuration</title>
+      <title>{{{ rule_title }}}</title>
       <affected family="unix">
         <platform>Red Hat Enterprise Linux 7</platform>
         <platform>Red Hat Enterprise Linux 8</platform>
         <platform>multi_platform_fedora</platform>
       </affected>
-      <description>Look for 'nousb' argument in the kernel line in /etc/default/grub</description>
+      <description>{{{ rule_description }}}</description>
```
yuumasato: This should be: `Look for parameter '{{{ parameter }}}' with value '{{{ value }}}' in '{{{ path }}}'`

ggbecker: Thanks for the catch. I'll also update fix the jenkins problem.

yuumasato:
@ggbecker Thank you

PR #4595, opened 2019-07-05
Simplify loading of Jinja macros

cipherboy:
There's no real reason to return a new dictionary instead of extending
the existing dictionary when we're going to end up modifying the
existing dictionary at some point anyways.

At best this helped to not pollute `substitutions_dict` when an exception
occurred, however exceptions from this function are usually passed back
up to CMake and not otherwise handled.

### Rationale

We have a number of Jinja macros already. We also have some code in the `ssg` module which could be helpful. The next PR I'll open introduces select `ssg` functions into Jinja, allowing us to use them from the Jinja macros.

Note that we'll only ever be able to cal into Python from Jinja and not visa-versa.

### DIFF
--- a/ssg/jinja.py
+++ b/ssg/jinja.py
@@ -71,20 +71,19 @@ def _get_jinja_environment(substitutions_dict):
 _get_jinja_environment.env = None
 
 
-def extract_substitutions_dict_from_template(filename, substitutions_dict):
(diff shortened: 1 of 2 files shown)


jan-cerny:
lgtm, thanks

PR #4596, opened 2019-07-05
Jinja: OVAL platforms

cipherboy:
#### Description:

Depends on #4595; will be rebased once that is discussed and merged. Only the last commit in this series is new.

Add support for generating OVAL's `affected` element from one or more prodtype-like identifiers (e.g., `rhel7`, `rhel`, etc.).

#### Rationale

This will help with Jinja-based OVAL generation if we're looking to generate the entire OVAL contents. In #4593, @ggbecker used `multi_platform_all` as the platform; in some cases this could not be correct (the example I [gave](https://github.com/ComplianceAsCode/content/pull/4593/files#r300729945) was Debian-like vs RHEL-like naming of Apache's `httpd` configuration directory). 

Being able to pass prodtypes instead of full product name will keep the macro call short while keeping it unambiguous. 

I tested this with the `sudo_vdsm_nopasswd` OVAL in `linux_os/guide/system/software/sudo/sudo_vdsm_nopasswd/oval/shared.xml`, but I think updating any number of OVAL definitions to use this is kinda out of the scope of this PR. 

Thoughts?


ssg/utils.py
```diff
@@ -40,6 +43,38 @@ def map_name(version):
                        % (version))
 
 
+def to_name(prod):
```
cipherboy: One thought is that we could move this to a new file, `ssg/jinja-macros.py`. Note that we can't actually add these to `products.py` since that'd create a cyclic dependency: `ssg/jinja.py -> ssg/products.py -> ssg/yaml.py -> ssg/jinja.py`

jan-cerny: @cipherboy Could it have a better name? For example `prodtype_to_name`?

cipherboy: @jan-cerny Sure, I'll fix this today if I get time.

ssg/utils.py
```diff
@@ -40,6 +43,38 @@ def map_name(version):
...
+    """
+    for name, prod_type in FULL_NAME_TO_PRODUCT_MAPPING.items():
+        if prod == prod_type:
+            return name
+    if prod in MULTI_PLATFORM_LIST or prod == 'all':
+        return "multi_platform_" + prod
+    raise RuntimeError("Unknown product name: %s" % prod)
+
+
+def to_platform(names):
```
yuumasato: Also, I suggest `name_to_platform`

ssg/jinja.py
```diff
@@ -98,26 +97,29 @@ def process_file(filepath, substitutions_dict):
     return template.render(substitutions_dict)
 
 
+def add_python_functions(substitutions_dict):
+    substitutions_dict['prodtype_to_name'] = to_name
+    substitutions_dict['name_to_platform'] = to_platform
+    substitutions_dict['prodtype_to_platform'] = prodtype_to_platform
```
yuumasato: :open_mouth:

ggbecker:
@cipherboy I took the liberty to fix the requests and rebase with master to resolve conflicts. 

I also incorporated the `oval_affected` macro to the OVAL jinja macros.

ssg/utils.py
ggbecker: Renamed function to `prodtype_to_name`.

ssg/utils.py
ggbecker: Renamed as suggested.

ggbecker:
LGTM. I integrated with OVAL jinja macros and they work perfect.

shared/macros-oval.jinja
```diff
@@ -224,10 +214,9 @@
...
     - multi_value (boolean): If set, it means that the parameter can accept multiple values and the expected value must be present in the current list of values.
     - missing_config_file_fail (boolean): If set, the check will fail if the configuration is not existent in the system.
-    - supported_platforms (String List): List of supported platforms the OVAL will applicable to. (e.g.: ['rhel7', 'rhel8', 'fedora'])
 #}}
 {{%- macro oval_grub_config(parameter='', value='', missing_parameter_pass=false, multi_value=true, missing_config_file_fail=false, supported_platforms=['all']) %}}
-{{{ oval_check_config_file("/etc/default/grub", prefix_regex="^[\s]*", parameter=parameter, separator_regex='=', value=value, missing_parameter_pass=missing_parameter_pass, application="grub", multi_value=multi_value, missing_config_file_fail=missing_config_file_fail, supported_platforms=supported_platforms) }}}
+{{{ oval_check_config_file("/etc/default/grub", prefix_regex="^[\s]*", parameter=parameter, separator_regex='=', value=value, missing_parameter_pass=missing_parameter_pass, application="grub", multi_value=multi_value, missing_config_file_fail=missing_config_file_fails) }}}
```
yuumasato: extra `s` in `missing_config_file_fails`

ggbecker: Fixed. Thanks for the catch.

ggbecker:
@openscap-ci test this please

ggbecker:
@cipherboy I'm merging this so the `supported_platforms` option in the OVAL macros is not misused by others (because it has been removed), as this PR introduces the capability of expanding `prodtypes` into OVAL.

cipherboy:
Cool, thanks for the updates!

PR #4597, opened 2019-07-05
Introduce Bash lineinfile equivalent

cipherboy:
Bash remediations lack a way of doing the exact same task as the Ansible
`lineinfile` macro; replace_or_append is focused on trying to do
everything automatically instead of doing small tasks. The new
`lineinfile` does the following:

 - Separates the regex to match against from the new line to insert
 - Supports insertion at a particular place
 - Supports case-insensitive comparisons
 - Creates high-level wrappers over the `lineinfile` function

This can progressively replace replace_or_append as time permits.

No Jinja was harmed in the making of these functions.. :-) 

### DIFF
--- a/shared/bash_remediation_functions/lineinfile.sh
+++ b/shared/bash_remediation_functions/lineinfile.sh
@@ -1,0 +1,132 @@
+# Functions which minic Ansible's lineinfile module.
+# dependencies: die.sh
+
+# Internal helper function to remove a line, if it is present.
+function lineinfile_absent() {
+    local path="$1"
+    local regex="$2"
+    local insensitive="${8:-true}"
+
+    if [ "$insensitive" == "true" ]; then
+        LC_ALL=C sed "/$regex/Id" "$path"
+    else
+        LC_ALL=C sed "/$regex/d" "$path"
+    fi
+}
+
+# Internal helper function to add a line at the desire location.
+function lineinfile_present() {
+    local path="$1"
+    local regex="$2"
+    local line="$3"
+    local insert_after="$6"
+    local insert_before="$7"
+    local insensitive="${8:-true}"
+
+    local grep_args=("-q" "-m" "1")
+    if [ "$insensitive" == "true" ]; then
+        grep_args+=("-i")
+    fi
+
+    if [ -z "$regex" ] || ! LC_ALL=C grep "${grep_args[@]}" "$regex" "$path"; then
+        if [ -z "$insert_after" ] && [ -z "$insert_before" ] || [ "$insert_after" == "EOF" ]; then
+            # Insert at the end of the file
+            cp "$path" "$path.bak"
+            printf '%s\n' "$line" >> "$path"
+        elif [ "$insert_before" == "BOF" ]; then
+             # Insert at the beginning of the file
+             cp "$path" "$path.bak"
+             printf '%s\n' "$line" > "$path"
+             cat "$path.bak" >> "$path"
+         elif [ -n "$insert_after" ]; then
+             # Insert after the line matching the regex $insert_after.
+             cp "$path" "$path.bak"
+             local line_number="$(LC_ALL=C grep -n "$insert_after" "$path.bak" | LC_ALL=C sed 's/:.*//g')"
+             if [ -z "$line_number" ]; then
+                 # We used insert_after, but there was no match: insert at
+                 # the end of the file.
+                 printf '%s\n' "$line" >> "$path"
+             else
+                 head -n "$(( line_number ))" "$path.bak" > "$path"
+                 printf '%s\n' "$line" >> "$path"
+                 tail -n "+$(( line_number + 1 ))" "$path.bak" >> "$path"
+             fi
+
+             # Clean up after ourselves.
+             rm "$path.bak"
+         elif [ -n "$insert_before" ]; then
+             # Insert before the line matching the regex $insert_before.
+              cp "$path" "$path.bak"
+             local line_number="$(LC_ALL=C grep -n "$insert_before" "$path.bak" | LC_ALL=C sed 's/:.*//g')"
+             if [ -z "$line_number" ]; then
+                 # We used insert_after, but there was no match: insert at
+                 # the end of the file.
+                 printf '%s\n' "$line" >> "$path"
+             else
+                 head -n "$(( line_number - 1 ))" "$path.bak" > "$path"
+                 printf '%s\n' "$line" >> "$path"
+                 tail -n "+$(( line_number ))" "$path.bak" >> "$path"
+             fi
+
+             # Clean up after ourselves.
+             rm "$path.bak"
+         fi
+     fi
+}
+
+function lineinfile() {
+    local path="$1"
+    local regex="$2"
+    local line="$3"
+    local create="${4:-false}"
+    local state="${5:-present}"
+    local insert_after="$6"
+    local insert_before="$7"
+    local insensitive="${8:-true}"
+
+    [ ! -e "$path" ] && [ "$create" != "true" ] && die "Path '$path' wasn't found on this system and not creating. Refusing to continue."
+    [ ! -e "$path" ] && [ "$create" == "true" ] && touch "$path"
+
+    if [ "$state" == "absent" ]; then
+        lineinfile_absent "$path" "$regex" "$insensitive"
+    elif [ "$state" == "present" ]; then
+        lineinfile_present "$path" "$regex" "$line" "$insert_after" "$insert_before" "$insensitive"
+    fi
+}
+
+function only_lineinfile() {
+    local path="$1"
+    local line_regex="$2"
+    local new_line="$3"
+    local create="$4"
+    local insert_after="$5"
+    local insert_before="$6"
+    local insensitive="$7"
+
+    lineinfile "$path" "$line_regex" '' 'false' 'absent' '' '' "$insensitive"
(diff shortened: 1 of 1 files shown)


cipherboy:
I'm not sure why this is failing; the error is cryptic to me. I've introduced two new files, but that's it.

```
cd /home/jenkins/workspace/scap-security-guide-pull-requests/label/fedora/build/rhel6 && env PYTHONPATH=/home/jenkins/workspace/scap-security-guide-pull-requests/label/fedora /bin/python3 /home/jenkins/workspace/scap-security-guide-pull-requests/label/fedora/build-scripts/generate_fixes_xml.py --remediation_type bash --build_dir /home/jenkins/workspace/scap-security-guide-pull-requests/label/fedora/build --output /home/jenkins/workspace/scap-security-guide-pull-requests/label/fedora/build/rhel6/bash-fixes.xml /home/jenkins/workspace/scap-security-guide-pull-requests/label/fedora/build/rhel6/fixes/bash
Error performing XCCDF <sub> substitution for function variables in accounts_logon_fail_delay fix. Exiting...
[304/648] [rhel6-content] generating xccdf-unlinked-resolved.xml
```

That uses the old `replace_or_append`, but I've not touched that (yet). Thoughts @jan-cerny, @matejak, or @yuumasato?

mildas:
@cipherboy at `ssg/build_remediations.py` is being checked if bash remediation function substitution happened correctly. The problem is it checks string which contains `# Set variables` and word `variables` wasn't substituted because it's a comment and it fails.

The easiest solution for now is renaming `variables.sh` file (the check looks for substings, so filename `var.sh` etc. also won't work). But I will take a look at the build for fixing this problem.

cipherboy:
@mildas Ahhh cool. If I get time today, I'll fix this up. :)

shared/bash_remediation_functions/lineinfile.sh
```diff
@@ -0,0 +1,106 @@
+# Functions which minic Ansible's lineinfile module.
+# dependencies: die.sh, variables.sh
+
+function lineinfile() {
+    local path="$1"
+    local regex="$2"
+    local new_line="$3"
```
matejak: What is this argument for? The Ansible module doesn't have it.

shared/bash_remediation_functions/lineinfile.sh
```diff
@@ -0,0 +1,106 @@
...
+    local create="$(normalize_boolean "$4" "false")"
+    local state="$(normalize_state "$5" "present")"
+    local insert_after="$6"
+    local insert_before="$7"
+    local insensitive="$(normalize_boolean "$8" "true")"
+
+    [ ! -e "$path" ] && [ "$create" != "true" ] && die "Path '$path' wasn't found on this system and not creating. Refusing to continue."
+    [ ! -e "$path" ] && [ "$create" == "true" ] && touch "$path"
+
+    if [ "$state" == "absent" ]; then
```
matejak: The complexity of this function is too big. I guess that we need `lineinfile_present`, `lineinfile_absent` etc. We will need a couple of small functions that are unit-testable e.g. using [BATS](https://github.com/bats-core/bats-core).

shared/bash_remediation_functions/variables.sh
```diff
@@ -0,0 +1,28 @@
...
+  if [ "$value" == "true" ] || [ "$value" == "yes" ]; then
+    echo "true"
+  elif [ "$value" == "false" ] || [ "$value" == "no" ]; then
+    echo "false"
+  else
+    echo "$default"
+  fi
+}
+
+function normalize_state() {
```
matejak: I think that as we are in 100% control of usage of the function, why not to enforce its arguments instead of normalizing them?

shared/bash_remediation_functions/lineinfile.sh
```diff
@@ -0,0 +1,106 @@
...
+        [ -n "$new_line" ] && echo "Ignoring parameter new_line in absent mode..."
+        echo "Saving $path to $path.bak and removing line defined by the regex '$regex'..."
+        if [ "$insensitive" == "true" ]; then
+            LC_ALL=C sed --in-place=.bak "/$regex/Id" "$path"
+        else
+            LC_ALL=C sed --in-place=.bak "/$regex/d" "$path"
```
yuumasato: For discussion: Does it make sense to keep the `.bak` files? What is expected of it?
If multiple rules touch the same file, the `.bak` file contain all previous changes.

shared/bash_remediation_functions/lineinfile.sh
yuumasato: @matejak And you mean to have `lineinfile` macro use `lineinfile_present`and `lineinfile_absent`?
This makes sense to me.

yuumasato:
Since #4598 is merged.
@openscap-ci test this please

shared/bash_remediation_functions/lineinfile.sh
matejak: I would be OK to have the `linenfile` function calling other functions - my main objection was against complex functions.
What you propose is one step further - IIUC the macro would use and expand to only a subset of the bash code. That's of course even better.

shared/bash_remediation_functions/lineinfile.sh
yuumasato: @matejak I'm confused, what argument are you talking about? `new_line` is the actual line want to there.

shared/bash_remediation_functions/variables.sh
yuumasato: @cipherboy What are you worried with that requires these normalizing functions?

shared/bash_remediation_functions/lineinfile.sh
matejak: Yes, I meant the `new_line` argument. I thought that it is the preferred `\n` form :-) Therefore, I suggest renaming that to `line`, which is how it is called in the Ansible module.

shared/bash_remediation_functions/lineinfile.sh
```diff
@@ -0,0 +1,106 @@
...
+        fi
+        if [ -z "$regex" ] || ! LC_ALL=C grep "${grep_args[@]}" "$regex" "$path"; then
+            if [ -z "$insert_after" ] && [ -z "$insert_before" ] || [ "$insert_after" == "EOF" ]; then
+                # Insert at the end of the file
+                cp "$path" "$path.bak"
+                echo "$new_line" >> "$path"
+            elif [ "$insert_before" == "BOF" ]; then
+                # Insert at the beginning of the file
+                cp "$path" "$path.bak"
+                echo "$new_line" > "$path"
```
matejak: This is a nitpick, but `printf '%s\n' "$new_line"` is [slightly better](https://unix.stackexchange.com/a/65819/260749).

shared/bash_remediation_functions/lineinfile.sh
yuumasato: Oh, I see, I agree.

shared/bash_remediation_functions/variables.sh
cipherboy: Sorry folks. I'm on PTO and likely won't have time to address these PR comments until Friday (I go back to work officially on Monday). :/ 

I was trying to make this more "Ansible"-like to make writing Bash remediations more friendly to non-programmers (or even those not familiar with bash). 

IIRC, Ansible handles several variations of the Booleans (`yes`/`no`/`true`/`false`/`True`/`False`/...). These differences aren't caught at compile time (in Bash remediations) because we don't execute them. It would be up to PR author to discover how our module works and ensure they use it correctly, and up to the PR reviewers that they double check all arguments are correct &c.. ...and potentially run the test suite, when present for a rule, to ensure the remediation works as expected.

Seems like we could be "friendly" (for new contributors) and accept more options rather than fewer. Especially for high-level helper functions that we're hoping for new contributors to use. My 2c.

shared/bash_remediation_functions/lineinfile.sh
cipherboy: Ah right, we had this discussion last time I went modifying bash... :-) I'll change this when I update the PR.

shared/bash_remediation_functions/lineinfile.sh
cipherboy: Ah, so I do agree too. Must've been misremembering the docs. Will fix.

cipherboy:
I'll work on adding BATS tests in the next round of updates. @yuumasato / @matejak -- does this address your feedback?

shared/bash_remediation_functions/lineinfile.sh
cipherboy: IMO, I wouldn't turn this into an Ansible macro. I do like splitting it out into functions. I could make the `present` cases into separate functions as well if you'd like. 

Most configuration files will use both (`_absent` and `_present`) anyways: you'll usually want to remove conflicting values before adding the desired value.

yuumasato: Looks good to me.

matejak:
I am fine with the current state. I don't see any other issues by @yuumasato that are in the strict scope of this PR. Therefore, I am merging this.
Thank you very much for this advanced PR!

shared/bash_remediation_functions/variables.sh
matejak: I now understand the motivation, and I am still in favor of simple strict interface vs "friendly" interface. To some extent, this is a matter of taste, especially normalizing those booleans. However, present/absent are pretty clear.

PR #4598, opened 2019-07-08
Remove comments before performing sanity check for substituted bash remediations

mildas:
#### Description:
When bash remediation contains comment with string, that is same as name of file from `shared/bash_remediations_functions` folder, then sanity check during build fails.

E.g. fail happened during build of #4597 because there was file with name `variables.sh` and one of bash remediations contains string with `# Set variables`.


jan-cerny:
@mildas thanks

PR #4599, opened 2019-07-09
Add remaining rules from CC to OSPP

jan-cerny:
#### Description:
Adds remaining rules from CC to OSPP.

#### Rationale:

The rules added in this patch are already present in the project.


yuumasato:
Thank you, @jan-cerny!

PR #4601, opened 2019-07-10
Deduplicate profile files.

matejak:
Removed comments in .profile files that contained information that was part of rule definition, and thus redundant. Those comments cause concerns e.g. when deduplicating CCEs, and they made addition of new rules to the profile a more laborious job.

@shawndwells are you OK with removing this, or is there any major use case to see those CCEs? If it is the latter case, perhaps some script could do the trick without duplicating information in our content.

### DIFF
--- a/rhel8/profiles/ospp.profile
+++ b/rhel8/profiles/ospp.profile
@@ -13,11 +13,11 @@ selections:
     ## SELinux Configuration
     #################################################################
 
-    ## RHEL 8 CCE-80869-1: Ensure SELinux is Enforcing
+    ## Ensure SELinux is Enforcing
     - var_selinux_state=enforcing
     - selinux_state
 
-    ## RHEL 8 CCE-80868-3: Configure SELinux Policy
+    ## Configure SELinux Policy
     - var_selinux_policy_name=targeted
     - selinux_policytype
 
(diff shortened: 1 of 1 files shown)


shawndwells:
> Removed comments in .profile files that contained information that was part of rule definition, and thus redundant. Those comments cause concerns e.g. when deduplicating CCEs, and they made addition of new rules to the profile a more laborious job.
> 
> @shawndwells are you OK with removing this, or is there any major use case to see those CCEs? If it is the latter case, perhaps some script could do the trick without duplicating information in our content.

Inclusion of the CCEs was probably overkill. I was using them to also make PRs to the ospp kickstart file from sgrubb (to self document the kickstart file). 

To help self-document the SCAP profile, having the rule title would be nice but the CCEs aren't particularly needed. However once set they should never change, so having them doesn't seem to introduce much burden. 

If asking for clear(er) thoughts.... no objection to dropping CCE, but the rule title should remain as a comment. IMHO we need to get better about inline developer documentation.

jan-cerny:
@matejak Let's keep the rule titles for now, sometimes they're helpful, but in ideal world the rule titles should be self-explaining. In future, the comments should be more explaining why we included the rule in profile instead of what they check. Something similar to ANSSI in #4615 . Please drop the CCEs for the profile yaml, they're really redundant information.

matejak:
Thank you for your feedback, I have modified the PR so it removes only those CCEs.

yuumasato:
This is waiting for a non-existent job in Jenkins to finish.
@openscap-ci test this please

PR #4602, opened 2019-07-10
Fixing No newline at end of file, introduced by 38fe5cf50fa6cc73a700fb779c7983845415b79d.

adelton:
#### Description:

- Fixing No newline at end of file, introduced by 38fe5cf50fa6cc73a700fb779c7983845415b79d.

#### Rationale:

- Files without the last newline are hard to work with, noise is generated by git tools as well.

PR #4603, opened 2019-07-10
Disable Transparent Inter Process Communication (TIPC) Support.

adelton:
#### Description:

- Disable Transparent Inter Process Communication (TIPC) Support.

#### Rationale:

- Disabling TIPC protects the system against exploitation of any flaws in its implementation.


adelton:
Note that this is on top of https://github.com/ComplianceAsCode/content/pull/4602.

adelton:
If the machine is a node in a cluster, this TIPC MAY be enabled. Should we make this optional, without remediation?

yuumasato:
> If the machine is a node in a cluster, this TIPC MAY be enabled. Should we make this optional, without remediation?

@adelton That is a good question, I guess it will depend if a node in a cluster is part of the Target Of Evaluation.

adelton:
> it will depend if a node in a cluster is part of the Target Of Evaluation.

When / how will we know if it's the case?

shawndwells:
> > it will depend if a node in a cluster is part of the Target Of Evaluation.
> 
> When / how will we know if it's the case?

IIRC this is being included mostly for Common Criteria profile. If that's the case, the Common Criteria profile is for a "general purpose operating system." Specialty deployments like HPC or containers or hypervisors are reasonably expected to change certain settings. 

Suggest adding a ``warning`` noting that under the Common Criteria evaluation, TIPC is allowed under certain conditions and detail what those conditions are.

adelton:
I've added the warning in e427b77f209b7accb993e596305ad38e15887f26. Also rebased on master.

PR #4606, opened 2019-07-11
Improve OVAL for OpenSSH version check

jan-cerny:
#### Description:
The content for RPM-based distributions shouldn't contain dpkginfo test because on these systems OpenSCAP is typically shipped without dpkginfo probe and produces warning "OVAL object not supported". Using Jinja macros, we can include only the rpminfo test in SCAP content for RPM-based distros or include only dpkginfo test in SCAP content for DEB-based distros. This also simplifies the resulting OVAL definition.

#### Rationale:
Related to https://bugzilla.redhat.com/show_bug.cgi?id=1640889


jan-cerny:
@openscap-ci test this please

yuumasato:
Nice catch!

Issue #4272, opened 2019-04-11
Beginners Starting Point?


kclinden:
#### Description of problem:
I have never written SCAP content before, and am looking at how to get started. I would like to write SCAP content to test compliance on Photon OS against DISA SRGs. I have been all over the wiki pages, but I am still not sure how to get started. The main page makes it look super easy for writing OVAL and XCCDF files using YAML, but I am not sure where to build those yaml files, and how to run them through the framework to generate a XCCDF DS files. I would like to just start with a basic check to validate that root logon is not permitted. 

What am I missing?

#### SCAP Security Guide Version:
Latest

#### Operating System Version:
Photon OS 2.0 and 3.0

yuumasato:
> I have never written SCAP content before, and am looking at how to get started. 
Hello, that is great!

> I would like to write SCAP content to test compliance on Photon OS against DISA SRGs. 

You will need to create a new product for Photon OS. This should guide you: https://github.com/ComplianceAsCode/content/tree/master/example
If anything is not clear, let us know, or suggest changes, :)

> I have been all over the wiki pages, but I am still not sure how to get started. The main page makes it look super easy for writing OVAL and XCCDF files using YAML, but I am not sure where to build those yaml files, and how to run them through the framework to generate a XCCDF DS files.

Have you checked the developer guide? It should give some more insight: https://github.com/ComplianceAsCode/content/blob/master/docs/manual/developer_guide.adoc

> I would like to just start with a basic check to validate that root logon is not permitted.

If the rules already exisits, which is the case, you'd basically create a profile, add the rules you want there, and build.

In the example [profile](https://github.com/ComplianceAsCode/content/blob/master/example/profiles/example.profile#L23), there is already a rule for SSH root login. If you are thinking about console root login, this would be the [rule](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/accounts/accounts-restrictions/root_logins/no_direct_root_logins).

kclinden:
I am going through the Devleoper Guide, and I am hung up on step 3.3. I downloaded the tar from the releases page per step 3.2. When I unpack the file there is no build directory. I see there is one in the git repo if I were to take the clone repo. Am I missing something with the release route, or is the doc msising something?

kclinden:
So I am using the repo clone route, and when I try to do a build on rhel7 I am getting errors that WARNING: OVAL check `'some check' was not found, removing <check-content> element from the xccdf file.`

[...]

PR #4610, opened 2019-07-11
New product documentation

ggbecker:
Introduce documentation on how to create a new product in the project.

Also contains simple bash script to help creating the directory structure and some of the needed files.

Relates to #4272

### DIFF
--- a/example/README.md
+++ b/example/README.md
@@ -23,7 +23,7 @@ testing purposes.
 
 Follow these steps to create a new product in the project:
 
-1. Create a new folder in the root directory which will holds the files related to your new product. To illustrate the process we will use the name `custom6` which basically means that the product is called `custom` and the major version is `6`.
+1. Create a new folder in the root directory which will hold the files related to your new product. To illustrate the process we will use the name `custom6` which basically means that the product is called `custom` and the major version is `6`.
 <pre>
 cd content
 mkdir custom6
@@ -48,7 +48,7 @@ message(STATUS "JBoss EAP 6: ${SSG_PRODUCT_EAP6}")
 if (SSG_PRODUCT_DEBIAN8)
     add_subdirectory("debian8")
 endif()
-<b>if (SSG_PRODUCT_CUSTOM6
+<b>if (SSG_PRODUCT_CUSTOM6)
       add_subdirectory("custom6")
 endif()</b>
 if (SSG_PRODUCT_EAP6)
--- a/example/README.md
+++ b/example/README.md
@@ -10,7 +10,7 @@ This product was introduced to show a few aspects:
 To build this system, follow the steps for building regularly, but
 pass `-DSSG_PRODUCT_EXAMPLE=ON` to the CMake command:
 
-    cd scap-security-guide/build
+    cd content/build
     cmake -DSSG_PRODUCT_EXAMPLE=ON ..
     make example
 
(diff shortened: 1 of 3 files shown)


example/README.md
```diff
@@ -10,11 +10,49 @@ This product was introduced to show a few aspects:
...
 This will build only the example profile; this can then be used
 with [OpenSCAP](https://github.com/OpenSCAP/openscap) or
 [scap-workbench](https://github.com/OpenSCAP/scap-workbench) for
 testing purposes.
+
+## Creating a new Product
+
+Follow these steps to create a new product in the project:
+
+1. Create a new folder in the root directory which will holds the files related to your new product. To illustrate the process we will use the name `custom6` which basically means that the product is called `custom` and the major version is `6`.
```
jan-cerny: holds -> hold

example/README.md
```diff
@@ -10,11 +10,49 @@ This product was introduced to show a few aspects:
...
+message(STATUS "Custom 6: ${SSG_PRODUCT_CUSTOM6}")
+message(STATUS "JBoss EAP 6: ${SSG_PRODUCT_EAP6}")
+...
+```
+```
+...
+if (SSG_PRODUCT_DEBIAN8)
+    add_subdirectory("debian8")
+endif()
+if (SSG_PRODUCT_CUSTOM6
```
jan-cerny: missing )

example/README.md
```diff
@@ -10,11 +10,49 @@ This product was introduced to show a few aspects:
...
+      add_subdirectory("custom6")
+endif()
+if (SSG_PRODUCT_EAP6)
+    add_subdirectory("eap6")
+endif()
+...
+```
```
jan-cerny: they also should add it in `./build_product`

example/README.md
```diff
@@ -10,11 +10,49 @@ This product was introduced to show a few aspects:
...
+1. Create a new folder in the root directory which will holds the files related to your new product. To illustrate the process we will use the name `custom6` which basically means that the product is called `custom` and the major version is `6`.
+```
+cd content
+mkdir custom6
+```
+2. Add the product to [CMakeLists.txt](../CMakeLists.txt) by adding the following lines:
+```
+...
+option(SSG_PRODUCT_DEBIAN8 "If enabled, the Debian 8 SCAP content will be built" TRUE)
+option(SSG_PRODUCT_CUSTOM6 "If enabled, the Custom 6 SCAP content will be built" TRUE)
```
jan-cerny: I suggest doing this in a "diff"-like format so that it will obvious they will add only the line with custom

ggbecker: I have changed to make the addition bold instead. What do you think about?

I didn't find an easy way to make it "diff"-like, do have any suggestions? I would find painful to put minus (-) and plus (+) manually.

example/README.md
ggbecker: I'll add this instruction as well. Thanks.

example/README.md
ggbecker: fixed

example/README.md
ggbecker: fixed

example/README.md
ggbecker: Assuming that setting text to `bold` is enough. Resolving conversation.

example/README.md
ggbecker: Added.

ggbecker:
I have addressed review requests. But the documentation is still not complete. Further work needs to be done.

ggbecker:
Added four more items to the documentation. Still in progress.

ggbecker:
More sections added. I looks that only few items are missing now, mainly in the `ssgconstants.py` file.

ggbecker:
- Added remaining sections for xslt transformation files
- Added script which create every file needed for a new product but files which need manual editing.

It is ready to review.

Note: the script is there in a best effort case, it lacks documentation on how to use it but it might help. Maybe it can evolve to something more useful.

jan-cerny:
@ggbecker This is a great contribution!

PR #4612, opened 2019-07-11
Drop ccc profile

yuumasato:
#### Description:

- Remove `ccc.profile`, as all its rules were integrated into `ospp.profile`.
- Actually not all, but:
  - `dconf_gnome_banner_enabled` - dconf rules should not be part of it
  - `dconf_gnome_login_banner_text` - dconf rules should not be part of it
  - `securetty_root_login_console_only` - [`securetty` is disabled by default in RHEL8] (https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/considerations_in_adopting_rhel_8/index#securetty_security)
    - being added into `ospp` by this PR.
  - `sshd_disable_rhosts_rsa` - this setting is deprecated in `openssh-server` > 7.4, RHEL8 has 7.8
  - `sshd_enable_warning_banner` - being added into `ospp` by this PR

#### Rationale:

- `ccc.profile` is incorporated in `ospp.profile`.


shawndwells:
> * 
> * `securetty_root_login_console_only` - [`securetty` is disabled by default in RHEL8](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/considerations_in_adopting_rhel_8/index#securetty_security)

Can it be re-enabled? If so, a check may still be needed.

yuumasato:
> Can it be re-enabled? If so, a check may still be needed.

It can be enabled just by editing a few files.
I have added the rule to the `ospp` profile, along with a few tests for it.
I wanted to make sure the check returned `pass` when file was missing.

shawndwells:
Thanks, @yuumasato

PR #4614, opened 2019-07-12
new rule timer_dnf-automatic_enabled

### LINKED ISSUE #4478: new rule: yum dnf-automatic.timer
#systemctl enable --now dnf-automatic.timer

jan-cerny:
#### Description:
This rule checks whether `dnf-automatic.timer` is enabled, which means the new updates are automatically checked.

`dnf-automatic.timer` systemd unit is a timer, not a service, so we can't use the services templates.  I have created a new template which generates OVALs, Bash and Ansible remediations for systemd timers which is very similar to service templates.

#### Rationale:
This configuration setting is a part of RHEL 8 CC.
Fixes #4478


linux_os/guide/system/software/updating/timer_dnf-automatic_enabled/rule.yml
```diff
@@ -0,0 +1,26 @@
...
+title: Enable dnf-automatic Timer
+
+description: |-
+    {{{ describe_timer_enable(timer="dnf-automatic") }}}
+
+rationale: |-
+    The <tt>dnf-automatic</tt> is an alternative command line interface (CLI) to <tt>dnf upgrade</tt> with specific facilities to make it suitable to be executed automatically and regularly from systemd timers, cron jobs and similar.
+    The tool is controlled by <tt>dnf-automatic.timer</tt> SystemD timer.
+
+
```
yuumasato: There are some extra newlines here.

jan-cerny:
@yuumasato 

They're very similar. The OVAL template for services check if both SERVICENAME.service and SERVICENAME.socket are a dependency of multi-user.target, but the OVAL template for timers check only TIMERNAME.timer. The Ansible template for services uses service Ansible module, the template for timers uses systemd Ansible module. The templates for timers have only systemd version because I think timers don't exist in other systems.

I think be reworked that they will share the same template, it probably will need to add a new column in CSV indicating the type of systemd unit, and the tempaltes will contain more if/else conditions.

Do you prefer merging the templates together or keeping them separate?

shared/templates/template_OVAL_timer_enabled
```diff
@@ -0,0 +1,47 @@
+<def-group>
+
+{{% if init_system == "systemd" and target_oval_version >= [5, 11] %}}
```
yuumasato: Any special reason not to add support for OVAL 5.10?

ggbecker: @jan-cerny you can probably reuse the implementation of service enabled for OVAL 5.10. See: https://github.com/ComplianceAsCode/content/blob/master/shared/templates/template_OVAL_service_enabled#L60

jan-cerny: @yuumasato I haven't looked into that yet.

@ggbecker That one probably not, because that's for SysV init and my understaning is that timers are a feature of SystemD. But something like https://github.com/ComplianceAsCode/content/blob/master/shared/templates/template_OVAL_service_disabled#L199  could work.

jan-cerny:
I have added support for OVAL 5.10 in the template. I have also rebased on master and resolved conflicts.

jan-cerny:
I have rebased on master and resolved conflicts.

rhel8/profiles/ospp.profile
```diff
@@ -1222,5 +1222,5 @@ selections:
     ## Configure dnf-automatic to Install Available Updates Automatically
     - dnf-automatic_apply_updates
 
-    ## TO DO: https://github.com/ComplianceAsCode/content/issues/4478
-    #systemctl enable --now dnf-automatic.timer
+    ## RHEL 8 CCE-82360-9: Enable dnf-automatic Timer
```
yuumasato: There is no need for the CCE number

jan-cerny:
I have removed CCEs from the profile YAML and I have raised an CSVLineError.

jan-cerny:
@openscap-ci test this please

shared/templates/template_OVAL_timer_enabled
```diff
@@ -0,0 +1,81 @@
...
+
+  <definition class="compliance" id="timer_{{{ TIMERNAME }}}_enabled" version="1">
+    <metadata>
+      <title>Timer {{{ TIMERNAME }}} Enabled</title>
+      <affected family="unix">
+        <platform>multi_platform_all</platform>
+      </affected>
+      <description>The {{{ TIMERNAME }}} timer should be enabled if possible.</description>
+    </metadata>
+    <criteria comment="package {{{ PACKAGENAME }}} installed and timer {{{ TIMERNAME }}} is  configured to start" operator="AND">
```
ggbecker:
```suggestion
    <criteria comment="package {{{ PACKAGENAME }}} installed and timer {{{ TIMERNAME }}} is configured to start" operator="AND">
```

rhel8/templates/csv/timers_enabled.csv
```diff
@@ -0,0 +1 @@
+dnf-automatic,dnf-automatic
```
ggbecker: Can you add the documentation header similar to other templates? It is hard to know which format it accepts and which parameters are optional.

fedora/templates/csv/timers_enabled.csv
```diff
@@ -0,0 +1 @@
+dnf-automatic,dnf-automatic
```
ggbecker: Can you add the documentation header similar to other templates? It is hard to know which format it accepts and which parameters are optional.

jan-cerny:
@ggbecker I have removed the space and I have added a documentation header to all timers_enabled,csv files.

PR #4609, opened 2019-07-11
Expand documentation of xccdf rule severity

yuumasato:
#### Description:

- Describe what the severity levels of a rule mean

#### Rationale:

- These descriptions were sourced from XCCDF v1.2

PR #4616, opened 2019-07-12
Revert "Expand documentation of xccdf rule severity"

shawndwells:
Reverts ComplianceAsCode/content#4609

NACK to original PR. These are not how the severity codes are currently defined. Should use the DISA language.

(and they should be a conversation about using something else, like NIST that Trevor mentioned awhile ago).

### DIFF
--- a/docs/manual/developer_guide.adoc
+++ b/docs/manual/developer_guide.adoc
@@ -710,28 +710,9 @@ A rule YAML file has one implied attribute:
 
 A rule itself contains these attributes:
 
-* `severity`: Is used for metrics and tracking. It can have one of the following values: `unknown`, `info`, `low`, `medium`, or `high`.
-+
-[cols="2", options="header"]
-|===
-|Level | Description
-|`unknown`
-|Severity not defined (default)
+* `severity`: Can be `unknown`, `low`, `medium`, or `high`.
 
-|`info`
-|Rule is informational only. Failing the rule doesn't imply failure to conform to the security guidance of the benchmark.
-
-|`low`
-|Not a serious problem
-
-|`medium`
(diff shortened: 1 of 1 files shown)


yuumasato:
> NACK to original PR. These are not how the severity codes are currently defined. 

We use the XCCDF Rule severity, as defined in XCCDF 1.2 ([NISTIR 7275 Rev. 4](https://csrc.nist.gov/CSRC/media/Publications/nistir/7275/rev-4/final/documents/nistir-7275r4_updated-march-2012_clean.pdf)).
Is there some other way to assign severity to rules?

>  Should use the DISA language.

The language can just be improved, the severity codes should follow XCCDF.
I don't understand how the content will be SCAP compliant without following the XCCDF standard.

> (and they should be a conversation about using something else, like NIST that Trevor mentioned awhile ago).

I'm not familiar with `Special Publication 800-30`, maybe you could document how to use it to map to what we use.

matejak:
This is tricky, as only a fraction of rules are associated with DISA, and the guide in question is a general contributors guide.
I am OK to merge the PR, as the to-be-reverted description did not contain anything surprising or insightful anyway, and we don't feature the `info` severity anywhere.

yuumasato:
> I am OK to merge the PR, as the to-be-reverted description did not contain anything surprising or insightful anyway,

On the other hand, it is only documented in `oscap` manual, and very sparsely.

> and we don't feature the `info` severity anywhere.

Up until now... :)

redhatrises:
Rather than reverting... I think that we should add a note about DISA's suggested severity levels. @shawndwells can you provide a link to those?

redhatrises:
Closing in favor of #4818

PR #4618, opened 2019-07-15
Disable storing core dumps.

adelton:
#### Description:

- Disable storing core dumps.

#### Rationale:

- A core dump includes a memory image taken at the time the operating system terminates an application. The memory image could contain sensitive data and is generally useful only for developers trying to debug problems.


jan-cerny:
@openscap-ci test this please

PR #4620, opened 2019-07-15
Remove supported_platforms parameter from dnf-automatic_apply_updates rule

mildas:
#### Description:
Remove supported_platforms parameter from macro.

Parameter was removed in #4596 but this rule was merged shortly before that.


ggbecker:
LGTM. It fixes problem building the content. Thanks for the fix.

PR #4622, opened 2019-07-15
Install fapolicyd for OSPP.

adelton:
#### Description:

- Install fapolicyd for OSPP.

#### Rationale:

- fapolicyd (File Access Policy Daemon) implements application whitelisting to decide file access rights.


jan-cerny:
Thanks!

PR #4624, opened 2019-07-15
Start using oval_sshd_config jinja macros for sshd rules

mildas:
#### Description:
#4593 introduced `oval_sshd_config` macro, which checks value for specified parameter in `/etc/sshd_config` file.

In this PR I want to change rules, which exactly fit for this macro.
There's more rules for sshd, which remain unchanged for now - they use external variable [sshd_set_keepalive](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/ssh/ssh_server/sshd_set_keepalive/oval/shared.xml#L55) or conditions [sshd_disable_empty_passwords](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/ssh/ssh_server/sshd_disable_empty_passwords/oval/shared.xml#L16).


mildas:
It's ready to review.

linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers/oval/shared.xml
```diff
@@ -1,45 +1 @@
-<def-group>
-  <definition class="compliance" id="sshd_use_approved_ciphers" version="1">
-    <metadata>
-      <title>Use Only Approved Ciphers</title>
-      <affected family="unix">
-        <platform>multi_platform_wrlinux</platform>
```
ggbecker: Please review the prodtype from `rule.yml` according to affected platform. Here I can see that is multi platform wrlinux but in `prodtype` it has only `prodtype: wrlinux1019,rhel6,rhel7,rhel8,ol7,rhv4`. Missing the wrlinux8.

ggbecker:
I've have seen that you used missing_parameter_pass=true for OVAL macros. Just to confirm that those are the default values, right?

linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers/oval/shared.xml
ggbecker: Actually, I'm not sure it was the intention to restrict to `wrlinux1019`. We need to investigate.

mildas: I will take a look and compare oval platforms and `rule.yml` platforms. It's possible that someone changed oval platforms but forgot about `rule.yml`.

mildas:
@ggbecker yes. Just to be sure - e.g.
`oval_sshd_config(parameter="HostbasedAuthentication", value="no", missing_parameter_pass=true)`

sshd_config manual page: 
> HostbasedAuthentication
             Specifies whether rhosts or /etc/hosts.equiv authentication together with successful public key client host authentication is allowed (host-based authentication).  The default is **no**.

Is it what you meant, right?

linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers/oval/shared.xml
mildas: I checked sshd rules and for most of them platforms in ovals don't correspond to `rule.yml` prodtypes and it's not possible to visit complete history of `rule.yml`. In some cases `prodtype` in `rule.yml` is missing (so it includes all products), but in OVAL it doesn't include all products and vice versa (e.g. example you wrote).

I will include to `rule.yml` products, which were in OVAL but not in `prodtype`. But I won't restrict `prodtype` if it includes products, which were not mentioned in OVAL `platform`.

ggbecker: Ok. I'm okay with that.

ggbecker:
> @ggbecker yes. Just to be sure - e.g.
> `oval_sshd_config(parameter="HostbasedAuthentication", value="no", missing_parameter_pass=true)`
> 
> sshd_config manual page:
> 
> > HostbasedAuthentication
> > Specifies whether rhosts or /etc/hosts.equiv authentication together with successful public key client host authentication is allowed (host-based authentication).  The default is **no**.
> 
> Is it what you meant, right?

Yes, it is.

mildas:
Updated with missing `wrlinux8` prodtype.

ggbecker:
LGTM. @yuumasato Just want to double check, is it ok to discard the content from `comment` and `description` in favor of the ones generated by the macro? Such as: https://github.com/ComplianceAsCode/content/pull/4624/files#diff-9f9924a321e28afe0bbbbfdeb04d7168L8

Assuming that the rule.yml content is enough source of information to the user.

mildas:
@ggbecker Updated PR with more replaced rules, because of new product check in lineinfile macro (#4639) 
Please, check last commit, if it's still in good shape. Thank you

ggbecker:
> @ggbecker Updated PR with more replaced rules, because of new product check in lineinfile macro (#4639)
> Please, check last commit*, if it's still in good shape. Thank you
> 
> Edit: *Last 2 commits

It looks great. Thanks for the patch. Lots of lines have been removed in favor of using the macros.

PR #4625, opened 2019-07-15
Add developer documentation for OVAL jinja macros.

ggbecker:
#### Description:

Add developer documentation for OVAL jinja macros. Explains which macros are available, what they do and briefly how to proceed when creating new high level macros.

Relates to merged PR: https://github.com/ComplianceAsCode/content/pull/4593

### DIFF
--- a/docs/manual/developer_guide.adoc
+++ b/docs/manual/developer_guide.adoc
@@ -1,4 +1,5 @@
 # ComplianceAsCode Developer Guide
+:rootdir: ../..
 :imagesdir: ./images
 :templatesdir: ../../shared/templates
 :rhel7dir: ../../rhel7/templates/csv
@@ -970,11 +971,12 @@ This is an example of a check, written using the custom OVAL syntax, that checks
 ----
 
 ===== Macros
+* See the  for more
 
-Jinja macros for OVAL checks are located in `/shared/macros-oval.jinja`. These currently include the following high-level macros:
+Jinja macros for OVAL checks are located in link:{rootdir}/shared/macros-oval.jinja[macros-oval.jinja]. These currently include the following high-level macros:
 
-- `oval_sshd_config` -- check a parameter and value in the sshd configuration
-- `oval_grub_config` -- check a parameter and value in the grub configuration
+- `oval_sshd_config` -- check a parameter and value in the sshd configuration file
+- `oval_grub_config` -- check a parameter and value in the grub configuration file
 - `oval_check_config_file` -- check a parameter and value in a given configuration file
 
 Always consider reusing `oval_check_config_file` when creating new macros, it has some logic that will save you some time (e.g.: platform applicability).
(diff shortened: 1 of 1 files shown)


docs/manual/developer_guide.adoc
```diff
@@ -969,6 +969,39 @@ This is an example of a check, written using the custom OVAL syntax, that checks
   </unix:file_object>
 ----
 
+===== Macros
+
+Jinja macros for OVAL checks are located in `/shared/macros-oval.jinja`. These currently include the following high-level macros:
+
+- `oval_sshd_config` -- check a parameter and value in the sshd configuration
+- `oval_grub_config` -- check a parameter and value in the grub configuration
```
yuumasato: `configuration file`

docs/manual/developer_guide.adoc
```diff
@@ -969,6 +969,39 @@ This is an example of a check, written using the custom OVAL syntax, that checks
   </unix:file_object>
 ----
 
+===== Macros
+
+Jinja macros for OVAL checks are located in `/shared/macros-oval.jinja`. These currently include the following high-level macros:
```
yuumasato: I think we should refer to project directories without `/` in the beginning, because that signalizes an absolute path.

yuumasato: I noticed that `macros-ansible.jinja` file is referred to with `/shared`, we should fix them.

docs/manual/developer_guide.adoc
```diff
@@ -969,6 +969,39 @@ This is an example of a check, written using the custom OVAL syntax, that checks
...
+
+- set of low-level macros to build the OVAL checks to test if a file exists:
+```
+oval_config_file_exists_criterion
+oval_config_file_exists_test
+oval_config_file_exists_object
+```
+
+====== Platform applicability
+Platform applicability is given by the `prodtype` property in the rule.yml file. If you are using `oval_check_config_file` macro directly or indirectly, it should be enough to define `prodtype`. Default is `all` platforms. If you intend to define your own OVAL check structure please consider using `oval_affected` macro from `/shared/macros.jinja`.
```
yuumasato: `define your own OVAL check structure please` what do you mean by structure?

docs/manual/developer_guide.adoc
cipherboy: Ah sorry. I used absolute paths to refer to locations from the root of the `ComplianceAsCode/content` directory, wherever that is for the user. We can use relative locations if you think that's better though.

docs/manual/developer_guide.adoc
yuumasato: I guess just removing `structure` from the sentence solves my confusion.

ggbecker: done.

docs/manual/developer_guide.adoc
ggbecker: done

docs/manual/developer_guide.adoc
ggbecker: @yuumasato I changed to reference the file using `link`. What do you think?

yuumasato: Better than my suggestion.

PR #4626, opened 2019-07-16
Add OVAL macro to check INI files

jan-cerny:
#### Description:
Introduces macro oval_check_ini_file to easily generate OVAL which checks options within INI files.

As an example it is used in OVAl for rule dnf-automatic_apply_updates.

#### Rationale:
Will speed up creating OVALs for rules which check configuration in an INI file.

### DIFF
--- a/shared/macros-oval.jinja
+++ b/shared/macros-oval.jinja
@@ -6,7 +6,7 @@
 
 {{#
   High level macro which checks if a particular combination of parameter and value in a configuration file is set.
-  This function can take nine parameters:
+  This macro can take ten parameters:
     - path (String): Path to the configuration file to be checked.
     - prefix_regex (String): Regular expression to be used in the beginning of the OVAL text file content check.
     - parameter (String): The parameter to be checked in the configuration file.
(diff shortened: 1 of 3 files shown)


shared/macros-oval.jinja
```diff
@@ -230,3 +247,19 @@
...
+    - path (String): Path to the configuration file to be checked.
+    - section (String): The parameter will be checked only within the given section defined by [section].
+    - parameter (String): The parameter to be checked in the configuration file.
+    - value (String): The value to be checked. This can also be a regular expression (e.g: value1|value2 can match both values).
+    - missing_parameter_pass (boolean): If set, the check will also pass if the parameter is not present in the configuration file (default is applied).
+    - application (String): The application which the configuration file is being checked. Can be any value and does not affect the OVAL check.
+    - multi_value (boolean): If set, it means that the parameter can accept multiple values and the expected value must be present in the current list of values.
+    - missing_config_file_fail (boolean): If set, the check will fail if the configuration is not existent in the system.
+#}}
+{{%- macro oval_check_ini_file(path='', section='', parameter='', value='', missing_parameter_pass=false, application='', multi_value=false, missing_config_file_fail=false) %}}
```
ggbecker: As devel doc was added for these macros, please add doc for this macro as well. Here: https://github.com/ComplianceAsCode/content/blob/master/docs/manual/developer_guide.adoc#macros

jan-cerny:
rebased and added doc

shared/macros-oval.jinja
ggbecker: I've seen that you updated accordingly. Thanks.

shared/macros-oval.jinja
```diff
@@ -115,15 +118,28 @@
 
 {{#
   Macro to check if a parameter in a configuration file is set (Object definition).
-  This function can take five parameters:
+  This macro can take six parameters:
```
ggbecker: Please change every occurrence of `This function` to `This macro` in the file.

jan-cerny: done

shared/macros-oval.jinja
```diff
@@ -115,15 +118,28 @@
...
+{{%- macro oval_line_in_file_object(path='', section='', prefix_regex='^[\s]*', parameter='', separator_regex='[\s]+', missing_parameter_pass=false) -%}}
 {{%- set suffix_id = "" -%}}
+{{%- if section %}}
+{{%- if missing_parameter_pass %}}
+{{#
+  There is no need for having a regular expression with a capture
+  group "(\S*)" when checking if the parameter is not present in
+  the configuration file.
+#}}
+{{%- set regex = "^\s*\["+section+"\].*(?:\\n\s*[^[\s].*)*\\n"+prefix_regex+parameter+separator_regex+"\S*\s*(?:|(?:#.*))?$" -%}}
```
ggbecker: Can you make improve the `regex` variable construction? Trying to reuse as much the regular expressions. I might help decreasing maintenance costs.

yuumasato: Is the trailing part `"\S*\s*(?:|(?:#.*))?$"` of the regex needed?

jan-cerny: @yuumasato great point, it might not be needed, I'll check it

jan-cerny: @ggbecker @yuumasato I have tried with regex101 and I think it's not needed in situation we only want to check the option is not present. I have removed it in 4918c60.

ggbecker: Great. It looks way better now.

ggbecker:
LGTM. Please fix the conflicts so we can merge it. Thanks.

jan-cerny:
@ggbecker rebased

yuumasato:
Merging!

PR #4628, opened 2019-07-16
WIP: Preserve correct configurations

cipherboy:
#### Problem Statement

Our bash and ansible remediations currently match in effect. First they strip all lines with the configuration parameter. Then they add the line with the correct value, regardless of whether or not they stripped a line with the correct value already.

One could envision a capable admin who already has a compliant configuration getting annoyed at this behavior, since SCAP remediations could be continually rewriting his config. We have the capabilities to only remove non-compliant lines and add the compliant line only if it is missing, so we probably should.

It is definitely possible with a three-line check (remove-incorrect, check, add-if-missing), but I'm wondering if anyone has ideas about how to make it a two-line check (remove-incorrect, add-if-missing). Thoughts?

#### Help wanted:

This does the easy part: update it to not remove a configuration line with the correct value.

What remains is checking if the file has the correct line, and only add it if it doesn't. 

Also, renames `ansible_etc_profile_set` to `ansible_etc_profile_set_variable` and updates its description -- it really can't be used with a command with `=` as the separator and no way to change the separator. 

### DIFF
--- a/docs/manual/developer_guide.adoc
+++ b/docs/manual/developer_guide.adoc
@@ -1093,8 +1093,8 @@ The Playbook is generated in `/build/ansible/<product>-playbook-<profile_id>.yml
 
 Jinja macros for Ansible content are located in `/shared/macros-ansible.jinja`. These currently include the following high-level macros:
 
+- `ansible_etc_profile_set_variable` -- ensure a variable gets set in /etc/profile or /etc/profile.d
 - `ansible_sshd_set` -- set a parameter in the sshd configuration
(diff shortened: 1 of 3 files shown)


shawndwells:
Hi @cipherboy , this seems like a really great improvement. Is this still in WIP? If not, mind if this PR is closed? Can always be re-opened if you have the time to continue this work!

cipherboy:
My ansible-fu escaped me on this one. I couldn't figure out how to do the
last part (mentioned under the help wanted section).

If anyone has ideas, feel free to take the current code and extend it as
necessary.

Until then, perhaps an issue is better than a WIP PR.

PR #4629, opened 2019-07-17
Make the Bash lineinfile functions callable from remediations

jan-cerny:
#### Description:
The functions called in Bash remediation need to have a corresponding
.sh file with the same name as the name of the function located
in bash_remediation_functions directory. Otherwise they will not be
expanded. To be able to use the high-level functions such as
set_config_file, we will add an including mechanism.  Users will call
include_lineinfile and the build system will include the whole file
include_lineinfile.sh. This patch also copies the die function because
the build system does not support dependent functions in other files.

#### Rationale:

Enables using  Bash functions introduced in #4597 in our remediations.

### DIFF
--- a/shared/bash_remediation_functions/include_lineinfile.sh
+++ b/shared/bash_remediation_functions/include_lineinfile.sh
@@ -1,6 +1,19 @@
 # Functions which minic Ansible's lineinfile module.
 # dependencies: die.sh
 
+function include_lineinfile() {
+    :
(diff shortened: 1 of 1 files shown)


ggbecker:
LGTM. Unfortunately when we have multiple functions inside the bash remediation, one needs to create a new function with the same name of the file so it can be properly embedded into the datastream.

PR #4630, opened 2019-07-17
Fixes of Jinja macro `oval_check_config_file`

jan-cerny:
#### Description:

This pull request proposes 2 changes in Jinja macro `oval_check_config_file` that generates OVAL to check config files:

1. Set multi_value to false instead of an empy string
The variable is used as boolean and is also documented as boolean.

2. Change default separator regex
The former default separator regex `(?-i)[\s]+` was wrong, because `(?-i)` means "stop being case-insensitive". It needs a pair `(?i)` before in the regular expression and is ineffective but confusing alone.
It should be harmless to change the default value now because all of the occurrences are calling the macro with the `separator_regex` parameter set.

#### Rationale:
Makes the default values more sensible and therefore it makes the macro generally usable.

### DIFF
--- a/shared/macros-oval.jinja
+++ b/shared/macros-oval.jinja
@@ -203,7 +203,7 @@
   sshd_config has case-insensitive parameters (but case-sensitive values).
 #}}
 {{%- macro oval_sshd_config(parameter='', value='', missing_parameter_pass=false, multi_value=false, missing_config_file_fail=false, supported_platforms=['all']) %}}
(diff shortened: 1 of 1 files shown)


ggbecker:
Please change `prefix_regex="(?i)^[\s]*"` to `prefix_regex="^[\s]*(?i)"` at line 206: https://github.com/ComplianceAsCode/content/pull/4630/files#diff-b0e4d720e043f25a5f0538373c127eb6L206

jan-cerny:
@ggbecker changed

ggbecker:
LGTM. Great, thank you for the patch.

PR #4631, opened 2019-07-17
New rule disable_tmux_status_line

jan-cerny:
#### Description:
Check if status line is turned off in tmux configuration.

#### Rationale:
This rule is a part of RHEL 8 CC configuration, however, I can't
create a rationale.


linux_os/guide/system/accounts/accounts-physical/screen_locking/console_screen_locking/disable_tmux_status_line/oval/shared.xml
```diff
@@ -0,0 +1,9 @@
+{{{
+oval_check_config_file(
```
mildas: What do you think about creating new high level macro? Similar as `ansible_tmux_set`?

jan-cerny: Cool idea! It leads to a question that we need to answer: Is it a good idea to create a small macro for every configured application or do we want to use one generic macro everywhere? I lean towards the former option, because they are more readable when used.

mildas: These small macros make sense if they will be used several times. And because in OSPP profile are 3 `tmux.conf` rules now and we already have high level ansible ans bash macros for them, then I'd say it makes sense.

Question is - is 3 rules enough to create new high level macro for them? Hard to say for next rule. But for this one as I mentioned we already existing ansible/bash macro, it makes sense for me (either no high level macro or high level macro for oval/ansible/bash)

jan-cerny:
This isn't a security setting and doesn't need to be checked by XCCDF rule.

PR #4632, opened 2019-07-17
Remove unnecessary packages from ospp

comps:
These don't affect any security claims so we shouldn't check for them.

### DIFF
--- a/rhel8/profiles/ospp.profile
+++ b/rhel8/profiles/ospp.profile
@@ -191,17 +191,11 @@ selections:
     ## Required Packages
     #################################################################
 
-    ## RHEL 8 CCE-82995-2: Install cryptsetup-luks Package
-    - package_cryptsetup-luks_installed
-
     ## RHEL 8 CCE-82994-5: Install sssd-ipa Package
     - package_sssd-ipa_installed
 
     ## RHEL 8 CCE-80844-4: Install aide Package
     - package_aide_installed
-
-    ## RHEL 8 CCE-82989-5: Install binutils Package
-    - package_binutils_installed
 
     ## RHEL 8 CCE-82985-3: Install dnf-automatic Package
     - package_dnf-automatic_installed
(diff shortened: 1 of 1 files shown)


dahaic:
Thanks!

rhel8/profiles/ospp.profile
```diff
@@ -191,18 +191,12 @@ selections:
     ## Required Packages
     #################################################################
 
-    ## RHEL 8 CCE-82995-2: Install cryptsetup-luks Package
-    - package_cryptsetup-luks_installed
```
shawndwells: Isn't this one needed to ensure hard drives are encrypted at rest with LUKS? 

Or rather, LUKS is the only FIPS 140-2 evaluated encryption at rest in RHEL. How do we ensure the system is configured with LUKS?

rhel8/profiles/ospp.profile
```diff
@@ -270,12 +258,6 @@ selections:
     ## RHEL 8 CCE-82943-2: Uninstall gssproxy Package
     - package_gssproxy_removed
 
-    ## RHEL 8 CCE-82939-0: Uninstall geolite2-city Package
-    - package_geolite2-city_removed
```
shawndwells: IIRC, the geocity packages need to be explicitly removed to ensure the system does not contain geolocation capabilities. 

Is there another way to ensure this?

rhel8/profiles/ospp.profile
comps: I was told that OSPPv4.2(.1) doesn't have, and we don't claim, any HDD encryption, hence the package doesn't need to be installed.

rhel8/profiles/ospp.profile
comps: I'm not aware of such requirement and I've been told their removal has no implication on OSPP claims. If you know of a specific SFR or other identifier, please do share.

PR #4633, opened 2019-07-18
Add mask_service parameter to services disabled template.

### LINKED ISSUE #4386: Service disabled checking ignores systemd masked units
(May be related to #2632.)

TL;DR: `xccdf_org.ssgproject.content_rule_service_kdump_disabled` reports failure when the service is masked instead of being disabled. My expected result (based on what the OVAL already checks for) would be a reported success as the service cannot start.

Looking at `shared/templates/template_OVAL_service_disabled`, the checking seems to be based on wheter `multi-user.target` "wants" the given service, which is the same as `systemctl enable/disable <service>`.
```
$ systemctl disable kdump
Removed symlink /etc/systemd/system/multi-user.target.wants/kdump.service.

$ systemctl enable kdump
Created symlink from /etc/systemd/system/multi-user.target.wants/kdump.service to /usr/lib/systemd/system/kdump.service.

$ ls -l /etc/systemd/system/multi-user.target.wants/kdump.service
lrwxrwxrwx. 1 root root 37 May 31 03:19 /etc/systemd/system/multi-user.target.wants/kdump.service -> /usr/lib/systemd/system/kdump.service
```
Masking, however, is different; it doesn't touch the symlink above, it instead creates a dummy service file with higher priority (`/etc` overrides `/usr/lib`) and makes it point to `/dev/null` so that systemctl literally cannot do any operations on the unit file other than unmasking.
```
$ systemctl mask kdump
Created symlink from /etc/systemd/system/kdump.service to /dev/null.

$ systemctl cat kdump
# /dev/null

$ systemctl is-enabled kdump
masked

$ systemctl unmask kdump
Removed symlink /etc/systemd/system/kdump.service.

$ systemctl cat kdump
# /usr/lib/systemd/system/kdump.service
[Unit]
Description=Crash recovery kernel arming
<snip>
```
As you can see, a service can be both "enabled" (started by `multi-user.target`) and "masked" at the same time - the enable/disable is like an ON/OFF switch, masking is like removing the power supply. In either case, the service is not automatically started.

Additionally, the big difference is that **disabled services can still be started** by the sysadmin or as a dependency and some certifications might need the "never start" guarantee that masked gives.

...

Therefore I propose:
 1. Modifying `template_OVAL_service_disabled` so that it detects masked services (`/etc/systemd/system/<servicename>.service` is a symlink to `/dev/null`)
 1. Adding a profile variable for `template_BASH_service_disabled` that would allow to specify whether a service should be just disabled or masked as well

PS: Some modification to `systemd_ocil_service_disabled` may also be needed.

Thanks!

ggbecker:
#### Description:

- Introduces `mask_service` parameter to services disabled template allowing to build checks+remediations which consider the `masked` attribute of services (meaning that the service cannot be started at all)

#### Rationale:

- Only deactivating a service may be enough as it can be started as dependency of other services. Masking the service ensures it cannot be started at all, be it by manually starting or by dependency of another service.

- Fixes #4386 

Tasks:

- [x] OVAL check for version 5.11 - LoadState property is set to masked
- [x] Bash+Ansible mask the service is `mask_service` is set
- [x] ~~OVAL check for version 5.10 - symlink to /dev/null -~~ in this OVAL version does not support checking the symlink path, so it is not possible to implement.
- [x] Change service_disabled.csv from products other RHEL8? at least documentation
- [x] update rule text and jinja macros such as: `systemd_ocil_service_disabled`
- [x] Update all test scenarios for services disabled template
- [x] Fix ansible remediation

### DIFF
--- a/debian8/templates/csv/services_disabled.csv
+++ b/debian8/templates/csv/services_disabled.csv
@@ -1,2 +1,2 @@
-# <service_name, package_name, daemon_name[, dont_mask_service]> (as recognized by chkconfig / systemd. To be used when daemon_name differs from service_name) (if dont_mask_service is set the service will not be masked, can be any string)
+# <service_name, package_name, daemon_name[, mask_service]> (as recognized by chkconfig / systemd. To be used when daemon_name differs from service_name) (mask_service: mask the service. Acceptable values are: true or false. Default value is true.)
 sshd,openssh-server,ssh
--- a/debian8/templates/csv/services_disabled.csv
+++ b/debian8/templates/csv/services_disabled.csv
@@ -1,2 +1,2 @@
-# service_name, package_name, daemon_name (as recognized by chkconfig / systemd. To be used when daemon_name differs from service_name)
+# <service_name, package_name, daemon_name[, dont_mask_service]> (as recognized by chkconfig / systemd. To be used when daemon_name differs from service_name) (if dont_mask_service is set the service will not be masked, can be any string)
 sshd,openssh-server,ssh
--- a/shared/templates/create_services_disabled.py
+++ b/shared/templates/create_services_disabled.py
@@ -37,7 +37,7 @@ class ServiceDisabledGenerator(FilesGenerator):
             print("\tError unpacking servicename, packagename, daemonname " +
                   "and mask_service: " + str(e))
             print("\tEntry: %s\n" % serviceinfo)
-            sys.exit(1)
+            raise CSVLineError()
 
         if not daemonname:
             daemonname = servicename
--- a/shared/templates/create_services_disabled.py
+++ b/shared/templates/create_services_disabled.py
@@ -4,19 +4,31 @@
 
 import sys
 
-from template_common import FilesGenerator, UnknownTargetError
+from template_common import FilesGenerator, UnknownTargetError, CSVLineError
 
 
 class ServiceDisabledGenerator(FilesGenerator):
     def generate(self, target, serviceinfo):
         try:
-            # get the items out of the list
-            servicename, packagename, daemonname = serviceinfo
+
+        # get the items out of the list
+            # items can be in format
+            # <service_name, package_name, daemon_name> or
+            # <service_name, package_name, daemon_name, mask_service>
+            mask_service = False
+            if len(serviceinfo) == 3:
+                servicename, packagename, daemonname = serviceinfo
+            elif len(serviceinfo) == 4:
+                servicename, packagename, daemonname, mask_service = serviceinfo
+                # use boolean instead of any string that came from csv file
+                mask_service = True
+            else:
+                raise CSVLineError()
             if not packagename:
                 packagename = servicename
         except ValueError as e:
             print("\tEntry: %s\n" % serviceinfo)
-            print("\tError unpacking servicename, packagename, and daemonname: " + str(e))
+            print("\tError unpacking servicename, packagename, daemonname and mask_service: " + str(e))
             sys.exit(1)
 
         if not daemonname:
@@ -27,7 +39,8 @@ class ServiceDisabledGenerator(FilesGenerator):
                 "./template_BASH_service_disabled",
                 {
                     "SERVICENAME": servicename,
-                    "DAEMONNAME": daemonname
+                    "DAEMONNAME": daemonname,
+                    "MASK_SERVICE": mask_service
                 },
                 "./bash/service_{0}_disabled.sh", servicename
             )
@@ -37,7 +50,8 @@ class ServiceDisabledGenerator(FilesGenerator):
                 "./template_ANSIBLE_service_disabled",
                 {
                     "SERVICENAME": servicename,
-                    "DAEMONNAME": daemonname
+                    "DAEMONNAME": daemonname,
+                    "MASK_SERVICE": mask_service
                 },
                 "./ansible/service_{0}_disabled.yml", servicename
             )
@@ -47,7 +61,8 @@ class ServiceDisabledGenerator(FilesGenerator):
                 "./template_PUPPET_service_disabled",
                 {
                     "SERVICENAME": servicename,
-                    "DAEMONNAME": daemonname
+                    "DAEMONNAME": daemonname,
+                    "MASK_SERVICE": mask_service
                 },
                 "./puppet/service_{0}_disabled.yml", servicename
             )
@@ -58,7 +73,8 @@ class ServiceDisabledGenerator(FilesGenerator):
                 {
                     "SERVICENAME": servicename,
                     "DAEMONNAME":  daemonname,
-                    "PACKAGENAME": packagename
+                    "PACKAGENAME": packagename,
+                    "MASK_SERVICE": mask_service
                 },
                 "./oval/service_{0}_disabled.xml", servicename
             )
@@ -67,4 +83,4 @@ class ServiceDisabledGenerator(FilesGenerator):
 
     def csv_format(self):
         return("CSV should contains lines of the format: " +
-               "servicename,packagename")
+               "servicename,packagename,daemonname[,mask_service]")
--- a/shared/templates/create_services_disabled.py
+++ b/shared/templates/create_services_disabled.py
@@ -16,21 +16,27 @@ class ServiceDisabledGenerator(FilesGenerator):
             # get the items out of the list
             # items can be in format
             # <service_name, package_name, daemon_name> or
-            # <service_name, package_name, daemon_name, dont_mask_service>
+            # <service_name, package_name, daemon_name, mask_service>
             if len(serviceinfo) == 3:
                 servicename, packagename, daemonname = serviceinfo
             elif len(serviceinfo) == 4:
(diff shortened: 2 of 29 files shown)


ggbecker:
OVAL 5.10 doesn't have support to which path the symlink is referencing, so it is not possible to implement the check in this version. OVAL 5.11 has the systemd support and the service's properties are checked instead.

By default, the service will be masked unless the user set in the `services_disabled.csv` the value `dont_mask_service`.

ssg/build_templates.py
```diff
@@ -73,7 +73,6 @@ def __init__(self, env_yaml):
             "audit_rules_login_events.csv":  AuditRulesLoginEventsGenerator(),
             "audit_rules_privileged_commands.csv":  AuditRulesPrivilegedCommandsGenerator(),
             "audit_rules_usergroup_modification.csv":  AuditRulesUserGroupModificationGenerator(),
-            "audit_rules_usergroup_modification.csv":  AuditRulesUserGroupModificationGenerator(),
```
matejak: Is this change really intended?

shared/templates/template_OVAL_service_disabled
```diff
@@ -17,7 +17,10 @@
       <criteria operator="AND" comment="service {{{ SERVICENAME }}} is not configured to start">
         <criterion comment="{{{ SERVICENAME }}} not wanted by multi-user.target" test_ref="test_{{{ SERVICENAME }}}_not_wanted_by_multi_user_target" />
         <criterion comment="{{{ SERVICENAME }}} socket not wanted by multi-user.target" test_ref="test_{{{ SERVICENAME }}}_socket_not_wanted_by_multi_user_target" />
-        <criterion comment="{{{ SERVICENAME }}} is not running" test_ref="test_service_not_running_{{{ SERVICENAME }}}" />
```
matejak: Why was this criterion removed?

shared/templates/create_services_disabled.py
```diff
@@ -4,19 +4,33 @@
...
-            servicename, packagename, daemonname = serviceinfo
+            # items can be in format
+            # <service_name, package_name, daemon_name> or
+            # <service_name, package_name, daemon_name, dont_mask_service>
+            if len(serviceinfo) == 3:
+                servicename, packagename, daemonname = serviceinfo
+            elif len(serviceinfo) == 4:
+                servicename, packagename, daemonname, dont_mask_service = serviceinfo
+                # use boolean instead of any string that came from csv file
+                mask_service = False if dont_mask_service != "" else True
```
matejak: I would advise strict interfaces, as they catch typos. So let's decide what values of the final argument are allowed, and let's enforce that.

shared/templates/template_BASH_service_disabled
```diff
@@ -5,18 +5,25 @@
...
+"$SYSTEMCTL_EXEC" list-unit-files | grep -q '^{{{ DAEMONNAME }}}.socket' && "$SYSTEMCTL_EXEC" disable '{{{ DAEMONNAME }}}.socket' {{%- if MASK_SERVICE %}} && "$SYSTEMCTL_EXEC" mask '{{{ DAEMONNAME }}}.socket' {{%- endif %}}
 # The service may not be running because it has been started and failed,
 # so let's reset the state so OVAL checks pass.
 # Service should be 'inactive', not 'failed' after reboot though.
 "$SYSTEMCTL_EXEC" reset-failed '{{{ DAEMONNAME }}}.service'
 {{% elif init_system == "upstart" %}}
 
 /sbin/service '{{{ DAEMONNAME }}}' disable
+{{%- if MASK_SERVICE %}}
+/sbin/service '{{{ DAEMONNAME }}}' mask
```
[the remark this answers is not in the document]
ggbecker: I see that even `disable` is not in the list. In this case it should be probably only `stop`

debian8/templates/csv/services_disabled.csv
```diff
@@ -1,2 +1,2 @@
-# service_name, package_name, daemon_name (as recognized by chkconfig / systemd. To be used when daemon_name differs from service_name)
+# <service_name, package_name, daemon_name[, dont_mask_service]> (as recognized by chkconfig / systemd. To be used when daemon_name differs from service_name) (if dont_mask_service is set the service will not be masked, can be any string)
```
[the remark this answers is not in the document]
ggbecker: This was my first proposal if you see here: https://github.com/ComplianceAsCode/content/pull/4633/commits/295b88a0ee80d943eeca16b26a5d2e40f9c4162b#diff-2949e8573ac963346b7b2c124319d617R1

But I was not sure how to handle the default value and that would require to change all service_disabled.csv files. It makes sense to have `mask_service` instead of this. I'll work on that.

shared/templates/template_OVAL_service_disabled
ggbecker: It was removed by mistake. Readding.

shared/templates/template_ANSIBLE_service_disabled
```diff
@@ -17,6 +20,9 @@
     name: "{{{ DAEMONNAME }}}.socket"
     enabled: "no"
     state: "stopped"
+    {{%- if MASK_SERVICE %}}
+    masked: "yes"
```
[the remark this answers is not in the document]
ggbecker: changed to systemd

ssg/build_templates.py
ggbecker: Yes, is a duplicated line. Unless is has some shady meaning.

shared/templates/create_services_disabled.py
ggbecker: Restricted to `true` or `false`.

shared/templates/template_BASH_service_disabled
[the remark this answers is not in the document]
ggbecker: I removed `mask` and also changed `disable` to `stop`.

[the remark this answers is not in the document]
matejak: [Disabling upstart services](https://askubuntu.com/questions/19320/how-to-enable-or-disable-services) is really a crazy procedure, we should get rid of the upstart code (used in RHEL6) as soon as possible.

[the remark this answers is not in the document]
ggbecker: Agreed. ;)

ggbecker:
Ansible remediation does not met all needed modifications so the OVAL check pass. Needs to be revisited. Please do not merge this.

tests/data/group_services/group_avahi/group_disable_avahi_group/rule_service_avahi-daemon_disabled/service_disabled.pass.sh
```diff
@@ -6,3 +6,4 @@ yum -y install avahi
 
 systemctl disable avahi-daemon.service
 systemctl disable avahi-daemon.socket
+systemctl mask avahi-daemon
```
matejak: We miss the `systemctl stop` here.

ggbecker: Fixed

shared/templates/template_ANSIBLE_service_disabled
```diff
@@ -4,19 +4,18 @@
...
+  systemd:
+    name: "{{{ DAEMONNAME }}}.{{ item }}"
     enabled: "no"
     state: "stopped"
+{{%- if MASK_SERVICE %}}
+    masked: "yes"
+{{%- endif %}}
   register: service_result
-  failed_when: "service_result is failed and ('Could not find the requested service' not in service_result.msg)"
+  failed_when: "service_result is failed and ('Could not find the requested service' not in service_result.msg) and ('Unit file {{{ DAEMONNAME }}}.{{ item }} does not exist' not in service_result.msg)"
```
ggbecker: @matejak Unfortunately it fails on stopping when the socket file does not exist causing the ansible to fail. I tried to find some function in the module to check this condition, but couldn't find anything. Do you have any suggestions?

ggbecker:
I have updated the ansible remediation and now it checks if the unit file (`*.socket` or `*.service`) exists before trying to stop/disable/mask.

I also have updated the OVAL check to the systemd service attribute `FragmentPath` instead of `UnitFileState` which is more reliable and now it checks if points to path equals `/dev/null`.

Tests scenarios executed:
- RHEL7:
  - Tested with service_avahi-daemon_disabled using C2S profile with both BASH and Ansible remediation.
- RHEL8:
  - Tested with service_debug-shell_disabled using OSPP profile with both BASH and Ansible remediation.

I consider again ready to review. Thanks for the requests.

ggbecker:
Note: The services disabled template tries to disable both the service and socket. In some cases, the socket might not exist. To handle this case, the remediation check if the unit file exist before disabling. The ideal would be to solve this in the template level where one could select the disabling of the socket if it is known that such service also uses socket. But change this behavior would cause too many changes as each csv template file for each product would need to be updated.

matejak:
Great, the applicable tests finally pass, so I am happy to merge this.

PR #4636, opened 2019-07-18
New rule auditd_local_events

jan-cerny:
#### Description:
Checks if local_events = yes in /etc/audit/auditd.conf.

#### Rationale:
This rule is a part of RHEL 8 Common Criteria configuration.


linux_os/guide/system/auditing/auditd_local_events/oval/shared.xml
```diff
@@ -0,0 +1,6 @@
+{{{
+oval_auditd_config(
+    parameter="local_events",
+    value="yes"
```
ggbecker: Is this the default value? if so, then include missing_parameter_pass=true.

jan-cerny: Man page `man 5 auditd.conf` says default value is yes. But I don't know what happens if the option isn't present in config file at all.

ggbecker: Based on the source code of `audit` we can infer that even if the option does not exist in the file it will still have the default value set:
https://github.com/linux-audit/audit-userspace/blob/be5c84c19171dda26f565205aec38376b6d9834a/src/auditd-config.c#L306

IMO, it is safe to set `missing_parameter_pass=true`.

jan-cerny: @ggbecker Great finding! I'll add it.

jan-cerny: @ggbecker added

ggbecker: Great. Thanks. Resolving conversation.

shared/macros-ansible.jinja
```diff
@@ -162,3 +162,14 @@
...
+
+{{#
+  High level macro to set a command in auditd configuration file /etc/audit/auditd.conf.
+  Parameters:
+  - msg: the name for the Ansible task
+  - parameter: parameter to be set in the configuration file
+  - value: value of the parameter
+#}}
+{{%- macro ansible_auditd_set(msg='', parameter='', value='') %}}
+{{{ ansible_set_config_file(msg, "/etc/audit/auditd.conf", parameter=parameter, value=value, create="yes", separator=" = ", separator_regex="\s*=\s*") }}}
```
ggbecker: From: https://www.systutorials.com/docs/linux/man/5-auditd.conf/ - "All option names and values are case insensitive"
That means we should accept "YES" as valid, the same for parameter. Is that the case?
```suggestion
{{{ ansible_set_config_file(msg, "/etc/audit/auditd.conf", prefix_regex="^[\s]*(?i)", parameter=parameter, separator_regex='(?-i)[\s]+', value="(?i)"+value+"(?-i)", create="yes", separator=" = ", separator_regex="\s*=\s*") }}}
```

shared/macros-oval.jinja
```diff
@@ -206,6 +206,20 @@
...
+  This function can take five parameters:
+    - parameter (String): The parameter to be checked in the configuration file.
+    - value (String): The value to be checked. This can also be a regular expression (e.g: value1|value2 can match both values).
+    - missing_parameter_pass (boolean): If set, the check will also pass if the parameter is not present in the configuration file (default is applied).
+    - multi_value (boolean): If set, it means that the parameter can accept multiple values and the expected value must be present in the current list of values.
+    - missing_config_file_fail (boolean): If set, the check will fail if the configuration is not existent in the system.
+
+#}}
+{{%- macro oval_auditd_config(parameter='', value='', missing_parameter_pass=false, multi_value=false, missing_config_file_fail=false) %}}
+{{{ oval_check_config_file("/etc/audit/auditd.conf", prefix_regex="^\s*", parameter=parameter, separator_regex='\s*=\s*', value=value, missing_parameter_pass=missing_parameter_pass, application="auditd", multi_value=multi_value, missing_config_file_fail=missing_config_file_fail) }}}
```
ggbecker: From: https://www.systutorials.com/docs/linux/man/5-auditd.conf/ - "All option names and values are case insensitive"
That means we should accept "YES" as valid, the same for parameter. Is that the case?
```suggestion
{{{ oval_check_config_file("/etc/audit/auditd.conf", prefix_regex="^[\s]*(?i)", parameter=parameter, separator_regex='(?-i)[\s]+', value="(?i)"+value+"(?-i)", , missing_parameter_pass=missing_parameter_pass, application="auditd", multi_value=multi_value, missing_config_file_fail=missing_config_file_fail) }}}
```

jan-cerny: Yes, they are insensitive, but our macros don't have a parameter to switch case sensitivity,

jan-cerny:
@ggbecker I have added a new patch which adds case-intensity.

rhel8/profiles/ospp.profile
```diff
@@ -702,7 +702,8 @@ selections:
     ## Audit Configuration
     #################################################################
 
-    # TODO: local_events=YES
+    ## RHEL 8 CCE-82233-8: Include Local Events in Audit Logs
```
yuumasato: Let's not add the CCE numbers here.

tests/data/group_system/group_auditing/rule_auditd_local_events/missing_value.pass.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+# profiles = xccdf_org.ssgproject.content_profile_ospp
+sed -i "/local_events/d" "/etc/audit/auditd.conf"
```
yuumasato: Does the file exist by default?
In the test scenario I'd make sure it exists.

jan-cerny:
I have removed CCE from the OSPP profile and I have ensured /etc/audit/auditd.conf exists.

adelton:
The rule suffers from the following problem: https://github.com/ComplianceAsCode/content/issues/4645.

PR #4637, opened 2019-07-18
Fix problems with creating files in include_lineinfile.sh

jan-cerny:
#### Description:

1. Pass the value of create to lineinfile. If only_lineinfile is called with create=true then lineinfile was called with create=false, which leads to calling die on line 100, but we wanted to create the file.
2. Use correct value in create paramater because the code in function lineinfile accepts "true" not "yes".

#### Rationale:
Remediation that should create the file shouldn't fail if the file doesn't exist.

### DIFF
--- a/shared/bash_remediation_functions/include_lineinfile.sh
+++ b/shared/bash_remediation_functions/include_lineinfile.sh
@@ -97,12 +97,18 @@ function lineinfile() {
     local insert_before="$7"
     local insensitive="${8:-true}"
 
-    [ ! -e "$path" ] && [ "$create" != "true" ] && die "Path '$path' wasn't found on this system and not creating. Refusing to continue."
-    [ ! -e "$path" ] && [ "$create" == "true" ] && touch "$path"
-
-    if [ "$state" == "absent" ]; then
-        lineinfile_absent "$path" "$regex" "$insensitive"
+    if [ "$state" == "absent" ] ; then
+        if [ -e "$path" ] ; then
+            lineinfile_absent "$path" "$regex" "$insensitive"
+        fi
     elif [ "$state" == "present" ]; then
+        if [ ! -e "$path" ] ; then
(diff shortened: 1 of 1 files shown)


shared/bash_remediation_functions/include_lineinfile.sh
```diff
@@ -116,7 +116,7 @@ function only_lineinfile() {
     local insert_before="$6"
     local insensitive="$7"
 
-    lineinfile "$path" "$line_regex" '' 'false' 'absent' '' '' "$insensitive"
+    lineinfile "$path" "$line_regex" '' "$create" 'absent' '' '' "$insensitive"
```
yuumasato: Isn't it strange to pass `create=true` and `state=absent`? It will create an empty file just to check that something is not there.
If the state is `absent` it makes sense that the file is not created.
What if instead of passing `$create` with `absent`, the function `lineinfile` in line 100, also checked `if state == "present"`?

jan-cerny:
@yuumasato I have slightly reworked it, hopefully it's more understandable now.

shared/bash_remediation_functions/include_lineinfile.sh
```diff
@@ -116,7 +122,7 @@ function only_lineinfile() {
     local insert_before="$6"
     local insensitive="$7"
 
-    lineinfile "$path" "$line_regex" '' 'false' 'absent' '' '' "$insensitive"
+    lineinfile "$path" "$line_regex" '' "false" 'absent' '' '' "$insensitive"
```
jan-cerny: Hmmm, this is probably unintended, I'll change this change.

jan-cerny:
@yuumasato I have removed the unwanted change of quotes.

jan-cerny:
@openscap-ci test this please

yuumasato:
LGTM, thank you!

PR #4643, opened 2019-07-23
Use sshd_config_set function for sshd remediations

mildas:
#### Description:
Remediate sshd rules with `sshd_config_set` function, which is based on setting values for parameters with `set_config_file` function.

#### Rationale:
Remediation functions for sshd rules were inconsistent.


matejak:
We have decided to go for a template in #4661, so after that PR is merged, you can adapt yours to use the template functionality.

mildas:
I will update it (remove templated remediations) after merging #4667

ggbecker:
@mildas, does the #4667 do everything this PR does? If yes, then we can close this PR. Thanks

mildas:
@ggbecker yes, because of changes in bash remediations, templates etc. we can close it.

PR #4646, opened 2019-07-23
Fix regex which looks for line in file configuration.

### LINKED ISSUE #4645: oscap xccdf eval  --rule xccdf_org.ssgproject.content_rule_auditd_local_events does not work
#### Description of problem:

When `local_events = yes` line is followed by `local_events = no`, `xccdf_org.ssgproject.content_rule_auditd_local_events` still reports pass.

Upstream master, f22dc360663418ff0856664490e7341487f66b8f.

RHEL 8.0.

#### Steps to Reproduce:

1. `./build_product rhel8`
2. To `/etc/audit/auditd.conf`, append lines
```
local_events = yes
local_events = no
```
3. `systemctl restart auditd`
4. `tail -f /var/log/audit/audit.log &`
5. `ssh root@localhost date`
6. `oscap xccdf eval --profile ospp --rule xccdf_org.ssgproject.content_rule_auditd_local_events build/ssg-rhel8-ds.xml`

#### Actual Results:

```
Title   Include Local Events in Audit Logs
Rule    xccdf_org.ssgproject.content_rule_auditd_local_events
Ident   CCE-82233-8
Result  pass
```

#### Expected Results:

```
Title   Include Local Events in Audit Logs
Rule    xccdf_org.ssgproject.content_rule_auditd_local_events
Ident   CCE-82233-8
Result  fail
```

There is no output produced by that `tail -f` because the last `local_events = no` wins.

The `xccdf_org.ssgproject.content_rule_auditd_local_events` should not say that `local_events` is enabled when it only looked at one line but missed the line that overrides the value.

ggbecker:
#### Description:

- The regex is now capable of finding multiple occurrences and only
considers the last occurrence.

Inspired by https://github.com/ComplianceAsCode/content/pull/4430/files#diff-693acc0bf8562e378862c734b05544a6R27 Thanks @adelton

#### Rationale:

- Only the last occurrence of a parameter should be taken into account when looking for line in file.

- Fixes #4645

### DIFF
--- a/shared/macros-oval.jinja
+++ b/shared/macros-oval.jinja
@@ -150,7 +150,7 @@
 #}}
 {{%- set regex = prefix_regex+parameter+separator_regex -%}}
 {{%- else %}}
-{{%- set regex = prefix_regex+parameter+separator_regex+"(.*)[\s]*(?:|(?:#.*))?$" -%}}
+{{%- set regex = ".*(?:\\n\s*[^[\s].*)*\s*"+prefix_regex+parameter+separator_regex+"(.*)[\s]*(?:|(?:#.*))?$" -%}}
 {{%- endif %}}
 {{%- endif %}}
   <ind:textfilecontent54_object id="obj_{{{ rule_id }}}{{{ suffix_id }}}" version="1">
--- a/tests/data/group_system/group_auditing/rule_auditd_local_events/multiple_values.fail.sh
+++ b/tests/data/group_system/group_auditing/rule_auditd_local_events/multiple_values.fail.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+# profiles = xccdf_org.ssgproject.content_profile_ospp
+echo "local_events = yes" > "/etc/audit/auditd.conf"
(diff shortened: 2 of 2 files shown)


shared/macros-oval.jinja
```diff
@@ -150,7 +150,7 @@
 #}}
 {{%- set regex = prefix_regex+parameter+separator_regex -%}}
 {{%- else %}}
-{{%- set regex = prefix_regex+parameter+separator_regex+"(.*)[\s]*(?:|(?:#.*))?$" -%}}
+{{%- set regex = ".*(?:\\n\s*[^[\s].*)*\s*"+prefix_regex+parameter+separator_regex+"(.*)[\s]*(?:|(?:#.*))?$" -%}}
```
yuumasato: This seems to complex, it deserves a comment.

yuumasato: I have proposed an alternate fix, https://github.com/ComplianceAsCode/content/pull/4647 :)

ggbecker:
I'm in favor of the fix proposed by #4647, it's simpler and fix the problem in a OVAL level and not in using regex tricks.

Closing this one.

PR #4647, opened 2019-07-23
Macro OVAL lineinfile to collect all objects, and make sure only one exists.

### LINKED ISSUE #4645: oscap xccdf eval  --rule xccdf_org.ssgproject.content_rule_auditd_local_events does not work

yuumasato:
#### Description:

- Update test to look for only one object
- Update object to return all objects found

#### Rationale:

- The test, as it was, was passing if all objects existed. And the object defined that only one object would be returned. Thus, with multiple `parameter = value` assignments, only one of them was considered, probably the first.

- Fixes #4645

### DIFF
--- a/shared/macros-oval.jinja
+++ b/shared/macros-oval.jinja
@@ -156,7 +156,7 @@
   <ind:textfilecontent54_object id="obj_{{{ rule_id }}}{{{ suffix_id }}}" version="1">
     <ind:filepath>{{{ path }}}</ind:filepath>
     <ind:pattern operation="pattern match">{{{ regex }}}</ind:pattern>
-    <ind:instance datatype="int">1</ind:instance>
+    <ind:instance operation="greater than or equal" datatype="int">1</ind:instance>
   </ind:textfilecontent54_object>
 {{%- endmacro %}}
 
--- a/tests/data/group_system/group_auditing/rule_auditd_local_events/double_assignment.fail.sh
+++ b/tests/data/group_system/group_auditing/rule_auditd_local_events/double_assignment.fail.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+# profiles = xccdf_org.ssgproject.content_profile_ospp
+echo "local_events = yes" >> "/etc/audit/auditd.conf"
+echo "local_events = no" >> "/etc/audit/auditd.conf"


yuumasato:
I'm not sure if it is worth while to have a note in the rule explicitly stating that any multiple assignment will fail the scan.
Taking the issue in question, for example, a `/etc/audit/auditd.conf` file with
```
local_events = no
local_events = yes
```
or 
```
local_events = yes
local_events = yes
```
will fail the scan.
Technically it should pass, but it is not a desirable state.

yuumasato:
Tested for some rules that are already using the macros: https://paste.fedoraproject.org/paste/83SV5djA-T88cZi87znQlw
Nothing seems to be broken by this patch.

jan-cerny:
@yuumasato Shouldn't it be that all occurrences of the object conform to the state?

matejak:
As the remediation is safely able to restore the "exactly one occurrence", it is absolutely harmless to fail on this one. This strict approach ensures that there is no mess in those config files, so it makes things easier for us.
I would not see an advanced behavior as a no-go, but this is a new implementation, and we should IMO focus on having something that works reliably first, before proposing additional enhancements.

yuumasato:
> @yuumasato Shouldn't it be that all occurrences of the object conform to the state?

@jan-cerny The quantity of objects that should conform to the state is determined by `check`, not by `check_existence`.
[checkenumeration](https://github.com/OVALProject/Language/blob/master/docs/oval-common-schema.md#---checkenumeration---)

`check_existence` determines how many objects should exist in the system.
[existenceenumeration](https://github.com/OVALProject/Language/blob/master/docs/oval-common-schema.md#---existenceenumeration---)

yuumasato:
> @yuumasato Shouldn't it be that all occurrences of the object conform to the state?

Hmm, I guess the main problem was that the `object` was only collecting the first `instance` match.
The change of `check_existence = "only_one_exists"` may not be necessary for the fix.

jan-cerny:
@yuumasato 

> Hmm, I guess the main problem was that the object was only collecting the first instance match.

Yes, I fully agree.

check_existence="only_one_exists" causes that if the option is present multiple times and each time it is set to a correct value then it returns a false positive.

I believe it we should try check_existence="at_least_one_exists" . The following test scenario fails if 
we expect result pass and check_existence="only_one_exists", but passes if we expect result pass and  check_existence="at_least_one_exists".

```
#!/bin/bash
# profiles = xccdf_org.ssgproject.content_profile_ospp
echo "local_events = yes" >> "/etc/audit/auditd.conf"
echo "local_events = yes" >> "/etc/audit/auditd.conf"
echo "local_events = yes" >> "/etc/audit/auditd.conf"
```

@matejak 

> This strict approach ensures that there is no mess in those config files, so it makes things easier for us.

I don't fully understand. If you have the following config file:

```
local_events = yes
local_events = no
local_events = yes
local_events = no
local_events = no
```
In my opinion it should find all occurrences of local_events and check if they're all set correctly to "yes". If you check only the first occurrence of local_events, which is "yes" in this case, you report a pass, but in fact it's overwritten by the last "no".

adelton:
How easy is it to check only the last object in OVAL?

If it's not easy, you can always use a prefix to the regexp as `.*` or something similar to skip to any previous content and find only the last line.

jan-cerny:
> How easy is it to check only the last object in OVAL?

It this it isn't possible. OVAL allows us to check specific instance of regex match, but the problem is that you don't know in advance how many matches will be there, so you don't know the index of the last instance.

adelton:
Right. That's a pity.

Could you try the approach with making the actual regexp match the last occurence of that key?

yuumasato:
> check_existence="only_one_exists" causes that if the option is present multiple times and each time it is set to a correct value then it returns a false positive.

Yes, this is technically correct. But is there a reason to keep multiple identical assignments in the config file?
It is easier to check compliance status if the file is not messy.

Anyway, I dropped 8547a556c636a36c1577e7ce3e25c6c5fed5950e, just so this fix can get in.
The issue will be fixed.

Further discussion can follow regarding matching first or last.

jan-cerny:
It fixes one of failing tests in #4660. Thanks!

PR #4648, opened 2019-07-24
Compare suid/sgid files with the RPM database

jan-cerny:
#### Description:
It is difficult to maintain the list to list paths of all possible suid and sgid binaries in a Linux distribution. Instead, we can check if the suid or sgid file is owned by an RPM package by consulting the RPM database. Another advantage of this solution is that we can have a single OVAL for all RPM-related Linux distributions.  The patch modifies OVAL for rules file_permissions_unauthorized_suid and
file_permissions_unauthorized_sgid and also adds test scenarios for these rules.

#### Rationale:
Some files were reported as offending the rule even if they were owned by an RPM because they were not present in the whitelist.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1693026


shawndwells:
For the US, the requirement is no SUID/SGID. Coming from an RPM or not does not matter. Can this be updated to look for all files, vs only those from RPM?

yuumasato:
> For the US, the requirement is no SUID/SGID. Coming from an RPM or not does not matter.

@shawndwells  This seems fit into a new rule, where no sticky bit is expected.

This rule is expects their presence, and checks the sticky bits against an "authorized" source (the vendor).

jan-cerny:
@yuumasato 

The current rule description contains: 

"A straightforward means of identifying unauthorized SGID files is determine if any were not installed as part of an RPM package, which is cryptographically verified. Investigate the origin of any unpackaged SGID files.".

My understanding was that if the SGID binary comes from an RPM it is expected to not be reported as a finding. Before this PR, the OVAL contained a whitelist, commented as `  <!-- list of all setgid files included with base RHEL6,RHEL7,OL7 system -->` . I though the intention of this whitelist is to enumerate all the SGID binaries coming from an RPM.

The PR changes the code so that we don't have to maintain the whitelist. Unfortunately, the new solution also allows binaries that come from third-party RPMs.

Who defines which binaries are allowed?

jan-cerny:
We will need to check what the profiles require.
The rules are included in these profiles:
```
 gg file_permissions_unauthorized_s[gu]id *.profile | cut -f1 -d: |sort -u
ol7/profiles/standard.profile
ol8/profiles/standard.profile
rhel6/profiles/C2S.profile
rhel6/profiles/CSCF-RHEL6-MLS.profile
rhel6/profiles/standard.profile
rhel6/profiles/usgcb-rhel6-server.profile
rhel7/profiles/anssi_nt28_enhanced.profile
rhel7/profiles/C2S.profile
rhel7/profiles/standard.profile
rhel8/profiles/standard.profile

```

The BZ https://bugzilla.redhat.com/show_bug.cgi?id=1693026 is about RHEL 7 STIG but the rules aren't present in STIG profile anymore.

shawndwells:
> The PR changes the code so that we don't have to maintain the whitelist. Unfortunately, the new solution also allows binaries that come from third-party RPMs.
> 
> Who defines which binaries are allowed?

Something how only admins can install packages, and the packages must be signed by an organizationally-recognized key, and how YUM enforces that key signing on all ``yum install`` commands.

Even if we can't force the native ``rpm`` command to use gpg checks, it will still require an admin to install things.

Seems legit enough. We cannot protect against malicious insider. If someone has sudo perms to install packages we have to assume that individual can make decisions on what is "allowed."

shawndwells:
This is a clever and clean way not to maintain the whitelist anymore. Good idea!

yuumasato:
Cool, so it seems we have consensus that the authorized status is determined by the RPM package.
This looks good to be merged.

@jan-cerny You wanted to add something else?

jan-cerny:
@yuumasato I wanted to check what the profiles say, but I haven't started with that yet.

shawndwells:
Because this PR slightly changes the behavior of the evaluation, could a note be added into XCCDF describing the logic of the check?

e.g.
``This configuration check whitelists {SUID SGID} files which were installed via RPM. It is assumed that when an individual has sudo access to install an RPM, combined with CCE-XXXX requiring all packages to be signed with an organizationally-recognized GPG key, the software should be considered an approved package on the system. 

Any {SUID SGID} file not deployed through an RPM will be flagged for further review.``

linux_os/guide/system/permissions/files/file_permissions_unauthorized_sgid/rule.yml
```diff
@@ -8,6 +8,11 @@ description: |-
     unauthorized SGID files is determine if any were not installed as part of an
     RPM package, which is cryptographically verified. Investigate the origin
     of any unpackaged SGID files.
+    This configuration check whitelists SGID files which were installed via RPM.
```
yuumasato: I suggest changing 'whitelists' to 'considers authorized'.
Pleaes, do the same change in the other rule description.

jan-cerny:
@yuumasato it isn't done and it isn't clear to me as well

matejak:
What is the status of this PR? It looks like that the heavy lifting has been done by @jan-cerny, and now the more subtle issues have been uncovered. I see two directions for solutions:

- The PR is merged (as it presents a huge improvement over the previous state) and an issue is reported against the rule about the "what does it mean authorized", so it is tracked at least somewhere.
- We make a new rule from this check (SU/SGID binaries come from RPMs), and we then reuse its check in the "Authorized" rule in connection with some blacklist in form of a variable.

jan-cerny:
I face a problem that my solution reports `/usr/sbin/mount.nfs` as a SUID file that isn't owned by any RPM, but actually it is owned by `nfs-utils`.

I guess it's related to fact that `/sbin` directory is a symlink to `/usr/sbin`.  And when I list files in `nfs-utils` package I get the path from `/sbin`:

```
rpm -ql  nfs-utils | grep 'sbin/mount.nfs$'
/sbin/mount.nfs
```

I'm going to investigate further.

jan-cerny:
I was thinking about excluding "/sbin" and "/bin" because they're only symlinks to "/usr/sbin" and "/usr/bin", but then I realized that it won't help because RPMs can use either path with "/usr" or without "/usr".

jan-cerny:
I also checked using rpminfo instead of rpmverifyfile.

But again, for this rpminfo object I get only `/sbin/mount.nfs`

```
 26     <lin-def:rpminfo_object id="oval:x:obj:1" version="1" comment="Object">
 27         <lin-def:behaviors filepaths="true"/>
 28         <lin-def:name>nfs-utils</lin-def:name>
 29     </lin-def:rpminfo_object>
```

jan-cerny:
This OVAL needs a fix in OpenSCAP rpmverifyfile probe to work properly with symlinks. The patch for OpenSCAP has been submitted in https://github.com/OpenSCAP/openscap/pull/1412

ggbecker:
> This OVAL needs a fix in OpenSCAP rpmverifyfile probe to work properly with symlinks. The patch for OpenSCAP has been submitted in [OpenSCAP/openscap#1412](https://github.com/OpenSCAP/openscap/pull/1412)

I've tested this PR against the one from OpenSCAP and it works as expected.

PR #4649, opened 2019-07-24
Add new rule auditd_write_logs

jan-cerny:
#### Description:
Checks write_logs = yes in /etc/audit/auditd.conf.

#### Rationale:

This rule is a part of RHEL 8 CC configuration.


yuumasato:
Hmm, good point.
By templating `lineinfile` we can have benefit of testing by template.
And as I see it coming already, I'd rather investing extending test coverage for `auditd_local_events` rather than adding new, repeated ones here.

yuumasato:
So nothing left on this PR, merging!

PR #4650, opened 2019-07-24
Disable storing core dumps.

adelton:
#### Description:

- Disable storing core dumps.

#### Rationale:

- A core dump includes a memory image taken at the time the operating system terminates an application. The memory image could contain sensitive data and is generally useful only for developers trying to debug problems.


adelton:
Technically the coredump.conf is an ini file, not a plain config file. However, there is only one section, `[Coredump]` there, so we won't be looking for the value in wrong section. However, if there is no section in that file, me might match and/or add the value outside of the expected section.

Is that a problem?

shared/macros-ansible.jinja
```diff
@@ -173,3 +173,14 @@
...
+
+{{#
+  High level macro to set a parameter in /etc/systemd/coredump.conf.
+  Parameters:
+  - msg: the name for the Ansible task
+  - parameter: parameter to be set in the configuration file
+  - value: value of the parameter
+#}}
+{{%- macro ansible_coredump_config_set(msg='', parameter='', value='') %}}
+{{{ ansible_set_config_file(msg, "/etc/systemd/coredump.conf", parameter=parameter, value=value, create="yes", separator="=", separator_regex="\s*=\s*") }}}
```
ggbecker: Unfortunately this macro is not able to handle `ini` configuration file. Ansible provides a module to handle ini files:
https://docs.ansible.com/ansible/latest/modules/ini_file_module.html

In the project it is being used in some cases:
https://github.com/ComplianceAsCode/content/blob/3ea35b6431430e81fdf9303fc58598454999b3ca/linux_os/guide/services/sssd/sssd_offline_cred_expiration/ansible/shared.yml#L13-L20

We would need to create a new macro wrapping this `ini_file` from ansible. For now it is ok to leave like that.

shared/bash_remediation_functions/include_lineinfile.sh
```diff
@@ -156,3 +156,10 @@ function auditd_config_set() {
 
     set_config_file "/etc/audit/auditd.conf" "$parameter" "$value" "true" "" "" "true" " = " "\s*=\s*"
 }
+
+function coredump_config_set() {
+    local parameter="$1"
+    local value="$2"
+
+    set_config_file "/etc/systemd/coredump.conf" "$parameter" "$value" "true" "" "" "true" "=" "\s*=\s*"
```
ggbecker: Bash `lineinfile` functions also don't support `ini` configuration files. It is ok to leave like that.

adelton:
I've renamed the rules, added the `section` to OVAL, added the tests -> 4cf02c2c49661e5823d7bf7efca18ea19f86e3f9.

ggbecker:
> I think we should treat it like an `ini` file, and mention correct section in `rule.yml`
> 
> By default `/etc/systemd/coredump.conf` already exists and `[Coredump]` section is there.
> For this rules, we can state that the supported use case is remediation from default installation state, and remediation when file doesn't exist or doesn't have `[Coredump]` section is unsupported.
> I suggest adding a general warning in the `rule.yml`, describing supported use case.

To illustrate the `warnings` attribute:

https://github.com/ComplianceAsCode/content/blob/f7ca4ca04f13665db7ebb069a731f55658fe0f82/chromium/guide/chromium/chromium_policy_file/rule.yml#L20-L24

yuumasato:
I also suggest creating an issue to track that this rule needs to be updated to use `ini` capabilities when ready.

adelton:
I've now rebased on master and added `service_systemd-coredump_disabled` -> a0ad45d4af70e20a8e9cb8f5cb3c90b0966fe5e9.

adelton:
Added an explicit commit to add the bits about `[Coredump]` section and to also not create the file when it is missing (as we would not add that section line) -> bb59ff9b628c70e6369e74a10f455213c7852681.

yuumasato:
Thanks for the changes.

PR #4653, opened 2019-07-25
Document --url and --extra-repo of install_vm.py script

matusmarhefka:
#### Description:

- Document possibility to install VM with custom repositories using `install_vm.py` script.


yuumasato:
LGTM, thanks!

PR #4655, opened 2019-07-25
List build outputs in Developer Manual

yuumasato:
#### Description:

- List build outputs in Developer Manual


jan-cerny:
Thank you for documenting this!

PR #4656, opened 2019-07-25
Testing of set_config_file function with BATS

mildas:
#### Description:
Add basic test scenarios for `set_config_file` function with BATS.
Also includes it in ctest if `bats` package was successfully loaded.


matejak:
Closing, as this PR reincarnated as #4659

PR #4657, opened 2019-07-25
Test suite autodetect datastream.

matejak:
The frequent use case is that a single datastream is built in the build directory, and one wants it to be tested by the test suite.
This change add autodetection capabilities, so the --datastream option can be left out. If it is left out and there is not a single datastream in the build dir, an error is displayed instead.

As a side-effect, the `--datastream` option won't be required any longer, which is [generally recommended](https://docs.python.org/3/library/argparse.html#required).

> Required options are generally considered bad form because users expect options to be optional, and thus they should be avoided when possible.

### DIFF
--- a/tests/test_suite.py
+++ b/tests/test_suite.py
@@ -7,6 +7,7 @@ import os
 import os.path
 import time
 import sys
+from glob import glob
 
 ssg_dir = os.path.join(os.path.dirname(__file__), "..")
 sys.path.append(ssg_dir)
@@ -43,7 +44,6 @@ def parse_args():
     common_parser.add_argument("--datastream",
                                dest="datastream",
                                metavar="DATASTREAM",
-                               required=True,
                                help=("Path to the Source DataStream on this "
                                      "machine which is going to be tested"))
     benchmarks = common_parser.add_mutually_exclusive_group()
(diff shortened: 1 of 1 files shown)


tests/test_suite.py
```diff
@@ -216,10 +216,32 @@ def auto_select_xccdf_id(datastream, bench_number):
...
+    datastreams = glob(os.path.sep.join(build_dir_path + [ds_glob]))
+    return datastreams
+
+
+def get_unique_datastream():
+    datastreams = get_datastreams()
+    if len(datastreams) == 1:
+        return datastreams[0]
+    msg = ("Unable to auto-detect the datastream, "
+           "expected to find a single one in the build dir, but")
```
yuumasato: I wouldn't say that it is expected that there is only single datastream in build dir, I have othen more than one.

tests/test_suite.py
```diff
@@ -216,10 +216,32 @@ def auto_select_xccdf_id(datastream, bench_number):
...
+    if not datastreams:
+        raise RuntimeError(msg + " found none.")
+    raise RuntimeError(
+        msg + " found multiple ones: {0}".format(datastreams))
```
yuumasato: If multiple DataStreams are found, the tool should recommend use of `--datastream` option.

matejak:
Thanks for the feedback, comments are addressed.

tests/test_suite.py
```diff
@@ -218,21 +218,23 @@ def auto_select_xccdf_id(datastream, bench_number):
...
-    msg = ("Unable to auto-detect the datastream, "
-           "expected to find a single one in the build dir, but")
+    msg = ("Autodetection of the datastream file is possible only when there is "
+           "a single one in the build dir, but")
     if not datastreams:
-        raise RuntimeError(msg + " found none.")
+        raise RuntimeError(msg + " there is none.")
     raise RuntimeError(
-        msg + " found multiple ones: {0}".format(datastreams))
+        msg + " there are {0} of them. Use the --datastream option to select "
```
yuumasato: The line can get rather long:
```shell
ERROR - Error occurred during options normalization: Autodetection of the datastream file is possible only when there is a single one in the build dir, but there are 28 of them. Use the --datastream option to select e.g. ['../build/ssg-debian8-ds.xml', '../build/ssg-rhel7-ds.xml', '../build/ssg-wrlinux8-ds.xml', '../build/ssg-wrlinux1019-ds.xml', '../build/ssg-ubuntu1404-ds.xml', '../build/ssg-ol8-ds.xml', '../build/ssg-centos6-ds.xml', '../build/ssg-fuse6-ds.xml', '../build/ssg-centos7-ds.xml', '../build/ssg-eap6-ds.xml', '../build/ssg-ubuntu1604-ds.xml', '../build/ssg-ubuntu1804-ds.xml', '../build/ssg-rhv4-ds.xml', '../build/ssg-rhosp13-ds.xml', '../build/ssg-chromium-ds.xml', '../build/ssg-fedora-ds.xml', '../build/ssg-ol7-ds.xml', '../build/ssg-sl6-ds.xml', '../build/ssg-rhel8-ds.xml', '../build/ssg-sle12-ds.xml', '../build/ssg-firefox-ds.xml', '../build/ssg-sl7-ds.xml', '../build/ssg-sle11-ds.xml', '../build/ssg-rhel6-ds.xml', '../build/ssg-opensuse-ds.xml', '../build/ssg-jre-ds.xml', '../build/ssg-ocp3-ds.xml', '../build/ssg-centos8-ds.xml']
```
And in the terminal it actually wraps up.
I like that they short and easy to copy-paste when adding `--datastream` option.

PR #4658, opened 2019-07-25
Fixed an omission.

matejak:
The undefined `script` variable inside the method is an obvious omission.


yuumasato:
Good catch! Thanks for the fix.

PR #4659, opened 2019-07-26
Testing of set_config_file function with BATS 2

matejak:
This PR is a successor to #4656 

Differences:

- The cmake setup is simpler, analogous to e.g. `linkchecker` setup.
- The test that fails has been removed, so the PR can be merged. Another tests can be added later.


mildas:
@matejak I forgot commit 1 small fix for set_config_file which probably causes failing of that one test scenario.
The problem is [here](https://github.com/ComplianceAsCode/content/blob/master/shared/bash_remediation_functions/include_lineinfile.sh#L21). `lineinfile_absent` function is called with 3 arguments, so there's no 8th argument. The change should be easy `"${8:-true}"` -> `"${3:-true}"`. Can you please try it, if with this change the test will pass so we can use it? (sorry, I can't push commit now, otherwise I'd create PR with fix)

matejak:
@mildas Great point, I have added that fix to the PR.

tests/CMakeLists.txt
```diff
@@ -48,3 +48,10 @@ add_test(
     NAME "machine-only-rules"
     COMMAND env "PYTHONPATH=$ENV{PYTHONPATH}" "${PYTHON_EXECUTABLE}" "${CMAKE_CURRENT_SOURCE_DIR}/test_machine_only_rules.py" --source_dir "${CMAKE_SOURCE_DIR}" --build_dir "${CMAKE_BINARY_DIR}"
 )
+
+if (SSG_BATS_TESTS_ENABLED AND BATS_EXECUTABLE)
+    add_test(
+        NAME "bash-unit-tests"
```
mildas: Test name should be more specific, because if we are going to add test scenarios for other functions, this name will be misleading.

ggbecker: I think it is okay to leave like, but instead of running only one test in the `COMMAND` option, run tests for a directory as described here: https://github.com/sstephenson/bats#test-suites. So basically it should be something like moving tests to a folder called `unit/bats/` and setting:
`COMMAND "${BATS_EXECUTABLE}" "${CMAKE_CURRENT_SOURCE_DIR}/unit/bats/"`

matejak: Good point, I am changing the target to a directory.

matejak:
Seems to work:

```
       Start  60: bash-unit-tests
59/308 Test  #60: bash-unit-tests ...........................................................   Passed    0.36 sec
```

tests/test_set_config_file.bats
```diff
@@ -0,0 +1,118 @@
...
+
+    set_config_file "$tmp_file" "Compression" "no"
+
+    run diff "$tmp_file" <(printf "$expected_output")
+    [ "$status" -eq 0 ]
+
+    rm "$tmp_file"
+}
+
+@test "No remediation happend" {
```
ggbecker:
```suggestion
@test "No remediation happened" {
```

tests/CMakeLists.txt
ggbecker: What do you think of moving the tests to a specific folder? if the amount of `bats` files increase the `tests` folder will be polluted. I suggest that you create a directory called `tests/unit/bats` and place the tests there.

tests/test_set_config_file.bats
```diff
@@ -0,0 +1,118 @@
+# Source set_config_file function
+. $BATS_TEST_DIRNAME/../shared/bash_remediation_functions/include_lineinfile.sh
```
ggbecker: This would need to be updated if the tested is moved to another directory.
CMake could export the source dir variable to be used within the tests, and as a fallback use this relative path to find the remediation file

matejak: Yes, but why not? If something changes, failing tests themselves will alert the user that they require an update. Tests should IMO be executable easily, without ctest, just by calling `bats`.
The absolute path to remediation functions dir is known at configure-time, but embedding it to the script at that time looks too much of a hassle.

tests/test_set_config_file.bats
matejak: fixed by rebase.

tests/test_set_config_file.bats
ggbecker: I'm ok with the explanation. Solving the conversation.

ggbecker:
LGTM, thanks for the patch.

PR #4660, opened 2019-07-26
Introduce tests for macro that generates OVAL

jan-cerny:
#### Description:

The test contains test cases to test macro `oval_check_config_file` which we use to generate OVAL definitions. To test the macro, use `OVALTester.test()` method which takes the following parameters: short description, OVAL document as a string, content of checked configuration file and the expected OVAL result.

~Now the test scenarios don't pass because macro  `oval_check_config_file` doesn't create OVAL which would handle the following situations:~
1. ~Parameter and value are correct, but there is trailing white space at the end of line. Generated OVAL should evaluate as true, but it evaluates as false.~
2. ~Parameter and value are correct, but there is a comment at the end of line.  Generated OVAL should evaluate as true, but it evaluates as false.~
3. ~There are multiple occurrences of the parameter, but only the first have a correct value.  Generated OVAL should evaluate as false, but it evaluates as true. (fixed in #4647)~

EDIT: the macro has been updated to make the test cases pass.

#### Rationale:
We are going to use the Jinja macro `oval_check_config_file` very often everywhere, therefore we need to be sure that the macro behaves correctly.


jan-cerny:
I suspect that the regex that matches the value and the rest of the line is wrong.

See what is appended to the regex:

 https://github.com/ComplianceAsCode/content/blob/d7ef32885ee0dcd564b1848c0b25f427ab77fe09/shared/macros-oval.jinja#L153

In `(.*)[\s]*(?:|(?:#.*))?$` the `.*` is greedy which means the group matches the whole line, including training whitespace and comments.

At the first sight it looks like it should be `([^#\s]+)` but that doesn't solve the problems:
1. configuration file uses something different to comment than `#`
2. configuration file allows quoted strings which contain white-space inside
3. configuration file allows multiple values of the same parameter separated by a white-space
4. configuration file uses `=` or other nonwhite-space character as a separator and the spaces are allowed or needed inside the value as a part of the value

ggbecker:
It looks like the this part `[\s]*(?:|(?:#.*))?$` does nothing with the regex because of capture group `(.*)` from:
 https://github.com/ComplianceAsCode/content/blob/d7ef32885ee0dcd564b1848c0b25f427ab77fe09/shared/macros-oval.jinja#L153

One different approach we could take to solve this issue would to keep capturing everything and match the expected value under state test, this would avoid constructing crazy capturing group regexs.

jan-cerny:
rebased on the top of current master

jan-cerny:
@ggbecker 
> It looks like the this part [\s]*(?:|(?:#.*))?$ does nothing with the regex because of capture group (.*) from:

Yes, I agree, that's basically the same that I meant in my previous comment when I said that the star is greedy.

> One different approach we could take to solve this issue is to keep capturing everything and match the expected value under state test, this would avoid constructing crazy capturing group regexs.

That's a great idea. But I think that you just move the problem from the regex that is used in object to the regex that is used in state. Anyway you will have to create a regex that filters out the comments and whitespaces or handles it any way. 

Construction of the state regex as `^.*value.*$` is not an option because that allows values as substrings of values, which will lead to false results.

jan-cerny:
> Construction of the state regex as ^.*value.*$ is not an option because that allows values as substrings of values, which will lead to false results.

There is probably also another bug in the current implementation of multi_value=true:

```
{{%- if multi_value %}}
{{%- set regex = "^.*"+value+".*$" -%}}
```
Consider the following line in /etc/ssh/sshd_config
```
MACs hmac-sha2-512,hmac-sha2-256,hmac-sha1,hmac-sha1-etm@openssh.com,hmac-sha2-256-etm@openssh.com,hmac-sha2-512-etm@openssh.com
```
What would match?

jan-cerny:
I have added many new test cases, including cases that test INI-style file. 

Based on the test scenarios we have found out that our regular expressions aren't correct. We have to make the regular expressionmore precise. We will not allow new lines as separators
and prefixes. We will also check word boundaries when we use multi_value parameter set to true. We will avoid matching trailing whitespaces and trailing comments to prevent false results.

I suggest reviewing if the test scenarios make sense.

tests/test_macros_oval.py
```diff
@@ -0,0 +1,1042 @@
...
+            path='CONFIG_FILE',
+            section="vehicle",
+            parameter='speed',
+            value='100',
+            missing_parameter_pass=false,
+            application='',
+            multi_value=false,
+            missing_config_file_fail=false,
+        ) }}}""",
+        "[vehicle]\nspeed =\n100",
```
matejak: You can use [dedent](https://docs.python.org/3.6/library/textwrap.html#textwrap.dedent) to preserve nice input while removing leading whitespace.

tests/test_macros_oval.py
```diff
@@ -0,0 +1,1042 @@
...
+            tmp_dir = tempfile.mkdtemp()
+            config_file_path = self._create_config_file(
+                config_file_content, tmp_dir)
+            oval_path = self._create_oval(
+                oval_content, config_file_path, tmp_dir)
+            result, results_path = self._evaluate_oval(oval_path)
+            msg = ("OVAL Definition was evaluated as %s, but expected result "
+                   "is %s.") % (result, expected_result)
+            assert result == expected_result, msg
+            print("Test: %s: PASS" % (description))
```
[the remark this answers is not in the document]
jan-cerny: @matejak For me it shows the output only if `--verbose` is provided. Am I wrong?

jan-cerny:
@matejak I have extracted the test to a separate file and I have called dedent. I will provide an option to disable output when the script is run separately outside of ctest. 

I suggest focusing on reviewing the test cases now.

jan-cerny:
@matejak I have added a verbose option.

Please tell me which of the test cases you like and which of the test cases you don't like.

tests/oval_tester.py
```diff
@@ -0,0 +1,153 @@
...
+        config_file_content: content of the text configuration file that the
+                             OVAL will check
+        expected_result: expected result of evaluation of the OVAL definition
+        """
+        try:
+            tmp_dir = tempfile.mkdtemp()
+            config_file_path = self._create_config_file(
+                config_file_content, tmp_dir)
+            oval_path = self._create_oval(
+                textwrap.dedent(oval_content), config_file_path, tmp_dir)
```
matejak: I have just this dedent-related remark - please dedent test cases only where they are defined. That way, one can get rid of those `\n` in strings.
The problem with dedenting inside the framework is that one may forget about it, and then the test case definition will look differently from what is actually tested.

jan-cerny:
@matejak I have removed the commit that added dedent.

matejak:
Great, thanks again!

PR #4661, opened 2019-07-26
Add line in file template for sshd and auditd configuration files

ggbecker:
#### Description:

- Add line in file template for sshd and auditd configuration files. The template makes usage of line in file macros for OVAL, Bash and Ansible.

#### Rationale:

- Introduce template for similar line in file configuration to diminish boiler plate work.

Additional notes:
Both templates are pretty similar, they practically change only the high level macro function which is being called.
Also, `rule_title` is being populated with only `rule_id` information. It would require to load the corresponding `rule.yml` file in order to use the actual rule attributes, such as rule title, `prodtype`, etc.


shared/templates/create_auditd_lineinfile.py
```diff
@@ -0,0 +1,78 @@
...
+#
+# create_auditd_lineinfile.py
+#   automatically generate checks for auditd line in file configuration
+
+import sys
+
+from template_common import FilesGenerator, UnknownTargetError, CSVLineError
+
+
+class AuditdLineinfileGenerator(FilesGenerator):
```
mildas: What about creating super class e.g. `LineinfileGenerator`, and these Auditd/sshd classes would inherit from it? Because the first part from `generate` (try/catch block) will be same for every lineinfile generator, right?

shared/templates/create_auditd_lineinfile.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python2
```
mildas: Why `python2`?

ggbecker: Honestly, it was copied from `create_services_enabled.py` so it came from there. It is probably due to some RHEL6 support, but I'm not sure. Looking at `create_*.py` files, some of them use `#!/usr/bin/env python2`, some of them use `#!/usr/bin/python2` and some of them don't have the shebang at all. It is probably safe to remove.

shared/templates/create_auditd_lineinfile.py
ggbecker: That sounds interesting. There will be some differences but I think it is still worthy to have the super class. I'll think of something here.

mildas: Great, thank you!

ggbecker: I've updated the PR, please check it.

shared/templates/create_auditd_lineinfile.py
ggbecker: I removed from the file `create_lineinfile.py` which is the super class file for both `auditd` and `sshd` templates. Marking as resolved.

shared/templates/create_lineinfile.py
```diff
@@ -0,0 +1,94 @@
+
```
mildas: empty line

shared/templates/csv/auditd_lineinfile.csv
```diff
@@ -0,0 +1,2 @@
+# <rule_id, parameter, value[, missing_parameter_pass]> (missing_parameter_pass: If set, the OVAL check will also pass if the parameter is not present in the configuration file. Values can be true or false)
```
mildas: Can you please add, that false is default value? (also for sshd description)

shared/templates/csv/sshd_lineinfile.csv
```diff
@@ -0,0 +1,2 @@
+# <rule_id, parameter, value[, missing_parameter_pass]> (missing_parameter_pass: If set, the OVAL check will also pass if the parameter is not present in the configuration file. Values can be true or false)
+# sshd_enable_strictmodes,StrictModes,yes,true
```
mildas: Was it intention to comment this? I see you didn't delete oval, bash and ansible for the rule, but just to be sure.

shared/templates/template_OVAL_auditd_lineinfile
```diff
@@ -0,0 +1,7 @@
+{{{
+oval_auditd_config(
+	parameter=PARAMETER,
+	value=VALUE,
+	missing_parameter_pass=MISSING_PARAMETER_PASS
+)
```
mildas: trailing space

shared/templates/template_ANSIBLE_sshd_lineinfile
```diff
@@ -3,9 +3,9 @@
 # strategy = restrict
 # complexity = low
 # disruption = low
-{{{
-ansible_auditd_set(
-   parameter="local_events",
-   value="yes"
-)
+{{{
```
mildas: trailing space

shared/templates/template_OVAL_sshd_lineinfile
```diff
@@ -0,0 +1,7 @@
+{{{
+oval_sshd_config(
+	parameter=PARAMETER,
+	value=VALUE,
+	missing_parameter_pass=MISSING_PARAMETER_PASS
+)
```
mildas: trailing space

shared/templates/template_ANSIBLE_auditd_lineinfile
```diff
@@ -0,0 +1,11 @@
+# platform = multi_platform_all
+# reboot = false
+# strategy = restrict
+# complexity = low
+# disruption = low
+{{{
```
mildas: trailing space

shared/templates/create_lineinfile.py
ggbecker: fixed

shared/templates/template_OVAL_auditd_lineinfile
ggbecker: fixed

shared/templates/template_OVAL_sshd_lineinfile
ggbecker: fixed

shared/templates/template_ANSIBLE_auditd_lineinfile
ggbecker: fixed

shared/templates/template_ANSIBLE_sshd_lineinfile
ggbecker: fixed

shared/templates/csv/sshd_lineinfile.csv
ggbecker: It was supposed to serve as example doc. I added doc explaining that it's an example of usage.

shared/templates/csv/auditd_lineinfile.csv
ggbecker: Added for both `csv` files.

mildas:
Thank you

PR #4662, opened 2019-07-26
OSPP requirements and selections

yuumasato:
#### Description:

- Update OSPP profile with regards to requirements and rule selections.


rhel8/profiles/ospp.profile
```diff
@@ -1091,6 +1053,10 @@ selections:
     ## Enable FIPS Mode
     - enable_fips_mode
 
+    ## Set up Crypto policy
+    - var_system_crypto_policy=fips
+    - configure_crypto_policy
```
mildas: Not needed, because it's already there - #4676

PR #4663, opened 2019-07-29
Harden sshd crypto policy

vojtapolasek:
#### Description:

Added a new rule to harden SSHD by applying stricter crypto policy.

#### Rationale:

Athering to the CC requirements, based on Kickstart file.


yuumasato:
@openscap-ci add to whitelist

linux_os/guide/system/software/integrity/crypto/harden_sshd_crypto_policy/rule.yml
```diff
@@ -0,0 +1,20 @@
...
+    This can be done by dropping a file named <tt>opensshserver-xxx.config</tt>, replacing <tt>xxx</tt> with arbitrary identifier, into <tt>/etc/crypto-policies/local.d</tt>. This has to be followed by running <tt>update-crypto-policies</tt> sothat changes are applied.
+    Changes are propagated into <tt>/etc/crypto-policies/back-ends/opensshserver.config</tt>. This rule checks if this file contains predefined <tt>CRYPTO_POLICY</tt> environment variable configured with predefined value.
+
+rationale: |-
+    The Common Criteria requirements specify that certain parameters for OpenSSH Server are configured e.g. supported ciphers, accepted host key algorithms, public key types, key exchange algorithms, HMACs and GSSAPI key exchange is disabled.
+
+severity: medium
+
+identifiers:
+    cce@rhel8: 82881-4
```
yuumasato: Also remove the CCE value from the `shared/references/cce-redhat-avail.txt` file.

linux_os/guide/system/software/integrity/crypto/harden_sshd_crypto_policy/bash/shared.sh
```diff
@@ -0,0 +1,10 @@
+# platform = multi_platform_fedora,Red Hat Enterprise Linux 8,Oracle Linux 8
+
+
+cp="CRYPTO_POLICY='-oCiphers=aes128-ctr,aes256-ctr,aes128-cbc,aes256-cbc -oMACs=hmac-sha2-256,hmac-sha2-512 -oGSSAPIKeyExchange=no -oKexAlgorithms=diffie-hellman-group14-sha1,ecdh-sha2-nistp256,ecdh-sha2-nistp384,ecdh-sha2-nistp521 -oHostKeyAlgorithms=rsa-sha2-256,ecdsa-sha2-nistp256,ecdsa-sha2-nistp256-cert-v01@openssh.com,ecdsa-sha2-nistp384,ecdsa-sha2-nistp384-cert-v01@openssh.com,rsa-sha2-512,ecdsa-sha2-nistp521,ecdsa-sha2-nistp521-cert-v01@openssh.com -oPubkeyAcceptedKeyTypes=ssh-rsa,ecdsa-sha2-nistp256,ecdsa-sha2-nistp384'"
+file=/etc/crypto-policies/local.d/opensshserver-ospp.config
+
+#blank line at the begining to ease later readibility
+echo '' > $file
+echo $cp >> $file
+update-crypto-policies
```
yuumasato: Please, make sure there is a newline at the end of the files.

linux_os/guide/system/software/integrity/crypto/harden_sshd_crypto_policy/rule.yml
```diff
@@ -0,0 +1,20 @@
...
+title: 'Harden SSHD Crypto Policy'
+
+description: |-
+    Crypto Policies are means of enforcing certain cryptographic settings for selected applications including OpenSSH server.
+    The SSHD service is by default configured to enhance its configuration based on currently configured Crypto-Policy. However, in certain cases it is required to override the Crypto Policy specific to OpenSSH Server.
+    This can be done by dropping a file named <tt>opensshserver-xxx.config</tt>, replacing <tt>xxx</tt> with arbitrary identifier, into <tt>/etc/crypto-policies/local.d</tt>. This has to be followed by running <tt>update-crypto-policies</tt> sothat changes are applied.
+    Changes are propagated into <tt>/etc/crypto-policies/back-ends/opensshserver.config</tt>. This rule checks if this file contains predefined <tt>CRYPTO_POLICY</tt> environment variable configured with predefined value.
+
+rationale: |-
+    The Common Criteria requirements specify that certain parameters for OpenSSH Server are configured e.g. supported ciphers, accepted host key algorithms, public key types, key exchange algorithms, HMACs and GSSAPI key exchange is disabled.
```
yuumasato: I think the rationale for this rule is that algorithms required by Common Criteria are stricter than what is set in any current crypto policy.

linux_os/guide/system/software/integrity/crypto/harden_sshd_crypto_policy/rule.yml
```diff
@@ -0,0 +1,20 @@
...
+
+severity: medium
+
+identifiers:
+    cce@rhel8: 82881-4
+
+references:
+    OSPP:FCS_SSHS_EXT.1
```
yuumasato: The key should be all lower case: `ospp`. Also add a space between key and value.

linux_os/guide/system/software/integrity/crypto/harden_sshd_crypto_policy/rule.yml
```diff
@@ -0,0 +1,20 @@
+documentation_complete: true
+
+title: 'Harden SSHD Crypto Policy'
+
+description: |-
+    Crypto Policies are means of enforcing certain cryptographic settings for selected applications including OpenSSH server.
+    The SSHD service is by default configured to enhance its configuration based on currently configured Crypto-Policy. However, in certain cases it is required to override the Crypto Policy specific to OpenSSH Server.
```
yuumasato: `The SSHD service is by default configured to enhance its configuration based on currently configured Crypto-Policy.` I'm not sure I understand this thoroughly.
Mayby `The SSHD service is by default configured to extend its configuration ...` is what you mean?

vojtapolasek: The configuration does not need to be extended necessarily in terms of being nlarged, I changed the wording to "modified".

matejak:
test this please

linux_os/guide/system/software/integrity/crypto/harden_sshd_crypto_policy/rule.yml
```diff
@@ -0,0 +1,28 @@
+documentation_complete: true
+
+title: 'Harden SSHD Crypto Policy'
+
+description: |-
+    Crypto Policies are means of enforcing certain cryptographic settings for selected applications including OpenSSH server.
+    The SSHD service is by default configured to modify its configuration based on currently configured Crypto-Policy. However, in certain cases it might be needed to override the Crypto Policy specific to OpenSSH Server and leave rest of the Crypto Policy in tact.
```
yuumasato: Typo:`intact`

linux_os/guide/system/software/integrity/crypto/harden_sshd_crypto_policy/rule.yml
```diff
@@ -0,0 +1,28 @@
...
+    This can be done by dropping a file named <tt>opensshserver-xxx.config</tt>, replacing <tt>xxx</tt> with arbitrary identifier, into <tt>/etc/crypto-policies/local.d</tt>. This has to be followed by running <tt>update-crypto-policies</tt> sothat changes are applied.
```
yuumasato: Typo: `so that`

linux_os/guide/system/software/integrity/crypto/harden_sshd_crypto_policy/rule.yml
yuumasato: Suggestion for text: `However, it may be necessary to customize the Crypto Policy only for OpenSSH Server.`

linux_os/guide/system/software/integrity/crypto/harden_sshd_crypto_policy/rule.yml
```diff
@@ -0,0 +1,28 @@
...
+    cce@rhel8: 82176-9
+
+references:
+    ospp : FCS_SSHS_EXT.1
+
+ocil_clause: 'Crypto Policy for OpenSSH Server is not configured according to CC requirements'
+
+ocil: |-
+    To verify if the OpenSSH server uses defined Crypto Policy, run:
+    <pre>$ grep 'CRYPTO_POLICY' /etc/crypto-policies/back-ends/opensshserver.txt | tail -n 1</pre>
```
yuumasato: This should be: `opensshserver.config`

rhel8/profiles/ospp.profile
```diff
@@ -1226,3 +1226,5 @@ selections:
 
     ## Enable dnf-automatic Timer
     - timer_dnf-automatic_enabled
+
+    - harden_sshd_crypto_policy
```
yuumasato: Please, don't forget about newline at end of file.

linux_os/guide/system/software/integrity/crypto/harden_sshd_crypto_policy/rule.yml
```diff
@@ -0,0 +1,28 @@
...
+
+title: 'Harden SSHD Crypto Policy'
+
+description: |-
+    Crypto Policies are means of enforcing certain cryptographic settings for selected applications including OpenSSH server.
+    The SSHD service is by default configured to modify its configuration based on currently configured Crypto-Policy. However, in certain cases it might be needed to override the Crypto Policy specific to OpenSSH Server and leave rest of the Crypto Policy in tact.
+    This can be done by dropping a file named <tt>opensshserver-xxx.config</tt>, replacing <tt>xxx</tt> with arbitrary identifier, into <tt>/etc/crypto-policies/local.d</tt>. This has to be followed by running <tt>update-crypto-policies</tt> sothat changes are applied.
+    Changes are propagated into <tt>/etc/crypto-policies/back-ends/opensshserver.config</tt>. This rule checks if this file contains predefined <tt>CRYPTO_POLICY</tt> environment variable configured with predefined value.
+
+rationale: |-
```
yuumasato: I would say : `...compared to any existing Crypto Policy.`

yuumasato:
@vojtapolasek Thank you!

ggbecker:
Customizing the crypto policy file for `opensshserver` will require changes in our `enable_fips_mode` OVAL check as it tests the file if it is a symlink to the default file as you can see here:

https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/software/integrity/fips/enable_fips_mode/oval/shared.xml#L16
https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/oval/shared.xml#L59

Otherwise the `enable_fips_mode` rule will always fail when this new rule is selected.

@yuumasato what do you think about this?

adelton:
@ggbecker, yes, you are right.

yuumasato:
> Customizing the crypto policy file for `opensshserver` will require changes in our `enable_fips_mode` OVAL check as it tests the file if it is a symlink to the default file as you can see here:

Yes, customization of crypto policies fails the check for those symlinks in `configure_crypto_policy`.

We can consider dropping all the symlink checks, not being symlinked to a back-end policy does not imply malicious tampering of the policy.

adelton:
Alternatively, the check could be for the symlink, *or* the drop-in snippet file we know we support.

vojtapolasek:
@adelton  I agree with this, I find is as the safest option. However, in this case we need to track somewhere what policies we modify. Just saying.

yuumasato:
> the drop-in snippet file we know we support

How will the rule know that what is in the drop-in snippet is OK?
This seems out of scope of what `configure_crypto_policy` should be doing.

If a rule adds a drop-in replacement for a back-end, shouldn't that rule check it?

adelton:
> How will the rule know that what is in the drop-in snippet is OK?
It won't. But it will know that the snippet is allowed, and thus the check for the symlink does not make sense.

> This seems out of scope of what `configure_crypto_policy` should be doing.
> 
> If a rule adds a drop-in replacement for a back-end, shouldn't that rule check it?
Well, it does, functionally. The `harden_sshd_crypto_policy` rule checks if the result in `/etc/crypto-policies/back-ends/opensshserver.config` (be it symlink or file) has the content it expects. So it checks the ultimate result that the sshd will use.

PR #4664, opened 2019-07-29
Quick fix for CombinedMode _modify_parameters()

yuumasato:
#### Description:
- Update interface of `CombinedMode` method  `_modify_parameter()` to  match its parent's.

#### Rationale:
- `CombinedMode` inherits `_get_scenarios()` from RuleChecker but doesn't override it.
-  `RuleChecker`'s method  `_get_scenarios()` calls `_modify_parameters()` with two arguments.


matejak:
Good catch, this is a completion of #4658.

yuumasato:
Credits should go to @matusmarhefka, :)

PR #4665, opened 2019-07-29
Manual page create_module(2) says that this system call is present only in kernels before Linux 2.6.

adelton:
#### Description:

- The `create_module` syscal should not appear on any recent system and thus it does not need to be checked in the auditd rules.

#### Rationale:

- Manual page create_module(2) says that this system call is present only in kernels before Linux 2.6.
- RHEL 6 shipped with kernel 2.6.32, so it did not have that syscall, and neither do newer versions of RHEL or other operating systems.


yuumasato:
Hello, @tedbrunell 

Can you look into removing checks for syscal `create_module` from the STIG?
From RHEL6, this `syscal` is not present in the kernel. 
Syscal Ref: [create_module(2)](http://man7.org/linux/man-pages/man2/create_module.2.html)
kernel version ref: [rhel6](https://access.redhat.com/articles/3078#RHEL6)

Thank you.

comps:
Note that the syscall is still allocated in the kernel;
```
arch/x86/entry/syscalls/syscall_32.tbl:127      i386    create_module
arch/x86/entry/syscalls/syscall_64.tbl:174      64      create_module
```
(and similar)

Which means it's also recognized by `ausyscall`:
```
$ ausyscall create_module
create_module      174
```
But I couldn't find any linked code to it (`SYSCALL_DEFINE*()` or otherwise), so it probably is dead.

shawndwells:
On 7/29/19 11:54 AM, Jiri Jaburek wrote:
>
> Note that the syscall is still allocated in the kernel;
>
> |arch/x86/entry/syscalls/syscall_32.tbl:127 i386 create_module 
> arch/x86/entry/syscalls/syscall_64.tbl:174 64 create_module |
>
> (and similar)
>
> Which means it's also recognized by |ausyscall|:
>
> |$ ausyscall create_module create_module 174 |
>
> But I couldn't find any linked code to it (|SYSCALL_DEFINE*()| or 
> otherwise), so it probably is dead.
>

If a 3rd party application makes the create_module call, would it work 
on RHEL 7 or RHEL 8?

comps:
```c
#include <unistd.h>
#include <syscall.h>

int main()
{
    syscall(__NR_create_module, "abcd", 12345);
    return 0;
}
```
(to bypass glibc)
```
$ strace ./a.out 
...
create_module(0x402010, 0x3039)         = -1 ENOSYS (Function not implemented)
...
```
No.

shawndwells:
> ```c
> #include <unistd.h>
> #include <syscall.h>
> 
> int main()
> {
>     syscall(__NR_create_module, "abcd", 12345);
>     return 0;
> }
> ```
> 
> (to bypass glibc)
> 
> ```
> $ strace ./a.out 
> ...
> create_module(0x402010, 0x3039)         = -1 ENOSYS (Function not implemented)
> ...
> ```
> 
> No.

Thorough. Thanks! Looks like the call was already removed in the upstream audit.rules file:
https://github.com/linux-audit/audit-userspace/blob/master/rules/43-module-load.rules

Seems like this can be removed from the profiles.

adelton:
> Just don't edit file `shared/references/disa-stig-rhel7-v2r3-xccdf-manual.xml`, this file is the official reference file from DOD released in [STIG V2R3](https://dl.dod.cyber.mil/wp-content/uploads/stigs/zip/U_Red_Hat_Enterprise_Linux_7_V2R3_STIG.zip).
> The overlay file should not be changed as well.

Removed those two from the commit -> fba624ecbef4f7377dfe4b1d0d4b807aa8f0f279.

yuumasato:
@adelton thank you!

PR #4666, opened 2019-07-29
Semi automate release

yuumasato:
#### Description:

- Scripts to semi-automate release of content.
  - This is heavily inspired by OpenSCAP's release scripts

### DIFF
--- a/release_tools/README.adoc
+++ b/release_tools/README.adoc
@@ -1,0 +1,92 @@
+= How to perform a semi-automatic release =
+
+=== Prerequisities ===
+
+* System with `python3.6` or greater, installed https://pypi.python.org/pypi/PyGithub/1.35[PyGithub] module.
+* A GitHub token - get one on https://github.com/settings/tokens. Choose `repo` scope.
+* Your Jenkins user - your User ID should be the same as your GitHub handle, to be sure, check your profile page in Jenkins.
+* A Jenkins token - gnerate a token on Configure page in Jenkins
+
+=== Prep ===
+
+* Clone a clean `content` repository:
+
+   git clone https://github.com/ComplianceAsCode/content.git
+
+* Create `.env` file
++
+Create an `.env` file in the `release_tools` directory (i.e. where the release scripts are). +
+This will contain sensitive information, s.a. GitHub access tokens. +
+Put your GitHub token in the `.env` file, so it contains the following line: +
+
+   GITHUB_TOKEN='<your github token here>'
+   JENKINS_USER='<you jenkins user ID>'
+   JENKINS_TOKEN='<you jenkins token here>'
+
+[NOTE]
+====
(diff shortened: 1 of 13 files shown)


release_tools/README.adoc
```diff
@@ -0,0 +1,92 @@
+= How to perform a semi-automatic release =
+
+=== Prerequisities ===
+
+* System with `python3.6` or greater, installed https://pypi.python.org/pypi/PyGithub/1.35[PyGithub] module.
+* A GitHub token - get one on https://github.com/settings/tokens. Choose `repo` scope.
+* Your Jenkins user - your User ID should be the same as your GitHub handle, to be sure, check your profile page in Jenkins.
+* A Jenkins token - gnerate a token on Configure page in Jenkins
```
mildas: I'd say there's no need for indefinite article `A` in list of items (same for "**A** GitHub token").
+ misspelled `generate` (`gnerate`).

redhatrises:
Can we make this all a python script rather than mixing and matching between bash and python?

yuumasato:
> Can we make this all a python script rather than mixing and matching between bash and python?

Probably, I think the `cmake`, and  `git commands would still require bash snippets. But the python script could run it under the hood.

release_tools/release-utils.sh
```diff
@@ -0,0 +1,143 @@
...
+    echo :: Check rhel6-stig-ids.log for rules missing STIG IDs
+    (PYTHONPATH=.. python3 ../build-scripts/profile_tool.py stats -b ../build/ssg-rhel7-xccdf.xml --missing-stig-ids --profile stig) > rhel7-stig-ids.log
+    echo :: Check rhel7-stig-ids.log for rules missing STIG IDs
+
+}
+
+make_dist()
+{
+    if [ -e artifacts/scap-security-guide-$version.tar.bz2 ]; then
+        echo Tarball already exists
```
jan-cerny: If I understand it well this mean there is a leftover file with unknown contents. Do you think it should terminate the script? Or remove the file?

yuumasato: > If I understand it well this mean there is a leftover file with unknown contents.

Not necessarily. `make_dist()` is called by `build.sh`, which can be called multiple times while waiting for builds to finish. This is to check whether the tarball already exists, and avoid repeatedly generating it.

The tarball is removed when `cleanup.sh` is run, whole `artifacts/` directory is removed. See this [line](https://github.com/ComplianceAsCode/content/pull/4666/files#diff-184c7a1349c20ffd1f249962c52075a9R138).

> Do you think it should terminate the script? Or remove the file?

The `make_dist()` command can probably be moved to a different script.
I'll consider this when getting rid of the `.sh` scripts and everything is "pythonized".

shawndwells:
Closing due to approximately 4 months of inactivity. If you'd like to reopen and continue this work, please do!

PR #4667, opened 2019-07-30
Use of sshd_lineinfile template

mildas:
#### Description:
Start using sshd_lininfile template for sshd rules, that already used `oval_sshd_config` Jinja macro and deletes oval, bash and ansible folders for templated rules.

- Changes only rules, that are working with parameter and defined value, e.g. `HostbasedAuthentication no`.

- Doesn't change rules, that are working with external_variable ([sshd_set_mac_auth_tries](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/ssh/ssh_server/sshd_set_max_auth_tries/oval/shared.xml)) or regular expression ([sshd_use_approved_ciphers](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers/oval/shared.xml) although it was already using Jinja macro, because template can't handle regular expression in remediations).

Parameters, values and missing_parameter_pass were taken from OVALs.

#### Rationale:
Simplifies sshd rules definition and shows how to use sshd_lineinfile macro.


shared/templates/csv/sshd_lineinfile.csv
```diff
@@ -1,3 +1,18 @@
 # <rule_id, parameter, value[, missing_parameter_pass]> (missing_parameter_pass: If set, the OVAL check will also pass if the parameter is not present in the configuration file. Values can be true or false. Default value is false.)
 # example of usage:
```
ggbecker: You can remove the example of usage now as there will be multiple occurrences already that can be used as example

ggbecker:
@openscap-ci test this please

PR #4668, opened 2019-07-30
Remove watches since syscall rules cover all cases, and reconcile to single rule

adelton:
#### Description:

- Remove watches since syscall rules cover all cases.
- The existing rule audit_rules_kernel_module_loading covers the kernel module loading syscalls.

#### Rationale:

- Having both watches and syscall rules affects readability of the audit trail as there are two events for each operation.
- We use a single rule for the related syscalls to minimize the overhead of the audit system and of the compliance checking, as well as making the checks consistent.


adelton:
Similar change was made in upstream auditd: https://github.com/linux-audit/audit-userspace/commit/faba74ed367d5f8f2b6b5c8dcf6afca63a23c466.

adelton:
This will need to be rebased once https://github.com/ComplianceAsCode/content/pull/4665 is in.

adelton:
I merged all references, including `stigid@rhel7` -- not sure if that is correct.

shawndwells:
> I merged all references, including `stigid@rhel7` -- not sure if that is correct.

Each specific configuration that's being evalauted should be broken out into its own rule.

adelton:
Are you commenting on my `stigid@rhel7` question, or on the rule layout in general?

If it's about the rules, then the question is what forms a specific configuration. The current split to (say) `audit_rules_kernel_module_loading_init` and `audit_rules_kernel_module_loading_finit` leads to profiles only having one of those rules listed, which in turn leads to false sense of security for the admin when they see a rule there but in fact, the other API of loading the kernel modules is not covered. It just adds the extra work in maintaining the content when in reality, the user / admin will always want both enabled.

In a similar way, having separate rule for `audit_rules_kernel_module_loading_init` and `audit_rules_kernel_module_loading_delete` is just an extra complexity both for admins and for content authors, because without the module removal events, there is no real audit trail of what modules were loaded at any given time.

From this point of view, the configuration we are after is "kernel module loading and unloading is audited", even if it may be implemented by multiple audit rule lines for multiple syscalls.

And in fact, the single `audit_rules_kernel_module_loading` rule has already been there in the content, used by many profiles, so what we are doing here is really just unifying on that single rule for the kernel module manipulation audit.

redhatrises:
Kind nack.

There appears to be some confusion here on the differences of the all encompassing rule vs the finer grain rules. The `audit_rules_kernel_module_loading` rule is an older rule that is being used by profiles that don't follow CCE (1 check/configuration change per CCE identifier) guidelines, and really, should be phased out. `audit_rules_kernel_module_loading_finit` and `audit_rules_kernel_module_loading_init` for example are the newer rules that should be used over `audit_rules_kernel_module_loading` as they follow CCE guidelines and allow security departments to tailor to their security policy if needed.

yuumasato:
I would like to understand more this 1 check/configuration per CCE.

I couldn't find much info about CCEs. The [official page](https://nvd.nist.gov/config/cce) is quite shallow, unfortunately.
The [archive](https://cce.mitre.org/) has more stuff, ;), so pardon me, if I base myself on deprecated data.

I found some nice examples in [CCE List Editorial Policies](https://cce.mitre.org/lists/cce_list_editorialpolicies.html).

The rule could fit in [CD.5 Issue Decomposition (Split)](https://cce.mitre.org/lists/cce_list_editorialpolicies.html#cd5), more precisely, [example 5.4](https://cce.mitre.org/lists/cce_list_editorialpolicies.html#cd5.4).
Where it would be something of the sorts of:
```
+ The loading of kernel modules via syscal init_module should be logged 
+ The loading of kernel modules via syscal finit_module should be logged 
+ The unloading of kernel modules via syscal delete_module should be logged 
```

But I argue that decomposition of this rule actually makes it more similar to [CD.2 One Effect/Multiple Technical Mechanisms (Combine)](https://cce.mitre.org/lists/cce_list_editorialpolicies.html#cd2).
Where there are multiple mechanisms (syscals) to do the same thing (load/unload kernel moules).
```
+ The loading and unloading of kernel modules should be logged
```

Also, [CD.1 Effect vs. Technical Mechanism (Basic CD)](https://cce.mitre.org/lists/cce_list_editorialpolicies.html#cd1) promotes `configuration effect` over `technical mechanism`.

shawndwells:
On 8/5/19 1:21 PM, Watson Yuuma Sato wrote:
>
> I would like to understand more this 1 check/configuration per CCE.
>
> I couldn't find much info about CCEs. The official page 
> <https://nvd.nist.gov/config/cce> is quite shallow, unfortunately.
> The archive <https://cce.mitre.org/> has more stuff, ;), so pardon me, 
> if I base myself on deprecated data.
>
> I found some nice examples in CCE List Editorial Policies 
> <https://cce.mitre.org/lists/cce_list_editorialpolicies.html>.
>
> The rule could fit in CD.5 Issue Decomposition (Split) 
> <https://cce.mitre.org/lists/cce_list_editorialpolicies.html#cd5>, 
> more precisely, example 5.4 
> <https://cce.mitre.org/lists/cce_list_editorialpolicies.html#cd5.4>.
> Where it would be something of the sorts of:
>
> |+ The loading of kernel modules via syscal init_module should be 
> logged + The loading of kernel modules via syscal finit_module should 
> be logged + The unloading of kernel modules via syscal delete_module 
> should be logged |
>
> But I argue that decomposition of this rule actually makes it more 
> similar to CD.2 One Effect/Multiple Technical Mechanisms (Combine) 
> <https://cce.mitre.org/lists/cce_list_editorialpolicies.html#cd2>.
> Where there are multiple mechanisms (syscals) to do the same thing 
> (load/unload kernel moules).
>
> |+ The loading and unloading of kernel modules should be logged |
>
> Also, CD.1 Effect vs. Technical Mechanism (Basic CD) 
> <https://cce.mitre.org/lists/cce_list_editorialpolicies.html#cd1> 
> promotes |configuration effect| over |technical mechanism|.
>

It's a challenging situation.

Because we're mapping specific configuration actions to multiple 
standards (NIST 800-53, DoD SRG, PCI-DSS, HIPAA, ANSSI....), we need to 
be granular as possible.

Having "audit kernel modules unloads" be mapped to the same CCE that 
also covers "audit kernel loads" would be exceptionally confusing.

adelton:
> There appears to be some confusion here on the differences of the all encompassing rule vs the finer grain rules. The `audit_rules_kernel_module_loading` rule is an older rule that is being used by profiles that don't follow CCE (1 check/configuration change per CCE identifier) guidelines, and really, should be phased out. `audit_rules_kernel_module_loading_finit` and `audit_rules_kernel_module_loading_init` for example are the newer rules that should be used over `audit_rules_kernel_module_loading` as they follow CCE guidelines and allow security departments to tailor to their security policy if needed.

What is the point of distinguishing `audit_rules_kernel_module_loading_init` from `audit_rules_kernel_module_loading_finit`, with different rule and with different CEEs ids? I would assume that the goal is not to primarily audit the syscall invocation but to get an audit trail of modules that were loaded at a particular time. From that point of view, will the security department ever want to enable `audit_rules_kernel_module_loading_init` but not enable `audit_rules_kernel_module_loading_finit`?

Currently, `ol8/profiles/ospp.profile` lists `audit_rules_kernel_module_loading_init` but not `audit_rules_kernel_module_loading_finit`. Is that a deliberate decision or just an omission?

adelton:
Since the main discussion seems to revolve around single or multiple rules for the syscalls but not around the removal of the watches, I've not factored removal of the watches out to separate pull request: https://github.com/ComplianceAsCode/content/pull/4706. Hopefully that will allow us to move forward.

linux_os/guide/system/auditing/auditd_configure_rules/audit_kernel_module_loading/audit_rules_kernel_module_loading/rule.yml
```diff
@@ -56,18 +63,7 @@ references:
 
 ocil: |-
     {{{ ocil_audit_syscall(syscall="init_module") }}}
+    {{{ ocil_audit_syscall(syscall="finit_module") }}}
     {{{ ocil_audit_syscall(syscall="delete_module") }}}
```
shawndwells: This needs to be broken into multiple rules, one per syscall (init_module, finit_module, delete_module).

yuumasato: But this rule is aligning so nicely with item `Audit Kernel Module Loading and Unloading Events (Success/Failure) ` from https://github.com/ComplianceAsCode/content/pull/4680#issuecomment-521703084.

adelton:
With https://github.com/ComplianceAsCode/content/pull/4706 merged, I'm withdrawing this pull request.

PR #4670, opened 2019-08-01
add note about potential problem with nested VMX while running testsuite

vojtapolasek:
#### Description:

Added a paragraph describing solution when encountering error message about nested VMX while running test suite with Libvirt.

#### Rationale:

I encountered this problem on Arch Linux and I think it might be helpful for others upstream.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -64,34 +64,6 @@ $ python install_vm.py --domain test-suite-rhel8 --distro rhel8 --url http://bas
 in case the test suite breaks something and fails to revert. Do not use snapshot names starting with `ssg_`.\
 You can create a snapshot using `virsh` or `virt-manager`.
 
-*TIP*: It might happen, especially while using other distros than Fedora or RHEL (for example Arch), that you encounter the following error:
-
-```
-libvirt: QEMU Driver error : operation failed Failed to take snapshot: Error: Nested VMX virtualization does not support live migration yet
-```
-
-This error might be followed by Python tracebacks where the above message is repeated. First make sure that you are running the test suite on the physical machine and that you really DO NOT use nested virtualization by running VM in VM.
-
-If you pass this requirement, it might happen that nested virtualization is enabled for your KVM kernel module. Libvirt will refuse to do live migration in this case. You can check this by running:
-
-```
-$ systool -m kvm_intel -v | grep nested
-```
-
-If you see:
-
-```
-    nested              = "Y"
-```
-
-then the nested virtualization is enabled for the KVM kernel module and it needs to be disabled. In Arch Linux this can be done by putting 
-
-```
-options kvm_intel nested=1
-```
-
-into /etc/modprobe.d/modprobe.conf.
-
 ### Container backends
 
 There are 2 container backends supported, Podman and Docker.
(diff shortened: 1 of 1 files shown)


tests/README.md
```diff
@@ -150,6 +149,35 @@ Specify backend and image to use:
...
+If you see:
+
+```
+    nested              = "Y"
+```
+
+then the nested virtualization is enabled for the KVM kernel module and it needs to be disabled. In Arch Linux this can be done by putting 
+
+```
+options kvm_intel nested=1
```
mildas: By this you are enabling nested virtualization, so the sentence `...KVM kernel module and it needs to be disabled` contradicts with the command.

tests/README.md
```diff
@@ -150,6 +149,35 @@ Specify backend and image to use:
...
+```
+libvirt: QEMU Driver error : operation failed Failed to take snapshot: Error: Nested VMX virtualization does not support live migration yet
+```
+
+This error might be followed by Python tracebacks where the above message is repeated. First make sure that you are running the test suite on the physical machine and that you really DO NOT use nested virtualization by running VM in VM.
+
+If you pass this requirement, it might happen that nested virtualization is enabled for your KVM kernel module. Libvirt will refuse to do live migration in this case. You can check this by running:
+
+```
+$ systool -m kvm_intel -v | grep nested
```
mildas: Please add that this command is for Arch Linux and add how to find it on RHEL/Fedora - it should be achievable by `cat /sys/module/kvm_intel/parameters/nested` for Intel

vojtapolasek: It was valid for Fedora too, you just need to install sysfsutils, but I rewrote it to make it as generic as possible.

tests/README.md
```diff
@@ -150,6 +149,36 @@ Specify backend and image to use:
     - `hypervisor`: Typically, you will use the `qemu:///session`, or `qemu:///system`.
        It depends on where your VM resides.
     - `domain`: `libvirt` domain, which is basically name of the virtual machine.
+
+*NOTE*: It might happen, especially while using other distros than Fedora or RHEL (for example Arch), that you encounter the following error:
```
mildas: Can you add this whole part about nested virtualization error to Admonition block? You can find how to create Admontion block at https://asciidoctor.org/docs/asciidoc-writers-guide/#admonition-blocks
(I'm not sure about readability of asciidoc documentation for you, so if you have any problem, just let me know and I will help you with fix)

tests/README.md
```diff
@@ -150,6 +149,36 @@ Specify backend and image to use:
...
+```
+libvirt: QEMU Driver error : operation failed Failed to take snapshot: Error: Nested VMX virtualization does not support live migration yet
+```
+
+This error might be followed by Python tracebacks where the above message is repeated. First make sure that you are running the test suite on the physical machine and that you really DO NOT use nested virtualization by running VM in VM.
+
+If you pass this requirement, it might happen that nested virtualization is enabled for your KVM kernel module. Libvirt will refuse to do live migration in this case. You can check this by running:
+
+```
+$ cat /sys/modules/kvm_intel/parameters/nested
```
mildas: Wrong path to `nested` file. The path is `/sys/module/kvm_intel/parameters/nested`. It's under `module` folder not `modules`.

mildas:
Thank you

PR #4671, opened 2019-08-01
Clean up, finalize and rename rhel7/ccc profile to ospp

comps:
This hopefully finishes the ccc (ospp) profile, matching it against the kickstart.

Per email discussion with @shawndwells , I've also removed the old ospp* profiles and renamed the ccc one to just "ospp", matching the same profile name in rhel8.

It might be easier to review the changes on a per-commit basis so the removal/rename is more clearly split from the actual changes.


comps:
The tests are failing on
```
RuntimeError: Profile 'ospp42' is required to be in the 'RHEL-7' benchmark. It is a stable profile that can't be renamed or removed!
```
@shawndwells are you sure the profile should be removed?

yuumasato:
> The tests are failing on
> 
> ```
> RuntimeError: Profile 'ospp42' is required to be in the 'RHEL-7' benchmark. It is a stable profile that can't be renamed or removed!
> ```

The test is there to make sure the profile is not removed and causes problems downstream.

In this case, removal of the profile is for the benefit of the downstream users, so the test can be  updated.

https://github.com/ComplianceAsCode/content/blob/f7cc2908e41e727982f161b3070f39a07ed85013/tests/stable_profile_ids.py#L25

comps:
> Looks like most of the audit rules were removed, and should be added back in.

Please read the changeset more closely. The commit removes only lines that were commented out, not any active rules.

comps:
> > The tests are failing on
> > ```
> > RuntimeError: Profile 'ospp42' is required to be in the 'RHEL-7' benchmark. It is a stable profile that can't be renamed or removed!
> > ```
> 
> The test is there to make sure the profile is not removed and causes problems downstream.
> 
> In this case, removal of the profile is for the benefit of the downstream users, so the test can be updated.

OK, I've removed it from the list.

rhel7/profiles/ospp.profile
```diff
@@ -1,438 +1,183 @@
 documentation_complete: true
 
-title: 'United States Government Configuration Baseline'
+title: 'Protection Profile for General Purpose Operating Systems'
 
 description: |-
```
redhatrises: Please keep the description and just bump the OSPP version, or does the additional text fail NIAP validation?

rhel7/profiles/ospp.profile
```diff
@@ -1,438 +1,183 @@
...
-    security setting was chosen. Security Requirement Traceability Guides (RTMs) and
-    sample System Security Configuration Guides are provided via the
-    scap-security-guide-docs package.
-
-    This profile reflects U.S. Government consensus content and is developed through
-    the OpenSCAP/SCAP Security Guide initiative, championed by the National
-    Security Agency. Except for differences in formatting to accommodate
-    publishing processes, this profile mirrors OpenSCAP/SCAP Security Guide
-    content as minor divergences, such as bugfixes, work through the
-    consensus and release processes.
```
shawndwells: All of this language needs to stay.

comps:
* rebased
* ospp42 fixes squashed to the ospp42 removal commit
* ccc->ospp rename split from ospp42 removal
* ospp.profile now contains the rules from ccc.profile, but with title and description of the original usgcb ospp.profile

The original title+description I used was taken from rhel8/profiles/ospp.profile, but if that is somehow unsuitable for you, the current version keeps the original rhel7/profiles/ospp.profile wording.

comps:
This PR has been superseded by #4691 which contains only the Common Criteria relevant changes.

Any other profile shuffling or description decisions for `ospp` or `ospp42` can be made separately later.

PR #4674, opened 2019-08-02
Test suite: Set additional SSH options when testing ansible remediations

matusmarhefka:
#### Description:

- When using Ansible remediations, test suite should now consume additional options for SSH set in the `SSH_ADDITIONAL_OPTIONS` env variable.


yuumasato:
@matusmarhefka Does it work for you?
These options didn't work for me in the past and I had to use 
`ANSIBLE_HOST_KEY_CHECKING=False`. 
https://docs.ansible.com/ansible/latest/user_guide/intro_getting_started.html#host-key-checking

yuumasato:
Not sure which version of Ansible I was using though, [2.4](https://github.com/ansible/ansible/issues/31321) had a bug related to this.

yuumasato:
Testing it right now.

yuumasato:
It works.

PR #4675, opened 2019-08-02
Add new rhel8 aux gpg pubkey

ggbecker:
#### Description:

-  Add check for auxiliary rpm gpg key on RHEL8.

The auxiliary key is different from the one used in RHEL7 and therefore
has different version+release.

Also added a couple of test scenarios so it can be easily tested.

There might be different ways on how to implement the check, it could be handle at platform level, for example testing if the product is RHEL7 or RHEL8 and so on. I open to suggestions on improving the OVAL check.

#### Rationale:

- RHEL8 auxiliary key is different from the one used in RHEL7

### DIFF
--- a/linux_os/guide/system/software/updating/ensure_redhat_gpgkey_installed/oval/shared.xml
+++ b/linux_os/guide/system/software/updating/ensure_redhat_gpgkey_installed/oval/shared.xml
@@ -19,8 +19,12 @@
         </criteria>
         <criterion comment="package gpg-pubkey-fd431d51-4ae0493b is installed"
         test_ref="test_package_gpgkey-fd431d51-4ae0493b_installed" />
-        <criterion comment="package gpg-pubkey-2fa658e0-45700c69 is installed"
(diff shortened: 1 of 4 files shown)


yuumasato:
Ack.

@ggbecker A way to make the check simpler is to separate the key data from the logic.
Fedora check is like this, and the [macros](https://github.com/ComplianceAsCode/content/blob/9a5bf1ee098d78b0f495ab8e38cac77984508647/linux_os/guide/system/software/updating/ensure_fedora_gpgkey_installed/oval/shared.xml#L1) can be reused:
Check [logic](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/software/updating/ensure_fedora_gpgkey_installed/oval/shared.xml) and [data](https://github.com/ComplianceAsCode/content/blob/9a5bf1ee098d78b0f495ab8e38cac77984508647/fedora/product.yml#L14).

The simplification of the check can be done in subsequent PR.

PR #4676, opened 2019-08-05
Use crypto-policies to configure RHEL8 sshd algorithms

yuumasato:
#### Description:

- In RHEL8, configure SSHD Ciphers and MACs via system-wide crypto policies

#### Rationale:

- This is inline with done in `harden_sshd_crypto_policy` #4663

### DIFF
--- a/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers/rule.yml
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers/rule.yml
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: wrlinux8,wrlinux1019,rhel6,rhel7,rhel8,ol7,rhv4
+prodtype: wrlinux8,wrlinux1019,rhel6,rhel7,ol7,rhv4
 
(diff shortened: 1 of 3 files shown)


yuumasato:
See also: https://github.com/ComplianceAsCode/content/pull/3677

dahaic:
@vojtapolasek can you review this one? :)

yuumasato:
> Seems like more is needed. For example, how do we ensure the running system is using the FIPS ciphers? What provides assurance sshd is using crypto-policies?

@shawndwells I'm not sure what kind of assurance you are looking for.
I would say that setting a crypto-policy is very similar to setting a config file. In both cases, you set it, and the application will use it. It just happens that with crypto policies there is an extra layer that sets it for a bunch of libraries.

From [Chapter 3. Using system-wide cryptographic policies](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/security_hardening/using-the-system-wide-cryptographic-policies_security-hardening):
`
Once a system-wide policy is set up, applications in RHEL follow it and refuse to use algorithms and protocols that do not meet the policy, unless you explicitly request the application to do so.
`

matusmarhefka:
LGTM

PR #4679, opened 2019-08-05
Account for /tmp being on tmpfs in rhel8

comps:
(Same as 0296f371.)

/tmp is mounted as tmpfs by systemd (and will be explicitly remediated
anyway by the create_fstab_entry_if_needed code).

This fixes an anaconda oscap addon error/question that would stall
an automated install:

  Wrong configuration detected!
  /tmp must be on a separate partition or logical volume and has to be
  created in the partitioning layout before installation can occur with
  a security profile The installation should be aborted.
  Do you wish to continue anyway?
  Please respond 'yes' or 'no':


yuumasato:
Nice catch, thanks.

PR #4680, opened 2019-08-06
Single rule to configure audit rules for OSPP

yuumasato:
#### Description:

- Check and remediate required audit rules for OSPP profile as a whole.
- Select `audit_rules_for_ospp` in RHEL8
- Remove individual audit rules from RHEL8 OSPP
- Select `audit_rules_for_ospp` in RHEL7
  - This may cause conflicts with #4671
- What does `audit_rules_for_ospp` do?
  - Evaluates following files `10-base-config.rules`, `11-loginuids.rules`,`30-ospp-v42.rules` and `43-module-load.rules`,  "pre-configured rules".
  - Checks compares contents from "pre-configured rules" in `/etc/audit/rules.d` to what is in `/usr/share/doc/audit-VERSION/rules`.
  - Remediation copies "pre-configured rules"  from `/usr/share/doc/audit-VERSION/rules/` to `/etc/audit/rules.d/`.

#### Rationale:

- Individually checking and remediating the audit rules is a very complex task, considering that order and file in which they are affects their order, and thus, auditing behavior.


linux_os/guide/system/auditing/policy_rules/audit_rules_for_ospp/rule.yml
```diff
@@ -0,0 +1,52 @@
+documentation_complete: true
+
+prodtype: rhel7,rhel8
+
+title: 'Configure audit according to OSPP requirements'
+
+description: |-
+    Configure audit to meet requirements for Operating System Protection Profile (OSPP) v4.2.
+
+    Audit defines groups of rules in <tt>/usr/share/doc/audit-VERSION/rules</tt> to satisifay specific policies.
```
dahaic: typo, "satisifay"

linux_os/guide/system/auditing/policy_rules/audit_rules_for_ospp/rule.yml
```diff
@@ -0,0 +1,52 @@
+documentation_complete: true
+
+prodtype: rhel7,rhel8
+
+title: 'Configure audit according to OSPP requirements'
+
+description: |-
```
dahaic: Description could use the same hash treatment of OCIL, so from it is obvious what is expected. Even better would be version, but those files in audit package are not internally versioned, so we have to go with the hash for now.

linux_os/guide/system/auditing/policy_rules/audit_rules_for_ospp/bash/shared.sh
```diff
@@ -0,0 +1,310 @@
+# platform = Red Hat Enterprise Linux 7,Red Hat Enterprise Linux 8
+
+# let's start with a clean slate
+rm -f /etc/audit/rules.d/*
```
dahaic: We do not preserve any customization this way, but that's fine by me. It would be unnecessarily complex otherwise.

redhatrises:
100% against this change unless this is temporary. This makes tailoring impossible now (yes people will tailor OSPP).

dahaic:
> 100% against this change unless this is temporary. This makes tailoring impossible now (yes people will tailor OSPP).

Technical complexity of "per event" audit rule granularity is overwhelming. We can take a look at the possibility to allow tailoring of the hashes, or something of that sort. But I predict we won't reimplement OSPP requirements as they are to the old supergranular approach.

shawndwells:
Absolutely not. One configuration evaluation per rule. Cannot have a massive rule that checks all the audit rules. 

A webex/video conference may be better to talk through all this. Seeing many PRs lately which will have to be removed before OSPP is acceptable.

dahaic:
> Absolutely not. One configuration evaluation per rule. Cannot have a massive rule that checks all the audit rules.
> 
> A webex/video conference may be better to talk through all this. Seeing many PRs lately which will have to be removed before OSPP is acceptable.

Truly looking forward to it. Please schedule :)

dahaic:
@shawndwells @redhatrises Would it be acceptable, if each configuration action of OSPP configuration annex would have its separate file that we would then check as a bulk, via hash? So there would be something like `30-ospp-1-all-logons.rules` `30-ospp-2-file-object-unsuccessful.rules` etc.?

linux_os/guide/system/auditing/policy_rules/audit_rules_for_ospp/rule.yml
yuumasato: Fixed.

linux_os/guide/system/auditing/policy_rules/audit_rules_for_ospp/bash/shared.sh
```diff
@@ -0,0 +1,310 @@
...
+
+{{% if product == "rhel8" %}}
+# This is needed for RHEL8 until the file is updated
```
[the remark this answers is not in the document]
yuumasato: I have extended the comment to RHEL7 as well.

linux_os/guide/system/auditing/policy_rules/audit_rules_for_ospp/bash/shared.sh
yuumasato: Remediation doesn't remove everything now. There is no strict need for this. Other audit rules can co-exist with OSPP audit rules.

yuumasato:
> This makes tailoring impossible now (yes people will tailor OSPP).

Tailoring can be to remove `audit_rules_for_ospp`, and add the individual rules.
There could be a sample tailoring file with individual audit rules selected.

> @shawndwells @redhatrises Would it be acceptable, if each configuration action of OSPP configuration annex would have its separate file that we would then check as a bulk, via hash? So there would be something like `30-ospp-1-all-logons.rules` `30-ospp-2-file-object-unsuccessful.rules` etc.?

This would be similar to what #4668 is doing.

shawndwells:
On 8/7/19 6:42 AM, Watson Yuuma Sato wrote:
> +description: |-
> +    Configure audit to meet requirements for Operating System Protection Profile (OSPP) v4.2.


OSPP 4.2.1 was released. Has this PR been updated/reviewed against the 
now-current version?

Not sure if anything changed, but if the content aligns with 4.2.1, we 
should bump the version number in the description.

yuumasato:
Quoting our SME, from 4.2 to 4.2.1 there are no differences in auditing requirements.

yuumasato:
Hello @shawndwells 

As we had discussed last week splitting the audit rules according to NIAP's OSPP Configuration Annex seems reasonable:
- Audit All Logons (Success/Failure) and Logoffs (Successful)
- Audit File and Object Events (Unsuccessful)
- Audit User and Group Management Events (Success/Failure)
- Audit Privilege or Role Escalation Events (Success/Failure)
- Audit All Audit and Log Data Accesses (Success/Failure)
- Audit Cryptographic Verification of Software (Success/Failure)
- Audit Program Initiations (Success/Failure)
- Audit System Reboot, Restart, and Shutdown Events (Success/Failure)
- Audit Kernel Module Loading and Unloading Events (Success/Failure)

Each item above would have a rule for successful and unsuccessful events. That would mean around 18 rules to cover audit rules in OSPP profile.
I believe this will be a good step in granularity, improving manageability while maintaining reasonable control over the rules.

Upstream audit also plans to make similar steps and we can plan with them how to split the pre-configured audit rules for OSPP, and how we can use them for check and remediation.

Unfortunately this move will take time, and will have to be planned for next RHEL release.

We are in need of a rule to check audit rules for OSPP , so I'd like to push for this rule for now.
It can be dropped when file 30-ospp-v42.rules is split according to NIAPs configuration requirements.

yuumasato:
> For example, the kernel module and file watches should remain independent checks.

Sure, as an unreviewed  *draft* **preliminary** things could be broken down into:

- Audit All Logons (Success/Failure) and Logoffs (Successful)
  - Handled by pam
- Audit File and Object Events (Unsuccessful)
  - 30-ospp-v42.rules in:
    - unsuccessful-create
    - unsuccessful-modification
    - unsuccessful-access
    - unsuccessful-delete
    - unsuccessful-perm-change
- Audit User and Group Management Events (Success/Failure)
  - 30-ospp-v42.rules, in:
    - user-modify
    - group-modify keys
- Audit Privilege or Role Escalation Events (Success/Failure)
  - Handled by pam
- Audit All Audit and Log Data Accesses (Success/Failure)
- 30-ospp-v42.rules, in:
    - audit-trail
- Audit Cryptographic Verification of Software (Success/Failure)
  - Handled by rpm
- Audit Program Initiations (Success/Failure)
  - Handled by fapolicyd
- Audit System Reboot, Restart, and Shutdown Events (Success/Failure)
  - Handled by systemd
- Audit Kernel Module Loading and Unloading Events (Success/Failure)
  - 43-module-load.rules

> How do we break this apart to only check things that are dependent on order?

The critical ones *will* **probably** fall together in `Audit File and Object Events (Unsuccessful)`. So if they are treated together as a whole block, problem solved! :)

But I don't see these changes as part of this PR.

linux_os/guide/system/auditing/policy_rules/audit_rules_for_ospp/rule.yml
yuumasato: Now, description doesn't mention hashes. It suggests to remediate by copy the pre-configured audit rules from `/usr/share/doc/audit/rules/`. And to check by using diff command.

adelton:
Could you please rebase on the latest master?

yuumasato:
I have completely changed approach from using file hashes to actually comparing the file contents.
The remediation will not dump `30-ospp-v42.rule` into `/etc/audit/rules.d`, it will copy from `/usr/share/docs/audit-VERSION/rules/`.

adelton:
> it will copy from `/usr/share/docs/audit-VERSION/rules/`.

On RHEL 8, the directory name is `/usr/share/doc/audit/rules/`, without the version element. Is that different on RHEL 7?

yuumasato:
> Is that different on RHEL 7?

Yes, RHEL 7 has versioned docs dir.

dahaic:
Thank you @yuumasato ! I know this has been painful road, but it's finally ready to be merged. Looking forward to the next generation check, utilizing new Audit split (when it gets available).

PR #4681, opened 2019-08-06
Harden ssh client crypto policy

vojtapolasek:
#### Description:

Implementing rule and checks for hardening of SSH client Crypto Policies as required by Common Criteria.
Unfortunately overriding the Crypto Policy is not the way to go in this case, because of the way in which SSH client parses the configuration files.
Therefore a file in /etc/ssh/ssh_config.d/ will be created and chosen options of Crypto Policy should be overided.

#### Rationale:

Based on Common Criteria, this rule checks and offer a remediation for SSH client configuration. The CC restricts ciphers, accepted public key algorithms, GSSAjPI key exchange, HMAC algorithms, KEX algorithms.


vojtapolasek:
This is work in progress! Currently it just builds, it is not tested at all.
I borrowed Jinja Oval macros and rewrote them so that they can create multiple criteria, tests, objects and states in one file. the problem was that the IDs of criteria, objects, tests and states were dependent only on rule ID, in my modified macros they depend also on the config parameter checked.

vojtapolasek:
Work finished, need review please.

rhel8/profiles/ospp.profile
```diff
@@ -1048,8 +1048,8 @@ selections:
     ## Enable rngd Service
     #################################################################
 
-    ## TO DO: https://github.com/ComplianceAsCode/content/issues/4472
-    #sysctl enable rngd.service
+    ## Enable the Hardware RNG Entropy Gatherer Service
+    - service_rngd_enabled
```
ggbecker: This rule is already selected in master branch. Do you know why it is showing as a new addition? Below is the link to the master file pointing to the modification:
https://github.com/ComplianceAsCode/content/blob/777b143344da542fce58ab4bdbc1578a352845c1/rhel8/profiles/ospp.profile#L1054

vojtapolasek: I don't know. Maybe I messed up the ospp.profile file. What I did in the end was that I took this file from the actual master and copied it into my branch and only added my rule at the end, see the last commit. How to fix it?

ggbecker: No idea what happened here and I don't know how to fix. Maybe reverting this last commit and rebasing on top of master could help.

vojtapolasek:
should be fixed now

adelton:
I did not dig into changes in detail but overall this is the approach we want to take to be able to use the FIPS crypto policy and override the options that need to be stricter.

rhel8/profiles/ospp.profile
```diff
@@ -851,3 +851,5 @@ selections:
 
     ## Enable dnf-automatic Timer
     - timer_dnf-automatic_enabled
+
+    - harden_ssh_client_crypto_policy
```
yuumasato: One more thing, add the selection of the rule close the other related rules, below `harden_sshd_crypto_policy` for example.

tests/data/group_system/group_software/group_integrity/group_crypto/rule_harden_ssh_client_crypto_policy/wrong_pubkey.fail.sh
```diff
@@ -7,6 +7,6 @@ echo -e "Match final all\n\
 RekeyLimit 512M 1h\n\
 GSSAPIAuthentication no\n\
 Ciphers aes256-ctr,aes256-cbc,aes128-ctr,aes128-cbc\n\
-PubkeyAcceptedKeyTypes ssh-rsa,ecdsa-sha2-nistp384\n\
+PubkeyAcceptedKeyTypes ssh-rsa,ecdsa-sha2-nistp384,ecdsa-sha2-nistp256,malicious_pubkdy\n\
```
yuumasato: Typo that doesn't matter much: `pubkdy`.

linux_os/guide/system/software/integrity/crypto/harden_ssh_client_crypto_policy/rule.yml
```diff
@@ -5,11 +5,13 @@ title: 'Harden SSH client Crypto Policy'
...
     The SSH client is by default configured to modify its configuration based on currently configured Crypto-Policy. However, in certain cases it might be needed to override the Crypto Policy specific to OpenSSH Client and leave rest of the Crypto Policy intact.
-    This is unfortunately not possible through the Crypto Policy mechanism directly due to the way in which the SSH Client parses configuration files. It is needed to apply Crypto Policy modifications by dropping a file into the <tt>/etc/ssh/ssh_config.d/</tt> directory. The file needs to be named in a way such that it is parsed by SSH client before the actual Crypto Policy.
+    This is unfortunately not possible through the Crypto Policy mechanism directly due to the way in which the SSH Client parses configuration files. During the parsing process, as soon as the client parses some configuration option and its value, it remembers it and ignores any subsequent overides. The mechanism of crypto policies appends eventual customizations at the end of the system wide crypto policy. Therefore customizations will be ignored by the client.
```
yuumasato: Suggestion: `not possible through the Crypto Policy mechanism directly due to ...` to `not directly possible through the Crypto Policy mechanism due to...` 
Typo: `overrides`.

yuumasato: `The mechanism of crypto policies appends eventual customizations at the end of the system wide crypto policy. Therefore customizations will be ignored by the client.`

Not sure what this means. crypto-policy mechanism append customizations at the end of crypto-policy, and because of this, they are ignored by the client?

Maybe you want to say that "due to how SSH Client reads the crypto-poicy, any setting that a "official" crypto-policy would configure, will be ignored by the client" ?

linux_os/guide/system/software/integrity/crypto/harden_ssh_client_crypto_policy/rule.yml
```diff
@@ -5,11 +5,13 @@ title: 'Harden SSH client Crypto Policy'
...
+    Considering supported platforms, the default configuration file in <tt>/etc/ssh_ssh_config</tt> includes files matching the pattern <tt>/etc/ssh_ssh_config.d/*.conf</tt>. Currently, by default, there is present only one file called <tt>05-redhat.conf</tt>. This file subsequently includes the actual system-wide crypto policy placed in <tt>/etc/crypto-policies/back-ends/openssh.config</tt>.
+    To overide the system wide crypto policy, it is needed to place a file in the <tt>/etc/ssh/ssh_config.d/</tt> so that it is loaded before the <tt>05-redhat.conf</tt>. In this case it is file named <tt>05-ospp.conf</tt> containing parameters wchich need to be changed with respect to the crypto policy.
```
yuumasato: Typo: `override`.
I see no need for `it is needed to`: `To override the system wide crypto policy place a file`
And the file we drop in should be called `02-ospp.conf`, right?

linux_os/guide/system/software/integrity/crypto/harden_ssh_client_crypto_policy/rule.yml
```diff
@@ -0,0 +1,33 @@
...
+references:
+    ospp : FCS_SSHC_EXT.1
+
+ocil_clause: 'Crypto Policy for OpenSSH Client is not configured according to CC requirements'
+
+ocil: |-
+    To verify if the OpenSSH Client uses defined Crypto Policy, run:
+    <pre>$ cat /etc/ssh/ssh_config.d/02-ospp.conf</pre>
+    and verify that the line matches
+    <pre>Ciphers aes256-ctr,aes256-cbc,aes128-ctr,aes128-cbc</pre>
```
yuumasato: Match final all, RekeyLimit, and GSSAPIAuthentication are not mentioned in `ocil`.

linux_os/guide/system/software/integrity/crypto/harden_ssh_client_crypto_policy/rule.yml
vojtapolasek: I want to say that if you go the /etc/crypto-policies/local.d way, in the case of Openssh client this approach will not work if you try to override a parameter which is already configured by the system wide policy.

yuumasato: Aha! I understand this as part of `This file subsequently includes the actual system-wide crypto policy placed in <tt>/etc/crypto-policies/back-ends/openssh.config</tt>.`
As any file in ` /etc/crypto-policies/local.d` will become part of what is in `back-ends/openssh.config`.

To me this seems a bit too deep in details for the rule description.

linux_os/guide/system/software/integrity/crypto/harden_ssh_client_crypto_policy/rule.yml
```diff
@@ -0,0 +1,36 @@
...
+    Considering supported platforms, the default configuration file in <tt>/etc/ssh_ssh_config</tt> includes files matching the pattern <tt>/etc/ssh_ssh_config.d/*.conf</tt>. Currently, by default, there is present only one file called <tt>05-redhat.conf</tt>. This file subsequently includes the actual system-wide crypto policy placed in <tt>/etc/crypto-policies/back-ends/openssh.config</tt>.
+    To override the system wide crypto policy for Openssh client, place a file in the <tt>/etc/ssh/ssh_config.d/</tt> so that it is loaded before the <tt>05-redhat.conf</tt>. In this case it is file named <tt>02-ospp.conf</tt> containing parameters wchich need to be changed with respect to the crypto policy.
+    This rule checks if the file exists and if it contains required parameters and values which modify the Crypto Policy.
+
```
yuumasato: I think the rule description is a bit convoluted.
By swapping paragraph in line 10 with paragraph in line 8, description of what to do to override the crypto policy is more evident.
After that, merge paragraph currently in lines 8 and 9 and  make them shorter. This way the justification of why it needs to be done this way is shorter and secondary.

yuumasato:
Good stuff, thank you @vojtapolasek!

PR #4446, opened 2019-06-24
update OSPP

shawndwells:
Update OSPP profile with common criteria rule selections.

PR #4682, opened 2019-08-07
Reintroduce crypto policy rules in the OSPP profile for RHEL8

matusmarhefka:
#### Description:

Crypto policy rules were removed in the PR #4446. This commit
reintroduces them back into the profile.


yuumasato:
LGTM, thanks.

PR #4684, opened 2019-08-07
Delete unused/unwanted dconf_use_text_backend rule.

ggbecker:
#### Description:

- As part of previous dconf bugfix, the first attempt was to set dconf to use text backend as its database. It turned out that this didn't solve the problem thus leaving this rule as unused/unwanted. I propose to remove all artifacts related to it.

#### Rationale:

- Remove unused/unwanted dconf_use_text_backend rule.


ggbecker:
@openscap-ci test this please

shared/references/cce-redhat-avail.txt
```diff
@@ -706,3 +706,4 @@ CCE-82900-2
 CCE-82901-0
 CCE-82902-8
 CCE-82903-6
+CCE-80782-6
```
ggbecker: returns back the CCE to the available pool of numbers.

yuumasato:
LGTM.
I'm just not sure about returning the CCE to the pool, as the rule is part of release 0.1.45.

ggbecker:
> I'm just not sure about returning the CCE to the pool, as the rule is part of release 0.1.45.

That's true, it doesn't make sense to return it to the pool. I'll remove it.

shared/references/cce-redhat-avail.txt
ggbecker: Will remove as it was part of at least 0.1.45 release and could cause conflict.

matejak:
LGTM, thank you for this cleanup!

PR #4686, opened 2019-08-07
Jinja bash support

matejak:
- Added support for bash jinja macros in a similar way as we do with OVAL macros.
- Introduced a sample macro usage.


jan-cerny:
@matejak LGTM

PR #4687, opened 2019-08-07
Updated Fedora GPG info.

matejak:
The Fedora 30 is out, and info about Fedora 31 is available.
However, how does one determine value of `..._pkg_release`?

### DIFF
--- a/fedora/product.yml
+++ b/fedora/product.yml
@@ -10,13 +10,21 @@ pkg_manager: "dnf"
 
 init_system: "systemd"
 
-# The fingerprints below are retrieved from https://getfedora.org/keys/
-latest_version: 29
-latest_release_fingerprint: "5A03B4DD8254ECA02FDA1637A20AA56B429476B4"
-latest_pkg_release: "5a886537"
-latest_pkg_version: "429476b4"
+# The fingerprint and pkg_version are retrieved from https://getfedora.org/keys/
+# Future version is not used, but it makes updating latest_version easier.
+future_version: 31
+future_release_fingerprint: "7D22D5867F2A4236474BF7B850CB390B3C3359C4"
+future_pkg_version: "3c3359c4"
+# Obtain the pkg_release like this:
+# sudo rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-<version>-x86_64
+# rpm -q gpg-pubkey --qf '%{VERSION}-%{RELEASE}\n' | grep <pkg_version> | cut -f 2 -d -
 
-previous_version: 28
-previous_release_fingerprint: "128CF232A9371991C8A65695E08E7E629DB62FB1"
-previous_pkg_release: "59920156"
-previous_pkg_version: "9db62fb1"
+latest_version: 30
+latest_release_fingerprint: "F1D8EC98F241AAF20DF69420EF3C111FCFC659B9"
+latest_pkg_version: "cfc659b9"
+latest_pkg_release: "5b6eac67"
+
+previous_version: 29
+previous_release_fingerprint: "5A03B4DD8254ECA02FDA1637A20AA56B429476B4"
+previous_pkg_version: "429476b4"
+previous_pkg_release: "5a886537"


ggbecker:
> However, how does one determine value of `..._pkg_release`?

On RHEL8 I usually import the keys and then query gpg-pubkey rpm package:
```
rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release 
rpm -q gpg-pubkey --qf '%{NAME}-%{VERSION}-%{RELEASE}\n'
gpg-pubkey-fd431d51-4ae0493b
gpg-pubkey-d4082792-5b32db75
```

I think you can do similarly on Fedora.

redhatrises:
https://getfedora.org/security/ may or may not be what you are looking for.

yuumasato:
> > However, how does one determine value of `..._pkg_release`?
> 
> On RHEL8 I usually import the keys and then query gpg-pubkey rpm package:
> 
> ```
> rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release 
> rpm -q gpg-pubkey --qf '%{NAME}-%{VERSION}-%{RELEASE}\n'
> gpg-pubkey-fd431d51-4ae0493b
> gpg-pubkey-d4082792-5b32db75
> ```
> 
> I think you can do similarly on Fedora.

Unfortunately the package release is not listed in the keys page.
The way to find it out seems to be to install the key and check for what `rpm` command reports.

matejak:
Thanks, I have updated it, along with the guide.

matejak:
Sorry for the confusion with delete/restore.

fedora/product.yml
```diff
@@ -10,13 +10,20 @@ pkg_manager: "dnf"
...
 
-# The fingerprints below are retrieved from https://getfedora.org/keys/
-latest_version: 29
-latest_release_fingerprint: "5A03B4DD8254ECA02FDA1637A20AA56B429476B4"
-latest_pkg_release: "5a886537"
-latest_pkg_version: "429476b4"
+# The fingerprint and pkg_version are retrieved from https://getfedora.org/keys/
+future_version: 31
+future_release_fingerprint: "7D22D5867F2A4236474BF7B850CB390B3C3359C4"
+future_pkg_version: "3c3359c4"
```
yuumasato: @matejak What is the intended use of future key?
The OVAL check doesn't use them.

fedora/product.yml
```diff
@@ -10,13 +10,20 @@ pkg_manager: "dnf"
...
-previous_pkg_version: "9db62fb1"
+latest_version: 30
+latest_release_fingerprint: "F1D8EC98F241AAF20DF69420EF3C111FCFC659B9"
+latest_pkg_version: "cfc659b9"
+latest_pkg_release: "5b6eac67"
+
+previous_version: 29
+previous_release_fingerprint: "5A03B4DD8254ECA02FDA1637A20AA56B429476B4"
+previous_pkg_release: "5a886537"
+previous_pkg_version: "429476b4"
```
yuumasato: Reorder like in fedora 30 keys, version, then release.

fedora/product.yml
matejak: I have just filled that, as the information is available on the website.

fedora/product.yml
matejak: Done by rebase.

fedora/product.yml
yuumasato: @matejak Should we just have it there unused?

I think it should be used in check and remediations (If we would like to test that this rule works in future fedora 31), or have it removed.

matejak: What do you mean? If I just remove that, then anybody updating this rule for Fedora 31 five months from now will have to pull exactly that data from the website. Having it in the file will save time, and it doesn't break anything.

yuumasato: I'm not strongly opposed to this, I just don't see much use in storing data for future use.

> Having it in the file will save time, and it doesn't break anything.

The time spent will be the same for every Fedora release, one will be filling in key data for `future_version` or `current_version`.

yuumasato:
@matejak Thank you for updating this and adding comments about how to obtain the `pkg_release`.

fedora/product.yml
matejak: Actually one can use `future_version` to populate `current_version` every other release, therefore not having to look at the website :-)

yuumasato: Yes, for `current_version` it will be faster, but `future_version` will have to be populated again.
One may even run into the issue that `future_version` key might not exist yet, not sure what are the odds of this.

PR #4688, opened 2019-08-08
Test suite change logging prefix to warning

mildas:
#### Description:
Change logging prefix to `WARNING` when profile is missing or test is not applicable on given platform.

### DIFF
--- a/tests/ssg_test_suite/rule.py
+++ b/tests/ssg_test_suite/rule.py
@@ -47,7 +47,7 @@ def get_viable_profiles(selected_profiles, datastream, benchmark):
                 valid_profiles += [ds_profile]
 
     if not valid_profiles:
-        logging.error('No profile ends with "{0}"'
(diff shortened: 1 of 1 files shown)


ggbecker:
What is the rationale of these changes?

mildas:
@ggbecker We have test that runs test suite and then checks for `^ERROR` in output. When some profile is included in test header, but the profile isn't built (either it doesn't exist - e.g. C2S in RHEL8 - or we just don't want to build the profile), then the test suite prints error messages with the information. And the overall result is fail.

ggbecker:
> @ggbecker We have test that runs test suite and then checks for `^ERROR` in output. When some profile is included in test header, but the profile isn't built (either it doesn't exist - e.g. C2S in RHEL8 - or we just don't want to build the profile), then the test suite prints error messages with the information. And the overall result is fail.

Sounds reasonable. Thank you for the information.

PR #4690, opened 2019-08-08
Convert functions from include_lineinfile.sh to Jinja macros

jan-cerny:
Currently, the Bash remediation functions are saved in an XCCDF Value
and their code is substituted by OpenSCAP when the remediation is
executed.  However, the way of using the functions in individual
remediation is tedious and hard to understand. People have to source a
non-existent file /usr/share/scap-security-guide/remediation_functions
and then call an empty function to make the function code available in
the remediation. We have discovered that it does not allow to use
multiple functions in a single remediation. Another drawback is that
the function code is duplicated in HTML report each time it is used
which makes the report bigger. This patch aims to simplify creating the
Bash remediation and address the aforementioned problems by converting
the functions to Jinja macro. In Bash remediation you can now simply
call a Jinja macro and a simple, stand-alone Bash code will be generated
for you.  In this patch only the functions from include_lineinfile.sh
are converted to Jinja. Other shared modules can be converted in future
in a similar way.

### DIFF
--- a/linux_os/guide/system/permissions/restrictions/coredumps/coredump_disable_storage/bash/shared.sh
+++ b/linux_os/guide/system/permissions/restrictions/coredumps/coredump_disable_storage/bash/shared.sh
@@ -3,7 +3,4 @@
 # strategy = restrict
 # complexity = low
 # disruption = low
-. /usr/share/scap-security-guide/remediation_functions
-include_lineinfile
-
 {{{ bash_coredump_config_set("Storage", "none") }}}
--- a/linux_os/guide/system/permissions/restrictions/coredumps/coredump_disable_storage/bash/shared.sh
+++ b/linux_os/guide/system/permissions/restrictions/coredumps/coredump_disable_storage/bash/shared.sh
@@ -6,4 +6,4 @@
 . /usr/share/scap-security-guide/remediation_functions
 include_lineinfile
 
-coredump_config_set Storage none
+{{{ bash_coredump_config_set("Storage", "none") }}}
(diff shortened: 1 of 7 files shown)


shared/macros-bash.jinja
```diff
@@ -7,3 +56,95 @@
...
+{{%- macro set_config_file(path, parameter, value, create, insert_after, insert_before, insensitive, separator, separator_regex, prefix_regex) -%}}
+    {{%- set line_regex = prefix_regex+parameter+separator_regex -%}}
+    {{%- set new_line = parameter+separator+value -%}}
+    {{{ lineinfile(path, line_regex, "", false, false, "", "", insensitive) }}}
+    {{{ lineinfile(path, "", new_line, create, true, insert_after, insert_before, insensitive) }}}
+{{%- endmacro -%}}
+
+{{%- macro lineinfile(path, regex, line, create=false, present=true, insert_after="", insert_before="", insensitive=true) -%}}
+    {{%- if not present -%}}
+        if [ -e "{{{ path }}}" ] ; then
```
matejak: The previous `-%}}` eat the leading whitespace from `... if`, but there is nothing eaten from the corresponding `... fi`.

linux_os/guide/system/permissions/restrictions/coredumps/coredump_disable_storage/bash/shared.sh
```diff
@@ -6,4 +6,4 @@
 . /usr/share/scap-security-guide/remediation_functions
 include_lineinfile
```
mildas: include is not needed anymore with macro

jan-cerny: thanks

matejak:
LGTM, thank you for the PR!

PR #4691, opened 2019-08-08
Cleanup the RHEL7 ccc.profile, minimally

comps:
This is a minimalist version of #4671 touching just `rhel7/ccc.profile`, updating it for Common Criteria needs.

The only remaining TODO we know of is auditing (mentioned in the profile, see changes).


yuumasato:
LGTM.

PR #4692, opened 2019-08-09
Added stripping of new line when obtaining IP addr by podman inspect

matusmarhefka:
#### Description:

- The new line character in IP addr of a container was breaking podman backend of the SSG Test Suite.


matejak:
ACK, Thank you for the PR!

PR #4693, opened 2019-08-09
Fix generated SCAP 1.2 content 

yuumasato:
#### Description:

- Fix bug where built SCAP 1.2 DS was actually SCAP 1.3
- Also, renamed script to generic update sds version.
- The script receives target SCAP version and handles update of version.
  - When target version is 1.3, it is updated.
  - When target version is 1.2, nothing is done.

#### Rationale:

Enpower DS version update script to decide when to update or not,
instead of controling this decision on CMake script

This is to avoid defining 4 custom commands to build the DS
in CMake macro ssg_build_ds(). Namely:
1. command for rhel6|7 with 1.2 (includes pci-dss benchmark)
2. command for rhel6|7 with 1.3 (includes pci-dss benchmark)
3. command for other products with 1.2
4. command for other produts with 1.3

### DIFF
--- a/build-scripts/update_sds_version.py
+++ b/build-scripts/update_sds_version.py
@@ -110,6 +110,7 @@ def move_patches_up_to_date_to_source_data_stream_component(datastreamtree):
 
 def parse_args():
     p = argparse.ArgumentParser()
+    p.add_argument("--version", required=True)
     p.add_argument("--input", required=True)
     p.add_argument("--output", required=True)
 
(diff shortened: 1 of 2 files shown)


yuumasato:
> On the other hand loading the datastream using `ElementTree` and writing it back introduces the namespace to some XML entries and generates both datastream in a similar way

Yes. the reason to load and write the DS is consistency in namespaces.

PR #4697, opened 2019-08-09
Set number of records to cause an explicit flush to audit logs.

adelton:
#### Description:

- Set number of records to cause an explicit flush to audit logs.

#### Rationale:

- If option freq isn't set to 50, the flush to disk may happen after higher number of record, increasing the danger of audit loss.


adelton:
Fixed typo -> 5ac8f19ec909099e5976d929b0402b856296834b.

linux_os/guide/system/auditing/auditd_freq/rule.yml
```diff
@@ -0,0 +1,30 @@
...
+title: 'Set number of records to cause an explicit flush to audit logs'
+
+description: |-
+    To configure Audit daemon to issue an explicit flush to disk command
+    after writing 50 records, set <tt>freq</tt> to <tt>50</tt>
+    in <tt>/etc/audit/auditd.conf</tt>.
+
+rationale: |-
+    If option <tt>freq</tt> isn't set to <tt>50</tt>, the flush to disk
+    may happen after higher number of record, increasing the danger
```
jan-cerny: probably records

adelton: Fixed in 41a2debcab5fae88ccd15a62e87a5aef9381f3a2.

adelton:
Rebased on master -> 111ee162900cc8f2a4ed2c2181949e95bcee571e.

PR #4702, opened 2019-08-09
Enable unit-testing of bash shared jinja macros

matejak:
This PR enables to use jinja macros with the BATS framework.
A recipe how to expand the template in `tests/unit/bash` into `build/tests` is created by `cmake`, and `ctest -R bash` then calls a helper script that performs the expansion and executes the test suite run.


shared/macros-bash.jinja
```diff
@@ -53,8 +53,8 @@
 # rc: The error code (optional, default is 1)
 # action: What to do (optional, default is 'exit', can be also 'return' or anything else)
 #}}
-{{% macro die(message, rc=1, action="exit") -%}}
-printf '%s\n' '{{{ message }}}' >&2
+{{% macro die(message, rc=1, action="exit") %}}
```
jan-cerny: Is the removal of `-` before `%` intended?

shared/macros-bash.jinja
```diff
@@ -71,7 +71,7 @@ printf '%s\n' '{{{ message }}}' >&2
 # separator_regex: regular expression that describes the separator and surrounding whitespace
 # prefix_regex: regular expression describing allowed leading characters at each line
 #}}
-{{%- macro set_config_file(path, parameter, value, create, insert_after, insert_before, insensitive, separator, separator_regex, prefix_regex) -%}}
+{{%- macro set_config_file(path, parameter, value, create, insert_after, insert_before, insensitive=True, separator=" ", separator_regex="\s\+", prefix_regex="^\s*") -%}}
```
jan-cerny: Have we think about insensitive should be default True or False?

shared/macros-bash.jinja
```diff
@@ -101,41 +101,40 @@ LC_ALL=C sed -i "/{{{ regex }}}/{{{ modifier }}}" "{{{ path }}}"
     {{%- else -%}}
         {{%- set grep_args="-q -m 1" -%}}
     {{%- endif -%}}
-    {{% if (insert_after == "" and insert_before == "")  or insert_after == "EOF" -%}}
-# Insert at the end of the file
 cp "{{{ path }}}" "{{{ path }}}.bak"
+    {{% if not (insert_after or insert_before)  or insert_after == "EOF" -%}}
+# Insert at the end of the file
```
jan-cerny: This comment is shifted by 4 spaces to the right, probably because of 4 spaces before the if on previous line.  That breaks our beautiful indentation of the generated code. Adding `-` after `{{%` won't help because that would merge the comment with the `cp` command. Could you take a look into the indentation?

shared/macros-bash.jinja
matejak: I would leave it like it is. Tests will assure that it won't flip unnoticed.

shared/macros-bash.jinja
matejak: Good, I hope that I have fixed that.

shared/macros-bash.jinja
matejak: No, but other dashes had to be added elsewhere :-)

shared/macros-bash.jinja
jan-cerny: I have checked that it doesn't occur anymore.

jan-cerny:
ACK.

PR #4704, opened 2019-08-09
configure_crypto_policy: Do not check back-end symlink

yuumasato:
#### Description:

- Don't check for crypto-policy back-end symlinks.

#### Rationale:

- ~~The fact that they are not symlinks doesn't imply that they haven't been
tampered.~~
EDIT:
- Checking that the back-end `config` is a symlink to policy `txt` doesn't add value, as we don't check that policy `txt`s was not modified. 
- Plus, there are valid scenarios where it is ok not to have symlinks there. For example, by adding configs into `/etc/crypto-policies/local.d/` and running `update-crypto-policy`, the file in `/etc/crypto-policy/back-ends/` is not a symlink anymore.
  - This is done in `harden_sshd_crypto_policy`

### DIFF
--- a/linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/oval/shared.xml
+++ b/linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/oval/shared.xml
@@ -1,37 +1,3 @@
-{{%- if target_oval_version == [5, 11] -%}}
-{{# there is no good alternative for symlink_test for OVAL 5.10 #}}
-{{%- macro crypto_policy_symlink_criterion(library) %}}
-      <criterion comment="Check if /etc/crypto-policies/back-ends/{{{ library }}}.config is linked to /usr/share/crypto-policies/POLICY/{{{ library }}}.txt" test_ref="test_crypto_policy_{{{ library }}}_symlink" />
-{{%- endmacro %}}
-
-{{%- macro crypto_policy_symlink_check(library) %}}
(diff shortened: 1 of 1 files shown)


redhatrises:
Why?

The check is checking that there are no customizations to the crypto-policies that are delivered in the RPM. The tampering check is when RPM validates that those files haven't been modified (assuming we haven't made changes that breaks that), so the rationale for this change is not accurate.

yuumasato:
> The check is checking that there are no customizations to the crypto-policies that are delivered in the RPM.

@redhatrises We will require that some crypto-policies be customized. For example the rule in `harden_sshd_crypto_policy` for OSPP Profile.

> The tampering check is when RPM validates that those files haven't been modified (assuming we haven't made changes that breaks that), so the rationale for this change is not accurate.

I understand, so the check is not complete, and misses RPM part.
Updated the rationale.

vojtapolasek:
I think there is a middle road. I suggest that we keep the simlink checks, but we can modify the expression as:
there is either a simlink (as it is now)
or
there is a file instead and at the same time a file with $filename in /etc/crypto-policies/local.d/ exists.
where $filename would be a variable containing our dropin file, currently opensshserver-ospp.config.
What do you think?

yuumasato:
Looks reasonable.
Check that there is a symlink for the "library", or that there is a file in `/etc/crypto-policy/local.d/` for "library".

PR #4706, opened 2019-08-12
Remove watches since syscall rules cover all cases.

adelton:
Having both watches and syscall rules affects readability
of the audit trail as there are two events for each operation.

#### Description:

- Remove watches since syscall rules cover all cases.

#### Rationale:

- Having both watches and syscall rules affects readability of the audit trail as there are two events for each operation.
- Similar change was made in upstream auditd: https://github.com/linux-audit/audit-userspace/commit/faba74ed367d5f8f2b6b5c8dcf6afca63a23c466.


adelton:
Rebased on master -> b6eb80b07c492453ed004ef44d426cfe4c813e6d.

linux_os/guide/system/auditing/auditd_configure_rules/audit_kernel_module_loading/audit_rules_kernel_module_loading_finit/rule.yml
```diff
@@ -39,6 +39,8 @@ references:
     srg: SRG-OS-000471-GPOS-00216,SRG-OS-000477-GPOS-00222
     vmmsrg: SRG-OS-000477-VMM-001970
     stigid@rhel7: "030821"
+    stigid@rhel7: "030840"
+    stigid@rhel7: "030860"
```
jan-cerny: The `stigid@rhel7` key should be there only once - `references` is  a dictionary and the keys owerwrite each other. To see the results, build rhel7 product, open the HTML guide for STIG in `build/guides/ssg-rhel7-guide-stig.html`, find `ccdf_org.ssgproject.content_rule_audit_rules_kernel_module_loading_delete` and you'll see that only one of the IDs is there.

I think there should be used the comma separated list as in other references (see above and below). However, I tried it and it doesn't work. I'm investigating that.

jan-cerny: https://github.com/ComplianceAsCode/content/pull/4725 will fix the infrastructure so that you will be able to use a comma separated list for the STIG IDs same way as we use for other references.

adelton: I can of course just leave the original `swigid` there. My goal was to preserve the original ids with this rule that supersedes the old ones. But maybe it's not even needed?

jan-cerny: That's a great point. @yuumasato Do you know if a single rule is allowed to have multiple STIG IDs assigned? I mean from a logical standpoint, not whether it works in our tools.

yuumasato: > Do you know if a single rule is allowed to have multiple STIG IDs assigned?

I think no, but I'm not the expert in this matter.
@shawndwells or @redhatrises ^

jan-cerny:
If all rules will have just 1 STIG ID you won't be blocked by #4725 anymore.

linux_os/guide/system/auditing/auditd_configure_rules/audit_kernel_module_loading/audit_rules_kernel_module_loading_finit/rule.yml
redhatrises: @jan-cerny @yuumasato there can only be 1 STIG ID to a rule. Having multiple STIGIDs to a rule is a bug.

adelton:
Removed the duplicated `stigid`'s, rebased on master -> 0dbd1464552c9dd9ae5c989d2b8c53382048e93b.

linux_os/guide/system/auditing/auditd_configure_rules/audit_kernel_module_loading/audit_rules_kernel_module_loading_finit/rule.yml
jan-cerny: Thanks

jan-cerny:
@adelton thank you

Issue #4703, opened 2019-08-09
tests/test_suite.py rule fails with Rule xccdf_org.ssgproject.content_rule_coredump_disable_backtraces has not been evaluated! Wrong profile selected in test scenario?


adelton:
#### Description of problem:

Running `tests/test_suite.py rule` now fails with
```
Rule xccdf_org.ssgproject.content_rule_coredump_disable_backtraces has not been evaluated! Wrong profile selected in test scenario?
```

#### SCAP Security Guide Version:

The master: v0.1.45-195-g68aacd11a

#### Operating System Version:

On the host, Fedora 30 with docker-1.13.1-67.git1185cfd.fc30.x86_64.
In the container, RHEL 8.

#### Steps to Reproduce:

1.
```
docker build --build-arg=CLIENT_PUBLIC_KEY="$( cat /dir/.ssh/id_rsa.pub )" -t test_suite-rhel8 -f Dockerfiles/test_suite-rhel8 Dockerfiles
```
2.
```
./build_product rhel8
```
3.
```
tests/test_suite.py rule --docker test_suite-rhel8 --datastream build/ssg-rhel8-ds.xml coredump_disable_backtraces
```

#### Actual Results:
```
Setting console output to log level INFO
INFO - The base image option has been specified, choosing Docker-based test environment.
INFO - Logging into /dir/scap-content/logs/rule-custom-2019-08-09-1655/test_suite.log
INFO - xccdf_org.ssgproject.content_rule_coredump_disable_backtraces
ERROR - Failed to cleanup file '/dir/scap-content/logs/rule-custom-2019-08-09-1655/xccdf_org.ssgproject.content_rule_coredump_disable_backtraces-coredumps_processsizemax_nonzero.fail.sh-initial.html'
ERROR - Script coredumps_processsizemax_nonzero.fail.sh using profile xccdf_org.ssgproject.content_profile_ospp found issue:
ERROR - Rule xccdf_org.ssgproject.content_rule_coredump_disable_backtraces has not been evaluated! Wrong profile selected in test scenario?
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_coredump_disable_backtraces'.
ERROR - Failed to cleanup file '/dir/scap-content/logs/rule-custom-2019-08-09-1655/xccdf_org.ssgproject.content_rule_coredump_disable_backtraces-coredumps_processsizemax_0.pass.sh-initial.html'
[...]

PR #4708, opened 2019-08-12
Add option in build_product to build SCAP 1.2 content.

ggbecker:
#### Description:

- Add option to build SCAP content version 1.2 using `build_product` script.

#### Rationale:

- OpenSCAP versions prior to 1.3.1 are not able to consume the DS without providing `--fetch-remote-resources` option. This option provides easy way to create SCAP 1.2 content and which can be consumed by older versions of OpenSCAP.

- Relates to #4703

### DIFF
--- a/build_product
+++ b/build_product
@@ -34,7 +34,7 @@ handle_wrong_argument() {
 	echo
 	echo
 	printf 'Usage:\n'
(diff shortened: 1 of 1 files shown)


jan-cerny:
It works OK for me, I was able to build both 1.2 and 1.3 Fedora DS with the updated build_product script. Thank you!

PR #4709, opened 2019-08-12
Disable client alive messages.

adelton:
#### Description:

- Disable client alive messages.

#### Rationale:

- No waiting for keepalives, just drop the idle session after 10 minutes.


mildas:
Thank you

PR #4710, opened 2019-08-12
Convert Bash functions package_install|remove to Jinja macros

jan-cerny:
#### Description:
This patch removes package_install and package_remove and replaces them by new macros bash_package_install and bash_package_remove.  As a side benefit, the remediation will not need to detect the packaging system during runtime, but it will be done during the build based on product
information. 

#### Rationale:
Another step to get rid of the poorly designed mechanism of including shared bash functions. People have to source a non-existent file /usr/share/scap-security-guide/remediation_functions. We have discovered that it does not allow to use multiple functions in a single remediation. Another drawback is that the function code is duplicated in HTML report each time it is used which makes the report bigger. This patch aims to simplify creating the Bash remediation and address the aforementioned problems by converting the functions to Jinja macro. 

Similar to: https://github.com/ComplianceAsCode/content/pull/4690


linux_os/guide/services/ntp/service_chronyd_or_ntpd_enabled/bash/shared.sh
```diff
@@ -2,7 +2,7 @@
 . /usr/share/scap-security-guide/remediation_functions
 
 if ! `rpm -q --quiet chrony` && ! `rpm -q --quiet ntp-`; then
```
matejak: When you are at it, I would apply the DeMorgan rule on the condition, and those backticks are redundant.

matejak:
This is a great PR, please just make that condition less convoluted.

jan-cerny:
@matejak I have made that condition less convoluted.

matejak:
Great, thank you!

PR #4711, opened 2019-08-12
Force frequent session key renegotiation.

adelton:
#### Description:

- Force frequent session key renegotiation.

#### Rationale:

- By decreasing the limit based on the amount of data and enabling time-based limit, effects of potential attacks against encryption keys are limited.


adelton:
The OVAL check does not seem to work for this value which has space in it.

adelton:
Addressed in e4e1c8273e1af54ea8e9dfc43db28da95c5e0aac.

jan-cerny:
now it matches for me

PR #4715, opened 2019-08-13
When HostbasedAuthentication is disabled using disable_host_auth, sshd_disable_rhosts and sshd_disable_user_known_hosts are redundant.

adelton:
#### Description:

- When HostbasedAuthentication is disabled using disable_host_auth, sshd_disable_rhosts and sshd_disable_user_known_hosts are redundant.

#### Rationale:

- Both `IgnoreRhosts` and `IgnoreUserKnownHosts` only apply when `HostbasedAuthentication` is enabled. And in these profiles, it is disabled via `disable_host_auth`.


jan-cerny:
Indeed, man 5 sshd_config says the options work during HostbasedAuthentication.

PR #4716, opened 2019-08-13
Fix disk configuration in OSPP anaconda kickstart file.

ggbecker:
#### Description:

- Fix the disk configuration to be aligned with the rules required by OSPP profile. Removed the CCEs as the are referencing CCEs from RHEL 6 and 7 versions.

#### Rationale:

- Kickstart can be properly used to install RHEL8.1 machine using OSPP profile

- Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1740621


ggbecker:
@matejak this might help when creating installing the OS, as extra argument you can provide:

`inst.ks=http://192.168.122.1:8000/ssg-rhel8-ospp-ks.cfg inst.addrepo=AppStream,<repository_url>`

matejak:
I can confirm that the installation passes, and there are plenty of fixed rules as a result, and just a handful of failed.

PR #4718, opened 2019-08-13
Update OL8 HIPAA profile

iokomin:
#### Description:

Removed ensure_gpgcheck_repo_metadata rule, added dconf_db_up_to_date rule.

#### Testing:
- checked OL8 build to pass
- checked updated rules as part of hipaa evaluation on OL8


dahaic:
LGTM

PR #4721, opened 2019-08-13
Always set OVAL version using the variable CACHE

yuumasato:
#### Description:

- Set target OVAL version variables using the variable CACHE.

#### Rationale:
- When SCAP target version was 1.2, the OVAL target version was set without considering the  variable cache. 
  This makes it impossible to overide OVAL version when building SCAP 1.2
content, for example:
`cmake -G Ninja -DSSG_TARGET_OVAL_MINOR_VERSION:STRING=11 -DSSG_TARGET_SCAP_VERSION:STRING=1.2`


matejak:
I can confirm that it works, thank you for the PR!

PR #4723, opened 2019-08-14
The use of ed25519 is disabled via HostKeyAlgorithms in FIPS crypto policy.

adelton:
#### Description:

- No need to control the use of keys based on key filenames.

#### Rationale:

- The use of ed25519 is disabled via HostKeyAlgorithms in FIPS crypto policy.

### DIFF
--- a/rhel8/profiles/ospp.profile
+++ b/rhel8/profiles/ospp.profile
@@ -996,9 +996,6 @@ selections:
     ## Setup SSH Server
     #################################################################
 
-    ## TO DO: https://github.com/ComplianceAsCode/content/issues/4466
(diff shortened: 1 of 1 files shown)


matejak:
I see that it is disabled for `gnutls` in the crypto policy, what indicates that it is disabled for SSH?

adelton:
```
# sed 's/ /\n/g' /usr/share/crypto-policies/FIPS/opensshserver.txt | grep HostKeyAlgorithms | wc -l
1
# sed 's/ /\n/g' /usr/share/crypto-policies/FIPS/opensshserver.txt | grep ed25519 | wc -l
0
```

matejak:
OK, I can confirm that. Thank you for the PR!

PR #4724, opened 2019-08-14
Remove GNOME rules from rhel7/ospp

comps:
login_banner_text doesn't seem to be used by anything else
in the profile.

### DIFF
--- a/rhel7/profiles/ospp.profile
+++ b/rhel7/profiles/ospp.profile
@@ -138,9 +138,6 @@ selections:
     - accounts_passwords_pam_faillock_interval
     - var_accounts_passwords_pam_faillock_unlock_time=never
     - accounts_passwords_pam_faillock_unlock_time
-    - dconf_gnome_banner_enabled
-    - login_banner_text=usgcb_default
-    - dconf_gnome_login_banner_text
     - disable_users_coredumps
     - var_accounts_max_concurrent_login_sessions=10
     - accounts_max_concurrent_login_sessions


rhel7/profiles/ospp.profile
```diff
@@ -138,9 +138,6 @@ selections:
     - accounts_passwords_pam_faillock_interval
     - var_accounts_passwords_pam_faillock_unlock_time=never
     - accounts_passwords_pam_faillock_unlock_time
-    - dconf_gnome_banner_enabled
-    - login_banner_text=usgcb_default
```
dahaic: Help me understand - we have `sshd_enable_warning_banner` enabled in the profile - do we need specific wording of the warning? Becase `login_banner_text` contains wording, that should be then applied through `banner_etc_issue` rule that is not enabled by the profile. Can you check the requirements?

comps: According to my greps, `sshd_enable_warning_banner` does not use `login_banner_text`, it just sets `sshd_config` to do `Banner /etc/issue`.

As far as the ccc/ospp profile and the wording is concerned, it was brought up before, the response from stakeholders being:
> The configuration annex has specific text. I am not using the text for DoD 
systems. It is text for everyone else. I really do not think SCAP should 
check the contents of the banner. It should check that there is a banner.

dahaic: Perfect, thanks!

dahaic:
LGTM, thanks!

ggbecker:
The profile still selects `sshd_enable_warning_banner` which sets the `sshd` configuration to consume the content of `/etc/issue`. I wonder if instead of removing the `login_banner_text=usgcb_default`, we should add `banner_etc_issue` instead. 

See: https://github.com/ComplianceAsCode/content/pull/4728

PR #4725, opened 2019-08-14
Fix STIG IDs reference processing

jan-cerny:
#### Description:

This PR ~allows to assign a rule multiple STIG IDs using a comma-separated list which is the same way we use for the other references.~ prevents people from  assigning a rule multiple STIG IDs using a comma-separated list because a rule can have only 1 STIG ID. 

There were 2 basic problems:
1. Python code converting YAML to shorthand was not splitting the string at comma. 
2. XSLT template that converts shorhand to XCCDF had a special treatment for STIG IDs which always copies the whole attribute value instead of each identifier which lead to duplicate `reference` elements.

Note: there was some `stigid-concat` which I can't find anywhere and it seems to be evaluated as empty string which means result of concatenating it with any string is that string.

#### Rationale:
In https://github.com/ComplianceAsCode/content/pull/4706 we encountered that using a comma separated list for RHEL7 STIG ID doesn't work.

### DIFF
--- a/shared/transforms/shared_shorthand2xccdf.xslt
+++ b/shared/transforms/shared_shorthand2xccdf.xslt
@@ -176,7 +176,7 @@
                 <xsl:attribute name="system">
                   <xsl:value-of select="$disa-stigs-uri" />
                 </xsl:attribute>
-                <xsl:value-of select="concat(stigid-concat, .)" />
+                <xsl:value-of select="." />
               </xsl:when>
               <xsl:when test="name() = 'custom-cce'">
                 <xsl:attribute name="system">
@@ -341,18 +341,11 @@
           </xsl:if>
         </xsl:attribute>
         <xsl:choose>
-          <xsl:when test="name() = 'stigid'">
-            <xsl:value-of select="normalize-space(concat(stigid-concat, $refitem))" />
+          <xsl:when test="name() = 'disa'">
+            <xsl:value-of select='format-number($refitem, "CCI-000000")' />
           </xsl:when>
           <xsl:otherwise>
-            <xsl:choose>
-              <xsl:when test="name() = 'disa'">
-                <xsl:value-of select='format-number($refitem, "CCI-000000")' />
-              </xsl:when>
-              <xsl:otherwise>
-                <xsl:value-of select="normalize-space($refitem)" />
-              </xsl:otherwise>
-            </xsl:choose>
+            <xsl:value-of select="normalize-space($refitem)" />
           </xsl:otherwise>
         </xsl:choose>
       </reference>
--- a/ssg/build_yaml.py
+++ b/ssg/build_yaml.py
@@ -683,17 +683,15 @@ class Rule(object):
         return rule
 
     def _make_stigid_product_specific(self, product):
+        stig_id = self.references.get("stigid", None)
+        if not stig_id:
+            return
+        if "," in stig_id:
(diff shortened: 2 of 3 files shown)


matejak:
The `tests/unit/ssg-module/test_build_yaml.py` unit test needs to be updated, as we now don't have STIG IDs where the ID has the product-prefix s.a. `RHEL-07-27445-6`.

redhatrises:
Yes. This is very intentional and should not be changed.

matejak:
@redhatrises Could you please explain in greater detail? This PR allows rules to have multiple STIG IDs, and it doesn't change anything for rules that have one or zero STIG IDs. The product-prefix I was referring to is related to input, i.e. to yaml files, the STIG ID appearance in the datastream won't change.

redhatrises:
> @redhatrises Could you please explain in greater detail? This PR allows rules to have multiple STIG IDs, and it doesn't change anything for rules that have one or zero STIG IDs. The product-prefix I was referring to is related to input, i.e. to yaml files, the STIG ID appearance in the datastream won't change.

There can only be 1 STIGID to a rule not multiple. Enabling multiple STIG IDs to a rule is introducing a bug.

jan-cerny:
@redhatrises Aha, OK, I didn't know that. Then this patch isn't needed.

But, because this limitation is not obvious, I suggest adding a check to the Python code to verify that there is a single ID and its format is correct.

We also still can merge the unit test change from in 2ba63a9 and then we could remove the code that handled inconsistent usage of `RHEL-06-` prefixes, because we already changed all `rule.yml`s to not use prefixes for both RHEL6 and RHEL7 STIG IDs. (The prefixes are added automatically during the build.)

matejak:
I think that there is nothing wrong with this PR - it gets some things right, and as a side-effect, it opens door to multiple STIG IDs in the content.

If you add some assert `build_yaml.py` that will make sure that at most one STIG ID is allowed, we are good.

@redhatrises Could you please provide a reference that we could use to document the test, so it is easy to determine why two and more STIG IDs are a no-go? I was unable to find this claim on my own.

shawndwells:
> @redhatrises Could you please provide a reference that we could use to document the test, so it is easy to determine why two and more STIG IDs are a no-go? I was unable to find this claim on my own.

Not sure what is meant -- reference to document the test?

Each configuration rule must attest a single configuration action, and receive a single CCE.

Each configuration rule should have a single DISA reference as well. Sometimes DISA's rules accidentally evaluate two configuration settings in a single rule; in those cases we alert them to the mistake and they're separated out.

matejak:
@shawndwells Why have you closed that PR? Right now, there is nothing that could stop content authors to define multiple STIG IDs to one rule. This PR can fix this easily.
By reference I meant an answer to a question that an uninformed content creator may have in the future - "why it is not possible for one rule to have multiple STIG IDs - is it an artificial limitation, or is there a real reason for it"?

jan-cerny:
@matejak I have changed the code to raise an exception when somebody tries to have multiple STIG IDs in a rule.

redhatrises:
I believe that these changes are good to go. I will create a PR to talk about reference handling in the documentation... I thought I did already did that to reduce confusion for authors, but I must have dreamt it!

jan-cerny:
@matejak rebased on the top of master

matejak:
Great, thank you for the effort and everybody for remarks!

PR #4726, opened 2019-08-14
Move some system-tools rules to organized with their respective configuration rules

redhatrises:
Just reorganized some of the rules so that they weren't all in many different places.


jan-cerny:
Nice improvement, it makes HTML guides more cohesive.

PR #4728, opened 2019-08-15
Set login banner message to /etc/issue in RHEL8 OSPP profile.

ggbecker:
#### Description:

- Set login banner message to /etc/issue in RHEL8 OSPP profile. This banner is consumed by rule sshd_enable_warning_banner.

#### Rationale:

- Unless the banner message is set to /etc/issue, SSH logins will show default message from `issue` file:
```
\S
Kernel \r on an \m

```

### DIFF
--- a/rhel8/profiles/ospp.profile
+++ b/rhel8/profiles/ospp.profile
@@ -1222,3 +1222,13 @@ selections:
 
     ## Enable dnf-automatic Timer
     - timer_dnf-automatic_enabled
+
+    #################################################################
+    ## Login Banner
+    #################################################################
+
+    ## Login banner text
+    - login_banner_text=usgcb_default
+
+    ## Set login banner text to /etc/issue file
+    - banner_etc_issue
--- a/rhel8/profiles/ospp.profile
+++ b/rhel8/profiles/ospp.profile
@@ -1228,7 +1228,7 @@ selections:
     #################################################################
 
     ## Login banner text
-    - login_banner_text=usgcb_default
+    - login_banner_text=dod_default
 
     ## Set login banner text to /etc/issue file
     - banner_etc_issue


comps:
Is the usgcb banner exactly the same as the one from OSPP v4.2.1 Configuration Annex? ... https://www.niap-ccevs.org/MMO/PP/-442ConfigAnnex-/#logontext

ggbecker:
> Is the usgcb banner exactly the same as the one from OSPP v4.2.1 Configuration Annex? ... https://www.niap-ccevs.org/MMO/PP/-442ConfigAnnex-/#logontext

It's not. The value of `login_banner_text` should be `dod_banners`. I'll update that.

yuumasato:
> It's not. The value of `login_banner_text` should be `dod_banners`. I'll update that.

`dod_banners` or `dod_default`? 
`dod_banners` is a combination of `dod_default` and `dod_short`, and allows either one.
The text in linked URL is `dod_default`.

ggbecker:
> > It's not. The value of `login_banner_text` should be `dod_banners`. I'll update that.
> 
> `dod_banners` or `dod_default`?
> `dod_banners` is a combination of `dod_default` and `dod_short`, and allows either one.
> The text in linked URL is `dod_default`.

I've updated to use the `dod_default` which should be the right case, unless requirement changes.

yuumasato:
I was going to suggest that `login_banner_text` be interactive, so that one can add its own banner during tailoring with SCAP Workbench. 
https://github.com/ComplianceAsCode/content/blob/30e5cd50448e63e99f59a137c41362b796ebd66b/linux_os/guide/system/accounts/accounts-banners/login_banner_text.var#L13

But the banner values are actually nasty regexes... so I don't think so.

ggbecker:
The banner content is supposed to be defined by the organization itself. That means we cannot set a default there as it will be different for each case.

Ensuring that `sshd_enable_warning_banner` is properly checked and remediated is enough.

PR #4731, opened 2019-08-15
Improve build section in developer guide

mildas:
#### Description:
Merges part how to build all security content and security content for one product and removes unnecessarily `cd content` commands.

#### Rationale:
- The process of building all security content and one product content is almost same,
- `cd content` command is unnecessary because user should already be in project base directory and we don't need to explicitly mention it every time for every build task
- it is shorter, but with same amount of information for developer


matejak:
Thank you for the improvement!

PR #4732, opened 2019-08-15
Remove excess escape characters

jan-cerny:
#### Description:

Remove excess escape characters

#### Rationale:

They are displayed in HTML report which doesn't look nice.


matejak:
Nice catch!

PR #4733, opened 2019-08-15
Enable the rngd service for OSPP.

adelton:
#### Description:

- Enable the rngd service for OSPP.

#### Rationale:

- The Hardware RNG Entropy Gatherer service should be enabled.


linux_os/guide/services/rng/service_rngd_enabled/rule.yml
```diff
@@ -0,0 +1,25 @@
+documentation_complete: true
+
+prodtype: rhel8
```
jan-cerny: You should add `ol8` and `fedora` here because you add the rule in Fedora and OL8 profile.

redhatrises: Agreed with @jan-cerny but looks good so far. Nice job @adelton!

adelton: Added, rebased on master -> 560ef9499bd31c684fb71796f56be479cb76d868.

fedora/templates/csv/services_enabled.csv
```diff
@@ -5,5 +5,6 @@ firewalld,firewalld,
 ntpd,ntp,
 pcscd,pcsc-lite,
 postfix,,
+rngd,rng-tools,
```
yuumasato: Generation of definition for `rng-tools` is covered by 
``` shell
shared/templates/csv/packages_installed.csv:rng-tools
```

PR #4735, opened 2019-08-15
Adding SELinux rules into OSPP profile

dahaic:
Basic expectation that SELinux is in enforcing, and policy is set to
targeted should be part of the profile.


adelton:
LGTM.

PR #4737, opened 2019-08-16
Configure tmux to lock session after inactivity

adelton:
#### Description:

- Configure tmux to lock session after inactivity.

#### Rationale:

- Locking the session after a period of inactivity limits the potential exposure if the session is left unattended.


linux_os/guide/system/accounts/accounts-physical/screen_locking/console_screen_locking/configure_tmux_lock_after_time/ansible/shared.yml
```diff
@@ -0,0 +1,6 @@
+# platform = multi_platform_all
```
mildas: Please restrict platforms to fedora, rhel8 and ol8 as in oval and bash.

adelton:
Updated platform, rebased on master -> 3dbe8c1b4f09de559aada68c2f459d19282e9472.

mildas:
Thank you

Issue #4722, opened 2019-08-14
Rules not selected in any profile do not have OVAL and fix included in datastream


matusmarhefka:
#### Description of problem:
If the rule is not selected in any profile of some product, build system is not adding OVAL and fix for it into the datastream.

#### Steps to Reproduce:
1. Find a rule not selected in any profile (of some product) and build that product
2. Open a datastream file and check that the OVAL+fix for such rule is missing

yuumasato:
@matusmarhefka It is not clear to me if:
  - the build system is not adding the checks for the rules that are not selected anywhere, or;
  - rules which happen to be not selected anywhere don't have check implemented (not all rules have checks, nor will have)

Can you give example of rule which is not selected in any profile, that is expected to have OVAL check, but the datastream doesn't have it?

matusmarhefka:
@yuumasato The original report here is actually a root cause of the issue, basically when a templated rule was not selected in the specific product we forgot to update csv file for this rule in that product which caused that templated check and fix for such a rule was not added into the datastream.

Example can be the templated `service_sshd_enabled` rule which is not selected in any `rhel8` profile, but it is selected in some of the `rhel7` profiles. It has been fixed by the https://github.com/ComplianceAsCode/content/pull/4739

ggbecker:
My understanding is that only rules which are templated suffer from this problem. If you don't have the respective line in the csv file for some specific rule then it won't get OVAL and remediation into the datastream. Rules which have normal OVAL and remediation (under oval/bash/ansible directory) will not be affected by this issue.

matusmarhefka:
There are also some `notchecked` rules which are missing OVAL checks, should these be included into datastreams?

yuumasato:
> There are also some `notchecked` rules which are missing OVAL checks, should these be included into datastreams?

I think so.
All rules that can be checked should be part of the datastream, and almost all rules are check-able.
Not all are remediable, though.

The value I see is that a rule without check makes it evident that a profile's coverage is not complete, it doesn't give false sense of completeness.

I'm updating title of PR

Mab879:
No longer valid.

PR #4739, opened 2019-08-16
Expanded and cleaned up csv templates.

matejak:
Synced, expanded and reordered the RHEL7 and RHEL8 `.csv` files.

May_fix #4722.


rhel8/templates/csv/services_enabled.csv
```diff
@@ -1,10 +1,14 @@
-auditd,audit,
-chronyd,chrony,
+auditd,,
+chronyd_or_ntpd,chrony,
```
yuumasato: Comment in https://github.com/ComplianceAsCode/content/pull/4739/files#diff-6359728e005606555ad968d6eb8d980bR4 says it needs to be static.

Is it still the case?

rhel7/templates/csv/services_enabled.csv
```diff
@@ -1,5 +1,5 @@
 auditd,,
-auditd,audit,
+auditd,,
```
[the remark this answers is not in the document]
matejak: Oh, so the rule `linux_os/guide/system/auditing/package_auditd_installed` is wrong and should be renamed, right?

rhel8/templates/csv/services_enabled.csv
matejak: I see now - there is no need for template here. The rule name is `service_chronyd_or_ntpd_enabled`, so it can't have a template, as templates require the rule ID to contain the service name.

rhel7/templates/csv/services_enabled.csv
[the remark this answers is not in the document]
yuumasato: Yes, nice catch.

rhel8/templates/csv/services_enabled.csv
yuumasato: Definition for `service_chronyd_or_ntpd_enabled` still depends on the service checks for `chronyd`and `ntpd` though.

matejak: I see, it is about creating OVALs for extended definitions. Fixed by rebase.

rhel8/templates/csv/sysctl_values.csv
```diff
@@ -38,3 +38,4 @@ net.ipv6.conf.all.disable_ipv6,1
 net.ipv6.conf.default.accept_source_route,
 net.ipv6.conf.all.accept_source_route,
 net.ipv6.conf.all.forwarding,
+user.max_user_namespaces,1
```
yuumasato: Correct value is `0`, https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/permissions/restrictions/sysctl_user_max_user_namespaces/rule.yml.

Plus, can you leave this one commented out for now? This rule is optional.

matejak: I have thought that the idea is that any rule included in the benchmark should have a check, and if it is not needed for e.g. OSPP profile, it simply won't be selected. Is this incorrect?

rhel7/templates/csv/services_disabled.csv
```diff
@@ -45,6 +45,7 @@ squid,,
 sshd,openssh-server,
 sssd,,
 sysstat,sysstat,
+systemd-coredump,systemd,
```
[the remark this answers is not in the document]
matejak: Good catch, thanks!

matejak:
I have rebased, so those two fixes (remove RHEL7 line + arg of sysctl set to 0) fixes to be part of this commit: https://github.com/ComplianceAsCode/content/pull/4739/commits/a6a05e225cc3da959a920bf5429f3c4ffd43b8a7

rhel8/templates/csv/sysctl_values.csv
yuumasato: The optional rules are to be in the profile, but without check or remediation. So that compliance to is not enforced.

yuumasato: This can be approached in different way, by using `refine-rule` to make changes in a rules `role` within a profile.

PR #4742, opened 2019-08-19
Prevent user from disabling the screen lock.

adelton:
#### Description:

- Prevent user from disabling the screen lock.

#### Rationale:

- Not listing tmux among permitted shells prevents malicious program running as user from lowering security by disabling the screen lock.


adelton:
Rebased on master -> d33a304d71c11b2c08ad5c4388630c2558c0b1bb.

jan-cerny:
@adelton You usually add rules also to Fedora profile, why not this time?

adelton:
The `configure_bashrc_exec_tmux` (which this one is copy of) did not have it either. That one is a copy of `configure_tmux_lock_after_time` which is a copy of `configure_tmux_lock_command` which finally is a rule that I did not author, and it does not have Fedora as well.

mildas:
People can tailor Fedora profile with these tmux rules. If we would want them in, we can add them all together. Right now it would be mess, if we add this rule to the profile and other tmux rules wouldn't be there.
This PR LGTM and if you want tmux rules in Fedora profile, add them all together in new PR. Thank you

PR #4746, opened 2019-08-20
Convert include_dconf_settings.sh to Jinja macros

jan-cerny:
#### Description:
Rewrite Bash functions from include_dconf_settings to Jinja macros that generate them.

#### Rationale:
This is another part of the Jinja converting effort.
Similar to #4690


linux_os/guide/system/software/gnome/gnome_screen_locking/dconf_gnome_screensaver_idle_delay/bash/shared.sh
```diff
@@ -2,7 +2,5 @@
 . /usr/share/scap-security-guide/remediation_functions
 populate inactivity_timeout_value
 
-include_dconf_settings
-
-dconf_settings 'org/gnome/desktop/session' 'idle-delay' "uint32 ${inactivity_timeout_value}" 'local.d' '00-security-settings'
-dconf_lock 'org/gnome/desktop/session' 'idle-delay' 'local.d' '00-security-settings-lock'
+{{{ dconf_settings("org/gnome/desktop/session", "idle-delay" "uint32 ${inactivity_timeout_value}" "local.d", "00-security-settings") }}}
```
jan-cerny: missing commas

linux_os/guide/system/software/gnome/gnome_screen_locking/dconf_gnome_screensaver_lock_delay/bash/shared.sh
```diff
@@ -2,7 +2,5 @@
 . /usr/share/scap-security-guide/remediation_functions
 populate var_screensaver_lock_delay
 
-include_dconf_settings
-
-dconf_settings 'org/gnome/desktop/screensaver' 'lock-delay' "uint32 ${var_screensaver_lock_delay}" 'local.d' '00-security-settings'
-dconf_lock 'org/gnome/desktop/screensaver' 'lock-delay' 'local.d' '00-security-settings-lock'
+{{{ dconf_settings("org/gnome/desktop/screensaver", "lock-delay" "uint32 ${var_screensaver_lock_delay}" "local.d", "00-security-settings") }}}
```
jan-cerny: missing commas

shared/macros-bash.jinja
```diff
@@ -107,6 +107,55 @@ if test -x /usr/sbin/prelink; then
...
 {{%- endmacro -%}}
 
+{{#
+# Macro to configure DConf settings for RHEL and Fedora systems.
+#}}
+{{%- macro dconf_settings(path, key, value, db, setting_file) -%}}
+# Check for setting in any of the DConf db directories
+# If files contain ibus or distro, ignore them.
+# The assignment assumes that individual filenames don't contain :
+readarray -t SETTINGSFILES < <(grep -r "\\[{{{ path }}}\\]" "/etc/dconf/db/" | grep -v "distro\\|ibus" | cut -d":" -f1)
```
matejak: When you migrate from double quotes to single quotes in bash, double backslashes inside those double quotes can be converted to single backslashes (in single quotes). However, this line contains the `path` macro expansion which can be a bash variable, so quotes should remain double. But `"distro\\|ibus"` can be converted to `'distro\|ibus'` and so on.

shared/macros-bash.jinja
```diff
@@ -107,6 +107,55 @@ if test -x /usr/sbin/prelink; then
...
+DCONFFILE="/etc/dconf/db/{{{ db }}}/{{{ setting_file }}}"
+DBDIR="/etc/dconf/db/{{{ db }}}"
+
+mkdir -p "${DBDIR}"
+
+if [ ${#SETTINGSFILES[@]} -eq 0 ]
```
matejak: Please add quoting to variable expansion, so the code is somewhat more robust and we are not spammed by `shellcheck` later.

jan-cerny:
@matejak done

docs/manual/developer_guide.adoc
```diff
@@ -1208,6 +1208,9 @@ Available high-level Jinja macros to generate Bash remediations:
 - `bash_coredump_config_set` -  Set Coredump configuration in `/etc/systemd/coredump.conf`
 - `bash_package_install` - Install a package
 - `bash_package_remove` - Remove a package
+- `disable_prelink` - disables prelinking
```
cipherboy: Question: if these are going to be bash remediations, do we want to prefix them with `bash_`? Then if we get a similar macro for Ansible, we can call it `ansible_`. I think it'd be cool if we could have a `disable_prelink` macro that'll call `ansible_disable_prelink` or `bash_disable_prelink` depending on execution context but I don't know if that's possible.  (The other issue with that is that sometimes the ansible mode might support more features than the bash version does, e.g., `lineinfile` in Ansible uses PCRE regexes which I haven't found a good way of supporting in bash without requiring Python or Perl or something and "rewriting" that portion of the Ansible module).

cipherboy: Ah I see, `disable_prelink` was added in a previous PR but wasn't documented. Fixing that is out of scope for this PR then. Should I open a separate PR when this one merges and fix that for all three?

jan-cerny: @cipherboy great point, I agree, I didn't think about that so I kept the original name, but adding a prefix would greatly improve it. Yes, a separate PR would be great!

jan-cerny:
I have added `bash_` prefix to Jinja Dconf Bash macros.

cipherboy:
\o/ Thanks @jan-cerny! I'm fine if @matejak is. :)

jan-cerny:
@matejak Are you fine?

matejak:
Sure, thank you for the PR!

PR #4747, opened 2019-08-20
Log USBGuard daemon audit events using Linux Audit.

adelton:
#### Description:

- Log USBGuard daemon audit events using Linux Audit.

#### Rationale:

- Using the Linux Audit logging allows for centralized trace of events.


adelton:
Added `reference` -> 9dd6b4d74ed5adce61f1000a291b66c9f62482b3.

mildas:
`separator_regex` in OVAL differs from `separator_regex` in bash. Can you also create ansible remediation?

adelton:
The `separator_regex` in `bash/shared.sh` fixed -> 0462ad0291d70edf09b13cb7e31a0e53b43d4536.

I'd prefer not to block on the ansible remediation at this point.

mildas:
@adelton Ok, I won't insist and block on ansible remediation, but I still don't know about the `separator_regex`. You removed `\s*`, but the correct is with `\s*` (I mean `\s*=\s*`) in my opinion. Are you sure that with `\s*` it's wrong? I didn't find any documentation about usbguard-daemon.conf syntax and tried usbguard-daemon with spaces in config, and it successfully started.

adelton:
The default `.conf` file is without spaces and the documentation does not say anything about spaces. It's possible that it works with spaces but it does not seem guaranteed, and given the rest of the `.conf` file does not use them, I don't see a reason to accept them as valid config lines. The worst thing that might happen is we will mark such line and `fail` and remediate it to non-spaced form.

mildas:
I'm okay with that. Created test scenarios and they are passing, so looks good to me. Thank you

PR #4749, opened 2019-08-20
Prevent failures of Ansible task for service_enabled and timer_enabled

jan-cerny:
#### Description:
If the package which provides the service or timer is not present the Ansible task that enables the service or timer will fail because it can not find the unit file which will cause termination of the whole playbook. We can prevent the failure by executing the task only if the given package is present. The code uses Ansible package_facts module to determine if package is present.

#### Rationale:

Partially fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1741455

Addressing:
```
TASK [Enable service fapolicyd] ************************************************
fatal: [192.168.122.24]: FAILED! => {"changed": false, "msg": "Could not find the requested service fapolicyd: host"}
```

### DIFF
--- a/shared/templates/template_ANSIBLE_service_enabled
+++ b/shared/templates/template_ANSIBLE_service_enabled
@@ -5,7 +5,7 @@
 # disruption = low
 - name: Enable service {{{ SERVICENAME }}}
   block:
-  - name: Gather the rpm package facts
+  - name: Gather the package facts
     package_facts:
       manager: auto
 
--- a/shared/templates/template_ANSIBLE_service_enabled
+++ b/shared/templates/template_ANSIBLE_service_enabled
@@ -5,17 +5,14 @@
 # disruption = low
 - name: Enable service {{{ SERVICENAME }}}
   block:
-  - name: Check if {{{ PACKAGENAME }}} is installed
-    package:
-      name: {{{ PACKAGENAME }}}
-      state: present
-    check_mode: true
-    register: pkg
+  - name: Gather the rpm package facts
+    package_facts:
+      manager: auto
+
   - name: Enable service {{{ SERVICENAME }}}
     service:
       name: "{{{ DAEMONNAME }}}"
       enabled: "yes"
       state: "started"
-    # package would be installed if package module was not run in check mode
-    when: not pkg.changed
-
+    when: |-
+      '{{{ PACKAGENAME }}}' in ansible_facts.packages
--- a/shared/templates/template_ANSIBLE_service_enabled
+++ b/shared/templates/template_ANSIBLE_service_enabled
@@ -14,5 +14,5 @@
       name: "{{{ DAEMONNAME }}}"
       enabled: "yes"
       state: "started"
-    when: |-
-      '{{{ PACKAGENAME }}}' in ansible_facts.packages
+    when:
+    - '"{{{ PACKAGENAME }}}" in ansible_facts.packages'
--- a/shared/templates/template_ANSIBLE_service_enabled
+++ b/shared/templates/template_ANSIBLE_service_enabled
@@ -4,8 +4,18 @@
 # complexity = low
 # disruption = low
 - name: Enable service {{{ SERVICENAME }}}
-  service:
-    name: "{{{ DAEMONNAME }}}"
-    enabled: "yes"
-    state: "started"
+  block:
+  - name: Check if {{{ PACKAGENAME }}} is installed
+    package:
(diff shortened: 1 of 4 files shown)


shared/templates/template_ANSIBLE_service_enabled
```diff
@@ -4,8 +4,18 @@
...
-    enabled: "yes"
-    state: "started"
+  block:
+  - name: Check if {{{ PACKAGENAME }}} is installed
+    package:
+      name: {{{ PACKAGENAME }}}
+      state: present
+    check_mode: true
+    register: pkg
+  - name: Enable service {{{ SERVICENAME }}}
```
redhatrises: Add a blank line for readability please.

redhatrises:
I wonder if it would be better to run `package_facts.packages[my_package_name]` instead of the `package` module.

jan-cerny:
@openscap-ci test this please

shared/templates/template_ANSIBLE_timer_enabled
```diff
@@ -4,8 +4,15 @@
...
 - name: Enable timer {{{ TIMERNAME }}}
-  systemd:
-    name: "{{{ TIMERNAME }}}.timer"
-    enabled: "yes"
-    state: "started"
+  block:
+  - name: Gather the package facts
+    package_facts:
+      manager: auto
+  - name: Enable timer {{{ TIMERNAME }}}
```
redhatrises: Need a blank line before `-name` here as well

shared/templates/template_ANSIBLE_timer_enabled
```diff
@@ -4,8 +4,15 @@
...
+    systemd:
+      name: "{{{ TIMERNAME }}}.timer"
+      enabled: "yes"
+      state: "started"
+    when: |-
+      '{{{ PACKAGENAME }}}' in ansible_facts.packages
```
redhatrises: Let's make this:
```
   when:
     - '{{{ PACKAGENAME }}}' in ansible_facts.packages
```

shared/templates/template_ANSIBLE_service_enabled
```diff
@@ -4,8 +4,15 @@
...
+    package_facts:
+      manager: auto
 
+  - name: Enable service {{{ SERVICENAME }}}
+    service:
+      name: "{{{ DAEMONNAME }}}"
+      enabled: "yes"
+      state: "started"
+    when: |-
+      '{{{ PACKAGENAME }}}' in ansible_facts.packages
```
redhatrises: Let's make this:
```
   when:
     - '{{{ PACKAGENAME }}}' in ansible_facts.packages
```

shared/templates/template_ANSIBLE_timer_enabled
jan-cerny: fixed

shared/templates/template_ANSIBLE_timer_enabled
jan-cerny: fixed

shared/templates/template_ANSIBLE_service_enabled
jan-cerny: fixed

yuumasato:
LGTM, but every task for service enabled will be checking for the list of packages installed.
Similar to what happens here:
https://github.com/ComplianceAsCode/content/blob/3147bfe95b4154de447f92b82c44e222ea6d4995/shared/templates/template_ANSIBLE_sebool_var#L8

I'm afraid there is no easy way to avoid this, unless generation of Ansible playbooks is smart enough to de-duplicate tasks. And the block may actually make it harder to de-duplicate.

redhatrises:
> LGTM, but every task for service enabled will be checking for the list of packages installed.
> Similar to what happens here:
> 
> https://github.com/ComplianceAsCode/content/blob/3147bfe95b4154de447f92b82c44e222ea6d4995/shared/templates/template_ANSIBLE_sebool_var#L8
> 
> I'm afraid there is no easy way to avoid this, unless generation of Ansible playbooks is smart enough to de-duplicate tasks. And the block may actually make it harder to de-duplicate.

@yuumasato this is more of the correct way of checking if a package is installed rather than what is in the sebool_var. You will always have to run package_facts more than once if validating that a package is installed.

jan-cerny:
> generation of Ansible playbooks is smart enough to de-duplicate tasks

We build that using `oscap xccdf generate fix` so it would need to greatly improve this part of `oscap` preferably in a  way that it understands YAML

PR #4750, opened 2019-08-20
Add suport for rule refinement

yuumasato:
#### Description:

- Implement support for `xccdf:refine-rule`
  - Allows `weight`, `role` and `severity` of a rule to be modified on Profile basis.
- Syntax is: `- rule_id.property=value`
  - Usage is illustrated by rule `sysctl_user_max_user_namespaces` in `ospp` profile.

#### Rationale:
- Optional rules in `ospp` profile can be in the Benchmark as `role=unscored` and its evaluation results will not affect result of evaluation. Rule will be reported as "information"
- `severity=info` is set just for consistency, `role=unscored` is enough to make rule totally informational.
- `informational` rules are never fixed by OpenSCAP
- Kind of expected caveat, all profiles extending `ospp` profile that would like to use the rule "for real" will have to refine the rule again.
  - For example, profile `cui`, also ended up with rule `sysctl_user_max_user_namespaces` as informational.

Report for informational rule:
- Report when evaluation fails: [unscored_report.html.txt](https://github.com/ComplianceAsCode/content/files/3522793/unscored_report.html.txt)
- Report when evaluation passes: [unscored_report_pass.html.txt](https://github.com/ComplianceAsCode/content/files/3522807/unscored_report_pass.html.txt)

I recommend to remove `.txt` extension before pointing your favorite HTML reader to it.

### DIFF
--- a/rhel8/profiles/ospp.profile
+++ b/rhel8/profiles/ospp.profile
@@ -433,6 +433,7 @@ selections:
 
     ## Disable the use of user namespaces
     - sysctl_user_max_user_namespaces
+    - sysctl_user_max_user_namespaces.role=unscored
 
     ## Disable Access to Network bpf() Syscall From Unprivileged Processes
     - sysctl_kernel_unprivileged_bpf_disabled
--- a/rhel8/profiles/ospp.profile
+++ b/rhel8/profiles/ospp.profile
@@ -434,6 +434,7 @@ selections:
     ## Disable the use of user namespaces
     - sysctl_user_max_user_namespaces
     - sysctl_user_max_user_namespaces.role=unscored
+    - sysctl_user_max_user_namespaces.severity=info
 
     ## Disable Access to Network bpf() Syscall From Unprivileged Processes
     - sysctl_kernel_unprivileged_bpf_disabled
--- a/build-scripts/yaml_to_shorthand.py
+++ b/build-scripts/yaml_to_shorthand.py
@@ -70,9 +70,9 @@ def main():
 
         profiles = loader.loaded_group.profiles
         for p in profiles:
-            p.validate_refine_rules(loader.all_rules)
             p.validate_variables(loader.all_values)
             p.validate_rules(loader.all_rules, loader.all_groups)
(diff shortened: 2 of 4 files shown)


yuumasato:
> 1 - Is there any documentation on how to use these refinements?

Do you mean documentation on syntax and properties that can be refined?

>  2 - Does it make sense to check whether a rule is selected within a profile when doing refinement?

I think the XCCDF standard doesn't care.
But it makes sense to our use case, to check that developer intentions will be fulfilled.

ggbecker:
> > 1 - Is there any documentation on how to use these refinements?
> 
> Do you mean documentation on syntax and properties that can be refined?
> 
Exactly, more important the syntax part as the properties I assume they can be found in XCCDF standard.
> > 2 - Does it make sense to check whether a rule is selected within a profile when doing refinement?
> 
> I think the XCCDF standard doesn't care.
> But it makes sense to our use case, to check that developer intentions will be fulfilled.

Please add the check then. Thank you.

yuumasato:
@ggbecker I have added the check for refinement of rule selected.

Regarding documentation, I feel appropriate place would be a "profile" section in DevGuide, but currently there is no such thing.
I propose to document in subsequent PR.

rhel8/profiles/ospp.profile
```diff
@@ -432,7 +432,8 @@ selections:
     - sysctl_kernel_kexec_load_disabled
 
     ## Disable the use of user namespaces
-    - sysctl_user_max_user_namespaces
```
ggbecker: whoops ;) can you reintroduce the rule?

ggbecker:
> @ggbecker I have added the check for refinement of rule selected.
> 
> Regarding documentation, I feel appropriate place would be a "profile" section in DevGuide, but currently there is no such thing.
> I propose to document in subsequent PR.

Okay, I'm fine with not blocking the PR. Documentation can be added later in a bigger scope of profile docs.

rhel8/profiles/ospp.profile
yuumasato: Fixed by rebase.

PR #4751, opened 2019-08-21
Remove remaining CCE identifiers from RHEL derivatives.

ggbecker:
#### Description:

- Remove remaining CCE identifier references from RHEL datastream derivatives under the `definition` (OVAL check) tag.

#### Rationale:

- CCE identifiers are bound to RHEL and should not appear in derivatives datastreams

Additional information:
- Values such as:
```
<ns3:reference ref_id="CCE-80657-0" source="CCE"/>
```

### DIFF
--- a/ssg/build_derivatives.py
+++ b/ssg/build_derivatives.py
@@ -108,6 +108,17 @@ def remove_idents(tree_root, namespace, prod="RHEL"):
                 fix.text = re.sub(r"CCE-[0-9]*-[0-9]*", "", fix.text)
 
 
+def remove_cce_reference(tree_root, namespace, prod="RHEL"):
+    """
+    Remove CCE identifiers from OVAL checks in XML tree
+    """
+    for definition in tree_root.findall(".//{%s}definition" % (namespace)):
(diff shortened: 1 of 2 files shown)


ssg/build_derivatives.py
```diff
@@ -108,6 +108,17 @@ def remove_idents(tree_root, namespace, prod="RHEL"):
                 fix.text = re.sub(r"CCE-[0-9]*-[0-9]*", "", fix.text)
 
 
+def remove_cce_reference(tree_root, namespace, prod="RHEL"):
```
matejak: The `prod` argument is not used in the function, so what about removing it?

ggbecker: Removed it.

ggbecker:
> There is just a minor style nitpick, but besides that, I have verified that the PR works, so it is good to merge.

@matejak  I fixed the style issue. Can you review it again?

jan-cerny:
@ggbecker I'm taking this because @matejak isn't available

jan-cerny:
LGTM

PR #4689, opened 2019-08-08
Build only SCAP 1.3 content

yuumasato:
#### Description:

- Build only one version of SCAP content at a time. No more `-1.3` suffixes.
- Allow control of which SCAP version to build with `SSG_TARGET_SCAP_VERSION`
  - By default, build SCAP 1.3 content, with OVAL 5.11
  - If SCAP 1.2 is selected, use OVAL 5.10

#### Rationale:

- SCAP 1.3 is the latest version and supersedes and deprecates SCAP 1.2
- With OpenSCAP `--fetch-remote-resources` bug on SCAP 1.3 content fixed, let's default to building SCAP 1.3.

PR #4755, opened 2019-08-21
Build both scap versions, but 1.3 is default

yuumasato:
#### Description:

- Follow up from #4689, build both SCAP versions, instead of only one of them.
- Default means that the datastream named `ssg-<product>-ds.xml` is SCAP 1.3.
- The SCAP 1.2 datastream will be named `ssg-<product>-ds-1.2.xml`.
- Updated `build_script` command with option to generate fully compliant SCAP 1.2 with `--oval510`
- Docs also updated

EDIT:
- Also, `nist-zipfile` target changed to `vendor-zipfile` and documented.

#### Rationale:
- SCAP 1.3 is still not widely supported, until then build SCAP 1.2 content

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,16 +36,8 @@ set(SSG_VERSION "${SSG_MAJOR_VERSION}.${SSG_MINOR_VERSION}.${SSG_PATCH_VERSION}"
 
 set(SSG_VENDOR "ssgproject" CACHE STRING "Specify the XCCDF 1.2 vendor string.")
 
-set(SSG_TARGET_SCAP_VERSION "1.3" CACHE STRING "Which SCAP version we are targetting. Possible choices are 1.3 and 1.2")
-
 set(SSG_TARGET_OVAL_MAJOR_VERSION "5" CACHE STRING "Which major version of OVAL are we targetting. Only 5 is supported at the moment.")
 set(SSG_TARGET_OVAL_MINOR_VERSION "11" CACHE STRING "Which minor version of OVAL are we targetting. Possible choices are 10 or 11.")
-
-# If targetting SCAP 1.2, target OVAL version is 5.10
-if (SSG_TARGET_SCAP_VERSION VERSION_EQUAL "1.2")
-    set(SSG_TARGET_OVAL_MAJOR_VERSION "5" CACHE STRING "Which major version of OVAL are we targetting. Only 5 is supported at the moment.")
-    set(SSG_TARGET_OVAL_MINOR_VERSION "10" CACHE STRING "Which minor version of OVAL are we targetting. Possible choices are 10 or 11.")
-endif()
 
 set(SSG_TARGET_OVAL_VERSION "${SSG_TARGET_OVAL_MAJOR_VERSION}.${SSG_TARGET_OVAL_MINOR_VERSION}")
 
@@ -140,10 +132,6 @@ if (NOT SSG_TARGET_OVAL_VERSION VERSION_EQUAL "5.10" AND NOT SSG_TARGET_OVAL_VER
     message(WARNING "You are targetting OVAL version ${SSG_TARGET_OVAL_VERSION}. In SSG we support/test 5.10 and 5.11 only!")
 endif()
 
-if (NOT SSG_TARGET_SCAP_VERSION VERSION_EQUAL "1.3" AND NOT SSG_TARGET_SCAP_VERSION VERSION_EQUAL "1.2")
-    message(WARNING "You are targetting SCAP version ${SSG_TARGET_SCAP_VERSION}. In SSG we support/test 1.3 and 1.2 only!")
-endif()
-
 execute_process(
     COMMAND "${PYTHON_EXECUTABLE}" "${SSG_BUILD_SCRIPTS}/oscap_svg_support.py" "${OPENSCAP_OSCAP_EXECUTABLE}"
     RESULT_VARIABLE OSCAP_SVG_SUPPORT_RESULT
@@ -222,7 +210,6 @@ message(STATUS " ")
 
 message(STATUS "Build options:")
 message(STATUS "SSG vendor string: ${SSG_VENDOR}")
-message(STATUS "Target SCAP version: ${SSG_TARGET_SCAP_VERSION}")
 message(STATUS "Target OVAL version: ${SSG_TARGET_OVAL_VERSION}")
 message(STATUS "OVAL schematron validation: ${SSG_OVAL_SCHEMATRON_VALIDATION_ENABLED}")
 message(STATUS "shellcheck bash fixes validation: ${SSG_SHELLCHECK_BASH_FIXES_VALIDATION_ENABLED}")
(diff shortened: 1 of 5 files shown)


ggbecker:
Please add ctest for datastream validation to 1.2 datastream:

in SSGCommon.cmake:
```

    add_test(
        NAME "validate-ssg-${PRODUCT}-ds-1.2.xml"
        COMMAND "${OPENSCAP_OSCAP_EXECUTABLE}" ds sds-validate "${CMAKE_BINARY_DIR}/ssg-${PRODUCT}-ds-1.2.xml"
    )
```

yuumasato:
> Please add ctest for datastream validation to 1.2 datastream:

Should be addressed in b7e4437 

I also updated the cmake commands for the derivative products.

cmake/SSGCommon.cmake
```diff
@@ -571,7 +577,7 @@ macro(ssg_build_sds PRODUCT)
             DEPENDS generate-ssg-${PRODUCT}-cpe-dictionary.xml
             DEPENDS "${CMAKE_BINARY_DIR}/ssg-${PRODUCT}-cpe-dictionary.xml"
             DEPENDS "${CMAKE_BINARY_DIR}/ssg-${PRODUCT}-cpe-oval.xml"
-            COMMENT "[${PRODUCT}-content] generating ssg-${PRODUCT}-ds.xml"
+            COMMENT "[${PRODUCT}-content] generating ssg-${PRODUCT}-ds.xm and lssg-${PRODUCT}-ds-1.2.xml"
```
ggbecker:
```suggestion
            COMMENT "[${PRODUCT}-content] generating ssg-${PRODUCT}-ds.xm and ssg-${PRODUCT}-ds-1.2.xml"
```

ggbecker:
As the datastream naming convention is changing, please change here as well:

https://github.com/ComplianceAsCode/content/blob/365a295ca049429e3d7611296b03f94e9b671a6e/tests/run_scapval.py#L112

yuumasato:
@ggbecker Nice catch, should be fixed.

cmake/SSGCommon.cmake
```diff
@@ -571,7 +577,7 @@ macro(ssg_build_sds PRODUCT)
             DEPENDS generate-ssg-${PRODUCT}-cpe-dictionary.xml
             DEPENDS "${CMAKE_BINARY_DIR}/ssg-${PRODUCT}-cpe-dictionary.xml"
             DEPENDS "${CMAKE_BINARY_DIR}/ssg-${PRODUCT}-cpe-oval.xml"
-            COMMENT "[${PRODUCT}-content] generating ssg-${PRODUCT}-ds.xml"
+            COMMENT "[${PRODUCT}-content] generating ssg-${PRODUCT}-ds.xm and ssg-${PRODUCT}-ds-1.2.xml"
```
ggbecker:
```suggestion
            COMMENT "[${PRODUCT}-content] generating ssg-${PRODUCT}-ds.xml and ssg-${PRODUCT}-ds-1.2.xml"
```

CMakeLists.txt
```diff
@@ -367,11 +354,11 @@ endif()
 
 # ZIP only contains source datastreams and kickstarts, people who
 # want sources to build from should get the tarball instead.
-if(SSG_TARGET_SCAP_VERSION VERSION_EQUAL "1.3")
+if(SSG_TARGET_OVAL_VERSION VERSION_EQUAL "5.11")
     ssg_build_zipfile("scap-security-guide-${SSG_VERSION}")
     ssg_build_nist_zipfile("scap-security-guide-${SSG_VERSION}-SCAP-1.3")
 else()
-    ssg_build_zipfile("scap-security-guide-${SSG_VERSION}-SCAP-1.2")
+    ssg_build_zipfile("scap-security-guide-${SSG_VERSION}-oval-${SSG_TARGET_OVAL_VERSION}")
```
ggbecker: `${SSG_TARGET_OVAL_VERSION}` contains `dot` in the value (`5.11`). Release artifacts for OVAL 5.10 are in the form of `scap-security-guide-0.1.45-oval-510.zip`. I wonder if this can impact anyone somehow.

ggbecker:
```suggestion
    ssg_build_zipfile("scap-security-guide-${SSG_VERSION}-oval-${SSG_TARGET_OVAL_MAJOR_VERSION}${SSG_TARGET_OVAL_MINOR_VERSION}")
```
Would this be acceptable?

yuumasato: I just reverted to state before #4689 
https://github.com/ComplianceAsCode/content/pull/4689/files#diff-af3b638bc2a3e6c650974192a53c7291L268

This is a nice catch, as up to version 0.1.40, the 5.10 zip was named suffix `oval-5.10.zip`.

Do you feel strongly about this change?

cmake/SSGCommon.cmake
```diff
@@ -1195,34 +1211,68 @@ macro(ssg_build_zipfile_target ZIPNAME)
...
-        DEPENDS rhel6
-        DEPENDS rhel7
-        DEPENDS rhel8
-        DEPENDS rhosp13
-        DEPENDS rhv4
-        )
+    # When SCAP1.2 zipfiles are needed no more, these custom commands can be unified
+    if("${ZIPNAME}" MATCHES "SCAP-1.3")
+        # Zip only the SCAP1.3 datastream
+        add_custom_command(
```
ggbecker: The only different thing from these two custom commands is the suffix `ds-1.2.xml`. Can you infer from `ZIPNAME` the right suffix and deduplicate the code using some kind of CMake variable?

CMakeLists.txt
ggbecker: It seems that no harm was caused by using a different names. So I'm ok leaving as it is right now.

cmake/SSGCommon.cmake
yuumasato: Good idea, I'll look into it.

ggbecker:
It's also a good opportunity to mention the target `nist-zipfile`, please add a section next to regarding the target availability:
https://github.com/ComplianceAsCode/content/blob/master/docs/manual/developer_guide.adoc#optional-building-a-zip-file

cmake/SSGCommon.cmake
yuumasato: It works, thanks for idea.

yuumasato:
> It's also a good opportunity to mention the target `nist-zipfile`, please add a section regarding the target availability next to:
> https://github.com/ComplianceAsCode/content/blob/master/docs/manual/developer_guide.adoc#optional-building-a-zip-file

@ggbecker This is a bit of scope creep, but I'm taking it.

I have changed target name `nist-zipfile` to `vendor-zipfile`, and added a target specific for Red Hat products.

cmake/SSGCommon.cmake
```diff
@@ -1251,6 +1251,6 @@ macro(ssg_build_vendor_zipfile ZIPNAME)
         redhat-zipfile
         DEPENDS "${CMAKE_BINARY_DIR}/vendor-zipfile/${ZIPNAME}-RedHat.zip"
     )
-    add_custom_target(vendor-zipfile ALL)
+    add_custom_target(vendor-zipfile)
```
ggbecker: I was wondering why my build was always generating the vendor-zipfile when triggering ALL build. ;P

yuumasato:
@ggbecker Thank you for the great review!

PR #4756, opened 2019-08-21
Create configuration files if they don't exist

jan-cerny:
#### Description:
Add "create=yes" to Ansible Tasks where `lineinfile` module is used.

#### Rationale:
Prevents Ansible Playbook from terminating.
Partially fixes https://bugzilla.redhat.com/show_bug.cgi?id=1741455


jan-cerny:
Shouldn't we rather covert everything to macro? Thoughts? @matejak @yuumasato @matusmarhefka @ggbecker @redhatrises @adelton @vojtapolasek @mildas @dahaic

matejak:
The template is tricky, as we have one template per filetype, so the conversion has to be done on per-rule basis, possibly introducing new template types.
Anyway, I don't see any reason why not to merge this PR.

jan-cerny:
We have discussed this with @matusmarhefka and we found that I might miss some files. I havechecked again and I have found that some templates don't create the file. That's probably because I have examined only `linux_os` directory before. I have added modifications of templates in afda9d8.

matusmarhefka:
LGTM

PR #4757, opened 2019-08-21
extend oval check of configure_crypto_policy

vojtapolasek:
#### Description:

This extends oval check of configure_crypto_policy rule. Additionally it allows file in the /etc/crypto-policies/back-ends/ to be regular files, but appropriate file in /etc/crypto-policy/local.d/ must exist. This is with accordance to customizations of crypto policies.

#### Rationale:

This customization of crypto policies is used in some other rule and therefore should not fail.

### DIFF
--- a/linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/oval/shared.xml
+++ b/linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/oval/shared.xml
@@ -1,14 +1,14 @@
 {{%- if target_oval_version == [5, 11] -%}}
 {{# there is no good alternative for symlink_test for OVAL 5.10 #}}
 {{%- macro crypto_policy_symlink_criterion(library) %}}
-      <criteria operator="OR">
+      <criteria operator="OR" comment="checks if /etc/crypto-policies/back-ends/{{{ library }}}.config is either a symlink and in that case a drop in file in /etc/crypto-policies/local.d must not exist or /etc/crypto-policies/back-ends/{{{ library }}}.config is a regular file and the drop in file must exist">
         <criteria operator="AND">
           <criterion comment="Check if /etc/crypto-policies/back-ends/{{{ library }}}.config is linked to /usr/share/crypto-policies/POLICY/{{{ library }}}.txt" test_ref="test_crypto_policy_{{{ library }}}_symlink" />
           <criterion comment="Checks if there is NOT a drop in file in /etc/crypto-policies/local.d associated with {{{ library }}}." test_ref="test_{{{ library }}}_dropin_file" negate="true" />
         </criteria>
         <criteria operator="AND">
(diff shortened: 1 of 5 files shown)


linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/oval/shared.xml
```diff
@@ -1,7 +1,10 @@
 {{%- if target_oval_version == [5, 11] -%}}
 {{# there is no good alternative for symlink_test for OVAL 5.10 #}}
 {{%- macro crypto_policy_symlink_criterion(library) %}}
-      <criterion comment="Check if /etc/crypto-policies/back-ends/{{{ library }}}.config is linked to /usr/share/crypto-policies/POLICY/{{{ library }}}.txt" test_ref="test_crypto_policy_{{{ library }}}_symlink" />
+      <criteria operator="OR">
```
yuumasato: What is the expected behavior if both criterion evaluate to true?

Presence of drop in `local.d/library.config` file and `back-end/library.config`being a symlink would indicates that `update-crypto-policies` was not run recently, and remediation should be run.
So appropriate operator for criteria would be `XOR`.

yuumasato: The approach to explicitly check for them also works.

linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/rule.yml
```diff
@@ -26,6 +27,10 @@ ocil: |-
     following command:
     <pre>$ update-crypto-policies --show</pre>
     The output should return <pre><sub idref="var_system_crypto_policy" /></pre>.
+    Moreover, check if crypto policy customizations are applied correctly.
+    <pre>$ ls -l /etc/crypto-policies/back-ends/</pre>
+    For every file, check if it is a symbolic link or a regular file. If it is a symbolic link, verify that the target resides in the <pre>/usr/share/crypto-policies/<sub idref="var_system_crypto_policy" />/</pre> directory. If it is a regular file, check that there exists a drop in file in the <pre>/etc/crypto-policies/local.d</pre>. The file name of a drop in file must be formed as <pre>LIBRARY_NAME-ARBITRARY_ID.config</pre>.
+    For example, the drop in file for Openssh server can be named <pre>opensshserver-ospp.config</pre>.
```
yuumasato: Make this more direct, with commands and expected output.
For example, command below will give any file which is not a symlink:
```
ls -l /etc/crypto-policies/back-ends | grep "^[^lt]"
-rw-r--r--. 1 root root 429 Jul  9 14:16 nss.config
```
And if any result is returned, check that there are files in `/etc/crypto-policies/local.d` whose name match the libraries which is not symlink'ed.
`ls /etc/crypto-policies/local.d/ | awk -F'-' '{print $1}'` will print only the library names in `local.d` directory.

yuumasato:
Thank you for extending the check, @vojtapolasek.

PR #4758, opened 2019-08-21
Update linux_os guide for ol8

iokomin:
#### Description:

Minor linux_os guide update to generate correct content for ol8 product.

#### Testing:
- checked ol8 build
- checked corresponding guide items to include expected content


ggbecker:
LGTM, thanks for the patch.

PR #4761, opened 2019-08-22
Update CSRF cookie secure

redhatrises:
- Update CSRF_COOKIE_SECURE checks to use standard rule.yml template
- Assign CCEs
- Create NIST mappings.
- Create Ansible snippets

Replaces #4383


jan-cerny:
Thanks

PR #4764, opened 2019-08-24
Update ncp profile

### LINKED ISSUE #4707: ID and Title of new RHEL7 USGCB profile
#### Description of problem:
With the PR #4705 we are cleaning up OSPP profile so it's purely implementation of OSPP guidance. We need to preserve the USGCB profile (originally under ID `ospp`). Aforementioned PR moves it to ID `ncp`. It's not fully understood what is the best ID, but it's preferable to merge the PR and fix this a bit later.

I am setting the BLOCKER label here, as we really need to solve this before the release. And I am assigning Shawn as I believe he has most to say about this :)

dahaic:
#### Description:

- Update NCP profile to have better title and builds upon OSPP
- Only duplicated rules has been removed from the NCP defining profile - the profile itself has only additions inherited from OSPP

#### Rationale:

- NCP profile is meant to be expansion of OSPP to cover more usecases.

- Fixes #4707


yuumasato:
Looks good, thanks @dahaic

PR #4765, opened 2019-08-27
Add SELinux booleans CSV and remove RHEL8 from rules for packages not available

yuumasato:
#### Description:

- Add RHEL8 CSV file for SELinux booleans
- Remove `rhel8` from rules concerning packages not available in RHEL8:
  - `pam_pkcs11`
  - `screen`
  - `python-rhsm`

#### Rationale:
- All rules for SELinux boooleans were "checkless"
- Not applicable rules were available for RHEL8
- Helps with #4722


linux_os/guide/system/software/system-tools/package_python-rhsm_installed/rule.yml
```diff
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: rhel6,rhel7,rhel8,rhv4
```
jan-cerny:
```
$ git grep -l python-rhsm
linux_os/guide/system/software/system-tools/package_python-rhsm_installed/rule.yml
rhel6/kickstart/ssg-rhel6-stig-ks.cfg
```

It seems like this rule has no OVAL in all products anyway.

yuumasato: Yes, just description for the rules is there.

PR #4766, opened 2019-08-27
Fix rpm_verify_hashes and rpm_verify permissions ansible remediations

mildas:
#### Description:
Remediation was failing because pipeline returns code 1.

#### Rationale:
`set -o pipefail` causes, that pipe returns non-zero value, if any command from pipeline ends with non-zero return code. `rpm -Va` gives me every time return code 1 (not sure why, didn't find any info about that, but it's probably because `-Va` return list of files with incorrect hash and if the list is non-empty then it returns non-zero value) and because of remediation pipeline returned 1, it interrupted remediation run (and also ansible playbook).

### DIFF
--- a/linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_hashes/ansible/shared.yml
+++ b/linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_hashes/ansible/shared.yml
@@ -15,7 +15,7 @@
 
 - name: "Read files with incorrect hash"
   shell: |
-    set -o pipefail
+    # Do not set 'set -o pipefail' - 'rpm -Va' returns non-zero value because non-empty list of files
     rpm -Va | grep -E '^..5.* /(bin|sbin|lib|lib64|usr)/' | awk '{print $NF}'
   args:
     warn: False # Ignore ANSIBLE0006, we can't fetch files with incorrect hash using rpm module
--- a/linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_permissions/ansible/shared.yml
+++ b/linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_permissions/ansible/shared.yml
@@ -5,7 +5,7 @@
 # disruption = medium
 - name: "Read list of files with incorrect permissions"
   shell: |
-    set -o pipefail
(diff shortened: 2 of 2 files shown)


linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_hashes/ansible/shared.yml
```diff
@@ -15,7 +15,6 @@
 
 - name: "Read files with incorrect hash"
   shell: |
-    set -o pipefail
```
yuumasato: Can you add a comment why we cannot have `set -o pipefail`?
If possible, can you add comment to suppress warnings from Ansible lint?
Motivation for pipefail is explained in #4230

mildas: @yuumasato Added comment about `set -o pipefail`, but what do you mean with suppress warning from Ansible lint? Ansible doesn't show any warning to me (tried also without `warn: False` line)

yuumasato: @mildas `ansible-lint` command, provided by `python3-ansible-lint`

With your changes I get:
```bash
$ ansible-lint rhel8/playbooks/pci-dss/rpm_verify_hashes.yml 
[306] Shells that use pipes should set the pipefail option
rhel8/playbooks/pci-dss/rpm_verify_hashes.yml:36
Task/Handler: Read files with incorrect hash
```

To suppress this:
https://docs.ansible.com/ansible-lint/rules/rules.html#false-positives-skipping-rules

mildas: @yuumasato Well, I tried everything and I could suppress this error with `# noqa 306` comment on `name:` line or after `args:` line, but problem with it is that build deletes these comments - final playbook doesn't contain the suppress comment :-(

yuumasato: @mildas LOL, OK then.
Thanks for checking.
I'm merging, hopefully the comment will be enough to explain why there is not pipefail.

PR #4770, opened 2019-08-28
Fix sshd_required logic in OVAL macro

yuumasato:
#### Description:

- Add macro for `sshd_required_or_requirement_unset`
  - This is the missing counter part of `sshd_not_required_or_unset`

#### Rationale:

- Evaluation of SSH rules that use `sshd_required`are not evaluating correctly


ggbecker:
Looks like it was an omission when converting the OVAL checks to use jinja macros. Thanks for the fix. I'll run some tests to check if everything is ok.

yuumasato:
@ggbecker Thank you for review!

PR #4773, opened 2019-08-28
Enable fapolicyd to watch all system mountpoints.

adelton:
#### Description:

- Enable fapolicyd to watch all system mountpoints.

#### Rationale:

- To control file access for all possible rules, all system mountpoints need to be listed in `/etc/fapolicyd/fapolicyd.mounts`.


adelton:
Note that the kickstart suggests
```
mount | egrep '^tmpfs| ext4| ext3| xfs' | awk '{ printf "%s\n", $3 }' | egrep -v '^\/run|^\/sys' >> /etc/fapolicyd/fapolicyd.mounts
```
to filter out `/run` and `/sys` but that's not what the comment in `/etc/fapolicyd/fapolicyd.mounts` itself (both in RHEL 8 and upstream at https://github.com/linux-application-whitelisting/fapolicyd/blob/master/init/fapolicyd.mounts) suggests. Since consistency beats pretty much everything, I'm not filtering these prefixes in this pull request.

linux_os/guide/services/fapolicyd/configure_fapolicyd_mounts/rule.yml
```diff
@@ -0,0 +1,31 @@
...
+
+title: 'Enable fapolicyd to watch all system mountpoints'
+
+description: |-
+    Configure the <tt>fapolicyd</tt> (File Access Policy Daemon)
+    to watch on all system filesystem mountpoints
+    in <tt>/etc/fapolicyd/fapolicyd.mounts</tt>.
+
+rationale: |-
+    To control file access for all possible rules, all system
```
yuumasato: Do you mean  `all possible mountpoints`?

yuumasato:
> I'm not filtering these prefixes in this pull request.

I'll be checking on these.

linux_os/guide/services/fapolicyd/configure_fapolicyd_mounts/rule.yml
adelton: I think I had rules in mind -- to support any paths you might want to put to the `.rules` file that may land on any filesystem, we need to list all filesystems.

adelton:
I've added check for missing `fapolicyd.mounts` file -> 085349df39b62d43db03825b8c61075c8190187e.

@yuumasato, it seems to fix the `error` result when the file is missing. However, the check still `error`s when the file has no paths in it, for example when the file only has comments.

@redhatrises, is the check for file size needed? Merely checking for existence works to make the check `fail` when the file is missing, and even if the file is nonempty but only has comments (so no match for the `^\s*([^#\s]+)`), I still get `error` ...

adelton:
We could go with `^\s*([^#\s]+|)`, adding an empty string to the mix (which will always exist in the file) so that `obj_configure_fapolicyd_mounts` gets something and can be used in the test, but it feels a bit hacky.

yuumasato:
> Please add a `file_test` to the OVAL that checks that the file exists and that its size is greater than zero.

During development of another check, I tried a similar approach, to avoid errors and it didn't work. I didn't try checking for size, although it probably wouldn't matter.

I thought it was due to OVAL design, but looking at specification of [OperatorEnumerator](https://github.com/OVALProject/Language/blob/master/docs/oval-common-schema.md#---operatorenumeration---), it should work.
This could a bug in OpenSCAP behavior.

yuumasato:
> it seems to fix the `error` result when the file is missing.

Great, so it works.

>  the check still `error`s when the file has no paths in it, for example when the file only has comments.

Problem is accessing variable a variable that is "collected" as empty. 

> We could go with `^\s*([^#\s]+|)`, adding an empty string to the mix (which will always exist in the file) so that `obj_configure_fapolicyd_mounts` gets something and can be used in the test, but it feels a bit hacky.

If this prevents the variable from being empty, and doesn't break the check.
I don't think it will, it iis unlikely that an object with empty mount point will be returned.

adelton:
> I'm fine if this prevents the variable from being empty, and doesn't break the check.
> I don't think it will, it iis unlikely that an object with empty mount point will be returned.

Right. So isn't there a better way to say "empty variable (set) is okay, fail, don't error"?

jan-cerny:
I recall diving into that in past and I haven't found any.

yuumasato:
> Right. So isn't there a better way to say "empty variable (set) is okay, fail, don't error"?

I'm not aware of a way to do so. 

From `var_check` attribute described in [EntityAttributeGroup](https://github.com/OVALProject/Language/blob/master/docs/oval-definitions-schema.md#---entityattributegroup---):
`If a variable does not have any value value when referenced by an OVAL Object the object should be considered to not exist. If a variable does not have any value when referenced by an OVAL State an error should be reported during OVAL `

linux_os/guide/services/fapolicyd/configure_fapolicyd_mounts/rule.yml
yuumasato: Not sure I understand correctly.
So `fapolicyd.mounts` lists partitions to watch, and paths in `fapolicyd.rules` should fall within watched mounts?
Otherwise the rule won't have any effect?

adelton: That's my understanding of the rather sparse `fapolicyd(8)` man page.

adelton:
Added commit to handle empty `fapolicy.mounts` file well, rebased on master -> 7b91dd35c9f7eba4828c85957cceaf78c37e19dd.

adelton:
Added exclude of `/run$` and `/sys` -> 80ca1c04398816f2d6d56d3ef4e9753be4fd4660.

adelton:
Better yet -> 785850126ad981268a0d0e3fb38496d1c4dd6ddb.

yuumasato:
@adelton thank you for the check the doesn't error.

PR #4776, opened 2019-08-29
Fix remediation for rule configure_opensc_nss_db

ggbecker:
#### Description:

- Aims to move away from `pkcs11-switch` tool as it may not configure the way our OVAL check for this particular rule works. The OVAL check looks for content in file `/etc/pki/nssdb/pkcs11.txt` which is only modified when using `modutil -dbdir sql:/etc/pki/nssdb/ ...` (notice the prefix `sql:`). 

- Also the rule got removed from RHEL8 because in RHEL8 there is no coolkey and it seems to be already configured to use `opensc` library by default. Update1: source: https://access.redhat.com/articles/4253861 (See sections: `Smart cards in Firefox browser or Thunderbird` and `Register third party PKCS #11 module to p11-kit`)

#### Rationale:

- Remediation configures database correctly so OVAL check pass.

#### Additional info:

There are still some uncertainty on how this databases works, but so far they way is proposed it seems to work.

### DIFF
--- a/linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/configure_opensc_nss_db/bash/shared.sh
+++ b/linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/configure_opensc_nss_db/bash/shared.sh
@@ -4,8 +4,5 @@
 # complexity = low
 # disruption = low
 
-PKCSSW=$(/usr/bin/pkcs11-switch)
-
-if [ ${PKCSSW} != "opensc" ] ; then
-    ${PKCSSW} opensc
-fi
+modutil -delete "CoolKey PKCS #11 Module" -dbdir sql:/etc/pki/nssdb/ -force || true # ignore errors
+modutil -add "OpenSC PKCS #11 Module" -dbdir sql:/etc/pki/nssdb/ -libfile opensc-pkcs11.so -force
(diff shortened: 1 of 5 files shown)


linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/configure_opensc_nss_db/bash/shared.sh
```diff
@@ -1,11 +1,8 @@
...
 # complexity = low
 # disruption = low
 
-PKCSSW=$(/usr/bin/pkcs11-switch)
-
-if [ ${PKCSSW} != "opensc" ] ; then
-    ${PKCSSW} opensc
-fi
+modutil -delete "CoolKey PKCS #11 Module" -dbdir sql:/etc/pki/nssdb/ -force || true # ignore errors
+modutil -add "OpenSC PKCS #11 Module" -dbdir sql:/etc/pki/nssdb/ -libfile opensc-pkcs11.so -force
```
ggbecker: It might need some kind of condition to test if it is already setup correctly.

yuumasato:
Removing milestone assignment, as this is still in draft.

shawndwells:
@ggbecker closing this as abandoned PR. Feel free to re-open if you find the time in the future!

PR #4777, opened 2019-08-29
Revert PR #4766

redhatrises:
This reverts PR #4766 as it introduced an Ansible regression and is not the correct fix.

### DIFF
--- a/linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_hashes/ansible/shared.yml
+++ b/linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_hashes/ansible/shared.yml
@@ -15,7 +15,7 @@
 
 - name: "Read files with incorrect hash"
   shell: |
-    # Do not set 'set -o pipefail' - 'rpm -Va' returns non-zero value because non-empty list of files
+    set -o pipefail
     rpm -Va | grep -E '^..5.* /(bin|sbin|lib|lib64|usr)/' | awk '{print $NF}'
   args:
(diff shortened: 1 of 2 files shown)


jan-cerny:
Hi @redhatrises, could you explain the regression it introduced? Also, do you have any suggestion what would be a correct fix?

yuumasato:
@redhatrises Reverting is not an option, as the remediation was causing the playbook run to abort.
Are you mostly concerned about warning of ansible-lint? Is it for Ansible Galaxy?

redhatrises:
> @redhatrises Reverting is not an option, as the remediation was causing the playbook run to abort.
> Are you mostly concerned about warning of ansible-lint? Is it for Ansible Galaxy?

It is 100% the option when the solution is not correct.

redhatrises:
> Hi @redhatrises, could you explain the regression it introduced? Also, do you have any suggestion what would be a correct fix?

There are a couple of options that exist. The `failed_when` can be set correctly based off of the rc code or get rid of shell and use command using Ansible filters instead of grep and awk.

shawndwells:
@redhatrises agree this should be reverted, but what's the plan on a correct fix? Looks like merging this PR will block work on implementing an alternative implementation, but simultaneously we don't want content in a broken state for to long (even if upstream). 

Will you be able to submit a PR with an alternative implementation later today?

redhatrises:
> @redhatrises agree this should be reverted, but what's the plan on a correct fix? Looks like merging this PR will block work on implementing an alternative implementation, but simultaneously we don't want content in a broken state for to long (even if upstream).
> 
> Will you be able to submit a PR with an alternative implementation later today?

PR will follow really quick once merged.

PR #4779, opened 2019-08-30
Remove rule configure_opensc_nss_db from RHEL8 product.

ggbecker:
#### Description:

- Remove rule configure_opensc_nss_db from RHEL8 product. This PR is part of #4776, but I decided to make the RHEL8 rule removal separated as it can be part of release 0.1.46. This rule is failing on RHEL8 PCI-DSS. See sections `Smart cards in Firefox browser or Thunderbird` and `Register third party PKCS #11 module to p11-kit` on RHEL8 article: https://access.redhat.com/articles/4253861 which confirms that OpenSC is set by default on RHEL8

#### Rationale:

- The rule is not applicable to RHEL8. OpenSC is set by default on RHEL8. Additionally RHEL8 does not contain pkcs11-switch tool.

### DIFF
--- a/linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/configure_opensc_nss_db/ansible/shared.yml
+++ b/linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/configure_opensc_nss_db/ansible/shared.yml
@@ -1,4 +1,4 @@
-# platform = Red Hat Enterprise Linux 7,Red Hat Enterprise Linux 8,multi_platform_fedora,multi_platform_rhv,Oracle Linux 7
+# platform = Red Hat Enterprise Linux 7,multi_platform_fedora,multi_platform_rhv,Oracle Linux 7
 # reboot = false
 # strategy = configure
 # complexity = low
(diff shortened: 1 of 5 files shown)


yuumasato:
LGTM, thanks.

ggbecker:
> #### Description:
> 
>     * Remove rule configure_opensc_nss_db from RHEL8 product. This PR is part of #4776, but I decided to make the RHEL8 rule removal separated as it can be part of release 0.1.46. This rule is failing on RHEL8 PCI-DSS. See sections `Smart cards in Firefox browser or Thunderbird` and `Register third party PKCS #11 module to p11-kit` on RHEL8 article: https://access.redhat.com/articles/4253861 which confirms that OpenSC is set by default on RHEL8

I discovered a few things: OpenSC is set by default but it needs to be installed (`opensc` package). When you install `opensc`, it creates the file `/usr/share/p11-kit/modules/opensc.module` which contains the entry `module: opensc-pkcs11.so` and `p11-kit` knows how to load the library for the applications which will be using the library, e.g. nss (which is consumed by firefox/thunderbird)

We could change the approach of this rule to something like:

- check if the `/usr/share/p11-kit/modules/opensc.module` which points to correct library, and the remediation would be installing `opensc` package.

@yuumasato any thoughts on that?

PR #4780, opened 2019-08-30
Install 1.2 datastreams

yuumasato:
#### Description:

- Install the SCAP 1.2 DataStream
- Copy the SCAP 1.2 DataStream to pre-generate content zipfile
- Fix up some dependencies

#### Rationale:

- 1.2 DataStream is not installed by `cmake install` command


jan-cerny:
I have verified it installs both datastream versions and also both datastream versions are present in the zip file. Thank you.

PR #4782, opened 2019-08-30
Unselect rule directory_access_var_log_audit in OSPP Profile

yuumasato:
#### Description:

- The rule is already covered by `audit_rules_for_ospp`

#### Rationale:
- The audit rule is part of `30-ospp-v42.rules` which is configured by `audit_rules_for_ospp`


jan-cerny:
Indeed, it's in https://github.com/linux-audit/audit-userspace/blob/8b9028cb51e58dca0db00ba108598a6a7f55bdee/rules/30-ospp-v42.rules#L121 . Nice catch!

PR #4785, opened 2019-08-30
Fail rpm_verify ansible fixes only if rc returns error code greater than 1

redhatrises:
- We can expect a 0 or 1 rc code returned in the rpm_verify ansible fixes.
  We should only fail if the rc code is not 0 or not 1.


shawndwells:
Ack pending jenkins.

PR #4788, opened 2019-09-02
Add test scenarios for rsyslog_remote_tls

jan-cerny:
This rule uses a complex regular expression so it might be beneficial to cover the rule by few test scenarios.


jan-cerny:
I have rebased it on master and also moved it to the appropriate directory.

matejak:
Do we need that `profile = ` line in the metadata? The test suite will pick the virtual `(all)` profile, which is OK in this case, as the rule is not parametrized.

jan-cerny:
@matejak That's needed to use the test scenarios within SSGTS to test the specific rule.

jan-cerny:
@matejak I have removed profile IDs from the scenarios headers.

matejak:
Those tests are valid and work as expected. Thank you for the PR!

PR #4789, opened 2019-09-03
ssg_test_suite: Fix SSH port handling for podman backend in rootless mode

evgenyz:
This patch will allow to use podman backend of the SSG Test Suite without elevated privileges.

### DIFF
--- a/tests/ssg_test_suite/test_env.py
+++ b/tests/ssg_test_suite/test_env.py
@@ -224,11 +224,15 @@ class ContainerTestEnv(TestEnv):
         # Get the container time to fully start its service
         time.sleep(0.2)
 
-        self.domain_ip = self._get_container_ip(new_container) or 'localhost'
-        ports = self._get_container_ports(new_container)
-        if self.internal_ssh_port in ports:
-            self.ssh_port = ports[self.internal_ssh_port]
-            common.SSH_ADDITIONAL_OPTS = ('-o', 'Port={}'.format(self.ssh_port), ) + common.SSH_ADDITIONAL_OPTS
+        self.domain_ip = self._get_container_ip(new_container)
+        if not self.domain_ip:
+            self.domain_ip = 'localhost'
+            ports = self._get_container_ports(new_container)
+            if self.internal_ssh_port in ports:
+                self.ssh_port = ports[self.internal_ssh_port]
+        else:
+            self.ssh_port = self.internal_ssh_port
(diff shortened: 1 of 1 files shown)


matejak:
This unfortunately duplicates #4544 to some extent, but it is just a couple of lines of code, so no big harm.
However, we have functions that use `scp`, which will need to be adopted as well for the rootless podman to work.

jan-cerny:
@openscap-ci add to whitelist

evgenyz:
Aha, OK. It worked for me like that `./test_suite.py profile --container ssg_test_suite --datastream ../build/ssg-fedora-ds-1.3.xml standard`. Anyway, I'll check previous approach and `scp`-related code.
Thanks for the review.

evgenyz:
So, this is kind of hackiesh fix. I admit it. It is pretty simple, though.

Personally I would like to create some abstraction over ssh/scp and actively move it to DRY state, thus being able to affect SSH options without tapping into global variables. The question is — is it worth it?

evgenyz:
My test command was `./test_suite.py rule --container ssg_test_suite --datastream ../build/ssg-fedora-ds-1.3.xml accounts_tmout`

matejak:
I have tried the same command as you on Fedora.

It didn't work for me as root (which can be regarded as regression), I got the `ssh: connect to host 10.88.0.92 port 44875: Connection refused` error message. I got somewhat further as a normal user, but also no success. In the output, I got

```
ERROR - Cannot extract data tarball ./ssgts/./data.tar.
ERROR - Terminating due to error: Unable to upload test scripts: Cannot extract data tarball ./ssgts/./data.tar..
```

and the `data.upload.log` contain lots of erros s.a. `tar: ./utils.sh: Cannot change ownership to uid 102264, gid 102264: Invalid argument`. Could you please double-check?

Regarding optimal solution, I think that moving the SSH and SCP functions to `TestEnv` methods could do the trick, but if your approach works, I would call it good for the time being.

jan-cerny:
@evgenyz You have to change `ssg-fedora-ds-1.3.xml` to `ssg-fedora-ds.xml` and then it works fine.

matejak:
It looks like that the last error is more my `podman` setup. So the last think that remain is the case when `podman` is executed by root.

evgenyz:
Fix for root is here.

I've checked it with podman (rootless) and regular podman as well as docker backends. And I did it from a fresh start (cleaned and rebuilt all containers and images). Everything works for me.

matejak:
I can confirm that the root mode works for me. The approach cuts some corners, but it saved a lot of effort, good job!
@jan-cerny has confirmed that the rootless mode works for him, which is great. Thank you for the PR!

PR #4790, opened 2019-09-03
Add a tool to test generated datastreams

jan-cerny:
#### Description:
This script can compare 2 datastreams for presence of rules, OVALs and remediations, it can also show the diff between the remediation contents and also partially compares OVAL (only the criteria elements).

#### Rationale:
This script will help us to ensure no regression is made during the templating system redesign, but it might be useful also elsewhere.

### DIFF
--- a/utils/compare_ds.py
+++ b/utils/compare_ds.py
@@ -1,0 +1,274 @@
+#!/usr/bin/python3
+
+import argparse
+import sys
+import xml.etree.ElementTree as ET
+import difflib
+
+import ssg.constants
+
+ns = {
+    "ds": "http://scap.nist.gov/schema/scap/source/1.2",
+    "xccdf": "http://checklists.nist.gov/xccdf/1.2",
+    "oval": "http://oval.mitre.org/XMLSchema/oval-definitions-5",
+    "catalog": "urn:oasis:names:tc:entity:xmlns:xml:catalog",
+    "xlink": "http://www.w3.org/1999/xlink",
+}
+remediation_type_to_uri = {
+    "bash": "urn:xccdf:fix:script:sh",
+    "ansible": "urn:xccdf:fix:script:ansible",
+    "puppet": "urn:xccdf:fix:script:puppet",
+    "anaconda": "urn:redhat:anaconda:pre",
+}
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description="Compares two datastreams with regards to presence of"
+        "OVAL checks and all remediations")
+    parser.add_argument(
+        "--old", required=True, metavar="PATH",
+        help="Path to the old datastream")
+    parser.add_argument(
+        "--new", required=True, metavar="PATH",
+        help="Path to the new datastream")
+    parser.add_argument(
+        "--rule", metavar="RULE_ID",
+        help="Compare only the rule specified by given RULE_ID"
+    )
+    parser.add_argument(
+        "--no-diffs", action="store_true",
+        help="Do not perform detailed comparison of checks and "
+        "remediations contents."
+    )
+    return parser.parse_args()
+
+
+def get_benchmarks(root):
+    for component in root.findall("ds:component", ns):
+        for benchmark in component.findall("xccdf:Benchmark", ns):
+            yield benchmark
+
+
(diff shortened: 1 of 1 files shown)


utils/compare_ds.py
```diff
@@ -0,0 +1,274 @@
+#!/usr/bin/python3
+
+import argparse
+import sys
+import xml.etree.ElementTree as ET
+import difflib
+
+import ssg.constants
+
+ns = {
```
matejak: Couldn't we put this into constants?

utils/compare_ds.py
```diff
@@ -0,0 +1,274 @@
...
+    "ds": "http://scap.nist.gov/schema/scap/source/1.2",
+    "xccdf": "http://checklists.nist.gov/xccdf/1.2",
+    "oval": "http://oval.mitre.org/XMLSchema/oval-definitions-5",
+    "catalog": "urn:oasis:names:tc:entity:xmlns:xml:catalog",
+    "xlink": "http://www.w3.org/1999/xlink",
+}
+remediation_type_to_uri = {
```
matejak: ditto

utils/compare_ds.py
```diff
@@ -0,0 +1,274 @@
...
+    "puppet": "urn:xccdf:fix:script:puppet",
+    "anaconda": "urn:redhat:anaconda:pre",
+}
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description="Compares two datastreams with regards to presence of"
+        "OVAL checks and all remediations")
+    parser.add_argument(
```
matejak: Make the old and new DS filenames required positional arguments rather than required optional arguments. The current approach is somehow [non-standard](https://docs.python.org/3/library/argparse.html#required).

utils/compare_ds.py
```diff
@@ -0,0 +1,274 @@
...
+        old_benchmark, new_benchmark, old_oval_defs, new_oval_defs,
+        rule_id, show_diffs):
+    missing_rules = []
+    if rule_id:
+        if not rule_id.startswith(ssg.constants.OSCAP_RULE):
+            rule_id = ssg.constants.OSCAP_RULE + rule_id
+        rules_in_old_benchmark = old_benchmark.findall(
+            ".//xccdf:Rule[@id='%s']" % (rule_id), ns)
+    else:
+        rules_in_old_benchmark = old_benchmark.findall(".//xccdf:Rule", ns)
```
matejak: Ideally, split this into "get_rules_to_compare" and "compare_rules" functions - the original one is quite big and function names act as comments pointing towards what's going on.

utils/compare_ds.py
```diff
@@ -0,0 +1,274 @@
...
+        "xccdf:check[@system='%s']" % (ssg.constants.oval_namespace), ns)
+    new_oval_ref = new_rule.find(
+        "xccdf:check[@system='%s']" % (ssg.constants.oval_namespace), ns)
+    rule_id = old_rule.get("id")
+    if (old_oval_ref is None and new_oval_ref is not None):
+        print("New datastream adds OVAL for rule '%s'." % (rule_id))
+    elif (old_oval_ref is not None and new_oval_ref is None):
+        print("New datastream is missing OVAL for rule '%s'." % (rule_id))
+    elif (old_oval_ref is not None and new_oval_ref is not None):
+        old_check_content_ref = old_oval_ref.find(
```
matejak: Please extract this into a separate function with a good name, we have some serious `if` nesting here.

jan-cerny:
@matejak I have used constants from ssg.constants module whenever possible, refactored the code in order to extract functions `get_rules_to_compare` and `compare_rules`, then I have refactored the code to extract function `compare_fix_elements`. I have also renamed function `compare_remediation_contents` to `compare_fix_texts`. I have made `--old` and `--new` optional arguments.

matejak:
Good job! There is some confusion in the reply - the `old` and `new` arguments were converted to positional and they are required, as it was suggested. The new code is also easier to understand, which is great.

I have thought that you will put the whole namespace dict into the constant module as it is very userful in general, but this is also fine, and it can be moved later when conditions are even better.

Thank you for the PR!

PR #4792, opened 2019-09-03
SSG Test Suite: Fix (all) profile execution when running test suite in rule mode

ggbecker:
#### Description:

- Fix (all) profile execution when running test suite in rule mode.

### DIFF
--- a/tests/ssg_test_suite/oscap.py
+++ b/tests/ssg_test_suite/oscap.py
@@ -113,6 +113,13 @@ def get_result_id_from_arf(arf_path, verbose_path):
     return res_id
 
 
+def single_quote_string(input):
+    result = input
+    for char in "\"'":
+        result = result.replace(char, "")
+    return "'{}'".format(result)
+
+
 def generate_fixes_remotely(formatting, verbose_path):
     command_base = ['oscap', 'xccdf', 'generate', 'fix']
     command_options = [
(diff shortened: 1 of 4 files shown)


tests/ssg_test_suite/rule.py
```diff
@@ -18,7 +18,11 @@
 import data
 
 
-ALL_PROFILE_ID = "(all)"
+# Single quotes are needed because this value is used in
+# a bash environment when running oscap-ssh command line utility.
+# This has been fixed on OpenSCAP side but the fix has not been released yet:
+# https://github.com/OpenSCAP/openscap/pull/1366
```
matejak: ditto to making this revert-friendly.

tests/ssg_test_suite/oscap.py
```diff
@@ -117,7 +117,10 @@ def generate_fixes_remotely(formatting, verbose_path):
     ]
     command_operands = ['/{arf_file}'.format(** formatting)]
     if 'result_id' in formatting:
-        command_options.extend(['--result-id', formatting['result_id']])
+        # sanitize id because when using "(all)" profile the parenthesis
```
matejak: I propose to mark it somehow, so when a fix in `oscap-ssh` is released, these workarounds can be reverted quickly.

tests/ssg_test_suite/rule.py
```diff
@@ -226,7 +230,10 @@ def _test_target(self, target):
 
     def _modify_parameters(self, script, params):
         if self.scenarios_profile:
-            params['profiles'] = [self.scenarios_profile]
```
matejak: Let's tell users to specify the actual virtual profile name, i.e. `(all)`, like that, there won't be any need to treat the string `all` in a special way.

tests/ssg_test_suite/rule.py
```diff
@@ -343,7 +350,8 @@ def perform_rule_check(options):
     checker.scenarios_profile = options.scenarios_profile
     # check if target is a complete profile ID, if not prepend profile prefix
     if (checker.scenarios_profile is not None and
-            not checker.scenarios_profile.startswith(OSCAP_PROFILE)):
+            not checker.scenarios_profile.startswith(OSCAP_PROFILE) and
+            not checker.scenarios_profile.lower() == "all"):
```
matejak: Generally, we don't want to prefix anything to virtual profiles. A simple check could be based on brackets being present only in virtual profiles.

tests/ssg_test_suite/rule.py
ggbecker: done

tests/ssg_test_suite/rule.py
ggbecker: done

tests/ssg_test_suite/rule.py
ggbecker: done

tests/ssg_test_suite/oscap.py
ggbecker: still needed. keeping the change in a nicer way.

matejak:
Great, I can confirm that it works also on the system with quoting fixed in `oscap-ssh`.

PR #4794, opened 2019-09-04
Rule harden_sshd_crypto_policy added to the unselect_rules_list

matusmarhefka:
#### Description:

- Rule `harden_sshd_crypto_policy` is breaking access to machines using standard `rsa-sha2-256` and `rsa-sha2-512` keys, therefore add it to the `unselect_rules_list` which is used to unselect rules when testing using SSG TS in the profile mode.


mildas:
Looks good to me. Thank you

PR #4797, opened 2019-09-05
Changed owner of files in the test suite tarball.

matejak:
Made the owner of scenarios and shared functaionality ther root user, as tests are executed by root anyway.
High UIDs and GIDs could cause problems with the rootless podman backend.


jan-cerny:
Thanks for this patch.

PR #4798, opened 2019-09-05
ssg_test_suite: Update documentation about Test Suite usage with the information about rootless podman backend

evgenyz:


tests/README.md
```diff
@@ -88,35 +88,24 @@ To use Podman backend, you need to have:
...
-Go into the `Dockerfiles` directory of the project, and execute the following:
+Now when we all set with SSH keys let's build the container. Go into the `Dockerfiles` directory of the project, and execute the following:
 
 ```
-public_key="$(sudo cat /root/.ssh/id_rsa.pub)"
+public_key="$(cat ~/.ssh/id_rsa.pub)"
 podman build --build-arg CLIENT_PUBLIC_KEY="$public_key" -t ssg_test_suite -f test_suite-rhel .
 ```
 
+*NOTE*: If you are setting up the suite as the superuser use *public_key="$(sudo cat /root/.ssh/id_rsa.pub)"* instead of the first command.
```
matejak: Please add a mention that this typically applies when one wants to use `docker` as opposed to `podman`. What I recall, it could be a good idea to also mention that the build command should be also executed under root. This doesn't make a difference with `docker`, but with `podman`, the user who builds the container is the one who gets the image.

evgenyz: Mmm, I think this should clarify superuser usage. I'm trying to not to mix podman and docker in the guide, while stressing out the fact that Docker backend will require *sudo* in the appropriate section.

matejak:
Looks good, thank you for the PR, merging!

PR #4800, opened 2019-09-05
Improved the build_product script.

matejak:
* Added GNU command-line interface to parametrize the script.
* Leveraged the Argbash command-line code generator.
* Added the `--derivatives` option to be able to opt-in to build derivatives.
* Fixed shellcheck errors.

Derivatives are not built by default, which means that there is only one datastream built per product. As a result, when running tests after a build, there is no need to supply the `--datastream` option.


build_product
```diff
@@ -76,47 +270,49 @@ all_cmake_products=(
 	WRLINUX1019
 )
 
-if (( $# == 0 )); then
-	handle_wrong_argument
-fi
+NO_DERIVATIVES_OPTIONS=('-DSSG_SCIENTIFIC_LINUX_DERIVATIVES_ENABLED:BOOL=OFF' '-DSSG_CENTOS_DERIVATIVES_ENABLED:BOOL=OFF')
```
jan-cerny: If I run `./build_product rhel8` I get this warning from CMake:

```
CMake Warning:
  Manually-specified variables were not used by the project:

    SSG_SCIENTIFIC_LINUX_DERIVATIVES_ENABLED
```

I think you need to handle RHEL8 in a special way because the Scientific Linux 8 will not exist.

build_product
```diff
@@ -1,11 +1,201 @@
...
+_arg_debug="off"
+_arg_derivatives="off"
+
+
+print_help()
+{
+	printf '%s\n' "Wipes out contents of the 'build' directory and builds only and only the given products."
+	printf 'Usage: %s [-o|--oval <VERSION>] [-b|--builder <BUILDER>] [-j|--jobs <arg>] [--(no-)debug] [--(no-)derivatives] [-h|--help] <product-1> [<product-2>] ... [<product-n>] ...\n' "$0"
+	printf '\t%s\n' "<product>: Products to build"
+	printf '\t%s\n' "-o, --oval: OVAL minor version. Can be one of: '10', '11' and 'auto' (default: 'auto')"
```
jan-cerny: Could we put the whole version there? Or can we add a note that it's OVAL 5.x?

matejak:
Good suggestions, I have

- made the --oval option accept the full OVAL version string
- started to treat RHEL<=7 and RHEL>7 differently wrt disabling derivatives.

PR #4801, opened 2019-09-06
Small fixes for compare_ds.py

jan-cerny:
1. Add a `break`
If a comparison was a successful match we shouldn't continue with comparing, otherwise we would match all the items with same value.

2. Concatenate all text within a `fix` element
If there is no string between opening tag of `fix` element and the `sub` element, the `Element.text` property is `None`, which caused a traceback in `compare_fix_texts` function. To get all the text without skipping the child elements we can use `Element.itertext()` method.


matejak:
LGTM, thank you for the fix and its description!

PR #4804, opened 2019-09-06
Migrate templating input data from CSVs into the rules

yuumasato:
#### Description:

- Adds a script to migrate data store in templating CSV files to their respective rule.yml files.
  - For now, the script is limited to packages_installed template.
- [x] Load CSV data from all products
- [x] Normalize CSV Data (consolidate across all produts, make them product specific if necessary)
- ~~Walk through the product's Benchmark and put the Templating Data into the rules.~~
- [x] Move CSV data to rule.ymls

#### Rationale:

- The changes in templating should make it easier to identify that a rule is templated
- Maintenance of CSVs should be simpler when a new product is created
- @jan-cerny is working on the template infrastructure itself


utils/migrate_template_csv_to_rule.py
```diff
@@ -0,0 +1,638 @@
...
+    name = escape_path(os.path.basename(path))
+    rule_id = f"audit_rules_execution_{name}"
+
+    # create_audit_rules_execution.py escapes the '/' when generating the OVAL
+    # This is not actually necessary
+    audit_rules_execution["PATH"] = path
+    audit_rules_execution["NAME"] = name
+
+    audit_rules_execution["ID"] = f"audit_rules_execution_{name}"
+    audit_rules_execution["TITLE"] = f"Record Any Attempts to Run {name}"
```
jan-cerny: @yuumasato  Is it expected that the people will always fill this in when they write `rule.yml`? I think that isn't user friendly. Should #4809 contain similar logic to fill it the id and title during the build?

utils/migrate_template_csv_to_rule.py
```diff
@@ -0,0 +1,638 @@
...
+        package_name = service_name
+    daemon_name = csv_line[2]
+    if not daemon_name:
+        daemon_name = service_name
+
+    rule_id = f"service_{service_name}_enabled"
+
+    service_enabled["SERVICENAME"] = service_name
+    service_enabled["PACKAGENAME"] = package_name
+    service_enabled["DAEMONNAME"] = daemon_name
```
jan-cerny: Similar to previous comment, do we want content creators to provide all 3 items when they write a new `rule.yml` or do we allow them to provide only service name, meaning the package and daemon will be inferred by the templating engine in that situation?

utils/migrate_template_csv_to_rule.py
yuumasato: I agree, I have this already. I didn't push the next commits yet.

jan-cerny: @yuumasato So it's expected that the people will always need to fill the ID and title or it isn't expected? If the ID and title should be derived by the build system, I need to add that into #4809 . Actually I planned to do that, but then I saw this, so I'm asking about if it's needed. I would like to prevent duplicate work.

yuumasato: > So it's expected that the people will always need to fill the ID and title or it isn't expected?

It is not expected that people will write the ID.

jan-cerny: ok thanks

shared/templates/create_accounts_password.py
```diff
@@ -8,41 +8,41 @@
...
-            SIGN = "-?" if VARIABLE.endswith("credit") else ""
+            SIGN = "-?" if variable.endswith("credit") else ""
             self.file_from_template(
                 "./template_OVAL_accounts_password",
                 {
                     "SIGN": SIGN,
-                    "VARIABLE": VARIABLE,
-                    "OPERATION": OPERATION
+                    "variable": variable,
+                    "operation": operation
```
yuumasato: `SIGN` should probably be made lowercase as well.

shared/templates/create_accounts_password.py
```diff
@@ -8,41 +8,41 @@
 
 class AccountsPasswordGenerator(FilesGenerator):
     def generate(self, target, pam_info):
-        VARIABLE, OPERATION = pam_info
+        variable, operation = pam_info
```
yuumasato: For the build to work, the `create_*.py` scripts need to be update as well...

linux_os/guide/system/accounts/accounts-pam/password_quality/password_quality_pwquality/accounts_password_pam_dcredit/rule.yml
```diff
@@ -54,3 +54,9 @@ ocil: |-
     The DoD requires at least one digit in a password. This would appear as <tt>dcredit = -1</tt>.
 
 platform: pam
+
+template:
+    name: package_installed
+    vars:
+        operation: less than or equal
+        variable: dcredit
```
yuumasato: These are the only ones made lowercase for now...

linux_os/guide/services/base/service_portreserve_disabled/rule.yml
```diff
@@ -33,3 +33,10 @@ references:
 ocil: '{{{ ocil_service_disabled(service="portreserve") }}}'
 
 platform: machine
+
+template:
+    name: service_disabled
+    vars:
+        mask_service: 'true'
+        packagename@rhel6: portreserve
+        servicename: portreserve
```
jan-cerny: The `packagename@rhel6` key is superflous because implicit behavior will be that the package name is the same as service name.

linux_os/guide/services/fapolicyd/service_fapolicyd_enabled/rule.yml
```diff
@@ -23,3 +23,8 @@ references:
 ocil_clause: 'the service is not enabled'
 
 ocil: '{{{ ocil_service_enabled(service="fapolicyd") }}}'
+
+template:
+    name: service_enabled
+    vars:
+        servicename: fapolicyd
```
jan-cerny: Ansible remediation for service fapolicyd should not be generated. See `rhel8/templates/csv/services_enabled.csv`. That means you should add:
```
backends:
    ansible: off
```

linux_os/guide/system/permissions/partitions/mount_option_dev_shm_nodev/rule.yml
```diff
@@ -30,3 +30,26 @@ references:
...
+        mount_has_to_exist@wrlinux1019: 'yes'
+        mount_has_to_exist@wrlinux8: 'yes'
+        mountoption: nodev
+        mountpoint: /dev/shm
+        type: tmpfs
+        type@rhel6: ''
+        type@sle12: ''
+        type@wrlinux1019: ''
+        type@wrlinux8: ''
+    template_backends:
```
jan-cerny: We currently use "backends", not "template_backends".

linux_os/guide/services/fapolicyd/service_fapolicyd_enabled/rule.yml
yuumasato: This should have worked, I'll look into what went wrong.

linux_os/guide/services/base/service_portreserve_disabled/rule.yml
yuumasato: Looking into this.

linux_os/guide/system/permissions/partitions/mount_option_dev_shm_nodev/rule.yml
yuumasato: Fixed to `backends`.

linux_os/guide/services/base/service_portreserve_disabled/rule.yml
yuumasato: Fixed.

linux_os/guide/services/fapolicyd/service_fapolicyd_enabled/rule.yml
yuumasato: Fixed, diff data can be seen in https://github.com/ComplianceAsCode/content/pull/4804/commits/90864b0279b38905391e29aebbc6fc0fbe29b96f

yuumasato:
@jan-cerny I have added the fixed templated data in new commits, this way it is easier to review them.

Do we wan't to have a single commit for the CSV data migration?
I can regenerate the commit rebase easily if so.

jan-cerny:
@yuumasato I think you don't have to squash the commits.

linux_os/guide/system/software/updating/timer_dnf-automatic_enabled/rule.yml
```diff
@@ -23,3 +23,9 @@ references:
 ocil_clause: 'the dnf-automatic.timer is not enabled'
 
 ocil: "{{{ ocil_timer_enabled(timer='dnf-automatic') }}}"
+
+template:
+    name: timer_enabled
+    vars:
+        package_name: dnf-automatic
```
jan-cerny: @yuumasato It's PACKAGENAME in the templates for timer_enabled, so the key should ideally be without the `_`.

yuumasato: Fixed, diff data can be seen in 0ae0e65cbaa61fcf23eac67283a8fbd96e24a6ab

jan-cerny:
Thank you very much. Kudos!

yuumasato:
@jan-cerny Thank you very much for the review!

PR #4805, opened 2019-09-06
Rename disable_prelink -> bash_disable_prelink

cipherboy:
Per conversation in #4746, we should probably prefix bash remediation
helpers with the `bash_` prefix. This lets us quickly identify which
language a particular macro is for, especially when macros with similar
functionality behave differently across languages.


redhatrises:
LGTM

PR #4806, opened 2019-09-06
Remove lineinfile deduplication task in template

redhatrises:
#### Description:

Removes the Ansible deduplication task in the macros

#### Rationale:

- lineinfile handles only an existence of one line, and deduplication has unintended consequences for a config option existing multiple times.

### DIFF
--- a/shared/macros-ansible.jinja
+++ b/shared/macros-ansible.jinja
@@ -84,11 +84,8 @@
 {{%- macro ansible_only_lineinfile(msg, path, line_regex, new_line, create='no', block=False, validate='', insert_after='', insert_before='') -%}}
 {{%- if block %}}
 - name: "{{{ msg or rule_title }}}"
-  block:
-    {{{ ansible_lineinfile("Deduplicate values from " + path, path, regex=line_regex, new_line=None, create='no', state='absent')|indent }}}
-    {{{ ansible_lineinfile("Insert correct line to " + path, path, regex=None, new_line=new_line, create=create, state='present', validate=validate, insert_after=insert_after, insert_before=insert_before)|indent }}}
+  {{{ ansible_lineinfile("Insert correct line to " + path, path, regex=None, new_line=new_line, create=create, state='present', validate=validate, insert_after=insert_after, insert_before=insert_before)|indent }}}
 {{%- else %}}
-{{{ ansible_lineinfile("Deduplicate values from " + path, path, regex=line_regex, new_line=None, create='no', state='absent') }}}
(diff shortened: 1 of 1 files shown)


cipherboy:
That was actually kinda intended: if someone specifies e.g., `PasswordAuthentication` multiple times with different values, `sshd_config` will only take the last one (sans any `Host` config blocks in really complex configurations).

In this context, specifying the same parameter multiple times is an anti-pattern.


What context are you trying to use `ansible_lineinfile` that you don't want the deduplication? Perhaps we should create a different macro (which I thought we already had)?

`ansible_lineinfile` was also a low-level macro, and you should be writing a higher-level abstraction over the top of it.

redhatrises:
> That was actually kinda intended: if someone specifies e.g., `PasswordAuthentication` multiple times with different values, `sshd_config` will only take the last one (sans any `Host` config blocks in really complex configurations).
> 
> In this context, specifying the same parameter multiple times is an anti-pattern.
> 
> What context are you trying to use `ansible_lineinfile` that you don't want the deduplication? Perhaps we should create a different macro (which I thought we already had)?
> 
> `ansible_lineinfile` was also a low-level macro, and you should be writing a higher-level abstraction over the top of it which fits your particular usage.

Basically, you are using the wrong module for multi-line instantiations and single option configured multiple times.

cipherboy:
@redhatrises So the way this works is that the `ansible_lineinfile` macro is a thin wrapper over the Ansible [`lineinfile`](https://docs.ansible.com/ansible/latest/modules/lineinfile_module.html) module. Which yes, only handles one-line configuration parameters really well.

In this PR, you're updating `ansible_only_lineinfile` (which is an extension of the `ansible_lineinfile` macro) to not deduplicate lines. However, that's the explicit purpose of that macro. That macro *explicitly* checks that if `line_regex` is matched, any matching lines get removed, and at the end, the correct line (`new_line`) is always explicitly added. If you don't want to deduplicate, and perhaps only conditionally add, then you should create a separate macro.

The same holds for `ansible_set_config_file_dir`, except that it is a higher-level wrapper. It assumes you want to deduplicate the configuration file+dir structure (e.g., `/etc/yum.repos` and `/etc/yum.repos.d/`). If you don't, you should create a separate macro.  

If you don't want something to be deduplicated, you can use `ansible_lineinfile` directly (without the removal options) and explicitly only add (or, conditionally add).

In what usages are you seeing that `ansible_only_lineinfile` or `ansible_set_config_file_dir` is used incorrectly?

redhatrises:
Closing after talking to Alex offline.

cipherboy:
To clarify: I think we should focus on adding `replace` and `ini` module support and updating https://github.com/ComplianceAsCode/content/pull/4628 appropriately.

PR #4807, opened 2019-09-06
ansible_set_sshd: add before first commented Match

cipherboy:
Previously, we added new configuration values before the first instance
of an uncommented Match block. However, if an admin had a commented-out
Match block and later uncommented it, the configuration file would
likely become invalid.

We thus move the addition to before the first Match, regardless of
whether or not it is commented out.

I'd appreciate additional review on this. I *think* the resulting regex is correct...

Per discussion with @redhatrises.


redhatrises:
Tested it, and LGTM

PR #4809, opened 2019-09-09
[WIP] New templating system

jan-cerny:
#### Description:
The new templating system will use new keys from `rule.yml`: `template`,`template_data`, `template_backends`.

Script `build-scripts/build_templated_content.py` will generate the templated content for each product separately.  It will iterate over already resolved rule YAMLs (in `build/<product>/rhel8`) and for each rule it will generate all contents (OVAL, Bash, Ansible, Puppet, Anaconda) if the rule is templated.

The templates will remain intact, but the code from `create_` Python scripts needs to be ported over.

As an example, we add a rule `package_gcc_installed` which is templated using the new system.

#### Rationale:
The current system which is based on CSV files has certain problems and limitations.

1. the CSV data is located far away from rule definitions which it's cumbersome to determine which rule is templated or not and with which data are the templates populated.
2. multiple products have their own CSV although they are not much different from each other
3. when a new product is created the CSVs need to be recreated


linux_os/guide/system/software/system-tools/package_gcc_installed/rule.yml
```diff
@@ -0,0 +1,16 @@
...
+description: The gcc package should be installed.
+
+rationale: GCC can compile programs to binary executables
+
+template: package_installed
+template_data:
+    PKGNAME: gcc
+    EVR: 0:9.2.1-1
+template_backends:
+    bash: off
```
redhatrises: All content around the template should be sub-keys of template. And really, I don't think that we need to have the `template_data` or `template_backends` keys at all. A better and cleaner implementation of the template key would be:
```
template: 
    name: package_installed
    pkgname: gcc
    evr: 0:9.2.1-1
    bash: off
```
or
```
template: 
    name: package_installed
    vars:
        pkgname: gcc
        evr: 0:9.2.1-1
    bash: off
````

All keys should be lowercase.

Comment (matejak) [reply linux_os/guide/system/software/system-tools/package_gcc_installed/rule.yml]: I like the idea to have just one key for all of the template-related stuff. However, sub-keys for name, vars and backends are good, mainly because a person who is not familiar with the content will immediately understand what's going on, that `bash` is not a variable, but it switches `bash` remediation off, and that `name` is special.
It has also pleasant side-effects that data are better organized, parsing code will not have to have to deduce what category a key belongs to, and there won't be key name conflicts.

Comment (jan-cerny) [reply linux_os/guide/system/software/system-tools/package_gcc_installed/rule.yml]: The current proposal is:
```
template:
    name: template_name
    vars:
        param_name@rhel7: value1
        param_name@rhel8: value2
        other_param: other_value
    backends:
        ansible:  off
        bash: on # default (implicit)
```

eg.

```
template: 
    name: services_enabled
    vars:
        package@rhel: audit
        package@ubuntu: auditd
        daemon_name: auditd
    backends:
        ansible: off
````

Comment (jan-cerny) [reply linux_os/guide/system/software/system-tools/package_gcc_installed/rule.yml]: Regarding the request that the keys should be lowercase. In this PR they're uppercase because I pass they directly to Jinja which processes the templates and in templates we now use uppercase Jinja variables everywhere. We have briefly discussed that with @matejak an d @yuumasato and I the solution will be to use lowercase in Jinja and batch convert all the templates to use lowercase variables. That change can be a part of commit that fills the data to rule YAMLs.

Comment (redhatrises) [reply linux_os/guide/system/software/system-tools/package_gcc_installed/rule.yml]: Not seeing a need for the `backends` key at all since `ansible: off` relates to the template key itself making `backends` superfluous.... unless we need to add `ansible@rhel7` to the key.

Comment (jan-cerny) [reply linux_os/guide/system/software/system-tools/package_gcc_installed/rule.yml]: The separate `backends` key makes the code in `ssg/templates.py` that uses it slightly easier and less error-prone, which was @matejak 's point I believe. On the other hand it forces the person that uses it in `rule.yml` to write one more key, but it will happen rarely, because by default we plan to generate every type of content, so it will be used only if they want to not generate something (eg. turn off Ansible remediations for a specific rule). I'm fine with either way. We just need to decide. @yuumasato @matejak

Comment (redhatrises) [reply linux_os/guide/system/software/system-tools/package_gcc_installed/rule.yml]: Still don't see how it makes it error prone since building all templates will be the default unless something like `ansible: off` is added.

Comment (yuumasato) [reply linux_os/guide/system/software/system-tools/package_gcc_installed/rule.yml]: We will be fine with or without `backends` key.  I don't have a strong opinion for either side.

File: ssg/templates.py [hunk 2]
```diff
@@ -0,0 +1,193 @@
...
+        """
+        Builds templated content for a given rule for a given language,
+        based on parameters obtained from template_data key in rule.yml.
+        Writes the output to the correct build directories.
+        """
+        template_file_name = "template_{0}_{1}".format(
+            lang.upper(), rule.template)
+        template_file_path = os.path.join(
+            self.templates_dir, template_file_name)
+        if not os.path.exists(template_file_path):
```
yuumasato: I suggest to add an error message here.

linux_os/guide/system/software/system-tools/package_gcc_installed/rule.yml
matejak: Let's have the `backends` key there, as it will make the template definition more readable, and it will allow the parsing code to be a bit simpler. We will need it only for a very small amount of rules anyway.

ssg/templates.py
```diff
@@ -0,0 +1,193 @@
...
+                # OVAL checks need to be put to a different directory because
+                # they are processed differently than remediations later in the
+                # build process
+                output_dir = self.checks_dir
+            else:
+                output_dir = self.remediations_dir
+            dir_ = os.path.join(output_dir, lang)
+            self.output_dirs[lang] = dir_
+
+    def preprocess_data(self, template, raw_parameters):
```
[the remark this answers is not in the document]
jan-cerny: It could, but it would make the code more convoluted. I would prefer investigating if we can merge the templates together.

ssg/templates.py
jan-cerny: It will not work now. By default, we want to try to generate every type of content. Some templates don't have an Anaconda or Puppet option, so it's expected that the file doesn't exist.

I think if we want a meaningful error message here we would have to explicitly enumerate which templates are available for each template_name.

ssg/templates.py
```diff
@@ -0,0 +1,193 @@
...
+    "service_disabled": None,
+    "service_enabled": None,
+    "sshd_lineinfile": None,
+    "sysctl": None,
+    "sysctl_ipv6": None,
+    "sysctl_runtime": None,
+    "sysctl_runtime_var": None,
+    "sysctl_static": None,
+    "sysctl_static_var": None,
+    "sysctl_var": None,
```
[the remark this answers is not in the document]
jan-cerny: We had a discussion about that and the outcome is to consolidate the sysctl templates in a separate PR.

PR #4810, opened 2019-09-09
Migrate more Bash remediations to Jinja2 macros

cipherboy:
#### Description:

Per discussion in previous PRs, we'd like to migrate _all_ of the content under `shared/bash_remediations` to become Jinja2 macros under `shared/macros-bash.jinja`. This PR will do two tasks:

 1. ShellCheck each remediation helper prior to migration,
 2. Migrate it from a shell script to a Jinja2 template.

I've scoped this PR as follows:

 - `service_command` -> `bash_service_command`
 - `firefox_js_setting` -> `bash_firefox_js_setting`
 - `firefox_cfg_setting` -> `bash_firefox_cfg_setting`

#### Rationale:

 - Simplifies dependency tracking.
 - Ensures generated bash scripts contain only necessary code.


matejak:
I would start with removing the `. /usr/share/scap-security-guide/remediation_functions` lines in remediations where we can be sure that shared functions won't be used. This XSLT machinery is quite arcane, so I would say that although we may not be able to solve the issue, it is good enough if we manage to avoid it.

cipherboy:
Ah, looks like that worked @matejak! Thanks :)

cipherboy:
OK, if I migrate everything in one PR, I'll be done with it. So this PR is scoped to the above three functions. More to come later... :)

shared/macros-bash.jinja
```diff
@@ -156,6 +156,170 @@ fi
...
+# bash_service_command("enable", "bluetooth")
+# bash_service_command("disable", "bluetooth.service")
+# bash_service_command("disable", "rsh.socket", xinetd="rsh")
+#}}
+{{%- macro bash_service_command(service_state, service, xinetd="") -%}}
+local service_state="{{{ service_state }}}"
+local service="{{{ service }}}"
+local xinetd="{{{ xinetd }}}"
+
+# If systemctl is installed, use systemctl command; otherwise, use the service/chkconfig commands
```
matejak: At compile-time, we know what init system the concerned product uses, so you can go more the `bash_package_install` way, where the expanded macro is super simple.

shared/macros-bash.jinja
```diff
@@ -156,6 +156,170 @@ fi
...
+# Macro to enable or disable a particular service.
+#
+# Examples:
+#
+# bash_service_command("enable", "bluetooth")
+# bash_service_command("disable", "bluetooth.service")
+# bash_service_command("disable", "rsh.socket", xinetd="rsh")
+#}}
+{{%- macro bash_service_command(service_state, service, xinetd="") -%}}
+local service_state="{{{ service_state }}}"
```
matejak: Heads up - this work is about getting rid of functions, so the `local` keyword should typically be removed.

cipherboy:
How does this look? :)

matejak:
Looks great, thanks!

PR #4811, opened 2019-09-10
SSG Test Suite: Continue even when rule is not found on benchmark.

ggbecker:
#### Description:

- Continue evaluating rules even when some rule is not found on benchmark.

#### Rationale:

- When using `ALL` in rule mode, the execution will not stop if it encounters a rule which is not present in the tested benchmark from the datastream.

#### Additional Information:

```
ERROR - Rule 'xccdf_org.ssgproject.content_rule_sshd_use_approved_ciphers' isn't present in benchmark 'xccdf_org.ssgproject.content_benchmark_RHEL-8' in './content/build/ssg-rhel8-ds.xml'
```


matejak:
LGTM

Issue #2195, opened 2017-08-01
Fixes applies but oval check fails


MollyJoBault:
I've been doing some testing today.  I have some errors that seem to be similar so I figured I would make tickets for the ones I didn't know how to fix.

The below rules say the fix was applied successfully.  If I look in the filesystem, it looks like the fix was applied successfully but oval check says it failed.  Are these incomplete or is the oval check broken?

```
	xccdf_org.ssgproject.content_rule_aide_use_fips_hashes - fix applied, oval check fails - NA for CentOS 7.3 since it isn't a cerified FIP system
	xccdf_org.ssgproject.content_rule_mount_option_tmp_nodev - fix applied, oval check fails
	xccdf_org.ssgproject.content_rule_mount_option_tmp_noexec - fix applied, oval check fails 
	xccdf_org.ssgproject.content_rule_mount_option_tmp_nosuid - fix applied, oval check fails
	xccdf_org.ssgproject.content_rule_mount_option_var_tmp_bind - fix applied, oval check fails
	xccdf_org.ssgproject.content_rule_file_permissions_etc_shadow - fix applied, oval check fails
	xccdf_org.ssgproject.content_rule_audit_rules_privileged_commands - fix applied, oval check fails (format different than description, does this matter?)
	xccdf_org.ssgproject.content_rule_audit_rules_kernel_module_loading_init - fix applied, oval check fails (format different than description)
	xccdf_org.ssgproject.content_rule_audit_rules_kernel_module_loading_delete - fix applied, oval check fails (format different than description)
	xccdf_org.ssgproject.content_rule_umask_for_daemons - fix applied, oval check fails
```

redhatrises:
Can you provide more information such as the fstab and mtab file and what the actual audit fixes were? I am also assuming that this is the latest off of the master?

MollyJoBault:
You are correct, this is the latest off of the master.  

In fstab, 

- nodev, noexec, nosuid are applied to the tmp dir.  
- /var/tmp and /tmp are bind'ed. 

[fstab.txt](https://github.com/OpenSCAP/scap-security-guide/files/1191263/fstab.txt)
[mounts.txt](https://github.com/OpenSCAP/scap-security-guide/files/1191266/mounts.txt)

[root@devgobah-1112 openscap_data]# ls -al /etc/shadow
----------. 1 root root 948 Jul 31 11:03 /etc/shadow

redhatrises:
[...]

PR #4813, opened 2019-09-10
Fix /etc/shadow permissions documentation

70k10:
#### Description:

Fix the description and ocil permissions to match what is set in the template csv.

#### Rationale:
Check does not match documentation leading to user confusion and potential breakage.

Probably related to #2195


redhatrises:
@openscap-ci test this please

redhatrises:
@70k10 great catch! It is however not really related to #2195

redhatrises:
ACK.

PR #4817, opened 2019-09-10
update kubelet_enable_streaming_connections variable handling

### LINKED ISSUE #4316: update kubelet_enable_streaming_connections variable handling
The OVAL check should evaluate if the setting is > 0 < variable-defined setting.

redhatrises:
- Make sure that timeouts are not disabled
- Allow variable to be edited
- Fixed #4316

### DIFF
--- a/applications/openshift/kubelet/kubelet_enable_streaming_connections/oval/shared.xml
+++ b/applications/openshift/kubelet/kubelet_enable_streaming_connections/oval/shared.xml
@@ -17,6 +17,7 @@
   <ind:textfilecontent54_test check="all" check_existence="any_exist" comment="streaming-connection-timeout is configured" id="test_kubelet_enable_streaming_connections" version="1">
     <ind:object object_ref="object_kubelet_enable_streaming_connections" />
     <ind:state state_ref="state_kubelet_enable_streaming_connections" />
+    <ind:state state_ref="state_kubelet_enable_streaming_connections_nonzero" />
   </ind:textfilecontent54_test>
 
   <ind:textfilecontent54_object id="object_kubelet_enable_streaming_connections" version="1">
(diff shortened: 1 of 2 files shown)


redhatrises:
Something to be aware of that the OVAL will not handle checking if the timeout is less than the variable as it is extremely difficult to handle the string timeouts e.g. 10h30m20s

shawndwells:
> Something to be aware of that the OVAL will not handle checking if the timeout is less than the variable as it is extremely difficult to handle the string timeouts e.g. 10h30m20s

Code is good. Where should this limitation be documented? Maybe a note in the xccdf?

redhatrises:
> > Something to be aware of that the OVAL will not handle checking if the timeout is less than the variable as it is extremely difficult to handle the string timeouts e.g. 10h30m20s
> 
> Code is good. Where should this limitation be documented? Maybe a note in the xccdf?

@shawndwells it is documented by Kubernetes to use the string format at https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/

shawndwells:
> > > Something to be aware of that the OVAL will not handle checking if the timeout is less than the variable as it is extremely difficult to handle the string timeouts e.g. 10h30m20s
> > 
> > 
> > Code is good. Where should this limitation be documented? Maybe a note in the xccdf?
> 
> @shawndwells it is documented by Kubernetes to use the string format at https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/

Note so much on the string format, but the contents ability to only look for exact match vs "less than or equal to".

redhatrises:
> > > > Something to be aware of that the OVAL will not handle checking if the timeout is less than the variable as it is extremely difficult to handle the string timeouts e.g. 10h30m20s
> > > 
> > > 
> > > Code is good. Where should this limitation be documented? Maybe a note in the xccdf?
> > 
> > 
> > @shawndwells it is documented by Kubernetes to use the string format at https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/
> 
> Not so much on the string format, but the contents ability to only look for exact match vs "less than or equal to".

I don't think that most people expect that ability especially with the rule providing an exact match in output. It can be an area for further research though, but I don't think that we should hold up the PR to try and figure out how to do a less than or equal for `10h30m23s`.

shawndwells:
> > > > > Something to be aware of that the OVAL will not handle checking if the timeout is less than the variable as it is extremely difficult to handle the string timeouts e.g. 10h30m20s
> > > > 
> > > > 
> > > > Code is good. Where should this limitation be documented? Maybe a note in the xccdf?
> > > 
> > > 
> > > @shawndwells it is documented by Kubernetes to use the string format at https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/
> > 
> > 
> > Not so much on the string format, but the contents ability to only look for exact match vs "less than or equal to".
> 
> I don't think that most people expect that ability especially with the rule providing an exact match in output. It can be an area for further research though, but I don't think that we should hold up the PR to try and figure out how to do a less than or equal for `10h30m23s`.

Agreed - don't want to hold up the PR to figure out how to regex that string to an int and do comparisons. Was after if we should drop an informational note in the XCCDF to remind openshift operators the OVAL will be after an exact match.

If the answer is no, that operators would not expect ``less than or equal``, then lets merge and move forward :)

PR #4820, opened 2019-09-11
Port old template preprocessing into new system - simple templates

jan-cerny:
#### Description:
For each template it tries to port logic from `shared/templates/create_<template_name>.py` to the new templating system code.

It doesn't have to port everything from these scripts, because most of the stuff like template filepaths or names of generated files are handled by the new templating system automatically. So the functions introduced in this PR contain mostly computing of default values or escaping selected characters in arguments that will be passed to the Jinja engine.

This PR converts only the `create_*.py` scripts that were easy. The pathches the remaining templates which are more difficult or convoluted will be submitted in future as separate PRs.

#### Rationale:
Some templates can't use the data from `template: vars:`in `rule.yml` directly, but there is processing needed. For example escaping values, inferring default values. Currently this is handled by classes defined in `shared/templates/create_<template_name>.py`. It is difficult to reuse these classes in the new system. First, they depend on order of arguments, but we want to have YAML dictionaries of arbitrary order of keys. Second, they have certain paths hardcoded. Third, we will have to create a lot of boilerplate code (importing classes, mapping classes to templates).

### DIFF
--- a/shared/templates/template_OVAL_service_disabled
+++ b/shared/templates/template_OVAL_service_disabled
@@ -1,5 +1,7 @@
 <def-group>
 
+{{%- set package_removed_test_id = "test_service_" + SERVICENAME + "_package_" + PACKAGENAME + "_removed" -%}}
+
 {{% if init_system == "systemd" and target_oval_version >= [5, 11] %}}
 
   {{# we are using systemd and our target OVAL version does support the systemd related tests #}}
@@ -13,7 +15,7 @@
       <description>The {{{ SERVICENAME }}} service should be disabled if possible.</description>
     </metadata>
     <criteria comment="package {{{ PACKAGENAME }}} removed or service {{{ SERVICENAME }}} is not configured to start" operator="OR">
-      <extend_definition comment="{{{ PACKAGENAME }}} removed" definition_ref="package_{{{ PACKAGENAME }}}_removed" />
+      <criterion comment="{{{ PACKAGENAME }}} removed" test_ref="{{{ package_removed_test_id }}}" />
       <criteria operator="AND" comment="service {{{ SERVICENAME }}} is not configured to start">
         <criterion comment="{{{ SERVICENAME }}} is not running" test_ref="test_service_not_running_{{{ SERVICENAME }}}" />
         {{%- if MASK_SERVICE %}}
@@ -101,7 +103,7 @@
       <description>The {{{ SERVICENAME }}} service should be disabled if possible.</description>
     </metadata>
    <criteria comment="package {{{ PACKAGENAME }}} removed or service {{{ SERVICENAME }}} is not configured to start" operator="OR">
-    <extend_definition comment="{{{ PACKAGENAME }}} removed" definition_ref="package_{{{ PACKAGENAME }}}_removed" />
+    <criterion comment="{{{ PACKAGENAME }}} removed" test_ref="{{{ package_removed_test_id }}}" />
 
     <!-- OVAL <runlevel_test> is not implemented in OpenSCAP when compiled on Ubuntusystems yet:
          https://github.com/OpenSCAP/openscap/blob/maint-1.2/src/OVAL/probes/unix/runlevel.c#L210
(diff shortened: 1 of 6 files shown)


yuumasato:
Looking good so far. @jan-cerny would you like help with this?

jan-cerny:
@yuumasato 

Unfortunately, it doesn't work at this moment because I rebased on the master which already contains template data in rule.yml.

First reason is that we haven't added a processor for the variables with `@`, eg. `pkgname@fedora:`. That should be done in a separate PR.

Second, for this moment we should somehow skip the templates that don't have the callback yet to be able to always finish the build without errors. Because at this moment, the if there is no callback from a template, the generated content might be incomplete or invalid, and some checks later in the process are triggered. I suggest changing the system in a way that if no callback is present the templated content is not generated. That could be done also in separate PR.

After the 2 changes are done this PR can be rebased on master and then reviewed and merged, and the remaining callbacks probably could be done in multiple separate pull requests.

jan-cerny:
This PR adds new callbacks only for some of the templates - those that were very easy to port. The remaining templates will be done later in separate PRs. Therefore this is now ready for a review.

ssg/templates.py
```diff
@@ -30,38 +112,70 @@ def package_installed(data, lang):
     return data
 
 
+def package_removed(data, lang):
+    return data
+
+
+def service_disabled(data, lang):
```
yuumasato: OVAL checks for some `services_enabled` rules became incomplete, this is because OVAL template for `service_disabled` has an `extend_definition` for package remove.

For these cases, these services/packages have an entry for the service and package in both `services_disabled` and `packages_removed.csv. But there is no `rule.yml` for `package_removed`.
Examples: `acpid`, `autofs`.

For most cases, `service_disabled` can mean `package_removed`, unless the package of the service brings in some dependencies. I don't know of any such case right now, but it could happen.

A workaround would be for `service_disabled` preprocessor to also "inject" `package_removed` template and also generate this definition. I don't think the current mechanism allows for that though, :(

ssg/templates.py
```diff
@@ -25,6 +25,17 @@ def accounts_password(data, lang):
     return data
 
 
+def audit_rules_privileged_commands(data, lang):
+    path = data["path"]
+    name = os.path.basename(path).replace(r'[-\./]', '_')
+    data["name"] = name
+    if lang == "oval":
+        data["id"] = "audit_rules_execution_" + name
+        data["title"] = "Record Any Attempts to Run " + name
```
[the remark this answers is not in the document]
jan-cerny: Let's pass the rule ID into the template data by the templates script (simple oneliner).  Then, change the affected OVAL templates. At some moment, we should change all the OVAL templates to use the rule ID is the OVAL definition IDs.

ssg/templates.py
jan-cerny: The OVAL template should be able to be handle to generate both, so we should merge them together. We can leverage the Jinja macros to not repeat ourselves.

jan-cerny:
I have added commits that hopefully resolve the audit_rules_privileged_commands / audit_rules_execution problem and the service_disabled problem as well.

ssg/templates.py
```diff
@@ -19,6 +19,88 @@
...
+
+def audit_rules_unsuccessful_file_modification_o_creat(data, lang):
+    return data
+
+
+def audit_rules_unsuccessful_file_modification_o_trunc_write(data, lang):
+    return data
+
+
+def audit_rules_unsuccessful_file_modification_rule_order(data, lang):
```
matejak: What about defining a function `def noop(data, lang): return lang` and using that `noop` wherever applicable?

jan-cerny: That would be awesome if it wouldn't be already done the other way.

shared/templates/template_OVAL_service_disabled
```diff
@@ -278,4 +280,6 @@
 
 {{% endif %}}
 
+{{{ oval_test_package_removed(package=PACKAGENAME, test_id=package_removed_test_id) }}}
+
```
jan-cerny: The template template_OVAL_service_enabled has a similar problem which needs to be addressed.

jan-cerny:
Reviewers, should the last 2 commits (which change the templates themselves and Jinja macro) be separated to a separate PR?

jan-cerny:
I have rebased this on the top of master branch.

shared/templates/template_OVAL_service_disabled
jan-cerny: Should be fixed in 2c0f8e3

shared/templates/template_OVAL_service_enabled
```diff
@@ -11,7 +13,7 @@
       <description>The {{{ SERVICENAME }}} service should be enabled if possible.</description>
     </metadata>
     <criteria comment="package {{{ PACKAGENAME }}} installed and service {{{ SERVICENAME }}} is configured to start" operator="AND">
-    <extend_definition comment="{{{ PACKAGENAME }}} installed" definition_ref="package_{{{ PACKAGENAME }}}_installed" />
+    <criterion comment="{{{ PACKAGENAME }}} installed" test_ref="{{{ package_installed_test_id }}}" />
```
yuumasato: If a rule for the package installed already exists, this will end up creating two similar OVAL objects:
`test_service_fapolicyd_package_fapolicyd_installed`and
`test_package_fapolicyd_installed`

```
<ns8:rpminfo_object id="oval:ssg-obj_test_service_fapolicyd_package_fapolicyd_installed:obj:1" version="1">
    <ns8:name>fapolicyd</ns8:name>
</ns8:rpminfo_object>
```

``` 
<ns8:rpminfo_object id="oval:ssg-obj_test_package_fapolicyd_installed:obj:1" version="1">
    <ns8:name>fapolicyd</ns8:name>
</ns8:rpminfo_object>
```

PR #4826, opened 2019-09-12
Ignore Visual Studio Code configuration files

jan-cerny:
#### Description:
Ignore Visual Studio Code configuration files


ggbecker:
LGTM. I use VS Code and can confirm that `.vscode` is the right folder for configuration files.

PR #4828, opened 2019-09-12
Document new templating system in Developer Guide

jan-cerny:
#### Description:
In the new Sections of Developer Guide we will describe the new templating system, using it in rules and developing a new template.

#### Rationale:
#4809 introduces a change in templating

### DIFF
--- a/docs/manual/developer_guide.adoc
+++ b/docs/manual/developer_guide.adoc
@@ -1677,7 +1677,7 @@ The function must always return the (modified) `data` dictionary.
 The function must be always defined even if no processing of data is needed.
 In that situation the function just returns `data` parameter.
 
-The following example shows callback for template `mount_option`:
+The following example shows the callback function for the template `mount_option`:
 
 [source,python]
 ----
--- a/docs/manual/developer_guide.adoc
+++ b/docs/manual/developer_guide.adoc
@@ -1452,7 +1452,7 @@ audit_rules_unsuccessful_file_modification_o_creat::
 * Languages: OVAL
 
 audit_rules_unsuccessful_file_modification_o_trunc_write::
-* Ensure there is an Audit rule to record unsuccessful access attempts to files when O_TRUNC_WRITE flag is specified.
+* Ensure there is an Audit rule to record unsuccessful attempts to access files when O_TRUNC_WRITE flag is specified.
 * Parameters:
 ** *syscall* - name of the unsuccessful system call, eg. `openat`
 ** *pos* - position of the O_TRUNC_WRITE argument in the syscall, as specified by `-F` audit rule argument, eg. `a2`
--- a/docs/manual/developer_guide.adoc
+++ b/docs/manual/developer_guide.adoc
@@ -1439,7 +1439,7 @@ audit_rules_privileged_commands::
 * Languages: Ansible, Bash, OVAL
 
 audit_rules_unsuccessful_file_modification::
-* Ensure there is an Audit rule to record unsuccessful access attempts to files
+* Ensure there is an Audit rule to record unsuccessful attempts to access files
 * Parameters:
 ** *name* - name of the unsuccessful system call, eg. `creat`
 * Languages: Ansible, Bash, OVAL
--- a/docs/manual/developer_guide.adoc
+++ b/docs/manual/developer_guide.adoc
@@ -1337,26 +1337,406 @@ Available low-level Jinja macros that can be used in Bash remediations:
 - `die` - Function to terminate the remediation
 - `set_config_file` - Add an entry to a text configuration file
 
-==== Templating
+=== Templating
+
+Writing OVAL checks, Bash, or any other content can be tedious work. For
+certain types of rules we provide templates. If there is a template that can be
+used for the new rule you only need to specify the template name and its parameters in
+`rule.yml` and the content will be generated during the build.
+
+The templating system currently supports generating OVAL checks and Ansible,
+Bash, Anaconda and Puppet remediations.  All templates can be found in
+`link:{rootdir}/shared/templates[shared/templates]` directory. The files are
+named `template_<TYPE>_<NAME>`, where `<TYPE>` should be OVAL, ANSIBLE, BASH,
+ANACONDA or PUPPET and `<NAME>` is the template name.
+
+==== Using Templates
+
+To use a template in `rule.yml` add `template:` key there and fill it
+accordingly. The general form is the following:
+
+----
+template:
+    name: template_name
+    vars:
+        param_name: value # these parameters are individual for each template
+        param_name@rhel7: value1
+        param_name@rhel8: value2
+    backends: # optional
+        ansible: "off"
+        bash: "on" # on is implicit value
+----
+
+The `vars:` key contains template parameters and their values which will be
+substituted into the template.  Each template has specific parameters.  To use
+different values of parameters based on product, append `@` followed by product
+ID to the parameter name.
+
+The `backends:` key is optional. By default, all languages supported by a given template will be generated.
+with given name exist will be generated. This key can be used to explicitly opt
+out from generating a certain type of content for the rule.
+
+For example, to generate templated content except Bash remediation for rule
+"Package GCC is Installed" using `package_installed` template, add the
+following to `rule.yml`:
+
+----
+template:
+    name: package_installed
+    vars:
+        pkgname: gcc
+    backends:
+        bash: "off"
+----
+
+==== List of available templates
+
+accounts_password::
+* Checks if PAM enforces password quality requirements. Checks the configuration in `/etc/pam.d/system-auth` (for RHEL 6 systems) or `/etc/security/pwquality.conf` (on other systems).
+* Parameters:
+** *variable* - PAM `pam_cracklib` (on RHEL 6) or `pam_pwquality` (on other systems) module name, eg. `ucredit`, `ocredit`
+** *operation* - OVAL operation, eg. `less than or equal`
+* Languages: OVAL
+
+auditd_lineinfile::
+* Checks configuration options of the Audit Daemon in `/etc/audit/auditd.conf`.
+* Parameters:
+** *parameter* - auditd configuration item
+** *value* - the value of configuration item specified by parameter
+** *missing_parameter_pass* - effective only in OVAL checks, if set to `"true"` and the parameter is not present in the configuration file the OVAL check will return true.
+* Languages: Ansible, Bash, OVAL
+
+audit_rules_dac_modification::
+* Checks Audit Discretionary Access Control rules
+* Parameters:
+** *attr* - value of `-S` argument in Audit rule, eg. `chmod`
+* Languages: Ansible, Bash, OVAL
+
+audit_rules_file_deletion_events::
+* Ensure auditd Collects file deletion events
+* Parameters:
+** *name* - value of `-S` argument in Audit rule, eg. `unlink`
+* Languages: Ansible, Bash, OVAL
+
+audit_rules_login_events::
+* Checks if there are Audit rules that record attempts to alter logon and logout events.
+* Parameters:
+** *path* - value of `-w` in the Audit rule, eg. `/var/run/faillock`
+* Languages: Ansible, Bash, OVAL
+
+audit_rules_path_syscall::
+* Check if there are Audit rules to record events that modify user/group information via a syscall on a specific file.
+* Parameters:
+** *path* - path of the protected file, eg `/etc/shadow`
+** *pos* - position of argument, eg. `a2`
+** *syscall* - name of the system call, eg. `openat`
+* Languages: Ansible, Bash, OVAL
+
+audit_rules_privileged_commands::
+* Ensure Auditd collects information on the use of specified privileged command.
+* Parameters:
+** *path* - the path of the privileged command - eg. `/usr/bin/mount`
+* Languages: Ansible, Bash, OVAL
+
+audit_rules_unsuccessful_file_modification::
+* Ensure there is an Audit rule to record unsuccessful access attempts to files
+* Parameters:
+** *name* - name of the unsuccessful system call, eg. `creat`
+* Languages: Ansible, Bash, OVAL
+
+audit_rules_unsuccessful_file_modification_o_creat::
+* Ensure there is an Audit rule to record unsuccessful access attempts to files when O_CREAT flag is specified.
+* Parameters:
+** *syscall* - name of the unsuccessful system call, eg. `openat`
+** *pos* - position of the O_CREAT argument in the syscall, as specified by `-F` audit rule argument, eg. `a2`
+* Languages: OVAL
+
+audit_rules_unsuccessful_file_modification_o_trunc_write::
+* Ensure there is an Audit rule to record unsuccessful access attempts to files when O_TRUNC_WRITE flag is specified.
+* Parameters:
+** *syscall* - name of the unsuccessful system call, eg. `openat`
+** *pos* - position of the O_TRUNC_WRITE argument in the syscall, as specified by `-F` audit rule argument, eg. `a2`
+* Languages: OVAL
+
+audit_rules_unsuccessful_file_modification_rule_order::
+* Ensure that Audit rules for unauthorized attempts to a specific system call are ordered correctly.
+* Parameters:
+** *syscall* - name of the unsuccessful system call, eg. `openat`
+** *pos* - position of the flag parameter in the syscall, as specified by `-F` audit rule argument, eg. `a2`
+* Languages: OVAL
+
+audit_rules_usergroup_modification::
+* Check if Audit is configured to record events that modify account changes.
+* Parameters:
+** *path* - path that should be part of the audit rule as a value of `-w` argument, eg. `/etc/group`.
+* Languages: Ansible, Bash, OVAL
+
+file_groupowner::
+* Check group that owns the given file.
+* Parameters:
+** *filepath* - File path to be checked. If the file path ends with `/` it describes a directory.
+** *file_regex* - Regular expression that matches file names in a directory specified by *filepath*. Can be set only if *filepath* parameter specifies a directory.
+** *filegid* - group ID (GID)
+* Languages: Ansible, Bash, OVAL
+
+file_owner::
+* Check user that owns the given file.
+* Parameters:
+** *filepath* - File path to be checked. If the file path ends with `/` it describes a directory.
+** *file_regex* - Regular expression that matches file names in a directory specified by *filepath*. Can be set only if *filepath* parameter specifies a directory.
+** *fileuid* - user ID (UID)
+* Languages: Ansible, Bash, OVAL
+
+file_permissions::
+* Checks permissions (mode) on a given file.
+* Parameters:
+** *filepath* - File path to be checked. If the file path ends with `/` it describes a directory.
+** *file_regex* - Regular expression that matches file names in a directory specified by *filepath*. Can be set only if *filepath* parameter specifies a directory.
+** *filemode* - File permissions in a hexadecimal format, eg. `'0640'`.
+* Languages: Ansible, Bash, OVAL
+
+grub2_bootloader_argument::
+* Checks kernel command line arguments in GRUB 2 configuration.
+* Parameters:
+** *arg_name* - argument name, eg. `audit`
+** *arg_value* - argument value, eg. `'1'`
+* Languages: Ansible, Bash, OVAL
+
+kernel_module_disabled::
+* Checks if the given Linux kernel module is disabled.
+* Parameters:
+** *kernmodule* - name of the Linux kernel module, eg. `cramfs`
+* Languages: Ansible, Bash, OVAL
+
+mount::
+* Checks that a given mount point is located on a separate partition.
+* Parameters:
+** *mountpoint* - path to the mount point, eg. `/var/tmp`
+* Languages: Anaconda, OVAL
+
+mount_option::
+* Checks if a given partition is mounted with a specific option such as "nosuid".
+* Parameters:
+** *mountpoint* - mount point on the filesystem eg. `/dev/shm`
+** *mountoption* - mount option, eg. `nosuid`
+** *filesystem* - filesystem in `/etc/fstab`, eg. `tmpfs`. Used only in Bash remediation.
+** *type* - filesystem type. Used only in Bash remediation.
+** *mount_has_to_exist* - Used only in Bash remediation. Specifies if the *mountpoint* entry has to exist in `/etc/fstab` before the remediation is executed. If set to `yes` and the *mountpoint* entry is not present in `/etc/fstab` the Bash remediation terminates. If set to `no` the *mountpoint* entry will be created in `/etc/fstab`.
+* Languages: Anaconda, Ansible, Bash, OVAL
+
+mount_option_remote_filesystems::
+* Checks if all remote filesystems (NFS mounts in `/etc/fstab`) are mounted with a specific option.
+* Parameters:
+** *mountpoint* - always set to `remote_filesystems`
+** *mountoption* - mount option, eg. `nodev`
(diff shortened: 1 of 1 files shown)


jan-cerny:
I'm thinking about documenting templates themselves. Should it be a part of developer guide? Or should the templates have Jinja comments at their beginning? Do you have any preference?

jan-cerny:
We have discussed this with @ggbecker , we think that documenting it in developer guide will be more consumable by users, we can incorporate it also to the rethinked documentation.

docs/manual/developer_guide.adoc
```diff
@@ -1334,13 +1334,109 @@ Available low-level Jinja macros that can be used in Bash remediations:
...
+"Package GCC is Installed" using `package_installed` template, add the
+following to `rule.yml`:
+
+----
+template:
+    name: package_installed
+    vars:
+        pkgname: gcc
+    backends:
+        bash: off
```
matejak: Unfortunately we have to quote the boolean.

docs/manual/developer_guide.adoc
```diff
@@ -1337,13 +1337,321 @@ Available low-level Jinja macros that can be used in Bash remediations:
...
+* Check if there are Audit rules to record events that modify user/group information via a syscall on a specific file.
+* Parameters:
+** *path* - path of the protected file, eg `/etc/shadow`
+** *pos* - position of argument, eg. `a2`
+** *syscall* - name of the system call, eg. `openat`
+* Languages: Ansible, Bash, OVAL
+
+audit_rules_privileged_commands::
+* Ensure Auditd collects information on the use of privileged commands
+* Parameters:
```
vojtapolasek: Ensure Auditd collects information on the use of specified privileged command

docs/manual/developer_guide.adoc
```diff
@@ -1337,13 +1337,321 @@ Available low-level Jinja macros that can be used in Bash remediations:
...
+* Languages: Anaconda, OVAL
+
+mount_option::
+* Checks if a given partition is mounted with a specific option such as "nosuid".
+* Parameters:
+** *mountpoint* - mount point on the filesystem eg. `/dev/shm`
+** *mountoption* - mount option, eg. `nosuid`
+** *filesystem* - filesystem in `/etc/fstab`, eg. `tmpfs`. Used only in Bash remediation.
+** *type* - filesystem type. Used only in Bash remediation.
+** *mount_has_to_exist* - Used only in Bash remediation. Specifies if the *mountpoint* entry has to exist in `/etc/fstab`. If set to `yes` and the *mountpoint* entry is not present in `/etc/fstab` the Bash remediation terminates.
```
vojtapolasek: This line confuses me. If mount has to exist (set to yes) and it is not there, I would expect the remediation to add it instead of terminating...

docs/manual/developer_guide.adoc
```diff
@@ -1337,13 +1337,321 @@ Available low-level Jinja macros that can be used in Bash remediations:
...
+* Languages: Anaconda, Ansible, Bash, OVAL
+
+mount_option_remote_filesystems::
+* Checks if all remote filesystems (NFS mounts in `/etc/fstab`) are mounted with a specific option.
+* Parameters:
+** *mountpoint* - always set to `remote_filesystems`
```
vojtapolasek: What is the purpose of this parameter if it is ALWAYS set to something?

docs/manual/developer_guide.adoc
```diff
@@ -1337,13 +1337,321 @@ Available low-level Jinja macros that can be used in Bash remediations:
...
+** *mountoption* - mount option, eg. `nodev`
+** *filesystem* - filesystem of new mount point (used when adding new entry in `/etc/fstab`), eg. `tmpfs`. Used only in Bash remediation.
+* Languages: Ansible, Bash, OVAL
+
+mount_option_removable_partitions::
+* Checks if all removable media mounts are mounted with a specific option.
+* Parameters:
+** *mountpoint* - always set to `remote_filesystems`
```
vojtapolasek: What is the purpose of this parameter? Why talking about remote_systems when the template is concerning removable filesystems?

docs/manual/developer_guide.adoc
```diff
@@ -1337,13 +1337,321 @@ Available low-level Jinja macros that can be used in Bash remediations:
...
+** *mountoption* - mount option, eg. `nodev`
+** *filesystem* - filesystem of new mount point (used when adding new entry in `/etc/fstab`), eg. `tmpfs`. Used only in Bash remediation.
+* Languages: Anaconda, Ansible, Bash, OVAL
+
+package_installed::
+* Checks if a given package is installed. Optionally, it can also check whether a specific version or newer is installed.
+* Parameters:
+** *pkgname* - name of the RPM or DEB package, eg. `tmux`
+** *evr* - Optional parameter. Use to check a specific epoch, version, release in `epoch:version-release` format. Used only in OVAL checks.
```
vojtapolasek: Could you please give example of such parameter?

docs/manual/developer_guide.adoc
```diff
@@ -1337,13 +1337,321 @@ Available low-level Jinja macros that can be used in Bash remediations:
...
+mount_option_removable_partitions::
+* Checks if all removable media mounts are mounted with a specific option.
+* Parameters:
+** *mountpoint* - always set to `remote_filesystems`
+** *mountoption* - mount option, eg. `nodev`
+** *filesystem* - filesystem of new mount point (used when adding new entry in `/etc/fstab`), eg. `tmpfs`. Used only in Bash remediation.
+* Languages: Anaconda, Ansible, Bash, OVAL
+
+package_installed::
+* Checks if a given package is installed. Optionally, it can also check whether a specific version or newer is installed.
```
vojtapolasek: In the parameters below, the evr parameter is said to check for specific value. How to specify that we want newer version, eg greater or equal to the evr?

docs/manual/developer_guide.adoc
```diff
@@ -1337,13 +1337,321 @@ Available low-level Jinja macros that can be used in Bash remediations:
...
+* Checks if the given package is not installed.
+* Parameters:
+** *pkgname* - name of the RPM or DEB package, eg. `tmux`
+* Languages: Anaconda, Ansible, Bash, OVAL, Puppet
+
+sebool::
+* Checks values of SELinux booleans.
+* Parameters:
+** *seboolid* - name of SELinux boolean, eg. `cron_userdomain_transition`
+** *sebool_bool* - the value of the SELinux Boolean. Can be either `"true"` or `"false"`. If this parameter is not specified, the rule will use XCCDF Value `var_<seboolid>`
```
vojtapolasek: Could you explain more about var_seboolid? Where is it specified, what is it? Or is it supposed that an user already knows that?

docs/manual/developer_guide.adoc
```diff
@@ -1337,13 +1337,321 @@ Available low-level Jinja macros that can be used in Bash remediations:
...
+* Checks SSH server configuration items in `/etc/ssh/sshd_config`.
+* Parameters:
+** *parameter* - name of the SSH configuration option, eg. `KerberosAuthentication`
+** *value* - value of the SSH configuration option specified by *parameter*, eg. `"no"`.
+** *missing_parameter_pass* - If set to `"true"` the OVAL check will pass if the parameter is not present in `/etc/ssh/sshd_config`.
+* Languages: Ansible, Bash, OVAL
+
+sysctl::
+* Checks sysctl parameters.
+* Parameters:
```
vojtapolasek: Is it worth mentioning where are these parameters checked? file names, etc?

docs/manual/developer_guide.adoc
```diff
@@ -1337,13 +1337,321 @@ Available low-level Jinja macros that can be used in Bash remediations:
...
+** *sysctlvar* - name of the sysctl value, eg. `net.ipv4.conf.all.secure_redirects`.
+** *datatype* - data type of the sysctl value, eg. `int`.
+** *sysctlval* - value of the sysctl value, eg. `'1'`. If this parameter is not specified, XCCDF Value is used instead.
+* Languages: Ansible, Bash, OVAL
+
+timer_enabled::
+* Checks if a SystemD timer unit is enabled.
+* Parameters:
+** *timername* - name of the SystemD timer unit, without the `timer` suffix, eg. `dnf-automatic`.
+** *packagename* - name of the RPM package which provides the SystemD timer unit. This parameter is optional, if it is not provideded it is assumed that the name of the RPM package is the same as the name of the SystemD timer unit.
```
vojtapolasek: just a simple mistake, not provideded

docs/manual/developer_guide.adoc
```diff
@@ -1337,13 +1337,321 @@ Available low-level Jinja macros that can be used in Bash remediations:
...
+named `template_<TYPE>_<NAME>`. Where `<TYPE>` should be OVAL, ANSIBLE, BASH,
+ANACONDA or PUPPET and `<NAME>` is the what we will call the template name.
+
+Use the Jinja syntax we use elsewhere in the project; refer to the earlier
+section on Jinja macros for more information.  The parameters should be named
+using lowercase letters.
+
+Notice that OVAL should be written in shorthand format.  This is an example of
+an OVAL template file called _template_OVAL_package_installed_:
+
```
vojtapolasek: I am confused. Why does the file name has underscores at the begining and end of name? Is this something adoc specific?

docs/manual/developer_guide.adoc
```diff
@@ -1337,13 +1337,321 @@ Available low-level Jinja macros that can be used in Bash remediations:
...
+----
+include::{templatesdir}/template_OVAL_package_installed
```
vojtapolasek: Is this supposed to be verbatim? Or should it show the content of some file? If yes the content is not shown, at least not on github.

docs/manual/developer_guide.adoc
```diff
@@ -1337,13 +1337,321 @@ Available low-level Jinja macros that can be used in Bash remediations:
...
+
+
+2) Implement a callback function which will process the
+parameters before passing them to Jinja engine.  For example, this callback can
+provide default values, escape characters, check if parameters are correct, or
+any other processing of the parameters specific for the template.  The callback
+function should be put into `ssg/templates.py` and must have 2 parameters:
+
+- `data` - will contain the contents of `vars:` dictionary from `rule.yml`
+- `lang` - the language (`anaconda`, `ansible`, `bash`, `oval`, `puppet`)
```
vojtapolasek: How is this parameter used? Will it be populated with some list / array?

docs/manual/developer_guide.adoc
```diff
@@ -1337,13 +1337,321 @@ Available low-level Jinja macros that can be used in Bash remediations:
 - `die` - Function to terminate the remediation
 - `set_config_file` - Add an entry to a text configuration file
 
-==== Templating
```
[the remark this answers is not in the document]
ggbecker: I agree, you cannot find it from the table of contents, maybe increasing the table of content depth can also be beneficial for other sections.

docs/manual/developer_guide.adoc
```diff
@@ -1337,13 +1337,321 @@ Available low-level Jinja macros that can be used in Bash remediations:
...
+
+==== Using Templates
+
+Writing OVAL checks, Bash, or any other content can be tedious work. For
+certain types of rules we provide templates. If there is a template that can be
+used for the new rule you only specify the template name and its parameters in
```
ggbecker:
```suggestion
used for the new rule you only need to specify the template name and its parameters in
```

docs/manual/developer_guide.adoc
```diff
@@ -1337,13 +1337,321 @@ Available low-level Jinja macros that can be used in Bash remediations:
...
+is the template name.
+
+To use a template in `rule.yml` add `template:` key there and fill it
+accordingly. The general form is the following:
+
+----
+template:
+    name: template_name
+    vars:
+        param_name: value # these parameters individual for each template
```
ggbecker: Didn't understand the `these parameters individual for each template`.
Maybe:
```suggestion
        param_name: value # these parameters are individual for each template
```

docs/manual/developer_guide.adoc
```diff
@@ -1337,13 +1337,321 @@ Available low-level Jinja macros that can be used in Bash remediations:
...
+        param_name@rhel7: value1
+        param_name@rhel8: value2
+    backends: # optional
+        ansible: "off"
+        bash: "on" # on is implicit value
+----
+
+The `vars:` key contain template parameters and their values which will be
```
ggbecker:
```suggestion
The `vars:` key contains template parameters and their values which will be
```

docs/manual/developer_guide.adoc
```diff
@@ -1337,13 +1337,321 @@ Available low-level Jinja macros that can be used in Bash remediations:
...
+substituted into the template.  Each template has specific parameters.  To use
+different values of parameters based on product, append `@` followed by product
+ID to the parameter name.
+
+The `backends` key is optional. By default, every language for which a template
```
ggbecker:
```suggestion
The `backends` key is optional. By default, all languages supported by a given template will be generated
```

docs/manual/developer_guide.adoc
ggbecker: This is actually the usual XCCDF variables that you can define in profiles, there might be some documentation available on how to use them but I'm not sure where to right now. The seboolid will be replaced by some selinux boolean, for example: selinuxuser_execheap and in the profile you can use var_selinuxuser_execheap to turn on or off the boolean.

ggbecker: @jan-cerny Referencing the section which tells on how to use variables can be a good thing here.

docs/manual/developer_guide.adoc
ggbecker: It is, it formats the text to italic.

docs/manual/developer_guide.adoc
ggbecker: From adoc docs: https://github.com/asciidoctor/asciidoctor.org/blob/master/docs/_includes/include-directive.adoc
```suggestion
include::{templatesdir}/template_OVAL_package_installed[]
```

ggbecker: @jan-cerny please fix all `include::` directives

docs/manual/developer_guide.adoc
```diff
@@ -1337,13 +1337,321 @@ Available low-level Jinja macros that can be used in Bash remediations:
...
+- `data` - will contain the contents of `vars:` dictionary from `rule.yml`
+- `lang` - the language (`anaconda`, `ansible`, `bash`, `oval`, `puppet`)
+
+The function must always return the (modified) `data` dictionary.
+
+The function must be always defined even if no processing of data is needed.
+In that situation the function just returns `data` parameter.
+
+3) Add the template name to `templates` dictionary in `ssg/templates.py`. This
+dictionary maps template name to the callback function name.
```
ggbecker:
```suggestion
dictionary maps the template name to the callback function name.
```

docs/manual/developer_guide.adoc
```diff
@@ -1337,13 +1337,321 @@ Available low-level Jinja macros that can be used in Bash remediations:
...
+----
+include::{templatesdir}/template_ANSIBLE_package_installed
+----
+
+
+2) Implement a callback function which will process the
+parameters before passing them to Jinja engine.  For example, this callback can
+provide default values, escape characters, check if parameters are correct, or
+any other processing of the parameters specific for the template.  The callback
+function should be put into `ssg/templates.py` and must have 2 parameters:
```
ggbecker:
```suggestion
function should be put into `link:{rootdir}/ssg/templates.py[templates.py]` and must have 2 parameters:
```

docs/manual/developer_guide.adoc
ggbecker: @jan-cerny does the callback function get executed once for every supported language for current template? So in theory I could process the data differently for each supported language. Is that the current behavior?

docs/manual/developer_guide.adoc
ggbecker: That's strange, I have tested using `off` without quoting and the build system was able to switch off correctly the templated content. Do you remember what was the scenario where quoting is required?

jan-cerny: @ggbecker The problem is when you use Python 2 and you don't use quotes you will get a traceback due to some explicit "unicode" typecast in the generated processed rule.yml YAML file. 

For this moment, we will use quotes in the documentation.

But in future I wish to remove https://github.com/ComplianceAsCode/content/blob/master/ssg/yaml.py#L27 , which is causing this, but it needs to be investigated what else it could possibly break.

docs/manual/developer_guide.adoc
jan-cerny: It's really like that :smile:  I'll try to improve the wording.

docs/manual/developer_guide.adoc
jan-cerny: this is actually a copy-paste error, it should be always set to `var_removable_partition`

docs/manual/developer_guide.adoc
jan-cerny: I don't know. I think we should take a closer look at the templates. I have opened an issue in upstream https://github.com/ComplianceAsCode/content/issues/4914

docs/manual/developer_guide.adoc
jan-cerny: yes

docs/manual/developer_guide.adoc
jan-cerny: Actually, it checks if the given version or a newer version is installed, because OVAL check uses  operation "greater than or equal".    It's here:  https://github.com/ComplianceAsCode/content/blob/85bdd8bbb8753f77f6c728811c89f026b76b4d8b/shared/macros-oval.jinja#L371

At this moment it isn't possible to parametrize the operation, eg. to check for older version or for the exactly the same version.


I'll try to improve the wording.

docs/manual/developer_guide.adoc
jan-cerny: Do we have any section like that?

ggbecker: To be honest I don't know, I just assumed there was something.

docs/manual/developer_guide.adoc
jan-cerny: @vojtapolasek it's a string, it can be exactly one of the aforementioned values, the function is called for each supported language.

@ggbecker Yes, it gets executed once for every supported language for current template. Yes, you can process data differently, that is the current behavior, it's done like that in some of the callbacks.

jan-cerny: I'll try to improve the wording

docs/manual/developer_guide.adoc
```diff
@@ -1337,34 +1337,421 @@ Available low-level Jinja macros that can be used in Bash remediations:
...
+
+audit_rules_unsuccessful_file_modification_o_trunc_write::
+* Ensure there is an Audit rule to record unsuccessful access attempts to files when O_TRUNC_WRITE flag is specified.
+* Parameters:
+** *syscall* - name of the unsuccessful system call, eg. `openat`
+** *pos* - position of the O_TRUNC_WRITE argument in the syscall, as specified by `-F` audit rule argument, eg. `a2`
+* Languages: OVAL
+
+audit_rules_unsuccessful_file_modification_rule_order::
+* Ensure that Audit rules for unauthorized attempts to a specific system call are ordered correctly.
```
vojtapolasek:
```suggestion
* Ensure that Audit rules for unauthorized attempts to use a specific system call are ordered correctly.
```

docs/manual/developer_guide.adoc
```diff
@@ -1337,34 +1337,421 @@ Available low-level Jinja macros that can be used in Bash remediations:
...
+** *mountpoint* - always set to `remote_filesystems`
+** *mountoption* - mount option, eg. `nodev`
+** *filesystem* - filesystem of new mount point (used when adding new entry in `/etc/fstab`), eg. `tmpfs`. Used only in Bash remediation.
+** *mount_has_to_exist* - Used only in Bash remediation. Specifies if the *mountpoint* entry has to exist in `/etc/fstab` before the remediation is executed. If set to `yes` and the *mountpoint* entry is not present in `/etc/fstab` the Bash remediation terminates. If set to `no` the *mountpoint* entry will be created in `/etc/fstab`.
+* Languages: Ansible, Bash, OVAL
+
+mount_option_removable_partitions::
+* Checks if all removable media mounts are mounted with a specific option.
+* Parameters:
+** *mountpoint* - always set to `var_removable_partition`
```
vojtapolasek: is this an XCCDF variable?

docs/manual/developer_guide.adoc
```diff
@@ -1337,34 +1337,421 @@ Available low-level Jinja macros that can be used in Bash remediations:
...
+** *value* - value of the SSH configuration option specified by *parameter*, eg. `"no"`.
+** *missing_parameter_pass* - If set to `"true"` the OVAL check will pass if the parameter is not present in `/etc/ssh/sshd_config`.
+* Languages: Ansible, Bash, OVAL
+
+sysctl::
+* Checks sysctl parameters. The OVAL definition checks both configuration and runtime settings and require both of them to be set to the desired value to return true.
+* Parameters:
+** *sysctlvar* - name of the sysctl value, eg. `net.ipv4.conf.all.secure_redirects`.
+** *datatype* - data type of the sysctl value, eg. `int`.
+** *sysctlval* - value of the sysctl value, eg. `'1'`. If this parameter is not specified, XCCDF Value is used instead.
```
vojtapolasek: Do you mean xccdf variable?

docs/manual/developer_guide.adoc
```diff
@@ -1337,34 +1337,421 @@ Available low-level Jinja macros that can be used in Bash remediations:
...
+template.
+
+1) Create the template files, one for each type of file. Each one should be
+named `template_<TYPE>_<NAME>`. Where `<TYPE>` should be OVAL, ANSIBLE, BASH,
+ANACONDA or PUPPET and `<NAME>` is the what we will call the template name.
+Create these files in `link:{rootdir}/shared/templates[shared/templates]` directory.
+
+Use the Jinja syntax we use elsewhere in the project; refer to the earlier
+section on Jinja macros for more information.  The parameters should be named
+using lowercase letters.
```
vojtapolasek: What do you mean by named using lower case letters? I am confused because in the example oval template the parameters are in upper case.

docs/manual/developer_guide.adoc
jan-cerny: yes, I'll add it

docs/manual/developer_guide.adoc
jan-cerny: Yes, I mean a variable, but unfortunately, variables in XCCDF terminology are called XCCDF Values, and they're represented by `<xccdf:Value>` element :scream:

docs/manual/developer_guide.adoc
jan-cerny: Should be uppercase there at this moment. I'll fix that. But I think using lowercase letters this is the desired future state, because using uppercase letters is only a temporary workaround for easier transition from old system to the new system.

https://github.com/ComplianceAsCode/content/blob/85bdd8bbb8753f77f6c728811c89f026b76b4d8b/ssg/templates.py#L318-L323

jan-cerny:
@openscap-ci test this please

docs/manual/developer_guide.adoc
```diff
@@ -1337,34 +1337,422 @@ Available low-level Jinja macros that can be used in Bash remediations:
...
+- `lang` - string, describes language, can be one of: `"anaconda"`, `"ansible"`, `"bash"`, `"oval"`, `"puppet"`
+
+The callback function is executed for every supported language, so it can process the data differently for each language.
+
+The function must always return the (modified) `data` dictionary.
+
+The function must be always defined even if no processing of data is needed.
+In that situation the function just returns `data` parameter.
+
+The following example shows callback for template `mount_option`:
```
ggbecker:
```suggestion
The following example shows the callback function for template `mount_option`:
```

PR #4829, opened 2019-09-12
Unified OVAL sysctl templates and outputs to one file.

matejak:
Formerly, three templates (static, runtime, plain) were instantiated
to three OVAL files. This code generates one output that contains contents
of all three kinds, and it generates that from one file.

This PR therefore makes sure that there is one source template ~and one output file per templated rule~ and either

- the same output files as from before (except missing/added newlines), or
- only one output file that contains everything (see the `ONE_FILE_OVAL_OUTPUT` var).

Like that, the redesign can be verified against previous code, and if it is found OK, the one file output can be switched on.


shared/templates/create_sysctl.py
```diff
@@ -101,29 +96,32 @@ def generate(self, target, serviceinfo):
...
-                for sysctlfile, prefix in self.get_files_var_for_id(sysctl_var_id).items():
+                for stem, ttype in self.get_files_for_id(sysctl_var_id).items():
                     self.file_from_template(
-                        sysctlfile,
+                        "template_OVAL_sysctl",
                         {
+                            "FLAGS": ttype,
                             "SYSCTLID":  sysctl_var_id,
                             "SYSCTLVAR": sysctl_var,
+                            "SYSCTLVAL": "",
```
yuumasato: Now, the difference between both calls `self.file_from_template()` seems to be only  on this `SYSCTLVAL` assignment, either `""` or `sysctl_val`.
The `if` can wrap only an assignment to `SYSCTLVAL` instead of whole `file_from_template()`

shared/templates/create_sysctl.py
```diff
@@ -3,34 +3,29 @@
...
-            'template_OVAL_sysctl_static_var': 'sysctl_static_',
-            'template_OVAL_sysctl_runtime_var': 'sysctl_runtime_',
-            template_name: 'sysctl_'
-        }
+            template_type = 'P'  # P = plain
 
-    def get_files_for_id(self, var_id):
-        if self.is_ipv6_id(var_id):
-            template_name = 'template_OVAL_sysctl_ipv6'
+        if ONE_FILE_OVAL_OUTPUT:
```
yuumasato: Is this for purposes of having an easy switch back to multiple outputs? But still using one template file?

matejak: Exactly. I have updated the PR description.

shared/templates/create_sysctl.py
matejak: Yes, fixed.

matejak:
I have built content from master, moved the build dir and then compared contents of the `build/checks` directory. As there are some newline mismatches, one needs to filter those out.
So I have used the 

```
diff -r build{,-master}/<product>/checks/ | grep '[<>]\s*\w\+'
```

command to watch out for regressions.

yuumasato:
I'm not a fan of how complicate the OVAL template for sysctl will become.
@jan-cerny @matejak is it worth it?

matejak:
This PR is optimized towards the output and ability to switch, i.e. it behaves the same way as the previous three template files, and it is easy to modify the script to generate one-file output as well.
Once RHEL6 content is dropped and the one-file output will be set up as the only alternative, the jinja code will become much more simple.

yuumasato:
@matejak  It is probably better to have it simpler for content writer. Only one template for `sysctl`.
Ack!

PR #4832, opened 2019-09-12
Postfix network listening to loopback-only

70k10:
#### Description:

Add remediation and update the OVAL checks to support loopback-only inet_interface line.

#### Rationale:

This is a CIS requirement. Also, still accepting localhost for RHEL6 per the RHEL6 STIG so we support both that and CIS for RHEL6/7/8.


70k10:
@redhatrises Updated the check/remediation with a new variable. I think it is right, but scap-workbench was acting strange so it was hard to tell if the STIG applied the variable properly. I did see my changes in the ds-1.2.xml though.

linux_os/guide/services/mail/postfix_client/postfix_network_listening_disabled/oval/shared.xml
```diff
@@ -22,12 +22,14 @@
...
-    <ind:pattern operation="pattern match">^[\s]*inet_interfaces[\s]*=[\s]*(localhost|loopback-only)[\s]*$</ind:pattern>
-    {{% else %}}
-    <ind:pattern operation="pattern match">^[\s]*inet_interfaces[\s]*=[\s]*loopback-only[\s]*$</ind:pattern>
-    {{% endif %}}
+    <ind:pattern operation="pattern match">^[\s]*inet_interfaces[\s]*=[\s]*(.*)[\s]*$</ind:pattern>
     <ind:instance datatype="int">1</ind:instance>
   </ind:textfilecontent54_object>
 
+  <ind:textfilecontent54_state id="state_postfix_network_listening_disabled" version="1">
+    <ind:subexpression operation="equals" var_ref="var_postfix_inet_interfaces" />
```
redhatrises: Probably should add `datatype="string"` here.

redhatrises:
> @redhatrises Updated the check/remediation with a new variable. I think it is right, but scap-workbench was acting strange so it was hard to tell if the RHEL6 STIG applied the variable properly. I did see my changes in the ds-1.2.xml though.

@70k10 IIRC you need to add the datatype to the state. Also, looks like there are conflicts with `cce-redhat-avail.txt`.

70k10:
@redhatrises I changed the state  and fixed up the CCEs. Wasn't exactly sure if the datatype was necessary there as I found another example where that was omitted.

redhatrises:
LGTM

PR #4835, opened 2019-09-13
Updated the rule description of configure_fapolicyd_mounts

vojtapolasek:
#### Description:

Due to limitations in Fapolicyd, it is problematic to correctly work with dynamically created user mountpoints. Therefore a warning is added to the rule.


linux_os/guide/services/fapolicyd/configure_fapolicyd_mounts/rule.yml
```diff
@@ -8,6 +8,13 @@ description: |-
     Configure the <tt>fapolicyd</tt> (File Access Policy Daemon)
     to watch on all system filesystem mountpoints
     in <tt>/etc/fapolicyd/fapolicyd.mounts</tt>.
+An initial configuration may be generated by running
+the following command:
+<pre>mount | egrep '^tmpfs| ext4| ext3| xfs' | awk '{ printf "%s\n", $3 }' | egrep
+-v '^\/run$|^\/sys' > /etc/fapolicyd/fapolicyd.mounts</pre>
+Any mountpoints not listed in /etc/fapolicyd/fapolicyd.mounts will not be
+watched.
+
```
yuumasato: Maintain the indentation, I think it will fail the build.

redhatrises: Should the last sentence be an actual XCCDF warning?

yuumasato: I think warnings are more relate to issues caused by complying with the rule.
Not that if configure it incorrectly, it won't work.

redhatrises: I think that a warning type of `general` or `functionality` actually apply here.

yuumasato: I don't think so. Does a warning saying that ciphers no listed in SSH configuration will not be used?
This rule is about adding mountpoints to be watched, if you don't add them, they won't be watched.
The sentence itself is a bit redundant, and seems ok to be in description, but too much to be in separate warning note.

PR #4837, opened 2019-09-13
Skip content generation when callback isn't implemented

jan-cerny:
#### Description:
Skip content generation when callback isn't implemented

#### Rationale:
Currently, if the callback is not present the data are passed directly
to Jinja which sometimes lead to generating invalid OVAL and build
cannot be finished. This change will enable us to add the callbacks one
by one in multiple separate pull requests and we will be able to
complete the build using the new templating system without errors.

Blocks #4820


yuumasato:
Looks good to me.

PR #4840, opened 2019-09-13
Convert mount_option template family to the new system.

matejak:
Convert `mount_option` template family to the new system.
    
- Renamed `*mount_option_var templates` to `*mount_option_removable_partitions` as they have not been used anywhere else.
- OVALs and remediations present in the old system are same in the new one, byte-by-byte.
- New remediations are generated for these `rhel7` rules (I have not looked into another products):
```
 mount_option_boot_nodev
 mount_option_boot_nosuid
 mount_option_var_log_audit_nodev
 mount_option_var_log_audit_noexec
 mount_option_var_log_audit_nosuid
 mount_option_var_log_nodev
 mount_option_var_log_noexec
 mount_option_var_log_nosuid
 mount_option_var_nodev
 mount_option_var_tmp_bind
```


ssg/templates.py
```diff
@@ -19,6 +19,29 @@
...
+    if lang != "oval":
+        data["mountoption"] = re.sub(" ", ",", data["mountoption"])
+    else:
+        data["pointid"] = re.sub(r"[-\./]", "_", data["mountpoint"]).lstrip("_")
+    return data
+
+
+def mount_option_special(data, lang):
+    data = mount_option(data, lang)
+    if lang != "oval":
```
yuumasato: I think it should be `lang == "oval"`. `pointid` is only used by OVAL template.

ssg/templates.py
```diff
@@ -19,6 +19,29 @@
...
+        data["pointid"] = re.sub(r"^var_(.*)", r"\1s", data["mountpoint"])
+    return data
+
+
+def mount_option_var(data, lang):
```
yuumasato: This function is unused, is it expected?

ssg/templates.py
matejak: You are correct, but it turns out that it makes no difference, so I have removed that one as well.

ssg/templates.py
matejak: Yes, I remove it.

matejak:
I have cleaned up `ssg/templates.py` further.

yuumasato:
I'll only be able to review this tomorrow...

jan-cerny:
@yuumasato I have resolved the conflict and rebased on the top of upstream master because @matejak isn't available.

jan-cerny:
Using `compare_ds.py` on RHEL8 datastream I have checked it doesn't change content in the old system and generates templated content in the new system at the same time. ACK,

yuumasato:
@jan-cerny thanks

PR #4841, opened 2019-09-13
Made templates product-specific.

matejak:
Template variables and backends can be product-qualified.
Value of foo@rhel7 overwrites foo on rhel7, while value of foo will be the default
for other products.

Needed for #4820 to work properly.

### DIFF
--- a/ssg/build_yaml.py
+++ b/ssg/build_yaml.py
@@ -747,6 +747,7 @@ class Rule(object):
     def normalize(self, product):
         try:
             self.make_refs_and_identifiers_product_specific(product)
+            self.make_template_product_specific(product)
         except Exception as exc:
             msg = (
                 "Error normalizing '{rule}': {msg}"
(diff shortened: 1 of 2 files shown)


ssg/build_yaml.py
```diff
@@ -766,6 +767,22 @@ def _get_product_only_references(self):
...
+
+        not_specific_vars = self.template.get("vars", dict())
+        specific_vars = self._make_items_product_specific(
+            not_specific_vars, product_suffix, True)
+        self.template["vars"] = specific_vars
+
+        not_specific_backends = self.template.get("backends", dict())
+        specific_backends = self._make_items_product_specific(
+            not_specific_backends, product_suffix, True)
+        self.template["backends"] = specific_backends
```
jan-cerny: Note to myself: This means that "backends" will be always defined, so i we have some time we can drop the else branch or update the code in https://github.com/ComplianceAsCode/content/blob/master/ssg/templates.py#L152

PR #4843, opened 2019-09-17
Ported the sysctl macros to the new system.

matejak:
- Extended Bash and Ansible templates to handle both variable and value cases.
- Removed the *_var templates for Ansible and Bash, as they were made obsolete by the former action.
- Converted all templates in rule.yml to sysctl.
- Added a processing callback for sysctl, cleaned the callback list up.


jan-cerny:
Thank you, I have checked a couple of rules and they generate the same content with the old system as before and they also generate the same content in the new system.

PR #4848, opened 2019-09-18
Migrate OCP runtime check template to a Jinja macro

jan-cerny:
#### Description:
Migrate OCP runtime check template to a Jinja macro

#### Rationale:
The template template_OVAL_ocp_service_runtime_config generates OVAL
content which isn't owned by any rule, but generated definitions are
referenced as extend_definition in some rules. That means this template
is an exception which goes against the logic of the new templating
system which is supposed to generate the whole OVAL for a given rule (is
rule-based). It feels more natural to use Jinja macros than templates
for this way of reusing OVAL code. If we convert this to Jinja macros we
can keep the new templating system simple and straightforward.
Otherwise, we would have to extend it to support generating OVALs with a
custom name.


jan-cerny:
The problem here is that the build system currently does not expect more than 1 OVAL definition in a single shorthand file.

jan-cerny:
I have rebased it on the top of the upstream master

jan-cerny:
I have rebased it on the top of the upstream master

PR #4850, opened 2019-09-18
Add require_emergency_target_auth and update require_singleuser_auth

70k10:
#### Description:

Added a new check for the Systemd emergency.target/service authentication requirement.
Updated the checks and remediations for single-user authentication.

#### Rationale:
Changes made to assist verification and comply with the CIS.
I had them both merged into the require_singleuser_auth, but it got fairly large and I figured someone might want to enable/disable them separately.

Note: I did not add a CCE for the new check because I'm assuming since these checks are two halves of the same requirement they can share. Maybe not?


redhatrises:
> #### Description:
> 
> Added a new check for the Systemd emergency.target/service authentication requirement.
> Updated the checks and remediations for single-user authentication.

Since we are looking at a different target talking about a different configuration, we should create a new rule.

> #### Rationale:
> 
> Changes made to assist verification and comply with the CIS.
> I had them both merged into the require_singleuser_auth, but it got fairly large and I figured someone might want to enable/disable them separately.
> 
> Note: I did not add a CCE for the new check because I'm assuming since these checks are two halves of the same requirement they can share. Maybe not?

CCE's are for a global identifier for the lifetime of a rule for an OS/version, so needs a new CCE?

70k10:
@redhatrises Maybe I was unclear, I did add a new rule for the new check on the emergency target piece.
There are also some check updates to the require_single_user check.
I have modified the new check to have different CCEs.

redhatrises:
@70k10 thanks! One more thing, there is a merge conflict as some of the CCEs now exist in other rules.

70k10:
@redhatrises Fixed the conflict and pushed again.

redhatrises:
Great thanks!

PR #4851, opened 2019-09-19
Fix sourcing of shared functions in test scenarios for gui_login_banner group

matusmarhefka:
#### Description:

Test scenarios in `linux_os/guide/system/accounts/accounts-banners/gui_login_banner`
have been fixed to source shared functions from the variable `$SHARED` exported by
the SSG Test Suite. Also some test scenarios profile selection has been updated
from the `ospp` to the `ncp` as the original `ospp` profile has been renamed to `ncp`.

### DIFF
--- a/linux_os/guide/system/accounts/accounts-banners/gui_login_banner/dconf_gnome_banner_enabled/tests/correct_value.pass.sh
+++ b/linux_os/guide/system/accounts/accounts-banners/gui_login_banner/dconf_gnome_banner_enabled/tests/correct_value.pass.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
-# profiles = xccdf_org.ssgproject.content_profile_ospp
+# profiles = xccdf_org.ssgproject.content_profile_ncp
 
-. ../../../../group_software/group_gnome/dconf_test_functions.sh
+source $SHARED/dconf_test_functions.sh
 
 install_dconf_and_gdm_if_needed
 
--- a/linux_os/guide/system/accounts/accounts-banners/gui_login_banner/dconf_gnome_banner_enabled/tests/wrong_value.fail.sh
+++ b/linux_os/guide/system/accounts/accounts-banners/gui_login_banner/dconf_gnome_banner_enabled/tests/wrong_value.fail.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
-# profiles = xccdf_org.ssgproject.content_profile_ospp
+# profiles = xccdf_org.ssgproject.content_profile_ncp
 
-. ../../../../group_software/group_gnome/dconf_test_functions.sh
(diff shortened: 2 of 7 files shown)


linux_os/guide/system/accounts/accounts-banners/gui_login_banner/dconf_gnome_banner_enabled/tests/correct_value.pass.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 # profiles = xccdf_org.ssgproject.content_profile_ospp
```
vojtapolasek: This rule appears also in other profiles, in particular rhel7 c2s,stig,ncp

linux_os/guide/system/accounts/accounts-banners/gui_login_banner/dconf_gnome_banner_enabled/tests/wrong_value.fail.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 # profiles = xccdf_org.ssgproject.content_profile_ospp
```
vojtapolasek: again this rule appears in stig, c2s, ncp profile in rhel7

linux_os/guide/system/accounts/accounts-banners/gui_login_banner/dconf_gnome_login_banner_text/tests/correct_value.pass.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
-# profiles = xccdf_org.ssgproject.content_profile_ospp
+# profiles = xccdf_org.ssgproject.content_profile_ncp
```
vojtapolasek: this rule appears also in other profiles, OSPP for OL8 and Fedora, C2S for RH7

matusmarhefka:
@vojtapolasek So for the rule `dconf_gnome_login_banner_text`, it is setting the banner text based on the `login_banner_text` variable which selects either `usgcb_default` or `dod_banners`, both of these value selections are covered by the test scenarios (first by test scenarios specifying `ncp` in profiles, second by test scenarios specifying `stig` in profiles). By adding more profiles to the test scenarios we would be testing basically the same as all the profiles are selecting either `usgcb_default` or `dod_banners` (therefore, we usually only list one profile in a test scenario which represents the group of profiles which are the same regarding to selected rule and/or value of variable).

I forgot to update `dconf_gnome_banner_enabled` rule test scenarios because they are also specifying `ospp` profile, so I have updated them to specify `ncp`.

vojtapolasek:
test this please

PR #4853, opened 2019-09-20
fips rules passing for rhel8

vojtapolasek:
#### Description:

added rhel8 to oval check of the rule installed_OS_is_FIPS_certified
changed warning for rules from fips group

### DIFF
--- a/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers/oval/shared.xml
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers/oval/shared.xml
@@ -1,1 +1,45 @@
-{{{ oval_sshd_config(parameter="Ciphers", value="((aes128-ctr|aes192-ctr|aes256-ctr|aes128-cbc|aes192-cbc|aes256-cbc|3des-cbc|rijndael-cbc@lysator\.liu\.se),?)+") }}}
+<def-group>
+  <definition class="compliance" id="sshd_use_approved_ciphers" version="1">
+    <metadata>
+      <title>Use Only Approved Ciphers</title>
+      <affected family="unix">
+        <platform>multi_platform_wrlinux</platform>
+        <platform>Red Hat Enterprise Linux 6</platform>
+        <platform>Red Hat Enterprise Linux 7</platform>
+        <platform>multi_platform_rhv</platform>
+        <platform>Oracle Linux 7</platform>
+      </affected>
+      <description>Limit the ciphers to those which are FIPS-approved.</description>
+    </metadata>
+    <criteria operator="AND">
+      <extend_definition comment="Installed OS is FIPS certified" definition_ref="installed_OS_is_FIPS_certified" />
+      <criteria comment="SSH is configured correctly or is not installed"
+      operator="OR">
+        <criteria comment="sshd is not installed" operator="AND">
+          <extend_definition comment="sshd is not required or requirement is unset"
+          definition_ref="sshd_not_required_or_unset" />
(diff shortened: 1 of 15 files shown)


redhatrises:
@vojtapolasek please add/update the banner for the following rules:

- sshd_use_approved_macs
- sshd_use_approved_ciphers
- configure_crypto_policy
- installed_OS_is_FIPS_certified
- aide_use_fips_hashes

yuumasato:
@vojtapolasek I apologize. on Friday, I might have miss guided you a bit.
I think it makes sense to have all warning updated and synchronized.

@redhatrises As he will be out for a while, I'll address the issues.
Hope you don't mind, Vojta.

yuumasato:
@redhatrises FIPS regulatory warnings updated.

I found an issue with check for `sshd_use_approved_ciphers`, it used the macro `oval_sshd_config`, but it doesn't extend `installed_OS_is_FIPS_certified`.
I hope it is okay to piggyback the fix in this PR.

yuumasato:
@redhatrises What do you think of a macro for these FIPS regulatory warnings? Something like https://github.com/yuumasato/scap-security-guide/commit/d36f907eb2824f129ef2bed27ffdf961d38f1b7d

I think you mentioned that checks for FIPS rules should be explicit, that they cannot be templated. Would this restriction apply to the rule text as well?
These warning are currently repeated 12 times in the repo.

redhatrises:
> @redhatrises What do you think of a macro for these FIPS regulatory warnings? Something like [yuumasato@d36f907](https://github.com/yuumasato/scap-security-guide/commit/d36f907eb2824f129ef2bed27ffdf961d38f1b7d)
> 
> I think you mentioned that checks for FIPS rules should be explicit, that they cannot be templated. Would this restriction apply to the rule text as well?
> These warning are currently repeated 12 times in the repo.

Personally, not a fan of macros for duplicated word text as we are always going to have duplicated content, and there is such a thing as over-macro'ing something. However if you choose to macro, only macro the value and not the yaml key.

yuumasato:
> Personally, not a fan of macros for duplicated word text as we are always going to have duplicated content, and there is such a thing as over-macro'ing something. 

In this case, it would make it very easy for content developer to update the warning message.
But it can be tricky for whoever is reading the `rule.yml` with a macro to understand what it does.

> However if you choose to macro, only macro the value and not the yaml key.

The macro in https://github.com/yuumasato/scap-security-guide/commit/d36f907eb2824f129ef2bed27ffdf961d38f1b7d is just adding a value for `regulatory` warning.

We have changed the FIPS warning so frequently that I though it would deserve a macro. But I also think that this warning needs to be as visible as possible.
So I won't macro it... now. Maybe next a change is needed.

redhatrises:
> > Personally, not a fan of macros for duplicated word text as we are always going to have duplicated content, and there is such a thing as over-macro'ing something.
> 
> In this case, it would make it very easy for content developer to update the warning message.
> But it can be tricky for whoever is reading the `rule.yml` with a macro to understand what it does.
> 
> > However if you choose to macro, only macro the value and not the yaml key.
> 
> The macro in [yuumasato@d36f907](https://github.com/yuumasato/scap-security-guide/commit/d36f907eb2824f129ef2bed27ffdf961d38f1b7d) is just adding a value for `regulatory` warning.
> 
> We have changed the FIPS warning so frequently that I though it would deserve a macro. But I also think that this warning needs to be as visible as possible.
> So I won't macro it... now. Maybe next time a change is needed.

Good point. Although, now I don't see us changing to really at all. (I know.... famous last words.)

redhatrises:
LGTM

PR #4854, opened 2019-09-20
Add sshd_lineinfile and auditd_lineinfile to new templating system

jan-cerny:
#### Description:
This PR adds templates sshd_lineinfile and auditd_lineinfe to new templating system.

First, I had to fix wrong template names in some `rule.yml`s. Some rules that use templates have a wrong template name, which was probably a mistake during the mass conversion of CSVs.

Second, when we process the Jinja macros in OVALs or remediations which are defined in rule directories we add variables (eg. rule title, platforms) to the Jinja environment. The Jinja macros that are used in templates for auditd_lineinfile and templates for sshd_lineinfile require that these variables are defined. If they aren't defined they silently generate an invalid OVAL which is then dropped later in the build process.  Therefore we need to define these variables also in the env_yaml that is used in the templating engine to get a similar environment for generating content from templates as when processing not templated content.  This is similar to code in checks() in
ssg/build_ovals.py. Accidentaly, it fixes the weird task names in Ansible remediations for the affected rules.

Third, we had to add a rule which generates OVAL from template for check if OpenSSH isn't present. The rule package_openssh_removed is for SUSE, the rule
package_openssh-server_removed is for other operating systems.

The new callbacks in `templates.py` are very simple, but I don't like that they are working around the fact that we don't support boolean values in YAML.

~Needs to be rebased on master after #4820 is merged.~

#### Rationale:

Part of the "new templating system" effort.

### DIFF
--- a/linux_os/guide/services/ssh/package_openssh-server_installed/rule.yml
+++ b/linux_os/guide/services/ssh/package_openssh-server_installed/rule.yml
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: wrlinux1019,rhel7,rhel8,ol7,ol8,rhv4
+prodtype: debian8,fedora,ol7,ol8,rhel6,rhel7,rhel8,rhv4,ubuntu1404,ubuntu1604,ubuntu1804,wrlinux8,wrlinux1019
 
 title: 'Install the OpenSSH Server Package'
 
--- a/linux_os/guide/services/ssh/package_openssh-server_removed/rule.yml
+++ b/linux_os/guide/services/ssh/package_openssh-server_removed/rule.yml
@@ -15,7 +15,7 @@ rationale: |-
 
 severity: medium
 
-ocil_clause: 'the package is removed'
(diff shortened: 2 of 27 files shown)


jan-cerny:
I have rebased it on the top of upstream master.

jan-cerny:
> What is the limiting factor? Is is just that the data was migrated with "true" instead of True?

No, the problem isn't with the data migration. The problem is that we process YAML boolean values in a non-standard way, we don't convert them to booleans internally but we treat them as strings. It's here:

https://github.com/ComplianceAsCode/content/blob/75e0b2bf89102bf4a725a0e7935cb47997311895/ssg/yaml.py#L26-L27

It was introduced by https://github.com/ComplianceAsCode/content/pull/2737

yuumasato:
@jan-cerny Thank you for the port. Generated content looks good.

> No, the problem isn't with the data migration. The problem is that we process YAML boolean values in a non-standard way, we don't convert them to booleans internally but we treat them as strings. It's here:

I see, will give it some thought.

PR #4855, opened 2019-09-20
Fix regexes in template_oval_service_disabled and template_oval_service_enabled

vojtapolasek:
#### Description:

The regex used to find the unit files in the above mentioned template was lacking the ^ character, I added it.

I also created tests for the service_atd_disabled rule to verify it.

#### Rationale:

therefore in some cases it matched incorrect services as seen here:

https://bugzilla.redhat.com/show_bug.cgi?id=1662627

### DIFF
--- a/linux_os/guide/services/cron_and_at/service_atd_disabled/tests/service_disabled.pass.sh
+++ b/linux_os/guide/services/cron_and_at/service_atd_disabled/tests/service_disabled.pass.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+#
+# profiles = xccdf_org.ssgproject.content_profile_ospp,xccdf_org.ssgproject.content_profile_standard
+
+yum -y install at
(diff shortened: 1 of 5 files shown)


yuumasato:
@vojtapolasek I think similar fix can be made for `services_enabled` template.

linux_os/guide/services/cron_and_at/service_atd_disabled/tests/service_disabled.pass.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+#
+# profiles = xccdf_org.ssgproject.content_profile_ospp,xccdf_org.ssgproject.content_profile_standard
```
yuumasato: Rule `service_atd_disabled` is not selected in any `ospp` profile, why have it here?

PR #4857, opened 2019-09-20
Enable file permissions templates in new templating system

jan-cerny:
Ports `shared/templates/create_permissions.py`. This script populated
multiple templates at once, generating content for up to 3 rules.  In
the new system there is a different approach - the `rule.yml`s contain
specifically in their template `name` key the type of check to be
generated: file_owner, file_groupowner, file_permissions.

Files template_BASH_permissions and template_OVAL_permissions are never
used and no content was generated from them, so we can remove them.

In the old system, there were special Bash and Ansible templates used
(template_ANSIBLE_file_regex_permissions and
template_BASH_file_regex_permissions) if the file name was specified
using a regular expression. However, OVAL template was shared, meaning
there was no special OVAL template for regular expressions.
Unfortunately, the new system doesn't allow sharing templates between
multiple templates. This problem is solved by creating a new special
OVAL template for template file_regex_permissions.

To disambiguate filepath, path, and filename among these templates the
variable `filepath` has been renamed to `path` in template
`file_regex_permissions`, because it is in line with name of OVAL
`<unix:path>` element which also specifies directory paths. This way
we prevent people confusing it with OVAL `<unix:filepath>` element,
which specifies the full path instead.

### DIFF
--- a/applications/openshift/ocp-permissions/ocp-files/file_permissions_master_cni_conf/rule.yml
+++ b/applications/openshift/ocp-permissions/ocp-files/file_permissions_master_cni_conf/rule.yml
@@ -29,6 +29,6 @@ ocil: |-
 template:
     name: file_regex_permissions
     vars:
-        filemode: '0644'
+        path: /etc/cni/net.d
         filename: ^.*$
-        filepath: /etc/cni/net.d
+        filemode: '0644'
--- a/linux_os/guide/services/http/securing_httpd/httpd_configure_os_protect_web_server/httpd_restrict_file_dir_access/file_permissions_httpd_server_conf_d_files/rule.yml
+++ b/linux_os/guide/services/http/securing_httpd/httpd_configure_os_protect_web_server/httpd_restrict_file_dir_access/file_permissions_httpd_server_conf_d_files/rule.yml
@@ -31,6 +31,6 @@ ocil: '{{{ ocil_file_permissions(file="/etc/http/conf.d/*", perms="-rw-r-----") 
 template:
     name: file_regex_permissions
     vars:
-        filemode: '0640'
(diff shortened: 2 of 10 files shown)


jan-cerny:
The CI fail seems to be infrastructure problem.

jan-cerny:
@openscap-ci test this please

jan-cerny:
When I was reviewing #4858 I realized that the new template doesn't support checking permissions of directories. I think that's because the template has "filepath" key which merges the directory and basename part of the path. We now use it populate the `filepath`element in OVAL `file_object`. But this element can't be used to describe directories (see OVAL Unix file_object specification).  To describe directory you have to use `path` element instead and set `xsi:nil` attribute of `filename` element to true. The template or the callback should do that. However, we can't determine if a full path describes a directory or a file.

There are multiple possible solutions to this problem:

1. Split the `filepath` key in `rule.yml`s to 2 new YAML keys: `path` and `filename`, empty filename would mean that` path` describes a directory. The callbacks would merge the 2 when needed.

2. Keep `filepath` key, but add a new key `is_directory`, with implicit value of `false`. When set to `true` this key will cause generating OVAL check for directories.

3. Add some logic that will ensure if the value of `filepath` key ends with a `/` the OVAL check for directories would be generated


Thoughts? Suggestions?

yuumasato:
> 3. Add some logic that will ensure if the value of `filepath` key ends with a `/` the OVAL check for directories would be generated

This seems the most intuitive for me.

jan-cerny:
@yuumasato I have added the option 3 in b45c26f

yuumasato:
LGTM, thank you @jan-cerny

PR #4862, opened 2019-09-23
Fix oval applicability for wrlinux and ubuntu

yuumasato:
#### Description:

- This fixes an issue where WRLinux and Ubuntu content would not contain some OVAL, although applicable.
  - Update static content's affected platform with correct product's full name
  - Correct mapping `wrlinux` product to WRLinux product stem name
  - Correct mapping of `ubuntu`'s versions to Ubuntu's full name version

#### Rationale:
- ` _check_is_applicable_for_product()`, used by `build-scripts/combine_ovals.py`, was not able to match WRLinux and Ubuntu OVAL affected platform value to their full names.
- The OVAL affected platform platform value should match the product's full name.


ssg/utils.py
```diff
@@ -90,6 +90,10 @@ def parse_name(product):
     if match:
         _product = match.group(1)
         _product_version = match.group(2)
+        # Ubuntu versions have a dot in between the numbers
+        # While the prodtype doesn't have the dot, the full product name does
+        if _product == "ubuntu":
+            _product_version = _product_version[:2] + "." + _product_version[2:]
```
yuumasato: As `_parse_name()` is only used by `_check_is_applicable_for_product()`, I decided to update Ubuntu's version here.
But design-wise, it may be better to do it after call for `_parse_name()`:
https://github.com/ComplianceAsCode/content/blob/0e5705bedb3ccb3e6840a862598260ce42dae38e/ssg/build_ovals.py#L27

Thoughts?

yuumasato: I've decided to move the Ubuntu version handling out of `_parse_name()` and into `_check_is_applicable_for_product()`.

I think it is cleaner this way.

shared/templates/csv/packages_installed.csv
```diff
@@ -7,6 +7,7 @@ firewalld
 iptables
 libcap-ng-utils
 openscap-scanner
+openssh-server
```
jan-cerny: Do you think it can make a problem on SUSE systems where it's just `openssh`?

linux_os/guide/system/logging/ensure_rsyslog_log_file_configuration/rsyslog_files_permissions/oval/shared.xml
```diff
@@ -6,9 +6,9 @@
         <platform>multi_platform_rhel</platform>
         <platform>multi_platform_fedora</platform>
         <platform>Debian 8</platform>
-        <platform>Ubuntu 1404</platform>
-        <platform>Ubuntu 1604</platform>
-	<platform>Ubuntu 1804</platform>
+        <platform>Ubuntu 14.04</platform>
```
jan-cerny: I still can grep `Ubuntu 1404` without the dot in shared/templates/template_OVAL_service_disabled and elsewhere.

shared/templates/csv/packages_installed.csv
jan-cerny: I think it shouldn't make a problem because there are 2 different rules - first for SUSE and second for other operating systems.

yuumasato: It would bloat up SUSE's content with useless check for `openssh-server` package.
I have moved the entry to `wrlinux`'es CSV files to avoid bloating up content.

linux_os/guide/system/logging/ensure_rsyslog_log_file_configuration/rsyslog_files_permissions/oval/shared.xml
yuumasato: Fixed.

PR #4865, opened 2019-09-23
Add coreos NCP profile

redhatrises:
#### Description:

- Adds new CoreOS draft ncp profile. Everything in the profile is subject to change, and the profile is
  purposefully not buildable for now.

### DIFF
--- a/ocp4/profiles/coreos-ncp.profile
+++ b/ocp4/profiles/coreos-ncp.profile
@@ -1,0 +1,488 @@
+documentation_complete: true
+
+title: 'NIST National Checklist for Red Hat Enterprise Linux CoreOS'
+
+description: |-
+    This compliance profile reflects the core set of security
+    related configuration settings for deployment of Red Hat Enterprise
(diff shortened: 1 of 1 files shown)


coreos/profiles/ncp.profile
```diff
@@ -0,0 +1,490 @@
+documentation_complete: true
+
+title: 'NIST National Checklist Program Security Guide'
```
shawndwells:
```suggestion
title: 'NIST National Checklist for Red Hat Enterprise Linux CoreOS'
```

shawndwells: Ideally, a version number or range (e.g. 8.x) should be included.

redhatrises: Versioning will probably change so much that this would be an exercise in futility.

shawndwells: Yeah, you're right, and I see they've been removed for other profiles. Looks good to me!

PR #4868, opened 2019-09-24
Use defaults/main.yml and not vars/main.yml

redhatrises:
Ansible roles prefer `defaults/main.yml` over `vars/main.yml`


shawndwells:
Matches the language in Ansible training. Ack.

PR #4870, opened 2019-09-25
Use correct function to handle wrong args in build_product.

ggbecker:
#### Description:

- Fix build product tool when using nonexistent product names

#### Rationale:

- The function's name is `handle_wrong_products` instead of `handle_wrong_argument` therefore there was a syntax error.


yuumasato:
Nice catch!

PR #4871, opened 2019-09-25
Fix typo in installed_OS_is_rhv4.xml file.

ggbecker:
#### Description:

- Fix typo in installed_OS_is_rhv4.xml file.


redhatrises:
wow... that's a great catch! thanks for the fix.

PR #4874, opened 2019-09-27
Add rules to generate OVAL dependencies

jan-cerny:
#### Description:

Add new rules that will not be used in any profile for now but they are used to generate OVALs which are used in other rules.

#### Rationale:
Some of the rules depend on content which was generated from CSVs but no
corresponding rule exists. This is not possible in the new templating
system.  Since the new templating system is rule-based, we need a
rule.yml that contains a `template:` key to generate templated content.

### DIFF
--- a/linux_os/guide/services/avahi/package_avahi_installed/rule.yml
+++ b/linux_os/guide/services/avahi/package_avahi_installed/rule.yml
@@ -1,0 +1,24 @@
+documentation_complete: true
+
+title: Install avahi Package
+
+description: |-
+    {{{ describe_package_install(package="avahi") }}}
+
+rationale: |-
+    Avahi is a system which facilitates service discovery on
+    a local network -- this means that you can plug your laptop or
+    computer into a network and instantly be able to view other people who
+    you can chat with, find printers to print to or find files being
+    shared.
+
+severity: medium
+
+ocil_clause: 'the package is not installed'
+
+ocil: '{{{ ocil_package(package="avahi") }}}'
+
+template:
+    name: package_installed
+    vars:
+        pkgname: avahi
--- a/linux_os/guide/services/ldap/openldap_client/package_pam_ldap_removed/rule.yml
+++ b/linux_os/guide/services/ldap/openldap_client/package_pam_ldap_removed/rule.yml
@@ -1,0 +1,22 @@
+documentation_complete: true
+
+prodtype: rhel6
+
+title: Remove pam_ldap Package
+
+description: |-
+    {{{ describe_package_remove(package="pam_ldap") }}}
+
+rationale: |-
+    The pam_ldap module is a Pluggable Authentication Module (PAM) which provides authentication, authorization and password changing against LDAP servers.
+
+severity: medium
+
+ocil_clause: 'the package is installed'
+
+ocil: '{{{ ocil_package(package="pam_ldap") }}}'
+
+template:
+    name: package_removed
+    vars:
+        pkgname: pam_ldap
--- a/linux_os/guide/services/ntp/service_chronyd_enabled/rule.yml
+++ b/linux_os/guide/services/ntp/service_chronyd_enabled/rule.yml
@@ -1,0 +1,17 @@
+documentation_complete: true
+
+title: 'Enable the chronyd Daemon'
+
+description: '{{{ describe_service_enable(service="chronyd") }}}'
+
+rationale: |-
+    chronyd is a daemon for synchronisation of the system clock.
+
+severity: medium
+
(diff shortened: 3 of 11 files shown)


redhatrises:
> #### Description:
> 
> Add new rules that will not be used in any profile for now but they are used to generate OVALs which are used in other rules.
> #### Rationale:
> 
> Some of the rules depend on content which was generated from CSVs but no
> corresponding rule exists. This is not possible in the new templating
> system. Since the new templating system is rule-based, we need a
> rule.yml that contains a `template:` key to generate templated content.

This is a bug in the templating system and will be considered a bug and regression and needs to be fixed.

yuumasato:
> Add new rules that will not be used in any profile for now but they are used to generate OVALs which are used in other rules.

These rules will be used when applicability of rules is better implemented, as these checks are currently mostly working as an applicability check.
Except that the rule return `pass` instead of `notapplicable`.

> This is a bug in the templating system and will be considered a bug and regression and needs to be fixed.

I cannot imagine a place to put the template name and data. 
Adding a second template data for the rules that need these checks will duplicate a lot of code, as many rules rely on one common `package_installed` check.

A simple workaround for now can be to have the checks statically in `shared/checks`.

matejak:
It is incorrect to blame the new templating system - rules in question check for multiple things - often for package installed, service running and service configured. The approach that feels more correct is to only check for correct configuration, and to move the installed + running part into CPE applicability. This approach will be work fine with the new templating system.

Therefore, the solution before this CPE redesign happens will be dirty in one way or another. To sum it up, these possibilities were suggested by the team:

- Go with this PR that creates new rules.
- Use `shared/checks` to ensure that  those OVAL definitions in question exist.
- Extract the code used in templates package_installed, package_removed, service_enabled and service_disabled to Jinja macros, and use it to create self-contained definitions.
- Add a mechanism to the build system that would generate OVALs without a need to have corresponding rules - eg. to have some hardcoded list of "always generated templated content".

@redhatrises What of those options do you like more? I would go for this PR as it is, it is not perfect, but those newly added rules won't clutter anything, they won't mislead anybody, and we can simply remove them after the CPE Applicability Language is implemented.

redhatrises:
> It is incorrect to blame the new templating system: Rules in question check for multiple things - often for package installed, service running and service configured. Approach that feels more correct is to only check for correct configuration, and to move the installed + running part into CPE applicability checks. That approach would work fine with the new templating system.
> 
> Therefore, the solution before this CPE redesign happens will be dirty in one way or another. To sum it up, these possibilities were suggested by the team:
> 
>     * Go with this PR that creates new rules.
> 
>     * Use `shared/checks` to ensure that  those OVAL definitions in question exist.
> 
>     * Extract the code used in templates package_installed, package_removed, service_enabled and service_disabled to Jinja macros, and use it to create self-contained definitions.
> 
>     * Add a mechanism to the build system that would generate OVALs without a need to have corresponding rules - eg. to have some hardcoded list of "always generated templated content".
> 
> 
> @redhatrises What of those options do you like more? I would go for this PR as it is, it is not perfect, but those newly added rules won't clutter anything, they won't mislead anybody, and we can simply remove them after the CPE Applicability Language is implemented.

Templating system needs to handle all of the following scenarios:
1. Be able to generate extended_definitions
2. Be able to use CPE applicability language

yuumasato:
> and we can simply remove them after the CPE Applicability Language is implemented.

Wait, even with CPE applicability language we will need the rules for installed packages. Otherwise which rule will install a package which is need for applicability of another rule?

> Templating system needs to handle all of the following scenarios:
> 
>     1. Be able to generate extended_definitions

My opinion is that a definition is worth making independent when it is reusable by other definitions. Otherwise it should just be a `test` part of the definition.
On the other hand, with this approach definitions can get pretty complex. It is not clear to me where to draw the line.

Rant/example:
For `package_installed` cases, the definition is only extended in services rules because it wants to check whether the package is installed or not (it is effectively doing the applicability check).
But then, when applicability check is implemented, there will not be the need to check whether the package is installed.



>     2. Be able to use CPE applicability language

What does use CPE applicability language in templating system means?
Does it means linking the platform element in the rule(.yml) to the OVAL that verifies applicability?

yuumasato:
> Templating system needs to handle all of the following scenarios:
> 
>     1. Be able to generate extended_definitions

@redhatrises Do you have example of definition that should be independent but doesn't fit into any rule?

redhatrises:
Basically these rules should not be created and keys need to be added to the templating system to generate extended_definitions or CPE. So, this PR is a NACK.

jan-cerny:
I will create an alternative solution.

PR #4878, opened 2019-09-30
Remove extraneous *.* from description and check of rule 'rsyslog_remote_tls_cacert'

yuumasato:
#### Description:

- Rule `rsyslog_remote_tls_cacert`: remove extra `*.*`from rule description and check 
- Add simple test scenario for this

### DIFF
--- a/linux_os/guide/system/logging/rsyslog_sending_messages/rsyslog_remote_tls_cacert/oval/shared.xml
+++ b/linux_os/guide/system/logging/rsyslog_sending_messages/rsyslog_remote_tls_cacert/oval/shared.xml
@@ -22,7 +22,7 @@
 
   <ind:textfilecontent54_object id="obj_rsyslog_remote_tls_cacert" version="1">
     <ind:filepath operation="pattern match">^/etc/rsyslog\.(conf|d/.+\.conf)$</ind:filepath>
-    <ind:pattern operation="pattern match">^\s*global\(DefaultNetstreamDriverCAFile="(.+?)"\) +\*\.\*\s*\n</ind:pattern>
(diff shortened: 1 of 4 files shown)


comps:
I don't know if the `*.*` is valid or not - it might be valid rsyslog syntax, but it's certainly not required for the directive to work, so we should check just for the minimum required string.

yuumasato:
I have the impression that `*.*` was meant for the other rule `rsyslog_remote_tls`
Something like this:
```
 *.* action(type="omfwd" protocol="tcp" Target="&lt;remote system>" port="6514"
        StreamDriver="gtls" StreamDriverMode="1" StreamDriverAuthMode="x509/name")
```
I'll check it out.

comps:
See also the official documentation,
https://www.rsyslog.com/doc/v8-stable/tutorials/tls.html#client-setup
(no `*.*` by default)

yuumasato:
Hmm, maybe `*.*` is not needed at all.

> See also the official documentation,
> https://www.rsyslog.com/doc/v8-stable/tutorials/tls.html#client-setup
> (no `*.*` by default)

@jvymazal Can you advise us on this?

jvymazal:
@yuumasato tanks for notice, it seems that if you use recent enough version and new config, `*.*` is assumed by default if you do not write any selector.

yuumasato:
> @yuumasato thanks for notice, it seems that if you use recent enough version and new config, `*.*` is assumed by default if you do not write any selector.

What would be a recent version number? Is it supported in RHEL8 already?

jvymazal:
RHEL8 definitely, newer RHEL7 version (say 7.6+) as well.

yuumasato:
Thanks for the info!

yuumasato:
So this PR is removing `*.*` from `rsyslog_remote_tls_cacert` and it is not expected to be part of `rsyslog_remote_tls`.

PR #4881, opened 2019-10-01
Add -t parameter for readarray in fix_audit_syscall_rule to remove trailing newline

mildas:
#### Description:
readarray man page:

> -t     Remove a trailing delim (default newline) from each line read.

#### Rationale:
Without `-t` `readarray` left newline after each element from array.


matejak:
This one slipped past #4565, good job nailing this one out!

PR #4882, opened 2019-10-01
Fix dnf timer rule

redhatrises:
- Rule name was inconsistent with the other dnf rules


jan-cerny:
I have built RHEL8 datastream with your patch and the rule doesn't contain OVAL and remediations.

I think that's because the rule ID needs to be in a specific format for templated rules. If you change the rule ID the generated content isn't found by the build system. It's hard-coded in the scripts that process the templates. For example, here: https://github.com/ComplianceAsCode/content/blob/c71298a30ff10e388a599c7e0ed7e23d6d101947/shared/templates/create_timers_enabled.py#L31

By the way, after we finish the new template system we will be able to remove this limitation easily.

yuumasato:
> Rule name was inconsistent with the other dnf rules

But then it becomes inconsistent with other templated rules:
- `package_<package>_installed`
- `service_<service>_enabled`
- `timer_<timer>_enabled`

yuumasato:
I think it makes sense to rename `package_<packagename>_installed` to `<packagename>_package_installed`, then it is consistent with `<timername>_timer_enabled`

But changing rule ids can break tailoring.
If we go for changing IDs, it should be done all at once (within one release cycle), and announced so people are prepared.

redhatrises:
These are valid points from @yuumasato and @jan-cerny and updating rule ids would be very invasive for sure. Thanks for pointing this out guys!

PR #4883, opened 2019-10-02
Fix missing content in datastreams generated by new templating system

jan-cerny:
#### Description:

The missing OVALs were caused mostly by a mismatch in `prodtype` tag in the rule and the rule that generates the dependent content. This is solved by adding the prodtypes from the dependent rule to its dependency. However, a fix for 4 missing OVALs in OCP3  datastream will be submitted in a separate PR.

There was a mistake in port of CSV data for mount options, which caused that Anaconda remediation was disabled on all platforms. But according to CSVs it should be disabled only on RHEL 7  and 8.

For more details please read the commit messages of each commit. 

#### Rationale:

This prevents missing OVALs and Anaconda remediations in comparison of datatstreams generated by old and new templating.

### DIFF
--- a/linux_os/guide/system/permissions/partitions/mount_option_tmp_nodev/rule.yml
+++ b/linux_os/guide/system/permissions/partitions/mount_option_tmp_nodev/rule.yml
@@ -47,4 +47,5 @@ template:
         type@rhel7: tmpfs
         type@rhel8: tmpfs
     backends:
-        anaconda: 'off'
+        anaconda@rhel7: 'off'
+        anaconda@rhel8: 'off'
--- a/linux_os/guide/services/ftp/disabling_vsftpd/package_vsftpd_removed/rule.yml
+++ b/linux_os/guide/services/ftp/disabling_vsftpd/package_vsftpd_removed/rule.yml
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: wrlinux1019,rhel6,rhel7,rhel8,rhv4
+prodtype: fedora,sle12,wrlinux1019,rhel6,rhel7,rhel8,rhv4
 
 title: 'Uninstall vsftpd Package'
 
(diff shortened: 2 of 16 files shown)


matejak:
An alternative view of this is that it updates rule metadata, so they are included in relevant benchmarks.
Thank you for this PR!

linux_os/guide/system/permissions/partitions/mount_option_tmp_nodev/rule.yml
```diff
@@ -47,4 +47,5 @@ template:
         type@rhel7: tmpfs
         type@rhel8: tmpfs
     backends:
-        anaconda: 'off'
+        anaconda@rhel7: 'off'
+        anaconda@rhel8: 'off'
```
yuumasato: Fedora and Oracle also had anaconda remediation for `/dev/shm` mount points disabled.

> To get the same result using both old and new templating system,
we need to change the backends list.

But why do we want same results from old and new templating system?

In this case I think the content is better with anaconda disabled for all products.

The reason this is disabled is because `/dev/shm`doesn't exist at the moment of installation, they are managed by [systemd](https://github.com/lnykryn/systemd-rhel/blob/747f3af70c0ee9089e520ccd8917ded397a5cd2a/src/core/mount-setup.c#L91) and created during first boot.
So any product using systemd should have anaconda remediation disabled.

yuumasato: I couldn't find any resource documenting management of `/dev/shm` by systemd.

matejak: Content equivalence between the old and new systems makes it easy to evaluate the new system's state.

yuumasato: > Content equivalence between the old and new systems makes it easy to evaluate the new system's state.

I understand this reasoning, but I disagree with hindering generation of content with new templating system.

How about upping the CSV so that it generates the same content as the new system would?

PR #4884, opened 2019-10-02
Fix templates file_owner, file_groupowner and merge templates file_permissions and file_regex_permissions

jan-cerny:
#### Description:
During the port of template data to CSVs we have forgotten about 4 rules in OCP3 benchmark which check for owner and group owner of files specified by a wild card. The result was that we haven't port the feature that enabled generating this type of check as we incorrectly thought it isn't used. To avoid regression in OCP3 Benchmark after the new templating system is switched on, we need to reintroduce this feature in templates `file_owner` and `file_groupowner`. During the migration of the data the base name and directory path has been merged. So we propose an alternative way: there will be a new optional variable `file_regex` in the template which can contain a regular expression describing files in a given directory.  If `file_regex` is set the `filepath` must describe a directory, which we recognize by the trailing backslash.

#### Rationale:

Prevents breaking of OCP3 rules if we switch to the new templating system. `file_owner_master_openvswitch`, `file_owner_master_cni_conf`, `file_groupowner_master_openvswitch`, `file_groupowner_master_cni_conf`
The potentially missing OVALs were revealed by `utils/compare_ds.py`.

### DIFF
--- a/applications/openshift/ocp-permissions/ocp-files/file_groupowner_master_cni_conf/rule.yml
+++ b/applications/openshift/ocp-permissions/ocp-files/file_groupowner_master_cni_conf/rule.yml
@@ -23,3 +23,10 @@ references:
 ocil_clause: '{{{ ocil_clause_file_group_owner(file="/etc/cni/net.d/*", group="root") }}}'
 
 ocil: '{{{ ocil_file_group_owner(file="/etc/cni/net.d/*", group="root") }}}'
+
+template:
+    name: file_groupowner
+    vars:
+        filepath: /etc/cni/net.d/
+        file_regex: ^.*$
+        filegid: '0'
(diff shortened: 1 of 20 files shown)


yuumasato:
In #4857 the approach was to change `filepath` to `path` and `filename` contained the regex.

The approach in this PR seems simpler, as there is only one template, and a variables determine if the files to check should be collected via regular expression.

Do you think it makes sense to consolidate the approach to check for permission | owner | groupowner of files collected via regular expression in this PR?
i.e.: Also change `shared/templates/template_OVAL_file_permissions` and get rid of  `shared/templates/template_OVAL_file_regex_permissions`.

jan-cerny:
@yuumasato I agree because that way the templates will be consistent.

jan-cerny:
I have merged templates file_permissions and file_regex_permissions.

matejak:
It is a bit unfortunate that we specify `filepath`, which can be either an absolute filename path, or a directory, which is denoted as `path` in OVAL. However, the trailing slash which is checked in the template preprocessing remediates this trickiness. I can also see that it is an improvement over the previous state, as the terminology was not perfect in template files.

matejak:
Great, the code works for OCP3 and there I have observed no regressions, so LGTM.

PR #4887, opened 2019-10-02
Remove installation of packages from RHEL8 kickstarts as it is handled by rules

matusmarhefka:
#### Description:

- `aide` is handled by the rule `package_aide_installed`
* `libreswan` is handled by the rule `package_libreswan_installed` + it has been removed from the `ospp` profile

#### Rationale:

- Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1751681


redhatrises:
Should this be done with all kickstarts not just rhel8?

redhatrises:
Cool! LGTM

PR #4888, opened 2019-10-02
Update audit_rules_unsuccessful_file_modification regex to match multiple "-S" syscall args

yuumasato:
#### Description:

- Update OVAL definitions of valid and correct audit rule

#### Rationale:
-  After changes in `fix_audit_syscal_rule`, it started generating audit rules, with different syntax:
`-a always,exit -F arch=b32 -S creat -S open -S openat -S open_by_handle_at -S truncate -S ftruncate -F exit=-EACCES -F auid>=1000 -F auid!=unset -F key=access`
For which the OVAL check doesn't recognize as correct.


matusmarhefka:
@yuumasato LGTM, I have tested on RHEL8 and all test scenarios are passing. Btw can you list commits with these changes to the `fix_audit_syscal_rule`? Just for the record.

yuumasato:
@matusmarhefka I don't exactly which changes, but probably #4403, along with #4565 and #4881 changed the behavior of the function, thus resulting audit rules.

matusmarhefka:
Thanks @yuumasato

PR #4897, opened 2019-10-04
Made the new templating system work with Python2.6.

matejak:
Python 2.6 doesn't allow dict comprehensions.
Also the `package_gcc_installed` rule has been removed, as it was introduced for demonstration purposes only and the unquoted YAML bool causes problems on Python2.


ssg/templates.py
```diff
@@ -328,7 +328,10 @@ def preprocess_data(self, template, lang, raw_parameters):
             parameters = raw_parameters.copy()
         # TODO: Remove this right after the variables in templates are renamed
         # to lowercase
-        parameters = {k.upper(): v for k, v in parameters.items()}
+        uppercases = dict()
+        for k, v in parameters.items():
+            uppercases[k.upper()] = v
+        parameters.update(uppercases)
```
jan-cerny: I'd prefer to return `uppercases` right away.

konstruktoid:
Why support python2.6? 
https://pythonclock.org/

jan-cerny:
@konstruktoid Because Red Hat Enterprise Linux 6 has only Python 2.6, and we still support building on RHEL6.

matejak:
I have rebased, so no more updating of the dict with uppercase keys - now only the uppercase dict is returned.

PR #4898, opened 2019-10-04
New templating - generate OVAL definitions without creating rules 

jan-cerny:
#### Description:

This PR introduces a way to generate templated OVALs that aren't "owned" by any rule into the new templating system.

For more details, please see the commit messages of each commit.

#### Rationale:

Some templated OVALs are needed as dependency in other OVALs referenced by extend_definition. The new templating system is based on rules, so to generate an OVAL definition a rule should be created. But for some OVALs it isn't acceptable to have a special rule in the benchmark.

This is a temporary solution until applicability of all rules is handled by using CPE applicability checks.

See #4874

### DIFF
--- a/shared/templates/extra_ovals.yml
+++ b/shared/templates/extra_ovals.yml
@@ -1,0 +1,54 @@
+package_avahi_installed:
+  name: package_installed
+  vars:
+    pkgname: avahi
+
+package_esc_installed:
+  name: package_installed
+  vars:
+    pkgname: esc
+
+package_pam_pkcs11_installed:
+  name: package_installed
+  vars:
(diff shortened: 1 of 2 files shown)


matejak:
Could you please also add those? I have found out that they are missing based on analysis of build logs:

```
package_aide_installed
package_bind_removed
package_net-snmp_removed
package_prelink_removed
package_samba-common_installed
package_tftp-server_removed
package_vsftpd_removed
service_postfix_enabled
```

jan-cerny:
We have rule.yml for all of them expect package_prelink_removed which is actually solved by this PR. The problem is inconsistent `prodtype` which we can solve separately.

PR #4906, opened 2019-10-07
Fixed the remediation for rsyslog_files_permissions

matejak:
- Stripped quotes and brackets from extracted paths.
- Dropped apparent config files from extracted list of logfile paths.

The old remediation shows errors on a fresh RHEL8 install.

### DIFF
--- a/linux_os/guide/system/logging/ensure_rsyslog_log_file_configuration/rsyslog_files_permissions/tests/perms_0600.pass.sh
+++ b/linux_os/guide/system/logging/ensure_rsyslog_log_file_configuration/rsyslog_files_permissions/tests/perms_0600.pass.sh
@@ -9,10 +9,13 @@ source $SHARED/rsyslog_log_utils.sh
 PERMS=0600
 
 # setup test data
-create_rsyslog_test_logs 1
+create_rsyslog_test_logs 4
 
-# setup test log file and permissions
-chmod $PERMS ${RSYSLOG_TEST_LOGS[0]}
+# setup all files with incorrect permission
+chmod 0601 "${RSYSLOG_TEST_LOGS[@]}"
+
+# setup the real logfile with correct permissions
+chmod $PERMS "${RSYSLOG_TEST_LOGS[0]}"
 
 # add rule with 0600 permissions log file
 cat << EOF > $RSYSLOG_CONF
@@ -21,5 +24,13 @@ cat << EOF > $RSYSLOG_CONF
 #### RULES ####
 
 *.*        ${RSYSLOG_TEST_LOGS[0]}
+
+ *.*        ${RSYSLOG_TEST_LOGS[1]}
+
+authpriv.*        /nonexistent_file
+
+# *.*        /irrelevant_file
+
+\$something /irrelevant_file
 
 EOF
--- a/linux_os/guide/system/logging/ensure_rsyslog_log_file_configuration/rsyslog_files_permissions/bash/shared.sh
+++ b/linux_os/guide/system/logging/ensure_rsyslog_log_file_configuration/rsyslog_files_permissions/bash/shared.sh
@@ -16,14 +16,21 @@ do
 	# From each of these files extract just particular log file path(s), thus:
 	# * Ignore lines starting with space (' '), comment ('#"), or variable syntax ('$') characters,
 	# * Ignore empty lines,
+	# * Strip quotes and closing brackets from paths.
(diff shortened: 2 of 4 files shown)


yuumasato:
@matejak Thanks for the fix, would you consider adding test scenarios?

matejak:
Oops, accidental branch removal

matejak:
I have updated some test scenarios, but existing scenarios fail unexpectedly, and it is not trivial to fix them. So we can either merge the PR as-is, or wait a bit until the OVAL is fixed to be more robust.

linux_os/guide/system/logging/ensure_rsyslog_log_file_configuration/rsyslog_files_permissions/tests/perms_0600.pass.sh
```diff
@@ -9,10 +9,13 @@ source $SHARED/rsyslog_log_utils.sh
 PERMS=0600
 
 # setup test data
-create_rsyslog_test_logs 1
+create_rsyslog_test_logs 4
```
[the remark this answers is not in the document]
jan-cerny: Aha, I see that this will be addressed by https://github.com/ComplianceAsCode/content/pull/4379

linux_os/guide/system/logging/ensure_rsyslog_log_file_configuration/rsyslog_files_permissions/tests/perms_0600.pass.sh
```diff
@@ -22,4 +25,12 @@ cat << EOF > $RSYSLOG_CONF
 
 *.*        ${RSYSLOG_TEST_LOGS[0]}
 
+ *.*        ${RSYSLOG_TEST_LOGS[1]}
```
jan-cerny: I think that there is a problem that the OVAL for this rule always returns false if `/etc/rsyslog.conf` doesn't contain any inlcusion. Please add `\$IncludeConfig /etc/rsyslog.d/*.conf` to both test scenarios.

jan-cerny: @matejak this is the problem

yuumasato: > I think that there is a problem that the OVAL for this rule always returns false if `/etc/rsyslog.conf` doesn't contain any inlcusion.

This may be an actual finding in the OVAL check.
If `rsyslog.conf` or `/etc/rsyslog.d/*.conf` without includes are correct, the OVAL check should just pass.

@jvymazal Are includes for `rsyslog` `*.conf` file mandatory?

matejak: After consulting with @jvymazal, `rsystlog` will run even if there is no include directive present.
As an interesting fact, in a default installation, the directive is there, but it doesn't do anything, because the `/etc/rsyslog.d/` directory is empty.

linux_os/guide/system/logging/ensure_rsyslog_log_file_configuration/rsyslog_files_permissions/tests/perms_0600.pass.sh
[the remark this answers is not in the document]
matejak: Yes, that another PR is what I was referring to as waiting for more robust OVAL.
However, this PR fixes errors in remediation, so I propose to get it merged as-is.

jan-cerny:
We have discussed this with @matejak . We think that the remediation fix is basically a display fix. Currently, by looking to the HTML report people might think that the remediation is defective. If the patch is applied, it will not make the rule passing. The report from `oscap xccdf eval --remediate` will still show error, because the scan on RHEL 8 will fail, because OVAL expects IncludeConfig directive in `/etc/rsyslog.conf` and RHEL 8 uses the new syntax with `include()` instead. The remediation fix will only reveal the OVAL is defective.

We think that it isn't a release blocker, and it's not a regression from previous released version of RHEL8  content.

jan-cerny:
On my F30 laptop it doesn't produce the weird messages s. a. `/tmp/oscap.3j86mL/fix-XXJMZJ3X: line 48: [: : integer expression expected` anymore.

PR #4908, opened 2019-10-08
ocp4: Initial build system support for the OCP4 product

evgenyz:
It's just a basic CMake support. I plan to use this PR to create something usable by the Operator before merging.

My concerns about the current content:

1) This criteria defines RHCOS as being equal to RHEL:
```
<criterion comment="Installed operating system is part of the unix family"
      test_ref="test_rhel8_unix_family" />
      <criteria operator="OR">
        <criterion comment="RHEL 8 is installed" test_ref="test_rhel8" />
        <criterion comment="RHEL 8 CoreOS is installed" test_ref="test_rhel8_coreos" />
</criteria>
```
Which is a bit disturbing. The components might be the same, but OS managements is different (ostree-update vs yum etc).

2) I have no idea how to isolate something working out of the profile in place to start building, testing and providing this to the Operator. I'd like to have early integration.


redhatrises:
> It's just a basic CMake support. I plan to use this PR to create something usable by the Operator before merging.
> 
> My concerns about the current content:
> 
>     1. This criteria defines RHCOS as being equal to RHEL:
> 
> 
> ```
> <criterion comment="Installed operating system is part of the unix family"
>       test_ref="test_rhel8_unix_family" />
>       <criteria operator="OR">
>         <criterion comment="RHEL 8 is installed" test_ref="test_rhel8" />
>         <criterion comment="RHEL 8 CoreOS is installed" test_ref="test_rhel8_coreos" />
> </criteria>
> ```
> 
> Which is a bit disturbing. The components might be the same, but OS managements is different (ostree-update vs yum etc).

Don't see a problem here really. There is a difference in selected rules, so the yum rules are going to be replaced with ostree rules.

>     1. I have no idea how to isolate something working out of the profile in place to start building, testing and providing this to the Operator. I'd like to have early integration.

So far the PR looks good, but the `transforms` directory needs to be added. We also need to make the benchmark root able to handle a list of directories.

evgenyz:
So, now this PR is at least build-able. I completed the stub for OCP4 to the build-able state and moved `ocp4/profiles/coreos-ncp.profile => rhel8/profiles/ncp.profile` because *coreos* is *rhel8* (as stated above) and you can't use rules that are designed for *rhel8* in *ocp4* product. The *ocp4* product right now is for upper-level "applications configuration" checks, as it was with *ocp3*. I did it only to be able to build the branch with minimal changes.

Now, @redhatrises you said that you would like to "make the benchmark root able to handle a list of directories". I believe that you mentioned it because you don't want to move `coreos-ncp.profile` to the *rhel8* sub-folder, and you want to have all checks related to the CoreOS inside the *ocp4* product, right? But if I get it correctly it is not only about multiple roots. There are rules in the `coreos-ncp.profile` that belongs to *rhel8*, like *chronyd_client_only*, which is `prodtype: rhel8,fedora,ol8`. Does it make any sense to add *ocp4* to all these rules (and checks)?

jan-cerny:
@evgenyz @redhatrises How will the built datastream be used? Can the scanner evaluate both CoreOS configuration and OpenShift configuration in a single run? If oscap can't scan both things in a single run they shouldn't be in a single datastream.

redhatrises:
> So, now this PR is at least build-able. I completed the stub for OCP4 to the build-able state and moved `ocp4/profiles/coreos-ncp.profile => rhel8/profiles/ncp.profile` because _coreos_ is _rhel8_ (as stated above) and you can't use rules that are designed for _rhel8_ in _ocp4_ product. The _ocp4_ product right now is for upper-level "applications configuration" checks, as it was with _ocp3_. I did it only to be able to build the branch with minimal changes.

Got it. `coreos-ncp.profile` should actually stay in the `ocp4` directory and not be used at all under `RHEL8` as coreos is embedded into OCP4.

> Now, @redhatrises you said that you would like to "make the benchmark root able to handle a list of directories". I believe that you mentioned it because you don't want to move `coreos-ncp.profile` to the _rhel8_ sub-folder, and you want to have all checks related to the CoreOS inside the _ocp4_ product, right? 

That's correct since CoreOS is embedded.

> But if I get it correctly it is not only about multiple roots. There are rules in the `coreos-ncp.profile` that belongs to _rhel8_, like _chronyd_client_only_, which is `prodtype: rhel8,fedora,ol8`. Does it make any sense to add _ocp4_ to all these rules (and checks)?

Yup! That's exactly it.

jan-cerny:
I prefer merging the ocp4 product skeleton first, only with a single benchmark root (`../applications`) in product.yml. Then add the second benchmark to the product in a separate pull request. I expect that  adding a second benchmark will require multiple changes across the build scripts because at this moment all products use only a single benchmark. If the change is done as a separate PR it will be easier to review.

matejak:
I am in favor of @jan-cerny comment - the separation of the two benchmark roots is a limitation, and we would like to be able to use multiple *benchmark roots* and to produce datastreams with a single *benchmark*. But that has to be done in another PRs. However, the failing test should be addressed somehow in this PR.
Aside from that, it looks like that we could introduce a new Linux OS of Red Hat CoreOS, since that's what OCP4 runs on and what we may encounter in use cases outside of OpenShift too.

redhatrises:
> I prefer merging the ocp4 product skeleton first, only with a single benchmark root (`../applications`) in product.yml. Then add the second benchmark to the product in a separate pull request. I expect that adding a second benchmark will require multiple changes across the build scripts because at this moment all products use only a single benchmark. If the change is done as a separate PR it will be easier to review.

Completely agree and makes sense.

rhel8/profiles/ncp.profile
```diff
@@ -177,7 +177,7 @@ selections:
 
     ### Module Blacklist
```
redhatrises: The `coreos-ncp.profile` needs to stay in the `ocp4` product directory.

matejak:
Also the `build_product` script needs to be updated with OCP4.

jhrozek:
So I'm still fuzzy on the terminology, but here's my 2c: I think it might make sense to at least allow decoupling OCP and RHCOS in the future if it's not done from the start for two reasons: 1) Even though RHCOS might be used as OCP host only for now, this might change (e.g. the OTA upgrades might maybe make it an attractive platform for an IoT device) and 2) IIRC it still is possible to deploy OCP4 on RHEL (bring-your-own-infra).

jhrozek:
oh, btw - the reason I explicitly said that I'm still not clear on all the terminology is that it might be that I'm just paraphrasing what you already agreed on earlier. In that case, just ignore me :-)

redhatrises:
> So I'm still fuzzy on the terminology, but here's my 2c: I think it might make sense to at least allow decoupling OCP and RHCOS in the future if it's not done from the start for two reasons: 1) Even though RHCOS might be used as OCP host only for now, this might change (e.g. the OTA upgrades might maybe make it an attractive platform for an IoT device) and 2) IIRC it still is possible to deploy OCP4 on RHEL (bring-your-own-infra).

There would be a little work to do, but it can be done in the future. In the meantime, we have been asked by OCP product management to keep RHCOS as part of OCP, so for now, I think that we should honor this request.

rhel8/profiles/ncp.profile
redhatrises: @evgenyz can you please make this change and then take out of draft mode?

evgenyz: It won't be build-able with only this change

redhatrises: @evgenyz understood, but this PR is going to start holding up ocp4 content enablement soon.

evgenyz: This profile uses rules that are not compatible with 'ocp4' product. And I don't really like the idea of adding 'ocp4' as a product to all these rules in `linux_os/guide`, because 'ocp4' is not in fact Linux OS. 

Right now I'm trying to add a way to list multiple `benchmark_root` elements with a sane result, and the deeper I'm going into the problem less and less it is clear for me what is the actual benefit.

matejak:
test this please

rhel8/profiles/ncp.profile
jan-cerny: We think that this could be solved by temporarily setting documentation_complete to false until the other PR is merged.

yuumasato:
I think there are two things to consider, 1)location of rules, 2)products

For 1). rules related to OCP4 make sense to be in `applications/`, and rules for `RHCOS can be in `linux_os/`

For 2), as RHCOS is very closely tied together with OCP4, it makes sense to have OCP4 product leverage rules from both benchmark roots.

evgenyz:
Test please

jan-cerny:
@openscap-ci test this please

jan-cerny:
@evgenyz Great, now you can rebase #4936 on the top of master.

PR #4909, opened 2019-10-08
Update remetiation of mount_option_tmp rules, /tmp is not tmpfs in RHEL

yuumasato:
#### Description:

- In RHEL systems, do not create `/tmp` entry in `/etc/fstab` as a `tmpfs`
  - By default RHEL creates `/tmp` as normal filesystem.
- Enable Anaconda remediations so that partition existence is checked during installation
  - Remediation can only add the mount options, it cannot create the partition.
anaconda_tmp_warning

#### Rationale:
- Align with RHEL filesystem.


shawndwells:
Ack pending jenkins.

matejak:
ACK.

yuumasato:
@shawndwells @matejak Thanks for reviewing!

PR #4910, opened 2019-10-09
Added OCIL for mount option-type of rules.

matejak:


matejak:
Oops, accidental branch removal

shared/macros.jinja
```diff
@@ -598,7 +618,7 @@ ocil_clause: "the correct value is not returned"
 
 
 {{% macro body_of_warning_about_dependent_rule(rule_id, why) -%}}
-        When selecting this rule in a profile, 
+        When selecting this rule in a profile,
```
yuumasato: This change doesn't seem to make a difference, this is what is in the DS:
```
                <ns10:warning category="dependency" xml:lang="en-US">When selecting this rule in a profile,
    make sure that rule with ID <html:code/> is selected as well: because we need it</ns10:warning>
```
and rendered HTML guide:
warning_body_macro
Note that the rule ID is not correctly shown, either I tested the macro incorrectly, or there is bug. It doesn't affect the line continuation, though.

@matejak I'd rather keep the PR changes focused and not touch this line.

matejak: OK, I have just removed a trailing whitespace. I have reverted that by a rebase.

yuumasato: Thanks, sorry If I was a bit pedantic here.

PR #4911, opened 2019-10-09
Fix ansible lint issues

ggbecker:
#### Description:

- This pull request fixes linting issues and ignore some expected issues. Also changes the behavior to go through all generated playbook for given product, which means testing playbooks for each rule individually based on current product.

The rules ignored are: 204,301,303,403
Here you can find more details on their meaning:
https://docs.ansible.com/ansible-lint/rules/default_rules.html

#### Rationale:

- Makes the lint check all green so it can be enabled on All Green Jenkins view tab.

Note1: The ansible --syntax-check uses the same bash script and I tried to keep the original behavior which checks the syntax for each profile ansible playbook.
Note2: Considering how long the test takes to run, it is not feasible to enable it into PR gating.

The jenkins job which actually tests the build is this one, this job had the cmake variable ANSIBLE_CHECKS ON by default, so it triggered the ansible linting tests:
https://jenkins.complianceascode.io/job/scap-security-guide-pull-requests/2171

But as we are not going to enable by default, I'm removing the commit which had enabled by default.

### DIFF
--- a/cmake/SSGCommon.cmake
+++ b/cmake/SSGCommon.cmake
@@ -302,20 +302,20 @@ macro(ssg_build_remediations PRODUCT)
     if (ANSIBLE_PLAYBOOK_EXECUTABLE AND "${OSCAP_VERSION}" VERSION_GREATER "1.2.16")
         add_test(
             NAME "ansible-playbook-syntax-check-${PRODUCT}"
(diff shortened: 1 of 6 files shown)


ggbecker:
Testing ansible playbooks per rule made the complete build in a total of ~36min (13min more than without doing any ansible linting).

My only concern is that if we test only using the playbooks from "all" folder, that it can miss something regarding variables refined by profiles.

jan-cerny:
@ggbecker Thank you very much for your improvement!

PR #4916, opened 2019-10-10
Remove old templating system

jan-cerny:
#### Description:
Remove unused code.

#### Rationale:
This logic has been replaced by the new system introduced recently.


cmake/SSGCommon.cmake
```diff
@@ -348,51 +327,16 @@ macro(ssg_build_xccdf_with_remediations PRODUCT)
...
-            COMMAND "${CMAKE_COMMAND}" -E remove_directory "${BUILD_CHECKS_DIR}/shared/oval"
-            COMMAND env "PYTHONPATH=$ENV{PYTHONPATH}" "${PYTHON_EXECUTABLE}" "${SSG_BUILD_SCRIPTS}/generate_from_templates.py" --languages oval --input "${CMAKE_CURRENT_SOURCE_DIR}/templates" "${SSG_SHARED}/templates" --output "${BUILD_CHECKS_DIR}" "${BUILD_CHECKS_DIR}/shared" --shared "${SSG_SHARED}" --build-config-yaml "${CMAKE_BINARY_DIR}/build_config.yml" --product-yaml "${CMAKE_CURRENT_SOURCE_DIR}/product.yml" build
-            COMMAND env "PYTHONPATH=$ENV{PYTHONPATH}" "${PYTHON_EXECUTABLE}" "${SSG_BUILD_SCRIPTS}/combine_ovals.py" --build-config-yaml "${CMAKE_BINARY_DIR}/build_config.yml" --product-yaml "${CMAKE_CURRENT_SOURCE_DIR}/product.yml" --output "${CMAKE_CURRENT_BINARY_DIR}/oval-unlinked.xml" ${OVAL_COMBINE_PATHS}
-            COMMAND "${XMLLINT_EXECUTABLE}" --format --output "${CMAKE_CURRENT_BINARY_DIR}/oval-unlinked.xml" "${CMAKE_CURRENT_BINARY_DIR}/oval-unlinked.xml"
-            DEPENDS ${OVAL_CHECKS_DEPENDS}
-            DEPENDS ${EXTRA_OVAL_DEPS}
-            DEPENDS ${EXTRA_SHARED_OVAL_DEPS}
```
matejak: Could it be that those `DEPENDS` are also applicable for the new templating system?

jan-cerny: Yes, they're applicable, nice catch. But the files in variables in these DEPENDS aren't generated files, they're files from the source tree. It would be required if we supported incremental rebuilds. But incremental rebuilds don't work anyway for other reasons. So I don't think we should add these DEPENDS back.

PR #4918, opened 2019-10-10
Generate OCIL check for cramfs kernel module

yuumasato:
#### Description:

- Generate OCIL check for cramfs kernel module


jan-cerny:
I have build the RHEL8 datastream with your patch applied and the OCIL question for rule kernel_module_cramfs_disabled is there. Thanks.

Issue #4912, opened 2019-10-10
Some rules in Essential Eight (e8) profile are missing CCE/OVAL/OCIL (RHEL8)


matusmarhefka:
#### Description of problem:
Stats for the Australian Cyber Security Centre (ACSC) Essential Eight (e8) profile in RHEL8:

The following rules are missing CCE:
```
package_talk-server_removed
package_ypbind_removed
package_telnet-server_removed
package_rsh_removed
package_rsh-server_removed
package_quagga_removed
service_avahi-daemon_disabled
package_squid_removed
service_squid_disabled
rpm_verify_ownership
network_sniffer_disabled
sudo_remove_nopasswd
sudo_remove_no_authenticate
sudo_require_authentication
audit_rules_execution_setfiles
sshd_print_last_log
sshd_set_loglevel_info

```

The following rules are missing OVAL:
```
file_permissions_unauthorized_suid
```

The following rules are missing OCIL:
```
security_patches_up_to_date
mount_option_dev_shm_nodev
mount_option_dev_shm_nosuid
mount_option_dev_shm_noexec

```

#### SCAP Security Guide Version:
upstream master

matejak:
Missing OCIL has been fixed by #4910, OCIL for `security_patches_up_to_date` is not needed (see [comment](https://github.com/ComplianceAsCode/content/issues/4913#issuecomment-540450954) in #4913).

matusmarhefka:
All issues have been solved. Thanks @matejak

PR #4919, opened 2019-10-11
Added missing CCEs.

matejak:
Rules featured in RHEL8 profiles and reported in #4912 were missing RHEL8 CCEs.


linux_os/guide/services/obsolete/r_services/package_rsh-server_removed/rule.yml
```diff
@@ -20,6 +20,7 @@ severity: high
 identifiers:
     cce@rhel6: 27062-9
     cce@rhel7: 27342-5
+    cce@rhel8: 82184-3
```
redhatrises: @matejak I am not seeing this package for RHEL8. Should we remove it for RHEL8? Or should we have it enabled still "just in case"? Probably should check the other packages in the list as well.

matejak: Good catch. The rule was in a RHEL8 profile, so I have subsequently removed it from the product and profile in a separate commit.

redhatrises: @matejak do you mind checking to see if the other package_*_removed in this PR exist in RHEL8 or EPEL8?

matejak:
I went in-depth with RHEL8 package_installed and package_removed rules, and I have found that we have package_removed rules in the OSPP CC profile that assure that packages that are not part of RHEL8 are removed.
Thinking it over, one may install a package on own system from EPEL or even another RHEL version, so those rules may still be applicable. I will propose their removal in another PR after this one gets merged.

linux_os/guide/services/obsolete/r_services/package_rsh-server_removed/rule.yml
matejak: This package_removed cleanup started to interfere with the OSPP profile, so I propose not to solve those in the PR.

redhatrises:
> I went in-depth with RHEL8 package_installed and package_removed rules, and I have found that we have package_removed rules in the OSPP CC profile that assure that packages that are not part of RHEL8 are removed.
> Thinking it over, one may install a package on own system from EPEL or even another RHEL version, so those rules may still be applicable. I will propose their removal in another PR after this one gets merged.
> I have at least removed a RHEL8 package_installed rule for non-existent package.

Got it. Makes sense. Ack.

PR #4920, opened 2019-10-11
Remove overly restrictive FIPS checks

### LINKED ISSUE #4917: Excessive FIPS checks
Many of the FIPS related rules unnecessarily check whether or not the installed OS is FIPS certified.

For example, take the rule to check whether AIDE is configured to use FIPS 140-2 hashes.  The criteria are:
```xml
<oval:criteria operator="AND">
  <oval:extend_definition comment="Installed OS is FIPS certified" definition_ref="oval:ssg-installed_OS_is_FIPS_certified:def:1"/>
  <oval:extend_definition comment="Aide is installed" definition_ref="oval:ssg-package_aide_installed:def:1"/>
  <oval:criterion comment="non-FIPS hashes are not configured" test_ref="oval:ssg-test_aide_non_fips_hashes:tst:1"/>
  <oval:criterion comment="FIPS hashes are configured" test_ref="oval:ssg-test_aide_use_fips_hashes:tst:1"/>
</oval:criteria>
```
Whether or not the OS is FIPS certified is orthogonal to whether or not the FIPS approved algorithms are being used.

This is problematic when evaluating benchmarks on `CentOS` since it's not `RHEL` proper and thus not FIPS certified.  Certainly it should fail the"Installed OS is FIPS certified" test, which is it's own stand alone rule, but that shouldn't preclude validation that various crypto components like ssh and AIDE use FIPS approved crypto algorithms and hashes.

chuckatkins:
#### Description:

- This removes the OS FIPS certification check from testing the FIPS related configuration of various individual components.

#### Rationale:

- Whether or not a package is configured to use FIPS approved crypto algorithms and hashes or fips-related packages are installed is orthogonal to whether or not the installed operating system is FIPS certified.   Whether or not the OS is FIPS certified still remains a stand alone rule by itself and should continue to be used in environments where necessary.

- Fixes #4917

### DIFF
--- a/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers/oval/shared.xml
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers/oval/shared.xml
@@ -11,24 +11,21 @@
       </affected>
       <description>Limit the ciphers to those which are FIPS-approved.</description>
     </metadata>
-    <criteria operator="AND">
-      <extend_definition comment="Installed OS is FIPS certified" definition_ref="installed_OS_is_FIPS_certified" />
-      <criteria comment="SSH is configured correctly or is not installed"
-      operator="OR">
-        <criteria comment="sshd is not installed" operator="AND">
-          <extend_definition comment="sshd is not required or requirement is unset"
-          definition_ref="sshd_not_required_or_unset" />
-          <extend_definition comment="rpm package openssh-server removed"
-          definition_ref="package_openssh-server_removed" />
-        </criteria>
-        <criteria comment="sshd is installed and configured" operator="AND">
-          <extend_definition comment="sshd is required or requirement is unset"
-          definition_ref="sshd_required_or_unset" />
-          <extend_definition comment="rpm package openssh-server installed"
-          definition_ref="package_openssh-server_installed" />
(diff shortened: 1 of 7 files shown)


redhatrises:
Sorry but NACK. FIPS cipher checks are required to meet FIPS validated ciphers of which CentOS is not FIPS validated.

redhatrises:
Specifically NIST 800-53 SI-1 and SC-13 `in accordance with applicable federal laws, Executive Orders, directives, policies, regulations, and standards`.

jamescassell:
This same issue causes all new point releases of RHEL to fail all fips rules until DISA adds the most recent version to the "supported" list... lots of false positives.

chuckatkins:
Aren't the rules in question though just supposed to be checking that "only fips approved ciphers are being used", not "only fips approved ciphers are being used on a fips certified is"?  Forgetting about RHEL derivatives for the moment, developing content for other distros, you'd still want to be able to test the crypto configuration regardless of whether or not the os is approved.  There's already a separate rule that covers whether the os is approved or not.  So doesn't it make sense then for the profiles to include all the fips related rules and the ones checking technical configuration should be able to pass on any distro if configured appropriately while the "os is certified" would always fail unless rhel proper at a certified minor release?

PR #4921, opened 2019-10-11
[object_api_server_experimental_encryption_provider_cipher] Fix unclosed capture group & subexpression

alexisfacques:
#### Description

Regex was not working due to an unclosed capture group +  Character ':' is matched in the capture group as it is not a whitespace, but was not matched in the subexpression regex.


redhatrises:
@openscap-ci test this please

redhatrises:
@alexisfacques great catch! Thanks for the PR!

PR #4922, opened 2019-10-11
Prevent an exception when processing empty derived profiles

chuckatkins:
#### Description:

- Prevent an exception by ensuring that only non-empty selections are parsed

#### Rationale:

- When a profile is derived from another but adds no additional rules then the `selections` key returns `None` instead of an empty list, this an exception is thrown what that gets iterated over.  This ensures that the selections are only iterated over when it's a non-empty list, skipping `None` and empty lists.


redhatrises:
@openscap-ci test this please

yuumasato:
LGTM, thanks

PR #4923, opened 2019-10-11
Optionally allow RHEL derivatives to use non-standard profiles

chuckatkins:
#### Description:

- Add a build option that allows content for non-standard profiles to be generated for RHEL derivatives.

#### Rationale:

- Currently only "standard" profiles are generated for RHEL derivatives like CentOS.  This allows all RHEL profiles to be generated for RHEL derivatives.  By using the `enable_derivatives.py` script then the appropriate warnings are already inserted into all of the generated benchmarks noting that the it will not inherit the appropriate certifications that RHEL does.  This option is off by default do there is no change in behavior unless a developer explicitly enables the feature.


redhatrises:
Sorry, but if you are envisioning that any of the government baselines gets built for CentOS; then, this is a NACK.

redhatrises:
Specifically NIST 800-53 SI-1 and SC-13 `in accordance with applicable federal laws, Executive Orders, directives, policies, regulations, and standards`.

PR #4924, opened 2019-10-14
Remove old templating from Developer Guide

jan-cerny:
The old templating system has been removed in https://github.com/ComplianceAsCode/content/pull/4916 but at that time the documentation of the new templating system hasn't been merged, to avoid conflicts the removal of the old documentation is done as a separate PR.


ggbecker:
Thank you for the patch.

PR #4927, opened 2019-10-15
Fix missing OVAL in some of  RHEL 8 rules

jan-cerny:
#### Description:
Some of the rules in RHEL 8 becnhmark which aren't part of any RHEL 8 profile don't contain OVALs. This usually happens when OVAL hasn't been written yet. That's fine, but sometimes it's because of wrong prodtype elements or platform elements in OVAL or their mismatch.  In this PR we don't add new OVALs but we try to address the second problem for some of the rules.

Read commit messages of each commit for specific details.

#### Rationale:

Users can add the rules to their profiles by tailoring, but if the rule is missing OVAL it has limited usability for them.


matusmarhefka:
LGTM.

PR #4929, opened 2019-10-16
Set rule service_timesyncd_enabled prodtype to ubuntu 16.04 and 18.04

matusmarhefka:
#### Description:

- This rule enables service which is not in RHEL and the rule is only
selected in profiles for ubuntu1604 and ubuntu1804 products.
- Also added missing `service_enabled` template.

### DIFF
--- a/linux_os/guide/services/ntp/service_timesyncd_enabled/rule.yml
+++ b/linux_os/guide/services/ntp/service_timesyncd_enabled/rule.yml
@@ -1,5 +1,7 @@
 documentation_complete: true
 
+prodtype: ubuntu1604,ubuntu1804
+
 title: 'Enable systemd_timesyncd Service'
 
 description: '{{{ describe_service_enable(service="systemd_timesyncd") }}}'
@@ -30,3 +32,9 @@ references:
     cis-csc: 1,14,15,16,3,5,6
 
 ocil: '{{{ ocil_service_enabled(service="systemd_timesyncd") }}}'
+
+template:
+    name: service_enabled
+    vars:
+        servicename: systemd-timesyncd
+        packagename: systemd


linux_os/guide/services/ntp/service_timesyncd_enabled/rule.yml
```diff
@@ -1,5 +1,7 @@
 documentation_complete: true
 
+prodtype: multi_platform_ubuntu
```
jan-cerny: you can't use multi_platform_* in prodtype, that's for platform elements in OVAL. You must enumerate ubuntu1404,ubuntu1604,ubuntu1804.

matusmarhefka: Aha, so that is why the build failed, thanks.

DominiqueDevinci:
@matusmarhefka Hello, I was also working on this problem today, and I think that your solution don't works. At compile time (with your patch), I still see that :

```
WARNING: OVAL check 'service_timesyncd_enabled' was not found, removing <check-content> element from the XCCDF rule.
```
And running it on ubuntu1804 ...
```
Title   Enable systemd_timesyncd Service
Rule    xccdf_org.ssgproject.content_rule_service_timesyncd_enabled
W: oscap: Skipping rule that requires an unregistered check system or incorrect content reference to evaluate. Please consider providing a valid SCAP/OVAL instead of http://scap.nist.gov/schema/ocil/2
```

In fact i've spent the day to find the problem, and i've realize that it's only works if the name of the parent directory is exactly the save that the service (for instance with service_systemd_timesyncd_enabled , and the template servicename=systemd_timesyncd , the compilation is good, but the test always fail because the service name is systemd-timesyncd ).
The problem is it's not possible to name the directory service_systemd-timesyncd_enabled (the compilation crash). So i think that we should modify this behavior of the templating system (a desired behavior I guess ?).

jan-cerny:
@matusmarhefka  @DominiqueDevinci 

Nice catch! Thank you. The rule needs to be renamed to `service_systemd-timesyncd_enabled`. That means renaming the rule directory and also to change all occurrences of it in `.profile` files. This solution works for me, at least I tried ubuntu1804. 

In my opinion the desired state of the templating system would be that the OVAL definition ID won't depend on template parameters but instead will be the same as rule ID. That's a bigger change because it would mean changes in all OVAL templates.

matejak:
And the problem doesn't apply only to OVAL IDs, but also to OVAL affected platforms. We have already a jinja macro for that, and we need to eventually convert OVAL templates to use it.

matusmarhefka:
@DominiqueDevinci Thank you for reporting this!

@jan-cerny @matejak Guys, does it make sense to document this into the developer guide? So everybody is aware of this issue when writing templated rules.

jan-cerny:
@matusmarhefka You can report it as an issue and then link the issue from the developer guide.

PR #4931, opened 2019-10-17
Update upload_ansible_roles_to_galaxy.py

redhatrises:
- PyGithub Repository now uses get_contents instead of get_file_contents
- Don't override variables with boolean


jan-cerny:
I usually use the abandoned https://github.com/Ansible-Security-Compliance to test similar PRs.

PR #4933, opened 2019-10-17
Multiroot benchmarks for products

evgenyz:
@jan-cerny, @matejak, @redhatrises

This is an approach to add a way to use rules from multiple benchmarks. Most likely it is not complete yet, but I wanted to show the approach and raise a question in general if this feature is needed at all. It won't make #4908 any easier. Everything is rotating around the product, not the benchmark root.

It might be that I actually did not was really meant. If it is the case, let me know.

### DIFF
--- a/build-scripts/combine_remediations.py
+++ b/build-scripts/combine_remediations.py
@@ -55,15 +55,15 @@ def main():
     product = ssg.utils.required_key(env_yaml, "product")
 
     product_dir = os.path.dirname(args.product_yaml)
-    relative_guide_dir = ssg.utils.required_key(env_yaml, "benchmark_root")
-    guide_dir = os.path.abspath(os.path.join(product_dir, relative_guide_dir))
+    relative_guide_dirs = ssg.utils.required_key(env_yaml, "benchmark_root")
+    guide_dirs = [os.path.abspath(os.path.join(product_dir, d)) for d in relative_guide_dirs]
 
     # As fixes is continually updated, the last seen fix that is applicable for a
     # given fix_name is chosen to replace newer fix_names
     remediation_cls = remediation.REMEDIATION_TO_CLASS[args.remediation_type]
 
     rule_id_to_remediation_map = collect_fixes(
-        product, guide_dir, args.fix_dirs, args.remediation_type)
+        product, guide_dirs, args.fix_dirs, args.remediation_type)
 
     fixes = dict()
     for rule_id, fix_path in rule_id_to_remediation_map.items():
@@ -83,7 +83,7 @@ def main():
     sys.exit(0)
 
 
-def collect_fixes(product, guide_dir, fix_dirs, remediation_type):
+def collect_fixes(product, guide_dirs, fix_dirs, remediation_type):
     # path -> remediation
     # rule ID -> assoc rule
     rule_id_to_remediation_map = dict()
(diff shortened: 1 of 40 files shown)


chromium/product.yml
```diff
@@ -2,6 +2,6 @@ product: chromium
 full_name: Chromium
 type: product
 
-benchmark_root: "./guide"
+benchmark_root: ["./guide"]
```
redhatrises: Let's remove the brackets and make this a comma separated list having the python code convert it into a list.

evgenyz: Any particular reason for not using existing YAML format for lists and going imaginative here?

jan-cerny: evgenyz++, if YAML format is used, then no Python code is needed, because the conversion is done by pyyaml library.

ocp3/product.yml
```diff
@@ -2,7 +2,7 @@ product: ocp3
 full_name: Red Hat OpenShift Container Platform 3
 type: platform
 
-benchmark_root: "../applications"
+benchmark_root: ["../applications","../linux_os/guide/system"]
```
jan-cerny: I envision a problem that we will have to spend a lot of time with "prodtype" elements in "rule.yml"s. Currently, this adds rules eg. "Disable WIFI Network Connection Creation in GNOME" to the OCP content, which is  a rule that doesn't make sense in OpenShift context.

ocp3/profiles/coreos-ncp.profile
```diff
@@ -0,0 +1,62 @@
+documentation_complete: true
+
```
jan-cerny: Why do you change OCP 3? Isn't the request for multiple benchmarks about OCP 4?

evgenyz: Because OCP4 stuff is in different PR. This is just an example, I'll get rid of it before merging.

ocp3/product.yml
evgenyz: That is exactly my concern in #4908. I have some doubts about this multiroot thing being actually useful.

build-scripts/combine_remediations.py
```diff
@@ -55,15 +55,15 @@ def main():
     product = ssg.utils.required_key(env_yaml, "product")
 
     product_dir = os.path.dirname(args.product_yaml)
-    relative_guide_dir = ssg.utils.required_key(env_yaml, "benchmark_root")
-    guide_dir = os.path.abspath(os.path.join(product_dir, relative_guide_dir))
+    relative_guide_dirs = ssg.utils.required_key(env_yaml, "benchmark_root")
```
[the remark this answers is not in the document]
evgenyz: Actually I already regret messing with it. I'm thinking about adding another key like `extra_group_roots`.

chromium/product.yml
redhatrises: Then this needs to be the following:
```
benchmark_root: 
    - "./guide"
    - "./other_root"

ocp3/product.yml
redhatrises: There are more mutiroot benchmarks coming, so it is useful, yes.

chromium/product.yml
evgenyz: No, not necessarily. The `benchmark_root: ["./guide","./other_root"]` abbreviated form is legitimate YAML as well.

evgenyz:
Closing in favour of #4936.

PR #4936, opened 2019-10-20
Extra rules groups for Benchmarks

evgenyz:
This is an approach to add a way to use rules from multiple benchmarks. Rework of #4933.

### DIFF
--- a/docs/manual/developer_guide.adoc
+++ b/docs/manual/developer_guide.adoc
@@ -536,6 +536,22 @@ benchmark_root: "../linux_os/guide"
 .....
 ----
 
+Rules from multiple locations can be used for a single Benchmark. There is an optional key `additional_content_directories` for a list of paths to some arbitrary Groups of Rules
+to be included in the benchmark. Of all the rules collected only following would become a part of the benchmark:
+
+- rules that have the `prodtype` specified in correspondence with the benchmark;
+- rules that have no `prodtype` metadata.
+
+----
+.....
+
+benchmark_root: "../applications"
+additional_content_directories:
+    - "../linux_os/guide"
+
+.....
+----
+
 The Benchmarks are organized into directory structure. The directories represent either groups or rules. The group directories contain `group.yml` and rule directories `rule.yml`.
 The name of the group directory is the group ID, without the prefix.
 Similarly, the name of the rule directory if the rule ID, without the prefix.
--- a/ocp4/product.yml
+++ b/ocp4/product.yml
@@ -4,6 +4,10 @@ type: platform
 
 benchmark_root: "../applications"
 
+additional_content_directories:
+    - "../linux_os/guide/services"
+    - "../linux_os/guide/system"
+
 profiles_root: "./profiles"
 
 pkg_system: "rpm"
--- a/ssg/playbook_builder.py
+++ b/ssg/playbook_builder.py
@@ -30,6 +30,8 @@ class PlaybookBuilder():
                                                        "profiles_root")
         self.profiles_dir = os.path.abspath(
             os.path.join(product_dir, relative_profiles_dir))
+        additional_content_directories = product_yaml.get("additional_content_directories", [])
+        self.add_content_dirs = [os.path.abspath(os.path.join(product_dir, rd)) for rd in additional_content_directories]
 
     def get_profile_selections(self, profile):
         """
@@ -118,7 +120,12 @@ class PlaybookBuilder():
         variable values.
         """
         variables = dict()
-        for dirpath, dirnames, filenames in os.walk(self.guide_dir):
+        for cur_dir in [self.guide_dir] + self.add_content_dirs:
+            variables.update(self._get_rules_variables(cur_dir))
+        return variables
+
+    def _get_rules_variables(self, base_dir):
+        for dirpath, dirnames, filenames in os.walk(base_dir):
             for filename in filenames:
                 root, ext = os.path.splitext(filename)
                 if ext == ".var":
@@ -128,8 +135,7 @@ class PlaybookBuilder():
                     options = dict()
                     for k, v in xccdf_value.options.items():
                         options[str(k)] = str(v)
-                    variables[xccdf_value.id_] = options
-        return variables
+                    yield (xccdf_value.id_, options,)
 
     def _find_rule_title(self, rule_id):
         rule_path = os.path.join(self.rules_dir, rule_id + ".yml")
--- a/ssg/rules.py
+++ b/ssg/rules.py
@@ -85,10 +85,20 @@ def get_rule_dir_ovals(dir_path, product=None):
 
 def find_rule_dirs(base_dir):
     """
-    Generator which yields all rule_directories within a given base_dir
+    Generator which yields all rule directories within a given base_dir, recursively
     """
     for root, dirs, _ in os.walk(base_dir):
         for dir_name in dirs:
             dir_path = os.path.join(root, dir_name)
             if is_rule_dir(dir_path):
                 yield dir_path
+
+
+def find_rule_dirs_in_paths(base_dirs):
+    """
+    Generator which yields all rule directories within a given directories list, recursively
+    """
+    if base_dirs:
+        for cur_dir in base_dirs:
+            for d in find_rule_dirs(cur_dir):
+                yield d
--- a/tests/test_machine_only_rules.py
+++ b/tests/test_machine_only_rules.py
@@ -19,12 +19,14 @@ def main():
         product_yaml = ssg.yaml.open_raw(product_yaml_path)
         guide_dir = os.path.abspath(
             os.path.join(product_dir, product_yaml['benchmark_root']))
-        if not check_product(args.build_dir, product, guide_dir):
+        additional_content_directories = product_yaml.get("additional_content_directories", [])
+        add_content_dirs = [os.path.abspath(os.path.join(product_dir, rd)) for rd in additional_content_directories]
+        if not check_product(args.build_dir, product, [guide_dir] + add_content_dirs):
             sys.exit(1)
 
 
-def check_product(build_dir, product, guide_dir):
-    input_groups, input_rules = scan_rules_groups(guide_dir, False)
+def check_product(build_dir, product, rules_dirs):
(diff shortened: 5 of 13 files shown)


evgenyz:
Test please

evgenyz:
@openscap-ci test this please

ocp4/product.yml
```diff
@@ -0,0 +1,15 @@
+product: ocp4
+full_name: Red Hat OpenShift Container Platform 4
+type: platform
+
+benchmark_root: "../applications"
+
+extra_rule_group_roots:
```
redhatrises: Can this be `additional_directories`?

evgenyz: Well it of course can, but I'd rather prefer it won't. 

The name is sort of a message. First of all `roots` means that it behaves like `benchmark_root` key (diving into subdirs while collecting rules etc) while being a list of these roots (hence the plural form). Next, this option is working in such a way that it will bring into base benchmark only rules, and nothing else from other benchmarks will be inherited; you can have there any directory with a group.yml file and a subdir of rules (so there is `rule_group`). And, finally, these rules will be `extra` rules, nothing would be shadowed, overridden etc, etc.

evgenyz: BTW, do you want to change the behaviour along with the renaming?

jan-cerny: @evgenyz++, I think that @redhatrises's proposal `additional_directories` doesn't directly explain the content of the directories.

matejak:
Please rebase against current master, so there are less changes.

evgenyz:
@matejak Done

docs/manual/developer_guide.adoc
```diff
@@ -536,6 +536,18 @@ benchmark_root: "../linux_os/guide"
 .....
 ----
 
+Rules from multiple locations can be used for a single Benchmark. There is an optional key `extra_rule_group_roots` for a list of paths to some arbitrary Groups of Rules to be processed.
```
matejak: We need to say here how it works with inheritance. For example, the `platform: machine` can be specified both in the rule and in one of parent groups of the rule. Obviously, one can select a group that is not explicit on platform, because it is inherited from its parent group, or I can select a group that defines platform explicitly. How will those rules end up?

evgenyz: Right now it works pretty much like if you just symlink this external group directory into the benchmark. Not sure how it should work, thought.

evgenyz: If this behaviour is what we would like from this improvement I can add there something along the lines that rules would be processed as if they were located inside the directory structure of the benchmark in question and/or relate to a section in the manual that describes how rules are processed in general.

ocp4/product.yml
redhatrises: `extra_rule_group_roots` is equally not accurate. The rules are `additional` rules not `extra` rules. There is more that just rules and groups so `rule_group` is not accurate either so it should be `content`. Finally, we are adding addition group and rule directories, and not `roots` of a benchmark which is also incorrect. `additional_content_directories` is a wayyyyyyy better key name since we are adding additional content to an existing benchmar, and documentation should explain directly in more detail what `additional_content_directories` is and does.

ssg/rules.py
```diff
@@ -83,12 +83,15 @@ def get_rule_dir_ovals(dir_path, product=None):
     return results
 
 
-def find_rule_dirs(base_dir):
+def find_rule_dirs(base_dir, extra_dirs=None):
     """
-    Generator which yields all rule_directories within a given base_dir
+    Generator which yields all rule_directories within a given base_dir and, optionally, extra_dirs directories list
     """
-    for root, dirs, _ in os.walk(base_dir):
```
matejak: ditto, please rename the original function and call it from the new `find_rule_dirs`.

tests/test_machine_only_rules.py
```diff
@@ -88,23 +90,24 @@ def check_if_machine_only(dirpath, name, is_machine_only_group):
...
-    if check_if_machine_only(dirpath, "rule.yml", parent_machine_only):
-        rules.add(name)
-    for dir_item in os.listdir(dirpath):
-        subdir_path = os.path.join(dirpath, dir_item)
-        if os.path.isdir(subdir_path):
-            subdir_groups, subdir_rules = scan_rules_groups(
-                subdir_path, is_machine_only)
-            groups |= subdir_groups
-            rules |= subdir_rules
+    for dir_path in [root_dir_path] + extra_dir_paths:
```
matejak: ditto, extract and call :-)

tests/test_parse_affected.py
```diff
@@ -40,40 +40,43 @@ def main():
         ssg.jinja.add_python_functions(env_yaml)
 
         guide_dir = os.path.join(product_dir, product_yaml['benchmark_root'])
-        if guide_dir in guide_dirs:
-            continue
+        relative_extra_rules_dirs = product_yaml.get("extra_rule_group_roots", [])
+        extra_rules_dirs = [os.path.abspath(os.path.join(product_dir, rd)) for rd in relative_extra_rules_dirs]
 
-        for rule_dir in ssg.rules.find_rule_dirs(guide_dir):
```
matejak: ditto

utils/rule_dir_json.py
```diff
@@ -48,21 +47,19 @@ def walk_products(root, all_products):
         guide_dir = os.path.join(product_dir, product_yaml['benchmark_root'])
         guide_dir = os.path.abspath(guide_dir)
 
-        if guide_dir in visited_guide_dirs:
-            continue
+        relative_extra_rules_dirs = product_yaml.get("extra_rule_group_roots", [])
+        extra_rules_dirs = [os.path.abspath(os.path.join(product_dir, rd)) for rd in relative_extra_rules_dirs]
 
-        new_rule_dirs = sorted(ssg.rules.find_rule_dirs(guide_dir))
```
matejak: ditto

docs/manual/developer_guide.adoc
matejak: I have checked, and I agree with your comments. The only case to watch out for is when you specify a directory of a group that normally inherits properties from its parent group. I have verified that it indeed works as if you symlink the directory into the main benchmark - the target group is disconnected from its parent.
Next, please add a note that if you include an additional directory to the benchmark, following rules will become part of the benchmark:

- rules that have the corresponding prodtype specified
- rules that have no prodtype metadata

ssg/playbook_builder.py
```diff
@@ -118,17 +120,18 @@ def get_benchmark_variables(self):
         variable values.
         """
         variables = dict()
-        for dirpath, dirnames, filenames in os.walk(self.guide_dir):
```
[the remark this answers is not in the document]
evgenyz: Are you proposing to change it to something like that?

```
def _find_rule_dirs(base_dir):
    """
    Generator which yields all rule_directories within a given base_dir
    """
    for root, dirs, _ in os.walk(base_dir):
        for dir_name in dirs:
            dir_path = os.path.join(root, dir_name)
            if is_rule_dir(dir_path):
                yield dir_path

def find_rule_dirs(base_dir, extra_dirs=None):
    """
    Generator which yields all rule_directories within a given base_dir and, optionally, extra_dirs directories list
    """
    if extra_dirs is None:
        extra_dirs = []
    for cur_dir in [base_dir] + extra_dirs:
        for d in _find_rule_dirs(cur_dir):
            yield d
```
Looks kinda weird for me. I see an opportunity to clean it up a bit by leaving only one argument – a list of dirs. And event that would only move the problem a bit upwards.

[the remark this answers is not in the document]
matejak: Yes, you got it right. To make it clear, my suggestion aims for two functions and less `for` loop nesting. Moreover, you can think of better naming. The problem with three-level `for` loop nesting is that it takes a lot of effort to figure out intention of the code.
What you mention is something else - the existing code doesn't have the optimization that you mention, and you can design the `find_rule_dirs` that it accepts a single argument (list of directories), while the body uses single for loop.

ocp4/product.yml
evgenyz: As far as I know and perceive – extra == additional. And extra is a bit shorter. I don't really care, I could change it.

Then, it the middle, we have 'content', 'group' and 'rule' options. I'm non-native speaker and have not much of experience in the project, but for me 'content' word does not transmit the idea of collecting only rules from the directories in question. Can we say that rules == content? If everybody else is comfortable with it I'll change.

And finally 'root' vs 'directories'. Directories better describes what is expected here. Root links behaviour to the benchmark option (recursive scan) and also ties them together to some degree.

Again, I don't *really* care how it would be named, just trying to make most of the sense.

evgenyz:
Updated developer's documentation and changed the code according to the reviews.

ssg/products.py
```diff
@@ -26,10 +26,14 @@ def get_all(ssg_root):
...
-        else:
-            other_products.add(product)
+        relative_extra_rules_dirs = product_yaml.get("extra_rule_group_roots", [])
+        extra_rules_dirs = [os.path.abspath(os.path.join(product_dir, rd)) for rd in relative_extra_rules_dirs]
+
+        for cur_dir in [guide_dir] + extra_rules_dirs:
+            if 'linux_os' in guide_dir:
+                linux_products.add(product)
+            else:
+                other_products.add(product)
```
yuumasato: I don't think it makes sense to have a product in multiple categories, I think it should be on the category of the root benchmark.
These categories are used in the `json` utilities and will confuse the tool if a product is assigned to multiple categories.

evgenyz: I think that you're right. And, if you look closer, you can find that the code is working exactly like that. I just forgot to revert this file completely because all extended extra_rule_group_roots logic is not needed here. Thanks for noticing, though!

ocp4/product.yml
evgenyz: Okay, `additional_content_directories` it is.

evgenyz:
Oh my god. I screwed this PR.

evgenyz:
Okay, unscrewed and squashed. Including renamed vars and key.

matejak:
LGTM, thank you for this enhancement!

PR #4938, opened 2019-10-21
Fix various typos/incorrect descriptions in rules/groups metadata.

matejak:
- [x] Fix weblink macro call.
- [x] Disable Automatic Configuration group description is too specific.
- [x] Enable Kernel Page-Table Isolation rule doesn't logically belong to the Set Boot Loader Password group
- [ ] ~Set Password Quality Requirements group description is too specific.~ this is 
- [x] `linux_os/guide/services/sssd/group.yml` group's description has no RHEL8 link for "more info".


redhatrises:
I believe that the `Disable Automatic Configuration` group is supposed to be a rule and not a group. For PWQuality group, it depends on what the description change is to as it looks fine as is.

matejak:
I have removed the `disabling_ipv6_autoconfig` group, and made the former group a WIP rule.

matejak:
Needs #4945 merged in order to pass tests.

yuumasato:
@openscap-ci test this please

ggbecker:
First this PR needs to be against stabilization branch and then it needs to be rebased. I'll take a look and see what I can do. I may have the permissions to push here.

ggbecker:
Now there is a problem with ctest. Looking into it right now.

ggbecker:
> Now there is a problem with ctest. Looking into it right now.

Minor adjustment to parse-affected test was required.

PR #4939, opened 2019-10-21
Add rule to check cockpit service status

jcpunk:
#### Description:

This adds a rule to examine the status of cockpit remote management server (cockpit-project.org) and disable it.

#### Rationale:

Cockpit is starting to come enabled by default in Fedora, but it presents another vector of attack for login and possible privilege escalation.  This adds a rule to help worried folks assess this status.


jan-cerny:
@openscap-ci ok to test

jan-cerny:
Thank you very much for your patch.

PR #4944, opened 2019-10-23
Remove openshift api_server_profiling check

### LINKED ISSUE #4943: Remove openshift api_server_profiling check
The ocp3 opencis-master profile has a check that expects profiling for the API server to be explicitly disabled.  Red Hat has changed their guidance in this area, as API server profiling is protected via RBAC and can be left enabled or disabled as desired.  From a security perspective, we only want to make sure that the web interface for profiling is disabled for the controller.  We already have the controller_disable_profiling check for this.  We should remove the api_server_profiling check.

nkinder:
The ocp3 opencis-master profile has a check that expects profiling
for the API server to be explicitly disabled.  Red Hat has changed
their guidance in this area, as API server profiling is protected
via RBAC and can be left enabled or disabled as desired.  From a
security perspective, we only want to make sure that the web
interface for profiling is disabled for the controller.  We already
have the controller_disable_profiling check for this.

This patch removes the api_server_profiling check.

- Fixes #4943

### DIFF
--- a/ocp3/profiles/opencis-master.profile
+++ b/ocp3/profiles/opencis-master.profile
@@ -95,7 +95,6 @@ selections:
     - api_server_kubelet_client_cert
     - api_server_kubelet_client_key
     - api_server_kubelet_https
-    - api_server_profiling
     - api_server_request_timeout
     - api_server_secure_port
     - api_server_service_account_ca


shawndwells:
@openscap-ci add to whitelist

shawndwells:
> The ocp3 opencis-master profile has a check that expects profiling
> for the API server to be explicitly disabled. Red Hat has changed
> their guidance in this area, as API server profiling is protected
> via RBAC and can be left enabled or disabled as desired. From a
> security perspective, we only want to make sure that the web
> interface for profiling is disabled for the controller. We already
> have the controller_disable_profiling check for this.
> 
> This patch removes the api_server_profiling check.
> 
> * Fixes #4943

@nkinder noticed this is your first contribution. Welcome!

In the past there has only been one profile for OpenShift 3.x, which ended up being used by commercial, government, and other regulated customers. Chances are we need to create multiple compliance profiles for each of these segments moving forward.

Some of them, like the U.S. government, do not want/allow profiling tools on production systems. Akin to how kdump is disabled on RHEL. To preserve the ability for future profiles to disable profiling, what do you think about removing the rule from the existing profile but keep the OVAL and XCCDF?

This would buy us time to document how profiling-related RBAC controls could be enforced in the mid/long-term, while simultaneously letting more stringent markets like U.S. Government disable profiling in the short-term.

nkinder:
> To preserve the ability for future profiles to disable profiling, what do you think about removing the rule from the existing profile but keep the OVAL and XCCDF?

This sounds like a good approach, and I wondered about just removing the check from the profile.  I will modify the patch accordingly.

yuumasato:
LGTM, Shawn's proposal make sense, and is addressed.

I have changed the base branch to stabilization, so the fix is part of next release.

PR #4945, opened 2019-10-24
Allow incomplete rules.

matejak:
If a rule is marked as not `documentation_complete`, skip it on non-debug builds, like we do with "incomplete" profiles.

### DIFF
--- a/ssg/build_ovals.py
+++ b/ssg/build_ovals.py
@@ -7,7 +7,7 @@ import sys
 from copy import deepcopy
 import collections
 
-from .build_yaml import Rule
+from .build_yaml import Rule, DocumentationNotComplete
 from .constants import oval_namespace as oval_ns
 from .constants import oval_footer
 from .constants import oval_header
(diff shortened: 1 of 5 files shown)


matejak:
I see, force pushed another part of the fix.

jan-cerny:
@matejak That's great, but it haven't fixed my scenario from my comment.

matejak:
I see, I have fixed remaining cases of handling `Rule.from_yaml` throwing an exception.

jan-cerny:
@matejak I have reviewed it again and now it works for me. However, this blocks #4938 . Should this and #4938 got to the stabilization branch instead of master? @yuumasato

yuumasato:
> Should this and #4938 got to the stabilization branch instead of master?

I see both PRs as fixing bugs in the content.
- #4938 fixes many weblink issues
- And this PR fixes the release build when a rule is marked as `documentation_complete: false`

So they make sense to go into stabilization, and thus next release.

PR #4946, opened 2019-10-25
Fail when the rule provided by --rule-id doesn't exist

jan-cerny:
#### Description:
When an user wants to compare only 1 rule, and the rule isn't present in the datastream, for example the user makes a typo in --rule-id value, the script will fail.

### DIFF
--- a/utils/compare_ds.py
+++ b/utils/compare_ds.py
@@ -212,6 +212,8 @@ def get_rules_to_compare(benchmark, rule_id):
             rule_id = ssg.constants.OSCAP_RULE + rule_id
         rules = benchmark.findall(
             ".//xccdf:Rule[@id='%s']" % (rule_id), ns)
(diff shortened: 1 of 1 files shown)


utils/compare_ds.py
```diff
@@ -212,6 +212,8 @@ def get_rules_to_compare(benchmark, rule_id):
             rule_id = ssg.constants.OSCAP_RULE + rule_id
         rules = benchmark.findall(
             ".//xccdf:Rule[@id='%s']" % (rule_id), ns)
+        if len(rules) == 0:
+            raise ValueError("Can't find rule %s" % (rule_id))
```
yuumasato: How about catching this exceptions somewhere and printing a nice message instead of tracebacking?

jan-cerny:
I have caught the exception.

PR #4947, opened 2019-10-28
Add check to set /etc/motd similar to /etc/issue

jcpunk:
#### Description:

Add setting to permit setting /etc/motd in a similar way as /etc/issue.

#### Rationale:

Sites may want the "short" message for `/etc/motd` and the "long" one for `/etc/issue` so that the user notifications are highly visible in multiple places.


jan-cerny:
@openscap-ci ok to test

jan-cerny:
@jcpunk thanks

PR #4948, opened 2019-10-28
Set default to match syslog default

jcpunk:
#### Description:

The auditsp log can have a more useful default.

#### Rationale:

The rsyslog default points to `logcollector` this patch brings the audit default into sync so that folks using this alias can "just work" with the default audit rules.

### DIFF
--- a/linux_os/guide/system/auditing/configure_auditd_data_retention/var_audispd_remote_server.var
+++ b/linux_os/guide/system/auditing/configure_auditd_data_retention/var_audispd_remote_server.var
@@ -14,4 +14,4 @@ type: string
 interactive: true
 
 options:
-    default: myhost.mydomain.com
+    default: logcollector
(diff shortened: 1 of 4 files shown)


jan-cerny:
@openscap-ci test this please

jan-cerny:
@openscap-ci ok to test

linux_os/guide/system/auditing/configure_auditd_data_retention/var_audispd_remote_server.var
```diff
@@ -14,4 +14,4 @@ type: string
 interactive: true
 
 options:
-    default: myhost.mydomain.com
+    default: logcollector
```
jan-cerny: The test scenarios in `linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_audispd_configure_remote_server/tests` need to be updated as well.

jcpunk: I'll fix shortly.

jan-cerny:
@jcpunk thanks

PR #4949, opened 2019-10-28
Fill in the samples with the value from our variable

jcpunk:
#### Description:

Make the documentation cleaner

#### Rationale:

The sample text now uses the value from the variable rather than a hardcoded sample name.

### DIFF
--- a/linux_os/guide/system/logging/rsyslog_sending_messages/rsyslog_remote_loghost/rule.yml
+++ b/linux_os/guide/system/logging/rsyslog_sending_messages/rsyslog_remote_loghost/rule.yml
@@ -10,19 +10,19 @@ description: |-
     Along with these other directives, the system can be configured
     to forward its logs to a particular log server by
     adding or correcting one of the following lines,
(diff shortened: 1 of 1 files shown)


jan-cerny:
@openscap-ci ok to test

jan-cerny:
@openscap-ci test this please

linux_os/guide/system/logging/rsyslog_sending_messages/rsyslog_remote_loghost/rule.yml
```diff
@@ -16,13 +16,13 @@ description: |-
     they may not be supported in all environments.
     <br />
     To use UDP for log message delivery:
-    <pre>*.* @<i>loghost.example.com</i></pre>
+    <pre>*.* @<i><sub idref="rsyslog_remote_loghost_address" /></i></pre>
```
jan-cerny: Can you do the same change on line 13, please?

jcpunk: I'll fix shortly.

PR #4957, opened 2019-10-30
Remove directory_access_var_log_audit from RHEL 7 OSPP

jan-cerny:
#### Description:
Remove directory_access_var_log_audit from RHEL 7 OSPP

#### Rationale:

The audit rule `-a always,exit -F dir=/var/log/audit/
-F perm=r -F auid>=1000 -F auid!=unset -F key=access-audit-trail`
is present in `/usr/share/doc/audit-2.8.5/rules/30-ospp-v42.rules`
(checked on audit-2.8.5-4.el7.x86_64). That means this audit rule
is already checked and remediated by rule `audit_rules_for_ospp`.

### DIFF
--- a/rhel7/profiles/ospp.profile
+++ b/rhel7/profiles/ospp.profile
@@ -285,13 +285,11 @@ selections:
     ## AU-2(a) / FAU_GEN.1.1.c
     ## Audit Kernel Module Loading and Unloading Events (Success/Failure)
     ## AU-2(a) / FAU_GEN.1.1.c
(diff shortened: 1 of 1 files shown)


yuumasato:
This is similar to https://github.com/ComplianceAsCode/content/pull/4782.
Are we nit picky enough to make the comments the same?

yuumasato:
>  Are we nit picky enough to make the comments the same?

For consistency I'd say yes! :D

PR #4958, opened 2019-10-31
remove rule package_telnet_removed from profiles in rhel7, rhel8, rhv4

vojtapolasek:
#### Description:

The rule package_telnet_removed was removet from all profiles shipped with rhel7, rhel8 and rhv4

#### Rationale:

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1729222

The rule package_telnet_removed removes Telnet client which is needed by some installations, especially fence agents. The Telnet server, which introduces much bigger security problem, is removed by a different rule. It should be fine to keep the Telnet client as it does not introduce direct security risk.

### DIFF
--- a/rhel7/profiles/C2S.profile
+++ b/rhel7/profiles/C2S.profile
@@ -101,7 +101,6 @@ selections:
     - package_ypbind_removed
     - package_rsh_removed
     - package_talk_removed
-    - package_telnet_removed
     - sysctl_net_ipv4_ip_forward
     - sysctl_net_ipv4_conf_all_send_redirects
     - sysctl_net_ipv4_conf_default_send_redirects
--- a/rhel7/profiles/e8.profile
+++ b/rhel7/profiles/e8.profile
@@ -19,7 +19,6 @@ selections:
   - package_xinetd_removed
   - service_xinetd_disabled
   - package_ypbind_removed
-  - package_telnet_removed
   - service_telnet_disabled
   - package_telnet-server_removed
   - package_rsh_removed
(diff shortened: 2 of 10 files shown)


yuumasato:
I wonder if other Profiles should not remove `telnet` as well.

redhatrises:
This is a NACK. Any use of telnet whether client or server brings unencrypted remote authentication.

yuumasato:
>  Any use of telnet whether client or server brings unencrypted remote authentication and is a huge problem.

@redhatrises Note it is just removing rule for the "client removed", any rule for server removed is preserved. 

Can you clarify what is the problematic use case with the telnet client installed?

vojtapolasek:
Just to clarify why I am suggesting this, telnet is a dependency of several fence agents, to be exact 11 packages in rhel7 and 10 packages in rhel8.

redhatrises:
> > Any use of telnet whether client or server brings unencrypted remote authentication and is a huge problem.
> 
> @redhatrises Note it is just removing rule for the "client removed", any rule for server removed is preserved.
> 
> Can you clarify what is the problematic use case with the telnet client installed?

The problem is that it initiates an unencrypted remote session which can pass username and password in the clear. The fence agents a part of layered products need to start following good security practices as well now which means they need to start removing `telnet` functionality.

redhatrises:
Closing. See Steve Grubb's added comments about this https://lists.fedorahosted.org/archives/list/scap-security-guide@lists.fedorahosted.org/message/CYAR26WBNZIZHIO76LNKQINORX3UPWL7/

PR #4960, opened 2019-10-31
Don't require that files exist when path is regex

yuumasato:
#### Description:

- Add option to change `check_existence` of `file_test` to `any_exist` via `missing_file_pass` template option.
- Fixes evaluation of following rules when by run oscap-anaconda-addon. During installation phase, there are no SSH keys in `/etc/ssh/`.
  - `file_permissions_sshd_pub_key`
  - `file_permissions_sshd_private_key`
- Note: I have checked that after installation and first boot, any keys in `/etc/ssh/` has compliant permissions `0644`.
- Also updated rules `file_ownership_cron_allow` and `file_groupownership_cron_allow` to use the template.

#### Rationale:

- When checking for permissions of files determined by regular expression, accept that there can be no matches, and pass.
- When a regular expression is used, and no file is found, the test evaluates to fail:
https://github.com/OVALProject/Language/blob/master/docs/oval-definitions-schema.md#-testtype-


jan-cerny:
@yuumasato This is great! Can we have a test scenario for that?

jan-cerny:
@yuumasato And if there is an exact file path specified, and the file doesn't exist, should the result be fail or pass?

jan-cerny:
For example rule `file_groupowner_cron_allow`:

https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/cron_and_at/restrict_at_cron_users/file_groupowner_cron_allow/rule.yml

says in the description:
https://github.com/ComplianceAsCode/content/blob/d883b3d139dc310b6f649a1e6571a1a16cdcf355/linux_os/guide/services/cron_and_at/restrict_at_cron_users/file_groupowner_cron_allow/rule.yml#L8

However, it uses template `file_groupowner` which uses `check_existence="all_exist"`:

https://github.com/ComplianceAsCode/content/blob/d883b3d139dc310b6f649a1e6571a1a16cdcf355/shared/templates/template_OVAL_file_groupowner#L14

That is in contradiction with the rule description which clearly suggest the rule should be evaluated as "pass" if the file doesn't exist.

Fun fact: the problem has been fixed for RHEL 7 by providing manually created oval which uses `check_existence="any_exist"`: 

https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/cron_and_at/restrict_at_cron_users/file_groupowner_cron_allow/oval/rhel7.xml

And moreover, there is a test scenario that checks this, but passes only on RHEL7, but isn't marked as RHEL7 only.
https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/cron_and_at/restrict_at_cron_users/file_groupowner_cron_allow/tests/file_not_there.pass.sh

That suggests this check_existence is a broader problem in multiple similar rules.

redhatrises:
For all intents and purposes, this was intentional that file permissions fail if the configuration file doesn't exist for most cases. Also, `file_permissions_sshd_private_key` might have different permissions (`0600` vs `0640` vs `0644`) which will fail on rescan and will catch people off guard. @shawndwells what are your thoughts? If a configuration file doesn't exist, should a rule automatically pass? Or should there be a determination of which rules should pass and/or fail based off of if a configuration file exists?

yuumasato:
> This is great! Can we have a test scenario for that?

No, when there are no SSH keys, the Test Suite doesn't work.
```
kex_exchange_identification: read: Connection reset by peer
Failed to connect!
```

> @yuumasato And if there is an exact file path specified, and the file doesn't exist, should the result be fail or pass?

Good question, the rules are about file permissions, not file existence. At the same time a rule that passes when the file doesn't actually exist is misleading. 
Applicability of the rule should probably be based on file existence, but that is a different problem.
For now I'd just leave the static file paths as must exist.

> Fun fact: the problem has been fixed for RHEL 7 by providing manually created oval which uses `check_existence="any_exist"`:

Nice finding!

yuumasato:
> Also, `file_permissions_sshd_private_key` might have different permissions (`0600` vs `0640` vs `0644`) which will fail on rescan and will catch people off guard.

If the keys don't exist during Anaconda installation phase, there is nothing we can do.
And it is confusing if report from installation says `fail, but after first boot, the keys are there with correct permissions.

jan-cerny:
> Good question, the rules are about file permissions, not file existence. At the same time a rule that passes when the file doesn't actually exist is misleading.

I agree, that would be misleading, I prefer failing when the file doesn't exist.
If we go that way, in `file_groupowner_cron_allow` the rule description should be changed and the special RHEL7 OVAL removed, I don't understand the reason why the OVAL is different on RHEL7 than on other OSs.

yuumasato:
> If we go that way, in `file_groupowner_cron_allow` the rule description should be changed and the special RHEL7 OVAL removed

On the other hand, this rule `file_groupowner_cron_allow` is the only one which mentions in the description `if file exists`.

```
$ for file in $(grep -lR "name: file_" linux_os/); do grep "exists" $file; done;
    If <tt>/etc/cron.allow</tt> exists, it must be group-owned by <tt>root</tt>.
    If the owner of the cron.allow file is not set to root, the possibility exists for an
    If <tt>/etc/cron.allow</tt> exists, it must be owned by <tt>root</tt>.
    If the owner of the cron.allow file is not set to root, the possibility exists for an
```
> , I don't understand the reason why the OVAL is different on RHEL7 than on other OSs.

Other OSs should have it similar to what RHEL7 has.

shared/templates/template_OVAL_file_permissions
```diff
@@ -13,7 +13,12 @@
       <criterion comment="Check file mode of {{{ FILEPATH }}}" test_ref="test_file_permissions{{{ FILEID }}}" />
     </criteria>
   </definition>
-  <unix:file_test check="all" check_existence="all_exist" comment="Testing mode of {{{ FILEPATH }}}" id="test_file_permissions{{{ FILEID }}}" version="1">
+  {{%- if FILE_REGEX -%}}
+    {{# When a regex doesn't match any files, and check_existence is all_exist, the test evaluates to fail. #}}
```
jan-cerny: We also have templates template_OVAL_file_owner and template_OVAL_file_groupowner. For example, in OpenShift benchmark there are 3 rules suing regex for /etc/cni/net.d/.* and 3 rules for /etc/origin/openvswitch/.*  When you do this change only for template_OVAL_file_permissions, and the files don't exist, in each group the permissions check will pass but the owner and group checks will fail. Is it a problem?

jan-cerny: I suggest to solve it this way: add a new parameter to the template eg. fail_if_not_present.

yuumasato:
@jan-cerny Thanks for the review.
I have updated all three templates, and added a new option to the template: `missing_file_pass`, inspired by `missing_parameter_pass` [option](https://github.com/ComplianceAsCode/content/blame/master/docs/manual/developer_guide.adoc#L1415) in template `audit_lineinfile`.

jan-cerny:
Thanks!

PR #4970, opened 2019-11-04
OCP4 content cleanup

evgenyz:
Make OCP4 content usable


jhrozek:
Unfortunately this still does not work for me. I pushed the content into this repo/branch:
https://github.com/jhrozek/content/tree/ek_ocp4
from there I built this container: `quay.io/jhrozek/ocp4-openscap-content:ek_test`
you can view all the builds here:
https://quay.io/repository/jhrozek/ocp4-openscap-content?tab=builds
and still I'm getting not applicable.

I used this pod to test:
```
apiVersion: "v1"
kind: Pod
metadata:
    name: scanner-test
spec:
    initContainers:
        - name: content-container
          image: quay.io/jhrozek/ocp4-openscap-content:ek_test
          command: ["sh", "-c", "cp /var/lib/content/*.xml /content"]
          volumeMounts:
              - mountPath: "/content"
                name: content-volume
    containers:
        - name: openscap-scanner
          image: quay.io/jhrozek/openscap-ocp
          command: ["sh", "-c", "sleep 3600"]
          env:
              - name: HOSTROOT
                value: "/host"
              - name: PROFILE
                value: "xccdf_org.ssgproject.content_profile_coreos-fedramp"
              - name: CONTENT
                value: "ssg-ocp4-ds.xml"
              - name: REPORT_DIR
                value: "/reports"
              - name: RULE
                value: "xccdf_org.ssgproject.content_rule_selinux_state"
          volumeMounts:
              - mountPath: "/reports"
                name: report-volume
              - mountPath: "/content"
                name: content-volume
              - mountPath: "/host"
                name: host-root
    restartPolicy: Never
    volumes:
        - name: report-volume
          emptyDir: {}
        - name: content-volume
          emptyDir: {}
        - name: host-root
          hostPath:
              path: /
```
Then you can create the pod with:
```
oc create -f openscap-pod-with-init.yaml
```
This would first copy the content from the init container and then the scanner container would just sleep. So you can rsh into it:
```
oc rsh --container openscap-scanner pods/scanner-test sh
``` 
And then you can just do whatever you like, inspect the content:
```
oscap info /content/ssg-ocp4-ds.xml 
Document type: Source Data Stream
Imported: 2019-11-05T12:06:43

Stream: scap_org.open-scap_datastream_from_xccdf_ssg-ocp4-xccdf-1.2.xml
Generated: (null)
Version: 1.3
Checklists:
        Ref-Id: scap_org.open-scap_cref_ssg-ocp4-xccdf-1.2.xml
                Status: draft
                Generated: 2019-11-04
                Resolved: true
                Profiles:
                        Title: Open Computing Information Security Profile for OpenShift Node
                                Id: xccdf_org.ssgproject.content_profile_opencis-node
                        Title: Open Computing Information Security Profile for OpenShift Node
                                Id: xccdf_org.ssgproject.content_profile_coreos-fedramp
                Referenced check files:
                        ssg-ocp4-oval.xml
                                system: http://oval.mitre.org/XMLSchema/oval-definitions-5
                        ssg-ocp4-ocil.xml
                                system: http://scap.nist.gov/schema/ocil/2
Checks:
        Ref-Id: scap_org.open-scap_cref_ssg-ocp4-oval.xml
        Ref-Id: scap_org.open-scap_cref_ssg-ocp4-ocil.xml
        Ref-Id: scap_org.open-scap_cref_ssg-ocp4-cpe-oval.xml
Dictionaries:
        Ref-Id: scap_org.open-scap_cref_ssg-ocp4-cpe-dictionary.xml
```

Or run the scan:
```
oscap-chroot /host xccdf eval --fetch-remote-resources --profile xccdf_org.ssgproject.content_profile_coreos-fedramp --report /tmp/report.xml --rule xccdf_org.ssgproject.content_rule_selinux_state /content/ssg-ocp4-ds.xml
Title   Ensure SELinux State is Enforcing                                 
Rule    xccdf_org.ssgproject.content_rule_selinux_state
E: oscap:       RPM: cannot open Packages database in    
                                                                               
error: cannot open Packages database in                      
E: oscap:       RPM: cannot open Packages database in 
                                                                               
error: cannot open Packages database in 
Result  notapplicable
```

jhrozek:
feel free to ping me if you need access to an openshift cluster.
btw if I can ask for one more modification: we should name the file fedramp-moderate not just fedramp. This is my fault, I named the file originally, but it's better to include the level in the name from the start.

jan-cerny:
If there isn't any RPM database in the scanned container we will have to invent a different way of determining platform applicability, eg. based on reading /etc/os-release.

jhrozek:
@ashcrow had some idea how could a RHCOS node be detected, but sadly, I didn't take notes. Do you still remember what you proposed @ashcrow ?

jhrozek:
Yeah, I think this is good, we're not trying to write content for RHCOS in any other context than OCP worker, so at least for now we can assume that RHCOS == OCP. If the `REDHAT_SUPPORT_PRODUCT` and/or `REDHAT_BUGZILLA_PRODUCT` fields are at least sort of stable we could also look at them to distinguish between 'bare' RHCOS and OCP.

That said, we're likely need to go a step further (not needed /right now/, but still..) because AFAIK it's possible to use RHEL for the OCP worker nodes. So we'll also need to allow the OCP content to run on RHEL nodes. But, that's not needed for this first iteration, we should just keep that in mind to not shoot ourselves in the foot now.

ashcrow:
@jhrozek :+1:. The fields should be stable enough. You could also look at `NAME` which should be "Red Hat Enterprise Linux CoreOS"

shared/checks/oval/installed_app_is_ocp4.xml
```diff
@@ -1,28 +0,0 @@
-<def-group>
```
redhatrises: Instead of being removed, this needs to be updated to handle coreos as well as rhel nodes.

ocp4/profiles/coreos-fedramp.profile
```diff
@@ -0,0 +1,17 @@
+documentation_complete: true
```
redhatrises: This profile should not be added at this point in time. While Fedramp is the starting target, there are other profiles that are named differently that are essentially the same profile.

shared/checks/oval/installed_env_is_ocp4.xml
```diff
@@ -0,0 +1,15 @@
+<def-group>
+  <definition class="inventory" id="installed_env_is_ocp4" version="1">
```
redhatrises: This doesn't need to be added. The existing one needs to be updated.

redhatrises:
The rpm database definitely exists on CoreOS and the `redhat-release-coreos` rpm can be used to identify CoreOS. If the scanning container doesn't have access to scan all of CoreOS, this is a problem.

shared/checks/oval/installed_OS_is_rhel8.xml
```diff
@@ -41,14 +41,16 @@
...
-  <linux:rpminfo_object id="obj_rhel8_coreos" version="1">
-    <linux:name>redhat-release-coreos</linux:name>
-  </linux:rpminfo_object>
+  <ind:textfilecontent54_test check="all" comment="redhat-release-coreos is version 8" id="test_rhel8_coreos" version="1">
+    <ind:object object_ref="obj_rhel8_coreos" />
+    <ind:state state_ref="state_rhel8_coreos" />
+  </ind:textfilecontent54_test>
+  <ind:textfilecontent54_object id="obj_rhel8_coreos" version="1">
+    <ind:filepath>/etc/os-release</ind:filepath>
+    <ind:pattern operation="pattern match">^PRETTY_NAME=&quot;Red Hat Enterprise Linux CoreOS \d+\.(\d+)\.\d+\.\d+ \([\w\s]+\)&quot;$</ind:pattern>
```
JAORMX: @ashcrow what do you think of this, to check if a node is RHCOS?

ashcrow: `NAME` would be a simpler check if all that matters is if it's RHCOS or not ... but I think this would work. NOTE: I just eyeballed the regex :slightly_smiling_face:

evgenyz: I hope that it did not put too much strain to your eyes. I'm open to suggestions for improvements of the regex used in the check :)

shared/checks/oval/installed_OS_is_rhel8.xml
```diff
@@ -41,14 +41,16 @@
...
+    <ind:instance operation="greater than or equal" datatype="int">1</ind:instance>
+  </ind:textfilecontent54_object>
+  <ind:textfilecontent54_state id="state_rhel8_coreos" version="1">
+    <ind:subexpression operation="pattern match">80</ind:subexpression>
```
jhrozek: As discussed on chat, this needs to be more generic, OCP 4.3 already contains a different version tag, so the regex in PR would only work for 4.2

evgenyz:
So, first of all, the RPM-based check is no good at this moment because, while there is an RPM database in the CoreOS edition of RHEL (as a backward compatibility) it is at least located in a different place and current OVAL probe of the oscap scanner is unable to work with it. We will most likely be fixing the probe as there are a lot of RPM checks in the rules and they might be needed in the future. But, to continue with the OCP4 content evolution, I suggest to stick with the `os-release` file check for now.

Next, the version. Original check, submited by @redhatrises, was checking for package version of `redhat-release-coreos` to be 8. But the CoreOS edition actually bears two versions mixed together, the CoreOS itself and RHEL (e.g. 4.3 and 8.1) forming a very peculiar string `VERSION="42.80.20191022.0"`. I used the `(\d)\d+` part out of it, in coherence with previous RPM-based check, but I'm not sure if it's a good idea. This check is a complimentary check to the base RHEL's, and maybe we'd like to check here for CoreOS version instead.

shared/checks/oval/installed_OS_is_rhel8.xml
evgenyz: I've updated the check to accept all 8.x versions of the RHEL part of the version (sigh). The question about using CoreOS versions in this check still stands.

jhrozek:
On 11/7/19 9:22 AM, Evgeny Kolesnikov wrote:
>
> Next, the version. Original check, submited by @redhatrises 
> <https://github.com/redhatrises>, was checking for package version of 
> |redhat-release-coreos| to be 8. But the CoreOS edition actually bears 
> two versions mixed together, the CoreOS itself and RHEL (e.g. 4.3 and 
> 8.1) forming a very peculiar string |VERSION="42.80.20191022.0"|.
>

It sounds to me like|the 42 in||VERSION="42.80.20191022.0" |is 
actually|the OCP version and 8.0 is the base RHEL version. My OCP 4.3 
cluster has 43.81....
|

> I used the |(\d)\d+| part out of it, in coherence with previous 
> RPM-based check, but I'm not sure if it's a good idea. This check is a 
> complimentary check to the base RHEL's, and maybe we'd like to check 
> here for CoreOS version instead.
>
I'm not sure how easy it is, but checking if the OCP version is 4[digit] 
and the RHEL version is 8[digit] sounds the safest to me in the sense 
that it checks both version substrings.

But maybe we're overdoing this and as was said already earlier in the 
PR, just checking the RHCOS is OK. Let me test your latest PR and if it 
works, then I guess it could be good enough.

jan-cerny:
Looks good to me, I agree with you that we need to check the version == 8.*, because the check is a part of a check that checks if the operating system is RHEL 8. I think a check that would check only CoreOS without version might match future versions that will not be based on RHEL8.

jhrozek:
I tested the latest version and can confirm that it works:
```
oscap-chroot /host xccdf eval --fetch-remote-resources --profile xccdf_org.ssgproject.content_profile_coreos-fedramp --report /tmp/report.xml /content/ssg-ocp4-ds.xml
Title   Ensure SELinux State is Enforcing
Rule    xccdf_org.ssgproject.content_rule_selinux_state
E: oscap:       RPM: cannot open Packages database in 
                                                                               
error: cannot open Packages database in 
E: oscap:       RPM: cannot open Packages database in 
                                                                               
error: cannot open Packages database in 
WARNING: Skipping rule that requires an unregistered check system or incorrect content reference to evaluate. Please consider providing a valid SCAP/OVAL instead of http://scap.nist.gov/schema/ocil/2
Result  notchecked                
                                                                               
Title   Enable auditd Service                                     
Rule    xccdf_org.ssgproject.content_rule_service_auditd_enabled
E: oscap:     RPM: cannot open Packages database in 
                                                                               
error: cannot open Packages database in 
E: oscap:     RPM: cannot open Packages database in                                                                                                           
                                                                               
error: cannot open Packages database in 
Result  fail             
                                                                               
Title   Verify All Account Password Hashes are Shadowed
Rule    xccdf_org.ssgproject.content_rule_accounts_password_all_shadowed
Result  pass
```

shared/checks/oval/installed_app_is_ocp4.xml
```diff
@@ -10,19 +10,6 @@
     </metadata>
     <criteria>
       <extend_definition comment="RHEL8 OS installed" definition_ref="installed_OS_is_rhel8" />
-      <criterion comment="OpenShift Node is installed" test_ref="test_ocp4_node" />
```
redhatrises: You still need to check if OCP4 is installed. Checking just the OS install is not enough.

JAORMX: @redhatrises we don't support running RHCOS standalone. So OC4 has to be there.

redhatrises: @JAORMX agreed. That is my original statement in this part of the review.

evgenyz: If you tell me how exactly this node can be detected, I'd happily add this check.

evgenyz: I mean *beside* the obvious way of copying the checks from ocp3 and replacing the number, because RPM-based checks would not work.

ashcrow: Off the top of my head the only way I'm aware of that would be representative to if the node was _actively_ in a cluster would be to use `oc` itself. More or less `oc status` and check the return code. However, RHCOS isn't used in any other context other than OCP and that's by design which is why OCP 4 components, such as the cri-o runtime, mcd, and kubelet are baked in.

JAORMX: Thanks for chiming in folks! If checking `oc status` is acceptable then let's use that to check the node. If that's not an acceptable solution, I'm fine with the PR merging as-is.

jhrozek: If we don't support RHCOS in other context than OCP4 then why is the additional check required? Future-proofing?

redhatrises: The content needs to check the following and oc status doesn't work without an `oc` probe last I checked:

1. RHEL version
2. CoreOS version
3. OCP version installed on CoreOS
4. OCP version installed on RHEL node

ashcrow: Correct. For `oc` to work the node must be in the cluster. I believe the following would get you the information on both:
- Version: `VERSION` in /etc/os-release
- OCP version: Check `openshift-clients` package
```
rpm -q --qf "%{VERSION}" openshift-clients
```

jhrozek: Why can't we use the OCP version from the VERSION line in /etc/os-release again? We're already parsing that file..
VERSION="42.80.20191022.0 => OCP version=4.2, RHCOS version=8.0

redhatrises: Because that actually doesn't guarantee that the openshift bits are installed... just coreos, and it also won't work on a RHEL node.

ashcrow: Correct. The bits will be there on RHCOS as it's a component of Openshift, but RHEL may not have them since RHEL is general purpose.

redhatrises: Still doesn't matter if the bits are already installed on CoreOS because CoreOS is going to have a CPE (which is currently missing) that isn't the OpenShift CPE. CPE is part of this which is why just checking `/etc/os-release` isn't correct for checking the OCP parts of the check. In fact when CoreOS gets a CPE, we are going to have to split out the coreos installed check anyway.

redhatrises: In fact while we are having this discussion, there is going to have to be a way to validate with a gpg check not only the software installed in CoreOS, but also a container that is installed on coreos.

ashcrow: > In fact while we are having this discussion, there is going to have to be a way to validate with a gpg check not only the software installed in CoreOS, but also a container that is installed on coreos.

redhatrises: Still have to validate regardless of source against a key to meet compliance requirements.

JAORMX: I say we merge this for now so we can start using this to build content already. And we start iterating on this as we go. I'm fine with merging this as it is.

jhrozek: I understand where you're coming from and especially the points about RHEL nodes are valid. But we need to start working on the content and even just RHCOS-based is good enough.

On the RHCOS nodes: just knowing that "a node is RHCOS" is good enough at least for now because 1) RHCOS only ever exists as an OCP node 2) The content is supposed to be used from an operator anyway which can only run in OCP.

Again, I agree with your points. But if possible, I would prefer to merge the stub of the content ASAP and then iterate on it to provide a technically more sound solution.

jan-cerny:
The support for RHEL nodes can be implemented later in separate PR. This is good enough for this moment.

PR #4962, opened 2019-11-03
Fix OCP4 SCAP content validation problem

evgenyz:
Will fix #4961

PR #4972, opened 2019-11-05
Merging v0.1.47 into master after release

yuumasato:
#### Description:

- Merge fixes done for v0.1.47 into master, so they are part of next release.


tests/test_parse_affected.py
```diff
@@ -52,7 +52,11 @@ def main():
 def parse_affected(cur_dir, env_yaml):
     for rule_dir in ssg.rules.find_rule_dirs(cur_dir):
         rule_path = os.path.join(rule_dir, "rule.yml")
-        rule = ssg.build_yaml.Rule.from_yaml(rule_path, env_yaml)
+        try:
+            rule = ssg.build_yaml.Rule.from_yaml(rule_path, env_yaml)
+        except ssg.build_yaml.DocumentationNotComplete:
+            # Happens on non-debug build when a rule is "documentation-incomplete"
+            continue
```
yuumasato: @evgenyz @matejak There were conflicts from #4945 and #4962. 
Please double check that I resolved it correctly.

redhatrises:
Why is this a separate branch from the existing stabilization branch?

jan-cerny:
@yuumasato It seems that commits have been squashed which means we lost commit history.

yuumasato:
> Why is this a separate branch from the existing stabilization branch?

My mistake, sorry.

yuumasato:
> @yuumasato It seems that commits have been squashed which means we lost commit history.

@jan-cerny As far as I understand no history was lost. This was a merge commit of two diverging branches.

tests/test_parse_affected.py
ggbecker: it looks like it is correct, I added this piece of code actually.

PR #4989, opened 2019-11-07
make aide_periodic_cron_checking accepting broader array of time specs

vojtapolasek:
#### Description:

Modified the oval so that more ways of specifying the desired periodicity are passing.
- specifying weekdays by name
- ranges of weekdays
- asterisk in weekday column

Changed regexes for all crontab locations.
Removed acceptance of `@monthly` because it violates the rule description.
Modified remediation so that it removes invalid crontab entries pertaining to this rule.
Extended tests.

#### Rationale:

The rule specifies that the check should be run at least weekly and at most daily. However, the rule didn't accept for example
21 21 * * 6 ...
But this is valid specification of a weekly job.

- Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1658036

### DIFF
--- a/linux_os/guide/system/software/integrity/software-integrity/aide/aide_periodic_cron_checking/oval/shared.xml
+++ b/linux_os/guide/system/software/integrity/software-integrity/aide/aide_periodic_cron_checking/oval/shared.xml
@@ -29,7 +29,7 @@
   </ind:textfilecontent54_test>
   <ind:textfilecontent54_object comment="run aide with cron" id="object_test_aide_periodic_cron_checking" version="1">
     <ind:filepath>/etc/crontab</ind:filepath>
-    <ind:pattern operation="pattern match">^(([0-9]*[\s]*[0-9]*[\s]*\*[\s]*\*[\s]*\*)|@(hourly|daily|weekly|monthly))[\s]*root[\s]*/usr/sbin/aide[\s]*\-\-check.*$</ind:pattern>
+    <ind:pattern operation="pattern match">^(([0-9]*[\s]*[0-9]*[\s]*\*[\s]*\*[\s]*[\*,0-9])|@(hourly|daily|weekly|monthly))[\s]*root[\s]*/usr/sbin/aide[\s]*\-\-check.*$</ind:pattern>
     <ind:instance datatype="int" operation="greater than or equal">1</ind:instance>
   </ind:textfilecontent54_object>
 
--- a/linux_os/guide/system/software/integrity/software-integrity/aide/aide_periodic_cron_checking/oval/shared.xml
+++ b/linux_os/guide/system/software/integrity/software-integrity/aide/aide_periodic_cron_checking/oval/shared.xml
@@ -29,7 +29,7 @@
   </ind:textfilecontent54_test>
   <ind:textfilecontent54_object comment="run aide with cron" id="object_test_aide_periodic_cron_checking" version="1">
     <ind:filepath>/etc/crontab</ind:filepath>
-    <ind:pattern operation="pattern match">^(([0-9]*[\s]*[0-9]*[\s]*\*[\s]*\*[\s]*[\*,0-9])|@(hourly|daily|weekly|monthly))[\s]*root[\s]*/usr/sbin/aide[\s]*\-\-check.*$</ind:pattern>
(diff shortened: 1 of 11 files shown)


redhatrises:
@vojtapolasek is there a way to make a XCCDF variable for this? That way we could probably handle this better for more use cases.

jan-cerny:
@openscap-ci test this please

vojtapolasek:

vojtapolasek:
> @vojtapolasek is there a way to make a XCCDF variable for this? That way we could probably handle this better for more use cases.

There are two slightly different regexes in the oval, the one for /var/spool/crontab is slightly different. I am not sure if using variable would bring such a big benefit here. @yuumasato ?

yuumasato:
> Tre are two slightly different regexes in the oval, the one for /var/spool/crontab is slightly different.

Technically this is not a problem, you can define a local_variable and concatenate the the XCCDF Value with the fixed parts of the regex. See, [crypto_policy OVAL](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/oval/shared.xml#L17) for example.

For what I've checked, the regex seems to be generic enough to allow the pass cases.
I noticed the regex doesn't allow for 'mon-fri`, but `1-5` should cover it.

> That way we could probably handle this better for more use cases.

@redhatrises What kind of customization would you like to allow?

vojtapolasek:
I believe that regex not allowing mon-fri is right, as per this part of the Crontab (5) man page:
Names can also be used for the ’month’ and ’day of week’ fields. Use the first three letters of the particular day or month (case does not matter). Ranges or lists of names are not allowed.

linux_os/guide/system/software/integrity/software-integrity/aide/aide_periodic_cron_checking/oval/shared.xml
```diff
@@ -29,7 +29,7 @@
   </ind:textfilecontent54_test>
   <ind:textfilecontent54_object comment="run aide with cron" id="object_test_aide_periodic_cron_checking" version="1">
     <ind:filepath>/etc/crontab</ind:filepath>
-    <ind:pattern operation="pattern match">^(([0-9]*[\s]*[0-9]*[\s]*\*[\s]*\*[\s]*\*)|@(hourly|daily|weekly|monthly))[\s]*root[\s]*/usr/sbin/aide[\s]*\-\-check.*$</ind:pattern>
+    <ind:pattern operation="pattern match">^(([0-9]*[\s]*[0-9]*[\s]*\*[\s]*\*[\s]*(\*|([0-7]|mon|tue|wed|thu|fri|sat|sun)|[0-7]-[0-7]))|@(hourly|daily|weekly))[\s]*root[\s]*/usr/sbin/aide[\s]*\-\-check.*$</ind:pattern>
```
jan-cerny: @vojtapolasek I think that `@hourly` shouldn't be allowed by the regex because  rule description says: `At a minimum, AIDE should be configured to run a weekly scan. At most, AIDE should be run daily.`

vojtapolasek: From what I understand, we are not actively preventing users from setting stricter parameters than suggested... Do I understand it wrongly?

jan-cerny: Aha, we usually do that, but then the "at most ... " part of the rule description shouldn't be there. :smile:

linux_os/guide/system/software/integrity/software-integrity/aide/aide_periodic_cron_checking/tests/crontab_weekly_on_exact_day.pass.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+# profiles = xccdf_org.ssgproject.content_profile_pci-dss
+
+echo '21    21    *    *    *    root    /usr/sbin/aide --check &>/dev/null' >> /etc/crontab
```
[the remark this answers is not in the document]
vojtapolasek: Thank you, fixed.

PR #4991, opened 2019-11-07
Add draft RHEL8 STIG profile

redhatrises:
- Content is extended from the OSPP profile with missing rules added for SRG submission

### DIFF
--- a/rhel8/profiles/stig.profile
+++ b/rhel8/profiles/stig.profile
@@ -1,0 +1,35 @@
+documentation_complete: true
+
+title: '[DRAFT] DISA STIG for Red Hat Enterprise Linux 8'
+
+description: |-
+    This profile contains configuration checks that align to the
+    [DRAFT] DISA STIG for Red Hat Enterprise Linux 8.
+
+    In addition to being applicable to Red Hat Enterprise Linux 8, DISA recognizes this
+    configuration baseline as applicable to the operating system tier of
+    Red Hat technologies that are based on Red Hat Enterprise Linux 8, such as:
+
+    - Red Hat Enterprise Linux Server
+    - Red Hat Enterprise Linux Workstation and Desktop
+    - Red Hat Enterprise Linux for HPC
+    - Red Hat Storage
+    - Red Hat Containers with a Red Hat Enterprise Linux 8 image
+
+extends: ospp
+
+selections:
+    - login_banner_text=dod_banners
+    - dconf_gnome_banner_enabled
+    - banner_etc_issue
+    - accounts_password_set_min_life_existing
+    - accounts_password_set_max_life_existing
+    - account_disable_post_pw_expiration
+    - account_temp_expire_date
+    - audit_rules_usergroup_modification_passwd
+    - sssd_enable_smartcards
+    - sssd_offline_cred_expiration
+    - smartcard_configure_cert_checking
+    - encrypt_partitions
+    - sysctl_net_ipv4_tcp_syncookies
+    - clean_components_post_updating


rhel8/profiles/stig.profile
```diff
@@ -0,0 +1,36 @@
...
+    [DRAFT] DISA STIG for Red Hat Enterprise Linux 8.
+
+    In addition to being applicable to Red Hat Enterprise Linux 8, DISA recognizes this
+    configuration baseline as applicable to the operating system tier of
+    Red Hat technologies that are based on Red Hat Enterprise Linux 8, such as:
+
+    - Red Hat Enterprise Linux Server
+    - Red Hat Enterprise Linux Workstation and Desktop
+    - Red Hat Enterprise Linux for HPC
+    - Red Hat Storage
```
shawndwells: How does this profile apply to Ceph?

shawndwells: e.g. by that same logic, this would apply to OpenShift.

rhel8/profiles/stig.profile
```diff
@@ -0,0 +1,36 @@
...
+
+selections:
+    - login_banner_text=dod_banners
+    - dconf_gnome_banner_enabled
+    - banner_etc_issue
+    - accounts_password_set_min_life_existing
+    - accounts_password_set_max_life_existing
+    - account_disable_post_pw_expiration
+    - account_temp_expire_date
+    - audit_rules_usergroup_modification_passwd
```
shawndwells: Should this be in NCP?

rhel8/profiles/stig.profile
```diff
@@ -0,0 +1,36 @@
...
+    - sssd_enable_smartcards
+    - sssd_offline_cred_expiration
+    - smartcard_configure_cert_checking
+    - encrypt_partitions
+    - sysctl_net_ipv4_tcp_syncookies
```
shawndwells: Should this be in NCP?

rhel8/profiles/stig.profile
```diff
@@ -0,0 +1,36 @@
...
+    - package_openssh-server_installed
```
shawndwells: NACK. Baselines should not mandate the installation of SSH.

rhel8/profiles/stig.profile
```diff
@@ -0,0 +1,36 @@
...
+    - clean_components_post_updating
```
shawndwells: Another NCP candidate?

rhel8/profiles/stig.profile
redhatrises: > How does this profile apply to Ceph?

@shawndwells because the el7 has the statement that it applies to `Red Hat Storage`, so logically, the same statement applies to el8

redhatrises: > e.g. by that same logic, this would apply to OpenShift.

Partially true, that is of course if the node is pure RHEL.

rhel8/profiles/stig.profile
redhatrises: A lot of our checks and remediations check if openssh is installed and install if it isn't. I understand your comment about not "requiring" ssh, but why not add the rule and have it tailored out for those that don't use ssh?

rhel8/profiles/stig.profile
redhatrises: @shawndwells there is no `ncp` profile yet for el8. Also the stig profile extends `ospp` profile and not `ncp` profile.

rhel8/profiles/stig.profile
redhatrises: @shawndwells there is no `ncp` profile yet for el8. Also the stig profile extends `ospp` profile and not `ncp` profile.

rhel8/profiles/stig.profile
redhatrises: @shawndwells there is no `ncp` profile yet for el8. Also the stig profile extends `ospp` profile and not `ncp` profile.

rhel8/profiles/stig.profile
redhatrises: Talked with @shawndwells offline and removed the rule.

shawndwells:
@redhatrises Ack pending jenkins. Will keep an eye out for Jenkins finishing later today.... but feel free to self-merge once all green.

shawndwells:
+cc @tedbrunell as FYI

PR #4992, opened 2019-11-07
Fix command module typo

redhatrises:
This was missed in the stabilization effort


jan-cerny:
This addresses warning ` [WARNING]: As of Ansible 2.4, the parameter 'executable' is no longer supported with the 'command' module. Not using '/bin/bash'.`

Thank you.

PR #4993, opened 2019-11-07
Use service_facts to get services in template_ANSIBLE_service_disabled

redhatrises:
- Rather than running the command module use the service_facts instead


matejak:
This PR avoids usage of the `command` module, as `ansible_facts.services["<systemd unit name>"]` is defined for individual services. The socket unit names are not exposed as ansible_facts, so that part had to stay.
I have verified that the code works both for existing and non-existing service names.

Merging, thank you for the PR!

PR #4994, opened 2019-11-08
Remove coreos-fedramp profile

redhatrises:
This profile should never have been added. There are multiple moderate profiles all based off of NIST. It was asked to be removed in a previous PR for this and other reasons.


redhatrises:
This is what the `coreos-ncp` profile is for btw. Per the description:

`NIST 800-53 control selections for MODERATE impact systems (NIST 800-53)`

PR #4996, opened 2019-11-11
add double quote for rule to prevent globbing and word splitting

tianzhenjia:
add double quote for $rule to prevent globbing and word splitting


tianzhenjia:

yes, ok to test

jan-cerny:
@openscap-ci ok to test

jan-cerny:
Thanks

PR #4999, opened 2019-11-11
Replace function ensure_there_are_servers_... by Jinja macro

jan-cerny:
#### Description:

Replace Bash remediation function ensure_there_are_servers_in_ntp_compatible_config_file by a Jinja macro.

#### Rationale:

This is another part of our effort to convert Bash functions to Jinja macros. Once all macros will be converted we will be able to remove the legacy code and confusing sourcing of non-existent file.


matejak:
Works great, thanks!

PR #5000, opened 2019-11-11
Enable copr builds and add packit config

dhodovsk:
Let us introduce [packit service](https://packit.dev) to you - the automation to integrate upstream open source projects into Fedora operating system.

After merging this PR, you are just a few steps away from RPM builds being automatically triggered on your PR's.
It means, that you'll be able to try and play with your change, packaged as an RPM.

But there is more. By using packit, you can for example enable adding new releases into Fedora Rawhide.

What are the next steps?
* Install [Packit-as-a-Service github application](https://github.com/marketplace/packit-as-a-service) in your repo
* In case you are first user, wait for account approval
* Enjoy the built RPMs!

For more info, please:
 * check out the documentation: https://packit.dev/docs/
 * contact @packit-service

The spec you see in this PR was fetched from your package's Fedora dist-git.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -12,7 +12,7 @@ jobs:
     - fedora-all-x86_64
 - job: propose_downstream
   metadata:
-  dist-git-branch: fedora-all
+    dist-git-branch: fedora-all
   trigger: release
 - job: sync_from_downstream
   trigger: commit
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,16 +1,18 @@
+upstream_package_name: content
 downstream_package_name: scap-security-guide
-jobs:
-- job: copr_build
-  metadata:
-    targets:
-    - fedora-30-x86_64
-    - fedora-31-x86_64
-    - fedora-rawhide-x86_64
-  trigger: pull_request
 specfile_path: scap-security-guide.spec
 synced_files:
-- scap-security-guide.spec
-- .packit.yaml
-upstream_package_name: content
-actions:
-  post-upstream-clone: wget -N https://src.fedoraproject.org/rpms/scap-security-guide/raw/master/f/scap-security-guide.spec
+  - scap-security-guide.spec
+  - .packit.yaml
+jobs:
+- job: copr_build
+  trigger: pull_request
+  metadata:
+    targets:
+    - fedora-all-x86_64
+- job: propose_downstream
+  metadata:
+  dist-git-branch: fedora-all
+  trigger: release
+- job: sync_from_downstream
+  trigger: commit
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,9 +1,8 @@
-upstream_package_name: content
 downstream_package_name: scap-security-guide
 specfile_path: scap-security-guide.spec
-synced_files:
-  - scap-security-guide.spec
-  - .packit.yaml
+actions:
+  post-upstream-clone:
+    - wget -N https://src.fedoraproject.org/rpms/scap-security-guide/raw/master/f/scap-security-guide-0.1.48-xml_parsing.patch
 jobs:
 - job: copr_build
   trigger: pull_request
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,0 +1,16 @@
+downstream_package_name: scap-security-guide
+jobs:
+- job: copr_build
+  metadata:
+    targets:
+    - fedora-30-x86_64
+    - fedora-31-x86_64
+    - fedora-rawhide-x86_64
+  trigger: pull_request
+specfile_path: scap-security-guide.spec
+synced_files:
+- scap-security-guide.spec
+- .packit.yaml
+upstream_package_name: content
+actions:
+  post-upstream-clone: wget -N https://src.fedoraproject.org/rpms/scap-security-guide/raw/master/f/scap-security-guide.spec
--- a/scap-security-guide.spec
+++ b/scap-security-guide.spec
@@ -1,0 +1,66 @@
+Name:		scap-security-guide
+Version:	0.1.47
+Release:	2%{?dist}
+Summary:	Security guidance and baselines in SCAP formats
+License:	BSD-3-Clause
+URL:		https://github.com/ComplianceAsCode/content/
+Source0:	https://github.com/ComplianceAsCode/content/releases/download/v%{version}/scap-security-guide-%{version}.tar.bz2
+BuildArch:	noarch
+
+Patch0:		scap-security-guide-0.1.48-xml_parsing.patch
+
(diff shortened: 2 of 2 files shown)


redhatrises:
@openscap-ci test this please

matejak:
The only thing  I don't like about this is the .spec file - why to have it here, as it is already [publicly accessible](https://src.fedoraproject.org/rpms/scap-security-guide/blob/master/f/scap-security-guide.spec)?

The packit documentation mentions that `synced_files` are "copied to the dist-git", but the documentation doesn't clearly say what it implies - does running packit mean that the Fedora spec file will be effectively moved from Fedora dist-git here to the upstream?

dahaic:
Hello @dhodovsk (or should I reference @packit-service ? :) ), can you take a look at what Matej posted above?

(And thank you for proactive PR, and congrats for having item with number 5000!)

redhatrises:
> The only thing I don't like about this is the .spec file - why to have it here, as it is already [publicly accessible](https://src.fedoraproject.org/rpms/scap-security-guide/blob/master/f/scap-security-guide.spec)?

The packit service handles the the Fedora workflow for you. The reason that the spec file is upstream and downstream is because in many repos that keep their spec file upstream to mininize their workload downstream (except for changelog) means that changes made downstream can be synced upstream.

> 
> The packit documentation mentions that `synced_files` are "copied to the dist-git", but the documentation doesn't clearly say what it implies - does running packit mean that the Fedora spec file will be effectively moved from Fedora dist-git here to the upstream?

It means that the Fedora spec file will be kept "in sync" between dist-git and repo.  https://packit.dev/docs/configuration/#more-examples-of-synced-files

dhodovsk:
@redhatrises is right :)

There is an option to use packit actions to always just get the spec from the dist-git before attempting the srpm build, so it doesn't have to be specified here. - similarly as in [this example](https://github.com/packit-service/hello-world/pull/60/files). Would you be interested in that?

matejak:
> There is an option to use packit actions to always just get the spec from the dist-git before attempting the srpm build, ...

This sounds good - having the spec file here would open the possibility of having it modified by accident or by contributors not fully aware of Fedora packaging guidelines.

dhodovsk:
The packit config is now changed, so before every build, it fetches newest spec from dist-git, instead of storing it in repo.

shawndwells:
> The only thing I don't like about this is the .spec file - why to have it here, as it is already [publicly accessible](https://src.fedoraproject.org/rpms/scap-security-guide/blob/master/f/scap-security-guide.spec)?
> 
As stewards of open source, it's important for projects that Red Hat participates in be upstream first.

Having build files buried/hidden in downstream places only encumbers the community. There seems no downside to bringing the .spec file upstream.

> The packit documentation mentions that `synced_files` are "copied to the dist-git", but the documentation doesn't clearly say what it implies - does running packit mean that the Fedora spec file will be effectively moved from Fedora dist-git here to the upstream?

redhatrises:
> > The only thing I don't like about this is the .spec file - why to have it here, as it is already [publicly accessible](https://src.fedoraproject.org/rpms/scap-security-guide/blob/master/f/scap-security-guide.spec)?
> 
> As stewards of open source, it's important for projects that Red Hat participates in be upstream first.
> 
> Having build files buried/hidden in downstream places only encumbers the community. There seems no downside to bringing the .spec file upstream.
> 
> > The packit documentation mentions that `synced_files` are "copied to the dist-git", but the documentation doesn't clearly say what it implies - does running packit mean that the Fedora spec file will be effectively moved from Fedora dist-git here to the upstream?

Especially as their are several in the community who are actual Fedora packagers.

matejak:
I have sent an e-mail to the mailing list, and based on the feedback, we conclude this in the first week of December.

matejak:
test this plesase

yuumasato:
Loved the testing RPMs per PR and the capability to propose updates to Fedora.

We used to have the `.spec` file in upstream. It was used to allow community to build test RPMs, when CPack came in, we removed the spec file from upstream.
One of the burdens was synchronizing and keeping it working, as in the actual dist-git.
If `Packit` can keep them in sync, I don't mind having it upstream.

@dhodovsk If we have the `.spec` in upstream and "syncing it downstream" to Fedora. Would it mean that we would have to update Changelog in upstream before release and `propose-downstream`?

redhatrises:
@openscap-ci test this please

.packit.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - .packit.yaml
+jobs:
+- job: copr_build
+  trigger: pull_request
+  metadata:
+    targets:
+    - fedora-all-x86_64
+- job: propose_downstream
+  metadata:
+  dist-git-branch: fedora-all
```
jpopelka:
```suggestion
    dist-git-branch: fedora-all
```

dhodovsk:
Sorry about the super late answer, here is an example of how the spec would change after propose -> https://src.fedoraproject.org/rpms/packit/pull-request/39#request_diff

yuumasato:
@openscap-ci test this please

scap-security-guide.spec
```diff
@@ -0,0 +1,66 @@
...
+
+%description	doc
+The %{name}-doc package contains HTML formatted documents containing
+hardening guidances that have been generated from XCCDF benchmarks
+present in %{name} package.
+
+%prep
+%setup -q
+%patch0 -p1
+mkdir build
```
ggbecker:
```suggestion
mkdir -p build
```

ggbecker:
I have pushed d4dcd60 and now the builds are passing. It was failing on creating the dir `build` because it already existed.

ggbecker:
@openscap-ci test this please

ggbecker:
> I have pushed [d4dcd60](https://github.com/ComplianceAsCode/content/commit/d4dcd60f9aaf2d55716716f9388b45d85dea34c3) and now the builds are passing. It was failing on creating the dir `build` because it already existed.

I think I fully understood what was the problem now.
The "scap-security-guide.spec" in the official fedora builds use the `tar.bz2` compressed files found under artifact releases and those are generated by running `cmake .. && make package_source` which in the end doesn't contain the `build` directory, so that's why the spec file creates it before compiling the code.

And Packit creates the compressed file by running something like: `git archive --output content-v0.1.47.tar.gz --prefix content-v0.1.47 HEAD` and then generates the source rpm using `content-v0.1.47.tar.gz`. This `tar.gz` file contains the `build` directory thus leaving us with the necessity of using `mkdir -p build` as in d4dcd60.

IMO, I think we should also update the spec files under fedora channels, because if in the future we update the one which is added by this PR, it will start to fail again.

shawndwells:
Closing this as abandoned PR. Perhaps someone will want to take this on in the future.

PR #5001, opened 2019-11-11
Make coreos-ncp.profile buildable

jhrozek:
It was suggested to use the coreos-ncp.profile as a basis for checks of
RHCOS nodes in an OCP4 environment.

Because the profile is incomplete, it was marked as
`documentation_complete:false`. So in order to build the content, the `--debug`
flag should be passed to `build_content`.

That didn't work, though. One reason was that one ipv6-related rule was missing
two required attributes, then the profile itself was referencing rules that do
not apply to the profile.

The two patches in this PR work around that, but the build still fails with:
```
[ 28%] Built target generate-internal-ocp4-linked-xccdf-oval-ocil.xml
Scanning dependencies of target generate-ssg-ocp4-ocil.xml
Scanning dependencies of target generate-ssg-ocp4-xccdf.xml
Scanning dependencies of target generate-ssg-ocp4-oval.xml
[ 28%] Built target generate-ssg-ocp4-cpe-dictionary.xml
[ 28%] [ocp4-content] generating ssg-ocp4-ocil.xml
[ 28%] [ocp4-content] generating ssg-ocp4-xccdf.xml
[ 28%] [ocp4-content] generating ssg-ocp4-oval.xml
[ 36%] Built target generate-ssg-ocp4-ocil.xml
[ 43%] Built target generate-ssg-ocp4-oval.xml
File '/home/jhrozek/devel/compliance-as-code-content/build/ssg-ocp4-xccdf.xml' line 193: Element '{http://checklists.nist.gov/xccdf/1.1}select': This element is not expected. Expected is one of ( {http://checklists.nist.gov/xccdf/1.1}status, {http://checklists.nist.gov/xccdf/1.1}version, {http://checklists.nist.gov/
xccdf/1.1}title ).
OpenSCAP Error: Invalid XCCDF Checklist (1.1) content in /home/jhrozek/devel/compliance-as-code-content/build/ssg-ocp4-xccdf.xml. [/builddir/build/BUILD/openscap-1.3.1/src/source/oscap_source.c:346]
make[2]: *** [ocp4/CMakeFiles/generate-ssg-ocp4-xccdf.xml.dir/build.make:67: ssg-ocp4-xccdf.xml] Error 1
make[2]: *** Deleting file 'ssg-ocp4-xccdf.xml'
make[1]: *** [CMakeFiles/Makefile2:1825: ocp4/CMakeFiles/generate-ssg-ocp4-xccdf.xml.dir/all] Error 2
make: *** [Makefile:151: all] Error 2
```

I'm not sure how to fix that, so I'm submitting the PR as a WIP for now,
hopefully we can iterate on it and get a working profile soon.

### DIFF
--- a/ocp4/profiles/coreos-ncp.profile
+++ b/ocp4/profiles/coreos-ncp.profile
@@ -38,6 +38,13 @@ selections:
     ### Things needed to meet OSPP functional requirements.
     #######################################################
 
+    ### SELinux Configuration
+    - var_selinux_state=enforcing
+    - selinux_state
+
+    ### Audit
+    - service_auditd_enabled
+
     ### Partitioning
     #- mount_option_home_nodev
     #- mount_option_home_nosuid
@@ -159,7 +166,6 @@ selections:
     #- sysctl_fs_protected_symlinks
 
     ### Audit
-    #- service_auditd_enabled
     #- var_auditd_flush=incremental_async
     #- auditd_data_retention_flush
     #- auditd_local_events
@@ -233,8 +239,6 @@ selections:
     #- accounts_password_pam_unix_remember
 
     ### SELinux Configuration
-    #- var_selinux_state=enforcing
-    #- selinux_state
     #- var_selinux_policy_name=targeted
     #- selinux_policytype
 
--- a/ocp4/profiles/coreos-ncp.profile
+++ b/ocp4/profiles/coreos-ncp.profile
@@ -56,216 +56,216 @@ selections:
     #- partition_for_home
     #- partition_for_var
     #- mount_option_var_nodev
-    - partition_for_var_log
+    #- partition_for_var_log
     #- mount_option_var_log_nodev
     #- mount_option_var_log_nosuid
     #- mount_option_var_log_noexec
-    - partition_for_var_log_audit
+    #- partition_for_var_log_audit
     #- mount_option_var_log_audit_nodev
     #- mount_option_var_log_audit_nosuid
     #- mount_option_var_log_audit_noexec
 
     ### Services
     # sshd
-    - sshd_disable_root_login
-    - sshd_enable_strictmodes
-    - disable_host_auth
-    - sshd_disable_empty_passwords
-    - sshd_disable_kerb_auth
-    - sshd_disable_gssapi_auth
-    - var_sshd_set_keepalive=0
-    - sshd_set_keepalive
-    - sshd_enable_warning_banner
-    - sshd_rekey_limit
+    #- sshd_disable_root_login
+    #- sshd_enable_strictmodes
+    #- disable_host_auth
+    #- sshd_disable_empty_passwords
+    #- sshd_disable_kerb_auth
+    #- sshd_disable_gssapi_auth
+    #- var_sshd_set_keepalive=0
+    #- sshd_set_keepalive
+    #- sshd_enable_warning_banner
+    #- sshd_rekey_limit
 
     # Time Server
-    - chronyd_client_only
-    - chronyd_no_chronyc_network
+    # - chronyd_client_only
+    # - chronyd_no_chronyc_network
 
     ### Network Settings
-    - sysctl_net_ipv6_conf_all_accept_ra
-    - sysctl_net_ipv6_conf_default_accept_ra
-    - sysctl_net_ipv4_conf_all_accept_redirects
-    - sysctl_net_ipv4_conf_default_accept_redirects
-    - sysctl_net_ipv6_conf_all_accept_redirects
-    - sysctl_net_ipv6_conf_default_accept_redirects
-    - sysctl_net_ipv4_conf_all_accept_source_route
-    - sysctl_net_ipv4_conf_default_accept_source_route
-    - sysctl_net_ipv6_conf_all_accept_source_route
-    - sysctl_net_ipv6_conf_default_accept_source_route
-    - sysctl_net_ipv4_conf_all_secure_redirects
-    - sysctl_net_ipv4_conf_default_secure_redirects
-    - sysctl_net_ipv4_conf_all_send_redirects
-    - sysctl_net_ipv4_conf_default_send_redirects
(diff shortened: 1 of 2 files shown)


jhrozek:
CC @shawndwells @redhatrises @evgenyz

redhatrises:
> CC @shawndwells @redhatrises @evgenyz
> then the profile itself was referencing rules that do
not apply to the profile.

That's not true. Many of the capabilities exist in CoreOS and therefore apply to CoreOS. This includes rules like SSH configuration as well as password requirements handling, audit rule configuration, audit rule offloading, firewall enablement and configuration, FIPS checking and validation, etc. Having no rules in CoreOS profile, will prevent a build.

jan-cerny:
@redhatrises I believe @jhrozek references to the error messages such as `ValueError: Rule 'sysctl_net_ipv6_conf_all_accept_ra' was not found in the benchmark. Please remove rule 'sysctl_net_ipv6_conf_all_accept_ra' from profile 'coreos-ncp' before proceeding.` They can be solved by adding `ocp4` to `prodtype` keys.

jan-cerny:
Hi @jhrozek , there is a problem that rules specify in their `rule.yml` file on `prodtype:` line the "products" that they are part of. "Product" means operating system or layered product.
eg.
https://github.com/ComplianceAsCode/content/blob/d12afbb3d1a03addd5111a2592067ac472af551c/linux_os/guide/services/ntp/chronyd_client_only/rule.yml#L3
It's a comma-separated list of products. You need to add `ocp4` to this `prodtype` list for all rules in your profile. I suggest doing that by a script.

Note If a `rule.yml` doesn't have `prodtype:` key, it means the rule is part of every product.
The name of the directory that `rule.yml` is located in is always the same as the rule ID that you have in the profile.

jhrozek:
On 11/12/19 8:44 AM, Jan Černý wrote:
>
> @redhatrises <https://github.com/redhatrises> I believe @jhrozek 
> <https://github.com/jhrozek> references to the error messages such as 
> |ValueError: Rule 'sysctl_net_ipv6_conf_all_accept_ra' was not found 
> in the benchmark. Please remove rule 
> 'sysctl_net_ipv6_conf_all_accept_ra' from profile 'coreos-ncp' before 
> proceeding.| They can be solved by adding |ocp4| to |prodtype| keys.
>
Yes and no. Yes, I was getting the error initially, but I worked around 
that by commenting out the rules, because I don't think that blindly 
adding rules without verifying that they should be applied for RHCOS is 
the best thing to do. Then I was getting a different error because they 
were no rules..oops.

Now the content builds with the latest commits I pushed.

Can we merge them, please?

jhrozek:
On 11/12/19 12:57 AM, Gabe Alford wrote:
>
> That's not true. Many of the capabilities exist in CoreOS and 
> therefore apply to CoreOS. This includes rules like SSH configuration 
> as well as password requirements handling, audit rule configuration, 
> audit rule offloading, firewall enablement and configuration, FIPS 
> checking and validation, etc. Having no rules in CoreOS profile, will 
> prevent a build.
>
True, the latest commit re-adds the `selinux_state` rule, so that the 
content builds.

But having all the rules stuffed into the content is not a good idea 
either, because a) currently they don't apply for this profile, we need 
to add the `ocp4` prodtype and b) While RHCOS is kind-of sort-of RHEL, 
it's really tailored to a specific use-case and some rules that are 
currently in the profile do not apply. Just picking some from the top: 
The rules to have a separate partition for /home or /var do not apply 
because currently there is no easy way to repartition the RHCOS install 
(you can use a separate /disk/ but not a separate /partition/).

Mostly for the b) reason I commented out the rules and would like to go 
through them one-by-one to see which of them should stay. I explicitly 
didn't remove the rules, because I think the work that was done to put 
the rules in the profile in the first place can be valuable. But each 
rule should be evaluated and tested.

jan-cerny:
@jhrozek Thanks for explanation. In other words we need to evaluate every rule in the profile individually whether the rule makes sense in CoreOS context. Then I think your patch is fine and in future PRs people can either:
* remove the rules from the profile
* uncomment them and add prodtype to rule.yml

Waiting for CI.

jan-cerny:
@openscap-ci test this please

jhrozek:
On 11/12/19 9:11 AM, Jan Černý wrote:
>
> @jhrozek <https://github.com/jhrozek> Thanks for explanation. In other 
> words we need to evaluate every rule in the profile individually 
> whether the rule makes sense in CoreOS context. Then I think your 
> patch is fine and in future PRs people can either:
>
>   * remove the rules from the profile
>   * uncomment them and add prodtype to rule.yml
>

Right. (There's more to that, we'll probably want to also run some 
sanity tests to see if the rules don't break the functionality, but yeah)

jan-cerny:
I have tried the debug build  and it has built successfully. I think that we can merge it because the profile is `documentation_complete: false` anyway.

ocp4/profiles/coreos-ncp.profile
```diff
@@ -56,216 +60,210 @@ selections:
...
-    - sshd_disable_empty_passwords
-    - sshd_disable_kerb_auth
-    - sshd_disable_gssapi_auth
-    - var_sshd_set_keepalive=0
-    - sshd_set_keepalive
-    - sshd_enable_warning_banner
-    - sshd_rekey_limit
-
-    # Time Server
-    - chronyd_client_only
```
jan-cerny: @jhrozek Have the chronyd_ rules disappeared by purpose?

jhrozek: I re-added them, thanks for catching that.

jhrozek:
OK, new patches pushed. I also added the auditd service check because for some reason the SELinux check returns result=notchecked. That's something to fix, but the audit check works fine.

jan-cerny:
@openscap-ci add to whitelist

jan-cerny:
@openscap-ci test this please

jan-cerny:
Great, we are looking forward to further contributions.

PR #5002, opened 2019-11-12
Small Bash style fixes.

matejak:
- Use explicit curly braces when deferencing a variable when it is followed by square brackets.
- Removed ls | grep pattern in favor of wildcard expression, fixing the remediation in the process.
- Added robust [backslash quoting](https://www.gnu.org/software/bash/manual/bash.html#Double-Quotes) inside of double quotes.


redhatrises:
Nice! Looks good to me unless someone else has something.

PR #5004, opened 2019-11-13
Ansible tasks fixes

jan-cerny:
#### Description:
Fixes for RHEL 7 STIG profile Ansible Playbook.

For more information please read the commit messages of all commits.

#### Rationale:
Ansible playbook terminated before finishing all tasks due to errors.


redhatrises:
Looks good so far!

jan-cerny:
I still have a problem:

```
TASK [Add Kerberos security to nfs and nfs4 mount points] ***********************************************
fatal: [localhost]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: list object has no element 2\n\nThe error appears to be in '/root/rhel7-playbook-stig.yml': line 125
76, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n        - DISA-STIG-RHEL-07-040750\n    - name: Add Kerberos security to nfs and nfs
4 mount points\n      ^ here\n"}

```

ggbecker:
> I still have a problem:
> 
> ```
> TASK [Add Kerberos security to nfs and nfs4 mount points] ***********************************************
> fatal: [localhost]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: list object has no element 2\n\nThe error appears to be in '/root/rhel7-playbook-stig.yml': line 125
> 76, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n        - DISA-STIG-RHEL-07-040750\n    - name: Add Kerberos security to nfs and nfs
> 4 mount points\n      ^ here\n"}
> ```

It looks like the
```
  with_items:
    - "{{ points_register.stdout_lines }}"
```
got removed and there is no item to be consumed. From here:
https://github.com/ComplianceAsCode/content/commit/e268f1e07192a5cf343b6ac36053553d1074bd3b#diff-99cb1a68439a2c5b9929ff397a410194L21

ggbecker:
I think it needs to be extended the line which checks for the mount options. Something like that: `findmnt --fstab --types nfs,nfs4 -O nosec=krb5:krb5i:krb5p -n -o TARGET,SOURCE,FSTYPE,OPTIONS`

redhatrises:
> I think it needs to be extended the line which checks for the mount options. Something like that: `findmnt --fstab --types nfs,nfs4 -O nosec=krb5:krb5i:krb5p -n -o TARGET,SOURCE,FSTYPE,OPTIONS`

You fix is probably adding something like `when: points_register is defined`

ggbecker:
> > I think it needs to be extended the line which checks for the mount options. Something like that: `findmnt --fstab --types nfs,nfs4 -O nosec=krb5:krb5i:krb5p -n -o TARGET,SOURCE,FSTYPE,OPTIONS`
> 
> You fix is probably adding something like `when: points_register is defined` to the `name: "Add {{{ MOUNTOPTION }}} to nfs and nfs4 mount points"` task.

I see the problem now. The template seems to be ok. But the rule which is still failing is not using the template at all. And the `shared.yml` from this rule `mount_option_krb_sec_remote_filesystems` seems to be outdated.

PR #5009, opened 2019-11-14
ComplianceAsCode:fix fragile for loops

tianzhenjia:
For loops over find output are fragile.
Use $(..) instead of legacy `..`.

### DIFF
--- a/tests/ansible_playbook_check.sh
+++ b/tests/ansible_playbook_check.sh
@@ -3,7 +3,7 @@
 pushd "$2" > /dev/null
 
 DIRS="" # directories which might contain playbooks
-for dir in `find . -type d`
+for dir in {$(find . -type d)}
 do
 	# tries to find which directories contain valid playbooks
 	FILES=$(ls $dir/*.yml 2> /dev/null | wc -l)
@@ -12,7 +12,6 @@ do
 		DIRS="$DIRS $dir/*.yml"
 	fi
 done
-
 if [ -z "$CONTAINS_VALID_FILES" ]; then
 	echo "$2 does not contain any valid YAML files. Skipping the test."
 	popd > /dev/null


tianzhenjia:

yes

redhatrises:
@openscap-ci test this please

matejak:
Thank you for your contribution, to be honest though, I have never seen curly braces used like this with for. Could you please point out some references of this usage? This prominent resource doesn't mention this pattern: https://stackoverflow.com/questions/9612090/how-to-loop-through-file-names-returned-by-find

jan-cerny:
NACK. This adds superfluous braces:

```
[jcerny@thinkpad sshd_use_approved_macs{master}]$ ls
ansible  bash  oval  rule.yml
[jcerny@thinkpad sshd_use_approved_macs{master}]$ for x in {$(find . -type d)} ; do echo $x ; done
{.
./bash
./oval
./ansible}
[jcerny@thinkpad sshd_use_approved_macs{master}]$ 
```

tianzhenjia:
> Thank you for your contribution, to be honest though, I have never seen curly braces used like this with for. Could you please point out some references of this usage? This prominent resource doesn't mention this pattern: https://stackoverflow.com/questions/9612090/how-to-loop-through-file-names-returned-by-find

tianzhenjia:
Can I ask a question? 
`for dir in '`find . -type d``
this line of code will run "find . -type d" first, and then run for loop. The result is shown:
image
Can we white like this？
```
directory=$(find . -type d)
for dir in "$directory"
```

matejak:
These two approaches are the same. The only slight improvement to this loop I can think of would be to use `readarray` to capture `find` findings to a Bash array.

However, this test code will work just fine if the path doesn't contain spaces in file/directory names.

PR #5018, opened 2019-11-14
Coreos build - enable more rules

redhatrises:
#### Description:

- Enable audit rules and audit configuration checks
- Enable sysctl and kernel checks
- Certain software has to be installed and running
- Time has to be configured in chrony

### DIFF
--- a/linux_os/guide/system/network/network-kernel/network_host_parameters/sysctl_net_ipv4_ip_forward/rule.yml
+++ b/linux_os/guide/system/network/network-kernel/network_host_parameters/sysctl_net_ipv4_ip_forward/rule.yml
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: wrlinux1019,rhel6,rhel7,rhel8,fedora,ol7,ol8,rhv4
+prodtype: wrlinux1019,rhel6,rhel7,rhel8,fedora,ol7,ol8,rhv4,ocp4
 
 title: 'Disable Kernel Parameter for IP Forwarding on IPv4 Interfaces'
 
--- a/linux_os/guide/system/network/network-kernel/network_host_parameters/sysctl_net_ipv4_ip_forward/rule.yml
+++ b/linux_os/guide/system/network/network-kernel/network_host_parameters/sysctl_net_ipv4_ip_forward/rule.yml
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: wrlinux1019,rhel6,rhel7,rhel8,fedora,ol7,ol8,rhv4,ocp4
+prodtype: wrlinux1019,rhel6,rhel7,rhel8,fedora,ol7,ol8,rhv4
 
 title: 'Disable Kernel Parameter for IP Forwarding on IPv4 Interfaces'
 
@@ -39,6 +39,12 @@ references:
 ocil: |-
     {{{ ocil_sysctl_option_value(sysctl="net.ipv4.ip_forward", value="0") }}}
     The ability to forward packets is only appropriate for routers.
+
+warnings:
+    - functionality: |-
+        Certain technologies such as virtual machines, containers, etc. rely on IPv4 forwarding to enable and use networking.
+        Disabling IPv4 forwarding would cause those technologies to stop working. Therefore, this rule should not be used in
+        profiles or benchmarks that target usage of IPv4 forwarding.
 
 platform: machine
 
--- a/ocp4/profiles/coreos-ncp.profile
+++ b/ocp4/profiles/coreos-ncp.profile
@@ -1,4 +1,4 @@
-documentation_complete: false
+documentation_complete: true
 
 title: 'NIST National Checklist for Red Hat Enterprise Linux CoreOS'
 
@@ -37,13 +37,6 @@ selections:
     ### GENERAL REQUIREMENTS
     ### Things needed to meet OSPP functional requirements.
     #######################################################
-
-    ### SELinux Configuration
-    - var_selinux_state=enforcing
-    - selinux_state
-
-    ### Audit
-    - service_auditd_enabled
 
     ### Partitioning
     #- mount_option_home_nodev
@@ -86,37 +79,37 @@ selections:
     #- sshd_rekey_limit
 
     # Time Server
-    # - chronyd_client_only
-    # - chronyd_no_chronyc_network
+    - chronyd_client_only
+    - chronyd_no_chronyc_network
 
     ### Network Settings
-    #- sysctl_net_ipv6_conf_all_accept_ra
-    #- sysctl_net_ipv6_conf_default_accept_ra
-    #- sysctl_net_ipv4_conf_all_accept_redirects
-    #- sysctl_net_ipv4_conf_default_accept_redirects
-    #- sysctl_net_ipv6_conf_all_accept_redirects
-    #- sysctl_net_ipv6_conf_default_accept_redirects
-    #- sysctl_net_ipv4_conf_all_accept_source_route
-    #- sysctl_net_ipv4_conf_default_accept_source_route
-    #- sysctl_net_ipv6_conf_all_accept_source_route
-    #- sysctl_net_ipv6_conf_default_accept_source_route
-    #- sysctl_net_ipv4_conf_all_secure_redirects
-    #- sysctl_net_ipv4_conf_default_secure_redirects
-    #- sysctl_net_ipv4_conf_all_send_redirects
-    #- sysctl_net_ipv4_conf_default_send_redirects
-    #- sysctl_net_ipv4_conf_all_log_martians
-    #- sysctl_net_ipv4_conf_default_log_martians
-    #- sysctl_net_ipv4_conf_all_rp_filter
-    #- sysctl_net_ipv4_conf_default_rp_filter
-    #- sysctl_net_ipv4_icmp_ignore_bogus_error_responses
-    #- sysctl_net_ipv4_icmp_echo_ignore_broadcasts
-    #- sysctl_net_ipv4_ip_forward
-    #- sysctl_net_ipv4_tcp_syncookies
+    - sysctl_net_ipv6_conf_all_accept_ra
+    - sysctl_net_ipv6_conf_default_accept_ra
+    - sysctl_net_ipv4_conf_all_accept_redirects
+    - sysctl_net_ipv4_conf_default_accept_redirects
+    - sysctl_net_ipv6_conf_all_accept_redirects
+    - sysctl_net_ipv6_conf_default_accept_redirects
+    - sysctl_net_ipv4_conf_all_accept_source_route
+    - sysctl_net_ipv4_conf_default_accept_source_route
+    - sysctl_net_ipv6_conf_all_accept_source_route
+    - sysctl_net_ipv6_conf_default_accept_source_route
+    - sysctl_net_ipv4_conf_all_secure_redirects
+    - sysctl_net_ipv4_conf_default_secure_redirects
+    - sysctl_net_ipv4_conf_all_send_redirects
+    - sysctl_net_ipv4_conf_default_send_redirects
+    - sysctl_net_ipv4_conf_all_log_martians
+    - sysctl_net_ipv4_conf_default_log_martians
+    - sysctl_net_ipv4_conf_all_rp_filter
+    - sysctl_net_ipv4_conf_default_rp_filter
+    - sysctl_net_ipv4_icmp_ignore_bogus_error_responses
+    - sysctl_net_ipv4_icmp_echo_ignore_broadcasts
+    - sysctl_net_ipv4_ip_forward
+    - sysctl_net_ipv4_tcp_syncookies
 
     ### systemd
-    #- disable_ctrlaltdel_reboot
-    #- disable_ctrlaltdel_burstaction
-    #- service_debug-shell_disabled
+    - disable_ctrlaltdel_reboot
+    - disable_ctrlaltdel_burstaction
+    - service_debug-shell_disabled
 
     ### umask
     #- var_accounts_user_umask=027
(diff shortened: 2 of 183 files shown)


ocp4/profiles/coreos-ncp.profile
```diff
@@ -86,37 +79,37 @@ selections:
...
+    - sysctl_net_ipv4_conf_default_secure_redirects
+    - sysctl_net_ipv4_conf_all_send_redirects
+    - sysctl_net_ipv4_conf_default_send_redirects
+    - sysctl_net_ipv4_conf_all_log_martians
+    - sysctl_net_ipv4_conf_default_log_martians
+    - sysctl_net_ipv4_conf_all_rp_filter
+    - sysctl_net_ipv4_conf_default_rp_filter
+    - sysctl_net_ipv4_icmp_ignore_bogus_error_responses
+    - sysctl_net_ipv4_icmp_echo_ignore_broadcasts
+    - sysctl_net_ipv4_ip_forward
```
jhrozek: Are you sure it is safe to disable forwarding? All the pods in OCP are NAT-ed, right?

ocp4/profiles/coreos-ncp.profile
```diff
@@ -140,75 +133,75 @@ selections:
...
-    #- grub2_page_poison_argument
-    #- grub2_vsyscall_argument
-    #- grub2_vsyscall_argument.role=unscored
-    #- grub2_vsyscall_argument.severity=info
-    #- grub2_pti_argument
+    - grub2_audit_argument
+    - grub2_audit_backlog_limit_argument
+    - grub2_slub_debug_argument
+    - grub2_page_poison_argument
+    - grub2_vsyscall_argument
```
jhrozek: I understand the hardening potential of this setting, but IIRC vsyscalls were used in RHEL-6 and per Ben Briard, this is still a use-case we need to keep in mind. I'm not sure if can, for the purpose of the certification say "too bad, can't use this". If we can, then fine.

ocp4/profiles/coreos-ncp.profile
```diff
@@ -140,75 +133,75 @@ selections:
...
-    #- sysctl_kernel_unprivileged_bpf_disabled
-    #- sysctl_net_core_bpf_jit_harden
+    - sysctl_kernel_kptr_restrict
+    - sysctl_kernel_dmesg_restrict
+    - sysctl_kernel_kexec_load_disabled
+    - sysctl_kernel_yama_ptrace_scope
+    - sysctl_kernel_perf_event_paranoid
+    - sysctl_user_max_user_namespaces
+    - sysctl_user_max_user_namespaces.role=unscored
+    - sysctl_user_max_user_namespaces.severity=info
```
jhrozek: Does this rule make sense for RHCOS since its only use-case is to run containers?

ocp4/profiles/coreos-ncp.profile
```diff
@@ -140,75 +133,75 @@ selections:
...
-
+    - service_auditd_enabled
+    - var_auditd_flush=incremental_async
+    - auditd_data_retention_flush
+    - auditd_local_events
+    - auditd_write_logs
+    - auditd_log_format
+    - auditd_freq
+    - auditd_name_format
+    - var_auditd_action_mail_acct=root
```
jhrozek: Is it reasonable to expect that root will read their mail on a cluster node?

ocp4/profiles/coreos-ncp.profile
```diff
@@ -140,75 +133,75 @@ selections:
...
+    - var_auditd_space_left_action=email
+    - auditd_audispd_configure_remote_server
```
jhrozek: Note that with OCP, we'll be using the cluster-logging-operator to forward the logs off the cluster.

ocp4/profiles/coreos-ncp.profile
```diff
@@ -140,75 +133,75 @@ selections:
...
+    - package_sudo_installed
+    - package_usbguard_installed
+    - package_audispd-plugins_installed
     #- package_scap-security-guide_installed
-    #- package_audit_installed
-    #- package_libreswan_installed
-    #- package_rsyslog_installed
-    #- package_rsyslog-gnutls_installed
+    - package_audit_installed
+    - package_libreswan_installed
```
jhrozek: Why is libreswan needed?

ocp4/profiles/coreos-ncp.profile
```diff
@@ -140,75 +133,75 @@ selections:
...
+    - package_rsyslog_installed
```
jhrozek: Why is rsyslog needed? Cluster-logging-operator uses fluentd.

ocp4/profiles/coreos-ncp.profile
```diff
@@ -227,49 +220,51 @@ selections:
...
+    - var_selinux_policy_name=targeted
+    - selinux_policytype
 
     ### Application Whitelisting (RHEL 8)
-    #- package_fapolicyd_installed
-    #- service_fapolicyd_enabled
-    #- configure_fapolicyd_mounts
+    - package_fapolicyd_installed
+    - service_fapolicyd_enabled
+    - configure_fapolicyd_mounts
```
jhrozek: Very interesting. I wonder if a) anyone has tested fapolicyd with OCP and b) if fapolicyd should run on the host or in a deamonset pods. Something to think about and test I guess..

ocp4/profiles/coreos-ncp.profile
```diff
@@ -227,49 +220,51 @@ selections:
...
 
     ### Enable the Hardware RNG Entropy Gatherer Service
-    #- service_rngd_enabled
+    - service_rngd_enabled
```
jhrozek: This is correct and needed now, but in future (with newer kernels or in RHEL with backports), I understand that rngd won't be needed anymore,see https://github.com/openshift/machine-config-operator/issues/854 and https://gitlab.cee.redhat.com/coreos/redhat-coreos/merge_requests/685

ocp4/profiles/coreos-ncp.profile
```diff
@@ -227,49 +220,51 @@ selections:
...
-    #- configure_openssl_crypto_policy
-    #- configure_libreswan_crypto_policy
-    #- configure_kerberos_crypto_policy
-    #- enable_dracut_fips_module
+    - enable_fips_mode
+    - var_system_crypto_policy=fips
+    - configure_crypto_policy
+    - harden_sshd_crypto_policy
+    - harden_ssh_client_crypto_policy
+    - configure_bind_crypto_policy
```
jhrozek: I really hope nobody would ever run BIND on RHCOS directly :-)

ocp4/profiles/coreos-ncp.profile
jhrozek: Actually, upon reading the rule, it seems just informational?

ocp4/profiles/coreos-ncp.profile
shawndwells: ``vsyscall`` is already disabled in other government profiles, such as the DoD baselines for RHEL.

Having this rule in the ComplianceAsCode catalog allows for per-profile selection. For example the Government baselines may require this, and accept the implications, however more general-purpose baselines like PCI-DSS the rule may not need to be selected.

ocp4/profiles/coreos-ncp.profile
shawndwells: If a configuration option could be set then it must be evaluated, regardless of the intended product use case.

ocp4/profiles/coreos-ncp.profile
shawndwells: Good question! Generally, root email is aliased/forwarded to a distribution group (e.g. cluster-admins@customer.com). For the Government baselines this is expected and accepted. But something to revisit for the commercial profiles.

ocp4/profiles/coreos-ncp.profile
shawndwells: Showing my lack of product knowledge here - does that operator also pickup the CoreOS logs?

jhrozek:
> On 15 Nov 2019, at 17:48, Shawn Wells <notifications@github.com> wrote:
> 
> @shawndwells commented on this pull request.
> 
> In ocp4/profiles/coreos-ncp.profile:
> 
> > -    #- auditd_audispd_configure_remote_server
> -    #- auditd_audispd_encrypt_sent_records
> -    #- auditd_audispd_disk_full_action
> -    #- auditd_audispd_network_failure_action
> -
> +    - service_auditd_enabled
> +    - var_auditd_flush=incremental_async
> +    - auditd_data_retention_flush
> +    - auditd_local_events
> +    - auditd_write_logs
> +    - auditd_log_format
> +    - auditd_freq
> +    - auditd_name_format
> +    - var_auditd_action_mail_acct=root
> +    - var_auditd_space_left_action=email
> +    - auditd_audispd_configure_remote_server
> 
> Showing my lack of product knowledge here - does that operator also pickup the CoreOS logs?

If you mean `/var/log/audit/audit.log` then yes. Otherwise, I’d have to look up the answer.

jhrozek:
> On 15 Nov 2019, at 17:42, Shawn Wells <notifications@github.com> wrote:
> 
> @shawndwells commented on this pull request.
> 
> In ocp4/profiles/coreos-ncp.profile:
> 
> > @@ -140,75 +133,75 @@ selections:
>  
>      ### Kernel Config
>      ## Boot prompt
> -    #- grub2_audit_argument
> -    #- grub2_audit_backlog_limit_argument
> -    #- grub2_slub_debug_argument
> -    #- grub2_page_poison_argument
> -    #- grub2_vsyscall_argument
> -    #- grub2_vsyscall_argument.role=unscored
> -    #- grub2_vsyscall_argument.severity=info
> -    #- grub2_pti_argument
> +    - grub2_audit_argument
> +    - grub2_audit_backlog_limit_argument
> +    - grub2_slub_debug_argument
> +    - grub2_page_poison_argument
> +    - grub2_vsyscall_argument
> 
> vsyscall is already disabled in other government profiles, such as the DoD baselines for RHEL.
> 
> Having this rule in the ComplianceAsCode catalog allows for per-profile selection. For example the Government baselines may require this, and accept the implications, however more general-purpose baselines like PCI-DSS the rule may not need to be selected.

If it’s possible to opt-out in the unlikely case a customer would want to run a RHEL-6 container on an OCP4 cluster, then it’s fine. We might want to make this clear in the rule description, though. (Not as part of this PR of course)

ocp4/profiles/coreos-ncp.profile
JAORMX: This won't work with OpenShift, the kubelet and kubeproxy won't work without ip forwarding enabled.

ocp4/profiles/coreos-ncp.profile
JAORMX: Right, it's fine to disable it in this profile. It's expected.

ocp4/profiles/coreos-ncp.profile
JAORMX: The operator doesn't do it. We rely on the Cluster Logging Operator. audispd is not installed in CoreOS.

ocp4/profiles/coreos-ncp.profile
JAORMX: it's not installed nor available in CoreOS

ocp4/profiles/coreos-ncp.profile
JAORMX: yes

ocp4/profiles/coreos-ncp.profile
JAORMX: If this is required by a control, we gotta bring it up to the coreOS team and request the daemon.

ocp4/profiles/coreos-ncp.profile
JAORMX: they don't. Also, Kubernetes comes with its own DNS service (CoreDNS), which runs on a container.

ocp4/profiles/coreos-ncp.profile
redhatrises: This is required functionality... It will also have to be expanded to containers and the openshift env.... but that is for another time.

ocp4/profiles/coreos-ncp.profile
redhatrises: @jhrozek @JAORMX is lfuentd using FIPS libraries and encryptions and is it realtime off loading?

ocp4/profiles/coreos-ncp.profile
redhatrises: This is going to be required for CoreOS to do IPSec.

ocp4/profiles/coreos-ncp.profile
redhatrises: @jhrozek @JAORMX is the cluster operator using FIPS libraries and encryptions, and is it realtime off loading?

ocp4/profiles/coreos-ncp.profile
redhatrises: Great! Good feedback on this control. Will make a comment on it.

ocp4/profiles/coreos-ncp.profile
JAORMX: It's real time off loading, I'll get back to you on the FIPS part

ocp4/profiles/coreos-ncp.profile
JAORMX: What control requires Ipsec? Isn't the usage of TLS everywhere enough?

ocp4/profiles/coreos-ncp.profile
JAORMX: Fluentd is what the cluster-logging-operator uses to send logs out. It's real time off loading. I'll get back to you on the FIPS part

ocp4/profiles/coreos-ncp.profile
JAORMX: Fluentd uses OpenSSL underneath, so it should be FIPS compliant.

redhatrises: @JAORMX one more question I forgot to ask. How is fluentd getting the audit logs from a node?

ocp4/profiles/coreos-ncp.profile
redhatrises: @JAORMX it is part of AC-4 and others. IPSec is needed at the network layer which TLS doesn't touch.

JAORMX: Can you elaborate?

ocp4/profiles/coreos-ncp.profile
JAORMX: Fluentd runs in a pod on the node and reads the file. It subsequently forwards whatever is written to that file. It asistió keeps track of the last position it read.

ocp4/profiles/coreos-ncp.profile
JAORMX: @redhatrises could you create a Jira for us to start working with the CoreOS team to include this in the distribution?

ocp4/profiles/coreos-ncp.profile
shawndwells: @redhatrises if OpenShift comes with its own DNS service, would all the DNS SRG checks be applicable?

JAORMX: @shawndwells @redhatrises are those checks applicable for DNS that apply to internal services? CoreDNS doesn't resolve external domains. For that, the deployment relies on whatever is configured for the private network. Which means for AWS, they might rely on Route53, or in a private cloud, whatever DNS nameserver is configured by the deployer.

ocp4/profiles/coreos-ncp.profile
redhatrises: @JAORMX yes. Can do. This is part of the combined team assessing CoreOS/OCP.

ocp4/profiles/coreos-ncp.profile
redhatrises: @shawndwells @JAORMX Let's not have this conversation here. Let's have this separately, but now we need to have that.

linux_os/guide/system/network/network-kernel/network_host_parameters/sysctl_net_ipv4_ip_forward/rule.yml
```diff
@@ -40,6 +40,12 @@ ocil: |-
     {{{ ocil_sysctl_option_value(sysctl="net.ipv4.ip_forward", value="0") }}}
     The ability to forward packets is only appropriate for routers.
 
+warnings:
+    - functionality: |-
+        Certain technologies such as virtual machines, containers, etc. rely on IPv4 forwarding to enable and use networking.
+        Disabling IPv4 forwarding would cause those technologies to stop working. Therefore, this rule should not be used in
+        profiles or benchmarks that target usage of IPv4 forwarding.
```
shawndwells: Thanks for creating this warning. Should be very helpful for those tailoring or justifying dropping this rule locally based on how the endpoint is being used.

PR #5021, opened 2019-11-18
Add missing CCEs to rules from STIG profile

mildas:
#### Description:
Few rules from STIG profile were missing CCE for RHEL8.


shawndwells:
Thanks @mildas

PR #5023, opened 2019-11-19
Add waiting for RUNNING state of virtual machine

mildas:
#### Description:
Add 30sec waiting for virtual machine.

#### Rationale:
When running SSGTS multiple times in row, it can happen that virtual machine is in PAUSED state (and it's reverting snapshot to previous machine state) longer than expected. But a new run can be started (previous SSGTS process is finished) and it will crash, because it can't determine machine IP when machine is in PAUSED state.

If the state isn't still RUNNING after 30sec., then just try get IP and throw error, if it's not possible.

### DIFF
--- a/tests/ssg_test_suite/virt.py
+++ b/tests/ssg_test_suite/virt.py
@@ -107,6 +107,14 @@ def determine_ip(domain):
                        "          state='connected'/>"
                        "</channel>")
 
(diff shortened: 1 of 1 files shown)


matusmarhefka:
LGTM, thanks @mildas for fixing this, it was sometimes causing ugly race conditions in automated tests.

PR #5024, opened 2019-11-21
Updated the e8 profile for RHEL8.

matejak:
- removed obsolete `sshd` settings.
- added rules for crypto policies to set up strong ciphers + MACs for `sshd`.

### DIFF
--- a/rhel8/profiles/e8.profile
+++ b/rhel8/profiles/e8.profile
@@ -123,13 +123,15 @@ selections:
   - sshd_print_last_log
   - sshd_use_priv_separation
   - sshd_do_not_permit_user_env
-  - sshd_disable_rhosts_rsa
   - sshd_disable_rhosts
-  - sshd_allow_only_protocol2
   - sshd_set_loglevel_info
   - sshd_disable_empty_passwords
   - sshd_disable_user_known_hosts
   - sshd_enable_strictmodes
+
+  - var_system_crypto_policy=default
+  - configure_crypto_policy
+  - configure_ssh_crypto_policy
 
   ### Application whitelisting
   - package_fapolicyd_installed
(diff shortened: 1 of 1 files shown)


rhel8/profiles/e8.profile
```diff
@@ -123,14 +123,16 @@ selections:
...
   - sshd_do_not_permit_user_env
-  - sshd_disable_rhosts_rsa
   - sshd_disable_rhosts
-  - sshd_allow_only_protocol2
   - sshd_set_loglevel_info
   - sshd_disable_empty_passwords
   - sshd_disable_user_known_hosts
   - sshd_enable_strictmodes
 
+  - var_system_crypto_policy=default
```
redhatrises: @shaneboulden is there a crypto policy requirement for Essential 8 like FIPS is for US Gov?

shaneboulden: @redhatrises There's nothing explicit for the Essential 8, though we'd try to follow the [ISM guidelines for using crypto](https://www.cyber.gov.au/ism/guidelines-using-cryptography), and specifically the 'ASD Approved Crypto Algorithms'

It looks like the `default` policy meets a lot of these requirements with the exception that it permits SHA-1 in digital signatures (ASD approves SHA-2 family).

I think we'd probably prefer to use a `future` system-wide policy. The guidelines provided (linked above) are generally conservative, and this meets the requirement for SHA-2 (and explicitly disallowing SHA-1).

matejak:
I have reflected those suggestions in the comment.

redhatrises:
@shaneboulden do we need to add more cyrpto-policy checks for bind, krb, java, etc? Or is ssh just good enough.

shaneboulden:
> @shaneboulden do we need to add more cyrpto-policy checks for bind, krb, java, etc? Or is ssh just good enough.
I think ssh is fine - this is designed to be a baseline policy, and these services are pretty specific to server roles (DNS, domain-joined, Java app server).  

My preferred approach would be for users to create a tailoring file based on the E8 policy that caters for specific server roles.

redhatrises:
@shaneboulden sounds good. Thanks!

Ack and thanks for the work @matejak

PR #5025, opened 2019-11-21
Various minor fixes

matejak:
- Removed incorrect prodtype from a rule (the `unprivileged_bpf_disabled` kernel param doesn't work on RHEL7)
- Added proper quoting to bash array.


redhatrises:
LGTM... Pending Jenkins build passing

PR #5027, opened 2019-11-21
Enable more STIG/SRG table builds in RHEL8

redhatrises:
Now that the STIG and OSPP profiles exist, we can create the nistrefs tables


jan-cerny:
Thanks

PR #5029, opened 2019-11-22
SSGTS: Use wildcards instead of matching substring

jan-cerny:
#### Description:

This commit changes the way test suite users specify rules to be tested
in rule mode. Instead of matching a substring in rule ID, we will accept
wildcards as rule mode arguments. The wildcards are similar to shell
globs, and use Python fnmatch library. For example, argument
accounts_passwords_pam_faillock_deny will run test scenarios for a
single rule and accounts_passwords* will run test scenarios for all
rules which ID starts with accounts_passwords.

The rule ID prefix (xccdf_org.ssgproject.content_rule_) does not need to
be provided by the user because it will be added automatically.

#### Rationale:

Using substring match to find rules to be tested caused problems in
situation when we have 2 rule IDs and one of them is a substring of
another. For example, when an user requested to run test scenarios for
rule accounts_passwords_pam_faillock_deny, the test suite run tests also
for rule accounts_passwords_pam_faillock_deny_root. That was annoying
because testing the latter takes some time.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -246,9 +246,9 @@ Example usage:
 
 ### Rule-based testing
 
-In this mode, you specify the `rule` command and you supply part of the rule
-title as a positional argument. Unlike the profile mode, here each rule from
-the matching rule set is addressed independently, one-by-one.
+In this mode, you specify the `rule` command and you supply one or more rule
+IDs or wildcards as positional arguments. Unlike the profile mode, here each
+rule from the matching rule set is addressed independently, one-by-one.
 
 Rule-based testing enables to perform two kinds of tests:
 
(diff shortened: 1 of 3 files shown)


ggbecker:
Thanks for the patch. I tested and it works great.

ggbecker:
excerpt from one test run:

```
./run_test_rule.sh bash "*hd_disable_*"
INFO - Logging into /home/ggasparb/workspace/github/content/tests/logs/rule-custom-2019-11-25-1146/test_suite.log
INFO - xccdf_org.ssgproject.content_rule_sshd_disable_gssapi_auth
```

PR #5031, opened 2019-11-25
Format OS SRG

shawndwells:
Ran the OS SRG through an XML formatter.


redhatrises:
@shawndwells is this necessary? As the SRG xml file is updated over time, this type of update will get lost.

shawndwells:
> @shawndwells is this necessary? As the SRG xml file is updated over time, this type of update will get lost.

Necessary? Maybe not.  But super useful. Otherwise devs who want to review this file have to manually (re)parse each time this file is accessed.

redhatrises:
> > @shawndwells is this necessary? As the SRG xml file is updated over time, this type of update will get lost.
> 
> Necessary? Maybe not. But super useful. Otherwise devs who want to review this file have to manually (re)parse each time this file is accessed.

Does `oscap xccdf generate guide` not work in this case?

shawndwells:
> > > @shawndwells is this necessary? As the SRG xml file is updated over time, this type of update will get lost.
> > 
> > 
> > Necessary? Maybe not. But super useful. Otherwise devs who want to review this file have to manually (re)parse each time this file is accessed.
> 
> Does `oscap xccdf generate guide` not work in this case? I imagine that most devs would use that or parse in other ways.

No, generate guide creates an HTML report. Sometimes need to review the XML directly.

jan-cerny:
@shawndwells Please remove unrelated commits and remove also merge commits from this PR. Interactive git rebase on top of fresh fetched upstream master usually does the trick.

shawndwells:
Was addressed elsewhere.

PR #5038, opened 2019-11-26
Fix RHV4 DS applicability

yuumasato:
#### Description:

- Limit applicability check of RHV4 DS to RHVH

#### Rationale:

- A Virtualization Host can be either a RHVH or a standard RHEL installation, but each should use its own DS.

### DIFF
--- a/shared/checks/oval/installed_OS_is_rhv4.xml
+++ b/shared/checks/oval/installed_OS_is_rhv4.xml
@@ -12,11 +12,7 @@
       Red Hat Virtualization Host 4 or Red Hat Enterprise Host.</description>
     </metadata>
     <criteria>
-      <extend_definition comment="RHEL7 OS installed" definition_ref="installed_OS_is_rhel7" />
-      <criteria operator="OR">
-        <criterion comment="Red Hat Virtualization Host (RHVH)" test_ref="test_rhvh4_version" />
-        <criterion comment="Red Hat Virtualization Host is based on RHEL (RHELH)" test_ref="test_rhelh4_version" />
-      </criteria>
+      <criterion comment="Red Hat Virtualization Host (RHVH)" test_ref="test_rhvh4_version" />
     </criteria>
   </definition>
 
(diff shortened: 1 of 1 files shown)


shared/checks/oval/installed_OS_is_rhv4.xml
```diff
@@ -11,12 +11,9 @@
       <description>The operating system installed on the system is
       Red Hat Virtualization Host 4 or Red Hat Enterprise Host.</description>
     </metadata>
-    <criteria>
-      <extend_definition comment="RHEL7 OS installed" definition_ref="installed_OS_is_rhel7" />
-      <criteria operator="OR">
-        <criterion comment="Red Hat Virtualization Host (RHVH)" test_ref="test_rhvh4_version" />
-        <criterion comment="Red Hat Virtualization Host is based on RHEL (RHELH)" test_ref="test_rhelh4_version" />
-      </criteria>
```
redhatrises: @yuumasato should we just remove `test_rhelh4_version` and keep the `installed_OS_is_rhel7`?

yuumasato: We can use the `installed_OS_is_rhel7`, but I'd put it in the place of `test_rhelh4_version`.

redhatrises: @yuumasato basically we need to handle this:

- RHVH and RHVM in RHV4 DS
- RHELH (aka RHVH) in RHEL7 DS.

yuumasato: @redhatrises Something like this?

yuumasato: @redhatrises I think we should be fine limiting RHV4 DS to RHVH and RHVM.

But I'd rather not limit RHEL7 DS, and still allow it to evaluate on RHVH due to https://bugzilla.redhat.com/show_bug.cgi?id=1420038.

PR #5046, opened 2019-12-02
SRG Mapping: Do not flip CCI and SRGID when we do not have content

isimluk:
This is follow-up on V1R6 changes.

Kudos to @tedbrunell for noticing this.

See the flip in the collumns 1 and 2 below.

image


isimluk:
Got two other minor bugfixes in.

PR #5052, opened 2019-12-04
Update Ansible when statement to handle only containers

### LINKED ISSUE #5048: DISA RHEL 7 STIG FIPS Rules incorrectly applied to virtual machines
#### Description of problem:
Virtual Machines (ec2 instances in this case) have the variable `ansible_virtualization_role` set to `guest` by default. This causes the Ansible Role to skip all of the FIPS enabling tasks. 

Utilizing the RedHatOfficial role
https://github.com/RedHatOfficial/ansible-role-rhel7-stig

3.10.0-957.21.2.el7.x86_64

#### Steps to Reproduce:

1. Create VM or EC2 instance
2. Run the STIG Role
3. Observe the FIPS tasks are skipped. 
4.

#### Actual Results:
FIPS tasks are skipped and FIPS is not enabled

#### Expected Results:
FIPS tasks run and FIPS is correctly enabled for VMs.

redhatrises:
- Remediations apply to Virtual Machines as well
- Fixes #5048

### DIFF
--- a/ssg/build_remediations.py
+++ b/ssg/build_remediations.py
@@ -332,7 +332,7 @@ class AnsibleRemediation(Remediation):
     def update_when_from_rule(self, to_update):
         additional_when = ""
         if self.associated_rule.platform == "machine":
-            additional_when = ('ansible_virtualization_role != "guest" '
+            additional_when = ('ansible_virtualization_type != "podman" '
                                'or ansible_virtualization_type != "docker"')
         to_update.setdefault("when", "")
         new_when = ssg.yaml.update_yaml_list_or_string(to_update["when"], additional_when)
--- a/ssg/ansible.py
+++ b/ssg/ansible.py
@@ -12,9 +12,9 @@ from .constants import min_ansible_version
 from .constants import REF_PREFIX_MAP
 
 
-def add_minimum_version(ansible_src):
+def add_pre_tasks(ansible_src):
     """
-    Adds minimum ansible version to an Ansible script
+    Adds pre_tasks to an Ansible script
     """
     pre_task = (""" - hosts: all
    pre_tasks:
@@ -23,6 +23,10 @@ def add_minimum_version(ansible_src):
          that: "ansible_version.full is version_compare('%s', '>=')"
          msg: >
            "You must update Ansible to at least version %s to use this role."
+
+     - name: Set Ansible podman fact
+       set_fact:
+         container_env: "{{ lookup('env', 'container') }}"
           """ % (ansible_version_requirement_pre_task_name,
                  min_ansible_version, min_ansible_version))
 
--- a/ssg/build_profile_remediations.py
+++ b/ssg/build_profile_remediations.py
@@ -5,7 +5,7 @@ import os
 import sys
 from collections import namedtuple
 
-from .ansible import add_minimum_version, remove_multiple_blank_lines, \
+from .ansible import add_pre_tasks, remove_multiple_blank_lines, \
                      remove_trailing_whitespace
 from .shims import subprocess_check_output, Queue
 from .build_guides import _is_blacklisted_profile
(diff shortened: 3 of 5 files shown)


ssg/build_remediations.py
```diff
@@ -332,8 +332,7 @@ def _get_rule_reference(self, ref_class):
     def update_when_from_rule(self, to_update):
         additional_when = ""
         if self.associated_rule.platform == "machine":
-            additional_when = ('ansible_virtualization_role != "guest" '
```
yuumasato: @matejak Was the goal of `ansible_virtualization_role != "guest"` to prevent it from running on VMs?
I remember it was related to podman containers, but my memory is not clear.

Maybe at the time it was not clear what would be set in a podman container.
Found this [ansible issue](https://github.com/ansible/ansible/pull/64981/files), and seems that it will be fixed soon.

matejak: I think that Ansible was broken in the same way as it is now - it doesn't even list the system as guest - Ansible inside a podman container thinks that it is a `kvm` host.
When Ansible detects that it runs in docker, it sets role to `guest` *and* type to `docker`.
The `when` clause determines when to run the remediation. Prior to this PR, anything else than `docker` and `guest` **at the same time** resulted in that condition to be true, therefore executing the task.

matejak:
The current code doesn't disable the task on all guests - it does so only on Docker guests. The bugreport suggests otherwise, but it is either a bug in Ansible, or some kind of confusion.
Here is what `when:` does: https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#the-when-statement

redhatrises:
> The current code doesn't disable the task on all guests - it does so only on Docker guests. The bugreport suggests otherwise, but it is either a bug in Ansible, or some kind of confusion.
> Here is what `when:` does: https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#the-when-statement

That's not actually true. `'ansible_virtualization_role != "guest"` handles both container and vms and with the `or` it makes it true. What is in our codebase says essentially: `When not a virtual machine OR when not docker, you can run this task`

matejak:
>> The current code doesn't disable the task on all guests - it does so only on Docker guests...

> That's not actually true.
> ...
> What is in our codebase says essentially: When not a virtual machine OR when not docker, you can run this task

Your last sentence says something very similar to my statement. This double negation is quite mind-boggling, so let's try to make it more understandable: We can use [de Morgan's rule](https://en.wikipedia.org/wiki/De_Morgan%27s_laws) to rephrase the statement to this form:

```
when: not (ansible_virtualization_role == "guest" and ansible_virtualization_role  == "docker")
```
Then, our condition would read "Don't run this task if we are a Docker AND a guest at the same time".

redhatrises:
> > > The current code doesn't disable the task on all guests - it does so only on Docker guests...
> 
> > That's not actually true.
> > ...
> > What is in our codebase says essentially: When not a virtual machine OR when not docker, you can run this task
> 
> Your last sentence says something very similar to my statement. This double negation is quite mind-boggling, so let's try to make it more understandable: We can use [de Morgan's rule](https://en.wikipedia.org/wiki/De_Morgan%27s_laws) to rephrase the statement to this form:
> 
> ```
> when: not (ansible_virtualization_role == "guest" and ansible_virtualization_role  == "docker")
> ```

This is not what our when statement does currently.

> Then, our condition would read "Don't run this task if we are a Docker AND a guest at the same time".

Whatever works.

redhatrises:
Updated the PR for a more accurate fix.

matejak:
You got it wrong, your condition is always true.

matejak:
I like addition of `podman` support, but you have not explained why you think that `ansible_virtualization_type != "docker"` is more likely to be evaluated to true than `ansible_virtualization_role != "guest" or ansible_virtualization_type != "docker"` - isn't this the original purpose of the PR?

Also why to use the `container_env != "podman"` check, when the [PR referenced by Yuuma](https://github.com/ansible/ansible/pull/64981/files) indicates that `ansible_virtualization_type != "podman"` should be used?

Finally, you can run unit-tests quickly by executing `ctest -R unit` in the `build` directory, no need to wait for the CI or for the whole `make check` to finish.

shawndwells:
@redhatrises Closing this PR because it has not been touched since December. Feel free to re-open if/when you find the time! Looks promising.

PR #5054, opened 2019-12-05
Fix XML "parsing" of the remediation functions file.

matejak:
A proper fix is not worth the effort, as we aim to kill shared Bash remediation
with Jinja2 macros.

This fixes the build for Python>=3.7 that saves XMLs slightly differently - attributes have different ordering.

### DIFF
--- a/ssg/build_remediations.py
+++ b/ssg/build_remediations.py
@@ -56,11 +56,11 @@ def get_available_functions(build_dir):
     remediation_functions = []
     with codecs.open(xmlfilepath, "r", encoding="utf-8") as xmlfile:
         filestring = xmlfile.read()
-        # This regex looks implementation dependent but we can rely on
-        # ElementTree sorting XML attrs alphabetically. Hidden is guaranteed
-        # to be the first attr and ID is guaranteed to be second.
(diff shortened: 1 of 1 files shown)


ssg/build_remediations.py
```diff
@@ -60,7 +60,7 @@ def get_available_functions(build_dir):
         # ElementTree sorting XML attrs alphabetically. Hidden is guaranteed
         # to be the first attr and ID is guaranteed to be second.
         remediation_functions = re.findall(
-            r'<Value hidden=\"true\" id=\"function_(\S+)\"',
+            r'<Value.*id=\"function_(\S+)\"',
```
yuumasato: I guess we can trust that XCCDF Values with id beginning with `function_` are remediation functions.

Can you also update the comment in the lines above?

matejak: Yes, it is funny how things that look implementation-dependent eventually turn out to really be implementation-dependent :-)

jan-cerny: :octopus:

jan-cerny:
I have verified that this fixes the build on Fedora Rawhide and that the built SSG datastream contains the Bash functions as XCCDF Values. Thanks.

PR #5055, opened 2019-12-05
modify usbguard_allow_* rules to use new match-all keyword

vojtapolasek:
#### Description:

Usbguard_allow_hid and usbguard_allow_hub now use the new match-all keyword provided by Usbguard rule syntax.
Both rules DO NOT check for actual content of /etc/usbguard/rules.conf file, they just check if the file is not empty or is not missing. If file is empty or missing, they add respective lines. In other cases, intentional modification by administrator is assumed and therefore rule passes. This is reflected by warning in rule.yml.
OCIL, OVAL and Bash remediations are modified.
Also a new rule usbguard_allow_hid_and_hub is created because combining these two separate rules turns out to be difficult. Oval, rule file, Bash remediation and tests are created for this new rule. All three rules use common OVAL check and therefore only one test is provided.

#### Rationale:

IMPORTANT: usbguard_allow_* rules should be understood primarily as convenience 
Rules. They help administrators to ease management of machines.
The match-all keyword allows better matching of USB devices, for example
match-all { 03:*:*  will match HID devices with possibly multiple HID interfaces but will not match devices including any other USB class interface
This was not reasonably easy to do before.


yuumasato:
> Sorry, this won't be as simple as that - these two rules need to be able to work together - when both are enabled, it should result in `allow with-interface match-all { 03:*:* 09:*:*}`

Nice catch. An use case that should be allowed is devices with any combination of `03:*:*` and `09:*:*`.
In this case, having them as separate rules will make remediation and check complicated. As there can be interaction with any other existing rules.

Merging them and treating as a single config seems the sane approach.
The individual rules can still exist, for folks who only want to check for keyboards or hubs.

linux_os/guide/services/usbguard/usbguard_allow_hid/bash/shared.sh
```diff
@@ -1,8 +1,5 @@
 # platform = multi_platform_fedora,Red Hat Enterprise Linux 8,Oracle Linux 8
 
-if ! grep -Eq '^[ \t]*allow[ \t]+with-interface[ \t]+equals[ \t]+\{[ \t]+03:\*:\*[ \t]+\}[ \t]*$' /etc/usbguard/rules.conf ; then
-	echo "allow with-interface equals { 03:*:* }" >> /etc/usbguard/rules.conf
-fi
-if ! grep -Eq '^[ \t]*allow[ \t]+with-interface[ \t]+equals[ \t]+\{[ \t]+03:\*:\*[ \t]+03:\*:\*[ \t]+\}[ \t]*$' /etc/usbguard/rules.conf ; then
-	echo "allow with-interface equals { 03:*:* 03:*:* }" >> /etc/usbguard/rules.conf
+if ! grep -Eq '^[ \t]*allow[ \t]+with-interface[ \t]+all-of[ \t]+\{[ \t]+03:\*:\*[ \t]+\}[ \t]*$' /etc/usbguard/rules.conf ; then
```
ggbecker: Shouldn't be `match-all` instead of `all-of`?
```suggestion
if ! grep -Eq '^[ \t]*allow[ \t]+with-interface[ \t]+match-all[ \t]+\{[ \t]+03:\*:\*[ \t]+\}[ \t]*$' /etc/usbguard/rules.conf ; then
```

vojtapolasek: Thank you, I overlooked this and it is fixed.

dahaic:
Also, I have checked there is SRG mapped to this rule - `SRG-OS-000114-GPOS-00059`. This reads as `The operating system must uniquely identify peripherals before establishing a connection.`

This should be mapped to the `service_usbguard_enabled`, not to the convenience functions making holes in the strict default.

dahaic:
Thank you @vojtapolasek , looks great!

PR #5057, opened 2019-12-05
Remove configure_fapolicyd_mounts rules from profiles.

ggbecker:
#### Description:
Remove configure_fapolicyd_mounts rules from profiles.
- rhel8/ospp.profile
- rhel8/e8.profile
- ocp4/coreos-ncp.profile

#### Rationale:

- fapolicyd doesn't use "fapolicyd.mounts" file anymore so it doens't make sense to check this file.

References: 
- https://github.com/linux-application-whitelisting/fapolicyd/commit/7bf3822820a7e418ec516f5eb204e0249035addc#diff-1f4c31be8741e7d006e123cba8b57556
- Changelog: https://github.com/linux-application-whitelisting/fapolicyd/blob/b69200eff9ea9d878e82f43d226ee8ca10e455b2/ChangeLog#L18


redhatrises:
Should the rules and remediations still exist as well?

ggbecker:
> Should the rules and remediations still exist as well?

I guess we could remove it as well.

jan-cerny:
@ggbecker Nice discovery! Thank you.

PR #5063, opened 2019-12-10
oscap: modify using variables in the printf format

tianzhenjia:
Using variables in the printf format isn't a good choice.

### DIFF
--- a/tests/ds_unselect_rules.sh
+++ b/tests/ds_unselect_rules.sh
@@ -10,11 +10,11 @@ DS=$1
 UNSELECT_LIST=$2
 
 
-printf "Copying %s file to /tmp\n" "$DS"
+printf 'Copying %s file to /tmp\n' "$DS"
 cp $DS /tmp || exit 1
 DS="/tmp/$(basename $DS)"
 
-printf "Unselecting rules listed in the %s from the %s\n" "$UNSELECT_LIST" "$DS"
+printf 'Unselecting rules listed in the %s from the %s\n' "$UNSELECT_LIST" "$DS"
 for rule in $(cat $UNSELECT_LIST); do
 	sed -i "/<.*Rule id=\"$rule/s/selected=\"true\"/selected=\"false\"/g" $DS || exit 1
 	sed -i "/<.*select idref=\"$rule/s/selected=\"true\"/selected=\"false\"/g" $DS || exit 1
--- a/tests/ds_unselect_rules.sh
+++ b/tests/ds_unselect_rules.sh
@@ -10,11 +10,11 @@ DS=$1
 UNSELECT_LIST=$2
 
 
-printf "Copying $DS file to /tmp\n"
+printf "Copying %s file to /tmp\n" "$DS"
 cp $DS /tmp || exit 1
 DS="/tmp/$(basename $DS)"
 
-printf "Unselecting rules listed in the $UNSELECT_LIST from the $DS\n"
+printf "Unselecting rules listed in the %s from the %s\n" "$UNSELECT_LIST" "$DS"
 for rule in $(cat $UNSELECT_LIST); do
 	sed -i "/<.*Rule id=\"$rule/s/selected=\"true\"/selected=\"false\"/g" $DS || exit 1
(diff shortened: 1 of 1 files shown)


tianzhenjia:
yes




Best Regards 
贾天震
中移（苏州）软件技术有限公司
中国移动云能力中心，云计算产品部
手机：18896726192
邮箱 ：jiatianzhen@cmss.chinamobile.com
 
From: OpenSCAP Public Jenkins
Date: 2019-12-10 14:06
To: ComplianceAsCode/content
CC: jiatianzhen; Author
Subject: Re: [ComplianceAsCode/content] oscap: modify using variables in the printf format (#5063)
Can one of the admins verify this patch?

jan-cerny:
@openscap-ci test this please

tianzhenjia:
when to merge this commit?

tianzhenjia:
when to merge this commit?

tianzhenjia:
> @openscap-ci test this please



yes

tests/ds_unselect_rules.sh
```diff
@@ -10,11 +10,11 @@ DS=$1
 UNSELECT_LIST=$2
 
 
-printf "Copying $DS file to /tmp\n"
+printf "Copying %s file to /tmp\n" "$DS"
```
matejak: Please use single quotes for the first argument - that way you won't need to escape the backslash in `\n` - with double quotes, it should actually be `\\n`.

tests/ds_unselect_rules.sh
```diff
@@ -10,11 +10,11 @@ DS=$1
...
 cp $DS /tmp || exit 1
 DS="/tmp/$(basename $DS)"
 
-printf "Unselecting rules listed in the $UNSELECT_LIST from the $DS\n"
+printf "Unselecting rules listed in the %s from the %s\n" "$UNSELECT_LIST" "$DS"
```
matejak: ditto

matejak:
ok to test

tests/ds_unselect_rules.sh
tianzhenjia: printf "Copying %s file to /tmp\n" "$DS" 
and
printf 'Copying %s file to /tmp\n' "$DS" 
The results of these two methods are the same.   
And I don't understand what "the first argument" means.
Look forward to your reply.

matejak: Sure, see double quotes section of the Bash manual: https://www.gnu.org/software/bash/manual/html_node/Double-Quotes.html, especially the backslash-related part.

tianzhenjia: “\n” is interpreted as special character by printf, so there is no need to use  a backslash "\".
 You can see   https://www.computerhope.com/unix/uprintf.htm.
In addition，"printf "Copying %s file to /tmp\n" "$DS" "  what I modified  is based on the usage of printf，although the results are the same.

matejak: Please read the linked piece again. It reads:

> Enclosing characters in double quotes (‘"’) preserves the literal value of all characters within the quotes, with the exception of ‘$’, ‘`’, ‘\’ 
> ...
> The backslash retains its special meaning only when followed by one of the following characters: ‘$’, ‘`’, ‘"’, ‘\’, or newline.

Behavior of backslash followed by literal `n`, i.e. `\n`, is therefore not defined, and `shellcheck` actually complains about that. You are right about what `printf` does, but it has to go through the shell processing first. If you use single quotes, there is no shell processing, so `\n` gets straight to printf.

tianzhenjia: OK，I will use single quotes.
In addition，
`Thebackslash retains its special meaning only when followed by one of the following characters: ‘$’, ‘`’, ‘"’, ‘\’, or newline. Within double quotes, backslashes that are followed by one of these characters are removed. `
The character "newline" means “\n”, it's a whole.  According to the second sentence , I don't think there is any need to add \ in front of  \n.

matejak: `\n` is backslash followed by letter `n`, not a newline. By newline they mean literal newline. As newline is difficult to show in the manual, they describe it by that word.

matejak:
Thank you for your contribution, when we enable shellcheck in CI, there will be less warnings.

PR #5065, opened 2019-12-10
Update references of USBGuard rules

dahaic:
As there is effectively only one security related rule -
`service_usbguard_enabled`, correct OSPP and SRG references should be
there. `package_usbguard_installed` is prerequisite for the service rule.

Other rules, `usbguard_allow_*` are convenience and usability
and as such should not be linked to security requirements.

### DIFF
--- a/linux_os/guide/services/usbguard/usbguard_allow_hid/rule.yml
+++ b/linux_os/guide/services/usbguard/usbguard_allow_hid/rule.yml
@@ -24,10 +24,6 @@ severity: medium
 identifiers:
     cce@rhel8: 82274-2
 
-references:
-    ospp: FMT_SMF_EXT.1
-    srg: SRG-OS-000114-GPOS-00059
-
 ocil_clause: 'USB devices of class 3 are not authorized'
 
 ocil: |-
--- a/linux_os/guide/services/usbguard/package_usbguard_installed/rule.yml
+++ b/linux_os/guide/services/usbguard/package_usbguard_installed/rule.yml
@@ -20,7 +20,8 @@ identifiers:
     cce@rhel8: 82959-8
 
 references:
-    srg: SRG-OS-000378-GPOS-00163
+    ospp: FMT_SMF_EXT.1
+    srg: SRG-OS-000114-GPOS-00059,SRG-OS-000378-GPOS-00163
 
 ocil_clause: 'the package is not installed'
 
--- a/linux_os/guide/services/usbguard/service_usbguard_enabled/rule.yml
+++ b/linux_os/guide/services/usbguard/service_usbguard_enabled/rule.yml
@@ -19,7 +19,7 @@ identifiers:
 
 references:
     ospp: FMT_SMF_EXT.1
-    srg: SRG-OS-000378-GPOS-00163
+    srg: SRG-OS-000114-GPOS-00059,SRG-OS-000378-GPOS-00163
 
 ocil_clause: 'the service is not enabled'
 
--- a/linux_os/guide/services/usbguard/usbguard_allow_hid_and_hub/rule.yml
+++ b/linux_os/guide/services/usbguard/usbguard_allow_hid_and_hub/rule.yml
@@ -25,10 +25,6 @@ severity: medium
 identifiers:
     cce@rhel8: 82368-2
 
-references:
-    ospp: FMT_SMF_EXT.1
-    srg: SRG-OS-000114-GPOS-00059
-
 ocil_clause: 'USB devices of class 3 and 9:00 are not authorized'
 
 ocil: |-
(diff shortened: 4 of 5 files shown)


dahaic:
@redhatrises @tedbrunell hi, can you take a look at this PR? :)

redhatrises:
The only comment I would have is that the `ospp` and `srg` identifiers can stay in the rules themselves as you need all of them to satisfy the requirement.

linux_os/guide/services/usbguard/usbguard_allow_hid/rule.yml
```diff
@@ -24,10 +24,6 @@ severity: medium
 identifiers:
     cce@rhel8: 82274-2
 
-references:
-    ospp: FMT_SMF_EXT.1
-    srg: SRG-OS-000114-GPOS-00059
```
redhatrises: @dahaic why do you feel these references need to be removed? Seems like they should stay so that all of the reference is met.

dahaic: Allowing Class 03 USB devices is not a security feature. It is a convenience feature. And as such it does not seem right to be referenced as something that covers the requirement?

dahaic: (And the `usbguard_allow_*` rules are actually written as a convenience - OVALs there don't enforce anything, they pass when usbguard.rules file is non-empty)

redhatrises: @dahaic does specify a Class 03 USB device "identify" a type of device?

redhatrises: Also don't these rules assist in the configuring of USB interfaces i.e. allowing/disallowing what devices are recognized by the OS?

dahaic: Identification is done by the `usbguard` service - that running is enough to fulfill the requirements. Unfortunately default is too strict (nothing is allowed) so we need to sneak in something to soften it for general purpose system to be usable right after the installation :)

You are right that these rules are, in a way, related to the requirements . But they are optional extension. Should we still mark them with the references? These rules does not make system any more aligned with the requirement, and omission of the rule will make the system safer.

Keeping the references seems counter-intuitive to me.

jan-cerny: @redhatrises What do you think about @dahaic's explanation?

redhatrises: The references still apply because organizationally I may allow class 3 devices whereas another organization may just want a strict policy that doesn't allow them. So the references need to be kept.

yuumasato: @redhatrises So the use case for references is that they are used to identify rules related to a requirement that one may be interested to include in its policy?

yuumasato:
My initial understanding was that references served the purpose of justifying existence of the rule, that the rule existed to support a requirement from a policy.
I understand that not all rules referencing a requirement may be necessary, as one may want to adapt for the situation and "make" its own policy.

But it seems that references can also be used to find other interesting rules, something in the lines of:
"If you selected this rule, you may want to look at these other rules which reference the same requirement."

If what I said in the previous paragraph is a common/valid use case, we can leave the references in the convenience usbguard rules.

jan-cerny:
@dahaic If you agree feel free to close it.

jan-cerny:
@dahaic What are your plans?

matejak:
I am in favor of merging this because of this thought process: The rule is a convenient feature which doesn't got against the OSPP regulation, which is the reason why we put it into the profile. On the other hand, it doesn't make the system safer, i.e. it doesn't implement OSPP.
And we can communicate this exactly by removing the reference - if I see that a reference-less rule is part of a profile, I can see immediately that the rule must be a convenience feature, not part of implementation of the regulation.

yuumasato:
>  if I see that a reference-less rule is part of a profile, I can see immediately that the rule must be a convenience feature, not part of implementation of the regulation.

I guess this is the point of the discussion, some requirements are up to the organization to define.

The rules that allow HIDs and Hubs don't make the system any safer, but they allow the profile to define list of allowed external devices.
Now, my understanding is that when RHEL8 OSPP Profile selects these rules, the profile is composing its policy of authorized devices.
Another organization defining its OSPP may choose to authorize a different set of devices/interfaces.

And from the point of view of discoverability, I think we should keep the references.
Someone implementing a policy may want to search for rules that can support a requirement and help extend the list of allowed devices/interface, I see the references as a good way to find these related rules.

jan-cerny:
@yuumasato is more convincing. Moreover, if we remove them sooner or later someone adds them.

matejak:
OK, I see the point, and I am closing this PR.

PR #5066, opened 2019-12-11
WIP: Split audit rules

vojtapolasek:
#### Description:

Rule ensuring right files in /etc/audit/rules.d is splitted into individual rules.

#### Rationale:

This improves management of audit policy rules, allows more granularity.


ggbecker:
I notice that the jenkins build is failing because of rule file `linux_os/guide/system/auditing/policy_rules/audit_rules_10-base-config_rule_file/rule.yml` should be actually `linux_os/guide/system/auditing/policy_rules/audit_rules_10-base-config/rule.yml` as OSPP profile selects `audit_rules_10-base-config` rule.

redhatrises:
Thanks for starting this work @vojtapolasek. But there should be no more adding of rules. The rules already exist. ` linux_os/guide/system/auditing/policy_rules/` will be going away. 1 rule for 1 audit configuration is the requirement.

vojtapolasek:
@redhatrises  I am not sure I understand you. I thought we are going to split the big rule linux_os/guide/system/auditing/policy_rules/audit_rules_for_ospp into several separate rules, where 1 rule = 1 file. This was suggested so that administrators can have greater control over enabled auditing rules. @dahaic  ?

vojtapolasek:

redhatrises:
> @redhatrises I am not sure I understand you. I thought we are going to split the big rule linux_os/guide/system/auditing/policy_rules/audit_rules_for_ospp into several separate rules, where 1 rule = 1 file. This was suggested so that administrators can have greater control over enabled auditing rules. @dahaic ?

This was commented in the Epic. The audit rules should already exist. 1 rule for 1 audit rule.

dahaic:
As communicated in #4680 significance of the ordering in the audit rule files makes one SCAP rule per one audited event fool's errand - we would have to build pretty significant tooling to generate ovals that care about preserving order. The split as it is followed here and was previewed in #4680 is technical compromise that to our knowledge satisfies both sides sufficiently.

matejak:
@redhatrises said:
> This was commented in the Epic. The audit rules should already exist. 1 rule for 1 audit rule.

I agree with both of you - while Gabe is correct that the ideal state is to have one rule per one Audit setting, @dahaic is right on the money that there is no way how a check or remediation could be implemented - we would indeed need improvements to the existing Audit tooling, and also a special OVAL Audit test.

Unless there is a deus ex machina solution that we are not aware of, then we really have no choice than to continue the way of this PR, which moves the state somehow closer to that ideal.

linux_os/guide/system/auditing/policy_rules/audit_rules_10-base-config_rule_file/rule.yml
```diff
@@ -0,0 +1,45 @@
...
+
+
+
+rationale: |-
+    The audit rules defined in <tt>/usr/share/audit/sample-rules</tt> are the recommended way to meet compliance with OSPP v4.2.1.
+
+severity: medium
+
+identifiers:
+    cce@rhel8: 82368-2
```
matejak: This CCE has been reused, that's why tests fail.

vojtapolasek:
test this please

matejak:
Good job - the template looks fine, it is documented, and a test that demonstrates the functionality is provided. I can imagine that rule IDs or rule descriptions could be changed, but this can be done in another PR later, as this kind of changes is easy to review.
Merging, thank you for your PR!

PR #5071, opened 2019-12-12
Fix mod fixes script

ggbecker:
#### Description:

- Parse files without expanding jinja macros as it doesn't contain the correct environment to expand macros.
- Do not add platform if remediation already contains `multi_platform_all` to avoid redundancy.

#### Rationale:

- Some files were being skipped and they shouldn't.


matejak:
LGTM, thank you for this fix!

PR #5072, opened 2019-12-12
Use the FIPS:OSPP Crypto Policy

matejak:
- Defined the FIPS:OSPP Value
- As Crypto Policy modules cause the crypto backends configuration not to use symlinks, the OVAL check had to be streamlined.
- Symlink-focused test scenarios had to be dropped from the test suite.
- The harden SSH/SSHD rules were removed in favor of the new policy.


matejak:
The OCIL is correct, there is still a symlink from `/etc/krb5.conf.d/crypto-policies` to `/etc/crypto-policies/back-ends/krb5.config`. The new think is that in the past, the target `/etc/crypto-policies/back-ends/krb5.config` always was a symlink, whereas now, it may be a file, so the test had to be extended.

ggbecker:
> The OCIL is correct, there is still a symlink from `/etc/krb5.conf.d/crypto-policies` to `/etc/crypto-policies/back-ends/krb5.config`. The new think is that in the past, the target `/etc/crypto-policies/back-ends/krb5.config` always was a symlink, whereas now, it may be a file, so the test had to be extended.

Ok, now I get it. Thank you for the explanation.

ggbecker:
LGTM, test scenarios pass accordingly for both `configure_crypto_policy` and `configure_kerberos_crypto_policy` using both `bash` and `ansible` remediation.

PR #5076, opened 2019-12-16
[wip] Update NIST 800-53 mappings

shawndwells:
- Removed AC-17(7) and AC-17(8) because they were removed from NIST 800-53rev4
- Using CM-6(a) vs CM-2 because CM-6(a) is about IT products, whereas CM-2 is about config management of whole information system
- [ ] CM-6(b) should be swapped to CM-6(a)


shawndwells:
Using another PR

Issue #5012, opened 2019-11-14
Split audit rules into unique checks


shawndwells:
linux_os/guide/system/auditing/policy_rules/audit_rules_for_ospp/rule.yml should be split into unique checks

marcusburghardt:
Long time without interaction. This rule is used in one single profile.
Please, reopen if still relevant.

PR #5077, opened 2019-12-16
Revert "WIP: Split audit rules"

shawndwells:
Reverts ComplianceAsCode/content#5066

Having multiple audit events evaluated in a single rule is not acceptable for US Government baselines. Reverting this PR.


yuumasato:
@shawndwells #5066 is related to #5012, and we discussed this in https://github.com/ComplianceAsCode/content/pull/4680#issuecomment-521635195

Managing ordered audit rule individually is costly and error prone. Grouping the audit events by their intention is much more manageable and should still allow for meaningful tailoring.

PR #5085, opened 2020-01-02
Added machine-only CPEs to rules relevant only to non-virtualized systems

matejak:
This update ensures that rules that container/chrooted scanning is free from noise caused by notapplicable rules.

### DIFF
--- a/linux_os/guide/services/imap/configure_dovecot/dovecot_enabling_ssl/dovecot_disable_plaintext_auth/rule.yml
+++ b/linux_os/guide/services/imap/configure_dovecot/dovecot_enabling_ssl/dovecot_disable_plaintext_auth/rule.yml
@@ -16,6 +16,8 @@ rationale: |-
 
 severity: unknown
 
+platform: machine  # The check uses service_... extended definition, which doesnt support offline mode
+
 identifiers:
     cce@rhel6: 27144-5
     cce@rhel7: 80299-1
--- a/linux_os/guide/services/imap/configure_dovecot/dovecot_enabling_ssl/dovecot_enable_ssl/rule.yml
+++ b/linux_os/guide/services/imap/configure_dovecot/dovecot_enabling_ssl/dovecot_enable_ssl/rule.yml
@@ -19,6 +19,8 @@ rationale: |-
 
 severity: unknown
 
+platform: machine  # The check uses service_... extended definition, which doesnt support offline mode
+
 identifiers:
     cce@rhel6: 27571-9
     cce@rhel7: 80296-7
--- a/linux_os/guide/services/mail/postfix_client/postfix_network_listening_disabled/rule.yml
+++ b/linux_os/guide/services/mail/postfix_client/postfix_network_listening_disabled/rule.yml
@@ -17,6 +17,8 @@ rationale: |-
 
 severity: medium
 
+platform: machine  # The check uses service_... extended definition, which doesnt support offline mode
+
 identifiers:
     cce@rhel6: 26780-7
     cce@rhel7: 80289-2
--- a/linux_os/guide/system/accounts/accounts-restrictions/password_storage/accounts_password_all_shadowed/rule.yml
+++ b/linux_os/guide/system/accounts/accounts-restrictions/password_storage/accounts_password_all_shadowed/rule.yml
@@ -15,6 +15,8 @@ rationale: |-
 
 severity: medium
 
+platform: machine  # The oscap password probe doesn't support offline mode
+
 identifiers:
     cce@rhel6: 26476-2
     cce@rhel7: 27352-4
--- a/linux_os/guide/system/permissions/files/no_files_unowned_by_user/rule.yml
+++ b/linux_os/guide/system/permissions/files/no_files_unowned_by_user/rule.yml
@@ -21,6 +21,8 @@ rationale: |-
 
 severity: medium
 
+platform: machine  # The oscap password probe doesn't support offline mode
+
 identifiers:
     cce@rhel6: 27032-2
     cce@rhel7: 80134-0


redhatrises:
Agreed with most of these except for the following:

```
linux_os/guide/services/imap/configure_dovecot/dovecot_enabling_ssl/dovecot_disable_plaintext_auth/rule.yml
linux_os/guide/services/imap/configure_dovecot/dovecot_enabling_ssl/dovecot_enable_ssl/rule.yml
linux_os/guide/services/mail/postfix_client/postfix_network_listening_disabled/rule.yml
linux_os/guide/system/accounts/accounts-restrictions/password_storage/accounts_password_all_shadowed/rule.yml
linux_os/guide/system/accounts/accounts-session/root_paths/accounts_root_path_dirs_no_write/rule.yml
linux_os/guide/system/accounts/accounts-session/root_paths/root_path_no_dot/rule.yml
linux_os/guide/system/permissions/files/no_files_unowned_by_user/rule.yml
```

BTW... we should avoid marking rules as `not applicable` because of lack of support in the offline probes. This is considered a bad security practice.

ggbecker:
> Agreed with most of these except for the following:
> 
> ```
> linux_os/guide/services/imap/configure_dovecot/dovecot_enabling_ssl/dovecot_disable_plaintext_auth/rule.yml
> linux_os/guide/services/imap/configure_dovecot/dovecot_enabling_ssl/dovecot_enable_ssl/rule.yml
> linux_os/guide/services/mail/postfix_client/postfix_network_listening_disabled/rule.yml
> linux_os/guide/system/accounts/accounts-restrictions/password_storage/accounts_password_all_shadowed/rule.yml
> linux_os/guide/system/accounts/accounts-session/root_paths/accounts_root_path_dirs_no_write/rule.yml
> linux_os/guide/system/accounts/accounts-session/root_paths/root_path_no_dot/rule.yml
> linux_os/guide/system/permissions/files/no_files_unowned_by_user/rule.yml
> ```
> 
> BTW... we should avoid marking rules as `not applicable` because of lack of support in the offline probes. This is considered a bad security practice.

@matejak what happens if someone tries to scan those rules using online mode within a container? Do we expect to be `notapplicable`? I think we should not mark as such just because of the unsupported offline mode as Gabe mentioned.

I also identified potentially candidates to mark as machine:
`linux_os/guide/services/nfs_and_rpc/disabling_nfs/disabling_netfs/service_netfs_disabled/rule.yml`
`linux_os/guide/services/nfs_and_rpc/disabling_nfs/disabling_nfs_services/group.yml` (check the rules in this group, I think you can remove `machine` from the rules and keep only in the group.yml file)
`kernel_disable_entropy_contribution_for_solid_state_drives`

What do you think?

matejak:
@redhatrises 
I agree with you on the big picture - in an ideal world, those rules would produce an accurate result. However, OpenSCAP probes that don't support offline mode produce misleading results - as a rule is composed of multiple tests and the NOTAPPLICABLE OVAL result has low priority in criteria evaluation, results of rules that involve those probes are likely to be inaccurate.
Rules that you reference largely use the systemd probe, and envvar probe, which don't support offline scanning in RHEL7. The password probe [is even more complicated](https://github.com/OpenSCAP/openscap/issues/344).
The problem is that including those rules and checks in the content applicable to the offline use case would result in much worse security practice than marking them as not applicable.

@ggbecker Good suggestions, I have extended those rules.

yuumasato:
@evgenyz @matejak What is the result of the OVAL definition when a probe is not supported in offline mode?
Does the probe return impracticable results? Does it consistently return `unknown`?

matejak:
I don't know, and the question is whether it is important - if we know that test that is part of a check reports either some error or potentially incorrect results, we probably don't want to show the check result to the customer, because it is corrupted by that test.
False positives/negatives are evil, and the only remaining question is whether it is better to allow returning error-class result, or to ensure that we return notapplicable thanks to the machine-only CPE.

yuumasato:
> I don't know, and the question is whether it is important - if we know that test that is part of a check reports either some error or potentially incorrect results, we probably don't want to show the check result to the customer, because it is corrupted by that test.

I understand, my rationale was on the line that returning `unknown` makes more sense to me, than `notapplicable`.
But it is up to the scanner to return `unknown` if it finds itself doing evaluation in offline mode and the content requests use of an unsupported probe/test.
On the content side we can only do as much as preventing the scanner from evaluating content that we know will yield unreliable results in certain environments.

> False positives/negatives are evil, and the only remaining question is whether it is better to allow returning error-class result, or to ensure that we return notapplicable thanks to the machine-only CPE.

I agree that returning `notapplicable` is better than leaving the scanner to return unreliable results.

> BTW... we should avoid marking rules as `not applicable` because of lack of support in the offline probes. This is considered a bad security practice.

@redhatrises I agree that it is not ideal practice.
I suggest filling a ticket so we revisit these rules when offline probe support is enhanced.

redhatrises:
> > I don't know, and the question is whether it is important - if we know that test that is part of a check reports either some error or potentially incorrect results, we probably don't want to show the check result to the customer, because it is corrupted by that test.
> 
> I understand, my rationale was on the line that returning `unknown` makes more sense to me, than `notapplicable`.
> But it is up to the scanner to return `unknown` if it finds itself doing evaluation in offline mode and the content requests use of an unsupported probe/test.
> On the content side we can only do as much as preventing the scanner from evaluating content that we know will yield unreliable results in certain environments.
> 
> > False positives/negatives are evil, and the only remaining question is whether it is better to allow returning error-class result, or to ensure that we return notapplicable thanks to the machine-only CPE.
> 
> I agree that returning `notapplicable` is better than leaving the scanner to return unreliable results.
> 
> > BTW... we should avoid marking rules as `not applicable` because of lack of support in the offline probes. This is considered a bad security practice.
> 
> @redhatrises I agree that it is not ideal practice.
> I suggest filling a ticket so we revisit these rules when offline probe support is enhanced.

This is a problem against the scanner and not the content. Therefore, the approach is wrong. The content shouldn't be fixed to meet the scanner's failure for proper probe support. The scanner should be fixed to address these. Security content is more authoritative than probe support. This means that saying `not applicable` means that a rule is truly not applicable. Part of security is evaluating your results and understanding the accuracies and inaccuracies that the scanner provides you. The content needs to be and stay authoritative and patching the content to return green or not applicable results is a error and inaccuracy in the content. The proper fix would be to either enable offline support in the scanner and seperately have the scanner return `probe not supported in offline mode` debug message and stop the rule evaluation there with an `unknown` result.

matejak:
@redhatrises: Let me object to some of rules that you have named:

`dovecot_disable_plaintext_auth`, `dovecot_enable_ssl`, `postfix_network_listening_disabled`: These rules are relevant only if the respective service is enabled. If dovecot or postfix are running in a container, it is like that because those processes were set up as container entrypoint/command, not because they are enabled systemd services. Therefore, even if our scanner could check for offline systemd unit setup, it would produce false positives/negatives, which is worse than notapplicable.
`accounts_password_all_shadowed` and `no_files_unowned_by_user`: Aside from logging in to containers is an antipattern, I assume that you have missed my link to [#344](https://github.com/OpenSCAP/openscap/issues/344), which explains why offline check is a principal, not implementation problem. If one is only after passwd file, the check should have been implemented using a `textfilecontent` test. So this is a content-side issue again.

You are right that rules with the envvar test are indeed rules that could work in offline mode use cases, so I have kicked those two rules out of the PR.

matejak:
The non-controversial part of this PR has been exposed as #5104

jan-cerny:
@matejak What are your plans?

matejak:
I don't plan to modify the PR, as the main objection is out of scope for it. So a reviewer either merges it, or closes it, or the PR is split into OK and not OK parts even further.
@redhatrises objections are valid, but I think that so are my - there is no obvious easy solution.

matejak:
Closing, as there are ambiguous rules and no consensus how to handle them.

PR #5089, opened 2020-01-03
Set sshd private key permission to 0600 for Ubuntu 18.04

### LINKED ISSUE #5088: Permissions on SSH Server Private *_key Key Files not sufficient for Ubuntu 18.04
#### Description of problem:
The OpenSSH server requires stricter permissions than the current `0640` and will ignore private keys with those permissions.

0.1.47

Ubuntu 18.04

#### Steps to Reproduce:

1. Ensure permissions `0640` are set on a private key in `/etc/ssh/` on an Ubuntu 18.04 machine
2. Start the `sshd` daemon

#### Actual Results:
The keys are ignored as they are not set to a permissive enough version
```
@         WARNING: UNPROTECTED PRIVATE KEY FILE!          @
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
Permissions 0640 for '/etc/ssh/ssh_host_ed25519_key' are too open.
It is required that your private key files are NOT accessible by others.
This private key will be ignored.
key_load_private: bad permissions
Could not load host key: /etc/ssh/ssh_host_ed25519_key
```

#### Expected Results:
the openSSH daemon will load the private keys

This appears to have been addressed for OpenSUSE, as can be seen here 
https://github.com/ComplianceAsCode/content/blob/1c026ffd0908a0e5fdfc2a99b02cb8b9f353ce60/linux_os/guide/services/ssh/file_permissions_sshd_private_key/rule.yml#L6-L10

Is there a reason why we should use `0640` and not universally enforce `0600`? If the former, then Ubuntu 18.04 needs must be specifically updated to `0600`

Scronkfinkle:
#### Description:
This PR addresses the fact that the OpenSSH server will ignore all private key files if they have permissions less strict than `0600` on Ubuntu 18.04. This appears to also have been the case for openSUSE, and I have updated the rule file to include Ubuntu 18.04 to this scenario.

- Fixes #5088

### DIFF
--- a/linux_os/guide/services/ssh/file_permissions_sshd_private_key/rule.yml
+++ b/linux_os/guide/services/ssh/file_permissions_sshd_private_key/rule.yml
@@ -3,7 +3,7 @@ documentation_complete: true
 title: 'Verify Permissions on SSH Server Private *_key Key Files'
 
 description: |-
-    {{% if product in ['opensuse', 'sle11', 'sle12'] %}}
+    {{% if product in ['ubuntu1804','opensuse', 'sle11', 'sle12'] %}}
     {{{ describe_file_permissions(file="/etc/ssh/*_key", perms="0600") }}}
     {{% else %}}
     {{{ describe_file_permissions(file="/etc/ssh/*_key", perms="0640") }}}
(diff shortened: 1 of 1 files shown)


redhatrises:
@openscap-ci test this please

jan-cerny:
Thanks

PR #5093, opened 2020-01-06
Fix CPE to properly check /etc/login.defs on Ubuntu & Debian systems

### LINKED ISSUE #5092: `/etc/login.defs` is provided by a different package in Ubuntu 18.04
#### Description of problem:
Ubuntu 18.04 is provided the `/etc/login.defs` file by the `login` package. As a result, the CPE entry does not resolve the file correctly, and the check is marked as `notapplicable`

0.1.47

Ubuntu 18.04

#### Steps to Reproduce:

1. Install Ubuntu 18.04 on a machine
2. Run a SCAP scan with the `Set Password Maximum Age` option selected on it

#### Actual Results:
Scan returns `notapplicable`

#### Expected Results:
Scan returns `pass` or `fail`, depending on condition.

By running `dpkg -S /etc/login.defs` on the machine it can be seen that the login package provides the file:
```
$ dpkg -S /etc/login.defs 
login: /etc/login.defs
```

Scronkfinkle:
Checking for the proper lines inside of `/etc/login.defs` requires the file to be provided `shadow-utils` due to the CPE definition for it. On Ubuntu 18.04, the file is installed by the `login` package. This results in the check not occurring and the scan being marked as `notapplicable`.  

This PR resolves this by having Ubuntu 18.04 systems check for the `login` package instead.

- Fixes #5092


redhatrises:
@Scronkfinkle doesn't debian and ubuntu use login over shadow-utils? Seems there is duplicated code here.

jan-cerny:
@Scronkfinkle 

Your patch fixes the issue, but I would prefer a different solution, because you modify the definition for "package shadow-utils is installed". Look at the comments, titles, IDs, etc. in that file -- they all mention package shadow-utils, and suddenly it checks for a different package. That will be confusing.

I think that a correct solution is to create a new CPE entry cpe:/a:login, create OVAL for it, add it to the CPE dictioniary. That will be all similar to the CPE check for shadow-utils. And then, use something like this in rules YAMLs, eg. in `linux_os/guide/system/accounts/accounts-restrictions/password_expiration/accounts_maximum_age_login_defs/rule.yml`

```
{{% if product == "ubuntu1804" %}}
platform: login
{{% else %}}
platform: shadow-utils
{{% endif %}}
```

Scronkfinkle:
@redhatrises Looking into the packages for various Ubuntu distributions and Debian, it appears that login is used by all of them (Ubuntu 16.04-19.10 & Debian jessie-buster), and it provides the `/etc/login.defs`. 

I'm going to follow up on @jan-cerny 's advice and re-write it as a new CPE entry. Thanks for the advice!

Scronkfinkle:
@jan-cerny I think the changes i've made match the description you gave. I didn't add the login package to the RHEL stuff though because it's Debian/Ubuntu specific.

redhatrises:
Yeah.... sorry but no to this change... There should be no jinja templating around yaml keys.

Scronkfinkle:
@redhatrises 
To fix that, The jinja templating can assign the correct platform to a variable and then parse it. This would avoid wrapping the key in the template

Something like:
```
{{% set login_defs_platform = "shadow-utils" %}} 
{{% if product in ["ubuntu1404", "ubuntu1604", "ubuntu1804", "debian8", "debian9", "debian10"] %}}
  {{% set login_defs_platform = "login" %}}
{{% endif %}}
platform: {{ selected_login_defs_platform }}
```

Scronkfinkle:
@redhatrises The PR has been updated to better match the description you provided.

PR #5094, opened 2020-01-07
Use replace module in gpgcheck ansible

### LINKED ISSUE #3810: Ansible lineinfile doesn't rememdiate gpgcheck settings with multiple repos in a single file
#### Description of problem:
Ansible lineinfile doesn't rememdiate gpgcheck settings with multiple repos in a single file.
It just does the last found match. Users would expect to be secure and compliance with policy after running an Ansible remediation, but that is not the case.

Master

All

#### Steps to Reproduce:

1.
2.
3.
4.

#### Actual Results:
Only the last gpgcheck is changed in a multi-repo yum/dnf repo file.

#### Expected Results:
All repos have gpgcheck added or changed to gpgcheck=1

caitelatte:
Closes #3810

This will comment out all lines beginning with 'gpgcheck'
instead of only changing the final instance as lineinfile
does.

#### Description:

- This change modifies the play to use the [Ansible replace module](https://docs.ansible.com/ansible/latest/modules/replace_module.html#replace-module) instead which will address each line in each file.
- The original ensure_gpgcheck_never_disabled ansible remediation play uses the [Ansible lineinfile module](https://docs.ansible.com/ansible/latest/modules/lineinfile_module.html) which only affects one line at a time.
- One change in behaviour is that this play now comments out the entire line instead of replacing it with `gpgcheck=1`. Happy to modify this to replace the line with `gpgcheck=1` instead.

#### Rationale:

- This change will address a faulty playbook which did not fix all gpgcheck configurations.
- Fixes #3810

### DIFF
--- a/linux_os/guide/system/software/updating/ensure_gpgcheck_never_disabled/ansible/shared.yml
+++ b/linux_os/guide/system/software/updating/ensure_gpgcheck_never_disabled/ansible/shared.yml
@@ -16,4 +16,4 @@
   replace:
     dest: "{{ item.path }}"
     regexp: '^gpgcheck.+'
-    replace: 'gpgcheck = 1'
+    replace: 'gpgcheck=1'
--- a/linux_os/guide/system/software/updating/ensure_gpgcheck_never_disabled/ansible/shared.yml
+++ b/linux_os/guide/system/software/updating/ensure_gpgcheck_never_disabled/ansible/shared.yml
@@ -15,5 +15,5 @@
   with_items: "{{ yum_find.files }}"
   replace:
     dest: "{{ item.path }}"
-    regexp: '^(gpgcheck.+)'
-    replace: '# \1'
+    regexp: '^gpgcheck.+'
+    replace: 'gpgcheck = 1'
--- a/linux_os/guide/system/software/updating/ensure_gpgcheck_never_disabled/ansible/shared.yml
+++ b/linux_os/guide/system/software/updating/ensure_gpgcheck_never_disabled/ansible/shared.yml
@@ -13,8 +13,7 @@
 
 - name: Ensure gpgcheck Enabled For All {{{ pkg_manager }}} Package Repositories
   with_items: "{{ yum_find.files }}"
-  lineinfile:
-    create: yes
+  replace:
     dest: "{{ item.path }}"
-    regexp: '^gpgcheck'
-    line: 'gpgcheck=1'
+    regexp: '^(gpgcheck.+)'
+    replace: '# \1'
--- a/linux_os/guide/system/software/updating/ensure_gpgcheck_never_disabled/ansible/shared.yml
+++ b/linux_os/guide/system/software/updating/ensure_gpgcheck_never_disabled/ansible/shared.yml
@@ -1,19 +1,20 @@
 # platform = multi_platform_rhel,multi_platform_ol,multi_platform_fedora,multi_platform_rhv
 # reboot = false
-# strategy = unknown
+# strategy = enable
 # complexity = low
 # disruption = medium
-#
-- name: Find All {{{ pkg_manager }}} Repositories
-  find:
-    paths: "/etc/yum.repos.d/"
-    patterns: "*.repo"
-    contains: ^\[.+]$
-  register: yum_find
+- name: Grep for {{{ pkg_manager }}} repo section names
+  shell: grep -HEr '^\[.+\]' -r ./etc/yum.repos.d/
+  register: repo_grep_results
+  ignore_errors: True
+  changed_when: False
 
-- name: Ensure gpgcheck Enabled For All {{{ pkg_manager }}} Package Repositories
-  with_items: "{{ yum_find.files }}"
-  replace:
-    dest: "{{ item.path }}"
-    regexp: '^gpgcheck.+'
-    replace: 'gpgcheck=1'
+- name: Set gpgcheck=1 for each {{{ pkg_manager }}} repo
+  ini_file:
+    path: "{{item[0]}}"
+    section: "{{item[1]}}"
+    option: gpgcheck
+    value: '1'
+    no_extra_spaces: True
+  # regex filters grep output for files ending in .repo and matching section names.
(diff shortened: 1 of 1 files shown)


jan-cerny:
@openscap-ci ok to test

jan-cerny:
Works great for me, but I prefer to modify it to replace the line with gpgcheck = 1 instead.

redhatrises:
Also... what if the line doesn't exist for each repo entry? Does replace handle this?

caitelatte:
> Works great for me, but I prefer to modify it to replace the line with gpgcheck = 1 instead.

Sounds good @jan-cerny, I have pushed a new commit with gpgcheck=1 instead.

> Also... what if the line doesn't exist for each repo entry? Does replace handle this?

Hm, replace will not add a line to a file if it's missing - it just treats the file as not needing a change.

In this case the repo settings should use the global setting, which is checked in the `ensure_gpgcheck_globally_activated` check. The current behaviour of `lineinfile` would add `gpgcheck=1` to the end of the file if there is no previous occurence, however it would still be inconsistent in the case of multiple repos defined in a single file.

I think it would be better to ensure there are no `0` settings at all than to make there there are `1` settings for each repo.

redhatrises:
> > Works great for me, but I prefer to modify it to replace the line with gpgcheck = 1 instead.
> 
> Sounds good @jan-cerny, I have pushed a new commit with gpgcheck=1 instead.
> 
> > Also... what if the line doesn't exist for each repo entry? Does replace handle this?
> 
> Hm, replace will not add a line to a file if it's missing - it just treats the file as not needing a change.
> 
> In this case the repo settings should use the global setting, which is checked in the `ensure_gpgcheck_globally_activated` check. The current behaviour of `lineinfile` would add `gpgcheck=1` to the end of the file if there is no previous occurence, however it would still be inconsistent in the case of multiple repos defined in a single file.
> 
> I think it would be better to ensure there are no `0` settings at all than to make there there are `1` settings for each repo.

Be careful with assuming that `ensure_gpgcheck_globally_activated` is actually enabled.... this may not be the case... and could be tailored out of the security profile which would mean it is never enabled.

> I think it would be better to ensure there are no `0` settings at all than to make there there are `1` settings for each repo.

The default is that `gpgcheck` is not enabled. So, if `gpgcheck` does not exist or is not set to `1`, `gpgcheck` is disabled. I like your solution here for `replace` because it ensure that when gpgcheck is set that it is set to 1. Have an idea for handling the case where `gpgcheck` doesn't exist for all or any repos? I imagine it to be a separate task. If not, it would probably be good to open a ticket to fix the case where `gpgcheck` doesn't exist.

caitelatte:
> The default is that gpgcheck is not enabled. So, if gpgcheck does not exist or is not set to 1, gpgcheck is disabled. I like your solution here for replace because it ensure that when gpgcheck is set that it is set to 1. Have an idea for handling the case where gpgcheck doesn't exist for all or any repos? I imagine it to be a separate task. If not, it would probably be good to open a ticket to fix the case where gpgcheck doesn't exist.

Hm I see what you mean @jan-cerny. I didn't know that the default was no gpgcheck :o I have a solution to ensure it exists but it's got some complex multiline regex involved! Here's a regex101 link with an example test file: https://regex101.com/r/Wk0fM3/8

The play would read like the following:

```yaml
- name: Ensure gpgcheck Enabled For All Yum Package Repositories
  with_items: "{{ yum_find.files }}"
  replace:
    dest: "{{ item.path }}"
    regexp: '(?P<section_start>^\[.+\](?:\n^(?!(\[|gpgcheck)).*)*)(?P<gpgcheck_line>\n^gpgcheck.+)?'
    replace: '\g<section_start>\ngpgcheck=1'
```

It looks for the beginning line of a section, then matches new lines that are not a new section header or begin with gpgcheck until it reaches an optional gpgcheck line. Then it replaces the matched part of the section with the same content, minus the gpgcheck line, plus `gpgcheck=1`.

The only issue I could find is that it can't match a section which has `gpgcheck` declared twice, but I think that's a fair enough thing to miss out on.

If we change the behaviour of this check to explicitly set `gpgcheck=1` in every repo file, I'd like to reword the `rule.yml` to say it explicitly sets the behaviour rather than removing lines.

An alternative solution without a weird regex, but changing the method to the `ini_file` module would be as follows, but I think the suggested regex is pretty solid:

1. grep for lines starting with `[` in all `yum.repos.d` files
2. loop through each section in each file for an [`ini_files`](https://docs.ansible.com/ansible/latest/modules/ini_file_module.html) module call to explicitly set `gpgcheck = 1`?

jan-cerny:
@caitelatte The approach with "ini_file" module looks better to me, we already use the ini_file module in other yum related tasks, and you will not have to create any crazy regex.

caitelatte:
Ok, I've figured out the grep+ini_file approach and pushed it up in 2051bc3. It runs grep, pulling out the filenames and section headers within, then loops over each filename+section header to set gpgcheck=1.

This is a change in behaviour from how the `ensure_gpgcheck_never_disabled` rule is named and described, as well as how the bash solution works. A more accurate name for this would be `ensure_gpgcheck_always_enabled`. I'd be happy to rewrite the description blocks in the ensure_gpgcheck_never_disabled rules.yml and shared.xml but don't understand how that would affect the rule's coverage of identifiers and STIG rules, or where else the name would need to be changed so may need some help with that?

linux_os/guide/system/software/updating/ensure_gpgcheck_never_disabled/ansible/shared.yml
```diff
@@ -1,20 +1,20 @@
...
 # disruption = medium
-#
-- name: Find All {{{ pkg_manager }}} Repositories
-  find:
-    paths: "/etc/yum.repos.d/"
-    patterns: "*.repo"
-    contains: ^\[.+]$
-  register: yum_find
+- name: Grep for {{{ pkg_manager }}} repo section names
+  shell: grep -HEr '^\[.+\]' -r ./etc/yum.repos.d/
```
jan-cerny: the dot before /etc shouldn't be there

caitelatte: Oops, thanks for the catch @jan-cerny - left over from testing. Just pushed a fix.

PR #5097, opened 2020-01-07
Fix broken OCIL in STIG overlay guide

isimluk:
STIG overlay transformation expected OCIL to be inlined in XCCDF. This is no
longer true. This is simple bugfix that brings overlay transition to par with
latest XCCDF whereas OCIL is extracted to separate file.

### DIFF
--- a/cmake/SSGCommon.cmake
+++ b/cmake/SSGCommon.cmake
@@ -1105,7 +1105,7 @@ macro(ssg_build_html_stig_tables PRODUCT STIG_PROFILE)
     )
     add_custom_command(
         OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/unlinked-stig-xccdf.xml"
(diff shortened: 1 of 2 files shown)


isimluk:
Note: I just noticed this bug was very well visible in `build/tables/table-rhel7-stig.html`

jan-cerny:
@isimluk Cool. But I like the `If , this is a finding.` very much :smiley:

jan-cerny:
@isimluk Do you think we could do something with the formatting of the OCIL text?

isimluk:
> @isimluk Do you think we could do something with the formatting of the OCIL text?

No idea. Do You know who consumes `build/tables/table-rhel7-stig.html`?

jan-cerny:
@isimluk I don't know who consumes it, but it might be important because it's shipped in RHEL.

jan-cerny:
@isimluk Regarding the formatting, the OCIL isn't formatted in the datastream as well, so it looks liket we don't process the "rule.yml"s correctly.

Also I have noticed funny content eg. ` Is it the case that it is not the case? ` :smile_cat:

PR #5100, opened 2020-01-09
Make the srg_group.xml work again

isimluk:
# A picture is worth 1000 words
scrot

See also `tables/table-rhel7-srgmap-flat.html`.

# Overview

This re-implements SRG overlay.

SRG overlays have been broken since the repo was converted from the shorthand
format. This brings the support for `overlays/srg_support.xml` back to the game.

Overlay is used to generate more complete SRG Mapping table. The mapping tables
provides overview how the SRG requirements are implemented in scope of the
product.


isimluk:
Thanks!

PR #5101, opened 2020-01-09
Switch overlay for CCI-0000015 within RHEL

isimluk:
## RHEL supports account management functions

hence the CCI-000015 should be labeled as inherently-meets.

Quoting @tedbrunell:

        (...) we support account creation in RHEL. This should probably be
        marked at "Applicable - inherently meets" instead of "Guidance does not
        meet this requirement due to impracticality or scope".


shawndwells:
Merged https://github.com/ComplianceAsCode/content/pull/5100. Should be able to rebase when ready.

PR #5103, opened 2020-01-10
Fix: Yield empty table columns explicitly

isimluk:
Ommiting table fields leads these fields do not have background color.

See second row in [this image](https://user-images.githubusercontent.com/6666052/72068901-efc44500-32dd-11ea-8367-dbfa37872421.jpg). Grey background disappears in the middle of the row.


jan-cerny:
@isimluk Great, it fixes the gray background for me. Thanks. Please don't create branches in upstream repository.

PR #5105, opened 2020-01-12
E8: ensure there is a single account with uid zero

shaneboulden:
The [Essential 8 in Linux Environments guide](https://www.cyber.gov.au/publications/essential-eight-in-linux-environments) specifically mentions organisations should check for multiple accounts with UID zero. 

This PR adds the `accounts_no_uid_except_zero` check to the E8 profiles.


matejak:
LGTM, merging!

PR #5107, opened 2020-01-13
Add security autoupdates to the RHEL8 E8 profile.

matejak:
The E8 requires that systems are continuously updated, so a rule that ensures so should be part of the profile.


yuumasato:
LGTM, thanks!

PR #5109, opened 2020-01-13
Add a shell lineinfile template

matejak:
Extended our jinja macros to support shell-style assignments for Bash, Ansible and OVAL.

Added and documented a template that uses those macros.
Macros have `no_quotes` option for cases when quotes may be kept as part of the value, e.g. in `.env` files or in some cases in the `/etc/sysconfig` config files.


shared/macros-oval.jinja
```diff
@@ -17,8 +17,9 @@
     - multi_value (boolean): If set, it means that the parameter can accept multiple values and the expected value must be present in the current list of values.
     - missing_config_file_fail (boolean): If set, the check will fail if the configuration is not existent in the system.
     - section (String): If set, the parameter will be checked only within the given section defined by [section].
+    - quotes (String): If non-empty, one level of matching quotes is considered when checking the value. See comment of oval_line_in_file_state for more info.
 #}}
-{{%- macro oval_check_config_file(path='', prefix_regex='^[ \\t]*', parameter='', separator_regex='[ \\t]+', value='', missing_parameter_pass=false, application='', multi_value=false, missing_config_file_fail=false, section='') -%}}
+{{%- macro oval_check_config_file(path='', prefix_regex='^[ \\t]*', parameter='', separator_regex='[ \\t]+', value='', missing_parameter_pass=false, application='', multi_value=false, missing_config_file_fail=false, section='', quotes='') -%}}
```
jan-cerny: Do you think that we should extend the tests of this macro (`macro oval_check_config_file`) with the new parameter to make sure it's correctly propagated and not ignored?

shared/macros-ansible.jinja
```diff
@@ -141,6 +141,26 @@
 {{{ ansible_set_config_file(msg, "/etc/ssh/sshd_config", parameter, value=value, create="yes", prefix_regex='(?i)^\s*', validate="/usr/sbin/sshd -t -f %s", insert_before="^[#\s]*Match") }}}
 {{%- endmacro %}}
 
+{{#
+  High level macro to set a value in a shell-related file that contains var assignments. This
+  takes these values: msg (the name for the Ansible task), path to the file, a parameter to set
+  in the configuration file, and the value to set it to. We specify a case
+  sensitive comparison in the prefix since this is used to deduplicate since
```
jan-cerny: I think there is something missing at the end of the line.

Where is the case sensitive comparison done?

shared/macros-ansible.jinja
```diff
@@ -141,6 +141,26 @@
...
+{{% if no_quotes -%}}
+{{%- else -%}}
+{{%- set quotes = "\"'" -%}}
+  {{% if "$" in value %}}
+  {{% set value = '"%s"' % value %}}
+  {{% else %}}
+  {{% set value = "'%s'" % value %}}
+  {{% endif %}}
+{{%- endif -%}}
+{{{ ansible_set_config_file(msg, path, parameter, value=value, create="yes", prefix_regex='^\s*', validate="/usr/bin/bash -n %s", insert_before="^[#\s]*Match") }}}
```
jan-cerny: What is `insert_before="^[#\s]*Match"`? Is it a copy-paste omission?

ssg/templates.py
```diff
@@ -290,6 +290,30 @@ def sshd_lineinfile(data, lang):
     return data
 
 
+@template(["ansible", "bash", "oval"])
+def shell_lineinfile(data, lang):
+    value = data["value"]
+    if value[0] in ("'", '"') and value[0] == value[1]:
```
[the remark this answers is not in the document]
matejak: Yes, good catch.

shared/macros-oval.jinja
matejak: There is a list for quotes and no-quotes OVAL, is that what you meant, or do you still miss some other tests?

shared/macros-oval.jinja
```diff
@@ -173,12 +174,21 @@
   This macro can take two parameters:
     - value (String): The value to be checked. This can also be a regular expression (e.g: value1|value2 can match both values).
     - multi_value (boolean): If set, it means that the parameter can accept multiple values and the expected value must be present in the current list of values.
+    - quotes (String): If non-empty, one level of matching quotes is considered when checking the value. Specify one or more quote types as a string.
+      For example, for shell quoting, specify quotes="'\""), which will make sure that value, 'value' and "value" are matched, but 'value" or '"value"' won't be.
 #}}
-{{%- macro oval_line_in_file_state(value='', multi_value='') -%}}
+{{%- macro oval_line_in_file_state(value='', multi_value='', quotes='') -%}}
```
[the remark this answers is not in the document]
matejak: There is an almost identical macro that has the same name defined in the crypto policy check (`harden_ssh_client_crypto_policy`). That macro allows to instantiate multiple states in the OVAL def, as it uses the supplied parameter (that wouldn't be needed otherwise) as part of the state ID.
Unifying those two macros would be outside of scope of this PR.

shared/macros-ansible.jinja
matejak: This was a copy-paste error.

shared/macros-ansible.jinja
matejak: Thanks to pointing this out, I have fixed that after looking at the [Ansible documentation](https://docs.ansible.com/ansible/latest/modules/lineinfile_module.html).

shared/macros-oval.jinja
```diff
@@ -232,6 +242,25 @@
...
 
+{{#
+  High level macro to check if a particular shell variable is set.
+  This macro can take five parameters:
+    - path (String): Path to the file.
+    - parameter (String): The shell variable name.
+    - value (String): The variable value WITHOUT QUOTES.
+    - missing_parameter_pass (boolean): If set, the check will also pass if the parameter is not present in the configuration file (default is applied).
+    - multi_value (boolean): If set, it means that the parameter can accept multiple values and the expected value must be present in the current list of values.
+    - missing_config_file_fail (boolean): If set, the check will fail if the configuration is not existent in the system.
```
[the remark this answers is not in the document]
matejak: updated here and in other places as well.

shared/templates/template_ANSIBLE_shell_lineinfile
```diff
@@ -0,0 +1,21 @@
...
+{{%- else -%}}
+	{{% set msg = "Setting shell-quoted " ~ msg %}}
+{{%- endif -%}}
+{{{
+    ansible_shell_set(
+        msg=msg,
+        path=PATH,
+        parameter=PARAMETER,
+        value=VALUE,
+	no_quotes=NO_QUOTES
```
jan-cerny: indentation

shared/macros-ansible.jinja
```diff
@@ -141,6 +141,28 @@
...
+{{% if no_quotes -%}}
+{{%- else -%}}
+{{%- set quotes = "\"'" -%}}
+  {{% if "$" in value %}}
+  {{% set value = '"%s"' % value %}}
+  {{% else %}}
+  {{% set value = "'%s'" % value %}}
+  {{% endif %}}
+{{%- endif -%}}
+{{{ ansible_set_config_file(msg, path, parameter, value=value, create="yes", prefix_regex='^\s*', validate="/usr/bin/bash -n %s", insert_before="^# " ~ parameter) }}}
```
jan-cerny: add separator and separator_regex

shared/macros-oval.jinja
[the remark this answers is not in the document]
jan-cerny: That's unfortunate conflict, we probably should rename the custom macro.

shared/macros-oval.jinja
jan-cerny: You add a lot of tests for oval_check_shell_file, which is great. But, you modify also the parent macro oval_check_config_file, this addition is tested only indirectly. Do you think that we should add this new option also to test cases that call "oval_check_config_file"?

docs/manual/developer_guide.adoc
```diff
@@ -1591,6 +1591,28 @@ service_enabled::
...
+* Parameters:
+** *path* - What file to check.
+** *parameter* - name of the shell variable, eg. `SHELL`.
+** *value* - value of the SSH configuration option specified by *parameter*, eg. `"/bin/bash"`. Don't pass extra shell quoting - that will be handled on the lower level.
+** *no_quotes* - If set to `"true"`, the assigned value has to be without quotes during the check and remediation doesn't quote assignments either.
+** *missing_parameter_pass* - If set to `"true"` the OVAL check will pass if the parameter is not present in the target file.
+* Languages: Ansible, Bash, OVAL
+
+For example, template invocation specifying that parameter `HISTSIZE` should be set to value `500` in `/etc/profile` would produce a check that would pass if any of following lines are present in `/etc/profile`:
+
```
jan-cerny: I think that the example section should be moved inside a bullet point otherwise it's on the same level as other templates, which kind of breaks the typography.

docs/manual/developer_guide.adoc
```diff
@@ -1591,6 +1591,28 @@ service_enabled::
...
+* `HISTSIZE=500`
+* `HISTSIZE="500"`
+* `HISTSIZE='500'`
+
+and the remediation would insert the last form (with single-quotes) if the line was not present.
+
+If the `no_quotes` would be set in the template, only the first form would go be checked for, and also inserted if not present.
```
jan-cerny: please reword: "form would go be checked for"

matejak:
Yes, an exemplary rule is incoming.

shared/templates/template_ANSIBLE_shell_lineinfile
```diff
@@ -0,0 +1,21 @@
+# platform = multi_platform_all
+# reboot = false
+# strategy = restrict
+# complexity = low
+# disruption = low
+{{% set msg = "shell-style assignment of '" ~ PARAMETER ~ "' to '" ~ VALUE ~ "' in '" ~ PATH ~ "'." -%}}
```
jan-cerny: Remove the dot.

shared/macros-oval.jinja
```diff
@@ -232,15 +242,33 @@
 {{{ oval_check_config_file("/etc/ssh/sshd_config", prefix_regex="^[ \\t]*(?i)", parameter=parameter, separator_regex='(?-i)[ \\t]+', value=value, missing_parameter_pass=missing_parameter_pass, application="sshd", multi_value=multi_value, missing_config_file_fail=missing_config_file_fail) }}}
 {{%- endmacro %}}
 
+{{#
+  High level macro to check if a particular shell variable is set.
+  This macro can take five parameters:
+    - path (String): Path to the file.
+    - parameter (String): The shell variable name.
+    - value (String): The variable value WITHOUT QUOTES.
```
jan-cerny: application and no_quotes aren't documented

shared/templates/template_OVAL_shell_lineinfile
```diff
@@ -0,0 +1,10 @@
+{{{
+oval_check_shell_file(
+	path=PATH,
+	parameter=PARAMETER,
+	value=VALUE,
+	no_quotes=NO_QUOTES,
+	missing_parameter_pass=MISSING_PARAMETER_PASS
```
jan-cerny: I think `application` should be set otherwise it generates incomplete comments in OVAL, eg. `<criteria comment=" is configured correctly" operator="OR">`. I don't know what value set it to, though.

shared/templates/template_BASH_shell_lineinfile
```diff
@@ -0,0 +1,6 @@
+# platform = multi_platform_all
+# reboot = false
+# strategy = restrict
+# complexity = low
+# disruption = low
+{{{ bash_shell_file_set(path=PATH, parameter=PARAMETER, value=VALUE, no_quotes=NO_QUOTES) }}}
```
jan-cerny: I used it as a replacement for our custom code for accounts_tmout. The template generates weird code. Notice the `^Match` regex. It probably should be `^\s*TMOUT` in my example.

```
  1 # platform = multi_platform_all
  2 # reboot = false
  3 # strategy = restrict
  4 # complexity = low
  5 # disruption = low
  6 
  7   
  8   if [ -e "/etc/profile" ] ; then
  9     LC_ALL=C sed -i "/^\s*TMOUT=/d" "/etc/profile"
 10 else
 11     touch "/etc/profile"
 12 fi
 13 cp "/etc/profile" "/etc/profile.bak"
 14 # Insert before the line matching the regex '^Match'.
 15 line_number="$(LC_ALL=C grep -n "^Match" "/etc/profile.bak" | LC_ALL=C sed 's/:.*//g')"
 16 if [ -z "$line_number" ]; then
 17     # There was no match of '^Match', insert at
 18     # the end of the file.
 19     printf '%s\n' "TMOUT='100'" >> "/etc/profile"
 20 else
 21     head -n "$(( line_number - 1 ))" "/etc/profile.bak" > "/etc/profile"
 22     printf '%s\n' "TMOUT='100'" >> "/etc/profile"
 23     tail -n "+$(( line_number ))" "/etc/profile.bak" >> "/etc/profile"
 24 fi
 25 # Clean up after ourselves.
 26 rm "/etc/profile.bak"
```

linux_os/guide/services/ssh/ssh_server/sshd_use_strong_rng/rule.yml
```diff
@@ -0,0 +1,37 @@
...
+ocil_clause: |-
+    The SSH_USE_STRONG_RNG is not set to 1 in /etc/sysconfig/sshd
+
+template:
+    name: shell_lineinfile
+    vars:
+        path: '/etc/sysconfig/sshd'
+        parameter: 'SSH_USE_STRONG_RNG'
+        value: '1'
+        no_quotes: 'true'
```
jan-cerny: The generated Ansible remediation contains an empty line with leading whitespace between `insertbefore` and `validate`.

matejak:
I have addressed bunch of style and whitespace-related issues.

linux_os/guide/services/ssh/ssh_server/sshd_use_strong_rng/rule.yml
```diff
@@ -0,0 +1,37 @@
...
+identifiers:
+    cce@rhel8: 82462-3
+
+references:
+    ospp: FIA_AFL.1
+
+ocil: |-
+    To determine whether SSH is configured to use strong entropy seed, run
+    <pre>$ sudo grep SSH_USE_STRONG_RNG /etc/sysconfig/sshd</pre>
+    If a line indicating that SSH_USE_STRONG_RNG is set to 1 is returned,
```
jan-cerny: The man page https://linux.die.net/man/1/ssh-add says it should be at least 6

jan-cerny:
Thank you very much for addressing the issues, and the template now looks good to me.

Please check if your new rule sets the correct value.

matejak:
I have double-checked the value, and improved the description as well.

jan-cerny:
Great! Thanks.

PR #5110, opened 2020-01-13
Split rule for audit sample rules according to audit component

yuumasato:
#### Description:

- Split `audit_rules_for_ospp` according to audit sample rules
- ~This is a re-submission of #5066 but without selecting the rules.~
- Check existence and contents of "audit sample rules" according to https://github.com/linux-audit/audit-userspace/blob/master/rules
  - Files relevant to OSPP policy are covered by a CaC rule, check and remediation.
  - Check and remediation is done by verbatim content.
    - Including verbatim config content within SCAP content may lead to "escaping hell", I'd like to avoid this, but I'm willing to cede on this aspect for now.

#### Notes:
- This checks for verbatim literal contenst, as requested by @shawndwells and @redhatrises
- I understand this still doesn't allow for individual tailoring per syscall, but "tailor-ability" is improved, compared to `audit_rules_for_ospp`.
  - @shawndwells proposed a way move further and  handle finer granularity and not be swamped by complexity of ordering of audit rules (can you provide branch with prototype/idea?)


redhatrises:
The OCIL statements are completely wrong. System auditors do not diff. They verify the auditing configuration.

shawndwells:
> The OCIL statements are completely wrong. System auditors do not diff. They verify the auditing configuration.

+1. The underlying OVAIL and OCIL needs to be updated to explicitly test for what is inside the files.

Worth explicitly noting that moving from evaluating ospp.rules towards the new stub files is positive and worth merging to make incremental progress.

shawndwells:
Also, since the rules are now broken out into atomic groupings ([ref upstream audit](https://github.com/linux-audit/audit-userspace/tree/master/rules)), the existing rules can be updated to evaluate the proper file vs generic audit.rules, audit.d, etc.

yuumasato:
> +1. The underlying OVAL and OCIL needs to be updated to explicitly test for what is inside the files.

@shawndwells Could you clarify how you envision this?

> Also, since the rules are now broken out into atomic groupings ([ref upstream audit](https://github.com/linux-audit/audit-userspace/tree/master/rules)), the existing rules can be updated to evaluate the proper file vs generic audit.rules, audit.d, etc.

Could you clarify what what is the `proper file` and what is the `generic audit.rules, audit.d, etc`?

yuumasato:
> evaluate _and remediate_ the proper file.
> 
> Such as https://github.com/linux-audit/audit-userspace/blob/master/rules/30-ospp-v42-1-create-success.rules vs audit.rules.

I'm confused.
@shawndwells  Are you suggesting to evaluate 
`/usr/share/audit/sample-rules/30-ospp-v42-1-create-success.rules` instead of 
`/etc/audit/rules.d/30-ospp-v42-1-create-success.rules`?

matejak:
CCEs need to be updated - check the Jenkins output for hints. CCEs used in the past PR probably made it into another files created in the process.

vojtapolasek:
@shawndwells  I've updated a description, rationale and ocil of a sample rule. I would like to apply similar principle on remaining rules. Is that acceptable?

linux_os/guide/system/auditing/policy_rules/audit_basic_configuration/rule.yml
```diff
@@ -0,0 +1,59 @@
...
+prodtype: rhel8
+
+title: 'Configure basic parameters of Audit system'
+
+description: |-
+    Perform basic configuration of Audit system.
+    Make sure that any previously defined rules are cleared, the auditing system is configured to handle sudden bursts of events, and in cases of failure, messages are configured to be directed to Syslog.
+
+    The following rules configure audit as described above:
+    <pre>
```
matejak: There are indentation/quote matching issues in the file - I recommend to use an [online YAML parser](https://yaml-online-parser.appspot.com/) to make sure that the result meets expectations.

vojtapolasek: I used yamllint to find the redundant apostrophe. Now it compiles. Does it have some wrong indentation now?

redhatrises:
> @shawndwells I've updated a description, rationale and ocil of a sample rule. I would like to apply similar principle on remaining rules. Is that acceptable?

@vojtapolasek yes! Way better. The comments in `<pre>` probably don't need to be there, but not a deal breaker.

vojtapolasek:
@shawndwells  @redhatrises  after internal discussion an alternative way of checking the rule by comparing SHA-256 checksum has been added. This makes thinks more stable because it is easier to detect if the file was modified. The implementation of the check is also easier and less prone to errors. Additionally, it improves accessibility of the checking process.

linux_os/guide/system/auditing/policy_rules/audit_basic_configuration/tests/file_not_identical.fail.sh
```diff
@@ -0,0 +1,4 @@
+# profiles = xccdf_org.ssgproject.content_profile_ospp
+
+cp /usr/share/audit/sample-rulesaú10-base-config.rules /etc/audit/rules.d/
```
jan-cerny: @vojtapolasek This line contains ú after sample-rules instead of slash.

shawndwells:
> @shawndwells @redhatrises after internal discussion an alternative way of checking the rule by comparing SHA-256 checksum has been added. This makes thinks more stable because it is easier to detect if the file was modified. The implementation of the OVAL check is also easier and less prone to errors. Additionally, it improves accessibility of the checking process. Are you OK with that?

Looks like hashes are still being used. Instead of hashes, the underlying OVAL templates can be updated. For example, updating https://github.com/ComplianceAsCode/content/blob/master/shared/templates/template_OVAL_audit_rules_dac_modification#L46 to evaluate the file expected to contain DAC rules.
 
Will most likely require refactoring the XCCDF groups and rules.

yuumasato:
> the underlying OVAL templates can be updated. For example, updating https://github.com/ComplianceAsCode/content/blob/master/shared/templates/template_OVAL_audit_rules_dac_modification#L46 to evaluate the file expected to contain DAC rules.

--
TL;DR
- I'd rather check for verbatim lines instead of using these templates. But then, using checksums is a simpler and safer way to check verbatim lines.
- Why use a complex check to look for individual lines of a configuration piece that should be atomic?
--

And my reasoning is that these templates add complexity that is not necessary.

These individual rule templates were designed to handle differences in:
- operating system bit computing size
- string key used for logging `-F key=.*`
- `-S` syntax differences
  - single -S: `-S sys1` 
  - single -S with multiple syscalls: `-S sys1,sys2`
  - multiple -S: `-S sys1 -S sys2`

What is the benefit of checking for all of these when the configuration piece is atomic?
The remediation will just copy the sample file from `/usr/share/audit/sample-rules/`.

And if we are going to check for the verbatim actual lines, checking their checksum is much simpler and less error prone.

linux_os/guide/system/auditing/policy_rules/audit_basic_configuration/rule.yml
```diff
@@ -0,0 +1,83 @@
...
+    cce@rhel8: 82462-3
+
+references:
+    ospp: FAU_GEN.1.1.c
+    nist: AU-2(a)
+    srg: SRG-OS-000365-GPOS-00152,SRG-OS-000475-GPOS-00220
+
+ocil_clause: 'the content or the hash of file differs or file does not exist'
+
+ocil: |-
```
yuumasato: Just to be sure this is not an oversight in the discussions.
The Rule OCIL and the description contains both the file content and the checksum.

linux_os/guide/system/auditing/policy_rules/audit_basic_configuration/rule.yml
```diff
@@ -0,0 +1,77 @@
...
+## Make this bigger for busy systems
+-b 8192
+
+## This determine how long to wait in burst of events
+--backlog_wait_time 60000
+
+## Set failure mode to syslog
+-f 1
+
+' %}}
```
yuumasato: The syntax of this Jinja variable can be finicky. I can move this to a separate file, but then, the contents won't be as much as visible.

yuumasato:
I have added a template called `file_contents`, which checks for the contents of a file as a whole, and I have used in rule `audit_basic_configuration`.
The line contents of the rule were put in a Jinja variable to be reused, and easily replaced when the file changes.

shared/templates/template_BASH_file_contents
```diff
@@ -0,0 +1,8 @@
+# platform = multi_platform_all
+# reboot = false
+# strategy = restrict
+# complexity = low
+# disruption = low
+
+cat << EOF > {{{ FILEPATH }}}
```
adelton: Should this be
```
cat << 'EOF' > {{{ FILEPATH }}}
```
?

I believe this whole approach is inherently fragile because of potential escaping issues.

yuumasato: Thank you for pointing the potential problem with escaping.
Indeed, having the actual file contents in the datastream can open up another can of worms.

We already have a similar situation with the login banners, where 1) the banner needs to be escaped for inclusion xml 2) and also escaped for bash remediation, but 3 ) the same escaping breaks Ansible remediation.

yuumasato:
@openscap-ci test this please

yuumasato:
@openscap-ci test this please

shared/templates/template_BASH_file_contents
yuumasato: The variable expansion issue has been fixed in https://github.com/ComplianceAsCode/content/pull/5110/commits/1bb775638cd694750016b8ecb567e7f3e2b103a2.

It seems we will go the fragile way for now, hopefully not for long.

yuumasato:
@shawndwells and @redhatrises I'ld love to hear from you WRT to this PR.
I suggest we go this path while we discuss how to allow for finer granularity of checking and tailoring.

shared/macros-bash.jinja
```diff
@@ -509,3 +509,14 @@ if ! grep -qE '^\s*account\s+required\s+pam_faillock\.so.*$' "{{{ pam_file }}}"
...
+
+{{#
+  Generates bash script code that puts 'contents' into a file at 'filepath'
+  Parameters:
+    - filepath - filepath of the file to check
+    - contents - contents that should be in the file
+#}}
+{{%- macro bash_file_contents(filepath='', contents='') %}}
+cat << 'EOF' > {{{ filepath}}}
+{{{ contents}}}EOF
```
matusmarhefka: Just a nitpick, but you can add a space here after both `filepath` and `contents` variables.

linux_os/guide/system/auditing/policy_rules/audit_delete_success/tests/no_rule.fail.sh
```diff
@@ -0,0 +1 @@
+rm -rf /etc/audit/rules.d/*.rules
```
matusmarhefka: Here I would rather delete only the specific file from the `/etc/audit/rules.d/` - `30-ospp-v42-4-delete-success.rules`.

tests/shared/audit/10-base-config.rules
```diff
@@ -0,0 +1,13 @@
...
+## Increase the buffers to survive stress events.
+## Make this bigger for busy systems
+-b 8192
+
+## This determine how long to wait in burst of events
+--backlog_wait_time 60000
+
+## Set failure mode to syslog
+-f 1
+
```
matusmarhefka: Just for the record, in case audit configuration is updated in the future, can this data be obtained somehow automatically from the audit package? If so how and is it documented somewhere?

yuumasato:
Hello, @tedbrunell I've updated the SRGs in the split rules, would appreciate your review.

shared/macros-bash.jinja
```diff
@@ -509,3 +509,14 @@ if ! grep -qE '^\s*account\s+required\s+pam_faillock\.so.*$' "{{{ pam_file }}}"
...
 fi
 {{%- endmacro -%}}
+
+{{#
+  Generates bash script code that puts 'contents' into a file at 'filepath'
+  Parameters:
+    - filepath - filepath of the file to check
+    - contents - contents that should be in the file
+#}}
+{{%- macro bash_file_contents(filepath = '', contents = '') %}}
```
matusmarhefka:
```suggestion
{{%- macro bash_file_contents(filepath='', contents='') %}}
```

shared/macros-bash.jinja
```diff
@@ -509,3 +509,14 @@ if ! grep -qE '^\s*account\s+required\s+pam_faillock\.so.*$' "{{{ pam_file }}}"
...
+cat << 'EOF' > {{{ filepath}}}
```
matusmarhefka:
```suggestion
cat << 'EOF' > {{{ filepath }}}
```

shared/macros-bash.jinja
```diff
@@ -509,3 +509,14 @@ if ! grep -qE '^\s*account\s+required\s+pam_faillock\.so.*$' "{{{ pam_file }}}"
...
+{{{ contents}}}EOF
```
matusmarhefka:
```suggestion
{{{ contents }}}EOF
```

vojtapolasek: Aha Now I understand it. Thanks.

tests/shared/audit/10-base-config.rules
yuumasato: I started this thing to see how automating updated of content could look like:
https://gitlab.cee.redhat.com/wsato/content-extractor

It probably can be integrated in the utils directory.

yuumasato:
@matusmarhefka Thanks for the review.

@shawndwells I hope current implementation in this PR is aligned to what we discussed in person.
Compromises were made, and it explicitly checks for the file contents of the sample audit rules.
It doesn't allow for individual syscall customization, but I believe this is a reasonable step forward.
See more in https://github.com/ComplianceAsCode/content/pull/5110#issue-362263403

I believe next step is to investigate viability of modifying the templates to look for a specific audit rule in a specific audit rule filename, as discussed with Shawn.

matusmarhefka:
@yuumasato Now it LGTM. This is a nice improvement over the `audit_rules_for_ospp` rule.

PR #5111, opened 2020-01-14
Updated reference URL for gconf

yuumasato:
#### Description:

- Updated reference URL for gconf

#### Rationale:

- The previous URL when 404.
  Contents are not the same, but this is the closest thing available.
- It's been failing on Jenkins test for a few days:
  https://jenkins.complianceascode.io/view/SCAP%20Security%20Guide/job/scap-security-guide-linkcheck/


matusmarhefka:
LGTM, thanks.

PR #5113, opened 2020-01-15
Sshd disable empty passwords

jukito:
#### Description:
- To explicitly disallow SSH login from accounts with
    empty passwords, add or correct the following line in **/etc/ssh/sshd_config**:
    ´´´PermitEmptyPasswords no´´´
    Any accounts with empty passwords should be disabled immediately, and PAM configuration
    should prevent users from being able to assign themselves empty passwords.

#### Rationale:
- Configuring this setting for the SSH daemon provides additional assurance
    that remote login via SSH will require a password, even in the event of 
    misconfiguration elsewhere.


jukito:
Sorry, incorrect PR

Issue #5098, opened 2020-01-07
rsyslog_remote_tls does not check for / mention EKU


comps:
The current e3d6f00f3e version of `rsyslog_remote_tls` as used by RHEL-8 only has:

```
    <ind:subexpression datatype="string" operation="pattern match">(?=[\S\s]*\sprotocol="tcp")(?=[\S\s]*\sTarget="[^"]+?")(?=[\S\s]*\sport="6514")(?=[\S\s]*\sStreamDriver="gtls")(?=[\S\s]*\sStreamDriverMode="1")(?=[\S\s]*\sStreamDriverAuthMode="x509/name")</ind:subexpression>
```
and its decription matches that:
```
    <pre>echo 'action(type="omfwd" protocol="tcp" Target="&lt;remote system>" port="6514"
        StreamDriver="gtls" StreamDriverMode="1" StreamDriverAuthMode="x509/name")' >> /etc/rsyslog.conf
    </pre>
```
However, both need to also include
```
streamdriver.CheckExtendedKeyPurpose="on"
```
so that the configuration snippet looks more like
```
action(type="omfwd" protocol="tcp" Target="<remote system>" port="6514"
    StreamDriver="gtls" StreamDriverMode="1" StreamDriverAuthMode="x509/name"
    streamdriver.CheckExtendedKeyPurpose="on")
```

comps:
As a reference note, this is a fairly recent rsyslog upstream feature, see https://github.com/rsyslog/rsyslog/pull/3845 .

comps:
Also note that it may be easier to use/check a file inside `/etc/rsyslog.d` instead of appending to `/etc/rsyslog.conf`.

comps:
> Also note that it may be easier to use/check a file inside `/etc/rsyslog.d` instead of appending to `/etc/rsyslog.conf`.

Nevermind, the OVAL already checks for that. Neat.

PR #5119, opened 2020-01-20
Check EKU in rsyslog remote configuration

jan-cerny:
This patch adds option "streamdriver.CheckExtendedKeyPurpose" to the
list of mandatory settings in rule rsyslog_remote_tls and requires this
option to be set to "on".  This option configures rsyslog to check for
the ExtendedKeyUsage field of the certificate.  This is a new feature in
rsyslog introduced in https://github.com/rsyslog/rsyslog/pull/3845

Fixes: #5098


yuumasato:
Thanks! :)

PR #5122, opened 2020-01-21
Don't check for path length of logs directory

yuumasato:
#### Description:

- Don't check for path length of logs directory

#### Rationale:

- The logs are not part of the tarball, nor used to build the content.

### DIFF
--- a/tests/ensure_paths_are_short.py
+++ b/tests/ensure_paths_are_short.py
@@ -15,7 +15,8 @@ def main():
     for dir_, _, files in os.walk(ssg_root):
         # Don't check for path len of log files
         # They are not shipped nor used during build
-        if "tests/logs/" in dir_:
+        current_relative_path = os.path.relpath(dir_, ssg_root)
+        if current_relative_path.startswith("tests/logs/"):
             continue
         for file_ in files:
             path = os.path.relpath(os.path.join(dir_, file_), ssg_root)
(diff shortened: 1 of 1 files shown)


tests/ensure_paths_are_short.py
```diff
@@ -13,6 +13,10 @@ def main():
     ssg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
     max_path = ""
     for dir_, _, files in os.walk(ssg_root):
+        # Don't check for path len of log files
+        # They are not shipped nor used during build
+        if "tests/logs/" in dir_:
```
ggbecker: I wonder if something like `linux_os/guide/tests/logs/.../` will also be skipped.

yuumasato: It will be skipped.
```
>>> path = 'linux_os/guide/tests/logs/.../'
>>> "tests/logs/" in path
True
```

ggbecker: maybe something like this:
```suggestion
        current_relative_path = os.path.relpath(dir_, ssg_root)
        if current_relative_path[:11] == "tests/logs/":
            continue
```

yuumasato: @ggbecker Thanks for your suggestion,
Implemented slightly different, but should work the same way.

ggbecker: Great. It works perfectly.

PR #5127, opened 2020-01-27
Add a rule for the openssl strong entropy wrapper

matejak:


linux_os/guide/system/software/integrity/crypto/openssl_use_strong_entropy/oval/shared.xml
```diff
@@ -0,0 +1,34 @@
+<def-group>
+  <definition class="compliance" id="openssl_use_strong_entropy" version="1">
+    <metadata>
+      <title>Configure Openssl to use strong entropy</title>
```
jan-cerny: It should be OpenSSL with capital O, S, S and L.

linux_os/guide/system/software/integrity/crypto/openssl_use_strong_entropy/rule.yml
```diff
@@ -0,0 +1,38 @@
...
+    save the following shell snippet to the <tt>/etc/profile.d/cc-config.sh</tt>:
+    <pre>
+    {{{ openssl_strong_entropy_config_file() | indent(4) }}}
+    </pre>
+
+rationale: |-
+    The <tt>openssl</tt> default configuration uses less robust entropy sources for seeding.
+    The referenced script is sourced to every login shell, and it transparently adds an option
+    that enforces strong entropy to every <tt>openssl</tt> invocation,
+    which makes <tt>openssl</tt> more secure by default.
```
jan-cerny: This feels vague. What is the default entropy source and why is it weak?

matejak:
I have removed the indent, and it works. Tests don't pass, as they expect the file to contain blank lines, but the remediation result is syntactically correct.

linux_os/guide/system/software/integrity/crypto/openssl_use_strong_entropy/rule.yml
matejak: That's the best I was able to come up with. I have asked @comps for help.

comps: I don't think we should differentiate between "less robust" or "weak" or "strong". The default is good enough for any reasonable use case (to the point that credit card companies use it), this "strong entropy source" really just means using a SP800-90A compliant DRBG.

linux_os/guide/system/software/integrity/crypto/openssl_use_strong_entropy/bash/shared.sh
```diff
@@ -0,0 +1,5 @@
+# platform = Red Hat Enterprise Linux 8
+
+cat > /etc/profile.d/cc-config.sh <<- 'EOM'
```
comps: I would call it `openssl-rand.sh` or something like that. `cc-config.sh` seems like something that could conflict with some other rule / logic.

shared/macros.jinja
```diff
@@ -618,10 +618,42 @@ ocil_clause: "the correct value is not returned"
...
-        When selecting this rule in a profile, 
+        When selecting this rule in a profile,
         {{%- if why %}}
             make sure that rule with ID <code>{{{ rule_id }}}</code> is selected as well: {{{ why }}}
         {{%- else %}}
             rule <code>{{{ rule_id }}}</code> has to be selected as well.
         {{%- endif %}}
 {{% endmacro %}}
+
+{{% macro openssl_strong_entropy_config_file() -%}}
```
yuumasato: What is the reason to define a macro? Wouldn't a variable suffice?

linux_os/guide/system/software/integrity/crypto/openssl_use_strong_entropy/rule.yml
jan-cerny: Thanks @comps for explanation.

shared/macros.jinja
vojtapolasek: The variable could be theoretically used, but you would need to take care about escaping characters, in particular ' or " because you use one of those characters to enclose the variable value. Macro is foolproof in this case.

linux_os/guide/system/software/integrity/crypto/openssl_use_strong_entropy/tests/file_modified.fail.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+# platform = Red Hat Enterprise Linux 8
+# profiles = xccdf_org.ssgproject.content_profile_ospp
+
+echo "wrong data" > /etc/profile.d/cc-config.sh
```
comps: Forgot to rename this one?

jan-cerny:
@vojtapolasek Please remove the merge commit. Also, remove the conflicts. You can use interactive rebase and force push to do that.

shared/macros.jinja
yuumasato: Ok, that is an interesting side effect of using macros.
Long term wise, I'm not sure if we should rely on this aspect.

linux_os/guide/system/software/integrity/crypto/openssl_use_strong_entropy/ansible/shared.yml
```diff
@@ -0,0 +1,12 @@
+# platform = multi_platform_all
+# reboot = false
+# strategy = restrict
+# complexity = low
+# disruption = low
+
+- name: "copy a file with shell snippet to configure openssl strong entropy"
```
yuumasato: `copy` is the Ansible module in use, but we are not copying content but dumping / putting.
Suggestion: `Put (or setup) a shell snippet to configure OpenSSL to always use strong entropy`.

Also, mind the OpenSSL capital letters. :)

linux_os/guide/system/software/integrity/crypto/openssl_use_strong_entropy/rule.yml
```diff
@@ -0,0 +1,34 @@
+documentation_complete: true
+
+# TODO: The plan is not to need this for RHEL>=8.4
+prodtype: rhel8
+
+title: 'OpenSSL uses strong entropy source'
+
+description: |-
+    To set up an <tt>openssl</tt> wrapper that adds a <tt>-rand /dev/random</tt> option to the <tt>openssl</tt> invocation,
```
yuumasato: Before this line I'd say that the wrapper is a way to achieve fulfill this rule. The goal of the rule is not to setup a wrapper, but to make sure OpenSSL is using a strong entropy source.

Raw suggestion: `By default, OpenSSL doesn't always use a SP800-90A compliant random number generator.
A way to configure OpenSSL to always use a strong source is to setup a wrapper that adds ...`

linux_os/guide/system/software/integrity/crypto/openssl_use_strong_entropy/rule.yml
```diff
@@ -0,0 +1,34 @@
...
+identifiers:
+    cce@rhel8: 82721-2
+
+references:
+    ospp: FIA_AFL.1
+
+ocil: |-
+    To determine whether the <tt>openssl</tt> wrapper is configured correctly,
+    make sure that the <tt>/etc/profile.d/openssl-rand.sh</tt> file contains contents
+    that are included in the rule's description.
```
yuumasato: Similar to comment above, the goals is to make sure OpenSSL always use the compliant source of entropy. And the wrapper is a way to achieve this.

linux_os/guide/system/software/integrity/crypto/openssl_use_strong_entropy/rule.yml
```diff
@@ -0,0 +1,34 @@
+documentation_complete: true
+
+# TODO: The plan is not to need this for RHEL>=8.4
+prodtype: rhel8
+
+title: 'OpenSSL uses strong entropy source'
+
+description: |-
+    To set up an <tt>openssl</tt> wrapper that adds a <tt>-rand /dev/random</tt> option to the <tt>openssl</tt> invocation,
+    save the following shell snippet to the <tt>/etc/profile.d/openssl-rand.sh</tt>:
```
yuumasato: I'd say `save the following shell snippet exactly "as is"`.
My thinking is: Someone copy pasting the snippet may change indentation, comments or blank lines.

PR #5128, opened 2020-01-27
Add moderate profile

JAORMX:
This creates the moderate profile. This is where we can start enabling
the NIST moderate controls as we go.

It's a copy of the coreos-ncp.profile


redhatrises:
@openscap-ci add to whitelist

PR #5132, opened 2020-01-27
Ansible remediation for configure_openssl_crypto_policy

mildas:
#### Description:
Add ansible remediation that uses `lineinfile` module.

`ini_file` module wasn't possible to use, because it automatically adds `=` delimiter and according to https://github.com/ansible/ansible/issues/40786 it is not standard INI file without `=`.

#### Rationale:
The rule is in `ospp` and `pci-dss` profiles.


matejak:
Re-raised the issue in https://github.com/ansible/ansible/issues/66853 However, test scenarios work, so merging, thank you for the PR!

PR #5137, opened 2020-01-28
Add support to Ignition remediation type.

ggbecker:
#### Description:

- Add support to Ignition remediation type

How to generate the remediation meanwhile OpenSCAP doesn't support this specific type:

`oscap xccdf generate fix --profile coreos-ncp --template urn:xccdf:fix:script:ignition  build/ssg-ocp4-ds.xml`

### DIFF
--- a/linux_os/guide/system/accounts/accounts-restrictions/root_logins/no_direct_root_logins/ignition/shared.yml
+++ b/linux_os/guide/system/accounts/accounts-restrictions/root_logins/no_direct_root_logins/ignition/shared.yml
@@ -1,0 +1,18 @@
+# platform = multi_platform_all
+apiVersion: machineconfiguration.openshift.io/v1
+kind: Ignition
+metadata:
+  labels:
+    machineconfiguration.openshift.io/role: worker
+  name: 50-worker-empty-securetty
+spec:
+  config:
+    ignition:
+      version: 2.2.0
+    storage:
+      files:
+      - contents:
+          source: data:,
+        filesystem: root
+        mode: 0600
+        path: /etc/securetty
--- a/ocp4/profiles/coreos-ncp.profile
+++ b/ocp4/profiles/coreos-ncp.profile
@@ -77,6 +77,7 @@ selections:
     #- sshd_set_keepalive
     #- sshd_enable_warning_banner
     #- sshd_rekey_limit
+    
 
     # Time Server
     - chronyd_client_only
--- a/cmake/SSGCommon.cmake
+++ b/cmake/SSGCommon.cmake
@@ -273,8 +273,8 @@ macro(ssg_build_ansible_playbooks PRODUCT)
 endmacro()
 
 macro(ssg_build_remediations PRODUCT)
-    message(STATUS "Scanning for dependencies of ${PRODUCT} fixes (bash, ansible, puppet and anaconda)...")
-    _ssg_build_remediations_for_language(${PRODUCT} "bash;ansible;puppet;anaconda")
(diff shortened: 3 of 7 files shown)


ocp4/profiles/coreos-ncp.profile
```diff
@@ -77,6 +77,7 @@ selections:
     #- sshd_set_keepalive
     #- sshd_enable_warning_banner
     #- sshd_rekey_limit
+
```
yuumasato: nitpick: Unchaning change. But I'd rather not wait for Jenkins to test it again.

isimluk:
Nice! LGTM. :+1:

linux_os/guide/system/accounts/accounts-restrictions/root_logins/no_direct_root_logins/ignition/shared.yml
```diff
@@ -0,0 +1,18 @@
+# platform = multi_platform_all
+apiVersion: machineconfiguration.openshift.io/v1
+kind: Ignition
```
jhrozek: The Kind should be "MachineConfig", not "Ignition". 
Sorry, I know this is confusing. Ignition is the yaml-based language that is used under spec.config. But together with the other things, the spec makes up a "MachineConfig" Kubernetes type.

jhrozek: Great work overall by the way. I'm testing the result with the operator now.

ggbecker: It was actually a find-replace error. I will fix that.

shawndwells: @jhrozek while OCP may aggregate these snippets into MachineConfig's, would it be reasonable to expect use cases where other Kubernetes distributions may take the Ignition scripts for other purposes?

Or in other words -- should these remain to be called Ignition scripts, and let downstreams such as OpenShift convert to product-specific terms like Machine Configs? Trying to avoid having remediation snippets for OCP vs Docker vs Pivotal vs Rancher if possible, by letting Ignition be the commonality between them.

shawndwells: updated: chatted with @jhrozek and @JAORMX in person. Lose agreement that keeping ignition namespace might make sense as it will simplify code in the compliance operator.

@jhrozek double checking compliance-operator code and reporting back.

JAORMX: Right, if we were to keep this as ignition, we would then want to keep only from line 10 to 18. And remove the headers. Those should be handled by the compliance-operator as an implementation detail.

shawndwells:
@ggbecker Code was reviewed and jenkins passed - this can be merged.

To enable others to begin building remediation content this afternoon, what do you think about merging this now and having you update the namespace/naming back to ignition this afternoon? 

If agreeable, feel free to create a ticket to track the work and self-merge.

ggbecker:
> @ggbecker Code was reviewed and jenkins passed - this can be merged.
> 
> To enable others to begin building remediation content this afternoon, what do you think about merging this now and having you update the namespace/naming back to ignition this afternoon?
> 
> If agreeable, feel free to create a ticket to track the work and self-merge.

I agree. The code part which enables the capability to use a new type of remediation is seems to correct. We just need to figure out what is the format of the remediation, eg. which kind of data is expected to be there.

PR #5138, opened 2020-01-28
Enable rules that cover AU-9 better in OCP4 moderate profile

nkinder:
This enables a few rules that pertain to the AU-9 control in the OCP4
moderate profile.  Note that the logrotate check will need to be updated
in a future patch to allow a variable for the rotation cadence, as it is
currently hardcoded to 'daily'.  AU-9 does not require a specific rotation
cadence.

#### Description:

#### Rationale:


shawndwells:
Jenkins passed. Merging.

PR #5139, opened 2020-01-28
Made the affected OVAL metadata macro-generated in shared OVALs

matejak:
As we have filename-based distinction between different products
i.e. shared.xml vs fedora.xml, manual enumeration inside shared
check files is redundant and error-prone.

So far done for rules that are included in OCP4 profiles, as we don't have consistent rule applicability and prodtypes in all rules, so it can happen that the affected autogeneration hits extend definitions.


shawndwells:
Waiting on jenkins to merge.

shawndwells:
Jenkins complete. Merging.

PR #5142, opened 2020-01-28
Add Machine config remediation for rule: no_empty_passwords

ggbecker:
#### Description:

- Change Ignition remediation to `kind: MachineConfig` and adds another one for `no_empty_passwords`.
#### Rationale:

- The Compliance-Operator POC needs at least two rules with remediation to work.

Note: Exact format of the remediation is still a working progress so for now we are using `kind: MachineConfig`


jhrozek:
As discussed in person, @ggbecker would also amend the PR to drop the metadata section from both remediations.

ggbecker:
> As discussed in person, @ggbecker would also amend the PR to drop the metadata section from both remediations.

Just did: 81ae4284b83c5b9b915bd60d9462fc72db0bef3a

jhrozek:
That was quick, let me try the PR out..

linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/ignition/shared.yml
```diff
@@ -1,10 +1,6 @@
 # platform = multi_platform_wrlinux,multi_platform_rhel,multi_platform_fedora,multi_platform_ol,multi_platform_rhv,multi_platform_ocp
```
jhrozek: The only question about this whole PR is that the `no_empty_passwords` contains an enumeration of all the platforms..

linux_os/guide/system/accounts/accounts-restrictions/root_logins/no_direct_root_logins/ignition/shared.yml
```diff
@@ -1,10 +1,6 @@
 # platform = multi_platform_all
```
jhrozek: ..but the `no_direct_root_logins` uses `multi_platform_all`. Which of them is correct?

jhrozek: Oh please ignore me, the bash remediations use the same `platform` definitions.

linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/ignition/shared.yml
yuumasato: It can be ` # platform = multi_platform_all `. The `rule.yml` for `no_empty_passwords` doesn't define any `prodtype`.

jan-cerny:
Thanks @jhrozek for the review. ACK.

jan-cerny:
@ggbecker @jhrozek Could we refrain from the `%20` usage? It isn't readable and therefore harder to review. Can we store the code in git without the escape symbols and have a Jinja macro that adds the escape symbols during the build process?

jhrozek:
On Wed, Jan 29, 2020 at 01:31:40AM -0800, Jan Černý wrote:
> @ggbecker @jhrozek Could we refrain from the `%20` usage? It isn't readable and therefore harder to review. 

Ignition as a format stores the content of the file urlencoded. So
whatever we push to the OCP cluster via MC objects has to use this
strange format.

> Can we store the code in git without the escape symbols and have a Jinja macro that adds the escape symbols during the build process?

Perhaps? It would be nice and more readable, but I'm not familiar enough
with the scap build system to tell if that's feasible.

yuumasato:
>  Can we store the code in git without the escape symbols and have a Jinja macro that adds the escape symbols during the build process?

OTOH, storing the content urlencoded may be safer (in terms of escaping and newlines) than plain text.

PR #5143, opened 2020-01-28
add new CCE allocation into pool

shawndwells:
Additional CCE allocations from NIST.

+cc @redhatrises


matejak:
Thanks, much appreciated.

PR #5146, opened 2020-01-28
ocp4/moderate: Remove check for AIDE package

JAORMX:
AIDE will run in openshift as an operator, so no need to check the
package in the node.


redhatrises:
Ack

PR #5152, opened 2020-01-29
Update Docker content

konstruktoid:
#### Description:

This PR updates the documentation regarding building and running the Docker images, it also removes CentOS 6 and replaces it with CentOS 8.


docs/manual/developer_guide.adoc
```diff
@@ -399,14 +399,15 @@ make vendor-zipfile
...
 
-Use a suitable Dockerfile present in the
+Find a suitable Dockerfile present in the
 link:https://github.com/ComplianceAsCode/content/tree/master/Dockerfiles[Dockerfiles]
-directory and build the image.
+directory, copy the file to the top directory and build the image.
 This will take care of the build environment and all necessary setup.
 
 ```bash
+cp Dockerimages/ubuntu ./Dockerfile
```
jan-cerny: Instead of copying the Dockerfile you can use `--file Dockerfiles/ubuntu` in the `docker` command. Note also that the directory name is `Dockerfiles`, not `Dockerimages`.

Dockerfiles/centos8
```diff
@@ -0,0 +1,31 @@
...
+
+ENV OSCAP_USERNAME oscap
+ENV OSCAP_DIR content
+ENV BUILD_JOBS 4
+
+RUN true \
+  && dnf -y upgrade \
+  && dnf -y install cmake openscap-utils python3-jinja2 python3-pip python3-pyyaml \
+  && dnf -y --enablerepo=PowerTools install ninja-build \
+  && pip3 install argparse \
```
jan-cerny: Isn't argparse a part of python3-libs? Is there any reason to install it by pip?

docs/manual/developer_guide.adoc
konstruktoid: Thanks for catching that. Fixed.

Dockerfiles/centos8
konstruktoid: So true, just couldn't find `python3-libs` earlier for some reason, `python3-pip` is however still required.

jan-cerny:
Thank you very much for your contribution.

PR #5153, opened 2020-01-29
Create script to build datastream images from source

JAORMX:
This script runs the build_product script (taking the product as a
parameter), and it pushes the `ssg-$PRODUCT-ds.xml` file to a container.

The container is created by using a BuildConfig and an ImageStream,
these are updated by pushing the local XML file to the container build.

This is just one step of the build process which helps with local
development. In further iterations, I'll create a BuildConfig that builds
inside a container as opposed to requiring a local file.

Example:

    ./utils/build_ds_container.sh ocp4

### DIFF
--- a/ocp-resources/ds-build.yaml
+++ b/ocp-resources/ds-build.yaml
@@ -1,0 +1,28 @@
+kind: ImageStream
+apiVersion: image.openshift.io/v1
+metadata:
+  name: "openscap-$PRODUCT-ds" 
+  namespace: openshift-compliance
+---
+kind: BuildConfig
+apiVersion: build.openshift.io/v1
+metadata:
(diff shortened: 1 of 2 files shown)


JAORMX:
Note that this makes some assumptions (e.g. the naming scheme of the ds file is ssg-<product name>-ds.xml). This works for ocp4 and that's what I tested. I haven't tested this for rhel yet.

JAORMX:
Just tested it with RHEL and did some fixes.

jan-cerny:
@JAORMX Great! Thanks. Please move it to `utils` directory or elsewhere.

JAORMX:
@jan-cerny done! thanks for the feedback

PR #5160, opened 2020-02-03
Add Kickstart file for STIG profile

yuumasato:
#### Description:

- Add Kickstart file for STIG profile
  - Based on OSPP KS


matusmarhefka:
LGTM, thanks.

PR #5165, opened 2020-02-04
utils: A helper script to generate ignition remediations for files

jhrozek:
#### Description:

Adds a script that attempts to help with ignition creating and review by
offering an encode and decode actions that can either generate an ignition
file based on a prepared file or printing out how the file would look like
based on the ignition snippet.

#### Rationale:

It is tedious to craft the ignition remediations for files, mostly due
to having to urlencode the file input. What is even more difficult is to
see the resulting file.

The script is not exactly a pinnacle of Python programming but it seems
to do the job for me. Please see the --help output of the script for examples
and reference of the options.

### DIFF
--- a/utils/ignition-remediation.py
+++ b/utils/ignition-remediation.py
@@ -1,0 +1,306 @@
+#!/usr/bin/env python3
+
+import errno
+import os
+import sys
+import argparse
+import urllib.parse
+from string import Template
+
+description = \
+"""
+Encode a given file to an ignition based remediation or decode an ignition
+based file remediation to the resulting file.
+
+Encoding
+========
+To encode a file, you need to provide a number of attributes. At the minimum,
+you need to provide the version of the file that would be applied with
+MachineConfigs. Then you also need to provide the path where the file would
+be uploaded to using the MC and the file mode. Finally, either provide the
+platforms the remediation applies to or provide the rule name. If the rule
+name is provided, then the script reads the platforms from existing
+remediations. The output is either written to the rule directory if a rule
+is provided or to a file.
+
+Examples
+--------
+    * Encode a remediation based on a "golden file" at /tmp/chrony.conf. The
+      remediation will replace the file at /etc/chrony.conf with the mode 0644.
+      The resulting MachineConfig would be automatically written to the
+      directory where the rule chronyd_no_chronyc_network is stored. The
+      platforms that the remediation is applied to would be read from the
+      existing bash remediations.
+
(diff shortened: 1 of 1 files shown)


isimluk:
LGTM. This is good start! Thanks!

Do You think we should use script inside the build system or should each rule author run this by hand? (This may be resolved in subsequent PRs, but just trying to get your opinion...).

JAORMX:
Perhaps it would be useful to have the html results have a decoded version of the files that are being changed. That would surely make it easier to both review and audit. This could be done as subsequent work.

jhrozek:
On Wed, Feb 05, 2020 at 12:38:00AM -0800, Šimon Lukašík wrote:
> LGTM. This is good start! Thanks!
> 
> Do You think we should use script inside the build system or should each rule author run this by hand? (This may be resolved in subsequent PRs, but just trying to get your opinion...).

(Sorry for the late reply, I'm still trying to catch on after last week
while trying to fix our CI at the same time)

I'm not sure myself. At the moment I just wanted to scratch an itch.

It might be nice to have the not-encoded files stored somewhere in the repo,
but you'd still have to store the metadata about the file somewhere. Could
the remediation itself then be a stub that references the files and the
remediation itself would be generated during build?

Something like:

apiVersion: machineconfiguration.openshift.io/v1
kind: MachineConfig
spec:
  config:
    ignition:
      version: 2.2.0
    storage:
      files:
      - contents:
          source: %myfile1%
        filesystem: root
        mode: 0644
        path: /etc/myfile.conf

Then the build system would look for myfile1 in the remediation folder,
encode its contents and use them instead of %myfile1% ?

isimluk:
> It might be nice to have the not-encoded files stored somewhere in the repo, but you'd still have to store the metadata about the file somewhere. Could the remediation itself then be a stub that references the files and the remediation itself would be generated during build?

:+1: That was my line of thinking as well. I think we can try to iterate with current state of things and learn the best approach during the work on policies.

jan-cerny:
I think we can have a Jinja2 macro that would load the not-encoded file and encode it.

Then, we could have another Jinja 2 macro that would contain the stub and fill in the metadata values.

We can then create a new [SSG Template](https://github.com/ComplianceAsCode/content/blob/master/docs/manual/developer_guide.adoc#templating) which would use this macro, but the problem is that some of the rules are already templated and in the current implementation of the engine a single rule can't use multiple templates, which means the engine would have to be extended.

PR #5166, opened 2020-02-04
fix regex in accounts_passwords_pam_faillock_deny

vojtapolasek:
#### Description:

oval is updated to ignore mentions of pam_unix in commented lines
test is updated to showcase that it works also for inline comments

#### Rationale:

In case some comment contained string "pam_unix", a rule might have been evaulated as passed although the PAM configuration was not correct.


jan-cerny:
@vojtapolasek Thanks, nice catch!

PR #5174, opened 2020-02-05
Add "Scientific" to Ensure GPG check Enabled for Local Packages check

opoplawski:
#### Description:

- Add "Scientific" to Ensure GPG check Enabled for Local Packages check

#### Rationale:

- We use Scientific Linux


jan-cerny:
@openscap-ci ok to test

jan-cerny:
@opoplawski Thank you very much for your patch!

Issue #5172, opened 2020-02-05
A few rules part of STIG profile lack SRG mapping


yuumasato:
#### Description of problem:

Following rules lack SRG mapping:
##### RHEL8
- package_gnutls-utils_installed
- package_nss-tools_installed
##### RHEL7
- dconf_db_up_to_date

#### SCAP Security Guide Version:
0.1.48

This is a follow up from #5170

PR #5177, opened 2020-02-06
Map missing SRG rules

yuumasato:
#### Description:

- Add missing SRG references

Fixes: #5172 

Thanks @tedbrunell for the reference.


vojtapolasek:
SRG mappings make sense for me. Thank you for this.

PR #5178, opened 2020-02-06
create checks for (grub2|uefi)_no_removeable_media

vojtapolasek:
#### Description:

Created oval checks for rules grub2_no_removeable_media and uefi_no_removeable_media. Created test only for one instance as the checks are the same, only the file path is different. Fixed rule titles, although Boat Loader is nice, it is not correct.

#### Rationale:

Rules were missing checks.

### DIFF
--- a/linux_os/guide/system/bootloader-grub2/grub2_no_removeable_media/rule.yml
+++ b/linux_os/guide/system/bootloader-grub2/grub2_no_removeable_media/rule.yml
@@ -2,7 +2,7 @@ documentation_complete: true
 
 prodtype: wrlinux1019,rhel7,rhel8,ol7,ol8,rhv4
 
-title: 'Boat Loader Is Not Installed On Removeable Media'
+title: 'Boot Loader Is Not Installed On Removeable Media'
 
 description: |-
     The system must not allow removable media to be used as the boot loader.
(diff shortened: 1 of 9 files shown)


linux_os/guide/system/bootloader-grub2/grub2_no_removeable_media/rule.yml
```diff
@@ -2,7 +2,7 @@ documentation_complete: true
 
 prodtype: wrlinux1019,rhel7,rhel8,ol7,ol8,rhv4
 
-title: 'Boat Loader Is Not Installed On Removeable Media'
+title: 'Boot Loader Is Not Installed On Removeable Media'
```
jan-cerny: should be removable

vojtapolasek: ACtually maybe not necessarily:
https://www.collinsdictionary.com/dictionary/english/removeable

jan-cerny:
@vojtapolasek Is a remediation planned to be created? If there will be no remediation, please add `# remediation = none` to the test scenarios at the beginning.

Otherwise it's great.

vojtapolasek:
I don't think that remediation is planed, we don't know the user's machine setup and there can be many variants.

PR #5183, opened 2020-02-10
Permit enforcement of nosuid on /var

jcpunk:
#### Description:

The use of SUID binaries from `/var` strikes me as terribly odd.

#### Rationale:

My super quick scan of Fedora 31 shows no SUID binaries in `/var` so adding a rule to permit marking this partition as `nosuid` should be fairly safe while adding additional safeguards should one of the various daemons writing to `/var` attempt something tricky.


jan-cerny:
Thank you.

PR #5184, opened 2020-02-10
Permit setting sshd GSSAPI to yes

jcpunk:
#### Description:

Set GSSAPIAuth to enabled for SSHD

#### Rationale:

Sites using Kerberos/GSSAPI would benefit from having a rule to ensure GSSAPI is enabled.

### DIFF
--- a/linux_os/guide/services/ssh/ssh_server/sshd_enable_gssapi_auth/rule.yml
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_enable_gssapi_auth/rule.yml
@@ -1,0 +1,38 @@
+documentation_complete: true
+
+title: 'Enable GSSAPI Authentication'
+
+description: |-
+    Sites setup to use Kerberos or other GSSAPI Authenticaion require setting
+    sshd to accept this authentication.
+    To enable GSSAPI authentication, add or correct the following line in the
+    <tt>/etc/ssh/sshd_config</tt> file:
+    <pre>GSSAPIAuthentication yes</pre>
+
+rationale: |-
+    Kerberos authentication for SSH is often implemented using GSSAPI. If
+    Kerberos is enabled through SSH, the SSH daemon provides a means of access
+    to the system's Kerberos implementation. Vulnerabilities in the system's
+    Kerberos implementations may be subject to exploitation.
+
+    For enterprises, Kerberos is often enabled and used with GSSAPI for 
+    centralized user account management which may necessitate enabling of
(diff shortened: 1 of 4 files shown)


linux_os/guide/services/ssh/ssh_server/sshd_enable_gssapi_auth/tests/correct_value.pass.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+#
+# profiles = xccdf_org.ssgproject.content_profile_ospp
```
jan-cerny: Please remove `# profiles` line from all 3 tests. This rule isn't a part of any profile. And the test suite won't run the test scenarios unless the user explicitly overrides this setting.

jcpunk: Can do, I'll have a fix up shortly.

linux_os/guide/services/ssh/ssh_server/sshd_enable_gssapi_auth/rule.yml
```diff
@@ -0,0 +1,32 @@
...
+description: |-
+    Sites setup to use Kerberos or other GSSAPI Authenticaion require setting
+    sshd to accept this authentication.
+    To enable GSSAPI authentication, add or correct the following line in the
+    <tt>/etc/ssh/sshd_config</tt> file:
+    <pre>GSSAPIAuthentication yes</pre>
+
+rationale: |-
+    GSSAPI authentication is used to provide additional authentication mechanisms to
+    applications.
```
redhatrises: Please update the rationale to be something similar:

```
    Kerberos authentication for SSH is often implemented using GSSAPI. If Kerberos
    is enabled through SSH, the SSH daemon provides a means of access to the
    system's Kerberos implementation. Vulnerabilities in the system's Kerberos
    implementations may be subject to exploitation.

    For enterprises, Kerberos is often enabled and used with GSSAPI for 
    centralized user account management which may necessitate enabling of
    GSSAPI functionality in SSH. 
```

jcpunk: I'll switch this out and force push shortly.

jan-cerny:
Thank you for this patch and thanks for including the test scenarios :+1:

PR #5185, opened 2020-02-10
Add check for read-write SNMP users

jcpunk:
#### Description:

Add a test to look for Read Write SNMP users

#### Rationale:

Read Write SNMP users may present a risk to the system as this permits remote users to alter the system.


linux_os/guide/services/snmp/snmp_configure_server/snmpd_no_rwusers/rule.yml
```diff
@@ -0,0 +1,24 @@
...
+
+title: 'Ensure SNMP Read Write is disabled'
+
+description: |-
+    Edit <tt>/etc/snmp/snmpd.conf</tt>, remove any <tt>rwuser</tt> entries.
+    Once the read write users have been removed, restart the SNMP service:
+    <pre>$ sudo service snmpd restart</pre>
+
+rationale: |-
+    Certian SNMP settings can permit users to execute system behaviors from user
```
jan-cerny:
```suggestion
    Certain SNMP settings can permit users to execute system behaviors from user
```

linux_os/guide/services/snmp/snmp_configure_server/snmpd_no_rwusers/rule.yml
```diff
@@ -0,0 +1,24 @@
...
+    writes to the community strings.
+    This may permit a compromised accout to execute commands on a remote system.
```
jan-cerny:
```suggestion
    This may permit a compromised account to execute commands on a remote system.
```

linux_os/guide/services/snmp/snmp_configure_server/snmpd_no_rwusers/bash/shared.sh
```diff
@@ -0,0 +1,5 @@
+# platform = multi_platform_wrlinux,Red Hat Enterprise Linux 7,Red Hat Enterprise Linux 8
```
jan-cerny: You proposed to enable the rule also on Debian 9, Debian 10, RHEL 6 and Fedora, but the Bash script is more restricted. Is the SNMP configuration different on these distributions or is this an omission?

jcpunk: For the moment I'll restrict this to RH like systems as Debian can put things into a number of locations.  My initial plan was to be more aware, but during devel I hit a snag with Debian... forgot to drop them...

linux_os/guide/services/snmp/snmp_configure_server/snmpd_no_rwusers/rule.yml
```diff
@@ -0,0 +1,24 @@
...
+    Once the read write users have been removed, restart the SNMP service:
+    <pre>$ sudo service snmpd restart</pre>
+
+rationale: |-
+    Certain SNMP settings can permit users to execute system behaviors from user
+    writes to the community strings.
+    This may permit a compromised account to execute commands on a remote system.
+
+severity: medium
+
```
shawndwells: Could CCEs also be added?

jcpunk: Done.

jan-cerny:
@openscap-ci test this please

linux_os/guide/services/snmp/snmp_configure_server/snmpd_no_rwusers/rule.yml
```diff
@@ -0,0 +1,30 @@
...
+identifiers:
+    cce@rhel6: 82731-1
+    cce@rhel7: 82732-9
+    cce@rhel8: 82733-7
+    cce@fedora: 82734-5
```
jan-cerny: It's great that you added the CCEs, but we don't assign CCEs for Fedora.

jcpunk: I'll force push with a fix.

jan-cerny:
Thank you!

PR #5187, opened 2020-02-10
Remove krb5-server

jcpunk:
#### Description:

The krb5-server packages are not required in most instances, but can be problematic in unexpected places.

#### Rationale:

Similar to "remove the ldap servers if you don't need them", the kerberos-server packages can provide confusion and possibly setup rogue authentication.


jan-cerny:
Thank you!

PR #5188, opened 2020-02-10
Add RADIUS group and rule to remove server

jcpunk:
#### Description:

RADIUS is another network authentication protocol.

#### Rationale:

Similar to the "remove LDAP Server" and "Remove Kerberos Server" PRs I've submitted, this has the potential to be a confusing service to find on a system.

RADIUS - unlike other auth systems - is UDP based and uses a non-privileged port.  The potential for a compromised non-root account to abuse the binaries to confuse the network seems real to me.


linux_os/guide/services/radius/group.yml
```diff
@@ -0,0 +1,11 @@
...
+
+title: 'Radius'
+
+description: |-
+    Remote Authentication Dial-In User Service (RADIUS) is a networking
+    protocol, operating on port 1812 that provides centralized
+    Authentication, Authorization, and Accounting (AAA or Triple A)
+    management for users who connect and use a network service. 
+
+platform: machine
```
jan-cerny: Can a Radius server run in a container?

jcpunk: Good call, it can run in a container!

How would I note that here?  Just drop `platform`?

redhatrises: @jcpunk yeah... just drop platform.

linux_os/guide/services/radius/group.yml
```diff
@@ -0,0 +1,11 @@
+documentation_complete: true
+
+title: 'Radius'
```
shawndwells:
```suggestion
title: 'Remote Authentication Dial-In User Service (RADIUS)'
```

shawndwells: Small nit to expand group title, since this will likely be the first occurrence of the term.

linux_os/guide/services/radius/package_freeradius_removed/rule.yml
```diff
@@ -0,0 +1,31 @@
...
+    clients which use RADIUS for authentication. If the system is not
+    intended for use as a RADIUS Server it should be removed.
+
+rationale: |-
+    Unnecessary packages should not be installed to decrease the attack
+    surface of the system.  While this software is clearly essential on a
+    RADIUS server, it is not necessary on typical desktop or workstation systems.
+
+severity: low
+
```
shawndwells: Could you also assign some RHEL CCEs?

Format would be similar to:
`````
identifiers:
    cce@rhel6: xxxx
    cce@rhel7: yyyy
    cce@rhel8: zzzz
    cce@ocp4: aaaa
`````
Where the actual values can be taken from https://github.com/ComplianceAsCode/content/blob/master/shared/references/cce-redhat-avail.txt

Just remember to remove the CCE- bit, eg ``cce@rhel6: 82725-3``

shawndwells: (and remove and used values from the cce-redhat-avail file)

linux_os/guide/services/radius/group.yml
jcpunk: can do.

linux_os/guide/services/radius/group.yml
jcpunk: Will do.

linux_os/guide/services/radius/package_freeradius_removed/rule.yml
jcpunk: Will push a new version shortly.

jan-cerny:
@openscap-ci test this please

PR #5190, opened 2020-02-11
network_host_and_router_parameters group as machine-only

mildas:
#### Description:
Add machine-only platform to `network_host_and_router_parameters` `group.yml`  and remove platforms from all rules in the group (it does not need to be defined in both places).

#### Rationale:
All rules in `network_host_and_router_parameters` group should be machine-only.


jan-cerny:
Sounds reasonable because all of them are sysctl.

Issue #2811, opened 2018-04-30
Refactor Bash Remedation Functions


redhatrises:
1. Refactor Bash Remedation Function code to be less error prone e.g. #2794 
2. Refactor Bash Remedation Function code to use jinja2 templating
3. Refactor Bash Remedation Function code to not need include functions like #2810

matejak:
I have verified that it is possible to use macro substitutions in shared remediation functions on up-to-date `master`.

matejak:
Postponing to 0.1.41, as the fix is not trivial.

PR #5199, opened 2020-02-15
Fix hardcoded product in remediation script comments

iokomin:
#### Description:

Replaced hardcoded product from shared remediation script comments:
- rules audit_rules_unsuccessful_file_modification_*
- bash_remediation_function create_audit_remediation_unsuccessful_file_modification_detailed.sh
- template template_BASH_grub2_bootloader_argument

Just let me know if suggested change is not applicable for RHEL to wrap original ones into conditional statements.

#### Rationale:

Some shared scripts contain hardcoded product reference in comments not applicable to other supported products.

#### Testing:

- Checked successful ol7, ol8, rhel7, rhel8 builds and generated content

### DIFF
--- a/shared/bash_remediation_functions/perform_audit_rules_privileged_commands_remediation.sh
+++ b/shared/bash_remediation_functions/perform_audit_rules_privileged_commands_remediation.sh
@@ -6,7 +6,6 @@
 # 			One of 'auditctl' or 'augenrules'
 #
 # min_auid		Minimum original ID the user logged in with
-# 			'500' for RHEL-6 and before, '1000' for RHEL-7 and after.
 #
 # Example Call(s):
 #
--- a/linux_os/guide/system/auditing/auditd_configure_rules/audit_file_modification/audit_rules_unsuccessful_file_modification_creat/ansible/shared.yml
+++ b/linux_os/guide/system/auditing/auditd_configure_rules/audit_file_modification/audit_rules_unsuccessful_file_modification_creat/ansible/shared.yml
@@ -9,7 +9,7 @@
     path: /etc/audit/rules.d/30-ospp-v42-remediation.rules
     create: yes
     block: |
-        ## This content is a section of an Audit config snapshot recommended for RHEL8 sytems that target OSPP compliance.
+        ## This content is a section of an Audit config snapshot recommended for {{{ full_name }}} systems that target OSPP compliance.
         ## The following content has been retreived on 2019-03-11 from: https://github.com/linux-audit/audit-userspace/blob/master/rules/30-ospp-v42.rules
 
         ## The purpose of these rules is to meet the requirements for Operating
--- a/linux_os/guide/system/auditing/auditd_configure_rules/audit_file_modification/audit_rules_unsuccessful_file_modification_open_by_handle_at_o_creat/ansible/shared.yml
+++ b/linux_os/guide/system/auditing/auditd_configure_rules/audit_file_modification/audit_rules_unsuccessful_file_modification_open_by_handle_at_o_creat/ansible/shared.yml
@@ -9,7 +9,7 @@
     path: /etc/audit/rules.d/30-ospp-v42-remediation.rules
     create: yes
     block: |
-        ## This content is a section of an Audit config snapshot recommended for RHEL8 sytems that target OSPP compliance.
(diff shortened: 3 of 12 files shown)


shared/bash_remediation_functions/perform_audit_rules_privileged_commands_remediation.sh
```diff
@@ -5,8 +5,7 @@
 # audit_tool		tool used to load audit rules
 # 			One of 'auditctl' or 'augenrules'
 #
-# min_auid		Minimum original ID the user logged in with
-# 			'500' for RHEL-6 and before, '1000' for RHEL-7 and after.
```
jan-cerny: Why was this removed? I think that value of 500 for RHEL6 is still valid, and its still used in linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands/bash/rhel6.sh.

matejak: We have the `uid_min` available along e.g. `full_name` - check the RHEL6 `product.yml`. The remediation should probably reflect it.

iokomin: @jan-cerny suggested to remove according to the following reasoning:
- perform_audit_rules_privileged_commands_remediation function is shared between product types, while min_auid description comment references to the RHEL versions only, as a result generated scripts for other applicable products would contain RHEL-* values binding
- value for {{{ auid }}} specified in product global variables (or by default set to 1000) and clearly available in generated rules description supposed to use this function as part of remediation scripts
- with this comment limitation for product's {{{ uid_min }}} value is implicit while in general the function itself might handle any value given as min_auid argument

All in all, if it is considered reasonable to mention product min_auid value in comments I would propose following update:
"min_auid		Minimum original ID the user logged in with {{{ auid }}} for {{{ full_name }}}." 
Will this update work for you?

yuumasato: @iokomin Do the `{{{ auid }}}` get to the final text of the function rendered? I'd rather not have them un-rendered in the DS.
If not, the build system will need to be tweaked to render Jinja templates in `shared/bash_functions`...

matejak: For reference, `auid` is basically the same as `uid_min` (see `ssg/yaml.py:89`). I think that the first line of the comment is valid, because it describes what `min_auid` is. The second line that talks about RHEL-specific values should be taken out, or substituted by what @iokomin suggests.
OTOH, the remediation is defined in two files - the `shared.sh` and `rhel6.sh`, and the RHEL6 special case can be discarded if the shared gets the `min_auid` argument in form of the `{{{ auid }}}` macro that works for every product. It would be nice to make this addition to the pull request.

iokomin: @yuumasato Thanks for reminder, my suggested comment change definitely won't work as jinja2 templates in Bash Remediation Function still not available due to #2811. (Suggested text was just an idea which came into my mind, I would have verified if solution works in the project before pushing update to this PR :)) 
@matejak I agree that {{{ auid }}} usage in one shared remediation script looks reasonable, noticed this as well. I'll update PR with this change.

iokomin: @yuumasato, @matejak please find the PR updated:
- audit_rules_privileged_commands/bash/shared.sh used for all platforms with {{{ auid }}} template, rhel6.sh removed
- bash_remediation_functions/perform_audit_rules_privileged_commands_remediation.sh slightly updated to leave only first line for min_auid argument description

- checked ol{7,8},rhel{6,7,8},fedora,wrlinux builds w.r.t. to generated audit_rules_privileged_commands remediation content: rhel6 uses '500', other platforms use '1000' as min_auid value in bash scripts.

matejak: Great, I like that, and I am in favor of merging this PR. @yuumasato do you have any additional comments?

yuumasato: No comments, LGTM.

PR #5200, opened 2020-02-17
merge Firewalld rules into one group

vojtapolasek:
#### Description:

There were two subgroups in the network-firewalld group; firewalld_activation and ruleset_modifications. They are now merged into one group. Description of the network-firewalld group was a bit extended by merging in parts of former subgroups.

#### Rationale:

I don't think there is need for separate groups. Also this setup prevents the case where remediations are executed in wrong order.


redhatrises:
Something to consider is that this is considered equivalent to API breakage, and the many customers who use the content now have to evaluate, analyze, and document why this was done in their environment. We lose valuable information and organization in the groups being removed. I do agree that we should probably rework all the groups and do the breakage in a single release. There are a couple of things to be aware of that we can do that might be a better solution:

1. We are proposing in the latest release of the XCCDF spec to have profile based rule order execution which should help to solve some of these issues.
2. XCCDF has the capability to specify a requires for rule/group dependency. I don't know if we have implemented this functionality in OpenSCAP (not seeing an existing ticket), but we can for sure implement it in our build and add it to the content which IMO is better than removing valuable grouping and breaking custom tools and such way downstream from us.
3. Matej and Watson have done some rule ordering in the python code itself that may or may not help.

vojtapolasek:
Thank you very much for the feedback. I will try to use features Matej and Yuuma implemented and keep groups as they are so far.

PR #5204, opened 2020-02-20
Update rpm_verification group rules with OL support

iokomin:
#### Description

Added OL support to rpm_verify_ownership rule.
As rhbz1277603 is RHEL6 specific make gdm warning conditional in rpm_verify_ownership and rpm_verify_permissions rules.

#### Rationale:

- Set of shared packages and services rules is applicable to Oracle Linux releases.

#### Testing:
- Checked successful ol7, ol8, rhel6 builds and generated rules content
- Checked rules to work on ol7 and ol8 systems


yuumasato:
LGTM.
Thanks for the restriction of the warning to rhel6, it makes sense.

PR #5209, opened 2020-02-27
Introduce resolved profiles, and test for profile stability

matejak:
This PR introduces functionality that resolves relationship between profiles (selection, exclusion, extension) early in the build process, and saves resolved/compiled profiles to `build/<product>/profiles/` directory. Then, the build process works with resolved profiles only.
The XSLT code that resolves profiles has been leaved untouched for the time being.

This modification allows for simple test whether a profile selections have changed or not - reordering of entries in `.profile` files may hide an accidental selection falling out, and profiles that are extended by other profiles may be unexpectedly changed as a side-effect of their dependencies being changed. The introduced test watches for changes in RHEL8 `ospp`, `stig` and `pci-dss` profiles and the `e8` profile, other profiles/products can be added just by copying resolved profiles from the build folder to appropriate directories.


redhatrises:
Trying to understand why are we testing if profiles change? This seems like a waste of testing resources. Profile changes are expected.

matejak:
> Trying to understand why are we testing if profiles change? This seems like a waste of testing resources. Profile changes are expected.

Not all profile changes are expected - for example, we recently had a tough time with the RHEL8 STIG profile, when one party was reviewing something, while the profile was changing due to its dependency on the OSPP profile. On a related note, you don't expect selections to change in PRs that just regroup selections in a profile file, but it is difficult to review whether it is really the case (and we had issues with this during development of the OSPP profile).

In general, test of profile stability is extremely fast, and I think that only critical profiles with stakeholders should be protected by the test. That way, respective stakeholders can review easily what went on with "their" profile.

build-scripts/compile_profiles.py
```diff
@@ -0,0 +1,101 @@
...
+import ssg.build_yaml
+
+
+class ResolvableProfile(ssg.build_yaml.Profile):
+    def __init__(self, * args, ** kwargs):
+        super(ResolvableProfile, self).__init__(* args, ** kwargs)
+        self.raw_selections = set()
+        self.resolved = False
+
+    def resolve(self, all_profiles):
```
yuumasato: This should take `refine_rules` into account as well.

tests/test_profile_stability.py
```diff
@@ -0,0 +1,137 @@
...
+    parser.add_argument("test_data_root")
+    args = parser.parse_args()
+
+    reference_files = get_references(args.test_data_root)
+    if not reference_files:
+        raise RuntimeError("Unable to find any reference profiles in {test_root}"
+                           .format(test_root=args.test_data_root))
+    no_changes = True
+    for ref in reference_files:
+        sample = get_matching_sample(args.build_root, ref)
```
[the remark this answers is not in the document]
yuumasato: Or `compiled_profile`.

build-scripts/compile_profiles.py
```diff
@@ -0,0 +1,101 @@
...
+        if self.resolved:
+            return self.raw_selections
+        if self.extends:
+            extended_profile = all_profiles[self.extends]
+            extended_selects = extended_profile.resolve(all_profiles)
+
+            self.raw_selections.update(extended_selects)
+            self.variables.update(extended_profile.variables)
```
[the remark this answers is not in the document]
matejak: Right, good catch!

build-scripts/compile_profiles.py
matejak: Good point, what exactly is this refining? The `refine_rules` dictionary is available at that time, but I don't know what it does. How would this refinement affect profile compilation?
Refinement is effective when a rule is saved as an XML, i.e. later in the process.

tests/test_profile_stability.py
```diff
@@ -0,0 +1,137 @@
...
+        if not sample:
+            msg = ("Unexpectedly unable to find compiled profile corresponding"
+                   "to the test file {ref}. "
+                   "This indicates that a profile we have tests for hasn't been built."
+                   .format(ref=ref))
+            raise RuntimeError(msg)
+        changed = compare_reference_and_sample_selections(ref, sample)
+        no_changes = no_changes and changed
```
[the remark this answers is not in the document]
matejak: I could break when there is a change, but if I track the state continuously, I am able to produce a an error message that contains everything. Will try to do something about the naming.

tests/test_profile_stability.py
```diff
@@ -0,0 +1,137 @@
...
+def get_profile_name_from_reference_filename(fname):
+    path_components = fname.split(os.path.sep)
+    product_id = path_components[-2]
+    profile_id = os.path.splitext(path_components[-1])[0]
+    name = "{product_id}'s {profile_id}".format(
+        product_id=product_id, profile_id=profile_id)
+    return name
+
+
+def compare_reference_and_sample_selections(reference_fname, sample_fname):
```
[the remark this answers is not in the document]
matejak: OK, I originally thought that this function is abstract, but it is actually quite specialized already.

cmake/SSGCommon.cmake
```diff
@@ -96,16 +96,26 @@ macro(ssg_build_shorthand_xml PRODUCT)
     )
     string(REPLACE "\n" ";" SHORTHAND_INPUTS "${SHORTHAND_INPUTS_STR}")
 
+    add_custom_command(
+        # The command also produces the directory with rules, but this is done before the the shorthand XML.
+        OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/profiles"
+        COMMAND ${CMAKE_COMMAND} -E make_directory "${CMAKE_CURRENT_BINARY_DIR}/profiles"
+        COMMAND env "PYTHONPATH=$ENV{PYTHONPATH}" "${PYTHON_EXECUTABLE}" "${SSG_BUILD_SCRIPTS}/compile_profiles.py" --build-config-yaml "${CMAKE_BINARY_DIR}/build_config.yml" --product-yaml "${CMAKE_CURRENT_SOURCE_DIR}/product.yml" -o "${CMAKE_CURRENT_BINARY_DIR}/profiles/{name}.profile" "${CMAKE_CURRENT_SOURCE_DIR}/profiles/*.profile"
```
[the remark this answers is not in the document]
matejak: No, you are right.

cmake/SSGCommon.cmake
```diff
@@ -96,16 +96,26 @@ macro(ssg_build_shorthand_xml PRODUCT)
     )
     string(REPLACE "\n" ";" SHORTHAND_INPUTS "${SHORTHAND_INPUTS_STR}")
 
+    add_custom_command(
+        # The command also produces the directory with rules, but this is done before the the shorthand XML.
```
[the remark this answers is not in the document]
matejak: just a copy-paste error.

build-scripts/compile_profiles.py
```diff
@@ -0,0 +1,101 @@
...
+    return profile_files
+
+
+def main():
+    parser = create_parser()
+    args = parser.parse_args()
+    env_yaml = get_env_yaml(args.build_config_yaml, args.product_yaml)
+
+    profile_files = get_profile_files_from_root(env_yaml)
+    profile_files.extend(args.profile_file)
```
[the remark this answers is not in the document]
matejak: Think of it as of a convenience feature - you can generate a compiled profile by just supplying a list of profile files.

build-scripts/compile_profiles.py
yuumasato: > Good point, what exactly is this refining? The `refine_rules` dictionary is available at that time, but I don't know what it does. How would this refinement affect profile compilation?

I believe it would behave the same variables.
If profile B extends profile A, and profile B overrides a rule refinement from Profile A, the refinement from Profile B prevails.

For example, `ospp` profile refines the severity of two rules and set them as informational. A profile extending `ospp` could re-refine the rule and say they are not informational.

> Refinement is effective when a rule is saved as an XML, i.e. later in the process.

This applies for variables too, right?

matejak: I see, so we basically need to do the same as with variables, while making sure that the "selected" yaml key gets the list that contains proper refinements.

build-scripts/compile_profiles.py
```diff
@@ -15,17 +16,30 @@ def __init__(self, * args, ** kwargs):
...
-            extended_selects = extended_profile.resolve(all_profiles)
+            extended_profile.resolve(all_profiles)
 
+            extended_selects = extended_profile.raw_selections
             self.raw_selections.update(extended_selects)
-            self.variables.update(extended_profile.variables)
+
+            updated_variables = dict(extended_profile.variables)
+            updated_variables.update(self.variables)
+            self.variables = updated_variables
```
yuumasato: I'm thinking right now that we probably shouldn't be messing with the original variables from `Profile`.
This should be `self.raw_variables = updated_variables`.

And how about changing `raw` to `resolved`?

build-scripts/compile_profiles.py
```diff
@@ -15,17 +16,30 @@ def __init__(self, * args, ** kwargs):
...
+
         self.raw_selections.update(set(self.selected))
```
yuumasato: To keep related lines close, move line above after `self.raw_selections.update(extended_selects)`.

ssg/build_yaml.py
```diff
@@ -144,6 +144,10 @@ def dump_yaml(self, file_name, documentation_complete=True):
             selections.append("!"+item)
         for varname in self.variables.keys():
             selections.append(varname+"="+self.variables.get(varname))
+        for rule, refinements in self.refine_rules.items():
+            for prop, val in refinements:
+                selections.append("{rule}.{property}={value}"
+                                  .format(rule=rule, property=prop, value=val))
```
yuumasato: Nice catch.

tests/test_profile_stability.py
```diff
@@ -0,0 +1,145 @@
...
+        difference = get_reference_vs_built_difference(ref, compiled_profile)
+        if not difference.empty:
+            no_changes = False
+            comprehensive_profile_name = get_profile_name_from_reference_filename(ref)
+            report_comparison(comprehensive_profile_name, difference)
+
+    if not no_changes:
+        msg = (
+            "If changes to mentioned profiles are intentional, "
+            "edit the appropriate files in '{test_root}' directory."
```
yuumasato: It may be worthy to print out a `cp` command to make it easy for the user to align the profile with the reference file.

build-scripts/compile_profiles.py
matejak: When I look at it again, another thought appears - why to bother with `raw_selections` - the `resolve` method could update selections, so they reflect the extended profile, and they take into account unselections and discard them in the process.
In another words, why not to mess with original attributes, when messing means they are just unified with the extended profile and unselections are applied?

jan-cerny:
@matejak How do you prevent people from unexpected changes of test normatives?

build-scripts/compile_profiles.py
yuumasato: You have a point, it should be fine to update the original attributes.

matejak:
> @matejak How do you prevent people from unexpected changes of test normatives?

One can't prevent it, but as normatives tend to be resolved and alphabeticaly sorted, changes will be easy to spot and to discuss in PRs. It is even possible to [configure Github](https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners) to force explicit approvals to PRs if files are touched.

build-scripts/compile_profiles.py
matejak: I have updated the code to use the existing attributes-only approach.

yuumasato: Looks good

yuumasato:
I have an usability improvement suggestion. Idon't consider this blocking this PR, but I'll ask if you consider this useful and are willing to do it in this PR.
I have edited only OSPP profile, but it shows that STIG profile also changed.
```
Following selections were removed from the rhel8's ospp profile:
 - grub2_vsyscall_argument.role=unscored
 - grub2_vsyscall_argument.severity=info
Following selections were removed from the rhel8's stig profile:
 - grub2_vsyscall_argument.role=unscored
 - grub2_vsyscall_argument.severity=info
```

How about adding message that STIG extends OSPP profile.

matejak:
> How about adding message that STIG extends OSPP profile.

I think that this not trivial to do really right - the test would have to do more than just compare two yamls - it would have to read and resolve the original profile files in order to be able to generate a helpful message.
I have added a general reminder, and I propose to leave this idea to anybody else who finds it compelling. One option would be to let compiled profiles save the information about their chain of inheritance.

yuumasato:
> I think that this not trivial to do really right - the test would have to do more than just compare two yamls - it would have to read and resolve the original profile files in order to be able to generate a helpful message.

Oh right, once it is resolved it doesn't keep track of who it extends anymore.

yuumasato:
> I have added a general reminder, and I propose to leave this idea to anybody else who finds it compelling. One option would be to let compiled profiles save the information about their chain of inheritance.

Thanks for also adding instructions how to keep the extended profile the same.

PR #5217, opened 2020-03-06
SSGTS: profile mode extended to reboot VM before performing the final scan

matusmarhefka:
#### Description:

Some rules (for example `configure_crypto_policy`) require reboot of a machine
after a remediation to fully pass. Therefore, this commit adds reboot functionality
into the libvirt backend of the SSGTS. Reboot is performed only when testing using
the SSGTS profile mode after the remediation of a profile and before the final scan.
    
Other SSGTS updates:
* `tests/install_vm.py`: updated url to point to Fedora 31
* Addressed various pep8 code issues.

### DIFF
--- a/tests/ssg_test_suite/virt.py
+++ b/tests/ssg_test_suite/virt.py
@@ -174,8 +174,9 @@ def reboot_domain(domain, domain_ip, ssh_port):
     while domain.isActive():
         time.sleep(1)
         if time.perf_counter() - start_time >= timeout:
-            raise TimeoutError("Timeout reached: '{0}' domain failed to shutdown."
-                               .format(domain.name()))
+            str_err = "Timeout reached: '{0}' domain failed to shutdown.".format(domain.name())
+            logging.debug(str_err)
+            raise TimeoutError(str_err)
 
     logging.debug("Starting domain '{0}'".format(domain.name()))
     domain.create()
@@ -190,5 +191,7 @@ def reboot_domain(domain, domain_ip, ssh_port):
         except OSError as exc:
             time.sleep(1)
             if time.perf_counter() - start_time >= timeout:
-                raise TimeoutError("Timeout reached: '{0}' ({1}:22) domain does not "
-                                   "accept connections.".format(domain.name(), domain_ip)) from exc
+                str_err = ("Timeout reached: '{0}' ({1}:{2}) domain does not "
+                           "accept connections.".format(domain.name(), domain_ip, ssh_port))
+                logging.debug(str_err)
+                raise TimeoutError(str_err) from exc
--- a/tests/ssg_test_suite/virt.py
+++ b/tests/ssg_test_suite/virt.py
@@ -4,6 +4,7 @@ from __future__ import print_function
 import libvirt
 import logging
 import time
+from socket import create_connection
 
 import xml.etree.ElementTree as ET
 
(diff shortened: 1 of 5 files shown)


yuumasato:
@matusmarhefka how did you test this feature?
I tried running RHEL8 STIG Profile but I have problems login in after the reboot, probably because SSGTS tries to login as root.

And during one test I triggered the test and I asked me for the password, I typed it in, and it immediately returned me this:
```
INFO - Evaluation of profile xccdf_org.ssgproject.content_profile_stig.
INFO - Rebooting domain 'rhel8.2-night' before final scan.
root@192.168.122.125's password: 
ERROR - Profile run should end with return code 0 or 2 not "1" as it did!
```
Very likely the reboot timeout had already expired.

matusmarhefka:
@yuumasato Its profile mode, so there are rules which might break the ssh access to the VM as root. Have you used the `ds_unselect_rules.sh` script with the list from the `unselect_rules_list` to unselect these kind of rules from the datastream you are using for testing?

tests/ssg_test_suite/virt.py
```diff
@@ -158,3 +159,36 @@ def start_domain(domain):
...
+    logging.debug("Waiting for domain to boot")
+    start_time = time.perf_counter()
+    while True:
+        try:
+            with create_connection((domain_ip, ssh_port), timeout=connection_timeout):
+                break
+        except OSError as exc:
+            time.sleep(1)
+            if time.perf_counter() - start_time >= timeout:
+                raise TimeoutError("Timeout reached: '{0}' ({1}:22) domain does not "
```
yuumasato: How about also logging the Timeout message?

tests/ssg_test_suite/virt.py
```diff
@@ -158,3 +159,36 @@ def start_domain(domain):
...
+    domain.shutdown()
+
+    # Wait until domain shuts down.
+    logging.debug("Waiting for domain to shutdown")
+    start_time = time.perf_counter()
+    while domain.isActive():
+        time.sleep(1)
+        if time.perf_counter() - start_time >= timeout:
+            raise TimeoutError("Timeout reached: '{0}' domain failed to shutdown."
+                               .format(domain.name()))
```
yuumasato: I'm not a fan of the traceback it prints:
```
INFO - Rebooting domain 'rhel8.2-night' before final scan.
  File "test_suite.py", line 323, in <module>
    main()
  File "test_suite.py", line 319, in main
    options.func(options)
  File "/home/wsato/git/content/tests/ssg_test_suite/profile.py", line 43, in perform_profile_check
    checker.test_target(options.target)
  File "/home/wsato/git/content/tests/ssg_test_suite/oscap.py", line 621, in test_target
    self._test_target(target)
  File "/home/wsato/git/content/tests/ssg_test_suite/profile.py", line 22, in _test_target
    self.run_test_for_all_profiles(profiles)
  File "/home/wsato/git/content/tests/ssg_test_suite/oscap.py", line 635, in run_test_for_all_profiles
    self._run_test(profiles[0], test_data)
  File "/home/wsato/git/content/tests/ssg_test_suite/profile.py", line 33, in _run_test
    result = runner.run_stage(stage)
  File "/home/wsato/git/content/tests/ssg_test_suite/oscap.py", line 341, in run_stage
    result = self.final()
  File "/home/wsato/git/content/tests/ssg_test_suite/oscap.py", line 419, in final
    self.environment.reboot()
  File "/home/wsato/git/content/tests/ssg_test_suite/test_env.py", line 156, in reboot
    virt.reboot_domain(self.domain, self.domain_ip, self.ssh_port)
  File "/home/wsato/git/content/tests/ssg_test_suite/virt.py", line 178, in reboot_domain
    .format(domain.name()))
TimeoutError: Timeout reached: 'rhel8.2-night' domain failed to shutdown.
```

Someone should probably catch this exception.

matusmarhefka:
@yuumasato We already have it documented at https://github.com/ComplianceAsCode/content/blob/master/tests/README.md#unselecting-problematic-rules
Where do you think it would be appropriate to print this information about unselecting rules? Help output of the profile mode?

I agree with adding support for https://github.com/OpenSCAP/openscap/pull/1485 but not in this PR, we should probably also wait until it gets into distributions.

tests/ssg_test_suite/virt.py
matusmarhefka: Can you elaborate more on this? Do you want the traceback to be hidden and to have just the error message printed? I think we should terminate testing in this case anyway.

tests/ssg_test_suite/virt.py
matusmarhefka: Will do.

yuumasato:
> Where do you think it would be appropriate to print this information about unselecting rules? Help output of the profile mode?

Could be in help of profile mode.
A reminder when the Test Suite fails to connect after boot could work as well.

> I agree with adding support for [OpenSCAP/openscap#1485](https://github.com/OpenSCAP/openscap/pull/1485) but not in this PR, we should probably also wait until it gets into distributions.

Sure, didn't meant for support for non-root users to be added in this PR.

tests/ssg_test_suite/virt.py
yuumasato: > Do you want the traceback to be hidden and to have just the error message printed?

Yes, UX of traceback is not nice. 

> I think we should terminate testing in this case anyway.

Agreed.

matusmarhefka:
@yuumasato Profile mode help output extended and no more tracebacks on reboot timeouts, exceptions should be catched and test environment cleaned-up, example error message on timeout:

```
ERROR - Terminating due to timeout: Timeout reached: 'ssgts' (192.168.122.10:22) domain does not accept connections.
```

PR #5262, opened 2020-03-07
warning: failed to load external entity "shared/references/disa-os-srg-v1r5.xml"

isimluk:
Addressing build warning (non fatal failure):
```
    warning: failed to load external entity "shared/references/disa-os-srg-v1r5.xml"
```
    
This file has last changed in a6b377368b7157c1ceeb8b83c22913b2171c63b7.

Also, refactoring the same setting to the single place.


yuumasato:
LGTM, thanks!

PR #5266, opened 2020-03-09
fix OVAL template for audit_rules_unsuccessful_file_modification

vojtapolasek:
#### Description:

A negative lookahead was added to the regex in OVAL, therefore the cases where the rule contains for example -F a1&0111 are not matched.

#### Rationale:

The regex was too broad and could not distinguish between lines containing and not containing pattern described above.

### DIFF
--- a/shared/templates/template_OVAL_audit_rules_unsuccessful_file_modification
+++ b/shared/templates/template_OVAL_audit_rules_unsuccessful_file_modification
@@ -46,10 +46,10 @@
 
   <!-- General rule boiler plate -->
   <constant_variable id="var_32bit_arufm_{{{ NAME }}}_head" version="1" datatype="string" comment="audit rule arch and syscal">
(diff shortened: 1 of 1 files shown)


shared/templates/template_OVAL_audit_rules_unsuccessful_file_modification
```diff
@@ -46,10 +46,10 @@
 
   <!-- General rule boiler plate -->
   <constant_variable id="var_32bit_arufm_{{{ NAME }}}_head" version="1" datatype="string" comment="audit rule arch and syscal">
-      <value>^[\s]*-a[\s]+always,exit[\s]+(?:-F[\s]+arch=b32[\s]+)(?:.*(-S[\s]+{{{ NAME }}}[\s]+|([\s]+|[,]){{{ NAME }}}([\s]+|[,]))).*</value>
+      <value>^[\s]*-a[\s]+always,exit[\s]+(?:-F[\s]+arch=b32[\s]+)(?:.*(-S[\s]+{{{ NAME }}}[\s]+|([\s]+|[,]){{{ NAME }}}([\s]+|[,])))((?!-F[\s]+a[1234]&amp;).)*</value>
```
yuumasato: Instead of `[1234]` I suggest to use `\d`, it will be more robust.

yuumasato: And both groups should be non-capturing.

PR #5268, opened 2020-03-10
ocp4: Remove the rule that disables user namespaces

jhrozek:
#### Description:
Do not disable user namespaces via sysctl for RHCOS.

#### Rationale:
As the rule itself says, we should not disable the support for user
namespaces on container hosts such as RHCOS.

### DIFF
--- a/ocp4/profiles/moderate.profile
+++ b/ocp4/profiles/moderate.profile
@@ -143,9 +143,6 @@ selections:
     - sysctl_kernel_kexec_load_disabled
     - sysctl_kernel_yama_ptrace_scope
     - sysctl_kernel_perf_event_paranoid
-    - sysctl_user_max_user_namespaces
-    - sysctl_user_max_user_namespaces.role=unscored
-    - sysctl_user_max_user_namespaces.severity=info
     - sysctl_kernel_unprivileged_bpf_disabled
     - sysctl_net_core_bpf_jit_harden
 


redhatrises:
@jhrozek should we actually specify the values that are needed for `sysctl_user_max_user_namespaces`?

jhrozek:
On Tue, Mar 10, 2020 at 07:57:11AM -0700, Gabe Alford wrote:
> @jhrozek should we actually specify the values that are needed for `sysctl_user_max_user_namespaces`?

Do you mean the value for the actual sysctl parameter? Do you think we
should override the default?

redhatrises:
> On Tue, Mar 10, 2020 at 07:57:11AM -0700, Gabe Alford wrote: @jhrozek should we actually specify the values that are needed for `sysctl_user_max_user_namespaces`?
> Do you mean the value for the actual sysctl parameter? Do you think we should override the default?

@jhrozek was thinking if the RHCOS content should be opinionated in setting the expected `sysctl_user_max_user_namespaces` value for RHCOS rather than disabling it. Like the value should be greater than 0, "1500", or some other value. It may not make sense as `sysctl_user_max_user_namespaces` might need to be dynamic. Just curious if the policy should be opinionated about a value instead of disabling it.

jhrozek:
On Tue, Mar 10, 2020 at 01:22:40PM -0700, Gabe Alford wrote:
> > On Tue, Mar 10, 2020 at 07:57:11AM -0700, Gabe Alford wrote: @jhrozek should we actually specify the values that are needed for `sysctl_user_max_user_namespaces`?
> > Do you mean the value for the actual sysctl parameter? Do you think we should override the default?
> 
> @jhrozek was thinking if the RHCOS content should be opinionated in setting the expected `sysctl_user_max_user_namespaces` value for RHCOS rather than disabling it. Like the value should be greater than 0, "1500", or some other value. It may not make sense as `sysctl_user_max_user_namespaces` might need to be dynamic. Just curious if the policy should be opinionated about a value instead of disabling it.

Since I cannot think of any hardening or compliance reason to diverge
from the default, my opinion is that we should just stick with the
default.

jhrozek:
@JAORMX do you have an opinion on this PR?

JAORMX:
I think we should stick with the default. User namespaces are crucial for containers and the numbers will increase in the foreseeable future due to the move to rootless containers. So my take is : let's not mess with this so we don't break deployments.

redhatrises:
Sounds good. Thanks @jhrozek @JAORMX ! We can re-review later anyways if some security reason pops up. Ack

PR #5270, opened 2020-03-11
Added support for rawhide gpgkey check/remediation.

matejak:
Rawhide may be used for testing, and there is no reason to not support it, although we have all bits that are required already in place.


yuumasato:
LGTM, thanks.

PR #5271, opened 2020-03-11
Add Ignition remediation for sysctl parameters in the OCP4 moderate profile

jhrozek:
#### Description:
The patchset adds remediations for sysctl rules in the OCP4 moderate
profile. There is a patch per rule, which might look a bit odd, but I
thought it would provide for an easier review as is't then clearer which
file is created and even the content is quite readable. If you think this
many patches would pollute the git history too much, just let me know and
I can squash them.

#### Rationale:
We need to increase the content coverage of the OCP4 profile in
general. Adding the rules is also a good way of finding issues with the
tooling and UI/UX.


yuumasato:
The granularity of the commits is fine to me. 

These could be quickly covered by a sysctl template for ignition.
I assume that you'd rather have these merged sooner, than wait for the template to come.

jhrozek:
I would honestly prefer to have them merged, yes, that way we can start using the content in CI and verify that a cluster using these remediations would actually pass the OCP tests.

yuumasato:
Ack.

PR #5272, opened 2020-03-11
Fix the utils/ignition-remediation.py helper script

jhrozek:
#### Description:
The decode action would decode the urlencoded part of the ignition remediation,
but not print the result. Oops.

#### Rationale:
The script is handy to review ignition remediations. For example, to review
a rule from PR #5271, you could call:
```
python3 utils/ignition-remediation.py decode --rule=$rulename
```


yuumasato:
:)

PR #5274, opened 2020-03-11
Add dconf_db_up_to_date to RHEL8 STIG profile.

ggbecker:
#### Description:

- The rule dconf_db_up_to_date makes sure that the dconf database is up to date with regards to keyfiles. Rules that touch dconf settings only mention configuration under keyfiles, so if someone follows only the rules description guidance, value won't be reflected to the actual dconf database. So we need to make sure that the values are compiled to the binary database.


yuumasato:
LGTM.

PR #5276, opened 2020-03-11
Extracted a grub superuser username rule from the grub2_password rule

matejak:
Having an account protected by a password is a different thing both in terms of implementation and security:

- A well-chosen password makes the system secure, even if the username is known.
- In case of grub2 rules, username becomes knonw at the moment it appears in the kickstart, which would mean that such rule would be always failing on systems installed using a public kickstart.
- As of now, it is not even possible to change the username by other means than by performing a manual admin task.

Fixes: RHBZ#1810603


linux_os/guide/system/bootloader-grub2/grub2_admin_username/tests/grub-passwords.sh
```diff
@@ -0,0 +1,16 @@
...
+
+
+function set_superusers {
+	echo "set superusers=\"$1\"" > "$GRUB_CFG_ROOT/grub.cfg"
+}
+
+
+function remove_grub_password {
+	rm -f "$GRUB_CFG_ROOT/user.cfg"
+}
```
jan-cerny: Having a function that only renames rm command seems unnecessary. Moreover, its name is misleading.

linux_os/guide/system/bootloader-grub2/grub2_password/oval/shared.xml
```diff
@@ -31,12 +25,12 @@
...
 
-  <ind:textfilecontent54_test check="all" check_existence="all_exist" comment="superuser is defined in /boot/grub2/grub.cfg files. Superuser is not root, admin, or administrator" id="test_bootloader_superuser" version="1">
+  <ind:textfilecontent54_test check="all" check_existence="all_exist" comment="superuser is defined in /boot/grub2/grub.cfg files." id="test_bootloader_superuser" version="2">
     <ind:object object_ref="object_bootloader_superuser" />
   </ind:textfilecontent54_test>
-  <ind:textfilecontent54_object id="object_bootloader_superuser" version="1">
+  <ind:textfilecontent54_object id="object_bootloader_superuser" version="2">
     <ind:filepath>/boot/grub2/grub.cfg</ind:filepath>
-    <ind:pattern operation="pattern match">^[\s]*set[\s]+superusers=\"(?i)(?!root|admin|administrator)(?-i).*\"$</ind:pattern>
+    <ind:pattern operation="pattern match">^[\s]*set[\s]+superusers=\".*\"$</ind:pattern>
```
jan-cerny: the regex allows username `""` or invalid user names

linux_os/guide/system/bootloader-grub2/grub2_uefi_password/oval/shared.xml
```diff
@@ -25,12 +25,12 @@
...
 
-  <ind:textfilecontent54_test check="all" check_existence="all_exist" comment="superuser is defined in /boot/efi/EFI/redhat/grub.cfg. Superuser is not root, admin, or administrator" id="test_bootloader_uefi_superuser" version="1">
+  <ind:textfilecontent54_test check="all" check_existence="all_exist" comment="superuser is defined in /boot/efi/EFI/redhat/grub.cfg." id="test_bootloader_uefi_superuser" version="2">
     <ind:object object_ref="object_bootloader_uefi_superuser" />
   </ind:textfilecontent54_test>
-  <ind:textfilecontent54_object id="object_bootloader_uefi_superuser" version="1">
+  <ind:textfilecontent54_object id="object_bootloader_uefi_superuser" version="2">
     <ind:filepath operation="pattern match">^/boot/efi/EFI/(redhat|fedora)/grub.cfg$</ind:filepath>
-    <ind:pattern operation="pattern match">^[\s]*set[\s]+superusers=\"(?i)(?!root|admin|administrator)(?-i).*\"$</ind:pattern>
+    <ind:pattern operation="pattern match">^[\s]*set[\s]+superusers=\".*\"$</ind:pattern>
```
jan-cerny: ditto

linux_os/guide/system/bootloader-grub2/grub2_admin_username/oval/shared.xml
```diff
@@ -0,0 +1,23 @@
+<def-group>
+  <definition class="compliance" id="grub2_admin_username" version="1">
+    <metadata>
+      <title>Set Boot Loader Superuser Username to Unique Value</title>
+      {{{- oval_affected(products) }}}
+      <description>The grub2 boot loader superuser should have a username that is hard to guess.</description>
+    </metadata>
+
+    <criteria operator="OR">
+      <criterion comment="Pass if /boot/grub2/grub.cfg does not exist" test_ref="test_bootloader_grub_cfg" />
```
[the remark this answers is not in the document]
jan-cerny: I would also like to have a test scenario that checks if the rule passes when the file does not exist.

linux_os/guide/system/bootloader-grub2/grub2_admin_username/tests/unique-username.pass.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+. grub-passwords.sh
+
+set_superusers "koskic"
```
jan-cerny: What about testing invalid entry and no entry?

linux_os/guide/system/bootloader-grub2/grub2_password/oval/shared.xml
matejak: I think that this is fine - we are supposed to check whether the username doesn't violate a requirement, and it's grub's responsibility to handle nuances s.a. invalid or empty usernames.

matejak: Good point.

matejak:
Thanks for the feedback, I hope concerns were addressed.

linux_os/guide/system/bootloader-grub2/grub2_admin_username/rule.yml
```diff
@@ -0,0 +1,75 @@
...
+    admin, or administrator for the grub2 superuser account.
+    <br /><br />
+    Change the superuser to a different username (The default is 'root').
+    <pre>$ sed -i 's/\(set superuser=\).*/\1"&lt;unique user ID&gt;"/g' /etc/grub.d/01_users</pre>
+    <br /><br />
+    Once the superuser account has been added,
+    update the
+    <tt>grub.cfg</tt> file by running:
+    <pre>grub2-mkconfig -o /boot/grub2/grub.cfg</pre>
+    NOTE: Do NOT manually add the superuser account and password to the
```
redhatrises: Please make this a warning as well instead of a note.

linux_os/guide/system/bootloader-grub2/grub2_uefi_admin_username/rule.yml
```diff
@@ -0,0 +1,76 @@
...
+    admin, or administrator for the grub2 superuser account.
+    <br /><br />
+    Change the superuser to a different username (The default is 'root').
+    <pre>$ sed -i 's/\(set superuser=\).*/\1"&lt;unique user ID&gt;"/g' /etc/grub.d/01_users</pre>
+    <br /><br />
+    Once the superuser account has been added,
+    update the
+    <tt>grub.cfg</tt> file by running:
+    <pre>grub2-mkconfig -o /boot/efi/EFI/redhat/grub.cfg</pre>
+    NOTE: Do NOT manually add the superuser account and password to the
```
redhatrises: Please make this a warning as well instead of a note

linux_os/guide/system/bootloader-grub2/grub2_admin_username/rule.yml
```diff
@@ -0,0 +1,76 @@
...
+    nist-csf: PR.AC-1,PR.AC-4,PR.AC-6,PR.AC-7,PR.PT-3
+    ospp: FIA_AFL.1
+    srg: SRG-OS-000080-GPOS-00048
+    stigid@rhel7: "010480"
+    isa-62443-2013: 'SR 1.1,SR 1.10,SR 1.11,SR 1.12,SR 1.13,SR 1.2,SR 1.3,SR 1.4,SR 1.5,SR 1.6,SR 1.7,SR 1.8,SR 1.9,SR 2.1,SR 2.2,SR 2.3,SR 2.4,SR 2.5,SR 2.6,SR 2.7'
+    isa-62443-2009: 4.3.3.2.2,4.3.3.5.1,4.3.3.5.2,4.3.3.5.3,4.3.3.5.4,4.3.3.5.5,4.3.3.5.6,4.3.3.5.7,4.3.3.5.8,4.3.3.6.1,4.3.3.6.2,4.3.3.6.3,4.3.3.6.4,4.3.3.6.5,4.3.3.6.6,4.3.3.6.7,4.3.3.6.8,4.3.3.6.9,4.3.3.7.1,4.3.3.7.2,4.3.3.7.3,4.3.3.7.4
+    cobit5: DSS05.02,DSS05.04,DSS05.05,DSS05.07,DSS05.10,DSS06.03,DSS06.06,DSS06.10
+    iso27001-2013: A.18.1.4,A.6.1.2,A.7.1.1,A.9.1.2,A.9.2.1,A.9.2.2,A.9.2.3,A.9.2.4,A.9.2.6,A.9.3.1,A.9.4.1,A.9.4.2,A.9.4.3,A.9.4.4,A.9.4.5
+    cis-csc: 1,11,12,14,15,16,18,3,5
+    anssi: NT28(R17)
```
shawndwells: This is an irregular requirement. Have these policy mappings been reviewed? Where did they come from?

linux_os/guide/system/bootloader-grub2/grub2_password/rule.yml
```diff
@@ -8,28 +8,15 @@ description: |-
     The grub2 boot loader should have a superuser account and password
     protection enabled to protect boot-time settings.
     <br /><br />
-    To do so, select a superuser account name and password and and modify the
-    <tt>/etc/grub.d/01_users</tt> configuration file with the new account name.
-    <br /><br />
```
shawndwells: Why would this text be removed? Is it no longer accurate?

linux_os/guide/system/bootloader-grub2/grub2_password/rule.yml
```diff
@@ -8,28 +8,15 @@ description: |-
...
     <br /><br />
-    NOTE: It is recommended not to use common administrator account names like root,
-    admin, or administrator for the grub2 superuser account.
-    <br /><br />
-    Change the superuser to a different username (The default is 'root').
-    <pre>$ sed -i s/root/bootuser/g /etc/grub.d/01_users</pre>
-    <br /><br />
-    To meet FISMA Moderate, the bootloader superuser account and password MUST
-    differ from the root account and password.
-    Once the superuser account and password have been added,
```
shawndwells: Why would this text be removed? Is it no longer accurate?

linux_os/guide/system/bootloader-grub2/grub2_uefi_admin_username/rule.yml
```diff
@@ -0,0 +1,77 @@
...
+
+description: |-
+    The grub2 boot loader should have a superuser account and password
+    protection enabled to protect boot-time settings.
+    <br /><br />
+    To maximize the protection, select a password-protected superuser account with unique name, and modify the
+    <tt>/etc/grub.d/01_users</tt> configuration file to reflect the account name change.
+    <br /><br />
+    Do not to use common administrator account names like root,
+    admin, or administrator for the grub2 superuser account.
```
shawndwells:
```suggestion
    It is highly suggested not to use common administrator account names like root,
    admin, or administrator for the grub2 superuser account.
```

shawndwells: Unless we can point to something that explicitly forbids this, the language should be adjusted to reflect this is a suggestion.

linux_os/guide/system/bootloader-grub2/grub2_uefi_admin_username/rule.yml
```diff
@@ -0,0 +1,77 @@
...
+    nist-csf: PR.AC-4,PR.AC-6,PR.PT-3
+    ospp: FIA_AFL.1
+    srg: SRG-OS-000080-GPOS-00048
+    stigid@rhel7: "010490"
+    isa-62443-2013: 'SR 1.1,SR 1.10,SR 1.11,SR 1.12,SR 1.13,SR 1.2,SR 1.3,SR 1.4,SR 1.5,SR 1.6,SR 1.7,SR 1.8,SR 1.9,SR 2.1,SR 2.2,SR 2.3,SR 2.4,SR 2.5,SR 2.6,SR 2.7'
+    isa-62443-2009: 4.3.3.2.2,4.3.3.5.1,4.3.3.5.2,4.3.3.5.3,4.3.3.5.4,4.3.3.5.5,4.3.3.5.6,4.3.3.5.7,4.3.3.5.8,4.3.3.6.1,4.3.3.6.2,4.3.3.6.3,4.3.3.6.4,4.3.3.6.5,4.3.3.6.6,4.3.3.6.7,4.3.3.6.8,4.3.3.6.9,4.3.3.7.1,4.3.3.7.2,4.3.3.7.3,4.3.3.7.4
+    cobit5: DSS05.02,DSS05.04,DSS05.05,DSS05.07,DSS06.03,DSS06.06
+    iso27001-2013: A.6.1.2,A.7.1.1,A.9.1.2,A.9.2.1,A.9.2.3,A.9.4.1,A.9.4.4,A.9.4.5
+    cis-csc: 11,12,14,15,16,18,3,5
+    anssi: NT28(R17)
```
shawndwells: Where did these policy mappings come from?

linux_os/guide/system/bootloader-grub2/grub2_uefi_password/rule.yml
```diff
@@ -8,28 +8,15 @@ description: |-
...
     <br /><br />
-    NOTE: It is recommended not to use common administrator account names like root,
-    admin, or administrator for the grub2 superuser account.
-    <br /><br />
-    Change the superuser to a different username (The default is 'root').
-    <pre>$ sed -i s/root/bootuser/g /etc/grub.d/01_users</pre>
-    <br /><br />
-    To meet FISMA Moderate, the bootloader superuser account and password MUST
-    differ from the root account and password.
-    Once the superuser account and password have been added,
```
shawndwells: The removed text still seems valid. What has changed?

linux_os/guide/system/bootloader-grub2/grub2_password/rule.yml
matejak: The rule has been split, and this modification has been moved to the other rule.

linux_os/guide/system/bootloader-grub2/grub2_uefi_admin_username/rule.yml
matejak: They came from the original rule that this one was split from.

linux_os/guide/system/bootloader-grub2/grub2_uefi_password/rule.yml
matejak: Partly the description has been moved to the new rule.

I took out those FISMA lines, because profile reference rules and not vice versa - if a rule is required by a security policy, it is selected by the appropriate profile. Generally, we want every rule should fit seamlessly to any profile, and mentioning a policy in rule's description is exactly a seam.
The grub superuser password != system root password requirement could become a new rule, but there is no way to check this by OVAL, so is it even worth it to have in the guide just for the description? And if so, the description would yet have to be written, which is out of scope of this PR.

linux_os/guide/system/bootloader-grub2/grub2_password/rule.yml
matejak: The same case as text commented below.

linux_os/guide/system/bootloader-grub2/grub2_admin_username/rule.yml
matejak: This is basically the same as the change commented on below - I have simply transferred references from the original rule. Help on the reference handling is appreciated.

jan-cerny:
@matejak please resolve the conflict.

linux_os/guide/system/bootloader-grub2/grub2_admin_username/oval/shared.xml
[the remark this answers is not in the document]
jan-cerny: @matejak do you have any reaction to the first comment?

linux_os/guide/system/bootloader-grub2/grub2_admin_username/rule.yml
jan-cerny: Sometimes it might happen that the origin policy described only what is covered by the other rule.

linux_os/guide/system/bootloader-grub2/grub2_admin_username/oval/shared.xml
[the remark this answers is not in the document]
matejak: Sorry, forgot for a moment, but it should be good now.

shared/macros-oval.jinja
```diff
@@ -218,6 +218,32 @@
...
+    - filepath (String): Path to the file to be checked.
+    - stem of the test name
+#}}
+{{%- macro oval_config_file_absent_criterion(filepath, stem) -%}}
+	<criterion comment="Pass if {{{ filepath }}} does not exist" test_ref="test_{{{ stem }}}" />
+{{%- endmacro %}}
+
+{{#
+  Macro to define the OVAL test to check if the configuration file exists (Test definition).
+  This macro can take one parameter:
```
jan-cerny:
```suggestion
  This macro can take up to four parameters:
```

shared/macros-oval.jinja
```diff
@@ -218,6 +218,32 @@
...
+    - filepath (String): Path to the configuration file to be checked.
+    - test_version: OVAL test version metadata
+    - object_version: OVAL object version metadata
```
jan-cerny: parameter `stem` is not described

jan-cerny: Also, please don't use random strings as parameter or variable names. Good naming is crucial.

shared/macros-oval.jinja
```diff
@@ -218,6 +218,32 @@
...
+#}}
+{{%- macro oval_config_file_absent(filepath, stem, test_version=1, object_version=1) -%}}
```
jan-cerny: I suggest ignoring the `version` attributes and always set them to 1. Then, the macro could have only 2 parameters.

shared/macros-oval.jinja
```diff
@@ -218,6 +218,32 @@
...
 {{%- endmacro %}}
 
+{{#
+  Macro to define the OVAL criterion that requires a file not to exist.
+  This macro can take one parameter:
+    - filepath (String): Path to the file to be checked.
+    - stem of the test name
+#}}
+{{%- macro oval_config_file_absent_criterion(filepath, stem) -%}}
+	<criterion comment="Pass if {{{ filepath }}} does not exist" test_ref="test_{{{ stem }}}" />
```
jan-cerny: This macro is inappropriate. Please remove it and remove all of its occurrences.

ssg/oval.py
```diff
@@ -98,7 +98,17 @@ def applicable_platforms(oval_file):
...
+
+    oval_filename_components = oval_file.split(os.path.sep)
+    if len(oval_filename_components) > 3:
+        subst_dict["rule_id"] = oval_filename_components[-3]
+    else:
+        msg = "Unable to get rule ID from OVAL path '{path}'".format(path=oval_file)
+        print(msg, file=sys.stderr)
+
+    body = process_file_with_macros(oval_file, subst_dict)
+
```
jan-cerny: What is this change good for?

shared/macros-oval.jinja
jan-cerny:
```suggestion
{{%- macro oval_file_absent(filepath, id) -%}}
```

ssg/oval.py
matejak: A test was failing due to `{{{ rule_id }}` usage in the OVAL snippet. Which is BTW strange, as that is not the only snippet that contains it.

shared/macros-oval.jinja
matejak: Why? We have already a similar concept used successfully elsewhere, see e.g. the `oval_config_file_exists_criterion` macro and more like that in the OVAL jinja file.

shared/macros-oval.jinja
matejak: OK.

shared/macros-oval.jinja
matejak: stem is not a random string: https://en.wiktionary.org/wiki/stem#Noun (see point 7)

shared/macros-oval.jinja
matejak: I went for a copy-paste-friendly "Parameters:"

shared/macros-oval.jinja
jan-cerny: maybe, but nobody knows it, please remove it

shared/macros-oval.jinja
jan-cerny: Yes, please remove that as well.

shared/macros-oval.jinja
jan-cerny: OK, thanks, great idea!

shared/macros-oval.jinja
matejak: OK, relabeled `stem` to `id`.

ssg/oval.py
matejak: Usage of the `rule_id` substitution in the OVAL made a test that uses this functionality to fail. That change makes it passing again.

PR #5277, opened 2020-03-11
Add resources to install the compliance-operator

JAORMX:
This adds the necessary resources (and a script) to install the
compliance operator in an OpenShift cluster. This is currently using the
alpha channel for the operator, and we'll eventually move it to a stable
channel once we have one.

This will eventually be used in CI to deploy the operator and another
(incoming) script will take it into use to test the ocp4 profiles.


mrogers950:
Looks good to me.

shawndwells:
Blindly merging since @jhrozek and @mrogers950 ack'd.

PR #5278, opened 2020-03-11
Fix template mount_option_removable_partitions

jan-cerny:
There are multiple problems that we try to address by this patch.

1. The test scenario dvd_bad_opts.fail.sh for rule
mount_option_noexec_removable_partitions was failing because scanner
returned pass instead of expected fail. That is because the rule checks
if path specified by var_removable_partition exists and if it doesn't
exist the rule fail. This variable is always set to /dev/cdrom in all
profiles, as the variable definition has only this value. See
/linux_os/guide/system/permissions/partitions/var_removable_partition.var
Unless tailoring is used, the rule always passes if /dev/cdrom doesn't
exist. As the test scenario didn't create /dev/cdrom, but /dev/dvd,
the rule passed.

2. The test scenario dvd_good_opts.pass.sh for rule
mount_option_noexec_removable_partitions was also broken even if it
returned OK. The test scenario was creating /dev/dvd. So again, the
OVAL check passed because /dev/cdrom didn't exist.  Therefore the test
scenario was not testing the actual regular expression matching
in /etc/fstab.

3. The Bash and Ansible remediation weren't functional, because they
operated with a mount point (second column of /etc/fstab) but they
were passed values of the block device (which is in first column of
/etc/fstab). I think that people can mount the removable media to
arbitrary directories so we should base our checks on the block device
path instead of mount points. Actually, in all 3 rules we passed the
device path (/dev/cdrom) as a mount point.  That was probably a wrong
understanding of the remediation code. The remediations are reworked
to find entry starting with var_removable_partition value.

4. The OVAL check wasn't checking only /etc/fstab configuration, but was
also checking runtime state. That didn't conform to the rule
description. My understanding of the rule description is that the rule
is only about configuration, there is no mention of runtime.  Moreover,
the removable device doesn't have to be mounted at the moment of system
evaluation.

5. The test scenarios didn't cover multiple situation, so apart of
fixing the broken scenarios, new test scenarios have been added.

We still have a problem that /dev/cdrom is not the only possible
removable device that could exist. Think of USB flash drives, memory
cards, etc. The rule descriptions speak about removable media and
mention USB keys, but the checks don't check them. The challenge
will be recognizing removable media.

Resolves: RHBZ#1691579

This is opened for discussion. I wasn't able to understand this rule. Maybe I accidentally removed something that is intentional. Please review and let me know.

### DIFF
--- a/shared/templates/template_ANSIBLE_mount_option_removable_partitions
+++ b/shared/templates/template_ANSIBLE_mount_option_removable_partitions
@@ -3,31 +3,11 @@
 # strategy = configure
 # complexity = low
 # disruption = high
-- (xccdf-var {{{ MOUNTPOINT }}})
+- (xccdf-var var_removable_partition)
 
-- name: get back mount information associated to mountpoint
-  command: findmnt --fstab '{{ {{{ MOUNTPOINT }}} }}'
-  register: device_name
-  failed_when: device_name.rc > 1
-  changed_when: False
-
-- name: create mount_info dictionary variable
-  set_fact:
-    mount_info: "{{ mount_info|default({})|combine({item.0: item.1}) }}"
-  with_together:
-    - "{{ device_name.stdout_lines[0].split() | list | lower }}"
-    - "{{ device_name.stdout_lines[1].split() | list }}"
-  when:
-    - device_name.stdout is defined and device_name.stdout_lines is defined
-    - (device_name.stdout | length > 0)
-
-- name: Ensure permission {{{ MOUNTOPTION }}} are set on {{{ MOUNTPOINT }}}
-  mount:
-    path: "{{ {{{ MOUNTPOINT }}} }}"
-    src: "{{ mount_info.source }}"
-    opts: "{{ mount_info.options }},{{{ MOUNTOPTION }}}"
-    state: "mounted"
-    fstype: "{{ mount_info.fstype }}"
-  when:
-    - device_name.stdout is defined
-    - (device_name.stdout | length > 0)
+- name: Ensure permission {{{ MOUNTOPTION }}} are set on var_removable_partition
+  lineinfile:
+    path: /etc/fstab
+    regexp: '^\s*({{ var_removable_partition }})\s+([^\s]*)\s+([^\s]*)\s+([^\s]*)(.*)$'
+    backrefs: yes
+    line: '\1 \2 \3 \4,{{{ MOUNTOPTION }}} \5'
--- a/shared/templates/template_BASH_mount_option_removable_partitions
+++ b/shared/templates/template_BASH_mount_option_removable_partitions
@@ -4,19 +4,15 @@
 # Include source function library.
 . /usr/share/scap-security-guide/remediation_functions
 
-populate {{{ MOUNTPOINT }}}
+populate var_removable_partition
 
-include_mount_options_functions
+device_regex="^\s*$var_removable_partition\s\+"
+mount_option="{{{ MOUNTOPTION }}}"
 
-function perform_remediation {
-	# test "$mount_has_to_exist" = 'yes'
-	if test "{{{ MOUNT_HAS_TO_EXIST }}}" = 'yes'; then
-		assert_mount_point_in_fstab "${{{ MOUNTPOINT }}}" || { echo "Not remediating, because there is no record of ${{{ MOUNTPOINT }}} in /etc/fstab" >&2; return 1; }
-	fi
-
-	ensure_mount_option_in_fstab "${{{ MOUNTPOINT }}}" "{{{ MOUNTOPTION }}}" "{{{ FILESYSTEM }}}" "{{{ TYPE }}}"
-
-	ensure_partition_is_mounted "${{{ MOUNTPOINT }}}"
-}
-
-perform_remediation
+if grep -q $device_regex /etc/fstab ; then
+    previous_opts=$(grep $device_regex /etc/fstab | awk '{print $4}')
+    sed -i "s|\($device_regex.*$previous_opts\)|\1,$mount_option|" /etc/fstab
+else
+    echo "Not remediating, because there is no record of $var_removable_partition in /etc/fstab" >&2
+    return 1
+fi
--- a/shared/templates/template_OVAL_mount_option_removable_partitions
+++ b/shared/templates/template_OVAL_mount_option_removable_partitions
@@ -1,39 +1,31 @@
 <def-group>
-  <definition class="compliance" id="mount_option_{{{ MOUNTOPTION }}}_removable_partitions" version="4">
+  <definition class="compliance" id="mount_option_{{{ MOUNTOPTION }}}_removable_partitions" version="5">
     <metadata>
       <title>Add {{{ MOUNTOPTION }}} Option to Removable Media Partitions</title>
       {{{- oval_affected(products) }}}
       <description>The {{{ MOUNTOPTION }}} option should be enabled for all removable devices mounts in /etc/fstab.</description>
     </metadata>
     <criteria operator="OR">
-      <!-- First check if specified removable partition truly exists on the system. If not, don't check /etc/fstab & runtime configuration
+      <!-- First check if specified removable partition truly exists on the system. If not, don't check /etc/fstab
            since there's no device to check against -->
       <extend_definition comment="Check if removable partition really exists on the system"
       definition_ref="removable_partition_doesnt_exist" />
       <!-- Removable device exists. Check if it's CD/DVD drive. If so, verify that at least one from all of the possible its alternative
-           names in /etc/fstab & runtime configuration are configured with '{{{ MOUNTOPTION }}}' option -->
+           names in /etc/fstab are configured with '{{{ MOUNTOPTION }}}' option -->
       <criteria operator="AND">
         <extend_definition comment="Check if removable partition value represents CD/DVD drive"
         definition_ref="var_removable_partition_is_cd_dvd_drive" />
         <criteria operator="OR">
-          <criteria operator="AND">
-            <criterion test_ref="test_{{{ MOUNTOPTION }}}_etc_fstab_cd_dvd_drive"
+          <criterion test_ref="test_{{{ MOUNTOPTION }}}_etc_fstab_cd_dvd_drive"
             comment="Check if at least one from CD/DVD drive alternative names is using '{{{ MOUNTOPTION }}}' mount option in /etc/fstab" />
-            <criterion test_ref="test_{{{ MOUNTOPTION }}}_runtime_cd_dvd_drive"
-            comment="Check if at least one from CD/DVD drive alternative names is using '{{{ MOUNTOPTION }}}' mount option in runtime configuration" />
-          </criteria>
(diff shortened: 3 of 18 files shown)


jan-cerny:
I wonder what happens if there is no explicit entry in fstab. Will defaults will be used? Where can we find them? Should we check for them as well? Should the remediation create a fstab entry if it doesn't exist? If it should, which directory should be used as mountpoint?

shared/templates/template_BASH_mount_option_removable_partitions
```diff
@@ -4,19 +4,15 @@
...
+device_regex="^\s*$var_removable_partition\s\+"
+mount_option="{{{ MOUNTOPTION }}}"
 
-function perform_remediation {
-	# test "$mount_has_to_exist" = 'yes'
-	if test "{{{ MOUNT_HAS_TO_EXIST }}}" = 'yes'; then
-		assert_mount_point_in_fstab "${{{ MOUNTPOINT }}}" || { echo "Not remediating, because there is no record of ${{{ MOUNTPOINT }}} in /etc/fstab" >&2; return 1; }
-	fi
-
-	ensure_mount_option_in_fstab "${{{ MOUNTPOINT }}}" "{{{ MOUNTOPTION }}}" "{{{ FILESYSTEM }}}" "{{{ TYPE }}}"
```
yuumasato: I cannot find where `{{{ FILESYSTEM }}}` was set.

shared/templates/template_ANSIBLE_mount_option_removable_partitions
```diff
@@ -3,31 +3,11 @@
...
-    fstype: "{{ mount_info.fstype }}"
-  when:
-    - device_name.stdout is defined
-    - (device_name.stdout | length > 0)
+- name: Ensure permission {{{ MOUNTOPTION }}} are set on var_removable_partition
+  lineinfile:
+    path: /etc/fstab
+    regexp: '^\s*({{ var_removable_partition }})\s+([^\s]*)\s+([^\s]*)\s+([^\s]*)(.*)$'
+    backrefs: yes
+    line: '\1 \2 \3 \4,{{{ MOUNTOPTION }}} \5'
```
yuumasato: I'm not sure we should expect `/etc/fstab` to already contain an entry for the removable device in question.
I tend to think the remediation should add it.

jan-cerny: Probably yes, sounds reasonable

jan-cerny: But, I have realized another thing now. The rule should pass when `/etc/fstab` doesn't contain the entry. We have a test scenario `no_partitions.pass.sh` and this scenario simply erases all contents of `/etc/fstab`. I understand it was created to cover https://bugzilla.redhat.com/show_bug.cgi?id=1403905. That is an old BZ  where the reporter asks to no fail the scan if the devices are not mounted. I am not sure if the scenario meets his expectation.

yuumasato: Is it possible for the device not to be mounted and still have the entry in `/etc/fstab`?

yuumasato:
> I wonder what happens if there is no explicit entry in fstab. Will defaults will be used? Where can we find them? 

I think yes, they are probably defined by `systemd`. I could not find where though.

> Should we check for them as well? 

If the device is mounted, I think it will be visible through the `partition_test`, it has access to the info as `mount` command.

> Should the remediation create a fstab entry if it doesn't exist?

The way to customize default mount options is by creating an entry in `/etc/fstab`.

> If it should, which directory should be used as mountpoint?

I would probably use some distro default place.

yuumasato:
>     1. The test scenario dvd_bad_opts.fail.sh for rule
>        mount_option_noexec_removable_partitions was failing because scanner
>        returned pass instead of expected fail. That is because the rule checks
>        if path specified by var_removable_partition exists and if it doesn't
>        exist the rule fail. This variable is always set to /dev/cdrom in all
>        profiles, as the variable definition has only this value. See
>        /linux_os/guide/system/permissions/partitions/var_removable_partition.var
>        Unless tailoring is used, the rule always passes if /dev/cdrom doesn't
>        exist. As the test scenario didn't create /dev/cdrom, but /dev/dvd,
>        the rule passed.

I'm thinking right now that the variable is actually useless.

CIS benchmarks mentions removable media like Floppy and CD-ROM.
STIG mentions removable media like USB sticks.

I think the rule is expected to check for *all* types of removable devices, with *all* being defined as the pretty common ones, like floppy disks, CD-ROM, DVD-ROM, USB sticks.

shared/templates/template_BASH_mount_option_removable_partitions
jan-cerny: it wasn't set anywhere

shared/templates/template_ANSIBLE_mount_option_removable_partitions
jan-cerny: I think yes, when the removable media is removed.

jan-cerny:
> I think the rule is expected to check for all types of removable devices, with all being defined as the pretty common ones, like floppy disks, CD-ROM, DVD-ROM, USB sticks.

Yes, I concur, that's mentioned later in the pull request description.

shared/templates/template_OVAL_mount_option_removable_partitions
```diff
@@ -1,39 +1,31 @@
...
         </criteria>
       </criteria>
       <!-- Removable device exists & isn't CD/DVD drive. Check the particular devices is configured with '{{{ MOUNTOPTION }}}' mount option in both
-           /etc/fstab & runtime configuration -->
-      <criteria operator="AND">
-        <criterion test_ref="test_{{{ MOUNTOPTION }}}_etc_fstab_not_cd_dvd_drive"
+           /etc/fstab -->
+      <criterion test_ref="test_{{{ MOUNTOPTION }}}_etc_fstab_not_cd_dvd_drive"
         comment="Check if removable partition is using '{{{ MOUNTOPTION }}}' mount option in /etc/fstab" />
-        <criterion test_ref="test_{{{ MOUNTOPTION }}}_runtime_not_cd_dvd_drive"
```
redhatrises: Runtime needs to be kept and doesn't really need to be mentioned in the prose. There is a expectation which is mentioned in the prose to reboot after configuration has happened when a rule fails.

jan-cerny:
@openscap-ci test this please

jan-cerny:
@openscap-ci test this please

jan-cerny:
@openscap-ci test this please

shared/templates/template_OVAL_mount_option_removable_partitions
matejak: Maybe I am missing something - it seems to me that the check won't pass if there is no removable partition mounted (no objects would be found in that case). Removable devices are not supposed to be mounted most of the time, so forcing them to be present during the scan feels inappropriate.

matejak:
For your reference, this subject has been [discussed on the public mailing list](https://lists.fedoraproject.org/archives/list/scap-security-guide@lists.fedorahosted.org/thread/Q33GUFAVN73AA672ZGDHSFFPZDLHV62D/), and one thing is clear - removable drives landscape in Linux is huge, it pertains many areas, and it is full of exceptions and false friends. Therefore, I propose to restrict this PR to only fix issues reported by the [referenced Bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1691579), and do any additional enhancement to the removable devices handling in other pull requests.

matejak:
test this please

shared/templates/template_ANSIBLE_mount_option_removable_partitions
matejak: Concerning e.g. DVDs, if there is no device file, it implies that the computer doesn't have a drive, so there is nothing needed in fstab and the rule should simply pass.
The remediation could add the fstab line if the device file exists, but that's outside of what Ansible can do easily, and we aren't even sure how exactly should the line components look like. So I think that this is OK.

jan-cerny: The rule passes when /dev/cdrom doesn't exist and this isn't changed by this PR.

However, the rule also passes when /dev/cdrom exists but there is no line with /dev/cdrom in /etc/fstab. This is because of the old bugzilla https://bugzilla.redhat.com/show_bug.cgi?id=1403905. See the regression test for this BZ in https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/permissions/partitions/mount_option_noexec_removable_partitions/tests/no_partitions.pass.sh. From the test scenario you can see that we expect the rule to pass when /dev/cdrom exists but it isn't configured in /etc/fstab. I have revisited this old BZ and I think that this interpretation might be incorrect, because the mount option isn't enforced in this situation. Instead, I think that the rule should always require explicit configuration in /etc/fstab. On the other hand, if we change the behavior in this situation, we create a regression for the customer that reported the old BZ.

I don't know if it's possible to add the fstab line when the fstab line doesn't exist, because we don't know to which mountpoint the user wishes to mount the device. Should we use some predefined directory eg. /media/cdrom?

shared/templates/template_OVAL_mount_option_removable_partitions
jan-cerny: I also don't understand the meaning of the runtime check in this OVAL. Would that mean that users are supposed to pass this check only with CD inserted in the drive?

jan-cerny:
I think that based on the test scenarios the BZ is fixed by this PR. I think that the problem about passing when /etc/fstab doesn't contain the removable media entry isn't directly related to the BZ and therefore can be fixed separately.

jan-cerny:
@openscap-ci test this please

shared/templates/template_ANSIBLE_mount_option_removable_partitions
matejak: Thanks for your answer - the request in that BZ is really strange. However, as you point out, that rule doesn't have a remediation for cases when the entry in `/etc/fstab` is missing, which means no all-green state after the remediation. Creating a remediation that would fix `fstab` and that wouldn't introduce an antipattern is out of scope of this PR, as nobody complains about that aspect of the check.

matejak:
I can confirm that tests from the test suite pass as well. Therefore - it's time to merge the PR. Remaining issues can be addressed in subsequent PRs, however be sure to review the [discussion on the public mailing list](https://lists.fedoraproject.org/archives/list/scap-security-guide@lists.fedorahosted.org/thread/Q33GUFAVN73AA672ZGDHSFFPZDLHV62D/) and a [old bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1403905) before going for a fix, as the information there provides the essential context.

shared/templates/template_OVAL_mount_option_removable_partitions
yuumasato: I think the idea is that if a removable partition is found, it needs to have the correct mount options, not to force that it is mounted.

jan-cerny: I agree with @yuumasato

PR #5281, opened 2020-03-11
Add XCCDF conflicts and requires

redhatrises:
#### Description:

- Adds the capability of specifying `<xccdf:requires/>` and `<xccdf:conflicts/>` in benchmark writing to allow finer control over rules/groups that either depend on each other or conflict with each other.

### DIFF
--- a/docs/manual/developer_guide.adoc
+++ b/docs/manual/developer_guide.adoc
@@ -791,6 +791,10 @@ A rule itself contains these attributes:
 The <tt>/tmp</tt> partition is used as temporary storage by many programs. Placing <tt>/tmp</tt> in its own partition enables the setting of more restrictive mount options, which can help protect programs which use it.
 * `description`: Human-readable HTML description, which provides broader context for non-experts than the rationale. For example, description of the `partition_for_tmp` rule states that:
 +
+* `requires`: The `id` of another rule or group that must be selected and enabled in a profile.
++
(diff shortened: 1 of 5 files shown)


docs/manual/developer_guide.adoc
```diff
@@ -791,6 +791,10 @@ A rule itself contains these attributes:
 The <tt>/tmp</tt> partition is used as temporary storage by many programs. Placing <tt>/tmp</tt> in its own partition enables the setting of more restrictive mount options, which can help protect programs which use it.
 * `description`: Human-readable HTML description, which provides broader context for non-experts than the rationale. For example, description of the `partition_for_tmp` rule states that:
 +
+* `requires`: The `id` of another rule or group that must be selected and enabled in a profile.
```
yuumasato: What is the difference between selected and enabled in a Profile?

redhatrises: @yuumasato trying to not use the XCCDF terms. Basically a rule has to exist in the profile and will flip the switch to selected='true' (aka enabled for processing in the benchmark/profile as selected in this case means to score and evaluate)

yuumasato: I see, thanks for clarifying the language.

In the code you are not actually selection any required rule in the build system, did I miss something?
It would be cool if the build system could auto-select rules for you, based on requires.

And actually, the scanner behavior works the other way. lt won't to select the required rule if it is not selected, it will flip the rule whose requirements are not fulfilled to false.

PR #5283, opened 2020-03-12
Introduce a rule that mandates usage of subset of FIPS SSHD ciphers

matejak:
* Introduce the `sshd_use_stig_ciphers` rule.
That rule is a stricter version of the `sshd_use_approved_ciphers` rule. See also [related ML discussion](https://lists.fedorahosted.org/archives/list/scap-security-guide@lists.fedorahosted.org/thread/WG5DKAUG5DOUZGLFYEKMOEIRWNUGSQ37/)

* Convert the `sshd_use_approved_ciphers` rule to use the template system instead of individual remediations.


redhatrises:
Please make the ciphers a tailored value rather than a new rule.

matejak:
> Please make the ciphers a tailored value rather than a new rule.

I have thought of that as well, as I also don't like duplication, but I see those reasons why a duplicated rule is lesser evil:

- The final set of algorithms would have to be present in the description, OVAL checks and remediations. Each of those occurences requires different formatting, and transforming the variable value is not straightforward.
- As the rule description says that the rule's ciphers have to be subset of FIPS approved ciphers, any variable customization would have to be based on restriction of the set of all FIPS-approved ciphers. We don't want to make it possible for people to define what list of ciphers this rule should check if it mentions FIPS in the description.

If anyone would like to help with implementation, please do so.

yuumasato:
> Please make the ciphers a tailored value rather than a new rule.

The rule `sshd_use_approved_macs` already went down that path.

> * The final set of algorithms would have to be present in the description, OVAL checks and remediations. Each of those occurences requires different formatting, and transforming the variable value is not straightforward.

You can get inspired by how the OVAL and remediations are handled in https://github.com/ComplianceAsCode/content/tree/master/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_macs

To mention the list of ciphers in the rule descripton, `<sub>` may be enough, line in our old friend accounts timeout:
https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/accounts/accounts-session/accounts_tmout/rule.yml#L11

> * As the rule description says that the rule's ciphers have to be subset of FIPS approved ciphers, any variable customization would have to be based on restriction of the set of all FIPS-approved ciphers. We don't want to make it possible for people to define what list of ciphers this rule should check if it mentions FIPS in the description.

Are you concerned that some one would alter the variable and claim that these are now the FIPS approved ciphers?
I guess organizations and auditors would review the content used for the audit.
If the variable is exposed in the rule description, this becomes even clearer.

Is the use case where someone wants even stricter set of ciphers valid?
A customizable variable would allow for that.

matejak:
Thank you for the example, this makes the implementation of that rule easy.

> Are you concerned that some one would alter the variable and claim that these are now the FIPS approved ciphers?
I guess organizations and auditors would review the content used for the audit.
If the variable is exposed in the rule description, this becomes even clearer.

Yes, the "MACs" rule allows any MACs that are specified in the variable, so in some cases the rule may diverge quite heavily from the description. However if it is not a concern, I am fine with it, it would make the content simpler to maintain.

matejak:
Closing in favor of #5308 that adopts the approach suggested by @yuumasato . Thanks!

PR #5282, opened 2020-03-12
add ntpd_configure_restrictions for rhel7

vojtapolasek:
#### Description:

A rule ntpd_configure_restrictions is added for rhel7. Also additional CPE for rhel7 is added called ntp and it checks if ntp package is installed. This is done because the mentioned rule is applicable only if ntp is installed.

#### Rationale:

The rule is part of CIS rhel7 benchmark.

PR #5291, opened 2020-03-16
add rule ntpd_run_as_ntp_user

vojtapolasek:
#### Description:

Create new rule, including oval, bash, ansible, tests. Create also new cpe for Fedora and Rhel7 to check if ntp is installed.

#### Rationale:

Rule is part of CIS profile.


redhatrises:
We should deprecate ntp in favor of chrony in Fedora as well as what has been done in RHEL.

linux_os/guide/services/ntp/ntpd_run_as_ntp_user/oval/shared.xml
```diff
@@ -0,0 +1,36 @@
+<def-group>
+  <definition class="compliance" id="ntpd_run_as_ntp_user" version="1">
+    <metadata>
+      <title>Check if ntpd is configured to run under ntp user account</title>
```
redhatrises:
```suggestion
      <title>Configure ntpd To Run As ntp User</title>
```

linux_os/guide/services/ntp/ntpd_run_as_ntp_user/oval/shared.xml
```diff
@@ -0,0 +1,36 @@
...
+      {{{- oval_affected(products) }}}
+      <description>check two files which can set ntpd command line options for correct values</description>
```
redhatrises:
```suggestion
      <description>Ensure ntpd is configured to run correctly under the ntp user.</description>
```

linux_os/guide/services/ntp/ntpd_run_as_ntp_user/oval/shared.xml
```diff
@@ -0,0 +1,36 @@
...
+    </metadata>
+    <criteria operator="OR">
+      <criterion comment="check /etc/sysconfig/ntpd for correct value" test_ref="test_ntpd_run_as_ntp_user_etc_sysconfig_ntpd" />
```
redhatrises: `check /etc/sysconfig/ntpd is configured correctly`

linux_os/guide/services/ntp/ntpd_run_as_ntp_user/oval/shared.xml
```diff
@@ -0,0 +1,36 @@
...
+      <criterion comment="check /usr/lib/systemd/system/ntpd.service for correct value" test_ref="test_ntpd_run_as_ntp_user_systemd" />
```
redhatrises:
```suggestion
      <criterion comment="check /usr/lib/systemd/system/ntpd.service is configured correctly" test_ref="test_ntpd_run_as_ntp_user_systemd" />
```

linux_os/guide/services/ntp/ntpd_run_as_ntp_user/oval/shared.xml
```diff
@@ -0,0 +1,36 @@
...
+
+    <ind:textfilecontent54_object id="obj_ntpd_run_as_ntp_user_etc_sysconfig_ntpd"
+  version="1" comment="check /etc/sysconfig/ntpd">
+    <ind:filepath>/etc/sysconfig/ntpd</ind:filepath>
+    <ind:pattern operation="pattern match">^[\s]*OPTIONS=.*-u ntp:ntp.*$</ind:pattern>
+    <ind:instance datatype="int">1</ind:instance>
+  </ind:textfilecontent54_object>
+
+
+  <ind:textfilecontent54_test id="test_ntpd_run_as_ntp_user_systemd" version="1" check="all" comment="check /usr/lib/systemd/system/ntpd.service">
```
redhatrises: Why not use the systemd probes?

linux_os/guide/services/ntp/ntpd_run_as_ntp_user/rule.yml
```diff
@@ -0,0 +1,40 @@
...
+    {{{ weblink(link="http://www.ntp.org")}}}.
+    ntp can be configured to be a client and/or a server.
+    To ensure that ntpd is running under ntp user account, Add or edit the
+    <tt>OPTIONS</tt> variable in <tt>/etc/sysconfig/ntpd</tt> to include ' -u ntp:ntp ':
+    <pre>OPTIONS="-u ntp:ntp"</pre>
+    This recommendation only applies if ntp is in use on the system.
+
+rationale: |-
+    If ntp is in use on the system proper configuration is vital to ensuring time synchronization
+    is working properly.
```
redhatrises: I think a better rationale here is needed for why ntp should run as the ntp user.

linux_os/guide/services/ntp/ntpd_run_as_ntp_user/rule.yml
```diff
@@ -0,0 +1,40 @@
+documentation_complete: true
+
+prodtype: rhel7,fedora
+
+title: 'Ensure that ntpd is running under ntp user account'
```
redhatrises:
```suggestion
title: 'Configure ntpd To Run As ntp User'
```

linux_os/guide/services/ntp/ntpd_run_as_ntp_user/rule.yml
```diff
@@ -0,0 +1,40 @@
...
+
+ocil_clause: 'ntpd is not running under ntp user account'
+
+ocil: |-
+    Run the following commands and verify that ' -u ntp:ntp ' is included in OPTIONS or
+    ExecStart as listed:
+    <pre># grep "^OPTIONS" /etc/sysconfig/ntpd
+    OPTIONS="-u ntp:ntp"
+    # grep "^ExecStart" /usr/lib/systemd/system/ntpd.service
+    ExecStart=/usr/sbin/ntpd -u ntp:ntp $OPTIONS</pre>
```
redhatrises:
```suggestion
    To verify that <tt>ntpd</tt> is configured to correctly run under the <tt>ntp</tt> user,
    run the following commands:
    <pre>$ sudo grep "^OPTIONS" /etc/sysconfig/ntpd</pre>
    The output should contain
    <pre>OPTIONS="-u ntp:ntp"</pre>
    <pre>$ sudo grep "^ExecStart" /usr/lib/systemd/system/ntpd.service</pre>
    The output should contain
    <pre>ExecStart=/usr/sbin/ntpd -u ntp:ntp $OPTIONS</pre>
```

redhatrises: In general, always recommend good security practices by referencing `sudo` when providing verification commands e.g.
`$ sudo` instead of `# my_commands`.

linux_os/guide/services/ntp/ntpd_run_as_ntp_user/oval/shared.xml
vojtapolasek: This is a valid question, it did not occur to me that I could use systemd probes. However, I did some investigation and found out that rewriting this single test to use systemd_unit_property is really not a trivial task, especially when talking about ExecStart property. Therefore I suggest that now we leave as it is, because technically it is doing the right thing - checking the unit file. I will open an issue that  similar tests should be rewritten from filecontent tests to systemd_unit_property tests in future as. Is this acceptable for you?

jan-cerny:
Is this blocked by #5282?

vojtapolasek:
No, they should be independent.

linux_os/guide/services/ntp/ntpd_run_as_ntp_user/oval/shared.xml
jan-cerny: @vojtapolasek Can you explain why do you think it's difficult? I think you can get the ExecStart property and then compare it against a regex in an OVAL state. What is the catch?

vojtapolasek: Well the catch is that if you do for example:
$ systemctl show cups.service
You will get array of various values, not just the ExecStart line.
According to OVAL documentation the output of this probe is the same as running
$ systemctl show <unit>
As I said, this should be changed in future for all checks checking for systemd unit files, but it will not be a trivial work.

redhatrises: It should be something actually quite trivial. For example:
````
  <linux:systemdunitproperty_test check="all" check_existence="all_exist" id="test_execstart" comment="test" version="1">
    <linux:object object_ref="obj_execstart" />
    <linux:state state_ref="state_execstart" />
  </linux:systemdunitproperty_test>
  <linux:systemdunitproperty_object id="obj_execstart" version="1">
    <linux:unit>ntpd.service</linux:unit>
    <linux:property>ExecStart</linux:property>
  </linux:systemdunitproperty_object>
  <linux:systemdunitproperty_state id="state_execstart" version="1">
    <linux:unit>ntpd.service</linux:unit>
    <linux:property>ExecStart</linux:property>
    <linux:value>some_value</linux:value>
  </linux:systemdunitproperty_state>
```
However, `ExecStart` is not being gotten by the probe like it should. @jan-cerny is this expected?

jan-cerny:
@openscap-ci test this please

linux_os/guide/services/ntp/ntpd_run_as_ntp_user/oval/shared.xml
jan-cerny: @vojtapolasek You should be able to get only the ExecStart by specifying `ExecStart` in `property` element as @redhatrises has shown. The problem is that this doesn't work in OpenSCAP. I have tried it and the probe simply doesn't return this property. I think it's a bug. I will check it and report this problem as a BZ. Until it's fixed we have to use the textfilecontent probe to read the unit file.

jan-cerny: The problem has been reported as a bug in https://bugzilla.redhat.com/show_bug.cgi?id=1819773.

I think that now we can stay with the check proposed in this PR.

PR #5292, opened 2020-03-16
Select package_audit_installed

### LINKED ISSUE #5261: CIS 4.1.1.1 Ensure auditd is installed (Scored)
CIS calls for libaudit vs the audit RPMs. Need to explore which should be used.

yuumasato:
Package audit depends on audit-libs.

#### Description:

- Select rule to install audit package.

#### Rationale:

- `audit` package requires `audit-libs`, so I think the requirement is pretty much covered.

- Fixes #5261


ggbecker:
Shouldn't we add the CIS reference to the rule? (https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/auditing/package_audit_installed/rule.yml)
Something like:
```
references:
    cis@rhel8: 4.1.1.1
```

yuumasato:
Nice catch, fixed by ~~brute~~ push force.

PR #5294, opened 2020-03-16
Update audit data retention selects and variables

yuumasato:
#### Description:

- Select values for variables related to audit data retention
- Fix rule selected twice


redhatrises:
LGTM

Issue #2851, opened 2018-05-19
pam_faillock tally persistence


sshimko:
pam_faillock modifications don't appear to specify a location on the filesystem to store the account tally state. The default location seems to be /var/run/faillock, which isn't persistent by default on some distros.

I think any content touching faillock should probably set the location for tally state via "dir=" parameter in the PAM configuration to ensure the content reflects distro defaults.

yuumasato:
Interesting issue,
one way to address this would be to define an XCCDF:Value for `dir=`, that will only be used as part of remediation. Advantage of this approach is that one can tailor this value.
Another option is to have a value defined in `product.yml`, specifying default tally log dir for each product. Advantage of this approach is that it is streamlined to work out of the box, but without tailoring capabilities.
A third approach can combine both above, the value in `product.yml` would set the default XCCDF:Value for each product.

ggbecker:
@yuumasato In the first option you could put the tailored directory in the rule description.

marcusburghardt:
Just linking some STIG references to the topic:
https://www.stigviewer.com/stig/red_hat_enterprise_linux_8/2021-12-03/finding/V-230339
https://www.stigviewer.com/stig/red_hat_enterprise_linux_8/2021-12-03/finding/V-250315

marcusburghardt:
Solved by the inclusion of a new rule for `pam_faillock.so` `dir` option.
#9170

PR #5296, opened 2020-03-16
Select rules for audit login events

### LINKED ISSUE #5263: CIS 4.1.4 Ensure login and logout events are collected (Scored)
Need rule to audit /var/log/faillog (or do we?)

yuumasato:
#### Description:

- Select rules to audit:
  - `/var/run/faillock`: audit_rules_login_events_faillock
  - `/var/log/lastlog`: audit_ruels_login_events_lastlog

#### Rationale:

- Lets not use `audit_rules_login_events`as it also audits `/var/run/tallylog`
- `/var/log/faillog` doesn't exist, and seems to be a typo. In RHEL7 CIS Benchmark, it is correctly listed as `/var/run/faillock`.
- Fixes #5263


redhatrises:
We should open a new ticket to configure `/var/run/faillock` under `/var/log/faillock`. The default is a bad setting, because it is cleared on reboot. This allows for n number of tries.

yuumasato:
> We should open a new ticket to configure `/var/run/faillock` under `/var/log/faillock`. The default is a bad setting, because it is cleared on reboot. This allows for n number of tries.

@redhatrises I don't understand what you are asking here.

vojtapolasek:
In audit_rules_login_events_faillog, I see there is a reference
cis: 5.2.8
Looking into benchmarks it is unrelated to this rule.

vojtapolasek:
> We should open a new ticket to configure `/var/run/faillock` under `/var/log/faillock`. The default is a bad setting, because it is cleared on reboot. This allows for n number of tries.

@redhatrises  Do you mean opening a ticket against pam which owns the faillock utility? Theoretically yes. But it is not related to this PR.

yuumasato:
> In audit_rules_login_events_faillog, I see there is a reference
> cis: 5.2.8
> Looking into benchmarks it is unrelated to this rule.

Nice catch, I took the chance to piggyback RHEL7 CIS references.

vojtapolasek:
thanks, rest looks good

redhatrises:
> > We should open a new ticket to configure `/var/run/faillock` under `/var/log/faillock`. The default is a bad setting, because it is cleared on reboot. This allows for n number of tries.
> 
> @redhatrises Do you mean opening a ticket against pam which owns the faillock utility? Theoretically yes. But it is not related to this PR.

No. This is already configurable in `/etc/security/faillock.conf`. There needs to be a new rule and check that changes from `/var/run/faillock` to `/var/log/faillock` in our security benchmarks.

redhatrises:
Looks like the ticket already exists upstream. See https://github.com/ComplianceAsCode/content/issues/2851

yuumasato:
> No. This is already configurable in `/etc/security/faillock.conf`. There needs to be a new rule and check that changes from `/var/run/faillock` to `/var/log/faillock` in our security benchmarks.

> Looks like the ticket already exists upstream. See #2851

@redhatrises I see, thanks for pointing the new setting.
From what I see the issue itself would require a new rule, to make the change in `faillock.conf`. And `audit_rules_login_events_faillock` would have to be changed to follow any setting defined by this new rule.
It would probably become a new variable, and the rule would probably move out of the `audit_rules_login_events` template.

This is more than I envisioned for this PR, and I consider to be out of scope.

redhatrises:
> > No. This is already configurable in `/etc/security/faillock.conf`. There needs to be a new rule and check that changes from `/var/run/faillock` to `/var/log/faillock` in our security benchmarks.
> 
> > Looks like the ticket already exists upstream. See #2851
> 
> @redhatrises I see, thanks for pointing the new setting.
> From what I see the issue itself would require a new rule, to make the change in `faillock.conf`. And `audit_rules_login_events_faillock` would have to be changed to follow any setting defined by this new rule.
> It would probably become a new variable, and the rule would probably move out of the `audit_rules_login_events` template.
> 
> This is more than I envisioned for this PR, and I consider to be out of scope.

Agreed. That's why I asked for a ticket to be opened. However, we already have one open.

PR #5297, opened 2020-03-16
Select rules for SSH and add references

yuumasato:
#### Description:

- Select existing SSH rules
- Add references for CIS benchmark


redhatrises:
If the CIS references are literally the same, I don't think it makes sense to break them out. There are some breakouts in this PR that do make sense because the references are different.

yuumasato:

Now they are specific for RHEL instead of being totally generic. The requirement number varies greatly from product and version.

vojtapolasek:
@redhatrises  See my comment at https://github.com/ComplianceAsCode/content/pull/5302#issuecomment-601091692

linux_os/guide/services/ssh/ssh_server/sshd_enable_x11_forwarding/rule.yml
```diff
@@ -22,7 +22,8 @@ identifiers:
     cce@rhel8: 82421-9
 
 references:
-    cis: 5.2.4
+    cis@rhel7: 5.2.4
```
vojtapolasek: This reference is about disabling x11 forwarding, but the rule is about enabling x11 forwarding. Same for rhel8.

rhel8/profiles/cis.profile
```diff
@@ -710,52 +710,67 @@ selections:
...
+    - file_permissions_sshd_pub_key
+    # TO DO: check owner of pub keys in /etc/ssh is root:root
 
     ### 5.2.5 Ensure SSH LogLevel is appropriate (Scored)
-
+    - sshd_set_loglevel_info
 
     ### 5.2.6 Ensure SSH X11 forward is diabled (Scored)
-
+    - sshd_enable_x11_forwarding
```
vojtapolasek: We want the forwarding disabled, not enabled

yuumasato: Nice catch, thank you. Removed the rule.

linux_os/guide/services/ssh/ssh_server/sshd_enable_x11_forwarding/rule.yml
yuumasato: Removed as well.

vojtapolasek:
Thank you, merging

yuumasato:
Thanks for the review

PR #5299, opened 2020-03-17
add ntp and chrony cpes

vojtapolasek:
#### Description:

Add CPEs for package_ntp_installed and package_chrony_installed to relevant platforms I know about.

#### Rationale:

These applicability CPEs will be used in batch of rules pertaining to ntp and chrony.


yuumasato:
LGTM.

PR #5300, opened 2020-03-17
Ticket40400 sshd use approved macs

cyarbrough76:
#### Description:

- Edited rhel7 stig profile to reference the .var file for current values. Edited the .var file to include current stig-required values for rhel7.

#### Rationale:

- The default values are no longer stig compliant. Scan and remediation previously passed on rhel 7 targets with incorrect values for  FIPS approved macs.

### DIFF
--- a/linux_os/guide/services/ssh/sshd_approved_macs.var
+++ b/linux_os/guide/services/ssh/sshd_approved_macs.var
@@ -11,6 +11,5 @@ operator: equals
 interactive: false
 
 options:
-    sle12_stig: hmac-sha2-512,hmac-sha2-256
-	rhel7_stig: hmac-sha2-512,hmac-sha2-256
+    stig: hmac-sha2-512,hmac-sha2-256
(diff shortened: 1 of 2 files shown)


linux_os/guide/services/ssh/sshd_approved_macs.var
```diff
@@ -12,4 +12,5 @@ interactive: false
 
 options:
     sle12_stig: hmac-sha2-512,hmac-sha2-256
+	rhel7_stig: hmac-sha2-512,hmac-sha2-256
```
redhatrises: @cyarbrough76 as the variables are the exact same, please merge the `sle12_stig` and `rhel7_stig` into the same variable called `stig`.

redhatrises: @cyarbrough76 use spaces and not tabs. That's why your indentation is wrong here.

cyarbrough76: Concur, will do.

cyarbrough76: Did see that last message in time, will redo to address spacing

cyarbrough76: Spacing should be good now

redhatrises:
ACk

PR #5301, opened 2020-03-17
Select rules for system file permissions

yuumasato:
#### Description:

- Select existing rules for system permissions
- Add references
- I've added other related references to `rpm_verify_ownership` and `rpm_verify_permissions`, as these rules can contribute to them.
  I don't think this rule removes the need for specific rules for the other referenced reqs, as the package setting may deviate from policy expectation, and vice-versa.


redhatrises:
If the CIS references are literally the same, I don't think it makes sense to break them out.

yuumasato:

Now they are specific for RHEL instead of being totally generic. The requirement number varies from product and version.

matejak:
I think that literally nothing can go wrong when references are converted to be product-specific - they won't change, so they don't represent maintenance burden, and it makes rules future-proof - adding support for one more product will mean adding one line - no need for rule refactoring is guaranteed.

matejak:
LGTM, merging!

yuumasato:
Thanks for the review.

PR #5302, opened 2020-03-17
Banner and cron permissions and owners

yuumasato:
#### Description:

- Update CIS profile with rules and cron file permissions and owners
- Add rules for banner permissions
- Update CIS references for all rules mentioned above


redhatrises:
If the CIS references are literally the same, I don't think it makes sense to break them out.

redhatrises:
If the CIS references are literally the same, I don't think it makes sense to break them out.

yuumasato:

Now they are specific for RHEL instead of being totally generic. The requirement number varies greatly from product and version.

vojtapolasek:
I am not sure if there is any benefit in adding the chmod / chown commands directly to the profile as a comment. From my point of view it is another thing which needs to be checked if it matches with the corresponding rule... othervise LGTM. Thank you.

vojtapolasek:
Concerning the cis references, I see a benefit in having them split now even if they are currently the same for rhel7 and rhel8. I think this is a futureproof solution. We have no certainty that they won't change for example for rhel9. If we keep it as it is, we will have to go through all of them later and create variants for rhel7 / rhel8 / rhel9 in case they change in rhel9 or even later.

yuumasato:
> I am not sure if there is any benefit in adding the chmod / chown commands directly to the profile as a comment. From my point of view it is another thing which needs to be checked if it matches with the corresponding rule...

I followed the practice in the PR for other rules.
From my PoV, it makes it very clear what rules are covering what aspects of the requirement.
But it also means more work for the author, and also for the reviewer, as you mentioned.
If the Benchmark is updated these comments can go out of date.

I don't have a strong opinion on this.

vojtapolasek:
I agree that these notes are very ilustrative. Maybe we could later, when the profile is finalized, state somewhere, that these notes might not be exactly up-to-date. Other things look goot in this PR.

PR #5304, opened 2020-03-17
Ticket40450 sshd enable strictmodes

cyarbrough76:
#### Description:

- Edited rule.yml, changed missing parameter pass to false per stig 40450

#### Rationale:

- Scan was showing pass even though parameter was not present in file. Stig shows this as a finding.


cyarbrough76:
Unclear why both commits above appear in this PR. The first commit (40440) was covered under a separate PR.

redhatrises:
@cyarbrough76 can you remove the offending commit from this PR?

cyarbrough76:
I would like to. I cannot see a way to do that.

redhatrises:
`git rebase -i HEAD~3` then select the commit to drop by changing the word pick to drop.

PR #5306, opened 2020-03-18
Add Initial RHEL 7 CIS profile

yuumasato:
#### Description:

- Initial RHEL7 CIS profile

~~[ ] Update RHEL7 CIS references~~

### DIFF
--- a/rhel7/profiles/cis.profile
+++ b/rhel7/profiles/cis.profile
@@ -1,0 +1,729 @@
+documentation_complete: true
+
+title: 'CIS Red Hat Enterprise Linux 7 Benchmark'
+
+description: |-
+    This baseline aligns to the Center for Internet Security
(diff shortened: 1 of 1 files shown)


redhatrises:
@yuumasato have you also compared this profile to the C2S profile?

yuumasato:
> @yuumasato have you also compared this profile to the C2S profile?

Yes, they are very similar...but as far as I know (or told) they shouldn't be the same.

yuumasato:
I've decide to update the references later, when there is consensus whether having `cis@rhel7` and `cis@rhel8` when references are similar is worth it.

yuumasato:
Also, let's not block parallel work that can leverage a merged profile.

redhatrises:
No argument here. Ack

PR #5309, opened 2020-03-18
remove ntp mention from rule title

vojtapolasek:
#### Description:

Remove reference to ntp from title of rule chronyd_specify_remote_server.

#### Rationale:

It might be confused with the ntp package.


redhatrises:
Thanks

PR #5310, opened 2020-03-18
changed missing parameter pass to false per stig requirement

cyarbrough76:
#### Description:

- SSHD enable strictmodes, missing parameter pass changed to false

#### Rationale:

- Current STIG requires parameter be present

### DIFF
--- a/linux_os/guide/services/ssh/ssh_server/sshd_enable_strictmodes/rule.yml
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_enable_strictmodes/rule.yml
@@ -46,7 +46,7 @@ ocil: |-
 template:
     name: sshd_lineinfile
     vars:
-        missing_parameter_pass: 'true'
+        missing_parameter_pass: 'false'
         parameter: StrictModes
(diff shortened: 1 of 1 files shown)


shawndwells:
@redhatrises @cyarbrough76 Since strictmodes defaults to enabled there is no need to explicitly check for it. 

This represents a bug in the DISA content, not ComplianceAsCode, and should be reverted.

thenefield:
> @redhatrises @cyarbrough76 Since strictmodes defaults to enabled there is no need to explicitly check for it.
> 
> This represents a bug in the DISA content, not ComplianceAsCode, and should be reverted.

These content changes are to support compliance to the RHEL7 STIG. The change in question is specific to the RHEL7 STIG. Although in theory, the absence of the parameter may still exhibit the required default behavior, but to a person validating compliance against the STIG, it will be deemed as a failure. Differences of opinion with DISA should be brought up to them. But the goal of this project is to support compliance with a number of different security standards, in this case the DISA STIG. I would politely ask that you add back the required setting that was proposed. Its not hurting anything by being there to meet compliance.

PR #5314, opened 2020-03-19
Mention Compliance as Code blog in main README page.

ggbecker:
Make our blog more visible by adding it to our main README page.

Please tell me if you agree with where the link was put.

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,3 +287,4 @@ The SSG homepage is [https://www.open-scap.org/security-policies/scap-security-g
 
 * [SSG User Manual](docs/manual/user_guide.adoc)
 * [SSG Developer Guide](docs/manual/developer_guide.adoc)
+* [Compliance As Code Blog](https://complianceascode.github.io/)


redhatrises:
The blog and the website need to be integrated together. Having 2 separate websites is a bad user design and experience. This link should go in the `Further Reading` section.

ggbecker:
> Needs to move to the further reading section

I moved it to the further reading section.

And the problem with the OpenSCAP website is that it's really difficult to maintain since runs on wordpress. The blog was created with the idea of being simple and easy to contribute. And it would contain topics related to ComplianceAsCode and I'm not sure if fits `scap-security-guide`. @matejak do you have any other points regarding this topic?

redhatrises:
Our website is very dated anyway... sounds like we need a backlog item for updating it and moving it to be not wordpress. Jeykll might make things a lot easier. But, all this can be tackled later in life.

PR #5315, opened 2020-03-19
40470 sshd disable compression

cyarbrough76:
#### Description:

- SSHD disable compression. Add reference to var file in rhel7 profile. Missing param pass changed to false per STIG in oval.

#### Rationale:

- Rhel7 stig profile did not list the .var. STIG 40470 requires parameter to be present in file.

### DIFF
--- a/linux_os/guide/services/ssh/ssh_server/sshd_disable_compression/oval/shared.xml
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_disable_compression/oval/shared.xml
@@ -1,1 +1,1 @@
-{{{ oval_sshd_config(parameter="Compression", value="(no|delayed)", missing_parameter_pass=true) }}}
+{{{ oval_sshd_config(parameter="Compression", value="(no|delayed)", missing_parameter_pass=false) }}}
--- a/rhel7/profiles/stig.profile
+++ b/rhel7/profiles/stig.profile
@@ -22,7 +22,6 @@ selections:
     - var_screensaver_lock_delay=5_seconds
     - sshd_idle_timeout_value=10_minutes
     - sshd_approved_macs=stig
-    - var_sshd_disable_compression=no
     - var_accounts_fail_delay=4
     - var_selinux_state=enforcing
     - var_selinux_policy_name=targeted
--- a/rhel7/profiles/stig.profile
+++ b/rhel7/profiles/stig.profile
@@ -22,6 +22,7 @@ selections:
     - var_screensaver_lock_delay=5_seconds
     - sshd_idle_timeout_value=10_minutes
     - sshd_approved_macs=stig
+    - var_sshd_disable_compression=no
     - var_accounts_fail_delay=4
     - var_selinux_state=enforcing
     - var_selinux_policy_name=targeted


redhatrises:
@cyarbrough76 the default is no. So, the variable does not need to be added to the `stig.profile`.

cyarbrough76:
Understood and corrected. Thank you.

mildas:
Tests are failing now:
```
./test_suite.py rule --libvirt qemu:///system test-suite-rhel7 --profile stig --datastream ../build/ssg-rhel7-ds.xml sshd_disable_compression
INFO - Logging into /home/mlysonek/SCAP/content/tests/logs/rule-custom-2020-03-19-1951-1/test_suite.log
INFO - xccdf_org.ssgproject.content_rule_sshd_disable_compression
ERROR - Script comment.pass.sh using profile xccdf_org.ssgproject.content_profile_stig found issue:
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_sshd_disable_compression'.
ERROR - Script line_not_there.pass.sh using profile xccdf_org.ssgproject.content_profile_stig found issue:
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_sshd_disable_compression'.
INFO - Script other_value.pass.sh using profile xccdf_org.ssgproject.content_profile_stig OK
INFO - Script wrong_value.fail.sh using profile xccdf_org.ssgproject.content_profile_stig OK                            
INFO - Script correct_value.pass.sh using profile xccdf_org.ssgproject.content_profile_stig OK
```

If it's correct that the rule should fail when the parameter is not present (even when the default value is correct) then fix related test scenarios.

ggbecker:
I agree that STIG says it's a finding if the parameter is missing or commented out. But this has been discussed in the past and we decided to not check for default values. The idea is to check only when it deviates from the secure defaults. Here is the discussion: https://github.com/ComplianceAsCode/content/pull/4424

So I see two options here: we revert all PRs that touched these sshd configuration recently or we update all rules that are not checking for the defaults.

What do you think @redhatrises ?

Also @mildas has pointed out that test scenarios need to be updated. I know that if you are not familiar with linux development it might be hard to setup a test environment. But in this case it should be enough to change the expected result of a test for example the following file:

- `linux_os/guide/services/ssh/ssh_server/sshd_disable_compression/tests/comment.pass.sh`
would change to `linux_os/guide/services/ssh/ssh_server/sshd_disable_compression/tests/comment.fail.sh`

More details on: [SSG Test Suite](https://github.com/ComplianceAsCode/content/tree/master/tests) and more specifically on [how-rule-validation-scenarios-work](https://github.com/ComplianceAsCode/content/tree/master/tests#how-rule-validation-scenarios-work)

yuumasato:
Before continuing with switching rules to explicitly add default settings, we should discuss what is expected.
It seems different parties have different expectations, this was just filed: https://github.com/ComplianceAsCode/content/issues/5322

redhatrises:
@cyarbrough76 the community in general is against this as we saw with tickets being opened up, and it was my fault for merging in this regard. The best thing to do is to open tickets with DISA to request an update in language to say something like `If the "Compression" keyword is set to "yes", or is not commented out, this is a finding.` This should be done for a default settings.

redhatrises:
Closing per previous comment.

PR #5319, opened 2020-03-19
Fix rule `banner_etc_motd`

yuumasato:
#### Description:

- Improve `banner_etc_motd` to be as great as `banner_etc_issue`.
- Updated test scenarios


jan-cerny:
Thanks for updating the test scenarios - they pass for me with both Ansible and Bash remediations.

PR #5320, opened 2020-03-19
Update rule.yml

cyarbrough76:
#### Description:

- Changed missing parameter pass to false per stig 10470 requirement

#### Rationale:

- Scan reported pass when parameter was not present. STIG requires parameter to be present.

### DIFF
--- a/linux_os/guide/services/ssh/ssh_server/disable_host_auth/rule.yml
+++ b/linux_os/guide/services/ssh/ssh_server/disable_host_auth/rule.yml
@@ -50,7 +50,7 @@ references:
 template:
     name: sshd_lineinfile
     vars:
-        missing_parameter_pass: 'true'
+        missing_parameter_pass: 'false'
         parameter: HostbasedAuthentication
         rule_id: disable_host_auth
         value: 'no'


redhatrises:
@openscap-ci add to whitelist

redhatrises:
@openscap-ci test this please

mildas:
Tests are failing now:
```
$ ./test_suite.py rule --libvirt qemu:///system test-suite-rhel7 --profile stig --datastream ../build/ssg-rhel7-ds.xml disable_host_auth
INFO - Logging into /home/mlysonek/SCAP/content/tests/logs/rule-custom-2020-03-19-2004/test_suite.log
INFO - xccdf_org.ssgproject.content_rule_disable_host_auth
ERROR - Script comment.pass.sh using profile xccdf_org.ssgproject.content_profile_stig found issue:
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_disable_host_auth'.
ERROR - Script line_not_there.pass.sh using profile xccdf_org.ssgproject.content_profile_stig found issue:
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_disable_host_auth'.
INFO - Script wrong_value.fail.sh using profile xccdf_org.ssgproject.content_profile_stig OK
INFO - Script correct_value.pass.sh using profile xccdf_org.ssgproject.content_profile_stig OK
```
If it's correct that the rule should fail when the parameter is not present (even when the default value is correct) then fix related test scenarios.

cyarbrough76:
Based on feedback from similar rules. It appears that the community's powers that be do not want to add parameters if the default setting is technically correct. Even though the STIG wording specifies that the parameter be present and a missing parameter will continue to fail other scans and remain problematic. But if that is the consensus, I suppose I should just close this?

redhatrises:
@cyarbrough76 the community in general is against this as we saw with tickets being opened up, and it was my fault for merging in this regard. The best thing to do is to open tickets with DISA to request an update in language to say something like `If the "HostbasedAuthentication" keyword is set to "yes", or is not commented out, this is a finding.` This should be done for a default settings.

PR #5325, opened 2020-03-20
Fix service check service_chronyd_enabled to use proper rhel package name

Klaas-:
#### Description:
Fix service check service_chronyd_enabled to use proper rhel package name
currently it checks for packagename chronyd but package is chrony in rhel7/8

#### Rationale:
Use proper package name. Thanks for the help on irc with locating this.


yuumasato:
@openscap-ci add to whitelist

yuumasato:
Thank you for the patch.

PR #5326, opened 2020-03-20
Introduced a rule that uses non-standard yaml checks

matejak:
This rule introduces the to-be-standardized, but openscap-supported yamlfilecontent test.

Remarks:

- The rule is present in the `linux_os`  benchmark tree without prodtype restriction, although it is clearly `ocp4`-only rule. This is intentional - the point is to make it really easy to test e.g. on a Fedora container. When the integration part of the testing is stable, the PR will be marked as ready for review, and the rule will move to a correct location and get correct metadata.
- The rule doesn't scan files on the scanned system. It scans cluster setting, i.e. it assumes that a layer above will export them to yaml files that will end up in a directory. That's what the `ocp_data_root` is for.
- The content project is able to build e.g. a Fedora datastream that is formally invalid (the yaml OVALs are not OVAL 5.11),  but otherwise it is fine.
- The upstream-built scanner is able to evaluate the datastream when `--skip-valid` is supplied. You can execute a test setup script and then scan using 
  ```
  ./oscap_wrapper xccdf eval --rule xccdf_org.ssgproject.content_rule_ocp_data_root --profile '(all)' --report r.html <DS>
  ```
  That works, but I have found the OVAL state to be completely ignored by the scanner. Our scanner unit tests don't suffer from this "screw the state" phenomenon. I have tried to update the scanner's schemas, but I wasn't able to get rid of validation errors.


redhatrises:
Nice!!! As this is an ocp rule, please move under `applications/openshift`. Understand the point about testing, but that is what the OCP test system is for.

mrogers950:
@matejak Looks great!  Regarding the yaml export layer, the yaml-fetching component will need some way to source from the content a list of objects to fetch for a particular check. Is there a tag or some metadata that we can use?

linux_os/guide/system/ocp_idp_no_htpasswd/oval/shared.xml
```diff
@@ -0,0 +1,33 @@
...
+        <title></title>
+        {{{- oval_affected(products) }}}
+        <description>Sample rule</description>
+      </metadata>
+      <criteria operator="AND">
+        <criterion comment="Test that no element of spec.identityProviders has type htpasswd" test_ref="test_{{{ rule_id }}}"/>
+      </criteria>
+    </definition>
+
+    <ind:yamlfilecontent_test id="test_{{{ rule_id }}}" check_existence="none_exist" check="all" comment="No match of htpasswd in ..." version="1">
```
yuumasato: Maybe the state is ignored because the test expects not objects to exist (`check_existence="none_exist`). The mere existence of the object is enough to result in fail, and the state doesn't need to evaluated.

linux_os/guide/system/ocp_idp_no_htpasswd/oval/shared.xml
```diff
@@ -0,0 +1,33 @@
+{{% set YAML_TEST_OVAL_VERSION = [5, 11] %}}
```
yuumasato: Shouldn't this be 5.11.3?
To align with, https://github.com/OpenSCAP/openscap/pull/1482

linux_os/guide/system/ocp_idp_no_htpasswd/oval/shared.xml
matejak: I see, thanks for the insight! So what about this - look for at least one object of the state that we don't want, and then negate the result of the test. Does this sound like a good way to have it done?

yuumasato: I think so, some of our tests are written this way.

linux_os/guide/system/ocp_idp_no_htpasswd/oval/shared.xml
matejak: We don't operate with "patch" versions of the OVAL content. The oscap validation is performed on the datastream-level, and the datastream XSD decides what OVAL schema to use for validation (see also https://github.com/OpenSCAP/openscap/pull/1500)

matejak:
> @matejak Looks great! Regarding the yaml export layer, the yaml-fetching component will need some way to source from the content a list of objects to fetch for a particular check. Is there a tag or some metadata that we can use?

I see, you would like to have an easy way how to assure that export of data from a cluster is complete wrt rules, and maybe also vice versa - that every export has a rule. I am not so sure about metadata, but we can certainly encode that in the rule ID (i.e. `ocp_<file to be exported>_<...>`)

linux_os/guide/system/ocp_idp_no_htpasswd/oval/shared.xml
matejak: I have rebased. The passing scenario now behaves as expected, but the failing scenario doesn't work - when there are two occurrences of `type` when one is OK and one is forbidden (and supposed to be matched by the state), the test concludes that the state doesn't match the object. Could you please help me with this? It's quite easy to test even locally as long as one passes the `--skip-valid` option.

mrogers950:
> > @matejak Looks great! Regarding the yaml export layer, the yaml-fetching component will need some way to source from the content a list of objects to fetch for a particular check. Is there a tag or some metadata that we can use?
> 
> I see, you would like to have an easy way how to assure that export of data from a cluster is complete wrt rules, and maybe also vice versa - that every export is associated with at least one rule. I am not so sure about metadata, but we can certainly encode that in the rule ID (i.e. `ocp_<file to be exported>_<...>`)

@matejak I think that would be one part of it, we also need to be able to define in the rule the OpenShift/Kube object[s] (by name and namespace) that is expected. That way, the fetching component can find the object name and namespace, fetch it, and create the correct file.

yuumasato:
> @matejak I think that would be one part of it, we also need to be able to define in the rule the OpenShift/Kube object[s] (by name and namespace) that is expected. That way, the fetching component can find the object name and namespace, fetch it, and create the correct file.

Maybe the object can be implied from the yaml file name it wants to check, defined by path: https://github.com/ComplianceAsCode/content/pull/5326/files#diff-87becaa7ccca4c70f4c1b623dc92336fR22

mrogers950:
> > @matejak I think that would be one part of it, we also need to be able to define in the rule the OpenShift/Kube object[s] (by name and namespace) that is expected. That way, the fetching component can find the object name and namespace, fetch it, and create the correct file.
> 
> Maybe the object can be implied from the yaml file name it wants to check, defined by path: https://github.com/ComplianceAsCode/content/pull/5326/files#diff-87becaa7ccca4c70f4c1b623dc92336fR22

That might be a bit unwieldy as the object path would be something like `apis/apiextensions.k8s.io/v1/customresourcedefinitions/apiservers.config.openshift.io`. (My previous comment I mentioned name/namespace but we should probably include a full API path, which can be used directly by the collector component and handles non-namespaced resources just the same. )

linux_os/guide/system/ocp_idp_no_htpasswd/oval/shared.xml
yuumasato: Set  set attribute `entity_check` of element `value_of` in the state to `at least one`:
`<ind:value_of check_existence="at least one" >htpasswd</ind:value_of>`

The default value of `entity_check` is `all`, [reference](https://github.com/OVALProject/Language/blob/7fa7bba7b48f09decb732d00b2be032a487ff9fc/docs/oval-definitions-schema.md#-entitystatesimplebasetype-).

So when there are multiple entities, it will only evaluate to `true` when all occurrences of `type` have value `htpasswd`.

linux_os/guide/system/ocp_idp_no_htpasswd/oval/shared.xml
jan-cerny: I see, but doesn't that mean we will start generating invalid content by default?

matejak: The YAML test qualifies as an [unofficial OVAL extension](https://oval-community-guidelines.readthedocs.io/en/latest/proposal-process/release-process.html#how-to-release-streams). We can say with high certainty that it will take a lot of time before it is part of the official OVAL - the process works like this: unofficial -> development -> stable -> official.

matejak:
The rule is now considered to be in a good shape. It works with the upstream-built scanner if you pass it the `--skip-valid` additional option to the `xccdf eval` arguments module.
I will relocate the rule and set its prodtype to `ocp4` tomorrow if there is no other feedback.

linux_os/guide/system/ocp_idp_no_htpasswd/oval/shared.xml
jan-cerny: @matejak So do I understand it well that you want to build, release and ship only invalid content? Or will it be a special option that OCP developers have to turn on manually?

jan-cerny:
How is it possible that this passes Jenkins?

linux_os/guide/system/ocp_idp_no_htpasswd/oval/shared.xml
```diff
@@ -0,0 +1,33 @@
...
+
+    <ind:yamlfilecontent_test id="test_{{{ rule_id }}}" check="at least one" comment="Find at least one match of type: httpasswd among all identity providers." version="1">
+	    <ind:object object_ref="object_{{{ rule_id }}}"/>
+	    <ind:state state_ref="state_{{{ rule_id }}}"/>
+    </ind:yamlfilecontent_test>
+
+    <ind:yamlfilecontent_object id="object_{{{ rule_id }}}" version="1">
+      <ind:path var_ref="ocp_data_root"/>
+      <ind:filename operation="pattern match">^idp.(yaml|yml)$</ind:filename>
+      <ind:yamlpath>.spec.identityProviders[:].type</ind:yamlpath>
```
mrogers950: @matejak is this the area where we can define the collector metadata? I.E. adding `<ind:apipath>/api/v1/foo</ind:apipath>`
Or is this something that we would need to have set in rule.yml that is built into the datastream?

matejak: This is the part where the exact check is. XML Elements here are subjects to schemas, and the [yamlfilecontent check OVAL schema](https://github.com/OVAL-Community/OVAL/pull/90/files) won't allow any other elements here.
What we were discussing is a XCCDF metadata, which would go to the rule.yml, and it could look like this:

```
metadata: 
  - 'ocp_api_endpoint=/api/v1/foo'
```

The resulting metadata element would become a sub-element of the appropriate `Rule` XCCDF element of the datastream.
For reference, the code that you have pointed to would end up scattered in the OVAL section of the datastream (tests, objects and states end up in different locations within the OVAL datastream section). The "parent" OVAL element is the the OVAL definition, and that one is referenced by that XCCDF rule.

mrogers950: @matejak perfect, thanks for the explanation!

matejak:
Our CI obviously has issues with the yamlfilecontent check, which is not valid OVAL yet.

linux_os/guide/system/ocp_idp_no_htpasswd/oval/shared.xml
redhatrises: If setting `ocp_api_endpoint=/api/v1/foo` changes or influences evaluation in any way, this use of metadata isn't allowed per the XCCDF spec. You'd have to use <xccdf:Value> instead, so I take back my metadata suggestion.

matejak: My understanding is that the metadata wouldn't influence the OVAL check in any way. It is about another tool operating on a higher level could use the rule metadata to prepare the environment for the check itself.

JAORMX:
This causes an error in the ocp4 tests: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/ComplianceAsCode_content/5326/pull-ci-ComplianceAsCode-content-master-e2e-aws-moderate/17

mrogers950:
> This causes an error in the ocp4 tests: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/ComplianceAsCode_content/5326/pull-ci-ComplianceAsCode-content-master-e2e-aws-moderate/17

These errors are consistent with what I'm seeing when running this PR with the prototype resource collector:
```
status:
  errormsg: |
    OpenSCAP Error: File '/content/ssg-ocp4-ds.xml' line 12262: Element '{http://oval.mitre.org/XMLSchema/oval-definitions-5#independent}yamlfilecontent_test': This element is not expected.
     [/builddir/build/BUILD/openscap-1.3.2/src/XCCDF/xccdf_session.c:613]
    File '/content/ssg-ocp4-ds.xml' line 17203: Element '{http://oval.mitre.org/XMLSchema/oval-definitions-5#independent}yamlfilecontent_object': This element is not expected.
     [/builddir/build/BUILD/openscap-1.3.2/src/XCCDF/xccdf_session.c:613]
    File '/content/ssg-ocp4-ds.xml' line 23749: Element '{http://oval.mitre.org/XMLSchema/oval-definitions-5#independent}yamlfilecontent_state': This element is not expected.
     [/builddir/build/BUILD/openscap-1.3.2/src/XCCDF/xccdf_session.c:613]


```

matejak:
I have added a warning to the rule about necessity of retrieval of cluster config before the actual scan. As the warning is in the datastream, one can extract the relevant information using xpath expressions. I have managed to get so close using `xmllint` xpath capabilities:

`xmllint --xpath "//*[local-name()='Rule']/*[local-name()='warning']/*[local-name()='code']" build/ssg-ocp4-ds.xml  | grep ocp`

linux_os/guide/system/ocp_idp_no_htpasswd/oval/shared.xml
matejak: So now, the rule is not ocp4-only, and as the choice is between incomplete content and content with unofficial extensions, let's go for the latter.

matejak:
> These errors are consistent with what I'm seeing when running this PR with the prototype resource collector:

As of recently, the upstream version works out of the box. Previous upstream versions required the `--skip-valid` argument to perform scans.

matejak:
Are there any outstanding issues, preventing this PR to be merged?

CMakeLists.txt
```diff
@@ -94,6 +94,8 @@ option(SSG_PRODUCT_WRLINUX1019 "If enabled, the WRLinux1019 SCAP content will be
 option(SSG_CENTOS_DERIVATIVES_ENABLED "If enabled, CentOS derivative content will be built from the RHEL content" TRUE)
 option(SSG_SCIENTIFIC_LINUX_DERIVATIVES_ENABLED "If enabled, Scientific Linux derivative content will be built from the RHEL content" TRUE)
 
+option(VALIDATE_EVERYTHING "If enabled, include all validation tests regardless of oscap version requirements" FALSE)
```
yuumasato, line 97: I got confused by the help text and thought that no content was validated by default until I read code of `function(define_validate_product PRODUCT)` or commit message.

I suggest to updated the help text to clarify that validation of certain content is dependent on the scanner version, and include OVERRIDE or FORCE in the variable name.

cmake/SSGCommon.cmake
```diff
@@ -60,6 +60,17 @@ endif()
 set(SSG_HTML_GUIDE_FILE_LIST "")
 set(SSG_HTML_TABLE_FILE_LIST "")
 
+# Define validate_product FALSE for appropriate combinations
```
yuumasato, line 63: ` # Define validate_product FALSE for inappropriate combinations` ?

cmake/SSGCommon.cmake
```diff
@@ -60,6 +60,17 @@ endif()
...
+# of products and oscap versions.
+function(define_validate_product PRODUCT)
+    set(validate_product, TRUE)
+    if ("${OSCAP_VERSION}" VERSION_LESS "1.3.4")
+        if ("${PRODUCT}" MATCHES "^(ocp4|)$")
```
yuumasato, line 68: Is there a use case where `PRODUCT` is empty?

cmake/SSGCommon.cmake
```diff
@@ -60,6 +60,17 @@ endif()
 set(SSG_HTML_GUIDE_FILE_LIST "")
 set(SSG_HTML_TABLE_FILE_LIST "")
 
+# Define validate_product FALSE for appropriate combinations
+# of products and oscap versions.
+function(define_validate_product PRODUCT)
+    set(validate_product, TRUE)
```
yuumasato, line 66: As `validate_product` is a variable it should be all capital letters.

cmake/SSGCommon.cmake
matejak: Not really, it was an indication how to expand list of products easily.

yuumasato:
@JAORMX @mrogers950 @jhrozek This PR will make `ci/prow/e2e-aws-moderate` always fail until openscap 1.3.3 is used.
Are you guys ok merging this?

mrogers950:
@yuumasato this looks good to me now.

jhrozek:
> @yuumasato this looks good to me now, if @JAORMX / @jhrozek agree then let's merge

I'm sorry for the late review. While I don't pretend I understand all in this patchset (I only skimmed through the build system changes), the rule and the mechanism to populate the data so that the rule can actually run look good to me!

JAORMX:
Could we not add this to the moderate profile yet? It makes it crash cause we still don't have a new enough scanner in the operator. Let's merge this without adding the check to the profile so you can move forward, and subsequently let's work on getting a newer scanner on the operator so we could finally take this into use.

JAORMX:
/retest
@jhrozek updated the image to use a newer scanner

JAORMX:
So... where do we stand on adding this as metadata? Seems that now the implementation is using a `code` tag in a warning... which might not be very intuitive tbh

linux_os/guide/system/ocp_idp_no_htpasswd/oval/shared.xml
JAORMX: @redhatrises responding to your concerns:

> Metadata SHOULD NOT replace the functionality of existing fields within XCCDF. If the information matches the common use of some other XCCDF field, the information MUST appear in that XCCDF field. It MAY also appear in the `<xccdf:metadata>` element.

It doesn't replace functionality since a setting such as this one is missing from the spec. On the other hand, is clearly is data about the object that will be parsed, as it represents the location where the object can be found.

> Metadata SHALL NOT change the processing model as outlined in Section 7. Benchmark consumer products SHALL perform XCCDF assessments in the same way regardless of the presence of metadata. Metadata MAY still be used to support assessment features that are outside the purview of XCCDF—for example, to hold post-processing instructions to be performed on the completed XCCDF results or to display additional information to the user before or during the assessment.

It does not change the processing model (Tailoring, Document Generation or Assessing), since you can still put the file in the right place by yourself (with or without the metadata), it however supports assessment features (in this case it's pre-processing of the instructions in order to fetcht he objects and put them in the right place).

> Metadata SHOULD NOT alter the character content of the XCCDF properties used to generate output during document generation.Contents of the `<xccdf:metadata>` element MAY be added to the output above and beyond the XCCDF properties. Metadata MAY contain instructions that cause different stylistic conventions to be adopted in the conversion of an XCCDF document to prose.

While the URL can be displayed.. it doesn't need to.

applications/openshift/ocp_idp_no_htpasswd/rule.yml
```diff
@@ -0,0 +1,16 @@
...
+rationale: TBD
+
+identifiers:
+    cce@ocp4: 84209-6
+
+severity: medium
+
+warnings:
+    - general: |-
+        {{{ openshift_cluster_setting("/apis/config.openshift.io/v1/oauths/cluster", "idp.yml") | indent(8) }}}
```
JAORMX, line 16: Why wasn't the path used as a file name instead of having to specify a filename with every rule?

e.g. the file could have just been `/apis/config.openshift.io/v1/oauths/cluster` instead of having to specify `idp.yml`. The parent directory creation would be handled by the operator.

applications/openshift/ocp_idp_no_htpasswd/oval/shared.xml
```diff
@@ -0,0 +1,33 @@
...
+    </ind:yamlfilecontent_test>
+
+    <ind:yamlfilecontent_object id="object_{{{ rule_id }}}" version="1">
+      <ind:path var_ref="ocp_data_root"/>
+      <ind:filename operation="pattern match">^idp.yml$</ind:filename>
+      <ind:yamlpath>.spec.identityProviders[:].type</ind:yamlpath>
+    </ind:yamlfilecontent_object>
+
+    <ind:yamlfilecontent_state id="state_{{{ rule_id }}}" version="1">
+	    <ind:value_of entity_check="at least one">htpasswd</ind:value_of>
```
mrogers950, line 28: I didn't realize this is case-sensitive, so the check we need here is for "HTPasswd".

applications/openshift/ocp_idp_no_htpasswd/rule.yml
mrogers950: this would work, it actually might be better at preventing duplicate file names in rules

matejak: So just a maybe stupid check whether I understood you correctly - is it like that you would be happier to save (and scan) the file to `$ROOT/apis/config.openshift.io/v1/oauths/cluster` instead to `$ROOT/idp.yml`?

mrogers950: @matejak that's correct.

applications/openshift/ocp_idp_no_htpasswd/oval/shared.xml
matejak: done

matejak:
Suggestions have been addressed. For the time being, we have to abuse the XCCDF warning as metadata, as #5552 revealed that support of XCCDF 1.2 metadata in our content would require huge refactoring on the build system side.

mrogers950:
@matejak Thanks for the update, everything works well for me.

mrogers950:
@matejak please see https://github.com/matejak/scap-security-guide/pull/4 , merging this in should fix the e2e tests.

JAORMX:
I need to cut a release for the compliance-operator before we can merge this. Else the openshift CI jobs will start failing.

JAORMX:
Just released, giving this a try

JAORMX:
we did changes to the API. This is needed: https://github.com/ComplianceAsCode/content/pull/5774

JAORMX:
This needs a rebase.

mildas:
**Changes identified:**
Rule ocp_idp_no_htpasswd:
 OVAL check is newly added.
Profile platform-moderate on ocp4:
 Newly added profile.
Macro ocil_clause_entry_sshd_option:
 
Macro openshift_cluster_setting:
 
Others:
 Python abstract syntax tree change found in ssg/oval.py.

 (cd build && ctest -j4)
 build_product ocp4
 test_suite.py rule --libvirt qemu:///system test-suite-vm --remediate bash --datastream build/ssg-ocp4-ds.xml ocp_idp_no_htpasswd
 test_suite.py profile --libvirt qemu:///system test-suite-vm --datastream build/ssg-ocp4-ds.xml platform-moderate

JAORMX:
nice!

/lgtm

PR #5328, opened 2020-03-23
Restrict kernel_module and service_rsyncd_disabled rules as machine-only

mildas:
#### Description:
Restrict rules which are not applicable for containers as machine-only.

#### Rationale:
Containers use host OS kernel thus `kernel_module_*` rules don't make sense to be checked in them.

`service_rsyncd_disabled` is a newly added rule where probably was forgotten machine-only specification.


jan-cerny:
@openscap-ci test this please

jan-cerny:
@openscap-ci test this please

jan-cerny:
@openscap-ci test this please

jan-cerny:
Thank you, nice catch!

PR #5330, opened 2020-03-23
Extend timeout for VM restarts

mildas:
#### Description:
Extent timeout to 5min.

#### Rationale:
The previous timeout wasn't enough for slow machines and tests are failing because of that.


jan-cerny:
@openscap-ci test this please

PR #5338, opened 2020-03-24
ocp4: Add go dependencies to test directory

JAORMX:
CI expects to find the dependencies already. So we need to include them
here... unfortunately.


yuumasato:
Holy, that is a lot of files.

I'm not familiar with how go works.
Is there some way to keep them aligned with what the operator needs?
Can all this files be exchanged by a script that fetches them?

PR #5339, opened 2020-03-24
add rules for checking legacy "+" entries in passwd related files

vojtapolasek:
#### Description:

Added three rules which do basically the same thing, but for different files. They  ensure that there are no legacy entries in:
- /etc/passwd
- /etc/group
- /etc/shadow

I also added them to rhel7 and rhel8 CIS profiles.
Unfortunately due to the nature of check it was not possible to utilize macros.

#### Rationale:

CIS effort


linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_legacy_plus_entries_shadow/rule.yml
```diff
@@ -0,0 +1,31 @@
...
+rationale: |-
+    Using this method to include entries into <tt>/etc/shadow</tt> is considered legacy
+    and should be avoided. These entries may provide a way for an attacker
+    to gain access to the system.
+
+severity: medium
+
+identifiers:
+    cce@rhel7: 83390-5
+    cce@rhel8: 83391-3
```
yuumasato: CEE `83391-3` is not present in the CCE reference file, it is probably already taken by another rule.

vojtapolasek:
Good point, I renamed the rules and fixed the cce problem.

PR #5346, opened 2020-03-25
Various macro and description fixes

redhatrises:
#### Description:

- Unused macros shouldn't print error messages in content as macros do not have to be used
- Remove unnecessary operator usages
- Use compliance type language in description
- Strip regex from string values in description

### DIFF
--- a/shared/macros-highlevel.jinja
+++ b/shared/macros-highlevel.jinja
@@ -1,4 +1,5 @@
 {{% macro ocil_package(package) -%}}
+  {{% if pkg_system is defined %}}
     {{%- if pkg_system == "rpm" -%}}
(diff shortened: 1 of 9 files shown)


shared/macros-highlevel.jinja
```diff
@@ -1,15 +1,18 @@
 {{% macro ocil_package(package) -%}}
+  {{% if pkg_system is defined %}}
```
yuumasato: I might know the answer, but could you clarify for which products `pkg_system` would be undefined?

redhatrises: Any product that doesn't have one. Basically, macros should not be required to be used in the content.

jan-cerny:
Thank you

PR #5349, opened 2020-03-26
Update rhel7 CIS selections

yuumasato:
#### Description:

- Update CIS rule selection
- Add comment on requirements that don't need to fulfilled.


vojtapolasek:
LGTM

PR #5350, opened 2020-03-27
Revert "changed missing parameter pass from true to false per stig 40440"

ggbecker:
Reverts ComplianceAsCode/content#5303

Similar to ComplianceAsCode/content#5327

KerberosAuthentication is disable by default so there is no need to explicitly check for it. This represents a bug in the DISA content, not ComplianceAsCode, and should be reverted.


redhatrises:
@openscap-ci test this please

PR #5353, opened 2020-03-27
fix ansible remediation for postfix_network_listening_disabled

vojtapolasek:
#### Description:

Ansible remediation checks for presence of postfix package and proceeds only if it is installed.

#### Rationale:

This rule is in CIS but CIS does not mandate installation of postfix.


vojtapolasek:

PR #5354, opened 2020-03-27
ocp4: e2e test continuation

JAORMX:
This is a continuation of the e2e test suite for ocp4:

* Enables auto-applying remediations
* Waits until all the nodes are ready (they get restarted by the
  machine-config-operator)
* Does a second scan
* Checks that the remediations have diminished


JAORMX:
we're gonna need a release for this to pass. I'll arrange that on the compliance-operator side.

JAORMX:
@jhrozek yep... The remediations mess up the nodes on CI as well

PR #5547, opened 2020-03-31
fix rule disabling ipv6 through grub2

vojtapolasek:
#### Description:

Modify the oval template of grub2_argument to use rule ID as an oval definition ID.

While processing data for grub2_argument template, replace . with \. in case of oval.

Rename grub2_disable_ipv6 to grub2_ipv6_disable_argument

add tests to that rule

#### Rationale:

This solves two problems:

1. dot has to be escaped within regular expressions used in oval

2. the default naming of templated rules can't be applied here because the rule ID would contain dot. This is not acceptable.


mildas:
retest this please

mildas:
retest this please

redhatrises:
> I think a specific variable should be added that is explicitly escaped, `ARG_NAME_VALUE` is used in multiple text descriptions and comments, e.g.:
> 
> ```
>             <ns3:description>Ensure ipv6\.disable=1 is configured in the kernel line in /etc/default/grub.</ns3:description>
> ...
>             <ns3:criterion comment="Check if ipv6\.disable=1 is present in the GRUB2 environment variable block in /boot/grub2/grubenv" test_ref="oval:ssg-test_grub2_ipv6.disable_argument_grub_env:tst:1"/>
> ```
> 
> Although allowed, I particularly don' like that IDs of tests, objects and states would have a dot in it.
> 
> ```
>           <ns6:object object_ref="oval:ssg-object_grub2_ipv6.disable_argument_grub_env:obj:1"/>
>           <ns6:state state_ref="oval:ssg-state_grub2_ipv6.disable_argument:ste:1"/>
> ```

Agree with @yuumasato . Let's be consistent and not have `.` in IDs.

vojtapolasek:
Yes that makes sense. I created two new variables:
escaped_arg_name_value which is used only in the test state
sanitized_arg_name which is used in names of tests, objects, states

PR #5550, opened 2020-04-01
Added possibility of the test suite to expand platforms of the benchmark

matejak:
TLDR: Allow datastream of platform A to be tested using our test suite on a system (container, VM) of platform B != A.

This PR features these changes/improvements:

- Put the datastream into a temp dir upon tests initialization, so it can be modified, and one can do rebuilds while tests run.
- Test suite now can modify the datastream prior to the testing.
- Remove machine-only platform from rules/groups in the datastream unless it is explicitly requested to be preserved. This allows testing tests of all textfile-based rules in containers (e.g. sshd config, audit config etc.)
- If a CPE regex is supplied, skim the datastream for matching CPEs, and expand the benchmark platform list with them. If there is no matching CPE, treat the supplied argument as a CPE and extend the benchmark accordingly.
- Become able to take tests from other benchmark directories, as platform is not an obstacle any more.

Notes:

- Our test suite won't be able to handle yaml tests, as invalid datastreams are a major obstacle for e.g. `oscap-ssh`.
- All RHEL CPEs are present literally in the datastream, but for Fedora, there is a parametrized CPE. So if you would like to scan e.g. RHEL6 content on RHEL7 VM, you can pass `--add-platform enterprise_linux`, but for Fedora, you have to use e.g. `--add-platform 'cpe:/o:fedoraproject:fedora:30'`. If you pass `--add-platform fedora`, only Fedora 28 CPE will be added, which won't work other Fedoras than 28.

Special thanks to @yuumasato and @jan-cerny that shared their wisdom, so this could be hacked together.

### DIFF
--- a/tests/ssg_test_suite/common.py
+++ b/tests/ssg_test_suite/common.py
@@ -21,7 +21,11 @@ Scenario_conditions = namedtuple(
 Rule = namedtuple(
     "Rule", ["directory", "id", "short_id", "files"])
 
-_BENCHMARK_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), '../../linux_os/guide'))
+_BENCHMARK_DIRS = [
+        os.path.abspath(os.path.join(os.path.dirname(__file__), '../../linux_os/guide')),
+        os.path.abspath(os.path.join(os.path.dirname(__file__), '../../applications')),
+        ]
+
 _SHARED_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), '../shared'))
 
 REMOTE_USER = "root"
@@ -38,6 +42,12 @@ SSH_ADDITIONAL_OPTS = (
     "-o", "StrictHostKeyChecking=no",
     "-o", "UserKnownHostsFile=/dev/null",
 ) + SSH_ADDITIONAL_OPTS
+
+
+def walk_through_benchmark_dirs():
+    for dirname in _BENCHMARK_DIRS:
+        for dirpath, dirnames, filenames in os.walk(dirname):
+            yield dirpath, dirnames, filenames
 
 
 class Stage(object):
@@ -221,7 +231,7 @@ def create_tarball():
             "wb", suffix=".tar.gz", delete=False) as fp:
         with tarfile.TarFile.open(fileobj=fp, mode="w") as tarball:
             tarball.add(_SHARED_DIR, arcname="shared", filter=_make_file_root_owned)
-            for dirpath, dirnames, _ in os.walk(_BENCHMARK_DIR):
+            for dirpath, dirnames, _ in walk_through_benchmark_dirs():
                 rule_id = os.path.basename(dirpath)
                 if "tests" in dirnames:
                     tests_dir_path = os.path.join(dirpath, "tests")
@@ -284,7 +294,7 @@ def iterate_over_rules():
                         containing the test scenarios in Bash
             files -- list of executable .sh files in the "tests" directory
     """
-    for dirpath, dirnames, filenames in os.walk(_BENCHMARK_DIR):
+    for dirpath, dirnames, filenames in walk_through_benchmark_dirs():
         if "rule.yml" in filenames and "tests" in dirnames:
             short_rule_id = os.path.basename(dirpath)
             tests_dir = os.path.join(dirpath, "tests")
--- a/tests/ssg_test_suite/xml_operations.py
+++ b/tests/ssg_test_suite/xml_operations.py
@@ -4,6 +4,7 @@ import xml.etree.cElementTree as ET
 import logging
 import contextlib
 import re
+import subprocess
 
 from ssg.constants import OSCAP_RULE
 from ssg.constants import PREFIX_TO_NS
@@ -92,16 +93,31 @@ def remove_machine_only_from_element(root, element_spec):
                 el.remove(p)
 
 
-def get_all_cpe_refs(root):
-    query = (
-            ".//ds:component/oval-def:oval_definitions/oval-def:definitions"
-            "/oval-def:definition[@class='inventory']/oval-def:metadata"
-            "/oval-def:reference[@source='CPE']")
-    references = root.findall(query, PREFIX_TO_NS)
-    cpes = set()
-    for ref in references:
-        cpes.add(ref.get("ref_id"))
-    return cpes
+def get_oscap_supported_cpes():
+    """
+    Obtain a list of CPEs that the scanner supports
+    """
+    result = []
+    proc = subprocess.Popen(
+            ("oscap", "--version"), text=True,
+            stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    try:
+        outs, errs = proc.communicate(timeout=3)
+    except subprocess.TimeoutExpired:
+        logging.warn("Scanner timeouted when asked about supported CPEs")
+        proc.kill()
+        return []
+
+    if proc.returncode != 0:
+        first_error_line = errs.split("\n")[0]
+        logging.warn("Error getting CPEs from the scanner: {msg}".format(msg=first_error_line))
+
+    cpe_regex = re.compile(r'\bcpe:\S+$')
+    for line in outs.split("\n"):
+        match = cpe_regex.search(line)
+        if match:
+            result.append(match.group(0))
+    return result
 
 
 def add_platform_to_benchmark(root, cpe_regex):
@@ -113,7 +129,7 @@ def add_platform_to_benchmark(root, cpe_regex):
         )
         raise RuntimeError(msg)
 
-    all_cpes = get_all_cpe_refs(root)
+    all_cpes = get_oscap_supported_cpes()
     regex = re.compile(cpe_regex)
 
     cpes_to_add = []
--- a/tests/ssg_test_suite/xml_operations.py
+++ b/tests/ssg_test_suite/xml_operations.py
@@ -2,12 +2,14 @@
 import xml.etree.cElementTree as ET
 
 import logging
+import contextlib
+import re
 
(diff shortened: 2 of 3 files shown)


tests/ssg_test_suite/common.py
```diff
@@ -21,7 +21,11 @@
 Rule = namedtuple(
     "Rule", ["directory", "id", "short_id", "files"])
 
-_BENCHMARK_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), '../../linux_os/guide'))
+_BENCHMARK_DIRS = [
+        os.path.abspath(os.path.join(os.path.dirname(__file__), '../../linux_os/guide')),
+        os.path.abspath(os.path.join(os.path.dirname(__file__), '../../applications')),
+        ]
+
```
yuumasato, line 28: Are changes on this file intended to be here? They are not mentioned in the PR description.
It adds possibility to run test scenarios for rules in `applications` benchmark.

tests/ssg_test_suite/xml_operations.py
```diff
@@ -71,6 +74,71 @@ def infer_benchmark_id_from_component_ref_id(datastream, ref_id):
...
+    elements = root.findall(query, NAMESPACES)
+    for el in elements:
+        platforms = el.findall("./xccdf:platform", NAMESPACES)
+        for p in platforms:
+            if p.get("idref") == "cpe:/a:machine":
+                el.remove(p)
+
+
+def get_all_cpe_refs(root):
+    query = ".//ds:component/oval:oval_definitions/oval:definitions/oval:definition[@class='inventory']/oval:metadata/oval:reference[@source='CPE']"
```
yuumasato, line 106: This is actually querying for the checks used by CPE items, not the CPE items themselves.

tests/ssg_test_suite/common.py
matejak: Yes, I have updated the description.

matejak:
> I would actually expect the Test Suite to remove machine platform when requested, not by default.

Plenty of rules that are not applicable to containers in production can actually be tested in containers. This behavior is driven by assumption that if I want to test a rule using the container backend, I know what I am doing and the machine CPE just stands in my way.

tests/ssg_test_suite/xml_operations.py
matejak: I am a bit lost, what would be the right thing to do here?

yuumasato:
> Plenty of rules that are not applicable to containers in production can actually be tested in containers.

I understand it.
But won't it be confusing if a datastream behaves one way when ran via Test Suite and another when via scanner?

> This behavior is driven by assumption that if I want to test a rule using the container backend, I know what I am doing and the machine CPE just stands in my way.

Is it safe to assume that everybody knows what they are doing?

My thinking is that the Test Suite should evaluate the tested content and target system as close to default as possible.
And people who know what they are doing can use these kinds of options.

tests/ssg_test_suite/xml_operations.py
yuumasato: Apologies, the code itself is not wrong, but the naming is a bit off. But I may just be nitpicking.

But, I honestly wonder how it is working, because the DS doesn't contain `cpe-item` for other OS versions.
Maybe it is falling back to use the `cpe-dict` from the scanner?
And in this case, all this logic of `add_platform_to_benchmark` wouldn't be necessary...

yuumasato: And the OpenSCAP's CPE Dictionary contains rhel7, and other Fedoras:
https://github.com/OpenSCAP/openscap/blob/maint-1.3/cpe/openscap-cpe-dict.xml#L16

matejak: I need to add something to the benchmark platform. Right now, I get it from the datastream, but what I hear is that I may get the platform CPE string from the `oscap --version` output, i.e. I don't have to search the datastream, hoping to find occurrences of CPE IDs. Do I understand that right?

matejak:
> My thinking is that the Test Suite should evaluate the tested content and target system as close to default as possible.
And people who know what they are doing can use these kinds of options.

I see this dillema - should the test suite eagerly test everything that is testable, or should it be conservative by default, so nobody is misled by having an sshd test executed in a container with a non-notaplicable result? Let's ask the QA expert @matusmarhefka - if a QE engineer has a strong opinion about this, that would be the way to follow.

tests/ssg_test_suite/xml_operations.py
yuumasato: @matejak That is my suspicion.

Because the `platform` entry added doesn't have a counterpart `cpe-item` in the DS. 
So maybe openscap is falling back to its own CPE dictionary, and there it finds the `cpe-item` entry mathing the `platform` added.

matusmarhefka:
> > My thinking is that the Test Suite should evaluate the tested content and target system as close to default as possible.
> > And people who know what they are doing can use these kinds of options.
> 
> I see this dillema - should the test suite eagerly test everything that is testable, or should it be conservative by default, so nobody is misled by having an sshd test executed in a container with a non-notaplicable result? Let's ask the QA expert @matusmarhefka - if a QE engineer has a strong opinion about this, that would be the way to follow.

TBH I am more inclined to the original behavior of the test suite. I think it is more transparent to have a separate option to request removing machine CPEs from the content than doing it "auto-magically", I find it confusing. Removing machine CPEs might cause that unaware users might encounter strange issues when using container backed for testing. I also prefer to test rules in an environment for which they are meant, testing a machine-only rule in a container might produce different results than testing it in a full-fledged bare metal machine or a VM. Also testing in `ubi` vs `ubi-init` container with systemd might produce different results.

So I wouldn't remove machine CPEs by default, rather through a separate option. And maybe it would be better to reevaluate rules which are machine-only. If we think some rules can be tested in containers maybe they can be also evaluated/remediated in containers.

matejak:
Thanks for your feedback, it has been addressed in the last commit.

jan-cerny:
I also prefer to remove and modify CPEs only on request, not by default.

tests/ssg_test_suite/xml_operations.py
```diff
@@ -71,10 +65,93 @@ def infer_benchmark_id_from_component_ref_id(datastream, ref_id):
...
+    cpes_to_add = []
+    for cpe_str in all_cpes:
+        if regex.search(cpe_str):
+            cpes_to_add.append(cpe_str)
+
+    if not cpes_to_add:
+        cpes_to_add = [cpe_regex]
+
+    for benchmark in benchmarks:
+        existing_platform_element = benchmark.find("xccdf-1.2:platform", PREFIX_TO_NS)
```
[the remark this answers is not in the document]
matejak: Good catch. As the platform needs to come before some sub-elements and after another ones, no action will be taken.

tests/ssg_test_suite/xml_operations.py
```diff
@@ -71,10 +65,93 @@ def infer_benchmark_id_from_component_ref_id(datastream, ref_id):
...
+def add_platform_to_benchmark(root, cpe_regex):
+    benchmark_query = ".//ds:component/xccdf-1.2:Benchmark"
+    benchmarks = root.findall(benchmark_query, PREFIX_TO_NS)
+    if not benchmarks:
+        msg = (
+            "No benchmarks found in the datastream"
+        )
+        raise RuntimeError(msg)
+
+    all_cpes = get_oscap_supported_cpes()
```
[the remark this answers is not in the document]
matejak: You are right with the host/guest confusion, but I think that this is acceptable as part of the testing convenience enhancement.

tests/test_suite.py
```diff
@@ -63,6 +65,21 @@ def parse_args():
...
                                help="Selection number of reference ID related "
                                     "to benchmark to be used.")
+    common_parser.add_argument(
+            "--add-platform",
+            metavar="<CPE REGEX>",
+            default=None,
+            help="Find all CPEs in the datastream that match the provided regex, "
+            "and add them as platforms to all datastream benchmarks. "
+            "If the regex doesn't match anything, it will be treated "
+            "as a literal CPE, and added as a platform. "
```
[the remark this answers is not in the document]
matejak: I have updated the description.

tests/ssg_test_suite/xml_operations.py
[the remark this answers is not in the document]
jan-cerny: Can it be obtained from the definitions in SSG repository?

jan-cerny:
@matejak I have problems with using it. Can you take a look at this?

```
[jcerny@thinkpad tests{pr/5550}]$ python3 test_suite.py rule --libvirt qemu:///system ssgts_rhel8 --datastream /usr/share/xml/scap/ssg/content/ssg-fedora-ds.xml --add-platform enterprise_linux selinux_state                                                                               
INFO - Logging into /home/jcerny/work/git/scap-security-guide/tests/logs/rule-custom-2020-04-29-0919/test_suite.log                                                                           
WARNING - Couldn't find platform element in a benchmark, not adding any additional platforms as a result.                                                                                     
INFO - xccdf_org.ssgproject.content_rule_selinux_state                                         
ERROR - Script selinux_enforcing.pass.sh using profile xccdf_org.ssgproject.content_profile_ospp found issue:                                                                                 
ERROR - Rule evaluation resulted in notapplicable, instead of expected pass during initial stage                                                                                              
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_selinux_state'. 
```

The datastream that I used comes from scap-security-guide-0.1.49-1.fc31.noarch.

matejak:
Who would have guessed that an element returned by `find` may have a "falsey" value. I have fixed that by amending the last commit.

jan-cerny:
Thank you, I confirm the latest update fixes my problem.

PR #5552, opened 2020-04-02
[WIP] Add rule metadata

matejak:
Added rule metadata support to `rule.yml`.

Rule metadata are supported by the XCCDF standard (see the [XCCDF spec](https://csrc.nist.gov/CSRC/media/Publications/nistir/7275/rev-4/final/documents/nistir-7275r4_updated-march-2012_markup.pdf), page number 10:

> When used, the `<xccdf:metadata>` element SHALL contain document metadata expressed in XML. The metadata element can appear one or more times as a child of the `<xccdf:Benchmark>`, `<xccdf:Rule>`, ...

Judging by [the schema](https://github.com/OpenSCAP/openscap/blob/maint-1.3/schemas/xccdf/1.2/xccdf_1.2.xsd#L511), it looks like that it is expected that the metadata tag has at least one XML children of a non-xccdf namespace, so I have proposed to specify them as YAML dictionaries. FYI, the XCCDF standard suggest usage of the DCMI metadata schema for the metadata itself, but that's completely optional.

The PR has currently these problems:

1. The PR is stuck on the scanner refusing to compile the datastream.
2. I have used the "xml" namespace as the namespace for metadata sub-elements, which is a temp workaround.
3. It may be that rule's metadata element is supported in XCCDF 1.2, but not in XCCDF 1.1. If it turns out to be the case, what does this imply?

### DIFF
--- a/ssg/build_yaml.py
+++ b/ssg/build_yaml.py
@@ -25,7 +25,7 @@ from .xml import ElementTree as ET
 from .shims import unicode_func
 
 
-def add_sub_element(parent, tag, data):
+def add_sub_element(parent, tag, data=''):
     """
     Creates a new child element under parent with tag tag, and sets
     data as the content under the tag. In particular, data is a string
@@ -781,6 +781,7 @@ class Rule(object):
         "conflicts": lambda: list(),
         "requires": lambda: list(),
         "platform": lambda: None,
+        "metadata": lambda: dict(),
         "template": lambda: None,
     }
 
@@ -800,6 +801,7 @@ class Rule(object):
         self.requires = []
         self.conflicts = []
         self.platform = None
+        self.metadata = dict()
         self.template = None
 
     @classmethod
@@ -1022,6 +1024,12 @@ class Rule(object):
         add_sub_element(rule, 'title', self.title)
         add_sub_element(rule, 'description', self.description)
         add_sub_element(rule, 'rationale', self.rationale)
+
+        if self.metadata:
+            meta = add_sub_element(rule, 'metadata')
+            for meta_tag, meta_text in self.metadata.items():
+                meta_tag_with_ns = "xml:{tag}".format(tag=meta_tag)
+                add_sub_element(meta, meta_tag_with_ns, str(meta_text))
 
         main_ident = ET.Element('ident')
         for ident_type, ident_val in self.identifiers.items():
--- a/linux_os/guide/services/ssh/ssh_server/sshd_disable_root_login/rule.yml
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_disable_root_login/rule.yml
@@ -52,6 +52,9 @@ references:
 
 {{{ complete_ocil_entry_sshd_option(default="no", option="PermitRootLogin", value="no") }}}
 
+metadata:
+    meta_cathegory: SAMPLE_METADATA
+
 template:
     name: sshd_lineinfile
     vars:
--- a/shared/transforms/shared_shorthand2xccdf.xslt
+++ b/shared/transforms/shared_shorthand2xccdf.xslt
@@ -73,6 +73,7 @@
           <xsl:apply-templates select="description[contains(@prodtype, $prod_type) or not(@prodtype)]"/>
           <xsl:apply-templates select="warning[contains(@prodtype, $prod_type) or not(@prodtype)]"/>
           <xsl:apply-templates select="ref[contains(@prodtype, $prod_type) or not(@prodtype)]"/>
+          <xsl:apply-templates select="metadata[contains(@prodtype, $prod_type) or not(@prodtype)]"/>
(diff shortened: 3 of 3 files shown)


matejak:
> This makes the ocp4 build fail:

Yes, it doesn't build at all for any product that contains the rule that features the metadata content. We have to figure out what to do to make it passing.

yuumasato:
> The PR is stuck on the scanner refusing to compile the datastream.

I see that it is trying to validate against XCCDF 1.1 schema.
```
OpenSCAP Error: Invalid XCCDF Checklist (1.1) content in /home/jenkins/workspace/scap-security-guide-pull-requests/label/rhel8/build/debian8/xccdf-unlinked-resolved.xml. [/builddir/build/BUILD/openscap-1.3.1/src/source/oscap_source.c:346]
```

I think the issue is that the build system doesn't generate XCCDF 1.2 right out of the bat.

It first generates an XCCDF 1.1, then during [ssg_build_xccdf_final](https://github.com/ComplianceAsCode/content/blob/master/cmake/SSGCommon.cmake#L410) it converts the XCCDF to [version 1.2](https://github.com/ComplianceAsCode/content/blob/master/cmake/SSGCommon.cmake#L458), using an [XSL from OpenSCAP](https://github.com/ComplianceAsCode/content/blob/master/cmake/FindOpenSCAP.cmake#L16.
)

I'm not aware of anyone using XCCDF 1.1.
And I don't if `oscap` can generate  XCCDF 1.2 directly.

ssg/build_yaml.py
```diff
@@ -1004,6 +1006,12 @@ def to_xml_element(self):
         add_sub_element(rule, 'description', self.description)
         add_sub_element(rule, 'rationale', self.rationale)
 
+        if self.metadata:
+            meta = add_sub_element(rule, 'metadata')
+            for meta_tag, meta_text in self.metadata.items():
+                meta_tag_with_ns = "xml:{tag}".format(tag=meta_tag)
```
redhatrises, line 1012: Is there a way to not hard-code the namespace here? Will need dublin-core elements here as well.

redhatrises:
> > The PR is stuck on the scanner refusing to compile the datastream.
> 
> I see that it is trying to validate against XCCDF 1.1 schema.
> 
> ```
> OpenSCAP Error: Invalid XCCDF Checklist (1.1) content in /home/jenkins/workspace/scap-security-guide-pull-requests/label/rhel8/build/debian8/xccdf-unlinked-resolved.xml. [/builddir/build/BUILD/openscap-1.3.1/src/source/oscap_source.c:346]
> ```
> 
> I think the issue is that the build system doesn't generate XCCDF 1.2 right out of the bat.
> 
> It first generates an XCCDF 1.1, then during [ssg_build_xccdf_final](https://github.com/ComplianceAsCode/content/blob/master/cmake/SSGCommon.cmake#L410) it converts the XCCDF to [version 1.2](https://github.com/ComplianceAsCode/content/blob/master/cmake/SSGCommon.cmake#L458), using an [XSL from OpenSCAP](https://github.com/ComplianceAsCode/content/blob/master/cmake/FindOpenSCAP.cmake#L16.)
> 
> I'm not aware of anyone using XCCDF 1.1.
> And I don't if `oscap` can generate XCCDF 1.2 directly.

Great catch by both Matej and Yuma! There was some sort of reason for this way back in the day. Don't remember what. Possibly because the 1.2 version wasn't finalized yet. It would be nice to rectify oscap not being able to generate 1.2 and 1.3 content.

jan-cerny:
Great finding!

I think that to fix it you will have to start creating XCCDF 1.2 instead of 1.1 from the beginning of the build process. Our final product is datastream, which uses XCCDF 1.2.

I haven't looked into how to do that yet. Probably you will have to change our build system and XSLTs.

The main difference seems to be the ID format: short in 1.1 and long in 1.2.


> I'm not aware of anyone using XCCDF 1.1.

Downstream distributions ship it. For example, RHEL8: scap-security-guide-0.1.49-1.el8.noarch contains XCCDF files in 1.1 namespace. Perhaps the distributions can start shipping XCCDF 1.2 files instead.

> And I don't if oscap can generate XCCDF 1.2 directly.

I think oscap isn't important here, it's more about the scap-security-guide build system. We need to investigate if our templates and build scripts can generate XCCDF 1.2 directly. There is the `oscap xccdf resolve` step, but that seems to use the same code for all XCCDF versions. Later, the datastream is composed from the XCCDF 1.2 file, therefore that is already supported in oscap.

jan-cerny:
If we will still need to produce the XCCDF 1.1 files we will probably need to have a XSLT to convert from 1.2 to 1.1.

mildas:
**Changes identified:**
Others:

redhatrises:
Closing this hasn't been worked on in a long while.

PR #5553, opened 2020-04-02
Make the datastream namespaces readable again

matejak:
Build your favourite product and examine the built datastream - there should be no occurrences of `ns0`, `ns1` and other namespaces - there should be human-readable identifiers instead.

What this PR does:

- Consolidates known XML namespaces in a dictionary.
- Registers those namespaces with the XML parser, if it supports that.
- Makes sure that all build scripts use functions that register those namespaces.

### DIFF
--- a/ssg/constants.py
+++ b/ssg/constants.py
@@ -92,6 +92,7 @@ ALL_NS = {
     "xccdf-1.2": XCCDF12_NS,
     "xlink": xlink_namespace,
     "cpe-dict": "http://cpe.mitre.org/dictionary/2.0",
+    "cat": "urn:oasis:names:tc:entity:xmlns:xml:catalog",
 }
 
 
--- a/ssg/constants.py
+++ b/ssg/constants.py
@@ -83,7 +83,7 @@ OVAL_SUB_NS = dict(
 )
 
 
-ALL_NS = {
+PREFIX_TO_NS = {
     "oval-def": oval_namespace,
     "oval": "http://oval.mitre.org/XMLSchema/oval-common-5",
     "ds": datastream_namespace,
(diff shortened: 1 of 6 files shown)


ssg/constants.py
```diff
@@ -75,6 +75,32 @@
     "Verify Ansible meets SCAP-Security-Guide version requirements."
 standard_profiles = ['standard', 'pci-dss', 'desktop', 'server']
 
+
+OVAL_SUB_NS = dict(
```
redhatrises, line 79: Does this mean that we are going to have to manually handle the OVAL sub namespaces? If so, can this be automated to would mean it is one less thing to worry about in having to update or maintain.

ssg/constants.py
```diff
@@ -75,6 +75,32 @@
...
+
+ALL_NS = {
+    "oval-def": oval_namespace,
+    "oval": "http://oval.mitre.org/XMLSchema/oval-common-5",
+    "ds": datastream_namespace,
+    "ocil": ocil_namespace,
+    "xccdf-1.1": XCCDF11_NS,
+    "xccdf-1.2": XCCDF12_NS,
+    "xlink": xlink_namespace,
+    "cpe-dict": "http://cpe.mitre.org/dictionary/2.0",
```
yuumasato, line 94: `cat_namespace` should be added here too.

ssg/constants.py
matejak: This PR doesn't introduce any new obligation - it just takes what's here and reshuffles it. If another general namespace is introduced, it may be aliased as e.g. `ns6` in the end-products instead of being referred to by nice names.
In case that another OVAL sub-namespace is introduced and we would like to support it in our content, then I think more manual work will be required, and updating the dictionary will be even one of the things needed, and it will be needed for another reasons than to make beautiful datastreams.

ssg/constants.py
yuumasato: Why not reuse `cat_namespace` defined above?

ssg/constants.py
yuumasato: I don't see new name spaces added so frequently that it warrants automation.

ssg/constants.py
```diff
@@ -75,6 +75,33 @@
...
 
+
+OVAL_SUB_NS = dict(
+    ind="independent",
+    unix="unix",
+    linux="linux",
+)
+
+
+ALL_NS = {
```
yuumasato, line 86: How about following pattern in the file and name it something like `PREFIX_to_NS`?

matejak:
Sure, I like suggestions, so they are implemented.

PR #5559, opened 2020-04-03
Fix ol8 condition in accounts-physical rules

iokomin:
#### Description:

systemd conditions in account-physical OVAL definitions applicable to ol8, extending product list.

#### Testing:
- checked ol7 and ol8 builds


iokomin:
Would appreciate for explanation on content-test-filtering test failure, unfortunately I couldn't get the reason from logs.

mildas:
@iokomin Hello, the content-test-filtering is an analysis of changes made in a Pull request and it recommends a user which tests should be run to test the changes.

But looks like it has a problem with parsing XML and it's highly possible that it is content-test-filtering issue. Don't worry about that, this check is not required to pass and it doesn't block reviewer from merging your Pull request.

mildas:
retest this please

PR #5566, opened 2020-04-06
shared/partition.sh: Increase the size of a test device

matusmarhefka:
#### Description:

In some test scenarios (`fstab.fail.sh`) we create a new device
with size of 50MB to test rule `mount_option_tmp_noexec` and it
is re-mounted as the `/tmp`. This causes that there is not enough
space in `/tmp` where our test suite produces logs and result files.
This commit increases the size of the created device so logs can
fit in. Logs are copied from the guest to the host before the testing
device mounted as `/tmp` is unmounted and removed so logs are not lost
~~-- because of this the `separate.fail.sh` test scenario's remediation
is also allowed from now.~~

### DIFF
--- a/tests/shared/partition.sh
+++ b/tests/shared/partition.sh
@@ -3,7 +3,9 @@
 PARTITION="/root/new_partition"
 
 create_partition() {
-	dd if=/dev/zero of=$PARTITION bs=1M count=50
+	# Some scenarios re-mount this device as '/tmp', therefore its size
+	# needs to be enough for the produced logs.
+	dd if=/dev/zero of=$PARTITION bs=1M count=200
 	mkfs.ext2 -F $PARTITION
 }
 


linux_os/guide/system/permissions/partitions/mount_option_tmp_noexec/tests/separate.fail.sh
```diff
@@ -1,15 +1,10 @@
 #!/bin/bash
 # profiles = xccdf_org.ssgproject.content_profile_C2S
 
-# Remediating would mount /tmp, which would break the test environment.
-# remediation = none
```
ggbecker, line 5: I guess `remediation = none` here is essential. Unless it gives the error @yuumasato is facing

ggbecker: Actually I'm not facing any issue in SSG test suite running something like:
`$python test_suite.py rule --remediate-using bash --libvirt qemu:///session rhel79 --datastream ../build/ssg-rhel7-ds.xml mount_option_tmp_noexec`
```
INFO - Logging into /home/ggasparb/workspace/github/content/tests/logs/rule-custom-2020-04-14-1736/test_suite.log
INFO - xccdf_org.ssgproject.content_rule_mount_option_tmp_noexec
/INFO - Script runtime.pass.sh using profile xccdf_org.ssgproject.content_profile_C2S OK
INFO - Script separate.fail.sh using profile xccdf_org.ssgproject.content_profile_C2S OK
INFO - Script fstab.fail.sh using profile xccdf_org.ssgproject.content_profile_C2S OK
```

yuumasato: @ggbecker Could you try without `--remediate-using bash`?

matejak: The `oscap` remediation performs fix during the scan, so it may interfere with the scan report. OTOH, `bash` remediation executes a bash script after the initial scan finishes. So I think that this explains what's going on.

jan-cerny:
@matusmarhefka Are you able to reproduce @yuumasato 's problems?

jan-cerny:
@openscap-ci test this please

jan-cerny:
@openscap-ci test this please

jan-cerny:
@openscap-ci test this please

jan-cerny:
@openscap-ci test this please

jan-cerny:
@openscap-ci test this please

linux_os/guide/system/permissions/partitions/mount_option_tmp_noexec/tests/separate.fail.sh
matusmarhefka: Yes, I haven't realized this, remediations causes that `/tmp` is re-mounted so the files stored in the `/tmp` during remediation are shadowed, good catch @yuumasato 

I will return back the `remediation = none`.

PR #5569, opened 2020-04-07
Fix audit_rules_privileged_commands remediation

vojtapolasek:
Be ware that running the rule in profile mode might still fail because the rule catches some temporary binaries created by Dracut after boot which are deleted in several seconds. This is addressed in https://github.com/ComplianceAsCode/content/pull/5572.

#### Description:

- fix regex in the bash remediation function which had problems with binaries having similar name base (sudo and sudoedit)

- Make tests for the rule applicable to rhel8.

- add two tests which catch this fixed case so that it does not appear again.

- apply forced ordering of some groups to avoid problems, e.g. screen package having SUID being installed after the rule is remediated

#### Rationale:

This PR fixes a bug reported at https://bugzilla.redhat.com/show_bug.cgi?id=1691877.

Certain combination of rules caused this rule to fail to remediate properly or to be marked as failed while remediated properly.


vojtapolasek:
I merged the tests together.
I also removed the reordering trick with policy_rules and auditd_configure_rules. This fix is specific to ncp and it will be addressed in the separte commit.

PR #5570, opened 2020-04-07
Fix covscan issues.

ggbecker:
#### Description:

- Fix covscan issues, problems like:
  - variable assignment
  - using wildcards when storing values

#### Rationale:

- Gives a cleaner covscan results

### DIFF
--- a/shared/bash_remediation_functions/fix_audit_syscall_rule.sh
+++ b/shared/bash_remediation_functions/fix_audit_syscall_rule.sh
@@ -81,6 +81,7 @@ then
 elif [ "$tool" == 'augenrules' ]
 then
 	# Extract audit $key from audit rule so we can use it later
+	matches=()
 	key=$(expr "$full_rule" : '.*-k[[:space:]]\([^[:space:]]\+\)' '|' "$full_rule" : '.*-F[[:space:]]key=\([^[:space:]]\+\)')
 	readarray -t matches < <(sed -s -n -e "\;${pattern};!d" -e "/${arch}/!d" -e "/${group}/!d;F" /etc/audit/rules.d/*.rules)
 	if [ $? -ne 0 ]
--- a/shared/bash_remediation_functions/perform_audit_rules_privileged_commands_remediation.sh
+++ b/shared/bash_remediation_functions/perform_audit_rules_privileged_commands_remediation.sh
@@ -56,6 +56,7 @@ fi
 
 # Obtain the list of SUID/SGID binaries on the particular system (split by newline)
 # into privileged_binaries array
+privileged_binaries=()
 readarray -t privileged_binaries < <(find / -xdev -type f -perm -4000 -o -type f -perm -2000 2>/dev/null)
 
 # Keep list of SUID/SGID binaries that have been already handled within some previous iteration


linux_os/guide/system/permissions/restrictions/enable_execshield_settings/sysctl_kernel_exec_shield/bash/shared.sh
```diff
@@ -19,7 +19,7 @@ if [ "$(getconf LONG_BIT)" = "64" ] ; then
   if grep --silent noexec /boot/grub2/grub*.cfg ; then 
         sed -i "s/noexec.*//g" /etc/default/grub
         sed -i "s/noexec.*//g" /etc/grub.d/*
-        GRUBCFG=/boot/grub2/*.cfg
+        GRUBCFG="/boot/grub2/*.cfg"
```
matejak, line 22: This doesn't solve the problem - the assumption is that there is exactly one file ending with `.cfg`, and it truly won't work if there are more files like that. If you quote it, the asterisk won't be expanded, so it will break completely.
From the top of the `if` statement, it looks like that we may be after `/boot/grub2/grub.cfg` or `/boot/grub2/grub2.cfg`, so it may make sense to determine the config file name early in the remediation, and using it as a simple variable later on.
It can be a for loop that iterates through all occurrences of `/boot/grub2/*.cfg`, assigns the first one to that variable, and throws an error if there are more occurrences.

shared/bash_remediation_functions/fix_audit_syscall_rule.sh
```diff
@@ -81,6 +81,7 @@ then
 elif [ "$tool" == 'augenrules' ]
 then
 	# Extract audit $key from audit rule so we can use it later
+	declare -a matches=()
```
matejak, line 84: What was the error here? Normally, it is OK to just write `matches=()`.

ggbecker: `covscan` thinks that there is no assignment for `matches`. But I guess we can use just `matches=()` then.

linux_os/guide/system/permissions/restrictions/enable_execshield_settings/sysctl_kernel_exec_shield/bash/shared.sh
ggbecker: I agree with your statements. Also I wanted to point out that the OVAL check only checks for a single file:

https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/permissions/restrictions/enable_execshield_settings/sysctl_kernel_exec_shield/oval/shared.xml#L60

ggbecker: I'll remove this piece of code since it needs a little bit more of developing so we can merge the other fixes right away.

shared/bash_remediation_functions/fix_audit_syscall_rule.sh
matejak: So let's do the same with `privileged_binaries`.

linux_os/guide/system/permissions/restrictions/enable_execshield_settings/sysctl_kernel_exec_shield/bash/shared.sh
matejak: I am fine with the remediation being in line with the OVAL check.

shared/bash_remediation_functions/fix_audit_syscall_rule.sh
ggbecker: Done

matejak:
LGTM, thanks!

PR #5571, opened 2020-04-07
Skip generation of remediation when using special the default profile

ggbecker:
#### Description:

- Skips the generation of remediation when profile id is empty (also known as the `default` profile)

#### Rationale:

- It generates empty files that are useless.

Below is a sample when generating remediation and there is no profile specified:

```
$ oscap xccdf generate fix /usr/share/xml/scap/ssg/content/ssg-fedora-xccdf.xml
      #!/bin/bash
      ###############################################################################
      #
      # Bash Remediation Script for No profile (default benchmark)
      #
      # Profile Description:
      # Not available
      #
      # Profile ID:  (null)
      # Benchmark ID:  FEDORA
      # Benchmark Version:  0.1.49
      # XCCDF Version:  1.1
      #
      # This file was generated by OpenSCAP 1.3.2 using:
      # $ oscap xccdf generate fix --profile (null) --fix-type bash xccdf-file.xml
      #
      # This Bash Remediation Script is generated from an OpenSCAP profile without preliminary evaluation.
      # It attempts to fix every selected rule, even if the system is already compliant.
      #
      # How to apply this Bash Remediation Script:
      # $ sudo ./remediation-script.sh
      #
      ##############################################################################
```

### DIFF
--- a/ssg/build_profile_remediations.py
+++ b/ssg/build_profile_remediations.py
@@ -70,7 +70,7 @@ def get_output_paths(benchmarks, benchmark_profile_pairs, path_base, extension,
 
     for benchmark_id, profile_id, _ in benchmark_profile_pairs:
         # profile (default) is not applicable when generating remediation
-        if profile_id == "":
+        if not profile_id:
             continue
 
         if _is_blacklisted_profile(profile_id):
@@ -97,7 +97,7 @@ def fill_queue(benchmarks, benchmark_profile_pairs, input_path, path_base,
 
     for benchmark_id, profile_id, _ in benchmark_profile_pairs:
         # profile (default) is not applicable when generating remediation
-        if profile_id == "":
+        if not profile_id:
             continue
 
         if _is_blacklisted_profile(profile_id):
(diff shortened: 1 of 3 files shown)


ssg/build_profile_remediations.py
```diff
@@ -69,6 +69,10 @@ def get_output_paths(benchmarks, benchmark_profile_pairs, path_base, extension,
     paths = []
 
     for benchmark_id, profile_id, _ in benchmark_profile_pairs:
+        # profile (default) is not applicable when generating remediation
+        if profile_id == "":
```
matejak, line 73: It's not a big deal, but the so-called Pythonic way is to write `if not profile_id`: https://www.python.org/dev/peps/pep-0008/#programming-recommendations

ggbecker: I agree, but in other parts of the code it does this way, I guess the intention of the original author was to make explicit that it is the `default` profile when it is an empty string. But I'm fine changing that.

ggbecker: I replaced every occurrence of `profile_id == ""`

jan-cerny:
It really doesn't generate it! Thanks!

PR #5572, opened 2020-04-07
ssg test suite: wait 30 seconds for reboot to finish

vojtapolasek:
#### Description:

add 30 seconds delay after machine reboot

#### Rationale:

While working on https://github.com/ComplianceAsCode/content/pull/5569 it appeared that Dracut is creating some temporary files which are deleted shortly after reboot. They interfered with test results.

### DIFF
--- a/tests/ssg_test_suite/oscap.py
+++ b/tests/ssg_test_suite/oscap.py
@@ -281,7 +281,6 @@ class GenericRunner(object):
         # the system to finish startup, there were problems with
         # temporary files created by Dracut during image generation interfering
         # with the scan
-        # expressed in seconds
         self.time_to_finish_startup = 30
 
     def _make_arf_path(self):
@@ -431,7 +430,7 @@ class ProfileRunner(GenericRunner):
             logging.info("Rebooting domain '{0}' before final scan."
                          .format(self.environment.domain_name))
             self.environment.reboot()
-            logging.info("Waiting for {0} seconds to let the system finish startup and delete possible Dracut temporary files"
+            logging.info("Waiting for {0} seconds to let the system finish startup."
                          .format(self.time_to_finish_startup))
             time.sleep(self.time_to_finish_startup)
         return GenericRunner.final(self)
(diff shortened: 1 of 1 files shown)


tests/ssg_test_suite/oscap.py
```diff
@@ -424,6 +425,8 @@ def final(self):
             logging.info("Rebooting domain '{0}' before final scan."
                          .format(self.environment.domain_name))
             self.environment.reboot()
+            logging.info("Sleeping for 60 seconds.")
```
yuumasato, line 428: Could you print a more meaningful message? Like "Waiting for reboot to finish"

yuumasato: It would be a good idea to have the time amount in a variable.

tests/ssg_test_suite/oscap.py
```diff
@@ -276,6 +277,12 @@ def __init__(self, environment, profile, datastream, benchmark_id):
         self.command_base = []
         self.command_options = []
         self.command_operands = []
+        # number of seconds to sleep after reboot of vm to let
+        # the system to finish startup, there were problems with
+        # temporary files created by Dracut during image generation interfering
+        # with the scan
+        # expressed in seconds
```
yuumasato, line 284: Nitpicky: these two comment lines can be made a one line comment.

tests/ssg_test_suite/oscap.py
```diff
@@ -424,6 +431,9 @@ def final(self):
             logging.info("Rebooting domain '{0}' before final scan."
                          .format(self.environment.domain_name))
             self.environment.reboot()
+            logging.info("Waiting for {0} seconds to let the system finish startup and delete possible Dracut temporary files"
+                         .format(self.time_to_finish_startup))
```
yuumasato, line 435: Now I think the printed message is too specific, :-P.

The comment above definition of `time_to_finish_startup` should be enough to justify the delay.
For the log message itself, I think just `Waiting for {0} seconds to let the system finish startup` is enough.

Issue #5568, opened 2020-04-07
Rule grub2_ipv6_disable_argument removes sysctl ipv6 variables and causes rules to fail


yuumasato:
#### Description of problem:

Remediation of rule `grub2_ipv6_disable_argument` in CIS profile makes all sysctl ipv6 variable unavailable, causing sysctl ipv6 rules to `fail`.
After remediation of `grub2_ipv6_disable_argument`, the following rules evaluate to `fail`, and will result in `error` if tried to be remediated.
- sysctl_net_ipv6_conf_all_accept_redirects
- sysctl_net_ipv6_conf_default_accept_ra
- sysctl_net_ipv6_conf_all_accept_ra
- sysctl_net_ipv6_conf_default_accept_redirects

The remediation fails with:
```
sysctl: cannot stat /proc/sys/net/ipv6/conf/all/accept_ra: No such file or directory
```

#### SCAP Security Guide Version:
master (future 0.1.50)

#### Actual Results:
The sysctl ipv6 rules evaluate to `fail`, and `error` if remediated.

#### Expected Results:
The sysctl ipv6 rules evaluate to pass and the sysctl ipv6 variables are unavailable.

matejak:
Is it fixed by that PR, or is the issue more about having those rules more robust, so both the check and remediation can handle the case when a sysctl variable is not available?

yuumasato:
Its up to discussion.

The requirements are:
```
3.3.1 Ensure IPv6 router advertisements are not accepted (Not Scored) 
3.3.2 Ensure IPv6 redirects are not accepted (Not Scored) 
3.3.3 Ensure IPv6 is disabled (Not Scored)
```
In 3.3.3 it is clear that disabling IPv6 is not a hard requirement:
> If IPv6 is not to be used, it is recommended that it be disabled to reduce attack surface of the system.

I see three use cases:
A. IPv6 protocol is not used and ipv6 kernel module is not needed - tailoring needed to remove all three rules and add `grub2_ipv6_disable_argument` 
B. IPv6 protocol is not used but ipv6 kernel module is needed - no tailoring needed
C. IPv6 protocol is used - tailoring needed to remove all three rule

I think use case A needs either:
- Documentation that when Rule `grub2_ipv6_disable_argument` is added, the mentioned IPv6 rules need to be removed as well, or;
- That the mentioned IPv6 rules don't fail when `grub2_ipv6_disable_argument` is added.

marcusburghardt:
[...]

PR #5574, opened 2020-04-08
change rules for disabling ipv6 in CIS profile

vojtapolasek:
#### Description:

Replace grub2_ipv6_disable_argument with kernel_module_ipv6_option_disabled in both rhel7 and rhel8 cis profiles.

#### Rationale:

Although the benchmark recommends to use the grub2_ipv6_disable_argument approach, at the same time it breaks other rules, see https://github.com/ComplianceAsCode/content/issues/5568.

The final solution will be probably to document several options for disabling IPv6 and letting user pick the right one.


matejak:
LGTM, merging.

PR #5576, opened 2020-04-09
Fix ncp audit rules

vojtapolasek:
#### Description:

- remove audit_- rules from ncp which are already covered by audit_rules_from_ospp. This rule is in ospp profile which ncp profile extends.

- reorder groups so that audit_rules_for_ospp is executed before other smaller audit rules.

#### Rationale:

The fix is connected to https://github.com/ComplianceAsCode/content/pull/5569 and it tries  to fix problems with audit bash remediations.

The rule audit_rules_for_ospp applies large set of changes to audit configuration. These checks were duplicated in the ncp profile by smaller rules. But the way in which audit_rules_for_ospp and rules from auditd_configure_rules check and remediate the audit configuration slightly differs. Therefore, rules were conflicting and there appeared duplicate configuration entries or some rules were reporting failure although technically teh system was configured correctly.


shawndwells:
NACK. US Government baselines need independent audit rules.

vojtapolasek:
> NACK. US Government baselines need independent audit rules.

OK, I understand. So let's take it from the opposite side. Would it be acceptable if specific Audit rules in the NCP profile are left and the large rule audit_rules_from_ospp which is originally in the OSPP profile would be deselected?

vojtapolasek:
I experimentally tried this and after removal of audit_rules_for_ospp no more Audit related rules are failing. Now the question is if the removal would keep the same level of compliance. Note that there is a huge if not total overlap between audit_rules_for_ospp and individual rules present in ncp. Opinions? @shawndwells  @redhatrises

shawndwells:
@vojtapolasek really hate the never-ending hack of grouping audit rules. But until a better solution is made, it does seem to make sense for the CoreOS profile to have whatever audit rules are present from RHEL OSPP. Better to ship something than nothing.

@redhatrises What are your thoughts?

matejak:
AFAIK this PR doesn't aim to have a functional impact, it just presents the way to reduce duplication that is close to the approach used in OSPP.

@redhatrises Are you OK with that, or would you rather propose another solution?

mildas:
**Changes identified:**
Profile ncp on rhel7:
 Rule audit_rules_unsuccessful_file_modification_open, audit_rules_mac_modification, audit_rules_file_deletion_events, audit_rules_dac_modification_fsetxattr, audit_rules_unsuccessful_file_modification_openat, audit_rules_privileged_commands_unix_chkpwd, audit_rules_privileged_commands_crontab, audit_rules_file_deletion_events_unlink, audit_rules_dac_modification_chmod, audit_rules_dac_modification_lsetxattr, audit_rules_privileged_commands_userhelper, audit_rules_file_deletion_events_rename, audit_rules_privileged_commands_passwd, audit_rules_dac_modification_fremovexattr, audit_rules_privileged_commands_umount, audit_rules_kernel_module_loading_init, audit_rules_file_deletion_events_rmdir, audit_rules_usergroup_modification_group, audit_rules_session_events, audit_rules_kernel_module_loading_delete, audit_rules_unsuccessful_file_modification_truncate, audit_rules_usergroup_modification_shadow, audit_rules_dac_modification_lchown, audit_rules_privileged_commands_newgrp, audit_rules_dac_modification_fchownat, audit_rules_dac_modification_lremovexattr, audit_rules_dac_modification_fchmod, audit_rules_unsuccessful_file_modification_creat, audit_rules_unsuccessful_file_modification_ftruncate, audit_rules_dac_modification_fchmodat, audit_rules_dac_modification_chown, audit_rules_unsuccessful_file_modification_open_by_handle_at, audit_rules_usergroup_modification_passwd, audit_rules_file_deletion_events_unlinkat, audit_rules_dac_modification_setxattr, audit_rules_dac_modification_fchown, audit_rules_usergroup_modification_gshadow, audit_rules_privileged_commands_gpasswd, audit_rules_file_deletion_events_renameat, audit_rules_dac_modification_removexattr removed from ncp profile.
Others:

 build_product rhel7
 tests/test_suite.py profile --libvirt qemu:///system test-suite-vm --datastream build/ssg-rhel7-ds.xml ncp

redhatrises:
This PR and how OSPP handles breaks expectations around being able to fine grain and fine tune audit rules. Organizations are supposed to be able to choose which events are handled whereas the OSPP rules do not allow that. I don't see a way forward with this PR.

redhatrises:
Closing as this doesn't allow auditors to choose auditing rules in scap policy per NIST and DISA expectations.

PR #5567, opened 2020-04-07
Feature sle15 cis

eradot4027:
#### Description:

- Added CIS Benchmark Profile for SLE15 and enabled checks to support that profile
#### Rationale:

- CIS SLE15 Benchmark is in draft and the only one out there at the moment under development by and standards organization.

PR #5578, opened 2020-04-09
Feature suse15 cis

eradot4027:
#### Description:
Description:

    Added CIS Benchmark Profile for SLE15 and enabled checks to support that profile. In place of closed #5567

Rationale:

    CIS SLE15 Benchmark is in draft and the only one out there at the moment under development by and standards organization.


redhatrises:
Don't see why this can't be merged. Ack.

yuumasato:
@redhatrises thanks for reviewing it.
I think this was a re-submission of #5567 that fell through the cracks.

PR #5580, opened 2020-04-12
don't add boot UUID if same as root UUID

### LINKED ISSUE #2248: grub2_enable_fips_mode prose guidance causes boot fail on an EFI system
(Tested on RHEL 7.4.)

The guidance lacks an additional `grub.cfg` argument identifying the rootfs location.

The guidance in https://access.redhat.com/solutions/137833 has been tested and found to work properly. If that guidance is determined to be doctrinally correct, it should be stolen for SSG.

Separately and additionally, the OVAL test for FIPS mode dfoes not actually determine if FIPS mode is in effect. It should be augmented (or replaced) with a sysctl test for `crypto.fips_enabled = 1`.

enjaygee:
#### Description:

- Only add boot UUID to the GRUB command-line and add as boot arg to kernel command line if different from root UUID.

#### Rationale:

- Boot UUID should only be added to the GRUB command-line (and added as boot arg to kernel command line) if different from root UUID. Adding UUID when the same results in boot fail.  Discussed in https://github.com/ComplianceAsCode/content/issues/2248

- Fixes #2248

### DIFF
--- a/linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/bash/shared.sh
+++ b/linux_os/guide/system/software/integrity/fips/grub2_enable_fips_mode/bash/shared.sh
@@ -21,16 +21,25 @@ else
 	sed -i 's/\(^GRUB_CMDLINE_LINUX=".*\)"/\1 fips=1"/'  /etc/default/grub
 fi
 
+# Get the UUID of the device mounted at root (/).
+ROOT_UUID=$(findmnt --noheadings --output uuid --target /)
+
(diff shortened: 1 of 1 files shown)


redhatrises:
@openscap-ci test this please

jan-cerny:
I have installed a fresh RHEL 7.8 Server virtual machine. During installation, I have created only a single partition for /, therefore /boot was on the same partition as /. The rule that failed before remediation started to pass after remediation. Thank you very much for your patch.

Issue #5561, opened 2020-04-03
Rule deselection syntax doesn't work for extended profiles


Massi78:
#### Description of problem:
The deselection syntax for profiles it is not working if I try to deselect a rule which is present in a profile I'm extending from.

#### SCAP Security Guide Version:
0.1.49

#### Operating System Version:
ubuntu1804

#### Steps to Reproduce:

1. Writing a couple of rules for *profile_1*
```
documentation_complete: true
title: 'Profile 1'
description: 'something'
selections:
    - sysctl_fs_suid_dumpable
    - sysctl_kernel_randomize_va_space
```
2. Writing *profile_2* which extends *profile_1*
```
documentation_complete: true
title: 'Profile 2'
description: 'something'
extends: profile_1
selections:
    - "!sysctl_kernel_randomize_va_space"
```

#### Result
The evaluation of the command:
```
$> oscap xccdf eval --profile profile_2 ...
```
Includes both the rules even if I'm expecting just *sysctl_fs_suid_dumpable*

PR #5581, opened 2020-04-14
Fix the profile compilation.

matejak:
After resolving the actual "extends" relationship, remove the reference to the extended profile.
One can either keep unselections and extends, or one has to get rid of both.

Fixes: #5561


vojtapolasek:
Thank you for this fix. It looks good.

PR #5582, opened 2020-04-14
fix sshd_allow_only_protocol2

vojtapolasek:
#### Description:

This rule was templated. However, the template does not agree with the rule description. Therefore, it was decided to revert to prior checks and remediations which were used before templating.

#### Rationale:

The template searched for the "protocol" parameter in the /etc/ssh/sshd_config. OpenSSH servers with version higher than 7.4 support only protocol 2 and therefore the config file doesn't need to contain this configuration.

Fixes: RHBZ#1823576

### DIFF
--- a/linux_os/guide/services/ssh/ssh_server/sshd_allow_only_protocol2/oval/shared.xml
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_allow_only_protocol2/oval/shared.xml
@@ -2,14 +2,7 @@
   <definition class="compliance" id="sshd_allow_only_protocol2" version="1">
     <metadata>
       <title>Ensure Only Protocol 2 Connections Allowed</title>
-      <affected family="unix">
-        <platform>multi_platform_wrlinux</platform>
-        <platform>multi_platform_rhel</platform>
-        <platform>multi_platform_rhv</platform>
-        <platform>multi_platform_debian</platform>
-        <platform>multi_platform_ubuntu</platform>
-        <platform>multi_platform_ol</platform>
-      </affected>
+      {{{- oval_affected(products) }}}
       <description>The OpenSSH daemon should be running protocol 2.</description>
     </metadata>
     <criteria comment="SSH is configured correctly or is not installed"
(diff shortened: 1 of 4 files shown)


linux_os/guide/services/ssh/ssh_server/sshd_allow_only_protocol2/oval/shared.xml
```diff
@@ -0,0 +1,45 @@
...
+        <platform>multi_platform_rhv</platform>
+        <platform>multi_platform_debian</platform>
+        <platform>multi_platform_ubuntu</platform>
+        <platform>multi_platform_ol</platform>
+      </affected>
+      <description>The OpenSSH daemon should be running protocol 2.</description>
+    </metadata>
+    <criteria comment="SSH is configured correctly or is not installed"
+    operator="OR">
+      <criteria comment="sshd is not installed" operator="AND">
```
jan-cerny, line 17: We could use Jinja macros application_not_required_or_requirement_unset and application_required_or_requirement_unset here. The latter doesn't even contain the closing criteria tag, so we could easily sneak in the extend_definition element for the version check.

But I don't know if it would be useful for anything.

jan-cerny:
It's interesting that we already have the test scenario for this situation. openssh-7.4.pass.sh. It fails on current master and starts to pass with this PR.

linux_os/guide/services/ssh/ssh_server/sshd_allow_only_protocol2/oval/shared.xml
vojtapolasek: I would not use this macro because it has hardcoded descriptions connected with sshd and it would only confuse people. I don't think it brings any benefit. But I will add macro affected_oval.

jan-cerny: OK, makes sense

jan-cerny:
@openscap-ci test this please

PR #5589, opened 2020-04-15
ocp4: Enable ipv6-specific sysctl checks in moderate profile

JAORMX:
Some sysctl settings were causing issues with the remediation
application. These are not the problematic ones.


jan-cerny:
@openscap-ci test this please

PR #5598, opened 2020-04-15
Use jinja macros for bash remediation in rule sshd_set_keepalive.

ggbecker:
#### Description:

- Use jinja macros in favor of plain bash functions.

#### Rationale:

- Less error prone


redhatrises:
Looks like a huge bug is in the Jinja templates, CCE has been removed from remediations. Nack until this has been fixed.

ggbecker:
> Looks like a huge bug is in the Jinja templates, CCE has been removed from remediations. Nack until this has been fixed.

Started working on a fix here: #5599

PR #5599, opened 2020-04-15
Add @CCENUM@ placeholder for bash jinja macros

ggbecker:
#### Description:

- Add placeholder for CCE in bash remediation. These identifiers are replaced by the xslt transformation from here: https://github.com/ComplianceAsCode/content/blob/master/shared/transforms/xccdf-addremediations.xslt#L69

#### Rationale:

- Reintroduces back the functionality implemented by the `replace_or_append` bash function.

~~There is still some strange behavior which only one remediation is getting the right `@CCENUM` placeholder. Needs investigation. Keeping it draft.~~
Update1: It was missing the preprocessing of templated rules which uses bash jinja macros
Update2: This pull request considers that a comment is always the character: `#`. So, for other types of configuration files that use a different symbol to represent comment, it will not work.

Fixes #5599

### DIFF
--- a/shared/macros-bash.jinja
+++ b/shared/macros-bash.jinja
@@ -443,6 +443,7 @@ line_number="$(LC_ALL=C grep -n "{{{ insert_after }}}" "{{{ path }}}.bak" | LC_A
 if [ -z "$line_number" ]; then
     # There was no match of '{{{ insert_after }}}', insert at
     # the end of the file.
+    printf '# Per %s: Set %s in %s\n' "{{{ "@CCENUM" if has_cce else 'CCE' }}}" "{{{ line }}}" "{{{ path }}}" >> "{{{ path }}}"
     printf '%s\n' "{{{ line }}}" >> "{{{ path }}}"
 else
     head -n "$(( line_number ))" "{{{ path }}}.bak" > "{{{ path }}}"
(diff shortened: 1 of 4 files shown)


ggbecker:
@matejak This PR only proposes comments using bash syntax. If we want to support other kinds of comment syntax, we would need to have the comment symbol character as an input parameter. Do you think this is feasible?

jan-cerny:
Is it possible to do it without using the XSLT? Can the CCE be obtained from the `rule.yml` of the respective rule?

ggbecker:
> Is it possible to do it without using the XSLT? Can the CCE be obtained from the `rule.yml` of the respective rule?

It should be fairly simple to achieve this, since the code already identifies which rule contain the CCE before putting the placeholder in place.

But there is still the issue with comment syntax that might break other types of configuration files than the ones which have `#` as a comment symbol.

jan-cerny:
> It should be fairly simple to achieve this, since the code already identifies which rule contain the CCE before putting the placeholder in place.

Great news, I envision removing the XSLT in long term.

> But there is still the issue with comment syntax that might break other types of configuration files than the ones which have # as a comment symbol.

Great point! I'm afraid that this is the weak point of this feature because remediation authors can forget about this feature.

It probably needs to be a parameter of the macro. It can have the default value `#`.

ggbecker:
I have discovered that some cases are not covered by the current implementation. For example, the bash macro for [coredump_disable_storage](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/permissions/restrictions/coredumps/coredump_disable_storage/bash/shared.sh#L6) is not considering the CCE in the final generated bash remediation, so there is some extra work needed besides the one pointed out by Jan Cerny.

jan-cerny:
@ggbecker OK, good point. I'm afraid that implementing this feature correctly will require some more effort.

ggbecker:
I don't think the effort needed to make this PR work for all cases that use lineinfile type of remediation is worthy, since it was never really well implemented before, causing duplicates, etc... and that this is only a cosmetic thing, I propose to close this PR and if there are any objections feel free to reopen again and propose a proper justification on why we need this feature.

matejak:
Thanks for getting back to this PR. I agree that it is better not to have this feature than having it "to some degree", i.e. with exceptions and pitfalls that were discovered.
We can get back here if demand for it arises.

PR #5610, opened 2020-04-16
Update sshd disable x11 forwarding

ggbecker:
#### Description:

- Add CCE identifiers for RHEL7 and RHEL8
- Correct CIS reference number for RHEL7 according to [RHEL7 CIS profile comment](https://github.com/ComplianceAsCode/content/blob/master/rhel7/profiles/cis.profile#L580)


shawndwells:
> Nice catch!

hah. looks like we reviewed at the same time

PR #5619, opened 2020-04-16
Change permissions to 644 for passwd- file from rule file_permissions_backup_etc_passwd

### LINKED ISSUE #5611: Permissions of file remediated by file_permissions_backup_etc_passwd are reverted when a user is added
#### Description of problem:
[file_permissions_backup_etc_passwd](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/permissions/files/permissions_important_account_files/file_permissions_backup_etc_passwd/rule.yml)

This rule sets the permission `600` for the file `/etc/passwd-`, but as soon as a user is added to the system, the permissions are reverted back to `644`. It's either a misalignment on CIS or a bug in the component which takes care of this file.
```
rpm -qf /etc/passwd
setup-2.12.2-1.el8.noarch
```

Opening issue for awareness

0.1.49

RHEL8

#### Steps to Reproduce:

1. `chmod 600 /etc/passwd-`
2. `adduser user1`
3. `ls -la /etc/passwd-`

#### Actual Results:
Permissions are reverted to 644

#### Expected Results:
Permissions are 600

Report generated during RHEL8 installation phase selecting CIS profile
[eval_remediate_report.txt](https://github.com/ComplianceAsCode/content/files/4487156/eval_remediate_report.txt)
Report generated by a scan after installation of RHEL8 finished:
[xccdf_org.ssgproject.content_profile_cis.txt](https://github.com/ComplianceAsCode/content/files/4487162/xccdf_org.ssgproject.content_profile_cis.txt)
Note: change the extension back to `html` to be able to open the report in a browser.

KB for RHEL7 created in 2017:
https://access.redhat.com/solutions/3190922

ggbecker:
#### Description:

- Update expected permissions to `644` from file `/etc/passwd-`

#### Rationale:

- According to latest CIS RHEL7 benchmark (v2.2.0) the actual value is `644` instead of `600`

From section `6.1.6 Ensure permissions on /etc/passwd- are configured (Scored)`
```
Run the following command and verify Uid and Gid are both 0/root and Access is 644 or
more restrictive:
# stat /etc/passwd-
Access: (0644/-rw-------)
```
The problem here was that when the backup file is regenerated by `adduser` command for example, it has the same permissions as the original file (`/etc/passwd` has `644`), then the check would always fail after the backup file gets updated.

Now, RHEL8 benchmark still says to check for `600` even in the latest version, but we believe this is a finding, since for RHEL7 had the same behavior, also there was a support ticket created in order to update to `644` in the RHEL7 benchmark. I have created a ticket under official CIS ticket system to update this configuration for RHEL8.

Fixes #5611

### DIFF
--- a/linux_os/guide/system/permissions/files/permissions_important_account_files/file_permissions_backup_etc_passwd/rule.yml
+++ b/linux_os/guide/system/permissions/files/permissions_important_account_files/file_permissions_backup_etc_passwd/rule.yml
@@ -3,7 +3,7 @@ documentation_complete: true
 title: 'Verify Permissions on Backup passwd File'
 
 description: |-
-    {{{ describe_file_permissions(file="/etc/passwd-", perms="0600") }}}
+    {{{ describe_file_permissions(file="/etc/passwd-", perms="0644") }}}
 
 rationale: |-
     The <tt>/etc/passwd-</tt> file is a backup file of <tt>/etc/passwd</tt>, and as such,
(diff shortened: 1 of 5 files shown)


redhatrises:
And making it 0600 seems wrong considering that passwd in general is 0644.

redhatrises:
This looks good to me actually. The default is 0644, and as the bug correctly identifies, will always change to the file back to the original configuration.

PR #5624, opened 2020-04-16
Fix test scenarios for auditd_data_retention_flush rule

matusmarhefka:
#### Description:

Based on the current ospp in RHEL-7 and RHEL-8 the correct
value for flush is "incremental_async".


matusmarhefka:
@shawndwells You are right, we need to add platform for RHEL8 as well. I will do it in an additional PR, thanks for noticing.

PR #5625, opened 2020-04-16
Add check and remediation for xwindows_runlevel_target and select in profiles that remove package xorg-x11-server-common

ggbecker:
#### Description:

~~-Disable remediation provided by `package removed` template for rule `package_xorg-x11-server-common_removed`.~~

- Add check and remediation for `xwindows_runlevel_target` and select in profiles that remove package `xorg-x11-server-common`.

~~Add custom remediation rule `package_xorg-x11-server-common_removed` which removes the package and additionally set the default target of the system to `multi-user.target` so system does not end with a black screen that looks like the system is somehow broken.~~

#### Rationale:

This PR attempts to fix/mitigate the issue [BZ#1648162](https://bugzilla.redhat.com/show_bug.cgi?id=1648162) where many people hit the issue of installing RHEL7 using GUI and selecting the `STIG` profile then ending with a black screen after the installation finishes.

This happens because `STIG` profile selects the rule [package_xorg-x11-server-common_removed](https://github.com/ComplianceAsCode/content/blob/master/rhel7/profiles/stig.profile#L275) and this package is required when installing a system with GUI. The system doesn't break, you need just to switch to a different target (`console`).

There is no obvious fix for this problem without changing `OSCAP-Anaconda-Addon` to block this combination of `System with GUI` + Profile which selects the rule `package_xorg-x11-server-common_removed`. But opting for `System with GUI` can be a valid use case of hardening, then preventing an user from installing using this combination doesn't seem to be a good option.

~~This PR keeps the OVAL check in order to keep the automated profile validation, in most of the cases users won't use a system with GUI, so the check will naturally pass since the package won't be present. For those who would like to have GUI, then it will fail and results can either waived or manually remediated in a controlled environment.~~

I also would like to mention that RHEL8 defaults to installation using GUI, and [CIS profile](https://github.com/ComplianceAsCode/content/blob/master/rhel8/profiles/cis.profile#L263) for RHEL8 selects this rule.

Additionally, even though the bugzilla only mentions STIG profile, there are other profiles that can cause the same problem, this is the current list of profiles that selects this rule:

```
rhv4/profiles/rhvh-stig.profile
rhel8/profiles/cis.profile
rhel7/profiles/C2S.profile
rhel7/profiles/stig.profile
rhel7/profiles/rhelh-stig.profile
rhel7/profiles/cis.profile
sle15/profiles/cis.profile
rhel6/profiles/C2S.profile
rhel6/profiles/server.profile
rhel6/profiles/stig.profile
rhel6/profiles/CS2.profile
```


linux_os/guide/services/xwindows/disabling_xwindows/package_xorg-x11-server-common_removed/rule.yml
```diff
@@ -47,7 +47,18 @@ ocil: |-
     The output should be:
     <pre>package xorg-x11-server-common is not installed</pre>
 
+warnings:
+    - general: |-
+        If you use a system with Graphical User Interface (GUI), by removing the package xorg-x11-server-common, it can
+        potentially remove the graphical target and it might bring your system to an inconsistent state where
+        additional configuration can be required to access the system again. Only remove this package if you are sure.
```
redhatrises, line 54: Warning should be of type `functionality`.

Text should be updated:
````
        The installation and use of a Graphical User Interface (GUI) increases your attack vector and decreases your
        overall security posture. Removing the package xorg-x11-server-common package can
￼        potentially remove the graphical target which might bring your system to an inconsistent state requiring
￼        additional configuration to access the system again. If a GUI is an operational requirement, a tailored profile
        that removes this rule should used before continuing installation.
````

File: linux_os/guide/services/xwindows/disabling_xwindows/package_xorg-x11-server-common_removed/rule.yml [hunk 2]
```diff
@@ -47,7 +47,18 @@ ocil: |-
...
+
 template:
     name: package_removed
     vars:
         pkgname: xorg-x11-server-common
+    backends:
```
redhatrises, line 60: Remediations should be kept. Tailoring should be used instead.

yuumasato: > Sounds like the installer has a bug or should otherwise notify the user about package conflicts.

This is acknowledged in the PR description. 
I see this as a mitigation for the technical problem, which is the machine seems broken after installation.

Another possible approach is to remove this rule from "templation" and extend the bash remediation to also change the target to `multi-user`.

To solve the conflict of expectations (with or without GUI) a change in OAA is needed anyway.

yuumasato: > Another possible approach is to remove this rule from "templation" and extend the bash remediation to also change the target to `multi-user`.

Difference between the approaches being that in one case the GUI is there (the intended server install config wins), in the other the GUI is not there (the intended policy wins).

linux_os/guide/services/xwindows/disabling_xwindows/package_xorg-x11-server-common_removed/rule.yml
ggbecker: Updated accordingly.

linux_os/guide/services/xwindows/disabling_xwindows/package_xorg-x11-server-common_removed/rule.yml
ggbecker: I have reverted back the remediations and took a different approach. I added checks and remediation for rule [xwindows_runlevel_target](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/xwindows/disabling_xwindows/xwindows_runlevel_target/rule.yml) which will select the correct target for the system and it will prevent from ending up with a black screen. This rule has been selected in all profiles (but the one from RHEL6 because it is not applicable) that select the rule: `package_xorg-x11-server-common_removed`.

ggbecker:
I have changed the approach to fix the issue by selecting the rule that changes the target of a system to `multi-user.target` preventing the system to end up with a black screen and having to manually update the target after accessing a terminal with `Ctrl-Alt-F2` for example.

Remediations of package removal were reverted back.

linux_os/guide/services/xwindows/disabling_xwindows/xwindows_runlevel_target/bash/shared.sh
```diff
@@ -0,0 +1,7 @@
+# platform = multi_platform_sle,multi_platform_rhv,multi_platform_fedora,Red Hat Enterprise Linux 7,Red Hat Enterprise Linux 8
+# reboot = true
+# strategy = restrict
+# complexity = low
+# disruption = low
+
+systemctl set-default multi-user.target
```
yuumasato, line 7: Have you checked that `systemctl` command works during Anaconda post-install phase?

yuumasato:
> Remediations should be kept. Tailoring should be used instead.

> These baselines require removal of this package. 

Do I understand correctly that from these comments that you'd rather support the installation a Server with GUI hardened to STIG only via tailoring?

linux_os/guide/services/xwindows/disabling_xwindows/xwindows_runlevel_target/bash/shared.sh
ggbecker: Not yet. I'm still thinking how I can test this scenario.

linux_os/guide/services/xwindows/disabling_xwindows/xwindows_runlevel_target/ansible/shared.yml
```diff
@@ -0,0 +1,12 @@
...
+# strategy = restrict
+# complexity = low
+# disruption = low
+
+- name: get default target
+  command: systemctl get-default
+  register: default_target
+- name: Switch to multi-user runlevel
+  command: systemctl set-default multi-user.target
+  when: default_target.stdout != "multi-user.target"
```
redhatrises, line 12: This should use the `file` module and not the `command` module

rhel7/profiles/stig.profile
```diff
@@ -273,6 +273,7 @@ selections:
     - sshd_enable_x11_forwarding
     - tftpd_uses_secure_mode
     - package_xorg-x11-server-common_removed
+    - xwindows_runlevel_target
```
redhatrises, line 276: Does this rule exist in the STIG?

linux_os/guide/services/xwindows/disabling_xwindows/xwindows_runlevel_target/oval/shared.xml
```diff
@@ -0,0 +1,49 @@
...
+    <unix:filepath>/etc/systemd/system/default.target</unix:filepath>
+    <unix:canonical_path operation="pattern match">^/usr/lib/systemd/system/multi-user.target$</unix:canonical_path>
+  </unix:symlink_state>
+  {{%- else %}}
+  <ind:textfilecontent54_test check="all" comment="default runlevel is 3" id="test_etc_inittab_default_runlevel_target" version="1">
+    <ind:object object_ref="object_etc_inittab_default_runlevel_target" />
+  </ind:textfilecontent54_test>
+  <ind:textfilecontent54_object id="object_etc_inittab_default_runlevel_target" version="1">
+    <ind:filepath>/etc/inittab</ind:filepath>
+    <ind:pattern operation="pattern match">^[\s]*id:3:initdefault:[\s]*$</ind:pattern>
```
redhatrises, line 43: This shouldn't be here. It should stay under `xwindows_runlevel_setting`

linux_os/guide/services/xwindows/disabling_xwindows/xwindows_runlevel_target/tests/correct_target.pass.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+# platform = multi_platform_sle,multi_platform_rhv,multi_platform_fedora,Red Hat Enterprise Linux 7,Red Hat Enterprise Linux 8
+
+rm -f /etc/systemd/system/default.target
+ln -s /usr/lib/systemd/system/multi-user.target /etc/systemd/system/default.target
```
redhatrises, line 5: Why not use the systemctl commands here?

linux_os/guide/services/xwindows/disabling_xwindows/xwindows_runlevel_target/tests/wrong_target.fail.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+# platform = multi_platform_sle,multi_platform_rhv,multi_platform_fedora,Red Hat Enterprise Linux 7,Red Hat Enterprise Linux 8
+
+rm -f /etc/systemd/system/default.target
+ln -s /usr/lib/systemd/system/graphical.target /etc/systemd/system/default.target
```
redhatrises, line 5: Why not use the systemctl commands here?

linux_os/guide/services/xwindows/disabling_xwindows/xwindows_runlevel_target/ansible/shared.yml
```diff
@@ -0,0 +1,12 @@
+# platform = multi_platform_sle,multi_platform_rhv,multi_platform_fedora,Red Hat Enterprise Linux 7,Red Hat Enterprise Linux 8
+# reboot = true
+# strategy = restrict
+# complexity = low
+# disruption = low
+
+- name: get default target
+  command: systemctl get-default
+  register: default_target
+- name: Switch to multi-user runlevel
```
redhatrises, line 10: There always needs to be a space in between new plays. Readability is very important.

linux_os/guide/services/xwindows/disabling_xwindows/xwindows_runlevel_target/oval/shared.xml
```diff
@@ -0,0 +1,49 @@
+{{%- if init_system == "systemd" and target_oval_version == [5, 10] -%}}
+{{# this is the only scenario this definition cannot handle, there is no good alternative for symlink_test for OVAL 5.10 #}}
+{{%- else -%}}
+<def-group>
+  <definition class="compliance" id="xwindows_runlevel_target" version="1">
+    <metadata>
+      <title>Disable X Windows Startup By Setting Default SystemD Target</title>
+      {{{- oval_affected(products) }}}
+      {{%- if init_system == "systemd" %}}
+      <description>Checks /etc/systemd/system/default.target to ensure that the default runlevel target is set to multi-user.target.</description>
```
redhatrises, line 10: Remove `/etc/systemd/system/default.target`

linux_os/guide/services/xwindows/disabling_xwindows/xwindows_runlevel_target/oval/shared.xml
```diff
@@ -0,0 +1,49 @@
...
+    <criteria>
+      <criterion comment="default.target systemd softlink exists" test_ref="test_disable_xwindows_runlevel_target" />
+    </criteria>
+    {{%- else %}}
+    <criteria>
+      <criterion comment="default runlevel is 3" test_ref="test_etc_inittab_default_runlevel_target" />
+    </criteria>
+    {{%- endif %}}
+  </definition>
+  {{%- if init_system == "systemd" %}}
```
redhatrises, line 25: Don't need this.

rhel7/profiles/stig.profile
ggbecker: Per: `U_Red_Hat_Enterprise_Linux_7_STIG_V2R6_Manual-xccdf.xml`, there isn't.

linux_os/guide/services/xwindows/disabling_xwindows/xwindows_runlevel_target/ansible/shared.yml
ggbecker: Do you mean to use the `file` module to create the symlink? Isn't there anything else that systemctl would do when setting the default target?

linux_os/guide/services/xwindows/disabling_xwindows/xwindows_runlevel_target/oval/shared.xml
ggbecker: agreed

linux_os/guide/services/xwindows/disabling_xwindows/xwindows_runlevel_target/tests/correct_target.pass.sh
ggbecker: No specific reason. I'll change it.

linux_os/guide/services/xwindows/disabling_xwindows/xwindows_runlevel_target/tests/wrong_target.fail.sh
ggbecker: No specific reason. I'll change it.

linux_os/guide/services/xwindows/disabling_xwindows/xwindows_runlevel_target/ansible/shared.yml
ggbecker: ok

linux_os/guide/services/xwindows/disabling_xwindows/xwindows_runlevel_target/oval/shared.xml
ggbecker: ack

linux_os/guide/services/xwindows/disabling_xwindows/xwindows_runlevel_target/oval/shared.xml
ggbecker: ok

ggbecker:
I haven't noticed that STIG for RHEL7 doesn't contain any configuration item for setting `multi-user.target` as default (RHEL6 does), thus I'm not sure if having this rule is the ideal solution.

I have the impression that customizing the remediation for `package_xorg-x11-server-common_removed` by adding the `systemctl` command to set the `multi-user.target` as default is the best option.

linux_os/guide/services/xwindows/disabling_xwindows/package_xorg-x11-server-common_removed/rule.yml
```diff
@@ -9,8 +9,10 @@ description: |-
     installed. If X Windows is not installed then the system cannot boot into graphical user mode.
     This prevents the system from being accidentally or maliciously booted into a <tt>graphical.target</tt>
     mode. To do so, run the following command:
-    <pre>$ sudo yum groupremove "X Window System"</pre>
-    <pre>$ sudo yum remove xorg-x11-server-common</pre>
+    {{%- if product != "rhel8" and product != "rhv4" -%}}
+    <pre>$ sudo {{{ pkg_manager }}} groupremove "X Window System"</pre>
```
ggbecker, line 13: "X Window System" group is not available on RHEL8

ggbecker:
I added custom bash and ansible remediation for `package_xorg-x11-server-common_removed` and deselected the rule `xwindows_runlevel_target` from profiles since it doesn't make part of STIG benchmark.

The custom remediation removes the package as the old one would do and additionally set the default target as `multi-user.target`, this way when remediating this rule in a GUI system, it will fallback to a terminal system.

I kept the modifications to the rule `xwindows_runlevel_target` in this PR because I think it is still valid and can be used in further occasions.

linux_os/guide/services/xwindows/disabling_xwindows/package_xorg-x11-server-common_removed/ansible/shared.yml
```diff
@@ -0,0 +1,18 @@
...
+    name: "xorg-x11-server-common"
+    state: absent
+
+- name: get default target
+  command: systemctl get-default
+  register: default_target
+
+- name: Switch to multi-user runlevel
+  command: systemctl set-default multi-user.target
+  when: default_target.stdout != "multi-user.target"
```
shawndwells, line 18: NACK. This appears to do much more than remove a package, which is all this check should do.

linux_os/guide/services/xwindows/disabling_xwindows/package_xorg-x11-server-common_removed/bash/shared.sh
```diff
@@ -0,0 +1,7 @@
+# platform = multi_platform_sle,multi_platform_rhv,multi_platform_fedora,Red Hat Enterprise Linux 7,Red Hat Enterprise Linux 8,Oracle Linux 7,Oracle Linux 8
+
+# Install required packages
+{{{ bash_package_remove("xorg-x11-server-common") }}}
+
+systemctl set-default multi-user.target
```
shawndwells, line 6: NACK. This snippet should only remove a package.

linux_os/guide/services/xwindows/disabling_xwindows/package_xorg-x11-server-common_removed/rule.yml
shawndwells: What is the RHEL 8 equivalent?

linux_os/guide/services/xwindows/disabling_xwindows/package_xorg-x11-server-common_removed/rule.yml
```diff
@@ -47,7 +49,18 @@ ocil: |-
     The output should be:
     <pre>package xorg-x11-server-common is not installed</pre>
 
+warnings:
+    - functionality: |-
+        The installation and use of a Graphical User Interface (GUI) increases your attack vector and decreases your
+        overall security posture. Removing the package xorg-x11-server-common package can
+        potentially remove the graphical target which might bring your system to an inconsistent state requiring
+        additional configuration to access the system again. If a GUI is an operational requirement, a tailored profile
+        that removes this rule should used before continuing installation.
```
shawndwells, line 58:
```suggestion
        overall security posture. Removing the package xorg-x11-server-common package will
        remove the graphical target which might bring your system to an inconsistent state requiring
        additional configuration to access the system again. If a GUI is an operational requirement, a tailored profile
        that removes this rule should used before continuing installation.
```

shawndwells: This check will absolutely remove the package, so dropped the "can potentially" language.

linux_os/guide/services/xwindows/disabling_xwindows/xwindows_runlevel_setting/oval/shared.xml
```diff
@@ -35,7 +35,7 @@
   </unix:symlink_object>
   <unix:symlink_state comment="default.target is linked to multi-user.target" id="state_disable_xwindows_runlevel" version="1">
     <unix:filepath>/etc/systemd/system/default.target</unix:filepath>
-    <unix:canonical_path operation="pattern match">^/lib/systemd/system/multi-user.target$</unix:canonical_path>
+    <unix:canonical_path operation="pattern match">^/usr/lib/systemd/system/multi-user.target$</unix:canonical_path>
```
shawndwells, line 38: Any idea when this change was made? Meaning should we create checks for RHEL6-7 using the old path, and RHEL8 for the new?

ggbecker: In my tests, RHEL7.7 demonstrated the same behavior of RHEL8.X. Though the old location is still valid and the file exists, when using `systemctl set-default multi-user.target` it uses the new location with the prefix `/usr/`. I guess it needs to be extended to check for both locations if it is still a valid use case. I'll do some more testing around this to confirm all possible scenarios.

linux_os/guide/services/xwindows/disabling_xwindows/xwindows_runlevel_target/ansible/shared.yml
redhatrises: No. Ansible expects you to use the `file` module here. It's an ansible ism that users will expect.

linux_os/guide/services/xwindows/disabling_xwindows/package_xorg-x11-server-common_removed/rule.yml
ggbecker: It seems to be `base-x` they have similar package list, from RHEL8 machine:

```
[root@localhost gen]# yum group info base-x
Updating Subscription Management repositories.
Unable to read consumer identity
This system is not registered to Red Hat Subscription Management. You can use subscription-manager to register.
Last metadata expiration check: 0:04:26 ago on Wed 22 Apr 2020 11:55:27 AM EDT.

Group: base-x
 Description: Local X.org display server
 Mandatory Packages:
   glx-utils
   mesa-dri-drivers
   plymouth-system-theme
   xorg-x11-drv-ati
   xorg-x11-drv-evdev
   xorg-x11-drv-fbdev
   xorg-x11-drv-intel
   xorg-x11-drv-libinput
   xorg-x11-drv-nouveau
   xorg-x11-drv-qxl
   xorg-x11-drv-vesa
   xorg-x11-drv-vmware
   xorg-x11-drv-wacom
   xorg-x11-server-Xorg
   xorg-x11-utils
   xorg-x11-xauth
   xorg-x11-xinit
   xorg-x11-xinit-session
```

From RHEL7 machine:
```
[root@localhost ~]# yum group info 'X Window System'
Loaded plugins: product-id, search-disabled-repos, subscription-manager
This system is not registered with an entitlement server. You can use subscription-manager to register.
There is no installed groups file.
Maybe run: yum groups mark convert (see man yum)

Group: X Window System
 Group-Id: x11
 Description: X Window System Support.
 Mandatory Packages:
    glx-utils
   +initial-setup-gui
    mesa-dri-drivers
   +plymouth-system-theme
   +spice-vdagent
   +xorg-x11-drivers
   +xorg-x11-server-Xorg
   +xorg-x11-utils
    xorg-x11-xauth
    xorg-x11-xinit
   +xvattr
 Optional Packages:
   mesa-libGLES
   tigervnc-server
   wayland-protocols-devel
   xorg-x11-drv-evdev
   xorg-x11-drv-keyboard
   xorg-x11-drv-libinput
   xorg-x11-drv-mouse
   xorg-x11-drv-openchrome
```

Funny thing is that neither of those groups include `xorg-x11-server-common` package.
Removing this package manually remove many other dependencies: https://pastebin.com/yd1LwvyZ

ggbecker: I selected the package group `base-x` for RHEL8 based products.

matejak:
The PR attempts to solve the issue that if we remove the Xorg server package while leaving the boot target to GUI boot, the system will end up in a state in which it is not functional. The problem is that the STIG requires removal of the X server, while it doesn't say anything about the default boot target.

There are obviously these solutions available:
1. Don't do anything - it's not a bug, it's a STIG feature.
2. Don't really remove the X server, just let the scan fail, and let this dangerous and potentially destructive operation be performed manually by a skilled admin.
3. Introduce a convenience rule that would set the default boot target to command-line, and select it in this (and other affected) profile.
4. Make sure that the boot target is set to command-line right after the X server package is removed.

Each of those approaches got some kind of pushback, and @ggbecker basically went from scenario 2 to 4, but the pushback was unceasing. We are at risk of ending up with scenario 1, unless we reach a conclusion.

This situation is analogous with the usbguard hardening in e.g. RHEL8 OSPP profile, where we include a convenience rule that configures usbguard to allow a wide range of harmless and essential devices: https://github.com/ComplianceAsCode/content/pull/5055 What are again the reasons why we can't repeat this in the STIG profile? I can't find objections in the PR discussion, but there must have been some.

linux_os/guide/services/xwindows/disabling_xwindows/xwindows_runlevel_setting/oval/shared.xml
ggbecker: I have extended the check to allow both locations since both ways work. I have tested in RHEL7.1 and RHEL7.7 the behavior and on a fresh installed system the target points to `/lib` but as soon as the default target is changed by `systemctl` it points to `/usr`

```
[0 root@ci-vm-10-0-137-213 ~]# cat /etc/redhat-release 
Red Hat Enterprise Linux Server release 7.1 (Maipo)
[0 root@ci-vm-10-0-137-213 ~]# systemctl get-default
multi-user.target
[0 root@ci-vm-10-0-137-213 ~]# ls -la /etc/systemd/system/default.target
lrwxrwxrwx. 1 root root 37 Jul 17  2014 /etc/systemd/system/default.target -> /lib/systemd/system/multi-user.target
[0 root@ci-vm-10-0-137-213 ~]# systemctl set-default multi-user.target
rm '/etc/systemd/system/default.target'
ln -s '/usr/lib/systemd/system/multi-user.target' '/etc/systemd/system/default.target'
[0 root@ci-vm-10-0-137-213 ~]# ls -la /etc/systemd/system/default.target
lrwxrwxrwx. 1 root root 41 Apr 23 11:02 /etc/systemd/system/default.target -> /usr/lib/systemd/system/multi-user.target
[0 root@ci-vm-10-0-137-213 ~]# systemctl get-default
multi-user.target






[root@ci-vm-10-0-136-215 ~]# cat /etc/redhat-release 
[root@ci-vm-10-0-136-215 ~]# systemctl get-default
multi-user.target
[root@ci-vm-10-0-136-215 ~]# ls -la /etc/systemd/system/default.target
lrwxrwxrwx. 1 root root 37 Aug  6  2019 /etc/systemd/system/default.target -> /lib/systemd/system/multi-user.target
[root@ci-vm-10-0-136-215 ~]# systemctl set-default multi-user.target
Removed symlink /etc/systemd/system/default.target.
Created symlink from /etc/systemd/system/default.target to /usr/lib/systemd/system/multi-user.target.
[root@ci-vm-10-0-136-215 ~]# ls -la /etc/systemd/system/default.target
lrwxrwxrwx. 1 root root 41 Apr 23 11:07 /etc/systemd/system/default.target -> /usr/lib/systemd/system/multi-user.target
[root@ci-vm-10-0-136-215 ~]# systemctl get-default
multi-user.target
```

linux_os/guide/services/xwindows/disabling_xwindows/package_xorg-x11-server-common_removed/rule.yml
ggbecker: done

linux_os/guide/services/xwindows/disabling_xwindows/xwindows_runlevel_target/ansible/shared.yml
ggbecker: done

linux_os/guide/services/xwindows/disabling_xwindows/package_xorg-x11-server-common_removed/ansible/shared.yml
ggbecker: reverted

linux_os/guide/services/xwindows/disabling_xwindows/package_xorg-x11-server-common_removed/bash/shared.sh
ggbecker: reverted

linux_os/guide/services/xwindows/disabling_xwindows/xwindows_runlevel_setting/oval/shared.xml
ggbecker: I confirm that `systemctl set-default` will prefer to user files from `/usr`. Fresh installed system has target pointing to `/lib`. So I updated the check to cover both cases.

ggbecker:
> The PR attempts to solve the issue that if we remove the Xorg server package while leaving the boot target to GUI boot, the system will end up in a state in which it is not functional. The problem is that the STIG requires removal of the X server, while it doesn't say anything about the default boot target.
> 
> There are obviously these solutions available:
> 
>     1. Don't do anything - it's not a bug, it's a STIG feature.
> 
>     2. Don't really remove the X server, just let the scan fail, and let this dangerous and potentially destructive operation be performed manually by a skilled admin.
> 
>     3. Introduce a convenience rule that would set the default boot target to command-line, and select it in this (and other affected) profile.
> 
>     4. Make sure that the boot target is set to command-line right after the X server package is removed.
> 
> 
> Each of those approaches got some kind of pushback, and @ggbecker basically went from scenario 2 to 4, but the pushback was unceasing. We are at risk of ending up with scenario 1, unless we reach a conclusion.
> 
> This situation is analogous with the usbguard hardening in e.g. RHEL8 OSPP profile, where we include a convenience rule that configures usbguard to allow a wide range of harmless and essential devices: #5055 What are again the reasons why we can't repeat this in the STIG profile? I can't find objections in the PR discussion, but there must have been some.

I reverted back the custom remediation for `xwindows_runlevel_target` and chose scenario 3. There were a few discoveries in the process and they are also fixed, for example the OVAL check for this new rule considers target files from both locations: `/lib/systemd/system/multi-user.target` and `/usr/lib/systemd/system/multi-user.target`. I have tested a RHEL7 installation using the datastream containing this new rule (selecting STIG profile) and the system ends with a terminal console as expected instead of a black screen.

linux_os/guide/services/xwindows/disabling_xwindows/xwindows_runlevel_target/bash/shared.sh
ggbecker: Tested RHEL7 installation providing the datastream through a webserver and customizing the kickstart to point to the custom datastream and it worked perfectly. The system ends up on a regular terminal login console

ggbecker:
@openscap-ci test this please

linux_os/guide/services/xwindows/disabling_xwindows/xwindows_runlevel_target/bash/shared.sh
yuumasato: That's great!

ggbecker:
@openscap-ci test this please

matejak:
test this please

rhel7/profiles/stig.profile
ggbecker: The rule has been removed from rule selection list

matejak:
I am merging this PR, as all reviewer comments were addressed.
This PR improves the `` rule by adding checks/remediations and tests. This PR doesn't alter in STIG profiles, for that, refer to https://github.com/ComplianceAsCode/content/pull/5659. Only C2S and CIS profiles now select the rule.

rhel7/profiles/C2S.profile
```diff
@@ -78,6 +78,7 @@ selections:
     - service_xinetd_disabled
     - service_chronyd_or_ntpd_enabled
     - package_xorg-x11-server-common_removed
+    - xwindows_runlevel_target
```
shawndwells, line 81: Change reverted in https://github.com/ComplianceAsCode/content/commit/ab0a9490dcab448372b93e1364b5b4570e4d6630

PR #5637, opened 2020-04-17
Fix malformed MachineConfigs

JAORMX:
Some audit related Machineconfig objects were introduced as ignition
remediations. They were malformed. This fixes that.

This makes the unit tests pass. Which can be verified as follows:

    make -f tests/unit/kubernetes/Makefile unit

These tests were recently introduced in the following PR: https://github.com/ComplianceAsCode/content/pull/5636


shawndwells:
Crap, that's on me. Will go pack and check these.

PR #5642, opened 2020-04-17
Fix audit_basic_configuration ignition remediation

jhrozek:
#### Description:

The ignition remediation when applied was still failing the rule check
becuase the underlying oval check compares the file with a precise
expected value and the remediation had an extra newline at the end.

Also, I added ocp4 as a product type. I'm not sure if it's needed,
but since ignition is mostly used in OCP4, I think it makes sense if
the rule is usable for the OCP4 product.

#### Rationale:

- the ignition remediation, while technically correct, was still causing
the rule to fail when applied.

### DIFF
--- a/linux_os/guide/system/auditing/policy_rules/audit_basic_configuration/ignition/shared.yml
+++ b/linux_os/guide/system/auditing/policy_rules/audit_basic_configuration/ignition/shared.yml
@@ -8,7 +8,7 @@ spec:
     storage:
       files:
       - contents:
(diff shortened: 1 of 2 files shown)


linux_os/guide/system/auditing/policy_rules/audit_basic_configuration/ignition/shared.yml
```diff
@@ -1,4 +1,4 @@
-# platform = multi_platform_rhel,multi_platform_fedora,multi_platform_ocp
+# platform = multi_platform_ocp
```
shawndwells, line 1: These are still needed for RHEL worker nodes. Trying to be a good upstream OKD steward and added Fedora too. The ignition scripts won't get added to the RHEL and Fedora datastreams without these lines.

jhrozek: Thank you, makes sense. Fixed.

PR #5648, opened 2020-04-21
ocp4: Expand unit tests to validate profile selections

JAORMX:
Some checks are not valid for OCP. Let's start checking those and
erroring out with a reasonable message so folks don't try to enable
them.

As a result:

* This comments out the check related to user namespaces in the coreos-ncp profile, since this isn't a valid test for OCP (It'll cripple the deployment).

### DIFF
--- a/ocp4/profiles/coreos-ncp.profile
+++ b/ocp4/profiles/coreos-ncp.profile
@@ -148,9 +148,6 @@ selections:
     - sysctl_kernel_kexec_load_disabled
     - sysctl_kernel_yama_ptrace_scope
     - sysctl_kernel_perf_event_paranoid
-    - sysctl_user_max_user_namespaces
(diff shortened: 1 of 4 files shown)


ocp4/profiles/coreos-ncp.profile
```diff
@@ -148,9 +148,9 @@ selections:
     - sysctl_kernel_kexec_load_disabled
     - sysctl_kernel_yama_ptrace_scope
     - sysctl_kernel_perf_event_paranoid
-    - sysctl_user_max_user_namespaces
-    - sysctl_user_max_user_namespaces.role=unscored
-    - sysctl_user_max_user_namespaces.severity=info
+    #- sysctl_user_max_user_namespaces
```
jhrozek, line 151: I would outright remove the rules or at least added a comment why are the rules disabled

PR #5651, opened 2020-04-21
Fix several audit-related ignition remediations

jhrozek:
#### Description:

The OVAL checks use the filetextcontent54 probe and check against a blob
defined in the DS, so any divergence from the blob fails the rule. In these
remediations, the ignition files rendered trailing whitespace. Also,
some characters like ">=" were already encoded when the resulting file
was urlencoded.

#### Rationale:

Applying the remediations was still resulting in a failed check during a
subsequent run.


shawndwells:
@jhrozek how are you URL encoding the data?

We used the python snippet from @JAORMX's blog at https://jaosorior.dev/2019/modifying-node-configurations-in-openshift-4.x/:

````bash
$ cat "string to encode" | python3 -c "import sys, urllib.parse; print(urllib.parse.quote(''.join(sys.stdin.readlines())))"
````

**Comment by shawndwells** [comment]
@openscap-ci retest this please

**Comment by jhrozek** [comment]
> @jhrozek how are you URL encoding the data?
> 
> We used the python snippet from @JAORMX's blog at https://jaosorior.dev/2019/modifying-node-configurations-in-openshift-4.x/:
> 
> ```shell
> $ cat "string to encode" | python3 -c "import sys, urllib.parse; print(urllib.parse.quote(''.join(sys.stdin.readlines())))"
> ```

For quick and dirty one-liners I'm using a script from the CaC repo:
```
$ python3 utils/ignition-remediation.py encode --mode=0600 --infile=/tmp/etc/audit/rules.d/30-ospp-v42.rules --target=/etc/audit/rules.d/30-ospp-v42.rules --rule=audit_ospp_general --platform=multi_platform_rhel,multi_platform_fedora,multi_platform_ocp
```

I've been also developing this other tool where you keep a chroot-like tree of all files you want to encode together with metadata in the files about which rule they apply to and then you just run a single command which generates all the ignitions for you. This way, you can view, grep and inspect the raw files and also it's easier to then keep tabs on a single file used for multiple rules. I'm not sure if it's usable for anyone else, the tool is definitely very raw, but it's been handy for me.

**Comment by jhrozek** [comment]
btw for review of the PR, the `ingition-remediation.py` script can also be used. e.g. calling:
```
python3 utils/ignition-remediation.py decode --rule=audit_modify_failed
```
before and after the applying the patch could clearly show what are the differences.

shawndwells:
> > @jhrozek how are you URL encoding the data?
> > We used the python snippet from @JAORMX's blog at https://jaosorior.dev/2019/modifying-node-configurations-in-openshift-4.x/:
> > ```shell
> > $ cat "string to encode" | python3 -c "import sys, urllib.parse; print(urllib.parse.quote(''.join(sys.stdin.readlines())))"
> > ```
> 
> For quick and dirty one-liners I'm using a script from the CaC repo:
> 
> ```
> $ python3 utils/ignition-remediation.py encode --mode=0600 --infile=/tmp/etc/audit/rules.d/30-ospp-v42.rules --target=/etc/audit/rules.d/30-ospp-v42.rules --rule=audit_ospp_general --platform=multi_platform_rhel,multi_platform_fedora,multi_platform_ocp
> ```
> 
> I've been also developing this other tool where you keep a chroot-like tree of all files you want to encode together with metadata in the files about which rule they apply to and then you just run a single command which generates all the ignitions for you. This way, you can view, grep and inspect the raw files and also it's easier to then keep tabs on a single file used for multiple rules. I'm not sure if it's usable for anyone else, the tool is definitely very raw, but it's been handy for me.

Thanks! There doesn't seem to be mention of this in the docs. Will work on adding it.

PR #5652, opened 2020-04-21
remove remediations for configure_etc_hosts_deny

vojtapolasek:
#### Description:

Remove remediations, add warning and improve description to explain how to remediate the rule manually.

#### Rationale:

If default /etc/hosts.allow is used, the remediation of this rule block SSH access to the computer, effectively preventing management. This happens actually right after installation. Therefore, I suggest that the rule should not be remediated automatically. I added instructions for manual remediation into the description and management warning.

### DIFF
--- a/linux_os/guide/services/obsolete/inetd_and_xinetd/configure_etc_hosts_deny/rule.yml
+++ b/linux_os/guide/services/obsolete/inetd_and_xinetd/configure_etc_hosts_deny/rule.yml
@@ -36,8 +36,7 @@ ocil: |-
 
 warnings:
     - functionality: |-
-        This rule affects all access to serviceswhich honor <tt>/etc/hosts.allow</tt> and <tt>/etc/hosts.deny</tt> files.
+        This rule affects all access to services which honor <tt>/etc/hosts.allow</tt> and <tt>/etc/hosts.deny</tt> files.
         Connections to services originating from hosts not explicitly mentioned in <tt>/etc/hosts.allow</tt> will be rejected.
-        As the <tt>/etc/hosts.allow</tt> is empty by default, make sure it is appropriately configured before applying remediation for this rule.
         To avoid locking down all network access to the system, this rule doesn't perform automated remediation.
         For information about manual process of remediation see the rule description.
--- a/linux_os/guide/services/obsolete/inetd_and_xinetd/configure_etc_hosts_deny/rule.yml
+++ b/linux_os/guide/services/obsolete/inetd_and_xinetd/configure_etc_hosts_deny/rule.yml
@@ -12,7 +12,7 @@ description: |-
     <pre>ALL: ALL</pre>
     It is advised to inspect available network services which might be affected by modification of file mentioned above prior to performing the remediation of this rule.
     If there exist services which might be affected and access to them should not be blocked,
-    modify the <tt>/etc/hosts.deny</tt> file appropriately before performing the remediation.
+    modify the <tt>/etc/hosts.allow</tt> file appropriately before performing the remediation.
 
 
 rationale: |-
@@ -35,9 +35,9 @@ ocil: |-
     <pre>ALL: ALL</pre>
 
 warnings:
-    - management: |-
-        enabling this rule affects all connections to serviceswhich honor <tt>/etc/hosts.allow</tt> and <tt>/etc/hosts.deny</tt> files.
-        Connections to such servicesfrom any hosts which are not explicitly mentioned in <tt>/etc/hosts.allow</tt> will be rejected.
-        As the <tt>/etc/hosts.allow</tt> file is often left empty, there is a chance that remediation of this rule might prevent the system from accepting SSH connections and therefore limiting management access.
-        Therefore, this rule will not be remediated automatically. For information about manual process
-        of remediation see the rule description.
+    - functionality: |-
+        This rule affects all access to serviceswhich honor <tt>/etc/hosts.allow</tt> and <tt>/etc/hosts.deny</tt> files.
+        Connections to services originating from hosts not explicitly mentioned in <tt>/etc/hosts.allow</tt> will be rejected.
+        As the <tt>/etc/hosts.allow</tt> is empty by default, make sure it is appropriately configured before applying remediation for this rule.
+        To avoid locking down all network access to the system, this rule doesn't perform automated remediation.
+        For information about manual process of remediation see the rule description.
--- a/linux_os/guide/services/obsolete/inetd_and_xinetd/configure_etc_hosts_deny/rule.yml
+++ b/linux_os/guide/services/obsolete/inetd_and_xinetd/configure_etc_hosts_deny/rule.yml
@@ -10,6 +10,10 @@ description: |-
     The following line in the file ensures that access to services supporting this mechanism is denied to any clients
     not mentioned in <tt>/etc/hosts.allow</tt>:
     <pre>ALL: ALL</pre>
+    It is advised to inspect available network services which might be affected by modification of file mentioned above prior to performing the remediation of this rule.
+    If there exist services which might be affected and access to them should not be blocked,
+    modify the <tt>/etc/hosts.deny</tt> file appropriately before performing the remediation.
+
 
 rationale: |-
     Correct configuration in <tt>/etc/hosts.deny</tt> ensures that no explicitly mentioned clients will be able to connect to services supporting this access control mechanism.
@@ -29,3 +33,11 @@ ocil: |-
     <pre>cat /etc/hosts.deny</pre>
     Verify that the output contains the following line:
     <pre>ALL: ALL</pre>
+
+warnings:
+    - management: |-
+        enabling this rule affects all connections to serviceswhich honor <tt>/etc/hosts.allow</tt> and <tt>/etc/hosts.deny</tt> files.
+        Connections to such servicesfrom any hosts which are not explicitly mentioned in <tt>/etc/hosts.allow</tt> will be rejected.
+        As the <tt>/etc/hosts.allow</tt> file is often left empty, there is a chance that remediation of this rule might prevent the system from accepting SSH connections and therefore limiting management access.
+        Therefore, this rule will not be remediated automatically. For information about manual process
+        of remediation see the rule description.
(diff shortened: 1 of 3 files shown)


redhatrises:
Sounds like the profile needs to be tailored rather than removing remediations. Not everyone wants or uses or needs SSH access to be enabled.

vojtapolasek:
I understand your point. However, where should we document that so that admins notice and are not surprised that hardened machine is not accessible through SSH right after installation?

yuumasato:
> Sounds like the profile needs to be tailored rather than removing remediations.

I think the concern here is about the experience of someone just taking the profile, remediating, and then not having remote access to the system, even on a test system. For this not to happen `/etc/hosts.allow` would have to be configured too. And what hosts would be allowed?

yuumasato:
>  For this not to happen `/etc/hosts.allow` would have to be configured too. And what hosts would be allowed?

Actually, would be acceptable if a rule that configures `/etc/hosts.allow` was parametrized by a variable, and the default value was `sshd: ALL`?

redhatrises:
There should be a separate rule that configures `/etc/hosts.allow` to configure allowing SSH.

linux_os/guide/services/obsolete/inetd_and_xinetd/configure_etc_hosts_deny/rule.yml
```diff
@@ -10,6 +10,10 @@ description: |-
     The following line in the file ensures that access to services supporting this mechanism is denied to any clients
     not mentioned in <tt>/etc/hosts.allow</tt>:
     <pre>ALL: ALL</pre>
+    It is advised to inspect available network services which might be affected by modification of file mentioned above prior to performing the remediation of this rule.
+    If there exist services which might be affected and access to them should not be blocked,
+    modify the <tt>/etc/hosts.deny</tt> file appropriately before performing the remediation.
```
yuumasato, line 15: Shouldn't `/etc/hosts.allow` be the file modified to grant access?

Suggestion: `Every service that is essential for your organization should be allowed access before applying remediation for this rule.
Edit <tt>/etc/hosts.allow</tt> and add the allowed services and hosts.`

shawndwells:
> > For this not to happen `/etc/hosts.allow` would have to be configured too. And what hosts would be allowed?
> 
> Actually, would be acceptable if a rule that configures `/etc/hosts.allow` was parametrized by a variable, and the default value was `sshd: ALL`?

Most profiles have rules that prohibit /etc/hosts.allow. What is the underlying issue this PR helps resolve? Creating a warning certainly seems reasonable though.

redhatrises:
> > > For this not to happen `/etc/hosts.allow` would have to be configured too. And what hosts would be allowed?
> > 
> > 
> > Actually, would be acceptable if a rule that configures `/etc/hosts.allow` was parametrized by a variable, and the default value was `sshd: ALL`?
> 
> Most profiles have rules that prohibit /etc/hosts.allow. What is the underlying issue this PR helps resolve? Creating a warning certainly seems reasonable though.

If using a `/etc/hosts.deny` with `ALL: ALL`, you kinda need `/etc/hosts.allow` if you are going to allow a service, say SSH, through. Really configuring `hosts.allow` and `hosts.deny` just needs to go away in profiles... which luckily they are.

redhatrises:
Personally fine with dropping this rule completely.

vojtapolasek:
OK I see five options here:
1. keeping the rule as it is, but without tayloring, it will block SSH and possibly more network services directly after remediation. I don't think this is a good state. However, if we decide to use it, we should clearly and visibly document somewhere that without tayloring this will happen.
2. Drop remediations and keep the rule with warning and a good description, so that if someone desires to use this rule they have chance to modify /etc/hosts.allow appropriately and then remediate /etc/hosts.deny. That's what this PR is trying to achieve.
3. Drop only the Anaconda remediation, so that the rule will fail after installation, but the machine won't lose any network connections. I assume that if a user sees failing rule, they read warnings and description and they can make appropriate modifications to /etc/hosts.allow before remediating this rule.
4. Add a new rule / change this rule so that it adds some exceptions to /etc/hosts.allow. This could solve the problem by whitelisting some hosts and services... but which hosts and which services? Can we afford to make such assumptions?
5. Drop the rule completely. I am not sure if this is what we want because the rule is scored in CIS.
Personally I am for number 3.

dahaic:
> OK I see five options here:
> 
>     1. keeping the rule as it is, but without tailoring, it will block SSH and possibly more network services directly after remediation. I don't think this is a good state. However, if we decide to use it, we should clearly and visibly document somewhere that without tailoring this will happen.
> 
>     2. Drop remediations and keep the rule with warning and a good description, so that if someone desires to use this rule they have chance to modify /etc/hosts.allow appropriately and then remediate /etc/hosts.deny. That's what this PR is trying to achieve.
> 
>     3. Drop only the Anaconda remediation, so that the rule will fail after installation, but the machine won't lose any network connections. I assume that if a user sees failing rule, they read warnings and description and they can make appropriate modifications to /etc/hosts.allow before remediating this rule.
> 
>     4. Add a new rule / change this rule so that it adds some exceptions to /etc/hosts.allow. This could solve the problem by whitelisting some hosts and services... but which hosts and which services? Can we afford to make such assumptions?
> 
>     5. Drop the rule completely. I am not sure if this is what we want because the rule is scored in CIS.
>        Personally I am for number 3.

So given this is a rule used in CIS, by dropping the rule, we move away from the alignment. And adding additional rule is not only cumbersome, but also would include stuff that is not supposed to be there. So I agree with the original intention of getting rid of the remediation, with providing explanation why we refrain from doing that.

Preserving SSH access by default is not only increasing usability (we had bugs reported in the past that system is not accessible after installation, and we consider this behavior to be a blocker issue), but it also reduces the costs of development, as having SSH access streamlines our productization mechanisms.

Honestly,  I also don't think this rule is of an importance to warrant this extensive discussions - if it's OK to drop the rule, why it's not ok to drop just the remediation part of the rule?

vojtapolasek:
Just adding that after discussions I realized that the solution 3 is not possible. Therefore I am for 2.

shawndwells:
> So given this is a rule used in CIS, by dropping the rule, we move away from the alignment. And adding additional rule is not only cumbersome, but also would include stuff that is not supposed to be there. So I agree with the original intention of getting rid of the remediation, while providing explanation why we refrain from doing that.
> 
> Preserving SSH access by default is not only increasing usability (we had bugs reported in the past that system is not accessible after installation, and we consider this behavior to be a blocker issue), but it also reduces the costs of development, as having SSH access streamlines our productization mechanisms.
> 
> Honestly, I also don't think this rule is of an importance to warrant this extensive discussions - if it's OK to drop the rule, why it's not ok to drop just the remediation part of the rule?

There's no need to blindly follow random third party recommendations.

@dahaic why don't you work with CIS to remove the rule from their content? Work upstream first, so to say.

dahaic:
> > So given this is a rule used in CIS, by dropping the rule, we move away from the alignment. And adding additional rule is not only cumbersome, but also would include stuff that is not supposed to be there. So I agree with the original intention of getting rid of the remediation, while providing explanation why we refrain from doing that.
> > Preserving SSH access by default is not only increasing usability (we had bugs reported in the past that system is not accessible after installation, and we consider this behavior to be a blocker issue), but it also reduces the costs of development, as having SSH access streamlines our productization mechanisms.
> > Honestly, I also don't think this rule is of an importance to warrant this extensive discussions - if it's OK to drop the rule, why it's not ok to drop just the remediation part of the rule?
> 
> There's no need to blindly follow random third party recommendations.
> 
> @dahaic why don't you work with CIS to remove the rule from their content? Work upstream first, so to say.

Well, if we want to claim this profile follows that particular random third party recommendation, then yes, the need is there.

Though you have a point that we should talk to CIS and try to align it better with our understanding. As usual, discussions like that might take time, and we would like to be done with the basic CIS support as soon as possible. So yeah, we will do a round of polish later on, but for now we are sticking to what has been recommended.

So I believe the right approach is to remove the remediations,  and iterate later on, by either adding more rules, or by dropping the rule altogether.

linux_os/guide/services/obsolete/inetd_and_xinetd/configure_etc_hosts_deny/rule.yml
```diff
@@ -29,3 +33,11 @@ ocil: |-
     <pre>cat /etc/hosts.deny</pre>
     Verify that the output contains the following line:
     <pre>ALL: ALL</pre>
+
+warnings:
+    - functionality: |-
+        This rule affects all access to serviceswhich honor <tt>/etc/hosts.allow</tt> and <tt>/etc/hosts.deny</tt> files.
```
yuumasato, line 39: Missing space in `serviceswhich`.

linux_os/guide/services/obsolete/inetd_and_xinetd/configure_etc_hosts_deny/rule.yml
```diff
@@ -10,6 +10,10 @@ description: |-
     The following line in the file ensures that access to services supporting this mechanism is denied to any clients
     not mentioned in <tt>/etc/hosts.allow</tt>:
     <pre>ALL: ALL</pre>
+    It is advised to inspect available network services which might be affected by modification of file mentioned above prior to performing the remediation of this rule.
+    If there exist services which might be affected and access to them should not be blocked,
+    modify the <tt>/etc/hosts.allow</tt> file appropriately before performing the remediation.
+
```
yuumasato, line 16: Isn't this part redundant with the warning?

vojtapolasek: It is similar, but does it hurt? Are the warnings and descriptions visible in similar way in terms of visual style? I just don't want users to miss the information that if they want to allow some services, they have to place them into hosts.allow.

yuumasato: >  if they want to allow some services, they have to place them into hosts.allow.

I believe this is already covered in the description, starting from line 10.

yuumasato: I see you removed it from the warning, I'm fine with that.

yuumasato:
> So I believe the right approach is to remove the remediations, and iterate later on, by either adding more rules, or by dropping the rule altogether.

I share the same opinion here.

> Though you have a point that we should talk to CIS and try to align it better with our understanding.

Also agree with iterating with CIS on better aligning the benchmark, but current effort is focused on quickly bring a profile up to fill the gap.

PR #5655, opened 2020-04-22
Add DISA STIG OL7 v1r1 stig_overlay

iokomin:
#### Description:

Add DISA STIG OL7 v1r1 stig_overlay support including stig html tables

#### Testing:
- checked generated stig_overlay.xml: matches STIG ids, SCAP rule_id
- checked generated table-ol7-stig.html to match ids and descriptions


iokomin:
@shawndwells thanks for reviewing this PR!
Overlay and generated ref tables look extremely useful to understand coverage and binding to know security controls. I just have couple of questions on result html and stig-overlay formats, would scap-security-guide mail group be the right place for this discussion?

PR #5656, opened 2020-04-23
use freq param for both 'incremental' and 'incremental_async' of var_…

enjaygee:
…auditd_flush

#### Description:

-  This PR modifies the associated auditd bash remediation rule to use the freq parameter along with an "incremental_async" auditd flush value. 

#### Rationale:

- As noted in the RHEL 7 security guide (https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/security_guide/sec-configuring_the_audit_service), a flush value of "incremental_async" should be used in combination with the freq parameter.
- Current code comments out the freq parameter for "incremental_async", which generates the error "Error - incremental flushing chosen, but 0 selected for freq" in auditd sanity_check (see https://github.com/linux-audit/audit-userspace/blob/master/src/auditd-config.c)


jan-cerny:
@openscap-ci ok to test

jan-cerny:
@openshift-ci-robot 
/ok-to-test

jan-cerny:
great! thanks

PR #5657, opened 2020-04-23
Warn about only local user backends being considered

matejak:
Rules that deal with uids/gids don't take into account remote user backends s.a. AD or general LDAP setups.

In the ideal case, the scanner and the content would work together to work even with remote users, but as it is not the case, let's at least display a warning.

As more rule may be affected by this, I have made it a macro.


shared/macros.jinja
```diff
@@ -15,6 +15,30 @@ ocil_clause: "the required value is not set"
 {{%- endmacro %}}
 
 
+{{% macro local_users_considered_only() -%}}
+This rule only considers local users.
+If you have your users defined outside <code>/etc/passwd</code>, the rule won't consider those.
```
redhatrises, line 20: Lines 21-26 aren't need and are very wordy. 19-20 are really good and sufficient. Drop lines 21-26.

matejak:
OK, no problem. Fixed by a rebase.

redhatrises:
Looks good! Thank you very much!

PR #5660, opened 2020-04-24
Reference should not point to OS version.

adelton:
#### Description:

- Reference should not point to OS version.

#### Rationale:

- All other references point to standards, not to products.


redhatrises:
Great Catch! Thanks @adelton ! Ack

PR #5691, opened 2020-04-25
revert platform:machine for service_cockpit_disabled

shawndwells:
Services run inside containers

eg
`````
# cat Dockerfile

FROM fedora

RUN dnf -y install cockpit; dnf clean all; systemctl enable cockpit

EXPOSE 80

CMD [ "/sbin/init" ]
`````

see also https://developers.redhat.com/blog/2019/04/24/how-to-run-systemd-in-a-container/


redhatrises:

PR #5693, opened 2020-04-25
revert platform:machine on chronyd_client_only

shawndwells:
This rule is for a configuration check. What if the container is running a chrony service? Needs to ensure it's configured properly.


redhatrises:

PR #5694, opened 2020-04-25
revert platform:machine on chronyd_no_chronyc_network

shawndwells:
This check evaluates a configuration item. What if chrony is running inside a container? Need to ensure it is configured properly.


redhatrises:

PR #5695, opened 2020-04-25
revert platform:machine for chronyd_or_ntpd_set_maxpoll

shawndwells:
Need to ensure chrony configuration is properly set, even when running in containers.


redhatrises:

PR #5696, opened 2020-04-25
revert platform:machine in chronyd_or_ntpd_specify_multiple_servers

shawndwells:
Need to ensure proper configuration no matter where chrony or ntp is deployed.


redhatrises:

PR #5697, opened 2020-04-25
revert platform:machine for chronyd_or_ntpd_specify_remote_server

shawndwells:
Need to ensure proper config wherever time services are deployed.


redhatrises:

PR #5698, opened 2020-04-25
revert platform:machine in service_chronyd_or_ntpd_enabled

shawndwells:
Containers run services too. Looks like the underlying check needs to be updated, not removed.


redhatrises:

PR #5699, opened 2020-04-25
revert platform:machine for service_sssd_enabled

shawndwells:
containers run services too. Underlying OVAL may need updating, but the rule is still applicable.


redhatrises:

PR #5700, opened 2020-04-25
revert platform:machine for sssd_enable_smartcards

shawndwells:
Need to ensure proper configuration of the service however/wherever it is deployed. Looks like the underlying OVAL needs to be updated - but this rule is still applicable.


redhatrises:

PR #5701, opened 2020-04-25
revert platform:machine on sssd_memcache_timeout

shawndwells:
Need to ensure service is properly configured wherever/however it is deployed.


redhatrises:

PR #5702, opened 2020-04-25
revert platform:machine for sssd_offline_cred_expiration

shawndwells:
Need to ensure service is properly configured however/wherever it is deployed. Looks like underlying OVAL must be updated, but this rule is still applicable.


redhatrises:

PR #5704, opened 2020-04-25
revert platform:machine for service_usbguard_enabled

shawndwells:
Containers run services too. Looks like underlying OVAL needs to be updated, but this rule is still applicable to containers.


redhatrises:

PR #5705, opened 2020-04-25
revert platform:machine for smartcard_auth

shawndwells:
Underlying OVAL needs to be rewritten, but the rule is still applicable to containers.


redhatrises:

PR #5706, opened 2020-04-25
revert platform:machine for network_sniffer_disabled

shawndwells:
Still applicable to containers. Looks like a different OVAL technique must be used.


redhatrises:

PR #5707, opened 2020-04-25
revert platform:machine for service_systemd-coredump_disabled

shawndwells:
Containers run services too. Looks like OVAL technique needs to be updated.


redhatrises:

PR #5650, opened 2020-04-21
fix ansible remediations to avoid creating duplicate entries

vojtapolasek:
#### Description:

Ansible remediations for audit_rules_kernel_module_loading and audit_rules_media_export are modified to check if needed Audit lines already exist. Therefore, they should not create duplicates.
There is also new test checking for that in audit_rules_kernel_module_loading, as the remediation is quite complex.

#### Rationale:

avoid duplicate Audit rules

PR #5709, opened 2020-04-27
Introduce Ansible macros for remediating Audit syscall rules

vojtapolasek:
#### Description:

Macros are based on https://github.com/ComplianceAsCode/content/pull/5650

They should provide similar functionality as Bash remediation functions. They ensure that there are appropriate Audit rules for chosen X86 architecture with defined list of syscalls and with defined key.

#### Rationale:

Improve creation of Ansible remediations for Audit rules.


shared/macros-ansible.jinja
```diff
@@ -346,3 +346,195 @@ The macro requires following parameters:
...
+    audit_arch: "b{{ ansible_architecture | regex_replace('.*(\\d\\d$)','\\1') }}"
+
+- name: Declare list of syscals
+  set_fact:
+    syscalls: {{{ syscalls }}}
+
+- name: Declare number of syscalls
+  set_fact: audit_syscalls_number_of_syscalls="{{ syscalls|length|int }}"
+
+- name: Check existence of syscalls for 32 bit architecture in /etc/audit/rules.d/
```
yuumasato, line 374: The tasks for 32 bits are being included twice. As they are not wrapped in an `arch` conditional, when the call for b64 happens they are added too.

shared/macros-ansible.jinja
```diff
@@ -346,3 +346,195 @@ The macro requires following parameters:
...
+  loop: "{{ syscalls }}"
+
+- name: Get number of matched 32 bit syscalls in /etc/audit/rules.d/
+  set_fact: audit_syscalls_matched_32_rules_d="{{audit_syscalls_found_32_rules_d.results|sum(attribute='matched')|int }}"
+
+{{% if arch == "64" %}}
+- name: Check existence of syscalls for 64 bit architecture in /etc/audit/rules.d/
+  find:
+    paths: "/etc/audit/rules.d"
+    contains: '^[\s]*-a[\s]+always,exit[\s]+(?:.*-F[\s]+arch=b64[\s]+)(?:.*(-S[\s]+{{ item }}[\s]+|([\s]+|[,]){{ item }}([\s]+|[,]))).*(-k[\s]+|-F[\s]+key=)[\S]+[\s]*$'
```
yuumasato, line 389: Instead of having two tasks for "Check existence of syscalls ...", they could be merged into one.
The `arch` parameter can be used to render the actual string, for example `arch=b{{{ arch }}}`, this can also be done for the task names and variables.

shared/macros-ansible.jinja
vojtapolasek: This is done for a reason. Here is an excerpt from man page for audit.rules:
---
When you specify a syscall name, auditctl will look up the name and get its syscall number. This leads to some problems on bi-arch machines. The 32 and 64 bit syscall numbers sometimes, but not always, line up. So, to solve this problem, you would generally need to break the rule into 2 with one specifying −F arch=b32 and the other specifying −F arch=b64. This needs to go in front of the −S option so that auditctl looks at the right lookup table when returning the number.
---
I agree that it is not clear, it deserves some explanation. Of course other suggestions how to solve this are welcomed.

yuumasato: I understand that the remediation needs to add audit rules for both 32 and 64 bit syscalls.
But the original remediation didn't have task "Check existence of syscalls for 32 bit architecture in /etc/audit/rules.d/" twice, as the one generated now.
This is happening because the macro is called twice, but "Check existence of syscalls for 32 bit architecture in /etc/audit/rules.d/" is not wrapped in a conditional.

yuumasato: Another way of handling this would be to have one macro that handles both 32 and 64 bits at once, basically taking the original remediation and parametrizing `syscalls` and `key`.

One benefit of this is that tasks like `Set architecture for audit modules tasks` and `Set architecture for audit modules tasks` won't be duplicated.

JAORMX:
@vojtapolasek note that the openshift CI runs don't run ansible remediations and also don't block your patch from merging. So you could just merge this as-is.

shared/macros-ansible.jinja
```diff
@@ -346,3 +346,189 @@ The macro requires following parameters:
...
+  loop: "{{ syscalls }}"
+
+- name: Get number of matched 64 bit syscalls in /etc/audit/rules.d/
+  set_fact: audit_syscalls_matched_64_rules_d="{{audit_syscalls_found_64_rules_d.results|sum(attribute='matched')|int }}"
+
+- name: Search /etc/audit/rules.d for other rules with the key {{{ key }}}
+  find:
+    paths: "/etc/audit/rules.d"
+    recurse: no
+    contains: '(-F key=)|(-k\s+){{{ key }}}$'
```
yuumasato, line 401: We need test scenarios for the augenrules case.
This regex doesn't seem to be working, the remediation is not adding the rule together with other rules with same key.

yuumasato: Adding anchor to beginning of file and another parenthesis group solved for me:  '^.*((-F key=)|(-k\s+)){{{ key }}}$'

vojtapolasek: Nice catch, thanks. I created test cases for augen-rules. However, I don't think we can test this as I believe Oval doesn't care into which file in rules.d directory the rule is placed.

yuumasato: > However, I don't think we can test this as I believe Oval doesn't care into which file in rules.d directory the rule is placed.

Right, nice catch too, :).
As the remediation is not adding duplicates I don't consider this to be a blocking issue.

yuumasato:
Thank you for all the effort on this!

yuumasato:
@openscap-ci test this please

yuumasato:
Thank you for all the work on the Ansible remediations for audit!

PR #5718, opened 2020-04-29
ocp4: Add missing AC-1 checks to moderate profile

JAORMX:
These checks are enabled for RHEL, but are also valid for RHCOS. So
let's enable these checks.

Note that the sshd-related checks have been commented out since we still
need to come up with a remediation strategy, given that those touch the
same file.


JAORMX:
@jhrozek @mrogers950 PTAL

redhatrises:
LGTM

PR #5722, opened 2020-04-29
Ubuntu1804/cis kernel module rules

willumpie:
#### Description:

Enable kernel module rules for ubuntu 1804 according CIS profile:
cramfs, freevxfs, jffs2, hfs, hfsplus, udf.

Remove tests-criteria for Ubuntu1804:

#### Rationale:
- Adding these rules to complete the cis- hardening profile. 
- Remove testcriteria only for Ubuntu1804, because I only tested it on this version (Desktop version). 

  according the man-pages from ubuntu, /etc/modprobe.conf is still deprecated. But when tested, this rule is still not working. /etc/modprobe.d/[myfilename].conf is still working correctly.


jan-cerny:
@openscap-ci test this please

jan-cerny:
@openscap-ci ok to test

PR #5723, opened 2020-04-29
macro for remediating audit syscall rules requires arch to be specified

vojtapolasek:
#### Description:

Modified macros for ansible remediation of Audit syscalls to require parameter arch to be set.
Rewrite rule audit_rules_kernel_module_loading to use this new way.

#### Rationale:

Currently the macro is "smart" and determines if the rule should be remediated for 32 bit syscalls or both 32 and 64 bit syscalls. Unfortunately, this is not always desired. Fore example the stime syscall is not defined for 64 bit architecture and current macro would therefore produce wrong Audit configuration.

### DIFF
--- a/linux_os/guide/system/auditing/auditd_configure_rules/audit_kernel_module_loading/audit_rules_kernel_module_loading/ansible/shared.yml
+++ b/linux_os/guide/system/auditing/auditd_configure_rules/audit_kernel_module_loading/audit_rules_kernel_module_loading/ansible/shared.yml
@@ -3,11 +3,31 @@
 # strategy = restrict
 # complexity = low
 # disruption = low
+#
+# What architecture are we on?
+#
+- name: Set architecture for audit tasks
+  set_fact:
+    audit_arch: "b{{ ansible_architecture | regex_replace('.*(\\d\\d$)','\\1') }}"
 
-{{% if product == "rhel6" %}}
-{{{ ansible_audit_augenrules_add_syscall_rule(syscalls=["init_module", "delete_module"], key="modules") }}}
-{{{ ansible_audit_auditctl_add_syscall_rule(syscalls=["init_module", "delete_module"], key="modules") }}}
-{{% else %}}
-{{{ ansible_audit_augenrules_add_syscall_rule(syscalls=["init_module", "finit_module", "delete_module"], key="modules") }}}
-{{{ ansible_audit_auditctl_add_syscall_rule(syscalls=["init_module", "finit_module", "delete_module"], key="modules") }}}
-{{% endif %}}
+
+- name: perform remediation of Audit rules for kernel module loading for x86 platform
+  block:
+  {{% if product == "rhel6" %}}
+    {{{ ansible_audit_augenrules_add_syscall_rule(arch="b32", syscalls=["init_module", "delete_module"], key="modules")|indent(4) }}}
+    {{{ ansible_audit_auditctl_add_syscall_rule(arch="b32", syscalls=["init_module", "delete_module"], key="modules")|indent(4) }}}
+  {{% else %}}
+    {{{ ansible_audit_augenrules_add_syscall_rule(arch="b32", syscalls=["init_module", "finit_module", "delete_module"], key="modules")|indent(4) }}}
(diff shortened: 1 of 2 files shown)


shared/macros-ansible.jinja
```diff
@@ -349,49 +349,29 @@ The macro requires following parameters:
...
-    contains: '^[\s]*-a[\s]+always,exit[\s]+(?:.*-F[\s]+arch=b64[\s]+)(?:.*(-S[\s]+{{ item }}[\s]+|([\s]+|[,]){{ item }}([\s]+|[,]))).*(-k[\s]+|-F[\s]+key=)[\S]+[\s]*$'
+    contains: '^[\s]*-a[\s]+always,exit[\s]+(?:.*-F[\s]+arch={{{ arch }}}[\s]+)(?:.*(-S[\s]+{{ item }}[\s]+|([\s]+|[,]){{ item }}([\s]+|[,]))).*(-k[\s]+|-F[\s]+key=)[\S]+[\s]*$'
     patterns: "*.rules"
-  register: audit_syscalls_found_64_rules_d
+  register: audit_syscalls_found_{{{ arch }}}_rules_d
   loop: "{{ syscalls }}"
 
-- name: Get number of matched 64 bit syscalls in /etc/audit/rules.d/
-  set_fact: audit_syscalls_matched_64_rules_d="{{audit_syscalls_found_64_rules_d.results|sum(attribute='matched')|int }}"
+- name: Get number of matched syscalls for architecture {{{ arch }}}in /etc/audit/rules.d/
```
yuumasato, line 373: Missing space after `}}}`

linux_os/guide/system/auditing/auditd_configure_rules/audit_kernel_module_loading/audit_rules_kernel_module_loading/ansible/shared.yml
```diff
@@ -3,11 +3,31 @@
...
+
+- name: Perform remediation of Audit rules for kernel module loading for x86 platform
+  block:
+  {{% if product == "rhel6" %}}
+    {{{ ansible_audit_augenrules_add_syscall_rule(arch="b32", syscalls=["init_module", "delete_module"], key="modules")|indent(4) }}}
+    {{{ ansible_audit_auditctl_add_syscall_rule(arch="b32", syscalls=["init_module", "delete_module"], key="modules")|indent(4) }}}
+  {{% else %}}
+    {{{ ansible_audit_augenrules_add_syscall_rule(arch="b32", syscalls=["init_module", "finit_module", "delete_module"], key="modules")|indent(4) }}}
+    {{{ ansible_audit_auditctl_add_syscall_rule(arch="b32", syscalls=["init_module", "finit_module", "delete_module"], key="modules")|indent(4) }}}
+  {{% endif %}}
```
yuumasato, line 22: The only difference between these calls are the list of syscalls, so this can be simplified by defining a variable and passing it to the macro.

yuumasato: Something like this:
https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/permissions/files/permissions_important_account_files/file_permissions_backup_etc_shadow/rule.yml#L3

shared/macros-ansible.jinja
```diff
@@ -418,7 +418,7 @@ The macro requires following parameters:
 - syscalls: list of syscalls supplied as a list ["syscall1", "syscall2"] etc.
 - key: a key to use as rule identifier.
 #}}
-{{% macro ansible_audit_auditctl_add_syscall_rule(arch="", syscalls=[], key="") -%}}
+{{%- macro ansible_audit_auditctl_add_syscall_rule(arch="", syscalls=[], key="") -%}}
```
yuumasato, line 421: Any reason for the auditctl macro to have the leading dash `-` and augenrules macro to miss it?

vojtapolasek: Thanks, I removed it, I tried to fix something but this approach didn't help and I forgot about it.

PR #5724, opened 2020-04-29
Add audit field to the Ansible syscall macros

yuumasato:
#### Description:

- Allows the macro to handle general audit fields, thus to remediate a varied range of audit syscall rules with or without fields, like `auid`, `exit` or `a`.
  Audit fields reference: https://github.com/linux-audit/audit-documentation/blob/master/specs/fields/field-dictionary.csv

#### Note:
- ~~This needs to be rebased over #5723~~ Rebased


vojtapolasek:
Thank you very much for this addition. Macros are very flexible now. I ran tests for audit_rules_time_clock_settime and also audit_rules_kernel_module_loading and nothing is broken.

yuumasato:
Thanks for the review!

PR #5729, opened 2020-05-04
kube/ignition: Add service_disabled template definition

JAORMX:
This adds a template definition for the kubernetes and ignition
remediations. This results in having fixes generated automatically for
disabled services.


matejak:
test this please

ssg/templates.py
```diff
@@ -286,7 +286,7 @@ def sebool(data, lang):
     return data
 
 
-@template(["ansible", "bash", "oval", "puppet"])
+@template(["ansible", "bash", "oval", "puppet", "ignition", "kubernetes"])
```
jan-cerny, line 289: This needs to be added to the documentation in docs/manual/developer_guide.adoc line 1622.

shared/templates/template_IGNITION_service_disabled
```diff
@@ -0,0 +1,23 @@
...
+    ignition:
+      version: 2.2.0
+    systemd:
+      units:
+      - name: {{{ DAEMONNAME }}}.service
+        enabled: false 
+{{%- if MASK_SERVICE %}}
+        mask: true
+{{%- endif %}}
+      - name: {{{ DAEMONNAME }}}.socket
```
jan-cerny, line 19: What will happen if the socket file doesn't exist?

shared/templates/template_IGNITION_service_disabled
```diff
@@ -0,0 +1,23 @@
+# platform = multi_platform_rhel,multi_platform_fedora,multi_platform_ocp
+# reboot = true
```
jan-cerny, line 2: Is a reboot needed?

JAORMX: Yes. It happens whenever you add/modify a MachineConfig. The machine-config-operator does that.

shared/templates/template_IGNITION_service_disabled
JAORMX: AFAIU, it shouldn't be a problem as it should just remove the file in the targets directory. Or if "masked" is enabled, it should just point it to /dev/null

ssg/templates.py
JAORMX: Done

jan-cerny:
Thank you very much!

PR #5734, opened 2020-05-04
Added warning about profile not working with GUI systems.

matejak:
Right now, if one selects the Server With GUI RHEL7 Software Selection, the profile boots into a defective state - see https://bugzilla.redhat.com/show_bug.cgi?id=1648162
It is not easy to fix this, as adding convenience selections to the profile is a tricky thing, so let's at least warn users in the description.


redhatrises:
Closing as per our discussion we are handling this in the rule itself, a KBase article, and a ticket with DISA.

PR #5743, opened 2020-05-07
OCP4: Add remediation equality unit tests

JAORMX:
The tests check that remediations don't change files in inconsistent ways. This way we can keep a golden configuration for a specific service, and so remediations will work in the way that machine-config-operator expects.


jan-cerny:
@openscap-ci test this please

PR #5744, opened 2020-05-07
Provide --dry-run option to the Ansible Role generator script

isimluk:
Also, rename the script. This script does not touch Ansible Galaxy and thus it shall perhaps not have the galaxy word in its name.


mildas:
**Changes identified:**
Others:
 Python abstract syntax tree change found in utils/ansible_playbook_to_role.py. - ctest will be selected.

 (cd build && ctest -j4)

PR #5745, opened 2020-05-07
Validate & Test Ansible Roles

isimluk:
Builds ansible roles and runs `ansible-lint` on each.

This avoids us releasing versions that fail on https://galaxy.ansible.com/RedHatOfficial

### DIFF
--- a/.github/workflows/validate_ansible_roles.yaml
+++ b/.github/workflows/validate_ansible_roles.yaml
@@ -1,0 +1,26 @@
+name: Validate Ansible Roles
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+jobs:
+  validate:
+    name: Validate Ansible Roles
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Install Deps
+        uses: mstksg/get-package@master
+        with:
+                apt-get: cmake ninja-build libopenscap8 libxml2-utils expat xsltproc python3-jinja2 python3-yaml ansible-lint python3-github
+      - name: Checkout
+        uses: actions/checkout@v1
+      - name: Build
+        run: |
+          ./build_product rhel7 rhel8 rhv4
+      - name: Build Ansible Roles
+        run: |
+          PYTHONPATH=. python3 utils/ansible_playbook_to_role.py --build-playbooks-dir ./build/ansible/ --dry-run ./build/ansible_roles
+      - name: Lint Ansible Roles
+        run: |
+          ansible-lint -x 204 ./build/ansible_roles/*


mildas:
**Changes identified:**
Others:
 Python abstract syntax tree change found in utils/ansible_playbook_to_role.py. - ctest will be selected.

 (cd build && ctest -j4)

mildas:
**Changes identified:**
Others:
 Python abstract syntax tree change found in utils/ansible_playbook_to_role.py. - ctest will be selected.

 (cd build && ctest -j4)

jan-cerny:
@isimluk I think that instead of introducing a new check mechanism this should be added to upstream test suite.

.github/workflows/validate_ansible_roles.yaml
```diff
@@ -0,0 +1,24 @@
...
+  validate:
+    name: Validate Ansible Roles
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Deps
+        uses: mstksg/get-package@master
+        with:
+                apt-get: cmake ninja-build libopenscap8 libxml2-utils expat xsltproc python3-jinja2 python3-yaml ansible-lint
+      - name: Checkout
+        uses: actions/checkout@v1
```
redhatrises, line 13: @isimluk why not use https://github.com/marketplace/actions/ansible-lint ?

redhatrises: @isimluk on second thought.... could we just do https://github.com/marketplace/actions/ansible-molecule ? ansible-lint exists already in molecule. It may not be feasible or work to have Molecule, but would be really nice to have.

konstruktoid: my 2c @redhatrises, but `ansible-molecule` does a acceptable job on specific use cases and might work on this project but will fail as soon as some actions, like operational testing (restarting services etc), has to be performed.

I personally use the following and it will result in a 5/5 on Ansible Galaxy (https://galaxy.ansible.com/konstruktoid/hardening).

```
  if ! find ./ -type f -name '*.y*ml' ! -name '.*' -print0 | \
    xargs -0 ansible-lint; then
      echo 'ansible-lint failed.'
      exit 1
  fi

  if ! find ./ -type f -name '*.y*ml' ! -name '.*' -print0 | \
    xargs -0 yamllint -d "{extends: default, rules: {line-length: {level: warning}}}"; then
      echo 'yamllint failed.'
      exit 1
  fi
```

jan-cerny:
@isimluk Do you have any plans?

isimluk:
Hello @jan-cerny, Didn't get chance to follow-up on this, but definitely plan to get this towards successful merge. :sauropod:

Issue #5730, opened 2020-05-04
python2 test_suite.py - UnboundLocalError exception


mildas:
#### Description of problem:
`test_suite.py` throws exception when run with python2:
```
Traceback (most recent call last):
  File "/tmp/tmp.LvWxy5TK8S/rpmbuild/BUILD/scap-security-guide-0.1.50/tests/test_suite.py", line 368, in <module>
    main()
  File "/tmp/tmp.LvWxy5TK8S/rpmbuild/BUILD/scap-security-guide-0.1.50/tests/test_suite.py", line 358, in main
    with xml_operations.datastream_root(stashed_datastream, stashed_datastream) as root:
  File "/usr/lib64/python2.7/contextlib.py", line 17, in __enter__
    return self.gen.next()
  File "/tmp/tmp.LvWxy5TK8S/rpmbuild/BUILD/scap-security-guide-0.1.50/tests/ssg_test_suite/xml_operations.py", line 78, in datastream_root
    tree.write(save_location)
UnboundLocalError: local variable 'tree' referenced before assignment
```

#### SCAP Security Guide Version:
14fde08

#### Operating System Version:
RHEL 7

#### Steps to Reproduce:
1. python2 tests/test_suite.py combined --libvirt qemu:///system test-suite-rhel7 --datastream build/ssg-rhel7-ds.xml --remediate-using bash xccdf_org.ssgproject.content_profile_ospp

#### Actual Results:
```
Setting console output to log level INFO
INFO - The DataStream contains 2 Benchmarks
INFO - 0 - scap_org.open-scap_cref_ssg-rhel7-xccdf-1.2.xml
INFO - 1 - scap_org.open-scap_cref_ssg-rhel7-pcidss-xccdf-1.2.xml
INFO - Selected Benchmark is 0
INFO - To select a different Benchmark, use --xccdf-id-number option.
INFO - The base image option has not been specified, choosing libvirt-based test environment.
INFO - Logging into /home/mlysonek/SCAP/content/logs/combined-custom-2020-05-04-1128/test_suite.log
Traceback (most recent call last):
  File "tests/test_suite.py", line 368, in <module>
    main()
  File "tests/test_suite.py", line 358, in main
    with xml_operations.datastream_root(stashed_datastream, stashed_datastream) as root:
  File "/usr/lib64/python2.7/contextlib.py", line 17, in __enter__
    return self.gen.next()
[...]

PR #5747, opened 2020-05-07
Fix incomplete temporary file

jan-cerny:
Ensure all the data has been written before reusing the temporary
file. Occurred with Python 2 only.

Fixes: #5730


mildas:
**Changes identified:**
Others:
 Python abstract syntax tree change found in tests/test_suite.py. - ctest will be selected.

 (cd build && ctest -j4)

matejak:
LGTM, I would do the same :-)

PR #5748, opened 2020-05-07
Flush the write buffers after write.

matejak:
Flushing doesn't seem to be needed in Python3, but the code doesn't always work for Python2.
Closing the temp file makes it deleted, so that's why flushing is used instead.

Fixes: #5730


mildas:
**Changes identified:**
Others:
 Python abstract syntax tree change found in tests/test_suite.py. - ctest will be selected.

 (cd build && ctest -j4)

jan-cerny:
This is a duplicate of #5747

Issue #5712, opened 2020-04-27
RHEL7 - rpm_verify_ownership fails after kickstart installation


mildas:
#### Description of problem:
Rule `rpm_verify_ownership` fails after rhel7 kickstart installation with "Server with GUI" software selection and stig, e8, or cis profile.

#### SCAP Security Guide Version:
9e403dd

#### Operating System Version:
RHEL 7.8

#### Steps to Reproduce:
1. Install RHEL 7.8 with `Server with GUI` software selection and `stig`, `e8`, or `cis` security policy.
2. After installation scan the machine with the profile selected during the installation.

#### Actual Results:
`rpm_verify_ownership` fails.

#### Expected Results:
`rpm_verify_ownership` passes.

#### Addition Information/Debugging Steps:
All three profiles show that the `rpm_verify_ownership` failed because of `gdm` package.

redhatrises:
Looks like a bug against GDM.

matusmarhefka:
This is also issue in RHEL8 for `rpm_verify_ownership` and `rpm_verify_permissions` rules which fail on `gdm` package, profiles set different group ownership and permission for `/var/log/gdm` file compared to the one from package (happens for `hipaa`, `e8` and `pci-dss` with `Server with GUI`).

redhatrises:
This seems to be a continuing issue with GDM. Is there a bug for this and can we press the issue for them to fix it?

ggbecker:
We have https://bugzilla.redhat.com/show_bug.cgi?id=1878119 which is under `scap-security-guide` at the moment of this comment but we intend to move to GDM component.

ggbecker:
https://bugzilla.redhat.com/show_bug.cgi?id=1878119 has been verified and gdm build gdm-40.0-19.el8 should include the fix. @mildas can you check if this is still not fixed?

mildas:
gdm bug got fixed, but now there is problem in fprintd package - https://bugzilla.redhat.com/show_bug.cgi?id=2069297

Closing this issue, as it is not issue in content but in other packages.

PR #5755, opened 2020-05-12
Warn about findings from rpm_verify_permissions and rpm_verify_ownership

yuumasato:
#### Description:
There can be cases in which a Profile requires that a file permission be
more strict than package default permissions, or a file be owned by root.
In these cases this rule will report the file changed by the Profile itself as a finding.

#### Rationale:
Not all permission or ownership requirements make sense to be incorporated by the
package, and currently there is no mechanism to waive these findings.

### DIFF
--- a/linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_permissions/rule.yml
+++ b/linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_permissions/rule.yml
@@ -69,7 +69,10 @@ ocil: |-
 
 warnings:
     - general: |-
-        Profiles may require that specific files have stricter file permissions than defined by the vendor. Such files will be reported as a finding and need to be evaluated according to your policy and deployment environment.
+        Profiles may require that specific files have stricter file permissions than defined by the
+        vendor.
+        Such files will be reported as a finding and need to be evaluated according to your policy
+        and deployment environment.
 {{% if product == "rhel6" %}}
     - general: |-
         <b>Note: Due to a bug in the <tt>gdm</tt> package,
--- a/linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_permissions/rule.yml
+++ b/linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_permissions/rule.yml
@@ -67,8 +67,10 @@ ocil: |-
     is expected by the RPM database:
     <pre>$ rpm -Va | awk '{ if (substr($0,2,1)=="M") print $NF }'</pre>
 
-{{% if product == "rhel6" %}}
 warnings:
+    - general: |-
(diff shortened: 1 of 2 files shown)


linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_permissions/rule.yml
```diff
@@ -67,8 +67,10 @@ ocil: |-
     is expected by the RPM database:
     <pre>$ rpm -Va | awk '{ if (substr($0,2,1)=="M") print $NF }'</pre>
 
-{{% if product == "rhel6" %}}
 warnings:
+    - general: |-
+        Profiles may require that specific files have stricter file permissions than defined by the vendor. Such files will be reported as a finding and need to be evaluated according to your policy and deployment environment.
+{{% if product == "rhel6" %}}
```
redhatrises, line 73: This doesn't appear to be RHEL6 specific, and there appears to be a problem for RHEL7, possibly RHEL8. Should we remove this?

matejak: The referenced BZ is filed against RHEL6. Are you sure about that?

redhatrises:
There's #5712 opened against RHEL7. We should update the message while we are here if there are multiple problems against gdm in RHEL.

matejak:
I would love to have something similar for `rpm_verify_ownership`, as ownership by `root` can be our stricter requirement than ownership defined by the RPM package.

linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_permissions/rule.yml
yuumasato: @redhatrises You mean to make the warning applicable to all products, right?

redhatrises: @yuumasato yeah

yuumasato: Hmm, the referenced BZ is a clone of RHEL7 bug, and the RHEL7 bug was solved: https://bugzilla.redhat.com/show_bug.cgi?id=1275532

I would still keep the warning specific for RHEL6.

yuumasato:
I have added similar warning to `rpm_verify_ownership` rule.

matejak:
I am merging the PR. If more warnings with greater specificity are desired, they can be discussed in subsequent PRs, but this one already has a significant positive impact.

linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_permissions/rule.yml
redhatrises: Ahhh... okay. Got it thanks for that.

PR #5756, opened 2020-05-13
fix grub2_bootloader_argument template

### LINKED ISSUE #5754: Kernel audit parameter change should be idempotent
#### Description of problem:
The ansible fragment in template_ANSIBLE_grub2_bootloader_argument is not idempotent.  
Running it multiple times adds additional "audit=1" , and eventually you will fill up the allocated environment space.  The rule in my case is xccdf_org.ssgproject.content_rule_grub2_audit_argument .

Upstream

Any that uses this template.  

#### Steps to Reproduce:

1. Run these tasks multiple times....

```
- name: get current kernel parameters
  command: /usr/bin/grub2-editenv - list
  register: kernelopts
  changed_when: false
  when: ansible_virtualization_role != "guest" or ansible_virtualization_type != "docker"
  tags:
    - grub2_audit_argument
    - medium_severity
    - restrict_strategy
    - medium_complexity
    - low_disruption
    - reboot_required
    - NIST-800-53-AC-17(1)
    - NIST-800-53-AU-14(1)
    - NIST-800-53-AU-10
    - NIST-800-53-CM-6(a)
    - NIST-800-53-IR-5(1)
    - NIST-800-171-3.3.1
    - PCI-DSS-Req-10.3
    - CJIS-5.4.1.1

- name: Update the bootloader menu
  command: /usr/bin/grub2-editenv - set "{{ item }} audit=1"
  with_items: '{{ kernelopts.stdout_lines | select(''match'', ''^kernelopts.*'') |
    list }}'
  when:
    - kernelopts.stdout_lines is defined
    - kernelopts.stdout_lines | length > 0
    - ansible_virtualization_role != "guest" or ansible_virtualization_type != "docker"
  tags:
    - grub2_audit_argument
    - medium_severity
    - restrict_strategy
    - medium_complexity
    - low_disruption
    - reboot_required
    - NIST-800-53-AC-17(1)
    - NIST-800-53-AU-14(1)
    - NIST-800-53-AU-10
    - NIST-800-53-CM-6(a)
    - NIST-800-53-IR-5(1)
    - NIST-800-171-3.3.1
    - PCI-DSS-Req-10.3
    - CJIS-5.4.1.1
```

vojtapolasek:
#### Description:

- fix regex in ansible remediation for rhel7

- add limiting condition to rhel8 version of ansible and bash remediation

- add tests for the rule grub2_audit_backlog_limit_argument to catch the case where argument and its value were there, but directly after the value there were extra characters, effectively changing the value. See rationale.

- modify regular expression in oval state to be strict. Argument together with value have to be surrounded by white space characters, or eventually begining / end of line.

#### Rationale:

This fixes several problems:

- ansible and bash remediation of grub2 arguments through grub2-editenv command did not check if the correct value is already there, creating duplicate entries. Multiple remediations could exhaust grubenv name space.

- ansible remediation of /etc/default/grub performed wrong replacement of characters. For example, in case we were remediating audit_backlog_limit=8192 and the value was already present there, the result would be audit_backlog_limit=8192192.

- the previous problem let to the fix of error in oval check. The regex used in the check was correctly checking for the arg=value, but it did not care about surroundings. So audit=1 was correct, myaudit=1 was correct, audit=10 was correct... but we want only audit=1 to be correct.

- Fixes #5754

### DIFF
--- a/shared/templates/template_BASH_grub2_bootloader_argument
+++ b/shared/templates/template_BASH_grub2_bootloader_argument
@@ -14,7 +14,7 @@ fi
 grubby --update-kernel=ALL --args="{{{ ARG_NAME_VALUE }}}"
 {{% else %}}
 # Correct grub2 kernelopts value using grub2-editenv
-if ! grub2-editenv - list | grep -qE '^kernelopts(=|=.*\s){{{ ARG_NAME_VALUE }}}($|\s).*$'; then
+if ! grub2-editenv - list | grep -qE '^kernelopts=(.*\s)?{{{ ARG_NAME_VALUE }}}(\s.*)?$'; then
   grub2-editenv - set "$(grub2-editenv - list | grep kernelopts) {{{ ARG_NAME_VALUE }}}"
 fi
 {{% endif %}}
--- a/shared/templates/template_BASH_grub2_bootloader_argument
+++ b/shared/templates/template_BASH_grub2_bootloader_argument
@@ -14,5 +14,7 @@ fi
 grubby --update-kernel=ALL --args="{{{ ARG_NAME_VALUE }}}"
 {{% else %}}
 # Correct grub2 kernelopts value using grub2-editenv
-grub2-editenv - set "$(grub2-editenv - list | grep kernelopts) {{{ ARG_NAME_VALUE }}}"
+if ! grub2-editenv - list | grep -qE '^kernelopts(=|=.*\s){{{ ARG_NAME_VALUE }}}($|\s).*$'; then
+  grub2-editenv - set "$(grub2-editenv - list | grep kernelopts) {{{ ARG_NAME_VALUE }}}"
+fi
 {{% endif %}}
--- a/shared/templates/template_OVAL_grub2_bootloader_argument
+++ b/shared/templates/template_OVAL_grub2_bootloader_argument
@@ -92,7 +92,7 @@
 
   <ind:textfilecontent54_state id="state_grub2_{{{ SANITIZED_ARG_NAME }}}_argument"
   version="1">
-    <ind:subexpression datatype="string" operation="pattern match">^.*{{{ ESCAPED_ARG_NAME_VALUE }}}.*$</ind:subexpression>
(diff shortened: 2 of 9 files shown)


shared/templates/template_OVAL_grub2_bootloader_argument
```diff
@@ -92,7 +92,7 @@
 
   <ind:textfilecontent54_state id="state_grub2_{{{ SANITIZED_ARG_NAME }}}_argument"
   version="1">
-    <ind:subexpression datatype="string" operation="pattern match">^.*{{{ ESCAPED_ARG_NAME_VALUE }}}.*$</ind:subexpression>
+    <ind:subexpression datatype="string" operation="pattern match">^.*(?:^|\s){{{ ESCAPED_ARG_NAME_VALUE }}}(?:$|\s).*$</ind:subexpression>
```
yuumasato, line 95: `>^.*(?:^|\s({{{ ESCAPED_` it is a bit strange to expect beginning of line in the group.

Suggestion for the regex:
`^(?:.*\s)?{{{ ESCAPED_`

yuumasato: Suggestion for end part of the regex:
`(?:\s.*)?$`

shared/templates/template_BASH_grub2_bootloader_argument
```diff
@@ -14,5 +14,7 @@ fi
 grubby --update-kernel=ALL --args="{{{ ARG_NAME_VALUE }}}"
 {{% else %}}
 # Correct grub2 kernelopts value using grub2-editenv
-grub2-editenv - set "$(grub2-editenv - list | grep kernelopts) {{{ ARG_NAME_VALUE }}}"
+if ! grub2-editenv - list | grep -qE '^kernelopts(=|=.*\s){{{ ARG_NAME_VALUE }}}($|\s).*$'; then
```
yuumasato, line 17: Unless I miss something, I guess the regex can be simplified a bit:
`'^kernelopts=(.*\s)?{{{ ARG_NAME_VALUE }}}(\s.*)?$'`

Same thing for the Ansble template.

vojtapolasek:
You are right, regexes got a bit overengineered. Thanks for pointing that out.

yuumasato:
Nice work on the fix! :)

PR #5758, opened 2020-05-15
Template for yamlfilecontent checks

evgenyz:
This PR will add the `yamlfile_value` OVAL template. And then the rule from #5326 was changed to utilize this template.


matejak:
Pretty good first attempt! I would suggest this:

- It is questionable whether it makes sense to have a macro involved. Basically you can have a template that calls a macro, or the body of that macro can be the template itself. I am fine with either, as we already have macros that are rather specialized than reusable, so there is no precedent.
- Document your template by expanding the corresponding adoc: https://github.com/ComplianceAsCode/content/blob/master/docs/manual/developer_guide.adoc#732-list-of-available-templates

shared/macros-oval.jinja
```diff
@@ -515,3 +515,72 @@
...
+      <ind:path var_ref="{{{ rule_id }}}_file_location"/>
+      <ind:filename>{{{ file }}}</ind:filename>
+      <ind:yamlpath>{{{ yamlpath }}}</ind:yamlpath>
+    </ind:yamlfilecontent_object>
+
+    <ind:yamlfilecontent_state id="state_{{{ rule_id }}}" version="1">
+	    <ind:value_of {{% if type %}}datatype="{{{ type }}}"{{% endif %}} {{% if check %}}entity_check="{{{ check }}}"{{% endif %}}>{{{ value }}}</ind:value_of>
+    </ind:yamlfilecontent_state>
+
+    <external_variable comment="Root of OCP data dump" datatype="string" id="ocp_data_root" version="1" />
```
yuumasato, line 582: This probably should be wrapped in a conditional for `ocp_data`

shared/macros-oval.jinja
```diff
@@ -515,3 +515,72 @@
...
+  <def-group>
+    <definition class="compliance" id="{{{ rule_id }}}" version="1">
+      <metadata>
+        <title>Check the value of '{{{ yamlpath }}}' in the "{{{ file }}}" file</title>
+        {{{- oval_affected(products) }}}
+        <description>The file {{{ file }}} in {{{ path }}} should {{% if negate %}}not{{% endif %}} contain value "{{{ value }}}" of type {{{ type }}} at {{{ yamlpath }}}.</description>
+      </metadata>
+      <criteria>
+          <criterion comment="Make sure that {{% if negate %}}no{{% endif %}} {{% if check %}}({{{ check }}}){{% endif %}} element of '{{{ yamlpath }}}' has value '{{{ value }}}'' of type {{{ type }}}." {{% if negate %}}negate="true"{{% endif %}} test_ref="test_{{{ rule_id }}}"/>
+          <criterion comment="Make sure that there is the file '{{{ file }}}' exists in {{{ path }}}." test_ref="test_file_for_{{{ rule_id }}}"/>
```
yuumasato, line 543: What is the reason for this extra test for `{{{ file }}}` existence?
Won't the `yamlfilecontent_test` already fail if it doesn't exist?

shared/macros-oval.jinja
```diff
@@ -515,3 +515,72 @@
...
+       <concat>
+	       <variable_component var_ref="ocp_data_root"/>
+	       <literal_component>{{{ path }}}</literal_component>
+       </concat>
+       {{% else %}}
+       <literal_component>{{{ path }}}</literal_component>
+       {{% endif %}}
+    </local_variable>
+
+    <ind:yamlfilecontent_test id="test_{{{ rule_id }}}" check="at least one" comment="Find at least one match of '{{{ value }}}' in '{{{ yamlpath }}}'." version="1">
```
yuumasato, line 558: > `check="at least one"`

If multiple objects are found is it enough that at least one conforms to the state? Can the others be non-compliant?

evgenyz:
OCP configuration files also contain some metadata about its origins. For example the configuration from `apis/config.openshift.io/v1/oauths/cluster` (from #5326) would have:
```yaml
apiVersion: config.openshift.io/v1
kind: OAuth
name: cluster
```
Ideally we would like to check if the file we are checking is 'correct', i.e. originated from expected API endpoint.

evgenyz:
Some notes from @jhrozek:

> OK, so the singulars and plurals would best be autodiscovered, it is possible by HTTP GET for the group "above", so in our example you could GET /apis/config.openshift.io/v1/ and then it would show you all the object underneath with their singulars and plurals.

> I'm thinking there should be more metadata included like we include the endpoint
one other thing to note, the paths can be different depending on whether the object exists in a namespace or not, for example the oauths are cluster-scoped so they do not belong to a namespace. But some other objects, like pods live in a namespace and then the namespace is part of the path. But I'm thinking then the namespace should be also included in the `ocp-dump-location`.

shared/macros-oval.jinja
matejak: The test won't find any objects. Sometimes you want the whole rule to fail is objects are found, so the inability to fetch the actual yaml file would cause a false negative.

yuumasato: Right, so the check for the file is to be sure that when you don't objects to be found, it is not because the file is missing.

evgenyz: Maybe it would make sense to add it only for `negate=true` case.

shared/macros-oval.jinja
evgenyz: I'm not sure if there is a way to fetch multiple objects for this probe. I'll check the code.

shared/macros-oval.jinja
evgenyz: Right

mildas:
**Changes identified:**
Others:

 (cd build && ctest -j4)

shared/macros-oval.jinja
evgenyz: So I think that for this template it would make sense to have 'only one' there.

mildas:
**Changes identified:**
Others:

 (cd build && ctest -j4)

shared/macros-oval.jinja
yuumasato: Looks good.

shared/templates/template_OVAL_yamlfile_value
```diff
@@ -0,0 +1,58 @@
...
+      <concat>
+        <variable_component var_ref="ocp_data_root"/>
+        <literal_component>{{{ PATH }}}</literal_component>
+      </concat>
+      {{% else %}}
+      <literal_component>{{{ PATH }}}</literal_component>
+      {{% endif %}}
+  </local_variable>
+
+  <ind:yamlfilecontent_test id="test_{{{ rule_id }}}" check="only one" comment="Find only one '{{{ FILE }}}' with '{{{ VALUE }}}' at '{{{ YAMLPATH }}}'." version="1">
```
yuumasato, line 28: @evgenyz You are correct, only one object will be returned.
Reference, https://github.com/OpenSCAP/openscap/pull/1480#discussion_r382595294
In this case, `check="only one"` can be omited.

If still you'd like to make it clear that only one object is expected, the attribute that should be there is `check_existence="only_one_exists"`.
https://github.com/OVALProject/Language/blob/master/docs/oval-common-schema.md#---existenceenumeration---

shared/templates/template_OVAL_yamlfile_value
```diff
@@ -0,0 +1,58 @@
...
+    <ind:object object_ref="object_{{{ rule_id }}}"/>
+    <ind:state state_ref="state_{{{ rule_id }}}"/>
+  </ind:yamlfilecontent_test>
+
+  {{% if NEGATE %}}
+  <unix:file_test id="test_file_for_{{{ rule_id }}}" check="only one" comment="Find the file to be checked ('{{{ FILE }}}' in '{{{ PATH }}}')." version="1">
```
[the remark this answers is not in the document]
evgenyz: Hmm, if I omit `check` I have this: `File '/home/ekolesni/Devel/content/build/ssg-ocp4-ds.xml' line 26839: Element '{http://oval.mitre.org/XMLSchema/oval-definitions-5#unix}file_test': The attribute 'check' is required but missing.
 [/builddir/build/BUILD/openscap-1.3.3/src/XCCDF/xccdf_session.c:613]`

shared/templates/template_OVAL_yamlfile_value
evgenyz: `OpenSCAP Error: File '/home/ekolesni/Devel/content/build/ssg-ocp4-ds.xml' line 26835: Element '{http://oval.mitre.org/XMLSchema/oval-definitions-5#independent}yamlfilecontent_test': The attribute 'check' is required but missing.
 [/builddir/build/BUILD/openscap-1.3.3/src/XCCDF/xccdf_session.c:613]`

mildas:
**Changes identified:**
Rule ocp_idp_no_htpasswd:
 OVAL check for ocp_idp_no_htpasswd was deleted.
Others:

docs/manual/developer_guide.adoc
```diff
@@ -1673,6 +1673,19 @@ timer_enabled::
...
 
+yamlfile_value::
+* Check if value(s) of certain type is (are) present in a YAML (or JSON) file at a given path.
+* Parameters:
+** *path* - path of the file to check.
+** *file* - name of the file to check.
+** *yamlpath* - OVAL's link:https://github.com/OpenSCAP/yaml-filter/wiki/YAML-Path-Definition[YAML Path] expression.
+** *value* - the value to check.
+** *type* (string/SimpleDatatypeEnumeration) - datatype for state's value_of, optional.
+** *check* (string/CheckEnumeration) - entity_check value for state's value_of, optional.
```
ggbecker, line 1684:
```suggestion
** *entity_check* (string/CheckEnumeration) - entity_check value for state's value_of, optional.
```

evgenyz: Ah, nice catch, thanks

applications/openshift/ocp_idp_no_htpasswd/rule.yml
```diff
@@ -14,3 +14,13 @@ severity: medium
 warnings:
     - general: |-
         {{{ openshift_cluster_setting("/apis/config.openshift.io/v1/oauths/cluster") | indent(8) }}}
+
+template:
+    name: yamlfile_value
+    vars:
+        file: cluster.yaml
+        yamlpath: ".spec.identityProviders[:].type"
```
ggbecker, line 22: Just a guess. `rhel6` and `rhel7` CI nodes use `python2` to build the content. These `[:]` might be the problem when it's trying to load the `yaml` file.

This is the exact file that is causing the exception:
`Exception while handling file: /home/jenkins/workspace/scap-security-guide-pull-requests/label/rhel7/build/ocp4/rules/ocp_idp_no_htpasswd.yml`

ggbecker: maybe this is related: https://github.com/google/rekall/pull/323/files

ggbecker: I found the issue. I'll point it out in the right place.

applications/openshift/ocp_idp_no_htpasswd/rule.yml
```diff
@@ -14,3 +14,13 @@ severity: medium
...
+        value: "HTPasswd"
+        entity_check: "at least one"
+        negate: True
```
ggbecker, line 25: the build system doesn't support using directly `True` here when building with `python2`. We need to use something like:
```suggestion
        negate: 'yes'
```

applications/openshift/ocp_idp_no_htpasswd/rule.yml
```diff
@@ -14,3 +14,13 @@ severity: medium
...
+        ocp_data: True
```
ggbecker, line 26: Same here:
```suggestion
        ocp_data: 'yes'
```

applications/openshift/ocp_idp_no_htpasswd/rule.yml
ggbecker: I remember having some sort of documentation talking about this, but I can't find it. Maybe @matejak

jan-cerny: @ggbecker  I also can't find it.
I think it's caused by this: https://github.com/ComplianceAsCode/content/blob/141f810e048bb0dfbc15ae162c76ba6ee5f39078/ssg/yaml.py#L27
This can be relevant: https://github.com/ComplianceAsCode/content/pull/2737

matejak: Yes, back at the day when we moved to YAML, there was pushback against the implicit conversion of bool strings to actual bools, so it is turned off. The YAML parsing lib IMO goes too far - it converts both keys and values. But at the same time, turning this off for our content may violate expectations as well as cause problems like you have experienced. The template documentation should be clear that one should use quoted strings as arguments.

mildas:
**Changes identified:**
Rule ocp_idp_no_htpasswd:
 OVAL check for ocp_idp_no_htpasswd was deleted.
Others:

applications/openshift/ocp_idp_no_htpasswd/rule.yml
evgenyz: Done and done.

applications/openshift/ocp_idp_no_htpasswd/rule.yml
evgenyz: Thanks!

applications/openshift/ocp_idp_no_htpasswd/rule.yml
```diff
@@ -14,3 +14,13 @@ severity: medium
 warnings:
     - general: |-
         {{{ openshift_cluster_setting("/apis/config.openshift.io/v1/oauths/cluster") | indent(8) }}}
+
+template:
+    name: yamlfile_value
+    vars:
+        file: cluster.yaml
```
mrogers950, line 21: Right now compliance-operator expects the file path to match what's in openshift_cluster_setting(), which helps avoid conflicts. Otherwise we would have to verify that no other rule specifies cluster.yaml.

evgenyz: Aha. I'll change the template to use `filepath` in case when it is rendered for OCP checks.

mildas:
**Changes identified:**
Rule ocp_idp_no_htpasswd:
 OVAL check for ocp_idp_no_htpasswd was deleted.
Others:

evgenyz:
> After running some tests I have discovered that the way this template is implemented it doesn't work with values such as: `negate: 'no'`, because it will be evaluated as `True` in jinja macros for expressions such as: `{{% if NEGATE %}}`. In this case, these values need to be parsed by the function: `def yamlfile_value(data, lang):` first, similar to what's being done in: https://github.com/ComplianceAsCode/content/blob/master/ssg/templates.py#L310
> 
> Unfortunately the templating system cannot handle such kind of conversions automatically, a more robust system would have to implemented.

But this argument is optional, I was thinking that when you don't want it to be true, you could just omit it completely.

ggbecker:
> > After running some tests I have discovered that the way this template is implemented it doesn't work with values such as: `negate: 'no'`, because it will be evaluated as `True` in jinja macros for expressions such as: `{{% if NEGATE %}}`. In this case, these values need to be parsed by the function: `def yamlfile_value(data, lang):` first, similar to what's being done in: https://github.com/ComplianceAsCode/content/blob/master/ssg/templates.py#L310
> > Unfortunately the templating system cannot handle such kind of conversions automatically, a more robust system would have to implemented.
> 
> But this argument is optional, I was thinking that when you don't want it to be true, you could just omit it completely.

I understand your point. But I would prefer to make it explicit, people may come and don't know exactly what's the default and just want to make sure that the option is disabled.

Also, the documentation says default is `no` for some of the options, so I would just assume that this option is valid and I could use it if want. Which is not the case.

@matejak What do you think about this?

mildas:
**Changes identified:**
Rule ocp_idp_no_htpasswd:
 OVAL check for ocp_idp_no_htpasswd was deleted.
Others:

applications/openshift/ocp_idp_no_htpasswd/rule.yml
```diff
@@ -14,3 +14,13 @@ severity: medium
...
+
+template:
+    name: yamlfile_value
+    vars:
+        filepath: /apis/config.openshift.io/v1/oauths/cluster
+        yamlpath: ".spec.identityProviders[:].type"
+        value: "HTPasswd"
+        entity_check: "at least one"
+        negate: 'yes'
+        ocp_data: 'yes'
```
ggbecker, line 26:
```suggestion
        negate: 'true'
        ocp_data: 'true'
```

To align with current expected format.

evgenyz: Geez, thanks!

mildas:
**Changes identified:**
Rule ocp_idp_no_htpasswd:
 OVAL check for ocp_idp_no_htpasswd was deleted.
Others:

shared/templates/template_OVAL_yamlfile_value
```diff
@@ -0,0 +1,58 @@
+{{% if target_oval_version >= [5, 11] %}}
+<def-group>
+  <definition class="compliance" id="{{{ rule_id }}}" version="1">
+    <metadata>
+      <title>Check the value(s) of '{{{ YAMLPATH }}}' in the file "{{{ FILE }}}" (located in "{{{ PATH }}}")</title>
+      {{{- oval_affected(products) }}}
+      <description>
+        The file "{{{ FILEPATH }}} should {{% if NEGATE %}}not {{% endif %}}contain value "{{{ VALUE }}}" of type {{{ TYPE }}} at '{{{ YAMLPATH }}}'.
```
ggbecker, line 8:
```suggestion
        The file "{{{ FILEPATH }}}" should {{% if NEGATE %}}not {{% endif %}}contain value "{{{ VALUE }}}" {{% if TYPE %}}of type {{{ TYPE }}} {{% endif %}} at '{{{ YAMLPATH }}}'.
```

shared/templates/template_OVAL_yamlfile_value
```diff
@@ -0,0 +1,58 @@
...
+      </description>
+    </metadata>
+    <criteria>
+        <criterion comment="Make sure that {{% if NEGATE %}}no {{% endif %}}{{% if ENTITY_CHECK %}}({{{ ENTITY_CHECK }}}) {{% endif %}} element(s) of '{{{ YAMLPATH }}}' contain value '{{{ VALUE }}}'' of type '{{{ TYPE }}}''."{{% if NEGATE %}} negate="true"{{% endif %}} test_ref="test_{{{ rule_id }}}"/>
```
ggbecker, line 12:
```suggestion
        <criterion comment="Make sure that {{% if NEGATE %}}no {{% endif %}}{{% if ENTITY_CHECK %}}({{{ ENTITY_CHECK }}}) {{% endif %}} element(s) of '{{{ YAMLPATH }}}' contain value '{{{ VALUE }}}' {{% if TYPE %}}of type {{{ TYPE }}} {{% endif %}}."{{% if NEGATE %}} negate="true"{{% endif %}} test_ref="test_{{{ rule_id }}}"/>
```

mildas:
**Changes identified:**
Rule ocp_idp_no_htpasswd:
 OVAL check for ocp_idp_no_htpasswd was deleted.
Others:

shared/templates/template_OVAL_yamlfile_value
```diff
@@ -0,0 +1,58 @@
+{{% if target_oval_version >= [5, 11] %}}
+<def-group>
+  <definition class="compliance" id="{{{ rule_id }}}" version="1">
+    <metadata>
+      <title>Check the value(s) of '{{{ YAMLPATH }}}' in the file "{{{ FILE }}}" (located in "{{{ PATH }}}")</title>
```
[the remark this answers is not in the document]
evgenyz: Done

shared/templates/template_OVAL_yamlfile_value
evgenyz: Done

shared/templates/template_OVAL_yamlfile_value
evgenyz: Done

shared/templates/template_OVAL_yamlfile_value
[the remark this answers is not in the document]
ggbecker: I took a look at this and all other `filetest` checks use `check="all"`, so even though it's the default, OpenSCAP doesn't seem to like the absence of this attribute. So I would recommend just to put `check="all"` to the `filetest` checks.

shared/templates/template_OVAL_yamlfile_value
```diff
@@ -0,0 +1,58 @@
...
+      <title>Check the value(s) of '{{{ YAMLPATH }}}' in the file "{{{ FILEPATH }}}"</title>
+      {{{- oval_affected(products) }}}
+      <description>
+        The file "{{{ FILEPATH }}}" should {{% if NEGATE %}}not {{% endif %}}contain value "{{{ VALUE }}}"{{% if TYPE %}} of type {{{ TYPE }}}{{% endif %}} at '{{{ YAMLPATH }}}'.
+      </description>
+    </metadata>
+    <criteria>
+        <criterion comment="Make sure that {{% if NEGATE %}}no {{% endif %}}{{% if ENTITY_CHECK %}}({{{ ENTITY_CHECK }}}) {{% endif %}} element(s) of '{{{ YAMLPATH }}}' contain value '{{{ VALUE }}}'{{% if TYPE %}} of type {{{ TYPE }}}{{% endif %}}."{{% if NEGATE %}} negate="true"{{% endif %}} test_ref="test_{{{ rule_id }}}"/>
+        {{% if NEGATE %}}
+        <criterion comment="Make sure that there is the file '{{{ FILEPATH }}}' exists." test_ref="test_file_for_{{{ rule_id }}}"/>
```
ggbecker, line 14:
```suggestion
        <criterion comment="Make sure that the file '{{{ FILEPATH }}}' exists." test_ref="test_file_for_{{{ rule_id }}}"/>
```

shared/templates/template_OVAL_yamlfile_value
```diff
@@ -0,0 +1,58 @@
...
+      {{% endif %}}
+  </local_variable>
+
+  <ind:yamlfilecontent_test id="test_{{{ rule_id }}}" check="only one" check_existence="only_one_exists" comment="Find only one '{{{ FILEPATH }}}' with '{{{ VALUE }}}' at '{{{ YAMLPATH }}}'." version="1">
+    <ind:object object_ref="object_{{{ rule_id }}}"/>
+    <ind:state state_ref="state_{{{ rule_id }}}"/>
+  </ind:yamlfilecontent_test>
+
+  {{% if NEGATE %}}
+  <unix:file_test id="test_file_for_{{{ rule_id }}}" check="only one" check_existence="only_one_exists" comment="Find the file to be checked ('{{{ FILEPATH }}}')." version="1">
```
ggbecker, line 36:
```suggestion
  <unix:file_test id="test_file_for_{{{ rule_id }}}" check="all" check_existence="only_one_exists" comment="Find the file to be checked ('{{{ FILEPATH }}}')." version="1">
```

shared/templates/template_OVAL_yamlfile_value
```diff
@@ -0,0 +1,58 @@
...
+      <concat>
+        <variable_component var_ref="ocp_data_root"/>
+        <literal_component>{{{ FILEPATH }}}</literal_component>
+      </concat>
+      {{% else %}}
+      <literal_component>{{{ FILEPATH }}}</literal_component>
+      {{% endif %}}
+  </local_variable>
+
+  <ind:yamlfilecontent_test id="test_{{{ rule_id }}}" check="only one" check_existence="only_one_exists" comment="Find only one '{{{ FILEPATH }}}' with '{{{ VALUE }}}' at '{{{ YAMLPATH }}}'." version="1">
```
ggbecker, line 30:
```suggestion
  <ind:yamlfilecontent_test id="test_{{{ rule_id }}}" check="all" check_existence="only_one_exists" comment="Find only one '{{{ FILEPATH }}}' with '{{{ VALUE }}}' at '{{{ YAMLPATH }}}'." version="1">
```

mildas:
**Changes identified:**
Rule ocp_idp_no_htpasswd:
 OVAL check for ocp_idp_no_htpasswd was deleted.
Others:

PR #5765, opened 2020-05-15
Ansible mount_option: split mount and option task

yuumasato:
#### Description:

- Separate task that adds mount options mounts the mountpoint into two tasks.
  - Now there is one task to "add" the mount point, and another to make sure the mount point is mounted

#### Rationale:

-  Test scenario `separate.fail.sh` for `mount_option_var_tmp_noexec` started to fail after #5752 
  - Conditioning the "mount" task on the absence of the target mount option caused the task to always be skipped when mount option was alredy present, and could result in the mount point not 
being mounted.

### DIFF
--- a/shared/templates/template_ANSIBLE_mount_option
+++ b/shared/templates/template_ANSIBLE_mount_option
@@ -26,14 +26,19 @@
     - device_name.stdout is defined and device_name.stdout_lines is defined
     - (device_name.stdout | length > 0)
 
-- name: Ensure permission {{{ MOUNTOPTION }}} are set on {{{ MOUNTPOINT }}}
+- name: Make sure {{{ MOUNTOPTION }}} option is part of the to {{{ MOUNTPOINT }}} options
+  set_fact:
+    mount_info: "{{ mount_info | combine( {'options':''~mount_info.options~',{{{ MOUNTOPTION }}}' }) }}"
+  when:
+    - mount_info is defined and "{{{ MOUNTOPTION }}}" not in mount_info.options
+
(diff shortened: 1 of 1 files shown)


matusmarhefka:
LGTM, thanks for fixing it @yuumasato 

I have tested the rules which were failing because of this issue on RHEL-8.2 and I can confirm that this PR fixes their Ansible remediations:
```
$ python3 test_suite.py rule --libvirt qemu:///session rhel82_ssgts --mode online --remediate-using ansible --datastream ../build/ssg-rhel8-ds.xml --profile ospp mount_option_home_nosuid mount_option_var_tmp_noexec
...
INFO - xccdf_org.ssgproject.content_rule_mount_option_var_tmp_noexec
INFO - Script runtime.pass.sh using profile xccdf_org.ssgproject.content_profile_ospp OK
INFO - Script separate.fail.sh using profile xccdf_org.ssgproject.content_profile_ospp OK
INFO - Script fstab.fail.sh using profile xccdf_org.ssgproject.content_profile_ospp OK
INFO - xccdf_org.ssgproject.content_rule_mount_option_home_nosuid
INFO - Script runtime.pass.sh using profile xccdf_org.ssgproject.content_profile_ospp OK
INFO - Script separate.fail.sh using profile xccdf_org.ssgproject.content_profile_ospp OK
INFO - Script fstab.fail.sh using profile xccdf_org.ssgproject.content_profile_ospp OK
```

yuumasato:
> LGTM, thanks for fixing it @yuumasato

No problem, I introduced the issue in the first place, :stuck_out_tongue: . The test scenario caught it!

And thank you for the review!

PR #5768, opened 2020-05-18
Use https and .com for openssh website.

ggbecker:
#### Description:

- Update the weblink to `https` and uses `.com` extension which is the correct one nowadays.

Relates to https://jenkins.complianceascode.io/job/scap-security-guide-linkcheck/605/console even though it is probably not the cause but it'll help without a doubt to prevent future faults.


redhatrises:
The OCP e2e failure appears to be unrelated to this PR and has not affect on the OCP test.

PR #5770, opened 2020-05-19
Fix possible collision of remediations of file permissions

vojtapolasek:
#### Description:

Force ordering of "software" group before "bootloader-grub2" group.

#### Rationale:

If the "software" group is ordered after the "bootloader-grub2" group and certain profiles are remediated with Ansible playbooks, it happens that rules requesting file permissions different from the ones dictated by packagers start failing.
This problem is fixed by this PR in the following profiles:
- rhel7 cis

- rhel7 hipaa

- rhel7 ncp

- rhel7 rhelh-stig

- rhel8 cis

- rhel8 hipaa

### DIFF
--- a/ssg/build_yaml.py
+++ b/ssg/build_yaml.py
@@ -700,6 +700,11 @@ class Group(object):
         # audit_rules_privileged_commands, othervise the rule
         # does not catch newly installed screeen binary during remediation
         # and report fail
+        # the software group should come before the
+        # bootloader-grub2 group because of conflict between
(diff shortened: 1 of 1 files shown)


mildas:
**Changes identified:**
Others:

 (cd build && ctest -j4)

yuumasato:
LGTM.

I see this as a temporary fix as this doesn't solve the general problem of the package permissions.
But I think it makes sense to apply generic fixes before the specialized ones.

PR #5776, opened 2020-05-20
Move "ignition" remediations to "kubernetes" ones

JAORMX:
General "kubernetes" remediations are favored instead of having
"ignition" specific ones. The same format for the remediation can be
used.


matejak:
If I understand correctly, it is technically about moving files to a different directory, and we already have rules converted to the `kubernetes` remediation type (either directly or via a template), so it looks like that we can assume that no surprises will come.
Is everybody on board with the move, so when the PR gets rebased, it will be merged without delays? @redhatrises @evgenyz @mrogers950?

matejak:
I see no objections through the entire history of the PR, so I am merging it before conflicts arise. In the worst case, one can always undo a merge.

PR #5779, opened 2020-05-21
Attribute credit for CIS content

yuumasato:
#### Description:

- Attribute credit to Center for Internet Security for Benchmark contents
- Update the description text a bit.

This gets correctly rendered in OAA, and in HTML Guides.
trademarks

guide_trademarks

### DIFF
--- a/rhel7/profiles/cis.profile
+++ b/rhel7/profiles/cis.profile
@@ -3,9 +3,11 @@ documentation_complete: true
 title: 'CIS Red Hat Enterprise Linux 7 Benchmark'
 
 description: |-
-    This baseline aligns to the Center for Internet Security
-    Red Hat Enterprise Linux 7 Benchmark, v2.2.0, released
-    12-27-2017.
+    This profile defines a baseline that aligns to the Center for Internet Security®
+    Red Hat Enterprise Linux 7 Benchmark, v2.2.0, released 12-27-2017.
+
+    This profile includes Center for Internet Security®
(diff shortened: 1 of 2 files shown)


rhel7/profiles/cis.profile
```diff
@@ -3,9 +3,11 @@ documentation_complete: true
 title: 'CIS Red Hat Enterprise Linux 7 Benchmark'
 
 description: |-
-    This baseline aligns to the Center for Internet Security
-    Red Hat Enterprise Linux 7 Benchmark, v2.2.0, released
-    12-27-2017.
+    This profile defines a baseline that aligns to the Center for Internet Security®
+    Red Hat Enterprise Linux 7 Benchmark, v2.2.0, released 12-27-2017.
```
redhatrises, line 7: Is this date and benchmark version accurate?

yuumasato: They are correct.
But I just found out there is a 2.2.0.1, from Oct 29th 2018, :/

redhatrises:
Merging as is. Thanks.

rhel7/profiles/cis.profile
yuumasato: v2.2.0.1 seems to be unavailable for download.
And support discussion mentions there is virtually no difference between 2.2.0 and 2.2.0.1

ggbecker, line 7: small nitpick `Benchmark` here doesn't have the ™ (trademark symbol as there  is for rhel8 profile)

PR #5780, opened 2020-05-22
Fix TM typo in RHEL7 CIS

yuumasato:
#### Description:
- Brings consistency between RHEL7 and RHEL8 CIS profile descriptions
- Typo, typo, typo...


matejak:
The fix LGTM, and the test failure is unrelated, so merging.

PR #5789, opened 2020-05-27
WIP: ocp4: Enable template to generate sysctl remediations

JAORMX:
Instead of relying on manual remediations, let's just use a template.

This effectively creates a template that generates MachineConfigs that set sysctl values.

This also enables substitutions for kubernetes remediations.


mildas:
**Changes identified:**
Others:

mildas:
**Changes identified:**
Others:

mildas:
**Changes identified:**
Others:

mildas:
**Changes identified:**
Others:

jhrozek:
I wonder what do you think about the approach taken in #5760 where much of the expected file content is urlencoded already and only the variable parts are inserted through variable expansion? This way there is less logic in the `ssg/templates.py` which IMO shouldn't know much about what the templates contain..

But either way, this works, so just let me know your preference...

mildas:
**Changes identified:**
Others:

JAORMX:
> I wonder what do you think about the approach taken in #5760 where much of the expected file content is urlencoded already and only the variable parts are inserted through variable expansion? This way there is less logic in the `ssg/templates.py` which IMO shouldn't know much about what the templates contain..
> 
> But either way, this works, so just let me know your preference...

I agree. I changed it accordingly. Good idea!

mildas:
**Changes identified:**
Others:

jhrozek:
As discussed on Slack, this still doesn't work until you extract the remediations as the DS still has the `<sub>` elements in...

jhrozek:
By the way linking the generated fixes and the rules doesn't seem that hard. The rule looks something like this:
```
<fix complexity="low" disruption="medium" reboot="true" rule="sysctl_net_ipv6_conf_default_accept_source_route" strategy="disable">apiVersion: machineconfiguration.openshift.io/v1
kind: MachineConfig
spec:
  config:
    ignition:
      version: 2.2.0
    storage:
      files:
      - contents:
          source: data:,net.ipv6.conf.default.accept_source_route%20%3D%20<sub idref="sysctl_net_ipv6_conf_default_accept_source_route_value" />
        filesystem: root·
        mode: 0644·
        path: /etc/sysctl.d/75-sysctl_net_ipv6_conf_default_accept_source_route.conf
</fix>
```
Note the `rule` attribute in the `fix` element. We could match this with the `Rule` element's `id` attribute that looks like this: `<xccdf-1.2:Rule id="xccdf_org.ssgproject.content_rule_sysctl_net_ipv6_conf_default_accept_source_route" selected="false" severity="medium">`

Do you think this is worth at least a ticket or just pursing right away?

yuumasato:
LGTM.
The only thing that is not covered, is URL encoding the actual value of the variable. As the values for sysctl are simple, this shouldn't be a problem.

@JAORMX do you still plan to make changes to this PR?

JAORMX:
@yuumasato let's not merge this yet. I marked it as Work In Progress... so the problem is that even if we are able to generate correct remediations... it's unclear still how to get them from the operator. Normally we read them from the xccdf and now we'll need to fetch them from elsewhere. So, we need to sort that out first.

yuumasato:
> it's unclear still how to get them from the operator. Normally we read them from the xccdf and now we'll need to fetch them from elsewhere. So, we need to sort that out first.

So you kinda need the functionality that `oscap xccdf generate fix` would provide?

JAORMX:
> > it's unclear still how to get them from the operator. Normally we read them from the xccdf and now we'll need to fetch them from elsewhere. So, we need to sort that out first.
> 
> So you kinda need the functionality that `oscap xccdf generate fix` would provide?

Kinda... and we did look into that. Except that we still need to be able to trace what's the check and xccdf ID for the fix that's being generated. Which `oscap xccdf generate fix` doesn't provide.

PR #5791, opened 2020-05-28
ocp4: add sshd golden config as sshd_set_loglevel_info remediation

JAORMX:
The golden config was missing the loglevel setting, so it was added.


jhrozek:
I know that this is marked as WIP, but the decoded file doesn't seem to be different from the others...

JAORMX:
@jhrozek it isn't. The decoded file indeed address this specific configuration.

matejak:
test this please

PR #5792, opened 2020-05-28
ocp4: Change auditd flush frequency to incremental_async by default

JAORMX:
This addresses the requirements for e8 and moderate


jhrozek:
It's hard to actually test the remediations as quay.io is down, but in the meantime, I did a diff of the decoded files and it looks good:
```
@@ -7,7 +7,7 @@
 log_file = /var/log/audit/audit.log
 log_group = root
 log_format = ENRICHED
-flush = DATA
+flush = incremental_async
 freq = 50
 max_log_file = 8
 num_logs = 5
```

PR #5775, opened 2020-05-20
Add RHCOS4 product

JAORMX:

PR #5784, opened 2020-05-25
Add rules to configure zIPL

yuumasato:
#### Description:

- Add `bootloader-zip` group
- Add few rules for kernel options in zIPL
  - Ensure `audit=1` is set
  - Ensure `audit_backlog_limit=8192` is set
  - Ensure `selinux=0` is not set

PR #5793, opened 2020-05-28
Add 'bls_audit_option' rule

### LINKED ISSUE #5285: RHCOS does not write kernel parameters into grub.conf
#### Description of problem:
The probes that check for kernel arguments such as whether the audit=1 kernel argument has been added do not work because they seem to be checking for `/boot/grub2/grubenv`. This is not what's being used on RHCOS, the file is more or less empty there. We need to adjust the probes to check for something different.

as built from git as of today

RHCOS

#### Steps to Reproduce:

1. set kernel parameters with a MC
2. reboot the node
3. make sure the kernel parameters are set e.g. with cat /proc/cmdline
4. 

#### Actual Results:
the rule would still fail

#### Expected Results:
the rule should now pass

I guess the question is how to check the kernel arguments. The first thing that comes to my mind is `/proc/cmdline`. Is there a reason not to check that in the first place?

I did a little digging how is the kernel argument set on RHCOS and it goes like this:
 - a MachineConfig with an Ignition spec is defined by the admin
 - the MC is read by the machineConfig deamon (MCD) that is running on each cluster node
 - the MCD aggregates the resulting kernel arguments from all the applied MCs and calls `rpm-ostree kargs` and passes the list of the kernel arguments
 - `rpm-ostree` seems to write the kernel arguments to `/boot/loader/entries/ostree-$version-rhcos.conf` but I'm unsure how to tell which is the "current" version

rpm-ostree also allows you to print the current kernel arguments with `rpm-ostree kargs` but I'm unsure if it's realistic to call this from the scanner because of library loading issues and all that.

Before digging even more about how the ostree works, is using `/proc/cmdline` acceptable? We seem to be using the proc filesystem for the sysctls, so..

evgenyz:
For checking BLS-compatible boot loader entries for presence of audit=1 parameter. The rule is based on the 'bls_entries_option' template.

The `prodtype` of the rule is set to `fedora,rhel8` but it should be changed to `fcos, rhcos` (when we will have them), this rule and template won't work on regular Fedora and RHEL. See #5775, #5285.

This will fix #5285.


mildas:
**Changes identified:**
Others:

mildas:
**Changes identified:**
Others:

JAORMX:
@evgenyz could you change the directory to `kubernetes` instead of `ignition`? Also, could you add this to a profile like `moderate` or `e8`?

evgenyz:
> Let's add this to a profile

I assume that you want to add this rule to `ocp4` profiles? But here is the problem: OpenShift 4 could have nodes with RHEL and RHCOS, so, depending on the OS, the profile should contain either `grub2_audit_argument` rule or `bls_audit_option` (respectively). They are mutually exclusive, you can't have both of them as one of them will always fail.

JAORMX:
> > Let's add this to a profile
> 
> I assume that you want to add this rule to `ocp4` profiles? But here is the problem: OpenShift 4 could have nodes with RHEL and RHCOS, so, depending on the OS, the profile should contain either `grub2_audit_argument` rule or `bls_audit_option` (respectively). They are mutually exclusive, you can't have both of them as one of them will always fail.

Got it, let's wait until we have the RHCOS product then. which comes here https://github.com/ComplianceAsCode/content/pull/5775 . And then add it to a profile there.

yuumasato:
> But here is the problem: OpenShift 4 could have nodes with RHEL and RHCOS, so, depending on the OS, the profile should contain either `grub2_audit_argument` rule or `bls_audit_option` (respectively).

We plan to update RHEL rules to check BLS settings in the close future. It is not clear to me if we should write new rules or change the current ones. And zIPL is also coming into the scene, see #5784 

With that in mind, I'm thinking that the basic building blocks for BLS should be done as Jinja macros and the templates made specific for platforms.
Thoughts on this?

JAORMX:
Well, the RHCOS product merged. Can you add a check that uses it to a profile in that product?

linux_os/guide/system/auditing/bls_audit_option/rule.yml
```diff
@@ -0,0 +1,54 @@
...
+    isa-62443-2009: 4.2.3.10,4.3.2.6.7,4.3.3.3.9,4.3.3.5.8,4.3.3.6.6,4.3.4.4.7,4.3.4.5.6,4.3.4.5.7,4.3.4.5.8,4.4.2.1,4.4.2.2,4.4.2.4
+    cobit5: APO10.01,APO10.03,APO10.04,APO10.05,APO11.04,APO12.06,APO13.01,BAI03.05,BAI08.02,DSS01.04,DSS02.02,DSS02.04,DSS02.07,DSS03.01,DSS05.02,DSS05.03,DSS05.04,DSS05.07,MEA01.01,MEA01.02,MEA01.03,MEA01.04,MEA01.05,MEA02.01
+    iso27001-2013: A.11.2.6,A.12.4.1,A.12.4.2,A.12.4.3,A.12.4.4,A.12.7.1,A.13.1.1,A.13.2.1,A.14.1.3,A.15.2.1,A.15.2.2,A.16.1.4,A.16.1.5,A.16.1.7,A.6.2.1,A.6.2.2
+    cis-csc: 1,11,12,13,14,15,16,19,3,4,5,6,7,8
+    srg: SRG-OS-000254-GPOS-00095
+
+ocil_clause: 'auditing is not enabled at boot time'
+
+ocil: |-
+    Inspect the form of BLS (Boot Loader Specification) 'options' lines for the Linux operating system
```
redhatrises, line 44: nitpick: replace 'options' with <tt>options</tt>

evgenyz: Sure

evgenyz:
> With that in mind, I'm thinking that the basic building blocks for BLS should be done as Jinja macros and the templates made specific for platforms.
> Thoughts on this?

I seems to me that the most of recent bootloaders would be using BLS entries, including zIPL. And it would make sense to convert it to a Jinja template. But I have no idea how to handle that pesky `$kernelopts` case.

evgenyz:
Okay, now when we are done with `rhcos4` product it is time to get to the main course. There is a BZ (https://bugzilla.redhat.com/show_bug.cgi?id=1814210) which describes the problem we have when cluster has a RHEL (7.x in that case) node. Please also check out the link to the ML discussion in that BZ.

Now, in ocp4 profiles (let's take moderate as an example) we have the rule `grub2_audit_argument` which would check for presence of `audit=1` argument for the kernel. This rule is suitable for RHEL8 and RHEL7, but not for RHCOS. For RHOCS we have `bls_audit_option` rule (this PR).

The only way out of this I could imagine is to have separate profiles for nodes and for the cluster itself. Thoughts?

JAORMX:
> Okay, now when we are done with `rhcos4` product it is time to get to the main course. There is a BZ (https://bugzilla.redhat.com/show_bug.cgi?id=1814210) which describes the problem we have when cluster has a RHEL (7.x in that case) node. Please also check out the link to the ML discussion in that BZ.
> 
> Now, in ocp4 profiles (let's take moderate as an example) we have the rule `grub2_audit_argument` which would check for presence of `audit=1` argument for the kernel. This rule is suitable for RHEL8 and RHEL7, but not for RHCOS. For RHOCS we have `bls_audit_option` rule (this PR).
> 
> The only way out of this I could imagine is to have separate profiles for nodes and for the cluster itself. Thoughts?

that's exactly where we're heading and why I introduced the RHCOS product in the first place.

OCP4 will change to be for API/platform checks only. and node checks will use rhcos4/rhel7/rhel8 profiles respectively.

JAORMX:
> By the way, another question is how to _remediate_ RHEL7 nodes in the cluster?

This is a good question.

We could apply MachineConfigs to RHEL7 nodes just like we do with RHCOS and RHEL8 nodes... but we'll need to have MachineConfig remediations that apply and are relevant to RHEL7.

evgenyz:
> that's exactly where we're heading and why I introduced the RHCOS product in the first place.
> 
> OCP4 will change to be for API/platform checks only. and node checks will use rhcos4/rhel7/rhel8 profiles respectively.

Gosh, I just remember my introductory PR for OCP4 and regret not standing my ground. All this madness (including extra rules argument) could have been dodged.

I'm sorry. I don't know what is wrong with me. I edit comments instead if quoting them recently. :(

mildas:
**Changes identified:**
Profile moderate on rhcos4:
 Rule bls_audit_option added to moderate profile.
 Rule grub2_audit_argument removed from moderate profile.
Others:

 build_product rhcos4
 tests/test_suite.py profile --libvirt qemu:///system test-suite-vm --datastream build/ssg-rhcos4-ds.xml moderate

mildas:
**Changes identified:**
Profile moderate on rhcos4:
 Rule bls_audit_option added to moderate profile.
 Rule grub2_audit_argument removed from moderate profile.
Others:

 build_product rhcos4
 tests/test_suite.py profile --libvirt qemu:///system test-suite-vm --datastream build/ssg-rhcos4-ds.xml moderate

JAORMX:
Seems you might need to rebase.

rhcos4/profiles/moderate.profile
```diff
@@ -548,7 +548,7 @@ selections:
 
     # AU-3
     - package_audit_installed
-    - grub2_audit_argument
+    - bls_audit_option
```
yuumasato, line 551: The way how to check BLS entry options may be similar across OSes or arches, but description and OCIL can vary a lot. Take for example [`zipl_audit_argument`](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/bootloader-zipl/zipl_audit_argument/rule.yml), description and OCIL mention a few commands and config files specific for zipl.

So I wonder if the rule should be renamed to `grub2_bls_audit_option`.
And conversely, `zipl_audit_argument` renamed to `zipl_bls_audit_option`.

The template name can remain boot loader neutral, but I would put in singular: `bls_entry_option`.
@evgenyz Thoughts?

mildas:
**Changes identified:**
Profile moderate on rhcos4:
 Rule bls_audit_option added to moderate profile.
 Rule grub2_audit_argument removed from moderate profile.
Others:

 build_product rhcos4
 tests/test_suite.py profile --libvirt qemu:///system test-suite-vm --datastream build/ssg-rhcos4-ds.xml moderate

rhcos4/profiles/moderate.profile
evgenyz: Entries are plural because we are checking ALL entries: `<ind:filepath operation="pattern match">^/boot/loader/entries/.*\.conf$</ind:filepath>`.

JAORMX: oh, wouldn't it be the case that this won't pass even after the remediation is applied? The deployment keeps the previous configuration around which wouldn't have the compliant option.

evgenyz: Now, `zipl_audit_argument` (as it seems to me) holds 2 different *unrelated* checks: for option in BLS-compatible entry and for settings in /etc/zipl.conf. I thinks that they should be separated (`bls_audit_option` + `zipl_config_blah_blah`). The `bls_audit_option` is bootloader-agnostic and could be reused.

evgenyz: > oh, wouldn't it be the case that this won't pass even after the remediation is applied? The deployment keeps the previous configuration around which wouldn't have the compliant option.

Yes, it would be the case.

JAORMX: why make it so strict? it would make it almost impossible for customers to get to a compliant state.

jhrozek: I think we discussed this with Gabe some time ago and reached a compromise where it would have been enough to check the "current" bootloader settings plus the running kernel via `/proc/cpuinfo`. We can still follow up with either BLS or rpm-ostree developers to introduce a mode to replace all entries, but it shouldn't be the default anyway because you /want/ to be able to go back to the configuration before the changes were applied in case the node wouldn't boot up properly with the additional kargs.

evgenyz: Basically because of: https://github.com/ComplianceAsCode/content/issues/5285#issuecomment-617363140

But, also, this rule is a generic BLS-compatible check, it does not take into account configuration shenanigans of CoreOS.

evgenyz: Yet!

evgenyz: Later we could add something like `current=True` parameter to this template.

yuumasato: Later, this template can also be enhanced to consider cases when `$kernelopts` is present in the entry.

evgenyz: > Later, this template can also be enhanced to consider cases when `$kernelopts` is present in the entry.

I would be careful with that, it will make the rule not compatible with BLS.

evgenyz: But, on the other hand we can parametrize the template to create `grub2_bls_audit_option`.

yuumasato: The template can have conditionals for how different products handle BLS.
RHEL for example may check for all entries, not just the current.

PR #5799, opened 2020-06-01
utils: Add a helper script for creating platform checks

mrogers950:
#### Description:

Add utils/add_platform_rule.py which can be used by content developers to bootstrap new platform check content. It contains subcommands to create a new rule and verify the check criteria both in-cluster with compliance-operator and locally with podman.

Use example:
```
 $ utils/add_platform_rule.py create --rule=ocp_proxy_has_ca --type="proxies.config" --name="cluster" --yamlpath=".spec.trustedCA.name" --match="[a-zA-Z0-9]*"
 creating check for "/apis/config.openshift.io/v1/proxies/cluster" with yamlpath ".spec.trustedCA.name" satisfying match of "[a-zA-Z0-9]*"
 wrote applications/openshift/ocp_proxy_has_ca/rule.yml
 wrote applications/openshift/ocp_proxy_has_ca/oval/shared.xml

 $ utils/add_platform_rule.py cluster-test --rule=ocp_proxy_has_ca
 testing rule ocp_proxy_has_ca in-cluster
 deploying compliance-operator
 pushing image build to cluster
 waiting for cleanup from previous test run
 output from last phase check: LAUNCHING NOT-AVAILABLE
 output from last phase check: RUNNING NOT-AVAILABLE
 output from last phase check: AGGREGATING NOT-AVAILABLE
 output from last phase check: DONE COMPLIANT
 COMPLIANT
```

#### Rationale:

Remove some of the guesswork in creating OCP4 content and make manual testing easier for new content developers.

### DIFF
--- a/utils/add_platform_rule.py
+++ b/utils/add_platform_rule.py
@@ -1,0 +1,383 @@
+#!/usr/bin/python3
+
+import argparse
+import subprocess
+import pathlib
+import sys
+import textwrap
+import os
+import time
+
+PROG_DESC = (''' Create and test content files for Kubernetes API checks.
+
+This script is intended to help content writers create a new application check
+for OCP4/Kubernetes.
+
+- The 'create' subcommand creates the initial files for a new rule and fetches
+  the raw URL of the object in question (unless you specify the URL).
+
+- The 'test' subcommand builds your content locally and tests directly using an
+  openscap podman container. The scan container will test against yaml files
+  staged under --objectdir.
+
+- The 'cluster-test' subcommand pushes the content to your cluster, and then
+  runs a Platform scan for your rule with compliance-operator.
+
+Example workflow:
+
+$ utils/add_platform_rule.py create --rule=ocp_proxy_has_ca \
+  --type="proxies.config" --name="cluster" \
+  --yamlpath=".spec.trustedCA.name" --match="[a-zA-Z0-9]*"
+creating check for "/apis/config.openshift.io/v1/proxies/cluster" with yamlpath ".spec.trustedCA.name" satisfying match of "[a-zA-Z0-9]*"
+wrote applications/openshift/ocp_proxy_has_ca/rule.yml
+wrote applications/openshift/ocp_proxy_has_ca/oval/shared.xml
+
+$ mkdir -p /tmp/apis/config.openshift.io/v1/proxies/
+$ oc get proxies.config/cluster -o yaml > /tmp/apis/config.openshift.io/v1/proxies/cluster
+$ utils/add_platform_rule.py test --rule=ocp_proxy_has_ca
+testing rule ocp_proxy_has_ca locally
+Title
+        None
+Rule
+        xccdf_org.ssgproject.content_rule_ocp_proxy_has_ca
+Ident
+        CCE-84209-6
+Result
+        pass
+
+$ utils/add_platform_rule.py cluster-test --rule=ocp_proxy_has_ca
+testing rule ocp_proxy_has_ca in-cluster
+deploying compliance-operator
+pushing image build to cluster
+waiting for cleanup from previous test run
+output from last phase check: LAUNCHING NOT-AVAILABLE
+output from last phase check: RUNNING NOT-AVAILABLE
+output from last phase check: AGGREGATING NOT-AVAILABLE
+output from last phase check: DONE COMPLIANT
+COMPLIANT
+
+''')
+
+PLATFORM_RULE_DIR = 'applications/openshift'
+OSCAP_TEST_IMAGE = 'quay.io/compliance-operator/openscap-ocp:1.3.3'
+OSCAP_CMD_OPTS = 'oscap xccdf eval --verbose INFO --fetch-remote-resources --profile xccdf_org.ssgproject.content_profile_test --results-arf /tmp/report-arf.xml /content/ssg-ocp4-ds.xml'
+PROFILE_PATH = 'ocp4/profiles/test.profile'
+
+RULE_TEMPLATE = ('''prodtype: ocp4
+
+title: {TITLE}
+
+description: TBD
+
+rationale: TBD
+
+identifiers: {{}}
+
+severity: {SEV}
+
+warnings:
+    - general: |-
+        {{{{{{ openshift_cluster_setting("{URL}") | indent(8) }}}}}}
+''')
+
+
+OVAL_TEMPLATE = ('''{{{{% set YAML_TEST_OVAL_VERSION = [5, 11] %}}}}
+
+{{{{% if target_oval_version >= YAML_TEST_OVAL_VERSION %}}}}
+<def-group>
+      <definition class="compliance" version="1" id="{{{{{{ rule_id }}}}}}">
+      <metadata>
+        <title>{TITLE}</title>
+        {{{{{{- oval_affected(products) }}}}}}
+        <description>{DESC}</description>
+      </metadata>
+      <criteria operator="AND">
+        <criterion comment="{DESC}" negate="{NEGATE}" test_ref="test_{{{{{{ rule_id }}}}}}"/>
+        <criterion comment="Make sure that there is the actual file to scan" test_ref="test_file_for_{{{{{{ rule_id }}}}}}"/>
+      </criteria>
+    </definition>
+
+    <ind:yamlfilecontent_test id="test_{{{{{{ rule_id }}}}}}" check="at least one" comment="Find one match" version="1">
+            <ind:object object_ref="object_{{{{{{ rule_id }}}}}}"/>
+            <ind:state state_ref="state_{{{{{{ rule_id }}}}}}"/>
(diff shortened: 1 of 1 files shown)


mildas:
**Changes identified:**
Others:
 Python file utils/add_platform_rule.py is newly added.

mildas:
**Changes identified:**
Others:
 Python file utils/add_platform_rule.py is newly added.

mildas:
**Changes identified:**
Others:
 Python file utils/add_platform_rule.py is newly added.

mrogers950:
@jhrozek @JAORMX PTAL.

utils/add_platform_rule.py
```diff
@@ -0,0 +1,321 @@
...
+RULE_TEMPLATE = ('''prodtype: ocp4
+
+title: {TITLE}
+
+description: TBD
+
+rationale: TBD
+
+identifiers:
+    cce@ocp4: 84209-6
```
jhrozek, line 46: What does this identifier mean? Is it OK for all generated rules to have the same?

utils/add_platform_rule.py
```diff
@@ -0,0 +1,321 @@
...
+
+severity: medium
```
jhrozek, line 48: Maybe the severity should be TBD to force the creator to set it

utils/add_platform_rule.py
```diff
@@ -0,0 +1,321 @@
...
+    - general: |-
+        {{{{{{ openshift_cluster_setting("{URL}") | indent(8) }}}}}}
+''')
+
+
+OVAL_TEMPLATE = ('''{{{{% set YAML_TEST_OVAL_VERSION = [5, 11] %}}}}
+
+{{{{% if target_oval_version >= YAML_TEST_OVAL_VERSION %}}}}
+<def-group>
+      <definition class="compliance" version="1" id="{{{{{{ rule_id }}}}}}">
```
jhrozek, line 60: I wonder if this could be an OVAL macro, there are already several under `shared/`

utils/add_platform_rule.py
```diff
@@ -0,0 +1,321 @@
...
+    cluster_test_parser.add_argument(
+        '--deploy', default=True, help='Deploy the compliance-operator.')
+    cluster_test_parser.set_defaults(func=clusterTestFunc)
+
+    test_parser = subparser.add_parser(
+        'test', help='Test a rule locally against a directory of mocked object files using podman and an oscap container.')
+    test_parser.add_argument('--rule', required=True,
+                             help='The name of the rule to test.')
+    test_parser.add_argument(
+        '--file', required=True, default="build/ssg-ocp4-ds.xml", help='The path to the datastream file.')
```
jhrozek, line 311: If there's a default, the argument doesn't have to be required, right?

utils/add_platform_rule.py
```diff
@@ -0,0 +1,321 @@
...
+
+
+def testFunc(args):
+    if shutil.which('podman') is None:
+        print('podman is required')
+        return 1
+
+    print('testing rule %s locally' % args.rule)
+    ret_code, out = subprocess.getstatusoutput(
+        'podman run -it -v %s:/content/:z -v %s:/tmp/:z %s %s' % (args.file, args.objectdir, OSCAP_TEST_IMAGE, OSCAP_CMD_OPTS))
```
jhrozek, line 261: I'm having issues with this command. It seems like the objectdir is actually required, the default `/tmp` doesn't work for me, I'm getting `Error: relabel failed "/tmp": SELinux relabeling of /tmp is not allowed`. Also, just using the file as the volume doesn't work either, I'm getting errors like `Error: error creating named volume "build/ssg-ocp4-ds.xml": error running volume create option: names must match [a-zA-Z0-9][a-zA-Z0-9_.-]*: invalid argument`. Did you maybe want to share the parent directory of the DS file instead?

jhrozek: Also the test profile doesn't exist by default..

utils/add_platform_rule.py
```diff
@@ -0,0 +1,321 @@
...
+{{{{% if target_oval_version >= YAML_TEST_OVAL_VERSION %}}}}
+<def-group>
+      <definition class="compliance" version="1" id="{{{{{{ rule_id }}}}}}">
+      <metadata>
+        <title>{TITLE}</title>
+        {{{{{{- oval_affected(products) }}}}}}
+        <description>{DESC}</description>
+      </metadata>
+      <criteria operator="AND">
+        <criterion comment="{DESC}" negate="{NEGATE}" test_ref="test_{{{{{{ rule_id }}}}}}"/>
```
jhrozek, line 67: This is causing the build to fail, because the Python boolean gets converted to True/False. My crude fix was:
```
@@ -168,7 +169,7 @@ def createFunc(args):
     print('wrote ' + rule_yaml_path)
     with open(shared_xml_path, 'w') as file:
         file.write(OVAL_TEMPLATE.format(TITLE=args.title, DESC=args.description,
-                                        URL=url, YAMLPATH=args.yamlpath, MATCH=args.match, NEGATE=args.negate))
+                                        URL=url, YAMLPATH=args.yamlpath, MATCH=args.match, NEGATE=str(args.negate).lower()))
```

mildas:
**Changes identified:**
Others:
 Python file utils/add_platform_rule.py is newly added.

mildas:
**Changes identified:**
Others:
 Python file utils/add_platform_rule.py is newly added.

utils/add_platform_rule.py
mrogers950: We might be able to macrotize but I'd rather wait until we get some more rules created and find the common need macros

utils/add_platform_rule.py
mrogers950: Oh right, It's the unique CCE id for the rule. So each rule needs one from shared/references/cce-redhat-avail.txt . When you add a rule you have to cut from that file. I added a flag to be able to fill in the reference, but leaving the value and removal from cce-redhat-avail.txt up to the user for now..

mrogers950:
@JAORMX @jhrozek thanks for the reviews! The test command was indeed bogus, I've fixed it up and added a use example to the description text. Other comments are addressed as well.

utils/add_platform_rule.py
```diff
@@ -0,0 +1,383 @@
...
+RULE_TEMPLATE = ('''prodtype: ocp4
+
+title: {TITLE}
+
+description: TBD
+
+rationale: TBD
+
+identifiers:
+    {IDENT}
```
JAORMX, line 74: This wouldn't be a valid type for identifiers. a dictionary of identifiers is what's really expected here. IMO you should just leave this as an empty dict in yaml, and let the user edit it later. If not, somehow you'd need to parse the dict of identifiers.

utils/add_platform_rule.py
```diff
@@ -0,0 +1,383 @@
...
+    print('testing rule %s locally' % args.rule)
+
+    if args.build:
+        createTestProfile(args.rule)
+        ret_code, out = subprocess.getstatusoutput('./build_product ocp4')
+        if ret_code != 0:
+            print('build failed: %s' % out)
+            return 1
+
+    pod_cmd = 'podman run -it --security-opt label=disable -v "%s:/content" -v "%s:/tmp" %s %s' % (args.file, args.objectdir, OSCAP_TEST_IMAGE, OSCAP_CMD_OPTS)
```
jhrozek, line 313: The test doesn't work for me, I'm getting:
```
utils/add_platform_rule.py test --rule=ocp_proxy_has_ca
testing rule ocp_proxy_has_ca locally
  File "utils/add_platform_rule.py", line 383, in <module>
    sys.exit(main())
  File "utils/add_platform_rule.py", line 379, in main
    return args.func(args)
  File "utils/add_platform_rule.py", line 313, in testFunc
    pod_cmd = 'podman run -it --security-opt label=disable -v "%s:/content" -v "%s:/tmp" %s %s' % (args.file, args.objectdir, OSCAP_TEST_IMAGE, OSCAP_CMD_OPTS)

```

mildas:
**Changes identified:**
Others:
 Python file utils/add_platform_rule.py is newly added.

jhrozek:
Thank you this works for me. I have one more suggestion, but it's really just a suggestion, so feel free to ignore it or make use of it in a subsequent PR: I was wondering if it makes sense to add `--verbose INFO` after the `xccdf eval` because this is a devel tool and chances are you want to see some more verbose output. Or even provide an optional flag for the `test` command that sets the verbosity if needed.

But otherwise LGTM, really nice work.

mildas:
**Changes identified:**
Others:
 Python file utils/add_platform_rule.py is newly added.

mildas:
**Changes identified:**
Others:
 Python file utils/add_platform_rule.py is newly added.

mrogers950:
@jhrozek @redhatrises thanks for the reviews, I've updated this and also fixed a bug with the flags to skip building/pushing for quicker re-tests. They are now --skip-deploy and --skip-build.

PR #5800, opened 2020-06-01
Remove ocp4 prodtype from applications/openshift/api-server rules

mrogers950:
The rules that are currently defined for OCP4 under applications/openshift are incorrect; It's not feasible to check or modify the underlying static-pod files because these resources are immutable and have no API guarantees. We'll need to replace these with YAML checks, but for now I think we should remove the existing definitions.


redhatrises:
@mrogers950 actually working on updating these right now for ocp4, so I don't think that it would make sense to remove these and then re-add them in a few days if not next week.

PR #5801, opened 2020-06-02
Add RHEL datastream to ocp4 content image

JAORMX:
RHEL nodes can be OCP4 worker nodes, so let's add those datastreams to
the container image so they can be taken into use.


JAORMX:
@isimluk @jhrozek PTAL

PR #5804, opened 2020-06-03
Add support for blacklisting directories when doing system-wide file scans

matejak:
Initial support added to rpm_verify_ownership via the var_blacklisted_paths XCCDF value.

The PR **doesn't work out of the box - please help**: for some reason, the build system doesn't add the `check-export` element to the rule XCCDF, without which the external variable used in the OVAL doesn't work. If I add the `check-export` manually into the datastream, it starts to work.

I think that this mechanism could be ported to macros and used in more rules to enable users to e.g. exclude `/proc` etc. What I don't know is whether this filtering could prevent scanning (which would be really nice), or just collecting, i.e. the scanner would scan, but throw away findings and not eat more RAM, or neither, when the scanner scans, collects and waits for the end to throw everything away.
The fix would be beneficial even for the last case, as  `/proc` may cause false positives.

### DIFF
--- a/linux_os/guide/system/permissions/files/file_permissions_unauthorized_world_writable/oval/shared.xml
+++ b/linux_os/guide/system/permissions/files/file_permissions_unauthorized_world_writable/oval/shared.xml
@@ -2,13 +2,7 @@
   <definition class="compliance" id="file_permissions_unauthorized_world_writable" version="1">
     <metadata>
       <title>Find Unauthorized World-Writable Files</title>
-      <affected family="unix">
-        <platform>Red Hat Virtualization 4</platform>
-        <platform>multi_platform_ol</platform>
-        <platform>multi_platform_opensuse</platform>
-        <platform>multi_platform_rhel</platform>
-        <platform>multi_platform_wrlinux</platform>
-      </affected>
+      {{{- oval_affected(products) }}}
       <description>The world-write permission should be disabled for all files.</description>
     </metadata>
     <criteria>
@@ -31,6 +25,7 @@
     <filter action="exclude">state_file_permissions_unauthorized_world_write_exclude_special_selinux_files</filter>
     <filter action="exclude">state_file_permissions_unauthorized_world_write_exclude_proc</filter>
     <filter action="exclude">state_file_permissions_unauthorized_world_write_exclude_sys</filter>
+    <filter action="exclude">state_files_in_blacklisted_dirs</filter>
   </unix:file_object>
   <unix:file_state id="state_file_permissions_unauthorized_world_write" version="1">
     <unix:type operation="equals">regular</unix:type>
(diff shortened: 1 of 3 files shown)


mildas:
**Changes identified:**
Rule rpm_verify_ownership:
 Text changed in OVAL check.
Profile standard on fedora:
 Rule rpm_verify_ownership added to standard profile.

 build_product ol8
 tests/test_suite.py rule --libvirt qemu:///system test-suite-vm --remediate-using bash --datastream build/ssg-ol8-ds.xml rpm_verify_ownership
 build_product fedora
 tests/test_suite.py profile --libvirt qemu:///system test-suite-vm --datastream build/ssg-fedora-ds.xml standard

redhatrises:
This looks like another scanner problem and not a content problem. If OpenSCAP is scanning /proc and /sys for rpm permissions when rpms are not checking for these directories, this needs to be solved and fixed in the scanner. Remember many people use our content without using OpenSCAP and don't have this problem.

matejak:
I agree that this is not the best rule to apply this principle, as there are other rules that scan the whole filesystem, and part of the scanned directory tree may be a network mount that can't be identified without guessing.

However, a Fedora container has `/proc` and `/sys` folders owned by `nobody` and not by `root` as the `filesystem` RPM package says, so this enhancement is not completely irrelevant to this rule.

redhatrises:
> I agree that this is not the best rule to apply this principle, as there are other rules that scan the whole filesystem, and part of the scanned directory tree may be a network mount that can't be identified without guessing.
> 
> However, a Fedora container has `/proc` and `/sys` folders owned by `nobody` and not by `root` as the `filesystem` RPM package says, so this enhancement is not completely irrelevant to this rule.

This is a security and audit problem that `nobody` is used. No audit trail and `nobody` user is susceptible to attack. Sounds like this needs to be fixed on the container side.

jan-cerny:
@matejak I concur that it can be a bug in OpenSCAP, or specifically in the way in which we query the rpm database. I think the probe shouldn't read files that aren't part of any RPM package. Please post a reproducer.

matejak:
What bug do you mean? `/proc` or `/sys` are owned by other user than the RPM package prescribes, so it would be a valid finding.
Anyway, application of this principle to this rule is inappropriate, I will open a PR interacting with another rule.

jan-cerny:
@matejak My understanding was that you proposed this patch because the probes are scanning all files under /proc and /sys recursively and it took a long time or consumed a lot of memory. That's what I meant by "bug". Is it the case?

mildas:
**Changes identified:**
Rule file_permissions_unauthorized_world_writable:
 Text changed in OVAL check.
 Node deleted from OVAL check.
 Node moved within OVAL check.

 build_product ol7
 tests/test_suite.py rule --libvirt qemu:///system test-suite-vm --remediate-using bash --datastream build/ssg-ol7-ds.xml file_permissions_unauthorized_world_writable

matejak:
This PR doesn't address any specific bugs. I think that the probe queries the RPM DB, and it obtains list of files that have different ownership, so it is not like that every file is scanned and then compared to it.
Anyway, I am closing this PR, as I ported the code to a rule that causes real issues.

matejak:
See #5845 for the updated proposal - discussion over this PR moved to who should own `/proc` and whether there is an openscap bug involved, so I have opened a new PR.

PR #5811, opened 2020-06-04
Add RHEL content to quay_publish container image

JAORMX:
This is needed to have parity with the ocp4_content image.


JAORMX:
@isimluk I missed this one before. Could you check this out?

PR #5812, opened 2020-06-04
Remove result files when test scenarios pass.

ggbecker:
#### Description:

- Removes result `xml` files from `initial` and `final` steps of SSG Test Suite when **no** option `--dontclean` is selected.

#### Rationale:

- Saves up disk space.

#### Additional information:

It was a little bit confusing to identify exactly which files to remove... but deleting `self.results_path` seems to be a good start. Also I don't know exactly when `remediation` phase generates files, so if anyone knows which other files can be safely removed, feedback is welcomed.

### DIFF
--- a/tests/ssg_test_suite/oscap.py
+++ b/tests/ssg_test_suite/oscap.py
@@ -368,6 +368,19 @@ class GenericRunner(object):
 
         if result:
             LogHelper.log_preloaded('pass')
+            if self.clean_files:
+                files_to_remove = [self.verbose_path]
+                if stage in ['initial', 'final']:
+                    files_to_remove.append(self.results_path)
+
+                for fname in tuple(files_to_remove):
+                    try:
+                        if os.path.exists(fname):
+                            os.remove(fname)
+                    except OSError:
+                        logging.error(
+                            "Failed to cleanup file '{0}'"
+                            .format(fname))
         else:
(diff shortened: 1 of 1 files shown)


tests/ssg_test_suite/oscap.py
```diff
@@ -368,6 +368,13 @@ def run_stage(self, stage):
 
         if result:
             LogHelper.log_preloaded('pass')
+            if self.clean_files and stage in ['initial', 'final']:
```
matusmarhefka, line 371: Seems that xml results are generated only in `initial` and `final` stage, but there are also verbose log files and these are generated in every stage and I think we should remove them too (`self.verbose_path`).

What about doing cleanup in every stage and have a condition if it is remediation stage only remove `self.verbose_path`?

ggbecker: Ok, will try that.

matusmarhefka:
LGTM. I tested and I think this will greatly reduce the amount of generated logs and will also save disk space for automated tests. There is one more file generated for each tested rule - `RULE_ID.prescripts.log`, but this file contains output of all test scenarios setup so we cannot just remove it. We can try to address this somehow in the future, but right now I think that changes introduced in this PR are really helpful so I am going to merge. Thanks @ggbecker

PR #5813, opened 2020-06-04
Avoid duplicates in the srg mapping 

isimluk:
It has been discovered by @tedbrunell that `table-rhel8-srgmap-flat.html` may contain some duplicate records.
    
Here we amend the overlay transformation to include SSG Rule only once, even if it covers multiple SRGs.


redhatrises:
LGTM

isimluk:
Thank You :pray:

PR #5817, opened 2020-06-05
Policy source data format proposal

jan-cerny:
This is the policy source data format that we will use to improve
development of our profiles. It will allow us to store security controls
and requirements in the repository and then define profiles by using
their IDs instead of separate rules.

This is done in order to solve the problem that there is no easy way to
demonstrate to profile stakeholder the status of their profile.

Intended workflow:

* SME identifies security controls the policy consists of. Those
  controls serve as direct input for our profiles.
* SME goes through controls, and makes sure that they are sufficiently
  covered by rules.
* SME fine-tunes the profile by overriding a couple of individual rules
  in the profile file.

Once the format is accepted we can start developing tools that support
this new workflow.

In future, we can also use it for further refactoring, for example
streamlining the generation of HTML tables.

Add comments as if you review the code.

### DIFF
--- a/docs/policy_source_format.adoc
+++ b/docs/policy_source_format.adoc
@@ -1,0 +1,233 @@
+= Policy source data format
+
+The policy source data format will help us to align policy with an SCAP profile.
+Being able to understand the alignment between policy and the implementing
+profile is important for both developers and SMEs.
+
+The policy source data format adds another layer on top of profiles which stores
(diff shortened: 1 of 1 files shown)


redhatrises:
NACK.

shawndwells:
Other standards exist for this, nor is this a problem for content authors or consumers. NACK.

PR #5818, opened 2020-06-05
Add error message to the crypto policy remediation

matejak:
We have noticed that recently, the `update-crypto-policies` script has been rewritten into Python, which means that it is not available on systems without the Python environment (e.g. containers). Therefore, it may happen that the remediation for this rule can fail in many use cases.

So this PR introduces these changes to the update-crypto-policies invocation handling:

- If the script is not there, come up with a specific warning:
image

- If another error occurs, just make it clear that it happened during update-crypto-policies invocation:
image

### DIFF
--- a/linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/bash/shared.sh
+++ b/linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/bash/shared.sh
@@ -5,4 +5,17 @@
 
 populate var_system_crypto_policy
 
-update-crypto-policies --set ${var_system_crypto_policy}
+stderr_of_call=$(update-crypto-policies --set ${var_system_crypto_policy} 2>&1 > /dev/null)
+rc=$?
+
+if test "$rc" = 127; then
+	echo "$stderr_of_call" >&2
+	echo "Make sure that the script is installed on the remediated system." >&2
+	echo "See output of the 'dnf provides update-crypto-policies' command" >&2
+	echo "to see what package to (re)install" >&2
+
+	false  # end with an error code
+elif test "$rc" != 0; then
+	echo "Error invoking the update-crypto-policies script: $stderr_of_call" >&2
+	false  # end with an error code
+fi


mildas:
**Changes identified:**
Rule configure_crypto_policy:

 build_product ol8
 tests/test_suite.py rule --libvirt qemu:///system test-suite-vm --remediate-using bash --datastream build/ssg-ol8-ds.xml configure_crypto_policy

linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/bash/shared.sh
```diff
@@ -5,4 +5,21 @@
 
 populate var_system_crypto_policy
 
-update-crypto-policies --set ${var_system_crypto_policy}
+stderr_of_unsuccessful_call=$(update-crypto-policies -h 2>&1 > /dev/null)
+
+if test "$?" = 0; then
+	stderr_of_unsuccessful_call=$(update-crypto-policies --set ${var_system_crypto_policy} 2>&1 > /dev/null)
```
yuumasato, line 11: Is there a reason why not to make this call directly?

matejak: The problem with error messages is that e.g. `Error: Unknown policy ...` doesn't make it obvious what program could have said it, so I captured the error, and used it in a more explanatory error message.

jan-cerny: Optionally, you can replace the previous occurrence of `update-crypto-policies` by `which`.

yuumasato: I'm sorry I was not clear.
What I originally meant was: In line 8, why not do the actual set policy command there?
And condition if on line 17 can be based value of `$?`?

matejak: Interestingly, `which` may not be available (check out e.g. the Fedora container). Anyway, I gave it a shot, assuming that RC of 127 means command not found.

mildas:
**Changes identified:**
Rule configure_crypto_policy:

 build_product ol8
 tests/test_suite.py rule --libvirt qemu:///system test-suite-vm --remediate-using bash --datastream build/ssg-ol8-ds.xml configure_crypto_policy

PR #5819, opened 2020-06-07
Fail the build when some of html guides was not generated properly

isimluk:
In order to understand this change, let me please drive your attention to the cmake macro called `ssg_build_html_guides`. The said macro builds all the HTML guides for a given product and the guide index. The existence of the guide index makes the `make` command skip the macro and hence, we need to ensure all the guides are generated properly, before ever generating the guide index.

Previously, when guide generation failed no error was reported. But more importantly, subsequent run of the `make` did not generate the guide as the index was well in place.

Looking at the code, I get a feeling that original authors wanted to make the build error out on failure (given the exception handling code in the thread main).

:flamingo:

### DIFF
--- a/build-scripts/build_all_guides.py
+++ b/build-scripts/build_all_guides.py
@@ -96,6 +96,9 @@ def main():
     for worker in workers:
         worker.join()
 
+    if queue.unfinished_tasks > 0:
+        raise RuntimeError("Some of the guides were not exported successfully")
+
     index_source = ssg.build_guides.build_index(benchmarks, input_basename,
                                                 index_links, index_options,
                                                 index_initial_src)
--- a/ssg/build_guides.py
+++ b/ssg/build_guides.py
@@ -80,7 +80,6 @@ def builder(queue):
                 "Fatal error encountered when generating guide '%s'. "
                 "Error details:\n%s\n\n" % (guide_path, error)
             )
-            queue.task_done()
             with queue.mutex:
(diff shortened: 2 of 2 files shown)


mildas:
**Changes identified:**
Others:
 Python abstract syntax tree change found in build-scripts/build_all_guides.py.
 Python abstract syntax tree change found in ssg/build_guides.py.

isimluk:
Network timeouts.

/retest

jan-cerny:
@isimluk I have simulated the situation by breaking OpenSCAP XSLT and it fails successfully. Thank you! Out of curiosity, in which situation does this happen?

isimluk:
> @isimluk I have simulated the situation by breaking OpenSCAP XSLT and it fails successfully. Thank you! Out of curiosity, in which situation does this happen?

I have seen this failing fairly consistently in environments with insufficient memory resources. Given the enforced parallelism memory consumption of this build step is fairly significant.

jan-cerny:
@isimluk Thanks. Wouldn't it be better to stop enforcing parallelism, then?

isimluk:
Different mechanism for parallelism would be certainly welcome on my end.

However, I cannot comment on whether it is better or not. It is challenging to weight those two options against each other. One being the bug fix (with one known and perhaps many unknown triggers) other being conceptual change trying to fix the known trigger while leaving the all unknown triggers unaddressed.

PR #5822, opened 2020-06-08
WIP: Add rule that verifies if OCP4 cluster is upgradeable

JAORMX:


mildas:
**Changes identified:**
Rule ocp_cluster_upgradeable:
 OVAL check is newly added.

applications/openshift/ocp_cluster_upgradeable/oval/shared.xml
```diff
@@ -0,0 +1,48 @@
...
+            <unix:object object_ref="object_file_for_{{{ rule_id }}}"/>
+    </unix:file_test>
+
+    <unix:file_object id="object_file_for_{{{ rule_id }}}" version="1">
+      <unix:filepath var_ref="{{{ rule_id }}}_dump_location"/>
+    </unix:file_object>
+
+    <ind:yamlfilecontent_object id="object_{{{ rule_id }}}" version="1">
+      <ind:filepath var_ref="{{{ rule_id }}}_dump_location"/>
+      <ind:yamlpath>.status.conditions[type=Upgradeable].status</ind:yamlpath>
```
mrogers950, line 39: if you use `<ind:yamlpath>.status.conditions[:]</ind:yamlpath>` then the regex match can look for `'status: "True", type: Upgradeable'`

evgenyz: It won't work, the probe would only return scalar values or OVAL sets of scalar values.

redhatrises:
@JAORMX you going to get back to this?

redhatrises:
Closing as the work on this PR has stalled and hasn't been worked on in more than a while.

PR #5832, opened 2020-06-09
ocp4 e2e: Enable testing for different products

mrogers950:
This generalizes the e2e test to be able to run tests for different products.

The logic is as follows:

* If the product uses the `linux_os/guide` benchmark, it's marked as a node scan
* If the product uses the `application` benchmark, it's marked as a paltform scan

The intent is to start checking for OCP4 checks as well in its own test.

This also starts verifying that no invalid or errored checks happen. The assertion is that any check throwing an Error or None is an unexpected result.  

Co-Authored-By: @JAORMX

### DIFF
--- a/tests/ocp4e2e/Makefile
+++ b/tests/ocp4e2e/Makefile
@@ -1,5 +1,7 @@
 PROFILE?=
-TEST_TYPE?=
+# Defines the product that the test aims to test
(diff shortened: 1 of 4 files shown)


JAORMX:
With this change, a run would look as follows:

`make -f tests/ocp4e2e/Makefile e2e PROFILE=moderate PRODUCT=rhcos4`

PR #5833, opened 2020-06-09
Add OCP XCCDF CIS policy rules

redhatrises:
#### Description:

- Adds some of the CIS benchmark rules from the policies section. This is needed to submit a draft 
  benchmark for OCP4 for NCP. More PRs to come.

### DIFF
--- a/applications/openshift/accounts/accounts_restrict_service_account_tokens/rule.yml
+++ b/applications/openshift/accounts/accounts_restrict_service_account_tokens/rule.yml
@@ -20,9 +20,10 @@ severity: medium
 references:
     cis: 5.1.6
 
-#ocil_clause: 'service account tokens are mounted in pods in the cluster'
-#
-#ocil: |-
-#    Run the following command on the worker node(s):
-#    <pre>$ sudo grep -A1 anonymous /etc/kubernetes/kubernetes.conf</pre>
-#    The output should return <pre>enabled: false</pre>.
+ocil_clause: 'service account tokens are automounting'
+
+ocil: |-
+    For each pod in the cluster, review the pod specification and
+    ensure that pods that do not need to explicitly communicate with
+    the API server have <tt>automountServiceAccountToken</tt>
+    configured to <tt>false</tt>.
--- a/applications/openshift/accounts/accounts_restrict_service_account_tokens/rule.yml
+++ b/applications/openshift/accounts/accounts_restrict_service_account_tokens/rule.yml
@@ -1,0 +1,28 @@
+documentation_complete: true
+
+prodtype: ocp4
+
+title: 'Restrict Automounting of Service Account Tokens'
+
+description: |-
+    Service accounts tokens should not be mounted in pods except where the workload
+    running in the pod explicitly needs to communicate with the API server.
+    To ensure pods do not automatically mount tokens, set
+    <tt>automountServiceAccountToken</tt> to <tt>false</tt>.
+
+rationale: |-
+    Mounting service account tokens inside pods can provide an avenue
+    for privilege escalation attacks where an attacker is able to
+    compromise a single pod in the cluster.
+
+severity: medium
+
+references:
+    cis: 5.1.6
+
+#ocil_clause: 'service account tokens are mounted in pods in the cluster'
+#
+#ocil: |-
+#    Run the following command on the worker node(s):
+#    <pre>$ sudo grep -A1 anonymous /etc/kubernetes/kubernetes.conf</pre>
+#    The output should return <pre>enabled: false</pre>.
--- a/applications/openshift/api-server/api_server_anonymous_auth/rule.yml
+++ b/applications/openshift/api-server/api_server_anonymous_auth/rule.yml
@@ -7,9 +7,9 @@ title: 'Disable Anonymous Authentication to the API Server'
 description: |-
     By default, anonymous access to the OpenShift API is enabled. This
     configuration check ensures that anonymous requests to the OpenShift
-    API server are disabled. Edit the API server pod specification file
+    API server are disabled. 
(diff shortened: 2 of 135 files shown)


applications/openshift/accounts/accounts_restrict_service_account_tokens/rule.yml
```diff
@@ -0,0 +1,28 @@
...
+
+prodtype: ocp4
+
+title: 'Restrict Automounting of Service Account Tokens'
+
+description: |-
+    Service accounts tokens should not be mounted in pods except where the workload
+    running in the pod explicitly needs to communicate with the API server.
+    To ensure pods do not automatically mount tokens, set
+    <tt>automountServiceAccountToken</tt> to <tt>false</tt>.
```
JAORMX, line 11: Is the intent of this rule to check every single pod?

applications/openshift/secrets/secrets_no_environment_variables/rule.yml
```diff
@@ -0,0 +1,27 @@
+documentation_complete: true
+
+prodtype: ocp4
+
+title: 'Do Not Use Environment Variables with Secrets'
+
+description: |-
+    Secrets should be mounted as data volumes unstead of environment
```
jan-cerny, line 8:
```suggestion
    Secrets should be mounted as data volumes instead of environment
```

JAORMX:
/lgtm cancel

This is consistently failing the platform test

redhatrises:
@JAORMX 

I am seeing this which doesn't appear to be a content issue:

````
level=error
level=error msg="Error: Unable to find matching route for Route Table (rtb-01bb3be7b4d6ad8d5) and destination CIDR block (0.0.0.0/0)."
level=error
level=error msg="  on ../tmp/openshift-install-840758359/vpc/vpc-public.tf line 34, in resource \"aws_route\" \"igw_route\":"
level=error msg="  34: resource \"aws_route\" \"igw_route\" {"
level=error
level=error
````

**Comment by mildas** [comment]
**Changes identified:**
Profile cis on ocp4:
 Rule etcd_wal_dir removed from cis profile.

 build_product ocp4
 tests/test_suite.py profile --libvirt qemu:///system test-suite-vm --datastream build/ssg-ocp4-ds.xml cis

File: applications/openshift/api-server/api_server_anonymous_auth/rule.yml [hunk 3]
```diff
@@ -7,9 +7,9 @@ title: 'Disable Anonymous Authentication to the API Server'
 description: |-
     By default, anonymous access to the OpenShift API is enabled. This
     configuration check ensures that anonymous requests to the OpenShift
-    API server are disabled. Edit the API server pod specification file
+    API server are disabled. 
 {{%- if product == "ocp4" %}}
-    <tt>/etc/kubernetes/static-pod-resources/configmaps/config/config.yaml</tt> on the master node(s)
+    Edit the <tt>openshift-kube-apiserver</tt> configmap on the master node(s)
```
mrogers950, line 12: This configmap is the current running config, and whenever you try to edit this configmap manually the kube-apiserver-operator will just override it. It pieces it together through other APIs and here not a lot of it exposed for configuration purposes. It might be that most of these would have to be implemented as an unsupportedConfigOverrides (see `oc explain kubeapiserver.spec` ).

applications/openshift/kubelet/kubelet_configure_event_creation/rule.yml
```diff
@@ -9,11 +9,17 @@ description: |-
     reconstucting the chain-of-events.
 
     To prevent log creation limiting, edit the kubelet configuration
+{{%- if product == "ocp4" %}}
+    file <tt>/etc/kubernetes/kubernetes.conf</tt>
```
mrogers950, line 13: Editing a config file directly on a node is not permanent, these tunables would have to be exposed through the cluster's API (i.e., for general node configuration, that is through MachineConfigs) in order for them to stick and also be in effect for all nodes.

applications/openshift/api-server/api_server_anonymous_auth/rule.yml
JAORMX: @redhatrises let's comment out this explanation, and eventually switch it to refer to the instructions @mrogers950 is giving.

applications/openshift/kubelet/kubelet_configure_event_creation/rule.yml
JAORMX: @mrogers950 do you know a way to edit this configuration from the API side?

JAORMX:
@redhatrises the issue you were seeing is AWS being difficult :slightly_smiling_face:  gotta retest :confused:

mildas:
**Changes identified:**
Profile cis on ocp4:
 Rule etcd_wal_dir removed from cis profile.

 build_product ocp4
 tests/test_suite.py profile --libvirt qemu:///system test-suite-vm --datastream build/ssg-ocp4-ds.xml cis

applications/openshift/kubelet/kubelet_configure_event_creation/rule.yml
redhatrises: While this will be done through machine config, many users will understand that it is not configurable on the API side. What's probably going to make sense is to get all the rules into a profile, and then figure out how to correctly split out the rules that are node only vs API.

mrogers950: > While this will be done through machine config, many users will understand that it is not configurable on the API side. What's probably going to make sense is to get all the rules into a profile, and then figure out how to correctly split out the rules that are node only vs API.

Sounds good to me.

PR #5841, opened 2020-06-11
ocp4 e2e: Fix way of checking that nodes are ready

JAORMX:
The current way of checking that nodes are ready is error prone. This
attempts to fix it by also verifying that the configurations have
converged (desiredConfig is the currentconfig).

This also unties the check from MCO, and will only check this if the
MCO-expected annotation exists on the nodes.


jan-cerny:
@jhrozek What do you think about this?

JAORMX:
uhm... Seems to me that the node checks are a bit racy and we don't have enough info from the nodes themselves. Might need to start checking for the machineconfigpool instead.

PR #5849, opened 2020-06-16
Include OCP4 CPE check only into OCP4 datastream.

ggbecker:
#### Description:

- This OVAL check requires yamlfilecontent probe and that might not be
available for certain versions of OpenSCAP and it can cause the content
to fail the validation.

#### Rationale:

- Keep this check out of other datastreams than OCP4 one.

### DIFF
--- a/shared/checks/oval/installed_app_is_ocp4.xml
+++ b/shared/checks/oval/installed_app_is_ocp4.xml
@@ -3,7 +3,7 @@
     <metadata>
       <title>Red Hat OpenShift Container Platform</title>
       <affected family="unix">
-        <platform>multi_platform_all</platform>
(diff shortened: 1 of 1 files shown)


mildas:
**Changes identified:**
Rule checks:
 Text changed in OVAL check.
 Node deleted from OVAL check.
 Node moved within OVAL check.

matejak:
The "pending" test is probably just a result not being sent back to GH, as if you click it, you learn that 10/10 tests have passed.
Therefore, I am merging it.

mildas:
**Changes identified:**
Rule checks:
 Node deleted from OVAL check.
 Node moved within OVAL check.
 Text changed in OVAL check.

PR #5862, opened 2020-06-22
Convert ocp_allowed_registries_for_import into templated rule

evgenyz:


mildas:
**Changes identified:**
Rule ocp_allowed_registries_for_import:
 OVAL check for ocp_allowed_registries_for_import was deleted.

evgenyz:
@mrogers950, @ggbecker This is the evolution of yamlfile_value template and conversion of the `ocp_allowed_registries_for_import`.

I changed the `yaml_path` argument to point to a scalar value (otherwise it won't find anything). And then I `pattern match`ed it against `.*`. It might be a good idea to check if it is actually a valid URL.

mildas:
**Changes identified:**
Rule ocp_allowed_registries_for_import:
 OVAL check for ocp_allowed_registries_for_import was deleted.
Others:
 Python abstract syntax tree change found in utils/add_platform_rule.py.

mildas:
**Changes identified:**
Rule ocp_allowed_registries_for_import:
 OVAL check for ocp_allowed_registries_for_import was deleted.
Others:
 Python abstract syntax tree change found in utils/add_platform_rule.py.

evgenyz:
@mrogers950 I also changed your `pattern_check` into more flexible `operation` parameter.

docs/manual/developer_guide.adoc
```diff
@@ -1732,7 +1732,7 @@ yamlfile_value::
 ** *value* - the value to check.
 ** *type* (SimpleDatatypeEnumeration) - datatype for state's value_of, optional.
 ** *entity_check* (CheckEnumeration) - entity_check value for state's value_of, optional.
-** *pattern_check* - if set to `"true"`, the value is handled as a pattern match for state's value_of, optional.
+** *operation* (OperationEnumeration) - operation value for state's value_of, optional.
```
jan-cerny, line 1735: I suggest mentioning that the default operation is "equals".

evgenyz: Technically it would be empty (not set). As far as I can see, the document's default values are explained from the point of view of the template, i.e. what would be written into the DS by the template if parameter is omitted, not what OVAL would imply if nothing would be added there by the template.

Should we then also mention that `datatype` would be *string* by default, `entity_check` would be *all*, etc?

jan-cerny: Thanks for clarification that it's implicit OVAL value. Based on that, I think it isn't necessary to mention it.

It might improve the understanding for the template user, though.

applications/openshift/registry/ocp_allowed_registries_for_import/rule.yml
```diff
@@ -17,3 +17,13 @@ severity: medium
 warnings:
     - general: |-
         {{{ openshift_cluster_setting("/apis/config.openshift.io/v1/images/cluster") | indent(8) }}}
+
+template:
+    name: yamlfile_value
+    vars:
+        filepath: /apis/config.openshift.io/v1/oauths/cluster
```
shaneboulden, line 24: This should be `/apis/config.openshift.io/v1/images/cluster`, and not `oauths`

evgenyz: Good catch, thanks!

ggbecker:
> @mrogers950, @ggbecker This is the evolution of yamlfile_value template and conversion of the `ocp_allowed_registries_for_import`.
> 
> I changed the `yaml_path` argument to point to a scalar value (otherwise it won't find anything). And then I `pattern match`ed it against `.*`. It might be a good idea to check if it is actually a valid URL.

Apart from the requested changes by others... this LGTM. Also agree that some valid URL needs to be checked, since this value can be customized in any way, maybe we would have to support the use of variables in this case so they can override by tailoring for example.

mildas:
**Changes identified:**
Rule ocp_allowed_registries_for_import:
 OVAL check for ocp_allowed_registries_for_import was deleted.
Others:
 Python abstract syntax tree change found in utils/add_platform_rule.py.

docs/manual/developer_guide.adoc
evgenyz: Added an explanation

PR #5866, opened 2020-06-22
Make OCP4 CPE check more lax

JAORMX:
The check was verifying the version contained in the clusteroperators
resource. While this correct, it doesn't hold true in some CI
environments. This resulted in us struggling to do checks in CI...

Talking to the team in charge of building this component, they mentioned
that it's sufficient to check for the existence of this object in order
to verify that the cluster is indeed OpenShift 4.x. So this is what
we're checking now for the cpe.


mildas:
**Changes identified:**
Rule checks:
 Text changed in OVAL check.
 Node deleted from OVAL check.
 Node moved within OVAL check.

redhatrises:
One concern with this change is that we actually remove a literal version check to line up with the CPE.

JAORMX:
The OpenShift team changed the release number they use in CI. Now this shouldn't be an issue anymore.

PR #5871, opened 2020-06-23
disable_host_auth rule edit

cyarbrough76:
#### Description:

- _Changed missing_parameter_pass value to true._

#### Rationale:

- _STIG explicitly calls for the parameter to be present even though the OS default configuration is secure. STIG scans will continuously fail without it._

### DIFF
--- a/linux_os/guide/services/ssh/ssh_server/disable_host_auth/rule.yml
+++ b/linux_os/guide/services/ssh/ssh_server/disable_host_auth/rule.yml
@@ -52,7 +52,7 @@ references:
 template:
     name: sshd_lineinfile
     vars:
-        missing_parameter_pass: 'true'
+        missing_parameter_pass: 'false'
         parameter: HostbasedAuthentication
         rule_id: disable_host_auth
         value: 'no'


cyarbrough76:
I couldn't recall if there was ever a consensus on this. If we want to make them pass STIG scans, I can make the additional edits like this for similar remaining ssh server rules.

redhatrises:
The consensus of the project has been not to check secure by default settings. In other words, `HostbasedAuthentication` in `/etc/ssh/sshd_config` is configured out-of-the-box as disabled.

redhatrises:
> The consensus of the project has been not to check secure by default settings. In other words, `HostbasedAuthentication` in `/etc/ssh/sshd_config` is configured out-of-the-box as disabled. Basically, it is a bug in the STIG to not pass when this parameter is not set.

thenefield:
> > The consensus of the project has been not to check secure by default settings. In other words, `HostbasedAuthentication` in `/etc/ssh/sshd_config` is configured out-of-the-box as disabled.
> 
> Basically, it is a bug in the STIG to not pass when this parameter is not set.

Is it possible we could leave this configuration in until the STIG is updated to reflect your statement? In the meantime, it will still be considered non-compliant if not configured.

Furthermore, configuring an option that is already enforced by default isn't exactly hurting anything if the configuration is explicitly defined.

It easier just to pass the scan than to explain why the failure is a pass because a guy at Red Hat said so.

redhatrises:
> > > The consensus of the project has been not to check secure by default settings. In other words, `HostbasedAuthentication` in `/etc/ssh/sshd_config` is configured out-of-the-box as disabled.
> > 
> > 
> > Basically, it is a bug in the STIG to not pass when this parameter is not set.
> 
> Is it possible we could leave this configuration in until the STIG is updated to reflect your statement? In the meantime, it will still be considered non-compliant if not configured.
> 
> Furthermore, configuring an option that is already enforced by default isn't exactly hurting anything if the configuration is explicitly defined.
> 
> It easier just to pass the scan than to explain why the failure is a pass because a guy at Red Hat said so.

Sorry but no. This is bad security practice. We should never introduce a bug that is in someone else's content, and the ssh man page says that it is configured by default especially as the project as a whole works to make sure that default configurations are only failed if they are set to insecure settings.

cyarbrough76:
We understand that the STIG required change is unnecessary and that the STIG is technically incorrect. But this change does not actually impact the security of the system. DOD STIG scans are looking specifically for STIG compliance, not necessarily our interpretation of what the STIG should be. The current situation results in false positives by anyone using this tool. Adding an explicit definition of a parameter is not really a bug just because it is technically unneeded. We understand your position and the logic behind it but in this case it just seems more practical to meet the STIG until the STIG is corrected since it poses no actual increase of risk or impact to functionality. Either way, I will close the PR if were not making the change.

thenefield:
And this is exactly why it becomes difficult working with this user group.

The goal of this content is to bring a system compliant with the defined requirements. Regardless of the default behavior in the system, if the requirement explicitly states that a configuration should be defined, otherwise its a finding, our system and every other Red Hat system will be considered non-compliant. The end user will then need to manually go and apply the setting to meet the intent required by DoD. That defeats the purpose if this content if we cant automate checking and remediating a configuration that is required by DoD. 

Should we instead duplicate this rule with an _stig in the rule name so we can apply it this way just for the STIG? This will ensure that the default behavior takes precedence for other profiles\platforms, but still enable us to be compliant with the STIG.

redhatrises:
@cyarbrough76 @thenefield I understand your frustration, but there is a misunderstanding of how this repo works in relation and flow to the STIGs and DISA RME. This repo is considered upstream and will always differ as this is vendor provided guidance and content. Yes, the rhel6 & 7 content hasn't been updated in a while, this is in the process of being fixed. As this repo is part of the Vendor STIG process, we welcome you to contribute to make things better. However, this means that contributing to this repo will require review and conformance to vendor expectations, and part of those expectations, is making the content flexible for operations personnel. It is this content that gets turn around an submitted to DISA. There are currently conversations with DISA RME on how to do this more effectively and accurately that what is seen now. DISA themselves are moving away from configuring a default setting unless it is configured to a non-default setting. If you need an exact match to the DISA STIG with authoritative content, the DISA site provides this and is what you need.

thenefield:
If this is the intent of Red Hat, why not have a separate upstream-stig profile.

This way you can maintain what you would like it to be and what it currently is.

As a consumer of Red Hat products, this view point does not align with customer needs. Your current userbase

PR #5873, opened 2020-06-24
profile_tool.py: No Bash, no Ansible parity

yuumasato:
#### Description:

- Ansible parity percentage is 0 when there are no Bash remediations.

#### Rationale:

- Maybe it should be "N/A", but `pct` should be a number, something down the line may break if not.


mildas:
**Changes identified:**
Others:
 Python abstract syntax tree change found in ssg/build_profile.py.

PR #5875, opened 2020-06-24
Edit remediation for use_approved_ciphers

cyarbrough76:
#### Description:

- _Added RHEL6 remediation to shared.sh rather than having separate rhel6.sh_

#### Rationale:

- _The separate rhel6 remediation script shouldnt be necessary in this case._

### DIFF
--- a/linux_os/guide/services/ssh/sshd_approved_ciphers.var
+++ b/linux_os/guide/services/ssh/sshd_approved_ciphers.var
@@ -11,5 +11,10 @@ operator: equals
 interactive: false
 
 options:
+    {{#- STIG is product-dependent, so the selector has to be made available for individual products selectively -#}}
+    {{%- if product in ("rhel7", "ol7") %}}
     stig: aes128-ctr,aes192-ctr,aes256-ctr
+    {{%- elif product in ("rhel6", "ol6") %}}
+    stig: aes128-ctr,aes192-ctr,aes256-ctr,aes128-cbc,3des-cbc,aes192-cbc,aes256-cbc
+    {{%- endif %}}
     default: aes128-ctr,aes192-ctr,aes256-ctr,aes128-cbc,3des-cbc,aes192-cbc,aes256-cbc,rijndael-cbc@lysator.liu.se
--- a/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers/bash/rhel6.sh
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers/bash/rhel6.sh
@@ -1,6 +1,0 @@
-# platform = Red Hat Enterprise Linux 6
-grep -q ^Ciphers /etc/ssh/sshd_config && \
-  sed -i "s/Ciphers.*/Ciphers aes128-ctr,aes192-ctr,aes256-ctr,aes128-cbc,3des-cbc,aes192-cbc,aes256-cbc/g" /etc/ssh/sshd_config
-if ! [ $? -eq 0 ]; then
-    echo "Ciphers aes128-ctr,aes192-ctr,aes256-ctr,aes128-cbc,3des-cbc,aes192-cbc,aes256-cbc" >> /etc/ssh/sshd_config
-fi
--- a/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers/bash/shared.sh
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers/bash/shared.sh
@@ -1,4 +1,4 @@
-# platform = multi_platform_wrlinux,Red Hat Enterprise Linux 7,Oracle Linux 7
+# platform = multi_platform_wrlinux,Red Hat Enterprise Linux 7,Red Hat Enterprise Linux 6,Oracle Linux 7
 
 # Include source function library.
 . /usr/share/scap-security-guide/remediation_functions
--- a/linux_os/guide/services/ssh/sshd_approved_macs.var
+++ b/linux_os/guide/services/ssh/sshd_approved_macs.var
@@ -11,5 +11,8 @@ operator: equals
 interactive: false
 
 options:
+    {{#- STIG is product-dependent, so the selector has to be made available for individual products selectively -#}}
+    {{%- if product in ("rhel6", "rhel7", "ol6", "ol7") %}}
     stig: hmac-sha2-512,hmac-sha2-256
+    {{%- endif %}}
     default: hmac-sha2-512,hmac-sha2-256,hmac-sha1,hmac-sha1-etm@openssh.com,hmac-sha2-256-etm@openssh.com,hmac-sha2-512-etm@openssh.com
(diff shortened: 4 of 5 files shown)


mildas:
**Changes identified:**
Rule rhel6:

jan-cerny:
Hi @cyarbrough76 

Due to https://github.com/ComplianceAsCode/content/blob/b37a12fdd8593c8f3975b23bc9445a980f61f107/linux_os/guide/services/ssh/sshd_approved_ciphers.var#L15

this will mean adding the `rijndael-cbc@lysator.liu.se` cipher suite to RHEL 6 profiles.

Is that expected?

mildas:
**Changes identified:**
Rule rhel6:
Profile stig on rhel6:
 Rule  added to stig profile.

 build_product rhel6
 tests/test_suite.py profile --libvirt qemu:///system test-suite-vm --datastream build/ssg-rhel6-ds.xml stig

cyarbrough76:
> 
> 
> Hi @cyarbrough76
> 
> Due to
> 
> https://github.com/ComplianceAsCode/content/blob/b37a12fdd8593c8f3975b23bc9445a980f61f107/linux_os/guide/services/ssh/sshd_approved_ciphers.var#L15
> 
> this will mean adding the `rijndael-cbc@lysator.liu.se` cipher suite to RHEL 6 profiles.
> 
> Is that expected?

My intention was to set the stig ciphers for use with stig profile. I have now included the var setting in the rhel6 stig profile to use the stig settings. I'm not sure if the ciphers need to change for other applications (default .var settings).

matejak:
This PR reduces the list of approved ciphers to `aes128-ctr,aes192-ctr,aes256-ctr`, which may be unexpected, but I like the approach of this PR.

matejak:
ok to test

mildas:
**Changes identified:**
Rule rhel6:
Profile stig on rhel6:
 Rule  added to stig profile.

 build_product rhel6
 tests/test_suite.py profile --libvirt qemu:///system test-suite-vm --datastream build/ssg-rhel6-ds.xml stig

mildas:
**Changes identified:**
Rule rhel6:
Profile stig on rhel6:
 Rule  added to stig profile.

 build_product rhel6
 tests/test_suite.py profile --libvirt qemu:///system test-suite-vm --datastream build/ssg-rhel6-ds.xml stig

matejak:
I have introduced jinja to variable definitions, so if somebody uses the stig selector in profiles of a product that doesn't handle the list of ciphers/macs explicitly, they will get an appropriate build-time error.

redhatrises:
Please undo the Jinja changes. They aren't needed as the ciphers are the same across 6 & 7.

mildas:
**Changes identified:**
Rule rhel6:
Profile stig on rhel6:
 Rule  added to stig profile.

 build_product rhel6
 tests/test_suite.py profile --libvirt qemu:///system test-suite-vm --datastream build/ssg-rhel6-ds.xml stig

matejak:
I have streamlined the macs variable, as requested.

redhatrises:
> I have streamlined the macs variable, as requested.
> Jinja is still needed there, as it is the only way to prevent using the `stig` selector with products s.a. Fedora or Debian, i.e. products that either don't have STIG, or they have, but the project is not reflecting it.

They aren't needed, and content wise, we shouldn't be doing this. Please remove them for this PR to be merged. There is no need to prevent others from using a cipher selection based off of the STIG.

matejak:
AFAIK, there is no such thing as one size fits all STIG - there is STIG for RHEL6, for RHEL7 etc. Same rules that are selected in STIG profiles for different products even have different STIG IDs.
You are right that in the case of macs, there is only one value which is the same for RHEL6 and RHEL7, but that's because for RHEL8, this setting is done by Crypto Policies. Trying to use `sshd_approved_macs=stig` in RHEL8 should result in compile-time error, as the list of approved ciphers exists, but is different and part of the Crypto Policies package, not part of this project.
I can imagine that if there is a new STIG for Debian, Debian maintainers should edit files in question and explicitly validate that the actual list of ciphers/macs is correct. And they won't miss it if they get a compile-time error.

matejak:
@redhatrises please tell us your opinion on using `sshd_approved_ciphers=stig` on products that don't have STIG.

redhatrises:
It should 100% be allowed. There should be nothing in the build system to prevent variables from being used in this way. Just because there is no STIG for a product doesn't mean that I cannot use the variables.

mildas:
**Changes identified:**
Rule rhel6:
Profile stig on rhel6:
 Rule  added to stig profile.

 build_product rhel6
 tests/test_suite.py profile --libvirt qemu:///system test-suite-vm --datastream build/ssg-rhel6-ds.xml stig

matejak:
I see a similarity with e.g. a C compiler - that one also allows to deference null pointers. Is this what you meant?

I remove the macro for that case, as the possible inappropriate usage of the `stig` variable is not something of our concern.

linux_os/guide/services/ssh/sshd_approved_ciphers.var
```diff
@@ -11,5 +11,10 @@ operator: equals
 interactive: false
 
 options:
+    {{#- STIG is product-dependent, so the selector has to be made available for individual products selectively -#}}
+    {{%- if product in ("rhel7", "ol7") %}}
     stig: aes128-ctr,aes192-ctr,aes256-ctr
+    {{%- elif product in ("rhel6", "ol6") %}}
+    stig: aes128-ctr,aes192-ctr,aes256-ctr,aes128-cbc,3des-cbc,aes192-cbc,aes256-cbc
+    {{%- endif %}}
```
redhatrises, line 19:
```suggestion
    el7stig: aes128-ctr,aes192-ctr,aes256-ctr
    el6stig: aes128-ctr,aes192-ctr,aes256-ctr,aes128-cbc,3des-cbc,aes192-cbc,aes256-cbc
```

Variables need to be user friendly. They don't need to have a build selector for individual products. Selectors bread complexity and are generally not useful for variables.

matejak: I don't understand what you mean by "Variables ... don't need to have a build selector for individual products" while you suggest to add two build selectors for two classes of products. Could you please go into greater detail?
Regarding your approach, I generally wouldn't be against it, but in this case, it breaks backward compatibility - we use the `stig` selector for this variable in the RHEL7 STIG profile file.

redhatrises: > Regarding your approach, I generally wouldn't be against it, but in this case, it breaks backward compatibility - we use the stig selector for this variable in the RHEL7 STIG profile file.

Then use the following:

```
    stig: aes128-ctr,aes192-ctr,aes256-ctr
    el6stig: aes128-ctr,aes192-ctr,aes256-ctr,aes128-cbc,3des-cbc,aes192-cbc,aes256-cbc
```

Technically, the RHEL6 content will be considered EOL in November anyway.

redhatrises:
Requested changes haven't been made. In the next two weeks, the RHEL6 removal PR will be opened since RHEL6 is EOL. Closing.

PR #2404, opened 2017-10-10
Login banners regex

yuumasato:
#### Description:

- Fix remediation of Rule `banners_etc_issue` for RHEL7 `stig-rhel7-disa` Profile.
- Add comment explaining meaning of first banner in `dod_banners`.

#### Rationale:

- Expansion from `login_text_bannter` to actual text for remediation was not working for Values with more than one banner.

- Fixes #2327 

Thanks @sradigan for the fix.

PR #5878, opened 2020-06-24
Edit test checks for banner formatting

cyarbrough76:
#### Description:

- _Edit tests for banner_etc_issue and banner_etc_motd for banner formatting_

#### Rationale:

- _tests cases were incomplete or banner formatting was failing scan_

### DIFF
--- a/linux_os/guide/system/accounts/accounts-banners/banner_etc_motd/tests/banner_etc_motd_ospp_usbcg_banner.pass.sh
+++ b/linux_os/guide/system/accounts/accounts-banners/banner_etc_motd/tests/banner_etc_motd_ospp_usbcg_banner.pass.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+#
+# profiles = xccdf_org.ssgproject.content_profile_ospp
+
+# usgcb_default banner
+echo "-- WARNING -- This system is for the use of authorized users only. Individuals 
+using this computer system without authority or in excess of their authority 
+are subject to having all their activities on this system monitored and 
+recorded by system personnel. Anyone using this system expressly consents to 
+such monitoring and is advised that if such monitoring reveals possible 
+evidence of criminal activity system personal may provide the evidence of such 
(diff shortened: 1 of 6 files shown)


matejak:
For a bit of context - original banners were introduced in https://github.com/ComplianceAsCode/content/pull/2404, i.e. more than 2 years ago. Anyway, what was wrong with them?

cyarbrough76:
> 
> 
> For a bit of context - original banners were introduced in #2404, i.e. more than 2 years ago. Anyway, what was wrong with them?

I believe it was just spacing and line returns causing it to fail scans.

redhatrises:
The banners currently comply with what DISA provided in their notice consent document as well as what DoD CIO distributed out to the various DoD departments including what is in the SRG. These changes appear to undo that, so an actual policy document would be beneficial here.

linux_os/guide/system/accounts/accounts-banners/banner_etc_motd/tests/banner_etc_motd_ospp_usbcg_banner.pass.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+#
+# profiles = xccdf_org.ssgproject.content_profile_ospp
```
matejak, line 3: Are you sure about the profile here? `xccdf_org.ssgproject.content_profile_ospp` is not STIG, except that our RHEL8 STIG draft is based on it.

matejak:
Editing tests in this way is strange - it doesn't change the behavior of the content, and usually if we want to test for something else, we add tests, we don't delete old ones.
If the old behavior became undesirable, we should mark old scenarios as "expected failures", and also change the OVAL check.
Therefore, I propose to close this PR if this concern or @redhatrises concern aren't answered in a reasonable timeframe.

redhatrises:
PR is stale. Closing

PR #5883, opened 2020-06-24
edit bash for smartcard_configure_cert_checking

cyarbrough76:
#### Description:

- _add condition to check for and install pcsc-lite_

### DIFF
--- a/linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/smartcard_configure_cert_checking/bash/shared.sh
+++ b/linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/smartcard_configure_cert_checking/bash/shared.sh
@@ -2,6 +2,7 @@
 
 # Install required packages
 if ! rpm --quiet -q pam_pkcs11; then yum -y -d 1 install pam_pkcs11; fi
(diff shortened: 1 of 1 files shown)


mildas:
**Changes identified:**
Rule smartcard_configure_cert_checking:

 build_product ol7
 tests/test_suite.py rule --libvirt qemu:///system test-suite-vm --remediate-using bash --datastream build/ssg-ol7-ds.xml smartcard_configure_cert_checking

vojtapolasek:
Hello, thank you for the proposed change. As this rule does not have oval check, it is not possible to test it. Could you please explain why do you propose this change?

matejak:
Regarding what they write at https://pcsclite.apdu.fr/ it looks like that it is a package that is related to smart cards. Therefore, I think that it can't cause regressions, and it looks quite reasonable.

matejak:
I take it back - running `yum whatprovides /etc/pam_pkcs11/pam_pkcs11.conf` (file referenced later in the remediation) returns the `pam_pkcs11` package, `pcsc-lite` provides different files.
Therefore, the proposed change breaks the remediation, and the change hasn't been explained. Therefore, I close this PR.

PR #5889, opened 2020-06-26
Change the CPE identifier for RHCOS to indicate it's an OS

jhrozek:
#### Description:
The "part" after the cpe:// "uri" identifier was set to "a" which means
"application", but should probably say "o" to indicate that RHCOS is in
fact an OS.

#### Rationale:
This would help the Compliance Operator add proper metadata to the
objects it uses to represents profiles.


mildas:
**Changes identified:**
Rule checks:
Others:

jhrozek:
@evgenyz this is what we discussed yesterday on chat. Could you PTAL?

evgenyz:
LGTM! Thanks!

PR #5893, opened 2020-06-29
edit tests and bash for accounts_tmout

cyarbrough76:
#### Description:

- _Edited path in acounts_tmout for shared.sh and tests._

### DIFF
--- a/linux_os/guide/system/accounts/accounts-session/accounts_tmout/bash/shared.sh
+++ b/linux_os/guide/system/accounts/accounts-session/accounts_tmout/bash/shared.sh
@@ -2,9 +2,11 @@
 . /usr/share/scap-security-guide/remediation_functions
 populate var_accounts_tmout
 
-if grep --silent ^TMOUT /etc/profile ; then
-        sed -i "s/^TMOUT.*/TMOUT=$var_accounts_tmout/g" /etc/profile
+if grep --silent ^TMOUT /etc/profile.d/tmout.sh ; then
+        sed -i "s/^TMOUT.*/TMOUT=$var_accounts_tmout/g" /etc/profile.d/tmout.sh
 else
-        echo -e "\n# Set TMOUT to $var_accounts_tmout per security requirements" >> /etc/profile
-        echo "TMOUT=$var_accounts_tmout" >> /etc/profile
+        echo -e "\n# Set TMOUT to $var_accounts_tmout per security requirements" >> /etc/profile.d/tmout.sh
+        echo "TMOUT=$var_accounts_tmout" >> /etc/profile.d/tmout.sh
+        echo "readonly TMOUT" >> /etc/profile.d/tmout.sh
+        echo "export TMOUT" >> /etc/profile.d/tmout.sh
 fi
--- a/linux_os/guide/system/accounts/accounts-session/accounts_tmout/bash/shared.sh
+++ b/linux_os/guide/system/accounts/accounts-session/accounts_tmout/bash/shared.sh
@@ -2,11 +2,14 @@
 . /usr/share/scap-security-guide/remediation_functions
 populate var_accounts_tmout
 
-if grep --silent ^TMOUT /etc/profile.d/tmout.sh ; then
+if grep grep -q "^TMOUT=" /etc/profile.d/tmout.sh ; then
         sed -i "s/^TMOUT.*/TMOUT=$var_accounts_tmout/g" /etc/profile.d/tmout.sh
 else
         echo -e "\n# Set TMOUT to $var_accounts_tmout per security requirements" >> /etc/profile.d/tmout.sh
         echo "TMOUT=$var_accounts_tmout" >> /etc/profile.d/tmout.sh
         echo "readonly TMOUT" >> /etc/profile.d/tmout.sh
         echo "export TMOUT" >> /etc/profile.d/tmout.sh
+fi
+if grep grep -q "^TMOUT=" /etc/profile ; then
+        sed -i "/^TMOUT=.*/d" /etc/profile
 fi
--- a/linux_os/guide/system/accounts/accounts-session/accounts_tmout/rule.yml
+++ b/linux_os/guide/system/accounts/accounts-session/accounts_tmout/rule.yml
@@ -9,6 +9,8 @@ description: |-
     all user sessions will terminate based on inactivity. The <tt>TMOUT</tt>
     setting in <tt>/etc/profile.d/tmout.sh</tt> should read as follows:
     <pre>TMOUT=<sub idref="var_accounts_tmout" /></pre>
+    <pre>readonly TMOUT" /></pre>
+    <pre>export TMOUT" /></pre>
 
 rationale: |-
     Terminating an idle session within a short time period reduces
@@ -46,6 +48,8 @@ ocil_clause: 'value of TMOUT is not less than or equal to expected setting'
 ocil: |-
     Run the following command to ensure the <tt>TMOUT</tt> value is configured for all users
     on the system:
-    <pre>$ sudo grep TMOUT /etc/profile.d/tmout.sh</pre>
+    <pre>$ sudo grep TMOUT /etc/profile /etc/profile.d/*</pre>
     The output should return the following:
     <pre>TMOUT=<sub idref="var_accounts_tmout" /></pre>
+    <pre>readonly TMOUT" /></pre>
+    <pre>export TMOUT" /></pre>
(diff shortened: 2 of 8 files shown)


mildas:
**Changes identified:**
Rule accounts_tmout:

 build_product ol7
 tests/test_suite.py rule --libvirt qemu:///system test-suite-vm --remediate-using bash --datastream build/ssg-ol7-ds.xml accounts_tmout

redhatrises:
Only problem with this change, is that it needs to check and remediate both `/etc/profile` AND `/etc/profile.d`

vojtapolasek:
As I see it, you are moving the remediation from /etc/profile into /etc/profile.d/tmout.sh. What is the motivation behind this change? Change in STIG?

thenefield:
> Only problem with this change, is that it needs to check and remediate both `/etc/profile` AND `/etc/profile.d`

The STIG restricts the check to just /etc/profile.d/*

Why would we apply the remediation in both places?

thenefield:
> As I see it, you are moving the remediation from /etc/profile into /etc/profile.d/tmout.sh. What is the motivation behind this change? Change in STIG?

This is what the STIG and associated benchmark specifically look for.

vojtapolasek:
One more thing. Please also update the rule description and ocil to reflect the change. Thank you.

mildas:
**Changes identified:**
Rule accounts_tmout:

 build_product ol7
 tests/test_suite.py rule --libvirt qemu:///system test-suite-vm --remediate-using bash --datastream build/ssg-ol7-ds.xml accounts_tmout

mildas:
**Changes identified:**
Rule accounts_tmout:

 build_product ol7
 tests/test_suite.py rule --libvirt qemu:///system test-suite-vm --remediate-using bash --datastream build/ssg-ol7-ds.xml accounts_tmout

redhatrises:
This change needs to handle both `/etc/profile` AND `/etc/profile.d/`. You can overwrite the configuration in `/etc/profile.d/` by appending TMOUT to /etc/profile bringing the system out of compliance.

vojtapolasek:
> This change needs to handle both `/etc/profile` AND `/etc/profile.d/`. You can overwrite the configuration in `/etc/profile.d/` by appending TMOUT to /etc/profile bringing the system out of compliance.

You can't change the variable by appending to /etc/profile, because the variable in /etc/profile.d/tmout.sh is made readonly.
You would need to define the variable and make it readonly before the point where files from /etc/profile.d directory are included.
Are we going to check also for this condition?

linux_os/guide/system/accounts/accounts-session/accounts_tmout/rule.yml
```diff
@@ -5,9 +5,9 @@ prodtype: fedora,ol7,ol8,rhel6,rhel7,rhel8,rhv4,sle15,wrlinux1019
 title: 'Set Interactive Session Timeout'
 
 description: |-
-    Setting the <tt>TMOUT</tt> option in <tt>/etc/profile</tt> ensures that
+    Setting the <tt>TMOUT</tt> option in <tt>/etc/profile.d/tmout.sh</tt> ensures that
     all user sessions will terminate based on inactivity. The <tt>TMOUT</tt>
-    setting in <tt>/etc/profile</tt> should read as follows:
+    setting in <tt>/etc/profile.d/tmout.sh</tt> should read as follows:
     <pre>TMOUT=<sub idref="var_accounts_tmout" /></pre>
```
vojtapolasek, line 11: Please put here all three lines used in remediation, as I think that the part making the variable readonly is important.

vojtapolasek:
It seems that you have committed some backup file, please remove it. Thank you.

linux_os/guide/system/accounts/accounts-session/accounts_tmout/rule.yml
redhatrises: @vojtapolasek Making it readonly is a new rule and check, and it needs to go there and not in this rule.

vojtapolasek: My understanding is that this rule ensures that the TMOUT env variable is configured to some predefined value. The way proposed in latest STIG for rhel7 and also in this pull request ensures it very well, the readonly line makes sure that it is not trivial to overwrite the variable. I find this desirable. What is the reason to separate this into multiple rules?

mildas:
**Changes identified:**
Rule accounts_tmout:

 build_product ol7
 tests/test_suite.py rule --libvirt qemu:///system test-suite-vm --remediate-using bash --datastream build/ssg-ol7-ds.xml accounts_tmout

linux_os/guide/system/accounts/accounts-session/accounts_tmout/rule.yml
redhatrises: Better support in operational environments which as the vendor we are moving towards. And actually, changing this to a one-liner like `declare -xr TMOUT=600` would be way better. DISA was in favor of this alternative.

vojtapolasek: I am okay with changing the line just to declare -xrj TMOUT= something. @cyarbrough76  @thenefield  is that OK for you? I also suggest that both /etc/profile and /etc/profile.d/tmout.sh are checked, while the remediation ensures that the line exists ONLY in /etc/profile.d/tmout.sh.

linux_os/guide/system/accounts/accounts-session/accounts_tmout/bash/shared.sh
```diff
@@ -2,9 +2,14 @@
 . /usr/share/scap-security-guide/remediation_functions
 populate var_accounts_tmout
 
-if grep --silent ^TMOUT /etc/profile ; then
-        sed -i "s/^TMOUT.*/TMOUT=$var_accounts_tmout/g" /etc/profile
+if grep grep -q "^TMOUT=" /etc/profile.d/tmout.sh ; then
```
vojtapolasek, line 5:
```suggestion
if grep -q "^TMOUT=" /etc/profile.d/tmout.sh ; then
```

linux_os/guide/system/accounts/accounts-session/accounts_tmout/bash/shared.sh
```diff
@@ -2,9 +2,14 @@
...
+        sed -i "s/^TMOUT.*/TMOUT=$var_accounts_tmout/g" /etc/profile.d/tmout.sh
```
vojtapolasek, line 6:
```suggestion
        sed -i "s/^TMOUT.*/declare -xr TMOUT=$var_accounts_tmout/g" /etc/profile.d/tmout.sh
```

linux_os/guide/system/accounts/accounts-session/accounts_tmout/bash/shared.sh
```diff
@@ -2,9 +2,14 @@
...
 else
-        echo -e "\n# Set TMOUT to $var_accounts_tmout per security requirements" >> /etc/profile
-        echo "TMOUT=$var_accounts_tmout" >> /etc/profile
+        echo -e "\n# Set TMOUT to $var_accounts_tmout per security requirements" >> /etc/profile.d/tmout.sh
+        echo "TMOUT=$var_accounts_tmout" >> /etc/profile.d/tmout.sh
```
vojtapolasek, line 9:
```suggestion
        echo "declare -xr TMOUT=$var_accounts_tmout" >> /etc/profile.d/tmout.sh
```

linux_os/guide/system/accounts/accounts-session/accounts_tmout/bash/shared.sh
```diff
@@ -2,9 +2,14 @@
...
+        echo "readonly TMOUT" >> /etc/profile.d/tmout.sh
+        echo "export TMOUT" >> /etc/profile.d/tmout.sh
+fi
+if grep grep -q "^TMOUT=" /etc/profile ; then
```
vojtapolasek, line 13:
```suggestion
if grep -q "^TMOUT=" /etc/profile ; then
```

matejak:
I am closing this, as the PR
- is incomplete: The bash remediation is not complemented by OVAL or Ansible updates, and
- needs clarification: The readonly/export of the env variable has to be either agreed upon to put into one rule, or it has to be split into multiple rules that are complement each other.

As the PR author hasn't responded to reviewer feedback for a long time, it is appropriate to close the PR. Please reopen if you intend to continue its development.

PR #5905, opened 2020-07-02
Add zipl and grub2 CPEs

yuumasato:
#### Description:

- Add zipl and grub2 CPEs
- Update relevant CPE-Dictionaries

#### Rationale:
- This will make it possible to distinguish between GRUB2 and zIPL rule.

### DIFF
--- a/linux_os/guide/system/bootloader-zipl/group.yml
+++ b/linux_os/guide/system/bootloader-zipl/group.yml
@@ -8,4 +8,4 @@ description: |-
     options to it.
     The default {{{ full_name }}} boot loader for s390x systems is called zIPL.
 
-platform: machine
+platform: zipl
--- a/debian10/cpe/debian10-cpe-dictionary.xml
+++ b/debian10/cpe/debian10-cpe-dictionary.xml
@@ -27,6 +27,11 @@
             <!-- the check references an OVAL file that contains an inventory definition -->
             <check system="http://oval.mitre.org/XMLSchema/oval-definitions-5" href="filename">installed_env_has_gdm_package</check>
       </cpe-item>
+      <cpe-item name="cpe:/a:grub2">
+            <title xml:lang="en-us">Package grub2 is installed</title>
+            <!-- the check references an OVAL file that contains an inventory definition -->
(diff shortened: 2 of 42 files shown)


mildas:
**Changes identified:**
Rule checks:
 OVAL check is newly added.
Others:

linux_os/guide/system/bootloader-zipl/group.yml
```diff
@@ -8,4 +8,4 @@ description: |-
     options to it.
     The default {{{ full_name }}} boot loader for s390x systems is called zIPL.
 
-platform: machine
+platform: zipl
```
yuumasato, line 11: I've added tests for this behavior in https://github.com/OpenSCAP/openscap/pull/1551

ggbecker: That's interesting. Does it mean that rules under this group have automatically the `zipl` CPE check? This was going to be my first question when I first looked to the pull request.

yuumasato: > Does it mean that rules under this group have automatically the `zipl` CPE check?

In practice yes, I don't know if that is correct behavior though. I couldn't find details about this aspect in the XCCDF Standard.

yuumasato: What I know is that processing of an item with a platform that doesn't match should be ceased. From XCCDF 1.2, in 7.2.3.3.1 Basics:
> The processing type is Assessing, and the current platform (if known by the product) is not a
member of the set of platforms for this item. If this item is a rule, its result becomes
notapplicable (see Table 26).

This should cause all the Rules in the Group with `zipl` CPE not to be processed.
I'm not sure why the Rules in the Group would evaluate to `notapplicable`.

ggbecker: I don't understand your concern. It says that if doesn't match the platform and the item is a rule, the result becomes `notapplicable`. Am I missing something?

yuumasato:
Thank you for the through testing.

PR #5915, opened 2020-07-08
change URI of CIS benchmark to general one

vojtapolasek:
#### Description:

- URI pointing to CIS benchmark is changed so that it points to general CIS benchmark for Red Hat Enterprise Linux

#### Rationale:

- the old URL was not working anymore. Because CIS does not allow direct access to benchmark documents, the general URI is chosen.

### DIFF
--- a/rhel6/transforms/constants.xslt
+++ b/rhel6/transforms/constants.xslt
@@ -9,7 +9,7 @@
 <xsl:variable name="prod_type">rhel6</xsl:variable>
 
 <!-- Define URI of official CIS Red Hat Enterprise Linux 6 Benchmark -->
-<xsl:variable name="cisuri">https://benchmarks.cisecurity.org/tools2/linux/CIS_Red_Hat_Enterprise_Linux_6_Benchmark_v1.1.0.pdf</xsl:variable>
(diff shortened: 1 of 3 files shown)


yuumasato:
Thanks!

redhatrises:
@yuumasato @vojtapolasek  This just now introduced a bug in the content and removed CIS reference mapping to the benchmark document itself. The reference links have to be mapped to the documents themselves and not some generic url.

vojtapolasek:
> @yuumasato @vojtapolasek This just now introduced a bug in the content and removed CIS reference mapping to the benchmark document itself. The reference links have to be mapped to the documents themselves and not some generic url.

I understand your point. The problem is that although CIS benchmarks are available for everyone for free, you have to go through a form where you input your name, email, company, position etc. And they send you link to the benchmark through email. There is NO direct link to their benchmarks. Actually if you try to access the link for rhel7 CIS benchmark prior to this PR, you will find out that it does not work as expected. Probably CIS changed their web site since the URL was introduced. Any ideas? Maybe I missed something on CIS web site.

yuumasato:
> The reference links have to be mapped to the documents themselves and not some generic url.

@redhatrises I'm happy to update the reference links if you have suggestions.

PR #5920, opened 2020-07-10
Fix file regex in OCP3 content.

ggbecker:
#### Description:

- Fix file regex in OCP3 content. Yaml parser thinks that the backslash is escaping the symbol dot and
that raises an excetpion.

#### Rationale:

- Fix build of OCP3 content.

Tested on https://regex101.com/ and works as expected. Texts that begin with `.` are not matched

### DIFF
--- a/applications/openshift/ocp-permissions/ocp-files/file_permissions_master_openvswitch/rule.yml
+++ b/applications/openshift/ocp-permissions/ocp-files/file_permissions_master_openvswitch/rule.yml
@@ -32,3 +32,5 @@ template:
         filepath: /etc/origin/openvswitch/
         file_regex: ^[^.].*$
         filemode: '0644'
+    backends:
+        bash: 'off'
--- a/applications/openshift/ocp-permissions/ocp-files/file_permissions_master_openvswitch/rule.yml
+++ b/applications/openshift/ocp-permissions/ocp-files/file_permissions_master_openvswitch/rule.yml
@@ -30,5 +30,5 @@ template:
     name: file_permissions
     vars:
         filepath: /etc/origin/openvswitch/
-        file_regex: ^[^\.].*$
+        file_regex: ^[^.].*$
         filemode: '0644'


applications/openshift/ocp-permissions/ocp-files/file_permissions_master_openvswitch/rule.yml
```diff
@@ -30,5 +30,5 @@ template:
     name: file_permissions
     vars:
         filepath: /etc/origin/openvswitch/
-        file_regex: ^[^\.].*$
+        file_regex: ^[^.].*$
```
yuumasato, line 33: This doesn't work as intended for the Bash remediation.

Unlike in Ansible and OVAL, in Bash the `file_regex` is matched against the whole path, not just the base name.
`-regex` from `find` is matched against whole path. `-name` could be used, but it takes a shell pattern, not a regex.
https://github.com/ComplianceAsCode/content/blob/master/shared/templates/template_BASH_file_permissions#L7

applications/openshift/ocp-permissions/ocp-files/file_permissions_master_openvswitch/rule.yml
```diff
@@ -30,5 +30,7 @@ template:
...
         filemode: '0644'
+    backends:
+        bash: 'off'
```
ggbecker, line 36: @rmetrich Unfortunately the current file permissions bash remediation template doesn't support this `file_regex` as @yuumasato mentioned above. The bash remediation applies the regex to the full path, instead of only the base file name. Ansible remediation should be fine. Do you need the bash remediation? Or can we disable for the moment?

rmetrich: @ggbecker I guess my customers use Ansible remediation mostly, I will update here if I know more.

PR #5922, opened 2020-07-10
Remove listing functions from yaml_to_shorthand.py

jan-cerny:
This patch removes the ability of the `yaml_to_shorthand.py` script to
list inputs and outputs of shorthand generation to its stdout. It's
removing the action argument from the CLI of the script. It leaves the
build action as the only feature of the script. As a result, removing
it simplifies the code in `ssg.build_yaml` module.

Action `list-outputs` only prints the value `--output` command line
argument, therefore it is useless.

Action `list-inputs` lists some of the input files from which the
shorthand.xml will be generated. For example profile files, group files,
and rule files. I find this feature useless, as generating shorthand
involves many other files such as Jinja macro definitions, shared
checks, remediations and others.

The `list-inputs` command is used during configure time in CMake.
However, removing this step from the CMake macro is safe, because all
the files aren't generated, but are part of the source repository
(profile files, group files, and rule files are stored in git).  An
exception is the bash-remediation-functions.xml, but that is already
passed by using `DEPENDS ${BASH_REMEDIATION_FNS_DEPENDS}` in the CMake
command. As the affected CMake command already depends on compiled
profiles, it is sure that the built shorthand.xml file will contain
complete profile data.

The `list-inputs` doesn't work correctly anyway, because during CMake configuration it consumes the original profiles, ignoring the `extends:` relation, but during the build it consumes the compiled profiles.

### DIFF
--- a/build-scripts/yaml_to_shorthand.py
+++ b/build-scripts/yaml_to_shorthand.py
@@ -34,9 +34,6 @@ def parse_args():
     parser.add_argument("--output", required=True,
                         help="Output XCCDF shorthand file. "
                         "e.g.: /tmp/shorthand.xml")
-    parser.add_argument("action",
-                        choices=["build", "list-inputs", "list-outputs"],
-                        help="Which action to perform.")
     parser.add_argument("--resolved-rules-dir", "-l",
                         help="To which directory to put processed rule YAMLs.")
     parser.add_argument("--profiles-root",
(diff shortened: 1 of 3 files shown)


mildas:
**Changes identified:**
Others:
 Python abstract syntax tree change found in build-scripts/yaml_to_shorthand.py.

yuumasato:
The `list-inputs` were used to track the shorthand dependencies to allow incremental CMake builds, but support for this has been broken and incomplete for a while.
I'm okay with removing them.
But I'd love to have incremental builds again in the future, it would save a lot of development time.

> The `list-inputs` doesn't work correctly anyway, because during CMake configuration it consumes the original profiles, ignoring the `extends:` relation, but during the build it consumes the compiled profiles.

Yeah! This was very confusing for me when I was looking to extend the Profile resolution to "unselect" the rule refinements.

Issue #5874, opened 2020-06-24
Playbooks are built for unselected rules 


jan-cerny:
#### Description of problem:
The rule playbook builder ignores un-selections (using `!`) in extended profiles.

For example, RHEL 8 STIG profile extends the OSPP but unselects rule `ssh_client_rekey_limit`. However, a playbook for this rule is built in `stig` directory.

#### SCAP Security Guide Version:
current upstream as of HEAD 82bdae0c2fe038fb0c635a420428d7af62b8cac2

#### Operating System Version:
F32

#### Steps to Reproduce:

1.  ./build_product rhel8
2. ls build/rhel8/playbooks/stig

#### Actual Results:
In the aforementioned example, the directory `build/rhel8/playbooks/stig` contains `ssh_client_rekey_limit.yml` even if the rule isn't selected.

#### Expected Results:
In the aforementioned example, the directory `build/rhel8/playbooks/stig` doesn't contain `ssh_client_rekey_limit.yml` because the rule isn't selected.

#### Addition Information/Debugging Steps:
Using the ResolvableProfile class might help to fix this.

PR #5927, opened 2020-07-13
Fix build of Playbooks for extended profiles

jan-cerny:
#### Description:

ResolvableProfile class can properly resolve the extension relation in
profiles.  We don't need to have a second implementation of profile
resolution, we can reuse the ResolvableProfile class. Moreover, the
implementation in playbook_builder.py doesn't support rules removal
using exclamation mark, as described in #5874.

#### Rationale:

Fixes: #5874

### DIFF
--- a/build-scripts/compile_profiles.py
+++ b/build-scripts/compile_profiles.py
@@ -3,67 +3,9 @@ from __future__ import print_function
 import argparse
 import sys
 import os.path
-from copy import deepcopy
 from glob import glob
 
 import ssg.build_yaml
-
-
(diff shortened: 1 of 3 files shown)


mildas:
**Changes identified:**
Others:
 Python abstract syntax tree change found in build-scripts/compile_profiles.py.
 Python abstract syntax tree change found in ssg/playbook_builder.py.

jan-cerny:
@vojtapolasek @yuumasato Do you think that this is a candidate for cherry picking for 0.1.51 stabilization branch?

matejak:
Great fix that reuses and streamlines the existing code while fixing a bug.

PR #5936, opened 2020-07-15
The problem with Kerberos and OSPP is in the trusted channel requirement not allowing it, not really in the crypto.

adelton:
#### Description:

- The problem with Kerberos and OSPP is in the trusted channel requirement not allowing it, not really in the crypto.

#### Rationale:

- Linking OSPP requirements that the rules implement helps navigate the content.


redhatrises:
@openscap-ci test this please

PR #5939, opened 2020-07-16
The rules enable expected seeding, not handling of authentication failures.

adelton:
#### Description:

- The rules enable expected seeding, not handling of authentication failures.

#### Rationale:

- Linking OSPP requirements that the rules implement helps navigate the content.


yuumasato:
Thanks!

PR #5941, opened 2020-07-16
Update aide_verify_acls

ggbecker:
#### Description:

-  Update aide_verify_acls OVAL and bash to consider all cases (lowercase and uppercase).

AIDE allows configuration to use mixed cases (lowercase and uppercase) in both group definition and values. Couple of test scenarios were added to verify that it actually works with mixed cases. Also, changed the OVAL operation to also match one entry which was being ignored by only using `greater than 1`.

Reference: https://manpages.debian.org/testing/aide/aide.conf.5.en.html

#### Rationale:

- Cover all possible cases of AIDE configuration as described in the reference. If a uppercase `S` for example was in the value part (which is a valid attribute), the OVAL check simply didn't match the line for example.


mildas:
**Changes identified:**
Rule aide_verify_acls:
 Text changed in OVAL check.

 build_product ol7
 tests/test_suite.py rule --libvirt qemu:///system test-suite-vm --remediate-using bash --datastream build/ssg-ol7-ds.xml aide_verify_acls

redhatrises:
LGTM.

PR #5945, opened 2020-07-17
Update aide_verify_ext_attributes OVAL and Bash

ggbecker:
#### Description:

- Update OVAL and Bash to consider mixed cases where both group and value can have lowercase and uppercase characters.

#### Rationale:

- Similar to #5941


mildas:
**Changes identified:**
Rule aide_verify_ext_attributes:
 Text changed in OVAL check.

 build_product ol7
 tests/test_suite.py rule --libvirt qemu:///system test-suite-vm --remediate-using bash --datastream build/ssg-ol7-ds.xml aide_verify_ext_attributes

PR #5947, opened 2020-07-20
Remove OVAL check from rule install_antivirus

ggbecker:
#### Description:

- Remove OVAL check that checks for presence of McAfee antivirus.

#### Rationale:

- DISA STIG RHEL7 doesn't mention what antivirus should be installed anymore, so we cannot automatically check.

Reference: https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-72213?version=v2r7
or
https://dl.dod.cyber.mil/wp-content/uploads/stigs/zip/U_RHEL_7_V2R7_STIG.zip

#### Open Questions

What about other profiles that select this rule? Are they inline with this change? If no, how can we remove the check only for RHEL STIG profile? use something like `- install_antivirus.serity=info`?

Those are some of the questions I could not answer and would appreciate some advices.


mildas:
**Changes identified:**
Rule install_antivirus:
 OVAL check for install_antivirus was deleted.

ggbecker:
> DoD has mandated McAfee. OVAL has to stay. The OVAL is flexible enough that other A/V scanners can be added to the OVAL. The severity of high is correct. This is a requirement across FedRAMP, FISMA, and DoD.

Ok, that makes sense. I'll remove the commit that removes the OVAL check. Thanks.

redhatrises:
Great. Thank you!

PR #5951, opened 2020-07-21
Update dconf_gnome_banner_enabled to use local.d dconf database.

ggbecker:
#### Description:

- Update dconf_gnome_banner_enabled to use `local.d` dconf database.

#### Rationale:

- Open the questioning around these type of rules that still use `gdm.d` dconf databases. The thing is that RHEL8 does not `look` for databases in `gdm.d` as RHEL7 does, that means in RHEL8 by default the contents of `/usr/share/dconf/profile/gdm` are
```
$ cat /usr/share/dconf/profile/gdm
user-db:user
system-db:local
system-db:site
system-db:distro
file-db:/usr/share/gdm/greeter-dconf-defaults
```
This means it doesn't consider keys and values under `/etc/dconf/gdm.d/`. Even RHEL7 STIG says we should use `local.d`, but so far we didn't have any problems because RHEL7 still considers the `gdm.d` folder. But RHEL8 doesn't.

I would risk to say that this rule selected by [RHEL8 CIS profile](https://github.com/ComplianceAsCode/content/blob/master/rhel8/profiles/cis.profile#L237) doesn't work at all. I mean, the checks and remediation work but the actual configuration doesn't happen in the system. To make it work the user needs to include `system-db:gdm` into  `/usr/share/dconf/profile/gdm` manually.

References:
RHEL7 STIG - https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-71859?version=v2r7
RHEL8 Draft STIG - https://dl.dod.cyber.mil/wp-content/uploads/stigs/zip/U_RHEL_8_V1R0-1_IDraftSTIG.zip
Related bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1631378 (especially [comment13](https://bugzilla.redhat.com/show_bug.cgi?id=1631378#c13))

### DIFF
--- a/linux_os/guide/system/accounts/accounts-banners/gui_login_banner/dconf_gnome_banner_enabled/ansible/shared.yml
+++ b/linux_os/guide/system/accounts/accounts-banners/gui_login_banner/dconf_gnome_banner_enabled/ansible/shared.yml
@@ -5,7 +5,7 @@
 # disruption = medium
 - name: "Enable GNOME3 Login Warning Banner"
   ini_file:
-    dest: "/etc/dconf/db/gdm.d/00-security-settings"
+    dest: "/etc/dconf/db/local.d/00-security-settings"
     section: "org/gnome/login-screen"
     option: banner-message-enable
     value: "true"
@@ -14,7 +14,7 @@
 
 - name: "Prevent user modification of GNOME banner-message-enabled"
   lineinfile:
-    path: /etc/dconf/db/gdm.d/locks/00-security-settings-lock
+    path: /etc/dconf/db/local.d/locks/00-security-settings-lock
     regexp: '^/org/gnome/login-screen/banner-message-enable'
     line: '/org/gnome/login-screen/banner-message-enable'
     create: yes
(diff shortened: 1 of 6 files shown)


mildas:
**Changes identified:**
Rule dconf_gnome_banner_enabled:
 Ansible remediation changed.
 Template usage changed in ansible remediation.
 Text changed in OVAL check.

 build_product fedora
 tests/test_suite.py rule --libvirt qemu:///system test-suite-vm --remediate-using ansible --datastream build/ssg-fedora-ds.xml dconf_gnome_banner_enabled
 tests/test_suite.py rule --libvirt qemu:///system test-suite-vm --remediate-using bash --datastream build/ssg-fedora-ds.xml dconf_gnome_banner_enabled

ggbecker:
@halfline Is there any specific reason why RHEL8 doesn't include `system-db:gdm` into the `gdm` profile?

halfline:
> @halfline Is there any specific reason why RHEL8 doesn't include `system-db:gdm` into the `gdm` (`/usr/share/dconf/profile/gdm`) profile?
Just a bug:

https://bugzilla.redhat.com/show_bug.cgi?id=1833158

Should hopefully get fixed in 8.3

redhatrises:
As this is a bug identified in GDM, I don't see a need for this PR since the documentation recommends gdm.d over local.d

ggbecker:
> As this is a bug identified in GDM, I don't see a need for this PR since the documentation recommends gdm.d over local.d

I agree. We can keep all the rules that use `gdm.d` database since they are going to fix this issue on RHEL8.3.

PR #5955, opened 2020-07-23
Introduced a macro for the whole OVAL metadata section.

matejak:
Added definition of the `oval_metadata` macro, which:

- pulls the respective rule title from the XCCDF data.
- uses the `oval_affected` macro that takes care of platform metadata for most rules.
- requires definition of the check's description, as the one that is part of the XCCDF may be too long.

In other words, this macro reduces boilerplate as well as duplication of information.


redhatrises:
I would like to see the macro have a little more functionality with the ability to optionally set the title, affected family, a single platform, and/or the description without having to set indent in the macro itself.

matejak:
test this please

matejak:
test this please

jan-cerny:
@matejak I like this change very much. Is there anything else that you want to improve?

matejak:
No, I think that it is good enough as it is now.

Issue #379, opened 2015-01-13
[BugFix] Fix audit_rules_privileged_commands OVAL check to support other permissions than just exactly 'x'


iankko:
As mentioned in: https://github.com/OpenSCAP/scap-security-guide/pull/378 current version of `audit_rules_privileged_commands` OVAL check doesn't support the following scenarios:
- case when there exists rule in `/etc/audit/audit.rules` containing multiple SUID/SGID binaries listed as the `-F path=.*` arguments,
- case when particular existing `/etc/audit/audit.rules` rule contains also other permission specified in `-F perm=.*` clause other than just plain `'x'` (e.g. `'rx', 'wx'`)

In both of these cases current implementation returns `fail` even in case the system is configured properly. Therefore it's necessary to enhance the existing OVAL check to properly handle these cases.

isimluk:
The issue seems to be fixed pr #378. Can we close this?

If not, what exactly is missing?

yuumasato:
PR #1764 improved the situation but the first item is still an issue to be fixed.
Won't likely be fixed in this release.

shawndwells:
Can we use lookahead assertions?

e.g. current:
``<ind:pattern operation="pattern match">^[\s]*-a[\s]+always,exit[\s]+-F[\s]+path=\/usr\/bin\/sudo[\s]+-F[\s]+**perm=x[\s]**+-F[\s]+auid>=1000[\s]+-F[\s]+auid!=4294967295[\s]+(-k[\s]+|-F[\s]+key=)[\S]+[\s]*$</ind:pattern>``

to:
``<ind:pattern operation="pattern match">^[\s]*-a[\s]+always,exit[\s]+-F[\s]+path=\/usr\/bin\/sudo[\s]+-F[\s]+**perm=(?=.*x)[\s]**+-F[\s]+auid>=1000[\s]+-F[\s]+auid!=4294967295[\s]+(-k[\s]+|-F[\s]+key=)[\S]+[\s]*$</ind:pattern>``

yuumasato:
We can use lookahead, but any special reason to want to use it?
And I just notice we don't cover `rwx` permission in the audit rule...

shawndwells:
4-5 year old ticket - closing due to lack of activity.

redhatrises:
This is still completely valid.

marcusburghardt:
[...]

PR #5956, opened 2020-07-28
remove perm=x from rules about auditing of privileged commands

vojtapolasek:
#### Description:

remove the argument perm=x from rules audit_rules_privileged_commands_*

- templates (Ansible, Oval, Bash)

- tests (audit_rules_privileged_commands_sudo)
- rule descriptions

#### Rationale:

This is done to be alligned with latest rhel7 STIG guide. It does not introduce regression from the security point of view.


redhatrises:
@vojtapolasek for documentation purposes, does removing perm in the audit rules log read, write, execute, and attribute for audit rules? There is a ticket open for expansion to add 'r | w | a' to the audit rules in https://github.com/ComplianceAsCode/content/issues/379

vojtapolasek:
> @vojtapolasek for documentation purposes, does removing perm in the audit rules log read, write, execute, and attribute for audit rules? There is a ticket open for expansion to add 'r | w | a' to the audit rules in #379

Yes, every auditable action will be audited if perm=x is removed.

PR #5957, opened 2020-07-28
Update hipaa description

vojtapolasek:
#### Description:

Update profile description for rhel7 and rhel8 HIPAA profiles.

#### Rationale:

We do not claim that applying of the profile ensures that the system is HIPAA compliant. We help to make the system compliant.


redhatrises:
LGTM

PR #5963, opened 2020-07-29
Update kernel_module_disabled template to add modules into exclude list.

ggbecker:
#### Description:

- Removed non-existent/malformed `srg` entry: `SRG-OS-000378-GPOS-0016`. It looks like it's missing the last number, in this case `3` as you can notice from the subsequent entry: SRG-OS-000378-GPOS-0016**3**.
- Add a couple of test scenarios to validate changes.

#### Rationale:

Reference: https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-71983?version=v2r7


redhatrises:
> #### Description:
> * Update kernel_module_disabled template to add modules into exclude list.
> * Removed non-existent/malformed `srg` entry: `SRG-OS-000378-GPOS-0016`. It looks like it's missing the last number, in this case `3` as you can notice from the subsequent entry: SRG-OS-000378-GPOS-0016**3**.
> * Add a couple of test scenarios to validate changes.
> 
> #### Rationale:
> * According to STIG we also have to include the module into the exclude list as described:
> 
> ```
> Configure the operating system to disable the ability to use USB mass storage devices.
> 
> # vi /etc/modprobe.d/blacklist.conf
> 
> Add or update the line:
> 
> blacklist usb-storage
> ```
> 
> Reference: https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-71983?version=v2r7

Use of `blacklist` has been removed from the RHEL8 STIG as it doesn't do the job compared to use of `install`. This should be done with the RHEL7 STIG as well. So really on the GPOS needs and update and the tests.

ggbecker:
> 
> Use of `blacklist` has been removed from the RHEL8 STIG as it doesn't do the job compared to use of `install`. This should be done with the RHEL7 STIG as well. So really on the GPOS needs and update and the tests.

Ok, will do that.

redhatrises:
> > Use of `blacklist` has been removed from the RHEL8 STIG as it doesn't do the job compared to use of `install`. This should be done with the RHEL7 STIG as well. So really on the GPOS needs and update and the tests.
> 
> Ok, will do that.

Great! Thank you! Ack when build passes

redhatrises:
Merging as the problem is with the OCP CI

PR #5969, opened 2020-07-30
Fix syntax in OCIL checking command for accounts_user_dot_no_world_writable_programs

ggbecker:
#### Description:
- Fix syntax in OCIL checking command for accounts_user_dot_no_world_writable_programs

#### Rationale:

- Without space it throws: `find: missing argument to `-exec'`


vojtapolasek:
LGTM, thank you

PR #5959, opened 2020-07-29
Update rule dconf_gnome_screensaver_idle_activation_locked

ggbecker:
#### Description:

- Add ansible remediation for dconf_gnome_screensaver_lock_locked, fix applicability of bash remediation and add cce identifier for RHEL8

PR #5970, opened 2020-07-30
Use helper functions to install dconf and gdm.

ggbecker:
A follow up from #5959

#### Description:

- Use helper functions to install dconf and gdm.

#### Rationale:

- Make sure that rules are applicable before testing


jan-cerny:
This is a great idea!

vojtapolasek:
Thank you, nice improvement

PR #5971, opened 2020-07-30
Look for private keys in the whole file system.

ggbecker:
#### Description:

- Start discussion around the fact that we only check for private keys that are in `/etc/ssh/` and STIG says we should check the whole system for files that match a given regex.

Note: The check is not recursive and it doesn't look for all files in the filesystem. So they way it is doesn't work. It's more to open the discussion.

@redhatrises What do you think about this?

Reference: https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-72257?version=v2r7

### DIFF
--- a/linux_os/guide/services/ssh/file_permissions_sshd_private_key/rule.yml
+++ b/linux_os/guide/services/ssh/file_permissions_sshd_private_key/rule.yml
@@ -43,7 +43,7 @@ ocil: '{{{ ocil_file_permissions(file="/etc/ssh/*_key", perms="-rw-r-----") }}}'
 template:
     name: file_permissions
     vars:
-        filepath: /etc/ssh/
+        filepath: /
         missing_file_pass: 'true'
         file_regex: ^.*_key$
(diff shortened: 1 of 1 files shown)


matejak:
I think that this is a bug in STIG. If we assume that *host* keys are not in their default location, then we should also assume that they can have a custom filename, and searching for them in the whole system is therefore annoying (takes a lot of time and resources of the scanning machine) and pointless (we don't know what exactly are we searching for).
As parsing the config file and checking the configured keys in one rule is very difficult to write in OVAL (files can get included), I would advocate for two rules - one that keys are configured to be in `/etc/ssh` and that they have the expected filename pattern, and the second rule checking those files.

redhatrises:
Agreed that this is a STIG bug and don't see us moving forward with these changes as the vendor guidance is running and configuring ssh for `/etc/ssh` use.

ggbecker:
Thanks for the comments.

PR #5975, opened 2020-07-31
Remove rules that are not present on RHEL STIG v2r7 anymore.

ggbecker:
#### Description:

- Remove RHEL7 STIG rules that are not present in the benchmark anymore. Version v2r7 of the benchmark.

References:
https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-72143?version=V1R4&compareto=v2r7
https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-72169?version=V1R4&compareto=v2r7
https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-72271?version=V1R4&compareto=v2r7

### DIFF
--- a/rhel7/profiles/stig.profile
+++ b/rhel7/profiles/stig.profile
@@ -193,7 +193,6 @@ selections:
     - audit_rules_execution_setsebool
     - audit_rules_execution_chcon
     - audit_rules_execution_setfiles
-    - audit_rules_login_events_tallylog
     - audit_rules_login_events_faillock
     - audit_rules_login_events_lastlog
     - audit_rules_privileged_commands_passwd
@@ -206,7 +205,6 @@ selections:
     - audit_rules_sysadmin_actions
     - audit_rules_privileged_commands_newgrp
     - audit_rules_privileged_commands_chsh
-    - audit_rules_privileged_commands_sudoedit
     - audit_rules_media_export
     - audit_rules_privileged_commands_umount
     - audit_rules_privileged_commands_postdrop
@@ -254,7 +252,6 @@ selections:
     - sshd_use_priv_separation
     - sshd_disable_compression
     - chronyd_or_ntpd_set_maxpoll
-    - configure_firewalld_rate_limiting
     - service_firewalld_enabled
     - display_login_attempts
     - no_user_host_based_files


mildas:
**Changes identified:**
Profile stig on rhel7:
 Rule configure_firewalld_rate_limiting, audit_rules_privileged_commands_sudoedit, audit_rules_login_events_tallylog removed from stig profile.

 build_product rhel7

rhel7/profiles/stig.profile
```diff
@@ -206,7 +205,6 @@ selections:
     - audit_rules_sysadmin_actions
     - audit_rules_privileged_commands_newgrp
     - audit_rules_privileged_commands_chsh
-    - audit_rules_privileged_commands_sudoedit
```
redhatrises, line 209: This seems like a bug and probably shouldn't be removed.

rhel7/profiles/stig.profile
```diff
@@ -254,7 +252,6 @@ selections:
     - sshd_use_priv_separation
     - sshd_disable_compression
     - chronyd_or_ntpd_set_maxpoll
-    - configure_firewalld_rate_limiting
```
redhatrises, line 257: This is definitely a bug as SRG-OS-000420-GPOS-00186 references rate limiting.

ggbecker: This has been explicitly removed in STIG v2r4 as can be seen in the changelog and benchmark here: https://dl.dod.cyber.mil/wp-content/uploads/stigs/zip/U_RHEL_7_V2R8_STIG.zip

`- V-72271 - Removed this requirement from the STIG.`

Should we keep the rule even if the benchmark says otherwise?

rhel7/profiles/stig.profile
ggbecker: This one can be a bug since it got dropped between v1r4 and v2r1, similar to `audit_rules_login_events_tallylog`.

rhel7/profiles/stig.profile
redhatrises: Rate limiting is a form of Denial of Service protection internally and externally to the system. As the plan is to resubmit the RHEL7 content, we shouldn't drop rules unless it is no longer a technical feasibility e.g. dconf is used instead of gconf.

ggbecker:
It seems that we don't want to remove rules from STIG profile even if it is not present in the benchmark anymore. Closing for now.

PR #5986, opened 2020-08-06
Update STIG id of grub2_password

ggbecker:
#### Description:

~~- Since v2r1, RHEL7 STIG has a specific item to check for the `GRUB2_PASSWORD` in `/boot/grub2/user.cfg` ([V-81005](https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-81005?version=v2r7)). So far we have this check embedded with the `grub2_password` rule from [V-71961](https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-71961?version=v2r7), the ideia here is to open the discussion if we should split into two separate rules.~~

- Update STIG id of `grub2_password` to match STIG item that is RHEL >= 7.2 only ([V-81005](https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-81005?version=v2r7)).

Similarly this apply for the UEFI related rules.


mildas:
**Changes identified:**
Rule grub2_password:
 Text changed in OVAL check.
 Node deleted from OVAL check.
 Node moved within OVAL check.
Rule grub2_superuser_password:
 OVAL check is newly added.

 build_product fedora
 tests/test_suite.py rule --libvirt qemu:///system test-suite-vm --remediate-using bash --datastream build/ssg-fedora-ds.xml grub2_password

redhatrises:
What is the correct way to handle configure grub passwords? Obviously, there are multiple ways to configure, but we need to define the 1 correct way of checking this.

ggbecker:
> What is the correct way to handle configure grub passwords? Obviously, there are multiple ways to configure, but we need to define the 1 correct way of checking this. So, if user.cfg is the correct way, creating a new rule rather than fixing the broken one probably isn't the correct way.

Actually both ways are correct. The thing is that one check is applicable only to RHEL < 7.2 and the other to RHEL >= 7.2. One quick way to solve this would be to add both STIG ID to the same rule and that's would be it.

redhatrises:
As we are now at RHEL 7.9, let's fix this for RHEL >= 7.2.

ggbecker:
> As we are now at RHEL 7.9, let's fix this for RHEL >= 7.2.

The current rule already checks for RHEL >= 7.2 configuration, but also for RHEL < 7.2. The selected STIG ID is from the item that is RHEL < 7.2. So the proposal can be just to change the STIG ID to reference the item which covers RHEL >= 7.2.

redhatrises:
> > As we are now at RHEL 7.9, let's fix this for RHEL >= 7.2.
> 
> The current rule already checks for RHEL >= 7.2 configuration, but also for RHEL < 7.2. The selected STIG ID is from the item that is RHEL < 7.2. So the proposal can be just to change the STIG ID to reference the item which covers RHEL >= 7.2.

Sounds good. Should also do a spot check to verify that the rule.yml is correct for RHEL >= 7.2

ggbecker:
> > > As we are now at RHEL 7.9, let's fix this for RHEL >= 7.2.
> > 
> > 
> > The current rule already checks for RHEL >= 7.2 configuration, but also for RHEL < 7.2. The selected STIG ID is from the item that is RHEL < 7.2. So the proposal can be just to change the STIG ID to reference the item which covers RHEL >= 7.2.
> 
> Sounds good. Should also do a spot check to verify that the rule.yml is correct for RHEL >= 7.2

The only changed was to update the STIG id. Tests scenarios already cover correctness of rule for RHEL >= 7.2.

PR #5987, opened 2020-08-07
Select uefi_no_removeable_media in DISA RHEL7 STIG profile

ggbecker:
#### Description:

- Select uefi_no_removeable_media in DISA RHEL7 STIG profile

#### Rationale:

- The UEFI counterpart of this rule was not being selected by the profile and didn't contain some of the references.
- DISA STIG reference: https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-72075?version=v2r7

### DIFF
--- a/linux_os/guide/system/bootloader-grub2/uefi_no_removeable_media/rule.yml
+++ b/linux_os/guide/system/bootloader-grub2/uefi_no_removeable_media/rule.yml
@@ -23,6 +23,7 @@ identifiers:
 references:
     disa: CCI-001814
     srg: SRG-OS-000364-GPOS-00151
+    stigid@rhel7: RHEL-07-021700
 
 ocil_clause: 'it is not'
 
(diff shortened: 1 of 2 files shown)


mildas:
**Changes identified:**
Profile stig on rhel7:
 Rule uefi_no_removeable_media added to stig profile.

 build_product rhel7

redhatrises:
Technically, there should be a separate STIG id for this. Personally, I wish there was a single grub.cfg file vs efi and others etc. makes for a bad user experience.

ggbecker:
> Technically, there should be a separate STIG id for this. Personally, I wish there was a single grub.cfg file vs efi and others etc. makes for a bad user experience.

That would imply to send a request to DISA and they would consider this to next versions, right? I think we can select it for now and when this gets its own item we update accordingly. Otherwise I don't know what to do.

PR #5991, opened 2020-08-07
Remove extra zero on SRG ref mapping from kernel_module_dccp_disabled.

ggbecker:
#### Description:

- Remove extra zero on SRG ref mapping from kernel_module_dccp_disabled.

#### Rationale:

- Reference:https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-77821?version=V1R4&compareto=v2r7


redhatrises:
Yup, we may have more of those type of typos. When I looked at this PR, my gut reaction was to say that it ends in 000163 which is incorrect compared to the SRG.

ggbecker:
> Yup, we may have more of those type of typos. When I looked at this PR, my gut reaction was to say that it ends in 000163 which is incorrect compared to the SRG.

It looks like it is/was a bug in the Oracle Linux STIG benchmark: https://vaulted.io/library/disa-stigs-srgs/oracle_linux_7_security_technical_implementation_guide/V-99165

PR #5995, opened 2020-08-10
Update ansible additional when statement to fix issues with rules not being applied to vm's

### LINKED ISSUE #5048: DISA RHEL 7 STIG FIPS Rules incorrectly applied to virtual machines

carlosmmatos:
#### Description:

Update additional when clause and introduce new container checks.

#### Rationale:

- Additional when clause was confusing and not targeting a big issue - ensuring certain tasks don't run on containers. Trying to approach this - it became apparent that we needed another variable to be added to playbooks that contains a list of containers (as seen in upstream https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/facts/virtual/linux.py). This allows us to target the list in a clean manner when doing our `when:` clause check.

- Fixes #5048

### DIFF
--- a/ssg/build_remediations.py
+++ b/ssg/build_remediations.py
@@ -346,8 +346,7 @@ class AnsibleRemediation(Remediation):
     def update_when_from_rule(self, to_update):
         additional_when = ""
         if self.associated_rule.platform == "machine":
-            additional_when = ('ansible_virtualization_role != "guest" '
-                               'or ansible_virtualization_type != "docker"')
+            additional_when = 'ansible_virtualization_type not in ["docker", "lxc", "openvz"]'
         to_update.setdefault("when", "")
         new_when = ssg.yaml.update_yaml_list_or_string(to_update["when"], additional_when)
         if not new_when:
--- a/tests/unit/ssg-module/data/ansible-resolved.yml
+++ b/tests/unit/ssg-module/data/ansible-resolved.yml
@@ -2,7 +2,7 @@
   stat:
     path: /boot/grub2/grub.cfg
   register: file_exists
-  when: ansible_virtualization_role != "guest" or ansible_virtualization_type != "docker"
+  when: ansible_virtualization_type not in ["docker", "lxc", "openvz"]
   tags:
     - file_owner_grub2_cfg
     - medium_severity
@@ -22,7 +22,7 @@
     owner: 0
   when:
     - file_exists.stat.exists
-    - ansible_virtualization_role != "guest" or ansible_virtualization_type != "docker"
+    - ansible_virtualization_type not in ["docker", "lxc", "openvz"]
   tags:
     - file_owner_grub2_cfg
     - medium_severity


mildas:
**Changes identified:**
Others:
 Python abstract syntax tree change found in ssg/ansible.py.
 Python abstract syntax tree change found in ssg/build_profile_remediations.py.

redhatrises:
So far looks good to me, the build is failing for the `test_ansible_conformance` tests, so make sure to update the tests.

ssg/build_remediations.py
```diff
@@ -346,8 +346,7 @@ def _get_rule_reference(self, ref_class):
     def update_when_from_rule(self, to_update):
         additional_when = ""
         if self.associated_rule.platform == "machine":
-            additional_when = ('ansible_virtualization_role != "guest" '
-                               'or ansible_virtualization_type != "docker"')
+            additional_when = ('ansible_virtualization_role not in container_list')
```
yuumasato, line 349: When generating the Ansible Playbook with `oscap xccdf generate fix --fix-type ansible --profile pci-dss --output my.yml ./ssg-rhel8-ds.xml`, the `container_list` var is not defined.
For this use case to work, every Ansible snippet that add the `when` clause for `container_list`, also needs to define it.

There exists a syntax to make Ansible variables out of XCCDF variables, `xccdf-var`, consumed [here](https://github.com/ComplianceAsCode/content/blob/6a25c66822e34abc6aaa7346c362cb48032784ca/ssg/build_remediations.py#L559).
It takes an XCCDF Var and puts it into a syntax that openscap understands, so that it can handle it correctly when generating an Ansible fix.
Maybe there could be a variant of this that would just place a fixed value, `container-list-var`?

With the above, injecting `container_list` in `build_profile_remediations.py` with `add_extra_vars()`, may not be needed, as it would be handled like the other Ansible variables (testing needed).

ssg/ansible.py
```diff
@@ -39,6 +39,26 @@ def add_minimum_version(ansible_src):
...
 
+def add_extra_vars(ansible_src):
+    """
+    Adds extra variables to playbooks. Contains a pre-determined set of
+    additional variables to be passed in.
+    """
+    # TBD: Remove this function if https://github.com/ansible/ansible/pull/70697
+    # gets meged upstream - in future Ansible release.
+    extra_vars = (""" vars:
+      container_list:
```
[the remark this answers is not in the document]
carlosmmatos: I'll double check this again..

ssg/build_remediations.py
carlosmmatos: Interesting - so If I'm understanding you correctly, there is a potential to make this (or future additional vars) more consumable by turning it into an XCCDF var? I'll look more into this today. Thanks for pointing this out.

yuumasato: Not necessarily an XCCDF Var, but it can be added statically.

mildas:
**Changes identified:**
Others:

carlosmmatos:
After discussing more - and with current Ansible limitations, this change should reflect a better way of testing against current Ansible container virtual facts.

ggbecker (changes requested):
There is one failing test (python-unit-ssg-module), you need to update the following file with `ansible_virtualization_type not in ["docker", "lxc", "openvz"]` in both `when` statements:

https://github.com/ComplianceAsCode/content/blob/master/tests/unit/ssg-module/data/ansible-resolved.yml

ssg/build_remediations.py
```diff
@@ -346,8 +346,7 @@ def _get_rule_reference(self, ref_class):
     def update_when_from_rule(self, to_update):
         additional_when = ""
         if self.associated_rule.platform == "machine":
-            additional_when = ('ansible_virtualization_role != "guest" '
-                               'or ansible_virtualization_type != "docker"')
+            additional_when = 'ansible_virtualization_type not in ["docker", "lxc", "openvz"]'
```
yuumasato, line 349: How about `podman` and `container` types?

carlosmmatos: @yuumasato current versions of Ansible don't know what Ansible is, so I didn't want to mislead anyone in thinking that we are somehow checking for it. Make sense? This won't be available until 2.10 stable release. I'll create an issue to track that when stable 2.10 is available, we can make the changes.

carlosmmatos:
> There is one failing test (python-unit-ssg-module), you need to update the following file with `ansible_virtualization_type not in ["docker", "lxc", "openvz"]` in both `when` statements:
> 
> https://github.com/ComplianceAsCode/content/blob/master/tests/unit/ssg-module/data/ansible-resolved.yml

working on it now.

mildas:
**Changes identified:**
Others:

redhatrises:
Looks good. Bugs in ansible have to be fixed. We cannot fix that here.

ssg/build_remediations.py
yuumasato: Makes sense, for the `podman` and `container` types.

Ping  @matejak for awareness of the change in `ansible_virtualization_role`.

PR #5998, opened 2020-08-12
remove nonexistent srg from audit_rules_usergroup_modification_opasswd

vojtapolasek:
#### Description:

#### Rationale:

stig effort


matejak:
ACK, I can't find any evidence that the SRG exists.

PR #6000, opened 2020-08-13
Update network_sniffer_disabled

vojtapolasek:
#### Description:

- minor change in ocil

- add tests

#### Rationale:

per https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-72295?version=v2r7

stig effort

### DIFF
--- a/linux_os/guide/system/network/network_sniffer_disabled/rule.yml
+++ b/linux_os/guide/system/network/network_sniffer_disabled/rule.yml
@@ -46,4 +46,4 @@ ocil_clause: 'any network device is in promiscuous mode'
 
 ocil: |-
     Promiscuous mode of an interface can be disabled with the following command:
-    <pre>$ sudo ip link set dev <tt>device_name</tt> promisc off</pre>
+    <pre>$ sudo ip link set dev <tt>device_name</tt> multicast off promisc off</pre>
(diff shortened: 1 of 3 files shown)


redhatrises:
Since we are updating the rule, let's update the description while we are here.

```
    The system should not be acting as a network sniffer, which can
    capture all traffic on the network to which it is connected. Run the following
    to determine if any interface is running in promiscuous mode:
    <pre>$ ip link | grep PROMISC</pre>
    Promiscuous mode of an interface can be disabled with the following command:
￼    <pre>$ sudo ip link set dev <tt>device_name</tt> promisc off</pre>
```

linux_os/guide/system/network/network_sniffer_disabled/rule.yml
```diff
@@ -46,4 +46,4 @@ ocil_clause: 'any network device is in promiscuous mode'
 
 ocil: |-
     Promiscuous mode of an interface can be disabled with the following command:
-    <pre>$ sudo ip link set dev <tt>device_name</tt> promisc off</pre>
+    <pre>$ sudo ip link set dev <tt>device_name</tt> multicast off promisc off</pre>
```
redhatrises, line 49: This needs to be updated to be more like OCIL:

```suggestion
    To verify that Promiscuous mode of an interface is disabled, run the following command:
    <pre>$ ip link | grep PROMISC</pre>
    The output returned should not return any network device containing <tt>PROMISC</tt>.
```

vojtapolasek:
I have modified ocil and description.

PR #6006, opened 2020-08-17
ocp4/e2e test: Take into account if node is unschedulable

JAORMX:
If node is unschedulable, the node clearly isn't ready. So let's take
into account in the test.


jhrozek (approved):
/lgtm

PR #6014, opened 2020-08-19
Check sssd conf.d files and fix bash remediation for sssd_enable_pam_services

ggbecker:
#### Description:
Rule `sssd_enable_pam_services`:
- Update OVAL check to consider `/etc/sssd/conf.d/` directory and to correctly detect the configuration.
- Fix BASH remediation. Use `awk` combined with `grep` and `uniq` to get exact lines that need to be changed in the configuration file. Inspired from: https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/sssd/sssd_run_as_sssd_user/bash/shared.sh

#### Rationale:

- Better align with [V-72427](https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-72427?version=v2r7)

#### Help needed:

~~- Since `/etc/sssd/conf.d/` can be empty, we need a way to skip the check in this case. Using `existence=any_exist` does not help because if a given file exists and it doesn't comply with the requirement it will evaluate as `True`.~~
~~- Since we are using this multi line regex, I didn't find a way to combine this with `sed` to use in the remediation. Right now it will look for the line `services = ` and append pam when needed.~~
~~- Maybe we don't need to be strict in many ways so we could drop the check of `/etc/sssd/conf.d/`, but the remediation fix part is definitely something that needs to be fixed because the it is right now, it will always append `services = pam` to the file.~~

~~Note: to test this it is needed to have `sssd` installed. This is something to be added to test scenarios yet.~~


mildas:
**Changes identified:**
Rule sssd_enable_pam_services:
 Node deleted from OVAL check.
 Text changed in OVAL check.
 Node moved within OVAL check.

 build_product ol7
 tests/test_suite.py rule --libvirt qemu:///system test-suite-vm --remediate-using bash --datastream build/ssg-ol7-ds.xml sssd_enable_pam_services

linux_os/guide/services/sssd/sssd_enable_pam_services/oval/shared.xml
```diff
@@ -2,28 +2,40 @@
...
+    <ind:filename operation="pattern match">{{{ file_pattern }}}</ind:filename>
+    {{% endif %}}
     <ind:pattern operation="pattern match">^[\s]*\[sssd]([^\n\[\]]*\n+)+?[\s]*services.*pam.*$</ind:pattern>
     <ind:instance datatype="int">1</ind:instance>
   </ind:textfilecontent54_object>
+  {{% endmacro %}}
+
+  {{{ check_sssd_pam_test("file", "/etc/sssd/sssd.conf") }}}
+  {{{ check_sssd_pam_test("dir", "/etc/sssd/conf.d/", "^.*\.conf$") }}}
+
```
redhatrises, line 40: I don't think we need all these checks. A simple regex in `<ind:filepath>` will do.

jan-cerny:
> Since /etc/sssd/conf.d/ can be empty, we need a way to skip the check in this case. Using existence=any_exist does not help because if a given file exists and it doesn't comply with the requirement it will evaluate as True.

A single regex which describes all possible locations together with check_existence="at_least_one_exists". Can that work?

ggbecker:
> > Since /etc/sssd/conf.d/ can be empty, we need a way to skip the check in this case. Using existence=any_exist does not help because if a given file exists and it doesn't comply with the requirement it will evaluate as True.
> 
> A single regex which describes all possible locations together with check_existence="at_least_one_exists". Can that work?

That might improve the situation, but I'm not sure if it would entirely solve it. If I understood correctly, if there is no `services = something` at all, the check should pass (according to STIG), so in this case it would be `check_existence="any_exist"`, but then if it happens that there is a `services = nss` (without `pam`) line in one file for example, it would pass even if it doesn't comply with the check because of the `any_exist`, is that a correct assumption?

I think we should split the check in two then and make one to check if there is any `services = something` line evaluating to true if no entry is found and the other one checking for the actual values of `services = something`. And combine then with OR operation.

yuumasato (commented):
>     * Since `/etc/sssd/conf.d/` can be empty, we need a way to skip the check in this case. Using `existence=any_exist` does not help because if a given file exists and it doesn't comply with the requirement it will evaluate as `True`.

I think the problem is that the test is already looking for the object with the desired state. If the test has an explicit state it can work. 

For example, if the object is the services line, and we capture the services list:
```<ind:pattern operation="pattern match">^[\s]*\[sssd]([^\n\[\]]*\n+)+?[\s]*services[\s]*=[\s]*(.*)$</ind:pattern> ```
And then state makes sure that `pam` is one of the captured services:
```  <ind:textfilecontent54_state id="" version="1">
    <ind:subexpression operation="pattern_match">.*pam.*</ind:subexpression>
```
Edit: added explanation below.

This would work with `check_existence="any_exist"`, any number of `services = ` line can be found.
And with `check="all"`, all objects need to conform to the state, i.e. any `services = ` line found need to contain `pam`.

ggbecker:
> <ind:subexpression operation="pattern_match">.*pam.*</ind:subexpression>

Thanks for the thoroughly explanation, I'll run some tests based on feedback and then I'll submit new changes.

linux_os/guide/services/sssd/sssd_enable_pam_services/oval/shared.xml
ggbecker: Done. Thanks.

ggbecker:
I have pushed new changes but the OVAL is not working properly, I can't make the check pass, even with `sssd.conf` containing correct content. @yuumasato can you take a look at the OVAL check if you can find any discrepancies? Thank you.

linux_os/guide/services/sssd/sssd_enable_pam_services/oval/shared.xml
```diff
@@ -2,28 +2,29 @@
...
     <ind:object object_ref="obj_sssd_enable_pam_services" />
+    <ind:state state_ref="state_sssd_enable_pam_services" />
   </ind:textfilecontent54_test>
+
   <ind:textfilecontent54_object id="obj_sssd_enable_pam_services" version="1">
-    <ind:filepath>/etc/sssd/sssd.conf</ind:filepath>
-    <ind:pattern operation="pattern match">^[\s]*\[sssd]([^\n\[\]]*\n+)+?[\s]*services.*pam.*$</ind:pattern>
+    <ind:filepath operation="pattern match">^/etc/sssd/(sssd|conf\.d/.*)\.conf$</ind:filepath>
+    <ind:pattern operation="pattern match">^[\s]*\[sssd]([^\n\[\]]*\n+)+?[\s]*services[\s]*=[\s]*(.*)$</ind:pattern>
     <ind:instance datatype="int">1</ind:instance>
```
yuumasato, line 24: I'll run a few tests, but it could be that this `instance` is the problem, it should be `greater than or equal 1`

ggbecker: I have tried with `greater than or equal 1` as well, but it didn't work as well. Maybe I'm doing something wrong.

linux_os/guide/services/sssd/sssd_enable_pam_services/oval/shared.xml
```diff
@@ -2,28 +2,29 @@
...
   id="test_sssd_enable_pam_services" version="1">
     <ind:object object_ref="obj_sssd_enable_pam_services" />
+    <ind:state state_ref="state_sssd_enable_pam_services" />
   </ind:textfilecontent54_test>
+
   <ind:textfilecontent54_object id="obj_sssd_enable_pam_services" version="1">
-    <ind:filepath>/etc/sssd/sssd.conf</ind:filepath>
-    <ind:pattern operation="pattern match">^[\s]*\[sssd]([^\n\[\]]*\n+)+?[\s]*services.*pam.*$</ind:pattern>
+    <ind:filepath operation="pattern match">^/etc/sssd/(sssd|conf\.d/.*)\.conf$</ind:filepath>
+    <ind:pattern operation="pattern match">^[\s]*\[sssd]([^\n\[\]]*\n+)+?[\s]*services[\s]*=[\s]*(.*)$</ind:pattern>
```
yuumasato, line 23: Another possible issue is that there are two capturing groups in the regex.

ggbecker: That might be it.

```
("subexpression" " " ) ("subexpression" "nss,pam" ) ) ) () ) ) [oscap(8378):unknown(7f7fecf0b880):seap-packet.c:903:SEAP_packet_recv]
```

ggbecker: I'm going to use https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/sssd/sssd_run_as_sssd_user/oval/shared.xml#L23 as a basis for the regex, it seems much more accurate. With one minor change to the capturing group it seems to work just fine.

mildas:
**Changes identified:**
Rule sssd_enable_pam_services:
 Node moved within OVAL check.
 Text changed in OVAL check.
 Node deleted from OVAL check.

 build_product ol7
 tests/test_suite.py rule --libvirt qemu:///system test-suite-vm --remediate-using bash --datastream build/ssg-ol7-ds.xml sssd_enable_pam_services

linux_os/guide/services/sssd/sssd_enable_pam_services/bash/shared.sh
```diff
@@ -3,17 +3,16 @@
...
-	echo "$SSSD_PAM_SERVICES" >> $SSSD_CONF
+for f in $( ls $SSSD_CONF $SSSD_CONF_DIR 2> /dev/null ) ; do
+	# finds all services entries under [sssd] configuration category, get a unique list so it doesn't add redundant fix
+	services_list=$( awk '/^\s*\[/{f=0} /^\s*\[sssd\]/{f=1}f' $f | grep -P '^services[ \t]*=' | uniq )
 
+	while IFS= read -r services; do
+		if [[ ! $services =~ "pam" ]]; then
+			sed -i "s/$services$/&, pam/" $f
+		fi
+	done <<< "$services_list"
```
yuumasato, line 17: What happens if there are two identical `services` lines?
I guess only one of the lines has `pam` appended.

ggbecker: both of the lines will have `pam` appended. That's my understanding of this `sed` command, every occurrence matching the `value` will have `pam` appended.

ggbecker: if these two lines are under the same `[sssd]` section, only the last line of `services` will be considered by the OVAL check, but if they are somehow part of different `[sssd]` sections, OVAL will test both of them.

yuumasato: @ggbecker You are correct, :)

yuumasato (approved):
Nice work.

PR #6022, opened 2020-08-26
Update OVAL check and remediations for sshd_use_priv_separation.

ggbecker:
#### Description:

- Add ansible remediation for `sshd_use_priv_separation` and remove template entry from its `rule.yml` as it is not needed.

~~- Per [V-72265](https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-72265?version=v2r7) `sshd_use_priv_separation` should accept both `sandbox` and `yes` values for `UsePrivilegeSeparation`:~~

~~`Uncomment the "UsePrivilegeSeparation" keyword in "/etc/ssh/sshd_config" (this file may be named differently or be in a different location if using a version of SSH that is provided by a third-party vendor) and set the value to "sandbox" or "yes":`~~

~~So the OVAL check has to use the regex `(sandbox|yes)` similar to what is done in [sshd_disable_compression](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/ssh/ssh_server/sshd_disable_compression/oval/shared.xml)~~

~~The variable `var_sshd_priv_separation` is still being used so remediations (including the new Ansible one) know how to remediate.~~

~~And last but not least, the default value for this parameter is `yes` so we have to switch `missing_parameter_pass` to `true`:~~

~~`UsePrivilegeSeparation
Specifies whether sshd(8) separates privileges by creating an unprivileged child process to deal with incoming network traffic. After successful authentication, another process will be created that has the privilege of the authenticated user. The goal of privilege separation is to prevent privilege escalation by containing any corruption within the unprivileged processes. The default is ''yes''. `~~
~~Reference: https://linux.die.net/man/5/sshd_config~~

#### Rationale:

- STIG Reference: https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-72265?version=v2r7

### DIFF
--- a/linux_os/guide/services/ssh/ssh_server/sshd_use_priv_separation/ansible/shared.yml
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_use_priv_separation/ansible/shared.yml
@@ -1,0 +1,8 @@
+# platform = multi_platform_all
+# reboot = false
+# strategy = restrict
+# complexity = low
+# disruption = low
+- (xccdf-var var_sshd_priv_separation)
+
+{{{ ansible_sshd_set(parameter="UsePrivilegeSeparation", value="{{ var_sshd_priv_separation }}") }}}
--- a/linux_os/guide/services/ssh/ssh_server/sshd_use_priv_separation/rule.yml
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_use_priv_separation/rule.yml
@@ -42,11 +42,3 @@ ocil: |-
     following command:
     <pre>$ sudo grep UsePrivilegeSeparation /etc/ssh/sshd_config</pre>
     If configured properly, output should be <tt><sub idref="var_sshd_priv_separation" /></tt>.
-
-template:
-    name: sshd_lineinfile
-    vars:
-        missing_parameter_pass: 'false'
-        parameter: UsePrivilegeSeparation
-        rule_id: sshd_use_priv_separation
-        value: sandbox


mildas:
**Changes identified:**
Rule sshd_use_priv_separation:
 Ansible remediation newly added.
 Templatization usage changed in linux_os/guide/services/ssh/ssh_server/sshd_use_priv_separation/oval/shared.xml.

 build_product ol7
 tests/test_suite.py rule --libvirt qemu:///system test-suite-vm --remediate-using ansible --datastream build/ssg-ol7-ds.xml sshd_use_priv_separation
 tests/test_suite.py rule --libvirt qemu:///system test-suite-vm --remediate-using bash --datastream build/ssg-ol7-ds.xml sshd_use_priv_separation

linux_os/guide/services/ssh/ssh_server/sshd_use_priv_separation/oval/shared.xml
```diff
@@ -1,36 +1 @@
...
-  {{{ oval_line_in_file_test(path='/etc/ssh/sshd_config', parameter='UsePrivilegeSeparation') }}}
-  {{{ oval_line_in_file_object(path='/etc/ssh/sshd_config', prefix_regex="^[ \\t]*(?i)", parameter='UsePrivilegeSeparation', separator_regex='(?-i)[ \\t]+') }}}
-  
-  <ind:textfilecontent54_state id="state_sshd_use_priv_separation" version="1">
-    <ind:subexpression operation="equals" var_ref="var_sshd_priv_separation" />
-  </ind:textfilecontent54_state>
-
-  <external_variable comment="external variable for the desired value of UsePrivilegeSeparation" datatype="string" id="var_sshd_priv_separation" version="1" />
-</def-group>
+{{{ oval_sshd_config(parameter="UsePrivilegeSeparation", value="(sandbox|yes)", missing_parameter_pass=true) }}}
```
redhatrises, line 1: This isn't going to work. If `var_sshd_priv_separation` is set to `sandbox` and `sshd_config` isn't configured, pretty sure the oval will pass which means that the pass is really a false positive when `sandbox` is used.

ggbecker: From the STIG, both `sandbox` and `yes` values should be accepted and `yes` is the default value of `UsePrivilegeSeparation`, so that's why `missing_parameter_pass=true`

[V-72265](
https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-72265?version=v2r7) 
`Uncomment the "UsePrivilegeSeparation" keyword in "/etc/ssh/sshd_config" (this file may be named differently or be in a different location if using a version of SSH that is provided by a third-party vendor) and set the value to "sandbox" or "yes":`

[sshd_config](https://linux.die.net/man/5/sshd_config)
`UsePrivilegeSeparation`
`Specifies whether sshd(8) separates privileges by creating an unprivileged child process to deal with incoming network traffic. After successful authentication, another process will be created that has the privilege of the authenticated user. The goal of privilege separation is to prevent privilege escalation by containing any corruption within the unprivileged processes. The default is ''yes''. `

jan-cerny:
@openscap-ci test this please

mildas:
**Changes identified:**
Rule sshd_use_priv_separation:
 Ansible remediation newly added.
 Templatization usage changed in linux_os/guide/services/ssh/ssh_server/sshd_use_priv_separation/oval/shared.xml.

 build_product ol8
 tests/test_suite.py rule --libvirt qemu:///system test-suite-vm --remediate-using ansible --datastream build/ssg-ol8-ds.xml sshd_use_priv_separation
 tests/test_suite.py rule --libvirt qemu:///system test-suite-vm --remediate-using bash --datastream build/ssg-ol8-ds.xml sshd_use_priv_separation

linux_os/guide/services/ssh/ssh_server/sshd_use_priv_separation/oval/shared.xml
ggbecker: One thing I noticed is that if this change is correct, we need to update the rule text to explicitly mention that both options are accepted.

redhatrises: Doesn't matter. If I purposefully choose `sandbox` which affords more protections around child process sandboxing, and `UsePrivilegeSeparation` is not configured, it is a false positive because it's default is `yes` and less secure which means my scan and scan results are wrong.

ggbecker: Ok agreed, so I'll keep both check and remediation to expect for the variable `var_sshd_priv_separation`.

mildas:
**Changes identified:**
Rule sshd_use_priv_separation:
 Ansible remediation newly added.

 build_product ol8
 tests/test_suite.py rule --libvirt qemu:///system test-suite-vm --remediate-using ansible --datastream build/ssg-ol8-ds.xml sshd_use_priv_separation

PR #6026, opened 2020-08-28
Fix ocp4 e2e tests

JAORMX:
We needed to change our deployment strategy and upgrade the operator...


jhrozek:
During a local run, I saw:
```
    --- PASS: TestE2e/Parameter_setup_and_validation (0.01s)
    --- PASS: TestE2e/Operator_setup (78.14s)
    --- PASS: TestE2e/Run_first_compliance_scan (456.44s)
    --- PASS: TestE2e/Wait_for_Remediations_to_apply (30.07s)
    --- FAIL: TestE2e/Run_second_compliance_scan (581.19s)
```
And the nodes were still updating which seems like we still need to tune the `isNodeReady` function. I'm going to look into that..

JAORMX:
Meeting this. Seems we have an issue waiting for the nodes... But that's unrelated. This gets us further

PR #6033, opened 2020-09-01
reword ocils of password rules wrt existing accounts

vojtapolasek:
#### Description:

- reword ocils to use awk instead of chage

#### Rationale:

- it is aligned with STIG for rhel7 and also it is prone to problems caused by different than english locale

### DIFF
--- a/linux_os/guide/system/accounts/accounts-restrictions/password_expiration/accounts_password_set_max_life_existing/rule.yml
+++ b/linux_os/guide/system/accounts/accounts-restrictions/password_expiration/accounts_password_set_max_life_existing/rule.yml
@@ -34,7 +34,6 @@ ocil_clause: 'existing passwords are not configured correctly'
 
 ocil: |-
     Check whether the maximum time period for existing passwords is restricted
-    to 60 days by running the following command for each user:
-    <pre>$ sudo chage -l <i>USER</i> | grep Maximum</pre>
-    The output for each user should return something similary to the following:
-    <pre>Maximum number of days between password change\t\t: 60</pre>
+    to 60 days:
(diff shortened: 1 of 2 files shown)


linux_os/guide/system/accounts/accounts-restrictions/password_expiration/accounts_password_set_max_life_existing/rule.yml
```diff
@@ -34,7 +34,6 @@ ocil_clause: 'existing passwords are not configured correctly'
 
 ocil: |-
     Check whether the maximum time period for existing passwords is restricted
-    to 60 days by running the following command for each user:
-    <pre>$ sudo chage -l <i>USER</i> | grep Maximum</pre>
-    The output for each user should return something similary to the following:
-    <pre>Maximum number of days between password change\t\t: 60</pre>
+    to 60 days:
+    <pre>sudo awk -F: '$5 &gt; 60 {print $1 " " $5}' /etc/shadow</pre>
+    The output must not contain names of any system account.
```
redhatrises, line 39: This output shows that service accounts which contain both `/sbin/nologin` (i.e. not an interactive account) and `!!` (i.e. locked accounts) are shown in the output. Checking for a 60 day password on a non-interactive, locked service account makes no sense as it doesn't change anything in terms of security in addition to the fact that there is no password or a need for a password to begin with.

vojtapolasek:
OK makes sense. Do you think that the previous ocil was OK?

redhatrises:
> OK makes sense. Do you think that the previous ocil was OK?

For now. I think that a better solution like an enhancement to chage would be a better way to go. Not to be handled here of course.

PR #6035, opened 2020-09-02
add CUI kickstart for rhel8

vojtapolasek:


matejak:
Merging - the kickstart is based on the OSPP one, whith only the profile ID being different. That's the expected behavior, as the CUI profile is identical to OSPP from the kickstart perspective.

PR #6037, opened 2020-09-02
Fix CPE check that detects if package manager is installed.

ggbecker:
#### Description:

- Fix CPE check that detects if package manager is installed. In fedora, the returned valid for this rpmtest is `dnf`, so we can use the product.yml variable `pkg_manager` to detect correctly.

#### Rationale:

- Checks that have `platform: yum` in fedora are returning `notapplicable`.

`oscap xccdf eval --profile "(all)" --rule xccdf_org.ssgproject.content_rule_ensure_gpgcheck_local_packages build/ssg-fedora-ds.xml`

```
Title   Ensure gpgcheck Enabled for Local Packages
Rule    xccdf_org.ssgproject.content_rule_ensure_gpgcheck_local_packages
Result  notapplicable
```

With the changes it returns different then `notapplicable`:

```
Title   Ensure gpgcheck Enabled for Local Packages
Rule    xccdf_org.ssgproject.content_rule_ensure_gpgcheck_local_packages
Result  fail
```

### DIFF
--- a/shared/checks/oval/installed_env_has_yum_package.xml
+++ b/shared/checks/oval/installed_env_has_yum_package.xml
@@ -21,7 +21,7 @@
     <linux:object object_ref="obj_env_has_yum_installed" />
   </linux:rpminfo_test>
   <linux:rpminfo_object id="obj_env_has_yum_installed" version="1">
-    <linux:name>yum</linux:name>
+    <linux:name>{{{ pkg_manager }}}</linux:name>
   </linux:rpminfo_object>
 {{% elif pkg_system == "dpkg" %}}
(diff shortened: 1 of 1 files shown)


mildas:
**Changes identified:**
Rule checks:
 Node deleted from OVAL check.
 Text changed in OVAL check.

redhatrises:
This OVAL check is for yum. Yum != DNF

ggbecker:
> This OVAL check is for yum. Yum != DNF

I think I've misunderstood the purpose of this cpe a little bit. Closing this PR. Thanks.

PR #6047, opened 2020-09-04
Introduce platform to package mapping

### LINKED ISSUE #6041: Ansible remediations are not applied when using platform and the platform name is not the package name
#### Description of problem:

I stumbled upon this while writing tests for the rule accounts_logon_fail_delay. See this draft PR:
https://github.com/ComplianceAsCode/content/pull/6040
As you can see, the rule is applicable when platform "login_defs" is available. This platform uses a check which checks for shadow-utils package on systems using RPM. Unfortunately the Ansible remediation is created in a way so that it looks for the "login_defs" package, which does not exist.

master

rhel 7
#### Steps to Reproduce:

1. Build rhel7 product from the PR linked
2. run upstream testsuite
python3 test_suite.py rule --libvirt ... --datastream ../build/ssg-rhel7-ds.xml --remediate-using ansible --debug accounts_logon_fail_delay

#### Actual Results:
You will see that in case the line is missing or is wrong the new line is not added. Inspect test suite output and you will see that the task was skipped.

#### Expected Results:
The remediation is applied correctly.

yuumasato:
#### Description:

Introduce a mapping of CPE package platform name to a package name.

Each linux distro or version may have its specific name for a package,
this mapping allows a product to override the package name of a
platform.

In `ssg/constants.py` a default `package` name is mapped when it is different from the `platform` name.
And if the default `package` name still needs to be tweaked, it can be done in the `product.yml`.

The package name override mapping can also be extended to the actual OVALs that check for the package installed.
But I'd leave this for another PR.

#### Rationale:

- Allows larger flexibility of package names. Debian and RHEL often have different package names for example..

- Fixes #6041

### DIFF
--- a/rhcos4/product.yml
+++ b/rhcos4/product.yml
@@ -9,3 +9,7 @@ profiles_root: "./profiles"
 pkg_system: "rpm"
 
 init_system: "systemd"
+
+# Mapping of CPE platform to package
+platform_package_overrides:
+  login_defs: "shadow-utils"
--- a/rhel6/product.yml
+++ b/rhel6/product.yml
@@ -20,3 +20,7 @@ aux_pkg_version: "2fa658e0"
 
 release_key_fingerprint: "567E347AD0044ADE55BA8A5F199E2F91FD431D51"
 auxiliary_key_fingerprint: "43A6E49C4A38F4BE9ABF2A5345689C882FA658E0"
+
+# Mapping of CPE platform to package
+platform_package_overrides:
+  login_defs: "shadow-utils"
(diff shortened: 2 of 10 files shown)


mildas:
**Changes identified:**
Others:

mildas:
**Changes identified:**
Others:

vojtapolasek:
Looks great as a temporary solution. I think we should later propose something more universla, which would cover OVAL, Bash and Ansible. Meanwhile, could you please also add something similar for rhel7?

mildas:
**Changes identified:**
Others:
 Python abstract syntax tree change found in ssg/yaml.py.

vojtapolasek (approved):
LGTM, thank you very much.

yuumasato:
Thanks for the review!

PR #6054, opened 2020-09-07
tests/ocp4e2e: Wait for MCP to update

jhrozek:
#### Description:

This should prevent running the second suite while the first is still
updating.

#### Rationale:

We were looking only at node status but what can happen is that all the nodes
report ready for a bit before the next one even has a chance to start updating.


JAORMX:
YES!!!

PR #6060, opened 2020-09-08
Fix bash remediation for rule accounts_no_uid_except_zero.

ggbecker:
#### Help needed
~~- This remediation doesn't fix the problem because it just `soft` locks offending accounts, it doesn't change the UID of them as suggested by the fix text. Do you have any suggestions on how to improve this remediation?~~

#### Description:
 - Fix bash remediation for rule accounts_no_uid_except_zero.
   - When multiple offending accounts were found in the system, xargs
couldn't process properly each account found.

#### Rationale:

- Reference: https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-72005?version=v2r7

### DIFF
--- a/linux_os/guide/system/accounts/accounts-restrictions/root_logins/accounts_no_uid_except_zero/bash/rhel6.sh
+++ b/linux_os/guide/system/accounts/accounts-restrictions/root_logins/accounts_no_uid_except_zero/bash/rhel6.sh
@@ -1,2 +1,0 @@
-# platform = Red Hat Enterprise Linux 6
-awk -F: '$3 == 0 && $1 != "root" { print $1 }' /etc/passwd | xargs passwd -l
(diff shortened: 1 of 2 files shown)


mildas:
**Changes identified:**
Rule rhel6:
Rule accounts_no_uid_except_zero:

 build_product ol8
 tests/test_suite.py rule --libvirt qemu:///system test-suite-vm --remediate-using bash --datastream build/ssg-ol8-ds.xml accounts_no_uid_except_zero

jan-cerny:
Perhaps  find the smallest unused UID and then change the UID using usermod -u?

redhatrises:
It should lock the account, and then system admins can fix the problems that are associated with changing user ids. That's why the user ID is never changed.

carlosmmatos:
@redhatrises Is it known to the user that the account will be locked? This seems like one of those "manual" remediations as we wouldn't want to arbitrarily assign a new uid to an account. I saw this a lot when people where running Oracle DB2 products - it literally said in the manual to assign uid 0 to db2 accounts 🤦🏻‍♂️

redhatrises:
> @redhatrises Is it known to the user that the account will be locked? This seems like one of those "manual" remediations as we wouldn't want to arbitrarily assign a new uid to an account. I saw this a lot when people where running Oracle DB2 products - it literally said in the manual to assign uid 0 to db2 accounts 🤦🏻‍♂️

Yeah... this isn't a user notification thing. This is on the security department. If a user account is locked, the security department must investigate.

ggbecker:
Ok, if we don't want to remediate the UIDs, this PR is ready for review.

yuumasato:
> It should lock the account, and then system admins can fix the problems that are associated with changing user ids. That's why the user ID is never changed.

How about an XCCDF warning about the remediation behavior and expected admin intervention?

PR #6061, opened 2020-09-08
Introduce machine and package platform conditionals to Bash remediations

yuumasato:
#### Description:

- Trickle down the CPE Platform to the Bash remediations
  - When a rule has a `platform` be it `machine` or a `Package CPE` platform, the whole Bash remediation is wrapped around an `if` condition
  - Also adds conditionals for `machine` platform

#### Rationale:
When a rule has a Package CPE platform, like `platform: dconf` for example.
It is meant to be evaluated and remediated when the  systems has `dconf` package installed.
Evaluation and remediation done with OpenSCAP respects these CPEs, as the scanner understands the `CPE` and the `XCCDF:Platform` field.
But when the Bash scripts are generated, the Package applicability of the remediation is lost.

#### Rationale:

- Analogous to #6025

### DIFF
--- a/ssg/build_remediations.py
+++ b/ssg/build_remediations.py
@@ -736,14 +736,16 @@ def expand_xccdf_subs(fix, remediation_type, remediation_functions):
         patcomp = re.compile(pattern, re.DOTALL)
         fixparts = re.split(patcomp, fix.text)
         if fixparts[0] is not None:
-            # Split the portion of fix.text from fix start to first call of
-            # remediation function, keeping only the third part:
-            # * tail        to hold part of the fix.text after inclusion,
-            #               but before first call of remediation function
+            # Split the portion of fix.text at the string remediation_functions,
+            # and remove preceeding comment whenever it is there.
+            # * head        holds part of the fix.text before
+            #               remediation_functions string
+            # * tail        holds part of the fix.text after the
+            #               remediation_functions string
             try:
-                rfpattern = '(.*remediation_functions)(.*)'
-                rfpatcomp = re.compile(rfpattern, re.DOTALL)
-                _, _, tail, _ = re.split(rfpatcomp, fixparts[0], maxsplit=2)
+                rfpattern = r'((?:# Include source function library\.\n)?.*remediation_functions)'
+                rfpatcomp = re.compile(rfpattern)
+                head, _, tail = re.split(rfpatcomp, fixparts[0], maxsplit=1)
             except ValueError:
                 sys.stderr.write("Processing fix.text for: %s rule\n"
                                  % fix.get('rule'))
@@ -751,9 +753,10 @@ def expand_xccdf_subs(fix, remediation_type, remediation_functions):
                                  "after inclusion of remediation functions."
                                  " Aborting..\n")
                 sys.exit(1)
-            # If the 'tail' is not empty, make it new fix.text.
+            # If the 'head' is not empty, make it new fix.text.
             # Otherwise use ''
-            fix.text = tail if tail is not None else ''
+            fix.text = head if head is not None else ''
+            fix.text += tail if tail is not None else ''
             # Drop the first element of 'fixparts' since it has been processed
             fixparts.pop(0)
             # Perform sanity check on new 'fixparts' list content (to continue
--- a/ssg/build_remediations.py
+++ b/ssg/build_remediations.py
@@ -273,9 +273,11 @@ class BashRemediation(Remediation):
         self.local_env_yaml.update(env_yaml)
         result = super(BashRemediation, self).parse_from_file_with_jinja(self.local_env_yaml)
 
-        # There can be repeated inherited platforms and rule platforms
-        rule_platforms = set(self.associated_rule.inherited_platforms)
-        rule_platforms.add(self.associated_rule.platform)
+        rule_platforms = set()
+        if self.associated_rule:
+            # There can be repeated inherited platforms and rule platforms
+            rule_platforms.update(self.associated_rule.inherited_platforms)
+            rule_platforms.add(self.associated_rule.platform)
 
         platform_conditionals = []
         for platform in rule_platforms:
--- a/ssg/build_remediations.py
+++ b/ssg/build_remediations.py
@@ -27,6 +27,13 @@ REMEDIATION_TO_EXT_MAP = {
     'kubernetes': '.yml'
 }
 
+PKG_MANAGER_TO_PACKAGE_CHECK_COMMAND = {
+    'apt_get': 'dpkg-query -s {} &>/dev/null',
+    'dnf': 'rpm --quiet -q {}',
+    'yum': 'rpm --quiet -q {}',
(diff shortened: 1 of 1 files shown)


mildas:
**Changes identified:**
Others:

mildas:
**Changes identified:**
Others:

ssg/build_remediations.py
```diff
@@ -262,6 +269,46 @@ class BashRemediation(Remediation):
...
+                if platform in self.local_env_yaml["platform_package_overrides"]:
+                    platform = self.local_env_yaml["platform_package_overrides"].get(platform)
+
+                # Adjust package check command according to the pkg_manager
+                pkg_manager = self.local_env_yaml["pkg_manager"]
+                pkg_check_command = PKG_MANAGER_TO_PACKAGE_CHECK_COMMAND[pkg_manager]
+                platform_conditionals.append(pkg_check_command.format(platform))
+
+        if platform_conditionals:
+            platform_fix_text = "# Remediation is applicable only in certain platforms\n"
```
matejak, line 300: It is advisable to construct a multi-line string by creating a list of lines and then joining it. The primary aspect are performance reasons, but I believe that it would be nicer as well.
Next, I would suggest placing a blank line between the header and the remediation body for greater readability.

ssg/build_remediations.py
```diff
@@ -262,6 +269,46 @@ class BashRemediation(Remediation):
...
+
+            cond = platform_conditionals.pop(0)
+            platform_fix_text += "if {}".format(cond)
+            for cond in platform_conditionals:
+                platform_fix_text += " && {}".format(cond)
+            platform_fix_text += '; then\n{}\nelse\necho "Remediation is not applicable, nothing was done"\nfi'.format(result.contents)
```
matejak, line 306: 1. I would use `textwrap. indent` to indent the innner original body of the remediation if it doesn't break our dark remediation generation magic.
2. What do you think of redirecting the echo to stderr?

ssg/build_remediations.py
```diff
@@ -262,6 +269,46 @@ class BashRemediation(Remediation):
...
+                # Adjust package check command according to the pkg_manager
+                pkg_manager = self.local_env_yaml["pkg_manager"]
+                pkg_check_command = PKG_MANAGER_TO_PACKAGE_CHECK_COMMAND[pkg_manager]
+                platform_conditionals.append(pkg_check_command.format(platform))
+
+        if platform_conditionals:
+            platform_fix_text = "# Remediation is applicable only in certain platforms\n"
+
+            cond = platform_conditionals.pop(0)
+            platform_fix_text += "if {}".format(cond)
```
matejak, line 303: I think that it could be slightly more elegant to join the `platform_conditionals` list by ` && `.

ssg/build_remediations.py
```diff
@@ -691,24 +738,27 @@ def expand_xccdf_subs(fix, remediation_type, remediation_functions):
         patcomp = re.compile(pattern, re.DOTALL)
         fixparts = re.split(patcomp, fix.text)
         if fixparts[0] is not None:
```
matejak, line 740: I can't wait for the moment when we take some of the existing magic out and simplify this function.

matejak (changes requested):
Well done, there are some minor remarks, but the functionality works as expected.

The `{}` in format strings can be probably replaced by `{0}` to keep the code working on Python<2.7.
I see potential problems with tests, as I can imagine that sometest scenarios may misbehave due to the newly introduced "do nothing when it's not my business" functionality, but I see this as unlikely - OVALs are already platform-aware.

ssg/build_remediations.py
yuumasato: I suspect that when all bash remediation functions are converted to Jinja macros `expand_xccdf_subs` won't be needed anymore.

mildas:
**Changes identified:**
Others:

ssg/build_remediations.py
yuumasato: Indenting the body of the the remediation didn't break the build.
Unfortunately, not everything is indented.
When the scanner performs the XCCDF substitutions, the Bash remediation functions are not properly indented.
Main issue being that they have multiple lines, and can be used "platform wrapped" or not.

yuumasato: @matejak `indent` is not available in python2.
Do you know of a trivial substitute?

matejak: Oh, that Python2 not being able to do this - that's a surprise. We can have an indent as an optional feature in the try-except block, but I wouldn't do it now, as it conflicts with XCCDF substitutions, and there is a small likelihood that some remediations may not be indent-safe - consider [here-docs](https://unix.stackexchange.com/questions/76481/cant-indent-heredoc-to-match-code-blocks-indentation).

matejak:
> Can you show us an example of the conditional?

That's easy - compile e.g. Fedora and then grep `build/fixes/bash` with `&&`, and you will see all sorts of platforms.

Regarding the PR, I gave a second thought on the indentation, and I suggest to put there a comment that there is such possibility, but there is a Python2/3 catch, plus a small risk for remediations themselves.

matejak:
It may be that if you rebase against master, OCP tests will start to pass :-)

ssg/build_remediations.py
yuumasato: Interesting find. I'll drop adcb827 for now.

mildas:
**Changes identified:**
Others:

mildas:
**Changes identified:**
Others:

mildas:
**Changes identified:**
Others:

ssg/build_remediations.py
```diff
@@ -262,6 +269,56 @@ class BashRemediation(Remediation):
...
+            if not result.contents.startswith("\n"):
+                wrapped_fix_text.append("")
+
+            # It is possible to indent the original body of the remediation with textwrap.indent(),
+            # however, it is not supported by python2, and there is a risk of breaking remediations
+            # For example, remediations with a here-doc block could be affected.
+            wrapped_fix_text.append("{0}".format(result.contents))
+            wrapped_fix_text.append("")
+            wrapped_fix_text.append("else")
+            wrapped_fix_text.append("    >&2 echo 'Remediation is not applicable, nothing was done'")
```
matejak, line 315: Is this leading `>&2` intentional? I would place it after the echo.

yuumasato: Yes, it follows same syntax similar to what OpenSCAP does when generating Bash fixes.
https://github.com/OpenSCAP/openscap/blob/maint-1.3/src/XCCDF_POLICY/xccdf_policy_remediate.c#L631

matejak: Apparently you can place redirects before of after "simple commands" in Bash. Interesting!

matejak (approved):

matejak:
This is a great step and historic moment for the project - our most common remediations are now platform-aware, and therefore consistent with OVALs and Ansible, who became fully platform-aware recently.
Great work!

yuumasato:
@matejak Thank you very much for the review!

PR #6067, opened 2020-09-09
Set a lower bound value for accounts_passwords_pam_faillock_deny check.

ggbecker:
#### Description:

- Set a lower bound value (greater than zero) for `accounts_passwords_pam_faillock_deny` check.

#### Rationale:
- `pam` option `deny` shouldn't be set to zero.

- Reference: https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-71943?version=V2R8

`If the "deny" parameter is set to "0" or a value greater than "3" on both "auth" lines with the "pam_faillock.so" module, or is missing from these lines, this is a finding.`

### DIFF
--- a/linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_passwords_pam_faillock_deny/oval/shared.xml
+++ b/linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_passwords_pam_faillock_deny/oval/shared.xml
@@ -45,9 +45,10 @@
   <!-- step 1 and 3 test -->
   <ind:textfilecontent54_test id="test_accounts_passwords_pam_faillock_numeric_default_check_system-auth"
   check="all" check_existence="all_exist"
-  comment="Checks if pam_faillock authfail is hit even if pam_unix skips lines by defaulting, and also authfail deny value" version="1">
+  comment="Checks if pam_faillock authfail is hit even if pam_unix skips lines by defaulting, and also authfail deny value" version="2">
     <ind:object object_ref="object_accounts_passwords_pam_faillock_when_lines_skipped_system-auth" />
-    <ind:state state_ref="state_var_accounts_passwords_pam_faillock_deny_value" />
+    <ind:state state_ref="state_var_accounts_passwords_pam_faillock_deny_value_upper_bound" />
+    <ind:state state_ref="state_var_accounts_passwords_pam_faillock_deny_value_lower_bound" />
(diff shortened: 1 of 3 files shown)


mildas:
**Changes identified:**
Rule accounts_passwords_pam_faillock_deny:
 Text changed in OVAL check.
 Node moved within OVAL check.

 build_product ol8
 tests/test_suite.py rule --libvirt qemu:///system test-suite-vm --remediate-using bash --datastream build/ssg-ol8-ds.xml accounts_passwords_pam_faillock_deny

linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_passwords_pam_faillock_deny/oval/shared.xml
```diff
@@ -113,17 +115,22 @@
   <external_variable id="var_accounts_passwords_pam_faillock_deny" datatype="int"
   comment="number of failed login attempts allowed" version="1" />
 
-  <ind:textfilecontent54_state id="state_var_accounts_passwords_pam_faillock_deny_value" version="1">
+  <ind:textfilecontent54_state id="state_var_accounts_passwords_pam_faillock_deny_value_upper_bound" version="1">
     <ind:subexpression datatype="int" operation="less than or equal" var_ref="var_accounts_passwords_pam_faillock_deny" />
   </ind:textfilecontent54_state>
 
+  <ind:textfilecontent54_state id="state_var_accounts_passwords_pam_faillock_deny_value_lower_bound" version="1">
+    <ind:subexpression datatype="int" operation="greater than">0</ind:subexpression>
```
jan-cerny, line 123: The rule description in rule.yml doesn't say anything about allowing other values than value of var_accounts_passwords_pam_faillock_deny.

jan-cerny (commented):
The code looks great to me. The test scenario passes, thanks for adding it. But I feel it's not in line with rule.yml and I'm wondering how it affects other profiles that use this rule.

linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_passwords_pam_faillock_deny/oval/shared.xml
redhatrises: I don't see a problem with it at all, and it is quite in line with rule.yml. In fact, this fixes the bug where you could disable deny.

PR #6070, opened 2020-09-11
Improve the description of ocp_allowed_registries tests

jhrozek:
It was unclear how the admin is supposed to remediate the issue
without poking at the warning or the tests.

### DIFF
--- a/applications/openshift/registry/ocp_allowed_registries/rule.yml
+++ b/applications/openshift/registry/ocp_allowed_registries/rule.yml
@@ -8,6 +8,17 @@ description: |-
     and pods. This configuration setting ensures that all registries other than
     those specified are blocked.
 
+    You can set the allowed repositories by applying the following manifest using
+    <pre>oc patch</pre>, e.g. if you save the following snippet to
+    <pre>/tmp/allowed-registries-patch.yaml</pre>
+    <pre>
+    spec:
+      registrySources:
+        allowedRegistries:
+        - my-trusted-registry.internal.example.com
+    </pre> you would call
+    <pre>oc patch image.config.openshift.io cluster --patch="$(cat /tmp/allowed-registries-patch.yaml)" --type=merge</pre>
+
 rationale: |-
     Allowed registries should be configured to restrict the registries that the
     OpenShift container runtime can access, and all other registries should be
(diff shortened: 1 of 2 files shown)


jhrozek:
CC @mrogers950

JAORMX (approved):

applications/openshift/registry/ocp_allowed_registries/rule.yml
```diff
@@ -8,6 +8,15 @@ description: |-
...
     those specified are blocked.
 
+    You can set the allowed repositories by applying the following manifest using
+    <code>oc patch</code>:
+    <code>
+    spec:
+      registrySources:
+        allowedRegistries:
+        - my-trusted-registry.internal.example.com
+    </code>
```
redhatrises, line 18: I believe that these tags need to be `<pre/>` instead of `<code/>`. I have seen `<code/>` not render multiple lines of code correctly.

applications/openshift/registry/ocp_allowed_registries_for_import/rule.yml
```diff
@@ -7,6 +7,14 @@ description: |-
...
     to control, as a user who can stand up a malicious registry can then import
     content which claims to include the SHAs of legimitate content layers.
+    You can set the allowed repositories for import by applying the following
+    manifest using <code>oc patch</code>:
+    <code>
+    spec:
+      allowedRegistriesForImport:
+      - domainName: my-trusted-registry.internal.example.com
+        insecure: false
+    </code>
```
redhatrises, line 17: I believe that these tags need to be `<pre/>` instead of `<code/>`. I have seen `<code/>` not render multiple lines of code correctly.

applications/openshift/registry/ocp_allowed_registries/rule.yml
```diff
@@ -8,6 +8,15 @@ description: |-
     and pods. This configuration setting ensures that all registries other than
     those specified are blocked.
 
+    You can set the allowed repositories by applying the following manifest using
+    <code>oc patch</code>:
```
mrogers950, line 12: We might want to include the patch command itself just for ease-of-use

applications/openshift/registry/ocp_allowed_registries/rule.yml
jhrozek: Pre is now used

applications/openshift/registry/ocp_allowed_registries_for_import/rule.yml
jhrozek: Pre is now used

applications/openshift/registry/ocp_allowed_registries/rule.yml
jhrozek: Done, does it look any better now?

PR #6075, opened 2020-09-14
Migrate MachineConfigs to use ignition 3.1.0

JAORMX:
OCP 4.6 now uses ignition 3.1 instead of 2.2. So migration is needed.


JAORMX:
This won't work out of the box. We need to upgrade our MCO version in the compliance-operator. Doing so here: https://github.com/openshift/compliance-operator/pull/441 This will need a release there.

PR #6078, opened 2020-09-15
OCP4: Turn off documentation complete for CIS profile

JAORMX:
This profile is not actually complete nor usable. So let's only enable
it when it's ready.


JAORMX:
@jhrozek @mrogers950 PTAL

jhrozek (approved):
Verified that the cis profile is no longer included in the DS
/lgtm

PR #6080, opened 2020-09-15
Use resolved profiles in rule playbooks

jan-cerny:
#### Description:
Use resolved profiles in rule playbooks generator.

#### Rationale:
The profiles are resolved in previous build steps and saved to build/product/profiles. We don't need to resolve the profiles again when generating the playbooks. We can use the resolved profiles instead.

### DIFF
--- a/build-scripts/build_rule_playbooks.py
+++ b/build-scripts/build_rule_playbooks.py
@@ -29,6 +29,13 @@ def parse_args():
         "--ssg-root."
     )
     p.add_argument(
+        "--resolved-profiles-dir",
+        help="Directory that contains preprocessed profiles in YAML format "
+        "eg. ~/scap-security-guide/build/fedora/profiles. "
+        "If --resolved-profiles-dir is not specified, it is derived from "
+        "--ssg-root."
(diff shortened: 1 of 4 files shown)


mildas:
**Changes identified:**
Others:
 Python abstract syntax tree change found in build-scripts/build_rule_playbooks.py.
 Python abstract syntax tree change found in ssg/playbook_builder.py.
 Python abstract syntax tree change found in tests/unit/ssg-module/test_playbook_builder.py.

matejak:
This PR extends the `build_rule_playbooks`, so it now optionally accepts a path to compiled profiles, which it uses as an input for Ansible Playbooks.
Before this PR, profiles from the product directory were used, and they had to be resolved in the code of the script. As the script now reuses already-resolved profiles, it shortens the build time and makes the build process less convoluted.

In case when the script is invoked in the old way, it picks up profiles from the product directory, but doesn't attempt to resolve them.

matejak:
Merging - the change of behavior is inseparable from deduplication of build steps, and the build process as such stays the same.

PR #6088, opened 2020-09-15
Change OCP4 (RHCOS) audit=1 kernel option rule to check only the latest entry

evgenyz:
Change the audit=1 kernel option rule for RHCOS to be able to pass after remediation.

The `coreos_kernel_option` template only checks if the latest boot entry is compliant (acc. to https://github.com/ComplianceAsCode/content/issues/5285#issuecomment-624676603).

@JAORMX I ended up with a different template for CoreOS. Once the problem of stalled boot entry will be solved this rule (and other kernel option related rules) could be switched to `bls_entries_option` template (which would check all boot entries).


mildas:
**Changes identified:**
Profile moderate on rhcos4:
 Rule bls_audit_option removed from moderate profile.
 Rule coreos_audit_option added to moderate profile.
Others:

 build_product rhcos4
 tests/test_suite.py profile --libvirt qemu:///system test-suite-vm --datastream build/ssg-rhcos4-ds.xml moderate

redhatrises:
What's the follow up to fix BLS for all entries? Not having all entries configured is usually an audit finding.

JAORMX:
> What's the follow up to fix BLS for all entries? Not having all entries configured is usually an audit finding.

Hadn't we mentioned agreed that this would the solution while we figured out a way to fix this in coreos? Let me add a task for the team to follow up on this.

JAORMX:
/test e2e-aws-rhcos4-moderate
cluster bootstrapping error

shared/templates/template_OVAL_coreos_kernel_option
```diff
@@ -0,0 +1,71 @@
...
+  version="1">
+    <unix:object object_ref="object_coreos_{{{ SANITIZED_ARG_NAME }}}_entry_2_does_not_exists" />
+  </unix:file_test>
+
+  <unix:file_object id="object_coreos_{{{ SANITIZED_ARG_NAME }}}_entry_2_does_not_exists"
+  version="1">
+    <unix:filepath operation="pattern match">^/boot/loader/entries/ostree-2-*\.conf</unix:filepath>
+  </unix:file_object>
+
+</def-group>
```
JAORMX, line 71: For completeness; what about checking /proc/cmdline as well? That would ensure that the current configuration is compliant.

evgenyz: I'm still not sure if it should be a part of this check. Other bootloader-related rules don't have the runtime counterpart (because runtime check might be different even when bootloaders are the same). While this is not really applicable to this particular hackish rule, it would definitely be a problem for generic BLS-compatible check `bls_entries_option` in the future.

evgenyz: I would suggest creating a `proc_cmdline_option` template and a new rule based on this. It might be more preferable to have composite rules that could combine multiple templates into a single runtime+configuration check, but there is no way to make such a thing ATM, AFAIK.

JAORMX: Right; however, this is already a coreos specificy check. So maybe in this case it would make sense to add the proc/cmdline check here.

jhrozek: btw contrary to applying remediations on RHEL, applying a MachineConfig remediation through MachineConfigOperator always reboots the nodes, so when you apply this remediation, the machines in the cluster would automatically reboot into the approved config and at that point the `/proc/cmdline` check should pass as well. There is no extra disruption for the admin.

yuumasato (approved):
LGTM.

CoreOS replaces the first boot entry with the second boot entry when updating right? So that there are  only 2 boot entries.
I would a comment somewhere in the template mentioning that the check relies on this behavior.

shared/templates/template_OVAL_coreos_kernel_option
```diff
@@ -0,0 +1,71 @@
...
+        <criteria operator="AND">
+            <criterion test_ref="test_coreos_{{{ SANITIZED_ARG_NAME }}}_entry_2_options"
+            comment="Check if {{{ ARG_NAME_VALUE }}} is present in the 'options' line in /boot/loader/entries/ostree-2-*.conf (if it does exists)" />
+        </criteria>
+        <criteria operator="AND">
+            <criterion test_ref="test_coreos_{{{ SANITIZED_ARG_NAME }}}_entry_1_options"
+            comment="Check if {{{ ARG_NAME_VALUE }}} is present in the 'options' line in /boot/loader/entries/ostree-1-*.conf" />
+            <criterion test_ref="test_coreos_{{{ SANITIZED_ARG_NAME }}}_entry_2_does_not_exists"
+            comment="Check if /boot/loader/entries/ostree-2-*.conf is not present" />
+        </criteria>
```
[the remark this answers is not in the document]
evgenyz: I continued the work on the rules in #6100, let's move there.

evgenyz:
I think we can merge it, follow-up improvements are in #6100.

shared/templates/template_OVAL_coreos_kernel_option
evgenyz: I'm working on it in #6100.

jan-cerny:
LGTM

PR #6090, opened 2020-09-16
Remove rhel8 COPY statement from ocp4_content and quay_publish Dockerfiles

JAORMX:
This was a leftover and is causing issues.


matejak:
The `COPY` command tries to copy things that aren't there (judging based on the `build_product` invocation), so removing this is a good move.

jhrozek (approved):
/lgtm

PR #6091, opened 2020-09-16
Add CCE identifiers to all RHEL7 and RHEL8 rules that were missing

ggbecker:
#### Description:
- Add CCE identifiers to all RHEL7 and RHEL8 rules that were missing.
  - The logic used was to check if the rule does not contain `prodtype`
attribute which means applicable to all products or if specific product `rhel7` or `rhel8`
was present. It also considers only rules from `linux_os/guide` directory.
- Doesn't consider rules having `documentation_complete: false` (not enabled in the project)

#### Rationale:

- No more missing CCE in RHEL7 and RHEL8 ;)

#### Additional Information

- Our [`shared/references/cce-redhat-avail.txt`](https://github.com/ggbecker/content/blob/cceb92d42735511f1d0293f168823a5461b88965/shared/references/cce-redhat-avail.txt) will be almost empty though (16 identifiers left)


redhatrises:
If we can do this only for rules in use by profiles atm, that would be better. We are currently going through an exercise where we are our submitting usage of CCE to NIST for approval.

ggbecker:
> If we can do this only for rules in use by profiles atm, that would be better. We are currently going through an exercise where we are our submitting usage of CCE to NIST for approval.

That's where we stand now. But we don't have any test to enforce that new rules added to profiles will have their CCE identifier in place when pull request checks are done.

ggbecker:
@openscap-ci test

ggbecker:
@openscap-ci test this

ggbecker:
@openscap-ci test this please

jan-cerny:
> But we don't have any test to enforce that new rules added to profiles will have their CCE identifier in place when pull request checks are done.

Check out this: https://github.com/ComplianceAsCode/content/pull/6097

ggbecker:
Let's close this for now. The test from #6097 should prevent us from introducing new rules into rhel6,7,8 profiles without having the CCE assigned.

PR #6097, opened 2020-09-17
Add a test for missing CCEs

jan-cerny:
#### Description:
This test checks if all rules that are a part of a profile have a CCE identifier properly assigned. It checks the built datastream. It will check RHEL 6, RHEL 7 and RHEL 8. The test is a part of CTest. It will be executed on every PR. 

#### Rationale:
It will prevent introducing rules without CCEs, so it will prevent last-minute PRs eg. https://github.com/ComplianceAsCode/content/pull/6079

PR #6103, opened 2020-09-18
Fix missing CCE in rules selected by RHEL6 profiles.

ggbecker:
#### Description:

- Fix missing CCE in rules selected by RHEL6 profiles.
- It also adds one CCI reference to RHEL6 rule `grub_legacy_enable_selinux`, reference:[V-51337](https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_6_security_technical_implementation_guide/V-51337)

#### Rationale:

- No more missing CCE for rules selected in RHEL6,7,8 profiles.

Related to #6097


jan-cerny:
The OpenShift CI failure is probably an infrastructure problem, I believe this change shouldn't affect OCP content.

PR #6104, opened 2020-09-18
Update DISA STIG RHEL7 reference files to latest version (v2r8).

ggbecker:
#### Description:

- Update DISA STIG RHEL7 reference files to latest version (v2r8).


rhel7/overlays/stig_overlay.xml
```diff
@@ -1,991 +1,2 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
yuumasato, line 1: Why did the overlay become a non-null terminated one liner? :laughing:

ggbecker: Something went wrong with the `create-stig-overlay.py` utility. Let me check.

ggbecker: I guess there is something going on with the ElementTree python class and library versions of `xml.etree`. It's not pretty printing the XML.

ggbecker: I couldn't make it work with `xml.etree` module [from here](https://github.com/ComplianceAsCode/content/blob/master/ssg/xml.py#L10). I had to use `lxml` combined with `pretty_print=True` when writing the contents to the file.

yuumasato (approved):
LGTM

rhel7/overlays/stig_overlay.xml
yuumasato: Do you think it makes sense to update the generating code?

ggbecker: I've opened: https://github.com/ComplianceAsCode/content/pull/6105 please check it out

PR #6105, opened 2020-09-18
Update create-stig-overlay to pretty print the XML content

ggbecker:
#### Description:

- Update create-stig-overlay to pretty print the XML content. It uses `lxml` and `python3`.

#### Rationale:

- Previous implementation was printing one liner XML.

### DIFF
--- a/utils/create-stig-overlay.py
+++ b/utils/create-stig-overlay.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python3
 
 from __future__ import print_function
 
@@ -8,6 +8,7 @@ import argparse
 import os
 import ssg
 import ssg.xml
+import xml.dom.minidom
 
 ET = ssg.xml.ElementTree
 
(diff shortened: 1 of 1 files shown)


mildas:
**Changes identified:**
Others:
 Python abstract syntax tree change found in utils/create-stig-overlay.py.

matejak:
I am a bit worried about the lxml dependency - we got rid of it in 2017, and it's not a current build dependency. What about [minidom pretty XML output](https://docs.python.org/3.8/library/xml.dom.minidom.html#xml.dom.minidom.Node.toprettyxml)?

Re-introducing it is not a bad idea, but I would bring this up on public mailing lists.

redhatrises:
> I am a bit worried about the lxml dependency - we got rid of it in 2017, and it's not a current build dependency. What about [minidom pretty XML output](https://docs.python.org/3.8/library/xml.dom.minidom.html#xml.dom.minidom.Node.toprettyxml)?
> 
> Re-introducing lxml is not a bad idea though, but I would bring this up on public mailing lists first before merging this PR.

We should continue the removal of lxml especially if minidom xml is provided by default in python. Readding a way to pretty print for readability is helpful.

jan-cerny:
@ggbecker yes, please replace it by minidom

ggbecker:
I have replaced with minidom. Please take a look.

redhatrises:
This file isn't managed by the CI/CD workflow and can be merged as long as others are good with the code changes.

PR #6107, opened 2020-09-21
Throw warning when ocp4 and rhcos4 content fail on scapval

ggbecker:
#### Description:

- Throw warning when `ocp4` content fail on `scapval`.

#### Rationale:

- At this moment, the content validation always fail because it contains the non-standardized content (`yamlfilecontent_test`). With a warning instead, we can keep the job always green and whenever a fail occurs, we can assume it's a legit one.


mildas:
**Changes identified:**
Others:
 Python abstract syntax tree change found in tests/run_scapval.py.

mildas:
**Changes identified:**
Others:
 Python abstract syntax tree change found in tests/run_scapval.py.

PR #6114, opened 2020-09-23
Removed profile from test metadata where not needed

matejak:
Historically, the profile had to be specified in test scenarios - oscap needs a profile that selects the rule to be able to scan for that rule using the `--rule` option. This became unnecessary when the `(alll)` virtual profile got introduced.

Nowadays, the `--profile` option is needed iff the rule is parametrized by an XCCDF variable that assigns a specific value to that variable. 

This PR removes the profile metadata from rules that have been identified as not parametrized using this thought: If a rule doesn't have `external` in their OVALs, or it doesn't have any `oval` directory at all, then the rule is not parametrized.

Here is the supportive bash code that was used to identify trees of test scenarios that have `profiles` metadata, but that don't need them.

```
test_trees_of_rules_with_profiles=$(grep -l profiles $(fd .pass.sh) $(fd .notapplicable.sh) $(fd .fail.sh) | sed -e 's|/tests/.*$|/tests|' | sort | uniq)
trees_with_external_vars=''
for tree in $test_trees_of_rules_with_profiles; do
    test -d "$tree/../oval" && grep -q external "$tree"/../oval/*.xml && trees_with_external_vars="$tree $trees_with_external_vars"
    grep -q '\<name:\s\+\(accounts_password\|mount_option_removable_partitions\|sebool\|sysctl\|yamlfile_value\)' "$tree"/../rule.yml && trees_with_external_vars="$tree $trees_with_external_vars"
done
trees_without_external_vars=
for tree in $test_trees_of_rules_with_profiles; do
    grep -q "\<$tree\>" <<< "$trees_with_external_vars" || trees_without_external_vars="$tree $trees_without_external_vars"
done
```

### DIFF
--- a/linux_os/guide/services/cron_and_at/service_atd_disabled/tests/service_disabled.pass.sh
+++ b/linux_os/guide/services/cron_and_at/service_atd_disabled/tests/service_disabled.pass.sh
@@ -1,6 +1,5 @@
 #!/bin/bash
 #
-# profiles = xccdf_org.ssgproject.content_profile_standard
 
 yum -y install at
 systemctl disable atd.service
--- a/linux_os/guide/system/accounts/accounts-session/user_umask/accounts_umask_etc_login_defs/tests/wrong_value.fail.sh
+++ b/linux_os/guide/system/accounts/accounts-session/user_umask/accounts_umask_etc_login_defs/tests/wrong_value.fail.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 #
+# profiles = xccdf_org.ssgproject.content_profile_ospp
 
 if grep -q "^UMASK" /etc/login.defs; then
 	sed -i "s/^UMASK.*/UMASK 027/" /etc/login.defs
--- a/linux_os/guide/system/accounts/accounts-session/user_umask/accounts_umask_etc_login_defs/tests/wrong_value.fail.sh
+++ b/linux_os/guide/system/accounts/accounts-session/user_umask/accounts_umask_etc_login_defs/tests/wrong_value.fail.sh
@@ -1,6 +1,5 @@
 #!/bin/bash
 #
-# profiles = xccdf_org.ssgproject.content_profile_ospp
 
 if grep -q "^UMASK" /etc/login.defs; then
 	sed -i "s/^UMASK.*/UMASK 027/" /etc/login.defs
--- a/linux_os/guide/services/avahi/disable_avahi_group/service_avahi-daemon_disabled/tests/service_disabled.pass.sh
+++ b/linux_os/guide/services/avahi/disable_avahi_group/service_avahi-daemon_disabled/tests/service_disabled.pass.sh
@@ -1,6 +1,5 @@
 #!/bin/bash
 #
-# profiles = xccdf_org.ssgproject.content_profile_C2S
 
 yum -y install avahi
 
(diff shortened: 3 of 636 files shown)


jan-cerny:
> or it doesn't have any oval directory at all, then the rule is not parametrized.

Is it correct to assume that these rules aren't parametrized? They are often templated rules. I have seen that some templates reference an external variable. I haven't checked if they're tested, though.

> $(fd .pass.sh) $(fd .fail.sh)

What about the tests with expected result other than pass or fail, for example error?

jan-cerny:
The following templates are referencing an external variable, so they might be worth checking.

template_OVAL_accounts_password
template_OVAL_mount_option_removable_partitions
template_OVAL_sebool
template_OVAL_sysctl
template_OVAL_yamlfile_value


We have noticed that if you remove a profile tag from all scenario for templated rules you can accidentally remove it even if it would be needed.

linux_os/guide/services/cron_and_at/service_atd_disabled/tests/service_disabled.pass.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# profiles = xccdf_org.ssgproject.content_profile_standard
+# packages = at
```
jan-cerny, line 3: Should the `yum` command be removed, then?

matejak: Not yet, there is no support for package metadata, but it is coming.

linux_os/guide/system/accounts/accounts-pam/password_quality/password_quality_pwquality/accounts_password_pam_ocredit/tests/ospp.pass.sh
```diff
@@ -1,5 +1,4 @@
 #!/bin/bash
 #
-# profiles = xccdf_org.ssgproject.content_profile_ospp
```
jan-cerny, line 3: There is variable `var_password_pam_ocredit.var` involved in this and for some profiles, this shouldn't pass, eg. rht-ccp profile requires ocredit to be set to -2.

linux_os/guide/system/accounts/accounts-session/user_umask/accounts_umask_etc_login_defs/tests/wrong_value.fail.sh
```diff
@@ -1,6 +1,5 @@
 #!/bin/bash
 #
-# profiles = xccdf_org.ssgproject.content_profile_ospp
```
jan-cerny, line 3: This depends on var_accounts_user_umask. In some profiles, it's set to 027 and in other profiles it's set to 077, so in some profiles this test should pass.

matejak:
I have rebased the PR with following changes to the previous state:

- Package installation and package metadata was not touched at all.
- Profile metadata is kept if a rule uses external variables or it uses a parametrized template.

linux_os/guide/system/accounts/accounts-session/user_umask/accounts_umask_etc_login_defs/tests/wrong_value.fail.sh
matejak: This is a good catch, cases like this are hard to handle automatically. I have exempted this rule's tests from the PR.

jan-cerny:
@openscap-ci test this please

jan-cerny:
Hi @matejak!

Thank you.

PR #6118, opened 2020-09-23
Introduce new template for dconf configuration files

ggbecker:
#### Description:

- Introduce new template for `dconf` configuration files.

#### Rationale:

- Reduce similar code in the project

- [x] Documentation


mildas:
**Changes identified:**
Rule dconf_gnome_disable_user_admin:
 Ansible remediation for dconf_gnome_disable_user_admin was deleted.
Others:

matejak:
I would consider using / leveraging [macros for OVAL ini content](https://github.com/ComplianceAsCode/content/blob/master/shared/macros-oval.jinja#L376) - we have some [OVAL ini test cases](https://github.com/ComplianceAsCode/content/blob/master/tests/test_macros_oval.py#L657), which are fast and easy to extend with new test cases, and which provide good confidence to the check.

ggbecker:
> I would consider using / leveraging [macros for OVAL ini content](https://github.com/ComplianceAsCode/content/blob/master/shared/macros-oval.jinja#L376) - we have some [OVAL ini test cases](https://github.com/ComplianceAsCode/content/blob/master/tests/test_macros_oval.py#L657), which are fast and easy to extend with new test cases, and which provide good confidence to the check.

I've managed to reuse some of the jinja macros (eg. `oval_line_in_file_test`), but `dconf`'s case is a bit special because it checks for two things, the configuration, and the configuration lock. With that being said, it was not possible to simply reuse some of the more high level macros without major rewriting. The same applies to OVAL unit tests.

The ideal scenario would be to split these `dconf` configuration checks into single configuration item checks that would allow for much better reuse of the macros, but that would mean double the rules we already have for `dconf` and also despite this is the desired state, most of the policies pack these two configurations into one single requirement, which makes harder for us to reason the splitting.

ggbecker:
Rendered documentation. Look how cool the example section looks like:

Screenshot from 2021-01-11 18-26-16

matejak:
> I've managed to reuse some of the jinja macros (eg. `oval_line_in_file_test`), but `dconf`'s case is a bit special because it checks for two things, the configuration, and the configuration lock. With that being said, it was not possible to simply reuse some of the more high level macros without major rewriting. The same applies to OVAL unit tests.

I see, it would be great to have those macros structured in a way that they would allow reusing the test-object-state tuple without instantiating the definition/criteria container.

> Rendered documentation. Look how cool the example section looks like ...

Great indeed, computers are very clever nowadays.

ggbecker:
RHEL8 STIG has now a dedicated requirement which group all dconf locks. It's finally apart from the dconf configuration but groups all of them into one single item, I don't know if this is a good or bad thing.

matejak:
Reusing OVAL code using `extend_definition` is an option that could help us to have a thin STIG-only rule, while granular rules in the benchmark would use the template.

matejak:
LGTM, tests pass. Great work!

PR #6127, opened 2020-09-28
add ocp4-node product as top level product

JAORMX:
This adds the ocp4-node product. This identifies the pieces of OCP4 that are only
accessible from the host as opposed to the platform.

This will be built alongside the rest of the ocp4 content. e.g. we only need to run `./build_product --debug ocp4`

This is an alternative approach to https://github.com/ComplianceAsCode/content/pull/6124 which
creates the `ocp4-node` product as a top-level product (instead of an embedded one) in the content.

The advantage of this approach is that we won't need to add the "platform" element to the Profile itself,
all the profiles will match in this benchmark.


mildas:
**Changes identified:**
Rule checks:
 OVAL check is newly added.
Profile cis on ocp4-node:
 Newly added profile.
Others:

 build_product ocp4-node
 tests/test_suite.py profile --libvirt qemu:///system test-suite-vm --datastream build/ssg-ocp4-node-ds.xml cis

redhatrises:
I think the embedded approach is the correct way vs having another top-level product directory here. IMO it creates too much confusion to have another top level for openshift that is just for the node.

PR #6129, opened 2020-09-29
Move grub2_vsyscall_argument to grub2 group

yuumasato:
#### Description:

- Move `grub2_vsyscall_argument` rule to `bootloader-grub2` group.
- This will put the rule under grub2 platform, so the rule is only applicable on a machine system with grub2.

#### Rationale:

- The current `restrictions` group has a miscellaneous of rules, and it is not possible to assign `grub2` and `machine`.

#### Discussion:
- Other Grub2 rules are scattered around and hard to find. I think having them together in one place would make them easier to find... Thoughts?


yuumasato:
@openscap-ci test this please

yuumasato:
@openscap-ci test this please

redhatrises:
The failures of the CIs have no relation to this change. Merging.

I think that centralizing the grub2 rules in the appropriate group would be a good thing to do.

PR #6134, opened 2020-09-30
Remove platform all from kubernetes remediation templates.

ggbecker:
#### Description:

- Remove platform `all` from kubernetes remediation templates.

#### Rationale:

- Do we really want to include kubernetes remediation in all product datastreams? Some rules (e.g. [sshd_enable_strictmodes](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/ssh/ssh_server/sshd_enable_strictmodes/rule.yml)) don't have `prodtype` set, then it will be included in every datastream increasing their size.

### DIFF
--- a/shared/templates/template_KUBERNETES_service_disabled
+++ b/shared/templates/template_KUBERNETES_service_disabled
@@ -1,4 +1,4 @@
-# platform = multi_platform_rhel,multi_platform_fedora,multi_platform_ocp,multi_platform_rhcos4
+# platform = multi_platform_rhel,multi_platform_fedora,multi_platform_ocp,multi_platform_rhcos
 # reboot = true
 # strategy = disable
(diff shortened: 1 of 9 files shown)


shared/templates/template_KUBERNETES_service_disabled
```diff
@@ -1,4 +1,4 @@
-# platform = multi_platform_rhel,multi_platform_fedora,multi_platform_ocp,multi_platform_rhcos4
+# platform = multi_platform_rhel,multi_platform_fedora,multi_platform_ocp,multi_platform_rhcos
```
ggbecker, line 1: in theory `multi_platform_rhcos4` doesn't work.

JAORMX (approved):

Issue #6113, opened 2020-09-22
RHEL7 STIG ids assigned to mutiple rules


ggbecker:
#### Description of problem:

Following rules contain the same STIG ID (FYI @redhatrises):

- `RHEL-07-010010`
  - [rpm_verify_ownership](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_ownership/rule.yml#L38)
  - [rpm_verify_permissions](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/software/integrity/software-integrity/rpm_verification/rpm_verify_permissions/rule.yml#L56)

- `RHEL-07-010320`
  - [accounts_passwords_pam_faillock_deny](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_passwords_pam_faillock_deny/rule.yml#L47)
  - [accounts_passwords_pam_faillock_interval](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_passwords_pam_faillock_interval/rule.yml#L53)
  - [accounts_passwords_pam_faillock_unlock_time](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_passwords_pam_faillock_unlock_time/rule.yml#L51)

- `RHEL-07-010481`
  - [require_emergency_target_auth](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/accounts/accounts-physical/require_emergency_target_auth/rule.yml#L28)
  - [require_singleuser_auth](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/accounts/accounts-physical/require_singleuser_auth/rule.yml#L40)

- `RHEL-07-021700`
  - [grub2_no_removeable_media](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/bootloader-grub2/grub2_no_removeable_media/rule.yml#L27)
  - [uefi_no_removeable_media](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/bootloader-grub2/uefi_no_removeable_media/rule.yml#L26)

- `RHEL-07-030340`
[...]

PR #6135, opened 2020-09-30
Add test to check for duplicated STIG ids

ggbecker:
#### Description:

- Add test to check for duplicated STIG ids.
- Refactor existent `tests/assert_cces_unique.sh` to accept arguments and rename it to `tests/assert_reference_unique.sh`

#### Rationale:

- We should strive for rules without any duplicated STIG id, meaning that rules that are using duplicated STIG ids should have a new STIG id requested from DISA. This test includes all rules from the project, some of them we might not want to touch, we can remove them from the test later.

Relates to: #6113

#### Additional Information

The test is failing due to many duplicated STIG ids present in the project.


matejak:
The code and test integration LGTM.

I have a small nitpick - we should probably make it clear that we want to eliminate duplicate *identifiers*, i.e. not *references* - when an identifier is duplicated, it doesn't identify anything very reliably, whereas one reference can relate to multiple rules. STIG IDs are identifiers that are nested under "references", but in fact, they are supposed to be identifiers.

Next, we can't merge a PR that breaks the CI. What about skipping it in the test, and adding a comment that references that issue? I can see that we have some sort of idea how to fix it, but it can take months, and the code doesn't have to wait for such long time to be merged.

ggbecker:
> The code and test integration LGTM.
> 
> I have a small nitpick - we should probably make it clear that we want to eliminate duplicate _identifiers_, i.e. not _references_ - when an identifier is duplicated, it doesn't identify anything very reliably, whereas one reference can relate to multiple rules. STIG IDs are identifiers that are nested under "references", but in fact, they are supposed to be identifiers.
> 

I don't agree that STIG ids are identifiers within the CaC/content context, they reference an identifier from another document. In fact, these ids are not duplicated in the original STIG benchmark, but we are assigning them to multiple rules in our project because sometimes one STIG item covers multiple configuration items when actually in the ideal scenario would be one configuration item per STIG item. We can state that we strive for one STIG item per configuration item, but we cannot control that too much since we have to request DISA to make this change.

> Next, we can't merge a PR that breaks the CI. What about skipping it in the test, and adding a comment that references that issue? I can see that we have some sort of idea how to fix it, but it can take months, and the code doesn't have to wait for such long time to be merged.

It makes sense, do you know how to disable the test using some `ctest` option, or should I just comment the test invocation?

yuumasato:
> It makes sense, do you know how to disable the test using some `ctest` option, or should I just comment the test invocation?

The `-E` option can help:
```
-E <regex>, --exclude-regex <regex>

    Exclude tests matching regular expression.

    This option tells ctest to NOT run the tests whose names match the given regular expression.
```

ggbecker:
> > It makes sense, do you know how to disable the test using some `ctest` option, or should I just comment the test invocation?
> 
> The `-E` option can help:
> 
> ```
> -E <regex>, --exclude-regex <regex>
> 
>     Exclude tests matching regular expression.
> 
>     This option tells ctest to NOT run the tests whose names match the given regular expression.
> ```

I have change the jenkins job configuration to skip the test, let's see if it works.

ggbecker:
@openscap-ci test this please

cmake/SSGCommon.cmake
```diff
@@ -1261,9 +1261,14 @@ macro(ssg_define_guide_and_table_tests)
         endforeach()
         add_test(
             NAME "unique-cces"
-	    COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tests/assert_cces_unique.sh"
+	    COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tests/assert_reference_unique.sh" "cce"
+        )
+        add_test(
+            NAME "unique-stigids"
+	    COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tests/assert_reference_unique.sh" "stigid"
```
jan-cerny, line 1268: weird indentation

jan-cerny (approved):
works fine

cmake/SSGCommon.cmake
```diff
@@ -1261,9 +1261,14 @@ macro(ssg_define_guide_and_table_tests)
         endforeach()
         add_test(
             NAME "unique-cces"
-	    COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tests/assert_cces_unique.sh"
+	    COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tests/assert_reference_unique.sh" "cce"
```
jan-cerny, line 1264: Please don't mix tabs and spaces.

jan-cerny (changes requested):

cmake/SSGCommon.cmake
ggbecker: I'm not mixing them. It was like this before. Anyway, I'll fix both indentation issues

cmake/SSGCommon.cmake
ggbecker: done

tests/assert_reference_unique.sh
```diff
@@ -0,0 +1,29 @@
...
+readarray -t duplicated_list < <(uniq -d <<< "$all_matches")
+
+function handle_duplicate {
+	local dup="$1" occurrence
+	# Find all files containing duplicated references, and
+	# shorten filenames, so they are relative to the project root.
+	readarray -t occurrence < <(grep -l "${REF}.*:\\s*\\<${dup}" "${all_rule_files[@]}" | sed -e "s|^$PROJECT_ROOT/||")
+
+	echo >&2
+	printf "${REF} %s is included in files: \n" "$dup" >&2
```
matejak, line 22:
```suggestion
	printf '%s %s is included in files: \n' "${REF}" "${dup}" >&2
```

matejak:
LGTM, I just suggest to use `printf` according to [shell best practices](https://github.com/koalaman/shellcheck/wiki/SC2059]

matejak:
@jan-cerny I think that your concern was addressed as well.

jan-cerny:
I'm sorry for being late. I think they have been addressed.

PR #6137, opened 2020-09-30
Remove Debian8

redhatrises:
It's EOL


tests/unit/ssg_test_suite/test_matches_platform.py
```diff
@@ -80,16 +80,16 @@ def test_list_combined_match():
 
 
 def test_list_combined_match_2():
-    scenario_platforms = ["Debian 8", "multi_platform_ubuntu", "openSUSE",
+    scenario_platforms = ["multi_platform_ubuntu", "openSUSE",
```
matejak, line 83:
```suggestion
    scenario_platforms = ["Debian 8", "multi_platform_ubuntu", "openSUSE",
```

If the Debian platform is removed, then the CPE has no reason to match any of the remaining platforms.

matejak:
It is surprising how complex it is to remove a product. Ideally, it should be just about removing the product directory and finding and eliminating occurrences that are pointed out by the build system.

I made a suggestion that should make the tests passing again, and it LGTM oterwise.

redhatrises:
Updates made. Test failure with OCP has no bearing on this PR.

matejak:
ACK, thank you for taking the time to separate this product from the rest of the project, so it can be removed!

Issue #2642, opened 2018-03-03
Allow specifying rsyslog package name


jokajak:
#### Description of problem:
We use rsyslog7 on RHEL6 which means the shared package_rsyslog_installed check fails. It would be nice if the package name were a variable that could be customized.

#### SCAP Security Guide Version:
0.1.28
xccdf_org.ssgproject.content_profile_stig-rhel6-server-upstream

#### Operating System Version:
scap-security-guide-0.1.28-3.el6.noarch

#### Steps to Reproduce:

1. Install RHEL6
2. Install rsyslog7 using yum shell (See https://access.redhat.com/solutions/1260253 for more details)
3. Set package_rsyslog_name variable to rsyslog7
4. Perform SCAP check

#### Actual Results:
package_rsyslog_installed fails

#### Expected Results:
Set package_rsyslog_name variable to rsyslog7
package_rsyslog_installed passes

ggbecker:
https://github.com/ComplianceAsCode/content/pull/6142 updates the package name to use `rsyslog7` instead since it's the latest available version in RHEL6

PR #6142, opened 2020-10-01
Update package_rsyslog_installed in RHEL6 to consider both rsyslog and rsyslog7 package

ggbecker:
#### Description:

- Update package_rsyslog_installed in RHEL6 to consider both rsyslog and rsyslog7 package
#### Rationale:

- Relates to #2642

### DIFF
--- a/linux_os/guide/system/logging/package_rsyslog_installed/rule.yml
+++ b/linux_os/guide/system/logging/package_rsyslog_installed/rule.yml
@@ -38,4 +38,5 @@ template:
     name: package_installed
     vars:
         pkgname: rsyslog
-        pkgname@rhel6: rsyslog7
+    backends:
+        oval@rhel6: "off"
--- a/linux_os/guide/system/logging/package_rsyslog_installed/rule.yml
+++ b/linux_os/guide/system/logging/package_rsyslog_installed/rule.yml
@@ -38,3 +38,4 @@ template:
     name: package_installed
     vars:
         pkgname: rsyslog
+        pkgname@rhel6: rsyslog7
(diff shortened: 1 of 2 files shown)


redhatrises:
Is rsyslog or rsyslog7 the default install for RHEL6 latest?

ggbecker:
> Is rsyslog or rsyslog7 the default install for RHEL6 latest?

It's `rsyslog`. Feel free to close if that should be the correct value when default installed package is the one we should aim for.

redhatrises:
> > Is rsyslog or rsyslog7 the default install for RHEL6 latest?
> 
> It's `rsyslog`. Feel free to close if that should be the correct value when default installed package is the one we should aim for.

Either we drop or we update the OVAL to check for either rsyslog or rsyslog7. Your call as RHEL6 is EOL 11/30/2020 (outside of EUS which I doubt we are going to anything at this point.)

ggbecker:
I have created a specific OVAL check for RHEL6 which considers both packages `rsyslog` and `rsyslog7` and the special trick:

```
    backends:
        oval@rhel6: "off"
```
in the rule.yml seems to work just fine.

yuumasato (approved):
> The downside of this implementation is that I had to drop the `EVR` parameter from the template. So it cannot check for a particular version of the package.

I don't see lack of EVR support as a problem. The original rule didn't require a specific version, and updates are unlikely.

linux_os/guide/system/logging/package_rsyslog_installed/rule.yml
```diff
@@ -38,3 +38,5 @@ template:
     name: package_installed
     vars:
         pkgname: rsyslog
+    backends:
+        oval@rhel6: "off"
```
yuumasato, line 42: The build system already has a preference for product specific implementations (`rhel6.xml` vs `shared.xml`).
So this is not strictly necessar.
But I like it as it makes explicit in the `rule.yml` that `rhel6` is not using the template implementation.

PR #6143, opened 2020-10-02
Add ISM references

shaneboulden:
Adds Australian Information Security Manual references to corresponding controls


linux_os/guide/services/usbguard/service_usbguard_enabled/rule.yml
```diff
@@ -23,6 +23,7 @@ identifiers:
 references:
     ospp: FMT_SMF_EXT.1
     srg: SRG-OS-000378-GPOS-00163
+    ism: 1418
```
ggbecker, line 26:
```suggestion
    ism: "1418"
```

When you have a single number you need to enclose with quotes to transform into a string, otherwise the build system will think this is a integer and it fails to build the content.

redhatrises:
Sweet! Glad we are adding these.

shaneboulden:
@ggbecker I think this is good to go now

redhatrises:
PR LGTM

PR #6146, opened 2020-10-02
SSGTS combined mode: use all profile where applicable

matusmarhefka:
This commit introduces a small change into the SSGTS combined mode.
When testing a selected profile and test scenario does not contain
`profiles` metadata the test is rather performed using ALL virtual
profile instead of the selected (tested) profile.
This will prevent failures which might be caused by the tested profile
selecting different values in tested variables compared to defaults.
The ALL profile is always selecting default values.

What this change means?
- If rule is general and behaves the same way no matter the profile
  (i.e. does not use variables) do not list profiles in the profiles
  metadata - rule will always be tested using the ALL profile.
- If rule is applicable only to some profiles or you would like to test
  some specific regression when certain values are selected use the
  profiles metadata to limit test scenario to be ran only when testing
  profiles listed in the metadata.

### DIFF
--- a/tests/ssg_test_suite/combined.py
+++ b/tests/ssg_test_suite/combined.py
@@ -39,14 +39,21 @@ class CombinedChecker(rule.RuleChecker):
         self._current_result = None
 
     def _modify_parameters(self, script, params):
-        # If there is no profiles metadata for a script we add the tested
-        # profile id into it. Otherwise we check the metadata and set it
-        # to self.profile (the tested profile) only if the  metadata
-        # contains self.profile - otherwise scenario is not supposed to be
-        # tested using the self.profile and we return empty profiles
+        # If there is no profiles metadata in a script we will use
+        # the ALL profile - this will prevent failures which might
+        # be caused by the tested profile selecting different values
+        # in tested variables compared to defaults. The ALL profile
+        # is always selecting default values.
+        # If there is profiles metadata we check the metadata and set
+        # it to self.profile (the tested profile) only if the metadata
+        # contains self.profile - otherwise scenario is not supposed to
+        # be tested using the self.profile and we return empty profiles
         # metadata.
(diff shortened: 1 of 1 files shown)


mildas:
@openscap-ci test this please

mildas:
Before:
```
$ python3 test_suite.py combined --libvirt qemu:///system test-suite-rhel8 --datastream ../bu
ild/ssg-rhel8-ds.xml ospp
INFO - Performing combined test using profile: xccdf_org.ssgproject.content_profile_ospp
INFO - xccdf_org.ssgproject.content_rule_grub2_pti_argument
INFO - Script correct.pass.sh using profile xccdf_org.ssgproject.content_profile_ospp OK
INFO - Script arg_not_there.fail.sh using profile xccdf_org.ssgproject.content_profile_ospp OK
INFO - xccdf_org.ssgproject.content_rule_grub2_uefi_password
INFO - Script password-set.pass.sh using profile xccdf_org.ssgproject.content_profile_ospp OK
```
After:
```
$ python3 test_suite.py combined --libvirt qemu:///system test-suite-rhel8 --datastream ../build/ssg-rhel8-ds.xml ospp
INFO - Performing combined test using profile: xccdf_org.ssgproject.content_profile_ospp
INFO - xccdf_org.ssgproject.content_rule_grub2_pti_argument
INFO - Script correct.pass.sh using profile (all) OK
INFO - Script arg_not_there.fail.sh using profile (all) OK
INFO - xccdf_org.ssgproject.content_rule_grub2_uefi_password
INFO - Script password-set.pass.sh using profile (all) OK
```

I like this change because this often caused error results in test scenarios that were using default values different from profile-specific values.

In future, we would like to have test scenarios with variables that will have profile-specific values (or default).

mildas:
@openscap-ci test this please

PR #6147, opened 2020-10-02
Adds a BuildConfig for OpenShift

jhrozek:
#### Description:
Creating these manifests allows to build a content image just by
pushing to a repo and then either triggering a build or setting
and automatic trigger.

#### Rationale:

This can be useful for folks who can't develop content locally
on their machine (e.g. because they are using a platform where
the dependencies can't be fullfilled).


JAORMX (approved):
/lgtm

PR #6153, opened 2020-10-05
Fix applicability or grub2 rules for aarch64 and ppc64le

yuumasato:
#### Description:

- Fix `grub2` platform applicability for arches aarch64 and ppc64le.
- ppc64le has a small caveat that grub2 packages will be installed even when using OPAL
  - So for ppc64le, we also check that OPAL is not used.

#### Rationale:

- `grub2` platform was checking for presence of `grub2-pc`, but this package is x86_64 only.


vojtapolasek:
Thanks for this quick fix, I tested on both ppc64le and aarch64 and it seems that rules are applied as expected now.

PR #6157, opened 2020-10-07
remove rngd related rules from rhel8 ospp and stig

vojtapolasek:
#### Description:

- remove rules related to rngd from rhel8 ospp and stig profiles

#### Rationale:

- rngd is no longer required for sufficient entropy

- CC effort


JAORMX (approved):
LGTM

PR #6159, opened 2020-10-07
remove rngd related rules from rhcos profiles

JAORMX:
Description:

* Remove rules related to rngd from rhel8 ospp and stig profiles
* Similar to #6157

Rationale:

* rngd not longer required for sufficient entropy


vojtapolasek:
LGTM and tests are passing, merging.

PR #6161, opened 2020-10-07
ocp4: Don't fail on transcient error

JAORMX:
When checking for updates on the MachineConfigPools, we were failing if
an error happened while getting the current satus. This is not ideal as
there might be transcient issues (e.g. AWS having a flaky network).

Instead of failing, this logs the error and continues. The error is
captured in the end, and will be reflected if there is indeed an issue
that the cluster can't recover from.


jhrozek (approved):
/lgtm
the fact that errorf fails the test will never stop biting me..

PR #6083, opened 2020-09-15
Drop "esc" package from install_smartcard_packages rule

ggbecker:
#### Description:

- Drop "esc" package from install_smartcard_packages rule and use [package_installed](https://github.com/ComplianceAsCode/content/blob/master/docs/manual/developer_guide.adoc#732-list-of-available-templates) template to generate new content since now it contains only the instruction to install `pam_pkcs11` package.

#### Rationale:

- Reference: https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-72417?version=V1R4&compareto=v2r8

- This package is not recommended anymore by STIG. Unless we want to keep it in a similar way as `smartcard_auth` has when `GConf2/dconf` is installed: [smartcard_auth/oval/shared.xml#L16](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/smartcard_auth/oval/shared.xml#L16)

- It also fixes some SRG mappings that are still typos in official RHEL7 STIG, these are the actual SRG that were used as reference:
    - [SRG-OS-000377-GPOS-00162](https://vaulted.io/library/disa-stigs-srgs/general_purpose_operating_system_srg/V-56813)
    - [SRG-OS-000376-GPOS-00161](https://vaulted.io/library/disa-stigs-srgs/general_purpose_operating_system_srg/V-56815)

PR #6162, opened 2020-10-07
Set the OVAL ID of templated OVALs to the rule ID

matejak:
Provide a fallback if the rule ID is not known - this may happen in cases when the OVAL is generated using the template for another reason than because a rule needs it.

### DIFF
--- a/shared/templates/template_OVAL_accounts_password
+++ b/shared/templates/template_OVAL_accounts_password
@@ -1,5 +1,5 @@
 <def-group>
-  <definition class="compliance" id="accounts_password_pam_{{{ VARIABLE }}}" version="3">
+  <definition class="compliance" id="{{{ _RULE_ID }}}" version="3">
     {{{ oval_metadata("The password " + VARIABLE + " should meet minimum requirements") }}}
 {{% if product == "rhel6" %}}
     <criteria>
--- a/shared/templates/template_OVAL_audit_rules_dac_modification
+++ b/shared/templates/template_OVAL_audit_rules_dac_modification
@@ -1,5 +1,5 @@
 <def-group>
-  <definition class="compliance" id="{{{ rule_id }}}" version="1">
+  <definition class="compliance" id="{{{ _RULE_ID }}}" version="1">
     {{{ oval_metadata("The changing of file permissions and attributes should be audited.") }}}
     <criteria operator="OR">
 
--- a/shared/templates/template_OVAL_audit_rules_file_deletion_events
+++ b/shared/templates/template_OVAL_audit_rules_file_deletion_events
@@ -1,5 +1,5 @@
 <def-group>
-  <definition class="compliance" id="audit_rules_file_deletion_events_{{{ NAME }}}" version="1">
+  <definition class="compliance" id="{{{ _RULE_ID }}}" version="1">
     {{{ oval_metadata("The deletion of files should be audited.") }}}
 
     <criteria operator="OR">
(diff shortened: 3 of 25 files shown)


shared/templates/template_OVAL_accounts_password
```diff
@@ -1,5 +1,11 @@
+{{% if rule_id -%}}
+{{% set oval_id = rule_id -%}}
+{{% else -%}}
+{{% set oval_id = "accounts_password_pam_" + VARIABLE -%}}
+{{% endif -%}}
+
 <def-group>
-  <definition class="compliance" id="accounts_password_pam_{{{ VARIABLE }}}" version="3">
+  <definition class="compliance" id="{{{ oval_id }}}" version="3">
```
jan-cerny, line 8: @matejak 

I think that @jhrozek has problem with file_permissions template and not with this one right now, but anyway it's a perfect start because I think we will fix all the templates we have.

shared/templates/template_OVAL_accounts_password
```diff
@@ -1,5 +1,11 @@
+{{% if rule_id -%}}
+{{% set oval_id = rule_id -%}}
+{{% else -%}}
+{{% set oval_id = "accounts_password_pam_" + VARIABLE -%}}
+{{% endif -%}}
```
jan-cerny, line 5: I assume this construction is here due to so-called unowned templates (I mean shared/templates/extra_ovals.yml). 
At this moment it isn't used there. I know we need to be ready for the future. However, I don't want it here. I would prefer a generic solution instead.

The top level keys in shared/templates/extra_ovals.yml are fake rule IDs in fact. They would be rule IDs if the generated OVAL definition would be owned by a rule. So I think that the ssg/templates.py script can use the top level keys and substitute them to rule_id or some different variable so that it can substitute it into the OVAL template as the OVAL definition ID.

jan-cerny (commented):
@matejak 

I think that @jhrozek has problem with file_permissions template and not with this one right now, but anyway it's a perfect start because I think we will fix all the templates we have.

shared/templates/template_OVAL_accounts_password
matejak: Yes, I would like to port all templates, but I have started with one and then just apply the same technique to the rest.

shared/templates/template_OVAL_accounts_password
matejak: I see, this is a good point. Technically I would prefer to use `oval_id`, as `rule_id` creates an expectation that it is indeed an ID of a rule.
I agree that it is more appropriate to put the OVAL ID definition to `templates.py`.

matejak: I can confirm that `extra_ovals` act as an input for templates that defines a `rule_id`, although technically there is no rule.
Therefore, one can use the rule_id as oval_id right away, with one exception - the sysctl rules are somehow convoluted, and the substitution generates conflicting IDs.

jan-cerny:
It looks good to me and I have randomly checked couple of rules and they're generated.

matejak:
Fun fact: `linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/install_smartcard_packages`  rule used to define a `package_pam_pkcs11_installed` OVAL. So while the rule was notchecked (XCCDF/OVAL ID mismatch), the associated OVAL is reused by other rules.

ggbecker:
> Fun fact: `linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/install_smartcard_packages` rule used to define a `package_pam_pkcs11_installed` OVAL. So while the rule was notchecked (XCCDF/OVAL ID mismatch), the associated OVAL is reused by other rules.

Actually it was not. In this pull request you can find all the details about this little journey: https://github.com/ComplianceAsCode/content/pull/6083

matejak:
I see, so the rule was operational and in line with its title and ID before #6083, but it got into that weird state only for a short period after that PR. Therefore, changing the rule ID is indeed a bad idea.

PR #6163, opened 2020-10-07
OpenShift: Move all OVAL to be ocp3 specific

JAORMX:
None of it applies to OCP4, so let's work on that in separate commit.
Most of them will be worked on as part of the CIS work.

This also marks the ocp4 CIS profile with "documetation_complete", with
the intent of enabling its building without the `--debug` flag and adding a CI
job for it in the near future.


JAORMX:
Pull-request updated, HEAD is now fc950300608334932fb7160f076aefe418eb88e0

jhrozek (approved):
/lgtm

yuumasato:
Some of the checks moved to `ocp3.xml` had jinja conditionals for `ocp3` or `ocp4`.
For sake of consistency, it would have been good to remove the conditionals and `ocp4` code blocks.

PR #6166, opened 2020-10-08
Remove platform net-snmp from the group and use it in individual rules.

ggbecker:
#### Description:

- Remove platform net-snmp from the group and use it in individual rules.

#### Rationale:

- This was causing the platform being introduced in every datastream, since the `group.yml` doesn't use `prodtype`, we produce but the respective CPE dictionaries didn't contain the platform definition. I guess restricting the platform to individual rules is the most reasonable way to fix the issue, instead of adding the CPE definition to every CPE dictionary we have in the project.

- Fixes failing jobs from:
  - https://jenkins.complianceascode.io/view/All%20Green/job/scap-security-guide-scapval-scap-1.2/
  - https://jenkins.complianceascode.io/view/All%20Green/job/scap-security-guide-scapval-scap-1.3/


vojtapolasek (approved):
lgtm, thank you. We will have to mind this when using platforms in group.yaml in future.

PR #6168, opened 2020-10-08
OCP: Remove jinja references to ocp4 in ocp3 OVAL files

JAORMX:
These were leftovers from a previous commit that made these files
ocp3-specific.


yuumasato (approved):
Thank you for these changes.

PR #6170, opened 2020-10-08
 stop using rules about ssh idle timeout

vojtapolasek:
#### Description:

- remove rules and variable configurations from all profiles:

    - sshd_set_idle_timeout

    - sshd_set_keepalive

- replace rule with accounts_tmout where current prodtype of accounts_tmout allows it

#### Rationale:

It turned out that rules sshd_set_idle_timeout and sshd_set_keepalive are not designed in a way so that they could enhance security. Therefore they are being removed and replaced by accounts_tmout in cases when it makes sense. Configuration options connected to these rules are not tied to SSH interactive session timeouts at all.


vojtapolasek:
@redhatrises  @JAORMX  please review changes. Thank you.

redhatrises:
100% nack.

matejak:
If we want to keep that rules, we will need to come up with different description.

The current incorrect assumption was that the setting terminates SSH sessions that are idle for a given period of time.
In reality, those settings terminate the connection if there is no activity from the SSH client. If the session is active, there will be activity from the client for sure. However, there may be activity from the client even if the session is idle.

redhatrises:
> If we want to keep that rules, we will need to come up with different description.
> 
> The current incorrect assumption was that the setting terminates SSH sessions that are idle for a given period of time.
> In reality, those settings terminate the connection if there is no activity from the SSH client. If the session is active, there will be activity from the client for sure. However, there may be activity from the client even if the session is idle.

Which is the same thing. Active sessions are active user sessions.

matejak:
> Active sessions are active user sessions.

I don't think so, the session can be lively even if the user is away from the keyboard - the system may post messages to the console, and those get forwarded back to the user. But I think that you should discuss the matter with Mr. Poole in https://bugzilla.redhat.com/show_bug.cgi?id=1873547 as he seem to be better informed than us.

redhatrises:
> > Active sessions are active user sessions.
> 
> I don't think so, the session can be lively even if the user is away from the keyboard - the system may post messages to the console, and those get forwarded back to the user. But I think that you should discuss the matter with Mr. Poole in https://bugzilla.redhat.com/show_bug.cgi?id=1873547 as he seem to be better informed than us.

And that is the definition of an active remote user session.

redhatrises:
Closing this as this goes against NSA SNAC guidance which is also a part of NIST 800-53 requirements.

PR #6172, opened 2020-10-09
Adopt changes in yamlfilecontent_* check for yamlfile_value template

evgenyz:
These changes would allow to use `yamlfile_value` macro with the new version of the YAML check (openscap 1.3.4).


evgenyz:
@openshift-ci-robot test this please

jhrozek (approved):
This seems to unblock the scanner in the sense that it now runs, the rules are checked and report results.
Thank you very much.

JAORMX (approved):
Waiting for CI run before merging.

PR #6175, opened 2020-10-09
utils/add_platform_rule: Fix created rule identation

JAORMX:
A recent commit #6172 changed the values but didn't take into account
the identation. This fixes that.

In addition, this fixes other funky issues with the generated yaml
identation.


jhrozek (approved):
/lgtm
tested with scap 1.3.4

PR #6177, opened 2020-10-09
Introduce ability to set check_existence to yaml template

JAORMX:
This will only override the `check_existence` setting for the yamltest
and not the file test.


evgenyz:
If it would be working for you, it's a good idea to update document this new argument in `docs/manual/developer_guide.adoc` and update `add_platform_rule.py` in some way.

JAORMX:
Pull-request updated, HEAD is now 7e5bbb2cad03d886fc4449c1cbab4e824ad8c70d

PR #6178, opened 2020-10-09
Add a temporary rhcos4/profiles/cis.profile until we get the ocp4-node profile working

jhrozek:
#### Description:

Because the ocp4-node profile still has some issues, I'm submitting the example
rules in a throwaway rhcos4/profiles/cis.profile

The profile contains some example rules for rules that can be added on these
example rules comments how to do it.

There's also an example etcd rule we take clone to check for other etcd
parameters util we take advantage of the richer yaml querying that's been
introduced in the latest scap release.

#### Rationale:

- We need something for next week's hackfest

### DIFF
--- a/rhcos4/profiles/cis.profile
+++ b/rhcos4/profiles/cis.profile
@@ -1,0 +1,63 @@
+documentation_complete: true
+
+title: 'CIS Red Hat OpenShift Container Platform 4 Benchmark'
+
+description: |-
+    This profile defines a baseline that aligns to the Center for Internet Security®
+    Red Hat OpenShift Container Platform 4 Benchmark™, V0.3, currently unreleased.
+
+    This profile includes Center for Internet Security®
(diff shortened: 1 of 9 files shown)


rhcos4/profiles/cis.profile
```diff
@@ -0,0 +1,63 @@
+documentation_complete: true
+
+title: 'CIS Red Hat OpenShift Container Platform 4 Benchmark'
+
+description: |-
```
jan-cerny, line 5: I think that the description should mention that this is going to be thrown away.

jhrozek: Thank you for taking a look, in general I agree but with @yuumasato 's help we've been able to fix the cis-node profile, so I'm going to close this PR and continue there.

jhrozek:
This PR is deprecated in favour of the cis-node profile.

PR #6180, opened 2020-10-12
Fixes for the cis-node profile

jhrozek:
#### Description:

- The cis-node profile used to only check for the OCP4 kubernetes CPE which
  relies on fetching an object from OCP. Some checks are node-only though.

#### Rationale:

- Enabled OCP node checks. There is also one example rule, more to follow.

### DIFF
--- a/applications/openshift/worker/file_owner_worker_service/rule.yml
+++ b/applications/openshift/worker/file_owner_worker_service/rule.yml
@@ -4,10 +4,20 @@ prodtype: ocp4
 
 title: 'Verify User Who Owns The OpenShift Node Service File'
 
-description: '{{{ describe_file_owner(file="/etc/systemd/system/atomic-openshift-node.service", owner="root") }}}'
+description: |-
+{{%- if product == "ocp4" %}}
+  '{{{ describe_file_owner(file="/etc/systemd/system/kubelet.service", owner="root") }}}'
+{{% else %}}
+  '{{{ describe_file_owner(file="/etc/systemd/system/atomic-openshift-node.service", owner="root") }}}'
+{{%- endif %}}
 
 rationale: |-
-    The <tt>/etc/systemd/system/atomic-openshift-node.service</tt> file contains information about the configuration of the
+{{%- if product == "ocp4" %}}
+    The <tt>/etc/systemd/system/kubelet.service</tt>
+{{% else %}}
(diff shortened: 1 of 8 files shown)


JAORMX (approved):
/lgtm

applications/openshift/worker/file_owner_worker_service/rule.yml
```diff
@@ -19,12 +29,31 @@ severity: medium
...
+    vars:
+        filepath: /etc/systemd/system/kubelet.service
+        fileuid: '0'
+{{% else %}}
 template:
     name: file_owner
     vars:
         filepath: /etc/systemd/system/atomic-openshift-node.service
         fileuid: '0'
+{{%- endif %}}
```
yuumasato, line 59: @JAORMX @jhrozek Do I understand it correctly that this rule is expected to be present and behave differently for `ocp4` and `ocp4-node`?

Hey, I thought the rule sets  for `ocp4` and `ocp4-node` mutually exclusive. That a rule present in one wouldn't be in the other...

jhrozek:
On Mon, Oct 12, 2020 at 04:04:18AM -0700, Watson Yuuma Sato wrote:
> @yuumasato commented on this pull request.
> 
> 
> 
> > +{{%- if product == "ocp4" %}}
> +template:
> +    name: file_owner
> +    vars:
> +        filepath: /etc/systemd/system/kubelet.service
> +        fileuid: '0'
> +{{% else %}}
>  template:
>      name: file_owner
>      vars:
>          filepath: /etc/systemd/system/atomic-openshift-node.service
>          fileuid: '0'
> +{{%- endif %}}
> 
> @JAORMX @jhrozek Do I understand it correctly that this rule is expected to be present and behave differently for `ocp4` and `ocp4-node`?
> 
> Hey, I thought the rule sets  for `ocp4` and `ocp4-node` mutually exclusive. That a rule present in one wouldn't be in the other...

No, the else is for ocp3. I thought we talked about the rules being
mass-converted to ocp3 despite still being valid? That's why I didn't
just nuke the ocp3 content but used if-else conditions. The atomic
service is ocp3 only, kubelet is ocp4.

yuumasato:
Cool, ok. I got confused by the `prodtype` in the rule, which is just `ocp4`.
In this case, I think the `else` case won't match `ocp3`, or anything.

It should not be a problem for `ocp4` content.

Issue #6185, opened 2020-10-12
Missing referenced OVAL


matejak:
During the build, following rules are reported as missing target of `extend_definition`. You can get the list by running `./build_product 2>&1 | grep 'extends non-existing'`.

Here is an exhaustive list for all products:

- audit_rules_file_deletion_events_rmdir
- audit_rules_kernel_module_loading_init
- audit_rules_login_events_tallylog
- audit_rules_unsuccessful_file_modification_creat
- enable_dconf_user_profile
- firewalld_sshd_port_enabled
- installed_OS_is_FIPS_certified
- install_mcafee_cma_rt
- ocp_service_runtime_config_streaming_connection_timeout
- package_aide_installed
- package_MFEhiplsm_installed
- package_openssh-server_removed
- package_pam_pkcs11_installed
- service_postfix_enabled
- sysctl_static_net_ipv6_conf_all_disable_ipv6

Concerning RHEL7,8 RHCOS4 and Fedora, there are only two issues:

- [x] package_MFEhiplsm_installed
- [x] package_pam_pkcs11_installed

An OVAL being taken out of the build because the required content is not available is a red flag, and either the referenced content should be made available, or the rule should be removed.

vojtapolasek:
Note that this kind of issue is fixed for rhel7, rhel8 and fedora in current master.

matejak:
A test would be nice. I will think of something.

PR #6186, opened 2020-10-12
Fix rules so no there are no "missing extend_definition" warnings during the build

matejak:
This PR addresses Fedora, RHEL7,8 and RHCOS4 products.

- The rule `install_smartcard_packages` has been renamed to `package_pam_pkcs11_installed`. The former rule didn't have an OVAL check associated, and the new ID reflect what the rule does more accurately. If this change is rejected, an alternative solution would be to fix references, and reference `install_smartcard_packages` rather than `package_pam_pkcs11_installed`.
- Prodtypes were added to pass the build without dropped OVALs on Fedora and CoreOS.

See also: #6185

### DIFF
--- a/linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/install_smartcard_packages/rule.yml
+++ b/linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/install_smartcard_packages/rule.yml
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: ol7,rhel7
+prodtype: fedora,ol7,rhel7
 
 title: 'Install Smart Card Packages For Multifactor Authentication'
 
--- a/linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/smartcard_auth/oval/shared.xml
+++ b/linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/smartcard_auth/oval/shared.xml
@@ -2,7 +2,7 @@
   <definition class="compliance" id="smartcard_auth" version="3">
     {{{ oval_metadata("Enable Smart Card logins") }}}
     <criteria comment="smart card authentication is configured" operator="AND">
-      <extend_definition comment="pam_pkcs11 package is installed" definition_ref="package_pam_pkcs11_installed" />
+      <extend_definition comment="packages needed for smartcard support are installed" definition_ref="install_smartcard_packages" />
       <extend_definition comment="pcscd service is enabled" definition_ref="service_pcscd_enabled" />
       <criteria operator="OR">
         <extend_definition comment="esc package is installed" definition_ref="package_esc_installed" />
--- a/linux_os/guide/system/software/integrity/endpoint_security_software/mcafee_security_software/mcafee_hbss_software/package_MFEhiplsm_installed/rule.yml
+++ b/linux_os/guide/system/software/integrity/endpoint_security_software/mcafee_security_software/mcafee_hbss_software/package_MFEhiplsm_installed/rule.yml
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: fedora,rhel6,rhel7,rhel8,rhv4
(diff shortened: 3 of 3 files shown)


linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/package_pam_pkcs11_installed/rule.yml
```diff
@@ -1,6 +1,6 @@
 documentation_complete: true
```
ggbecker, line 1: I'm worried about the renaming breaking existent tailorings.

redhatrises: Exactly. Not only does it break backwards compatibility, it is a non-relevant change. It also breaks CCE mapping, and there are more packages in RHEL7 than just pkcs11 getting installed. This PR introduces more problems than it solves.

ggbecker: Actually, latest version of STIG for RHEL7 now only requires `pam_pkcs11` package to be installed, that's why this rule now uses the `package_installed` template: https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-72417

So the only problem here is the backwards compatibility issue

redhatrises: Ummm.... no not true... CCE breaking is a problem and backwards compatibility is a problem. The appropriate fix is to have the OVAL generated.

matejak: OK, can do it the other way around.
I see that tailoring compatibility can be an issue even for rules without OVAL - one may want to tailor them out exactly for this reason.

ggbecker: This rule had OVAL check before, but since now it checks only for a specific package it didn't make sense to keep the custom OVAL, so I changed some time ago to use the `package_installed` template instead, but I didn't foresee that the OVAL wouldn't get generated... that's because the generated definition ID of the check was different from the rule's ID. But the template seems to be using the rule's id now: https://github.com/ComplianceAsCode/content/blob/master/shared/templates/template_OVAL_package_installed#L2 so the OVAL check is supposed to be correctly assigned to this rule.

ggbecker: @matejak 
I've identified the issue. Actually we are still using the OVAL definition `package_pam_pkcs11_installed` in  https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/smartcard_auth/oval/shared.xml#L5

And this definition was removed in https://github.com/ComplianceAsCode/content/commit/c73da05b65742ca6820a9b7c0fc9ee22097c805a#diff-52ce1eba7a9574ad3c71d2419debb827

So we have two options, revert back this change to have this extra OVAL generated or change this line https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/smartcard_auth/oval/shared.xml#L5

to:
`      <extend_definition comment="pam_pkcs11 package is installed" definition_ref="install_smartcard_packages" />`

matejak: Indeed, I think that the second option is better.

ggbecker: I guess everything from here has been addressed.

ggbecker (approved):

PR #6189, opened 2020-10-13
add rule configuring kernel to trust CPU RNG into rhel8 OSPP

vojtapolasek:
#### Description:

- add new rule which checks if the kernel is configured to trust CPU RNG

- remediations are templated, oval is custom

#### Rationale:

- rhel8 CC effort


linux_os/guide/system/software/integrity/kernel_trust_cpu_rng/rule.yml
```diff
@@ -0,0 +1,49 @@
+documentation_complete: true
+
+prodtype: rhel8
+
+title: 'Configure kernel to trust the CPU random number generator'
+
+description: |-
+    To configure the Linux kernel to trust the CPU hardware random number
+    generator, make sure that <tt>random.trust_cpu=on</tt> is configured as boot
```
redhatrises, line 9:
```suggestion
    generator, make sure that <tt>random.trust_cpu=on</tt> is configured as a boot
```

linux_os/guide/system/software/integrity/kernel_trust_cpu_rng/rule.yml
```diff
@@ -0,0 +1,49 @@
...
+    parameter. If using Grub2 bootloader, you can add it with the following
+    command:
```
redhatrises, line 11:
```suggestion
    parameter by running the following command:
```

linux_os/guide/system/software/integrity/kernel_trust_cpu_rng/rule.yml
```diff
@@ -0,0 +1,49 @@
...
+    Make sure that the kernel is configured to trust the CPU RNG by following
+    commands. To check if the option was correctly configured at kernel compile
+    time, run the following command:
+    <pre>grep -q CONFIG_RANDOM_TRUST_CPU=y /boot/config-`uname -r`</pre>
+    If the command does not return any output, then the option is not compiled
+    in. To check if the option is configured through boot parameter, run the
+    following command:
+    <pre>sudo grep 'kernelopts.*random\.trust_cpu=on.*' /boot/grub2/grubenv</pre>
+    If the command does not return any output, then the boot parameter is
+    missing.
```
redhatrises, line 41:
```suggestion
    To ensure that the kernel is compiled to trust the CPU random number generator, 
    run the following command:
    <pre>grep -q CONFIG_RANDOM_TRUST_CPU=y /boot/config-`uname -r`</pre>
    If the command does not return any output, then the option is not compiled
    into the kernel.
    In addition, to ensure that the bootloader correctly enables the CPU
    random number generator trust,  if the option is configured through boot parameter, run the
    following command:
    <pre>sudo grep 'kernelopts.*random\.trust_cpu=on.*' /boot/grub2/grubenv</pre>
    If the command does not return any output, then the boot parameter is
    missing and the random number generator is not trusted.
```

vojtapolasek:
I extended the rule, because even if the option is compiled in, it can be disabled through boot parameter.

yuumasato (approved):
I see that feedback by Gabe has been addressed.
Test coverage is good and the rule is working.

PR #6191, opened 2020-10-13
remove package_tuned_removed from rhel8 ospp

vojtapolasek:
#### Description:

- remove the rule from rhel8 ospp profile

#### Rationale:

- rhel8 CC effort


redhatrises:
Ack

PR #6192, opened 2020-10-13
e2e/ocp4: Display more verbose logs for e2e tests

JAORMX:
This is aimed at helping folks figure out if the tests they're writing
are actually being executed and what the results are.


jhrozek (approved):
I think this is good, if someone is looking at the logs chances are they are trying to figure out something in which case it is better to be more verbose
/lgtm

PR #6193, opened 2020-10-13
Fix rule api_server_token_auth for ocp4

JAORMX:
This rule was missing the actual check for ocp4. Now it's there.


ggbecker:
@openscap-ci test this please

PR #6195, opened 2020-10-13
Add api_server_auth_mode_node rule to ocp4/cis profile

JAORMX:
This rule checks that the authorization-mode is set to Node in the
api-server.

### DIFF
--- a/applications/openshift/api-server/api_server_auth_mode_node/rule.yml
+++ b/applications/openshift/api-server/api_server_auth_mode_node/rule.yml
@@ -1,0 +1,38 @@
+prodtype: ocp4
+
+title: Must have authorization-mode Node set
+
+description: 'Restrict kubelet nodes to reading only objects associated with them.'
+
+rationale: |-
+  The Node authorization mode only allows kubelets to read Secret,
+  ConfigMap, PersistentVolume, and PersistentVolumeClaim objects
+  associated with their nodes.
+
+severity: medium
+
+references:
+    cis: 1.2.8
+
+ocil_clause: 'The Node authorization-mode is configured and enabled on the master node'
+
+ocil: |-
+    Run the following command on the master node(s):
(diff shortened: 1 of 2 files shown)


redhatrises:
@JAORMX Should `applications/openshift/api-server/api_server_authorization_mode/rule.yml` be fixed instead.

dahaic:
> @JAORMX Should `applications/openshift/api-server/api_server_authorization_mode/rule.yml` be fixed instead.

Each of 1.2.7, 1.2.8 and 1.2.9 has its own specifics to check. Thus we have decided to go with three rules, and revert the `api_server_authorization_mode` back to OCP3 only. See #6209 where I am doing the revert.

applications/openshift/api-server/api_server_auth_mode_node/rule.yml
```diff
@@ -0,0 +1,38 @@
...
+
+template:
+  name: yamlfile_value
+  vars:
+    ocp_data: "true"
+    entity_check: "at least one"
+    filepath: /api/v1/namespaces/openshift-kube-apiserver/configmaps/config
+    yamlpath: '.data["config.yaml"]'
+    values:
+    - value: '"authorization-mode":\[.*Node"\]'
```
dahaic, line 36: This construct is dependable on the order of the values, isn't it? I propose slightly different regex `'"authorization-mode":\[[^]]*"Node"'`

JAORMX: I'm not acquainted with this notation: `\[[^]]*` what's that?

JAORMX:
Pull-request updated, HEAD is now e01d571096fdca59f3f98f7e8b361864dca86b13

applications/openshift/api-server/api_server_auth_mode_node/rule.yml
JAORMX: thanks! added that

PR #6196, opened 2020-10-13
Cis node 4.1.8

yuumasato:
#### Description:

- Select rules for OCP4 `cis-node` 4.1.8.
  Rules to check owner and groupowner of `/etc/kubernetes/kubelet-ca.crt`
- Ad CCEs

#### Notes
- Rules already existed, were added in https://github.com/ComplianceAsCode/content/pull/5833


applications/openshift/worker/file_groupowner_worker_ca/rule.yml
```diff
@@ -13,8 +13,8 @@ rationale: |-
 
 severity: medium
 
-#identifiers:
-#    cce@ocp4: 80619-0
```
yuumasato, line 17: @redhatrises was this CCE added as a place holder?
Is the CCE valid and assigned tot this rule actually?

redhatrises: no, it's not a valid CCE. It is a placeholder since the content is still considered alpha / beta/

PR #6197, opened 2020-10-13
ocp4: Fix api_server_admission_control_plugin_AlwaysAdmit rule

JAORMX:
This rule was not checking for the right file. Now it checks the correct
API value.


applications/openshift/api-server/api_server_admission_control_plugin_AlwaysAdmit/rule.yml
```diff
@@ -6,10 +6,11 @@ title: 'Disable the AlwaysAdmit Admission Control Plugin'
 
 description: |-
     To ensure OpenShift only responses to requests explicitly allowed by the
-    admissions control plugin, edit the API Server pod specification file
 {{%- if product == "ocp4" %}}
-    <tt>/etc/kubernetes/static-pod-resources/configmaps/config/config.yaml</tt> on the master node(s)
+    admission control plugin. Check that the 'config' ConfigMap object does not
```
redhatrises, line 10: nitpick:

```suggestion
    admission control plugin. Check that the <code>config</code> ConfigMap object does not
```

PR #6207, opened 2020-10-13
Fix api server authorization mode

rflorenc:
#### Description:

- Update api_server_authorization_mode rule for OCP 4


dahaic:
Hello @rflorenc , as discussed in #6196 and #6209, it's better to leave rule `api_server_authorization_mode` alone (or leave it on the #6209 to revert, and to introduce a new rule. It will be cleaner solution.

PR #6208, opened 2020-10-13
Decorate my recently added OCP4 CIS rules with CCE identifiers

jhrozek:
#### Description:

- Adds CCE identifiers to a couple of rules I added recently to the OCP4 CIS profile

#### Rationale:

- All rules should have CCE identifiers


redhatrises:
The OCP tests failures have nothing to do with this PR. ack

PR #6214, opened 2020-10-13
ocp4: Add api_server_admission_control_plugin_NamespaceLifecycle rule

JAORMX:
This rule didn't work in ocp4. Now it does!


redhatrises:
I think we need to work on the default message as it feels needing for more detail. But, that's something I can work on later.

PR #6218, opened 2020-10-13
[WIP DO NOT MERGE] CIS Node 4.2.9 add template to kubelet_configure_event_creation/rule.yml

lsm5:
Signed-off-by: Lokesh Mandvekar <lsm5@fedoraproject.org>

#### Description:

- Adds a template to CIS Node 4.2.9 kubelet_configure_event_creation/rule.yml

#### Rationale:

- Necessary to run the scan

### DIFF
--- a/applications/openshift/kubelet/kubelet_configure_event_creation/rule.yml
+++ b/applications/openshift/kubelet/kubelet_configure_event_creation/rule.yml
@@ -2,28 +2,26 @@ documentation_complete: true
 
 prodtype: ocp3,ocp4
 
-title: 'kubelet - Do Not Limit Event Creation'
+title: 'kubelet - Limit Event Creation'
 
 description: |-
-    All events should be captured and not restricted as this helps in
-    reconstucting the chain-of-events.
+    A maximum of 50 events should be captured to protect against DoS attacks.
 
     To prevent log creation limiting, edit the kubelet configuration
 {{%- if product == "ocp4" %}}
     file <tt>/etc/kubernetes/kubernetes.conf</tt>
     on the kubelet node(s) and set the below parameter:
-    <pre>eventRecordQPS: 0</pre>
+    <pre>kubeAPIQPS: 50</pre>
 {{% else %}}
     file <tt>/etc/origin/node/node-config.yaml</tt>
     on the kubelet node(s) and set the below parameter:
     <pre>kubeletArguments:
       event-qps:
-      - '0'</pre>
+      - '50'</pre>
 {{%- endif %}}
 
 rationale: |-
-    All events should be captured and not restricted as this helps in
-    reconstucting the chain-of-events.
+    A maximum of 50 events should be captured to protect against DoS attacks.
 
 severity: medium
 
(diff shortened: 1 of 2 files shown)


jhrozek:
> @redhatrises so there seem to be inconsistencies between the CIS standard and rule.yml. The YAML keys [kubeAPIQPS v/s eventRecordQPS] are not the same and not sure what's a good value to set for event creation [50 v/s 0].
> 
> /cc @jhrozek @yuumasato

Right, it looks like the current rule.yaml is looking at the setting from the point of "We don't want to lose any events therefore let's not limit the rate of events" while the CIS benchmark text seems to be looking at the control from the point of "We don't want to allow DoS-ing the server, therefore let's limit the rate of events". I can see both points, but the CIS standard seems to be after DoS protection, so the current rule seems to be wrong. What do you think @redhatrises?

lsm5:
alright, so `kubeAPIQPS: 50` works for OCP4. I'm not sure what to use for ocp3, so I just reused what's in the rule.yml.

PTAL all.

lsm5:
I think the title would need to be changed too..

jhrozek:
@redhatrises would appreciate your thoughts on this PR.

redhatrises:
This one is a hard one... as it can be set to 0 or greater based off of how the organization rate limits in their infrastructure which is really how the rule should be written.... so either a) we merge this now as is and make the rule flexible later or b) we make it flexible now. I don't care either way.

lsm5:
> This one is a hard one... as it can be set to 0 or greater based off of how the organization rate limits in their infrastructure which is really how the rule should be written.... so either a) we merge this now as is and make the rule flexible later or b) we make it flexible now. I don't care either way.

option b sounds much better to me. I'll look into it. Thanks!

applications/openshift/kubelet/kubelet_configure_event_creation/rule.yml
```diff
@@ -2,28 +2,26 @@ documentation_complete: true
 
 prodtype: ocp3,ocp4
 
-title: 'kubelet - Do Not Limit Event Creation'
+title: 'kubelet - Limit Event Creation'
```
redhatrises, line 5:
```suggestion
title: 'kubelet - Configure Event Creation Limits'
```

jhrozek:
Hmm, how do I kick the CaC Jenkins job? The patch now seems to be OK in the sense that it is in line with what CIS 4.2.9 prescribes.

redhatrises:
@openscap-ci test this please

jhrozek:
@redhatrises Hi, would you mind checking this PR? Re-reading your earlier comments, it's unclear to me what you meant by making the rule flexible, did you mean we should introduce a variable? What the rule now does is that it adheres strictly to the letter of the CIS benchmark and sets the value to 50 instead of previously used 0.

redhatrises:
> @redhatrises Hi, would you mind checking this PR? Re-reading your earlier comments, it's unclear to me what you meant by making the rule flexible, did you mean we should introduce a variable? What the rule now does is that it adheres strictly to the letter of the CIS benchmark and sets the value to 50 instead of previously used 0.

Yes. A variable needs to be introduced to make this rule more usable to those who not only want to configure the value per their organizational policy and needs. An organization could set it to 0 and perform rate limiting another way.

PR #6227, opened 2020-10-14
fix remediation of audit_rules_privileged_commands

vojtapolasek:
#### Description:

- change command which searches for suid and sgid binaries to ignore remote file systems. The list of file systems is the same as in Openscap (https://github.com/OpenSCAP/openscap/blob/58415789b5a2dd48c0cb0b7cff0fc1dbcb26b4ab/src/OVAL/probes/fsdev.c#L109)

- this is fixed in Bash and Ansible remediations and in tests.

#### Rationale:

- up to now, the command which searched for privileged binaries started in the root file system and did not traverse any other file systems. But there might be setups where for example the /usr folder is on a different file system. In this cases the remediation would not work properly.

- This fixes https://bugzilla.redhat.com/show_bug.cgi?id=1886037

### DIFF
--- a/linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands/ansible/shared.yml
+++ b/linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands/ansible/shared.yml
@@ -4,14 +4,57 @@
 # complexity = low
 # disruption = low
 
-- name: Search for privileged commands
-  shell: find / -not -fstype afs -not -fstype ceph -not -fstype cifs -not -fstype smb3 -not -fstype smbfs -not -fstype sshfs -not -fstype ncpfs -not -fstype ncp -not -fstype nfs -not -fstype nfs4 -not -fstype gfs -not -fstype gfs2 -not -fstype glusterfs -not -fstype gpfs -not -fstype pvfs2 -not -fstype ocfs2 -not -fstype lustre -not -fstype davfs -type f -perm -4000 -o -type f -perm -2000 2>/dev/null
-  args:
-    warn: False
-    executable: /bin/bash
-  check_mode: no
-  register: find_result
-  changed_when: false
+- name: "Configure excluded (non local) file systems"
+  set_fact:
+    excluded_fstypes:
+      - afs
+      - ceph
+      - cifs
+      - smb3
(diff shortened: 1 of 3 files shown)


vojtapolasek:
According to my investigation, the oval and remediations were not aligned. Oval was traversing local file systems, ignoring remote ones. Remediations were considering only single filesystem; the root one. Now remediations behave in the same way as oval, I believe it is correct.

vojtapolasek:
Maybe I was not clear enough. Until this PR, the -xdev option was used. This option prevents the find command to descend other filesystems than the initial one. The remediation uses find command with starting directory of /. When the -xdev option was used and the /usr directory was on a different filesystem, binaries under /usr directory were not identified and no Audit rules were created for them. I consider this an error.
After this PR, the -xdev option is no longer used, but the find command will not descend explicitly mentioned types of filesystems which are considered as remote (NFS, Samba etc). Therefore it will now work correctly even in case the /usr directory is on a different filesystem, of course if the filesystem is not remote.
The OVAL check does exactly the same - it scans all non-remote filesystems for binaries and then checks if an Audit rule exists for each of them. I don't think that wording needs to be changed, because in the description and ocil, the find command is run for each partition separately. No -xdev is mentioned there.
Well, yes, the rule is complex because it is kinda automagical. I am not happy about that but it was here already so I try to keep it sane.

vojtapolasek:
I have extended the Bash remediation to ignore also fuse.sshfs file system. I have also redesigned the Ansible remediation so that it is robust.

linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands/ansible/shared.yml
```diff
@@ -4,14 +4,57 @@
...
+    excludes: excluded_paths
+    file_type: file
+    hidden: yes
+    recurse: yes
+  register: found_files
+
+- name: "construct list of suid or sgid binaries"
+  set_fact:
+    suid_sgid_binaries: "{{ suid_sgid_binaries | union([item.path]) }}"
+  when: item.mode is match("2.*")
```
yuumasato, line 56: @vojtapolasek The `when` condition also needs to match mode SUID mode (`4000`)

yuumasato (approved):

PR #6228, opened 2020-10-14
Fix rule admission_control_plugin_SecurityContextDeny for ocp4

ralvares:
Update rule admission_control_plugin_SecurityContextDeny - CIS 1.2.13 to support ocp4

### DIFF
--- a/applications/openshift/api-server/api_server_admission_control_plugin_SecurityContextDeny/rule.yml
+++ b/applications/openshift/api-server/api_server_admission_control_plugin_SecurityContextDeny/rule.yml
@@ -8,8 +8,12 @@ description: |-
     Instead of using a customized SecurityContext for pods, a Pod Security
     Policy (PSP) should be used. PSP is a cluster-level resource that controls
     the actions that a pod can perform and what resource the pod may access.
-    The <tt>SecurityContextDeny</tt> admission control policy enables PSP. To
-    configure OpenShift to use PSP, edit the API Server pod specification file
+    The <tt>SecurityContextDeny</tt> admission control policy enables PSP.    
+{{%- if product == "ocp4" %}}
+    Ensure that the SecurityContextConstraint plugin is enabled:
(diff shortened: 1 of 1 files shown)


applications/openshift/api-server/api_server_admission_control_plugin_SecurityContextDeny/rule.yml
```diff
@@ -19,22 +23,56 @@ description: |-
...
+
+template:
+  name: yamlfile_value
+  vars:
+    ocp_data: "true"
+    entity_check: "none satisfy"
+    filepath: /api/v1/namespaces/openshift-kube-apiserver/configmaps/config
+    yamlpath: '.data["config.yaml"]'
+    values:
+    - value: '"enable-admission-plugins":\[[^]]*"SecurityContextConstraint"'
```
JAORMX, line 76: This one is a little confusing to me. Isn't this about SecurityContextDeny?

ralvares: If my understanding of the documentation is correct, For 4.6 and above, verify that the set of admission plugins includes SecurityContextConstraint, is that correct ?

redhatrises: This rule is about SecurityContextDeny and not SecurityContextConstraint. SecurityContextConstraint rules are found under `applications/openshift/scc`. If we need to verify that SecurityContextConstraint is enabled, we need a new rule.

ralvares: Fixed!

JAORMX (approved):

jhrozek:
We can either merge the PR now as-is or drop the ocp4 conditionals since they are no longer needed since we dropped ocp3 recently.

JAORMX:
I'll merge this and fix this up in a separate PR.

PR #6229, opened 2020-10-14
rpm_verify_permissions - update shared.sh to handle files managed by more the one RPM and use --restore rather then --setperms

itewk:
#### Description:

If a file, like /var/log/lastlog, is managed by more then one RPM the current script can't handle it.
Also, the man pages for rpm suggest using `--restore` rather then `--setperms` to fix not just perms but ownership.

#### Rationale:

* fixes a bug
* better remediates issues


isimluk (approved):
LGTM.

isimluk:
@redhatrises thoughts?

redhatrises:
@openscap-ci test this please

redhatrises:
`--restore` looks to be the better option as well as replacement for some of these individual changes. We will probably have to create a new rule as well.

PR #6204, opened 2020-10-13
Add yamlfile_value template to api_server_kubelet_certificate_authority.

ggbecker:
#### Description:

- Add yamlfile_value template to `api_server_kubelet_certificate_authority`.

#### Rationale:

- Since OCP4.6 the correct value to check for a certificate authority in
the client configuration is apiServerArguments['kubelet-certificate-authority']
from the config map.

PR #6230, opened 2020-10-14
Enhance regex and template data for api_server_kubelet_certificate_authority

ggbecker:
#### Description:

- Enhance regex and template data for api_server_kubelet_certificate_authority
  - The new regex looks for a specific JSON key and not just the text and it requires to have something set. Spaces are allowed between key and value

#### Rationale

Follow up from: #6204


matejak:
test this please

matejak:
LGTM, as the point is to check that there is *some* certificate authority, the enhancement is good as it is.
I recommend merging if the CI passes.

PR #6233, opened 2020-10-14
Add option quick to build_product script.

ggbecker:
#### Description:

- When this option is enabled, it builds the datastream only. Other
artifacts are not generated, for example guides, tables, etc. Thus
saving time when the datastream is the only thing desired to
generate.

#### Rationale:

- Speed up build time when only the datastream is desired to be built.

- This can also be used in OCP content testing here: https://github.com/ComplianceAsCode/content/blob/master/utils/build_ds_container.sh#L65


build_product
```diff
@@ -250,6 +257,14 @@ set_no_derivatives_options() {
 	fi
 }
 
+set_quick_options() {
+	if test "$_arg_quick" = on ; then
+		for chosen_product in "${_arg_product[@]}"; do
+			QUICK+="generate-ssg-${chosen_product}-ds.xml "
```
matejak, line 263: Usage of `+=` is interesting in this context, and apparently correct. I would consider using arrays - you would need to declare an empty array first.

```suggestion
			QUICK+=( "generate-ssg-${chosen_product}-ds.xml" )
```

build_product
```diff
@@ -324,11 +339,12 @@ done
...
 CMAKE_OPTIONS=(${ADDITIONAL_CMAKE_OPTIONS} "${build_type_option}" "${oval_major_version_option}" "${oval_minor_version_option}" '-DSSG_PRODUCT_DEFAULT=OFF' "${cmake_enable_args[@]}" -G "$cmake_generator")
 set_no_derivatives_options
+set_quick_options
 
 set -e
 rm -rf build/*
 cd build
 cmake .. "${CMAKE_OPTIONS[@]}"
-$build_command "-j${jobs}"
+$build_command "-j${jobs}" ${QUICK}
```
matejak, line 348: With an array, you could use `"${QUICK[@]}"`, which is generally considered better, as it can handle whitespaces.

matejak (changes requested):
I suggest renaming `QUICK` to `EXPLICIT_BUILD_TARGETS` or something like that, so the name represents variable contents.

It is possible to use Bash arrays for this purpose, but that's not necessary, as those targets won't contain namespaces.

build_product
ggbecker: Done

ggbecker:
> I suggest renaming `QUICK` to `EXPLICIT_BUILD_TARGETS` or something like that, so the name represents variable contents.
> 
> It is possible to use Bash arrays for this purpose, but that's not necessary, as those targets won't contain namespaces.

I have addressed your requests, please take a look again.

matejak:
Might be counter-intuitive, but the best array expension would be `"${EXPLICIT_BUILD_TARGETS[@]}"`. Check out https://stackoverflow.com/a/8880633 comment 1.

evgenyz:
I can't help but wonder, why not `--ds` or `--ds-only` or something else with *ds*. It is short and actually describes what would be done.

evgenyz:
I can envision that, because of the blanket-covering nature of the term `quick`, we might introduce `not-so-quick` and `even-faster` options if we would be in a mood to add something to the DS or skip building flavors of the DS.

matejak (approved):

matejak:
I see that the PR has hit the committee :-) I don't have any additional remarks except quoting, so I will leave it to Becker and Evgany.

ggbecker:
> I can envision that, because of the blanket-covering nature of the term `quick`, we might introduce `not-so-quick` and `even-faster` options if we would be in a mood to add something to the DS or skip building flavors of the DS.

I've renamed the option to `datastream-only`

> Might be counter-intuitive, but the best array expension would be `"${EXPLICIT_BUILD_TARGETS[@]}"`. Check out https://stackoverflow.com/a/8880633 comment 1.

Fixed.

evgenyz:
Great!

PR #6234, opened 2020-10-14
[WIP]: A yaml check for whether at least one SCC with allowPrivilegedContainer=false exists

jhrozek:
#### Description:

- so far this is a Wip

#### Rationale:

- We need more checks!


lsm5:
rebase needed probably.

evgenyz (approved):

jhrozek:
Closing. The correct rule that implements CIS 5.2.6 is in PR #6273 
The rule skeleton (ocil and such) as it is in master now is wrong and instead refers to 5.2.1, not 5.2.6

PR #6236, opened 2020-10-14
ocp4: Add check for CIS 1.2.19

JAORMX:
This checks that the insecure-port option is set to 0 or "0" on the
apiserver configuration.

### DIFF
--- a/applications/openshift/api-server/api_server_insecure_port/rule.yml
+++ b/applications/openshift/api-server/api_server_insecure_port/rule.yml
@@ -11,7 +11,7 @@ description: |-
     has not been enabled, the <tt>insecure-port</tt> setting should not exist
 {{%- if product == "ocp4" %}}
     Edit the <tt>openshift-kube-apiserver</tt> configmap on the master node(s)
-    and remove the <tt>insecure-bind-address</tt> if it exists:
+    and remove the <tt>insecure-port</tt> if it exists:
     <pre>
     "apiServerArguments":{
       ...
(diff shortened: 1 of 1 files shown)


applications/openshift/api-server/api_server_insecure_port/rule.yml
```diff
@@ -42,8 +42,27 @@ ocil: |-
...
+# This will check that insecure-port is "0" or 0.
+template:
+  name: yamlfile_value
+  vars:
+    ocp_data: "true"
+    entity_check: "all"
+    filepath: /api/v1/namespaces/openshift-kube-apiserver/configmaps/config
+    yamlpath: '.data["config.yaml"]'
+    values:
+    - value: '"insecure-port":\[["|]0["|]\]'
```
matejak, line 66: What is the intention behind the `["|]`? Would an integer value be really quoted by pipes instead of quotes?

JAORMX: regex format is weird. Sorry for that. The format intended here is as follows: `[<value 1>|<value 2>]` -> `<value 1> or <value 2>`. in this case `["|]` will check for either `"` or nothing. This can match either `"0"` or `0`.

matejak: I am pretty sure that it's not the case - OVAL defines that the PCRE family of regexes should be used, and it can be tested e.g. on https://regex101.com You can see that the pipe inside square brackets is taken literally, but you can express that using `"?`.

JAORMX: @matejak thanks for the review! I'll change it then

JAORMX:
Pull-request updated, HEAD is now 1cd657e18073c1188e8e297221b05e601657755a

yuumasato (approved):

PR #6238, opened 2020-10-14
Add commented section for OCP4 CIS etcd node checks

lbragstad:
Some of the etcd checks in OCP4 are node checks that we can automate,
but we need to add them to the cis-node.profile instead of the
cis.profile.

This commit prepares for those changes by adding the applicable checks
as commented sections in the cis-node.profile. Subsequent patches will
fill in each etcd check.


JAORMX (approved):
LGTM

matejak:
I merge the PR, as it has been approved, and doesn't introduce any functional changes.

PR #6240, opened 2020-10-14
Update api_server_audit_log_path to use different apiserver conf file.

ggbecker:
#### Description:

- Change `api_server_audit_log_path` to check for `openshift-apiserver` api configuration file.
- Create a new rule to check the `openshift-kube-apiserver` namespace.
#### Rationale:

- Currently we are checking only the `openshift-kube-apiserver` but we also need to check the `openshift-apiserver` namespace for some rules.
- It looks like this PR is introducing something new and I don't know exactly we are going to handle these rules, if we are going to use the same group and such.


applications/openshift/api-server/api_server_audit_log_path/rule.yml
```diff
@@ -45,9 +31,23 @@ ocil_clause: '<tt>audit-log-path</tt> does not contain a valid audit file path'
...
+    - general: |-
+        {{{ openshift_cluster_setting("/api/v1/namespaces/openshift-apiserver/configmaps/config") | indent(8) }}}
 {{%- endif %}}
-    The output should return a valid audit log path.
+
+template:
+    name: yamlfile_value
+    vars:
+        ocp_data: "true"
+        filepath: '/api/v1/namespaces/openshift-apiserver/configmaps/config'
```
JAORMX, line 47: I think this rule should check the kube apiserver.

ocp4/profiles/cis.profile
```diff
@@ -60,6 +60,7 @@ selections:
     - api_server_profiling
   # 1.2.22 Ensure that the --audit-log-path argument is set
     - api_server_audit_log_path
+    - kube_api_server_audit_log_path
```
JAORMX, line 63: I think we should instead explicitly create an extra rule that checks  for the openshift apiserver.

Something such as `openshift_apiserver_audit_log_path`. The rest of the rules that start with `api_server` check the kube-apiserver already.

JAORMX (changes requested):
I'm still not sure how to handle rules that should apply to the openshift-apiserver. Maybe we could have a group for those? called `openshift-api-server`

ocp4/profiles/cis.profile
ggbecker: Yeah, I'm going to redo some stuff with these in mind. Thanks

applications/openshift/api-server/api_server_audit_log_path/rule.yml
ggbecker: Got it. I will revert and improve how we are handling these rules.

PR #6241, opened 2020-10-14
OCP4 CIS 5.2.5 (SCC privilege escalation)

nkinder:
#### Description:

- This adds a rule to check if SCC privilege escalation is disabled for at least one SCC.

#### Rationale:

- This is required to satisfy CIS 5.2.5.


jhrozek (approved):
The rule and the check look good. The default OCP4 installation is not compliant at the moment so I put a note to the draft CIS document that we need to follow up and see why even the restricted SCC doesn't prevent privilege escalation.

PR #6243, opened 2020-10-14
OVAL yamlfile_value: apply entity check on the field rather then on record

yuumasato:
#### Description:

- Change the template to apply `entity_check` to the `field`, instead of applying it to the `record`.

#### Rationale:

- It fixes behavior of the template when the `yamlpath` returns a `record` with multiples `field`s.
  E.g.: https://github.com/ComplianceAsCode/content/pull/6234

### DIFF
--- a/applications/openshift/api-server/api_server_insecure_port/rule.yml
+++ b/applications/openshift/api-server/api_server_insecure_port/rule.yml
@@ -11,7 +11,7 @@ description: |-
     has not been enabled, the <tt>insecure-port</tt> setting should not exist
 {{%- if product == "ocp4" %}}
     Edit the <tt>openshift-kube-apiserver</tt> configmap on the master node(s)
-    and remove the <tt>insecure-bind-address</tt> if it exists:
+    and remove the <tt>insecure-port</tt> if it exists:
     <pre>
     "apiServerArguments":{
       ...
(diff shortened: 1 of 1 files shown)


yuumasato:
Actually, from what I checked in the [OVAL schema](https://github.com/OpenSCAP/openscap/blob/maint-1.3/schemas/oval/5.11.2/oval-definitions-schema.xsd), I think both the `record` State and `field` State  can have an `entity_check` attribute:
  The [`EntityObjectRecordType`](https://github.com/OpenSCAP/openscap/blob/maint-1.3/schemas/oval/5.11.2/oval-definitions-schema.xsd#L1524) is an [`EntityBaseComplexType`](https://github.com/OpenSCAP/openscap/blob/maint-1.3/schemas/oval/5.11.2/oval-definitions-schema.xsd#L1385) ; and the [`EntityObjectFieldType`](https://github.com/OpenSCAP/openscap/blob/maint-1.3/schemas/oval/5.11.2/oval-definitions-schema.xsd#L1545) is an `anySimpleType`+`entity_check`.
  The [`EntityStateRecordType`](https://github.com/OpenSCAP/openscap/blob/maint-1.3/schemas/oval/5.11.2/oval-definitions-schema.xsd#L1795) is an [`EntityStateComplexBaseType`](https://github.com/OpenSCAP/openscap/blob/maint-1.3/schemas/oval/5.11.2/oval-definitions-schema.xsd#L1584); and the [`EntityStateFieldType`](https://github.com/OpenSCAP/openscap/blob/maint-1.3/schemas/oval/5.11.2/oval-definitions-schema.xsd#L1816) is an `anySimpleType`+`entity_check`. 

In the end we may need both:
-  `entity_check` in the record, for yamlpaths that return more than one record
- `entity_check` in the field, for records that return more than one field

PR #6244, opened 2020-10-14
OCP4 CIS 5.2.3

nkinder:
#### Description:

- This adds a rule to check if host IPC access is disabled for at least one SCC.

#### Rationale:

- This is required to satisfy CIS 5.2.3.


jhrozek (approved):
LGTM

PR #6247, opened 2020-10-14
ocp4 - CIS 1.2.21: Add relevant information for ocp4

JAORMX:
This check doesn't really apply to ocp4 due to the defaults we use. So
let's just make it informational.

### DIFF
--- a/applications/openshift/api-server/api_server_profiling/rule.yml
+++ b/applications/openshift/api-server/api_server_profiling/rule.yml
@@ -2,22 +2,17 @@ documentation_complete: true
 
 prodtype: ocp3,ocp4
 
+{{%- if product == "ocp4" %}}
+title: 'OpenShift Profiling considerations'
+{{% else %}}
 title: 'Disable Profiling on the API server'
+{{%- endif %}}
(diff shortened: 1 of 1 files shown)


applications/openshift/api-server/api_server_profiling/rule.yml
```diff
@@ -28,24 +23,37 @@ description: |-
 {{%- endif %}}
 
 rationale: |-
+{{%- if product == "ocp4" %}}
+    Profiling is enabled by default in OpenShift. The API server operators exposes
+    Prometheus metrics via the metrics service.  Profiling data is sent to <tt>healthzPort</tt>,
+    the port of the localhost healthz endpoint. Changing this value may disrupt components
+    that monitor the kubelet health. The default port value is <tt>10248</tt>, and the healthz
+
+    BindAddress is <tt>127.0.0.1</tt>.
```
redhatrises, line 32: lines 27-32 can go in the description.

JAORMX:
@redhatrises so, we shouldn't even put the info here then?

PR #6245, opened 2020-10-14
Cis 5.2.7

lsm5:
#### Description:

- Add template to scc_limit_net_raw_capability to check if NET_RAW or ALL are dropped in at least one SCC

#### Rationale:

- Critical to drop dangerous capabilities

NOTE: We aren't yet compliant with this.

@redhatrises @jhrozek @yuumasato PTAL

PR #6250, opened 2020-10-14
Cis 5.2.9

lsm5:
#### Description:

- Adds a template to scc/scc_limit_net_raw_capability/rule.yml to check if atleast one SCC limits containers which have all capabilities set.

#### Rationale:

- At least one SCC should drop "ALL" capabilities.

NOTE: non-compliant as well like CIS 5.2.7

### DIFF
--- a/applications/openshift/scc/scc_drop_container_capabilities/rule.yml
+++ b/applications/openshift/scc/scc_drop_container_capabilities/rule.yml
@@ -31,3 +31,23 @@ ocil: |-
     Review each SCC and determine that all capabilities are either
     completely disabled as a list entry under <tt>requiredDropCapabilities</tt>,
     or that all the un-required capabilities are dropped for containers and SCCs.
+
+{{%- if product == "ocp4" %}}
+warnings:
+    - general: |-
+        {{{ openshift_cluster_setting("/apis/security.openshift.io/v1/securitycontextconstraints") | indent(8) }}}
+{{%- endif %}}
+
+template:
+{{%- if product == "ocp4" %}}
+    name: yamlfile_value
+    vars:
+        ocp_data: "true"
+        filepath: /apis/security.openshift.io/v1/securitycontextconstraints
+        yamlpath: ".items[:]['requiredDropCapabilities']"
+        check_existence: "at_least_one_exists"
+        entity_check: "at least one"
+        values:
+          - type: "string"
+            value: "ALL"
+{{%- endif %}}
--- a/applications/openshift/scc/scc_drop_container_capabilities/rule.yml
+++ b/applications/openshift/scc/scc_drop_container_capabilities/rule.yml
@@ -44,10 +44,12 @@ template:
     vars:
         ocp_data: "true"
         filepath: /apis/security.openshift.io/v1/securitycontextconstraints
-        yamlpath: ".items[:]['requiredDropCapabilities']"
+        yamlpath: ".items[:]['requiredDropCapabilities','#'][:]"
         check_existence: "at_least_one_exists"
         entity_check: "at least one"
         values:
-          - type: "string"
-            value: "ALL"
+          - key: "requiredDropCapabilities"
+            value: "^ALL$"
+            entity_check: "at least one"
+            operation: "pattern match"
 {{%- endif %}}


lsm5:
builds on https://github.com/ComplianceAsCode/content/pull/6245

lsm5:
need to change this one..

lsm5:
builds on #6245 and #6261

applications/openshift/scc/scc_limit_net_raw_capability/rule.yml
```diff
@@ -27,7 +27,26 @@ ocil_clause: 'NET_RAW does not exist or is configured in defaultAddCapabilities
...
+    vars:
+        ocp_data: "true"
+        filepath: /apis/security.openshift.io/v1/securitycontextconstraints
+        yamlpath: ".items[:]['requiredDropCapabilities']"
+        check_existence: "at_least_one_exists"
+        entity_check: "at least one"
+        values:
+          - type: "string"
+            value: "NET_RAW"
+            value: "ALL"
```
jhrozek, line 52: I'm unsure if the multiple values would work like this..but let's figure out the yamlpath first

applications/openshift/scc/scc_limit_net_raw_capability/rule.yml
```diff
@@ -27,7 +27,26 @@ ocil_clause: 'NET_RAW does not exist or is configured in defaultAddCapabilities
...
+    - general: |-
+        {{{ openshift_cluster_setting("/apis/security.openshift.io/v1/securitycontextconstraints") | indent(8) }}}
+{{%- endif %}}
+
+template:
+    name: yamlfile_value
+    vars:
+        ocp_data: "true"
+        filepath: /apis/security.openshift.io/v1/securitycontextconstraints
+        yamlpath: ".items[:]['requiredDropCapabilities']"
```
jhrozek, line 46: Hi @evgenyz can you help us out here? We have a list of SCCs and then the `requiredDropCapabilities` is a list, too. The `requiredDropCapabilities` attributes is a list itself, too, it might look like:
```
  requiredDropCapabilities:
  - KILL
  - MKNOD
  - SETUID
  - SETGID
```
So I tried to adjust the YAML-path like this:
```
".items[:]['requiredDropCapabilities'][:]"
```
but this causes an error in the scanner:
```
I: oscap:     Opening file '/kubernetes-api-resources/apis/security.openshift.io/v1/securitycontextconstraints'.
I: oscap:     YAML path '.items[:]['requiredDropCapabilities'][:]' points to a multi-dimensional structure (sequence containing another sequence)
I: oscap:     Test 'oval:ssg-test_scc_drop_container_capabilities:tst:1' requires that at least one object defined by 'oval:ssg-object_scc_drop_container_capabilities:obj:1' exists on the system.
I: oscap:     0 objects defined by 'oval:ssg-
```
Can you suggest if we can do anything?

jhrozek (commented):
Looks like we might have an issue with the multi-dimensional array, let's see what idea does @evgenyz have

applications/openshift/scc/scc_limit_net_raw_capability/rule.yml
evgenyz: As stupid as it might sound, this would help: `.items[:]['requiredDropCapabilities','#'][:]`. And even `.items[:]['requiredDropCapabilities','#']` might be enough, as ending `[:]` is usually redundant.

evgenyz: Result would be: 
```xml
           <ind-sys:value datatype="record">
              <field name="required^drop^capabilities">KILL</field>
              <field name="required^drop^capabilities">MKNOD</field>
            </ind-sys:value>
            <ind-sys:value datatype="record">
              <field name="required^drop^capabilities">SETGUID</field>
            </ind-sys:value>
            ...
```

evgenyz: This '#' is kind of a bug. Or an edge case. But right now, if you want an array behid just one key you have to pretend that you're interested in multiple keys (i.e. keys selection path section type). Of course this would work for multiple keys as well, '#' is just a dummy key.

evgenyz: To be precise, if you want an array, behind single key, which is also in array and you're collecting all of them ([:]['key','dummy-key'][:]). Array index, for example, would work without this hack ([0]['key'][:]).

evgenyz: Also, `type='string'` is redundant, as default type is always string.

applications/openshift/scc/scc_limit_net_raw_capability/rule.yml
evgenyz: This should work
```
values:
   - value: "NET_RAW|ALL"
     operation: "pattern match"
```

evgenyz: Hold on a second. What actually was meant here? Array should contain both or having any of them is OK?

jhrozek: Either. The attribute describes capabilities that are dropped. At a minimum we want NEW_RAW to be dropped specifically, but if anyone drops ALL, that's fine, too.

evgenyz: Then yes, this would work:
```
values:
   - value: "^NET_RAW|ALL$"
     operation: "pattern match"
```

jhrozek:
I rebased the patch atop origin/master and added fixups that were both possible and needed since PR#6278

PR #6256, opened 2020-10-15
add rule use_pam_wheel_for_su

vojtapolasek:
#### Description:

- add a rule which enforces usage of pam_wheel module for su authentication

- rule, oval, bash, ansible, tests

- add rule to rhel8 ospp

#### Rationale:

- rhel CC effort


linux_os/guide/system/accounts/accounts-restrictions/root_logins/use_pam_wheel_for_su/bash/shared.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+# platform = multi_platform_rhel,multi_platform_fedora,multi_platform_ol,multi_platform_rhv
+
+if ! grep -q pam_wheel /etc/pam.d/su; then
```
mildas, line 4: This does not do what you expect - I tried this condition and it got evaluated as true when there was no `pam_wheel` and evaluated as false when `pam_wheel` was present in `/etc/pam.d/su`.

mildas:
The bash remediation script adds the required line only when a line with `pam_wheel.so` is present so it does not fix a case when there is no `pam_wheel` line.

However, I see that you have created test scenario that removes all `auth required pam_wheel.so use_uid` lines. This scenario does not hit the issue I have pointed out because the default `/etc/pam.d/su` file has `#auth       sufficient  pam_wheel.so trust use_uid ` like there.

linux_os/guide/system/accounts/accounts-restrictions/root_logins/use_pam_wheel_for_su/bash/shared.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+# platform = multi_platform_rhel,multi_platform_fedora,multi_platform_ol,multi_platform_rhv
+
+# uncomment the option if commented
+  sed '/^[[:space:]]*#[[:space:]]*auth[[:space:]]\+required[[:space:]]\+pam_wheel\.so[[:space:]]\+use_uid$/s/^#//' -i /etc/pam.d/su
```
yuumasato, line 5: The sed command won't uncomment the line when the line has spaces before '#'.

yuumasato (changes requested):
> The bash remediation script adds the required line only when a line with `pam_wheel.so` is present so it does not fix a case when there is no `pam_wheel` line.

Actually, I think we shouldn't remediate cases where the commented line is not there as we don't know where in the stack to put the  config  for `pam_wheel.so`.

linux_os/guide/system/accounts/accounts-restrictions/root_logins/use_pam_wheel_for_su/bash/shared.sh
```diff
@@ -0,0 +1,9 @@
...
+
+if ! grep -q '^[\s]*auth[\s]+required[\s]+pam_wheel\.so\[s]+use_uid$' /etc/pam.d/su; then
+  echo "auth             required        pam_wheel.so use_uid" >> /etc/pam.d/su
```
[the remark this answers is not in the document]
mildas: I didn't know that and it's important information for reviewing pam rules. Thanks

yuumasato (approved):

PR #6262, opened 2020-10-15
Add rule for OCP4 CIS 1.3.2

mrogers950:
Adds a check for CIS 1.3.2 to demonstrate intentional protection of the
profiling endpoints with an RBAC role, since profiling cannot be disabled.


jhrozek (approved):
looks good and works fine.

PR #6267, opened 2020-10-15
Add option to enable filepath regex in file permissions template.

ggbecker:
#### Description:

- Add option to enable filepath regex in file permissions template. Also adds the same option to groupowner and owner templates.

#### Rationale:

- Enable template to check multiple files with same pattern but from different directories.


JAORMX (approved):
works like a charm

jan-cerny:
The generated OVAL looks OK.

PR #6272, opened 2020-10-15
OCP4 - CIS 1.1.3 Add check

JAORMX:
This checks for the file permissions of the kube-controller-manager
static pod definition.


JAORMX:
Pull-request updated, HEAD is now 8ff5f84cd448e4293942feb42062a03c133f29d5

PR #6276, opened 2020-10-19
Create custom OVAL check for uefi_no_removeable_media

### LINKED ISSUE #6028: Rule uefi_no_removeable_media fails even on non-UEFI systems
#### Description of problem:
Rule `uefi_no_removeable_media` fails even on non-UEFI systems, rule check should be made more robust to first verify if the system uses UEFI and only then to check the configuration.

master

RHEL7

#### Steps to Reproduce:
Run the following on non-UEFI system:
`oscap xccdf eval --profile '(all)' --rule xccdf_org.ssgproject.content_rule_uefi_no_removeable_media ssg-rhel7-ds.xml`
and the rule will fail even though it should rather be not applicable or pass.

ggbecker:
#### Description:

- Create custom OVAL check for uefi_no_removeable_media.
- Add the extended definition [system_boot_mode_is_uefi](https://github.com/ComplianceAsCode/content/blob/master/shared/checks/oval/system_boot_mode_is_uefi.xml) to check if the system is UEFI or not

#### Rationale:

- With current approach CPE applicability is implemented, we cannot easily assign multiple platform checks, and since other rules ([file_permissions_efi_grub2_cfg](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/bootloader-grub2/file_permissions_efi_grub2_cfg/oval/shared.xml#L5)) are already using this approach to make the rule pass on systems that are not UEFI we can keep this way until better support for CPE applicability language is supported.

- Fixes #6028

### DIFF
--- a/linux_os/guide/system/bootloader-grub2/uefi_no_removeable_media/oval/shared.xml
+++ b/linux_os/guide/system/bootloader-grub2/uefi_no_removeable_media/oval/shared.xml
@@ -6,7 +6,7 @@
 
 <def-group>
   <definition class="compliance" id="uefi_no_removeable_media" version="1">
-    {{{ oval_metadata("Ensure 'set root' is configured with value '(?!fd)(?!cd)(?!usb).*' in /boot/efi/EFI/redhat/grub.cfg") }}}
+    {{{ oval_metadata("Ensure the system is not configured to use a boot loader on removable media.") }}}
     <criteria comment="The respective application or service is configured correctly or system boot mode is not UEFI" operator="OR">
       <extend_definition definition_ref="system_boot_mode_is_uefi" negate="true" comment="Pass if system boot mode is not UEFI" />
       <criterion comment="Check the set root in {{{ grub_cfg_prefix + "/grub.cfg" }}}" test_ref="test_uefi_no_removeable_media" />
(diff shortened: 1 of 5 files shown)


linux_os/guide/system/bootloader-grub2/uefi_no_removeable_media/oval/shared.xml
```diff
@@ -1 +1,35 @@
-{{{ oval_check_config_file(path='/boot/efi/EFI/redhat/grub.cfg', prefix_regex='^[ \\t]*', parameter='set root', separator_regex='=', value="'(?!fd)(?!cd)(?!usb).*'", missing_parameter_pass=false, missing_config_file_fail=true) }}}
+{{% if product == "fedora" %}}
+{{% set grub_cfg_prefix = "/boot/efi/EFI/fedora" %}}
+{{% else %}}
+{{% set grub_cfg_prefix = "/boot/efi/EFI/redhat" %}}
+{{% endif %}}
+
+<def-group>
+  <definition class="compliance" id="uefi_no_removeable_media" version="1">
+    {{{ oval_metadata("Ensure 'set root' is configured with value '(?!fd)(?!cd)(?!usb).*' in /boot/efi/EFI/redhat/grub.cfg") }}}
```
redhatrises, line 9:
```suggestion
    {{{ oval_metadata("Ensure the system is not configured to use a boot loader on removable media.") }}}
```

yuumasato (approved):
LGTM

PR #6277, opened 2020-10-19
Disable OVAL backend from file_permissions grub2_cfg rules

ggbecker:
#### Description:

- Disable OVAL template for file_permissions grub2_cfg rules and fix a typo in installed_env_has_grub2_package.xml

#### Rationale:

- There is not need to let OVAL template enabled for these rules since they have custom [shared1](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/bootloader-grub2/file_permissions_efi_grub2_cfg/oval/shared.xml) [shared2](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/bootloader-grub2/file_permissions_grub2_cfg/oval/shared.xml) OVAL

### DIFF
--- a/docs/manual/developer_guide.adoc
+++ b/docs/manual/developer_guide.adoc
@@ -1773,7 +1773,7 @@ yamlfile_value::
 ** *yamlpath* - OVAL's link:https://github.com/OpenSCAP/yaml-filter/wiki/YAML-Path-Definition[YAML Path] expression.
 ** *entity_check* (link:https://github.com/OVALProject/Language/blob/master/docs/oval-common-schema.md#CheckEnumeration[CheckEnumeration]) - entity_check value for state's value_of, optional. If omitted, entity_check attribute would not be set and will be treated by OVAL as 'all'. +
     Possible options are `all`, `at least one`, `none satisfy` and `only one`.
-** *check_existence* (link:https://github.com/OVALProject/Language/blob/master/docs/oval-common-schema.md#ExistenceEnumeration[ExistenceEnumeration]) - `check_existence` value for the `yamlfilecontent_test`, optional. If omitted, check_existence attribute will defalut to 'only_one_exists'. +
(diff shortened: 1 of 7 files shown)


vojtapolasek:
Thank you for the fix, I have few questions:
1. In file_permissions_efi_grub2_cfg  there is a conditional in the rule description. If the product is Fedora, the location of the file is different than for other products. This is, however, not reflected in OVAL check nor in the Bash remediation. Can you please fix it? I wonder if the template makes sense at all here because the same problem will be with Ansible remediation.
2. Can you please add tests to file_permissions_grub2_cfg? It is not using template oval and it should be easy to test.
3. How about similar rules for owner and group of Grub files? They also have their custom ovals but are effectively using templated one. Can you disable it for them as well?

ggbecker:
> Thank you for the fix, I have few questions:
> 
>     1. In file_permissions_efi_grub2_cfg  there is a conditional in the rule description. If the product is Fedora, the location of the file is different than for other products. This is, however, not reflected in OVAL check nor in the Bash remediation. Can you please fix it? I wonder if the template makes sense at all here because the same problem will be with Ansible remediation.
> 
The OVAL check uses `^/boot/efi/EFI/(redhat|fedora)/grub.cfg$` which should cover Fedora as well.
The existing remediation is applicable only to RHEL and OL, but I'll fix the remediation to cover Fedora as well using template tricks, I know a nice solution.

>     2. Can you please add tests to file_permissions_grub2_cfg? It is not using template oval and it should be easy to test.
> 
I'll see what I can do.
>     3. How about similar rules for owner and group of Grub files? They also have their custom ovals but are effectively using templated one. Can you disable it for them as well?

They are not using the templated OVAL because the custom `shared.xml` always takes precedence, but I thought it would be nice to explicitly set that we don't want to use OVAL from the template... to make it more obvious.

ggbecker:
I took a different approach instead, see: #6286 
This may become obsolete

PR #6279, opened 2020-10-20
fix conflicts of audit rules for privileged commands

### LINKED ISSUE #5984: Rule audit_rules_privileged_commands fails after remediation (RHEL7 STIG)
#### Description of problem:
When remediating `stig` profile on RHEL7 or `ism_o` profile RHEL8 the rule `audit_rules_privileged_commands` is reported as fixed, but an additional scan reports it as failed. This is most likely caused by the following remediations because the rule is reported as fixed during remediation (related to [RHBZ#1750755](https://bugzilla.redhat.com/show_bug.cgi?id=1750755)).

master

RHEL7 (stig), RHEL8 (ism_o)

vojtapolasek:
#### Description:

- modify audit_rules_privileged_commands to unify with templated rules audit_rules_privileged_commands_*

    - remove checking for perm=x

    - remove remediationg of perm=x

    - unify checks and remediations to use -F key=privileged

#### Rationale:

The "automagic" rule audit_rules_privileged_commands and individual templated rules audit_rules_privileged_commands_*  were not cooperating together, although they are are often selected in one profile. It resulted in one of rules failing although the configuration was actually correct. It also resulted in duplicate Audit rule entries.

- fixes #5984

### DIFF
--- a/shared/bash_remediation_functions/perform_audit_rules_privileged_commands_remediation.sh
+++ b/shared/bash_remediation_functions/perform_audit_rules_privileged_commands_remediation.sh
@@ -103,13 +103,12 @@ do
 		base_search=$(sed -e '/-a always,exit/!d' -e '/-F path='"${sbinary_esc}"'[^[:graph:]]/!d'		\
 				-e '/-F path=[^[:space:]]\+/!d'						\
 				-e '/-F auid>='"${min_auid}"'/!d' -e '/-F auid!=\(4294967295\|unset\)/!d'	\
+				-e '/-F perm=[rwxa]\+/d' \
 				-e '/-k \|-F key=/!d' "$afile")
 
 		# Increase the count of inspected files for this sbinary
 		count_of_inspected_files=$((count_of_inspected_files + 1))
 
-		# Require execute access type to be set for existing audit rule
-		exec_access='x'
 
 		# Search current audit rules file's content for presence of rule pattern for this sbinary
 		if [[ $base_search ]]
--- a/shared/bash_remediation_functions/perform_audit_rules_privileged_commands_remediation.sh
+++ b/shared/bash_remediation_functions/perform_audit_rules_privileged_commands_remediation.sh
@@ -79,7 +79,7 @@ do
 	local count_of_inspected_files=0
 
 	# Define expected rule form for this binary
-	expected_rule="-a always,exit -F path=${sbinary} -F perm=x -F auid>=${min_auid} -F auid!=unset -k privileged"
+	expected_rule="-a always,exit -F path=${sbinary} -F auid>=${min_auid} -F auid!=unset -F key=privileged"
 
 	# If list of audit rules files to be inspected is empty, just add new rule and move on to next binary
 	if [[ ${#files_to_inspect[@]} -eq 0 ]]; then
@@ -101,7 +101,7 @@ do
 		#   them in arbitrary order)
 	
 		base_search=$(sed -e '/-a always,exit/!d' -e '/-F path='"${sbinary_esc}"'[^[:graph:]]/!d'		\
-				-e '/-F path=[^[:space:]]\+/!d'   -e '/-F perm=.*/!d'						\
+				-e '/-F path=[^[:space:]]\+/!d'						\
 				-e '/-F auid>='"${min_auid}"'/!d' -e '/-F auid!=\(4294967295\|unset\)/!d'	\
 				-e '/-k \|-F key=/!d' "$afile")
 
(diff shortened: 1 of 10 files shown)


yuumasato:
@vojtapolasek So this is a regression after https://github.com/ComplianceAsCode/content/pull/5956?

vojtapolasek:
> @vojtapolasek So this is a regression after #5956?

Only partially. Yes the removal of perm=x was not introduced into the automagic rule, that's true. But additionally there was a problem that the templated rule and the automagic rule used different rule keys.

yuumasato (changes requested):
Test scenarios are failing because the script that generates the correct solution has not been updated.
https://github.com/ComplianceAsCode/content/blob/e5a65a1de7eda39e237401417196e87fbc6ea840/linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands/tests/generate_privileged_commands_rule.sh#L7

And I suggest to add a new `fail` test scenario with audit rules with `-F perm=x`. To assure it removes the `perm` correctly.
In extension of previous PR, I suggest adding similar same test scenario for `audit_rules_privileged_commands_sudo`.

vojtapolasek:
You are right, I forgot to update tests. I fixed it and this led to discovering of a typo in oval. I also updated rule description. I will do separate update for templated audit_rules_privileged_commands rules.

shared/bash_remediation_functions/perform_audit_rules_privileged_commands_remediation.sh
```diff
@@ -103,13 +103,12 @@ do
 		base_search=$(sed -e '/-a always,exit/!d' -e '/-F path='"${sbinary_esc}"'[^[:graph:]]/!d'		\
 				-e '/-F path=[^[:space:]]\+/!d'						\
 				-e '/-F auid>='"${min_auid}"'/!d' -e '/-F auid!=\(4294967295\|unset\)/!d'	\
+				-e '/-F perm=[rwxa]\+/d' \
```
yuumasato, line 106: @vojtapolasek Could you clarify what issue this is fixing?

yuumasato: The `base_search` still needs to look for audit rules setting up a specific `-F perm=`.
But then, instead of removing the code block below (former lines 130 to 154) the block should be changed to remove the `perm` filter.

Test scenario `augenrules_rules_with_perm_x.fail.sh`  scenario is passing, but in the end it duplicates the audit rules:
```
-a always,exit -F path=/usr/bin/passwd -F perm=x -F auid>=1000 -F auid!=unset -k privileged
...
-a always,exit -F path=/usr/bin/passwd -F auid>=1000 -F auid!=unset -F key=privileged
```

shared/bash_remediation_functions/perform_audit_rules_privileged_commands_remediation.sh
```diff
@@ -127,31 +126,6 @@ do
...
-			then
-
-				# If not, append the 'x' (execute) permission to the existing access type bits
-				access_type="$access_type$exec_access"
-				# Reconstruct the permissions section for the rule
-				new_rule_perm="-F perm=$access_type"
-				# Update existing rule in current audit rules file with the new permission section
-				sed -i "s#${rule_head}\(.*\)${rule_tail}#${rule_head}${new_rule_perm}${rule_tail}#" "$afile"
-
-			fi
```
yuumasato, line 129: Instead of being removed, this block needs to be changed to remove any `-F perm=`.

yuumasato (changes requested):

vojtapolasek:
I modified the remediation. The base_search does not pick rules with perm=x explicitly, I don't think it makes sense here as they are covered by rest of regexes. I modified the remediation so that it removes the perm flag as requested.

yuumasato (approved):
Thank for addressing he comments, looks good to me.
I found a few typo issues, though.

shared/bash_remediation_functions/perform_audit_rules_privileged_commands_remediation.sh
```diff
@@ -122,10 +122,10 @@ do
 			readarray -t handled_sbinaries < <(grep -o -e "-F path=[^[:space:]]\+" <<< "$concrete_rule")
 			handled_sbinaries=("${handled_sbinaries[@]//-F path=/}")
 
-		# 		Merge the list of such SUID/SGID binaries found in this iteration with global list ignoring duplicates
+			# 		Merge the list of such SUID/SGID binaries found in this iteration with global list ignoring duplicates
```
yuumasato, line 125:
```suggestion
			# Merge the list of such SUID/SGID binaries found in this iteration with global list ignoring duplicates
```

PR #6280, opened 2020-10-20
fix srgs for new rhel8 stig rules

### LINKED ISSUE #6275: Rule package_chrony_installed is missing SRG reference
#### Description of problem:
Rule `package_chrony_installed` got selected into the OSPP profile by the https://github.com/ComplianceAsCode/content/pull/6181 which means it is also selected in STIG profile, but it is missing SRG reference.

master

RHEL8

vojtapolasek:
#### Description:

- adds SRG to package_chrony_installed and kernel_trust_cpu_rng rules

#### Rationale:

- these rules were added to rhel8 ospp and because stig extends ospp, they appear also in stig. However, srgs were not added.

- fixes #6275

### DIFF
--- a/linux_os/guide/services/ntp/package_chrony_installed/rule.yml
+++ b/linux_os/guide/services/ntp/package_chrony_installed/rule.yml
@@ -26,7 +26,7 @@ references:
     cis@rhel8: 2.2.1.1
     ism: 0988,1405
     ospp: FMT_SMF_EXT.1
-    srg: SRG-OS-000355-GPOS-00143,SRG-OS-000356-GPOS-00144
+    srg: SRG-OS-000355-GPOS-00143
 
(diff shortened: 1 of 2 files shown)


linux_os/guide/services/ntp/package_chrony_installed/rule.yml
```diff
@@ -25,7 +25,8 @@ references:
     cis@rhel7: 2.2.1.1
     cis@rhel8: 2.2.1.1
     ism: 0988,1405
-    ospp: package_audispd-plugins_installed
+    ospp: FMT_SMF_EXT.1
+    srg: SRG-OS-000355-GPOS-00143,SRG-OS-000356-GPOS-00144
```
redhatrises, line 29:
```suggestion
    srg: SRG-OS-000355-GPOS-00143
```

PR #6285, opened 2020-10-21
allow install_vm.py to create UEFI based machines

vojtapolasek:
#### Description:

- add a new parameter to install_vm.py:

    - --uefi normal - will create UEFI based machine without secureboot

    - --uefi=secureboot will create UEFI based machine with secure boot

- kickstart is modified during this action, the  size of /boot partition is decreased and there is created extra /boot/efi partition

- this feature is mentioned in the documentation

- you need e2k-ovmf package to use this feature.

- Note that such machines can't be used with the test suite directly, because Libvirt does not support internal snapshots of UEFI machines

#### Rationale:

- this will help to test rules dealing with UEFI


yuumasato (approved):
Thanks for the addition!
Works for me, :)

PR #6288, opened 2020-10-22
Specify the applicable OpenShift version for the CIS profiles

JAORMX:
This helps establish expectations about the profiles.

### DIFF
--- a/ocp4/profiles/cis.profile
+++ b/ocp4/profiles/cis.profile
@@ -13,6 +13,8 @@ description: |-
 
     Note that this part of the profile is meant to run on the Platform that
     Red Hat OpenShift Container Platform 4 runs on top of.
+
+    This profile is applicable to OpenShift versions 4.6 and greater.
 selections:
   ### 1 Control Plane Components
   ###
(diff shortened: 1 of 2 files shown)


ocp4/profiles/cis.profile
```diff
@@ -13,6 +13,8 @@ description: |-
 
     Note that this part of the profile is meant to run on the Platform that
     Red Hat OpenShift Container Platform 4 runs on top of.
+
+    The applicable version of OpenShift that this profile addresses is 4.6.
```
redhatrises, line 17: We've traditionally stayed away from minor version tracking upstream since it not only adds a large maintenance burden, but usually, downstream is responsible for maintaining the minor version.... however, that is in a traditional RHEL sense. Also, upstream traditionally lags behind or way ahead of a downstream release. Does it make sense to handle minor versions?

JAORMX: uhm... Maybe what we really want to explain to folks is that this profiles are for 4.6+ and won't be applicable for 4.5 and lower releases.

redhatrises: okay. That makes sense. How about something like:

`This profile is applicable to OpenShift versions 4.6 and greater.`

JAORMX:
Pull-request updated, HEAD is now b99e175877b44e11bffb51fb0df21bbf909069b6

PR #6290, opened 2020-10-22
Parametrize rule for login.defs hashing algorithm

yuumasato:
#### Description:

- Parametrize rule `set_password_hashing_algorithm_logindefs`.
  - Allow choice between `SHA512` and `SHA256`.
- Add simple tests.

#### Rationale:

- More flexibility for the rule.

### DIFF
--- a/linux_os/guide/system/accounts/accounts-pam/var_password_hashing_algorithm.var
+++ b/linux_os/guide/system/accounts/accounts-pam/var_password_hashing_algorithm.var
@@ -1,0 +1,18 @@
+documentation_complete: true
+
+title: Password Hashing algorithm
+
+description: |-
+    Specify the system default encryption algorithm for encrypting passwords.
+    Defines the value set as ENCRYPT_METHOD in /etc/login.defs.
+
+type: string
+
+operator: equals
+
+interactive: false
+
+options:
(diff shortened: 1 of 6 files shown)


linux_os/guide/system/accounts/accounts-pam/var_password_hashing_algorithm.var
```diff
@@ -0,0 +1,17 @@
+documentation_complete: true
+
+title: Password Hassing algorithm
+
+description: |-
+    Specify the hashing algorithm.
```
redhatrises, line 6: Can you flush this description out a bit more?

linux_os/guide/system/accounts/accounts-pam/var_password_hashing_algorithm.var
```diff
@@ -0,0 +1,17 @@
+documentation_complete: true
+
+title: Password Hassing algorithm
```
mildas, line 3: `title: Password Hashing Algorithm`

yuumasato: Should be fixed, thanks.

mildas:
LGTM

PR #6293, opened 2020-10-23
Fixed the sshd_set_idle_timeout rule

matejak:
The configuration works if and only if `ClientAliveCountMax` is set to 0.

Therefore, the check and remediations for this rule take both configuration items into account.
The rule description has been clarified to warn from situations when the timeout isn't effective.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1859551


linux_os/guide/services/ssh/ssh_server/sshd_set_idle_timeout/bash/shared.sh
```diff
@@ -1,5 +1,8 @@
-# platform = multi_platform_wrlinux,multi_platform_rhel,multi_platform_fedora,multi_platform_ol,multi_platform_rhv,multi_platform_sle
+# platform = multi_platform_all
+
 . /usr/share/scap-security-guide/remediation_functions
+
 {{{ bash_instantiate_variables("sshd_idle_timeout_value") }}}
 
-replace_or_append '/etc/ssh/sshd_config' '^ClientAliveInterval' $sshd_idle_timeout_value '@CCENUM@' '%s %s'
+{{{ bash_sshd_config_set("ClientAliveInterval", "$sshd_idle_timeout_value") }}}
+{{{ bash_sshd_config_set("ClientAliveCountMax", "0") }}}
```
redhatrises, line 8: ClientAliveCountMax needs to maintain itself in a new rule. One configuration change per rule only.

linux_os/guide/services/ssh/ssh_server/sshd_set_idle_timeout/rule.yml
```diff
@@ -9,6 +9,7 @@ description: |-
     To set an idle timeout interval, edit the following line in <tt>/etc/ssh/sshd_config</tt> as
     follows:
     <pre>ClientAliveInterval <b>{{{ xccdf_value("sshd_idle_timeout_value") }}}</b></pre>
+    while making sure that <tt>ClientAliveCountMax</tt> is set to <tt>0</tt>.
```
redhatrises, line 12: ClientAliveCountMax needs to maintain itself in a new rule. One configuration change per rule only.

redhatrises (changes requested):
ClientAliveCountMax needs to maintain itself in a new rule. One configuration change per rule only. This is a requirement.

linux_os/guide/services/ssh/ssh_server/sshd_set_idle_timeout/rule.yml
```diff
@@ -67,6 +68,9 @@ ocil: |-
     <pre>ClientAliveInterval {{{ xccdf_value("sshd_idle_timeout_value") }}}</pre>
 
 warnings:
-    - dependency: |-
-        SSH disconnecting idle clients will not have desired effect without also
-        configuring ClientAliveCountMax in the SSH service configuration.
+    - general: |-
```
redhatrises, line 71: The original warning still applies. This warning can be appended to it.

linux_os/guide/services/ssh/ssh_server/sshd_set_idle_timeout/rule.yml
```diff
@@ -67,6 +68,9 @@ ocil: |-
...
+        Following conditions may prevent the SSH session to time out:
```
redhatrises, line 72:
```suggestion
      Some processes may stop SSH from correctly detecting that the user is idle. The following conditions may prevent the SSH session to time out:
```

linux_os/guide/services/ssh/ssh_server/sshd_set_idle_timeout/rule.yml
```diff
@@ -67,6 +68,9 @@ ocil: |-
...
+        <ul>
+        <li>Remote processes on the remote machine generates output. As the output has to be transferred over the network to the client, the timeout is reset every time such transfer happens.</li>
+        <li>Any <tt>scp</tt> or <tt>sftp</tt> activity by the same user to the host resets the timeout.</li>
+        </ul>
```
redhatrises, line 76: `requires:` needs to be added to the rule for ClientAliveCountMax

matejak: Would you propose to introduce a new rule, or to alter an existing one? We have a rule for `ClientAliveCountMax`, but that one is parametrized, and it may set the value of that configuration option to other values than 0.

matejak:
Could you please clarify that "one rule = change to one configuration line" requirement?

I see that we can choose between this requirement and a clean, self-contained rule. We know what a self-contained rule is, but we don't know anything about that requirement.

linux_os/guide/services/ssh/ssh_server/sshd_set_idle_timeout/rule.yml
redhatrises: Then the rule for ClientAliveCountMax needs to be updated instead of a new rule.

linux_os/guide/services/ssh/ssh_server/sshd_set_idle_timeout/rule.yml
```diff
@@ -8,7 +8,7 @@ description: |-
     <br /><br />
     To set an idle timeout interval, edit the following line in <tt>/etc/ssh/sshd_config</tt> as
     follows:
-    <pre>ClientAliveInterval <b>{{{ xccdf_value("sshd_idle_timeout_value") }}}</b></pre>
+    <pre>ClientAliveInterval <b>{{{ xccdf_value("sshd_idle_timeout_value") }}}</b></pre>.
```
redhatrises, line 11: nitpick... this period `.` should be removed as it doesn't read right.

redhatrises:
Fixed and merged

matejak:
I assume that closing of the PR without merge was not intentional, so I reopen it after removing the period.

matejak:
So this is a really weird glitch - the branch has been merged, but this PR doesn't show it. See e.g. https://github.com/ComplianceAsCode/content/tree/master/linux_os/guide/services/ssh/ssh_server/sshd_set_idle_timeout for the proof that this PR has been really merged.

ggbecker:
> So this is a really weird glitch - the branch has been merged, but this PR doesn't show it. See e.g. https://github.com/ComplianceAsCode/content/tree/master/linux_os/guide/services/ssh/ssh_server/sshd_set_idle_timeout for the proof that this PR has been really merged.

The reason is because @redhatrises has fixed the request by himself and I'm guessing he pushed straight to master since you can see that he has co-authored the commits, for example: https://github.com/ComplianceAsCode/content/commit/5ab82b482aca0e7dbf86216e00e75569950d6e1a

matejak:
OK, that makes sense.
I would prefer to push to the branch, or to suggest a typo-fixing commit like that and merge it, as that would make the history cleaner.

PR #6297, opened 2020-10-26
ocp4: Use ScanSettingBindings for e2e tests

JAORMX:
These are the objects we recommend folks to use instead of raw
ComplianceSuites. It also allows us for auto-detection of platform/node
checks.

### DIFF
--- a/tests/ocp4e2e/helpers.go
+++ b/tests/ocp4e2e/helpers.go
@@ -6,10 +6,8 @@ import (
 	"flag"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"os"
 	"path"
-	"strings"
 	"testing"
 	"time"
 
@@ -19,10 +17,10 @@ import (
 	mcfg "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	mcfgconst "github.com/openshift/machine-config-operator/pkg/daemon/constants"
-	"gopkg.in/yaml.v2"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	extscheme "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/scheme"
+	"k8s.io/apimachinery/pkg/api/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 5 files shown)


JAORMX:
Pull-request updated, HEAD is now dd56b9d5b7bb2033d8aca4fa6e4c57a8fb03839d

tests/ocp4e2e/helpers.go
```diff
@@ -292,92 +249,135 @@ func (ctx *e2econtext) waitForOperatorToBeReady() {
...
+		},
 	}
 
-	// We shouldn't get here.
-	return nil
+	bo := backoff.WithMaxRetries(backoff.NewConstantBackOff(apiPollInterval), 180)
+	err := backoff.RetryNotify(func() error {
+		return ctx.dynclient.Create(goctx.TODO(), binding)
+	}, bo, func(err error, d time.Duration) {
+		fmt.Printf("Couldn't create binding after %s: %s\n", d.String(), err)
```
jhrozek, line 366: Is %s OK to use for formatting an error? I've usually seen %v being used.

jhrozek: Hmm, git grep across some code shows it is, please disregard.

tests/ocp4e2e/helpers.go
```diff
@@ -443,9 +443,68 @@ func (ctx *e2econtext) waitForMachinePoolUpdate(name string) error {
...
+	var lastErr error
+	err = wait.PollImmediate(2*time.Second, 5*time.Minute, func() (bool, error) {
+		suite := &cmpv1alpha1.ComplianceSuite{}
+		key := types.NamespacedName{Name: s, Namespace: ctx.OperatorNamespacedName.Namespace}
+		lastErr = ctx.dynclient.Get(goctx.TODO(), key, suite)
+		if lastErr != nil {
+			return false, nil
+		}
+		if suite.Status.Phase != cmpv1alpha1.PhasePending &&
+			suite.Status.Phase != cmpv1alpha1.PhaseLaunching {
```
jhrozek, line 487: Hmm this sounds racy, what if the phase goes to running very fast? Wouldn't it be more robust to just check that the state changes after applying the annotation?

jhrozek (commented):
Some nits inline, mostly looks good

tests/ocp4e2e/helpers.go
```diff
@@ -292,92 +249,135 @@ func (ctx *e2econtext) waitForOperatorToBeReady() {
...
+		found := &cmpv1alpha1.ProfileBundle{}
+		if err := ctx.dynclient.Get(goctx.TODO(), key, found); err != nil {
+			if errors.IsNotFound(err) {
+				return ctx.dynclient.Create(goctx.TODO(), pb)
+			}
+			return err
+		}
+		// Update the spec in case it differs
+		found.Spec = pb.Spec
+		return ctx.dynclient.Update(goctx.TODO(), found)
```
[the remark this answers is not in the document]
JAORMX: It is, we don't have a cached client enabled here

tests/ocp4e2e/helpers.go
JAORMX: Should I just check that the phase is not "done" then?

jhrozek: I think that would be best. Pending can be extremely fast, Launching should have some delay but...
Checking for != Done seems like safest.

tests/ocp4e2e/helpers.go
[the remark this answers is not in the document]
jhrozek: Ah, sure, dynclient.

JAORMX:
Pull-request updated, HEAD is now 5a26d848ddcfed55b27198d64a9893a4e6395964

jhrozek (approved):
/lgtm

PR #6298, opened 2020-10-26
Remove the option to enable/disable "mask" a service

ggbecker:
#### Description:

- Remove the option to enable/disable `mask` a service and always mask the service on `service_disabled` related rules.

#### Rationale:

- We are not using `disable` option only in any of the rules. So it doesn't make sense to keep this since we use `mask` for all rules. This can cause confusion for users as the description says that the user can either `disable` or `mask`. This way is set in stone that they need to `mask` the service to be compliant with the rule.

- Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1891435

### DIFF
--- a/shared/macros.jinja
+++ b/shared/macros.jinja
@@ -379,9 +379,7 @@ ocil_clause: "the package is installed"
 
 {{%- macro systemd_describe_socket_disable(socket) %}}
     The <code>{{{ socket }}}</code> socket can be disabled with the following command:
-    <pre>$ sudo systemctl disable {{{ socket }}}.socket</pre>
-    The <code>{{{ socket }}}</code> socket can be masked with the following command:
-    <pre>$ sudo systemctl mask {{{ service }}}.socket</pre>
+    <pre>$ sudo systemctl mask {{{ socket }}}.socket</pre>
 {{%- endmacro %}}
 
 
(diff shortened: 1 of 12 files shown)


shared/macros.jinja
```diff
@@ -379,9 +379,7 @@ ocil_clause: "the package is installed"
 
 {{%- macro systemd_describe_socket_disable(socket) %}}
     The <code>{{{ socket }}}</code> socket can be disabled with the following command:
-    <pre>$ sudo systemctl disable {{{ socket }}}.socket</pre>
-    The <code>{{{ socket }}}</code> socket can be masked with the following command:
-    <pre>$ sudo systemctl mask {{{ service }}}.socket</pre>
+    <pre>$ sudo systemctl mask {{{ socket }}}.socket</pre>
```
redhatrises, line 382: To mask, we either need to disable the service first, or add `--now`. Otherwise, service is still running.

ggbecker: Yes, we have to add `--now`, but `disable` has the same behavior, so the instruction wasn't accurate before either:

From the `systemctl` manpage
```
       disable UNIT...
           Disables one or more units. ...

           This command implicitly reloads the system manager configuration after completing the operation. Note that this command does not implicitly stop the units that are being disabled. If this is desired, either combine this
           command with the --now switch, or invoke the stop command with appropriate arguments later.
```

PR #6301, opened 2020-10-27
Some minor cleanups after ocp3 removal

jhrozek:
#### Description:

- Removes a bunch of ocp4 versioned conditionals which are not needed anymore
- Removes the ansible and bash remediations from application/openshift

#### Rationale:

- The version conditionals are removed because ocp4 is currently the only supported version and having the version conditionals just muddies the code
- The ansible and bash remediations were removed because they are not used or tested and untested code means broken code and should just be removed


JAORMX (approved):
awww yes! code removals!

redhatrises:
Thanks! Missed those!

PR #6302, opened 2020-10-27
OCP4: Verify CIS 1.3 section

JAORMX:
This pass fixes some minor issues about the section:

* Set references where needed
* Fix OCIL strings to use appropriate commands
* Fix comments to reflect new titles used in the draft

### DIFF
--- a/applications/openshift/controller/controller_bind_address/rule.yml
+++ b/applications/openshift/controller/controller_bind_address/rule.yml
@@ -34,11 +34,15 @@ ocil_clause: |-
     <tt>bind-address</tt> is not configured to a secure IP address
 
 ocil: |-
-    To verify that <tt>bind-address</tt> is configured correctly,
+    To verify that <tt>secure-port</tt> is configured correctly,
     run the following command:
-    <pre>$ oc get configmaps config -n openshift-kube-controller-manager -ojson | jq -r '.data["config.yaml"]' | jq -r '.extendedArguments["bind-address"]'</pre>
-    Verify that the <tt>bind-address</tt> argument is set to a secure
-    IP address.
+    <pre>$ oc get configmaps config -n openshift-kube-controller-manager -ojson |   jq -r '.data["config.yaml"]' | jq '.extendedArguments["secure-port"][]'</pre>
+    Verify that it's using an appropriate port (the value is not <pre>0</pre>).
+
+    To verify that <tt>port</tt> is configured correctly,
+    run the following command:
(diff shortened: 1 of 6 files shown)


applications/openshift/controller/controller_bind_address/rule.yml
```diff
@@ -34,11 +34,15 @@ ocil_clause: |-
     <tt>bind-address</tt> is not configured to a secure IP address
 
 ocil: |-
-    To verify that <tt>bind-address</tt> is configured correctly,
+    To verify that <tt>secure-port</tt> is configured correctly,
```
redhatrises, line 37: So, this rule is about `bind-address` and not `secure-port`. The change doesn't match the rule and description.

applications/openshift/controller/controller_bind_address/rule.yml
```diff
@@ -34,11 +34,15 @@ ocil_clause: |-
...
     run the following command:
-    <pre>$ oc get configmaps config -n openshift-kube-controller-manager -ojson | jq -r '.data["config.yaml"]' | jq -r '.extendedArguments["bind-address"]'</pre>
-    Verify that the <tt>bind-address</tt> argument is set to a secure
-    IP address.
+    <pre>$ oc get configmaps config -n openshift-kube-controller-manager -ojson |   jq -r '.data["config.yaml"]' | jq '.extendedArguments["secure-port"][]'</pre>
+    Verify that it's using an appropriate port (the value is not <pre>0</pre>).
+
+    To verify that <tt>port</tt> is configured correctly,
+    run the following command:
+    <pre>$ oc get configmaps config -n openshift-kube-controller-manager -ojson |   jq -r '.data["config.yaml"]' | jq '.extendedArguments["port"][]'</pre>
```
redhatrises, line 44: Rules that check multiple different configurations need to be broken out into separate rules. This makes the content more flexible, is a requirement for CCEs, and makes it easier for our end users to understand.

JAORMX:
Pull-request updated, HEAD is now 2d7749ccc88fb0362882a4ea1645e9b8b6d917c5

applications/openshift/controller/controller_bind_address/rule.yml
JAORMX: Done

applications/openshift/controller/controller_bind_address/rule.yml
JAORMX: Done

JAORMX:
CI is broken because it's already using Fedora 33, and for some reason Fedora 33 has only openscap 1.3.3 and not 1.3.4 which we need.

PR #6304, opened 2020-10-27
Generate the CPE Dictionary dynamically

yuumasato:
#### Description: 

- This PR attempts to simplify definition syntax of CPE
  - Removes the products' static cpe-dictionary and generate them dynamically based on the `platform` elements of the content in the shorthand xml.
  - The build system identifies the required CPEs and generates the appropriate product `cpe-dictionary`.
- The the main product doesn't need to contain cpe-items related to derivate product anymore.
  - CPEs for derivative products are injected in the derivative product's cpe-dictionary.
- Product related CPEs are now defined in their respective `product.yml` file.
- Rule related CPEs are sourced from `cpes_root` defined in `products.yml` file.
  - For the CPEs defined in upstream, `cpes_root` is `shared/applicability/cpes.yml`.
- Remove constants `XCCDF_PLATFORM_TO_CPE` and `PRODUCT_TO_CPE_MAPPING`.
  These mapping are now generated dynamically, based on the `cpes` defined to the product.

#### Rationale:
- Simplify CPE declaration
- Avoids the need to manually add a new cpe-item for each product.


shared/applicability/cpes.yml
```diff
@@ -0,0 +1,325 @@
...
+
+  "cpe:/o:example":
+    id: example
+    title: "Example"
+    check_id: installed_OS_is_part_of_Unix_family
+
+  "cpe:/o:fedoraproject:fedora:32":
+    id: fedora32
+    title: "Fedora 32"
+    check_id: installed_OS_is_fedora
```
redhatrises, line 50: This should be in the product.yml file.

shared/applicability/cpes.yml
```diff
@@ -0,0 +1,325 @@
...
+
+  "cpe:/a:mozilla:firefox":
+    id: firefox
+    title: "Mozilla Firefox"
+    check_id: installed_app_is_firefox
```
redhatrises, line 55: This should be in the product.yml file.

shared/applicability/cpes.yml
```diff
@@ -0,0 +1,325 @@
...
+
+  "cpe:/a:redhat:jboss_fuse:6.0":
+    id: fuse6
+    title: "JBoss Fuse 6.0"
+    check_id: installed_app_is_fuse6
```
redhatrises, line 60: This should be in the product.yml file.

shared/applicability/cpes.yml
```diff
@@ -0,0 +1,325 @@
...
+
+  "cpe:/o:debianproject:debian:9":
+    id: debian9
+    title: "Debian release 9 (Stretch)"
+    check_id: installed_OS_is_debian9
+
+  "cpe:/o:debian:debian_linux:10":
+    id: debian10
+    title: "Debian Linux 10"
+    check_id: installed_OS_is_debian10
```
redhatrises, line 40: This should be in the product.yml file.

shared/applicability/cpes.yml
```diff
@@ -0,0 +1,325 @@
...
+
+  "cpe:/a:container":
+    id: container
+    title: "Container"
+    check_id: installed_env_is_a_container
+
+  "cpe:/o:debianproject:debian:9":
+    id: debian9
+    title: "Debian release 9 (Stretch)"
+    check_id: installed_OS_is_debian9
```
redhatrises, line 35: This should be in the product.yml file.

shared/applicability/cpes.yml
```diff
@@ -0,0 +1,325 @@
...
+
+  "cpe:/o:centos:centos:8":
+    id: centos8
+    title: "CentOS 8"
+    check_id: installed_OS_is_centos8
+
+  "cpe:/a:google:chromium-browser":
+    id: chromium
+    title: "Google Chromium Browser"
+    check_id: installed_app_is_chromium
```
redhatrises, line 20: This should be in the product.yml file.

shared/applicability/cpes.yml
```diff
@@ -0,0 +1,325 @@
...
+
+  "cpe:/a:grub2":
+    id: grub2
+    title: "Package grub2 is installed"
+    check_id: installed_env_has_grub2_package
+
+  "cpe:/a:oracle:jre:":
+    id: jre-oracle
+    title: "Oracle's Java Runtime Environment"
+    check_id: installed_app_is_java
```
redhatrises, line 75: This should be in the product.yml file.

shared/applicability/cpes.yml
```diff
@@ -0,0 +1,325 @@
...
+
+  "cpe:/a:ibm:jre:":
+    id: jre-ibm
+    title: "IBM Java Runtime Environment"
+    check_id: installed_app_is_java
```
redhatrises, line 80: This should be in the product.yml file.

shared/applicability/cpes.yml
```diff
@@ -0,0 +1,325 @@
...
+
+  "cpe:/a:redhat:openjdk:":
+    id: openjdk-redhat
+    title: "OpenJDK Java Runtime Environment<"
+    check_id: installed_app_is_java
```
redhatrises, line 85: This should be in the product.yml file.

shared/applicability/cpes.yml
```diff
@@ -0,0 +1,325 @@
...
+
+  "cpe:/a:sun:jre:":
+    id: jre-oracle
+    title: "Sun Java Runtime Environment"
+    check_id: installed_app_is_java
```
redhatrises, line 90: This should be in the product.yml file.

shared/applicability/cpes.yml
```diff
@@ -0,0 +1,325 @@
...
+
+  "cpe:/o:apple:macos:10.15":
+    id: macos
+    title: "Apple macOS 10.15"
+    check_id: installed_OS_is_macos1015
+
+  "cpe:/a:mcafee:virusscan_enterprise_for_linux:1.9":
+    id: mcafee-1.9
+    title: "McAfee's VirusScan Enterprise for Linux (VSEL) 1.9.x"
+    check_id: installed_app_is_vsel
```
redhatrises, line 115: This should be in the product.yml file.

shared/applicability/cpes.yml
```diff
@@ -0,0 +1,325 @@
...
+
+  "cpe:/a:mcafee:virusscan_enterprise_for_linux:2.0":
+    id: mcafee-2.0
+    title: "McAfee's VirusScan Enterprise for Linux (VSEL) 2.0.x"
+    check_id: installed_app_is_vsel
```
redhatrises, line 120: This should be in the product.yml file.

shared/applicability/cpes.yml
```diff
@@ -0,0 +1,325 @@
...
+
+  "cpe:/a:ntp":
+    id: ntp
+    title: "Package ntp is installed"
+    check_id: installed_env_has_ntp_package
+
+  "cpe:/a:redhat:openshift_container_platform:4.1":
+    id: ocp4.1
+    title: "Red Hat OpenShift Container Platform 4"
+    check_id: installed_app_is_ocp4
```
redhatrises, line 140: This should be in the product.yml file.

shared/applicability/cpes.yml
```diff
@@ -0,0 +1,325 @@
...
+
+  "cpe:/o:redhat:openshift_container_platform_node:4":
+    id: ocp4.1
+    title: "Red Hat OpenShift Container Platform 4 Node"
+    check_id: installed_app_is_ocp4_node
```
redhatrises, line 145: This should be in the product.yml file.

shared/applicability/cpes.yml
```diff
@@ -0,0 +1,325 @@
...
+
+  "cpe:/o:oracle:linux:7":
+    id: ol7
+    title: "Oracle Linux 7"
+    check_id: installed_OS_is_ol7_family
```
redhatrises, line 150: This should be in the product.yml file.

shared/applicability/cpes.yml
```diff
@@ -0,0 +1,325 @@
...
+
+  "cpe:/o:oracle:linux:8":
+    id: ol8
+    title: "Oracle Linux 8"
+    check_id: installed_OS_is_ol8_family
```
redhatrises, line 155: This should be in the product.yml file.

shared/applicability/cpes.yml
```diff
@@ -0,0 +1,325 @@
...
+
+  "cpe:/a:pam":
+    id: pam
+    title: "Package pam is installed"
+    check_id: installed_env_has_pam_package
+
+  "cpe:/o:redhat:enterprise_linux_coreos:4":
+    id: rhcos4
+    title: "Red Hat Enterprise Linux CoreOS 4"
+    check_id: installed_OS_is_rhcos4
```
redhatrises, line 185: This should be in the product.yml file.

shared/applicability/cpes.yml
```diff
@@ -0,0 +1,325 @@
...
+
+  "cpe:/o:redhat:enterprise_linux:6::client":
+    id: rhel6-client
+    title: "Red Hat Enterprise Linux 6 Client"
+    check_id: installed_OS_is_rhel6
+
+  "cpe:/o:redhat:enterprise_linux:6::computenode":
+    id: rhel6-computenode
+    title: "Red Hat Enterprise Linux 6 ComputeNode"
+    check_id: installed_OS_is_rhel6
```
redhatrises, line 200: rhel6 should be dropped

shared/applicability/cpes.yml
```diff
@@ -0,0 +1,325 @@
...
+
+  "cpe:/o:redhat:enterprise_linux:7":
+    id: rhel7
+    title: "Red Hat Enterprise Linux 7"
+    check_id: installed_OS_is_rhel7
```
redhatrises, line 205: This should be in the product.yml file.

shared/applicability/cpes.yml
```diff
@@ -0,0 +1,325 @@
...
+
+  "cpe:/o:redhat:enterprise_linux:7::server":
+    id: rhel7-server
+    title: "Red Hat Enterprise Linux 7 Server"
+    check_id: installed_OS_is_rhel7
```
redhatrises, line 210: This should be in the product.yml file.

ssg/constants.py
```diff
@@ -173,6 +173,15 @@
...
     ],
+    "centos6": [
+        "cpe:/o:centos:centos:6",
+    ],
+    "centos7": [
+        "cpe:/o:centos:centos:7",
+    ],
+    "centos8": [
+        "cpe:/o:centos:centos:8",
+    ],
```
redhatrises, line 184: all cpes need to be dropped from constants.py

shared/applicability/cpes.yml
```diff
@@ -0,0 +1,325 @@
...
+
+  "cpe:/a:redhat:openstack:13":
+    id: rhosp13
+    title: "Red Hat OpenStack Platform 13"
+    check_id: installed_app_is_rhosp13
+
+  "cpe:/a:redhat:enterprise_virtualization_manager:4":
+    id: rhvm4
+    title: "Red Hat Virtualization 4 Manager"
+    check_id: installed_app_is_rhv4
```
redhatrises, line 250: Overall structure needs to be cleaner and clearer like so:

```
￼    - id: rhvm4
  ￼    title: "Red Hat Virtualization 4 Manager"
      cpe:  "cpe:/a:redhat:enterprise_virtualization_manager:4"
￼      check_id: installed_app_is_rhv4

**Review by redhatrises — CHANGES_REQUESTED** [review]

File: shared/applicability/cpes.yml (hunk 22 above)
Comment (redhatrises) [reply shared/applicability/cpes.yml:250]: there also needs to be an additional key called `additional_cpes` or `include_cpes` that maps to `shared/applicability/cpes.yml`

File: shared/applicability/cpes.yml (hunk 1 above)
Comment (yuumasato) [reply shared/applicability/cpes.yml:50]: Yeah, It think it would be better placed at the `product.yml`.
But I didn't want to overhaul it all at once.

This first step is just to unify the source of the CPEs.

File: ssg/constants.py (hunk 21 above)
Comment (yuumasato) [reply ssg/constants.py:184]: Actually a lot of CPE mappings can be dropped from `ssg/constants.py`.
I'm not sure if it would better be done in this PR or a subsequent one.
I lean towards doing it in a separate PR.

File: shared/applicability/cpes.yml (hunk 22 above)
Comment (yuumasato) [reply shared/applicability/cpes.yml:250]: > there also needs to be an additional key called `additional_cpes` or `include_cpes` that maps to `shared/applicability/cpes.yml`

Why, if the CPEs are picked up automatically based on the content included in the benchmark?

File: ssg/constants.py (hunk 21 above)
Comment (redhatrises) [reply ssg/constants.py:184]: sounds good

File: shared/applicability/cpes.yml (hunk 22 above)
Comment (redhatrises) [reply shared/applicability/cpes.yml]: To not only make the content readable and understandable, but also remove the abstraction of the build system. Makes it easy for content authors and everyone to read and understand additional cpe is being added in `product.yml`

File: shared/applicability/cpes.yml (hunk 1 above)
Comment (redhatrises) [reply shared/applicability/cpes.yml]: It is better to overhaul and move the CPEs into `product.yml` and move the shared cpes into the `shared/applicability/cpes.yml` now.

Comment (redhatrises) [reply shared/applicability/cpes.yml]: Otherwise, down with these changes btw as this helps to make the content easier to use for sure.

File: shared/applicability/cpes.yml [hunk 23]
```diff
@@ -0,0 +1,324 @@
+centos6:
```
redhatrises, line 1: Should still have this be `cpe:` and then a yaml lists of dictionaries.

yuumasato: Why?

redhatrises: Readability and understanding

shared/applicability/cpes.yml
yuumasato: I find it rather cumbersome to, after selecting a rule in the profile, to have to add any `cpe-item` it uses to the `product.yml`.
I'd expect the build-system to be handle that automatically for me (include the CPE in the dictionary of the product).

I think the readability and understandability  should be solved in another way.
Maybe "compiled" products yamls? Build logs?

shared/applicability/cpes.yml
yuumasato: Why jump ahead of ourselves? :)

shared/applicability/cpes.yml
yuumasato: Moved them.

shared/applicability/cpes.yml
redhatrises: having the cpe in the product.yml makes it easy for content authors to understand.... and no... expecting the build system to handle something is a developerism and not a content authoring-ism.

shared/applicability/cpes.yml
yuumasato: Right, I'm fine with that.

shared/applicability/cpes.yml
yuumasato: > having the cpe in the product.yml makes it easy for content authors to understand....

Understand what?

The main point of the PR (making it dynamically generated) is to avoid the need to edit every products' file to add a new CPE every time a new CPE is created. Because otherwise, it fails SCAPVal, and may fail during scan with "unknown CPE".
I don't want to keep the same "add something here" gating-like mechanism for content to be correctly built.
The use case it handles is:
1. I created a new rule and it needs to use a new CPE, which I also just created;
1. The rule is for a component very common in all linux distros, so I have not specified a `prodtype`.
1. I selected the rule in the profiles of products I care;
1. Oh no, suddenly all other linux products that picked this rule up into the benchmark started to fail SCAP validation, :(
1. Now someone needs to add the CPE to all other products.



> expecting the build system to handle something is a developerism

I call this convenience of automation.

> content authoring-ism.

Where can I find more about this?
I'm very interested in helping content authors and their use cases.

yuumasato:
I'm currently updating documentation around CPEs.
Don't merge yet, please.

shared/applicability/cpes.yml
matejak: > ... having the cpe in the product.yml makes it easy for content authors to understand...

Imagine that there is a CPE in `product.yml`, and I think that it got there by mistake, or that the reason why it was introduced is not valid any more. What would I have to do in order to resolve those doubts?

yuumasato: Maybe I don't understand where you are heading...

If a CPE entry in a `product.yml` was not trivial, I would expect comment lines, if there weren't any I'd look into the commit that added it.

matejak: Everything has a cost - if the automation performs some magic behind the content author's back, that can have a cost, especially when the magic doesn't work right, or if the author would like something slightly different.

Conversely, if there is no automation, the obvious cost is that you have to add config items s.a. CPEs manually. But that's not where it ends - you get risk of copy-paste errors, and you also have to deal with removal of items that are no longer needed or that have been removed by refactoring.

I think I understand @redhatrises concern of automating too much, and I think that we should look both ways - what are the risks of automating too little?
So far it seems that we should require non-trivial manually added CPEs to be described by comments.

yuumasato: I agree that automation can make it hard to do something slightly different. But that was not the original point of the question, and I think it can be handled in a different way.

>  So far it seems that we should require non-trivial manually added CPEs to be described by comments. 

Now I'm not sure if your previous question was about:
- the manually added CPE for the product in `product.yml` or
- the automatically added CPEs in the product's `cpe-dictionary`.

To clarify, the automation doesn't add anything to `product.yml` file, it will include `cpe-item` in the `cpe-dictionary`, as the files removed by this PR (rhel8/cpe/rhel8-cpe-dictionary.xml).
And to check the CPEs added by the build system one can check the built dictionary. It should have the same complexity as reviewing the old static files.

matejak: My question was about items that would have to be added to `product.yml` if there is no automation. The automation is supposed to make sure that only and only CPEs that are needed get added. Conversely, when a CPE is added manually, it has to be kept in sync with the actual content.

You have answered the question, thanks.

chromium/product.yml
```diff
@@ -5,3 +5,9 @@ type: product
 benchmark_root: "./guide"
 
 profiles_root: "./profiles"
+
+cpes:
+  chromium:
```
redhatrises, line 10: nitpick: please make this `- chromium`.

chromium/product.yml
```diff
@@ -5,3 +5,9 @@ type: product
...
+    name: "cpe:/a:google:chromium-browser"
+    title: "Google Chromium Browser"
+    check_id: installed_app_is_chromium
```
redhatrises, line 13: Otherwise, these changes are incredibly better than the cpe files.

jre/product.yml
```diff
@@ -5,3 +5,24 @@ type: product
 benchmark_root: "./guide"
 
 profiles_root: "./profiles"
+
+cpes:
+  jre-oracle:
+    name: "cpe:/a:oracle:jre:"
+    title: "Oracle's Java Runtime Environment"
+    check_id: installed_app_is_java
```
redhatrises, line 13: do we need `jre-oracle`? Why not skip it and start with `- name`?

chromium/product.yml
redhatrises, line 13: Just like the `profiles_root`, I want to have `additional_cpes` in this file, because as a content author, I don't want to choose to add `cpes.yml` or not. I might come with my own `cpes.yml` file that doesn't contain what we have now.

jre/product.yml
yuumasato: The product CPES don't need to have these `jre-oracle` and `jre-ibm`... They are like IDs for the CPEs, but are not actually used by the build system.

But, the IDs of the CPEs in `shared/applicability/cpes.yml` are needed, these are IDs are referenced by rules and profiles.
I just tried to keep same declaration format.

chromium/product.yml
yuumasato: I can make this change, but I don't see the point in making it a list of CPEs.
Is this request aesthetic or functional?

redhatrises: both. It makes it easier to read a file full of them.

chromium/product.yml
redhatrises: > So you'd like to be able to choose the file to source the CPE's from?
> 
> So if `product.yml` specified a directory, and any `yml` (or even `xml`) file there was sourced. Would it satisfy your needs?
> 
> ```
> cpes_root: "../shared/applicability"
> ```

It definitely would address the needs!

jre/product.yml
redhatrises: > The product CPES don't need to have these `jre-oracle` and `jre-ibm`... They are like IDs for the CPEs, but are not actually used by the build system.
> 
> But, the IDs of the CPEs in `shared/applicability/cpes.yml` are needed, these are IDs are referenced by rules and profiles.
> I just tried to keep same declaration format.

So, if I am understanding you correctly, `jre-oracle`, `jre-ibm`, etc. aren't needed by the build system in `product.yml` but are needed by the build system in `cpes.yml`?

yuumasato: Well, the IDs are there for any content to refer to them.
But as they are already declared in the `product.yml` they are for, nothing else references them.
Rules and profiles refer to the IDs of `cpes` declared in `cpes.yml`.

I'd rather keep the syntax the same for `cpes` in `product.yml` and `cpes.yml`.

chromium/product.yml
yuumasato: Made the change in c16efcd

chromium/product.yml
yuumasato: Possibility to load multiple source CPE yml files added in 05506be

jre/product.yml
yuumasato: Actually, the CPE IDs are needed.
The `ocp4` profiles reference CPE's defined in the `product.yml`.

ssg/constants.py
yuumasato: `PRODUCT_TO_CPE_MAPPING` was removed

redhatrises:
At this point, looks good to me. I also like how you broke out the shared CPEs into categories. I think that that is a nice touch.

PR #6308, opened 2020-10-28
Adding app-security-dev to stig_ns for JRE updates

carlosmmatos:
#### Description:

Adding the app-sec-dev filter to the stig_ns allows creating/updating overlay files for JRE. Otherwise, it would fail to map the RULE ID's appropriately


redhatrises:
no need to waste testing resources on this.

PR #6313, opened 2020-10-29
utils/build_ds_container.sh: Enable in-cluster builds

JAORMX:
This enables a mode of building the content in-cluster, as opposed to
having folks build the content locally and then pushing it.

To enable this mode, the `-c` flag was introduced.


JAORMX:
Pull-request updated, HEAD is now 00e353956b8cb99144a0b67cabb0a6307996814a

jhrozek (approved):
/lgtm

PR #6314, opened 2020-10-29
ocp4: Use Fedora 33 for container files

JAORMX:
Fedora 33 now has OpenSCAP 1.3.4 which is what we need to build. So we
can take it into use. This updates the image tag to use 33 specifically
as opposed to using "latest" in order to avoid stability issues that
might come wit upgrades.


jhrozek (approved):
/lgtm

PR #6315, opened 2020-10-29
OCP4: Enhance e2e tests to check individual rules

JAORMX:
This enhances the e2e tests to be able to checkf or individual rules.

This requires folks to create a `/tests/ocp4` directory in the rule
that's to be tested. If the directory doesn't exist, the verification
will be skipped.

If the directory exists, it'll then look for a `e2e.yml` file that
contains the test definition.

The current format is:

```
---
default_result: <result>
result_after_remediation: <result>
```

With result being either `PASS` or `FAIL`.

If there were remediations applied the test will either check the value
of `result_after_remediation` if it's not empty or `default_result`.

This enables us to test that rules don't regress and that they issue the
expected results.

### DIFF
--- a/applications/openshift/controller/controller_use_service_account/tests/ocp4/e2e.yml
+++ b/applications/openshift/controller/controller_use_service_account/tests/ocp4/e2e.yml
@@ -1,0 +1,2 @@
+---
+default_result: PASS 
--- a/tests/ocp4e2e/e2e_test.go
+++ b/tests/ocp4e2e/e2e_test.go
@@ -55,7 +55,7 @@ func TestE2e(t *testing.T) {
 		ctx.waitForComplianceSuite(suite)
 		numberOfRemediations = ctx.getRemediationsForSuite(suite)
 		numberOfFailuresInit = ctx.getFailuresForSuite(suite)
-		numberOfCheckResultsInit = ctx.getCheckResultsForSuite(suite)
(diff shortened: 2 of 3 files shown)


ggbecker (approved):
From the ComplianceAsContent project perspective this looks good to me. I think we should concentrate tests in the `tests` folder so others are aware of their existence. We need to think about tests specific to `linux` if it would be better to move to a separate folder as well. The same as this PR proposes. The downside is that thousand of files have to be moved though.

jhrozek (approved):
lgtm

This looks good to me as well and is pretty much what I had in mind. One question though (non-blocking), do you have an idea how to extend the format in case different OCP releases differ in what we expect? Something like default_result@ocp4.6: pass?

JAORMX:
> lgtm
> 
> This looks good to me as well and is pretty much what I had in mind. One question though (non-blocking), do you have an idea how to extend the format in case different OCP releases differ in what we expect? Something like default_result@ocp4.6: pass?

I thought about it but decided not to address this as of yet.

One idea is to have a list of tests:
```
- default_result: PASS
  applicable_versions: 4.6
- default_result: FAIL
  applicable_versions: 4.7
```

In the test, we would iterate until finding the appropriate test....

What you propose (value@version) is also a viable idea.

I think it would be better if we implement this once a case comes up.

PR #6316, opened 2020-10-29
Bump scap-workbench and openscap version in README doc.

ggbecker:
#### Description:

- Bump scap-workbench and openscap version in README doc.

#### Rationale:

- Reference newer documentation


redhatrises:
Can we update these links so that they go to the projects themselves rather than versions and versions docs so that we don't have to keep updating this file?

PR #6318, opened 2020-10-30
OCP4: Add manual remediation capabilities to e2e tests

JAORMX:
this adds the ability to run manual remediations as part of the e2e
tests. These remediations cannot be currently done via the
compliance-operator, because either they only work on partial objects or
they require an opinionated decision about the deployment which the
operator cannot make.

These remediations will live in the ocp4 tests directory with the
filename `e2e-remediation.sh`. When this file is found, it'll get run.

The remediation is expected to wait for the status to converge.

There is a default timeout of 15 mins for the remediation script(s) to
finish.


jhrozek (approved):
/lgtm

PR #6323, opened 2020-10-30
Implement variable metadata for test scenarios

matejak:
- Modify datastreams, so XCCDF values have defaults that we want.
- Use XSLT transform to alter datastreams - OpenSCAP requires xsltproc, so the utility is installed.
- ~Datastreams are part of snapshots, so no worries with overwriting them.~ Datastreams have to be quasi-snapshotted on the host - there are no datastreams uploaded to the scanned guest, as `oscap-ssh` does the scan.

This PR requires #6322 to be merged in order to work.

### DIFF
--- a/tests/ssg_test_suite/rule.py
+++ b/tests/ssg_test_suite/rule.py
@@ -2,12 +2,15 @@ from __future__ import print_function
 
 import logging
 import os
+import shutil
 import os.path
 import re
 import subprocess
 import collections
 import json
 import fnmatch
+import tempfile
+import contextlib
 
 from ssg.constants import OSCAP_PROFILE, OSCAP_PROFILE_ALL_ID, OSCAP_RULE
 from ssg_test_suite import oscap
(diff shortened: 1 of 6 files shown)


jan-cerny:
@matejak It would be nice to show us an explample

matejak:
Sure, it's still draft, I will move it forward once the dependency is merged.

jan-cerny:
@matejak you can rebase

matejak:
So it turned out that there was no dependency to this PR.
Anyway, you can check the example in those test scenarios.

tests/ssg_test_suite/rule.py
```diff
@@ -344,16 +363,45 @@ def _check_and_record_rule_scenario(self, scenario, remote_rule_dir, rule_id, re
...
-        self._check_rule_scenario(scenario, remote_rule_dir, rule_id, remediation_available)
+        with self.copy_of_datastream():
+            self._check_rule_scenario(scenario, remote_rule_dir, rule_id, remediation_available)
         self.results.append(self._current_result.save_to_dict())
 
+    @contextlib.contextmanager
+    def copy_of_datastream(self, new_filename=None):
+        old_filename = self.datastream
+        if not new_filename:
+            _, new_filename = tempfile.mkstemp(prefix="ssgts_ds_modified", dir="/tmp")
```
jan-cerny, line 374: Why is a temporary datastream created once here and second time in _change_variable_value? How does it work?

tests/ssg_test_suite/rule.py
```diff
@@ -344,16 +363,45 @@ def _check_and_record_rule_scenario(self, scenario, remote_rule_dir, rule_id, re
...
+        shutil.copy(old_filename, new_filename)
+        self.datastream = new_filename
+        yield new_filename
+        self.datastream = old_filename
+        os.unlink(new_filename)
+
+    def _change_variable_value(self, varname, value):
+        XSLT_FILENAME = "/tmp/change.xslt"
```
jan-cerny, line 382: This would be better as a temporary file instead

tests/ssg_test_suite/rule.py
matejak: Copying of datastereams works as follows:

- Copy upon start of the test suite: When the test suite starts, it copies the datastream, so the user can e.g. rebuild without affecting tests.
- Copy before executing a rule scenario: Newly introduced, this code assumes that any test scenario may make edits to the datastream that shouldn't be persistent, so every scenario just gets a copy.
- Copy during value change: That one is just an implementation detail - it wouldn't be needed if `xsltproc` could do an in-place modification.

tests/ssg_test_suite/rule.py
matejak: OK

jan-cerny:
@matejak Thank you very much for you pull request!

PR #6326, opened 2020-10-30
Remove unused RHEL7 profiles

redhatrises:
- While these profiles need to be developed, they won't be in completed in time for being able to be used downstream


jan-cerny:
Thanks for removing them.

PR #6329, opened 2020-11-02
OCP4/CIS 1.4.1: Remove invalid rule and add reference to actual check

JAORMX:
The actual check was being done in another rule, so let's mark the
reference.

On the other hand, 1.4.1 was outputting an "info" rule which didn't even
describe the correct thing to do. So let's remove it instead.


jhrozek (approved):
lgtm,
the resulting built DS contains references to both CIS controls and the rbac_debug_role_protects_pprof does belong to both controls (although it doesn't cover them fully)

PR #6331, opened 2020-11-02
OCP4/CIS 2: Fix OCIL and add tests

JAORMX:
For rules that were covered during the hackfest, this does some minor fixes and
adds tests.

Note that some rules were left untouched because they need actual checks
entirely.

### DIFF
--- a/applications/openshift/etcd/etcd_key_file/rule.yml
+++ b/applications/openshift/etcd/etcd_key_file/rule.yml
@@ -6,10 +6,10 @@ title: 'Ensure That The etcd Key File Is Correctly Set'
 
 description: |-
     To ensure the <tt>etcd</tt> service is serving TLS to clients,
-    edit the <tt>etcd</tt> configuration file
-    <tt>/etc/etcd/etcd.conf</tt> on the master and
-    adding a key file to <tt>ETCD_KEY_FILE</tt>:
-    <pre>ETCD_CERT_FILE=/etc/ssl/etcd/system:etcd-server:<i>etcd_dns_name</i>.key</pre>
+    make sure the <tt>etcd-pod*</tt> <tt>ConfigMaps</tt> in the
+    <tt>openshift-etcd</tt> namespace contain the following argument
+    for the <tt>etcd</tt> binary in the <tt>etcd</tt> pod:
+    <pre>oc get -nopenshift-etcd cm etcd-pod -oyaml | grep "\-\-key-file=/etc/kubernetes/static-pod-certs/secrets/etcd-all-serving/etcd-serving-NODE_NAME.key"</pre>
 
 rationale: |-
     Without cryptographic integrity protections, information can be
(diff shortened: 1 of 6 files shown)


applications/openshift/etcd/etcd_key_file/rule.yml
```diff
@@ -26,6 +26,20 @@ references:
...
+        {{{ openshift_cluster_setting("/api/v1/namespaces/openshift-etcd/configmaps/etcd-pod") | indent(8) }}}
+
+template:
+    name: yamlfile_value
+    vars:
+        ocp_data: "true"
+        filepath: /api/v1/namespaces/openshift-etcd/configmaps/etcd-pod
+        yamlpath: ".data['pod.yaml']"
+        values:
+          - value: ".*--key-file=/etc/kubernetes/static-pod-certs/secrets/etcd-all-serving/etcd-serving-NODE_NAME.key \\.*"
```
mrogers950, line 44: what happens with NODE_NAME here? I thought this would not be a substituted value. 

The --cert-file and --key-file checks can be a check that these flags are set with _any_ value, since they are required arguments.

JAORMX: the literal value is NODE_NAME. I don't know why. But that's how the file actually looks like.

jhrozek:
On Mon, Nov 02, 2020 at 10:48:15PM -0800, Juan Osorio Robles wrote:
> @JAORMX commented on this pull request.
> 
> 
> 
> > +    Run the following command:
> +    <pre>oc get -nopenshift-etcd cm etcd-pod -oyaml | grep "\-\-key-file=/etc/kubernetes/static-pod-certs/secrets/etcd-all-serving/etcd-serving-NODE_NAME.key"</pre>
> +    Verify that there is a certificate configured.
> +
> +warnings:
> +    - general: |-
> +        {{{ openshift_cluster_setting("/api/v1/namespaces/openshift-etcd/configmaps/etcd-pod") | indent(8) }}}
> +
> +template:
> +    name: yamlfile_value
> +    vars:
> +        ocp_data: "true"
> +        filepath: /api/v1/namespaces/openshift-etcd/configmaps/etcd-pod
> +        yamlpath: ".data['pod.yaml']"
> +        values:
> +          - value: ".*--key-file=/etc/kubernetes/static-pod-certs/secrets/etcd-all-serving/etcd-serving-NODE_NAME.key \\.*"
> 
> the literal value is NODE_NAME. I don't know why. But that's how the file actually looks like.

I don't know where the expansion happens and by whom, but when the CM is
turned into a static pod definition, then NODE_NAME is expanded. The
static pod definitions on the FS have the per-node value.

This is why I was thinking we might want to check the files on the FS
in some later content iteration and match the values against the node
name, which would probably require hand-written oval though and iterate
over the nodes. Not really required to do now for sure.

PR #6334, opened 2020-11-03
Add rule for password hashing rounds in pam_unix

yuumasato:
#### Description:

- Add variable (`var_password_pam_unix_rounds`) to set number of rounds required for the password hashing algorithm.
- Add rule to configure the number of hashing rounds in pam_unix.so (`accounts_password_pam_unix_rounds`)
  - ... with oval, remediations and tests.
- The system default number of rounds is `5000`.
  - When `rounds=` option is absent in `pam_unix.so` module, the system default is compared against `var_password_pam_unix_rounds`.

#### Rationale:

- The number of `rounds` for password hashing algorithm should be set via PAM modules. Setting it via `/etc/login.defs` is not effective.


linux_os/guide/system/accounts/accounts-restrictions/password_storage/accounts_password_pam_unix_rounds/rule.yml
```diff
@@ -0,0 +1,41 @@
...
+
+prodtype: fedora,ol7,ol8,rhel7,rhel8,rhv4
+
+title: 'Set number of Password Hashing Rounds'
+
+description: |-
+    Configure the number or rounds for the password hashing algorithm. This can be
+    accomplished by using the <tt>rounds</tt> option for the <tt>pam_unix</tt> PAM module.
+    <br /><br />
+    In the files <tt>/etc/pam.d/system-auth</tt> and <tt>/etc/pam.d/password-auth</tt>, append <tt>rounds=<sub idref="var_password_pam_unix_rounds" /></tt>
```
redhatrises, line 11: There needs to be 2 rules here... one for system-auth and the other for password-auth. Not all systems configure password-auth as they don't always have network connectivity.

yuumasato: Alright, I split the rule.

jan-cerny (approved):
It's split into 2 rules now and it looks fine.

PR #6335, opened 2020-11-03
ocp4: Add a couple of checks for CIS 1.4.2

JAORMX:
While these checks are not exchaustive, they're a couple of things that need to
be checked for.


jhrozek (approved):
The rules look in line with 1.4.2 and seem to work fine.
lgtm

PR #6336, opened 2020-11-03
Add more OCP4/CIS 4.1 test results.

mrogers950:
Adds tests results and a correction to the 4.1.5 check.

### DIFF
--- a/applications/openshift/worker/file_permissions_kubelet_conf/rule.yml
+++ b/applications/openshift/worker/file_permissions_kubelet_conf/rule.yml
@@ -5,7 +5,7 @@ prodtype: ocp4
 title: 'Verify Permissions on The Kubelet Configuration File'
 
 description: |-
-    {{{ describe_file_permissions(file="/etc/kubernetes/kubelet.conf", perms="0600") }}}
+    {{{ describe_file_permissions(file="/etc/kubernetes/kubelet.conf", perms="0644") }}}
 
 rationale: |-
     If the kubelet configuration file is writable by a group-owner or the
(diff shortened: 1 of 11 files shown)


applications/openshift/worker/file_permissions_kubelet_conf/rule.yml
```diff
@@ -5,7 +5,7 @@ prodtype: ocp4
 title: 'Verify Permissions on The Kubelet Configuration File'
 
 description: |-
-    {{{ describe_file_permissions(file="/etc/kubernetes/kubelet.conf", perms="0600") }}}
+    {{{ describe_file_permissions(file="/etc/kubernetes/kubelet.conf", perms="0644") }}}
```
redhatrises, line 8: This is a good time to ask: why permissions are moving from more secure to insecure? This is kinda backwards, and if we are doing default file permissions also makes me think that there is no value in file permission checks now. This is a question as to why this is being done.

JAORMX: The check was probably written to check for 0600, only to now discover that the default is actually 644. We might want to start restricting these, as was done for the audit logs https://github.com/openshift/cluster-openshift-apiserver-operator/pull/397

jhrozek: +1 to tracking hardening the permissions in a separate ticket. Currently the changes to the rule reflect both the CIS benchmark and the permissions on the cluster.

jhrozek:
Can you also add the rule to the cis-node profile? Currently it is unused.

mrogers950:
> Can you also add the rule to the cis-node profile? Currently it is unused.

Done, thanks

applications/openshift/worker/file_permissions_kubelet_conf/rule.yml
redhatrises: If we can open a ticket for all permissions in CIS benchmark to be `0600` instead of `0644` and add the ticket here, that would be great.

JAORMX:
The kubeconfig files don't contain credentials, and instead are just references to the apiserver. Those are not sensitive files.

redhatrises:
> The kubeconfig files don't contain credentials, and instead are just references to the apiserver. Those are not sensitive files.

(playing devil's advocate) If that's the case, why create rules for ownership and permissions if the files are not sensitive at all? Can an attacker discover any sort of cluster information from them? If so, the method of "discovery" is a source of information an attacker can use to attack a cluster based off of information from the kubeconfig files.

JAORMX:
> > The kubeconfig files don't contain credentials, and instead are just references to the apiserver. Those are not sensitive files.
> 
> (playing devil's advocate) If that's the case, why create rules for ownership and permissions if the files are not sensitive at all? 

@redhatrises the protection comes from the need for those files to not be writable by any other user.

> Can an attacker discover any sort of cluster information from them? If so, the method of "discovery" is a source of information an attacker can use to attack a cluster based off of information from the kubeconfig files.

They can discover the cluster URL for a given endpoint... although, if they already got into a node in the cluster, they would know this already.

JAORMX:
@redhatrises also, to that extend, I rather have these files more restricted. Giving no visibility at all to an attacker is always better. That's why I proposed https://github.com/openshift/machine-config-operator/pull/2202 some days ago.

PR #6347, opened 2020-11-09
Remove capture_output option from subprocess.run in SSGTS

mildas:
#### Description:
The `capture_output` option was added in Python 3.7 - https://docs.python.org/3/library/subprocess.html

> Changed in version 3.7: Added the text parameter, as a more understandable alias of universal_newlines. Added the capture_output parameter.

However, RHEL 7 has Python 3.6 thus SSGTS does not work there at all.

This PR changes the `capture_output` option to `stdout=PIPE` + `stderr=PIPE` that has the same effect as `capture_output=True`.


yuumasato (approved):
LGTM.

PR #6348, opened 2020-11-09
Select sshd_set_keepalive where sshd_set_idle_timeout is selected.

ggbecker:
#### Description:

- Select sshd_set_keepalive where sshd_set_idle_timeout is selected. Update document reference for `rhel8/pci-dss` profile

#### Rationale:

- sshd_set_keepalive is required to be selected in order to
sshd_set_idle_timeout to work properly.

Relates to: https://bugzilla.redhat.com/show_bug.cgi?id=1873547

PR #6349, opened 2020-11-09
add rule sshd_set_keepalive as a dependency

vojtapolasek:
#### Description:

- add rule sshd_set_keepalive to a profile in case it contains sshd_set_idle_timeout rule

#### Rationale:

The check of rule sshd_set_idle_timeout references rule sshd_set_keepalive. In case the whole profile is remediated and this rule is not present, the status of rule sshd_set_idle_timeout will be alway fail, because there does not exist any rule which would remediate the dependency.

### DIFF
--- a/fedora/profiles/pci-dss.profile
+++ b/fedora/profiles/pci-dss.profile
@@ -100,6 +100,7 @@ selections:
     - dconf_gnome_screensaver_lock_enabled
     - dconf_gnome_screensaver_mode_blank
     - sshd_set_idle_timeout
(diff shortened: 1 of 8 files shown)


ggbecker:
I opened a very similar pull request: #6348 

We can go with any of these pull requests... mine also change the example profile to include the rule... but it doesn't include the reference part in the stability test data for pci-dss.

ggbecker:
I have updated #6348 with the differences from here. No need to keep duplicated pull requests. Closing this one.

PR #6352, opened 2020-11-09
Add rule to configure group owner of /usr/bin/sudo

yuumasato:
#### Description:
- Add new rule `sudo_dedicated_group`, set by variable  `var_sudo_dedicated_group`.
  - The group is created if it doesn't exist. (I'm not sure if this is a good idea)
  - I've added warnings about the remediation. It may set up an unmanageable system, if:
    -  the dedicated group has no members;
    - root login is not allowed.

#### Rationale:
- Restricting group owner of `/usr/bin/sudo` helps reduce surface attack.
- Helps cover ANSSI R57


redhatrises:
@yuumasato can you fix the CCE conflict?

yuumasato:
CCE conflict resolved.

matejak (changes requested):
I have proposed an additional test as a separate commit, and some small modifications to the boilerplate code as suggestions.

Other than that, the PR looks just right.

matejak (approved):

matejak:
Great, thank you for the PR!

PR #6360, opened 2020-11-10
ocp4/CIS 1.1.(19-21): Add checks for Certificates and Keys

JAORMX:
This checks that the certificates and keys that lie on-disk and are used by the
kube control plane have the appropriate permissions and ownership.

Note that a regex is used to check for all certs and keys, so further additions
will get automatically picked up.

* OCP4/CIS 1.1.19 & 1.1.21: Add checks for OpenShift PKI keys
* OCP4/CIS 1.1.20 & 1.1.21: Add checks for OpenShift PKI cert files
* OCP4/CIS 1.1.20 & 1.1.21: Add checks for Etcd pki cert files

### DIFF
--- a/applications/openshift/master/file_owner_openshift_pki_cert_files/rule.yml
+++ b/applications/openshift/master/file_owner_openshift_pki_cert_files/rule.yml
@@ -23,7 +23,7 @@ ocil_clause: '{{{ ocil_clause_file_owner(file="/etc/kubernetes/static-pod-resour
 ocil: |-
   {{{ ocil_file_owner(file="/etc/kubernetes/static-pod-resources/.*/.*/.*/tls\.crt", owner="root") }}}
 
-# Note that this recursively checks for files, and the formas is as follows:
+# Note that this recursively checks for files, and the form is as follows:
 #
 #  /etc/kubernetes/static-pod-resources/.*/.*/.*/tls\.key
 #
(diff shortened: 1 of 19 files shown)


applications/openshift/master/file_owner_openshift_pki_cert_files/rule.yml
```diff
@@ -0,0 +1,45 @@
...
+title: 'Verify User Who Owns The OpenShift PKI Certificate Files'
+
+description: |-
+  {{{ describe_file_owner(file="/etc/kubernetes/static-pod-resources/.*/.*/.*/tls\.crt", owner="root") }}}
+
+rationale: |-
+  OpenShift makes use of a number of certificates as part of its operation.
+  You should verify the ownership of the directory containing the PKI
+  information and all files in that directory to maintain their integrity.
+  The directory and files should be owned by the system administrator.
```
mrogers950, line 14: Nit: Other rules say owned by root:root instead of system administrator, should we say root here also?

JAORMX: Gabe corrected me in another patch saying that "root:root" was too specific.

redhatrises: In general, there should be not statement of who a directory or files are to be owned by at all in the rationale.

JAORMX: @redhatrises got it, I'll remove it in a subsequent PR. Might want to add guidance on how to write these rationales to the docs.

PR #6361, opened 2020-11-10
ocp4/CIS: Address 1.2.24 and 1.2.25

JAORMX:
This addresses the aforementioned controls by checking the configMap for both
the kube apiserver and the openshift apiserver.

* ocp4/CIS 1.2.25: Add check for OpenShift APIServer audit-log-maxsize
* ocp4/CIS 1.2.25: Fix description and add check for audit-log-maxsize setting
* ocp4/CIS 1.2.24: Add check for OpenShift APIServer audit-log-maxbackup
* ocp4/CIS 1.2.24: Fix description and add rule for audit-log-maxbackup

### DIFF
--- a/applications/openshift/api-server/ocp_api_server_audit_log_maxsize/rule.yml
+++ b/applications/openshift/api-server/ocp_api_server_audit_log_maxsize/rule.yml
@@ -35,6 +35,10 @@ ocil: |-
     <pre>$ oc get configmap config -n openshift-apiserver -ojson | jq -r '.data["config.yaml"]' | jq '.apiServerArguments["audit-log-maxsize"]'</pre>
     The output should return a value of <pre>["100"]</pre> or as appropriate.
 
+warnings:
+- general: |-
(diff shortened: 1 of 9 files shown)


applications/openshift/api-server/ocp_api_server_audit_log_maxsize/rule.yml
```diff
@@ -0,0 +1,48 @@
...
+
+ocil: |-
+    Run the following command on the master node(s):
+    <pre>$ oc get configmap config -n openshift-apiserver -ojson | jq -r '.data["config.yaml"]' | jq '.apiServerArguments["audit-log-maxsize"]'</pre>
+    The output should return a value of <pre>["100"]</pre> or as appropriate.
+
+template:
+  name: yamlfile_value
+  vars:
+    ocp_data: "true"
```
mrogers950, line 41: With these you'll need the 'openshift_cluster_setting' warning section

JAORMX: yep! I do... these fails happen to pass because other rules take the same file into use. But if I would run them individually they wouldn't get the configmap on to the scan. Thanks for catching this!

PR #6362, opened 2020-11-10
ocp4/CIS: Remove file ownership sentences

JAORMX:
This enforces consistency in our rules and these sentences were not relevant for the rationale.

### DIFF
--- a/applications/openshift/master/file_groupowner_scheduler_kubeconfig/rule.yml
+++ b/applications/openshift/master/file_groupowner_scheduler_kubeconfig/rule.yml
@@ -11,7 +11,6 @@ rationale: |-
   The kubeconfig for the Scheduler contains paramters for the scheduler
   to access the Kube API.
   You should set its file ownership to maintain the integrity of the file.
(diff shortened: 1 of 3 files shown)


redhatrises:
The entire sentence should be removed and should not exist at all in the rationale.

JAORMX:
Pull-request updated, HEAD is now 2075b184c5036875a4bb6b6a81e869d817f43968

JAORMX:
> The entire sentence should be removed and should not exist at all in the rationale.

Done, thanks for the suggestion

JAORMX:
Pull-request updated, HEAD is now 6ce9a31cb599ea0333aaad1551f17447f022fcaf

redhatrises:
> > The entire sentence should be removed and should not exist at all in the rationale.
> 
> Done, thanks for the suggestion

Thanks. It just doesn't make sense to add a statement for what the expected configuration is when the rationale is for why the rule exists and what security considerations are in the rationale.

PR #6363, opened 2020-11-10
Add rule to drop sudo 'other' execution permisson

yuumasato:
#### Description:

- This rule removes execution permission for any user that is not the owner or member of group.
I.e.: chmod o-x /usr/bin/sudo

#### Rationale:
- Restricting group owner of `/usr/bin/sudo` helps reduce surface attack.
- Helps cover ANSSI R57


yuumasato:
Any tips for the rule ID?
I don't like much the current one.

ggbecker:
> Any tips for the rule ID?
> I don't like much the current one.

`restrict_sudo_binary_permissions`
`restrict_sudo_permissions`
`restrict_sudo_executable_permissions`
`sudo_executable_stricter_permissions`

Maybe you can get inspired by one of theses

linux_os/guide/system/software/sudo/sudo_no_others_execute_permission/rule.yml
```diff
@@ -0,0 +1,31 @@
+documentation_complete: true
+
+prodtype: rhel7,rhel8
+
+title: 'Ensure only owner and groupowners can excute sudo'
```
ggbecker, line 5:
```suggestion
title: 'Ensure only owner and groupowners can execute sudo'
```

linux_os/guide/system/software/sudo/sudo_no_others_execute_permission/rule.yml
```diff
@@ -0,0 +1,31 @@
...
+
+ocil_clause: '{{{ ocil_clause_file_permissions(file="/usr/bin/sudo", perms="---s--x---") }}}'
+
+ocil: '{{{ ocil_file_permissions(file="/usr/bin/sudo", perms="---s--x---") }}}'
+
+template:
+    name: file_permissions
+    vars:
+        filepath: /usr/bin/sudo
+        filemode: '4110'
```
ggbecker, line 31: Could it conflict with `rpm_verify_permissions` rule?

yuumasato: Hmm, yeah, it surely will.

redhatrises: > Could it conflict with `rpm_verify_permissions` rule?

Same thing with the other rule changing group ownership. The ultimate problem is that sudo is a setuid program. Wish that would change for the better, but alas, we are where we are.

yuumasato: Well, goal of this PR is not to solve the permission conflicts with another rule.

One way to mitigate this is to force evaluation of `rpm_verify_permission` before all other permission changing rules.
This way, the end result is the expected (permissions are reset to default and any deviation is done by the rules in the profile).

The Profile evaluation wont' ever converge, though. Every time a scan is done, the Profile will reset the file permissions according to RPM and then change file permissions according to the profile.
On a second though, I guess there is no way around this, unless we tracked all file permissions in the system, or maybe used AIDE to detect these kinds of changes...

yuumasato:
> > Any tips for the rule ID?
> > I don't like much the current one.
> 
> `restrict_sudo_binary_permissions`
> `restrict_sudo_permissions`
> `restrict_sudo_executable_permissions`
> `sudo_executable_stricter_permissions`
> 
> Maybe you can get inspired by one of theses

Thank you for the suggestions.
I went with `sudo_restrict_others_executable_permission`, as I wanted to make it explicit that the rule restricts others execution permission.

linux_os/guide/system/software/sudo/sudo_no_others_execute_permission/rule.yml
yuumasato: In 8048bf9 I force `rpm_verification` group to be first in the Benchmark.
This way the file permissions are reset to default state before the profile hardens them.

Is there any other rule that resets configuration to system default?

linux_os/guide/system/software/sudo/sudo_no_others_execute_permission/rule.yml
yuumasato: Fixed, thanks, :P

ssg/build_yaml.py
```diff
@@ -781,26 +781,33 @@ def to_xml_element(self):
...
-        # The FIPS group should come before Crypto - if we want to set a different (stricter) Crypto Policy than FIPS.
-        # the firewalld_activation must come before ruleset_modifications, othervise
-        # remediations for ruleset_modifications won't work
-        # rules from group disabling_ipv6 must precede rules from configuring_ipv6,
-        # otherwise the remediation prints error although it is successful
         priority_order = [
+            # Make sure rpm_verify_(hashes|permissions|ownership) are run before any other rule.
+            # Due to conflicts between rules rpm_verify_* rules and any rule that configures
+            # stricter settings, like file_permissions_grub2_cfg and sudo_dedicated_group,
+            # the rules deviating from the system detault should be evaluated later.
```
vojtapolasek, line 788:
```suggestion
            # the rules deviating from the system default should be evaluated later.
```

ssg/build_yaml.py
```diff
@@ -781,26 +781,33 @@ def to_xml_element(self):
...
+            # So that in the end the system has contents, permissions and ownership reset, and
+            # any deviations or stricter settings are applied by the rules in the profile.
+            "software", "integrity", "integrity-software", "rpm_verification",
+
+            # The account group has to precede audit group because
+            # the rule package_screen_installed is desired to be executed before the rule
+            # audit_rules_privileged_commands, othervise the rule
+            # does not catch newly installed screeen binary during remediation
```
vojtapolasek, line 796:
```suggestion
            # does not catch newly installed screen binary during remediation
```

vojtapolasek (changes requested):
Just small typos, everything looks good. I am still doing some tests if the change in ordering does not break anything. Did you test it on some profiles?

yuumasato:
>  Did you test it on some profiles?

I haven't done extensive testing, just evaluated RHEL-7 STIG profile and results seemed fine for me.

vojtapolasek:
I tested on rhel8 CIS and also no changes. Merging.

vojtapolasek (approved):

PR #6364, opened 2020-11-10
Fix bash_dconf_settings to grep whole keyword alike.

ggbecker:
#### Description:

- Fix `bash_dconf_settings` to grep whole keyword alike.

#### Rationale:

- Avoid cases where `automount-open` could be misinterpreted as a match for `automount` thus no applying correctly the fix.


mildas (approved):
LGTM, can be merged when tests are done.

PR #6365, opened 2020-11-10
Part 1 of remediation updates/additions for Java JRE

carlosmmatos:
#### Description:

Remediation updates and additions for Java JRE. More to come in another PR.

### DIFF
--- a/jre/guide/java/java_jre_blacklist_check/ansible/jre.yml
+++ b/jre/guide/java/java_jre_blacklist_check/ansible/jre.yml
@@ -1,0 +1,18 @@
+# platform = Java Runtime Environment
+# reboot = false
+# strategy = configure
+# complexity = low
+# disruption = low
+- name: Ensure Java Deployment Directory Exists
+  file:
+    path: /etc/.java/deployment
+    state: directory
+    mode: '0755'
+
+- name: Prevent the Download of Prohibited Mobile Code
+  lineinfile:
+    create: yes
+    mode: '0644'
+    dest: "/etc/.java/deployment/deployment.properties"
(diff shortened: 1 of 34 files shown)


jre/guide/java/java_jre_blacklist_check/ansible/jre.yml
```diff
@@ -0,0 +1,18 @@
+# platform = Java Runtime Environment
+# reboot = false
+# strategy = configure
+# complexity = low
+# disruption = low
+- name: Ensure Java Deployment Directory Exists
+  file:
+    path: /etc/.java/deployment
+    state: directory
+    mode: '0755'
```
redhatrises, line 10: Alternatively, you can `stat` the file and add a `when` statement to the next task so that the next task will run when the file exists. It can be sometimes better to take this approach rather than creating the file/directory every time. Thoughts?

carlosmmatos: For me, I feel like the `stat` is useful in certain scenarios where I might have a block option to run, but in this instance, this task will be minimal throughout the controls. Either way, you are still going to have 1 task run to determine this state. The `file` module is quick and will just verify that the directory exists, or else create it. No additional logic to be added.

carlosmmatos: Oh and also we can't assume that the `/etc/.java/deployment` will be there already - so this is another reason stat can't work here.

redhatrises: Alright. Sounds good.

PR #6366, opened 2020-11-10
Add or tidy up rules for CIS 1.2.4, 1.2.5 and 1.2.40

jhrozek:
Please see the commit messages for more details.

#### Description:

- All three controls touch the TLS settings of either of kube apiservers

#### Rationale:

- Increases the CIS benchmark coverage


applications/openshift/api-server/api_server_oauth_https_serving_cert/rule.yml
```diff
@@ -0,0 +1,50 @@
+documentation_complete: true
+
+prodtype: ocp4
+
+title: 'Make sure that the openshift-oauth-apiserver service uses TLS'
```
redhatrises, line 5:
```suggestion
title: 'Ensure the openshift-oauth-apiserver service uses TLS'
```

applications/openshift/api-server/api_server_openshift_https_serving_cert/rule.yml
```diff
@@ -0,0 +1,50 @@
...
+
+severity: medium
+
+references:
+    cis: 1.2.4
+
+ocil_clause: |-
+    Verify that the serving-cert for the openshift-apiserver is type
+    <tt>kubernetes.io/tls</tt> and that returned Data includes <tt>tls.crt</tt>
+    and <tt>tls.key</tt>.
```
redhatrises, line 29: The `ocil_clause` when fully built looks like:

```
Is it the case that Verify that the serving-cert for the openshift-apiserver
is type <tt>kubernetes.io/tls</tt> and that returned Data includes
<tt>tls.crt</tt> and <tt>tls.key</tt>?
```

which is a little weird, so would recommend something like:

```suggestion
    the openshift-apiserver serving-cert is not set to type
    <tt>kubernetes.io/tls</tt> and that returned Data does not include <tt>tls.crt</tt>
    and <tt>tls.key</tt>
```

applications/openshift/api-server/api_server_oauth_https_serving_cert/rule.yml
```diff
@@ -0,0 +1,50 @@
...
+
+severity: medium
+
+references:
+    cis: 1.2.4
+
+ocil_clause: |-
+    Verify that the serving-cert for the openshift-oauth-apiserver is type
+    <tt>kubernetes.io/tls</tt> and that returned Data includes <tt>tls.crt</tt>
+    and <tt>tls.key</tt>.
```
redhatrises, line 29: The `ocil_clause` when fully built looks like:

```
Is it the case that Verify that the serving-cert for the openshift-apiserver
is type <tt>kubernetes.io/tls</tt> and that returned Data includes
<tt>tls.crt</tt> and <tt>tls.key</tt>.?
```

which is a little weird, so would recommend something like:

```suggestion
    the openshift-apiserver serving-cert is not set to type
    <tt>kubernetes.io/tls</tt> and that returned Data does not include <tt>tls.crt</tt>
    and <tt>tls.key</tt>
```

applications/openshift/api-server/api_server_oauth_https_serving_cert/rule.yml
jhrozek: thanks

applications/openshift/api-server/api_server_openshift_https_serving_cert/rule.yml
jhrozek: Thank you, I used your suggestion. But I have two questions:

- are the `ocil_clause`s supposed to be written in a negative? I've seen some of them written in a positive ("...permissions are hardened...") and some of them in a negative ("...the permissions are open..."). Which is it?
- How do I view the ocil clauses rendered? To render the description and rationale I just view the built HTML guide (`build/guides/ssg-ocp4-guide-cis.html`) but I don't see the OCIL attributes there...

applications/openshift/api-server/api_server_oauth_https_serving_cert/rule.yml
jhrozek: fixed, thanks

jhrozek:
Thank you for the review, I pushed new patches with your suggestion.

JAORMX (approved):

applications/openshift/api-server/api_server_openshift_https_serving_cert/rule.yml
redhatrises: > Thank you, I used your suggestion. But I have two questions:
> 
> * are the `ocil_clause`s supposed to be written in a negative? I've seen some of them written in a positive ("...permissions are hardened...") and some of them in a negative ("...the permissions are open..."). Which is it?

There is no hard and fast rule as the answers to these questions are true/false. Generally, we have stuck to the negative more so than the positive. Basically, you are asking the auditor here to say "True, it is configured correctly" or "False, it isn't configured correctly".

> * How do I view the ocil clauses rendered? To render the description and rationale I just view the built HTML guide (`build/guides/ssg-ocp4-guide-cis.html`) but I don't see the OCIL attributes there...

Probably because the guides (purposefully) are XCCDF only. Currently, would have to check the OCIL file itself.

PR #6368, opened 2020-11-11
OCP4/CIS 4.2 test result additions

mrogers950:
Adds tests for more rules

### DIFF
--- a/applications/openshift/kubelet/kubelet_configure_tls_cert/tests/ocp4/e2e.yml
+++ b/applications/openshift/kubelet/kubelet_configure_tls_cert/tests/ocp4/e2e.yml
@@ -1,0 +1,2 @@
+---
+default_result: PASS
--- a/applications/openshift/kubelet/kubelet_disable_readonly_port/tests/ocp4/e2e.yml
+++ b/applications/openshift/kubelet/kubelet_disable_readonly_port/tests/ocp4/e2e.yml
@@ -1,0 +1,2 @@
+---
+default_result: PASS
--- a/applications/openshift/kubelet/kubelet_enable_cert_rotation/tests/ocp4/e2e.yml
+++ b/applications/openshift/kubelet/kubelet_enable_cert_rotation/tests/ocp4/e2e.yml
@@ -1,0 +1,2 @@
+---
+default_result: PASS
--- a/applications/openshift/kubelet/kubelet_enable_client_cert_rotation/tests/ocp4/e2e.yml
+++ b/applications/openshift/kubelet/kubelet_enable_client_cert_rotation/tests/ocp4/e2e.yml
@@ -1,0 +1,2 @@
+---
+default_result: PASS
--- a/applications/openshift/kubelet/kubelet_enable_iptables_util_chains/tests/ocp4/e2e.yml
+++ b/applications/openshift/kubelet/kubelet_enable_iptables_util_chains/tests/ocp4/e2e.yml
@@ -1,0 +1,2 @@
+---
+default_result: PASS
--- a/applications/openshift/kubelet/kubelet_enable_server_cert_rotation/tests/ocp4/e2e.yml
+++ b/applications/openshift/kubelet/kubelet_enable_server_cert_rotation/tests/ocp4/e2e.yml
@@ -1,0 +1,2 @@
+---
+default_result: PASS
--- a/applications/openshift/kubelet/kubelet_enable_streaming_connections/tests/ocp4/e2e.yml
+++ b/applications/openshift/kubelet/kubelet_enable_streaming_connections/tests/ocp4/e2e.yml
@@ -1,0 +1,2 @@
+---
+default_result: PASS


applications/openshift/kubelet/kubelet_protect_kernel_defaults/rule.yml
```diff
@@ -0,0 +1,41 @@
...
+
+description: |-
+    By default, the kubelet can adjust kernel parameters automatically.
+    If kernel default protection is enabled, the kubelet will instead
+    error if kernel parameters do not match what it expects.
+
+rationale: |-
+    OpenShift tunes the kernel and kubelet configurations dynamically. If
+    kernel default protection is enabled, changes could lead to the kubelet
+    crash-looping.
```
redhatrises, line 15: Okay, but what is the security rationale? So, for example, if the kubelet can change any kernel parameter, that means an attacker can figure out a way to disable kernel host protections through the kubelet.

applications/openshift/kubelet/kubelet_protect_kernel_defaults/rule.yml
```diff
@@ -0,0 +1,41 @@
+documentation_complete: true
+
+prodtype: ocp4
+
+title: 'Kubelet - Disable Kernel Default Protection'
+
+description: |-
+    By default, the kubelet can adjust kernel parameters automatically.
+    If kernel default protection is enabled, the kubelet will instead
+    error if kernel parameters do not match what it expects.
```
redhatrises, line 10: Question: If for some reason, it is set to True, how can I change this?

applications/openshift/kubelet/kubelet_protect_kernel_defaults/rule.yml
mrogers950: What we want to check for is that the kubelet _can_ adjust kernel parameters  - if this option is turned on for the kubelet, then a change in the kernel's parameters (not by the kubelet) can crash-loop the kubelet. The security rational in this case would be to avoid a DoS of the kubelet.

applications/openshift/kubelet/kubelet_protect_kernel_defaults/rule.yml
mrogers950: It's enforced by the platform, so unless there is an exposed knob there is not a persistent way to change it.

redhatrises: If it is enforced by the platform and cannot be changed say on install, update, etc, is there an actual reason to have this rule at all?

applications/openshift/kubelet/kubelet_protect_kernel_defaults/rule.yml
redhatrises: This is actually a bad thing if I am understanding you correctly. If the kubelet can can adjust kernel parameters, it means the kubelet can be used to attack a node, and an attacker can figure out how to hijack a kubelet to change kernel parameters to enable more insecure settings or disable secure settings.

mrogers950: Possibly, but that's not the intent of the protection recommendation in the draft. If it is turned on you probably won't have a running kubelet.

applications/openshift/kubelet/kubelet_protect_kernel_defaults/rule.yml
mrogers950: Perhaps not? It's still a valid option to the kubelet, but it should never appear in practice. There are other rules where we check for essentially deprecated flags

JAORMX: it's in the benchmark... I guess the intent is to not have folks wreck their deployments thinking it'll bring them extra security.

applications/openshift/kubelet/kubelet_protect_kernel_defaults/rule.yml
JAORMX: Let's change the rationale to explain to folks that not allowing the kubelet to tune kernel parameters could lead to the kubelet itself crashing and you losing availability in your deployment. Kernel parameters could tuned in such a way as to restrict a container even further... and when running a container that does security scanning (such as sysdig) you'd need to relax some tunables as well (thinking about sysctl's)

applications/openshift/kubelet/kubelet_protect_kernel_defaults/rule.yml
redhatrises: This is the question. Can you actually configure it in OCP. If so, yes this rule should be there. If there is no way, adding this rule makes no sense.

applications/openshift/kubelet/kubelet_protect_kernel_defaults/rule.yml
redhatrises: Utlimately, this needs to be fixed in OCP. A node should be flagged when kernel tunables are disabled to OCP so that it doesn't continue to be polled, and an alternative method through machineconfigs should set the tunables vs the kubelet doing it. As of this point, the security risk in OCP is for enabling or disabling this component. You enable it, you risk DOS and kubelet problems because of a couple of sysctls not set. You disable it, you risk node security by allowing a kubelet configure kernel parameters. As a point of reference, this is not a problem for other kube distros. For more security conscious customers and requirements, this is going to have to be fixed.

applications/openshift/kubelet/kubelet_protect_kernel_defaults/rule.yml
JAORMX: It's possible to configure as an "unsupportedOverride".

JAORMX: oh... actually this isn't even an unsupported override. This is settable and even supported via https://docs.openshift.com/container-platform/4.6/nodes/nodes/nodes-nodes-managing.html (it allows you to set almost any setting for the kubelet).

JAORMX: So yeah, it's relevant for us to check this.

JAORMX: We tested this in my local deployment and it makes the kubelet crash...

JAORMX: Upon further investigation, it seems that the arguments that the kubelet needs to set are not many... maybe we can fix this outside of the kubelet and disallow it from changing this.

redhatrises: @JAORMX it would be best to support preventing the DoS and the kubelet from making changes to the nodes. Otherwise, we open a vulnerability to close a vulnerability so to speak. If we can fix this outside of the kubelet, that would be amazing.

JAORMX: @redhatrises researching this. I'll probably come back with results later today.

JAORMX:
@mrogers950 I suggest we split this patch and handle 4.2.6 in its own PR.

JAORMX:
FTR, I've been researching this, and it seems possible to remediate (so far):

Remediating this on a running cluster requires a two step process:

1. Set the appropriate sysctl's via MachineConfig (and wait for the pools to update)
2. Set the kubelet config override via the KubeletConfig resource (this will create a MachineConfig which will update the pools too).

You can't do this in one step because kubelet runs before the MachineConfig daemon on normal operation. So if you would update the nodes with all the necessary objects in one go, the kubelet won't be able to start as it won't be able to change the sysctl's. However, if remediation is done in the order I mentioned, the sysctl's will already be there and the kubelet will start normally.

This is not a problem for newly provisioned nodes, because on first-boot, the MachineConfigDaemon runs before the kubelet.

mrogers950:
@JAORMX I dropped the rule commit

JAORMX:
Awesome. LGTM. Let's include the rule in a separate PR.

JAORMX (approved):

PR #6372, opened 2020-11-12
ocp4/CIS 1.2.13: Fix SecurityContextDeny check

JAORMX:
This also removes the templates differentiating between ocp versions, as we no longer need them.


JAORMX:
Pull-request updated, HEAD is now d6705254940245e5d666a98569f31567a188d2a3

PR #6373, opened 2020-11-12
Add RHEL7 STIG profile stability test source.

ggbecker:
#### Description:

- Add RHEL7 STIG profile stability test source.

#### Rationale:

- Any change in the RHEL7 STIG profile rule set willl be double checked by the profile stability test.


redhatrises:
Sorry, but what is the point of this test? It clearly doesn't indicate stability since a PR adding a rule to this profile and the non-test profile makes it impossible for "stability" tracking and adding a rule to a stability profile AND the actual profile is redundant.

ggbecker:
> Sorry, but what is the point of this test? It clearly doesn't indicate stability since a PR adding a rule to this profile and the non-test profile makes it impossible for "stability" tracking and adding a rule to a stability profile AND the actual profile is redundant.

The main reason of this test is to prevent undesired changes on profiles that extend other profiles. For example, if you add a rule in the parent profile it will automatically get into the child profile and that might be wanted.

The RHEL7 STIG profile doesn't extend any other profile, so this PR doesn't make much sense actually. Closing it.

PR #6377, opened 2020-11-12
Add documentation for variables option in test scenarios.

ggbecker:
#### Description:

- Add documentation for variables option in test scenarios.

#### Rationale:

- Follow up from #6323

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -345,10 +345,13 @@ The header consists of comments (starting by `#`). Possible values are:
 - `remediation` is a string specifying one of the allowed remediation types (eg. `bash`, `ansible`, `none`).
   The `none` value means that the tested rule has no implemented remediation.
 - `templates` has no effect at the moment.
+- `variables` is a comma-separated list of XCCDF values that sets a different default value for XCCDF variables.
 
 After the header, bash commands that prepare the scenario follow.
 
-Example of a scenario:
+Examples of test scenario:
+
+Using `platform` and `profiles` metadata:
 
 ```
 #!/bin/bash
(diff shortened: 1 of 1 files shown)


redhatrises:
Why is there this separate readme for testing over the existing documentation?

JAORMX (approved):

ggbecker:
> Why is there this separate readme for testing over the existing documentation?

I don't know the answer. But we have plans to incorporate it to: https://complianceascode.readthedocs.io/en/latest/ so it will be unified into a single place.

tests/README.md
```diff
@@ -345,10 +345,13 @@ The header consists of comments (starting by `#`). Possible values are:
 - `remediation` is a string specifying one of the allowed remediation types (eg. `bash`, `ansible`, `none`).
   The `none` value means that the tested rule has no implemented remediation.
 - `templates` has no effect at the moment.
+- `variables` is a comma-separated list of XCCDF values that sets a different default value for XCCDF variables.
```
matejak, line 348:
```suggestion
- `variables` is a comma-separated list of XCCDF values that sets a different default value for XCCDF variables in a form `<variable name>=<value>`. Typically, you use only one of `profile` or `variables` in scenario metadata - default values are effective only if the variable is not defined using a selector, which is what profiles do.
```

matejak:
> Why is there this separate readme for testing over the existing documentation?

This works nicely on Github, as it renders the README when you enter the `tests` folder. I think that it makes sense, moreover Becker promises seamless integration into the online documentation.

redhatrises:
tests have not bearing on this PR.

PR #6385, opened 2020-11-14
Add Ubuntu 20.04 support

Caligatio:
#### Description:

This patch adds support for Ubuntu 20.04 using the Standard profile.  I still have one or two lingering issues that need some help with which will be in separate comments/issues.

#### Rationale:

- Ubuntu 20.04 is completely absent from SSG


Caligatio:
I have most of the Ubuntu 18.04 "Standard" Profile implemented for 20.04 but have two lingering issues:

**Partition problems**
All of the partition checks get an "Unknown" status with an error in the form of
```
No items have been found conforming to the following objects:
Object oval:ssg-object_mountvar_log_audit_own_partition:obj:1 of type partition_object`
```
I unfortunately do not understand where this is originating from.

**sshd Checks**
All of my sshd checks are coming back as "notchecked" with an error of `No candidate or applicable check found.`  In the OVAL, it makes reference to `package_openssh-server_installed` but I don't really understand how the rule interacts with the OVAL.

Any help would be greatly appreciated!

Caligatio:
Ran into another issue that I don't quite know how to solve:

The `sysctl_fs_suid_dumpable` rule seems straight forward but I noticed that it failed on reboot.  [Googling around](https://estl.tech/ensuring-core-dumps-are-restricted-138f77759fc9),  I found that Ubuntu's error reporting services deliberately changes this value to "2" (we want it to be 1).  The fix is straight forward but I don't know how to integrate it given that this rule uses a template.

ggbecker:
> Ran into another issue that I don't quite know how to solve:
> 
> The `sysctl_fs_suid_dumpable` rule seems straight forward but I noticed that it failed on reboot. [Googling around](https://estl.tech/ensuring-core-dumps-are-restricted-138f77759fc9), I found that Ubuntu's error reporting services deliberately changes this value to "2" (we want it to be 0). The fix is straight forward but I don't know how to integrate it given that this rule uses a template.

You have to add the line:

`sysctlval@ubuntu2004: '2'`

here https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/permissions/restrictions/coredumps/sysctl_fs_suid_dumpable/rule.yml#L34

ggbecker:
@openscap-ci ok to test

Caligatio:
Sorry for the poor explanation @ggbecker; this isn't just a case of needing a different value for `sysctl`.  The problem is that there is an Ubuntu service (Apport) that "hijacks" the `suid_dumpable` value after `sysctl.conf` gets read.  The result is that you can't set `suid_dumpable = 0` without disabling Apport.  This means there are two options:

1. Get rid of this rule from the Standard Profile for Ubuntu
2. Add documentation that you need to add `enabled=0` to `/etc/default/apport` on Ubuntu installs.

I lean towards 1 since Ubuntu has decided to actively manage this setting.  2 is handy because it keeps the setting uniform across all profiles but this means that this rule cannot use the generic template (I think).

redhatrises:
> suid_dumpable

From a security & compliance perspective, it is better to do #2 rather than enable a setting. As vendors, we don't always implement the right security configurations by default.

Caligatio:
OK, I added a new rule that disables Apport.  Ideally, I'd like the `sysctl_fs_suid_dumpable` rule to do a `requires@ubuntu2004` but it appears you can't do "@" in the requires as the compile errors out saying there's unread values.

linux_os/guide/services/apport/service_apport_disabled/rule.yml
```diff
@@ -0,0 +1,23 @@
+documentation_complete: true
+
+title: 'Disable Apport Service'
+
+prodtype: ubuntu2004
+
+description: |-
+    The Apport modifies certain kernel configuration values at
+    runtime which conflicts with other hardening.
```
redhatrises, line 9:
```suggestion
    runtime which may decrease the overall security of the system and expose sensitive data.
```

ubuntu2004/transforms/constants.xslt
```diff
@@ -0,0 +1,21 @@
+<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
+
+<xsl:include href="../../shared/transforms/shared_constants.xslt"/>
+
+<xsl:variable name="product_long_name">Ubuntu 2004</xsl:variable>
+<xsl:variable name="product_short_name">Ubuntu 2004</xsl:variable>
+<xsl:variable name="product_stig_id_name">UBUNTU_FOCAL_FOSSA</xsl:variable>
```
redhatrises, line 7:
```suggestion
<xsl:variable name="product_stig_id_name">Canonical_Ubuntu_20-04</xsl:variable>
```

ubuntu2004/transforms/constants.xslt
```diff
@@ -0,0 +1,21 @@
...
+<xsl:variable name="product_guide_id_name">UBUNTU-FOCAL</xsl:variable>
```
redhatrises, line 8:
```suggestion
<xsl:variable name="product_guide_id_name">UBUNTU_20-04</xsl:variable>
```

konstruktoid:
Thanks for doing this @Caligatio.
Perhaps `UBUNTU2004` should be added to https://github.com/ComplianceAsCode/content/blob/master/build_product#L291-L292 as well.

linux_os/guide/services/ntp/package_timesyncd_installed/rule.yml
```diff
@@ -0,0 +1,29 @@
+documentation_complete: true
+
+prodtype: ubuntu2004
+
+title: 'Install the systemd_timesyncd service'
```
konstruktoid, line 5: Very much nitpicking, but `service` -> `Service` so it matches the other rules.

Caligatio:
@konstruktoid Changes made!  Apparently my grep'ing missed the `build_product` file :( 

Assuming this PR is accepted, my plan was to begin chipping away at the CIS Level 1 profiles this weekend and submit them as separate PRs.

redhatrises:
Merging as there is further work that is going on. Ack! Thanks for contributing!

PR #6387, opened 2020-11-16
Fix release tooling

vojtapolasek:
#### Description:

- while creating release notes, ignore files ending in .profile if they contain "tests" in their filepath.

- the version used in the name of a new commit which bumps the version after release was hardcoded. Now it is dynamically detected.

- while generating release notes, take into account only merged pull requests.

#### Rationale:

- generation of release notes ended with an exception if there was a PR which introduced change to some file ending in .profile and that file was not part of a product.

- closed but not merged pull requests were part of release notes

- the branch name for bumping of version after release was hardcoded and had to be changed manually, annoying.


jan-cerny:
I think it's fine

PR #6390, opened 2020-11-16
Make Bash Jinja macros documentation pretty

ggbecker:
#### Description:

- Make Bash Jinja macros documentation pretty.

#### Rationale:

- One more example of improved Jinja Macro documentation that can be used as inspiration.

  - Preview Before: https://complianceascode.readthedocs.io/en/latest/jinja_macros/bash.html#template-bash_package_install
  - Preview After: https://ggbecker-content.readthedocs.io/en/pretty-jinja-macros-doc/jinja_macros/bash.html#template-bash_package_install

### DIFF
--- a/shared/macros-bash.jinja
+++ b/shared/macros-bash.jinja
@@ -102,9 +102,12 @@ populate {{{ name }}}
 {{%- endmacro -%}}
 
 {{#
-# Install a package
-# Uses the right command based on pkg_manger proprerty defined in product.yaml.
-# package: name of the package
+    Install a package
+
+    Uses the right command based on pkg_manager property defined in product.yml.
+
+    :param package: name of the package
+    :type package: str
 #}}
 {{%- macro bash_package_install(package) -%}}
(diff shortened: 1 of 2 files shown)


jan-cerny:
It looks better. I like very much that you posted the example links. Great!

matejak:
It may be worth considering the [Napoleon extension](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html) for Google-style docstrings that are nice to read by humans as well. That extension is widely supported.

ggbecker:
> It may be worth considering the [Napoleon extension](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html) for Google-style docstrings that are nice to read by humans as well. That extension is widely supported.

I'm not sure if that would work since we are using the `sphinxcontrib.jinjadomain` which has its own parser

ggbecker:
@matejak see this: https://github.com/ggbecker/sphinxcontrib.jinjadomain/pull/1

PR #6392, opened 2020-11-17
OCP4/CIS 4.2.9: Update kubelet_configure_event_creation rule

mrogers950:
The benchmark does not recommend a setting of 0, but to a value that is appropriate for the scale of the deployment. This means any value would be appropriate (including the default of 5 when eventRecordQPS is omitted as per the cluster default). So here we check that there has at least been a configured value. The OCP4 kubelet default excludes eventRecordQPS, so the default compliance state for this item is a fail.


redhatrises:
@mrogers950 can we add a `Value` to this PR so that the value of `eventRecordQPS` is configurable by customers through tailoring?

redhatrises:
lgtm

PR #6400, opened 2020-11-19
OCP4/CIS 2.7: Check that etcd has its own CA

JAORMX:
This introduces a rule that checks that etcd is using its own CA which
is different from Kube's CA. It does this by using custom OVAL to check
that the two CA files are indeed different.

Since both components (etcd and the kube apiserver) only run on the
master nodes, this introduces a new cpe `ocp4-master-node`. This allows
us to differentiate between master nodes and other nodes. However, take
into account that it's on a node level and it isn't applicable to
"platform checks"

### DIFF
--- a/applications/openshift/etcd/etcd_unique_ca/rule.yml
+++ b/applications/openshift/etcd/etcd_unique_ca/rule.yml
@@ -5,17 +5,15 @@ prodtype: ocp4
 title: 'Configure A Unique CA Certificate for etcd'
 
 description: |-
-    A unique and different CA certificate should be created for <tt>etcd</tt>.
-    To ensure the <tt>etcd</tt> service is using a unique certificate,
-    set <tt>ETCD_TRUSTED_CA_FILE</tt> to <tt>/etc/etcd/ca.crt</tt>
-    in <tt>/etc/etcd/etcd.conf</tt> on the master node that does NOT match
-    the OpenShift CA certificate:
-    <pre>ETCD_TRUSTED_CA_FILE=/etc/ssl/etcd/ca.crt</pre>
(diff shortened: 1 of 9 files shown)


mrogers950:
LGTM!

JAORMX:
I'll come up with a way to set expectations on the role for specific results. This will have to wait for that solution.

JAORMX:
Pull-request updated, HEAD is now 20f692811184c7334b19ca70afa94631dc3f77e7

applications/openshift/etcd/etcd_unique_ca/rule.yml
```diff
@@ -5,30 +5,31 @@ prodtype: ocp4
 title: 'Configure A Unique CA Certificate for etcd'
 
 description: |-
-    A unique and different CA certificate should be created for <tt>etcd</tt>.
-    To ensure the <tt>etcd</tt> service is using a unique certificate,
-    set <tt>ETCD_TRUSTED_CA_FILE</tt> to <tt>/etc/etcd/ca.crt</tt>
-    in <tt>/etc/etcd/etcd.conf</tt> on the master node that does NOT match
-    the OpenShift CA certificate:
-    <pre>ETCD_TRUSTED_CA_FILE=/etc/ssl/etcd/ca.crt</pre>
+  A unique and different CA certificate should be created for <tt>etcd</tt>.
```
mrogers950, line 8: nit: this paragraph says 'unique and different' and also 'different and separate', I think we can just say separate in these cases

JAORMX: That came over from the previous content.

PR #6402, opened 2020-11-19
Abort the build if an OVAL is not included due to extend_definition

matejak:
If an extending definition is missing, then the OVAL won't be there as well.
This indicates an incorrect setup of applicability - either extending definitions are not applicable as they should, or the compound definition should not be applicable.

Fixes: #6185

### DIFF
--- a/ssg/build_renumber.py
+++ b/ssg/build_renumber.py
@@ -114,7 +114,6 @@ class OVALFileLinker(FileLinker):
 
     def __init__(self, translator, xccdftree, checks):
         super(OVALFileLinker, self).__init__(translator, xccdftree, checks)
-        self.tolerate_missing_extending_defs = False
         self.oval_groups = None
 
     def _get_checkid_string(self):
@@ -143,20 +142,12 @@ class OVALFileLinker(FileLinker):
 
         defs_miss = get_oval_checks_extending_non_existing_checks(self.tree, indexed_oval_defs)
         if defs_miss:
-            if self.tolerate_missing_extending_defs:
-                drop_oval_definitions(self.tree, defs_miss.keys(), self.oval_groups, indexed_oval_defs)
-                for broken, missed in defs_miss.items():
-                    print("WARNING: OVAL definition '{0}' extended by non-existing {1}, "
-                          "removing it from OVAL definitions."
-                          .format(broken.get("id"), list(missed)),
-                          file=sys.stderr)
-            else:
-                msg = ["Following extending definitions are missing:"]
-                for missing, broken in transpose_dict_with_sets(defs_miss).items():
-                    broken = [b.get("id") for b in broken]
-                    msg.append("\t'{missing}' needed by: {broken}"
-                               .format(missing=missing, broken=broken))
-                raise RuntimeError("\n".join(msg))
+            msg = ["Following extending definitions are missing:"]
+            for missing, broken in transpose_dict_with_sets(defs_miss).items():
+                broken = [b.get("id") for b in broken]
+                msg.append("\t'{missing}' needed by: {broken}"
+                           .format(missing=missing, broken=broken))
+            raise RuntimeError("\n".join(msg))
 
         self._add_cce_id_refs_to_oval_checks(xccdf_to_cce_id_mapping)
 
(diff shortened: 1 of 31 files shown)


matejak:
test this please

ggbecker:
your pull requested has detected that some `prodtypes` are missing in [disable_prelink](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/software/integrity/disable_prelink/rule.yml#L3)

it should be: `prodtype: rhcos4,ol7,rhel7,wrlinux1019`

matejak:
It was not so simple with OSP, as some OVALs are generated by templates, but it is resolved now.

ggbecker:
> It was not so simple with OSP, as some OVALs are generated by templates, but it is resolved now.

Great. But now there is some issue with OCP4 as it only build on fedora and when using certain version of openscap

ocp4/product.yml
```diff
@@ -3,6 +3,10 @@ full_name: Red Hat OpenShift Container Platform 4
 type: platform
 
 benchmark_root: "../applications"
+additional_content_directories:
+  - "../linux_os/guide/system/software/integrity/endpoint_security_software/mcafee_security_software"
+  - "../linux_os/guide/system/network/network-ipv6/disabling_ipv6"
+  - "../linux_os/guide/system/selinux"
```
JAORMX, line 9: This doesn't sound right... why are these being included?

JAORMX:
> > It was not so simple with OSP, as some OVALs are generated by templates, but it is resolved now.
> 
> Great. But now there is some issue with OCP4 as it only build on fedora and when using certain version of openscap

Right, ocp4 content needs openscap 1.3.4

ocp4/product.yml
matejak: That's a good catch. So we have some "shared" OVALs that don't belong to any rules. One would expect them to be included in the benchmark if they are needed by rule checks, but that's not the case - they are included unconditionally. Those incriminated definitions are the `` and `` ones.
The quick fix of these cases would be to set "affected platform" of those snippets to Linux OSs only, as they don't make sense for OCP, OSP and other application-level products.

In the longer run, we should do something about our messed-up system of platforms vs prodtypes vs products.
Finally, those shared checks should be included only on-demand, and there should be a distinction between an applicability of the check and whether the check is actually needed by some other check or by an existing rule.

JAORMX:
@matejak using `/retest` is only applicable for the OpenShift CI. But that CI is passing. the ansible role test is handled differently and I'm not sure what command you'd need to run there.

matejak:
RHV has been affected as well - it is just a flavor of RHEL after all.

ssg/build_renumber.py
```diff
@@ -139,8 +141,22 @@ def _link_oval_tree(self):
         indexed_oval_defs = map_elements_to_their_ids(
             self.tree, ".//{0}".format(self._get_checkid_string()))
 
-        drop_oval_checks_extending_non_existing_checks(
-            self.tree, self.oval_groups, indexed_oval_defs)
+        defs_miss = get_oval_checks_extending_non_existing_checks(self.tree, indexed_oval_defs)
+        if defs_miss:
+            if self.tolerate_missing_extending_defs:
```
jan-cerny, line 146: This is never set which means this block is never executed. What are the plans?

matejak: There is no plan - the code is just prepared for the possibility that missing extending definitions become tolerable.

jan-cerny: Aha, and can it ever happen that it will be tolerable? What is the scenario in which it would be used?

matejak: Hard to say - you know how it is with assumptions. There must have been some reason why it is tolerated since it was introduced.

jan-cerny: But it's you who introduces it so you should be able to explain why you introduce it.

matejak: I have removed the conserved code in a commit that can be reverted in case that there is an interest to build content that has "definition holes".

jan-cerny:
@matejak please refresh my mind about why this PR is hanging

matejak:
This PR can prevent products with inconsistent OVAL content from building :-( Perhaps we could start introducing it after the release.

matejak:
The PR doesn't break any builds now, so please @jan-cerny give it a go.

JAORMX:
/retest

CI failed gathering logs; it's unrelated to the CaC tests.

PR #6404, opened 2020-11-20
add rule ssh_client_use_strong_rng

vojtapolasek:
#### Description:

- add rule, remediations, tests

- add rule to ospp profile

- update profile stability tests

#### Rationale:

Common criteria effort - fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1884067


redhatrises:
Since these rules apply to ssh, please add them under the ssh services group as they apply to configuring ssh.

linux_os/guide/system/software/integrity/crypto/ssh_client_use_strong_rng/oval/shared.xml
```diff
@@ -0,0 +1,40 @@
...
+    <ind:instance datatype="int" operation="greater than or equal">1</ind:instance>
+  </ind:textfilecontent54_object>
+
+  <ind:textfilecontent54_test check="all" check_existence="all_exist" comment="check correct entropy configuration in /etc/profile.d/cc-ssh-strong-rng.csh" id="test_ssh_client_strong_rng_csh" version="1">
+    <ind:object object_ref="obj_ssh_client_strong_rng_csh"/>
+    <ind:state state_ref="state_ssh_client_strong_rng" />
+  </ind:textfilecontent54_test>
+
+  <ind:textfilecontent54_object id="obj_ssh_client_strong_rng_csh" version="1">
+    <ind:filepath>/etc/profile.d/cc-ssh-strong-rng.csh</ind:filepath>
```
redhatrises, line 29: You also need to verify that /etc/profile is set correctly for this configuration. Otherwise, it can be overwritten and the scanner wouldn't know about it at all. That's a bad thing.

linux_os/guide/system/software/integrity/crypto/ssh_client_use_strong_rng/rule.yml
```diff
@@ -0,0 +1,44 @@
...
+title: 'SSH client uses strong entropy to seed'
+
+description: |-
+    To set up SSH client to use entropy from a high-quality source, make sure
+    that appropriate shell environment variables are configured. The
+    <tt>SSH_USE_STRONG_RNG</tt> environment variable determines how many bytes
+    of entropy to use. Make sure that the file
+    <tt>/etc/profile.d/cc-ssh-strong-rng.sh</tt> contains line
+    <pre>export SSH_USE_STRONG_RNG=32</pre>
+    and that the file <tt>/etc/profile.d/cc-ssh-strong-rng.csh</tt> contains line
```
redhatrises, line 14: there are 2 configuration changes here which means there needs to be separate rules for them.

linux_os/guide/system/software/integrity/crypto/ssh_client_use_strong_rng/rule.yml
```diff
@@ -0,0 +1,44 @@
...
+    <pre>setenv SSH_USE_STRONG_RNG 32</pre>.
+
+rationale: |-
+    SSH implementation in RHEL8 uses the openssl library, which doesn't use high-entropy sources by default.
+    Randomness is needed to generate data-encryption keys, and as plaintext padding and initialization vectors
+    in encryption algorithms, and high-quality entropy elliminates the possibility that the output of
+    the random number generator used by SSH would be known to potential attackers.
```
redhatrises, line 21:
```suggestion
    Some SSH implementations use the openssl library for entropy, which by default, doesn't use high-entropy sources.
    Randomness is needed to generate considerably more secure data-encryption keys. Plaintext padding, initialization vectors
    in encryption algorithms, and high-quality entropy eliminates the possibility that the output of
    the random number generator used by SSH would be known to potential attackers.
```

linux_os/guide/system/software/integrity/crypto/ssh_client_use_strong_rng/rule.yml
```diff
@@ -0,0 +1,44 @@
...
+severity: medium
+
+identifiers:
+    cce@rhel8: CCE-83346-7
+
+references:
+    ospp: FCS_CKM.1.1
+    srg: SRG-OS-000480-GPOS-00227
+
+ocil_clause: 'SSH client is not configured to use 32 bytes of entropy'
```
redhatrises, line 32:
```suggestion
ocil_clause: 'SSH client is not configured correctly or is not configured to use at least 32 bytes of entropy'
```

vojtapolasek:
I have splitted the rule into two and created a special group for SSH client rules. I have also implemented check that the variable is not overridden in /etc/profile. Also numbers >= 32 are allowed.

matejak:
Conversations are resolved, the rule LGTM (there are two rules because of the one configuration = one rule convention), so merging.
Thank you for this addition!

PR #6405, opened 2020-11-20
Introduce test suite script wrappers

matejak:
This PR introduces a script that builds the test suite Fedora Podman container, and another script that executes the test suite in container mode, and passes it flags that remove disadvantages of the (Fedora) container backend, so a user that has `podman` can test a rule by performing two steps as simple as one can imagine.
The test suite execution wrapper adds option that neutralize the machine-only platform restriction by passing `--remove-machine-only`, and the platform of the benchmark is extended based on the OS of the testing image using the `--add-platform` option.

Next, when legitimately testing machine-only rules on containers, it is not enough to make checks work - it is needed to enable remediations as well.
In order to do so, the machine platform checking condition is replaced by a statement that mocks bare-metal systems. This needed to be introduced after #6061


tests/test_rule_in_container.sh
```diff
@@ -0,0 +1,141 @@
...
+public_key=$_arg_public_key
+podman images | grep -q "$_arg_name" || die "Couldn't find the podman image '$_arg_name'"
+test_image_cpe_product=$(podman run --rm "$_arg_name" cat /etc/os-release | grep cpe | cut -d : -f 4)
+test -n "$test_image_cpe_product" || die "Unable to deduce the product CPE from the container's /etc/os-release file."
+
+scenario_args=()
+test -n "$_arg_scenarios" && scenario_args=(--scenario "$_arg_scenarios")
+
+# build without context
+python ${script_dir}/test_suite.py rule --remove-machine-only "${scenario_args[@]}" --add-platform "$test_image_cpe_product" --container "$_arg_name" -- "${_arg_rule}"
```
jan-cerny, line 139: I wonder why it doesn't have `--datastream` to choose the datastream. I have found that `test_suite.py` has some sort of datastream autodetection. There is a message that it works only when there is single datastream in the build directory. I wasn't aware of it and it isn't documented. Would it be better to be explicit about the datastream?

matejak: Indeed, it's not documented - I will fix that one. OTOH, I think that as it is an optional argument, it's OK that there is a default that works.
I am not against propagating the option to the wrapper, but the idea is to build and test with minimal effort, so I see the need to specify a datastream as an exceptional case.

jan-cerny: That's it - up until now I think that it is a required argument, it's in all examples in the readme, so I wasn't aware about the convenience!

But thanks for explanation, it's probably fine to not have it :D

jan-cerny:
@matejak It's great! Overall it looks good to me. Please fix the PEP8 problems before I merge it.

PR #6409, opened 2020-11-23
SRG for ssh_client_rekey_limit

mildas:
#### Description:
Add SRG identifier https://vaulted.io/library/disa-stigs-srgs/general_purpose_operating_system_srg/V-56861 to `ssh_client_rekey_limit` rule.

#### Rationale:
The rule is in the ospp profile and stig extends ospp.


redhatrises:
I think that SRG-OS-000423-GPOS-00187 is a better fit here.

mildas:
Sure. It's difficult to find best fit. I've updated the SRG it to your suggestion.

PR #6410, opened 2020-11-23
remove extended check from accounts_passwords_pam_faillock_enforce_local

vojtapolasek:
#### Description:

- remove extending definition  checking for accounts_password_pam_faillock from oval check of accounts_passwords_pam_faillock_enforce_local

#### Rationale:

- the extending check is not directly connected with the rule description. Usage of faillock should be verified with extra rule.

### DIFF
--- a/linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_passwords_pam_faillock_enforce_local/oval/shared.xml
+++ b/linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_passwords_pam_faillock_enforce_local/oval/shared.xml
@@ -2,7 +2,6 @@
   <definition class="compliance" id="accounts_passwords_pam_faillock_enforce_local" version="1">
     {{{ oval_metadata("Failed password attempts are enforced for local users only.") }}}
     <criteria operator="AND" comment="conditions for accounts_passwords_pam_faillock_enforce_local are satisfied">
-      <extend_definition comment="faillock.so exists in system-auth" definition_ref="accounts_password_pam_faillock" />
       <criterion comment="faillock.conf" test_ref="test_accounts_passwords_pam_faillock_enforce_local" />
     </criteria>
   </definition>


redhatrises:
Sorry but nack The rule does not have to describe the fact that pam_faillock has to exist in pam for this to work. All of our pam settings work this way.

vojtapolasek:
> Sorry but nack The rule does not have to describe the fact that pam_faillock has to exist in pam for this to work. All of our pam settings work this way.

I understand. The problem is that this is the ONLY rule which uses this extended check. Other rules related to faillock do not. And actually the remediation of this rule does NOT do anything influencing this extended check. So the solution could be to add this to the rule remediation. The check basically wants the line
password required pam_faillock.so
or
password requisite pam_faillock.so
in /etc/pam.d/system-auth
But it is not written anywhere how to remediate this. At which position in the PAM stack should the line be presennt? I would expect that the rule accounts_passwords_pam_faillock_enforce_local would care only about the file /etc/security/faillock.conf and no other files. One rule - one configuration. Any suggestions?

redhatrises:
Please reference the pwquality checks for examples of this... the existing pam faillock rules are now wrong in rhel8 and should be changed to follow how these checks are done. If pam isn't configured correctly, the check doesn't work. The check is currently correct and this PR is introducing a bug. See pwquality rules for how the faillock rules will be done in the future.

vojtapolasek:
I am not saying that the check is wrong, but I am saying that I don't know how to remediate the rule so that this check starts passing. Currently pwquality rules do NOT remediate the files in a way which would influence the result of the extending check. So we will have to come up with something. Or I see that there might be rule dependency.

yuumasato:
The existing rules for `pam_faillock` enable and configure it at the same time, as the options and parameters were added together with the module in the stack.

The new approach of using `/etc/security/faillock` is much simpler to check and remediate, no doubt.
But enabling and configuration of the module happen in different places now, so they should be made separate rules.
And AFAIK there is no rule or remediation that just enables `pam_faillock.so` in `/etc/pam.d/system-auth`.

PR #6411, opened 2020-11-23
ocp4: Add automatic remediation for etcd encryption provider

JAORMX:
Instead of this relying on a manual remediation, we should instead
create an automated one that the compliance-operator can deploy.

Depends on: https://github.com/openshift/compliance-operator/pull/487

### DIFF
--- a/applications/openshift/api-server/api_server_encryption_provider_config/kubernetes/shared.yml
+++ b/applications/openshift/api-server/api_server_encryption_provider_config/kubernetes/shared.yml
@@ -1,0 +1,8 @@
+# platform = multi_platform_ocp
+apiVersion: config.openshift.io/v1
+kind: APIServer
+metadata:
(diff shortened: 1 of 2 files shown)


redhatrises:
@openscap-ci test this please

JAORMX:
@redhatrises we need to add extra permissions to the compliance-operator for this to work. The patch just merged, so now we just need to do an upstream release for it to get picked up here.

JAORMX:
Still haven't done the upstream release. We're investigating some performance issues. I'll update here once that's done

JAORMX:
Pull-request updated, HEAD is now 589aabca902d84e96f5d33cf823767e5c70385ae

JAORMX:
Pull-request updated, HEAD is now 7b5809d79a95ed7992fc712ad24b5dff2cad5bbd

PR #6412, opened 2020-11-23
Rename .github/issue_template folder to .github/ISSUE_TEMPLATE

ggbecker:
Github expects it to be uppercase, otherwise it doesn't work.

A follow up from: https://github.com/ComplianceAsCode/content/pull/6395

PR #6413, opened 2020-11-23
Use correct parameter for name field in issue template.

ggbecker:
#### Description:

- Use correct parameter for name field in issue template.

#### Rationale:

- Github will now be able to process this file.


ggbecker:
The issue has been uncovered after #6412 got merged. Merging this as well since there is no code change and we don't need for tests, same reason as in #6412

PR #6286, opened 2020-10-21
Add bios and uefi CPE applicability for grub2 rules

ggbecker:
#### Description:

- Create two new groups to hold uefi grub2 and non-uefi grub2 related rules.
- Add new CPE applicability check based on existing check for UEFI

#### Rationale:

- Checks return not applicable instead of passing if checking a UEFI rule on a Non-UEFI based system for example.
- No need to have a custom check with extended definition for UEFI

- Relates #6028

PR #6414, opened 2020-11-24
Move uefi_no_removeable_media tests to correct place.

ggbecker:
#### Description:

- Move uefi_no_removeable_media tests to correct place.

#### Rationale:

- Omission from #6286


jan-cerny:
Nice

PR #6415, opened 2020-11-25
make rule kernel_trust_cpu_rng grub2 only

vojtapolasek:
#### Description:

- change platform to from machine to grub2

- moved rule to more appropriate group

- rename the rule

#### Rationale:

- the change of platform was needed because the rule was applied also on s390 with zipl bootloader, that does not make sense

- change of group and name reflect the rule purpose better. Potential (although I believe small) chance of breaking already existing tayloring. Opinions?


redhatrises:
merging... changes are rhel only.

PR #6416, opened 2020-11-25
WIP: Implement CIS Level 1 for Ubuntu 20.04

Caligatio:
#### Description:

This is a work in progress PR to add the CIS Level 1 profile for Ubuntu 20.04

#### Rationale:

I initially implemented the "Standard" profile for Ubuntu 20.04 by copying over the existing profile for Ubuntu 18.04  Unfortunately there were only a few tests so this PR should bring Ubuntu 20.04 up to par with the other supported distributions.


ggbecker:
@Caligatio I took the liberty to update the conflicts and to change the code so it now builds correctly. Overall the pull request looks good. Please take a look at the individual commits to understad the changes that were required.

Caligatio:
Looks good to me.  I stalled a bit on porting more rules but hope to get back into it later this month.

More of a note to myself, I'll have to fiddle a bit with `linux_os/guide/system/software/updating/security_patches_up_to_date/rule.yml` to bring it inline with the generalized CVE URL.

ggbecker:
@openscap-ci test this please

linux_os/guide/system/software/disk_partitioning/partition_for_tmp/rule.yml
```diff
@@ -21,9 +21,7 @@ identifiers:
 references:
     stigid@ol7: OL07-00-021340
     anssi: BP28(R12)
-    cis@rhel8: 1.1.2
-    cis@ubuntu1804: 1.1.2
-    cis@ubuntu2004: 1.1.2
+    cis: 1.1.2
```
yuumasato, line 24: Tempting as it may be to unify these references, I lean towards not merging them.
There is no guarantee that CIS profiles added in the future will match these references numbers.

Caligatio: Wouldn't it make sense to unify the references and then just have the outliers be overrides? This is done is many, many places with references defaulting to RHEL and then @ubuntu-whatever being overrides.

redhatrises: Yes, it does make sense to unify references until there are deviations, and they should always be unified unless it deviates. Deviations are exceptions not the rule.

yuumasato: I mainly would like to avoid inconsistencies.
If the reference is not product specific it will show up in every product, even for products that don't have a CIS profile implemented.
Another point is that when adding a new CIS profile to a product, one will have to process every rule with a `cis` reference in it, either to accept the reference or remove it.

yuumasato: > This is done is many, many places with references defaulting to RHEL and then @ubuntu-whatever being overrides.

Well, yes, maybe it would not be a problem to unify them.

> Deviations are exceptions not the rule.

When it comes to sharing descriptions, checks and fixes, it makes sense. A lot of code and practices is shared among distros.
I know that CIS benchmarks have some kind of hierarchy and relationship between them, So unifying them may not be a big problem, and my concerns are not relevant.

linux_os/guide/system/permissions/files/permissions_important_account_files/file_groupowner_backup_etc_group/rule.yml
```diff
@@ -16,8 +16,7 @@ identifiers:
     cce@rhel8: CCE-83475-4
 
 references:
-    cis@rhel7: 6.1.8
-    cis@rhel8: 6.1.8
+    cis: 6.1.8
```
cipherboy, line 19: Since other section `6.1.x` rules differ between RHEL and Ubuntu, I might suggest we keep these two separate for the time being.

linux_os/guide/system/permissions/files/permissions_important_account_files/file_groupowner_backup_etc_passwd/rule.yml
```diff
@@ -16,8 +16,7 @@ identifiers:
     cce@rhel8: CCE-83324-4
 
 references:
-    cis@rhel7: 6.1.6
-    cis@rhel8: 6.1.6
+    cis: 6.1.6
```
cipherboy, line 19: As above.

linux_os/guide/system/permissions/files/permissions_important_account_files/file_groupowner_backup_etc_shadow/rule.yml
```diff
@@ -22,8 +22,7 @@ identifiers:
     cce@rhel8: CCE-83415-0
 
 references:
-    cis@rhel7: 6.1.7
-    cis@rhel8: 6.1.7
+    cis: 6.1.7
```
cipherboy, line 25: As above.

linux_os/guide/system/permissions/files/permissions_important_account_files/file_owner_backup_etc_group/rule.yml
```diff
@@ -16,8 +16,7 @@ identifiers:
     cce@rhel8: CCE-83473-9
 
 references:
-    cis@rhel7: 6.1.8
-    cis@rhel8: 6.1.8
+    cis: 6.1.8
```
cipherboy, line 19: Same.

linux_os/guide/system/permissions/mounting/kernel_module_cramfs_disabled/rule.yml
```diff
@@ -27,9 +27,7 @@ identifiers:
     cce@rhcos4: CCE-82514-1
 
 references:
-    cis@rhel8: 1.1.1.1
-    cis@sle15: 1.1.1.1
-    cis@ubuntu1804: 1.1.1.1
+    cis: 1.1.1.1
```
cipherboy, line 30: cramfs remains the same, but 1.1.1.x change between 18.04 and 20.04, so I'm inclined to leave these separate.

linux_os/guide/system/software/updating/security_patches_up_to_date/rule.yml
```diff
@@ -21,6 +21,9 @@ description: |-
     If the system is configured for online updates, invoking the following command will list available
     security updates:
     <pre>$ sudo zypper refresh &amp;&amp; sudo zypper list-patches -g security</pre>
+{{% elif product in ["ubuntu2004"] %}}
```
cipherboy, line 24: I'd add both other Ubuntu versions while you're here.

linux_os/guide/system/software/disk_partitioning/partition_for_tmp/rule.yml
cipherboy: I'm fine with either, but the draft is a living specification and it often depends on what version of a particular OS spec you're looking at (e.g., v1.1.0 of the Ubuntu 20.04 spec changes a number of section numbers than v1.0.0). 

My concern is someone going ahead and changing the `cis` reference because, e.g., RHEL changed, but e.g., Ubuntu didn't, without looking at all products the rule currently applies to.

If we make deviations the exception, not the rule, we need to make sure updating the default involves more cross-checking. :-)

My 2c.

ubuntu2004/profiles/cis_level1_server.profile
```diff
@@ -0,0 +1,653 @@
...
+    - mount_option_tmp_nodev
+
+    ### 1.1.4 Ensure nosuid option set on /tmp partition (Automated)
+    - mount_option_tmp_nosuid
+
+    ### 1.1.5 Ensure noexec option set on /tmp partition (Automated)
+    - mount_option_tmp_noexec
+
+    ### 1.1.6 Ensure /dev/shm is configured (Automated)
+    # Skip due to being handled by systemd and ensured by follow-on tests
```
cipherboy, line 47: Yeah, it'd be nice to have an entry here for `/dev/shm` (and also `/tmp`) to have it configured in `systemd` correctly, rather than adding `/etc/fstab` changes.

ubuntu2004/profiles/cis_level1_server.profile
```diff
@@ -0,0 +1,653 @@
...
+    - mount_option_dev_shm_nosuid
+
+    ### 1.1.9 Ensure noexec option set on /dev/shm partition (Automated)
+    - mount_option_dev_shm_noexec
+
+    ### 1.1.10 Ensure separate partition exists for /var (Automated)
+    # Skip due to being Level 2
+
+    ### 1.1.11 Ensure separate partition exists for /var/tmp (Automated)
+    # This is a Level 2 test but is needed for 1.1.12-14
```
cipherboy, line 62: I filed ticket#12460 at CIS for tracking this.

ubuntu2004/profiles/cis_level1_server.profile
```diff
@@ -0,0 +1,653 @@
...
+
+    ## 1.3 Configure sudo ##
+    ### 1.3.1 Ensure sudo is installed (Automated)
+
+    ### 1.3.2 Ensure sudo commands use pty (Automated)
+
+    ### 1.3.3 Ensure sudo log file exists (Automated)
+
+    ## 1.4 Filesystem Integrity Checking ##
+    ### 1.4.1 Ensure AIDE is installed (Automated)
```
cipherboy, line 120: Just in case you're interested, I think [`package_aide_installed`](https://github.com/ComplianceAsCode/content/tree/master/linux_os/guide/system/software/integrity/software-integrity/aide/package_aide_installed) together with [`aide_build_database`](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/software/integrity/software-integrity/aide/aide_build_database/rule.yml) would work here.

Though, perhaps `aide_build_database` should be updated to what's in the current guidance:

> ```console
> # aideinit
> # mv /var/lib/aide/aide.db.new /var/lib/aide/aide.db
> ```

ubuntu2004/profiles/cis_level1_server.profile
```diff
@@ -0,0 +1,653 @@
...
+
+    ### 1.4.2 Ensure filesystem integrity is regularly checked (Automated)
```
cipherboy, line 122: Same as above; perhaps best to leave as CIS benchmark allows either cron or systemd timers, but we do have [`aide_periodic_cron_checking`](https://github.com/ComplianceAsCode/content/tree/master/linux_os/guide/system/software/integrity/software-integrity/aide/aide_periodic_cron_checking).

ubuntu2004/profiles/cis_level1_server.profile
```diff
@@ -0,0 +1,653 @@
...
+
+    ## 1.5 Secure Boot Settings ##
+    ### 1.5.1 Ensure bootloader password is set (Automated)
```
cipherboy, line 125: For future reference, there are some [`bootloader-grub2`](
https://github.com/ComplianceAsCode/content/tree/master/linux_os/guide/system/bootloader-grub2) tests available as well.

cipherboy (approved):
I think this is a good start. Obviously needs to be rebased on top of latest HEAD. I think I agree with @yuumasato's original assessment that its better to leave CIS references alone (and separated out per target). This way too, CIS references don't appear on an OS that lacks a CIS benchmark.

Other than that, there's more rules that could be added into later sections (that already exist in CaC), but this is mergeable as-is and later PRs can always add more rules incrementally.


Looks great, thanks!

linux_os/guide/system/software/disk_partitioning/partition_for_tmp/rule.yml
yuumasato: @cipherboy So you are slightly in favor of defaulting to specific `cis` references then?

cipherboy: Yeah, on the other side as an intern back in '18 I definitely would've preferred a single `cis` reference with OS-specific overrides when necessary.

But now having seen more of the CIS process, I don't think that is correct, as helpful as it would be ;-) 

Each OS needs to go through CIS process and have a spec published before a CIS reference becomes valid for it. So having no default `cis` reference makes it explicit and obvious that this has been done for a given OS/rule combination. 

My 2c.

cipherboy:
Closing this in favor of #7181 as that is a completed set of CIS profiles. @Caligatio You have been credited as co-author on several commits both internally (during development and rebases) and on that pull request. Thanks for all your help getting the initial profiles started! Feel free to contribute to discussion on that pull request and/or upstream in subsequent pull requests once it merges.

PR #6419, opened 2020-11-25
Moved OVAL CVE Feed metadata from the rule to individual products.

matejak:
This change brings following benefits:

- Changes of feeds don't require to touch the rule file.
- The build aborts if a feed doesn't start with https.
- Rule gets XCCDF warnings if there is no feed, or if the feed doesn't end with `.xml`.

### DIFF
--- a/linux_os/guide/system/software/updating/security_patches_up_to_date/rule.yml
+++ b/linux_os/guide/system/software/updating/security_patches_up_to_date/rule.yml
@@ -74,5 +74,5 @@ warnings:
 {{% else %}}
 {{# The rule will be "notchecked" #}}
 warnings:
-    - general: "{{{ full_name }}} doesn't have a OVAL CVE Feed associated, so this rule doesn't have a corresponding check."
+    - general: '{{{ full_name }}} does not have a corresponding OVAL CVE Feed. Therefore, this will result in a "not checked" result during a scan.'
 {{% endif %}}
--- a/linux_os/guide/system/software/updating/security_patches_up_to_date/rule.yml
+++ b/linux_os/guide/system/software/updating/security_patches_up_to_date/rule.yml
@@ -64,9 +64,6 @@ references:
 # SCAP 1.3 content should reference flat non compressed xml files
 {{% if oval_feed_url %}}
 oval_external_content: {{{ oval_feed_url }}}
-  {{% if not oval_feed_url.startswith("https") %}}
-  {{{ raise("OVAL feed of product '" + product + "' is not available through an encrypted channel: " + oval_feed_url) }}}
-  {{% endif %}}
   {{% if not oval_feed_url.endswith(".xml") %}}
 warnings:
     - general: "The OVAL feed of {{{ full_name }}} is not a XML file, which may not be understood by all scanners."
--- a/linux_os/guide/system/software/updating/security_patches_up_to_date/rule.yml
+++ b/linux_os/guide/system/software/updating/security_patches_up_to_date/rule.yml
@@ -60,21 +60,19 @@ references:
     cis-csc: 18,20,4
     anssi: BP28(R08)
 
-{{# Make sure all the external OVAL content links are secured via TLS! #}}
 
 # SCAP 1.3 content should reference flat non compressed xml files
-{{% if product == "rhel7" %}}
-oval_external_content: "https://www.redhat.com/security/data/oval/com.redhat.rhsa-RHEL7.xml"
-{{% elif product == "rhel8" %}}
-oval_external_content: "https://www.redhat.com/security/data/oval/com.redhat.rhsa-RHEL8.xml"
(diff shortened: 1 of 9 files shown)


linux_os/guide/system/software/updating/security_patches_up_to_date/rule.yml
```diff
@@ -60,21 +60,19 @@ references:
...
-oval_external_content: "https://people.canonical.com/~ubuntu-security/oval/com.ubuntu.xenial.cve.oval.xml"
-{{% elif product in ["ol7", "ol8"] %}}
-oval_external_content: "https://linux.oracle.com/security/oval/com.oracle.elsa-all.xml.bz2"
-{{% elif product == "sle12" %}}
-oval_external_content: "https://support.novell.com/security/oval/suse.linux.enterprise.12.xml"
-{{% elif product == "sle15" %}}
-oval_external_content: "https://support.novell.com/security/oval/suse.linux.enterprise.15.xml"
+{{% if oval_feed_url %}}
+oval_external_content: {{{ oval_feed_url }}}
+  {{% if not oval_feed_url.startswith("https") %}}
```
redhatrises, line 67: This needs to fail in python code rather than through the jinja template.

linux_os/guide/system/software/updating/security_patches_up_to_date/rule.yml
```diff
@@ -60,21 +60,19 @@ references:
...
+  {{{ raise("OVAL feed of product '" + product + "' is not available through an encrypted channel: " + oval_feed_url) }}}
+  {{% endif %}}
+  {{% if not oval_feed_url.endswith(".xml") %}}
+warnings:
+    - general: "The OVAL feed of {{{ full_name }}} is not a XML file, which may not be understood by all scanners."
+  {{% endif %}}
 {{% else %}}
 {{# The rule will be "notchecked" #}}
+warnings:
+    - general: "{{{ full_name }}} doesn't have a OVAL CVE Feed associated, so this rule doesn't have a corresponding check."
```
redhatrises, line 77:
```suggestion
    - general: '{{{ full_name }}} doesn't have a corresponding OVAL CVE Feed. Therefore, this will result in a "not checked" result during a scan.'
```

linux_os/guide/system/software/updating/security_patches_up_to_date/rule.yml
matejak: Where specifically would you suggest to handle the issue? What's wrong with having it here, where other aspects of the feed URL are translated to the rule's XCCDF?

linux_os/guide/system/software/updating/security_patches_up_to_date/rule.yml
matejak: Thanks, I felt that something wasn't right with that sentence.

linux_os/guide/system/software/updating/security_patches_up_to_date/rule.yml
yuumasato: What if the `oval_feed_url` was process in [`open_environment`](https://github.com/ComplianceAsCode/content/blob/master/ssg/yaml.py#L141) ?

That is the function that is, in general, loading `product.yml`.

matejak: > What if the oval_feed_url was processed in open_environment ?

Technically, it could go there, but how would it be better than the current approach which works? Right now, we don't make any assumptions about the schema of `product.yml` - we just propagate its contents to the environment without asking questions. 

Acknowledging that the product can have e.g. the `oval_feed_url` property that, for some reason, shouldn't start with `http` means that the Python code becomes responsible for validation of `product.yml`. As we don't do any validation of products now, it is likely that starting to do so for the last property that caught attention would be inconsistent with how other properties are handled.

yuumasato: > Technically, it could go there, but how would it be better than the current approach which works?

The Rule becomes more readable, and the build would still fail if the feed url is not secure.

> Acknowledging that the product can have e.g. the `oval_feed_url` property that, for some reason, shouldn't start with `http` means that the Python code becomes responsible for validation of `product.yml`. As we don't do any validation of products now, it is likely that starting to do so for the last property that caught attention would be inconsistent with how other properties are handled.

Not validating every property may be inconsistent, but it should be ok to only validate the properties that need validation.

redhatrises: Yes. The readability of this rule is now quite atrocious due to these changes.

matejak: OK, thank you for the clarification.

matejak:
FYI, the validation has been moved to Python.

yuumasato (approved):

PR #6420, opened 2020-11-25
Add xwindows_runlevel_target to RHEL7 STIG profile.

ggbecker:
#### Description:

- Add xwindows_runlevel_target to RHEL7 STIG profile.

#### Rationale:

- This rule is now part of the check/fix text from RHEL7 STIG v3: https://www.stigviewer.com/stig/red_hat_enterprise_linux_7/2020-09-03/finding/V-204624

- It doesn't have a new STIG ID assigned though and it's treated in the same item configuration for removal of `xorg-x11-server-common` package


redhatrises:
Thanks... yeah. That needs to be fixed with DISA. Merging. OCP tests have no bearing on this PR.

PR #6424, opened 2020-11-25
Fix accounts tmout regex and remediation

freddieRv:
#### Description:

- Modified regex pattern on oval definition to capture only value of interest. Also to not match white-space arround "="
- Modified remediation script so it does not replace whole line with TMOUT variable assignment

#### Rationale:

- Regex pattern on oval definition caused oscap tool to fail when more statements followed the variable assignment (e.g. `TMOUT=600; readonly TMOUT; export TMOUT`)
- Replacing the whole line where the TMOUT var assignment is found could remove statements after the assignment

### DIFF
--- a/linux_os/guide/system/accounts/accounts-session/accounts_tmout/bash/shared.sh
+++ b/linux_os/guide/system/accounts/accounts-session/accounts_tmout/bash/shared.sh
@@ -2,8 +2,8 @@
 . /usr/share/scap-security-guide/remediation_functions
 {{{ bash_instantiate_variables("var_accounts_tmout") }}}
 
-if grep --silent ^TMOUT /etc/profile ; then
-        sed -i "s/^TMOUT.*/TMOUT=$var_accounts_tmout/g" /etc/profile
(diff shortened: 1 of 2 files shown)


redhatrises:
Ideally, `declare -xr TMOUT=600` would be used instead of the 3 statement handling which is a pita.

jan-cerny:
@openscap-ci ok to test

jan-cerny:
@openshift-ci-robot 
/ok-to-test

jan-cerny:
LGTM. The test scenarios pass and I have created 2 more to test your problem and it's fine. Thanks!

matejak:
Every now and then we get this problem with exporting and marking the variable read-only - I hope that we will be able to solve it.
Next, our check doesn't check for things like `export TMOUT=9999`, which would be a natural (i.e. incorrect, but not manufactured with an evil intent) configuration that could override the value.
This PR doesn't make things worse, so I am not against merging it.

@jan-cerny Just don't forget to push those scenarios.

PR #6425, opened 2020-11-26
Update Ubuntu support in kernel_module_disabled

Caligatio:
#### Description:

All the now supported versions of Ubuntu use systemd so the oval unnecessarily excluded kernel module loading config files and did not include Ubuntu families for the remediation templates.

#### Rationale:

Ubuntu 14.04 support was recently removed due to it being quasi-EOL (there is paid extended support available) which did not use systemd.  As such, all current and presumably future Ubuntu releases now use systemd. The `kernel_module_disabled` OVAL template unnecessarily excluded systemd-flavored files under 16.04 and will cause issues for my eventual Ubuntu 20.04 merge.


Caligatio:
Closing this in favor of #6427.

PR #6426, opened 2020-11-26
Fix NTP group.yml conditional statements

Caligatio:
#### Description:

The NTP group.yml has a condition block that checks the product type to then link off to documentation about NTP.  It currently checks against "debian" and "ubuntu" which are never valid products. By default, it links to Fedora documentation so there is no need to do a specific fedora check.

#### Rationale:

Built Ubuntu and Debian documentation is currently incorrect as it links to Fedora docs.


jan-cerny:
@openscap-ci OK to test

jan-cerny:
@openshift-ci-robot 
/ok-to-test

jan-cerny:
Thank you very much!

PR #6427, opened 2020-11-26
Add Ubuntu support to several remediations

Caligatio:
#### Description:

This is a superset of changes from #6425 that adds Ubuntu support to a number of remediations.  I did not touch the auditd-related rules as they need in-depth testing and the existing grub-related rules do not work on Ubuntu.

#### Rationale:

I'm not sure why Ubuntu support was not in several of these to begin with but these changes should work fine.


yuumasato:
@openscap-ci Ok to test

matejak:
Looks OK. We can't test that functionally, but the change doesn't introduce any risk of false negatives, so I merge it.
Thank you for the PR!

PR #6428, opened 2020-11-26
ocp4: Add check for audit log forwarding

JAORMX:
This checks for the ClusterLogForwardings resource in OCP. This resource
can forward audit logs to external stores.

This addresses CIS 1.2.23 and NIST SP 800-53 AU-9(2).

Depends on https://github.com/openshift/compliance-operator/pull/499


applications/openshift/api-server/api_server_audit_log_maxage/rule.yml
```diff
@@ -1,32 +0,0 @@
-documentation_complete: true
-
-prodtype: ocp4
-
-title: 'Configure the Audit Log Maximum Retention Days (maximumFileRetentionDays)'
```
redhatrises, line 5: @JAORMX you are still going to need this rule in addition to offloading audit logs. Security conscious environments like to have this configured in case offloading fails for some reason (which does happen). Loss of audit logs is a bad thing.

applications/openshift/api-server/audit_log_forwarding_is_enabled/rule.yml
```diff
@@ -0,0 +1,32 @@
+prodtype: ocp4
```
redhatrises, line 1: s/audit_log_forwarding_is_enabled/audit_log_forward_enabled/g 

Don't need verbs like `is, are, etc.` in rule names.

applications/openshift/api-server/audit_log_forwarding_is_enabled/rule.yml
```diff
@@ -0,0 +1,32 @@
...
+title: Ensure that the audit logs are forwarded off the cluster for retention
+
+description: |-
+  OpenShift audit works at the API server level, logging all requests coming to the server.
+  Audit is on by default and the best practice is to ship audit logs off the cluster for retention. 
+
+rationale: |-
+  Retaining logs ensures that you can go back in time and investigate or correlate any events.
+  Doing so off cluster ensures that an attacker that has access to the cluster won't be able to
+  tamper with the logs, as they'll be stored off-site.
```
redhatrises, line 12:
```suggestion
  Retaining logs ensures the ability to go back in time to investigate or correlate any events.
  Offloading audit logs from the cluster ensures that an attacker that has access to the cluster will not be able to
  tamper with the logs because of the logs being stored off-site.
```

It's best not to use contractions or "you" in these rationale as it makes them more official.

applications/openshift/api-server/audit_log_forwarding_is_enabled/rule.yml
```diff
@@ -0,0 +1,32 @@
+prodtype: ocp4
+
+title: Ensure that the audit logs are forwarded off the cluster for retention
```
redhatrises, line 3: Ensure that Audit Log Forwarding Is Enabled

applications/openshift/api-server/audit_log_forwarding_is_enabled/rule.yml
```diff
@@ -0,0 +1,32 @@
...
+
+description: |-
+  OpenShift audit works at the API server level, logging all requests coming to the server.
+  Audit is on by default and the best practice is to ship audit logs off the cluster for retention.
```
redhatrises, line 7: Remember: How does one does this? Can't tell by the existing rule description.

ocp4/profiles/moderate.profile
```diff
@@ -41,3 +41,5 @@ selections:
     - ocp_idp_no_htpasswd
     - ocp_allowed_registries_for_import
     - ocp_allowed_registries
+    # AU-9(2)
```
redhatrises, line 44: I would stop tracking NIST refs in the profiles. The reason is for this is that there are usually more than 1 NIST reference to a rule and then the question is... which one is in the profile?

JAORMX:
Pull-request updated, HEAD is now 7c4b566b4569d3d179398d975264e46f8d276956

applications/openshift/api-server/api_server_audit_log_maxage/rule.yml
JAORMX: there is a hardcode default for that already. Didn't think it was necessary given that.

applications/openshift/api-server/audit_log_forwarding_is_enabled/rule.yml
JAORMX: Let's write guidelines for writing content.

JAORMX:
Pull-request updated, HEAD is now 4426652f6a6ba647a2b42b3a3225e46a8c5f0b5c

JAORMX:
Pull-request updated, HEAD is now 3df2b6f6a6c1d097e12fc5f8b6446974d7171a26

JAORMX:
Pull-request updated, HEAD is now 14fac960dbb802c7b7ab33c4bd6575addbe5d4e7

JAORMX:
Pull-request updated, HEAD is now 9222bb3c405754a49a528080f3d2afdc4e4bfd5d

JAORMX:
Pull-request updated, HEAD is now 2d2bce269e86ea7e5813c4c5594a6d515b7ed5ac

applications/openshift/api-server/api_server_audit_log_maxage/rule.yml
redhatrises: ahh... got it. Thanks.

jhrozek (approved):
/lgtm
The PR looks good to me and it looks like all of Gabe's comments were addressed.

PR #6432, opened 2020-11-26
Skip irrelevant template directories

matejak:
The templating system threw an error whenever it came across a directory where a template could be, but no template was defined there.
This PR introduces skipping of directories that clearly aren't supposed to contain templates - they contain no `template.yml` nor any `*.template` files.
Any other directory is passed on, and if the template is incomplete, there will be a build-time error.

This PR extracted the second part of the `Template` constructor to a separate `load` method. The newly-introduced `looks_like_template` method reuses attributes set by the constructor.

### DIFF
--- a/ssg/templates.py
+++ b/ssg/templates.py
@@ -4,6 +4,7 @@ from __future__ import print_function
 import os
 import sys
 import imp
+import glob
 
 import ssg.build_yaml
 import ssg.utils
@@ -37,6 +38,8 @@ class Template():
         self.template_path = os.path.join(self.template_root_directory, self.name)
         self.template_yaml_path = os.path.join(self.template_path, "template.yml")
         self.preprocessing_file_path = os.path.join(self.template_path, preprocessing_file_name)
+
+    def load(self):
         if not os.path.exists(self.preprocessing_file_path):
             self.preprocessing_file_path = None
         self.langs = []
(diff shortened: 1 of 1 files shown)


ssg/templates.py
```diff
@@ -64,6 +67,16 @@ def preprocess(self, parameters, lang):
...
+    def looks_like_template(self):
+        if not os.path.isdir(self.template_root_directory):
+            return False
+        if os.path.islink(self.template_root_directory):
+            return False
+        template_sources = glob.glob(os.path.join(self.template_path, "*.template"))
+        if not os.path.isfile(self.template_yaml_path) and not template_sources:
+            return False
+        return True
+
```
yuumasato, line 79: I think this check could be done during `Template` instantiation, sot that it every instance is checked.

yuumasato: This is not a big deal, merging to unblock CI.

matejak: I was thinking of it, but the constructor can't return any value - it just initializes the object, and we want to prevent its addition to the templates dictionary.
Then, I was thinking of a static method that one would use before instantiating `Template`, but that one would have to calculate all paths again, and that would be either duplication, or it would require even more code extractions.

yuumasato: Yeap, I came to similar conclusion.
Ideally all instances would come valid and wouldn't need one to remember to call a function to check it.
This may warrant a comment or documentation note...

PR #6434, opened 2020-11-27
Revert hardcoding of ClientAliveCountMax to 0

### LINKED ISSUE #6408: How to parameterize a rule/OVAL without breaking existing use?
I'm working on implementing the CIS Level 1 profile for Ubuntu 20.04 and its guidance has changed on the SSH `ClientAliveCountMax` value. It appears in all the existing rules that it is recommended to be set to `0` whereas in this guidance it is set to `3`.  I had previously implemented the Ubuntu 20.04 "Standard" profile which used this set to `0` recommendation.

How to I change the rule to allow passing in a `xccdf_value` but also having default value so I do not need to change all the pre-existing rules?  Since I already used an instance of it being `0`, I rather not set this at the `prodtype` level.

Caligatio:
## Description:

This PR reverts the previous change that hardcoded `ClientAliveCountMax` to `0`.  It also removes several tests from `sshd_set_idle_timeout` as they appear to assume that `ClientAliveCountMax` will be `0`.  The previous OVAL also allowed for a config value less than or equal to the specified value which I changed to a direct equals comparison.

EDIT: I would really appreciate a second set of eyes on the tests.  I can read the bash but I have no idea how they're used and whether the values in there need to match the rule.

#### Rationale:

Per #6408, there was a discussion at https://bugzilla.redhat.com/show_bug.cgi?id=1873547 how `ClientAliveCountMax` should always be `0` to make the desired timeout work as anticipated.  Starting with OpenSSH v8.2 (which is used in Ubuntu 20.04), the meaning of a `0` value changed to mean that timeouts are disabled.  This PR reverts the previous changes so that the desired value can be specified in the rule.

Fixes #6408


matejak:
First of all, let me thank you for bringing this issue up!

Next, please, not so fast - the new behavior doesn't apply to products that use OpenSSH<8.2, which is, among others, RHEL<=8. For those products, we have to consider dependency of this rule and `sshd_set_idle_timeout`.

The new behavior of SSH is, however, fundamentally different. Would you agree that

- no configuration will make SSH to terminate idle sessions, so it is not possible to implement the `sshd_set_idle_timeout` rule on newer systems any more, and that
- SSH will continue to terminate unresponsive sessions, but only if `ClientAliveCountMax` > 0?

If that's the case, I think that the following needs to be done:

1. The current `sshd_set_idle_timeout` rule needs to have its applicability revised - only products that ship the old SSH are applicable. Tests are correct, I think that there is no reason to delete them.
2. The `sshd_set_keepalive` can be parametrized again, but the `sshd_set_idle_timeout` needs to continue to work the same way as it works now. What about creating a copy of the rule that you are reverting, call it `sshd_set_keepalive_0`?
3. (optional) A `sshd_set_unresponsive_timeout` should be created, as a new complement of `sshd_set_keepalive`.

Caligatio:
I'm honestly not familiar with the innards of OpenSSH to tell you how the new behavior works.  However, I can affirmatively say that the CIS Ubuntu 20.04 recommended settings are `ClientAliveInterval` = `300` and `ClientAliveCountMax` = `3`.  The CIS Ubuntu 18.04 recommended settings are `ClientAliveInterval` = `300` and `ClientAliveCountMax` = `0` so I presume they did their research?  I'm thinking the setting is still viable?

I feel like it's very duplicative to have essentially the same "function" duplicated twice but having one where the input hardcoded to 0.  If you abstract this a bit, the guidance specifies a pair of values that effects the hardening you want.  For some distros/versions, that means setting `ClientAliveCountMax` = `0` and others have it set to `ClientAliveCountMax` = `3`.  I see these as being the same rule with just different parameterized input?

konstruktoid:
I created an issue and a proposal of change to the CIS benchmarks (OpenSSH, and the distributions) a few months ago.
It basically boils down to that the ClientAliveCountMax should be set between 1 and 3 (or according to your policy of course).

And ClientAliveCountMax set to 0 "works" for OpenSSH 7.9 but does not work anymore with OpenSSH 8.x.

https://bugzilla.mindrot.org/show_bug.cgi?id=3172#c6:
"ClientAliveCountMax=0 has never been specified to work as an idle timeout. If it did that then it was by accident and would be unreliable. E.g. if the client specified it's own ServerAliveTimeout or kept a forwarded TCP connection open then it would never fire."

matejak:
> I feel like it's very duplicative to have essentially the same "function" duplicated twice but having one where the input hardcoded to 0. If you abstract this a bit, the guidance specifies a pair of values that effects the hardening you want. For some distros/versions, that means setting `ClientAliveCountMax` = `0` and others have it set to `ClientAliveCountMax` = `3`. I see these as being the same rule with just different parameterized input?

I think that I can understand your concern - why to have two rules, when one parametrized rule can behave both ways? And I also agree with @konstruktoid that the `sshd_set_idle_timeout` rule was backed by undocumented behavior.
My only concern is that the `sshd_set_idle_timeout` still works as expected on a huge range of supported systems, but it works only if the `ClientAliveCountMax` is set to 0. With regard to those cases, it shouldn't be possible to parametrize that check, because it could lead to false positives if the XCCDF value is altered e.g. by the tailoring to have positive values.
Having a separate rule would make sure that a profile that contains `sshd_set_idle_timeout` and that hypothetical `sshd_set_keepalive_0` would remediate the system into compliance. Having just a separate check would mean that the remediation will be missing.

Gradually, products that ship the old SSH server will reach their end of life, which means that the respective rules will get dropped eventually. It just requires patience.

Caligatio:
@matejak I'm relatively new to this whole area and neglected to consider modifying the XCCDF value at runtime.

I'm trying to think how to clean up the git history here a bit.  Should I just close out this PR/branch and make a new one?  That way I can just do a git move from the "old" `sshd_set_keepalive` to your recommended name of `sshd_set_keepalive_0` and then reintroduce my changes as the "new" `sshd_set_keepalive`.  I'm not super excited about needing to tweak all the existing profiles to use the new `sshd_set_keepalive_0` rule but I think the `sshd_set_keepalive_0` name is usefully descriptive.

matejak:
It's great the we understand each other. I suggest to leave this PR open for a couple of days, so e.g. @redhatrises can also express his opinion. You have uncovered another problem, which is the impact of the proposed change on existing profiles, specifically on existing tailoring.

This is a really weird situation, because:

1. There was a parametrized rule `sshd_set_keepalive`.
2. We have discovered that in the given context, parametrization of the rule is a bug, the parameter should be frozen to 0, which we did. The rule ID stayed.
3. You showed up, pointed out that the context is invalid, and some kind of parametrization makes sense in the new context.

The parametrized rule from point 3 will be identical to the parametrized rule from point 1, so it makes sense that they should have the same IDs, while the "hardcoded" rule should be introduced with a different ID.
As a counter argument, the intention of rule from point 1 and of the "hardcoded" rule is the same, so it is counter-intuitive for them to have different IDs, as this difference would imply that existing profiles and tailoring files need to be updated for no reason.

Having more opinions helps when there are strong arguments both ways.

Caligatio:
As an alternative, we could make the rule be SSH version-aware and ignore the XCCDF value if SSH is older than v8.2.  This isn't without i's own drawbacks as it would look like it's misbehaving should people actually want `ClientAliveCountMax` to be something other than `0` on older versions.

matejak:
> As an alternative, we could make the rule be SSH version-aware and ignore the XCCDF value if SSH is older than v8.2. This isn't without its own drawbacks as it would look like it's misbehaving should people actually want `ClientAliveCountMax` to be something other than `0` on older versions.

The problem with this is that the configuration option changed its meaning, so rules that apply to SSH before and after 8.2 would have different description and rationale,. Checks and remediations would be much more complex, too.
I have encouraged more people to post their feedback here to help us to see consequences of rule IDs pick.

matejak:
So, as a nobody seems to have strong opinions, I propose to go through the path of parametrizing the current rule and keeping it applicable for older products, while conserving the current hardcoded rule under that different ID, and making that one applicable only for products that ship OpenSSH<8.2.

So what about the tailoring and ID continuity? We can reintroduce the `var_sshd_set_keepalive=0` to respective profiles, so even if the parametrized rule ends up in a profile because of tailoring, it will do the right thing by default. If both rules end up in a profile because of tailoring, it shouldn't cause a problem, as they should aim to configure the same thing, only the reasoning is different, which may confuse readers of scan reports.

What do you think of this?

matejak:
Hello @Caligatio, are you still interested in the PR?

It needs these adjustments:

- Don't delete tests - the idle timeout rule should stay as it is, but only for products that ship OpenSSH<8.2
- Please create `sshd_set_keepalive_to_0` as a copy of the (current upstream master) hardcoded keepalive rule. We will have to pay some attention to IDs/references of those rule, and we will have to make sure that they are selected correctly in profiles, but that's just a couple of lines to pay attention to.
- Parametrizing  `sshd_set_keepalive` is a good move - thanks!

It is obvious that there is not much to be done on this PR, so please stick with us a little longer.

carlosmmatos:
@matejak Is the CI stuck? What's the status of this PR or are we looking to close it for now unless the OP want's to keep it alive?

Caligatio:
Sorry, I've undergone some major computer changes and lost my setup to easily work on this project.  I think the above two commits implement what @matejak asked?

matejak:
test this please

linux_os/guide/services/ssh/ssh_server/sshd_set_keepalive_0/rule.yml
```diff
@@ -0,0 +1,71 @@
...
+    This ensures a user login will be terminated as soon as the <tt>ClientAliveInterval</tt>
+    is reached.
+
+severity: medium
+
+identifiers:
+    cce@rhel7: CCE-27082-7
+    cce@rhel8: CCE-80907-9
+    cce@rhcos4: CCE-82464-9
+    cce@sle12: CCE-83034-9
```
matejak, line 27: Those CCEs are duplicated in the parametrized rule. Please delete four random CCEs from https://github.com/ComplianceAsCode/content/blob/master/shared/references/cce-redhat-avail.txt and use them as replacement for these ones.
This will make the CI to pass.

matejak:
ok to test

Caligatio:
I'm going to have to claim complete ignorance on CCE numbers but the last commit did what you asked!

linux_os/guide/services/ssh/ssh_server/sshd_set_keepalive/oval/shared.xml
```diff
@@ -0,0 +1,52 @@
...
+    </criteria>
+  </definition>
+  <ind:textfilecontent54_test check="all" check_existence="all_exist"
+  comment="Tests the value of the ClientAliveCountMax setting in the /etc/ssh/sshd_config file"
+  id="test_sshd_clientalivecountmax" version="1">
+    <ind:object object_ref="obj_sshd_clientalivecountmax" />
+    <ind:state state_ref="state_sshd_clientalivecountmax" />
+  </ind:textfilecontent54_test>
+  <ind:textfilecontent54_state id="state_sshd_clientalivecountmax" version="1">
+    <ind:subexpression datatype="int" operation="equals" var_check="all"
```
matejak, line 40: The original check could handle supercompliance, so I propose to follow that.

```suggestion
    <ind:subexpression datatype="int" operation="less than or equal" var_check="all"
```

linux_os/guide/services/ssh/ssh_server/sshd_set_keepalive/rule.yml
```diff
@@ -9,10 +9,13 @@ description: |-
     each <tt>ClientAliveCountMax</tt> message. If the SSH server does not
     receive a response from the client, then the connection is considered idle
     and terminated.
-
     To ensure the SSH idle timeout occurs precisely when the
     <tt>ClientAliveInterval</tt> is set, set the <tt>ClientAliveCountMax</tt> to
-    value of <tt>0</tt>.
+    value of <tt>0</tt>. This profile sets <tt>ClientAliveCountMax</tt> to
```
matejak, line 14: This and the previous lines are incorrect, but the OCIL text is correct, so it just needs to be copy-pasted here.

matejak (changes requested):
Almost there sorry for the delay!
I have found one minor issue in OVAL that I could propose a fix for, and one issue in the rule description that can be fixed easily, but the fix can't be suggested using the GH interface.

I have also updated selections in profiles, so the entire rule switch is contained in this PR.

linux_os/guide/services/ssh/ssh_server/sshd_set_keepalive/rule.yml
Caligatio: Changes made!

Caligatio:
@matejak: your Github wizardry puts me to shame :)

I think I accepted your change on the OVAL and updated the rule description with the new behavior.

matejak (approved):

matejak:
Thanks @Caligatio for staying tuned!

I have checked the content, and that's fine by me. I just need someone to check the profile change that I have proposed.

ggbecker (approved):
Changes from @matejak look good.

I'll just leave this comment here for posterity: OpenSSH >= 8.2 does not terminate idle sessions anymore. The previous behavior was not intentional. Fundamentally the security requirement of idle sessions in OpenSSH cannot be achieved at the moment. Any combination of ClientAliveMaxCount and ClientAliveInterval will **not** work.

PR #6435, opened 2020-11-27
CIS Benchmark 4.2.13 (kubelet_configure_tls_cipher_suites)

tmishina:
#### Description:

- implement a rule for [CIS Benchmark for OpenShift 4.2.13](https://workbench.cisecurity.org/sections/514741/recommendations/839648).

#### Rationale:

- (TBD)

#### discussion topic: check values in a list

Currently, this code fails with following error.  [yamlfile_value template](https://github.com/ComplianceAsCode/content/blob/master/shared/templates/yamlfile_value/oval.template) now supports checking value under a map, but (as far as I understand) it does not support checking value in a list.

```
$ oc logs -f platform-scan-api-checks-pod -c scanner
HOSTROOT not set, using normal oscap
Running oscap-chroot as oscap xccdf eval --fetch-remote-resources --profile xccdf_org.ssgproject.content_profile_cis --results-arf /tmp/report-arf.xml /content/ssg-ocp4-ds.xml
The scanner returned 1
OpenSCAP Error: File '/content/ssg-ocp4-ds.xml' line 8128: Element '{http://oval.mitre.org/XMLSchema/oval-definitions-5}field': Duplicate key-sequence ['#'] in unique identity-constraint '{http://oval.mitre.org/XMLSchema/oval-definitions-5#independent}UniqueYamlFileValueFieldName'.
 [/builddir/build/BUILD/openscap-1.3.4/src/XCCDF/xccdf_session.c:681]
File '/content/ssg-ocp4-ds.xml' line 8129: Element '{http://oval.mitre.org/XMLSchema/oval-definitions-5}field': Duplicate key-sequence ['#'] in unique identity-constraint '{http://oval.mitre.org/XMLSchema/oval-definitions-5#independent}UniqueYamlFileValueFieldName'.
 [/builddir/build/BUILD/openscap-1.3.4/src/XCCDF/xccdf_session.c:681]
...
Invalid SCAP Source Datastream (1.3) content in /content/ssg-ocp4-ds.xml. [/builddir/build/BUILD/openscap-1.3.4/src/source/oscap_source.c:353]
Invalid SCAP Source Datastream (1.3) content in /content/ssg-ocp4-ds.xml [/builddir/build/BUILD/openscap-1.3.4/src/XCCDF/xccdf_session.c:797]
OpenSCAP Error: Unable to open file: '/tmp/report-arf.xml' [/builddir/build/BUILD/openscap-1.3.4/src/source/oscap_source.c:288]
Could not create Result DataStream session: File is not Result DataStream. [/builddir/build/BUILD/openscap-1.3.4/src/DS/ds_rds_session.c:53]
Failed to split given result datastream '/tmp/report-arf.xml'.
The rds-split operation returned 1
```

### DIFF
--- a/applications/openshift/kubelet/kubelet_configure_tls_cipher_suites/ocp4/e2e-remediation.sh
+++ b/applications/openshift/kubelet/kubelet_configure_tls_cipher_suites/ocp4/e2e-remediation.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+cat << EOS | oc apply -f -
+apiVersion: machineconfiguration.openshift.io/v1
+kind: KubeletConfig
+metadata:
+  name: tls-ciphers
+spec:
+  machineConfigPoolSelector:
+    matchLabels:
+      custom-kubelet: tls-ciphers
+  kubeletConfig:
+    tlsCipherSuites:
+    - TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
+    - TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
+    - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
+    - TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
+EOS
+
+overwrite_flag=''
+if [ "$(oc get machineconfigpool worker -o=jsonpath='{.metadata.labels.custom-kubelet}')" != "" ] ; then
+  overwrite_flag='--overwrite'
+fi
+
+oc label ${overwrite_flag} machineconfigpool worker custom-kubelet=tls-ciphers
--- a/applications/openshift/kubelet/kubelet_configure_tls_cipher_suites_ingresscontroller/rule.yml
+++ b/applications/openshift/kubelet/kubelet_configure_tls_cipher_suites_ingresscontroller/rule.yml
@@ -1,0 +1,42 @@
+documentation_complete: true
+
+prodtype: ocp4
+
+title: "Ensure that the Ingress Controller only makes use of Strong Cryptographic Ciphers"
+
+description: |-
+  Ensure that the Ingress Controller is configured to only use strong cryptographic ciphers.
+
+rationale: |-
+  TLS ciphers have had a number of known vulnerabilities and weaknesses,
+  which can reduce the protection provided by them. By default Kubernetes
+  supports a number of TLS ciphersuites including some that have security
+  concerns, weakening the protection provided.
+
+severity: medium
+
+#identifiers:
+#    cce@ocp4:
+
+references:
+  cis@ocp4: 4.2.13
+
+ocil_clause: "TLS cipher suite configuration is not configured"
+
+ocil: |-
+  Run the following comman on the kubelete nodes(s):
+  {{% raw %}}<pre>oc -n openshift-ingress-operator patch ingresscontroller/default --type merge -p '{"spec":{"tlsSecurityProfile":{"type":"Custom","custom":{"ciphers":["ECDHE-ECDSA-AES128-GCM-SHA256","ECDHE-RSA-AES128-GCM-SHA256","ECDHE-RSA-AES256-GCM-SHA384"],"minTLSVersion":"VersionTLS12"} } } }'</pre>{{% endraw %}}
+
+warnings:
+- general: |-
+    {{{ openshift_cluster_setting("/apis/operator.openshift.io/v1/namespaces/openshift-ingress-operator/ingresscontrollers/default") | indent(4) }}}
+
+template:
+  name: yamlfile_value
+  vars:
+    ocp_data: "true"
+    filepath: '/apis/operator.openshift.io/v1/namespaces/openshift-ingress-operator/ingresscontrollers/default'
+    yamlpath: ".status.tlsProfile.ciphers[:]"
+    values:
+    - value: '^(ECDHE-ECDSA-AES128-GCM-SHA256|ECDHE-RSA-AES128-GCM-SHA256|ECDHE-ECDSA-CHACHA20-POLY1305|ECDHE-RSA-AES256-GCM-SHA384|ECDHE-RSA-CHACHA20-POLY1305|ECDHE-ECDSA-AES256-GCM-SHA384|AES256-GCM-SHA384|AES128-GCM-SHA256)$'
+      operation: 'pattern match'
--- a/applications/openshift/kubelet/kubelet_configure_tls_cipher_suites_kubeapiserver_operator/rule.yml
+++ b/applications/openshift/kubelet/kubelet_configure_tls_cipher_suites_kubeapiserver_operator/rule.yml
@@ -2,10 +2,10 @@ documentation_complete: true
 
 prodtype: ocp4
 
-title: "Ensure that the Kubernetes API Server only makes use of Strong Cryptographic Ciphers"
+title: "Ensure that the Kubernetes API Server Operator only makes use of Strong Cryptographic Ciphers"
 
 description: |-
-  Ensure that the Kubernetes API Server is configured to only use strong cryptographic ciphers.
+  Ensure that the Kubernetes API Server Operator is configured to only use strong cryptographic ciphers.
 
 rationale: |-
   TLS ciphers have had a number of known vulnerabilities and weaknesses,
--- a/applications/openshift/kubelet/kubelet_configure_tls_cipher_suites_kubeapiserver_operator/rule.yml
+++ b/applications/openshift/kubelet/kubelet_configure_tls_cipher_suites_kubeapiserver_operator/rule.yml
@@ -1,0 +1,42 @@
+documentation_complete: true
+
+prodtype: ocp4
+
+title: "Ensure that the Kubernetes API Server only makes use of Strong Cryptographic Ciphers"
+
+description: |-
+  Ensure that the Kubernetes API Server is configured to only use strong cryptographic ciphers.
+
+rationale: |-
+  TLS ciphers have had a number of known vulnerabilities and weaknesses,
+  which can reduce the protection provided by them. By default Kubernetes
+  supports a number of TLS ciphersuites including some that have security
+  concerns, weakening the protection provided.
+
+severity: medium
+
+#identifiers:
+#    cce@ocp4:
+
+references:
+  cis@ocp4: 4.2.13
+
+ocil_clause: "TLS cipher suite configuration is not configured"
+
+ocil: |-
+  Run the following comman on the kubelete nodes(s):
+  {{% raw %}}<pre>oc patch kubeapiservers.operator.openshift.io cluster --type merge -p '{"spec":{"unsupportedConfigOverrides":{"servingInfo":{"cipherSuites":["TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256","TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256","TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305","TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384","TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305","TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384","TLS_RSA_WITH_AES_256_GCM_SHA384","TLS_RSA_WITH_AES_128_GCM_SHA256"]} } } }'</pre>{{% endraw %}}
+
+warnings:
+- general: |-
+    {{{ openshift_cluster_setting("/apis/operator.openshift.io/v1/kubeapiservers/cluster") | indent(4) }}}
+
+template:
+  name: yamlfile_value
+  vars:
+    ocp_data: "true"
+    filepath: '/apis/operator.openshift.io/v1/kubeapiservers/cluster'
+    yamlpath: ".spec.unsupportedConfigOverrides.servingInfo.cipherSuites[:]"
+    values:
+    - value: '^(TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256|TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256|TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305|TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384|TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305|TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384|TLS_RSA_WITH_AES_256_GCM_SHA384|TLS_RSA_WITH_AES_128_GCM_SHA256)$'
+      operation: 'pattern match'
--- a/applications/openshift/kubelet/kubelet_configure_tls_cipher_suites_openshiftapiserver_operator/rule.yml
+++ b/applications/openshift/kubelet/kubelet_configure_tls_cipher_suites_openshiftapiserver_operator/rule.yml
@@ -1,0 +1,42 @@
+documentation_complete: true
+
+prodtype: ocp4
+
+title: "Ensure that the OpenShift API Server only makes use of Strong Cryptographic Ciphers"
+
+description: |-
+  Ensure that the OpenShift API Server is configured to only use strong cryptographic ciphers.
+
+rationale: |-
+  TLS ciphers have had a number of known vulnerabilities and weaknesses,
+  which can reduce the protection provided by them. By default Kubernetes
+  supports a number of TLS ciphersuites including some that have security
+  concerns, weakening the protection provided.
+
+severity: medium
+
+#identifiers:
+#    cce@ocp4:
+
+references:
+  cis@ocp4: 4.2.13
+
+ocil_clause: "TLS cipher suite configuration is not configured"
+
+ocil: |-
+  Run the following comman on the kubelete node(s):
+  {{% raw %}}<pre>oc patch openshiftapiservers.operator.openshift.io cluster --type merge -p '{"spec":{"unsupportedConfigOverrides":{"servingInfo":{"cipherSuites":["TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256","TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256","TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305","TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384","TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305","TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384","TLS_RSA_WITH_AES_256_GCM_SHA384","TLS_RSA_WITH_AES_128_GCM_SHA256"] } } } }'</pre>{{% endraw %}}
+
+warnings:
+- general: |-
+    {{{ openshift_cluster_setting("/apis/operator.openshift.io/v1/openshiftapiservers/cluster") | indent(4) }}}
+
+template:
+  name: yamlfile_value
+  vars:
+    ocp_data: "true"
+    filepath: '/apis/operator.openshift.io/v1/openshiftapiservers/cluster'
+    yamlpath: ".spec.unsupportedConfigOverrides.servingInfo.cipherSuites[:]"
+    values:
+    - value: '^(TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256|TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256|TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305|TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384|TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305|TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384|TLS_RSA_WITH_AES_256_GCM_SHA384|TLS_RSA_WITH_AES_128_GCM_SHA256)$'
+      operation: 'pattern match'
--- a/applications/openshift/kubelet/kubelet_configure_tls_cipher_suites_openshiftapiserver_operator/rule.yml
+++ b/applications/openshift/kubelet/kubelet_configure_tls_cipher_suites_openshiftapiserver_operator/rule.yml
@@ -2,10 +2,10 @@ documentation_complete: true
 
 prodtype: ocp4
 
-title: "Ensure that the OpenShift API Server only makes use of Strong Cryptographic Ciphers"
+title: "Ensure that the OpenShift API Server Operator only makes use of Strong Cryptographic Ciphers"
 
 description: |-
-  Ensure that the OpenShift API Server is configured to only use strong cryptographic ciphers.
+  Ensure that the OpenShift API Server Operator is configured to only use strong cryptographic ciphers.
 
 rationale: |-
   TLS ciphers have had a number of known vulnerabilities and weaknesses,
--- a/ocp4/profiles/cis.profile
+++ b/ocp4/profiles/cis.profile
@@ -174,7 +174,7 @@ selections:
     # Like kubelet_disable_readonly_port but check for .apiServerArguments["kubelet-client-certificate"]
     - kubelet_configure_tls_key
     # Like kubelet_disable_readonly_port but check for .apiServerArguments["kubelet-client-key"]
-  # 4.2.13 Ensure that the Kubelet only makes use of Strong Cryptographic Ciphers
+
 
   ### 5 Policies
   ###
--- a/applications/openshift/kubelet/kubelet_configure_tls_cipher_suites/ocp4/e2e.yml
+++ b/applications/openshift/kubelet/kubelet_configure_tls_cipher_suites/ocp4/e2e.yml
@@ -1,0 +1,4 @@
+---
+default_result: FAIL
+result_after_remediation: PASS
+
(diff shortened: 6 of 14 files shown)


evgenyz:
You are trying to make sure that a sequence in `.status.tlsProfile.ciphers[:]` only contain one or more of the 8 listed strings (`TLS_AES...`), but not any other. Am I right?

evgenyz:
This could be a workaround:

```yaml
template:
  name: yamlfile_value
  vars:
    ocp_data: "true"
    filepath: /apis/operator.openshift.io/v1/namespaces/openshift-ingress-operator/ingresscontrollers/default
    yamlpath: ".status.tlsProfile.ciphers[:]"
    values:
    - value: '^TLS_AES_128_GCM_SHA256|TLS_AES_256_GCM_SHA384|TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305|TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384|TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384|TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305|TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384|TLS_RSA_WITH_AES_256_GCM_SHA38|TLS_RSA_WITH_AES_128_GCM_SHA256$'
      operation: 'pattern match'
```

evgenyz:
Meanwhile I'll think about how we could improve the template to get better support for this case, because there is a nice OVAL feature for this type of checks: https://github.com/OpenSCAP/openscap/blob/5de37831defc4f7615f45f4184bf2cb5427e9c82/tests/probes/yamlfilecontent/test_probes_yamlfilecontent_array.xml#L136

evgenyz:
But then it has a bit different semantics. Hmm...

tmishina:
@evgenyz 

> You are trying to make sure that a sequence in `.status.tlsProfile.ciphers[:]` only contain one or more of the 8 listed strings (`TLS_AES...`), but not any other. Am I right?
 
Yes, that's what I'd like to check.
I will try your suggested workaround, thanks.

Before the trial, let me confirm whether my understanding about the yamlpath is correct.

> `".status.tlsProfile.ciphers[:]"`

In my understanding, this yamlpath means `["a", "b", "c"]`, not `"a b c"` for a yaml data below. Am I correct?

```yaml
status:
  tlsProfile:
    ciphers:
    - a
    - b
    - c
```

evgenyz:
Yes, that is correct. The OVAL from my workaround would basically go through all elements of the sequence (list) and make sure that each value match the regexp (^TLS..|TLS..|...$).

evgenyz:
@tmishina Actually, the regexp should be `^(TLS_AES_128_GCM_SHA256|TLS_AES_256_GCM_SHA384|TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305|TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384|TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384|TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305|TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384|TLS_RSA_WITH_AES_256_GCM_SHA38|TLS_RSA_WITH_AES_128_GCM_SHA256)$`. 

And, by the way, are these strings case-sensitive?

tmishina:
@evgenyz Thank you, your suggestion works file with several test patterns!
(and, according to the standard, they look case-sensitive)
I pushed [a new commit](https://github.com/tmishina/content/commit/2e00408c5d9d11f85fb0d0365d61944b8f3d0811).

Now I'm trying to remaining 4 rules in section 4.3.12.  Two of the three rules can be implemented as similar to the first one, but other two are challenging.

1. how to parse json in a string
We need to describe a pattern in a text in JSON (yaml).  In other words, two-level JSON parsing is required.
(details can be found in [my wiki page](https://github.com/tmishina/content/wiki/cis-4.2.13#openshift-authentication))
Is it possible for the current `yamlfile_value` template to do so?

2. how to implement "OR" condition
In the final rule we need to say that the entire rule will be PASS if
```
.tlsProfile is null OR each of `.tlsProfile.cipherSuites[:]` is in a permitted cipher suite list
```
The question here is how to express this kind of `OR` rule in `yamlfile_value` template; is it possible for the current `yamlfile_value` template?

tmishina:
@JAORMX CI returns error not only for the target rule (`kubelet_configure_tls_cipher_suites_ingresscontroller`) but also other rules.
1. can I ignore the errors on non-target rules?
2. is it possible to change the configuration of CI container to make it be compliant?

evgenyz:
> 1. how to parse json in a string
> We need to describe a pattern in a text in JSON (yaml).  In other words, two-level JSON parsing is required.
> (details can be found in [my wiki page](https://github.com/tmishina/content/wiki/cis-4.2.13#openshift-authentication))
> Is it possible for the current `yamlfile_value` template to do so?

This is currently impossible, sorry. We are trying draft a way to deal with this kind of Russian Dolls, but it won't be implemented any time soon.
 
> 2. how to implement "OR" condition
> In the final rule we need to say that the entire rule will be PASS if
> ```
> .tlsProfile is null OR each of `.tlsProfile.cipherSuites[:]` is in a permitted cipher suite list
> ```

This is rather simple, `check_existence: 'any_exist'` (instead of the default `all_exists`) will evaluate the test as successful if there are no objects in the system.
```yaml
template:
  name: yamlfile_value
  vars:
    ocp_data: "true"
    filepath: /apis/operator.openshift.io/v1/namespaces/openshift-ingress-operator/ingresscontrollers/default
    yamlpath: ".status.tlsProfile.ciphers[:]"
    check_existence: 'any_exist'
    values:
    - value: '^(TLS_AES_128_GCM_SHA256|TLS_AES_256_GCM_SHA384|TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305|TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384|TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384|TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305|TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384|TLS_RSA_WITH_AES_256_GCM_SHA38|TLS_RSA_WITH_AES_128_GCM_SHA256)$'
      operation: 'pattern match'
```

tmishina:
@evgenyz thank you, then for the first rule, is this (below) kind of terrible regexp pattern the only workaround if cipher suite `A`, `B`, and `C` is approved cipher?

```yaml
template:
  name: yamlfile_value
  vars:
    ocp_data: "true"
    filepath: /apis/operator.openshift.io/v1/namespaces/openshift-authentication/configmaps/v4-0-config-system-cliconfig
    yamlpath: '.data['"v4-0-config-system-cliconfig"]'
    values:
    - value: '\"cipherSuite\": \[(A|A,B|A,C|A,B,C|A,C,B|B|B,A|B,C|B,A,C|B,C,A|C|C,A|C,B|C,A,B|C,B,A)\]'
      operation: 'pattern match'
```

evgenyz:
I'd use something like `\"cipherSuite\": \[(A|B|C|[,\s])+\]`

tmishina:
@evgenyz thank you, yes, I can't believe my previous pattern...yours is much more better than mine.
Finally I found this pattern: `"\"cipherSuites\":\[(\"TLS_AES_128_GCM_SHA256\"|\"TLS_AES_256_GCM_SHA384\"|\"TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305\"|\"TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384\"|\"TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384\"|\"TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305\"|\"TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384\"|\"TLS_RSA_WITH_AES_256_GCM_SHA38\"|\"TLS_RSA_WITH_AES_128_GCM_SHA256\"|,| )*\]"`

tmishina:
Now I've tested the rules and verified that they works fine. Please review this up-to-date rules, thanks.
And I have two remaining questions for this PR...

1. Current OCP on CodeReady Container does not comply with the CIS Benchmark for OpenShift, and therefore the results for my rules are "FAIL".  I added "PASS" patterns to the rule (e.g., [for kubelet_configure_tls_cipher_suites_ingresscontroller](https://github.com/tmishina/content/blob/e8a65341eb173ee51bf76fc1b3cf21343a7b3179/applications/openshift/kubelet/kubelet_configure_tls_cipher_suites_ingresscontroller/rule.yml#L42)).  Are there better ways to test this rule?
2. This rule also need [changes on Compliance Operator](https://github.com/tmishina/compliance-operator/commit/5005c7cf8d270632da00f09af487e73f2e9a4850).  Can I create a Pull Request for compliance-operator?  Is there a process to request changes from viewpoint of ComplianceAsCode?

JAORMX:
> Now I've tested the rules and verified that they works fine. Please review this up-to-date rules, thanks.
> And I have two remaining questions for this PR...
> 
>     1. Current OCP on CodeReady Container does not comply with the CIS Benchmark for OpenShift, and therefore the results for my rules are "FAIL".  I added "PASS" patterns to the rule (e.g., [for kubelet_configure_tls_cipher_suites_ingresscontroller](https://github.com/tmishina/content/blob/e8a65341eb173ee51bf76fc1b3cf21343a7b3179/applications/openshift/kubelet/kubelet_configure_tls_cipher_suites_ingresscontroller/rule.yml#L42)).  Are there better ways to test this rule?


> 
>     2. This rule also need [changes on Compliance Operator](https://github.com/tmishina/compliance-operator/commit/5005c7cf8d270632da00f09af487e73f2e9a4850).  Can I create a Pull Request for compliance-operator?  Is there a process to request changes from viewpoint of ComplianceAsCode?

Yes! please do

tmishina:
@JAORMX I found another issue when remediation has been performed.  When a remediation script changes the cipher suites of the ingress controller to the ones permitted in [the benchmark for OpenShift](https://workbench.cisecurity.org/sections/514741/recommendations/839648), services through the ingress controller (such as web console, internal registry) becomes unreachable.

Is there workaround (e.g., client configuration change) for this? or should we modify the set of permitted cipher suites of the benchmark to make the services be available?
(The set of permitted cipher suites is same as [the benchmark for kubernetes](https://www.cisecurity.org/benchmark/kubernetes/), and therefore we need to adjust the difference when we modify the set for OCP...)

tmishina:
OK, now test is passed at my local (CodeReady Containers) cluster.  Something goes wrong on CI/CD test environment...
Let me summarize the rules.


### Goal
Goal of section 4.2.13 is configuring following TLS Cipher Suites to permitted ones in the CIS Benchmark specification.

- Ingress Operator
- kube API Operator
- OpenShift API Operator

These configurations are inherited from the configuration openshift authentication configuration which is registered as a configmap `v4-0-config-system-cliconfig` in namespace `openshift-authentication`.

To remediate the configurations of the operators, we just need to change `spec` of each operator, and we don't need to change `v4-0-config-system-cliconfig`.

### Details of check and remediation
#### Ingress Operator
- check: all of `.status.tlsProfile.ciphers[:]` are permitted in CIS Benchmark
    - **Note**: cipher suite names for ingress operator must be speficied in the names of [Mozilla's profile definition](recommendation), while [CIS Benchmark](https://workbench.cisecurity.org/sections/665108/recommendations/1090244) uses [TLS specification name](https://tools.ietf.org/html/rfc8446).  [OpenSSL cipher table](https://www.openssl.org/docs/man1.1.1/man1/ciphers.html) can map between the Mozilla's names and TLS spec names.
- remediation: specify `type: custom` profile and specify cipher suites which are supported by OpenShift **and** permitted by the CIS Benchmark

#### kube API Operator
- check: if `.spec.unsupportedConfigOverrides` is defined, check all of `.spec.unsupportedConfigOverrides.servingInfo.cipherSuites` are permitted in the CIS Benchmark. If it is not defined, check all of `.spec.observedConfig.servingInfo.cipherSuites` are permitted in the CIS Benchmark.
- remediation: specify permitted cipher suites to  `.spec.unsupportedConfigOverrides.servingInfo.cipherSuites` 
    - **Note**: use `TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256` instead of `TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305`, and also, use `TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256` instead of `TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305` because the latters are old name. See [the RFC draft](https://tools.ietf.org/wg/tls/draft-ietf-tls-chacha20-poly1305/draft-ietf-tls-chacha20-poly1305-04-from-03.wdiff.html) in details.

#### OpenShift API Operator
as similar to kube API Operator; only one difference is the target resource (`openshiftapiservers.operator.openshift.io/cluster` instead of `kubeapiservers.operator.openshift.io/cluster`.

### Remaining Discussion Topics
#### more precise check for kube/OpenShift API Operators
Current rule does not support the OR rule.

- precise rule (as described avobe): if `.spec.unsupportedConfigOverrides` is defined, check all of `.spec.unsupportedConfigOverrides.servingInfo.cipherSuites` are permitted in the CIS Benchmark. If it is not defined, check all of `.spec.observedConfig.servingInfo.cipherSuites` are permitted in the CIS Benchmark.
- current rule: check all of `.spec.unsupportedConfigOverrides.servingInfo.cipherSuites` are permitted in the CIS Benchmark.

I don't know how to specify this type of `or` rule...I'd appreciate it if someone knows how to write this with `yamlfile_value` template.

#### description of each rule
The description that I wrote here should be also written in `rule.yml`.  Is there any standard to write the description, rationale and ocli in `rule.yml`?

ocp4/profiles/cis.profile
```diff
@@ -157,7 +157,14 @@ selections:
     - kubelet_configure_tls_key
     # Like kubelet_disable_readonly_port but check for .apiServerArguments["kubelet-client-key"]
   # 4.2.13 Ensure that the Kubelet only makes use of Strong Cryptographic Ciphers
+    - kubelet_configure_tls_cipher_suites_ingresscontroller
+    - kubelet_configure_tls_cipher_suites_kubeapiserver_operator
+    - kubelet_configure_tls_cipher_suites_openshift_authentication
```
JAORMX, line 162: This rule seems to be missing and it's failing when building the image because of it:

```
 ValueError: Rule 'kubelet_configure_tls_cipher_suites_openshift_authentication' was not found in the benchmark. Please remove rule 'kubelet_configure_tls_cipher_suites_openshift_authentication' from profile 'cis' before proceeding. 
```

tmishina:
I added OCIL clauses, but I failed to suppress jinja2 expansion (`{{{` `}}}'`).
Does anyone know how to do so? thanks in advance.

JAORMX:
@tmishina folks are leaving on PTO for a couple of weeks. Let's get back to this when folks are back.

JAORMX:
I am still fairly concerned at this check and recommendation. Even though we're following what CIS originally proposed, this is a custom TLS profile which I'm not sure if the OpenShift team will support. I'd be more comfortable recommending the intermediate profile which is supported. Instead, we should reach to CIS and see if it's fine.

mrogers950:
I think the OCP recommendation and check for this rule needs to mirror what is in the Kubernetes 1.6 benchmark. It has 4.2.13 focused on the TLSCipherSuites setting for the kubelet, ensuring that weak ciphers and CBC mode is excluded.
This is made a little more complicated by the defaulting of kubelet.conf, as it doesn't normally include a TLSCipherSuites option at all, so I can see the check composed similar to the following:
- A node-level yaml probe of /etc/kubernetes/kubelet.conf
  - if kubelet.conf contains TLSCipherSuites option, and it contains in its set anything other than the allowed set, fail
  - If kubelet.conf does not contain TLSCipherSuites, then fail (this is the default scenario). 
  - remediation would set the CIS recommended set.
 
The default that the kubelet uses is from: https://github.com/kubernetes/kubernetes/blob/faaae714274c2fd0eb6ce71a77d54106b7689e3d/staging/src/k8s.io/component-base/cli/flag/ciphersuites_flag.go#L29 excluding the ciphers from https://github.com/kubernetes/kubernetes/blob/faaae714274c2fd0eb6ce71a77d54106b7689e3d/staging/src/k8s.io/component-base/cli/flag/ciphersuites_flag.go#L52
This still leaves a few CBC ciphers, and the CIS general advice for cipher selection is to exclude all CBC. So in the case of the default, I think we should consider that a fail.

The checks for api/ingress will still be useful but I think for other sections (There is an API cipher check needed for 1.2.35). But for 4.2.13 in particular we should keep it focused on just the kubelet ciphers.

tmishina:
I wrote down a logic to check kubelet cipher suites.
https://github.com/tmishina/content/wiki/cis-4.2.13-(trial-2)#check-logic

I think the logic below is not supported by current `yamlfile_value` template...is there any workaround for this?

```
kubeletconfig <- kubectl get kubeletconfig
if kubeletconfig is null; then
  kubeletconfig <- /etc/kubernetes/kubelet.conf
fi
```

mrogers950:
> I wrote down a logic to check kubelet cipher suites.
> https://github.com/tmishina/content/wiki/cis-4.2.13-(trial-2)#check-logic
> 
> I think the logic below is not supported by current `yamlfile_value` template...is there any workaround for this?
> 
> ```
> kubeletconfig <- kubectl get kubeletconfig
> if kubeletconfig is null; then
>   kubeletconfig <- /etc/kubernetes/kubelet.conf
> fi
> ```

It's not possible to make a fall-back with two different check types to my knowledge, so a rule will either have to be a platform check or a node check. In this case, looking at just /etc/kubernetes/kubelet.conf with a node check should suffice. Using the platform check to examine MachineConfigs might not be as reliable (I'm thinking that looking through the whole list might have older mc entries, plus the rule cannot select a resource by label).

The rest of the logic you noted sounds good to me.

applications/openshift/kubelet/kubelet_configure_tls_cipher_suites/ocp4/e2e-remediation.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+cat >> /etc/kubernetes/kubelet.conf << EOF
+tlsCipherSuites:
+- TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
+- TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
+- TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
+- TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
+EOF
```
JAORMX, line 9: Note that the remediation does not run on the host itself. So... while this will run on the CI node, it doesn't really do anything but create this file on the test-runner. If you want to set this you'd need to set it via a custom `KubeletConfig` object: https://docs.openshift.com/container-platform/4.6/rest_api/machine_apis/kubeletconfig-machineconfiguration-openshift-io-v1.html or via `MachineConfig` directly

tmishina: Thank you, I misunderstood the behaviour of remediation run.

As far as I investigated, each of workaround has pros and cons.

### Using KubeletConfig
- document: [TLS configuration in OpenShift Container Platform \- Red Hat Customer Portal](https://access.redhat.com/articles/5348961)
- `.metadata.name` of `KubeletConfig` can be arbitrary defined (with specifying it as a label value for `machineconfigpool`), but names in check rule and remediation script must be predefined.  The check script may make false alert when user explicitly defines `KubeleteConfig` with a name which is not identical to the one in the check script.

### Using machineconfig
There are several `MachineConfig` resources in a CRC cluster.

```sh
$ oc get machineconfig
NAME                                               GENERATEDBYCONTROLLER                      IGNITIONVERSION   AGE
00-master                                          721141df0005d407d15b6664982f22be0604eb72   3.1.0             67d
00-worker                                          721141df0005d407d15b6664982f22be0604eb72   3.1.0             67d
01-master-container-runtime                        721141df0005d407d15b6664982f22be0604eb72   3.1.0             67d
01-master-kubelet                                  721141df0005d407d15b6664982f22be0604eb72   3.1.0             67d
01-worker-container-runtime                        721141df0005d407d15b6664982f22be0604eb72   3.1.0             67d
01-worker-kubelet                                  721141df0005d407d15b6664982f22be0604eb72   3.1.0             67d
...
```

In my understanding all of the `XX-XX-kubelet` `MachineConfig` resources should be checked.  However, [API Resource Collector of Compliance Operator](https://github.com/openshift/compliance-operator/blob/master/cmd/manager/api-resource-collector.go) does not support  fetching multiple resources (same issue happens [on authoring section 5 rules](https://github.com/ComplianceAsCode/content/issues/6529#issuecomment-773846031)), 

Any comments or suggestions? thanks in advance.

JAORMX: @mrogers950 any idea what we could do here? ^^

ocp4/profiles/cis.profile
```diff
@@ -169,7 +169,13 @@ selections:
     - kubelet_configure_tls_key
     # Like kubelet_disable_readonly_port but check for .apiServerArguments["kubelet-client-key"]
   # 4.2.13 Ensure that the Kubelet only makes use of Strong Cryptographic Ciphers
+    - kubelet_configure_tls_cipher_suites
```
mrogers950, line 172: Since this is a node-level check it should be in the cis-node.profile instead

applications/openshift/kubelet/kubelet_configure_tls_cipher_suites/ocp4/e2e-remediation.sh
mrogers950: The remediation could still be a mc/KubeletConfig even though the check is looking at the /etc/kubernetes/kubelet.conf on the node. This way we're checking the running config of the kubelet as opposed to a MC that is only describing the intended state of the machines, so I think it makes for a more accurate check.

tmishina: I tried to apply KubeletConfig to confirm that applying KubeletConfig also changes `/etc/kubernetes/kubelet.conf`, but the method described in [this document](https://docs.openshift.com/container-platform/4.6/scalability_and_performance/recommended-host-practices.html#create-a-kubeletconfig-crd-to-edit-kubelet-parameters_) does not work on my CodeReady Container environment.  I pushed the remediation script to check the remediation script behaviour on CI test enviornment.

ocp4/profiles/cis.profile
tmishina: @mrogers950 thank you, I changed target profile from `cis` to `cis-node`.
Now the check looks fine (no error happened in the CI test).  Are current description of `rule.yml` (comment and logic), test and remediation script OK to be merged?

applications/openshift/kubelet/kubelet_configure_tls_cipher_suites_ingresscontroller/rule.yml
```diff
@@ -0,0 +1,42 @@
+documentation_complete: true
+
+prodtype: ocp4
+
+title: "Ensure that the Kubelet only makes use of Strong Cryptographic Ciphers"
```
mrogers950, line 5: For these rules being added that we are leaving out of the profile for now (ingresscontroller, kubeapiserver_operator, openshiftapiserver_operator), let's update the titles and descriptions and related text to be more specific about the component. i.e., this would be "Ensure that the Ingress Controller only.."

mrogers950: Otherwise, the rest of it looks good to me!

tmishina: Thank you for pointing out, I modified the titles.
Also, I reorganized my commits, and done `git rebase master`.

@mrogers950 @JAORMX 
Could you review the final content and merge it if it's OK? thanks.

applications/openshift/kubelet/kubelet_configure_tls_cipher_suites_kubeapiserver_operator/rule.yml
```diff
@@ -0,0 +1,42 @@
+documentation_complete: true
+
+prodtype: ocp4
+
+title: "Ensure that the Kubernetes API Server only makes use of Strong Cryptographic Ciphers"
+
+description: |-
+  Ensure that the Kubernetes API Server is configured to only use strong cryptographic ciphers.
```
mrogers950, line 8: More accurately, this is covering the Kubernetes API Server _Operator_ TLS config. The API Server itself is covered by other rules. But for this, it is concerned with the serving endpoints of the operator.

applications/openshift/kubelet/kubelet_configure_tls_cipher_suites_openshiftapiserver_operator/rule.yml
```diff
@@ -0,0 +1,42 @@
+documentation_complete: true
+
+prodtype: ocp4
+
+title: "Ensure that the OpenShift API Server only makes use of Strong Cryptographic Ciphers"
+
+description: |-
+  Ensure that the OpenShift API Server is configured to only use strong cryptographic ciphers.
```
mrogers950, line 8: Same comment here, this is covering the OpenShift API Server Operator TLS config.

applications/openshift/kubelet/kubelet_configure_tls_cipher_suites_kubeapiserver_operator/rule.yml
tmishina: @mrogers950 I agree that "Operator" should be added; I updated the content (both for kube API server operator and OpenShift API server operator).

JAORMX:
@openscap-ci test this please

JAORMX:
We just need the OpenSCAP CI to return results before merging.

JAORMX:
@openscap-ci test this please

ggbecker:
@openscap-ci test this pleases

PR #6437, opened 2020-11-27
More precise modified time comparison in "configure_crypto_policy"

yuumasato:
#### Description:

- Change the way we compare the timestamps of `/etc/crypto-policies/config` and `/etc/crypto-policies/state/current` to be a bit more precise.
  - Instead of comparing the difference between each file and current time, and ensuring that time difference to `config` file is greater (i.e. `config` file was changed before `current` file).
  - Let's directly compare their timestamp (in seconds since epoch), and ensure that `config` timestamp is less than `current` timestamp.

#### Rationale:

- There is no way to calculate the `time_difference` against the same moment in time.

### DIFF
--- a/linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/tests/config_and_current_same_time.pass.sh
+++ b/linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/tests/config_and_current_same_time.pass.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+# platform = Red Hat Enterprise Linux 8
+
+# IMPORTANT: This is a false negative scenario.
+# File /etc/crypto-policies/config is newer than /etc/crypto-policies/state/current, thus incompliant,
+# but the OVAL evaluation restuls in pass.
+# With a precision of seconds in OVAL, there is not really much we can do to detect this.
+update-crypto-policies --set "DEFAULT"
+touch /etc/crypto-policies/config
--- a/linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/oval/shared.xml
+++ b/linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/oval/shared.xml
@@ -19,16 +19,12 @@
     <unix:filepath datatype="string">/etc/crypto-policies/config</unix:filepath>
   </unix:file_object>
 
-  <local_variable id="variable_crypto_policies_current_file_age" version="1" comment="Age of /etc/crypto-policies/state/current" datatype="int">
-      <time_difference format_2="seconds_since_epoch">
-        <object_component object_ref="crypto_policies_current_file" item_field="m_time"/>
-      </time_difference>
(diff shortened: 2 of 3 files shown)


linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/tests/config_and_current_same_time.pass.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+# platform = Red Hat Enterprise Linux 8
+
+# This is false negative scenario
```
yuumasato, line 4: @matusmarhefka @mildas Do you see value in this test scenario?
Strictly speaking, this should be a `fail` scenario. But AFAIK, there is no way to detect this, as smallest time unit in OVAL is one second.

matusmarhefka: I see, if we want to fake the files to have the same modification timestamps I would do the following trick:
```
update-crypto-policies --set "DEFAULT"
CDATE=$(date +%Y%m%d%H%M)
touch -a -m -t "$CDATE" /etc/crypto-policies/config /etc/crypto-policies/state/current
```

matusmarhefka: This is strictly to just test the situation when file modification time is the same, but I understand that this is OVAL limitation and we cannot really catch situations when modification time difference is less than 1s.

yuumasato: Thanks for the tip for the timestamp, I'll add this to the test scnerio.
But my concern was more about whether we should keep a test like this, to illustrate the situation we cannot cover well.

mildas: I don't know what you mean with `false negative scenario` in this case (shouldn't it be `false positive`? but probably I'm just confused). However, I like this test scenario and I see value in it because we can test that this specific case is passing and not causing failures from <1sec roundings.

The next comment about precision of seconds is very important there and probably I would add to it something like `WARNING:` or `IMPORTANT:` just to emphasize its importance.

yuumasato: @mildas Hmm, I think it is a false negative, as it reports a no-issue situation where there is one.

> However, I like this test scenario and I see value in it because we can test that this specific case is passing and not causing failures from <1sec roundings.

I think the rounding issue is solved in 3df3833f8191f8c261d02f9dd958a23f2619bfbb, it was caused by comparing the timestamp of the file to the current time.

In this test scenario, the <1sec difference is an actual issue.

PR #6438, opened 2020-11-27
Bump RHEL7 STIG version to V3R1 and update stig_overlay.xml.

ggbecker:
#### Description:

- Bump RHEL7 STIG version to V3R1 and update stig_overlay.xml.


jan-cerny:
@openscap-ci test this please

PR #6441, opened 2020-11-30
Add custom OVAL for partition_for_tmp to check for tmp.mount service.

ggbecker:
#### Description:

- Create custom OVAL check for `/tmp` partition. This enables the check to comply with checking of `tmp.mount` service status.

#### Rationale:

- Reference: https://www.stigviewer.com/stig/red_hat_enterprise_linux_7/2020-09-03/finding/V-204496

#### Questions:

- I'm not entirely sure if we need such check. I suspect that `partition_test` is enough which comes from the `mount` template. What do you think?

### DIFF
--- a/linux_os/guide/system/software/disk_partitioning/partition_for_tmp/oval/shared.xml
+++ b/linux_os/guide/system/software/disk_partitioning/partition_for_tmp/oval/shared.xml
@@ -1,0 +1,32 @@
+<def-group>
+    <definition class="compliance" id="partition_for_tmp" version="1">
+      {{{ oval_metadata("If stored locally, create a separate partition for
+        /tmp. If /tmp will be mounted from another
+        system such as an NFS server, then creating a separate partition is not
+        necessary at this time, and the mountpoint can instead be configured
+        later.") }}}
+      <criteria operator="OR">
+        <criterion test_ref="test_tmp_partition" comment="/tmp on own partition" />
+        <criterion comment="tmp.mount is running" test_ref="test_service_running_tmp.mount" />
+      </criteria>
+    </definition>
+    <linux:partition_test check="all" check_existence="all_exist"
+    id="test_tmp_partition" version="1" comment="/tmp on own partition">
+      <linux:object object_ref="object_mount_tmp_own_partition" />
+    </linux:partition_test>
+    <linux:partition_object id="object_mount_tmp_own_partition" version="1">
+      <linux:mount_point>/tmp</linux:mount_point>
+    </linux:partition_object>
+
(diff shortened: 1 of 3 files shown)


yuumasato:
I would make the check for `tmp.mount` service a new rule.

If there is a second way to have `/tmp` in a separate partition. Which one would be the remediated state of this rule?

Also, the `tmp.mount` service doesn't play well with current rules for partition options. With the `systemd.mount` the [options](https://www.freedesktop.org/software/systemd/man/systemd.mount.html#Options) are defined in the `Unit file`.

ggbecker:
> I would make the check for `tmp.mount` service a new rule.
> 
I tried to use the service enabled template, but `tmp.mount` is kind of a special service and doesn't fit with our existent template. Also it's an `OR` condition between these two items, I don't know how it would be handled as two separate rules.

yuumasato:
> I tried to use the service enabled template, but `tmp.mount` is kind of a special service and doesn't fit with our existent template.

Yeah, some different configuration is required in a unit file.

> Also it's an `OR` condition between these two items, I don't know how it would be handled as two separate rules.

Handling `OR`s in a profile is tricky, the choice of what path to take is not easy automate.
Checking that either condition is satisfied is fine in a rule. But when it comes to fixing it, what should the rule do `fstab` or `tmp.mount`?

I **think** the way these situations were dealt with in the past was by making the profiles opinionated.
The profile author, or SME, makes the default choice, and leads the profile that way.

Ideally, it would be easy to tailor the profile to the other choices, or have pre-defined tailorings (or other profiles).
But I don't know how that would be done without getting too complex.

jan-cerny:
@ggbecker Any updates?

ggbecker:
> @ggbecker Any updates?

Not really, I don't know if this fix is correct and if it makes sense.

jan-cerny:
@ggbecker Let's summarize what we already know.

Before this change the rule used the `mount` template which uses OVAL partition test which reads data from `/proc/mounts`.

The STIG rule allows 2 options: enabled tmp.mount service or a separate partition defined in /etc/fstab.

If you don't have any of these, you won't find any `/tmp` entry in `/proc/mounts` so the OVAL partition test will fail.
If you enable tmp.mount systemd unit, a `/tmp` entry will appear in  `/proc/mounts` so the OVAL partition test will pass.
If you instead create a separate partition, a `/tmp` entry will also appear in  `/proc/mounts` so the OVAL partition test will pass.

That means that the partition test covers both situation required by the rule. I have checked that on a RHEL 7 VM.

It seems that the OVAL partition check in `mount` template is sufficient for the rule.

@ggbecker @yuumasato What else do we need to know?

jan-cerny:
@ggbecker @yuumasato it seems you missed this one

jan-cerny:
@ggbecker please resolve the conflicts

jan-cerny:
Closing because of inactivity.

PR #6444, opened 2020-12-01
fix oval of grub2_kernel_trust_cpu_rng

vojtapolasek:
#### Description:

- an old name was left as oval ID

- also changed platform to machine, as grub2 platform is inherited from the group

#### Rationale:


mildas:
LGTM, it works properly and not resulting in `notchecked`. Thanks

Issue #6446, opened 2020-12-02
CentOS8 derivative product returning 'not applicable' result for every rule


Chalmiller:
#### Description of problem:
I am creating the CentOS8 derivative product via `./build_product rhel8 --derivatives`, but when running the profile with `sudo oscap xccdf eval --results results.xml --oval-results --report report.html --profile ccs_centos8 build/ssg-centos8-ds.xml` the results are all being returned as `not applicable`. 

I previously generated a centOS7 derivative product in the same manner, but it ran successfully so I'm unsure why this isn't the same?

Thank you for the help!

Chalmiller:
Awesome as always @ggbecker! Thanks again for the help!

PR #6447, opened 2020-12-02
Translate CPE check ids for derivative products.

ggbecker:
#### Description:

- Translate CPE check ids for derivative products.

#### Rationale:

- Follow up fix from: #6304. Since the introduction of dynamically generated CPE, there is an issue for CPEs from derivative products. Without this fix checks always return `notapplicable` for derivative products.

- Applies the same logic as regular CPE id transformation from: https://github.com/ComplianceAsCode/content/blob/master/build-scripts/cpe_generate.py#L191

- Fixes: #6446


Chalmiller:
Just tested the above code changes and they appear to have fixed the bug. Thank you again @ggbecker!

PR #6449, opened 2020-12-03
RHCOS4: Fix sysctl remediations and add tests

JAORMX:
They were missing an expected newline at the end.

This will ensure that we're running the remediations as expected.


JAORMX:
Pull-request updated, HEAD is now a5bcbb701c081c6e7c6fea2a875985b756459b97

jhrozek (approved):
LGTM

PR #6450, opened 2020-12-03
kubernetes: Fix kernel argument template

JAORMX:
The jinja was missing braces.

This also adds an e2e test to one of the rules that uses this so we
don't regress.

We'll add the rest of the e2e tests in a subsequent commit.


jhrozek (approved):
LGTM

PR #6453, opened 2020-12-03
Add extended definition to check for OpenSSH 7.4 in sshd_disable_compression

ggbecker:
#### Description:

- Add extended definition to check for OpenSSH 7.4 or higher in `sshd_disable_compression`

#### Rationale:
- The pre-authentication vulnerability was fixed since OpenSSH 7.4 so the SSH configuration represented by this rules doesn't make sense if the version is equals or higher to that.

- [Changelog reference](https://www.openssh.com/releasenotes.html): 
```
 * sshd(8): Remove support for pre-authentication compression.
   Doing compression early in the protocol probably seemed reasonable
   in the 1990s, but today it's clearly a bad idea in terms of both
   cryptography (cf. multiple compression oracle attacks in TLS) and
   attack surface. Pre-auth compression support has been disabled by
   default for >10 years. Support remains in the client.
```

- STIG requirement: https://www.stigviewer.com/stig/red_hat_enterprise_linux_7/2020-09-03/finding/V-204602

#### Additional Information

- The test scenarios still have to be updated, hence the draft PR.


matejak:
This looks really great, especially the test coverage. Is there any reason for keeping it as a draft PR?

ggbecker:
> This looks really great, especially the test coverage. Is there any reason for keeping it as a draft PR?

On my tests, the package downgrade commands were not working. I tested on RHEL7.9 and I'm not sure if it's not possible to downgrade them anymore or if I was doing something wrong.

ggbecker:
I've flipped to ready to review. The solution proposed is to remove all the `fail` test scenarios as we cannot reproduce them anymore because the package downgrade doesn't work anymore for recent versions of RHEL (e.g 7.9).

PR #6458, opened 2020-12-07
Use /proc/sys/kernel/osrelease instead of uname test to check system arch

yuumasato:
#### Description:

- Check `/proc/sys/kernel/osrelease` to get arch of the system.
- Follow up from #6442 
#### Rationale:

 - `uname` test is not supported in offline mode and breaks the compliance-operator workflow


jan-cerny:
Looks fine to me, the def returns true and matches x86_64 on my system.

JAORMX:
Awesome! I'll give this a try!

PR #6460, opened 2020-12-07
CIS profiles update for multiple OS

truzzon:
#### Description:

- Adding cis.profile for sle12, which was missing. (Took huge inspiration from rhel7)
- Update of sle15 profile to better represent the current state of the Benchmark


jan-cerny:
@openscap-ci ok to test

@openshift-ci-robot 
/ok-to-test

sle15/profiles/cis.profile
```diff
@@ -696,184 +794,217 @@ selections:
...
+    ###       configured (Automated)
+    - var_accounts_passwords_pam_faillock_unlock_time=900
+    - var_accounts_passwords_pam_faillock_deny=5
+    - accounts_passwords_pam_faillock_unlock_time
+    - accounts_passwords_pam_faillock_deny
 
-    ### 5.3.3 Ensure password reuse is limited (Scored)
+    ### 5.3.3 Ensure password reuse is limited (Automated)
+    - var_password_pam_unix_remember=5
+    - accounts_password_pam_unix_remembe
```
ggbecker, line 832:
```suggestion
    - accounts_password_pam_unix_remember
```

ggbecker:
@openscap-ci test this please

ggbecker:
@truzzon the CI indicates a problem with the rule `account_disable_post_pw_expiration`. It is not included in the SUSE12 benchmark. That means you have to update: https://github.com/cipherboy/scap-security-guide/blob/master/linux_os/guide/system/accounts/accounts-restrictions/account_expiration/account_disable_post_pw_expiration/rule.yml#L3 to include `sle12`. There might be more cases of the same. I suggest you try to build it locally first otherwise it can get too many iterations until the build finishes successfully.

I suggest reading the documentation on how to build the project here: https://complianceascode.readthedocs.io/en/latest/manual/developer/02_building_complianceascode.html

This way you will be able to quickly iterate over the problems.

truzzon:
@ggbecker Will the mentioned error show up one at a time? So should I check the other used rules as well? Or will it stay the only one?

ggbecker:
@truzzon there is still something going on with the build. I'll take a look later and see what I can do.

But if you manage to setup your developer environment. you can run `./build_product sle12 sle15 --datastream-only` and that will successfully finish the build.

carlosmmatos:
@truzzon @ggbecker please rebase and resolve any issues. Thanks!

truzzon:
Hello, I merged the upstream into my local branch. I'll push it in a few days. Then the issue should be resolved.

truzzon:
Hello, have finished up the first steps and added some checks. I also have an updated profile for rhel7 and 8 waiting. Some files still have placeholder texts, that need to be filled. Can someone help me out there?
Please review the new commits carefully. I only had the CIS Documents as a reference and changed existing values to suite its needs. I don't want to break something for someone else.
The local Test Builds where working fine for rhel7,rhel8,sle12 and sle15.

truzzon:
Hello, I found the CCE Reference table. What do I do, if a check that is Part of CIS is not referenced by CCE, and therefore will cause the build checks to fail?

truzzon:
@ggbecker What are the dependencies of the CCE Reference Check? Is the existence if the rule enough or does it need to be called in through a profile? 
I removed all rules regarding issue.net, because it is usually only called by telnet, which should be disabled anyway.

But I still have the problem, that some things are not referenced in CCE at all.
E.g. Web servers besides httpd (nginx, apache2,lighthttpd,etc.) or the user, group and mode for ssh keyfiles in "/etc/ssh/ssh_host_*_key".

Another example is nftables. There is not a single CCE Reference. So do I have to remove everyting?

What is your opinion on cases like this and how should I proceed?

ggbecker:
> @ggbecker What are the dependencies of the CCE Reference Check? Is the existence if the rule enough or does it need to be called in through a profile?

- The CCE check (if it exists) will only trigger if it selected in a profile.
- Duplicated CCE will always be checked
- Rules that do not have CCE assigned to and belong to RHEL realm should have a CCE assigned from this pool: https://github.com/ComplianceAsCode/content/blob/master/shared/references/cce-redhat-avail.txt . For other products, you should contact the responsible person who has the pool of CCE available for these products. There was someone assigning CCE for SLE products, I don't remember exactly who it was but you can find them in this repository throughout the history of it.

> I removed all rules regarding issue.net, because it is usually only called by telnet, which should be disabled anyway.
> 
> But I still have the problem, that some things are not referenced in CCE at all.
> E.g. Web servers besides httpd (nginx, apache2,lighthttpd,etc.) or the user, group and mode for ssh key files in "/etc/ssh/ssh_host_*_key".
> 
> Another example is nftables. There is not a single CCE Reference. So do I have to remove everything?
> 
> What is your opinion on cases like this and how should I proceed?

As I said, for RHEL products they should have a new CCE assigned from the available pools of CCE issued by NIST.

> 
> In my next push, I will correct the references if available and remove the duplicates. I'll set comments in the files where no Reference is available.
> 
> The build test will fail but I hope, that it will make my problem more understandable.
> 
> P.S.: https://nvd.nist.gov/config/cce/index
> I'm using the excel sheets available here as reference point.

I think we should take a step back here and consider a simpler approach to these changes. As you can see there is already a bunch of conflicts and they will grow even more with time. I suggest breaking this pull request into smaller pieces, to be more product centric and easier to review. Let's start small, and then you can gain confidence in the requirements to get things merged. The update of CIS profile for SLE15 seems to be a good start for that. There are a lot of mixed-up things that make it hard to review and get things.

truzzon:
Hello @ggbecker,

I'm closing this PR for now.

I have to rework the structure of my additions and break them up into several smaller PRs as you recommended.

Since I haven't had much time lately, I'll have to catch up on the latest changes and see if topics I was adressing were already implemented. 

Best regards
Nico

PR #6462, opened 2020-12-09
Set platform package mapping to None for not_s390x_arch CPE.

ggbecker:
#### Description:

- Set platform package mapping to None for not_s390x_arch CPE.

#### Rationale:

- Remediation that would check the rpm package condition always fail because `not_s390x_arch` does not translate to anything related to `rpm` database.


yuumasato (approved):
Great catch, most of the existing CPE's were about packages installed, but recently we've seen the need for other types of checks for CPEs, like checking the contents of a file.

This will fix the error.
But we'll need better support for Bash and Ansible for these "other" platforms.

PR #6463, opened 2020-12-10
Fix RHEL8 CIS Benchmark version

yuumasato:
#### Description:

- Fix version of RHEL8 CIS Benckmark
  Nice catch @ggbecker


ggbecker (approved):
v1.1.0 has not been published yet so it makes sense that we keep the current released one. Thanks for the changes.

PR #6464, opened 2020-12-10
Remove useless build-time noise

matejak:
Make the build experience clean - remove useless warnings or announcements.

- Don't warn that a rule doesn't have an OVAL - there are many rules like that, some of them are like that on purpose, and ones that shouldn't be OVAL-less probably get lost in the noise.
- Don't print messages about how many remediations/OVALs were merged / collected.
- Don't warn about rules that don't reference PCI-DSS.

What remains is `oscap` warning about remote resources. It looks like that we could use an opt-out switch that would silence those warnings, as we don't want to download anything during the content build. Or, in a more extreme note, we aim to avoid to use the scanner in order to produce guides. Anyway, I don't think that those warnings could be removed without introducing drastic measures to the content build.

### DIFF
--- a/build-scripts/combine_remediations.py
+++ b/build-scripts/combine_remediations.py
@@ -80,8 +80,6 @@ def main():
     remediation.write_fixes_to_dir(fixes, args.remediation_type,
                                    args.output_dir)
 
-    sys.stderr.write("Collected %d %s remediations.\n" % (len(fixes), args.remediation_type))
-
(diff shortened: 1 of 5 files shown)


jan-cerny:
>  Or, in a more extreme note, we aim to avoid to use the scanner in order to produce guides. 

That would be the best, I assume that we could process the file directly by xsltproc.

Nice improvement!

PR #6465, opened 2020-12-10
Fix some missing extend_definition dependencies

matejak:
Fixed cases of products having rules with checks that require checks that are were not enabled for the product.

### DIFF
--- a/linux_os/guide/system/software/integrity/disable_prelink/rule.yml
+++ b/linux_os/guide/system/software/integrity/disable_prelink/rule.yml
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: ol7,rhel7
+prodtype: rhcos4,ol7,rhel7,wrlinux1019
 
 title: 'Disable Prelinking'
 
--- a/shared/checks/oval/install_mcafee_hbss.xml
+++ b/shared/checks/oval/install_mcafee_hbss.xml
@@ -4,7 +4,16 @@
     <metadata>
       <title>Install McAfee Host-Based Intrusion Detection Software (HBSS)</title>
       <affected family="unix">
-        <platform>multi_platform_all</platform>
+	<platform>multi_platform_debian</platform>
+	<platform>multi_platform_fedora</platform>
+	<platform>multi_platform_opensuse</platform>
+	<platform>multi_platform_ol</platform>
+	<platform>multi_platform_rhcos</platform>
+	<platform>multi_platform_rhel</platform>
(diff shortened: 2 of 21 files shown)


linux_os/guide/system/software/integrity/disable_prelink/rule.yml
```diff
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: ol7,rhel7
+prodtype: rhcos4,ol7,rhel7,wrlinux1019
```
yuumasato, line 3: `disable_prelink` is not enabled for `rhel8`, so I wonder what requires that on `rhcos4`.

shared/checks/oval/install_mcafee_hbss.xml
```diff
@@ -4,7 +4,16 @@
...
+	<platform>multi_platform_debian</platform>
+	<platform>multi_platform_fedora</platform>
+	<platform>multi_platform_opensuse</platform>
+	<platform>multi_platform_ol</platform>
+	<platform>multi_platform_rhcos</platform>
+	<platform>multi_platform_rhel</platform>
+	<platform>multi_platform_rhv</platform>
+	<platform>multi_platform_sle</platform>
+	<platform>multi_platform_ubuntu</platform>
+	<platform>multi_platform_wrlinux</platform>
```
yuumasato, line 16: @matejak Could you clarify why this is needed?
Why `multi_platform_all` doesn't work?

matejak: Those are Linux OS products. You don't want this OVAL to appear in products s.a. OSP, Firefox and whatnot. I think that a shared OVAL makes it into a benchmark even if there is no rule for it, and then it can happen that it is "incomplete". As no makes use of it, it's not a real issue, but as we don't distinguish between useful and useless OVALs, a missing extending definition is treated in the same way for all OVALs.

linux_os/guide/system/software/integrity/disable_prelink/rule.yml
matejak: So the answer is that `grub2_enable_fips_mode` needs it. Sounds legit :-)

shared/checks/oval/install_mcafee_hbss.xml
yuumasato: Thanks for the clarification, and I think your observations are correct.

yuumasato (approved):

PR #6468, opened 2020-12-11
Add ignition version to coreos_kernel_option template

JAORMX:
While not strictly necessary, it does show up in the kube object. So
let's add it for completeness.


jhrozek (approved):
LGTM

PR #6475, opened 2020-12-14
Add custom OVAL check file_permissions_multus_conf to check stricter permissions

ggbecker:
#### Description:

- Add custom OVAL check file_permissions_multus_conf to check stricter permissions

#### Rationale:

- Some security requirements say that file must have a specific file permissions mode or stricter.

### DIFF
--- a/applications/openshift/master/file_permissions_multus_conf/oval/shared.xml
+++ b/applications/openshift/master/file_permissions_multus_conf/oval/shared.xml
@@ -1,0 +1,27 @@
+<def-group>
+    <definition class="compliance" id="file_permissions_multus_conf" version="1">
+      {{{ oval_metadata(description="This test makes sure that ^/var/run/multus/cni/net.d/.*$ has mode 0644. If the target file or directory has an extended ACL, then it will fail the mode check.") }}}
+      <criteria>
+        <criterion comment="Check file mode of ^/var/run/multus/cni/net.d/.*$" test_ref="test_file_permissions_multus_conf" negate="true"/>
+      </criteria>
+    </definition>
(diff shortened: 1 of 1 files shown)


applications/openshift/master/file_permissions_multus_conf/oval/shared.xml
```diff
@@ -0,0 +1,27 @@
+<def-group>
+    <definition class="compliance" id="file_permissions_multus_conf" version="1">
+      {{{ oval_metadata(description="This test makes sure that ^/var/run/multus/cni/net.d/.*$ has mode 0644. If the target file or directory has an extended ACL, then it will fail the mode check.") }}}
```
redhatrises, line 3: Please come up with a more compliance-like statement here for why this checks exists. For example:

```Ensure all files in /var/run/multus/cni/net.d/ have permissions 0644 or stricter.```

redhatrises:
In general security and compliance realms, permissions of 644 are weak and should be 600 instead.

applications/openshift/master/file_permissions_multus_conf/oval/shared.xml
ggbecker: That comes from the rule.yml fields. https://github.com/ComplianceAsCode/content/blob/master/applications/openshift/master/file_permissions_multus_conf/rule.yml and are being populated by the template. This pull request is going to be replaced by https://github.com/ComplianceAsCode/content/pull/6476

ggbecker:
> In general security and compliance realms, permissions of 644 are weak and should be 600 instead.

In this case we should reevaluate the permissions set in https://github.com/ComplianceAsCode/content/blob/master/applications/openshift/master/file_permissions_multus_conf/rule.yml

I'm going to close this PR as it is superseded by #6476

PR #6477, opened 2020-12-14
OCP: Cleanup rules in section 1.1 of CIS profile

jhrozek:
NOTE: Please pay special attention to the last commit about cert and keys.
This is the only one that actually changed the probe, the rest is more or less
just cosmetic changes and making sure the rules run on master nodes only without
ignoring missing files.

#### Description:

- Hardens the rules in section 1.1 of the CIS standard by running the rules on
  master nodes only and dropping the `missing_file_pass: true` attribute except
  for several files under /var or /run like pidfiles which might be ephemeral
- Adjusts tests accordingly
- Uses wildcards instead of regexes in OCILs and description to make sure them
  usable better for copy-paste

#### Rationale:

- Let's make the CIS profile production-ready

### DIFF
--- a/applications/openshift/master/file_groupowner_cni_conf/rule.yml
+++ b/applications/openshift/master/file_groupowner_cni_conf/rule.yml
@@ -4,7 +4,7 @@ prodtype: ocp4
 
 title: 'Verify Group Who Owns The OpenShift Container Network Interface Files'
 
-description: '{{{ describe_file_group_owner(file="/etc/cni/net.d/.*", group="root") }}}'
+description: '{{{ describe_file_group_owner(file="/etc/cni/net.d/*", group="root") }}}'
 
 rationale: |-
     CNI (Container Network Interface) files consist of a specification and libraries for
@@ -20,9 +20,9 @@ identifiers:
 references:
     cis: 1.1.10
 
-ocil_clause: '{{{ ocil_clause_file_group_owner(file="/etc/cni/net.d/.*", group="root") }}}'
+ocil_clause: '{{{ ocil_clause_file_group_owner(file="/etc/cni/net.d/*", group="root") }}}'
 
-ocil: '{{{ ocil_file_group_owner(file="/etc/cni/net.d/.*", group="root") }}}'
+ocil: '{{{ ocil_file_group_owner(file="/etc/cni/net.d/*", group="root") }}}'
 
 template:
     name: file_groupowner
--- a/applications/openshift/master/file_owner_etcd_member/rule.yml
+++ b/applications/openshift/master/file_owner_etcd_member/rule.yml
@@ -7,10 +7,12 @@ title: 'Verify User Who Owns The etcd Member Pod Specification File'
 description: '{{{ describe_file_owner(file="/etc/kubernetes/static-pod-resources/etcd-pod-*/etcd-pod.yaml", owner="root") }}}'
 
 rationale: |-
-    If the etcd specification file contains the configuration of
-    the etcd highly-available key-value store which Kubernetes uses for
-    persistent storage of all of its REST API object. Protection of this
-    file is critical for OpenShift security.
+    The etcd pod specification file controls various parameters that
+    set the behavior of the etcd service in the master node. etcd is a
+    highly-available key-value store which Kubernetes uses for persistent
+    storage of all of its REST API object. You should restrict its file
+    permissions to maintain the integrity of the file. The file should be
+    writable by only the administrators on the system.
 
 severity: medium
 
--- a/applications/openshift/master/file_owner_etcd_member/rule.yml
+++ b/applications/openshift/master/file_owner_etcd_member/rule.yml
@@ -2,14 +2,15 @@ documentation_complete: true
 
 prodtype: ocp4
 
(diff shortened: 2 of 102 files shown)


applications/openshift/master/file_owner_etcd_member/rule.yml
```diff
@@ -2,14 +2,15 @@ documentation_complete: true
...
+description: '{{{ describe_file_owner(file="/etc/kubernetes/static-pod-resources/etcd-pod-*/etcd-pod.yaml", owner="root") }}}'
 
 rationale: |-
-    The ETCD specification file contains information about the configuration of the
-    OpenShift scheduler that is configured on the system. Protection of this file is
-    critical for OpenShift security.
+    If the etcd specification file contains the configuration of
+    the etcd highly-available key-value store which Kubernetes uses for
+    persistent storage of all of its REST API object. Protection of this
+    file is critical for OpenShift security.
```
redhatrises, line 13: Check each instances of this rationale. The english is off.

applications/openshift/master/file_owner_openshift_pki_cert_files/rule.yml
```diff
@@ -5,22 +5,25 @@ prodtype: ocp4
...
 
 description: |-
-  {{{ describe_file_owner(file="/etc/kubernetes/static-pod-resources/.*/.*/.*/tls\.crt", owner="root") }}}
+  {{{ describe_file_owner(file="/etc/kubernetes/static-pod-resources/*/*/*/tls.crt", owner="root") }}}
 
 rationale: |-
   OpenShift makes use of a number of certificates as part of its operation.
   You should verify the ownership of the directory containing the PKI
   information and all files in that directory to maintain their integrity.
+  The directory and files should be owned by the system administrator.
```
redhatrises, line 14: This doesn't need to be added to the rationale.

applications/openshift/master/file_permissions_etcd_member/rule.yml
```diff
@@ -2,16 +2,17 @@ documentation_complete: true
 
 prodtype: ocp4
 
-title: 'Verify Permissions on the ETCD Member Pod Specification File'
+title: 'Verify Permissions on the etcd Member Pod Specification File'
 
 description: |-
-    {{{ describe_file_permissions(file="/etc/kubernetes/static-pod-resources/etcd-pod-.*/etcd-pod.yaml", perms="0644") }}}
+    {{{ describe_file_permissions(file="/etc/kubernetes/static-pod-resources/etcd-pod-*/etcd-pod.yaml", perms="0644") }}}
```
redhatrises, line 8: What are the plans to move beyond 644? In security and compliance realm, you really don't check files that are system default file permissions, and the files themselves should really be 0600 for security.

applications/openshift/master/file_groupowner_cni_conf/rule.yml
```diff
@@ -4,7 +4,7 @@ prodtype: ocp4
 
 title: 'Verify Group Who Owns The OpenShift Container Network Interface Files'
 
-description: '{{{ describe_file_group_owner(file="/etc/cni/net.d/.*", group="root") }}}'
+description: '{{{ describe_file_group_owner(file="/etc/cni/net.d/*", group="root") }}}'
```
JAORMX, line 7: when introducing this, I was explicitly asked to add the "." so it matches the regex that's actually being used.

applications/openshift/master/file_permissions_etcd_member/rule.yml
JAORMX: They don't contain any sensitive information, they're merely a way to tell the kubelet how to start the etcd pods. any sensitive information is stored as kubernetes secrets with appropriate permissions.

jhrozek: also, CIS asks us to just check if they are not group or world writable, too.

applications/openshift/master/file_owner_etcd_member/rule.yml
jhrozek: Please check this version, I more or less reused the rationale from CIS with the upstream kubernetes specific path removed. If would like this version to be improved, I would appreciate a suggestion.

applications/openshift/master/file_owner_openshift_pki_cert_files/rule.yml
jhrozek: Thank you, removed.

jhrozek:
The current patches are removing more instances of missing_file_pass=true from the SDN related rules.

jhrozek:
`   * could not run steps: step e2e-aws-ocp4-cis failed: failed to acquire lease: status 502 Bad Gateway, status code 502 `

JAORMX:
@jhrozek seems like a legitimate issue:
```
     helpers.go:640: Result - Name: e2e-cis-node-worker-file-owner-master-admin-kubeconfigs - Status: SKIP - Severity: medium
    helpers.go:643: E2E-FAILURE: The expected result for the file_owner_master_admin_kubeconfigs rule didn't match. Expected 'map[master:PASS worker:SKIP]', Got 'SKIP' 
```

PR #6478, opened 2020-12-15
OCP4: Add CCEs to rules used by the CIS profile

JAORMX:
A lot of the CCEs were missing, this adds them.


jhrozek (approved):
ack, I verified all the identifiers are unique and all of them were taken from and removed from the reserved list.

JAORMX:
@openprivacy test this please

PR #6479, opened 2020-12-15
Fix FIPS checks for RHCOS

JAORMX:
This adds RHCOS as a FIPS certified OS (it inherits that from RHEL) and
it enables OVAL that's specific to checking RHCOS.

The relevant changes are:

modify sysctl check
-------------------

Given that in the Compliance Operator the OpenSCAP command runs in
offline mode, we can't use the sysctl probe directly. So, instead I
modified that check to use the value from
`/proc/sys/crypto/fips_enabled` directly.

Remove dracut checks
--------------------

Initramfs is managed by ostree, which in turn wraps dracut. It doesn't
use the dracut-fips module directly as that part is handled by ignition
itself. So, these checks have been removed.

### DIFF
--- a/linux_os/guide/system/software/integrity/certified-vendor/installed_OS_is_FIPS_certified/oval/shared.xml
+++ b/linux_os/guide/system/software/integrity/certified-vendor/installed_OS_is_FIPS_certified/oval/shared.xml
@@ -6,6 +6,7 @@
     <criteria comment="Installed operating system is a certified operating system" operator="OR">
       <extend_definition comment="Installed OS is RHEL7" definition_ref="installed_OS_is_rhel7" />
       <extend_definition comment="Installed OS is RHEL8" definition_ref="installed_OS_is_rhel8" />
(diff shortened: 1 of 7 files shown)


yuumasato (approved):
LGTM.

Not a big deal, but I noticed that the descriptions of rules [enable_fips_mode](https://github.com/ComplianceAsCode/content/blob/1e7fc874c5587fc1cde3c9c017e43a36ef4035db/linux_os/guide/system/software/integrity/fips/enable_fips_mode/rule.yml) and [sysctl_crypto_fips_enabled](https://github.com/ComplianceAsCode/content/blob/1e7fc874c5587fc1cde3c9c017e43a36ef4035db/linux_os/guide/system/software/integrity/fips/sysctl_crypto_fips_enabled/rule.yml) are a bit off from the RHCOS procedures, like the way how to enable FIPS mode and the OCIL to check `crypto.fips_enabled`.
These can be addressed at another opportunity if preferred.

jhrozek (approved):
lgtm

PR #6481, opened 2020-12-15
Add s390x (IBM Z) as a 64bit architecture

rmetrich:
#### Description:

- s390x (IBM Z) is not recognized as a 64bit architecture, causing some rules to fail.

Without this patch:
~~~
Title   Record Attempts to Alter Time Through stime
Rule    xccdf_org.ssgproject.content_rule_audit_rules_time_stime
Ident   CCE-27299-7
Result  fail
~~~

With this patch:
~~~
Title   Record Attempts to Alter Time Through stime
Rule    xccdf_org.ssgproject.content_rule_audit_rules_time_stime
Ident   CCE-27299-7
Result  pass
~~~
There are likely other rules impacted by not having IBM Z systems detected as 64bit systems.


yuumasato:
@openscap-ci add to whitelist

yuumasato (approved):
Thank you for the patch!

PR #6492, opened 2020-12-18
Bug 1909081: Add `overwrite: true` to all ignition configurations

JAORMX:
This is a parameter that's now required in ignition v3.1.0, as, if it's
not set, ignition may fail if there's already a file in the system and
you're trying to overwrite it.

We set it here as we expect compliant configurations to take precedence
over anything else.


jhrozek (approved):
/lgtm
tested with a bunch of remediations that overwrite files (e.g. sshd_config) and seemed to work OK with a 4.7 nightly cluster

PR #6495, opened 2020-12-18
Create new rules for ANSSI R39

jan-cerny:
#### Description:
ANSSI R39: Temporary directories dedicated to each account
Each user or service account must have its own temporary directory and dispose of it exclusively.

We can satisfy this requirement by setting up the temporary directories as polyinstantiated, according to https://access.redhat.com/blogs/766093/posts/3169121.

This PR adds new rules that check for the configuration changes.


yuumasato (changes requested):
The rules look good and the test scenarios pass, but I think you forgot to enable the SELinux boolean `polyinstantiation_enabled=1`.
Without it the polyinstantiation didn't work, and no other user than `root` was able to login.

DraugurHundur:
I'm confused why create the inst directory in /var/tmp under /var/tmp while you create the inst directory for /tmp under / (which may be a different partition) ?

yuumasato:
That's a good point. I guess it would be safer in (terms of partition capacity) to have the instance in `/tmp/tmp-inst`?

DraugurHundur:
I'm guessing because /tmp is wiped on reboot (whereas /var/tmp is not) this can cause an issue as the inst directory would have to be recreated on restart.

linux_os/guide/system/accounts/accounts-session/accounts_polyinstantiated_tmp/oval/shared.xml
```diff
@@ -4,6 +4,7 @@
     <criteria operator="AND" comment="Check Polyinstantiation of /tmp Directories">
       <criterion comment="Check that /tmp-inst exists and has mode 000" test_ref="test_tmp_inst" />
       <criterion comment="Check configuration of /tmp in /etc/security/namespace.conf file" test_ref="test_tmp_in_namespace_conf" />
+      <criterion comment="Check SELinux boolean polyinstantiation_enabled is enabled" test_ref="test_accounts_polyinstantiated_tmp_sebool" />
```
yuumasato, line 7: Should the check and remediation for sebool `polyinstantiation_enabled` be present in multiple rules?
It makes more sense embed it on `enable_pam_namespace`.
It could also become a rule of its own.

jan-cerny: Yes, but we already have a rule for this selinux boolean, it's a templated rule, its ID is sebool_polyinstantiation_enabled https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/selinux/selinux-booleans/sebool_polyinstantiation_enabled/rule.yml. But we're adding this rule to the profile.

I like the idea about embedding the enable_pam_namespace, but I would even go firther and merge the check from enable_pam_namespace into accounts_polyinstantiated_tmp and accounts_polyinstantiated_var_tmp.

What is your advise?

yuumasato: Oh, I missed that a rule for sebool `polyinstantiaion_enabled` already existed, and it uses an XCCDF Value for the value.

In this case, just selecting it with the appropriate value should do, right?

yuumasato:
> I'm confused why create the inst directory in /var/tmp under /var/tmp while you create the inst directory for /tmp under / (which may be a different partition) ?

@jan-cerny What do you think of the suggestion to create the instance dir in `/tmp/tmp-inst`? So that data intended for polyinstantiated `/tmp`, is actually put in  the same partition as /tmp` would.

linux_os/guide/system/accounts/accounts-session/accounts_polyinstantiated_tmp/oval/shared.xml
yuumasato: Sorry for the confusion and extra work.

jan-cerny: It should be enough when the profile selects all 4 rules.

yuumasato (approved):
LGTM,  thank you, @jan-cerny :)

PR #6496, opened 2020-12-21
Add rule to check if OS is 64-bit when supported by CPU

yuumasato:
#### Description:

- Add rule that checks if the system supports 64-bit operating systems
  - The rule fails if CPU supports for 64-bit operations, but installed OS is not a 64-bit one.

#### Rationale:
- Supports ANSSI BP-028 R10
  - Preference to 64-bit operating systems should be given when supported


jan-cerny (commented):
This looks fine. It passes on my 64 bit system.

yuumasato:
@jan-cerny Thank you for the review, rule description and rationale are updated.

jan-cerny:
@yuumasato thanks, LGTM

PR #6498, opened 2020-12-22
Add the sudoers_no_command_negation rule - ANSSI R62

matejak:
The rule makes sure that sudoers user specifications don't define commands using negations, which creates a false sense of security.

Again, remediations are questionable here - removing offending negations wouldn't fix any security issues, as the job to do in those cases is to redesign the user specification, or even to modify the system group layout,

### DIFF
--- a/linux_os/guide/system/software/sudo/sudoers_no_command_negation/oval/shared.xml
+++ b/linux_os/guide/system/software/sudo/sudoers_no_command_negation/oval/shared.xml
@@ -1,0 +1,26 @@
+<def-group>
+     <definition class="compliance" id="{{{ rule_id }}}" version="1">
+     {{{ oval_metadata("Check that sudoers doesn't contain command negations") }}}
+     <criteria operator="AND">
+	     <criterion comment="Make sure that no command in user spec contains negation" test_ref="test_{{{ rule_id }}}" />
+     </criteria>
+  </definition>
+
+  <ind:textfilecontent54_test check="all" check_existence="none_exist"
+  comment="Make sure that no user spec in sudoers has a runas spec that includes root or ALL"
+	  id="test_{{{ rule_id }}}" version="1">
+  <ind:object object_ref="object_{{{ rule_id }}}" />
+  </ind:textfilecontent54_test>
+
+  <ind:textfilecontent54_object id="object_{{{ rule_id }}}" version="1">
+    <ind:filepath operation="pattern match">^/etc/sudoers(\.d/.*)?$</ind:filepath>
+    <!-- The regex idea: <user list> <host list> = (<the whole command without negation>,)* <command with negation> <whatever>
+         where a command is <runas spec>?<anything except , or !>+,
+           - ',' is a command delimiter, while
+           - '!' outside of a runas spec is a command negation we are after,
+         The capturing group holds the offending command.
+    -->
+    <ind:pattern operation="pattern match">^(?:\s*[^#][^=]+)=(?:\s*(\([^\)]+\))?\s*(?!\s*\()[^,!\n]+,)*\s*(?:\([^\)]+\))?\s*(?!\s*\()([^,\n]*!\S+).*</ind:pattern>
+    <ind:instance datatype="int">1</ind:instance>
+  </ind:textfilecontent54_object>
+</def-group>
--- a/linux_os/guide/system/software/sudo/sudoers_no_command_negation/oval/shared.xml
+++ b/linux_os/guide/system/software/sudo/sudoers_no_command_negation/oval/shared.xml
@@ -7,7 +7,7 @@
   </definition>
 
   <ind:textfilecontent54_test check="all" check_existence="none_exist"
-  comment="Make sure that no user spec in sudoers has a runas spec that includes root or ALL"
+  comment="Make sure that no command in user spec contains negation"
 	  id="test_{{{ rule_id }}}" version="1">
   <ind:object object_ref="object_{{{ rule_id }}}" />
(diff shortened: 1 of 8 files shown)


linux_os/guide/system/software/sudo/sudoers_no_command_negation/oval/shared.xml
```diff
@@ -0,0 +1,26 @@
+<def-group>
+     <definition class="compliance" id="{{{ rule_id }}}" version="1">
+     {{{ oval_metadata("Check that sudoers doesn't contain command negations") }}}
+     <criteria operator="AND">
+	     <criterion comment="Make sure that no command in user spec contains negation" test_ref="test_{{{ rule_id }}}" />
+     </criteria>
+  </definition>
+
+  <ind:textfilecontent54_test check="all" check_existence="none_exist"
+  comment="Make sure that no user spec in sudoers has a runas spec that includes root or ALL"
```
yuumasato, line 10: This comment needs to be updated.

linux_os/guide/system/software/sudo/sudoers_no_command_negation/oval/shared.xml
```diff
@@ -0,0 +1,26 @@
...
+
+  <ind:textfilecontent54_object id="object_{{{ rule_id }}}" version="1">
+    <ind:filepath operation="pattern match">^/etc/sudoers(\.d/.*)?$</ind:filepath>
+    <!-- The regex idea: <user list> <host list> = (<the whole command without negation>,)* <command with negation> <whatever>
+         where a command is <runas spec>?<anything except , or !>+,
+           - ',' is a command delimiter, while
+           - '!' outside of a runas spec is a command negation we are after,
+         The capturing group holds the offending command.
+    -->
+    <ind:pattern operation="pattern match">^(?:\s*[^#][^=]+)=(?:\s*(\([^\)]+\))?\s*(?!\s*\()[^,!\n]+,)*\s*(?:\([^\)]+\))?\s*(?!\s*\()([^,\n]*!\S+).*</ind:pattern>
```
yuumasato, line 23:
```suggestion
    <ind:pattern operation="pattern match">^(?:\s*[^#=]+)=(?:\s*(\([^\)]+\))?\s*(?!\s*\()[^,!\n]+,)*\s*(?:\([^\)]+\))?\s*(?!\s*\()([^,\n]*!\S+).*</ind:pattern>
```
In the test set there are no commented sudoers rules, testing on "regex101" the regex was matching commented infringing rules as a finding.
With the suggestion above the regex doesn't match infringing commented rules anymore.

yuumasato, line 23:
```suggestion
    <ind:pattern operation="pattern match">^(?:\s*[^#][^=]+)=(?:\s*(\(?:[^\)]+\))?\s*(?!\s*\()[^,!\n]+,)*\s*(?:\([^\)]+\))?\s*(?!\s*\()([^,\n]*!\S+).*</ind:pattern>
```
This avoids capturing runas specs with a `!`.

yuumasato (changes requested):
The regex looks good, I have just a few suggestions for it.

linux_os/guide/system/software/sudo/sudoers_no_command_negation/oval/shared.xml
matejak: I was hoping that the captured spec would somehow pop up in OVAL results, but that was never the case, so I will do it like that.

yuumasato: Hmm, I know that it emerges in the OpenSCAP DEVEL logs, in the collected item as `subexpression`.
It should show up in  the OVAL results too.

yuumasato: It should avoid capturing the runas spec, like `(bob !alice)`.
And actually, the suggestion is above.
Here is correct one:
```suggestion
    <ind:pattern operation="pattern match">^(?:\s*[^#][^=]+)=(?:\s*(?:\([^\)]+\))?\s*(?!\s*\()[^,!\n]+,)*\s*(?:\([^\)]+\))?\s*(?!\s*\()([^,\n]*!\S+).*</ind:pattern>
```

linux_os/guide/system/software/sudo/sudoers_no_command_negation/oval/shared.xml
```diff
@@ -0,0 +1,26 @@
...
+
+  <ind:textfilecontent54_object id="object_{{{ rule_id }}}" version="1">
+    <ind:filepath operation="pattern match">^/etc/sudoers(\.d/.*)?$</ind:filepath>
+    <!-- The regex idea: <user list> <host list> = (<the whole command without negation>,)* <command with negation> <whatever>
+         where a command is <runas spec>?<anything except , or !>+,
+           - ',' is a command delimiter, while
+           - '!' outside of a runas spec is a command negation we are after,
+         The last non-capturing group holds the offending command.
+    -->
+    <ind:pattern operation="pattern match">^(?:\s*[^#=]+)=(?:\s*(\([^\)]+\))?\s*(?!\s*\()[^,!\n]+,)*\s*(?:\([^\)]+\))?\s*(?!\s*\()(?:[^,\n]*!\S+).*</ind:pattern>
```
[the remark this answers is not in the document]
matejak: I see, I have edited the wrong group. Thanks for updating the suggestion!

PR #6089, opened 2020-09-15
Introduce centralised policy definitions

matejak:
This PR presents a fully-functional PoC of a build system extension that allows to introduce product-agnostic elements to profile definitions.
Right now, profiles are tightly coupled with products. If there are two products that are very similar, respective profiles are almost or entirely identical. This creates following problems:

- Profile development is product-dependent - if a profile is defined for more than one products, then profile experts will have to choose a "primary" product to develop the profile. This means that maintainers of other products and profiles will have to be aware of the development effort and that a particular profile is the "primary" one, as they will have to watch it and react to its development.
- When a profile issue is discovered in a profile outside of the concentrated development effort, an there is therefore not a "primary" profile that everybody watches, there is no easy way how to propose the fix to other products. A proper way that ensures that the information is not overlooked is to open pull requests to equivalent profiles so maintainers can merge them. In case of ANSSI, that profile is duplicated 8 times in the project, so one could have to open up to 7 PRs.

The centralised policy definition is able to overcome this - products automatically "take" selections from the policy, so it can be the target of interest when working with profile experts, as well as target of bugfixes, when a bug is discovered by a product user. Individual profiles can add product-specific rules outside of the centralised policy if they wish, and profiles can be protected against introduction of unwanted rules using the [profile stability tests](https://complianceascode.github.io/template/2020/03/06/profile_guard.html).

Additional benefit of centralised policies is that it brings together the structure of our content to the structure of policies how they are defined in respective documents - see this example from the ANSSI policy:

image

Regarding the implementation aspect of the PR:

The `controls/anssi.yml` file contains ANSSI BP28 requirements stored in YAML format. The file is both machine-readable and human-readable.  Each entry in the file represents an ANSSI requirement and contains the hardening level and a list of XCCDF rules mapped to this requirement. It concentrates the most important information into a single file. This form enables easier discussion and PRs on the rules mapping because it follows the logical structure of the original document.

The profile definitions leverage the fact that the `controls/anssi.yml` contains the rules and the profile levels they belong to. The profiles selection list in all 4 ANSSI profile files is replaced by a single statement that selects rules for the profile automatically.

[...]

PR #6499, opened 2020-12-22
Introduce centralised policy definitions

matejak:
This is a follow-up to #6089, as the source branch was updated. There are indications that the predecessor PR was misunderstood by reviewers at that time, and this is the most straightforward way of introducing it for reveiw again.

This PR presents a fully-functional PoC of a build system extension that allows to introduce product-agnostic elements to profile definitions.
Right now, profiles are tightly coupled with products. If there are two products that are very similar, respective profiles are almost or entirely identical. This creates following problems:

- Profile development is product-dependent - if a profile is defined for more than one products, then profile experts will have to choose a "primary" product to develop the profile. This means that maintainers of other products and profiles will have to be aware of the development effort and that a particular profile is the "primary" one, as they will have to watch it and react to its development.
- When a profile issue is discovered in a profile outside of the concentrated development effort, an there is therefore not a "primary" profile that everybody watches, there is no easy way how to propose the fix to other products. A proper way that ensures that the information is not overlooked is to open pull requests to equivalent profiles so maintainers can merge them. In case of ANSSI, that profile is duplicated 8 times in the project, so one could have to open up to 7 PRs.

The centralised policy definition is able to overcome this - products automatically "take" selections from the policy, so it can be the target of interest when working with profile experts, as well as target of bugfixes, when a bug is discovered by a product user. Individual profiles can add product-specific rules outside of the centralised policy if they wish, and profiles can be protected against introduction of unwanted rules using the [profile stability tests](https://complianceascode.github.io/template/2020/03/06/profile_guard.html).

Additional benefit of centralised policies is that it brings together the structure of our content to the structure of policies how they are defined in respective documents - see this example from the ANSSI policy:

image

Regarding the implementation aspect of the PR:

The `controls/anssi.yml` file contains ANSSI BP28 requirements stored in YAML format. The file is both machine-readable and human-readable.  Each entry in the file represents an ANSSI requirement and contains the hardening level and a list of XCCDF rules mapped to this requirement. It concentrates the most important information into a single file. This form enables easier discussion and PRs on the rules mapping because it follows the logical structure of the original document.

The profile definitions leverage the fact that the `controls/anssi.yml` contains the rules and the profile levels they belong to. The profiles selection list in all 4 ANSSI profile files is replaced by a single statement that selects rules for the profile automatically.

The core part of the PR is in `ssg/controls.py` Python module. It contains simple classes Control, Policy and ControlsManager. These classes implement parsing the controls file and contains methods that resolve them to a list of rules. ControlsManager is then used by the build system. There is also a simple unit test in `tests/unit/ssg-module/test_controls.py` which tests these classes. There are a few example files used by the unit test in `tests/unit/ssg-module/data/`.

The `ssg/build_yaml.py` has been refactored . Then, there is a new class `ProfileWithInlinePolicies` which inhertis from the ResolvableProfile class. It can resolve the controls in the profile selections list, but it can still process the profiles that consist of individual rules. The class replaces its parent it during profile compilation `ProfileWithInlinePolicies`.

The build system change is subtle. The order of steps is still the same. Only the `build-scripts/compile_profiles.py`  script which compiles the profiles is changed to use the controls during profile resolution. The compiled profiles still look the same, they contain a resolved list of rules, which means next build steps aren’t changed.

To avoid any confusion regarding possible unwanted implications PRs of this scale could introduce wrt content authors or developers, I emphasize that this PR
- doesn't change anything regarding evaluation (scanning), it
- doesn't change anything for profile authors that don't like this new format - they can still use the old one, and it even
- doesn't change anything for people that would like to analyze profile composition using the traditional "profile = linear list of rules" approach, because one can analyze the compiled profile file. In fact, one should set up [compiled profile stability test](https://complianceascode.github.io/template/2020/03/06/profile_guard.html) and have this partially automated.

Its main use case is to assist content development of ANSSI profiles, where the ability to record rationale of rule selections has a high added value.


carlosmmatos (approved):
As per our discussions - this is fine to implement as it doesn't impact previous mechanisms of profile generation. Will merge based on the outcome of CI

PR #6502, opened 2021-01-04
configure_crypto_policy test scenario - ensure that both files have same timestamp

mildas:
#### Description:
Improve the test scenario because in the past it was occasionally causing failures when the `/etc/crypto-policies/config` timestamp got rounded to second greater than the `/etc/crypto-policies/state/current` timestamp.

### DIFF
--- a/linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/tests/config_and_current_same_time.pass.sh
+++ b/linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/tests/config_and_current_same_time.pass.sh
@@ -2,9 +2,11 @@
 # platform = Red Hat Enterprise Linux 8
 
 # IMPORTANT: This is a false negative scenario.
-# File /etc/crypto-policies/config can be newer than /etc/crypto-policies/state/current
-# but the difference can be in milliseconds and after rounding to seconds timestamps of
-# those files can be same, thus incompliant.
+# File /etc/crypto-policies/config can be newer than /etc/crypto-policies/state/current,
+# which should always be a finding.
+# But the difference can be in a couple of milliseconds, and timestamps of
+# those files rounded to seconds can be same, thus making the OVAL test pass, although the system is incompliant.
+# The scenario equalizes those timestamps, so the corresponding test always passes.
 # With a precision of seconds in OVAL, there is not really much we can do to detect this.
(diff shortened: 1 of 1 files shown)


matejak:
I am a bit lost, the comment says that the scenario is a "false negative" one, so I assume that although it tends to pass, it should fail, but we can't fix the check.
Does your fix to the test scenario assure that there will be always the false negative, or does it modify the scenario to a point where it passes legitimately?

mildas:
@matejak Yes, the test scenario assures that this will be always false negative. The `/etc/crypto-policies/config` timestamp must be less than `/etc/crypto-policies/state/current` timestamp. However, smallest time unit in OVAL is one second and it was common that the rule check was failing even after remediation because the timestamps difference was so small that both timestamps were rounded to the time since epoch.

linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/tests/config_and_current_same_time.pass.sh
```diff
@@ -2,8 +2,9 @@
 # platform = Red Hat Enterprise Linux 8
 
 # IMPORTANT: This is a false negative scenario.
-# File /etc/crypto-policies/config is newer than /etc/crypto-policies/state/current, thus incompliant,
-# but the OVAL evaluation restuls in pass.
+# File /etc/crypto-policies/config can be newer than /etc/crypto-policies/state/current
+# but the difference can be in milliseconds and after rounding to seconds timestamps of
+# those files can be same, thus incompliant.
```
matejak, line 7:
```suggestion
# File /etc/crypto-policies/config can be newer than /etc/crypto-policies/state/current,
# which should always be a finding.
# But the difference can be in a couple of milliseconds, and timestamps of
# those files rounded to seconds can be same, thus making the OVAL test pass, although the system is incompliant.
# The scenario equalizes those timestamps, so the corresponding test always passes.
```

matejak:
I see, I have suggested a reword to the comment. The test scenario is not that useful now, but at least it corresponds to its name.
I think that overall, we are fine, because it is impossible to have a less than a second timestamp difference between those files by accident or by omission.

mildas:
Sounds good. I've commited suggested changes.

matejak:
Thanks for looking into the problem!

PR #6506, opened 2021-01-05
Dconf fixes

ggbecker:
#### Description:

- ac49ea6 fixes the problem with spaces in remediation because the OVAL as the way it is doesn't support spaces between `=`
- Also add `dconf update` to ansible remediation similarly to what `bash` remediation does.

#### Additional information

These gdm configurations should really become a template. There is a draft PR #6118 to implement the template, I should get back to it soon.


matejak:
We solve this in Bash with less copy-pasting using a macro, but a template would clearly be the best solution.

Good catch with the `no_extra_spaces`!

PR #6507, opened 2021-01-05
add rule checking if world writable directories are owned by root

vojtapolasek:
#### Description:

- add rule, check, remediations, tests

#### Rationale:

anssi effort


redhatrises:
LGTM

ggbecker:
Why not to select in the https://github.com/ComplianceAsCode/content/blob/master/controls/anssi.yml already.

Also I was taking a looking at the requirement and I had the impression that these permissions should apply to the directories that have the sticky bit set. Is that the case here with this new rule?

vojtapolasek:
I plan to add all rules pertaining to R40 into the policy file in a separate PR. I don't think that this rule is related to sticky bit situation, I created this rule based on 6.16 from the policy.

ggbecker:
> I plan to add all rules pertaining to R40 into the policy file in a separate PR. I don't think that this rule is related to sticky bit situation, I created this rule based on 6.16 from the policy.

You're right, I re-read the requirement and it's clear that directories that are world-writable should be owned by root, despite if sticky bit is set or not.

ggbecker (approved):

PR #6511, opened 2021-01-06
Fix jinja expansion on installed_OS_is_vendor_supported

ggbecker:
#### Description:

- Fix Jinja Expansion.

#### Rationale:

- Condition was always returning true and `SUSE` was getting into every product than `ol7` and `ol8` in this case.


jan-cerny:
Looks great! Thanks for noticing and fixing this problem!

PR #6523, opened 2021-01-11
Require separate filesystem for /var/tmp

kresss:
#### Description:

- Require a separate filesystem for the /var/tmp mount point. 
- 
#### Rationale:

- There are already rules in the ospp profiles for RHEL8 that require the nodev, nosuid, and noexec mount options on /var/tmp. /var/tmp should have its own filesystem instead of falling under the filesystem for /var.


yuumasato:
@openscap-ci ok to test

tests/data/profile_stability/rhel8/stig.profile
```diff
@@ -185,6 +185,7 @@ selections:
 - partition_for_var
 - partition_for_var_log
 - partition_for_var_log_audit
+- partition_for_var_tmp
```
yuumasato, line 188: RHEL-8 STIG profile also includes rules for mount options for `/var/tmp/`, so this change should be good for now.
CC: @ggbecker and @carlosmmatos for awareness.

ggbecker: Despite RHEL8 STIG having the mount options for `/var/tmp/`, the V1R1 doesn't mention this as a separate partition. Which I believe is a bug in the STIG. I guess since they have based the STIG on OSPP, that might explain why they miss the separate partition there. We have to bring this topic to the table. Whatever gets into this STIG profile that is extending OSPP will get overridden by the new standalone profile. So I don't have a strong opinion here.

ggbecker:
> Shouldn't there be at least a `nist` reference added to the rule if it is included in the OSPP profile?

I'm not sure about that, but if you take a look at the mount options rules for this partition, they also don't have any `nist` reference
```
    - mount_option_var_tmp_nodev
    - mount_option_var_tmp_noexec
    - mount_option_var_tmp_nosuid
```

carlosmmatos:
@kresss please resolve the conflicts and push the new changes.

PR #6525, opened 2021-01-12
Added a rule for having commands with arguments in sudoers - ANSSI R63

matejak:
The rule assures that commands in user specs are always followed by at least one argument.

The regular expression is already monstrous, so although the rule is somehow remediable in theory (one could add "" as an argument to commands that don't have one), there is no remediation, as having one that would work reliably would be very difficult.

The rule has some limitations - it doesn't handle commas in arguments well - `command one\,two` would be the correct way of specifying that only the argument `one,two` is allowed to be specified, and the current regex treats that escaped comma as a commands delimiter.


linux_os/guide/system/software/sudo/sudoers_no_command_args/tests/simple.fail.sh
```diff
@@ -0,0 +1,5 @@
+# platform = multi_platform_all
+# packages = sudo
+# remediation = none
+
+echo 'jen !fred		ALL, SERVERS = /bin/sh' > /etc/sudoers
```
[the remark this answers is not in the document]
yuumasato: And `visudo` complains about `jen !fred ALL, servers =`. The User_List is comma separated.
So it should be `jen,!fred ALL,Servers =`.
And `jen, !fred ALL, Servers =` was accepted too.

yuumasato (changes requested):
The check seems to working as expected, great work on the regex.
I've uncovered a minor issue on the User_List in the test scenarios that don't seem require any change, but would be good to double check.

> The regular expression is already monstrous, so although the rule is somehow remediable in theory (one could add "" as an argument to commands that don't have one), there is no remediation, as having one that would work reliably would be very difficult.

How about adding the note above as a warning? Similar to done in https://github.com/ComplianceAsCode/content/blob/5573ab62e16808bdb1f879422d2cca91c54e65a1/linux_os/guide/system/software/sudo/sudoers_no_command_negation/rule.yml#L41-L43


> The rule has some limitations - it doesn't handle commas in arguments well - `command one\,two` would be the correct way of specifying that only the argument `one,two` is allowed to be specified, and the current regex treats that escaped comma as a commands delimiter.

What happens when there is a command argument with comma? From what I tested it becomes a finding, i.e.: a false positive, which less of a problem than a false negative.
How about adding the limitation of comma handling to the rule as warning? So that people are less surprised when something valid comes up as a finding?

matejak:
Thanks, I think that I have addressed your concerns in a separate commit:

- Added warning about remediation and limitations to the rule, added a false positive test.
- Improved the check not to be confused by trailing whitespace, added a test.
- Fixed tests, so they correspond to real sudoers syntax.

linux_os/guide/system/software/sudo/sudoers_no_command_args/oval/shared.xml
```diff
@@ -0,0 +1,25 @@
...
+  </ind:textfilecontent54_test>
+
+  <ind:textfilecontent54_object id="object_{{{ rule_id }}}" version="1">
+    <ind:filepath operation="pattern match">^/etc/sudoers(\.d/.*)?$</ind:filepath>
+    <!-- The regex idea: <user list> <host list> = (<the whole command with at least an arg>,)* <command with no arg> <end of the line or next command spec we don't care about>
+         where a command is <runas spec>?<anything except ,>+,
+           - ',' is a command delimiter, while
+         The last capturing group holds the offending command without args.
+    -->
+    <ind:pattern operation="pattern match">^(?:\s*[^#=]+)=(?:\s*(?:\([^\)]+\))?\s*(?!\s*\()[^,\s]+(?:[ \t]+[^,\s]+)+,)*(\s*(?:\([^\)]+\))?\s*(?!\s*\()[^,\s]+(?:,|[ \t]$))</ind:pattern>
```
yuumasato, line 22:
```suggestion
    <ind:pattern operation="pattern match">^(?:\s*[^#=]+)=(?:\s*(?:\([^\)]+\))?\s*(?!\s*\()[^,\s]+(?:[ \t]+[^,\s]+)+[ \t]*,)*(\s*(?:\([^\)]+\))?\s*(?!\s*\()[^,\s]+[ \t]*(?:,|$))</ind:pattern>
```

There are two changes in the suggestion above:
- the first should allow matching any number of spaces after arguments of correct command calls
  sample test scenario: `nobody ALL=/bin/ls , (!bob,alice) /bin/dog, /bin/cat arg`
- the second allows matching the last command call without arguments nor spaces
  See comment in `simple.fail.sh`.

linux_os/guide/system/software/sudo/sudoers_no_command_args/rule.yml
```diff
@@ -0,0 +1,46 @@
...
+  cce@rhel8: CCE-83632-0
+
+references:
+    anssi: BP28(R63)
+
+ocil_clause: '/etc/sudoers file contains user specifications that allow execution of commands with any arguments'
+
+ocil: |-
+    To determine if arguments that commands can be executed with are restricted, run the following command:
+    <pre>$ sudo grep -PR '^(?:\s*[^#=]+)=(?:\s*(?:\([^\)]+\))?\s*(?!\s*\()[^,\s]+(?:[ \t]+[^,\s]+)+,)*(\s*(?:\([^\)]+\))?\s*(?!\s*\()[^,\s]+(?:,|$))' /etc/sudoers /etc/sudoers.d/</pre>
```
yuumasato, line 35:
```suggestion
    <pre>$ sudo grep -PR '^(?:\s*[^#=]+)=(?:\s*(?:\([^\)]+\))?\s*(?!\s*\()[^,\s]+(?:[ \t]+[^,\s]+)+[ \t]*,)*(\s*(?:\([^\)]+\))?\s*(?!\s*\()[^,\s]+[ \t]*(?:,|$))' /etc/sudoers /etc/sudoers.d/</pre>
```

linux_os/guide/system/software/sudo/sudoers_no_command_args/tests/simple.fail.sh
```diff
@@ -0,0 +1,5 @@
+# platform = multi_platform_all
+# packages = sudo
+# remediation = none
+
+echo 'jen,!fred		ALL,SERVERS = /bin/sh ' > /etc/sudoers
```
yuumasato, line 5: The original test scenario `jen,!fred		ALL,SERVERS = /bin/sh` started to fail.

controls/anssi.yml
```diff
@@ -845,7 +845,8 @@ controls:
   - id: R63
     level: intermediary
     title: Explicit arguments in sudo specifications
-    # rules: TBD
+    rules:
+    - sudoers_no_command_args
```
yuumasato, line 849: How about changing rule id to `sudoers_explicit_command_args`, per rule title?

yuumasato (changes requested):
The fix introduced a new issue, see `simple.fail.sh`, I made a suggestion for that.
And also added suggestion to handle spaces after args in compliant commands.

linux_os/guide/system/software/sudo/sudoers_no_command_args/rule.yml
matejak: I agree with this suggestion, thanks!

controls/anssi.yml
matejak: I agree, that would be a more descriptive ID.

yuumasato (approved):
Thank you for the rule, great work! :)

PR #6528, opened 2021-01-12
remove accounts_passwords_pam_faillock_enforce_local from rhel8 stig

vojtapolasek:
#### Description:

- remove the rule from profile and from data for stable profile test

#### Rationale:

- The rhel8 STIG does not include such requirement


mildas:
After looking into rhel8 stig, it turns out that `local_users_only` can be set if we want to track failed user authentication attempts only for local users and ignore centralized users. But it's only a note and no such requirement is there. Thanks and LGTM

PR #6530, opened 2021-01-13
fix selinux_policytype oval regex

vojtapolasek:
#### Description:

- relax the oval to allow space characters after the policy name
- add test for this case

#### Rationale:

The rule was too strict.


jan-cerny:
Thank you very much, the test scenario is mazing and it passes for me.

PR #6531, opened 2021-01-13
sshd_set_keepalive PCI DSS requirement reference

mildas:
#### Description:
Add `sshd_set_keepalive` same PCI DSS requirement reference as [sshd_set_idle_timeout](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/ssh/ssh_server/sshd_set_idle_timeout/rule.yml#L44) has.
8.1.8 requirement:
> If  a session has been idle for more than 15 minutes, require the user to re-authenticate to re-activate the terminal or session

Because the `sshd_set_keepalive` rule already was in the PCI DSS profile, the missing reference caused that the rule was not properly categorized, see

Before
wrong-pcidss

After
correct-pcidss

Side effect - both rules start to pass during profile remediation because this corrects their ordering - first remediate `sshd_set_keepalive`, then `sshd_set_idle_timeout`.

#### Rationale:
As [sshd_set_idle_timeout description](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/ssh/ssh_server/sshd_set_idle_timeout/rule.yml#L68) mentions, the rule does not have desired effect without configuring `ClientAliveCountMax` (configured by `sshd_set_keepalive`). Thus, if `sshd_set_idle_timeout` belongs to the requirement, `sshd_set_keepalive` also has to.


matejak:
Nice fix for the RHEL7 PCI-DSS-centric benchmark. Thanks!

PR #6532, opened 2021-01-13
Render policy to HTML

matejak:
This PR adds a capability to render a policy definition into HTML in a context of some product.

In order to accomplish this, the build system needed to be extended a bit:

- Rule records definition_location, which points to the actual rule file in the source tree - this is used to be able to deduce a Github URL from a compiled rule file.
- Security Policy now records `title` and `source`.
- Security Policy now records a list of controls under `controls`.
  Formerly, that attribute recorded mapping of control ID to controls, and has been moved to `controls_by_id`.
- Security Control constructor initializes `title`, `description`, and `automated`.

### DIFF
--- a/utils/controls-template.html
+++ b/utils/controls-template.html
@@ -1,7 +1,9 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
 <head>
-<style type="text/css">
+  <meta charset="UTF-8">
+  <title>Definition of {{{ policy.title }}} for {{{ full_name }}}</title>
+<style>
 </style>
 </head>
 <body>
@@ -19,7 +21,6 @@ based on <a href="{{{ policy.source }}}">{{{ policy.source }}}</a>
   <p>Description: {{{ control.description | safe }}}</p>
   <p>Level: {{{ control.level }}}</p>
   <p>Automated: {{{ control.automated }}}</p>
-  <p>
   {{% if control.rules -%}}
   Selections:
   <ul>
@@ -32,9 +33,10 @@ based on <a href="{{{ policy.source }}}">{{{ policy.source }}}</a>
     {{%- endfor %}}
   </ul>
   {{% else -%}}
+  <p>
   <strong>No rules selected</strong>
-  {{%- endif %}}
   </p>
+  {{%- endif %}}
 </div>  
 {{% endfor -%}}
 </div>  
--- a/utils/controls-template.html
+++ b/utils/controls-template.html
@@ -1,0 +1,39 @@
+<!DOCTYPE html>
+<html>
+<head>
+<style type="text/css">
+</style>
+</head>
+<body>
+
+<h1>Definition of {{{ policy.title }}} for {{{ full_name }}}</h1>
+
+{{% if policy.source %}}
+based on <a href="{{{ policy.source }}}">{{{ policy.source }}}</a>
+{{% endif %}}
+
+<div class="all_controls">
+{{% for control in policy.controls -%}}
+<div class="one_control">
+  <h2>{{{ control.id }}}: {{{ control.title }}}</h2>
+  <p>Description: {{{ control.description | safe }}}</p>
+  <p>Level: {{{ control.level }}}</p>
+  <p>Automated: {{{ control.automated }}}</p>
+  <p>
+  {{% if control.rules %}}
+  Selections:
+  <ul>
+    {{% for rulid in control.rules -%}}
+    <li><a href="https://github.com/ComplianceAsCode/content/tree/master/{{{ rules[rulid].relative_definition_location }}}">{{{ rulid }}}</a>: {{{ rules[rulid].title }}}</li>
+    {{% endfor -%}}
+  </ul>
+  {{% else %}}
+  <strong>No rules selected</strong>
+  {{% endif %}}
+  </p>
+</div>  
+{{% endfor -%}}
+</div>  
+
+</body>
+</html>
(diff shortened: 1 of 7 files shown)


utils/render-policy.py
```diff
@@ -0,0 +1,86 @@
...
+    def get_result(self):
+        subst_dict = dict(rules=self.rules_dict, product=self.product, policy=self.policy)
+        subst_dict.update(self.env_yaml)
+        html_jinja_template = os.path.join(os.path.dirname(__file__), "controls-template.html")
+        return ssg.jinja.process_file(html_jinja_template, subst_dict)
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description="Generate reference-rule mapping.")
+    parser.add_argument("policy", help="The policy YAML")
```
jan-cerny, line 72: should controls be used instead?

utils/render-policy.py
```diff
@@ -0,0 +1,86 @@
...
+    parser.add_argument("product", help="What product to consider")
```
jan-cerny, line 73: change it to `--product`

utils/render-policy.py
jan-cerny: change it to --controls instead

utils/render-policy.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def get_result(self):
+        subst_dict = dict(rules=self.rules_dict, product=self.product, policy=self.policy)
+        subst_dict.update(self.env_yaml)
+        html_jinja_template = os.path.join(os.path.dirname(__file__), "controls-template.html")
+        return ssg.jinja.process_file(html_jinja_template, subst_dict)
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description="Generate reference-rule mapping.")
```
jan-cerny, line 71: A better description would be better.

utils/render-policy.py
```diff
@@ -0,0 +1,86 @@
...
+        for r_file in rule_files:
+            rule = ssg.build_yaml.Rule.from_yaml(r_file, self.env_yaml)
+            rule.relative_definition_location = (
+                pathlib.PurePath(rule.definition_location)
+                .relative_to(self.project_directory))
+            rules_dict[rule.id_] = rule
+        return rules_dict
+
+    def get_env_yaml(self, product, build_dir):
+        product_yaml = self.project_directory / product / "product.yml"
```
jan-cerny, line 50: this is wonderful!

jan-cerny (changes requested):
Documentation is needed.

I haven't done a full review because I haven't figured out how to use it.
I tried something but I only got an error.

```
[jcerny@thinkpad scap-security-guide{pr/6532}]$ utils/render-policy.py --build-dir build/ --output policy.html controls/anssi.yml rhel8
  File "utils/render-policy.py", line 81, in <module>
    result = HtmlOutput(args.product, args.build_dir, args.policy).get_result()
  File "utils/render-policy.py", line 27, in __init__
    self.rules_dict = self.get_rules_with_metadata(product, build_dir)
  File "utils/render-policy.py", line 44, in get_rules_with_metadata
    pathlib.PurePath(rule.definition_location)
  File "/usr/lib64/python3.8/pathlib.py", line 907, in relative_to
    raise ValueError("{!r} does not start with {!r}"
ValueError: '/home/jcerny/work/git/scap-security-guide/linux_os/guide/system/permissions/partitions/mount_option_home_nodev/rule.yml' does not start with ''
```

Can you help me?

utils/render-policy.py
matejak: I don't know, the associated class in the source code is called "Policy", and the file allows definition of policy metadata.

matejak: I don't know, the associated class in the source code is called "Policy", and the file allows definition of policy metadata.

utils/render-policy.py
matejak: As the product is a required argument, I would rather keep it positional.

matejak:
We have figured out the error, and I have significantly updated the description that was really lacking.

utils/render-policy.py
```diff
@@ -0,0 +1,92 @@
...
+        subst_dict.update(self.env_yaml)
+        html_jinja_template = os.path.join(os.path.dirname(__file__), "controls-template.html")
+        return ssg.jinja.process_file(html_jinja_template, subst_dict)
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description="Render a policy file typically located in 'controls' directory "
+        "of the project to HTML in a context of a product. "
+        "The product must be built")
```
jan-cerny, line 74:
```suggestion
        "The product must be built.")
```

utils/controls-template.html
```diff
@@ -0,0 +1,43 @@
...
+{{% for control in policy.controls -%}}
+<div class="one_control">
+  <h2>{{{ control.id }}}: {{{ control.title }}}</h2>
+  <p>Description: {{{ control.description | safe }}}</p>
+  <p>Level: {{{ control.level }}}</p>
+  <p>Automated: {{{ control.automated }}}</p>
+  <p>
+  {{% if control.rules -%}}
+  Selections:
+  <ul>
```
jan-cerny, line 25: Paragraphs in HTML can't contain lists.

utils/controls-template.html
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE html>
+<html>
+<head>
+<style type="text/css">
+</style>
+</head>
```
jan-cerny, line 6: The HTML validator https://validator.w3.org/ shows multiple errors and warnings.

jan-cerny (changes requested):
It works great, I have only minor problems.

I have additional question: What about XCCDF Values selections? Did you consider them? What are pros and cons of having them shown in the generated document?

matejak:
Thanks for the validator tip.

matejak:
Regarding values, it would be nice to have them there, but I consider that out of scope of this PR that introduces the rendering as a PoC.

jan-cerny:
@matejak Thanks. It's a nice feature!

PR #6534, opened 2021-01-13
add rule sshd_x11_use_localhost

vojtapolasek:
#### Description:

- add templated rule
-add it to rhel7 stig profile

#### Rationale:

stig effort


shared/references/cce-redhat-avail.txt
```diff
@@ -1,8 +1,5 @@
-CCE-83392-1
```
ggbecker, line 1: These CCEs from line 1 and 2 seem to be from a different piece of work. They are not included in the new rule.

vojtapolasek: You are right, I fixed it. Thank you.

linux_os/guide/services/ssh/ssh_server/sshd_x11_use_localhost/rule.yml
```diff
@@ -0,0 +1,43 @@
...
+
+ocil_clause: "the display proxy is listening on wildcard address"
+
+ocil: |-
+    {{{ ocil_sshd_option(default="yes", option="X11UseLocalhost", value="yes") }}}
+
+template:
+    name: sshd_lineinfile
+    vars:
+        missing_parameter_pass: 'false'
```
ggbecker, line 40:
```suggestion
        missing_parameter_pass: 'true'
```

According to sshd manual and code, "yes" is the default setting.

https://github.com/openssh/openssh-portable/blob/master/servconf.c#L351
https://linux.die.net/man/5/sshd_config

ggbecker: I know that STIG in this case says if the line is missing or commented then it's a finding. But we decided to ignore this in past for all SSHD configurations, so customers don't have to reconfigure all their sshd config files. If the default is what the configuration wants then we make the OVAL check pass.

ggbecker (approved):

PR #6536, opened 2021-01-14
Improved regex and tests

matejak:
Negated commands start with an exclamation mark. As `!` may be present in the command's allowed argument, the regexp has to reflect that.

The exclamation mark in command argument doesn't lead to false positives any more - without the change, the modified test scenario would end up as a positive finding.

Complements #6498


yuumasato (approved):
LGTM, thank you.

PR #6537, opened 2021-01-15
RHEL-07-040710 now configures X11Forwarding to disable

ggbecker:
#### Description:

- RHEL-07-040710 now configures X11Forwarding to disable.

#### Rationale:

- Change from the upcoming version of RHEL7 STIG
- Severity has been changed to `medium`. Previously when the setting was to enable it was `high`


carlosmmatos:
LGTM

PR #6087, opened 2020-09-15
Merge dev shm rules into one STIG item.

ggbecker:
- This is a discussion starter pull request, do not merge as it is.

#### Description:

- Rules `mount_option_dev_shm_nodev` and `mount_option_dev_shm_nosuid` were merged together with `mount_option_dev_shm_noexec`. Do we want to combine them into one single rule or should we keep as it is? but then which STIG id should we use for other rules, do keep it like it is right now?

#### Rationale:

- Reference: https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-81013?version=V2R7&compareto=v2r8

PR #6538, opened 2021-01-15
OL7 DISA STIG v2r1 update

freddieRv:
#### Description:

- Updated OL7 STIG profile to conform with OL7 DISA STIG v2r1. Updated CCIs and SRGs where necessary. Generated new stig overlay and added DISA STIG OL7 v2r1 reference manual. 

#### Rationale:

- STIG profile for OL7 had not been updated since it was added.


linux_os/guide/system/permissions/partitions/mount_option_dev_shm_nodev/rule.yml
```diff
@@ -24,7 +24,7 @@ identifiers:
 references:
     cis@rhel8: 1.1.5
     cis@ubuntu1804: 1.1.14
-    stigid@ol7: OL07-00-021022
+    stigid@ol7: OL07-00-021024
```
ggbecker, line 27: Duplicated STIGIDs pose a bit of a problem. when results are translated into STIG viewer format, for example, only one of the rules will have the results incorporated. Also, `stig_overlay.xml` will have one of these three rules. I know that DISA has deliberately merged these rules into one single requirement, but that's the opposite we ultimately want to achieve. We want to have a single configuration item per rule. I have tried to merge these three rules together in the past but with no success, see the discussion here: https://github.com/ComplianceAsCode/content/pull/6087

I'm not complaining about this change itself, but more about explaining the context behind the problems that people might have with STIG results and mapping.

ggbecker:
I have a small remark regarding STIG ID, otherwise, it looks good.

linux_os/guide/system/permissions/partitions/mount_option_dev_shm_nodev/rule.yml
freddieRv: Got it. I'll follow the same approach and leave the old ids.

ggbecker:
Changes look good to me. That's a good improvement for OL users. With the new STIG benchmark v2r1, the content now can generate compatible STIG viewer results to be used with the latest version since DISA has come up with new identifiers for the requirements.

ggbecker (approved):

PR #6540, opened 2021-01-18
Select rules for ANSSI R37

yuumasato:
#### Description:

- These rules are better fit for R37 than R38.
   R37 is about binaries designed to be used with setuid or setgid bits.
   R38 is about reducing number of binaries with setuid root.

#### Rationale:

- Better alignment of R37 and R38

### DIFF
--- a/controls/anssi.yml
+++ b/controls/anssi.yml
@@ -590,8 +590,17 @@ controls:
 
   - id: R37
     level: minimal
-    title: Executables with setuid and/or setgid bits
-    # rules: TBD
+    title: Executables with setuid and setgid bits
+    notes: >-
+      Only programs specifically designed to be used with setuid or setgid bits can have these privilege bits set.
+      This requirement considers apropriate for setuid and setgid bits the binaries that are installed from
+      recognized and authorized repositories (covered in R15).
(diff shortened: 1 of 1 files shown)


controls/anssi.yml
```diff
@@ -590,19 +590,22 @@ controls:
...
+    - file_permissions_unauthorized_suid
+    - file_permissions_unauthorized_sgid
 
   - id: R38
     level: enhanced
     title: Executable setuid root
-    description: >-
-      Setuid executables should be as small as possible. When it is expected
-      that only the administrators of the machine execute them, the setuid bit
-      must be removed and prefer them commands like su or sudo, which can be monitored
```
jan-cerny, line 602: Why is the description removed?

jan-cerny (commented):
looks convincing

controls/anssi.yml
yuumasato: My bad, I'll add it back.

yuumasato: The description for R38 is not removed anymore.

matejak:
Merging, the CI failure has nothing to do with the PR.

PR #6541, opened 2021-01-18
add rhel7 stig specific rule for ssh ciphers

vojtapolasek:
#### Description:

- create new rule only for rhel7 with remediations and tests
- replace sshd_use_approved_ciphers in rhel7 stig profile with this new rule

#### Rationale:

part of rhel7 stig update: https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-204578?version=v3r1


linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers_ordered_stig/rule.yml
```diff
@@ -0,0 +1,63 @@
...
+prodtype: rhel7
+
+title: 'Use Only FIPS 140-2 Validated Ciphers'
+
+description: |-
+    Limit the ciphers to those algorithms which are FIPS-approved.
+    The following line in <tt>/etc/ssh/sshd_config</tt>
+    demonstrates use of FIPS-approved ciphers:
+    <pre>Ciphers aes256-ctr,aes192-ctr,aes128-ctr</pre>
+    This rule ensures that there are configured ciphers mentioned above (or dheir subset), keeping the given order of algorithms.
```
ggbecker, line 12:
```suggestion
    This rule ensures that there are configured ciphers mentioned
    above (or their subset), keeping the given order of algorithms.
```

Typo in `their` and add a new line

ggbecker:
Other than the typo it looks good to me.

vojtapolasek:
Thank you, accepted.

vojtapolasek:
Please DO NOT merge yet, I found a problem with the oval regex.

vojtapolasek:
Hopefuly fixed, you can review it. Thank you.

ggbecker:
I was thinking about turning the regex into a macro. This same logic is applied for the HMAC rule, but will also be applied for rules in RHEL8 STIG.

My proposal would be a jinja function that accepts the parameter, separator, and which items should be evaluated, then the macro would expand to the regex described in this PR and then can be reused by these other rules.

What do you think?

matejak:
I am in favor of this - hardly anybody can say that a macro is difficult and not approachable, while the regular expression is clear. Although the regex may be really good, purpose of some groups may not be obvious. This could be easily described in form of comments in the macro definition. Not to mention mitigation of the copy-paste problem - as the regex is complex enough, propagating any of its refinements to its other instances is difficult to achieve.

redhatrises:
What's the point of this change? Seems superfluous, and the existing rule should be updated.

vojtapolasek:
> What's the point of this change? Seems superfluous, and the existing rule should be updated.

This rule is stricter than the original rule. The original rule is also used in other policies than stig, e.g. ospp, hipaa... and they do not have such strict requirement in place.

vojtapolasek:
I am also in favor of the macro, but I believe that that work is out of scope of this PR. But we should definitely turn it into macro in the future.> What's the point of this change? Seems superfluous, and the existing rule should be updated.

This rule is stricter than the original rule. The original rule is also used in other policies than stig, e.g. ospp, hipaa... and they do not have such strict requirement in place.

ggbecker:
> > What's the point of this change? Seems superfluous, and the existing rule should be updated.
> 
> This rule is stricter than the original rule. The original rule is also used in other policies than stig, e.g. ospp, hipaa... and they do not have such strict requirement in place.

The upcoming revision of RHEL7 STIG will require that the Ciphers allowed (HMAC is also in the list) must be in a specific order, from the strongest to the weakest in this case. Previously the order was not considered at all, so any combination of supported Ciphers would match and they were described from the weakest to the strongest.

And as Vojtech mentioned, the rule is selected by other profiles that use a different set of Ciphers hence the need for a new rule.

matejak:
> What's the point of this change? Seems superfluous, and the existing rule should be updated.

Probably the strongest reason why to have a new rule instead of just making the existing one ordered is that the ordering introduces complexity that forced us to drop the rule parametrization by an XCCDF value. The old, unordered rule can be parametrized, and introducing ordering while removing parametrization would break existing tailoring.

ggbecker (approved):
Everything seems to be solved here. The need for a new rule has proven correct as there is no other way to use the old one without disrupting the use case.

PR #6542, opened 2021-01-18
Supress Ansible lint error 503

jan-cerny:
It says that Tasks that run when changed should likely be handlers.
However, we don't use handlers, and developer guide says that handlers
aren't supported. I assume handlers would cause problems for SCAP
scanners. Unless we start to support handlers this error isn't fixable
for us therefore we can suppress it globally.

Addressing problems in scap-security-guide-lint-check Jenkins job:
30/48 Test #260: ansible-playbook-ansible-lint-check-rhel8 .........***Failed  630.77 sec
all/sudo_add_umask.yml:30: [E503] Tasks that run when changed should likely be handlers
anssi_bp28_enhanced/sudo_add_umask.yml:30: [E503] Tasks that run when changed should likely be handlers
anssi_bp28_high/sudo_add_umask.yml:30: [E503] Tasks that run when changed should likely be handlers
anssi_bp28_intermediary/sudo_add_umask.yml:30: [E503] Tasks that run when changed should likely be handlers

### DIFF
--- a/tests/ansible-lint_config.yml
+++ b/tests/ansible-lint_config.yml
@@ -3,3 +3,4 @@ skip_list:
   - '301' # Commands should not change things if nothing needs doing
   - '303' # Using command rather than module
   - '403' # Package installs should not use latest
+  - '503' # Tasks that run when changed should likely be handlers


matejak:
One argument against this PR could be that content with such issues could cause problems down the line, as other consumers of our Ansible content would have to address this, and it could be quite serious for them.
On the other hand, I see that this is another of (already many) cases when we don't strictly follow the Ansible way of doing things, and I don't see how such another addition could become a noticeably bigger problem than the already existing waivers.
Therefore, I am slightly in favor of merging this.

yuumasato (approved):

PR #6546, opened 2021-01-19
add rhel7 stig specific rule for sshd approved macs

vojtapolasek:
#### Description:

- add rule, remediations, check, tests
- modify rhel7 stig profile
- remove stigid from original sshd_use_approved_macs

#### Rationale:

stig effort, similar to #6541


linux_os/guide/services/ssh/ssh_server/sshd_use_approved_macs_ordered_stig/oval/shared.xml
```diff
@@ -0,0 +1,38 @@
...
+
+  <ind:textfilecontent54_test check="all" check_existence="all_exist"
+  comment="tests the value of MACs setting in the /etc/ssh/sshd_config file"
+  id="test_sshd_use_approved_macs_ordered_stig" version="1">
+    <ind:object object_ref="obj_sshd_use_approved_macs_ordered_stig" />
+  </ind:textfilecontent54_test>
+
+  <ind:textfilecontent54_object id="obj_sshd_use_approved_macs_ordered_stig" version="1">
+    <ind:filepath>/etc/ssh/sshd_config</ind:filepath>
+    <ind:pattern operation="pattern match">^[\s]*(?i)MACs(?-i)[\s]+(?=[\w,-@]+)(hmac-sha2-512(?=[\w,-@]+|$),?)?(hmac-sha2-256(?=[\w,-@]+|$),?)?[\s]*(?:#.*)?$</ind:pattern>
```
matejak, line 34: Couldn't the regex be simplified?

- Why is there `[\w,-@]` in `hmac-sha2-256(?=[\w,-@]+|$),?` and not just `\w` ? Ditto for the lookahead between MACs and hmac-sha2-512.
- Why isn't there just `(hmac-sha2-256)?` at the end?

linux_os/guide/services/ssh/ssh_server/sshd_use_approved_macs_ordered_stig/bash/shared.sh
```diff
@@ -0,0 +1,7 @@
+# platform = multi_platform_wrlinux,Red Hat Enterprise Linux 7,Oracle Linux 7
+
+if grep -q -P '^[[:space:]]*MACs[[:space:]]+' /etc/ssh/sshd_config; then
```
matejak, line 3: FYI, you don't have to use `[[:space:]]` with `grep -P`, `\s` will do the same. And you could even copy-paste parts of the OVAL regex.

linux_os/guide/services/ssh/ssh_server/sshd_use_approved_macs_ordered_stig/oval/shared.xml
vojtapolasek: correct, thank you. I will apply this also to the other PR.

ggbecker (approved):
LGTM. The need for a new rule has the same foundation as the rule introduced by #6541. There is no other way of keeping this new requirement to the old rule without breaking the use case there.

PR #6548, opened 2021-01-19
Add metadata to ANSSI R35

yuumasato:
#### Description:

- Add metadata to R35 - umask value.
  Current implementation cannot differentiate between system and standard user `umask`, they are both set to the same value.


matejak:
LGTM, thanks!

PR #6553, opened 2021-01-21
Add variable selector and notes for ANSSI R29

yuumasato:
#### Description:

- Update time selection for 
- Add notes about checking remote user session inactivity with OpenSSH.


ggbecker (approved):
LGTM. Hopefully RHEL8 will keep using openssh8.0 where the idle timeout mechanism is still present.

PR #6555, opened 2021-01-22
Update testing kickstart file partitions

mildas:
#### Description:
Update ks file partitions to be also aligned with ANSSI profiles.

#### Rationale:
This ks is used for preparation of testing machines where profiles are tested. ANSSI profiles require to have `/srv`, `/opt`, and `/usr` as separate partitions thus they need to be specified as separate partitions or the rules will fail.

### DIFF
--- a/tests/kickstarts/test_suite.cfg
+++ b/tests/kickstarts/test_suite.cfg
@@ -79,18 +79,24 @@ zerombr
 clearpart --linux --initlabel
 
 # Create primary system partitions (required for installs)
-part /boot --fstype=xfs --size=512
+part /boot --fstype=xfs --size=512 --fsoptions="noauto,nosuid,noexec"
 part pv.01 --grow --size=1
 
 # Create a Logical Volume Management (LVM) group (optional)
 volgroup VolGroup --pesize=4096 pv.01
 
 # Create particular logical volumes (optional)
-logvol / --fstype=xfs --name=root --vgname=VolGroup --size=11264 --grow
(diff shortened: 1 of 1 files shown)


matejak:
When I run a scan, it shows that `/boot` should be on the same partition as well, and there are mount options that are missing, but I think that those are handled by remediations.
So what about `/boot`? That one is under R12.

mildas:
The `/boot` partition is problematic. ANSSI profile has [a rule](https://github.com/ComplianceAsCode/content/blob/master/controls/anssi.yml#L184) that requires the `noauto` option for `/boot`:

> **noauto**
> Can only be mounted explicitly (i.e., the -a option will not cause the filesystem to be mounted). 

and this option I have added to kickstart.

Thus, when you install a machine using this kickstart, restart the machine, and run scan, then scan won't find `/boot` partition because it is not mounted and all rules related to `/boot` options also fail. When you mount it manually, then it's there and has all correct options.

Do you think that `noauto` should not be added by kickstart?

yuumasato:
> Thus, when you install a machine using this kickstart, restart the machine, and run scan, then scan won't find `/boot` partition because it is not mounted and all rules related to `/boot` options also fail. When you mount it manually, then it's there and has all correct options.

Heh, that is a funny side effect of `noauto` that is not taken into account by the rule that checks for the `/boot` mount options.
I noticed now that the `noauto` mount option is optional for ANSSI, and should be deployed with care.

Issue #6544, opened 2021-01-18
ANSSI BP-028 (high) and ANSSI DAT-BP28 (high) - sudo issue after kickstart installation


mildas:
#### Description of problem:
It is not possible to run `sudo` on a machine that was installed using ANSSI BP-028 (high) kickstart.

#### SCAP Security Guide Version:
6013bc0

#### Operating System Version:
RHEL 7, RHEL 8

#### Steps to Reproduce:

1. Install a machine using ANSSI BP-028 (high) or ANSSI DAT-BP28 (high) kickstart
2. Connect to the machine and run e.g. `sudo ls`

#### Actual Results:
`sudo: /usr/bin/sudo must be owned by uid 0 and have the setuid bit set`

#### Expected Results:
Output from a command

#### Additional Information:
I suspect these two rules
- ~~sudoers_explicit_command_args~~
- ~~sudoers_no_root_target~~

because they are also failing when the profile is tested using `test_suite.py profile` - they don't come with remediation, failure is expected in this test.

mildas:
Fixed, thank you @yuumasato

PR #6556, opened 2021-01-22
Drop remediation for sudo_dedicated_group

yuumasato:
#### Description:

- Drop remediation for `sudo_dedicated_group`
  The rule cannot be simply remediated without admin preparation first.

#### Rationale:

- This rule combined with no root login can prevent all users from getting adminitrative privileges.
  This rule requires administrative configurations before being remediated, like adding admin accounts and adding them to the `sudogrp` or dedicated group of choice.


mildas (approved):
Tests are passing and the remediation is not available anymore.

mildas:
LGTM, fixes #6544

yuumasato:
Thanks for the review.

PR #6557, opened 2021-01-22
Add cap_audit_write to be able to run sshd in containers

matusmarhefka:
Podman drops `cap_audit_write` which causes that it is not possible
run sshd by default. Therefore, we need to add the capability.

For more details see:
https://github.com/containers/podman/issues/3651


matejak:
LGTM. Note that you need another `podman` than 2.1 to test this PR due to https://github.com/containers/podman/issues/8444

PR #6558, opened 2021-01-22
fix remediations of dir_perms_world_writable_root_owned

vojtapolasek:
#### Description:

- rewrite ansible remediation to use modules rather than shell command
- add fuse.sshfs to set of non-local file systems
- rename tests

#### Rationale:

The ansible remediation was failing the "find" command displayed some harmless errors during execution (pseudo file filesystems). This caused fatal failure of whole playbook.


linux_os/guide/system/permissions/files/dir_perms_world_writable_root_owned/ansible/shared.yml
```diff
@@ -4,5 +4,52 @@
...
+    hidden: yes
+    recurse: yes
+  register: found_dirs
+
+- name: "Change owner to root on directories which are world writable"
+  file:
+    path: '{{ item.path }}'
+    owner: root
+  loop:  '{{ found_dirs.files }}'
+  when: item.woth is defined and item.woth is true
```
yuumasato, line 55: The check for `is defined` should be done on `item` before accessing `woth`, and I think `is true` is not needed.
`when: item is defined and item.woth`

linux_os/guide/system/permissions/files/dir_perms_world_writable_root_owned/ansible/shared.yml
```diff
@@ -4,5 +4,52 @@
 # complexity = low
 # disruption = medium
 
-- name: "Find all world writable files not owned by root on local partitions and change their owner to root"
-  shell: find / -not -fstype afs -not -fstype ceph -not -fstype cifs -not -fstype smb3 -not -fstype smbfs -not -fstype sshfs -not -fstype ncpfs -not -fstype ncp -not -fstype nfs -not -fstype nfs4 -not -fstype gfs -not -fstype gfs2 -not -fstype glusterfs -not -fstype gpfs -not -fstype pvfs2 -not -fstype ocfs2 -not -fstype lustre -not -fstype davfs -type d -perm -0002 -uid +0 -exec chown root {} \;
+
+- name: "Configure excluded (non local) file systems"
+  set_fact:
+    excluded_fs:
```
yuumasato, line 10: How about naming it `excluded_fstypes`?

yuumasato (commented):
Looks good.
How about putting the tasks in a block?

linux_os/guide/system/permissions/files/dir_perms_world_writable_root_owned/ansible/shared.yml
yuumasato: As `item` is brought in by `loop`, `item is defined` may not be needed.

vojtapolasek: I will leave there item.woth is true, I think it is more understandable.

linux_os/guide/system/permissions/files/dir_perms_world_writable_root_owned/ansible/shared.yml
```diff
@@ -4,5 +4,52 @@
...
+    hidden: yes
+    recurse: yes
+  register: found_dirs
+
+- name: "Change owner to root on directories which are world writable"
+  file:
+    path: '{{ item.path }}'
+    owner: root
+  loop:  '{{ found_dirs.files }}'
+  when: item.woth is true
```
yuumasato, line 55: Ansible-2.9.17 doesn't like comparisons with `is true`
```
TASK [Change owner to root on directories which are world writable] **********************************************************************************************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"msg": "The conditional check 'item.woth is true' failed. The error was: template error while templating string: no test named 'true'. String: {% if item.woth is true %} True {% else %} False {% endif %}\n\nThe error appears to be in '/root/rhel7-playbook-anssi_nt28_enhanced.yml': line 4179, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n    - name: Change owner to root on directories which are world writable\n      ^ here\n"}
```

This works: `when: item.woth`

linux_os/guide/system/permissions/files/dir_perms_world_writable_root_owned/ansible/shared.yml
```diff
@@ -4,5 +4,52 @@
...
+
+- name: "Create empty list of excluded paths"
+  set_fact:
+    excluded_paths: "[]"
+
+- name: "Detect nonlocal file systems and add them to excluded paths"
+  set_fact: 
+    excluded_paths: "{{ excluded_paths | union([item.mount]) }}"
+  loop: "{{ ansible_mounts }}"
+  when: item.fstype is in excluded_fstypes
```
yuumasato, line 39: Ansible 2.9.17 also complained about this syntax.
Change it to:
`when: item.fstype in excluded_fstypes`

yuumasato (changes requested):
Found some syntax issues related to Ansible-2.9.17, :(

yuumasato (approved):

PR #6545, opened 2021-01-19
WIP: ocp4/CIS 1.3.1: Add rules for eviction thresholds

JAORMX:
This adds the necessary rules to check for the kubelet eviction
threshold values.

Signed-off-by: Juan Antonio Osorio Robles <jaosorior@redhat.com>

PR #6563, opened 2021-01-25
Add variable support to yamlfile_value template

ggbecker:
#### Description:

- Add variable support to `yamlfile_value` template.
  - Two new options were added to the template:
    - `xccdf_variable`: must have an XCCDF variable identifier
    - `embedded_data`: the data retrieved by the `yamlpath` returns a BLOB content, so the OVAL check is completely different in this case.
      - in case `embedded_data` is true, the value field must contain a regex with a capture group, this captured value will be used to compare with the variable.

#### Rationale:

- Some of the OCP checks have to use XCCDF variables and without this option in the template, manually crafted OVAL would have to be created.
- The feature design could be much better if we didn't have many use cases to cover. This feature can be simplified when yamlfile probe is able to parse embedded JSON/YAML data.

kudos to @JAORMX for testing this feature.

This supersedes #6545


applications/openshift/api-server/api_server_request_timeout/rule.yml
```diff
@@ -40,3 +40,17 @@ ocil: |-
...
+    ocp_data: "true"
+    entity_check: "at least one"
+    filepath: /api/v1/namespaces/openshift-kube-apiserver/configmaps/config
+    yamlpath: '.data["config.yaml"]'
+    xccdf_variable: var_api_min_request_timeout
+    embedded_data: "true"
+    values:
+    - value: '"apiServerArguments":{.*"min-request-timeout":\["(\d*)"\]'
+      operation: "pattern match"
+      type: "string"
```
JAORMX, line 56: Let's add an e2e test to ensure this is passing [1]. No remediation should be needed for this one with this default. 

All you need for the test is:

```
---
default_result: PASS
```

[1] https://complianceascode.readthedocs.io/en/latest/manual/developer/06_contributing_with_content.html#writing-e2e-tests-for-specific-rules

JAORMX (changes requested):

applications/openshift/api-server/api_server_request_timeout/rule.yml
ggbecker: I will add the test.

mrogers950:
@ggbecker very cool, thank you for this feature! I have suggested changes to the description of the request timeout check that I will open a follow up PR for.

ggbecker:
@openscap-ci test this please

JAORMX:
A recent update in OCP seems to have broken some rules:

```
     helpers.go:640: Result - Name: e2e-cis-node-master-file-groupowner-ovs-conf-db - Status: FAIL - Severity: medium
    helpers.go:643: E2E-FAILURE: The expected result for the file_groupowner_ovs_conf_db rule didn't match. Expected 'PASS', Got 'FAIL'
    helpers.go:640: Result - Name: e2e-cis-node-master-file-groupowner-ovs-conf-db-lock - Status: FAIL - Severity: medium
    helpers.go:643: E2E-FAILURE: The expected result for the file_groupowner_ovs_conf_db_lock rule didn't match. Expected 'PASS', Got 'FAIL' 
```
Need to investigate

applications/openshift/api-server/api_server_request_timeout/rule.yml
ggbecker: Done

JAORMX:
/retest

I'm unable to reproduce this in 4.6... maybe it's a CI-only failure. I guess I'll have to re-kick the deployment and log into that cluster.

JAORMX:
Identified the issue. Fixing it now.

JAORMX:
https://github.com/ComplianceAsCode/content/pull/6566 Fixes the issue.

JAORMX:
@ggbecker could you rebase? the fix merged.

ggbecker:
> @ggbecker could you rebase? the fix merged.

Done

JAORMX:
/retest

Timed out

shared/templates/yamlfile_value/oval.template
```diff
@@ -25,12 +25,39 @@
...
+  <ind:variable_object id="variable_object_{{{ rule_id }}}" version="1">
+    <ind:var_ref>local_variable_{{{ rule_id }}}</ind:var_ref>
+  </ind:variable_object>
+
+  <ind:variable_state id="variable_state_{{{ rule_id }}}" version="1">
+    <ind:value datatype="string" operation="equals" var_ref="{{{ XCCDF_VARIABLE }}}"/>
+  </ind:variable_state>
+
+  {{% for val in VALUES %}}
+  <local_variable id="local_variable_{{{ rule_id }}}" datatype="string" comment="comment1" version="1">
```
yuumasato, line 49: This may not work with multiple `values`.
It seems to me that there will be multiple `local_variable` with the same ID.

JAORMX: oh! this might need an enhancement then.

ggbecker: I think it doesn't make sense to keep this multiple value option here. We should consider only the first element and make sure in the template data processing that the condition is met. Partially, this condition is already met... when using variables and embedded the data processing checks if values is set, I think we would have to expand to check if only a single value is present

PR #6564, opened 2021-01-25
Reassign a new unique CCE identifier to approved macs STIG rule.

ggbecker:
#### Description:

- Reassign a new unique CCE identifier to the approved macs STIG rule.

#### Rationale:

- Fix issue with duplicated CCE.


ggbecker:
Self merging since this issue makes all jobs fail in the testing for unique CCEs. No need to wait.

ggbecker:
@vojtapolasek I've removed the bugfix label since we don't need to backport it to the stabilization branch.

PR #6565, opened 2021-01-25
CIS 1.2.34: update api_server_encryption_provider_cipher

mrogers950:
The previous manual recommendation of checking the EncryptionConfig directly, as well as the Status of the encryption process would be too error prone to make for a reliable check. For one, the automatic check cannot reference a secret, and verifying based on status output would likely return true if the encryption was disabled after being previously enabled.

Instead, do the same check as api_server_encryption_provider_config, since selecting `aescbc` at the top level config is both the method to enable and start encryption of etcd as well as configure the strongest cipher.


jhrozek (approved):
/lgtm

PR #6570, opened 2021-01-27
Remove noauto for boot partition from test kickstart and ANSSI profiles

vojtapolasek:
#### Description:

- remove "noauto" option from definition of /boot partition in test suite kickstart

#### Rationale:

This causes many problems during testing because /boot partition has to be manually mounted.


yuumasato (approved):
LGTM.
As we discussed, I pushed a change to comment `mount_option_boot_noauto` for the time being.

vojtapolasek:
Looks good, thanks for completing the change.

PR #6573, opened 2021-01-27
Extend list of rules of unselected rules for testing

mildas:
#### Description:
Add `sudo_add_noexec` to the list because it prevents from running oscap scan on machine.


yuumasato (approved):
Looks good.

Adding `noexec` to `/etc/sudoers` was preventing some testing `sudo commands` to spawn a shell, :smile:.
That is the intended behavior, but we can exclude it during testing.
The rule is templated, so its behavior can be tested through rule mode.

PR #6574, opened 2021-01-27
fix ansible remediation of dir_perms_world_writable_root_owned

vojtapolasek:
#### Description:

- fix wrong list declaration in remediation
- add test which uses nfs4 as a nonlocal file system


yuumasato (approved):
Nice catch.
Thanks for the test scenario, it exercises the fix.

PR #6575, opened 2021-01-28
remove "noauto" option from boot partition in anssi kickstarts

vojtapolasek:
#### Description:

- remove option from rhel7 and rhel8 kickstarts

#### Rationale:

- rule was removed from the profile and presence of this option causes errors when checking / remediating


yuumasato (approved):
The `noauto` mount option is optional, and the other rules checking mandatory mount options for `/boot` are not ready for it.

PR #6576, opened 2021-01-28
Bump RHEL7 STIG version to v3r2.

ggbecker:
Replace all related artifacts with the new version of them.

#### Description:

- Bump RHEL7 STIG version to v3r2.
  - Replace all related artifacts with the new version of them.


yuumasato (approved):
LGTM

PR #6579, opened 2021-01-29
Select RHEL8 STIG V1R1 existing content

ggbecker:
#### Description:

- Select content that matches with RHEL8 STIG and add STIG id references to rules.
- The profile no longer extends OSPP.

#### Rationale:

- RHEL8 STIG has been released


ggbecker:
I'm moving to draft again because I identified some issues with the STIG ids assignment. There are duplicates that are not intentional. Please hold back until I address this and switch to ready to review again.

@yuumasato ^^^

ggbecker:
> I'm moving to draft again because I identified some issues with the STIG ids assignment. There are duplicates that are not intentional. Please hold back until I address this and switch to ready to review again.
> 
> @yuumasato ^^^

The STIG ids should be fixed now. I've decided also to remove intentionally duplicated STIG ids. These ids were assigned for rules that cover a single requirement in STIG, but these requirements should be split. Duplicate ids do not add any value and can cause misinterpretation.

yuumasato (approved):
LGTM.

PR #6580, opened 2021-01-29
Remove rule for /boot noauto from R13

yuumasato:
#### Description:

- Disable the rule until mount options for `/boot` can be checked without the need for the partition to be mounted.

#### Rationale:

- This causes false positives in other rules for `/boot` mount options


vojtapolasek:
lgtm, thank you

mildas:
This rule still causes problems with checking other `/boot` options and that `/boot` is on separate partition because `/boot` is not mounted automatically. Until we have different way for checking `/boot` rules, this one should be disabled. Thank you, LGTM

mildas (approved):

PR #6586, opened 2021-02-02
Drop remediation for sysctl_kernel_modules_disabled

yuumasato:
#### Description:

- Remove Bash remediation for `sysctl_kernel_modules_disabled`.

#### Rationale:

- Remediating this during kickstart install time renders the machine unbootable.
- This makes it impossible to apply the fix during kickstart install. Unfortunately it also makes it impossible to fix it with Bash _force_.
  - There is no way to make a rule remediate with Bash during normal runs but not remediate during kickstart install. Thoughts on this?
- From what I tested, applying the fix after first boot is fine.

### DIFF
--- a/linux_os/guide/system/permissions/restrictions/sysctl_kernel_modules_disabled/rule.yml
+++ b/linux_os/guide/system/permissions/restrictions/sysctl_kernel_modules_disabled/rule.yml
@@ -26,6 +26,11 @@ references:
 
 platform: machine
 
+warnings:
+  - general:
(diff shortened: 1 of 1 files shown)


vojtapolasek:
Hi, thanks for this. I believe this deserves a warning in the rule.yml, so that people are not surprised why the rule is not remediated.

yuumasato:
@vojtapolasek Warning added.

vojtapolasek:
lgtm, thank you

PR #6588, opened 2021-02-02
ocp4/CIS: Address 3.2

JAORMX:
This addresses 3.2.1 and 3.2.2 with a single rule that checks that
desired audit profile is set in the cluster.


JAORMX:
/retest

error gathering logs

PR #6600, opened 2021-02-04
Extend /var partition to 3GB in rhel8 kickstarts

mildas:
#### Description:
In rhel8 kickstarts, allocate 1GB more for `/var` instead of `/`.

#### Rationale:
In rhel8 installation with GUI packages, seems like that 2GB are not enough for `/var` because lately kickstart installations started to fail with those errors:
```
Disk Requirements:
   At least 34MB more space needed on the /var filesystem. The installer will now terminate.
```

### DIFF
--- a/rhel8/kickstart/ssg-rhel8-anssi_bp28_enhanced-ks.cfg
+++ b/rhel8/kickstart/ssg-rhel8-anssi_bp28_enhanced-ks.cfg
@@ -110,7 +110,7 @@ part pv.01 --grow --size=1
 volgroup VolGroup --pesize=4096 pv.01
 
 # Create particular logical volumes (optional)
-logvol / --fstype=xfs --name=LogVol06 --vgname=VolGroup --size=4216 --grow
+logvol / --fstype=xfs --name=LogVol06 --vgname=VolGroup --size=3192 --grow
 # Ensure /usr Located On Separate Partition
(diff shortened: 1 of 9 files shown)


matusmarhefka:
Do we want to change this for all the rhel8 kickstarts or only for ANSSI where we got this issue?

yuumasato:
The issue with GUI can be independent of the profile applied. My only concern is if `/` has enough storage for install.

mildas:
I'm also concerned if 3GB for `/` are enough. It works now (I have tested the change on ANSSI enhanced profile) but in few weeks/months same issue with disk can appear for `/`. What do you think about taking disk space from `/usr` instead of `/`?

yuumasato:
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/performing_a_standard_rhel_installation/partitioning-reference_installing-rhel

The recommended size for `/` (not including `/home` and `swap`) is 10GB, minimal is 5GB.

3 GB for `/` seems enough, given that other paths like `/usr` and `/var` have their own partitions.
And we can increase it if is not enough any more.

yuumasato (approved):
@matusmarhefka Are you ok with not changing just the ANSSI Kickstarts?

matusmarhefka:
@yuumasato @mildas Yes I am fine with that. If that would be a problem our testing will catch it anyway.

PR #6613, opened 2021-02-12
remove kernel_module_vfat_disabled from CIS profiles

vojtapolasek:
#### Description:

- remove the rule from all cis profiles

#### Rationale:

- the rule is marked as manual / unscored and additionally it prevents UEFI systems from booting


carlosmmatos:
LGTM

PR #6619, opened 2021-02-16
adds remediation for 1.2.25 api_server_audit_log_maxsize

swarred:
#### Description:

-  _This PR is to add remediation for `api_server_audit_log_maxsize`_

#### Rationale:

-  _Currently, there is not remediation for [this rule
](https://github.com/ComplianceAsCode/content/tree/master/applications/openshift/api-server/api_server_audit_log_maxsize)_
-  _Fixes #6618_


shared/templates/dconf_ini_file/template.py
```diff
@@ -1,6 +1,7 @@
 
 import os
 
+
```
JAORMX, line 4: nit: Let's get rid of this extra space.

applications/openshift/api-server/api_server_audit_log_maxsize/kubernetes/shared.yml
```diff
@@ -0,0 +1,7 @@
+# platform = multi_platform_ocp
+apiVersion: config.openshift.io/v1
+kind: APIServer
+metadata:
+  name: cluster
+spec:
+  maximumFileSizeMegabytes: 100
```
JAORMX, line 7: nice! This looks really good! To ensure that this gets tested and we don't regress, do you mind adding an e2e test [1]?

The main idea is to check what was the default value and to check that the check should pass once it gets remediated.

So...

if the test was failing to begin with and then it gets remediated, the `e2e.yml` file would look as follows:

```
---
default_result: FAIL
result_after_remediation: PASS
```

There is no need to create an `e2e-remediation.sh` file since this already has an automated remediation, which is this file.

[1] https://complianceascode.readthedocs.io/en/latest/manual/developer/06_contributing_with_content.html#end-to-end-tests

JAORMX (changes requested):
This is looking great! Let's just add an e2e test to avoid regressions and this will be ready to ship

JAORMX:
/retest

The CI cluster seems to be having issues:

```
 The connection to the server api.ci-op-xsx5dh84-1d877.origin-ci-int-aws.dev.rhcloud.com:6443 was refused - did you specify the right host or port? 
```

This error is unrelated to the PR.

applications/openshift/api-server/api_server_audit_log_maxsize/kubernetes/shared.yml
mrogers950: Should `maximumFileSizeMegabytes` be variable-ized and default to 100? (Not sure if we can do that substitution with remediations).

JAORMX: We currently can't. We would need an enhancement in the OpenSCAP command to be able to do those. The problem is that variables only get replaced on the remediate command from `oscap` and since these are `kubernetes` remediations we would need to enhance how that works... then hook that up to the operator. Certainly something we should tackle at some point.

shared/templates/dconf_ini_file/template.py
swarred: @JAORMX When opening a PR previously, we recieved a linting error on this file: [pep8 E501 expected 2 blank lines](https://www.python.org/dev/peps/pep-0008/#blank-lines). 

Do you still want us to change this back?

applications/openshift/api-server/api_server_audit_log_maxsize/kubernetes/shared.yml
swarred: @JAORMX Will get this `e2e.yml` test in.

swarred:
@JAORMX I apologize for the extra commits, I have a new line in the `e2e.yml` added

JAORMX:
Ah! I didn't know there was a test already. Apparently this passes by default and it was already reflected in the e2e tests

shared/templates/dconf_ini_file/template.py
JAORMX: Nah, that's alright; let's keep it then

applications/openshift/api-server/api_server_audit_log_maxsize/kubernetes/shared.yml
JAORMX: @swarred Seems that there was already an e2e.yml and the test was passing to begin with. This addition actually makes the test fail. Let's remove the changes to e2e.yml and keep the test passing, as it seems that the default was just fine.

swarred:
@JAORMX `e2e.yml` changes have been reverted

```
---
default_result: PASS
```

JAORMX:
@openscap-ci test this please

JAORMX:
@openscap-ci test this please

PR #6620, opened 2021-02-17
ocp4: TLS: Add NIST control reference and add to moderate profile

JAORMX:
Add checks to cover SC-8(1) and add them to the moderate profile.


matejak:
@openscap-ci test this please

PR #6623, opened 2021-02-18
fix wrong description of sshd_limit_user_access

vojtapolasek:
#### Description:

- correct the description - replace DenyUsers with AllowUsers

#### Rationale:

- https://bugzilla.redhat.com/show_bug.cgi?id=1859551
- using DenyUsers did not make sense with the rest of description, AllowUsers should be used instead


ggbecker (approved):
LGTM

PR #6627, opened 2021-02-22
ocp4/unit: Get rid of MachineConfig specific tests

JAORMX:
We no longer require these, as we're able to detect issues like this in
the e2e tests now


jhrozek (approved):
/lgtm

PR #6628, opened 2021-02-22
Fix STIG id reference for sshd_x11_use_localhost.

ggbecker:
#### Description:

- Fix STIG id reference for sshd_x11_use_localhost.

#### Rationale:

- The format for STIG id references is `stigid`


yuumasato (approved):
Nice catch

PR #6638, opened 2021-02-24
Add partition_for_var_tmp SRG

mildas:
#### Description:
This requirement is not explicitly listed in DISA STIG requirements but its group should be same as other partitions rules.
See https://github.com/ComplianceAsCode/content/pull/6523 where the rule was added.

#### Rationale:
Rule must have SRG because it is in stig profile.


ggbecker:
As you said, this requirement is not listed in the STIG RHEL8 v1r1, although we believe it should. The request was sent to DISA, and hopefully, we will see this in V1R2.

PR #6639, opened 2021-02-24
OCP: Add OVAL for api_server_anonymous_auth

jhrozek:
#### Description:

This covers control 1.2.1 of CIS. We deviate from CIS in the sense that
CIS itself prescribes that anonymous access should be disabled. OCP
allows anonymous access, but all requests are instead authorized and any
requests that are not authenticated reach the authorization layer as
user=system:anonymous and group=system:unauthenticated.

This rule takes that into account by making sure that there exists at
least one clusterrolebinding that binds the system:unauthenticated group
to a role, which in turn ensures that requests bound to the
system:unauthenticated group cannot do more that what that role allows
them to.

#### Rationale:

- Increases the CIS benchmark coverage


JAORMX (approved):

JAORMX:
@openscap-ci test this please

PR #6646, opened 2021-02-25
kubernetes: Get rid of unit tests

JAORMX:
These weren't as useful as initially thought. So let's rely on the e2e
tests instead.


jhrozek (approved):
/lgtm

PR #6642, opened 2021-02-25
Allow to override build date with SOURCE_DATE_EPOCH

bmwiedemann:
#### Description:

- Allow to override build date with `SOURCE_DATE_EPOCH`
in order to make builds reproducible.

This PR was done while working on [reproducible builds for openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds).

#### Rationale:

- See https://reproducible-builds.org/ for why this is good and https://reproducible-builds.org/specs/source-date-epoch/ for the definition of this variable.
- Without this patch, `.xml` files varied for each build in

```
    <oval:timestamp>2021-02-22T21:05:22</oval:timestamp>
```

PR #6647, opened 2021-02-25
Sort file lists

bmwiedemann:
#### Description:

- Sort file lists

#### Rationale:

- so that xml files build in a reproducible way
in spite of indeterministic filesystem readdir order
and http://bugs.python.org/issue30461

See https://reproducible-builds.org/ for why this is good.

Together with #6642 and https://github.com/OpenSCAP/openscap/pull/1699 this allowed for bit-identical results.
Probably some of this change is not needed, but build was slow, so I did not try to reduce the patch.

This PR was done while working on [reproducible builds for openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds).


jan-cerny:
Thank you!

PR #6649, opened 2021-03-01
Introduce packit support

matejak:
Packit enables integration with downstream Fedora packaging, and potentially with packaging of derived systems.

The setup uses the Rawhide spec file to create the srpm, so the upstream project doesn't require any adjustments.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,0 +1,17 @@
+# See the documentation for more information:
+# https://packit.dev/docs/configuration/
+
+specfile_path: scap-security-guide.spec
+
+# add or remove files that should be synced
+synced_files:
+    - scap-security-guide.spec
+    - .packit.yaml
+
+# name in upstream package repository/registry (e.g. in PyPI)
+upstream_package_name: scap-security-guide
+# downstream (Fedora) RPM package name
+downstream_package_name: scap-security-guide
+
+actions:
+  post-upstream-clone: "wget https://src.fedoraproject.org/rpms/scap-security-guide/raw/rawhide/f/scap-security-guide.spec -O scap-security-guide.spec"


TomasTomecek (commented):
I can see the SRPM build failed due to infra problem - I'll investigate tomorrow what the problem was.

TomasTomecek:
okay, I can see the problem now, we failed to pull image from docker.io due to quota; sorry about that

TomasTomecek:
I can see the RPM failed to be built due to cmake - is there some special way to create archives or doing `git archive` is good enough?

jan-cerny:
@TomasTomecek Only the F32 build fails. For some reason it tries to use the Rawhide spec file on F32. Do you have any ideas why it didn't use the F32 spec file on F32?

jan-cerny:
Aha I can see what can be the problem, it's in the pull request description:

> The setup uses the Rawhide spec file to create the srpm, so the upstream project doesn't require any adjustments.

You can't use the Rawhide spec file to build RPM on F32 because we use CMake and the RPM macro `%cmake` behaves differently starting from F33. 

If you are curious you can read details about this F33 change: https://fedoraproject.org/wiki/Changes/CMake_to_do_out-of-source_builds

@matejak Please change your "setup" so that a correct spec according to the Fedora version is used.

.packit.yaml
```diff
@@ -0,0 +1,17 @@
...
+    - scap-security-guide.spec
+    - .packit.yaml
+
+# name in upstream package repository/registry (e.g. in PyPI)
+upstream_package_name: scap-security-guide
+# downstream (Fedora) RPM package name
+downstream_package_name: scap-security-guide
+
+actions:
+  post-upstream-clone: "wget https://src.fedoraproject.org/rpms/scap-security-guide/raw/rawhide/f/scap-security-guide.spec -O scap-security-guide.spec"
```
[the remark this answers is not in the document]
matejak: This won't be a problem here - required dependencies don't change often.

matejak:
> @matejak Please change your "setup" so that a correct spec according to the Fedora version is used.

That's not a bug, it's a feature - the point is to use the Rawhide spec file for everything. You are right that it doesn't work for Fedora 32, and it also doesn't work for RHEL8 due to the cmake build issue as well, but it will work for anything newer than Fedora 33 and derived.

jan-cerny:
> > @matejak Please change your "setup" so that a correct spec according to the Fedora version is used.
> 
> That's not a bug, it's a feature - the point is to use the Rawhide spec file for everything. You are right that it doesn't work for Fedora 32, and it also doesn't work for RHEL8 due to the cmake build issue as well, but it will work for anything newer than Fedora 33 and derived.
> 


I think that maybe you can add an if else construction that depend on the system version to the rawhide spec

matejak:
> I think that maybe you can add an if else construction that depend on the system version to the rawhide spec

I don't think that a product that will be EOLed in two months deserves to have an `if` statement in the rawhide spec file. The goal is to have the rawhide spec file as simple as possible as long as it can be used on RHEL, CentOS Stream, and Fedora at the same time. If there is a conflict between simplicity and product support, one has to pick sides, and F32 alone is not a convincing side.

TomasTomecek:
The easiest solution is to configure `jobs` and precisely pick chroots:
```
  - job: copr_build
    trigger: pull_request
    metadata:
      targets:
        - fedora-rawhide
        - fedora-34
        - fedora-33
  - job: tests
    trigger: pull_request
    metadata:
      targets:
        - fedora-rawhide
        - fedora-34
        - fedora-33
```
we also have aliases for development and stable versions of Fedora Linux: https://packit.dev/docs/configuration/#available-copr-build-targets

Let me know if this is acceptable for you.

.packit.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+
+# TODO: Remove the "jobs" key as soon as F34 is out
```
jan-cerny, line 20: My understanding was that this key disables F32 builds so I would say that it should be disabled after F32 is removed from the infrastructure. Am I correct or are you correct?

jan-cerny: Also, is F33 under fedora-development?

matejak: I think that both of us are correct - when F34 is released, F32 will disappear, and we can return to default jobs.

F33 is clearly not a development version, as you can see when you check out test results. I have tried to avoid mentioning explicit version numbers, so things work even if nobody removes the `jobs` key.

jan-cerny:
@matejak Any updates?

matejak:
I would say that it simply works, so we can merge it. If we wait until the F34 release, we can simplify the config even further.

jan-cerny:
@matejak Fedora 34 was released in April

matejak:
> @matejak Fedora 34 was released in April

Time flies, it's actually really F35 which is around the corner

matejak:
And it doesn't work because of patches in the Fedora spec file, so let's close the PR.

PR #6652, opened 2021-03-01
Adds classification-banner rule

### LINKED ISSUE #6651: No rule written for openshift to check classification banner
Description of problem:

There is not currently a check for a classification banner in OpenShift in accordance with AC-8 SYSTEM USE NOTIFICATION under NIST 800-53.

SCAP Security Guide Version:
N/A

Operating System Version:
OCP 4.6

Steps to Reproduce:
N/A

Actual Results:
Rule is NOT present at applications/openshift/classification-banner/rule.yml

Expected Results:
Rule IS present at applications/openshift/classification-banner/rule.yml

swarred:
#### Description:

This PR is to add a rule to check for a classification banner in OpenShift.

#### Rationale:

- Currently, there is not a rule to check for a classification banner in OpenShift
- Fixes #6651


swarred:
@JAORMX Do you we need to add in e2e tests for this rule? We do not intend on adding associated remediation to this rule, given that the Console Notification configuration would not be considered universal for all environments.

applications/openshift/classification-banner/rule.yml
```diff
@@ -0,0 +1,25 @@
+prodtype: ocp4
+
+title: Security Banner must be configured on OpenShift console
+
+description: test
+
+rationale: TBD
+
+identifiers: {}
+
```
JAORMX, line 10: given that we know what controls this is addressing, let's add the `references` key and state the nist control.

applications/openshift/classification-banner/rule.yml
```diff
@@ -0,0 +1,25 @@
...
+severity: unknown
```
JAORMX, line 11: let's add severity: medium

JAORMX (commented):
The rule makes sense! Let's make sure it works in CI. Here's the docs:

https://complianceascode.readthedocs.io/en/latest/manual/developer/06_contributing_with_content.html#end-to-end-tests

swarred:
@JAORMX I have made requested changes

applications/openshift/classification-banner/rule.yml
```diff
@@ -0,0 +1,30 @@
...
+
+template:
+  name: yamlfile_value
+  vars:
+    ocp_data: "true"
+    entity_check: "at least one"
+    filepath: /apis/console.openshift.io/v1/consolenotifications/classification-banner
+    yamlpath: ".spec.text"
+    values:
+    - value: "Unclassified"
```
JAORMX, line 30: this seems to only verify that the banner has the text "Unclassified" is this expected? Wouldn't we want to verify that folks have put at least something? you could use a regex here.

swarred: @JAORMX This is expected! The text itself is custom depending on what the user set as the value. I'm open to a better way if you think I am missing something. The value could be `text: Anything`. I chose `text: Unclassified` as this rule is expecting a CRD is created, via the `e2e-remediation.sh`, that creates the CRD with `text: Unclassified`

https://docs.openshift.com/container-platform/4.7/web_console/customizing-the-web-console.html#creating-custom-notification-banners_customizing-web-console

JAORMX: So, this will try to match "Unclassified" for the text and will fail on any other value. I think you instead want to use a regex to indicate that you want to match anything. So let's put:

```
    - value: ".*"
      operation: "pattern match"
      type: "string"
```

Here `.*` will be the thing that matches any set of characters.

JAORMX:
oho! all of a sudden this PR got a lot bigger

swarred:
@JAORMX Got the PR cleaned up, sorry about that!

JAORMX (changes requested):
This is looking great!

One last thing, in order to test this, let's add this to the moderate profile: https://github.com/ComplianceAsCode/content/blob/master/ocp4/profiles/moderate.profile

Right now it's not in any profile, and thus, not tested in CI yet.

swarred:
@JAORMX Added rule to moderate profile. I went ahead and changed folder structure to match what I am seeing accross the project. It appears I need a `group.yml` file though?

https://complianceascode.readthedocs.io/en/latest/manual/developer/03_creating_content.html?highlight=group.yml#benchmark-structure-layout

JAORMX:
@swarred yep, it seems you do need that if you're introducing a new group (which is what the new sub-folder represents)

swarred:
@JAORMX group.yml file was added. Thanks!

JAORMX:
/retest

OpenShift CI seems to be overloaded

JAORMX:
There seems to be something wrong with the banner:
```
     helpers.go:676: Result - Name: e2e-moderate-classification-banner - Status: FAIL - Severity: medium
    helpers.go:679: E2E-FAILURE: The expected result for the classification_banner rule didn't match. Expected 'PASS', Got 'FAIL' 
```

swarred:
@JAORMX Fixed!

JAORMX:
The rule is fine and the remediation works locally. It's currently failing due to an implementation detail in CI (the pod runs with a readonly filesystem for security) so it'll fail in its current implementation. Since most of this is fine. I'll merge this and submit a separate PR fixing CI. Thanks for your work here!

PR #6660, opened 2021-03-05
Add missing CIS references

yuumasato:
#### Description:

- Extend test for missing STIG IDs to check CIS, HIPAA and ANSSI references
  - I think `profile_tools.py` needs a rewrite to easily support new references, but I didn't want to go through it right now.
- Add CIS references to rules selected in RHEL CIS profiles.

#### Rationale:

- The references provide tracking and rationale for the rule selection in the Profile.

#### Note:
- Missing references for HIPAA and ANSSI will be added in following PRs.


ssg/build_profile.py
```diff
@@ -453,6 +483,33 @@ def show_profile_stats(self, profile, options):
...
+                      "CIS Refs: %d of %d have them [%d%% missing]"
+                      % (profile, rules_count - missing_cis_refs_count,
+                         rules_count,
+                         (100.0 * missing_cis_refs_count / rules_count)))
+                self.console_print(profile_stats['missing_cis_refs'],
+                                   console_width)
+
+            if options.missing_hipaa_refs and profile_stats['missing_hipaa_refs']:
+                print("*** rules of '%s' profile missing "
+                      "CIS Refs: %d of %d have them [%d%% missing]"
```
matusmarhefka, line 497: `HIPAA Refs:...`

ssg/build_profile.py
```diff
@@ -453,6 +483,33 @@ def show_profile_stats(self, profile, options):
...
+                      % (profile, rules_count - missing_hipaa_refs_count,
+                         rules_count,
+                         (100.0 * missing_hipaa_refs_count / rules_count)))
+                self.console_print(profile_stats['missing_hipaa_refs'],
+                                   console_width)
+
+            if options.missing_anssi_refs and profile_stats['missing_anssi_refs']:
+                print("*** rules of '%s' profile missing "
+                      "CIS Refs: %d of %d have them [%d%% missing]"
```
matusmarhefka, line 506: `ANSSI Refs:...`

matusmarhefka:
I have double checked and there is no rule in CIS profiles (both RHEL7 and RHEL8) which is missing reference anymore.

ssg/constants.py
```diff
@@ -61,6 +61,9 @@
 anaconda_system = "urn:redhat:anaconda:pre"
 cce_uri = "https://nvd.nist.gov/cce/index.cfm"
 stig_ns = "https://public.cyber.mil/stigs/srg-stig-tools/"
+cis_ns = "https://www.cisecurity.org/benchmark/red_hat_linux/"
```
JAORMX, line 64: So... would the CIS references for OCP show this as well? How could we select a different link for different products? We'd need an OpenShift-specific one, or https://www.cisecurity.org/benchmark/kubernetes/ in the meantime since the OCP one is still not on the CIS website.

ssg/build_profile.py
yuumasato: XD, Fixed

ssg/constants.py
yuumasato: Hmm, great observation, the References URIs are not managed centrally.

>  How could we select a different link for different products?

The products can define their own reference URIs in `<product>/transforms/constants.xslt`.
This is the [rhel8](https://github.com/ComplianceAsCode/content/blob/master/rhel8/transforms/constants.xslt) one, and the [ocp4](https://github.com/ComplianceAsCode/content/blob/master/ocp4/transforms/constants.xslt) is empty.

The `profile_tool.py` script should be improved to be able to get URI specific to the be product it is handling.
I'd leave this for future PRs.

yuumasato: "Un-emptying' the CIS URI for OCP4 content in #6663

PR #6662, opened 2021-03-05
Update ANSSI references

yuumasato:
#### Description:

- Add ANSSI references to rules selected in the `controls/anssi.yml`

#### Rationale:
- The references provide tracking and rationale for the rule selection in the Profile.


jan-cerny (approved):
I have checked it with controls/anssi.yml and it's OK.

I think in future we can automate it.

PR #6665, opened 2021-03-05
Add new rules to sle12/profiles/stig.profile

brett060102:
This adds all rules in last PR sle12 stig.profile

Fix banner oval check:
   Appropriately encode GDM banner
   The OVAL check doesn't allow the banner text to be in single quote
   so use the approprate macro to format it correctly.

Add missing bash remediation. 

#### Description:

- add sl12 rules to stig.profile, fix a found bug

#### Rationale:

-  Add newly supported sle12 rules to stig.profile
-  Add shared/bash_remediation_functions/ensure_pam_module_options.sh which was dropped from earlier changes.

- banner oval check: Appropriately encode GDM banner. The OVAL check doesn't allow the banner text to be in single quote  so use the appropriate macro to format it correctly.

### DIFF
--- a/linux_os/guide/system/accounts/accounts-banners/gui_login_banner/banner_etc_gdm_banner/ansible/shared.yml
+++ b/linux_os/guide/system/accounts/accounts-banners/gui_login_banner/banner_etc_gdm_banner/ansible/shared.yml
@@ -8,5 +8,5 @@
 - name: "{{{ rule_title }}}"
   lineinfile:
     dest: /etc/gdm/banner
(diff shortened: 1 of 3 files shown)


ggbecker (approved):

ggbecker:
@brett060102  I think now is a good time to update the stig_overlay.xml. Check the reference: https://complianceascode.readthedocs.io/en/latest/manual/developer/04_updating_reference_and_overlay.html?highlight=create-stig-overlay#stig-overlay-content

brett060102:
@ggbecker thank you @abergmann do you want to take this one or shall I?

PR #6668, opened 2021-03-08
V-204578 change order of SSH cipher suites for stig

amdonov:
Latest stig requires specific order for SSH ciphers.

#### Description:

Change the order of SSH cipher suites.

#### Rationale:

https://www.stigviewer.com/stig/red_hat_enterprise_linux_7/2020-12-08/finding/V-204578

Inspect the "Ciphers" configuration with the following command:

```
# grep -i ciphers /etc/ssh/sshd_config
Ciphers aes256-ctr,aes192-ctr,aes128-ctr

```
If any ciphers other than "aes256-ctr", "aes192-ctr", or "aes128-ctr" are listed, **the order differs from the example above**, the "Ciphers" keyword is missing, or the returned line is commented out, this is a finding.

### DIFF
--- a/linux_os/guide/services/ssh/sshd_approved_ciphers.var
+++ b/linux_os/guide/services/ssh/sshd_approved_ciphers.var
@@ -11,5 +11,5 @@ operator: equals
 interactive: false
 
 options:
-    stig: aes128-ctr,aes192-ctr,aes256-ctr
+    stig: aes256-ctr,aes192-ctr,aes128-ctr
     default: aes128-ctr,aes192-ctr,aes256-ctr,aes128-cbc,3des-cbc,aes192-cbc,aes256-cbc,rijndael-cbc@lysator.liu.se


ggbecker:
We have already fixed this issue by creating a new rule that considers ordering and a subset of ciphers. Please use it instead: [sshd_use_approved_ciphers_ordered_stig](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers_ordered_stig/rule.yml)

Only changing the order in the variable does not solve all the issues, which includes allowing a subset of ciphers in this case.

For more details please check the pull request: https://github.com/ComplianceAsCode/content/pull/6541

This new rule should be part of the next release v0.1.55.

Feel free to reopen this or start a discussion at https://github.com/ComplianceAsCode/content/discussions

PR #6672, opened 2021-03-09
fix rule accounts_password_pam_unix_remember

vojtapolasek:
#### Description:

- add tests
- make oval check for both system-auth and password-auth like the description states
- rewrite remediations to keep pam_unix.so if already used but prefer pam_pwhistory.so, as pam_unix.so is not being supported in later pam versions

#### Rationale:

- remediations and checks were not aligned with rule description

### DIFF
--- a/linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_password_pam_unix_remember/ansible/shared.yml
+++ b/linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_password_pam_unix_remember/ansible/shared.yml
@@ -5,14 +5,53 @@
 # disruption = medium
 {{{ ansible_instantiate_variables("var_password_pam_unix_remember") }}}
 
-- name: "Do not allow users to reuse recent passwords - system-auth (change)"
-  replace:
-    dest: /etc/pam.d/system-auth
-    regexp: '^(password\s+sufficient\s+pam_unix\.so\s.*remember\s*=\s*)(\S+)(.*)$'
-    replace: '\g<1>{{ var_password_pam_unix_remember }}\g<3>'
+# this macro is used to correct existing instances of unix or pwhistory module
 
-- name: "Do not allow users to reuse recent passwords - system-auth (add)"
+{{% macro prevent_password_reuse(file, module) -%}}
+
+- name: "detect if {{{ module }}} module is used"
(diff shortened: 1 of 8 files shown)


yuumasato (commented):
NAck.

The approach to keep adding `remember` to `pam_unix` if it is present is not a good one, `pam_unix` will most likely be always present.
Similar approach would be to check if `pam_unix` is already used for `remember` purposes, but that doesn't work either.
Changing the password after it expires doesn't work when `pam_unix .* remember` is configured in `rhel7` and `rhel8`.
Related BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1778929

yuumasato:
Snack for discussion: A rule should probably address `system-auth` and `password-auth` independently.

yuumasato:
How about taking [accounts_password_pam_pwhistory_remember](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_password_pam_pwhistory_remember/rule.yml) and making `system-auth` and `password-auth` version?

ggbecker:
> How about taking [accounts_password_pam_pwhistory_remember](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_password_pam_pwhistory_remember/rule.yml) and making `system-auth` and `password-auth` version?

To add a bit of context, this rule was introduced by Suse folks and they were following SLE12 STIG strictly. We are in the process of asking DISA to split these similar rules in RHEL8. Hopefully, when they accept it, these changes will also be reflected in STIG for other products.

yuumasato:
> > How about taking [accounts_password_pam_pwhistory_remember](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_password_pam_pwhistory_remember/rule.yml) and making `system-auth` and `password-auth` version?
> 
> To add a bit of context, this rule was introduced by Suse folks and they were following SLE12 STIG strictly. We are in the process of asking DISA to split these similar rules in RHEL8. Hopefully, when they accept it, these changes will also be reflected in STIG for other products.

I see now that rule actually configurew two things, `remember` and `use_authtok`.

yuumasato:
This has been superseded by #6672.

PR #6675, opened 2021-03-10
Make test suite tests more accessible

matejak:
The PR introduces fixes to the test suite as well as to the container testing wrapper:

Fixed machine-only handling in the test suite.
- Fixed query for Bash fixes.
- Support for empty remediations.

Refreshed the container test wrapper.
- Added support for `--dontclean`
- Simplified the end-result command-line composition.
- Enabled the `--dry-run` option that prints the full test suite command-line.

### DIFF
--- a/tests/test_rule_in_container.sh
+++ b/tests/test_rule_in_container.sh
@@ -2,7 +2,7 @@
 
 # Created by argbash-init v2.9.0
 # ARG_OPTIONAL_SINGLE([name],[n],[Name of the test image],[ssg_test_suite])
-# ARG_OPTIONAL_SINGLE([scenarios],[s],[Regex to reduce selection of tested scenarios],[.*])
+# ARG_OPTIONAL_SINGLE([scenarios],[s],[Regex to reduce selection of tested scenarios],[])
 # ARG_OPTIONAL_SINGLE([datastream],[d],[Path to the datastream to use in tests. Autodetected by default.])
 # ARG_OPTIONAL_SINGLE([remediate-using],[r],[What to remediate with],[oscap])
 # ARG_OPTIONAL_BOOLEAN([dontclean],[],[Dont remove HTML reports from the log directory.])
@@ -52,7 +52,7 @@ _positionals=()
 _arg_rule=
 # THE DEFAULTS INITIALIZATION - OPTIONALS
 _arg_name="ssg_test_suite"
-_arg_scenarios=".*"
+_arg_scenarios=""
 _arg_datastream=
 _arg_remediate_using="oscap"
 _arg_dontclean="off"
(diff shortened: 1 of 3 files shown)


tests/test_rule_in_container.sh
```diff
@@ -53,17 +55,21 @@ _arg_name="ssg_test_suite"
...
 {
 	printf '%s\n' "Test a rule using the container backend."
-	printf 'Usage: %s [-n|--name <arg>] [-s|--scenarios <arg>] [-d|--datastream <arg>] [-r|--remediate-using <REMEDIATION>] [-h|--help] <rule>\n' "$0"
+	printf 'Usage: %s [-n|--name <arg>] [-s|--scenarios <arg>] [-d|--datastream <arg>] [-r|--remediate-using <REMEDIATION>] [--(no-)dontclean] [--(no-)dry-run] [-h|--help] <rule>\n' "$0"
 	printf '\t%s\n' "<rule>: The short rule ID. Wildcards are supported."
 	printf '\t%s\n' "-n, --name: Name of the test image (default: 'ssg_test_suite')"
 	printf '\t%s\n' "-s, --scenarios: Regex to reduce selection of tested scenarios (default: '.*')"
 	printf '\t%s\n' "-d, --datastream: Path to the datastream to use in tests. Autodetected by default. (no default)"
 	printf '\t%s\n' "-r, --remediate-using: What to remediate with. Can be one of: 'oscap', 'bash' and 'ansible' (default: 'oscap')"
+	printf '\t%s\n' "--dontclean, --no-dontclean: Dont remove HTML reports from the log directory. (off by default)"
```
jan-cerny, line 71: :zany_face: 

Totally unsure what "off by default" means here. What is "no dontclean"?

tests/test_rule_in_container.sh
```diff
@@ -53,17 +55,21 @@ _arg_name="ssg_test_suite"
...
+	printf '\t%s\n' "--dry-run, --no-dry-run: Just print the test suite command-line. (off by default)"
```
jan-cerny, line 72: This is similar but not so terrible. However, I really don't understand the purpose of `--no-dry-run`. If I don't want to run a dry run I simply don't type `--dry-run`.

jan-cerny (changes requested):

tests/test_rule_in_container.sh
matejak: Off refers to the option, i.e. it means that `--dontclean` is not considered by default. I agree that this double negative (don't + no) is unfortunate.

tests/test_rule_in_container.sh
matejak: There is a theoretical possibility that you have command options stored in an array, but you want to override any prior `--dry-run` that may or may not be there, and you would do that by adding `--no-dry-run` to the array. But again, I agree that it is unfortunate.

jan-cerny:
@matejak I have a new problem: the command generated by the dry run option doesn't work for me.

```
[jcerny@localhost scap-security-guide{pr/6675}]$ tests/test_rule_in_container.sh --dry-run ensure_logrotate_activated
python3 /home/jcerny/work/git/scap-security-guide/tests/test_suite.py rule --remove-machine-only --scenario .* --remediate-using oscap --add-platform fedora --container ssg_test_suite -- ensure_logrotate_activated
[jcerny@localhost scap-security-guide{pr/6675}]$ python3 /home/jcerny/work/git/scap-security-guide/tests/test_suite.py rule --remove-machine-only --scenario .* --remediate-using oscap --add-platform fedora --container ssg_test_suite -- ensure_logrotate_activated
usage: test_suite.py [-h] {profile,rule,combined} ...
test_suite.py: error: unrecognized arguments: -- ensure_logrotate_activated
```

jan-cerny:
I think it's because the `.*` needs quotes

PR #6677, opened 2021-03-10
mount_option_* test scenarios

mildas:
#### Description:
Most of the test scenarios are copy-paste from other mount rules with small adjustments.

Previous `mount_option_dev_shm_noexec` test scenarios were removed because they did not test the rule properly - they just changed line in fstab/mtab and that isn't enough. All these test scenarios were passing because the default configuration of a machine is exactly the same as the test scenarios expected as result.

#### Rationale:
Increase test coverage of rules from CIS profile.

### DIFF
--- a/tests/shared/partition.sh
+++ b/tests/shared/partition.sh
@@ -53,5 +53,7 @@ clean_up_partition() {
     path="$1"
     escaped_path=${path//$'/'/$'\/'}
     sed -i "/${escaped_path}/d" /etc/fstab
-    umount ${path} || true  # no problem if not mounted
+    if mountpoint -q -- "${path}"; then
+        umount -l ${path}
+    fi
 }
--- a/linux_os/guide/system/permissions/partitions/mount_option_dev_shm_nodev/tests/fstab.fail.sh
+++ b/linux_os/guide/system/permissions/partitions/mount_option_dev_shm_nodev/tests/fstab.fail.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+. $SHARED/partition.sh
+
+clean_up_partition /dev/shm
+
+create_partition
+
+make_fstab_given_partition_line /dev/shm ext2 noexec
+
+mount_partition /dev/shm
--- a/linux_os/guide/system/permissions/partitions/mount_option_dev_shm_nodev/tests/runtime.pass.sh
+++ b/linux_os/guide/system/permissions/partitions/mount_option_dev_shm_nodev/tests/runtime.pass.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+. $SHARED/partition.sh
+
+clean_up_partition /dev/shm
+
+create_partition
+
+make_fstab_correct_partition_line /dev/shm
+
+mount_partition /dev/shm
--- a/linux_os/guide/system/permissions/partitions/mount_option_dev_shm_nodev/tests/separate.fail.sh
+++ b/linux_os/guide/system/permissions/partitions/mount_option_dev_shm_nodev/tests/separate.fail.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+. $SHARED/partition.sh
+
+clean_up_partition /dev/shm
+
+create_partition
+
+make_fstab_correct_partition_line /dev/shm
+
+# $SHARED/fstab is correct, but we are not mounted.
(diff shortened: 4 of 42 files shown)


JAORMX (approved):

matusmarhefka:
When testing on RHEL8 VM I see few errors/warnings:
```
...
INFO - xccdf_org.ssgproject.content_rule_mount_option_nosuid_removable_partitions
WARNING - Script no_partitions.pass.sh - profile xccdf_org.ssgproject.content_profile_C2S not found in datastream
...
INFO - xccdf_org.ssgproject.content_rule_mount_option_dev_shm_noexec
ERROR - Script runtime.pass.sh using profile (all) found issue:
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_mount_option_dev_shm_noexec'.
INFO - Script separate.fail.sh using profile (all) OK
INFO - Script fstab.fail.sh using profile (all) OK
ERROR - Rule evaluation resulted in error, instead of expected fixed during remediation stage 
ERROR - The remediation failed for rule 'xccdf_org.ssgproject.content_rule_mount_option_dev_shm_noexec'.
...
INFO - xccdf_org.ssgproject.content_rule_mount_option_dev_shm_nosuid
INFO - Script runtime.pass.sh using profile (all) OK
ERROR - Script separate.fail.sh using profile (all) found issue:
ERROR - Rule evaluation resulted in pass, instead of expected fail during initial stage 
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_mount_option_dev_shm_nosuid'.
ERROR - Script fstab.fail.sh using profile (all) found issue:
ERROR - Rule evaluation resulted in pass, instead of expected fail during initial stage 
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_mount_option_dev_shm_nosuid'.
...
INFO - xccdf_org.ssgproject.content_rule_mount_option_nodev_removable_partitions
WARNING - Script no_partitions.pass.sh - profile xccdf_org.ssgproject.content_profile_C2S not found in datastream
...
INFO - xccdf_org.ssgproject.content_rule_mount_option_dev_shm_nodev
INFO - Script runtime.pass.sh using profile (all) OK
ERROR - Script separate.fail.sh using profile (all) found issue:
ERROR - Rule evaluation resulted in pass, instead of expected fail during initial stage 
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_mount_option_dev_shm_nodev'.
ERROR - Script fstab.fail.sh using profile (all) found issue:
ERROR - Rule evaluation resulted in pass, instead of expected fail during initial stage 
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_mount_option_dev_shm_nodev'.
```

mildas:
@matusmarhefka I've added lazy umount option to `partition.sh` to force umount when device is busy.

tests/shared/partition.sh
```diff
@@ -53,5 +53,5 @@ clean_up_partition() {
     path="$1"
     escaped_path=${path//$'/'/$'\/'}
     sed -i "/${escaped_path}/d" /etc/fstab
-    umount ${path} || true  # no problem if not mounted
+    umount -l ${path} || true  # no problem if not mounted
```
matusmarhefka, line 56: This might mask some other issues, what about checking if path is mounted and unmount only in that case?

mildas: I don't think it masks issues and I also don't think that checking if path is mounted will help us.
In case of `/dev/shm`, it is mounted but `umount` fails:
```
# umount /dev/shm
umount: /dev/shm: target is busy.
```
However `# umount -l /dev/shm` passes because it detach a filesystem from the file hierarchy now and clean up references as soon as it is not busy anymore.

In case of `umount` without any option, it doesn't detach a filesystem, we can't remount it with new options, and thus we can't perform test scenarios properly.

matusmarhefka: I agree with the lazy umount. My point was, what if even lazy umount fails? This way you will hide the error, SSGTS will report that it expected rule to pass but got fail. If you would rather umount only in case the path is mounted then these kind of issues would be correctly detected by the SSGTS which would report that script failed to prepare environment.

mildas: Aha, I think I get it now. Updated with check if the path is mounted

matusmarhefka:
I can confirm that all issues are fixed now, thank you @mildas :)

PR #6679, opened 2021-03-11
ocp4/e2e: fix classification_banner remediation

JAORMX:
This fixes the remediation by not writing the object to /tmp which is
read-only in the container.


JAORMX:
@swarred here's the fix for the remediation.

swarred:
@JAORMX Awesome, great to know!

JAORMX:
@swarred the script was also missing execute permissions. So doing `chmod +x <filename>` did the trick.... the problem right now is OpenShift CI, which seems to be broken in general.

Issue #5213, opened 2020-03-03
PCI-DSS profile on RHEL7/RHEL8 - many failing test scenarios


mildas:
#### Description of problem:
Many test scenarios are failing during the initial stage in the PCI-DSS profile.

It can be caused by incorrect test scenarios or by incorrect OVAL checks.

#### SCAP Security Guide Version:
6ba3fcf

#### Operating System Version:
RHEL 8

#### Steps to Reproduce:

1. `python3 tests/test_suite.py combined --libvirt qemu:///system test_suite_vm                 --datastream /usr/share/xml/scap/ssg/content/ssg-rhel8-ds.xml --mode online --remediate-using bash                 xccdf_org.ssgproject.content_profile_pci-dss`

#### Actual Results:
Output from test suite (passed test scenarios, failed test scenarios during the final stage and irrelevant information have been removed from output):

RHEL7
**xccdf_org.ssgproject.content_rule_smartcard_auth**
ERROR - Script installed_enabled.pass.sh using profile (all) found issue:
ERROR - Rule evaluation resulted in fail, instead of expected pass during initial stage 
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_smartcard_auth'.

RHEL8
**xccdf_org.ssgproject.content_rule_ensure_logrotate_activated**
ERROR - Script logrotate_conf_extra_monthly.fail.sh using profile (all) found issue:
ERROR - Rule evaluation resulted in pass, instead of expected fail during initial stage 
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_ensure_logrotate_activated'.

RHEL7/RHEL8
**xccdf_org.ssgproject.content_rule_file_permissions_etc_group**
ERROR - Script narrow_permissions.pass.sh using profile (all) found issue:
ERROR - Rule evaluation resulted in fail, instead of expected pass during initial stage 
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_file_permissions_etc_group'.

RHEL7/RHEL8
**xccdf_org.ssgproject.content_rule_file_permissions_etc_passwd**
ERROR - Script narrow_permissions.pass.sh using profile (all) found issue:
ERROR - Rule evaluation resulted in fail, instead of expected pass during initial stage 
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_file_permissions_etc_passwd'.

#### Expected Results:
Test scenarios pass during the initial stage.

[...]

PR #6683, opened 2021-03-11
allow stricter permissions in certain file_permissions_* rules

vojtapolasek:
#### Description:

- enable stricter file permissions for the following rules:
  - file_permissions_etc_group
  - file_permissions_etc_passwd

#### Rationale:

There are tests which should be passing, but without enabling of this option they are failing.

- Fixes part of #5213


vojtapolasek:
After discussion this will be fixed in a different PR. Closing now.

PR #6685, opened 2021-03-11
Add ansible and bash remediation for wireless_disable_interfaces.

ggbecker:
#### Description:

- Add ansible and bash remediation for wireless_disable_interfaces.

#### Rationale:

- Fixes [bz#1859487](https://bugzilla.redhat.com/show_bug.cgi?id=1859487)

#### Notes

I don't how to create a test scenario for that since the machine needs to have a wifi device enabled. I've manually verified that the `/proc/net/wireless` on my laptop doesn't contain any wireless devices after running `nmcli radio wifi off` and the OVAL check would pass.

### DIFF
--- a/linux_os/guide/system/network/network-wireless/wireless_software/wireless_disable_interfaces/ansible/shared.yml
+++ b/linux_os/guide/system/network/network-wireless/wireless_software/wireless_disable_interfaces/ansible/shared.yml
@@ -1,0 +1,8 @@
+# platform = multi_platform_all
+# reboot = false
+# strategy = unknown
+# complexity = low
+# disruption = medium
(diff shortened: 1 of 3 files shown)


yuumasato (approved):
LGTM.

@ggbecker Do you think it makes sense to add `platform: machine` to this rule?
I see that other rules in the same group have it...

ggbecker:
> LGTM.
> 
> @ggbecker Do you think it makes sense to add `platform: machine` to this rule?
> I see that other rules in the same group have it...

We should probably add it. Let me add it.

yuumasato (approved):
Thanks!

PR #6693, opened 2021-03-12
remove ensure_logrotate_configured from CIS profiles

vojtapolasek:
#### Description:

- remove the rule from cis for rhel7, rhel8, sle15

#### Rationale:

- the rule is marked as unscored/ manual.

### DIFF
--- a/rhel7/profiles/cis.profile
+++ b/rhel7/profiles/cis.profile
@@ -537,7 +537,6 @@ selections:
     ### 4.2.4 Ensure permissions on all logfiles are configured (Scored)
 
     ## 4.3 Ensure logrotate is configured (Not Scored)
-    - ensure_logrotate_activated
 
     # 5 Access, Authentication and Authorization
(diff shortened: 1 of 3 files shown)


vojtapolasek:
@brett060102  @guangyee  is that okay for sle15?

guangyee:
> @brett060102 @guangyee is that okay for sle15?

@abergmann what do you think? Not sure if we want to keep those type of rules around for a complete profile. I am fine with removing them. But please feel free to disagree.

brett060102:
@vojtapolasek @abergmann is mostly out until Monday, I don't see any reason for not removing it for sle15. We have not done anything with that file in almost a year, but I know nothing about its function or purpose, but since we are only now getting sle15 rules in place, I can't see how removing that one would effect us.

ggbecker (approved):
sle15 folks are not against the removal from sle15/cis.profile and the rest LGTM.

PR #6698, opened 2021-03-15
Make the root of the content available in build_config.yml

yuumasato:
#### Description:

- Make `ssg_root` available in `build_config.yml`
- Update a few of the `build-scripts` to use the `build_config.yml`

#### Rationale:

- Makes it easy for scripts to know the path to the "root of content"
- This avoids the need to manually craft and deduced the path to the root of the project.

The `build_config.yml` and the `product.yml` makes for a good input source for all the information needed to build the contents.


yuumasato:
I realize that what we need is not the `ssg_root`, but the `benchmark_root`.

PR #6699, opened 2021-03-15
Profiles generated from policy contains only rules applicable for the product

yuumasato:
#### Description:

- When generating profiles based on the policy file, include only rules that are applicable to the product.
  i.e.: Only include rules with `prodtype` matching the `product` being built.

#### Rationale:

- The control should be useful for many products, adding a lot of product Jinja conditionals would make it hard to read.


yuumasato:
This is based on few changes already proposed in #6698

yuumasato:
Closing in favor of #6725

PR #6701, opened 2021-03-15
Updated test scenarios to work on containers

matejak:
Added package metadata, and made scenarios more robust - not all files may really be there all the time.

### DIFF
--- a/linux_os/guide/system/logging/log_rotation/ensure_logrotate_activated/tests/logrotate_conf_extra_monthly.fail.sh
+++ b/linux_os/guide/system/logging/log_rotation/ensure_logrotate_activated/tests/logrotate_conf_extra_monthly.fail.sh
@@ -1,4 +1,6 @@
 #!/bin/bash
 
+# packages = logrotate,crontabs
+
 sed -i "s/weekly/daily/g" /etc/logrotate.conf
 echo "monthly" >> /etc/logrotate.conf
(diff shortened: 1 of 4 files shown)


linux_os/guide/system/logging/log_rotation/ensure_logrotate_activated/tests/logrotate_conf_extra_monthly.fail.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash
 
+# packages: logrotate,crontabs
```
ggbecker, line 3: I don't think `colon` is the right syntax here:

https://github.com/ComplianceAsCode/content/blob/22ef2f61b24ff4686d745ba68656ca3b88a99e27/tests/ssg_test_suite/rule.py#L311

ggbecker: @matejak this should be the only change required here. Can you update the PR?

matejak: Correct, thanks!

ggbecker:
I've tested in a few environments and they all look good. If they contain changes from https://github.com/ComplianceAsCode/content/pull/6710 which is already in main branch.

LGTM

PR #6707, opened 2021-03-16
ocp4: Fix file_permissions_etcd_data_dir rule

JAORMX:
This is done by setting the `allow_stricter_permissions` flag to false.

### DIFF
--- a/applications/openshift/master/file_permissions_etcd_data_dir/rule.yml
+++ b/applications/openshift/master/file_permissions_etcd_data_dir/rule.yml
@@ -33,3 +33,4 @@ template:
     vars:
         filepath: /var/lib/etcd/member/
         filemode: '0700'
+        allow_stricter_permissions: "false"


JAORMX:
well... this didn't work :(

ggbecker:
> well... this didn't work :(

the template parameter parsing is wrong. I should have an update soon to make this work

yuumasato:
> the template parameter parsing is wrong. I should have an update soon to make this work

Would it also fix the original problem with `file_permissions_etcd_data_dir`?

ggbecker:
> > the template parameter parsing is wrong. I should have an update soon to make this work
> 
> Would it also fix the original problem with `file_permissions_etcd_data_dir`?

I'm investigating that. It looks like the template does not work properly when it evaluates a directory.

JAORMX:
Closing in favor of https://github.com/ComplianceAsCode/content/pull/6709

PR #6708, opened 2021-03-16
postfix_network_listening_disabled tests

mildas:
#### Description:
Add a new test scenario and fix existing ones. The existing ones didn't work as expected because when `postfix` is not running, the `/etc/postfix/main.cf` file is not checked at all.

### DIFF
--- a/linux_os/guide/services/mail/postfix_client/postfix_network_listening_disabled/tests/correct_commented.fail.sh
+++ b/linux_os/guide/services/mail/postfix_client/postfix_network_listening_disabled/tests/correct_commented.fail.sh
@@ -2,4 +2,5 @@
 # packages = postfix
 
 echo "#inet_interfaces = loopback-only" > /etc/postfix/main.cf
-echo "inet_interfaces = something_different" >> /etc/postfix/main.cf
+systemctl enable postfix
+systemctl start postfix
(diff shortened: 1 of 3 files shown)


linux_os/guide/services/mail/postfix_client/postfix_network_listening_disabled/tests/wrong.fail.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash
 # packages = postfix
 
-echo "inet_interfaces = some_different_interface,loopback-only" > /etc/postfix/main.cf
```
yuumasato, line 4: I see value in testing if a correct substring is not mistaken as a correct configuration, the rule description mentions that the only value configured should be the one from the variable.

yuumasato: I see that the original test "tested" two things at the same time, so it could be put into a test scenario of its own.

mildas: That's a good point. I have added a new test scenario similar to the previous one.

Unfortunately, in the new test scenario, I have to start postfix first and then add the configuration. It's because having `inet_interfaces = some_different_interface,loopback-only` requires to have interface called `loopback-only`. Without that, you can't start postfix.

yuumasato (approved):

PR #6706, opened 2021-03-16
Revert "change default for allow_stricter_permissions to true"

JAORMX:
This reverts commit d57f4f95801c3b5a9dafa7d6bfa715f9b61f6722.

This breaks the ocp4 check `file_permissions_etcd_data_dir`

PR #6709, opened 2021-03-16
Fix file permissions template when checking directories and allowing stricter permissions

ggbecker:
#### Description:

- Update1: Fixed the template to properly assess directory permissions

### DIFF
--- a/shared/templates/file_permissions/oval.template
+++ b/shared/templates/file_permissions/oval.template
@@ -1,23 +1,27 @@
+{{%- if ALLOW_STRICTER_PERMISSIONS %}}
+    {{#
+        When allowing stricter permissions, the check looks for files that do not match the criteria.
+        In this case none_exist is used to make sure that there are no matches.
+    #}}
+  {{% set FILE_EXISTENCE = "none_exist" %}}
+{{%- else -%}}
+  {{%- if MISSING_FILE_PASS -%}}
+    {{# Any number of files can exist, from zero to any #}}
+    {{% set FILE_EXISTENCE = "any_exist" %}}
+  {{%- else -%}}
+    {{# All defined files must exist. When using regex, at least one file must match #}}
+    {{% set FILE_EXISTENCE = "all_exist" %}}
+  {{%- endif -%}}
+{{%- endif -%}}
 <def-group>
   <definition class="compliance" id="{{{ _RULE_ID }}}" version="1">
     {{{ oval_metadata("This test makes sure that " + FILEPATH + " has mode " + FILEMODE + ".
       If the target file or directory has an extended ACL, then it will fail the mode check.
       ") }}}
     <criteria>
-      <criterion comment="Check file mode of {{{ FILEPATH }}}" test_ref="test_file_permissions{{{ FILEID }}}"{{{ ' negate="true"' if ALLOW_STRICTER_PERMISSIONS }}}/>
+      <criterion comment="Check file mode of {{{ FILEPATH }}}" test_ref="test_file_permissions{{{ FILEID }}}"/>
     </criteria>
   </definition>
-  {{%- if ALLOW_STRICTER_PERMISSIONS %}}
-    {{% set FILE_EXISTENCE = "at_least_one_exists" %}}
-  {{%- else -%}}
-    {{%- if MISSING_FILE_PASS -%}}
-      {{# Any number of files can exist, from zero to any #}}
-      {{% set FILE_EXISTENCE = "any_exist" %}}
-    {{%- else -%}}
-      {{# All defined files must exist. When using regex, at least one file must match #}}
-      {{% set FILE_EXISTENCE = "all_exist" %}}
-    {{%- endif -%}}
-  {{%- endif -%}}
       <unix:file_test check="all" check_existence="{{{ FILE_EXISTENCE }}}" comment="Testing mode of {{{ FILEPATH }}}" id="test_file_permissions{{{ FILEID }}}" version="2">
       <unix:object object_ref="object_file_permissions{{{ FILEID }}}" />
       <unix:state state_ref="state_file_permissions{{{ FILEID }}}_mode_{{{ 'not_' if ALLOW_STRICTER_PERMISSIONS }}}{{{ FILEMODE }}}" />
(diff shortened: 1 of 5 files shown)


JAORMX:
Well, CI seems happy

JAORMX:
/test e2e-aws-ocp4-moderate

The test passed but the pod that uploads logs (provided by the CI infra) failed.

yuumasato:
I think the issue is actually on the scanner side, see linked PR in OpenSCAP.

JAORMX:
@yuumasato should we then merge https://github.com/ComplianceAsCode/content/pull/6706 and only merge that default when the scanner fix lands?

yuumasato:
> @yuumasato should we then merge #6706 and only merge that default when the scanner fix lands?

Either that, or we merge #6709, as it seems to provide a reasonable workaround.

JAORMX:
> > @yuumasato should we then merge #6706 and only merge that default when the scanner fix lands?
> 
> Either that, or we merge #6709, as it seems to provide a reasonable workaround.

@yuumasato @ggbecker whatever you guys think is best.

shared/templates/file_permissions/oval.template
```diff
@@ -32,6 +34,9 @@
         <unix:filename operation="pattern match">{{{ FILE_REGEX }}}</unix:filename>
       {{%- else %}}
         <unix:filename xsi:nil="true" />
+      {{%- if ALLOW_STRICTER_PERMISSIONS %}}
+        <filter action="include">state_file_permissions{{{ FILEID }}}_mode_not_{{{ FILEMODE }}}</filter>
+      {{%- endif %}}
```
yuumasato, line 39: The use of `filter action=include` here is not wrong, but may require some explanation as to why we go the approach of "defining that the object with a certain state doesn't exist" opposed "defining the object and checking that it doesn't match a state".

yuumasato (changes requested):

yuumasato:
> @yuumasato @ggbecker whatever you guys think is best.

I'd go with this PR.
The use of `filter action="include"` enables OpenSCAP < 1.3.5 to use the file_permissions template with `allow stricter permissions`.

Revertint the template to `allow_stricter_permissions: false` won't make it work for rules who enable it.

shared/templates/file_permissions/oval.template
yuumasato: The filter should be kept until 1.3.5 is widely deployed, but there is no harm in keeping the filter indefinitely.

shared/templates/file_permissions/oval.template
```diff
@@ -20,7 +20,9 @@
   {{%- endif -%}}
       <unix:file_test check="all" check_existence="{{{ FILE_EXISTENCE }}}" comment="Testing mode of {{{ FILEPATH }}}" id="test_file_permissions{{{ FILEID }}}" version="2">
       <unix:object object_ref="object_file_permissions{{{ FILEID }}}" />
+      {{%- if not ALLOW_STRICTER_PERMISSIONS %}}
       <unix:state state_ref="state_file_permissions{{{ FILEID }}}_mode_{{{ 'not_' if ALLOW_STRICTER_PERMISSIONS }}}{{{ FILEMODE }}}" />
```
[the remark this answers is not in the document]
ggbecker: I would rather remove this condition now that we know this should  work for openscap >= 1.3.5

shared/templates/file_permissions/oval.template
ggbecker: I'll add a comment. I'll also try to unify this filter with the existing one that is being used for the file option (when the check is configured to look for a file)

ggbecker:
> > @yuumasato @ggbecker whatever you guys think is best.
> 
> I'd go with this PR.
> The use of `filter action="include"` enables OpenSCAP < 1.3.5 to use the file_permissions template with `allow stricter permissions`, and will work after the fix is in the wild too.
> 
> Revertint the template to `allow_stricter_permissions: false` won't make it work for rules who enable it.

I've updated the pull requests with some improvements to the code structure and applied suggestions. I've also switched to read to review, please take a look again. Commits that enabled/disabled the `allow_stricter_permissions` were stripped away from the PR.

ggbecker:
@yuumasato please advise if e87eb6b is the best solution for:

https://github.com/ComplianceAsCode/content/runs/2130339562?check_suite_focus=true#step:5:141

`oval:ssg-test_file_permissions_backup_etc_group:tst:1 - No state should be referenced when check_existence has a value of 'none_exist'.`

shared/templates/file_permissions/oval.template
```diff
@@ -1,3 +1,19 @@
+{{%- if ALLOW_STRICTER_PERMISSIONS %}}
+    {{#
+        When allowing stricter permissions, the check looks for files/directories that
+        do not match the permissions (STATEMODE). The criterion is then negated to make
+        sure there are no files/directories matched.
+    #}}
+  {{% set FILE_EXISTENCE = "at_least_one_exists" %}}
```
yuumasato, line 7: I think this should be `any_exist`.
If there are no files/directories with extra permissions bits set, we are fine.

shared/templates/file_permissions/oval.template
```diff
@@ -1,3 +1,19 @@
+{{%- if ALLOW_STRICTER_PERMISSIONS %}}
+    {{#
+        When allowing stricter permissions, the check looks for files/directories that
+        do not match the permissions (STATEMODE). The criterion is then negated to make
```
yuumasato, line 4: Text suggestion:
`looks for files/directories that have any extra permission bit set.`

yuumasato (commented):
> @yuumasato please advise if [e87eb6b](https://github.com/ComplianceAsCode/content/commit/e87eb6b3ba5ca0e4d3aadcc1f56975b442db01be) is the best solution for:

This looks to be most future proof and easier to update later, :)

With OpenSCAP < 1.3.5, the filter will include the objects we want (files/directories with extra permission bits set), and although the State comparison is buggy, we know all objects evaluated there will have at least one Entity trespassing the mask, thus matching and working as expected.

With OpenSCAP >= 1.3.5, the Entity matching will work correctly and we can just remove the filter. And forgetting to remove the filter will cause no harm.

shared/templates/file_permissions/oval.template
yuumasato: Extra suggestion: `The permissions (STATEMODE) is negated (inverted?) and becomes a mask for permissions that should not be set.`

shared/templates/file_permissions/oval.template
ggbecker: If we use `any_exist` every check would fail. For example, if no object is collected (there are no files/directories with wrong permissions), this check would pass (`any_exist` allows) and subsequentially fail because the criterion has `negate=true`. That means, we should keep the `at_least_one_exists` to cover this case.

yuumasato: You are right, the `negate` confused me, :)

yuumasato (approved):
Nice fix @ggbecker :)

LGTM, just pending results from OCP rules.

JAORMX (approved):
Thanks for taking a look at this folks! Let's wait for CI now.

JAORMX:
@ggbecker yeah... the failure is unrelated. It was the cluster logging stack failing to install (yet another operator).

JAORMX:
ci/prow/e2e-aws-ocp4-cis-node passed, which was the affected job. Merging.

PR #6710, opened 2021-03-16
fix remediations of ensure_logrotate_activated

vojtapolasek:
#### Description:

- fix regexes in Bash and Ansible remediation
- fix regex in test

#### Rationale:

It turned out that regex was too strict and some occurences of "weekly / monthly / yearly" did not get removed properly.


yuumasato (approved):
LGTM.

PR #6713, opened 2021-03-17
mount_option ansible remediation - remediate when mount point is not in mounted

mildas:
#### Description:
Add task that crafts line for fstab if a mount point is not mounted.
Add new test scenario that tests if a remediation is able to create partition from scratch.

#### Rationale:
If a mount point is not expected to exist and it is not mounted, then ansbile remediation didn't perform any remediation. This case occurred when  `mount_has_to_exist: 'no'` (only moun_option_dev_shm_* rules).
Bash remediation doesn't require any change, its behavior is different:

Bash remediation:

1. if `mount_has_to_exist: 'yes'`, check the mount point has record in fstab - if no record, no remediation is performed
2. ensure the mount option is in fstab
3. ensure the mount point is mounted

Ansible remediation:

1. if `mount_has_to_exist: 'yes'`, check the mount point has record in fstab - if no record, other tasks are skipped
2. **if `mount_has_to_exist: 'no'`, check the mount point is mounted, if not mounted, other tasks are skipped**
3. create a variable with mount information
4. ensure the mount option is in fstab
5. ensure the mount point is mounted

PR changes ansible remediation step 2 to:
- if `mount_has_to_exist: 'no'`, check the mount point is mounted, if not mounted, craft fstab record

### DIFF
--- a/linux_os/guide/system/permissions/partitions/mount_option_dev_shm_nodev/tests/no_partition.fail.sh
+++ b/linux_os/guide/system/permissions/partitions/mount_option_dev_shm_nodev/tests/no_partition.fail.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+. $SHARED/partition.sh
+
+clean_up_partition /dev/shm
--- a/linux_os/guide/system/permissions/partitions/mount_option_dev_shm_noexec/tests/no_partition.fail.sh
+++ b/linux_os/guide/system/permissions/partitions/mount_option_dev_shm_noexec/tests/no_partition.fail.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+. $SHARED/partition.sh
+
+clean_up_partition /dev/shm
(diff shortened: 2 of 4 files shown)


mildas:
Tested with few randomly selected mount_option_* rules:
`$ python3 tests/test_suite.py rule --profile cis --libvirt qemu:///system test-suite-rhel8-partitions --remediate-using ansible --datastream build/ssg-rhel8-ds.xml mount_option_var_log_audit_nosuid mount_option_var_noexec mount_option_nosuid_removable_partitions mount_option_dev_shm_nosuid mount_option_dev_shm_nodev mount_option_tmp_nodev`

vojtapolasek:
In my opinion this brings the Ansible remediation to the same level of Bash remediation. Tests are passing. Thank you for changes.

vojtapolasek:
@mildas  please investigate failing ansible-role-test, there are errors reported.

mildas:
@vojtapolasek aha, thanks, I didn't check results from the ansible-role test.
I had to update the condition that was checking if variable is empty. Now it's passing.

vojtapolasek:
lgtm, thank you

PR #6730, opened 2021-03-19
Need to set pipefail when using shell in ansible

brett060102:
Upstream ansible lint tests require pipefail when using shell
in ansible.

#### Description:

- fix ansible-lint pipefail check failures
- Added set pipefail where possible
- Added #noqa 306 in cases were grep was used
#### Rationale:

- no ansible link failures


ggbecker:
There is still one more issue here: https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/accounts/accounts-banners/gui_login_banner/gui_login_dod_acknowledgement/ansible/shared.yml#L33

`[E601] Don't compare to literal True/False`

vojtapolasek:
Thank you for the fix, now all test pass. Merging.

PR #6732, opened 2021-03-22
ocp4/etcd: Fix rule checks for 4.8

JAORMX:
From 4.7 to 4.8 the directory with the etcd certificate/keys changed. So
an update to the rules is needed in order for us to cater for the
changes and still have these checks applicable in order versions.


jhrozek (approved):
/lgtm

PR #6737, opened 2021-03-23
Add rule for checking selinux is not disabled in coreos

JAORMX:
This also enables the `oval_argument_value_in_line_criterion` template
to be negated and it adds to the `coreos_kernel_option` OVAL template as well.

### DIFF
--- a/docs/manual/developer/06_contributing_with_content.md
+++ b/docs/manual/developer/06_contributing_with_content.md
@@ -1268,6 +1268,21 @@ the following to `rule.yml`:
 
 -   Languages: OVAL
 
+#### coreos_kernel_option
+-   Checks that `argument=value` pair is present in the kernel arguments.
+    Note that this applies to Red Hat CoreOS.
+
+-   Parameters:
+
+    -   **arg_name** - Argument name, eg. `audit`
+
+    -   **arg_value** - Argument value, eg. `'1'`
+
+    -   **negate** - negates the check, which then ensures that
+        `argument=value` is not present in the kernel arguments.
+
+-   Languages: OVAL, Kubernetes
+
 #### dconf_ini_file
 -   Checks for `dconf` configuration. Additionally checks if the
     configuration is locked so it cannot be overriden by the user.
--- a/docs/manual/developer/06_contributing_with_content.md
+++ b/docs/manual/developer/06_contributing_with_content.md
@@ -1278,7 +1278,7 @@ the following to `rule.yml`:
 
     -   **arg_value** - Argument value, eg. `'1'`
 
-    -   **negate** - negates the check, which then ensures that
+    -   **arg_negate** - negates the check, which then ensures that
         `argument=value` is not present in the kernel arguments.
 
 -   Languages: OVAL, Kubernetes
--- a/linux_os/guide/system/selinux/coreos_enable_selinux_kernel_argument/rule.yml
+++ b/linux_os/guide/system/selinux/coreos_enable_selinux_kernel_argument/rule.yml
@@ -1,0 +1,53 @@
+documentation_complete: true
+
+prodtype: fedora,rhcos4,ol7,ol8,rhel7,rhel8,rhv4,sle15
+
+title: 'Ensure SELinux Not Disabled in the kernel arguments'
+
+description: |-
+    SELinux can be disabled at boot time by disabling it via a kernel argument.
+    Remove any instances of <tt>selinux=0</tt> from the kernel arguments in that
+    file to prevent SELinux from being disabled at boot.
+
+rationale: |-
+    Disabling a major host protection feature, such as SELinux, at boot time prevents
+    it from confining system services at boot time.  Further, it increases
+    the chances that it will remain off during system operation.
+
+severity: medium
+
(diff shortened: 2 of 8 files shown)


yuumasato:
@openscap-ci test this please

evgenyz:
Looks good. @yuumasato what do you think about the template argument name `negate`?

evgenyz (approved):

linux_os/guide/system/selinux/coreos_enable_selinux_kernel_argument/rule.yml
```diff
@@ -0,0 +1,52 @@
...
+    <tt>MachineConfig</tt> object.
+
+template:
+    name: coreos_kernel_option
+    vars:
+        arg_name: selinux
+        arg_value: '0'
+        arg_negate: 'true'
+    backends:
+        kubernetes: "off"
```
ggbecker, line 52: please add a new line here

shared/templates/coreos_kernel_option/kubernetes.template
```diff
@@ -1,3 +1,4 @@
+{{%- if not ARG_NEGATE %}}
```
ggbecker, line 1: I don't think we need this macro, the `backends` should do trick. Or do you still want to have this?

linux_os/guide/system/selinux/grub2_enable_selinux/rule.yml
```diff
@@ -20,7 +20,6 @@ severity: medium
 identifiers:
     cce@rhel7: CCE-26961-3
     cce@rhel8: CCE-80827-9
-    cce@rhcos4: CCE-82666-9
```
ggbecker, line 23: Was this rule ever shipped in a release? If yes, I don't think it makes sense to remove the CCE even if it's not part of a profile anymore.

ggbecker (changes requested):
For the new template argument `arg_negate`, we have to include a pre-processing exactly like this: https://github.com/ComplianceAsCode/content/blob/master/shared/templates/sshd_lineinfile/template.py otherwise it would fail to detect `false` parameter.

I also notice that the template `coreos_kernel_option` misses documentation in the official devel doc page. It would probably go here: https://github.com/ComplianceAsCode/content/blame/master/docs/manual/developer/06_contributing_with_content.md#L1270

shared/templates/coreos_kernel_option/kubernetes.template
JAORMX: nope, let me remove it.

linux_os/guide/system/selinux/grub2_enable_selinux/rule.yml
JAORMX: OK, I'll change it.

linux_os/guide/system/selinux/coreos_enable_selinux_kernel_argument/rule.yml
JAORMX: Done

JAORMX:
> For the new template argument `arg_negate`, we have to include a pre-processing exactly like this: https://github.com/ComplianceAsCode/content/blob/master/shared/templates/sshd_lineinfile/template.py otherwise it would fail to detect `false` parameter.
> 
> I also notice that the template `coreos_kernel_option` misses documentation in the official devel doc page. It would probably go here: https://github.com/ComplianceAsCode/content/blame/master/docs/manual/developer/06_contributing_with_content.md#L1270

Done

JAORMX:
/test e2e-aws-ocp4-cis

remediation timed out. Unrelated to this.

JAORMX:
this passed the e2e test:
```
    helpers.go:676: Result - Name: e2e-moderate-master-coreos-enable-selinux-kernel-argument - Status: PASS - Severity: medium
    helpers.go:747: Rule coreos_enable_selinux_kernel_argument matched expected result
```

docs/manual/developer/06_contributing_with_content.md
```diff
@@ -1268,6 +1268,21 @@ the following to `rule.yml`:
...
+-   Checks that `argument=value` pair is present in the kernel arguments.
+    Note that this applies to Red Hat CoreOS.
+
+-   Parameters:
+
+    -   **arg_name** - Argument name, eg. `audit`
+
+    -   **arg_value** - Argument value, eg. `'1'`
+
+    -   **negate** - negates the check, which then ensures that
```
ggbecker, line 1281:
```suggestion
    -   **arg_negate** - negates the check, which then ensures that
```

The actual parameter name is `arg_negate` although I don't know which of them makes more sense, probably something more verbose like `negate_check` or something

ggbecker: But I'm fine with `arg_negate` as well

ggbecker (changes requested):
A minor issue in the documentation. Other than that, it looks good.

ggbecker (approved):
LGTM. There are some tests still running but they were verified to be working before the last commit, no need to wait since the last commit only addresses one minor issue in the text docs.

PR #6738, opened 2021-03-23
Add tests for no files unowned by user/group rules

matusmarhefka:
#### Description:

- Add tests for rule `no_files_unowned_by_user`.
- Add tests for rule `file_permissions_ungroupowned`.
- Update descriptions of these rules as the `-fstype local` is not supported in the GNU find, see: https://www.gnu.org/software/findutils/manual/html_node/find_html/Filesystems.html#Filesystems


vojtapolasek:
LGTM, thank you

PR #6743, opened 2021-03-23
rhcos4: Add check for nousb kernel argument

JAORMX:
This adds a check that verifies if the nousb kernel argument is set. If
it isn't it should generate an appropraite remediation.

This also enhances the `coreos_kernel_option` template and the
`oval_argument_value_in_line_criterion` and
`oval_argument_value_in_line_test` macros to not require `arg_value`
anymore. If this parameter is not given, it'll only use `arg_name`.


ggbecker (commented):
Since you are making the template parameter optional, you can update the documentation as well here: https://github.com/ComplianceAsCode/content/blame/master/docs/manual/developer/06_contributing_with_content.md#L1279

shared/templates/coreos_kernel_option/template.py
```diff
@@ -3,4 +3,6 @@
 
 def preprocess(data, lang):
     data["arg_negate"] = parse_template_boolean_value(data, parameter="arg_negate", default_value=False)
+    if "arg_value" not in data:
+        data["arg_value"] = ""
```
ggbecker, line 7:
```suggestion
```

This is probably not necessary. Jinja will be able to properly evaluate `{{% if ARG_VALUE %}}`

JAORMX: I got some error at some point and ended up adding it here.

JAORMX:
@ggbecker I think this is good to go, can you revisit this?

docs/manual/developer/06_contributing_with_content.md
```diff
@@ -1276,7 +1276,7 @@ the following to `rule.yml`:
 
     -   **arg_name** - Argument name, eg. `audit`
 
-    -   **arg_value** - Argument value, eg. `'1'`
+    -   **arg_value** - Argument value, eg. `'1'` (Optional)
```
ggbecker, line 1279: I feel like we should explain a bit more the template behavior when `arg_value` is not provided. For example, that the template will render a test that only looks for the presence of `arg_name`. The whole template is built around the idea of checking `argument=value`, so it might confuse people seeing the word `optional` with no further explanation. Sorry to nitpick here ;)

ggbecker (changes requested):
One more nitpick regarding documentation.

docs/manual/developer/06_contributing_with_content.md
JAORMX: OK, added more explanation on this doc, what do you think?

docs/manual/developer/06_contributing_with_content.md
```diff
@@ -1276,7 +1276,8 @@ the following to `rule.yml`:
 
     -   **arg_name** - Argument name, eg. `audit`
 
-    -   **arg_value** - Argument value, eg. `'1'`
+    -   **arg_value** - Argument value, eg. `'1'`. This parameter is optional,
+        and if ommited, this template will only use **arg_name**.
```
ggbecker, line 1280:
```suggestion
        and if omitted, this template will only use **arg_name**.
```

ggbecker (approved):
LGTM, thanks for addressing the requests.

Issue #6729, opened 2021-03-19
Generic python shebang?


cipherboy:
#### Description of problem:

A lot of shebangs are currently using `#!/usr/bin/env python2`:

```console
shared/transforms/pcidss/generate_pcidss_json.py:1:#!/usr/bin/env python2
shared/transforms/pcidss/transform_benchmark_to_pcidss.py:1:#!/usr/bin/env python2
build-scripts/relabel_ids.py:1:#!/usr/bin/env python2
build-scripts/unselect_empty_xccdf_groups.py:1:#!/usr/bin/env python2
build-scripts/build_templated_content.py:1:#!/usr/bin/env python2
build-scripts/generate_bash_remediation_functions.py:1:#!/usr/bin/env python2
build-scripts/verify_references.py:1:#!/usr/bin/env python2
build-scripts/oscap_svg_support.py:1:#!/usr/bin/env python2
build-scripts/combine_ovals.py:1:#!/usr/bin/env python2
build-scripts/enable_derivatives.py:1:#!/usr/bin/env python2
build-scripts/build_all_guides.py:1:#!/usr/bin/env python2
build-scripts/combine_remediations.py:1:#!/usr/bin/env python2
build-scripts/sds_move_ocil_to_checks.py:1:#!/usr/bin/env python2
build-scripts/add_stig_references.py:1:#!/usr/bin/env python2
build-scripts/build_profile_remediations.py:1:#!/usr/bin/env python2
build-scripts/cpe_generate.py:1:#!/usr/bin/env python2
build-scripts/generate_fixes_xml.py:1:#!/usr/bin/env python2
utils/fix-rules.py:1:#!/usr/bin/env python2
utils/count_oval_objects.py:1:#!/usr/bin/env python2
utils/find_duplicates.py:1:#!/usr/bin/env python2
utils/generate_contributors.py:1:#!/usr/bin/env python2
utils/xccdf2csv-stig.py:1:#!/usr/bin/env python2
utils/ansible_playbook_to_role.py:1:#!/usr/bin/env python2
utils/testoval.py:1:#!/usr/bin/env python2
tests/ensure_paths_are_short.py:1:#!/usr/bin/env python2
tests/test_suite.py:1:#!/usr/bin/env python2
tests/ssg_test_suite/xml_operations.py:1:#!/usr/bin/env python2
tests/ssg_test_suite/oscap.py:1:#!/usr/bin/env python2
[...]

PR #6744, opened 2021-03-23
Switch to generic python shebang

cipherboy:
This should let us autodetect whichever python version is default on a
given platform. As long as we continue to maintain python2/python3
compatibility, this should be fine.

Resolves: #6729

### DIFF
--- a/build-scripts/add_stig_references.py
+++ b/build-scripts/add_stig_references.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
 
 from __future__ import print_function
 
--- a/build-scripts/build_all_guides.py
+++ b/build-scripts/build_all_guides.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
 
 from __future__ import print_function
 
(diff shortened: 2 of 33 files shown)


JAORMX (approved):

yuumasato (approved):
Thanks @cipherboy

I'm not an expert, but it looks like there is no python shebang that works for all cases, nonetheless this is an enhancement and makes things consistent.

> As long as we continue to maintain python2/python3
> compatibility, this should be fine.

We have test for this: https://jenkins.complianceascode.io/view/SCAP%20Security%20Guide/job/scap-security-guide-py2-py3-sanity/
And I actually found out that it is giving up false positives due to missing python2 dependencies, :laughing:.
We should look into it soon.

The build for `rhel7` node was successful, so all things look fine for now.

cipherboy:
Thanks @yuumasato! I think RPM packaging in Fedora mangles Python shebangs from `/usr/bin/env <x>` over to `/usr/bin/<x>`. Perhaps we want to template them using CMake and provide all scripts under `/build/bin`? I don't know, that seems less useful on average.

Regardless, with a python-means-python3 package providing `/usr/bin/python -> /usr/bin/python3` symlink, this shebang seems to be the lesser of all evils :)

yuumasato:
Yes, Fedora changes the Python shebangs.
But we don't package the build-scripts or utilities in Fedora, and I'm not aware of other distro shipping the tooling.
I think the packages have mostly shipped the DataStreams, Playbooks / Bash Scripts and HTML guides / tables.

The shebangs may have to be revisited if / when these scripts start being packaged.

PR #6747, opened 2021-03-24
Add rule for RHCOS to check for interactive boot being disabled

JAORMX:
This was implemented for grub2 but we need a check for RHCOS. This adds
that.


ggbecker (approved):
LGTM

PR #6748, opened 2021-03-24
Updated no_rsh_trust_files rule and added tests

matusmarhefka:
This rule should cover only rhosts files. The OVAL has
been updated to not search for shosts files which are
handled by different rules:
* `no_host_based_files`
* `no_user_host_based_files`  

Both Bash and Ansible remediations have been aligned
with the OVAL check and new tests have been added.

### DIFF
--- a/linux_os/guide/services/obsolete/r_services/no_rsh_trust_files/bash/shared.sh
+++ b/linux_os/guide/services/obsolete/r_services/no_rsh_trust_files/bash/shared.sh
@@ -1,6 +1,5 @@
 # platform = Red Hat Virtualization 4,multi_platform_ol,multi_platform_rhel
-find /home -maxdepth 2 -type f -name .rhosts -exec rm -f '{}' \;
 
-if [ -f /etc/hosts.equiv ]; then
-	/bin/rm -f /etc/hosts.equiv
-fi
+find /root -xdev -type f -name ".rhosts" -exec rm -f {} \;
+find /home -maxdepth 2 -xdev -type f -name ".rhosts" -exec rm -f {} \;
+rm -f /etc/hosts.equiv
--- a/linux_os/guide/services/obsolete/r_services/no_rsh_trust_files/oval/shared.xml
+++ b/linux_os/guide/services/obsolete/r_services/no_rsh_trust_files/oval/shared.xml
@@ -7,26 +7,26 @@
       <criterion test_ref="test_no_rsh_trust_files_etc" negate="true" />
     </criteria>
   </definition>
-  <unix:file_test check="all" check_existence="at_least_one_exists" comment="look for .rhosts or .shosts in /root" id="test_no_rsh_trust_files_root" version="1">
+  <unix:file_test check="all" check_existence="at_least_one_exists" comment="look for .rhosts in /root" id="test_no_rsh_trust_files_root" version="1">
     <unix:object object_ref="object_no_rsh_trust_files_root" />
   </unix:file_test>
-  <unix:file_test check="all" check_existence="at_least_one_exists" comment="look for .rhosts or .shosts in /home" id="test_no_rsh_trust_files_home" version="1">
+  <unix:file_test check="all" check_existence="at_least_one_exists" comment="look for .rhosts in /home" id="test_no_rsh_trust_files_home" version="1">
     <unix:object object_ref="object_no_rsh_trust_files_home" />
   </unix:file_test>
-  <unix:file_test check="all" check_existence="at_least_one_exists" comment="look for /etc/hosts.equiv or /etc/shosts.equiv" id="test_no_rsh_trust_files_etc" version="1">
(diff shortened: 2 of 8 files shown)


linux_os/guide/services/obsolete/r_services/no_rsh_trust_files/oval/shared.xml
```diff
@@ -7,26 +7,26 @@
...
-  <unix:file_test check="all" check_existence="at_least_one_exists" comment="look for .rhosts or .shosts in /home" id="test_no_rsh_trust_files_home" version="1">
+  <unix:file_test check="all" check_existence="at_least_one_exists" comment="look for .rhosts in /home" id="test_no_rsh_trust_files_home" version="1">
     <unix:object object_ref="object_no_rsh_trust_files_home" />
   </unix:file_test>
-  <unix:file_test check="all" check_existence="at_least_one_exists" comment="look for /etc/hosts.equiv or /etc/shosts.equiv" id="test_no_rsh_trust_files_etc" version="1">
+  <unix:file_test check="all" check_existence="at_least_one_exists" comment="look for /etc/hosts.equiv" id="test_no_rsh_trust_files_etc" version="1">
     <unix:object object_ref="object_no_rsh_trust_files_etc" />
   </unix:file_test>
-  <unix:file_object comment="look for .rhosts or .shosts in /root" id="object_no_rsh_trust_files_root" version="1">
+  <unix:file_object comment="look for .rhosts in /root" id="object_no_rsh_trust_files_root" version="1">
```
matusmarhefka, line 19: I am just not sure about this part? Is it enough to look for `.rhosts` files just within `/root` and `/home` or should we search the whole filesystem?

yuumasato: The `.rhosts` file are used by the "r-commands' and are effective when present on the user's home directory of the users.
https://www.mkssoftware.com/docs/man4/rhosts.4.asp

linux_os/guide/services/obsolete/r_services/no_rsh_trust_files/bash/shared.sh
```diff
@@ -1,6 +1,10 @@
...
-	/bin/rm -f /etc/hosts.equiv
-fi
+# Identify local mounts
+MOUNT_LIST=$(df --local | awk '{print $6}' | grep -v Mounted | grep -v '^/dev')
+
+# Find file on each listed mount point
+for cur_mount in ${MOUNT_LIST}
+do
+	find ${cur_mount} -xdev -type f \( -name ".rhosts" -o -name "hosts.equiv" \) -exec rm -f {} \;
+done
```
yuumasato, line 10: I think it should be enough to look into the `/home` and `/root`.

matusmarhefka: Is it? because if you look at the `/etc/passwd` not all users have their home directories only under `/home` or `/root`.

matusmarhefka: We can probably write remediation based on the `6.2.13` (https://workbench.cisecurity.org/sections/300333/recommendations/536114) but if we would like to have similar OVAL check, wouldn't it be too complex?

yuumasato: I see, you are correct, :).
So the check and remediation could look for the users home and act only in those directories.

The OVAL wouldn't be too complex, but would require defining a variable based on contents of `/etc/passwd`.

matusmarhefka: I've looked into this and if we would try to check it for every user from `/etc/passwd` the check would be traversing the whole FS so I've decided to leave the OVAL check as it is. Just for the record, following users have `/` as their home directory on my system (most of them are services anyway):
```
nobody /
dbus /
systemd-coredump /
systemd-network /
systemd-resolve /
polkitd /
qemu /
usbmuxd /
radvd /
nm-openvpn /
nm-openconnect /
tcpdump /
systemd-timesync /
flatpak /
firebird /
```
I even found that some users like `rtkit` have `/proc` as their home directory which doesn't make sense at all to check within this rule.

matusmarhefka: I will update remediations so they are aligned with the OVAL.

matusmarhefka:
@yuumasato Now it should be ready for review. I think I will also update OVAL checks and remediations of `no_host_based_files` and `no_user_host_based_files` rules as they are also checking the whole FS (but in a separate PR).

yuumasato (approved):
LGTM.

@matusmarhefka Can you just rebase to trigger CI?
Once tests pass I'll merge it.

PR #6749, opened 2021-03-25
Add tests for rule account_unique_name

matusmarhefka:
New test scenarios for the rule `account_unique_name`.

### DIFF
--- a/linux_os/guide/system/accounts/accounts-restrictions/account_expiration/account_unique_name/tests/all_unique.pass.sh
+++ b/linux_os/guide/system/accounts/accounts-restrictions/account_expiration/account_unique_name/tests/all_unique.pass.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+cut -d: -f1 /etc/passwd | sort | uniq -d | while read user_name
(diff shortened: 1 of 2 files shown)


matusmarhefka:
@yuumasato This rule is not selected in the RHEL7 CIS profile but it is scored in the official RHEL7 CIS benchmark. What would you say if we add it into the RHEL7 profile?

jan-cerny:
I believe that we can have the test scenarios even if the rule isn't part of the CIS profile because it's a part of other profiles.

```
[jcerny@localhost scap-security-guide{pr/6749}]$ tests/test_rule_in_container.sh account_unique_name
INFO - Logging into /home/jcerny/work/git/scap-security-guide/logs/rule-custom-2021-03-30-0956-1/test_suite.log
INFO - xccdf_org.ssgproject.content_rule_account_unique_name
INFO - Script all_unique.pass.sh using profile (all) OK
INFO - Script no_unique.fail.sh using profile (all) OK
[jcerny@localhost scap-security-guide{pr/6749}]$ 

```

LGTM

PR #6752, opened 2021-03-25
Add package_* test scenarios

mildas:
#### Description:
Add test scenarios to `package_*_installed` and to `package_*_removed` rules from CIS profile.

Each rule has test scenario for installation and test scenario for removal of the package except `libselinux`. The `libselinux` package has only installation test scenario because after removal of the package, the system becomes unusable - commands start to print `libselinux.so.1: cannot open shared object file: No such file or directory` (libselinux is dependency of several core packages, e.g. systemd).

### DIFF
--- a/linux_os/guide/services/ldap/openldap_client/package_openldap-clients_removed/tests/installed.fail.sh
+++ b/linux_os/guide/services/ldap/openldap_client/package_openldap-clients_removed/tests/installed.fail.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+yum install -y openldap-clients
--- a/linux_os/guide/services/ldap/openldap_client/package_openldap-clients_removed/tests/installed.fail.sh
+++ b/linux_os/guide/services/ldap/openldap_client/package_openldap-clients_removed/tests/installed.fail.sh
@@ -1,3 +1,4 @@
 #!/bin/bash
+# package = yum
 
 yum install -y openldap-clients
--- a/linux_os/guide/services/ldap/openldap_client/package_openldap-clients_removed/tests/removed.pass.sh
+++ b/linux_os/guide/services/ldap/openldap_client/package_openldap-clients_removed/tests/removed.pass.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+yum remove -y openldap-clients
--- a/linux_os/guide/services/ntp/package_chrony_installed/tests/installed.pass.sh
+++ b/linux_os/guide/services/ntp/package_chrony_installed/tests/installed.pass.sh
@@ -1,3 +1,4 @@
 #!/bin/bash
+# package = yum
 
 yum install -y chrony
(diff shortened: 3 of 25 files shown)


linux_os/guide/services/ldap/openldap_client/package_openldap-clients_removed/tests/installed.fail.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+yum install -y openldap-clients
```
matusmarhefka, line 3: What about using `# packages = openldap-clients`? I think it would perform better in case of combined mode as packages from this metadata should be installed at once before running the tests. Is that correct @matejak ?

mildas: Benefit I see in the implemented approach is that it's clear that the installation is the actual test scenario.

With the `# packages = ...` approach, the installation will be part of test environment preparation and scenario will be empty (or it might be required to call some command in the scenario like `true` but nothing relevant to the test).

jan-cerny: I prefer Milan's approach with this test

matusmarhefka: > Benefit I see in the implemented approach is that it's clear that the installation is the actual test scenario.
> 
> With the `# packages = ...` approach, the installation will be part of test environment preparation and scenario will be empty (or it might be required to call some command in the scenario like `true` but nothing relevant to the test).

Yes, my idea was to have it like that as fetching yum database in every test scenario will significantly increase the tests run time when testing all rules from a profile, for example in `combined` mode. Isn't it actually obvious from the filename of the test scenario that it is about installation/removal?
To summarize, I am not strictly against this approach, but I would still prefer if we would use the `packages` metadata. I think that test suite run times are pretty high and I don't want to increase them even more if we don't have to.

matejak: I see both sides, and I think that there may be a solution - if you declare that you need the `yum` package, the test environment will have the package metadata up-to-date, so it will be only the installation that takes time for that scenario, and that's IMO OK. Moreover, the test scenario clearly needs `yum`, so that's not a hack.
Without that, I agree that every scenario would fetch the metadata again, which I see as a great waste of time and of electrical power.

mildas: Interesting idea. I have added that in last commit, can you confirm that it's what you meant?

matusmarhefka: @mildas Yes, thank you for updating test scenarios and thanks to @matejak for this clever idea :)

PR #6753, opened 2021-03-25
Update tests of account_disable_post_pw_expiration

matusmarhefka:
Updated pass scenario so it works with all profiles,
corrected wrong value scenario and added a new scenario
for case when the feature is disabled.


mildas:
Works great. Thank you

PR #6754, opened 2021-03-25
Reorganize env and product yaml

yuumasato:
#### Description:

- This adds 393fbdc and ea7dad8 on top of #6725 reorganzing a bit the `env_yaml` and the `product_yaml`.
- Moves loading of `product_cpes` to `get_product_yaml()`.

#### Rationale:

- A few aspects that were product specific were being amalgamated in `open_environment()`

### DIFF
--- a/build-scripts/compile_profiles.py
+++ b/build-scripts/compile_profiles.py
@@ -54,7 +54,7 @@ def get_env_yaml(build_config_yaml, product_yaml):
     if build_config_yaml is None or product_yaml is None:
         return None
 
-    env_yaml = ssg.yaml.open_environment(build_config_yaml, product_yaml)
+    env_yaml = ssg.environment.open_environment(build_config_yaml, product_yaml)
     return env_yaml
 
 
--- a/build-scripts/cpe_generate.py
+++ b/build-scripts/cpe_generate.py
@@ -10,6 +10,7 @@ import argparse
 
 import ssg.build_cpe
 import ssg.id_translate
+import ssg.products
 import ssg.xml
 import ssg.yaml
 
@@ -29,7 +30,12 @@ def parse_args():
         "definitions and the tests, states objects and variables it "
         "references, and then write them into a standalone OVAL CPE file, "
         "along with a synchronized CPE dictionary file.")
-    p.add_argument("product", help="Name of the product")
+    p.add_argument(
+        "--product-yaml",
+        help="YAML file with information about the product we are building. "
+        "e.g.: ~/scap-security-guide/rhel7/product.yml "
+        "needed for autodetection of profile root"
+    )
     p.add_argument("idname", help="Identifier prefix")
     p.add_argument("cpeoutdir", help="Artifact output directory")
     p.add_argument("shorthandfile", help="shorthand xml to generate "
@@ -106,7 +112,9 @@ def main():
     translator = ssg.id_translate.IDTranslator(args.idname)
     ovaltree = translator.translate(ovaltree)
 
-    newovalfile = args.idname + "-" + args.product + "-" + os.path.basename(args.ovalfile)
+    product_yaml = ssg.products.get_product_yaml(args.product_yaml)
+    product = product_yaml["product"]
+    newovalfile = args.idname + "-" + product + "-" + os.path.basename(args.ovalfile)
     newovalfile = newovalfile.replace("oval-unlinked", "cpe-oval")
     ssg.xml.ElementTree.ElementTree(ovaltree).write(args.cpeoutdir + "/" + newovalfile)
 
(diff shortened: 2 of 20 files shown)


yuumasato:
Does this reorg make sense?

build-scripts/compile_profiles.py
```diff
@@ -54,7 +54,7 @@ def get_env_yaml(build_config_yaml, product_yaml):
     if build_config_yaml is None or product_yaml is None:
         return None
 
-    env_yaml = ssg.yaml.open_environment(build_config_yaml, product_yaml)
+    env_yaml = ssg.environment.open_environment(build_config_yaml, product_yaml)
```
vojtapolasek, line 57: I notice that in other files where you did this change, you also altered the import statement. Here you did not. Nothing seems to break during build, but I rather ask.

vojtapolasek (commented):
Looks good, thank you. This is important work. It seems that it builds correctly.
There are some notices from pep8 bot, could you check it please?
Also there has been another utility merged recently - render-rules.py. Is it also affected?

build-scripts/compile_profiles.py
yuumasato: The script didn't have an `import ` for the `yaml` module. So I just followed the same format and replaced `ssg.yaml` with `ssg.environment`.

vojtapolasek:
Seems that there is a problem with platform called ocp4.8.

build-scripts/compile_profiles.py
cipherboy: I think this happens because `ssg.build_yaml` already imports `ssg.{yaml->environment}` and so you don't strictly need to duplicate the import.

ssg/products.py
```diff
@@ -4,17 +4,70 @@
...
+    if "uid_min" not in existing_properties:
+        result["uid_min"] = DEFAULT_UID_MIN
+
+    if "auid" not in existing_properties:
+        result["auid"] = existing_properties.get("uid_min", DEFAULT_UID_MIN)
+
+    return result
+
+
+def get_product_yaml(product_yaml_path):
```
cipherboy, line 48: Two questions here:

1. Does this need to be tempered by the build YAML? If you look at e.g., `combine_ovals.py`:
    ```python3
    env_yaml = ssg.yaml.open_environment(args.build_config_yaml, args.product_yaml)
    ```
2. Would something like `load_product_yaml` be clearer? Especially if we add a `get_product_yaml_path(product)` -- we could make the product YAML path be inferred by the actual product. Just a thought.

build-scripts/cpe_generate.py
```diff
@@ -29,7 +30,12 @@ def parse_args():
         "definitions and the tests, states objects and variables it "
         "references, and then write them into a standalone OVAL CPE file, "
         "along with a synchronized CPE dictionary file.")
-    p.add_argument("product", help="Name of the product")
+    p.add_argument(
+        "--product-yaml",
+        help="YAML file with information about the product we are building. "
+        "e.g.: ~/scap-security-guide/rhel7/product.yml "
+        "needed for autodetection of profile root"
+    )
```
cipherboy, line 38: Not sure how much of a point there is, but for consistency it might be nice to pass in build config as well and use `open_environment` below and call it `env_yaml`. My 2c.

tests/ssg_test_suite/common.py
```diff
@@ -171,7 +173,9 @@ def _get_platform_cpes(platform):
             raise ValueError
         platform_cpes = set()
         for p in products:
-            p_cpes = ProductCPEs(p)
+            product_yaml_path = os.path.join(ssg_root, p, "product.yml")
```
cipherboy, line 176: Here's where I was thinking a `get_product_yaml(product)` form would be most useful. Then it would be `load_product_yaml(path)` elsewhere. My 2c.

cipherboy (approved):
I probably should've batched into a review but then I kept looking :-)

I didn't follow all the changes in `build_yaml.py` but otherwise looks good to me. Will let others do the final review and go-ahead, seems like there's still a build failure possibly and some conflicts?

ssg/products.py
yuumasato: I like the idea of the two functions.

yuumasato (commented):
Thank you for the review, folks.
I've just rebased to fix the conflict, and addressed the build issue with platform ocp-4.8

build-scripts/cpe_generate.py
cipherboy: Yeah I'm fine leaving this. Just thought I'd mention. :)

ssg/products.py
yuumasato: I just renamed the current function to `load_product_yaml`.

yuumasato:
@cipherboy @vojtapolasek This is ready to merge if your reviews are still positive.

I'll look into adding `get_product_yaml_path` in a subsequent PR.

PR #6661, opened 2021-03-05
implement support for multiple platforms connected with disjunction

vojtapolasek:
#### Description:

- modify build system so that it consumes platforms in two possible way
either
```
 platform: <single_platform>
```
or
```
platforms:
  - platform_1
  - platform_2
```
- internally, the list of platforms in compiled rules is called "platforms"
- platforms are added in the xccdf rule definition, implicitly connected with OR
- modify generation of Bash remediations so that rule-specific platforms are placed after inherited platforms and correctly connected with OR
- the same for Ansible remediations
- modify ssg test suite unit tests, in particular precompiled rules
- update documentation

#### Rationale:

This PR is part of an initiative which plans to implement support for CPE applicability language. This support will allow content authors to specify more granular definition of rule applicability.

PR #6760, opened 2021-03-26
fix build system to correctly parse platforms

vojtapolasek:
#### Description:
- read both "platform" and "platforms" while parsing yaml files
- then later work only with "platforms" but make sure that content of "platform" gets propagated into "platforms"

#### Rationale:

- previous introduction of multiple independent platforms in #6661  caused a breakage in platform applicability. This is the fix.

### DIFF
--- a/ssg/build_yaml.py
+++ b/ssg/build_yaml.py
@@ -117,7 +117,7 @@ class Profile(object):
         # self.platforms is used further in the build system
         # self.platform is merged into self.platforms
         # it is here for backward compatibility
-        self.platforms = None
+        self.platforms = set()
         self.platform = None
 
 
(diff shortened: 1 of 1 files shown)


yuumasato (changes requested):
What do you think about making `platforms` a `set()`?
I think it will simplify the code a bit.

vojtapolasek:
I pushed changes which initialize platforms as set. They are also stored as set within compiled rules... I am not sure if it breaks something, it looks like not at the first glance.

ssg/build_yaml.py
```diff
@@ -172,7 +172,7 @@ def dump_yaml(self, file_name, documentation_complete=True):
         if self.extends is not None:
             to_dump["extends"] = self.extends
 
-        if self.platforms is not None:
+        if len(self.platforms) > 0:
```
yuumasato, line 175: Nitpick: an empty platform evaluates to False:
`if self.platforms:` will work the same.

yuumasato (changes requested):

yuumasato (approved):

PR #6762, opened 2021-03-26
Change dhcp server package name to dhcp-server in rhel8

mildas:
#### Description:
There is no `dhcp` package in rhel8 repositories, the package with the dhcpd service is `dhcp-server`.
On rhel7 and sle15, no change is needed because the `dhcp` package with dhcpd is there.

Also add test scenarios for service_dhcpd_disabled to increase CIS profile test coverage.


yuumasato (approved):
Nice touch with the tests for multiple distros.

PR #6764, opened 2021-03-29
Optimize rules that traverse whole root filesystem by exclusion of /dev, /proc, and /sys where appropriate

evgenyz:
#### Rationale:

When these entries are excluded with the help of state entity using `<filter action="exclude" ...>` the data is being collected from the filesystem and then discarded, which is sub-optimal.

- Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1942208

### DIFF
--- a/linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands/oval/shared.xml
+++ b/linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands/oval/shared.xml
@@ -23,21 +23,17 @@
 
   <!-- First define OVAL entities that can be reused across tests below -->
   <unix:file_object id="object_system_privileged_commands" comment="system files with setuid or setgid permission set" version="1">
-    <unix:behaviors recurse="directories" recurse_direction="down" recurse_file_system="local" max_depth="-1" />
-    <unix:path operation="equals">/</unix:path>
+    <unix:behaviors recurse_file_system="local"/>
+    <!-- don't search /dev, /proc, /sys -->
+    <unix:path operation="pattern match">^\/(?!(dev|proc|sys)(\/|$)).*$</unix:path>
     <!-- [a-z]+ regex below is a workaround for OpenSCAP https://fedorahosted.org/openscap/ticket/457 bug -->
     <unix:filename operation="pattern match">[a-z]+</unix:filename>
     <filter action="include">state_setuid_or_setgid_set</filter>
-    <filter action="exclude">state_dev_proc_sys_dirs</filter>
   </unix:file_object>
 
   <unix:file_state id="state_setuid_or_setgid_set" version="1" operator="OR">
     <unix:suid datatype="boolean">true</unix:suid>
     <unix:sgid datatype="boolean">true</unix:sgid>
-  </unix:file_state>
-
-  <unix:file_state id="state_dev_proc_sys_dirs" version="1">
-    <unix:filepath operation="pattern match">^\/(dev|proc|sys)\/.*$</unix:filepath>
   </unix:file_state>
 
   <local_variable id="variable_all_privileged_commands" comment="All privileged commands" datatype="string" version="1">
--- a/linux_os/guide/system/permissions/files/file_permissions_unauthorized_sgid/oval/shared.xml
+++ b/linux_os/guide/system/permissions/files/file_permissions_unauthorized_sgid/oval/shared.xml
@@ -11,8 +11,9 @@
     </unix:file_test>
 
     <unix:file_object comment="files with sgid set which are not owned by any RPM package" id="obj_file_permissions_unauthorized_sgid_unowned" version="1">
-        <unix:behaviors recurse="directories" recurse_direction="down" max_depth="-1" recurse_file_system="local" />
-        <unix:path operation="equals">/</unix:path>
+        <unix:behaviors recurse_file_system="local" />
+        <!-- don't search /dev, /proc, /sys -->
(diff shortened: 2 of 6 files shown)


JAORMX:
/retest

```
     helpers.go:458: The Compliance Suite 'e2e-moderate' didn't get to DONE phase: Still waiting for the scans to be done 
```

I wonder what happened here... The scan timed out executing.

evgenyz:
Okay, this is a bit controversial. For the former implementation for my Fedora it took 1:52 to complete. The "optimized" one is 3:11. Command: `oscap oval eval --id=oval:ssg-audit_rules_privileged_commands:def:1 ssg-fedora-oval.xml`.

linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands/oval/shared.xml
```diff
@@ -23,23 +23,19 @@
 
   <!-- First define OVAL entities that can be reused across tests below -->
   <unix:file_object id="object_system_privileged_commands" comment="system files with setuid or setgid permission set" version="1">
-    <unix:behaviors recurse="directories" recurse_direction="down" recurse_file_system="local" max_depth="-1" />
-    <unix:path operation="equals">/</unix:path>
+    <unix:behaviors recurse_file_system="local"/>
+    <!-- don't search /dev, /proc, /sys -->
+    <unix:path operation="pattern match">^\/(?!(dev|proc|sys)(\/|$)).*$</unix:path>
     <!-- [a-z]+ regex below is a workaround for OpenSCAP https://fedorahosted.org/openscap/ticket/457 bug -->
```
yuumasato, line 29: Do we have a clue what https://fedorahosted.org/openscap/ticket/457 was about?
Can we do `filename` like this?

https://github.com/ComplianceAsCode/content/blob/da57b9e50cfd6630116f169355aa14a4f5f54adf/linux_os/guide/system/logging/rsyslog_sending_messages/rsyslog_remote_loghost/oval/shared.xml#L42

linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands/oval/shared.xml
```diff
@@ -23,23 +23,19 @@
 
   <!-- First define OVAL entities that can be reused across tests below -->
   <unix:file_object id="object_system_privileged_commands" comment="system files with setuid or setgid permission set" version="1">
-    <unix:behaviors recurse="directories" recurse_direction="down" recurse_file_system="local" max_depth="-1" />
-    <unix:path operation="equals">/</unix:path>
+    <unix:behaviors recurse_file_system="local"/>
+    <!-- don't search /dev, /proc, /sys -->
+    <unix:path operation="pattern match">^\/(?!(dev|proc|sys)(\/|$)).*$</unix:path>
```
yuumasato, line 28: I think the forward slash need to be escaped, we don't escape it in this rule, for example.
https://github.com/ComplianceAsCode/content/blob/da57b9e50cfd6630116f169355aa14a4f5f54adf/linux_os/guide/system/software/integrity/software-integrity/aide/aide_scan_notification/oval/shared.xml#L43

And the second group doesn't need to capture anything. (Not sure if this would speed up the regex matching)

```suggestion
    <unix:path operation="pattern match">^/(?!(dev|proc|sys)(?:/|$)).*$</unix:path>
```

JAORMX:
Based on the findings of this work, I'm removing the rule from RHCOS profiles: https://github.com/ComplianceAsCode/content/pull/6769 the concept of the rule itself doesn't seem to be feasible to check in RHCOS given that its main use-case is to run Kubernetes. We'll need to find another way to address that requirement. Seems that auditd is not suitable for that.

jan-cerny:
Hi @evgenyz , I think that this can also help us with https://bugzilla.redhat.com/show_bug.cgi?id=1932833 where the customer has a problem caused by file_permissions_ungroupowned.

linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands/oval/shared.xml
jan-cerny: I can't find it but I have found this comment referencing it: https://bugzilla.redhat.com/show_bug.cgi?id=1237197#c2

linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands/oval/shared.xml
matejak: I would like to have this as an OVAL macro, so it could be reused with less regex copy-pasting, and the intention would be totally obvious.

evgenyz: Err, well, after https://github.com/OpenSCAP/openscap/pull/1726 we can throw away `/sys` and `/proc`. And this approach is in general 10-20% slower (most likely because of the regexp handling during traversal).

jan-cerny:
@evgenyz Any updates?

Is the slow down because of the more complex regex or the partial match optimization in OpenSCAP fts module for the simpler regex? If it means a slow down, do we want to close this?

evgenyz:
Let's close it. `/dev/` is not that big and speed reduction is huge.

evgenyz:
https://github.com/OpenSCAP/openscap/pull/1726 takes care of `/proc` and `/sys`.

PR #6765, opened 2021-03-29
Update tests of accounts_tmout to work when overriding profiles

matusmarhefka:
When overriding test scenario profiles in rule mode
the variables metadata are not used. Therefore, this
commit updates 2 scenarios to work even in case when
a specific profile is forced in the rule mode using
the `--profile` option.

### DIFF
--- a/linux_os/guide/system/accounts/accounts-session/accounts_tmout/tests/multiline.fail.sh
+++ b/linux_os/guide/system/accounts/accounts-session/accounts_tmout/tests/multiline.fail.sh
@@ -1,9 +1,9 @@
 #!/bin/bash
 
-# variables = var_accounts_tmout=700
+# variables = var_accounts_tmout=900
 
 if grep -q "TMOUT" /etc/profile; then
-	sed -i "s/.*TMOUT.*/TMOUT=900; readonly TMOUT; export TMOUT/" /etc/profile
+	sed -i "s/.*TMOUT.*/TMOUT=950; readonly TMOUT; export TMOUT/" /etc/profile
 else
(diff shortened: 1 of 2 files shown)


ggbecker:
It sounds like test scenarios that use `variables` should be skipped when running in `profile` mode. But I guess that would decrease considerably the amount of eligible tests.

matusmarhefka:
> It sounds like test scenarios that use `variables` should be skipped when running in `profile` mode. But I guess that would decrease considerably the amount of eligible tests.

Yes, that would make sense in the `rule` mode when profile is overridden to a specific profile or if a test scenario specifies profile(s) in `profiles` metadata.

The amount of eligible tests would decrease in such case, but it might be better than having such tests failing.

yuumasato:
So it looks like test scenarios with variables should use the variable's default value for most of the test cases, and deviate only when the specifically testing a change in the value of the variable.

I don't have anything against merging this PR.
How about you @ggbecker ?

ggbecker:
> So it looks like test scenarios with variables should use the variable's default value for most of the test cases, and deviate only when the specifically testing a change in the value of the variable.
> 
> I don't have anything against merging this PR.
> How about you @ggbecker ?

Me neither. But I think we should register this situation with an issue.

ggbecker (approved):
LGTM

ggbecker:
> > So it looks like test scenarios with variables should use the variable's default value for most of the test cases, and deviate only when the specifically testing a change in the value of the variable.
> > I don't have anything against merging this PR.
> > How about you @ggbecker ?
> 
> Me neither. But I think we should register this situation with an issue.

I've opened #6781

matusmarhefka:
Thank you @ggbecker for creating an issue.

PR #6769, opened 2021-03-29
RHCOS4: Remove instances of audit_rules_privileged_commands

JAORMX:
This rule traverses the whole file system in order to find instances of
executables that have setuid and setgid enabled. This is not scalable in
hosts running container orchestration engines since we'll have assorted
filesystems of the containers which might differ at a given point in
time.

Therefore, it's no use trying to audit this. We'll have to find another
way to address this.


JAORMX:
/test e2e-aws-rhcos4-e8

CI pod failed sending logs.. the test actually passed

PR #6771, opened 2021-03-30
RHCOS4: Explicitly use OSPP profile for rules covered by it

JAORMX:
This ensures that we don't try to remediate individually the audit rules
for syscalls that are already covered by the OSPP audit rule profile.
This way we could address the explicit ordering requirements


linux_os/guide/system/auditing/auditd_configure_rules/audit_file_modification/audit_rules_unsuccessful_file_modification_creat/kubernetes/shared.yml
```diff
@@ -0,0 +1,7 @@
+# platform = multi_platform_ocp,multi_platform_rhcos
+# reboot = true
+# strategy = disable
+# complexity = low
+# disruption = medium
+
+{{{ kubernetes_machineconfig_ospp_audit_rules() }}}
```
yuumasato, line 7: Should we explicitly say in 
https://github.com/ComplianceAsCode/content/blob/2a632ada26e2365700032fcad3dca37ed208e637/linux_os/guide/system/auditing/auditd_configure_rules/audit_file_modification/audit_rules_unsuccessful_file_modification_creat/rule.yml that `kubernetes` backend is off?

JAORMX: we could... but as far as I can tell, this takes precedence.

yuumasato: Yes, it does.
Was just concerned with possibility of content authors being confused.

yuumasato: I leave it up to you.

JAORMX: I'll leave it as-is.

PR #6772, opened 2021-03-30
make build system handle inherited platforms also for groups

vojtapolasek:
#### Description:

- create a new attribute for group called inherited_platforms
- this is a set which is filled up with platforms of parent groups during the parsing of directory structure
- the contents of this set propagate to inherited platforms of groups

#### Rationale:

Previous implementation propagated group platforms to rules only from the closest group in hierarchy. This fixes it. you can see a good example with rule file_groupowner_efi_grub2_cfg. Before this PR, the rule inherited only platform "uefi". After this PR, it inherits also "grub2". This manifests in the rule remediation. Before this PR the remediation was actually not checking if grub2 package is installed.

### DIFF
--- a/ssg/build_yaml.py
+++ b/ssg/build_yaml.py
@@ -820,9 +820,6 @@ class Benchmark(object):
 class Group(object):
     """Represents XCCDF Group
     """
-    ATTRIBUTES_TO_PASS_ON = (
-        "platforms",
-    )
 
     def __init__(self, id_):
         self.id_ = id_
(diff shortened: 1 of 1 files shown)


ssg/build_yaml.py
```diff
@@ -1441,6 +1442,8 @@ def _load_group_process_and_recurse(self, guide_directory):
         if self.loaded_group:
             if self.parent_group:
                 self.parent_group.add_group(self.loaded_group)
+                if hasattr(self.parent_group, "platforms") and self.parent_group.platforms:
+                    self.loaded_group.inherited_platforms.update(self.parent_group.platforms)
```
yuumasato, line 1446: The `add_group()` method from `Group` class calls a method named `_pass_our_properties_on_to()`, which passes down to its child group or rule its `platform`.

ssg/build_yaml.py
```diff
@@ -1499,8 +1502,12 @@ def _process_rules(self):
             self.loaded_group.add_rule(rule)
 
             if self.loaded_group.platforms:
-                rule.inherited_platforms += self.loaded_group.platforms
+                rule.inherited_platforms.update(self.loaded_group.platforms)
 
+            if self.loaded_group.inherited_platforms:
+                rule.inherited_platforms.update(self.loaded_group.inherited_platforms)
+
+
```
yuumasato, line 1510: Same thing here, the two `ifs` are handling the platforms a rule inherits from the group it gets inserted into.
It seems to me that these two conditionals should be moved to the Group's `add_rule()` method.

yuumasato (changes requested):
I think your changes are good, just made comments about repositioning them.

With inheritance done explicitly, we probably have no use for `_pass_our_properties_on_to()`, as the `platforms` and `inherited_platforms` of the parent become the `inherited_platforms` of the child.

yuumasato (approved):
Thanks for the fix!

PR #6779, opened 2021-03-30
pipefail not really compatible with grep

brett060102:
The pipefail option traps on all non 0 reurn codes, but grep exits
with 1 if the pattern is not found. So, just disable that ansible-lint
check if the shell snippet is using grep.

#### Description:

- Disable ansible-lint pipefail test, if going grep
#### Rationale:

- The pipefail option traps on all non 0 reurn codes, but grep exits with 1 if the pattern is not found. So, just disable that ansible-lint check if the shell snippet is using grep.

- Without this the ansible play will hit pipefail trap if pattern is not found, which pretty much breaks, if not found, add cases.

### DIFF
--- a/linux_os/guide/services/obsolete/r_services/no_host_based_files/ansible/shared.yml
+++ b/linux_os/guide/services/obsolete/r_services/no_host_based_files/ansible/shared.yml
@@ -5,8 +5,9 @@
 # disruption = low
 - block:
     - name: "Find local mount points"
-      shell: | #noqa 306 we don't care about grep failure in this case
-          df --local | awk '{print $6}' | grep -v Mounted | grep -v '^/dev'
+      shell: |
+          set -o pipefail
+          { df --local | awk '{print $6}' | grep -v Mounted | grep -v '^/dev' || true; }
       register: local_mount_points
 
     - name: "Detect the shosts.equiv files on the system"
--- a/linux_os/guide/services/obsolete/r_services/no_host_based_files/ansible/shared.yml
+++ b/linux_os/guide/services/obsolete/r_services/no_host_based_files/ansible/shared.yml
@@ -5,8 +5,7 @@
 # disruption = low
 - block:
     - name: "Find local mount points"
-      shell: |
-          set -o pipefail
+      shell: | #noqa 306 we don't care about grep failure in this case
           df --local | awk '{print $6}' | grep -v Mounted | grep -v '^/dev'
       register: local_mount_points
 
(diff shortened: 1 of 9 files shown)


brett060102:
We did find and fix this in our development branch, but I dropped that part of the fix when doing https://github.com/ComplianceAsCode/content/pull/6730

linux_os/guide/services/obsolete/r_services/no_host_based_files/ansible/shared.yml
```diff
@@ -5,8 +5,7 @@
 # disruption = low
 - block:
     - name: "Find local mount points"
-      shell: |
-          set -o pipefail
+      shell: | #noqa 306 we don't care about grep failure in this case
```
yuumasato, line 8: Unfortunately the yaml libraries strip out comments when processing them.
The exception for ansible-lint should be added here: https://github.com/ComplianceAsCode/content/blob/master/tests/ansible-lint_config.yml

vojtapolasek:
Here is an alternative approach which does not require removal of ansible-lint checks. If it works.
https://stackoverflow.com/questions/6550484/prevent-grep-returning-an-error-when-input-doesnt-match

shared/templates/pam_options/ansible.template
```diff
@@ -24,8 +24,8 @@
 
 - name: Check to see if '{{{ MODULE }}}' module is configured in '{{{ PATH }}}'
   shell: |
-    set -o pipefail
-    grep -E '^\s*{{{ TYPE }}}\s+\S+\s+{{{ MODULE }}}' {{{ PATH }}} | cat
+     set -o pipefail
+     { grep -E '^\s*{{{ TYPE }}}\s+\S+\s+{{{ MODULE }}}' {{{ PATH }}} || true; }
```
brett060102, line 28: not sure why this one can't be done as:
shell: grep -E '^\s*{{{ TYPE }}}\s+\S+\s+{{{ MODULE }}}' {{{ PATH }}} || true

But ansible-lint gave a pipe error when I tried that.

brett060102:
@vojtapolasek Think this should be OK, now. In pam_options/ansible.template and set_password_hashing_min_rounds_logindefs/ansible/shared.yml I tried usimh a single line command since there is no piping used there, but ansible-lint still objected. This code now makes more sense than the original. The "| cat" was just being used to eat the "1" return code from grep, the execute "true" on failure is really a better way to do that. Thanks for your time on this.

brett060102:
@vojtapolasek should be done messing with this now.

brett060102:
@vojtapolasek How does this one look now?

vojtapolasek:
I am checking it, give me a bit more time, there are other warnings and I need to filter them out.

vojtapolasek:
I think it looks good. Thank you for fixing this.

brett060102:
@vojtapolasek Thank you.

PR #6783, opened 2021-03-31
service_rsyncd_disabled - update package name to rsync-daemon

mildas:
#### Description:
Some products have the `rsyncd` service in the `rsync-daemon` package, not in `rsync`.

- rhel8 and Fedora - I have verified that these two products have `rsyncd` in the `rsync-daemon` package,
- ol8 and rhv4 - this is guess (they should be based on rhel8 so I'd say it should be similar as rhel8),
- sle15 - I didn't find any `rsync-daemon` package for sle15, @brett060102 can you verify that the `rsyncd` service is in the `rsync` package?

The `rhcos4` prodtype was removed because `rsyncd` is not available in RHCOS4 (neither `rsync` nor `rsync-daemon` is available there). And this rule is not present in any RHCOS4 profile.

### DIFF
--- a/linux_os/guide/services/obsolete/service_rsyncd_disabled/rule.yml
+++ b/linux_os/guide/services/obsolete/service_rsyncd_disabled/rule.yml
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: rhel7,ol7,rhel8,ol8,fedora,rhv4,rhcos4,sle15
+prodtype: rhel7,ol7,rhel8,ol8,fedora,rhv4,sle15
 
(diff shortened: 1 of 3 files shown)


JAORMX (approved):

vojtapolasek:
@freddieRv  could you please confirm the guess for ol8?

brett060102:
@mildas for SLES rsyncd is in rsync-*.x86_64

carlosmmatos:
Don't merge until we've confirmed OL8.

freddieRv:
@vojtapolasek guess is correct, `rsyncd` service for OL8 is in `rsync-daemon` package

PR #6792, opened 2021-04-05
Allow product-version-agnostic checks/remediations

cipherboy:
#### Description:

This allows product-version-agnostic checks/remediations to be added to
CaC and used ahead of shared content, while still allowing
version-specific content to be favored ahead of both.

That is, given the following:

 - `ovals/shared.xml`
 - `ovals/ubuntu.xml`
 - `ovals/ubuntu2004.xml`

The priority order returned from:

```python3
    get_rule_dir_ovals(..., product="ubuntu")
```

would be:

  1. `ubuntu2004.xml`
  2. `ubuntu.xml`
  3. `shared.xml`

#### Rationale:

This is useful when OVAL/... content differs between products (e.g.,
RHEL, Ubuntu, ...) but still applies to other versions of the same
product (e.g., 18.04 and 20.04). Duplication thus occurs only at a
product level (if say, Debian chose to reuse Ubuntu content or
visa-versa).

### DIFF
--- a/ssg/build_remediations.py
+++ b/ssg/build_remediations.py
@@ -655,18 +655,42 @@ def get_rule_dir_remediations(dir_path, remediation_type, product=None):
     if not has_remediations_dir:
         return []
 
-    results = []
+    # Two categories of results: those for a product and those that are
+    # shared to multiple products. Within common results, there's two types:
+    # those shared to multiple versions of the same type (added up front) and
+    # those shared across multiple product types (e.g., RHEL and Ubuntu).
+    product_results = []
+    common_results = []
     for remediation_file in sorted(os.listdir(remediations_dir)):
         file_name, file_ext = os.path.splitext(remediation_file)
         remediation_path = os.path.join(remediations_dir, remediation_file)
 
         if file_ext == ext and rules.applies_to_product(file_name, product):
+            # rules.applies_to_product ensures we only have three entries:
+            # 1. shared
+            # 2. <product>
+            # 3. <product><version>
+            #
+            # Note that the product variable holds <product><version>.
             if file_name == 'shared':
-                results.append(remediation_path)
+                # Shared are the lowest priority items, add them to the end
+                # of the common results.
+                common_results.append(remediation_path)
+            elif file_name != product:
+                # Here, the filename is a subset of the product, but isn't
+                # the full product. Product here is both the product name
+                # (e.g., ubuntu) and its version (2004). Filename could be
+                # either "ubuntu" or "ubuntu2004" so we want this branch
+                # to trigger when it is the former, not the latter. It is
+                # the highest priority of common results, so insert it
+                # before any shared ones.
+                common_results.insert(0, remediation_path)
             else:
-                results.insert(0, remediation_path)
+                # Finally, this must be product-specific result.
+                product_results.append(remediation_path)
 
-    return results
+    # Combine the two sets in priority order.
+    return product_results + common_results
 
 
 def expand_xccdf_subs(fix, remediation_type, remediation_functions):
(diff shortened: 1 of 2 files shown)


ssg/build_remediations.py
```diff
@@ -656,17 +656,20 @@ def get_rule_dir_remediations(dir_path, remediation_type, product=None):
...
+    common_results = []
     for remediation_file in sorted(os.listdir(remediations_dir)):
         file_name, file_ext = os.path.splitext(remediation_file)
         remediation_path = os.path.join(remediations_dir, remediation_file)
 
         if file_ext == ext and rules.applies_to_product(file_name, product):
             if file_name == 'shared':
-                results.append(remediation_path)
+                common_results.append(remediation_path)
+            elif file_name != product:
```
yuumasato, line 667: Nitpick: I think it is a bit confusing that it checks that `file_name != product`.
As `file_name` passed through `applies_to_product()` it is clearly a sub-string of `product`, i.e. "ubuntu", when product is "ubuntu2004".

I'd add a comment here, or flip the order, `elif file_name == product:` and the `else` handles any product sub
 -string

ssg/build_remediations.py
```diff
@@ -656,17 +656,20 @@ def get_rule_dir_remediations(dir_path, remediation_type, product=None):
...
+                common_results.insert(0, remediation_path)
             else:
-                results.insert(0, remediation_path)
+                results.append(remediation_path)
```
yuumasato, line 670: Another nitpick, `results` will contain the `remediation_path` for the very product specific remediation, right? i.e. "ubuntu2004".

In this case `results` doesn't strictly need to be a list, we don't expect to have multiple `bash/2004.sh`.

yuumasato (commented):
Looks good, I just have two nitpicks that apply to both remediation and oval handling.

ssg/build_remediations.py
cipherboy: Sure. :) Yeah, I grouped it this way since grouping `common_results` modifications ahead of `results` modifications made most sense to me but I'll clarify with a comment.

ssg/build_remediations.py
cipherboy: Yeah I wasn't sure what I wanted to do here. In the future, if there were say, SCE content, you could envision having multiple entries for the same product (say, Perl, Python, and Bash content for `ubuntu2004`). With fancy CPE checks &c, you could maybe determine that the system (at runtime) doesn't have Perl or Python but does have Bash... so you'd want to take the first of the unique extension entries out of there. Maybe the SCE content would be in separate directories, I don't know. 

Since we're ultimately coercing the results to a list (and not performing the selection here), leaving `results` as a list allows for this possibility and allows same-typed addition later.

Otherwise, we'd use an assignment here but just coerce it to a list for the return eventually anyways (`[result] + common_results` or `common_results.insert(0, result)`).

I figured it was a wash. But happy to do whatever. 

Preferences?

cipherboy:
@yuumasato Addressed the first one, let me know if you want me to do the second change. Thanks!

ssg/build_remediations.py
yuumasato: You kinda overdid it,but looks ok, :)

ssg/build_remediations.py
yuumasato: But SCE is about checking, right? Not remediating.

> Maybe the SCE content would be in separate directories, I don't know.

I think that would be the case....

> Since we're ultimately coercing the results to a list (and not performing the selection here),

Yes, but `results` aims to contain the top pick. I just found it strange that we would have multiple matches for top pick.
It was just a nitpick anyway.

yuumasato (approved):

PR #6793, opened 2021-04-05
Updates to support ansible_playbook_to_role utility

carlosmmatos:
#### Description:

Numerous updates were made to enhance the functionality
and to fix bugs:
* Updated GH actions to run on push events for verification. Also added a new tag skip.
* Updated galaxy meta template. Includes additions of variables
to be used as substitutions to support automating the creation
of that file, and to prevent manual additions as was handled in
the past.
* Added the anssi profiles to the whitelist (have not done any testing
on these roles yet)
* Updates/bug fixes to support meta templating. Fixed issue with
meta and readme files being overwritten with the wrong output.
* Updated syntax for python3 support. No need to use python2 for
this IMO.

#### Rationale:

Need to continue to provide a mechanism of translating our Ansible playbooks into roles that can be ingested into Ansible Galaxy.

### DIFF
--- a/.github/workflows/validate_ansible_roles.yaml
+++ b/.github/workflows/validate_ansible_roles.yaml
@@ -1,7 +1,7 @@
 name: Validate Ansible Roles
 on:
   push:
-    branches: [ master ]
+    branches: [ '*' ]
   pull_request:
     branches: [ master ]
 jobs:
@@ -12,15 +12,19 @@ jobs:
       - name: Install Deps
         uses: mstksg/get-package@master
         with:
-                apt-get: cmake ninja-build libopenscap8 libxml2-utils expat xsltproc python3-jinja2 python3-yaml ansible-lint python3-github
+                apt-get: cmake ninja-build libopenscap8 libxml2-utils expat xsltproc python3-jinja2 python3-yaml python3-github
+      - name: Install upstream ansible-lint
+        run: |
+          python -m pip install --upgrade pip
+          pip install ansible-lint ansible
       - name: Checkout
         uses: actions/checkout@v1
       - name: Build
         run: |
-          ./build_product rhel7 rhel8 rhv4
+          ./build_product rhel7 rhel8
       - name: Build Ansible Roles
         run: |
           PYTHONPATH=. python3 utils/ansible_playbook_to_role.py --build-playbooks-dir ./build/ansible/ --dry-run ./build/ansible_roles
       - name: Lint Ansible Roles
         run: |
-          ansible-lint -x 204 ./build/ansible_roles/*
+          ansible-lint -x 204 -x experimental ./build/ansible_roles/*
--- a/utils/ansible_playbook_to_role.py
+++ b/utils/ansible_playbook_to_role.py
@@ -57,11 +57,19 @@ yaml.add_constructor(_mapping_tag, dict_constructor)
 PRODUCT_WHITELIST = set([
     "rhel7",
     "rhel8",
-    "rhv4",
 ])
 
 PROFILE_WHITELIST = set([
+    "anssi_nt28_enhanced",
+    "anssi_nt28_high",
+    "anssi_nt28_intermediary",
+    "anssi_nt28_minimal",
+    "anssi_bp28_enhanced",
+    "anssi_bp28_high",
+    "anssi_bp28_intermediary",
+    "anssi_bp28_minimal",
     "C2S",
+    "cis",
     "cjis",
     "hipaa",
     "cui",
@@ -102,7 +110,7 @@ def create_empty_repositories(github_new_repos, github_org):
 
 def clone_and_init_repository(parent_dir, organization, repo):
     os.system(
-        "git clone https://github.com/%s/%s" % (organization, repo))
+        "git clone git@github.com:%s/%s" % (organization, repo))
     os.system("ansible-galaxy init " + repo + " --force")
     os.chdir(repo)
     try:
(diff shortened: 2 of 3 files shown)


carlosmmatos:
OHmy lord... what is happening to OpenShift CI? Not that I care about it for this PR since my changes should have no impact on this but clearly something is messed up.

utils/ansible_playbook_to_role.py
```diff
@@ -61,7 +61,12 @@ def dict_constructor(loader, node):
 ])
 
 PROFILE_WHITELIST = set([
+    "anssi_nt28_enhanced",
+    "anssi_nt28_high",
+    "anssi_nt28_intermediary",
+    "anssi_nt28_minimal",
```
yuumasato, line 67: The ANSSI profiles for RHEL8 have a slightly different ID, with `bp28` instead of `nt28`.
And note that High is still in draft state.

utils/ansible_playbook_to_role.py
```diff
@@ -252,6 +270,36 @@ def _description(self):
...
+            "openscap",
+            "ssg",
+            "scap",
+            "security",
+            "compliance",
+            "complianceascode",
+            "redhatofficial",
+            "redhat",
+        ]
+        prod, prof = self.name.split("_", 1)
```
yuumasato, line 289: Just a comment that it may be worthy storing `product` and `profile` as a property, or something, of the class for easy access.

utils/ansible_playbook_to_role.py
```diff
@@ -229,6 +234,19 @@ def _raw_playbook(self):
...
+    @memoize
+    def platform_version(self):
+        prod = self.name.split("_")
+        platform = prod[0]
+        # Check to see if this is RHEL product
+        if platform in PRODUCT_WHITELIST:
+            # For RHEL, we can get what version
+            if 'rhel' in platform:
+                return platform[len(platform)-1]
+            return "7\n    - 8"
```
yuumasato, line 247: I think you don't plan on updating RHEV4 roles for the moment, but `PRODUCT_WHITELIST` contains `rhv4`, and this may lead to platform version 7 and 8 for RHEV.
Better disarm this trap now, :)

.github/workflows/validate_ansible_roles.yaml
```diff
@@ -1,7 +1,7 @@
 name: Validate Ansible Roles
 on:
   push:
-    branches: [ master ]
+    branches: [ '*' ]
```
yuumasato, line 4: ` > Updated GH actions to run on push events for verification.`

I'm not sure what is the intention here. The other branch we have is `stable`
Are you looking to cover releases as well? Maybe `release` the activity type is better suited.

yuumasato (commented):
I haven't tried to run the script, not sure if there is a good dry-run mode...
@carlosmmatos Did you run it somehow?

utils/ansible_playbook_to_role.py
carlosmmatos: got it.. will update bp28 and understand that high is in draft.. as mentioned prior, this was just to add some new profiles that may benefit from this in the near future. Currently I plan on supporting the STIG profiles, and will add jobs in our GH workflow to only build and produce roles for RHEL STIG.

utils/ansible_playbook_to_role.py
carlosmmatos: I agree

.github/workflows/validate_ansible_roles.yaml
carlosmmatos: I was trying to make sure that when someone is working on this in their forked repo, they get some validation. Obviously, this was only working if you were pushing to the master branch, which no one does unless they're merging. We can chat more about this to see what makes sense.

yuumasato: Thanks for clarification.
I don't know what would be the best way to achieve this.
I guess this looks fine to me.... Anyone has other thoughts?

utils/ansible_playbook_to_role.py
```diff
@@ -229,6 +247,19 @@ def _raw_playbook(self):
         with io.open(self._local_playbook_filename, 'r', encoding="utf-8") as f:
             return f.read()
 
+    @property
+    @memoize
+    def platform_version(self):
+        prod = self.name.split("_")
```
yuumasato, line 253: You can just use `self.product` here.

utils/ansible_playbook_to_role.py
```diff
@@ -157,7 +165,17 @@ def __init__(self, local_playbook_filename):
...
+
+    @property
+    @memoize
+    def product(self):
+        return self.name.split("_")[0]
+
+    @property
+    @memoize
+    def profile(self):
+        return self.name.partition("_")[-1]
```
yuumasato, line 178: Is there a reason to use different functions? (split vs partition)

carlosmmatos: example: rhe7_anssi_nt28_enhanced
```python
>>> name = "rhe7_anssi_nt28_enhanced"
>>> name.split("_")
['rhe7', 'anssi', 'nt28', 'enhanced']
>>> name.partition("_")
('rhe7', '_', 'anssi_nt28_enhanced')
```

partition gives me the actual profile name as the last index, as opposed to chopping together the profile with split(). At least that was my goal with profile, was to get <prod>_[<profile>] as is passed in.

utils/ansible_playbook_to_role.py
carlosmmatos: I agree, let me update this func

utils/ansible_playbook_to_role.py
carlosmmatos: @yuumasato my bad, I brainfarted this... so yes, I can use split with the positional arg of 1, and that will give me the second index as my result:
```python
>>> name.split("_", 1)
['rhe7', 'anssi_nt28_enhanced']
```
This way we keep split as the common method of choice...

yuumasato: Sorry, I should have given an example use of split when inquiring about it.

yuumasato (approved):
LGTM.

PR #6794, opened 2021-04-05
Add Ubuntu as a known platform for ssg_test_suite

cipherboy:
#### Description:

In order to run on Ubuntu (as a host and a guest) I ended up needing two
things:

 1. Add Ubuntu to the package installation capabilities,
 2. Fix libvirt<->xml interactions: `tostring` rendered as `bytes` not as a
    `str` object but the libvirt API expected a `str`.

I also needed to add `CPE_NAME` to the Ubuntu guest VM as our
`/etc/os-release` doesn't contain it by default.

### DIFF
--- a/tests/ssg_test_suite/virt.py
+++ b/tests/ssg_test_suite/virt.py
@@ -139,7 +139,7 @@ def determine_ip(domain):
         domain_xml = ET.fromstring(domain.XMLDesc())
         for guest_agent_node in domain_xml.iter('channel'):
             if guest_agent_node.attrib['type'] == 'unix':
-                guest_agent_xml_string = ET.tostring(guest_agent_node)
+                guest_agent_xml_string = ET.tostring(guest_agent_node, encoding='unicode')
                 break
         if guest_agent_xml_string:
             domain.detachDevice(guest_agent_xml_string)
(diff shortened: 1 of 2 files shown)


cipherboy:
I'm curious if anyone else ran into the `bytes`/`str` difference and what the cause of it ends up being.

tests/ssg_test_suite/virt.py
```diff
@@ -142,6 +142,8 @@ def determine_ip(domain):
                 guest_agent_xml_string = ET.tostring(guest_agent_node)
                 break
         if guest_agent_xml_string:
+            if isinstance(guest_agent_xml_string, bytes):
+                guest_agent_xml_string = guest_agent_xml_string.decode('utf-8')
```
jan-cerny, line 146: I think you can use `encoding` argument in `tostring` that  creates this variable above. https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.tostring

cipherboy: Ohhhhhhhhhhhhh I see now. From the doc:

> If encoding is "unicode", a string is returned. Otherwise a bytestring is returned.

I used `encoding='utf-8'` so the result was a bytestring NOT a string. d'oh!

Thanks @jan-cerny!

cipherboy:
Thanks @jan-cerny! Updated :)

jan-cerny:
Thanks!

Issue #6781, opened 2021-03-31
Certain test scenarios have to be skipped by SSGTS when selecting a profile


ggbecker:
In some test scenarios we assume that a certain value is wrong (to provoke a fail) and it can happen that this assumed value is actually selected by a profile which can make the test have unexpected results when running SSGTS with a selected profile.

These tests should either be skipped or treated in a different way. Further investigation is needed.

_Originally posted by @ggbecker in https://github.com/ComplianceAsCode/content/issues/6765#issuecomment-810042481_

cipherboy:
Two thoughts:

Perhaps add a mechanism for selecting a test case only under certain scenarios. Either in dotted form in the script name (like `test_name.profile_name.pass/fail.sh`) or as part of a header section in the test (similar to platform, perhaps do `profile@platform` or similar). By default applies to everything, and could have negations perhaps (`!profile@platform` or `!platform`).

Perhaps we need to add jinja templating into the test cases. This might require we add a no-op status code but we could evaluate each test case in the environment of the rule+profile+product it is being run under (special casing `(all)` to the default), separately for each execution.

 This would allow us to:
   - Pull in the actual variable value we care about,
   - No-op execution when the profile/.. doesn't match one we care about (if we execute it, the former idea would allow us to skip these tests entirely),
   - Handle any (small) differences in product more easily than creating a new rule -- potentially handling different package names too.

One follow up question: do templates have test cases? If not yet, perhaps this would be a way to add them in a product-agnostic manner -- then we can validate that the prodtypes we _think_ they apply to, they actually will. Just a thought :)

cipherboy:
Note that Jinja support for templates is going to be added in #7210 and support for tests under shared/templates is added under #7211. 

Currently Jinja support for templates doesn't do any detection of profile and variable values. But the infrastructure is now in place to pass through a platform specifier so passing through a profile as well wouldn't be too much work (though, some care as to how variables get loaded and exposed to Jinja templates at build time will need to be taken).

Mab879:
This should be fixed now.

PR #6795, opened 2021-04-05
Handle profile selection correctly in SSG Test Suite

cipherboy:
#### Description:

    Validate scripts against combined's target profile
    
    When combined is executed, a specific list of profiles are supplied.
    Because these scripts are executed for each profile (within rule.py), we
    end up executing these scripts against each profile in the benchmark.
    However, combined already takes a profile argument and thus these
    scripts should be limited to the profiles the user already selected.
    
---

    Handle OSCAP_PROFILE_ALL_ID in rule testing
    
    When testing a script which applies to multiple profiles, we want to
    execute OSCAP against each profile in the DS, not once. Previously, we'd
    use '(all)' to indicate all profiles (e.g., in combined.py's
    _modify_parameters), but check it against ALL. IMO this is incorrect
    behavior and we should instead be adding all DS profiles individually.
    
#### Rationale:

I encountered both of these issues while attempting to run `combined` on the WIP Ubuntu profile. Note that Ubuntu currently ships with OpenSCAP version 1.2 (so perhaps `(all)` pseudo-profile wasn't added until 1.3)?

### DIFF
--- a/tests/ssg_test_suite/combined.py
+++ b/tests/ssg_test_suite/combined.py
@@ -56,10 +56,10 @@ class CombinedChecker(rule.RuleChecker):
         # be tested using the self.profile and we return empty profiles
         # metadata.
         if not params["profiles"]:
-            params["profiles"].append(rule.OSCAP_PROFILE_ALL_ID)
+            params["profiles"].extend(self.profile)
             logging.debug(
                 "Added the {0} profile to the list of available profiles for {1}"
-                .format(rule.OSCAP_PROFILE_ALL_ID, script))
+                .format(self.profile, script))
         else:
             params['profiles'] = [item for item in params['profiles'] if re.search(self.profile, item)]
         return params
--- a/tests/ssg_test_suite/rule.py
+++ b/tests/ssg_test_suite/rule.py
@@ -39,7 +39,7 @@ def get_viable_profiles(selected_profiles, datastream, benchmark, script=None):
     all_profiles.append(OSCAP_PROFILE_ALL_ID)
 
     for ds_profile in all_profiles:
-        if 'ALL' in selected_profiles:
+        if 'ALL' in selected_profiles or OSCAP_PROFILE_ALL_ID in selected_profiles:
             valid_profiles += [ds_profile]
             continue
         for sel_profile in selected_profiles:


ggbecker:
According to https://github.com/OpenSCAP/openscap/pull/1193 the "(all)" profile was supposed to be backported to openscap 1.2.18, but this release never happened. RHEL7 has a downstream patch for this feature though.

cipherboy:
@ggbecker Thanks for the pointers! Yeah, our openscap is still at 1.2.17 and until 22.04, openscap 1.3 won't hit a LTS release. 20.04 will be supported for quite a while and that's what I'm running on for a while...

I can look into getting a bugfix in openscap but I'm not yet sure who owns the package. :)

I still think that the intent of the user, when run against a specific list of profiles, is to use those profiles and not all profiles, but I'll leave it up to y'all :)

ggbecker:
> @ggbecker Thanks for the pointers! Yeah, our openscap is still at 1.2.17 and until 22.04, openscap 1.3 won't hit a LTS release. 20.04 will be supported for quite a while and that's what I'm running on for a while...
> 
> I can look into getting a bugfix in openscap but I'm not yet sure who owns the package. :)
> 
> I still think that the intent of the user, when run against a specific list of profiles (edit: in the case where there is a variable that takes different values based on profile), is to use those profiles and not all profiles, but I'll leave it up to y'all :)

I'm trying to understand the intention behind this pull request, but first I want to confirm that you understand that `(all)` profile, is a single virtual profile that OpenSCAP understands and interprets as a profile that selects all the rules in the benchmark and uses default values for XCCDF variables. Do you have the same understanding?

cipherboy:
Yep :)

My intention lies here:

> and uses default values for XCCDF variables

If I run, say:

```bash
$ ./test_suite.py combined .... cis_level1_server
```

And I add a profile definition say:

```
sections:
 - var_ssh_access_timeout=10
```

and in the variable I have:

```
values:
  default: "0"
  10: "10"
```

Then the `(all)` selection will cause the rule to validate and remediate with a value of 0, correct? Even though, AFAICT, from the intent of the CLI invocation, the caller expected a value of 10 right (because that's what the profile actually specifies)?

If the user executes:


```bash
$ ./test_suite.py combined .... '(all)'
```

yes, they would expect the default to be used.

My 2c., but I think the behavior with a single profile listed is wrong and I'm trying to correct. Maybe we need to make `(all)` an explicitly understood argument with separate behavior?

(And I guess user here is a SSG developer :D)

ggbecker:
> Yep :)
> 
> My intention lies here:
> 
> > and uses default values for XCCDF variables
> 
> If I run, say:
> 
> ```shell
> $ ./test_suite.py combined .... cis_level1_server
> ```
> 
> And I add a profile definition say:
> 
> ```
> sections:
>  - var_ssh_access_timeout=10
> ```
> 
> and in the variable I have:
> 
> ```
> values:
>   default: "0"
>   10: "10"
> ```
> 
> Then the `(all)` selection (edit: as is the current behavior of the test suite, prior to my PR AFAICT) will cause the rule to validate and remediate with a value of 0, correct? Even though, AFAICT, from the intent of the CLI invocation, the caller expected a value of 10 right (because that's what the profile actually specifies)?
> 
> If the user executes:
> 
> ```shell
> $ ./test_suite.py combined .... '(all)'
> ```
> 
> yes, they would expect the default to be used.
> 
> My 2c., but I think the behavior with a single profile listed is wrong and I'm trying to correct. Maybe we need to make `(all)` an explicitly understood argument with separate behavior?
> 
> (And I guess user here is a SSG developer :D)
> 
> Edit: I guess this causes bigger problems if we have multiple profiles with conflicting value selections. The test suite (`/tests` subdir of a rule) doesn't know how to handle different values for different profiles, right? Does it have access to the templated variable value? Maybe I'm missing something obvious though :)

This problem with different values for different profiles is still open, it even manifested a few days ago in https://github.com/ComplianceAsCode/content/pull/6765 where a test scenario assumed a wrong value but that value was actually correct from the selected profile's perspective. I believe that the combined mode is somehow broken due to these variables' uncertainties.

To be honest, there are multiple corner cases with these combinations of options and test scenarios that are hard to remember everything. I need to run a few tests before I really grasp what the problem is. Please bear with me for a couple of days more ;)

cipherboy:
Of course :)

> This problem with different values for different profiles is still open

Got a ticket?

I'm wondering if perhaps we need two things: another dotted section for profile (`test_name.profile_name.pass/fail.sh`) and a make all these tests jinja-macroized and pull in the relevant rules from the profile.

matusmarhefka:
> Of course :)
> 
> > This problem with different values for different profiles is still open
> 
> ~Got a ticket?~ Edit: I see #6781.
> 
> I'm wondering if perhaps we need two things: another dotted section for profile (`test_name.profile_name.pass/fail.sh`) and a make all these tests jinja-macroized and pull in the relevant rules from the profile. My understanding was that there was already a product check in the rule (if one was present, right?).

We already have profile specific test scenarios, for example https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/ssh/ssh_server/sshd_rekey_limit/tests/rhel8_ok.pass.sh#L2 defines a specific profile in `profiles` metadata (but feel free to use profile id also in the filename of such test scenario like `test_name.profile_name.pass/fail.sh`).

I agree that situation with SSGTS is more complex, current situation is that the test suite behaves differently in `rule` and `combined` mode.

In the `rule` mode we test selected rules and the test suite follows `profiles` metadata (so it will run test scenario for each profile listed in this metadata) and in case there is no `profiles` metadata in a test scenario it will run with the virtual `(all)` profile. There is also the option `--profile` which can be used to override the profile selection from `profiles` metadata so every test scenario will be executed only against the profile selected by this command-line argument.

In the `combined` mode we test all rules selected by a specific profile and the behavior of the test suite is a little bit different. Let's say we test `hipaa` profile. In case there is `hipaa` profile listed under `profiles` metadata in a test scenario the test scenario will be executed, otherwise it will be skipped. In case there is no `profiles` metadata in a test scenario (most of the test scenarios do not have `profiles` metadata) this mode will (instead of skipping such test scenario) run the test scenario with the virtual `(all)` profile - this way we can re-use test scenarios when testing any profile (the only disadvantage is that in this case variables will have default values). In case you want to have a specific regression test only for a certain profile(s) which relies on a specific value being selected by some variable in this profile you need to use `profiles` metadata:
`# profiles = xccdf_org.ssgproject.content_profile_ID1,xccdf_org.ssgproject.content_profile_ID2,...`
to limit test scenario so it is executed only when testing profiles listed in `profiles` metadata.

tests/ssg_test_suite/combined.py
```diff
@@ -56,10 +56,10 @@ def _modify_parameters(self, script, params):
         # be tested using the self.profile and we return empty profiles
         # metadata.
         if not params["profiles"]:
-            params["profiles"].append(rule.OSCAP_PROFILE_ALL_ID)
+            params["profiles"].extend(self.profile)
             logging.debug(
                 "Added the {0} profile to the list of available profiles for {1}"
-                .format(rule.OSCAP_PROFILE_ALL_ID, script))
+                .format(self.profile, script))
```
yuumasato, line 62: It makes sense to default to the profile initially provided, instead of `'(all)'`, but it doesn't to avoid inaccurate results with profiles that select different values than what the test scenario is prepared for.
And unfortunately this is badly documented, I think just in the comment above.

But I guess it could be argued that defaulting to `'(all)' doesn't provide much value when you actually want to test a specific one. Maybe it should skip the test scenario.

I think a workaround for your case would be to set `# profile = xccdf_org....content_profile_cis_level1_server` in the test scenario, as it would be preparing it for the profile value anyway?

cipherboy: Maybe it is best to start with documentation on how the existing test suite works currently and in what scenarios various rules and profiles are tested and what the intent of the three CLI invocations are. Maybe that's where the disconnect is. :-)

From my understanding:

 - There are two types of rules in the current CaC system (well, technically a third): 
    1. Rules which depend on a value of a variable e.g., `sudo_add_umask`
    2. Rules which are independent and stand-alone entities. E.g., `sudo_remove_nopasswd`
    3. Rules which depend on one or more other rules being enforced and passing to be passing as well. E.g., something like `aide_build_database` or other related rules which technically depends on `package_aide_installed`.  Sometimes they're written in a way that they pass if the package isn't installed. --- Note that we do have `requires` and `conflicts` in the rule itself (though appear to be used only very rarely such as some `sshd` and `auditd` rules and in the `endpoint_security` rules.
 - Rule two is the easiest: its pass/fail should be strictly independent of profile selection and the values of any and all variables.
     - In reality though, its pass/fail status may be dependent on other external rules and whether or not they've been evaluated.
     - Here, it doesn't/shouldn't matter what invocation path we choose. If we choose a rule invocation, a profile invocation, or a combined invocation, it should behave the same (+/- differences in currently hidden dependencies).
     - So, I'm inclined to ignore this type of rule as it isn't interesting.
 - Rule 1 needs some work under the current system (as I understand it):
     - I do see that there's a `# profiles` selection in the test case itself, in conjunction with a `# platform`.
         - While useful in limited cases, I think having the _value_ of the variables explicitly available is more important. This means:
             - Understanding the default value of variables (for use with `(all)`),
             - Loading and parsing the profile itself (when one was explicitly provided).
             - Jinja to load the current value of the variable into the test case.
             - This ensures the _test_ for the rule gets evaluated in the same environment the _check_ and _remediations_ get built in.
     - Here, like I've been saying above, the _intent_ of the execution (and well, documenting what each scenario is supposed to do) matters most. 
         - For `./test_suite.py rule` based execution there is the `--profile` which, given the above, would suffice.
         - For `./test_suite.py profile` based execution, profile choice is a required argument.
         - For `./test_suite.py combined` based execution, profile choice is again a required argument.
         - IMO, this data is _available_ and hence we should be using it. When I call `test_suite combined <profile>` (with `<profile>` not equal to `ALL` or `(all)`), I shouldn't be evaluating rules and conditions under the `(all)` profile; it should always be under the profile I selected on the CLI. The same principal holds, IMO, elsewhere.
         - For bulk execution, an explicit `ALL` or `(all)` should be used instead.
         - When multiple profiles are selected (with `profile` or `combined` -- note that `rule` doesn't support such), I think we could execute each rule/scenario separately under each profile. My 2c though.
 - Once rule 1 work is done, the test system could be extended to understand the `requires` and `conflicts` options and execute all required dependencies prior to executing the given rule.

Concretely, based on my current understanding, I'd propose:
 - Adding better documentation on _how_ the test suite is organized and the intent between different execution methods.
 - Add Jinja2 support to test cases to avoid the need for `profile` and `platform` as much -- to allow for _scenario_ dependent tests, making them value/platform independent.
 - Clean up profile selection for OSCAP invocations and passing that info into the test case iteslf (for use with jinja macros).
 - Updating the test suite to understand the `requires` (and `conflicts`? I'm not sure) fields. 

Thoughts?

It'll be a good bit of work... I'd like someone with familiarity with the test system as it currently exists to do the documentation... To make sure I'm not missing something obvious :D  But I'd be interested in contributing jinja support, better profile execution understanding, and dependency tracking support eventually if it is of interest the project.

cipherboy:
(Just marking as draft so it is clear it shouldn't be merged without further discussion).

tests/ssg_test_suite/rule.py
```diff
@@ -39,7 +39,7 @@ def get_viable_profiles(selected_profiles, datastream, benchmark, script=None):
     all_profiles.append(OSCAP_PROFILE_ALL_ID)
 
     for ds_profile in all_profiles:
-        if 'ALL' in selected_profiles:
+        if 'ALL' in selected_profiles or OSCAP_PROFILE_ALL_ID in selected_profiles:
```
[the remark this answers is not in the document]
cipherboy: Just bringing this up because I hit it again: there's no 1.2.18 release of OpenSCAP with support for `(all)` which makes this effectively broken on systems without that patch backported (Debian, all Ubuntu...).

jan-cerny:
@cipherboy any plans?

cipherboy:
Hi @jan-cerny -- this one didn't seem like we had consensus around it. I think from a call, the position was largely that this was working as intended, even though this was sub-optimal in some cases, as it covers other cases more broadly.

I'm going ahead and closing it. The patch is really minimal; someone else can pick it up if they're interested and new consensus is reached around behavior.

PR #6796, opened 2021-04-06
RHCOS4: Enhance instructions to better reflect how to work with the platform

JAORMX:
This changes various descriptions and macros in the case of rhcos4 to better
reflect what an admin would really need to do to check or remediate.


JAORMX:
@yuumasato @ggbecker could you take a look at this? I'm starting to change instructions for RHCOS4 to better guide users reading the content. this is a bit bigger change since I'm modifying the macros, and before going forward, I want to know what you folks think of this approach.

linux_os/guide/system/accounts/accounts-physical/disable_ctrlaltdel_burstaction/rule.yml
```diff
@@ -9,8 +9,37 @@ description: |-
...
+            path: /etc/systemd/system.conf.d/disable_ctrlaltdelete_burstaction.conf
+            overwrite: true
+    EOF
+    </pre>
+    <p>
+    This will add the relevant configuration to <tt>/etc/systemd/system.conf.d/</tt>,
+    thus configuring Systemd apropriately.
+    </p>
+    {{{ machineconfig_description_footer() | indent(4) }}}
+    {{% endif %}}
```
yuumasato, line 42: Side comment: The Jinja conditionals in between the text make things tricky to read...

One way to increase readability, would be to make it possible to split the description in two parts, one which pertains to description of the setting and its behavior, and another about product specific configuration steps. For example, `description` and `configuration`.

Another approach would be to allow entire product specific descriptions, e.g. `description@rhcos4`

linux_os/guide/system/accounts/accounts-physical/disable_ctrlaltdel_burstaction/rule.yml
```diff
@@ -9,8 +9,37 @@ description: |-
...
+        ignition:
+          version: 3.1.0
+        storage:
+          files:
+          - contents:
+              source: data:,CtrlAltDelBurstAction%3Dnone
+            mode: 0644
+            path: /etc/systemd/system.conf.d/disable_ctrlaltdelete_burstaction.conf
+            overwrite: true
+    EOF
```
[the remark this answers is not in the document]
JAORMX: Yeah, that makes sense. I'll get to that.

linux_os/guide/system/accounts/accounts-physical/disable_ctrlaltdel_burstaction/rule.yml
JAORMX: maybe the former approach is better, (having `description@<product>`

JAORMX:
@yuumasato, thanks for the review! I'll continue adding changes then! I'll target the ocil macros and service_disabled next.

linux_os/guide/services/usbguard/package_usbguard_installed/rule.yml
```diff
@@ -5,7 +5,31 @@ prodtype: fedora,rhcos4,ol7,ol8,rhel7,rhel8,rhv4
...
+          version: 3.1.0
+      extensions:
+        - usbguard
+    </pre>
+    <p>
+    This will install the <tt>usbguard</tt> package in all the
+    nodes labeled with the "master" role.
+    </p>
+    {{{ machineconfig_description_footer() | indent(4) }}}
+    {{% endif %}}
```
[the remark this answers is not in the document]
yuumasato: I see, thanks for clarification

JAORMX:
Pull-request updated, HEAD is now 8636b28429c248f9d2ef840cfbd18b645630b35c

JAORMX:
@yuumasato added changes to both descriptions and coil, WDYT?

shared/macros.jinja
```diff
@@ -679,3 +740,23 @@ openssl()
...
+{{{ weblink(link="https://docs.openshift.com/container-platform/4.6/post_installation_configuration/machine-configuration-tasks.html",
+            text="the relevant documentation") }}}.
+</p>
+{{% endmacro %}}
+
+{{% macro rhcos_node_login_instructions() -%}}
+    As a user with administrator privileges, log into a node in the relevant pool:
+    <pre>
+    $ oc debug node/$NODE_NAME
+    sh-4.4# chroot /host
```
mrogers950, line 760: This step might be clearer if we say:

At the `sh-4.4#` prompt, run:
# chroot /host

mrogers950:
one nit, looks good to me otherwise

shared/macros.jinja
JAORMX: could you propose a snippet? I'm not sure what you mean :confused:

yuumasato (approved):
LGTM

Issue #6798, opened 2021-04-06
Building OCP4 on Ubuntu 20.10 with OpenSCAP 1.2


cipherboy:
#### Description of problem:

`master` build fails with:

```
[243/907] [ocp4-content] generating ssg-ocp4-xccdf.xml
FAILED: ssg-ocp4-xccdf.xml 
cd /home/cipherboy/GitHub/cipherboy/scap-security-guide/build/ocp4 && /usr/bin/xsltproc --output /home/cipherboy/GitHub/cipherboy/scap-security-guide/build/ssg-ocp4-xccdf.xml /home/cipherboy/GitHub/cipherboy/scap-security-guide/shared/transforms/shared_xccdf-removeaux.xslt /home/cipherboy/GitHub/cipherboy/scap-security-guide/build/ocp4/xccdf-linked.xml && /usr/bin/sed -i s/oval-linked.xml/ssg-ocp4-oval.xml/g /home/cipherboy/GitHub/cipherboy/scap-security-guide/build/ssg-ocp4-xccdf.xml && /usr/bin/sed -i s/ocil-linked.xml/ssg-ocp4-ocil.xml/g /home/cipherboy/GitHub/cipherboy/scap-security-guide/build/ssg-ocp4-xccdf.xml && env PYTHONPATH=/home/cipherboy/GitHub/cipherboy/scap-security-guide /usr/bin/python3 /home/cipherboy/GitHub/cipherboy/scap-security-guide/build-scripts/unselect_empty_xccdf_groups.py --input /home/cipherboy/GitHub/cipherboy/scap-security-guide/build/ssg-ocp4-xccdf.xml --output /home/cipherboy/GitHub/cipherboy/scap-security-guide/build/ssg-ocp4-xccdf.xml && /usr/bin/oscap xccdf resolve -o /home/cipherboy/GitHub/cipherboy/scap-security-guide/build/ssg-ocp4-xccdf.xml /home/cipherboy/GitHub/cipherboy/scap-security-guide/build/ssg-ocp4-xccdf.xml && /usr/bin/xmllint --nsclean --format --output /home/cipherboy/GitHub/cipherboy/scap-security-guide/build/ssg-ocp4-xccdf.xml /home/cipherboy/GitHub/cipherboy/scap-security-guide/build/ssg-ocp4-xccdf.xml
[...]

PR #6799, opened 2021-04-06
Disable OCP4 building on older OpenSCAP releases

cipherboy:
#### Description

SSIA.

#### Rationale:

OCP4 fails to build on Ubuntu; we should only build it if the relevant OpenSCAP is present.

----

Note: this will need to be updated once OpenSCAP 1.3.x is released with
OCP4 content.

Resolves: #6798

----

I'm curious if there's a different version check we should use, or if everyone developing OCP4 content is running on OpenSCAP 1.4.0?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -155,6 +155,14 @@ if (SSG_SVG_IN_XCCDF_ENABLED AND NOT OSCAP_SVG_SUPPORT_RESULT EQUAL 0)
     message(WARNING "Your version of OpenSCAP does not support having the SVG logo in the XCCDF, disabling SVG logo.")
 endif()
 
+# OCP4 requires non-standard extensions. Vanilla OpenSCAP 1.2 doesn't support
+# it. See also: https://github.com/ComplianceAsCode/content/issues/6798
+if ("${OSCAP_VERSION}" VERSION_LESS "1.3.4" AND SSG_PRODUCT_OCP4)
+    set(SSG_PRODUCT_OCP4 OFF)
+    message(WARNING "Won't build OCP4 content as it requires an OpenSCAP version with OCP4 support. See also: https://github.com/ComplianceAsCode/content/issues/6798. n.b.: if 1.3.4 fails to build OCP4 content, please update to 1.3.5")
+endif()
+
+
 if (SSG_JINJA2_CACHE_ENABLED)
     file(MAKE_DIRECTORY "${SSG_JINJA2_CACHE_DIR}")
     if (NOT EXISTS "${SSG_JINJA2_CACHE_DIR}")


CMakeLists.txt
```diff
@@ -154,6 +154,16 @@ if (SSG_SVG_IN_XCCDF_ENABLED AND NOT OSCAP_SVG_SUPPORT_RESULT EQUAL 0)
     message(WARNING "Your version of OpenSCAP does not support having the SVG logo in the XCCDF, disabling SVG logo.")
 endif()
 
+# OSP4 requires non-standard extensions. Vanilla OpenSCAP 1.2 doesn't support
+# it. See also: https://github.com/ComplianceAsCode/content/issues/6798
+# TODO: this should be updated when a 1.3.x version is released with ocp4
+# content support.
+if ("${OSCAP_VERSION}" VERSION_LESS "1.4.0")
```
cipherboy, line 161: Looks like this version is wrong. CI is running with 1.3.4 and it otherwise works. Does anyone know what the correct version is here? Did 1.3.3 ship the required changes?

JAORMX: What if we just use 1.3.4? I honestly don't remember when the changes were shipped.

cipherboy: The worst case is someone with an older oscap can't build it. We don't yet have a way to override this disable... So we'd want the version to be precise. I wonder if we should do the opposite? Disable OCP4 by default and if the default is on and we have a  new enough oscap, turn it on. That doesn't help if you explicitly disable OCP4 (and leave default as on) but... I dunno.

cipherboy: (As mentioned in #6798 -- this could be in the event of a heavily patched OSCAP) -- we'd need a feature check if we wanted to.

CMakeLists.txt
```diff
@@ -154,6 +154,16 @@ if (SSG_SVG_IN_XCCDF_ENABLED AND NOT OSCAP_SVG_SUPPORT_RESULT EQUAL 0)
     message(WARNING "Your version of OpenSCAP does not support having the SVG logo in the XCCDF, disabling SVG logo.")
 endif()
 
+# OSP4 requires non-standard extensions. Vanilla OpenSCAP 1.2 doesn't support
```
yuumasato, line 157: Typo: OCP4

CMakeLists.txt
yuumasato: 1.3.4 is the oldest OpenSCAP version that can build current OCP4 content.
See these logs for example: https://jenkins.complianceascode.io/view/All%20Green/job/scap-security-guide-nightly-zip/953/consoleFull

> I wonder if we should do the opposite? Disable OCP4 by default and if the default is on and we have a new enough oscap, turn it on.

As tendency is for adoption of oscap with OCP building capability to spread, I'd leave it ON by default and check for oscap version

> (As mentioned in #6798 -- this could be in the event of a heavily patched OSCAP) -- we'd need a feature check if we wanted to.

I don't understand what "in the event of a heavily patched OSCAP" means

cipherboy: >> (As mentioned in #6798 -- this could be in the event of a heavily patched OSCAP) -- we'd need a feature check if we wanted to.

> I don't understand what "in the event of a heavily patched OSCAP" means

@evgenyz [commented](https://github.com/ComplianceAsCode/content/issues/6798#issuecomment-814719100) saying that oscap in Fedora and Red Hat were heavily patched.

Let's say upstream release 1.3.5 actually adds the missing features/fixes some bugs for OCP4 content. Upstream release 1.3.4 didn't have them. I think @evgenyz's point was that the Fedora/RHEL build _might_ have version 1.3.4 _plus_ a bunch of patches to make it behave like 1.3.5 for OCP4 content building. In this case, the "proper" upstream CaC/content version number would be OSCAP 1.3.5, _but_ that'd break all Fedora users, because they actually have 1.3.4+fedora_patches. -- and the fedora_patches part isn't detected/represented in the three-segment version number.

This is a hypothetical situation where the version number itself isn't sufficient and feature detection (does OCSAP actually support the yaml probes or whatever is missing) would work better.

My 2. :) Hope that's clearer. Just making sure if we check for any upstream version x.y.z, we don't have a x.y.z-1 with a bunch of patches that make it work running around in Fedora/RHEL. I know Ubuntu is still running 1.2.17 so there's no risk of us having OCP4 patches there :)

yuumasato: I see now, It seems I'm having trouble understanding things today, :-P

I had not realized that other fixes in unreleased 1.3.5 were also required.

cipherboy: No problem :) So I think perhaps we just ship with 1.3.4 as the version check and deal with any mix-ups in versions and hope that anyone who is at 1.3.4 and can't build OCP4 can upgrade to 1.3.5?

CMakeLists.txt
cipherboy: Thanks!

CMakeLists.txt
yuumasato: Looks good to me.
AFAIK (and I may be just ignorant), Fedora tends to ship the latest greatest OpenSCAP, so it would be the only one affected by this.

Still, we can update this to 1.3.5 once it is out and have this settled, :)

CMakeLists.txt
```diff
@@ -155,6 +155,14 @@ if (SSG_SVG_IN_XCCDF_ENABLED AND NOT OSCAP_SVG_SUPPORT_RESULT EQUAL 0)
     message(WARNING "Your version of OpenSCAP does not support having the SVG logo in the XCCDF, disabling SVG logo.")
 endif()
 
+# OCP4 requires non-standard extensions. Vanilla OpenSCAP 1.2 doesn't support
+# it. See also: https://github.com/ComplianceAsCode/content/issues/6798
+if ("${OSCAP_VERSION}" VERSION_LESS "1.3.4")
+    set(SSG_PRODUCT_OCP4 OFF)
+    message(WARNING "Won't build OCP4 content as it requires an OpenSCAP version with OCP4 support. See also: https://github.com/ComplianceAsCode/content/issues/6798. n.b.: if 1.3.4 fails to build OCP4 content, please update to 1.3.5")
```
yuumasato, line 162: I think it would make sense to print this message only if SSG_PRODUCT_OCP4 was ON, would be less annoying for configs that already disabled it.

CMakeLists.txt
yuumasato: @evgenyz @JAORMX  thoughts about approach above?

JAORMX: We use UBI (RHEL-based) to build the content in OCP4, not Fedora. If we get 1.3.5 available we could take it into use.

CMakeLists.txt
cipherboy: Ah good point. Thanks @yuumasato! Updated :)

CMakeLists.txt
yuumasato: I see, but it should be aligned with what gets into RHEL, right? So version 1.3.4.

yuumasato (approved):

commandline-be:
same issue with oscap 1.3.5

ggbecker:
> same issue with oscap 1.3.5

@commandline-be 

that would be strange... which distribution are you using? openscap may have be built without the support for the yamlcontentfile: https://github.com/OpenSCAP/openscap/blob/367170182664bf34536266b5495c12580b4ddb8b/CMakeLists.txt#L271

That's the only reasonable explanation to me.

commandline-be:
> > same issue with oscap 1.3.5
> 
> @commandline-be
> 
> that is strange... which distribution are you using? openscap may have be built without the support for the yamlcontentfile: https://github.com/OpenSCAP/openscap/blob/367170182664bf34536266b5495c12580b4ddb8b/CMakeLists.txt#L271
> 
> That's the only reasonable explanation to me.

not long after I left the project where this was applicable, I'm unable to follow up because of this

PR #6800, opened 2021-04-06
Update Ubuntu CPEs to match published CPEs

cipherboy:
#### Description:

From both `official-cpe-dictionary_v2.3.xml` and
`official-cpe-dictionary_v2.2.xml`, Ubuntu's CPEs include the LTS suffix.
Update CaC/content's copies to match.

Note: this only applies for LTS releases. If non-LTS releases were to
be included later, they wouldn't include this suffix.

#### Rationale:

Also working on a PR to get `CPE_NAME` added to `base-files`; figured we should make it consistent with the NIST NVD entries everywhere.


JAORMX (approved):

cipherboy:
Thanks for the review @JAORMX! Since this is only a CPE update, I'll go ahead and merge.

PR #6801, opened 2021-04-07
Add the RHEL9 product

matejak:
RHEL9 will come one day, so it is better to be prepared.

The product has been created using RHEL8 as a basis, and making sure that the build works.
No profiles or STIG overlay for the time being.


JAORMX (approved):

rhel9/product.yml
```diff
@@ -0,0 +1,32 @@
...
+release_key_fingerprint: "567E347AD0044ADE55BA8A5F199E2F91FD431D51"
+auxiliary_key_fingerprint: "6A6AA7C97C8890AEC6AEBFE2F76F66C3D4082792"
+oval_feed_url: "https://www.redhat.com/security/data/oval/com.redhat.rhsa-RHEL9.xml"
+
+cpes_root: "../shared/applicability"
+cpes:
+  - rhel9:
+      name: "cpe:/o:redhat:enterprise_linux:9"
+      title: "Red Hat Enterprise Linux 9"
+      check_id: installed_OS_is_rhel9
```
yuumasato, line 28: I think the build will also fail if a check with this ID doesn't exist.

yuumasato (changes requested):
I get this error when trying to build:
```
-- Build files have been written to: /home/wsato/git/content/build
make[2]: *** No rule to make target '../rhel9/profiles', needed by 'rhel9/profiles'.  Stop.
make[1]: *** [CMakeFiles/Makefile2:1319: rhel9/CMakeFiles/generate-internal-rhel9-shorthand.xml.dir/all] Error 2
make: *** [Makefile:171: all] Error 2
```

I also noticed that CI didn't build the RHEL9 product, is that intended?

rhel9/product.yml
matejak: Both findings are related to git-related omissions, and have been fixed, thanks for noticing.

yuumasato (approved):
`> I also noticed that CI didn't build the RHEL9 product, is that intended?`

Discussed this with @matejak  and enabling build and test of it right now will impact currently open PRs. So this can be enabled in the short term future to minimize impact on others.

Edit: sorry for wrong mention.

PR #6802, opened 2021-04-07
SLES-15-030390 add rule, remediation and test

brett060102:
#### Description:

-  SLES-15-030390 'Ensure auditd Collects Information on the Use of Privileged Commands - rmmod'

#### Rationale:

- Create rule and tests audit_rules_privileged_commands_rmmod

### DIFF
--- a/linux_os/guide/services/ntp/chronyd_client_only/tests/ocp4/e2e.yml
+++ b/linux_os/guide/services/ntp/chronyd_client_only/tests/ocp4/e2e.yml
@@ -1,3 +1,0 @@
----
-default_result: FAIL
-result_after_remediation: PASS
--- a/linux_os/guide/services/ntp/chronyd_client_only/tests/ocp4/e2e.yml
+++ b/linux_os/guide/services/ntp/chronyd_client_only/tests/ocp4/e2e.yml
@@ -1,0 +1,3 @@
+---
+default_result: FAIL
+result_after_remediation: PASS
--- a/linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_rmmod/rule.yml
+++ b/linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_rmmod/rule.yml
@@ -1,0 +1,54 @@
+documentation_complete: true
+
+prodtype: sle15
+
+title: 'Ensure auditd Collects Information on the Use of Privileged Commands - rmmod'
+
+description: |-
+    At a minimum, the audit system should collect the execution of
+    privileged commands for all users and root. If the <tt>auditd</tt> daemon is
+    configured to use the <tt>augenrules</tt> program to read audit rules during
+    daemon startup (the default), add a line of the following form to a file with
+    suffix <tt>.rules</tt> in the directory <tt>/etc/audit/rules.d</tt>:
+    <pre>-w /sbin/rmmod -p x -k modules</pre>
+
+rationale: |-
+    Misuse of privileged functions, either intentionally or unintentionally by
+    authorized users, or by unauthorized external entities that have compromised system accounts,
+    is a serious and ongoing concern and can have significant adverse impacts on organizations.
+    Auditing the use of privileged functions is one way to detect such misuse and identify
+    the risk from insider and advanced persistent threats.
+    <br /><br />
+    Privileged programs are subject to escalation-of-privilege attacks,
+    which attempt to subvert their normal role of providing some necessary but
+    limited capability. As such, motivation exists to monitor these programs for
+    unusual activity.
+
+severity: medium
+
+identifiers:
+    cce@sle15: CCE-85732-6
+
+references:
+    disa@sle15: CCI-000130,CCI-000169,CCI-000172,CCI-002884
+    nist@sle15: AU-12(c),AU-12.1(iv),AU-3,AU-3.1,AU-12(a),AU-12.1(ii),MA-4(1)(a)
+    srg@sle15: SRG-OS-000037-GPOS-00015,SRG-OS-000062-GPOS-00031,SRG-OS-000392-GPOS-00172,SRG-OS-000462-GPOS-00206,SRG-OS-000471-GPOS-00215
+    stigid@sle15: SLES-15-030390
+
+ocil_clause: 'it is not the case'
+
+ocil: |-
+    To verify that auditing of privileged command use is configured, run the
+    following command:
+    <pre>
+       sudo auditctl -l | grep -w '/sbin/rmmod'
+       -w /sbin/rmmod -p x -k modules
+    </pre>
+    If the system is configured to audit the execution of the module management program "rmmod",
+    the command will return a line.
+
+
+template:
+    name: audit_rules_privileged_commands
+    vars:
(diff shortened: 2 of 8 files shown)


linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_rmmod/rule.yml
```diff
@@ -0,0 +1,54 @@
...
+    following command:
+    <pre>
+       sudo auditctl -l | grep -w '/sbin/rmmod'
+       -w /sbin/rmmod -p x -k modules
+    </pre>
+    If the system is configured to audit the execution of the module management program "rmmod",
+    the command will return a line.
+
+
+template:
```
vojtapolasek, line 51: I think that this template will achieve a different result than mentioned in the rule description. You can try to apply the rule on your machine, you will get:
-a always,exit -F path=/usr/sbin/rmmod -F auid>=1000 -F auid!=unset -F key=privileged

Consider creating your own remediation and check or modifying the description.

vojtapolasek (changes requested):
Thank you for the new rule.
I think that adding a test for ocp4 does not make much sense here since the rule is for sle.
If you want to add tests, please look at the following blog post for introduction:
https://complianceascode.github.io/template/2021/03/25/tests_howto.html
Please see other comments for more feedback.

linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_rmmod/rule.yml
brett060102: You are correct and will do.

brett060102: @vojtapolasek I have been looking at stigs, SLES-15-030370, SLES-15-030380, SLES-15-030390, SLES-15-030400,  SLES-15-030410 and then SLES-15-030420

SLES-15-030370 and SLES-15-030410, have the output of form:
-a always,exit -S all -F path=

There are number of other stigs that also use that output.

But SLES-15-030380, SLES-15-030390, SLES-15-030400,  and SLES-15-030410 look for:
-w /path/command -p x -k modules

based on https://documentation.suse.com/sles/12-SP4/html/SLES-all/cha-audit-scenarios.html I think that the "-a always,exit -S all -F path=..." Is what the is needed and not the "-w /path/command -p x -k modules" The latter is a watch and based on reading of the above doc, normally associated with things like  directories, config file, log files and not commands or kernel syscalls.

I am going to update this PR based on the assumption that we want:
-a always,exit -F path=/usr/sbin/rmmod -F auid>=1000 -F auid!=unset -F key=privileged

As for the testing and tests/ocp4/e2e.yml  we looked over the test doc and have written .sh tests for other rules. We based this rule off of another similar rule it had tests/ocp4/e2e.yml I could not find much about it in the docs, but in looking at the other occurrences of it it seemed to be a general way of testing things where the initial state was fail and the pass case was done via remediation. It did not seem that tests/ocp4/e2e.yml were distro specific. Is there a write up on it somewhere.
I have nothing against, writing specific tests, but since we are using a template, I am not sure what could be dome beyond simple pass and simple fail.

@abergmann  Do you have any comments on this?

brett060102:
The light is slowing starting to dawn here. /sbin/rmmod  and /usr/sbin/rmmod are symlinks to /usr/bin/kmod. So, it should be a "-w" because we need to audit if the link is changed. Still not sure why the rule only worries about /sbin'kmod a not /usr/sbin/kmod.

Same issue applies to  SLES-15-030380 to SLES-15-030410.

Thank you and sorry for our confusion.

linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_rmmod/rule.yml
rumch-se: @vojtapolasek  
In the realization of SLES-15-030390 I used a predefined template which provides "-a always...", while this particular rule is looking for "-w /path/command ...". I tried to find in the list of existing templates a good replacement of audit_rules_privileged_commands (to fix the issue), but unfortunately, I did not find any.  Any advice in this direction will be welcomed.  For me sounds strange your finding about SLES-15-030410 - because its realization is this https://github.com/ComplianceAsCode/content/tree/master/linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_kmod and according to the code of the rule the expected code is "-w /path/command -p x -k modules" (look at https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_kmod/ansible/shared.yml)

linux_os/guide/services/ntp/chronyd_client_only/tests/ocp4/e2e.yml
```diff
@@ -1,3 +0,0 @@
----
-default_result: FAIL
-result_after_remediation: PASS
```
brett060102, line 3: Not sue how this happened. looking at it now.

brett060102: above issue fixed

brett060102:
@vojtapolasek the rework is complete.

vojtapolasek:
Hello, thank you for changes. And thank you for tests! I tried to test the rule on rhel system and received the following fatail failure when running the remediation:
```
TASK [Restart auditd.service] **************************************************
fatal: [192.168.122.107]: FAILED! => {"changed": false, "msg": "Unable to restart service auditd.service: Failed to restart auditd.service: Operation refused, unit auditd.service may be requested by dependency only (it is configured to refuse manual start/stop).\nSee system logs and 'systemctl status auditd.service' for details.\n"}
```

Does remediation work for SLE?
Moreover, I have a suggestion. We already have Jinja templates for such watch rules, you can see example usage in this remediation:
https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/auditing/auditd_configure_rules/audit_rules_session_events/ansible/shared.yml
You can find the macro among others here:
https://complianceascode.readthedocs.io/en/latest/jinja_macros/ansible.html
Could you consider using of this macros? I know that they are not checking for usage of augen vs auditctl. The question is, if it is actually important.
Usage of these macros decreases code duplication, that's why I suggest it. What do you think about it?

brett060102:
@vojtapolasek Yes, the remediation runs fine on SLE, did not know about those macros, I will give those a shot. Just did not know about them. Thank you, I will work on recoding.

brett060102:
@vojtapolasek changes to use the watch file macros, it is a lot cleaner. Thank you.

vojtapolasek:
No problem. Now you know about macros. I suggest before implementing any oval / ansible / bash thing, have a look into the appropriate macro file. It can save you some time. Changes look good, Openshift CI failures are not relevant to this rule because it is sle15 only. Merging.

PR #6805, opened 2021-04-07
Add CPE OVAL check for SLES for SAP 12 and 15.

brett060102:
#### Description:

- Add an additional OVAL check to cover SLES for SAP 12 and 15.

The general SLE OVAL check is looking for the sles-release or sled-release package and is checking its version number. 
SLES for SAP comes with a different release file called SLES_SAP-release that needs to be verified. 

#### Rationale:

- need to support SLES for SAP


yuumasato (approved):
LGTM.
@brett060102 would you like someone on Suse side to double check it?

msmeissn:
@brett060102 actually works for suse. I can confirm the check is good, and the product SLES4SAP is using the SLES_SAP-release package as specified in the oval.

yuumasato:
Thanks @msmeissn 

I wanted to know if it made sense for someone else to check it as well.

PR #6810, opened 2021-04-08
Use v1beta1 for api_server_api_priority_flowschema_catch_all object

mrogers950:
v1alpha1 for FlowSchemas is deprecated and removed in kube 1.21.

PR #6813, opened 2021-04-09
OCP4: Address flowschema version change by handling different OCP versions

JAORMX:
OpenShift 4.8 rebased on top of Kubernetes 1.21. This version had a lot of nice
features, but one change affected our content.

The `flowschema` resource is now version `v1beta1` and the `v1alpha` version we
were using was completely removed.

Given that we've made the decision to use one branch for the content of all the
versions we support. We're in need of addressing this.

This is addressed as follows:

## Create CPE per OCP4 version to handle deprecations
    
Since OCP4 uses the same datastream for multiple version, the time has
come for us to start differentiating between what's supported in each
version. Thus, we have now specific CPEs for different versions that can
be used to delimit what's supported.

On the flowschema case, we now have two rules:

* The original one now reflects the new version that's supported in Kubernetes.
* A new rule, that reflects the `v1alpha1` version.
    
The each of these rules now has the appropriate supported versions
marked on the `platforms` key.

## Add ability to do version-dependent E2E tests
    
This enables the framework to do tests based on the version the cluster
is running. We can then add different expectations based on the cluster's
version.

The flowschema tests were enhanced with this capability.


JAORMX:
This PR supersedes https://github.com/ComplianceAsCode/content/pull/6810

JAORMX:
@yuumasato @ggbecker what do you guys think of this approach?

JAORMX:
@openscap-ci test this please

mrogers950:
LGTM! Will let others weigh in.

applications/openshift/api-server/api_server_api_priority_flowschema_catch_all/rule.yml
```diff
@@ -26,6 +26,10 @@ rationale: |-
 identifiers:
   cce@ocp4: CCE-83522-3
 
+platforms:
+  - ocp4.8
+  - ocp4.9
+  - ocp4.10
```
yuumasato, line 32: I worry that the list of platform will keep growing bigger and bigger.

Do you think it would make sense to have platforms named `flowschemas_v1alpha1` and `flowschemas_v1beta1`? And they would know and check the OCP version that use each flowschemas.

Another approach: If the platform check is written to check for version ranges, instead of individual versions, maintenance of the platform would happen for each change in flowschemas version opposed to every ocp release. For example, `flowschemas_v1beta1` is for ocp 4.8 or higher.
But this would be a problem if it is expected that older content versions work on newer ocp. And I think it would not be a problem for newer content version to work on older releases.

yuumasato (commented):
The approach looks great to me, :)

I just have a slight concern about amount of platforms it will generate in the long run.

JAORMX:
@yuumasato the idea would be to delete platforms as they loose support. So it would be many, but not as bad.

JAORMX:
@openscap-ci test this please

yuumasato (approved):

PR #6814, opened 2021-04-09
Add ignore and rotate options to auditd/audisp variables

mildas:
#### Description:
Add missing options to auditd and audisp variables.

### DIFF
--- a/linux_os/guide/system/auditing/configure_auditd_data_retention/var_audispd_network_failure_action.var
+++ b/linux_os/guide/system/auditing/configure_auditd_data_retention/var_audispd_network_failure_action.var
@@ -17,3 +17,4 @@ options:
     syslog: syslog
     warn_once: warn_once
     stop: stop
+    ignore: ignore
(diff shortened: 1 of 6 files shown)


JAORMX (approved):

vojtapolasek:
Now when I am thinking about it, I wonder if these options should be actually available in the content. The option "ignore" will just ignore the problem, potentially filling up the storage space and therefore losing ability to write further logs. The option "rotate" will cause loss of older log records, potentially losing evidence. What do others think? @matejak  @yuumasato

JAORMX:
@vojtapolasek I think that's still acceptable as you might have other mechanisms that provide supporting controls. e.g. you're using rsyslog or fluentd to ship audit logs immediately as they come. Rotation could also be happening externally, so in that case `ignore` might be an acceptable setting.

yuumasato:
@JAORMX has a good point.
Just looking at the rule by itself there is no direct security benefit.
But in the context of a whole profile, or deployed system, the requirement could be addressed by other means and the newly added variable options may be needed.

yuumasato (approved):

PR #6819, opened 2021-04-09
OL hipaa profiles

freddieRv:
#### Description:

- Add OL7 hipaa profile
- Update OL8 hipaa profile

#### Rationale:

- OL profiles effort


jan-cerny:
Thanks for your patch.

PR #6822, opened 2021-04-11
SLES-15-10352 rule

teacup-on-rockingchair:
#### Description:

- Verify that Shared Library Directories Have Restrictive Permissions

#### Rationale:

- Clone implementation of file_permissions_library_dirs to dir_permissions_library_dirs
- Add some tests to cover basic scenarios


linux_os/guide/system/permissions/files/permissions_within_important_dirs/dir_permissions_library_dirs/rule.yml
```diff
@@ -0,0 +1,57 @@
...
+    </pre>
+    Kernel modules, which can be added to the kernel during runtime, are
+    stored in <tt>/lib/modules</tt>. All sub-directories in these directories
+    should not be group-writable or world-writable. If any file in these
+    directories is found to be group-writable or world-writable, correct
+    its permission with the following command:
+    <pre>$ sudo chmod go-w <i>DIR</i></pre>
+
+rationale: |-
+    If the SUSE operating system were to allow any user to make changes to software libraries,
```
brett060102, line 27: Don't think this should be SUSE specific.

linux_os/guide/system/permissions/files/permissions_within_important_dirs/dir_permissions_library_dirs/oval/shared.xml
```diff
@@ -0,0 +1,46 @@
...
+    <filter action="exclude">dir_perms_state_symlink</filter>
+  </unix:file_object>
+
+  <unix:file_object comment="library files" id="dir_object_file_permissions_lib_files" version="1">
+    <!-- Check the files within /lib, /lib64, /usr/lib, /usr/lib64 directories have safe permissions (go-w) -->
+    <unix:path operation="pattern match">^\/lib(|64)|^\/usr\/lib(|64)</unix:path>
+    <unix:filename operation="pattern match">^.*$</unix:filename>
+    <filter action="include">dir_state_perms_nogroupwrite_noworldwrite</filter>
+    <filter action="exclude">dir_perms_state_symlink</filter>
+  </unix:file_object>
```
brett060102, line 35: It looks like the dir_object_file_permissions_lib_files and dir_object_file_permissions_lib_dir are doing the same thing. Am I missing something here? I guess from test name that one is for dirs and files, but test look there are the same and the rule is for dir's only so. I feel uncomfortable if the oval does something different than what is in the rule file.

sle15/profiles/stig.profile
```diff
@@ -147,6 +147,7 @@ selections:
     - dconf_db_up_to_date
     - dconf_gnome_banner_enabled
     - dconf_gnome_login_banner_text
+    - dir_permissions_library_dirs
```
brett060102, line 150: There is a conflict on this file that should be addressed.

linux_os/guide/system/permissions/files/permissions_within_important_dirs/dir_permissions_library_dirs/ansible/shared.yml
```diff
@@ -0,0 +1,18 @@
+# platform = SUSE Linux Enterprise 15
```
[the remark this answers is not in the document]
teacup-on-rockingchair: Thanks for the comment - should be ok in 86173c7

sle15/profiles/stig.profile
teacup-on-rockingchair: rebased the branch so conflict should be resolved now in 07143f3

linux_os/guide/system/permissions/files/permissions_within_important_dirs/dir_permissions_library_dirs/rule.yml
```diff
@@ -0,0 +1,57 @@
+documentation_complete: true
+
+title: |-
+{{% if product in ["sle15"] %}}
```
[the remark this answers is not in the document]
teacup-on-rockingchair: thanks - removed that in 4e4ea53

linux_os/guide/system/permissions/files/permissions_within_important_dirs/dir_permissions_library_dirs/rule.yml
teacup-on-rockingchair: thanks - removed that in 4e4ea53

ggbecker:
@openscap-ci ok to test

linux_os/guide/system/permissions/files/permissions_within_important_dirs/dir_permissions_library_dirs/oval/shared.xml
ggbecker: It probably needs to replace the `unix:filename` with `<unix:filename xsi:nil="true" />`. See how it's done in the [file_permissions template](https://github.com/ComplianceAsCode/content/blob/master/shared/templates/file_permissions/oval.template#L35-L41)

linux_os/guide/system/permissions/files/permissions_within_important_dirs/dir_permissions_library_dirs/oval/shared.xml
```diff
@@ -0,0 +1,46 @@
...
+    <unix:path operation="pattern match">^\/lib(|64)|^\/usr\/lib(|64)</unix:path>
+    <unix:filename xsi:nil="true" />
+    <filter action="include">dir_state_perms_nogroupwrite_noworldwrite</filter>
+    <filter action="exclude">dir_perms_state_symlink</filter>
+  </unix:file_object>
+
+  <unix:file_object comment="library files" id="dir_object_file_permissions_lib_files" version="1">
+    <!-- Check the files within /lib, /lib64, /usr/lib, /usr/lib64 directories have safe permissions (go-w) -->
+    <unix:path operation="pattern match">^\/lib(|64)|^\/usr\/lib(|64)</unix:path>
+    <unix:filename operation="pattern match">^.*$</unix:filename>
```
ggbecker, line 32:
```suggestion
    <unix:filename xsi:nil="true" />
```

teacup-on-rockingchair: The idea of the rule `dir_object_file_permissions_lib_files` was to match all files in the directories, as far as I understood `  <unix:filename xsi:nil="true" />` will match only the directories, which should have been done in `dir_object_file_permissions_lib_dir`, so my point here is to match all files with `dir_object_file_permissions_lib_files` and dirs and subdirs with `dir_object_file_permissions_lib_dir` object. I guess once you are suggested changes I misunderstood something in the syntax. When I implemented it I had in mind the file_object description in the oval spec (https://oval.mitre.org/language/version5.11/ovaldefinition/documentation/unix-definitions-schema.html), where it is stated that:
```
...
For example, one would set xsi:nil to true if the desire was to test the attributes or permissions associated with a directory. Setting xsi:nil equal to true is different than using a .* pattern match, which says to collect every file under a given path. 
```

brett060102 (changes requested):
test cases need to end in either .fail.sh or .pass.sh so for example:

rename:
linux_os/guide/system/permissions/files/permissions_within_important_dirs/dir_permissions_library_dirs/tests/world_writable_dir_on_usr_lib.sh

to
linux_os/guide/system/permissions/files/permissions_within_important_dirs/dir_permissions_library_dirs/tests/world_writable_dir_on_usr_lib.fail.sh
same for owner_only_writable_dir.sh, world_writable_dir_on_lib.sh, world_writable_dir_on_lib.sh, world_writable_dir_on_usr_lib.sh

linux_os/guide/system/permissions/files/permissions_within_important_dirs/dir_permissions_library_dirs/oval/shared.xml
ggbecker: I have probably misread the comment from brett. If the idea is to check for all files in the directories then it is probably fine. I'm will go through the PR once again soon, please bear with me.

ggbecker: Based on text from `SLES-15-010351`, I understand that no files should be checked within this rule, I think that was my original concern.

The library files test should be covered by this rule already:
https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/permissions/files/permissions_within_important_dirs/file_permissions_library_dirs/rule.yml#L43

My question is, do we really want to have the test `dir_object_file_permissions_lib_files` in this rule?

brett060102: @ggbecker  @teacup-on-rockingchair  SLES-15-010352 the STIG for this PR is explicit in covering directories only "-type d", SLES-15-010351 is explicit in covering files "-type f" So, in this case the oval should only be handling directories. So, if I understand this correctly <unix:filename xsi:nil="true" /> is what is needed here.

linux_os/guide/system/permissions/files/permissions_within_important_dirs/dir_permissions_library_dirs/tests/world_writable_dir_on_usr_lib.sh
```diff
@@ -0,0 +1,5 @@
+# platform = multi_platform_sle
+DIRS="/usr/lib /usr/lib64"
+for dirPath in $DIRS; do
+	mkdir -p "$dirPath/testme" && chmod 777 "$dirPath/test"
```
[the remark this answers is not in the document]
teacup-on-rockingchair: Thx should be ok in 8851e58

linux_os/guide/system/permissions/files/permissions_within_important_dirs/dir_permissions_library_dirs/oval/shared.xml
teacup-on-rockingchair: Thanks @brett060102 and @ggbecker , it seems I was totally confused on this one. Dropped completely the rule for files so now(252fa369214604e990f7375532e81448bfeb227e) only dirs are checked by this one.

ggbecker:
Unfortunately, there is a conflict with the stig.profile file, can you resolve it?

linux_os/guide/system/permissions/files/permissions_within_important_dirs/dir_permissions_library_dirs/rule.yml
```diff
@@ -0,0 +1,53 @@
...
+    individuals must be allowed to obtain access to information system components for purposes
+    of initiating changes, including upgrades and modifications.
+
+identifiers:
+    cce@sle15: CCE-85753-2
+
+severity: medium
+
+references:
+    cci@sle15: CCI-001499
```
ggbecker, line 39:
```suggestion
    disa: CCI-001499
```

Unfortunately the correct reference here is `disa`, see in the references table here: https://complianceascode.readthedocs.io/en/latest/manual/developer/06_contributing_with_content.html#rules

ggbecker (approved):
LGTM. Thanks for addressing all the requests. The next step is to remove the dirs checking from this rule: https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/permissions/files/permissions_within_important_dirs/file_permissions_library_dirs/oval/shared.xml#L21

PR #6823, opened 2021-04-12
OCP4: Add applicability warnings

JAORMX:
These will show up as part of the results and guide users to why a check
is marked as NOT-APPLICABLE.


mrogers950:
LGTM!

PR #6824, opened 2021-04-12
Add `dconf_gnome_lock_screen_on_smartcard_removal` to cover RHEL-08-020050

ggbecker:
#### Description:

- Add `dconf_gnome_lock_screen_on_smartcard_removal` to cover RHEL-08-020050

#### Discussion:
- The `check content` from [V-230351](https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_8_security_technical_implementation_guide/V-230351) suggests that the configured file is under `distro.d`, this is not a hard requirement and for sake of consistency I'm proposing to use `local.d` since most of rules use this dconf database source.


carlosmmatos:
LGTM - we can chat more about a new request of separating out the configuration from the lock.. but that will become another PR.

PR #6825, opened 2021-04-12
Render set of rules as HTML

matejak:
There may be cases when it is convenient to present a set of manually selected rules to people who are not familiar with the project (e.g. to component maintainers who could assess impact of changes to their components on rules).
This PR introduces a script that can produce a human-readable summary of rule descriptions based on a selection of rule IDs that were built for a particular product.
Introduction of component metadata to rules could automate this task almost entirely, right now, the only way of selecting relevant rules would use `grep` or `find` commands.

Aside from that, this PR introduces a class that allows writing such HTML renderers with minimal code duplication.

### DIFF
--- a/utils/render-rules.py
+++ b/utils/render-rules.py
@@ -1,0 +1,64 @@
+#!/usr/bin/python3
+
+from glob import glob
+import collections
+import os
+import pathlib
+
+import argparse
+
+import ssg.build_yaml
+import ssg.controls
+import ssg.yaml
+import ssg.jinja
+import template_renderer
+
+
+class HtmlOutput(template_renderer.Renderer):
+    TEMPLATE_NAME = "rules-template.html"
+
+    def _get_all_compiled_profiles(self):
+        compiled_profiles = glob(str(self.built_content_path / "profiles" / "*.profile"))
+        profiles = []
+        for p in compiled_profiles:
+            profiles.append(ssg.build_yaml.Profile.from_yaml(p))
+        return profiles
+
+    def _set_rule_profiles_membership(self, rule, profiles):
+        rule.in_profiles = []
+        for p in profiles:
+            if rule.id_ in p.selected:
+                rule.in_profiles.append(p)
+
+    def process_rules(self, rule_files, component):
+        rules = []
+        profiles = self._get_all_compiled_profiles()
+
+        for r_file in rule_files:
+            rule = ssg.build_yaml.Rule.from_yaml(r_file, self.env_yaml)
+            self._set_rule_relative_definition_location(rule)
+            self._set_rule_profiles_membership(rule, profiles)
+
+            rules.append(rule)
+
+        self.template_data["rules"] = rules
+        self.template_data["component"] = component
+
+
+def parse_args():
+    parser = HtmlOutput.create_parser(
+        "Pass a list of rule YAMLs (that are related to a component), and the script will "
+        "render their summary as an HTML along with links to the usage of these rules in profiles "
+        "and with links to the upstream rule source.")
+    parser.add_argument(
+        "--component", help="Name of the component.")
+    parser.add_argument(
+        "rule", metavar="FILENAME", nargs="+", help="The rule YAML files")
+    return parser.parse_args()
+
+
+if __name__ == "__main__":
+    args = parse_args()
+    renderer = HtmlOutput(args.product, args.build_dir)
+    renderer.process_rules(args.rule, args.component)
+    renderer.output_results(args)
--- a/utils/render-rules.py
+++ b/utils/render-rules.py
@@ -3,6 +3,7 @@
 from glob import glob
 import collections
 import os
+import re
 import pathlib
 
 import argparse
@@ -30,7 +31,10 @@ class HtmlOutput(template_renderer.Renderer):
             if rule.id_ in p.selected:
                 rule.in_profiles.append(p)
 
-    def process_rules(self, rule_files, component):
+    def _resolve_var_substitutions(self, rule):
+        rule.description = re.sub('<sub\s+idref="([^"]*)"\s*/>', r"<tt>$\1</tt>", rule.description)
+
+    def process_rules(self, rule_files):
         rules = []
         profiles = self._get_all_compiled_profiles()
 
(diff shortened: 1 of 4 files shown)


jan-cerny:
Hi @matejak. I haven't reviewed the code but I have a problem with the word component. I assume it's the component in a context of Bugzilla that's used in RHEL or Fedora. Other users might be confused. Anyway, it's used only as a header in the generated HTML. And I see that in future you might want to generate something different. For example, you might want to have a list of rules configuring user authentication. Therefore I propose to remove `--component` option and introduce `--header` instead. Them, you will be able to run eg:
```
./render-rules.py --header "Rules related to Rsyslog"
./render-rules.py --header "Rules using OVAL yamlfilecontent"
./render-rules.py --header "Rules configuring cryptograhy"
```
WDYT?

utils/rules-template.html
```diff
@@ -0,0 +1,36 @@
...
+</head>
+<body>
+
+<h1>Rules related to {{{ component }}}</h1>
+
+<div class="rules">
+{{% for rule in rules -%}}
+<div class="one_rule">
+	<h2>{{{ loop.index }}}. {{{ rule.title }}}</h2>
+  <p>{{{ rule.description | safe }}}</p>
```
yuumasato, line 17: I noticed one issue when rendering a rule description that references a variable, for example [accounts_password_pam_difok](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/accounts/accounts-pam/password_quality/password_quality_pwquality/accounts_password_pam_difok/rule.yml#L7).

It doesn't render any text as the variable's default value, and may cause any text to be subscript.
```html
The pam_pwquality module's <tt>difok</tt> parameter sets the number of characters
in a password that must not be present in and old password during a password change.
<br><br>
Modify the <tt>difok</tt> setting in <tt>/etc/security/pwquality.conf</tt>
to equal <sub idref="var_password_pam_difok"> to require differing characters
when changing passwords.</sub>
```

dahaic:
@matejak can you provide some obvious example for me? How it works, and what is the intermediate benefit, and what are the long term plans with it? I am a bit confused by the mentions of "component metadata".

matejak:
> @matejak can you provide some obvious example for me? How it works, and what is the intermediate benefit, and what are the long term plans with it? I am a bit confused by the mentions of "component metadata".
> 
> ...

This PR just introduces a rendering tool for manually selected rules - no new metadata are involved.
It first started with the ambition to produce output tailored for component maintainers, but it is a general rule renderer in its current form.

The use case is simply to present rules to people in the HTML form. Obviously, having rule-component mapping would make generation of that tailored output much easier, but it would introduce its own problems, so it is not in scope of this PR.

utils/template_renderer.py
```diff
@@ -0,0 +1,104 @@
...
+            print(result)
+        else:
+            with open(args.output, "w") as outfile:
+                outfile.write(result)
+
+    @staticmethod
+    def create_parser(description):
+        parser = argparse.ArgumentParser(description=description)
+        parser.add_argument(
+            "product", help="The product to be built")
```
jan-cerny, line 100: The missing dashes cause a problem that product is a positional argument when it should be optional.

```suggestion
            "--product", help="The product to be built")
```

utils/template_renderer.py
```diff
@@ -0,0 +1,104 @@
...
+            self.template_data["title"] = args.title
+        result = self.get_result()
+        if not args.output:
+            print(result)
+        else:
+            with open(args.output, "w") as outfile:
+                outfile.write(result)
+
+    @staticmethod
+    def create_parser(description):
```
jan-cerny, line 97: What is the benefit of having argument parser split into 2 places ie 2 different files?

utils/render-rules.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class HtmlOutput(template_renderer.Renderer):
+    TEMPLATE_NAME = "rules-template.html"
+
+    def _get_all_compiled_profiles(self):
+        compiled_profiles = glob(str(self.built_content_path / "profiles" / "*.profile"))
+        profiles = []
+        for p in compiled_profiles:
+            profiles.append(ssg.build_yaml.Profile.from_yaml(p))
```
jan-cerny, line 25: could be a generator

utils/render-rules.py
```diff
@@ -0,0 +1,65 @@
...
+        return profiles
+
+    def _set_rule_profiles_membership(self, rule, profiles):
+        rule.in_profiles = []
+        for p in profiles:
+            if rule.id_ in p.selected:
+                rule.in_profiles.append(p)
+
+    def _resolve_var_substitutions(self, rule):
+        rule.description = re.sub('<sub\s+idref="([^"]*)"\s*/>', r"<tt>$\1</tt>", rule.description)
```
jan-cerny, line 35: it requires a comment

jan-cerny (commented):
It's really amazing, I have only very small nitpick.

jan-cerny:
> The use case is simply to present rules to people in the HTML form. Obviously, having rule-component mapping would make generation of that tailored output much easier, but it would introduce its own problems, so it is not in scope of this PR.

With a more advanced template we could even generate HTML guides directly instead of using `oscap xccdf generate` feature.

utils/template_renderer.py
matejak: Because one part of the argument parsing will be the same for all renderers of compiled product data. Moreover, this allows `output_results` to make assumptions about a set of parsed arguments.

utils/template_renderer.py
matejak: Why? The product is required, so why should it be an optional argument?

utils/render-rules.py
matejak: I don't understand this suggestion, please expand it.

utils/render-rules.py
matejak: Do you mean a code comment, or that the generated HTML should feature a tooltip or something like that?

jan-cerny: A code comment

utils/render-rules.py
jan-cerny: For example it can be a list comprehension but it's OK
 https://www.w3schools.com/python/python_lists_comprehension.asp

utils/template_renderer.py
jan-cerny: Because you already have another positional argument and having 2 will be confusing

utils/template_renderer.py
jan-cerny: OK

utils/render-rules.py
matejak: I see, that's correct.

utils/template_renderer.py
matejak: I disagree, it's one product followed by whatever, in this case, that would be a set of one or more rules. The help message is rendered correctly, and what you propose is a required optional argument, which is [somehow against established conventions](https://docs.python.org/3/library/argparse.html#required) (see the box with the note).

utils/render-rules.py
matejak: I get it, that's a meaningful requirement.

matejak:
> With a more advanced template we could even generate HTML guides directly instead of using `oscap xccdf generate` feature.

I agree, I even think that we could technically generate XCCDF without much additional effort needed, and the only thing preventing the datastream generation is just the structure of OVAL.

jan-cerny:
I think it's fine now. Kodus.

PR #6827, opened 2021-04-13
RHCOS4: Add relevant instructions and e2e test for banner_etc_issue

JAORMX:
It was missing.

Note that this also changes the OVAL to take `/etc/issue.d` into account and look for the
legal notice there too.


jhrozek:
The E2E tests are failing with:
```
[31mERRO[0m[2021-04-14T12:03:10Z] 'ascii' codec can't encode character '\u2026' in position 87: ordinal not in range(128) 
```
this seems like some python utf8 issue, but we don't use python anywhere in our e2e tests do we? is this coming from prow itself?

jhrozek (approved):
The PR itself looks good FWIW

PR #6831, opened 2021-04-13
SLES-15-010220 updates for firewalld

brett060102:
SuSEfirewall2 was retired in SLE-12 in favor of firewalld
So, we need to remove SuSEfirewall2 references from SL15
and fix up the SLES-15-010220 rule. This includes change SLES-15-010370
to SLES-15-010220. SLES-15-010370 was specific to the firewall panic switch.

#### Description:

- Fix SLES-15-010220 
We found out about SuSEfirewall2 to firewalld change after initial SLE15 work.

#### Rationale:

- Use SLE-15 preferred firewall package.

### DIFF
--- a/linux_os/guide/system/network/network-firewalld/firewalld_activation/package_firewalld_installed/rule.yml
+++ b/linux_os/guide/system/network/network-firewalld/firewalld_activation/package_firewalld_installed/rule.yml
@@ -15,11 +15,15 @@ identifiers:
     cce@rhel7: CCE-82999-4
     cce@rhel8: CCE-82998-6
     cce@rhcos4: CCE-82521-6
+    cce@sle15: CCE-85698-9
 
 references:
     nist: CM-6(a)
+    nist@sle15: CM-7,CM-7.1(iii),CM-7(b),AC-17(1)
     srg: SRG-OS-000480-GPOS-00227,SRG-OS-000298-GPOS-00116
+    srg@sle15: SRG-OS-000096-GPOS-00050,SRG-OS-000297-GPOS-00115,SRG-OS-000480-GPOS-00232
     cis@rhel8: 3.4.1.1
+    stigid@sle15:  SLES-15-010220
 
 ocil_clause: 'the package is not installed'
 
--- a/sle15/profiles/stig.profile
+++ b/sle15/profiles/stig.profile
@@ -175,7 +175,6 @@ selections:
     - gui_login_dod_acknowledgement
     - installed_OS_is_vendor_supported
     - install_smartcard_packages
-    - kernel_module_dccp_disabled
     - kernel_module_usb-storage_disabled
     - mount_option_home_nosuid
     - mount_option_noexec_remote_filesystems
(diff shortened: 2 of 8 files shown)


brett060102:
@abergmann Can you please review this?

sle15/profiles/stig.profile
```diff
@@ -190,9 +189,8 @@ selections:
     - package_aide_installed
     - package_audit-audispd-plugins_installed
     - package_audit_installed
-    - package_SuSEfirewall2_installed
     - package_telnet-server_removed
-    - package_vsftpd_removed
```
brett060102, line 195: Not sure why this is showing up as a deleted line. I don't see it in the current sle15/profiles/stig.profile

ggbecker: I can see it here: https://github.com/ComplianceAsCode/content/blob/65885c18f89a2696358abf015eb98cb7e25dcff7/sle15/profiles/stig.profile#L195

You probably need to rebase your branch

brett060102: Could be, but this change was started https://github.com/ComplianceAsCode/content/master

brett060102: Fixed, I do see it in my remotes now, not sure why not before. I am religious about remote updates.

brett060102 (commented):
@ggbecker should be good now.

linux_os/guide/system/network/network-firewalld/firewalld_activation/package_firewalld_installed/rule.yml
```diff
@@ -15,11 +15,15 @@ identifiers:
...
     cce@rhcos4: CCE-82521-6
+    cce@sle15: CCE-85698-9
 
 references:
     nist: CM-6(a)
+    nist@sle15: CM-7,CM-7.1(iii),CM-7(b),AC-17(1)
     srg: SRG-OS-000480-GPOS-00227,SRG-OS-000298-GPOS-00116
+    srg@sle15: SRG-OS-000096-GPOS-00050,SRG-OS-000297-GPOS-00115,SRG-OS-000480-GPOS-00232
     cis@rhel8: 3.4.1.1
+    stigid@sle15:  SLES-15-010220
```
ggbecker, line 26: The same STIG id is assigned to `service_firewalld_enabled`. That's not ideal and we already requested DISA to split these for RHEL8. That's probably not going to be reflected to the SLE STIGs. Having a duplicate means your STIG results will map to only one of these if you import results into STIG Viewer for example. On the other hand, for openscap html reports, there will be no negative impact.

ggbecker: I'm not opposed to merging this, I just want to have your acknowledgement.

brett060102: acknowledged and agreed. Not quite sure what else to do given our stigs. it is better than what we have in SLE-12 where we have three rules attached to the same stig.

brett060102: @ggbecker let's go forward as it is. I was told that getting changes in the STIGs takes a while. We only have two STIGs related to firewalld in SLE-15 SLES-15-010220 and SLE-15 SLES-15-010370.
SLE-15 SLES-15-010220 really covers configuration which we attached to install as well, since we can't configure what is not installed.

SLE-15 SLES-15-010370 covers testing the panic feature to disable remote access. Not sure how to test that one. Direction really requires being attached to system console.

ggbecker (approved):

PR #6833, opened 2021-04-14
Extend profile_tool.py with option to check for missing OSPP and CUI references

mildas:
#### Description:
`profile_tool.py` already has options to check missing references for ANSSI, CIS, STIG, and HIPAA references. This extends the script with OSPP and CUI missing references check.

#### Rationale:
We want to have test that will check for missing references in rules. The `profile_tool.py` script will be used in the test and thus to be able to check OSPP and CUI missing references, the script needs to have option to check those references.


matusmarhefka:
@mildas Works great, thank you.

PR #6834, opened 2021-04-14
RHCOS4: Use RHEL8 OVAL feed and allow usage of `security_patches_up_to_date` rule

JAORMX:
RHCOS4 bases its packages on RHEL 8, so it may use RHEL 8's OVAL feed
for checking that the security patches are up to date.

This also adds rhcos4 to that rule's platforms so we may enable it if
needed.


jhrozek:
Trying out this rule gives me a MANUAL result. Do I need the CO patches as well for the rule results to be correctly evaluated?

JAORMX:
@jhrozek that's cause I messed up this PR. let me fix it.

JAORMX:
Fixed.

PR #6836, opened 2021-04-14
Add OCIL to sshd_limit_user_access

jhrozek:
#### Rationale:

OCP4 uses OCIL as instructions for manual checks, so it's prudent to
have OCIL clauses for all rules.


jhrozek:
I think adding the OCIL makes sense no matter what, but at the same time I wonder if we need this rule in the RHCOS4 moderate profile since normally you shouldn't manage users on RHCOS...
@JAORMX what do you think?

Issue #6791, opened 2021-04-02
Bad CIS-CSC URI? 


cipherboy:
#### Description of problem:

In the shared constants section of the xslt's there's the following:

> shared/transforms/shared_constants.xslt:32:<xsl:variable name="cis-cscuri">https://www.cisecurity.org/wp-content/uploads/2017/03/Poster_Winter2016_CSCs.pdf</xsl:variable>

I'm not sure what the correct URI is, but perhaps there is a better URL? Maybe https://www.cisecurity.org/controls/ ? 

If so, I can file a PR else someone else with a better URI can do it. 

#### SCAP Security Guide Version:

`master`

PR #6837, opened 2021-04-14
Fix CIS-CSC URI

cipherboy:
#### Description:

CIS-CSC: Center for Internet Security Critical Security Controls
for Effective Cyber Defense -- appears to have been renamed to
just CIS Controls and is now present at a new URL:

    https://www.cisecurity.org/controls/

Update ours to match rather than linking to a (now dated) poster.

Resolves: #6791


yuumasato (approved):
LGTM.
It seems that the proposed URL will always point to latest resources.

A side note, it doesn't seem to be called "Critical Security Controls" anymore, just Controls...
Should we change all `cis-cscuri` references? And to what?
I guess this actually should go to another PR?

cipherboy:
I mean, we could also update the references in the [rule.yml](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/ssh/ssh_server/sshd_set_keepalive/rule.yml#L51)...

cipherboy:
I've filed #6855 to do at a later point in time. Thanks @yuumasato!

Issue #6844, opened 2021-04-15
RHEL8 STIG rules - missing IDs


mildas:
#### Description of problem:
Some rules from rhel8 stig profile miss an ID.

List of the rules:
```
*** rules of 'stig' profile missing STIG IDs: 211 of 251 have them [15% missing]
   auditd_data_retention_space_left_action      
   configure_bind_crypto_policy                    configure_crypto_policy                      
   configure_kerberos_crypto_policy             
   configure_libreswan_crypto_policy            
   configure_openssl_crypto_policy              
   configure_ssh_crypto_policy                  
   dconf_gnome_banner_enabled                      enable_dracut_fips_module                    
   grub2_admin_username                            grub2_uefi_admin_username                    
   no_empty_passwords                           
   package_abrt-addon-ccpp_removed              
   package_abrt-addon-kerneloops_removed        
   package_abrt-addon-python_removed               package_abrt-cli_removed                     
   package_abrt-plugin-logger_removed           
   package_abrt-plugin-rhtsupport_removed       
   package_abrt-plugin-sosreport_removed           package_aide_installed                       
   package_audit_installed                      
   package_fapolicyd_installed                  
   package_firewalld_installed                  
   package_openssh-server_installed             
   package_rng-tools_installed                     package_tmux_installed                       
   package_usbguard_installed                      partition_for_var_tmp                        
   require_emergency_target_auth                
   set_password_hashing_algorithm_systemauth    
   sshd_disable_empty_passwords                    sshd_disable_kerb_auth                       
   sshd_set_keepalive_0                            sshd_use_strong_rng                          
   sysctl_crypto_fips_enabled                   
   sysctl_net_ipv4_conf_default_accept_source_route
   sysctl_net_ipv6_conf_all_accept_redirects    
   sysctl_net_ipv6_conf_all_accept_source_route 
   sysctl_net_ipv6_conf_default_accept_redirects
   sysctl_net_ipv6_conf_default_accept_source_route                 
```

#### SCAP Security Guide Version:
`master`

#### Steps to Reproduce:

`$ python3 build-scripts/profile_tool.py stats --benchmark build/ssg-rhel8-xccdf.xml --profile ospp --missing-ospp-refs --skip-stats`

#### Actual Results:
Rules miss IDs.

#### Expected Results:
No rule misses ID.

#### Additional Information/Debugging Steps:
[...]

PR #6849, opened 2021-04-15
Missing references test

mildas:
#### Description:
Add test for checking missing references in profiles.

Usage:
`$ tests/missing_refs.sh python3 build/ssg-rhel8-xccdf.xml ospp stig hipaa`

$1 - python executable (for cmake, `${PYTHON_EXECUTABLE}`)
$2 - product xccdf
$3+ - list of profiles

#### Rationale:
Test that will be added to gating.

### DIFF
--- a/tests/missing_refs.sh
+++ b/tests/missing_refs.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+PYTHON_EXECUTABLE="$1"
+BENCHMARK="$2"
+shift 2
+PROFILES=()
+while test $# -gt 0; do
+    PROFILES+=("$1")
+    shift
+done
+
+PROJECT_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")" && cd .. && pwd)"
+RETURN_CODE=0
+
+function check_missing_references() {
+    profile="$1"
+    if [[ "$profile" == "stig" ]]; then
+        refs_argument="--missing-stig-ids"
+    elif [[ "$profile" == "anssi"* ]]; then
+        refs_argument="--missing-anssi-refs"
+    else
+        refs_argument="--missing-$profile-refs"
+    fi
+
+    profile_stats="$("$PYTHON_EXECUTABLE" "$PROJECT_ROOT/build-scripts/profile_tool.py" stats --benchmark "$BENCHMARK" --profile $profile $refs_argument --skip-stats)"
+
+    if [ ! -z "$profile_stats" ]; then
+        printf '%s\n' "$profile_stats" >&2
+        RETURN_CODE=1
+    fi
+}
+
+for PROFILE in ${PROFILES[@]}; do
+    check_missing_references "$PROFILE"
+done
+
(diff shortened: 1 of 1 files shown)


mildas:
List of rules missing references, output from the test:
```55/73 Test #56: missing-references-ssg-rhel7-xccdf.xml ..............................***Failed    1.64 sec
*** rules of 'stig' profile missing STIG IDs: 241 of 252 have them [4% missing]
   audit_rules_login_events_tallylog            
   audit_rules_privileged_commands_sudoedit     
   configure_firewalld_rate_limiting               dconf_db_up_to_date                          
   dconf_gnome_disable_automount                
   dconf_gnome_disable_automount_open           
   dconf_gnome_disable_autorun                  
   mount_option_dev_shm_nodev                   
   mount_option_dev_shm_nosuid                     package_aide_installed                       
   xwindows_runlevel_target                     
*** rules of 'hipaa' profile missing HIPAA Refs: 142 of 143 have them [0% missing]
   dconf_db_up_to_date            

68/73 Test #69: missing-references-ssg-rhel8-xccdf.xml ..............................***Failed    1.37 sec
*** rules of 'hipaa' profile missing HIPAA Refs: 134 of 137 have them [2% missing]
   configure_crypto_policy                      
   configure_ssh_crypto_policy                     dconf_db_up_to_date          
```
@ggbecker Is the rhel7 stig doable? Or is it same as you mentioned in https://github.com/ComplianceAsCode/content/issues/6844#issuecomment-820429384 and I should exclude it for now and create issue?

mildas:
I will split this PR into two PRs - this one will add only the test and the new PR will add it to `SSGCommon.cmake`

jan-cerny:
Have you considered making this an CMake macro instead of a bash script to avoid having multiple abstraction levels?

tests/missing_refs.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+PYTHON_EXECUTABLE="$1"
+BENCHMARK="$2"
+PROFILES="$3"
```
jan-cerny, line 4: this is a risk of quoting problems

what about making every argument starting from 3 a profile? eg. use shift

mildas: Aha, that's much better approach. Thanks, I will update it

mildas:
@jan-cerny No. Actually, I didn't think about CMake macro approach at all because it didn't occur to me.
With the macro, we could avoid multiple abstraction levels, but do you think it will be readable? I mean, if implementing this in CMake macro won't be unnecessarily complicated when it can be easily done in Bash script + other tests are in a form of standalone script and this one will be as a macro.

jan-cerny:
I think it would be readable

> other tests are in a form of standalone script and this one will be as a macro.

That's a good argument against it.

mildas:
@jan-cerny What do you think? Should I implement it as CMake macro or are you fine with current approach?

jan-cerny:
@mildas I'm fine with either way.

PR #6856, opened 2021-04-16
Updating RHEL8 STIG profile for readability changes

carlosmmatos:
Direct sync of V1R1 for rules - along with current mapped rules.

PR #6858, opened 2021-04-19
Remove harden_sshd_crypto_policy from RHEL8 STIG profile

ggbecker:
#### Description:

- Remove harden_sshd_crypto_policy from RHEL8 STIG profile.

#### Rationale:

- Follow up from https://github.com/ComplianceAsCode/content/pull/6856#discussion_r615735419


carlosmmatos:
LGTM - I don't need that CI that always breaks to pass.. I know this is simple

PR #6860, opened 2021-04-19
add rule for disabling of GUI

vojtapolasek:
#### Description:

- add new rule xwindows_remove_packages_and_set_runlevel
  - the rule removes appropriate packages based on rhel version and also configures correct run level
- add this rule to rhel7 and rhel8 stig profiles

#### Rationale:

- stig effort


linux_os/guide/services/xwindows/disabling_xwindows/xwindows_remove_packages_and_set_runlevel/ansible/shared.yml
```diff
@@ -0,0 +1,25 @@
...
+# reboot = true
+# strategy = restrict
+# complexity = low
+# disruption = low
+
+- name: Ensure xorg packages are removed
+  package:
+    name: "{{ item }}"
+    state: absent
+  loop:
```
carlosmmatos, line 11: I would urge against using loops in Ansible yum package modules.. Instead, you should opt for passing in to name for efficiencies:
```
- name: Ensure xorg packages are removed
  package:
    name:
      - xorg-x11-server-Xorg
      - xorg-x11-server-common
      - xorg-x11-server-utils
{{% if product not in ["rhel7", "ol7"] %}}
      - xorg-x11-server-Xwayland
{{% endif %}}
    state: absent
```
see: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/yum_module.html#notes

carlosmmatos (changes requested):

carlosmmatos (approved):

carlosmmatos:
LGTM

PR #6861, opened 2021-04-19
Add rules related to permissions of `/var/log` and `/var/log/messages`

ggbecker:
#### Description:

- Add rules related to permissions of `/var/log` and `/var/log/messages`

@guangyee I renamed one of the SLE rules related `/var/log/messages` to this as I think it's more appropriate, see: a9d4350


JAORMX (approved):
these would be nice to have in the RHCOS4 moderate profile

PR #6862, opened 2021-04-19
Creating new RHEL 8 STIG GUI profile

carlosmmatos:
#### Description:

New RHEL 8 STIG with GUI profile. Tackles issues that allow existing customers to utilize the RHEL 8 STIG during install, and still maintain a graphical env if they chose to have one.

### DIFF
--- a/rhel8/profiles/stig_gui.profile
+++ b/rhel8/profiles/stig_gui.profile
@@ -11,7 +11,7 @@ title: 'DISA STIG with GUI for Red Hat Enterprise Linux 8'
 
 description: |-
     This profile contains configuration checks that align to the
-    DISA STIG with GUI for Red Hat Enterprise Linux 8.
+    DISA STIG with GUI for Red Hat Enterprise Linux 8 V1R1.
 
     In addition to being applicable to Red Hat Enterprise Linux 8, DISA recognizes this
     configuration baseline as applicable to the operating system tier of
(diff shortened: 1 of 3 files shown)


rhel8/profiles/stig_gui.profile
```diff
@@ -0,0 +1,36 @@
...
+    SMEs:
+        - carlosmmatos
+
+reference: https://public.cyber.mil/stigs/downloads/?_dl_facet_stigs=operating-systems%2Cunix-linux
+
+title: 'DISA STIG with GUI for Red Hat Enterprise Linux 8'
+
+description: |-
+    This profile contains configuration checks that align to the
+    DISA STIG with GUI for Red Hat Enterprise Linux 8.
```
ggbecker, line 14: I think we have missed the versioning for RHEL8 STIG profile. We should have `V1R1` stated in this line.
```suggestion
    DISA STIG with GUI for Red Hat Enterprise Linux 8 V1R1.
```

Would you care to change in the original stig.profile as well? On the other hand V1R2 is about to be released.

carlosmmatos: Roger that.

yuumasato (approved):
LGTM.
But I leave the final ack to @ggbecker

ggbecker (approved):

PR #6863, opened 2021-04-19
Creating new RHEL 7 STIG GUI profile

carlosmmatos:
#### Description:

New RHEL 7 STIG with GUI profile. Tackles issues that allow existing customers to utilize the RHEL 7 STIG during install, and still maintain a graphical env if they chose to have one.

### DIFF
--- a/rhel7/profiles/stig_gui.profile
+++ b/rhel7/profiles/stig_gui.profile
@@ -1,0 +1,35 @@
+documentation_complete: true
+
+metadata:
+    version: V3R1
+    SMEs:
+        - carlosmmatos
+
+reference: https://public.cyber.mil/stigs/downloads/?_dl_facet_stigs=operating-systems%2Cunix-linux
+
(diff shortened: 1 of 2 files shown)


rhel7/profiles/stig_gui.profile
```diff
@@ -0,0 +1,35 @@
+documentation_complete: true
+
+metadata:
+    version: V3R1
```
yuumasato, line 4: Update the version in the metadata too.

yuumasato: It is still V1R1, :grimacing:

carlosmmatos: hahaha my bad.. was looking at the wrong profile...whoops

ggbecker (approved):
LGTM. I hope profile id: `stig_gui` is good enough and will not require to be changed later :D

PR #6864, opened 2021-04-19
RHEL8 - ensuring stigid's and references are set where appropriate

carlosmmatos:
Cleaning up/adding RHEL8 stigid's and other disa refs.

I understand that we will have some duplicate STIG ID's, but we'll just have to deal with that for now until we either:

- Get DISA to update it OR
- provide a temporary rule that accounts for the multiple rules


carlosmmatos:
@ggbecker - I know we talked about the issues with having duplicate stigid's but what is the current alternative to this approach? Is it better to exclude a stigid on a rule than to have it duplicated? Should we tackle these as one-off stig rules that combine both?

linux_os/guide/system/auditing/service_auditd_enabled/rule.yml
```diff
@@ -55,7 +55,7 @@ references:
     srg@sle12: SRG-OS-000037-GPOS-00015,SRG-OS-000038-GPOS-00016,SRG-OS-000039-GPOS-00017,SRG-OS-000040-GPOS-00018,SRG-OS-000041-GPOS-00019,SRG-OS-000042-GPOS-00021,SRG-OS-000051-GPOS-00024,SRG-OS-000054-GPOS-00025,SRG-OS-000122-GPOS-00063,SRG-OS-000254-GPOS-00095,SRG-OS-000255-GPOS-00096,SRG-OS-000392-GPOS-00172,SRG-OS-000480-GPOS-00227
     disa@sle12: CCI-000130,CCI-000131,CCI-000132,CCI-000133,CCI-000134,CCI-000135,CCI-000154,CCI-000158,CCI-000366,CCI-001464,CCI-001487,CCI-001876,CCI-002884
     nist@sle12: AU-3,AU-3(1),AU-3(1).1(ii),AU-3.1,AU-6(4),AU-6(4).1,AU-7(1),AU-7(1).1,AU-7(a),AU-14(1),AU-14(1).1,CM-6(b),CM-6.1(iv),MA-4(1)(a)
-    stigid@rhel8: RHEL-08-010560
+    stigid@rhel8: RHEL-08-030180
```
ggbecker, line 58: I'd say the older one makes more sense in this rule: https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_8_security_technical_implementation_guide/V-230297

linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/rule.yml
```diff
@@ -47,7 +47,7 @@ references:
     cobit5: APO01.06,DSS05.04,DSS05.05,DSS05.07,DSS05.10,DSS06.02,DSS06.03,DSS06.10
     iso27001-2013: A.10.1.1,A.11.1.4,A.11.1.5,A.11.2.1,A.13.1.1,A.13.1.3,A.13.2.1,A.13.2.3,A.13.2.4,A.14.1.2,A.14.1.3,A.18.1.4,A.6.1.2,A.7.1.1,A.7.1.2,A.7.3.1,A.8.2.2,A.8.2.3,A.9.1.1,A.9.1.2,A.9.2.1,A.9.2.2,A.9.2.3,A.9.2.4,A.9.2.6,A.9.3.1,A.9.4.1,A.9.4.2,A.9.4.3,A.9.4.4,A.9.4.5
     cis-csc: 1,12,13,14,15,16,18,3,5
-    stigid@rhel8:  sshd_disable_empty_passwords
```
ggbecker, line 50: How could I do that? :D   probably some copy/paste issue

carlosmmatos: Lol I thought it was some magical way of referencing a rule

linux_os/guide/system/auditing/service_auditd_enabled/rule.yml
carlosmmatos: sounds good, I'll fix now

ggbecker (approved):

PR #6865, opened 2021-04-20
remove no longer relevant rules from rhel7 stig

vojtapolasek:
#### Description:

- remove three rules which seem no longer used

#### Rationale:

- reference taken from https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide?version=V3R2
- these rules were discovered when checking for rules without proper stig references

### DIFF
--- a/rhel7/profiles/stig.profile
+++ b/rhel7/profiles/stig.profile
@@ -203,7 +203,6 @@ selections:
     - audit_rules_execution_setsebool
     - audit_rules_execution_chcon
     - audit_rules_execution_setfiles
-    - audit_rules_login_events_tallylog
     - audit_rules_login_events_faillock
     - audit_rules_login_events_lastlog
     - audit_rules_privileged_commands_passwd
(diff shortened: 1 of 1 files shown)


rhel7/profiles/stig.profile
```diff
@@ -268,7 +266,6 @@ selections:
     - sshd_disable_compression
     - sshd_disable_user_known_hosts
     - chronyd_or_ntpd_set_maxpoll
-    - configure_firewalld_rate_limiting
```
ggbecker, line 271: this has a historical event: https://github.com/ComplianceAsCode/content/pull/5168

carlosmmatos: @vojtapolasek @ggbecker Please remove it as it's no longer relevant. If we want to re-add this, we need to go through the proper channels and communicate this to DISA.

ggbecker (approved):
LGTM

PR #6866, opened 2021-04-20
add sudo_restrict_privilege_elevation_to_authorized to rhel7 and rhel8 stig

vojtapolasek:
#### Description:

- update prodtype, references and identifiers for the rule
- fix wrong sle15 reference
- make tests applicable
- add rule to profiles

#### Rationale:

stig effort


JAORMX (approved):

linux_os/guide/system/software/sudo/sudo_restrict_privilege_elevation_to_authorized/rule.yml
```diff
@@ -18,11 +21,18 @@ severity: medium
...
 
 references:
     nist: CM-6(b),CM-6(iv)
     disa@sle15: CCI-000366
-    stig@sle15: SLES-15-020101
+    stigid@sle15: SLES-15-020101
+    disa@rhel7: CCI-000366
+    stigid@rhel7: CCI-000366
+    srg: SRG-OS-000480-GPOS-00227
+    disa@rhel8: CCI-000366
```
yuumasato, line 34: I was discussing with @ggbecker that the `disa` CCI reference is product independent, much like the `srg`.
Reference: https://public.cyber.mil/stigs/cci/

What do you think of removing the product specification?

vojtapolasek: No problem, I will fix it. Just note that we have about 100 cases of disa@something in the content. That's why I was not sure and I rather took this approach.

carlosmmatos: I'll also chime in on this subject since I've seen this recently... in the docs https://complianceascode.readthedocs.io/en/latest/manual/developer/06_contributing_with_content.html#rules

we don't mention the ability to use such a naming convention for `disa` refs, so how did this come to fruition? The only reference we allow such a convention by our docs is the `stigid@product`..

ggbecker: I guess every reference can have a product qualifier, the build system does not prevent this. And that probably happened when someone is contributing and wants to make sure that it doesn't affect other products unintentionally.

JAORMX (approved):

tests/data/profile_stability/rhel8/stig.profile
```diff
@@ -247,6 +247,7 @@ selections:
 - sssd_offline_cred_expiration
 - sudo_remove_no_authenticate
 - sudo_remove_nopasswd
+- sudo_restrict_privilege_elevation_to_authorized
```
ggbecker, line 250: please rebase this PR and add the same rule to the RHEL8 STIG with GUI profile, which is a new profile:

`tests/data/profile_stability/rhel8/stig_gui.profile`

ggbecker (changes requested):
Update profile stability data in stig_gui profile

vojtapolasek:
fixed

ggbecker:
Please resolve the conflicts so we can merge it.

vojtapolasek:
I rebased.

linux_os/guide/system/software/sudo/sudo_restrict_privilege_elevation_to_authorized/rule.yml
```diff
@@ -18,11 +21,17 @@ severity: medium
...
+    cce@rhel7: CCE-83423-4
+    cce@rhel8: CCE-83425-9
+
 
 references:
     nist: CM-6(b),CM-6(iv)
     disa: CCI-000366
-    stig@sle15: SLES-15-020101
+    stigid@sle15: SLES-15-020101
+    stigid@rhel7: CCI-000366
```
ggbecker, line 32:
```suggestion
    stigid@rhel7: RHEL-07-010341
```

ggbecker (approved):

PR #6867, opened 2021-04-20
Add PCI-DSS profile to RHV4

yuumasato:
#### Description:

-  Add PCI-DSS profile based on RHEL8 profile
- Remove libreswan from the profile, as it is not available for RHVH.


yuumasato:
Thanks @carlosmmatos

PR #6873, opened 2021-04-21
Add anaconda remediation for xwindows_remove_packages

ggbecker:
#### Description:

- Add anaconda remediation for xwindows_remove_packages

#### Rationale:

- This is required so oscap anaconda addon can preprocess the list of packages being removed and decide whether to prevent users from continue installation or not. In case they select Server with Gui and default STIG profile (which removes xorg packages and cause conflict in the installation)

It's not possible to add the agent switch the same as [bash](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/xwindows/disabling_xwindows/xwindows_remove_packages/bash/shared.sh#L17
) and ansible remediation are doing. But this does not pose a problem since the bash remediation will also get executed later in the installation phase.


ggbecker:
I'm going to self merge this as soon as the build/tests finish so it doesn't leave the tests in the main branch broken.

carlosmmatos:
LGTM

PR #6874, opened 2021-04-21
Fix RHEL8 STIG with GUI stable profile data.

ggbecker:
#### Description:

- Fix RHEL8 STIG with GUI stable profile data.

#### Rationale:

- Hotfix to the current main branch. `stable-profiles` test is failing without this.


ggbecker:
Self merging this as it unblocks the main branch testing

PR #6879, opened 2021-04-22
SLES-15-010250 add rule, remediation and tests.

teacup-on-rockingchair:
#### Description:

- Set PAM's Common Authentication Hashing Algorithm

#### Rationale:

- Add oval checks for common pam authentication to be based on sha512 and pam_unix.so module to be required
- Add ansible and bash remediation, bash one based on the ensure_pam_module_options macro
- Add 3 basic tests to create valid and two options of invalid configuration in the common-auth


linux_os/guide/system/accounts/accounts-pam/set_password_hashing_algorithm/set_password_hashing_algorithm_commonauth/rule.yml
```diff
@@ -0,0 +1,64 @@
...
+    <tt>sha512</tt>:
+    <pre>$ grep sha512 /etc/pam.d/common-auth</pre>
+
+platform: pam
+
+template:
+    name: pam_options
+    vars:
+      path: /etc/pam.d/common-auth
+      type: password
```
vojtapolasek, line 59: I suppose you want to check auth section, not password section.
```suggestion
      type: auth
```

vojtapolasek (changes requested):
Thank you for this rule. Please see review comments.
Just a note, you do not need to supply Ansible remediation and OVAL check in this case, because you use template. That is the goal of templates - no need to write remediations or checks if they have many thing in common. As you can see, the pam_options template provides oval and ansible:
https://github.com/ComplianceAsCode/content/tree/master/shared/templates/pam_options
However, if you want to supply Bash remediation, you can, because the template does not provide it.
You can read more about templates here:
https://complianceascode.readthedocs.io/en/latest/manual/developer/06_contributing_with_content.html#templating

linux_os/guide/system/accounts/accounts-pam/set_password_hashing_algorithm/set_password_hashing_algorithm_commonauth/rule.yml
teacup-on-rockingchair: Thanks :bow:, should be ok in edeb2c5

linux_os/guide/system/accounts/accounts-pam/set_password_hashing_algorithm/set_password_hashing_algorithm_commonauth/rule.yml
```diff
@@ -0,0 +1,64 @@
...
+    of a strong hashing algorithm that makes password cracking attacks more
+    difficult.
+
+severity: medium
+
+identifiers:
+    cce@sle15: CCE-85754-0
+
+references:
+    disa@sle15: CCI-000803
```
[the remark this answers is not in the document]
teacup-on-rockingchair: Thanks :bow:, should be ok in edeb2c5

vojtapolasek:
Hello, thanks for fixes. Please could you remove the oval and ansible remediation, is it is supplied by the template?

linux_os/guide/system/accounts/accounts-pam/set_password_hashing_algorithm/set_password_hashing_algorithm_commonauth/rule.yml
```diff
@@ -36,7 +36,7 @@ identifiers:
     cce@sle15: CCE-85754-0
 
 references:
-    disa@sle15: CCI-000803
+    disa: CCI-000803
     nist@sle15: IA-7,IA-7.1
```
vojtapolasek, line 40: This is also product independent. Sorry for not noticing that earlier.
```suggestion
    nist: IA-7,IA-7.1
```

vojtapolasek:
Just that one more reference change and the PR is good to merge. Thank you.

vojtapolasek (approved):

PR #6880, opened 2021-04-22
Show which rule failed during template builds

cipherboy:
#### Description

Previously, when a rule failed to build when applying the template
(e.g., using a glob in a package removal template), we wouldn't
necessarily know which rule failed to build. Print an error message
showing which rule this was and then re-throw the error so that the
builder can see which rule caused the build failure.

### DIFF
--- a/ssg/templates.py
+++ b/ssg/templates.py
@@ -217,7 +217,7 @@ class Builder(object):
                 self.build_lang(
                     rule_id, template_name, template_vars, lang, local_env_yaml)
             except Exception as e:
-                print("Error building rule {0}".format(rule_id), file=sys.stderr)
+                print("Error building templated {0} content for rule {1}".format(lang, rule_id), file=sys.stderr)
                 raise e
 
     def build_extra_ovals(self):
--- a/ssg/templates.py
+++ b/ssg/templates.py
@@ -213,8 +213,12 @@ class Builder(object):
         local_env_yaml["rule_title"] = rule_title
         local_env_yaml["products"] = self.env_yaml["product"]
         for lang in langs_to_generate:
-            self.build_lang(
-                rule_id, template_name, template_vars, lang, local_env_yaml)
+            try:
+                self.build_lang(
+                    rule_id, template_name, template_vars, lang, local_env_yaml)
+            except Exception as e:
+                print("Error building rule {0}".format(rule_id), file=sys.stderr)
+                raise e
 
     def build_extra_ovals(self):
         declaration_path = os.path.join(self.templates_dir, "extra_ovals.yml")


ssg/templates.py
```diff
@@ -213,8 +213,12 @@ def build_rule(self, rule_id, rule_title, template, langs_to_generate):
         local_env_yaml["rule_title"] = rule_title
         local_env_yaml["products"] = self.env_yaml["product"]
         for lang in langs_to_generate:
-            self.build_lang(
-                rule_id, template_name, template_vars, lang, local_env_yaml)
+            try:
+                self.build_lang(
+                    rule_id, template_name, template_vars, lang, local_env_yaml)
+            except Exception as e:
+                print("Error building rule {0}".format(rule_id), file=sys.stderr)
```
yuumasato, line 220: Could the language matter?
Depending on the template vars can be specific for a language.

Suggestion for message:
```suggestion
                print("Error building templated {0} content for rule {1}".format(lang, rule_id), file=sys.stderr)
```

yuumasato (commented):
> Previously, when a rule failed to build when applying the template
> (e.g., using a glob in a package removal template),

It is not clear to me how you triggered a templated content build exception with a glob.
I tried `pkgname: tuned*list`, but the build error happened only latter, when generating the guides...
```
OpenSCAP Error: File '/home/wsato/git/content/build/ssg-rhel8-ds.xml' line 11950: Element '{http://oval.mitre.org/XMLSchema/oval-definitions-5}criterion', attribute 'test_ref': [facet 'pattern'] The value 'oval:ssg-test_package_tuned*list_removed:tst:1' is not accepted by the pattern 'oval:[A-Za-z0-9_\-\.]+:tst:[1-9][0-9]*'.
```

I was only able to trigger the exception when I introduced a syntax error like this `pkgnameinvalid: tuned`.

```
Error building rule package_tuned_removed
  File "/home/wsato/git/content/build-scripts/build_templated_content.py", line 56, in <module>
    builder.build()
  File "/home/wsato/git/content/ssg/templates.py", line 260, in build
    self.build_all_rules()
  File "/home/wsato/git/content/ssg/templates.py", line 246, in build_all_rules
    self.build_rule(
  File "/home/wsato/git/content/ssg/templates.py", line 221, in build_rule
    raise e
  File "/home/wsato/git/content/ssg/templates.py", line 217, in build_rule
    self.build_lang(
  File "/home/wsato/git/content/ssg/templates.py", line 143, in build_lang
    filled_template = ssg.jinja.process_file_with_macros(
  File "/home/wsato/git/content/ssg/jinja.py", line 178, in process_file_with_macros
    return process_file(filepath, substitutions_dict)
  File "/home/wsato/git/content/ssg/jinja.py", line 134, in process_file
    return template.render(substitutions_dict)
  File "/usr/lib/python3.8/site-packages/jinja2/environment.py", line 1090, in render
    self.environment.handle_exception()
  File "/usr/lib/python3.8/site-packages/jinja2/environment.py", line 832, in handle_exception
    reraise(*rewrite_traceback_stack(source=source))
  File "/usr/lib/python3.8/site-packages/jinja2/_compat.py", line 28, in reraise
    raise value.with_traceback(tb)
  File "/home/wsato/git/content/shared/templates/package_removed/oval.template", line 4, in top-level template code
    {{{ oval_metadata("The " + pkg_system|upper + " package " + PKGNAME + " should be removed.", affected_platforms=["multi_platform_all"]) }}}
jinja2.exceptions.UndefinedError: 'PKGNAME' is undefined
ninja: build stopped: subcommand failed.
```

cipherboy:
@yuumasato said:
>  It is not clear to me how you triggered a templated content build exception with a glob.

Hmm maybe I didn't then. I've been poking at trying to update the other with 20.04 content. One different was in the xorg package rules, where the glob would actually at the end (per CIS guidance):

```yml
    pkgname@ubuntu2004: 'xserver-xorg*'
```

Though obviously this does not work, as you mentioned. :-)

The upstream contributor added one template:

```yml
template:
  name: sshd_lineinfile
  vars:
    parameter: "ClientAliveCountMax"
    value: "0"
    value@ubuntu2004: "3"
    missing_parameter_pass: "false"
    kubernetes: "off"
```

But AFAICT that worked.

So maybe the error was in the other one I've added:

```yml
template:
    name: sudo_defaults_option
    vars:
        option: logfile
        variable_name: "var_sudo_logfile"
```

But sadly, all commit history has been squashed so I'd have to go dig through dangling refs to try and find what I did and why it broke the build before I fixed it...

ssg/templates.py
cipherboy: Ahh good idea, thanks @yuumasato!

yuumasato:
> But sadly, all commit history has been squashed so I'd have to go dig through dangling refs to try and find what I did and why it broke the build before I fixed it...

@cipherboy No need to bother then.

yuumasato (approved):

JAORMX (approved):

PR #6883, opened 2021-04-23
Simplify the policy renderer

matejak:
It can now take advantage of the boilerplate provided by other renderers.


JAORMX (approved):
Nice!

PR #6886, opened 2021-04-23
Relax version checks for Oracle JRE only

### LINKED ISSUE #6885: Oracle JRE Version match too restrictive
#### Description of problem:
The version in the OVAL for JRE has a version "equals" for 1.8.0, while Oracle is now inserting the update # into the version string, so the version for 1.8u281 is 1.8.0_281 (as noted in the system info).

The using test 'oval:ssg-test_oracle_java_rhel:tst:1` requires that the version be exactly 1.8.0 to pass. 

0.1.55

Red Hat Enterprise Linux 7

#### Steps to Reproduce:

1. Install Oracle JRE 1.8.0_281 Linux x64 RPM Package from https://www.oracle.com/java/technologies/javase-jre8-downloads.html
2. Attempt to perform an XCCDF scan with the datastream.
```bash
oscap xccdf eval --profile xccdf_org.ssgproject.content_profile_stig ssg-jre-ds.xml
```

#### Actual Results:
All evaluation entries are `notapplicable`

#### Expected Results:
None of the evaluation entries are `notapplicable`

Suggest adding a test set to bound the version to [1.8.0, 1.9.0) and include it for the JRE checks when dealing with oracle_java.

lordofhyphens:
#### Description:
- Bounds the version check to be 1.8.0 <= x < 1.9.0 instead of strictly 1.8.0

#### Rationale:

- Oracle Java RPM package reports the update version number as "1.8.0_xxx" for update versions (so 1.8 u281 -> 1.8.0_281), which fails a check.
- Fixes #6885

### DIFF
--- a/jre/checks/oval/installed_app_is_java.xml
+++ b/jre/checks/oval/installed_app_is_java.xml
@@ -20,7 +20,8 @@
 
   <linux:rpminfo_test check="all" check_existence="at_least_one_exists" comment="Oracle JRE is installed" id="test_oracle_java" version="1">
     <linux:object object_ref="obj_oracle_java" />
-    <linux:state state_ref="state_java_version" />
+    <linux:state state_ref="state_java_version_minimum" />
+    <linux:state state_ref="state_java_version_maximum" />
   </linux:rpminfo_test>
   <linux:rpminfo_object id="obj_oracle_java" version="1">
     <linux:name datatype="string" operation="pattern match">^jre.*$</linux:name>
@@ -68,5 +69,11 @@
 
   <linux:rpminfo_state id="state_java_version" version="1">
     <linux:version datatype="version" operation="equals">1.8.0</linux:version>
+  </linux:rpminfo_state>
+  <linux:rpminfo_state id="state_java_version_minimum" version="1">
+    <linux:version datatype="version" operation="greater than or equal">1.8.0</linux:version>
+  </linux:rpminfo_state>
+  <linux:rpminfo_state id="state_java_version_maximum" version="1">
+    <linux:version datatype="version" operation="less than">1.9.0</linux:version>
   </linux:rpminfo_state>
 </def-group>


yuumasato:
@openscap-ci ok to test

jre/checks/oval/installed_app_is_java.xml
```diff
@@ -69,4 +70,10 @@
   <linux:rpminfo_state id="state_java_version" version="1">
     <linux:version datatype="version" operation="equals">1.8.0</linux:version>
   </linux:rpminfo_state>
+  <linux:rpminfo_state id="state_java_version_minimum" version="1">
+    <linux:version datatype="version" operation="greater than or equal">1.8.0</linux:version>
+  </linux:rpminfo_state>
+  <linux:rpminfo_state id="state_java_version_maximum" version="1">
+    <linux:version datatype="version" operation="less than">1.9.0</linux:version>
```
cipherboy, line 77: I don't quite understand why we have this? Why do we care which JDK version they're running? Is it because our interior rules only apply on JDK8? Do we just need rules for JDK11+?

We care that, whatever major JDK version they have is the latest, but I don't quite understand why we should care which major version they're running --- and assume `test_oracle_java` only passes if it is JDK1.8.0 --- it should be independent of JDK versions :) 

My 2c. -- either test is misnamed (another PR) or these version checks should be removed altogether.

(I also don't use JRE checks so perhaps I'm completely off base here)

carlosmmatos: @cipherboy yes, it's because our current ruleset applies to JDK version 8. This entire product was based on the Oracle JRE 8 STIG, and there are plans to do another round once a new STIG for ( I assume 11+) comes out. With that being said, I agree that we should probably start cleaning up our verbiage around this product. Perhaps make it `jre8` instead of `jre`, and start updating the test names so they make more sense of what is being checked.

carlosmmatos (approved):
sorry - didn't mean to start a review

jre/checks/oval/installed_app_is_java.xml
yuumasato, line 77: @lordofhyphens Does the check as proposed work for you?

Checking the OVAL standard for `datatype="version"`, it says that "[short numbers] should be padded [to match length]".
Reference: https://oval-community-guidelines.readthedocs.io/en/latest/oval-schema-documentation/oval-common-schema.html#simpledatatypeenumeration
But I don't know if OpenSCAP scanner does that automatically, or the content author should do it.
@evgenyz @jan-cerny Do you, by chance, know something about this?

A package version of  `1.8.0_281` would be equivalent of `1.8.0.281`, which would still be greater than `1.8.0.0` and smaller than `1.9.0.0`, if automatically padded by the scanner.

yuumasato: @carlosmmatos So I guess `test_oracle_java_rhel` should be updated as well?

carlosmmatos: I don't even know if we offer/support that anymore..I guess I'm not that concerned with it as I doubt we are pumping out new packages for it. However, I'm also not opposed to just implementing the min/max check as the standard in the event that future openjdk packages start doing the same thing.

carlosmmatos: @yuumasato I've verified that this does indeed work. As you mentioned, the addition of the `_XXX` just added an additional padding to the `1.8.0` original check which is why it failed.. It would be so much nicer if the OVAL check would simply check what you supplied it to check - ie `1.8.0` instead of adding an additional padded number! Kind of like a subset check.

lordofhyphens:
Yes, the check as proposed does work on my test workstation (RHEL 7.9).

jre/checks/oval/installed_app_is_java.xml
lordofhyphens: Yeah, I didn't touch `test_oracle_java_rhel` as I don't have a testbed for that and while I believe it'd have the similar interesting things going on, I was operating under a "change as little as necessary to get my system detected".

PR #6895, opened 2021-04-26
remove product specific disa and srg references

vojtapolasek:
#### Description:

I used following commands to remove product specific disa and srg references and replace them with ones without product:
```
find linux_os/guide/ -name rule.yml -exec sed -i 's/disa@.*:/disa:/' {} \;
find linux_os/guide/ -name rule.yml -exec sed -i 's/srg@.*:/srg:/' {} \;
```

#### Rationale:

These references are not product specific. More information here:
https://public.cyber.mil/stigs/cci/
https://disa.mil/-/media/Files/DISA/News/Events/Symposium-2019/2---Kreigline_STIG-Panel-Briefing1_approved_Final.ashx

PR #6896, opened 2021-04-26
Prevent global references to use product-qualifiers

ggbecker:
#### Description:

- Prevent global references to use product-qualifiers.
  - Starts with srg and disa references.

This will definitely fail during the build because we have many of these instances.

Relates to: #6895

### DIFF
--- a/docs/manual/developer/06_contributing_with_content.md
+++ b/docs/manual/developer/06_contributing_with_content.md
@@ -252,10 +252,10 @@ A rule may contain those reference-type attributes:
     `srg`, `nist`, etc., whose keys may be modified with a product
     (e.g., `stigid@rhel7`) to restrict what products a reference
     identifier applies to. Depending on the type of reference (e.g.
-    catalog, rulei, etc.) will depend on how many can be added to a
+    catalog, ruleid, etc.) will depend on how many can be added to a
     single rule. In addition, certain references in a rule such as
-    `stigid` only apply to a certain product and product version; they
-    cannot be used for multiple products and versions
+    `stigid` or `cis` only apply to a certain product and product version; they
+    cannot be used for multiple products and versions.
 
     <table>
     <colgroup>
@@ -274,7 +274,7 @@ A rule may contain those reference-type attributes:
     </thead>
     <tbody>
     <tr class="odd">
-    <td><p>cis</p></td>
+    <td><p>cis@&lt;product&gt;&lt;product_version&gt;</p></td>
     <td><p>Center for Internet Security (catalog identifier)</p></td>
     <td><p>0-to-many, 0-to-1 is preferred</p></td>
     <td><p>5.2.5</p></td>
--- a/ssg/build_yaml.py
+++ b/ssg/build_yaml.py
@@ -1044,6 +1044,7 @@ class Rule(object):
     }
 
     PRODUCT_REFERENCES = ("stigid", "cis",)
+    GLOBAL_REFERENCES = ("srg", "disa", "cis-csc",)
 
     def __init__(self, id_):
         self.id_ = id_
@@ -1229,6 +1230,12 @@ class Rule(object):
                 continue
 
             label = full_label.split("@")[0]
+
+            if label in Rule.GLOBAL_REFERENCES:
+                raise ValueError(
+                    "You cannot use product-qualified for '{item_u}' reference. Please remove the product-qualifier and merge values with existing reference. Original line: {item_q}: {value_q}"
+                    .format(item_u=label, item_q=full_label, value_q=value))
+
             if label in items_dict and not allow_overwrites and value != items_dict[label]:
                 msg = (
                     "There is a product-qualified '{item_q}' item, "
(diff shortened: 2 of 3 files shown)


ssg/build_yaml.py
```diff
@@ -1024,6 +1024,7 @@ class Rule(object):
     }
 
     PRODUCT_REFERENCES = ("stigid", "cis",)
+    GLOBAL_REFERENCES = ("srg", "disa",)
```
cipherboy, line 1027: `cis-csc` would be another one here that could be added. It has no product-specific uses right now.

jan-cerny:
I suggest to add this to the documentation.
https://complianceascode.readthedocs.io/en/latest/manual/developer/06_contributing_with_content.html#rules

ggbecker:
> I suggest to add this to the documentation.
> https://complianceascode.readthedocs.io/en/latest/manual/developer/06_contributing_with_content.html#rules

Definitely, it's also a good opportunity to bring every reference that fits into this condition. Mainly those described in the reference's table. And we can also sync whatever is missing on both sides.

cipherboy:
@jan-cerny and @ggbecker: How does one contribute to readthedocs.io's documentation? I'm not familiar at all with that. When I view the contents under `docs/developer` it says that it is deprecated and that the instructions are up there. But I have no idea how to contribute. Do you have a documentation contributors guide for me? :D I don't think we were using rtd.io when I was an intern.

ggbecker:
> @jan-cerny and @ggbecker: How does one contribute to readthedocs.io's documentation? I'm not familiar at all with that. When I view the contents under `docs/developer` it says that it is deprecated and that the instructions are up there. But I have no idea how to contribute. Do you have a documentation contributors guide for me? :D I don't think we were using rtd.io when I was an intern.
> 
> (Asking this so I can add some of my new contributions and fix_rules.py to it as well ;-) )

You can simply edit files in this [folder](https://github.com/ComplianceAsCode/content/tree/master/docs/manual/developer) (containing former `developer_guide.adoc` split into multiple files) and they should be automatically reflected to rtd when the changes get merged. You can build the documentation locally to verify that it renders ok by following [this](https://complianceascode.readthedocs.io/en/latest/manual/developer/02_building_complianceascode.html#sphinx-packages-developer-documentation) and [this](https://complianceascode.readthedocs.io/en/latest/manual/developer/02_building_complianceascode.html#generating-sphinx-documentation)

The general idea of readthedocs integration is that if you see something there and edit the correspondent doc within CaC *.md files, it should automatically update the rtd docs.

ssg/build_yaml.py
ggbecker: added

docs/manual/developer/06_contributing_with_content.md
```diff
@@ -274,7 +274,7 @@ A rule may contain those reference-type attributes:
     </thead>
     <tbody>
     <tr class="odd">
-    <td><p>cis</p></td>
+    <td><p>cis&lt;product&gt;&lt;product_version&gt;</p></td>
```
cipherboy, line 277: Missing an `@` like `stigid` below :)

And thanks for the docs updates!

cipherboy (approved):

docs/manual/developer/06_contributing_with_content.md
ggbecker: thanks for the catch

ssg/build_yaml.py
```diff
@@ -1197,6 +1198,12 @@ def _make_items_product_specific(self, items_dict, product_suffix, allow_overwri
                 continue
 
             label = full_label.split("@")[0]
+
+            if label in Rule.GLOBAL_REFERENCES:
+                raise ValueError(
+                    "You cannot use product-qualified for '{item_u}' reference. Please remove the product-qualifier and merge values with existing reference. Original line: {item_q}: {value_q}"
```
cipherboy, line 1204: @ggbecker If you want to fix this up to pass PEP8, I'd be happy to merge it.

ggbecker: Done. Let's wait for test results.

ggbecker:
There was a rule with an offending product-qualified global reference, so I removed it and also I realized that the condition should come before testing if the qualifier matched the product being built at the moment so it catches the problem even if you are building only one specific content for example and the offending reference is from another product (see 591e528)

ssg/build_yaml.py
```diff
@@ -1044,6 +1044,7 @@ class Rule(object):
     }
 
     PRODUCT_REFERENCES = ("stigid", "cis",)
+    GLOBAL_REFERENCES = ("srg", "disa", "cis-csc",)
```
cipherboy, line 1047: Should `vmmsrg` be added here as well since you fixed it in the `set_password_hashing...` rule?

ggbecker: probably yes ;)

ggbecker: Done

ggbecker: I think we can add more in the future as we understand more about the references

PR #6904, opened 2021-04-28
OCP: Fix description and OCIL in proxy-kubeconfig rules

jhrozek:
The proxy-kubeconfig rules were sort of a special case as they don't
check ownership or file permissions of files on the hosts, but mounted
inside a pod. The rules used autogenerated description and OCIL which
didn't make sense to check files inside a pod, let's use hand-crafted
instructions instead.

Resolves: rhbz#1954572


JAORMX (approved):
will merge once it passes CI

PR #6882, opened 2021-04-23
Sort references, identifiers in rule.yml

cipherboy:
#### Description:

This PR introduces two new utilities into (now renamed) `utils/fix_rules.py`:

 - `utils/fix_rules.py duplicate_subkeys` -- checks for duplicated subkeys of `references` or `identifiers` within rules.
 - `utils/fix_rules.py sort_subkeys` -- sorts subkeys of `references` and `identifiers` once they've been deduplicated.

Additionally we:

 - Make `ssg_root` an optional parameter to `utils/fix_rules.py` and automatically infer it based on the location of the python.
 - Fix an issue with loading a subset of a YAML file from `ssg/rule_yaml.py`, due to a missing `Loader=` parameter. 

This PR also applies both utilities.

 - Two files with duplicated keys were found and manually fixed.
 - A bunch of files were sorted. :-)

I've generated both and ran the test suite afterwards a build with this patchset applied. I've got:

<details>
    <summary>Test suite execution</summary>

```
[911/911] [man-page] generating man page
real 105.69
user 2149.27
sys 179.59
Test project /home/cipherboy/GitHub/cipherboy/scap-security-guide/build
        Start   1: sanity-ssg-tables-table-macos1015-nistrefs.html
        Start   2: sanity-ssg-tables-table-macos1015-nistrefs-moderate.html
        Start   3: sanity-ssg-tables-table-macos1015-cces.html
        Start   4: sanity-ssg-tables-table-rhcos4-nistrefs.html
        Start   5: sanity-ssg-tables-table-rhcos4-nistrefs-ospp.html
        Start   6: sanity-ssg-tables-table-rhcos4-nistrefs-stig.html
        Start   7: sanity-ssg-tables-table-rhcos4-cces.html
        Start   8: sanity-ssg-tables-table-rhcos4-srgmap.html
        Start   9: sanity-ssg-tables-table-rhcos4-srgmap-flat.html
        Start  10: sanity-ssg-tables-table-ol7-nistrefs.html
        Start  11: sanity-ssg-tables-table-ol7-cuirefs.html
        Start  12: sanity-ssg-tables-table-ol7-pcidssrefs.html
        Start  13: sanity-ssg-tables-table-ol7-anssirefs.html
        Start  14: sanity-ssg-tables-table-ol7-ospprefs.html
        Start  15: sanity-ssg-tables-table-ol7-nistrefs-standard.html
        Start  16: sanity-ssg-tables-table-ol7-nistrefs-stig.html
[...]

PR #6906, opened 2021-04-28
Automatically prodtype rules in a profile

cipherboy:
#### Description

The new `autoprodtyper.py` utility automatically prodtypes all rules in
a given profile, adding the specified product. This leverages the work
previously done in `utils/mod_protype.py` to seamlessly edit prodtypes
with minimal diff (modifying only the prodtype line). If we assume a
profile for a product was created with intent (and that the rules in it
were intended to work for the product the profile is in), then
automatically adding this prodtype to the rules saves a bit of effort.

#### Rationale:

Per discussion with Jan and Matej on `#openscap`, I'd like to share what I've been working on, and eventually work together with upstream to improve the controls format.

This is one step in our work to begin officially maintaining Ubuntu content in CaC. Nobody likes adding prodtypes and this utility helps with that.

### DIFF
--- a/utils/autoprodtyper.py
+++ b/utils/autoprodtyper.py
@@ -1,0 +1,84 @@
+#!/usr/bin/env python
+
+import sys
+import os
+import argparse
+import json
+
+import ssg.build_yaml
+import ssg.products
+import ssg.rules
+import ssg.yaml
+import ssg.utils
+import ssg.rule_yaml
+
+from mod_prodtype import add_products
+
+SSG_ROOT = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description="Automatically add a missing product "
+                                     "reference to the prodtype key of a rule.yml based "
+                                     "on inclusion in a specified profile")
+    parser.add_argument("-j", "--json", type=str, action="store",
+                        default="build/rule_dirs.json", help="File to read "
+                        "json output of rule_dir_json from (defaults to "
+                        "build/rule_dirs.json")
+    parser.add_argument("-c", "--build-config-yaml", default="build/build_config.yml",
+                        help="YAML file with information about the build configuration. "
+                        "Defaults to build/build_config.yml")
+    parser.add_argument("-p", "--profiles-root",
+                        help="Override where to look for profile files.")
+    parser.add_argument("product", type=str, help="Product to add prod types for")
+    parser.add_argument("profile", type=str, help="Profile to iterate over")
+
+    return parser.parse_args()
+
+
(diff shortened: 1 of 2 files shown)


utils/autoprodtyper.py
```diff
@@ -0,0 +1,80 @@
...
+    parser.add_argument("product", type=str, help="Product to add prod types for")
+    parser.add_argument("profile", type=str, help="Profile to iterate over")
+
+    return parser.parse_args()
+
+
+def autoprodtype(env_yaml, rule_dirs, profile_path, product):
+    profile = ssg.build_yaml.ProfileWithInlinePolicies.from_yaml(profile_path, env_yaml)
+
+    for rule_id in profile.selected + profile.unselected:
```
jan-cerny, line 39:
```suggestion
    for rule_id in profile.selected + profile.unselected:
        print(rule_id)
```


It would be nice to print the rule ID out here. Or maybe in add_products.

 Because currently the output is somehow weird:

```
Current prodtype is empty, not adding the new prodtype.
Current prodtype is empty, not adding the new prodtype.
Current prodtype is empty, not adding the new prodtype.
Current prodtype is empty, not adding the new prodtype.
/home/jcerny/work/git/scap-security-guide/ssg/rule_yaml.py:132: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  return yaml.load(new_file)
Current prodtype: fedora,ol7,ol8,rhel7,rhel8,rhv4,sle12,sle15,wrlinux1019,rhcos4
New prodtype: fedora,ol7,ol8,rhcos4,rhel7,rhel8,rhel9,rhv4,sle12,sle15,wrlinux1019
```
It isn't clear to which rules the messages are related.

jan-cerny (commented):
Nice utility! It works great for me.

utils/autoprodtyper.py
cipherboy: For the following line:

> /home/jcerny/work/git/scap-security-guide/ssg/rule_yaml.py:132: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
>   return yaml.load(new_file)

Note that this PR actually depends on a change made in #6882. 

And yes, I should add output based on which rule was modified.

cipherboy: I've updated the rule output to silence the empty prodtypes and make which rule was modified clearer.

cipherboy:
Thanks @jan-cerny! I've updated with your comments. I changed the output a little actually.

jan-cerny:
Thanks, LGTM :D

PR #6907, opened 2021-04-28
Check all rules in profile for required reference

cipherboy:
#### Description

The new `refchecker.py` utility automatically checks all rules in a given
profile for a required reference field. This is most commonly useful
when creating benchmarks for a specified system, such as CIS or STIG,
where it is expected that all rules in a given benchmark have references
related to them.

Unlike `tests/missing_refs.sh` (and its helper tool,
`build-scripts/profile_tool.py`), this operates on the YAML level and
should have fewer false-positives in the content for non-RHEL
identifiers. Additionally, this parses the present values of the rules,
meaning that it can be used between builds.

#### Rationale

I found `build-scripts/profile_tool.py` not to work for Ubuntu CIS references, what this was originally meant to help with. I also don't really like waiting for even the Ubuntu product to build, so it nice to have a much quicker utility for checking some things (like missing references). 

I also didn't want to write the next utility when I first wrote this one. But I eventually ended up doing it anyways... :-)

See also: #6906

### DIFF
--- a/utils/refchecker.py
+++ b/utils/refchecker.py
@@ -5,12 +5,15 @@ import os
 import argparse
 import json
 
+import ssg.build_profile
 import ssg.build_yaml
+import ssg.controls
+import ssg.environment
 import ssg.products
 import ssg.rules
+import ssg.rule_yaml
 import ssg.yaml
 import ssg.utils
-import ssg.rule_yaml
 
 SSG_ROOT = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
 
@@ -25,6 +28,8 @@ def parse_args():
     parser.add_argument("-c", "--build-config-yaml", default="build/build_config.yml",
                         help="YAML file with information about the build configuration. "
                         "Defaults to build/build_config.yml")
+    parser.add_argument("--controls", default="controls",
+                        help="Directory that contains control files with policy controls.")
     parser.add_argument("-p", "--profiles-root",
                         help="Override where to look for profile files.")
     parser.add_argument("product", type=str, help="Product to check has required references")
@@ -55,8 +60,14 @@ def load_for_product(rule_obj, product, env_yaml=None):
     return rule
 
 
-def reference_check(env_yaml, rule_dirs, profile_path, product, reference):
+def reference_check(env_yaml, rule_dirs, profile_path, product, product_yaml, reference,
+                    profiles_root, controls_manager=None):
     profile = ssg.build_yaml.ProfileWithInlinePolicies.from_yaml(profile_path, env_yaml)
+
+    if controls_manager:
+        profile_files = ssg.products.get_profile_files_from_root(env_yaml, product_yaml)
+        all_profiles = ssg.build_profile.make_name_to_profile_mapping(profile_files, env_yaml)
+        profile.resolve(all_profiles, controls_manager)
 
     ok = True
     for rule_id in profile.selected + profile.unselected:
@@ -91,7 +102,12 @@ def main():
 
     product_base = os.path.join(SSG_ROOT, args.product)
     product_yaml = os.path.join(product_base, "product.yml")
-    env_yaml = ssg.yaml.open_environment(args.build_config_yaml, product_yaml)
+    env_yaml = ssg.environment.open_environment(args.build_config_yaml, product_yaml)
+
+    controls_manager = None
+    if os.path.exists(args.controls):
+        controls_manager = ssg.controls.ControlsManager(args.controls, env_yaml)
+        controls_manager.load()
 
     profiles_root = os.path.join(product_base, "profiles")
     if args.profiles_root:
@@ -105,7 +121,8 @@ def main():
         msg = msg.format(args.profile)
         raise ValueError(msg)
 
-    ok = reference_check(env_yaml, all_rules, profile_path, args.product, args.reference)
+    ok = reference_check(env_yaml, all_rules, profile_path, args.product, product_yaml,
+                         args.reference, profiles_root, controls_manager)
     if not ok:
         sys.exit(1)
 
--- a/utils/refchecker.py
+++ b/utils/refchecker.py
@@ -1,0 +1,114 @@
+#!/usr/bin/env python
+
+import sys
+import os
+import argparse
+import json
+
+import ssg.build_yaml
+import ssg.products
+import ssg.rules
+import ssg.yaml
+import ssg.utils
+import ssg.rule_yaml
+
+SSG_ROOT = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description="Check all rule.yml referenced in a given"
+                                     "profile for a required reference identifier")
+    parser.add_argument("-j", "--json", type=str, action="store",
+                        default="build/rule_dirs.json", help="File to read "
+                        "json output of rule_dir_json from (defaults to "
+                        "build/rule_dirs.json")
+    parser.add_argument("-c", "--build-config-yaml", default="build/build_config.yml",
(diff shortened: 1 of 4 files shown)


utils/refchecker.py
```diff
@@ -0,0 +1,114 @@
...
+
+def load(rule_obj, env_yaml=None, sce_metadata=None):
+    """
+    From the given rule_dir object, load the build_yaml.Rule associated with
+    it.
+    """
+
+    yaml_file = ssg.rules.get_rule_dir_yaml(rule_obj['dir'])
+    return ssg.build_yaml.Rule.from_yaml(yaml_file, env_yaml=env_yaml,
+                                         sce_metadata=sce_metadata)
```
jan-cerny, line 44: The sce_metadata doesn't exist

utils/refchecker.py
```diff
@@ -0,0 +1,114 @@
...
+import ssg.rules
+import ssg.yaml
+import ssg.utils
+import ssg.rule_yaml
+
+SSG_ROOT = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
+
+
+def parse_args():
+    parser = argparse.ArgumentParser()
```
jan-cerny, line 19: Please add a help text

utils/refchecker.py
```diff
@@ -0,0 +1,114 @@
...
+    profile_filename = args.profile + ".profile"
+    profile_path = os.path.join(profiles_root, profile_filename)
+    if not os.path.exists(profile_path):
+        msg = "Unknown profile {0}: check profile, --profiles-root, and try again"
+        msg = msg.format(args.profile)
+        raise ValueError(msg)
+
+    ok = reference_check(env_yaml, all_rules, profile_path, args.product, args.reference)
+    if not ok:
+        print("One or more rules failed required reference check.", file=sys.stderr)
```
jan-cerny, line 109: We can see that from the previous output.

jan-cerny:
It's amazing.

>I found build-scripts/profile_tool.py not to work for Ubuntu CIS references, what this was originally meant to help with.

Have you found why?

> so it nice to have a much quicker utility for checking some things 

Do you think it can replace the profile_tool.py in the ctest suite?

utils/refchecker.py
cipherboy: Sorry, that's an unrelated change that crept in to this utility. :)

cipherboy:
@jan-cerny said:
> It's amazing.

Thanks! I've updated with your feedback, thanks! :) 

@jan-cerny said:
> @cipherboy said:
> > I found build-scripts/profile_tool.py not to work for Ubuntu CIS references, what this was originally meant to help with.
> 
> Have you found why?

I figured it'd be a hard problem to solve in general, since each product has its own CIS & STIG ids in the XCCDF layer. e.g,. `https://www.cisecurity.org/benchmark/ubuntu_linux/` vs `https://www.cisecurity.org/benchmark/red_hat_linux/`. This is all encoded at the XSLT layer `<product>/transforms/constants.xslt`  -- but the Python layer (`ssg/constants.py` -- `cis_ns`) doesn't know about each product's unique constants. We could pull this out at runtime (and require `--product` be specified on the CLI if it can't be inferred from the filename) from the XSLT... But I don't know. 

Thoughts?

> 
> > so it nice to have a much quicker utility for checking some things
> 
> Do you think it can replace the profile_tool.py in the ctest suite?

Sure, doable. I'd maybe run both since this is so quick. They do check different things and at the end of the build step (when you're ready to run the test suite), you _do_ have XCCDF available, so both are, IMO, useful. 

I'd let you or someone else propose test suite addition in a future PR. Not sure which profiles &c y'all want to test and Ubuntu doesn't yet have any upstream completed profiles ;-)

utils/refchecker.py
jan-cerny: Yes, but because of this variable it doesn't work.

```
  File "/home/jcerny/work/git/scap-security-guide/utils/refchecker.py", line 114, in <module>
    main()
  File "/home/jcerny/work/git/scap-security-guide/utils/refchecker.py", line 107, in main
    ok = reference_check(env_yaml, all_rules, profile_path, args.product, args.reference)
  File "/home/jcerny/work/git/scap-security-guide/utils/refchecker.py", line 68, in reference_check
    rule = load_for_product(rule_dirs[rule_id], product, env_yaml=env_yaml)
  File "/home/jcerny/work/git/scap-security-guide/utils/refchecker.py", line 53, in load_for_product
    rule = load(rule_obj, env_yaml=env_yaml, sce_metadata=sce_metadata)
  File "/home/jcerny/work/git/scap-security-guide/utils/refchecker.py", line 43, in load
    return ssg.build_yaml.Rule.from_yaml(yaml_file, env_yaml=env_yaml,
TypeError: from_yaml() got an unexpected keyword argument 'sce_metadata'
```

jan-cerny:
> Thanks! I've updated with your feedback, thanks! :)

You forgot to push it :)

> I figured it'd be a hard problem to solve in general, since each product has its own CIS & STIG ids in the XCCDF layer. e.g,. https://www.cisecurity.org/benchmark/ubuntu_linux/ vs https://www.cisecurity.org/benchmark/red_hat_linux/. This is all encoded at the XSLT layer <product>/transforms/constants.xslt -- but the Python layer (ssg/constants.py -- cis_ns) doesn't know about each product's unique constants. We could pull this out at runtime (and require --product be specified on the CLI if it can't be inferred from the filename) from the XSLT... But I don't know.

> Thoughts?

Thanks for the investigation. Yes, the problem seems to be the different URL in `<reference href="...">` in different products.

Requiring the product in profile_tool.py seems as the best way to fix the profile_tool.py, although it might be difficult. Another option would be to have multiple different checks for CIS IDs (`--missing-ubuntu-cis-refs`, `--missing-rhel-cis-refs`) which would be easy to implement but it doesn't scale.

> Sure, doable. I'd maybe run both since this is so quick. They do check different things and at the end of the build step (when you're ready to run the test suite), you do have XCCDF available, so both are, IMO, useful.

Thanks for the explanation.

> I'd let you or someone else propose test suite addition in a future PR. Not sure which profiles &c y'all want to test and Ubuntu doesn't yet have any upstream completed profiles ;-)

OK.

jan-cerny:
Also it would be great to document this in docs/manual/developer/05_tools_and_utilities.md

utils/refchecker.py
```diff
@@ -0,0 +1,114 @@
...
+    it, normalizing for the given product.
+    """
+
+    rule = load(rule_obj, env_yaml=env_yaml)
+    rule.normalize(product)
+    return rule
+
+
+def reference_check(env_yaml, rule_dirs, profile_path, product, reference):
+    profile = ssg.build_yaml.ProfileWithInlinePolicies.from_yaml(profile_path, env_yaml)
```
jan-cerny, line 59: I have reviewed it again and I have found one more thing. There's a situation in which it doesn't work form me. I have removed the anssi reference from rule selinux_state and then I tried if the tool reports it, but it doesn't. ` utils/refchecker.py  rhel8 anssi_bp28_high anssi` gives no output. I think it's because of the controls feature used in anssi profiles. I think you need to run `resolve` method on the `profile`.

cipherboy: Well that was a bit of work ;-)

I ended up refactoring `compile_profiles.py` to make this possible. Let me know what you think.

cipherboy:
> Requiring the product in profile_tool.py seems as the best way to fix the profile_tool.py, although it might be difficult. Another option would be to have multiple different checks for CIS IDs (--missing-ubuntu-cis-refs, --missing-rhel-cis-refs) which would be easy to implement but it doesn't scale.

Just figured I'd mention it here as well: I've opened #6941 fixing `profile_tool.py` use case for other CIS products.

jan-cerny:
amazing!

Issue #6889, opened 2021-04-26
Allow Remarks alongside profile entries and other Profile-related changes


cipherboy:
#### Description of problem:

Thanks to @mab879's work on https://github.com/OpenSCAP/openscap/pull/1744, I was thinking it might be nice to expose this functionality in CaC. As a parallel to https://github.com/OpenSCAP/openscap/issues/1743 consider this:

When building profiles, we might want to give a reason (to the end-consumer of the CaC/content artifacts) why certain rules are either selected or not selected by default in this profile and how this selection impacts them. `<xccdf:remark />` provides this capability. 

I could see one of two design paths for this:

 1. Modify the format of the profile YAML file. Either turn `selection` into an array of (optionally) a structure like:
     ```yaml
     selection:
      - my_rule
      - rule: my_other_rule
        remark: selected because we need it 
     ```
     The downside is this approach makes reading and parsing the profile harder. I have some tooling which assumes a certain structure of the YAML (grep/sed/awk) that would break. Small matter I guess :-) Maybe the SSG test suite would also break? I do not know!
 2.  Alternatively, stick this data somewhere else but close by. Either within the same `.profile` in a different section (`remarks` maybe, mapping `rule_id` into remark text? Or in a adjacent file (`cis_level1_server.remarks`). Either way could probably work.

I think the former would be the cleanest design, but maybe this breaks other assumptions in other places? 

So I'd personally lean towards either of the designs in path 2 above.

Thoughts?

JAORMX:
uh! this sounds awesome! We've been missing this from profiles and tailored profiles too.

yuumasato:
Either approach looks fine to me, but slightly leaning towards approach 1.

The first approach keeps the remarks close to the rule selection, keeping spacial locality of data tighter. I think this helps maintenance of consistency. For example, when removing a rule it is easier to note the remark and remove it too.

> I think the former would be the cleanest design, but maybe this breaks other assumptions in other places?

I don't think the change in the selection would affect the Test Suite.
It will impact the compiled/resolved profiles, which are used by the profile stability tests. But it is probably just a matter of ignoring the remarks during the test.

The second approach makes it easy to share and add remarks to already existing profiles, but keeps it a separate "thing".
In my opinion, this makes the remarks easier to forget and prone to being outdated.

cipherboy:
> Either approach looks fine to me, but slightly leaning towards approach 1.

Would you prefer the "new" format strictly enforced? Something like:

```yaml
[...]

PR #6908, opened 2021-04-28
Add utility to automatically reference rules

cipherboy:
#### Description

Nicely laid out profiles have a structure matching their corresponding
benchmarks. I'm thinking the CIS and STIG profiles here.

Let's formalize that structure a bit.

Given a profile with format:

```yaml
    selection:
        # <identifier> description
        - rule_choice
```

Where one or more comments precede one or more rules, and the closest
rule that matches the given format for a reference identifier wins.

Let's take this as an example:

```yaml
    ## 5.3 Configure PAM ##
    ### 5.3.1 Ensure password creation requirements are configured (Automated)
    - var_password_pam_minlen=14
    - accounts_password_pam_minlen
    - var_password_pam_minclass=4
    - accounts_password_pam_minclass
```

Here both `accounts_password_pam_minlen` and
`accounts_password_pam_minclass` should get CIS reference value 5.3.1. The
other two entries are vars, and since 5.3 is further away than 5.3.1,
5.3.1 should win.

Some ground rules:

 - We should avoid guessing when possible
 - We should create minimal diffs
 - Some rules lack a references section; we should add them in that
   case.
 - If we're not sure, ignore the rule and print info telling the caller
   about it.
 - Rules that don't belong to a section shouldn't be in the profile!
 - If we're adding a reference, don't clutter other products! Only do
   our current product.

Mostly this is built on top of the `ssg.rule_yaml` helpers I've added
previously during my internship.

The structure roughly follows the previous `utils/refchecker.py`. Walk all
rules in a profile, load their YAML, load their actual lines, use a bit
of magic to auto-detect the rule identifier, and write the results out.

Most of the heavy lifting was done in `rule_yaml.py` to add a new helper
method, add_or_modify_nested_section_key, that adds a new key in a
nested section (and optionally, the section itself) or -- if said key is
already there -- modifies the value of that key.

#### Rationale

I was talking with Jan and Matej on `#openscap` about the new controls file. We're working on porting our current Ubuntu CIS benchmark into CaC; this is just one step that will help with it.

I'm opening this in the interest of working together on a common solution to product-specific controls and benchmark-based profiles. Knowing what we're currently using should help show where we're coming from and make a common design possible.  

See also: #6906 and #6907

### DIFF
--- a/ssg/rule_yaml.py
+++ b/ssg/rule_yaml.py
@@ -175,3 +175,112 @@ def get_section_lines(file_path, file_contents, key_name):
         return section[0]
 
     return None
+
+
+def get_line_whitespace(line):
+    """
+    Get the exact whitespace used at the start of this line.
+    """
+    stripped_line = line.lstrip()
+    delta = len(line) - len(stripped_line)
+    return line[:delta]
+
+
+def guess_section_whitespace(file_contents, section_range, default='    '):
+    """
+    Hack: we need to figure out how much whitespace to add when adding a new key to
+    an existing section. Since different files might be parsed differently, take the
+    minimum key's whitespace length in this section.
+    """
+    whitespace = None
+    for line_num in range(section_range.start+1, section_range.end):
+        line = file_contents[line_num]
+        if line and ':' in line:
+            # Assume this is a key, so update our assumptions of whitespace. We ignore
+            # non-key lines.
+            this_whitespace = get_line_whitespace(line)
+
+            # Only take it if we have _less_ whitespace (to avoid dealing with nested
+            # sections) or if we have no whitespace yet.
+            if whitespace is None or len(this_whitespace) < len(whitespace):
+                whitespace = this_whitespace
+
+    # If we don't have any whitespace, use the default to show the YAML parser it
+    # is a nested section.
+    if whitespace is None:
+        whitespace = default
+
+    return whitespace
+
+
+def add_or_modify_nested_section_key(file_path, file_contents, section_title,
+                                     key, value, new_section_after_if_missing=None):
+    """
+    Either modify an existing nested section key (in key: value) form or
+    add it if missing. Optionally, take a section and add our new section
+    after the existing section.
+    """
+    new_contents = file_contents[:]
+    section = get_section_lines(file_path, file_contents, section_title)
+
+    if not section:
+        if not new_section_after_if_missing:
+            msg = "File %s lacks all instances of section %s; refusing to modify file."
+            msg = msg.format(file_path, section)
+            raise ValueError(msg)
+
+        previous_section = get_section_lines(file_path, file_contents,
+                                             new_section_after_if_missing)
+        if not previous_section:
+            msg = "File %s lacks all instances of sections %s and %s; refusing to modify file."
+            msg = msg.format(file_path, section, new_section_after_if_missing)
+            raise ValueError(msg)
+
+        new_section_header = get_line_whitespace(file_contents[previous_section.start])
+        new_section_header += section_title + ':'
+        new_section_kv = guess_section_whitespace(file_contents, previous_section)
+        new_section_kv += key + ': ' + value
+
+        new_section = [new_section_header, new_section_kv, '']
+
+        tmp_contents = new_contents[:previous_section.end+1]
+        tmp_contents += new_section
+        tmp_contents += new_contents[previous_section.end+1:]
+        new_contents = tmp_contents
+
+        return new_contents
+
+    # Nasty hacky assumption: assume key is 'unique' within the section and we can
+    # ignore whitespaces issues with this approach. Also assume (and validate!) that
+    # : does not appear in the key. This allows us to split the line by ':' and take
+    # the first as the actual key in the file.
+    assert ':' not in key
+    key_match = ' ' + key + ':'
+
+    found = None
+    for line_num in range(section.start, section.end+1):
+        line = file_contents[line_num]
+        if key_match in line:
+            if found:
+                msg = "Expected to only have key {0} appear once in file, but appeared "
+                msg += "twice: once on line {1} and once on line {2}."
+                msg = msg.format(key, found, line_num)
+                raise ValueError(msg)
+
+            # Preserve leading whitespace. :-)
+            key_prefix = line.split(':', maxsplit=1)[0]
+            new_line = key_prefix + ': ' + value
+            new_contents[line_num] = new_line
+            found = line
+
+    if not found:
+        # Be lazy and add it right after the section heading. Worst case we'll just
+        # come back and sort the section at a later time.
+        whitespace = guess_section_whitespace(file_contents, section)
+        new_line = whitespace + key + ': ' + value
+        tmp_contents = new_contents[:section.start+1]
+        tmp_contents += [new_line]
+        tmp_contents += new_contents[section.start+1:]
+        new_contents = tmp_contents
+
+    return new_contents
--- a/utils/autorefer.py
+++ b/utils/autorefer.py
@@ -1,0 +1,258 @@
+#!/usr/bin/env python
+
+import sys
+import os
+import argparse
+import json
+
+import ssg.build_yaml
+import ssg.products
+import ssg.rules
+import ssg.yaml
+import ssg.utils
+import ssg.rule_yaml
+
+from refchecker import load_for_product
+
+SSG_ROOT = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
+SECTION_KEY_FUNC = ssg.rule_yaml.add_or_modify_nested_section_key
+
+
+"""
+Nicely laid out profiles have a structure matching their corresponding
+benchmarks. I'm thinking the CIS and STIG profiles here.
+
+Let's formalize that structure a bit.
+
+Given a profile with format:
+
+```yaml
+    selection:
+        # <identifier> description
+        - rule_choice
+```
+
+Where one or more comments precede one or more rules, and the closest
+comment that matches the given format for a reference identifier wins.
+
+Let's take this as an example:
+
+```yaml
+    ## 5.3 Configure PAM ##
+    ### 5.3.1 Ensure password creation requirements are configured (Automated)
+    - var_password_pam_minlen=14
+    - accounts_password_pam_minlen
+    - var_password_pam_minclass=4
+    - accounts_password_pam_minclass
+```
+
+Here both accounts_password_pam_minlen and
+accounts_password_pam_minclass should get CIS reference value 5.3.1. The
+other two entries are vars, and since 5.3 is further away than 5.3.1,
+5.3.1 should win.
+
+Some ground rules:
+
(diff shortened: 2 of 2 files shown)


cipherboy:
Note that, if at the end you want a section of rules with multiple identifiers, you can do it like so:

```yaml
        ### 5.3.1,5.3.2 Multi-use rules
        - some_other_rule
```

This wouldn't quite fit in the existing benchmark structure I suppose, but could be inserted in a useful place. I didn't experiment with what happened with one rule that is placed multiple times in a profile; it would probably fail the build too.

matejak:
I don't like the approach for following reasons:

- In the real world, reference is not a rule’s property, but it depends on a relationship of a rule and a regulation that exists independently from the rule. Our project started with references as rule properties, but it is clearly visible that if the situation continues to develop at the this pace, we will end up with rule definition files that will contain a scary reference blob that won't be human-readable.
But the project has made a step in a direction away from this - we have control files that are supposed to be based on those regulations. Next, we have the concept of compiled rules as an intermediary, but still human-readable format. Therefore, I would much rather see the future in references being added by means of control files (or even profile files) rather than by means of rule keys.
- The proposal would introduce duplication of information - part of which would be present in comments.

I am sorry that you put quite some energy into this proposal, but I think that its elements could be leveraged when references are added to rule metadata as a step of the build process.

cipherboy:
Hi @matejak :-) Thanks for taking the time to look! I think this fits best in conjunction with my thoughts on #6889 and #6915 (the off-topic ones) if you haven't read them. 

> Our project started with references as rule properties, but it is clearly visible that if the situation continues to develop at the this pace, we will end up with rule definition files that will contain a scary reference blob that won't be human-readable.

I definitely hear and understand that. I'll start thinking about how to solve that. :-) 

> The proposal would introduce duplication of information - part of which would be present in comments.

I'd like to push back here. While the information is _duplicated_ in the strictest since (and in conjunction with #6889 and #6915) the intent is to have the source of truth ultimately be the profile. This utility is a way to make that profile _push_ references back to the rules themselves. The duplication is necessary for the build system but this tooling makes the profile the canonical reference location, allowing automatic update of the rules in question.

Where I'm coming from is that it is _really_ helpful to have content about a _rule_ centralized in a single place. That's why we created the rule directory (remember the old structure? I barely do!) in the first place: everything about the XCCDF rule entry (including references, ...) are in one place. I'd be fine splitting out a rule into a separate `references.yml` if you wanted to make auto-generated stuff exist in one place. But if you make _controls_ influence _profiles_ and _rules_'s references (which they don't do presently, correct?) --- now you need another build step and it gets really tough to see where everything (in the resulting XCCDF) is coming from.

The other intent with this is to make changes hmm _auditable_. You mentioned elsewhere about making sure changes don't unintentionally impact the generated content. I'm 100% on board with that. By checking in the resulting profiles and references, I feel it becomes easier to audit changes. You know any changes in content must reflect back into existing rules that are checked in, else they'll be ignored. For large changes (consider an upgrade from CIS benchmark at v1.0.0 to v3.0.x) --- being able to do so incrementally across multiple PRs (but having a large portion being automated such as references & prodtypes) would be a big help.

With things not happening until the actual build stage, you're stuck auditing _build artifacts_. The resulting XCCDF still isn't completely `diff`-able IIUC. We had that problem when I was an intern too. By taking some of the more risky moving parts (bulk changes of any sort -- profile generation, prodtypes, references, &c) and putting them back into the existing content system (rather than as another layer on top) --- you restore confidence in the content you're shipping. 

You could even go so far as to add some of these steps to the test system if you prefer. The goal of my tooling is to be _minimal_ and _reproducible_ -- this makes review simple: run the same tools locally and compare the generated output or see if they make any new changes. They shouldn't if the PR is valid (is the assertion I'm trying to bake in with these tool changes).

We're just beginning our CaC journey at Canonical, but already this tool has been _very_ helpful for importing our existing CIS tooling into the CaC build system.

---

The goal with this PR is to provide something that _works_ for us, now. IIUC this is a capability that the controls file currently lacks.

Over time, as we discuss and reach consensus over what changes we all agree to make, we can update our tooling and work towards that goal. This is, IMO, an important step towards making controls file (whether directly or via an intermediate, checked-in profile representation) _automatically_ generate references. I'm in no way saying that this tool is _perfect_ or _fully complete_. 

Yeah, there's still work to be done in the ultimate goal, but having _some_ starting code that we can then change, reuse, extend, whatever to fit the eventual end goal is a lot better than having none. :-)

(Or who knows, maybe it is better to throw it all away, and start over, but at least there's some existing code that solved a very similar problem already). 

----

Perhaps I should combine all my various discussions on what I'm envisioning into a single document for our call later, would that help? 


Hope this helps :-)

matejak:
> Hi @matejak :-) Thanks for taking the time to look! I think this fits best in conjunction with my thoughts on #6889 and #6915 (the off-topic ones) if you haven't read them.
> 
> > Our project started with references as rule properties, but it is clearly visible that if the situation continues to develop at the this pace, we will end up with rule definition files that will contain a scary reference blob that won't be human-readable.
> 
> I definitely hear and understand that. I'll start thinking about how to solve that. :-) Do you have ideas yet?
> 
> > The proposal would introduce duplication of information - part of which would be present in comments.
> 
> I'd like to push back here. While the information is _duplicated_ in the strictest since (and in conjunction with #6889 and #6915) the intent is to have the source of truth ultimately be the profile. This utility is a way to make that profile _push_ references back to the rules themselves. The duplication is necessary for the build system but this tooling makes the profile the canonical reference location, allowing automatic update of the rules in question.

I understand the thought that there would the primary source in the profile file, but as long as the build system looks into rules, then there is duplication. Moreover, some references are currently present only in rules, so a human being can't tell just by looking whether a reference in a rule YAML is derived from a profile file comment, or it exists just in the rule yaml. Obviously, one could enforce that every reference in rules has to be derived from a reference in a profile file, but we would be then talking about a backward-incompatible change of input formats in the project, which can't be introduced as a PR, but it needs public discussion in the ML at least.
> 
> Where I'm coming from is that it is _really_ helpful to have content about a _rule_ centralized in a single place. That's why we created the rule directory (remember the old structure? I barely do!) in the first place: everything about the XCCDF rule entry (including references, ...) are in one place. I'd be fine splitting out a rule into a separate `references.yml` if you wanted to make auto-generated stuff exist in one place. But if you make _controls_ influence _profiles_ and _rules_'s references (which they don't do presently, correct?) --- now you need another build step and it gets really tough to see where everything (in the resulting XCCDF) is coming from.

I agree that there would have to be another build step. But I don't see that as something bad and scary - just take a look how a compiled rule differs from a rule in the source tree - there would be one more key s.a. control-based references, and that would be it.
Again, reference is not a rule property. A rule is e.g. sshd can't allow root to log in. It has description, rationale, remediations, and checks. Then there are identifiers, which are supposed to uniquely identify the rule, so although they are derived, they are permanent and unambiguous, so it makes sense to store them close to the rule instead in some kind of rule ID -> other IDs mapping data structure.
References to policies are something entirely different - they are subjective, there is usually some sort of reasoning related to the relationship, and they can also change. I don't like edits to a blob in the rule yaml file, although the rule itself hasn't changed a bit.

> The other intent with this is to make changes hmm _auditable_. You mentioned elsewhere about making sure changes don't unintentionally impact the generated content. I'm 100% on board with that. By checking in the resulting profiles and references, I feel it becomes easier to audit changes. You know any changes in content must reflect back into existing rules that are checked in, else they'll be ignored. For large changes (consider an upgrade from CIS benchmark at v1.0.0 to v3.0.x) --- being able to do so incrementally across multiple PRs (but having a large portion being automated such as references & prodtypes) would be a big help.
>
> With things not happening until the actual build stage, you're stuck auditing _build artifacts_. The resulting XCCDF still isn't completely `diff`-able IIUC. We had that problem when I was an intern too. By taking some of the more risky moving parts (bulk changes of any sort -- profile generation, prodtypes, references, &c) and putting them back into the existing content system (rather than as another layer on top) --- you restore confidence in the content you're shipping.

prodtypes are another pain, yes. And I agree with the emphasis on the auditable behavior. Maybe I get it wrong, but I don't see the difference between auditing source rule files that have partial machine-generated content, and auditing compiled rule files that are a build artifact, but they are otherwise the same. We already have some kind of auditing of profile changes: https://complianceascode.github.io/template/2020/03/06/profile_guard.html And we already have a compiler of rule references to an ordered json: https://github.com/ComplianceAsCode/content/blob/master/utils/gen_tables.py that could be used in a similar way.

> You could even go so far as to add some of these steps to the test system if you prefer. The goal of my tooling is to be _minimal_ and _reproducible_ -- this makes review simple: run the same tools locally and compare the generated output or see if they make any new changes. They shouldn't if the PR is valid (is the assertion I'm trying to bake in with these tool changes).
> 
> We're just beginning our CaC journey at Canonical, but already this tool has been _very_ helpful for importing our existing CIS tooling into the CaC build system.
> 
> The goal with this PR is to provide something that _works_ for us, now. IIUC this is a capability that the controls file currently lacks.
> 
> Over time, as we discuss and reach consensus over what changes we all agree to make, we can update our tooling and work towards that goal. This is, IMO, an important step towards making controls file (whether directly or via an intermediate, checked-in profile representation) _automatically_ generate references. I'm in no way saying that this tool is _perfect_ or _fully complete_.

I think that you can benefit from the tooling without having it merged upstream. As I have outlined above, your proposal has big consequences, as basically changes the content input workflow and interpretation of the source yamls, and such proposal is much bigger than a PR. While it may improve workflows, I believe that the steps that the whole project makes should, among other things, play nice with backward compatibility, and they should introduce workflow changes that are likely not to change soon again, which I see as weaknesses of your proposal.

> Yeah, there's still work to be done in the ultimate goal, but having _some_ starting code that we can then change, reuse, extend, whatever to fit the eventual end goal is a lot better than having none. :-)
> 
> (Or who knows, maybe it is better to throw it all away, and start over, but at least there's some existing code that solved a very similar problem already).
> 
> Perhaps I should combine all my various discussions on what I'm envisioning into a single document for our call later, would that help?

As I have already said, I believe that some of the code should be reused in the bigger-picture task to enhance the build system in a way that is not clear yet. We think that something needs to be done wrt reference handling, but parsing comments in profile files is not something that we would like. Regarding your proposal to accumulate ideas into a document - that's a great idea, go for it.

cipherboy:
Just two quick follow up points:

 1. I think part of the question is around what _is_ a rule. In your mind, it sounds like it is purely a state+action that needs to be taken to a system. From my perspective, it is still the full `<XCCDF:Rule />` element, in a nicer form. 
 2. Which ML? I joined the public `open-scap-list <open-scap-list@redhat.com>` but its never had much traffic, especially on these type of discussions (at least, as far as I recall in the past) --- is there a separate public developer mailing list I've never found? :-) 

Happy to organize my thoughts into a document form if that benefits the community. I was hoping that GH Issue/PR with something concrete would be easy enough to see and respond to, but I guess not. :-)

cipherboy:
Closing this and carrying this downstream.

PR #6909, opened 2021-04-28
Fix usbguard match-all syntax for HID rule

nmaludy:
#### Description:

The description for the usbguard HID rule had a suggested fix with incorrect configuration syntax. 
Specifically it suggested a fix of: 
```
allow with-interface match_all { 03:*:* 09:00:* }`
```

Unfortunately `match_all` is incorrect syntax, instead it should be `match-all`, full example:
```
allow with-interface match-all { 03:*:* 09:00:* }
```

The `shared.sh` fix script has the correct syntax and matches the syntax adjusted in this PR for the descriptions.

#### Rationale:

Ensures that description has proper syntax and matches the `shared.sh` fix script.

### DIFF
--- a/linux_os/guide/services/usbguard/usbguard_allow_hid/rule.yml
+++ b/linux_os/guide/services/usbguard/usbguard_allow_hid/rule.yml
@@ -8,7 +8,7 @@ description: |-
     To allow authorization of Human Interface Devices (keyboard, mouse)
     by USBGuard daemon,
     add the line
-    <tt>allow with-interface match_all { 03:*:* }</tt>
+    <tt>allow with-interface match-all { 03:*:* }</tt>
     to <tt>/etc/usbguard/rules.conf</tt>.
 
 rationale: |-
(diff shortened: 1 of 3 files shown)


yuumasato:
@openscap-ci ok to test

yuumasato:
@openscap-ci test this please

nmaludy:
@yuumasato looks like this is failing on a bunch of things i didn't change, any ideas?

JAORMX:
@nmaludy the cos-node test is broken due to a regression in openvswitch. I'm looking into it. The failure is unrelated to this PR

JAORMX (approved):

yuumasato:
> @yuumasato looks like this is failing on a bunch of things i didn't change, any ideas?

One of the mandatory jobs had failed on rhel7, so I restarted.
For some it had failed to fetch the tree...Not sure why that happens sometimes, re-running the job does the trick.

PR #6911, opened 2021-04-28
Fix bugs in rule_dir_json.py

cipherboy:
**Note**: This fixes a correctness bug as well. Previously we'd generate an incorrect products applicability when parsing rules from products with `applications/` as the guide dir. This is because I had special-cased `linux_os`'s many-to-one product applicability. This should be fixed now for all products.

### DIFF
--- a/utils/rule_dir_json.py
+++ b/utils/rule_dir_json.py
@@ -37,7 +37,14 @@ def walk_products(root, all_products):
     all_rule_dirs = []
     product_yamls = {}
 
-    for product in all_products:
+    # There is a one-to-many mapping of guide_dir->product. This means that
+    # we first need to walk all products, gather their guide_dirs, and then
+    # come back and walk the guide_dirs so we have a complete list of products
+    # to evaluate each under.
+    product_to_guide = dict()
+    guide_to_products = defaultdict(set)
+
+    for product in sorted(all_products):
         product_dir = os.path.join(root, product)
         product_yaml_path = os.path.join(product_dir, "product.yml")
         product_yaml = ssg.products.load_product_yaml(product_yaml_path)
@@ -46,13 +53,29 @@ def walk_products(root, all_products):
         guide_dir = os.path.join(product_dir, product_yaml['benchmark_root'])
         guide_dir = os.path.abspath(guide_dir)
 
-        additional_content_directories = product_yaml.get("additional_content_directories", [])
-        add_content_dirs = [os.path.abspath(os.path.join(product_dir, rd)) for rd in additional_content_directories]
+        product_to_guide[product] = guide_dir
+        guide_to_products[guide_dir].add(product)
 
-        for cur_dir in [guide_dir] + add_content_dirs:
+    for product in sorted(all_products):
+        product_dir = os.path.join(root, product)
+        product_yaml = product_yamls[product]
+        guide_dir = product_to_guide[product]
+
+        potential_products = list(guide_to_products[guide_dir])
+        if guide_dir not in visited_dirs:
+            for rule_id, rule_dir in collect_rule_ids_and_dirs(guide_dir):
+                all_rule_dirs.append((rule_id, rule_dir, guide_dir, potential_products))
+            visited_dirs.add(guide_dir)
+
+        additional_content_directories = product_yaml.get("additional_content_directories", [])
+        if not additional_content_directories:
+            continue
+
+        for add_content_dirs in additional_content_directories:
+            cur_dir = os.path.abspath(os.path.join(product_dir, add_content_dirs))
             if cur_dir not in visited_dirs:
                 for rule_id, rule_dir in collect_rule_ids_and_dirs(cur_dir):
-                    all_rule_dirs.append((rule_id, rule_dir, cur_dir, product))
+                    all_rule_dirs.append((rule_id, rule_dir, cur_dir, [product]))
                 visited_dirs.add(cur_dir)
 
     return all_rule_dirs, product_yamls
@@ -158,19 +181,17 @@ def main():
     all_rule_dirs, product_yamls = walk_products(args.root, all_products)
 
     known_rules = {}
-    for rule_id, rule_dir, guide_dir, given_product in all_rule_dirs:
-        product_list = sorted(linux_products)
-        if 'linux_os' not in guide_dir:
-            product_list = [given_product]
-
+    for rule_id, rule_dir, guide_dir, given_products in all_rule_dirs:
         try:
-            rule_obj = handle_rule_yaml(product_list, product_yamls, rule_id, rule_dir, guide_dir)
+            rule_obj = handle_rule_yaml(given_products, product_yamls, rule_id,
+                                        rule_dir, guide_dir)
         except ssg.yaml.DocumentationNotComplete:
             # Happens on non-debug build when a rule is "documentation-incomplete"
             continue
 
-        rule_obj['ovals'], oval_products = handle_ovals(product_list, product_yamls, rule_obj)
-        rule_obj['remediations'], r_products = handle_remediations(product_list, product_yamls, rule_obj)
+        rule_obj['ovals'], oval_products = handle_ovals(given_products, product_yamls, rule_obj)
+        rule_obj['remediations'], r_products = handle_remediations(given_products, product_yamls,
+                                                                   rule_obj)
 
         # Validate oval products
         for key in oval_products:
--- a/utils/rule_dir_json.py
+++ b/utils/rule_dir_json.py
@@ -176,7 +176,10 @@ def main():
         for key in oval_products:
             oval_products[key] = sorted(oval_products[key])
             if len(oval_products[key]) > 1:
-                print("product has multiple ovals: %s - %s" % (key, ','.join(oval_products[key])), file=sys.stderr)
+                all_ovals = ','.join(oval_products[key])
+                msg = "Product {0} has multiple ovals in rule {1}: {2}"
+                msg = msg.format(key, rule_id, all_ovals)
+                print(msg, file=sys.stderr)
 
         rule_obj['oval_products'] = oval_products
 
@@ -186,7 +189,11 @@ def main():
             if len(r_products[key]) > 1:
                 exts = sorted(map(lambda x: os.path.splitext(x)[1], r_products[key]))
                 if len(exts) != len(set(exts)):
-                    print("product has multiple remediations of the same type: %s - %s" % (key, ','.join(r_products[key])), file=sys.stderr)
+                    all_fixes = ','.join(r_products[key])
+                    msg = "Product {0} has multiple remediations of the same type "
+                    msg += "in rule {1}: {2}"
+                    msg = msg.format(key, rule_id, all_fixes)
+                    print(msg, file=sys.stderr)
 
         rule_obj['remediation_products'] = r_products
 


cipherboy:
Just to note that I am getting some duplicate remediation output here form some SUSE artifacts:

```
Product sle12 has multiple remediations of the same type in rule no_empty_passwords: sle12.yml,sle15.yml
Product sle15 has multiple remediations of the same type in rule no_empty_passwords: sle12.yml,sle15.yml
Product sle12 has multiple remediations of the same type in rule accounts_tmout: sle12.sh,sle12.yml,sle15.yml
Product sle15 has multiple remediations of the same type in rule accounts_tmout: sle12.sh,sle12.yml,sle15.yml
Product sle12 has multiple remediations of the same type in rule clean_components_post_updating: sle12.sh,sle12.yml,sle15.yml
Product sle15 has multiple remediations of the same type in rule clean_components_post_updating: sle12.yml,sle15.yml
Product sle12 has multiple remediations of the same type in rule ensure_gpgcheck_globally_activated: sle12.yml,sle15.yml
Product sle15 has multiple remediations of the same type in rule ensure_gpgcheck_globally_activated: sle12.yml,sle15.yml
```

Suggested fix would be to use `SUSE Linux Enterprise 12` and `SUSE Linux Enterprise 15` instead of `multi_platform_sle` btw. 

/cc @guangyee @brett060102

utils/rule_dir_json.py
```diff
@@ -158,25 +183,26 @@ def main():
...
+        rule_obj['remediations'], r_products = handle_remediations(given_products, product_yamls,
+                                                                   rule_obj)
 
         # Validate oval products
         for key in oval_products:
             oval_products[key] = sorted(oval_products[key])
             if len(oval_products[key]) > 1:
-                print("product has multiple ovals: %s - %s" % (key, ','.join(oval_products[key])), file=sys.stderr)
+                all_ovals = ','.join(oval_products[key])
+                msg = "Product {0} has multiple ovals in rule %1}: {2}"
```
jan-cerny, line 203:
```suggestion
                msg = "Product {0} has multiple ovals in rule {1}: {2}"
```

utils/rule_dir_json.py
```diff
@@ -46,13 +53,31 @@ def walk_products(root, all_products):
         guide_dir = os.path.join(product_dir, product_yaml['benchmark_root'])
         guide_dir = os.path.abspath(guide_dir)
 
+        product_to_guide[product] = guide_dir
+        if guide_dir not in guide_to_products:
+            guide_to_products[guide_dir] = set()
+        guide_to_products[guide_dir].add(product)
```
jan-cerny, line 59: What about using collections.defaultdict for guide_to_products?

cipherboy: I keep forgetting it is allowed in Python 2... Even though it was already imported. :D

cipherboy:
Updated. Good catches, @jan-cerny!

matejak:
As the json file is not product-specific and useful only for developers, it would make sense to build it as a separate target, possibly along other things that we have in the project and that we don't build by default.
The time needed for the build is quite considerable, and I think that that the mechanism has its limitations s.a. not being aware of the templating system.

As a side note, the build produces so-called compiled rules, which are YAML files for all rules that are built for a particular product. Those rule files are located side-by-side in the `build/<product>/rules/` directory, they are "final" - all macros are expanded, and they contain deduced information that original rule yamls don't have. Maybe it would be possible to use those artifacts as an input for some other utilities as well?

cipherboy:
>  The time needed for the build is quite considerable, and I think that that the mechanism has its limitations s.a. not being aware of the templating system.
> As a side note, the build produces so-called compiled rules, which are YAML files for all rules that are built for a particular product. Those rule files are located side-by-side in the build/<product>/rules/ directory, they are "final" - all macros are expanded, and they contain deduced information that original rule yamls don't have. Maybe it would be possible to use those artifacts as an input for some other utilities as well?

Sorry, not sure I follow :-)

`rule_dir_json.py` is fairly quick and takes on the order of 12 seconds to run. It isn't the slowest portion of the build. Even building a single product in parallel (the default) it won't slow down the build much at all. Before and after this change, the build time doesn't increase much:

```
$ time ./build_product rhel8
real	1m13.629s
user	3m24.858s
sys	0m13.366s
$ time ./build_product rhel8
real	1m20.753s
user	3m47.656s
sys	0m13.590s
```

If you want, I can look into optimizing this some. I'm not quite sure why it takes this long. What would be an acceptable target? 5s?


I think `Debug` serves as a good target for content developers-only as any existing or future one. If you want, I can update it to run only in `Debug` builds. If you have another target that already exists, I'm happy to use it.

Lastly, I'm not entirely sure what your point is with the side note.

`rule_dir_json.py` provides a _mapping_ of `rule_id` to concrete location on disk, _in the content directory_. To refresh on the data structure, it is most closely a map of `rule_id` into objects:

```
id: rule_id
dir: absolute path on disk of this rule directory
guide: which guide it belongs to
products: which products this rule applies to
title: rule's title
ovals:
   map of oval's path to type, product (from file name), and computed applicable products and platforms
remediations:
   map of remediation's path to type, product (from file name), and computed applicable products and platforms
oval_products:
   map of product to ovals affecting the specified product
remedation_products:
   map of product to remediations affecting the specified product
```

It parses rule.yml and checks&ovals under the first product and parses the rule.yml and subsequent oval&checks under that. The assumption being that `prodtype` of rules and `platform`+`product` applicability of checks and fixes are more or less independent of the current environmental YAML and is more or less consistent across products.  

It does not:

 - Attempt to provide the full contents of the rule (in non-parsed form).
 - Attempt to parse the entire contents of the rule.
 - Attempt to provide the full parsed contents of checks and remediations.

Instead, it exists so that utilities that modify lots of rules have an easy way to look up:

 - What rules exist in the system,
 - What checks & ovals they have,
 - Where these rules are on disk,
 - If modifying a specific file (such as a check) -- having the _context_ the file should be parsed under. Oh it applies to Debian and Ubuntu? Probably ignore the file if we're modifying RHEL stuff. &c.

If you want, we can remove the title portion (under the assumption that certain rules like `clean_components_post_updating` would get parsed different (e.g., showing `Ensure apt_get Removes Previous Package Versions` for me) and that would be confusing on the whole). Nothing is using it right now... I think we were originally going to use it to check if rules had duplicated titles but that didn't yet materialize. I can add that functionality though if you think it'd be generally useful and is sufficient to keep the one piece of data that usually.

Additionally if you want, we can formalize the prodtype/platform/product assumptions with a test case and fail the build if content which doesn't respect this is introduced. It requires a N-way comparison of output though (for each rule with `n` prodtypes, we'd have to load all check/fix content and compare the computed platform & product applicability under each). It would be some work but doable. It will break other utilities operating on our content if this assumption doesn't hold. 

With all the utilities operating on it (`fix_rules.py` as of #6882, `refchecker.py` as of #6907, `autoprodtyper.py` as of #6906, `autorefer.py` in #6908 if it is included, `mod_prodtype.py`, `mod_checks.py`, `mod_fixes.py`, and `rule_dir_stats.py` and `rule_dir_diff.py`) -- it definitely makes sense to have the data stored in this format. With the exception of `rule_dir_stats` and `rule_dir_diff` and `refchecker.py`, none of these utilities could even operate on the `build/<product>/rules` directory -- without having to deal with nasty, nasty merges back into the existing content directory. 

My 2c., but I think we should optimize for developer/contributor experience here and this data structure is generally useful to improve it.

matejak:
Let me to react to elements of your reaction:

- build time of the json file: Those 15 seconds are fine given the purpose of the data. However, I wouldn't compare it to the rest of our build system that is very inefficient, then you could justify anything. My point would be that you build the content to use it with the scanner, or to run tests, and you may rebuild it if tests fail. Conversely, you need the json "database" if you want to heavily edit content metadata, and you don't want to rebuild it every time that you perform a build, as it is much less sensitive to typical code changes compared to a datastream.
In a nutshell - it's not about that number of seconds, but it is about being built by default although it is not needed, including CI and not very powerful machines that are used for training purposes. If you could get the time to 1/3 easily, then it would get to the rounding error, but I wouldn't do that.

- Separate target - I didn't mean debug builds - I meant something like `profile-stats`, when you are simply able to run `ninja profile-stats`, and it shows you the stats. I have been using the tool, and I even copied the json to the project directory, so it stays there, not needing to be rebuilt after the build_product erases everything in `build`.

- The json vs compiled rules - it was just a FYI. The json file contains everything in one file, while compiled rules are product-specific, and every rule ID is in a separate file - BTW a compiled rule knows its definition location as well. No need to change anything, I was just hoping that may find it useful.

So to sum it up, everything is fine, except that we don't like it to be built by default.

And could you please explain what you meant by

>Additionally if you want, we can formalize the prodtype/platform/product assumptions with a test case and fail the build if content which doesn't respect this is introduced. It requires a N-way comparison of output though (for each rule with n prodtypes, we'd have to load all check/fix content and compare the computed platform & product applicability under each). It would be some work but doable. It will break other utilities operating on our content if this assumption doesn't hold.

I don't know what you mean by prodtype/platform/product assumptions that could be verified automatically? I think that examples would help.

cipherboy:
> I don't know what you mean by prodtype/platform/product assumptions that could be verified automatically? I think that examples would help.

The existing `rule_dir_json.py` script's `rule_dirs.json` output _assumes_ that, given a parsed OVAL/Ansible/... check under a _single_ product X, the contents of the (OS) _platform_<->_product_ mapping won't change. The actual check/fix content might change (which isn't included in the JSON blob and thus doesn't matter), but not the check/fix's applicability.

That is to say, given a OVAL/Ansible/.. content Z for rule X:

    \forall product A \in all_products, and 
      \forall product B != A \in all_products, then
        X_Z eval (A).applicability == X_Z eval (B).applicability

This can be tested formally (on a really long-running test) in pseudo-Python like so:

```python
for guide_dir in all_shared_guide_dirs:
  for rule_dir in find_rules(guide_dir):
    all_products = products_for_shared_dir(guide_dir)
    our_prod = all_products[0]
    for their_prod in all_prodcuts[1:]:
      for obj in get_checks_and_fixes(rule_dir):
        our_obj = obj.eval_with_context(our_prod_ctx)
        their_obj = obj.eval_with_context(their_prod_ctx)
        assert our_obj.applicability == their_obj.applicability 
```

I don't think this was ever explicitly documented anywhere, and I'm curious if it actually holds.

jan-cerny:
@matejak @cipherboy Thanks for your opinions. I also lean towards not building it by default. Otherwise it's fine.

cipherboy:
> > I don't know what you mean by prodtype/platform/product assumptions that could be verified automatically? I think that examples would help.
> 
> The existing `rule_dir_json.py` script's `rule_dirs.json` output _assumes_ that, given a parsed OVAL/Ansible/... check under a _single_ product X, the contents of the (OS) _platform_<->_product_ mapping won't change. The actual check/fix content might change (which isn't included in the JSON blob and thus doesn't matter), but not the check/fix's applicability.
> 
> That is to say, given a OVAL/Ansible/.. content Z for rule X:
> 
> ```
> \forall product A \in all_products, and 
>   \forall product B != A \in all_products, then
>     X_Z eval (A).applicability == X_Z eval (B).applicability
> ```
> 
> This can be tested (on a really long-running test) in pseudo-Python like so:
> 
> ```python
> for guide_dir in all_shared_guide_dirs:
>   for rule_dir in find_rules(guide_dir):
>     all_products = products_for_shared_dir(guide_dir)
>     our_prod = all_products[0]
>     for their_prod in all_prodcuts[1:]:
>       for obj in get_checks_and_fixes(rule_dir):
>         our_obj = obj.eval_with_context(our_prod_ctx)
>         their_obj = obj.eval_with_context(their_prod_ctx)
>         assert our_obj.applicability == their_obj.applicability 
> ```
> 
> I don't think this was ever explicitly documented anywhere, and I'm curious if it actually holds.

There's a couple of easy counter-examples for `title` differing, though it doesn't matter too much:

```
jre/guide/benchmark.yml:4:title: Guide to the Secure Configuration of {{{ full_name }}}
jre/guide/java/java_jre_clean_previous_version/rule.yml:3:title: 'Ensure {{{ pkg_manager }}} Removes Previous Package Versions'
chromium/guide/benchmark.yml:4:title: Guide to the Secure Configuration of {{{ full_name }}}
vsel/guide/benchmark.yml:4:title: Guide to the Secure Configuration of {{{ full_name }}}
apple_os/benchmark.yml:4:title: Guide to the Secure Configuration of {{{ full_name }}}
fuse6/guide/benchmark.yml:4:title: Guide to the Secure Configuration of {{{ full_name }}}
applications/benchmark.yml:4:title: Guide to the Secure Configuration of {{{ full_name }}}
firefox/guide/benchmark.yml:4:title: Guide to the Secure Configuration of {{{ full_name }}}
linux_os/guide/benchmark.yml:4:title: Guide to the Secure Configuration of {{{ full_name }}}
linux_os/guide/system/bootloader-grub2/non-uefi/file_owner_grub2_cfg/rule.yml:5:title: 'Verify {{{ grub2_boot_path }}}/grub.cfg User Ownership'
linux_os/guide/system/bootloader-grub2/non-uefi/file_permissions_grub2_cfg/rule.yml:5:title: 'Verify {{{ grub2_boot_path }}}/grub.cfg Permissions'
linux_os/guide/system/bootloader-grub2/non-uefi/file_groupowner_grub2_cfg/rule.yml:5:title: 'Verify {{{ grub2_boot_path }}}/grub.cfg Group Ownership'
linux_os/guide/system/software/updating/ensure_gpgcheck_globally_activated/rule.yml:5:title: 'Ensure gpgcheck Enabled In Main {{{ pkg_manager }}} Configuration'
linux_os/guide/system/software/updating/clean_components_post_updating/rule.yml:5:title: 'Ensure {{{ pkg_manager }}} Removes Previous Package Versions'
linux_os/guide/system/software/updating/ensure_gpgcheck_never_disabled/rule.yml:5:title: 'Ensure gpgcheck Enabled for All {{{ pkg_manager }}} Package Repositories'
```

Issue #6913, opened 2021-04-29
OCP4 cis-node profile is broken due to openvswitch change


JAORMX:
#### Description of problem:

The e2e job `e2e-aws-ocp4-cis-node` is failing as of recently due to a recent change in openvswitch.

The logs look as follows:

```
 INFO[2021-04-28T17:57:41Z]     helpers.go:695: Result - Name: e2e-cis-node-worker-file-groupowner-ovs-pid - Status: FAIL - Severity: medium 
INFO[2021-04-28T17:57:41Z]     helpers.go:698: E2E-FAILURE: The expected result for the file_groupowner_ovs_pid rule didn't match. Expected 'PASS', Got 'FAIL' 
INFO[2021-04-28T17:57:41Z]     helpers.go:695: Result - Name: e2e-cis-node-worker-file-groupowner-ovs-sys-id-conf - Status: PASS - Severity: medium 
INFO[2021-04-28T17:57:41Z]     helpers.go:760: Rule file_groupowner_ovs_sys_id_conf matched expected result 
INFO[2021-04-28T17:57:41Z]     helpers.go:695: Result - Name: e2e-cis-node-worker-file-groupowner-ovs-vswitchd-pid - Status: FAIL - Severity: medium 
INFO[2021-04-28T17:57:41Z]     helpers.go:698: E2E-FAILURE: The expected result for the file_groupowner_ovs_vswitchd_pid rule didn't match. Expected 'PASS', Got 'FAIL' 
INFO[2021-04-28T17:57:41Z]     helpers.go:695: Result - Name: e2e-cis-node-worker-file-groupowner-ovsdb-server-pid - Status: FAIL - Severity: medium 
INFO[2021-04-28T17:57:41Z]     helpers.go:698: E2E-FAILURE: The expected result for the file_groupowner_ovsdb_server_pid rule didn't match. Expected 'PASS', Got 'FAIL' 
```
The rules expected the group to be openvswitch (of id 800), and a recent change in the openvswitch version shipped in RHCOS4 (based on RHEL 8.4) changed those pid files to be owned by the group hugetlbfs (id 801).

The content should be updated to reflect this.

PR #6914, opened 2021-04-29
ocp4: Enhance group ownership checks openvswitch processes pid files

JAORMX:
In a more recent version of Open vSwitch, the process pid files changed
ownership from "openvswitch" to "hugetlbfs". OCP 4.6 will stay in an
older version of OvS, so to keep this working with newer versions (e.g.
the one included in OCP 4.8), this fixes the content to check for both
groups. This way the content keeps being valid accross releases.

Closes https://github.com/ComplianceAsCode/content/issues/6913

### DIFF
--- a/applications/openshift/master/file_groupowner_ovs_vswitchd_pid/rule.yml
+++ b/applications/openshift/master/file_groupowner_ovs_vswitchd_pid/rule.yml
@@ -5,7 +5,9 @@ prodtype: ocp4
 title: 'Verify Group Who Owns The Open vSwitch Daemon PID File'
 
 description: |-
-    {{{ describe_file_group_owner(file="/run/openvswitch/ovs-vswitchd.pid", group="openvswitch") }}}
+    Ensure that the file <pre>/run/openvswitch/ovs-vswitchd.pid</pre>,
+    is owned by the group <pre>openvswitch</pre> or <pre>hugetlbfs</pre>,
+    depending on your settings and Open vSwitch version.
 
 rationale: |-
     CNI (Container Network Interface) files consist of a specification and libraries for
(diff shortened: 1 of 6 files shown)


applications/openshift/master/file_groupowner_ovs_vswitchd_pid/rule.yml
```diff
@@ -21,13 +23,26 @@ identifiers:
 references:
     cis@ocp4: 1.1.9
 
-ocil_clause: '{{{ ocil_clause_file_group_owner(file="/run/openvswitch/ovs-vswitchd.pid", group="openvswitch") }}}'
+ocil_clause: 'Is it the case that /run/openvswitch/ovs-vswitchd.pid has group owner openvswitch or hugetlbfs?'
```
ggbecker, line 26:
```suggestion
ocil_clause: '/run/openvswitch/ovs-vswitchd.pid has group owner openvswitch or hugetlbfs'
```

You may remove some parts of this sentence as they are added by the build system, from the `ocil_clause` [documentation](https://complianceascode.readthedocs.io/en/latest/manual/developer/06_contributing_with_content.html?highlight=ocil_clause) we have:


`
ocil_clause: This attribute contains the statement which describes how to determine whether the statement is true or false. Check out rule.yml in linux_os/guide/system/software/disk_partitioning/encrypt_partitions/: this contains a partitions do not have a type of crypto_LUKS value for ocil_clause. This clause is prefixed with the phrase Is it the case that <ocil_clause> ?.
`

ggbecker: This can be applied to the other occurrences of ocil_clause in this PR

JAORMX: Thanks! Will do

JAORMX: Done

yuumasato (approved):

PR #6915, opened 2021-04-29
Document controls

jan-cerny:
#### Description:

This PR adds documentation describing controls, controls files and using them in profiles. The PR aims to describe the current state of this feature.

#### Rationale:

During recent discussions with @cipherboy  we have realized that the controls and generated profiles aren't documented at all.

For example, https://github.com/ComplianceAsCode/content/blob/master/controls/anssi.yml is used to create RHEL ANSSI profiles which are shipped but in fact it is using an undocumented feature.

### DIFF
--- a/docs/manual/developer/03_creating_content.md
+++ b/docs/manual/developer/03_creating_content.md
@@ -355,3 +355,303 @@ A profile should define these attributes:
     or
 
 -   rule refinements, e.g. `accounts_tmout.severity=high`.
+
+
+## Controls
+
+The controls adds another layer on top of profiles. Controls files store the
+metadata for security controls and, more importantly, concentrate the mapping
+from requirement to rule at a single place.
+
+We will explain the format using the expected workflow and we will show examples
+of the format.
+
+### Storing controls data
+
+When we develop a new SCAP profile, usually we first get a PDF document with a
+policy that consists of controls (requirements). For example NIST 800-53
+controls, ANSSI requirements, CIS Benchmark, etc.
+
+To add the policy to our project repository, we will create a YAML
+file that represents this policy. There is a special directory, called
+`controls`, in the repository root to store these files. These files serve as a
+database of controls (requirements). They are independent from profiles and
+products. We can extract the interesting data from the PDF and save them in a
+YAML file. We can also create this file manually, or it can be converted from
+some other format (XML, OpenControl) if available and then submitted.
+
+There is a YAML file that represents the whole policy. Initially, the
+YAML file can look like this:
+
+```
+$ cat controls/abcd.yml
+
+id: abcd
+title: ABCD Benchmark for securing Linux systems
+version: 1.2.3
+source: https://www.abcd.com/linux.pdf
+controls:
+  - id: R1
+    title: User session timeout
+    description: |-
+      Remote user sessions must be closed after a certain
+      period of inactivity.
+  - id: R2
+    title: Minimization of configuration
+    description: |-
+      The features configured at the level of launched services
+      should be limited to the strict minimum.
+  - id: R3
+    title: Enabling SELinux targeted Policy
+    description: |-
+      It is recommended to enable SELinux in enforcing mode
+      and to use the targeted policy.
+```
+
+In the real world, controls (requirements) can be nested. For example, PCI-DSS
+has tree structure, within requirement 2.3 we can find 2.3.a, 2.3.b, etc.
+Therefore, each item in `controls` list can contain a `controls` list.
+
+Once we have the file, we can read through the policy requirements and assess
+each requirement. For the controls we will have to identify whether it can be
+automated by SCAP. If yes, we should look if we already have existing XCCDF
+rules in our project. 
+
+For example, let’s say that we identified that:
+
+* R1 can be automatically scanned by SCAP and we already have 3 existing rules
+in our repository. However, we want one of them to be selected only on RHEL 9,
+but the rule is applicable to all platforms.
+* R2 is up to manual checking, but we have systemd_target_multi_user which is
+related to this control.
+* R3 can be automatically scanned by SCAP but unfortunately we don’t have any
+rules and checks implemented yet.
+
+For each control we will add the `automated` key, which describes whether the
+control requirement can be automated by SCAP and scanning. Possible values are:
+`yes`, `no`, `partially`.
+
+When XCCDF rules exist, we will assign them to the controls. We will distinguish
+between XCCDF rules which directly implement the given controls (represented by
+`rules` YAML key) and rules that are only related or relevant to the control
+(represented by `related_rules` YAML key).
+
+The `rules` and `related_rules` keys consist of a list of rule IDs. If a rule
+needs to be chosen only in some of products despite its `prodtype` we can use
+Jinja macros inside the controls file to choose products.
+
+After we finish our analysys, we will insert our findings to the controls file,
+the file will look like this:
+
+```
+$ cat controls/abcd.yml
+ 
+id: abcd
+title: ABCD Benchmark for securing Linux systems
+version: 1.2.3
+source: https://www.abcd.com/linux.pdf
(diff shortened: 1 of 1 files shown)


vojtapolasek:
Well done. I have just two remarks:
1. Can you add example of nested controls?
2.  Does it have any effect on resulting datastream to mark rule as automatable, partially, or not automatable?
3. The same question as above for related_rules.

jan-cerny:
@vojtapolasek Thanks for the review. I have addressed the questions in the latest commit.

docs/manual/developer/03_creating_content.md
```diff
@@ -355,3 +355,372 @@ A profile should define these attributes:
...
+
+For example, let’s say that we identified that:
+
+* R1 can be automatically scanned by SCAP and we already have 3 existing rules
+in our repository. However, we want one of them to be selected only on RHEL 9,
+but the rule is applicable to all platforms.
+* R2 is up to manual checking, but we have systemd_target_multi_user which is
+related to this control.
+* R3 can be automatically scanned by SCAP but unfortunately we don’t have any
+rules and checks implemented yet.
```
cipherboy, line 428: Isn't "and checks" redundant as we'd have no way to have a check independent of a rule but still satisfying the control? That is, we could put a check under `shared/...` but we'd have no way of invoking it for the specified control without a rule entry?

cipherboy: But if you want to keep it: nit: `or checks` rather than `and checks` makes more sense to me.

docs/manual/developer/03_creating_content.md
```diff
@@ -355,3 +355,372 @@ A profile should define these attributes:
...
+          - var_password_pam_ocredit=1
+```
+
+### Using controls in profiles
+
+Later, we can use the policy requirements in profile YAML. Let’s say that we
+will define a “Desktop” profile built from the controls.
+
+To use controls, we add them under `selection` keys in the profile. The entry
+has the form `policy_id:control_id`.
```
cipherboy, line 635: Is this right? Looking at the ANSSI profiles, I think the form is:

```
policy_id:control_id[:level_id]
```

Where `level_id` is optional. 


(You state it here like it is the definitive form, but below you introduce another form -- I figured it makes sense to clear it up.)

docs/manual/developer/03_creating_content.md
```diff
@@ -355,3 +355,372 @@ A profile should define these attributes:
...
+For example, let's say that ABCD benchmark would define 2 levels: low and high.
+The low level would contain R1 and R2. The high level would contain everything
+from the low level and R3, ie. the high level would contain R1, R2 and R3.
+
+First, add the `levels` key to the YAML file and list the level names from
+lowest to highest. The tools working with the controls file assume that every
+control included in a level are also included in all subseqent levels listed in
+the `levels` list.
+
+Second, add `level` key to every control ID to specify level the control belongs to.
```
cipherboy, line 509: Completely out-of-scope to this PR but CIS is organized as follows:

- Level 1 Server
    - Level 2 Server inheriting all of Level 1 Server's rules
- Level 1 Workstation
    - Level 2 Workstation inheriting all of Level 1 Workstation's rules

Notably, there are two separate hierarchy of levels, or rather, that there's two separate "tags" (workstation, server) that are completely independent.

How would this be represented in a control file? My understanding is we'd need two separate control files (one for Workstation one for Server), which limits the amount of reuse between them.

Could levels be converted to an arbitrary hierarchy? Something like:

```yaml
levels:
 - low
 - medium
 - high
```
for a simple hierarchy and:

```yaml
levels:
 - id: l1-server
 - id: l1-workstation
 - id: l2-server
   inherit:
     - l1-server
 - id: l2-workstation
   inherit:
     - l1-workstation
```

For a custom-defined hierarchy with arbitrary nesting?

Then `level` in the section identifier itself would also have to take a list.


Just a thought.

docs/manual/developer/03_creating_content.md
```diff
@@ -355,3 +355,372 @@ A profile should define these attributes:
...
+selections:
+  - abcd:all
+  - security_patches_uptodate
+```
+
+It is possible to use levels if the levels are defined in the controls file. For
+example, `abcd:all:low` selects all rules for the ABCD low level or
+`abcd:all:high` selects all rules from the ABCD high level.
+
+Finally, when we build the content we will automatically get a SCAP profile
```
cipherboy, line 678: `s/content we/content, we/`

docs/manual/developer/03_creating_content.md
```diff
@@ -355,3 +355,372 @@ A profile should define these attributes:
...
+which contains XCCDF rules and variables from all controls selected in profile
```
cipherboy, line 679: This isn't explicitly mentioned above so I figured I'd call it out: `variables` is only discussed here. Can you select variables in the control file's `rules` section? Should that section be renamed if it isn't strictly only rules and is more or less similar to the Profile shorthand? What if they differ based on level?

e.g., Level 1 Server/Workstation in CIS would take AppArmor EnforceOrComplain variable with either value (enforce or complain) -- but Level 2 Server/Workstation could only take Enforce setting.

Would we just ignore the value of the variable in level 2? &c.

docs/manual/developer/03_creating_content.md
```diff
@@ -355,3 +355,372 @@ A profile should define these attributes:
...
+`sshd_set_idle_timeout`, `accounts_tmout`, `var_accounts_tmout=10_min` and
+`security_patches_uptodate`. The profile will not contain
+`systemd_target_multi_user` even if control `R2` is selected because that is
+listed under `related_rules`.  The profile will be compiled to a canonical
+form.
+
+Example of a compiled profile:
+
+```
+$ cat build/rhel8/profiles/abcd-desktop.profile
```
cipherboy, line 697: You might also want to call this out explicitly somewhere, but for debugging, knowing and listing this path explicitly (and how to find it) would be helpful, IMO.

docs/manual/developer/03_creating_content.md
```diff
@@ -355,3 +355,372 @@ A profile should define these attributes:
...
+  This profile contains configuration checks that align to
+  the ABCD benchmark.
+selections:
+  - sshd_set_idle_timeout
+  - accounts_tmout
+  - var_accounts_tmout=10_min
+  - security_patches_uptodate
+```
+
+We got a profile file that we could create manually and it's used seamleassly to
```
cipherboy, line 711: Maybe wordsmith this a little? Something like:

> ```markdown
> This profile is similar in content to one we could've created manually, but instead
> is automatically generated during build from a semantic data source. It seamlessly
> integrates with the build system to include the generated profile in the resulting
> SCAP data stream.
> ```

Also note typo: `s/seamleassly/seamlessly/g`.

cipherboy (commented):
I spent some time reading the docs to make sure I understood them and made a bunch of suggestions along the way. Feel free to take or leave whatever you want. :-)

Overall, makes sense to me so feel free to merge if you don't want to address my comments or want to do it at a later time.

(Perfect is the enemy of good enough after all... :D)

docs/manual/developer/03_creating_content.md
jan-cerny: Yes, it would make sense. It's a great idea.

docs/manual/developer/03_creating_content.md
jan-cerny: Yes, variables can be selected. In one of the examples we select `var_accounts_tmout=10_min`. It's more or less the same as the profile selections. I agree the key should be renamed, ideally to `selections`. It isn't possible to have the different selections in different levels at this moment. It would make sense to add the option to have it somehow.

jan-cerny:
@cipherboy Thanks for the review, it's very helpful.

cipherboy (approved):
Looks great, thanks @jan-cerny for documenting this!

Issue #6901, opened 2021-04-27
Approved sshd ciphers: dependency on installed_OS_is_FIPS_certified?


cipherboy:
#### Description of problem:

I'm working on adding Ubuntu 20.04 CIS rules. CIS has the notion of an "approved" set of SSH ciphers/kex/MACs.

However, the `sshd_approved_ciphers` and `sshd_approved_macs` rules both depend on `installed_OS_is_FIPS_certified`. Is this strictly required?

Ubuntu does have FIPS support (see https://ubuntu.com/security/certifications) so my issue is not there (I am working out language and will propose an upstream PR to add it to `installed_OS_is_FIPS_certified` eventually). More, my thoughts are that the `sshd_approved_ciphers` rules appear to be better written (e.g., having a variable for allowed ciphers) than the corresponding "non-FIPS" `sshd_use_strong_ciphers` rules.

I'd prefer to drop the dependency on `installed_OS_is_FIPS_certified` and have profiles which strictly require FIPS to pull in `installed_OS_is_FIPS_certified` when necessary. And also, drop the "strong" versions of the rules in favor of the "approved"+variable versions.

Thoughts?

#### SCAP Security Guide Version:

`master` as of 5b5fd67837246ffa293f107c58c6e448646c3292.

lordofhyphens:
I believe that DISA STIG also has moved to the "approved" list (at least it's own scanner is checking for an explicit list of ciphers/kex/MACs, so I don't think it'd go unused for RHEL.

cipherboy:
> I believe that DISA STIG also has moved to the "approved" list (at least it's own scanner is checking for an explicit list of ciphers/kex/MACs, so I don't think it'd go unused for RHEL.

I do agree but note that DISA STIG (being part of DoD) has strict requirements to use FIPS. :-) So the dependency from `sshd_approved_ciphers` onto `installed_OS_is_FIPS_certified` (in the OVAL) makes perfect sense in the DISA profile case but not in the e.g., CIS case which is used by public sector.

marcusburghardt:
FYI @dodys

dodys:
> FYI @dodys

@marcusburghardt that question is from Alex when he was working at Canonical, so it is not a question strict to Ubuntu but rather to the whole of ComplianceAsCode.

If this doesn't fit the project intention then we should close this ticket.

PR #6921, opened 2021-04-29
Broaden sshd_use_approved_* to non-FIPS use cases

cipherboy:
#### Description

Allow `sshd_use_approved_ciphers` to satisfy both FIPS and non-FIPS use
cases with the same rule. Benchmarks like CIS encourage the usage of
approved cipher suites as well (and do mention FIPS), but don't strictly
limit the caller to FIPS. For DISA STIG, other rules are more applicable
like sshd_use_approved_ciphers_ordered_stig so this change should be
safe.

Remove the dependency on the installed_OS_is_FIPS_certified OVAL as
well.

Also apply to sshd_use_approved_macs.

Resolves: #6901

```
Co-Authored-By: Richard Maciel Costa <richard.maciel.costa@canonical.com>
```

#### Rationale

As mentioned in #6901, we have duplicate rules for strictly DISA-STIG systems (`sshd_use_approved_ciphers_ordered_stig` and `sshd_use_approved_macs_ordered_stig`). Benchmarks like CIS can use but don't strictly require FIPS mode. But all benchmarks have the notion of an "approved" list of ciphers. So replace most instances of FIPS with Approved.

Note that I left the list of FIPS ciphers in place (along with the FIPS warning message) as the CIS benchmark does include FIPS-approved cipher listings.

### DIFF
--- a/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers/oval/shared.xml
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers/oval/shared.xml
@@ -1,8 +1,7 @@
 <def-group>
   <definition class="compliance" id="sshd_use_approved_ciphers" version="1">
-    {{{ oval_metadata("Limit the ciphers to those which are FIPS-approved.") }}}
+    {{{ oval_metadata("Limit the ciphers to those which are approved.") }}}
     <criteria operator="AND">
-      <extend_definition comment="Installed OS is FIPS certified" definition_ref="installed_OS_is_FIPS_certified" />
       <criteria comment="SSH is configured correctly or is not installed"
       operator="OR">
         <criteria comment="sshd is not installed" operator="AND">
@@ -56,5 +55,5 @@
     </split>
   </local_variable>
 
-  <external_variable comment="SSH Approved Ciphers by FIPS" datatype="string" id="sshd_approved_ciphers" version="1" />
+  <external_variable comment="SSH Approved Ciphers" datatype="string" id="sshd_approved_ciphers" version="1" />
 </def-group>
--- a/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers/rule.yml
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers/rule.yml
@@ -2,13 +2,13 @@ documentation_complete: true
 
 prodtype: ol7,ol8,rhel7,rhel8,wrlinux1019,wrlinux8,sle12
 
-title: 'Use Only FIPS 140-2 Validated Ciphers'
+title: 'Use Only Approved Ciphers'
 
 description: |-
-    Limit the ciphers to those algorithms which are FIPS-approved.
+    Limit the ciphers to those algorithms which are approved.
     Counter (CTR) mode is also preferred over cipher-block chaining (CBC) mode.
     The following line in <tt>/etc/ssh/sshd_config</tt>
-    demonstrates use of FIPS-approved ciphers:
+    demonstrates use of approved ciphers:
     <pre>Ciphers aes128-ctr,aes192-ctr,aes256-ctr,aes128-cbc,3des-cbc,aes192-cbc,aes256-cbc</pre>
     The man page <tt>sshd_config(5)</tt> contains a list of supported ciphers.
 {{% if product in ["rhel7","ol7"] %}}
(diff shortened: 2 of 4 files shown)


JAORMX (approved):

yuumasato:
I don't fully understand the FIPS use cases and some of the nuances...

But historically, the `sshd_use_approved_*` rules have been about FIPS approved algorithms, and the `sshd_use_strong_*` rules came in to be the counter part without  FIPS verbiage. So I'd tend to improve the `strong` rules instead of the `approved` ones... But I don't have a strong opinion here.
Do @carlosmmatos @ggbecker have thoughts on this?

And about the `approved` rules having a check for FIPS approved OSes. This came in from people in the field, that despite the warnings in the rule people still got confused and thought they were FIPS compliant because the rule was passing.
I'd really be interested to hear more stories about this, whether the OS check is useful or just frustrates people.

cipherboy:
> This came in from people in the field, that despite the warnings in the rule people still got confused and thought they were FIPS compliant because the rule was passing.

Hmmm...

So one of the things this does is drop "FIPS" from the title of the rule so hopefully that will be less-so the case.

However, I'm not suggesting that `installed_OS_is_FIPS_certified` be removed from all profiles that it is relevant in, merely that it (continue to) be explicitly included and be a separate and independent rule in its own right.

Rules like `enable_fips_mode` should definitely exist and depend on `installed_OS_is_FIPS_certified`, so I'm not sure how (unless they're building their own profile?) they'd end up with _this rule_ being what convinces them of FIPS support, without well, the other rules that include FIPS in their name. :D 

I dunno. :-)

Regarding the OS check in general, I think it is useful to weed out the CentOS/Rocky/Scientific Linux users (or in general, when a new OS is released like RHEL 9 but isn't yet certified from day-zero) -- in case they assume the OS is FIPS certified this will help confirm it isn't. But more OSes are undergoing FIPS certification and might not necessarily be tracked here. Ubuntu was the obvious missing case but maybe there's more elsewhere?

carlosmmatos (commented):
This is fine to me. I agree that there is no need to be explicit about the fact of using FIPS-140 ciphers. We have the required regulatory statements in the rule.yml's that say this. This allows more general usage and flexibility of the rule. If we need to "explicitly" toss FIPS verbiage around, then we should probably create a new rule or target the verbiage against products where its customer base might get confused without seeing the words FIPS.

jan-cerny:
@cipherboy any plans?

cipherboy:
Hi @jan-cerny, no plans from me :-) @dodys might be better if he's still at Canonical. But if there's enough community sentiment (it seemed like Carlos agreed), I'm more than happy to have someone else pick this up (@Mab879? :D)

dodys:
> Hi @jan-cerny, no plans from me :-) @dodys might be better if he's still at Canonical. But if there's enough community sentiment (it seemed like Carlos agreed), I'm more than happy to have someone else pick this up (@Mab879? :D)

I'm trying to catch-up on this, but is there anything else needed?

cipherboy:
@dodys Not that I can tell; from what I recall of discussions at the time, it was left for internal discussions at RH to decide if they were fine to remove it or not. From a project perspective, it made sense roughly that FIPS compliance should be a separate line item than "safe" SSH cipher suite selection and further that SSH cipher suite selection could be restricted in cases other than FIPS mode.

But IIRC, there was some discussion of their own benchmark plans if they were going to broaden this rule from strict FIPS compliance or not.

Maybe it just needs a rebase and we can merge it? If so, happy to do the rebase.

dodys:
> @dodys Not that I can tell; from what I recall of discussions at the time, it was left for internal discussions at RH to decide if they were fine to remove it or not. From a project perspective, it made sense roughly that FIPS compliance should be a separate line item than "safe" SSH cipher suite selection and further that SSH cipher suite selection could be restricted in cases other than FIPS mode.
> 
> But IIRC, there was some discussion of their own benchmark plans if they were going to broaden this rule from strict FIPS compliance or not. Otherwise, we might end up creating a separate role entirely (with a different name...); something I think we've tried to avoid in this project...
> 
> Maybe it just needs a rebase and we can merge it? If so, happy to do the rebase.

Yeah, I can also do it, if you want me to.

But it would be nice to have some confirmation from @yuumasato and others, that there isn't anything else impacting this on their side.

jan-cerny:
It seems that there is nothing happening recently, so I will closed this due to inactivity. But you can reopen it if you still want this change.

PR #6926, opened 2021-04-30
Introduce support for the distributed SSHd configuration

matejak:
As we can see in recent Fedora versions, the sshd configuration is not contained in the config file, but the config file just includes whatever is in the `/etc/ssh/sshd_config.d/` directory.

This PR adds this mechanics for Fedora and the RHEL9 products, and it introduces a rule that makes sure that such directive exists in the main `/etc/ssh/sshd_config` file.


JAORMX (approved):
awesomeness!! Wish we had this in RHCOS already

linux_os/guide/services/ssh/ssh_server/sshd_use_directory_configuration/rule.yml
```diff
@@ -0,0 +1,35 @@
...
+ocil_clause: "you don't include other configuration files from the main configuration file"
+
+# A setp-by-step guide how to modify the configuration to achieve compliance
+ocil: |-
+    To determine whether the SSH server includes configuration files from the right directory, run the following command:
+    <pre>$ sudo grep -i '^Include' /etc/ssh/sshd_config</pre>
+    If a line <tt>Include /etc/ssh/sshd_config.d/*</tt> is returned, then the configuration file inclusion is set correctly.
+
+template:
+    name: lineinfile
```
cipherboy, line 32: I don't think this does what you want sadly :/

The reason the `sshd`-specific template was introduced was due to `Match` blocks. Reading the `sshd_config` man page:

> An `Include` directive may appear inside a `Match` block to perform conditional inclusion.

So you're almost better off leaving `sshd_lineinfile` alone, updating it to take an optional file and using `sshd_lineinfile` here and then stub out a `sshd_config` that does what you've updated below (either using `sshd_lineinfile` directly on `/etc/ssh/sshd_config` or on a config directory). :/

cipherboy (changes requested):
Sometimes I really dislike the `Match` syntax of SSH config files but at least it is consistent... :-)

I like the intent of this though!

linux_os/guide/services/ssh/ssh_server/sshd_use_directory_configuration/rule.yml
matejak: I see. An alternative would be to extend this rule with a check that would make sure that there is no `Match` in the main config file - the direction set by default sshd config file in Fedora indicates that the preferable configuration should be done by the config files from the config directory, and that the main config file should just include them and do nothing else.
As an alternative to what you have suggested could be rebranding `sshd_lineinfile` to `sshd_config`, and just using macros for this rule. In other words, the `sshd_lineinfile` template in its current form would simply disappear - why to have a template, if there is only one rule using it?

cipherboy: I don't know about other products (such as RHEL8 and the SUSE versions) but at least we're currently preferring the main config file in our CIS and STIG for Ubuntu 20.04... So to technically be STIG compliant here, we'd prefer to keep a `sshd_lineinfile` around that can set stuff (and is cognizant of `Match`) in the main `/etc/ssh/sshd_config` file... Even if we do move to `sshd_config.d` for other newer content eventually. 16.04 is just nearing the ESM phase (where we mostly intend to stop rebasing to new CIS & STIG versions and definitely won't be seeing a CaC backport) --- so I'd expect 20.04 to be around for the next 4/5 years too with the current ruleset. :/

matejak: The change that I have introduced allows products to opt-in for the sshd config dir behavior. IOW, merging the PR shouldn't affect other products than Fedora and RHEL9, so a rule can behave differently based on the target product without changing its definition.

cipherboy: Ah right, I saw that and ignored it. :facepalm: 

Yes, so then I think it is just a question of how you want to handle this one-off weird thingy. I think directly calling macros from `oval/...` and `bash/...` is as good as anything to me.

cipherboy: (Or maybe introduce another rule as well, forbidding `Match` sections when this rule is selected?)

linux_os/guide/services/ssh/ssh_server/sshd_use_directory_configuration/rule.yml
```diff
@@ -0,0 +1,35 @@
...
+severity: medium
+
+identifiers:
+    cce@rhel9: CCE-84156-9
+
+# The second part of the sentence explaining what got wrong.
+# ... Is it the case that <YOUR ocil_clause>
+ocil_clause: "you don't include other configuration files from the main configuration file"
+
+# A setp-by-step guide how to modify the configuration to achieve compliance
```
jan-cerny, line 25:
```suggestion
# A step-by-step guide how to modify the configuration to achieve compliance
```

JAORMX:
Huh, there seem to be some regressions. This is from the `ci/prow/e2e-aws-rhcos4-e8` job:

```
 INFO[2021-04-30T11:18:09Z]     helpers.go:695: Result - Name: e2e-e8-master-sshd-disable-rhosts - Status: FAIL - Severity: medium 
INFO[2021-04-30T11:18:09Z]     helpers.go:698: E2E-FAILURE: The expected result for the sshd_disable_rhosts rule didn't match. Expected 'PASS', Got 'FAIL'
```

JAORMX:
The same failure appears in the `moderate` job.

linux_os/guide/services/ssh/ssh_server/sshd_use_directory_configuration/rule.yml
matejak: Could you please point me out to handling of the `Match` block? I took a look into definitions in `macros-oval.jinja`, and I couldn't find anything like that there.

matejak:
There are some mysterious failures - the rule should work exactly the same on other products than Fedora and RHEL9, and it also fails to build (presumably on Python2 systems) with a really weird error message.
I will investigate those.

linux_os/guide/services/ssh/ssh_server/sshd_use_directory_configuration/rule.yml
cipherboy: For Ansible it was here:

https://github.com/ComplianceAsCode/content/blob/master/shared/macros-ansible.jinja#L145-L156

I don't know if it ever got finished for shell. I don't think it was considered for OVAL because regardless of whether or not a value is in a Match block or not, it should be a finding if the value is wrong. But if the value is correct and the option _has_ to be present, then it must live outside a Match block, which makes the logic a little harder.

At the time I was contributing, I was most familiar with Ansible/Bash and not familiar at all with OVAL.

So perhaps my comments here are only relevant to the Ansible portion of the template. 

I'm not quite following this PR right now, but I think you didn't touch the Ansible? At least, that's what its looking like to me -- in which case, the Ansible wouldn't support `sshd_config.d` folders.

matejak: You are right that the new system won't work with Ansible. I have checked out what Ansible can do for us here, and I think that not much - there is still no sshd module, and the lineinfile module needs a filename rather than a regex with an indication of where to place potential lines. IOW, it looks like that we will need an Ansible "program" with variables, ifs (whens) and so on in order to handle this.
I wouldn't consider this lack of Ansible support as a PR blocker, because it is a missing feature rather than a introduced regression - a new configuration practice has appeared out there, and the content project just tries to catch up step by step.

yuumasato:
> There are some mysterious failures - the rule should work exactly the same on other products than Fedora and RHEL9, and it also fails to build (presumably on Python2 systems) with a really weird error message.
> I will investigate those.

@matejak Was there any conclusion on the mysterious failures?

yuumasato (commented):
@matejak Sorry, my misteake I made a comment but did not finish the review.

matejak:
> @matejak Was there any conclusion on the mysterious failures?

Yes, they were related to Jinja constructs that were not supported everywhere, and they were fixed.

linux_os/guide/services/ssh/ssh_server/sshd_use_directory_configuration/bash/shared.sh
```diff
@@ -1,12 +1,15 @@
 # platform = multi_platform_all
 
 {{% set target_file = "/etc/ssh/sshd_config.d/sshd_config_original.conf" -%}}
+{{% set base_config = "/etc/ssh/sshd_config" -%}}
 if test -f {{{ target_file}}}; then
 	{{{ die("Remediation probably already happened, '" ~ target_file ~ "' already exists, not doing anything.", action="false") }}}
+elif grep -Eq '^\s*Include\s+/etc/ssh/sshd_config\.d/\*\.conf' {{{ base_config }}} && ! grep -Eq '^\s*Match\s' {{{ base_config }}}; then
+	{{{ die("Remediation probably already happened, {{{ base_config }}} already contains the include directive.", action="false") }}}
```
yuumasato, line 8: This line needs some more Jinja foo, the remediation is printing:
`Remediation probably already happened, {{{ base_config }}} already contains the include directive.`

matejak: Good catch!

yuumasato (approved):

Issue #6925, opened 2021-04-30
Duplicate STIGID for RHEL 8 / require_emergency_target_auth


cipherboy:
#### Description of problem:

In `require_emergency_target_auth`:

```yaml
references:
    stigid@rhel7: RHEL-07-010481
    stigid@rhel8: RHEL-08-010151   # <-- Here?
    stigid@ol7: OL07-00-010481
    stigid@rhel8: RHEL-08-010152   # <-- Or here?
```

I'm not sure how to resolve, but it'll be a blocker for #6882. @carlosmmatos -- mind taking a look? I've got no idea which to choose... :) Thanks!

If I merge them together on one key I get:

```
RuntimeError: Error normalizing 'require_emergency_target_auth': Rules can not have multiple STIG IDs.
```

#### SCAP Security Guide Version:

`master` as of https://github.com/ComplianceAsCode/content/commit/6b8a62c644fbc4da1e08fa0cb7a8098057398e01.

ggbecker:
This one is bit tricky, it should be `RHEL-08-010152`, but this identifier is not released yet (we are working together with DISA to have better defined STIGs). So I think we should keep `RHEL-08-010151` for the moment since it maps to something that is already released (even though we will have a duplicate), see: https://github.com/ComplianceAsCode/content/blob/b0ba64583302d5cbf5d8cabe6ecccb7da99df57e/linux_os/guide/system/accounts/accounts-physical/require_singleuser_auth/rule.yml#L47

carlosmmatos:
Yup - @ggbecker is right - we jumped the gun on this one so that's my bad if I put that in there. Waiting for tests to complete on #6928 before closing this.

cipherboy:
Thanks! @ggbecker and @carlosmmatos :) I'll work on rebasing #6882 now that everything's landed...

PR #6928, opened 2021-04-30
Revert STIG id for require_emergency_target_auth.

ggbecker:
#### Description:

- Revert STIG id for require_emergency_target_auth.

#### Rationale:

- Keep the STIG id that is currently released. This new STIG id will be released in V1R3.

Fixes: #6925


vojtapolasek:
LGTM

PR #6930, opened 2021-04-30
Remove support statement from the manual page

jan-cerny:
This statement is outdated. It says that SSG is shipped in EPEL and that
it isn't supported by Red Hat. We don't ship it EPEL for many years.
When the man page is shipped in RHEL it isn't true that it isn't
supported by Red Hat. When the man page is packaged in other
distributions, the information about support by Red Hat is redundant.
The package usually follows the distro's support policy.  Other packages
also don't claim their support level in their manual page. People should
refer to the distribution support documentation instead.

### DIFF
--- a/docs/man_page_template.jinja
+++ b/docs/man_page_template.jinja
@@ -93,27 +93,6 @@ Contains Ansible Playbooks for SSG profiles.
 Contains Bash remediation scripts for SSG profiles.
 .RE
 
-.SH STATEMENT OF SUPPORT
-The SCAP Security Guide, an open source project jointly maintained by Red Hat
-and the NSA, provides XCCDF and OVAL content for Red Hat technologies. As an open
-source project, community participation extends into U.S. Department of Defense
-agencies, civilian agencies, academia, and other industrial partners.
-
-SCAP Security Guide is provided to consumers through Red Hat's Extended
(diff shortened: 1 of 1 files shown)


vojtapolasek:
Looks good, thank you. I see there are still some parts mentioning Red Hat specifically.
e.g. line 10, line 98, line 127
then there are sentences referencing rhel6, line 118 and 123
These parts are included even if the man page is built for non RH products. Is that desirable?

jan-cerny:
@vojtapolasek Nice catch.

I have removed the first occurrence of Red Hat which was in the "Description" section.

However, the other occurrences of Red Hat are in DEPLOYMENT TO U.S. CIVILIAN GOVERNMENT SYSTEMS and DEPLOYMENT TO U.S. MILITARY SYSTEMS. I'm not sure about these sections. They probably should be limited to builds for RHEL or they can be removed as well. @carlosmmatos Do you have some insights about these sections?

vojtapolasek:
Also on line 54 in the example there is ssg-rhel7-ds.xml explicitly mentioned. Later in the text, the rhel is replaced with Jinja macro for product. Can we put it also here?

vojtapolasek:
I believe that failing OCP test is not related to the change in the man page. Merging.

PR #6935, opened 2021-05-01
Add RHEL OS profiles to the rule/check to look for silent option in pam_lastlog

### LINKED ISSUE #6902: RHEL-07-040530 fails check after remediation with upstream XCCDF
#### Description of problem:
The DISA STIG also requires (at least w/r/t the SCAP checklist from pub.cyber.mil) that the `silent` flag is not present for `pam_lastlog.so` (as the STIG clearly states "must display" for STIG ID RHEL-07-040530).

This causes a failure when the system is later rescanned with the DISA checklist and OVAL definitions. 

master @ 5b5fd67837246ffa293f107c58c6e448646c3292

Red Hat Enterprise Linux  7.9

#### Steps to Reproduce:

1. Scan and remediate a RHEL7 system with `content_profile_stig`. 
2. Re-scan the system with DISA's own checklist and  the MAC1-Classified profile.

#### Actual Results:
V-204605 is marked as failed by DISA's SCAP benchmark (https://dl.dod.cyber.mil/wp-content/uploads/stigs/zip/U_RHEL_7_V3R3_STIG_SCAP_1-2_Benchmark.zip)

#### Expected Results:
V-204605 is not marked as failed and the `silent` flag is not present.

lordofhyphens:
#### Description:

- Added the check for `silent` in `display_login_attempts` to the OVAL when building for RHEL7/8/9, which was already present for SLE.

#### Rationale:
- [DISA STIG](https://www.stigviewer.com/stig/red_hat_enterprise_linux_7/2017-12-14/finding/V-72275) requires that silent not be present.
- NIST (ncp) [AC-7](https://www.stigviewer.com/controls/800-53/AC-9) indicates that it should also not be silenced.
- PCI DSS 10.2.4 is silent on the matter, so altering the behavior should be fine. 
- Fixes #6902

### DIFF
--- a/linux_os/guide/system/accounts/accounts-pam/display_login_attempts/oval/shared.xml
+++ b/linux_os/guide/system/accounts/accounts-pam/display_login_attempts/oval/shared.xml
@@ -3,7 +3,7 @@
     {{{ oval_metadata("Configure the system to notify users of last login/access using pam_lastlog.") }}}
     <criteria operator="AND">
       <criterion comment="Conditions for pam_lastlog are satisfied" test_ref="test_display_login_attempts" />
-      {{% if product in ['sle12', 'sle15'] %}}
+      {{% if product in ['sle12', 'sle15', 'rhel7', 'rhel8', 'rhel9'] %}}
       <criterion comment="silent option for pam_lastlog is set" test_ref="test_display_login_attempts_silent" />
       {{% endif %}}
     </criteria>
--- a/linux_os/guide/system/accounts/accounts-pam/display_login_attempts/rule.yml
+++ b/linux_os/guide/system/accounts/accounts-pam/display_login_attempts/rule.yml
@@ -6,7 +6,7 @@ description: |-
     To configure the system to notify users of last logon/access
     using <tt>pam_lastlog</tt>, add or correct the <tt>pam_lastlog</tt>
     settings in
-    {{% if product in ["sle12", "sle15"] %}}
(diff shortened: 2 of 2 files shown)


vojtapolasek:
I am afraid that the fix won't be so simple. By enabling this criteria, you will be checking for contents in the file /etc/pam.d/login, whereas stig talks about /etc/pam.d/postlogin.

jan-cerny:
> I am afraid that the fix won't be so simple. By enabling this criteria, you will be checking for contents in the file /etc/pam.d/login, whereas stig talks about /etc/pam.d/postlogin.

@vojtapolasek Do you request any changes?

vojtapolasek:
@jan-cerny  yes I do. The proposed implementation does not align with STIG. It correctly checks if the "silent" option is not present for the pam_lastlog.so module, but in a wrong file.

ggbecker:
This has been fixed in https://github.com/ComplianceAsCode/content/commit/803585c7254437587dcbeabce0f2d4f2fcf58022 already. 

Also the https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/accounts/accounts-pam/display_login_attempts/oval/shared.xml is now much cleaner and uses the configuration file depending on the OS. I will close this, if you disagree you can reopen it.

lordofhyphens:
No, I agree it should be fixed. I'll do a cross-check against DISA to validate that it's "fixed" to their satisfaction and follow up as necessary.

PR #6916, opened 2021-04-29
Document rule.yml modification utilities

cipherboy:
`Signed-off-by: Alexander Scheel <alex.scheel@canonical.com>`

---

This provides documentation for `mod_prodtype.py` and the utilities/modifications to utilities in #6882, ~~#6906~~, #6907, and #6908.

It should probably wait to merge until all of those utilities have merged.

PR #6936, opened 2021-05-01
Add support for Rocky Linux

Mab879:
#### Description:

This pull request adds support for Rocky Linux, which is currently in preview.

#### Rationale:

This pull request will enable Rocky Linux users to run scans on their Rocky Linux machines.


omok314159:
Hi, @Mab879 
We've already create ComplianceAsCode contents I think.
https://github.com/atomicturtle/openscap/tree/main/ComplianceAsCode
If you will create new rockylinux8 contents, it's better to merge our ComplianceAsCode contents.

OMO

omok314159:
It seems your contents approach is that "Add RockyLinux same as CentOS contents in RHEL config". My Approach is "Add RockyLinux as same as OracleLinux or other distribution". So, now I'm asking in Rocky's Mattermost #security channel.(I asked same questions several month ago in same channel.)

cipherboy:
Hi @omok314159 :)

Speaking for myself and not my past / present employers -- and perhaps badly guessing at what others do... :D

And also, dropping by since I was @Mab879's reference for his first PR to CaC, he's looking to get involved more in the upstream project and is also running Rocky Linux (he mentioned your MM channel).

---

It all really depends on the approach you want to take, in my view.

If you're looking to replicate CentOS's behavior, creating a derivative product is definitely the way to go. You'll mostly get bug-for-bug what CentOS delivered, and you'll pick up rebranded RHEL content. A derivative product (Rocky) inherits all content from the base product (RHEL), just with a changed CPE and a few warnings in some places (e.g., RHEL is FIPS certified, but CentOS wasn't). This is the closest to the RHEL experience you can get, without running RHEL.

If, however, you have enough of a team to maintain your own content sets -- and want to diverge potentially significantly from RHEL's -- then the approach of Oracle or SUSE is definitely the way to go. Here you'd be able to choose which rules you want to bring in, which compliance profiles, &c. It _could_ be bug for bug, but you'd either have to ask the Red Hat folks here to add `rocky8` to all new `rhel8` rules or you'd have to keep a close eye on the upstream repo and do so. If you have the team and are looking to deliver different content -- this is definitely the way to do.

If your team is smaller, I'd probably suggest taking the former approach and if you have any content changes, contribute them upstream (here) to the RHEL content sets so they'll get picked up next rebase.

---

However, looking at your [actual build](https://github.com/atomicturtle/openscap/blob/main/ComplianceAsCode/content_for_supporting_rocky8/tools/add_product_rocky8.sh#L10-L12) scripts... It looks like you're really taking the approach here -- just using `sed` and some other hacks, whereas the build system will allow you to do all that directly. i.e., you're creating a derivative product _manually_. :)

That approach would let you eventually fork the content (like OL) but keep on top of what RHEL 8 is doing for the time being. However, there's a lot of utilities that I've added to CaC that you've likely missed due to [lack of documentation](https://github.com/ComplianceAsCode/content/pull/6916) that helps with all this.


Hope that help explains the differences and why I suggested Matt go with this approach for an upstream contribution. 

Oh and feel free to hang out in `#openscap` on Freenode if you're packaging or contributing to OpenSCAP and ComplianceAsCode!

Mab879:
All,  thanks for the feedback. I couple points that I want to make:

1. From a quick look at the RHEL7 content it seems that Scientific Linux was added in similar way to how I added Rocky Linux to the RHEL8 content.
1. I think there is some value in adding Rocky Linux content here vs downstream. If the Rocky Linux content lived in this repo (As a part of the RHEL content or as its owns) then users of other distributions can use the Rocky Linux content in tools like scap-workbench.

Overall, I'm open to whatever approach the community wants.

omok314159:
Hi, @cipherboy @Mab879 
Thanks for your comment. Actually I suggested to Rocky community to contribute in ComplianceAsCode on Dec 2020 because I know it's the best way to contribute in upstream. In that time, we were not ready to contribute in ComplianceAsCode because of legal problem I think. Then we had plan to submit the contents later. 
Also I'll inform to community that your suggestion and ask which is a better way for Rocky(Actually we are discussing it now).

omok314159:
Hi, @cipherboy @Mab879 , we need to discuss this(how can we maintain, team, and so on) more in Rocky project. Then I think the better way is "(just for now) add Rocky Linux as same as CentOS" by this PR. And later we will change the scheme in ComplianceAsCode.

[5 years ago, I discussed this kind of topic](https://lists.fedorahosted.org/archives/list/scap-security-guide@lists.fedorahosted.org/message/WN7JUCIQ3PREEFN2SGJT7SBHRJODXUPO/) for CentOS when this ComplianceAsCode project was called as "Scap Security Guide". So, I wish to make sure how RocyLinux will consider about ComplianceAsCode scheme.

Mab879:
Closing this out. I might open this again with better community support from Rocky Linux.

PR #6939, opened 2021-05-03
Update dpkg package installed command

cipherboy:
#### Description:

Previously the command was:

    dpkg-query -s {0} &>/dev/null

This was replaced with:

    dpkg-query --show --showformat='${db:Status-Status}\n' '{0}' 2>/dev/null | grep -q installed

The rationale for this change is `dpkg-query` alone isn't sufficient for
detecting whether or not a package is installed. In particular, if the
package used to be installed but was removed, dpkg-query will still exit
with a status code of 0 (thus, succeeding), but will display output similar
to:

    Status: deinstall ok config-files

Outputting just the field we want (`${db:Status-Status}`) and then
quietly grepping for the status we want (`installed`) gives us the final
result we want.

```
Co-Authored-By: Richard Maciel Costa <richard.maciel.costa@canonical.com>
```

---

/cc @richardmaciel-canonical

### DIFF
--- a/ssg/build_remediations.py
+++ b/ssg/build_remediations.py
@@ -28,7 +28,8 @@ REMEDIATION_TO_EXT_MAP = {
 }
 
 PKG_MANAGER_TO_PACKAGE_CHECK_COMMAND = {
-    'apt_get': 'dpkg-query -s {0} &>/dev/null',
+    'apt_get': "dpkg-query --show --showformat='${{db:Status-Status}}\\n' '{0}' 2>/dev/null " +
+               "| grep -q installed",
     'dnf': 'rpm --quiet -q {0}',
     'yum': 'rpm --quiet -q {0}',
     'zypper': 'rpm --quiet -q {0}',


JAORMX (approved):

cipherboy:
Fix the commit message to clarify that `dpkg-query` actually returns status code 0 when the package is removed (thus, succeeding). 

Thanks @richardmaciel-canonical!

JAORMX (approved):

yuumasato (approved):
LGTM.

@cipherboy Just want to check, is this behavior is the same for Debian and Ubuntu?
I'd assume yes, but better ask.

cipherboy:
Yeah:

```
$ cr debian:latest
$ apt update
$ apt install -y git
$ dpkg-query -s git ; echo $?
--> 0
$ dpkg-query -s libjss-java ; echo $?
--> 1
$ apt remove -y git
$ dpkg-query -s git ; echo $?
Status: deinstall ok config-files
--> 0
$ dpkg-query --show --showformat='${db:Status-Status}' git | grep -q installed ; echo $?
--> 1
$ dpkg-query --show --showformat='${db:Status-Status}' libjss-java | grep -q installed ; echo $?
dpkg-query: no packages found matching libjss-java
--> 1
$ dpkg-query --show --showformat='${db:Status-Status}' bash | grep -q installed ; echo $?
--> 0
```

I think the only question is if we should silence stderr to hide the error from a package that hasn't yet been installed (`libjss-java` above). Thoughts?

yuumasato:
> I think the only question is if we should silence stderr to hide the error from a package that hasn't yet been installed (`libjss-java` above). Thoughts?

Good point, as the check of package installed is for applicability purposes and not for the remediation itself I'd try to silence them, as they can be misleading. 
Plus, these apt-get checks for package installed are more meaningful when using the Bash scripts, and could add unnecessary and misleading noise to a already very long log. (When using --remediate, the rule applicability will guarantee that the remediation checks are passing).

cipherboy:
@yuumasato Updated to include `2>/dev/null` redirect on the `dpkg-query` part. Thanks!

yuumasato (approved):
> Thanks!

Well, thank you! Happy to see Debian/Ubuntu content being uplifted.

cipherboy:
Seems like OpenShift/AWS failures:

```

INFO[2021-05-04T12:19:27Z] level=error msg=Error: [ERR]: Error building changeset: InvalidChangeBatch: [Tried to create resource record set [name='api.ci-op-4rsxt58h-da1ef.origin-ci-int-aws.dev.rhcloud.com.', type='A'] but it already exists] 
INFO[2021-05-04T12:19:27Z] level=error msg=	status code: 400, request id: 56fb94e0-72a7-4033-9e7e-4c1b72abe584 
INFO[2021-05-04T12:19:27Z] level=error                                  
INFO[2021-05-04T12:19:27Z] level=error msg=  on ../tmp/openshift-install-042506361/route53/base.tf line 42, in resource "aws_route53_record" "api_external_alias": 
INFO[2021-05-04T12:19:27Z] level=error msg=  42: resource "aws_route53_record" "api_external_alias" { 
```

cipherboy:
\o/ CI passed. I'm going ahead and merging between @yuumasato's review and @richardmaciel-canonical's out-of-band ACK.

PR #6940, opened 2021-05-03
Add rule for SLES-15-030110

teacup-on-rockingchair:
#### Description:
- Adopt audit_rules_privileged_commands_unix_chkpwd rule to cover the specification for SLES-15-030110

#### Rationale:

- Add relvant IDs and references to the rule
- Add to templates support to handle audit rules both for unix_chkpwd unix2_chkpwd tools 
  - note specifics of the location of the tools
  - note the option to support both tools instead of just unix_chkpwd as in original implementation
  - test against sle15 show that parsing the auditctl need to take into account numeric representation of the  auid and also the '-S all' option
- Add audit_rules_privileged_commands_unix_chkpwd to STIG profile for SLE15
- Assign CCE id CCE-85727-6
- Update docs for audit_rules_privileged_commands

### DIFF
--- a/linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_unix2_chkpwd/tests/ocp4/e2e.yml
+++ b/linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_unix2_chkpwd/tests/ocp4/e2e.yml
@@ -1,3 +1,0 @@
----
-default_result: FAIL
-result_after_remediation: PASS
--- a/linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_unix2_chkpwd/tests/ocp4/e2e.yml
+++ b/linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_unix2_chkpwd/tests/ocp4/e2e.yml
@@ -1,0 +1,3 @@
+---
+default_result: FAIL
+result_after_remediation: PASS
--- a/linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_unix_chkpwd/rule.yml
+++ b/linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_unix_chkpwd/rule.yml
@@ -40,9 +40,9 @@ identifiers:
 references:
     stigid@ol7: OL07-00-030640
     cui: 3.1.7
-    disa: CCI-000135,CCI-000172,CCI-002884
+    disa: CCI-000130,CCI-000135,CCI-000169,CCI-000172,CCI-002884
     hipaa: 164.308(a)(1)(ii)(D),164.308(a)(3)(ii)(A),164.308(a)(5)(ii)(C),164.312(a)(2)(i),164.312(b),164.312(d),164.312(e)
-    nist: AC-2(4),AU-2(d),AU-12(c),AC-6(9),CM-6(a)
+    nist: AC-2(4),AU-2(d),AU-3,AU-3.1,AU-12(a),AU-12(c),AU-12.1(ii),AU-12.1(iv),AC-6(9),CM-6(a),MA-4(1)(a)
     nist-csf: DE.CM-1,DE.CM-3,DE.CM-7,ID.SC-4,PR.PT-1
     ospp: FAU_GEN.1.1.c
     srg: SRG-OS-000042-GPOS-00020,SRG-OS-000392-GPOS-00172,SRG-OS-000471-GPOS-00215,SRG-OS-000037-GPOS-00015
@@ -54,8 +54,6 @@ references:
     cobit5: APO10.01,APO10.03,APO10.04,APO10.05,APO11.04,BAI03.05,DSS01.03,DSS03.05,DSS05.02,DSS05.04,DSS05.05,DSS05.07,MEA01.01,MEA01.02,MEA01.03,MEA01.04,MEA01.05,MEA02.01
     iso27001-2013: A.12.4.1,A.12.4.2,A.12.4.3,A.12.4.4,A.12.7.1,A.14.2.7,A.15.2.1,A.15.2.2
     cis-csc: 1,12,13,14,15,16,2,3,5,6,7,8,9
-    nist@sle15: AU-3,AU-3.1,AU-12(a),AU-12.1(ii),AU-12(c),AU-12.1(iv),MA-4(1)(a)
-    disa@sle15: CCI-000130,CCI-000169
     stigid@sle15: SLES-15-030110
 
 ocil_clause: 'it is not the case'
--- a/linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_unix_chkpwd/rule.yml
+++ b/linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_unix_chkpwd/rule.yml
@@ -68,4 +68,4 @@ template:
     name: audit_rules_privileged_commands
     vars:
         path: /usr/sbin/unix_chkpwd
-        path@sle15: (/sbin/unix_chkpwd|/sbin/unix2_chkpwd)
+        path@sle15: /sbin/unix_chkpwd
--- a/linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_unix_chkpwd/rule.yml
+++ b/linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_unix_chkpwd/rule.yml
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: fedora,ol7,ol8,rhcos4,rhel7,rhel8,rhel9,rhv4,sle12,wrlinux1019
+prodtype: fedora,ol7,ol8,rhcos4,rhel7,rhel8,rhel9,rhv4,sle12,sle15,wrlinux1019
 
 title: 'Ensure auditd Collects Information on the Use of Privileged Commands - unix_chkpwd'
 
@@ -35,6 +35,7 @@ identifiers:
     cce@rhel8: CCE-80740-4
     cce@rhcos4: CCE-82609-9
     cce@sle12: CCE-83109-9
+    cce@sle15: CCE-85727-6
 
 references:
     stigid@ol7: OL07-00-030640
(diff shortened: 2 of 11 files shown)


vojtapolasek:
@openscap-ci  test this pleasee

linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_unix_chkpwd/rule.yml
```diff
@@ -53,6 +54,10 @@ references:
     cobit5: APO10.01,APO10.03,APO10.04,APO10.05,APO11.04,BAI03.05,DSS01.03,DSS03.05,DSS05.02,DSS05.04,DSS05.05,DSS05.07,MEA01.01,MEA01.02,MEA01.03,MEA01.04,MEA01.05,MEA02.01
     iso27001-2013: A.12.4.1,A.12.4.2,A.12.4.3,A.12.4.4,A.12.7.1,A.14.2.7,A.15.2.1,A.15.2.2
     cis-csc: 1,12,13,14,15,16,2,3,5,6,7,8,9
+    nist@sle15: AU-3,AU-3.1,AU-12(a),AU-12.1(ii),AU-12(c),AU-12.1(iv),MA-4(1)(a)
+    disa@sle15: CCI-000130,CCI-000169
+    srg@sle15: SRG-OS-000037-GPOS-00015
```
vojtapolasek, line 59: Please merge this with existing SRG references, they are product independent.

vojtapolasek (changes requested):
Hello, this is very interesting PR. The idea with multiple paths sounds good, however, it needs to be tested. Can you please create tests for the rule to see if it works as expected?
See other comments for minor remarks. Thank you.

linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_unix_chkpwd/rule.yml
```diff
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: fedora,rhcos4,ol7,ol8,rhel7,rhel8,rhv4,wrlinux1019,sle12
+prodtype: fedora,rhcos4,ol7,ol8,rhel7,rhel8,rhv4,wrlinux1019,sle12,sle15
```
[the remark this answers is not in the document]
teacup-on-rockingchair: Thanks for the feedback :bow:  will add tests in the following days

linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_unix_chkpwd/rule.yml
teacup-on-rockingchair: This should be handled in d42fdb0

linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_unix_chkpwd/rule.yml
```diff
@@ -53,6 +54,10 @@ references:
     cobit5: APO10.01,APO10.03,APO10.04,APO10.05,APO11.04,BAI03.05,DSS01.03,DSS03.05,DSS05.02,DSS05.04,DSS05.05,DSS05.07,MEA01.01,MEA01.02,MEA01.03,MEA01.04,MEA01.05,MEA02.01
     iso27001-2013: A.12.4.1,A.12.4.2,A.12.4.3,A.12.4.4,A.12.7.1,A.14.2.7,A.15.2.1,A.15.2.2
     cis-csc: 1,12,13,14,15,16,2,3,5,6,7,8,9
+    nist@sle15: AU-3,AU-3.1,AU-12(a),AU-12.1(ii),AU-12(c),AU-12.1(iv),MA-4(1)(a)
+    disa@sle15: CCI-000130,CCI-000169
```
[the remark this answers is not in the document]
teacup-on-rockingchair: Should be handled in e98ba87

linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_unix_chkpwd/rule.yml
```diff
@@ -53,6 +54,10 @@ references:
     cobit5: APO10.01,APO10.03,APO10.04,APO10.05,APO11.04,BAI03.05,DSS01.03,DSS03.05,DSS05.02,DSS05.04,DSS05.05,DSS05.07,MEA01.01,MEA01.02,MEA01.03,MEA01.04,MEA01.05,MEA02.01
     iso27001-2013: A.12.4.1,A.12.4.2,A.12.4.3,A.12.4.4,A.12.7.1,A.14.2.7,A.15.2.1,A.15.2.2
     cis-csc: 1,12,13,14,15,16,2,3,5,6,7,8,9
+    nist@sle15: AU-3,AU-3.1,AU-12(a),AU-12.1(ii),AU-12(c),AU-12.1(iv),MA-4(1)(a)
```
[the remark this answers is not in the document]
teacup-on-rockingchair: Should be handled in e98ba87

teacup-on-rockingchair:
After spending some time on this adding extra tests, kudos for @vojtapolasek on this feedback, found a major problem with the current approach, I've chosen in implementation. The definition of the rule for SUSE(sle15) context need to cover both `unix2_chkpwd` and `unix_chkpwd` in same rule.
So I am exploring three approaches at the moment:
1. Modify template to support list of paths, and generate oval, ansible, bash, kubernetes rules for each element of the template, when the path parameter is a list of commands. It complicates the code quite a lot and still not convinced if it worth the effort, from point of view that the code is going to be reused in future.  
2. Not using the `audit_rules_privileged_commands template` for the SUSE sle15 if that is possible
3. Use existing `audit_rules_privileged_commands_unix_chkpwd` rule to implement `unix_chkpwd` of the STIG definition and create additional one, `audit_rules_privileged_commands_unix2_chkpwd` for `unix2_chkpwd`,  where both rules will be referring to the same STIG rule, if that is possible

vojtapolasek:
I think option 3 is fine. We try to promote the approach "one rule = one configuration change", so that rules can be more easily reused among profiles. Sometimes this might result in one stigid being in multiple rules.

linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_unix2_chkpwd/tests/ocp4/e2e.yml
```diff
@@ -0,0 +1,3 @@
+---
```
vojtapolasek, line 1: This file is not needed because the rule is not applicable for OCP4.

shared/templates/audit_rules_privileged_commands/oval.template
```diff
@@ -24,7 +24,7 @@
   <ind:textfilecontent54_object id="object_{{{ ID }}}_augenrules" version="1">
     <ind:filepath operation="pattern match">^/etc/audit/rules\.d/.*\.rules$</ind:filepath>
 {{% if product in ["sle12", "sle15"] %}}
-    <ind:pattern operation="pattern match">^[\s]*-a[\s]+always,exit[\s]+-F[\s]+path={{{ PATH }}}(?:[\s]+-F[\s]+perm=x)?[\s]+-F[\s]+auid>={{{ auid }}}[\s]+-F[\s]+auid!=(?:4294967295|unset)[\s]+(?:-k[\s]+|-F[\s]+key=)[\S]+[\s]*$</ind:pattern>
+    <ind:pattern operation="pattern match">^[\s]*-a[\s]+always,exit[\s]+(-S[\s]+all[\s]+)*-F[\s]+path={{{ PATH }}}(?:[\s]+-F[\s]+perm=x)?[\s]+-F[\s]+auid\>={{{ auid }}}[\s]+-F[\s]+auid!=(?:4294967295|unset|-1)[\s]+(?:-k[\s]+|-F[\s]+key=)[\S]+[\s]*$</ind:pattern>
```
vojtapolasek, line 27: What is the purpose of this change?

teacup-on-rockingchair: Well the DISA document definitions for SUSE, I've got, used as example bellow line:
```

-a always,exit -S all -F path=/sbin/unix_chkpwd -F perm=x -F auid>=1000 -F auid!=-1 -k privileged-unix-chkpwd
-a always,exit -S all -F path=/sbin/unix2_chkpwd -F perm=x -F auid>=1000 -F auid!=-1 -k privileged-unix2-chkpwd
```
So I decided to include also the possibility that the audit rule may include system calls as specification. It is is not mandatory part of the regex so should not interfere with the previous definition.

vojtapolasek: I see. According to auditctl man page, using no -S or using -S all should be equivalent - Audit will watch for all syscalls in both cases. @stevegrubb  can you confirm this?
I see one important point which needs to be considered. By changing this template you are changing OVAL checks for many rules, 36 to be exact. I wonder if this should be somehow mirrored in their description. Because suddenly you will have passing rules where their Audit lines will not correspond with the rule description.
Thought son this? @ggbecker  @yuumasato

ggbecker: Since this is a change that doesn't really affect the meaning of the audit line, I think the description doesn't need to be update, it would cause more confusion in my opinion. RHEL STIG doesn't mention the usage of this parameter for example.

Changing the subject, there is also a new backslash added to `auid\>={{{ auid }}}`. Is it really necessary? (The `-1` added to the `auid!=` should be ok though)

linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_unix2_chkpwd/tests/ocp4/e2e.yml
teacup-on-rockingchair: File dropped at 73b5042

shared/templates/audit_rules_privileged_commands/oval.template
teacup-on-rockingchair: The backslash seems to be an over-precaution in the regex, dropped it in 6c087d6

stevegrubb: > According to auditctl man page, using no -S or using -S all should be equivalent - Audit will watch for all syscalls in
> both cases. @stevegrubb can you confirm this?

The syscall is actually selected by the perm field for watches. You can't see any evidence of this when listing the rules because its done in a different place than the syscall filter. I think that in the absence of specifying a syscall, auditctl defaults to all. You do not need to specify it.

teacup-on-rockingchair: Ok so if I understand correctly I can rely on 100% that `-S all` will not end up in the configuration file, so dropping that redundant group from the pattern

vojtapolasek (approved):

vojtapolasek:
@openscap-ci test this please

PR #6941, opened 2021-05-03
Make profile-tool product-aware

cipherboy:
#### Description:

This introduces a product-specific form of `profile_tool.py`, taking the
optional `--product` argument (without a default). When the product
directory exists and there's a `templates/constants.xslt` under it, we'll
load that and attempt to influence our STIG and CIS benchmark namespaces
appropriately.

---

See also discussion on https://github.com/ComplianceAsCode/content/pull/6907

### DIFF
--- a/ssg/build_profile.py
+++ b/ssg/build_profile.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 from __future__ import print_function
 
+import os
 import sys
 
 from .xml import ElementTree
@@ -14,7 +15,8 @@ from .constants import puppet_system as puppet_rem_system
 from .constants import anaconda_system as anaconda_rem_system
 from .constants import cce_uri
 from .constants import ssg_version_uri
-from .constants import stig_ns, cis_ns, hipaa_ns, anssi_ns, ospp_ns, cui_ns
+from .constants import stig_ns, cis_ns, generic_stig_ns, hipaa_ns, anssi_ns
+from .constants import ospp_ns, cui_ns, xslt_ns
 console_width = 80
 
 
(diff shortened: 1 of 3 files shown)


ssg/build_profile.py
```diff
@@ -14,7 +15,8 @@
 from .constants import anaconda_system as anaconda_rem_system
 from .constants import cce_uri
 from .constants import ssg_version_uri
-from .constants import stig_ns, cis_ns, hipaa_ns, anssi_ns, ospp_ns, cui_ns
+from .constants import stig_ns, cis_ns, generic_stig_ns, hipaa_ns, anssi_ns
```
cipherboy, line 18: Does uh, the STIG NS actually _do_ anything? 

```bash
$ gif 'https://public.cyber.mil/stigs/srg-stig-tools/'
ssg/constants.py:63:stig_ns = "https://public.cyber.mil/stigs/srg-stig-tools/"
```

The only place I see it is in the constants...

In the generated RHEL8 XCCDF I see:

```
          <reference href="https://public.cyber.mil/stigs/downloads/?_dl_facet_stigs=operating-systems%2Cunix-linux">RHEL-08-040124</reference>
```

Which is the new `generic_stig_ns` I introduced.... Seems like RHEL 8 needs to update its constant?

jan-cerny:
I'm somehow unable to get it working. I run

```
build-scripts/profile_tool.py stats --profile standard  --product debian9 --benchmark build/ssg-debian9-xccdf.xml  --missing-cis-refs 
```

I would expect it will print out many rules that don't have cis reference because there is actually one rule in the whole repository that has `cis@debian9` in references.

Is it just a wrong example? Or do I miss anything?

cipherboy:
Note that CIS and STIG both only run on the CIS/STIG named profile:

https://github.com/ComplianceAsCode/content/blob/master/ssg/build_profile.py#L280

That tripped me up as well.

jan-cerny:
OK, thanks for explanation. Debian 9 has no CIS profile :D So I tried the SLE15 and it works. Before it reported that 0 rules have CIS refs and with your patch and with --product option it reports 92 have them. Great!

PR #6942, opened 2021-05-03
Add platform check for RHEL minor version.

ggbecker:
#### Description:

- Add platform check for RHEL minor version.

#### Rationale:

- This rule should be applicable to RHEL 8.0 and 8.1 only (RHEL-08-020022)

#### Problems

- The current way our `platform` implementation works does not allow me to combine the RHEL platform check with `pam` check using AND operator, so ultimately this does not solve the problem. But I'm starting here to ask for more opinions on how to tackle this problem of applicability only to certain minor versions of RHEL (which is one of the problems to be solved with minor version agnostic content)

#### Disclaimer

- This is a very simplistic implementation. ;)

### DIFF
--- a/shared/checks/oval/installed_OS_is_rhel8_0.xml
+++ b/shared/checks/oval/installed_OS_is_rhel8_0.xml
@@ -1,0 +1,42 @@
+<def-group>
+  <definition class="inventory"
+  id="installed_OS_is_rhel8_0" version="1">
+    <metadata>
+      <title>Red Hat Enterprise Linux 8.0</title>
+      <affected family="unix">
+        <platform>multi_platform_all</platform>
+      </affected>
+      <reference ref_id="cpe:/o:redhat:enterprise_linux:8.0"
+      source="CPE" />
+      <description>The operating system installed on the system is
+      Red Hat Enterprise Linux 8.0</description>
+    </metadata>
+    <criteria>
+      <criterion comment="Installed operating system is part of the unix family"
+      test_ref="test_rhel8_0_unix_family" />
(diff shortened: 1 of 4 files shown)


shared/checks/oval/installed_OS_is_rhel8_0.xml
```diff
@@ -0,0 +1,42 @@
+<def-group>
```
cipherboy, line 1: Are `chared/checks/oval/*.xml` JINJA-enabled? Could these be macro-ized?

ggbecker: I wondering that as well, but I haven't checked that yet.

cipherboy: No worries, it is outside the scope of the PR :-) I might suggest it to @Mab879 if he wants to try his hand at another PR. It seems like a lot of the (Linux) OS-related version checks could be simplified.

JAORMX: They are jinja enabled. We use this for OpenShift https://github.com/ComplianceAsCode/content/blob/master/shared/checks/oval/installed_app_is_ocp4.xml#L46

carlosmmatos:
@ggbecker we've drafted a new PR #7251 to take this off your place. I'll let you close this when you get back.

ggbecker:
> @ggbecker we've drafted a new PR #7251 to take this off your place. I'll let you close this when you get back.

Closing out in favor of the new PR.

PR #6944, opened 2021-05-04
Improve the test suite wrapper

matejak:
It now uses a whole, literal CPE for the purpose of CPE injection into the datastream.

The previous way that used part of the CPE stopped working for some reason.

### DIFF
--- a/tests/test_rule_in_container.sh
+++ b/tests/test_rule_in_container.sh
@@ -185,7 +185,7 @@ script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)" || { echo "Couldn't d
 # [ <-- needed because of Argbash
 
 podman images | grep -q "$_arg_name" || die "Couldn't find the podman image '$_arg_name'"
-test_image_cpe_product=$(podman run --rm "$_arg_name" cat /etc/os-release | grep cpe | cut -d : -f 4)
+test_image_cpe_product=$(podman run --rm "$_arg_name" sh -c '. /etc/os-release && echo "$CPE_NAME"')
 test -n "$test_image_cpe_product" || die "Unable to deduce the product CPE from the container's /etc/os-release file."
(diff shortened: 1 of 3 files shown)


jan-cerny:
Looks fine.

> The previous way that used part of the CPE stopped working for some reason.

 In which situation? Do you have a reproducer?

tests/test_rule_in_container.sh
```diff
@@ -185,7 +185,7 @@ script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)" || { echo "Couldn't d
 # [ <-- needed because of Argbash
 
 podman images | grep -q "$_arg_name" || die "Couldn't find the podman image '$_arg_name'"
-test_image_cpe_product=$(podman run --rm "$_arg_name" cat /etc/os-release | grep cpe | cut -d : -f 4)
+test_image_cpe_product=$(podman run --rm "$_arg_name" cat /etc/os-release | grep CPE_NAME | cut -f 2 -d = | tr -d "'\"")
```
cipherboy, line 188: What about something like:

```
bash -c 'source /etc/os-release && echo $CPE_NAME'
```

`/etc/os-release` & friends were IIRC meant [to be source-able](https://www.freedesktop.org/software/systemd/man/os-release.html) and this is shorter. :-)

matejak: So that's not so simple as source is a bash builtin, but something very similar works.

matejak:
and another one that improves error reporting, esp. in connection with the Podman 2.2 issue.

carlosmmatos:
I can confirm this addresses and fixes the issues faced with Python versions < 3.8. Also nice to have additional verbiage surrounding the podman version fiasco. If more people hit this issue, it may be prudent to add a link to some doc that goes more in-depth around this.

LGTM

PR #6945, opened 2021-05-04
make sure that rule platforms is a set

vojtapolasek:
#### Description:

- convert list of rules read from rule.yml to set

#### Rationale:

I know that this might not seem nice but writing sets into yaml directly looks ugly, you can see it in compiled rules if you build some product with this PR.
But internally we need platforms to be a set to eliminate possible duplicities. So let's do it in the background in this way.

### DIFF
--- a/ssg/build_yaml.py
+++ b/ssg/build_yaml.py
@@ -1091,6 +1091,10 @@ class Rule(object):
                    .format(fname=yaml_file, err=str(exc)))
             raise RuntimeError(msg)
 
+        # platforms are read as list from the yaml file
+        # we need them to convert to set again
+        rule.platforms = set(rule.platforms)
+
         for warning_list in rule.warnings:
             if len(warning_list) != 1:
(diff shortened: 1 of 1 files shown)


jan-cerny:
@vojtapolasek I which situation this happens? What problem does it cause?

vojtapolasek:
Interestingly, it does not currently have any effect right now. But it can cause problems in the future, when we would like to for example merge rule.inherited_platforms and rule.platforms, expecting rule.platforms to be a set (for elimination of duplicates).
You can verify it by putting a print statement
```
print (rule.id_)
print (type(rule.platforms))
```
Somewhere after line 1093 in build_yaml.py.
Watch for the rule chronyd_or_ntpd_set_maxpol. It contains multiple platforms (as the only rule in the content so far).

jan-cerny:
OK, thanks for explanation, indeed, for the chronyd_or_ntpd_set_maxpoll there is a list.

PR #6947, opened 2021-05-04
SLES-15-010355 rule

rumch-se:
#### Description:

- _SLES-15-010355 rule_

#### Rationale:

- _Review required_

### DIFF
--- a/linux_os/guide/system/permissions/files/permissions_within_important_dirs/root_permissions_syslibrary_files/ansible/shared.yml
+++ b/linux_os/guide/system/permissions/files/permissions_within_important_dirs/root_permissions_syslibrary_files/ansible/shared.yml
@@ -1,0 +1,10 @@
+# platform = multi_platform_sle
+# reboot = false
+# strategy = restrict
+# complexity = high
+# disruption = medium
+
+- name: "Set ownership to root of system-wide library files"
+  command: "find {{ item }} ! -group root -type f -exec chgrp root '{}' \\;"
+  with_items: [ '/lib', '/lib64', '/usr/lib', '/usr/lib64' ]
+
--- a/linux_os/guide/system/permissions/files/permissions_within_important_dirs/root_permissions_syslibrary_files/oval/shared.xml
+++ b/linux_os/guide/system/permissions/files/permissions_within_important_dirs/root_permissions_syslibrary_files/oval/shared.xml
@@ -1,0 +1,27 @@
+<def-group>
+  <definition class="compliance" id="root_permissions_syslibrary_files" version="2">
+    {{{ oval_metadata("
+        Checks that system-wide library files in /lib, /lib64, /usr/lib, /usr/lib64
+        are owned by root.
+      ") }}}
+    <criteria >
+      <criterion test_ref="test_root_permissions_for_syslibrary_files" />
+    </criteria>
+  </definition>
+
+  <unix:file_test  check="all" check_existence="none_exist" comment="test if system-wide files have root permissions" id="test_root_permissions_for_syslibrary_files" version="1">
+    <unix:object object_ref="root_permissions_for_system_wide_library_files" />
+  </unix:file_test>
+
+  <unix:file_object comment="system-wide directories" id="root_permissions_for_system_wide_library_files" version="1">
+    <!-- Checks that system-wide library files in /lib, /lib64, /usr/lib, /usr/lib64
+        are owned by root. -->
+    <unix:path operation="pattern match">^\/lib[64?]|^\/usr\/lib[64?]</unix:path >    
+    <unix:filename operation="pattern match">^.*$</unix:filename>
+    <filter action="include">group_permissions_for_system_wide_files_are_not_root</filter>
+  </unix:file_object>
+
+  <unix:file_state id="group_permissions_for_system_wide_files_are_not_root" version="1" >
+    <unix:group_id datatype="int" operation="not equal">0</unix:group_id>
+  </unix:file_state>
+</def-group>
--- a/linux_os/guide/system/permissions/files/permissions_within_important_dirs/root_permissions_syslibrary_files/oval/shared.xml
+++ b/linux_os/guide/system/permissions/files/permissions_within_important_dirs/root_permissions_syslibrary_files/oval/shared.xml
@@ -16,7 +16,6 @@
   <unix:file_object comment="system-wide directories" id="root_permissions_for_system_wide_library_files" version="1">
     <!-- Checks that system-wide library files in /lib, /lib64, /usr/lib, /usr/lib64
         are owned by root. -->
-    <unix:behaviors recurse_direction="down" />
     <unix:path operation="pattern match">^\/lib(64)?|^\/usr\/lib(64)?</unix:path >    
     <unix:filename operation="pattern match">^.*$</unix:filename>
     <filter action="include">group_permissions_for_system_wide_files_are_not_root</filter>
--- a/linux_os/guide/system/permissions/files/permissions_within_important_dirs/root_permissions_syslibrary_files/oval/shared.xml
+++ b/linux_os/guide/system/permissions/files/permissions_within_important_dirs/root_permissions_syslibrary_files/oval/shared.xml
@@ -16,6 +16,7 @@
   <unix:file_object comment="system-wide directories" id="root_permissions_for_system_wide_library_files" version="1">
     <!-- Checks that system-wide library files in /lib, /lib64, /usr/lib, /usr/lib64
         are owned by root. -->
+    <unix:behaviors recurse_direction="down" />
     <unix:path operation="pattern match">^\/lib(64)?|^\/usr\/lib(64)?</unix:path >    
     <unix:filename operation="pattern match">^.*$</unix:filename>
     <filter action="include">group_permissions_for_system_wide_files_are_not_root</filter>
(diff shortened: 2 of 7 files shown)


vojtapolasek:

linux_os/guide/system/permissions/files/permissions_within_important_dirs/root_permissions_syslibrary_files/rule.yml
```diff
@@ -0,0 +1,55 @@
...
+
+description: |-
+    System-wide library files are stored in the following directories
+    by default:
+    <pre>/lib
+    /lib64
+    /usr/lib
+    /usr/lib64
+    </pre>
+    All system-wide shared library files should be protected from unautorised
```
vojtapolasek, line 17:
```suggestion
    All system-wide shared library files should be protected from unauthorised
```

linux_os/guide/system/permissions/files/permissions_within_important_dirs/root_permissions_syslibrary_files/rule.yml
```diff
@@ -0,0 +1,55 @@
...
+ocil_clause: 'any system wide library directory is returned'
+
+ocil: |-
+    System-wide library files are stored in the following directories:
+    <pre>/lib
+    /lib64
+    /usr/lib
+    /usr/lib64
+    </pre>
+    To find system-wide library files sored in these directories are group-owned by
```
vojtapolasek, line 53:
```suggestion
    To find if system-wide library files stored in these directories are group-owned by
```

linux_os/guide/system/permissions/files/permissions_within_important_dirs/root_permissions_syslibrary_files/rule.yml
```diff
@@ -0,0 +1,55 @@
...
+    root run the following command for each directory <i>DIR</i>s:
```
vojtapolasek, line 54:
```suggestion
    root run the following command for each directory <i>DIR</i>:
```

linux_os/guide/system/permissions/files/permissions_within_important_dirs/root_permissions_syslibrary_files/ansible/shared.yml
```diff
@@ -0,0 +1,10 @@
+# platform = multi_platform_sle
+# reboot = false
+# strategy = restrict
+# complexity = high
+# disruption = medium
+
+- name: "Set ownership to root of system-wide library files"
+  command: "find {{ item }} ! -group root -type f -exec chgrp root '{}' \\;"
```
vojtapolasek, line 8: afaik we are trying to avoid usage of command and shell if possible. Can this be possibly rewritten to use the file module?

linux_os/guide/system/permissions/files/permissions_within_important_dirs/root_permissions_syslibrary_files/oval/shared.xml
```diff
@@ -0,0 +1,27 @@
...
+
+  <unix:file_test  check="all" check_existence="none_exist" comment="test if system-wide files have root permissions" id="test_root_permissions_for_syslibrary_files" version="1">
+    <unix:object object_ref="root_permissions_for_system_wide_library_files" />
+  </unix:file_test>
+
+  <unix:file_object comment="system-wide directories" id="root_permissions_for_system_wide_library_files" version="1">
+    <!-- Checks that system-wide library files in /lib, /lib64, /usr/lib, /usr/lib64
+        are owned by root. -->
+    <unix:path operation="pattern match">^\/lib[64?]|^\/usr\/lib[64?]</unix:path >    
+    <unix:filename operation="pattern match">^.*$</unix:filename>
```
vojtapolasek, line 20: If you want to recurse directories, please use file behaviors.
You can se an example here:
https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/selinux/selinux_all_devicefiles_labeled/oval/shared.xml#L12

vojtapolasek (changes requested):
Hello, thank you for the new rule. Please see comments.

linux_os/guide/system/permissions/files/permissions_within_important_dirs/root_permissions_syslibrary_files/ansible/shared.yml
rumch-se: @vojtapolasek - Thank you for your proposal, but before I do some changes here I would like to clarify something. In current code the find command finds/collects the files without a correct group ownership and changes it to the correct one We need 1 loop. The second loop is nested indirectly. If we want to rewrite this code in ansible and to avoid bash - we will have 2 loops. With the 1st loop we will collect the files and with the 2nd we will apply the change on them.  If you can propose me a solution/sample with 1 loop with ansible I will adapt the current code to it. I think that in real situations we have to implement the rules asap - imagine a data center with 10K+ servers and more i.e. the speed of rules' implementation will play important role. The realization in ansible will be slower than the realization in bash. If for bash we need 5 sec and for ansible 2 seconds more - in case with 10K servers we will need only for this rule  333 minutes - 5.55h more!Are we looking for speed and efficiency or for implementation which will be based on a specific language? Any feedback will be welcomed - because I used this approach in a few rules and in this case it will be necessary to change my realization(this part of rule's code) in all of them.

vojtapolasek: I think this might work:
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/file_module.html
```
- name: Recursively change ownership of a directory
  ansible.builtin.file:
    path: {{ item }}
    state: directory
    recurse: yes
    group: root
```
I understand your concerns. Can you test this?

rumch-se: @vojtapolasek  - According to the rule - we have to change the group ownership of a *file*  not of a directory => we have to collect all files in the /lib /lib64 /usr/lib and /usr/lib64 with owner != root and for them to apply change the group owner to root. The ansible file - doesn't provide ability to collect files. The ansible find - provides ability to collect files in a dictionary, but this dictionary does not have an attribute group. If you  can advice me how from a dictionary created by find to extact a file and to change its group - it will solve the problem. I was trying with this - and the result   files_matched - has a very complex structure 
 
- name: "Find system-wide library files"
  find:
    paths: "{{ item }}"
    file_type: "file"
    hidden: yes
  register: files_matched  
  with_items: [ '/lib' , '/lib64', '/usr/lib', '/usr/lib64' ]

https://docs.ansible.com/ansible/latest/collections/ansible/builtin/find_module.html

vojtapolasek: Hi, I am sorry, I misunderstood you before. You are right, the file module will set ownership for directories and files. I think using the command module is therefore fine in this case.

linux_os/guide/system/permissions/files/permissions_within_important_dirs/root_permissions_syslibrary_files/oval/shared.xml
vojtapolasek: This is the only remaining part to be fixed. This is excerpt from the OVAL specification:
Note that the default recurse_direction behavior is 'none' so even though max_depth specifies no limitation by default, the recurse_direction behavior turns recursion off.

rumch-se: Hi @vojtapolasek  - I have implemented the proposed change and I have corrected the regex.

vojtapolasek (approved):

vojtapolasek:
@openscap-ci ok to test

vojtapolasek:
@openscap-ci test this please

vojtapolasek:
@openscap-ci ok to test

ggbecker:
@openscap-ci test this please

linux_os/guide/system/permissions/files/permissions_within_important_dirs/root_permissions_syslibrary_files/oval/shared.xml
```diff
@@ -0,0 +1,28 @@
...
+
+  <unix:file_test  check="all" check_existence="none_exist" comment="test if system-wide files have root permissions" id="test_root_permissions_for_syslibrary_files" version="1">
+    <unix:object object_ref="root_permissions_for_system_wide_library_files" />
+  </unix:file_test>
+
+  <unix:file_object comment="system-wide directories" id="root_permissions_for_system_wide_library_files" version="1">
+    <!-- Checks that system-wide library files in /lib, /lib64, /usr/lib, /usr/lib64
+        are owned by root. -->
+    <unix:behaviors recurse_direction="down" />
+    <unix:path operation="pattern match">^\/lib(64)?|^\/usr\/lib(64)?</unix:path >
```
ggbecker, line 20: Unfortunately it's not possible to use recursion with pattern match:

`
oval:ssg-root_permissions_for_system_wide_library_files:obj:1 - the recurse_direction behavior MUST not be used when a pattern match is used with a path entity.
`

I believe you have to create an object for each of the directories you are looking for (`/lib`, `/lib64`, `/usr/lib` and `/usr/lib64`)

ggbecker: also we might wanna double check if recursion is also needed in other similar rules to this one.

ggbecker: > also we might wanna double check if recursion is also needed in other similar rules to this one. Maybe other rules will have the same behavior, such as the one being proposed here: https://github.com/ComplianceAsCode/content/pull/7002

vojtapolasek:
@openscap-ci test this please

PR #6948, opened 2021-05-05
Update VSEL references to remove qualifier from global references

ggbecker:
#### Description:

- Update VSEL references to remove qualifier from global references.

#### Rationale:

- Since we introduced this build condition #6896, some of the references cannot have product qualifier as they are so called global references.
- We missed these references because those rules belong only to VSEL product and our PR gating doesn't build that product ;(

### DIFF
--- a/vsel/guide/vsel/general_settings/dats_auto_update/rule.yml
+++ b/vsel/guide/vsel/general_settings/dats_auto_update/rule.yml
@@ -19,9 +19,9 @@ rationale: |-
 severity: medium
 
 references:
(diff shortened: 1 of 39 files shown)


JAORMX (approved):

ggbecker:
@openscap-ci retest this please

cipherboy (approved):
Ah my bad, I figured CI would build everything. 

I see all the vsel references and this appears to be all of them. ACK'd.

cipherboy:
Building all products works locally from this branch. CI passed, so merging.

PR #6953, opened 2021-05-05
Updating macros to support idempotency when deduplicating values

carlosmmatos:
#### Description:

- Current ansible lineinfile macros do not support idempotent behavior when invoking the `ansible_only_lineinfile` macro.

#### Rationale:

- If we're using Ansible, might as well use it better. Ensure that the rules that use these macros provide the end-user with a better Ansible experience.


jan-cerny:
The change looks fine but we need some test scenarios that test this macro. For example in rule accounts_tmout.

carlosmmatos:
> The change looks fine but we need some test scenarios that test this macro. For example in rule accounts_tmout.

Sounds good.. I'll look into adding these this morning.

PR #6966, opened 2021-05-06
Enable RHEL9 for kernel-related rules

matejak:
No sysctl variables were deprecated.
Rules for kernel modules are about disabling them, so there is little reason not to enable them for RHEL9 as well. Besides that, all modules are relevant, except the `dccp` and `freevxfs` ones.


jan-cerny:
@matejak I can't see any RHEL 9 in the sysctl rules.

matejak:
> @matejak I can't see any RHEL 9 in the sysctl rules.

Good catch! Fixed now.

PR #6969, opened 2021-05-07
Add rules to remove setroubleshoot server and plugin packages

yuumasato:
#### Description:

- Added rules to remove setroubleshoot-plugins and server.

#### Rationale:

- Helps satisfy ANSSI-BP-028 R68


JAORMX:
@yuumasato is this only used in the ANSSI profiles? Was wondering if this would be a relevant check to add to RHCOS. While setroubleshoot is not packaged nor enabled, at least we could verify that someone hasn't done an unsupported override.

yuumasato:
> @yuumasato is this only used in the ANSSI profiles? 

At the moment, and as far as I know yes.

> Was wondering if this would be a relevant check to add to RHCOS. While setroubleshoot is not packaged nor enabled, at least we could verify that someone hasn't done an unsupported override.

I see, I'll add them to RHCOS too.

vojtapolasek:
Hello, thank you for rules. Two comments:
1. Both rules are missing ocils.
2. Do you plan to add rules to the high profile later?

yuumasato:
@vojtapolasek Thanks for review, I added the OCIL clauses using macro.

I intended to add the rules later, but I see no harm in doing it now.

vojtapolasek:
lgtm, thank you

PR #6978, opened 2021-05-07
Check libpam-runtime on Ubuntu

cipherboy:
On Debian-based systems, the `pam` package doesn't exist as a binary
package. `libpam-runtime` is the equivalent, depending on `libpam0g` which
provides the actual binaries. Many packages depend on `libpam-runtime`,
making it the preferred package to check for.

/cc @richardmaciel-canonical

### DIFF
--- a/shared/checks/oval/installed_env_has_pam_package.xml
+++ b/shared/checks/oval/installed_env_has_pam_package.xml
@@ -31,7 +31,7 @@
     <linux:object object_ref="obj_env_has_pam_installed" />
   </linux:dpkginfo_test>
   <linux:dpkginfo_object id="obj_env_has_pam_installed" version="1">
-    <linux:name>pam</linux:name>
+    <linux:name>libpam-runtime</linux:name>
   </linux:dpkginfo_object>
 {{% endif %}}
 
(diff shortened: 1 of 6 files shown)


richardmaciel-canonical (approved):
LGTM

shared/checks/oval/installed_env_has_pam_package.xml
```diff
@@ -31,7 +31,7 @@
     <linux:object object_ref="obj_env_has_pam_installed" />
   </linux:dpkginfo_test>
   <linux:dpkginfo_object id="obj_env_has_pam_installed" version="1">
-    <linux:name>pam</linux:name>
+    <linux:name>libpam-runtime</linux:name>
```
yuumasato, line 34: I think you'll also need to map `pam` to `libpam-runtime` in a `package-override` for Ubuntu and Debian products.
Otherwise the remediations will still check for `pam`.
Something similar to:
https://github.com/ComplianceAsCode/content/blob/96cbe2466cc11cd9de20c6ed7061ae2a54e1ac60/sle15/product.yml#L27

cipherboy: Aha, ty :)

yuumasato:
@openscap-ci test this please

yuumasato:
These changes should not affect OCP4, and the error seems infrastructural (timeout).

PR #6985, opened 2021-05-10
OCP4: Introduce FIPS check for MachineConfigs

JAORMX:
This checks that FIPS is enabled in the relevant MachineConfig object.

This is not meant as a standalone check, but supporting the OS check.

Note that this needs: https://github.com/openshift/compliance-operator/pull/640


JAORMX:
This PR is required before merging this: https://github.com/openshift/release/pull/18778

JAORMX:
Pull-request updated, HEAD is now a00abc2d7e8d946d9a1f8716eb06532ee2ee24ce

Issue #6899, opened 2021-04-27
RHEL8 STIG with GUI installation - package conflicts 


mildas:
#### Description of problem:
I'm not able to install RHEL 8.4 with Server with GUI package selection and with the STIG security policy - "DISA STIG with GUI for Red Hat Enterprise Linux 8".

I have built rhel8 datastream and inserted it to the installation via network (Security Policy -> Change content)

#### SCAP Security Guide Version:
c43f689153c

#### Operating System Version:
RHEL 8.4

#### Steps to Reproduce:
1. Start rhel8.4 installation
2. Prepare partitions for STIG
3. Security Policy -> Change content, and put the link with rhel8 datastream to fetch 

#### Actual Results:
Not possible to begin the installation because of software selection error:
`Error checking software selection`

#### Expected Results:
Successful installation

#### Additional Information/Debugging Steps:
Most of the conflicts are about `xorg-x11-*` even though it is being unselected in the `stig_gui.profile`. However, `gssproxy` package conflict seems valid to me because the profile contains the `package_gssproxy_removed` rule.

Software selection details:
```
 Problem 1: package xorg-x11-xinit-1.3.4-18.el8.x86_64 requires xhost, but none of the providers can be installed
  - conflicting requests
  - package xorg-x11-server-utils-7.7-27.el8.x86_64 is filtered out by exclude filtering
 Problem 2: package gnome-session-wayland-session-3.28.1-10.el8.x86_64 requires xorg-x11-server-Xwayland(x86-64), but none of the providers can be installed
  - conflicting requests
  - package xorg-x11-server-Xwayland-1.20.10-1.el8.x86_64 is filtered out by exclude filtering
 Problem 3: package xorg-x11-drv-wacom-0.38.0-1.el8.x86_64 requires Xorg, but none of the providers can be installed
  - package xorg-x11-drv-wacom-0.38.0-1.el8.x86_64 requires xserver-abi(ansic-0) >= 4, but none of the providers can be installed
  - package xorg-x11-drv-wacom-0.38.0-1.el8.x86_64 requires xserver-abi(xinput-24) >= 1, but none of the providers can be installed
  - conflicting requests
  - package xorg-x11-server-Xorg-1.20.10-1.el8.x86_64 is filtered out by exclude filtering
 Problem 4: package xorg-x11-drv-vmware-13.2.1-8.el8.x86_64 requires Xorg, but none of the providers can be installed
[...]

PR #6993, opened 2021-05-11
Disable anaconda remediation from package_gssproxy_removed to prevent blocking installation

ggbecker:
#### Description:

-  Disable anaconda remediation from package_gssproxy_removed to prevent blocking installation

#### Rationale:

- It conflicts with nfs-utils which is included in GUI installations.

- Relates to: #6987

- With Anaconda remediation disabled, the bash remediation takes place during post install phase and any package dependency is nicely handled by `yum`.

- Fixes: #6899

### DIFF
--- a/linux_os/guide/system/software/system-tools/package_gssproxy_removed/rule.yml
+++ b/linux_os/guide/system/software/system-tools/package_gssproxy_removed/rule.yml
@@ -27,3 +27,5 @@ template:
     name: package_removed
     vars:
         pkgname: gssproxy
+    backends:
+        anaconda: "off"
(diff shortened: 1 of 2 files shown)


ggbecker:
The following test scenarios have been executed:

- Installed RHEL8 with GUI then remove package gssproxy. Everything seems to be ok but the packages that are removed because of dependencies are quite big, see: https://pastebin.com/rQrsdDG7

- Installed RHEL8 with GUI and then run oscap remediate (unselected `package_gssproxy_removed`)
  - Results are all good.

ggbecker:
Pull request is ready to review. Disabling the anaconda remediation of this specific rule ensured installation of RHEL with GUI finished successfully.

carlosmmatos:
@openscap-ci retest

PR #6998, opened 2021-05-11
OL7 DISA STIG v2r3 update

freddieRv:
#### Description:

- Bump version on stig profile
- Update stig manual reference xccdf file
- Update stig overlay
- Set var_accounts_tmout to 15_mins
- Add rules:
    - sudoers_validate_passwd
    - sudo_restrict_privilege_elevation_to_authorized
    - xwindows_remove_packages

#### Rationale:

- OL profiles effort


ol7/profiles/stig.profile
```diff
@@ -303,3 +303,6 @@ selections:
     - grub2_admin_username
     - grub2_uefi_admin_username
     - sshd_x11_use_localhost
+    - sudoers_validate_passwd
+    - sudo_restrict_privilege_elevation_to_authorized
+    - xwindows_remove_packages
```
ggbecker, line 308: you probably want to remove `package_xorg-x11-server-common_removed` from this profile and reassign its STIG id to this new `xwindows_remove_packages`. I haven't checked the OL STIG v2r3, but that was what happened with RHEL STIG, so I'm assuming something similar would occur with OL STIG

ggbecker: The same with `xwindows_runlevel_target`. `xwindows_remove_packages` should do everything needed from both these rules.

ggbecker: see this: https://github.com/ComplianceAsCode/content/pull/6860/files#diff-1e03dd31341183b6e46243526e1437e19ea87a73351bbd2b807ed763471641d3

freddieRv: You are right, `xwindows_remove_packages` covers what both these rules were covering. Thanks for the notice - addressed.

ggbecker (approved):
LGTM. Thanks

PR #7002, opened 2021-05-12
Add rule for SLES-15-010356

vtrubovics:
**Description:**

-  Add rule to 'Verify that system-wide directories Have Root Group Ownership'

**Rationale:**

System-wide directories should be group-owned by root.

- Add oval check for directories containing system library files having correct ownership
- Add ansible remediation based on combination of find and file modules in order to avoid lint warnings for file permissions when using find shell command and have no previous knowledge on the directories permission modes
- Add couple of tests to verify pass and fail behaviour
- Add rule for SLES-15-010356 to stig profile


ggbecker:
@openscap-ci ok to test

linux_os/guide/system/permissions/files/dir_syslibraries_root_owned_group/rule.yml
```diff
@@ -0,0 +1,54 @@
...
+    and "/usr/lib64" are group-owned by root.
+
+description: |-
+    System-wide library directories are:
+    <pre>/lib
+    /lib64
+    /usr/lib
+    /usr/lib64
+    </pre>
+    All system-wide directories should be protected from unautorised
```
ggbecker, line 16:
```suggestion
    All system-wide directories should be protected from unauthorised 
```

vtrubovics: Fixed.

ggbecker:
@openscap-ci add to whitelist

PR #7004, opened 2021-05-12
install_vm.py: add possibility to install GUI system

matusmarhefka:
This PR introduces `--install-gui` option into the `install_vm.py` script so it is easy to test also installations with GUI.

Additional changes:
* updated default Fedora compose in the `install_vm.py` script to Fedora 34
* updated SSGTS kickstart file as `install` command has been deprecated in Fedora 34 - https://docs.fedoraproject.org/my/fedora/f34/release-notes/sysadmin/Installation/#sect-kickstart-deprecated-removed
* updated SSGTS kickstart file to contain additional partitions (`/usr`, `/opt`, `/srv`) so the kickstart is compatible with all Fedora/RHEL profiles (`/usr` partition has increased size to be able to install Fedora 34 with GUI)

I have tested both installations, with and without GUI, for the following:
* RHEL7
* RHEL8
* Fedora 34

### DIFF
--- a/tests/install_vm.py
+++ b/tests/install_vm.py
@@ -94,6 +94,12 @@ def parse_args():
         choices=['secureboot', 'normal'],
         help="Perform UEFI based installation, optionally with secure boot support."
     )
+    parser.add_argument(
+        "--install-gui",
+        dest="install_gui",
+        action='store_true',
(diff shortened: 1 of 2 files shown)


jan-cerny:
Works amazing, I was able to install a F34 VM with GUI.

PR #7005, opened 2021-05-13
Improve test suite documentation

jan-cerny:
1. reorganize into a new structure. There are 3 main sections now:
- Preparing backends
- Test scenarios
- Running tests

2. add more information about the profiles key in test scenario
headers

3. explain more the combined and rule mode and usage of profiles in
   these modes

4. typo and grammar fixes

5. consistent usage of data stream

It's mostly reshuffling the text, there is only a little new text.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -9,19 +9,18 @@ Priority is to provide very simple system of test scenarios, easily extendable
 so everyone contributing remediation can also provide cases where this
 remediation works.
 
-The Test Suite executes the scans, remediations, and test scenarios on VMs or containers.\
-To test content using VMs, *Libvirt backend* is used;\
+The Test Suite executes the scans, remediations, and test scenarios on VMs or containers.
+To test content using VMs, *Libvirt backend* is used;
 to test content using containers, either *Podman* or *Docker* backend can be used.
 
 Once a domain or container is prepared it can be used indefinitely.\
 The Test Suite can perform tests focused on a profile or a specific rule.\
 Typically, for a test run, a sequence of scan, remediation and scan is peformed.
 
-# How to prepare a backend for testing
-For the Test Suite to work, you need to have a backend image prepared for testing.\
-You can use a powerful full-blown VM backend, or a lightweight container backend.
+# Preparing a backend for testing
 
-SSG Test Suite currently does not provide automated provisioning of VMs or containers.
+For the test suite to work, you need to have a backend image prepared for testing.
+You can use a powerful full-blown VM backend, or a lightweight container backend.
 
 ## Libvirt backend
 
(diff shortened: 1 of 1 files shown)


tests/README.md
```diff
@@ -130,16 +128,116 @@ To use Docker backend, you need to have:
...
+- `profiles` is a comma-separated list of profiles to which this scenario is
+  restricted. Use this only if the scenario makes sense only in a specific
+  profile. Typically, a rule doesn't depend on a profile and behaves the same
+  way regardless the profile it's a part of. If the rule is parametrized by
+  variables (XCCDF Values), use the `variables` key instead. This key is
+  intended to be used in regression testing of bugs in profiles, it isn't
+  intended for casual use.
+- `remediation` is a string specifying one of the allowed remediation types (eg.
+  `bash`, `ansible`, `none`). The `none` value means that the tested rule has no
+  implemented remediation.
```
matusmarhefka, line 175: We can add here that `none` can also be used in case that remediation breaks test environemnt (for example unmounting `/tmp` would break test suite as opnescap generates reports into this directory).

tests/README.md
```diff
@@ -299,111 +355,108 @@ Notice we didn't use full rule name on the command line. The prefix `xccdf_org.s
...
+Note that `profile-id` is matched by the suffix, so it works the same as in `oscap` tool
+(you can use `oscap info --profiles` to see available profiles).
 
-In this mode, you are testing the rules selected by a profile, using the contexts provided by each rule's test scenarios.
-This mode provides an easy way of performing rule-based testing on all rules that are part of a profile.
+### Unselecting problematic rules
 
-Any `# profiles` metadata in the test scenarios will be ignored, the Test Suite will assume that the test scenarios apply for the Profile, since the rule is selected by the profile.
+Sometimes you would like to skip a rule in the profile because they are too slow to test,\
+or you know a rule doesn't have a remediation and you get less value by testing it.
```
matusmarhefka, line 404: I would also add one more example case here - some rules might break the test suite, for example the rule `sshd_disable_root_login` which disables root login to the tested VM will break our test suite as it uses root user with all ssh commands (yes we should think about using some user in wheel group in the future so we don't need to unselect such rules).

tests/README.md
```diff
@@ -38,14 +37,13 @@ To use Libvirt backend, you need to have:
   - `virt-viewer`       (optional, to access graphical console of VMs)
   - `ansible`           (required if remediating via Ansible)
   - `edk2-ovmf`         (required if you want to install UEFI based machine)
-- A VM that fullfils following requirements:
+- A VM that fulfils the following requirements:
   - Package `qemu-guest-agent` installed
   - Package `openscap` version 1.2.15 or higher installed
   - `root` can login via ssh (it is recommended to setup key-based authentication)
   - `root` can install packages (for RHEL7, it means subscription enabled).
```
cipherboy, line 44: One thing not mentioned is that the VM needs to have `CPE_NAME` in `/etc/os-release`. I came across this because Ubuntu doesn't ship such in a stock image. It isn't a lot of work to add, but it is something to note.

https://bugs.launchpad.net/ubuntu/+source/base-files/+bug/1472288 

It didn't get picked up for Hirsute, so it'll likely be Impish (21.10) or later (22.04 LTS).

matusmarhefka:
Merging, thanks @jan-cerny for documenting all of the pitfalls of the SSGTS.

PR #7008, opened 2021-05-17
Remove audit_privileged_commands from RHEL7 STIG profile

ggbecker:
#### Description:

- Remove audit_privileged_commands from RHEL7 STIG profile.
  - This rule is not aligned with latest version of RHEL7 STIG and should be
removed from the profile.

#### Rationale:

- The previous assigned rule was how previous versions of RHEL7 STIG prior V2 handled the vulnerability. Since V2 the rule has been using audit to check the execve syscall
- Current text RHEL7: https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-204516


JAORMX (approved):

vojtapolasek:
LGTM

PR #7011, opened 2021-05-17
Enable dconf rules for RHEL9

matejak:
There are no changes besides the deprecation of the `gdm.d` directory, which is also part of this PR.

`gdm.d` has been removed as a solution of https://bugzilla.redhat.com/show_bug.cgi?id=1546644 and it is patched by e.g.
https://src.fedoraproject.org/rpms/gdm/blob/f34/f/0001-data-add-system-dconf-databases-to-gdm-profile.patch -
you dont see `system-db:gdm` in the `data/dconf/gdm.in` file (the whole file is shown in the diff).


shared/macros.jinja
```diff
@@ -3,6 +3,12 @@ ocil_clause: "the required value is not set"
 {{%- endmacro %}}
 
 
+{{% set dconf_gdm_dir = "distro.d" %}}
+{{% if product not in ("fedora", "rhel9") %}}
+{{% set dconf_gdm_dir = "gdm.d" %}}
```
jan-cerny, line 8: I have a question: Shouldn't it be a property of product.yml? What's our guidance?

matejak: That would probably be a better place, yes. Taking a closer look, it looks like that the issue is present only in RHEL7 and RHEL8, so a default value combined with override should work better.

yuumasato: Something similar was done recently for grub2's path:
https://github.com/ComplianceAsCode/content/pull/6986

matejak: > Something similar was done recently for grub2's path

Thanks, it is nice to have a provisional solution that is already used.

linux_os/guide/system/accounts/accounts-banners/gui_login_banner/dconf_gnome_banner_enabled/tests/correct_value.pass.sh
```diff
@@ -5,7 +5,7 @@ source $SHARED/dconf_test_functions.sh
 install_dconf_and_gdm_if_needed
 
 clean_dconf_settings
-add_dconf_setting "org/gnome/login-screen" "banner-message-enable" "true" "gdm.d" "00-security-settings"
-add_dconf_lock "org/gnome/login-screen" "banner-message-enable" "gdm.d" "00-security-settings"
+add_dconf_setting "org/gnome/login-screen" "banner-message-enable" "true" "distro.d" "00-security-settings"
+add_dconf_lock "org/gnome/login-screen" "banner-message-enable" "distro.d" "00-security-settings"
```
yuumasato, line 9: The changes in this test scenario make it only applicable for Fedora and RHEL9.

Is there a way to conditionally setup different gdm dconf directories?
Otherwise, this test should probably be duplicated, one copy for products using `gdm.d`, and another copy for products using `distro.d`.

linux_os/guide/system/accounts/accounts-banners/gui_login_banner/dconf_gnome_login_banner_text/tests/correct_value.pass.sh
```diff
@@ -10,7 +10,7 @@ login_banner_text="--[\s\n]+WARNING[\s\n]+--[\s\n]*This[\s\n]+system[\s\n]+is[\s
 expanded=$(echo "$login_banner_text" | sed 's/(\\\\\x27)\*/\\\x27/g;s/(\\\x27)\*//g;s/(\\\\\x27)/tamere/g;s/(\^\(.*\)\$|.*$/\1/g;s/\[\\s\\n\][+*]/ /g;s/\\//g;s/(n)\*/\\n/g;s/\x27/\\\x27/g;')
 
 clean_dconf_settings
-add_dconf_setting "org/gnome/login-screen" "banner-message-text" "'${expanded}'" "gdm.d" "00-security-settings"
-add_dconf_lock "org/gnome/login-screen" "banner-message-text" "gdm.d" "00-security-settings-lock"
+add_dconf_setting "org/gnome/login-screen" "banner-message-text" "'${expanded}'" "distro.d" "00-security-settings"
+add_dconf_lock "org/gnome/login-screen" "banner-message-text" "distro.d" "00-security-settings-lock"
```
yuumasato, line 14: This test scenario is only applicable for RHEL7, so it should configure `gdm.d`, not `distro.d`

linux_os/guide/system/accounts/accounts-banners/gui_login_banner/dconf_gnome_login_banner_text/tests/missing_value_stig.fail.sh
```diff
@@ -10,7 +10,7 @@ install_dconf_and_gdm_if_needed
 # expanded=$(echo "$login_banner_text" | sed 's/(\\\\\x27)\*/\\\x27/g;s/(\\\x27)\*//g;s/(\\\\\x27)/tamere/g;s/(\^\(.*\)\$|.*$/\1/g;s/\[\\s\\n\][+*]/ /g;s/\\//g;s/(n)\*/\\n/g;s/\x27/\\\x27/g;')
 
 clean_dconf_settings
-add_dconf_setting "org/gnome/login-screen" "banner-message-enabled" "true" "gdm.d" "00-security-settings"
-add_dconf_lock "org/gnome/login-screen" "banner-message-enabled" "gdm.d" "00-security-settings-lock"
+add_dconf_setting "org/gnome/login-screen" "banner-message-enabled" "true" "distro.d" "00-security-settings"
+add_dconf_lock "org/gnome/login-screen" "banner-message-enabled" "distro.d" "00-security-settings-lock"
```
yuumasato, line 14: Same comments here about applicability, either this test is moved to RHEL9 platform, or the test remains setting up `gdm.d`.

yuumasato (commented):
Changes look good to me, but the test scenarios need a bit more of work.
The tests applicabilities are mismatched with what they configure.

linux_os/guide/system/accounts/accounts-banners/gui_login_banner/dconf_gnome_banner_enabled/tests/correct_value.pass.sh
cipherboy: Did I suggest on one of my test system changes issues/PRs to use jinja for interpreting test cases prior to running them? It'd require reading product as well... But I think it might help to write platform-independent tests a little better (in addition to the test suite header)? Just a thought :)

cipherboy (commented):
Just commenting here to say I like the structure of the PR, thanks @matejak! :)

(I obviously haven't tested this on RHEL 9).

matejak:
As #7210 works really great, test scenarios now support both old and new RHELs.

yuumasato (approved):
LGTM

PR #7017, opened 2021-05-18
Fix configure_opensc_nss_db bash remediation

freddieRv:
#### Description:

- Fix remediation script to compare the output value of `pkcs11-switch` and to run `pkcs11-switch opensc` inside of `if` statement
- Added call to `modutil` since `pkcs11-switch opensc` does not update `/etc/pki/nssdb/pkcs11.txt`. Without this update rule results in `fail` even after successfully running the remediation

#### Rationale:

- This is a bug fix

### DIFF
--- a/linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/configure_opensc_nss_db/bash/shared.sh
+++ b/linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/configure_opensc_nss_db/bash/shared.sh
@@ -4,8 +4,11 @@
 # complexity = low
 # disruption = low
 
-PKCSSW=$(/usr/bin/pkcs11-switch)
+PKCSSW="/usr/bin/pkcs11-switch"
+MODULE=$(${PKCSSW})
 
-if [ ${PKCSSW} != "opensc" ] ; then
-    ${PKCSSW} opensc
+if [ "$MODULE" != "opensc" ] ; then
+    echo | ${PKCSSW} opensc
 fi
+
+modutil -force -add opensc -dbdir sql:/etc/pki/nssdb -libfile opensc-pkcs11.so


ggbecker:
I've tried to do something similar in the past (https://github.com/ComplianceAsCode/content/pull/4776) regarding this rule, but faced some technical aspects of this module that the way of configuring it was not really something that worked. I don't remember exactly the details but I can try to find conversations with maintainer to illustrate better.

Have you tried to do a functional testing with this using smart cards?

freddieRv:
@ggbecker Thanks for the feedback and the extra information. I was not aware of #4776.

> Have you tried to do a functional testing with this using smart cards?

I have not. My goal with this PR was to bring the bash remediation to what it intended to do. I am also not quite familiar with this module. Only after some digging I found the `modutil` tool and added the statement using it in the hopes of passing the oval check.

However after reading the info you provided:

> `... You can examine pkcs11.txt, but there's a pretty high chance that it will not be what the application is using to load the module.`

I believe the oval check could use some updating too.

linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/configure_opensc_nss_db/bash/shared.sh
```diff
@@ -4,8 +4,11 @@
 # complexity = low
 # disruption = low
 
-PKCSSW=$(/usr/bin/pkcs11-switch)
+PKCSSW="/usr/bin/pkcs11-switch"
```
mildas, line 7: This script is in the `opensc` package so if the package is not installed the remediation will do result in `error`.

@ggbecker do you think that installing the `opensc` package in this remediation is a good idea? Installation of `opensc` feels like a different rule, however without the package we can't perform remediation of this one.

jan-cerny: @mildas Probably yes, Imagine that somebody will update the rule in the future and this test will be executed by the CTF.

jan-cerny:
@mildas @ggbecker Can you summarize what we need to do to unblock this?

ggbecker:
> @mildas @ggbecker Can you summarize what we need to do to unblock this?

In theory, we have no confirmation that this rule is useful. Because the OVAL check looks for the `pkcs11.txt` file but that doesn't mean that the smartcard's module is actually considering contents of that file. This change is probably good because it updates the `pkcs11.txt` accordingly though, so from the OVAL/remediation alignment perspective it should be correct.

jan-cerny:
@mildas @ggbecker So would you be in favor of closing this PR? It's a stalled PR.

ggbecker:
> @mildas @ggbecker So would you be in favor of closing this PR? It's a stalled PR.

I would say so. The rule itself is very controversial and I couldn't find the proper solution back in the days. Although the check/remediation could be properly aligned, there is no evidence that the configuration actually works.

PR #7023, opened 2021-05-19
Fix issue 7019

### LINKED ISSUE #7019: SLE15 - Wrong STIG id assigned to rule
https://github.com/ComplianceAsCode/content/blob/c1cc3f23363105a886227bca1b905b71d795dcc0/linux_os/guide/system/software/sudo/sudo_require_authentication/rule.yml#L36

@brett060102 I believe this is not the correct STIG for this rule. At least from what the rule is looking for (!authenticate NOPASSWD)

From SLE15 STIG text there is:
`
Verify the operating system requires re-authentication when using the "sudo" command to elevate privileges. > sudo grep -i 'timestamp_timeout' /etc/sudoers /etc/sudoers.d/* /etc/sudoers:Defaults timestamp_timout=0 If "timestamp_timeout" is set to a negative number, is commented out, or no results are returned, this is a finding.
`

RHEL STIGs also have this rule but so far there is no implemented rule in the project.

brett060102:
SUSE STIG should have been SLES-15-010450 not SLES-15-020102

#### Description:

- We should have used STIG SLES-15-010450 for this rule

#### Rationale:

- Need to have correct STIG

- Fixes #7019


ggbecker (approved):
LGTM

PR #7025, opened 2021-05-20
Increase build system flexibility

jan-cerny:
1. Add options for building Ansible and Bash content
This patch adds 2 new options SSG_ANSIBLE_PLAYBOOKS_ENABLED and
SSG_BASH_SCRIPTS_ENABLED which will allow user to turn on or off
building and installing profile Bash remediation scripts and profile
Ansible Playbooks. They are enabled by default, therefore the default
behavior doesn't change, but people can turn them off to speed up the
build. These options can be useful when calling cmake in downstream spec
files.

2.  Do not build HTML guide for the virtual default profile
The virtual '(default)' profile is a profile that doesn't contain
any rules, so the built HTML guide also doesn't contain any rules
which means it contains only group descriptions. This HTML guide
has no use for the users and it only increases the built size.

3. Add option to disable SCAP 1.2 data streams
This commit adds a new option that enables to turn on building
the SCAP 1.2 source data streams (ssg-*-ds-1.2.xml). This option
will help people who don't want to build and ship this file.
The default setting is TRUE which means the default behavior
shouldn't change.


carlosmmatos:
@jan-cerny any plan to add a wrapper in `build_product` script for disabling/enabling of ansible/bash? Or will this just be passing it in as an ADDITIONAL_CMAKE_OPTIONS?

jan-cerny:
@openscap-ci test this please

carlosmmatos:
LGTM

matejak:
> @jan-cerny any plan to add a wrapper in `build_product` script for disabling/enabling of ansible/bash? Or will this just be passing it in as an ADDITIONAL_CMAKE_OPTIONS?

This has been addressed, and the PR plays nicely with the script's `--datastream-only` option, so LGTM as well.

matejak:
The CI failure seems to be related to infrastructure issues, so I merge this PR.

PR #7027, opened 2021-05-20
RHEL-08-020100 updating existing rule

### LINKED ISSUE #6954: RHEL-08-020100 remediation does not write "retry=3" to password-auth
#### Description of problem:  

RHEL-08-020100 remediation does not write "retry=3" to password-auth.  It does add it to system-auth correctly.

0.1.55

RHEL 8.3

#### Steps to Reproduce:

1.  Apply ansible playbook to fresh rhel8 system.
2.  Apply oscap xccdf eval --remediate to same system
3.   sudo cat /etc/pam.d/password-auth | grep pam_pwquality does not show deny=3
4.  

#### Actual Results:

deny=3 added to system-auth

#### Expected Results:

deny=3 added to system-auth AND password-auth

N/A

carlosmmatos:
#### Description:

- Retry=3 pam_pwquality module does not check/remediate the password-auth file in RHEL 8.

#### Rationale:

- The RHEL 8 STIG currently expects the retry=3 option to be in both system-auth and password-auth files

- Fixes #6954 

- ***NOTES*** I will be requesting a change for this rule with DISA to follow the same format for all the other pwquality rules. That is, to use the `/etc/security/pwquality.conf` file to set this rule.

### DIFF
--- a/linux_os/guide/system/accounts/accounts-pam/password_quality/password_quality_pwquality/accounts_password_pam_retry/oval/shared.xml
+++ b/linux_os/guide/system/accounts/accounts-pam/password_quality/password_quality_pwquality/accounts_password_pam_retry/oval/shared.xml
@@ -2,20 +2,37 @@
   <definition class="compliance" id="accounts_password_pam_retry" version="1">
     {{{ oval_metadata("The password retry should meet minimum requirements") }}}
     <criteria operator="AND" comment="Conditions for retry are satisfied">
-      <criterion comment="pam_pwquality" test_ref="test_password_pam_pwquality_retry" />
+      <criterion comment="pam_pwquality system-auth" test_ref="test_password_pam_pwquality_retry_system_auth" />
+      {{% if product == "rhel8" -%}}
+      <criterion comment="pam_pwquality password-auth" test_ref="test_password_pam_pwquality_retry_password_auth" />
+      {{%- endif %}}
     </criteria>
   </definition>
 
-  <ind:textfilecontent54_test check="all" comment="check the configuration of /etc/pam.d/system-auth" id="test_password_pam_pwquality_retry" version="1">
-    <ind:object object_ref="obj_password_pam_pwquality_retry" />
+  {{% macro test_pwquality_retry(path, test_ref) %}}
+  <ind:textfilecontent54_test check="all" comment="check the configuration of {{{ path }}}" id="test_{{{ test_ref }}}" version="1">
+    <ind:object object_ref="obj_{{{ test_ref }}}" />
     <ind:state state_ref="state_password_pam_retry" />
   </ind:textfilecontent54_test>
+  {{% endmacro %}}
 
-  <ind:textfilecontent54_object id="obj_password_pam_pwquality_retry" version="1">
-    <ind:filepath>/etc/pam.d/system-auth</ind:filepath>
+  {{{ test_pwquality_retry( path="/etc/pam.d/system-auth", test_ref="password_pam_pwquality_retry_system_auth") }}}
+  {{% if product == "rhel8" -%}}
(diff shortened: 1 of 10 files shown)


JAORMX:
/retest

@jhrozek do you think this is relevant for RHCOS?

linux_os/guide/system/accounts/accounts-pam/password_quality/password_quality_pwquality/accounts_password_pam_retry/oval/shared.xml
```diff
@@ -2,20 +2,37 @@
...
+  <ind:textfilecontent54_object id="obj_{{{ test_ref }}}" version="1">
+    <ind:filepath>{{{ path }}}</ind:filepath>
     <ind:pattern operation="pattern match">^\s*password\s+(?:(?:required)|(?:requisite))\s+pam_pwquality\.so.*retry=([0-9]*).*$</ind:pattern>
     <ind:instance datatype="int">1</ind:instance>
   </ind:textfilecontent54_object>
+  {{% endmacro %}}
+
+  {{{ object_pwquality_retry( path="/etc/pam.d/system-auth", test_ref="password_pam_pwquality_retry_system_auth") }}}
+  {{% if product == "rhel8" -%}}
+  {{{ object_pwquality_retry( path="/etc/pam.d/system-auth", test_ref="password_pam_pwquality_retry_password_auth") }}}
```
ggbecker, line 34:
```suggestion
  {{{ object_pwquality_retry( path="/etc/pam.d/password-auth", test_ref="password_pam_pwquality_retry_password_auth") }}}
```

linux_os/guide/system/accounts/accounts-pam/password_quality/password_quality_pwquality/accounts_password_pam_retry/oval/shared.xml
```diff
@@ -2,20 +2,37 @@
...
+    <ind:object object_ref="obj_{{{ test_ref }}}" />
     <ind:state state_ref="state_password_pam_retry" />
   </ind:textfilecontent54_test>
+  {{% endmacro %}}
 
-  <ind:textfilecontent54_object id="obj_password_pam_pwquality_retry" version="1">
-    <ind:filepath>/etc/pam.d/system-auth</ind:filepath>
+  {{{ test_pwquality_retry( path="/etc/pam.d/system-auth", test_ref="password_pam_pwquality_retry_system_auth") }}}
+  {{% if product == "rhel8" -%}}
+  {{{ test_pwquality_retry( path="/etc/pam.d/system-auth", test_ref="password_pam_pwquality_retry_password_auth") }}}
```
ggbecker, line 21:
```suggestion
  {{{ test_pwquality_retry( path="/etc/pam.d/password-auth", test_ref="password_pam_pwquality_retry_password_auth") }}}
```

carlosmmatos:
@ggbecker good catch.. I've added the changes, and cleaned up the commit history

carlosmmatos:
@ggbecker ci job failed due to timeout. It's up to you whether you want to retest it or not.

ggbecker:
I actually have two concerns about this pull request. First one is that we are considering `0` as a passing value and STIG explicitly states that value zero is a finding.

So we should update the OVAL accordingly... sorry it was not in the original scope of this PR, but I only realized that now.

The second one is that the test scenarios are not considering the `password-auth` file, even though the test passes because the default settings contain the `retry=3`, but we should ensure our test scenarios cover this. Btw, you can probably remove the profile metadata `ospp` profile from the test scenarios.

ggbecker (approved):
LGTM now. Thanks. I don't think we need to wait for the last ocp test. They are not related to changes in this PR.

JAORMX (approved):

PR #7029, opened 2021-05-21
Update NIST references for audit log forwarding and IDPs

jhrozek:
Many controls can be solved by offloading logs and having an IDP


JAORMX (approved):
LGTM

PR #7030, opened 2021-05-21
Fix disable_users_coredumps's limits.d exists

cipherboy:
When running profiles with this rule, the following message is
outputted:

> ```
> The 'oscap' process has written the following content to stderr:
>   W: oscap: Entity name 'subexpression' from state
>   (id: 'oval:ssg-state_core_dumps_limits_d_exists:ste:1') not found
>   in item (id: '145163290').
> ```

### DIFF
--- a/linux_os/guide/system/permissions/restrictions/coredumps/disable_users_coredumps/oval/shared.xml
+++ b/linux_os/guide/system/permissions/restrictions/coredumps/disable_users_coredumps/oval/shared.xml
@@ -45,11 +45,7 @@
   comment="Tests for existance of the ^[\s]*\*[\s]+(hard|-)[\s]+core setting in the /etc/security/limits.d directory"
   id="test_core_dumps_limits_d_exists" version="1">
     <ind:object object_ref="object_core_dumps_limits_d_exists" />
-    <ind:state state_ref="state_core_dumps_limits_d_exists" />
   </ind:textfilecontent54_test>
-  <ind:textfilecontent54_state id="state_core_dumps_limits_d_exists" version="1">
-    <ind:subexpression operation="equals">0</ind:subexpression>
-  </ind:textfilecontent54_state>
   <ind:textfilecontent54_object id="object_core_dumps_limits_d_exists" version="1">
     <ind:path>/etc/security/limits.d</ind:path>
     <ind:filename operation="pattern match">^.*\.conf$</ind:filename>


richardmaciel-canonical (changes requested):
Since you're removing the state, you should change the object regexp to match the value of the core parameter. Something like this:
^[\s]*\*[\s]+(?:hard|-)[\s]+core[\s]+0

Else, you risk validating a line like this:
hard core 100
which actually allows core dumps of, at most, 100kb.

cipherboy:
@richardmaciel-canonical said:
> Since you're removing the state, you should change the object regexp to match the value of the core parameter. Something like this:
> ```
> ^[\s]**[\s]+(?:hard|-)[\s]+core[\s]+0
> ```

I don't think this is necessary? Note that the `test_core_dumps_limits_d` test (before this change) entirely duplicated the `test_core_dumps_limits_d_exists` test and contains the missing capture group. :-) 

So in this case, the `_exists` elements that we're modifying here (to remove the state) are just checking whether a line with `hard core` exists in the file, and the non-`_exists` suffixed items are actually validating the value (above). 

While the updated `_exists` elements _will_ match on `100` value, I _think_ that's by design... and the entire OVAL check will fail if the line is 100.

I could be missing something, please tell me if I am :-)

richardmaciel-canonical:
Ah, I haven't noticed the test_core_dumps_limits_d test. 
I agree with you, then.

richardmaciel-canonical (approved):

cipherboy:
Looking at the last merged PR: https://github.com/ComplianceAsCode/content/pull/7037#partial-pull-merging

>  INFO[2021-05-24T14:18:57Z]     helpers.go:695: Result - Name: e2e-moderate-master-disable-users-coredumps - Status: FAIL - Severity: medium 

So its failure here isn't related AFAICT.

JAORMX:
> Looking at the last merged PR: [#7037 (comment)](https://github.com/ComplianceAsCode/content/pull/7037#partial-pull-merging)
> 
> > INFO[2021-05-24T14:18:57Z]     helpers.go:695: Result - Name: e2e-moderate-master-disable-users-coredumps - Status: FAIL - Severity: medium
> 
> So its failure here isn't related AFAICT.

The ocp test test for expected values for the rules, the rules can have a PASS or FAIL, and some might have remediations, so they should PASS after the remediation has been applied.

Actual test failures get reflected with the E2E-FAILURE identifier.

ggbecker:
> @richardmaciel-canonical said:
> 
> > Since you're removing the state, you should change the object regexp to match the value of the core parameter. Something like this:
> > ```
> > ^[\s]**[\s]+(?:hard|-)[\s]+core[\s]+0
> > ```
> 
> I don't think this is necessary? Note that the `test_core_dumps_limits_d` test (before this change) entirely duplicated the `test_core_dumps_limits_d_exists` test and contains the missing capture group. :-)
> 
> So in this case, the `_exists` elements that we're modifying here (to remove the state) are just checking whether a line with `hard core` exists in the file, and the non-`_exists` suffixed items are actually validating the value (above).
> 
> While the updated `_exists` elements _will_ match on `100` value, I _think_ that's by design... and the entire OVAL check will fail if the line is 100.
> 
> I could be missing something, please tell me if I am :-)

That's correct. It looks like it was a copy paste issue (see: https://github.com/ComplianceAsCode/content/commit/7f20eb7a6ecbdc8cd6d19edc7291f2dd62cd82a4)

ggbecker (approved):

PR #7032, opened 2021-05-21
Add bash package installated macro

cipherboy:
#### Description

This introduces a new macro, `bash_package_installed` that checks (in a platform-dependent way) whether a given package is installed. This macro is then used in `service_chronyd_or_ntpd_enabled`, to ensure the detection happens correctly on Debian/Ubuntu platforms. 

--- 

This macro could also be used in bash fixes for package inventory CPEs, whereby if a package isn't installed, the fix wouldn't be run (when running from a straight-profile hardening script instead of from oscap). This would be instead of the current Python-based code presently used. 

Just a thought for the future, outside the scope of this fix... :-)


richardmaciel-canonical (approved):

carlosmmatos:
LGTM

JAORMX (approved):

PR #7037, opened 2021-05-24
OCP4: Link ocp_idp_no_htpasswd with IA-5(4)

jhrozek:
The control is mostly an IDP-side control with the exception that htpasswd
must not be used so that the password policy can be set and controlled
centrally.

### DIFF
--- a/applications/openshift/master/file_permissions_openshift_pki_key_files/rule.yml
+++ b/applications/openshift/master/file_permissions_openshift_pki_key_files/rule.yml
@@ -20,6 +20,7 @@ severity: medium
 
 references:
   cis@ocp4: 1.1.21
+  nist: IA-5(2)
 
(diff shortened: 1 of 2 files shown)


applications/openshift/master/file_permissions_openshift_pki_key_files/rule.yml
```diff
@@ -14,6 +14,7 @@ rationale: |-
 
 identifiers:
   cce@ocp4: CCE-83580-1
+  nist: IA-5(2)
```
JAORMX, line 17: This is on the wrong section. It should be in `references` not `identifiers`

JAORMX (changes requested):

JAORMX (approved):
LGTM

PR #7039, opened 2021-05-24
Add option to enable installation of individual ansible tasks per rule.

ggbecker:
#### Description:

- Add option to enable installation of individual ansible tasks per rule.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,6 +49,7 @@ option(SSG_SVG_IN_XCCDF_ENABLED "If enabled, the built XCCDFs will include the S
 option(SSG_SEPARATE_SCAP_FILES_ENABLED "If enabled, separate SCAP files (OVAL, XCCDF, CPE dict, ...) will be installed alongside the source data-streams" TRUE)
 option(SSG_ANSIBLE_PLAYBOOKS_ENABLED "If enabled, Ansible Playbooks for each profile will be built and installed." TRUE)
 option(SSG_BASH_SCRIPTS_ENABLED "If enabled, Bash remediation scripts for each profile will be built and installed." TRUE)
+option(SSG_ANSIBLE_TASKS_ENABLED "If enabled, Ansible Tasks for each rule will be installed." FALSE)
 option(SSG_JINJA2_CACHE_ENABLED "If enabled, the jinja2 templating files will be cached into bytecode. Also see SSG_JINJA2_CACHE_DIR." TRUE)
 option(SSG_BATS_TESTS_ENABLED "If enabled, bats will be used to run unit-tests of bash remediations." TRUE)
 set(SSG_JINJA2_CACHE_DIR "${CMAKE_BINARY_DIR}/jinja2_cache" CACHE PATH "Where the jinja2 cached bytecode should be stored. This speeds up builds at the expense of disk space. You can use one location for multiple SSG builds for performance improvements.")
--- a/cmake/SSGCommon.cmake
+++ b/cmake/SSGCommon.cmake
@@ -914,6 +914,20 @@ macro(ssg_build_product PRODUCT)
             "
         )
     endif()
+    if(SSG_ANSIBLE_TASKS_ENABLED)
+        install(
+            CODE "
+            file(GLOB ROLE_FILES \"${CMAKE_BINARY_DIR}/${PRODUCT}/playbooks/*.yml\") \n
+            if(NOT IS_ABSOLUTE ${SSG_ANSIBLE_ROLE_INSTALL_DIR}/tasks)
+                file(INSTALL DESTINATION \"\${CMAKE_INSTALL_PREFIX}/${SSG_ANSIBLE_ROLE_INSTALL_DIR}/tasks\"
+                    TYPE FILE FILES \${ROLE_FILES})
+            else()
+                file(INSTALL DESTINATION \"${SSG_ANSIBLE_ROLE_INSTALL_DIR}/tasks\"
+                    TYPE FILE FILES \${ROLE_FILES})
+            endif()
+            "
+        )
+    endif()
 
     # grab all the kickstarts (if any) and install them
     file(GLOB KICKSTART_FILES "${CMAKE_CURRENT_SOURCE_DIR}/kickstart/ssg-${PRODUCT}-*-ks.cfg")


CMakeLists.txt
```diff
@@ -49,6 +49,7 @@ option(SSG_SVG_IN_XCCDF_ENABLED "If enabled, the built XCCDFs will include the S
 option(SSG_SEPARATE_SCAP_FILES_ENABLED "If enabled, separate SCAP files (OVAL, XCCDF, CPE dict, ...) will be installed alongside the source data-streams" TRUE)
 option(SSG_ANSIBLE_PLAYBOOKS_ENABLED "If enabled, Ansible Playbooks for each profile will be built and installed." TRUE)
 option(SSG_BASH_SCRIPTS_ENABLED "If enabled, Bash remediation scripts for each profile will be built and installed." TRUE)
+option(SSG_ANSIBLE_TASKS_ENABLED "If enabled, Ansible Tasks for each rule will be installed." FALSE)
```
jan-cerny, line 52: What about guarding the call of the ssg_build_ansible_playbooks macro by this option?

jan-cerny, line 52: In fact, these files are Ansible Playbooks, so using the term "tasks" is misleading.

cmake/SSGCommon.cmake
```diff
@@ -914,6 +914,20 @@ macro(ssg_build_product PRODUCT)
             "
         )
     endif()
+    if(SSG_ANSIBLE_TASKS_ENABLED)
+        install(
+            CODE "
+            file(GLOB ROLE_FILES \"${CMAKE_BINARY_DIR}/${PRODUCT}/playbooks/*.yml\") \n
```
jan-cerny, line 920: In fact, these files are Ansible Playbooks and you also have used the misleading word "tasks" for it before so I think using the term "role" for them increases confusion.

CMakeLists.txt
ggbecker: it makes sense

ggbecker: I understand that... but how should I call them since `SSG_ANSIBLE_PLAYBOOKS_ENABLED` is already being used?

maybe `SSG_ANSIBLE_INDIVIDUAL_PLAYBOOKS_ENABLED` or `SSG_ANSIBLE_PLAYBOOKS_PER_RULE_ENABLED`?

cmake/SSGCommon.cmake
ggbecker: that's correct... it was a copy paste problem

CMakeLists.txt
ggbecker: can you please check if this is what you have in mind? https://github.com/ComplianceAsCode/content/pull/7039/commits/1ec2353dc17c2c97e2c1a4ee66960f8723372afb

ggbecker: I've changed to SSG_ANSIBLE_PLAYBOOKS_PER_RULE_ENABLED, see: https://github.com/ComplianceAsCode/content/pull/7039/commits/81f9051433bec735f0ce915290d465ba98401f86#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR51

cmake/SSGCommon.cmake
```diff
@@ -914,6 +914,20 @@ macro(ssg_build_product PRODUCT)
             "
         )
     endif()
+    if(SSG_ANSIBLE_PLAYBOOKS_PER_RULE_ENABLED)
+        install(
+            CODE "
+            file(GLOB PLAYBOOK_PER_RULE_FILES \"${CMAKE_BINARY_DIR}/${PRODUCT}/playbooks/*\") \n
+            if(NOT IS_ABSOLUTE ${SSG_ANSIBLE_ROLE_INSTALL_DIR}/playbooks)
+                file(INSTALL DESTINATION \"\${CMAKE_INSTALL_PREFIX}/${SSG_ANSIBLE_ROLE_INSTALL_DIR}/playbooks/${PRODUCT}\"
```
jan-cerny, line 922: I'm sorry, I had a comment but I apparently didn't submit it or I misclicked. My comment was that the ${SSG_ANSIBLE_ROLE_INSTALL_DIR} is already used to install profile playbooks. But the downstream Linux distributions will want to place the rule playbooks to a different subpackage than the profile playbooks. Therefore I think it might be better to put the rule playbooks to a different directory than ${SSG_ANSIBLE_ROLE_INSTALL_DIR}. What is your opinion on that?

cmake/SSGCommon.cmake
```diff
@@ -769,7 +769,7 @@ macro(ssg_build_product PRODUCT)
     ssg_build_xccdf_unlinked(${PRODUCT})
     ssg_build_ocil_unlinked(${PRODUCT})
     ssg_build_remediations(${PRODUCT})
-    if ("${PRODUCT_ANSIBLE_REMEDIATION_ENABLED}")
+    if ("${PRODUCT_ANSIBLE_REMEDIATION_ENABLED}" OR SSG_ANSIBLE_PLAYBOOKS_PER_RULE_ENABLED)
```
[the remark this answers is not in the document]
ggbecker: Ok, will do.

cmake/SSGCommon.cmake
ggbecker: I understand your concern, but it would feel really strange to have ansible playbooks on a different folder than `{SSG_ANSIBLE_ROLE_INSTALL_DIR}` (which is usually `scap-security-guide/ansible`). That's why a created a subfolder called `playbooks` where this files would be installed. I guess the fact that downstream Linux distributions will use these files on a different subpackage doesn't necessarily mean we have to install these files on a different top-level directory.

cmake/SSGCommon.cmake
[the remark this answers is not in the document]
ggbecker: Done

cmake/SSGCommon.cmake
jan-cerny: >  but it would feel really strange to have ansible playbooks on a different folder than {SSG_ANSIBLE_ROLE_INSTALL_DIR}

It depends, something like /usr/share/scap-security-guide/rule_ansible_playbooks/ could work.

>  I guess the fact that downstream Linux distributions will use these files on a different subpackage doesn't necessarily mean we have to install these files on a different top-level directory.

That's correct, it's certainly possible to pick the right items for the (sub)packages in spec file.

I mean I think it's fine in the current form, I was curious if we could come up with something possibly better or not.

yuumasato (changes requested):
Just running `cmake` I get the following warning for each product:
```
CMake Warning (dev) at cmake/SSGCommon.cmake:804 (add_dependencies):          
  Policy CMP0046 is not set: Error on non-existent dependency in                                                                                             
  add_dependencies.  Run "cmake --help-policy CMP0046" for policy details.    
  Use the cmake_policy command to set the policy and suppress this warning.

  The dependency target "generate-chromium-ansible-playbooks" of target
  "chromium-content" does not exist.
Call Stack (most recent call first):
  chromium/CMakeLists.txt:6 (ssg_build_product)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

The problem is that `SSG_ANSIBLE_PLAYBOOKS_PER_RULE_ENABLED` being false by default, disables `ssg_build_ansible_playbooks(${PRODUCT})` which adds the custom_target `generate-${PRODUCT}-ansible-playbooks`, which is later consumed here: https://github.com/ComplianceAsCode/content/blob/a70543850e8411b267a6f0bcd3077f684b1a6f5e/cmake/SSGCommon.cmake#L806

ggbecker:
> Just running `cmake` I get the following warning for each product:
> 
> ```
> CMake Warning (dev) at cmake/SSGCommon.cmake:804 (add_dependencies):          
>   Policy CMP0046 is not set: Error on non-existent dependency in                                                                                             
>   add_dependencies.  Run "cmake --help-policy CMP0046" for policy details.    
>   Use the cmake_policy command to set the policy and suppress this warning.
> 
>   The dependency target "generate-chromium-ansible-playbooks" of target
>   "chromium-content" does not exist.
> Call Stack (most recent call first):
>   chromium/CMakeLists.txt:6 (ssg_build_product)
> This warning is for project developers.  Use -Wno-dev to suppress it.
> ```
> 
> The problem is that `SSG_ANSIBLE_PLAYBOOKS_PER_RULE_ENABLED` being false by default, disables `ssg_build_ansible_playbooks(${PRODUCT})` which would add the custom_target `generate-${PRODUCT}-ansible-playbooks`.

That's why I had initially an `OR` logic expression in the condition for `ssg_build_ansible_playbooks` call... I guess I'll have to reevaluate some things. Thanks for noticing that.

yuumasato:
> That's why I had initially an `OR` logic expression in the condition for `ssg_build_ansible_playbooks` call... I guess I'll have to reevaluate some things. Thanks for noticing that.

I think the `AND` makes sense. A solution could be to declare the custom target directly in the `ssg_build_product` macro, like in:
https://github.com/ComplianceAsCode/content/blob/a70543850e8411b267a6f0bcd3077f684b1a6f5e/cmake/SSGCommon.cmake#L755

ggbecker:
> > That's why I had initially an `OR` logic expression in the condition for `ssg_build_ansible_playbooks` call... I guess I'll have to reevaluate some things. Thanks for noticing that.
> 
> I think the `AND` makes sense. A solution could be to declare the custom target directly in the `ssg_build_product` macro, like in:
> 
> https://github.com/ComplianceAsCode/content/blob/a70543850e8411b267a6f0bcd3077f684b1a6f5e/cmake/SSGCommon.cmake#L755

hopefully that's what you had in mind :) https://github.com/ComplianceAsCode/content/pull/7039/commits/f7bb68f1a125d906eceb58a4c18960339e780cd0

cmake/SSGCommon.cmake
```diff
@@ -769,7 +765,13 @@ macro(ssg_build_product PRODUCT)
     ssg_build_xccdf_unlinked(${PRODUCT})
     ssg_build_ocil_unlinked(${PRODUCT})
     ssg_build_remediations(${PRODUCT})
-    if ("${PRODUCT_ANSIBLE_REMEDIATION_ENABLED}")
+
+    add_custom_target(
+        generate-${PRODUCT}-ansible-playbooks
+        DEPENDS "${ANSIBLE_PLAYBOOKS_DIR}"
```
yuumasato, line 771: I think the `DEPENDS` needs to stay within `ssg_build_ansible_playbooks`, as at this point it was not defined yet.

cmake/SSGCommon.cmake
```diff
@@ -277,10 +277,6 @@ macro(ssg_build_ansible_playbooks PRODUCT)
         DEPENDS "${SSG_BUILD_SCRIPTS}/build_rule_playbooks.py"
         COMMENT "[${PRODUCT}-content] Generating Ansible Playbooks"
     )
-    add_custom_target(
-        generate-${PRODUCT}-ansible-playbooks
-        DEPENDS "${ANSIBLE_PLAYBOOKS_DIR}"
-    )
```
yuumasato, line 283: Here the dependency can be added to the custom_target.
```
    add_dependencies(
        generate-${PRODUCT}-ansible-playbooks
        "${ANSIBLE_PLAYBOOKS_DIR}"
    )
```

yuumasato (changes requested):
> hopefully that's what you had in mind :) [f7bb68f](https://github.com/ComplianceAsCode/content/commit/f7bb68f1a125d906eceb58a4c18960339e780cd0)

Almost there, see my comments.

ggbecker:
> > hopefully that's what you had in mind :) [f7bb68f](https://github.com/ComplianceAsCode/content/commit/f7bb68f1a125d906eceb58a4c18960339e780cd0)
> 
> Almost there, see my comments.

Okay, done in https://github.com/ComplianceAsCode/content/pull/7039/commits/141dce878cc62f4a8f59f38e42148081075ea786 (amended commit)

yuumasato:
> > > hopefully that's what you had in mind :) [f7bb68f](https://github.com/ComplianceAsCode/content/commit/f7bb68f1a125d906eceb58a4c18960339e780cd0)
> > 
> > 
> > Almost there, see my comments.
> 
> Okay, done in [141dce8](https://github.com/ComplianceAsCode/content/commit/141dce878cc62f4a8f59f38e42148081075ea786) (amended commit)

Now I get warnings when doing this, :(
`cmake -DSSG_ANSIBLE_PLAYBOOKS_PER_RULE_ENABLED:BOOLEAN=ON ../`

@ggbecker I apologize, my suggestion doesn't work as expected. `add_dependencies()` only works on targets, not files.

I think you'll need to have a better look on why this dependency is added here:
https://github.com/ComplianceAsCode/content/blob/141dce878cc62f4a8f59f38e42148081075ea786/cmake/SSGCommon.cmake#L805-L808

To me it looks like it should be moved closer to where it is defined.

ggbecker:
> > > > hopefully that's what you had in mind :) [f7bb68f](https://github.com/ComplianceAsCode/content/commit/f7bb68f1a125d906eceb58a4c18960339e780cd0)
> > > 
> > > 
> > > Almost there, see my comments.
> > 
> > 
> > Okay, done in [141dce8](https://github.com/ComplianceAsCode/content/commit/141dce878cc62f4a8f59f38e42148081075ea786) (amended commit)
> 
> Now I get warnings when doing this, :(
> `cmake -DSSG_ANSIBLE_PLAYBOOKS_PER_RULE_ENABLED:BOOLEAN=ON ../`
> 
> @ggbecker I apologize, my suggestion doesn't work as expected. `add_dependencies()` only works on targets, not files.
> 
> I think you'll need to have a better look on why this dependency is added here:
> 
> https://github.com/ComplianceAsCode/content/blob/141dce878cc62f4a8f59f38e42148081075ea786/cmake/SSGCommon.cmake#L805-L808
> 
> To me it looks like it should be moved closer to where it is defined.

I brought the dependency definition closer to the target definition. That might fix the issue, although I'm not entirely sure if this would cause any undesired side effect. See: 02df119b92e31f2454e9ef30986c791639c65661

yuumasato:
@ggbecker The rule playbooks were not being built with 02df119.
I made a PR to your branch.
It would also be good to squash the last few commits to clean up a bit this back and forth with the target and dependencies.

ggbecker:
> @ggbecker The rule playbooks were not being built with [02df119](https://github.com/ComplianceAsCode/content/commit/02df119b92e31f2454e9ef30986c791639c65661).
> I made a PR to your branch.

Thank you very much.

> It would also be good to squash the last few commits to clean up a bit this back and forth with the target and dependencies.

I definitely will :D

cmake/SSGCommon.cmake
yuumasato: As the profile playbooks installed in `${SSG_ANSIBLE_ROLE_INSTALL_DIR}/` are actually playbooks (not roles), how about installing the rule playbooks in `${SSG_ANSIBLE_ROLE_INSTALL_DIR}/rule_playbooks/${PRODUCT}`?

yuumasato (approved):
LGTM, I just have one comment about the install path of rule playbooks.

cmake/SSGCommon.cmake
ggbecker: I agree that the naming is a bit confusing here. I guess it's better to have explicit than some hidden meaning. I'll update it accordingly. Thanks

ggbecker:
> LGTM, I just have one comment about the install path of rule playbooks.

I have updated the installing directory.

yuumasato:
@JAORMX Hey man, I see a a lot of "didn't match expected results" in the rhcos e2e, are these somehow expected?
I don't see how changes in this PR would cause it.
`The expected result for the audit_rules_etc_group_open rule didn't match. Expected 'PASS', Got 'FAIL' `

yuumasato:
Test results look good now, :)

PR #7040, opened 2021-05-24
Flip more rules to RHEL9

matejak:
Enable more rules - it is not about large groups that are enabled for one single reason, but it is rather about small "packs" of rules.

Those "packs" are grouped in individual commits, and the common rationale is that there are no notable changes in behavior in related components.


yuumasato (approved):
LGTM, all tests are passing

PR #7042, opened 2021-05-24
Ansible playbook to role updates

carlosmmatos:
#### Description:

- fixed issue with when clause ordering
- minor change to readme template
- fix bug where the script adds commits even when there are no changes
  - This should only commit to the repo when a change is detected, not all the time as it currently stands.

### DIFF
--- a/utils/ansible_playbook_to_role.py
+++ b/utils/ansible_playbook_to_role.py
@@ -212,7 +212,7 @@ class PlaybookToRoleConverter():
                 task["when"] = [task["when"]]
 
             variables_to_add = {tag for tag in task["tags"] if self._tag_is_valid_variable(tag)}
-            task["when"] = ["{varname} | bool".format(varname=v) for v in variables_to_add] + task["when"]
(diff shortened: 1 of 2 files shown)


utils/ansible_playbook_to_role.py
```diff
@@ -396,9 +396,8 @@ def _local_content(self, filepath):
 
     def _remote_content(self, filepath):
         remote = self.remote_repo.get_contents(filepath)
-        content = remote.decoded_content
-        if filepath == 'README.md':
-            content = content.decode("utf-8")
+        # We rant the raw string to compare against _local_content
```
yuumasato, line 399:
```suggestion
        # We want the raw string to compare against _local_content
```

yuumasato (approved):
LGTM, just found a small typo.

utils/ansible_playbook_to_role.py
carlosmmatos: lol whoops.

yuumasato (approved):

PR #7047, opened 2021-05-25
Fix typo in Ansible for idle-activation-enabled

yuumasato:
#### Description:

- The parameter name should have dashes instead of underscores.

#### Rationale:

- Typo causing the rule to fail.


jan-cerny (approved):

ggbecker (changes requested):
I've seen two occurrences of `idle_activation_enabled` in `ocil_clause`, please update them as well.

ggbecker:
> > I've seen two occurrences of `idle_activation_enabled` in `ocil_clause`, please update them as well.
> 
> @ggbecker Fixed, only found one occurence, thinking .

I didn't mention but it was on a different file ;(
https://github.com/ComplianceAsCode/content/blob/9d9b3b4cb3650eb66e2fa89c6dc82416bd9a7ff3/linux_os/guide/system/software/gnome/gnome_screen_locking/dconf_gnome_screensaver_idle_activation_locked/rule.yml#L42

yuumasato:
Alright, fixed.

ggbecker (approved):

ggbecker:
Merging it. No need to wait for OCP results

PR #7051, opened 2021-05-27
OCP4: Disable usbguard rules from RHCOS4 moderate profile

JAORMX:
These rules are very problematic and are the main reason for this job
being flaky... the problem is usbguard crashing which is a bug.

Let's disable these for now as a flaky service brings no value.


jhrozek:
content-test-filtering test says: `stderr: error: could not lock config file .git/config: No space left on device`

jhrozek (approved):
lgtm

JAORMX:
@openscap-ci test this please

PR #7054, opened 2021-05-27
SLES-15-020102 new STIG rule

vtrubovics:
#### Description:

Add rule for sudo Re-Authentication when using the sudo command

#### Rationale:

Add rule concerning sudo Re-Authentication timestamp_timeout to SLE15 stig profile

### DIFF
--- a/linux_os/guide/system/software/sudo/sudo_require_reauthentication/oval/shared.xml
+++ b/linux_os/guide/system/software/sudo/sudo_require_reauthentication/oval/shared.xml
@@ -1,0 +1,25 @@
+<def-group>
+  <definition class="compliance" id="{{{ rule_id }}}" version="1">
+    {{{ oval_metadata("'Ensure sudo timestamp_timeout is appropriate - sudo timestamp_timeout") }}}
+    <criteria comment="The timestamp_timeout should be configured" >
+      <criterion comment="check configuration in /etc/sudoers" test_ref="test_sudo_timestamp_timeout" />
+    </criteria>
+  </definition>
+
+  <ind:textfilecontent54_test check="all" check_existence="all_exist" comment="check correct configuration in /etc/sudoers" id="test_sudo_timestamp_timeout" version="1">
+    <ind:object object_ref="obj_sudo_timestamp_timeout"/>
+    <ind:state state_ref="state_sudo_timestamp_timeout" />
+  </ind:textfilecontent54_test>
+
+  <ind:textfilecontent54_object id="obj_sudo_timestamp_timeout" version="1">
+    <ind:filepath>/etc/sudoers</ind:filepath>
+    <ind:pattern operation="pattern match">^[\s]*Defaults[\s]+timestamp_timeout=([\d]+)$</ind:pattern>
+    <ind:instance datatype="int" operation="greater than or equal">1</ind:instance>
+  </ind:textfilecontent54_object>
+
+  <ind:textfilecontent54_state id="state_sudo_timestamp_timeout"
+  version="1">
+    <ind:subexpression datatype="int" operation="greater than or equal">0</ind:subexpression>
+  </ind:textfilecontent54_state>
+
+</def-group>
--- a/linux_os/guide/system/software/sudo/sudo_require_reauthentication/rule.yml
+++ b/linux_os/guide/system/software/sudo/sudo_require_reauthentication/rule.yml
@@ -1,0 +1,42 @@
+documentation_complete: true
+
+prodtype: sle15
+
+title: 'The operating system must require Re-Authentication when using the sudo command.
+ Ensure sudo timestamp_timeout is appropriate - sudo timestamp_timeout'
+
+description: |-
+    The sudo <tt>timestamp_timeout</tt> tag sets the amount of time sudo password prompt waits.
+    The default <tt>timestamp_timeout</tt> value is 5 minutes.
+    The timestamp_timeout should be configured by making sure that the
+    <tt>timestamp_timeout</tt> tag exists in
+    <tt>/etc/sudoers</tt> configuration file or any sudo configuration snippets
+    in <tt>/etc/sudoers.d/</tt>.
+    If the value is set to an integer less than 0, the user's time stamp will not expire 
+    and the user will not have to re-authenticate for privileged actions until the user's session is terminated.
+
+rationale: |-
+    Without re-authentication, users may access resources or perform tasks for which they
+    do not have authorization.
+    <br /><br />
+    When operating systems provide the capability to escalate a functional capability, it
+    is critical that the user re-authenticate.
+
+severity: medium
+
+identifiers:
+    cce@sle15: CCE-85764-9
+
+references:
+    disa: CCI-002038
+    stigid@sle15: SLES-15-020102
+    srgid: SRG-OS-000373-GPOS-00156
+
+ocil_clause: 'timestamp_timeout is not set with the appropriate value for sudo.'
+
+ocil: |-
+    Verify the operating system requires re-authentication 
+    when using the "sudo" command to elevate privileges, run the following command:
+    <pre>sudo grep -ri '^Defaults.*timestamp_timeout' /etc/sudoers /etc/sudoers.d</pre>
+    The output should be:
+    <pre>/etc/sudoers:Defaults timestamp_timout=0</pre> or "timestamp_timout" is set to a positive number.
--- a/linux_os/guide/system/software/sudo/sudo_require_reauthentication/rule.yml
+++ b/linux_os/guide/system/software/sudo/sudo_require_reauthentication/rule.yml
@@ -48,3 +48,4 @@ template:
         variable_name: "var_sudo_timestamp_timeout"
     backends:
         oval: "off"
+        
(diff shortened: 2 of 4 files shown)


ggbecker (changes requested):
Let's create a variable to hold the timeout value. The STIG defines that any positive value and zero should be considered a valid value. Check this PR that uses a similar approach https://github.com/ComplianceAsCode/content/pull/7058/files

The template does not support integer variables, so we can leave this custom OVAL you are proposing but take advantage of the template to generate the remediations, it should work. I will suggest the change to the rule.yml with the template section.

linux_os/guide/system/software/sudo/sudo_require_reauthentication/rule.yml
```diff
@@ -0,0 +1,42 @@
...
+    srgid: SRG-OS-000373-GPOS-00156
+
+ocil_clause: 'timestamp_timeout is not set with the appropriate value for sudo.'
+
+ocil: |-
+    Verify the operating system requires re-authentication 
+    when using the "sudo" command to elevate privileges, run the following command:
+    <pre>sudo grep -ri '^Defaults.*timestamp_timeout' /etc/sudoers /etc/sudoers.d</pre>
+    The output should be:
+    <pre>/etc/sudoers:Defaults timestamp_timout=0</pre> or "timestamp_timout" is set to a positive number.
```
ggbecker, line 42:
```suggestion
    <pre>/etc/sudoers:Defaults timestamp_timout=0</pre> or "timestamp_timout" is set to a positive number.

template:
    name: sudo_defaults_option
    vars:
        option: timestamp_timeout
        variable_name: "var_sudo_timestamp_timeout"
    backends:
        oval: "off"
```

linux_os/guide/system/software/sudo/sudo_require_reauthentication/rule.yml
```diff
@@ -0,0 +1,42 @@
...
+
+identifiers:
+    cce@sle15: CCE-85764-9
+
+references:
+    disa: CCI-002038
+    stigid@sle15: SLES-15-020102
+    srgid: SRG-OS-000373-GPOS-00156
+
+ocil_clause: 'timestamp_timeout is not set with the appropriate value for sudo.'
```
ggbecker, line 35:
```suggestion
ocil_clause: 'timestamp_timeout is not set with the appropriate value for sudo'
```
You don't need to include `period` here. The ocil clause is expanded to this: `Is it the case that <ocil_clause> ?` Check: https://complianceascode.readthedocs.io/en/latest/manual/developer/06_contributing_with_content.html?highlight=ocil_clause#rules

linux_os/guide/system/software/sudo/sudo_require_reauthentication/rule.yml
vtrubovics: For some reasons I cannot add var_sudo_timestamp_timeout variable to the stig.profile.
Created issue:
https://github.com/SUSE/ComplianceAsCodeContent/issues/373

ggbecker: You have to create a new file similar to this one: https://github.com/ComplianceAsCode/content/pull/7058/files#diff-ed2c51c8061e7bb15626a40dadb3c414320fa152992463ac5acda720d61f5639

named: `var_sudo_timestamp_timeout.var`

vtrubovics: Did suggested fixes.

linux_os/guide/system/software/sudo/var_sudo_timestamp_timeout.var
```diff
@@ -0,0 +1,22 @@
+documentation_complete: true
```
ggbecker, line 1: Please change the description and other items accordingly. This one references the passwd_timeout and we want to describe timestamp_timeout, which has a slightly different meaning

sle15/profiles/stig.profile
```diff
@@ -19,6 +19,7 @@ selections:
     - var_accounts_fail_delay=4
     - var_accounts_tmout=15_min
     - inactivity_timeout_value=15_minutes
+    - var_sudo_timestamp_timeout=1_minute
```
ggbecker, line 22:
```suggestion
    - var_sudo_timestamp_timeout=always_prompt
```

STIG recommends 0, even though positive values, I think the default one should be 0 and if people want to have different value here then they can customize it.

linux_os/guide/system/software/sudo/sudo_require_reauthentication/oval/shared.xml
```diff
@@ -0,0 +1,25 @@
...
+
+  <ind:textfilecontent54_object id="obj_sudo_timestamp_timeout" version="1">
+    <ind:filepath>/etc/sudoers</ind:filepath>
+    <ind:pattern operation="pattern match">^[\s]*Defaults[\s]+timestamp_timeout=([\d]+)$</ind:pattern>
+    <ind:instance datatype="int" operation="greater than or equal">1</ind:instance>
+  </ind:textfilecontent54_object>
+
+  <ind:textfilecontent54_state id="state_sudo_timestamp_timeout"
+  version="1">
+    <ind:subexpression datatype="int" operation="greater than or equal">0</ind:subexpression>
```
ggbecker, line 22: You need to use the variable you have just created here. See this: https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/accounts/accounts-restrictions/password_storage/accounts_password_pam_unix_rounds_system_auth/oval/shared.xml#L31

linux_os/guide/system/software/sudo/var_sudo_timestamp_timeout.var
vtrubovics: Fixed.

linux_os/guide/system/software/sudo/sudo_require_reauthentication/oval/shared.xml
ggbecker: Actually it doesn't make sense to use the variable here in the OVAL since any positive value should be valid.

linux_os/guide/system/software/sudo/sudo_require_reauthentication/rule.yml
```diff
@@ -0,0 +1,51 @@
...
+    srgid: SRG-OS-000373-GPOS-00156
+
+ocil_clause: 'timestamp_timeout is not set with the appropriate value for sudo'
+
+ocil: |-
+    Verify the operating system requires re-authentication 
+    when using the "sudo" command to elevate privileges, run the following command:
+    <pre>sudo grep -ri '^Defaults.*timestamp_timeout' /etc/sudoers /etc/sudoers.d</pre>
+    The output should be:
+    <pre>/etc/sudoers:Defaults timestamp_timout=0</pre> or "timestamp_timout" is set to a positive number.
```
ggbecker, line 42:
```suggestion
    <pre>/etc/sudoers:Defaults timestamp_timeout=0</pre> or "timestamp_timeout" is set to a positive number.
```

I know that this is a copy paste from the STIG, but can you please fix this typo?

linux_os/guide/system/software/sudo/sudo_require_reauthentication/oval/shared.xml
vtrubovics: Yes. The OVAL code is not needed here because ansible and OVAL code are genererated by template. I will do the fix.

vtrubovics: Commited fix.

ggbecker: In my opinion we need to have this custom OVAL here. The OVAL generated by the template has a different subexpression comparison and that will not work with this rule (this rule is about checking positive integers). That's why I suggested the `oval: "off"` in the template data section.

See this: https://github.com/ComplianceAsCode/content/blob/master/shared/templates/sudo_defaults_option/oval.template#L22

You can revert back to the original implementation.

vtrubovics: Reverted previous commit and did recommended fix.

linux_os/guide/system/software/sudo/sudo_require_reauthentication/oval/shared.xml
```diff
@@ -0,0 +1,27 @@
...
+    <ind:pattern operation="pattern match">^[\s]*Defaults[\s]+timestamp_timeout=([\d]+)$</ind:pattern>
+    <ind:instance datatype="int" operation="greater than or equal">1</ind:instance>
+  </ind:textfilecontent54_object>
+
+  <ind:textfilecontent54_state id="state_sudo_timestamp_timeout"
+  version="1">
+    <ind:subexpression datatype="string" operation="equals" var_ref="var_sudo_timestamp_timeout" />
+  </ind:textfilecontent54_state>
+
+  <external_variable id="var_sudo_timestamp_timeout" datatype="string" version="1" comment="Variable value for sudo timestamp_timeout" />
```
ggbecker, line 25: Sorry for the confusion here, but the OVAL actually doesn't need to check for the variable. For example, if someone selects the variable to be `3_minutes`, these OVAL check will invalid `0, 1 and 2`, which should still pass. The variable in this case should be only used for the remediation as it's already being used by the template. You can revert the last commit.

linux_os/guide/system/software/sudo/sudo_require_reauthentication/oval/shared.xml
vtrubovics: Reverted OVAL check for variable.

linux_os/guide/system/software/sudo/sudo_require_reauthentication/oval/shared.xml
vtrubovics: Reverted commit (a9e7f71).

ggbecker (approved):

PR #7055, opened 2021-05-27
RHEL-08-020220 updates

### LINKED ISSUE #6955: RHEL-08-020220 remediation does not write "remember=5" to password-auth
#### Description of problem:

RHEL-08-020100 remediation does not write "remember=5" to password-auth, nor does it write it to the correct line in system-auth.  It writes it to "password sufficient pam_unix.so" when the STIG specifies to write to "password required pam_pwhistory.so" line in both password-auth and system-auth.

0.1.55

rhel8 8.3

#### Steps to Reproduce:

1.  Apply ansible playbook to fresh rhel8 system.
2.  Apply oscap xccdf eval --remediate to same system
3.  sudo grep -i remember /etc/pam.d/system-auth /etc/pam.d/password-auth does not show remember=5

#### Actual Results:

password sufficient pam_unix.so  remember=5

#### Expected Results:

password required pam_pwhistory.so remember=5 in both  password-auth and system-auth.

N/A

carlosmmatos:
#### Description:

- Fix issue where the rule was only targeting system-auth file, and not both system-auth and password-auth.
- Also added test scenarios.

#### Rationale:

- This particular rule needed to be changed. I didn't see the need to maintain the checks for pam_unix.so since the remember option for that module is deprecated in favor of pam_pwhistory.so. If this needs to be added back, I suggest a new rule, or we change the names to be more specific.
- The bash and ansible were extremely inflexible, probably because pam_unix.so is generally a standard module in the pam stack. These needed to be overhauled.
  - I might also consider turning some of this into a macro/template to be used for - I was looking at the pam_options but that doesn't give you the ability to run on more than one file along with some other reasons.

- Fixes #6955

### DIFF
--- a/linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_password_pam_pwhistory_remember_password_auth/rule.yml
+++ b/linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_password_pam_pwhistory_remember_password_auth/rule.yml
@@ -24,16 +24,25 @@ identifiers:
     cce@rhel8: CCE-83478-8
 
 references:
+    stigid@ol7: OL07-00-010270
+    cis@rhel7: 5.3.3
+    cis@rhel8: 5.4.3
+    cjis: 5.6.2.1.1
+    cui: 3.5.8
     disa: CCI-000200
     nist: IA-5(f),IA-5(1)(e)
     nist-csf: PR.AC-1,PR.AC-6,PR.AC-7
+    nist@sle15: IA-5(1)(e),IA-5(1).1(v)
+    pcidss: Req-8.2.5
     srg: SRG-OS-000077-GPOS-00045
     vmmsrg: SRG-OS-000077-VMM-000440
+    stigid@rhel7: RHEL-07-010270
     isa-62443-2013: 'SR 1.1,SR 1.10,SR 1.2,SR 1.3,SR 1.4,SR 1.5,SR 1.7,SR 1.8,SR 1.9,SR 2.1'
     isa-62443-2009: 4.3.3.2.2,4.3.3.5.1,4.3.3.5.2,4.3.3.6.1,4.3.3.6.2,4.3.3.6.3,4.3.3.6.4,4.3.3.6.5,4.3.3.6.6,4.3.3.6.7,4.3.3.6.8,4.3.3.6.9,4.3.3.7.2,4.3.3.7.4
     cobit5: DSS05.04,DSS05.05,DSS05.07,DSS05.10,DSS06.03,DSS06.10
     iso27001-2013: A.18.1.4,A.7.1.1,A.9.2.1,A.9.2.2,A.9.2.3,A.9.2.4,A.9.2.6,A.9.3.1,A.9.4.2,A.9.4.3
     cis-csc: 1,12,15,16,5
+    stigid@rhel8: RHEL-08-020220
 
 ocil_clause: |-
      the value of remember is not set equal to or greater than {{{ xccdf_value("var_password_pam_remember") }}}
--- a/linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_password_pam_pwhistory_remember_password_auth/rule.yml
+++ b/linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_password_pam_pwhistory_remember_password_auth/rule.yml
@@ -1,0 +1,51 @@
+documentation_complete: true
+
+prodtype: fedora,ol7,ol8,rhel7,rhel8,rhel9,rhv4,wrlinux1019
+
+title: 'Limit Password Reuse: password-auth'
+
+description: |-
+    Do not allow users to reuse recent passwords. This can be
+    accomplished by using the <tt>remember</tt> option for the
+    <tt>pam_pwhistory</tt> PAM modules.
+    <br /><br />
+    In the file <tt>/etc/pam.d/password-auth</tt>, make sure the parameter
+    <tt>remember</tt> is present, and that the value
+    for the <tt>remember</tt> parameter is {{{ xccdf_value("var_password_pam_remember") }}} or greater. For example:
+    <pre>password {{{ xccdf_value("var_password_pam_remember_control_flag") }}} pam_pwhistory.so <i>...existing_options...</i> remember={{{ xccdf_value("var_password_pam_remember") }}} use_authtok</pre>
+    The DoD STIG requirement is 5 passwords.
+
+rationale: 'Preventing re-use of previous passwords helps ensure that a compromised password is not re-used by a user.'
+
+severity: medium
+
+identifiers:
+    cce@rhel7: CCE-83476-2
+    cce@rhel8: CCE-83478-8
+
+references:
+    disa: CCI-000200
+    nist: IA-5(f),IA-5(1)(e)
+    nist-csf: PR.AC-1,PR.AC-6,PR.AC-7
+    srg: SRG-OS-000077-GPOS-00045
+    vmmsrg: SRG-OS-000077-VMM-000440
+    isa-62443-2013: 'SR 1.1,SR 1.10,SR 1.2,SR 1.3,SR 1.4,SR 1.5,SR 1.7,SR 1.8,SR 1.9,SR 2.1'
+    isa-62443-2009: 4.3.3.2.2,4.3.3.5.1,4.3.3.5.2,4.3.3.6.1,4.3.3.6.2,4.3.3.6.3,4.3.3.6.4,4.3.3.6.5,4.3.3.6.6,4.3.3.6.7,4.3.3.6.8,4.3.3.6.9,4.3.3.7.2,4.3.3.7.4
+    cobit5: DSS05.04,DSS05.05,DSS05.07,DSS05.10,DSS06.03,DSS06.10
+    iso27001-2013: A.18.1.4,A.7.1.1,A.9.2.1,A.9.2.2,A.9.2.3,A.9.2.4,A.9.2.6,A.9.3.1,A.9.4.2,A.9.4.3
+    cis-csc: 1,12,15,16,5
+
+ocil_clause: |-
+     the value of remember is not set equal to or greater than {{{ xccdf_value("var_password_pam_remember") }}}
+
+ocil: |-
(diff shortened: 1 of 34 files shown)


linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_password_pam_unix_remember/ansible/shared.yml
```diff
@@ -5,14 +5,77 @@
...
-    dest: /etc/pam.d/system-auth
-    regexp: '^(password\s+sufficient\s+pam_unix\.so\s.*remember\s*=\s*)(\S+)(.*)$'
-    replace: '\g<1>{{ var_password_pam_unix_remember }}\g<3>'
+- name: "{{{ rule_title }}}: set_fact"
+  set_fact:
+    {{% if product == "rhel8" -%}}
+    control: 'required'
+    {{%- else -%}}
+    control: 'requisite'
+    {{%- endif %}}
```
yuumasato, line 14: I think these could just be build time jinja variables.
Once content is built an Ansible task is not expected to handle multiple platforms.

Or do you see value in having this customizable by an admin?

linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_password_pam_unix_remember/ansible/shared.yml
```diff
@@ -5,14 +5,77 @@
...
+        - password-auth
+  when: pam_pwhistory.results[0].found == 1 or pam_pwhistory.results[1].found == 1
+
+# if entry does not exist - also to prevent duplicating
+- name: "{{{ rule_title }}}: Add pam_pwhistory"
+  pamd:
+    name: "{{ item }}"
+    type: password
+    control: required
+    module_path: pam_deny.so
```
yuumasato, line 71: When remediating RHEL8 right after install, the `pam_history.so` module is placed after a few `sufficient` modules, resulting in no password remember check. 

```
password   requisite pam_pwquality.so try_first_pass local_users_only
password   sufficient pam_unix.so sha512 shadow nullok try_first_pass use_authtok
password   sufficient pam_sss.so use_authtok
password   required pam_pwhistory.so use_authtok remember=5
password   required pam_deny.so
```

The pam_pwhistory.so needs to be added before pam_unix.so. The bash remediation does that, and reused passwords are checked.

yuumasato (changes requested):
>     * I didn't see the need to maintain the checks for pam_unix.so since the remember option for that module is deprecated in favor of pam_pwhistory.so. If this needs to be added back, I suggest a new rule, or we change the names to be more specific.

I' don't think it is a good idea to have `pam_unix` in the rule ID, and deal with `pam_pwhistory`.
I think we need new rules deailng with `remember` in `pam_pwhistory` module.

Note: With new rules, I think this supersedes #6672.

.github/workflows/validate_ansible_roles.yaml
```diff
@@ -8,23 +8,28 @@ jobs:
   validate:
     name: Validate Ansible Roles
     runs-on: ubuntu-20.04
+    container: fedora:latest
```
vojtapolasek, line 11: Is this change somehow directly related to this PR?

vojtapolasek (changes requested):
Hello, thank you very much for these rules, they will be very helpful.
Just one nitpick and one question, see comments please.

linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_password_pam_pwhistory_remember_system_auth/rule.yml
```diff
@@ -0,0 +1,60 @@
...
+references:
+    stigid@ol7: OL07-00-010270
+    cis@rhel7: 5.3.3
+    cis@rhel8: 5.4.3
+    cjis: 5.6.2.1.1
+    cui: 3.5.8
+    disa: CCI-000200
+    nist: IA-5(f),IA-5(1)(e)
+    nist-csf: PR.AC-1,PR.AC-6,PR.AC-7
+    nist@sle15: IA-5(1)(e),IA-5(1).1(v)
```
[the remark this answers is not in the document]
carlosmmatos: Yes, I can merge those those

.github/workflows/validate_ansible_roles.yaml
carlosmmatos: For some reason, this PR seemed to fail with the existing ubuntu-20.04 setup. I tracked it down to the complaint of pyYAML, which upstream fedora had later version and this did not fail.
https://github.com/ComplianceAsCode/content/pull/7055/checks?sha=0e2d887d8e8feabf073c2e97cee7c2e788497cae#step:6:28

However, let me double-check this issue and make sure it's not content-related, in which case I can just revert this change.

carlosmmatos: @vojtapolasek Okay - for context reference, there were a couple of issues at play here.. 

1. Having a block in the following order caused issues:
```yaml
    - name: 'Limit Password Reuse: system-auth'
      block:

        - name: 'Limit Password Reuse: system-auth - Ensure correct control value in pam_pwhistory'
          pamd:
            name: system-auth
            type: password
        ...
```

-   It doesn't like the two `- name` stanza's having single quotes
-   I had to remove the `{{{ rule_title }}}` from the second `-name` to get it to work. 

2. There was some sort of discrepancy in the reference section for the `accounts_password_pwhistory_remember_password_auth` rule.yml file that was causing a failure with checking for valid yaml tags in https://github.com/ComplianceAsCode/content/blob/master/utils/ansible_playbook_to_role.py#L322
-  I cleaned up the references and that fixed the issue.

Finally, just a point about using the container approach I suggested - none of the above issues were present, seems like newer versions of PyYAML fixed the above issues. I believe I can get the latest version of PyYAML using pip, I will be submitting a PR with this change as soon as I test it out.

vojtapolasek (approved):

.github/workflows/validate_ansible_roles.yaml
yuumasato: > I had to remove the `{{{ rule_title }}}` from the second `-name` to get it to work.

Could it be related to this https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#when-to-quote-variables-a-yaml-gotcha ?

carlosmmatos: Not sure how, when the ansible playbook to role script runs, the `{{{ rule_title }}}` has already been converted, so there would be no gotcha's for yaml at that point. My guess is, maybe instead of:
```yaml
- name: "{{{ rule_title }}}"
  ...
```
I could try removing the double quotes?:
```yaml
- name: {{{ rule_title }}}
  ...
```

linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_password_pam_pwhistory_remember_password_auth/rule.yml
```diff
@@ -0,0 +1,56 @@
...
+    nist: IA-5(f),IA-5(1)(e),IA-5(1).1(v)
+    nist-csf: PR.AC-1,PR.AC-6,PR.AC-7
+    pcidss: Req-8.2.5
+    srg: SRG-OS-000077-GPOS-00045
+    vmmsrg: SRG-OS-000077-VMM-000440
+    isa-62443-2013: 'SR 1.1,SR 1.10,SR 1.2,SR 1.3,SR 1.4,SR 1.5,SR 1.7,SR 1.8,SR 1.9,SR 2.1'
+    isa-62443-2009: 4.3.3.2.2,4.3.3.5.1,4.3.3.5.2,4.3.3.6.1,4.3.3.6.2,4.3.3.6.3,4.3.3.6.4,4.3.3.6.5,4.3.3.6.6,4.3.3.6.7,4.3.3.6.8,4.3.3.6.9,4.3.3.7.2,4.3.3.7.4
+    cobit5: DSS05.04,DSS05.05,DSS05.07,DSS05.10,DSS06.03,DSS06.10
+    iso27001-2013: A.18.1.4,A.7.1.1,A.9.2.1,A.9.2.2,A.9.2.3,A.9.2.4,A.9.2.6,A.9.3.1,A.9.4.2,A.9.4.3
+    cis-csc: 1,12,15,16,5
```
ggbecker, line 41:
```suggestion
    cis-csc: 1,12,15,16,5
    stigid@rhel8: RHEL-08-020220
```

I think we should add this stigid here as well.

linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_password_pam_pwhistory_remember_password_auth/tests/argument_missing.fail.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+# platform = Red Hat Enterprise Linux 7,Red Hat Virtualization 4,multi_platform_fedora,multi_platform_ol,multi_platform_wrlinux
+
```
ggbecker, line 3:
```suggestion

# packages = pam
```

Please add this package metadata to all test scenarios.

ggbecker (changes requested):
Can you please purge the changes related to the github actions from commit 76d78a2905b3140c963c16fe0c4d33b59831542a and 812c48ae6c31a15f1795954d699df32a5a7cb6a ?

Otherwise LGTM

linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_password_pam_pwhistory_remember_password_auth/rule.yml
carlosmmatos: I thought we agreed not to have 2 rules with the same STIG ID?

carlosmmatos:
> Can you please purge the changes related to the github actions from commit [76d78a2](https://github.com/ComplianceAsCode/content/commit/76d78a2905b3140c963c16fe0c4d33b59831542a) and [812c48a](https://github.com/ComplianceAsCode/content/commit/812c48ae6c31a15f1795954d699df32a5a7cb6a4) ?
> 
> Otherwise LGTM

Done

ggbecker (approved):
LGTM

PR #7056, opened 2021-05-28
Update cis sle15 profile to better represent the release version 1.0.0

truzzon:
#### Description:

- Update of sle15 cis.profile

Hello, not all headlines have a rule. The goal was to get the profile up-to-date with the available rules.
I'll create more PRs later to add the Rules that are still missing.


yuumasato:
@openscap-ci add to whitelist

linux_os/guide/system/accounts/accounts-pam/password_quality/password_quality_pamcracklib/cracklib_accounts_password_pam_dcredit/rule.yml
```diff
@@ -35,7 +35,7 @@ ocil: |-
     To check how many digits are required in a password, run the following
     command:
     <pre># grep pam_cracklib /etc/pam.d/common-password
-    password requisite pam_cracklib.so dcredit={{{ xccdf_value("var_password_pam_delay") }}}</pre>
+    password requisite pam_cracklib.so dcredit={{{ xccdf_value("var_password_pam_dcredit") }}}</pre>
```
yuumasato, line 38: Nice catch

linux_os/guide/system/accounts/accounts-restrictions/root_logins/root_path_default/rule.yml
```diff
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: fedora,ol7,ol8,rhel7,rhel8
+prodtype: fedora,ol7,ol8,rhel7,rhel8,sle15
```
yuumasato, line 3: You didn't add a `cis` reference for this rule. Is that intentional?

yuumasato (approved):
LGTM, but I didn't check the SUSE CIS references in the rules and profile.
@truzzon would you like to wait for someone to review them?

truzzon:
Hello @yuumasato, I used the PDF file provided by CIS as a reference.

I added comments inside the profile, where in my opinion the check is not 100% true to the benchmark. (E.g.: A service should only be disabled, where CIS wants the whole package to be removed.)

A review of someone with more expertise couldn't hurt and I'd appreciate the input.

linux_os/guide/system/accounts/accounts-restrictions/root_logins/root_path_default/rule.yml
truzzon: No it is not. I'll add it later today. Thanks for pointing it out.

truzzon: After checking it, I removed the prodtype 'sle15' for now and commented out the reference in the profile.

Because I'm not sure that this rule can return the correct result for the use case ("Ensure root PATH Integrity").

sle15/profiles/cis.profile
```diff
@@ -692,184 +737,232 @@ selections:
...
-    ### 6.1.4 Ensure permissions on /etc/group are configured (Scored)
+    ### 6.1.4 Ensure permissions on /etc/group are configured (Automated)
     # chown root:root /etc/group
     - file_owner_etc_group
     - file_groupowner_etc_group
 
     # chmod 644 /etc/group
     - file_permissions_etc_group
 
-    ### 6.1.5 Ensure permissions on /etc/gshadow are configured (Scored)
```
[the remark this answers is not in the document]
yuumasato: If they are not part of the CIS Benchmark anymore I'd remove them from the profile, including the references in the rule.

yuumasato:
I think there is no need to hold this off any longer.

PR #7057, opened 2021-05-30
Select rule for RHEL8 CIS 1.3.2 (fixes #5220)

### LINKED ISSUE #5220: CIS 1.3.2 Ensure sudo commands use pty (Scored)
[CIS_Red_Hat_Enterprise_Linux_8_Benchmark_v1.0.0 (1).pdf](https://github.com/ComplianceAsCode/content/files/4300871/CIS_Red_Hat_Enterprise_Linux_8_Benchmark_v1.0.0.1.pdf)

alexhaydock:
#### Description:

- Add rule for RHEL 8 CIS control 1.3.2

#### Rationale:

- There is already a rule for ensuring that sudo commands only work when using a PTY. We just need to enable it.

- Fixes #5220

### DIFF
--- a/linux_os/guide/system/software/sudo/sudo_add_use_pty/rule.yml
+++ b/linux_os/guide/system/software/sudo/sudo_add_use_pty/rule.yml
@@ -20,10 +20,9 @@ identifiers:
     cce@rhel8: CCE-83798-9
 
 references:
-    cis@rhel8: 1.3.2
-
-references:
     anssi: BP28(R58)
+    cis@rhel7: 5.2.2
+    cis@rhel8: 1.3.2
 
 ocil_clause: 'use_pty is not enabled in sudo'
 
--- a/linux_os/guide/system/software/sudo/sudo_add_use_pty/rule.yml
+++ b/linux_os/guide/system/software/sudo/sudo_add_use_pty/rule.yml
@@ -18,6 +18,8 @@ severity: medium
 identifiers:
     cce@rhel7: CCE-83797-1
     cce@rhel8: CCE-83798-9
+
+references:
     cis@rhel8: 1.3.2
 
 references:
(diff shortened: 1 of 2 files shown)


yuumasato:
@openscap-ci test this please

linux_os/guide/system/software/sudo/sudo_add_use_pty/rule.yml
```diff
@@ -19,6 +19,9 @@ identifiers:
     cce@rhel7: CCE-83797-1
     cce@rhel8: CCE-83798-9
 
+references:
+    cis@rhel8: 1.3.2
```
yuumasato, line 23: Please add the `cis` reference together with the `anssi` entry below.

yuumasato (changes requested):

linux_os/guide/system/software/sudo/sudo_add_use_pty/rule.yml
alexhaydock: I've addressed this above by putting the CIS reference for RHEL 8 in the right place. I noticed a CCE for RHEL 7 was already assigned, so I played the CIS reference for RHEL 7 in the references section as well.

But when it comes to adding this rule to the RHEL 7 CIS profile, I'm facing the same problem as I faced in #7058, which is that there's no obvious place to put it or way to tag it correctly since it seems like the references in that profile are outdated. `5.2.2`, which would be the current reference for this rule, is already present as a different rule:

https://github.com/ComplianceAsCode/content/blob/8ba767fedfc5e79ce72c5fdd8c31d718162b4e07/rhel7/profiles/cis.profile#L584

yuumasato: The RHEL7 CIS profile will soon be realigned with latest benchmark.
So it should be fine to have the inconsistency of the rule having a cis@rhel7 reference` aligned with the latest benchmark, while the profile still mentions 2.2.0.

yuumasato (approved):
Looks good. Sorry for delay in coming back to this.

PR #7058, opened 2021-05-30
Add sudo custom logfile control for RHEL 8 CIS (fixes #5221)

### LINKED ISSUE #5221: CIS 1.3.3 Ensure sudo log file exists (Scored)
Needs rule
[CIS_Red_Hat_Enterprise_Linux_8_Benchmark_v1.0.0 (1).pdf](https://github.com/ComplianceAsCode/content/files/4300883/CIS_Red_Hat_Enterprise_Linux_8_Benchmark_v1.0.0.1.pdf)

alexhaydock:
#### Description:

- CIS for RHEL 8 requires enabling a custom sudo logfile.
- This is my first attempt at implementing a new rule that didn't already exist. I have tried to follow the contributor documentation but my apologies if there is still anything missing!

#### Rationale:

- This satisfies the RHEL 8 CIS control 1.3.3 which does not currently have a rule yet.

- Fixes #5221


yuumasato:
@openscap-ci add to white list

yuumasato:
@openscap-ci add to whitelist

linux_os/guide/system/software/sudo/sudo_custom_logfile/rule.yml
```diff
@@ -0,0 +1,27 @@
+documentation_complete: true
+
+title: 'Ensure Sudo Logfile Exists - sudo logfile'
+
+description: |-
+    A custom logfile can be configured for sudo with the logfile tag.
```
yuumasato, line 6: Could you also add in the description what is expected to be configured and how?
Something like what is in the `ocil` is enough.

Having this in the description makes it easy to understand what this rule is about in the HTML guides and reports.

linux_os/guide/system/software/sudo/sudo_custom_logfile/rule.yml
```diff
@@ -0,0 +1,27 @@
...
+
+rationale: |-
+    A sudo log file simplifies auditing of sudo commands.
+
+severity: low
+
```
yuumasato, line 12: Would you mind adding a CCE too?
Something like:
https://github.com/ComplianceAsCode/content/blob/8ba767fedfc5e79ce72c5fdd8c31d718162b4e07/linux_os/guide/system/software/sudo/sudo_add_use_pty/rule.yml#L20

Take a random number from `shared/references/cce-redhat-avail.txt`.

yuumasato (changes requested):
Thank you for the rule addition, I just have a few comments.

linux_os/guide/system/software/sudo/sudo_custom_logfile/rule.yml
yuumasato: Actually, the tests failed due to missing CCE.

alexhaydock:
Thanks for the helpful feedback, @yuumasato :smile: I had missed the CCE identifiers requirement but I'll keep that one in mind for future rule additions.

I've tried to resolve your comments in the latest commit, and at the same time I tried to enable this rule for RHEL 7 as well. But I'm hitting an issue enabling this for RHEL 7 since I think the latest version of the CIS benchmark has caused some of the rule identifiers to change.

I'm not sure which version of the benchmark the CIS profile for RHEL 7 in CaC is based on, but this rule is `5.2.3` in the latest RHEL 7 benchmark, according to CIS. But when I look at the `cis.profile` file, there's already a `5.2.3` for a different rule:

https://github.com/ComplianceAsCode/content/blob/8ba767fedfc5e79ce72c5fdd8c31d718162b4e07/rhel7/profiles/cis.profile#L587

I think I should be able to include this rule as-is in the CIS profile for RHEL 7 the same way I have for RHEL 8, but I'm not sure where to put it as a result of this conflict.

alexhaydock:
@yuumasato - When you (or any other reviewers) get the chance, it would be great if you could let me know whether there's anything else above what I've got here that might be expected or before a PR like this could be merged.

My aim is to go through all of the remaining "NEEDS RULE"s for RHEL 8 CIS (and RHEL 7 where it applies), and I just want to make sure I've got the process streamlined before I start firing off a load of PRs. I don't want to go creating a load of extra work for maintainers. Thanks! 😄

linux_os/guide/system/software/sudo/var_sudo_logfile.var
```diff
@@ -0,0 +1,16 @@
+documentation_complete: true
+
+title: 'Sudo - logfile value'
+
+description: |-
+    Specify the sudo logfile to use. The default value used here matches the example
+    location from CIS, which uses /var/log/sudo.log.
+
+interactive: false
```
yuumasato, line 9: Setting this to true will enable SCAP Workbench to edit/add values during tailoring. In case one doesn't want to use suggested value from CIS.

yuumasato (commented):
> I'm not sure which version of the benchmark the CIS profile for RHEL 7 in CaC is based on, but this rule is `5.2.3` in the latest RHEL 7 benchmark, according to CIS. But when I look at the `cis.profile` file, there's already a `5.2.3` for a different rule:

Yeah, the current RHEL7 CIS profile is based on version 2.2.0, so it needs an update.

> I think I should be able to include this rule as-is in the CIS profile for RHEL 7 the same way I have for RHEL 8, but I'm not sure where to put it as a result of this conflict.

I agree, this rule applies for RHEL7 CIS 3.1.0.
Not sure what is the best way to proceed with the RHEL 7 CIS update.
One way I see is to update the metadata version to 3.1.0 and reorganize the item numbers in the profile file.
As updating the profile file from 2.2.0 to 3.1.0 may require a significant effort, we could flip the RHEL7 CIS to title to`DRAFT` until all the commented items numbers and tittles are aligned to v3.1.0.

> @yuumasato - When you (or any other reviewers) get the chance, it would be great if you could let me know whether there's anything else above what I've got here that might be expected or before a PR like this could be merged.

The PR looks pretty good and clean to me. I just added one more thing I noticed that may be interesting to change (interactive: true)
In general, tests are good to have too. But as this rule is using a template for which there is at least one rule with tests, they are not required.

linux_os/guide/system/software/sudo/var_sudo_logfile.var
alexhaydock: Good catch! Yeah it's definitely useful for the user to be able to override this location and put the logfile where they want. I've pushed a new commit that sets this to `true` now. :+1:

yuumasato (approved):

PR #7063, opened 2021-06-01
improve documentation

vojtapolasek:
#### Description:

- remove hardcoded macro descriptions and rather link to the generated ones
- add doc strings for high level macros
- incorporate example/readme.md into the documentation, it describes step by step how to create a new product
- link to Content labs
- remove mention of Copr repository since we no longer maintain scap-security-guide in copr
- other minor fixes

#### Rationale:

This PR tries to get rid of unneeded stuff and make more things available in the central readthedocs.io documentation.


carlosmmatos:
My suggestions for improving documentation:

- Move Templates to top level navigation under 'DEVELOPER GUIDE' - probably above Legacy Notice
- Create documentation around variables
  - How do I know how to create a variable and how to use it?
  - Is there a difference between variables defined in profiles, as opposed to rules?

vojtapolasek:
> My suggestions for improving documentation:
> 
>     * Move Templates to top level navigation under 'DEVELOPER GUIDE' - probably above Legacy Notice
> 
>     * Create documentation around variables
>       
>       * How do I know how to create a variable and how to use it?
>       * Is there a difference between variables defined in profiles, as opposed to rules?
I don't understand this last bullet. As far as I know, you can define variable's value only in the profile, not within the rule it self.

cipherboy:
@vojtapolasek said:
> @carlosmmatos said:
> >   * Is there a difference between variables defined in profiles, as opposed to rules?
> 
> I don't understand this last bullet. As far as I know, you can define variable's value only in the profile, not within the rule it self.

Perhaps Carlos is asking about explaining XCCDF variable versus say, `local_variable`? Can you have an external/extended variable in an OVAL under `shared/checks/oval/...` and use it in a rule's OVAL?

carlosmmatos:
> @vojtapolasek said:
> 
> > @carlosmmatos said:
> > > * Is there a difference between variables defined in profiles, as opposed to rules?
> > 
> > 
> > I don't understand this last bullet. As far as I know, you can define variable's value only in the profile, not within the rule it self.
> 
> Perhaps Carlos is asking about explaining XCCDF variable versus say, `local_variable`? Can you have an external/extended variable in an OVAL under `shared/checks/oval/...` and use it in a rule's OVAL (I think so)?
> 
> Former would be one where users should change the values, `local` would be for temporary intermediates, and `extended` would be for something that might differ on a per-OS basis (e.g., I used JINJA+`product.yml` for the grub2 `/boot` sub-folder path, but I could've put it in a `shared/checks/oval` and defined it as an external_variable and updated all the OVALs using more OVAL junk rather than using the easier JINJA substitution.

Yes - basically there is 0 documentation about variables in general, whether its an XCCDF variable vs some other variable.. I just had this conversation with @yuumasato and @ggbecker and we all agreed that we don't have anything in the docs to help explain these at any level. 

So my suggestion is to document the different types of variables that can be used in the project, how they are used, how to interact with them, design considerations, etc.. just making it easier for someone to understand how to use them, and how they are used.

vojtapolasek:
OK I understand now. I confirm we have no documentation about variables. I will add it.

jan-cerny:
It's amazing!

Do you plan to add the variables in this PR or will it be a separate PR?

I suggest changing the commit message of 4897a88 from `bash: n: příkaz nenalezen` to something sensible.

vojtapolasek:
Yes I will add documentation to this PR, I am just busy with other things. And I will change the commit message.

cipherboy:
@vojtapolasek What order do we want to merge this and #7309/#7312 in?

vojtapolasek:
I must say that I currently don't have enough time to improve this PR. I will work on it later definitely, but not in following days / weeks. So feel free to merge other PRs, do not wait for this one.

cipherboy:
Do you want to merge this PR now, as-is? The variable discussion could
definitely be moved to a new PR and this one definitely has some good
stuff. No sense holding up much needed documentation in my mind.

jan-cerny:
@vojtapolasek I still see most of the changes valuable. I think we should merge it as is and the variables can be addressed in another PR. What do you think? Can you rebase so that we can proceed with merging?

vojtapolasek:
I am OK with thath, I am glad you find it useful. I have just rebased.

jan-cerny (approved):

PR #7065, opened 2021-06-01
Remove deprecated install command from RHEL8 kickstarts

matusmarhefka:
#### Description:

* Remove deprecated install command from RHEL8 kickstarts, Anaconda prints the following warning when installing RHEL8:
```   
    The install command has been deprecated and no longer has any effect.
    It may be removed from future releases, which will result in a fatal
    error from kickstart.  Please modify your kickstart file to remove
    this command.
```
* Increased default memory in `install_vm.py` as for RHEL8 min. memory is 1.5 GB per logical CPU core: https://access.redhat.com/articles/rhel-limits

There is one more warning from Anaconda about authconfig:
```
org.fedoraproject.Anaconda.Modules.Security[1733]: INFO:program:IMPORTANT: authconfig is replaced by authselect, please update your scripts.
```
but I am not sure if we can just replace `authconfig` with `authselect` as there might be also some rules in profiles
which would need the similar update. This needs more investigation (outside of this PR).

### DIFF
--- a/rhel8/kickstart/ssg-rhel8-cis-ks.cfg
+++ b/rhel8/kickstart/ssg-rhel8-cis-ks.cfg
@@ -6,9 +6,6 @@
 # https://pykickstart.readthedocs.io/en/latest/
 # https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/performing_an_advanced_rhel_installation/index#performing_an_automated_installation_using_kickstart
 
-# Install a fresh new system (optional)
-install
-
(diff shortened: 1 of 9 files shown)


vojtapolasek:
Looks good, thank you. Just a question, are you going to solve this somehow also for the kickstart used for installing of test machines?

matusmarhefka:
@vojtapolasek KS for test suite has already been updated in https://github.com/ComplianceAsCode/content/pull/7004

vojtapolasek:
I see. Merging.

PR #7067, opened 2021-06-01
Fix oval check in uefi_no_removeable_media

freddieRv:
#### Description:

- The group `['|\(]` already matches the closing `'`, so the extra `.*'` at the end of the regex was causing false negatives

#### Rationale:

- This fixes false negatives


ggbecker (approved):
Good catch.

PR #7068, opened 2021-06-02
OCP4: Upgrade builder and target images

JAORMX:
The image we use to generate upstream content has been upgraded to use
Fedora 34. Also, the target image now uses ubi-micro.


jhrozek (approved):
The change itself is trivial, if tests pass, go ahead and merge.

PR #7069, opened 2021-06-02
Add rule to configure MaxStartups for sshd

### LINKED ISSUE #5528: 5.2.18 Ensure SSH MaxStartups is configured (Scored)

alexhaydock:
#### Description:

- Adds rule for CIS "Ensure SSH MaxStartups is configured"

#### Rationale:

- Based on pre-existing sshd configuration rules, this adds a new rule and appropriate tests to configure MaxStartups for the SSH daemon.
- I have made this variable configurable as CIS is not too opinionated about the value set here. It notes that the output should be "10:30:60 or matches site policy".
- Fixes #5528


rhel7/profiles/cis.profile
```diff
@@ -625,6 +625,10 @@ selections:
     ### 5.2.15 Ensure SSH warning banner is configured (Scored)
     - sshd_enable_warning_banner
 
+    ### 5.2.18 Ensure SSH MaxStartups is configured (Scored)
+    - sshd_set_maxstartups
+    - var_sshd_set_maxstartups=10:30:60
```
alexhaydock, line 630: The build process is struggling particularly where I have configured this variable. I tried quoting the strings but it appears to fail still.

Strictly speaking, we don't _need_ to specify the variable in the profile itself, as this is already the value that will be set by the rule, but I wanted to configure it in the actual profile so that the CIS-recommended value is always set even if the rule defaults change elsewhere in the project in future.

yuumasato: What happens is that `10:30:60` is not a selector for the variable, but the value of the variable. (Also, `default` is not a valid selector).
To explicitly list the variable selection in the profile you to need add a selector with the same value.
Check the example below, `default` has the same value as the `no` selector.
https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/ssh/ssh_server/var_sshd_disable_compression.var

alexhaydock: Ah, I had assumed that the variable value entered into the `Profile` was a "freeform" field that the user could enter anything into, but now I've had a look at a few other `vars` files I understand how this works now. Thanks!

alexhaydock: @yuumasato Interestingly though I seem to have still broken the CI job even though I tried quoting the strings in the `Profile` when explicitly setting the variable:
```
KeyError: 'var_sshd_set_maxstartups=10'
```
It seems to hit the `:` character and ignore the remainder of the line, even single-quoted.

Is there something I can do to escape the colon character for the build process, or should I just remove these lines from the `Profile` files? We don't technically need them in there, but it'll help the CIS configuration not drift if the rule gets updated in future for other standards.

linux_os/guide/services/ssh/ssh_server/sshd_set_maxstartups/rule.yml
```diff
@@ -0,0 +1,33 @@
...
+    authentication connection attempts, use the rate limiting function of MaxStartups
+    to protect availability of sshd logins and prevent overwhelming the daemon.
+
+rationale: |-
+    The MaxStartups parameter specifies the maximum number of concurrent
+    unauthenticated connections to the SSH daemon. Additional connections will be
+    dropped until authentication succeeds or the LoginGraceTime expires for a
+    connection. The sshd default is 10. By default, this rule uses the
+    CIS-recommended value of 'MaxStartups 10:30:60', which adds 'random early drop'
+    functionality.
```
yuumasato, line 16: What value is expected for the parameter `MaxStartups` depends more on the variable than the rule itself.
Here you can mention the variable using following syntax: `{{{ xccdf_value("var_sshd_set_maxstartups") }}}`.
Example: https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/ssh/ssh_server/sshd_disable_compression/rule.yml#L12

This will render the rule nicely in the reports and guides with the actual value selected for the profile.
Also, I think this should be part of the description, rather then rationale.

rhel7/profiles/cis.profile
yuumasato: So the issue is that `:` is the delimiter used to designate control selections.
But this is a variable selection and we should be able to not mistake if for a control.
I filed #7073 to address this issue.

linux_os/guide/services/ssh/ssh_server/var_sshd_set_maxstartups.var
```diff
@@ -0,0 +1,14 @@
...
+description: 'Configure parameters for maximum concurrent unauthenticatd connections to the SSH daemon.'
+
+type: string
+
+operator: equals
+
+interactive: true
+
+options:
+    '10:30:60': '10:30:60'
```
yuumasato, line 14: You can keep the default entry here with the same value.

linux_os/guide/services/ssh/ssh_server/sshd_set_maxstartups/tests/correct_value.pass.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+#
+# profiles = xccdf_org.ssgproject.content_profile_ospp
+
+if grep -q "^MaxStartups" /etc/ssh/sshd_config; then
+	sed -i "s/^MaxStartups.*/UsePrivilegeSeparation 10:30:60/" /etc/ssh/sshd_config
```
yuumasato, line 6: Copy-pasta error, `UsePrivilegeSeparation`

linux_os/guide/services/ssh/ssh_server/sshd_set_maxstartups/tests/comment.fail.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+#
+# profiles = xccdf_org.ssgproject.content_profile_ospp
```
yuumasato, line 3: The profiles in these test scenarios should be cis. I get errors when trying the test suite.
` # profiles = xccdf_org.ssgproject.content_profile_cis`

By the way, at the moment the tests in the `tests/' directory are not run by our CI. We hope to set this up soon.
So these need to be run locally for now, you can check how to set the Test Suite here:
https://complianceascode.readthedocs.io/en/latest/tests/README.html#

yuumasato (changes requested):

linux_os/guide/services/ssh/ssh_server/sshd_set_maxstartups/ansible/shared.yml
```diff
@@ -0,0 +1,8 @@
+# platform = multi_platform_all,multi_platform_sle
+# reboot = false
+# strategy = restrict
+# complexity = low
+# disruption = low
+- (xccdf-var var_sshd_set_maxstartups)
+
+{{{ ansible_sshd_set(parameter="UsePrivilegeSeparation", value="{{ var_sshd_set_maxstartups }}") }}}
```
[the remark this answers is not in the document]
alexhaydock: Fixed - thank you :+1:

linux_os/guide/services/ssh/ssh_server/sshd_set_maxstartups/rule.yml
alexhaydock: This makes sense. I also think you're right and maybe the rationale and description fields I'd settled on here possibly work better the other way around, so I've swapped things a bit.

linux_os/guide/services/ssh/ssh_server/sshd_set_maxstartups/ansible/shared.yml
```diff
@@ -0,0 +1,8 @@
+# platform = multi_platform_all,multi_platform_sle
```
cipherboy, line 1: `multi_platform_sle` is extraneous -- it is implied by `_all`.

linux_os/guide/services/ssh/ssh_server/var_sshd_set_maxstartups.var
alexhaydock: Re-added, thanks.

rhel7/profiles/cis.profile
alexhaydock: Great. Thanks for doing that! We can keep this PR open until #7073 is merged and the CI jobs are able to run properly.

linux_os/guide/services/ssh/ssh_server/sshd_set_maxstartups/oval/shared.xml
```diff
@@ -0,0 +1,30 @@
...
+      </criteria>
+      <criterion comment="Check the MaxStartups value in /etc/ssh/sshd_config"
+      test_ref="test_sshd_set_maxstartups" />
+    </criteria>
+  </definition>
+  
+  {{{ oval_line_in_file_test(path='/etc/ssh/sshd_config', parameter='MaxStartups') }}}
+  {{{ oval_line_in_file_object(path='/etc/ssh/sshd_config', prefix_regex="^[ \\t]*(?i)", parameter='MaxStartups', separator_regex='(?-i)[ \\t]+') }}}
+  
+  <ind:textfilecontent54_state id="state_sshd_set_maxstartups" version="1">
```
cipherboy, line 25: Just a general comment... It makes this OVAL check much harder, but technically CIS is worded to allow inequality comparisons... It would require splitting the variable into three parts (using local variables and pattern matching) -- but would allow a stricter-than-necessary config to pass. However, it is hard to determine intent -- if the admin specifically set a value and the existing value was stricter, the rule would pass even though the admin might want to still remediate (thought they could always do this manually by generating a profile-based remediation without scanning).

Thoughts?

linux_os/guide/services/ssh/ssh_server/sshd_set_maxstartups/tests/correct_value.pass.sh
alexhaydock: Fixed - thanks.

linux_os/guide/services/ssh/ssh_server/sshd_set_maxstartups/rule.yml
```diff
@@ -0,0 +1,34 @@
...
+
+severity: low
+
+identifiers:
+    cce@rhel7: CCE-83610-6
+    cce@rhel8: CCE-83609-8
+
+references:
+    cis@rhel7: 5.3.21
+    cis@rhel8: 5.2.18
```
cipherboy, line 27:
```suggestion
    cis@rhel8: 5.2.18
    cis@ubuntu2004: 5.2.21
```

While you're here... :-)

linux_os/guide/services/ssh/ssh_server/sshd_set_maxstartups/tests/comment.fail.sh
alexhaydock: I've switched this over to the `cis` profile now. I'm going to get a Podman-based environment set up for testing like you suggest. Thanks for the link!

linux_os/guide/services/ssh/ssh_server/sshd_set_maxstartups/rule.yml
```diff
@@ -0,0 +1,34 @@
...
+title: 'Ensure SSH MaxStartups is configured'
+
+description: |-
+    The MaxStartups parameter specifies the maximum number of concurrent
+    unauthenticated connections to the SSH daemon. Additional connections will be
+    dropped until authentication succeeds or the LoginGraceTime expires for a
+    connection. To confgure MaxStartups, you should add or correct the following
+    line in the
+    <tt>/etc/ssh/sshd_config</tt> file:
+    <pre>MaxStartups {{{ xccdf_value("var_sshd_set_maxstartups") }}}</pre>
```
cipherboy, line 12: Technically, for CIS, it also includes this note:

```
> Note: Local site policy may be more restrictive.
```

But it seems like you're making this "equals" and not allowing "more restrictive" in the OVAL -- just something to discuss. What are your thoughts on the "or more restrictive" clause? Even if the OVAL doesn't, if someone was just reading the XCCDF it might be nice to mention it.

alexhaydock (commented):
@yuumasato Thanks for the helpful review and for spotting my embarrassing copy-paste fails. I'm getting the hang of it more with each PR so I do appreciate your help!

linux_os/guide/services/ssh/ssh_server/sshd_set_maxstartups/rule.yml
```diff
@@ -0,0 +1,34 @@
...
+
+rationale: |-
+    To protect a system from denial of service due to a large number of pending
+    authentication connection attempts, use the rate limiting function of MaxStartups
+    to protect availability of sshd logins and prevent overwhelming the daemon.
+
+severity: low
```
cipherboy, line 19: Since both `set_max_auth_tries` and `set_max_sessions` are medium, would you make this medium as well? Or is there justification for low that I'm overlooking?

linux_os/guide/services/ssh/ssh_server/var_sshd_set_maxstartups.var
```diff
@@ -0,0 +1,15 @@
...
+description: 'Configure parameters for maximum concurrent unauthenticatd connections to the SSH daemon.'
+
+type: string
+
+operator: equals
+
+interactive: true
+
+options:
+    default: "10:30:60"
```
cipherboy, line 14: The sshd default is actually `10:30:100` -- perhaps leave `10:30:60` as an option and set it explicitly in the profile (like you already do)?

cipherboy (commented):
Just some thoughts :) Take them or leave them, I don't consider them important enough to hold up merging if others are in favor of merging.

linux_os/guide/services/ssh/ssh_server/sshd_set_maxstartups/rule.yml
alexhaydock: This is a good point. I've added wording to this effect into the rule and will investigate strengthening the OVAL check also.

linux_os/guide/services/ssh/ssh_server/sshd_set_maxstartups/rule.yml
alexhaydock: It was a struggle to decide, honestly. I based this one on the DISA Category guidelines [in the developer docs](https://complianceascode.readthedocs.io/en/latest/manual/developer/06_contributing_with_content.html):

The guidelines suggest this for **medium**:
> Any vulnerability, the exploitation of which has a potential to result in loss of Confidentiality, Availability, or Integrity.

And this for **low:**
> Any vulnerability, the existence of which degrades measures to protect against loss of Confidentiality, Availability, or Integrity.

Not having the more restrictive `MaxStartups` setting in place doesn't necessarily impact Confidentiality, Availability, or Integrity directly, so it seemed to fit more into the `low` category to me. But since it's quite close between the two, I think it might be sensible to do what you've done and assess the severity with reference to the values we're already using for other similar `sshd` rules.

I've updated it to medium to keep it in line with the other rules, but I think either is probably valid.

linux_os/guide/services/ssh/ssh_server/sshd_set_maxstartups/oval/shared.xml
alexhaydock: I agree with the idea that it would be sensible to include a test that allows stricter-than-necessary rules to pass the test.

I do also see what you're suggesting though, that an admin might want to end a remediation run with the CIS-recommended value of `10:30:60` regardless of whether a more strict value is already in place. But reading the wording of the CIS benchmark, CIS does not appear to be too opinionated about the specific value here so I think maybe it's fine for a validation run to consider a stricter value as a pass.

I'll have to look closer into splitting the variable and validating the separate segments to allow the stricter configs to pass the rule. I'll have to look through other tests and see whether I can find any which already do something similar.

alexhaydock (commented):
Thanks for the comments and helpful review, @cipherboy - much appreciated! :smile: 

I think I've addressed them all here apart from the more specific OVAL check which will need some additional work.

linux_os/guide/services/ssh/ssh_server/sshd_set_maxstartups/rule.yml
cipherboy: Isn't `MaxStartups` mean to prevent system overload, so it goes to directly address availability? :) But thanks!

linux_os/guide/services/ssh/ssh_server/sshd_set_maxstartups/oval/shared.xml
cipherboy: Well, I don't have the answer to the local variable part, but the rest of it would look something like:


```xml
<def-group>
  <definition class="compliance" id="{{{ rule_id }}}" version="1">
    {{{ oval_metadata("The MaxStartups parameter specifies the maximum number of concurrent unauthenticated connections to the SSH daemon.") }}}
    <criteria operator="OR">
      <extend_definition definition_ref="package_openssh-server_removed" negate="true" comment="SSH daemon is not installed" applicability_check="true" />
      <criteria operator="AND">
        <criterion test_ref="tst_maxstartups_start_parameter" comment="SSH MaxStartups start parameter is less than or equal to 10" />
        <criterion test_ref="tst_maxstartups_rate_parameter" comment="SSH MaxStartups rate parameter is greater than or equal to 30" />
        <criterion test_ref="tst_maxstartups_full_parameter" comment="SSH MaxStartups full parameter is less than or equal to 100" />
      </criteria>
    </criteria>
  </definition>
  <ind:textfilecontent54_object id="obj_sshd_config_maxstartups_first_parameter" version="1">
    <ind:filepath operation="equals">/etc/ssh/sshd_config</ind:filepath>
    <ind:pattern operation="pattern match" datatype="string">(?i)^\s*MaxStartups\s+(\d+):\d+:\d+\s*$</ind:pattern>
    <ind:instance operation="greater than or equal" datatype="int">1</ind:instance>
  <ind:textfilecontent54_object id="obj_sshd_config_maxstartups_second_parameter" version="1">
    <ind:filepath operation="equals">/etc/ssh/sshd_config</ind:filepath>
    <ind:pattern operation="pattern match" datatype="string">(?i)^\s*MaxStartups\s+\d+:(\d+):\d+\s*$</ind:pattern>
    <ind:instance operation="greater than or equal" datatype="int">1</ind:instance>
  <ind:textfilecontent54_object id="obj_sshd_config_maxstartups_third_parameter" version="1">
    <ind:filepath operation="equals">/etc/ssh/sshd_config</ind:filepath>
    <ind:pattern operation="pattern match" datatype="string">(?i)^\s*MaxStartups\s+\d+:\d+:(\d+)\s*$</ind:pattern>
    <ind:instance operation="greater than or equal" datatype="int">1</ind:instance>
  <ind:textfilecontent54_state id="ste_sshd_config_start_parameter_valid" version="1">
    <ind:subexpression datatype="int" operation="less than or equal">10</ind:subexpression>
  <ind:textfilecontent54_state id="ste_sshd_config_rate_parameter_valid" version="1">
    <ind:subexpression datatype="int" operation="greater than or equal">30</ind:subexpression>
  <ind:textfilecontent54_state id="ste_sshd_config_full_parameter_valid" version="1">
    <ind:subexpression datatype="int" operation="less than or equal">100</ind:subexpression>
  <ind:textfilecontent54_test check="all" check_existence="only_one_exists" id="tst_maxstartups_start_parameter" version="1" comment="SSH MaxStartups start parameter is less than or equal to 10">
    <ind:object object_ref="obj_sshd_config_maxstartups_first_parameter" />
    <ind:state state_ref="ste_sshd_config_start_parameter_valid" />
  </ind:textfilecontent54_test>
  <ind:textfilecontent54_test check="all" check_existence="only_one_exists" id="tst_maxstartups_rate_parameter" version="1" comment="SSH MaxStartups rate parameter is greater than or equal to 30">
    <ind:object object_ref="obj_sshd_config_maxstartups_second_parameter" />
    <ind:state state_ref="ste_sshd_config_rate_parameter_valid" />
  </ind:textfilecontent54_test>
  <ind:textfilecontent54_test check="all" check_existence="only_one_exists" id="tst_maxstartups_full_parameter" version="1" comment="SSH MaxStartups full parameter is less than or equal to 100">
    <ind:object object_ref="obj_sshd_config_maxstartups_third_parameter" />
    <ind:state state_ref="ste_sshd_config_full_parameter_valid" />
  </ind:textfilecontent54_test>
</def-group>
```

I think you're looking for [something like this](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/accounts/accounts-restrictions/root_logins/no_shelllogin_for_systemaccounts/oval/shared.xml#L143-L150) for the variable parts. :-) 

(Above OVAL shared with permission from internal contributors, feel free to use all or part of it here in this CaC PR). 

If it is more than you want to bite off, just LMK and I can open a second PR after this merges.

alexhaydock: I see #7073 has been merged now so technically this PR could be as well I think, but I still plan on tackling this one when I get a chance. I like a challenge...

alexhaydock: @cipherboy I've tried to integrate this now. Hopefully I understood the local variable part properly. The regex there should match the value assigned to the last instance of `MaxStartups` found in the `/etc/ssh/sshd_config` file.

Your test lines above seemed to make sense as-is, so I've incorporated that mostly verbatim into the OVAL file.

linux_os/guide/services/ssh/ssh_server/sshd_set_maxstartups/oval/shared.xml
```diff
@@ -0,0 +1,71 @@
...
+  <ind:textfilecontent54_test check="all" check_existence="only_one_exists" id="tst_maxstartups_rate_parameter" version="1" comment="SSH MaxStartups rate parameter is greater than or equal to 30">
+    <ind:object object_ref="obj_sshd_config_maxstartups_second_parameter" />
+    <ind:state state_ref="ste_sshd_config_rate_parameter_valid" />
+  </ind:textfilecontent54_test>
+  <ind:textfilecontent54_test check="all" check_existence="only_one_exists" id="tst_maxstartups_full_parameter" version="1" comment="SSH MaxStartups full parameter is less than or equal to 100">
+    <ind:object object_ref="obj_sshd_config_maxstartups_third_parameter" />
+    <ind:state state_ref="ste_sshd_config_full_parameter_valid" />
+  </ind:textfilecontent54_test>
+
+  <local_variable id="variable_sshd_config_maxstartups" datatype="int"
```
cipherboy, line 64: @alexhaydock is this done? I see this variable sitting here but I don't actually see it used. :-)

cipherboy: I think I was thinking about splitting this into three portions (start/rate/full) and then using it in the `ste_sshd_config_{start,rate,full}_parameter_valid` state checks using the `var_ref` parameter. 

Here, for the `local_variable` I'd use a `variable_component` rather than an `object_component` and simplifying the regex based off of the actual value of the `var_sshd_set_maxstartups` variable.

yuumasato: This is missing an `external_variable` definition to actually "pull in" the value of the XCCDF variable.
E.g.: https://github.com/ComplianceAsCode/content/blob/607b04549c5c542042af19fda6b5df6c7a87520d/linux_os/guide/system/selinux/selinux_state/oval/shared.xml#L25

> I think I was thinking about splitting this into three portions (start/rate/full) and then using it in the `ste_sshd_config_{start,rate,full}_parameter_valid` state checks using the `var_ref` parameter.

Once the external variable is available, I'd split the variable into three `local_variables`, with regexes similar to what you do for `obj_sshd_config_maxstartups_{first,second,third}_parameter`.
And then update `ste_sshd_config_{start,rate,full}_parameter_valid` to use a `var_ref` to the `local_variable`s defined above.

> If you don't want to, I think this is OK to merge as-is and I can add tests for this case and update the test to it. Was there a design decision towards this route that I'm missing? :)

Well, at the moment the OVAL check is not using the variable, which is mentioned in the rule description and used in the remediations.
So for consistency, I'd rather update the OVAL to use the variable or remove it from rule and remediation.

cipherboy: Yep, exactly what I was thinking about doing. :) If @alexhaydock doesn't want to get to it, I can push one. :-)

cipherboy:
@alexhaydock Hey! If you don't mind rebasing to clean up the conflict, we'd be happy to merge this and then I'll tackle the last few OVAL changes. :)

alexhaydock:
> @alexhaydock Hope all is well! If you don't mind rebasing to clean up the conflict, we'd be happy to merge this and then I'll tackle the last few OVAL changes. :)

I had hoped to be less busy recently and did want to try and tackle your (very helpful, thank you!) suggestions head-on, but I'll be a little busy for the coming week so we might as well get these merged. I've rebased the PR to account for the merge conflict caused by #7108 and I'm planning on refactoring #6976 as soon as I get the chance, to perform the same kind of CIS splitting for RHEL 8.

linux_os/guide/services/ssh/ssh_server/sshd_set_maxstartups/oval/shared.xml
```diff
@@ -0,0 +1,71 @@
...
+  comment="Value of last MaxStartups from /etc/ssh/sshd_config" version="1">
+    <regex_capture pattern="^\s*MaxStartups\s+(\d+:\d+:\d+\s*$(?!\s*MaxStartups\s+(\d+:\d+:\d+\s*$)))">
+      <object_component item_field="subexpression"
+      object_ref="object_last_maxstartups_from_sshd_config" />
```
yuumasato, line 68: We are not be able to merge this as is, this `object_ref` points to a non-existing object.

@alexhaydock Another alternative to merging this PR is for @cipherboy to push changes directly here, would you be okay with that?

alexhaydock: Hi @yuumasato - I've attempted to resolve the merge conflict here by removing the whole local variable. If this still isn't enough to let merging happen. If this still isn't enough then I'm happy for @cipherboy to push changes directly. Thanks again for your help :smile:

yuumasato: Hi, sorry I was out yesterday.

The conflict with the CCEs file is solved. But some other rule was updated and took the same CCEs that were assigned to this rule. I could push a quick CCE update commit and merge this...

alexhaydock: > Hi, sorry I was out yesterday.
> 
> The conflict with the CCEs file is solved. But some other rule was updated and took the same CCEs that were assigned to this rule. I could push a quick CCE update commit and merge this...

Ah I'd missed that when resolving the conflict. Feel free to update the CCE list and merge if that is the easiest fix.

yuumasato: @alexhaydock Sorry, taking a second look at the branch history and 8aa40ccb71a79876c97ba90427740c8728e3a1b7, there is something strange with the rebase. A lot of CEEs are showing as removed and that should not be the change set done by this PR.

Could you please try rebasing on top of latest master? And while you are at it, assign new CCEs to the rule?
(Sorry for the trouble, I haven't tried recently, but I think only original authors can do force-pushes to PRs)

cipherboy: @yuumasato I'll take it over at some point. I just haven't had time yet. It will probably be next week after the holidays. I can clean it up then. :-)

alexhaydock: @yuumasato I think I fixed it. I've assigned new CCEs and managed to rebase on upstream properly. Not quite sure what went on with the resolution of that merge conflict earlier.

yuumasato: I see what is strange, I think you are only updating the `cce-redhat-avail.txt`, the diff against master is ok, but there are commits removing a bunch of CEEs that were consumed in the master branch.

And now I see a merge from master into your branch, I'll try to clean the history.

alexhaydock: @yuumasato Would merging as a squash commit work if the diff against master is okay? Or have I managed to make too much of a mess of it? :facepalm:

yuumasato: I got into a meeting right after I sent the message, :laughing: 

@alexhaydock I removed the merge commits and was able to force-push into your branch, check if it seems ok.

I just dropped the changes to `controls\cis_rhel7.yml`, 299380fdadd06d3d9c2688a23001bd688adf92e9 to avoid conflicts with #7134 and subsequent PRs @vojtapolasek has.

alexhaydock: @yuumasato No problem 😆 and thanks for sorting this out. I think just dropping this one from the in-progress RHEL7 control file is sensible for now. I know about the work @vojtapolasek has been doing there.

yuumasato: @alexhaydock Cool, tests passed! I'm merging this one. :)

yuumasato (approved):
@alexhaydock Thank you for the rule addition. :)

Approving as is (with check hard coded to `10:30:100`) . As discussed with @cipherboy, he plans to update the rule to use the variable. Ideally this should be done before the next release.

PR #7073, opened 2021-06-02
Don't mistake ':' in variable selector for control

yuumasato:
#### Description:

- Check if the item is not a variable selection before assuming it is a control selection.

#### Rationale:

- This allows variable to use ":" in the selector.

### DIFF
--- a/ssg/build_yaml.py
+++ b/ssg/build_yaml.py
@@ -514,7 +514,8 @@ class ProfileWithInlinePolicies(ResolvableProfile):
         self.controls_by_policy = defaultdict(list)
 
     def apply_selection(self, item):
-        if ":" in item:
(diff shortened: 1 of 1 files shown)


cipherboy (approved):
I agree with pep8speaks but otherwise LGTM :)

yuumasato:
Fixed, :)

JAORMX (approved):
This LGTM. The ocp4 e2e failure is unrelated and I'm investigating it. This is fine to merge IMO.

cipherboy:
Per confirmation from @JAORMX and seeing as above test fails somewhat frequently, likely unrelated. :-) Merged! Thanks @yuumasato for the fix!

PR #7076, opened 2021-06-03
Fix import of open_environment method

yuumasato:
#### Description:

- The function was relocated from `ssg.yaml` to `ssg.environment`.

#### Rationale:

- `utils/render-policy.py` got left out of the relocation and was broken
- Follow up from #6754


vojtapolasek (approved):
It works now, thanks for the fix.

PR #7077, opened 2021-06-03
OCP4: Check that either the OAuth server or any of the clients have inactivity timeout set

jhrozek:
The patch adds two rules (auth_inactivity_timeout and
oauthclient_inactivity_timeout) that check the inactivity timeout for the
OAuth server or the clients respectively. These rules are not included
in any of the profiles, but are referenced from the hand-crafted OVAL of
oauth_or_oauthclient_inactivity_timeout.

The oauth_or_oauthclient_inactivity_timeout rule is included from the
OCP4 moderate profile.

At the moment, the remediation uses a hardcoded timeout of 10 minutes.


applications/openshift/authentication/oauth_or_oauthclient_inactivity_timeout/kubernetes/shared.yml
```diff
@@ -0,0 +1,8 @@
+# platform = multi_platform_ocp
+apiVersion: config.openshift.io/v1
+kind: OAuth
+metadata:
+  name: cluster
+spec:
+  tokenConfig:
+    accessTokenInactivityTimeout: 600s
```
jhrozek, line 8: Even if the kubernetes remediations can't be configured, I thought it was better to have an automated remediation with some reasonable default. Let me know if you think this would cause more problems for the customers and I can remove it

applications/openshift/authentication/oauth_inactivity_timeout/rule.yml
```diff
@@ -0,0 +1,67 @@
...
+
+template:
+    name: yamlfile_value
+    vars:
+        ocp_data: 'true'
+        filepath: /apis/config.openshift.io/v1/oauths/cluster
+        yamlpath: ".spec.tokenConfig.accessTokenInactivityTimeout"
+        check_existence: "only_one_exists"
+        values:
+            - value: ".*"
```
jhrozek, line 66: At the moment, the rule checks for ANY timeout. I was going back and forth on whether we should already introduce a variable and check for its value to make sure the timeout is not higher than some organizational setting. But 1) I guess the timeouts can be different for every client which would make the check difficult and 2) since the remediation is static anyway, setting the variable to a custom value might make the check fail because the remediation would set it to something else than the check looks for.
Thoughts?

applications/openshift/authentication/oauth_or_oauthclient_inactivity_timeout/kubernetes/shared.yml
JAORMX: Nope, I think this is a good approach. Can the Compliance Operator already remediate this? didn't know we had already set the RBAC permissions for this.

applications/openshift/authentication/oauth_inactivity_timeout/rule.yml
JAORMX: I'm fine with this, but ultimately it should be a variable. If you plan to leave this as it is, let's add a TODO comment.

JAORMX:
helpers.go:707: E2E-FAILURE: The expected result for the oauth_or_oauthclient_inactivity_timeout rule didn't match. Expected 'PASS', Got 'FAIL'

applications/openshift/authentication/oauth_inactivity_timeout/rule.yml
jhrozek: Hmm, the more I think about it the more I think the variable would be better. When we are able to template the remediations, it would be easier to see which must be templated if we just look for the existing variables. I'll change the PR.

applications/openshift/authentication/oauth_or_oauthclient_inactivity_timeout/kubernetes/shared.yml
jhrozek: Oh you're right. When I was looking at what RBAC perms we need for the fedramp work, I did add read perms for the OAuth objects, but forgot that we also need write perms for the remediations.

I suggest we split the PR into two, merge the check without remediation and when a new CO release is ready (I guess we might find more RBAC issues), then add the remediation.

jhrozek:
It looks like CI is passing (great!) but please don't merge the PR just yet, I'll introduce the variable first, I forgot to do that earlier

jhrozek:
> It looks like CI is passing (great!) but please don't merge the PR just yet, I'll introduce the variable first, I forgot to do that earlier

Done two changes: the rule now fetches both objects and a variable is used for the global `oauth/cluster` object. I did not ad a variable for the oauthclients because the clients can be configured each with a different timeout (that's sort of the point) so a single variable wouldn't work very well.

JAORMX (approved):
LGTM

jhrozek:
Oops, the remediation was wrong in the previous iteration, let's try again.

JAORMX:
/retest

Route53 issues

JAORMX:
/retest
AWS quota issue

jhrozek:
/retest
going to watch this PR live, something fishy is going on

jhrozek:
/test e2e-aws-ocp4-moderate
aws is really flaky

PR #7081, opened 2021-06-03
Mount option template updates

cipherboy:
#### Description:

This pull request does three things:

1. Aligns Bash remediation with other languages, checking for `no` rather than `yes` for consistency.
2. Fixes a bug in the OVAL where by `MOUNT_HAS_TO_EXIST=no` wasn't taken into account.
3. Per our discussion on the community call yesterday, allows this OVAL template to be used in separate rules -- one where `MOUNT_HAS_TO_EXIST=no` and one where `MOUNT_HAS_TO_EXIST=yes` --- note that, in that case, previously the builld would fail as interior tests had duplicate values.

This is a fix brought upstream as a result of our internal CaC port.


shared/templates/mount_option/bash.template
```diff
@@ -6,8 +6,8 @@
 include_mount_options_functions
 
 function perform_remediation {
-	# test "$mount_has_to_exist" = 'yes'
-	if test "{{{ MOUNT_HAS_TO_EXIST }}}" = 'yes'; then
+	# test "$mount_has_to_exist" = 'no'
+	if test "{{{ MOUNT_HAS_TO_EXIST }}}" = 'no'; then
```
jan-cerny, line 10: you should also flip the condition, right?

cipherboy: Ah, d'oh. Yes.

cipherboy:
@jan-cerny Good catch, thanks! Updated :)

jan-cerny:
There are some test scenarios that started to fail in rule mount_option_dev_shm_nodev separate.fail.sh and no_partition.fail.sh. I think that with this fix the scenarios need to be changed to .pass.sh. @mildas What do you think?

mildas:
@jan-cerny Yes, I agree.
Now, I see that `mount_option_dev_shm_*` rules has `mount_has_to_exist: 'no'` so it is expected that the rules will pass when the partition doesn't exist. `separate.fail.sh` and `no_partition.fail.sh` test scenarios have to be `.pass.sh`

cipherboy:
Sorry about the delay. This PR has been rebased and the test cases updated @jan-cerny @mildas.

jan-cerny (approved):
great

PR #7082, opened 2021-06-03
Add option_regex_suffix to sudo_defaults_option template

cipherboy:
#### Description

In the course of porting Ubuntu CIS tooling to CaC, we too used the
`sudo_defaults_option` template for the logfile rule. However, we found a
shortcoming in the existing template: `\w` matches word characters and
not `/` that many of the log file paths will use. Introduce an optional
template variable, `option_regex_suffix`, to control the regex portion
_after_ the option name in this template. This will allow us to specify
a different value in the `sudo_custom_logfile` rule.

---

I was wanting to review the pull request prior to it merging but I forgot to assign myself as a reviewer and it got merged with this bug :D No matter. 

/cc @alexhaydock

### DIFF
--- a/docs/manual/developer/06_contributing_with_content.md
+++ b/docs/manual/developer/06_contributing_with_content.md
@@ -1773,7 +1773,7 @@ The remediations add the `Defaults` option  to `/etc/sudoers` file.
 -   Parameters:
 
     - **option** - name of sudo `Defaults` option to enable.
-    - **option_regex_suffix** - suffix to the pattern-match to use after **option**; defaults to `=(\w+)`.
+    - **option_regex_suffix** - suffix to the pattern-match to use after **option**; defaults to `=(\w+)\b`.
     - **parameter_variable** - name of the XCCDF variable to get the value for the option parameter.\
       (optional, if not set the check and remediation won't use parameters)
     - **default_is_enabled** -  set to `"true"` if the option is enabled by default for the product.
--- a/docs/manual/developer/06_contributing_with_content.md
+++ b/docs/manual/developer/06_contributing_with_content.md
@@ -1773,6 +1773,7 @@ The remediations add the `Defaults` option  to `/etc/sudoers` file.
 -   Parameters:
 
     - **option** - name of sudo `Defaults` option to enable.
+    - **option_regex_suffix** - suffix to the pattern-match to use after **option**; defaults to `=(\w+)`.
     - **parameter_variable** - name of the XCCDF variable to get the value for the option parameter.\
       (optional, if not set the check and remediation won't use parameters)
     - **default_is_enabled** -  set to `"true"` if the option is enabled by default for the product.
--- a/linux_os/guide/system/software/sudo/sudo_custom_logfile/rule.yml
+++ b/linux_os/guide/system/software/sudo/sudo_custom_logfile/rule.yml
@@ -31,4 +31,5 @@ template:
     name: sudo_defaults_option
     vars:
         option: logfile
+        option_regex_suffix: '=(.*)'
         variable_name: var_sudo_logfile
--- a/linux_os/guide/system/software/sudo/sudo_custom_logfile/rule.yml
+++ b/linux_os/guide/system/software/sudo/sudo_custom_logfile/rule.yml
@@ -31,5 +31,41 @@ template:
     name: sudo_defaults_option
     vars:
         option: logfile
-        option_regex_suffix: '=(.*)'
+        # Description of option_regex_suffix
+        # Outer capture group for the value (which we need to compare against an
+        # XCCDF variable).
+        #
+        # Inside it is an OR of two paths:
+        #
+        #  - Either we have a quoted value, or
+        #  - We don't have a quoted value.
+        #
+        # In the quoted path, we match the start and end quote (and therefore, the
+        # user running against this rule MUST specify quotes in the variable value
+        # if necessary!). Then we match (in between these quotes) any (potentially
+        # empty) group of character that:
+        #
+        #  - Is an escaped double quote,
+        #  - Is an escaped backslash,
+        #  - Or isn't one of those characters.
+        #
+        # Finally, we match on \B: since we know " is not a word character,
+        # it'll only match if the following character is also not a word
+        # character. This ensures we don't have a string such as "quoted"d, which
+        # would (presumably) be invalid in a sudoers entry.
+        #
+        # In the non-quoted path, we strictly match the value. However, we have a
+        # few more escaped characters to deal with. Thus we match any (potentially
+        # empty) group of characters that:
+        #
+        #  - Is an escaped comma,
+        #  - Is an escaped double quote,
+        #  - Is an escaped space (per `man sudoers`, this needs escaping without
+        #    double quotes),
+        #  - Is an escaped backslash,
+        #  - Or isn't one of those characters.
+        #
+        # Finally, we check for regular word boundary (with \b), ensuring the
+        # next character isn't yet another word character.
+        option_regex_suffix: '=("(?:\\"|\\\\|[^"\\\n])*"\B|[^"](?:(?:\\,|\\"|\\ |\\\\|[^", \\\n])*)\b)'
         variable_name: var_sudo_logfile
--- a/linux_os/guide/system/software/sudo/sudo_custom_logfile/tests/logfile_enabled.pass.sh
+++ b/linux_os/guide/system/software/sudo/sudo_custom_logfile/tests/logfile_enabled.pass.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+# platform = multi_platform_all
+
+echo "Defaults logfile=/var/log/sudo.log" >> /etc/sudoers
--- a/linux_os/guide/system/software/sudo/sudo_custom_logfile/tests/logfile_absent.fail.sh
+++ b/linux_os/guide/system/software/sudo/sudo_custom_logfile/tests/logfile_absent.fail.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+# platform = multi_platform_all
+
+# Code taken from macro bash_sudo_remove_config()
+for f in $( ls /etc/sudoers /etc/sudoers.d/* 2> /dev/null ) ; do
+  matching_list=$(grep -P '^(?!#).*[\s]+logfile.*$' $f | uniq )
+  if ! test -z "$matching_list"; then
+    while IFS= read -r entry; do
+      # comment out "{{{ parameter }}}" matches to preserve user data
+      sed -i "s/^${entry}$/# &/g" $f
+    done <<< "$matching_list"
+
+    /usr/sbin/visudo -cf $f &> /dev/null || echo "Fail to validate $f with visudo"
+  fi
+done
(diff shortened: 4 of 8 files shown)


alexhaydock:
Thanks for spotting this one and for the CC. 👍 I'll keep an eye out for similar issues when working with vars in future.

cipherboy:
No worries :-) I just pushed a set of test cases that should help verify it btw.

linux_os/guide/system/software/sudo/sudo_custom_logfile/tests/logfile_enabled.pass.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+# platform = multi_platform_all
+
+echo "Defaults logfile=/var/log/sudo.log" >> /etc/sudoers
```
yuumasato, line 4: Should we also consider cases when multiple options are defined in a single `Defaults`?
The current `option_regex_suffix` will not in these cases. I think it will match the following options as part of the path.
E.g: `Defaults logfile=/var/log/sudo.log,noexec`

linux_os/guide/system/software/sudo/sudo_custom_logfile/rule.yml
```diff
@@ -31,4 +31,5 @@ template:
     name: sudo_defaults_option
     vars:
         option: logfile
+        option_regex_suffix: '=(.*)'
```
yuumasato, line 34: And could you add this new parameter in the template documentation?

yuumasato (commented):
@cipherboy Nice catch and sorry for the early merge.

This looks good, my only concern is with regards to multiple options defined in a single line.

linux_os/guide/system/software/sudo/sudo_custom_logfile/rule.yml
yuumasato: Most future proof would be to craft a regex that matches the logfile path without going further into other options.

linux_os/guide/system/software/sudo/sudo_custom_logfile/tests/logfile_enabled.pass.sh
cipherboy: Do our current `sudo` rules support that? My understanding is no, because we match on `Defaults <option>` -- so if you have an entry like `Defaults logfile=/var/log/sudo.log,noexec` -- it'll only match for `logfile` and any rule for `noexec` will fail because `Defaults noexec` is missing from the sudo rules?

Regarding compliance... If there isn't a `noexec` rule and it was added by the admin, I'm not sure how I feel about that. CIS doesn't appear to mention that scenario and you do get into dangerous territory -- what if your value is `Defaults logfile=/var/log/noexec/sudo.log,something`? Would you match on `Defaults noexec` and `Defaults.*,noexec`? Is there a way to escape a comma in a parameter value? &c. I'd prefer to stick with simple matches, IMO. I dunno. 

My 2c. but if we do want to support it, I think it is outside the scope of this particular PR. Thoughts?

linux_os/guide/system/software/sudo/sudo_custom_logfile/rule.yml
cipherboy: Added to the docs :-)

See commentary above about the general case. Would the preferred solution be `=([^,]*)` or do you have another regex in mind?

linux_os/guide/system/software/sudo/sudo_custom_logfile/tests/logfile_enabled.pass.sh
yuumasato: > Do our current `sudo` rules support that? My understanding is no, because we match on `Defaults <option>` -- so if you have an entry like `Defaults logfile=/var/log/sudo.log,noexec` -- it'll only match for `logfile` and any rule for `noexec` will fail because `Defaults noexec` is missing from the sudo rules?

They do work with multiple options. When `OPTION_REGEX` is `\w+`, it doesn't match the commas.
And the OVAL accounts for boundaries too, which should help ensure it is not taking the whole line:
https://github.com/ComplianceAsCode/content/blob/master/shared/templates/sudo_defaults_option/oval.template#L16

This rule has tests with multiple values per line:
https://github.com/ComplianceAsCode/content/tree/master/linux_os/guide/system/software/sudo/sudo_add_umask/tests

> My 2c. but if we do want to support it, I think it is outside the scope of this particular PR. Thoughts?

Well, it could be outside of scope of this PR.

cipherboy: Ah, d'oh, I missed the `.*` :D 

I agree, I'll make changes :-)

linux_os/guide/system/software/sudo/sudo_custom_logfile/rule.yml
yuumasato: I think lazy solution could be: `=([\w\/.]+)`. But it should work if the regex matches any character allowed in a path.
Hmm, can a filename have a comma?

yuumasato: Commas are allowed.
But then, if a logfile path can have commas, I don't know how sudo itself would parse the config correctly.
Unless you could wrap the path in quotes...

cipherboy: > Values may be enclosed in double quotes ("") when they contain multiple words.  Special characters may be escaped with a backslash (‘\’).

It is underspecified as to whether or not quotes remove the need for `\` backslash escpaing. I dunno. :-)

cipherboy: Grab some :popcorn:, what do you think of:

> ```
> =("(?:(?:\\"|[^"])*)*"|[^"](?:(?:\\,|\\"|[^",])*))
> ```

Outer capture group for the value.

Inside its an OR:

 - Either we have quotes.
 - Or we don't have quotes. 

In the quote path, we match the start and end quote (admin must specify in the XCCDF variable). Then we match any character group that's either an escaped double quote or isn't a double quote. Repeat until we have our value.

Same for the non-double-quote path: we match any character group that's either an escaped double-quote, an escaped comma, or any non-comma non-double-quote character. Repeat until we have our value. 

---

Discussed this with @monoidmusician and he suggested something along the lines of:

> ```
> (([^",]|\\,)*|"([^"]|\\")*")*
> ```

But I had issues getting that to work (it broke on escaped commas) and came up with the above.

WDYT?

cipherboy: Take another bite of :popcorn:

> ```
> =("(?:(?:\\"|[^"])*)*"|[^"](?:(?:\\,|\\"|\\ |[^", ])*))
> ```

To the non-quote path I:

 - Added space as an escapable-character
 - Added space to the non-match (for standalone space) group at the end.

Thanks @MonoidMusician for the catch.

cipherboy: Ooops... Still broken. I think this one works better:

> ```
> =("(?:\\"|\\\\|[^"\\\n])*"\B|[^"](?:(?:\\,|\\"|\\ |\\\\|[^", \\\n])*)\b)
> ```

Note that this will require pulling the `\b` out of the [OVAL pattern](https://github.com/ComplianceAsCode/content/blob/master/shared/templates/sudo_defaults_option/oval.template#L16) and putting it into the default `option_regex_suffix`. 

<details>
<summary>
Test cases:
</summary>


```
Defaults logfile="/var/this_path"
Defaults logfile="/var/this path"
Defaults logfile="/var/this_path",noexec
Defaults logfile="/var/this path",noexec
Defaults logfile=/var/log,noexec
Defaults logfile=/var/log\,noexec,noexec
Defaults logfile=/var/log/\ noexec,noexec
Defaults logfile=/var/log/insane\"quotes
Defaults logfile=/var/log/insane\"quotes,noexec
Defaults logfile=/var/log no quotes
Defaults logfile=/var/log no quotes,noexec
Defaults logfile=/var/log\ with\ escaped\ spaces
Defaults logfile=/var/log\ with\ escaped\ spaces,noexec
Defaults noshm,logfile="/var/this_path"
Defaults noshm,logfile="/var/this path"
Defaults noshm,logfile="/var/this_path",noexec
Defaults noshm,logfile="/var/this path",noexec
Defaults noshm,logfile=/var/log,noexec
Defaults noshm,logfile=/var/log\,noexec,noexec
Defaults noshm,logfile=/var/log/\ noexec,noexec
Defaults noshm,logfile=/var/log/insane\"quotes
Defaults noshm,logfile=/var/log/insane\"quotes,noexec
Defaults noshm,logfile=/var/log no quotes
Defaults noshm,logfile=/var/log no quotes,noexec
Defaults noshm,logfile=/var/log\ with\ escaped\ spaces
Defaults noshm,logfile=/var/log\ with\ escaped\ spaces,noexec
```

</details>

richardmaciel-canonical: Regexp reviewed. Looks solid.

yuumasato: Love the test set.
Any chance of adding the explanation as comment in the rule?

> Note that this will require pulling the `\b` out of the [OVAL pattern](https://github.com/ComplianceAsCode/content/blob/master/shared/templates/sudo_defaults_option/oval.template#L16) and putting it into the default `option_regex_suffix`.

Ok, this should not be a problem if incorporated into the default `option_regex_suffix`, right?

> Take another bite of popcorn

The emoji looks more like beer to me...  Was thinking who bites beer?

cipherboy: > Ok, this should not be a problem if incorporated into the default `option_regex_suffix`, right?

Yep, that's exactly what I did :-) Just something to keep in mind if anyone else wishes to do something similar. 

@yuumasato Did you want me to somehow add the test set to the SSGTS? Or could I leave it out? Wouldn't I effective need a ton of different profiles to set the variable value differently? Just curious how that'd work. Or does it suffice to leave the test suite here/in the commit?

yuumasato: > Did you want me to somehow add the test set to the SSGTS? Or could I leave it out? 

Thinking about, it would be good to have at least one test scenario, for sanity checking purposes.

> Wouldn't I effective need a ton of different profiles to set the variable value differently? Just curious how that'd work.

Not profiles, but individual test scenarios overwriting the variable with `# variable` Like this:
https://github.com/ComplianceAsCode/content/blob/95d892d7d058698c945115042f4901f8d5496215/linux_os/guide/system/software/sudo/sudo_add_umask/tests/0027_var_and_default_state.fail.sh#L3

> Or does it suffice to leave the test suite here/in the commit?

The bulk of tests can be in the commit.

docs/manual/developer/06_contributing_with_content.md
```diff
@@ -1773,6 +1773,7 @@ The remediations add the `Defaults` option  to `/etc/sudoers` file.
 -   Parameters:
 
     - **option** - name of sudo `Defaults` option to enable.
+    - **option_regex_suffix** - suffix to the pattern-match to use after **option**; defaults to `=(\w+)`.
```
yuumasato, line 1776: The `\b' is missing in the default `option_regex_suffix`.

cipherboy: Ah, d'oh thanks for the catch :)

yuumasato:
There was some unrelated weird failure in Jenkins.
@openscap-ci test this please

yuumasato (approved):
Nice work! :)

cipherboy:
Thanks @yuumasato!

PR #7083, opened 2021-06-03
Add some more tests

cipherboy:
#### Description:

This pull requests adds some test cases into CaC for some rules which already exist. Let me know if any should be dropped. Some of these address issues w.r.t. `yum` vs `apt`.

The kernel tests (as pointed out by Richard) also didn't really have a "wrong value" -- any non-empty command would really suffice.  So we changed them to actually _loading_ the module (closet thing to a wrong value) and added an empty file test elsewhere. 

Not tested on RHEL or Fedora. Could someone check that please? :-)

### DIFF
--- a/linux_os/guide/system/accounts/accounts-banners/file_owner_etc_motd/tests/missing.pass.sh
+++ b/linux_os/guide/system/accounts/accounts-banners/file_owner_etc_motd/tests/missing.pass.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+# platform = multi_platform_all
+
+rm -rf /etc/motd
--- a/linux_os/guide/system/bootloader-grub2/non-uefi/file_owner_grub2_cfg/tests/correct.pass.sh
+++ b/linux_os/guide/system/bootloader-grub2/non-uefi/file_owner_grub2_cfg/tests/correct.pass.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-chown 0:0 /boot/grub2/grub.cfg
+chown 0:0 {{{ grub2_boot_path }}}/grub.cfg
--- a/linux_os/guide/system/accounts/accounts-banners/file_groupowner_etc_issue/tests/correct.pass.sh
+++ b/linux_os/guide/system/accounts/accounts-banners/file_groupowner_etc_issue/tests/correct.pass.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+# platform = multi_platform_all
+
+touch /etc/issue
+chgrp 0 /etc/issue
--- a/linux_os/guide/system/accounts/accounts-banners/file_groupowner_etc_issue/tests/incorrect.fail.sh
+++ b/linux_os/guide/system/accounts/accounts-banners/file_groupowner_etc_issue/tests/incorrect.fail.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+# platform = multi_platform_all
+
+touch /etc/issue
+chgrp 1 /etc/issue
--- a/linux_os/guide/system/accounts/accounts-banners/file_groupowner_etc_motd/tests/correct.pass.sh
+++ b/linux_os/guide/system/accounts/accounts-banners/file_groupowner_etc_motd/tests/correct.pass.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+# platform = multi_platform_all
+
+touch /etc/motd
+chgrp 0 /etc/motd
--- a/linux_os/guide/system/accounts/accounts-banners/file_groupowner_etc_motd/tests/incorrect.fail.sh
+++ b/linux_os/guide/system/accounts/accounts-banners/file_groupowner_etc_motd/tests/incorrect.fail.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+# platform = multi_platform_all
+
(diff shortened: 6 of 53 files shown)


linux_os/guide/system/bootloader-grub2/non-uefi/file_owner_grub2_cfg/tests/correct.pass.sh
```diff
@@ -1,3 +1,7 @@
 #!/bin/bash
 
-chown 0:0 /boot/grub2/grub.cfg
+if [ -e /boot/grub2 ]; then
```
jan-cerny, line 3: Does anybody know whether we can use Jinja macros also in the test scenarios?

linux_os/guide/system/bootloader-grub2/non-uefi/file_owner_grub2_cfg/tests/correct.pass.sh
```diff
@@ -1,3 +1,7 @@
...
+    chown 0:0 /boot/grub2/grub.cfg
```
jan-cerny, line 4: So if I accidentally create this file on Ubuntu and accidentally the rule won't pick the correct path the test scenario will pass. I think it's unlikely to happen, but shouldn't we rather create different test scenarios that would have the `# platform =` specified in the header?

jan-cerny (commented):
They all pass on RHEL 8  virtual machine except the 2 rules linux_os/guide/system/accounts/accounts-banners/file_owner_etc_motd/tests/missing.pass.sh linux_os/guide/system/accounts/accounts-banners/file_groupowner_etc_motd/tests/missing.pass.sh

linux_os/guide/system/accounts/accounts-banners/file_owner_etc_motd/tests/missing.pass.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+# platform = multi_platform_all
+
+rm -rf /etc/motd
```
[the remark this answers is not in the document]
jan-cerny: Aha, I see that it has been changed in the meanwhile in https://github.com/ComplianceAsCode/content/pull/7131/commits/6bbfad35d925003a48313c79416ae82c6e6be7a4

linux_os/guide/system/bootloader-grub2/non-uefi/file_owner_grub2_cfg/tests/correct.pass.sh
cipherboy: We can't (yet). That is something I'm working on enabling, to eventually also support templated test cases. :-) I think it requires some changes to the API. See:

https://github.com/ComplianceAsCode/content/blob/master/tests/ssg_test_suite/common.py#L236-L253

linux_os/guide/system/bootloader-grub2/non-uefi/file_owner_grub2_cfg/tests/correct.pass.sh
cipherboy: See above; I'd rather we use jinja macros eventually, but that'll have to wait for the enablement to finish on my part. Then we can pull `grub2_boot_path` from the `product.yml`.

Also, who would do that to their system? :o Poor, poor computer.

linux_os/guide/system/bootloader-grub2/non-uefi/file_owner_grub2_cfg/tests/correct.pass.sh
cipherboy: I plan to revisit this shortly now that #7210 has merged. :-)

linux_os/guide/system/software/sudo/package_sudo_installed/tests/installed.pass.sh
```diff
@@ -1,4 +1,8 @@
 #!/bin/bash
 # package = yum
 
-yum install -y sudo
+if command -v yum; then
+    yum install -y sudo
+elif command -v apt-get; then
+    DEBIAN_FRONTEND=noninteractive apt-get install -y sudo
```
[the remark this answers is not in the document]
cipherboy: Dropping this commit as we'd prefer to address it in #7211 where templated tests can be used instead.

linux_os/guide/system/bootloader-grub2/non-uefi/file_owner_grub2_cfg/tests/correct.pass.sh
cipherboy: @jan-cerny This has been updated.

linux_os/guide/system/bootloader-grub2/non-uefi/file_owner_grub2_cfg/tests/correct.pass.sh
cipherboy: Does this suffice? Note that as we currently audit in CaC, it wouldn't catch this anyways, so IMO, testing should be aligned with what we can currently audit.

Though, with #7260, theoretically we could handle both paths... I dunno. Up to you.

cipherboy:
@jan-cerny This PR has been updated.

jan-cerny (approved):

PR #7084, opened 2021-06-04
Kubernetes Remediations: Add YAML separator

JAORMX:
This ensures that a correct YAML separator is added to all kubernetes
remediations.

### DIFF
--- a/applications/openshift/api-server/api_server_audit_log_maxsize/kubernetes/shared.yml
+++ b/applications/openshift/api-server/api_server_audit_log_maxsize/kubernetes/shared.yml
@@ -1,3 +1,4 @@
+---
 # platform = multi_platform_ocp
 apiVersion: config.openshift.io/v1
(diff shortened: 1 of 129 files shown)


jhrozek:
Maybe a silly question, but does it matter if the magic comments that say which platform the remediation is for are after the separator?

JAORMX:
> Maybe a silly question, but does it matter if the magic comments that say which platform the remediation is for are after the separator?

I don't know :/

jhrozek:
On Fri, Jun 04, 2021 at 04:22:58AM -0700, Juan Osorio Robles wrote:
> > Maybe a silly question, but does it matter if the magic comments that say which platform the remediation is for are after the separator?
> 
> I don't know :/

np, let me check the built output.

jhrozek (approved):
the metadata seem to be retained, so lgtm

Issue #6898, opened 2021-04-26
Bugged script for CCE-82024-1


BloodBlight:
#### Description of problem:
When running the remediation script for CCE-82024-1, it results in the following error:
sed: -e expression #1, char 33: unknown option to `s'

#### SCAP Security Guide Version:
Not sure

#### Operating System Version:
RHEL 7

#### Steps to Reproduce:
Create a script for CCE-82024-1 and run it.  Code:
```
# Remediation is applicable only in certain platforms
if [ ! -f /.dockerenv ] && [ ! -f /run/.containerenv ]; then

if LC_ALL=C grep -q -m 1 "^install dccp" /etc/modprobe.d/dccp.conf ; then
	sed -i 's/^install dccp.*/install dccp /bin/true/g' /etc/modprobe.d/dccp.conf
else
	echo -e "\n# Disable per security requirements" >> /etc/modprobe.d/dccp.conf
	echo "install dccp /bin/true" >> /etc/modprobe.d/dccp.conf
fi

else
    >&2 echo 'Remediation is not applicable, nothing was done'
fi
```

#### Actual Results:
ERROR: sed: -e expression #1, char 33: unknown option to `s'

#### Expected Results:
File should be altered.

#### Additional Information/Debugging Steps:

Ran  's/^install dccp.*/install dccp /bin/true/g' trough https://sed.js.org/ and got the same error.  Something is bugged on that line.  I'm not very good with sed, so I'm not sure what it should look like.

BloodBlight:
Recommended fix from one of my co-workers:

`sed -i 's#^install dccp.*#install dccp /bin/true#g' /etc/modprobe.d/dccp.conf`

Seems to have worked after testing.

BloodBlight:
Looks like CCE-27277-3 has this issue as well.  Maybe someone using an odd version of sed when making/testing these?

cipherboy:
@BloodBlight Good find! I think this is caused by the [template](https://github.com/ComplianceAsCode/content/blob/master/shared/templates/kernel_module_disabled/bash.template#L13). It'd be a very easy fix if you want to open a PR for it :-)

PR #7086, opened 2021-06-04
Fix kernel module disable template

cipherboy:
#### Description:

Kernel module disabling includes `/bin/true` -- but we use `/` as the path separator in our `sed` invocation. 

Fix this bug and make sure we don't introduce similar in the future. 

--- 

Resolves: #6898


yuumasato (approved):
Looks great to me.

PR #7088, opened 2021-06-07
Enable auditd_audispd_configure_sufficiently_large_partition in RHEL8 (RHEL-08-030660)

ggbecker:
#### Description:
- Enable auditd_audispd_configure_sufficiently_large_partition in RHEL8.
  - Select it into RHEL8 STIG profile.

@matusmarhefka @mildas This rule will probably have an impact on STIG tests because it requires 10GB on /var/log/audit and we usually setup 512MB
https://github.com/ComplianceAsCode/content/blob/95d892d7d058698c945115042f4901f8d5496215/rhel8/kickstart/ssg-rhel8-stig-ks.cfg#L123

I'm updating the kickstart file as well, but that means disk size will grow.


JAORMX (approved):

carlosmmatos:
@ggbecker Can you fix the conflicts (probably due to products directory change) and then I will merge.

ggbecker:
> @ggbecker Can you fix the conflicts (probably due to products directory change) and then I will merge.

Done.

@matusmarhefka @mildas reminding you that we'll probably need to take some actions for the tests for RHEL8 STIG because of disk space.

PR #7091, opened 2021-06-07
Enable sudo_require_reauthentication in RHEL8 (RHEL-08-010384)

ggbecker:
#### Description:

- Enable `sudo_require_reauthentication` in RHEL8
  - Select it into RHEL8 STIG profile.

The `[-]?` was added to regexes so it makes sure negative numbers are also matched.


carlosmmatos (approved):
LGTM

ggbecker:
I have rebased the pull request. I also gave preference on using the template option `option_regex_suffix: '=[-]?\w+\b'` instead of hardcoding this value into the template.

shared/templates/sudo_defaults_option/ansible.template
```diff
@@ -8,7 +8,7 @@
 - name: Ensure {{{ OPTION }}} is enabled with the appropriate value in /etc/sudoers
   lineinfile:
     path: /etc/sudoers
-    regexp: '^[\s]*Defaults\s(.*)\b{{{ OPTION }}}=\w+\b(.*)$'
+    regexp: '^[\s]*Defaults\s(.*)\b{{{ OPTION }}}=[-]?\w+\b(.*)$'
```
cipherboy, line 11: Ah we might need to re-evaluate our `OPTION_REGEX_SUFFIX` then. 

Maybe it should be replaced with `OPTION_VALUE_REGEX` and `OPTION_SUFFIX_REGEX`? Then `OPTION_REGEX_SUFFIX = OPTION_VALUE_REGEX + OPTION_SUFFIX_REGEX` and in bash/ansible we can add-in capture group (around the `OPTION_SUFFIX_REGEX`)? 

Thoughts?

linux_os/guide/system/software/sudo/sudo_require_reauthentication/rule.yml
```diff
@@ -46,6 +48,10 @@ template:
     vars:
         option: timestamp_timeout
         variable_name: "var_sudo_timestamp_timeout"
+        # optional minus char added so remediation can detect properly if item is already configured
+        option_regex_suffix: '=[-]?\w+\b'
```
cipherboy, line 52: This is fine, but just note that this regex only currently applies to the OVAL (which you've turned off below and replaced with a custom one above). :-)

ggbecker: That's true. I overlooked this, it's kind of useless then. The quick fix is to update the regex directly to accept the minus char.

shared/templates/sudo_defaults_option/ansible.template
ggbecker: I'll see what I can do... I'm probably gonna get the shortcut though xD

linux_os/guide/system/software/sudo/sudo_require_reauthentication/rule.yml
ggbecker: actually it applies to the bash remediation as well: https://github.com/ComplianceAsCode/content/blob/master/shared/templates/sudo_defaults_option/template.py#L26 and without it the remediation doesn't work properly. I'll probably leave it like this.

cipherboy: Ah ok :) No reason to hold this up then. :)

PR #7095, opened 2021-06-08
GH Actions/Build: Sync the product list with Jenkins job

evgenyz:


JAORMX:
@evgenyz is it already possible to add `ocp4` to this product list? Or do we still have a limitation regarding the `oscap` version?

evgenyz:
> @evgenyz is it already possible to add `ocp4` to this product list? Or do we still have a limitation regarding the `oscap` version?

Not for Ubuntu 20.04. I'm working on another action that would be based on Fedora (containerized), where it should be possible. It'd be a different PR. Or maybe not. I'll convert it to a draft for a while.

JAORMX:
@evgenyz that's fine. I was just curious. I think this PR is good as it is.

We either way build the OCP4 content as part of the `ci/prow/images` job.

evgenyz:
@mildas Can you please check that the CTF is working properly? Also, would it be possible to use `actions/checkout` to fetch the repo to make it more aparent?

evgenyz:
Otherwise it LGTM.

mildas:
@evgenyz CTF works properly for me. What exactly doesn't work for you?

The CTF knows how to analyze changes in Ansible remediations, Bash remediations, Jinja macros, OVAL checks, profiles, and Python files. If it doesn't know all filetypes in a Pull request or it doesn't find any change that needs to be tested (e.g. PR changes only description in profile file) then it doesn't print anything. I did this to not create useless outputs like "nothing identified" or "no test selected".

Example - your PR that changes files unknown for CTF vs PR changing OVAL check:
```
$ python3 content_test_filtering.py pr --output-format markdown 7095
$ python3 content_test_filtering.py pr --output-format markdown 7103
**Changes identified:**
Rules:
 no_shelllogin_for_systemaccounts

<details>
<summary>Show details</summary>

Rule no_shelllogin_for_systemaccounts:
 Text changed in OVAL check.
</details>


 build_product rhel8
 tests/test_suite.py rule --libvirt qemu:///system test-suite-vm --remediate-using bash --datastream build/ssg-rhel8-ds.xml no_shelllogin_for_systemaccounts
```

> Also, would it be possible to use actions/checkout to fetch the repo to make it more aparent?

Yes, it should be. You have 2 options - `--local` + `--repository` or only `--repository`.
With `--local` and `--repository ${LOCAL_CONTENT_REPO}`, it will not clone the repository and use local one. But it expects that everything is already locally available:
- updated master branch - without updated master, it will analyze also commits between non-updated master and PR (if the PR was created from newer master than the locally available master)
- compared branch is available - `content-test-filtering.py pr .... XYZ` expects that local repository has `pr-XYZ` branch available, `content-test-filtering branch ... BRANCH` expects that `BRANCH` is there

Without `--local`, it will fetch branch/pr to local repo. Problem here is, that it doesn't remove the branch/pr if it already locally exists. It just fetches the branch. So if a branch already locally exists and remote branch is updated (e.g. by `git commit --amend`), CTF doesn't notice the new change.

evgenyz:
Okay, got it.

I tired 
> With --local and --repository ${LOCAL_CONTENT_REPO}, it will not clone the repository and use local one. But it expects that everything is already locally available:

but got stuck with absence of `pr-XYZ`.

I think that we can make it a TODO item to switch to `action/checkout@v2` and `mshick/add-pr-comment@v1` (or something similar).

mildas:
@evgenyz Sure, it can be as TODO now and I will update CTF if e.g. some new option is needed for `action/checkout@v2`. Actually, what is needed? If you need to use a specific branch name, you can use `content-test-filtering.py branch` instead of `pr`.

.github/workflows/build.yaml
```diff
@@ -1,23 +1,63 @@
...
-        run: |
-          ./build_product rhel7 rhel8
+        run: ./build_product chromium fedora firefox fuse6 rhcos4 rhel7 rhel8 rhel9 rhosp10 rhosp13 sle12 sle15 ubuntu2004
       - name: Test
         run: ctest -j2 --output-on-failure -E unique-stigids
         working-directory: ./build
+      - name: Build Ansible Roles
+        run: PYTHONPATH=. python3 utils/ansible_playbook_to_role.py --build-playbooks-dir ./build/ansible/ --dry-run ./build/ansible_roles
+      - name: Lint Ansible Roles
+        run: ansible-lint -x 204 -x experimental ./build/ansible_roles/*
```
yuumasato, line 26: @evgenyz We discussed this today, and I think we decided to have it as nightly job for now, right? Or it will remain here until we have something ran nightly on Jenkins?
So could you move the Ansible Roles Lint test out of the gating?

evgenyz: Yeah, yeah, forgot to remove it from PR. Done.

yuumasato: Is it a good idea to remove `.github/workflows/validate_ansible_roles.yaml` while there is not yet a replacement for it?

yuumasato: I'm fine removing it if you plan to add it to Jenkins soon-ish.

evgenyz: Err, it's already in the history of the code. For the lifetime of the repo :)

evgenyz: And we don't want it to be in Actions anyway

evgenyz: I'm planing to add it to Go-CD

yuumasato (approved):

Issue #7093, opened 2021-06-07
SSGTS - Quoting values in variables for a test case?


cipherboy:
#### Description of problem:

How are values for variables supposed to be encoded? The documentation under `tests/README.md` is lacking this piece of information.

Suppose I want a literal value `"/tmp/something.log"` (including the double quotes). `scap-workbench` generates the following:

```xml
    <xccdf:set-value idref="xccdf_org.ssgproject.content_value_var_sudo_logfile">"/tmp/something"</xccdf:set-value>
  </xccdf:Profile>
```

However, I'm unable to get this to work when done via the test suite:

```bash
linux_os/guide/system/software/sudo/sudo_add_logfile/tests/logfile_enabled_with_quotes.pass.sh
#!/bin/bash
# platform = multi_platform_all
# variables = var_sudo_logfile="/tmp/something"

echo 'Defaults logfile="/tmp/something"' >> /etc/sudoers
```

(Or, using `&quot;` to escape it). 

When running, I get the following:

```
DEBUG - Applying script logfile_enabled_with_quotes.pass.sh
DEBUG - Using test script logfile_enabled_with_quotes.pass.sh with context pass
DEBUG - Searching for viable /home/cipherboy/GitHub/cipherboy/ComplianceAsCode-content/tests/logs/rule-custom-2021-06-07-1420/xccdf_org.ssgproject.content_rule_sudo_add_logfile-logfile_enabled_with_quotes.pass.sh-initial.verbose.log
DEBUG - Found /home/cipherboy/GitHub/cipherboy/ComplianceAsCode-content/tests/logs/rule-custom-2021-06-07-1420/xccdf_org.ssgproject.content_rule_sudo_add_logfile-logfile_enabled_with_quotes.pass.sh-initial.verbose.log
DEBUG - Searching for viable /home/cipherboy/GitHub/cipherboy/ComplianceAsCode-content/tests/logs/rule-custom-2021-06-07-1420/xccdf_org.ssgproject.content_rule_sudo_add_logfile-logfile_enabled_with_quotes.pass.sh-initial.html
DEBUG - Found /home/cipherboy/GitHub/cipherboy/ComplianceAsCode-content/tests/logs/rule-custom-2021-06-07-1420/xccdf_org.ssgproject.content_rule_sudo_add_logfile-logfile_enabled_with_quotes.pass.sh-initial.html
[...]

PR #7097, opened 2021-06-08
Fix SSG test suite support for setting variables

cipherboy:
#### Description

When trying to set variables from within scripts, SSGTS has a very
limited regex for allowed variables and values:

    r'^# {0} = ([ =,_\.\-\w\(\)]*)$'.format(parameter)

This greatly restricted the ability to set certain non-word characters
such as quotes and percent signs (e.g., 25% remaining).

Loosen up the regex considerably to:

    r'^# {0} = (.*)$'.format(parameter)

This allows us to capture the entire value of this parameter, regardless
of its value. I feel this is safe because we explicitly check for

    # <parameter> =

Which should limit false-positives and is otherwise easy enough to work
around (remove spaces around the equals, remove initial leading space,
use two pound signs, &c).

Resolves: #7093

---

I came across this limitation on #7082 (which was since merged). I was trying to set a variable with literal double quote in it `"`. I also came across it internally today on a variable with a `%` in its value, causing me to actually investigate why :D 

Let me know if anyone disagrees with the change.

### DIFF
--- a/tests/ssg_test_suite/rule.py
+++ b/tests/ssg_test_suite/rule.py
@@ -312,7 +312,7 @@ class RuleChecker(oscap.Checker):
         with open(script, 'r') as script_file:
             script_content = script_file.read()
             for parameter in params:
-                found = re.search(r'^# {0} = ([ =,_\.\-\w\(\)]*)$'.format(parameter),
+                found = re.search(r'^# {0} = (.*)$'.format(parameter),
                                   script_content,
                                   re.MULTILINE)
                 if found is None:


yuumasato (approved):
I agree, this change looks safe to me.
The regex search already looks for the parameter name and it seems safe to capture whatever is after the  ` = `.

It has been like this since the [beginning](https://github.com/ComplianceAsCode/content/commit/55a878f53c48612f7017776deefc62c4595ec205#diff-9e608ba6628274b91631c6aa076d74c9626341916e7a700dba3bd835cc64a43cR25), not sure why though. 
Not merging yet, just  in case someone knows why.

matusmarhefka (approved):
LGTM, the regex is there since the beginning, it was updated only once. I don't see any reason for such complicated regex so I agree with this change.

PR #7099, opened 2021-06-08
Only enable ansible/yaml lint tests when playbooks are built

ggbecker:
#### Description:

- Only enable ansible/yaml lint tests when playbooks are built
- Add two tests to ansible playbooks per profile since there are a few post-processing made by python scripts and it makes sense to test them in addition to playbooks per rule.

Relates to #7039

### DIFF
--- a/cmake/SSGCommon.cmake
+++ b/cmake/SSGCommon.cmake
@@ -302,7 +302,7 @@ macro(ssg_build_remediations PRODUCT)
     if ("${PRODUCT_ANSIBLE_REMEDIATION_ENABLED}")
         # only enable the ansible syntax checks if we are using openscap 1.2.17 or higher
         # older openscap causes syntax errors, see https://github.com/OpenSCAP/openscap/pull/977
-        if (ANSIBLE_PLAYBOOK_EXECUTABLE AND "${OSCAP_VERSION}" VERSION_GREATER "1.2.16")
+        if (SSG_ANSIBLE_PLAYBOOKS_ENABLED AND ANSIBLE_PLAYBOOK_EXECUTABLE AND "${OSCAP_VERSION}" VERSION_GREATER "1.2.16")
             add_test(
                 NAME "ansible-playbook-syntax-check-${PRODUCT}"
                 COMMAND "${CMAKE_SOURCE_DIR}/tests/ansible_playbook_check.sh" "${ANSIBLE_PLAYBOOK_EXECUTABLE}" "${CMAKE_BINARY_DIR}/ansible" "${PRODUCT}"
--- a/cmake/SSGCommon.cmake
+++ b/cmake/SSGCommon.cmake
@@ -309,17 +309,33 @@ macro(ssg_build_remediations PRODUCT)
             )
         endif()
         if (ANSIBLE_CHECKS)
-            if (ANSIBLE_LINT_EXECUTABLE AND "${OSCAP_VERSION}" VERSION_GREATER "1.2.16")
-                add_test(
-                    NAME "ansible-playbook-ansible-lint-check-${PRODUCT}"
-                    COMMAND "${CMAKE_SOURCE_DIR}/tests/ansible_playbook_check.sh" "${ANSIBLE_LINT_EXECUTABLE}" "${CMAKE_BINARY_DIR}/${PRODUCT}/playbooks" "${CMAKE_SOURCE_DIR}/tests/ansible-lint_config.yml"
-                )
+            if (SSG_ANSIBLE_PLAYBOOKS_PER_RULE_ENABLED)
+                if (ANSIBLE_LINT_EXECUTABLE)
+                    add_test(
+                        NAME "ansible-playbook-per-rule-ansible-lint-check-${PRODUCT}"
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -98,6 +98,15 @@ option(SSG_SCIENTIFIC_LINUX_DERIVATIVES_ENABLED "If enabled, Scientific Linux de
 
 option(FORCE_VALIDATE_EVERYTHING "If enabled, perform all validation tests regardless of oscap version requirements. By default, the build system may not perform validation of some content types if the openscap used to build it is not up-to-date" FALSE)
 
+# Remove this option when we would like to run ansible-lint and yamllint against our playbooks by default.
+# Right now these checks are not performed and need to be enabled by adding -DANSIBLE_CHECKS=ON to cmake
+# before running ctest.
+option(ANSIBLE_CHECKS "Set to ON to enable ansible-lint and yamllint checks" OFF)
+if(ANSIBLE_CHECKS)
+# ANSIBLE_CHECKS needs to have individual playbooks per rule enabled
```
cipherboy, line 106: Maybe add a warning so if anyone sets `-DSSG_ANSIBLE_PLAYBOOKS_PER_RULE_ENABLED=OFF` explicitly on the CLI, they know why it was enabled? Though default is FALSE. :-) 

Optionally, you could move this even higher up and have the default value of `SSG_ANSIBLE_PLAYBOOKS_PER_RULE_ENABLED` differ based on `ANSIBLE_CHECKS` value. That'll correctly show the warning only when a user explicitly calls `-DANSIBLE_CHECKS=ON -DSSG_ANSIBLE_PLAYBOOKS_PER_RULE_ENABLED=OFF` and catch the conflict. :-) 

Just a thought.

ggbecker: I have added a more elaborated condition and a warning in case it's set to OFF

cipherboy: Looks good, thanks @ggbecker!

cipherboy (approved):

CMakeLists.txt
```diff
@@ -98,6 +98,18 @@ option(SSG_SCIENTIFIC_LINUX_DERIVATIVES_ENABLED "If enabled, Scientific Linux de
...
 
+# Remove this option when we would like to run ansible-lint and yamllint against our playbooks by default.
+# Right now these checks are not performed and need to be enabled by adding -DANSIBLE_CHECKS=ON to cmake
+# before running ctest.
+option(ANSIBLE_CHECKS "Set to ON to enable ansible-lint and yamllint checks" ON)
+if(ANSIBLE_CHECKS)
+    # ANSIBLE_CHECKS needs to have individual playbooks per rule enabled
+    if(NOT SSG_ANSIBLE_PLAYBOOKS_PER_RULE_ENABLED)
+        message(WARNING "ANSIBLE_CHECKS is set to ON but it requires SSG_ANSIBLE_PLAYBOOKS_PER_RULE_ENABLED. Enforcing its value to ON.")
+        set(SSG_ANSIBLE_PLAYBOOKS_PER_RULE_ENABLED ON)
```
yuumasato, line 109: Is there a special need relationship between `ANSIBLE_CHECKS` and `ANSIBLE_PLAYBOOKS_PER_RULE_ENABLED` that requires it build the rule playbooks?
Forcing build and installation of per rule playbooks por the purpose of running linters seems awkward to me.

yuumasato: I think the issue is that the testing scrip `tests/ansible_playbook_check.sh` doesn't care about which playbooks it is checking. It just searches the build directory.

ggbecker:
The PR has turned into a different direction as we realized it makes more sense to tests things that are actually being built and not force to build content to be tested. So the ansible/yamllint tests are enabled only the flags to build the ansible content is enabled.

yuumasato (changes requested):
Looking good, just found one issue with tests when profile playbooks are not built.

cmake/SSGCommon.cmake
```diff
@@ -309,17 +309,33 @@ macro(ssg_build_remediations PRODUCT)
...
+                        COMMAND "${CMAKE_SOURCE_DIR}/tests/ansible_playbook_check.sh" "${YAMLLINT_EXECUTABLE}" "${CMAKE_BINARY_DIR}/${PRODUCT}/playbooks" "${CMAKE_SOURCE_DIR}/tests/yamllint_config.yml"
+                    )
+                endif()
             endif()
-            if (YAMLLINT_EXECUTABLE AND "${OSCAP_VERSION}" VERSION_GREATER "1.2.16")
-                add_test(
-                    NAME "ansible-playbook-yamllint-check-${PRODUCT}"
-                    COMMAND "${CMAKE_SOURCE_DIR}/tests/ansible_playbook_check.sh" "${YAMLLINT_EXECUTABLE}" "${CMAKE_BINARY_DIR}/${PRODUCT}/playbooks" "${CMAKE_SOURCE_DIR}/tests/yamllint_config.yml"
-                )
+            if (SSG_ANSIBLE_PLAYBOOKS_ENABLED)
```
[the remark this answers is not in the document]
ggbecker: Got it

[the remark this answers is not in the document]
ggbecker: Done in: dd273ee831fc80244b9754ad30aa9dddde32337a

yuumasato (approved):
Looks good, just pending jenkins.

yuumasato:
I see some rules report `Status: FAIL`, but I don't think these are unrelated to this PR.
There are also some infra issues.
CC: @JAORMX

PR #7100, opened 2021-06-09
WIP: Explicitly mention the variable to tune the rule

jhrozek:
So far one rule was changed to get feedback.

### DIFF
--- a/applications/openshift/kubelet/kubelet_eviction_thresholds_set_hard_memory_available/rule.yml
+++ b/applications/openshift/kubelet/kubelet_eviction_thresholds_set_hard_memory_available/rule.yml
@@ -38,7 +38,10 @@ description: |-
 
   <p>
   This rule pertains to the <tt>memory.available</tt> setting of the <tt>evictionHard</tt>
-  section.
+  section. The value the rule checks can be tuned using the variable with the
+  XCCDF ID <tt>xccdf_org.ssgproject.content_value_var_kubelet_evictionhard_memory_available</tt>.
+  When used through the Compliance Operator, the variable would be named
+  <tt>$profile-var-kubelet-evictionhard-memory-available</tt>
   </p>
 
 rationale: |-


applications/openshift/kubelet/kubelet_eviction_thresholds_set_hard_memory_available/rule.yml
```diff
@@ -38,7 +38,10 @@ description: |-
 
   <p>
   This rule pertains to the <tt>memory.available</tt> setting of the <tt>evictionHard</tt>
-  section.
+  section. The value the rule checks can be tuned using the variable with the
+  XCCDF ID <tt>xccdf_org.ssgproject.content_value_var_kubelet_evictionhard_memory_available</tt>.
```
JAORMX, line 42: There is a jinja template to get variables. Might wanna use that instead.

jhrozek:
On Wed, Jun 09, 2021 at 05:51:18AM -0700, Juan Osorio Robles wrote:
> @JAORMX commented on this pull request.
> 
> 
> 
> > @@ -38,7 +38,10 @@ description: |-
>  
>    <p>
>    This rule pertains to the <tt>memory.available</tt> setting of the <tt>evictionHard</tt>
> -  section.
> +  section. The value the rule checks can be tuned using the variable with the
> +  XCCDF ID <tt>xccdf_org.ssgproject.content_value_var_kubelet_evictionhard_memory_available</tt>.
> 
> There is a jinja template to get variables. Might wanna use that instead.

I know there's a template to get the variable value, but what macro
would you use here?

matejak:
> I know there's a template to get the variable value, but what macro would you use here?

You are right, we don't have a macro that would supply you with "This rule is parametrized by the ... XCCDF value (which is currently set to ...)" text. We plan to add this to the build system at some point, but that's not any time soon, as we need to extract the parametrization from OVAL and bring it to the rule when it is processed.

PR #7103, opened 2021-06-09
Fix OVAL issues of the no_shelllogin_for_systemaccounts rule

richardmaciel-canonical:
- object_etc_passwd_entries was missing /usr/bin/false and /bin/false
  which are used by some system accounts in Ubuntu
- test_shell_defined_default_uid_range does not require object existence
  to pass anymore

Signed-off-by: Richard Maciel Costa <richard.maciel.costa@canonical.com>

#### Description:

This pull request fixes two problems found in the no_shelllogin_for_systemaccounts.

The first one was the lack of /usr/bin/false and /bin/false paths in the regular expression used by the object_etc_passwd_entries OVAL object. Since some Linux distros traditionally use the aforementioned paths, this test would fail on those systems.

Second one was the wrong use of check_existence=all_exist in the test_shell_defined_default_uid_range test. This could lead to false negatives when the system does not use SYS_UID_MIN and SYS_UID_MAX parameters in /etc/login.defs, because if the system doesn't have any user, the object_etc_passwd_entries will return 0 objects. With the all_exist modifier, the test_shell_defined_default_uid_range will fail, even if the system is in a valid state (i.e. no system user with a valid shell).

### DIFF
--- a/linux_os/guide/system/accounts/accounts-restrictions/root_logins/no_shelllogin_for_systemaccounts/oval/shared.xml
+++ b/linux_os/guide/system/accounts/accounts-restrictions/root_logins/no_shelllogin_for_systemaccounts/oval/shared.xml
@@ -4,8 +4,8 @@
       a login shell.") }}}
     <criteria operator="OR">
       <!-- If SYS_UID_MIN and SYS_UID_MAX is not defined in /etc/login.defs
-           perform the check that all /etc/passwd entries having shell defined
-           have UIDs outside the default <0, UID_MIN -1> range.
+           perform the check that any /etc/passwd entry with shell defined
+           has UID outside the default <0, UID_MIN -1> range.
            If at least one UID with shell defined exists within that range,
            the requirement isn't met -->
       <criteria operator="AND">
@@ -18,9 +18,9 @@
       </criteria>
       <!-- If both SYS_UID_MIN and SYS_UID_MAX are defined in /etc/login.defs
            perform both checks:
-           * That all /etc/passwd entries having shell defined have UIDs outside
+           * That any /etc/passwd entry with shell defined has UID outside
            the range for reserved system accounts <0, SYS_UID_MIN> range,
-           * That all /etc/passwd entries having shell defined have UIDs outside
+           * That any /etc/passwd entries with shell defined has UID outside
            the range for dynamically allocated system accounts <SYS_UID_MIN, SYS_UID_MAX>
            If at least one UID with shell defined exists within some of the two
            ranges, the requirement isn't met -->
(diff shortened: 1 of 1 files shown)


cipherboy:
@openscap-ci add-to-whitelist 
@openscap-ci ok-to-test

linux_os/guide/system/accounts/accounts-restrictions/root_logins/no_shelllogin_for_systemaccounts/oval/shared.xml
```diff
@@ -71,7 +71,7 @@
   <!-- Get all /etc/passwd entries having shell defined as OVAL object -->
   <ind:textfilecontent54_object id="object_etc_passwd_entries" version="1">
     <ind:filepath>/etc/passwd</ind:filepath>
-    <ind:pattern operation="pattern match">^(?!root).*:x:([\d]+):[\d]+:[^:]*:[^:]*:(?!\/usr\/sbin\/nologin|\/sbin\/nologin|\/bin\/sync|\/sbin\/shutdown|\/sbin\/halt).*$</ind:pattern>
+    <ind:pattern operation="pattern match">^(?!root).*:x:([\d]+):[\d]+:[^:]*:[^:]*:(?!\/usr\/sbin\/nologin|\/sbin\/nologin|\/bin\/sync|\/sbin\/shutdown|\/sbin\/halt|\/bin\/false|\/usr\/bin\/false).*$</ind:pattern>
```
yuumasato, line 74: Hi, this change has been reverted previously.
See https://github.com/ComplianceAsCode/content/pull/3767 and its comments.

If this is really needed for Ubuntu, please add a conditional based on the product.

If you agree with the justification for avoiding accepting `/bin/false` and `/usr/bin/false, and willing to go an extra mile. It seems that a warning in the rule would help understand why they are not there.

linux_os/guide/system/accounts/accounts-restrictions/root_logins/no_shelllogin_for_systemaccounts/oval/shared.xml
```diff
@@ -93,7 +93,7 @@
   <!-- Thus check that all /etc/passwd entries having shell defined
        have UID outside of <0, UID_MIN -1> range -->
   <ind:textfilecontent54_test state_operator="OR" id="test_shell_defined_default_uid_range" check="all"
-  check_existence="all_exist" comment="&lt;0, UID_MIN - 1&gt; system UIDs having shell set"
+  check_existence="any_exist" comment="&lt;0, UID_MIN - 1&gt; system UIDs having shell set"
```
yuumasato, line 96: Nice catch, do think it would make sense to change the comment above to:
 `<!-- Thus check that any /etc/passwd entry with shell defined`

If you agree, would you mind doing that for `test_shell_defined_reserved_uid_range` and `test_shell_defined_dynalloc_uid_range`?

yuumasato (changes requested):

linux_os/guide/system/accounts/accounts-restrictions/root_logins/no_shelllogin_for_systemaccounts/oval/shared.xml
richardmaciel-canonical: Sure, no problem. Will add the comment to the other tests as well.

linux_os/guide/system/accounts/accounts-restrictions/root_logins/no_shelllogin_for_systemaccounts/oval/shared.xml
richardmaciel-canonical: Well, as @cipherboy remind me, that should apply to /bin/sync and, by extension, /sbin/shutdown, /sbin/halt too. And the latters are handled as exceptions in the aforementioned regexp.

Also, the way I see it, the problem of using the account for port forwarding must actually be mitigated by a security rule that locks system accounts. Using this rule for that purpose is a bad fit.

cipherboy: There's also an explicit parameter to block port forwarding in general. (`DisableForwarding` being the strongest, but also the `Allow*Forwarding` parameters can be set to `no`).

yuumasato: > Well, as @cipherboy reminded me, that should apply to /bin/sync and, by extension, /sbin/shutdown, /sbin/halt too. And the latters are handled as exceptions in the aforementioned regexp.

Yeah, you have a point. And re-reading the article `nologin` is not any better.
 
> Also, the way I see it, the problem of using the account for port forwarding must actually be mitigated by a security rule that locks system accounts. Using this rule for that purpose is a bad fit.

I agree.

cipherboy (approved):
I've looked at this a couple of times and keep forgetting to click the approve button. :D

yuumasato (approved):
Rerunning test as rhcos4 selects this rule and has e2e tests.

yuumasato:
Finally all green, :)

PR #7105, opened 2021-06-10
Add new rule to generate initial usbguard policy (RHEL-08-040140)

ggbecker:
#### Description:

- Add new rule to generate initial usbguard policy.

#### Rationale:

- Should solve problems like: https://bugzilla.redhat.com/show_bug.cgi?id=1970481
- This implementation is what STIG defines: https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_8_security_technical_implementation_guide/V-230524
  - The ideal implementation would actually be to use https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/usbguard/usbguard_allow_hid_and_hub/rule.yml


linux_os/guide/services/usbguard/usbguard_generate_policy/ansible/shared.yml
```diff
@@ -0,0 +1,37 @@
...
+      manager: auto
+
+  - name: Enable service usbguard
+    service:
+      name: "usbguard"
+      enabled: "yes"
+      state: "started"
+      masked: "no"
+    when:
+    - '"usbguard" in ansible_facts.packages'
```
carlosmmatos, line 37: Shouldn't this type of check be at each layer here? For example, I wouldn't run any of these tasks unless `"usbguard" in ansible_facts.packages`. What do you think?

linux_os/guide/services/usbguard/usbguard_generate_policy/bash/shared.sh
```diff
@@ -0,0 +1,17 @@
+# platform = multi_platform_rhel
+# reboot = false
+# strategy = configure
+# complexity = low
+# disruption = low
+
+USBGUARD_CONF=/etc/usbguard/rules.conf
+if [ ! -f "$USBGUARD_CONF" ] || [ ! -s "$USBGUARD_CONF" ]; then
+    usbguard generate-policy > $USBGUARD_CONF
```
carlosmmatos, line 9: I would say the same is true here as well.

carlosmmatos (changes requested):

carlosmmatos (approved):

carlosmmatos:
LGTM

PR #7107, opened 2021-06-10
Fix chrony maxpoll description, include peer in bash check

cipherboy:
#### Description:

As reported by @mralph-rh, the description text is vague and makes it sound like `maxpoll` goes on its own line; it doesn't.

Additionally `peer` is a valid keyword and should also contain `maxpoll` entries. 

cc: @dodys -- this will affect Ubuntu as well. Curious what your thoughts on about 20.04 STIG as it doesn't appear to include `peer` entries.

### DIFF
--- a/linux_os/guide/services/ntp/chronyd_or_ntpd_set_maxpoll/bash/shared.sh
+++ b/linux_os/guide/services/ntp/chronyd_or_ntpd_set_maxpoll/bash/shared.sh
@@ -14,3 +14,8 @@ sed -i "s/^\(server.*maxpoll\) [0-9][0-9]*\(.*\)$/\1 $var_time_service_set_maxpo
 grep "^server" "$config_file" | grep -v maxpoll | while read -r line ; do
         sed -i "s/$line/& maxpoll $var_time_service_set_maxpoll/" "$config_file"
 done
+
+# Add maxpoll to peer entries without maxpoll
+grep "^peer" "$config_file" | grep -v maxpoll | while read -r line ; do
+        sed -i "s/$line/& maxpoll $var_time_service_set_maxpoll/" "$config_file"
+done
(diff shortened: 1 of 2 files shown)


dodys (approved):
LGTM

vojtapolasek (changes requested):
Thank you for the change. I like that you make the description more explicit. However, I have a problem with including the "peer" part only in the Bash remediation:
1. OVAL is not checking for it
2. there is no test for it
3. what about Ansible and Kubernetes remediations?
4. it is not mentioned in the description

Could you please improve the "peer" coverage by incorporating it at least into oval, preferably ansible? And definitely describing it.
Thank you.

vojtapolasek:
@cipherboy  do you plan to improve this PR?

cipherboy:
@vojtapolasek, no I think I had turned it over to @Mab879 if he wanted to do it, but not sure it found time on his plate yet.

dodys:
we might also need to deal with 'pool' entries?

cipherboy:
Just to clarify what's missing, per discussion on IRC with @dodys:

Benchmarks need to understand two things:

 - More than just `server` entries (also `peer` and `pool`). 
 - There's more than just `/etc/chrony.conf` -- there's a config directory (`/etc/chrony/conf.d`) and server/peer/pool directory iirc at something like `/etc/chrony/{servers,peers,pools}/*` just for configuring server/peer/pool statements.

Right now, this rule purely aligns with benchmarks; we could probably put the `peer` and `pool` entry into other rules and audit them separately (if desired) while encouraging benchmarks to adopt them. But I'd probably make sure _all_ configuration search paths are handled appropriately in a given rule. My 2c.

I'm guessing most of these directives would be used in more advanced configuration of chrony, but most people (benchmark authors included?) probably only do the bare minimum to configure chrony and don't use as complex of setup as it could support.

matejak:
Couldn't push to this branch, so watch the successor #8015

cipherboy:
@matejak Ah interesting, I wonder if they've changed the language. It now says:

> - [ ] Allow edits and access to secrets by maintainers
> 
> If checked, users with write access to ComplianceAsCode/content can add new commits to your fix-chrony branch. You can always change this setting later.
>
> Maintainers could potentially edit this repository's workflows to reveal values of secrets and gain access to other branches.

I've generally left it checked in the past, sorry about that!

PR #7109, opened 2021-06-11
Fix covscan shell findings

mildas:
#### Description:
```
Error: SHELLCHECK_WARNING (CWE-398):
/usr/share/scap-security-guide/bash/rhel7-script-anssi_nt28_enhanced.sh:1630:10: error[SC1066]: Don't use $ on the left side of assignments.
# 1628|       if grep --silent '^\s*TMOUT' $f; then
# 1629|           sed -i -E "s/^(\s*)TMOUT\s*=\s*(\w|\$)*(.*)$/\1TMOUT=$var_accounts_tmout\3/g" $f
# 1630|->         $tmout_found=1
# 1631|       fi
# 1632|   done
```

```
Error: SHELLCHECK_WARNING (CWE-88):
/usr/share/scap-security-guide/bash/rhel7-script-stig_gui.sh:191:14: error[SC2068]: Double quote array expansions to avoid re-splitting elements.
#  189|   	fi
#  190|   
#  191|-> 	for hash in ${forbidden_hashes[@]}
#  192|   	do
#  193|   		config=$(echo $config | sed "s/$hash//")
```

```
Error: SHELLCHECK_WARNING (CWE-398):
/usr/share/scap-security-guide/bash/rhel7-script-stig_gui.sh:1339:10: error[SC2045]: Iterating over ls output is fragile. Use globs.
# 1337|   (>&2 echo "Remediating rule 40/248: 'sudo_remove_nopasswd'")
# 1338|   
# 1339|-> for f in $( ls /etc/sudoers /etc/sudoers.d/* 2> /dev/null ) ; do
# 1340|     matching_list=$(grep -P '^(?!#).*[\s]+NOPASSWD[\s]*\:.*$' $f | uniq )
# 1341|     if ! test -z "$matching_list"; then
```

```
Error: SHELLCHECK_WARNING (CWE-398):
/usr/share/scap-security-guide/bash/rhel7-script-stig_gui.sh:19943:10: error[SC2045]: Iterating over ls output is fragile. Use globs.
#19941|   SSSD_CONF_DIR="/etc/sssd/conf.d/*.conf"
#19942|   
#19943|-> for f in $( ls $SSSD_CONF $SSSD_CONF_DIR 2> /dev/null ) ; do
#19944|   	# finds all services entries under [sssd] configuration category, get a unique list so it doesn't add redundant fix
#19945|   	services_list=$( awk '/^\s*\[/{f=0} /^\s*\[sssd\]/{f=1}f' $f | grep -P '^services[ \t]*=' | uniq )
```

```
Error: SHELLCHECK_WARNING (CWE-398):
/usr/share/scap-security-guide/bash/rhel8-script-anssi_bp28_high.sh:7011:17: warning[SC2125]: Brace expansions and globs are literal in assignments. Quote it or use an array.
# 7009|           sed -i "s/noexec.*//g" /etc/default/grub
# 7010|           sed -i "s/noexec.*//g" /etc/grub.d/*
# 7011|->         GRUBCFG=/boot/grub2/*.cfg
# 7012|           grub2-mkconfig -o "$GRUBCFG"
# 7013|     fi
```

### DIFF
--- a/linux_os/guide/services/sssd/sssd_enable_pam_services/bash/shared.sh
+++ b/linux_os/guide/services/sssd/sssd_enable_pam_services/bash/shared.sh
@@ -6,7 +6,10 @@
 SSSD_CONF="/etc/sssd/sssd.conf"
 SSSD_CONF_DIR="/etc/sssd/conf.d/*.conf"
 
-for f in $( ls $SSSD_CONF $SSSD_CONF_DIR 2> /dev/null ) ; do
+for f in $SSSD_CONF $SSSD_CONF_DIR; do
+	if [ ! -e "$f" ]; then
+		continue
+	fi
 	# finds all services entries under [sssd] configuration category, get a unique list so it doesn't add redundant fix
 	services_list=$( awk '/^\s*\[/{f=0} /^\s*\[sssd\]/{f=1}f' $f | grep -P '^services[ \t]*=' | uniq )
 
--- a/linux_os/guide/services/sssd/sssd_run_as_sssd_user/bash/shared.sh
+++ b/linux_os/guide/services/sssd/sssd_run_as_sssd_user/bash/shared.sh
@@ -1,7 +1,10 @@
 # platform = multi_platform_rhel,multi_platform_fedora,multi_platform_ol
 
 found=false
-for f in $( ls /etc/sssd/sssd.conf /etc/sssd/conf.d/*.conf 2> /dev/null ) ; do
+for f in /etc/sssd/sssd.conf /etc/sssd/conf.d/*.conf; do
+	if [ ! -e "$f" ]; then
+		continue
+	fi
 	user=$( awk '/^\s*\[/{f=0} /^\s*\[sssd\]/{f=1} f{nu=gensub("^\\s*user\\s*=\\s*(\\S+).*","\\1",1); if($0!=nu){user=nu}} END{print user}' $f )
 	if [ -n "$user" ] ; then
 		if [ "$user" != sssd ] ; then
--- a/linux_os/guide/system/permissions/restrictions/enable_execshield_settings/sysctl_kernel_exec_shield/bash/shared.sh
+++ b/linux_os/guide/system/permissions/restrictions/enable_execshield_settings/sysctl_kernel_exec_shield/bash/shared.sh
@@ -19,7 +19,6 @@ if [ "$(getconf LONG_BIT)" = "64" ] ; then
   if grep --silent noexec {{{ grub2_boot_path }}}/grub*.cfg ; then
         sed -i "s/noexec.*//g" /etc/default/grub
         sed -i "s/noexec.*//g" /etc/grub.d/*
-        GRUBCFG={{{ grub2_boot_path }}}/*.cfg
-        grub2-mkconfig -o "$GRUBCFG"
+        grub2-mkconfig -o "{{{ grub2_boot_path }}}"/grub*.cfg
   fi
 fi
--- a/linux_os/guide/system/software/sudo/sudo_add_noexec/tests/noexec_absent.fail.sh
+++ b/linux_os/guide/system/software/sudo/sudo_add_noexec/tests/noexec_absent.fail.sh
@@ -2,7 +2,10 @@
 # platform = multi_platform_all
 
 # Code taken from macro bash_sudo_remove_config()
-for f in $( ls /etc/sudoers /etc/sudoers.d/* 2> /dev/null ) ; do
+for f in /etc/sudoers /etc/sudoers.d/*; do
+  if [ ! -e "$f" ]; then
+    continue
+  fi
   matching_list=$(grep -P '^(?!#).*[\s]+noexec.*$' $f | uniq )
   if ! test -z "$matching_list"; then
     while IFS= read -r entry; do
--- a/linux_os/guide/system/software/sudo/sudo_custom_logfile/tests/logfile_absent.fail.sh
+++ b/linux_os/guide/system/software/sudo/sudo_custom_logfile/tests/logfile_absent.fail.sh
@@ -2,7 +2,10 @@
 # platform = multi_platform_all
 
 # Code taken from macro bash_sudo_remove_config()
-for f in $( ls /etc/sudoers /etc/sudoers.d/* 2> /dev/null ) ; do
+for f in /etc/sudoers /etc/sudoers.d/*; do
+  if [ ! -e "$f" ]; then
+    continue
+  fi
   matching_list=$(grep -P '^(?!#).*[\s]+logfile.*$' $f | uniq )
   if ! test -z "$matching_list"; then
     while IFS= read -r entry; do
--- a/linux_os/guide/system/accounts/accounts-session/accounts_tmout/bash/shared.sh
+++ b/linux_os/guide/system/accounts/accounts-session/accounts_tmout/bash/shared.sh
@@ -8,7 +8,7 @@ tmout_found=0
 for f in /etc/profile /etc/profile.d/*.sh; do
     if grep --silent '^\s*TMOUT' $f; then
         sed -i -E "s/^(\s*)TMOUT\s*=\s*(\w|\$)*(.*)$/\1TMOUT=$var_accounts_tmout\3/g" $f
-        $tmout_found=1
+        tmout_found=1
     fi
 done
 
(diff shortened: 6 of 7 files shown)


linux_os/guide/system/permissions/restrictions/enable_execshield_settings/sysctl_kernel_exec_shield/bash/shared.sh
```diff
@@ -19,7 +19,7 @@ if [ "$(getconf LONG_BIT)" = "64" ] ; then
   if grep --silent noexec {{{ grub2_boot_path }}}/grub*.cfg ; then
         sed -i "s/noexec.*//g" /etc/default/grub
         sed -i "s/noexec.*//g" /etc/grub.d/*
-        GRUBCFG={{{ grub2_boot_path }}}/*.cfg
-        grub2-mkconfig -o "$GRUBCFG"
+        GRUBCFG="{{{ grub2_boot_path }}}/*.cfg"
+        grub2-mkconfig -o $GRUBCFG
```
mildas, line 23: Not sure if this is correct approach.

linux_os/guide/services/sssd/sssd_enable_pam_services/bash/shared.sh
```diff
@@ -6,7 +6,7 @@
 SSSD_CONF="/etc/sssd/sssd.conf"
 SSSD_CONF_DIR="/etc/sssd/conf.d/*.conf"
 
-for f in $( ls $SSSD_CONF $SSSD_CONF_DIR 2> /dev/null ) ; do
+for f in $SSSD_CONF $SSSD_CONF_DIR; do
```
cipherboy, line 9: This is the wrong approach here. You'd need a glob and an existence check to accomplish this:

```bash
for f in $SSSD_CONF $SSSD_CONF_DIR/*; do
    if [ ! -e "$f" ]; then
        continue
    fi

    ... ## rest of check
done
```

Note that `ls` takes care of the existence check and the directory listing for you.

linux_os/guide/services/sssd/sssd_run_as_sssd_user/bash/shared.sh
```diff
@@ -1,7 +1,7 @@
 # platform = multi_platform_rhel,multi_platform_fedora,multi_platform_ol
 
 found=false
-for f in $( ls /etc/sssd/sssd.conf /etc/sssd/conf.d/*.conf 2> /dev/null ) ; do
+for f in /etc/sssd/sssd.conf /etc/sssd/conf.d/*.conf; do
```
cipherboy, line 4: As above; probably best to add the existence check in.

linux_os/guide/system/permissions/restrictions/enable_execshield_settings/sysctl_kernel_exec_shield/bash/shared.sh
cipherboy: I'd perhaps advocate for:

```
grub2-mkconfig -o "{{{ grub2_boot_path}}}"/grub*.cfg
```

and drop the intermediate variable. 

In practice, `grub2_boot_path` probably won't have spaces but shellcheck/covscan probably doesn't know that and so theoretically it could, hence the quoting, however, the old source is wrong as the quoted glob won't be expanded and instead passed directly to the program (leading it to write out something like `/boot/grub/*.cfg` as a real file on disk).

linux_os/guide/system/software/sudo/sudo_add_noexec/tests/noexec_absent.fail.sh
```diff
@@ -2,7 +2,7 @@
 # platform = multi_platform_all
 
 # Code taken from macro bash_sudo_remove_config()
-for f in $( ls /etc/sudoers /etc/sudoers.d/* 2> /dev/null ) ; do
+for f in /etc/sudoers /etc/sudoers.d/*; do
```
cipherboy, line 5: As above about `ls` and existence check.

linux_os/guide/system/software/sudo/sudo_custom_logfile/tests/logfile_absent.fail.sh
```diff
@@ -2,7 +2,7 @@
 # platform = multi_platform_all
 
 # Code taken from macro bash_sudo_remove_config()
-for f in $( ls /etc/sudoers /etc/sudoers.d/* 2> /dev/null ) ; do
+for f in /etc/sudoers /etc/sudoers.d/*; do
```
cipherboy, line 5: As above about `ls` and existence check.

cipherboy (changes requested):
@mildas Thanks for this! I think there's a few points you should probably address to prevent weird error messages. Also looked at the grub2 one, I think the original check was wrong and the fix is better, but I fear shellcheck might still complain, so I suggested a new fix.

linux_os/guide/services/sssd/sssd_enable_pam_services/bash/shared.sh
mildas: But the glob is already in variable `SSSD_CONF_DIR`
```
 SSSD_CONF_DIR="/etc/sssd/conf.d/*.conf" 
```
so it isn't needed in the `for`, right?

cipherboy: But `$SSSD_CONF` could just not exist :-)

cipherboy: Also:

```
for x in /tmp /tmp/sodfsdfasdf*; do
                         echo "$x"
                         done
/tmp
/tmp/sodfsdfasdf*
```

cipherboy: Edit: Sorry @mildas I misread what you said. yes, no need for the interior glob, but I'm worried shellcheck might complain in general, but maybe its analysis has gotten better understanding of literals/globs. :-)

cipherboy: It has gotten smarter since I last used it, nice! :-)

```
 cipherboy   .git  ~  tmp  46  shellcheck file.sh 

In file.sh line 8:
	services_list=$( awk '/^\s*\[/{f=0} /^\s*\[sssd\]/{f=1}f' $f | grep -P '^services[ \t]*=' | uniq )
                                                                  ^-- SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
	services_list=$( awk '/^\s*\[/{f=0} /^\s*\[sssd\]/{f=1}f' "$f" | grep -P '^services[ \t]*=' | uniq )


In file.sh line 12:
			sed -i "s/$services$/&, pam/" $f
                                                      ^-- SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
			sed -i "s/$services$/&, pam/" "$f"

For more information:
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
```

mildas: Ok, I'm going to add the if checks and update the variable

mildas:
@cipherboy Updated. I have kept same indentation for each file, so it might look strange that file is updated with tabs and other one with spaces :) And thank you for the comments, I've learned something new about tricky Bash.

linux_os/guide/system/permissions/restrictions/enable_execshield_settings/sysctl_kernel_exec_shield/bash/shared.sh
```diff
@@ -19,7 +19,6 @@ if [ "$(getconf LONG_BIT)" = "64" ] ; then
   if grep --silent noexec {{{ grub2_boot_path }}}/grub*.cfg ; then
         sed -i "s/noexec.*//g" /etc/default/grub
         sed -i "s/noexec.*//g" /etc/grub.d/*
-        GRUBCFG={{{ grub2_boot_path }}}/*.cfg
-        grub2-mkconfig -o "$GRUBCFG"
+        grub2-mkconfig -o "{{{ grub2_boot_path }}}/grub*.cfg"
```
cipherboy, line 22: Glob still needs to go outside of the quote, otherwise you end up passing the literal `*` as a file-path character to `grub2-mkconfig` instead of having the shell expand it. :-) Sorry!

mildas: Aha! Sorry for that. It should be fixed now

cipherboy (approved):

PR #7111, opened 2021-06-14
Splitting rhel7 cis profile - section 3

vojtapolasek:
#### Description:

- rewrite the benchmark into the controls file format
- update references where needed

#### Rationale:

Continuing efford to split CIS benchmark into 4 profiles. This covers section 3. Note that there are some rules missing for firewall configuration. We should also solve in future how to handle some rules which should be evaluated only when firewalld / nftables / iptables in installed.


controls/cis_rhel7.yml
```diff
@@ -965,3 +965,245 @@ controls:
...
+    automated: no # rule missing
+
+  - id: 3.5.3.3.6
+    title: Ensure ip6tables is enabled and running (Automated)
+    levels:
+    - l1_server
+    - l1_workstation
+    automated: yes
+    # rules:
+    #    - service_ip6tables_enabled
```
yuumasato, line 1209: These can be uncommented and the build system will handle it.

controls/cis_rhel7.yml
```diff
@@ -965,3 +965,245 @@ controls:
...
+    automated: no # rule missing
+
+  - id: 3.5.3.1.1
+    title: Ensure iptables packages are installed (Automated)
+    levels:
+    - l1_server
+    - l1_workstation
+    automated: yes
+    # rules:
+    # - package_iptables_installed
```
yuumasato, line 1107: Any reason why this rule is commented?
Is this commented because it is repeated? The control format is able to handle repeated rules just fine.

yuumasato: This item is missing `iptables-services' installed too.

vojtapolasek:
Thank you for review. I will fix typos. About iptables rules... in the CIS benchmark, there are three ways how to do the firewall; Firewalld, Nftables and Iptables. I wonder what will happen if I will allow installation of Firewalld and Iptables together or if Firewalld and IPtables service will be running at the same time.

yuumasato:
> Thank you for review. I will fix typos.

There are still 5 occurrences of `valueenabled`.

> About iptables rules... in the CIS benchmark, there are three ways how to do the firewall; Firewalld, Nftables and Iptables. I wonder what will happen if I will allow installation of Firewalld and Iptables together or if Firewalld and IPtables service will be running at the same time.

I see, the rules for installation of Firewalld, Nftables or Ip tables are not manual, like the chrony and NTP were.

In this situation we can either choose one firewall to be the default, and then have the rules to install its packages selected.
Or we remove all rules that install firewalls and let the sysadmin install the desired one manually.

cipherboy:
@yuumasato said:
> In this situation we can either choose one firewall to be the default, and then have the rules to install its packages selected.
Or we remove all rules that install firewalls and let the sysadmin install the desired one manually.

This is what I mentioned earlier w.r.t. variable-based pseudo-CPE and using it to select which Firewall to use (one of the upstream calls).

If you're interested in merging variable-based pseudo-CPEs, I can propose it as a PR. It is a nasty bit of work that will require close review to merge upstream, but I think it would be worthwhile, generally speaking. 

However, it will play tricks if @vojtapolasek decides to extend the `platforms` field to take OR/AND'd constraints including variable-based CPEs. I think the variable-based pseudo-CPEs might give Vojta a hint as to how to do it on the implementation side, though I don't have any guidance as for the language (in the `rule.yml`) itself. :)

vojtapolasek:
After discussion I decided to go through default selection of Firewalld. I don't see this as a permanent solution. But currently as the CPE applicability language support is still not complete, I see this as the best solution for keeping CIS profile "not wrong".

yuumasato:
> @yuumasato said:
> 
> > In this situation we can either choose one firewall to be the default, and then have the rules to install its packages selected.
> > Or we remove all rules that install firewalls and let the sysadmin install the desired one manually.
> 
> This is what I mentioned earlier w.r.t. variable-based pseudo-CPE and using it to select which Firewall to use (one of the upstream calls).
> 
> If you're interested in merging variable-based pseudo-CPEs, I can propose it as a PR. It is a nasty bit of work that will require close review to merge upstream, but I think it would be worthwhile, generally speaking.

We have something similar implemented for SSHDs at the OVAL level, and it is also nasty and complicates things a bunch. I'd lean towards not adding it upstream for now.

yuumasato:
@openscap-ci test this please

yuumasato (approved):
LGTM, just wating for mandatory tests to run.

PR #7112, opened 2021-06-14
Splitting rhel7 cis profile - section 2

vojtapolasek:
#### Description:

- rewrite section 2 of rhel7 cis into control file format
- update rule references where needed

#### Rationale:

This is another part of the process of splitting CIS benchmark into 4 profiles.


linux_os/guide/services/ntp/service_ntpd_enabled/rule.yml
```diff
@@ -20,6 +20,9 @@ rationale: |-
 
 severity: medium
 
+identifiers:
+    cce@rhel7: CCE-83433-3
```
jan-cerny, line 24: @vojtapolasek Why is the CCE moved from gnome_gdm_disable_xdmcp? I would keep the existing CCE in gnome_gdm_disable_xdmcp and assign the new one here instead. Or is there some reason for this?

vojtapolasek: Good point, this is a mistake. I will fix it.

controls/cis_rhel7.yml
```diff
@@ -567,3 +567,255 @@ controls:
...
+  - id: 2.2.1.3
+    title: Ensure ntp is configured (Automated)
+    levels:
+    - l1_server
+    - l1_workstation
+    automated: yes
+    rules:
+    - service_ntpd_enabled
+    - ntpd_configure_restrictions
+    - ntpd_specify_remote_server
```
yuumasato, line 607: In previous version we opinionated the profile toward `chrony`, as we had a rule to install it.
With the item `2.2.1.1` being manual, the choice whether to use ntpd or chrony is put on the sysadmin.

The `chrony` rules have `platform: chrony`.
But the only `ntpd' rule with `platform: ntp` is `ntpd_configure_restrictions`, could you add platform ntp to the other ntp rules so that they are only run when it is installed?

yuumasato (commented):
Looks good, I just have one improvement suggestion.

controls/cis_rhel7.yml
vojtapolasek: fixed

yuumasato (approved):

PR #7114, opened 2021-06-14
New rsyslog rule per RHEL-08-010070 STIG

carlosmmatos:
#### Description:

New STIG rule for RHEL 8.


ggbecker (approved):
LGTM. Tests work flawless.

PR #7116, opened 2021-06-15
ensure_pam_module_options now fix empty option value

richardmaciel-canonical:
Given a pam module option, this regexp will fix its value even if
there is no value associated previously (but the option exists).

For instance, for the module pam_tally2, option deny, this commit will allow
this script to fix 'deny=' adding a proper value like 'deny=3'.

Signed-off-by: Richard Maciel Costa <richard.maciel.costa@canonical.com>

#### Description:
Small change to the regexp responsible of the sed whose purpose is to fix the value of the option with the parameter `_defaultValue`.

#### Rationale:
If a pam module line has an option with `option=`, the bash script will fail to fix that option value.


cipherboy (approved):

ggbecker (commented):
merging as approved by @cipherboy

PR #7119, opened 2021-06-15
Add single quotes to dconf value "lock-screen" (RHEL-08-020050)

ggbecker:
#### Description:

- Add single quotes to dconf value "lock-screen" (RHEL-08-020050)

#### Rationale:

- Value without the single quotes is not accepted and `dconf update` fails.

### DIFF
--- a/linux_os/guide/system/software/gnome/gnome_login_screen/dconf_gnome_lock_screen_on_smartcard_removal/rule.yml
+++ b/linux_os/guide/system/software/gnome/gnome_login_screen/dconf_gnome_lock_screen_on_smartcard_removal/rule.yml
@@ -7,12 +7,12 @@ title: 'Enable the GNOME3 Screen Locking On Smartcard Removal'
 description: |-
     In the default graphical environment, screen locking on smartcard removal
     can be enabled by setting <tt>removal-action</tt>
-    to <tt>lock-screen</tt>.
+    to <tt>'lock-screen'</tt>.
     <br /><br />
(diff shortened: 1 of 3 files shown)


cipherboy (approved):
Should we add type-checking to the template? If int(value) parses fine, don't use quotes, but if it doesn't parse and there aren't quotes, throw an error? 

Otherwise, LGTM.

ggbecker:
> Should we add type-checking to the template? If int(value) parses fine, don't use quotes, but if it doesn't parse and there aren't quotes, throw an error?

I don't think the dconf configuration follows this precise rule, for example [dconf_gnome_disable_user_list](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/software/gnome/gnome_login_screen/dconf_gnome_disable_user_list/rule.yml#L16) uses `true` without quotes and it's a valid case. I think it would become quite complicated, I'm not sure it's worth it.

> 
> Otherwise, LGTM.

cipherboy:
> I don't think the dconf configuration follows this precise rule, for example dconf_gnome_disable_user_list uses true without quotes and it's a valid case. I think it would become quite complicated, I'm not sure it's worth it.

Ah we'd have to do int/string/bool types, gotcha. :/ Yeah, fine for now.

I confirmed no other existing rules have this particular problem.

yuumasato (approved):

carlosmmatos (approved):
LGTM

PR #7120, opened 2021-06-15
OCP4: Use minute notation for inactivity timeout setting

JAORMX:
We were setting `600s`, which equals to 10min, however **something** is
"prettifying" the setting to "10m0s", which makes the test fails and
contributes to the flakiness we've been seeing.

This uses the `10m0s` notation instead to fix this.


jhrozek (approved):
LGTM

PR #7121, opened 2021-06-15
Fix STIG reference in profile file.

ggbecker:
#### Description:

- Fix STIG reference in profile file.


carlosmmatos (approved):
Don't know how we missed that, but good catch!

PR #7122, opened 2021-06-16
RHEL8 - Removing profiles that aren't needed.

carlosmmatos:
#### Description:

IMO - we can remove these profiles. I have not seen any asks around them, and I don't want to have profiles just laying around for no good reason.


carlosmmatos:
> I've no horse in this race, but these two profiles look very easy to create by tailoring.

I agree.. it's not something I want to be held responsible for if something doesn't work right, especially since I'm not aware of the origins or requests for these profiles.. I'll see if anyone else has some context, more so around the VPP profile, but the STIG one is a definite no-go for me since I know for a fact we haven't submitted/started any official process to get that done.

yuumasato (approved):
Green light from me.

PR #7124, opened 2021-06-17
GH Actions: Reintroduce independent AR linter for RHELs

evgenyz:


.github/workflows/build-lint-ansible-roles.yaml
```diff
@@ -0,0 +1,23 @@
+name: Gating (AR / RHEL)
+on:
+  push:
+    branches: [ '*' ]
+  pull_request:
+    branches: [ master ]
+jobs:
+  validate-ubuntu-ar:
+    name: Build, Lint Ansible Roles on Ubuntu Latest
+    runs-on: ubuntu-latest
```
yuumasato, line 10: It was previously running on `ubuntu-20.04`, could that make a difference?

evgenyz: Latest is 20.04 ATM

evgenyz: I believe that the problem is in the content (I've re-based the branch today)

yuumasato: Oh, :-P

ggbecker:
I could not reproduce the problem locally on Fedora 33.

evgenyz:
> I could not reproduce the problem locally on Fedora 33.

Huh. Well, I can try and switch it to containerized Fedora I guess

evgenyz:
BTW, I remember that we have some sort of `--quick` option to build the content, which would skip building some rarely needed stuff. Can I use it here?

ggbecker:
> BTW, I remember that we have some sort of `--quick` option to build the content, which would skip building some rarely needed stuff. Can I use it here?

It's `--datastream-only`, but in this case it wouldn't generate the required ansible playbooks if I'm not mistaken.

ggbecker:
> > BTW, I remember that we have some sort of `--quick` option to build the content, which would skip building some rarely needed stuff. Can I use it here?
> 
> It's `--datastream-only`, but in this case it wouldn't generate the required ansible playbooks if I'm not mistaken.

But we can build specifically the ansible playbooks target using cmake and ninja (for faster builds)

Install ninja:
```
# Fedora/RHEL
yum install ninja-build

# Ubuntu/Debian
apt-get install ninja-build
```
CMake for specific products and use ninja:
`cmake -DSSG_PRODUCT_DEFAULT=OFF -DSSG_PRODUCT_RHEL7=ON -DSSG_PRODUCT_RHEL8=ON -G "Ninja" ../`

Then build only the profile playbooks (it takes almost the same time as building only datastreams):
`ninja -j2 rhel8-profile-playbooks rhel7-profile-playbooks`

ggbecker (approved):

PR #7128, opened 2021-06-17
Fix OCP worker group name

mrogers950:
This addresses https://bugzilla.redhat.com/show_bug.cgi?id=1919660
@JAORMX @jhrozek


JAORMX (approved):
Oops! Thanks for the fix! Feel free to merge once CI passes

PR #7129, opened 2021-06-17
Add /var/log/audit individual ownership rules

ggbecker:
#### Description:

- Split [file_ownership_var_log_audit](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/auditing/auditd_configure_rules/file_ownership_var_log_audit/rule.yml) into four new rules but preserve the older one so it doesn't affect other profiles.

#### Rationale:

- STIG references:
https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_8_security_technical_implementation_guide/V-230397
https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_8_security_technical_implementation_guide/V-230398
https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_8_security_technical_implementation_guide/V-230399
https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_8_security_technical_implementation_guide/V-230400


linux_os/guide/system/auditing/auditd_configure_rules/directory_group_ownership_var_log_audit/tests/correct_value.pass.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+chgrp root /var/log/audit
```
jan-cerny, line 3: What about testing also the case when log_group in auditd.conf is not root?

JAORMX:
/retest

```
level=error
level=error msg=Error: error waiting for Route53 Hosted Zone (Z0642338JI7G7CUAH96U) creation: timeout while waiting for state to become 'INSYNC' (last state: 'PENDING', timeout: 15m0s)
level=error
level=error msg=  on ../tmp/openshift-install-604179952/route53/base.tf line 22, in resource "aws_route53_zone" "new_int":
level=error msg=  22: resource "aws_route53_zone" "new_int" {
level=error
level=error
level=fatal msg=failed to fetch Cluster: failed to generate asset "Cluster": failed to create cluster: failed to apply Terraform: failed to complete the change 
```

AWS Route53 issue... this is done as a pre-req to getting a cluster and is related to issues with the account itself. Not to this PR.

linux_os/guide/system/auditing/auditd_configure_rules/directory_group_ownership_var_log_audit/rule.yml
```diff
@@ -0,0 +1,37 @@
+documentation_complete: true
+
+prodtype: rhel8
+
+title: 'System Audit Directories Must Be Group Owned By Root'
+
+description: |-
+    All audit directories must be group owned by root user. By default, the path for audit log is <pre>/var/log/audit/</pre>.
```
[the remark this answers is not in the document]
ggbecker: Yeah, this behavior was never described in the original rule but probably at some point someone realized that the rule should accept that behavior but security policies haven't been updated accordingly.

linux_os/guide/system/auditing/auditd_configure_rules/directory_ownership_var_log_audit/rule.yml
```diff
@@ -0,0 +1,37 @@
...
+    nist: CM-6(a),AC-6(1),AU-9(4)
+    nist-csf: DE.AE-3,DE.AE-5,PR.AC-4,PR.DS-5,PR.PT-1,RS.AN-1,RS.AN-4
+    pcidss: Req-10.5.1
+    srg: SRG-OS-000057-GPOS-00027,SRG-OS-000058-GPOS-00028,SRG-OS-000059-GPOS-00029,SRG-OS-000206-GPOS-00084
+    isa-62443-2013: 'SR 2.1,SR 2.10,SR 2.11,SR 2.12,SR 2.8,SR 2.9,SR 5.2,SR 6.1'
+    isa-62443-2009: 4.2.3.10,4.3.3.3.9,4.3.3.5.8,4.3.3.7.3,4.3.4.4.7,4.3.4.5.6,4.3.4.5.7,4.3.4.5.8,4.4.2.1,4.4.2.2,4.4.2.4
+    cobit5: APO01.06,APO11.04,APO12.06,BAI03.05,BAI08.02,DSS02.02,DSS02.04,DSS02.07,DSS03.01,DSS05.04,DSS05.07,DSS06.02,MEA02.01
+    iso27001-2013: A.10.1.1,A.11.1.4,A.11.1.5,A.11.2.1,A.12.4.1,A.12.4.2,A.12.4.3,A.12.4.4,A.12.7.1,A.13.1.1,A.13.1.3,A.13.2.1,A.13.2.3,A.13.2.4,A.14.1.2,A.14.1.3,A.16.1.4,A.16.1.5,A.16.1.7,A.6.1.2,A.7.1.1,A.7.1.2,A.7.3.1,A.8.2.2,A.8.2.3,A.9.1.1,A.9.1.2,A.9.2.3,A.9.4.1,A.9.4.4,A.9.4.5
+    cis-csc: 1,11,12,13,14,15,16,18,19,3,4,5,6,7,8
+    stigid@rhel8: RHEL-08-030100
```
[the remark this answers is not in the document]
ggbecker: The rules are a subset of the original rules, so that's why I kept them.

linux_os/guide/system/auditing/auditd_configure_rules/directory_group_ownership_var_log_audit/tests/correct_value.pass.sh
ggbecker: That's a good idea... I think so far we haven't tested this case in other rules that behave the same.

linux_os/guide/system/auditing/auditd_configure_rules/directory_ownership_var_log_audit/rule.yml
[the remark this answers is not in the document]
jan-cerny: OK, thanks for explanation, it sounds fine.

linux_os/guide/system/auditing/auditd_configure_rules/directory_group_ownership_var_log_audit/rule.yml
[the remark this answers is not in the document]
jan-cerny: So do you want to add a description here?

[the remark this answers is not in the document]
ggbecker: I think it makes more sense to remove this criterion.

ggbecker:
@jan-cerny I've greatly redesigned the OVAL in some aspects and kept the non-root group check only for the group ownership rules and added to the rule description. I also added test scenarios to cover the case where a non-root group is used. I also noticed a problem with the extend definition for non-root group checking... the default value for `log_group` is root, so it must be set in order to have a different value other than `root`. The check doesn't check the actual group id. It just cares that is different than zero (root group id).

linux_os/guide/system/auditing/auditd_configure_rules/directory_ownership_var_log_audit/tests/correct_value.pass.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+chgrp root /var/log/audit
```
jan-cerny, line 3: Shouldn't there be `chown` instead because the rule is about user ownership not group ownership?

linux_os/guide/system/auditing/auditd_configure_rules/file_group_ownership_var_log_audit/oval/shared.xml
```diff
@@ -0,0 +1,46 @@
+<def-group>
+  <definition class="compliance" id="file_group_ownership_var_log_audit" version="1">
+    {{{ oval_metadata("Checks that all /var/log/audit files are group owned by the root user.") }}}
+    <criteria operator="OR">
+      <criteria operator="AND" comment="files are root group owned">
+        <criterion test_ref="test_group_ownership_var_log_audit_files" />
+      </criteria>
```
jan-cerny, line 7: The criteria element is superfluous because it contains just a signle criterion.

linux_os/guide/system/auditing/auditd_configure_rules/directory_group_ownership_var_log_audit/oval/shared.xml
```diff
@@ -0,0 +1,46 @@
...
+  <definition class="compliance" id="directory_group_ownership_var_log_audit" version="1">
+    {{{ oval_metadata("Checks that all /var/log/audit directories are group owned by the root user.") }}}
+    <criteria operator="OR">
+      <criteria operator="AND" comment="directories are root group owned">
+        <criterion test_ref="test_group_ownership_var_log_audit_directories" />
+      </criteria>
+      <criteria operator="AND" comment="log_group in auditd.conf is not root">
+        <extend_definition comment="log_group in auditd.conf is not root"
+        definition_ref="auditd_conf_log_group_not_root" />
+        <criterion test_ref="test_group_ownership_var_log_audit_directories-non_root" />
```
jan-cerny, line 11: In this scenario https://paste.centos.org/view/7e2744b8 the /var/log/audit is owned by evil people but the rule passes but I think it should fail.

cipherboy: In my mind, shouldn't we accept whatever `log_group` is specified in that file and have a separate rule to audit that `log_group` is acceptably either `root` or `adm`... But I dunno :)

jan-cerny: this way the pastebin is going to expire so I will just copy paste it here:

```
#!/bin/bash
 
groupadd group_test
groupadd evil_people
 
sed -i "/\s*log_group.*/d" /etc/audit/auditd.conf
echo "log_group = group_test" >> /etc/audit/auditd.conf
 
chgrp evil_people /var/log/audit
```

linux_os/guide/system/auditing/auditd_configure_rules/directory_ownership_var_log_audit/tests/correct_value.pass.sh
ggbecker: That's correct.

linux_os/guide/system/auditing/auditd_configure_rules/file_group_ownership_var_log_audit/oval/shared.xml
ggbecker: Ok, I'll remove it

linux_os/guide/system/auditing/auditd_configure_rules/directory_group_ownership_var_log_audit/oval/shared.xml
ggbecker: This is a copy paste from the previous implementation. I don't think OVAL would be powerful enough to do such check where it would detect exactly the group that is set in the log file.

I don't know how to proceed here... maybe we could remove this criterion since STIG does not describe it.

linux_os/guide/system/auditing/auditd_configure_rules/directory_group_ownership_var_log_audit/rule.yml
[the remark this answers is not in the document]
cipherboy: Just to mention this, but we talked about `log_group` on last week's CIS benchmark call. I think a future revision might take it into account.

ggbecker:
The only thing open to be resolved is the `log_group` being different than `root`. The check that allows other groups than `root` has been present for a long time and there must be reason for that. It doesn't really align with STIG, but there are many profile that are using similar rules, so changing other related rules would affect a significant amount of profiles and products.

cipherboy:
Just as a point of information, in the Ubuntu STIG we have:

```
Group ID (Vulid): V-238247
Group Title: SRG-OS-000057-GPOS-00027
Rule ID: SV-238247r653916_rule
Severity: CAT II
Rule Version (STIG-ID): UBTU-20-010124


Check Content: 
Verify the group owner is set to own newly created audit logs in the audit configuration file with the following command:

$ sudo grep -iw log_group /etc/audit/auditd.conf
log_group = adm

If the value of the "log_group" parameter is other than "root" or "adm", this is a finding. 
```

So IMO, it is within scope to have `log_group` and future CIS benchmarks (like I noted above) could include this change as well.

jan-cerny:
> The only thing open to be resolved is the log_group being different than root. The check that allows other groups than root has been present for a long time and there must be reason for that. It doesn't really align with STIG, but there are many profile that are using similar rules, so changing other related rules would affect a significant amount of profiles and products.

Is that something that needs to be resolved before merging?

jan-cerny:
@ggbecker please resolve the conflicts

ggbecker:
> @ggbecker please resolve the conflicts

Done

ggbecker:
> > The only thing open to be resolved is the log_group being different than root. The check that allows other groups than root has been present for a long time and there must be reason for that. It doesn't really align with STIG, but there are many profile that are using similar rules, so changing other related rules would affect a significant amount of profiles and products.
> 
> Is that something that needs to be resolved before merging?

I think it isn't. But the `log_group != root` behavior is still there although I don't think it can cause any serious trouble.

cipherboy:
@ggbecker @jan-cerny Per my call with DISA, they're going to add the `log_group=adm` alternative permissions to the Ubuntu STIG in the near future. CIS will take a little more time but will also allow alternative permissions for alternative log groups. HTH.

jan-cerny (approved):
LGTM and all test scenarios passed on a RHEL8 VM.

carlosmmatos (commented):
My only suggestion would be to add Ansible remediation if possible.. If I'm looking at the existing bash remediation, I would think the ansible remediation could be something like this:
```
- name: Check for existing value
  slurp:
    src: /etc/audit/auditd.conf
  register: log_group

- set_fact:
    log_group_value: "{{ log_group['content']| b64decode | regex_search(regexp,'\\1') }}"
  vars:
    regexp: 'log_group\s*=\s*(.*)'

- file:
    path: /var/log/audit
    group: "{{ log_group_value }}"
...
...
```

PR #7139, opened 2021-06-23
change vim package name to vim-common

vojtapolasek:
#### Description:

- the rule package_vim_installed uses package name of "vim", but this package actually does not exist. Change name to "vim-common".

#### Rationale:

The rule was reported as failing eventhough the application was installed.


ggbecker (approved):
LGTM. The failing test is unrelated to this PR.

JAORMX (approved):
rebasing should fix the issue.

ggbecker:
I'm inclined to just merge it right away as we know for sure what was the problem and it's unrelated.

cipherboy (commented):
I'm late to the party, but ACK from an Ubuntu side as well, it has both `vim` and `vim-common` there.

PR #7141, opened 2021-06-23
Increase /var/log/audit partition size to 10GB in STIG GUI KS file

mildas:
#### Description:
Related to #7088 - new rule in STIG profiles, however the partition size has been changed only in stig kickstart


ggbecker (approved):

ggbecker:
@mildas please rebase this with latest master and the problem with tests should disappear

mildas:
@ggbecker rebased

PR #7144, opened 2021-06-24
Add new Red Hat CCEs.

matejak:
Add 5000 new CCEs for Red Hat.

Those were requested from NIST using the cce@nist.gov e-mail address.


ggbecker (approved):
LGTM

PR #7147, opened 2021-06-24
Enable generic rules for RHEL9

matejak:
Those rules interact with components on a low-level basis, so they are likely to work on RHEL9.

See individual commits for more details.

Fixes:  #7147


matejak:
@Jakuje Could you please check out whether the smart card rules modification makes sense?

linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/force_opensc_card_drivers/rule.yml
```diff
@@ -1,11 +1,11 @@
...
+prodtype: fedora,ol7,ol8,rhel7,rhel8,rhel9,rhv4
 
 title: 'Force opensc To Use Defined Smart Card Driver'
 
 description: |-
-    The OpenSC smart card tool can auto-detect smart card drivers; however by
+    The OpenSC smart card middleware can auto-detect smart card drivers; however by
     forcing the smart card driver in use by your organization, opensc will no longer
     autodetect or use other drivers unless specified. This helps to prevent
     users from using unauthorized smart cards. The default smart card driver for this
```
Jakuje, line 11: on the following line, you have also the `/etc/opensc-<i>ARCH</i>.conf`, which is already fixed in the previous rule. Please, change it also here. I just checked and on normal x86_64 it should work on since RHEL7.

linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/configure_opensc_card_drivers/rule.yml
```diff
@@ -9,9 +9,8 @@ description: |-
     setting the smart card drivers in use by your organization helps to prevent
     users from using unauthorized smart cards. The default smart card driver for this
     profile is <tt>{{{ xccdf_value("var_smartcard_drivers") }}}</tt>.
-    To configure the OpenSC driver, edit the <tt>/etc/opensc-<i>ARCH</i>.conf</tt> (where
-    <i>ARCH</i> is the architecture of your operating system) file. Look for a
-    line similar to:
+    To configure the OpenSC driver, edit the <tt>/etc/opensc.conf</tt>.
+    Look for a line similar to:
     <pre># card_drivers = old, internal;</pre>
```
Jakuje, line 14: Actually, since RHEL 7.7 we ship minimal `opensc.conf` which does not have these example lines. Even though your remediation scripts still work, the instructions here no longer work. It should probably say "add the following line into the file in the "app default" block so it will look like
```
app default {
   [...]
   card_drivers = {{{ xccdf_value("var_smartcard_drivers") }}};
}
```
This is the case also for the other occurrences modifying this file.

The other option might be to use the same steps as in the scripts using `opensc-tool --set-conf-entry`.

linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/configure_opensc_card_drivers/rule.yml
```diff
@@ -47,8 +47,7 @@ ocil_clause: 'the smart card driver is not configured correctly'
 
 ocil: |-
     To verify that <tt>{{{ xccdf_value("var_smartcard_drivers") }}}</tt> is configured
-    as the smart card driver, run the following command changing <i>ARCH</i> for
-    the architecture of your operating system:
-    <pre>$ grep card_drivers /etc/opensc-<i>ARCH</i></pre>
+    as the smart card driver, run the following command:
+    <pre>$ grep card_drivers /etc/opensc.conf</pre>
     The output should return something similar to:
     <pre>card_drivers = {{{ xccdf_value("var_smartcard_drivers") }}};</pre>
```
Jakuje, line 53: The verification steps might also use the `opensc-tool --get-conf-entry` to make it less complicated. But that is not mandatory

linux_os/guide/system/software/integrity/fips/etc_system_fips_exists/rule.yml
```diff
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: fedora,rhcos4,ol7,ol8,rhel7,rhel8,rhv4,wrlinux1019
+prodtype: fedora,rhcos4,ol7,ol8,rhel7,rhel8,rhel9,rhv4,wrlinux1019
```
vojtapolasek, line 3: I see you added no CCE to this rule, any reason for that?

vojtapolasek (commented):
LGTM, just one question, see comment.

linux_os/guide/system/software/integrity/fips/etc_system_fips_exists/rule.yml
matejak: That rule is not in any profile, so it doesn't need a CCE. However, its check is used by a rule that is selected in a profile.

linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/configure_opensc_card_drivers/rule.yml
matejak: Thanks!

vojtapolasek (commented):
I ran ssg test suite on rules which have Fedora in their prodtype. I used Fedora 34. There are 2 rules which are problematic, see comments.
I suggest rather leaving them out now.

linux_os/guide/system/accounts/accounts-physical/disable_ctrlaltdel_reboot/rule.yml
```diff
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: fedora,ol7,ol8,rhcos4,rhel7,rhel8,rhv4,sle12,sle15,ubuntu2004,wrlinux1019
+prodtype: fedora,ol7,ol8,rhcos4,rhel7,rhel8,rhel9,rhv4,sle12,sle15,ubuntu2004,wrlinux1019
```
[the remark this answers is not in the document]
matejak: This looks very similar to #7248

The rule is probably OK, but our templated "service disabled" content doesn't work well with Fedora / RHEL9.

linux_os/guide/system/accounts/accounts-pam/set_password_hashing_algorithm/set_password_hashing_algorithm_logindefs/rule.yml
```diff
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: fedora,ol7,ol8,rhel7,rhel8,rhv4,sle12,sle15,ubuntu2004,wrlinux1019
+prodtype: fedora,ol7,ol8,rhel7,rhel8,rhel9,rhv4,sle12,sle15,ubuntu2004,wrlinux1019
```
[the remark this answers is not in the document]
matejak: This is a classical textfilecontent rule which should work anywhere. I will check it out.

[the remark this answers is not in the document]
matejak: So it turns out that the check on Fedora fails because the rule is not applicable according to `shared/checks/oval/installed_env_has_login_defs.xml` as it is under the `login_defs` platform.

vojtapolasek (approved):
Merging, found errors have to be solved outside this PR.

PR #7149, opened 2021-06-25
GH Actions: Trigger on push to master and stabilization-* branches, don't try to post to the PR (results in log)

evgenyz:


ggbecker (commented):
I would rather move the `content-test-filtering` job to a different file and trigger only in PRs. Here whenever something is merged to master, the job also gets executed and from my POV it's useless.

I think we only want it to be executed on PRs.

evgenyz:
> I would rather move the `content-test-filtering` job to a different file and trigger only in PRs. Here whenever something is merged to master, the job also gets executed and from my POV it's useless.
> 
> I think we only want it to be executed on PRs.

Sure, can do that

ggbecker:
You can also remove the text `Validate Ansible Roles` from this title since it doesn't do that anymore:

https://github.com/ComplianceAsCode/content/blob/5ba05c306b39273048ec08732bdeb8d5dc64b6ee/.github/workflows/build.yaml#L9

.github/workflows/build.yaml
```diff
@@ -46,14 +46,18 @@ jobs:
         with:
           apt-get: git python3-jinja2 python3-yaml python3-deepdiff python3-git python3-github python3-requests xmldiff
       # TODO: Use action's checkout along with --local and --repository options of ctf
+      # See: https://github.community/t/how-to-commit-to-two-branches-from-an-action/17713/4
       #- name: Checkout
-      #  uses: actions/checkout@v2
+      #  uses: actions/checkout@v1
       - name: Checkout (CTF)
         uses: actions/checkout@v2
```
vojtapolasek, line 53: The comment above talks about actions/checkout@v1, why v2 is here?

evgenyz: The checkout@v1 is the only way to get multiple branches at the same time (which will allow us to switch to action-only workflow for content filter).

.github/workflows/build.yaml
```diff
@@ -46,14 +46,18 @@ jobs:
...
         with:
           repository: mildas/content-test-filtering
           path: ctf
-      - name: Process
-        run: python3 ./ctf/content_test_filtering.py pr --output-format markdown ${{ github.event.pull_request.number }} > ctf.md
-      - name: Update the PR
-        run: python3 ./ctf/utility_scripts/comment_pr.py --token ${{ github.token }} --pr ${{ github.event.pull_request.number }} --comment ctf.md
+      - name: Process (see the output for recommended tests)
+        run: python3 ./ctf/content_test_filtering.py pr --output-format markdown ${{ github.event.pull_request.number }} # > ctf.md
```
ggbecker, line 58: I can't see the output in the logs of this line. I think you have to explicitly echo the output, something like this:

```suggestion
        run: python3 ./ctf/content_test_filtering.py pr --output-format markdown ${{ github.event.pull_request.number }} # > ctf.md
        id: ctf
      - name: CTF Output
        run: echo ${{ steps.ctf.outputs.stdout }}
```

evgenyz: As far as I know the tool won't output anything if there are no relevant changes detected. The change you proposed could be used to carry some output over to the next step (which you created), but it is not needed in general.

ggbecker: It makes sense. Maybe the CTF can be improved in this case to print at least some message that there is nothing to be tested, otherwise it can confuse people because it says look for the output but there is nothing. But I'm fine with this approach

ggbecker (approved):

PR #7151, opened 2021-06-25
New rules for RHEL-08-010290

carlosmmatos:
#### Description:

- This change adds 2 new rules, one to deal with the openssh.conf crypto file, and the other with the opensshserver.conf crypto file for using the appropriate MACs.


linux_os/guide/system/software/integrity/crypto/harden_sshd_macs_openssh_conf_crypto_policy/oval/shared.xml
```diff
@@ -0,0 +1,35 @@
...
+  </ind:textfilecontent54_object>
+
+  <ind:textfilecontent54_state id="ste_{{{ rule_id }}}" version="1">
+    <ind:text var_ref="sshd_macs_crypto" operation="equals"></ind:text>
+  </ind:textfilecontent54_state>
+
+  <local_variable id="sshd_macs_crypto" datatype="string" comment="The regex of the directive" version="1">
+    <concat>
+      <literal_component>MACs </literal_component>
+      <variable_component var_ref="sshd_approved_macs"/>
```
ggbecker, line 30: This OVAL check doesn't allow a subset of the algorithms. The implementation for the sshd_config file accounts for that and that's why we have decided to use a hardcoded value for the regex:

https://github.com/ComplianceAsCode/content/blob/602a7e283c9a03e56c1921b76312cd86c16478be/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_macs_ordered_stig/oval/shared.xml#L34

From [V-230251](https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_8_security_technical_implementation_guide/V-230251)
`
If the MAC entries in the "openssh.config" and "opensshserver.config" files have any hashes other than "hmac-sha2-512" and "hmac-sha2-256", the order differs from the example above, if they are missing, or commented out, this is a finding.
`
For example if someone configures something `MACs hmac-sha2-512` (which is valid from a STIG standpoint), the check will fail.

The only way to solve this is to have a specialized STIG rule, unfortunately. Unless we are ok with our check forcing to always require both values.

cipherboy: Could we hoist regex creation into a Python-based jinja function (see `add_python_functions` in `ssg/jinja.py`)? There's been enough times that we've wanted to do _something_ along these lines (perhaps ordered or unordered subset) that perhaps its time to expand the helpers and turn this into a safe-ish (either handling escaping or explicitly warning about bad characters) Python helper. 

Thoughts?

cipherboy: Concretely, in the case where we're comparing against a variable, we'd need to know the order of things in the variable and possible entries. Then you can use a regex to select each portion (splitting on a comma) across both file contents and variable contents and then drop them into a `set` and use `subset of` as the comparison operator? Just a thought.

ggbecker: We have discussed these possibilities in the past. The problem with variable in this case is that it requires ordering and OVAL cannot guarantee that. I think we cannot achieve this, but maybe I'm wrong.

carlosmmatos: > This OVAL check doesn't allow a subset of the algorithms. The implementation for the sshd_config file accounts for that and that's why we have decided to use a hardcoded value for the regex:
> 
> https://github.com/ComplianceAsCode/content/blob/602a7e283c9a03e56c1921b76312cd86c16478be/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_macs_ordered_stig/oval/shared.xml#L34
> 
> From [V-230251](https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_8_security_technical_implementation_guide/V-230251)
> `If the MAC entries in the "openssh.config" and "opensshserver.config" files have any hashes other than "hmac-sha2-512" and "hmac-sha2-256", the order differs from the example above, if they are missing, or commented out, this is a finding.`
> For example if someone configures something `MACs hmac-sha2-512` (which is valid from a STIG standpoint), the check will fail.
> 
> The only way to solve this is to have a specialized STIG rule, unfortunately. Unless we are ok with our check forcing to always require both values.

So the way I interpret this is: `Make sure the only value in MACs is <exactly what we tell you it should be (ie, hmac-sha2-512,hmac-sha2-256)`. This is why using the variable makes sense because we don't need to check for anything else other than what DISA wants to see. There is no subset check because that falls under DISA's `if they are missing - this is a finding` statement.

cipherboy: I hadn't known about the "the order differs from the example above" -- so I am OK with a strict equality check.

linux_os/guide/system/software/integrity/crypto/harden_sshd_macs_openssh_conf_crypto_policy/rule.yml
```diff
@@ -0,0 +1,60 @@
...
+    Overriding the system crypto policy makes the behavior of the OpenSSH
+    client violate expectations, and makes system configuration more
+    fragmented.
+
+severity: medium
+
+identifiers:
+    cce@rhel8: CCE-85870-4
+
+references:
```
Mab879, line 27: These need to be sorted. `utils/fix_rules.py sort_subkeys` from `master` should be helpful here.

Mab879 (changes requested):
Looks good, just need to order the rules to get CI to pass.

linux_os/guide/system/software/integrity/crypto/harden_sshd_macs_opensshserver_conf_crypto_policy/rule.yml
```diff
@@ -0,0 +1,60 @@
...
+
+rationale: |-
+    Overriding the system crypto policy makes the behavior of the OpenSSH
+    daemon violate expectations, and makes system configuration more
+    fragmented.
+
+severity: medium
+
+identifiers:
+    cce@rhel8: CCE-85871-2
```
Mab879, line 25: Looks like there is some issuse with CCEs still. The follow rules 
  - `linux_os/guide/system/permissions/files/file_permissions_etc_audit_auditd/rule.yml`
 - `linux_os/guide/system/software/integrity/crypto/harden_sshd_macs_opensshserver_conf_crypto_policy/rule.yml`

Mab879 (changes requested):

linux_os/guide/system/software/integrity/crypto/harden_sshd_macs_opensshserver_conf_crypto_policy/rule.yml
carlosmmatos: should be good now.. also updated some text to rename daemon with server..

Mab879 (approved):

PR #7152, opened 2021-06-28
OCP4: Fix kubelet cert/key matching rules

JAORMX:
The location for these changed in OCP 4.9, so this makes it portable.


jhrozek (approved):
LGTM,
code looks good, let's see how CI goes..

JAORMX:
Pull-request updated, HEAD is now 745b8a30772885fb98084bdd30bf00f5b5c4258a

JAORMX:
Missed a spot

PR #7154, opened 2021-06-28
modify platform and description of no_empty_passwords

vojtapolasek:
#### Description:

- make the rule no_empty_passwords applicable only to "machine" platform
- extend description to explain the reason

#### Rationale:

- the rule does not make much sense within container environment, because it should not be possible to login into the container through external means anyway (ssh)

- fixes https://bugzilla.redhat.com/show_bug.cgi?id=1856466
-


mildas (approved):

mildas:
LGTM

PR #7164, opened 2021-06-29
Add SLES-12-030362 to sysctl_net_ipv6_conf_default_accept_source_route

anivan-suse:
#### Description:

- SLES-12-030362

#### Rationale:

- sysctl_net_ipv6_conf_default_accept_source_route

### DIFF
--- a/products/sle12/profiles/stig.profile
+++ b/products/sle12/profiles/stig.profile
@@ -260,6 +260,7 @@ selections:
     - sysctl_net_ipv4_icmp_echo_ignore_broadcasts
     - sysctl_net_ipv4_ip_forward
     - sysctl_net_ipv4_tcp_syncookies
+    - sysctl_net_ipv6_conf_all_accept_redirects
     - sysctl_net_ipv6_conf_all_accept_source_route
     - sysctl_net_ipv6_conf_default_accept_redirects
     - vlock_installed
--- a/products/sle12/profiles/stig.profile
+++ b/products/sle12/profiles/stig.profile
@@ -260,8 +260,8 @@ selections:
     - sysctl_net_ipv4_icmp_echo_ignore_broadcasts
     - sysctl_net_ipv4_ip_forward
     - sysctl_net_ipv4_tcp_syncookies
-    - sysctl_net_ipv6_conf_all_accept_redirects
     - sysctl_net_ipv6_conf_all_accept_source_route
+    - sysctl_net_ipv6_conf_default_accept_source_route
     - sysctl_net_ipv6_conf_default_accept_redirects
     - vlock_installed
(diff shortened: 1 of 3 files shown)


products/sle12/profiles/stig.profile
```diff
@@ -260,6 +260,7 @@ selections:
     - sysctl_net_ipv4_icmp_echo_ignore_broadcasts
     - sysctl_net_ipv4_ip_forward
     - sysctl_net_ipv4_tcp_syncookies
+    - sysctl_net_ipv6_conf_all_accept_redirects
```
cipherboy, line 263: I think this line is superfluous in this PR. The changes are to forwarding and source_route... And forwarding isn't in this STIG profile and should probably be added. 

So I'd change this line to:

```yaml
 - sysctl_net_ipv6_conf_default_forwarding
```

Or drop it altogether. Note that it is breaking the build currently as-is. :-)

cipherboy (changes requested):

products/sle12/profiles/stig.profile
cipherboy: Thanks for the PR though! :)

cipherboy: Ah it was probably a bad copy-paste with #7165.

anivan-suse: Hey! Yes you are right, looks like I just made a little copy+paste mistake. :) I will fix it shortly, thanks for your help!

cipherboy:
@anivan-suse Any chance you could do a rebase instead of an explicit merge commit? :-) 

I think that'd be something like:

```bash
# Check out the branch this PR is from
git checkout SLES-12-030362

# Update objects from all remotes; this is the first step of "git pull" and does not update any local branches.
git fetch --all

# Then, rebase off of the new upstream master; assumes this repo (ComplianceAsCode/content) is named `upstream`; use `git remote -v` to show all remotes.
git rebase upstream/master

# If any conflicts occur, fix them up (`git status` will show the files that need to be fixed) and then continue with the rebase:
# git rebase --continue

# Finally, force-push the result up if/once successful. 
git push --force
```

Thanks!

cipherboy:
@JAORMX I don't think CI failure is related to this PR:

>     helpers.go:720: Result - Name: e2e-moderate-worker-sysctl-net-ipv6-conf-default-accept-source-route - Status: FAIL - Severity: medium
>     helpers.go:789: Rule sysctl_net_ipv6_conf_default_accept_source_route matched expected result 

There's a lot of other failing rules, making me suspect a template is wrong or something.

(Are you the right person to tag? :D)

I'm going ahead and merging.

cipherboy (approved):

PR #7167, opened 2021-06-29
Generate HTML STIG tables for stig_gui profile

vojtapolasek:
#### Description:

- modify build system to generate HTML STIG reference tables also for stig_gui profile

#### Rationale:

- HTML tables for stig_gui were missing


ggbecker (approved):
LGTM

PR #7174, opened 2021-06-30
Add `rule_dir` to rule rendering in test-rule-dir-json

JAORMX:
This is missing to be able to render some macros such as
`openshift_filtered_path`

### DIFF
--- a/utils/rule_dir_json.py
+++ b/utils/rule_dir_json.py
@@ -93,7 +93,9 @@ def handle_rule_yaml(product_list, product_yamls, rule_id, rule_dir, guide_dir):
     prod_type = product_list[0]
     product_yaml = product_yamls[prod_type]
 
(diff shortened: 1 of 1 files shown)


cipherboy:
Hey Juan! :-) Always nice to see people using some of the utils I've built. :-) If you don't mind me asking, what are you using it for?

There are some limitations with `rule_dir.json`, so perhaps it might be worth evaluating if we want to improve it?

ggbecker:
> Hey Juan! :-) Always nice to see people using some of the utils I've built. :-) If you don't mind me asking, what are you using it for?
> 
> There are some limitations with `rule_dir.json`, so perhaps it might be worth evaluating if we want to improve it?

This is actually connected to a bug fixed in our tests: https://github.com/ComplianceAsCode/content/pull/7176

:)

cipherboy:
Ahhhh ok :) No matter then!

PR #7180, opened 2021-06-30
Adding file_groupownership_system_commands_dirs to RHEL8 STIG

carlosmmatos:
#### Description:

- This rule supports STIG ID: RHEL-08-010320


ggbecker (approved):
LGTM

PR #7182, opened 2021-06-30
Enable templates for RHEL9

matejak:
Concerned templates are low-level, underlying components are stable.

I have updated some of the applicability metadata to include all RHELs.


cipherboy (approved):
No clue about the zipl ones, but I definitely agree on the auditd and mount options ones. Seem like sane changes. 

But a RHEL person should probably also review it. :-)

yuumasato (approved):
LGTM

PR #7184, opened 2021-07-01
Map SLES-12-010877 to file_permissions_system_commands_dirs rule

teacup-on-rockingchair:
#### Description:
- Map SLES-12-010877 to file_permissions_system_commands_dirs rule

#### Rationale:

- Add rule file_permissions_system_commands_dirs to sle12
- Made a patch in oval definition to make the regex more robust. Turns out the `?` qualifier does not behave ok on sle12,
  so changed to usage of `|` and tested on sle12 and sle15


vojtapolasek (approved):
LGTM, thank you

PR #7189, opened 2021-07-01
Expand valid values available for auditd num_logs setting

alexhaydock:
#### Description:

- This commit expands the valid values available to the `auditd_num_logs` variable.

#### Rationale:

- This variable expands the values available for `num_logs` in `/etc/audit/auditd.conf`. In conjunction with the `max_log_file` setting, this determines the overall auditd log storage of a system.
- Currently the default (and maximum) configurable value available is to keep `5` logs. When combined with the highest `max_log_file` setting of 20MB, the maximum retention for auditd logs that can be expressed by policy or tailoring file is capped at 100MB. There are many scenarios where it is desirable to keep more audit logs then this.
- I think keeping the highest value of 20MB for `max_log_file` is sensible, so this commit expands the number of these logs that can be kept, to increase the retention capabilities that an administrator can choose.
- With the new `num_logs` parameters added by this commit in conjunction with a `max_log_file` value of `20`, administrators can now choose to keep a maximum of 2GB of audit logs which, in my opinion, is much more reasonable.


JAORMX (approved):
LGTM

cipherboy (approved):

ggbecker (approved):

PR #7193, opened 2021-07-02
Split RHEL 7 CIS profile - section 5

vojtapolasek:
#### Description:

- update the policy file
- add or correct cis references in rule.yml files

#### Rationale:

CIS split effort


yuumasato (changes requested):
Made some general comments.
The only change I see as essential is the `pam_unix.so` with `remember`.

controls/cis_rhel7.yml
```diff
@@ -1543,3 +1543,483 @@ controls:
...
+    - set_password_hashing_algorithm_systemauth
+
+  - id: 5.4.4
+    title: Ensure password reuse is limited (Automated)
+    levels:
+    - l1_server
+    - l1_workstation
+    automated: yes
+    notes: |-
+      Usage of pam_unix.so module is not deprecated and is not supported by this policy interpretation.
```
yuumasato, line 1920: I think you meant that use of pam_unix.so with remember is deprecated.

yuumasato (commented):
LGTM
@vojtapolasek please rebase and fix the conflicts, :)

vojtapolasek:
Thank you for the review. I think we do not have rule for checking of SSH exchange key algorithms as you suggested in one of your comments.

yuumasato (approved):

PR #7194, opened 2021-07-02
OCP4: Enable rules to to check each API server audit logs sit on a separate partition

jhrozek:
On RHCOS4, we need to make sure there is enough space for logs, especially
audit logs. Let's makre sure that /var/log is set on a separate partition.

At the moment, the rule is manual (meaning, the OVAL check is disabled for
RHCOS), because the OVAL mount probe checks /proc which really doesn't work
well when running oscap-chroot.

Jira: CMP-963


jhrozek:
@carlosmmatos would like to get your opinion here. The rule currently just checks if `/var/log` itself is on a separate partition. Looking at other profiles, they also check if `/var/log/audit` is a separate partition, which seems to be in line with AC-4:

> The organization allocates audit record storage capacity in accordance with [Assignment: organization-defined audit record storage requirements].

My question is: Given that each of the OpenShift API servers (`kube-apiserver`, `openshift-apiserver` and `openshift-oauth-apiserver`) has its own audit log, should we have a total of 4 separate partitions under `/var/log` for each of the audit directories (`/var/log/audit`, '/var/log/kube-apiserver`, '/var/log/oauth-apiserver` and `/var/log/openshift-apiserver`) ? That seems a bit like an overkill, but in general it's doable. Or can we get away with just having a separate `/var/log` partition?

linux_os/guide/system/software/disk_partitioning/partition_for_var_log/rule.yml
```diff
@@ -42,7 +67,12 @@ references:
 
 platform: machine
 
+# (jhrozek): at the moment, the mount probe checks the /proc filesystem
+# even if openscap looks at a chroot, which doesn't allow to check for
+# mount points when checking RHCOS4 via CO
```
JAORMX, line 72: So...right now this is effectively just a MANUAL check, right? Let's add a simple e2e test that merely checks that this results `MANUAL`. This way we'll remember to add a relevant test once the probe works.

JAORMX (changes requested):

JAORMX:
> > The organization allocates audit record storage capacity in accordance with [Assignment: organization-defined audit record storage requirements].
> 
> My question is: Given that each of the OpenShift API servers (`kube-apiserver`, `openshift-apiserver` and `openshift-oauth-apiserver`) has its own audit log, should we have a total of 4 separate partitions under `/var/log` for each of the audit directories (`/var/log/audit`, '/var/log/kube-apiserver`, '/var/log/oauth-apiserver` and `/var/log/openshift-apiserver`) ? That seems a bit like an overkill, but in general it's doable. Or can we get away with just having a separate `/var/log` partition?

uhm... where do these directories come from? Are they created at run-time or do they come from some rpm package?

JAORMX (approved):

linux_os/guide/system/software/disk_partitioning/partition_for_var_log/rule.yml
jhrozek: Done. I asked @evgenyz on the chat if the probe really can't work in chroot, but looking at openscap code, I think it can't, it really pokes at /proc and checks the mountpoints with `getmntent`, that won't work in a chroot.

In the end I don't think that a manual check matters here that much given that creating the mount points can only be done when the cluster is installed.

JAORMX: @jhrozek well, the manual check will appear in the kube results and int he guide if they download the RAW ones. So it's relevant information that folks need to take into account.

jhrozek:
> > > The organization allocates audit record storage capacity in accordance with [Assignment: organization-defined audit record storage requirements].
> > 
> > 
> > My question is: Given that each of the OpenShift API servers (`kube-apiserver`, `openshift-apiserver` and `openshift-oauth-apiserver`) has its own audit log, should we have a total of 4 separate partitions under `/var/log` for each of the audit directories (`/var/log/audit`, '/var/log/kube-apiserver`, '/var/log/oauth-apiserver` and `/var/log/openshift-apiserver`) ? That seems a bit like an overkill, but in general it's doable. Or can we get away with just having a separate `/var/log` partition?
> 
> uhm... where do these directories come from? Are they created at run-time or do they come from some rpm package?

/var/log/audit is owned by `auditd` RPM. The others are `hostPath` k8s volumeMounts which don't specify any `type` or other options, that means that the path is created if it doesn't already exist.

JAORMX:
> > > > The organization allocates audit record storage capacity in accordance with [Assignment: organization-defined audit record storage requirements].
> > > 
> > > 
> > > My question is: Given that each of the OpenShift API servers (`kube-apiserver`, `openshift-apiserver` and `openshift-oauth-apiserver`) has its own audit log, should we have a total of 4 separate partitions under `/var/log` for each of the audit directories (`/var/log/audit`, '/var/log/kube-apiserver`, '/var/log/oauth-apiserver` and `/var/log/openshift-apiserver`) ? That seems a bit like an overkill, but in general it's doable. Or can we get away with just having a separate `/var/log` partition?
> > 
> > 
> > uhm... where do these directories come from? Are they created at run-time or do they come from some rpm package?
> 
> /var/log/audit is owned by `auditd` RPM.

lol that's the only directory that I'm sure about.

> The others are `hostPath` k8s volumeMounts which don't specify any `type` or other options, that means that the path is created if it doesn't already exist.

Not necessarily, the hostPaths would need to have a `DirectoryOrCreate` option set for them to be created [1], else the mount would fail. So I wonder if this is something that's created before starting the apiserver pods, or if it's included in the kubernetes RPM.

https://kubernetes.io/docs/concepts/storage/volumes/#hostpath

jhrozek:
On Fri, Jul 02, 2021 at 05:31:09AM -0700, Juan Osorio Robles wrote:
> > > > > The organization allocates audit record storage capacity in accordance with [Assignment: organization-defined audit record storage requirements].
> > > > 
> > > > 
> > > > My question is: Given that each of the OpenShift API servers (`kube-apiserver`, `openshift-apiserver` and `openshift-oauth-apiserver`) has its own audit log, should we have a total of 4 separate partitions under `/var/log` for each of the audit directories (`/var/log/audit`, '/var/log/kube-apiserver`, '/var/log/oauth-apiserver` and `/var/log/openshift-apiserver`) ? That seems a bit like an overkill, but in general it's doable. Or can we get away with just having a separate `/var/log` partition?
> > > 
> > > 
> > > uhm... where do these directories come from? Are they created at run-time or do they come from some rpm package?
> > 
> > /var/log/audit is owned by `auditd` RPM.
> 
> lol that's the only directory that I'm sure about.
> 
> > The others are `hostPath` k8s volumeMounts which don't specify any `type` or other options, that means that the path is created if it doesn't already exist.
> 
> Not necessarily, the hostPaths would need to have a `DirectoryOrCreate` option set for them to be created [1], else the mount would fail. So I wonder if this is something that's created before starting the apiserver pods, or if it's included in the kubernetes RPM.

No, they wouldn't I tested this before replying. If you leave the type
unspecified, e.g.:
```
  volumes:
    - name: varlog
      hostPath:
        path: /var/log/mydirectory
```

The the directory `/var/log/mydirectory` is created.

JAORMX:
@jhrozek aha! I misunderstood the docs then. Thanks for clarifying.

jhrozek:
```
    helpers.go:748: Result - Name: e2e-moderate-node-master-file-permissions-var-log-kube-audit - Status: INCONSISTENT - Severity: medium
    helpers.go:751: E2E-FAILURE: The expected result for the file_permissions_var_log_kube_audit rule didn't match. Expected 'map[master:PASS worker:NOT-APPLICABLE]', Got 'INCONSISTENT'
```
huh?

jhrozek:
FYI, I reached out to Steve Grubb for some opinions on the partitioning as well, will update the PR when I know more.

jhrozek:
> FYI, I reached out to Steve Grubb for some opinions on the partitioning as well, will update the PR when I know more.

See https://issues.redhat.com/browse/CMP-963 for the full discussion, but tl;dr we need a separate partition per audit log directory.

jhrozek:
Found a stupid typo, but otherwise this should be ready...
@mrogers950 Can you take a look and merge if you're OK with these patches, please?

mrogers950:
LGTM

PR #7199, opened 2021-07-05
OCP4: Add rule that checks cluster-wide proxy

JAORMX:


jhrozek (commented):
Code-wise looks good. I'm not sure we'll be able to do much in terms of automated  testing in CI, so I wonder if we should tag the rule as FAIL-ing by default?

JAORMX:
Pull-request updated, HEAD is now 483fdea8b761944aac60bf1f3dfd9366269cf539

JAORMX:
Pull-request updated, HEAD is now 13177b2fd6e9140249115e196b2bcd777ef15e80

PR #7203, opened 2021-07-06
OCP4: Add manual rule that indicates folks to remove the kubeadmin user

JAORMX:
We don't want to give permissions to the operator to view that secret,
but we can still indicate in the security guide that folks should delete
it.


JAORMX:
Pull-request updated, HEAD is now 20f1eecb064b0f66701a807373567bc767971c51

jhrozek (approved):
I was just typing a response asking you to add the OCIL clauses and while you refreshed the PR. Magic :-)
This version looks good

PR #6990, opened 2021-05-10
display_login_attempts - DISA STIG compliance

freddieRv:
#### Description:

- update oval to:
    - check there is no 'silent' option for pam_lastlog
    - check for 'required' control value

#### Rationale:

- As per DISA STIG, following line should be added to PAM's postlogin config file:
```
session required pam_lastlog.so showfailed
```
- Also check content states: _If "pam_lastlog" is missing from "/etc/pam.d/postlogin" file, or the **silent** option is present, this is a finding._
- [RHEL7 DISA STIG implementation guide reference](https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-204605)

PR #7204, opened 2021-07-06
Make some rules descriptions generic and add ubuntu specific bits

dodys:
#### Description:

- A few rules mentioned specific distros, but such rules can also be used by other distros. So replaced the distro's name with {{{ full_name }}}.
- Also added Ubuntu checks and specific texts.

### DIFF
--- a/linux_os/guide/system/software/updating/clean_components_post_updating/rule.yml
+++ b/linux_os/guide/system/software/updating/clean_components_post_updating/rule.yml
@@ -9,10 +9,17 @@ description: |-
     new versions have been installed. To configure <tt>{{{ pkg_manager }}}</tt> to remove the
     {{% if product in ["sle12", "sle15"] %}}
     previous software components after updating, set the <tt>solver.upgradeRemoveDroppedPackages</tt>
+    {{% elif 'ubuntu' in product %}}
+    previous software components after updating, set the <tt>::Remove-Unused-Dependencies</tt> and
+    <tt>::Remove-Unused-Kernel-Packages</tt>
     {{% else %}}
     previous software components after updating, set the <tt>clean_requirements_on_remove</tt>
     {{% endif %}}
+    {{% if 'ubuntu' in product %}}
+    to <tt>true</tt> in <tt>{{{ pkg_manager_config_file }}}</tt>.
+    {{% else %}}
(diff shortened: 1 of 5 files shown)


JAORMX (approved):

cipherboy (approved):

linux_os/guide/system/software/updating/clean_components_post_updating/rule.yml
```diff
@@ -58,6 +68,8 @@ ocil: |-
     <pre>$ grep -i upgradeRemoveDroppedPackages {{{ pkg_manager_config_file }}}</pre>
     The output should return something similar to:
     <pre>solver.upgradeRemoveDroppedPackages=true</pre>
+    {{% elif 'ubuntu' in product %}}
+    <pre>$ grep -i remove-unused {{{ pkg_manager_config_file }}}</pre>
```
yuumasato, line 72: This is will render the `ocil` for Ubuntu as just the grep command. Is that intentional?

dodys: Nice catch! 
I just pushed a fix for that.

yuumasato: The first sentence will still be the `grep` command, :grin: 

Just wanted to point out that the other `ocil` start with:
`To verify that <tt>...</tt> is configured properly, run the
    following command:`

yuumasato (approved):
LGTM.

ggbecker:
@cipherboy We are rewriting some parts of the `display_login_attempts` in #6990 and I noticed you guys added `ubuntu` in the rule description but I think you also should edit the OVAL, ansible, bash... if it configures in `/etc/pam.d/login`

Can you please check this?

cipherboy:
@ggbecker Not sure I follow. :-) Right now we haven't upstreamed any of our non-templated hardening changes -- mostly they reside in `ubuntu.sh` files (including for `display_login_attempts`) and thus aren't impacted by #6990 merging. #6990 already merged, so I'm guessing it is a little late for us to review it, but are you asking us to fix anything wrong (if anything is)?

Note that while we use the same path as SUSE (which seems to have regressed in #6990), we have different arguments (we still allow silent for instance).

HTH

PR #7207, opened 2021-07-06
Fix sshd service name in Ubuntu

dodys:
#### Description:

- sshd service name in Ubuntu is ssh


cipherboy (approved):

Note that `systemctl status sshd` works on Ubuntu due to linking, but the OVAL only works on original file names and fails to operate correctly on linked names.

yuumasato (approved):
LGTM.

@cipherboy Feel free to merge PRs if you think they are fine, and won't affect much contents from other products.

PR #7208, opened 2021-07-06
Rename install_smartcard_packages to package_smartcard_installed 

dodys:
#### Description:

- Some rules like smartcard_configure_ca, smartcard_configure_cert_checking and others, actually depend on install_smartcard_packages being executed before them. To guarantee that the order is actually respected renaming the rule seems to be enough and I was not able to reproduce the issue of rules executing before the renamed install_smartcard_packages.
- Fixed references to install_smartcard_packages
- Also added Ubuntu specific smartcard package


cipherboy (approved):

Thanks to the community on `#openscap` for pointing us in the right direction on this one. Hopefully all the changes are expected and I didn't miss anything obvious in other products when merging this internally.

yuumasato:
Renaming rules is something we should avoid doing so that we don't break tailorings out there in the community.

OTOH, this seems like a strong case for a rule that would benefit from a rename.

cipherboy:
Yeah, to me, its either rename the rule to align with our existing ordering logic or expand our ordering logic.

Ideally, I'd like to see proper ordering addressed (before/after) dependencies in rules themselves and first/last ordered lists for pre/post conditions on groups. But I don't have the time to go implement that properly, so perhaps we either merge this or merge the equivalent python-side change.

ggbecker:
> Renaming rules is something we should avoid doing so that we don't break tailorings out there in the community.
> 
I second that. I would prefer going with the equivalent python-side change.

cipherboy:
Closing this PR; per agreement here and in discussion with @dodys, I'll open a new PR with the python changes.

PR #7209, opened 2021-07-06
Update 'Configure Notification of Post-AIDE Scan Details' for RHEL8 DISA STIG Alignment

Mab879:
#### Description:

Added
- Ansible fix for this rule
- Configurable email for sending notification email for AIDE alerts

### DIFF
--- a/linux_os/guide/system/software/integrity/software-integrity/aide/aide_scan_notification/bash/shared.sh
+++ b/linux_os/guide/system/software/integrity/software-integrity/aide/aide_scan_notification/bash/shared.sh
@@ -1,6 +1,16 @@
 # platform = Red Hat Enterprise Linux 7,Red Hat Enterprise Linux 8,Red Hat Virtualization 4,multi_platform_wrlinux,multi_platform_ol,multi_platform_sle
 
+. /usr/share/scap-security-guide/remediation_functions
+
 {{{ bash_package_install("aide") }}}
+{{{ bash_instantiate_variables("var_aide_scan_notification_email") }}}
+{{% if product in ["sle12", "sle15"] %}}
+    {{% set aide_path = "/usr/bin/aide" %}}
+{{% else %}}
+    {{% set aide_path = "/usr/sbin/aide" %}}
+{{% endif %}}
+
+
 
 CRONTAB=/etc/crontab
 CRONDIRS='/etc/cron.d /etc/cron.daily /etc/cron.weekly /etc/cron.monthly'
@@ -14,11 +24,7 @@ if [ -f /var/spool/cron/root ]; then
 	VARSPOOL=/var/spool/cron/root
 fi
 
-if ! grep -qR '^.*\/usr\/sbin\/aide\s*\-\-check.*|.*\/bin\/mail\s*-s\s*".*"\s*root@.*$' $CRONTAB_EXIST $VARSPOOL $CRONDIRS; then
-{{% if product in ["sle12", "sle15"] %}}
-	echo '0 5 * * * root /usr/bin/aide  --check | /bin/mail -s "$(hostname) - AIDE Integrity Check" root@localhost' >> $CRONTAB
-{{% else %}}
-	echo '0 5 * * * root /usr/sbin/aide  --check | /bin/mail -s "$(hostname) - AIDE Integrity Check" root@localhost' >> $CRONTAB
-{{% endif %}}
+if ! grep -qR '^.*{{{aide_path}}}\s*\-\-check.*|.*\/bin\/mail\s*-s\s*".*"\s*.*@.*$' $CRONTAB_EXIST $VARSPOOL $CRONDIRS; then
+	echo "0 5 * * * root {{{ aide_path }}}  --check | /bin/mail -s \"\$(hostname) - AIDE Integrity Check\" $var_aide_scan_notification_email" >> $CRONTAB
 fi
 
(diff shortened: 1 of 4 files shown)


linux_os/guide/system/software/integrity/software-integrity/aide/var_aide_scan_notification_email.var
```diff
@@ -0,0 +1,16 @@
+documentation_complete: true
+
+title: Integrity Scan Notification Email Address
+
+description: |-
+    Specify the email address for designated personnel if baseline
```
cipherboy, line 6: "for notifications to designated personnel" perhaps?

cipherboy (commented):
I think this looks fairly good.

linux_os/guide/system/software/integrity/software-integrity/aide/var_aide_scan_notification_email.var
Mab879: Fixed, thanks.

linux_os/guide/system/software/integrity/software-integrity/aide/aide_scan_notification/rule.yml
```diff
@@ -57,3 +57,5 @@ ocil: |-
     <pre>$ grep aide /etc/crontab</pre>
     The output should return something similar to the following:
     <pre>05 4 * * * root /usr/sbin/aide --check | /bin/mail -s "$(hostname) - AIDE Integrity Check" root@localhost</pre>
+    The email address that the notifications are sent to can changed by overriding
```
ggbecker, line 60:
```suggestion
    The email address that the notifications are sent to can be changed by overriding
```

linux_os/guide/system/software/integrity/software-integrity/aide/aide_scan_notification/bash/shared.sh
```diff
@@ -1,6 +1,9 @@
 # platform = Red Hat Enterprise Linux 7,Red Hat Enterprise Linux 8,Red Hat Virtualization 4,multi_platform_wrlinux,multi_platform_ol,multi_platform_sle
```
ggbecker, line 2: We can remove the jinja if else by introducing a jinja variable for the `sle` difference. Note the only difference is that the aide path is `bin` and `sbin` in the other.

So something like
```
{{%if product in ["sle12", "sle15"] %}}
    {{% set aide_bin_path="bin" %}}
{{% else %}}
    {{% set aide_bin_path="sbin" %}}
{{% endif %}}
```

linux_os/guide/system/software/integrity/software-integrity/aide/aide_scan_notification/bash/shared.sh
```diff
@@ -14,11 +17,11 @@ if [ -f /var/spool/cron/root ]; then
 	VARSPOOL=/var/spool/cron/root
 fi
 
-if ! grep -qR '^.*\/usr\/sbin\/aide\s*\-\-check.*|.*\/bin\/mail\s*-s\s*".*"\s*root@.*$' $CRONTAB_EXIST $VARSPOOL $CRONDIRS; then
+if ! grep -qR '^.*\/usr\/sbin\/aide\s*\-\-check.*|.*\/bin\/mail\s*-s\s*".*"\s*.*@.*$' $CRONTAB_EXIST $VARSPOOL $CRONDIRS; then
```
ggbecker, line 20:
```suggestion
if ! grep -qR '^.*\/usr\/{{{aide_bin_path}}}\/aide\s*\-\-check.*|.*\/bin\/mail\s*-s\s*".*"\s*.*@.*$' $CRONTAB_EXIST $VARSPOOL $CRONDIRS; then
```

linux_os/guide/system/software/integrity/software-integrity/aide/aide_scan_notification/bash/shared.sh
```diff
@@ -14,11 +17,11 @@ if [ -f /var/spool/cron/root ]; then
...
 {{% if product in ["sle12", "sle15"] %}}
-	echo '0 5 * * * root /usr/bin/aide  --check | /bin/mail -s "$(hostname) - AIDE Integrity Check" root@localhost' >> $CRONTAB
+	echo "0 5 * * * root /usr/bin/aide  --check | /bin/mail -s \"\$(hostname) - AIDE Integrity Check\" $var_aide_scan_notification_email" >> $CRONTAB
```
ggbecker, line 22:
```suggestion
	echo "0 5 * * * root /usr/{{{aide_bin_path}}}/aide  --check | /bin/mail -s \"\$(hostname) - AIDE Integrity Check\" $var_aide_scan_notification_email" >> $CRONTAB
```

linux_os/guide/system/software/integrity/software-integrity/aide/aide_scan_notification/bash/shared.sh
cipherboy: Just a suggestion from the peanut gallery, but I would make this the full path. On Ubuntu we use`/usr/bin/aide.wrapper`. :-)

ggbecker: it can be, then it probably needs some extra work around the grep command, which uses escaped version of the path

cipherboy: Just curious, but why would grep need escaping? 

```
$ grep  '/var/log/audit'  linux_os/guide/system/auditing/auditd_configure_rules/logs_group_ownership_var_log_audit/bash/shared.sh 
    chown :adm /var/log/audit/
    sed -i -e '$alog_file = /var/log/audit/audit.log' /etc/audit/auditd.conf
```

grep has no problems with the `/` character and the other content there will prevent it from being interpreted as a path.

vojtapolasek (changes requested):
Thank you for the rule. It looks good. But I agree with others that the variable should be defined differently. It should either contain the absolut path to the directory where Aide executable is placed or maybe even whole absolute path including executable.

vojtapolasek:
LGTM, tests are passing, merging

vojtapolasek (approved):

PR #7210, opened 2021-07-06
SSGTS: Jinja enablement for test cases

cipherboy:
#### Description:

This enables Jinja templating during test suite build time (precisely, prior to building the product tarball) via writing test scenarios out to a temporary directory.

We do require one semi-breaking API change to identify the target product; see the first commit message for more details:

```
Add --product argument to SSGTS

test_suite.py previously had no knowledge of which product was being
evaluated (from a content perspective). While it had a datastream and an
operating system image and thus could theoretically reverse back to a
content product based on the contents of either, doing so would be
difficult. Adding a parameter allows the caller to override
autodetection from datastream file name (which doesn't necessarily
always work).

Parameter: --product
Argument: name of a content product (e.g., rhel8, ubuntu2004, &c)
Notes: Automatically inferred from adequately named datastream file
       (e.g., ssg-<...product...>-ds*.xml)

Doing so will allow us to enable jinja macros for test cases (based on
product environment content) and later enable templated tests.
```

In my view, as long as a build datastream is specified in the pipelines and they're not renamed, this shouldn't be much of an issue. 

#### Rationale:

Existing tests try to be product independent but sometimes depend on properties about the operating system. While we have the `# product = ...` list, sometimes the differences in operating systems are already known to the build system (such as package managers or grub2 configuration path) and are more easily worked around in Jinja rather than creating separate files for each potential product.

Note: I mostly tested this against `file_owner_etc_shadow` and updated a test case to include `{{{ full_name }}}`; I've not run a complete battery of profile/... or against RHEL systems, so I'd appreciate some review on this PR in that department.

Later PRs would include updating some of the grub tests to use the relevant macro. Eventually I'd like to also include support for templated test cases, ~~but I've not yet gotten to adding it~~ -- see #7211 for a draft of this. Let me know what you think!

### DIFF
--- a/tests/ssg_test_suite/common.py
+++ b/tests/ssg_test_suite/common.py
@@ -59,7 +59,7 @@ def walk_through_benchmark_dirs(product=None):
     directories = _BENCHMARK_DIRS
     if product is not None:
         yaml_path = product_yaml_path(SSG_ROOT, product)
-        product_base = os.path.basename(yaml_path)
+        product_base = os.path.dirname(yaml_path)
         product_yaml = load_product_yaml(yaml_path)
         benchmark_root = os.path.join(product_base, product_yaml['benchmark_root'])
         directories = [os.path.abspath(benchmark_root)]
@@ -276,7 +276,6 @@ def template_tests(product=None):
         product_yaml = dict()
         if product:
             yaml_path = product_yaml_path(SSG_ROOT, product)
-            product_base = os.path.basename(yaml_path)
             product_yaml = load_product_yaml(yaml_path)
 
         # Below we could run into a DocumentationNotComplete error. However,
--- a/tests/ssg_test_suite/common.py
+++ b/tests/ssg_test_suite/common.py
@@ -8,12 +8,17 @@ import functools
 import tarfile
 import tempfile
 import re
+import shutil
 
 from ssg.build_cpe import ProductCPEs
+from ssg.build_yaml import Rule as RuleYAML
 from ssg.constants import MULTI_PLATFORM_MAPPING
 from ssg.constants import FULL_NAME_TO_PRODUCT_MAPPING
 from ssg.constants import OSCAP_RULE
+from ssg.jinja import process_file
 from ssg.products import product_yaml_path, load_product_yaml
+from ssg.rules import get_rule_dir_yaml, is_rule_dir
+from ssg.rule_yaml import parse_prodtype
 from ssg_test_suite.log import LogHelper
 
 Scenario_run = namedtuple(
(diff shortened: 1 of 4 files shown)


jan-cerny:
I get this error when I try to test any rule from RHEL 8 product:

```
[jcerny@thinkpad scap-security-guide{pr/7210}]$ python3 tests/test_suite.py rule --libvirt qemu:///system ssgts_rhel8 accounts_tmout
INFO - Logging into /home/jcerny/work/git/scap-security-guide/logs/rule-custom-2021-07-07-1017/test_suite.log
ERROR - Nothing has been tested!
  File "/home/jcerny/work/git/scap-security-guide/tests/test_suite.py", line 387, in <module>
    main()
  File "/home/jcerny/work/git/scap-security-guide/tests/test_suite.py", line 383, in main
    options.func(options)
  File "/home/jcerny/work/git/scap-security-guide/tests/ssg_test_suite/rule.py", line 455, in perform_rule_check
    checker.test_target(options.target)
  File "/home/jcerny/work/git/scap-security-guide/tests/ssg_test_suite/oscap.py", line 654, in test_target
    self._test_target(target)
  File "/home/jcerny/work/git/scap-security-guide/tests/ssg_test_suite/rule.py", line 286, in _test_target
    self._prepare_environment(scenarios_by_rule)
  File "/home/jcerny/work/git/scap-security-guide/tests/ssg_test_suite/rule.py", line 244, in _prepare_environment
    self.remote_dir = common.send_scripts(self.test_env)
  File "/home/jcerny/work/git/scap-security-guide/tests/ssg_test_suite/common.py", line 400, in send_scripts
    archive_file = create_tarball(test_env.product)
  File "/home/jcerny/work/git/scap-security-guide/tests/ssg_test_suite/common.py", line 368, in create_tarball
    templated_tests = template_tests(product=product)
  File "/home/jcerny/work/git/scap-security-guide/tests/ssg_test_suite/common.py", line 358, in template_tests
    raise exp
  File "/home/jcerny/work/git/scap-security-guide/tests/ssg_test_suite/common.py", line 304, in template_tests
    rule = RuleYAML.from_yaml(rule_path, product_yaml)
  File "/home/jcerny/work/git/scap-security-guide/tests/../ssg/build_yaml.py", line 1131, in from_yaml
    if env_yaml and env_yaml["product"] in parse_prodtype(rule.prodtype):
KeyError: 'product'
```

It complains about missing `product` in `env_yaml` but in the place where it's called (/tests/ssg_test_suite/common.py, line 304) the `env_yaml` dict contains only a signle entry `cmake_build_type`. It seems that it doesn't load the yaml from a file it because `product` is `None` in that function.

cipherboy:
@jan-cerny Ahh good catch. I don't usually test with implicit datastream selection. Updated! :-)

tests/ssg_test_suite/common.py
```diff
@@ -48,8 +55,16 @@
...
 
-def walk_through_benchmark_dirs():
-    for dirname in _BENCHMARK_DIRS:
+def walk_through_benchmark_dirs(product=None):
+    directories = _BENCHMARK_DIRS
+    if product is not None:
+        yaml_path = product_yaml_path(SSG_ROOT, product)
+        product_base = os.path.basename(yaml_path)
+        product_yaml = load_product_yaml(yaml_path)
+        benchmark_root = os.path.join(product_base, product_yaml['benchmark_root'])
```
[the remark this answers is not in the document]
cipherboy: _Iiinteresting_, that doesn't manifest on my system. I do agree that does seem to be a bug and I've pushed a fix that works for me.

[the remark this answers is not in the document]
cipherboy: (Just for future reference, we determined this was cause by a difference in directories that the test suite was executed under.)

jan-cerny (approved):
It works great for me - I have tried rule mode, profile mode and also testing using the test_rules_in container.sh wrapper. Great PR!

matejak:
What was it with the `file_owner_etc_shadow` and `{{{ full_name }}}`? I can't see it in this PR neither in #7083

tests/ssg_test_suite/common.py
```diff
@@ -233,29 +248,155 @@ def _make_file_root_owned(tarinfo):
...
+        # error.
+        product_yaml['cmake_build_type'] = 'Debug'
+
+        # Note that we're not exactly copying 1-for-1 the contents of the
+        # directory structure into the temporary one. Instead we want a
+        # flattened mapping with all rules in a single top-level directory
+        # and all tests immediately contained within it. That is:
+        #
+        # /group_a/rule_a/tests/something.pass.sh -> /rule_a/something.pass.sh
+        for dirpath, dirnames, _ in walk_through_benchmark_dirs(product):
```
matejak, line 295: It would be great to extract contents of this for loop to a separate function or method.

tests/ssg_test_suite/common.py
```diff
@@ -233,29 +248,155 @@ def _make_file_root_owned(tarinfo):
...
+            # directory, recursively.
+            tests_dir_path = os.path.join(dirpath, "tests")
+            tests_dir_path = os.path.abspath(tests_dir_path)
+            for dirpath, dirnames, filenames in os.walk(tests_dir_path):
+                for dirname in dirnames:
+                    # We want to recreate the correct path under the temporary
+                    # directory. Resolve it to a relative path from the tests/
+                    # directory.
+                    dir_path = _rel_abs_path(os.path.join(dirpath, dirname), tests_dir_path)
+                    assert '../' not in dir_path
```
matejak, line 333: Please also include a message clarifying what went wrong when there is an assert outside of a block that handles `AssertionError`.

tests/ssg_test_suite/common.py
```diff
@@ -233,29 +248,155 @@ def _make_file_root_owned(tarinfo):
     return tarinfo
 
 
-def create_tarball():
+def _rel_abs_path(current_path, base_path):
```
matejak, line 251: What is the benefit of resolving the examined path to the absolute path before the comparison?

tests/ssg_test_suite/common.py
```diff
@@ -233,29 +248,155 @@ def _make_file_root_owned(tarinfo):
...
+            # Create the destination directory.
+            dest_path = os.path.join(tmpdir, rule.id_)
+            os.mkdir(dest_path)
+
+            # Walk the test directory, writing all tests into the output
+            # directory, recursively.
+            tests_dir_path = os.path.join(dirpath, "tests")
+            tests_dir_path = os.path.abspath(tests_dir_path)
+            for dirpath, dirnames, filenames in os.walk(tests_dir_path):
+                for dirname in dirnames:
```
matejak, line 328: Do we expect subdirectories in the `tests` directory? I couldn't find any.

cipherboy: On my checkout I see tons of `tests/ocp4/e2e.yml` subdirectories:

```
./linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_sudo/tests/ocp4/e2e.yml
./linux_os/guide/system/auditing/package_audit_installed/tests/ocp4/e2e.yml
./linux_os/guide/system/logging/log_rotation/ensure_logrotate_activated/tests/ocp4/e2e.yml
./linux_os/guide/system/selinux/selinux_policytype/tests/ocp4/e2e.yml
```

(total of 393 results). 

I have no idea if/how they're used but my thought was we might as well jinja enable them while we're here.

cipherboy:
@matejak Just as a heads up, this already merged. :-) I'm happy to pull any fixes into #7211 though. Regarding the `..._etc_shadow` rule, #7211 has real tests I want to ship for the `package_..._installed` rulesets so I'd take a look at those rather than what I did for testing purposes on this PR. HTH :-)

tests/ssg_test_suite/common.py
cipherboy: Sorry, could you expand on this? Are you asking why call `abspath` prior to `relpath`? My understanding was the supplied path must have a path with prefix `base_path`, so the `abspath` ensures that we have a correctly-rooted path structure.

The intent of this is to make copying into the tarball correct, with subdirectories (like `tests/ocp4/...`).

tests/ssg_test_suite/common.py
cipherboy: We can do that in #7211.

tests/ssg_test_suite/common.py
cipherboy: I don't really think we need this assert, given our other stuff.

tests/ssg_test_suite/common.py
cipherboy: Ah I see now. This will be addressed in #7211.

PR #7211, opened 2021-07-06
SSGTS: tests for shared/templates

cipherboy:
#### Description:

Follow-up to #7210. WIP; still ~~needs to be extended~~ ~~needs to be rebased once #7210 merges~~.

~~Currently test cases are added to the tarball that is uploaded, but I've not yet figured out where tests are listed for execution and updated that.~~

I've finished adding support for shared/templated content. Feel free to test it :-)

No major changes to APIs were required as part of this. The templating system does not require `tests` to be listed under the template languages (so as not to impact existing content building). 

The only rule that has been updated with tests is currently `package_installed` containing three comprehensive test cases (package installed, package removed, and package installed+removed).

#### Rationale:

Many rules use templated tests. The combination of product.yml and rule.yml frequently specifies helpful additional context when executing test cases (package name on a platform or package manager for instance). This PR adds support to the test system to use tests under `shared/templates/<template_name>/tests` and augment rule-specific tests with them.


cipherboy:
I think this is ready for feedback on the design &c. Obviously still needs to wait for #7210 to merge. :-)

```
$ ./test_suite.py --libvirt 'qemu:///system' 'ubuntu-20.04-cac-test' '--datastream' '../build/ssg-ubuntu2004-ds-1.2.xml' rule --profile standard package_audit_installed 
/home/cipherboy/GitHub/cipherboy/scap-security-guide
INFO - Logging into /home/cipherboy/GitHub/cipherboy/scap-security-guide/tests/logs/rule-custom-2021-07-07-0659/test_suite.log
INFO - xccdf_org.ssgproject.content_rule_package_audit_installed
INFO - Script package-installed-removed.fail.sh using profile xccdf_org.ssgproject.content_profile_standard OK
INFO - Script package-installed.pass.sh using profile xccdf_org.ssgproject.content_profile_standard OK
INFO - Script package-removed.fail.sh using profile xccdf_org.ssgproject.content_profile_standard OK
```

tests/ssg_test_suite/common.py
```diff
@@ -289,21 +499,65 @@ def iterate_over_rules():
...
+            # like the behavior in template_tests, this will overwrite any
+            # templated tests with the same file name.
             tests_dir = os.path.join(dirpath, "tests")
             tests_dir_files = os.listdir(tests_dir)
+            for test_case in tests_dir_files:
+                test_path = os.path.join(tests_dir, test_case)
+                if os.path.isdir(test_path):
+                    continue
+
+                all_tests[test_case] = process_file(test_path, local_env_yaml)
```
cipherboy, line 549: Forgot this the first time. :-) But I was able to use this to update some package install clauses in some rules, e.g.:

```diff
diff --git a/linux_os/guide/system/auditing/service_auditd_enabled/tests/disabled.fail.sh b/linux_os/guide/system/auditing/service_auditd_enabled/tests/disabled.fail.sh
index 158ffcfc39..a2f6f6c60e 100644
--- a/linux_os/guide/system/auditing/service_auditd_enabled/tests/disabled.fail.sh
+++ b/linux_os/guide/system/auditing/service_auditd_enabled/tests/disabled.fail.sh
@@ -1,5 +1,10 @@
 #!/bin/bash
+
+{{% if 'audit' in platform_package_overrides %}}
+# packages = {{{ platform_package_overrides['audit'] }}}
+{{% else %}}
 # packages = audit
+{{% endif %}}
 
 systemctl stop auditd
 systemctl disable auditd
```

And it works on Ubuntu. :-)

cipherboy:
@jan-cerny / @matejak This is now ready for review :-)

cipherboy:
@matejak Documentation around the rules have been added to this PR (covering both templated and regular Jinja macros). Please take a look and let me know what you think :-)

Achilleasein:
After a discussion with the rest of the reviewers we concluded that there are some positives and some negatives to the proposed PR:

**Pros:**
1. Some of the code is really good code and could be reused.
2. With the written tests you can easily spot errors in the definition of the rules written in the yaml, for example a typo.

**Cons:**
1. The first point that can be made is that with this implementation there are redundant tests running, since there is extra code generation and is testing similar scenarios.
2. The second point is that the generation happens from the ruleset which means from the same code of the remediations making the whole idea a bit pointless. It’s not providing any new features.
3. To further drive the point in, we could achieve the same result with a simple txt comparison, since we have the expected values.
4. Even if we would go with the approach, there would be caveats - not all test scenarios are the same for all rules templated by the same template. E.g. look at sudo package test scenarios and other package rules' test scenarios, here is a [link](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/dhcp/disabling_dhcp_server/service_dhcpd_disabled/tests/enabled.fail.sh).

A few errors were found, 1 major and 1 minor:
1. Yum remove will fail and needs to be changed to rpm remove, it is just an edge case for sudo package, still needs correcting.
2. On tests/ssg_test_suite/common.py on line 513 the variable is declared again, probably could be squished with line 339.

If you want we can set up a call to further discuss the PR with the rest of the reviewing team.
Reply to the latest Alex’s question in the PR.
I would also like to say that I’m sorry for the late delayed reply, but I just started working on the project and needed some time to figure everything out. I will be at your disposal for anything else you may need!

cipherboy:
Thanks for the detailed reply, @Achilleasein!

I guess I have a few questions about your cons:



> 1. The first point that can be made is that with this implementation there are redundant tests running, since there is extra code generation and is testing similar scenarios.


> 2. The second point is that the generation happens from the ruleset which means from the same code of the remediations making the whole idea a bit pointless. It’s not providing any new features.

> 3. To further drive the point in, we could achieve the same result with a simple txt comparison, since we have the expected values.

This one I really don't understand. :-) 

The point of the test cases, as I understand it, are to help validate the OVAL/Hardening actually _applies_ on a system and executes correctly (whereby, in a failing scenario, the OVAL correctly fails, then the hardening applies correctly resulting in a pass ultimately,   and, in a passing scenario, the OVAL correctly passes). This can't simply be checked with a simple text comparison and, IMO, needs a running system (whether that's a container or VM or something else). Where would that comparison be? During build time to ensure rule X has a template with specified variables? How do you maintain RHEL vs SUSE vs ... differences in that text comparison? &c.

The benefit of SSGTS is that you can run it on the desired platform and ensure e.g. all packages or services which are claimed to exist (or paths or anything else) actually do. :-)

> 4. Even if we would go with the approach, there would be caveats - not all test scenarios are the same for all rules templated by the same template. E.g. look at sudo package test scenarios and other package rules' test scenarios, here is a link.

Of course :-) That's why non-templated tests (rule directory tests) take priority over templated tests. And, furthermore, why we already have a mechanism for selecting features of a template (er, languages) to enable/disable. 

If you're looking for more customizability, I'd suggest that we could use a locally sourced function for setup/teardown and turn it into a common pattern.

E.g., in `<rule_id>/tests/shared.sh`:

```
ssgts_setup() {
if [ ! -e /etc/dhcp/ ]; then
  return 0
fi

cat << EOF >> /etc/dhcp/dhcpd.conf
subnet 192.168.122.0 netmask 255.255.255.248 {
}
EOF
}
```

And have all tests eventually look something like:

```sh
#!/bin/bash

ssgts_setup

## ACTUAL BODY OF TEST HERE

ssgts_teardown
```

shared/templates/package_installed/tests/package-installed-removed.fail.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+{{{ bash_package_install(PKGNAME) }}}
```
cipherboy, line 3: If the comment about redundancy was here, note that Debian systems can differentiate between a package that has never been installed versus one that has been installed before. We want to make sure that the OVAL correctly detects that the package is still removed and doesn't e.g., behave incorrectly on packages that were previously installed.

Note that we actually had this bug in the package installed bash check before: `dpkg -L | grep <name>` is NOT sufficient of a check for whether a package is present because removed packages will show up in that list. That's how we ended up with the more [complex query presently](https://github.com/ComplianceAsCode/content/blob/master/shared/macros-bash.jinja#L675):

```bash

{{%- macro bash_package_installed(pkgname) -%}}
{{%- if pkg_manager == "apt_get" -%}}
dpkg-query --show --showformat='${db:Status-Status}\n' "{{{ pkgname }}}" 2>/dev/null | grep -q installed
{{%- else -%}}
rpm --quiet -q "{{{ pkgname }}}"
{{%- endif -%}}
{{%- endmacro -%}}
```

cipherboy:
@Achilleasein Also you mentioned:

> If you want we can set up a call to further discuss the PR with the rest of the reviewing team. 

I had talked with @matejak (perhaps before his PTO) about setting up a call for the usual time on Wednesday, July 28th, but this hasn't gotten scheduled yet. That'd be as good of time as any, from my PoV :-)

Achilleasein (commented):
> I had talked with @matejak (perhaps before his PTO) about setting up a call for the usual time on Wednesday, July 28th, but this hasn't gotten scheduled yet. That'd be as good of time as any, from my PoV :-)

I believe the best approach is to have a call and discuss this and then post the decisions here to make it easier for everyone to follow. Will contact you as soon as possible when he returns.
If that is delayed or you're in a hurry I can elaborate here and we can discuss this further.

tests/ssg_test_suite/common.py
```diff
@@ -248,15 +253,142 @@ def _make_file_root_owned(tarinfo):
...
+    # so they'll be overwritten later if necessary.
+    if rule.template and rule.template['vars']:
+        templated_tests = template_builder.get_all_tests(rule.id_, rule.template,
+                                                         local_env_yaml)
+
+        for relative_path in templated_tests:
+            test_content = templated_tests[relative_path]
+            write_rule_templated_tests(dest_path, relative_path, test_content)
+
+    write_rule_dir_tests(local_env_yaml, dest_path, dirpath)
```
cipherboy, line 391: Needs to be conditional on `tests/` directory existing.

tests/ssg_test_suite/common.py
```diff
@@ -297,61 +423,7 @@ def template_tests(product=None):
             if "tests" not in dirnames or not is_rule_dir(dirpath):
```
cipherboy, line 423: This needs to be removed for rules with template-only tests.

tests/ssg_test_suite/common.py
```diff
@@ -430,21 +502,65 @@ def iterate_over_rules(product=None):
...
+    # Begin by loading context about our execution environment, if any.
+    product_yaml = get_product_context(product)
+
+    # Initialize a mock template_builder.
+    empty = "/ssgts/empty/placeholder"
+    template_builder = ssg.templates.Builder(product_yaml, empty,
+                                             _SHARED_TEMPLATES, empty, empty)
+
     for dirpath, dirnames, filenames in walk_through_benchmark_dirs(product):
         if "rule.yml" in filenames and "tests" in dirnames:
```
cipherboy, line 522: As above; `tests/` may not exist with templated tests, needs to be removed.

tests/ssg_test_suite/common.py
```diff
@@ -430,21 +502,65 @@ def iterate_over_rules(product=None):
...
+            # templating system.
+            all_tests = dict()
+
+            # Start
+            if rule.template and rule.template['vars']:
+                templated_tests = template_builder.get_all_tests(
+                    rule.id_, rule.template, local_env_yaml)
+                all_tests.update(templated_tests)
+
+            # Add additional tests from the local rule directory. Note that,
```
cipherboy, line 542: Conditional on it existing.

cipherboy (commented):
Misc. changes necessary.

tests/ssg_test_suite/common.py
cipherboy: (in #7295 we provide a better macro for this).

cipherboy:
@Achilleasein @dahaic This has the option now that we wanted. Default is to **not** run rules with the same template, but there's an option for full coverage: 

```
$ ssgts rule --profile cis_level1_server package_audit_installed package_rsyslog_installed
/home/cipherboy/GitHub/cipherboy/scap-security-guide
INFO - Logging into /home/cipherboy/GitHub/cipherboy/scap-security-guide/tests/logs/rule-custom-2021-07-29-0745/test_suite.log
INFO - xccdf_org.ssgproject.content_rule_package_rsyslog_installed
INFO - Script package-installed-removed.fail.sh using profile xccdf_org.ssgproject.content_profile_cis_level1_server OK
INFO - Script package-installed.pass.sh using profile xccdf_org.ssgproject.content_profile_cis_level1_server OK
INFO - Script package-removed.fail.sh using profile xccdf_org.ssgproject.content_profile_cis_level1_server OK
$ ssgts rule --duplicate-templates --profile cis_level1_server package_audit_installed package_rsyslog_installed
/home/cipherboy/GitHub/cipherboy/scap-security-guide
INFO - Logging into /home/cipherboy/GitHub/cipherboy/scap-security-guide/tests/logs/rule-custom-2021-07-29-0747/test_suite.log
INFO - xccdf_org.ssgproject.content_rule_package_rsyslog_installed
INFO - Script package-installed-removed.fail.sh using profile xccdf_org.ssgproject.content_profile_cis_level1_server OK
INFO - Script package-installed.pass.sh using profile xccdf_org.ssgproject.content_profile_cis_level1_server OK
INFO - Script package-removed.fail.sh using profile xccdf_org.ssgproject.content_profile_cis_level1_server OK
INFO - xccdf_org.ssgproject.content_rule_package_audit_installed
ERROR - Script package-installed-removed.fail.sh using profile xccdf_org.ssgproject.content_profile_cis_level1_server found issue:
ERROR - Rule xccdf_org.ssgproject.content_rule_package_audit_installed has not been evaluated! Wrong profile selected in test scenario?
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_package_audit_installed'.
ERROR - Script package-installed.pass.sh using profile xccdf_org.ssgproject.content_profile_cis_level1_server found issue:
ERROR - Rule xccdf_org.ssgproject.content_rule_package_audit_installed has not been evaluated! Wrong profile selected in test scenario?
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_package_audit_installed'.
ERROR - Script package-removed.fail.sh using profile xccdf_org.ssgproject.content_profile_cis_level1_server found issue:
ERROR - Rule xccdf_org.ssgproject.content_rule_package_audit_installed has not been evaluated! Wrong profile selected in test scenario?
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_package_audit_installed'.
```

(Turns out package_audit_installed is CIS level 2, but I think it illustrates the point just as well).

dahaic (approved):
Thank you, from the usability perspective it looks great!

Achilleasein:
Working as intended and the added functionality is as discussed. Just merged it.

PR #7216, opened 2021-07-07
New rule for RHEL-08-010350

Mab879:
#### Description:

New rule to check the group ownership of library directories and files.


ggbecker:
I think this open PR is proposing the same rule: https://github.com/ComplianceAsCode/content/pull/7179 we can combine them together

Mab879:
This rule is slightly different. We need to check files not the directories like in #7179, per the STIG.

linux_os/guide/system/permissions/files/permissions_within_important_dirs/file_groupownership_library_dirs/tests/correct_permissions.pass.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+for LIBDIR in /lib/ /lib64/ /usr/lib/ /usr/lib64/
+do
+   if [[ ! -f $LIBDIR ]]
+   then
+     touch $LIBDIR/LIBDIR
```
vojtapolasek, line 7: Maybe this might be clearer?
```suggestion
     touch $LIBDIR/testfile
```

vojtapolasek (changes requested):
Hello, thank you for the rule. Please see comments. Most of them are just cosmetical. I am just not understanding how the test case correct_permissions.pass.sh should exactly work.

Mab879:
Closing this one out in favor of #7231.

PR #7218, opened 2021-07-08
Enable SSGTS into Github Actions

ggbecker:
#### Description:

- Enable SSGTS in GH actions.
- This test uses content-test-filtering as input for which rules and remediation types should be tested. The product built is also provided by the CTF (content-test-filtering).
- Logs are uploaded in case of `ERROR` during tests execution. These logs can be downloaded and inspected locally.
- Content is built in a fedora:latest container because it contains latest version of OpenSCAP
- The rules are tested using the run_rule_in_container.sh script which overrides the cpe so it can be tested in the fedora container.
- Note: the tests are skipped in case there are no changes detected in the pull request. That's the case here, I've tested this PR with changes to the content but for the merge we need to keep it clean.

### DIFF
--- a/.github/workflows/ssgts.yaml
+++ b/.github/workflows/ssgts.yaml
@@ -20,17 +20,26 @@ jobs:
         with:
           repository: mildas/content-test-filtering
           path: ctf
-      - name: Process
+      - name: Detect content changes in the PR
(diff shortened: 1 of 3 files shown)


vojtapolasek (approved):
Great job, thank you for this. I tested the functionality on this PR:
https://github.com/ComplianceAsCode/content/pull/7300
Merging.

PR #7223, opened 2021-07-08
Add Ubuntu to cron.allow, at.allow rules for CIS

cipherboy:
#### Description:

I missed that `file_owner_cron_allow` and `file_groupowner_cron_allow` already existed in the project when I created the initial CIS profiles. I've also included the missing `file_permissions_cron_allow` rule and the corresponding `at.allow` equivalents of the `cron.allow` rules since they're all templated content.

I'm guessing other distros' CIS benchmarks include the same rules, so these should be generally useful to the community.

### DIFF
--- a/linux_os/guide/services/cron_and_at/restrict_at_cron_users/file_groupowner_at_allow/rule.yml
+++ b/linux_os/guide/services/cron_and_at/restrict_at_cron_users/file_groupowner_at_allow/rule.yml
@@ -1,0 +1,30 @@
+documentation_complete: true
+
+prodtype: ubuntu2004
+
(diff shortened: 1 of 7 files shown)


cipherboy:
Tagging @brett060102 and @vojtapolasek as they have contributed to SLE and RHEL CIS rules respectively and probably will want to add this to those distros after it merges.

richardmaciel-canonical (approved):
LGTM

jan-cerny:
@cipherboy lgtm but please resolve the conflicts

cipherboy:
Thanks @jan-cerny, I've rebased this :-)

jan-cerny (approved):

PR #7224, opened 2021-07-08
Prioritize install_smartcard_packages like package_*_installed

cipherboy:
#### Description:

Several smartcard rules have the problem that install_smartcard_packages
is not appropriately ordered. In #7208, we renamed the rule to match the
pattern of package_*_installed. However, this approach breaks any
existing tailoring and thus wasn't desired. Move to updating the Python
to order this rule as a special case.


matejak (approved):
LGTM, this PR fixes an issue when a "package is installed" rule escaped the already-existing ordering, because it was slightly aside from our templated package-installed rules.

PR #7226, opened 2021-07-09
enable levels to override variables of inheriting levels

vojtapolasek:
#### Description:

- modify build system
- while getting list of controls of a certain level, make sure that only the top-most level variable definition is used

#### Rationale:

- this modification allows for example CIS level 2 profile to have a different variable than CIS level 1 profile

### DIFF
--- a/ssg/controls.py
+++ b/ssg/controls.py
@@ -198,9 +198,7 @@ class ControlsManager():
             raise ValueError(msg)
         return policy
 
-    def get_all_controls_of_level(self, policy_id, level_id, override_vars=True):
-        # if override_vars is enabled, then variables from higher levels will
-        # override variables defined in controls of lower levels
+    def get_all_controls_of_level(self, policy_id, level_id):
         policy = self._get_policy(policy_id)
         levels = policy.get_level_with_ancestors(level_id)
         # we use OrderedDict here with empty values instead of ordered set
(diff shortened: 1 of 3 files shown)


alexhaydock:
I don't really have the deep level of CaC understanding to work out exactly how this applies but just to check - will variables set in tailoring files still take precedence even when this change is merged?

cipherboy:
@alexhaydock Tailoring files are applied at the XCCDF level. You can think of CaC as one big build system to generate XCCDFs (and other artifacts). As long as @vojtapolasek doesn't rename the variables (unlikely), there's no impact to tailoring files. This just changes how the Controls format (a CaC-ism used to generate Profiles) selects the default value of a variable at a given level.

ssg/controls.py
```diff
@@ -202,9 +202,16 @@ def get_all_controls_of_level(self, policy_id, level_id):
 
         all_policy_controls = self.get_all_controls(policy_id)
         eligible_controls = []
-        for c in all_policy_controls:
-            if len(level_ids.intersection(c.levels)) > 0:
-                eligible_controls.append(c)
+        defined_variables = []
+        # we will go level by level, from top to bottom
+        # this is done to enable overriding of variables by higher levels
+        for lv in level_ids:
```
yuumasato, line 208: `level_ids` is a `set`, so we cannot assume any particular order over the IDs.

Additionally, we cannot assume that the levels returned by `get_level_with_ancestors` is ordered, it also returns the levels in a `set`.

yuumasato (changes requested):
I gave it quick try with ANSSI profiles and it doesn't work.
I took a random variable from the profile, included it in a control from each level and assigned each one a different  selector. The variables came assigned with incorrect values, and each time I built the content the assignments varied.

Maybe we need to convert those sets to a list? I don't know what else would be impacted.

It would be good to have tests for this as well.

vojtapolasek:
Hello, I updated the PR. I replaced set with ordereddict. I also added a new functionality which can explicitly enable level overriding. I am not sure if the functionality is in the right part of the build system.
@yuumasato  @matejak  @cipherboy  what do you think?

vojtapolasek:
I have updated the PR. I made overriding of variables hardcoded and added some tests. @yuumasato  @jan-cerny  what do you think now?

yuumasato (changes requested):

vojtapolasek:
I think I fixed it.

ssg/controls.py
```diff
@@ -198,13 +201,32 @@ def _get_policy(self, policy_id):
...
         eligible_controls = []
-        for c in all_policy_controls:
-            if len(level_ids.intersection(c.levels)) > 0:
-                eligible_controls.append(c)
+        defined_variables = []
+        # we will go level by level, from top to bottom
+        # this is done to enable overriding of variables by higher levels
+        for lv in levels.keys():
+            for c in all_policy_controls:
+                if lv.id in c.levels:
```
matejak, line 211: This `if` can be simplified by `if lv.id not in c.levels: continue` - one level of indentation less.

The following code is not complex, but it needs extraction into meaningfully named private functions. Right now, it is very difficult to guess what are we doing why.

yuumasato (approved):
LGTM, thank you for the enhancement, :)

matejak:
I like the test, and I believe that the code works, but I would like to see some code readability improvements. As the original author won't be available for some time, I can tackle that next week.

matejak:
I have made following adjustments:

- Renamed get_level_with_ancestors to get_level_with_ancestors_sequence, and made it return a list instead - a dictionary result is quite confusing.
- Removed some optimization in the variable deletion loops.
- Extracted functionality to a _get_control_without_variables static method.
- Defined variable removal steps using set operations.

yuumasato (approved):
@matejak Thank you for the improvements.
Looks great now.

PR #7227, opened 2021-07-09
Add oval check for bios_enable_execution_restrictions

Mab879:
#### Description:

Add oval check for bios_enable_execution_restrictions to if `nx` in the flags in `/proc/cpuinfo`

#### Rationale:

For DISA STIG RHEL-08-010420

### DIFF
--- a/linux_os/guide/system/permissions/restrictions/enable_nx/bios_enable_execution_restrictions/oval/shared.xml
+++ b/linux_os/guide/system/permissions/restrictions/enable_nx/bios_enable_execution_restrictions/oval/shared.xml
@@ -1,8 +1,10 @@
 <def-group>
     <definition class="compliance" id="bios_enable_execution_restrictions" version="2">
         {{{ oval_metadata("The NX (no-execution) bit flag should be set on the system.") }}}
-        <criteria>
+        <criteria operator="AND">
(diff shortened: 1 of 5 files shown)


cipherboy (changes requested):
In order to address CIS 1.5.2 (1.6.2 in my copy of the RHEL7 CIS benchmark v3.0.1) -- across many other distros including Ubuntu and I'm guessing SUSE as well, we need a check for whether or not the protection is active on boot. 

CIS suggests the following three-part audit:

```bash
# [[ -n $(grep noexec[0-9]*=off /proc/cmdline) || -z $(grep -E -i ' (pae|nx) ' /proc/cpuinfo) || -n $(grep '\sNX\s.*\sprotection:\s' /var/log/dmesg | grep -v active) ]] && echo "NX Protection is not active"
```

This is equivalent to:

 1. No entries of `'noexec[0-9]*=off'` in `/proc/cmdline`
 2. Positive match for `' (pae|nx)' ` in `/proc/cpuinfo`
 3. Positive match for `protection: active` in `/var/log/dmesg`.

I'm tempted to elide the PAE portion and bring it up on a CIS call regarding this rule. But, I think 1 and 3 are important to have, that wouldn't impact STIG except in non-compliant client configurations.

cipherboy (approved):

PR #7239, opened 2021-07-13
Pass the input directory to the playbook builder explicitly

matejak:
The default value assumes that build directory is "build", which is not always the case, and no playbooks get built in such situations.


jan-cerny:
I have verified that it fixes the problem if I remove the `build` directory and create a different one instead. But it would be amazing if the commit or PR description contained a reproducer or a link to reproducer.

jan-cerny (approved):

PR #7244, opened 2021-07-13
Replace gpg keys to support Fedora 33 to 35

heryxpc:
#### Description:

- Changed the GPG keys from Fedora 30-32 to Fedora 33-35, which are the current _previous, later and future_ versions.

#### Rationale:

- We are running openscap in our environments and the rule `xccdf_org.ssgproject.content_rule_ensure_fedora_gpgkey_installed` is failing due to the gpg fingerprint key for Fedora 33 is not defined.

Any guidance if this is indeed the fix or if there is a workaround to avoid this false positive will be really appreciated.

### DIFF
--- a/products/fedora/product.yml
+++ b/products/fedora/product.yml
@@ -18,20 +18,20 @@ cpes:
       check_id: installed_OS_is_fedora
 
 # The fingerprint and pkg_version are retrieved from https://getfedora.org/keys/
-future_version: 32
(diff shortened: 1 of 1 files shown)


heryxpc:
CC @gkleiman

matejak (approved):
LGTM, I have checked those keys against the website, and they match. Thanks!

PR #7249, opened 2021-07-14
Introduce automated CCE adder

matejak:
You can call e.g.

`python utils/fix-rules.py -p products/rhel9/product.yml add-cce sshd_disable_root_login service_ntpd_enabled`

and the script will make sure that all rules have RHEL9 CCEs.
- CCE is not assigned if the rule is not applicable for the given product or if it already has a CCE
- CCEs are picked randomly from the file. Support for CCE pools other than the redhat pool is there.

Aside from that, the script has been reworked to remove code that looked ugly.

### DIFF
--- a/utils/fix-rules.py
+++ b/utils/fix-rules.py
@@ -4,10 +4,23 @@ import sys
 import os
 import jinja2
 import argparse
+import re
+import random
 
-from ssg import yaml, checks
+from ssg import yaml, checks, products
 from ssg.shims import input_func
 import ssg
+
+
+_COMMANDS = dict()
+
+
+def command(name, description):
+    def wrapper(wrapped):
+        _COMMANDS[name] = wrapped
+        wrapped.description = description
+        return wrapped
+    return wrapper
 
 
 def has_empty_identifier(yaml_file, product_yaml=None):
@@ -20,6 +33,21 @@ def has_empty_identifier(yaml_file, product_yaml=None):
             if str(value).strip() == "":
                 return True
     return False
+
+
+def has_no_cce(yaml_file, product_yaml=None):
+    rule = yaml.open_and_macro_expand(yaml_file, product_yaml)
+    product = product_yaml["product"]
+    if "prodtype" in rule and product not in rule["prodtype"]:
+        return False
+    if 'identifiers' in rule and rule['identifiers'] is None:
+        return True
+
+    if 'identifiers' in rule and rule['identifiers'] is not None:
+        for ident in rule['identifiers']:
+            if ident == "cce@" + product:
+                return False
+    return True
 
 
 def has_empty_references(yaml_file, product_yaml=None):
@@ -74,7 +102,7 @@ def has_int_reference(yaml_file, product_yaml=None):
     return False
 
 
-def find_rules(directory, func):
+def find_rules(directory, func, first_product_yaml=None):
     # Iterates over passed directory to correctly parse rules (which are
     # YAML files with internal macros). The most recently seen product.yml
     # takes precedence over previous product.yml, e.g.:
@@ -93,7 +121,7 @@ def find_rules(directory, func):
     results = []
     product_yamls = {}
     product_yaml_paths = {}
-    product_yaml = None
+    product_yaml = first_product_yaml
     product_yaml_path = None
     for root, dirs, files in os.walk(directory):
         dirs.sort()
@@ -129,6 +157,8 @@ def find_rules(directory, func):
             except jinja2.exceptions.UndefinedError:
                 print("Failed to parse file %s (with product.yml: %s). Skipping"
                       % (path, product_yaml_path))
+                pass
+            except ssg.yaml.DocumentationNotComplete:
                 pass
 
     return results
@@ -245,7 +275,10 @@ def rewrite_value_remove_prefix(line):
 
 
 class CCEFile:
-    def __init__(self, project_root):
+    def __init__(self, project_root=None):
+        if not project_root:
+            project_root = os.path.join(
+                os.path.dirname(os.path.abspath(__file__)), "..")
         self.project_root = project_root
 
     @property
@@ -289,8 +322,8 @@ def add_to_the_section(file_contents, yaml_contents, section, new_keys):
 
     sec_ranges = find_section_lines(file_contents, section)
     if len(sec_ranges) != 1:
-        raise RuntimeError("Refusing to fix file: %s -- could not find one section: %d"
-                           % (path, sec_ranges))
+        raise RuntimeError("could not find one section: %s"
+                           % section)
 
     begin, end = sec_ranges[0]
     r_lines = set()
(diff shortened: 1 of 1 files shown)


cipherboy:
@matejak I'd be a little careful with some of these changes. @yuumasato expressed interest in merging #6882 after @vojtapolasek's RHEL 7 CIS controls format changes merged (which, to my knowledge, it has)... And that introduces significant changes to the `fix-rules.py` script. Just be aware there will be some rebase conflicts with that either way this ends up merging. :-)

matejak:
> @matejak I'd be a little careful with some of these changes. @yuumasato expressed interest in merging #6882 after @vojtapolasek's RHEL 7 CIS controls format changes merged (which, to my knowledge, it has)... And that introduces significant changes to the `fix-rules.py` script. Just be aware there will be some rebase conflicts with that either way this ends up merging. :-)
> 
> That said, this looks like a great improvement in automation!

Good point, and I am comfortable with handling those conflicts myself, as I think that they should be mostly trivial. We also don't have to hurry with this, interested people can checkout the PR and make a copy of the file outside of git.

utils/fix-rules.py
```diff
@@ -244,6 +275,92 @@ def rewrite_value_remove_prefix(line):
...
+    template_line = str(file_contents[end - 1])
+    leading_whitespace = re.match(r"^\s*", template_line).group()
+    for key, value in new_keys.items():
+        to_insert.append(leading_whitespace + key + ": " + value)
+
+    new_contents = file_contents[:end] + to_insert + file_contents[end:]
+    return new_contents
+
+
+def sort_section(file_contents, yaml_contents, section):
```
cipherboy, line 344: In #6882, we have:

https://github.com/ComplianceAsCode/content/pull/6882/commits/8e115a101af074a7d03de3f8bbf1fed34c29fb7d#diff-7e689958b33c2feea57ead09db20cb40ea8dc8ee5a20afdffec67ede2c8b4c25R226

`ssg.rule_yaml.sort_section_keys(...)`

You might consider seeing if we could have a single implementation for use by both scripts?

jan-cerny:
@matejak please resolve conflicts

alexhaydock:
I wasn't able to get this to run in its current state - what am I missing? @matejak 

I tried below in a clean `python:3` container with just this script, and `ssg` and `jinja2` installed via Pip:
```bash
root@1b8ca50b5943:/usr/src# python3 /opt/fix-rules.py -p products/rhel8/product.yml add-cce root_path_no_dot
  File "/opt/fix-rules.py", line 10, in <module>
    from ssg import yaml, checks, products
ImportError: cannot import name 'yaml' from 'ssg' (/usr/local/lib/python3.9/site-packages/ssg/__init__.py)
```

I had the same issue on my Ubuntu 20.10 host so I'm sure I'm missing something.

ggbecker:
> I wasn't able to get this to run in its current state - what am I missing? @matejak
> 
> I tried below in a clean `python:3` container with just this script, and `ssg` and `jinja2` installed via Pip:
> 
> ```shell
> root@1b8ca50b5943:/usr/src# python3 /opt/fix-rules.py -p products/rhel8/product.yml add-cce root_path_no_dot
> Traceback (most recent call last):
>   File "/opt/fix-rules.py", line 10, in <module>
>     from ssg import yaml, checks, products
> ImportError: cannot import name 'yaml' from 'ssg' (/usr/local/lib/python3.9/site-packages/ssg/__init__.py)
> ```
> 
> I had the same issue on my Ubuntu 20.10 host so I'm sure I'm missing something.

I believe you need to define `PYTHONPATH` first, in the project's root directory there's a file called `.pyenv.sh` that should do the trick. Run:

`source .pyenv.sh` and the variable should be defined, in other words you need to include the project's directory as a valid python library directory.

https://complianceascode.readthedocs.io/en/latest/manual/developer/05_tools_and_utilities.html?highlight=source%20.pyenv.sh#tools-and-utilities

matejak:
@cipherboy Could you please just skim through whether it looks reasonable to you?

utils/fix_rules.py
```diff
@@ -257,6 +286,77 @@ def rewrite_value_remove_prefix(line):
     return key + ": " + new_value
 
 
+class CCEFile:
```
cipherboy, line 289: @matejak Does it make sense to move this to say, `ssg/cce.py` and move the `is_cce_format_valid` and `is_cce_value_valid` into there as well (from `ssg/checks.py`)?

utils/fix_rules.py
```diff
@@ -257,6 +286,77 @@ def rewrite_value_remove_prefix(line):
...
+    def line_to_cce(self, line):
+        return line
+
+    def line_isnt_cce(self, cce, line):
+        return line != cce
+
+    def read_cces(self):
+        with open(self.absolute_path, "r") as f:
+            cces = f.read().splitlines()
+        return cces
```
cipherboy, line 309: Does it make sense to filter for valid CCEs and only return CCE-like entries?

What's the format of this file? If it is strictly one CCE per line, do we need to do the line 311? Could we instead make this class closeable and do something like:

```python3
with CCEFile() as picker:
   ... assign cces ... 
``` 

And have the file be read once (into this class) and cces removed on pick and the file re-written from `.close()`? It means you probably have to use `random.randint` instead of `random.shuffle` over all CCEs, but might be cleaner.

My 2c.

utils/fix_rules.py
```diff
@@ -257,6 +286,77 @@ def rewrite_value_remove_prefix(line):
...
+
+CCE_POOLS["redhat"] = RedhatCCEFile
+
+
+def add_to_the_section(file_contents, yaml_contents, section, new_keys):
+    to_insert = []
+
+    sec_ranges = find_section_lines(file_contents, section)
+    if len(sec_ranges) != 1:
+        raise RuntimeError("could not find one section: %s"
```
cipherboy, line 340: I think it is ideal to also include the rule id in this and the assertion below. See the other helper methods where the path to the rule is explicitly passed. This makes it easier for the caller to inspect which particular file failed. My 2c. :-)

cipherboy (changes requested):
Looks like some nice changes, thanks @matejak! Just a few random comments, but don't bother holding up merge of this if you don't like the feedback :-)

cipherboy (approved):
Oops, meant to only comment so I'll approve to even things out.

utils/fix_rules.py
matejak: It's not that simple, `with` calls the close by default even when there is an exception, which would delete a CCE from the file even though the code that is supposed to write it into the rule fails. Also the current architecture also requires to come up with a potential CCE - that may not end up being assigned at all.
So this explicit approach ask for CCE + ask for its deletion when we know that it has been used looks best to me under current conditions, although I understand that it is not ideal.

utils/fix_rules.py
matejak: The rule is not known here, but I have added the handling code to the upper levels.

cipherboy (approved):
Thanks for the updates @matejak — looks like a great utility :)

cipherboy:
@matejak looks like there’s ctest errors though:

```
70/194 Test  #71: fix_rules-sort_subkeys ...................................................***Failed    0.29 sec
212
213
  File "/home/runner/work/content/content/utils/fix_rules.py", line 757, in <module>
214
    __main__()
215
  File "/home/runner/work/content/content/utils/fix_rules.py", line 744, in __main__
216
    args = parse_args()
217
  File "/home/runner/work/content/content/utils/fix_rules.py", line 739, in parse_args
218
    create_other_parsers(subparsers)
219
  File "/home/runner/work/content/content/utils/fix_rules.py", line 715, in create_other_parsers
220
    subparser.add_argument("--cce-pool", "-p", default="redhat", choices=list(CCE_POOLS.keys()))
221
NameError: name 'CCE_POOLS' is not defined
```

matejak:
Feel free to merge it if you like it!

Issue #7247, opened 2021-07-14
Remediation of grub rules errors on RHEL 9


adelton:
#### Description of problem:

Remediation of grub-related rules fails on RHEL 9. It could be that the rules incorrectly fail to determine that they are actually met.

#### SCAP Security Guide Version:

scap-security-guide-0.1.57-1.git452b4c2.el9.noarch

#### Operating System Version:

Red Hat Enterprise Linux release 9.0 Beta (Plow)

#### Steps to Reproduce:

1. `cat /etc/default/grub`
2. `grub2-editenv - list`
3. `for i in xccdf_org.ssgproject.content_rule_grub2_audit_backlog_limit_argument xccdf_org.ssgproject.content_rule_grub2_audit_argument xccdf_org.ssgproject.content_rule_grub2_pti_argument xccdf_org.ssgproject.content_rule_grub2_kernel_trust_cpu_rng xccdf_org.ssgproject.content_rule_grub2_slub_debug_argument xccdf_org.ssgproject.content_rule_grub2_page_poison_argument ; do oscap xccdf eval --remediate --profile xccdf_org.ssgproject.content_profile_ospp --rule $i /usr/share/xml/scap/ssg/content/ssg-rhel9-ds.xml ; done`

#### Actual Results:

```
# cat /etc/default/grub 
GRUB_TIMEOUT=5
GRUB_DISTRIBUTOR="$(sed 's, release .*$,,g' /etc/system-release)"
GRUB_DEFAULT=saved
GRUB_DISABLE_SUBMENU=true
GRUB_TERMINAL_OUTPUT="console"
GRUB_CMDLINE_LINUX="audit=1 audit_backlog_limit=8192 slub_debug=P page_poison=1 pti=on vsyscall=none rd.lvm.lv=VolGroup/root rd.lvm.lv=VolGroup/lv_swap rd.shell=0"
GRUB_DISABLE_RECOVERY="true"
GRUB_ENABLE_BLSCFG=true
```
```
saved_entry=19e59c11b07447a6bcdaa94033c55aea-5.13.0-1.el9.aarch64
 audit_backlog_limit=8192
 audit=1
 pti=on
 random.trust_cpu=on
 slub_debug=P
 page_poison=1
boot_success=1
```
```
[...]

PR #7257, opened 2021-07-16
Fixed the remediations when there are no previous kernelopts

matejak:
The #7247 indicates that our current remediation can't add the first kernel option. The fix takes a different approach if it detects that no kernelopts are present so far.

This fix is not specific to RHEL9.


yuumasato (commented):
LGTM so far.
Do you plan to also look at the Ansible remediation?
It suffers from the same issue

matejak:
I would be in favor of merging it now, Ansible can be added later, as even the current Ansible code is non-trivial.

yuumasato (approved):
Sounds good.

PR #7258, opened 2021-07-16
Fix coverity problems

jan-cerny:

### DIFF
--- a/shared/macros-bash.jinja
+++ b/shared/macros-bash.jinja
@@ -623,7 +623,7 @@ fi
 {{%- endmacro %}}
 
 {{%- macro bash_sudo_remove_config(parameter, pattern) -%}}
-for f in $( ls /etc/sudoers /etc/sudoers.d/* 2> /dev/null ) ; do
+for f in /etc/sudoers /etc/sudoers.d/* ; do
   matching_list=$(grep -P '^(?!#).*[\s]+{{{ pattern }}}.*$' $f | uniq )
   if ! test -z "$matching_list"; then
     while IFS= read -r entry; do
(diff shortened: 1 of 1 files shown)


shared/macros-bash.jinja
```diff
@@ -623,7 +623,7 @@ fi
 {{%- endmacro %}}
 
 {{%- macro bash_sudo_remove_config(parameter, pattern) -%}}
-for f in $( ls /etc/sudoers /etc/sudoers.d/* 2> /dev/null ) ; do
+for f in /etc/sudoers /etc/sudoers.d/* ; do
```
cipherboy, line 626: Hi @jan-cerny! 

Please see: https://github.com/ComplianceAsCode/content/pull/7109/files

Notably, in this pattern, we have to do a `if [ ! -e "$f" ]; then continue ; fi` type pattern because `ls` handles the filtering of files existing for us, whereas the `for x in glob-with-no-matches` will report the glob itself:

```bash
>    for f in /etc/sudoers.d/*.cozznf /etc/sudoers; do
>        echo "[ $f ]"
>    done
[ /etc/sudoers.d/*.cozznf ]
[ /etc/sudoers ]
```

Otherwise we'll get a lot of other output below when trying to read/modify/.. a file that doesn't exist.

cipherboy (changes requested):

shared/macros-bash.jinja
cipherboy: Thanks @jan-cerny!

cipherboy (approved):

ggbecker (approved):

PR #7259, opened 2021-07-16
Correct some issues with the CIS ICMP redirects rule on RHEL 7/8

alexhaydock:
#### Description:

Corrects some issues with incorrect CIS references on rules relating to ICMP redirects, and adds some additional required rules to the control in the RHEL 7 controls file.

cc @vojtapolasek 

#### Rationale:

- The references for the two rules here were incorrect, and in order to fully comply with the CIS control for ICMP redirects, the RHEL 7 CIS profile also needs to apply the "default" sysctl.


alexhaydock:
Pushed a new merge commit which fixes the conflict caused by 1931f6ea9480924a51e251bdf1ee511e9651e3e7 being merged.

vojtapolasek (approved):
Thank you for noticing this. Looks good, merging.

PR #7261, opened 2021-07-20
Set package_rear_installed to notapplicable on s390x

jan-cerny:
Resolves: RHBZ#1958939


matejak:
LGTM, the platform is already used elsewhere in the project.

matejak (approved):

Issue #7238, opened 2021-07-13
account_unique_id results in 'unknown' in container


mildas:
#### Description of problem:
The xccdf_org.ssgproject.content_rule_account_unique_id results in unknown during container scan.

#### SCAP Security Guide Version:
452b4c2

#### Steps to Reproduce:
1. `oscap-docker image registry.access.redhat.com/ubi8/ubi-minimal xccdf eval --progress --benchmark-id xccdf_org.ssgproject.content_benchmark_RHEL-8 --profile "(all)" /usr/share/xml/scap/ssg/content/ssg-rhel8-ds.xml`

#### Actual Results:
xccdf_org.ssgproject.content_rule_account_unique_id:unknown 

#### Expected Results:
No unknown result

mildas:
Not sure if this rule should be machine only. 
In my opinion, it makes sense in containers, so I see a problem in OVAL or maybe in scanner.

yuumasato:
The unknown result happens because the password probe doesn't work in offline mode. 

Below is an excerpt from scan logs 
[account_unique_id_logs.txt](https://github.com/ComplianceAsCode/content/files/6839304/account_unique_id_logs.txt)

`W: oscap:     Requested offline mode is not supported by password probe. [oscap(1935):probe_worker(7f4a5c7f8700):worker.c:988:probe_worker]`

@evgenyz @jan-cerny Do you know the reasons why this probe is not enabled in offline mode?

evgenyz:
Maybe because of `getpwent` and co.?

evgenyz:
But we could use `fgetpwent` instead, at least on Linux. Do we want it to be implemented? How soon?

yuumasato:
> But we could use `fgetpwent` instead, at least on Linux. Do we want it to be implemented?

If possible, yes, I think.

> How soon?

This is not urgent.

Should we mark the rule as machine only for now?

mildas:
This has been already fixed. Closing

PR #7263, opened 2021-07-20
Mark as machine rules that collect password_object

yuumasato:
#### Description:

- Mark as machine rules that collect password_object

#### Rationale:

- The password probe doesn't support offline mode, and the rule results in unknown.
- Related to #7238


vojtapolasek (approved):
LGTM

vojtapolasek:
The failing OCP test is not related to any of the changed rules.

PR #7266, opened 2021-07-21
OCP4: The moderate profile requires the audit profile to be WriteRequestBodies

jhrozek:
This is needed to cover AU-3(1).

Unfortunately, our CI infra is not yet equipped to handle tests that
need different results for different benchmarks and in this case it
would be needed - CIS requires the Default audit profile to be used, so
the rule passes by default and doesn't need a remediation whereas
moderate checks for WriteRequestBodies so the rule fails by default and
needs to be remediated. Therefore the e2e test is removed for this rule
until we enhance the e2e test machinery.

### DIFF
--- a/applications/openshift/logging/audit_profile_set/tests/ocp4/e2e.yml
+++ b/applications/openshift/logging/audit_profile_set/tests/ocp4/e2e.yml
@@ -1,2 +1,0 @@
----
-default_result: PASS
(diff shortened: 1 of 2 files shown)


jhrozek:
btw I created https://issues.redhat.com/browse/CMP-1041 to track enhancing the e2e tests

jhrozek:
@mrogers950 PTAL
this is adding the per-profile auditing, fixing of the tests is filed as https://issues.redhat.com/browse/CMP-1041

mrogers950 (approved):

PR #7271, opened 2021-07-22
Fix ansible remediation of display_login_attempts

ggbecker:
#### Description:

- Fix ansible remediation of display_login_attempts.

### DIFF
--- a/linux_os/guide/system/accounts/accounts-pam/display_login_attempts/ansible/shared.yml
+++ b/linux_os/guide/system/accounts/accounts-pam/display_login_attempts/ansible/shared.yml
@@ -10,11 +10,13 @@
 # complexity = low
 # disruption = low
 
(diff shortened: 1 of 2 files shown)


vojtapolasek (approved):
Thank you, the fix works, verified manually on clean rhel installation. However, the added test scenario does not display error when tested with previous implementation... so maybe we are not testing the right thing. But I have no idea why it is like that.
I will merge the fix, as it clearly improves the experience.

ggbecker:
> Thank you, the fix works, verified manually on clean rhel installation. However, the added test scenario does not display error when tested with previous implementation... so maybe we are not testing the right thing. But I have no idea why it is like that.
I had the same behavior and don't why I can't reproduce it. But the file should be the same as in RHEL7 vanilla.

> I will merge the fix, as it clearly improves the experience.

PR #7272, opened 2021-07-22
Update regex for openssl TLS MinProtocol parameter

### LINKED ISSUE #7236: RHEL8 STIG - configure_openssl_tls_crypto_policy fail
#### Description of problem:
The configure_openssl_tls_crypto_policy rule fails in RHEL8 STIG profile.

452b4c2

RHEL 8

#### Steps to Reproduce:
1. `python3 tests/test_suite.py profile --libvirt qemu:///system test_suite_vm --datastream /tmp/ssg-rhel8-ds.xml                         --xccdf-id scap_org.open-scap_cref_ssg-rhel8-xccdf-1.2.xml --mode online --remediate-using oscap xccdf_org.ssgproject.content_profile_stig`

#### Actual Results:
`xccdf_org.ssgproject.content_rule_configure_openssl_tls_crypto_policy - fail`

#### Expected Results:
`xccdf_org.ssgproject.content_rule_configure_openssl_tls_crypto_policy - pass`

<span class="label label-default"><abbr title="OVAL details taken from arf:report with id='oval1'">OVAL test results details</abbr></span><div class="panel panel-default"><div class="panel-body"><h4><span class="label label-primary">tests the value of MinProtocol setting in the /etc/crypto-policies/back-ends/opensslcnf.config file</span> 
        <span class="label label-default">oval:ssg-test_configure_openssl_tls_crypto_policy:tst:1</span> 
        <span class="label label-danger">false</span></h4><h5>No items have been found conforming to the following objects:</h5><h5>Object <strong><abbr>oval:ssg-obj_configure_openssl_tls_crypto_policy:obj:1</abbr></strong> of type
                <strong>textfilecontent54_object</strong></h5>

Filepath | Pattern | Instance
-- | -- | --
/etc/crypto-policies/back-ends/opensslcnf.config | ^(?:.*\n)*\s*MinProtocol\s*=\s*(.+?)[ \t]*(?:$\|#) | 1

</div></div>

yuumasato:
#### Description:

- Change OVAL regex to accept `TLS.MinProtocol` along with just `MinProtocol`.
  - The change is on the prefix, so that comments in OVAL don't include the TLS regex.
- Change remediation to set the crypto-policy instead of editing the back-end

#### Rationale:

- From RHEL-8.5, the the opensslcnf back-end parameter name changed.
- Fixes #7236


yuumasato:
I'm not sure how to cover the remediation. All crypto-policies in RHEL are compliant, but I see 2 issues:
1. Does it make sense to edit the crypto-policy back-end?
1. How to make the content remediate differently for RHEL-8.4 and RHEL-8.5.

ggbecker:
> I'm not sure how to cover the remediation. All crypto-policies in RHEL are compliant, but I see 2 issues:
> 
>     1. Does it make sense to edit the crypto-policy back-end?

I don't have the answer, but this is where it comes from: https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_8_security_technical_implementation_guide/V-230255
> 
>     2. How to make the content remediate differently for RHEL-8.4 and RHEL-8.5.

I think the only way would be to check redhat-release rpm or /etc/redhat-release file to detect which version it is. But that's not a good solution since it can complicate for other distros. DISA usually splits rules like that in the STIG, keeping one item for older releases and one item for newer releases, maybe that's something to be considered.

cipherboy:
> Does it make sense to edit the crypto-policy back-end?

I think it'd be best to work with the crypto team and make sure a policy is shipped that aligns with DISA expectations and if the current policy isn't correct for anything, have them switch to the right policy and then regen everything.

i.e., audit the actual config you expect but remediate by switching to the desired policy of the correct name and regen crypto policies.

If someone ships a custom policy it won't remediate by default (if OSCAP is used properly). 

My 2c.

yuumasato:
> > Does it make sense to edit the crypto-policy back-end?
> 
> I think it'd be best to work with the crypto team and make sure a policy is shipped that aligns with DISA expectations and if the current policy isn't correct for anything, have them switch to the right policy and then regen everything.
> 
> i.e., audit the actual config you expect but remediate by switching to the desired policy of the correct name and regen crypto policies.
> 
> If someone ships a custom policy it won't remediate by default (if OSCAP is used properly).
> 
> My 2c.

Reply from crypto team was that it is bad idea to edit the back-end, one can corrupt the pre-generated crypto-policies or lose policy customizations.

linux_os/guide/system/software/integrity/crypto/configure_openssl_tls_crypto_policy/ansible/shared.yml
```diff
@@ -3,16 +3,14 @@
...
-}}}
+- name: "{{{ rule_title }}}"
+  lineinfile:
+    path: /etc/crypto-policies/config
+    regexp: '^(?!#)(\S+)$'
+    line: "{{ var_system_crypto_policy }}"
+    create: yes
+
+- name: Verify that Crypto Policy is Set (runtime)
+  command: /usr/bin/update-crypto-policies --set {{ var_system_crypto_policy }}
```
carlosmmatos, line 16: @yuumasato doesn't this reset the crypto-policy files?

carlosmmatos (commented):
I may need to better understand the issue here, but see my comment

linux_os/guide/system/software/integrity/crypto/configure_openssl_tls_crypto_policy/ansible/shared.yml
yuumasato: @carlosmmatos Yes, the remediation will set the crypto-policy defined by the profile.

yuumasato: I reckon that setting the crypto-policy may not be desired.
So I'll drop the remediation and a recommendation in the rule that the admin needs to look for a crypto-policy that satisfies the rule's requirement.

yuumasato: Done in 97d792e, please check it out.

cipherboy: Could choice of crypto policy be controlled via an XCCDF variable? It seems like certain profiles OSPP/STIG have very particular choices of profiles (probably, FIPS) and we could use the DEFAULT crypto policy by default.

yuumasato: @cipherboy There is a [variable](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/software/integrity/crypto/var_system_crypto_policy.var) that can be used together with rule [configure_crypto_policy](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/software/integrity/crypto/configure_crypto_policy/rule.yml). But I don't where you are heading with the question.

cipherboy: @yuumasato I think it was trying to address @carlosmmatos's concern :-) He was worried about it resetting crypto-policy, but we already use a variable here and it aligns more or less with the existing rule as you mentioned. I think removing the additional remediation is correct, but I think we're missing an actual dependency from this rule onto `configure_crypto_policy` right? My 2c., but I think we want to make sure the policy rule gets dragged into the profile as well as this rule.

carlosmmatos (approved):
LGTM

PR #7274, opened 2021-07-22
Update package_vim_installed to point to vim-enhanced package

ggbecker:
#### Description:

- Update package_vim_installed to point to vim-enhanced package.
  - This is actually the package that contains the vim binary.


yuumasato (approved):
LGTM.

```
[root@localhost ~]# yum whatprovides /usr/bin/vim
Updating Subscription Management repositories.
Unable to read consumer identity

This system is not registered with an entitlement server. You can use subscription-manager to register.

Last metadata expiration check: 5:26:58 ago on Thu 22 Jul 2021 05:45:10 EDT.
vim-enhanced-2:8.0.1763-15.el8.x86_64 : A version of the VIM editor which includes recent enhancements
Repo        : @System
Matched from:
Filename    : /usr/bin/vim
```

PR #7275, opened 2021-07-22
SSGTS - add rhel9 install command, add rhel9 platform to crypto policy rule test scenarios

mildas:
The rhel9 install command is needed for preparation phase of SSGTS.

The crypto policy rules have rhel9 prodtype, so their test scenarios should be enabled on rhel9 too.

### DIFF
--- a/tests/ssg_test_suite/common.py
+++ b/tests/ssg_test_suite/common.py
@@ -473,6 +473,7 @@ INSTALL_COMMANDS = dict(
     fedora=("dnf", "install", "-y"),
     rhel7=("yum", "install", "-y"),
     rhel8=("yum", "install", "-y"),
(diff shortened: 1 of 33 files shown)


tests/ssg_test_suite/common.py
```diff
@@ -473,6 +473,7 @@ def get_cpe_of_tested_os(test_env, log_file):
     fedora=("dnf", "install", "-y"),
     rhel7=("yum", "install", "-y"),
     rhel8=("yum", "install", "-y"),
+    rhel9=("yum", "install", "-y"),
```
yuumasato, line 476: :+1:

yuumasato (approved):
LGTM

PR #7276, opened 2021-07-22
New rule for RHEL-08-020270

carlosmmatos:
#### Description:

- Adds new rule for RHEL 8 STIG. Very similar to the existing temporary accounts rule, except this one explicitly calls out the use of emergency rules as a different means of temporary account.
- There is no remediation or checking of this, as this rule is primarily an organizational control and is used as a reference.


linux_os/guide/system/accounts/accounts-restrictions/account_expiration/account_emergency_expire_date/rule.yml
```diff
@@ -0,0 +1,52 @@
...
+    cce@rhel8: CCE-85910-8
+
+references:
+    cis-csc: 1,12,13,14,15,16,18,3,5,7,8
+    cobit5: DSS01.03,DSS03.05,DSS05.04,DSS05.05,DSS05.07,DSS06.03
+    disa: CCI-000016,CCI-001682
+    isa-62443-2009: 4.3.3.2.2,4.3.3.5.1,4.3.3.5.2,4.3.3.7.2,4.3.3.7.3,4.3.3.7.4
+    isa-62443-2013: 'SR 1.1,SR 1.2,SR 1.3,SR 1.4,SR 1.5,SR 1.7,SR 1.8,SR 1.9,SR 2.1,SR 6.2'
+    iso27001-2013: A.12.4.1,A.12.4.3,A.6.1.2,A.7.1.1,A.9.1.2,A.9.2.1,A.9.2.2,A.9.2.3,A.9.2.4,A.9.2.6,A.9.3.1,A.9.4.1,A.9.4.2,A.9.4.3,A.9.4.4,A.9.4.5
+    nist: AC-2(2),AC-2(3),CM-6(a)
```
Mab879, line 40: Checking on https://commoncriteria.github.io/pp/references/nistvscci.html it only lists AC-2 (2) for the two CCIs listed. But I see the case for leaving them there.

carlosmmatos: This was literally just a copy/paste from an existing rule, so I'm assuming those references were "correct".. the only difference is this rule is more explicit of using the term "emergency" account, rather than temporary.

Mab879 (approved):

PR #7277, opened 2021-07-23
Disable ctrl-alt-del.target before masking it

### LINKED ISSUE #7213: RHEL 8 HIPAA/OSPP/STIG/CUI kickstart installation - disable_ctrlaltdel_reboot fail
#### Description of problem:
The `disable_ctrlaltdel_reboot` rule is failing after installing RHEL8 using HIPAA security policy.

958709e

RHEL 8

#### Steps to Reproduce:

1. Install RHEL 8 with HIPAA security policy
2. Scan the machine with openscap

#### Actual Results:
disable_ctrlaltdel_reboot fails

#### Expected Results:
disable_ctrlaltdel_reboot passes

OVAL details from HTML report:
<h4><span class="label label-primary">Disable Ctrl-Alt-Del key sequence override exists</span> 
        <span class="label label-default">oval:ssg-test_disable_ctrlaltdel_exists:tst:1</span> 
        <span class="label label-danger">false</span></h4><h5>Following items have been found on the system:</h5>

Filepath | Canonical path
-- | --
/etc/systemd/system/ctrl-alt-del.target | /usr/lib/systemd/system/reboot.target

yuumasato:
#### Description:

- Disable `ctrl-alt-del.target` before masking it.

#### Rationale:
- From 8.5, the target is enabled and cannot be simply masked.
- Fixes #7213

### DIFF
--- a/linux_os/guide/system/accounts/accounts-physical/disable_ctrlaltdel_reboot/bash/shared.sh
+++ b/linux_os/guide/system/accounts/accounts-physical/disable_ctrlaltdel_reboot/bash/shared.sh
@@ -4,6 +4,7 @@
 # The process to disable ctrl+alt+del has changed in RHEL7. 
 # Reference: https://access.redhat.com/solutions/1123873
 {{% endif %}}
+systemctl disable --now ctrl-alt-del.target
 systemctl mask --now ctrl-alt-del.target
 {{%- else -%}}
 # If system does not contain control-alt-delete.override,


ggbecker (approved):
We didn't have a general problem with the service disabled rules because the template already does these steps of disabling the service first.

https://github.com/ComplianceAsCode/content/blob/98887e53006382830c28847d2a698da8535c682c/shared/templates/service_disabled/bash.template#L11

PR #7286, opened 2021-07-26
Update RHEL7 and RHEL8 STIG overlay

ggbecker:
#### Description:

- Update RHEL7 STIG overlay.
- Update RHEL8 STIG overlay.


ggbecker:
There are multiple issues with the create-stig-overlay.py script that it doesn't make sense to squeeze this into the 0.1.57 release. Closing for now and we should follow up later in regular main branch work.

PR #7287, opened 2021-07-26
Fix create-stig-overlay.py script and use it during build time

ggbecker:
#### Description:

- Fix create-stig-overlay.py script and use it during build time
  - The CMake file has been updated to generate the stig_overlay.xml file as it's needed. This removes the need of manually submitting the file every time an update to STIG references in rules is made.
  - This PR also update some of the STIG benchmarks that have been updated and were not in sync with latest DISA releases.
- The mapping tables are generated when the product [CMakeLists.txt](https://github.com/ComplianceAsCode/content/blob/5c80ec203537b2da5f995ba693bfbe46666e3217/products/rhel8/CMakeLists.txt#L30) file contains for example the following https://github.com/ComplianceAsCode/content/blob/5c80ec203537b2da5f995ba693bfbe46666e3217/products/rhel8/CMakeLists.txt#L30

#### Rationale:

- No more deprecated stig_overlay files that are used to generate STIG mapping tables.

For more insights in the work done see the commit messages.


jan-cerny (changes requested):
Firefox, Chromium and JRE overlays aren't built.

ggbecker:
> Firefox, Chromium and JRE overlays aren't built.

They are not built because these products don't need it as they don't generate the STIG mapping tables. Maybe we should include them in their respective `CMakeLists.txt`

jan-cerny:
>      They are not built because these products don't need it as they don't generate the STIG mapping tables. Maybe we should include them in their respective CMakeLists.txt

Aha, OK. Then why the overlays of these products were stored in the reporstiory? Was that an inconsistency, unused artifact or did they have purpose?

ggbecker:
> > ```
> >  They are not built because these products don't need it as they don't generate the STIG mapping tables. Maybe we should include them in their respective CMakeLists.txt
> > ```
> 
> Aha, OK. Then why the overlays of these products were stored in the reporstiory? Was that an inconsistency, unused artifact or did they have purpose?

Unused artifact. They are only used when `ssg_build_html_stig_tables(${PRODUCT}) ` is selected in the respective product.

jan-cerny (approved):

PR #7291, opened 2021-07-27
Document search order of rule content

cipherboy:

---

Note that the change at the end wasn't documented but is observed behavior. We ran into this as we were changing a rule with a `shared.yml` Ansible remediation with limited applicability (`platform = multi_platform_sle`) and were surprised that shared content with the wrong platform was included ahead of templated content with broader applicability, when building Ubuntu products.


ggbecker:
If I understood it correctly, this behavior sounds like a bug. If a shared.yml file exists but doesn't have the platform for the current built product, I would expect that the build system would search other sources for the remediation (e.g. templated one)

cipherboy:
@ggbecker I don't disagree, but this is well outside the scope of this PR ;-) I think it is a long-standing bug in the way the content is selected and would take a fair amount of work to fix.

jan-cerny (approved):

PR #7294, opened 2021-07-27
Add option to avoid cleanup in SSGTS

cipherboy:
When debugging failed test cases in SSGTS, it is frequently helpful
to save the state of the VM with provisioned test cases.

---

How do we want to document this? Is there anything else we should do here? I've only been able to test with libvirt based deployments, but it might be nice for someone to test container based deployments as well. 

Also, unlike `--debug`, we quit executing the test suite entirely and leave cleanup to the operator. This means they can debug about as much as they want and not have to worry about resuming the test suite later. They can revert to the origin snapshot and then remove it.

### DIFF
--- a/tests/ssg_test_suite/test_env.py
+++ b/tests/ssg_test_suite/test_env.py
@@ -68,6 +68,7 @@ class TestEnv(object):
         self.ssh_additional_options = []
 
         self.product = None
+        self.cleanup = True
 
     def start(self):
         """
@@ -217,6 +218,9 @@ class VMTestEnv(TestEnv):
         virt.reboot_domain(self.domain, self.domain_ip, self.ssh_port)
 
     def finalize(self):
+        if not self.cleanup:
+            return
+
         self._delete_saved_state(self._origin)
         # self.domain.shutdown()
         # logging.debug('Shut the domain off')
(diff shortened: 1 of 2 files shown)


matejak:
Hello, thank you for your contribution. We have the `--debug` option that pauses the execution and lets you to examine the tested system, but having a snapshot or a container saved sounds much more useful and scalable.
However, I am not in favor of simply leaving snapshots behind - the test suite should clean up snapshots that it uses internally, and if we want to leave something behind, it has to be an explicit test suite decision rather than omission of cleanup. Also those snapshots should be named after the respective test runs, so one can identify afterwards what to look at.

Do you think that you could work on this more, so we end up with a coherent solution?

cipherboy:
\o hey there @matejak :-)

Sadly I no longer have a SSGTS-capable development environment set up, so perhaps any such improvements are best left for someone else :/ 

I definitely agree some consideration about proper cleanup could be done but I'm not really sure what's the best approach. The `--no-cleanup` option was best suited for when a single rule was executed, but you're right that it might be nice to save execution after each rule in a separate snapshot/container image. I'm not quite sure how that'd interact with multiple runs (the caller might still need to manually cleanup these saved but non-default-named snapshots tho). Perhaps an intern project next summer? ;-)

Issue #7280, opened 2021-07-23
systemd not following symlinks on FragmentPath on Ubuntu


cipherboy:
#### Description of problem:

In testing with @Mab879... :-)

On a RHEL 8.4 machine (and Fedora 33 with systemd 246), the system reports the FragmentPath correctly:

```
[ladmin@localhost ~]$ systemctl --version  
systemd 239 (239-45.el8_4.2) 
+PAM +AUDIT +SELINUX +IMA -APPARMOR +SMACK +SYSVINIT +UTMP +LIBCRYPTSETUP +GCRYPT +GNUTLS +ACL +XZ +LZ4 +SECCOMP +BLKID +ELFUTILS +KMOD +IDN2 -IDN +PCRE2 default-hierarchy=legacy
[ladmin@localhost ~]$ sudo systemctl mask cups 
Created symlink /etc/systemd/system/cups.service → /dev/null.
[ladmin@localhost ~]$ systemctl show cups | grep FragmentPath 
FragmentPath=/dev/null
```

However, on a Ubuntu 20.04 (and Ubuntu 21.10 testing -- with `248.3-1ubuntu2`) box we don't follow FragmentPath:

```
root@ubuntu-2004-base:~# systemctl --version
systemd 245 (245.4-4ubuntu3.11)
+PAM +AUDIT +SELINUX +IMA +APPARMOR +SMACK +SYSVINIT +UTMP +LIBCRYPTSETUP +GCRYPT +GNUTLS +ACL +XZ +LZ4 +SECCOMP +BLKID +ELFUTILS +KMOD +IDN2 -IDN +PCRE2 default-hierarchy=hybrid
root@ubuntu-2004-base:~# systemctl disable cups
Synchronizing state of cups.service with SysV service script with /lib/systemd/systemd-sysv-install.
Executing: /lib/systemd/systemd-sysv-install disable cups
Removed /etc/systemd/system/multi-user.target.wants/cups.path.
Removed /etc/systemd/system/sockets.target.wants/cups.socket.
Removed /etc/systemd/system/printer.target.wants/cups.service.
root@ubuntu-2004-base:~# systemctl mask cups
Created symlink /etc/systemd/system/cups.service → /dev/null.
root@ubuntu-2004-base:~# systemctl show cups | grep -i FragmentPath
FragmentPath=/etc/systemd/system/cups.service
root@ubuntu-2004-base:~# ls -alh /etc/systemd/system/cups.service
[...]

Issue #7248, opened 2021-07-14
Rule xccdf_org.ssgproject.content_rule_service_systemd-coredump_disabled fails to work on RHEL 9


adelton:
#### Description of problem:

Even if systemd-coredump is masked, xccdf_org.ssgproject.content_rule_service_systemd-coredump_disabled seems to fail and remediation errors.

#### SCAP Security Guide Version:

scap-security-guide-0.1.57-1.git452b4c2.el9.noarch

#### Operating System Version:

Red Hat Enterprise Linux release 9.0 Beta (Plow)

#### Steps to Reproduce:

1. `systemctl status systemd-coredump`
2. `oscap xccdf eval --remediate --profile xccdf_org.ssgproject.content_profile_ospp --rule xccdf_org.ssgproject.content_rule_service_systemd-coredump_disabled /usr/share/xml/scap/ssg/content/ssg-rhel9-ds.xml`

#### Actual Results:

```
○ systemd-coredump.service
     Loaded: masked (Reason: Unit systemd-coredump.service is masked.)
     Active: inactive (dead)
```
```
WARNING: Datastream component 'scap_org.open-scap_cref_security-data-oval-com.redhat.rhsa-RHEL9.xml' points out to the remote 'https://www.redhat.com/security/data/oval/com.redhat.rhsa-RHEL9.xml'. Use '--fetch-remote-resources' option to download it.
WARNING: Skipping 'https://www.redhat.com/security/data/oval/com.redhat.rhsa-RHEL9.xml' file which is referenced from datastream
WARNING: Skipping ./security-data-oval-com.redhat.rhsa-RHEL9.xml file which is referenced from XCCDF content
Title   Disable acquiring, saving, and processing core dumps
Rule    xccdf_org.ssgproject.content_rule_service_systemd-coredump_disabled
Ident   CCE-83974-6
Result  fail


 --- Starting Remediation ---
WARNING: Skipping ./security-data-oval-com.redhat.rhsa-RHEL9.xml file which is referenced from XCCDF content
Title   Disable acquiring, saving, and processing core dumps
Rule    xccdf_org.ssgproject.content_rule_service_systemd-coredump_disabled
Ident   CCE-83974-6
Result  error
```

#### Expected Results:

Either the rule should be happy that the service is already masked, or it should mask it some more.

matejak:
[...]

PR #7296, opened 2021-07-27
Remove FragmentPath check from service_disabled

cipherboy:
In https://github.com/systemd/systemd/issues/582, it is documented that
systemd could eventually replace `FragmentPath=/dev/null` (on masked
services) with the actual service path -- not the fully (symlink)
resolved path as is currently the case.

This matches the behavior currently seen in Ubuntu (all versions) and
RHEL 9/Fedora 34.

Per discussion with Gabriel, Matej, Richard, and Matt, it is best to
remove this check, especially since `ActiveState=Masked` suffices.

Resolves: #7280
Resolves: #7248

### DIFF
--- a/shared/templates/service_disabled/oval.template
+++ b/shared/templates/service_disabled/oval.template
@@ -13,7 +13,6 @@
       <criteria operator="AND" comment="service {{{ SERVICENAME }}} is not configured to start">
         <criterion comment="{{{ SERVICENAME }}} is not running" test_ref="test_service_not_running_{{{ SERVICENAME }}}" />
         <criterion comment="Property LoadState of service {{{ SERVICENAME }}} is masked" test_ref="test_service_loadstate_is_masked_{{{ SERVICENAME }}}" />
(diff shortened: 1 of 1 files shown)


matejak:
The FragmentPath check has been introduced in #4633. @ggbecker do you recall whether there was a concrete reason why it was introduced, or was it just general caution?
If there is such a reason, another approach could involve the [symlink test](https://oval-community.readthedocs.io/en/latest/oval-schema-documentation/unix-definitions-schema.html#symlink-test).

ggbecker:
> The FragmentPath check has been introduced in #4633. @ggbecker do you recall whether there was a concrete reason why it was introduced, or was it just general caution?
> If there is such a reason, another approach could involve the [symlink test](https://oval-community.readthedocs.io/en/latest/oval-schema-documentation/unix-definitions-schema.html#symlink-test).

It was a general caution. I remember looking for sources on the internet and this was usually included, but IMO just checking the ActiveState=Masked should be enough.

matejak (approved):
LGTM, thanks for the fix!

PR #7299, opened 2021-07-28
Update References for RHEL8 STIG V1R3

Mab879:
#### Description:

This PR updates STIG references for RHE8 V1R3 STIG.

#### Rationale:

Make it easier to compare across other tools.


ggbecker:
I've noticed that `RHEL-08-010544` can be added to https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/software/disk_partitioning/partition_for_var_tmp/rule.yml

ggbecker (approved):
I have rebased your pull request to remove that conflict from the timeline, it's not good to have those conflicts in the git history.

linux_os/guide/system/auditing/service_auditd_enabled/rule.yml
```diff
@@ -52,7 +52,7 @@ references:
     srg: SRG-OS-000037-GPOS-00015,SRG-OS-000038-GPOS-00016,SRG-OS-000039-GPOS-00017,SRG-OS-000040-GPOS-00018,SRG-OS-000041-GPOS-00019,SRG-OS-000042-GPOS-00021,SRG-OS-000051-GPOS-00024,SRG-OS-000054-GPOS-00025,SRG-OS-000122-GPOS-00063,SRG-OS-000254-GPOS-00095,SRG-OS-000255-GPOS-00096,SRG-OS-000365-GPOS-00152,SRG-OS-000392-GPOS-00172,SRG-OS-000480-GPOS-00227,SRG-OS-000062-GPOS-00031
     stigid@ol7: OL07-00-030000
     stigid@rhel7: RHEL-07-030000
-    stigid@rhel8: RHEL-08-010560
```
ggbecker, line 55: We should probably tell DISA to remove one of these. We advised them to split the service auditd but there was already a STIG item that did that. The fix actually was just to remove the auditd service enabled from the install audit package STIG item.

ggbecker: I'm preparing the documentation to send them with requests.

PR #7301, opened 2021-07-29
SCE - shared/template support

cipherboy:
#### Dependencies:

- [x] #7075 for base SCE support.
- [x] #7211 for some common `ssg/template.py` improvements/changes

This pull request will have to be rebased after those two PRs merge. 

#### Description & Rationale:

Sometimes support for SCE isn't enough; we'd like to use the information already in the template system when generating SCE content. This is most commonly typical for auditd rules: SCE allows us to audit the runtime auditd configuration (where OVAL only allows auditing of the content on disk). Rather than writing SCE content per-rule manually, we can use the template system to automatically generate Bash SCE content for auditing this.

Unlike generic SCE support, the template system requires knowledge of the target language. This is most frequently bash, so that's what we've added here, though in a manner that if future SCE languages are desired in the future, they could be added.

This also extends the profile tool to support SCE content. Let me know if you'd prefer that in a separate PR.

### DIFF
--- a/ssg/build_yaml.py
+++ b/ssg/build_yaml.py
@@ -1408,6 +1408,7 @@ class Rule(object):
         if main_ref.attrib:
             rule.append(main_ref)
 
+        ocil_parent = rule
         check_parent = rule
 
         if self.sce_metadata:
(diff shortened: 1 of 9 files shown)


cipherboy:
I think this is ready for review :-)

ssg/build_yaml.py
```diff
@@ -1427,7 +1418,31 @@ def to_xml_element(self):
...
+            # Finally, before we begin, we might have an element with both SCE
+            # and OVAL. We have no way of knowing (right here) whether that is
+            # the case (due to a variety of issues, most notably, that linking
+            # hasn't yet occurred). So we must rely on the content author's
+            # good will, by annotating SCE content with a complex-check tag
+            # if necessary.
+
+            if 'complex-check' in self.sce_metadata:
+                # Here we have an issue: XCCDF allows EITHER one or more check
+                # elements OR a single complex-check. While we have an explicit
```
jan-cerny, line 1431: I have added an `sce-bash.template` to the package_installed template. When I build a datastream I get 3 `check` elements within the `Rule` elements. First`check` is for OVAL, second `check` for SCE and a third `check` for OCIL. Is that expected? When will the `complex-check` element created?

ssg/templates.py
```diff
@@ -108,10 +110,10 @@ def __init__(
         self.output_dirs = dict()
         for lang in languages:
             lang_dir = lang
-            if lang == "oval":
-                # OVAL checks need to be put to a different directory because
-                # they are processed differently than remediations later in
-                # the build process
+            if lang == "oval" or lang.startswith("sce-"):
+                # OVAL and SCE checks need to be put to a different directory
```
jan-cerny, line 114: Why isn't the directory called `sec-bash`?

ssg/build_yaml.py
cipherboy: From my reading of the XCCDF spec, multiple top-level checks in different languages are implicitly joined with an "OR" and it is up to the implementation/execution environment to choose which to execute.

See the documentation section (already merged) on complex-check [for adding one](https://github.com/ComplianceAsCode/content/blob/master/docs/manual/developer/06_contributing_with_content.md#sce-check-content). Note that `complex-check` XML element requires the creation of the correct header (`# complex-check = <operator>`) for instance, and then the OCIL will be dropped into one side and then any OVAL/SCE content will be dropped into the other with the operator from the header.

ssg/templates.py
cipherboy: The rule-dir SCE content is placed in a directory called `sce` with filename `{{{ rule_id }}}` and file extension of the language used (presumably -- we preserve whatever file extension was already in the rule_dir file -- e.g., `.py`, `.sh` &c). In the templating system, we lack knowledge of file extension given by the author (they're all `.template`) and so we need another mechanism to indicate file suffix. However, we still (for XCCDF and DS path) want to put it in the `sce/` directory. For now, Canonical only needs Bash content, so we've called it `sce-bash`. If someone else wanted to add, e.g., Python support, they _only_ need to do so in the templating system (adding `sce-python` and the correct `.py` suffix) and then the rest of the SCE infrastructure gets reused. 

HTH clarifies things.

jan-cerny (approved):
Thanks for clarification. I was able to add a SCE bash template for package_installed and use it in some rules.

PR #7303, opened 2021-07-29
Update existing rule for RHEL-08-020320

Mab879:
#### Description:

Add to profile, clean up, add tests, and moved the rule to a better location. Keeping under SAP did not seem like a good idea.

#### Rationale:
- https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_8_security_technical_implementation_guide/V-230379


linux_os/guide/system/accounts/accounts-restrictions/accounts_authorized_local_users/ansible/shared.yml
```diff
@@ -1,4 +1,4 @@
-# platform = multi_platform_sle
+# platform = multi_platform_sle,multi_platform_rhel
 # reboot = false
 # strategy = restrict
 # complexity = low
```
Mab879, line 4: This seem low to me, do we want to raise it to medium or high? Looking for more guidance on this.

ggbecker: I don't know how to answer this. Following the documentation [here](https://complianceascode.readthedocs.io/en/latest/manual/developer/06_contributing_with_content.html?highlight=complexity#remediations) it indicates it's just informative and there is no real impact on anything. This is translated into ansible tags but I doubt people are actually using this meta information in any way.

```
The estimated complexity or difficulty of applying the fix to the target. Only informative for now
```

Mab879: My issue is that without tailoring this rule will cause issues when remediated. Setting `disruption` to medium or high might be good idea or add warning like in some other rules.

ggbecker: I agree. You can add a warning to the rule.yml similar as it's done in: https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/software/sudo/sudo_dedicated_group/rule.yml#L14

Mab879: The more I think about it, just deleting users doesn't sit right with me, do want to disable remediation by default?

jan-cerny: > The more I think about it, just deleting users doesn't sit right with me, do want to disable remediation by default?

Why?

Mab879: By default this will delete all user added local users on the box, for a new system that is fine, but for existing systems this will be highly distributive.

jan-cerny: OK, so please remove it. Also, please make sure that the test scenarios will pass.

jan-cerny (dismissed):

jan-cerny:
unfortunately we've got some conflicting files, could you fix that?

Mab879:
@jan-cerny Done, thanks for pointing that you.

linux_os/guide/system/accounts/accounts-restrictions/var_accounts_authorized_local_users_regex.var
```diff
@@ -22,5 +22,6 @@ operator: pattern match
 interactive: true
 
 options:
+    rhel8: "^(root|bin|daemon|adm|lp|sync|shutdown|halt|mail|operator|games|ftp|nobody|pegasus|systemd-bus-proxy|systemd-network|dbus|polkitd|abrt|unbound|tss|libstoragemgmt|rpc|colord|usbmuxd$|pcp|saslauth|geoclue|setroubleshoot|rtkit|chrony|qemu|radvd|rpcuser|nfsnobody|pulse|gdm|gnome-initial-setup|postfix|avahi|ntp|sshd|tcpdump|oprofile|uuidd)$"
```
ggbecker, line 25: On a vanilla RHEL8 I'm getting fails and the following users are being reported. I tested this with an empty `bad_user.fail.sh` file. I don't know if this should be reported as fail but 
I think it's worth investigating... maybe some account is not being defined in the regex.
```
/etc/passwd | daemon:
/etc/passwd | bin:
/etc/passwd | adm:
/etc/passwd | lp:
/etc/passwd | sync:
/etc/passwd | shutdown:
/etc/passwd | halt:
/etc/passwd | mail:
/etc/passwd | operator:
/etc/passwd | games:
/etc/passwd | ftp:
/etc/passwd | nobody:
/etc/passwd | dbus:
/etc/passwd | systemd-coredump:
/etc/passwd | systemd-resolve:
/etc/passwd | tss:
/etc/passwd | polkitd:
/etc/passwd | unbound:
/etc/passwd | sssd:
/etc/passwd | chrony:
/etc/passwd | sshd:
```

ggbecker:
changes look good... I just wonder if we are using the correct users in the regex. From the STIG, it states that:

`Accounts such as "games" and "gopher" are not authorized accounts as they do not support authorized system functions.`

And in the default RHEL8 there is an account called `games` as described in the STIG and it should not be allowed... on the other hand this is on a RHEL default installation and we are not providing the remediation because it can be dangerous to remove legit accounts. We should gather more opinions on this. @yuumasato What do you think about this?

ggbecker:
> changes look good... I just wonder if we are using the correct users in the regex. From the STIG, it states that:
> 
> `Accounts such as "games" and "gopher" are not authorized accounts as they do not support authorized system functions.`
> 
> And in the default RHEL8 there is an account called `games` as described in the STIG and it should not be allowed... on the other hand this is on a RHEL default installation and we are not providing the remediation because it can be dangerous to remove legit accounts. We should gather more opinions on this. @yuumasato What do you think about this?

I have found a few discussions around this topic. It's not conclusive and mostly system users are legacy.
https://access.redhat.com/discussions/685053
https://access.redhat.com/solutions/31669

Since there is a XCCDF variable for users allowed, we can move forward with this. The user's list can be tailored according to the need.

ggbecker (approved):

PR #7304, opened 2021-07-30
Fix wrongly referenced python variable

ggbecker:
#### Description

- Fix wrongly referenced python variable. Interpreted language problems.


matejak (approved):
LGTM, nice fix!

PR #7309, opened 2021-07-30
Add documentation on SCE content

cipherboy:
`Signed-off-by: Alexander Scheel <alex.scheel@canonical.com>`

---

Note that `complex-check` is introduced in #7301.

PR #7312, opened 2021-07-30
Document the build system

cipherboy:
#### Description:

This aims to provide provide better inline documentation in the code as well as a high-level introduction to the build system in the developer guide.

Note that this technically documents the state of the system after #7301 is applied for the same reason as #7309, though this only applies to the comment in `build-scripts/build_sce.py`.


cmake/SSGCommon.cmake
```diff
@@ -608,6 +659,9 @@ macro(ssg_build_pci_dss_xccdf PRODUCT)
     )
 endmacro()
 
+# Build source datastreams (as opposed to result datastreams that occur after
```
jan-cerny, line 662:
```suggestion
# Build source data streams (as opposed to result data streams that occur after
```

docs/manual/developer/07_understanding_build_system.md
```diff
@@ -0,0 +1,141 @@
...
+   system.
+ - [OVAL](https://oval.cisecurity.org/), the Open Vulnerability and Assessment
+   Language; this is a standardized mechanism for auditing various compliance
+   checks.
+ - [OCIL](https://csrc.nist.gov/projects/security-content-automation-protocol/specifications/ocil),
+   the Open Checklist Interactive Language, an expressive language for
+   handling manual compliance checks.
+ - [CPE](https://nvd.nist.gov/products/cpe), the Common Platform Enumeration;
+   a scheme for identifying software and systems.
+ - A DS, Datastream format, a mechanism for combining the above into a single
```
jan-cerny, line 23:
```suggestion
 - SCAP source data stream format, a mechanism for combining the above into a single
```

docs/manual/developer/07_understanding_build_system.md
```diff
@@ -0,0 +1,141 @@
+# Understanding the ComplianceAsCode build system
+
+## Introduction
+
```
jan-cerny, line 4: We usually use the system "one sentence = one line". The advantage is that you don't have to re-wrap everything.

docs/manual/developer/07_understanding_build_system.md
```diff
@@ -0,0 +1,141 @@
...
+However, the specifics of building a particular product are contained in the
+shared module located at `cmake/SSGCommon.cmake`. This file contains all of
+the CMake logic to build a particular product and exposes the top-level macro
+`ssg_build_product(...)`. This macro generates per-product build, installation,
+and testing targets. While the specifics should be understood from this file
+directly, in general this takes the following outline of steps in rough order
+of occurrence:
+
+ 1. Generate SCE content and metadata,
+ 2. Generate a shorthand XML from various YAML files (like rules, variables,
```
jan-cerny, line 78: shorthand is a term that appears here without prior explanation in this document

jan-cerny (commented):
this is amazing

docs/manual/developer/07_understanding_build_system.md
cipherboy: I ended up writing this after I wrote the CMake documentation. I don't think shorthand is ever defined elsewhere either and is kinda in flux as the build process changes. Do you have a suggestion around how to introduce it? It isn't a standard we can link to externally, right?

jan-cerny: It isn't a standard, it's our invention and unfortunately the format isn't documented anywhere. I think you have introduced it nicely in the code comments: "a single XML document that resembles the structure of the XCCDF but isn't standards compliant", you can use it here as well.

cipherboy: @jan-cerny updated :)

docs/manual/developer/07_understanding_build_system.md
cipherboy: Sorry, thta's how I ended up writing tis one. Do you want me to reflow it now?

jan-cerny: No, that would be a waste of time. But you can try to use it in you future additions.

jan-cerny (approved):

PR #7323, opened 2021-08-03
Add new rule for RHEL-07-030330 and RHEL-08-030730

Mab879:
#### Description:

This new rule is copy of auditd_data_retention_space_left, but setup to allow for percentages.

#### Rationale:

Modifying the rule auditd_data_retention_space_left was going make things too complex. Just made a new rule

STIG References:
* https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_8_security_technical_implementation_guide/V-230483
* https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-204513


linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_data_retention_space_left_percentage/tests/no_percent_sign.fail.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+# profiles = xccdf_org.ssgproject.content_profile_stig
```
jan-cerny, line 2: According to https://complianceascode.readthedocs.io/en/latest/tests/README.html#scenarios-format you shouldn't use `profiles` in test scenarios headers unless you're making a regression test for a bug in a profile. You probably want to use `variables` instead and specify the value here.

linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_data_retention_space_left_percentage/ansible/shared.yml
```diff
@@ -0,0 +1,15 @@
+# platform = Red Hat Virtualization 4,multi_platform_fedora,multi_platform_rhel,multi_platform_ol,multi_platform_sle
```
jan-cerny, line 1: The rule has `prodtype:  fedora,rhel7,rhel8,rhel9` so I think that the remediation shouldn't have RHV, SLE and Oracle Linux.  I would prefer `platform = multi_platform_all`, the remediation script won't get into other products because the rule has `prodtype`.

linux_os/guide/system/auditing/configure_auditd_data_retention/var_auditd_space_left_percentage.var
```diff
@@ -0,0 +1,13 @@
...
+title: 'The percentage remaining in disk space before prompting space_left_action'
+
+description: 'The setting for space_left as a percentage in /etc/audit/auditd.conf'
+
+type: number
+
+interactive: false
+
+options:
+    25pc: 25
```
ggbecker, line 12:
```suggestion
    25pc: 25
    50pc: 50
    75pc: 75
```

let's add some more options here

linux_os/guide/system/auditing/configure_auditd_data_retention/var_auditd_space_left_percentage.var
```diff
@@ -0,0 +1,13 @@
+documentation_complete: true
+
+title: 'The percentage remaining in disk space before prompting space_left_action'
+
+description: 'The setting for space_left as a percentage in /etc/audit/auditd.conf'
+
+type: number
+
+interactive: false
```
ggbecker, line 9:
```suggestion
interactive: true
```

set to true here so in the scap-workbench users can customize to any other value they want. Otherwise it only allows to select the predefined values.

linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_data_retention_space_left_percentage/tests/no_percent_sign.fail.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
ggbecker, line 1:
```suggestion
#!/bin/bash
# variables = var_auditd_space_left_percentage=25pc
```

let's make these test scenarios future proof

ggbecker:
There is something strange going on with the test scenarios now that variables are selected. They are all failing in the remediation phase.

Also can you please address the feedback in the variable definition?

Thanks

ggbecker:
> There is something strange going on with the test scenarios now that variables are selected. They are all failing in the remediation phase.
> 
I've discovered we have to use the actual value in the `variables` metadata. So we need to set it to `=25`

linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_data_retention_space_left_percentage/tests/no_percent_sign.fail.sh
```diff
@@ -0,0 +1,6 @@
...
+# variables = var_auditd_space_left_percentage=25pc
```
ggbecker, line 2:
```suggestion
# variables = var_auditd_space_left_percentage=25
```

products/rhel8/profiles/stig.profile
```diff
@@ -856,8 +856,7 @@ selections:
     # RHEL-08-030720
 
     # RHEL-08-030730
-    # this rule expects configuration in MB instead percentage as how STIG demands
-    # - auditd_data_retention_space_left
+    - auditd_data_retention_space_left_percentage
```
ggbecker, line 859:
```suggestion
    - auditd_data_retention_space_left_percentage
    
    # RHEL-08-030731
```

The rule below `auditd_data_retention_space_left_action` has been split a there is a new STIG id assigned to it. The rule.yml itself has been updated accordingly. But the metadata in the profile file has not.

ggbecker (approved):
One minor request, but otherwise LGTM. Thanks

products/rhel8/profiles/stig.profile
Mab879: @ggbecker Done.

PR #7324, opened 2021-08-03
Automated STIG Control File Creation

Mab879:
#### Description:

Provides a scripted method to create control files based on DISA's XCCDF files and gives RHEL8 as an example.

#### Rationale:

This allow us better find gaps in our coverage on STIGs.

### DIFF
--- a/controls/stig_rhel8.yml
+++ b/controls/stig_rhel8.yml
@@ -1,2033 +1,2056 @@
 policy: Red Hat Enterprise Linux 8 Security Technical Implementation Guide
-source: https://public.cyber.mil/stigs/downloads/
 title: Red Hat Enterprise Linux 8 Security Technical Implementation Guide
 id: stig_rhel8
-version: V1R3
+source: https://public.cyber.mil/stigs/downloads/
 levels:
-  - id: high
-  - id: medium
-  - id: low
-
+    -   id: high
+    -   id: medium
+    -   id: low
 controls:
     -   id: RHEL-08-010000
         levels:
             - high
+        title: RHEL 8 must be a vendor-supported release.
         rules:
             - installed_OS_is_vendor_supported
-        title: RHEL 8 must be a vendor-supported release.
-        automated: 'yes'
+        status: automated
     -   id: RHEL-08-010010
         levels:
             - medium
-        rules:
-            - security_patches_up_to_date
         title: RHEL 8 vendor packaged system security patches and updates must be installed
             and up to date.
-        automated: 'yes'
+        rules:
+            - security_patches_up_to_date
+        status: automated
     -   id: RHEL-08-010020
         levels:
             - high
-        rules:
-            - enable_fips_mode
-            - sysctl_crypto_fips_enabled
         title: 'RHEL 8 must implement NIST FIPS-validated cryptography for the following:
     to provision digital signatures, to generate cryptographic hashes, and to protect
     data requiring data-at-rest protections in accordance with applicable federal
     laws, Executive Orders, directives, policies, regulations, and standards.'
-        automated: 'yes'
+        rules:
+            - enable_fips_mode
+            - sysctl_crypto_fips_enabled
+        status: automated
     -   id: RHEL-08-010030
         levels:
             - medium
-        rules:
-            - encrypt_partitions
         title: All RHEL 8 local disk partitions must implement cryptographic mechanisms
             to prevent unauthorized disclosure or modification of all information that requires
             at rest protection.
-        automated: 'yes'
+        rules:
+            - encrypt_partitions
+        status: automated
     -   id: RHEL-08-010040
         levels:
             - medium
-        rules:
-            - sshd_enable_warning_banner
         title: RHEL 8 must display the Standard Mandatory DoD Notice and Consent Banner
             before granting local or remote access to the system via a ssh logon.
-        automated: 'yes'
+        rules:
+            - sshd_enable_warning_banner
+        status: automated
     -   id: RHEL-08-010050
         levels:
             - medium
-        rules:
-            - dconf_gnome_banner_enabled
-            - dconf_gnome_login_banner_text
         title: RHEL 8 must display the Standard Mandatory DoD Notice and Consent Banner
             before granting local or remote access to the system via a graphical user logon.
-        automated: 'yes'
+        rules:
+            - dconf_gnome_login_banner_text
+        status: automated
     -   id: RHEL-08-010060
         levels:
             - medium
-        rules:
-            - banner_etc_issue
         title: RHEL 8 must display the Standard Mandatory DoD Notice and Consent Banner
             before granting local or remote access to the system via a command line user logon.
-        automated: 'yes'
+        rules:
+            - banner_etc_issue
+        status: automated
     -   id: RHEL-08-010070
         levels:
             - medium
+        title: All RHEL 8 remote access methods must be monitored.
         rules:
             - rsyslog_remote_access_monitoring
-        title: All RHEL 8 remote access methods must be monitored.
-        automated: 'yes'
+        status: automated
     -   id: RHEL-08-010090
         levels:
             - medium
         title: RHEL 8, for PKI-based authentication, must validate certificates by constructing
             a certification path (which includes status information) to an accepted trust
             anchor.
-        automated: 'no'
-
+        status: pending
     -   id: RHEL-08-010100
         levels:
             - medium
         title: RHEL 8, for certificate-based authentication, must enforce authorized access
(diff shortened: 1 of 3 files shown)


ggbecker:
This looks very nice. I created #7326 where I update both RHEL7/8 benchmark files. I think we can merge that one and then you rebase this one.

utils/build_stig_control.py
```diff
@@ -0,0 +1,114 @@
...
+                control['zzautomated'] = 'yes'
+            else:
+                control['zzautomated'] = 'no'
+            controls.append(control)
+
+    output = dict()
+    output['controls'] = controls
+    f = open(args.output, 'w')
+    out_text = yaml.dump(output)
+    out_text = out_text.replace('zzautomated:', 'automated:')
```
cipherboy, line 82: Why do this out of curiosity? It doesn’t look like the key is conditional on anything, so why not provision the automated key right away up above? :-)

utils/build_stig_control.py
```diff
@@ -0,0 +1,114 @@
...
+    f.write(out_text)
+    if not f.closed:
+        f.flush()
+        f.close()
+
+
+def get_implemented_stigs(args):
+    linux_products, other_products = ssg.products.get_all(args.root)
+    all_products = linux_products.union(other_products)
+    all_rule_dirs, product_yamls = walk_products(args.root, all_products)
```
cipherboy, line 92: As an alternative, you could use `rule_dirs.json` from the other utility and then walk only rules which apply to this OS. It would be faster than walking the entire `linux_os` tree and make the behavior align better :-)

utils/build_stig_control.py
```diff
@@ -0,0 +1,114 @@
...
+    prod_type = product_list[0]
+    product_yaml = product_yamls[prod_type]
+
+    rule_yaml = ssg.build_yaml.Rule.from_yaml(rule_file, product_yaml)
+    rule_obj['references'] = rule_yaml.references
+
+    return rule_obj
+
+
+def main():
```
cipherboy, line 56: I’d suggest moving main down so it is the last method.

utils/build_stig_control.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def handle_rule_yaml(product_list, product_yamls, rule_id, rule_dir, guide_dir):
+    rule_obj = {'id': rule_id, 'dir': rule_dir, 'guide': guide_dir}
+    rule_file = ssg.rules.get_rule_dir_yaml(rule_dir)
+
+    prod_type = product_list[0]
+    product_yaml = product_yamls[prod_type]
+
+    rule_yaml = ssg.build_yaml.Rule.from_yaml(rule_file, product_yaml)
```
cipherboy, line 50: You might need to call `update` so it resolves all references for the specified product. Needs to pass in a concrete product tho.

utils/build_stig_control.py
```diff
@@ -0,0 +1,114 @@
...
+    all_products = linux_products.union(other_products)
+    all_rule_dirs, product_yamls = walk_products(args.root, all_products)
+    known_rules = dict()
+    stigid_string = 'stigid@%s' % args.product
+
+    for rule_id, rule_dir, guide_dir, given_products in all_rule_dirs:
+        try:
+            rule_obj = handle_rule_yaml(given_products, product_yamls, rule_id,
+                                        rule_dir, guide_dir)
+        except ssg.yaml.DocumentationNotComplete:
```
cipherboy, line 100: I’ll leave it up to you how you want to handle this, but if you pass in a “Debug” field with value true, this exception won’t occur. You might want to condition it around an arg flag tho. (In the environment yaml).

utils/build_stig_control.py
```diff
@@ -0,0 +1,114 @@
...
+        f.flush()
+        f.close()
+
+
+def get_implemented_stigs(args):
+    linux_products, other_products = ssg.products.get_all(args.root)
+    all_products = linux_products.union(other_products)
+    all_rule_dirs, product_yamls = walk_products(args.root, all_products)
+    known_rules = dict()
+    stigid_string = 'stigid@%s' % args.product
```
cipherboy, line 94: If you call `update` above, you don’t need to specify by product here. Instead you can do it by just “stigid”.

utils/build_stig_control.py
cipherboy: You should also construct a properly formatted `local_env_yaml` Instead. Look at some of the build system stuff under `ssg` I think. Or the other utilities.

cipherboy (commented):
Some thoughts from the peanut gallery :-)

utils/build_stig_control.py
cipherboy: I guess I should say, generally we order top->bottom where bottom is main and it is a toposorted call tree… if you call a method a from method b, a should be above b. ;-0 `get_platform_rules` and some others are mis-organized. :-)

utils/build_stig_control.py
```diff
@@ -0,0 +1,144 @@
...
+
+    return parser.parse_args()
+
+
+def handle_rule_yaml(args, rule_id, rule_dir, guide_dir):
+    rule_obj = {'id': rule_id, 'dir': rule_dir, 'guide': guide_dir}
+    rule_file = ssg.rules.get_rule_dir_yaml(rule_dir)
+
+    product_dir = os.path.join(args.root, "products", args.product)
+    product_yaml_path = os.path.join(product_dir, "product.yml")
```
cipherboy, line 50: Not sure how long this takes to run, but as an optimization, `env_yaml` should be hoisted up and passed in once. Then it can be modified to include `rule_id` And such.

cipherboy: Look up `local_env_yaml` in the code base :-)

utils/build_stig_control.py
```diff
@@ -0,0 +1,145 @@
...
+
+def main():
+    args = parse_args()
+    check_files(args)
+
+    ns = {'checklist': 'http://checklists.nist.gov/xccdf/1.1'}
+    known_rules = get_implemented_stigs(args)
+    tree = ET.parse(args.manual)
+    root = tree.getroot()
+    controls = list()
```
cipherboy, line 113: Maybe refactor this out into a separate method?

utils/build_stig_control.py
```diff
@@ -0,0 +1,145 @@
...
+    output = dict()
+    output['policy'] = root.find('checklist:title', ns).text
+    output['title'] = root.find('checklist:title', ns).text
+    output['id'] = 'stig_%s' % args.product
+    output['source'] = 'https://public.cyber.mil/stigs/downloads/'
+    output['levels'] = list()
+    for level in ['high', 'medium', 'low']:
+        output['levels'].append({'id': level})
+    output['controls'] = controls
+    f = open(args.output, 'w')
```
cipherboy, line 137: Could use the with pattern:

```python3
with open(args.output, 'w') as f:
    f.write(...)
```

And you wouldn't need the flush/close stuff.

cipherboy (approved):
I think this looks fairly good. I've run the tool on Ubuntu and it seems to spit out a control file.

jan-cerny:
@ggbecker @Mab879 Why is this stuck?

Mab879:
I'm not sure, its ready to be merged.

JAORMX:
@Mab879 I'm doing some changes to the `controls` structure here https://github.com/ComplianceAsCode/content/pull/7506 maybe you might want to pick those up already?

JAORMX:
@Mab879 the aforementioned PR adding a status to the controls structure has merged. What do you think about using that instead of `automated`?

Mab879:
@JAORMX I can do that.

controls/stig_rhel8.yml
```diff
@@ -0,0 +1,2838 @@
...
+  - id: low
+
+controls:
+    -   id: RHEL-08-010000
+        levels:
+            - high
+        rules:
+            - installed_OS_is_vendor_supported
+        title: RHEL 8 must be a vendor-supported release.
+        automated: 'yes'
```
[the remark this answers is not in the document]
Mab879: Fixed.

jan-cerny (approved):

ggbecker:
Before start using the controls file within the profile file we need to update some rules that don't have the STIG id assigned and run the tool again:

```
configure_bind_crypto_policy
configure_crypto_policy
configure_kerberos_crypto_policy
configure_libreswan_crypto_policy
configure_ssh_crypto_policy
enable_dracut_fips_module
harden_sshd_ciphers_openssh_conf_crypto_policy
harden_sshd_macs_openssh_conf_crypto_policy
```

And then we need to bring variables into the controls file as well.

PR #7339, opened 2021-08-05
Move a SLES-12-010380 requirement from one rule to another

rumch-se:
#### Description:

- _With this update I moved SLES-12-030310 from one rule to another_

#### Rationale:

- _SLES-12-030310 was defined in rule gnome_gdm_disable_automatic_login, but according to its definition its correct place is in the rule gnome_gdm_disable_unattended_automatic_login_

### DIFF
--- a/linux_os/guide/system/software/gnome/gnome_login_screen/gnome_gdm_disable_unattended_automatic_login/rule.yml
+++ b/linux_os/guide/system/software/gnome/gnome_login_screen/gnome_gdm_disable_unattended_automatic_login/rule.yml
@@ -28,8 +28,8 @@ references:
     disa: CCI-000366
     nist: CM-6(b),CM-6.1(iv)
     srg: SRG-OS-000480-GPOS-00229
-    stigid@sle15: SLES-15-040430
     stigid@sle12: SLES-12-010380
+    stigid@sle15: SLES-15-040430
 
 ocil_clause: 'GDM allows users to automatically login or unattended login'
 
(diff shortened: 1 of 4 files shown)


ggbecker:
You need to update the profile files (e.g. stig.profile) that use this rule to select the new one.

rumch-se:
@ggbecker - Thank you for your message. I have done a second commit (a80abc0), I do believe that now everything will be OK.

products/sle12/profiles/stig.profile
```diff
@@ -185,7 +185,7 @@ selections:
     - file_permissions_system_commands_dirs
     - file_permission_user_init_files
     - ftp_present_banner
-    - gnome_gdm_disable_automatic_login
+    #   gnome_gdm_disable_automatic_login
```
ggbecker, line 188:
```suggestion
    - gnome_gdm_disable_unattended_automatic_login
```
I believe the idea is to select the new STIG referenced rule here.

linux_os/guide/system/software/gnome/gnome_login_screen/gnome_gdm_disable_unattended_automatic_login/rule.yml
```diff
@@ -21,13 +21,15 @@ rationale: |-
...
 identifiers:
+    cce@sle12: CCE-83245-1
     cce@sle15: CCE-85723-5
 
 references:
     disa: CCI-000366
     nist: CM-6(b),CM-6.1(iv)
     srg: SRG-OS-000480-GPOS-00229
     stigid@sle15: SLES-15-040430
+    stigid@sle12: SLES-12-010380
```
ggbecker, line 32:
```suggestion
    stigid@sle12: SLES-12-010380
    stigid@sle15: SLES-15-040430
```

we now have a gating test that detects if references are sorted and in this case sle12 needs to come first.

For future reference, you can run these two commands from the project's directory to sort references automatically:

`python utils/rule_dir_json.py`
`python utils/fix_rules.py sort_subkeys -y`

ggbecker (approved):

PR #7340, opened 2021-08-05
update version of rhel7 stig_gui profile

vojtapolasek:
#### Description:

- update version and sme of stig_gui in rhel7

#### Rationale:


ggbecker (approved):
LGTM

PR #7343, opened 2021-08-05
utils/add_platform_rule.py: fix selfLink discovery

mrogers950:
selfLink is now disabled on OCP, so we now look at oc's debug log to figure out the object path.


jhrozek (approved):
/lgtm
This is good, the only thing I'd like (eventually) would be a test that grabs a well-known object and checks that we can still get its API path. But let's unblock the script now!

PR #7344, opened 2021-08-05
Add a new rules RHEL-08-010001 and RHEL-07-020019

Mab879:
#### Description:

This adds a new rules for RHEL-08-010001 and RHEL-07-020019.

Looking for feedback on the `agent_mfetpd_running`, is there a better way to do this?

#### Rationale:
It seems that latest STIGs are not out on STIG Viewer or vaulted yet.


ggbecker (changes requested):
`mcafee-endpoint-security-software`, even though hyphen works for the folder name, it's more common to use underscore for the word separator aka [snake_case](https://en.wikipedia.org/wiki/Snake_case) in CaC.

ggbecker (changes requested):
For the package installation rule, I think it makes sense to rename it to `package_mcafeetp_installed` and you can probably reuse macros from other similar rules... I think there are macros for ocil and even description... of course the description can have the macro and additional text.

Another thing is that both rules contain almost the same text. It would be better if the description would match exactly what checks/remediations do.

For the process checking I think it makes sense to add at least a failing test scenario to confirm that the check is at least able to detect a failure, furthermore if it's possible to create a mock for the daemon for a passing test scenario, it would be even better.

linux_os/guide/system/software/integrity/endpoint_security_software/mcafee_security_software/mcafee_endpoint_security_software/agent_mfetpd_running/rule.yml
```diff
@@ -0,0 +1,40 @@
...
+    cce@rhel8: CCE-86261-5
+
+references:
+    disa: CCI-001233
+    nist: SI-2(2)
+    srg: SRG-OS-000191-GPOS-00080
+    stigid@rhel7: RHEL-07-020019
+    stigid@rhel8: RHEL-08-010001
+
+ocil_clause: 'virus scanning software is not installed or running'
```
ggbecker, line 29:
```suggestion
ocil_clause: 'virus scanning software is not running'
```

linux_os/guide/system/software/integrity/endpoint_security_software/mcafee_security_software/mcafee_endpoint_security_software/agent_mfetpd_running/rule.yml
```diff
@@ -0,0 +1,40 @@
...
+
+ocil: |-
+    To verify that McAfee Endpoint Security for Linux is installed
+    and running, run the following command(s):
+    <pre>$ sudo rpm -qa | grep -i mcafeetp</pre>
+    <pre>$ sudo ps -ef | grep -i mfetpd</pre>
```
ggbecker, line 35:
```suggestion
    To verify that McAfee Endpoint Security for Linux is
    running, run the following command:
    <pre>$ sudo ps -ef | grep -i mfetpd</pre>
```

linux_os/guide/system/software/integrity/endpoint_security_software/mcafee_security_software/mcafee_endpoint_security_software/package_mcafeetp_installed/rule.yml
```diff
@@ -0,0 +1,45 @@
...
+    - general: |-
+        Due to McAfee Endpoint Security for Linux (ENSL) being 3rd party software,
+        automated remediation is not available for this configuration check.
+
+platform: machine
+
+template:
+    name: package_installed
+    vars:
+        pkgname: mcafeetp
```
ggbecker, line 45:
```suggestion
        pkgname: mcafeetp
    backends:
        bash: "off"
        ansible: "off"
        anaconda: "off"
        puppet: "off"
        blueprint: "off"
```

let's disable the remediation completely.

ggbecker (approved):
LGTM. I tested this using mockup methods to run an arbitrary named program that the OVAL check was able to detect correctly.

Thanks for addressing all the feedback.

PR #7349, opened 2021-08-06
Sles 12 020130 1

rumch-se:
#### Description:

- _Review required_

#### Rationale:

- _The rule SLES-12-020130 was extended with ansible, bash and tests _


rumch-se:
I have to fix this request

PR #7351, opened 2021-08-06
Rule SLES-12-010890 added bash

rumch-se:
#### Description:

- _Revision required_

#### Rationale:

- _To the existing rule SLES-12-010890 was added - bash script for check and fix of the proposed recommendation_


ggbecker:
Since there are no test scenarios for this rule... the remediation cannot be tested. I'm assuming you did some tests on your own.

ggbecker (approved):

ggbecker:
> Since there are no test scenarios for this rule... the remediation cannot be tested. I'm assuming you did some tests on your own.

but actually our automated test would run this remediation on a fedora, so it probably wouldn't help much since the file permissions is handled differently (by `chkstat`).

PR #7353, opened 2021-08-06
Update CCEs and identifiers on rules that make up RHEL 8 CIS 4.1.15

alexhaydock:
#### Description:

- This updates CCEs and identifiers on rules which will be used to satisfy the RHEL 8 CIS 4.1.15 control as part of #6976 
- This relates to a comment made by @vojtapolasek here: https://github.com/ComplianceAsCode/content/pull/6976/files#r682440362

#### Rationale:

- These rules will be used to satisfy the RHEL 8 CIS 4.1.15 control but do not have RHEL 8 identifiers yet.


vojtapolasek (approved):
LGTM, failing checks are from OCP, I think they can be ignored here.

PR #7364, opened 2021-08-09
UBTU-20-010005: Adjust vlock_installed for Ubuntu

dodys:
#### Description:

- Move from hardcoded distro name to full_name variable.
- Add pkgname for Ubuntu.


linux_os/guide/system/accounts/accounts-physical/screen_locking/console_screen_locking/vlock_installed/rule.yml
```diff
@@ -32,9 +43,17 @@ references:
 
 ocil_clause: 'the package is not installed'
 
-ocil: '{{{ ocil_package(package="kbd") }}}'
+ocil: |-
+    {{% if 'ubuntu' in product %}}
+    {{{ ocil_package(package="vlock") }}}
```
jan-cerny, line 48: It would probably be better to declare it as a variable at the beginning.

jan-cerny (approved):

PR #7365, opened 2021-08-09
Adjust smartcard_configure_cert_checking OVAL for other distros

dodys:
#### Description:

- smartcard_configure_cert_checking OVAL was SUSE specific, changed it so other distros can use it as well.

### DIFF
--- a/linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/smartcard_configure_cert_checking/oval/shared.xml
+++ b/linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/smartcard_configure_cert_checking/oval/shared.xml
@@ -2,27 +2,25 @@
   <definition class="compliance" id="smartcard_configure_cert_checking" version="4">
     {{{ oval_metadata("Enable Smart Card Login") }}}
     <criteria comment="smart card authentication is configured" operator="AND">
-      {{% if product in ['sle12', 'sle15'] %}}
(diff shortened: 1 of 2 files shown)


linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/smartcard_configure_cert_checking/oval/shared.xml
```diff
@@ -1,34 +1,26 @@
 <def-group>
   <definition class="compliance" id="smartcard_configure_cert_checking" version="4">
-    <metadata>
-      <title>Enable Smart Card Login</title>
-      <affected family="unix">
-        <platform>multi_platform_sle</platform>
```
jan-cerny, line 6: I think that if you enable this on all platforms you should remove the rhel8 from the prodtype in the rule becasue on rhel8 neither the configuration file nor the package exist.

dodys: So, rhel8 is in the prodtype but the rule is not in any rhel8 profile. That's probably a leftover?
If that's right, I can remove rhel8 from the prodtype in a new commit.

dodys: I've removes rhel8 from prodtype in a new commit.

jan-cerny (approved):

PR #7228, opened 2021-07-10
Bug 1857179 - Improved ansible fix for banner files. Replace files on…

marcusburghardt:
#### Description:

This patch fix the Bug 1857179 - https://bugzilla.redhat.com/show_bug.cgi?id=1857179
The files / etc / issue and / etc / motd are replaced by the Ansible playbook, even if they are already correct, impacting the efficiency and integrity of the files once a new inode is created.
This patch improves the playbook using the copy module, which ensures that the file is only updated if the current content is different from the desired content.

#### Rationale:

This patch achieve the same goal with less code and more efficiency.
The copy module is part of ansible-base and included in all Ansible installations.

PR #7376, opened 2021-08-11
Check for Ansible Playbooks that remove a file but reintroduce it again

yuumasato:
#### Description:

- Check if any playbook removes a file and then adds it back again.
  The file removal is based on the 'file' module with 'state: absent', and the reintroduction of the file is based on 'lineinfile', 'blockinfile' and 'copy' modules.
- The test is only executed if `yamlpath` is installed.

#### Rationale:

- Related to #7228


matusmarhefka:
@yuumasato What about mentioning `yamlpath` installation through `pip` in the https://complianceascode.readthedocs.io/en/latest/manual/developer/02_building_complianceascode.html#installing-build-dependencies ?

matusmarhefka (approved):
Looks good :) Thanks for adding unit tests.

Issue #7348, opened 2021-08-06
Finish replacement of shared bash functions by Jinja macros


jan-cerny:
#### Description of problem:
To prevent code duplication in our Bash remediation, we use 2 ways of sharing code:
* shared bash functions, located in `/shared/bash_remediation_functions`
* [Jinja](https://jinja.palletsprojects.com/en/3.0.x/) macros, located in `shared/macros-bash.jinja` (see also other `.jinja` files in the `/shared` directory)

We prefer using Jinja macros. Over the time, we gradually converted some of the bash functions to Jinja macros. We would like to finish the conversion. We want to totally get rid of shared bash functions and use Jinja macros everywere. To do that, we need to finish conversion of the remaining shared bash functions, located in `/shared/bash_remediation_functions`. This will allow us to remove some code from the build system and remove the functions code from the generated data stream. Also, we could easily fix some shellcheck problems in future.

#### SCAP Security Guide Version:
current upstream as of 2020-08-06

#### Operating System Version:
all

#### Steps to Reproduce:

1. git grep '/usr/share/scap-security-guide/remediation_functions'

#### Actual Results:
Some bash remediation use the shared functions, some use Jinja macros.

#### Expected Results:
Shared bash remediation functions are not used anywhere. They're fully replaced by Jinja macros. The replacement is fully functionally equivalent. New Jinja macros are documented and tested.

#### Additional Information/Debugging Steps:
You can get inspired by PRs that replaced some bash by Jinja in past:
* https://github.com/ComplianceAsCode/content/pull/4810
* https://github.com/ComplianceAsCode/content/pull/4999
* https://github.com/ComplianceAsCode/content/pull/4998
* https://github.com/ComplianceAsCode/content/pull/4746
* https://github.com/ComplianceAsCode/content/pull/4734
* etc.

Keep in mind that in bash you can return only from a function. But if a Jinja macro doesn't expand to a function, you can't use return in that code.

jan-cerny:
Fixed in #7578

PR #7378, opened 2021-08-11
Convert Bash function replace_or_append to a Jinja macro

j-ode:
#### Description:

- Converted replace_or_append bash function and its calls to jinja macro

#### Rationale:

- #7348

#### Needs review:

- Some files that were calling the function still contain ". /usr/share/scap-security-guide/remediation_functions" --> I left it there and propose to remove this line from all files in the project once all the functions are converted to macros. Bad idea?


shared/templates/accounts_password/bash.template
```diff
@@ -3,7 +3,6 @@
 # strategy = restrict
 # complexity = low
 # disruption = low
-. /usr/share/scap-security-guide/remediation_functions
 {{{ bash_instantiate_variables("var_password_pam_" + VARIABLE) }}}
 
-replace_or_append '/etc/security/pwquality.conf' '^{{{ VARIABLE }}}' $var_password_pam_{{{ VARIABLE }}} '@CCENUM@' '%s = %s'
+{{{ bash_replace_or_append('/etc/security/pwquality.conf', '^{{{ VARIABLE }}}', "$var_password_pam_{{{ VARIABLE }}}", '@CCENUM@', '%s = %s') }}}
```
jan-cerny, line 8: I can see too much braces here. You are nesting them, that won't work. 

This file is a template. There are multiple rules that are using this template, they all have same bash remediation which differ only by variable `VARIABLE`  (I know, it's confusing :scream: ).

Check out the the "Templating" section in the documentation: https://complianceascode.readthedocs.io/en/latest/manual/developer/06_contributing_with_content.html#templating

For example, check `linux_os/guide/system/accounts/accounts-pam/password_quality/password_quality_pwquality/accounts_password_pam_dcredit/rule.yml`, then build the `rhel8` product, and then watch what got generated into `build/rhel8/fixes/bash/accounts_password_pam_dcredit.sh`. Currently, you have it broken.

You probably would have to do something similar as is already done on line 6, ie. concatenate the strings with the `VARIABLE` variable.

shared/templates/sysctl/bash.template
```diff
@@ -16,7 +15,7 @@
 # If {{{ SYSCTLVAR }}} present in /etc/sysctl.conf, change value to appropriate value
 #	else, add "{{{ SYSCTLVAR }}} = value" to /etc/sysctl.conf
 #
-replace_or_append '/etc/sysctl.conf' '^{{{ SYSCTLVAR }}}' "$sysctl_{{{ SYSCTLID }}}_value" '@CCENUM@'
+{{{ bash_replace_or_append('/etc/sysctl.conf', '^{{{ SYSCTLVAR }}}', "$sysctl_{{{ SYSCTLID }}}_value", '@CCENUM@') }}}
```
jan-cerny, line 18: This is a similar problem you have in the accounts_password template.

jan-cerny (changes requested):
Hi @j-ode, this is a great start!

Please change the PR title to something sensible, for example: "Convert Bash function 
replace_or_append to a Jinja macro".

Focus on the content generated using templates.

See other comments below.

shared/templates/accounts_password/bash.template
```diff
@@ -3,7 +3,6 @@
 # strategy = restrict
 # complexity = low
 # disruption = low
-. /usr/share/scap-security-guide/remediation_functions
 {{{ bash_instantiate_variables("var_password_pam_" + VARIABLE) }}}
```
jan-cerny, line 6: I have discovered that we can't remove the `. /usr/share/scap-security-guide/remediation_functions` line because there is the Jinja macro `bash_instantiate_variables` that gets expanded to a bash function `populate`. Please remove this removal. In other words, add it back.

This is a thing that could be revealed by SSGTS test suite. Unfortuantely, the GH action `SSGTS / Run Tests (pull_request)` has terminated unexpectedly, probably due to a problem that isn't related to your PR.

shared/templates/sysctl/bash.template
```diff
@@ -3,7 +3,6 @@
 # strategy = disable
 # complexity = low
 # disruption = medium
-. /usr/share/scap-security-guide/remediation_functions
```
jan-cerny, line 6: This is the same, we need this because the `bash_instantiante_variables` contains a bash function call. :scream_cat:

jan-cerny (changes requested):

jan-cerny (approved):
LGTM. Great achievement!

I would merge it but we will just wait with that for a while until the strategy is determined.

shared/macros-bash.jinja
```diff
@@ -686,3 +686,75 @@ dpkg-query --show --showformat='${db:Status-Status}\n' "{{{ pkgname }}}" 2>/dev/
...
+      {{{ bash_replace_or_append('/etc/sysconfig/selinux', '^SELINUX=', "$var_selinux_state", '@CCENUM@', '%s=%s') }}}
+ 
+#}}
+{{%- macro bash_replace_or_append(config_file, key, value, cce, format='%s = %s') -%}}
+case_insensitive_mode=yes sed_case_insensitive_option='' grep_case_insensitive_option=''
+
+if [ "$case_insensitive_mode" = yes ]; then
+    sed_case_insensitive_option="i"
+    grep_case_insensitive_option="-i"
+fi
```
matejak, line 726: Just out of curiosity, what is this? It looks like that we are always in case insensitive mode, and therefore `sed_case_insensitive_option` and `grep_case_insensitive_option` could be macros instead of shell variables.

shared/macros-bash.jinja
```diff
@@ -686,3 +686,75 @@ dpkg-query --show --showformat='${db:Status-Status}\n' "{{{ pkgname }}}" 2>/dev/
...
+    sed_command+=('--follow-symlinks')
+fi
+
+# Test that the cce arg is not empty or does not equal @CCENUM@.
+# If @CCENUM@ exists, it means that there is no CCE assigned.
+if [ -n "{{{ cce }}}" ] && [ "{{{ cce }}}" != '@CCENUM@' ]; then
+    cce="{{{ cce }}}"
+else
+    cce="CCE"
+fi
```
matejak, line 741: We are no longer in a function - we know whether `cce` is defined or not at the time when the macro is used, so there can be a cce assignment sub-macro.

matejak:
I like the PR, especially the tests. I have also added some remarks, which should be addressed, but that doesn't block the PR from merging.

shared/macros-bash.jinja
jan-cerny: Nice finding! It was there before. Case insensitive mode is always yes.  So I guess it could be removed completely, and substitute the bash variables by their values.

shared/macros-bash.jinja
jan-cerny: Hi @matejak , thanks for pointing this out. It's an interesting piece of code. 

The `@CCENUM@` is a string that is replaced at a later phase of the build process by the rule CCE number. It's done by the XSLT template /shared/transforms/xccdf-addremediations.xslt on lines 68 - 81. Notice that in `/build/<product>/fixes/bash` you will see the `@CCENUM@` string, but in the generated SCAP source data stream (`ssg-<product>-ds.xml`) you will see that it got replaced by the rule's CCE number. Or, it's replaced by an empty string if the rule doesn't have a CCE number. For example, in Fedora we don't assign CCEs.

Another observation is that the bash function or now the jinja macro is always called with the third parameter set to `@CCENUM@`, which means that this parameter is probably useless and we could always use `@CCENUM@` in the macros code.

We will probably need to change the logic here to get the desired result in the built SCAP source data stream. The desired result is: If the rule has a CCE number, set the `cce` bash variable to the value of the CCE number. If the rule doesn't have a CCE number, set the `cce` bash variable to string `"CCE"`

j-ode: Commit 2ea91b3 should achieve the desired result; however, it generates weird code in the resulting remediation. If anyone has a more elegant solution, it would be appreciated :eyes:

jan-cerny:
:+1:

shared/macros-bash.jinja
matejak: Answering to myself - #2689  explains that. Therefore, I suggest renaming the macro to `replace_or_append_case_insensitive`

PR #7383, opened 2021-08-12
add kickstarts for rhel8 CIS profiles

vojtapolasek:
#### Description:

- update the cis kickstart to reference level 2 - Server
- add other 3 kickstarts

#### Rationale:

Relates to #6976


jan-cerny:
Looks great but I don't know if it's OK to merge it before #6976 is merged because until that it would reference a non-existing profile.

yuumasato (approved):
Looks good, and they worked for me.

PR #7388, opened 2021-08-16
products/sle15/profiles/stig.profile: Update according to U_SLES_15_STIG_V1R3 Manual

apatard:
#### Description:

The current checks for the STIG manual for SLES15 is giving failures due to not check for what's
exactly in the manual and most importantly, it's doing a check which doesn't exist anymore.

In details:

- SLES-15-040270 has been removed
- SLES-15-010280 check is looking for ClientAliveInterval set to 600
- SLES-15-020290 is about only 'password requisite pam_cracklib.so', nothing about 'retry='
  and no rules are mentioning it
- SLES-15-020020 remediation is about maxlogins set to 10
- SLES-15-010400 checks and remediation talks about maxpoll set to 16


Mab879 (approved):
LGTM

PR #7389, opened 2021-08-16
locking_out_password_attempts/accounts_password_pam_unix_remember/oval/shared.xml: sles15 fix

apatard:
#### Description:

The STIG manual for SLES-15-020250 says:
    
Check that the SUSE operating system prohibits the reuse of a password for a minimum of five generations with the following command:

```
> grep pam_pwhistory.so /etc/pam.d/common-password
   password requisite pam_pwhistory.so remember=5 use_authtok
```

but the check is looking for /etc/pam.d/system-auth, so update the check to use the right file for sle15.


apatard:
fwiw, I've yet to understand why the bash test is failing and not the ansible one. Since my change should be transparent for non-sles15 distro, I think that the test was already failing/broken.

jan-cerny (approved):
Thanks for fixing the failing tests (rhel8_wrong_value.fail.sh and wrong_value.fail.sh). Now all the test scenarios pass on both RHEL 7 and RHEL 8 with both Bash and Ansible remediation.

PR #7391, opened 2021-08-16
OCP4: Disable `kubelet_disable_hostname_override` OVAL and update verbiage

### LINKED ISSUE #7390: OpenShift CI consistently fails on the `kubelet_disable_hostname_override` rule
#### Description of problem:

the OpenShift CI consistently fails with the following log:

```
     helpers.go:751: E2E-FAILURE: The expected result for the kubelet_disable_hostname_override rule didn't match. Expected 'PASS', Got 'FAIL' 
```

this is because a [recent change](https://github.com/openshift/machine-config-operator/pull/2694) modified the expected behavior to actually override the hostname which is against what the CIS guidelines specify. However, in this specific case, the overriding is needed to avoid upgrade issues and it's used to ensure that the hostname stays the same (and thus the PKI would keep working).

JAORMX:
This check is no longer applicable as we need the override to ensure
upgrades work for OpenStacka and AWS. The verbiage of the rule has been
updated to reflect it.

Closes #7390


JAORMX:
Pull-request updated, HEAD is now cc907ce5e649a91060f79431f4580048f601b81b

applications/openshift/kubelet/kubelet_disable_hostname_override/rule.yml
```diff
@@ -2,29 +2,25 @@ documentation_complete: true
...
+    and keeps working throughout the lifecycle of the cluster. IP addresses
+    are not used, and hence this makes it easier for security analysts to
+    associate kubelet logs with the appropriate node.
 
 rationale: |-
-    Allowing hostnames to be overrided creates issues around resolving nodes
+    Allowing hostnames to be overridden creates issues around resolving nodes
     in addition to TLS configuration, certificate validation, and log correlation
-    and validation.
+    and validation. However, in some cases explicit overriding of this parameter
```
nkinder, line 18: (nit) "overriding of this parameter **is** necessary".

mrogers950: aside from this, LGTM

matejak (approved):
LGTM, the PR removes check, some optional metadata, and fixes typos - business as usual.
Then, it changes the `default_result`, and preserves the absence of trailing newline, which is the most suspicious part of the PR.

Merging, thanks!

PR #7392, opened 2021-08-16
Add ISM Official kickstarts

matejak:
Kickstarts are based on their e8 siblibgs. The e8 profile requires Grub password, which I believe is not needed by any rule, so the ISM profile disables this.


Achilleasein (approved):
Looks good, approved.

PR #7393, opened 2021-08-16
Remove a spurious whitespace trim

matejak:
The first line of the if- block ended up in the metadata comment.


ggbecker:
Can you also remove it from the `else` and `endif` statements? I tested here using this [online jinja parser](http://jinja.quantprogramming.com/) and when `init_system` is not `systemd`, the problem is also present.

matejak:
The `endif` with trimming at the end of the file could discard trailing newline, which would be bad, but I think that the `else` trim is harmless - but so is a newline, so I have removed it as well.

ggbecker (approved):
LGTM

PR #7394, opened 2021-08-16
WIP: grub2 bootloader argument template uefi support

### LINKED ISSUE #6716: grub2_bootloader_argument OVAL template lacks support for UEFI systems
#### Description of problem:
This very issue has already been discussed on the [mailing list](https://lists.fedorahosted.org/archives/list/scap-security-guide@lists.fedorahosted.org/thread/VZDPWODPXZXKAWLYC6FCSS5NQMFARRLP/). However I couldn't find any issues/PRs directly addressing it.

In short: The grub2_bootloader_argument OVAL template has a hardocded path to the `grub.cfg` config file that is only valid for non-uefi systems, namely `/boot/grub2/grub.cfg`. While in UEFI systems it is `/boot/efi/EFI/redhat/grub.cfg`.

This causes false negatives on UEFI systems for the rules that use this template.

0.1.54

Testing with OL7.9

#### Steps to Reproduce:
1. On a UEFI system, run a scan with any of the rules that use the grub2_bootloader_argument template (e.g. `grub2_vsyscall_argument`, `grub2_page_poison_argument`, `grub2_audit_backlog_limit_argument`).

#### Actual Results:
Scan always results in `fail` and remediation in `error` since the OVAL check looks for the config entry in a file that does not exist.

#### Expected Results:
Scan should result in `pass`/`fail` depending on if the looked for configuration is present or not

A possible approach to fix this was suggested on the email thread linked above:

_"...So the approach forward could be to add UEFI versions of the rules
equivalent to the BIOS kernel argument rules; and updating the template to
handle both cases (by passing different arguments for each case, for
example)..."_

I will gladly implement this fix if it is an acceptable approach.

freddieRv:
#### Description:

- Add support fort both BIOS/UEFI in grub2_bootloader_argument oval template
-  Created uefi and non-uefi versions of following rules:
    - grub2_enable_iommu_force
    - grub2_kernel_trust_cpu_rng
    - grub2_pti_argument
    - grub2_vsyscall_argument
- Updated rule.yml files to use grub2_boot_path and grub2_uefi_boot_path respectively
- Added an 'id_suffix' to oval definition in grub2_bootloader_argument oval template so that there are no repeated id's between the uefi and non-uefi rules

#### Rationale:

- Even after #6986, the grub2_bootloader_argument template supported only the config path defined in `grub2_boot_path` at product level. With this update, the uefi and non-uefi versions of the rules use `grub2_boot_path` and `grub2_uefi_boot_path` respectively.

- Fixes #6716


shared/templates/grub2_bootloader_argument/template.py
```diff
@@ -8,4 +8,9 @@ def preprocess(data, lang):
         data["escaped_arg_name_value"] = data["arg_name_value"].replace(".", "\\.")
         # replace . with _, this is used in test / object / state ids
         data["sanitized_arg_name"] = ssg.utils.escape_id(data["arg_name"])
+
+        data['id_suffix'] = ''
+        if 'uefi' in data['_rule_id']:
+            data['id_suffix'] = 'uefi'
+
```
freddieRv, line 15: This feels a bit hacky. Another approach could be to add the `id_suffix` as another template parameter

linux_os/guide/system/bootloader-grub2/non-uefi/file_groupowner_grub2_cfg/rule.yml
```diff
@@ -8,7 +8,7 @@ description: |-
     The file <tt>{{{ grub2_boot_path }}}/grub.cfg</tt> should
     be group-owned by the <tt>root</tt> group to prevent
     destruction or modification of the file.
-    {{{ describe_file_group_owner(file="{{{ grub2_boot_path }}}/grub.cfg", group="root") }}}
+    {{{ describe_file_group_owner(file=grub2_boot_path+"/grub.cfg", group="root") }}}
```
cipherboy, line 11: Damn I messed that up. Great catch @freddieRv!

linux_os/guide/system/bootloader-grub2/uefi/grub2_uefi_kernel_trust_cpu_rng/oval/shared.xml
```diff
@@ -0,0 +1,65 @@
...
+    <criteria operator="OR" >
+      <criteria operator="AND">
+        <criterion comment="check if the option is compiled in" test_ref="test_trust_cpu_rng_compiled_in" />
+        <criterion comment="check if the option is not overridden through a boot parameter" test_ref="test_trust_cpu_rng_boot_param_off" negate="true" />
+      </criteria>
+      <criterion comment="check if the option is configured as a boot parameter" test_ref="test_trust_cpu_rng_boot_param_on" />
+    </criteria>
+  </definition>
+
+  <local_variable id="var_kernel_config_file" comment="the config file for currently running kernel" datatype="string" version="1">
```
matejak, line 13: Unfortunately, we can't detect identical OVAL objects and deduplicate them. The variable existing twice in the OVAL (although it is a "local" variable) creates problems, so you can work around this by duplicating it under a different name:

```suggestion
  <local_variable id="var_{{{ rule_id }}}_kernel_config_file" comment="the config file for currently running kernel" datatype="string" version="1">
```

linux_os/guide/system/bootloader-grub2/uefi/grub2_uefi_kernel_trust_cpu_rng/oval/shared.xml
```diff
@@ -0,0 +1,65 @@
...
+  </local_variable>
+
+  <unix:uname_object id="trust_cpu_rng_uname" comment="determine version and build of currently running kernel" version="1" />
+
+  <ind:textfilecontent54_test id="test_trust_cpu_rng_compiled_in" check="all" check_existence="all_exist" version="1" comment="check if the parameter was set at the compile time for current kernel">
+    <ind:object object_ref="object_trust_cpu_rng_compiled_in" />
+  </ind:textfilecontent54_test>
+
+  <ind:textfilecontent54_object id="object_trust_cpu_rng_compiled_in" version="1">
+    <ind:filepath operation="equals" var_ref="var_kernel_config_file" />
```
matejak, line 27:
```suggestion
    <ind:filepath operation="equals" var_ref="var_{{{ rule_id }}}_kernel_config_file" />
```

shared/templates/grub2_bootloader_argument/oval.template
```diff
@@ -3,63 +3,63 @@
     {{{ oval_metadata("Ensure " + ARG_NAME_VALUE + " is configured in the kernel line in /etc/default/grub.") }}}
     <criteria operator="AND">
       {{% if product in ["rhel7", "ol7"] or 'ubuntu' in product %}}
-        <criterion test_ref="test_grub2_{{{ SANITIZED_ARG_NAME }}}_argument_grub_cfg"
-        comment="Check if {{{ ARG_NAME_VALUE }}} is present in the boot parameters in the {{{ grub2_boot_path }}}/grub.cfg for all kernels" />
+        <criterion test_ref="test_grub2_{{{ SANITIZED_ARG_NAME + ID_SUFFIX }}}_argument_grub_cfg"
```
matejak, line 6: This `SANITIZED_ARG_NAME + ID_SUFFIX` is very ugly, just use jinja's `{{% set unique_oval_id_stem=SANITIZED_ARG_NAME + ID_SUFFIX %}} (or something like that) -  that will be much better

linux_os/guide/system/bootloader-grub2/uefi/grub2_uefi_enable_iommu_force/rule.yml
```diff
@@ -0,0 +1,35 @@
...
+rationale: |-
+    On x86 architectures, activating the I/OMMU prevents the system from arbritrary accesses potentially made by
+        hardware devices.
+
+severity: unknown
+
+identifiers:
+    cce@rhel7: CCE-82351-8
+    cce@rhel8: CCE-83920-9
+    cce@rhel9: CCE-83844-1
```
matejak, line 18: How come that those CCEs have appeared here, but they are not taken from the CCE pool?

You can either remove them, or you can assign them using the `utils/fix_rules.py` script (but you may need to rebase before that).

matejak (changes requested):
Thank you for the PR, we appreciate it, some things need to be fixed, but the direction is good overall.

freddieRv:
Thank you for the feedback @matejak ! I will address your comments and push the new changes shortly

matejak:
The PR has developed conflicts, and it hasn't been active for a long period of time. I am closing it, but don't hesitate to reopen if conflicts are resolved, and feedback is somehow addressed.

PR #7310, opened 2021-07-30
Add support for multiple profiles in TestSuite combined mode

yuumasato:
#### Description:

- Add support for testing multiple profiles at once on combined mode.
  - The combined mode is the only Test Suite run mode that didn't support multiple inputs

#### Rationale:

- After #7218,  execution of combined mode got broken, specifically https://github.com/ComplianceAsCode/content/commit/17f46a62362 made the `options.target` a list, which `combined` mode did not support.
```shell
$ python3 test_suite.py combined --libvirt qemu:///session rhel84 --datastream ../build/ssg-rhel8-ds.xml ospp
Setting console output to log level INFO
INFO - The base image option has not been specified, choosing libvirt-based test environment.
INFO - Logging into /home/wsato/git/content/tests/logs/combined-custom-2021-07-30-1619/test_suite.log
Traceback (most recent call last):
  File "/home/wsato/git/content/tests/test_suite.py", line 396, in <module>
    main()
  File "/home/wsato/git/content/tests/test_suite.py", line 392, in main
    options.func(options)
  File "/home/wsato/git/content/tests/ssg_test_suite/combined.py", line 101, in perform_combined_check
    if not profile.startswith(OSCAP_PROFILE):
AttributeError: 'list' object has no attribute 'startswith'
```

PR #7395, opened 2021-08-16
Fix combined mode execution in SSGTS

cipherboy:
While enabling GH-action based execution of the SSGTS, an option for
executing multiple rules was added, allowing for comma-separated rules
to be passed via the command line. This doesn't behave well with
combined mode, which expected a single target (a string) versus a list.

Update combined mode to understand multiple arguments via the CLI and
execute each profile separately.

### DIFF
--- a/tests/ssg_test_suite/combined.py
+++ b/tests/ssg_test_suite/combined.py
@@ -85,6 +85,8 @@ class CombinedChecker(rule.RuleChecker):
 def perform_combined_check(options):
     checker = CombinedChecker(options.test_env)
 
+    # These options are static between different runs under different
+    # profiles.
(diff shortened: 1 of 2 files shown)


ggbecker (approved):
LGTM. Thanks for the fix.

yuumasato:
There was https://github.com/ComplianceAsCode/content/pull/7310 which had a similar approach.

cipherboy:
@yuumasato Ah, apologies, hadn't seen it. :/ We can take yours if you want to push a revert commit of mine.

yuumasato:
I talked to @matusmarhefka as well and he was in favor of the approach without commas.
Proposing revert.

PR #7398, opened 2021-08-17
Pcidss Configure Crypto Rules

anivan-suse:
#### Description:

 - Add following PCI-DSS Rules to sle15
 -      package_libreswan_installed
 -      configure_libreswan_crypto_policy
 -      configure_openssl_crypto_policy
 -      configure_ssh_crypto_policy


products/sle15/profiles/pci-dss.profile
```diff
@@ -33,9 +33,6 @@ selections:
     - accounts_password_pam_unix_remember
     - accounts_passwords_pam_faillock_deny
     - accounts_passwords_pam_faillock_unlock_time
-    - accounts_password_pam_ucredit
-    - accounts_password_pam_minlen
-    - accounts_password_pam_dcredit
```
ggbecker, line 38: This probably shouldn't be here

anivan-suse: Yes - you are right. I didn't see this before, but it looks like those rules are already above in the file, I think I may have moved them but not sure how to get rid of this?

matejak: Simply copy-paste the selections from here to where they belong in the actual file, and make another commit. I think that we can squash commits in this PR at the end, as the PR represents only one logical change.

anivan-suse: Copy-pasted them, and it has the same issue as here just on different lines :)

Mab879 (changes requested):
Please rebase this request on `master` and resolve the conflicts.

Mab879 (approved):

PR #7400, opened 2021-08-17
Remove kubelet_disable_hostname_override rule

JAORMX:
This check is no longer applicable as we need the override to ensure
upgrades work for OpenStacka and AWS.


jhrozek:
FWIW, I think that removing the rule is better than having it around as manual but without any clear steps, because what we tend to see if that people don't like manual checks and expect more or less automated compliant/non-compliant answer. A manual check with no instructions would likely just increase the support load, so I think it's even better to remove the check completely.

jhrozek (commented):
don't you also need to remove the rule from the profiles?

JAORMX:
Pull-request updated, HEAD is now 1e98036fac6467e450f96d22ef26fbc06b1d1b9b

jhrozek (approved):
Locally the build went fine, let's see if CI goes through this time.

Issue #7408, opened 2021-08-18
sshd_lineinfile template is using wrong ssh_config.d filename


comps:
#### Description of problem:

```
$ grep 'sshd_config.d/hardening' -R *
shared/templates/sshd_lineinfile/bash.template:touch /etc/ssh/sshd_config.d/hardening
shared/templates/sshd_lineinfile/bash.template:        path="/etc/ssh/sshd_config.d/hardening",
```
The file should be called `<number>-hardening.conf` as the order of files in `ssh_config.d` matters and only files with `.conf` are included:
```
# grep Include /etc/ssh/sshd_config
Include /etc/ssh/sshd_config.d/*.conf
```

#### SCAP Security Guide Version:
Latest master, 9bdb45c683.

#### Operating System Version:
RHEL-9 / Fedora

ggbecker:
Update https://github.com/ComplianceAsCode/content/blob/9bdb45c683fc7a31392ea04f5861fc51e842e0bd/shared/templates/sshd_lineinfile/bash.template#L15

to something like:

`path="/etc/ssh/sshd_config.d/00-hardening.conf", `

matejak:
That name really sucks, so what about `60-redhat_hardening.conf` as there is `50-redhat.conf` already?

comps:
Note that you might need to comment out those options in `sshd_config` during remediation, and definitely use lower numbers as, for most options, the first occurence applies and others are ignored.

ggbecker:
> That name really sucks, so what about `60-redhat-hardening.conf` as there is `50-redhat.conf` already?

embedding `redhat` to the filename doesn't seem to be the best solution from a community POV. So far it's only enabled on Fedora and RHEL9, but in future it can be enabled in more platforms.

ggbecker:
> Note that you might need to comment out options changed by this file in `sshd_config` during remediation, and definitely use lower numbers as, for most options, the first occurence applies and others are ignored.

This should be done by the following lines if I'm not mistaken https://github.com/ComplianceAsCode/content/blob/9bdb45c683fc7a31392ea04f5861fc51e842e0bd/shared/templates/sshd_lineinfile/bash.template#L12-L13

but they completely remove the lines instead of commenting out.

matejak:
> embedding `redhat` to the filename doesn't seem to be the best solution from a community POV. So far it's only enabled on Fedora and RHEL9, but in future it can be enabled in more platforms.

[...]

PR #7410, opened 2021-08-18
Handle sshd directory config basenames better

matejak:
The check should consider only files matching `.*\.conf`.

The default basename should be valid as well - on Red Hat systems, there is `50-redhat.conf`, so `60-complianceascode.conf` seems to be a generally good fit.

Fixes: #7408

### DIFF
--- a/shared/templates/sshd_lineinfile/bash.template
+++ b/shared/templates/sshd_lineinfile/bash.template
@@ -11,8 +11,9 @@ mkdir -p /etc/ssh/sshd_config.d
 touch /etc/ssh/sshd_config.d/hardening
 {{{ lineinfile_absent("/etc/ssh/sshd_config", line_regex, insensitive=true) }}}
 {{{ lineinfile_absent_in_directory("/etc/ssh/sshd_config.d", line_regex, insensitive=true) }}}
+{{%- set hardening_config_basename = "00-complianceascode-hardening.conf" %}}
 {{{ set_config_file(
-        path="/etc/ssh/sshd_config.d/hardening",
+        path="/etc/ssh/sshd_config.d/" ~ hardening_config_basename,
         parameter=PARAMETER,
         value=VALUE,
         create=true,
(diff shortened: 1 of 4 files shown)


cipherboy (approved):
I like this idea.

matejak:
so for some reasons test scenarios don't pass, so please don't merge yet.

shared/templates/sshd_lineinfile/bash.template
```diff
@@ -11,8 +11,9 @@ mkdir -p /etc/ssh/sshd_config.d
 touch /etc/ssh/sshd_config.d/hardening
 {{{ lineinfile_absent("/etc/ssh/sshd_config", line_regex, insensitive=true) }}}
 {{{ lineinfile_absent_in_directory("/etc/ssh/sshd_config.d", line_regex, insensitive=true) }}}
+{{%- set hardening_config_basename = "60-complianceascode-hardening.conf" %}}
```
ggbecker, line 14: I think it's ok to have a higher number here since our remediation would remove any offending occurrences from any other file including the ones with lower number that would take precedence in the configuration loading. If *not* then I'd suggest we use a lower number.

ggbecker (commented):
I have found some more occurrences of `oval_line_in_file_object` in the project in custom SSHD rules that may need the `filepath_regex` option to be updated. Maybe they don't need any additional change, I just wanted to be sure that we don't miss anything.

shared/templates/sshd_lineinfile/bash.template
matejak: I somehow thought that it is the last directive that determines what's going to happen, but it's the exact opposite - as documented, and also as verified.
I will therefore change the value of `60-comp...` to a universal `00-comp...`.

matejak:
> I have found some more occurrences of `oval_line_in_file_object` in the project in custom SSHD rules that may need the `filepath_regex` option to be updated. Maybe they don't need any additional change, I just wanted to be sure that we don't miss anything.

That's OK, the default behavior is not to use any regex, unless `filepath_regex` is set to a value, nothing needs to change. Also sshd directory config is a relatively new thing, so it is not deeply embedded in our build system.
However, we have a terminology mess, when an innocently-looking argument `path` is actually `filepath` most of the time, but sometimes it may be indeed `path`. But an attempt to tackle this is work for another PR.

ggbecker (approved):
LGTM. Thank you for the comments.

PR #7411, opened 2021-08-18
Added rule for RHEL-08-010400

Mab879:
#### Description:

Added rule for RHEL-08-010400

#### Rationale:
https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_8_security_technical_implementation_guide/V-230274


linux_os/guide/services/sssd/sssd_certificate_verification/bash/shared.sh
```diff
@@ -0,0 +1,33 @@
...
+
+found=false
+for f in /etc/sssd/sssd.conf /etc/sssd/conf.d/*.conf; do
+	if [ ! -e "$f" ]; then
+		continue
+	fi
+	cert=$( awk '/^\s*\[/{f=0} /^\s*\[sssd\]/{f=1} f{nu=gensub("^\\s*certificate_verification\\s*=\\s*ocsp_dgst\\s*=\\s*(\\S+).*","\\1",1); if($0!=nu){cert=nu}} END{print cert}' $f )
+	if [ -n "$cert" ] ; then
+		if [ "$cert" != $var_sssd_certificate_verification_digest_function ] ; then
+			sed -i "s/^certificate_verification\s*=.*/certificate_verification = ocsp_dgst = $var_sssd_certificate_verification_digest_function/" $f
```
jan-cerny, line 20: I think that the "$f" should be quoted.

linux_os/guide/services/sssd/sssd_certificate_verification/oval/shared.xml
```diff
@@ -0,0 +1,30 @@
...
+
+    <ind:textfilecontent54_test check="all" check_existence="all_exist" comment="test the value of
+        certificate_verification in sssd configuration" id="test_{{{rule_id}}}" version="1">
+        <ind:object object_ref="obj_{{{rule_id}}}" />
+        <ind:state state_ref="state_{{{rule_id}}}" />
+    </ind:textfilecontent54_test>
+
+    <ind:textfilecontent54_object id="obj_{{{rule_id}}}" version="1">
+        <ind:filepath operation="pattern match">^/etc/sssd/(sssd|conf\.d/.*)\.conf$</ind:filepath>
+        <ind:pattern operation="pattern match">^[\s]*\[sssd](?:[^\n\[]*\n+)+?[\s]*certificate_verification\s*=\s*ocsp_dgst\s*=\s*(\w+)$</ind:pattern>
```
jan-cerny, line 19: I think it isn't needed to surround the `\s` by `[` and `]` if the `\s` is the only member of the group. Be consistent with the rest of the expression.

jan-cerny (commented):
It looks amazing.

I have only minor problems.

The tests pass for me locally:

```
[jcerny@thinkpad scap-security-guide{pr/7411}]$ python3 tests/test_suite.py rule --libvirt qemu:///system ssgts_rhel8 sssd_certificate_verification
INFO - Logging into /home/jcerny/work/git/scap-security-guide/logs/rule-custom-2021-08-19-0956/test_suite.log
INFO - xccdf_org.ssgproject.content_rule_sssd_certificate_verification
INFO - Script correct_value.pass.sh using profile (all) OK
INFO - Script not_configured.fail.sh using profile (all) OK
INFO - Script wrong_section.fail.sh using profile (all) OK
```

But, in the github Actions job, they're all notapplicable. @mildas Any idea why is that happening? Is it because the VM doesn't have SSSD installed? Should Mathhew add sssd to packages section in the scenarios?

mildas:
@jan-cerny The rule is in group with `platform: sssd` which needs to have installed `sssd-common` package. So yes, it is caused by the missing package.
Please, @Mab879 add `# packages = sssd-common` to all test scenarios.

ggbecker (commented):
One test scenario suggestion is to have something like this in the test:

```
echo -e "[sssd]\ndifferent_option = test\ncertificate_verification = ocsp_dgst=sha1" >> /etc/sssd/sssd.conf
```

To check if the OVAL is able to detect a configuration item even if it comes after other configuration options.

Also, I'm trying to find the documentation around the syntax for this particular case, if it's allowed to have spaces between the equal sign from `ocsp_dgst=sha1`, but with no success so far.

jan-cerny (approved):

PR #7427, opened 2021-08-20
OCP4: Add rule that checks for audit error alerts

JAORMX:
This checks that an alert exists by checking for prometheus rules. This
is needed to address AU-5.

From OCP 4.9 there is now a prometheus rule that checks this by default
given that https://github.com/openshift/cluster-kube-apiserver-operator/pull/1166
merged. So, the e2e test passes by default.

However, in older versions of Openshift this rule doesn't exist, so an
appropriate remediation was added. Note, that for that to work,
https://github.com/openshift/compliance-operator/pull/684 needs to merge
and be released.


jhrozek (approved):
/lgtm

PR #7435, opened 2021-08-20
Make possible to use pip3 installed modules when packaging

marcusburghardt:
#### Description:

This PR does a small update in the logic used by CMake to detect Python modules, giving more flexibility to use Python modules installed via pip3.

#### Rationale:

There are situations during the development and testing phases where is convenient to use Python modules installed via pip3.
For example, when the desired module is not available on the official repositories. However, these modules are not recognized when building packages. This FIX work around this in a flexible and simple way.

More details in the documentation update from this PR.

### DIFF
--- a/cmake/FindPythonModule.cmake
+++ b/cmake/FindPythonModule.cmake
@@ -7,6 +7,17 @@ function(find_python_module module)
 		if(ARGC GREATER 1 AND ARGV1 STREQUAL "REQUIRED")
 			set(PY_${module}_FIND_REQUIRED TRUE)
 		endif()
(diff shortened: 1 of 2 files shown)


jan-cerny:
Hi, What is the use case in which you want to build rpms with external dependencies?

marcusburghardt:
> Hi, What is the use case in which you want to build rpms with external dependencies?

Hi Jan, I detailed an use case in the documentation sent in this PR:
https://github.com/ComplianceAsCode/content/pull/7435/files#diff-125cb981d97ac7771e69135e66d4195d0f71b840d7ca6e05bd6c90fbfc495001

jan-cerny (approved):

PR #7450, opened 2021-08-23
FedRAMP SC-17: Add Ingress CA replacement rule

mrogers950:
@jhrozek @JAORMX


applications/openshift/networking/default_ingress_ca_replaced/rule.yml
```diff
@@ -0,0 +1,47 @@
...
+
+  The Ingress Operator is the component responsible for enabling external
+  access to OpenShift Container Platform cluster services. The aforementioned
+  operator creates an internal CA and issues a wildcard certificate that is
+  valid for applications under the .apps sub-domain. Both the web console
+  and CLI use this certificate as well. The certificate and key would need
+  to be replaced since a certificate coming from a trusted provider is
+  needed.
+
+  <tt>https://docs.openshift.com/container-platform/latest/security/certificates/replacing-default-ingress-certificate.html</tt>
```
JAORMX, line 21: There is a specific jinja template for this called `weblink`:

```
{{{ weblink(link="https://docs.oracle.com/en/operating-systems/oracle-linux/selinux/") }}}
```

You can use that instead.

applications/openshift/networking/default_ingress_ca_replaced/rule.yml
```diff
@@ -0,0 +1,47 @@
...
+
+severity: medium
+
+references:
+  nist: SC-17
```
JAORMX, line 26: Cany ou also add the NERC CIP mapping: `CIP-007-3 R5.1` ?

applications/openshift/networking/default_ingress_ca_replaced/tests/ocp4/e2e-remediation.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+# You can safely patch this to a non-existent secret name, it's a no-op, but OK for testing.
```
JAORMX, line 3: clever! I Was wondering what we could do to test this

JAORMX (changes requested):

JAORMX:
The rule has some issue:
```
     helpers.go:748: Result - Name: e2e-moderate-default-ingress-ca-replaced - Status: PASS - Severity: medium
    helpers.go:751: E2E-FAILURE: The expected result for the default_ingress_ca_replaced rule didn't match. Expected 'FAIL', Got 'PASS' 
```

applications/openshift/networking/default_ingress_ca_replaced/rule.yml
```diff
@@ -0,0 +1,47 @@
...
+warnings:
+- general: |-
+    {{{ openshift_cluster_setting("/apis/config.openshift.io/v1/proxies/cluster") | indent(4) }}}
+
+template:
+  name: yamlfile_value
+  vars:
+    ocp_data: "true"
+    filepath: /apis/config.openshift.io/v1/proxies/cluster
+    yamlpath: ".spec.trustedCA.name"
```
JAORMX, line 44: OpenSCAP's yaml probe has a caveat that for it to actually recognize that a rule's yamlpath doesn't match, the full path has to exist on the object. Now, `.spec.trustedCA.name` doesn't exist on the `cluster` object by default, which is why this reports PASS (confusingly enough).

In order to address this, we have the `check_existence_yamlpath` field which is documented here: https://complianceascode.readthedocs.io/en/latest/manual/developer/06_contributing_with_content.html#yamlfile-value

That field's purpose is to determine what's the minimum amount of yaml objects to gather... and so... if we compare it to what's being gathered they have to match. You can safely use `.metadata.name` which is always present in kubernetes objects. hence, in this case, `.metadata.name` would gather 1 object while `.spec.trustedCA.name` would gather 0 objects, and so it would fail which is what we want.

JAORMX (changes requested):

applications/openshift/networking/default_ingress_ca_replaced/rule.yml
mrogers950: Added, thanks!

JAORMX:
Well... this is weird:

```
error: build error: failed to pull image: After retrying 2 times, Pull image still failed due to error: Error initializing source docker://registry.fedoraproject.org/fedora-minimal:34: error pinging docker registry registry.fedoraproject.org: Get "https://registry.fedoraproject.org/v2/": dial tcp 38.145.60.20:443: i/o timeout
```

mrogers950:
oops, I have unsorted references
```
/home/runner/work/content/content/applications/openshift/networking/default_ingress_ca_replaced/rule.yml has one or more unsorted references

```

JAORMX:
huh: 
```
     helpers.go:748: Result - Name: e2e-moderate-default-ingress-ca-replaced - Status: PASS - Severity: medium
    helpers.go:751: E2E-FAILURE: The expected result for the default_ingress_ca_replaced rule didn't match. Expected 'FAIL', Got 'PASS' 
```

applications/openshift/networking/default_ingress_ca_replaced/rule.yml
```diff
@@ -0,0 +1,50 @@
...
+    check_existence_yamlpath: ".metadata.name"
+    check_existence: "all_exist"
+    values:
+    - value: ".*"
```
JAORMX, line 49: I see the issue now...
The default proxy called cluster has the following:
```
spec:
  trustedCA:
    name: ""
```

The `.*` matches everything, even an empty string.

You should instead use `.+` which matches at least one character.

JAORMX (changes requested):
found the issue

applications/openshift/networking/default_ingress_ca_replaced/rule.yml
mrogers950: Oh, I suspected that earlier, thanks, I fixed it.

PR #7451, opened 2021-08-24
Improve Ansible playbook linters test times

yuumasato:
#### Description:

- 100024a Fixes a bug where every product built tested all profile playbooks built.
- 18175d3 Reduces the scope per rule playbook Ansible and yaml linters to the `all` profile.
  - This drastically reduces the scope of playbooks tested and greatly reduces the test time.
  - The only differences, if any, between `all` and other profiles playbooks should be selected variables.

#### Rationale:

- The lint checkers take too long to run `ctest -R -lint-check --output-on-failure`
When all products are built, these are the test times  on my machine before this patch:
`Total Test time (real) = 24100.07 sec`
And after this patch:
`Total Test time (real) = 2150.24 sec`

### DIFF
--- a/cmake/SSGCommon.cmake
+++ b/cmake/SSGCommon.cmake
@@ -357,13 +357,13 @@ macro(ssg_build_remediations PRODUCT)
                 if (ANSIBLE_LINT_EXECUTABLE AND "${OSCAP_VERSION}" VERSION_GREATER "1.2.16")
                     add_test(
                         NAME "ansible-playbook-per-profile-ansible-lint-check-${PRODUCT}"
-                        COMMAND "${CMAKE_SOURCE_DIR}/tests/ansible_playbook_check.sh" "${ANSIBLE_LINT_EXECUTABLE}" "${CMAKE_BINARY_DIR}/ansible" "${CMAKE_SOURCE_DIR}/tests/ansible-lint_config.yml"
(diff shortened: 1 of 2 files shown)


yuumasato:
@matejak 
> . Do we really need it, or would it be possible to dissolve the wrapping functionality to those `cmake` invocations?

So.. we kinda need a script.
`add_test` makes every test command argument quoted and this prevents globs from expanding.
A test command defined like:
`"${ANSIBLE_LINT_EXECUTABLE}" -c "${CMAKE_SOURCE_DIR}/tests/ansible-lint_config.yml" "${CMAKE_BINARY_DIR}/ansible/${PRODUCT}*.yml`
Is run as:
`Test command: /usr/bin/ansible-lint "-c" "/home/wsato/git/content/tests/ansible-lint_config.yml" "/home/wsato/git/content/build/ansible/rhel8*.yml"`
This disables globs, the last argument is treated as an explicit path, and no files to test are found.

jan-cerny:
Can CMake pass expanded globs eg. as a list variable?

yuumasato:
> Can CMake pass expanded globs eg. as a list variable?

@jan-cerny Not sure what kind of glob you have in mind.
But from what I tested, CMake can do globs and pass them expanded. The problem is that the `file(GLOB ...)` command is run during configure time, and the playbooks for the test are only available after the build. (The correct solution would be to track/predict the playbooks that will be built and add them to the test as dependencies.)

jan-cerny:
Thanks for explanation. I think that it's not worth the effort. I'm OK with using the wrapper script.

jan-cerny (approved):
the tests run OK so it looks good to me

PR #7453, opened 2021-08-24
Removed unused bash function library include_merge_files_by_lines.sh

j-ode:
#### Description:

- Removed include_merge_files_by_lines.sh bash function library because it is not used anywhere.

#### Rationale:

- #7348


jan-cerny (approved):
Nice! I also can't find them anywhere.

PR #7454, opened 2021-08-24
Add sort prodtypes to fix_rules

Mab879:
#### Description:

This adds `sort_prodtypes` to `./utils/fix_rules`. This ensures that prodtypes are in alphabetical order.

#### Rationale:

To help make the rule files easier to read.

### DIFF
--- a/utils/fix_rules.py
+++ b/utils/fix_rules.py
@@ -272,6 +272,22 @@ def rewrite_value_int_str(line):
     return key + ": " + str_value
 
 
+def rewrite_keyless_section(file_contents, yaml_contents, section, content):
+    new_contents = file_contents[:]
+
+    sec_ranges = find_section_lines(file_contents, section)
+    if len(sec_ranges) != 1:
+        raise RuntimeError("Refusing to fix file: %s -- could not find one section: %d"
+                           % (path, sec_ranges))
+
+    if len(sec_ranges[0]) != 2:
+        raise RuntimeError("Section has more than one line")
+
+    new_contents[sec_ranges[0][0]] = f"{section}: {content}"
+
+    return new_contents
+
+
 def rewrite_value_remove_prefix(line):
     # Rewrites a key's value to remove a "CCE" prefix.
     key_end = line.index(':')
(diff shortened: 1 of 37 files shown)


utils/fix_rules.py
```diff
@@ -272,6 +272,25 @@ def rewrite_value_int_str(line):
     return key + ": " + str_value
 
 
+def rewrite_keyless_section(file_contents, yaml_contents, section, content):
+    # For a given section, rewrite the keys in int_keys to be strings. Refuses to
```
cipherboy, line 276: Is this comment correct?

utils/fix_rules.py
```diff
@@ -272,6 +272,25 @@ def rewrite_value_int_str(line):
...
+
+    sec_ranges = find_section_lines(file_contents, section)
+    if len(sec_ranges) != 1:
+        raise RuntimeError("Refusing to fix file: %s -- could not find one section: %d"
+                           % (path, sec_ranges))
+
+    if len(sec_ranges[0]) != 2:
+        raise RuntimeError("Section has more than one line")
+
+    new_contents[sec_ranges[0][0]] = f"{section}: {content}"
```
cipherboy, line 289: Are f-strings allowed? Are we expecting utilities to be py2/py3 compatible?

utils/fix_rules.py
```diff
@@ -562,6 +581,15 @@ def fix_callback(file_contents, yaml_contents):
             cce_pool.remove_cce_from_file(cce)
 
 
+def has_unsorted_prodtype(yaml_file, product_yaml=None):
+    rule = yaml.open_and_macro_expand(yaml_file, product_yaml)
+    if 'prodtype' in rule:
+        sorted_prodtypes = rule['prodtype'].split(',')
```
cipherboy, line 587: I’d maybe cache this as `split_prodtypes` Variable and then call `sorted(…)` instead of `.sort()`. My 2c.

utils/fix_rules.py
```diff
@@ -702,6 +730,31 @@ def sort_subkeys(args, product_yaml):
     return int(len(results) > 0)
 
 
+@command("sort_prodtypes", "sorts the products in prodtype")
```
cipherboy, line 733: Nit: “sorts products in the prodtype” or “sorts the products in the prodtype”. My 2c but I think the latter sounds best? As written, to me, it sounds like it is lacking a ‘the’.

cipherboy (commented):
Minor nits.

utils/fix_rules.py
matejak: I would allow it - even RHEL7 has Python3.6, moreover utilities are not needed for the build.

matejak (approved):
Great addition, works for me, many thanks!

PR #7456, opened 2021-08-25
SLE15 Add some audit rules to  sle15 PCI-DSS profile

teacup-on-rockingchair:
#### Description:

- Add some audit rules to  sle15 PCI-DSS profile

#### Rationale:
Add rules 
- audit_rules_networkconfig_modification
- audit_rules_session_events
- audit_rules_media_export
- audit_rules_mac_modification
- audit_rules_immutable
- audit-audispd-plugins_installed
- grub2_bootloader_argument

Made small improvement in the ansible template for non-rh based platforms for grub2_bootloader_argument rule to add kernelopts param if none is added


shared/templates/grub2_bootloader_argument/ansible.template
```diff
@@ -30,7 +30,8 @@
 {{% else %}}
 
 - name: get current kernel parameters
-  command: '/usr/bin/grub2-editenv - list'
+  ansible.builtin.shell:
+    cmd: '/usr/bin/grub2-editenv - list | echo "kernelopts="'
```
jan-cerny, line 34: But this will always print just `kernelopts=` and nothing else, won't it?

teacup-on-rockingchair: Thanks for bringing this up @jan-cerny :bow: 

Definitely very stupid mistake, thinking grep , writing echo :). Changed it in ea14ebd

jan-cerny (approved):

PR #7458, opened 2021-08-25
Remove package applicability from s390x_arch generated remediations

ggbecker:
#### Description:
- Remove package applicability from s390x_arch generated remediations.
  - This makes sure that there will be no package applicability check in the
remediations of rules that use the s390x_arch platform applicability,
since the check is made by checking a line in file instead. At this
moment the build system does not allow doing such checks. The side
effect is that Bash and Ansible roles will apply this remediation even
on a system that is not s390_arch, so using OpenSCAP scanner is highly
recommended.

### DIFF
--- a/ssg/constants.py
+++ b/ssg/constants.py
@@ -353,6 +353,7 @@ XCCDF_PLATFORM_TO_PACKAGE = {
   "uefi": None,
   "non-uefi": None,
   "not_s390x_arch": None,
+  "s390x_arch": None,
 }
 
 # _version_name_map = {


matejak:
So is it that either we have a RPM-based check that would be different from the OVAL, or we have no platform check in the remediation?
Anyway, the case when a remediation isn't triggered although it should have been is more problematic than the opposite case, so if we don't know of a package that is always installed on s390x systems, we may be better off with the no platform remediation check approach.

ggbecker:
> So is it that either we have a RPM-based check that would be different from the OVAL, or we have no platform check in the remediation?
> Anyway, the case when a remediation isn't triggered although it should have been is more problematic than the opposite case, so if we don't know of a package that is always installed on s390x systems, we may be better off with the no platform remediation check approach.

Maybe we can use the package that is used by the previously platform `zipl`:

https://github.com/ComplianceAsCode/content/blob/00397128c0fcd2d45521b34fb3ef421064e73dd7/shared/checks/oval/installed_env_has_zipl_package.xml#L24

@yuumasato what do you think? Is `s390utils-base` a good indication of a s390x machine?

yuumasato:
> @yuumasato what do you think? Is `s390utils-base` a good indication of a s390x machine?

Using package names for the remediation applicability check can lead to same issues we had with the CPE platform for `zipl`. Like package name changes and different names in different product versions.

yuumasato:
Closed by accident

yuumasato:
> Anyway, the case when a remediation isn't triggered although it should have been is more problematic than the opposite case, so if we don't know of a package that is always installed on s390x systems, we may be better off with the no platform remediation check approach.

From what I checked what will happen to non s390x systems is that non-existing binary may be run (`/usr/sbin/zipl`) and a new file created (`/etc/kernel/cmdline`).

These seem less problematic.

matejak:
I am not super happy about this, but it seems to be the best option that we have at the moment. I think that in the long run, we should be able to define applicability checks for OVAL and remediations next to each other, so we prevent desyncs like this.

matejak (approved):

yuumasato:
> I think that in the long run, we should be able to define applicability checks for OVAL and remediations next to each other, so we prevent desyncs like this.

Totally agree. There should be support for generic applicability checks too. Currently only machine and package checks are done.

PR #7471, opened 2021-08-26
Bug 1942281 - Postfix rules notapplicable when package is not installed

marcusburghardt:
Set postfix rules to notapplicable when package is not installed.

#### Description:

The rule "postfix_prevent_unrestricted_relay" is assessed even in systems where Postfix is not present, resulting in a "failed" result when actually should be "notapplicable".

This problem is also mentioned in this Bugzilla:
https://bugzilla.redhat.com/show_bug.cgi?id=1942281

#### Rationale:

It was included a CPE to check the postfix package and included this condition in the respective rule, via the "platform" parameter.


linux_os/guide/services/mail/postfix_harden_os/postfix_server_cfg/postfix_server_relay/postfix_prevent_unrestricted_relay/rule.yml
```diff
@@ -16,6 +16,8 @@ rationale: |-
 
 severity: medium
 
+platform: postfix
+
```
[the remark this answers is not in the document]
ggbecker: Then you can remove the `platform: postfix` from all rules that are within the group directory.

ggbecker:
The build is failing because `postfix_prevent_unrestricted_relay` rule does not have the `rhel9` in the `prodtype` attribute.

Also, when editing profiles that are in this folder: https://github.com/ComplianceAsCode/content/blob/master/tests/data/profile_stability/

means that you have to add the same rules in the profiles listed there. This is a test that prevents from updating profiles unintentionally. Think more like a safeguard for mistakenly introduced rules. When you build the project, you can just copy the processed profile file from `build/rhel8/profiles/stig.profile to https://github.com/ComplianceAsCode/content/blob/master/tests/data/profile_stability/rhel8/ and it should update accordingly.

marcusburghardt:
> The build is failing because `postfix_prevent_unrestricted_relay` rule does not have the `rhel9` in the `prodtype` attribute.
> 
> Also, when editing profiles that are in this folder: https://github.com/ComplianceAsCode/content/blob/master/tests/data/profile_stability/
> 
> means that you have to add the same rules in the profiles listed there. This is a test that prevents from updating profiles unintentionally. Think more like a safeguard for mistakenly introduced rules. When you build the project, you can just copy the processed profile file from `build/rhel8/profiles/stig.profile` to https://github.com/ComplianceAsCode/content/blob/master/tests/data/profile_stability/rhel8/ and it should update accordingly.

Interesting. Thanks for this information Gabriel. I am updating the PR.

linux_os/guide/services/mail/postfix_client/postfix_network_listening_disabled/rule.yml
```diff
@@ -42,3 +42,5 @@ ocil: |-
     Run the following command to ensure postfix accepts mail messages from only the local system:
     <pre>$ grep inet_interfaces /etc/postfix/main.cf</pre>
     If properly configured, the output should show only <tt>{{{ xccdf_value("var_postfix_inet_interfaces") }}}</tt>.
+
+platform: postfix
```
yuumasato, line 46: @marcusburghardt You can remove the platform from the rule.
It will inherit it from the Group it is in.

yuumasato (changes requested):
LGTM, just one change and should be good to merge.

linux_os/guide/services/mail/postfix_client/postfix_network_listening_disabled/rule.yml
marcusburghardt: @yuumasato , this rule is in the group `postfix_client` while the modified group is `postfix_harden_os`, in the same level. I can't include the platform parameter in the group `postfix_client`, without change the directories structure, because other rules there actually are not dependent of postfix package, differently of `postfix_harden_os` group, where all rules depend of postfix package.

yuumasato: @marcusburghardt It seems to me that all rules in `postfix_client` group only make sense when postfix is installed.
I think `postfix_client_configure_mail_alias` is the least obvious, but the rule and remediation mention `postconf` and `newaliases`, which are `postfix` binaries.

yuumasato: Discussed offline with @marcusburghardt , and Rule `postfix_client_configure_mail_alias` is more related to root mail alias then `postfix` itself.

yuumasato (approved):

PR #7491, opened 2021-08-31
Convert perform_audit_rules_privileged_commands_remediation bash function to a Jinja macro

j-ode:
#### Description:

- Converted the perform_audit_rules_privileged_commands_remediation bash remediation function to a jinja macro and replaced all function calls in other files with a macro invocation, as well as removed unnecessary sourcing of the /usr/share/scap-security-guide/remediation_functions folder.

#### Rationale:

- #7348


shared/macros-bash.jinja
```diff
@@ -39,6 +39,151 @@ populate {{{ name }}}
...
+output_audit_file="/etc/audit/audit.rules"
+{{%- else -%}}
+# If the audit tool is 'augenrules', then:
+# * add '/etc/audit/rules.d/*.rules' to the list of files to be inspected
+#   (split by newline),
+# * specify /etc/audit/rules.d/privileged.rules' as the output file, where
+#   missing rules should be inserted
+readarray -t files_to_inspect < <(find /etc/audit/rules.d -maxdepth 1 -type f -name '*.rules' -print)
+output_audit_file="/etc/audit/rules.d/privileged.rules"
+{{%- endif %}}
```
jan-cerny, line 74: I have a new idea that we could check that it's one of "auditctl" and "augenrules" and produce an error when the macro is used with some different value as a parameter.

Something like:

```
{{% if tool == "auditctl" %}}
...
{{% elif tool == "augenrules" %}}
...
{{%- else -%}}
{{{ raise("Unknown tool used: " + tool) }}}
{{%- endif %}}

```

The `raise()` is already used in other macros in this file so you can get inspired.

shared/macros-bash.jinja
```diff
@@ -39,6 +39,151 @@ populate {{{ name }}}
...
+
+# Keep list of SUID/SGID binaries that have been already handled within some previous iteration
+sbinaries_to_skip=()
+
+# For each found sbinary in privileged_binaries list
+for sbinary in "${privileged_binaries[@]}"
+do
+
+	# Check if this sbinary wasn't already handled in some of the previous sbinary iterations
+	# Return match only if whole sbinary definition matched (not in the case just prefix matched!!!)
```
jan-cerny, line 89: Use spaces for indentation.

shared/macros-bash.jinja
```diff
@@ -39,6 +39,151 @@ populate {{{ name }}}
...
+
+	# For each audit rules file from the list of files to be inspected
+	for afile in "${files_to_inspect[@]}"
+	do
+		# Search current audit rules file's content for match. Match criteria:
+		# * existing rule is for the same SUID/SGID binary we are currently processing (but
+		#   can contain multiple -F path= elements covering multiple SUID/SGID binaries)
+		# * existing rule contains all arguments from expected rule form (though can contain
+		#   them in arbitrary order)
+
```
jan-cerny, line 120: trailing whitespace

shared/macros-bash.jinja
```diff
@@ -285,11 +432,11 @@ fi
 
 {{%- macro _make_bash_variable_assignment(varname, value="", quoted_value="") -%}}
 {{% if value -%}}
-	{{{ varname }}}="{{{ value }}}"
+    {{{ varname }}}="{{{ value }}}"
```
jan-cerny, line 435: Please don't change unrelated macros.

jan-cerny (changes requested):

shared/macros-bash.jinja
j-ode: Reverted changes in commit 1a81452

jan-cerny (approved):

PR #7493, opened 2021-08-31
Allow HID USB in the ISM profile

matejak:
The usbguard is too strict without this rule, and its default setting blocks keyboard and mouse.

### DIFF
--- a/products/rhel8/profiles/ism_o.profile
+++ b/products/rhel8/profiles/ism_o.profile
@@ -52,6 +52,7 @@ selections:
   ## Identifiers 1418
   - package_usbguard_installed
   - service_usbguard_enabled
+  - usbguard_allow_hid_and_hub
 
   ## Authentication hardening
   ## Identifiers 1546 / 0974 / 1173 / 1504 / 1505 / 1401 / 1559 / 1560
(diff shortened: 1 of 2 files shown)


yuumasato (approved):
LGTM

matusmarhefka:
This might do the trick for the keyboard/mouse devices, but other devices connected to the machine (like card readers) might still be blocked by the usbguard. What about also including `usbguard_generate_policy` rule? But I am not sure how would this rule interact with the `usbguard_allow_hid_and_hub`.

yuumasato:
> This might do the trick for the keyboard/mouse devices, but other devices connected to the machine (like card readers) might still be blocked by the usbguard.

From what I checked, the ISM profile doesn't configure card readers.

>  But I am not sure how would this rule interact with the `usbguard_allow_hid_and_hub`.

[Remediation](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/usbguard/usbguard_generate_policy/bash/shared.sh#L12) of `usbguard_generate_policy` overwrites `/etc/usbguard/rules.conf`. So It doesn't seem to play well with `usbguard_allow_hid_and_hub`'s [remediation](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/usbguard/usbguard_allow_hid_and_hub/bash/shared.sh#L5).

matusmarhefka:
I can confirm that  keyboard/mouse is not blocked anymore with the rule `usbguard_allow_hid_and_hub` selected in the profile.

PR #7495, opened 2021-08-31
OL8 was prevent build all from working

Mab879:
This issue was created in b773be3a0e8042f62fe136a55ef1bdea5dda5287. This prevent the nightly builds from working. This quick fix to make the build work.


jan-cerny (approved):

yuumasato:
This may fix the build but I think it breaks the check logic for `ol8`.
CC: @freddieRv

freddieRv:
I will take a look into it. Thanks for the heads up @yuumasato

PR #7497, opened 2021-09-01
Add docs on status proposal for controls structure

JAORMX:
This adds docs about the `status` and `tickets` keys that are meant to
be introduced to the "controls" structure. These will help portray the
current status of the implementation of a control. And help people track
where work still needs to be done to complete it.

Note that when writing automation that actually takes this into use, I'll default to
using `status: pending`

### DIFF
--- a/docs/manual/developer/03_creating_content.md
+++ b/docs/manual/developer/03_creating_content.md
@@ -558,6 +558,65 @@ controls:
       and to use the targeted policy.
 ```
 
+### Reporting status
+
+In some cases, it's useful to know the status of a certain control for a
+specific product. In order to better portray this, it's possible to set
+such information on each control using the `status` key.
+
+The `status` key may hold the following values:
+
+* `pending`: The control is not yet evaluated for the product.
+
+* `not applicable`: The control is not applicable to this product.
+
+* `inherently met`: The control is inherently met by the product.
+
+* `documentation`: The control is addressed by product documentation.
+
+* `planned`: The control is not yet implemented, but is planned.
+
+* `partial`: While work has been done to address this control, there is still
+             work needed to fully address it.
+
+* `supported`: The control is addressed by the product (but is missing content
+                automation).
+
+* `automated`: The control is addressed by the product and can be automatically
+               checked for.
+
+When there is work on-going to address a specific control, it may be portrayed
+via the `tickets` key. The aforementioned key shall contain a list of URLs that
+may help the reader track what work needs to be done to address a specific
+control.
+
+```
+$ cat controls/abcd.yml
+
+id: abcd
+title: ABCD Benchmark for securing Linux systems
+version: 1.2.3
+source: https://www.abcd.com/linux.pdf
+levels:
+  - id: low
+  - id: high
+    inherits_from:
+    - low
+controls:
+  - id: R1
+    levels:
+    - low
+    title: User session timeout
+    description: |-
+      Remote user sessions must be closed after a certain
+      period of inactivity.
+    status: partial
+    tickets:
+    - https://my-ticket-tracker.com/issue/1
+    - https://my-ticket-tracker.com/issue/2
+```
+
+
 ### Controls file format
 
 This is a complete schema of the YAML file format.
@@ -581,6 +640,8 @@ controls: a list of controls (required key)
     related_rules: a list of related rules
     note: a short paragraph of text
     controls: a nested list of controls
+    status: a keyword that reflects the current status of the implementation of this control
+    tickets: a list of URLs reflecting the work that still needs to be done to address this control
 ```
 
 Full example of a controls file:


docs/manual/developer/03_creating_content.md
```diff
@@ -558,6 +558,61 @@ controls:
...
+* `inherently met`: The control is inherently met by the product.
+
+* `documentation`: The control is addressed by product documentation.
+
+* `planned`: The control is not yet implemented, but is planned.
+
+* `partial`: While work has been done to address this control, there is still
+             work needed to fully address it.
+
+* `complete`: The control is addressed by the product.
```
yuumasato, line 582: And `status: complete` seems equivalent to  `automated: yes`

docs/manual/developer/03_creating_content.md
```diff
@@ -558,6 +558,61 @@ controls:
...
+
+* `inherently met`: The control is inherently met by the product.
+
+* `documentation`: The control is addressed by product documentation.
+
+* `planned`: The control is not yet implemented, but is planned.
+
+* `partial`: While work has been done to address this control, there is still
+             work needed to fully address it.
+
```
yuumasato, line 581: `status: partial` seems equivalent to `automated: partial`

docs/manual/developer/03_creating_content.md
```diff
@@ -558,6 +558,61 @@ controls:
...
+The `status` key may hold the following values:
+
+* `pending`: The control is not yet evaluated for the product.
+
+* `not applicable`: The control is not applicable to this product.
+
+* `inherently met`: The control is inherently met by the product.
+
+* `documentation`: The control is addressed by product documentation.
```
yuumasato, line 575: Out of curiosity, could you point to a control addressed by documentation?

docs/manual/developer/03_creating_content.md
```diff
@@ -558,6 +558,61 @@ controls:
...
+specific product. In order to better portray this, it's possible to set
+such information on each control using the `status` key.
+
+The `status` key may hold the following values:
+
+* `pending`: The control is not yet evaluated for the product.
+
+* `not applicable`: The control is not applicable to this product.
+
+* `inherently met`: The control is inherently met by the product.
```
yuumasato, line 573: This is where the `automated` key used to fall short, an explanation in notes is necessary to explain why it is not automated. The`status` key seems to handle this better.

Question about `inherently met`, does this mean that a product satisfies a control by design and no configuration of any sort is needed? That compliance would no be broken by misconfigurations.

docs/manual/developer/03_creating_content.md
```diff
@@ -558,6 +558,61 @@ controls:
...
+
+* `partial`: While work has been done to address this control, there is still
+             work needed to fully address it.
+
+* `complete`: The control is addressed by the product.
+
+When there is work on-going to address a specific control, it may be portrayed
+via the `tickets` key. The aforementioned key shall contain a list of URLs that
+may help the reader track what work needs to be done to address a specific
+control.
```
yuumasato, line 587: To understand a bit more the nature of the tickets. Would they be about features that need to be implemented in the product to be able to satisfy a control, or about the actual writing of rules, or any kind of support in CaC? Or both?

yuumasato (commented):
It seems to me that `status` can completely supersede `automated` key while extending its capabilities. The `automated` key could not reflect the reasons why a controls was not automated ("Was it not evaluated yet?", "The product cannot satisfy it", "The product always satisfyes it").

docs/manual/developer/03_creating_content.md
JAORMX: @yuumasato pretty much. Or that: "misconfiguring this is non-trivial".

docs/manual/developer/03_creating_content.md
JAORMX: I was thinking more along the lines of "features that need to be implemented in the product to be able to satisfy a control"

docs/manual/developer/03_creating_content.md
JAORMX: uhm... now that I think about it, we might want to distinguish between "the product can do this" and "the product can do this & we can already check this through content"

docs/manual/developer/03_creating_content.md
JAORMX: https://github.com/ComplianceAsCode/redhat/blob/master/openshift-container-platform-4/policies/SA-System_and_Services_Acquisition/component.yaml#L146-L170

docs/manual/developer/03_creating_content.md
JAORMX: It is already quite trivial for us to determine if a control is automated since... well, this is in the same repo as the automation for the control. So all we need to do is build the content, check the rules pertaining to the control it have OVAL and that's it.

docs/manual/developer/03_creating_content.md
JAORMX: perhaps we should have something like:
`status: implemented` to portray that the system implements it... and then...
`status: automated` to portray that we have automation in place to check it's compliance in the content.

docs/manual/developer/03_creating_content.md
yuumasato: I see, I overlooked the narrative as  the controls are marked with `implementation_status: complete`.
So from what I could check, these are used for controls that are about documentation.

Do you imagine `status: documentation` being used for controls that will not have automation and are supposed to be remediated manually? And the control points to documentation.
For example, configuring GRUB or BIOS password.

docs/manual/developer/03_creating_content.md
yuumasato: That makes sense to me, although `implemented` could be confused with "we have rules implemented".
How about `supported`, as in the product has features that can support the control.

JAORMX: I dig that!

JAORMX:
Pull-request updated, HEAD is now ac3213ce6c879daaa3cd8ac203b0890060364c97

yuumasato (approved):
LGTM.
Should we also document that `status` is better suited for reporting than`automated`?

JAORMX:
> LGTM.
> Should we also document that `status` is better suited for reporting than`automated`?

I think we can do that in a subsequent PR. If we do plan to replace `automated`, then we can start by taking the `status: automated` into account in https://github.com/ComplianceAsCode/content/blob/master/ssg/controls.py

PR #7500, opened 2021-09-02
Add audit_rules_dac_modification_umount specific

### LINKED ISSUE #7386: SLES15 STIG umount syscall on 64bit
#### Description of problem:

The check for audit record of umount syscall is failing (rule SLES-15-030360).
From what I understand, it's looking for the umount syscall on 32 bits and 64 bits.
Except that the STIG manual is asking to check sys_umount only on 32bits and
for a good reason: it doesn't exist on 64bits.
See https://marcin.juszkiewicz.com.pl/download/tables/syscalls.html for instance.

SLES 15 STIG V1R3 Manual:
```
Check that the system calls are being audited by performing the following command:

> sudo auditctl -l | grep 'umount'

-a always,exit -F arch=b32 -S umount -F auid>=1000 -F auid!=-1 -k privileged-umount
-a always,exit -F arch=b32 -S umount2 -F auid>=1000 -F auid!=-1 -k privileged-umount
-a always,exit -F arch=b64 -S umount2 -F auid>=1000 -F auid!=-1 -k privileged-umoun
```

0.1.57

SLES15SP2

#### Steps to Reproduce:

1. Download and extract latest release on a sles15 with SLES15 STIG rules applied. In particular
```
# grep umount /etc/audit/rules.d/01-stig.rules
-a always,exit -F arch=b32 -S umount -F auid>=1000 -F auid!=4294967295 -k privileged-umount
-a always,exit -F arch=b32 -S umount2 -F auid>=1000 -F auid!=4294967295 -k privileged-umount
-a always,exit -F arch=b64 -S umount2 -F auid>=1000 -F auid!=4294967295 -k privileged-umount
```
2. Run the bench, for instance with:
```
# oscap xccdf eval \
    --fetch-remote-resources \
    --profile xccdf_org.ssgproject.content_profile_stig \
    --report /tmp/report.html ssg-sle15-ds.xml
```

#### Actual Results:

Look at the results and notice that the check for SLES-15-030360 is failing

#### Expected Results:

Check to succeed

teacup-on-rockingchair:
#### Description:

- Modify audit_rules_dac_modification_umount rule definition to use a special case on the audit_rules_dac_modification template to check/apply only 32bit relevant rules

#### Rationale:

- Fixes #7386

### DIFF
--- a/linux_os/guide/system/auditing/auditd_configure_rules/audit_dac_actions/audit_rules_dac_modification_umount/ansible/shared.yml
+++ b/linux_os/guide/system/auditing/auditd_configure_rules/audit_dac_actions/audit_rules_dac_modification_umount/ansible/shared.yml
@@ -1,0 +1,43 @@
+# platform = multi_platform_all
+# reboot = true
+# strategy = restrict
+# complexity = low
+# disruption = low
+
+
+- name: Perform remediation of Audit rules for umount for x86 platform
+  block:
+    {{{ ansible_audit_augenrules_add_syscall_rule(
+      action_arch_filters="-a always,exit -F arch=b32",
+      other_filters="",
+      auid_filters="-F auid>="~auid~" -F auid!=unset",
+      syscalls=["umount"],
+      key="perm_mod",
+      syscall_grouping=[],
+      )|indent(4) }}}
+    {{{ ansible_audit_auditctl_add_syscall_rule(
+      action_arch_filters="-a always,exit -F arch=b32",
+      other_filters="",
+      auid_filters="-F auid>="~auid~" -F auid!=unset",
+      syscalls=["umount"],
+      key="perm_mod",
+      syscall_grouping=[]
+      )|indent(4) }}}
+{{%- if CHECK_ROOT_USER %}}
+    {{{ ansible_audit_augenrules_add_syscall_rule(
+      action_arch_filters="-a always,exit -F arch=b32",
+      other_filters="",
+      auid_filters="-F auid=0",
+      syscalls=["umount"],
+      key="perm_mod",
+      syscall_grouping=[],
+      )|indent(4) }}}
+    {{{ ansible_audit_auditctl_add_syscall_rule(
+      action_arch_filters="-a always,exit -F arch=b32",
+      other_filters="",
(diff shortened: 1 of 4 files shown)


yuumasato (approved):
LGTM.
The template could also be modified to handle 32 and 64 bit cases only, but this fix is good for now.

teacup-on-rockingchair:
> LGTM.
> The template could also be modified to handle 32 and 64 bit cases only, but this fix is good for now.

I also was thinking about that option :
- having all the templates loop through list of supported architectures and create rules for each one of them
-  adding additional template var for supported architectures for a syscall.
-  If none specified, then assume all supported by the template are valid.

But then, considered that only 32bit syscalls will become obsolete with time and wasn't sure if the complication in the template code will have that much of a benefit in the general case. Also considered the recommendation of @ggbecker which sounded more like this is a special case that doesn't need to add complexity to the templates.

Still if you think it might look better with general template support might try to implement it in the context of the same PR

yuumasato:
> > LGTM.
> > The template could also be modified to handle 32 and 64 bit cases only, but this fix is good for now.
> 
> I also was thinking about that option :
> 
>     * having all the templates loop through list of supported architectures and create rules for each one of them
> 
>     * adding additional template var for supported architectures for a syscall.
> 
>     * If none specified, then assume all supported by the template are valid.
> 
> 
> But then, considered that only 32bit syscalls will become obsolete with time and wasn't sure if the complication in the template code will have that much of a benefit in the general case. Also considered the recommendation of @ggbecker which sounded more like this is a special case that doesn't need to add complexity to the templates.

The point about only b32 syscalls going obsolete is a good one. 
> 
> Still if you think it might look better with general template support might try to implement it in the context of the same PR

I don't have a strong opinion here, I agree taht additional complication may not be worth it.
There aren't many rules for syscalls that are b32 only. The other rule witch is b32 only is [audit_rules_time_stime](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/auditing/auditd_configure_rules/audit_time_rules/audit_rules_time_stime/rule.yml), and this rule has a custom check.

We can go with the simpler custom approach, and if more rules for b32 only syscall come up, we can consider a more general solution.

PR #7505, opened 2021-09-03
Fix the offline remediation of firewalld

matejak:
When the service is not running, we use the `firewall-offline-cmd`, and we also don't reload the config.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1812180

### DIFF
--- a/linux_os/guide/system/network/network-firewalld/ruleset_modifications/configure_firewalld_rate_limiting/bash/shared.sh
+++ b/linux_os/guide/system/network/network-firewalld/ruleset_modifications/configure_firewalld_rate_limiting/bash/shared.sh
@@ -1,4 +1,9 @@
-# platform = Red Hat Enterprise Linux 7,multi_platform_wrlinux,multi_platform_ol
+# platform = multi_platform_all
 
-firewall-cmd --permanent --direct --add-rule ipv4 filter INPUT_direct 0 -p tcp -m limit --limit 25/minute --limit-burst 100  -j INPUT_ZONES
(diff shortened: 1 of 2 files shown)


jan-cerny:
@matejak How can I test this? Now the rule is only present in the OL7 and WRLinux STIG profiles. Which profile did you use? Were you backporting this to RHEL 7.8 or did you use current upstream content for `rhel7` product with some modifications or even something different?

matejak:
> @matejak How can I test this? Now the rule is only present in the OL7 and WRLinux STIG profiles. Which profile did you use? Were you backporting this to RHEL 7.8 or did you use current upstream content for `rhel7` product with some modifications or even something different?

You are right, that rule is not selected in our profiles. I have added it to one for the testing purpose, then built the datastream and supplied it to an Anaconda installation, and I have finally examined the results file.

jan-cerny:
I confirm that in the results file the result of this rule is now "fixed".

jan-cerny:
@matejak Would it be possible to have a test scenario for this situation?

matejak:
> @matejak Would it be possible to have a test scenario for this situation?

Probably not upstream - only a functional test would discover that the old remediation was not good enough, but setting up a testing environment that would be able to test this situation and recover results would be difficult to set up. I suggest that we do the "best effort" approach in remediations that we can't test easily, which means that we fix bugs and hope that things won't break by themselves.

jan-cerny:
OK, I understand.

jan-cerny (approved):

PR #7508, opened 2021-09-06
OCP4: Mark all PCI-DSS section 9 as `not applicable`

JAORMX:
These are physical controls and not something that the platform itself
can do, nor something we can check for.


jhrozek (approved):
/lgtm

PR #7509, opened 2021-09-06
Fix test scenarios which should remove libselinux package

matusmarhefka:
Templated test scenarios cannot be used in this case.
Package libselinux cannot be uninstalled normally
as it would cause removal of `sudo` package which is
protected and package manager returns error in such
case. If the package would be removed forcefully it
would make the system unusable. Therefore, we will
remove the existing rpmdb and we will create a fake
empty one without libselinux package.


jan-cerny (approved):
@matusmarhefka it's amazing

PR #7513, opened 2021-09-07
OCP4: Update OpenShift CIS Benchmark version

JAORMX:
The official benchmark has been released and so we can now refer to the
version we're using.


jhrozek (approved):
LGTM

PR #7514, opened 2021-09-07
New test cases for audit rules

marcusburghardt:
#### Description:

Included missing testing scenarios for some rules related to audit.
Where it was possible, it were created templates to save code duplication.

#### Rationale:

One new test case introduced in this PR revealed an issue in the remediation rules, which was also fixed here.
Also, with the introduction of new test scenarios as "templates" we can have more examples to support a gradual reduction of duplicating code, at least related to test scenarios.

### DIFF
--- a/linux_os/guide/system/auditing/auditd_configure_rules/audit_login_events/audit_rules_login_events_faillog/tests/correct_rules.pass.sh
+++ b/linux_os/guide/system/auditing/auditd_configure_rules/audit_login_events/audit_rules_login_events_faillog/tests/correct_rules.pass.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+#
+
+rm -f /etc/audit/rules.d/*
+> /etc/audit/audit.rules
+echo "-w /var/log/faillog -p wa -k logins" >> /etc/audit/rules.d/logins.rules
--- a/linux_os/guide/system/auditing/auditd_configure_rules/audit_dac_actions/audit_rules_dac_modification_fchown/tests/correct_rules.pass.sh
+++ b/linux_os/guide/system/auditing/auditd_configure_rules/audit_dac_actions/audit_rules_dac_modification_fchown/tests/correct_rules.pass.sh
@@ -1,6 +1,0 @@
-#!/bin/bash
-
-
-echo "-a always,exit -F arch=b32 -S fchown -F auid>=1000 -F auid!=unset -F key=perm_mod" >> /etc/audit/rules.d/perm_mod.rules
-echo "-a always,exit -F arch=b64 -S fchown -F auid>=1000 -F auid!=unset -F key=perm_mod" >> /etc/audit/rules.d/perm_mod.rules
-
--- a/linux_os/guide/system/auditing/auditd_configure_rules/audit_file_deletion_events/audit_rules_file_deletion_events_rename/tests/correct_rules.pass.sh
+++ b/linux_os/guide/system/auditing/auditd_configure_rules/audit_file_deletion_events/audit_rules_file_deletion_events_rename/tests/correct_rules.pass.sh
@@ -1,5 +1,0 @@
-#!/bin/bash
-
-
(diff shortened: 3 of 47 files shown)


linux_os/guide/system/auditing/auditd_configure_rules/audit_login_events/audit_rules_login_events_faillog/tests/correct_rules.pass.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+#
+
```
jan-cerny, line 3: The rule audit_rules_login_events_faillog is only on sle12,ubuntu2004. Do you test it on any of these 2 systems?

jan-cerny (commented):
otherwise all the scenarios pass for me on a RHEL 8 VM backend

shared/templates/audit_rules_usergroup_modification/oval.template
```diff
@@ -18,7 +18,7 @@
   </ind:textfilecontent54_test>
   <ind:textfilecontent54_object id="object_audit_rules_usergroup_modification_{{{ NAME }}}_augen" version="1">
     <ind:filepath operation="pattern match">^/etc/audit/rules\.d/.*\.rules$</ind:filepath>
-    <ind:pattern operation="pattern match">^\-w[\s]+{{{ PATH }}}[\s]+\-p[\s]+\b([rx]*w[rx]*a[rx]*|[rx]*a[rx]*w[rx]*)\b[\s]+(-k[\s]+|-F[\s]+key=)\w+[\s]*$</ind:pattern>
+    <ind:pattern operation="pattern match">^\-w\s+{{{ PATH }}}\s+\-p\s+(wa|aw)\s+(-k[\s]+|-F[\s]+key=)[-\w]+\s*$</ind:pattern>
```
matusmarhefka, line 21: What is the reason of this regex update? I think the original intention was to accept more variations of permissions, e.g. when someone wants to have also read permission alongside the permissions required by the `audit_rules_usergroup_modification` rules it was accepted by the original OVAL check, now it would not be and the OVAL check would be asserted as fail in such cases.

marcusburghardt: Actually my intention was to make sure the both permissions related to write were assessed. Before the update, if the rule was only with a or w, it would be wrongly assessed as PASS. But, reading your comment now I realized that combinations of wa with other permissions were excluded with this update. I will fix it to solve both cases. Thanks

linux_os/guide/system/auditing/auditd_configure_rules/audit_login_events/audit_rules_login_events_faillog/tests/correct_rules.pass.sh
marcusburghardt: Hi Jan, I wasn't able to test in these 2 systems but I cant't see a reason to fail there, since the test is only exclusive. I mean that similar tests already passed there. In this one specifically we are only excluding other systems due that specific file (faillog) be typical for these two systems. I will try to prepare a test environment for these systems soon to confirm. But I don't think it is a blockage, considering the test case.

marcusburghardt:
The "SSGTS / Run Tests (pull_request)" test is failing because the container used for this test doesn't have `audit` installed, but it is expected by this test. It was relatively difficult to reproduce this and finally collect the relevant information for troubleshooting.

To avoid this in the future, I sent another PR improving the logging of remote commands on ssg_test_suites:
https://github.com/ComplianceAsCode/content/pull/7527

Regarding this failure, it something related to the test environment itself, since all tests were successful in a VM environment.
So, there is no blockage for this PR for now.

I will analyse the following ideas for other PRs:
- Review or maybe introduce Applicability of content for audit related rules.
- Review and maybe update the container files for the ssg_test_suite.

matusmarhefka (approved):

PR #7519, opened 2021-09-08
Fix test scenarios which should remove sudo package

matusmarhefka:
Templated test scenarios cannot be used in this case.
`sudo` is a special package which cannot be normally
removed so we need to remove it using rpm and without
removing other dependencies.

### DIFF
--- a/linux_os/guide/system/software/sudo/package_sudo_installed/tests/package-installed-removed.fail.sh
+++ b/linux_os/guide/system/software/sudo/package_sudo_installed/tests/package-installed-removed.fail.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+# platform = multi_platform_rhel,multi_platform_fedora
+
+# sudo is a special package which cannot be normally
+# removed so we need to remove it using rpm and without
+# removing other dependencies.
+rpm -e --nodeps sudo
--- a/linux_os/guide/system/software/sudo/package_sudo_installed/tests/package-removed.fail.sh
+++ b/linux_os/guide/system/software/sudo/package_sudo_installed/tests/package-removed.fail.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+# platform = multi_platform_rhel,multi_platform_fedora
+
+# sudo is a special package which cannot be normally
+# removed so we need to remove it using rpm and without
+# removing other dependencies.
+rpm -e --nodeps sudo


cipherboy:
@matusmarhefka Just a thought, but do we want to default to always removing without deps (in all test scenarios)? When we use the template to check for a package, we only check for _that_ package and not any or all deps of it (e.g., making up something but we only check for sudo and not sudo-common or sudo-base) to be present/removed.

matusmarhefka:
> @matusmarhefka Just a thought, but do we want to default to always removing without deps (in all test scenarios)? When we use the template to check for a package, we only check for _that_ package and not any or all deps of it (e.g., making up something but we only check for sudo and not sudo-common or sudo-base) to be present/removed.

Yeah, I think it would make sense. So if nobody is against let's do that, I can update it within this PR.

On Fedora/RHEL and SUSE we can use `rpm -e --nodeps` right?
What about Ubuntu/Debian? I found that it should be possible using `dpkg -r --force-depends PACKAGE`. Is that okay to use or do you have any other suggestion @cipherboy ?

cipherboy:
That sounds fine with me from my reading of the man page, but best to check with @dodys or @richardmaciel-canonical if they have any other suggestions. :)

richardmaciel-canonical:
`dpkg -r --force-depends PACKAGE` is good enough.

matusmarhefka:
I was thinking about this more and I am not sure if we really want to do package removals without dependencies. My concern is that when we just remove one specific package we might leave unneeded dependencies installed on the system which would be otherwise uninstalled. I would like to see more opinions on this change before going forward with it.

matusmarhefka:
Ok so I would say let's not make the decision within this PR as it is something which would change behavior in many rules just to workaround few exceptions. Therefore, I am leaving the PR as it is.

richardmaciel-canonical:
> I was thinking about this more and I am not sure if we really want to do package removals without dependencies. My concern is that when we just remove one specific package we might leave unneeded dependencies installed on the system which would be otherwise uninstalled. I would like to see more opinions on this change before going forward with it.

I agree with your decision. Executing this automatically could leave the affected system in an unstable state. Depending on the amount of rules executed together, would be a pain to debug.

cipherboy:
@richardmaciel-canonical said:
> 
> @matusmarhefka said:
> > 
> > I was thinking about this more and I am not sure if we really want to do package removals without dependencies. My concern is that when we just remove one specific package we might leave unneeded dependencies installed on the system which would be otherwise uninstalled. I would like to see more opinions on this change before going forward with it.
> 
> I agree with your decision. Executing this automatically could leave the affected system in an unstable state. Depending on the amount of rules executed together, would be a pain to debug.

Note that in both rule and combined execution, snapshots are taken before and after each test case and thus numbers of rules or tests doesn't impact system stability here. It makes debugging the failure of a test case potentially harder but it doesn't impact the stability across multiple rules.

marcusburghardt:
I think this approach of removing packages without consider dependencies as a standard may create more problems (or at least more work) than solutions due the possibility of something be broken and impact in the test itself. Maybe this happen eventually only, considering the packages we want to test, but even if happens rarely we would need to analyse case by case always. 

I think that our tests should be more realistic as possible to make sense. In a situation like that, I would opine to create exceptions for these special packages inside the test template itself. I saw two scenarios this week and I don't think we have much more different cases for package related rules:

1 - Protected package, like sudo: 
This package is protected but is not required by any other critical package. So, we could safely remove it slightly different via dnf using this command: dnf remove --setopt protected_packages= sudo or even rpm -e --nodeps sudo on RHEL and Fedora, at least.

2 - Package required by many critical packages, like libselinux:
If we remove this package in a RHEL or Fedora system (mostly in other distros too, but I couldn't test), the system won't even work in a productive way. So, looks reasonable to me only skip the test which tries to remove it or trick the OS by cleaning the rpmdb and equivalent in other distros. This doesn't look realistic.

So, regardless these examples and how we would "do" the test for these exceptional cases, I think it is more scalable to introduce conditional on templates than treat them in individual rules. Also, this `sudo` case looks to be more an exception instead of a rule. Looks strange to me change the entire standard due to an exception. So, to not be misinterpreted, : ) I am not against the technical approach in this PR related to this specific package, but I don't see it as a definitive solution or a candidate for a long-term standard removing packages without consider dependencies neither to work directly on rules.

matusmarhefka:
Closing this in favor of https://github.com/ComplianceAsCode/content/pull/7559

PR #7524, opened 2021-09-09
OCP4: Mark relevant rules to use the platform `ocp4-node`

JAORMX:
Some rules are only meant to run on the node itself. So let's mark them
as such as opposed to only relying on that distinction to happen on the
profile itself.

### DIFF
--- a/applications/openshift/etcd/etcd_unique_ca/rule.yml
+++ b/applications/openshift/etcd/etcd_unique_ca/rule.yml
@@ -26,6 +26,7 @@ references:
 
 platforms:
   - ocp4-master-node
+  - ocp4-node
 
 warnings:
   - dependency: |-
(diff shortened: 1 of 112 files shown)


JAORMX:
Pull-request updated, HEAD is now 085e6ba8f394f8462a15ae017511dc8cb4ca7394

JAORMX:
Well... this didn't work :(

matejak:
So wouldn't it be enough to simply not add the `ocp4-node` platform if there is already an `ocp4-master-node` one there? As an `ocp4-master-node` is a node, it is basically implied. It is true that checks for node and master node check different things, we don't have to reflect this in any way. Having those non-exhaustive platforms declared in rule files is IMO still beneficial, because they contribute to the rule's definition.
If the project supports declaration of platforms hierarchy, i.e. saying that `ocp4-master-node` is a special case of `ocp4-node`, that would even be enough for filtering.

PR #7507, opened 2021-09-03
Move the templated sshd OVAL check to a macro

matejak:
That way, it can be reused e.g. in tests.

PR #7525, opened 2021-09-09
Add templated test scenarios for sshd rules

matusmarhefka:
There are following templated scenarios added:
* `correct_value.pass.sh`
* `correct_value_directory.pass.sh`
* `param_conflict.fail.sh`
* `param_conflict_directory.fail.sh`
* `wrong_value.fail.sh`
* `wrong_value_directory.fail.sh`

The following tests are not templated as they might be pass or fail
test scenarios depending on a specific rule:
* `comment.{pass,fail}.sh`
* `line_not_there.{pass,fail}.sh`

This PR follows changes from the https://github.com/ComplianceAsCode/content/pull/7507.


matusmarhefka:
All affected rules were tested on RHEL8 and Fedora 34 VMs and I think I've found an issue
which manifests only on Fedora 34 (on RHEL8 all scenarios are passing).
When testing on RHEL8:
```
INFO - xccdf_org.ssgproject.content_rule_sshd_enable_x11_forwarding
INFO - Script correct_value.pass.sh using profile (all) OK
INFO - Script param_conflict.fail.sh using profile (all) OK
INFO - Script line_not_there.fail.sh using profile (all) OK
INFO - Script comment.fail.sh using profile (all) OK

INFO - xccdf_org.ssgproject.content_rule_sshd_disable_x11_forwarding
INFO - Script correct_value.pass.sh using profile (all) OK
INFO - Script param_conflict.fail.sh using profile (all) OK
INFO - Script line_not_there.pass.sh using profile (all) OK
INFO - Script comment.pass.sh using profile (all) OK
```
When testing on Fedora 34:
```
INFO - xccdf_org.ssgproject.content_rule_sshd_enable_x11_forwarding
INFO - Script correct_value_directory.pass.sh using profile (all) OK
INFO - Script correct_value.pass.sh using profile (all) OK
INFO - Script param_conflict_directory.fail.sh using profile (all) OK
INFO - Script wrong_value_directory.fail.sh using profile (all) OK
ERROR - Script param_conflict.fail.sh using profile (all) found issue:
ERROR - Rule evaluation resulted in pass, instead of expected fail during initial stage 
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_sshd_enable_x11_forwarding'.
ERROR - Script line_not_there.fail.sh using profile (all) found issue:
ERROR - Rule evaluation resulted in pass, instead of expected fail during initial stage 
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_sshd_enable_x11_forwarding'.
ERROR - Script comment.fail.sh using profile (all) found issue:
ERROR - Rule evaluation resulted in pass, instead of expected fail during initial stage 
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_sshd_enable_x11_forwarding'.

INFO - xccdf_org.ssgproject.content_rule_sshd_disable_x11_forwarding
INFO - Script correct_value_directory.pass.sh using profile (all) OK
INFO - Script correct_value.pass.sh using profile (all) OK
ERROR - Script param_conflict_directory.fail.sh using profile (all) found issue:
ERROR - Rule evaluation resulted in pass, instead of expected fail during initial stage 
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_sshd_disable_x11_forwarding'.
ERROR - Script wrong_value_directory.fail.sh using profile (all) found issue:
ERROR - Rule evaluation resulted in pass, instead of expected fail during initial stage 
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_sshd_disable_x11_forwarding'.
INFO - Script param_conflict.fail.sh using profile (all) OK
INFO - Script line_not_there.pass.sh using profile (all) OK
INFO - Script comment.pass.sh using profile (all) OK
```
So it seems as an issue in the OVAL template with the combination of
`missing_parameter_pass` and `config_is_distributed=true`.

mildas (changes requested):
Each `comment.*.sh` test uses wrong variable `PARAMETER` instead of `SSHD_PARAM`.


Idea for future SSGTS improvement:
Having some test scenarios templated and some not templated just because of different expected result is redundant. We might solve it with having templated both versions (pass + fail) and each of them would have condition
```
if [ PARAM == VALUE ];
    return SPECIAL_RETURN_CODE
fi
..... rest of the test scenario
```
and the SPECIAL_RETURN_CODE would cause `notapplicable` result for the test scenario. With this, both test scenarios would be executed, however only the correct one would be performed. But it is code redundancy vs test run time.
@matusmarhefka what do you think about the idea? Btw for now the current approach is fine.

matusmarhefka:
> Each `comment.*.sh` test uses wrong variable `PARAMETER` instead of `SSHD_PARAM`.
> 

Typo! thanks I will fix that.

> Idea for future SSGTS improvement:
> Having some test scenarios templated and some not templated just because of different expected result is redundant. We might solve it with having templated both versions (pass + fail) and each of them would have condition
> 
> ```
> if [ PARAM == VALUE ];
>     return SPECIAL_RETURN_CODE
> fi
> ..... rest of the test scenario
> ```
> 
> and the SPECIAL_RETURN_CODE would cause `notapplicable` result for the test scenario. With this, both test scenarios would be executed, however only the correct one would be performed. But it is code redundancy vs test run time.
> @matusmarhefka what do you think about the idea? Btw for now the current approach is fine.

I agree that this is something which we will need to address for the templated scenarios and your proposal sounds good. I was also thinking about test scenarios setting the expected test result through a variable and if the variable would not be set it would be taken from the file name (as it is currently).

mildas (approved):
I have tested selection of the sshd rules with:
```$ python3 tests/test_suite.py rule --no-reports --libvirt qemu:///system test-suite-rhel8 --datastream build/ssg-rhel8-ds.xml --duplicate-templates disable_host_auth sshd_disable_pubkey_auth sshd_disable_rhosts
 ssdh_enable_warning_banner sshd_disable_empty_passwords sshd_disable_gssapi_auth sshd_disable_kerb_auth sshd_disable_root_login sshd_disable_root_password_login sshd_disable_tcp_forwarding```
Everything works as expected. Thank you @matusmarhefka

ggbecker:
Test scenarios that echo `bad_val` to the ssh config file fail with Ansible remediation because the ansible remediation tries to validate the configuration file and `bad_val` is not a valid configuration.

`/usr/sbin/sshd -t -f %s`

https://github.com/ComplianceAsCode/content/blob/2d0d6d97760579506bca5027df5d6674f4e886b7/shared/macros-ansible.jinja#L179

PR #7526, opened 2021-09-09
Fix function call introduced by PR #7488

dodys:
#### Description:

- With the change from bash function to macro function the remediation of accounts_password_pam_unix_remember were broken for Ubuntu.


jan-cerny (approved):
Nice catch! I confirm that now it generates an expanded Jinja macro.

PR #7527, opened 2021-09-10
Allows logging stderr for remote commands in ssg_test_suite

marcusburghardt:
#### Description:

To prepare a container or a VM for ssg_test_suite remote commands are executed via ssh. Since they are actually executed remotely, in case of an error in the remote host or container, the information in the logfiles is very limited, making troubleshooting harder.

Here is a clear example, where we can see that the ssh command failed, but we have no clue about the failure cause without manual investigation:
https://github.com/ComplianceAsCode/content/pull/7514/checks?check_run_id=3544704088 (logs found in Artifacts)

We can't see detailed output because we are calling the `subprocess.run` function and its `check` parameter as `True`.
It means that if the command returns non-zero inside this subprocess function, it will raise an exception there, outside ssg_test_suite logic and consequently skipping to log the exception on ssg_test_suite logs.

In this PR I changed the `subprocess.run` function parameter `check` to `False` and treated the exception inside the ssg_test_suite to make sure we have the most valuable information for troubleshooting.

#### Rationale:

This simple update doesn't bring impacts in performance or the relevant logic of ssg_test_suites. Instead, it allows us to save a lot of time with troubleshooting for cases like this mentioned in the description or any other which may appear in the future.

### DIFF
--- a/tests/ssg_test_suite/common.py
+++ b/tests/ssg_test_suite/common.py
@@ -227,13 +227,15 @@ def run_with_stdout_logging(command, args, log_file):
     log_file.write("{0} {1}\n".format(command, " ".join(args)))
     result = subprocess.run(
             (command,) + args, encoding="utf-8", stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE, check=True)
+            stderr=subprocess.PIPE, check=False)
     if result.stdout:
         log_file.write("STDOUT: ")
         log_file.write(result.stdout)
     if result.stderr:
         log_file.write("STDERR: ")
         log_file.write(result.stderr)
+    if result.returncode:
+        raise RuntimeError("'%s' command returned non-zero." % command)
     return result.stdout
 
 


marcusburghardt:
To give more context, this was the log content when tried to test an `audit` related rule in a container without audit package installed:
```
##### audit_rules_networkconfig_modification / augen_correct_rules.pass.sh #####
ssh -o Port=37133 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null root@localhost cd /root/ssgts/audit_rules_networkconfig_modification; SHARED=/root/ssgts/shared bash -x augen_correct_rules.pass.sh
```

And here is the log content after this update:
```
##### audit_rules_networkconfig_modification / auditctl_correct_rules.pass.sh #####
ssh -o Port=38429 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null root@localhost cd /root/ssgts/audit_rules_networkconfig_modification; SHARED=/root/ssgts/shared bash -x auditctl_correct_rules.pass.sh
STDERR: Warning: Permanently added '[localhost]:38429' (ECDSA) to the list of known hosts.
+ sed -i 's%^ExecStartPost=.*%ExecStartPost=-/sbin/auditctl%' /usr/lib/systemd/system/auditd.service
sed: can't read /usr/lib/systemd/system/auditd.service: No such file or directory
+ rm -rf '/etc/audit/rules.d/*'
+ rm /etc/audit/audit.rules
rm: cannot remove '/etc/audit/audit.rules': No such file or directory
+ echo '-a always,exit -F arch=b32 -S sethostname,setdomainname -F key=audit_rules_networkconfig_modification'
auditctl_correct_rules.pass.sh: line 11: /etc/audit/audit.rules: No such file or directory
+ echo '-a always,exit -F arch=b64 -S sethostname,setdomainname -F key=audit_rules_networkconfig_modification'
auditctl_correct_rules.pass.sh: line 12: /etc/audit/audit.rules: No such file or directory
+ echo '-w /etc/issue -p wa -k audit_rules_networkconfig_modification'
auditctl_correct_rules.pass.sh: line 13: /etc/audit/audit.rules: No such file or directory
+ echo '-w /etc/issue.net -p wa -k audit_rules_networkconfig_modification'
auditctl_correct_rules.pass.sh: line 14: /etc/audit/audit.rules: No such file or directory
+ echo '-w /etc/hosts -p wa -k audit_rules_networkconfig_modification'
auditctl_correct_rules.pass.sh: line 15: /etc/audit/audit.rules: No such file or directory
+ echo '-w /etc/sysconfig/network -p wa -k audit_rules_networkconfig_modification'
auditctl_correct_rules.pass.sh: line 16: /etc/audit/audit.rules: No such file or directory
```

jan-cerny (approved):
great improvement

PR #7530, opened 2021-09-10
Make accounts_password_pam_retry work in Ubuntu

dodys:
#### Description:

- Adapt rule description
- Alter oval for Ubuntu
- Add Ubuntu specific bash
- Make tests platform generic
- Add tests for accounts_password_pam_enforcing

### DIFF
--- a/linux_os/guide/system/accounts/accounts-pam/password_quality/password_quality_pwquality/accounts_password_pam_retry/tests/correct_value.pass.sh
+++ b/linux_os/guide/system/accounts/accounts-pam/password_quality/password_quality_pwquality/accounts_password_pam_retry/tests/correct_value.pass.sh
@@ -1,8 +1,13 @@
 #!/bin/bash
-# platform = Red Hat Enterprise Linux 7,Red Hat Virtualization 4,multi_platform_fedora,multi_platform_ol,multi_platform_wrlinux
+# platform = Red Hat Enterprise Linux 7,Red Hat Virtualization 4,multi_platform_fedora,multi_platform_ol,multi_platform_wrlinux,multi_platform_ubuntu
 
 retry_cnt=3
+{{% if 'ubuntu' not in product %}}
 config_file="/etc/pam.d/system-auth"
+{{% else %}}
+config_file="/etc/pam.d/common-password"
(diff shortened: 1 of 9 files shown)


linux_os/guide/system/accounts/accounts-pam/password_quality/password_quality_pwquality/accounts_password_pam_retry/tests/correct_value.pass.sh
```diff
@@ -1,8 +1,12 @@
 #!/bin/bash
-# platform = Red Hat Enterprise Linux 7,Red Hat Virtualization 4,multi_platform_fedora,multi_platform_ol,multi_platform_wrlinux
```
jan-cerny, line 2: This change causes that this scenario will be executed also on RHEL8, but it shouldn't be executed on RHEL8 because it fails on RHEL8 and there is a different scenario only for RHEL8 rhel8_correct_value.pass.sh. You will have to list Ubuntu here instead of removing the line.

linux_os/guide/system/accounts/accounts-pam/password_quality/password_quality_pwquality/accounts_password_pam_retry/tests/wrong_value.fail.sh
```diff
@@ -1,8 +1,11 @@
 #!/bin/bash
-# platform = Red Hat Enterprise Linux 7,Red Hat Virtualization 4,multi_platform_fedora,multi_platform_ol,multi_platform_wrlinux
```
jan-cerny, line 2: Here there is also a different test scenario specific for RHEL8 so this one shouldn't be executed on RHEL8.

jan-cerny (changes requested):

linux_os/guide/system/accounts/accounts-pam/password_quality/password_quality_pwquality/accounts_password_pam_retry/tests/correct_value.pass.sh
dodys: fixed, thanks!

jan-cerny (approved):

PR #7532, opened 2021-09-10
ssg/build_yaml.py: Remove unused import

JAORMX:
when trying to use ssg.controls from another script, the fact that
build_yaml.py imported the `Controls` object resulted in a circular
dependency error. It turns out that the import wasn't actually being
used, so let's remove it.


matejak (approved):
The import is indeed not used in this module.

Thank you for the fix!

PR #7534, opened 2021-09-10
linux_os/guide/system/*/*/*/dconf_*/oval/shared.xml: Try fixing wrong regexp

apatard:
- '/' has to be escaped
- ']' has to be escaped too.

### DIFF
--- a/linux_os/guide/system/accounts/accounts-banners/gui_login_banner/dconf_gnome_banner_enabled/oval/shared.xml
+++ b/linux_os/guide/system/accounts/accounts-banners/gui_login_banner/dconf_gnome_banner_enabled/oval/shared.xml
@@ -20,7 +20,7 @@
   version="1">
     <ind:path>/etc/dconf/db/{{{ dconf_gdm_dir }}}/</ind:path>
     <ind:filename operation="pattern match">^.*$</ind:filename>
-    <ind:pattern operation="pattern match">^\[org/gnome/login-screen]([^\n]*\n+)+?banner-message-enable=true$</ind:pattern>
+    <ind:pattern operation="pattern match">^\[org/gnome/login-screen\]([^\n]*\n+)+?banner-message-enable=true$</ind:pattern>
     <ind:instance datatype="int">1</ind:instance>
   </ind:textfilecontent54_object>
 
--- a/linux_os/guide/system/software/gnome/gnome_login_screen/dconf_gnome_disable_restart_shutdown/oval/shared.xml
+++ b/linux_os/guide/system/software/gnome/gnome_login_screen/dconf_gnome_disable_restart_shutdown/oval/shared.xml
@@ -20,7 +20,7 @@
   version="1">
     <ind:path>/etc/dconf/db/{{{ dconf_gdm_dir }}}/</ind:path>
     <ind:filename operation="pattern match">^.*$</ind:filename>
-    <ind:pattern operation="pattern match">^\[org/gnome/login-screen]([^\n]*\n+)+?disable-restart-buttons=true$</ind:pattern>
+    <ind:pattern operation="pattern match">^\[org/gnome/login-screen\]([^\n]*\n+)+?disable-restart-buttons=true$</ind:pattern>
     <ind:instance datatype="int">1</ind:instance>
   </ind:textfilecontent54_object>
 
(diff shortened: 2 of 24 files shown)


apatard:
hm. trying to reproduce the issue here, but I'm getting different results:
```
bash tests/test_rule_in_container.sh --dontclean --logdir logs_dconf_gnome_screensaver_lock_enabled --remediate-using bash --name ssg_test_suite --datastream build/ssg-rhel8-ds.xml dconf_gnome_screensaver_lock_enabled

INFO - Logging into logs_dconf_gnome_screensaver_lock_enabled/test_suite.log
INFO - xccdf_org.ssgproject.content_rule_dconf_gnome_screensaver_lock_enabled
WARNING - Script correct_value.pass.sh using profile (all) notapplicable
WARNING - Rule xccdf_org.ssgproject.content_rule_dconf_gnome_screensaver_lock_enabled evaluation resulted in notapplicable
WARNING - Script comment.fail.sh using profile (all) notapplicable
WARNING - Rule xccdf_org.ssgproject.content_rule_dconf_gnome_screensaver_lock_enabled evaluation resulted in notapplicable
WARNING - Script setting_not_there.fail.sh using profile (all) notapplicable
WARNING - Rule xccdf_org.ssgproject.content_rule_dconf_gnome_screensaver_lock_enabled evaluation resulted in notapplicable
WARNING - Script correct_value_unlocked.fail.sh using profile (all) notapplicable
WARNING - Rule xccdf_org.ssgproject.content_rule_dconf_gnome_screensaver_lock_enabled evaluation resulted in notapplicable
WARNING - Script wrong_value.fail.sh using profile (all) notapplicable
WARNING - Rule xccdf_org.ssgproject.content_rule_dconf_gnome_screensaver_lock_enabled evaluation resulted in notapplicable
```

apatard:
ok. reproduced by running 2 tests instead of only one. Like:
```
tests/test_rule_in_container.sh ... dconf_gnome_banner_enabled,dconf_gnome_screensaver_lock_enabled
```
So, possibly:
- there's something wrong in the dconf_gnome_screensaver_lock_enabled files. Looking at that.
- the test suite doesn't work correctly since a previous test may impact the following ones.

apatard:
> ok. reproduced by running 2 tests instead of only one. Like:
> 
> ```
> tests/test_rule_in_container.sh ... dconf_gnome_banner_enabled,dconf_gnome_screensaver_lock_enabled
> ```
> 
> So, possibly:
> 
>     * there's something wrong in the dconf_gnome_screensaver_lock_enabled files. Looking at that.

found the issue. Local fix for SLES-15-010100 which should not have been merged here. (the fix will be provided in
a different PR)

> 
>     * the test suite doesn't work correctly since a previous test may impact the following ones.

While it's still true, it turns out that this particular issue is more an issue with this dconf_gnome_screensaver_lock_enabled test. It's not installing dconf so the rule is not check. The CI fails since the previous test dconf_gnome_banner_enabled is installing dconf.

Doing more testing. Once done will push the updates to my PR

matejak:
[Perl regexp metacharacters](https://perldoc.perl.org/perlre#Metacharacters) need to be escaped, which means that `]` can be escaped for "nice code" reasons (it is also not a metacharacter already since there is no preceding unescaped `[`), but there is absolutely no reason to escape the slash `/`.

@apatard are you willing to remove escapes of slashes from the PR? Although they don't affect the regex, they would create an impression that the escape is needed, and they would also lead to functionally harmless, but confusing occurrence of multiple escaping styles.

matejak (changes requested):
[Perl regexp metacharacters](https://perldoc.perl.org/perlre#Metacharacters) need to be escaped, which means that `]` can be escaped for "nice code" reasons (it is also not a metacharacter already since there is no preceding unescaped `[`), but there is absolutely no reason to escape the slash `/`.

@apatard are you willing to remove escapes of slashes from the PR? Although they don't affect the regex, they would create an impression that the escape is needed, and they would also lead to functionally harmless, but confusing occurrence of multiple escaping styles.

matejak:
@apatard please reply to our comments or update your PR, otherwise we will close it.

apatard:
@matejak hi. sorry, I've not yet found the time to work on that. It's still on my todo list, I "only" need to find to do that (mainly recreate my test lab and do the testing).

matejak:
> @matejak hi. sorry, I've not yet found the time to work on that. It's still on my todo list, I "only" need to find to do that (mainly recreate my test lab and do the testing).

OK, there is one more suggestion regarding test scenarios.

apatard:
hm. the PR looks "weird". There should only be 2 commits.

matejak:
Strange, but those commits should disappear if you rebase against master.

PR #7537, opened 2021-09-10
SLES-15-010570: Fix aide path

apatard:
The STIG documentation is wrong as it's looking for /usr/sbin/aide but it's /usr/bin/aide on (at least) SLE12 and SLE15 RPMs.

[ Had to fix the ansible remediation script too ]


linux_os/guide/system/software/integrity/software-integrity/aide/aide_periodic_cron_checking/ansible/shared.yml
```diff
@@ -1,3 +1,9 @@
+{{% if product in [ "sle12", "sle15" ] %}}
+{{% set aide_path = '/usr/bin/aide' %}}
+{{% else %}}
+{{% set aide_path = '/usr/sbin/aide' %}}
```
jan-cerny, line 4: To avoid defining this multiple times you should define it in products/sle15/product.yml, then add the deafult value to ssg/constants.py and handle the default value in ssg/products.py.

apatard: ok. Will update the PR to do that.

apatard: Done. I hope I did correctly what you've asked.

jan-cerny (approved):

PR #7539, opened 2021-09-10
products/sle15/product.yml: Fix grub.cfg path

apatard:
All tools are still using /boot/grub2, even in the UEFI case,
so change the value to match reality.

### DIFF
--- a/products/sle15/product.yml
+++ b/products/sle15/product.yml
@@ -17,7 +17,7 @@ release_key_fingerprint: "FEAB502539D846DB2C0961CA70AF9E8139DB7C82"
 oval_feed_url: "https://ftp.suse.com/pub/projects/security/oval/suse.linux.enterprise.15.xml"
 
 grub2_boot_path: "/boot/grub2"
-grub2_uefi_boot_path: "/boot/efi/EFI/sles"
+grub2_uefi_boot_path: "/boot/grub2"
 aide_bin_path: "/usr/bin/aide"
 
 cpes_root: "../../shared/applicability"


matejak:
Please check out e.g. https://github.com/ComplianceAsCode/content/blame/190cad8bc4ef957583b9e29c1508a1be43660388/linux_os/guide/system/bootloader-grub2/uefi/grub2_uefi_password/rule.yml - the code that indicated that `/boot/efi/EFI/sles` is the right SLE15 config file was touched less than a year ago - what could explain this?

apatard:
The problem is that the STIG documentation is asking for ``/boot/efi/EFI/sles/grub.cfg`` but from what I understand, all the tools are using ``/boot/grub2/grub.cfg``. The ``/boot/efi/EFI/sles/grub.cfg`` file merely contains:
```
set btrfs_relative_path="yes"
search --fs-uuid --set=root ....
set prefix=(${root})/boot/grub2
source "${prefix}/grub.cfg"
```
Of course, it's possible to follow the STIG documentation but at the next update, the tooling will update /boot/grub2/grub.cfg.
Here is a part of /var/log/pbl.log, which is the log file for the tool called by the kernel package after installation:
```
>>>>>>>>>>>>>>>>
<1> bootloader_entry-4067 new.126: bootloader_entry-4067 = /usr/lib/bootloader/bootloader_entry, version = 0.931, root = 0:47
<1> bootloader_entry-4067 main.269: /usr/lib/bootloader/bootloader_entry remove default 5.3.18-24.67-default vmlinuz-5.3.18-24.67-default initrd-5.3.18-24.67-default
<1> bootloader_entry-4067 main.271: bootloader = grub2-efi
<1> bootloader_entry-4067 main.277: locale = 
<1> bootloader_entry-4067 run_command.294: '/usr/lib/bootloader/grub2-efi/config' = 0, output:
<<<<<<<<<<<<<<<<
+ /usr/sbin/grub2-mkconfig -o /boot/grub2/grub.cfg
Generating grub configuration file ...
Found theme: /boot/grub2/themes/SLE/theme.txt
Found linux image: /boot/vmlinuz-5.3.18-24.75-default
Found initrd image: /boot/initrd-5.3.18-24.75-default
Found linux image: /boot/vmlinuz-5.3.18-24.70-default
Found initrd image: /boot/initrd-5.3.18-24.70-default
done
>>>>>>>>>>>>>>>>
```
So, the direct consequence will be that the system will have very few chances to run the updated kernel and not running an updated kernel seems against a lot of security practices.

jan-cerny (approved):
@apatard thanks for explanation

jan-cerny:
@apatard please resolve conflicts

PR #7549, opened 2021-09-13
SLES-15-010100: Fix check

apatard:
While the rule correctly states that the dconf key to use on SLES is
/org/gnome/desktop/lockdown, the check is using a different key:
/org/gnome/desktop/screensaver/lock-enabled.
Correct the oval test and bash remediation script to use the proper
commands / dconf path.


jan-cerny (approved):
Nice catch!

PR #7551, opened 2021-09-14
Removed deprecated populate remediation function and modified python build script to allow for the removal

j-ode:
#### Description:

- Removed the populate.sh remediation function, as it was never actually used; its calls were translated during build
- Modified the ssg/build_remediations.py build script to stop looking for the removed populate function and trimmed excess code that was used for deprecated remediation functions
- Fixed two chromium remediations and added a missing var file for one of them
- Changed all remaining populate calls to bash_instantiate_variables macro invocation
- Removed unnecessary sourcing of the remediation_functions directory

#### Rationale:

- #7348

#### To be continued:

- As this is the last remediation function being removed, in a follow-up pull request I will modify the documentation to reflect the changes made.


shared/macros-bash.jinja
```diff
@@ -8,7 +8,7 @@ echo "Setting=$varname1" >> config_file
 #}}
 {{%- macro bash_instantiate_variables() -%}}
 {{%- for name in varargs -%}}
-populate {{{ name }}}
+bash-populate {{{ name }}}
```
jan-cerny, line 11: I think that when you use this macro you want to make a bash variable assignment code generated.

For example, when there is this in a bash remediation:

```
{{{ bash_instantiate_variables("var_rekey_limit_size", "var_rekey_limit_time") }}}
```

you want to generate

```
var_rekey_limit_size="(bash-populate var_key_limit_size)"
var_rekey_limit_time="(bash-populate var_rekey_limit_time)"
```

Then, the code in `ssg/build_remediations.py` that would handle `bash-populate` replacement would be the same to the code that handles `ansible-populate`.

ssg/build_remediations.py
```diff
@@ -838,40 +798,23 @@ def expand_xccdf_subs(fix, remediation_type, remediation_functions):
...
 
-        # This remediation script utilizes some of internal remediation functions
-        # Expand shell variables and remediation functions calls with <xccdf:sub>
-        # elements
-        pattern = r'\n+(\s*(?:' + r'|'.join(remediation_functions) + r')[^\n]*)\n'
+        # This remediation script utilizes bash-populate calls
+        # Expand bash-populate calls with <xccdf:sub> elements
+
+        # Split the portion of fix.text at the bash-populate function call.
+        pattern = r'\n+(\s*(?:bash-populate)[^\n]*)\n'
```
jan-cerny, line 810: Can we avoid requiring the `\n`? It seems that the newlines causes troubles.

ssg/build_remediations.py
```diff
@@ -966,18 +871,17 @@ def expand_xccdf_subs(fix, remediation_type, remediation_functions):
...
-            # substituted both the original function call and XCCDF <sub> element
-            # for that function need to be present in the modified text of the fix
-            # Otherwise something went wrong, thus exit with failure
-            if func in modfixtext and funcxccdfsub not in modfixtext:
-                sys.stderr.write("Error performing XCCDF <sub> substitution "
-                                 "for function %s in %s fix. Exiting...\n"
-                                 % (func, fix.get("rule")))
-                sys.exit(1)
+        # Then efine expected XCCDF sub element form for this function
+        funcxccdfsub = "<sub idref=\"function_%s\"" % "bash-populate"
```
jan-cerny, line 875: We probably don't want to have any element with ID starting with `function` anywhere, we want to get rid of them.

ssg/build_remediations.py
```diff
@@ -716,15 +658,15 @@ def get_rule_dir_remediations(dir_path, remediation_type, product=None):
...
     """For those remediation scripts utilizing some of the internal SCAP
     Security Guide remediation functions expand the selected shell variables
     and remediation functions calls with <xccdf:sub> element
 
-    This routine translates any instance of the 'populate' function call in
+    This routine translates any instance of the 'bash-populate' function call in
     the form of:
 
-            populate variable_name
+            (bash-populate variable_name)
```
jan-cerny, line 669: These aren't functions anymore. And we don't have any other remediation functions anymore. You will have to rewrite this comment completely especially you will have to remove the things about the remediation functions below.

ssg/build_remediations.py
```diff
@@ -747,15 +689,15 @@ def expand_xccdf_subs(fix, remediation_type, remediation_functions):
             <sub idref="function_function_name"/>
             function_name "arg1" "arg2" ... "argN"
     """
-
+    if fix is not None:
+        fix_text = fix.text
```
jan-cerny, line 693: And if it's None it would probably make sense to terminate, otherwise the variable fix_text would be undefined.

jan-cerny (changes requested):
Please remove the merge commit from this PR by doing a rebase.

Unfortunately, there seems to be a problem with gating.

```
  File "/__w/content/content/build-scripts/generate_bash_remediation_functions.py", line 88, in <module>
    main()
  File "/__w/content/content/build-scripts/generate_bash_remediation_functions.py", line 52, in main
    for file_ in sorted(os.listdir(args.input_dir)):
FileNotFoundError: [Errno 2] No such file or directory: '/__w/content/content/shared/bash_remediation_functions'
```

shared/macros-bash.jinja
```diff
@@ -8,7 +8,7 @@ echo "Setting=$varname1" >> config_file
 #}}
 {{%- macro bash_instantiate_variables() -%}}
 {{%- for name in varargs -%}}
-populate {{{ name }}}
+{{{ name }}}="(bash-populate {{{ name }}})"
```
jan-cerny, line 11: I guess that you can now remove `{{# this line is intentionally left blank #}}` because you don't match the newlines anymore, but in the generated code there is quite a lot superfluous newlines.

jan-cerny (changes requested):

jan-cerny (approved):

jan-cerny:
Great job!

PR #7554, opened 2021-09-14
Fix set_tests_properties for fix_rules-sort_prodtypes

Mab879:
#### Description:

Fix `set_tests_properties` on the `fix_rules-sort_prodtypes` test. In some cases this will fail due to not having `test-rule-dir-json` done first.

#### Rationale:

Fix the test suite on new builds.


vojtapolasek (approved):
Looks good, thanks for noticing it.

PR #7555, opened 2021-09-14
sshd_use_approved_ciphers: Add test scenarios for RHEL7 CIS

matusmarhefka:
#### Description:

- Added new test scenarios for rule `sshd_use_approved_ciphers` to cover RHEL7 CIS.

### DIFF
--- a/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers/tests/cis_rhel7_correct_value_full.pass.sh
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_ciphers/tests/cis_rhel7_correct_value_full.pass.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+#
+# platform = Red Hat Enterprise Linux 7
+# profiles = xccdf_org.ssgproject.content_profile_cis
+
+if grep -q "^Ciphers" /etc/ssh/sshd_config; then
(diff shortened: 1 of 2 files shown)


jan-cerny (approved):
The changes look good to me and the test scenarios pass on a RHEL7 VM.

@matusmarhefka 

I have noticed that the SSGTS GH action doesn't execute the tests, which is unfortunate, because I had to do it locally. Do you have any idea why it happened?

I also have noticed that there are other test scenarios for this rule that are marked as tests for RHEL 7 STIG profile, but this rule isn't a part of the RHEL 7 STIG profile, so the test run produces errors. Instead there is rule  sshd_use_approved_ciphers_ordered_stig in the RHEL 7 STIG profile.

matusmarhefka:
> The changes look good to me and the test scenarios pass on a RHEL7 VM.
> 
> @matusmarhefka
> 
> I have noticed that the SSGTS GH action doesn't execute the tests, which is unfortunate, because I had to do it locally. Do you have any idea why it happened?
> 

It is because content test filtering doesn't support it ATM, but @mildas is working on it.

> I also have noticed that there are other test scenarios for this rule that are marked as tests for RHEL 7 STIG profile, but this rule isn't a part of the RHEL 7 STIG profile, so the test run produces errors. Instead there is rule sshd_use_approved_ciphers_ordered_stig in the RHEL 7 STIG profile.

Aha, so the rule got unselected from the profile and we should remove these scenarios as they are no longer applicable.

matusmarhefka:
@jan-cerny https://github.com/ComplianceAsCode/content/pull/7561

PR #7559, opened 2021-09-15
Allow specific Jinja processing for test suite

marcusburghardt:
#### Description:

There are some specific cases where a protected package should be removed, mostly for testing purposes, but this doesn't work with traditional ways, as expected.

A good example is the `sudo` package, where more details were discussed in this PR:
https://github.com/ComplianceAsCode/content/pull/7519

For treating this situation of protected packages, it was necessary to create rule specific test scenarios to override the templated ones. Although this approach works, it is reactive and demand manual analysis.

#### Rationale:

Since we can previously know which packages are protected in a system which use YUM or DNF, the approach suggested in this PR treats this exception directly on `bash-macros.jinja`. The main advantage of this is that test scenarios don't need any change. Even if the list of protected packages in a system changes, this will be automatically and properly treated in the same standard.

This macro can be easily updated for similar exceptions in other distros.

### DIFF
--- a/shared/macros-bash.jinja
+++ b/shared/macros-bash.jinja
@@ -519,6 +519,7 @@ zypper install -y "{{{ package }}}"
     Remove a package
 
     Uses the right command based on pkg_manager property defined in product.yml.
+    When used in a test scenario, the macro will remove even protected packages.
 
     :param package: name of the package
     :type package: str
@@ -527,11 +528,11 @@ zypper install -y "{{{ package }}}"
 {{%- if pkg_manager is defined -%}}
 {{%- if pkg_manager == "yum" or pkg_manager == "dnf" -%}}
 if rpm -q --quiet "{{{ package }}}" ; then
-    if grep -q "{{{ package }}}" $(find /etc/dnf/protected.d /etc/yum/protected.d -type f); then
-        {{{ pkg_manager }}} remove -y --setopt protected_packages= "{{{ package }}}"
-    else
-        {{{ pkg_manager }}} remove -y "{{{ package }}}"
-    fi
+{{% if SSG_TEST_SUITE_ENV %}}
+    {{{ pkg_manager }}} remove -y --setopt protected_packages= "{{{ package }}}"
+{{% else %}}
+    {{{ pkg_manager }}} remove -y "{{{ package }}}"
+{{% endif %}}
 fi
 {{%- elif pkg_manager == "apt_get" -%}}
 DEBIAN_FRONTEND=noninteractive apt-get remove -y "{{{ package }}}"
--- a/shared/macros-bash.jinja
+++ b/shared/macros-bash.jinja
@@ -527,7 +527,11 @@ zypper install -y "{{{ package }}}"
 {{%- if pkg_manager is defined -%}}
 {{%- if pkg_manager == "yum" or pkg_manager == "dnf" -%}}
 if rpm -q --quiet "{{{ package }}}" ; then
-    {{{ pkg_manager }}} remove -y "{{{ package }}}"
+    if grep -q "{{{ package }}}" $(find /etc/dnf/protected.d /etc/yum/protected.d -type f); then
+        {{{ pkg_manager }}} remove -y --setopt protected_packages= "{{{ package }}}"
+    else
+        {{{ pkg_manager }}} remove -y "{{{ package }}}"
+    fi
 fi
 {{%- elif pkg_manager == "apt_get" -%}}
 DEBIAN_FRONTEND=noninteractive apt-get remove -y "{{{ package }}}"
--- a/tests/ssg_test_suite/common.py
+++ b/tests/ssg_test_suite/common.py
@@ -277,6 +277,10 @@ def get_product_context(product=None):
     # property to bypass this error.
     product_yaml['cmake_build_type'] = 'Debug'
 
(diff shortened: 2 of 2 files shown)


shared/macros-bash.jinja
```diff
@@ -527,7 +527,11 @@ zypper install -y "{{{ package }}}"
 {{%- if pkg_manager is defined -%}}
 {{%- if pkg_manager == "yum" or pkg_manager == "dnf" -%}}
 if rpm -q --quiet "{{{ package }}}" ; then
-    {{{ pkg_manager }}} remove -y "{{{ package }}}"
+    if grep -q sudo $(find /etc/dnf/protected.d /etc/yum/protected.d -type f); then
```
yuumasato, line 530: Should it be like this?
```suggestion
    if grep -q {{{ package }}} $(find /etc/dnf/protected.d /etc/yum/protected.d -type f); then
```
Not sure I agree with all `package_removed` remediations having a hardcoded check to verify if `sudo` is protected, and then removing whatever package is defined in `{{{ package }}}` (which might be protected).

matusmarhefka (approved):

yuumasato (changes requested):

shared/macros-bash.jinja
matusmarhefka: Good point, it should indeed be `{{{ package }}}` instead of `sudo` here.

yuumasato:
> The main advantage of this is that test scenarios don't need any change.

Yes, but we are still changing the remediation of many rules.

shared/macros-bash.jinja
yuumasato: But then, the remediation becomes "remove any package, even if protected".
In this case, we might just clear the `protected_packages` list for all rules. (Not sure it is a good idea)

marcusburghardt: Yes, it definitively should be "{{{ package }}}". Thanks for catching. Updating.
Regarding cleaning the `protected_packages` list I also agree that additional test would be interesting, like validating "if it was required by a test scenario or not", but I am not sure about the best way to do this now.
Alternatively, with less risk but similar effect is to remove using RPM directly.

Since we don't have risk considering the current rules and its respective packages, I think we can polish this conditional later, but for now simplify the test scenarios.

marcusburghardt:
> > The main advantage of this is that test scenarios don't need any change.
> 
> Yes, but we are still changing the remediation of many rules.

Not actually, because all the packages currently impacted by this macro are not protected packages. Also, I don't think it is realistic to remove a protected package since this action would have direct impact in security and/or stability of the system.

shared/macros-bash.jinja
yuumasato: > Regarding cleaning the `protected_packages` list I also agree that additional test would be interesting, like validating "if it was required by a test scenario or not", but I am not sure about the best way to do this now.

I don't think this is possible, the check for the test scenario would have to happen at build time, and the macro evaluates the `if` conditional at remediation time. It seems to me that a content that behaves differently under a testing environment would be necessary, and for that we either add conditionals used only for testing purposes (that will end up being shipped) or have a dedicated build type for testing. Neither approach looks good to me.

> Since we don't have risk considering the current rules and its respective packages, I think we can polish this conditional later, but for now simplify the test scenarios.

The conditional as it is doesn't change anything, it will always remove a package, protected or not.
Additionally, I'm not sure there is anything the remediation can check (at remediation time) to decide whether to remove a protected package or not because it is in a testing environment.

marcusburghardt: Another alternative is to create a new macro only for testing, like `bash_package_remove_test(package)`. This, for sure, wouldn't affect remediation but would demand update in test scenario scripts. Also doable, I think.

yuumasato:
> Not actually, because all the packages currently impacted by this macro are not protected packages.

Well, behavior may still be the same, but they will have an unnecessary conditional.

shared/macros-bash.jinja
yuumasato: @marcusburghardt An analogous idea to `bash_package_remove_test` is to expand `bash_package_remove` differently for a build and a Test Suite run.
With the right Jinja conditionals this could be done. Is there a Jinja variable that is only set during an Test Suite expansion?

marcusburghardt: I am not sure if we any common Jinja variable for Test Suite. I did some tests yesterday with this approach, treating the conditionals inside the `template.py`, but it became too complex for the value included by it, IMHO. And, probably it would be necessary to include some additional variable inside the rule.yml when calling the template. Then I moved to this much simpler approach directly on bash-macros.

yuumasato: @marcusburghardt I guess a variable could be injected into `product_yaml` in https://github.com/ComplianceAsCode/content/blob/master/tests/ssg_test_suite/common.py#L429

Which can later be checked by `bash_package_remove` to expand with `protected_packages` option.

yuumasato:
Assigned blocker and bugfix labels, as the the `sudo` test scenario aborts any Test Suite run with the rule.
CC: @matusmarhefka @mildas

yuumasato:
@marcusburghardt To include this fix in 0.1.58 release, this PR should be proposed against `stabilization` branch.

yuumasato (approved):

PR #7561, opened 2021-09-15
sshd_use_approved_ciphers: Rule is no longer part of STIG, remove STIG test scenarios

matusmarhefka:
#### Description:

- Rule `sshd_use_approved_ciphers` is no longer selected in RHEL7 STIG profile, therefore remove RHEL7 STIG related test scenarios


jan-cerny:
The rule has been replaced in the rhel7 stig profile by rule sshd_use_approved_ciphers_ordered_stig which has its own test scenarios.

jan-cerny (approved):

PR #7563, opened 2021-09-16
Improved file ownership test scenario coverage

marcusburghardt:
#### Description:

File ownership related tests moved to templates in already templated rules.
Consequently enabled test scenarios for all other rules which use same template

#### Rationale:

80 test scripts removed in favor of 6 test scripts included as generic templates.


jan-cerny (approved):
LGTM, tests passes for both templates on a rhel8 VM.

PR #7570, opened 2021-09-16
Fix typo in rsyslog streamdriver remediations

yuumasato:
#### Description:

- Fix typo in rsyslog streamdriver remediations

#### Rationale:

- The remediations don't need to escape '$'.


Mab879 (changes requested):
It looks like the test suite for these rules is failing with these changes. This should be addressed before merging.

yuumasato:
Only the Ansible remediations need to drop the `\`, in Bash the backslash prevents variable substitution.

Mab879 (approved):

Issue #7292, opened 2021-07-27
RHEL8 STIG - service_auditd_enabled fail (pass in remediation phase, fail in final check)


mildas:
#### Description of problem:
The `service_auditd_enabled` rule check fails after remediating a machine with STIG profile and restarting it.

#### SCAP Security Guide Version:
7eb7ef1

#### Operating System Version:
RHEL 8

#### Steps to Reproduce:
1. `python3 tests/test_suite.py profile --libvirt qemu:///system test_suite_vm --datastream /tmp/ssg-rhel8-ds.xml --xccdf-id scap_org.open-scap_cref_ssg-rhel8-xccdf-1.2.xml --mode online --remediate-using oscap xccdf_org.ssgproject.content_profile_stig`

#### Actual Results:
xccdf_org.ssgproject.content_rule_service_auditd_enabled - fail 

#### Expected Results:
xccdf_org.ssgproject.content_rule_service_auditd_enabled - pass 

#### Additional Information/Debugging Steps:
In the remediation phase, it shows the rule passed (package installed and service is active). However, in the final phase, after machine restart, it fails (audit package is installed, but the service is not running).

It might be that some audit configuration breaks it and it can't be started.

freddieRv:
On OL7, the rule `auditd_overflow_action` breaks the auditd daemon after a reboot because the config option `overflow_action` is not supported. This might be an issue with the specific OL7 audit RPM though. The same option is supported on the audit shipped with OL8.

ggbecker:
I believe these following changes fix the issue described originally here:

https://github.com/ComplianceAsCode/content/pull/6910

https://github.com/ComplianceAsCode/content/pull/6910/files#diff-6518a0e9b97463fd5d2a337b46c053367785594713c979dab64d0f6e0363bf2d

https://github.com/ComplianceAsCode/content/pull/6910/files#diff-6d36e8d97141aafdfd999784156ba4ff387d3c4a6ac9594198bfc3d059b0fc3c

It was basically a syntax error when editing the `/etc/audit/audit.conf` file which caused the service to fail when starting.

Can you double check it @mildas

PR #7573, opened 2021-09-16
Update config path in auditd_overflow_action

freddieRv:
#### Description:

- Updated rule text, oval check, ansible and bash remediations to use the correct path to configure audisp plugin.
- Also updated auditd_audispd_syslog_plugin_activated rule text to mention correct path for OL8

#### Rationale:

- Config options for audisp plugin are placed in /etc/audisp/auditd.conf for audit package previous to v3
- Placing the config option `overflow_action` in /etc/audit/auditd.conf prevents the auditd service from starting on reboot
- This is the issue I mentioned in #7292

### DIFF
--- a/linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_overflow_action/ansible/shared.yml
+++ b/linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_overflow_action/ansible/shared.yml
@@ -4,7 +4,13 @@
 # complexity = low
 # disruption = low
 
-{{{ ansible_set_config_file(file="/etc/audit/auditd.conf",
+{{% if product in ["rhel8", "fedora", "ol8", "rhv4"] %}}
+{{% set audisp_conf_file = "/etc/audit/auditd.conf" %}}
+{{% else %}}
+{{% set audisp_conf_file = "/etc/audisp/audispd.conf" %}}
+{{% endif %}}
+
+{{{ ansible_set_config_file(file=audisp_conf_file,
                   parameter="overflow_action",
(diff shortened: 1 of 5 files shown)


linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_overflow_action/ansible/shared.yml
```diff
@@ -4,7 +4,13 @@
 # complexity = low
 # disruption = low
 
-{{{ ansible_set_config_file(file="/etc/audit/auditd.conf",
+{{% if product in ["rhel8", "fedora", "ol8", "rhv4"] %}}
```
Mab879, line 7: Instead of duplicating this logic for the config file location, I think we should create a variable for it.

Mab879 (changes requested):

linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_overflow_action/ansible/shared.yml
freddieRv: Hi @Mab879 , would the work on https://github.com/ComplianceAsCode/content/pull/7594 be a better approach?

Mab879: I think I like that better, I have requested some feedback from some others on that PR as well.

freddieRv:
Closing in favor of https://github.com/ComplianceAsCode/content/pull/7594

PR #7577, opened 2021-09-17
Improvements on wireless_disable_interfaces rule

marcusburghardt:
#### Description:

Included test scenarios
CPE test included to report notapplicable where there is no wifi
Greater accuracy to detect non-disabled wifi interfaces

#### Rationale:

The previous approach to validade that wifi interfaces are disabled relies on the content of /proc/net/wireless.
WiFi interfaces will only appear in this file if they are actually connected. It means that a wifi interfaces not present in this file can be disconnected but still active, ready to be connected, like in these examples:

**WiFi Interface Enabled and Connected**
```
2: wlo1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DORMANT group default qlen 1000
    link/ether xx:xx:xx:xx:xx:xx brd ff:ff:ff:ff:ff:ff
    altname wlp0s20f3
```

**WiFi Interface Enabled and Disconnected**
```
2: wlo1: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state DOWN mode DORMANT group default qlen 1000
    link/ether xx:xx:xx:xx:xx:xx brd ff:ff:ff:ff:ff:ff permaddr yy:yy:yy:yy:yy:yy
    altname wlp0s20f3
```

**WiFi Interface Disabled**
```
2: wlo1: <BROADCAST,MULTICAST> mtu 1500 qdisc noqueue state DOWN mode DEFAULT group default qlen 1000
    link/ether xx:xx:xx:xx:xx:xx brd ff:ff:ff:ff:ff:ff permaddr yy:yy:yy:yy:yy:yy
    altname wlp0s20f3
```

The new approach instead check for the UP FLAG of wifi interfaces to make sure they are actually disabled (not UP).

In parallel, a CPE has been included to report "not applicable" instead of "error" where there is no wireless interface.
This CPE relies on the presence of /proc/net/wireless file, which is always present when any wireless interfaces are detected by the system, even if the wireless interfaces are inactive.

### DIFF
--- a/linux_os/guide/system/network/network-wireless/wireless_software/wireless_disable_interfaces/oval/shared.xml
+++ b/linux_os/guide/system/network/network-wireless/wireless_software/wireless_disable_interfaces/oval/shared.xml
@@ -2,15 +2,20 @@
   <definition class="compliance" id="wireless_disable_interfaces" version="1">
     {{{ oval_metadata("All wireless interfaces should be disabled.") }}}
     <criteria>
-      <criterion comment="query /proc/net/wireless" test_ref="test_wireless_disable_interfaces" />
+      <criterion comment="check if wifi interfaces are disabled" test_ref="test_wireless_disable_interfaces" negate="true" />
     </criteria>
   </definition>
-  <ind:textfilecontent54_test check="all" check_existence="none_exist" comment="query /proc/net/wireless" id="test_wireless_disable_interfaces" version="1">
-    <ind:object object_ref="object_wireless_disable_interfaces" />
-  </ind:textfilecontent54_test>
-  <ind:textfilecontent54_object id="object_wireless_disable_interfaces" version="1">
-    <ind:filepath>/proc/net/wireless</ind:filepath>
-    <ind:pattern operation="pattern match">^\s*[-\w]+:</ind:pattern>
-    <ind:instance datatype="int" operation="equals">1</ind:instance>
-  </ind:textfilecontent54_object>
+
+  <unix:interface_test check="all" check_existence="at_least_one_exists" id="test_wireless_disable_interfaces" version="1" comment="check if UP flag is present on wifi interfaces">
+    <unix:object object_ref="object_active_wifi_interfaces" />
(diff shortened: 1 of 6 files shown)


linux_os/guide/system/network/network-wireless/wireless_software/wireless_disable_interfaces/oval/shared.xml
```diff
@@ -2,15 +2,20 @@
...
-  <ind:textfilecontent54_test check="all" check_existence="none_exist" comment="query /proc/net/wireless" id="test_wireless_disable_interfaces" version="1">
-    <ind:object object_ref="object_wireless_disable_interfaces" />
-  </ind:textfilecontent54_test>
-  <ind:textfilecontent54_object id="object_wireless_disable_interfaces" version="1">
-    <ind:filepath>/proc/net/wireless</ind:filepath>
-    <ind:pattern operation="pattern match">^\s*[-\w]+:</ind:pattern>
-    <ind:instance datatype="int" operation="equals">1</ind:instance>
-  </ind:textfilecontent54_object>
+
+  <unix:interface_test check="all" check_existence="none_exist" id="test_wireless_disable_interfaces" version="1" comment="check if UP flag is present on wifi interfaces">
```
jan-cerny, line 9: it looks like that the test failure is legit 
```
oval:ssg-test_wireless_disable_interfaces:tst:1 - No state should be referenced when check_existence has a value of 'none_exist'.
```

marcusburghardt:
> @marcusburghardt How can I run the test scenarios? Do I need some special VM?

Hi @jan-cerny , it is necessary a physical machine with a wireless interface. Unfortunately I didn't find any feasible way to simulate a wireless interface in virtual machines. I did some tests in my laptop while testing the modifications.

This was one of the reasons which motivated me to create this CPE, to report "notapplicable" where a wireless interface is not present. Otherwise it would return error.

jan-cerny (approved):

PR #7587, opened 2021-09-20
SSGTS: Switch message to warning when a rule has no test scenarios

### LINKED ISSUE #7483: SSGTS throws error when there are no test scenarios available for a given rule
#### Description of problem:
SSGTS reports error when there are no test scenarios for a given rule, for example:

`ERROR - No matching rule ID found for '['configure_opensc_card_drivers', 'force_opensc_card_drivers']'`

It should report correct information and in this case a warning message should be enough instead of `ERROR`.

#### Steps to Reproduce:

1. `tests/test_rule_in_container.sh --dontclean --logdir logs_ansible --remediate-using ansible --name ssg_test_suite --datastream ssg-rhel8-ds.xml configure_opensc_card_drivers,force_opensc_card_drivers`

#### Actual Results:
```
INFO - Logging into logs_ansible/test_suite.log
ERROR - No matching rule ID found for '['configure_opensc_card_drivers', 'force_opensc_card_drivers']'
WARNING - Nothing has been tested!
```

#### Expected Results:
```
INFO - Logging into logs_ansible/test_suite.log
WARNING - No test scenarios found for following rules: '['configure_opensc_card_drivers', 'force_opensc_card_drivers']'
WARNING - Nothing has been tested!
```

@matusmarhefka @mildas This should help SSGTS integration with Github Actions since it relies on output messages to decide when the test should fail. So, misleading `ERROR` messages can mess with test results.

ggbecker:
#### Description:
- SSGTS: Switch message to warning when a rule has no test scenarios
  - When the rule has no test scenarios the SSGTS will just report a warning
that there are not test scenarios to be tested. Additionally remove
unused code.

#### Rationale:

- Fixes #7483

### DIFF
--- a/tests/ssg_test_suite/combined.py
+++ b/tests/ssg_test_suite/combined.py
@@ -33,7 +33,6 @@ class CombinedChecker(rule.RuleChecker):
     """
     def __init__(self, test_env):
         super(CombinedChecker, self).__init__(test_env)
-        self._matching_rule_found = False
 
         self.rules_not_tested_yet = set()
         self.results = list()
--- a/tests/ssg_test_suite/rule.py
+++ b/tests/ssg_test_suite/rule.py
@@ -118,7 +118,6 @@ class RuleChecker(oscap.Checker):
     """
     def __init__(self, test_env):
         super(RuleChecker, self).__init__(test_env)
-        self._matching_rule_found = False
 
         self.results = list()
         self._current_result = None
@@ -282,10 +281,8 @@ class RuleChecker(oscap.Checker):
     def _test_target(self, target):
         rules_to_test = self._get_rules_to_test(target)
         if not rules_to_test:
-            self._matching_rule_found = False
-            logging.error("No matching rule ID found for '{0}'".format(target))
+            logging.error("No tests found matching the rule ID(s) '{0}'".format(", ".join(target)))
             return
-        self._matching_rule_found = True
 
         scenarios_by_rule = dict()
         for rule in rules_to_test:


mildas:
@ggbecker I've touched this topic with @dahaic and better approach will be to have missing test scenarios as SSGTS error to enforce contributors to create test scenarios.

In case, a rule can't have test scenario (e.g. `wireless_disable_interfaces`), there should be possibility to add exception for the rule. This exception might be for example in a form of `.no_tests` file in `tests/` folder.

ggbecker:
> @ggbecker I've touched this topic with @dahaic and better approach will be to have missing test scenarios as SSGTS error to enforce contributors to create test scenarios.
> 
> In case, a rule can't have test scenario (e.g. `wireless_disable_interfaces`), there should be possibility to add exception for the rule. This exception might be for example in a form of `.no_tests` file in `tests/` folder.

I understand the reasoning behind it, but I don't fully agree with this. We have many rules that don't have test scenarios for multiple historical reasons. To be able to enforce such policy, we would have to catch up with all those missing tests, otherwise sometimes just a simple fix could become a extra burden on a external contributor just because the rule, they are trying to submit a fix, doesn't have test scenarios.

If we manage to get to a point when every rule has either test scenarios or a waiver, then I'm ok with that.

Either way, I think we should have at least the log message updated.

mildas:
> I understand the reasoning behind it, but I don't fully agree with this. We have many rules that don't have test scenarios for multiple historical reasons. To be able to enforce such policy, we would have to catch up with all those missing tests, otherwise sometimes just a simple fix could become a extra burden on a external contributor just because the rule, they are trying to submit a fix, doesn't have test scenarios.

I see your point. From developer point of view, it is extra burden. But in many cases, the test scenario are pretty simple - remove/add line, create file or so - not much work. And with templated test scenarios, the situation with test scenarios will get better even faster.
A simple fix can break the rule. Implementing test scenarios right away can prevent these problems and save our time in future. And if the simple fix is not related to code, then content-test-filtering shouldn't suggest running tests in GH Actions.

> If we manage to get to a point when every rule has either test scenarios or a waiver, then I'm ok with that.

Without enforcing creation of tests, we will not get to this point soon.

> Either way, I think we should have at least the log message updated.

The warning type of message will not result in SSGTS fail, true?

dahaic:
> I understand the reasoning behind it, but I don't fully agree with this. We have many rules that don't have test scenarios for multiple historical reasons. To be able to enforce such policy, we would have to catch up with all those missing tests, otherwise sometimes just a simple fix could become a extra burden on a external contributor just because the rule, they are trying to submit a fix, doesn't have test scenarios.
> 
When contributor submits a fix, I'd expect to provide the respective test scenario anyways. If it's a fix that is not connected to a testable scenario, then we can help by providing some basic tests. In general though - if I'm touching something, it's in my own interest to make sure I don't break stuff. So necessity to write additional tests should not be perceived as a burden.

If nothing else, it's a way to ensure the rule behavior will stay the same.

ggbecker:
All right. I'm convinced, I've reverted the log message to error.

jan-cerny (approved):
I think that this is already a big improvement because it removes the confusion. We can revisit the topic later.

PR #7588, opened 2021-09-20
Implement `filter_rules` for Profiles

JAORMX:
This implements a mechanism to filter rules for a profile based on
attributes of the rule object.

This allows us to do stuff like this:

```
filter_rules: 'platform == "ocp4-node"'
```

... Since the filter takes the Python syntax, more complex rules can be
created. e.g. one is able to use all boolean operators from the
language.

The variables available are exactly the ones available in the `Rule`
object from *build_yaml.py*.

### DIFF
--- a/products/ocp4/profiles/pci-dss-node.profile
+++ b/products/ocp4/profiles/pci-dss-node.profile
@@ -15,7 +15,7 @@ title: 'PCI-DSS v3.2.1 Control Baseline for Red Hat OpenShift Container Platform
 description: |-
     Ensures PCI-DSS v3.2.1 security configuration settings are applied.
 
-filter_rules: 'platform != "ocp4-node"'
+filter_rules: 'platform == "ocp4-node"'
 
 selections:
     - pcidss_ocp4:all:base
--- a/ssg/build_yaml.py
+++ b/ssg/build_yaml.py
@@ -1049,7 +1049,7 @@ def rule_filter_from_def(filterdef):
     def filterfunc(rule):
         # Remove globals for security and only expose
         # variables relevant to the rule
-        return eval(filterdef, {}, rule.__dict__)
+        return eval(filterdef, {"__builtins__": None}, rule.__dict__)
     return filterfunc
 
 
--- a/ssg/build_yaml.py
+++ b/ssg/build_yaml.py
@@ -376,13 +376,23 @@ class ResolvableProfile(Profile):
         super(ResolvableProfile, self).__init__(* args, ** kwargs)
         self.resolved = False
         self.resolved_selections = set()
+        self.rule_filter = None
+
+    def read_yaml_contents(self, yaml_contents):
+        super(ResolvableProfile, self).read_yaml_contents(yaml_contents)
+        self.rule_filter = rule_filter_from_def(
+            yaml_contents.pop("filter_rules", None))
 
     def _controls_ids_to_controls(self, controls_manager, policy_id, control_id_list):
         items = [controls_manager.get_control(policy_id, cid) for cid in control_id_list]
         return items
 
     def _merge_control(self, control):
-        self.selected.extend(control.rules)
+        if self.rule_filter is not None:
+            filteredrules = (str(rule) for rule in control.rules if self.rule_filter(rule))
+            self.selected.extend(filteredrules)
+        else:
+            self.selected.extend(str(rule) for rule in control.rules)
         for varname, value in control.variables.items():
             if varname not in self.variables:
                 self.variables[varname] = value
@@ -1032,6 +1042,17 @@ class Group(object):
         return self.id_
 
 
+def rule_filter_from_def(filterdef):
+    if filterdef is None or filterdef == "":
+        return None
+
+    def filterfunc(rule):
+        # Remove globals for security and only expose
+        # variables relevant to the rule
+        return eval(filterdef, {}, rule.__dict__)
+    return filterfunc
+
+
 class Rule(object):
     """Represents XCCDF Rule
     """
@@ -1086,6 +1107,18 @@ class Rule(object):
         self.template = None
         self.local_env_yaml = None
         self.current_product = None
+
+    def __deepcopy__(self, memo):
+        cls = self.__class__
+        result = cls.__new__(cls)
+        memo[id(self)] = result
+        for k, v in self.__dict__.items():
+            # These are difficult to deep copy, so let's just re-use them.
+            if k != "template" and k != "local_env_yaml":
+                setattr(result, k, deepcopy(v, memo))
+            else:
+                setattr(result, k, v)
+        return result
 
     @classmethod
     def from_yaml(cls, yaml_file, env_yaml=None, sce_metadata=None):
@@ -1497,6 +1530,9 @@ class Rule(object):
         root = self.to_xml_element()
         tree = ET.ElementTree(root)
         tree.write(file_name)
+
+    def __str__(self):
+        return self.id_
 
 
 class DirectoryLoader(object):
--- a/ssg/controls.py
+++ b/ssg/controls.py
@@ -114,18 +114,19 @@ class Control():
                 varname, value = item.split("=", 1)
                 control.variables[varname] = value
             else:
+                rule_yaml = get_rule_path_by_id(content_dir, item)
+                if rule_yaml is None:
+                    # item not found in benchmark_root
+                    continue
+                rule = ssg.build_yaml.Rule.from_yaml(rule_yaml, env_yaml)
+
                 # Check if rule is applicable to product, i.e.: prodtype has product id
                 if product is None:
                     # The product was not specified, simply add the rule
-                    control.rules.append(item)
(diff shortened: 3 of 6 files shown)


products/ocp4/profiles/pci-dss-node.profile
```diff
@@ -0,0 +1,21 @@
...
+        - mrogers950
+
+reference: https://www.pcisecuritystandards.org/documents/PCI_DSS_v3-2-1.pdf
+
+title: 'PCI-DSS v3.2.1 Control Baseline for Red Hat OpenShift Container Platform 4'
+
+description: |-
+    Ensures PCI-DSS v3.2.1 security configuration settings are applied.
+
+filter_rules: 'platform == "ocp4-node"'
```
yuumasato, line 18: It feels strange to have this separated from the control selection item (i.e. `pcidss_ocp4:all:base`).

Do we also want/need the `filter_rules` to work independently of control based profiles?

ssg/build_yaml.py
```diff
@@ -1032,6 +1042,16 @@ def __str__(self):
         return self.id_
 
 
+def rule_filter_from_def(filterdef):
+    if filterdef is None or filterdef == "":
+        return None
+    def filterfunc(rule):
+        # Remove globals for security and only expose
+        # variables relevant to the rule
+        return eval(filterdef, {}, rule.__dict__)
```
yuumasato, line 1051: Neat!

products/ocp4/profiles/pci-dss-node.profile
yuumasato: Reading the description again there is no mention of the controls file. I read the PR with a different context in mind, :stuck_out_tongue:

JAORMX:
Pull-request updated, HEAD is now a5e4be3b471e5b99918f6bc5e6e668bbfe410188

JAORMX:
Pull-request updated, HEAD is now f75ff0507e90eb0503c1bad95b357a0ea8b6d159

JAORMX:
Pull-request updated, HEAD is now c7f387b56681f445e946c542806747e658de941a

ggbecker:
@JAORMX please check https://github.com/JAORMX/content/pull/4 it fixes the problems with the controls unit testing.

JAORMX:
/retest

IdP deployment error (took too long)

yuumasato (changes requested):
I like the filtering mechanism. But what is its expected scope? Is it expected to work with profiles defined without control files?

I tried restricting RHEL-8 `ospp` and `anssi_bp_28_minimal` profiles with `filter_rules: 'platform == "machine"'`, and it seems to have worked only with the `anssi` profile.

I think the behavior should be extended to all profiles, or the syntax mechanism integrated into the "control selection mechanism":

`     - pcidss_ocp4:all:base['platform == "ocp4-node"']`

Edit: testing with  `'platform != "machine"'` the difference is more evident.

products/ocp4/profiles/pci-dss-node.profile
matejak: We have discussed the question of interface - should this filtering clause be here, in the profile file, or should it be closer to the controls manipulation, because that's what we want - to influence what rules we get from those controls?

I suggest that we keep it as it is for now - we can easily change the interface to something else, as the code that asks rules about their platform would stay, and only the place it is invoked from would change.
By looking at `pcidss_ocp4:all:base`, it is clear that the control syntax has been abused enough, and `pcidss_ocp4:all:base:platform == something` could be the solution for now, but wouldn't benefit the project in the long-term. So let's come with a better-looking and more robust and intuitive instantiation syntax, and only then let's use it.

ssg/controls.py
```diff
@@ -114,18 +114,19 @@ def from_control_dict(cls, control_dict, env_yaml=None, default_level=["default"
                 varname, value = item.split("=", 1)
                 control.variables[varname] = value
             else:
+                rule_yaml = get_rule_path_by_id(content_dir, item)
```
matejak, line 117: This is OK for now, but generally rules loaded this way won't posses inherited group platforms, and in some cases they may not be valid YAMLs due to jinja2 constructs. However, this problem can be solved later.

ssg/build_yaml.py
```diff
@@ -1498,6 +1531,9 @@ def to_file(self, file_name):
         tree = ET.ElementTree(root)
         tree.write(file_name)
 
+    def __str__(self):
+        return self.id_
```
matejak, line 1535: Do we need a string representation of a rule? A human-readable representation would be rather its title, and a reliable representation is basically its yaml.

matejak:
> I like the filtering mechanism. But what is its expected scope? Is it expected to work with profiles defined without control files?
> 

I can imagine that we could get clean container profiles "for free", which could make sense.

> I tried restricting RHEL-8 `ospp` and `anssi_bp_28_minimal` profiles with `filter_rules: 'platform == "machine"'`, and it seems to have worked only with the `anssi` profile.

It could be caused by wrong detection of platforms because of group inheritance. I would expect `"machine" in / not in platforms` to do the right thing, but that's not possible ATM. But I don't see this as a blocker to the PR, as the target product OCP4 doesn't suffer from those problems, and the behavior can be fixed for the rest by making the build process more sane.

> 
> I think the behavior should be extended to all profiles, or the syntax mechanism integrated into the "control selection mechanism":

See also my reply https://github.com/ComplianceAsCode/content/pull/7588#discussion_r713162890 - I suggest not to abuse the control line any more, but I see your point and I can't say that I would disagree.

@JAORMX What was the reason that controls got the list of data structures rather than rule IDs?

products/ocp4/profiles/pci-dss-node.profile
yuumasato: > We have discussed the question of interface - should this filtering clause be here, in the profile file, or should it be closer to the controls manipulation, because that's what we want - to influence what rules we get from those controls?

TL;DR; I tend to think that that the filtering clause should be in the profile.
--
The control file keeps a set of controls organized according to the policy it is tracking, it organizes the controls into policy levels and maps rules to them. It can be seen as an organized "rule gatherer".

Now the pool of rules gathered by the control file need to make it into profiles.
And, although the profiles can be aligned to the policy (ANSSI), they can be organized in an unrelated way (PCI-DSS for OCP4). What I mean by that is that the separation into `master` and `node` is an OCP4 construct, the policy doesn't predict the product being "divided" into two parts.

So I think the filter should be defined in the profile to avoid adding product specific constructs in the control file.

> By looking at `pcidss_ocp4:all:base`, it is clear that the control syntax has been abused enough, and `pcidss_ocp4:all:base:platform == something` could be the solution for now, but wouldn't benefit the project in the long-term - that would be the third selection option added to the one-line syntax, as first we select `all` controls, we then restrict those with `level == base`, and then we have this platform filter. It is likely that another selection requirements are yet to be discovered. So let's come with a better-looking and more robust and intuitive instantiation syntax, and only then let's use it.

I agree that the control syntax is being pushed beyond what it was designed for, which, as far as I remember, was to be backward compatible with the profile syntax.
Maybe consider backward compatibility breaking changes?

JAORMX:
> > I like the filtering mechanism. But what is its expected scope? Is it expected to work with profiles defined without control files?
> 
> I can imagine that we could get clean container profiles "for free", which could make sense.
> 
> > I tried restricting RHEL-8 `ospp` and `anssi_bp_28_minimal` profiles with `filter_rules: 'platform == "machine"'`, and it seems to have worked only with the `anssi` profile.
> 
> It could be caused by wrong detection of platforms because of group inheritance. I would expect `"machine" in / not in platforms` to do the right thing, but that's not possible ATM. But I don't see this as a blocker to the PR, as the target product OCP4 doesn't suffer from those problems, and the behavior can be fixed for the rest by making the build process more sane.

OK, should we then address more functionality in a subsequent PR? We could document this until it's ready then. But this would be a first iteration.

> 
> > I think the behavior should be extended to all profiles, or the syntax mechanism integrated into the "control selection mechanism":
> 
> See also my reply [#7588 (comment)](https://github.com/ComplianceAsCode/content/pull/7588#discussion_r713162890) - I suggest not to abuse the control line any more, but I see your point and I can't say that I would disagree.
> 
> @JAORMX What was the reason that controls got the list of data structures rather than rule IDs?

It allowed me to easily traverse the rules and get all the data from them, such as the platform. I can't do that just with the rule IDs.

ssg/build_yaml.py
JAORMX: @matejak this is needed for utils/controls-template.html

ssg/build_yaml.py
matejak:
```suggestion
        return eval(filterdef, {"__builtins__": None}, rule.__dict__)
```

According to https://stackoverflow.com/a/9558001/592892 this can be improved, but eval is still dangerous, although that modification restricts the attack surface.

yuumasato (approved):
The intended use with control files work, but it will need to be iterated to work with profiles defined without controls.

PR #7590, opened 2021-09-20
Fix RHEL8 CIS reference of mount_option_dev_shm_nodev

matusmarhefka:
#### Description:

- Rule `mount_option_dev_shm_nodev` should have `1.1.15` RHEL8 CIS reference.


jan-cerny (approved):
This is in line with the controls file.

PR #7591, opened 2021-09-20
Add rule no_duplicate_uids

dodys:
#### Description:

- The Ubuntu operating system must uniquely identify interactive users.
- This rule doesn't have a remediation, as this requires manual interaction.

### DIFF
--- a/linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_duplicate_uids/rule.yml
+++ b/linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_duplicate_uids/rule.yml
@@ -1,0 +1,20 @@
+documentation_complete: true
+
+prodtype: ubuntu2004
+
+title: 'Ensure no duplicate UIDs exist'
+
+description: |-
+    Although the useradd program will not let you create a duplicate User ID (UID),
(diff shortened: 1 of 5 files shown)


linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_duplicate_uids/rule.yml
```diff
@@ -0,0 +1,23 @@
...
+description: |-
+    Although the useradd program will not let you create a duplicate User ID (UID),
+    it is possible for an administrator to manually edit the /etc/passwd file
+    and change the UID field. Users must be assigned unique UIDs for
+    accountability and to ensure appropriate access protections.
+
+rationale: |-
+    Users must be assigned unique UIDs for accountability and to ensure
+    appropriate access protections.
+
```
jan-cerny, line 16: I suggest adding a warning that no automated remediation is available because it would remove user accounts or change their UIDs which could break the system.

dodys: I added it in a new commit, let me know if it looks good :)

jan-cerny (approved):

PR #7592, opened 2021-09-20
Improvements on file permissions test scenarios

marcusburghardt:
Included templated test scrips for "file_permissions" template.
Reduced the number of scripts and increased the coverage by ensuring
that any templated rule using this template is now covered by test scenarios.


jan-cerny (approved):
Great work!

PR #7596, opened 2021-09-20
Break the shared bash install_smartcard_packages remediation

### LINKED ISSUE #7593: Fallback to template remediation when shared remediation doesn't list a platform
#### Description of problem:

With the recent commit 71664ae, and even though Ubuntu is not listed in the platform in the shared.sh, during build it is not falling back to the templated bash, so for that rule it doesn't get any remediation anymore. This can become a problem as some distros might use specific shared(.sh/.yml) remediations instead of the one from the template, affecting distros using the template.

There are two solutions here:
1. Rename shared remediation to their distro's name: (shared.sh -> sle12.sh/sle15.sh/ubuntu.sh/...
2. Make build system fallback to template remediation when platform is not listed in shared file.

I prefer to have the build system smart enough to handle this, but if this is not possible then please let us know in the comments.

Master

Ubuntu 20.04 LTS

#### Steps to Reproduce:

Any profile that has install_smartcard_packages and uses the template bash remediation can be affected by this. 

1. Build such profile
2. The final <product>-script-stig.sh won't have any remediation for install_smartcard_packages.

#### Actual Results:

Templated remediation not being picked during build.

#### Expected Results:

Build should fallback to template remediation when platform is not listed in shared remediation.

N/A

Adding @teacup-on-rockingchair to make SUSE aware of this.
CC @ggbecker : creating this ticket as we discussed a bit about this earlier.

teacup-on-rockingchair:
#### Description:

- Break the shared bash install_smartcard_packages remediation

#### Rationale:

- Rename the install_smartcard_packages bash shared remediation to sle15 and copy to sle12 so the bash remediation will be used only for those platforms
- Rest of the platforms should use the template bash remediation

- Fixes #7593


Mab879 (approved):

Mab879:
Thanks for the swift action!

Issue #7118, opened 2021-06-15
The build system does not warn when Rule ID and OVAL ID mismatch


ggbecker:
#### Description of problem:

The build system does not detect mismatches of Rule id and OVAL id. If these mismatches occur, then OVAL content won't get into the datastream and this rule will always return `notchecked`. A warning can be added in these cases to inform the user that there's a mismatch. Additionally the OVAL ids could be converted to use the Jinja variable [`{{{ rule_id }}}`](https://github.com/ComplianceAsCode/content/blob/master/ssg/build_ovals.py#L289) available during build time, so it would prevent any potential mismatch.

cipherboy:
I think the alternative to Jinja is just to set (or force-trample) the OVAL ID in the build system. Then we can eventually remove all IDs and just infer it during build.

PR #7598, opened 2021-09-21
Check that OVAL ID matches rule ID

jan-cerny:
#### Description:

The build system will print a message when the OVAL definition ID doesn't match the rule ID.

#### Rationale:

Before this change,  it silently ignored the mismatch and the rule missed any OVAL check in the built data stream.

Fixes: #7118

### DIFF
--- a/ssg/build_ovals.py
+++ b/ssg/build_ovals.py
@@ -254,6 +254,16 @@ def _check_oval_version_from_oval(oval_file_tree, oval_version):
         return True
 
 
+def _check_rule_id(oval_file_tree, rule_id, file_path):
+    for definition in oval_file_tree.findall(
+            "./{%s}def-group/{%s}definition" % (oval_ns, oval_ns)):
+        definition_id = definition.get("id")
+        if definition_id != rule_id:
+            msg = ("OVAL definition ID '%s' doesn't match "
+                "rule ID '%s' in '%s'\n" % (definition_id, rule_id, file_path))
+            sys.stderr.write(msg)
+
+
 def checks(env_yaml, yaml_path, oval_version, oval_dirs):
     """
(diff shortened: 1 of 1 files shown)


ssg/build_ovals.py
```diff
@@ -250,6 +254,17 @@ def _check_oval_version_from_oval(xml_content, oval_version):
...
 
+def _check_rule_id(oval_file_tree, rule_id, file_path):
+    for definition in oval_file_tree.findall(
+            "./{%s}def-group/{%s}definition" % (oval_ns, oval_ns)):
+        definition_id = definition.get("id")
+        if definition_id != rule_id:
+            msg = (
+                "OVAL definition ID '%s' doesn't match rule ID '%s' "
+                "in '%s'\n" % (definition_id, rule_id, file_path))
+            sys.stderr.write(msg)
```
matejak, line 265: It would be nice to have the issue discovery and issue handling separate, preferably in a separate function.

matejak (changes requested):
Thank you for the enhancement, it's a great contribution - and the only thing to do would be to decouple the reporting from the rest of the check.

ssg/build_ovals.py
jan-cerny: I would prefer to keep it here because I find the information in this message valuable.

matejak: I am not asking to drop anything, just to decouple the detection from reporting. Right now, the function is immobile - printing that exact message may not be desirable side-effect all of the time.

jan-cerny:
@matejak I have moved warning message out from the relevant function

matejak:
Great addition, I have verified that it works. Thank you very much!

PR #7600, opened 2021-09-21
OCP4: Simply moderate profiles

JAORMX:
Since they effectively include the CIS rules, let's directly use the
`extends` key to ensure they're always up-to-date.

This simplifies the profile since we no longer need to add all the rules
explicitly.


jhrozek (approved):
I think this is fine, we just need to be mindful about changing any rules in CIS and keep in mind that they'd affect fedramp as well.

Vincent056 (approved):
/lgtm

PR #7604, opened 2021-09-21
Add rules banner_etc_issue_net and sshd_enable_warning_banner_net

dodys:
#### Description:

- Ubuntu STIG profile has a banner in /etc/issue.net, because of that we created two rules to address this:
banner_etc_issue_net and sshd_enable_warning_banner_net

### DIFF
--- a/linux_os/guide/system/accounts/accounts-banners/banner_etc_issue_net/rule.yml
+++ b/linux_os/guide/system/accounts/accounts-banners/banner_etc_issue_net/rule.yml
@@ -1,0 +1,63 @@
+documentation_complete: true
+
+prodtype: ubuntu2004
+
+title: 'Modify the System Login Banner'
+
+description: |-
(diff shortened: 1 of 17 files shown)


linux_os/guide/system/accounts/accounts-banners/banner_etc_issue_net/rule.yml
```diff
@@ -0,0 +1,63 @@
+documentation_complete: true
+
+prodtype: ubuntu2004
+
+title: 'Modify the System Login Banner'
+
+description: |-
+    To configure the system login banner edit <tt>/etc/issue</tt>. Replace the
```
jan-cerny, line 8:
```suggestion
    To configure the system login banner edit <tt>/etc/issue.net</tt>. Replace the
```

dodys: nice catch!
fixed, thanks!

jan-cerny (approved):

PR #7605, opened 2021-09-21
controls: PCI-DSS section 10

mrogers950:
@jhrozek @JAORMX


controls/pcidss_ocp4.yml
```diff
@@ -1838,137 +1838,153 @@ controls:
   rules: []
 
 - id: Req-10.1
-  title: 10.1 Implement audit trails to link all access to system components to each
-    individual user.
+  title: >-
+    10.1 Implement audit trails to link all access to system components to each individual
+    user.
+  description: |-
+    All actions taken by users of OpenShift are logged and capable of being used to satisfy audit requirements.
```
JAORMX, line 1845: I've been doing these assessments on the `notes` section. What do you think about moving these to that key?

controls/pcidss_ocp4.yml
```diff
@@ -1979,82 +1995,117 @@ controls:
...
     title: 10.4.3 Time settings are received from industry-accepted time sources.
     levels:
     - base
-    notes: ''
-    status: pending
-    rules: []
+    status: automated
+    rules:
+      - chronyd_or_ntpd_specify_remote_servers
+      - chronyd_or_ntpd_specify_multiple_server
```
JAORMX, line 2024: interesting... we don't have an OpenShift-specific setting for this and these come from `linux_os/guide`... So they won't be included in the final profile as they're not applicable nor usable by the `ocp4` nor `ocp4-node` products. I think we should remove them and mark this rule as not applicable, as this is applicable to RHCOS instead. At some point we should have some RHCOS profile for this

JAORMX (changes requested):

controls/pcidss_ocp4.yml
mrogers950: Sounds good, I'll fix it

controls/pcidss_ocp4.yml
mrogers950: Ok, I wasn't sure if they were usable in this profile. Fixed.

JAORMX (approved):

PR #7609, opened 2021-09-21
Add remaining Blueprint templates

evgenyz:
Templates: `service_enabled`, `service_disabled`, `mount` and `grub2_bootloader_argument`.

This PR will also extend `mount` template with `min_size` argument, for the recommended minimal size of the partition.

### DIFF
--- a/shared/templates/grub2_bootloader_argument/blueprint.template
+++ b/shared/templates/grub2_bootloader_argument/blueprint.template
@@ -1,0 +1,4 @@
+# platform = multi_platform_rhel,multi_platform_fedora,multi_platform_ol,multi_platform_rhv,multi_platform_ubuntu,multi_platform_sle
+
+[customizations.kernel]
+append = "{{{ ARG_NAME_VALUE }}}"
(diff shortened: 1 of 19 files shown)


shared/templates/grub2_bootloader_argument/blueprint.template
```diff
@@ -0,0 +1,4 @@
+# platform = multi_platform_rhel,multi_platform_fedora,multi_platform_ol,multi_platform_rhv,multi_platform_ubuntu,multi_platform_sle
```
vojtapolasek, line 1: Any reason for not putting just multi_platform_all here?
https://complianceascode.readthedocs.io/en/latest/manual/developer/04_style_guide.html#header

vojtapolasek (commented):
Looks good, please see the question.

shared/templates/grub2_bootloader_argument/blueprint.template
evgenyz: Parts of BP won't be available for all platforms. It is to be refined further, but the current one I just copied from the sibling template.

vojtapolasek (approved):
Looks good, approving.

vojtapolasek:
The failing check is caused by the test running in the container, where you can't work with partitions very well. I confirmed it works on local VM.

PR #7610, opened 2021-09-22
Level 2 rules for CIS

Achilleasein:
#### Description:

- _Implementing the following CIS server lvl 2 tests require_emergency_target_auth and require_singleuser_auth_

### DIFF
--- a/linux_os/guide/system/permissions/partitions/mount_option_dev_shm_nodev/rule.yml
+++ b/linux_os/guide/system/permissions/partitions/mount_option_dev_shm_nodev/rule.yml
@@ -25,7 +25,7 @@ identifiers:
 references:
     cis-csc: 11,13,14,3,8,9
     cis@rhel7: 1.1.8
-    cis@rhel8: 1.1.5
+    cis@rhel8: 1.1.15
     cis@sle12: 1.1.8
     cis@sle15: 1.1.16
     cis@ubuntu1804: 1.1.14


Achilleasein:
We can probably squish the tests into one.

jan-cerny:
Hello, please follow https://chris.beams.io/posts/git-commit/.

matusmarhefka:
@Achilleasein Actually your test scripts are in wrong directory (`bash`), they should go into `tests` directory.

linux_os/guide/system/accounts/accounts-physical/require_emergency_target_auth/bash/cis_rhel7_require_emergency_target_auth.fail.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
matusmarhefka, line 1: I think we don't need another fail scenario, there is already one and we can make it applicable to all RHEL versions by modifying `platforms` metadata to contain `multi_platform_rhel`. (pass scenarios still need to be separate for RHEL7 and RHEL8).

linux_os/guide/system/accounts/accounts-physical/require_emergency_target_auth/bash/cis_rhel7_require_emergency_target_auth.pass.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+#
+# platform = Red Hat Enterprise Linux 7
+# profiles = xccdf_org.ssgproject.content_profile_cis
```
matusmarhefka, line 4: I think we can remove this, it should work for any profile selecting this rule.

matusmarhefka (changes requested):

linux_os/guide/system/accounts/accounts-physical/require_emergency_target_auth/bash/cis_rhel7_require_emergency_target_auth.fail.sh
matusmarhefka: Pretty much same applies also to the `require_singleuser_auth` rule.

matusmarhefka:
Please move your changes in your fork from master into another branch, I would also change the name of the test scripts, they don't need to contain rule ID, file names are missing pass/fail at the end and also remove `cis` from the file name as they are applicable also to other profiles. So change it to something like `rhel7_correct_value.pass.sh`.

PR #7613, opened 2021-09-22
Ground work for implementation of CPE applicability language

vojtapolasek:
#### Description:

- create new classes for CPEAL platform-specification, platform, logical-test fact-ref
- implement parsing mechanism for more complex platform specification (it will probably change in follow-up PR):
  - OR is done by using separate lines
  - and is done by using & in a line
  - negation is done with ! preceding the platform name
- the new functionality is added into CPEProduct class
- the XCCDF document now contains cpe-lang:platform-specification with all the platforms
- platforms are currently using fact-ref, which points to the good old cpe name we used before. This is done partially to prevent big breakage, but mainly because xccdf 1.1 does not support high enough version of CPE applicability language to use check-fact-ref directly
- groups and rules already use these new platforms, profiles and benchmarks not yet
- the goal is to just move from cpe-names to CPE platforms, no changes to applicability

### DIFF
--- a/ssg/build_cpe.py
+++ b/ssg/build_cpe.py
@@ -106,17 +106,18 @@ class ProductCPEs(object):
     def get_product_cpe_names(self):
         return [ cpe.name for cpe in self.product_cpes.values() ]
 
-    def parse_platform_definition(self, platforms):
+    def parse_platform_definition(self, platform_line):
         # let's construct the platform id
-        id_list = [self._convert_platform_to_id(platform) for platform in platforms]
-        id = "_or_".join(id_list)
-        id = "cpe_platform_" + id
+        id = "cpe_platform_" + self._convert_platform_to_id(platform_line)
         platform = CPEALPlatform(id)
-        # add initial test
-        initial_test = CPEALLogicalTest(operator="OR", negate="false")
-        platform.add_test(initial_test)
-        for platform_line in platforms:
+        # add initial test if the line does not contain AND or NEGATE operator
+        # othervise the presence of & or ! will create the test while parsing
+        if "&" not in platform_line and "!" not in platform_line:
+            initial_test = CPEALLogicalTest(operator="OR", negate="false")
+            platform.add_test(initial_test)
             initial_test.add_object(self.parse_platform_line(platform_line))
+        else:
+            platform.add_test(self.parse_platform_line(platform_line))
         return platform
 
     def parse_platform_line(self, platform_line):
--- a/ssg/build_cpe.py
+++ b/ssg/build_cpe.py
@@ -219,7 +219,11 @@ class CPEALPlatformSpecification(object):
         self.platforms = []
 
     def add_platform(self, platform):
-        self.platforms.append(platform)
+        """
+        we check if semantically equal platform is not already in the list of platforms
+        """
+        if platform not in self.platforms:
+            self.platforms.append(platform)
 
     def to_xml_element(self):
         cpe_platform_spec = ET.Element(
--- a/ssg/build_cpe.py
+++ b/ssg/build_cpe.py
@@ -108,15 +108,12 @@ class ProductCPEs(object):
 
     def parse_platform_definition(self, platforms):
         # let's construct the platform id
-        #print ("parsing platforms")
-        #print (platforms)
-        id_list = [self._convert_platform_to_id(platform) for platform in platforms ]
+        id_list = [self._convert_platform_to_id(platform) for platform in platforms]
         id = "_or_".join(id_list)
         id = "cpe_platform_" + id
-        #print (id)
         platform = CPEALPlatform(id)
         # add initial test
-        initial_test = CPEALTest(operator = "OR", negate = "false")
+        initial_test = CPEALTest(operator="OR", negate="false")
         platform.add_test(initial_test)
         for platform_line in platforms:
             initial_test.add_object(self.parse_platform_line(platform_line))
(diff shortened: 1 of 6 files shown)


ssg/build_cpe.py
```diff
@@ -104,6 +106,46 @@ def get_cpe_name(self, cpe_id):
     def get_product_cpe_names(self):
         return [ cpe.name for cpe in self.product_cpes.values() ]
 
+    def parse_platform_definition(self, platforms):
+        # let's construct the platform id
+        id_list = [self._convert_platform_to_id(platform) for platform in platforms]
+        id = "_or_".join(id_list)
```
yuumasato, line 112: This is "fusing" multiple 'OR'ed platform entries into a single `CPEPlatform`, but I think it should create a `CPEPlatform` per entry in`platforms`.

I think merging 'OR'ed platforms into a single 'CPEPlatform' will reduce their reusability and increase the number of unique CPEALPlatforms in the content.

For example:
```
Rule AB.yml
platforms:
  - Expression A
  - Expression B

Rule AC.yml
platforms:
  - Expression A
  - Expression C

Rule BC.yml
platforms:
  - Expression B
  - Expression C

Rule ABC.yml
platforms:
  - Expression A
  - Expression B
  - Expression C
```

When each rule gets a `CPEPlatform` that includes the 'OR'ed platforms, each rule may get a unique complex `logical-test'. From the example above we would get 4 CPEALPlatforms. (Each rule may end up having its own unique platform).

If we make a `CPEPlatform` from each entry in the `platforms` key, we get much smaller `CPEALPlatforms` that can be used by multiple rules.
From the example, we would get  3 CPEALPlatforms, one for each Expression.

ssg/build_cpe.py
```diff
@@ -104,6 +106,46 @@ def get_cpe_name(self, cpe_id):
...
+        id = "cpe_platform_" + id
+        platform = CPEALPlatform(id)
+        # add initial test
+        initial_test = CPEALTest(operator="OR", negate="false")
+        platform.add_test(initial_test)
+        for platform_line in platforms:
+            initial_test.add_object(self.parse_platform_line(platform_line))
+        return platform
+
+    def parse_platform_line(self, platform_line):
```
yuumasato, line 122: This function assumes that a `platform_line` has a single operator.
It should be possible to have  `platform_line` like 'A & B & !C`.

yuumasato (commented):
I know the PR is WIP, but just mentioning in case you are not aware:
When an XCCDF item references a CPEALPlatform, the ID in `@id` should be prefixed by `#`.
E.g.: `<xccdf-1.2:platform idref="cpe_platform_machine"/>`
Ref: XCCDF 6.2.5

yuumasato:
The prefix `CPEAL` seems a bit too long for me, how about shortening it to just `CPE`?

JAORMX:
Will this PR include docs on the expression format? If not, could you post it as part of this PR's description? That would help me review this as well. I'm hoping to take this into use as soon as it lands.

vojtapolasek:
Yes I plan to include documentation, I added it into the "what needs to be done" section.

JAORMX (dismissed):

vojtapolasek:
Thank you everyone for suggestions. The grammar parser and more human readable syntax sound really good. Due to recent changes and overall complexity of the change, I decided to steer this PR in a bit different direction.
I plan to achieve the same applicability as we have now, but expressed with CPE platforms instead of cpe-names. This will be a base for further changes in follow-up pull requests.

ssg/build_cpe.py
```diff
@@ -245,3 +357,39 @@ def extract_referred_nodes(tree_with_refs, tree_with_ids, attrname):
...
+        raise (NotImplementedError("not implemented yet"))
+    else:
+        # the line should contain a CPEAL ref name
+        cpealfactref = CPEALFactRef(product_cpe.get_cpe_name(platform_line))
+        return cpealfactref
+
+def convert_platform_to_id(platform):
+    id = platform.replace(" ", "")
+    id = id.replace("&", "_and_")
+    id = id.replace("!", "not_")
```
matejak, line 376: This can be removed, as `&` and `!` are not allowed anyway.

ssg/build_cpe.py
```diff
@@ -245,3 +357,39 @@ def extract_referred_nodes(tree_with_refs, tree_with_ids, attrname):
...
+    """
+    This function takes one line of platform definition from yaml file and returns a
+    CPE platform with appropriate tests and factrefs.
+    """
+    # let's construct the platform id
+    id = "cpe_platform_" + convert_platform_to_id(platform_line)
+    platform = CPEALPlatform(id)
+    # add initial test if the line does not contain AND or NEGATE operator
+    # othervise the presence of & or ! will create the test while parsing
+    if "&" not in platform_line and "!" not in platform_line:
```
matejak, line 389: This is also always true.

matejak:
This PR introduces the SCAP-side support for more complex rule applicability by means of the CPE Applicability Language.

I have proposed to the code that was originally present in this PR that bridged this new functionality with rule definitions. This is a problem that can be addressed separately, and that is, in itself, quite complex, as it has to appeal to content authors, and it has to be usable by remediations etc.

Therefore, the current PR doesn't contain code that is suitable for unit-testing - the ability to express a complex platform definition in XML can be tested by means of system tests using a scanner, whereas the ability to extract a platform definition from a rule yaml into its intermediate representation will need unit tests.

matejak (approved):

matejak:
I am merging the PR, so we can move on. One set of suggestions became not applicable because the scope has been reduced, while others are still somehow relevant, but the discussion around them got silent, and they can be addressed subsequently.
Thank you for the work!

yuumasato (approved):

PR #7617, opened 2021-09-22
profile_tool.py: Update fields displayed in HTML

ggbecker:
#### Description:
- profile_tool.py: Update fields displayed in HTML
  - New fields were introduced in the profile statistics but weren't updated
in the HTML page generation, which cause to fields being rendered its
whole list of rules to the HTML instead of having it written to a file
so it can be nice displayed.

#### Rationale:

- Makes page like this to be nicely displayed again: https://jenkins.complianceascode.io/job/scap-security-guide-stats/Statistics/rhel8/product-statistics/statistics.html

### DIFF
--- a/build-scripts/profile_tool.py
+++ b/build-scripts/profile_tool.py
@@ -240,7 +240,11 @@ def main():
             'missing_puppet_fixes',
             'missing_anaconda_fixes',
             'missing_cces',
-            'ansible_parity'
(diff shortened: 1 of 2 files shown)


JAORMX (approved):

jan-cerny (approved):
Yes, I built the stats and it's way more readable.

PR #7621, opened 2021-09-22
remove rule disable_prelink from rhel7 cis

vojtapolasek:
#### Description:

- remove rule from rhel7 cis control

#### Rationale:

It was not properly aligned with the benchmark.

### DIFF
--- a/controls/cis_rhel7.yml
+++ b/controls/cis_rhel7.yml
@@ -395,11 +395,9 @@ controls:
     levels:
     - l1_server
     - l1_workstation
-    status: partial
+    status: planned
     notes: >-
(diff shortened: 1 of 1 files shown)


controls/cis_rhel7.yml
```diff
@@ -395,11 +395,9 @@ controls:
     levels:
     - l1_server
     - l1_workstation
-    status: partial
+    automated: no
```
Mab879, line 398: This should be `status` see https://complianceascode.readthedocs.io/en/latest/manual/developer/03_creating_content.html#reporting-status

Mab879 (approved):

PR #7624, opened 2021-09-22
Included test scenarios for CIS

marcusburghardt:
#### Description:

Included CIS specific test scenarios for the following rules:
- accounts_maximum_age_login_defs
- banner_etc_issue
- banner_etc_motd

#### Rationale:

Missing test scenarios covered.

### DIFF
--- a/controls/cis_rhel7.yml
+++ b/controls/cis_rhel7.yml
@@ -477,7 +477,7 @@ controls:
     status: automated
     rules:
     - banner_etc_motd
-    - login_banner_text=usgcb_default
+    - login_banner_text=cis_default
 
   - id: 1.7.2
     title: Ensure local login warning banner is configured properly (Automated)
@@ -487,7 +487,7 @@ controls:
     status: automated
     rules:
     - banner_etc_issue
-    - login_banner_text=usgcb_default
+    - login_banner_text=cis_default
 
   - id: 1.7.3
     title: Ensure remote login warning banner is configured properly (Automated)
--- a/linux_os/guide/system/accounts/accounts-banners/banner_etc_issue/tests/banner_etc_issue_cis.fail.sh
+++ b/linux_os/guide/system/accounts/accounts-banners/banner_etc_issue/tests/banner_etc_issue_cis.fail.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+# profiles = xccdf_org.ssgproject.content_profile_cis, xccdf_org.ssgproject.content_profile_cis_server_l1, xccdf_org.ssgproject.content_profile_cis_workstation_l1, xccdf_org.ssgproject.content_profile_cis_workstation_l2
+
+# cis_default banner
+echo "Authorized uses only. All activity will not be monitored and reported." > /etc/issue
--- a/controls/cis_rhel8.yml
+++ b/controls/cis_rhel8.yml
@@ -473,7 +473,7 @@ controls:
     status: automated
     rules:
       - banner_etc_motd
-      - login_banner_text=usgcb_default
+      - login_banner_text=cis_default
 
   - id: 1.8.1.2
     title: Ensure local login warning banner is configured properly (Automated)
@@ -483,7 +483,7 @@ controls:
     status: automated
     rules:
       - banner_etc_issue
-      - login_banner_text=usgcb_default
+      - login_banner_text=cis_default
 
   # NEEDS RULE
   # https://github.com/ComplianceAsCode/content/issues/5225
--- a/linux_os/guide/system/accounts/accounts-banners/banner_etc_issue/tests/banner_etc_issue_cis.pass.sh
+++ b/linux_os/guide/system/accounts/accounts-banners/banner_etc_issue/tests/banner_etc_issue_cis.pass.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+# profiles = xccdf_org.ssgproject.content_profile_cis, xccdf_org.ssgproject.content_profile_cis_server_l1, xccdf_org.ssgproject.content_profile_cis_workstation_l1, xccdf_org.ssgproject.content_profile_cis_workstation_l2
+
+# cis_default banner
+echo "Authorized uses only. All activity may be monitored and reported." > /etc/issue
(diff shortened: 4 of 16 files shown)


linux_os/guide/system/accounts/accounts-restrictions/password_expiration/accounts_maximum_age_login_defs/tests/correct_cis.fail.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
matusmarhefka, line 1: I would rename this scenario to something like `incorrect_cis.fail.sh`.

controls/cis_rhel7.yml
```diff
@@ -477,7 +477,7 @@ controls:
     status: automated
     rules:
     - banner_etc_motd
-    - login_banner_text=usgcb_default
+    - login_banner_text=cis_default
```
matusmarhefka, line 480: Good catch.

linux_os/guide/system/accounts/accounts-restrictions/password_expiration/accounts_maximum_age_login_defs/tests/correct_cis.fail.sh
marcusburghardt: Thanks for catching @matusmarhefka. Copy and past without attention is a poison. : ))

marcusburghardt: Done

linux_os/guide/system/accounts/accounts-banners/banner_etc_issue/tests/banner_etc_issue_cis.fail.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+# profiles = xccdf_org.ssgproject.content_profile_cis
```
vojtapolasek, line 2: I suggest you add also other profiles derived from the CIS control file.

linux_os/guide/system/accounts/accounts-banners/banner_etc_motd/tests/banner_etc_motd_disa_dod_default_banner.pass.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+#
```
vojtapolasek, line 2: I suggest you remove these lines, you are removing them actually in tests for the other rule.

vojtapolasek (changes requested):
Thank you for the updates, please see comments.

linux_os/guide/system/accounts/accounts-banners/banner_etc_motd/tests/banner_etc_motd_disa_dod_default_banner.pass.sh
marcusburghardt: This empty `#` in the header of test scripts is very common (at least 190 scripts have this now). I would suggest we document the standard for this: with or without this `#` when header parameters are not present so we can avoid doubts in the future. For now, I didn't find something. Are you aware of something? If they are there without any reason, I agree to remove what is not necessary, but not only here, everywhere.

linux_os/guide/system/accounts/accounts-banners/banner_etc_issue/tests/banner_etc_issue_cis.fail.sh
marcusburghardt: Hi @vojtapolasek , do you mean `server_l1`, `server_l2`, `workstation_l1`, and workstation_l2`? If so, are they necessary in this header since they are derivations of the main profile? Anyways, I am testing here and will include them since this could be helpful and doesn't look to have any down side. : )

linux_os/guide/system/accounts/accounts-banners/banner_etc_motd/tests/banner_etc_motd_disa_dod_default_banner.pass.sh
vojtapolasek: I agree we should settle on something. My point is that in this PR, you are removing these lines for one set of tests and adding them for another set of tests. I just want to keep it consistent in the PR :-)

linux_os/guide/system/accounts/accounts-banners/banner_etc_issue/tests/banner_etc_issue_cis.fail.sh
marcusburghardt: > Well it could be helpful if you want to test the whole profile, but not the cis server level 2 which is marked as cis in the content. So remaining profiles are cis_server_l1, cis_workstation_l1 and cis_workstation_l2.

I have tested the rule against these other profiles, just to see the behavior and indeed it is good to include these derivations. Thanks for rise this. : ) Sending an update soon.

linux_os/guide/system/accounts/accounts-banners/banner_etc_motd/tests/banner_etc_motd_disa_dod_default_banner.pass.sh
marcusburghardt: Ah, got your point. : ) Actually, I am not removing it in some files and including in others. I am actually incrementing the line with a parameter for some files which didn't have a parameter before. Inconsistent is prone to interpretation here, since we don't have a definition. hehe. In many cases, we only have a single "#" where no header parameter is necessary (I am following this since most of the scripts I checked follow this). However, checking all files we can also find many cases where this empty '#' is there together with other parameters. And, we have even a third case where nothing is there besides the shebang. Although this is nit pick, I think we should define a standard and gradually adjust everything according to the defined standard. I will suggest something for the Style Guide.

marcusburghardt: https://github.com/ComplianceAsCode/content/pull/7639

jan-cerny (approved):

PR #7628, opened 2021-09-23
WIP: Add directory to add tailorings for OCP

JAORMX:
Now that we're introducing remediations that require values to be set,
we need to be able to pass those values somehow. Normally this is done
via a tailoring... so let's add a directory where we can do that. The
e2e framework needs to detect this directory and use the tailoring
insead when needed.


Vincent056 (approved):
Thanks for the PR, it /lgtm

PR #7630, opened 2021-09-23
RHCOS4: Mark wireless_disable_interfaces as NOT-APPLICABLE in e2e tests

JAORMX:
In our standard AWS-based E2E tests, the wireless interface is not
present, and so this rule will return NOT-APPLICABLE. The rule is still
relevant, but let's fix the test.


Vincent056 (approved):

marcusburghardt (approved):
LGTM. Thanks @JAORMX

PR #7631, opened 2021-09-23
ntp/chrony remove server remediations and test scenarios

mildas:
#### Description:
**chronyd_specify_remote_server**
Update platform to `multi_platform_rhel` and remove unnecessary empty lines from tests

**ntpd_specify_remote_server**
Add remediations and test scenarios that are basically same as for `chronyd_specify_remote_server`. Remediations use same remote server variable as chronyd - `var_multiple_time_servers`

### DIFF
--- a/linux_os/guide/services/ntp/chronyd_specify_remote_server/tests/correct.pass.sh
+++ b/linux_os/guide/services/ntp/chronyd_specify_remote_server/tests/correct.pass.sh
@@ -1,7 +1,5 @@
 #!/bin/bash
 # packages = chrony
-# platform = multi_platform_fedora,Red Hat Enterprise Linux 8
-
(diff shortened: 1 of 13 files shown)


mildas:
Updated ntpd test scenarios platform to RHEL 7, because `ntp` package is not available in RHEL8 and Fedora 33.

jan-cerny:
> Updated ntpd test scenarios platform to RHEL 7, because ntp package is not available in RHEL8 and Fedora 33.

If the `ntp` package isn't available on some systems, then we most likely should limit the platform of rule `ntpd_specify_remote_server` in `rule.yml` to only the systems where this package is available.

jan-cerny (approved):

PR #7632, opened 2021-09-23
service_enabled test scenarios templates

mildas:
#### Description:
Basically same test scenario templates as for [service_disabled](https://github.com/ComplianceAsCode/content/tree/master/shared/templates/service_disabled/tests) but with inverse result expected.

### DIFF
--- a/shared/templates/service_enabled/tests/service_disabled.fail.sh
+++ b/shared/templates/service_enabled/tests/service_disabled.fail.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+# packages = {{{ PACKAGENAME }}}
+
+systemctl stop {{{ DAEMONNAME }}}
+systemctl disable {{{ DAEMONNAME }}}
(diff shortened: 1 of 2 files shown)


shared/templates/service_enabled/tests/service_disabled.fail.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+# packages = {{{ PACKAGENAME }}}
+
+systemctl stop {{{ DAEMONNAME }}}
+systemctl disable {{{ DAEMONNAME }}}
+{{% if MASK_SERVICE %}}
```
jan-cerny, line 6: Unlike the  `service_disabled` template, the `service_enabled`  template doesn't use the `MASK_SERVICE` parameter.

jan-cerny (approved):
Thanks @matusmarhefka for a quick fix.

PR #7634, opened 2021-09-23
Add test scenarios specific for CIS

jan-cerny:
#### Description:
This commit adds test scenarios specific for the RHEL7 CIS Level 2
profile for rules accounts_password_pam_pwhistory_remember_system_auth
and accounts_password_pam_pwhistory_remember_password_auth. The test
scenarios were copied from existing correct value pass scenario
and the values have been set according to item 5.4.4 in
controls/cis_rhel7.yml.

#### Rationale:

Increase test coverage of CIS profile.


vojtapolasek (commented):
Thank you for well commented tests.
See comments.
Actually @matusmarhefka  doers it make sense to add remaining CIS profiles? Does it make sense when testing for example in profile mode in SSG test suite?

matusmarhefka:
> Thank you for well commented tests.
> See comments.
> Actually @matusmarhefka doers it make sense to add remaining CIS profiles? Does it make sense when testing for example in profile mode in SSG test suite?

Test scenarios are only executed in rule/combined mode. I think in this case it is enough to have just `cis` in profiles metadata, these rules are the same in all cis profiles so I don't see a value in running them 4 times and testing the same thing.

matusmarhefka (approved):

PR #7635, opened 2021-09-23
Add Variable-based CPE and UFW rules for Ubuntu 20.04 STIG

dodys:
#### Description:

- This is a long PR as it adds support for variable-based CPEs and as an example I've included the UFW rules which make use of it.
- The variable-based CPEs commits were mainly developed by @cipherboy. I will let him talk about it, if possible.
- Let me know if you think I should break this into multiple PRs instead. I also have a few commits that adds the platform bits to iptables rules and others, that I didn't include here but I can certainly add in a separate PR.

### DIFF
--- a/ssg/build_remediations.py
+++ b/ssg/build_remediations.py
@@ -307,7 +307,7 @@ class BashRemediation(Remediation):
             assert cpe.value
 
             if cpe.negated:
-                return '[ "${0}" != "{1}" ]'.format(cpe.variable, cpe.value)
+                return '{{ [ -n "${0}" ] && [ "${0}" != "{1}" ] ; }}'.format(cpe.variable, cpe.value)
             return '[ "${0}" == "{1}" ]'.format(cpe.variable, cpe.value)
         elif platform is not None:
             # Assume any other platform is a Package CPE
@@ -572,7 +572,7 @@ class AnsibleRemediation(Remediation):
             assert cpe.value
 
             if cpe.negated:
-                return '{0} != "{1}"'.format(cpe.variable, cpe.value)
+                return '{0} and {0} != "{1}"'.format(cpe.variable, cpe.value)
             return '{0} == {1}'.format(cpe.variable, cpe.value)
         elif platform is not None:
             # Assume any other platform is a Package CPE
--- a/ssg/build_remediations.py
+++ b/ssg/build_remediations.py
@@ -334,7 +334,9 @@ class BashRemediation(Remediation):
 
             cpe = self.local_env_yaml["product_cpes"].get_cpe(platform)
             assert cpe.variable
-            return "populate {0}".format(cpe.variable)
+            var_include = ". /usr/share/scap-security-guide/remediation_functions\n"
+            var_include += "populate {0}".format(cpe.variable)
+            return var_include
 
 
 class AnsibleRemediation(Remediation):
--- a/ssg/build_remediations.py
+++ b/ssg/build_remediations.py
@@ -335,8 +335,7 @@ class BashRemediation(Remediation):
 
             cpe = self.local_env_yaml["product_cpes"].get_cpe(platform)
             assert cpe.variable
-            var_include = ". /usr/share/scap-security-guide/remediation_functions\n"
-            var_include += "populate {0}".format(cpe.variable)
+            var_include = "{0}='(bash-populate {0})'".format(cpe.variable)
             return var_include
 
 
--- a/ssg/build_remediations.py
+++ b/ssg/build_remediations.py
@@ -247,14 +247,27 @@ class BashRemediation(Remediation):
         rule_specific_conditionals = [
             cpe_platforms[p].to_bash_conditional()
             for p in rule_specific_cpe_platform_names]
+        inherited_variables = [
+            self.generate_platform_variables(p) for p in inherited_cpe_platform_names]
+        rule_specific_variables = [
+            self.generate_platform_variables(p) for p in rule_specific_cpe_platform_names]
         # remove potential "None" from lists
         inherited_conditionals = sorted([
             p for p in inherited_conditionals if p != ''])
         rule_specific_conditionals = sorted([
             p for p in rule_specific_conditionals if p != ''])
+        inherited_variables = sorted([p for p in inherited_variables if p is not None])
+        rule_specific_variables = sorted([p for p in rule_specific_variables if p is not None])
 
         if inherited_conditionals or rule_specific_conditionals:
             wrapped_fix_text = ["# Remediation is applicable only in certain platforms"]
+            # By definition, we can only have variables if we have conditionals;
+            # thus we only need to consider this inside the above if.
+            variables = set(inherited_variables + rule_specific_variables)
+            if variables:
+                # Unlike the ansible remediation below, we always load these
+                # variables as they might not be populated in time otherwise.
+                wrapped_fix_text.extend(sorted(variables))
 
             all_conditions = ""
             if inherited_conditionals:
@@ -280,6 +293,48 @@ class BashRemediation(Remediation):
             result = remediation(contents="\n".join(wrapped_fix_text), config=result.config)
 
         return result
+
+    def generate_platform_conditional(self, platform):
+        if platform == "machine":
+            # Based on check installed_env_is_a_container
+            return '[ ! -f /.dockerenv ] && [ ! -f /run/.containerenv ]'
+        elif platform.startswith("var_") or platform.startswith("not_var_"):
+            if not self.local_env_yaml:
+                return None
+
+            cpe = self.local_env_yaml["product_cpes"].get_cpe(platform)
+            assert cpe.variable
+            assert cpe.value
+
+            if cpe.negated:
+                return '[ "${0}" != "{1}" ]'.format(cpe.variable, cpe.value)
+            return '[ "${0}" == "{1}" ]'.format(cpe.variable, cpe.value)
+        elif platform is not None:
+            # Assume any other platform is a Package CPE
+
+            # Some package names are different from the platform names
+            if platform in self.local_env_yaml["platform_package_overrides"]:
+                platform = self.local_env_yaml["platform_package_overrides"].get(platform)
+
+                # Workaround for platforms that are not Package CPEs
+                # Skip platforms that are not about packages installed
+                # These should be handled in the remediation itself
+                if not platform:
+                    return
+
+            # Adjust package check command according to the pkg_manager
+            pkg_manager = self.local_env_yaml["pkg_manager"]
+            pkg_check_command = PKG_MANAGER_TO_PACKAGE_CHECK_COMMAND[pkg_manager]
+            return pkg_check_command.format(platform)
+
+    def generate_platform_variables(self, platform):
+        if platform.startswith("var_") or platform.startswith("not_var_"):
+            if not self.local_env_yaml:
+                return None
+
+            cpe = self.local_env_yaml["product_cpes"].get_cpe(platform)
+            assert cpe.variable
+            return "populate {0}".format(cpe.variable)
 
 
 class AnsibleRemediation(Remediation):
@@ -409,11 +464,17 @@ class AnsibleRemediation(Remediation):
         rule_specific_conditionals = [
             cpe_platforms[p].to_ansible_conditional()
             for p in rule_specific_cpe_platform_names]
+        inherited_variables = [
+            self.generate_platform_variables(p) for p in inherited_cpe_platform_names]
+        rule_specific_variables = [
+            self.generate_platform_variables(p) for p in rule_specific_cpe_platform_names]
         # remove potential "None" from lists
         inherited_conditionals = sorted([
             p for p in inherited_conditionals if p != ''])
         rule_specific_conditionals = sorted([
             p for p in rule_specific_conditionals if p != ''])
+        inherited_variables = sorted([p for p in inherited_variables if p is not None])
+        rule_specific_variables = sorted([p for p in rule_specific_variables if p is not None])
 
         # remove conditionals related to package CPEs if the updated task
         # collects package facts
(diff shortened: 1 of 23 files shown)


dodys:
rebased to fix conflict in shared/macros-highlevel.jinja

cipherboy:
Well, happy to see this upstreamed, thanks @dodys! :-) Not really sure where to start any more. :D

Variable based pseudo-CPEs was a solution we came upon early in the development of Canonical's internal CaC port. One feature we wanted was the ability to use a XCCDF variable to switch between one of various choices, but for sure audit against it. 

That is, compared to the real CPE system, even if the package is not installed (or could reasonably co-exist), we still want to generate pass/fail messages for the user's selection -- and preferably generate not-applicable (really: pass) messages for everything else. Note that not-applicable isn't possible due to a certain long-standing OVAL bug that still hasn't gotten solved. This further has the consequence that, if a XCCDF variable pseudo-CPE selection does NOT align with the packages on the system, the remediation of the benchmark would result in following user's desires, though it might take two scan+remediate passes in order to do so (given that CPE and pseudo-CPE could co-exist in the rule).

That is, if UFW is selected in the XCCDF variable but iptables was present, and given that all rules included both CPEs and pseudo-CPEs, the iptables would pass (see note below -- less than ideal but given our constraints) and all other would return not-applicable (except the `package_ufw_installed` and `package_iptables_removed`). On the first remediation, this would remove iptables and install UFW. On the next scan, iptables would return not-applicable (since iptables-persistent is no longer installed) and then UFW rules would return fail (since presumably they don't comply with the benchmark) -- and on this remediation, now the remediate-able rules would then finally be remediated.

Since the CPE evaluation is independent of XCCDF variables (which is to say, CPE evaluation has no access to XCCDF variables), we end up creating a "pseudo-CPE": shared OVAL definitions placed in the main OVAL content. Each rule with a pseudo-CPE selector then has its OVAL modified to essentially include a block like so:

```xml
<criteria operator="OR">
   <criterion id="not_variable">
   <criteria operator="AND">
       <criterion id="variable">
       ....existing criteria/criterion....
   </criteria>
</criteria> 
```

Ideally the first criterion would result in a hard not-applicable, but due to https://github.com/OVAL-Community/OVAL/issues/24, this isn't possible. Thus, the first criterion results in a "pass". This gets put into both bespoke and templated OVALs. 

I think we ended up using pseudo-CPEs in a couple of places internally:

 - Firewall selection (ufw/iptables/nftables)
 - Timesync selection (chrony/ntpd/systemd-timesyncd)
 - One or two other places I'm forgetting.

Notably, in both of these categories, two or more of the selections could co-exist peacefully on the system (firewalld and ufw are both front-ends to either iptables/nftables and thus definitely could co-exist with them --- or systemd-tymesyncd could be installed and inactive and co-exist with either of chrony/ntpd). Without XCCDF variable selector, alignment with the benchmarks (which, as of time of creation, audited firewall frontends rather than backends) is harder, if we want to preserve that users still need to make an explicit choice.

Finally, note that this support only applies to the OVAL. Since SCE could in theory be any number of languages, and it is easy enough to provision+read a variable and return not-applicable manually, SCE does not have automatic pseudo-variable CPE support. IMO, if we restrict SCE to purely bash, this could be useful, but if we wish to support any other language (e.g., Python), it grows to become too complex to do automatically. 

Hope that helps to provide some context on the design and requirements for this system.

jan-cerny:
Thanks for a detailed explanation, I think the feature can be a great improvement.

Currently, @vojtapolasek and @yuumasato are working on some changes in CPEs, so it might be a good idea to consult this with them.

@vojtapolasek or @yuumasato What are your thoughts on this proposal?

ssg/build_remediations.py
```diff
@@ -304,6 +329,17 @@ def generate_platform_conditional(self, platform):
...
             return pkg_check_command.format(platform)
 
+    def generate_platform_variables(self, platform):
+        if platform.startswith("var_") or platform.startswith("not_var_"):
+            if not self.local_env_yaml:
+                return None
+
+            cpe = self.local_env_yaml["product_cpes"].get_cpe(platform)
+            assert cpe.variable
+            var_include = ". /usr/share/scap-security-guide/remediation_functions\n"
```
jan-cerny, line 339: This pseudo-include is something that @j-ode has removed recently and we don't want to introduce it again.

I think that you don't need it anymore and you can replace `populate` with the `bash_instantiate_variables` Jinja macro.

dodys: thanks! I added a commit addressing this.

yuumasato:
> One feature we wanted was the ability to use a XCCDF variable to switch between one of various choices, but for sure audit against it.

With `but for sure audit against it` you mean to still evaluate the rules for the choices not made?

 
> That is, compared to the real CPE system, even if the package is not installed (or could reasonably co-exist), we still want to generate pass/fail messages for the user's selection -- and preferably generate not-applicable (really: pass) messages for everything else. 

I guess this issue is mostly caused because rules that configure a certain package have the CPE for applicability of said certain packages.
I guess that is why in RHEL8 CIS the rules for package install are not automated.
 
> Each rule with a pseudo-CPE selector then has its OVAL modified to essentially include a block like so:
> 
> ```
> <criteria operator="OR">
>    <criterion id="not_variable">
>    <criteria operator="AND">
>        <criterion id="variable">
>        ....existing criteria/criterion....
>    </criteria>
> </criteria> 
> ```

I did a local build and none of the ufw checks had the above construct, but the remediations had the check for  variable=package.
 
> Notably, in both of these categories, two or more of the selections could co-exist peacefully on the system (firewalld and ufw are both front-ends to either iptables/nftables and thus definitely could co-exist with them --- or systemd-tymesyncd could be installed and inactive and co-exist with either of chrony/ntpd). Without XCCDF variable selector, alignment with the benchmarks (which, as of time of creation, audited firewall frontends rather than backends) is harder, if we want to preserve that users still need to make an explicit choice.

Not sure I understand, could you clarify why is it hard to align the benchmark?

I guess the XCCDF way of making a choice is via profile tailorings, and actually making a choice via XCCDF variable will in fact require a tailoring.
Maybe the difficulty is knowing witch rules need to be selected for each choice? 

It looks like the variable-based CPE is trying to generalize the approach that some SSHD rules have with regards to "system install status" and "profile intention" (or user intention).
For example: [sshd_use_approved_macs_ordered_stig](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_macs_ordered_stig/oval/shared.xml#L6-L22)
Together with the `sshd_*_or_unset.xml` in `shared/checks/oval`

These "sshd requires" mechanism were implemented by me a long time ago based on some quirky requirements (that may actually no have been true, meaning they can probably be removed.), they are a bit convoluted and confusing I admit. But it seems to me that the proposed variable-based CPEs can get quite complex and confusing as well.

I kinda got confused by "CPE" in the naming. I understand the proposed mechanism as a way to bundle up a few rules in a choice mechanism, but really determining applicability (validity or relevancy) of the rules.

I see the difficulty in making available "choices" aligned with the policy, and see value in trying to make it easier for users to pick those choices. But I have a slight feeling this is adding a bit more complexity than necessary.
Will think on it a bit more.

yuumasato:
> Currently, @vojtapolasek and @yuumasato are working on some changes in CPEs, so it might be a good idea to consult this with them.

I don't think this proposal affects the CPE Applicability being worked on by Vojta.
The CPE applicability is just a more powerful way to describe the relevancy (not to say applicability again) of the rules in a "static" way. The applicability language is not intended to allow for choices in a profile tailoring

cipherboy:
Many thanks for the thoughtful comments Watson!

@yuumasato said:
> @cipherboy said:
>
> > One feature we wanted was the ability to use a XCCDF variable to switch between one of various choices, but for sure audit against it.
> 
> With `but for sure audit against it` you mean to still evaluate the rules for the choices not made?

Close. Consider these rules (relevant for a benchmark):

 - package_iptables_installed
 - package_iptables_removed
 - iptables_default_deny (cpe: iptables)
 - package_ufw_installed
 - package_ufw_removed
 - ufw_default_deny (cpe: ufw)

In a CPE-only scenario, we'd have to make a choice here as to which rules to select in the profile. Let's assume the profile favors `iptables` and that's the package installed. Then there's two scenarios:

1. sysadmin wants `iptables` installed. Easy, gets audited, expected.
2. sysamdin wants `ufw` installed. Since `iptables` is selected and `ufw` isn't, the admin has to tailor the entire profile to **unselect** the `iptables` rules (ok, easier) and to select the `ufw` rules (which is harder! --- notably, the sysadmin has to understand the _benchmark_ behind the profile and find all relevant `ufw` rules among the set of _all_ `ufw` rules). Now the sysadmin must do a two-pass "audit+remediate" -- one to install `ufw` and the other to actually use the rules (due to CPE -- this is fine and expected). 

Now consider the case where we use variable-based pseudo-CPEs. Here, we add all rules relevant to the benchmark to the profile. That means all of the rules listed above. Let's again assume `iptables` is the method selected in the XCCDF variable.

1. sysadmin wants `iptables` installed. Easy, gets audited, expected. For the `ufw`-based rules, they all report pass (due to the bug mentioned above -- unless they use SCE, in which case they can report not applicable --- or if they use real CPE, then it might also be not applicable).
2. sysadmin wants `ufw` installed. Since `iptables` is selected in the XCCDF variable, the admin merely needs to change the variable to `ufw` instead. Now the sysadmin must do the same two-pass "audit+remediate". For `iptables`, we see pass/not-applicable, much like `ufw` above.

Notably, what this means is that a the _tailoring_ merely contains _organization-specific_ policy -- and NOT benchmark+organization-specific policy (in the case of CPE-only -- all `ufw` rules must be enabled aligning with the benchmark and then rules not desired by the organization must be removed). This means if the benchmark changes, the organization must manually pull in additional rules.

This doesn't lose any of the ability for CaC to be opinionated about auditing/remediation of the benchmark (it can still favor whatever approach/tools it desires), and CPE is still present and IMO, useful.

> > That is, compared to the real CPE system, even if the package is not installed (or could reasonably co-exist), we still want to generate pass/fail messages for the user's selection -- and preferably generate not-applicable (really: pass) messages for everything else.
> 
> I guess this issue is mostly caused because rules that configure a certain package have the CPE for applicability of said certain packages. I guess that is why in RHEL8 CIS the rules for package install are not automated.

Right, and with pseudo-variable based CPEs, we can now automate them and they only get executed when this variable says so. :-)

> 
> > Each rule with a pseudo-CPE selector then has its OVAL modified to essentially include a block like so:
> > ```
> > <criteria operator="OR">
> >    <criterion id="not_variable">
> >    <criteria operator="AND">
> >        <criterion id="variable">
> >        ....existing criteria/criterion....
> >    </criteria>
> > </criteria> 
> > ```
> 
> I did a local build and none of the ufw checks had the above construct, but the remediations had the check for variable=package.

Right, in the remediation it should all be `firewall_choice == "ufw"`. Was that not the case?

> 
> > Notably, in both of these categories, two or more of the selections could co-exist peacefully on the system (firewalld and ufw are both front-ends to either iptables/nftables and thus definitely could co-exist with them --- or systemd-tymesyncd could be installed and inactive and co-exist with either of chrony/ntpd). Without XCCDF variable selector, alignment with the benchmarks (which, as of time of creation, audited firewall frontends rather than backends) is harder, if we want to preserve that users still need to make an explicit choice.
> 
> Not sure I understand, could you clarify why is it hard to align the benchmark?
> 
> I guess the XCCDF way of making a choice is via profile tailorings, and actually making a choice via XCCDF variable will in fact require a tailoring. Maybe the difficulty is knowing witch rules need to be selected for each choice?

Exactly. When you build CaC, you'd get dozens of rules for a single firewall. CIS, STIG, ... everything all creates slightly different firewall rules and gets dumped into the datastream. When the sysadmin wants to change from the default firewalling agent to another (`iptables`->`ufw` in my above examples) -- they must not only be an expert in their organization's preferences/requirements -- AND in the benchmark itself -- BUT also in CaC to know exactly which rules to select that are actually relevant to the benchmark and not the other benchmarks so the tailoring can be created correctly.

Also, this is still a profile tailoring: it just changes it from a choice of N benchmark required rules + K organization policy rules, to a single variable change + K organization policy rules -- meaning the intersection/interaction of the N+K rules isn't muddied.

> 
> It looks like the variable-based CPE is trying to generalize the approach that some SSHD rules have with regards to "system install status" and "profile intention" (or user intention). For example: [sshd_use_approved_macs_ordered_stig](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/ssh/ssh_server/sshd_use_approved_macs_ordered_stig/oval/shared.xml#L6-L22) Together with the `sshd_*_or_unset.xml` in `shared/checks/oval`
> 
> These "sshd requires" mechanism were implemented by me a long time ago based on some quirky requirements (that may actually no have been true, meaning they can probably be removed.), they are a bit convoluted and confusing I admit. But it seems to me that the proposed variable-based CPEs can get quite complex and confusing as well.
> 
> I kinda got confused by "CPE" in the naming. I understand the proposed mechanism as a way to bundle up a few rules in a choice mechanism, but really determining applicability (validity or relevancy) of the rules.

Yeah, the name "pseudo-CPE" came about because originally, we were going to have the rules _not_ selected (by the variable_ return "not applicable" -- just like CPE would've if the package wasn't installed. It was more of mimicing the functionality of CPE (from a user perspective) than the _implementation_ from a benchmark creator's perspective. 

However, OVAL doesn't allow that right now. :/

Other name suggestions welcome :-) 

> 
> I see the difficulty in making available "choices" aligned with the policy, and see value in trying to make it easier for users to pick those choices. But I have a slight feeling this is adding a bit more complexity than necessary. Will think on it a bit more.

Yeah, if you know of another way of doing this, I'd be interested to hear it. We discussed it in broad terms on a face to face call, but at the time nobody came up with any alternative approaches and this is kinda the thing that worked.

yuumasato:
@cipherboy Just a short technical comment

> > > Each rule with a pseudo-CPE selector then has its OVAL modified to essentially include a block like so:
> > > ```
> > > <criteria operator="OR">
> > >    <criterion id="not_variable">
> > >    <criteria operator="AND">
> > >        <criterion id="variable">
> > >        ....existing criteria/criterion....
> > >    </criteria>
> > > </criteria> 
> > > ```
> > 
> > 
> > I did a local build and none of the ufw checks had the above construct, but the remediations had the check for variable=package.
> 
> Right, in the remediation it should all be `firewall_choice == "ufw"`. Was that not the case?

Yes, the remediation has the check for `var_firewall_package, but the OVAL doesn't get the extra criteria from [update_with_var_platforms()](https://github.com/ComplianceAsCode/content/pull/7635/files#diff-13820b5e497e3e55e121deb90bb035d65077f7f54487218e5583769b391f4501R281)

dodys:
@yuumasato is this build failure after my rebase related to PR #7808, that you mentioned about SCE checks?
or is it something else?

matejak:
Rebasing against current master may help.

dodys:
> Rebasing against current master may help.

After rebase it is still failing. 
It seems to be the same error I've reported here:
https://github.com/mildas/content-test-filtering/issues/24

ggbecker:
> > Rebasing against current master may help.
> 
> After rebase it is still failing. It seems to be the same error I've reported here: [mildas/content-test-filtering#24](https://github.com/mildas/content-test-filtering/issues/24)

They are similar but the cause is a little bit different. I've reported https://github.com/mildas/content-test-filtering/issues/27 so @mildas can take a look. It should be simple to fix.

dodys:
> > > Rebasing against current master may help.
> > 
> > 
> > After rebase it is still failing. It seems to be the same error I've reported here: [mildas/content-test-filtering#24](https://github.com/mildas/content-test-filtering/issues/24)
> 
> They are similar but the cause is a little bit different. I've reported [mildas/content-test-filtering#27](https://github.com/mildas/content-test-filtering/issues/27) so @mildas can take a look. It should be simple to fix.

thanks!

PR #7639, opened 2021-09-23
Updated style guide to define test scripts header

marcusburghardt:
#### Description:

This is not a real problem, but since we have a style guide, it would be good to have this defined.

#### Rationale:

Currently our test scripts are not consistent. We have at least 3 cases:
1 - Test scripts without unnecessary lines (suggestion in this PR), like in this example:
```
#!/bin/bash

<code here>
```
2 - Test scripts with an extra "#" but without any header parameter, like in this example:
```
#!/bin/bash
#

<code here>
```
3 - Test scripts with an extra "#" and header parameters, like in this example:
```
#!/bin/bash
#
# packages = audit

<code here>
```
The suggestion for the style guide is the have the simplest and shortest option, like in case 1.


jan-cerny (approved):
Great update!

PR #7641, opened 2021-09-23
Implementing test ssh_set_max_sessions for rhel7 profiles

Achilleasein:
#### Description:

- _Implementing the test sshd_set_max_sessions for rhel7 profiles, specifically this time for cis and in future for any other profile._

### DIFF
--- a/linux_os/guide/services/ssh/ssh_server/sshd_set_max_sessions/tests/rhel7_correct_value.pass.sh
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_set_max_sessions/tests/rhel7_correct_value.pass.sh
@@ -1,0 +1,9 @@
+# platform = multi_platform_all
+# reboot = false
+# strategy = configure
+# complexity = low
+# disruption = low
+
+{{{ bash_instantiate_variables("var_sshd_max_sessions") }}}
+
+{{{ bash_sshd_config_set(parameter="MaxSessions", value="$var_sshd_max_sessions") }}}
--- a/linux_os/guide/services/ssh/ssh_server/sshd_set_max_sessions/tests/rhel7_correct_value.pass.sh
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_set_max_sessions/tests/rhel7_correct_value.pass.sh
@@ -1,5 +1,3 @@
 # platform = Red Hat Enterprise Linux 7
 
-{{{ bash_instantiate_variables("var_sshd_max_sessions") }}}
-
-{{{ bash_sshd_config_set(parameter="MaxSessions", value="$var_sshd_max_sessions") }}}
+{{{ bash_sshd_config_set(parameter="MaxSessions", value="$10") }}}
--- a/linux_os/guide/services/ssh/ssh_server/sshd_set_max_sessions/tests/rhel7_correct_value.pass.sh
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_set_max_sessions/tests/rhel7_correct_value.pass.sh
@@ -1,3 +1,3 @@
 # platform = Red Hat Enterprise Linux 7
 
-{{{ bash_sshd_config_set(parameter="MaxSessions", value="$10") }}}
+{{{ bash_sshd_config_set(parameter="MaxSessions", value="10") }}}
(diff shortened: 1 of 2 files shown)


Achilleasein:
It looks like there are tests by watson more than a year ago, should they be removed? Looks like we have templated the tests.

linux_os/guide/services/ssh/ssh_server/sshd_set_max_sessions/tests/rhel7_correct_value.pass.sh
```diff
@@ -0,0 +1,9 @@
+# platform = multi_platform_all
```
jan-cerny, line 1: I can see that it has `rhel7` in the file name so you aren't consistent here. Moreover, `multi_platform_all` is the default value so we don't put it into our test scenarios.

linux_os/guide/services/ssh/ssh_server/sshd_set_max_sessions/tests/rhel7_correct_value.pass.sh
```diff
@@ -0,0 +1,9 @@
...
+# reboot = false
+# strategy = configure
+# complexity = low
+# disruption = low
```
jan-cerny, line 5: We don't use `reboot`, `strategy`, `complexity` and `disruption` keys in test scenarios. See https://complianceascode.readthedocs.io/en/latest/tests/README.html#scenarios-format for description of keys that can be used.

linux_os/guide/services/ssh/ssh_server/sshd_set_max_sessions/tests/rhel7_correct_value.pass.sh
```diff
@@ -0,0 +1,9 @@
...
+
+{{{ bash_instantiate_variables("var_sshd_max_sessions") }}}
```
jan-cerny, line 7: This macro can't be used in test scenarios.

jan-cerny (changes requested):
@Achilleasein 

>It looks like there are tests by watson more than a year ago, should they be removed?

These existing test scenarios are both for RHEL 8 CIS profile (see their headers). The difference between RHEL 7 and RHEL 8 CIS profiles is that RHEL 7 allows 10 sessions whereas RHEL 8 allows only 4 sessions. See the `var_sshd_max_sessions` settings in respective control files. The existing test scenarios look valid, they set the correct value and they also pass on a RHEL8 VM.

>Looks like we have templated the tests.

Tests might be templated only if the rule is templated. You can recognize templated rules easily, they have `template:` key in their `rule.yml`. The rule `ssh_set_max_sessions` doesn't have this key in its `rule.yml`, therefore the rule isn't templated, therefore the templated test scenarios don't apply here. 

Your test scenario doesn't pass for me on a RHEL 7 VM.  Can you take a look into this problem?

```
python3 tests/test_suite.py rule --no-reports --libvirt qemu:///system ssgts_rhel7 sshd_set_max_sessions
INFO - 1 - scap_org.open-scap_cref_--home--jcerny--work--git--scap-security-guide--build--ssg-rhel7-pcidss-xccdf-1.2.xml
INFO - Logging into /home/jcerny/work/git/scap-security-guide/logs/rule-custom-2021-09-24-0844/test_suite.log
WARNING - Script correct_value.pass.sh is not applicable on given platform
WARNING - Script wrong_value.fail.sh is not applicable on given platform
INFO - xccdf_org.ssgproject.content_rule_sshd_set_max_sessions
ERROR - Script rhel7_correct_value.pass.sh using profile (all) found issue:
ERROR - The initial scan failed for rule 'xccdf_org.ssgproject.content_rule_sshd_set_max_sessions'.
```

linux_os/guide/services/ssh/ssh_server/sshd_set_max_sessions/tests/rhel7_correct_value.pass.sh
```diff
@@ -0,0 +1,3 @@
+# platform = Red Hat Enterprise Linux 7
+
+{{{ bash_sshd_config_set(parameter="MaxSessions", value="$10") }}}
```
matusmarhefka, line 3: Typo, `$10` should be `10` here.

matusmarhefka (approved):

PR #7642, opened 2021-09-24
Add a fail test for accounts_password_all_shadowed

jan-cerny:
#### Description:
This rule missed a fail test scenario. We will add a fake user
with a random string different than x or asterisk in the second
field in /etc/passwd.

#### Rationale:

This PR will improve the test coverage of RHEL 7 CIS profiles.

### DIFF
--- a/linux_os/guide/system/accounts/accounts-restrictions/password_storage/accounts_password_all_shadowed/oval/shared.xml
+++ b/linux_os/guide/system/accounts/accounts-restrictions/password_storage/accounts_password_all_shadowed/oval/shared.xml
@@ -13,6 +13,6 @@
     <unix:username operation="pattern match">.*</unix:username>
   </unix:password_object>
   <unix:password_state id="state_accounts_password_all_shadowed" version="1">
-    <unix:password operation="pattern match">x|\*</unix:password>
+    <unix:password operation="pattern match">^[x*]$</unix:password>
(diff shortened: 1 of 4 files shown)


marcusburghardt (changes requested):
@jan-cerny , the test is fine. However, we should also update the OVAL here. For example, this line would also match as valid:
xtestuser:0fcf3792f5fffc4fcd1801fab37e0470c66d46cb0f01:8000:8000:testuser:/home/testuser:/bin/bash

In this case, the password is not shadowed but the test would pass because the username contains a x.

marcusburghardt:
> @jan-cerny , the test is fine. However, we should also update the OVAL here. For example, this line would also match as valid:
> xtestuser:0fcf3792f5fffc4fcd1801fab37e0470c66d46cb0f01:8000:8000:testuser:/home/testuser:/bin/bash
> 
> In this case, the password is not shadowed but the test would pass because the username contains a x.

This should solve:
`:x:|:\*:`

jan-cerny:
@marcusburghardt Amazing! I have changed the OVAL and added another test scenario. But I think the regex needs to be different than you propose because the OVAL `password` element contains only the password value , ie. without colons.

marcusburghardt (approved):
Great. Thanks @jan-cerny

PR #7645, opened 2021-09-24
Add sphinx missing dependency in the developer guide

ggbecker:
#### Description:

- Add sphinx missing dependency in the developer guide

#### Rationale:

- Adds the package that contains `sphinx-build` for example.


jan-cerny (approved):
@ggbecker Nice catch! Thanks!

PR #7647, opened 2021-09-24
Use --duplicate-templates in SSGTS integration with Github

ggbecker:
This option will ensure that all templated rules are properly
tested. Otherwise only one rule for such template would be tested.

Additionally its expands the test_rule_in_container.sh script to
support additional SSGTS parameters using env variable.

#### Description:

- Use --duplicate-templates in SSGTS integration with Github.
  - This option will ensure that all templated rules are properly
tested. Otherwise only one rule for such template would be tested.
  - Additionally its expands the test_rule_in_container.sh script to
support additional SSGTS parameters using env variable.


ggbecker:
Screenshot from 2021-09-24 15-57-13

testing commit with changes to two package installed rules

jan-cerny (approved):
I like the idea that users will be able to pass any argument easily.

Issue #7446, opened 2021-08-23
Improve CCE expansion in remediations


jan-cerny:
#### Description of problem:
During @j-ode 's work on #7378 we found that a lot of bash remeditations use the string `@CCENUM@`.

The `@CCENUM@` is a string that is replaced at a later phase of the build process by the rule CCE number. It's done by the XSLT template `/shared/transforms/xccdf-addremediations.xslt` on lines 68 - 81. Notice that in `/build/<product>/fixes/bash` you will see the `@CCENUM@` string, but in the generated SCAP source data stream (`ssg-<product>-ds.xml`) you will see that it got replaced by the rule's CCE number. Or, it's replaced by an empty string if the rule doesn't have a CCE number. For example, in Fedora we don't assign CCEs.

I think that the mechanism isn't transparent and it was surprising that something modifies the remediation. I think that the same resul could be achieved using Jinja macros which are our preferred mechanism for text expansion.

Also, the Jinja macro `bash_replace_or_append` should be changed to not have the `cce` parameter because the macro is always called with this parameter set to `@CCENUM@`.

#### SCAP Security Guide Version:
current upstream as of 2021-08-23

#### Operating System Version:
all

#### Steps to Reproduce:

1. use `@CCENUM@` in a bash remediation script.
2. build content
3. see the generated code in `/build/fixes/<product>/bash`
4. see the generated SCAP source data stream, eg. ssg-rhel8-ds.xml
5. compare step 3 with step 4

#### Actual Results:
`@CCENUM@` is used to put the CCE number into the remediation body

#### Expected Results:
a documented Jinja macro can be used to put the CCE number into the remediation body

#### Additional Information/Debugging Steps:
no

Mab879:
#5608 might be related

jan-cerny:
Fixed in #7657

PR #7657, opened 2021-09-27
Moved CCE number expansion from XSLT to jinja macro 

j-ode:
#### Description:

- Previously when we wanted to put a rule's CCE number into the changed files during remediations, we used the '@CCENUM@' string which was then searched and replaced by a XSLT template. This was not ideal, so I created the set_cce_value jinja macro which gets the CCE number from the environment provided by the build script.
- Removed the obsolete XSLT template and removed the 'cce' argument from all bash_replace_or_append macro invocations
- Fixed BATS unit tests for bash_replace_or_append

#### Rationale:

- #7446

### DIFF
--- a/shared/macros-bash.jinja
+++ b/shared/macros-bash.jinja
@@ -1150,7 +1150,6 @@ cce="CCE"
   config_file:		Configuration file that will be modified
   key:			Configuration option to change
   value:		Value of the configuration option to change
-  cce:			The CCE identifier or '@CCENUM@' if no CCE identifier exists
   format:		The printf-like format string that will be given stripped key and value as arguments,
              so e.g. '%s=%s' will result in key=value subsitution (i.e. without spaces around =)
  
@@ -1162,16 +1161,16 @@ cce="CCE"
   Example Call(s):
  
       With default format of 'key = value':
-      {{{ bash_replace_or_append('/etc/sysctl.conf', '^kernel.randomize_va_space', '2', '@CCENUM@') }}}
+      {{{ bash_replace_or_append('/etc/sysctl.conf', '^kernel.randomize_va_space', '2') }}}
  
       With custom key/value format:
-      {{{ bash_replace_or_append('/etc/sysconfig/selinux', '^SELINUX=', 'disabled', '@CCENUM@', '%s=%s') }}}
+      {{{ bash_replace_or_append('/etc/sysconfig/selinux', '^SELINUX=', 'disabled', '%s=%s') }}}
  
       With a variable:
-      {{{ bash_replace_or_append('/etc/sysconfig/selinux', '^SELINUX=', "$var_selinux_state", '@CCENUM@', '%s=%s') }}}
+      {{{ bash_replace_or_append('/etc/sysconfig/selinux', '^SELINUX=', "$var_selinux_state", '%s=%s') }}}
  
 #}}
-{{%- macro bash_replace_or_append(config_file, key, value, cce, format='%s = %s') -%}}
+{{%- macro bash_replace_or_append(config_file, key, value, format='%s = %s') -%}}
 
 # Test if the config_file is a symbolic link. If so, use --follow-symlinks with sed.
 # Otherwise, regular sed command will do.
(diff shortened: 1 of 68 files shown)


shared/macros-bash.jinja
```diff
@@ -1165,12 +1179,8 @@ if test -L "{{{ config_file }}}"; then
     sed_command+=('--follow-symlinks')
 fi
 
-# If the cce arg is empty, CCE is not assigned.
-if [ -z "{{{ cce }}}" ]; then
-    cce="CCE"
-else
-    cce="{{{ cce }}}"
-fi
+# Set the rule CCE value; if the rule has none, it is set to 'CCE'
```
jan-cerny, line 1182: It's great that you added a comment, but when I think of it, I realize that this comment will be shown also in the generated code, so the people that will see the generated code will see something like this:

```
# Set the rule CCE value; if the rule has none, it is set to 'CCE'
cce="CCE-80790-9"
```
and if they see something like this it doesn't make sense for them because the value is simply set on a single line right away, without any `if`s.

I would remove the comment. What is your opinion?

shared/macros-bash.jinja
```diff
@@ -1126,6 +1126,21 @@ rpm --quiet -q "{{{ pkgname }}}"
...
+  #
+  # This macro gets the var cce_identifiers from the environment created by the build scripts.
+  # The cce_identifiers is a dictionary that contains either the 'cce':'CCENUM' record for the
+  # product this remediation is being built for, or it is empty, in which case the string 'CCE' is used.
+  #}}
+{{%- macro set_cce_value() -%}}
+{{% if 'cce' in cce_identifiers -%}}
+cce="{{{ cce_identifiers['cce'] }}}"
+{{%- else -%}}
+cce="CCE"
```
jan-cerny, line 1140: I have found some rules in the built  RHEL 8 data stream (ssg-rhel8-ds.xml) which have `cce="CCE"` in the remediation script even if these rules have CCEs. For example, the rule `accounts_password_pam_lcredit` has a RHEL8 CCE number assigned, but it doesn't propagate to this macro. Can it be related to the fact that this rule is templated?

shared/macros-bash.jinja
```diff
@@ -1165,12 +1179,8 @@ if test -L "{{{ config_file }}}"; then
...
 fi
 
-# If the cce arg is empty, CCE is not assigned.
-if [ -z "{{{ cce }}}" ]; then
-    cce="CCE"
-else
-    cce="{{{ cce }}}"
-fi
+# Set the rule CCE value; if the rule has none, it is set to 'CCE'
+{{{ set_cce_value() }}}
```
jan-cerny, line 1183: I think that now when we don't use the XSLT we have an opportunity to make the ` bash_replace_or_append` Jinja macro less stupid or even more intelligent.

Currently,  when the rule doesn't have any CCE number assigned, it prints `Per CCE: Set something in somefile.`. However, this feels like a weird workaround for the situtation when there is no CCE. If there is no CCE, it doesn't make sense to talk about CCEs at all. In my opinion, when a rule has no CCE we shouldn't put this this comment at all. What do you think?

shared/macros-bash.jinja
JAORMX: This can probably be replaced by a jinja comment. e.g.

```
{{# Set the rule CCE value; if the rule has none, it is set to 'CCE' #}}
```
This way, we can still inform other developers about this, and it won't appear in the generated code

ssg/templates.py
```diff
@@ -360,7 +363,7 @@ def build_all_rules(self):
                 continue
             langs_to_generate = self.get_langs_to_generate(rule)
             self.build_rule(
-                rule.id_, rule.title, rule.template, langs_to_generate, platforms=rule.platforms)
+                rule.id_, rule.title, rule.template, langs_to_generate, rule, platforms=rule.platforms)
```
jan-cerny: When I see it I can see that there are these parameters passed: `rule.id_`, `rule.title`, `rule.template`, which are all attributes of the `rule` object and at the same time the whole `rule` object is passed. I feels superflous. I would rename the new parameter of the `build_rule` to `identifiers`, changed the line 307 accordingly and here I would pass `rule.identifies` as an argument.

jan-cerny (approved):
Great job

PR #7659, opened 2021-09-27
Implementing fail scenarios for tests that read /proc/cpuinfo

Achilleasein:
#### Description:

- _Implementation of test case: bios_enable_execution_restrictions_
- _~~Implementation of test case: install_PAE_kernel_on_x86-32~~_

#### Rationale:

Properly implementing a pass scenario means we have to create grub script which is pointless since in the test machines there is state reset after every reboot.
To that end we implemented fail scenarios that create a bind of an empty file to the desired file /proc/cpuinfo and that fails the tests correctly.


matusmarhefka (approved):

linux_os/guide/system/permissions/restrictions/enable_nx/bios_enable_execution_restrictions/tests/wrong_value.fail.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+#
```
matusmarhefka, line 2: Please also add `# remediation = none` as this rule has no remediation script.

matusmarhefka (changes requested):

marcusburghardt (changes requested):
Hi @Achilleasein, the idea of using `--bind` is interesting for this complicated test case. Nice! : )

However, I would suggest you to create some content in the fake `cpuinfo` instead of an empty file. The OVAL check of this rule has an AND condition which checks if the CPU supports NX through the CPU Flags in `/proc/cpuinfo`, if there is any log message matching `protection: disabled` in `/var/log/messages` and if `noexec[0-9]*=off` is found in `/proc/cmdline`. So, we can create fail and pass cases exploring this condition and creating faking files.

Examples:
Pass case with NX flag
`flags           : fpu vme acpi tm pbe syscall nx pdpe1gb rdtscp`

Fail case without NX flag
`flags           : fpu vme acpi tm pbe syscall pdpe1gb rdtscp`

Similar approach can be used with `/proc/cmdline` and `messages|dmesg`.
Note: I am not sure about the OS behavior when these proc files are corrupted (created manually with some content). To stay in the safe line, I would suggest to copy the original one and subtly apply the desired changes with `sed`.

Achilleasein:
@marcusburghardt What you say about using sed and possible impications is correct, but the idea is that it will take some regex to actually find the flag, which could potentially change between various systems. So this was the next best solution without over engineering or spending too much time to it.
My take is to leave it as is and create a better version in the future.
But we can do some research and see if there is point on doing the longer solution.

marcusburghardt:
> @marcusburghardt What you say about using sed and possible impications is correct, but the idea is that it will take some regex to actually find the flag, which could potentially change between various systems. So this was the next best solution without over engineering or spending too much time to it. My take is to leave it as is and create a better version in the future. But we can do some research and see if there is point on doing the longer solution.

@Achilleasein, actually it looks simpler. Please, take a look in these following examples adapted from your `wrong_value.fail.sh`.
`nx_flag_not_present.pass.sh`:
```
#!/bin/bash
# platform = multi_platform_rhel,multi_platform_fedora

cp /proc/cpuinfo /tmp/cpuinfo
sed -ir 's/(^flags.*:.*)(\s+nx\s+)(.*)/\1 \3/g' /tmp/cpuinfo
mount --bind /etc/cpuinfo /proc/cpuinfo
```

`nx_flag_present_not_disabled.fail.sh`:
```
#!/bin/bash
# platform = multi_platform_rhel,multi_platform_fedora

cp /proc/cpuinfo /tmp/cpuinfo
# This will make sure the nx flag is present. For this case, it is not a problem if it appears twice.
sed -i 's/^flags.*:.*/& nx/g' /tmp/cpuinfo

# This will make sure the it was not disabled, as expected by the OVAL test
cp /proc/cmdline /tmp/cmdline
sed -r 's/\s+noexec[0-9]*=off[\s]*//g' /tmp/cmdline

mount --bind /etc/cpuinfo /proc/cpuinfo
mount --bind /etc/cmdline /proc/cmdline
```

linux_os/guide/system/permissions/restrictions/enable_nx/bios_enable_execution_restrictions/tests/wrong_value.fail.sh
```diff
@@ -0,0 +1,28 @@
...
+# platform = multi_platform_rhel,multi_platform_fedora
+#
```
marcusburghardt, line 4: You can remove these (line 2 as well) empty "comments":
https://complianceascode.readthedocs.io/en/latest/manual/developer/04_style_guide.html#tests

linux_os/guide/system/permissions/restrictions/enable_nx/bios_enable_execution_restrictions/tests/wrong_value.fail.sh
```diff
@@ -0,0 +1,28 @@
...
+# remediation = none
+
+> /tmp/cpuinfo
+> /tmp/cmdline
+
+if [ -e /tmp/cpuinfo ] && [ -e /proc/cpuinfo ]; then
+cp /proc/cpuinfo /tmp/cpuinfo
```
marcusburghardt, line 11: You could simplify these lines with something like this:
```
if [ cp /proc/cpuinfo /tmp/cpuinfo ]; then
   sed ...
```

linux_os/guide/system/permissions/restrictions/enable_nx/bios_enable_execution_restrictions/tests/wrong_value.fail.sh
```diff
@@ -0,0 +1,28 @@
...
+if grep -Fxq '^flags.*:.*/& nx' /tmp/cpuinfo ; then
```
marcusburghardt, line 12: This `if` is not necessary since the `flags` field is always present in the file /proc/cpuinfo:
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/arch/x86/include/asm/cpufeature.h

Also, this filter will always return false. I would suggest to remove it.

linux_os/guide/system/permissions/restrictions/enable_nx/bios_enable_execution_restrictions/tests/wrong_value.fail.sh
```diff
@@ -0,0 +1,28 @@
...
+sed -i 's/^flags.*:.*/& nx/g' /tmp/cpuinfo
+else
+echo 'flags = nx' >> /tmp/cpuinfo
```
marcusburghardt, line 15: Could be also removed with the simplification previously suggested.

linux_os/guide/system/permissions/restrictions/enable_nx/bios_enable_execution_restrictions/tests/wrong_value.fail.sh
```diff
@@ -0,0 +1,28 @@
...
+fi
+mount --bind /tmp/cpuinfo /proc/cpuinfo
+fi
+
+if [ -e /tmp/cmdline ] && [ -e /proc/cmdline ]; then
+cp /proc/cmdline /tmp/cmdline
+if grep -Fxq sed -r '\s+noexec[0-9]*=off[\s]*' /tmp/cmdline ; then
```
marcusburghardt, line 22: This line should not work. Could you review it, please?

marcusburghardt (changes requested):
@Achilleasein , thanks for the update. Do you intend to create a "pass" script as well?
Please take a look in the comments.

linux_os/guide/system/permissions/restrictions/enable_nx/bios_enable_execution_restrictions/tests/wrong_value.fail.sh
Achilleasein: It is to make sure just in case, either someone else had the same destructive idea as us or there has been any strange changes.

linux_os/guide/system/permissions/restrictions/enable_nx/bios_enable_execution_restrictions/tests/wrong_value.fail.sh
Achilleasein: I will check it asap.

linux_os/guide/system/permissions/restrictions/enable_nx/bios_enable_execution_restrictions/tests/wrong_value.fail.sh
Achilleasein: Just left them there for easier reading. Wanna create a styleguide?

matusmarhefka (dismissed):
Also please indent your code.

marcusburghardt (approved):
LGTM. Thanks for these test scenarios.

PR #7660, opened 2021-09-28
Rework bash remediation for accounts_password_pam_unix_remember 

teacup-on-rockingchair:
#### Description:

- Rework bash remediation for accounts_password_pam_unix_remember rule to support the case when pam config file does not exist

#### Rationale:

- Use bash template function bash_ensure_pam_module_options to setup needed options


yuumasato (changes requested):
The macro behaviour doesn't seem right and needs some fixes.

PR #7666, opened 2021-09-28
Added Build, Test on OpenSUSE Leap 15 on pull requests

Mab879:
#### Description:

This adds a build and test job for OpenSUSE on pull requests.

#### Rationale:

Due to the amount of OpenSUSE contributions recently it seems appropriate to add.

### DIFF
--- a/.github/workflows/gate.yaml
+++ b/.github/workflows/gate.yaml
@@ -5,6 +5,22 @@ on:
   pull_request:
     branches: [ 'master', 'stabilization*' ]
 jobs:
+  validate-suse:
+    name: Build, Test on OpenSUSE Leap 15 (Container)
+    runs-on: ubuntu-latest
+    container:
+      image: opensuse/leap:15
+    steps:
+      - name: Install Deps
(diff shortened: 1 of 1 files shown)


.github/workflows/gate.yaml
```diff
@@ -5,6 +5,22 @@ on:
...
+    runs-on: ubuntu-latest
+    container:
+      image: opensuse/leap:15
+    steps:
+      - name: Install Deps
+        run: zypper install -y git cmake make openscap-utils python3-PyYAML bats python3-pytest python3-pytest-cov python3-Jinja2 libxslt-tools libxml2-tools libexpat1 libexpat-devel libexpat1 expat
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Build
+        run: ./build_product chromium fedora firefox fuse6 rhcos4 rhel7 rhel8 rhel9 rhosp10 rhosp13 sle12 sle15 ubuntu2004
```
jan-cerny, line 19: I wouldn't build the other products than SLE in this job because they are already built and tested by the other jobs.

Mab879: That can be changed but this also matches what are doing on Ubuntu and I wanted to match that.

jan-cerny: OK, consistency beats everything :smiley:

jan-cerny (approved):

PR #7667, opened 2021-09-28
Added ability to slice SSGTS rule checking runs

dahaic:
Added new option --slice (with parameters N M) to the SSGTS

This will slice the number of scenarios into M slices, and
will run only Nth slice. This allows for concurrent execution over
several machines, to reduce the time required in complex scenarios.

Execution without this option utilized:
[ssgts_slice_log_whole.txt](https://github.com/ComplianceAsCode/content/files/7245062/ssgts_slice_log_whole.txt)

Execution with a 9-way split. 
[ssgts_slice_log_split.txt](https://github.com/ComplianceAsCode/content/files/7245061/ssgts_slice_log_split.txt)

For whatever reason there are many notapplicables, so the logs are not that cool, but it's outside of scope of the change. And the messages are identical.
Also - notice that together, sliced scenarios are a bit more time consuming. It's because the initial parsing of the DS takes a lot of time

Combined mode also works:
[ssgts_slice_log_combined.txt](https://github.com/ComplianceAsCode/content/files/7245104/ssgts_slice_log_combined.txt)
.


dahaic:
Specifically might be useful to @matusmarhefka

tests/test_suite.py
```diff
@@ -211,14 +220,32 @@ def parse_args():
                                  dest="scenarios_regex",
                                  default=None,
                                  help="Regular expression matching test scenarios to run")
+    parser_combined.add_argument("--slice",
```
matusmarhefka, line 223: What about using https://docs.python.org/3/library/argparse.html#metavar ?
Right now this is how the help message looks like:
```
  --slice _SLICES _SLICES
                        Allows to run only Xth slice of Y in total, to enable stable parallelization
                        of the bigger test sets.
```
which is not that nice. With added metavar (e.g. `metavar=('X', 'Y')`) it would be:
```
  --slice X Y
```
Also `Note that metavar only changes the displayed name - the name of the attribute on the parse_args() object is still determined by the dest value`.

dahaic: *Facepalm* yeah, I believe I was searching for it, but for whatever reason missed the `metavar`.. Thanks, forcepushed new version with proper help output :)

PR #7668, opened 2021-09-29
Remove leftover remediation_functions calls

dodys:
#### Description:

- Remove leftover remediation_functions calls


jan-cerny:
@mildas The CTF thinks that ubuntu is a rule.

jan-cerny (approved):
Thanks for your PR! Nice catch!

PR #7672, opened 2021-09-29
Clarify doc for check-export in SCE

mgjadoul:
#### Description:

I try to explain how to use check_export as it was not exactly what I understood when reading the doc.

#### Rationale:

- SCE specifications are not really clear about env variables generated by check export before calling SCE scripts, thus this is what I experienced by using the generated DS with oscap.


jan-cerny (approved):
Thank you very much for this improvement!

PR #7679, opened 2021-09-30
Add new rule "service_ypserv_disabled"

### LINKED ISSUE #5232: CIS 2.2.17 Ensure NIS Server is not enabled (Scored)
Need new rule. Only have package removed, not service disabled.

Mab879:
#### Description:

Add rule _service_ypserv_disabled_ for RHEL8 CIS 2.2.17.

#### Rationale:

- Fixes #5232


jan-cerny (approved):

jan-cerny:
The problem with the failing tests is weird

```
##### service_ypserv_disabled / service_enabled.fail.sh #####
ssh -o Port=46395 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null root@localhost cd /root/ssgts/service_ypserv_disabled; SHARED=/root/ssgts/shared bash -x service_enabled.fail.sh
STDERR: Warning: Permanently added '[localhost]:46395' (ECDSA) to the list of known hosts.
+ systemctl start ypserv
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down
```

Nevertheless, the test scenarios pass locally when they're run on a RHEL 8 VM and this is a templated rule so it can be only affected by mistakes in the template. I think we are safe to proceed.

PR #7680, opened 2021-10-01
Update SSGTS so it can use mount in containers

ggbecker:
#### Description:

- Update SSGTS so it can use mount in containers


ggbecker:
`--privileged` works 8dd3a151c584c8aff87029769032cdf9e33a5860

matusmarhefka (approved):

Issue #5522, opened 2020-03-29
4.2.2.3 Ensure journald is configured to write logfiles to persistent disk (Scored)


shawndwells:
Needs rule.

May benefit other profiles too.

Issue #5521, opened 2020-03-29
4.2.2.2 Ensure journald is configured to compress large log files (Scored)


shawndwells:
needs rule

other profiles may benefit from this too

Issue #5520, opened 2020-03-29
4.2.2.1 Ensure journald is configured to send logs to rsyslog (Scored)


shawndwells:
Needs rule

shawndwells:
Note other profiles will benefit from this too

PR #7682, opened 2021-10-01
Add new rules for CIS Journald Config

Mab879:
#### Description:

Add three rules for CIS Journald Config

#### Rationale:

 Fixes 
- #5520
- #5521 
- #5522


linux_os/guide/system/logging/journald/journald_storage/rule.yml
```diff
@@ -0,0 +1,43 @@
+documentation_complete: true
+
+prodtype: rhel7,rhel8,rhel9
+
+title: Ensure journald is configured to write log files to persistent disk
+
+description: |-
+    The journald system may store log files in violate memory or locally on disk.
```
jan-cerny, line 8:
```suggestion
    The journald system may store log files in volatile memory or locally on disk.
```

linux_os/guide/system/logging/journald/journald_storage/rule.yml
```diff
@@ -0,0 +1,43 @@
...
+    If the logs are only stored in violate memory they will we lost upon reboot.
```
jan-cerny, line 9:
```suggestion
    If the logs are only stored in volatile memory they will we lost upon reboot.
```

linux_os/guide/system/logging/journald/journald_compress/tests/correct_value.pass.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+# platform = multi_platform_all
+JOURNALD_CONFIG=/etc/systemd/journald.conf
```
jan-cerny, line 3: When the rule is templated we prefer templated test scenarios.

Mab879: Looks like there is not any templated tests after looking at https://github.com/ComplianceAsCode/content/tree/master/shared/templates/shell_lineinfile.

PR #7689, opened 2021-10-04
Refactor handling of scap datastructures

matejak:
This PR refactors classes in `build_yaml.py`, and adds some tiny improvements that were added in the process.

Classes become more declarative, they get the `KEYS`, `MANDATORY_KEYS`, `GENERIC_FILENAME` and `ID_LABEL` class variables that mostly control their behavior. In a greater detail:

- Extract the common boilerplate, so there is less of duplicated code, and those structures can be ported to a Python `dataclass` in the future.
- Supply those classes with methods that allow them to be dumped as yamls.
- Allow modifying the load/dump behavior by extending the appropriate methods.
- Preserve backwards compatibility.

The shorthand generated as a result of the new code is the same as the previous shorthand, which indicates that the XCCDF was not negatively affected.


matejak:
I have addressed remarks in a rebase.

PR #7700, opened 2021-10-06
UBTU-20-010300: Verify there is a script that offloads audit data and that script runs weekly

dodys:
#### Description:

- Verify there is a script that offloads audit data and that script runs weekly.
Check if there is a script in the "/etc/cron.weekly" directory that offloads audit data:

`# sudo ls /etc/cron.weekly
audit-offload`

Check if the script inside the file does offloading of audit logs to external media.
If the script file does not exist or does not offload audit logs, this is a finding.

### DIFF
--- a/linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_offload_logs/tests/correct.pass.sh
+++ b/linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_offload_logs/tests/correct.pass.sh
@@ -1,4 +1,6 @@
 #!/bin/bash
+# packages = audit
+# remediation = none
 
 mkdir -p /etc/cron.weekly/
 touch /etc/cron.weekly/audit-offload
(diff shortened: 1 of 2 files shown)


linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_offload_logs/tests/correct.pass.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+touch /etc/cron.weekly/audit-offload
```
ggbecker, line 3: please make sure that the directory exists first. In the tests there is something:
```
+ touch /etc/cron.weekly/audit-offload
touch: cannot touch '/etc/cron.weekly/audit-offload': No such file or directory
```

```suggestion
mkdir -p /etc/cron.weekly/
touch /etc/cron.weekly/audit-offload
```

dodys: Thanks, I've fixed it.

PR #7701, opened 2021-10-06
Add missing disk full action variable to Ubuntu 20.04 STIG profile

dodys:
#### Description:

- Add missing disk full action variable to Ubuntu 20.04 STIG profile


ggbecker (approved):
LGTM

PR #7703, opened 2021-10-07
OCP4: Take CIS profiles into use as a base for PCI-DSS

JAORMX:
Requirement 2.2 recommends the usage of accepted industry standards for
hardening the system. We already have the CIS profiles done which serve this
purpose. So, this tags the rules appropriately and makes the profiles inherit
from CIS.


JAORMX:
Pull-request updated, HEAD is now 216bab9c4c1a81a34ed368b962e9a157232e3d2e

PR #7714, opened 2021-10-08
Migrate Satellite 6 NIST evaluation from OpenControl to controls structure

JAORMX:
This migrates the contents of OpenControl towards the control structure.
Currently there are no rules enabled for this, but this facilitates
that.


jhrozek:
So some controls in satellite (and looking at opencontrol only in satellite) use "implementation_status: none" for organizational controls which are translated to "in progress". Do you think it's correct?

PR #7717, opened 2021-10-08
Add `utils/create_scap_delta_tailoring.py`

Mab879:
#### Description:

The `utils/create_scap_delta_tailoring.py` utility makes so that we can crate tailoring files that only enable rules that are not covered by the DISA's SCAP content for STIG profiles.

#### Rationale:

Some users need to use DISA content and also want to use our content as well,

### DIFF
--- a/utils/create_scap_delta_tailoring.py
+++ b/utils/create_scap_delta_tailoring.py
@@ -21,21 +21,17 @@ NS = {'scap': ssg.constants.datastream_namespace,
 PROFILE = 'stig'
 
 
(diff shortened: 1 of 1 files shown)


Mab879:
I think the plan was to have the tailoring file on Red Hat Access, but we could change it to run outside the repo for the end user to run.

yuumasato:
> I think the plan was to have the tailoring file on Red Hat Access, but we could change it to run outside the repo for the end user to run.

@Mab879  Do you plan to change it to run outside of the repo?

PR #7722, opened 2021-10-11
Update CONTRIBUTING.md

evgenyz:
It is been a while since any of those four people contributed, reviewed or "ensured that the commits and contributions meet the standards" in any other meaningful way. While we recognize and appreciate the value they brought to the project in the past, I think that it is time to move on and stop misleading possible future contributors about ways to participate in the project.


matejak:
Thank you for the PR!

PR #7730, opened 2021-10-12
Enforce usage of product (fedora) option when running SSGTS on Github actions

ggbecker:
#### Description:

-  Enforce usage of product (fedora) option when running SSGTS on Github actions

#### Rationale:

- When testing dpkg content on Fedora, we want to make sure that package installation uses dnf instead of apt-get.


ggbecker:
> #### Description:
> 
>     * Enforce usage of product (fedora) option when running SSGTS on Github actions
> 
> 
> #### Rationale:
> 
>     * When testing dpkg content on Fedora, we want to make sure that package installation uses dnf instead of apt-get.

The optimal fix is to use debian/ubuntu based containers to test debian/ubuntu content.

PR #4786, opened 2019-09-02
Improve template grub2 argument

vojtapolasek:
#### Description:

The templates for checking and remediating of Grub2 arguments are improved. The way of updating Grub configuration between RHEL7 and RHEL8 has changed and therefore two ways of accomplishing checking and remediation were used. The documentation and OCIL was also updated.

#### Rationale:

The work started because of BZ [https://bugzilla.redhat.com/show_bug.cgi?id=1547642](https://bugzilla.redhat.com/show_bug.cgi?id=1547642), then it was found that the reworking because of changes between RH7 and RH8 is needed.

PR #7736, opened 2021-10-13
Align the RHEL8 STIG to the RHEL9 STIG

Mab879:
#### Description:

Copied over the RHEL8 STIG to RHEL9. 

Added prodtype and CCE to support the before mentioned operation.

#### Rationale:

Preparing for RHEL9.


products/rhel9/profiles/stig.profile
```diff
@@ -1049,10 +1178,10 @@ selections:
     - sshd_disable_x11_forwarding
 
     # RHEL-08-040341
-#    - sshd_x11_use_localhost  # not supported in RHEL9 ATM
+    - sshd_x11_use_localhost
```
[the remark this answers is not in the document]
Mab879: Its in `sshd_config(5)`

marcusburghardt (changes requested):
Looks good to me. I only saw that SSGTS is reporting an error due to the `configure_firewalld_ports` rule not having a test scenario. While a `fail` scenario would broke the connection during the tests, a `pass` scenario would be valid to make sure the OVAL is working as expected at least in compliant environments. And the error would be fixed. : )

Mab879:
Tests pass on a VM.
```
$ ./test_suite.py rule --libvirt qemu:///system test_ssg_rhel84_auto --datastream ../build/ssg-rhel8-ds.xml configure_firewalld_ports
INFO - Logging into /home/mburket/code/ComplianceAsCode/content/tests/logs/rule-custom-2021-11-11-0712/test_suite.log
INFO - xccdf_org.ssgproject.content_rule_configure_firewalld_ports
INFO - Script ssh_port_enabled.pass.sh using profile (all) OK
```

marcusburghardt:
> Tests pass on a VM.
> 
> ```
> $ ./test_suite.py rule --libvirt qemu:///system test_ssg_rhel84_auto --datastream ../build/ssg-rhel8-ds.xml configure_firewalld_ports
> Setting console output to log level INFO
> INFO - The base image option has not been specified, choosing libvirt-based test environment.
> INFO - Logging into /home/mburket/code/ComplianceAsCode/content/tests/logs/rule-custom-2021-11-11-0712/test_suite.log
> INFO - xccdf_org.ssgproject.content_rule_configure_firewalld_ports
> INFO - Script ssh_port_enabled.pass.sh using profile (all) OK
> ```

Maybe we should include `platform: machine` for this rule?

Mab879:
> Maybe we should include platform: machine for this rule?

The group is already `platform: machine`.

marcusburghardt:
Hi @Mab879 , I could reproduce the error locally, testing the rule in a similar container, and the error is because the container doesn't have the `firewalld` package.

```
...
+ firewall-cmd --add-port=22/tcp
ssh_port_enabled.pass.sh: line 3: firewall-cmd: command not found
+ firewall-cmd --add-port=22/tcp --permanent
ssh_port_enabled.pass.sh: line 4: firewall-cmd: command not found 
```
To solve this in test scenarios we can take advantage of the `package` header.
https://complianceascode.readthedocs.io/en/latest/tests/README.html#scenarios-format

Here is an example:
https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/ssh/ssh_server/firewalld_sshd_port_enabled/tests/no_nic_in_ssh_zone.fail.sh#L2

Mab879:
So that package statement won't fix the issuse. I can't get that rule (`firewalld_sshd_port_enabled`) to pass in a container as either. `firewall-cmd` needs dbus.

ggbecker:
I believe these firewall rules are not meant to work in containers anyway. Our testing fakes the environment so it can try to check if check and remediation are aligned, but we cannot run functional tests. I think it's safe to waive the SSGTS tests in this case. @marcusburghardt

marcusburghardt:
Thank you for the information @ggbecker.

Although this test will fail in containers, it is working in VMs and is a good test.
Therefore, I am approving this PR.

We recognize that the error will be shown for future PRs which change this specific rule.

marcusburghardt (approved):
Great work @Mab879 . Thanks

shared/templates/grub2_bootloader_argument/ansible.template
```diff
@@ -4,32 +4,32 @@
 # complexity = medium
 # disruption = low
 
-{{% if product in ["rhel7", "ol7"] or 'ubuntu' in product %}}
-- name: check {{{ ARG_NAME }}} argument exists
+{{% if product in ["rhel7", "ol7", "rhel9"] or 'ubuntu' in product %}}
```
ggbecker, line 7: It looks like this change needs to be applied for bash/OVAL as well.

I found out that RHEL8 STIG recommends using `grubby` to set kernel parameters, I wonder if it makes sense to include rhel8 in this list: https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_8_security_technical_implementation_guide/V-230279

Even though this started as a change a long time ago because RHEL7 and RHEL8 were configured in a different way: 
https://github.com/ComplianceAsCode/content/pull/4786

@yuumasato @Mab879 @vojtapolasek

ggbecker: https://github.com/ComplianceAsCode/content/pull/7931

matejak:
Some rules that don't work with RHEL9 were enabled in this PR (e.g. the harden sshd crypto ones), where they have blended with the crowd of rules safely established on RHEL9.
Next time, we have to do it differently than flipping some prodtypes, and hoping that it will work. It is also a lot of rules, so going manually through them in a small group won't work either. The work has to be somewhat distributed.

PR #7749, opened 2021-10-15
Fixed regex used in OVAL related to audit rules

marcusburghardt:
It was rightly caught by test scenarios where different variations
of valid permissions were used.


Achilleasein:
Looks good to me, everything works.

PR #7750, opened 2021-10-15
Add oval-graph support for the test suite

matejak:
This PR flips the result format to ARF, and uses oval-graph to generate HTML graphs locally.

If `oval-graph` is available on the system that runs the tests, the test suite generates the interactive HTML visualization of the OVAL check based on the ARF. Generated HTML graphs are small, so they are currently not cleaned from the folder (in contrast to  reports).

It looks like that the ARF that oscap produces when called with `--remediate` refers to the state before the fix, so those graphs aren't useful.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -345,6 +345,12 @@ UserKnownHostsFile /dev/null
 
 All logs of Test Suite are stored in `logs` directory. The specific diretory is shown at the beginning of each test run.
 
+If you want more verbose logs, pass the `--dontclean` argument that preserves result files, reports and verbose scanner output
+even in cases when the test result went according to the expectations.
+If your system has the [oval-graph](://github.com/OpenSCAP/oval-graph) package installed that provides the `arf-to-html` command,
+the test suite will use it to extract OVAL evaluation details from ARFs, and save those condensed reports to the `logs` directory
+even if the `--dontclean` argument has been specified.
+
 ## Rule-based testing
 
 ```
@@ -499,7 +505,6 @@ If you would like to test all profile's rules against their test scenarios:
 ```
 ./test_suite.py combined --libvirt qemu:///system ssg-test-suite-rhel8 --datastream ../build/ssg-rhel8-ds.xml ospp
 ```
-
 
 
 # Analysis of results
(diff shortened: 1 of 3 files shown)


jan-cerny:
@matejak @Honny1 I have discovered that it's because arf-to-graph refuses to consume the ARF.

```
arf-to-graph --all-in-one --output /home/jcerny/work/git/scap-security-guide/logs/rule-custom-2021-11-15-1124/selinux_state-selinux_permissive.fail.sh-initial-graph.html /home/jcerny/work/git/scap-security-guide/logs/rule-custom-2021-11-15-1124/selinux_state-selinux_permissive.fail.sh-initial-arf.xml .
Error: This file "/home/jcerny/work/git/scap-security-guide/logs/rule-custom-2021-11-15-1124/selinux_state-selinux_permissive.fail.sh-initial-arf.xml" is not arf report file or there are no results
[jcerny@thinkpad scap-security-guide{pr/7750}]$ rpm -q oval-graph
oval-graph-1.3.1-1.fc33.noarch
```

@matejak It would be better to consume the stderr/stdout of the arf-to-graph command in the `TestEnv.arf_to_html` method.

@Honny1 What do you think?

[arf.zip](https://github.com/ComplianceAsCode/content/files/7538002/arf.zip)

Honny1:
@jan-cerny It seems you have replicated the [issue](https://github.com/OpenSCAP/oval-graph/issues/204) in the `oval-graph`. I'll find out where the problem is.

Honny1:
@jan-cerny The issue with `oval-graph` is fixed in [PR](https://github.com/OpenSCAP/oval-graph/pull/216).

jan-cerny:
@Honny1 Thanks for finding the root cause which was that the ARF didn't contain any system_data because no OVAL object has been collected.

Thanks to that, I found it suspicous that oscap didn't collect any object and I have discovered that there is a problem in the test scenario `linux_os/guide/system/selinux/selinux_state/tests/selinux_permissive.fail.sh` which will be fixed by https://github.com/ComplianceAsCode/content/pull/7943.

jan-cerny (approved):

jan-cerny:
@matejak As the problem is not in this PR but in oval-graph I think we can happily merge this unless you know about something that would prevent that. Just please rebase it on master so that we have the F34 gating job.

jan-cerny:
@matejak bump

Honny1:
The new version of `oval-graph` is in testing.

PR #7751, opened 2021-10-15
Change `sshd_set_idle_timeout` to require `sshd_set_keepalive_0`

yuumasato:
#### Description:

- Change `requires` of `sshd_set_idle_timeout` to `sshd_set_keepalive_0`, which ensures that the undocumented SSH user session idle time out works.
- Remove inconsistencies with `ClientAliveInterval` and `ClientAliveCountMax` in RHEL9 and Fedora
- Move Ubuntu 20.04 profiles to use `sshd_set_keepalive_0`.

#### Rationale:
- See also https://github.com/ComplianceAsCode/content/pull/6434
- Related to https://github.com/OpenSCAP/openscap/issues/1814


yuumasato:
@dodys There are changes related to Ubuntu 20.04 profiles

dodys:
> @dodys There are changes related to Ubuntu 20.04 profiles

so this means that for OpenSSH >= 8.2 we should default to sshd_set_keepalive_0, right? 
Just trying to make sure I'm not confusing things.

richardmaciel-canonical:
> > > > @dodys There are changes related to Ubuntu 20.04 profiles
> > > 
> > > 
> > > so this means that for OpenSSH >= 8.2 we should default to sshd_set_keepalive_0, right? Just trying to make sure I'm not confusing things.
> > 
> > 
> > @dodys No, it is the opposite, this matter is confusing, :).
> > ```
> > * Until OpenSSH < 8.2, the profiles can use `sshd_set_keepalive_0` (`ClientAliveCountMax 0`) to terminate idle session.
> > 
> > * From OpenSSH >= 8.2, setting `ClientAliveCountMax 0` disables the session termination feature.
> >   
> >   * See https://www.openssh.com/txt/release-8.2 and https://bugzilla.mindrot.org/show_bug.cgi?id=2627
> > ```
> 
> Got it! Then for 20.04 we have OpenSSH >= 8.2, then we should not move to `sshd_set_keepalive_0`. Also, STIG for 20.04 expects `ClientAliveCountMax 1`, which means we should have `var_sshd_set_keepalive=1`.
> 
> @richardmaciel-canonical could you please confirm what is needed for CIS? And since you've been discussing this ClientAliveCountMax for a while, could you review my comment above for STIG?

@dodys CIS for Ubuntu 20.04 uses `sshd_set_keepalive` with `var_sshd_set_keepalive=3`, which is what the benchmark requires.

For STIG, you're right. We need `sshd_set_keepalive` with `var_sshd_set_keepalive=1`

yuumasato:
Per comments, I've dropped changes in Ubuntu 20.04 CIS and STIG profiles (from c7c58e1 to d45a4f0).

vojtapolasek (approved):
I think it looks good, all issues were resolved.

PR #7755, opened 2021-10-15
Fix Ansible for rsyslog_encrypt_offload_* 

### LINKED ISSUE #7741: rules rsyslog_encrypt_offload_* have wrong regex
#### Description of problem:
Rules are using Ansible macro which is a lineinfile wrapper. This wrapper uses the supplied parameter and value as regex as well as the line to be inserrted into the destination config file. This usually works nicely but not in case when the line contains e.g. $.
The $ must be escaped in case the line is used as regex. This is not happening now.

master commit 29c92c1b98cb123e99dca30c0adefb4b14ab9c32 

#### Steps to Reproduce:

1. setup a VM and make sure that there is no file /etc/rsyslog.d/encrypt.conf
2.  build content with Ansible playbooks
3. run ansible-playbook -u root -i vm_ip, --tags="rsyslog_encrypt_offload_defaultnetstreamdriver" build/ansible/rhel8-playbook-stig.yml
4. append some bogus text to eh line in /etc/rsyslog.d/encrypt.conf created by previous run, mangling the parameter
4.

#### Actual Results:
The mangled line stays there, new correct line is appended

#### Expected Results:
The mangled line is erplaced by the correct line

Regex escaping should be implemented somewhere in the build process of Ansible playbooks. I think we can use the regex_escape filter.

Mab879:
#### Description:

Escaped the `$` in `rsyslog_encrypt_offload_*`

#### Rationale:

- Fixes #7741


jan-cerny (dismissed):

vojtapolasek:
Just a thought, is there actually a case where we would NOT want the regex to be escaped? Would it make sense to do this implicitly?

Mab879:
I'm having a hard time getting `regexp: '{{ {{{ regex }}} | regex_escape }}'` to play nice with build system and Ansible lint. Not sure how best to make it work.

shared/macros-ansible.jinja
```diff
@@ -25,13 +25,15 @@ value: "Setting={{ varname1 }}"
...
+{{%- macro ansible_lineinfile(msg='', path='', regex='', new_line='', create='no', state='present', with_items='', register='', when='', validate='', insert_after='', insert_before='', check_mode=False, escape_regex=False) -%}}
 - name: "{{{ msg or rule_title }}}"
   lineinfile:
     path: '{{{ path }}}'
     create: {{{ create }}}
-    {{%- if regex %}}
+    {{%- if regex and not escape_regex %}}
     regexp: '{{{ regex }}}'
+    {{%- elif regex and escape_regex %}}
+    regexp: '{{ {{{ regex }}} | regex_escape }}'
```
vojtapolasek, line 36: Maybe this is the solution?
```suggestion
    regexp: {{{ regex }}} | regex_escape
```

shared/macros-ansible.jinja
```diff
@@ -25,13 +25,15 @@ value: "Setting={{ varname1 }}"
...
+{{%- macro ansible_lineinfile(msg='', path='', regex='', new_line='', create='no', state='present', with_items='', register='', when='', validate='', insert_after='', insert_before='', check_mode=False, escape_regex=False) -%}}
 - name: "{{{ msg or rule_title }}}"
   lineinfile:
     path: '{{{ path }}}'
     create: {{{ create }}}
-    {{%- if regex %}}
+    {{%- if regex and not escape_regex %}}
     regexp: '{{{ regex }}}'
+    {{%- elif regex and escape_regex %}}
+    regexp: {{{ regex }}} | regex_escape
```
yuumasato, line 36: Does the `regex_escape` filter work when out of double curly braces?

yuumasato: From what I could check the `regex_escape` here has no effect.

yuumasato:
Another thing I also noticed is that the OVAL checks for [rsyslog_encrypt_offload_defaultnetstreamdriver](https://github.com/ComplianceAsCode/content/blob/f6541126a4d19bfef8752028467659ab9d9f74ed/linux_os/guide/system/logging/ensure_rsyslog_log_file_configuration/rsyslog_encrypt_offload_defaultnetstreamdriver/oval/shared.xml#L6) and related don't handle cases when `/etc/rsyslog.conf` has the parameter witt correct, but /etc/rsyslog.d/somefile.conf` has the parameter with wrong value.

I guess we want to fail in these cases? As the Ansible remediation always removes configurations in `/etc/rsyslog.d/`.
And unfortunately the Bash remediation is a bit behind and only checks for `/etc/rsyslog.conf`.

But I think these are out of the original scope of this PR and can be handled later...

yuumasato (approved):

yuumasato:
@Mab879 LGTM, before I merge, could you address the ansible-lint issues?

PR #7766, opened 2021-10-19
Re-introduce smartcard_configure_cert_checking to rhel8

matejak:
The rule has been removed in #7365 - it probably fell out of a profile.
However, this is not a reason for removal from a benchmark.
Removals can break tailoring, and subsequently cause regressions.


matusmarhefka:
> I understand the reasoning to re-add the rule, but does it make sense though?
> 
> The rule is about a package not available in RHEL8: https://github.com/ComplianceAsCode/content/pull/7365/files#r689482465

Indeed, the `pam_pkcs11` package (from RHEL7) is replaced by `sssd` on RHEL8. In general we don't want to remove rules from data stream to not break tailorings, but in this case we can do an exception as this can be thought of as a bug fix, the rule would never work on RHEL8.

PR #7774, opened 2021-10-20
Use different path to configure audit's `overflow_action` in RHEL7/OL7

ggbecker:
#### Description:
- Use different path to configure overflow_action in RHEL7/OL7
  - The correct path for this option in these OSes is
/etc/audisd/audispd.conf.

#### Rationale:

- Audit service can be started successfully.


ggbecker:
FYI @freddieRv

PR #7779, opened 2021-10-21
accounts_no_uid_except_zero: Don't run `passwd` if `awk` returns nothing

### LINKED ISSUE #7457: accounts_no_uid_except_zero bash remediation reports error if no match
#### Description of problem:

bash remediation of accounts_no_uid_except_zero reports passwd help usage if run without check before.

awk -F: '$3 == 0 && $1 != "root" { print $1 }' /etc/passwd | xargs --max-lines=1 passwd -l

if no matches come out of awk, it will run passwd -l without arguments ... 

Perhaps use "xargs -r --max-lines=1 passwd -l"    that avoids running if input is empty?

0.1.57

SUSE Linux Enterprise Server 15 SP3

#### Steps to Reproduce:

1. run above remediation line if no other account has uid 0 besides root (uually default)
2. 
3.
4.

#### Actual Results:

passwd help output

#### Expected Results:

no output at all

Mab879:
#### Description:

Adds `--no-run-if-empty` so that you don't a usage error if passing.

#### Rationale:

Fixes #7457


vojtapolasek (approved):
LGTM, thank you

PR #7780, opened 2021-10-25
tests/install_vm.py: add timeouted wait for VM to shutdown

matusmarhefka:
Added timeout should prevent issues where a VM is still in
the `running` state after `virsh console` disconnects and
therefore subsequent `virsh start` fails, e.g.:

```
Starting Reboot...

dracut Warning: Killing all remaining processes
Rebooting.
[  522.430163] reboot: Restarting system

error: Domain is already active
```

### DIFF
--- a/tests/install_vm.py
+++ b/tests/install_vm.py
@@ -4,6 +4,7 @@ import argparse
 import os
 import sys
 import subprocess
+import time
 
 
 def parse_args():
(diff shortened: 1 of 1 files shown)


matejak:
LGTM, perhaps just shorten the offending line, so the bot is satisfied.

matejak (approved):
Thank you for the PR!

PR #7781, opened 2021-10-25
Integrate compare_ds.py tool to Github Actions

ggbecker:
#### Description:

- Integrate compare_ds.py tool to Github Actions

#### Rationale:

- Cool datastream diff should be produced ~~and posted as a comment to the PR~~!
- ~~Unfortunately the comment can't be placed on a PR coming from a different fork. I'm looking for alternatives.~~
- Using the `pull_request_target` event, the workflow should be able to add comments to the PR. There are some security concerns though. 
  - Some references: 
    - https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/#improvements-for-public-repository-forks
    - https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request_target

The workflow will only trigger after it's already merged. So the test commit should be purged because it has no effect right now.


evgenyz:
How about using https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#workflow_run with single `if: ${{ steps.ctf.outputs.CTF_OUTPUT_SIZE != '0' }}` for the real diff job?

ggbecker:
> > How about using https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#workflow_run with single `if: ${{ steps.ctf.outputs.CTF_OUTPUT_SIZE != '0' }}` for the real diff job?
> > We could be able to use single https://github.com/ComplianceAsCode/content/blob/master/.github/workflows/ctf.yaml workflow as a dependency for all other jobs that require its output.
> 
> we can definitely improve how jobs relate to each other and put them as dependency of each other. I will take a look at this.

I'm starting to think that this is not a good idea. There are some limitations, for example the job that runs on a workflow_run doesn't know the PR context and after some tries I was not even able to make it work properly (https://github.com/ggbecker/content/pull/18)

Maybe we can use the same workflow file to accommodate all of these jobs that depend on CTF and use the simple `needs` keyword.

I have to investigate more. But I think we shouldn't block this one to get merged as we can refactor and improve the jobs later.

evgenyz:
Yeah, you're probably right. I just don't like CTF sprawling and wedging all over the place unchecked.

evgenyz (approved):

PR #7786, opened 2021-10-25
Tools: Compare rule's platform in compare_ds.py

evgenyz:
It will properly handle names (like "cpe:/a:ntp") as well as cpe-lang platforms with fact-refs (<cpe-lang:platform id="cpe_platform_machine">).

The logic is based on the original test script from @vojtapolasek and should help in migration to new CPE platforms.

### DIFF
--- a/utils/compare_ds.py
+++ b/utils/compare_ds.py
@@ -13,7 +13,8 @@ ns = {
     "oval": ssg.constants.oval_namespace,
     "catalog": ssg.constants.cat_namespace,
     "xlink": ssg.constants.xlink_namespace,
-    "ocil": ssg.constants.ocil_namespace
+    "ocil": ssg.constants.ocil_namespace,
+    "cpe-lang": ssg.constants.cpe_language_namespace,
 }
 remediation_type_to_uri = {
     "bash": ssg.constants.bash_system,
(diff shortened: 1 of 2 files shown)


utils/compare_ds.py
```diff
@@ -95,6 +96,42 @@ def print_offending_elements(elements, sign):
...
+                for cpe_platform in cpe_platforms:
+                    fact_refs = cpe_platform.findall(".//cpe-lang:fact-ref", ns)
+                    for fact_ref in fact_refs:
+                        entry["cpe"].append(fact_ref.get("name"))
+            else:
+                entry["cpe"].append(idref)
+
+    if entries[0]["cpe"] != entries[1]["cpe"]:
+        print("Platform has been changed for rule '{0}': {1} -> {2}".format(
+            entries[0]["rule"].get("id"), repr(entries[0]["cpe"]), repr(entries[1]["cpe"])))
```
ggbecker, line 132: Let's make the output more like the other diffs

something like:

```suggestion
        print("Platform definition %s differs for:" % (entries[0]["rule"].get("id")))
        print("--- old datastream")
        print("+++ new datastream")
        print("-{}".format(repr(entries[0]["cpe"])))
        print("+{}".format(repr(entries[1]["cpe"])))
```

This way we have nice colors when displaying the diff... even though the old and new entries are not really code, it will popup for user's eyes

evgenyz: Sure, done

ggbecker (approved):

PR #7788, opened 2021-10-26
File permission rules

dodys:
#### Description:

This PR makes changes and add rules regarding file permissions. If you think I should break this into multiple commits let me know, but they are not extensively complex.

- Make file_permissions template oval exclude symlinks
- Add ubuntu platform to dir_permissions_library_dirs tests
- Add ubuntu platform to file_permissions_binary_dirs bash
- Add tests to file_permissions_binary_dirs and file_permissions_library_dirs
- Add new rule dir_permissions_binary_dirs
- Add new rule file_permissions_audit_binaries
- Make permissions_local_var_log use file_permissions template
- Add ubuntu specific bash for permissions_local_var_log
- Fix permissions_local_var_log tests
- Add ubuntu specific bash to file_permissions_var_log


shared/templates/file_permissions/oval.template
```diff
@@ -67,6 +67,11 @@
       #}}
       <filter action="include">state_file_permissions{{{ FILEID }}}_{{{ loop.index0 }}}_mode_not_{{{ FILEMODE }}}</filter>
     {{%- endif %}}
+      <filter action="exclude">exclude_symlinks_{{{ FILEID }}}</filter>
```
Mab879, line 70: Do we want to wrap this with an if as well? So that in cases that we can to check symlinks we can?

dodys: sorry, I completely missed your comment.
I don't fully remember now, but the way this oval is written, it is not following symbolic links I think.
Also what would be a great if for this case? platform specific?

Mab879: I don't have any case off the top my head. Was just thought. I think we can leave it as is.

linux_os/guide/system/permissions/files/permissions_local_var_log/bash/ubuntu.sh
```diff
@@ -0,0 +1,14 @@
+# platform = multi_platform_ubuntu
+
+readarray -t files < <(find /var/log/)
```
matejak, line 3: This is not robust, consider using this approach: https://github.com/koalaman/shellcheck/wiki/SC2044

dodys: readarray is extensively used in CaC's code. Maybe we should leave as is in this PR and fix it altogether in a separate PR.
Thoughts?

ggbecker: Our file owner, groupowner, permissions template is already using this approach. I'm fine with this approach for now and then we can fix everything related to this in a different PR.

matejak: `readarray` in itself is fine, it's just the combination of `find` and `readarray` that is problematic. I am fine to address it later, #8032 will remind us :-)

ggbecker (approved):
LGTM.

ggbecker:
@dodys can you please rebase on top of latest master? we have changed the required check to use Fedora34 and this code is based on Fedora33. The rebase should retrigger the test using the correct configuration and then we will be able to merge this one.

PR #7790, opened 2021-10-26
Home directories definition and existence

marcusburghardt:
#### Description:

This PR brings OVAL, tests and remediation for the following rules:

- accounts_user_interactive_home_directory_defined
- accounts_user_interactive_home_directory_exists

#### Rationale:

These rules are required by some standards, like STIG and CIS. Therefore, it is valuable to have them properly working.
This PR partially fix:

https://bugzilla.redhat.com/show_bug.cgi?id=1884687

This PR is also related to https://github.com/ComplianceAsCode/content/pull/7770


linux_os/guide/system/accounts/accounts-session/accounts_user_interactive_home_directory_defined/ansible/shared.yml
```diff
@@ -0,0 +1,24 @@
...
+    split: ':'
+
+- name: Create local_users variable from the getent output
+  ansible.builtin.set_fact:
+    local_users: '{{ ansible_facts.getent_passwd|dict2items }}'
+
+- name: Ensure interactive users have a home directory defined
+  ansible.builtin.user:
+    name: '{{ item.key }}'
+    home: '/home/{{ item.key }}'
```
Mab879, line 19: So this assumption may not always be true and if the interactive user has a non-standard home directory this appears to . change that. Maybe we can add when condition this to only run if the user has no home dir.

linux_os/guide/system/accounts/accounts-session/accounts_user_interactive_home_directory_defined/bash/shared.sh
```diff
@@ -0,0 +1,9 @@
+# platform = multi_platform_all
+# reboot = false
+# strategy = restrict
+# complexity = low
+# disruption = low
+
+for user in `awk -F':' '{ if ($4 >= {{{ uid_min }}} && $4 != 65534) print $1 }' /etc/passwd`; do
+    sed -i "s/\($user:x:[0-9]*:[0-9]*:.*:\).*\(:.*\)$/\1\/home\/$user\2/g" /etc/passwd;
```
Mab879, line 8: See my comments above. I have the same concerns.

Mab879 (changes requested):
Overall, I like this pr. However, I have concerns about the remediations. '

linux_os/guide/system/accounts/accounts-session/accounts_user_interactive_home_directory_exists/oval/shared.xml
```diff
@@ -0,0 +1,91 @@
...
+
+  <!-- #### prepare a password object for the two tests in this rule #### -->
+  <unix:password_object id="object_accounts_user_interactive_home_directory_exists_objects"
+                        version="1">
+    <unix:username datatype="string" operation="not equal">nobody</unix:username>
+    <filter action="include">state_accounts_user_interactive_home_directory_exists_uids</filter>
+  </unix:password_object>
+
+  <unix:password_state id="state_accounts_user_interactive_home_directory_exists_uids" version="1">
+    <unix:user_id datatype="int" operation="greater than or equal">{{{ uid_min }}}</unix:user_id>
```
yuumasato, line 20: Should the `root` user be checked too?

linux_os/guide/system/accounts/accounts-session/accounts_user_interactive_home_directory_defined/ansible/shared.yml
yuumasato: I think it should check the existence of the path defined passwd.

marcusburghardt: Hum, good point @Mab879 . Indeed, without a conditional, the remediation could also change something not expected to be changed. I will include this test. Thanks.

linux_os/guide/system/accounts/accounts-session/accounts_user_interactive_home_directory_defined/oval/shared.xml
```diff
@@ -0,0 +1,36 @@
...
+       #### creation of object #### -->
+  <unix:password_object id="object_accounts_user_interactive_home_directory_defined_objects"
+                        version="1">
+    <unix:username datatype="string" operation="not equal">nobody</unix:username>
+    <filter action="include">state_accounts_user_interactive_home_directory_defined_uids</filter>
+  </unix:password_object>
+
+  <unix:password_state id="state_accounts_user_interactive_home_directory_defined_uids"
+                       version="1">
+    <unix:user_id datatype="int" operation="greater than or equal">{{{ uid_min }}}</unix:user_id>
```
yuumasato, line 21: Is it clear what is an interactive user? Does STIG define it?

https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_8_security_technical_implementation_guide/V-230320
The STIG item related to this rule doesn't explicitly say to check `root` user, but defines a fallback to check users from UID 1000.

But int https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_8_security_technical_implementation_guide/V-230323 it seems to define interactive users the ones that can login and leave logon traces.

marcusburghardt: Hi @yuumasato , this is not explicit neither on CIS. However, I am much more prone to interpret `interactive users` as `regular users`. The fallback you mentioned also suggests that STIG interprets this way too.

Considering the rules are related to home directories "definition" and "existence" I don't think it makes sense to test root home directory too, since it should always be defined and present in any distro which follows the Filesystem Hierarchy Standard (FHS).

I would add that if we need the same tests for root (not a regular user), separate rules should be created because even the logic to assess it is a little bit different.

linux_os/guide/system/accounts/accounts-session/accounts_user_interactive_home_directory_defined/rule.yml
```diff
@@ -8,6 +8,10 @@ description: |-
     Assign home directories to all interactive users that currently do not
     have a home directory assigned.
 
+    This rule checks if the home directory is properly defined in a folder which has
+    at least one parent folder, like "user" in "/home/user" or "/remote/users/user".
+    Therefore, it will report finds for home directories like /users, /tmp or /.
```
Mab879, line 13:
```suggestion
    Therefore, this rule will report a finding for home directories like <tt>/users</tt>, <tt>/tmp</tt> or <tt>/</tt>,
```

Mab879 (changes requested):
Just one minor change.

linux_os/guide/system/accounts/accounts-session/accounts_user_interactive_home_directory_defined/rule.yml
marcusburghardt: Done. Only changed the last "," by a "." : )

Mab879 (approved):

PR #7791, opened 2021-10-26
File groupowner rules

dodys:
#### Description:

- Adapt file_groupowner_var_log to accept other gids
- Add ubuntu to file_groupownership_system_commands_dirs bash script
- Make file_groupownership_system_commands_dirs tests work in ubuntu
- Add new rule file_groupownership_audit_configuration
- Add new rule file_groupownership_audit_binaries
- Add new rule file_groupowner_var_log_syslog
- Add new rule dir_groupownership_binary_dirs
- Add ubuntu2004 to prodtype in root_permissions_syslibrary_files
- Make root_permissions_syslibrary_files use file_groupowner template
- Make root_permissions_syslibrary_files tests work in ubuntu and add it to Ubuntu 20.04 STIG profile


dodys:
@ggbecker I would appreciate your input here about the failing tests. Similar to the file owner, the missing_file_test.pass.sh fails for file_groupowner_var_log_syslog.
Also the template test incorrect_groupowner.fail.sh is passing for rule root_permissions_syslibrary_files, but it shouldn't be executed because the rule has regex and instead we should expect only the incorrect_group.fail.sh to be executed but since their names differ, both are executed. What would be the best solution here? Rename the rule's specific test?

ggbecker:
> @ggbecker I would appreciate your input here about the failing tests. Similar to the file owner, the missing_file_test.pass.sh fails for file_groupowner_var_log_syslog. Also the template test incorrect_groupowner.fail.sh is passing for rule root_permissions_syslibrary_files, but it shouldn't be executed because the rule has regex and instead we should expect only the incorrect_group.fail.sh to be executed but since their names differ, both are executed. What would be the best solution here? Rename the rule's specific test?

I'll look into this when I have some time. Please bear with me

linux_os/guide/system/permissions/files/permissions_within_important_dirs/root_permissions_syslibrary_files/tests/correct_group.pass.sh
```diff
@@ -1,4 +1,4 @@
-# platform = multi_platform_sle,Red Hat Enterprise Linux 8,multi_platform_fedora
+# platform = multi_platform_sle,Red Hat Enterprise Linux 8,multi_platform_fedora,multi_platform_ubuntu
```
ggbecker, line 1: We need to change the name of this file to `correct_groupowner.pass.sh` so it overrides properly the templated tests.

linux_os/guide/system/permissions/files/permissions_within_important_dirs/root_permissions_syslibrary_files/tests/incorrect_group.fail.sh
```diff
@@ -1,10 +1,11 @@
-# platform = multi_platform_sle,Red Hat Enterprise Linux 8,multi_platform_fedora
+# platform = multi_platform_sle,Red Hat Enterprise Linux 8,multi_platform_fedora,multi_platform_ubuntu
```
ggbecker, line 1: We need to change the name of this file to `incorrect_groupowner.pass.sh` so it overrides properly the templated tests.

ggbecker (changes requested):
This file needs to be updated to handle the test when a rule is not using the missing_file_pass attribute.

I have tested the following and it seems to work:

```
#!/bin/bash
#

{{% for path in FILEPATH %}}
    {{% if MISSING_FILE_PASS %}}
        rm -f {{{ path }}}
    {{% else %}}
        mkdir -p $(basename {{{ path }}})
        touch {{{ path }}}
        chgrp {{{ FILEGID }}} {{{ path }}}
    {{% endif %}}
{{% endfor %}}
```

It basically sets up the machine correctly and the rule passes normally when the missing_file_pass is set to false.

linux_os/guide/system/permissions/files/permissions_within_important_dirs/root_permissions_syslibrary_files/rule.yml
```diff
@@ -59,3 +60,14 @@ ocil: |-
...
+template:
+    name: file_groupowner
+    vars:
+        filepath:
+            - /lib/
+            - /lib64/
+            - /usr/lib/
+            - /usr/lib64/
+        file_regex: ^.*$
+        filegid: '0'
```
ggbecker, line 73:
```suggestion
        filegid: '0'
        recursive: 'true'
```

I believe we need to use the `recursive` option here. What do you think?

dodys: No need to as we have the file_regex, that is enough to search for every file under the specified filepath
The recursive flag is actually needed when you pass a filepath, that is a directory and there's no regex, so we make sure that we are searching for subdirectories as well

ggbecker (approved):
LGTM

PR #7794, opened 2021-10-27
Generate OCIL different way

jan-cerny:
#### Description:

OCIL will be generated directly to a file while generating shorthand XML (in other words while processing compiled/resolved rules). 

To facilitate review you can use the new feature of compare_ds.py to compare OCILs in the built data stream which is proposed in PR https://github.com/ComplianceAsCode/content/pull/7793.

#### Rationale:

Current way of generating OCIL is to put the OCIL data to shorthand XML, then extract it using XSLT, then replace it by reference using XSLT. This is complex and not necessary. The way proposed in this PR is simpler than the current way. We will generate OCIL while processing the data we have in memory. This reduces the amount of build steps to remove some build targets and removes some XSLT templates. In future, it can help when getting rid of shorthand because it removes some dependencies of shorthand.xml. 

The disadvantage is a possible confusion caused by the fact that now the build_shorthand.py script generates 2 output files.

### DIFF
--- a/ssg/build_yaml.py
+++ b/ssg/build_yaml.py
@@ -1609,7 +1609,7 @@ class Rule(XCCDFEntity):
         # The "ocil" key in compiled rules contains HTML and XML elements
         # but OCIL question texts shouldn't contain HTML or XML elements,
         # therefore removing them.
-        if self.ocil:
+        if self.ocil is not None:
             ocil_without_tags = re.sub(r"</?[^>]+>", "", self.ocil)
         else:
             ocil_without_tags = ""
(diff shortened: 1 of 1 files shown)


jan-cerny:
I have realized that build_shorthand.py probably isn't the best place where OCIL generation should happen. But I think that at this moment we don't have any "best place". OCIL generation can happen in any of our scripts that iterate over compiled rules. There are multiple scripts that iterate over compiled rules, for example the combine_remediations.py, build_shorthand.py, build_templated_content.py. Each of them loads compiled rules from the disk and processes them, which means that compiled rules are loaded and processed multiple times during a build of a single product. I think that one way that we could go in future is to merge these in a single script so that the compiled rules will be loaded and processed only once.

PR #7795, opened 2021-10-27
Add YAML Lint CI job

Mab879:
#### Description:

Adds YAML Lint for changed files to CI.

#### Rationale:

Improve our YAML formatting.

### DIFF
--- a/.yamllint.yml
+++ b/.yamllint.yml
@@ -5,3 +5,4 @@ rules:
     line-length:
         max: 99
         level: warning
+    document-start: disable


.yamllint.yml
```diff
@@ -0,0 +1,7 @@
+---
+extends: default
+
+rules:
+    line-length:
+        max: 99
+        level: warning
```
yuumasato, line 7: I would suggest to ignore `document-starts` for now:
`document-start: disable`

yuumasato (commented):
The 'yamllint' tool cannot parse/validate Jinja2 templates.
Ref: https://github.com/adrienverge/yamllint/issues/65

Would it make more sense to add this as a `ctest` to run over the compiled profiles/groups/rules?

.yamllint.yml
Mab879: > Would it make more sense to add this as a ctest to run over the compiled profiles/groups/rules?
Yes, and it would catch errors from our Jinja templating. 

> I would suggest to ignore document-starts for now:
Good idea.

Mab879: > Would it make more sense to add this as a ctest to run over the compiled profiles/groups/rules?

After trying it out (using `find build/*/groups build/*/rules build/*/profiles -name "*.yml" | xargs yamllint -c .yamllint.yml`) I don't think there is much value on using on the compiled rules. As it seems those are create with `yaml.dump` Ideally we would validate right after jinja runs.

PR #7796, opened 2021-10-27
Add rule schema

Mab879:
#### Description:

Adds a [JSON Schema](http://json-schema.org/) for rule.yml files.

#### Rationale:

If an IDE is configured correct allows for auto complete. Also, this can be used in the future for better linting of the project.

### DIFF
--- a/shared/schemas/rule.json
+++ b/shared/schemas/rule.json
@@ -1,0 +1,78 @@
+{
+  "$id": "http://complianceascode.io/schemas/1.0/rule.json",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "object",
+  "properties": {
+    "documentation_complete": {
+      "type": "boolean"
+    },
+    "title": {
+      "type": "string"
+    },
+    "prodtype": {
+      "type": "string"
+    },
+    "description": {
+      "type": "string"
+    },
+    "rationale": {
+      "type": "string"
+    },
+    "severity": {
+      "type": "string",
+      "enum": [
+        "high",
+        "medium",
+        "low",
+        "unknown"
+      ]
+    },
+    "identifiers": {
+      "type": "object",
+      "uniqueItems": true
+    },
+    "references": {
+      "type": "object",
+      "uniqueItems": true
+    },
+    "ocil_clause": {
+      "type": "string"
+    },
+    "ocil": {
+      "type": "string"
+    },
+    "platform": {
+      "type": "string"
+    },
+    "warnings": {
+      "type": "array"
+    },
+    "conflicts": {
+      "type": "array"
+    },
+    "requires": {
+      "type": "array"
+    },
+    "template": {
+      "type": "object",
+      "additionalProperties": false,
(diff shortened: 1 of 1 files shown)


jan-cerny:
> If an IDE is configured correct allows for auto complete. Also, this can be used in the future for better linting of the project.

How do I make it work? Can I do it in Visual Studio Code?

Also, can this schema be reused in our code that parses the YAML files? I mean in ssg/build_yaml.py on line 1242 we define the required keys. That defines the schema in a way. Maybe in future this definition could be sourced from JSON schema?

Mab879:
> How do I make it work? Can I do it in Visual Studio Code?

Yes, you can use VS Code. You will need [Red Hat's YAML Extension](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml)

And if you on this branch you can add the following to your settings.json.

```    
"yaml.schemas": {
        "shared/schemas/rule.json": "rule.yml"
    }
```

I have also tested with Jetbrains' PyCharm.


> Also, can this schema be reused in our code that parses the YAML files? I mean in ssg/build_yaml.py on line 1242 we define the required keys. That defines the schema in a way. Maybe in future this definition could be sourced from JSON schema

A possibility. There are Python [validators](https://json-schema.org/implementations.html#validator-python).

marcusburghardt:
> > How do I make it work? Can I do it in Visual Studio Code?
> 
> Yes, you can use VS Code. You will need [Red Hat's YAML Extension](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml)
> 
> And if you on this branch you can add the following to your settings.json.
> 
> ```
> "yaml.schemas": {
>         "shared/schemas/rule.json": "rule.yml"
>     }
> ```
> 
> I have also tested with Jetbrains' PyCharm.
> 
> > Also, can this schema be reused in our code that parses the YAML files? I mean in ssg/build_yaml.py on line 1242 we define the required keys. That defines the schema in a way. Maybe in future this definition could be sourced from JSON schema
> 
> A possibility. There are Python [validators](https://json-schema.org/implementations.html#validator-python).

FYI, I just included this mentioned extension in ansible-role-openscap:
https://galaxy.ansible.com/marcusburghardt/ansible_role_openscap

I will follow this PR to also automate the update of `settings.json`

jan-cerny:
It works amazing.

Except for the rules that contain Jinja conditions, for example when I open linux_os/guide/system/accounts/accounts-pam/display_login_attempts/rule.yml in VS Code it's all red. Can we do anything about that?

Mab879:
>Except for the rules that contain Jinja conditions, for example when I open linux_os/guide/system/accounts/accounts-pam/display_login_attempts/rule.yml in VS Code it's all red. Can we do anything about that?

@jan-cerny I don't think there is. Those errors appear to be from YAML validator, not the schema validator. Which is correct in stating that file is not valid YAML. That is the one issue with using Jinja in our YAML, that it makes it hard to use YAML tooling.

jan-cerny (approved):

PR #7809, opened 2021-11-01
utils/add_platform_rule.py: Allow specifying scan type

JAORMX:
The script was hardcoded to use `Platform`, while folks might also need
to test out `Node` scans.

### DIFF
--- a/utils/add_platform_rule.py
+++ b/utils/add_platform_rule.py
@@ -134,7 +134,7 @@ kind: ComplianceScan
 metadata:
   name: test
 spec:
-  scanType: Platform
+  scanType: {TYPE}
   profile: {PROFILE}
   content: ssg-ocp4-ds.xml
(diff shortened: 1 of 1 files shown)


lmzuccarelli (commented):
Applied the changes and tested using the following ` ./utils/add_platform_rule.py cluster-test --scan-type=Node --rule tls_version_check_masters_workers`

Results
```
Output from last phase check: DONE COMPLIANT
* The result is 'COMPLIANT'
```

Execute without the --scan-type (default is Platform) 
Result 
```
Output from last phase check: DONE NOT-APPLICABLE
* The result is 'NOT-APPLICABLE'
```

jhrozek (approved):
Even though it was @lmzuccarelli who reviewed and tested the patch, I'm rubber-stamping it so that it gets merged.

PR #7810, opened 2021-11-01
Improve combine_remediations.py

jan-cerny:
#### Description:

The script combine_remediations.py will be executed only once for each product instead of executing it for each remediation type. The script is now iterating directly over compiled rules. As a result, each compiled rule will be loaded and processed only once.

For more details please read the commit message of every commit.

#### Rationale:

Tries to simplify the build process, reduces the amount of CMake targets.

Possibly should speed up the build, but I haven't measured the time yet.


cmake/SSGCommon.cmake
```diff
@@ -255,6 +261,7 @@ endmacro()
 macro(ssg_build_remediations PRODUCT)
     message(STATUS "Scanning for dependencies of ${PRODUCT} fixes (bash, ansible, puppet, anaconda, ignition, kubernetes and blueprint)...")
 
+    _ssg_combine_remediations(${PRODUCT} "${PRODUCT_REMEDIATION_LANGUAGES}")
     _ssg_build_remediations_for_language(${PRODUCT} "${PRODUCT_REMEDIATION_LANGUAGES}")
```
yuumasato, line 265: The names of `_ssg_combine_remediations` and `combine_remdiations.py` mislead me a bit.
AFAIU the macro and script are selecting or filtering the appropriate remediations.

Same for `_ssg_build_remediations_for_language`, I think this the macro that is actually combining the fixes into a `${LANGUAGE}-fixes.xml`.

jan-cerny:
rebased && renamed cmake targets and script combine_remediations.py

yuumasato (changes requested):
Looks great to me, thank you for the great commit messages.

> Possibly should speed up the build, but I haven't measured the time yet.

It seems to have improved by almost a minute, from 7m20s to 6m35s.

build-scripts/collect_remediations.py
```diff
@@ -0,0 +1,117 @@
...
+import ssg.build_yaml
+import ssg.rules
+import ssg.jinja
+import ssg.environment
+import ssg.utils
+import ssg.xml
+
+
+def parse_args():
+    p = argparse.ArgumentParser()
```
matejak, line 18: What about describing the function that the script performs, i.e. that it puts all remediations, templated or not, to the output directory.

build-scripts/collect_remediations.py
```diff
@@ -0,0 +1,117 @@
...
+                    fix_path = templated_fix_path
+        if fix_path is None:
+            # neither static nor templated remediation found
+            continue
+        remediation_obj = remediation_cls(fix_path)
+        remediation_obj.associate_rule(rule)
+        fix = remediation.process(remediation_obj, env_yaml)
+        if fix:
+            output_file_path = os.path.join(output_dirs[lang], rule.id_ + ext)
+            remediation.write_fix_to_file(fix, output_file_path)
```
matejak, line 90: I propose to extract this to a function as well - given a path, we either write the fix into a directory, or we don't do anything.

build-scripts/collect_remediations.py
```diff
@@ -0,0 +1,117 @@
...
+        if fix_path is None:
+            # check if we have a templated remediation instead
+            if os.path.isdir(language_fixes_from_templates_dir):
+                templated_fix_path = os.path.join(
+                    language_fixes_from_templates_dir, rule.id_ + ext)
+                if os.path.exists(templated_fix_path):
+                    fix_path = templated_fix_path
+        if fix_path is None:
+            # neither static nor templated remediation found
+            continue
```
matejak, line 84: I propose to extract this into a function that returns the "best guess" fix path.
The interesting thing is that if a rule has a "manual" remediation, it is used in favor of the template remediation. I wonder whether this is a good approach - one can disable template content easily in the template invocation, so I would say that template should have precedence, or there could be a warning that there is ambiguous content available.

matejak (changes requested):
The PR introduces great changes, and the only things that I suggest are related to the code layout.

build-scripts/collect_remediations.py
yuumasato: But disabling a language in the `rule.yml` disables it for all products.

If there is a static remediation it may be because the template doesn't suit a specific product and (probably) cannot be easily incorporated into the template.

jan-cerny:
refactores, updated according to review and rebased

yuumasato (approved):
Looks great to me.
Pending approval from @matejak.

matejak (approved):

matejak:
Perfect, I have performed manual XCCDF XML diff, and there were no changes, so congratulations for finishing this great piece of work!

PR #7811, opened 2021-11-01
Add unit test for kubelet_enable_iptables_util_chains

Vincent056:
The unit test was missing for `kubelet_enable_iptables_util_chains`, this PR adds the unit test for it.

### DIFF
--- a/applications/openshift/kubelet/kubelet_enable_iptables_util_chains/tests/enable_iptables_util_chains.fail.sh
+++ b/applications/openshift/kubelet/kubelet_enable_iptables_util_chains/tests/enable_iptables_util_chains.fail.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+# remediation = none
+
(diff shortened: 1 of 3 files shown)


applications/openshift/kubelet/kubelet_enable_iptables_util_chains/rule.yml
```diff
@@ -44,5 +44,5 @@ template:
         filepath: /etc/kubernetes/kubelet.conf
         yamlpath: ".makeIPTablesUtilChains"
         values:
-         - value: "true"
-           operation: "equals"
```
JAORMX, line 48: This changes the rule back to its non-working state.

JAORMX (changes requested):
Let's also add a test that checks that the key is there with the `false` key

JAORMX (approved):
Thanks for adding unit tests!

PR #7815, opened 2021-11-02
Introduce better jinja-based tables

matejak:
This PR adds Python code that generate tables from compiled rules, profiles and variables.
Following issues have been fixed:

- Ordering - table rows are ordered by references in the proper ascending order.
- Variable substitutions - profile-based tables got a working column where profile variable substitutions are shown. The information is extracted from rule description, so it's not 100% reliable.
- Speed - profile-based scripts (anssirefs) are much faster.

Following issues have appeared though:

- Surprisingly, new reference-based scripts are slower. It's probably cheaper to read the XCCDF, process it and dump the output to a table. Proposed solution - be able to generate all reference tables in one step. WDYT?
- The regex description of reference format is a pain to include in cmake files. WDYT of hadcoding the reference-ref regex relationship e.g. in `constants.py`?


products/rhel8/CMakeLists.txt
```diff
@@ -9,15 +9,29 @@ set(DISA_SRG_TYPE "os")
 ssg_build_product(${PRODUCT})
 
 ssg_build_html_table_by_ref(${PRODUCT} "nist")
+set(NIST_REF_FORMAT "'.*-(\\\\d+)(?:\\\\((\\\\d+)\\\\))?'")
+ssg_build_html_ref_table("new-table-${PRODUCT}-nist" "${PRODUCT}" "nist" "${NIST_REF_FORMAT}")
 ssg_build_html_table_by_ref(${PRODUCT} "cui")
+set(CUI_REF_FORMAT "'(\\\\d+)(?:\\\\.(\\\\w+)(?:\\\\.(\\\\w+)(?:\\\\.(\\\\w+))?)?)?'")
+ssg_build_html_ref_table("new-table-${PRODUCT}-cui" "${PRODUCT}" "cui" "${CUI_REF_FORMAT}")
 ssg_build_html_table_by_ref(${PRODUCT} "cis")
+set(CIS_REF_FORMAT "'(\\\\d+)\\\\.(\\\\d+)(?:\\\\.(\\\\w+)(?:\\\\.(\\\\w+)(?:\\\\.(\\\\w+))?)?)?'")
```
jan-cerny, line 18: I wonder what the regexes are good for. Are there any references that don't match the regex? In this specific example, I believe you want to use all of the `cis@rhel8` references. Maybe you don't need them at all?

cmake/SSGCommon.cmake
```diff
@@ -1257,6 +1257,57 @@ macro(ssg_build_html_nistrefs_table PRODUCT PROFILE)
         DESTINATION "${SSG_TABLE_INSTALL_DIR}")
 endmacro()
 
+
+macro(ssg_build_html_ref_table BASENAME PRODUCT REFERENCE REF_FORMAT)
+    add_custom_command(
+        OUTPUT "${CMAKE_BINARY_DIR}/tables/${BASENAME}.html"
+        COMMAND "${CMAKE_COMMAND}" -E make_directory "${CMAKE_BINARY_DIR}/tables"
+        COMMAND env "PYTHONPATH=$ENV{PYTHONPATH}" "${PYTHON_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/utils/gen_reference_table.py" --build-dir "${CMAKE_BINARY_DIR}" --output "${CMAKE_BINARY_DIR}/tables/${BASENAME}.html" --ref-format "${REF_FORMAT}" "${PRODUCT}" "${REFERENCE}"
```
jan-cerny, line 1265: I would execute the script only once and generate them in a single step.

products/rhel8/CMakeLists.txt
matejak: The trick is that if I describe a reference using a regex with groups, those groups can be used when sorting them in a smart way.
But it looks terrible in the cmake config, and it also doesn't make sense to be repeated in all products.

jan-cerny: If I rephrase that, the purpose of these regular expressions is correct sorting of the items in the generated tables. Is that correct? And is there any other usage of these regular expressions?

matejak: It has also a priority purpose - if there are more reference strings in that reference category, ones that match the regex are listed first. One can observe this on ANSSI BP/NT references.

cmake/SSGCommon.cmake
matejak: That would mean introducing a `generate_references_tables ... <output filename template> <ref1> <ref2> <ref3> ...` script - I will check that out.

jan-cerny: That should be possible - you can get inspired in the PR https://github.com/ComplianceAsCode/content/blob/ef14ad67b67e58272889d410391f196a5f449842/cmake/SSGCommon.cmake#L196

matejak:
New tables need review, and if they pass, a new commit can nuke old tables, and substitute them with new ones where applicable.

jan-cerny:
It all looks great.

Make sure that the file names of the new tables are same as the names of the old table files to avoid problems with packaging, rpmdiff etc.

I noticed that the format of the NIST tables changed in build/tables/new-table-rhel8-cis.html. The change is up to the discussion.  The question is how people use this table. Neither the previous form nor the new form seems to be actually useful. In the old form, rules with multiple references repeated, in the new form, the references are grouped by rules. I would expect that the form of the table would follow the structure of the NIST controls document - ie. controls sorted the same way as in the document and for each control show all the relevant rules (that would mean to have some of the rules listed multiple times in the table). But that might be not worth the time needed to deleop it.

utils/gen_profile_table.py
```diff
@@ -0,0 +1,70 @@
...
+    def process_rules(self, reference):
+        super(HtmlOutput, self).process_rules(reference)
+
+        self.template_data["profile"] = self.profile
+
+
+def update_parser(parser):
+    parser.add_argument(
+        "profile", metavar="PROFILE_ID", help="The ID of the profile")
+
```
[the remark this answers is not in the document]
matejak: Do we need a wallet when we have only one banknote? That's the question!

matejak:
> Looks mostly good.
> 
> But I can see some weird output in table-rhel7-anssirefs.html. For example, "Ensure SELinux State is Enforcing" is listed there only once although it belongs into 2 requirements - R4 and R66. It's sorted by requriement, so if I scroll down around R66 I won't see "Ensure SELinux State is Enforcing" there so I can easily miss it.

This is actually a feature - a rule appears in the table only once, and references are aggregated in the row of the first occurrence. It is an improvement to the past state, when the order was not consistent, and one had to search for a requirement anyway. It is clear though that the UX can be improved in the future.

utils/gen_profile_table.py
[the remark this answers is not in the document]
jan-cerny: I don't understand

jan-cerny:
> This is actually a feature - a rule appears in the table only once, and references are aggregated in the row of the first occurrence. It is an improvement to the past state, when the order was not consistent, and one had to search for a requirement anyway. It is clear though that the UX can be improved in the future.

I have a hard time thinking about this as an feature as I don't see use case for this layout. I would think of it as an incovenience. However, I think it can be fixed in a next PR as the goal of us was to speed up the build and the previous organization was also bad.

jan-cerny (approved):

PR #7820, opened 2021-11-02
Improve ssgts gh action

ggbecker:
#### Description:

- Enable verbose option in CTF to improve debugging.
- Cache CTF output to be used by different jobs.
- Print common logs in case of failure.

#### Rationale:

- Depends on https://github.com/mildas/content-test-filtering/pull/21 for `verbose` mode to work correctly.


mildas:
@ggbecker What do you think about printing SSGTS output in `Fail in case of ERROR present in logs_bash/test_suite.log or logs_ansible/test_suite.log` step?
Current error doesn't say much: `Run exit 1 Error: Process completed with exit code 1.`
Probably `cat logs_bash/test_suite.log; cat logs_ansible/test_suite.log; exit 1` is enough

mildas (approved):

mildas:
LGTM

PR #7824, opened 2021-11-03
Permissions and ownership of files under home directories

marcusburghardt:
### Description:
This PR brings OVAL, tests and remediation for the following rules:

accounts_users_home_files_ownership
accounts_users_home_files_groupownership
accounts_users_home_files_permissions

### Rationale:
These rules are required by some standards, like STIG and CIS. Therefore, it is valuable to have them properly working.
This PR partially fix:

https://bugzilla.redhat.com/show_bug.cgi?id=1884687

This PR is also related to #7770 and #7790.


linux_os/guide/system/accounts/accounts-session/accounts_users_home_files_groupownership/bash/shared.sh
```diff
@@ -0,0 +1,7 @@
+# platform = multi_platform_all
+# reboot = false
+# strategy = restrict
+# complexity = low
+# disruption = low
+
+awk -F':' '{ if ($4 >= {{{ gid_min }}} && $4 != 65534) system("chgrp -Rf " $4" "$6) }' /etc/passwd
```
yuumasato, line 7: So if I understand correctly this is changing the group owner of all files in the user's home directory to his own user group.
But the rule is not about all files in a users home being owned by his own group, it is about all files being group owned by a group which the user is a member of, it doesn't need to be his own group.

Ref: https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-204472?version=V3R4

linux_os/guide/system/accounts/accounts-session/accounts_users_home_files_groupownership/ansible/shared.yml
```diff
@@ -0,0 +1,32 @@
...
+  ansible.builtin.set_fact:
+    local_users: '{{ ansible_facts.getent_passwd|dict2items }}'
+
+- name: Test for existence home directories to avoid creating them, but only fixing ownership
+  ansible.builtin.stat:
+    path: '{{ item.value[4] }}'
+  register: path_exists
+  loop: '{{ local_users }}'
+  when:
+    - item.value[2]|int >= {{{ uid_min }}}
```
yuumasato, line 22: It seems to be inconsistent with the bash remediation, which compares to GID.

linux_os/guide/system/accounts/accounts-session/accounts_users_home_files_groupownership/oval/shared.xml
```diff
@@ -0,0 +1,52 @@
...
+    </criteria>
+  </definition>
+
+  <unix:password_object id="object_accounts_users_home_files_groupownership_objects" version="1">
+    <unix:username datatype="string" operation="not equal">nobody</unix:username>
+    <filter action="include">state_accounts_users_home_files_groupownership_interactive_gids</filter>
+  </unix:password_object>
+
+  <unix:password_state id="state_accounts_users_home_files_groupownership_interactive_gids" version="1">
+    <unix:user_id datatype="int" operation="greater than or equal">{{{ gid_min }}}</unix:user_id>
```
yuumasato, line 16: Should the `root` not considered an interactive user?

linux_os/guide/system/accounts/accounts-session/accounts_users_home_files_permissions/oval/shared.xml
```diff
@@ -0,0 +1,52 @@
...
+
+  <!-- For detailed comments about logic used in this OVAL, check the
+       "file_ownership_home_directories" rule. -->
+  <unix:password_object id="object_accounts_users_home_files_permissions_objects" version="1">
+    <unix:username datatype="string" operation="not equal">nobody</unix:username>
+    <filter action="include">state_accounts_users_home_files_permissions_interactive_uids</filter>
+  </unix:password_object>
+
+  <unix:password_state id="state_accounts_users_home_files_permissions_interactive_uids" version="1">
+    <unix:user_id datatype="int" operation="greater than or equal">{{{ uid_min }}}</unix:user_id>
```
yuumasato, line 18: The RHEL7 STIG doesn't explicitly mention `root`: https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-204471?version=V3R4

But shouldn't  the `root` user be checked as well?

linux_os/guide/system/accounts/accounts-session/accounts_users_home_files_ownership/oval/shared.xml
```diff
@@ -0,0 +1,52 @@
...
+    </criteria>
+  </definition>
+
+  <unix:password_object id="object_accounts_users_home_files_ownership_objects" version="1">
+    <unix:username datatype="string" operation="not equal">nobody</unix:username>
+    <filter action="include">state_file_ownership_home_directories_interactive_uids</filter>
+  </unix:password_object>
+
+  <unix:password_state id="state_accounts_users_home_files_ownership_interactive_uids" version="1">
+    <unix:user_id datatype="int" operation="greater than or equal">{{{ uid_min }}}</unix:user_id>
```
yuumasato, line 16: The RHEL7 STIG doesn't explicitly mention `root`: https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-204471?version=V3R4

But shouldn't  the `root` user be checked as well?

linux_os/guide/system/accounts/accounts-session/accounts_users_home_files_permissions/ansible/shared.yml
```diff
@@ -0,0 +1,32 @@
...
+  register: path_exists
+  loop: '{{ local_users }}'
+  when:
+    - item.value[2]|int >= {{{ uid_min }}}
+    - item.value[2]|int != 65534
+
+- name: Ensure interactive local users are the group-owners of their respective home directories
+  ansible.builtin.file:
+    path: '{{ item.0.value[4] }}'
+    mode: '0700'
```
yuumasato, line 28: Why check for `750` but remediate for `700`?
Same for Bash remediation.

linux_os/guide/system/accounts/accounts-session/accounts_users_home_files_permissions/bash/shared.sh
```diff
@@ -0,0 +1,7 @@
+# platform = multi_platform_all
+# reboot = false
+# strategy = restrict
+# complexity = low
+# disruption = low
+
+awk -F':' '{ if ($4 >= {{{ uid_min }}} && $4 != 65534) system("chmod -Rf 700 "$6) }' /etc/passwd
```
yuumasato, line 7: Besides it setting the permission to `700` instead of `750`.
The remediation is changing the permissions of all files, but I think it should only change the permission of files that were not correct. If a file had a more strict permission than `700` (`750`) it would be loosened.

I think the same applies to Ansible.

yuumasato (changes requested):
Looks good in general, I have concerns about the scope of interactive users and behavior of a few of the remediations.

linux_os/guide/system/accounts/accounts-session/accounts_users_home_files_groupownership/bash/shared.sh
marcusburghardt: Hi @yuumasato , looks the rule you referred is written this way because there is an exception for primary groups in some distros. For example, since RHEL and Ubuntu, by default, automatically create a new group with the same name of the respective new user and set it as primary group for that user, the SLES default approach is to include new users in a shared group called `users`. This rule covers these two approaches since it relies on the primary group informed on `/etc/passwd`.

This also makes sense in a context of other rules from the same benchmark, like:
https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-204470?version=V3R4
https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-204473?version=V3R4

So, in an environment in compliance with these three rules (204470, 204472 and 204473), for example, from the same benchmark means that only the owner can `write` on files under its home directory (makes sense). In this environment it is also possible that other users may `read` files under its home directory (despite the impact on confidentiality, makes sense in some environments, as mentioned above). In this case, even if the group-owner of files under a home directory are different from the primary group, they wouldn't be accessible due the rules `204470` and `204473`. For that make sense the permissions and/or ownership of the home directory itself should also be changed, conflicting with other rules from the same benchmark.

In this context I defend the interpretation that the files should be group-owned by the user primary group and the description is only confused due to the case of distros which share the primary group among regular users. Otherwise, the other two rules (maybe more) should be updated in the benchmark description. But this would clearly impact in less security for the system and doesn't look to be the intention.

yuumasato: I agree with you, thank you for the detailed explanation.

linux_os/guide/system/accounts/accounts-session/accounts_users_home_files_groupownership/ansible/shared.yml
yuumasato: @marcusburghardt I see you resolved this conversation but I don't see any changes.

linux_os/guide/system/accounts/accounts-session/accounts_users_home_files_groupownership/oval/shared.xml
marcusburghardt: https://github.com/ComplianceAsCode/content/pull/7790#discussion_r744794969

linux_os/guide/system/accounts/accounts-session/accounts_users_home_files_permissions/ansible/shared.yml
marcusburghardt: Hi @yuumasato , thanks for alerting me on that. Indeed, the remediation can be improved. I will update the PR soon.

linux_os/guide/system/accounts/accounts-session/accounts_users_home_files_permissions/bash/shared.sh
marcusburghardt: Update is cumming. : ))

marcusburghardt: Update is coming. : ))

linux_os/guide/system/accounts/accounts-session/accounts_users_home_files_ownership/oval/shared.xml
```diff
@@ -0,0 +1,52 @@
...
+    {{{ oval_metadata("All User Files and Directories In The Home Directory Must Have a Valid Owner") }}}
+    <criteria>
+      <criterion test_ref="test_accounts_users_home_files_ownership"
+                 comment="All User Files and Directories In The Home Directory Must Have a Valid Owner"/>
+    </criteria>
+  </definition>
+
+  <unix:password_object id="object_accounts_users_home_files_ownership_objects" version="1">
+    <unix:username datatype="string" operation="not equal">nobody</unix:username>
+    <filter action="include">state_file_ownership_home_directories_interactive_uids</filter>
```
yuumasato, line 12: State `state_file_ownership_home_directories_interactive_uids` is not defined anywhere.

linux_os/guide/system/accounts/accounts-session/accounts_users_home_files_groupownership/bash/shared.sh
```diff
@@ -0,0 +1,12 @@
...
+# reboot = false
+# strategy = restrict
+# complexity = low
+# disruption = low
+
+for user in $(awk -F':' '{ if ($4 >= {{{ gid_min }}} && $4 != 65534) print $1 }' /etc/passwd); do
+    home_dir=$(getent passwd $user | cut -d: -f6)
+    group=$(getent passwd $user | cut -d: -f4)
+    # Only update the group-ownership when necessary.
+    find $home_dir -not -group $group -exec chgrp -f $group {} \;
```
yuumasato, line 11: I wonder if this is necessary, wasn't the previous approach enough?
`awk -F':' '{ if ($4 >= {{{ gid_min }}} && $4 != 65534) system("chgrp -Rf " $4" "$6) }' /etc/passwd`

linux_os/guide/system/accounts/accounts-session/accounts_users_home_files_ownership/bash/shared.sh
```diff
@@ -0,0 +1,11 @@
...
+# reboot = false
+# strategy = restrict
+# complexity = low
+# disruption = low
+
+for user in $(awk -F':' '{ if ($3 >= {{{ uid_min }}} && $3 != 65534) print $1 }' /etc/passwd); do
+    home_dir=$(getent passwd $user | cut -d: -f6)
+    # Only update the ownership when necessary.
+    find $home_dir -not -user $user -exec chown -f $user {} \;
+done
```
yuumasato, line 11: I wonder if this is necessary, wasn't the previous approach enough?
In the end all files' ownership need to converge to the primary group, no?
Doing `find` seems more costly then doing `chwon -Rf`

linux_os/guide/system/accounts/accounts-session/accounts_users_home_files_permissions/bash/shared.sh
```diff
@@ -0,0 +1,9 @@
+# platform = multi_platform_all
+# reboot = false
+# strategy = restrict
+# complexity = low
+# disruption = low
+
+for home_dir in $(awk -F':' '{ if ($4 >= {{{ uid_min }}} && $4 != 65534) print $6 }' /etc/passwd); do
+    find $home_dir -perm /027 -exec chmod g-w,o=- {} \;
```
yuumasato, line 8: Could we skip the `find` and just do `chmod -R g-w,o=-`?

Previous approach was doing the following, which could loosen permissions unintentionally.
` awk -F':' '{ if ($4 >= {{{ uid_min }}} && $4 != 65534) system("chmod -Rf 700 "$6) }' /etc/passwd`

But the following will just remove the offending permissions.
`awk -F':' '{ if ($4 >= {{{ uid_min }}} && $4 != 65534) system("chmod -Rf g-w,o=- "$6) }' /etc/passwd`

yuumasato (changes requested):
Just have a few comments about the remediations, maybe we can avoid doing `find` which can be costly.

linux_os/guide/system/accounts/accounts-session/accounts_users_home_files_groupownership/bash/shared.sh
yuumasato: Good point, thank you for the clarification and comment in the remediation.

yuumasato (approved):
@marcusburghardt great work!

Just pending CI for merge.

PR #7826, opened 2021-11-03
Upgrade to F34 in Gating

Mab879:
#### Description:

Move to Fedora 34 for gating

#### Rationale:

F33 will EOL soon.

### DIFF
--- a/.github/workflows/gate.yaml
+++ b/.github/workflows/gate.yaml
@@ -108,3 +108,36 @@ jobs:
       - name: Test
         run: ctest -j2 --output-on-failure -E unique-stigids
         working-directory: ./build
+
+  validate-fedora-rawhide:
+    name: Build, Test on Fedora Rawhide (Container)
+    runs-on: ubuntu-latest
+    container:
+      image: fedora:rawhide
+    steps:
+      -   name: Install Deps
+          run: dnf install -y cmake make openscap-utils python3-pyyaml python3-jinja2 bats python3-pytest python3-pytest-cov ansible python3-pip
+      -   name: Install deps python
+          run: pip install ruamel.yaml yamlpath
+      -   name: Checkout
+          uses: actions/checkout@v2
+      -   name: Build
+          run: |-
+            ./build_product \
+                chromium \
+                fedora \
+                firefox \
+                fuse6 \
+                rhcos4 \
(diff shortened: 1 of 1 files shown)


Mab879:
The issue in F35 appears to relate to Jinja2 3.0.x release.

.github/workflows/gate.yaml
```diff
@@ -75,10 +75,10 @@ jobs:
         working-directory: ./build
 
   validate-fedora:
-    name: Build, Test on Fedora 33 (Container)
+    name: Build, Test on Fedora 34 (Container)
```
yuumasato, line 78: When the job name changes the required test name in the project's settings need to be changed as well.
Should we remove the version from the name of the job?

Mab879: Sounds like a decent plan to me.

evgenyz: It will make it less obvious which F was used to execute the test. I'm not insisting if people who work with the content closer than me are not against this change, just explaining my decision to include it there in the first place.

evgenyz: NB: We could probably benefit from having a rolling Rawhide-based *optional* test to embrace things to come.

Mab879: Keeping it in the title is fine. Its only a once every month or year change. 

I have also added a raw hide test. However, that will fail since we not compatible with the new version of Jinja.

jan-cerny:
@Mab879 @yuumasato Is there anything that blocks us from merging?

yuumasato:
> @Mab879 @yuumasato Is there anything that blocks us from merging?

@jan-cerny I think this PR is good to go.
The only thing preventing merge is updating the Test this repo considers required, F34 instead of F33.
Changing the setting will probably require that F34 test be ran again on open PRs, but I think this wlll require a rebase of the PRs....

I think this will also affect the stabilization branch, as it won't have the F34 job configured.
We can schedule the gating switch to F34 after the release is done on Nov 25th.
Thoughts?

Mab879:
Is there a way we can mark the Rawhide job as as optional/allowed to fail? There is issue with the new Jinja in F35+ that seems to be causing some issues for us.

yuumasato (approved):

PR #7829, opened 2021-11-04
OCP/PCI-DSS: Add responses for Req-6.1 and Req-6.2

JAORMX:
Req-6.1 is an organizational control and therefore not applicable to OpenShift.

Req-6.2 can be met (in a way) inherently because OpenShift offers automatic
updates for both the platform and operators.


sherine-k (approved):

jhrozek (approved):
LGTM

PR #7832, opened 2021-11-04
Implemented buildsystem profiling

j-ode:
#### Description:

- For developers to be able to see how their changes affect the speed of the buildsystem, we need some sort of a profiling tool. I created the build_profiler.sh and build_profiler_report.py scripts located in the utils folder to make this a simple process.
- These tools run automatically from the .build_product script located in the content folder simply by adding the '-p' or '--profiling' flag. The build is then done using ninja with jobs limited to 1 and the output is saved in a .ninja_log, which is then placed in the .build_profiling directory inside a directory named after the products that were built. The file is numbered starting from 0 (the baseline reading).
- After that, the .build_profiler_report.py script is called with the logfile name and it parses the data from it and prints it to the CLI. It also creates a .webtreemap file, which is then used to generate an HTML file with the same number as the logfile. This interactive HTML can be used to quickly see what are the biggest items prolonging the build time. 

#### Rationale:

- We need a way to
1. See what build targets take the most time to finish
2. What changes to the build time did our code cause
and this tool helps with that.

### DIFF
--- a/build_product
+++ b/build_product
@@ -10,6 +10,7 @@
 # ARG_OPTIONAL_BOOLEAN([ansible-playbooks],[],[Build Ansible Playbooks for every profile],[on])
 # ARG_OPTIONAL_BOOLEAN([bash-scripts],[],[Build Bash remediation scripts for every profile],[on])
 # ARG_OPTIONAL_BOOLEAN([datastream-only],[],[Build the datastream only. Do not build any of the guides, tables, etc],[off])
+# ARG_OPTIONAL_BOOLEAN([profiling],[p],[Use ninja and call the build_profiler.sh util],[off])
 # ARG_USE_ENV([ADDITIONAL_CMAKE_OPTIONS],[],[Whitespace-separated string of arguments to pass to CMake])
 # ARG_POSITIONAL_INF([product],[Products to build, ALL means all products],[0],[ALL])
 # ARG_DEFAULTS_POS([])
@@ -59,7 +60,7 @@ builder_type()
 
 begins_with_short_option()
 {
-	local first_option all_short_options='objh'
+	local first_option all_short_options='objph'
 	first_option="${1:0:1}"
 	test "$all_short_options" = "${all_short_options/$first_option/}" && return 1 || return 0
 }
@@ -82,7 +83,7 @@ _arg_profiling="off"
 print_help()
 {
 	printf '%s\n' "Wipes out contents of the 'build' directory and builds only and only the given products."
-	printf 'Usage: %s [-o|--oval <VERSION>] [-b|--builder <BUILDER>] [-j|--jobs <arg>] [--(no-)debug] [--(no-)derivatives] [--(no-)ansible-playbooks] [--(no-)bash-scripts] [--(no-)datastream-only] [-h|--help] [<product-1>] ... [<product-n>] ...\n' "$0"
+	printf 'Usage: %s [-o|--oval <VERSION>] [-b|--builder <BUILDER>] [-j|--jobs <arg>] [--(no-)debug] [--(no-)derivatives] [--(no-)ansible-playbooks] [--(no-)bash-scripts] [--(no-)datastream-only] [-p|--(no-)profiling] [-h|--help] [<product-1>] ... [<product-n>] ...\n' "$0"
 	printf '\t%s\n' "<product>: Products to build, ALL means all products (defaults for <product>: 'ALL')"
 	printf '\t%s\n' "-o, --oval: OVAL version. Can be one of: '5.10', '5.11' and 'auto' (default: 'auto')"
 	printf '\t%s\n' "-b, --builder: Builder engine. Can be one of: 'make', 'ninja' and 'auto' (default: 'auto')"
@@ -92,8 +93,8 @@ print_help()
 	printf '\t%s\n' "--ansible-playbooks, --no-ansible-playbooks: Build Ansible Playbooks for every profile (on by default)"
 	printf '\t%s\n' "--bash-scripts, --no-bash-scripts: Build Bash remediation scripts for every profile (on by default)"
 	printf '\t%s\n' "--datastream-only, --no-datastream-only: Build the datastream only. Do not build any of the guides, tables, etc (off by default)"
+	printf '\t%s\n' "-p, --profiling, --no-profiling: Use ninja and call the build_profiler.sh util (off by default)"
 	printf '\t%s\n' "-h, --help: Prints help"
-	printf '\t%s\n' "-p, --profiling: Use ninja and call the build_profiler.sh util"
 	printf '\nEnvironment variables that are supported:\n'
 	printf '\t%s\n' "ADDITIONAL_CMAKE_OPTIONS: Whitespace-separated string of arguments to pass to CMake."
 
@@ -160,6 +161,18 @@ parse_commandline()
 				_arg_datastream_only="on"
 				test "${1:0:5}" = "--no-" && _arg_datastream_only="off"
 				;;
+			-p|--no-profiling|--profiling)
+				_arg_profiling="on"
+				test "${1:0:5}" = "--no-" && _arg_profiling="off"
+				;;
+			-p*)
+				_arg_profiling="on"
+				_next="${_key##-p}"
+				if test -n "$_next" -a "$_next" != "$_key"
+				then
+					{ begins_with_short_option "$_next" && shift && set -- "-p" "-${_next}" "$@"; } || die "The short option '$_key' can't be decomposed to ${_key:0:2} and -${_key:2}, because ${_key:0:2} doesn't accept value and '-${_key:2:1}' doesn't correspond to a short option."
+				fi
+				;;
 			-h|--help)
 				print_help
 				exit 0
@@ -167,9 +180,6 @@ parse_commandline()
 			-h*)
 				print_help
 				exit 0
-				;;
-			-p|--profiling)
-				_arg_profiling="on"
 				;;
 			*)
 				_last_positional="$1"
--- a/build_product
+++ b/build_product
@@ -76,6 +76,7 @@ _arg_derivatives="off"
 _arg_ansible_playbooks="on"
 _arg_bash_scripts="on"
 _arg_datastream_only="off"
+_arg_profiling="off"
 
 
 print_help()
@@ -92,6 +93,7 @@ print_help()
 	printf '\t%s\n' "--bash-scripts, --no-bash-scripts: Build Bash remediation scripts for every profile (on by default)"
 	printf '\t%s\n' "--datastream-only, --no-datastream-only: Build the datastream only. Do not build any of the guides, tables, etc (off by default)"
 	printf '\t%s\n' "-h, --help: Prints help"
+	printf '\t%s\n' "-p, --profiling: Use ninja and call the build_profiler.sh util"
 	printf '\nEnvironment variables that are supported:\n'
 	printf '\t%s\n' "ADDITIONAL_CMAKE_OPTIONS: Whitespace-separated string of arguments to pass to CMake."
 
@@ -165,6 +167,9 @@ parse_commandline()
 			-h*)
 				print_help
 				exit 0
+				;;
+			-p|--profiling)
+				_arg_profiling="on"
 				;;
 			*)
 				_last_positional="$1"
@@ -336,10 +341,13 @@ test "$_arg_debug" = on && build_type_option="-DCMAKE_BUILD_TYPE=Debug"
 jobs="$_arg_jobs"
 test "$jobs" = auto && jobs=$cores
 
-if test "$_arg_builder" = make; then
-	build_with_make
+if test "$_arg_profiling" = on; then
+	build_with_ninja
+	jobs=1
 elif test "$_arg_builder" = ninja; then
 	build_with_ninja
+elif test "$_arg_builder" = make; then
+	build_with_make
 else
 	autodetect_builder
 fi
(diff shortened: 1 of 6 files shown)


utils/build_profiler.sh
```diff
@@ -0,0 +1,70 @@
...
+    >&2 echo "Changing to the $product_string directory failed"
+    exit 1
+fi
+
+# Set log_number
+if [ -f "0.ninja_log" ]; then
+    # set log number to the number of the latest log file and add 1
+    log_number=$(ls *.ninja_log | cut -d'.' -f1 | sort -rn | head -n 1)
+    ((log_number=log_number+1))
+    echo "-----\nComparing logfile $log_number to baseline 0..."
```
jan-cerny, line 45: This will print the `\n` literally instead of a newline. You should use `echo -e` or maybe `printf`.

utils/build_profiler_report.py
```diff
@@ -0,0 +1,166 @@
...
+
+    # sort targets by % taken of build time
+    temp = {k: v for k, v in sorted(CurrentDict.items(), key=lambda item: item[1], reverse=True)}
+    CurrentDict = temp
+
+    for target in CurrentDict.keys():
+        # percentage of build time that the target took
+        perc = CurrentDict[target]/total_time_current * 100
+
+        # difference between perc in current and in baseline
```
jan-cerny, line 97: Maybe a difference in seconds would be more interesting or interesting as well?

utils/build_profiler_report.py
```diff
@@ -0,0 +1,166 @@
...
+def PrintReport(CurrentDict: dict, BaselineDict: dict = None) -> None:
+    """Print report with results of profiling to stdout"""
+
+    # If the targets have changed between baseline and current, we are using total_time_intersect
+    # to calculate delta (ratio of durations of targets) instead of total_time
+    if BaselineDict and BaselineDict.keys() != CurrentDict.keys():
+        print("Warning: the targets in the current logfile differ from those in the baseline\
+             logfile; therefore the Δ delta is calculated without taking the added/removed\
+                  targets into consideration. If the added/removed targets modify the behavior of\
+                      targets in both logfiles, the Δ delta may not make sense.\n-----")
```
jan-cerny, line 75: Coding style! You don't have to use `\` if you're inside parenthesis, you can put `"` at the end of the line, and then `"` at the beginning of the new line (after leading spaces), and the strings will get concatenated.

utils/build_profiler_report.py
```diff
@@ -0,0 +1,166 @@
...
+        delta = 0
+        if BaselineDict:
+            if target_mismatch:
+                if target in BaselineDict.keys():
+                    delta = CurrentDict[target]/total_time_current_intersect * 100 - \
+                        BaselineDict[target]/total_time_baseline_intesect * 100
+
+            else:
+                delta = perc - (BaselineDict[target]/total_time_baseline * 100)
+
```
jan-cerny, line 107: Try to extract this to a function.

build_product
```diff
@@ -166,6 +168,9 @@ parse_commandline()
 				print_help
 				exit 0
 				;;
+			-p|--profiling)
```
jan-cerny, line 171: We should add your new feature to the documentation. You can reuse / get inspired by the text that you wrote in the pull request description.

utils/build_profiler.sh
```diff
@@ -0,0 +1,70 @@
...
+
+# parse and compare log files; show results to user
+../../utils/build_profiler_report.py "$log_number.ninja_log"
+if [ ! -f "$log_number.ninja_log.webtreemap" ]; then
+    >&2 echo "utils/build_profiler_report.py failed to create $log_number.ninja_log.webtreemap"
+    exit 1
+fi
+
+# create HTML from .webtreemap file
+webtreemap -o "$log_number.html" < "$log_number.ninja_log.webtreemap"
```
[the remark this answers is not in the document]
jan-cerny: How can I install it?

jan-cerny (commented):
It's amazing. It's better than I expected. I like the idea of comparing 2 builds.

I have only minor problems. 

I recommend focusing on Python coding style.

utils/build_profiler_report.py
```diff
@@ -0,0 +1,166 @@
...
+def FormatTime(t):
+    """Converts a time into a human-readable format"""
+
+    t /= 1000
+    if t < 60:
+        return '%.1fs' % t
+    if t < 60 * 60:
+        return '%dm%.1fs' % (t / 60, t % 60)
+    return '%dh%dm%.1fs' % (t / (60 * 60), t % (60 * 60) / 60, t % 60)
```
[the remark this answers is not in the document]
matejak: A difference between two times is [timedelta](https://docs.python.org/3.8/library/datetime.html#datetime.timedelta) that provides a human-readable string representation that you can't tune, but perhaps it could be good enough.

utils/build_profiler.sh
```diff
@@ -0,0 +1,70 @@
...
+if [ "$#" -ne 1 ]; then
+    echo "build_profiler.sh requires one argument - product_string"
+    exit 1
+fi
+
+product_string="$1"
+
+# Create and change to .build_profiling dir 
+if [ ! -d ".build_profiling" ]; then
+    mkdir .build_profiling
```
matejak, line 12: It is a good idea to have a `die` function defined in your script, and then just to write

```
mkdir .build_profiling || die "Creating the .build_profiling directory failed"
```

utils/build_profiler_report.py
```diff
@@ -0,0 +1,166 @@
...
+    with open(file, 'r') as fp:
+        lines = fp.read().splitlines()
+
+    # {Target: duration} dict
+    TargetDurationDict = {}
+    for line in lines:
+        line = line.split()
+
+        # ToDo - create better comment detection
+        if line[0][0] != '#':
```
matejak, line 17:
```suggestion
        line = line.strip()
        if line.startswith('#'):
            continue
        else:
```

utils/build_profiler_report.py
```diff
@@ -0,0 +1,166 @@
...
+    if t < 60:
+        return '%.1fs' % t
+    if t < 60 * 60:
+        return '%dm%.1fs' % (t / 60, t % 60)
+    return '%dh%dm%.1fs' % (t / (60 * 60), t % (60 * 60) / 60, t % 60)
+
+
+def GenerateWebTreeMap(CurrentDict: dict, BaselineDict: dict) -> None:
+    """Create file for webtreemap to generate an HTML from; if target is new, append _NEW"""
+    with open(sys.argv[1] + ".webtreemap", 'w') as fp:
```
matejak, line 39: Don't force your function to get input from parsing arguments, delegate that to another layer and let that function to accept a string as an input.

utils/build_profiler_report.py
matejak: I would suggest to create a variable holding the message first, and then just print the variable.

Also I would use D or delta instead of the greek capital letter, that way nobody will scratch their heads because of mysterious triangles, and you will avoid potential encoding issues.

utils/build_profiler_report.py
```diff
@@ -0,0 +1,166 @@
...
+
+def main() -> None:
+    # Dict key order used by PrintReport only specified in 3.7.0+
+    if sys.version_info < (3, 7, 0):
+        sys.stderr.write("You need python 3.7 or later to run this script\n")
+        exit(1)
+
+    if len(sys.argv) != 2:
+        print("Incorrect number of arguments")
+        sys.exit(99)
```
matejak, line 152: Simply use argparse, it gives you standardized interface, meaningful separation of code into layers, and help in one package.

utils/build_profiler_report.py
```diff
@@ -0,0 +1,166 @@
...
+
+    # skip loading baseline if we are creating it
+    if sys.argv[1] != "0.ninja_log":
```
matejak, line 155: This looks strange, behavior of the program shouldn't depend on the exact filename.

utils/build_profiler_report.py
j-ode: I don't see the reason for moving this into a function, considering the fact that I would have to pass 7 variables to such function. That would just make the code less readable in my opinion.

jan-cerny:
@j-ode You have marked most of the items as resolved, but you should mark them as resolved only after you push the commits that fix the reviewer's problems or after you persuade them to change their opinion. :smiley:

utils/build_profiler_report.py
```diff
@@ -0,0 +1,166 @@
...
+             logfile; therefore the Δ delta is calculated without taking the added/removed\
+                  targets into consideration. If the added/removed targets modify the behavior of\
+                      targets in both logfiles, the Δ delta may not make sense.\n-----")
+        target_mismatch = True
+        total_time_current_intersect = GetTotalTimeIntersect(CurrentDict, BaselineDict)
+        total_time_baseline_intesect = GetTotalTimeIntersect(BaselineDict, CurrentDict)
+    else:
+        target_mismatch = False
+
+    header = [f'{"Target:":60}', f"{'%':4}", f"{'Δ':5}", f"{'T':8}", f"{'%Δ':5}", "Note"]
```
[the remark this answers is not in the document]
j-ode: The column seems to contain both filenames (rhel8/shorthand.xml) and directory names (rhel8/profiles), so renaming it to filename wouldn't make much sense, so I will keep it as "Target".

utils/build_profiler.sh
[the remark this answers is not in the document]
j-ode: Done, the scripts checks if the command is installed and if not, gives you a hint on how to install it and also gives you the entire command to generate the HTML after you install it.

utils/build_profiler_report.py
j-ode: good idea, the script now prints the difference in seconds as well in addition to the percentage change in time

build_product
j-ode: I added a section called "Profiling the buildsystem" to files 2 and 5 of the documentation, with the former containing a brief overview of how to use the -p|--profiling switch with the build_product script and what it is used for, while the latter contains details on how the two helper scripts build_profiler.sh and build_profiler_report.py actually work.

utils/build_profiler_report.py
j-ode: I delegated the checking for the baseline logfile to the build_profiler.sh script and added a "--baseline" flag to the build_profiler_report.py script which is set by build_profiler.sh if the baseline logfile is missing. When the flag is set, the build_profiler_report.py script does not compare the current logfile to a baseline logfile.

docs/manual/developer/02_building_complianceascode.md
```diff
@@ -446,6 +446,16 @@ ctest -LE quick -j4
 
 Note: CTest does not run [SSG Test Suite](https://github.com/ComplianceAsCode/content/tree/master/tests) which provides simple system of test scenarios for testing profiles and rule remediations.
 
+## Profiling the buildsystem
+
+To make sure your changes don't prolong the time that it takes to build products by using the `build_product` script too much,
```
jan-cerny, line 451: We now prefer style "1 line = 1 sentence", because people won't have to re-wrap when modyfing the text.

utils/build_profiler.sh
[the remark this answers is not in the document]
jan-cerny: Thanks, the message is amazing.

However, I'm new to the javascript applications, so I'm still unable to run it. I have successfully executed `sudo npm i webtreemap`, but then I can't execute the webtreemap command.

```
[jcerny@thinkpad scap-security-guide{pr/7832}]$ webtreemap -o .build_profiling/rhel8/1.html < .build_profiling/rhel8/1.ninja_log.webtreemap
bash: webtreemap: command not found...
```

I assume that some path needs to be added to my environment for the shell to find it. Is that correct? Or how can I enable it?

utils/build_profiler_report.py
```diff
@@ -0,0 +1,199 @@
...
+        if target_mismatch:
+            intersect_time_delta = total_time_current_intersect - total_time_baseline_intesect
+            if abs(intersect_time_delta) < 60:
+                intersect_time_delta = 0
+            intersect_time_delta = format_time(intersect_time_delta)
+            line.append(f'| intersect TD {intersect_time_delta:>8}')
+            intersect_perc_time_delta = (total_time_current_intersect /
+                                         total_time_baseline_intesect) * 100 - 100
+            line.append(f'| intersect %TD {intersect_perc_time_delta:+5.1f}')
+        print(' '.join(line))
```
jan-cerny, line 158: But, for some reason, now I can see everything twice, there are always 2 items that differ only by the name, but they obviously represent the same CMake target. For example

```
/home/jcerny/work/git/scap-security-guide/build/rhel8/collect-remediations-rhel8 |  6.6 |  -0.3 |    16.6s |    -1.4s |  -8.0
rhel8/collect-remediations-rhel8                             |  6.6 |  -0.3 |    16.6s |    -1.4s |  -8.0
```

Have you experienced this?

jan-cerny (commented):
I like it, the coding style improved greatly and the documentation is just amazing!

utils/build_profiler_report.py
jan-cerny: I use cmake-3.22.1-5.fc35.x86_64.

j-ode: To address the comment above (For some reason I can't reply to it), my npm packages get installed in /usr/local/bin, so try adding that to your PATH.

Concerning the duplicate targets, the cmake version shouldn't matter, as the profiling works with ninja. My ninja version is 1.10.2, which is the latest according to https://github.com/ninja-build/ninja/releases. Try running ninja --version and try updating to the latest version.

jan-cerny: Can you point me to some guide/docs? I can see some webtreemap only in /home/jcerny/node_modules/webtreemap but nothing like the in /usr/local/bin.

jan-cerny: my ninja is ninja-build-1.10.2-6.fc35.x86_64

jan-cerny: I have chechekd and it's duplicate also in ninja log. So it doesn't happen for you, right? Why do you think that CMake shouldn't matter? CMake is the system that defines the build targets.

jan-cerny (approved):

PR #7834, opened 2021-11-05
OCP: Ensure `ebs_encryption_enabled_on_machinesets` on AWS only

JAORMX:
Cloud-specific CPEs were recently introduced [1][2], and so, let's take
the AWS one into use so the aforementioned rule will only run on AWS.

As a result, if the rule is ran on Azure or another cloud, it'll be
marked as `NOT-APPLICABLE` in the result.

A subsequent commit will add the GCP and Azure rules to the disk
encryption catch-all rule `machine_volume_encrypted`.

[1] Commit 01b3ecdb84a92630c6f07fd9882a4a1eb4e7afe3
[2] Commit 12eb1bc8d1f3f6e21d45206bed21b185d74e9bb9


sherine-k (approved):
Nice one! 
Didn't know you could do that

Vincent056 (approved):
/lgtm

PR #7838, opened 2021-11-05
Add a POC handling of expressions based on simpleeval

matejak:
The code shows simpleeval parsing an expression into a tree.

The code allows for construction of instances that can be defined in the code, and then passed to the expression evaluator. The evaluator can then evaluate a user-supplied expression using Python syntax and referencing those instances. Result of the evaluation is a tree structure with and/or container nodes and leaf nodes. The idea is that the resulting tree can be used as an input for all sorts of text generators.

### DIFF
--- a/ssg/boolean_handling.py
+++ b/ssg/boolean_handling.py
@@ -1,0 +1,169 @@
+import ast
+import copy
+
+
+class GraphNode:
+    """
+    Base for both AND/OR or leaf nodes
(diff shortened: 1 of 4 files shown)


jan-cerny:
Is this going to be replaced by #7873?

matejak:
> Is this going to be replaced by #7873?

Yes, exactly.

PR #7843, opened 2021-11-08
Fix compatibility with python < 3.5

yuumasato:
#### Description:

- This change ensures that unmatched groups are substituted for empty strings.

#### Rationale:

- Up to python 3.5, substituting an unmatched group would case a unmatched group error.
   From python 3.5 an unmatched group is automatically substituted by an empty string.
  Ref: https://bugs.python.org/issue1519638


vojtapolasek (approved):
LGTM, thanks for the fix.

PR #7851, opened 2021-11-09
Add Delta Tailoring Files to the Build System

Mab879:
#### Description:

Builds the delta tailoring files for RHEL7 and RHEL8.

#### Rationale:

So that we can releases these files to downstream.


cmake/SSGCommon.cmake
```diff
@@ -761,6 +761,32 @@ macro(ssg_make_all_tables PRODUCT)
     )
 endmacro()
 
+macro(ssg_build_rule_dir_json)
+    add_custom_command(
+        OUTPUT "${CMAKE_BINARY_DIR}/rule_dirs.json"
+        COMMAND env "PYTHONPATH=$ENV{PYTHONPATH}" "${PYTHON_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/utils/rule_dir_json.py" --output "${CMAKE_BINARY_DIR}/rule_dirs.json" --root "${CMAKE_SOURCE_DIR}"
```
yuumasato, line 767: @jan-cerny @matejak Thoughts on generating the `build/rule_dirs.json` by default?

I know there were discussions about this in the past, but couldn't find them.

cmake/SSGCommon.cmake
```diff
@@ -761,6 +761,32 @@ macro(ssg_make_all_tables PRODUCT)
...
+endmacro()
+
+macro(ssg_build_disa_delta PRODUCT PROFILE)
+    file(GLOB DISA_SCAP_REF "${SSG_SHARED_REFS}/disa-stig-${PRODUCT}-v[0-9]*r[0-9]*-xccdf-scap.xml")
+        add_custom_command(
+            OUTPUT "${CMAKE_BINARY_DIR}/${PRODUCT}_{$PROFILE}_delta_tailoring.xml"
+            COMMAND env "PYTHONPATH=$ENV{PYTHONPATH}" "${PYTHON_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/utils/create_scap_delta_tailoring.py" --root "${CMAKE_SOURCE_DIR}" --product "${PRODUCT}" --manual "${DISA_SCAP_REF}" --json "${CMAKE_BINARY_DIR}/rule_dirs.json" --profile "${PROFILE}" --reference "stigid" --output "${CMAKE_BINARY_DIR}/${PRODUCT}_${PROFILE}_delta_tailoring.xml"
+            DEPENDS "${CMAKE_BINARY_DIR}/ssg-${PRODUCT}-ds.xml"
+            DEPENDS "${CMAKE_BINARY_DIR}/rule_dirs.json"
+            COMMENT "[${PRODUCT}-generate-ssg-delta] generating disa tailering file"
```
yuumasato, line 782: Typo: `tailering`

cmake/SSGCommon.cmake
```diff
@@ -761,6 +761,32 @@ macro(ssg_make_all_tables PRODUCT)
...
+        COMMENT "[rule_dir_json] generating rule_dirs.json"
+    )
+    add_custom_target("generate-ssg-rule_dir_json"
+        DEPENDS "${CMAKE_BINARY_DIR}/rule_dirs.json"
+    )
+endmacro()
+
+macro(ssg_build_disa_delta PRODUCT PROFILE)
+    file(GLOB DISA_SCAP_REF "${SSG_SHARED_REFS}/disa-stig-${PRODUCT}-v[0-9]*r[0-9]*-xccdf-scap.xml")
+        add_custom_command(
```
yuumasato, line 777: Indentantion seems to be off.

cmake/SSGCommon.cmake
```diff
@@ -761,6 +761,32 @@ macro(ssg_make_all_tables PRODUCT)
...
+            OUTPUT "${CMAKE_BINARY_DIR}/${PRODUCT}_{$PROFILE}_delta_tailoring.xml"
+            COMMAND env "PYTHONPATH=$ENV{PYTHONPATH}" "${PYTHON_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/utils/create_scap_delta_tailoring.py" --root "${CMAKE_SOURCE_DIR}" --product "${PRODUCT}" --manual "${DISA_SCAP_REF}" --json "${CMAKE_BINARY_DIR}/rule_dirs.json" --profile "${PROFILE}" --reference "stigid" --output "${CMAKE_BINARY_DIR}/${PRODUCT}_${PROFILE}_delta_tailoring.xml"
```
yuumasato, line 779: How about outputting the tailoring to `${CMAKE_BINARY_DIR}/${PRODUCT}/tailoring/${PRODUCT}_${PROFILE}_delta_tailoring.xml` ?

cmake/SSGCommon.cmake
```diff
@@ -761,6 +761,32 @@ macro(ssg_make_all_tables PRODUCT)
...
+            DEPENDS "${CMAKE_BINARY_DIR}/ssg-${PRODUCT}-ds.xml"
+            DEPENDS "${CMAKE_BINARY_DIR}/rule_dirs.json"
+            COMMENT "[${PRODUCT}-generate-ssg-delta] generating disa tailering file"
+         )
+
+        add_custom_target(generate-ssg-delta-${PRODUCT}-${PROFILE}
+            DEPENDS "${CMAKE_BINARY_DIR}/${PRODUCT}_{$PROFILE}_delta_tailoring.xml"
```
yuumasato, line 786: There is a typo (`{$PROFILE}`) which causes `create_scap_detal_tailoring.py` to run everytime.

yuumasato (changes requested):
You also need to add `install` commands so that the tailoring file is moved to its directory during package install.
Check out https://github.com/ComplianceAsCode/content/blob/master/CMakeLists.txt#L28 for inspiration.
I'm thinking a good path might be: `"${CMAKE_INSTALL_DATADIR}/scap-security-guide/tailoring"`

I'm also thinking that the whole output of `create_scap_delta_tailoring.py` is unnecessary when it is plugged into the build system. Maybe add an option to make it silent?

cmake/SSGCommon.cmake
jan-cerny: My concern would be that the `utils/rule_dir_json.py` is very slow, ~ 12 s on my beef laptop.

cmake/SSGCommon.cmake
```diff
@@ -761,6 +761,36 @@ macro(ssg_make_all_tables PRODUCT)
...
+            DEPENDS "${CMAKE_BINARY_DIR}/rule_dirs.json"
+            DEPENDS "${PRODUCT}-content"
+            COMMENT "[${PRODUCT}-generate-ssg-delta] generating disa tailoring file"
+         )
+
+        add_custom_target(generate-ssg-delta-${PRODUCT}-${PROFILE}
+            DEPENDS "${CMAKE_BINARY_DIR}/${PRODUCT}/tailoring/${PRODUCT}_${PROFILE}_delta_tailoring.xml"
+        )
+
+        install(FILES "${CMAKE_BINARY_DIR}/${PRODUCT}/tailoring/${PRODUCT}_{$PROFILE}_delta_tailoring.xml"
```
yuumasato, line 790: Typo again, :) `{$PROFILE}`

cmake/SSGCommon.cmake
```diff
@@ -761,6 +761,36 @@ macro(ssg_make_all_tables PRODUCT)
...
+                DESTINATION SSG_TAILORING_INSTALL_DIR)
```
yuumasato, line 791: Expand the variable `SSG_TAILORING_INSTALL_DIR`, :)

`- Installing: /tmp/SSG_TAILORING_INSTALL_DIR/rhel8_stig_delta_tailoring.xml`

cmake/SSGCommon.cmake
yuumasato: The `rule_dirs.json` is used to collect the Ids of rules applicable to the product:
https://github.com/ComplianceAsCode/content/blob/28a5fb54a949065fbed57a9c6e39e5ddc931961a/utils/create_scap_delta_tailoring.py#L56

But as we are plugging the script into the build system, we could use the data that is available as part of the build pipeline. So collecting the Ids of rules applicable to the product can done by listing the files in `build/${PRODUCT}/rules`.

But I also see the value in being able to run `create_scap_delta_tailoring.py` without doing the whole build.
So the script could either receive a `rule_dir_json.py` or a `--resolved-rules-dir`?

During the build the target `generate-ssg-delta-${PRODUCT}-${PROFILE}` would depend on `${PRODUCT}-compile-all`.

yuumasato (changes requested):

yuumasato (approved):
LGTM.

PR #7853, opened 2021-11-10
OCP/RHCOS: Mark SC-7(18) as `not applicable`

JAORMX:
This control refers to a boundary protection device, and thus is not
applicable to neither Openshift nor RHCOS. These devices are typically
installed on the cloud these are running and should be checked
separately.


Vincent056 (approved):
/lgtm

PR #7855, opened 2021-11-10
OCP/PCI-DSS: Add responses for section 7

JAORMX:
Most of these are related to RBAC and its usage in the platform.

We meet this :)


JAORMX:
Pull-request updated, HEAD is now 9ebcd0569309bd121f0d802cd4c0db53eaeec612

ggbecker:
> This datastream diff is auto generated by the check `Compare DS/Generate Diff`
> Click here to see the full diff
> 
> ```diff
> Platform has been changed for rule 'xccdf_org.ssgproject.content_rule_api_server_api_priority_flowschema_catch_all'
> --- old datastream
> +++ new datastream
> -['cpe:/a:redhat:openshift_container_platform:4.8', 'cpe:/a:redhat:openshift_container_platform:4.9', 'cpe:/a:redhat:openshift_container_platform:4.10']
> +['cpe:/a:redhat:openshift_container_platform:4.10', 'cpe:/a:redhat:openshift_container_platform:4.8', 'cpe:/a:redhat:openshift_container_platform:4.9']
> ```

we have introduced this new GH action, but I noticed two bugs here. One that is not replacing the existing comment, which it should do (fixed by https://github.com/ComplianceAsCode/content/pull/7856), the second is that the CPEs are the same but the tool is not able to identify them as the same, probably some sorting needs to happen before comparing stuff.

lmzuccarelli (commented):
Verification 
- Executed unit tests
  - rbac_cluster_roles_defined
  - rbac_roles_defined
All tests passing

controls/pcidss_ocp4.yml
```diff
@@ -1624,41 +1647,58 @@ controls:
...
     - base
-    notes: ''
-    status: pending
+    notes: |-
+      The OpenShift Container Platform has RBAC enabled by default. With this
+      feature, any requests to the API is denied unless an explicit
+      Role or ClusterRule is bound to the entity making the request.
+
+      All objects in Kubernetes/OpenShift are bound to RBAC permissions.
+    status: inherently met
```
lmzuccarelli, line 1666: More of a question about "inherently met" - this really means that "by virtue of the fact (note details) this is taken care of, but this is not automated

lmzuccarelli:
Overall LGTM. I will leave the final /lgtm to other team members

jhrozek (approved):
/lgtm

PR #7859, opened 2021-11-10
update description of grub2_uefi_password

vojtapolasek:
#### Description:

remove the part which talks about using grub2-mkconfig to update the documentation

#### Rationale:

according to documentation it is not needed

https://access.redhat.com/solutions/4575381

### DIFF
--- a/linux_os/guide/system/bootloader-grub2/non-uefi/grub2_password/rule.yml
+++ b/linux_os/guide/system/bootloader-grub2/non-uefi/grub2_password/rule.yml
@@ -25,13 +25,14 @@ description: |-
     </pre>
     NOTE: the bootloader superuser account and password MUST differ from the
     root account and password.
-    {{% endif %}}
-    <br /><br />
-    {{% if product in ["sle12", "sle15"] %}}
     Once the superuser password has been added,
     update the
     <tt>grub.cfg</tt> file by running:
-    <pre>grub2-mkconfig -o {{{ grub2_boot_path }}}/grub.cfg</pre>
+    {{% if "ubuntu" in product %}}
+    <pre>update-grub</pre>
+    {{% elif product in ["sle12", "sle15"] %}}
+    <pre>grub2-mkconfig -o {{{ grub2_uefi_boot_path }}}/grub.cfg</pre>
+    {{% endif %}}
     {{% endif %}}
 
 rationale: |-
--- a/linux_os/guide/system/bootloader-grub2/non-uefi/grub2_password/rule.yml
+++ b/linux_os/guide/system/bootloader-grub2/non-uefi/grub2_password/rule.yml
@@ -27,10 +27,12 @@ description: |-
     root account and password.
     {{% endif %}}
     <br /><br />
+    {{% if product in ["sle12", "sle15"] %}}
     Once the superuser password has been added,
     update the
     <tt>grub.cfg</tt> file by running:
     <pre>grub2-mkconfig -o {{{ grub2_boot_path }}}/grub.cfg</pre>
+    {{% endif %}}
 
 rationale: |-
     Password protection on the boot loader configuration ensures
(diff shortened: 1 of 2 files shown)


yuumasato (commented):
Should description of rule `grub2_password` be updated too?

linux_os/guide/system/bootloader-grub2/non-uefi/grub2_password/rule.yml
```diff
@@ -26,11 +26,13 @@ description: |-
     NOTE: the bootloader superuser account and password MUST differ from the
     root account and password.
     {{% endif %}}
+    {{% if "ubuntu" in product %}}
```
yuumasato, line 29: Is the text change also valid for other products?
Now the text is only applicable to Ubuntu while before it was applicable to all products.

linux_os/guide/system/bootloader-grub2/uefi/grub2_uefi_password/rule.yml
```diff
@@ -26,13 +26,11 @@ description: |-
...
     root account and password.
     {{% endif %}}
+    {{% if "ubuntu" in product %}}
     Once the superuser password has been added,
     update the
     <tt>grub.cfg</tt> file by running:
-    {{% if "ubuntu" in product %}}
     <pre>update-grub</pre>
-    {{% else %}}
-    <pre>grub2-mkconfig -o {{{ grub2_uefi_boot_path }}}/grub.cfg</pre>
```
yuumasato, line 35: Is the text change also valid to other products too? Like sle12 or wrlinux1019?
If not, instead of excluding the `else`, we should make it into a conditional that doesn't match rhel nor fedora.

yuumasato (changes requested):

vojtapolasek:
@abergmann  @teacup-on-rockingchair  @dodys  could you please check these changes if they are OK for SLE and Ubuntu?

dodys:
> @abergmann @teacup-on-rockingchair @dodys could you please check these changes if they are OK for SLE and Ubuntu?

sorry, I'm on leave today and was yesterday as well. I'll mark @richardmaciel-canonical here, to see if he can check it. Otherwise, I will check it on Monday, if that's ok.

linux_os/guide/system/bootloader-grub2/non-uefi/grub2_password/rule.yml
richardmaciel-canonical: Actually Ubuntu only uses `update-grub`, just like the UEFI case.

vojtapolasek:
@yuumasato  I think it is fixed.

linux_os/guide/system/bootloader-grub2/uefi/grub2_uefi_password/rule.yml
yuumasato: If I understand comment from @teacup-on-rockingchair correctly, configuration with `grub2-mkconfig` is valid for SLE.
So you should change the `else` to an `elif`.

linux_os/guide/system/bootloader-grub2/non-uefi/grub2_password/rule.yml
```diff
@@ -27,10 +27,12 @@ description: |-
     root account and password.
     {{% endif %}}
     <br /><br />
+    {{% if product in ["sle12", "sle15"] %}}
     Once the superuser password has been added,
     update the
     <tt>grub.cfg</tt> file by running:
     <pre>grub2-mkconfig -o {{{ grub2_boot_path }}}/grub.cfg</pre>
+    {{% endif %}}
```
yuumasato, line 35: This section is missing the `update-grub` command for `ubuntu`.
I guess you can copy it from the uefi rule.

yuumasato (changes requested):
The rule descriptions are still inconsistent.

vojtapolasek:
I think I got everything aligned now.

yuumasato (approved):

PR #7860, opened 2021-11-10
Sort CPEs before adding them to the datastream

ggbecker:
#### Description:

- Sort CPEs before adding them to the datastream.
- Do not show diff for rule security_patches_up_to_date.

#### Rationale

- Produce reproduceable builds and remove not useful information about security_patches_up_to_date in compare_ds.py tool

### DIFF
--- a/.github/workflows/compare-ds.yaml
+++ b/.github/workflows/compare-ds.yaml
@@ -47,6 +47,7 @@ jobs:
         if: ${{ steps.ctf.outputs.CTF_OUTPUT_SIZE != '0' }}
         uses: actions/checkout@v2
         with:
+          ref: ${{ github.event.pull_request.head.sha }}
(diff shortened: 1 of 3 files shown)


ggbecker:
Unfortunately changes to the workflow will only reflect when the PR gets merged as it uses `pull_request_target`. I have tested on my fork and they worked pretty well: https://github.com/ggbecker/content/pull/20#issuecomment-965700805

The workflow file similarly as it is proposed here: https://github.com/ggbecker/content/blob/master/.github/workflows/compare-ds.yaml

ggbecker:
@evgenyz can you please check this one? thanks

evgenyz (approved):

JAORMX:
@ggbecker thanks for this fix!

PR #7862, opened 2021-11-11
OCP4: Add rule for SC-5(1): enable routes rate limit

Vincent056:
This PR adds a rule: routes_rate_limit - Check if the rate limit is enabled for all routes


applications/openshift/networking/routes_rate_limit/tests/have_routes_rate_limit.pass.sh
```diff
@@ -0,0 +1,128 @@
...
+routes_apipath="apis/route.openshift.io/v1/routes?limit=500"
+
+cat <<EOF > "$kube_apipath/$routes_apipath"
+{
+    "apiVersion": "v1",
+    "items": [
+        {
+            "apiVersion": "route.openshift.io/v1",
+            "kind": "Route",
+            "metadata": {
```
JAORMX, line 20: shouldn't this fail because this doesn't have the annotation?

applications/openshift/networking/routes_rate_limit/tests/no_routes_rate_limit.pass.sh
```diff
@@ -0,0 +1,133 @@
+#!/bin/bash
```
JAORMX, line 1: why is this passing if no routes have rate limiting?

applications/openshift/networking/routes_rate_limit/rule.yml
```diff
@@ -0,0 +1,49 @@
...
+  rate-limiting annotations.
+
+  [1] https://docs.openshift.com/container-platform/4.9/networking/routes/route-configuration.html#nw-route-specific-annotations_route-configuration
+  
+rationale: |-
+  The usage of rate limit for Routes provides basic protection against distributed denial-of-service (DDoS) attacks.
+
+# TODO: @Vincent056 need to find out which cce
+#identifiers:
+#  cce@ocp4: CCE-xxx
```
JAORMX, line 17: we move them from https://github.com/ComplianceAsCode/content/blob/master/shared/references/cce-redhat-avail.txt

Since they are unique, you need to delete it from that list once you use it.

applications/openshift/networking/routes_rate_limit/rule.yml
```diff
@@ -0,0 +1,49 @@
+prodtype: ocp4
```
JAORMX, line 1: lets add this rule here: https://github.com/ComplianceAsCode/content/blob/master/controls/nist_ocp4.yml

JAORMX (changes requested):

applications/openshift/networking/routes_rate_limit/tests/no_routes_rate_limit.pass.sh
Vincent056: Because we are checking non openshift namespaces, so a new cluster will pass the test

applications/openshift/networking/routes_rate_limit/tests/have_routes_rate_limit.pass.sh
Vincent056: Because we skipped all the openshift related namespaces

applications/openshift/networking/routes_rate_limit/rule.yml
```diff
@@ -0,0 +1,49 @@
...
+
+title: Ensure that all Routes has rate limit enabled 
+
+description: |-
+  OpenShift has an option to set the rate limit for Routes [1] when creating new Routes.
+  All routes outside the openshift namespaces should use the
```
jhrozek, line 7: Also outside the kube- namespaces (nit I know)

applications/openshift/networking/routes_rate_limit/tests/have_routes_rate_limit.pass.sh
JAORMX: aaaah! Gotcha

jhrozek (commented):
Looks good! One nit inline

JAORMX (changes requested):
Let's add two more test cases:
* One where there are no routes in the system
* One where there's several routes: some of them are not compliant and some of them are compliant

controls/nist_ocp4.yml
```diff
@@ -13204,7 +13204,8 @@ controls:
     [2] https://docs.openshift.com/container-platform/4.7/applications/quotas/quotas-setting-per-project.html
     [3] https://docs.openshift.com/container-platform/4.7/applications/quotas/quotas-setting-across-multiple-projects.html
     [4] https://docs.openshift.com/container-platform/4.7/nodes/pods/nodes-pods-configuring.html#nodes-pods-configuring-bandwidth_nodes-pods-configuring
-  rules: []
+  rules: 
+  - routes_rate_limit
```
JAORMX, line 13208: could you please change the status as well? There is a description of the statuses here: https://complianceascode.readthedocs.io/en/latest/manual/developer/03_creating_content.html#reporting-status

JAORMX (changes requested):

JAORMX (approved):

PR #7872, opened 2021-11-12
OCP4:include sc-5 and sc5(2) in routes_rate_limit

Vincent056:
Only SC5 is included in FedRAMP, and since `routes_rate_limit` is also relevant with SC5(1) and SC5, we included  SC5(1) and SC5 in the rule.

### DIFF
--- a/applications/openshift/networking/routes_rate_limit/rule.yml
+++ b/applications/openshift/networking/routes_rate_limit/rule.yml
@@ -17,7 +17,7 @@ identifiers:
   cce@ocp4: CCE-90779-0
 
 references:
-  nist: SC-5(1)
+  nist: SC-5,SC-5(1),SC-5(2)
 
 {{% set jqfilter = '[.items[] | select(.metadata.namespace | startswith("kube-") or startswith("openshift-") | not) | select(.metadata.annotations["haproxy.router.openshift.io/rate-limit-connections"] == "true" | not) | .metadata.name]' %}}
 
--- a/applications/openshift/networking/routes_rate_limit/tests/ocp4/e2e.yml
+++ b/applications/openshift/networking/routes_rate_limit/tests/ocp4/e2e.yml
@@ -1,2 +1,3 @@
 ---
 default_result: PASS
+result_after_remediation: FAIL
--- a/controls/nist_ocp4.yml
+++ b/controls/nist_ocp4.yml
@@ -13172,7 +13172,8 @@ controls:
     [6] https://kubernetes.io/docs/concepts/cluster-administration/flow-control/
 
     https://issues.redhat.com/browse/CMP-427
-  rules: []
+  rules: 
+  - routes_rate_limit
   description: |-
     The information system protects against or limits the effects of the following types of denial of service attacks: [Assignment: organization-defined types of denial of service attacks or reference to source for such information] by employing [Assignment: organization-defined security safeguards].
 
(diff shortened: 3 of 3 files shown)


applications/openshift/networking/routes_rate_limit/rule.yml
```diff
@@ -17,7 +17,7 @@ identifiers:
   cce@ocp4: CCE-90779-0
 
 references:
-  nist: SC-5(1)
+  nist: SC-5, SC-5(1), SC-5(2)
```
JAORMX, line 20: can you remove the spaces from here?

JAORMX (changes requested):

Vincent056:
@JAORMX I tried this on my local cluster, it got passed, so do you think there might extra routes on e2e cluster

Vincent056:
@JAORMX Found it There is `test-route` from "namespace": "e2e-test-authentication-operator-kmtcm", do you think I should exclude `e2e-test` prefix in the rule check?
```
  {
            "apiVersion": "route.openshift.io/v1",
            "kind": "Route",
            "metadata": {
                "annotations": {
                    "openshift.io/host.generated": "true"
                },
                "creationTimestamp": "2021-11-14T08:26:12Z",
                "name": "test-route",
                "namespace": "e2e-test-authentication-operator-kmtcm",
                "resourceVersion": "37943",
                "uid": "89008422-5a03-403e-9635-7b0e8ac92da8"
            },
            "spec": {
                "host": "test-route-e2e-test-authentication-operator-kmtcm.apps.ci-op-80b910fj-1708d.origin-ci-int-aws.dev.rhcloud.com",
                "port": {
                    "targetPort": "https"
                },
                "tls": {
                    "insecureEdgeTerminationPolicy": "Redirect",
                    "termination": "reencrypt"
                },
                "to": {
                    "kind": "Service",
                    "name": "pod-svc",
                    "weight": 100
                },
                "wildcardPolicy": "None"
            },
            "status": {
                "ingress": [
                    {
                        "conditions": [
                            {
                                "lastTransitionTime": "2021-11-14T08:26:12Z",
                                "status": "True",
                                "type": "Admitted"
                            }
                        ],
                        "host": "test-route-e2e-test-authentication-operator-kmtcm.apps.ci-op-80b910fj-1708d.origin-ci-int-aws.dev.rhcloud.com",
                        "routerCanonicalHostname": "router-default.apps.ci-op-80b910fj-1708d.origin-ci-int-aws.dev.rhcloud.com",
                        "routerName": "default",
                        "wildcardPolicy": "None"
                    }
                ]
            }
        },
```

applications/openshift/networking/routes_rate_limit/tests/ocp4/e2e.yml
```diff
@@ -1,2 +1,2 @@
 ---
-default_result: PASS
+default_result: FAIL
```
JAORMX, line 2: Why is this set to FAIL now?

is the cluster not passing by default? we could create a manual remediation that makes this pass. Here's an example: https://github.com/ComplianceAsCode/content/blob/master/applications/openshift/networking/configure_network_policies_namespaces/tests/ocp4/e2e-remediation.sh

In that case, there are no networkpolicies in the cluster, so we create a new non-control plane namespace and create a network policy for it to make it pass.

JAORMX (changes requested):
Let's get the e2e test addressed

applications/openshift/networking/routes_rate_limit/tests/ocp4/e2e.yml
Vincent056: Sorry I forgot to update this, it is pass by default

Vincent056: There is `test-route`, from "namespace": `"e2e-test-authentication-operator-kmtcm"`, cause this rule to fail

Vincent056:
provisioning /retest

JAORMX (approved):

Vincent056:
We are accepting this PR to fail the e2e test in https://github.com/ComplianceAsCode/content/pull/7862, because only `SC-5(1)` was added for rule `routes_rate_limit`, and it is not part of FedRAMP Moderate, it was not tested in the above PR. we later added SC-5 control, which is part of Moderate profile, so that it will get included in FedRAMP Moderate profile. The reason why the e2e-test is falling after the second run is that the IdP tests will create a route in a non-kube/openshift namespaces after the first scan.

PR #7873, opened 2021-11-12
Introduce handling of versioned Boolean algebra expressions

evgenyz:
#### Description:

The `ssg/boolean_expression.py` module is based on *boolean.py* (in the `ext` directory) and *pkg_resources* from setuptools (version specifiers PEP440, PEP508).

Examples and tests are in `tests/unit/ssg-module/test_boolean_expressions.py`.

#### Rationale:

We need it for handling CPE platforms and potentially other things.

### DIFF
--- a/tests/unit/ssg-module/test_boolean_expressions.py
+++ b/tests/unit/ssg-module/test_boolean_expressions.py
@@ -1,0 +1,105 @@
+import pytest
+
+from ssg import boolean_expression
+from xml.dom import expatbuilder
+
+
+class PlatformFunction(boolean_expression.Function):
+    def as_cpe_lang_xml(self):
+        return '<cpe-lang:logical-test negate="' + ('true' if self.is_not() else 'false') + \
+               '" operator="' + ('OR' if self.is_or() else 'AND') + '">' + \
+               ''.join([arg.as_cpe_lang_xml() for arg in self.args]) + "</cpe-lang:logical-test>"
+
+
+class PlatformSymbol(boolean_expression.Symbol):
+    def as_cpe_lang_xml(self):
+        return '<cpe-lang:fact-ref name="cpe:/a:' + self.name + ':' + ':'.join([v for (op, v) in self.specs]) + '"/>'
+
+
+class PlatformAlgebra(boolean_expression.Algebra):
+    def __init__(self):
+        super(PlatformAlgebra, self).__init__(symbol_cls=PlatformSymbol, function_cls=PlatformFunction)
+
+    @staticmethod
+    def as_cpe_lang_xml(expr):
+        s = '<cpe-lang:platform id="' + expr.as_id() + '">' + expr.as_cpe_lang_xml() + '</cpe-lang:platform>'
+        # A primitive but simple way to pretty-print an XML string
+        return expatbuilder.parseString(s, False).toprettyxml()
+
+
+@pytest.fixture
+def algebra():
+    return PlatformAlgebra()
+
+
+@pytest.fixture
+def expression1(algebra):
+    return algebra.parse(u'(oranges==2.0 | banana) and not ~apple + !pie', simplify=True)
+
+
+def test_dyn():
+    alg = boolean_expression.Algebra(symbol_cls=PlatformSymbol, function_cls=PlatformFunction)
+    exp = alg.parse('not banana and not apple or anything')
+    assert str(exp) == '(~banana&~apple)|anything'
+    assert str(exp.simplify()) == 'anything|(~apple&~banana)'
+
+
+def test_id(expression1):
+    assert str(expression1.as_id()) == 'apple_and_banana_or_oranges_eq_2.0_or_not_pie'
+
+
+def test_cnf(algebra, expression1):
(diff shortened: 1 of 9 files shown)


vojtapolasek:
Hello, that looks great. I see that tests on SLE and Debian are failing. It seems that ordering of XML attributes got swapped for some reason on these platforms. I can't reproduce it on Fedora.

evgenyz:
> Hello, that looks great. I see that tests on SLE and Debian are failing. It seems that ordering of XML attributes got swapped for some reason on these platforms. I can't reproduce it on Fedora.

Yeah, that's because of how `xml.dom`s `toprettyxml` works in different versions.

tests/unit/ssg-module/test_boolean_expressions.py
```diff
@@ -0,0 +1,90 @@
...
+    exp_s = exp.simplify()
+    assert str(exp_s) == 'anything|(~apple&~banana)'
+
+
+def test_evaluate_simple_boolean_ops(exp_simple):
+    assert exp_simple(**{'oranges': '2.0', 'apple': True, 'pie': True})
+    assert not exp_simple(**{'oranges': '2.0', 'apple': False, 'pie': True})
+
+
+def test_evaluate_simple_version_ops(exp_simple):
```
matejak, line 57: I would copy-paste the expression definition, as having it close to the actual test is convenient.

evgenyz:
>     1. The result of parsing the expression is either `Function` or `Symbol`. `Function` is basically a graph node, its `args` are other nodes or leaf nodes. Therefore, the representation of a boolean expression is available in form of a tree.

Yes.
 
>     2. There are other files included - is it about dealing with version numbers?

Yes, the `packaging` module is related to version specifiers.

>     3. There is expression simplification included "for free" - we have to make it clear whether we want it in the full scale though. I remember that we also liked the idea of having platform definitions close to the form in which the platform was defined originally, so the generated content is somewhat recognizable to content creators. That feature comes with its own problems too, so we just have to decide which path to take.

Yes, maybe. But I'm leaning towards converting everything into DNF form, it feels natural for expressions we would be writing. We can try and choose later, when we will have more material to consider.

matejak:
> ...
> > 2. There are other files included - is it about dealing with version numbers?
> > 
> 
> 
> Yes, the `packaging` module is related to version specifiers.

What about setuptools? Its [pkg_resources](https://setuptools.pypa.io/en/latest/pkg_resources.html#parsing-utilities) package contains functionality implementing PEP 440, and setuptools are widely available.

matejak:
> > ```
> > 1. The result of parsing the expression is either `Function` or `Symbol`. `Function` is basically a graph node, its `args` are other nodes or leaf nodes. Therefore, the representation of a boolean expression is available in form of a tree.
> > ```
> 
> 
> Yes.

It would be more user-friendly if a result of parsing is `OperatorNode` instead of `Function` with children under `child_nodes` instead of `args`,  and `LeafNode` instead of a `Symbol`. Once the expression is parsed, there is no node, just the tree, and especially the `args` attribute to access children nodes is confusing.

evgenyz:
> It would be more user-friendly if a result of parsing is `OperatorNode` instead of `Function` with children under `child_nodes` instead of `args`, and `LeafNode` instead of a `Symbol`. Once the expression is parsed, there is no node, just the tree, and especially the `args` attribute to access children nodes is confusing.

Well, first of all all of this is inherited from the `boolean.py`. We, of course, could create proxy classes and methods to do the renaming, but, honestly, I doubt that it worth the effort.

evgenyz:
> What about setuptools? Its [pkg_resources](https://setuptools.pypa.io/en/latest/pkg_resources.html#parsing-utilities) package contains functionality implementing PEP 440, and setuptools are widely available.

Actually, we can use `pkg_resources`. During the first round of matching I overlooked the fact that there is a requirement parser (PEP 508) ready to be used: https://setuptools.pypa.io/en/latest/pkg_resources.html#requirements-parsing. Thanks for the hint for redoing my once-over. I'll update the PR.

vojtapolasek:
> > > ```
> > > 1. The result of parsing the expression is either `Function` or `Symbol`. `Function` is basically a graph node, its `args` are other nodes or leaf nodes. Therefore, the representation of a boolean expression is available in form of a tree.
> > > ```
> > 
> > 
> > Yes.
> 
> It would be more user-friendly if a result of parsing is `OperatorNode` instead of `Function` with children under `child_nodes` instead of `args`, and `LeafNode` instead of a `Symbol`. Once the expression is parsed, there is no node, just the tree, and especially the `args` attribute to access children nodes is confusing.

I think it is not necessary. These "Symbol" and "Function" classes will be anyway subclassed by CPEALLogicalTest and CPEALFactRef, at least that is my solution so far. So the technical details will be hidden anyway. Do you feel strongly about having it more user friendly?

vojtapolasek (approved):
Thank you very much for this implementation. I like it. I managed to incorporate it into the build system in my private branch and I did not find any serious obstacles.
I think this can be merged as soon as all questions get answered.

yuumasato (approved):
LGTM.

I understand @matejak's concern but also agree that creating proxy classes may not be necessary.
@vojtapolasek has some implementation using this code and we will see there if it will bee too confusing with the classes as they are.

PR #7877, opened 2021-11-14
OCP4: Add rule for SC-5: resource_requests_quota

Vincent056:
This PR adds a rule: resource_requests_quota- Check if either a cluster resource quota is set or a resource quota is set per namespace


applications/openshift/general/resource_requests_quota_per_project/oval/shared.xml
```diff
@@ -0,0 +1,94 @@
...
+
+  <!-- Comparison -->
+  <!-- This verifies that the filtered namespaces are equal to the namespaces with resourcequota -->
+  <ind:variable_state id="state_elements_count_for_resource_requests_quota_per_project" version="1">
+    <ind:value var_ref="local_variable_counter_resource_requests_quotas_filtered_namespaces"/>
+  </ind:variable_state>
+  
+  <!-- OCP data root declaration -->
+  <external_variable comment="Root of OCP data dump" datatype="string" id="ocp_data_root" version="1" />
+</def-group>
```
Vincent056, line 94: Credit to @JAORMX, I found a nice way of doing checks from `applications/openshift/networking/configure_network_policies_namespaces`

Vincent056:
For `cluster resource quota`, I only checked if there is at least one `cluster resource quota` presented, do you guys think this is good enough? @JAORMX @jhrozek @mrogers950

JAORMX:
```
     helpers.go:808: E2E-FAILURE: The expected result for the resource_requests_quota rule didn't match. Expected 'PASS', Got 'FAIL' 
```

the remediation didn't do the trick

Vincent056:
> ```
>      helpers.go:808: E2E-FAILURE: The expected result for the resource_requests_quota rule didn't match. Expected 'PASS', Got 'FAIL' 
> ```
> 
> 
>      
>   
> 
> the remediation didn't do the trick

Sorry, forgot to set file permission lol

applications/openshift/general/resource_requests_quota/oval/shared.xml
```diff
@@ -0,0 +1,11 @@
+<def-group oval_version="5.11">
+  <definition class="compliance" id="resource_requests_quota" version="1">
+    {{{ oval_metadata("The sysctl parameter needs to be set before enabling kernel protection") }}}
+
+    <criteria operator="OR">
+      <extend_definition comment="cluster quotas enabled" definition_ref="resource_requests_quota_cluster" />
+      <extend_definition comment="resource quota per project" definition_ref="resource_requests_quota_per_project" />
+    </criteria>
```
Vincent056, line 8: There is something that seems really weird, all the unit cases passed for  `resource_requests_quota` where either `resource_requests_quota_cluster` passes or `resource_requests_quota_per_project` passes, however, that's not the case when the check was running in the cluster, the main rule fails when `resource_requests_quota_cluster` passes, did I write anything wrong maybe in here? @JAORMX

applications/openshift/general/resource_requests_quota/rule.yml
```diff
@@ -0,0 +1,47 @@
...
+  application, it is important to tune based on memory and CPU consumption, 
+  allocating enough resources for the application to function properly.
+
+severity: medium
+
+identifiers:
+  cce@ocp4: CCE-90588-5
+
+references:
+  nist: SC-5,SC-5(1),SC-5(2)
```
JAORMX, line 47: You need to include the warning that tells the compliance operator that it needs to download the resources. Else it won't do it and it'll fail.

JAORMX (changes requested):

applications/openshift/general/resource_requests_quota/rule.yml
Vincent056: But this one is extending the other two rules? Do I need to include the warning for both of two rules, `resource_requests_quota_cluster`, `resource_requests_quota_per_project`

JAORMX: that's right. The Compliance Operator is not that smart :) So it will only read the warning from the rule that's being used in the data stream. Because of this, the `resource_requests_quota` needs to contain all of the resource inclusions you use.

We could fix this in CO... but it would require further parsing the DS to detect if you're importing other rules in the DS from a single rule...

Vincent056: Ok, that's really helpful. thanks for letting me know that! I think adding a warning will fix the problem!

Vincent056:
@JAORMX stack warning together did not work, I changed all the rules to use jq filter instead

JAORMX (approved):
NICE!

PR #7878, opened 2021-11-15
Add CentOS Stream 9 derivative product from RHEL9

ggbecker:
#### Description:

- Add CentOS Stream 9 derivative product from RHEL9.

#### Rationale:

- Do not exclude RHEL9 profiles.
- Use `cs9` as product name since OSCAP-Anaconda-Addon expects in this format.

CentOS Linux 8:
```python
>>> from pyanaconda.core import constants
>>> constants.shortProductName
'cl'
```

CentOS Stream 9:
```python
>>> from pyanaconda.core import constants
>>> constants.shortProductName
'cs'
```

In OOA we use `shortProductName` to find the default datastream to be used

https://github.com/OpenSCAP/oscap-anaconda-addon/blob/8229ad19588665d858dbdc1796be3bcd1bc79079/org_fedora_oscap/common.py#L80

I believe that for CentOS it never worked as we name the CentOS datastream as `ssg-centos8-ds.xml` for example, and the one that OAA tries to find is `ssg-cl8-ds.xml` for CentOS Linux 8 and `ssg-cs9-ds.xml` for CentOS Stream 9.

### DIFF
--- a/build-scripts/enable_derivatives.py
+++ b/build-scripts/enable_derivatives.py
@@ -97,7 +97,8 @@ def main():
         raise RuntimeError("No Benchmark found!")
 
     for namespace, benchmark in benchmarks:
-        ssg.build_derivatives.profile_handling(benchmark, namespace)
+        if args[1] != "cs9":
+            ssg.build_derivatives.profile_handling(benchmark, namespace)
         if not ssg.build_derivatives.add_cpes(benchmark, namespace, mapping):
             raise RuntimeError(
                 "Could not add derivative OS CPEs to Benchmark '%s'."
--- a/build-scripts/enable_derivatives.py
+++ b/build-scripts/enable_derivatives.py
@@ -98,6 +98,8 @@ def main():
 
     for namespace, benchmark in benchmarks:
         if args[1] != "cs9":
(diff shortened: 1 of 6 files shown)


build-scripts/enable_derivatives.py
```diff
@@ -97,7 +97,8 @@ def main():
         raise RuntimeError("No Benchmark found!")
 
     for namespace, benchmark in benchmarks:
-        ssg.build_derivatives.profile_handling(benchmark, namespace)
+        if args[1] != "cs9":
+            ssg.build_derivatives.profile_handling(benchmark, namespace)
```
jan-cerny, line 101: It would probably need to add a comment explaining why is CentOS Stream 9 different than other derivatives.

ggbecker:
I believe it's ready to be reviewed.

jan-cerny (approved):

jan-cerny:
It builds and the rules are applicable on a c9s VM.

ggbecker:
Screenshot from 2022-02-16 12-10-53

Left is old version of content without this patch and right version with the patch.

Content is now by default present on Centos Stream 9 and list all RHEL9 profiles.

PR #7879, opened 2021-11-15
Add CentOS 7 build to CI

Mab879:
#### Description:

Adds a CentOS 7 build to CI.

#### Rationale:

Make sure that our Python scripts work on CentOS 7.

### DIFF
--- a/.github/workflows/gate.yaml
+++ b/.github/workflows/gate.yaml
@@ -16,7 +16,22 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
       - name: Build
-        run: ./build_product rhel7 --derivatives
+        run: |-
+          ./build_product \
+              chromium \
+              fedora \
+              firefox \
+              fuse6 \
+              rhcos4 \
+              rhel7 \
+              rhel8 \
+              rhel9 \
+              rhosp10 \
+              rhosp13 \
+              sle12 \
+              sle15 \
+              ubuntu2004
+              --derivatives
       - name: Test
         run: ctest -j2 --output-on-failure -E unique-stigids
         working-directory: ./build
--- a/.github/workflows/gate.yaml
+++ b/.github/workflows/gate.yaml
@@ -28,9 +28,6 @@ jobs:
               rhel9 \
               rhosp10 \
               rhosp13 \
-              sle12 \
-              sle15 \
-              ubuntu2004
               --derivatives
       - name: Test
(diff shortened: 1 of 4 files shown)


.github/workflows/gate.yaml
```diff
@@ -5,6 +5,22 @@ on:
...
+    runs-on: ubuntu-latest
+    container:
+      image: centos:7
+    steps:
+      - name: Install Deps
+        run: yum install -y cmake make openscap-utils PyYAML libxslt expat xml-common python-jinja2
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Build
+        run: ./build_product rhel7
```
ggbecker, line 19:
```suggestion
        run: ./build_product rhel7 --derivatives
```

since we are building within centos, let's build the centos 7 derivatives as well

ggbecker:
It seems there are some problems with type hints in python2:
```
 [ 98%] [rhel7-generate-ssg-delta] generating disa tailoring file
  File "/__w/content/content/utils/create_scap_delta_tailoring.py", line 26
    def get_profile(product: str, profile_name: str) -> ET.Element:
                           ^
SyntaxError: invalid syntax
make[2]: *** [rhel7/tailoring/rhel7_stig_delta_tailoring.xml] Error 
```
https://stackoverflow.com/questions/35230635/type-hinting-in-python-2

.github/workflows/gate.yaml
```diff
@@ -5,6 +5,22 @@ on:
...
+    runs-on: ubuntu-latest
+    container:
+      image: centos:7
+    steps:
+      - name: Install Deps
+        run: yum install -y cmake make openscap-utils PyYAML libxslt expat xml-common python-jinja2
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Build
+        run: ./build_product rhel7 --derivatives
```
yuumasato, line 19: We risk missing build issues in other products, we should consider building `rhel8`, `jre` and `firefox` too.

For example, current master has an issue with Rule.to_ocil() and rule `service_systemd-coredump_disabled` when building RHEL8 content on python 2.

Mab879: I just added all the products back.

Mab879:
The CI is failing to an issue with Docker. Going to wait a bit and retry the jobs.

yuumasato (approved):
Thank you for adding the test and solving the build issues!

PR #7897, opened 2021-11-18
OCP: Add SI-6(d) file_integrity_notification_enabled

Vincent056:
SI-6(d) requires the configuration and support of alerts by our operators, This PR adds rule `file_integrity_notification_enabled` to check if a Prometheus rule object is created by File Integrity Operator.


applications/openshift/general/file_integrity_notification_enabled/tests/ocp4/e2e.yml
```diff
@@ -0,0 +1,3 @@
+---
+default_result: FAIL
+result_after_remediation: PASS
```
JAORMX, line 3: nice to see this passes. Where's the remediation though? I'm guessing this relies on the manual remediation we already have that installs FIO, does it?

JAORMX (changes requested):

applications/openshift/general/file_integrity_notification_enabled/tests/ocp4/e2e.yml
Vincent056: Yeah it does, `applications/openshift/integrity/file_integrity_exists/tests/ocp4/e2e-remediation.sh`

controls/nist_ocp4.yml
```diff
@@ -15802,6 +15802,14 @@ controls:
 
     https://issues.redhat.com/browse/CMP-544
   rules: []
+- id: SI-6(d)
```
JAORMX, line 15805: This is adding a new control (`SI-6(d)`) and not adding it to any level. So this won't get included in any profile. Let's add it to SI-6, the (d) section is not really a control but merely a bullet point in the control's verbiage.

JAORMX (changes requested):

controls/nist_ocp4.yml
Vincent056: Just fixed it

JAORMX (approved):
LGTM. I'll merge it once CI passes

PR #7899, opened 2021-11-18
Rename/remove rule for package abrt-addon-python

yuumasato:
#### Description:

- Remove rule package_abrt-addon-python_removed from RHEL-8 and RHEL-9.
- Add rule package_python3-abrt-addon_removed to RHEL-8.
- 
#### Rationale:

- The package was renamed in RHEL-8 and RHEL-9 has no such package.


linux_os/guide/system/software/system-tools/package_python3-abrt-addon_removed/rule.yml
```diff
@@ -0,0 +1,29 @@
...
+    {{{ describe_package_remove(package="python3-abrt-addon") }}}
+
+rationale: |-
+    <tt>python3-abrt-addon</tt> contains python hook and python analyzer
+    plugin for handling uncaught exceptions in python programs.
+
+severity: low
+
+identifiers:
+    cce@rhel8: CCE-82923-4
```
yuumasato, line 17: I have reused the CCE from RHEL-8, should I actually get a new one?

jan-cerny: I think that yes, you should, it's a different package, therefore a different configuration setting.

yuumasato: New CCE added.

jan-cerny:
@yuumasato the profile stability test fails

jan-cerny (approved):

PR #7908, opened 2021-11-22
Ubuntu specific bash and oval for dconf_gnome_login_banner_text

dodys:
#### Description:

- Ubuntu uses /etc/gdm3/greeter.dconf-defaults to set login-screen configuration.
- Made some changes to the tests, maybe will need to tweak based on the results here.


linux_os/guide/system/accounts/accounts-banners/gui_login_banner/dconf_gnome_login_banner_text/tests/ubuntu_correct_value_stig.pass.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+# platform = multi_platform_ubuntu
+# profiles = xccdf_org.ssgproject.content_profile_stig
+
+apt update
+apt install gdm3 -y
```
Mab879, line 6: Where you having some issues `# packages` with these tests?

Mab879 (commented):
Just few questions around how packages are installed in the testing.

Thanks for the PR.

dodys:
I've rebased and simplified the test changes as some of them were not even needed.

linux_os/guide/system/accounts/accounts-banners/gui_login_banner/dconf_gnome_login_banner_text/tests/wrong_value_stig.fail.sh
```diff
@@ -1,21 +1,30 @@
 #!/bin/bash
 # profiles = xccdf_org.ssgproject.content_profile_stig
-# packages = dconf,gdm
```
Mab879, line 3: Just making sure that this removal was intentional.

Mab879 (commented):
Just one last package line that I have a question on.

linux_os/guide/system/accounts/accounts-banners/gui_login_banner/dconf_gnome_login_banner_text/tests/wrong_value_stig.fail.sh
dodys: yes, otherwise it will fail on Ubuntu, and also you have "install_dconf_and_gdm_if_needed" on line 12 to install them if needed

Mab879 (approved):

PR #7910, opened 2021-11-22
Update `chronyd_or_ntpd_set_maxpoll` to add maxpoll option to chrony pool directives

yuumasato:
#### Description:

- This updates the rule to recognize and remediate `pool` directives with the appropriate `maxpoll` value.
- Enable and apply the same changes to Ansible remediation.

#### Rationale:

- The same option that apply to 'server' directive also applies to 'pool' directive.
  - https://chrony.tuxfamily.org/doc/3.4/chrony.conf.html#pool
- RHEL-8 chrony configuration defaults to using `pool`:
  - e.g.: `pool 2.rhel.pool.ntp.org iburst`

### DIFF
--- a/linux_os/guide/services/ntp/chronyd_or_ntpd_set_maxpoll/bash/shared.sh
+++ b/linux_os/guide/services/ntp/chronyd_or_ntpd_set_maxpoll/bash/shared.sh
@@ -8,9 +8,9 @@ config_file="/etc/ntp.conf"
 
 
 # Set maxpoll values to var_time_service_set_maxpoll
-sed -i "s/^\(server.*maxpoll\) [0-9][0-9]*\(.*\)$/\1 $var_time_service_set_maxpoll \2/" "$config_file"
+sed -i "s/^\(\(server\|pool\).*maxpoll\) [0-9][0-9]*\(.*\)$/\1 $var_time_service_set_maxpoll \3/" "$config_file"
 
-# Add maxpoll to server entries without maxpoll
-grep "^server" "$config_file" | grep -v maxpoll | while read -r line ; do
+# Add maxpoll to server or pool entries without maxpoll
+grep "^\(server\|pool\)" "$config_file" | grep -v maxpoll | while read -r line ; do
         sed -i "s/$line/& maxpoll $var_time_service_set_maxpoll/" "$config_file"
 done
--- a/linux_os/guide/services/ntp/chronyd_or_ntpd_set_maxpoll/rule.yml
+++ b/linux_os/guide/services/ntp/chronyd_or_ntpd_set_maxpoll/rule.yml
@@ -11,6 +11,8 @@ description: |-
     <tt>maxpoll</tt> in <tt>/etc/ntp.conf</tt> or <tt>/etc/chrony.conf</tt>
     add the following:
     <pre>maxpoll {{{ xccdf_value("var_time_service_set_maxpoll") }}}</pre>
+    to <pre>server</pre> direstives. If using chrony any <pre>pool</pre> directives
+    should be configured too.
     {{% if product == "rhcos4" %}}
     <p>
     Note that if the remediation shipping with this content is being used, the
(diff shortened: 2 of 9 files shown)


yuumasato:
I didn't create my branch from an updated master branch and now the datastream diff comment shows some unrelated changes.
Maybe instead of building the latest master, should the GH Actions build from the latest common ancestor?
CC @ggbecker

linux_os/guide/services/ntp/chronyd_or_ntpd_set_maxpoll/bash/shared.sh
```diff
@@ -8,9 +8,9 @@ config_file="/etc/ntp.conf"
 
 
 # Set maxpoll values to var_time_service_set_maxpoll
-sed -i "s/^\(server.*maxpoll\) [0-9][0-9]*\(.*\)$/\1 $var_time_service_set_maxpoll \2/" "$config_file"
+sed -i "s/^\(\(server\|pool\).*maxpoll\) [0-9][0-9]*\(.*\)$/\1 $var_time_service_set_maxpoll \3/" "$config_file"
```
marcusburghardt, line 11: @yuumasato , is this `\3` back-reference necessary in any situation I can't see? Or could be removed?

yuumasato: The last back reference is necessary to keep any `server` or `pool` otions defined after `maxpool`.

It increased from `\2` to `\3` because I did not manage to get non-capturing groups to work with `sed`.
And it seems it is not supported: https://stackoverflow.com/a/36546377

marcusburghardt: Ok. Thanks

marcusburghardt (approved):
Thanks for this improvement @yuumasato .

linux_os/guide/services/ntp/chronyd_or_ntpd_set_maxpoll/rule.yml
```diff
@@ -11,6 +11,8 @@ description: |-
     <tt>maxpoll</tt> in <tt>/etc/ntp.conf</tt> or <tt>/etc/chrony.conf</tt>
     add the following:
     <pre>maxpoll {{{ xccdf_value("var_time_service_set_maxpoll") }}}</pre>
+    to <pre>server</pre> direstives. If using chrony any <pre>pool</pre> directives
```
ggbecker, line 14: typo: direstives

ggbecker:
The reason why we didn't accept the `pool` keyword was mainly because STIG only allows the `server` option because United States Naval Observatory (USNO) time servers only provide `server` artifacts. I asked DISA to consider the `pool` option and they said they can't adopt this change.

yuumasato:
@ggbecker The rule is about chrony or ntp being configured with`maxpoll` option, it is not about not using `pool` directive.

Not sure if that is a security or practical concern, but it looks like there should be a rule specific to prevent use of `pool` directive.

ggbecker:
> @ggbecker The rule is about chrony or ntp being configured with`maxpoll` option, it is not about not using `pool` directive.
> 
> Not sure if that is a security or practical concern, but it looks like there should be a rule specific to prevent use of `pool` directive.

I understand, unfortunately STIG doesn't have a requirement for the `server` configuration option and this rule was basically being abused for this implicit requirement. I realized that if we make this rule passing by default on a RHEL8 system, people won't actually configure the ntp servers to point to something that https://www.usno.navy.mil/USNO/time/ntp provides and will make the STIG compliance kind of misleading. RHEL7 already passes with default configuration because the default config uses `server` entries (so this has probably manifested in the wild already).

yuumasato:
> I realized that if we make this rule passing by default on a RHEL8 system, people won't actually configure the ntp servers to point to something that https://www.usno.navy.mil/USNO/time/ntp provides and will make the STIG compliance kind of misleading.

This seem like a third configuration item that there should be a separate rule for. And we have a rule that seems fit for that `chronyd_or_ntpd_specify_remote_server`.
Maybe with an appropriate value in https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/ntp/var_multiple_time_servers.var that rule can be used.

linux_os/guide/services/ntp/chronyd_or_ntpd_set_maxpoll/rule.yml
yuumasato: Addressed in https://github.com/ComplianceAsCode/content/pull/7945

PR #7918, opened 2021-11-24
grub2_kernel_trust_cpu_rng was checking for wrong option

vojtapolasek:
#### Description:

- modify OVAL to accept y as well as Y
- modify / add tests

#### Rationale:

- we are checking for kernel config option. Usually it is set to "y", but we were checking only for "Y". Now we check for both cases.


yuumasato (approved):
LGTM.

yuumasato:
> This datastream diff is auto generated by the check `Compare DS/Generate Diff`
> Click here to see the full diff

The diff shows nothing, maybe it should it show the change in the OVAL?
I tried re-running the job, but got same results.

PR #7919, opened 2021-11-24
Clarify behaviour of SSHD rules

yuumasato:
#### Description:

- Mention in the description:
  - What are the default SSH config values.
  - What config file is checked/remediated.
- Move tracking of products with distributed SSH config to `product.yml`

#### Rationale:

- The rule does not add the configuration item if the rule's target value is the default value.
- It is not clear where the configuration is remediated.
- Related to [RHBZ#2020670]( https://bugzilla.redhat.com/show_bug.cgi?id=2020670)


linux_os/guide/services/ssh/ssh_server/disable_host_auth/rule.yml
```diff
@@ -7,8 +7,10 @@ description: |-
     more secure than <tt>.rhosts</tt> authentication. However, it is
     not recommended that hosts unilaterally trust one another, even
     within an organization.
-    <br /><br />
-    To disable host-based authentication, add or correct the
+    <br />
+    The default SSH configuration disables host-based authentication.
+    <br />
+    To explicitly disable host-based authentication, add or correct the
```
yuumasato, line 13: @adelton Hello, does the change above make it more clear what is the default value and the behaviour of the rule?

Do you think some more explicit is needed? For example:
`
The default SSH configuration disables host-based authentication. The appropriate value will be used by SSH if no configuration file sets a value for <tt>HostbasedAuthentication</tt>.
`

yuumasato: @vojtapolasek your feedback is also welcome, :)

vojtapolasek:
For some reason direct replying is not accessible... So replying here.
Looks good, but I have a question. Can we guarantee that mentioned options are default for all products?

yuumasato:
> Can we guarantee that mentioned options are default for all products?

That is a good question, and I think that was or should have been answered when the template keys were being changed.
This PR updates the description to follow what the rule already does for a while, hehe.

shared/macros-oval.jinja
```diff
@@ -844,8 +844,8 @@
 {{%- set sshd_config_path = "/etc/ssh/sshd_config" %}}
 {{%- set sshd_config_dir = "/etc/ssh/sshd_config.d" -%}}
 {{%- set description = "Ensure '" ~ parameter ~ "' is configured with value '" ~ value ~ "' in " ~ sshd_config_path %}}
-{{%- if config_is_distributed %}}
-{{%- set description = description  ~ " and in " ~ sshd_config_dir -%}}
```
yuumasato, line 848: @matejak I'd appreciate your review on the description variable above, I believe it should say:
`Ensure parameter '<parameter>' is configured with value '<value>' in /etc/ssh/sshd_config or in /etc/ssh/sshd_config.d`

Also, any fedback on the `sshd_distributed_config` parametrization is welcome.

yuumasato:
I noticed that the Ansible remediations don't handle the config directories on Fedora and RHEL9.
I'll take a look at this, but will probably do a separate PR.

linux_os/guide/services/ssh/ssh_server/disable_host_auth/rule.yml
jan-cerny: I like your idea of the more explicit description. For the users it's hard to find whether they have the default settings or whether the default settings have been modified by something.

vojtapolasek (approved):
I like the approach. I don't see any problem with the PR.

linux_os/guide/services/ssh/ssh_server/disable_host_auth/rule.yml
yuumasato: I think I'm leaning towards making the description of default config more explicit.

JAORMX (approved):

yuumasato:
> Can we guarantee that mentioned options are default for all products?

Hello @dodys @richardmaciel-canonical  @freddieRv @teacup-on-rockingchair @anivan-suse 
Could you provide any insight if Ubuntu, SLE or OL make changes to the default SSHD config?

linux_os/guide/services/ssh/ssh_server/sshd_enable_pam/rule.yml
```diff
@@ -8,8 +8,12 @@ description: |-
     PasswordAuthentication in addition to PAM account and session module processing for all
     authentication types.
 
-    To enable PAM authentication, add or correct the following line in the
-    <tt>/etc/ssh/sshd_config</tt> file:
+    To enable PAM authentication, add or correct the following line in
+    {{% if product in ['fedora', 'rhel9'] %}}
```
jan-cerny, line 12: Should sshd_distributed_config be used?

linux_os/guide/services/ssh/ssh_server/disable_host_auth/rule.yml
```diff
@@ -7,9 +7,17 @@ description: |-
...
+    The default SSH configuration disables host-based authentication. The appropriate
+    configuration is used if no value is set for <tt>HostbasedAuthentication</tt>.
+    <br />
+    To explicitly disable host-based authentication, add or correct the
+    following line in
+    {{% if sshd_distributed_config == "true" %}}
+    <tt>/etc/ssh/sshd_config.d/00-complianceascode-hardening.conf</tt>:
+    {{% else %}}
+    <tt>/etc/ssh/sshd_config</tt>:
+    {{% endif %}}
```
jan-cerny, line 20: This `if` block repeats often in many rules. Would it be better to have a macro that would contain this block?

jan-cerny (commented):
It's amazing

shared/macros.jinja
```diff
@@ -106,7 +106,11 @@ oc get {{% if all_namespaces %}}--all-namespaces{{% elif namespace %}}-n {{{ nam
 #}}
 {{% macro ocil_sshd_option(default, option, value) -%}}
     To determine how the SSH daemon's <tt>{{{ option }}}</tt> option is set, run the following command:
+    {{% if product in ['fedora', 'rhel9'] %}}
```
iokomin, line 109: @yuumasato shouldn't condition instead of checking prodtypes be `{{%- if sshd_distributed_config == "true" %}}`

freddieRv:
@yuumasato There are no Oracle patches modifying default configuration options for ssh/sshd. So you can consider the Oracle package as version compatible with the RHEL one.

linux_os/guide/services/ssh/ssh_server/disable_host_auth/rule.yml
yuumasato: I'm okay converting this to a macro, my only concern would be readability of the source code...

yuumasato: Macro added in e7efd94f67021da6b54749a1eea4f83aeac977ce

linux_os/guide/services/ssh/ssh_server/sshd_enable_pam/rule.yml
yuumasato: Fixed in e04685101eedd853e97bf4690afc862986495384

shared/macros.jinja
yuumasato: Fixed in c8e17a56ac66df0e143764037f78a4296c278345

jan-cerny (approved):

linux_os/guide/services/ssh/ssh_server/sshd_enable_strictmodes/rule.yml
```diff
@@ -42,13 +47,7 @@ references:
...
     vmmsrg: SRG-OS-000480-VMM-002000
 
-ocil_clause: 'it is commented out or is not enabled'
-
-ocil: |-
-    To check if StrictModes is enabled or set correctly, run the
-    following command:
-    <pre>$ sudo grep StrictModes /etc/ssh/sshd_config</pre>
-    If configured properly, output should be <pre>yes</pre>
+{{{ complete_ocil_entry_sshd_option(default="yes", option="StricModes", value="yes") }}}
```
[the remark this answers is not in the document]
yuumasato: Addressed in https://github.com/ComplianceAsCode/content/pull/7945

PR #7927, opened 2021-11-29
Use common ancestor commit in Compare DS job

ggbecker:
#### Description:

- Use the common ancestor commit of a branch as the basis for compare ds datastream. If the branch is not up to date with latest main branch it can happens that newer changes introduced in master will appear in the diff because it's comparing with outdated branch.

#### Rationale:

- Mitigates: https://github.com/ComplianceAsCode/content/pull/7910#issuecomment-975681893


yuumasato (approved):
LGTM

ggbecker:
Merging as we can only verify that the changes are correct when the code is already in main branch. I'll be monitoring compare-ds jobs to check if they behave correctly.

PR #7928, opened 2021-11-29
Enable customization of check target resource path at runtime

tmishina:
### Description

This PR enables the OCP4 checks for the control-plane components of managed OpenShift services such as HyperShift.

In a managed OpenShift environment, some check target API resources are moved to locations which is not expected in the rules of ComplianceAsCode/content.  This PR enables users to specify the check target resource location in XCCDF variables at runtime.

The PR consists of the following changes. 

1. a jinja macro to embed a variable which points to a resource
2. a jinja macro to specify a variable which points to the filepath for the resource
3. actural rules for HyperShift clusters

For item 1, as of now, we use a new macro `openshift_resource_path_xccdf` to embed a reference to a variable.
(This part will be replaced a new code because of the discussion in [the other PR](https://github.com/openshift/compliance-operator/pull/750))

For the item 2, my PR introduces a new element `filepath_xccdf_variable` in `.templates.yamlfile_value` of rule.yml.  However, is not smart - `filepath` is still required to avoid error.

I'd like to start discussion from this topics ([comment on the CO repo](https://github.com/openshift/compliance-operator/pull/750#issuecomment-979272103)) about `filepath_xccdf_variable` - how to avoid using two elements in a single yamlfile_value template?

### Rationale
- This PR expands the capability of ComplianceAsCode/content and Compliance Operator to various kinds of installations including managed OpenShift Services

### Related Topics
- [Enable runtime customization of API resource collection paths for HyperShift support by tmishina · Pull Request \#750 · openshift/compliance\-operator](https://github.com/openshift/compliance-operator/pull/750)


shared/macros.jinja
```diff
@@ -20,6 +20,9 @@ This rule's check operates on the cluster configuration dump.
 Therefore, you need to use a tool that can query the OCP API, retrieve the {{% if endpoint is string %}}<code class="ocp-api-endpoint">{{{ endpoint }}}</code> API endpoint to the local <code class="ocp-dump-location">{{{ xccdf_value("ocp_data_root") }}}/{{{ endpoint.lstrip("/") }}}</code> file.{{% else %}}{{% for item in endpoint %}}<code class="ocp-api-endpoint">{{{ item }}}</code> API endpoint to the local <code class="ocp-dump-location">{{{ xccdf_value("ocp_data_root") }}}/{{{ item.lstrip("/") }}}</code> file{{% endfor %}}.{{% endif %}}
 {{%- endmacro %}}
 
+{{% macro openshift_resource_path_xccdf(varname) -%}}
+Customized to: <code class="ocp-api-endpoint-var">{{{ varname }}}</code>
```
JAORMX, line 24: It would be good to add more verbiage here; this appears on the security guide and the report and it would be good to still give the appropriate context to people.

tmishina: I changed code to use the text substitution functionality discussed in https://github.com/openshift/compliance-operator/pull/750 and it avoids me from using a new macro `openshift_resource_path_xccdf`.  Instead, the code uses an existing macro `openshift_cluster_setting`.

Vincent056: Hi @tmishina, just looked at the PR, I am wondering is a reason why are we using `openshift_resource_path_xccdf`, instead of openshift_cluster_setting("/apis/apps/v1/namespaces/{{<somvar>}}"

tmishina: @Vincent056 sorry, my operation mistake.  I've just force-pushed the latest code which does not use `openshift_resource_path_xccdf`.

Vincent056: No worries, just let me know when you have the updated code.

tmishina: @Vincent056 Now the branch contains the updated code.  `macros.jinja` does not include `openshift_resource_path_xccdf`, and no rule uses it. Sorry for confusion.

applications/openshift/api-server/api_server_admission_control_plugin_AlwaysAdmit/rule.yml
```diff
@@ -29,20 +29,20 @@ ocil_clause: '<tt>AlwaysAdmit</tt> admission control plugin is not set'
...
     To verify that the <tt>AlwaysAdmit</tt> admission control plugin is not set, run the following command:
-    <pre>$ oc -n openshift-kube-apiserver get configmap config -o json | jq -r '.data."config.yaml"' | jq '.apiServerArguments."enable-admission-plugins"'</pre>
+    <pre>$ oc -n openshift-kube-apiserver get configmap config -o json | jq -r '.data."config.json"' | jq '.apiServerArguments."enable-admission-plugins"'</pre>
     The output should not contain <tt>AlwaysAdmit</tt>
 
 warnings:
 - general: |-
-    {{{ openshift_cluster_setting("/api/v1/namespaces/openshift-kube-apiserver/configmaps/config") | indent(4) }}}
-
+    {{{ openshift_cluster_setting("{{.var_openshift_kube_apiserver_config}}") | indent(4) }}}
```
Vincent056, line 37: Wondering do we want to have a global variable like {{{ openshift_cluster_setting("/api/v1/namespaces/{{<hyper-shift-ns>}}/openshift-kube-apiserver/configmaps/config" | indent(4) }}}, instead one variable for the whole api-path?

tmishina: Thank you, it is very important question which relates my initial comment (`how to avoid using two elements in a single yamlfile_value template?`).

There are two API resource path locations in a rule file: one is `.warnings[0]`, another is `.template.vars.filepath_xccdf_variable` (`.filepath_xccdf_variable` is a new element introduced by me).

`.warnings[0]` is used by the API resource collector of Compliance Operator. `parse_arf_result.go` substitutes a placeholder with actual values (e.g., `/api/{{.namespace}}.config{{.configname}}` to `/api/myns/config/myconfig`) before the resource collection, and therefore the API resource collector can read data from the correct resource path (`/api/myns/config/myconfig`).

In contrast, `.template.vars.filepath_xccdf_variable` is used by `oscap` command to read the API resource file. `oscap` command directly reads a data stream XML file, and therefore no substitution can be done before `oscap` opens the API resource file.  It means that we cannot use a description like (`filepath: /api/{{.namespace}}.config{{.configname}}`).

That's why I introduced a new element `.filepath_xccdf_variable` which contains full path for a resource instead of the combination of constant text and variables. This also force me use full path in `.warnings[0]`. 
(If we use notation like `/api/{{.namespace}}.config{{.configname}}` for `.warnings[0]`, then we need to maintain three variables. It just introduces complexity.)

I know that it is not smart, but I can't find any good workaround to avoid introducing the new element.
Do you have any idea to integrate `.template.vars.filepath` and `.template.vars.filepath_xccdf_variable`?

Vincent056: Hi @tmishina, I think that's an interesting point! I think there is also another issue, we have some rules that use jq filters, and we use `openshift_filtered_cluster_setting()` for warnings.

tmishina: @Vincent056 
One idea for solving the issue. I'd appreciate it if I can get your comments or other ideas!

---

What we'd like to do is converting an expression including variables like this:

```yaml
filepath: /api/v1/namespaces/{kas_config_namespace}/configmaps/{kas_config_name}
```

into this at the time of datastream build to avoid error at build and test:

```yaml
filepath: |-
	<local_variable>
	  <concat>
	    <variable_component var_ref="ocp_data_root"/>
	    <literal_component>/api/v1/namespaces/</literal_component>
	    <variable_component var_ref="kas_config_namespace"/>
	    <literal_component>/configmaps/</literal_component>
	    <variable_component var_ref="kas_config_name">
	  </concat>
	</local_variable>
```

Let's assume that we generate this value at [preprocess()](https://github.com/ComplianceAsCode/content/blob/master/shared/templates/yamlfile_value/template.py#L4-L31) and store it in `data['filepath_xml']`. In addition, modify [oval.template](https://github.com/ComplianceAsCode/content/blob/master/shared/templates/yamlfile_value/oval.template) as follows.

- use the new data FILEPATH_XML at the [conversion from FILEPATH into XML expression](https://github.com/ComplianceAsCode/content/blob/master/shared/templates/yamlfile_value/oval.template#L21-L30) 
- remain using FILEPATH when the template uses it as an identifier (5 occurrences: [here](https://github.com/ComplianceAsCode/content/blob/master/shared/templates/yamlfile_value/oval.template#L4), [here](https://github.com/ComplianceAsCode/content/blob/master/shared/templates/yamlfile_value/oval.template#L11),  [here](https://github.com/ComplianceAsCode/content/blob/master/shared/templates/yamlfile_value/oval.template#L13),  [here](https://github.com/ComplianceAsCode/content/blob/master/shared/templates/yamlfile_value/oval.template#L34), and [here](https://github.com/ComplianceAsCode/content/blob/master/shared/templates/yamlfile_value/oval.template#L75) ) .

Vincent056: I am wondering if we could feed the datastream to go template before we feed it to oscap as datastream, needs some research on it. It will be great if that could work!

tmishina: @Vincent056 I have pushed new commits which rules assume the implementation of the `feed the datastream to go-template before we feed it to oscap as datastream` feature and [the other PR on Compliance Operator](https://github.com/openshift/compliance-operator/pull/750). Please tell me if you have some progress on it, thanks!
Also, now branch has merged the latest contents of `master` branch as of today (2022-01-19).

tmishina:
Let me close this PR because we have moved to #8231

Issue #4354, opened 2019-05-17
Variables from OCIL defined in rule.yml are not being expanded


ggbecker:
#### Description of problem:
Variables used in OCIL section are not being expanded

linux_os/guide/system/accounts/accounts-physical/screen_locking/smart_card_login/configure_opensc_card_drivers/rule.yml:
```
ocil: |-
    To verify that <tt><sub idref="var_smartcard_drivers" /></tt> is configured
    as the smart card driver, run the following command changing <i>ARCH</i> for
    the architecture of your operating system:
    <pre>$ grep card_drivers /etc/opensc-<i>ARCH</i></pre>
    The output should return something similar to:
    <pre>card_drivers = <sub idref="var_smartcard_drivers" />;</pre>
```

#### SCAP Security Guide Version:
0.1.44

#### Steps to Reproduce:

1. cat build/ssg-rhel7-ds.xml | grep -A 2 -B 2 "card_drivers = ;"
2.
3.
4.

#### Actual Results:

$ grep card_drivers /etc/opensc-ARCH
The output should return something similar to:
card_drivers = ;
      Is it the case that the smart card driver is not configured correctly?
      </ns0:question_text>
...

#### Expected Results:
$ grep card_drivers /etc/opensc-ARCH
The output should return something similar to:
card_drivers = 300;
      Is it the case that the smart card driver is not configured correctly?
      </ns0:question_text>
...

jan-cerny:
We have discovered in https://github.com/ComplianceAsCode/content/pull/9662 that this also affects the HTML table `build/tables/table-rhel8-stig.html`.

The key fact here is that the generated OCIL files eg. `ssg-rhel8-ocil.xml` contain broken content. And this broken content propagates to the HTML table, but also to the SCAP source data stream that we ship.

Example of the OCIL element, notice the `"&lt;sub idref="var_accounts_passwords_pam_faillock_deny" /&gt;"` there:

```
22995       <ocil:question_text>Verify Red Hat Enterprise Linux 8 is configured to lock an account after 
22996 unsuccessful logon attempts with the command:
22997 
22998 
22999 $ grep 'deny =' /etc/security/faillock.conf
23000 deny = .
23001       Is it the case that the "deny" option is not set to "&lt;sub 
idref="var_accounts_passwords_pam_faillock_deny" /&gt;"
23002 or less (but not "0"), is missing or commented out?
23003       </ocil:question_text>
```

marcusburghardt:
[...]

PR #7936, opened 2021-11-30
fix enable_fips_mode remediations

vojtapolasek:
#### Description:

- modify Bash and Ansible remediations of enable_fips_mode
- apart from running fips-mode-setup --enable, it now applies also the cryptopolicy configured in the profile.

#### Rationale:

- the previous implementation only used the fips-mode-setup utility to enable the FIPS mode. However, this utility always configures the "fips" cryptopolicy. However, some profiles (such as OSPP), uses a policy based on "fips" - "fips:ospp". This policy is based on "fips", it is even stricter. But due to hardcoded configuration of "fips" policy done by fips-mode-setup, the remediation always resulted in error.
- This implementation should not cause problems even if the profile for some reason tries to use FIPS mode and at the same time it tries to use policy not based on fips. In that case, the check test_system_crypto_policy_value will always make rule fail.

### DIFF
--- a/linux_os/guide/system/software/integrity/fips/enable_fips_mode/ansible/shared.yml
+++ b/linux_os/guide/system/software/integrity/fips/enable_fips_mode/ansible/shared.yml
@@ -13,3 +13,13 @@
   command: /usr/bin/fips-mode-setup --enable
   when:
     - is_fips_enabled.stdout.find('FIPS mode is enabled.') == -1
+
+- name: "{{{ rule_title }}}"
+  lineinfile:
+    path: /etc/crypto-policies/config
+    regexp: '^(?!#)(\S+)$'
+    line: "{{ var_system_crypto_policy }}"
+    create: yes
+
+- name: Verify that Crypto Policy is Set (runtime)
+  command: /usr/bin/update-crypto-policies --set {{ var_system_crypto_policy }}
--- a/linux_os/guide/system/software/integrity/fips/enable_fips_mode/ansible/shared.yml
+++ b/linux_os/guide/system/software/integrity/fips/enable_fips_mode/ansible/shared.yml
@@ -3,6 +3,7 @@
 # strategy = restrict
 # complexity = medium
 # disruption = medium
+{{{ ansible_instantiate_variables("var_system_crypto_policy") }}}
 
 - name: Check to see the current status of FIPS mode
   command: /usr/bin/fips-mode-setup --check
--- a/linux_os/guide/system/software/integrity/fips/enable_fips_mode/rule.yml
+++ b/linux_os/guide/system/software/integrity/fips/enable_fips_mode/rule.yml
@@ -51,7 +51,7 @@ ocil: |-
     <pre>fips-mode-setup --check</pre>
     The output should contain the following:
     <pre>FIPS mode is enabled.</pre>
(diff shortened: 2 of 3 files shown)


vojtapolasek:
I am aware that I used copy paste. We can consider macro.

linux_os/guide/system/software/integrity/fips/enable_fips_mode/ansible/shared.yml
```diff
@@ -13,3 +13,13 @@
   command: /usr/bin/fips-mode-setup --enable
   when:
     - is_fips_enabled.stdout.find('FIPS mode is enabled.') == -1
+
+- name: "{{{ rule_title }}}"
+  lineinfile:
+    path: /etc/crypto-policies/config
+    regexp: '^(?!#)(\S+)$'
+    line: "{{ var_system_crypto_policy }}"
```
yuumasato, line 21: I believe you need to instantiate `var_system_crypto_policy`

yuumasato (commented):
I'd also update the rule description and mention that the crypto policy is expected to match `var_system_crypto_policy`

vojtapolasek:
I have updated the remediation and rule description. I was thinking about having there some Jinja conditional based on the value of var_system_crypto_policy... something like if the policy is set to FIPS, there will be no additional sentence. If there is a different policy, the sentence mentioning that the rule additionally configures a cryptography policy  would be added. But I am not sure if the build system actually supports different rule description for different profiles.

linux_os/guide/system/software/integrity/fips/enable_fips_mode/rule.yml
```diff
@@ -50,6 +51,10 @@ ocil: |-
     <pre>fips-mode-setup --check</pre>
     The output should contain the following:
     <pre>FIPS mode is enabled.</pre>
+    To verify that cryptography policy has been configured correctly, run the
```
yuumasato, line 54:
```suggestion
    To verify that the cryptographic policy has been configured correctly, run the
```

yuumasato (changes requested):

yuumasato:
> I was thinking about having there some Jinja conditional based on the value of var_system_crypto_policy... something like if the policy is set to FIPS, there will be no additional sentence. If there is a different policy, the sentence mentioning that the rule additionally configures a cryptography policy would be added. But I am not sure if the build system actually supports different rule description for different profiles.

I don't think this is possible with XCCDF.
The Jinja macros and conditionals are processed during content build time.
And the dynamicity of the rule descriptions at rule evaluation time are provided by `xccdf:sub` which only provides simple text substitution. AFAIK XCCDF doesn't support complex processing in the rule descriptions.

vojtapolasek:
I have accepted your suggestion. HOwever, I noticed that the variable does not show in ocil. However, it is displayed correctly in the rule description. Is that expected?

yuumasato:
> I have accepted your suggestion. HOwever, I noticed that the variable does not show in ocil. However, it is displayed correctly in the rule description. Is that expected?

Kind of :(
https://github.com/ComplianceAsCode/content/issues/4354

yuumasato (approved):

yuumasato:
Seems like `ci/prow/e2e-aws-rhcos4-moderate` is failing on infra.

PR #7455, opened 2021-08-24
Rule SLES-15-030710 - update

rumch-se:
#### Description:

- _Review required_

#### Rationale:

- _Code updates after proposed changes from upstream_

- Fixes # _Issue number here (e.g. #26) or remove this line if no issue exists._

PR #7937, opened 2021-11-30
compare-ds job: Do not post empty comments.

ggbecker:
#### Description:

- Do not post empty comments

#### Rationale:

- Sometimes it can happen that CTF will report a product to be built but the changes are actually only present in some certain product (jinja macros) and the diff generates an empty output. CTF can be enhanced to report exactly which product is changed. Usually CTF gives priority to RHEL distributions and the detection is based only on `prodtype` parameter.
- Example of empty comment: https://github.com/ComplianceAsCode/content/pull/7455#issuecomment-982437399


.github/workflows/compare-ds.yaml
```diff
@@ -57,8 +57,15 @@ jobs:
         run: utils/compare_ds.py ssg-${{steps.product.outputs.prop}}-ds.xml build/ssg-${{steps.product.outputs.prop}}-ds.xml | tee diff.log
         env:
           PYTHONPATH: ${{ github.workspace }}
-      - name: Get diff.log
+      - name: Test if there are datastream changes
         if: ${{ steps.ctf.outputs.CTF_OUTPUT_SIZE != '0' }}
+        run: echo "::set-output name=COMPARE_DS_OUTPUT_SIZE::$(stat --printf="%s" diff.log)"
+        id: compare_ds
+      - name: Print changes to datastream if any
```
yuumasato, line 64: Nitpick: The name gives me the impression that the changes will be added to the datastream.
Ssuggestion: `Print datastream changes if any`

yuumasato (approved):
LGTM.

.github/workflows/compare-ds.yaml
ggbecker: Ok, let me quickly rebase and add this modification

PR #7941, opened 2021-12-01
Remove OpenStack products and checks

JAORMX:
These have not been maintained for a long time and don't really work.

Let's remove them to avoid further confusion.

If/When OpenStack checks get created in CaC, they'll need to recreate
the `applications/openstack` directory and a product for the relevant
version (e.g. the current RHOSP version is `16.1` and work is on-going
for 17).


ggbecker:
I believe you have to remove occurrences such as: 

https://github.com/ComplianceAsCode/content/blob/a64f9de17fcb690b0ed21af26c12956915ffc0eb/CMakeLists.txt#L399

and maybe there are more in CMake files and python scripts.

JAORMX:
Pull-request updated, HEAD is now fa7e2578a8374302a9d4d0f1aa1485afe9f3e38d

matejak:
It looks like that this PR makes the CI fail. @jhrozek any thoughts on this?

jan-cerny (approved):

PR #7942, opened 2021-12-01
File attributes template: Create directory first if doesn't exist.

ggbecker:
#### Description:

- Templated tests for file attributes: Create directory of file if it doesn't exist.


JAORMX (approved):
Thanks!

PR #7943, opened 2021-12-01
Fix broken test scenario

jan-cerny:
During review of #7750 we have discovered that for this test scenario
the rule fails not because the `SELINUX=permissive` but because
no `SELINUX=` entry exists. The missing `SELINUX=` entry was caused
by a typo in the sed expression which effectively replaced
`SELINUX=enforcing` by just `permissive`. As a result, oscap didn't
find the OVAL object `oval:ssg-object_etc_selinux_config:obj:1`.

### DIFF
--- a/linux_os/guide/system/selinux/selinux_state/tests/selinux_enforcing.pass.sh
+++ b/linux_os/guide/system/selinux/selinux_state/tests/selinux_enforcing.pass.sh
@@ -2,6 +2,7 @@
 # profiles = xccdf_org.ssgproject.content_profile_C2S, xccdf_org.ssgproject.content_profile_ospp
 
 SELINUX_FILE='/etc/selinux/config'
+touch "$SELINUX_FILE"
 
 if grep -s '^[[:space:]]*SELINUX' $SELINUX_FILE; then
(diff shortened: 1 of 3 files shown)


ggbecker:
One of the tests reference a file that doesn't exist in the container environment:

```
##### selinux_state / selinux_missing.fail.sh #####
ssh -o Port=43163 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null root@localhost cd /root/ssgts/selinux_state; SHARED=/root/ssgts/shared bash -x selinux_missing.fail.sh
STDERR: Warning: Permanently added '[localhost]:43163' (ECDSA) to the list of known hosts.
+ SELINUX_FILE=/etc/selinux/config
+ sed -i '/^[[:space:]]*SELINUX/d' /etc/selinux/config
sed: can't read /etc/selinux/config: No such file or directory
```

It should be safe to create it before, or test if the file exists first and do just `touch` if it doesn't.

jan-cerny:
That's cute!

jan-cerny:
I have added touching the config file so now the test scenarios pass in the container in GitHub actions.

ggbecker (approved):

PR #7944, opened 2021-12-01
Fix jinja issue on fedora rawhide

ggbecker:
#### Description:

- Newer versions of Jinja in Fedora rawhide require that loader in yaml.load function, so every function that was missing the parameter now has the default Loader used.


JAORMX (approved):

JAORMX (dismissed):
Oh, seems there's still issues in rawhide

ggbecker:
> Oh, seems there's still issues in rawhide

I missed one occurrence of the current fix. It should be good now. Let's wait for the test results.

yuumasato (approved):

PR #7913, opened 2021-11-22
Ol7 stig v2r5

Xeicker:
#### Description:

- Remove grub password legacy rules from stig ol7
- Update grub2_admin_username and grub2_uefi_admin_username rules OVAL checks so they are capable of identifying next scenarios:
  - No super user defined : Fail (was already capable)
  - More than one superuser defined: fail (wasn't capable)
  - A superuser defined as admin, administrator, root: fail (was already capable)
  - A superuser defined with same username as aother OS user: fail (wasn't capable)
  - Non of the above: Pass
- Add sebool_ssh_sysadm_login rule to stig ol7 profile
- Create selinux_confine_to_least_privilege rule based on OL07-00-020021 disa stig ID
- Create selinux_context_elevation_for_sudo rule based on OL07-00-020023 disa stig ID

#### Rationale:

- This Updates ol7 stig profile to meet disa stig v2r5 requirements

PR #7948, opened 2021-12-02
Fix OVAL comments so they can be parsed by CTF tool

ggbecker:
#### Description:

- Fix OVAL comments so they can be parsed by CTF tool.

#### Rationale:

- CTF can parse the XML to find differences between content changes. Problem manifested here: https://github.com/ComplianceAsCode/content/pull/7913


mildas (approved):
`xml.etree.ElementTree.ParseError` in CTF has problem with quotes inside quotes. Workarounding these cases in the CTF analysis might be very complex and thus we should avoid such cases in CaC and do not use quotes in quotes

mildas:
Waiting for last check. When it's done, it can be merged

ggbecker:
merging as tests finished. The ones failing are expected because they are using broken content from master which should be fixed by this PR.

PR #7949, opened 2021-12-02
Fix CVE data URL

jan-cerny:
#### Description:

The CVE URL feed changed the location. It's still available also via
current URLs, but it's wise to use the new location:
* the uncompresesd XMLs are a redirect to compressed XMLs
* the files at www.redhat.com will redirect to access.redhat.com soon

#### Rationale:

Resolves: rhbz#2028428, rhbz#2028435

However, this changes means that we willl use bz2 files, which leads to a question on how we want to deal with warning "The OVAL feed of Red Hat Enterprise Linux 8 is not a XML file, which may not be understood by all scanners." That comes from https://github.com/ComplianceAsCode/content/blob/207eefd897a14e69e035846a99d22a792c284e07/linux_os/guide/system/software/updating/security_patches_up_to_date/rule.yml#L77 
I propose to keep the warning. WDYT?

### DIFF
--- a/products/rhel7/product.yml
+++ b/products/rhel7/product.yml
@@ -19,7 +19,7 @@ aux_pkg_version: "2fa658e0"
 
 release_key_fingerprint: "567E347AD0044ADE55BA8A5F199E2F91FD431D51"
 auxiliary_key_fingerprint: "43A6E49C4A38F4BE9ABF2A5345689C882FA658E0"
-oval_feed_url: "https://www.redhat.com/security/data/oval/com.redhat.rhsa-RHEL7.xml"
+oval_feed_url: "https://access.redhat.com/security/data/oval/com.redhat.rhsa-RHEL7.xml.bz2"
 
 grub2_uefi_boot_path: "/boot/efi/EFI/redhat"
 
--- a/products/rhel8/product.yml
+++ b/products/rhel8/product.yml
@@ -19,7 +19,7 @@ aux_pkg_version: "d4082792"
 
 release_key_fingerprint: "567E347AD0044ADE55BA8A5F199E2F91FD431D51"
 auxiliary_key_fingerprint: "6A6AA7C97C8890AEC6AEBFE2F76F66C3D4082792"
-oval_feed_url: "https://www.redhat.com/security/data/oval/com.redhat.rhsa-RHEL8.xml"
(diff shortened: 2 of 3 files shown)


ggbecker:
There's also one concern that SCAP standard does not allow usage of compressed files.

jan-cerny:
> There's also one concern that SCAP standard does not allow usage of compressed files.

But they're already compressed at this moment, the URL that we have in our code currently redirects to a compressed file. Try to click on it: https://www.redhat.com/security/data/oval/com.redhat.rhsa-RHEL8.xml and you'll get a bzip2 archive.

JAORMX (approved):

matejak:
We can't do anything else at the moment - our feeds simply are compressed, so we are just open about it instead of keeping it under the redirect.
The warning will be applicable as long as compression is not recognized by the SCAP standard, so I support keeping it there.
Thank you for the PR!

jan-cerny:
@matejak I don't like this because AFAIK it's not a warning, it's an error.

matejak:
> @matejak I don't like this because AFAIK it's not a warning, it's an error.

What do you mean by error? It shows up as a warning in the guide of the `security_patches_up_to_date` rule, and a link to that location is present in the datastream.

jan-cerny:
@matejak

> What do you mean by error? It shows up as a warning in the guide of the security_patches_up_to_date rule, and a link to that location is present in the datastream.

It's allegedly. That's what I heard from @ggbecker . Today, I tried to check if it's the case and I found that it fails SRC-169-1, but only when `scapval` is executed with `-online` option.

> An xccdf:Rule with @id "xccdf_NAMESPACE_rule_security_patches_up_to_date" and @multi-check=true SHALL reference an OVAL component that contains two or more oval definition of class 'patch'. This schematron assert also covers SRC-275-1 If your content contains external references, SCAPVal will attempt to resolve them in -online mode.

matejak:
> It's allegedly. That's what I heard from @ggbecker . Today, I tried to check if it's the case and I found that it fails SRC-169-1, but only when `scapval` is executed with `-online` option.

OK, that makes sense, we knew that `scapval` won't like it, and that's also what our warning in the rule description states is about. So all is as good as it can be under the current conditions.

PR #7956, opened 2021-12-03
Dockerfiles: Include EKS in Compliance Operator-related dockerfiles

JAORMX:
We never supported RHEL7 in the Compliance Operator, but EKS support is
on the works, so lets include that instead to make it easier for folks
to test it.


Vincent056 (approved):
/lgtm

PR #7957, opened 2021-12-03
utils/build_ds_container.sh: Add option to push to any image repository

JAORMX:
This adds the option to push to any given repository with the
aforementioned utility. This is useful for when developing content for
Kubernetes distros other than OpenShift.

A Dockerfile to take this into use was also created.

Co-Authored-By: Vincent Shen <wenshen@redhat.com>
Co-Authored-By: Lance Bragstad <lbragsta@redhat.com>

### DIFF
--- a/utils/build_ds_container.sh
+++ b/utils/build_ds_container.sh
@@ -14,6 +14,8 @@ print_usage() {
     echo "    $cmdname -c                      Build content in-cluster (NOTE: This ignores the products and debug flags)."
     echo "    $cmdname -d                      Build content using the --debug flag."
     echo "    $cmdname -P [product] (-P ...)   Specify applicable product(s) to build. This option can be specified multiple times. (Defaults to 'ocp4' 'rhcos4')"
+    echo "    $cmdname -r [repository]         Specify container image repository to push to"
+    echo "    $cmdname -t [tag]                The tag for the container image (defaults to latest)"
     exit 0
 }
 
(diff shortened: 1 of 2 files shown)


utils/build_ds_container.sh
```diff
@@ -72,7 +82,39 @@ if [ "$namespace" == "openshift-compliance" ]; then
...
+        # This enables us to create a configuration so we can build a container
+        # with the datastream
+        # If they already exist, this is not a problem
+        oc apply -n "$namespace" -f "$root_dir/ocp-resources/ds-from-local-build.yaml"
+
+        # Create output directory
+        from_dir=$(mktemp -d)
+
+        # Copy datastream files to output directory
+        cp "$root_dir/build/"*-ds.xml "$from_dir"
```
Vincent056, line 106: Do we still need those, since we are just copying from the build folder in Dockerfile?

Vincent056 (commented):
Just some nits.

utils/build_ds_container.sh
JAORMX: we do, this isn't for the registry push, but for the OpenShift push instead

Vincent056: Ahh, I think it is still used here https://github.com/ComplianceAsCode/content/blob/7b64c1e178e8d820894ccf7ba55fcb8f962716f5/utils/build_ds_container.sh#L161

Vincent056 (approved):
/lgtm

PR #7960, opened 2021-12-03
Fix issues with dividing a str by str in `utils/render-policy.py`

Mab879:
#### Description:

Use `os.path.join` for path building in `utils/render-policy.py`.

#### Rationale:

Without this fix the script will not work. Errors on trying to divide a `str ` by `str`.


ggbecker (approved):
This was probably when porting the script to python2. I think it was using `pathlib` previously

PR #7961, opened 2021-12-06
Filter out RHEL8 STIG rules on RHV hosts

yuumasato:
#### Description:

- Add CPE platform for `ovirt-host`, which is the base package for RHV virtualization hosts.
- Add the platform on RHEL8 STIG rules that should not be evaluated on RHV hosts.

#### Rationale:

- This enables specific rules from a profile to behave differently when the machine acts as a virtualization host.

### DIFF
--- a/linux_os/guide/services/ssh/ssh_server/sshd_disable_root_login/rule.yml
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_disable_root_login/rule.yml
@@ -56,13 +56,11 @@ references:
 
 {{{ complete_ocil_entry_sshd_option(default="no", option="PermitRootLogin", value="no") }}}
 
-{{% if product == "rhel8" %}}
 platforms:
 {{{ rule_notapplicable_when_ovirt_installed() | indent(4)}}}
 
 warnings:
 {{{ ovirt_rule_notapplicable_warning("RHV hosts require root access to be managed by RHV Manager") | indent(4) }}}
-{{% endif %}}
 
 template:
     name: sshd_lineinfile
--- a/linux_os/guide/services/ssh/ssh_server/sshd_disable_root_login/rule.yml
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_disable_root_login/rule.yml
@@ -61,9 +61,7 @@ platforms:
     - no_ovirt
 
 warnings:
-    - general: |-
-        This rule is disabled on Red Hat Virtualization Hosts and Managers, it will report not applicable.
-        RHV hosts require root access to be managed by RHV Manager.
+{{{ ovirt_rule_notapplicable_warning("RHV hosts require root access to be managed by RHV Manager") | indent(4) }}}
 {{% endif %}}
 
 template:
--- a/linux_os/guide/services/ssh/ssh_server/sshd_disable_root_login/rule.yml
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_disable_root_login/rule.yml
@@ -58,11 +58,11 @@ references:
 
 {{% if product == "rhel8" %}}
 platforms:
-    - not_ovirt-host
+    - no_ovirt
 
 warnings:
     - general: |-
-        This rule is disabled on Red Hat Virtualization Hosts, it will report not applicable.
+        This rule is disabled on Red Hat Virtualization Hosts and Managers, it will report not applicable.
         RHV hosts require root access to be managed by RHV Manager.
 {{% endif %}}
 
--- a/linux_os/guide/services/ssh/ssh_server/sshd_disable_root_login/rule.yml
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_disable_root_login/rule.yml
@@ -56,6 +56,11 @@ references:
 
 {{{ complete_ocil_entry_sshd_option(default="no", option="PermitRootLogin", value="no") }}}
 
+{{% if product == "rhel8" %}}
+platforms:
+    - not_ovirt-host
+{{% endif %}}
+
 template:
     name: sshd_lineinfile
     vars:
--- a/linux_os/guide/services/ssh/ssh_server/sshd_disable_root_login/rule.yml
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_disable_root_login/rule.yml
@@ -59,6 +59,11 @@ references:
 {{% if product == "rhel8" %}}
 platforms:
     - not_ovirt-host
+
+warnings:
+    - general: |-
+        This rule is disabled on Red Hat Virtualization Hosts, it will report not applicable.
+        RHV hosts require root access to be managed by RHV Manager.
 {{% endif %}}
 
 template:
--- a/linux_os/guide/services/ssh/ssh_server/sshd_disable_root_login/rule.yml
+++ b/linux_os/guide/services/ssh/ssh_server/sshd_disable_root_login/rule.yml
@@ -58,7 +58,7 @@ references:
 
 {{% if product == "rhel8" %}}
 platforms:
-    - no_ovirt
+{{{ rule_notapplicable_when_ovirt_installed() | indent(4)}}}
 
 warnings:
 {{{ ovirt_rule_notapplicable_warning("RHV hosts require root access to be managed by RHV Manager") | indent(4) }}}
--- a/linux_os/guide/system/software/sudo/sudo_remove_nopasswd/rule.yml
+++ b/linux_os/guide/system/software/sudo/sudo_remove_nopasswd/rule.yml
@@ -54,4 +54,9 @@ ocil: |-
 {{% if product == "rhel8" %}}
 platforms:
     - not_ovirt-host
+
+warnings:
+    - general: |-
+        This rule is disabled on Red Hat Virtualization Hosts, it will report not applicable.
+        RHV requires to perform operations as root without being asked for password.
 {{% endif %}}
--- a/linux_os/guide/system/software/sudo/sudo_remove_nopasswd/rule.yml
+++ b/linux_os/guide/system/software/sudo/sudo_remove_nopasswd/rule.yml
@@ -50,3 +50,8 @@ ocil: |-
     To determine if <tt>NOPASSWD</tt> has been configured for sudo, run the following command:
     <pre>$ sudo grep -ri nopasswd /etc/sudoers /etc/sudoers.d/</pre>
     The command should return no output.
+
+{{% if product == "rhel8" %}}
+platforms:
+    - not_ovirt-host
+{{% endif %}}
--- a/linux_os/guide/system/software/sudo/sudo_remove_nopasswd/rule.yml
+++ b/linux_os/guide/system/software/sudo/sudo_remove_nopasswd/rule.yml
@@ -56,7 +56,5 @@ platforms:
     - no_ovirt
 
 warnings:
-    - general: |-
-        This rule is disabled on Red Hat Virtualization Hosts and Managers, it will report not applicable.
-        RHV requires to perform operations as root without being asked for password.
+{{{ ovirt_rule_notapplicable_warning("RHV requires to perform operations as root without being asked for password") | indent(4) }}}
 {{% endif %}}
--- a/linux_os/guide/system/software/sudo/sudo_remove_nopasswd/rule.yml
+++ b/linux_os/guide/system/software/sudo/sudo_remove_nopasswd/rule.yml
@@ -53,7 +53,7 @@ ocil: |-
 
 {{% if product == "rhel8" %}}
 platforms:
-    - no_ovirt
+{{{ rule_notapplicable_when_ovirt_installed() | indent(4)}}}
 
 warnings:
 {{{ ovirt_rule_notapplicable_warning("RHV requires to perform operations as root without being asked for password") | indent(4) }}}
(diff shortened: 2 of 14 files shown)


yuumasato:
Here is an example report in a RHVH host:
[rhv-stig.html.txt](https://github.com/ComplianceAsCode/content/files/7660524/rhv-stig.html.txt)

(GH doesn't support HTML attachments, just download and rename it to `.html` extension)

linux_os/guide/services/ssh/ssh_server/sshd_disable_root_login/rule.yml
```diff
@@ -56,6 +56,16 @@ references:
...
 {{{ complete_ocil_entry_sshd_option(default="no", option="PermitRootLogin", value="no") }}}
 
+{{% if product == "rhel8" %}}
+platforms:
+    - not_ovirt-host
+
+warnings:
+    - general: |-
+        This rule is disabled on Red Hat Virtualization Hosts, it will report not applicable.
+        RHV hosts require root access to be managed by RHV Manager.
```
yuumasato, line 66: @almusil Let me know precise this is.

Any info about other rule is appreciated too.

shared/checks/oval/installed_env_has_ovirt-host_package.xml
```diff
@@ -0,0 +1,38 @@
...
+{{% if pkg_system == "rpm" %}}
+  <linux:rpminfo_test check="all" check_existence="at_least_one_exists"
+  id="test_env_has_ovirt-host_installed" version="1"
+  comment="system has package ovirt-host installed">
+    <linux:object object_ref="obj_env_has_ovirt-host_installed" />
+  </linux:rpminfo_test>
+  <linux:rpminfo_object id="obj_env_has_ovirt-host_installed" version="1">
+    <linux:name>ovirt-host</linux:name>
+  </linux:rpminfo_object>
+{{% elif pkg_system == "dpkg" %}}
```
almusil, line 27: We don't package for dpkg, this check can be removed if it's not required by the format.

shared/checks/oval/installed_env_has_ovirt-host_package.xml
```diff
@@ -0,0 +1,38 @@
...
+  </definition>
+
+{{% if pkg_system == "rpm" %}}
+  <linux:rpminfo_test check="all" check_existence="at_least_one_exists"
+  id="test_env_has_ovirt-host_installed" version="1"
+  comment="system has package ovirt-host installed">
+    <linux:object object_ref="obj_env_has_ovirt-host_installed" />
+  </linux:rpminfo_test>
+  <linux:rpminfo_object id="obj_env_has_ovirt-host_installed" version="1">
+    <linux:name>ovirt-host</linux:name>
```
almusil, line 25: Can this rule check multiple packages? Sorry I did not emphasize this enough, but the profile should apply to both hosts and manager, that would mean having installed `ovirt-host OR ovirt-engine`. Does it make sense?

shared/applicability/virtualization.yml
```diff
@@ -9,3 +9,13 @@ cpes:
       name: "cpe:/a:machine"
       title: "Bare-metal or Virtual Machine"
       check_id: installed_env_is_a_machine
+
+  - not_ovirt-host:
+      name: "cpe:/a:not_ovirt-host"
+      title: "OVirt is not installed"
```
almusil, line 15: Just small typo, it is oVirt (small o).

linux_os/guide/services/ssh/ssh_server/sshd_disable_root_login/rule.yml
almusil: This sounds like a good explanation.

linux_os/guide/system/software/sudo/sudo_remove_nopasswd/rule.yml
```diff
@@ -50,3 +50,8 @@ ocil: |-
     To determine if <tt>NOPASSWD</tt> has been configured for sudo, run the following command:
     <pre>$ sudo grep -ri nopasswd /etc/sudoers /etc/sudoers.d/</pre>
     The command should return no output.
+
+{{% if product == "rhel8" %}}
+platforms:
+    - not_ovirt-host
```
almusil, line 56: Something along: `This rule is disabled for Red Hat Virtualization, it will report not applicable. RHV requires to perform operations as a root without being asked for password.`

linux_os/guide/system/software/system-tools/package_gssproxy_removed/rule.yml
```diff
@@ -25,6 +25,11 @@ references:
 
 {{{ complete_ocil_entry_package(package="gssproxy") }}}
 
+{{% if product == "rhel8" %}}
+platforms:
+    - not_ovirt-host
```
almusil, line 30: Something along: `This rule is disabled for Red Hat Virtualization, it will report not applicable. RHV uses NFS storage, which has dependency on gssproxy.`

linux_os/guide/system/software/system-tools/package_tuned_removed/rule.yml
```diff
@@ -28,6 +28,11 @@ references:
 
 {{{ complete_ocil_entry_package(package="tuned") }}}
 
+{{% if product == "rhel8" %}}
+platforms:
+    - not_ovirt-host
```
almusil, line 33: Something along: `This rule is disabled for Red Hat Virtualization, it will report not applicable. RHV requires tuned package for tuning profiles that can enhance virtualization performance.`

shared/checks/oval/installed_env_has_ovirt-host_package.xml
yuumasato: > Can this rule check multiple packages? 

Yes, it is possible to check for multiple packages.

> Sorry I did not emphasize this enough, but the profile should apply to both hosts and manager, that would mean having installed `ovirt-host OR ovirt-engine`. Does it make sense?

It makes sense, but then the question I have is if the hosts and the manager will always have the same restrictions. Could a rule be ok for the manager but not for the host? Or vice-versa?

For example, does the system acting as the manager also need to allow for `root` login?

I guess one complicator is that the manager can be self-hosted right?
In this case a single system acts both as the manager and the host.
So it should be simpler to extend the platform to identify both host and manager.

almusil: > It makes sense, but then the question I have is if the hosts and the manager will always have the same restrictions. Could a rule be ok for the manager but not for the host? Or vice-versa?

It should not hurt if the rule applies for both.

> For example, does the system acting as the manager also need to allow for root login? 

Regular manager does not, but as you have mentioned the HE needs it. 

> I guess one complicator is that the manager can be self-hosted right?
In this case a single system acts both as the manager and the host.
So it should be simpler to extend the platform to identify both host and manager.

We have counted 6 or 7 rules that will be excluded (the 7th is not yet decided if we can fix it). And ~half of them is common for both I am not sure if it makes sense to split it. If it does I won't have anything against that.

shared/checks/oval/installed_env_has_ovirt-host_package.xml
yuumasato: @almusil So there are no Ubuntu or Debian packages for oVirt?

almusil: That's correct, we are officially building only for rpm targets.

shared/checks/oval/installed_env_has_ovirt-host_package.xml
yuumasato: One problem of having the same restrictions for both the manager and hosts is that they will not be as much aligned/hardened to the STIG standard as they could be. It could end up also misleading sys admins and auditors.

On a second thought, it is probably not that much complicated to distinguish rules that don't work for the manager, don't work for the hosts, or don't work for both.

On the self-hosted engine case (HE), are both `ovirt-host` and `ovirt-engine` installed on the machine?

almusil: No HE has only ovirt-host installed because it as a VM running on the first host.

shared/checks/oval/installed_env_has_ovirt-host_package.xml
yuumasato: Removed the dpkg tests.

linux_os/guide/system/software/sudo/sudo_remove_nopasswd/rule.yml
yuumasato: Thank you for all the suggestions!

shared/checks/oval/installed_env_has_ovirt-host_package.xml
yuumasato: So I guess the Host running the virtualized Manager needs the some kind of "freedom" as the Manager?

Anyway, we can go with simpler approach and improve if necessary.

almusil: Except some installation steps the host is not that special. But I agree, let's have common exclude for both.

almusil:
@yuumasato Can we please add:
`xccdf_org.ssgproject.content_rule_sysctl_net_ipv4_ip_forward` 
with explanation: `This rule is disabled on Red Hat Virtualization Hosts and Managers, it will report not applicable.
        RHV host requires IPv4 forwarding for the Hosted Engine bootstrap VM to reach network outside of the initial host.`
        
`xccdf_org.ssgproject.content_rule_xwindows_remove_packages`
with explanation: `This rule is disabled on Red Hat Virtualization Hosts and Managers, it will report not applicable.
        X11 graphic libraries are dependency of OpenStack Cinder storage.`
        
That should be everything for now, thank you

mwperina:
> `xccdf_org.ssgproject.content_rule_xwindows_remove_packages` with explanation: `This rule is disabled on Red Hat Virtualization Hosts and Managers, it will report not applicable. X11 graphic libraries are dependency of OpenStack Cinder storage.`

Just small rephrase, I'd change it to:

`This rule is disabled on Red Hat Virtualization Hosts and Manager, it will report not applicable. X11 graphic libraries are dependency of OpenStack Cinderlib storage provider.`

almusil:
@yuumasato  Also just for clarification, this skips will happen on all profiles right? Not only the DISA STIG. We are also interestedin PCI DSS, so if we need to also something specific to that profile or we would be good to go with those rules applying to it also.

yuumasato:
@almusil Hi, I have filtered out the two rules you mention.



> Also just for clarification, this skips will happen on all profiles right?

Yes, rule will be skipped no matter the profile. There is no need for something specific for PCI-DSS.

yuumasato:
I see that for PCI-DSS profile we only removed libreswan related rules:
https://github.com/ComplianceAsCode/content/pull/6867/commits/62a8cf02e78acd20aa17379c6f4a898508934275

matejak (changes requested):
LGTM, I have these questions/suggestions:

1. Would it make sense to reformulate the condition as `if "rhel" in product` to be more future-proof?
2. What about introducing helper macros `rhel_not_ovirt_platform()` and `rhel_not_ovirt_warning(define_warning_block=true)` that would add a platform, and that would define the whole warning block or provide just the "general" warning for cases when there already is one. Having a lot of code copy-pasted around is prone to copy-paste errors and typos.

yuumasato:
@matejak Thank you for your review and questions/suggestions.

> LGTM, I have these questions/suggestions:
> 
>     1. Would it make sense to reformulate the condition as `if "rhel" in product` to be more future-proof?

Probably not, each release of RHV is based on a specific version of RHEL.
@almusil thoughts on this?
 
>     2. What about introducing helper macros `rhel_not_ovirt_platform()` and `rhel_not_ovirt_warning(define_warning_block=true)` that would add a platform, and that would define the whole warning block or provide just the "general" warning for cases when there already is one. Having a lot of code copy-pasted around is prone to copy-paste errors and typos.

I have added two macros, one for the platform and another for the warning, together with a change in the build to make the rules mores readable.

almusil:
> @matejak Thank you for your review and questions/suggestions.
> 
> > LGTM, I have these questions/suggestions:
> > ```
> > 1. Would it make sense to reformulate the condition as `if "rhel" in product` to be more future-proof?
> > ```
> 
> Probably not, each release of RHV is based on a specific version of RHEL. @almusil thoughts on this?
> 

Yes as we need to be sure that this applies to specific version. 

> > ```
> > 2. What about introducing helper macros `rhel_not_ovirt_platform()` and `rhel_not_ovirt_warning(define_warning_block=true)` that would add a platform, and that would define the whole warning block or provide just the "general" warning for cases when there already is one. Having a lot of code copy-pasted around is prone to copy-paste errors and typos.
> > ```
> 
> I have added two macros, one for the platform and another for the warning, together with a change in the build to make the rules mores readable.

matejak:
> > Probably not, each release of RHV is based on a specific version of RHEL. @almusil thoughts on this?
> 
> Yes as we need to be sure that this applies to specific version.

We are good then, moreover as those conditions are now defined in macros, they are concentrated in one place, so future adjustments will be cheap and not error-prone.

matejak (approved):

yuumasato:
The ocp4 CI failures are related to https://github.com/ComplianceAsCode/content/issues/8012

yuumasato:
Rebased to include https://github.com/ComplianceAsCode/content/pull/8020

yuumasato:
'SSGTS / Run Tests (pull_request)' is affected by https://github.com/ComplianceAsCode/content/issues/8024

matejak:
> `SSGTS / Run Tests (pull_request)` is affected by #8024

Indeed, those errors are unrelated, so I merge the PR.

Issue #7890, opened 2021-11-16
`xccdf_org.ssgproject.content_rule_file_permissions_home_dirs` should it check for world write?


jcpunk:
#### Description of problem:

Should the check `xccdf_org.ssgproject.content_rule_file_permissions_home_dirs` ([Ensure that User Home Directories are not Group-Writable or World-Readable](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/accounts/accounts-session/file_permissions_home_dirs/rule.yml)) also check if the home area is world writable?  It seems an odd oversight.

A home area of `751` is suitable for my environment.

#### SCAP Security Guide Version:
scap-security-guide-0.1.58-1.fc35.noarch

#### Operating System Version:
Fedora 35

#### Steps to Reproduce:

1. `useradd -m exampleuser`
2. `chmod 753 /home/exampleuser`
3. Run rule `xccdf_org.ssgproject.content_rule_file_permissions_home_dirs`
4.

#### Actual Results:
`/home/exampleuser` meets the requirements of this rule

#### Expected Results:
An error that the directory permits group write.

Mab879:
@marcusburghardt do some insight here? Also, we two rules that seem very similar, `file_permissions_home_directories` and `file_permissions_home_dirs`.

lenox-joseph:
The rationale for the rule says that people shouldn't be able to write to other people's home directories.

marcusburghardt:
Hi @jcpunk , there is actually two very similar rules:
- [file_permissions_home_dirs
](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/accounts/accounts-session/file_permissions_home_dirs/rule.yml)
- [file_permissions_home_directories](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/accounts/accounts-session/file_permissions_home_directories/rule.yml)

The last is present in the current current versions of CIS and STIG benchmarks for RHEL8, for instance. I didn't check other distros, but it is possible that the first rule is outdated in favor of the second one. Could you inform which benchmark are you using, please?

Anyways, if any benchmark still recommends this rule as it is currently described, we have to fix it.
The current logic of the rule `file_permissions_home_dirs`, is checking for expected permissions instead of not expected permissions, like the `file_permissions_home_directories` rule correctly do, and this is causing this wrong result that you described.

jcpunk:
Hello,

[...]

PR #7963, opened 2021-12-06
Remediation and improvement for file_permissions_home_dirs rule

marcusburghardt:
Reviewing the issue https://github.com/ComplianceAsCode/content/issues/7890 it was noticed that this rule,
despite not being so used as the `file_permissions_home_directories`, was missing remediation and the OVAL
could be improved to be in alignment with other rules related to home directories owned by interactive users.


matejak (changes requested):
The rule got improved by pulling the OVAL code from the `file_permissions_home_directories` rule, and basically ditto for the Bash and Ansible remediations.
The OVAL code duplication could possibly be reduced by using the `extend_definition` mechanics, but it is probably not worth it, as Bash and Ansible will be duplicated as well.

What about adding a deprecation notice to the rule that it is substituted by the `file_permissions_home_directories` one to the rule yaml and perhaps to remediations? Perhaps a `..._deprecated_by(rule_id)` macro could come in handy :-)

jan-cerny:
@marcusburghardt ping :smile:

marcusburghardt:
After long time, I finally managed to resume the work in this PR. My last commit introduced a new macro.
I would like more opinions about pros and cons informing a release number on it. The intention it to make it explicit `when` the deprecation was set, but I am not sure if the release number is the best way to provide this information.

shared/macros/warning.jinja
```diff
@@ -12,6 +12,19 @@
...
 {{% endmacro %}}
 
+{{#
+    Includes a deprecation warning in rules which are no longer recommended in favor of a similar
+    and more updated rule.
+
+    :param rule: Rule which replaces the deprecated rule.
+    :param release: Release number since the rule is officially deprecated.
+#}}
+{{% macro warning_rule_deprecated_by(rule, release) %}}
```
matejak, line 22:
```suggestion
{{% macro warning_rule_deprecated_by(rule, release="") %}}
```

shared/macros/warning.jinja
```diff
@@ -12,6 +12,19 @@
...
+- general: |-
+    This rule is deprecated in favor of the <code>{{{ rule }}}</code> rule.
+    Please consider replacing this rule in your files as it is not expected to receive updates
+    as of version <code>{{{ release }}}</code>.
```
matejak, line 26:
```suggestion
    Please consider replacing this rule in your files as it is not expected to receive
    {{%- if release %}}
    updates as of version <code>{{{ release }}}</code>.
    {{%- else %}}
    further updates.
    {{% endif -%}}
```

linux_os/guide/system/accounts/accounts-session/file_permissions_home_dirs/bash/shared.sh
```diff
@@ -0,0 +1,12 @@
+# platform = multi_platform_all
+# reboot = false
+# strategy = restrict
+# complexity = low
+# disruption = low
+
+for home_dir in $(awk -F':' '{ if ($4 >= {{{ uid_min }}} && $4 != 65534) print $6 }' /etc/passwd); do
```
matejak, line 7: What is that `65534`? A max uid? Would it be possible to make an explanatory variable declared in the macro definitions file?

linux_os/guide/system/accounts/accounts-session/file_permissions_home_dirs/bash/shared.sh
```diff
@@ -0,0 +1,12 @@
...
+    # Only update the permissions when necessary. This will avoid changing the inode timestamp when
+    # the permission is already defined as expected, therefore not impacting in possible integrity
+    # check systems that also check inodes timestamps.
+    find $home_dir -perm /7027 -exec chmod u-s,g-w-s,o=- {} \;
```
matejak, line 11: Wouldn't it make sense to quote `$home_dir`?

linux_os/guide/system/accounts/accounts-session/file_permissions_home_dirs/bash/shared.sh
marcusburghardt: This uid is used in the special case of `nobody` user.

shared/macros/warning.jinja
marcusburghardt: Great. Thanks

matejak:
Thank you, I like those changes very much.

PR #7966, opened 2021-12-06
Fix configure_firewalld_ports bash to use jinja custom syntax

ggbecker:
#### Description:

- Fix configure_firewalld_ports bash to use jinja custom syntax

#### Rationale:

- We use custom syntax for jinja and we must use double curly braces


Mab879 (approved):
Thanks for the fix.

PR #7968, opened 2021-12-07
EKS: Introduce CIS controls file and take it into use

JAORMX:
This introduces `cis_eks.yml` in the controls folder. This will be used
to do the EKS CIS assessment.

The controls were already taken into use in the cis and cis-node
profiles.


controls/cis_eks.yml
```diff
@@ -0,0 +1,647 @@
...
+    Security is a shared responsibility between AWS and the Amazon EKS customer. The shared responsibility model describes this as security of the cloud and security in the cloud:
+
+    Security of the cloud – AWS is responsible for protecting the infrastructure that runs AWS services in the AWS Cloud. For Amazon EKS, AWS is responsible for the Kubernetes control plane, which includes the control plane nodes and etcd database. Third-party auditors regularly test and verify the effectiveness of our security as part of the AWS compliance programs. To learn about the compliance programs that apply to Amazon EKS, see AWS Services in Scope by Compliance Program.
+
+    Security in the cloud – Your responsibility includes the following areas.
+
+        - The security configuration of the data plane, including the configuration of the security groups that allow traffic to pass from the Amazon EKS control plane into the customer VPC
+        - The configuration of the worker nodes and the containers themselves
+
+          - The worker node guest operating system (including updates and security patches)
```
rhmdnd, line 25: This appears over indented compared to the rendered PDF.

controls/cis_eks.yml
```diff
@@ -0,0 +1,647 @@
...
+          - Setting up and managing network controls, such as firewall rules
+          - Managing platform-level identity and access management, either with or in addition to IAM
+
+        - The sensitivity of your data, your company’s requirements, and applicable laws and regulations
+
+    AWS is responsible for securing the control plane, though you might be able to configure certain options based on your requirements.
+  notes: |-
+    As noted, evaluating the security of the control plane is not the
+    responsibility of customers or deployers, but of Amazon instead. Thus, this
+    profile won't cover this.
```
rhmdnd, line 39: nit: s/won't cover this/will not implement controls from Section 1/ 

Or something of that nature?

controls/cis_eks.yml
```diff
@@ -0,0 +1,647 @@
...
+
+- id: '2'
+  title: '2 Control Plane Configuration'
+  description: |-
+    This section contains recommendations for Amazon EKS control plane logging
+    configuration. Customers are able to configure logging for control plane in
+    Amazon EKS.
+  levels:
+  - level_2
+  status: not applicable
```
rhmdnd, line 51: Is this applicable because there is an action that needs to be taken, even if it is manual at the moment?

jhrozek (approved):
/lgtm

controls/cis_eks.yml
```diff
@@ -0,0 +1,647 @@
...
+        Exporting logs and metrics to a dedicated, persistent datastore such as CloudTrail ensures
+        availability of audit data following a cluster security event, and provides a central location
+        for analysis of log and metric data collated from multiple sources.
+      # TODO: Evaluate if it's feasible to do an automated check for this
+      # control. This might require that the Compliance Operator acquire the
+      # ability to probe AWS endpoints (like the EKS one) to be able to do such
+      # checks.
+      notes: |-
+        It's not feasible to do such a check via automated means. e.g. via the
+        Compliance Operator using OpenSCAP. This has to be checked manually.
```
rhmdnd, line 81: I'm not sure if we can rely on the `aws` command line, but I was able to fetch this information using `aws eks describe-cluster $CLUSTER_NAME`
```
CLUSTERLOGGING  False
TYPES   api
TYPES   audit
TYPES   authenticator
TYPES   controllerManager
TYPES   scheduler
```

Manually enabling logging through the control plane gives me the following results afterwards:

```
CLUSTERLOGGING  True
TYPES   api
TYPES   audit
TYPES   authenticator
TYPES   controllerManager
TYPES   scheduler
```

controls/cis_eks.yml
```diff
@@ -0,0 +1,647 @@
...
+
+      If the --config argument is present, this gives the location of the Kubelet config file. This
+      config file could be in JSON or YAML format depending on your distribution.
+    status: pending
+    levels:
+    - level_2
+    controls:
+    - id: 3.2.1
+      title: >-
+        3.2.1 Ensure that the --anonymous-auth argument is set to false (Automated)
```
rhmdnd, line 202: Automated here doesn't mean we necessarily have an OpenSCAP control for it since the rules are empty, right?

controls/cis_eks.yml
```diff
@@ -0,0 +1,647 @@
...
+      - level_2
+      rules: []
+    - id: 3.2.11
+      title: >-
+        3.2.11 Ensure that the RotateKubeletServerCertificate argument is set to true (Manual)
+      status: pending
+      levels:
+      - level_1
+      rules: []
+- id: '4'
```
rhmdnd, line 277: Do we need to fill in the section description like we do for other sections?

controls/cis_eks.yml
Vincent056: I think we need control according to this? https://complianceascode.readthedocs.io/en/latest/manual/developer/03_creating_content.html?highlight=automated#reporting-status

Vincent056 (approved):
/lgtm

controls/cis_eks.yml
JAORMX: Right, but either way, we don't have the means to call `aws eks` unless we enhance the compliance operator. So, we need to evaluate the feasibility of this.

controls/cis_eks.yml
JAORMX: well, we could create a "MANUAL" rule and implement this. I can change it back to pending.

controls/cis_eks.yml
JAORMX: `status: automated` means we have an OpenSCAP control for this. the CIS benchmark has `(Automated) or `(Manual)` in the title for what they consider can be automatable. We could probably get rid of these.

controls/cis_eks.yml
JAORMX: We could. However, to not make this PR bigger than it already is, I suggest we do so as part of evaluating section 4.

JAORMX:
Pull-request updated, HEAD is now 6086e2b9f7959c3b6338a1a1c70fa6168384edd5

controls/cis_eks.yml
rhmdnd: Ack - sounds like we'll need something in CO to do that, then?

controls/cis_eks.yml
rhmdnd: Cool - yeah that sounds good. I was getting a little confused looking at the benchmark and seeing the inconsistency with what we have in this repository.

controls/cis_eks.yml
rhmdnd: Just wanted to make sure. Thanks for the update, let's do it in a follow up.

rhmdnd:
Looks good, I think we can fix the nits in a follow up. Thanks!

controls/cis_eks.yml
JAORMX: Yep, let's discuss this one outside of this PR

PR #7967, opened 2021-12-07
OCP4: Add unit test documentation

JAORMX:
This has been missing.

Signed-off-by: Juan Antonio Osorio Robles <jaosorior@redhat.com>

PR #7970, opened 2021-12-07
OCP4: Add documentation how to write unit test for rule uses jq filter

Vincent056:
In addition to https://github.com/ComplianceAsCode/content/pull/7967, added documentation on writing unit test for rule uses jq filter.


jhrozek (approved):
Thanks, this is helpful!
/lgtm

PR #7976, opened 2021-12-08
Add docs for create srg export

Mab879:
#### Description:

Add docs for SRG export and in the process work for the RHEL8 STIG.

#### Rationale:

Improve documentation for the SRG export.

### DIFF
--- a/linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_passwords_pam_faillock_deny_root/rule.yml
+++ b/linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_passwords_pam_faillock_deny_root/rule.yml
@@ -49,12 +49,29 @@ references:
     stigid@rhel7: RHEL-07-010330
     stigid@rhel8: RHEL-08-020022
 
-ocil_clause: 'that is not the case'
+ocil_clause: 'limiting the number of failed logon attempts for the root user is not configured'
 
 ocil: |-
     To ensure that even the <tt>root</tt> account is locked after a defined number of failed password
     attempts, run the following command:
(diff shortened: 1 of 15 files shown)


linux_os/guide/system/accounts/accounts-pam/locking_out_password_attempts/accounts_passwords_pam_faillock_deny_root/rule.yml
```diff
@@ -49,12 +49,29 @@ references:
...
-ocil_clause: 'that is not the case'
+ocil_clause: 'limiting the number of failed logon attempts for the root user is not configured'
 
 ocil: |-
     To ensure that even the <tt>root</tt> account is locked after a defined number of failed password
     attempts, run the following command:
     <pre>$ grep even_deny_root /etc/pam.d/system-auth</pre>
     The output should show <tt>even_deny_root</tt>.
 
+fix: |-
```
JAORMX, line 60: Is the intention, then, to have `fix` be manual steps to remediate something? Can't we somehow derive this from the "Ansible" or "bash" remediations?

Mab879: > Is the intention, then, to have fix be manual steps to remediate something? 

Currently, yes

> Can't we somehow derive this from the "Ansible" or "bash" remediations?

Possibly. I believe that we will still need to this key in some cases. But for cases where there are bash remediations we might be able to use those.

JAORMX: fair enough.

JAORMX (approved):

PR #7978, opened 2021-12-09
Fix `utils/rule_dir_json.py` when no product is detected when rendering OVAL

JAORMX:
`ssh/oval.py`'s `applicable_platform` fails in some cases where a
product is not detected since it needs to be rendered in jinja.

This adds a parameter that passes the product name and takes it into use
in `utils/rule_dir_json.py`.

### DIFF
--- a/utils/rule_dir_json.py
+++ b/utils/rule_dir_json.py
@@ -117,7 +117,7 @@ def handle_ovals(product_list, product_yamls, rule_obj):
         oval_product, _ = os.path.splitext(oval_name)
         oval_obj = {'name': oval_name, 'product': oval_product}
 
-        platforms = ssg.oval.applicable_platforms(oval_path)
+        platforms = ssg.oval.applicable_platforms(oval_path, product=oval_product)
         cs_platforms = ','.join(platforms)
 
         oval_obj['platforms'] = platforms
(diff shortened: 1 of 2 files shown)


utils/rule_dir_json.py
```diff
@@ -117,7 +117,7 @@ def handle_ovals(product_list, product_yamls, rule_obj):
         oval_product, _ = os.path.splitext(oval_name)
         oval_obj = {'name': oval_name, 'product': oval_product}
 
-        platforms = ssg.oval.applicable_platforms(oval_path)
+        platforms = ssg.oval.applicable_platforms(oval_path, product=oval_product)
```
ggbecker, line 120: oval_product contains the directory name such as `shared`, `firefox`, `jre`, etc... it doesn't really contain a product per se. If the product is really relevant for the script, we should make it as a parameter and let cmake pass the right product. If we don't need it then we can just put some default value so it doesn't traceback.

JAORMX: uhm... let me see what I can do

matejak: Probably the default `example` product introduced in #7979 will do the trick on a good-enough level - the utility in question is not designed to be product-specific, so for the time being let's hope that usage of the default product won't impact its result.

JAORMX: yeah, actually this PR is not needed any more.

PR #7979, opened 2021-12-09
Set default product name when loading product yaml

JAORMX:
If no product name is given, this uses the `example` product and adds it
to the implied properties.


matejak (approved):
LGTM, and works for me as well, thanks!

PR #7982, opened 2021-12-09
Fix OCP4 and EKS SCAP validation

ggbecker:
#### Description:

- Fix OCP4 and EKS SCAP validation
  - Requirement: SRC-330
    - No state should be referenced when check_existence has a value of 'none_exist'
    - inconsistent datatype between the variable and an associated var_ref


JAORMX (approved):
Thanks for fixing this!

PR #7983, opened 2021-12-09
honor CPE logical expressions when generating Bash and Ansible remediations

vojtapolasek:
What currently does not work is:
- when generating Ansible conditionals, platform_package_overrides is not honored. No idea why, I need help.
- one unit SSG test needs to be modified.
Other things should work very nicely, I did comparisons of datastreams generated from this branch and from master and we should not break anything.

#### Description:

- create new Jinja macros for creating of conditionals for installed packages
- put correct conditionals to definitions of CPE items (shared/applicability/...)
- get ridf of some magic functions in build_remediations.py (platform package overrides, selection of packaging system) this is now handled by JInja
- some tweaks to parsing of conditionals, especially ignore them if they are empty

#### Rationale:

### DIFF
--- a/shared/applicability/general.yml
+++ b/shared/applicability/general.yml
@@ -3,12 +3,13 @@ cpes:
       name: "cpe:/a:audit"
       title: "Package audit is installed"
       check_id: installed_env_has_audit_package
-#      bash_conditional: {{ pkg_bash_conditional(audit) }}
+      bash_conditional: {{{ pkg_bash_conditional("audit") }}}
 
   - chrony:
       name: "cpe:/a:chrony"
       title: "Package chrony is installed"
       check_id: installed_env_has_chrony_package
+      bash_conditional: {{{ pkg_bash_conditional("chrony") }}}
 
   - gdm:
       name: "cpe:/a:gdm"
--- a/shared/applicability/general.yml
+++ b/shared/applicability/general.yml
@@ -4,6 +4,7 @@ cpes:
       title: "Package audit is installed"
       check_id: installed_env_has_audit_package
       bash_conditional: {{{ bash_pkg_conditional("audit") }}}
+      ansible_conditional: {{{ ansible_pkg_conditional("audit") }}}
 
   - chrony:
       name: "cpe:/a:chrony"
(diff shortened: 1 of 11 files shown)


jan-cerny:
@vojtapolasek Regarding the Ansible:

> when generating Ansible conditionals, platform_package_overrides is not honored. No idea why, I need help.

It seems to work for me, for example in the `debian10` product which has multiple `platform_packaged_overrides`, the compiled platforms contain correct `ansible_conditional`. Except the `login_defs` platform, when I found that there is a typo in `shared/applicability/general.yml`: on line 35 there needs to be an underscore (`_`) in `login_defs` instead of a dash.

shared/applicability/general.yml
```diff
@@ -3,41 +3,57 @@ cpes:
...
       check_id: installed_env_has_libuser_package
+      bash_conditional: {{{ bash_pkg_conditional("libuser") }}}
+      ansible_conditional: {{{ ansible_pkg_conditional("libuser") }}}
 
   - login_defs:
       name: "cpe:/a:login_defs"
       title: "Package providing /etc/login.defs is installed"
       check_id: installed_env_has_login_defs
+      bash_conditional: {{{ bash_pkg_conditional("login_defs") }}}
+      ansible_conditional: {{{ ansible_pkg_conditional("login-defs") }}}
```
jan-cerny, line 35:
```suggestion
      ansible_conditional: {{{ ansible_pkg_conditional("login_defs") }}}
```

ssg/build_cpe.py
```diff
@@ -153,6 +153,12 @@ def __init__(self, cpeitem_data):
         self.name = cpeitem_data["name"]
         self.title = cpeitem_data["title"]
         self.check_id = cpeitem_data["check_id"]
+        self.bash_conditional = ""
+        if "bash_conditional" in cpeitem_data.keys():
+            self.bash_conditional = cpeitem_data["bash_conditional"]
+        self.ansible_conditional = ""
+        if "ansible_conditional" in cpeitem_data.keys():
+            self.ansible_conditional = cpeitem_data["ansible_conditional"]
```
jan-cerny, line 161: If  you look for a key in a dictionary, you don't have to use `x in d.keys()` it's just `x in d`. Also, you can use `.get()` to which gives a default value if a key isn't present in the dictionary.

```suggestion
        self.bash_conditional = cpeitem_data.get("bash_conditional", "")
        self.ansible_conditional = cpeitem_data.get("ansible_conditional", "")
```

jan-cerny (approved):

jan-cerny:
@vojtapolasek Can I merge now? Also, you mentioned in this PR description that one unit SSG test needs to be modified. Which test is that?

vojtapolasek:
You can merge if it helps you in writing test.
The failing test is python-unit-ssg-module .

vojtapolasek:
I have removed it.

yuumasato:
Removing milestone again, :)

PR #7987, opened 2021-12-09
Port build_ds_container.sh to python

rhmdnd:
The build_ds_container.sh script is useful for building content and
integrating it into your kubernetes clusters. Overtime, it's grown more
and more functionality, like building content in the cluster, locally,
or pushing built container images to repositories.

This commit attempts to reorganize the script and ports it to Python,
based on recommendations from the initial authors.

The primary changes include:

 - Using python kubernetes clients for interacting with clusters
 - Quieter logging by default, while maintaining verbosity using --debug
 - Auto-generated container image tags when publishing content to
   repositories (useful for development workflows)

This patch doesn't include support for OCP deployments as it was
originally developed to build content for EKS benchmarks. That
functionality will come in another revision of this patch and should
remain backwards compatible with `build_ds_container.sh`.

Signed-off-by: Lance Bragstad <lbragsta@redhat.com>

#### Description:

#### Rationale:


utils/build_ds_container.py
```diff
@@ -0,0 +1,257 @@
...
+    with client.ApiClient() as api_client:
+        body = client.V1Namespace(
+            api_version='v1', kind='Namespace', metadata={'name': namespace})
+
+        try:
+            api = client.CoreV1Api(api_client)
+            api.create_namespace(body)
+            log.info('Created namespace %s', namespace)
+        except exceptions.ApiException as err:
+            if err.status == 409:
```
Vincent056, line 103: Just curious, does 409 always only refer to when a namespace already exists.

utils/build_ds_container.py
```diff
@@ -0,0 +1,257 @@
...
+
+def create_build_config(build_config):
+    # FIXME(lbragstad) Fill this out using an openshift client, since we're
+    # dealing with an openshift-specific resource.
+    pass
+
+
+def build_content_locally(products, debug=False):
+    build_binary_path = os.path.join(REPO_PATH, 'build_product')
+    command = [build_binary_path] + products
```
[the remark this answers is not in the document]
Vincent056: Ok, I see, this is an array

Vincent056:
I think the codes looks good!

utils/build_ds_container.py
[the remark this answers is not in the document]
rhmdnd: Good question - it can be done either way in python. The python community issued guidance via a PEP that recommends using whitespace around operators.

https://www.python.org/dev/peps/pep-0008/#other-recommendations

utils/build_ds_container.py
mrogers950: Yeah, that's AlreadyExists. Might be good to use a constant or comment this number.

utils/build_ds_container.py
```diff
@@ -0,0 +1,257 @@
...
+
+def supports_custom_resource(resource_name):
+    with client.ApiClient() as api_client:
+        api = client.ApiextensionsV1Api(api_client)
+        try:
+            api.read_custom_resource_definition(resource_name)
+            return True
+        except exceptions.ApiException as err:
+            return False
+            if err.status == 404:
```
mrogers950, line 117: Hmm, can we reach this if the previous line is `return False` ?

utils/build_ds_container.py
```diff
@@ -0,0 +1,257 @@
...
+                return False
+            log.warning(
+                'Unable to determine if %s custom resources are supported in '
+                'this deployment', resource_name)
+            return False
+
+
+def create_image_stream():
+    # FIXME(lbragstad) Fill this out using an openshift client, since we're
+    # dealing with an openshift-specific resource.
```
mrogers950, line 127: Just wondering what we might gain by adding a second client lib for this. I think as long as the below works by specifying the api_version we'd be good.

utils/build_ds_container.py
```diff
@@ -0,0 +1,257 @@
...
+        'podman', 'push', 'localhost/localcontentbuild:latest',
+        repository_string]
+    capture_output = True
+    if debug:
+        capture_output = False
+    result = subprocess.run(command, capture_output=capture_output)
+    if result.returncode == 0:
+        log.info(f'Pushed image to {repository}:{tag}')
+    elif result.returncode == 125:
+        log.error(
```
mrogers950, line 187: Responded in the slack thread with some thoughts on the repo push.. It's unclear to me how the old script inherits repo and image tags to push this to the cluster.

utils/build_ds_container.py
rhmdnd: It should only be emitted when the namespace already exists. I added a comment to the code to make that more clear.

utils/build_ds_container.py
rhmdnd: Yep - good catch. I didn't finish smoothing out the wrinkles here before pushing this version.

I removed the if statement since I don't think it matters. Instead, I just logged something for the user and returned false.

utils/build_ds_container.py
rhmdnd: ack

utils/build_ds_container.py
rhmdnd: @JAORMX added that functionality recently  as command line arguments. In that implementation, the script relied on `podman login`.

I initially implemented this method with the ability to pass in a username/password and invoking the command with `--creds`, but afterwards I removed it to keep things simpler.

Instead, @JAORMX and I agreed that we can leave authentication to the user, prior to them running this script. In that case, I think the best we can do is just detect if authentication failed for some reason and give them guidance on how to fix it.

utils/build_ds_container.py
```diff
@@ -0,0 +1,256 @@
...
+result = subprocess.run(
+    ['git', 'rev-parse', '--show-toplevel'], capture_output=True, check=True)
+# Convert the file path from bytes to unicode since we might manipulate it
+# later. Also, strip off any newlines.
+REPO_PATH = result.stdout.decode().strip()
+
+
+def ensure_namespace_exists(namespace):
+    """Function that ensures there is a namespace for the content."""
+    with client.ApiClient() as api_client:
```
JAORMX, line 96: I have to admit, I'm not a big fan of using full blown kube APIs for simple scripts. using `subprocess.run` alongside `kubectl` or `oc` would be both more simple, and would make the declarations more obvious (since it would rely on YAML), thus easier to maintain. You could then do `oc apply -f ...` would would already take care of exception of objects already existing and such. So all of this would be reduced to just a couple of lines.

JAORMX (changes requested):

rhmdnd:
@Vincent056 I suppressed the output by default, but it should emit the same output as the bash script if you pass `--debug`.

Vincent056:
> @Vincent056 I suppressed the output by default, but it should emit the same output as the bash script if you pass `--debug`, but let me know if this isn't working how you expect it to.

Sorry didn't see that earlier, it looks good!

rhmdnd:
@Vincent056 I can take the opposite approach too and emit verbose logging by default, and supply an option to suppress logging using something like `--quiet` if you prefer.

jhrozek (approved):
The quiet mode was a bit of a surprise to me at first (I happened to have a broken profile in tree..) but I think it's fine and we can revert the debug option to quiet later on.
/lgtm

PR #7995, opened 2021-12-13
Fix compare-ds GH Action

ggbecker:
#### Description:

- The previous solution to find the forking point was not actually working and now it checks out specifically the commit from the Pull Request to compare. It was always comparing with latest master, and sometimes the PR has an outdated branch so we have to take this into consideration.
- It also adds logic to delete a previously put comment by `compare-ds` when a new update makes the pull request to have no changes regarding previous state.

I've tested the changes in my private fork as the changes are only reflected when they are merged because they use the special event trigger `pull_request_target`.

### DIFF
--- a/.github/workflows/compare-ds.yaml
+++ b/.github/workflows/compare-ds.yaml
@@ -84,7 +84,6 @@ jobs:
           body="${body//$'\r'/'%0D'}"
           echo ::set-output name=log::$body
       - name: Find Comment
-        if: ${{ steps.compare_ds.outputs.COMPARE_DS_OUTPUT_SIZE != '0'}}
         uses: peter-evans/find-comment@v1
         id: fc
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/compare-ds.yaml
```diff
@@ -108,3 +107,10 @@ jobs:
 
             </details>
           edit-mode: replace
+      - name: Delete comment in case no changes detected but there is already a comment from previous states
```
yuumasato, line 111: From the commit message I understand that this is to cover cases when some one force pushes and the new commit actually triggers no changes in compare-ds. 
But this is not much clear from the step `name`.

yuumasato:
>     * The previous solution to find the forking point was not actually working and now it checks out specifically the commit from the Pull Request to compare. It was always comparing with latest master, and sometimes the PR has an outdated branch so we have to take this into consideration.

Nitpick, the info in this item in the description could be put in the second commit's message.

yuumasato (approved):

PR #8001, opened 2021-12-15
CPEAL tests

jan-cerny:
#### Description:

* fix the broken test `tests/unit/ssg-module/test_build_remediations.py`
* add simple tests for `ProductCPEs` class
* add simple tests for `Platform` class

Also adds example data to `test/data` (product.yml, a few CPE definitions in `applicability`)

#### Rationale:
Increase test coverage of the CPE Applicability Language feature on unit level.

### DIFF
--- a/tests/unit/ssg-module/test_build_cpe.py
+++ b/tests/unit/ssg-module/test_build_cpe.py
@@ -94,13 +94,30 @@ def test_extract_referred_nodes():
     assert results[0].text == 'Source Code'
 
 
+#############################################
+# Unit tests for ProductCPEs.get_cpe() method
+#############################################
+#
+# Note that there are 2 types of CPE definitions that differ by the source they
+# come from:
+# * Product CPEs, loaded from product YAML
+# * Content CPEs, loaded from directory specified by the `cpes_root` key in
+#   product YML, usually from the `/applicability` directory
+#
+# This test case test that both types are used by the ProductCPEs class and
+# that both CPE types are handled equally.
 def test_product_cpes():
+
+    # CPEs are loaded from `DATADIR/product.yml` but also from
+    # `DATADIR/applicability` because `DATADIR/product.yml` references the
+    # `DATADIR/applicability` directory in the `cpes_root` key
     product_yaml_path = os.path.join(DATADIR, "product.yml")
     product_yaml = open_raw(product_yaml_path)
     product_yaml["product_dir"] = os.path.dirname(product_yaml_path)
     product_cpes = ssg.build_cpe.ProductCPEs(product_yaml)
 
-    # get CPE by name and verify it's loaded
+    # get a product CPE by name and verify it's loaded
+    # this CPE is defined in `DATADIR/product.yml`
     rhel7_cpe = product_cpes.get_cpe("rhel7")
     assert(rhel7_cpe.name == "cpe:/o:redhat:enterprise_linux:7")
     assert(rhel7_cpe.title == "Red Hat Enterprise Linux 7")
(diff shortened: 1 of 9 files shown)


vojtapolasek:
Hello, it looks great! Could you please just fix the pep8 issue?

tests/unit/ssg-module/test_build_yaml.py
```diff
@@ -127,3 +130,67 @@ def test_priority_ordering():
...
+def test_platform_from_text_unknown_platform(env_yaml):
+    with pytest.raises(ssg.build_cpe.CPEDoesNotExist):
+        ssg.build_yaml.Platform.from_text("something_bogus", env_yaml)
+
+
+def test_platform_from_text_simple(env_yaml):
+    platform = ssg.build_yaml.Platform.from_text("machine", env_yaml)
+    assert platform.to_ansible_conditional() == "ansible_virtualization_type not in [\"docker\", \"lxc\", \"openvz\", \"podman\", \"container\"]"
+    assert platform.to_bash_conditional() == "[ ! -f /.dockerenv ] && [ ! -f /run/.containerenv ]"
+    assert platform.to_xml_element() == b'<?xml version=\'1.0\' encoding=\'utf8\'?>\n<ns0:platform xmlns:ns0="http://cpe.mitre.org/language/2.0" id="machine"><ns0:logical-test operator="AND" negate="false"><ns0:fact-ref name="cpe:/a:machine" /></ns0:logical-test></ns0:platform>'
```
matejak, line 154: I would not write those asserts so literally, as the RHS can change for legitimate reasons, and it would require update of the test code for no reasons.
I suggest to simplify those asserts, so unit tests make sure that things look right, but they the check for the exact "right" to integration tests or system tests that are able to pass even when the syntax changes.

```suggestion
    assert "ansible_virtualization_type" in platform.to_ansible_conditional()
    assert "containerenv" in platform.to_bash_conditional()
    assert platform.to_xml_element()'
```

tests/unit/ssg-module/test_build_cpe.py
```diff
@@ -90,3 +92,49 @@ def test_extract_referred_nodes():
 
     assert len(results) == 1
     assert results[0].text == 'Source Code'
+
+
+def test_product_cpes():
```
matejak, line 97: What are tests in this file testing?
Is it the ability to finding a file and translate its contents?
If that's the case, I would suggest to reduce those checks, or to at least compare cpe2 attribute to cpe1 attributes and not to a literal value.
That way, tests will be more indicative regarding what they are testing, and they will be somewhat less dependent on the test data.

matejak (changes requested):
The idea of unit tests is to test a functionality very close to where it is implemented, and that's not what tests here do - the testing environment needs to be set up in a relatively opaque way, and the test can start to fail in case that implementation details change and e.g. the XML whitespace changes, another "non-machine" platform is added, etc.

Therefore, I propose to make tests in this PR less precise and more robust. And in the longer term, we should restructure the code so we can test it in a more streamlined approach.

tests/unit/ssg-module/test_build_cpe.py
jan-cerny: The tests in this file (`ssg-module/test_build_cpe.py`) are unit tests for classes and functions in `/ssg/build_cpe.py`. Specifically, the 2 test cases added in this PR (`test_product_cpes` and `test_content_cpes`) are testing the `ProductCPEs` class. These test cases are focused on testing the `get_cpe` method. This method has been chosen to be tested because other methods of this class are either private or used only inside the constructor of this class, except the `get_cpe_name` method, but that only uses the `get_cpe` method internally. Basically, the following features of the `get_cpe` method are tested in the tests:
* it accepts both CPE name and CPE ID
* it returns the same `CPEItem` object if a CPE name is passed to the CPE as if a CPE ID is passed to the method
* it returns `CPEItem` instances for CPEs that were defined as so-called "product CPEs" (CPEs defined in current product's `product.yml`)
* it returns `CPEItem` instances for CPEs that were defined as so-called "content CPEs" (CPEs defined under the `applicability` directory)
* the returned `CPEItem` instances contain all the data

So unfortunately, it tests also the constructor, which in turn uses the other methods in the class, which means that almost whole class is tested by this. I think that is probably the cause of your concern. Would you prefer to add other tests cases that would test also other methods?

> or to at least compare cpe2 attribute to cpe1 attributes and not to a literal value.

What do you mean by that?

tests/unit/ssg-module/test_build_yaml.py
jan-cerny: Great!

tests/unit/ssg-module/test_build_cpe.py
matejak: I see, I would prefer if it was more obvious what is going on, the test doesn't indicate at all that it depends on the contents of the "applicability" directory. Ideally, one should be able to make the code to load CPEs from mock product data and from mock applicability data. But this is an issue with the tested code, not with tests.
I would then just ask to put this as a comment to the content test.

tests/unit/ssg-module/test_build_cpe.py
```diff
@@ -90,3 +92,49 @@ def test_extract_referred_nodes():
...
+    rhel7_cpe = product_cpes.get_cpe("rhel7")
+    assert(rhel7_cpe.name == "cpe:/o:redhat:enterprise_linux:7")
+    assert(rhel7_cpe.title == "Red Hat Enterprise Linux 7")
+    assert(rhel7_cpe.check_id == "installed_OS_is_rhel7")
+    assert(rhel7_cpe.bash_conditional == "")
+    assert(rhel7_cpe.ansible_conditional == "")
+
+    # get CPE by ID and verify it's loaded
+    rhel7_cpe_2 = product_cpes.get_cpe("cpe:/o:redhat:enterprise_linux:7")
+    assert(rhel7_cpe_2.name == "cpe:/o:redhat:enterprise_linux:7")
```
matejak, line 113: >>    or to at least compare cpe2 attribute to cpe1 attributes and not to a literal value.
>
> What do you mean by that?

```suggestion
    assert(rhel7_cpe_2.name == rhel7_cpe.name)
```

and so on - this strongly indicates that those returned CPE classes should be the same, and there is less copy-pasting of text.

jan-cerny:
I have slightly updated the tests

matejak:
Great, that's definitely much better, and there is not much to improve here - thank you very much for this PR!

Rough edges in tests point to design issues of the code, so next time we develop new features, let's write tests concurrently.

matejak (approved):

yuumasato:
The PR was merged into `cpe_logical_expressions`, so I'm clearing the milestone.

PR #8008, opened 2021-12-17
Mark as Manual RHEL7 CIS items that are manual

yuumasato:
#### Description:

- Put `Manual` status to manual items from RHEL7 CIS benchmark.

#### Rationale:

- This way we are more explicit about items that we don't plan to implement
- The manual controls are reported as such in `controleval.py` and facilitate assessment of the implementation level

### DIFF
--- a/controls/cis_rhel7.yml
+++ b/controls/cis_rhel7.yml
@@ -264,14 +264,14 @@ controls:
     levels:
     - l1_server
     - l1_workstation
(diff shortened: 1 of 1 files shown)


marcusburghardt (approved):
Looks good. Just waiting the last missing test to finish.

Mab879 (approved):

PR #8015, opened 2021-12-21
Fix chrony maxpoll description, include peer in bash and oval

matejak:
This is a rebase + extension of #7107 to the new master.

Concerns raised in https://github.com/ComplianceAsCode/content/pull/7107#pullrequestreview-697300751 were mostly addressed by PRs that were merged in the meantime.

### DIFF
--- a/linux_os/guide/services/ntp/chronyd_or_ntpd_set_maxpoll/rule.yml
+++ b/linux_os/guide/services/ntp/chronyd_or_ntpd_set_maxpoll/rule.yml
@@ -9,7 +9,7 @@ description: |-
     {{{ xccdf_value("var_time_service_set_maxpoll") }}} in <tt>/etc/ntp.conf</tt> or
     <tt>/etc/chrony.conf</tt> to continuously poll time servers. To configure
     <tt>maxpoll</tt> in <tt>/etc/ntp.conf</tt> or <tt>/etc/chrony.conf</tt>
-    add the following after each `server` entry:
+    add the following after each `server`, `pool` or `peer` entry:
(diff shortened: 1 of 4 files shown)


linux_os/guide/services/ntp/chronyd_or_ntpd_set_maxpoll/rule.yml
```diff
@@ -9,7 +9,7 @@ description: |-
     {{{ xccdf_value("var_time_service_set_maxpoll") }}} in <tt>/etc/ntp.conf</tt> or
     <tt>/etc/chrony.conf</tt> to continuously poll time servers. To configure
     <tt>maxpoll</tt> in <tt>/etc/ntp.conf</tt> or <tt>/etc/chrony.conf</tt>
-    add the following:
+    add the following after each `server` entry:
```
marcusburghardt, line 12: Since we are actually correctly accepting `pool` and `peer` properties as well in the OVAL, looks reasonable make this information explicit in the description.

marcusburghardt (changes requested):

marcusburghardt (commented):
Tests worked fine using Ansible and Bash in RHEL8 and RHEL9 vms.

I only realized now that we don't have any test scenario considering the `peer` entry.
I wouldn't say this is mandatory for this case, since the already tested regex was only incremented. But also doesn't look complex to include. So, it is your call Matej. : )

Finally, the error reported looks to be related to the test environment, which apparently is missing the package `oval-graph`, but I couldn't investigate this deeply for now.

jan-cerny:
@matejak @marcusburghardt Any updates? Feel free to close if you want.

marcusburghardt (approved):

PR #8026, opened 2022-01-05
Use authselect to edit pam files if it is present

### LINKED ISSUE #8022: pam_faillock.so rules fail after remediation
#### Description of problem:
PAM related rules fail after hardened kickstart installation and also after clean system remediation. 

05831dc

RHEL7, RHEL8, RHEL9

#### Steps to Reproduce:

1. `python3 tests/test_suite.py profile --libvirt qemu:///system test_suite_vm --datastream /tmp/ssg-rhel9-ds.xml --xccdf-id scap_org.open-scap_cref_ssg-rhel9-xccdf-1.2.xml --mode online --remediate-using oscap xccdf_org.ssgproject.content_profile_ospp`

#### Actual Results:
```
There are unexpected "fail" results:
xccdf_org.ssgproject.content_rule_accounts_passwords_pam_faillock_deny - fail
xccdf_org.ssgproject.content_rule_accounts_passwords_pam_faillock_deny_root - fail
xccdf_org.ssgproject.content_rule_accounts_passwords_pam_faillock_interval - fail
xccdf_org.ssgproject.content_rule_accounts_passwords_pam_faillock_unlock_time - fail
```

#### Expected Results:
PAM rules pass

marcusburghardt:
#### Description:

The remediation included in no_empty_passwords rule was editing pam
files directly, making them inconsistent for the `authselect` tool.

#### Rationale:

It is not recommended to edit pam files directly when a proper tool exists.

- Partially Fixes #8022

### DIFF
--- a/linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/ansible/shared.yml
+++ b/linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/ansible/shared.yml
@@ -41,9 +41,9 @@
       - result_authselect_check_cmd is success
     fail_msg:
     - authselect integrity check failed. Remediation aborted!
-    - This remediation could not be applied because the authselect profile is not integer, probably due to manual edition.
+    - This remediation could not be applied because the authselect profile is not intact.
+    - It is not recommended to manually edit the PAM files when authselect is available
     - In cases where the default authselect profile does not cover a specific demand, a custom authselect profile is recommended.
-    - Where authselect is in place, it is not recommended to manually edit pam files.
     success_msg:
     - authselect integrity check passed
 
--- a/linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/ansible/shared.yml
+++ b/linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/ansible/shared.yml
@@ -35,14 +35,17 @@
     - result_authselect_check_cmd is success
     - result_authselect_features.stdout is not search("without-nullok")
 
-- name: Informative message if the authselect integrity check fails
-  debug:
-    msg:
-    - This remediation could not be applied because the authselect profile was manually modified.
+- name: Informative message based on the authselect integrity check result
+  ansible.builtin.assert:
+    that:
+      - result_authselect_check_cmd is success
+    fail_msg:
+    - authselect integrity check failed. Remediation aborted!
+    - This remediation could not be applied because the authselect profile is not integer, probably due to manual edition.
     - In cases where the default authselect profile does not cover a specific demand, a custom authselect profile is recommended.
     - Where authselect is in place, it is not recommended to manually edit pam files.
-  when:
-    - result_authselect_check_cmd is failed
+    success_msg:
+    - authselect integrity check passed
 
 - name: Ensure nullok property is absent by directly editing pam files
   replace:
--- a/linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/ansible/shared.yml
+++ b/linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/ansible/shared.yml
@@ -17,14 +17,6 @@
   when:
     - result_authselect_present.stat.exists
 
-- name: Get authselect current profile
-  ansible.builtin.shell:
-    cmd: authselect current -r | awk '{ print $1 }'
-  register: result_authselect_profile
-  changed_when: false
-  when:
-    - result_authselect_check_cmd is failed
-
 - name: Get authselect current features
   ansible.builtin.shell:
     cmd: authselect current | tail -n+3 | awk '{ print $2 }'
@@ -32,40 +24,27 @@
   changed_when: false
   when:
     - result_authselect_present.stat.exists
+    - result_authselect_check_cmd is success
 
-- name: Define the current authselect profile as a local fact
-  ansible.builtin.set_fact:
-    authselect_profile: "{{ result_authselect_profile.stdout }}"
-  when:
-    - result_authselect_check_cmd is failed
-    - result_authselect_profile is not skipped
-
-- name: Restore the authselect profile if manually modified
-  ansible.builtin.command:
-    cmd: authselect select {{ authselect_profile }} --force
-  register: result_pam_authselect_restore_profile
-  when:
-    - result_authselect_check_cmd is failed
-    - result_authselect_profile is not skipped
-
-- name: Restore the authselect features if manually modified
-  ansible.builtin.command:
-    cmd: authselect enable-feature {{ item }}
-  register: result_pam_authselect_restore_features
-  loop: "{{ result_authselect_features.stdout_lines }}"
-  when:
-    - result_authselect_check_cmd is failed
-    - result_authselect_features is not skipped
-
-- name: "Ensure nullok property is absent via authselect tool"
+- name: Ensure nullok property is absent via authselect tool
   ansible.builtin.command:
     cmd: authselect enable-feature without-nullok
   register: result_authselect_cmd
   when:
     - result_authselect_present.stat.exists
+    - result_authselect_check_cmd is success
     - result_authselect_features.stdout is not search("without-nullok")
 
-- name: "Ensure nullok property is absent by directly editing pam files"
+- name: Informative message if the authselect integrity check fails
+  debug:
+    msg:
+    - This remediation could not be applied because the authselect profile was manually modified.
+    - In cases where the default authselect profile does not cover a specific demand, a custom authselect profile is recommended.
+    - Where authselect is in place, it is not recommended to manually edit pam files.
+  when:
+    - result_authselect_check_cmd is failed
+
+- name: Ensure nullok property is absent by directly editing pam files
   replace:
     dest: "{{ item }}"
     regexp: 'nullok'
--- a/linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/ansible/shared.yml
+++ b/linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/ansible/shared.yml
@@ -1,14 +1,76 @@
-# platform = multi_platform_rhel,multi_platform_fedora,multi_platform_ol,multi_platform_rhv,multi_platform_sle
+# platform = multi_platform_rhel,multi_platform_fedora,multi_platform_ol,multi_platform_rhv
 # reboot = false
 # strategy = configure
 # complexity = low
 # disruption = medium
-- name: "Prevent Log In to Accounts With Empty Password - system-auth"
-  replace:
-    dest: /etc/pam.d/system-auth
-    regexp: 'nullok'
+- name: Check if system relies on authselect
+  ansible.builtin.stat:
+    path: /usr/bin/authselect
+  register: result_authselect_present
 
-- name: "Prevent Log In to Accounts With Empty Password - password-auth"
+- name: Check integrity of authselect current profile
+  ansible.builtin.command:
+    cmd: authselect check
+  register: result_authselect_check_cmd
+  changed_when: false
+  ignore_errors: yes
+  when:
+    - result_authselect_present.stat.exists
+
+- name: Get authselect current profile
+  ansible.builtin.shell:
+    cmd: authselect current -r | awk '{ print $1 }'
+  register: result_authselect_profile
+  changed_when: false
+  when:
+    - result_authselect_check_cmd is failed
+
+- name: Get authselect current features
+  ansible.builtin.shell:
+    cmd: authselect current | tail -n+3 | awk '{ print $2 }'
+  register: result_authselect_features
+  changed_when: false
+  when:
+    - result_authselect_present.stat.exists
+
+- name: Define the current authselect profile as a local fact
+  ansible.builtin.set_fact:
+    authselect_profile: "{{ result_authselect_profile.stdout }}"
+  when:
+    - result_authselect_check_cmd is failed
+    - result_authselect_profile is not skipped
+
+- name: Restore the authselect profile if manually modified
+  ansible.builtin.command:
+    cmd: authselect select {{ authselect_profile }} --force
+  register: result_pam_authselect_restore_profile
+  when:
+    - result_authselect_check_cmd is failed
+    - result_authselect_profile is not skipped
+
+- name: Restore the authselect features if manually modified
+  ansible.builtin.command:
+    cmd: authselect enable-feature {{ item }}
+  register: result_pam_authselect_restore_features
+  loop: "{{ result_authselect_features.stdout_lines }}"
+  when:
+    - result_authselect_check_cmd is failed
+    - result_authselect_features is not skipped
+
+- name: "Ensure nullok property is absent via authselect tool"
+  ansible.builtin.command:
+    cmd: authselect enable-feature without-nullok
+  register: result_authselect_cmd
+  when:
+    - result_authselect_present.stat.exists
+    - result_authselect_features.stdout is not search("without-nullok")
+
+- name: "Ensure nullok property is absent by directly editing pam files"
   replace:
-    dest: /etc/pam.d/password-auth
+    dest: "{{ item }}"
     regexp: 'nullok'
+  loop:
+    - /etc/pam.d/system-auth
+    - /etc/pam.d/password-auth
+  when:
+    - not result_authselect_present.stat.exists
--- a/linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/bash/shared.sh
+++ b/linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/bash/shared.sh
@@ -1,3 +1,20 @@
 # platform = multi_platform_wrlinux,multi_platform_rhel,multi_platform_fedora,multi_platform_ol,multi_platform_rhv,multi_platform_sle
-sed --follow-symlinks -i 's/\<nullok\>//g' /etc/pam.d/system-auth
-sed --follow-symlinks -i 's/\<nullok\>//g' /etc/pam.d/password-auth
+SYSTEM_AUTH="/etc/pam.d/system-auth"
+PASSWORD_AUTH="/etc/pam.d/password-auth"
+
+if [ -f /usr/bin/authselect ]; then
+    if [ $(authselect enable-feature without-nullok) ]; then
+        authselect apply-changes
+    else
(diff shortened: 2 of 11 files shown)


linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/ansible/shared.yml
```diff
@@ -1,14 +1,72 @@
...
+    - result_authselect_check_cmd is failed
+    - result_authselect_profile | default(false) | bool
+
+- name: Restore the authselect profile if manually modified
+  ansible.builtin.command:
+    cmd: authselect select {{ authselect_profile }} --force
+  register: result_pam_authselect_restore_profile
+  when:
+    - result_authselect_check_cmd is failed
+    - result_authselect_profile | default(false) | bool
```
yuumasato, line 46: Is it a debugging leftover? Won't the `default(false)` cause the when condition to always fail?

linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/ansible/shared.yml
```diff
@@ -1,14 +1,72 @@
...
+    - result_authselect_check_cmd is failed
+
+- name: Define the current authselect profile as a local fact
+  ansible.builtin.set_fact:
+    authselect_profile: "{{ result_authselect_profile.stdout }}"
+  when:
+    - result_authselect_check_cmd is failed
+    - result_authselect_profile | default(false) | bool
+
+- name: Restore the authselect profile if manually modified
```
yuumasato, line 40: Is it a good approach to remove any changes if the pam file was changed by a third party?

I think it makes sense for our remediations not to make direct edits when `authselect` is available in the system.
But if a change was made by something else, we should probably:
-  not make any changes, or
- fall back to the direct edit too.

Thoughts?

linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/ansible/shared.yml
```diff
@@ -1,14 +1,72 @@
...
-- name: "Prevent Log In to Accounts With Empty Password - system-auth"
-  replace:
-    dest: /etc/pam.d/system-auth
-    regexp: 'nullok'
+- name: Check if system relies on authselect
+  ansible.builtin.stat:
+    path: /usr/bin/authselect
+  register: result_authselect_present
+
+- name: Check integrity of authselect current profile
```
yuumasato, line 11: I think this task should have `changed_when: false`, so that it doesn't report as changed for a command that only collects data.

linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/ansible/shared.yml
```diff
@@ -1,14 +1,72 @@
...
+- name: Restore the authselect features if manually modified
+  ansible.builtin.command:
+    cmd: authselect enable-feature {{ item }}
+  register: result_pam_authselect_restore_features
+  loop: "{{ result_authselect_features.stdout_lines }}"
+  when:
+    - result_pam_authselect_restore_profile
+    - result_authselect_features | default(false) | bool
+
+- name: "Ensure nullok property is not present via authselect tool"
```
yuumasato, line 57: This task is not converging, it always reports as `changed.

linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/ansible/shared.yml
yuumasato: I made a manual change to the pam file and the task was skipped.
```
ansible-playbook -i "rhel85," --tags no_empty_passwords build/ansible/rhel8-playbook-stig.yml 

PLAY [all] ***********************************************************************************************************

TASK [Gathering Facts] ***********************************************************************************************
ok: [rhel85]

TASK [Check if system relies on authselect] **************************************************************************
ok: [rhel85]

TASK [Check integrity of authselect current profile] *****************************************************************
fatal: [rhel85]: FAILED! => {"changed": true, "cmd": ["authselect", "check"], "delta": "0:00:00.006320", "end": "2021-11-08 09:38:18.257149", "msg": "non-zero return code", "rc": 3, "start": "2021-11-08 09:38:18.250829", "stderr": "[error] [/etc/authselect/system-auth] has unexpected content!", "stderr_lines": ["[error] [/etc/authselect/system-auth] has unexpected content!"], "stdout": "Current configuration is not valid. It was probably modified outside authselect.", "stdout_lines": ["Current configuration is not valid. It was probably modified outside authselect."]}
...ignoring

TASK [Get authselect current profile] ********************************************************************************
changed: [rhel85]

TASK [Get authselect current features] *******************************************************************************
changed: [rhel85]

TASK [Define the current authselect profile as a local fact] *********************************************************
skipping: [rhel85]

TASK [Restore the authselect profile if manually modified] ***********************************************************
skipping: [rhel85]

TASK [Restore the authselect features if manually modified] **********************************************************
[DEPRECATION WARNING]: evaluating 'result_pam_authselect_restore_profile' as a bare variable, this behaviour will go 
away and you might need to add |bool to the expression in the future. Also see CONDITIONAL_BARE_VARS configuration 
toggle. This feature will be removed in version 2.12. Deprecation warnings can be disabled by setting 
deprecation_warnings=False in ansible.cfg.
skipping: [rhel85] => (item=without-nullok) 

TASK [Ensure nullok property is not present via authselect tool] *****************************************************
fatal: [rhel85]: FAILED! => {"changed": true, "cmd": ["authselect", "enable-feature", "without-nullok"], "delta": "0:00:00.007492", "end": "2021-11-08 09:38:19.853987", "msg": "non-zero return code", "rc": 4, "start": "2021-11-08 09:38:19.846495", "stderr": "[error] [/etc/authselect/system-auth] has unexpected content!\n[error] Unexpected changes to the configuration were detected.\n[error] Refusing to activate profile unless those changes are removed or overwrite is requested.\nUnable to enable feature [17]: File exists", "stderr_lines": ["[error] [/etc/authselect/system-auth] has unexpected content!", "[error] Unexpected changes to the configuration were detected.", "[error] Refusing to activate profile unless those changes are removed or overwrite is requested.", "Unable to enable feature [17]: File exists"], "stdout": "", "stdout_lines": []}

PLAY RECAP ***********************************************************************************************************
rhel85                     : ok=5    changed=3    unreachable=0    failed=1    skipped=3    rescued=0    ignored=1   
```

yuumasato (changes requested):
I wonder if adding a test scenario with manual changes would be interesting.

linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/ansible/shared.yml
```diff
@@ -1,14 +1,72 @@
-# platform = multi_platform_rhel,multi_platform_fedora,multi_platform_ol,multi_platform_rhv,multi_platform_sle
+# platform = multi_platform_rhel,multi_platform_fedora,multi_platform_ol,multi_platform_rhv
```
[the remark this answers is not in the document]
marcusburghardt: There are separate Ansible remediation playbooks for sle12 and sle15.

linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/ansible/shared.yml
marcusburghardt: This approach related to the `[DEPRECATION WARNING]` mentioned there. Let me check this conditional again, since I tested this case before... : (

linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/ansible/shared.yml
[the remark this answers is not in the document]
yuumasato: Oh, missed that, :stuck_out_tongue:

linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/ansible/shared.yml
marcusburghardt: Yes, this is a good approach in my point of view.

Doing manual changes when the `authselect` is present makes the pam files inconsistent. It means any other integration or even new changes can no longer happen automatically. It breaks the pam structure in a way that only manual changes are possible, which may bring at least two risks related to:
- Availability, considering how sensible the pam files are to ordering and syntax.
- Malicious activity, since manual editing of these files could be maliciously prepared to workaround the controls in place.

This remediation turns the system back to the consistent state preserving possible features previously enabled and also ensuring a validated configuration is in use.

Indeed, if any change was make directly in pam files, going against the best practices, they will be lost (for good reasons).

This specific rule deals with `nullok` parameter, which has a specific feature out-of-box in `authselect`. For other pam related rules (which I am already fixing too), the approach in cases `authselect` doesn't have a feature out-of-box, is to work with custom profiles managed by `authselect`.

linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/ansible/shared.yml
marcusburghardt: Good remind. I will do that. Thanks

marcusburghardt: Done.

marcusburghardt:
> I wonder if adding a test scenario with manual changes would be interesting.

Included `authselect_modified_pam.fail.sh`.

linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/ansible/shared.yml
marcusburghardt: Fixed. Thanks.

linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/ansible/shared.yml
marcusburghardt: Improved the conditional for this task. Fixed.

linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/bash/shared.sh
```diff
@@ -1,3 +1,20 @@
 # platform = multi_platform_wrlinux,multi_platform_rhel,multi_platform_fedora,multi_platform_ol,multi_platform_rhv,multi_platform_sle
-sed --follow-symlinks -i 's/\<nullok\>//g' /etc/pam.d/system-auth
-sed --follow-symlinks -i 's/\<nullok\>//g' /etc/pam.d/password-auth
+SYSTEM_AUTH="/etc/pam.d/system-auth"
+PASSWORD_AUTH="/etc/pam.d/password-auth"
+
+if [ -f /usr/bin/authselect ]; then
+    if [ $(authselect enable-feature without-nullok) ]; then
```
yuumasato, line 6: I wonder in what case this `enable-feature` would fail where the second `enable-feature`, in the `else` would not.

linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/ansible/shared.yml
yuumasato: > Yes, this is a good approach in my point of view.
> 
> Doing manual changes when the `authselect` is present makes the pam files inconsistent. It means any other integration or even new changes can no longer happen automatically. It breaks the pam structure in a way that only manual changes are possible, which may bring at least two risks related to:
> 
>     * Availability, considering how sensible the pam files are to ordering and syntax.
> 
>     * Malicious activity, since manual editing of these files could be maliciously prepared to workaround the controls in place.

When one edits the PAM files without `authselect` one assumes the risks you mention. Maybe this should be more clear in the rule description? (And BTW `authselect` is not mentioned at all in the rule description)

> This remediation turns the system back to the consistent state preserving possible features previously enabled and also ensuring a validated configuration is in use.
> 
> Indeed, if any change was make directly in pam files, going against the best practices, they will be lost (for good reasons).

I'm not convinced that reverting third-party changes is a good experience, we should at least mention this in rule description.

What I think the remediation should do is warn about inconsistent state and recommend the admin to include them in a custom profile. In other words, if an admin is not using `authselect` we should not force them, but invite them to use it.

> 
> This specific rule deals with `nullok` parameter, which has a specific feature out-of-box in `authselect`. For other pam related rules (which I am already fixing too), the approach in cases `authselect` doesn't have a feature out-of-box, is to work with custom profiles managed by `authselect`.

The change in the PAM files may be related to something else, not necessarily `nullok`.

yuumasato: > But if a change was made by something else, we should probably:
>
>    not make any changes, or
>    fall back to the direct edit too.

One could argue that we can just do direct edits when there are changes outside of `authselect`, because we need to maintain this approach for RHEL7 anyway.  The downside is that the rule will be remediated and the admin will not be nudged to use `authselect`.
So just I lean towards not making any changes.

linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/bash/shared.sh
marcusburghardt: The first case will result success (true) if the profile is integer. In this case, we don't need to force the profile selection. In case this command fails, it means the authselect profile is not integer. For this case, we reactivate it, preserving the features but removing manual modifications which caused the integrity check to fail.

In Ansible, the logic is the same, but instead or directly trying to apply the change, the check option is used in order to make the Ansible tasks easier to be executed by conditionals.

linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/ansible/shared.yml
marcusburghardt: > > Yes, this is a good approach in my point of view.
> > Doing manual changes when the `authselect` is present makes the pam files inconsistent. It means any other integration or even new changes can no longer happen automatically. It breaks the pam structure in a way that only manual changes are possible, which may bring at least two risks related to:
> > ```
> > * Availability, considering how sensible the pam files are to ordering and syntax.
> > 
> > * Malicious activity, since manual editing of these files could be maliciously prepared to workaround the controls in place.
> > ```
> 
> When one edits the PAM files without `authselect` one assumes the risks you mention. Maybe this should be more clear in the rule description? (And BTW `authselect` is not mentioned at all in the rule description)

I agree that this should be more explicit here.
> 
> > This remediation turns the system back to the consistent state preserving possible features previously enabled and also ensuring a validated configuration is in use.
> > Indeed, if any change was make directly in pam files, going against the best practices, they will be lost (for good reasons).
> 
> I'm not convinced that reverting third-party changes is a good experience, we should at least mention this in rule description.
> 
I can't remember an example of current integrations with deals directly in pam files for distros which relies on `authselect`. The integrations usually happen directly on sssd layer.

> What I think the remediation should do is warn about inconsistent state and recommend the admin to include them in a custom profile. In other words, if an admin is not using `authselect` we should not force them, but invite them to use it.
> 
> > This specific rule deals with `nullok` parameter, which has a specific feature out-of-box in `authselect`. For other pam related rules (which I am already fixing too), the approach in cases `authselect` doesn't have a feature out-of-box, is to work with custom profiles managed by `authselect`.
> 
> The change in the PAM files may be related to something else, not necessarily `nullok`.

That's what I meant. : ) In other words, where the `authselect` has a feature out-of-box, we can use it in the current profile. Otherwise, a custom profile should be used, like in https://github.com/ComplianceAsCode/content/pull/8030

linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/rule.yml
```diff
@@ -70,3 +70,11 @@ ocil: |-
...
     with empty passwords. Remove any instances of the <tt>nullok</tt> option to
     prevent logins with empty passwords.
+
+warnings:
+    - general: |-
+       If the system relies on <tt>authselect</tt> tool to manage pam settings, the remediation
+       will also use <tt>authselect</tt> tool. However, if any manual modification was made in
+       pam files, the <tt>authselect</tt> integrity check will fail and the remediation will be
+       aborted in order to preserve intentional changes. In this case, an informative message will
+       be shown in the remediation report.
```
yuumasato, line 80: Update any mention of `pam` to `PAM`.

linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/tests/authselect_modified_pam.fail.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+# packages = authselect
+# platform = Red Hat Enterprise Linux 8,Red Hat Enterprise Linux 9,multi_platform_fedora
+# remediation = none
+
+SYSTEM_AUTH_FILE="/etc/pam.d/system-auth"
+
+# This modification will break the integrity checks done by authselect. The remediation
+# should be robust enough to deal with possible manual and unvalidated modification.
```
yuumasato, line 9: This is comment is a bit outdated.
The remediation doesn't "deal with possible manual remediations", it aborts the remediation when it finds the authselect profiles have been edited by a third party.

linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/ansible/shared.yml
```diff
@@ -1,14 +1,58 @@
...
+
+- name: Informative message based on the authselect integrity check result
+  ansible.builtin.assert:
+    that:
+      - result_authselect_check_cmd is success
+    fail_msg:
+    - authselect integrity check failed. Remediation aborted!
+    - This remediation could not be applied because the authselect profile is not integer, probably due to manual edition.
+    - In cases where the default authselect profile does not cover a specific demand, a custom authselect profile is recommended.
+    - Where authselect is in place, it is not recommended to manually edit pam files.
```
yuumasato, line 46: I'd flip lines 45 and 46, while making edits on line 45:
`It is not recommended to manually edit the PAM files when authselect is available.`

yuumasato (changes requested):
The rule and remediation are working as expected.
Just a few text changes are needed.

yuumasato (approved):

PR #8033, opened 2022-01-06
Clarify local ssg package dependency

rhmdnd:
The ssg package within this repository is used within controleval.py and
it's also the name of an official package available via pypi. New users
might confuse the local import for the publicly available package, which
will lead to even more confusion since they're obviously two different
packages.

This commit adds a comment to the script to clarify those details for
readers. It also provides additional details as a RuntimeError exception
so that the caller knows what to do if the script fails to import the
local ssg package.

### DIFF
--- a/utils/controleval.py
+++ b/utils/controleval.py
@@ -3,8 +3,31 @@ import collections
 import argparse
 import os
 
-from ssg import controls
-import ssg.products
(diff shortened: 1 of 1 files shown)


Vincent056 (approved):
/lgtm

jan-cerny (approved):

PR #8029, opened 2022-01-05
OCP4: api_server_no_adm_ctrl_plugins_disabled ignore PodSecurity

Vincent056:
We are going to ignore PodSecurity from the disabled admission control plugins list because it is validating only and it is disabled by default on 4.10 due to the transition from SCCs to PSs

Related Link: https://github.com/openshift/enhancements/pull/899/files

PR #8035, opened 2022-01-06
OCP4: default_ingress_ca_replaced remove manual remediation

Vincent056:
We should remove manual remediation, as it will stop the IdP test routes from working.

### DIFF
--- a/applications/openshift/networking/default_ingress_ca_replaced/tests/ocp4/e2e.yml
+++ b/applications/openshift/networking/default_ingress_ca_replaced/tests/ocp4/e2e.yml
@@ -1,3 +1,2 @@
 ---
 default_result: FAIL
-result_after_remediation: PASS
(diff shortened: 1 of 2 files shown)


mrogers950:
Looks good to me considering your findings from the failed tests:
```
   The configuration is invalid for proxy 'cluster' (failed to validate configmap reference for proxy trustedCA 'testing': failed to get trustedCA configmap for proxy cluster: ConfigMap "testing" not found). Use 'oc edit proxy.config.openshift.io cluster' to fix.
```

applications/openshift/networking/default_ingress_ca_replaced/tests/ocp4/e2e.yml
```diff
@@ -1,3 +1,2 @@
 ---
-default_result: FAIL
-result_after_remediation: PASS
```
rhmdnd, line 3: Discussed this with @Vincent056 and @mrogers950 and it sounds like another potential option is to supply a valid ConfigMap with a legitimate CA, instead of a fake one for testing that fails later on.

This would allow us to keep the remediation in place and not clean it up later.

Vincent056:
Close this PR, included the commit in https://github.com/ComplianceAsCode/content/pull/8029

PR #8036, opened 2022-01-07
Suggest running command for setting grub2 password as root

mildas:
#### Description:
The grub bootloader password may only be set by root.

#### Rationale:
See `Additional info:` in https://bugzilla.redhat.com/show_bug.cgi?id=2021093


vojtapolasek (approved):
Good catch, thank you.

PR #8038, opened 2022-01-07
Add better error messages to utils/controleval.py and add does not meet to stats output

Mab879:
#### Description:

Adds an error message for the case when the a policy has a level
defined but has no controls. This commit also adds the 'does not meet'
to the stats output

#### Rationale:
Before the script would give an divide by zero error. Also, added a new status to the script to aid in creating control files.

### DIFF
--- a/utils/controleval.py
+++ b/utils/controleval.py
@@ -42,6 +42,10 @@ def calculate_stats(ctrls):
     ctrlstats = collections.defaultdict(int)
     ctrllist = collections.defaultdict(set)
 
+    if total == 0:
+        print("No controls founds with the given inputs. Maybe try another level.")
+        exit(1)
+
     for ctrl in ctrls:
         ctrlstats[str(ctrl.status)] += 1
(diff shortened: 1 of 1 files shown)


utils/controleval.py
```diff
@@ -60,6 +64,7 @@ def calculate_stats(ctrls):
     print_specific_stat("Supported", ctrlstats[controls.Status.SUPPORTED], applicable)
     print_specific_stat("Documentation", ctrlstats[controls.Status.DOCUMENTATION], applicable)
     print_specific_stat("Inherently Met", ctrlstats[controls.Status.INHERENTLY_MET], applicable)
+    print_specific_stat("Does Not Met", ctrlstats[controls.Status.DOES_NOT_MEET], applicable)
```
jan-cerny, line 67:
```suggestion
    print_specific_stat("Does Not Meet", ctrlstats[controls.Status.DOES_NOT_MEET], applicable)
```

jan-cerny (approved):

PR #8039, opened 2022-01-07
Improvements to controls and STIG export

Mab879:
#### Description:

- Adds fix and check to control
- Add the ability override the status of a control with rules in `create_srg_export.py`.

#### Rationale:
Improvements needed for RHEL9 STIG.


jan-cerny (changes requested):
The new items fix and checks need to be tested in the unit tests.

jan-cerny (approved):

PR #8041, opened 2022-01-10
Handle long diffs gracefully in compare-ds github action

ggbecker:
#### Description:

- Handle long diffs gracefully in compare-ds github action

#### Rationale:

- Apparently it's not possible to place comments larger than 65k characters so we have to limit them.

Tested in https://github.com/ggbecker/content/pull/26#issuecomment-1008924213
 https://github.com/ggbecker/content/runs/4762507114?check_suite_focus=true

Co-authored with @matejak


.github/workflows/compare-ds.yaml
```diff
@@ -106,6 +106,26 @@ jobs:
...
+          edit-mode: replace
+      - name: Create or update a trimmed comment
+        if: ${{ steps.compare_ds.outputs.COMPARE_DS_OUTPUT_SIZE > 65000 }}
+        uses: peter-evans/create-or-update-comment@v1
+        with:
+          comment-id: ${{ steps.fc.outputs.comment-id }}
+          issue-number: ${{ github.event.pull_request.number }}
+          body: |
+            This datastream diff is auto generated by the check `Compare DS/Generate Diff`.
+            Due to the excessive size of the diff, it has been trimmed to fit the 4096-character limit.
```
matejak, line 119: The limit is no longer 4096 characters.
It is probably based on 16-bit integer (0 to 65535), so setting the limit body to 65000 is sufficient - the surrounding text can't get longer that easily.
Initially I thought that 8000+ characters is too much, but I don't think so any more, the diff in the example PR is perfectly usable.

matejak (changes requested):
The comments needs a typo fixed, otherwise LGTM!

.github/workflows/compare-ds.yaml
ggbecker: Fixed by amending the commit. The limit is somehow connected to: https://github.com/actions/virtual-environments/issues/3257 but I haven't really spend too much time investigating it. I believe the 65k chars is enough for our purposes.

matejak (approved):

PR #8043, opened 2022-01-11
Implement handling of logical expressions in platform definitions

vojtapolasek:
#### Description:

- platform is now represented as a class, it is also stored along compiled rules
- a parsing mechanism is implemented which allows to use logical expressions when defining platforms e.g. chrony and (ssh or ntp)

#### Rationale:

This enables greater flexibility when defining applicability of rules / groups.

### DIFF
--- a/build-scripts/collect_remediations.py
+++ b/build-scripts/collect_remediations.py
@@ -130,7 +130,6 @@ def main():
             # Happens on non-debug build when a platform is
             # "documentation-incomplete"
             continue
-        print (platform)
         cpe_platforms[platform.name] = platform
 
     for rule_file in os.listdir(args.resolved_rules_dir):
--- a/build-scripts/collect_remediations.py
+++ b/build-scripts/collect_remediations.py
@@ -45,6 +45,9 @@ def parse_args():
         "--fixes-from-templates-dir", required=True,
         help="directory from which we will collect fixes generated from "
         "templates")
+    p.add_argument(
+        "--platforms-dir", required=True,
+        help="directory from which we collect compiled platforms")
 
     return p.parse_args()
 
@@ -116,6 +119,16 @@ def main():
 
     product = ssg.utils.required_key(env_yaml, "product")
     output_dirs = prepare_output_dirs(args.output_dir, args.remediation_type)
+
+    for platform_file in os.listdir(args.platforms_dir):
+        platform_path = os.path.join(args.platforms_dir, platform_file)
+        try:
+            platform = ssg.build_yaml.Platform.from_yaml(platform_path, env_yaml)
+        except ssg.build_yaml.DocumentationNotComplete:
+            # Happens on non-debug build when a platform is
+            # "documentation-incomplete"
+            continue
+        env_yaml["product_cpes"].platforms[platform.name] = platform
 
     for rule_file in os.listdir(args.resolved_rules_dir):
         rule_path = os.path.join(args.resolved_rules_dir, rule_file)
--- a/build-scripts/collect_remediations.py
+++ b/build-scripts/collect_remediations.py
@@ -12,6 +12,7 @@ import ssg.jinja
 import ssg.environment
 import ssg.utils
 import ssg.xml
+from ssg.build_cpe import ProductCPEs
 
 
 def parse_args():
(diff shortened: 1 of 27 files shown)


build-scripts/collect_remediations.py
```diff
@@ -117,6 +120,16 @@ def main():
...
 
+    for platform_file in os.listdir(args.platforms_dir):
+        platform_path = os.path.join(args.platforms_dir, platform_file)
+        try:
+            platform = ssg.build_yaml.Platform.from_yaml(platform_path, env_yaml)
+        except ssg.build_yaml.DocumentationNotComplete:
+            # Happens on non-debug build when a platform is
+            # "documentation-incomplete"
+            continue
+        env_yaml["product_cpes"].platforms[platform.name] = platform
```
jan-cerny, line 131: I don't like that we add it to the env_yaml. It's a dictionary already containing multiple different types of data from various different sources. There is arbitrary data with no structure. We use it as universal storage. Adding CPE to it increases already large confusion. Instead, we should define the structure of the data and keep everything organized. Try separating it from env_yaml, at least on high level.

shared/macros-bash.jinja
```diff
@@ -1677,3 +1677,26 @@ fi
 {{% endif %}}
 
 {{% endmacro %}}
+
+{{%- macro bash_pkg_conditional(package) -%}}
```
jan-cerny, line 1681: Missing docs for all new Jinja macros.

ssg/build_yaml.py
```diff
@@ -1987,3 +1998,89 @@ def export_ocil_to_file(self, filename):
...
+        "bash_conditional",
+        "ansible_conditional"
+    ]
+
+    prefix = "cpe-lang"
+    ns = PREFIX_TO_NS[prefix]
+
+    @classmethod
+    def from_text(cls, expression, env_yaml):
+        test = env_yaml["product_cpes"].algebra.parse(
```
jan-cerny, line 2027: I wonder why the `Algebra` instance is a member of the `ProductCPEs` class. Is it an optimization to prevent creating many instances of the `Algebra` class? Otherwise it seems that you could create an instance of `Algebra` here, without having to access the instance of `ProductCPEs` class.

vojtapolasek:
@jan-cerny  said:
I don't like that we add it to the env_yaml. It's a dictionary already containing multiple different types of data from various different sources. There is arbitrary data with no structure. We use it as universal storage. Adding CPE to it increases already large confusion. Instead, we should define the structure of the data and keep everything organized. Try separating it from env_yaml, at least on high level.
End of quote.
I understand your point, however, I thought that the env_yaml is actually suitable for this case because it is passed into various functions anyway.
cpe_products is currently used in several functions - when loading groups, rules, etc. If I separate it, it will increase number of parameters passed around, but it will not have any additional value. I am not sure if in the end it will make things less complicated.
Can you please explain more what wil be the benefit of separating cpe_products from env_yaml?

jan-cerny:
@vojtapolasek easier tracking of the data flow, clear responsibilities, easier location of data usage, prevention of conflicts, less clutter when debugging, named variables

Also, recently, we have found that jinja macros are loaded thousands times during the build which is caused by their passing into env_yaml, and I would like to avoid that this situation would happen also with CPE definitions at some point in future. Once something is put into env_yaml it's hard to track where its used and difficult to extract it from there.

OTOH You're right that I don't know if there will be any problems because I haven't tried it.

vojtapolasek:
@jan-cerny  I have separated the product_cpes from the env_yaml and updated tests accordingly. Update of documentation will come soon.

ssg/build_cpe.py
```diff
@@ -72,7 +71,7 @@ def load_content_cpes(self):
                 continue
 
             # Get past "cpes" key, which was added for readability of the content
-            cpes_list = open_raw(dir_item_path)["cpes"]
+            cpes_list = open_and_macro_expand(dir_item_path, self.product_yaml)["cpes"]
```
jan-cerny, line 74: Only a small thing: I have noticed that all the compiled platforms in `build/rhel8/platforms/` have the `definition_location` key set to an empty string. Is it OK or do you think we should fill it in? It seems that removing it won't be easy because it's defined in the `XCCDFEntity` class which is a parent class of many other classes in build_yaml.py.

jan-cerny (approved):
Looks amazing! Great!

Question: Will the documentation be a part of this PR or of a different PR?

jan-cerny (changes requested):
We can see some tracebacks during the build in CI. Can you check it?

vojtapolasek:
I have to fix it, building of ocp4 content got broken by the refactoring.
The macro documentation will be part of this PR, the general documentation is here: https://github.com/ComplianceAsCode/content/pull/8094

vojtapolasek:
I have updated macro documentation as well.

evgenyz:
> We can see some tracebacks during the build in CI. Can you check it?

For the current state of the PR observable problems are all related to CI flakiness.

evgenyz (approved):

jan-cerny (approved):

jan-cerny:
KUDOS

PR #8044, opened 2022-01-11
Add trivial bash fixes

matejak:
Implement trivial fixes mainly related to
- quoting
- unused variables
- bash shebangs that cause problems if they are present in our source snippets - the build system may not preserve them on the first line of the bash remediation content.

### DIFF
--- a/linux_os/guide/services/sssd/sssd_certificate_verification/bash/shared.sh
+++ b/linux_os/guide/services/sssd/sssd_certificate_verification/bash/shared.sh
@@ -11,7 +11,7 @@ for f in /etc/sssd/sssd.conf /etc/sssd/conf.d/*.conf; do
 	if [ ! -e "$f" ]; then
 		continue
 	fi
-	cert=$( awk '/^\s*\[/{f=0} /^\s*\[sssd\]/{f=1} f{nu=gensub("^\\s*certificate_verification\\s*=\\s*ocsp_dgst\\s*=\\s*(\\w+).*","\\1",1); if($0!=nu){cert=nu}} END{print cert}' "$f" )
+	cert=$(awk '/^\s*\[/{f=0} /^\s*\[sssd\]/{f=1} f{nu=gensub("^\\s*certificate_verification\\s*=\\s*ocsp_dgst\\s*=\\s*(\\w+).*","\\1",1); if($0!=nu){cert=nu}} END{print cert}' "$f")
 	if [ -n "$cert" ] ; then
 		if [ "$cert" != $var_sssd_certificate_verification_digest_function ] ; then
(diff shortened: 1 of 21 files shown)


shared/macros-bash.jinja
```diff
@@ -260,7 +260,7 @@ do
 
                 # Split concrete_rule into head, perm, and tail sections using hash '#' delimiter
                 rule_head=$(cut -d '#' -f 1 <<< "$concrete_rule")
-                rule_perm=$(cut -d '#' -f 2 <<< "$concrete_rule")
+                # rule_perm=$(cut -d '#' -f 2 <<< "$concrete_rule")  # not needed in the remediation, but that's what is stored there
```
yuumasato, line 263: I think you can remove the unused `rule_perm` variable and update the comment in line 261.

linux_os/guide/services/sssd/sssd_certificate_verification/bash/shared.sh
```diff
@@ -22,9 +22,9 @@ done
 
 if ! $found ; then
 	SSSD_CONF="/etc/sssd/conf.d/certificate_verification.conf"
-	mkdir -p $( dirname $SSSD_CONF )
-	touch $SSSD_CONF
-	chown root:root $SSSD_CONF
-	chmod 600 $SSSD_CONF
-	echo -e "[sssd]\ncertificate_verification = ocsp_dgst = $var_sssd_certificate_verification_digest_function" >> $SSSD_CONF
+	mkdir -p "$( dirname "$SSSD_CONF" )"
```
yuumasato, line 25: Are there any guidance about spaces after `(` and before `)`?

linux_os/guide/services/sssd/sssd_run_as_sssd_user/bash/shared.sh
```diff
@@ -16,9 +16,9 @@ done
 
 if ! $found ; then
 	SSSD_CONF="/etc/sssd/conf.d/ospp.conf"
-	mkdir -p $( dirname $SSSD_CONF )
-	touch $SSSD_CONF
-	chown root:root $SSSD_CONF
-	chmod 600 $SSSD_CONF
-	echo -e "[sssd]\nuser = sssd" >> $SSSD_CONF
+	mkdir -p "$( dirname "$SSSD_CONF" )"
```
yuumasato, line 19: Same thing here about the spaces after and before `(` and `)`.

yuumasato (changes requested):
Thanks for these fixes.

linux_os/guide/services/sssd/sssd_certificate_verification/bash/shared.sh
matejak: none that I am aware of, so removing those.

PR #8049, opened 2022-01-12
Update accounts_password_pam_retry rule to use pwquality.conf file

marcusburghardt:
The retry parameter from pam_pwquality was recently introduced in
pwquality.conf. Previously, it was only possible to set it directly in
PAM files. The OVAL, remediation and test scenarios were updated to be
in alignment with this new option to set the retry parameter.


linux_os/guide/system/accounts/accounts-pam/password_quality/password_quality_pwquality/accounts_password_pam_retry/oval/shared.xml
```diff
@@ -47,13 +59,31 @@
...
 
-  <ind:textfilecontent54_state comment="lower bound of password_pam_retry" id="state_password_pam_retry_lower_bound" version="1">
+  <ind:textfilecontent54_state id="state_password_pam_retry_lower_bound" version="1"
+                               comment="lower bound of password_pam_retry">
     <ind:subexpression datatype="int" operation="greater than">0</ind:subexpression>
   </ind:textfilecontent54_state>
 
-  <external_variable comment="External variable for pam_pwquality retry" datatype="int" id="var_password_pam_retry" version="1" />
+  <external_variable datatype="int" id="var_password_pam_retry" version="1"
+                     comment="External variable for pam_pwquality retry"/>
```
yuumasato, line 74: Please, avoid doing changes like indentation and line breaks on lines not related with actual fixes and enhancements. This increases the amount of lines one has to review WRT to a change.

Next time separate the beautification changes in a separate commit.

yuumasato (approved):
This rule's remediation was affecting all the other PAM rules. Thank you for the fix.

PR #8056, opened 2022-01-12
Close tempfiles

matejak:
In reality, we are just after their filenames and without closing them, those files are hanging around open, which can lead to problems.


vojtapolasek (approved):
LGTM, thank you

PR #8059, opened 2022-01-13
Fix OL7 CIS URI

yuumasato:
#### Description:

- The previous URI was resulting in 'Error 1016: Origin DNS error'.

#### Rationale:

- The new URI follows the same pattern RHEL8 and SLE15 have.


yuumasato:
@freddieRv CC for awareness

Mab879 (approved):

PR #8061, opened 2022-01-13
Increase audit_ospp_general and audit_immutable_login_uids test coverage

mildas:
#### Description:
These audit rules have only pass test scenarios. We need to have also fail tests to check that they fail when it's expected

#### Rationale:
Increase test coverage for audit rules from https://bugzilla.redhat.com/show_bug.cgi?id=2000264

### DIFF
--- a/linux_os/guide/system/auditing/policy_rules/audit_immutable_login_uids/tests/file_missing.fail.sh
+++ b/linux_os/guide/system/auditing/policy_rules/audit_immutable_login_uids/tests/file_missing.fail.sh
@@ -1,0 +1,3 @@
+# packages = audit
+
+rm -f /etc/audit/rules.d/11-loginuid.rules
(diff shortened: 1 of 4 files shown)


mildas:
@vojtapolasek I have tried `audit_immutable_login_uids` in container and got the same error. It's caused by `augenrules`:
```
[root@af414766d37c /]# augenrules --load                                                                              
/usr/sbin/augenrules: No change                                                                                        
You must be root to run this program.                                                                                  
[root@af414766d37c /]# echo $?                                                                                                                                                                                                                
4                                                    
```
and I see that audit rules are `platform: machine`, so I'd say we should ignore SSGTS in container and test it only in VM.

vojtapolasek (approved):
LGTM. I tested your new tests on local VM and they are working as expected. Meanwhile I discovered an error within the audit_ospp_general rule, I will fix it in a different PR.

PR #8064, opened 2022-01-13
Fix remediation of display_login_attempts rule

marcusburghardt:
Fix the remediation to properly configure PAM. The remediations were also updated to use `authselect` where the tool is present.


yuumasato (approved):
LGTM

PR #8066, opened 2022-01-14
Fix ansible lint issues and SLE variable selector using explicit forbidden `default` value

ggbecker:
#### Description:

- Minor fixes to content so they pass tests and build properly.

@truzzon this [commit](https://github.com/ComplianceAsCode/content/commit/bc9a779ad85951c2d4eb5bc2ab08f1adeee44669) fixes an issue with the variable selector and doesn't change the end result. I am just considering if the intention is actually to use the `suse` selector for this [variable](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/ntp/var_multiple_time_servers.var#L16).

### DIFF
--- a/controls/cis_sle12.yml
+++ b/controls/cis_sle12.yml
@@ -501,7 +501,6 @@ controls:
     automated: yes
     rules:
     - chronyd_specify_remote_server
-    - var_multiple_time_servers=default
     - chronyd_run_as_chrony_user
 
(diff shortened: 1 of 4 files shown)


vojtapolasek (approved):
Looks good. The failing test is expected, the test is not run in SLE environment so the rule is reported as "not checked".

PR #8069, opened 2022-01-15
The audit package and auditd service are needed for FAU_GEN.1 SFR.

adelton:
#### Description:

The audit package and auditd service are needed for FAU_GEN.1 SFR.

#### Rationale:

-  In OSPP, the SFR dealing with audit data generation (which auditd.service from audit package configures in its ExecStartPost) is FAU_GEN.1.


vojtapolasek (approved):
LGTM

PR #8070, opened 2022-01-16
These auditd configurations affect the whole SFR, not just its specific parts.

adelton:
#### Description:

- These auditd configurations affect the whole SFR, not just its specific parts.

#### Rationale:

- The FAU_GEN.1.1.c lists selections but other part of the SFR are also related to these configurations.


vojtapolasek (approved):
LGTM

PR #8071, opened 2022-01-16
Clarify that log_format and name_format affects specifically information included in the audit records, not events for which audit records get generated.

adelton:
#### Description:

- Clarify that log_format and name_format affects specifically information included in the audit records, not events for which audit records get generated.

#### Rationale:

- Where it is clear which part of SFR is related to the configuration checked and set by the rule, it makes sense to be more specific.


vojtapolasek (approved):
LGTM

PR #8072, opened 2022-01-16
Ensuring immutable UIDs is related to the subject identity required by FAU_GEN.1.2, it does not affect for wihch events audit records will be generated.

adelton:
#### Description:

- Ensuring immutable UIDs is related to the subject identity required by FAU_GEN.1.2, it does not affect for wihch events audit records will be generated.

#### Rationale:

- Avoid confusion about the purpose of the configuration.


vojtapolasek (approved):
LGTM

PR #8075, opened 2022-01-18
Add rules in ocp4/CM-8(3)

swghosh:
#### Description:

- Added the following rules to OCP4 CM-8(3) control
    - Reject Unsigned Container Images by Default
    - RBAC Limit Cluster Admin to Nodes
- Changed CM-8(3) status to "Partial" as RBAC rule is not automated

#### Rationale:

- CM - 8(3) requires addition of rules so that the following rules can be added:
    - OCP: only cluster admin can add a node
    - OCP: validate container is signed and coming from 
- This is inherently met in RHCOS, hence not added:
    - RHCOS: GPG key checking for software

### DIFF
--- a/controls/nist_ocp4.yml
+++ b/controls/nist_ocp4.yml
@@ -5467,18 +5467,24 @@ controls:
   levels:
   - high
 - id: CM-8(3)
-  status: pending
+  status: partial
   notes: |-
-    Section a: A complete control response is planned. Engineering progress can be
-    tracked via:
+    Section a: Red Hat OpenShift allows for configuring the platform in such a way that only signed containers will be accepted for scheduling. [1] [2]
+    Red Hat signs the images provided for running the platform, as well as the base images that are usable for building other workloads.
 
-    https://issues.redhat.com/browse/CMP-280
+    Customers should be careful to enforce image signing on their own images.
 
-    Section b: A complete control response is planned. Engineering progress can be
-    tracked via:
+    In addition, Red Hat OpenShift allows limiting he container image registries from which normal users can import images by setting allowedRegistriesForImport [3].
 
-    https://issues.redhat.com/browse/CMP-280
-  rules: []
+    [1] https://docs.openshift.com/container-platform/4.7/security/container_security/security-container-signature.html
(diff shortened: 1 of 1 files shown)


ggbecker:
ok-to-test

swghosh:
@jhrozek May I ask you to PTAL? I have updated the control response text with specific details about the rules.
Also, lemme know if we can update the control status to something apart from `partial`, I am not sure about that.

jhrozek:
> 

Sure! So the control response is what goes into the "notes" attribute. We shouldn't touch the "description" attribute, that comes from the standard and should be left as it is in the standard. If you don't mind, we can team up on this and I can provide the control response.

I would write something like:
```
Section a:  Red Hat OpenShift allows for configuring the platform in such a way that only signed containers will be accepted for scheduling. [1] [2]
Red Hat signs the images provided for running the platform, as well as the base images that are usable for building other workloads.

Customers should be careful to enforce image signing on their own images.

 In addition, Red Hat OpenShift allows limiting he container image registries from which normal users can import images by setting  allowedRegistriesForImport[3]

    [1] https://docs.openshift.com/container-platform/4.7/security/container_security/security-container-signature.html
    [2] https://docs.openshift.com/container-platform/4.7/security/container_security/security-deploy.html#security-deploy-image-sources_security-deploy
    [3] https://docs.openshift.com/container-platform/4.7/openshift_images/image-configuration.html

Section b: Red Hat OpenShift doesn't provide an integrated solution out-of-the-box. However, the customer can deploy Red Hat Advanced Cluster Security, or a third party product which offer such functionality.
```

If you agree with the text above, feel free to incorporate it or alternatively I can provide a patch you can cherry-pick into your PR.

swghosh:
Thank you so much @jhrozek for providing the response text which I've now updated in the notes field (really helpful) of c9bd182. I am hoping that this works out, lest you can send me the patch for the changes which I'd be happy to cherry-pick onto this PR :)

jhrozek (approved):
LGTM

jhrozek:
The fedora rawhide container CI failure is unrelated. I'm going to try merging the PR manually so we can close it.

PR #8079, opened 2022-01-19
Update test kickstarts wrt password lifetime restrictions

matusmarhefka:
Renew passwords for users in `%post` sections of test kickstarts
so the entries about password change are added into /etc/shadow
and remediation of the rule accounts_password_set_max_life_existing
does not request a password change in our automated tests.


matusmarhefka:
https://github.com/ComplianceAsCode/content/pull/8081

PR #8082, opened 2022-01-19
Fix ansible accounts umask ansible remediations

ggbecker:
#### Description:

- Fix ansible accounts umask ansible remediations
  - It now correctly handles the case when there is no line present in the
file by using the lineinfile ansible module.

### DIFF
--- a/linux_os/guide/system/accounts/accounts-session/user_umask/accounts_umask_etc_bashrc/ansible/shared.yml
+++ b/linux_os/guide/system/accounts/accounts-session/user_umask/accounts_umask_etc_bashrc/ansible/shared.yml
@@ -5,8 +5,16 @@
 # disruption = low
 {{{ ansible_instantiate_variables("var_accounts_user_umask") }}}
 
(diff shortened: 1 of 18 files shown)


vojtapolasek:
Hello, nice job, thank you. However, using only lineinfile will not remediate the system if there are multiple occurences of umask. I believe that the Bash remediation handles it. I think that both "replace" and "lineinfile" needs to be used for this. Could you please add the functionality?

ggbecker:
> Hello, nice job, thank you. However, using only lineinfile will not remediate the system if there are multiple occurences of umask. I believe that the Bash remediation handles it. I think that both "replace" and "lineinfile" needs to be used for this. Could you please add the functionality?

I've put the replace command back and now the lineinfile task will only execute if nothing was replaced. I've also updated the metadata of some tests to remove the OSPP profile because the profile doesn't contain certain rules. It was probably some legacy profile.

vojtapolasek (approved):
Amazing, thank you for the fix.

PR #8083, opened 2022-01-19
Determine the platform name for CIS profile descriptions and rationale

rhmdnd:
We recently added a profile for the EKS CIS benchmark. Luckily, we were
able to re-use many existing rules from the OCP CIS benchmark, but when
you render the rules in an EKS deployment, they reference "OpenShift" in
various descriptions or rationale. This can be confusing to end users
who might not understand the linkage between rules and profiles in this
repository.

This commit defines a variable for rules that are shared between the EKS
and OCP CIS benchmark and displays it appropriately depending on the
product type.

#### Description:

#### Rationale:

### DIFF
--- a/applications/openshift/kubelet/kubelet_anonymous_auth/rule.yml
+++ b/applications/openshift/kubelet/kubelet_anonymous_auth/rule.yml
@@ -6,8 +6,10 @@ platform: {{{ product }}}-node
 
 {{%- if product == "eks" %}}
 {{%- set kubeletconf_path = "/etc/kubernetes/kubelet/kubelet-config.json" %}}
+{{%- set platform_noun = "EKS" %}}
 {{%- else %}}
 {{%- set kubeletconf_path = "/etc/kubernetes/kubelet.conf" %}}
+{{%- set platform_noun = "OpenShift" %}}
 {{%- endif %}}
 
 title: 'Disable Anonymous Authentication to the Kubelet'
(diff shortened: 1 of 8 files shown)


applications/openshift/kubelet/kubelet_anonymous_auth/rule.yml
```diff
@@ -6,8 +6,10 @@ platform: {{{ product }}}-node
 
 {{%- if product == "eks" %}}
 {{%- set kubeletconf_path = "/etc/kubernetes/kubelet/kubelet-config.json" %}}
+{{%- set platform_noun = "EKS" %}}
```
yuumasato, line 9: Instead of defining the `platform_noun` in every rule, it should be defined in the products's `product.yml` file.
There is already `full_name` defined  that could be leveraged, if that is too long maybe a new `short_name` variable would make sense?
https://github.com/ComplianceAsCode/content/blob/master/products/eks/product.yml#L2

marcusburghardt: Hi @rhmdnd , do you plan to continue working on this sooner or later?

By the way, `short_name` would make sense, IMO.

jan-cerny: @rhmdnd any updates?

rhmdnd:
Closing this for now since I don't have the cycles to clean it up in the near future. I'll reopen this PR if/when we can address the issues here.

PR #8084, opened 2022-01-20
Fix require_emergency_target_auth and require_singleuser_auth in RHEL9

ggbecker:
#### Description:
- Fix require_emergency_target_auth and require_singleuser_auth in RHEL9
  - These rules didn't take RHEL9 into Jinja if statements. Then it was
using code that is used for RHEL7 which is not compatible.

#### Rationale:

- Related to: https://bugzilla.redhat.com/show_bug.cgi?id=1999576


vojtapolasek (approved):
Nice catch, thank you.

Issue #8074, opened 2022-01-17
rhel8 ospp is missing auditd_audispd_syslog_plugin_activated


comps:
#### Description of problem:
The current git version is missing `auditd_audispd_syslog_plugin_activated` in `rhel8/ospp.profile` even though other remote-rsyslog-logging related rules are there. Please add this rule to the profile, it seems it was missed while reverting the remote logging disablement, thanks.

#### SCAP Security Guide Version:
becb55f28

matejak:
Correct, removal in #5158 was followed by revert in #6167 and related additions in #5181 and #6181, but `auditd_audispd_syslog_plugin_activated` didn't make it back.

marcusburghardt:
The `auditd_audispd_syslog_plugin_activated` rule is still not in `rhel8/ospp.profile`. Is it intentional?

Mab879:
> The `auditd_audispd_syslog_plugin_activated` rule is still not in `rhel8/ospp.profile`. Is it intentional?

Still true as of today as well.

PR #5181, opened 2020-02-10
rhel8: modify rule selections for OSPP and STIG to meet baselines

vojtapolasek:
#### Description:

Following rules dropped from OSPP profile for rhel8:

- package_libcap-ng-utils_installed

- package_audispd-plugins_installed

- package_gnutls-utils_installed

- package_nss-tools_installed

- auditd_audispd_syslog_plugin_activated

Added following rules to OSPP profile for rhel8:

- package_crypto-policies_installed

- package_openssh-server_installed

- package_openssh-clients_installed

- package_policycoreutils-python-utils_installed

Added following rules to STIG profile for rhel8:

- package_audispd-plugins_installed

- package_libcap-ng-utils_installed

- auditd_audispd_syslog_plugin_activated

#### Rationale:

These changes are made to meet OSPP and STIG agreed baselines.

PR #8086, opened 2022-01-20
Revert temporary removal of rule

matejak:
The rule got removed in https://github.com/ComplianceAsCode/content/pull/5181 but it was not re-introduced later.

Fixes: #8074


Mab879:
Can you please update the `tests/data/profile_stability/rhel8/ospp.profile` to reflect this change as well?

Mab879 (approved):

Issue #8017, opened 2021-12-23
Requesting checksums be provided alongside the release binaries


mjmst74:
#### Description of problem:
Would it be possible to start providing checksums, or GPG signatures, along with the binaries on the https://github.com/ComplianceAsCode/content/releases/ page?

yuumasato:
Hi, I believe it should be simple to start providing checksums, not sure about GPG signatures, would have to take a look into it.

lenox-joseph:
I suspect if you want signatures and you're targeting a Linux platform, it's probably better to wait on a packager to get to it.

marcusburghardt:
Checksums are already provided.

PR #8087, opened 2022-01-21
Generate release artifacts' checksums

yuumasato:
#### Description:

- Generate and publish checksums of the source tarball and the pre-build content zipfiles.
  - Note: CMake checksums with `sha512` algorithm were only added in CMake 3.10 and `CPACK_PACKAGE_CHECKSUM` was only added in CPack 3.7.
  - As these commands are not part of the normal content build process, I'm not bumping `cmake_minimum_required`. It should be enough to ensure that the systems doing the release have CMake > 3.10.
  - The image used for the releases is a Fedora 34 and it has CMake 3.20.

#### Rationale:

- Partially addresses #8017

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -60,7 +60,10 @@ jobs:
           body: ${{ steps.build_changelog.outputs.changelog }}
           files: |
             build/scap-security-guide-*.tar.bz2
+            build/scap-security-guide-*.tar.bz2.sha512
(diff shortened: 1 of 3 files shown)


Mab879 (approved):
Looks good.

```
$ sha512sum -c *.sha512
scap-security-guide-0.1.60-oval-5.10.zip: OK
scap-security-guide-0.1.60.tar.bz2: OK
scap-security-guide-0.1.60.zip: OK
```

PR #8088, opened 2022-01-21
Improve error handling in the test suite

matejak:
- `run_with_stdout_logging` returns a CompleteProcess instance instead of just stdout
- `run_with_stdout_logging` doesn't raise an exception if return code is non-zero
- `execute_ssh_command` accepts an error template that allows better user feedback

As a result, remediation errors are made visible for ssgts tests from the terminal output.

### DIFF
--- a/tests/ssg_test_suite/test_env.py
+++ b/tests/ssg_test_suite/test_env.py
@@ -116,19 +116,25 @@ class TestEnv(object):
     def get_ssh_additional_options(self):
         return list(common.SSH_ADDITIONAL_OPTS)
 
-    def execute_ssh_command(self, command, log_file, error_msg=None):
+    def execute_ssh_command(self, command, log_file, error_msg_template=None):
+        """
+        Args:
+            - command: Command to execute remotely as a single string
+            - log_file
+            - error_msg_template: A string that can contain references to:
+                ``command``, ``remote_dest``, ``rc``, and ``stderr``
+        """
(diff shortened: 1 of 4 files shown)


tests/ssg_test_suite/test_env.py
```diff
@@ -116,19 +116,26 @@ def get_ssh_port(self):
...
-                "ssh", tuple(self.ssh_additional_options) + (remote_dest, command), log_file)
-        except Exception as exc:
-            logging.error(error_msg + ": " + str(exc))
+        result = common.run_with_stdout_logging(
+            "ssh", tuple(self.ssh_additional_options) + (remote_dest, command), log_file)
+        if result.returncode:
+            error_msg = error_msg_template.format(
+                command=command, remote_dest=remote_dest,
+                rc=result.returncode, stderr=result.stderr)
+            logging.error(error_msg)
```
jan-cerny, line 136: I see every error twice. I think it's because of this line.

jan-cerny (changes requested):
Amazing idea. I tried to break the remediation for the accounts_tmout rule and thanks to this I could instantly see where the problem is.

jan-cerny (approved):
LGTM. An important information is that it works only with `--remediate-using bash`, not with the default (oscap based) remediation. That's kind of obvious, because the `oscap` command doesn't propagate the remediation errors, but during my review I was able to forget about that completely.

Issue #7752, opened 2021-10-15
RHEL9 failing rules per profile


matusmarhefka:
#### Description of problem:
See the section per each RHEL9 profile which contains rules failing even after remediation.

#### SCAP Security Guide Version:
master (57fc344)

#### Operating System Version:
RHEL9 Beta

#### Steps to Reproduce:
1. Run installation of RHEL9 Beta with a profile kickstart file (from https://github.com/ComplianceAsCode/content/tree/master/products/rhel9/kickstart)
2. Scan the system after installation with the profile based on kickstart which was used to install the system.

#### OSPP, CUI
`service_auditd_enabled` (in SSGTS profile mode)
`no_tmux_in_shells` (in SSGTS profile mode)

#### CIS Workstation Level 2 (GUI)
~`dconf_gnome_login_banner_text`~
~`selinux_confinement_of_daemons` - no remediation, but finds many breakages~

#### HIPAA
`rpm_verify_hashes` - detects issue in `systemd` package (`/usr/lib/systemd/system/rescue.service`)
~`auditd_audispd_syslog_plugin_activated`~

#### HIPAA (GUI)
`rpm_verify_hashes` - detects issue in `systemd` package (`/usr/lib/systemd/system/rescue.service`)
`rpm_verify_permissions` - detects issues in `accountsservice`, `gdm` and `fprintd` packages
~`auditd_audispd_syslog_plugin_activated`~

#### PCI-DSS
~`configure_opensc_card_drivers` - no Bash remediation~
~`force_opensc_card_drivers` - no Bash remediation~

#### PCI-DSS (GUI)
`rpm_verify_permissions` - detects issues in `accountsservice`, `gdm` and `fprintd` packages
~`auditd_audispd_syslog_plugin_activated`~

#### STIG
`no_tmux_in_shells` (SSGTS profile mode)
`configure_firewalld_ports`
`accounts_password_set_min_life_existing` (SSGTS profile mode)
`sshd_set_idle_timeout`

#### STIG_GUI (GUI)
`no_tmux_in_shells` (SSGTS profile mode)
`accounts_password_set_min_life_existing` (SSGTS profile mode)
`configure_firewalld_ports`
`sshd_set_idle_timeout`

#### ANSSI BP-028 High (results for both, no GUI and GUI)
~`accounts_polyinstantiated_tmp`~ - fails only after kickstart installation, needs to be remediated once more
[...]

PR #8100, opened 2022-01-25
Fix configure firewalld ports

yuumasato:
#### Description:

- Update the path where the check `firewalld_sshd_port_enabled` checks for NIC configuration.
- Fixes failures on rule `configure_firewalld_ports`. (The rule check is the same as `firewalld_sshd_port_enabled`'s check.

#### Rationale:

- The path changed in recent Fedoras and RHEL9:
  - https://fedoramagazine.org/nmstate-a-declarative-networking-config-tool/
  - https://access.redhat.com/solutions/5313011

- Related to https://github.com/ComplianceAsCode/content/issues/7752. Addresses fails on rule `configure_firewalld_ports`.


Mab879:
Tests pass locally.

PR #8103, opened 2022-01-25
OL8 STIG severity adjustments

freddieRv:
#### Description:

- Update the severity of several rules to be aligned with the OL8 and RHEL8 STIG profiles.
- For rules that have a different severity on OL8 and RHEL8 STIG profiles, adjust the severity on the OL8 stig profile.
- Remove the 'draft' status from the OL8 stig profile and update the rule's selection for the OL08-00-010287 requirement.

#### Rationale:

- Oracle Linux content efforts, alignment with DISA STIG.


linux_os/guide/system/software/disk_partitioning/partition_for_home/rule.yml
```diff
@@ -14,7 +14,7 @@ rationale: |-
     setting of more restrictive mount options, and also helps ensure that
     users cannot trivially fill partitions used for log or audit data storage.
 
-severity: low
+severity: medium
```
Mab879, line 17: So SUSE benchmarks mark this as a low.

linux_os/guide/system/software/disk_partitioning/partition_for_tmp/rule.yml
```diff
@@ -12,7 +12,7 @@ rationale: |-
     Placing <tt>/tmp</tt> in its own partition enables the setting of more
     restrictive mount options, which can help protect programs which use it.
 
-severity: low
+severity: medium
```
Mab879, line 15: RHEL7 has this as a low.

Mab879 (commented):
I want to hear from others about what we should do for when we have different severity per OS.

linux_os/guide/system/software/disk_partitioning/partition_for_tmp/rule.yml
freddieRv: We can always make the severity adjustments on the profiles, that is definitely a less disruptive change.

freddieRv: Done

Mab879 (approved):

PR #8104, opened 2022-01-26
OCP4:fix controller_insecure_port_disabled rule

Vincent056:
This rule was checking if insecure port only binds to the loopback address, since the --port flag is deprecated from kube controller manager, the insecure port is disabled, therefore we updated this rule to check if the insecure port is present, or does it bind to a loopback address

Related Bugzilla link: https://bugzilla.redhat.com/show_bug.cgi?id=2040132, 

Github commit:https://github.com/openshift/cluster-kube-controller-manager-operator/commit/fc8d3bb3b90a22854e9d678351512823da725673#diff-eb046ac1c42be018b202cbf3b94b1f08716c318b257aa4931784c48fe0174372

### DIFF
--- a/applications/openshift/controller/controller_insecure_port_disabled/rule.yml
+++ b/applications/openshift/controller/controller_insecure_port_disabled/rule.yml
@@ -16,6 +16,12 @@ description: |-
       "port": ["0"],
     ...
     </pre>
+    It is also acceptable for a system to deprecate the insecure port:
+    <pre>
+    "extendedArguments": {
+    ...
+    ...
+    </pre>
 
 rationale: |-
     The Controller Manager API service is used for health and metrics
@@ -24,6 +30,8 @@ rationale: |-
     attack surface.
 
 severity: low
+
+{{% set jqfilter = '[.data."config.yaml" | fromjson | if .extendedArguments["port"]!=null then .extendedArguments["port"]==["0"] else true end]' %}}
 
 identifiers:
     cce@ocp4: CCE-83578-5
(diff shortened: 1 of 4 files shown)


jhrozek (approved):
I'm acking the PR with a question -- given how you read the arguments of controller-manager, shouldn't we also check (in a follow-up PR) that `--secure-port` is not 0?

jhrozek:
cluster installation is FUBAR..
```
error running backup collection: errors occurred while gathering data:
    [skipping gathering namespaces/openshift-machine-api due to error: one or more errors occurred while gathering pod-specific data for namespace: openshift-machine-api

    [one or more errors occurred while gathering container data for pod cluster-autoscaler-operator-8c58cddb7-txsn9:

    [container "cluster-autoscaler-operator" in pod "cluster-autoscaler-operator-8c58cddb7-txsn9" is waiting to start: trying and failing to pull image, previous terminated container "cluster-autoscaler-operator" in pod "cluster-autoscaler-operator-8c58cddb7-txsn9" not found], one or more errors occurred while gathering container data for pod cluster-baremetal-operator-5cb64fbfb5-wgg5f:
```

jhrozek:
huh, keycloak setup seems to be timing out again. Let's see if it was a fluke, otherwise we need to investigate properly

/retest

Vincent056:
> 

We didn't merge this one yet: https://github.com/ComplianceAsCode/ocp4e2e/pull/20, But I think it might be with something else with cluster authentication operator

Vincent056:
> huh, keycloak setup seems to be timing out again. Let's see if it was a fluke, otherwise we need to investigate properly
> 
> /retest


We didn't merge this one yet: https://github.com/ComplianceAsCode/ocp4e2e/pull/20, But I think it might be with something else with cluster authentication operator

PR #8106, opened 2022-01-26
Kickstart file for ANSSI using authselect

marcusburghardt:
Since rhel8, it is recommended to use authselect profiles to make PAM settings more manageable and less error prone.

This makes possible to properly apply the remediations for required PAM rules.


yuumasato (approved):
LGTM, thanks!

PR #8107, opened 2022-01-26
Put fapolicyd group before accounts group in the benchmark.

ggbecker:
#### Description:
- Put fapolicyd group before accounts group in the benchmark.
  - The rule that installs fapolicyd creates a new user in the system and
this new user needs to be configured by rules from the accounts group.
Otherwise it caused the classical problem of ordering of rules in the
benchmark.

### DIFF
--- a/ssg/build_yaml.py
+++ b/ssg/build_yaml.py
@@ -1146,6 +1146,11 @@ class Group(XCCDFEntity):
             # any deviations or stricter settings are applied by the rules in the profile.
             "software", "integrity", "integrity-software", "rpm_verification",
 
+            # Installing fapolicyd package creates a new user called fapolicyd
+            # and password_expiration (subgroup of accounts) rules configure parameters for
+            # all users in the system. So it should come first then accounts group.
+            "fapolicyd",
+
             # The account group has to precede audit group because
             # the rule package_screen_installed is desired to be executed before the rule
             # audit_rules_privileged_commands, othervise the rule


ssg/build_yaml.py
```diff
@@ -1146,6 +1146,11 @@ def to_xml_element(self, env_yaml=None):
             # any deviations or stricter settings are applied by the rules in the profile.
             "software", "integrity", "integrity-software", "rpm_verification",
 
+            # Installing fapolicyd package creates a new user called fapolicyd
+            # and password_expiration (subgroup of accounts) rules configure parameters for
+            # all users in the system. So it should come first then accounts group.
+            "fapolicyd",
```
yuumasato, line 1152: The `fapolicyd` package is still installed prior to configuring password remediation.
This happens because they are in different branches, :( 

```
./linux_os/guide/system/accounts/accounts-restrictions/password_expiration/accounts_password_set_max_life_existing
./linux_os/guide/services/fapolicyd/package_fapolicyd_installed
```

System and Service groups would have to be swapped.

yuumasato: And indeed swapping the groups fixed this particular issue.

yuumasato: "historically" the `system` group was placed before the  `services` group, and we kept it that way:
https://github.com/ComplianceAsCode/content/pull/4900

I'm not sure what other consequences swapping `system` and `services` would do...

ggbecker:
Closing this as it doesn't fix the problem. We have to find a better way to get `package_fapolicyd_installed` rule to become before rules in accounts group. The way it is, every rule from group `system` comes before rules from group `services`.

PR #8108, opened 2022-01-26
Fix ansible remediation for configure_gnutls_tls_crypto_policy

vojtapolasek:
#### Description:

- make tests applicable for rhel9 as well
- rewrite the remediation

#### Rationale:

The remediation did not apply changes if there was no line which would match regexp.

### DIFF
--- a/linux_os/guide/system/software/integrity/crypto/configure_gnutls_tls_crypto_policy/ansible/shared.yml
+++ b/linux_os/guide/system/software/integrity/crypto/configure_gnutls_tls_crypto_policy/ansible/shared.yml
@@ -15,7 +15,7 @@
     follow: yes
   register: gnutls_file
 
-- name: "{{{ rule_title }}}: Add"
+- name: "{{{ rule_title }}}: Create file and add line"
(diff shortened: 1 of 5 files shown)


vojtapolasek:
I believe the failing test is due to problem with the container. Also cryptopolicies are not part of SLE AFAIK.

yuumasato (changes requested):
> The remediation did not apply changes if there was no line which would match regexp.

The reason the regex does not match is because the syntax of `/etc/crypto-policies/back-ends/gnutls.config` changed significantly in RHEL9, it has the format an ini file now.

The rule should probably be re-evaluated, whether it makes sense for RHEL9, and if the customization to the gnutls TLS crypto policy still make sense.

yuumasato:
I see that this rule is only selected on STIG profiles.
CC @ggbecker and @Mab879 for awareness.

vojtapolasek:
The format of gnutls configuration for rhel9 has changed, it needs a new rule probably. It will be addressed in a different PR.

PR #8109, opened 2022-01-26
sshd_lineinfile ansible macro dir support and directory check fix

mildas:
#### Description:
Add support for directory check in sshd macro:

- `config_file="/etc/ssh/sshd_config"` - remove configuration from this file
- `config_dir="/etc/ssh/sshd_config.d"` - remove configuration from this directory
- `set_file="/etc/ssh/sshd_config.d/00-complianceascode-hardening.conf"` - configure in this file - we use the same file in Bash remediation

Fix for `ansible_set_config_file_dir`:

- it did not remove configuration line from `config_file` because of `dupes.found is defined and dupes.found > 1` condition was resulting in false

### DIFF
--- a/shared/macros-ansible.jinja
+++ b/shared/macros-ansible.jinja
@@ -174,8 +174,12 @@ value: :code:`Setting={{ varname1 }}`
   We also specify the validation program here; -t specifies test and -f allows
   Ansible to pass a file at a different path.
 #}}
-{{%- macro ansible_sshd_set(msg='', parameter='', value='') %}}
+{{%- macro ansible_sshd_set(msg='', parameter='', value='', config_is_distributed="false") %}}
+{{%- if config_is_distributed == "true" %}}
+{{{ ansible_set_config_file_dir(msg, config_file="/etc/ssh/sshd_config", config_dir="/etc/ssh/sshd_config.d", set_file="/etc/ssh/sshd_config.d/00-complianceascode-hardening.conf", parameter=parameter, separator_regex="\s+", value=value, prefix_regex="^\s*", create='yes', validate='/usr/sbin/sshd -t -f %s', insert_after='', insert_before="^[#\s]*Match") }}}
+{{%- else %}}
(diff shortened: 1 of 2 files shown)


mildas:
Without forcing the `ansible_set_config_file_dir` to remove line from `/etc/ssh/sshd_config`, the remediation did:
```
# grep -r HostbasedAuthen
sshd_config:HostbasedAuthentication yes
```
to
```
# grep -r HostbasedAuthen
sshd_config.d/00-complianceascode-hardening.conf:HostbasedAuthentication no
sshd_config:HostbasedAuthentication yes
```
And I don't think this is correct behavior.
But maybe I'm wrong and `sshd_config.d/` configurations are prioritized over `sshd_config` and we are expected to not touch it. Thoughts? @yuumasato @ggbecker

yuumasato:
> Without forcing the `ansible_set_config_file_dir` to remove line from `/etc/ssh/sshd_config`, the remediation did:
> 
> ```
> # grep -r HostbasedAuthen
> sshd_config:HostbasedAuthentication yes
> ```
> 
> to
> 
> ```
> # grep -r HostbasedAuthen
> sshd_config.d/00-complianceascode-hardening.conf:HostbasedAuthentication no
> sshd_config:HostbasedAuthentication yes
> ```
> 
> And I don't think this is correct behavior. But maybe I'm wrong and `sshd_config.d/` configurations are prioritized over `sshd_config` and we are expected to not touch it. Thoughts? @yuumasato @ggbecker

Maybe @matejak know whether the configurations in .d directory take precedence or not?

matejak:
> Maybe @matejak know whether the configurations in .d directory take precedence or not?

There is no implied precedence, drop-ins are effective only because the main config file includes them.
However, we have a rule that checks for that: https://github.com/ComplianceAsCode/content/tree/master/linux_os/guide/services/ssh/ssh_server/sshd_use_directory_configuration

That rule isn't very intelligent when it looks at that config file, so if the directive is there, it passes. There may be other config directives there though. OVALs of other sshd rules check the main config file and those drop-ins, so if there is any config contradiction caused by remediation not doing its job properly, the contradiction will be detected. And such situation won't arise if we talk about default settings or systems remediated from default setting.

yuumasato (approved):
Good job on updating the macro!

The Ansible remediations for SSH can now configure the drop in files.
And the other rules that used the macro `ansible_set_config_file_dir` are still working as expected.

PR #8118, opened 2022-01-27
Use var ansible_pkg_mgr to determine GPG checking

Ripolin:
#### Description:

Use package manager from Ansible facts to determine which task should be run for GPG checking

#### Rationale:

* GPG check will work on all EL distribution using yum or dnf, including RockyOS


Ripolin:
@ggbecker Ok I close this PR and try the latest release. Thx for your quick response

PR #8121, opened 2022-01-27
Removed criteria in OVAL check of require_singleuser_auth

Xeicker:
#### Description:

- Removed unnecessary criteria in OVAL check of **require_singleuser_auth** rule

#### Rationale:

- DISA, in its stig profile for OL08-00-010151 and RHEL-08-010151, only requires the presence of:
`ExecStart=-/usr/lib/systemd/systemd-sulogin-shell rescue `
in the `/usr/lib/systemd/system/rescue.service` file. The other criteria are not mentioned.


vojtapolasek (approved):
LGTM

PR #8125, opened 2022-01-28
Add place for site specific information

jcpunk:
#### Description:

Add a location to insert site specific information in tailoring files

#### Rationale:

For my site I'd like to add a link to our site computing policy for sys admins, admin training guides for PII, and system registration process.  There isn't a great place to put that information in a tailored profile today.

In my local tests I couldn't seem to get this to work, so I fear I'm doing something wrongly....

### DIFF
--- a/applications/intro/site-local/group.yml
+++ b/applications/intro/site-local/group.yml
@@ -1,0 +1,6 @@
+documentation_complete: true
+
+title: "Site Specific Information"
+
+description: |-
+  {{{ xccdf_value("site_local_message") }}}
--- a/applications/intro/site-local/group.yml
+++ b/applications/intro/site-local/group.yml
@@ -3,4 +3,4 @@ documentation_complete: true
 title: "Site Specific Information"
 
 description: |-
-  {{{ xccdf_value("site_local_message") }}}
+  {{{ xccdf_value("var_site_local_message") }}}
--- a/applications/intro/site-local/var_site_local_message.var
+++ b/applications/intro/site-local/var_site_local_message.var
@@ -1,0 +1,15 @@
+documentation_complete: true
+
+title: 'Site Local Message'
+
+description: |-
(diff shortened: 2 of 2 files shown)


applications/intro/site-local/group.yml
```diff
@@ -0,0 +1,6 @@
+documentation_complete: true
+
+title: "Site Specific Information"
+
+description: |-
+  {{{ xccdf_value("site_local_message") }}}
```
matejak, line 6:
```suggestion
  {{{ xccdf_value("var_site_local_message") }}}
```

This could help - the `.var` suffix is not part of the variable name, but the `var_` prefix is, as it has no special meaning.

matejak:
It is not clear what is the purpose of this PR. It looks like that you would like to add a rule that is parametrized by the XCCDF value, but the rule has no serious content. The build error could be related to the variable name and reference mismatch.

jcpunk:
My intent with this PR is to provide a clear place where I can add some site local content to the tailoring.

Specifically, I'm looking to add a link to the admin training guides for PII for our site.  This isn't a rule that can be audited against the system, so much as it is a way to interlink our policies.  Folks running the tailored rule set should be aware of this information, but reminders for humans and links for new staff can be very helpful.

jan-cerny:
Hi @jcpunk , you are adding this to the `intro` group. However, the contents of this group aren't visible in the HTML guide. In fact, we would like to get rid of intro group from our project, because the information isn't visible anywhere.

Instead, you could modify the contents of the `description` element in your tailoring file. The contents of the `description` would be visible in HTML guide and would work with any version of content. What do you think about it?

jan-cerny:
@jcpunk Did that work out? Can we help? Do you still want this PR?

jcpunk:
Tailoring the description seems to be sufficient for me.

PR #8126, opened 2022-01-28
Create krb5 related CPEs and applied to a rule for ol8 and rhel8

Xeicker:
#### Description:

- Introduce krb5_server_older_than_1_17-18 & krb5_workstation_older_than_1_17-18 applicability CPEs
- Apply both of them to kerberos_disable_no_keytab rule for ol8 and rhel8

#### Rationale:

- DISA, in its STIG profile, requires the rule kerberos_disable_no_keytab
to only be applicable if the system has a version of krb5-server or
krb5-workstation older than 1.17-18


jan-cerny:
@Mab879 What's our stance on this for RHEL 8?

Mab879:
> @Mab879 What's our stance on this for RHEL 8?

From the RHEL8 STIG V1R5 on STIG ID RHEL-08-010161

> If the system is a workstation utilizing krb5-workstation-1.17-18.el8.x86_64 or newer, this requirement is not applicable.

Seems acceptable to me.

Mab879 (changes requested):
Just a couple minor issues. But overall looks good.

jan-cerny (approved):

PR #8127, opened 2022-01-28
OL8 missing parameter pass

freddieRv:
#### Description:

- Add a `missing_parameter_pass@ol8` entry set to `false` to several rules to align them with DISA STIG requirements which have statements along the lines "_if the parameter is missing, or if the returned line is commented out, this is a finding_"
- The affected DISA requirements are:
  - OL08-00-040340
  - OL08-00-010500
  - OL08-00-010521
  - OL08-00-010522
  - OL08-00-010830
  - OL08-00-020350
  - OL08-00-030061

#### Rationale:

- Effort to align the OL8 "stig" profile with the OL8 DISA STIG one


ggbecker:
I believe the direction is to go this kind of changes project-wide. There were some resistance in the past because you generate a lot of new failures for programs that have the default config already compliant. Meanwhile, default values can change and might break this compliance, on one hand we should trust developers that their default is safe, on the other hand having them explicit will mean that a change will only occur if a change in the policy is actually made and not something that went under the radar as some default unset value changed. I struggle myself in taking sides, but I believe we should strive for the explicit config. Maybe you can propose these changes on a project level instead of product specific.

freddieRv:
@ggbecker just updated my changes to apply to all products.

I hope one commit for all files is also alright instead of one commit per file.

ggbecker:
We have transitioned to missing_parameter_pass: false (it's also the default value) in  #8179

The only rule that would still make sense in this pull request is:

[auditd_local_events/rule.yml](https://github.com/ComplianceAsCode/content/pull/8127/files#diff-c15beaa2f3a4758666a5aef6c3df5c382d5c51c73ce041c9165a55f9a6000efa)

freddieRv:
@ggbecker updated as suggested. Now only auditd_local_events is being updated.

marcusburghardt (approved):

marcusburghardt (changes requested):
The `commented.pass.sh`, the `missing_value.pass.sh` and the `missing_file.pass.sh` also need to be adjusted in alignment to this change.

marcusburghardt (approved):

PR #8130, opened 2022-01-31
Assign RHEL-08-010341 and RHEL-08-010351 to rules

ggbecker:
#### Description:

- Assign RHEL-08-010341 and RHEL-08-010351 to rules.
  - Convert rules to use template. Enhance file templates to disregard symlinks.


ggbecker:
The failing SSGTS test is legit, I'm working on it.

ggbecker:
> The failing SSGTS test is legit, I'm working on it.

The problem should be solved by f82c314

jan-cerny (approved):

PR #8134, opened 2022-01-31
OCP4 adds OVN,SDN networkType CPE

Vincent056:
We want to add OVN and SDN network type CPE so that OVN/SDN rules can be run on the correct platform accordingly

This PR is also to address: https://bugzilla.redhat.com/show_bug.cgi?id=1994609


JAORMX (approved):

jhrozek (approved):
/lgtm

jhrozek:
In tests, I see one cluster-up failure and the known IDP failure. Since this PR is about adding a new CPE, I think it's safe to merge.

PR #8136, opened 2022-02-01
Add github workflow to publish statistics, guides and tables

ggbecker:
#### Description:

- Add github workflow to publish statistics, guides and tables.

Job ported from https://jenkins.complianceascode.io/job/scap-security-guide-stats/

Preview: https://ggbecker.github.io/content/

Job should only start when the change is merged to the main branch.

### DIFF
--- a/.github/workflows/gh-pages.yaml
+++ b/.github/workflows/gh-pages.yaml
@@ -5,14 +5,14 @@ on:
 jobs:
   publish:
     name: Publish stats, tables and guides
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
+    container:
+      image: fedora:latest
     env:
       PAGES_DIR: __pages
     steps:
       - name: Install Deps
-        uses: mstksg/get-package@master
-        with:
-          apt-get: cmake ninja-build libopenscap8 libxml2-utils expat xsltproc python3-jinja2 python3-yaml python3-setuptools ansible-lint python3-github bats python3-pytest python3-pytest-cov python3-pip
+        run: dnf install -y cmake git ninja-build openscap-utils python3-pyyaml python3-jinja2 python3-pytest ansible-lint expat libxslt python3-pip
       - name: Install deps python
         run: pip3 install json2html
       - name: Checkout
--- a/.github/workflows/gh-pages.yaml
+++ b/.github/workflows/gh-pages.yaml
@@ -1,0 +1,36 @@
+name: Github Pages
+on:
+  push:
+    branches: [ 'master' ]
+jobs:
+  publish:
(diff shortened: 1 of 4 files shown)


.github/workflows/gh-pages.yaml
```diff
@@ -0,0 +1,122 @@
...
+        working-directory: ./build
+      - name: Build Guides and Mapping Tables
+        run: ninja -j2
+        working-directory: ./build
+      - name: Build Statistics
+        run: ninja html-stats html-profile-stats -j2
+        working-directory: ./build
+      - name: Generate HTML pages
+        run: |-
+          mkdir $PAGES_DIR
```
evgenyz, line 31: No way! Please create a script in the project!

evgenyz (changes requested):

.github/workflows/gh-pages.yaml
ggbecker: Fixed by rebase. I didn't spend much time on creating the script as it was a straight copy from the jenkins job. I particularly don't like to host this script in the project as it's not very useful for others and I'm afraid it would take too much time to turn into something reusable.

evgenyz: On the contrary, it is quite useful! First of all it can be tested and debugged without GH Actions involvement. And then, as the consequence, it could be used by some other CI solution if desired (like GitLab). By keeping the actions code to the minimum we make the project less dependent on GH.

Also if, even in this form, it is good enough to provide some value for project developers why we should be hesitant about adding it to the project?

evgenyz (approved):
I like it! Thank you!

jan-cerny:
@ggbecker What about reusing code from https://github.com/OpenSCAP/website/blob/master/utils/choosing_policy_page.py to make the Guides index.html nicer?

ggbecker:
> @ggbecker What about reusing code from https://github.com/OpenSCAP/website/blob/master/utils/choosing_policy_page.py to make the Guides index.html nicer?

That's probably a good idea. Let me take a look.

ggbecker:
> @ggbecker What about reusing code from https://github.com/OpenSCAP/website/blob/master/utils/choosing_policy_page.py to make the Guides index.html nicer?

Done in: 03514ecb86650197b75def8d30015b69ccaa7eec

ggbecker:
I've made one last improvement here: base the workflow on Fedora image as it contains the latest OpenSCAP scanner that is able to build OCP4 content, otherwise it wouldn't build the content and guides for this product (including EKS content)

556fbbf3fb4abcaddc9aedf2cd9466285ca05e7a

evgenyz (approved):

PR #8138, opened 2022-02-02
Update RHEL-08-010383 to require only one occurrence of a config.

ggbecker:
#### Description:

- Update RHEL-08-010383 to require only one occurrence of a config.


jan-cerny:
Great, but the commit message doesn't explain why.

ggbecker:
> Great, but the commit message doesn't explain why.

That's true. Actually I would like to expand more on this as it doesn't consider configuration files in the /etc/sudoers.d/ directory as it should. I'm moving the PR to draft for now.

ggbecker:
> 

The commit message is now updated and the remediations were updated to remove duplicates from configuration files.

yuumasato (approved):
LGTM!

PR #8141, opened 2022-02-02
Bug 1994609: OCP: Update SDN rules to the correct platform

Vincent056:
After we added OVN,SDN CPE from: https://github.com/ComplianceAsCode/content/pull/8134, and https://github.com/openshift/compliance-operator/pull/785 we will be able to addrees the issue where SDN rules was run on OVN cluster. This fix ensure the SDN rules will only run on SDN openshift clusters

Relate BUGZILLA Link: https://bugzilla.redhat.com/show_bug.cgi?id=1994609


Vincent056:
The CI will pass after: https://github.com/ComplianceAsCode/content/pull/8134 and https://github.com/openshift/compliance-operator/pull/785 merged.

jhrozek:
The code looks good, but we can't merge it until we release a new CO version which fetches the network resource right?

Vincent056:
> The code looks good, but we can't merge it until we release a new CO version which fetches the network resource right?

That's  right.

Vincent056:
@jhrozek the CI actually passed, looked at Compliance Result CM, for these platformscan, the results are `not applicable`, it turns out the CI didn't care about `not applicable` compliance check result, which makes sense since all the node rules are `not applicable` on a platform scan result. For node scan, since it was just checking for a file on the node to identify if it is on SDN cluster, these rules resulted in a pass.

```
          <rule-result idref="xccdf_org.ssgproject.content_rule_file_owner_proxy_kubeconfig" role="full" time="2022-02-10T23:21:10+00:00" severity="medium" weight="1.000000">
            <result>notapplicable</result>
          </rule-result>
```

jhrozek (approved):
Sorry for the delay in review
/lgtm

PR #8142, opened 2022-02-02
Update EKS CIS section 2

rhmdnd:
Section 2 of the EKS CIS profile can only be checked manually since it
requires using input and a network connection to an AWS endpoint.
Updating the comment and status of that section to reflect that status.

#### Description:

#### Rationale:


jhrozek (approved):
/lgtm

PR #8149, opened 2022-02-04
Evaluate CIS EKS benchmark section 5.2

rhmdnd:
This commit updates section 5.2 of the benchmark to manual since it
requires access to the AWS IAM endpoint and it requires
deployment-specific knowledge to perform the audit.

#### Description:

#### Rationale:

### DIFF
--- a/controls/cis_eks.yml
+++ b/controls/cis_eks.yml
@@ -707,14 +707,37 @@ controls:
   - id: '5.2'
     title: >-
       5.2 Identity and Access Management (IAM)
-    status: pending
+    status: manual
     levels:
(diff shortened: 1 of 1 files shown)


Vincent056 (commented):
I wonder if it is possible to collect aws endpoint resources in CO, to do that, user probably have to supply IAM service account token, I think we could add a option in the helm deployment to create such serect with user provided aws service token to use it in CO, and so we are able to gather such resources in CO during a scan.

jhrozek (approved):
LGTM

Vincent056 (approved):

PR #8150, opened 2022-02-04
Evaluate CIS EKS benchmark section 5.3

rhmdnd:
This commit updates section 5.3 to be a manual check since it requires
deployment-specific knowledge from the user and a network connection to
an AWS endpoint. It also corrects the level to be accurate with what's
documented in the benchmark.

#### Description:

#### Rationale:


jhrozek (approved):
LGTM

PR #8151, opened 2022-02-04
Add missing rsync dependency to gh-pages workflow

ggbecker:
#### Description:

- Add missing rsync dependency to gh-pages workflow.

#### Rationale:

- Without this package the github action is not able to publish the
generated artifacts.

The page preview led to a confusion because what was being presented there was something previously built.


ggbecker:
It builds correctly now: https://github.com/ggbecker/content/runs/5069129669?check_suite_focus=true

evgenyz (approved):

PR #8152, opened 2022-02-04
align ospp audit rules with the latest upstream release

vojtapolasek:
#### Description:

- unify the file content for rhel9 and rhel8 products
- remove extra test scenarios and test data for rhel9
- align files with sample-rules shipped in Audit 3.0.7

#### Rationale:

Both rhel8 and rhel9 content of audit rules related to ospp got updated.


jan-cerny:
@mildas @vojtapolasek @ggbecker  The test scenarios are all notapplicable, that's a similar problem we hit in #8146.

ggbecker:
> @mildas @vojtapolasek @ggbecker The test scenarios are all notapplicable, that's a similar problem we hit in #8146.

Found the problem, newer versions of the build system is generating cpe name for machine as simply: `#machine` whereas previously it generated `#cpe_platform_machine` thus the `test_suite.py` needs to be updated accordingly here:

https://github.com/ComplianceAsCode/content/blob/0d7ac62a4e1c85fa0c9b2b365ace5d2b7aa61f16/tests/ssg_test_suite/xml_operations.py#L102-L103

I'm going to submit a PR for that, but I wonder if any other place should be updated as well.

vojtapolasek:
I will rebase to reflect the changes in the test suite.

vojtapolasek:
The tests are now updated to be aligned with Audit 3.0.7 upstream version. I think these are final changes.

yuumasato (approved):

PR #8153, opened 2022-02-04
Evaluate EKS CIS section 5.4

rhmdnd:
This commit updates these checks to manual since they all require either
deployment-specific knowledge or API access to AWS or Kubernetes since
the api server configuration isn't available on the worker nodes.

#### Description:

#### Rationale:


jhrozek (approved):
LGTM

PR #8160, opened 2022-02-04
Create OVAL check for selinux_context_elevation_for_sudo [OL7]

Xeicker:
#### Description:

- Implement first version of OVAL check for the rule selinux_context_elevation_for_sudo. 

#### Rationale:

- This rule didn't have automated content, Done to meet DISA STIG OL07-00-020023  requirement


jan-cerny:
SSGTS fail is expected because it didn't find the rule because this rule in only for OL7.

ggbecker:
We should probably consider including this rule into RHEL7 STIG 

STIG id: RHEL-07-020023

PR #8161, opened 2022-02-04
Oval update for two rules to only allow results from only one file [ol7]

Xeicker:
#### Description:

- Update OVAL of sudoers_validate_passwd & sudo_require_reauthentication to only allow that the required configuration is located only in one file

#### Rationale:

- This is introduced in latest DISA STIG profile v2r6. Both OL07-00-010342 & OL07-00-010343 implemented in SSG in sudoers_validate_passwd & sudo_require_reauthentication respectevely added the statement "If results are returned from more than one file location, this is a finding"


ggbecker (changes requested):
We implemented this kind of change by changing the check_existence to "only_one_exists" which will have a very similar result. It wouldn't allow duplicates of the same config within the same file though. So the check only allows one occurrence of such configuration which basically ends up being one file only.

There is the case where multiple configurations are checked within one rule, so the check_existence alone is not enough, because if you have those configurations distributed among different files, it would still pass. So this extra check is needed.

I believe you can generalize this check to all products and add a test scenario for  `sudoers_validate_passwd` that would create a valid configuration among more than one file and this check should be able to capture that.

The same comment here applies for other PRs you are proposing that are similar to this one. Thanks.

ggbecker (approved):

PR #8163, opened 2022-02-04
Fix the directory for products stats script

Mab879:
#### Description:
Fixes the `utils/generate_html_pages.sh` so that per product statistics work.
#### Rationale:
Current the statistics folder is empty on GH pages.

### DIFF
--- a/utils/generate_html_pages.sh
+++ b/utils/generate_html_pages.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # suppress stdout from pushd and popd commands
 pushd () {
     command pushd "$@" > /dev/null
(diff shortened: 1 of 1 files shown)


utils/generate_html_pages.sh
```diff
@@ -31,7 +31,7 @@ products=$(echo -e "import ssg.constants\nprint(ssg.constants.product_directorie
 for product in $products
 do
     if [ -d build/$product ]; then
-        mkdir -p $STATS_DIR/$product
+        mkdir -p $STATS_DIR/$product/product-statistics
```
ggbecker, line 34: this shouldn't be really a problem. The problem is with the `python` in the line `30`. We need to use `python3` instead.

https://github.com/ComplianceAsCode/content/runs/5072085920?check_suite_focus=true
`Generate HTML pages`
```
Run utils/generate_html_pages.sh $PAGES_DIR
utils/generate_html_pages.sh: line 30: python: command not found
```

Mab879: Thanks for the review. Bet I believe both are needed. Without this `mkdir` I get the following errors. But with the `mkdir` I don't get the errors.

```$ ./utils/generate_html_pages.sh stats4
cp: cannot create directory 'stats4/statistics/chromium/product-statistics': No such file or directory
cp: cannot create directory 'stats4/statistics/chromium/profile-statistics': No such file or directory
cp: cannot create directory 'stats4/statistics/debian9/product-statistics': No such file or directory
cp: cannot create directory 'stats4/statistics/debian9/profile-statistics': No such file or directory```

ggbecker: Let's add this and see what happens :)

ggbecker (approved):

PR #8164, opened 2022-02-04
Introduce nonzero_check var to accounts_password template

Xeicker:
#### Description:

- Added nonzero_check variable with default value set to False to the accounts_password template. And implement conditionally this behavior in which OVAL checks if the value obtained is different to 0.This could be useful to future rules since zero is a special value in a few related configurations
- This new variable has been implemented in the accounts_password_pam_maxclassrepeat rule

#### Rationale:

- For OL7 the STIG id OL07-00-010190 needs to check both if maxclassrepeat is less than 4 and different to 0. For this reason included the variable nonzero_check with default value False this way the rule keeps using the template accounts_password without modifying other rules behaviour. 
- Since DISA's requirements go in this direction(also RHEL's new profile states the same) the variable has been updated to be True as default value in accounts_password_pam_maxclassrepeat rule

### DIFF
--- a/docs/templates/template_reference.md
+++ b/docs/templates/template_reference.md
@@ -12,7 +12,11 @@
 
     -   **operation** - OVAL operation, eg. `less than or equal`
 
--   Languages: OVAL
+    -   **non_zerocheck** - effective only in OVAL checks, if
+        set to `"true"` it will test if the value is different
+        than zero. (default value: `"false"`).
+
+-   Languages: Ansible, Bash, OVAL
 
 #### auditd_lineinfile
 -   Checks configuration options of the Audit Daemon in
--- a/docs/templates/template_reference.md
+++ b/docs/templates/template_reference.md
@@ -2,19 +2,22 @@
 
 #### accounts_password
 -   Checks if PAM enforces password quality requirements. Checks the
-    configuration in `/etc/pam.d/system-auth` (for RHEL 6 systems) or
-    `/etc/security/pwquality.conf` (on other systems).
+    configuration in `/etc/security/pwquality.conf`.
 
 -   Parameters:
 
-    -   **variable** - PAM `pam_cracklib` (on RHEL 6) or `pam_pwquality`
-        (on other systems) module name, eg. `ucredit`, `ocredit`
+    -   **variable** - PAM `pam_pwquality` password quality
+        requirement, eg. `ucredit`, `ocredit`
 
     -   **operation** - OVAL operation, eg. `less than or equal`
 
-    -   **non_zerocheck** - effective only in OVAL checks, if
-        set to `"true"` it will test if the value is different
-        than zero. (default value: `"false"`).
+    -   **check_zero_boundary** - effective only in OVAL checks, if
+        set to `"true"` it will test if the value respects the a zero
+        boundary operation. The operation is defined by the
+        **zero_boundary_operation** parameter (default value: `"false"`)
+
+    -   **zero_boundary_operation** - OVAL operation, eg. `greater than`
+        (default value: `"greater than"`)
 
 -   Languages: Ansible, Bash, OVAL
 
--- a/docs/templates/template_reference.md
+++ b/docs/templates/template_reference.md
@@ -11,13 +11,9 @@
 
     -   **operation** - OVAL operation, eg. `less than or equal`
 
-    -   **check_zero_boundary** - effective only in OVAL checks, if
-        set to `"true"` it will test if the value respects the a zero
-        boundary operation. The operation is defined by the
-        **zero_boundary_operation** parameter (default value: `"false"`)
-
-    -   **zero_boundary_operation** - OVAL operation, eg. `greater than`
-        (default value: `"greater than"`)
+    -   **zero_comparison_operation** - (optional) OVAL operation, eg. `greater than`.
+        When set, it will test if the **variable** value matches the OVAL operation
+        when compared to zero.
 
 -   Languages: Ansible, Bash, OVAL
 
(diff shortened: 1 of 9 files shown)


linux_os/guide/system/accounts/accounts-pam/password_quality/password_quality_pwquality/accounts_password_pam_maxclassrepeat/rule.yml
```diff
@@ -56,3 +56,5 @@ template:
     vars:
         variable: maxclassrepeat
         operation: less than or equal
+        nonzero_check: "false"
+        nonzero_check@ol7: "true"
```
ggbecker, line 60:
```suggestion
        nonzero_check: "true"
```

I believe this change can be applied to all products. RHEL STIG also brings this change in the latest releases. Please update the description accordingly as well.

ggbecker (changes requested):
1 - Can you please add documentation for this new template attribute here: https://github.com/ComplianceAsCode/content/blame/master/docs/templates/template_reference.md#L3

2 - At the same place above, please update the `Languages` to `Ansible, Bash, OVAL` since the template provides all these languages and the section was outdated apparently.

3 - Please update the rule's description to in sync with what the OVAL check does. Include in the description that the value `0` is not accepted as a compliant value.

ggbecker:
> 1 - Can you please add documentation for this new template attribute here: https://github.com/ComplianceAsCode/content/blame/master/docs/templates/template_reference.md#L3
> 
> 2 - At the same place above, please update the `Languages` to `Ansible, Bash, OVAL` since the template provides all these languages and the section was outdated apparently.
> 
> 3 - Please update the rule's description to in sync with what the OVAL check does. Include in the description that the value `0` is not accepted as a compliant value.

I took the liberty to push some commits to address the requests because it also benefits other products and we want to merge this ASAP. Thanks for opening this.

ggbecker (approved):

Xeicker:
> I took the liberty to push some commits to address the requests because it also benefits other products and we want to merge this ASAP. Thanks for opening this.

Sorry you needed to do so. I wasn't near to a computer these days

shared/templates/accounts_password/oval.template
```diff
@@ -24,5 +28,18 @@
...
 
+  {{% if CHECK_ZERO_BOUNDARY %}}
+  <ind:textfilecontent54_test check="all"
+  comment="check the configuration of /etc/security/pwquality.conf"
+  id="test_password_pam_pwquality_{{{ VARIABLE }}}_zero_boundary" version="1">
+    <ind:object object_ref="obj_password_pam_pwquality_{{{ VARIABLE }}}" />
+    <ind:state state_ref="state_password_pam_{{{ VARIABLE }}}_zero_boundary" />
+  </ind:textfilecontent54_test>
+
+  <ind:textfilecontent54_state id="state_password_pam_{{{ VARIABLE }}}_zero_boundary" version="1">
```
yuumasato, line 39: You can also add the state `state_password_pam_{{{ VARIABLE }}}_zero_boundary` to the test `test_password_pam_pwquality_{{{ VARIABLE }}}`. Then the object will have to conform to both states.

If you'd like to be more explicit `state_operator="AND"` could be added to the test too.
Example of `state_operator` use: https://github.com/ComplianceAsCode/content/blob/2b022d2af8c5c3f18b2bc123df3638b90248a532/linux_os/guide/system/accounts/accounts-restrictions/root_logins/no_shelllogin_for_systemaccounts/oval/shared.xml#L155

docs/templates/template_reference.md
```diff
@@ -2,17 +2,24 @@
...
+    -   **variable** - PAM `pam_pwquality` password quality
+        requirement, eg. `ucredit`, `ocredit`
 
     -   **operation** - OVAL operation, eg. `less than or equal`
 
--   Languages: OVAL
+    -   **check_zero_boundary** - effective only in OVAL checks, if
+        set to `"true"` it will test if the value respects the a zero
+        boundary operation. The operation is defined by the
+        **zero_boundary_operation** parameter (default value: `"false"`)
```
yuumasato, line 17: Would it make sense to get rid of `check_zero_boundary` and just use `zero_boundary_operation`?
(Suggestion) And maybe rename it to `zero_comparison_operation`

docs/templates/template_reference.md
```diff
@@ -2,17 +2,24 @@
 
 #### accounts_password
 -   Checks if PAM enforces password quality requirements. Checks the
-    configuration in `/etc/pam.d/system-auth` (for RHEL 6 systems) or
-    `/etc/security/pwquality.conf` (on other systems).
+    configuration in `/etc/security/pwquality.conf`.
```
yuumasato, line 5: Maybe we should add to the template description that it only compares the parameter value against the xccdf variable, that it does not check the value of the parameter in relation to zero.
And to compare the parameter value against zero, `zero_boundary_operation` needs to be set.

docs/templates/template_reference.md
```diff
@@ -2,17 +2,24 @@
...
-    -   **variable** - PAM `pam_cracklib` (on RHEL 6) or `pam_pwquality`
-        (on other systems) module name, eg. `ucredit`, `ocredit`
+    -   **variable** - PAM `pam_pwquality` password quality
+        requirement, eg. `ucredit`, `ocredit`
 
     -   **operation** - OVAL operation, eg. `less than or equal`
 
--   Languages: OVAL
+    -   **check_zero_boundary** - effective only in OVAL checks, if
+        set to `"true"` it will test if the value respects the a zero
```
yuumasato, line 15: text suggestion: test if the parameter value respects the `zero_boundary_operation` when compared to zero.

yuumasato (approved):

PR #8166, opened 2022-02-04
Update ensure_oracle_gpgkey_installed

freddieRv:
#### Description:

- Remove hardcoded fingerprints from bash remediation and add them as variables in the respective product.yml files
- Add logic to bash script to allow having a release key only, or both a release and backup keys
- Also modify the OVAL content to use the template approach to generate product-specific content
- Additionally add `pkg_release` and `pkg_version` entries to the OL product.yml files to be able to use the OVAL template approach

#### Rationale:

- Since the OL9 product was introduced, this update was needed
- Also using the template approach for the OVAL content makes it easier to maintain


jan-cerny (approved):
lgtm

PR #8170, opened 2022-02-07
Update rule to only remove the graphical interface

yuumasato:
#### Description:

- Removed configuration of runlevel target from rule `xwindows_remove_packages`.
- Remediations now only remove the graphical interface packages.
- The warning now mentions the rule that can be used to set the runlevel target.

#### Rationale:

- The rule was doing two configurations at once, removing graphical environment and setting the target runlevel.
  Now it only removes packages related to the graphical interface, the target runlevel is set by `xwindows_runlevel_target`.

### DIFF
--- a/linux_os/guide/services/xwindows/disabling_xwindows/xwindows_remove_packages/rule.yml
+++ b/linux_os/guide/services/xwindows/disabling_xwindows/xwindows_remove_packages/rule.yml
@@ -19,14 +19,6 @@ description: |-
     {{% else %}}
     <pre>sudo {{{ pkg_manager }}} remove xorg-x11-server-Xorg xorg-x11-server-common xorg-x11-server-utils xorg-x11-server-Xwayland</pre>
     {{% endif %}}
-    Additionally, setting the system's default target to
-    <tt>multi-user.target</tt> will prevent automatic startup of the X server.
-    To do so, run:
-    <pre>$ systemctl set-default multi-user.target</pre>
-    You should see the following output:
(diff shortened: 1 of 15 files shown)


linux_os/guide/services/xwindows/disabling_xwindows/xwindows_remove_packages/rule.yml
```diff
@@ -72,6 +64,8 @@ warnings:
         The installation and use of a Graphical User Interface (GUI) increases your attack vector and decreases your
         overall security posture. Removing the package xorg-x11-server-common package will remove the graphical target
         which might bring your system to an inconsistent state requiring additional configuration to access the system
-        again. If a GUI is an operational requirement, a tailored profile that removes this rule should used before
+        again. </br>
+        The rule 'xwindows_runlevel_target' can be used to configure the system to boot into the multi-user.target.</br>
```
ggbecker, line 68:
```suggestion
        The rule <tt>xwindows_runlevel_target</tt> can be used to configure the system to boot into the multi-user.target.</br>
```

Apparrently we can't use single quotes here.

yuumasato:
I need to investigate what is going on here

ggbecker:
In theory, the test scenarios could be updated to remove the `multi-user.target` setting.

yuumasato:
I'll update the tests, they should be aligned with the rule, otherwise it creates confusion.

ggbecker (approved):

PR #8171, opened 2022-02-07
Fix for bug 1195520

anivan-suse:
#### Description:

Add sle15.sh and sle12.sh bash remediation for rule: set_password_hashing_algorithm_systemauth

#### Rationale:

- Fixes bug #1195520 https://bugzilla.suse.com/show_bug.cgi?id=1195520

### DIFF
--- a/linux_os/guide/system/accounts/accounts-pam/set_password_hashing_algorithm/set_password_hashing_algorithm_systemauth/bash/shared.sh
+++ b/linux_os/guide/system/accounts/accounts-pam/set_password_hashing_algorithm/set_password_hashing_algorithm_systemauth/bash/shared.sh
@@ -1,13 +1,20 @@
-# platform = multi_platform_wrlinux,multi_platform_rhel,multi_platform_fedora,multi_platform_ol,multi_platform_rhv
+# platform = multi_platform_all
 
 AUTH_FILES[0]="/etc/pam.d/system-auth"
 {{%- if product == "rhel7" %}}
 AUTH_FILES[1]="/etc/pam.d/password-auth"
 {{%- endif %}}
 
+{{% if product in ["sle15", "sle12"] -%}}
+if ! grep -q "^password.*required.*pam_unix.so.*sha512" /etc/pam.d/common-password ; then
+   sed -i --follow-symlinks "/^password.*required.*pam_unix.so/ s/$/ sha512/" /etc/pam.d/common-password
+fi
+{{%- else -%}}
 for pamFile in "${AUTH_FILES[@]}"
 do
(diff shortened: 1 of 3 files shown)


ggbecker (commented):
I would be more in favor of merging these files with the shared.sh and using Jinja macros to define the filepath and the regexes used. They kind of look similar and we avoid duplicated code.

We probably should reevaluate all of the `sle12.*` `sle15.*` and see if can something similar.

jan-cerny (changes requested):
@anivan-suse After you have moved the SLE12/15 code to `shared.sh` you should now delete the files `sle12.sh` and `sle15.sh`.

anivan-suse:
Thanks for pointing that out - removed.

jan-cerny (approved):

jan-cerny (changes requested):
You have a test failing on SLE 15:

```
175/190 Test #171: sle15-bash-shellcheck ....................................................***Failed   10.24 sec

In /__w/content/content/build/sle15/fixes/bash/set_password_hashing_algorithm_systemauth.sh line 5:
AUTH_FILES[0]="/etc/pam.d/system-auth"
^-----------^ SC2034: AUTH_FILES appears unused. Verify use (or export if used externally).

For more information:
  https://www.shellcheck.net/wiki/SC2034 -- AUTH_FILES appears unused. Verify...
```

brett060102:
@anivan-suse You need to move the block
```
AUTH_FILES[0]="/etc/pam.d/system-auth"
{{%- if product == "rhel7" %}}
AUTH_FILES[1]="/etc/pam.d/password-auth"
{{%- endif %}}
```
under the {{% else %}}
to address the failing test issue.

jan-cerny:
bump!

brett060102:
@anivan-suse not sure you saw the last note the code need to change to the following to fix the failing test:
`# platform = multi_platform_all

{{% if product in ["sle15", "sle12"] -%}}
if ! grep -q "^password.*required.*pam_unix.so.*sha512" /etc/pam.d/common-password ; then
   sed -i --follow-symlinks "/^password.*required.*pam_unix.so/ s/$/ sha512/" /etc/pam.d/common-password
fi
{{%- else -%}}
AUTH_FILES[0]="/etc/pam.d/system-auth"
{{%- if product == "rhel7" %}}
AUTH_FILES[1]="/etc/pam.d/password-auth"
{{%- endif %}}
for pamFile in "${AUTH_FILES[@]}"
do
	if ! grep -q "^password.*sufficient.*pam_unix.so.*sha512" $pamFile; then
		sed -i --follow-symlinks "/^password.*sufficient.*pam_unix.so/ s/$/ sha512/" $pamFile
	fi
done
{{%- endif %}}

`

jan-cerny (approved):

PR #8174, opened 2022-02-07
Update references of RHEL8 STIG audit rules

ggbecker:
#### Description:

- The rules were grouped together during last release and they should have the same STIG id as other STIG ids have been deleted. We are keeping them as separate rules since we think it's still better for tailoring and our rules can remediate within the same line of audit config. The only concern is that the exact ordering might not be identical to what STIG has in its verbiage.

### DIFF
--- a/linux_os/guide/system/auditing/auditd_configure_rules/audit_dac_actions/audit_rules_dac_modification_fchmod/rule.yml
+++ b/linux_os/guide/system/auditing/auditd_configure_rules/audit_dac_actions/audit_rules_dac_modification_fchmod/rule.yml
@@ -58,7 +58,7 @@ references:
     stigid@ol7: OL07-00-030410
     stigid@ol8: OL08-00-030540
     stigid@rhel7: RHEL-07-030420
-    stigid@rhel8: RHEL-08-030540
+    stigid@rhel8: RHEL-08-030490
     stigid@sle12: SLES-12-020470
     stigid@sle15: SLES-15-030300
     stigid@ubuntu2004: UBTU-20-010153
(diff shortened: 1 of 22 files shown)


linux_os/guide/system/auditing/auditd_configure_rules/audit_dac_actions/audit_rules_dac_modification_fchmod/rule.yml
```diff
@@ -58,7 +58,7 @@ references:
     stigid@ol7: OL07-00-030410
     stigid@ol8: OL08-00-030540
     stigid@rhel7: RHEL-07-030420
-    stigid@rhel8: RHEL-08-030540
```
yuumasato, line 61: Should the comments in the profile be removed too?
https://github.com/ComplianceAsCode/content/blob/77730337dd834cefd8e7e8d410d60aa043bd42e2/products/rhel8/profiles/stig.profile#L857

yuumasato (commented):
> The only concert is that the exact ordering might not be identical to what STIG has in its verbiage.

A difference in the order of the syscalls whithin the audit rule will not affect its behavior.

linux_os/guide/system/auditing/auditd_configure_rules/audit_dac_actions/audit_rules_dac_modification_fchmod/rule.yml
ggbecker: That's correct

ggbecker: Done in d871cc8379ecf77429e651df104445c719543f03

ggbecker:
> > 
> 
> A difference in the order of the syscalls whithin the audit rule will not affect its behavior.

I meant the order between what's defined in the STIG document and the one that ends up in the final audit config when we remediate using CaC/content.

yuumasato:
> > > 
> > 
> > 
> > A difference in the order of the syscalls whithin the audit rule will not affect its behavior.
> 
> I meant the order between what's defined in the STIG document and the one that ends up in the final audit config when we remediate using CaC/content.

Yes, the order in the list may vary, yet it should not affect audit's behaviour.

yuumasato (approved):

PR #8175, opened 2022-02-07
Fix SSGTS platform machine overriding mechanism

ggbecker:
#### Description:

- Fix SSGTS platform machine overriding mechanism.

#### Rationale:

- https://github.com/ComplianceAsCode/content/pull/8152#issuecomment-1031703514

That was a silent one. Let's see what happens with the tests.


jan-cerny (approved):
Fixes notapplicable problem in 8152.

PR #8176, opened 2022-02-07
OCP fix insecureEdgeTerminationPolicy check

Vincent056:
We should check route insecureEdgeTerminationPolicy is None/Redirect instead of Disabled/Redirect according to: https://github.com/openshift/api/blob/31ffd77a8f025c73b5d19b390af353f9775e7ef1/route/v1/types.go#L270

Related BZ link: https://bugzilla.redhat.com/show_bug.cgi?id=2002695


jhrozek (approved):
LGTM

PR #8177, opened 2022-02-07
OCP enable variable support for sshd remediation

Vincent056:
We are adding variable support for sshd related remediation to address BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2049141

Without it, users set values will have no effect on generated sshd remediations


jhrozek (commented):
Looks good, but can you rebase so that we get a clean CI run?

jhrozek (approved):
The code looks good and works. Ack!

PR #8180, opened 2022-02-08
Change the grub2 template

matejak:
- Introduce the concept of product-specific bootloader config
  properties that determine the check/remediation form.
- Expand the RHEL8 remediation with a check for update of
  /etc/default/grub contents.
- Add a RHEL8 check that looks for kernelopts references in loader entries.
- Update tests.

The newly-added jinja product properties definitions should probably go in products or into a (so far not existent) file where components could add properties to products, but I would keep those there for the time being.

Aims to fix: https://bugzilla.redhat.com/show_bug.cgi?id=2030966


shared/templates/grub2_bootloader_argument/oval.template
```diff
@@ -54,8 +85,25 @@
...
+    <ind:state state_ref="state_grub2_{{{ SANITIZED_ARG_NAME }}}_argument" />
+  </ind:textfilecontent54_test>
+
+  <ind:textfilecontent54_object id="object_grub2_{{{ SANITIZED_ARG_NAME }}}_argument_grub_env"
+  version="1">
+    <ind:filepath>{{{ grub2_boot_path }}}/grubenv</ind:filepath>
+    <ind:pattern operation="pattern match">^kernelopts=(.*)$</ind:pattern>
+    <ind:instance datatype="int" operation="greater than or equal">1</ind:instance>
+  </ind:textfilecontent54_object>
+{{%- endif %}}
```
yuumasato, line 104: Any special reason to move this block up?

yuumasato (commented):
I like how the OVAL is structured now, the variables modularize the check in a nice way.

But I think the same variables are not applying that well to the remediations, which I think should rely more on `grubby`.
On the remediations, the Jinja variables should probably about the behaviour of `grubby` across the systems, like what it does and doesn't. So if an aspect is false, it means that grubby doesn't handle it, and we need to do it manually.

matejak:
> ...
> But I think the same variables are not applying that well to the remediations, which I think should rely more on `grubby`. On the remediations, the Jinja variables should probably about the behaviour of `grubby` across the systems, like what it does and doesn't. So if an aspect is false, it means that grubby doesn't handle it, and we need to do it manually.

I agree, I see it now as well.

shared/checks/oval/grub2_entries_reference_kernelopts.xml
```diff
@@ -0,0 +1,25 @@
...
+    <ind:textfilecontent54_test id="test_grub2_entries_reference_kernelopts"
+                                comment="check kernel command line parameters for referenced boot entries reference the $kernelopts variable."
+                                check="all" check_existence="all_exist" version="1">
+      <ind:object object_ref="object_grub2_entries_reference_kernelopts" />
+    </ind:textfilecontent54_test>
+
+  <ind:textfilecontent54_object id="object_grub2_entries_reference_kernelopts" version="1">
+    <ind:path>/boot/loader/entries/</ind:path>
+    <ind:filename operation="pattern match">^.*\.conf$</ind:filename>
+    <ind:pattern operation="pattern match">^options .*\b\$kernelopts\b.*$</ind:pattern>
```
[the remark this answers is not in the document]
matejak: Indeed, `$` is not a "word character".

shared/templates/grub2_bootloader_argument/oval.template
matejak: It minimizes datastream diffs across various RHEL versions that this PR introduces.

vojtapolasek:
I have tested the current implementation with grub2_audit_argument rule. All tests fail on rhel8, rhel7 and rhel9 are fine. I will investigate more.

shared/checks/oval/grub2_entries_reference_kernelopts.xml
vojtapolasek, line 22:
```suggestion
    <ind:pattern operation="pattern match">^(options \$kernelopts|options .*\s\$kernelopts)\b.*$</ind:pattern>
```

jhrozek:
/retest
the OCP CI should now be fixed (if not, maybe rebasing on master to make sure the current OCP fixes are pulled in should help)

shared/checks/oval/grub2_entries_reference_kernelopts.xml
```diff
@@ -0,0 +1,25 @@
...
+
+    <ind:textfilecontent54_test id="test_grub2_entries_reference_kernelopts"
+                                comment="check kernel command line parameters for referenced boot entries reference the $kernelopts variable."
+                                check="all" check_existence="all_exist" version="1">
+      <ind:object object_ref="object_grub2_entries_reference_kernelopts" />
+    </ind:textfilecontent54_test>
+
+  <ind:textfilecontent54_object id="object_grub2_entries_reference_kernelopts" version="1">
+    <ind:path>/boot/loader/entries/</ind:path>
+    <ind:filename operation="pattern match">^.*\.conf$</ind:filename>
```
matusmarhefka, line 21: I think we should skip a file ending with `rescue.conf` because that file is not touched by grubby when adding kernel options.

yuumasato: A filter_state should be able to handle that.

yuumasato (changes requested):
Looks good to me, but I think some tests need to be updated.

- [grub2_audit_argument/tests/blank_grubenv_rhel8.fail.sh](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/auditing/grub2_audit_argument/tests/blank_grubenv_rhel8.fail.sh) results in error.
  It seems that when `/etc/grub2/grubenv` doesn't have `kernelopts`, `grubby` doesn't set anything there. In this situation `grub2-editenv - set "kernelopts=audit=1"` needs to be run so that `kernelopts` are added back, but at this point we have already lost any other options previously set.
  I suggest we don't support the case when `grubenv` is cleread and set `# remediation=none` in this test scenario.

- [grub2_audit_argument/tests/correct_grubenv.pass.sh](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/auditing/grub2_audit_argument/tests/correct_grubenv.pass.sh) This test scenario doesn't work on RHEL9, as the `grep` for `kernelopts` finds nothing.

matejak:
I have added support for BIOS and UEFI setups, which affects RHEL8. I am confident that RHEL9-specific part that doesn't check the rescue entry works as intended.

I have modified scenario metadata according to your suggestions. Although I think that it would be nice to have some of them in template test scenarios, I would leave that for other PRs.

yuumasato (approved):

yuumasato:
Nice work on the template.

yuumasato:
The changes look good and test scenarios are passing for me.

PR #8185, opened 2022-02-09
New rules related to pam_pwquality

marcusburghardt:
Introduce two new rules for the following STIG IDs:
- RHEL-08-020100
- RHEL-08-020101

Also updates the rule related to the STIG ID RHEL-08-020104:
- accounts_password_pam_retry


Mab879 (changes requested):
Overall looks good, just one issue with the test stability file.

Mab879 (approved):

Mab879 (approved):

PR #8186, opened 2022-02-09
Extend package_pam_pwquality_installed rule for RHEL

marcusburghardt:
Extend the `rule package_pam_pwquality_installed` to be also applicable for RHEL.

This rule checks if the package which brings the `pam_pwquality.so` module is installed.

### DIFF
--- a/shared/macros-bash.jinja
+++ b/shared/macros-bash.jinja
@@ -638,7 +638,7 @@ zypper install -y "{{{ package }}}"
 {{%- if pkg_manager == "yum" or pkg_manager == "dnf" -%}}
 if rpm -q --quiet "{{{ package }}}" ; then
 {{% if SSG_TEST_SUITE_ENV %}}
-    {{{ pkg_manager }}} remove -y --setopt protected_packages= "{{{ package }}}"
+    rpm -e --nodeps "{{{ package }}}"
 {{% else %}}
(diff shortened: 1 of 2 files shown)


shared/macros-bash.jinja
```diff
@@ -638,7 +638,7 @@ zypper install -y "{{{ package }}}"
 {{%- if pkg_manager == "yum" or pkg_manager == "dnf" -%}}
 if rpm -q --quiet "{{{ package }}}" ; then
 {{% if SSG_TEST_SUITE_ENV %}}
-    {{{ pkg_manager }}} remove -y --setopt protected_packages= "{{{ package }}}"
+    rpm -e --nodeps "{{{ package }}}"
```
matejak, line 641: Won't this break the test suite on non-rpm systems?

matusmarhefka: Yes, here we can probably use similar approach as in https://github.com/ComplianceAsCode/content/blob/master/shared/macros-bash.jinja#L608

marcusburghardt: No, it is included in the conditional for `yum` and `dnf` only. Distros with other package managers are not affected.

matusmarhefka (approved):
LGTM, thank you!

PR #8187, opened 2022-02-09
Pass the rule when no time server nor pool is set

yuumasato:
#### Description:

- Add a test that removes all  `server` and `pool` options from `/etc/chrony.conf`
- Extend the check with another test that passes when there are not `server` or `pool` directives configured.
- Added a new rule to ensure that Chrony is configure with `server` directives only.

#### Rationale:

- If no time server or pool is configured, there is no entry to add maxpoll option to, so the rule should evaluate to pass.
- Follow up from #7910


Mab879:
In the [STIGs referenced](https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_8_security_technical_implementation_guide/V-230484) this rule states "If the parameter "server" is not set or is not set to an authoritative DoD time source, this is a finding.". 

Are we planning on creating a new rule to cover this?

yuumasato:
@ggbecker @Mab879 If I understand your comments, `chronyd_or_ntpd_set_maxpoll` can `pass` when there are no `server` or `pool` to configure. And there should be a new rule that will `fail` if no `server` is configured in chrony?

ggbecker:
> @ggbecker @Mab879 If I understand your comments, `chronyd_or_ntpd_set_maxpoll` can `pass` when there are no `server` or `pool` to configure. And there should be a new rule that will `fail` if no `server` is configured in chrony?

I think that's the most reasonable interpretation.

yuumasato:
There is already a rule about having remote time server specified, but this one also accepts `server` and `pool`.
https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/ntp/chronyd_specify_remote_server/rule.yml

Not sure how to proceed here.

linux_os/guide/services/ntp/chronyd_server_directive/rule.yml
```diff
@@ -0,0 +1,30 @@
...
+
+warnings:
+  - general: This rule doesn't come with a remediation, the time source needs to be added by the adminstrator.
+
+identifiers:
+    cce@rhel8: CCE-86077-5
+
+references:
+    disa: CCI-001891
+    stigid@rhel8: RHEL-08-030740
```
ggbecker, line 23: Please add CCE for RHEL9 and select this new rule in the RHEL9 STIG profile as well.

Also add the SRG references from https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_8_security_technical_implementation_guide/V-230484
`
Satisfies: SRG-OS-000355-GPOS-00143, SRG-OS-000356-GPOS-00144, SRG-OS-000359-GPOS-00146`

yuumasato:
Addressed feedback and rebased on latests master.

linux_os/guide/services/ntp/chronyd_server_directive/rule.yml
```diff
@@ -0,0 +1,32 @@
...
+warnings:
+  - general: This rule doesn't come with a remediation, the time source needs to be added by the adminstrator.
+
+identifiers:
+    cce@rhel8: CCE-86077-5
+    cce@rhel9: CCE-87077-4
+
+references:
+    disa: CCI-001891
+    srg: SRG-OS-000355-GPOS-00143, SRG-OS-000356-GPOS-00144, SRG-OS-000359-GPOS-00146
```
ggbecker, line 24:
```suggestion
    srg: SRG-OS-000355-GPOS-00143,SRG-OS-000356-GPOS-00144,SRG-OS-000359-GPOS-00146
```

PR #8189, opened 2022-02-09
Propose a different layout for kubernetes rules

rhmdnd:
Initially, this repository only contained profiles for OCP. With the
addition of EKS profiles, we now have two distributions and we should
think about how we want to organize the rules for kubernetes platforms.

The new layout must:

  - Render accurate content for each kubernetes distribution
  - Allow for rule re-use wherever possible
  - Keep existing rule names for backwards compatibility

The first point is important so that end-users are presented with
content that matches the official benchmarks word-for-word.

This commit adds a new group under `applications/` for `kubernetes`
distributions. Each rule added to this new layout should be written as
generically as possible if it pertains to more than one kubernetes
platform (e.g., EKS or OCP). Existing rules under
`applications/openshift/` will be ported to `applications/kubernetes/`
without changing the rule name to preserve backwards compatibility.
Those changes will also include parameterization templating in the rule
to support each platform currently using it.

This will result in template-heavy rules, but it provides the following
benefits:

  - All kubernetes-specific rules are in a platform agnostic location
  - We won't have issues with multiple sub-groups with the same name
    (e.g., `applications/kubernetes/eks/logging` and
    `applications/kubernetes/openshift/logging` see issue 8190) which
    breaks the content build process
  - We can still reuse rules across platforms

Additionally, this commit adds a new rule for the EKS CIS profile as an
example. Further, more complicated examples will be ported in future
patches.

#### Description:

#### Rationale:


applications/kubernetes/audit_logging/rule.yml
```diff
@@ -0,0 +1,56 @@
...
+    <li>Click Logging</li>
+    <li>Ensure all 5 choices are set to Enabled</li>
+  </ol>
+  If any of the choices are disabled, this is a finding.
+
+  Via the CLI
+  <tt>aws --region "<i>${REGION_CODE}</i>" eks describe-cluster --name "<i>${CLUSTER_NAME}</i>" --query 'cluster.logging.clusterLogging[?enabled==true].types'</tt>
+  If no results are returned, this is a finding.
+
+fix: |-
```
jhrozek, line 46: For my own education, how is this attribute translated into the DataStream? I know the OCIL is then available as `question_text` element, but how do we handle (esp. through CO) the `fix` attribute?

rhmdnd: That's a good question. I don't think we do (at least I haven't been able to find a way to get that to show up in rules created by the compliance-operator).

I started writing these following some of the advice in the CaC guides and `fix` seems like the most appropriate place for the remediation.

rhmdnd: Well - I think we can back out this rule from this PR since it landed in a separate patch [0] where I kept the remediation in the OCIL.

I'd be curious to know if someone from the CaC project could advise on using `fix` in this context. I think we'd need to add it to the datastream before we can update the CO to consume it.

[0] https://github.com/ComplianceAsCode/content/commit/07b62a3cd819c1c5d33c16153ba24a5eb79a163d

jan-cerny: AFAIK the contents of the "fix" key doesn't propagate to the built SCAP source data stream. It's used only by utils/create_srg_export.py. Recently, @Mab879 started to use the "fix"  for RHEL9 STIG, for example in https://github.com/ComplianceAsCode/content/pull/8321. @Mab879 do you have some more information about this "fix"?

Mab879: @jan-cerny That is correct. To the best of the knowledge only `utils/create_srg_export.py` uses the `fix` key. The `fix` was designed to be human readable.

jan-cerny:

PR #8192, opened 2022-02-10
Fix bug 1195525

anivan-suse:
#### Description:
Change bash/shared.sh  for rule file_permissions_home_directories

#### Rationale:
- Fixes #1195525: https://bugzilla.suse.com/show_bug.cgi?id=1195525


marcusburghardt (approved):
Thanks for this fix, it makes the rule also compatible with distros which users share a primary group.

PR #8194, opened 2022-02-10
Add missing_file_pass to owner/group-owner rules

dodys:
#### Description:

With commit fe36b355 file_groupowner and file_owner started to filter/
exclude symlinks. Unfortunately the rules touched here, have some main
paths that are actually a symlink (e.g. /lib64 on Ubuntu) and the filter
together with the all_exist rule in their OVAL, make the rule evaluation
to fail. By adding missing_file_pass, we change it to any_exist.

### DIFF
--- a/linux_os/guide/system/permissions/files/permissions_within_important_dirs/dir_group_ownership_library_dirs/rule.yml
+++ b/linux_os/guide/system/permissions/files/permissions_within_important_dirs/dir_group_ownership_library_dirs/rule.yml
@@ -66,3 +66,4 @@ template:
             - /usr/lib64/
         recursive: 'true'
         filegid: '0'
+        missing_file_pass: 'true'
--- a/linux_os/guide/system/permissions/files/permissions_within_important_dirs/dir_groupownership_binary_dirs/rule.yml
+++ b/linux_os/guide/system/permissions/files/permissions_within_important_dirs/dir_groupownership_binary_dirs/rule.yml
@@ -63,3 +63,4 @@ template:
             - /usr/local/sbin/
         recursive: 'true'
         filegid: '0'
+        missing_file_pass: 'true'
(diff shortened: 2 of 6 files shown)


ggbecker:
> Who will provide them?

let's ping @dodys and see if he replies in a couple of days... otherwise we can submit some tests to get this merged.

dodys:
> > Who will provide them?
> 
> let's ping @dodys and see if he replies in a couple of days... otherwise we can submit some tests to get this merged.

sorry, I haven't had the time to take a look at it.
Also, I'm not sure what kind of tests you think it would be good as the `missing_file_test.pass.sh` template test is just doing a `rm -rf`

ggbecker:
> > > Who will provide them?
> > 
> > 
> > let's ping @dodys and see if he replies in a couple of days... otherwise we can submit some tests to get this merged.
> 
> sorry, I haven't had the time to take a look at it. Also, I'm not sure what kind of tests you think it would be good as the `missing_file_test.pass.sh` template test is just doing a `rm -rf`

The logs have:
```
+ rm -f /lib/
rm: cannot remove '/lib/': Is a directory
+ rm -f /lib64/
rm: cannot remove '/lib64/': Is a directory
+ rm -f /usr/lib/
rm: cannot remove '/usr/lib/': Is a directory
+ rm -f /usr/lib64/
rm: cannot remove '/usr/lib64/': Is a directory
```

I believe the templated scenario should be updated to consider if the path is a directory, but apart from that... it's probably not a good idea to remove those directories even in a test, it might break the system. So a custom test scenario might help in this case.

dodys:
This PR doesn't seem needed anymore with the fixes done in PR #8456
As soon as that PR makes to master I will re-run the tests and confirm it here.

dodys:
> This PR doesn't seem needed anymore with the fixes done in PR #8456 As soon as that PR makes to master I will re-run the tests and confirm it here.

Confirmed and closing this PR.

PR #8196, opened 2022-02-10
Add rule to check for default sudoers includedir

yuumasato:
#### Description:

- Add rule that ensures sudo only includes drop-in files from `/etc/sudoers`.

#### Rationale:

- This rule supports RHEL-08-010379


Xeicker:
I was reviewing an equivalent requirement for OL7 and apparently this misses this part of the requirement:

> Verify the operating system does not have nested "include" files or directories within the /etc/sudoers.d directory with the following command:
> 
> $ sudo grep include /etc/sudoers.d/*
> 
> If results are returned, this is a finding. 

Is it intended to be included in another rule?

ggbecker:
> I was reviewing an equivalent requirement for OL7 and apparently this misses this part of the requirement:
> 
> > Verify the operating system does not have nested "include" files or directories within the /etc/sudoers.d directory with the following command:
> > $ sudo grep include /etc/sudoers.d/*
> > If results are returned, this is a finding.
> 
> Is it intended to be included in another rule?

I believe we should include the check in this rule. We need to extend the check to look for `#include` within `/etc/sudoers.d/*` files and make sure none exists. The remediation should erase any `#include` in those files.

ggbecker (approved):

Issue #8198, opened 2022-02-10
Kubernetes content is specific to OpenShift


rhmdnd:
#### Description of problem:

OpenShift was the first kubernetes distribution with content in this repository. As a result, many of the rules and rule groups in `applications/openshift` are OpenShift-specific.

Now that we are adding content for other distributions, we should find a way to organize the content so it's platform agnostic. This 
makes it easier to use the content for EKS, ASK, OpenShift, and vanilla kubernetes.

Ideally, it would be great to:

1.) Reuse as many existing rules as we can
2.) Have content to be specific to the benchmark in question (content shouldn’t bleed details about irrelevant distributions when in use)
3.) Easily navigate content

##### Group rules by distribution

The layout would adhere to the following convention, where groups could be nested as needed:

```applications/kubernetes/{distribution}/{group}/{rule}```

Examples of the directory structure, omitting rules and only focusing on layout:

```applications/kubernetes/``` - contains all common rules for kubernetes
```applications/kubernetes/eks/``` - contains rules specific to EKS
```applications/kubernetes/aks/```- contains rules specific to AKS
```applications/kubernetes/openshift/``` - contains rules specific to OpenShift
```applications/kubernetes/eks/logging/``` - contains rules specific to EKS logging
```applications/kubernetes/openshift/logging/``` - contains rules specific to OpenShift logging

Benefits

- Well organized
- Easy to find rules that are specific to a distribution
- Easy to reuse common rules that are generally applicable to kubernetes

Drawbacks

- Requires a lot of refactoring to get there
- Having two groups with the same name and product type [breaks](https://github.com/ComplianceAsCode/content/issues/8190) CaC content builds
  - Example: we can’t do applications/kubernetes/eks/logging and applications/kubernetes/openshift/logging
- We would need to be careful nesting groups under the distribution and it would be inconsistent to have some logging rules in ```logging/``` for one distribution and not another

##### Group rules by component subsystem

The layout would adhere to the following convention:

```applications/kubernetes/{group}/{distribution}/{rule}```

Examples of the directory structure, omitting rules and only focusing on layout:

```applications/kubernetes/``` - contains all common rules for kubernetes
```applications/kubernetes/logging/``` - contains rules specific to logging
```applications/kubernetes/worker/``` - contains rules specific to worker nodes
```applications/kubernetes/logging/eks/``` - contains rules specific to EKS logging
```applications/kubernetes/logging/openshift/``` - contains rules specific to OpenShift logging
[...]

PR #8199, opened 2022-02-10
Generalize OpenShift groups

rhmdnd:
We're now starting to use some OpenShift rules for EKS checks. This
commit updates the groups for those specific rules, which contain
`prodtype: eks`, to remove OpenShift and generalize it with Kubernetes.

This commit is related to issue #8198

#### Description:

#### Rationale:

### DIFF
--- a/applications/openshift/group.yml
+++ b/applications/openshift/group.yml
@@ -2,18 +2,20 @@ documentation_complete: true
 
 prodtype: ocp4,eks
 
-title: 'OpenShift Settings'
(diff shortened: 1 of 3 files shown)


Vincent056 (approved):

jhrozek:
@rhmdnd can you rebase atop origin/master and force-push so that we get a clean CI run and can be sure that we don't break anything with the patch?

jan-cerny (approved):

PR #8201, opened 2022-02-10
OCP4: Fix api_server_no_adm_ctrl_plugins_disabled jq filter

Vincent056:
The previous jq filter is outputting null instead of empty when there is no `disable-admission-plugins` present, this pr updated the jq filter, so that it will output empty instead of null.


jhrozek (approved):
LGTM

PR #8203, opened 2022-02-10
Parameterize kubernetes worker rules

rhmdnd:
Some of the worker rules are used in both EKS and OpenShift profiles.
This commit updates the rules to use parameters based on the product
(either EKS or OCP) so that the descriptions, rationale, and OCIL are
accurate for the user and don't bleed across products.

This is related to issue #8198

#### Description:

#### Rationale:


applications/openshift/worker/file_permissions_worker_kubeconfig/rule.yml
```diff
@@ -8,26 +8,77 @@ platforms:
 {{%- if product == "eks" %}}
 {{%- set octal_perms = "0644" %}}
```
Vincent056, line 9: the `octal_perms` doesn't seem to be used?

Vincent056:
No sure why did some of the unit test are failing

applications/openshift/worker/file_groupowner_kubelet_conf/rule.yml
```diff
@@ -6,22 +6,66 @@ platform: {{{ product }}}-node
...
+  Verify that the ownership is set to <tt>root:root</tt>
+{{% endset %}}
 {{%- else %}}
 {{%- set kubeletconf_path = "/etc/kubernetes/kubelet.conf" %}}
+{{% set description %}}
+  Ensure that the <tt>kubelet.conf</tt> file ownership is set to
+  <tt>root:root</tt>.
+{{% endset %}}
+{{% set rationale %}}
+  The <tt>kubelet.conf</tt> file is the kubeconfig file for the node, and
```
jhrozek, line 43: We have quite a few rules that check permissions and ownership, usually doubled for user and group checks. Wouldn't it be nicer to write macros in `shared/macros.jinja` that branch per distribution and expand to either ssh-ing to the nodes for EKS or generic Kubernetes distros and oc debug in a loop for OCP? That way, we could reduce the amount of boilerplate in the rules themselves.

applications/openshift/worker/file_groupowner_kubelet_conf/rule.yml
```diff
@@ -6,22 +6,66 @@ platform: {{{ product }}}-node
...
+  <tt>for node in $(oc get nodes -o jsonpath='{.items[*].metadata.name}')
+  do
+  oc debug node/${node} -- chroot /host stat -c %U:%G /etc/kubernetes/kubelet.conf
+  done</tt>
+  Verify that the ownership is set to <tt>root:root</tt>.
+{{% endset %}}
 {{%- endif %}}
 
-title: 'Verify Group Who Owns The Kubelet Configuration File'
+title: 'Verify Ownership of the Kubelet Configuration File'
```
jhrozek, line 59: Shouldn't the title still say verify group ownership?

applications/openshift/worker/file_groupowner_kubelet_conf/rule.yml
jhrozek: btw see both the definition and usage of `rhcos_node_login_instructions` for example.

rhmdnd: That's a good point. My initial goal here was to propose the content as written word-for-word in the various benchmarks.

You're suggesting we write macros that expand out different boilerplate content based on the product?

applications/openshift/worker/file_groupowner_kubelet_conf/rule.yml
rhmdnd: After reading the benchmark I decided to generalize the title since it says to match the user and group.

Somewhat related, could this just be the control section title (`Ensure that the kubelet configuration file ownership is set to root:root`)?

applications/openshift/worker/file_permissions_worker_kubeconfig/rule.yml
rhmdnd: Oh - good find. I'll remove that in a subsequent push.

applications/openshift/worker/file_groupowner_kubelet_conf/rule.yml
jhrozek: Yeah. I didn't do any research on how easy or hard it would be, so it's just a suggestion. But I was thinking something like `cluster_node_login_instructions` that would provide the boilerplate and branch based on the distro, `ssh` for generic kube and `oc debug` for OCP..

applications/openshift/worker/file_groupowner_kubelet_conf/rule.yml
jhrozek: I guess? I wanted to mostly point out that there's always two rules to check ownership, one for user ownership and one for group

jan-cerny:
@rhmdnd please resolve the conflicts

jan-cerny:

PR #8019, opened 2021-12-29
CFE-214 : SC-6 : CaC rule that checks workloads for resource requests/limits

Connect2naga:
CFE-214 : Inital commit with basic rule to validate resource request

#### Description:

  When deploying an application, it is important to tune based on memory and CPU consumption,
  allocating enough resources for the application to function properly. Images provided by
  OpenShift Dedicated behave properly within the confines of the memory they are allocated.
  However, any applicaiton images must pay attention to the specific resources required to
  ensure they are available.

  If the node where a Pod is running has enough of a resource available, it's possible (and allowed)
  for a container to use more resource than its request for that resource specifies.
  However, a container is not allowed to use more than its resource limit.

#### Rationale:

  Resource requests/limits provide constraints that limit aggregate resource consumption
  per container. This helps prevent resource starvation. When deploying your
  application, it is important to tune based on memory and CPU consumption,
  allocating enough resources for the application to function properly.

- Fixes # CFE-214

PR #8208, opened 2022-02-11
OCP4: add SC-6 rules under SC-6 controls

Vincent056:
Rules resource_requests_limits_in_daemonset, resource_requests_limits_in_statefulset and resource_requests_limits_in_deployment were not added under SC-6 control in a previous PR: https://github.com/ComplianceAsCode/content/pull/8019, therefore let's add it now.

### DIFF
--- a/controls/nist_ocp4.yml
+++ b/controls/nist_ocp4.yml
@@ -13251,7 +13251,7 @@ controls:
     applicable to the configuration of OpenShift.
   rules: []
 - id: SC-6
-  status: pending
+  status: automated
   notes: |-
     The customer is responsible for the protection and availability of
     resources by allocating processor and memory resources by process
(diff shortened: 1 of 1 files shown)


controls/nist_ocp4.yml
```diff
@@ -13251,7 +13251,7 @@ controls:
     applicable to the configuration of OpenShift.
   rules: []
 - id: SC-6
-  status: pending
+  status: supported
```
jhrozek, line 13254: why supported and not automated? (Genuine question, I don't remember the difference)

Vincent056: that's a good catch, it should be automated, supported means missing content automation

jhrozek (approved):
LGTM

PR #8210, opened 2022-02-11
Add EKS CIS section 2.1.1 rule

rhmdnd:
This rule is functionally trivial since it doesn't perform a scan using
the compliance operator and it doesn't remediate an issue, yet. To do
that, we require some additional functionality in compliance-operator to
interact with AWS endpoints.

In the mean time, the rule is informative about how to perform the audit
and how to remediate it.

#### Description:

#### Rationale:


Vincent056 (approved):
/lgtm

rhmdnd:
Just for other reviews since I spent some time digging into it. The SSGTS failure is because the rule proposed in this patch doesn't have an accompanying test since it's not automating anything.

Vincent056:
> > Just for other reviews since I spent some time digging into it. The SSGTS failure is because the rule proposed in this patch doesn't have an accompanying test since it's not automating anything.
> 
> Stupid question because I don't know the build system that well...is just this PR going to fail that test or will the test keep failing for other PRs if we merge it?

Only this PR will fail, it should be okay to merge

jhrozek:
> > > Just for other reviews since I spent some time digging into it. The SSGTS failure is because the rule proposed in this patch doesn't have an accompanying test since it's not automating anything.
> > 
> > 
> > Stupid question because I don't know the build system that well...is just this PR going to fail that test or will the test keep failing for other PRs if we merge it?
> 
> Only this PR will fail, it should be okay to merge

Thanks for confirming! I'm going to click the merge button, so that I get yelled at in case we break something :-)

PR #8211, opened 2022-02-12
Add new DISA STIG rules to Mozilla Firefox

lenox-joseph:
#### Description:

Implementation of new rules for DISA's STIG profile for Firefox. Includes a new "set locked config" template for the new items (existing ones can be converted if desired)

#### Rationale:

DISA added new hardening guidelines for Firefox and they weren't implemented here. I use the remediation on my systems, so here we are.


lenox-joseph:
> tests/test_rule_in_container.sh --dontclean --logdir logs_ansible --remediate-using ansible --name ssg_test_suite --datastream ssg-rhel8-ds.xml firefox_preferences-enhanced_tracking,firefox_preferences-extension_recommendation,firefox_preferences-cryptomining,firefox_preferences-disable_ssl_weak_ciphers,firefox_preferences-telemetry,firefox_preferences-telemetry-archive,firefox_preferences-fingerprinting-protection

Any idea what I did wrong for the test system to assume that these checks from the Firefox profile ended up being looked for in the rhel8 datastream? These are all of my new rules this was being run (and failing) before I had written any tests (and I do have templated tests now).

ggbecker:
> 

This is a limitation of our diff detection system. I believe in this case it's looking in the `prodtype` field, and since it has no value in the rules proposed here, it will consider `rhel8` as the default product. The diff would need to consider that the files are coming from a different benchmark directory than the one set in `rhel8` but it can become overly complicated.

Setting `prodtype: firefox` may workaround this issue but it wouldn't be the best scenario for the content per se since it resides in the firefox dir only.

shared/templates/firefox_lockparameter/tests/correct_value.pass.sh
```diff
@@ -0,0 +1,22 @@
...
+    # write our bad config in every location to ensure remediation fixes it.
+    mkdir -p ${firefox_dir}
+    touch ${firefox_dir}/${firefox_cfg}
+
+    # delete the parameter if it's present.
+    if LC_ALL=C grep -q "^lockPref(\"${FIREFOX_PARAM}\"" $1; then
+        sed "/^lockPref(\"${FIREFOX_PARAM}\"/Id" $1
+    fi
+    # put bad value in
+    echo 'lockPref("'"${FIREFOX_PARAM}"'", '"${FIREFOX_VAL}"');' &gt;&gt; "${firefox_dir}/${firefox_cfg}"
```
ggbecker, line 21:
```suggestion
    # put correct value in
    echo 'lockPref("'"${FIREFOX_PARAM}"'", '"${FIREFOX_VAL}"');' >> "${firefox_dir}/${firefox_cfg}"
```

lenox-joseph: I'm down with this; I had been cribbing off of the existing output. Ideally most of the existing rules should be converted to template.

lenox-joseph:
> > 
> 
> This is a limitation of our diff detection system. I believe in this case it's looking in the `prodtype` field, and since it has no value in the rules proposed here, it will consider `rhel8` as the default product. The diff would need to consider that the files are coming from a different benchmark directory than the one set in `rhel8` but it can become overly complicated.
> 
> Setting `prodtype: firefox` may workaround this issue but it wouldn't be the best scenario for the content per se since it resides in the firefox dir only.

I'll give that a shot just to see if it's an effective workaround. I don't have the scope to fix the diff detection system.

ggbecker:
> > > 
> > 
> > 
> > This is a limitation of our diff detection system. I believe in this case it's looking in the `prodtype` field, and since it has no value in the rules proposed here, it will consider `rhel8` as the default product. The diff would need to consider that the files are coming from a different benchmark directory than the one set in `rhel8` but it can become overly complicated.
> > Setting `prodtype: firefox` may workaround this issue but it wouldn't be the best scenario for the content per se since it resides in the firefox dir only.
> 
> I'll give that a shot just to see if it's an effective workaround. I don't have the scope to fix the diff detection system.

I believe you need to set the prodtype in all new files you are proposing. Otherwise it will always give priority to `rhel8` content If I'm not mistaken.

lenox-joseph:
> I believe you need to set the prodtype in all new files you are proposing. Otherwise it will always give priority to `rhel8` content If I'm not mistaken.

Thought process was that the one I marked the prodtype for should disappear from the error (which did not occur).

lenox-joseph:
@ggbecker  apparently yes everything not in a prod that is tested does need to be tagged with that prod for the entire diff. Good to know that I don't necessarily want to mix firefox changes and non with the same PR.

ggbecker (approved):

ggbecker (changes requested):
How about updating the file https://github.com/ComplianceAsCode/content/blob/master/products/firefox/profiles/stig.profile with information about version, SME, etc similarly as we have here: https://github.com/ComplianceAsCode/content/blob/master/products/rhel8/profiles/stig.profile#L3-L15

lenox-joseph:
@ggbecker are you offering yourself up as tribute to be SME, or were you implying that I should offer myself?

ggbecker:
> @ggbecker are you offering yourself up as tribute to be SME, or were you implying that I should offer myself?

I believe the most important metadata here is the version. But SME is always welcomed... feel free to add your name if you are willing to do. It's mostly to know who to contact in case there are unknowns and etc.

ggbecker (approved):

PR #8213, opened 2022-02-14
RHCOS: IA-2(4) control response

Vincent056:
Added control response for IA-2(4)

### DIFF
--- a/controls/nist_rhcos4.yml
+++ b/controls/nist_rhcos4.yml
@@ -6784,12 +6784,12 @@ controls:
   - high
   - moderate
 - id: IA-2(4)
-  status: pending
+  status: automated
   notes: |-
-    A control response is planned. Engineering progress can be tracked via:
-
-    https://issues.redhat.com/browse/CMP-303
-  rules: []
+    The only way to access the nodes should through '$oc debug', remote SSH should be disabled after the
+    Installation
+  rules: 
+  - package_openssh_removed
   description: |-
     The information system implements multifactor authentication for local access to non-privileged accounts.
   title: >-
--- a/linux_os/guide/services/ssh/package_openssh_removed/rule.yml
+++ b/linux_os/guide/services/ssh/package_openssh_removed/rule.yml
@@ -1,6 +1,6 @@
 documentation_complete: true
 
-prodtype: opensuse,sle12,sle15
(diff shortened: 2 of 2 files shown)


controls/nist_rhcos4.yml
```diff
@@ -6784,8 +6784,12 @@ controls:
   - high
   - moderate
 - id: IA-2(4)
-  status: pending
+  status: not applicable
   notes: |-
+    This is out of scope, RHCOS does not provide MFA natively, only through 3rd party IDM such as LDAP. 
+    CoreOS Tectonic User Auth
+    https://cloud.redhat.com/learn/topics/tectonic
```
jhrozek, line 6791: This page says that tectonic is EOL. I wouldn't mention it. Let's mention LDAP, AD or RHIDM instead.
But a bigger question is, should we just punt on the user management in RHCOS altogether like we did for STIG and even make sure that e.g. SSSD is disabled and the only way to access the nodes is through `oc debug`?
Because..if we do suggest that you can use 2FA to log in to RHCOS, what are our options other than letting someone install SSSD or hand-configuring SSH with certificates/yubikeys/whatnot?

Also, remove the blob before about control response is planned.

Vincent056: Thanks for checking out the PR, I think that makes sense, maybe we should say, the only way to access the nodes is through `oc debug` because SSSD is disabled.

Vincent056: So all the authentication is handled on OCP, through a third party IDP

jhrozek: That's really a design question of sorts. I don;t think that moderate had this kind of requirement for direct access to nodes. I think we should consider what our options are and also take into consideration that after the default installation, there is the `core` account which logs in with an ssh key...is it enough to make sure that the ssh key is protected by a passphrase? Or do we need to even disable ssh?

Let's gather all the answers and not rush the PR in..

Vincent056: That makes sense, let's hold this PR for now, yes there is a core account with ssh key, I think maybe we can also block the ssh port on the network firewall.

Vincent056: /hold

Vincent056:
close this one in favor of https://github.com/ComplianceAsCode/content/pull/8512

PR #8215, opened 2022-02-14
Fix bug 1195521

anivan-suse:
#### Description:
Rule: accounts_maximum_age_login_defs

#### Rationale:
- Fixes # bug 1195521: https://bugzilla.suse.com/show_bug.cgi?id=1195521

### DIFF
--- a/linux_os/guide/system/accounts/accounts-restrictions/password_expiration/accounts_maximum_age_login_defs/bash/shared.sh
+++ b/linux_os/guide/system/accounts/accounts-restrictions/password_expiration/accounts_maximum_age_login_defs/bash/shared.sh
@@ -1,4 +1,4 @@
-# platform = multi_platform_wrlinux,multi_platform_rhel,multi_platform_ol,multi_platform_rhv,multi_platform_fedora,multi_platform_ubuntu,multi_platform_sle
+# platform = multi_platform_all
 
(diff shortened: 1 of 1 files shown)


linux_os/guide/system/accounts/accounts-restrictions/password_expiration/accounts_maximum_age_login_defs/bash/shared.sh
```diff
@@ -1,4 +1,4 @@
-# platform = multi_platform_wrlinux,multi_platform_rhel,multi_platform_ol,multi_platform_rhv,multi_platform_fedora,multi_platform_ubuntu
+# platform = multi_platform_wrlinux,multi_platform_rhel,multi_platform_ol,multi_platform_rhv,multi_platform_fedora,multi_platform_ubuntu,multi_platform_sle
```
matejak, line 1: The rule is applicable everywhere, so technically, it makes sense to restrict the remediation. However, the remediation is generic and completely backed by the rule description, so I would rather suggest putting `multi_platform_all` here. WDYT @yuumasato or @ggbecker ?

anivan-suse: Fixed! Thanks :)

Mab879 (approved):

PR #8216, opened 2022-02-14
Fix for bug 1195652

anivan-suse:
#### Description:

Rule: aide_check_audit_tools

#### Rationale:
- Fixes # bug 1195652: https://bugzilla.suse.com/show_bug.cgi?id=1195652

### DIFF
--- a/linux_os/guide/system/software/integrity/software-integrity/aide/aide_check_audit_tools/bash/shared.sh
+++ b/linux_os/guide/system/software/integrity/software-integrity/aide/aide_check_audit_tools/bash/shared.sh
@@ -18,6 +18,12 @@
 {{% set auditfiles = auditfiles + ["/usr/sbin/audispd"] %}}
 {{% endif %}}
 
+{{% if 'rhel' not in product %}}
+{{% set configString = 'p+i+n+u+g+s+b+acl+selinux+xattrs+sha512' %}}
+{{% else %}}
+{{% set configString = "p+i+n+u+g+s+b+acl+xattrs+sha512" %}}
+{{% endif %}}
+
 {{% set configString = "p+i+n+u+g+s+b+acl+xattrs+sha512" %}}
 {{% for file in auditfiles %}}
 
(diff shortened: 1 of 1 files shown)


linux_os/guide/system/software/integrity/software-integrity/aide/aide_check_audit_tools/bash/shared.sh
```diff
@@ -18,6 +18,12 @@
 {{% set auditfiles = auditfiles + ["/usr/sbin/audispd"] %}}
 {{% endif %}}
 
+{{% if 'rhel' not in product %}}
+{{% set configString = 'p+i+n+u+g+s+b+acl+selinux+xattrs+sha512' %}}
+{{% else %}}
+{{% set configString = "p+i+n+u+g+s+b+acl+xattrs+sha512" %}}
+{{% endif %}}
+
 {{% set configString = "p+i+n+u+g+s+b+acl+xattrs+sha512" %}}
```
yuumasato, line 27: I think you'll need to remove definition of `configString` in line 27.

linux_os/guide/system/software/integrity/software-integrity/aide/aide_check_audit_tools/bash/shared.sh
```diff
@@ -18,6 +18,12 @@
 {{% set auditfiles = auditfiles + ["/usr/sbin/audispd"] %}}
 {{% endif %}}
 
+{{% if 'rhel' not in product %}}
```
yuumasato, line 21: I'm not sure how valid or expected this change is to all products.
Is your change in the context of SLE only? Or all products that are not RHEL?

If in context of SLE, suggest the following, as this keeps the `configString` the same for the rest of products.
```suggestion
{{% if 'sle' in product %}}
```

yuumasato (changes requested):

linux_os/guide/system/software/integrity/software-integrity/aide/aide_check_audit_tools/bash/shared.sh
anivan-suse: Done!

yuumasato (approved):
Thank you for the fix, the remediation is now aligned with the Ansible one.

PR #8217, opened 2022-02-14
Fix bug 1195533

anivan-suse:
#### Description:
Add bash remediation for rule: file_etc_security_opasswd

#### Rationale:
- Fixes # Bug 1195533: https://bugzilla.suse.com/show_bug.cgi?id=1195533

### DIFF
--- a/linux_os/guide/system/permissions/files/permissions_important_account_files/file_etc_security_opasswd/bash/shared.sh
+++ b/linux_os/guide/system/permissions/files/permissions_important_account_files/file_etc_security_opasswd/bash/shared.sh
@@ -1,0 +1,7 @@
+# platform = multi_platform_sle
+
+# Create /etc/security/opasswd if needed
+# Owner group mode root.root 0600
+[ -f  /etc/security/opasswd ] || touch /etc/security/opasswd
+chown root.root /etc/security/opasswd
+chmod 0600 /etc/security/opasswd
--- a/linux_os/guide/system/permissions/files/permissions_important_account_files/file_etc_security_opasswd/bash/shared.sh
+++ b/linux_os/guide/system/permissions/files/permissions_important_account_files/file_etc_security_opasswd/bash/shared.sh
@@ -1,7 +1,5 @@
-# platform = multi_platform_sle
+# platform = multi_platform_all
 
-# Create /etc/security/opasswd if needed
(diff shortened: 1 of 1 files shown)


linux_os/guide/system/permissions/files/permissions_important_account_files/file_etc_security_opasswd/bash/shared.sh
```diff
@@ -0,0 +1,7 @@
+# platform = multi_platform_sle
```
matejak, line 1: I suggest to leave `multi_platform_all` here - the applicability of the remediation is already restricted by the applicability of the rule. The fix doesn't contain anything that is SLE-specific and not part of the rule scope.

linux_os/guide/system/permissions/files/permissions_important_account_files/file_etc_security_opasswd/bash/shared.sh
```diff
@@ -0,0 +1,7 @@
...
+
+# Create /etc/security/opasswd if needed
+# Owner group mode root.root 0600
```
matejak, line 4: The remediation code explains itself quite clearly, so I suggest to get rid of those comments.

linux_os/guide/system/permissions/files/permissions_important_account_files/file_etc_security_opasswd/bash/shared.sh
```diff
@@ -0,0 +1,7 @@
...
+[ -f  /etc/security/opasswd ] || touch /etc/security/opasswd
+chown root.root /etc/security/opasswd
```
matejak, line 6: The user-group separator is `:` already for some time, and [usage of dot is discouraged](https://www.gnu.org/software/coreutils/manual/html_node/chown-invocation.html).

matejak (changes requested):
Thank you for your PR, there are just some minor issues that can be fixed easily, plus a suggestion to remove comments.

matejak:
Thank you for fixing the remediation, it looks good now, but please give us a test scenario as well.

brett060102:
This PR can be closed. The issue was addressed in PR https://github.com/ComplianceAsCode/content/pull/8552

PR #8219, opened 2022-02-14
Playbooks contain ansible.builtin requiring Ansible 2.9 or higher

dmc5179:
#### Description:

Playbooks now container ansible.builtin task definitions which require a newer version of Ansible. The ssg min version constant is currently set to 2.5.

#### Rationale:

With the addition of ansible.builtin task definitions, a version of Ansible newer than 2.5, the current ssg min default, is required. ansible.builtin requires Ansible version 2.9 or higher.


jan-cerny (approved):
This should be fine, because 2.9 is available even on RHEL7 and in fact this change only documents the actual state of the project.

PR #8236, opened 2022-02-16
Add rule for EKS CIS section 5.3

rhmdnd:
We can't automate anything in this section yet since the compliance
operator doesn't have the ability to query AWS endpoints. Until that
happens, let's at least provide the user with the manual steps they need
to take to audit and remediate.

#### Description:

#### Rationale:


Vincent056 (approved):

jan-cerny:
@rhmdnd Next time please remove the template text from the PR description and replace it by your own description and rationale.

PR #8237, opened 2022-02-16
Add rules for CIS EKS section 5.4

rhmdnd:
These controls don't have any automatic checks or rememdiations. If or
when the compliance-operator has the ability to query AWS endpoints, we
can consider updating these rules accordingly.

In the meantime, we can provide the user with manual instructions for
performing audits and remediations.

#### Description:

#### Rationale:


Vincent056 (approved):
I think the rules look good, maybe we could have a rule to check if `Calico` is installed, if that's the default go-to CNI

rhmdnd:
> I think the rules look good, maybe we could have a rule in the future to check if `Calico` is installed, if that's the default go-to CNI

Yeah - that's a good idea.

jan-cerny:
@rhmdnd @Vincent056 

>  I think the rules look good, maybe we could have a rule in the future to check if Calico is installed, if that's the default go-to CNI

Is this something you plan to add to this PR or will it be a separate PR?

Vincent056:
It looks like some rules are reusing `cce`, could you fix that?

applications/openshift/networking/configure_tls/rule.yml
```diff
@@ -0,0 +1,28 @@
...
+  Encrypt traffic to HTTPS load balancers using TLS certificates.
+
+rationale: |-
+  Encrypting traffic between users and your Kubernetes workload is fundamental
+  to protecting data sent over the web.
+
+severity: unknown
+
+identifiers:
+  cce@eks: CCE-88207-6
```
rhmdnd, line 17: This is duplicated from the above rule - replace it with  CCE-89133-3

rhmdnd:
> It looks like some rules are reusing `cce`, could you fix that?

Done - good catch.

PR #8240, opened 2022-02-16
Add rule for EKS CIS section 5.6

rhmdnd:
This control can't be automated, at least until the compliance operator
can communicate with AWS endpoints. Until then, we can at least supply
the user with the manual audit and remediation steps from the benchmark.


Vincent056 (commented):
No sure if it is the right question, I am just wondering if it is possible to check if Fargate is in use using Kube resources without accessing aws resources?

rhmdnd:
> No sure if it is the right question, I am just wondering if it is possible to check if Fargate is in use using Kube resources without accessing aws resources?

Potentially. I attempted to get CO running on Fargate in December, but hit some issues.

We could try it again since we have landed some changes in CO to use different labels.

jhrozek:
> > No sure if it is the right question, I am just wondering if it is possible to check if Fargate is in use using Kube resources without accessing aws resources?
> 
> Potentially. I attempted to get CO running on Fargate in December, but hit some issues.
> 
> We could try it again since we have landed some changes in CO to use different labels.

Should this PR go in as-is in the meantime?

rhmdnd:
> Should this PR go in as-is in the meantime?

At the very least we're providing the manual steps while we investigate ways to determine if we're running on Fargate. Worst case is we have to use the AWS client, which we'll need changes in CO to use.

jhrozek:
> > Should this PR go in as-is in the meantime?
> 
> At the very least we're providing the manual steps while we investigate ways to determine if we're running on Fargate. Worst case is we have to use the AWS client, which we'll need changes in CO to use.

I agree. Let me just try re-running the `ocp4` tests (I know the SSGTS test is "supposed" to fail)

Vincent056:
All the test passed

PR #8241, opened 2022-02-16
RHEL9 STIG Updates

Mab879:
#### Description:

Updates to the RHEL9 STIG control.

#### Rationale:

RHEL9 STIG


utils/create_srg_export.py
```diff
@@ -34,6 +34,209 @@
       'xccdf-1.1': ssg.constants.XCCDF11_NS}
 SEVERITY = {'low': 'CAT III', 'medium': 'CAT II', 'high': 'CAT I'}
 
+srgid_to_iacontrol = {
```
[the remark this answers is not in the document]
Mab879: It is, if we can somehow get this data into an external file that might help.

utils/create_srg_export.py
```diff
@@ -220,6 +444,17 @@ def get_rule_json(json_path: str) -> dict:
     return rule_json
 
 
+def add_variables(env_yaml, policy):
+    for control in policy.controls:
+        for rule in control.selections:
+            if rule.startswith('var_'):
+                parts = rule.split('=')
```
matejak, line 451: Selected item can be a rule, but here it is obvious that it can also be a variable or a refinement. Therefore, the name `rule` is misleading.
Next, you identify variables using the assumption that they start with `var_`, which is incorrect - there is e.g. the `sshd_approved_macs` variable.
However, you can use `ssg.build_yaml.SelectionHandler` and its `apply_selection` that identifies all variables for you.

utils/create_srg_export.py
```diff
@@ -220,6 +444,17 @@ def get_rule_json(json_path: str) -> dict:
...
+                if len(parts) != 2:
+                    raise ValueError('Invalid variable definition')
+                env_yaml[parts[0].replace('var_', '')] = parts[1]
```
matejak, line 454: Please use an explanatory variable here. And also double-check whether removal of `var_` is the right thing to do - why we want to do it in the first place?

matejak (changes requested):
I have requests regarding the Python code. I can't review content changes, so please ask somebody who has an idea about what is expected to help with that.

Mab879:
@jan-cerny Can please help review the content changes?

matejak (approved):

jan-cerny (approved):

ggbecker (approved):

PR #8246, opened 2022-02-21
Remove tmux process runinng check in configure_bashrc_exec_tmux

ggbecker:
#### Description:
- Remove tmux process runinng check.
  - This check can cause troubles since the user must be logged to show up
as tmux running. For example, an evaluation happening through a cron job
wouldn't be able to make this rule work, since no terminal is being
used.


vojtapolasek (approved):
LGTM

PR #8250, opened 2022-02-22
Introduce an "authselect enabled" rule

matejak:
The new rule features a check whether there is any authselect profile enabled. If it is not the case, it sets a profile according to an XCCDF value.

There are two test scenarios, the failing one is not remediated, as it is difficult to un-authselect a system - the authselect utility detects that there were previous modifications to other config files, and it won't apply the profile unless `--force` is used, which I wouldn't do in remediations.


ssg/build_yaml.py
```diff
@@ -1128,7 +1128,7 @@ def to_xml_element(self, env_yaml=None):
         regex = (r'(package_.*_(installed|removed))|' +
                  r'(service_.*_(enabled|disabled))|' +
                  r'install_smartcard_packages$')
-        priority_order = ["installed", "install_smartcard_packages", "removed",
+        priority_order = ["enable_authselect", "installed", "install_smartcard_packages", "removed",
```
mildas, line 1131: Is this ordering correct? Do we really want it before `installed` rules?

yuumasato: The priority ordering has effect within the group. Could it happen that `authselect` itself is not installed? 
This seems fine as the other package installed rule is `package_pam_pwquality_installed`.
https://github.com/ComplianceAsCode/content/tree/7ad2c970043370def2787a8ed773486384543821/linux_os/guide/system/accounts/accounts-pam

yuumasato: The group `accounts-restrictions` also has rules remediating via `authselect`.

So I think we should guarantee that group `accounts-pam` is placed before group `accounts-restriction`  in https://github.com/ComplianceAsCode/content/blob/7ad2c970043370def2787a8ed773486384543821/ssg/build_yaml.py#L1159

Or we move the rule `authselect_enabled` one group up, into the `accounts` group.

linux_os/guide/system/accounts/accounts-pam/enable_authselect/rule.yml
```diff
@@ -0,0 +1,32 @@
+documentation_complete: true
+
+prodtype: rhel7,rhel8,ol8,rhel9,fedora
```
marcusburghardt, line 3: Not necessary for rhel7.

linux_os/guide/system/accounts/accounts-pam/enable_authselect/tests/empty.fail.sh
```diff
@@ -0,0 +1,6 @@
+# platform = multi_platform_rhel,multi_platform_fedora
+# remediation = none
+
+rm -rf /etc/authselect/*
+touch /etc/authselect/authselect.conf
```
yuumasato, line 5: I manually installed a RHEL8 and chose minimal install.
The installed system didn't have file authselect configued and file `/etc/authselect/authselect.conf` did not exist.
image

Does it make sense to add a test where only `/etc/authselect/authselect.conf` is missing?

marcusburghardt: At least in my researches I didn't find a way to "disable" `authselect` once a profile is selected unless the symbolic links are manually removed. If this is the case, the system could still have a content in `/etc/authselect/authselect.conf` but the authselect profile are not actually in use.

I would recommend to assess if an authselect profile is selected by testing if the PAM files are symlinks to /etc/authselect

matejak: Done, plus the remediation now remediates with `--force` if the RPM check of the `pam` package discovers no changes to the package - it is able to detect changes to config files. So if there are no changes, there is no reason why not to force.

jan-cerny:
@matejak please rebase the PR on the latest upstream master so that the GitHub actions could execute the required CI job.
@vojtapolasek Is there anything else that needs to be done here?

vojtapolasek:
I think it looks good. One question: Do you plan to add Ansible remediation as well?

marcusburghardt:
> The issue could fix #8857.
> 
> I don't know how to write an Ansible remediation without an ugly shell code, so I would postpone this, as it can be done separately.

@matejak I can help with these Ansible remediation. I am actually working on the #8959 now and refactoring the bash macros for PAM. The next step will be the alignment between bash and Ansible macros, where I should work exclusively on Ansible improvements.

Finally, this PR plus the PR I am preparing for #8959 will fix the #8857 and help on the #8912.

linux_os/guide/system/accounts/enable_authselect/oval/shared.xml
```diff
@@ -0,0 +1,49 @@
...
+    comment="Check that /etc/authselect/authselect.conf is not empty"
+    id="test_authselect_conf_not_empty" version="1">
+  <ind:object object_ref="obj_authselect_conf_not_empty" />
+  </ind:textfilecontent54_test>
+
+  <ind:textfilecontent54_object id="obj_authselect_conf_not_empty" version="1">
+    <ind:filepath>/etc/authselect/authselect.conf</ind:filepath>
+    <ind:pattern operation="pattern match">^.+$</ind:pattern>
+    <ind:instance operation="greater than or equal" datatype="int">1</ind:instance>
+  </ind:textfilecontent54_object>
```
marcusburghardt, line 30: Is this test still necessary? It is not used anyways.

linux_os/guide/system/accounts/enable_authselect/oval/shared.xml
```diff
@@ -0,0 +1,49 @@
...
+    "password": "password-auth",
+    "postlogin": "postlogin",
+    "smartcard": "smartcard-auth",
+    "system": "system-auth",
+} -%}}
+
+<def-group>
+	<definition class="compliance" id="{{{ rule_id }}}" version="1">
+    {{{ oval_metadata("Check that authselect is enabled") }}}
+    <criteria operator="AND" comment="Check that /etc/authselect/authselect.conf is not empty">
```
marcusburghardt, line 12: This `comment` text is not coherent to the criteria.

marcusburghardt (changes requested):

marcusburghardt:
All tests passed locally, in VM and Container:
```
INFO - Logging into logs_bash/test_suite.log
INFO - xccdf_org.ssgproject.content_rule_enable_authselect
INFO - Script not_remediable.fail.sh using profile (all) OK
INFO - Script profile.pass.sh using profile (all) OK
INFO - Script remediable.fail.sh using profile (all) OK
```
```
INFO - Logging into /home/mburghardt/OpenSCAP/Forks/ComplianceAsCode/content/logs/rule-custom-2022-06-24-1742/test_suite.log
INFO - xccdf_org.ssgproject.content_rule_enable_authselect
INFO - Script not_remediable.fail.sh using profile (all) OK
INFO - Script profile.pass.sh using profile (all) OK
INFO - Script remediable.fail.sh using profile (all) OK
```

marcusburghardt (approved):
Thanks @matejak . This rule is pretty handy.

marcusburghardt:
> Code Climate has analyzed commit [969d456](https://github.com/ComplianceAsCode/content/commit/969d456c255d0225ab5c7ee4a865dff58f858865) and detected **11 issues** on this pull request.
> 
> Here's the issue category breakdown:
> 
> Category	Count
> Style	1
> Bug Risk	10
> The test coverage on the diff in this pull request is **100.0%** (**50%** is the threshold).
> 
> This pull request will bring the total coverage in the repository to **42.7%** (**0.0%** change).
> 
> View more on [Code Climate](https://codeclimate.com/github/ComplianceAsCode/content/pull/8250).

Checked. We probably have to adjust the CodeClimate filters.

PR #8253, opened 2022-02-22
Various bash fixes

matejak:
Mostly fix quoting and subshell invocations that captures output and so on.


jan-cerny:
The test scenarios pass locally

jan-cerny (approved):

PR #8254, opened 2022-02-22
Simplify generated augen bash expressions

matejak:
Certain checks can be performed at expansion time rather than at execution time.

The datastream diff is a good helper to see what this fix is about.


jan-cerny (approved):
The test scenarios pass locally, probably due to the container used in GH CI.

PR #8256, opened 2022-02-23
OCP4: Add control response for AC-12(1)

Vincent056:
Added control response for AC-12(1), section b is blocked by an RFE: https://issues.redhat.com/browse/RFE-2645

### DIFF
--- a/controls/nist_ocp4.yml
+++ b/controls/nist_ocp4.yml
@@ -1463,16 +1463,27 @@ controls:
   - high
   - moderate
 - id: AC-12(1)
-  status: pending
+  status: partial
   notes: |-
-    Section a: A control response is planned. Progress can be tracked via:
+    With any authentication mechanism the ability to revoke credentials if they
+    are compromised or no longer required, is a key control. Kubernetes client
+    certificate authentication does not allow for this due to a lack of support
+    for certificate revocation.
 
-    https://issues.redhat.com/browse/CMP-223
+    OpenShift's built-in OAuth server allows credential revocation by relying on
+    the Identity provider, as well as giving the administrators the ability to
+    revoke any tokens given to a specific user.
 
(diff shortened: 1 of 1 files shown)


controls/nist_ocp4.yml
```diff
@@ -1463,16 +1463,20 @@ controls:
...
 - id: AC-12(1)
-  status: pending
+  status: partial
   notes: |-
-    Section a: A control response is planned. Progress can be tracked via:
+    OpenShift provides a logout capability for user-initiated communications sessions whenever
+    authentication is used to gain access to the system when the default kubeadmin account is disabled.
+    
+    The OpenShift does not have the logout capability when cert is being used with the default kubeadmin account,
+    therefore we need to check if kubeadmin account is disabled.
```
jhrozek, line 1472: What do you think about using the text about certificate logins and revocation we have here: https://github.com/JAORMX/content/blob/be396cf36803f9890d6947d9688a74b3a9f535d7/applications/openshift/authentication/idp_is_configured/rule.yml#L42 ?

controls/nist_ocp4.yml
```diff
@@ -1463,16 +1463,20 @@ controls:
   - high
   - moderate
 - id: AC-12(1)
-  status: pending
+  status: partial
```
jhrozek, line 1466: This is marked as partial becuase there's an RFE right?

jhrozek (commented):
Looks good! One nit and one question inline

controls/nist_ocp4.yml
Vincent056: Thanks for the suggestion, I will update the content there

controls/nist_ocp4.yml
Vincent056: Yes, that's right

jhrozek (approved):
/lgtm

PR #8257, opened 2022-02-23
Introduce convenience macros for find and awk

matejak:
Macros enable easy creation of for loops using the [jinja2 call(...) syntax](https://jinja.palletsprojects.com/en/3.0.x/templates/#call)

Basically copied from https://github-wiki-see.page/m/koalaman/shellcheck/wiki/SC2044

### DIFF
--- a/shared/macros-bash.jinja
+++ b/shared/macros-bash.jinja
@@ -1683,6 +1683,7 @@ fi
 
 {{% endmacro %}}
 
+
 {{#
   This macro creates a bash conditional which is used to determine if a remediation is applicable.
   The macro takes package as an argument and chooses appropriate package manager.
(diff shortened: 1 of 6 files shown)


shared/macros-bash.jinja
```diff
@@ -1746,3 +1747,25 @@ else
...
+#}}
+{{% macro iterate_over_find_output(varname, find_args="") -%}}
+while IFS= read -r -d '' {{{ varname }}}; do
+    {{{ caller() | indent(4) }}}
+done <   <(find {{{ find_args }}} -print0)
+{{%- endmacro %}}
+
+
+{{#
+    Macro that lets you to define the body of a loop that iterates over the output of any command
```
jan-cerny, line 1764:
```suggestion
    Macro that lets you define the body of a loop that iterates over the output of any command
```

shared/macros-bash.jinja
```diff
@@ -1746,3 +1747,25 @@ else
    sed -i -e 's/\"'{{{ pol_setting }}}'.*/\"'{{{ pol_setting }}}'\": '{{{ pol_setting_val }}}',/g' {{{ chrome_pol_dir }}}{{{ chrome_pol_file }}}
 fi
 {{%- endmacro -%}}
+
+
+{{#
+    Macro that lets you to define the body of a loop that iterates over the output of the find command
```
jan-cerny, line 1753:
```suggestion
    Macro that lets you define the body of a loop that iterates over the output of the find command
```

jan-cerny (changes requested):

jan-cerny (approved):

PR #8261, opened 2022-02-23
Adjust partition sizes to acomodate a GUI install

yuumasato:
#### Description:

- Add 1536 MB to the `/usr` partition.
  This should also give it some room for future requirement increases.
- Reduce 512 MB from `/opt`, `/srv` and`/home` partitions.

#### Rationale:

- The GUI requires more room in the `/usr` partition:
  `At least 339MB more space needed on the /usr filesystem.`
- Let's reduce size of `/opt`, `/srv` and `/home`, which should not have any data right after install.


mildas (approved):
Tested on latest compose and the installation succeed. LGTM, merging

PR #8262, opened 2022-02-23
Ensure that get_implemented_stigs in utils/create_scap_delta_tailoring works in all cases

Mab879:
#### Description:

Fixes `get_implemented_stigs` in `utils/create_scap_delta_tailoring.py`

#### Rationale:

Before this commit using resolved_rules_dir would deselect all rules


ggbecker (approved):
Tested here and both options with and without --resolved-rules-dir produce the same amount of rules unselected.

PR #8264, opened 2022-02-24
NIST: AC-18(4) ensure wireless capability is turned disabled

Vincent056:
Add one rule to check if all the wireless kernels are disabled, added control response.


vojtapolasek (changes requested):
Hello, thank you for new rules. In the project, we try to athere to the "one rule = one config change" approach. I see that you created multiple templated rules and then one big rule which tries to enfroce multiple configuration changes.
Do you think you could go with the approach of using smaller templated rules only?

Vincent056:
> Hello, thank you for new rules. In the project, we try to athere to the "one rule = one config change" approach. I see that you created multiple templated rules and then one big rule which tries to enfroce multiple configuration changes. Do you think you could go with the approach of using smaller templated rules only?

Thanks for the review, that makes sense but I am wondering will I still need to separate them even they are just checking different wireless kernels? If that is the case, I can create one rule per wireless kernel, but that means remediation needs to be separated as well?

vojtapolasek:
So I checked the kernel_module_disabled template and it is unfortunately missing kubernetes remediation. I think that the best approach would be to:
1. create kubernetes remediation template for kernel_module_disabled. You can see example of already existing kubernetes platform here: https://github.com/ComplianceAsCode/content/blob/master/shared/templates/service_disabled/kubernetes.template
2. use only separate rules for each module. Utilize the template in these rules.
What do you think about it?

Vincent056:
> So I checked the kernel_module_disabled template and it is unfortunately missing kubernetes remediation. I think that the best approach would be to:
> 
>     1. create kubernetes remediation template for kernel_module_disabled. You can see example of already existing kubernetes platform here: https://github.com/ComplianceAsCode/content/blob/master/shared/templates/service_disabled/kubernetes.template
> 
>     2. use only separate rules for each module. Utilize the template in these rules.
>        What do you think about it?

Thanks for the suggestion, this config is a little tricky because of the URL encoding, but I think we might be able to create a template for it

So as for right now, there is one main rule `kernel_module_wireless_disabled`, this main rule only aggreate all the sub-rules, by extending their oval definitions, and this rule will only pass when all the sub-rule pass and there is only one remediation needed in the main rule.

If we are going to separate them(I guess I just need to remove the main rule), we would need to create one machine config remediation per module/sub-rule?

The sub-rule is there, we can always choose to use any sub-rules or just the main rule.

vojtapolasek:
Right. I suggest to remove the aggregating rule and create one rule per disabled kernel module. You can then use rules as needed.
As for urlencode, you can get inspired by other kubernetes templates. Actually, I have just noticed that there is a different template called coreos_kernel_option.
The example rule can be found here: https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/bootloader-grub2/coreos_pti_kernel_argument/rule.yml
Isn't this what you are looking for? If yes, you can easily create those small rules. If you decide to use the coreos template, please change also rule names appropriately.

Vincent056:
Thanks for the advice, I will separate them out

Vincent056:
> Right. I suggest to remove the aggregating rule and create one rule per disabled kernel module. You can then use rules as needed. As for urlencode, you can get inspired by other kubernetes templates. Actually, I have just noticed that there is a different template called coreos_kernel_option. The example rule can be found here: https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/bootloader-grub2/coreos_pti_kernel_argument/rule.yml Isn't this what you are looking for? If yes, you can easily create those small rules. If you decide to use the coreos template, please change also rule names appropriately.

Thanks for the advice, I will correct them

Vincent056:
> > Right. I suggest to remove the aggregating rule and create one rule per disabled kernel module. You can then use rules as needed. As for urlencode, you can get inspired by other kubernetes templates. Actually, I have just noticed that there is a different template called coreos_kernel_option. The example rule can be found here: https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/bootloader-grub2/coreos_pti_kernel_argument/rule.yml Isn't this what you are looking for? If yes, you can easily create those small rules. If you decide to use the coreos template, please change also rule names appropriately.
> 
> Thanks for the advice, I will correct them

@vojtapolasek Just updated the rule and added a template to generate remediation. Thanks again for the suggestion, it is really helpful!

jhrozek (approved):
/lgtm
the test failure seems to be expected because there are no tests (which is also expected)

Vincent056:
@vojtapolasek Could you help me to approve this PR? Thanks

linux_os/guide/system/network/network-wireless/wireless_software/kernel_module_cfg80211_disabled/rule.yml
```diff
@@ -0,0 +1,39 @@
+documentation_complete: true
+
+prodtype: fedora,rhcos4,rhel8,rhel9
+
+title: 'Disable Wireless Kernel cfg80211 Module'
+
+description: |-
+    The kernel's module loading system can be configured to prevent
```
vojtapolasek, line 8: The description states something different than the rule actually does. The rule (thanks to the platform) disables only one module - cfg80211.
I suggest you use this macro:
    {{{ describe_module_disable(module="cfg80211" }}}
You can review the macro here: https://github.com/ComplianceAsCode/content/blob/master/shared/macros.jinja#L528
Please fix this for all rules you create.

linux_os/guide/system/network/network-wireless/wireless_software/kernel_module_cfg80211_disabled/rule.yml
```diff
@@ -0,0 +1,39 @@
+documentation_complete: true
+
+prodtype: fedora,rhcos4,rhel8,rhel9
+
+title: 'Disable Wireless Kernel cfg80211 Module'
```
vojtapolasek, line 5: _ don't think that mentioning "wireless" makes sense here. Please fix titles for all rules you create._

vojtapolasek (changes requested):
Hello, the PR still needs some changes. See comments.
There are currently no tests for your rules. I mean no tests which could be used by our SSG Test suite.
Could you please add tests to each rule? You can get inspired by this:
https://github.com/ComplianceAsCode/content/tree/master/linux_os/guide/system/permissions/mounting/kernel_module_cramfs_disabled/tests
Of course the best would be to include tests for the template, but this is out of scope of this PR I believe.
Thank you.

controls/nist_rhcos4.yml
```diff
@@ -1759,6 +1759,10 @@ controls:
   rules:
   - wireless_disable_in_bios
   - kernel_module_bluetooth_disabled
+  - kernel_module_wireless_iwlmvm_disabled
+  - kernel_module_wireless_iwlwifi_disabled
```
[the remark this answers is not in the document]
Vincent056: That's a good catch, I remember I had to remove the wireless, because I used the yaml template, and the template's oval definition name does not include wireless, I forgot to remove wireless from here

vojtapolasek (approved):

vojtapolasek:
Thank you very much. Merging.

PR #8269, opened 2022-02-24
UEFI booting requires FAT support.

brett060102:
Rule kernel_module_vfat_disabled, disables the vfat drivers in the kernel,
but SUSE's UEFI boot requires FAT support. See:
https://documentation.suse.com/sles/15-SP2/html/SLES-all/cha-uefi.html

Applying this rule on a SUSE system, results in a system that fails
to boot.

#### Description:
-- Remove sle15 from kernel_module_vfat_disabled/

Rulle:
linux_os/guide/system/permissions/mounting/kernel_module_vfat_disabled/rule.yml

which is part of the cis profile should not include suse.
While testing the cis.profile remediation on SLES15, we found that disabling the vfat in the kernel causes the system to fail to come up after reboot.

This because SUSE is EUFI based and it requires FAT support:
https://documentation.suse.com/sles/15-SP1/html/SLES-all/cha-uefi.html

After removing the cis@sle15 tag and sle15 from the prodtype. The system was able to be reboot after running cis remediation.

#### Rationale:

- System should be able to reboot system after remediation.


vojtapolasek (changes requested):
Hello,
I understand your goal, however, I think the better way to accomplish it is to remove the rule from the profile e.g. controls/cis_sle15.profile. You can also remove the prodtype if you see it fit, but you will remove an option for users of sle15 datastreams to use this rule in tailoring.
Also I suggest that you add explanation of this chance into the cis control file.

brett060102:
@vojtapolasek Thank you. I will make the update to controls/cis_sle_15.yml did not know about that file at all. I am not sure what you mean by "suggest that you add explanation of this" in cis control file. I look at the other files in that directory and I don't see and example explanations. I do think we need to fully disable the rule for SUSE, since while we support legacy BIOS doing it on UEFI systems will brick the system. It would be best to does does not apply if UEFI, but that is beyond my knowledge.

vojtapolasek:
Hello, you can check out controls/cis_rhel8.yml, line 38 to seee the example. Read more about format of control files here:
https://complianceascode.readthedocs.io/en/latest/manual/developer/03_creating_content.html#controls
As for removing the prodtype, I suggest you leave the prodtype and the CIS reference in tact. There might be user who know that the system is running on BIOS and they might want to include the rule for example through tailoring.

brett060102:
@vojtapolasek Thank you. I think I have it now. I was surprised that tagging the rule as manual did not disable remediation. Also changes the use of automated: to status: since  automated: has been deprecated. Also, made the cis id match what was in the control file. Thank you for walking me through this.

vojtapolasek:
Hello, I was happy to help. Could you please also change the PR description and possibly the title? Cause now the change is much bigger than described.

brett060102:
@vojtapolasek done. Do you want me to squash the changes?

vojtapolasek (approved):

vojtapolasek:
Yes please, squash it. Thank you very much, that will be it.

brett060102:
@vojtapolasek squashed.

PR #8271, opened 2022-02-25
Add templates to handle detection of most Firefox Policy STIG items. (WIP)

lenox-joseph:
#### Description:

Implement detection OVAL and remediation scripts to cover several manual STIGs for Mozilla Firefox's administrative policy.

#### Rationale:

Firefox uses a more complex JSON structure than Chromium does (at least for currently implemented items), and I'd like to be able to extend automagical remediations to cover the manual DISA STIG (especially as the automatic upstream SCAP benchmark has gone poof).


lenox-joseph:
Failure is because the test system is trying to find the rules in the wrong prodtype, no idea how to address this in the build system.

lenox-joseph:
@Mab879 @ggbecker  any availability to do a review?

Mab879 (changes requested):
Thanks for the contribution. I did find a couple issues when running this on a Fedora 35 box. Let me know if you need any information.

```Title   Set Firefox Configuration File Location
Rule    xccdf_org.ssgproject.content_rule_firefox_preferences-lock_settings_config_file
Result  error

Title   Disable Firefox Configuration File ROT-13 Encoding
Rule    xccdf_org.ssgproject.content_rule_firefox_preferences-lock_settings_obscure
Result  error
````

**Comment by lenox-joseph** [comment]
> Title   Set Firefox Configuration File Location
> Rule    xccdf_org.ssgproject.content_rule_firefox_preferences-lock_settings_config_file
> Result  error
> 
> Title   Disable Firefox Configuration File ROT-13 Encoding
> Rule    xccdf_org.ssgproject.content_rule_firefox_preferences-lock_settings_obscure
> Result  error

I didn't touch anything anywhere near these rules (both are self-contained), so I have no idea what could have broken on your test system.

**Comment by lenox-joseph** [comment]
Thank you for the review @Mab879

File: products/firefox/guide/firefox/firefox_policy-search_suggestion/rule.yml [hunk 1]
```diff
@@ -0,0 +1,25 @@
+documentation_complete: true
+
+prodtype: firefox
+
+title: 'Firefox search suggestions must be disabled.'
+
+description: |-
+    Private browsing may be disabled in an administrative policy by setting
```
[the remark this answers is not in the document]
lenox-joseph: Fixed.

products/firefox/guide/firefox/firefox_policy-extension_update/rule.yml
```diff
@@ -0,0 +1,25 @@
...
+prodtype: firefox
+
+title: 'Firefox must be configured to not automatically update installed add-ons and plugins.'
+
+description: |-
+    The update check may be disabled in an administrative policy by setting
+    the <tt>ExtensionUpdate</tt> key under <tt>policies</tt> to <tt>false</tt>.
+
+rationale: |-
+    To protect privacy and sensitive data, Firefox provides the ability to configure the program so that data entered into forms is not saved. This mitigates the risk of a website gleaning private information from prefilled information.
```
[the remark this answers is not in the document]
lenox-joseph: Things I get for doing 20 rules at once -.-;

products/firefox/guide/firefox/firefox_policy-encrypted_media_extensions/rule.yml
```diff
@@ -0,0 +1,35 @@
...
+
+description: |-
+    Feedback reporting feature may be disabled via administrative policy by setting
+    <ul>
+    <li><tt>Enabled</tt> under <tt>EncryptedMediaExtensions</tt> to <tt>false</tt></li>
+    <li><tt>Locked</tt> under <tt>EncryptedMediaExtensions</tt> to <tt>true</tt></li>
+    </ul>
+
+rationale: |-
+    Enable or disable Encrypted Media Extensions and optionally lock it.
```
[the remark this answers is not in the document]
lenox-joseph: Done.

lenox-joseph:
@Mab879 should be good to look at again. I didn't realize that someone had extended the diff system to allow testing of the firefox datastream now.

lenox-joseph:
@Mab879  I'd also like to get this into remediation sets for work, although I understand if there's extra trepidation at adding these remediations (at this point I don't think it's feasible to add Ansible remediations without writing up a module to go with it).

lenox-joseph:
This failure seems like something in master as I don't have any Kubernetes remediations in this PR, especially since there weren't any failures before I rebased.

Mab879:
The content looks good.

@matejak can you take a look the remediation script?

lenox-joseph:
@matejak  those failed tests look like something in the baseline as I don't have any Kubernetes remediation implemented.

lenox-joseph:
@matejak  FYI, new commits are because there were some revisions from the upstream DISA changes.

shared/templates/firefox_policy-setting/bash.template
```diff
@@ -0,0 +1,58 @@
...
+   pass
+else:
+   _tree{{{ policy_item.path_python[loop.revindex0] }}} = dict()
+{{% endfor %}}
+_tree{{{ policy_item.path_python[0] }}}['{{{ policy_item.parameter }}}'] = {{{ policy_item.value_escaped }}}
+{{% endfor %}}
+_file=open('${firefox_dir}/${firefox_cfg}', 'w')
+json.dump(_tree, _file, indent=4, sort_keys=True)
+_file.close()
+""" | ${__REMEDIATE_PYTHON}
```
matejak, line 44: You use an interesting approach here that you call Python from a Bash remediation. This is not a blocker, but it would be probably better to have an all-Python remediation format that would get translated into Bash by the build system. Anyway, the Bash way of remediating this would [probably involve](https://stackoverflow.com/questions/24942875/change-json-file-by-bash-script) the `jq` tool - do you have any thoughts about this?

lenox-joseph:
I agonized over JSON when doing the initial implementation. If FFOX hadn't put JSON only policy configurations in place, I would have been happy to just do the preference file.

I didn't have a pure bash implementation available to me that actually let me set fields (and I looked) and jq is not installed "standard". I don't think it is even in the standard repos for RHEL7 (the platform I must support and the point of me making this PR).

There are some existing remediations for Chrome, but those assume all of the attributes are right at the top of the file (the simplest case), which quickly proved to be useless to me.

If someone wants to write/modify a pure shell implementation on their own time (there are a few to start from) that can in-place modify an existing file, I would appreciate it, but I could not justify the cost.

Python is near guaranteed to be installed on Linux across the distros one would also have Firefox on *and* evaluating the DISA STIG. That's why I try to be 2/3 compliant and only use the standard library.

What I *really* want is for OVAL spec gets a native JSON handler and on our side we had better assurances about handling JSON. The OVAL regex I have only really works for these configurations.

IMHO it's a cursed format and I wish that developers would stop using it for configuration files at rest. Or failing that, the standard tooling would get better.
________________________________
From: Matěj Týč ***@***.***>
Sent: Friday, May 13, 2022 9:39:32 AM
To: ComplianceAsCode/content ***@***.***>
Cc: Lenox, Joseph D Collins ***@***.***>; Mention ***@***.***>
Subject: [External] Re: [ComplianceAsCode/content] Add templates to handle detection of most Firefox Policy STIG items. (WIP) (PR #8271)


@matejak commented on this pull request.

________________________________

In shared/templates/firefox_policy-setting/bash.template<https://urldefense.com/v3/__https://github.com/ComplianceAsCode/content/pull/8271*discussion_r872473753__;Iw!!MvWE!RMYIzsRtx22PUwhaaiKhKspXzneaxBHFDxo2gNhNTbY4dDzYrN6qmEHZlpK0mVJl5NM$>:

> +_file=open('${firefox_dir}/${firefox_cfg}', 'r')
+_tree=json.load(_file)
+_file.close()
+{{% for policy_item in POLICIES %}}
+{{% for p in policy_item.subpath %}}
+if '{{{ p }}}' in _tree{{{ policy_item.path_python[loop.revindex] }}}:
+   pass
+else:
+   _tree{{{ policy_item.path_python[loop.revindex0] }}} = dict()
+{{% endfor %}}
+_tree{{{ policy_item.path_python[0] }}}['{{{ policy_item.parameter }}}'] = {{{ policy_item.value_escaped }}}
+{{% endfor %}}
+_file=open('${firefox_dir}/${firefox_cfg}', 'w')
+json.dump(_tree, _file, indent=4, sort_keys=True)
+_file.close()
+""" | ${__REMEDIATE_PYTHON}


You use an interesting approach here that you call Python from a Bash remediation. This is not a blocker, but it would be probably better to have an all-Python remediation format that would get translated into Bash by the build system. Anyway, the Bash way of remediating this would probably involve<https://urldefense.com/v3/__https://stackoverflow.com/questions/24942875/change-json-file-by-bash-script__;!!MvWE!RMYIzsRtx22PUwhaaiKhKspXzneaxBHFDxo2gNhNTbY4dDzYrN6qmEHZlpK04-pL6K8$> the jq tool - do you have any thoughts about this?

shared/templates/firefox_policy-setting/tests/correct_value.pass.sh
```diff
@@ -0,0 +1,32 @@
...
+{{{ find_python() }}}
+
+# Iterate over the possible Firefox install directories
+for firefox_dir in ${firefox_dirs}; do
+    # write our bad config in every location to ensure remediation fixes it.
+    mkdir -p ${firefox_dir}
+    echo "{ \"policies\" :{ } }" > ${firefox_dir}/${firefox_cfg}
+    echo """
+import json
+_file=open('${firefox_dir}/${firefox_cfg}', 'r')
```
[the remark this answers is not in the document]
lenox-joseph: Since I was using a Python snippet anyway, it made sense at thee time to re-use that embedded Python script so that I knew _a priori_ that it would be valid. 

Reimplementing with "just `bash` and `jinja`" means that all of the python here gets pushed up to the template process function and the entirety of the file write gets done as a block `cat << EOF  > ${firefox_dir}/${firefox_cfg} ... stuff EOF` and the resulting JSON is still done in Python.

Since it's just tests, if we had buy-in from the people maintaining all the build/test infrastructure to ensure `jq` is available and in PATH (and amended the docs so that other poor souls trying to run tests on their own), these could leverage `jq`.

shared/templates/firefox_policy-setting/bash.template
lenox-joseph: I agonized over JSON when doing the initial implementation. If FFOX hadn't put JSON only policy configurations in place, I would have been happy to just implement the new manual STIG items in the preference file.

I didn't have a pure bash implementation available to me that actually let me set fields (and I looked) and jq is not installed "standard". I don't think it is even in the standard repos for RHEL7 (the platform I must support and the point of me making this PR).

There are some existing remediation scripts for Chrome, but those assume all of the attributes are right at the top of the file (the simplest case), which quickly proved to be useless to me.

If someone wants to write/modify a pure shell implementation on their own time (there are a few to start from) that can in-place modify an existing file, I would appreciate it, but I could not justify the cost.

Python is near guaranteed to be installed on Linux across the distros one would also have Firefox on *and* evaluating the DISA STIG. That's why I try to be 2/3 compliant and only use the standard library.

What I *really* want is for OVAL spec gets a native JSON handler and on our side we had better assurances about handling JSON. The OVAL regex I have only really works for these configurations.

IMHO it's a cursed format and I wish that developers would stop using it for configuration files at rest. Or failing that, the standard tooling would get better.

lenox-joseph: As for "Python remediation format that gets translated to Bash by the build system", that's way more complexity than I was willing to bite off for this effort, as the Bash-form of the remediation _still_ has to interpret any existing files and modify with a 100% accuracy rate. Hacking things together with regex is a kludge I reserve for when I have zero other choice (such as for our OVAL checker).

shared/templates/firefox_policy-setting/tests/correct_value.pass.sh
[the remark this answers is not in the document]
lenox-joseph: Yeah, I went and triple checked and `jq` is an EPEL package, which makes it unobtanium in my use case.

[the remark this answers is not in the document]
lenox-joseph: So I spent some time pushing the actual test output generation up into `template.py` and that work actually makes the test harder to debug/review, because of the way the templates interact with tests, especially in a container.

I had to stick a `print(filled_templates)` statement into `ssg/templates.py` under `get_all_tests()` so that I could actually see and verify my text output and realize the refactoring mistake I had made. It isn't pretty *at all*.

shared/templates/firefox_policy-setting/template.py
```diff
@@ -0,0 +1,137 @@
...
+        if p == "policies":
+            _result = ''.join([r'"{0}"[\s]*:[\s]*'.format(p),
+                               r"\{.*[\s]*",
+                               _result,
+                               r"[^}]*\}"])
+        else:
+            _result = ''.join([r'"{0}"[\s]*:[\s]*'.format(p),
+                               r"(\{?(\{[^}]*\}?)*|[^}]*)[\s]*",
+                               _result,
+                               r"[^}]*\}"])
```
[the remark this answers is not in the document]
lenox-joseph: After the refactor and removal of unused methods (forgotten about when I gave up on trying to do pure `bash` nonsense), it's actually only the one function for OVAL.

shared/templates/firefox_policy-setting/tests/missing_value.fail.sh
```diff
@@ -0,0 +1,35 @@
...
+
+firefox_cfg="policies.json"
+firefox_dirs="/usr/lib/firefox/distribution /usr/lib64/firefox/distribution /usr/local/lib/firefox/distribution /usr/local/lib64/firefox/distribution"
+{{{ find_python() }}}
+
+# Iterate over the possible Firefox install directories
+for firefox_dir in ${firefox_dirs}; do
+    # write our bad config in every location to ensure remediation fixes it.
+    mkdir -p ${firefox_dir}
+    echo "{ \"policies\" :{ } }" > ${firefox_dir}/${firefox_cfg}
```
matejak, line 11:
```suggestion
    echo '{ "policies" :{ } }' > "${firefox_dir}/${firefox_cfg}"
```

I propose to reduce escaping as much as possible.

shared/templates/firefox_policy-setting/tests/correct_value.pass.sh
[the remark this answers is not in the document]
lenox-joseph: A here-document doesn't help when I still have to *write the here-document*, and that's non-trivial when I'm working *in the template*. I did write Python code to generate that here-document, but there's nowhere I can put that other than the Python template code. 

```python
def build_test_json(policy, _test, missing=False, wrong=False):
    """Generate a minimal JSON configuration for this particular policy object, based 
       on a correct (set to value), missing, or wrong value. 
       Returns a dict() containing this test entry.
       
       Parameters:
           policy: current policy item to generate for
           _test: dict() that has at least the value {"policies": {} }
    """
    _current = _test.get("policies")
    _key =  policy["parameter"]
    _value = make_json_boolean(policy["value"])

    for p in policy["subpath"]:
        if p not in _current:
           _current[p] = dict()
        _current = _current.get(p)
    
    if wrong:
        _current[_key] = "VERYVERYBADVALUE"
    elif not missing:
        _current[_key] = _value
    return _test
```

What I don't want is to spread the tests of this template across 10-20 individual rules; duplicating effort and separating the test from the implementing code. 
Otherwise I wouldn't have bothered with the template (which was the approach of the original writer of the lock parameter rules) in the first place.

This is the same approach (to keep the tests of the template with the template) as I took for a template for the [firefox_lockparameter](https://github.com/ComplianceAsCode/content/tree/master/shared/templates/firefox_lockparameter/tests) template and is also done in other places like `audit_rules_unsuccessful_file_modification`.

shared/templates/firefox_policy-setting/tests/missing_value.fail.sh
```diff
@@ -0,0 +1,35 @@
...
+# Remove the parameter
+try:
+    del _tree{{{ policy_item.path_python[0] }}}['{{{ policy_item.parameter }}}']
+except KeyError:
+    pass
+{{% endfor %}}
+_file=open('${firefox_dir}/${firefox_cfg}', 'w')
+json.dump(_tree, _file, indent=4, sort_keys=True)
+_file.close()
+""" | ${__REMEDIATE_PYTHON}
```
matejak, line 34:
```suggestion
{{% set json_lines = [] %}}
{{% for policy_item in POLICIES %}}
{{% for p in policy_item.subpath %}}
{{# the 'or ""' prevents None to be returned as the result of append #}}
{{{ json_lines.append('"p" = "always wrong value"') or "" }}}
{{% endfor %}} 
{{% endfor %}} 
cat > "${firefox_dir}/${firefox_cfg}" << EOF
{
{{{ ",\n".join(json_lines) }}}
}
EOF
```

shared/templates/firefox_policy-setting/bash.template
matejak: Thanks for the comment, it really clarifies that the Python approach is a right pick.

shared/templates/firefox_policy-setting/template.py
```diff
@@ -0,0 +1,131 @@
...
+
+def normalize_policy_path(path_string):
+    """Generate a normalized path array from yaml.
+       Paths are space-separated key names.
+       All firefox policies are under the policies key.
+    """
+    path = []
+    if len(path_string.strip()) > 0:
+        path = re.split(" ", path_string)
+    if 'policies' not in path:
```
matejak, line 14: Shouldn't it be `if path[0] != 'policies'`?

shared/templates/firefox_policy-setting/template.py
```diff
@@ -0,0 +1,131 @@
...
+        path = ['policies'] + path
+    return path
+
+
+def make_json_boolean(instr, oval_format=False):
```
matejak, line 19: Those are basically two different functions, one when `oval_format` is true, and the other one when it is false. So please make it into two functions.

Next, argument `instr` is not descriptive, and it is also not clear what should be the output - one code path returns actual boolean, one returns a regular expression, and other return a string. So the function name is also questionable. Please try to make also the function name descriptive.

shared/templates/firefox_policy-setting/template.py
lenox-joseph: I consider them equivalent and just naturally went to the "if not in X". Switching to the other form is fine.

shared/templates/firefox_policy-setting/template.py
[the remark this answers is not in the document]
matejak: Still, there are regexes s.a. `"{0}"[\s]*:[\s]*` and `[^}]*\}` that require about a minute to be deciphered, so please assign them to explanatory variables.

shared/templates/firefox_policy-setting/template.py
lenox-joseph: I must have misunderstood [you before](https://github.com/ComplianceAsCode/content/pull/8271#discussion_r872481374) then because I *did* have those two functions for rendering JSON booleans separate.

shared/templates/firefox_policy-setting/template.py
```diff
@@ -0,0 +1,131 @@
...
+                               r"[^}]*\}"])
+        else:
+            _result = ''.join([r'"{0}"[\s]*:[\s]*'.format(p),
+                               r"(\{?(\{[^}]*\}?)*|[^}]*)[\s]*",
+                               _result,
+                               r"[^}]*\}"])
+    return _result
+
+
+def build_python_json_notation(_path, tier=0):
```
matejak, line 48: I have no idea what this function is supposed to do. What is tier? And what is python-specific about the functionality?

shared/templates/firefox_policy-setting/tests/correct_value.pass.sh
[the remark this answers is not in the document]
matejak: I think that this is a step in a wrong direction. Previously, you had the Python code in the test scenario template, and  I think that you can actually generate the json file directly and easily at the same time. It also feels better to have different code for test scenario setup and the remediation.

Our test suite lacks a possibility to view a test scenario code after its jinja-expansion, which makes this quite annoying. I will look into it though.

shared/templates/firefox_policy-setting/template.py
lenox-joseph: When generating the inlined python script for remediation, I need to have the call notation written generated (where it's used in the template as `path_python`). Tier is the distance from the top of the document.
Adding a docstring now.

shared/templates/firefox_policy-setting/template.py
matejak: Probably, "split" means that one should end up with more functions. Generally, this is a good direction.

matejak:
The PR touches multiple aspects of the template and test capabilities of the project, so please be patient - it will take a few more iterations to get everything through.

shared/templates/firefox_policy-setting/tests/correct_value.pass.sh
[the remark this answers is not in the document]
lenox-joseph: This code *does* generate the json and makes it a string that is then used for a here-document. Unless you want to make `jq` a hard dependency to run these tests, there's not a solution that is going to work other than either what I did here (generating the json during template instantiation/preprocessing), generating bespoke JSON for every rule that uses the template and placing those with the rule documents, or going down the path of madness by trying to assemble valid JSON with just standard POSIX tools. 

Personally, I think that either pushing the "here-document generation" step into the preprocess or leveraging the mechanism the remediation used is the less complex manner that doesn't give up the benefits of having the template in the first place.

Using `jq` is not something I can decide, given it's non-existence in RHEL standard repos. 

Generating bespoke JSON is a route I am going to resist as it introduces more complexity to implementing rules for this STIG in the future (and more places to screw up). 

The path of madness is right out, as I have already spent far too much effort trying to do that (which led to using Python as a stand-in for `jq` in the first place).

>Our test suite lacks a possibility to view a test scenario code after its jinja-expansion, which makes this quite annoying. I will look into it though.

When I was checking this code I had stuck a print statement for the expanded templates in that function so that I could tell what was going on (because I was running in a container, which further obfuscated things). Being able to dump those expanded scripts to the log directory would be extremely useful.

shared/templates/firefox_policy-setting/template.py
lenox-joseph: https://github.com/ComplianceAsCode/content/pull/8271/commits/cb89dc72a97d03ab77febe9e2552d8fee049bf7f

shared/templates/firefox_policy-setting/template.py
lenox-joseph: https://github.com/ComplianceAsCode/content/pull/8271/commits/879ea5ffb21522c9f3a1b64782d78673eda36c6e

shared/templates/firefox_policy-setting/template.py
```diff
@@ -0,0 +1,137 @@
...
+    return "\'{0}\'".format(instr)
+
+
+def make_json_boolean_oval(instr):
+    if instr in ["false", "true"]:
+        return instr
+    return "\"{0}\"".format(instr)
+
+
+def parameter_find_oval(_path, parameter, value):
```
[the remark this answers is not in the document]
lenox-joseph: I think I've cleared up the actual issue in this comment; there was cruft left over from when I was trying to not use Python and I've cleared all of that old code out.

shared/templates/firefox_policy-setting/template.py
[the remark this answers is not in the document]
lenox-joseph: Replaced with variables that hopefully are understandable to someone other than me, and I collapsed the branches so there's less duplication and I think it's clearer what is going on (changing the match to be a lot less forgiving when I'm not trying to match from the root node).

lenox-joseph:
Rebase because some of the test hacks I was using that are unique to my dev environment were accidentally leaked into the branch.

lenox-joseph:
DISA changed things again on me, so I've gone through the entire STIG and will close this PR and submit a new branch.

PR #8273, opened 2022-02-25
Generate srgmap tables using Jinja

jan-cerny:
This commit changes the generator of table-product-srgmap.html and
table-product-srgmap-flat.html files. The current generator implemented
by XSLT is replaced by a Jinja template and a Python script
utils/gen_srg_table.py. Instead of processing XCCDF and OCIL XML files
we will load the data from compiled rules and compiled profiles.  This
change save us approximately 5 seconds during the build of a single
product.  The aim is to generate almost the same table as we generate
currently.

We remove some of the parameters of the CMake macro
ssg_build_html_srgmap_tables because we always call this macro with the
same parameter values. We won't generate xccdf-linked-srg-overlay.xml
because it's an interim file which isn't used or shipped anywhere.


jan-cerny:
rebased on master

utils/gen_srg_table.py
```diff
@@ -0,0 +1,90 @@
...
+
+Srg = collections.namedtuple(
+    "SRG", [
+        "cci", "requirement", "vuln_discussion", "check_content", "fixtext"])
+
+
+def parse_srgs(srgs_filename):
+    srgs = dict()
+    root = ET.parse(srgs_filename)
+    for rule in root.findall(".//xccdf-1.1:Rule", NS):
```
Mab879, line 22: `utils/create_srg_export.py` has some code that looks exactly like this. Should we move it to `ssg` and share?

utils/tables/srgmap_flat_template.html
```diff
@@ -0,0 +1,74 @@
+{{% set title = "SRGs from General Purpose Operating System Security Requirements Guide Mapped to DISA STIG for " ~ full_name -%}}
+
+{{% extends "tables_template_common.html" %}}
+
+{{% block table -%}}
+<table>
```
Mab879, line 6: This is almost exactly what `utils/create_srg_export.py` produces as a CSV. Is it worth trying to combine these outputs?

utils/tables/srgmap_flat_template.html
```diff
@@ -0,0 +1,74 @@
...
+    <td>{{{ rule.identifiers["cce"] }}}: {{{ rule.title }}}</td>
+{{% else %}}
+    <td>{{{ rule.title }}}</td>
+{{% endif %}}
+    <td>{{{ srg.vuln_discussion }}}</td>
+    <td>{{{ rule.description }}}</td>
+    <td>Applicable - Configurable</td>
+    <td>{{{ srg.check_content }}}</td>
+    <td>{{{ rule.ocil }}} Is it the case that {{{ rule.ocil_clause }}}?</td>
+    <td>{{{ srg.fixtext }}}</td>
```
Mab879, line 44: Some rules have `fix` on them now, should we try to use that?

utils/gen_srg_table.py
```diff
@@ -0,0 +1,90 @@
...
+            check_content=check_content, fixtext=fixtext)
+    return srgs
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description="Create a SRG map and a SRG flat map, both at once")
+    parser.add_argument(
+        "--build-dir", default="build", help="Path to the build directory")
+    parser.add_argument(
```
Mab879, line 49: Add a short argument `-p` for this?

Mab879 (commented):
Overall, I think moving to this method for is a great idea. However, I think we should looking centralizing some of this code in the `ssg` module.

utils/gen_srg_table.py
jan-cerny: great, thanks, done it the latest commit

utils/tables/srgmap_flat_template.html
jan-cerny: Can you explain this more?

Mab879: So the flat SRG table looks very similar to the output of `utils/create_srg_export.py`. They have the same columns, however `utils/create_srg_export.py` does include more statuses and includes more fields that are always blank in the srg mapping tables.

jan-cerny: Great idea. My problem is that I actually don't understand the tables. So I only tried to create a tool that generates the same tables as the tables that we generate currently by XSLTs. But the current tables are confusing. Why are there empty columns? Why are columns with same values on each row? Why the headers don't correspond the content? I wonder if anyone has answers to this.

Mab879: So the table columns are from the STIG spreadsheet. 

> Why are there empty columns?

Depending on the status column, some columns might be blank. For example if the status is "Applicable - Configurable", the last three columns do not need to be filled out.

> Why are columns with same values on each row?

If there is a CaC rule for a given SRGID it almost always "Applicable - Configurable".

> Why the headers don't correspond the content?

Not sure on this one.

matejak: What about merging the PR if it works, and do the unification of the codebase with `utils/create_srg_export.py` as a second step? The PR is quite big already.

vojtapolasek:
@ggbecker  @Mab879  could you please state if the resulting tables are OK? See @jan-cerny  comments above. Are we OK with blank / repeating column content?

jan-cerny:
I have rebased on the current master.

I have replaced the most eye beating thing which is the V number in CCI column by CCI.

But, let me rephrase the original question:

Should the tables be the same as until now or should they be different? If they should be different, what items should be different?

matejak:
@Mab879 What about Honza's question? If the tables are good enough, then let's merge them, otherwise let's try to find the shortest path to that "good enough".

Mab879:
> I have rebased on the current master.
> 
> I have replaced the most eye beating thing which is the V number in CCI column by CCI.
> 
> But, let me rephrase the original question:
> 
> Should the tables be the same as until now or should they be different? If they should be different, what items should be different?
So I that these tables are okay. If more distros use the SRG GPOS control then we should try to use the create_srg_export.

jan-cerny:
I have rebased this on the upstream master.

matejak:
LGTM, the tables are basically identical except occasional fixes.

The code coverage got slightly down because of new code introduced to the ssg module, but that's acceptable for the time being.
Thank you for your great PR!

PR #8278, opened 2022-02-28
Update OVAL content for grub2_admin_username

Xeicker:
#### Description:

- Remove the OVAL criterion that allow the absence of  grub.cfg file

#### Rationale:

- DISA's STIG profile expects to have a superuser defined with certain characteristics, from there can be inferred that the file is needed so it defines the superuser correctly

### DIFF
--- a/linux_os/guide/system/bootloader-grub2/non-uefi/grub2_admin_username/oval/shared.xml
+++ b/linux_os/guide/system/bootloader-grub2/non-uefi/grub2_admin_username/oval/shared.xml
@@ -4,9 +4,6 @@
               "The grub2 boot loader superuser should have a username that is hard to guess.") }}}
 
     <criteria operator="OR">
-      {{% if product not in ["ol8", "rhel8"] %}}
-      {{{ oval_file_absent_criterion(grub2_boot_path + "/grub.cfg") }}}
-      {{% endif %}}
       <criterion comment="Superuser is defined in {{{ grub2_boot_path }}}/grub.cfg and it isn't
             root, admin, administrator nor equal to any system username"
             test_ref="test_bootloader_superuser_differ_from_other_users"/>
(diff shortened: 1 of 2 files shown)


jan-cerny:
The CI is green because the test scenarios aren't applicable on containers so they weren't executed. When run on a virtual machine, this PR causes fail of test scenario no-grub.pass.sh : https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/bootloader-grub2/non-uefi/grub2_admin_username/tests/no-grub.pass.sh

Xeicker:
> The CI is green because the test scenarios aren't applicable on containers so they weren't executed. When run on a virtual machine, this PR causes fail of test scenario no-grub.pass.sh : https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/bootloader-grub2/non-uefi/grub2_admin_username/tests/no-grub.pass.sh

well, makes sense, the final purpose of this PR is that OVAL fails when grub.cfg is not present. So, should I rename that test to end with .fail.sh or would you consider this PR is not needed?

matejak:
The question is why there is such a check - I think that it has been introduced as a substitution of CPE checks - if there is no Grub or there is the other (efi/legacy) Grub installed, then make this rule pass, as another rule is supposed to catch this configuration.
I think that if it is the case, this should be fixed by means of CPEs that can additionally make sure that the proper NOTAPPLICABLE result is returned.

Xeicker:
@matejak, the CPE is actually implemented at the group.yml file under the bootloader-grub2 folder.
So I could make this change general and delete the no-grub.pass.sh test since it wouldn't be needed then.

jan-cerny:
Thank makes sense, so please remove the no-grub.pass.sh.

jan-cerny (approved):

PR #8285, opened 2022-03-03
Parametrize configuration of `kernel.kptr_restrict` and add rule for `kernel.panic_on_oops`

yuumasato:
#### Description:

- Parametrize rule `sysctl_kernel_kptr_restrict` to configure the system based on a variable.
  - Other values for  `kernel.kptr_restrict` are also relevant.
- Add rule to check for `kernel.panic_on_oops=1`
  - Kernel panicking on oops may help prevent attacks.

I can break these into separate PRs if preferred.


linux_os/guide/system/permissions/restrictions/sysctl_kernel_panic_on_oops/rule.yml
```diff
@@ -0,0 +1,31 @@
...
+
+description: '{{{ describe_sysctl_option_value(sysctl="kernel.panic_on_oops", value="1") }}}'
+
+rationale: |-
+    An attacker trying to exploit the kernel may trigger kernel OOPSes,
+    panicking the system will impede them from continuing.
+
+warnings:
+    - functionality: |-
+       The system may start to panic when it normally wouldn't. A non-catastrohic error that
```
jan-cerny, line 13:
```suggestion
       The system may start to panic when it normally wouldn't. A non-catastrophic error that
```

jan-cerny (approved):

PR #8287, opened 2022-03-03
fix aide_build_database rule and remediation to work with sles 12 and 15

brett060102:
aide_build_database check and remediation  which is in the cis profile for sle12 and sle15 does not work for SLE 12/15.

#### Description:

- There were a number of issues:
-  oval/shared.xml
   assumeed aide.conf had:
        @@define DBDIR /path
        database_out=file:@@{DBDIR}/filename
        database=file:@@{DBDIR}/filename
SLE 12 and 15 does not use: :@@{DBDIR} it just has:
         database_out=file:/pathfile
         database=file:/path/file

- in remediation
1:  bash was not enabled for sle
2: sle uses /var/lib/aide/aide.db.new and /var/lib/aide/aide.db not /var/lib/aide/aide.db.new.gz and /var/lib/aide/aide.db.gz
3: path to aide is /usr/bin/aide as opposed to /usr/sbin/aide --init

- in the rule.yml:
   Need to make sle aware for file name and path differences.
   Since .gx nv not gz also applied to ubuntu needed to re-code some
   {{% if not in product %}} to
   {{% if 'ubuntu' in product or 'sle' in product %}}

#### Rationale:

- get aide_build_database working for sle 12/15

### DIFF
--- a/linux_os/guide/system/software/integrity/software-integrity/aide/aide_build_database/ansible/shared.yml
+++ b/linux_os/guide/system/software/integrity/software-integrity/aide/aide_build_database/ansible/shared.yml
@@ -11,7 +11,11 @@
     - aide
 
 - name: "Build and Test AIDE Database"
+{{% if 'sle' in product %}}
+  command: /usr/bin/aide --init
+{{% else %}}
   command: /usr/sbin/aide --init
+{{% endif %}}
   changed_when: True
 
 # mainly to allow ansible's check mode to work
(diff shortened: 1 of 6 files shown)


linux_os/guide/system/software/integrity/software-integrity/aide/aide_build_database/ansible/shared.yml
```diff
@@ -22,8 +26,13 @@
 
 - name: "Stage AIDE Database"
   copy:
+{{% if 'sle' in product %}}
```
Mab879, line 29:
```suggestion
{{% if 'ubuntu' in product or 'sle' in product %}}
```

Mab879 (changes requested):
Thanks for the PR. Just one minor change to keep everything consistent.

brett060102:
Accepted suggestion. Thought about that initially, but don't like changing behavior for releases that I am not working on.

brett060102:
@Mab879 requested change has been made.

Mab879 (approved):
Thanks!

PR #8288, opened 2022-03-03
Include rule mount_option_proc_hidepid

marcusburghardt:
This rule uses a combination of parameter and value in the options field
of `fstab`. Therefore, the mount_option template was also updated to
support this format directly on the `mountoption` parameter or using
variables when calling the `mount_option` template from the `rule.yml` file.


marcusburghardt:
Regarding the error in `SSGTS / Run Tests (pull_request)`, it is related to a limitation in the container which is used for this test. The option is set in `fstab` file, but this file is not present in this image. This container specifically has the option `hidepid=invisible` predefined in its image and this cannot be changed by our test scenario scripts. Thus, we can waive this error.

Mab879 (changes requested):
Thanks for the PR.

The tests pass on a RHEL8 VM, but they are failing on a RHEL9 VM.

Also have few of minor suggestions as well.

Mab879 (approved):

PR #8296, opened 2022-03-06
gid_passwd_group_same oval does not allow ! in passwd field

brett060102:
password_storage/gid_passwd_group_same/oval/shared.xml is hard coded to only
allow for "x" in the passwd field in the /etc/group file.

SLE12/15 has two entries that use "!" in that field.
So, the test incorrectly fails.

#### Description:

- allow passwd field in /etc/group to be either "x" or "!"

#### Rationale:

- Correct test for SL12/15

### DIFF
--- a/linux_os/guide/system/accounts/accounts-restrictions/password_storage/gid_passwd_group_same/oval/shared.xml
+++ b/linux_os/guide/system/accounts/accounts-restrictions/password_storage/gid_passwd_group_same/oval/shared.xml
@@ -10,7 +10,7 @@
   <ind:textfilecontent54_object id="object_gid_passwd_group_same_var" version="1">
     <ind:filepath>/etc/group</ind:filepath>
     <!-- Group ID (GID) from /etc/group (3-rd column) captured as subexpression of this object -->
-    <ind:pattern operation="pattern match">^.*:[x!]:([0-9]+):</ind:pattern>
+    <ind:pattern operation="pattern match">^[^:]+:[^:]+:([0-9]+):</ind:pattern>
     <ind:instance operation="greater than or equal" datatype="int">1</ind:instance>
   </ind:textfilecontent54_object>
 
@@ -30,7 +30,7 @@
   <ind:textfilecontent54_object id="object_gid_passwd_group_same" version="1">
     <ind:filepath>/etc/passwd</ind:filepath>
     <!-- Group ID (GID) from /etc/passwd (4-th column) captured as subexpression of this object -->
-    <ind:pattern operation="pattern match">^.*:[0-9]+:([0-9]+):</ind:pattern>
+    <ind:pattern operation="pattern match">^[^:]+:[^:]+:[0-9]+:([0-9]+):</ind:pattern>
     <ind:instance operation="greater than or equal" datatype="int">1</ind:instance>
   </ind:textfilecontent54_object>
 
--- a/linux_os/guide/system/accounts/accounts-restrictions/password_storage/gid_passwd_group_same/oval/shared.xml
+++ b/linux_os/guide/system/accounts/accounts-restrictions/password_storage/gid_passwd_group_same/oval/shared.xml
@@ -10,7 +10,7 @@
   <ind:textfilecontent54_object id="object_gid_passwd_group_same_var" version="1">
     <ind:filepath>/etc/group</ind:filepath>
     <!-- Group ID (GID) from /etc/group (3-rd column) captured as subexpression of this object -->
-    <ind:pattern operation="pattern match">^.*:x:([0-9]+):</ind:pattern>
+    <ind:pattern operation="pattern match">^.*:[x!]:([0-9]+):</ind:pattern>
     <ind:instance operation="greater than or equal" datatype="int">1</ind:instance>
   </ind:textfilecontent54_object>
 


linux_os/guide/system/accounts/accounts-restrictions/password_storage/gid_passwd_group_same/oval/shared.xml
```diff
@@ -10,7 +10,7 @@
   <ind:textfilecontent54_object id="object_gid_passwd_group_same_var" version="1">
     <ind:filepath>/etc/group</ind:filepath>
     <!-- Group ID (GID) from /etc/group (3-rd column) captured as subexpression of this object -->
-    <ind:pattern operation="pattern match">^.*:x:([0-9]+):</ind:pattern>
+    <ind:pattern operation="pattern match">^.*:[x!]:([0-9]+):</ind:pattern>
```
marcusburghardt, line 13: If SLE uses two `!`, this regex wouldn't work for this case. It is also good to include a test scenario script for this case.

marcusburghardt (changes requested):

linux_os/guide/system/accounts/accounts-restrictions/password_storage/gid_passwd_group_same/oval/shared.xml
brett060102: @marcusburghardt I tested that regex and it seemed to handle both "x" and "!" in the passwd field in /etc/groups sle does have both "x" and "1" entries. I am pretty sure "[x!]" means mach a single char that is either a "x" or "!". 

I will look at adding a test case for this but not quite sure how to do it.  current tests are
```sampsone@bigubuntu:~/github/STIGS/content$ cat ./linux_os/guide/system/accounts/accounts-restrictions/password_storage/gid_passwd_group_same/tests/all_groups_exist.pass.sh 
#!/bin/bash

# remove offending users who don't have existing GIDs
while read -r line ; do
    login=$(echo $line | cut -f1 -d':')
    gid=$(echo $line | cut -f4 -d':')
    if ! cut -f3 -d':' /etc/group | grep -q "$gid" ; then
        userdel -f $login
    fi
done < "/etc/passwd"

sampsone@bigubuntu:~/github/STIGS/content$ cat ./linux_os/guide/system/accounts/accounts-restrictions/password_storage/gid_passwd_group_same/tests/user_with_nonexistent_group.fail.sh 
#!/bin/bash

groupdel 888888 || true
echo "testuser:x:8000:888888:testuser:/home/testuser:/bin/bash" >> /etc/passwd
```
Since all_groups_exist.pass.sh is not filtering on the passwd field I don't see how to do it there. sort of adding the filter and user_with_nonexistent_group.fail.sh only modifies /etc/passwd

I could add testuser2 to both /etc/group and /etc/passwd with "!" in the /etc/group passwd field, but that is all that I can think of.

brett060102: @marcusburghardt sorry misunderstood part of you comment. sle 12/15 only use a single "!" in the passwd field of /etc/group. I am still not sure what testing changes to make though, because it does not look like the existing check code is examining the passwd field entry.

marcusburghardt: Ok. When I firstly read the description I wrongly assumed the field would be composed by two `!`, like `!!`. : )

Regarding the test, I would suggest at least a `pass` test where the second field from `/etc/group` entries are replaced by `!`.
This test is to make sure the regex is working as expected both for `x` and `!` but also to catch any future OVAL changes that could affect this behavior. You could use the `all_groups_exist.pass.sh` as reference with minor changes.

brett060102: @marcusburghardt I think that the following will will make all_groups_exist.pass.sh match the oval.
```
sampsone@bigubuntu:~/github/STIGS/content$ cat ./linux_os/guide/system/accounts/accounts-restrictions/password_storage/gid_passwd_group_same/tests/all_groups_exist.pass.sh 
#!/bin/bash

# remove offending users who don't have existing GIDs
while read -r line ; do
    login=$(echo $line | cut -f1 -d':')
    gid=$(echo $line | cut -f4 -d':')
    if ! grep -P -e '^.*:[x!]' /tmp/group cut -f3 -d':' | grep -q "$gid" ; then
        userdel -f $login
    fi
done < "/etc/passwd"

But I am wondering the oval should change to:diff --git a/linux_os/guide/system/accounts/accounts-restrictions/password_storage/gid_passwd_group_same/oval/shared.xml b/linux_os/guide/system/accounts/accounts-restrictions/password_storage/gid_passwd_group_same/oval/shared.xml
index b04102b4df..1b5aea5b27 100644
--- a/linux_os/guide/system/accounts/accounts-restrictions/password_storage/gid_passwd_group_same/oval/shared.xml
+++ b/linux_os/guide/system/accounts/accounts-restrictions/password_storage/gid_passwd_group_same/oval/shared.xml
@@ -10,7 +10,7 @@
   <ind:textfilecontent54_object id="object_gid_passwd_group_same_var" version="1">
     <ind:filepath>/etc/group</ind:filepath>
     <!-- Group ID (GID) from /etc/group (3-rd column) captured as subexpression of this object -->
-    <ind:pattern operation="pattern match">^.*:[x!]:([0-9]+):</ind:pattern>
+    <ind:pattern operation="pattern match">^.*:*:([0-9]+):</ind:pattern>
     <ind:instance operation="greater than or equal" datatype="int">1</ind:instance>

This would match the test code and seems to match what the rule says. I can't see any reason for excluding GIDs with passwords. Although that happens so seldom, I doubt anyone though of it until now. 
I would think that all we care about is that the entry has three fields and that the last one is numbers,

brett060102:
@marcusburghardt What do you think about changing regex from:
-    <ind:pattern operation="pattern match">^.*:[x!]:([0-9]+):</ind:pattern>
to
+    <ind:pattern operation="pattern match">^.*:*:([0-9]+):</ind:pattern>

This will make the oval match the testing that we do and what is in the rule text.

brett060102:
@marcusburghardt after talking to one of our regex gurus, I updated the regex for both the group and passwd match expressions. This change does not require a test change since it does more closely how the test code works. 

The new regex follows the file structure up to the gid field in both files. 
Please tell me what you think.

marcusburghardt (approved):
LGTM

brett060102:
@marcusburghardt thank you.

PR #8297, opened 2022-03-07
Fix accounts_umask_etc_bashrc to work with sle12/15

brett060102:
sles 12/15 uses /etc/bash.bashrc not /etc/bashrc

#### Description:

- sles 12/15 uses /etc/bash.bashrc not /etc/bashrc

1: update tests and remediation to handle /etc/bash.bashrc
2: change platform to be multi_platform_all in bash remediation to match
   ansible code.

#### Rationale:

- get accounts_umask_etc_bashrc working for sle12/15


jan-cerny (changes requested):
@brett060102 Can you add this change also to `rule.yml` of this rule to make the rule description aligned with bash/ansible?

brett060102:
@jan-cerny Thank you and done.

jan-cerny (approved):
great!

PR #8300, opened 2022-03-07
Mark no_empty_passwords_etc_shadow as machine only

mildas:
#### Description:

Mark `no_empty_passwords_etc_shadow` same as `no_empty_passwords` as [machine only](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords/rule.yml#L28)

#### Rationale:
See https://bugzilla.redhat.com/show_bug.cgi?id=1856466#c16

### DIFF
--- a/linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords_etc_shadow/rule.yml
+++ b/linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords_etc_shadow/rule.yml
@@ -21,6 +21,8 @@ rationale: |-
 
 severity: high
 
+platform: machine
+
 identifiers:
     cce@sle12: CCE-83249-3
     cce@sle15: CCE-91155-2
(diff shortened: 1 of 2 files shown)


linux_os/guide/system/accounts/accounts-restrictions/password_storage/no_empty_passwords_etc_shadow/rule.yml
```diff
@@ -13,6 +13,10 @@ description: |-
     <pre>$ sudo passwd [username]</pre>
     Lock an account:
     <pre>$ sudo passwd -l [username]</pre>
+    Note that this rule is not applicable for systems running within a
+    container. Having user with empty password within a container is not
+    considered a risk, because it should not be possible to directly login into
+    a container anyway.
```
ggbecker, line 19: maybe a warning is better in this context. Something like: https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/software/sudo/sudoers_no_root_target/rule.yml#L42-L43

mildas: Okay, I will change it also in `no_empty_passwords` rule

jan-cerny (approved):

PR #8307, opened 2022-03-08
OCP4: Added control response for SI-4(24)

Vincent056:
A control response is added for SI-4(24), rule `file_integrity_exists`, `file_integrity_notification_enabled` and `compliance_notification_enabled` are added to the control.


jhrozek (approved):
/lgtm

Issue #8301, opened 2022-03-07
RHEL 7 - grub2_*_argument rules fail after kickstart installation


mildas:
#### Description of problem:
All grub2 argument rules fail after kickstart installations. The most affected profile is OSPP as it has the biggest portion of grub2 argument rules:
```
xccdf_org.ssgproject.content_rule_grub2_audit_argument - fail
xccdf_org.ssgproject.content_rule_grub2_audit_backlog_limit_argument - fail
xccdf_org.ssgproject.content_rule_grub2_vsyscall_argument - fail
xccdf_org.ssgproject.content_rule_grub2_page_poison_argument - fail
xccdf_org.ssgproject.content_rule_grub2_slub_debug_argument - fail
```

#### SCAP Security Guide Version:
4bca873

#### Operating System Version:
RHEL 7.9

#### Steps to Reproduce:
1. Install machine using kickstart, e.g. using OSPP profile
2. Scan machine:
`oscap xccdf eval --profile xccdf_org.ssgproject.content_profile_ospp --report xccdf_org.ssgproject.content_profile_ospp.html ssg-rhel7-ds.xml`

#### Actual Results:
grub2 rules fail

#### Expected Results:
All grub2 rules passes

#### Additional Information/Debugging Steps:
<details>
  <summary>Example of OVAL results - grub2_slub_debug_argument</summary>
  
<h4><span class="label label-primary">check kernel command line parameters for slub_debug=P in /boot/grub2/grub.cfg for all kernels</span> 
                <span class="label label-danger">failed</span> because of these items:</h4>

Path | Content
-- | --
/boot/grub2/grub.cfg | linux16 /vmlinuz-3.10.0-1160.el7.x86_64  root=/dev/mapper/VolGroup-LogVol06 ro rhgb quiet  rd.lvm.lv=VolGroup/LogVol06 rd.lvm.lv=VolGroup/lv_swap console=ttyS0  rd.shell=0 fips=1 boot=UUID=078b3f81-e34b-497f-8f11-db5e65a956b1 audit=1  audit_backlog_limit=8192 vsyscall=none page_poison=1 slub_debug=P  LANG=en_US.UTF-8
[...]

PR #8322, opened 2022-03-10
grub2_audit_argument test scenarios cleanup

mildas:
#### Description:
`grub2_audit_argument` is templated by `grub2_bootloader_argument`. The template has already implemented several test scenarios.
Removed tests:
`arg_not_there_etcdefaultgrub.fail.sh` - same as existing templated test
`arg_not_there_rhel8.fail.sh` - same as templated `arg_not_there_grubenv.fail.sh` test
`correct_grubenv.pass.sh` and `correct_value.pass.sh` - similar to templated `correct_value.pass.sh` (uses Jinja macro to ensure the correct command is used)

Note that several test scenarios doesn't pass on RHEL7 - see #8301.


jan-cerny:
But you didn't remove the invalid_rescue.sh, right?

mildas:
Aha, the error output was not relevant to this PR. The `invalidate_rescue.pass.sh` test scenario is templated test that was already updated in another PR - #8311 
So this PR is only about redundant test cleanup. I've updated the description

jan-cerny (approved):
OK, thanks for updating the description.

PR #8326, opened 2022-03-10
Add new GH job to generate XLSX table and HTML page with SRG mapping

ggbecker:
#### Description:

- Add new GH job to generate XLSX table and HTML page with SRG mapping

### DIFF
--- a/utils/create_srg_export.py
+++ b/utils/create_srg_export.py
@@ -11,9 +11,9 @@ import sys
 import string
 from typing.io import TextIO
 import xml.etree.ElementTree as ET
-import pandas as pd
 
 import convert_srg_export_to_xlsx
+import convert_srg_export_to_html
 
 try:
     import ssg.build_yaml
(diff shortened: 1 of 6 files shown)


utils/create_srg_export.py
```diff
@@ -249,6 +249,11 @@ def main() -> None:
                            used_rules)
         print(f"File written to {full_output}")
 
+    # export CSV to HTML table
+    import pandas as pd
```
Mab879, line 253: Do we want to have a switch for this?

Mab879 (commented):
I think this is a good idea and will be helpful in the RHEL9 STIG development.

utils/create_srg_export.py
ggbecker: Yes, we should definitely have an option for it. Also the script is in very early stages, it was just to have a proof of concept

ggbecker: updated

ggbecker:
@Mab879 FYI

utils/create_srg_export.py
```diff
@@ -36,6 +38,55 @@
...
+    'IA Control', 'CCI', 'SRGID', 'STIGID', 'SRG Requirement', 'Requirement',
+    'SRG VulDiscussion', 'Vul Discussion', 'Status', 'SRG Check', 'Check', 'SRG Fix',
+    'Fix', 'Severity', 'Mitigation', 'Artifact Description', 'Status Justification'
+    ]
+
+COLUMNS = string.ascii_uppercase[:17]  # A-Q uppercase letters
+
+COLUMN_MAPPINGS = dict(zip(COLUMNS, HEADERS))
+
+HTML_OUTPUT_TEMPLATE = '''
```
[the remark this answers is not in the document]
ggbecker: I've created a separate file to handle the HTML export

Mab879 (approved):

utils/create_srg_export.py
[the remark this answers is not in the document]
Mab879: That works. I was thinking about putting the template in another file but this works as well.

ggbecker:
I haven't tested the files publishing as I can't do it without actually pushing to master, so there might be needed some extra rounds of commits to fix any issue that may appear in the gh-pages deployment

PR #8327, opened 2022-03-10
Add auid criteria to rules related to syscall audit rules

Xeicker:
#### Description:

-  Added the auid criteria:` -F auid>=1000 -F auid!=unset`. To the rules `audit_rules_kernel_module_loading_init`, `audit_rules_kernel_module_loading_finit` and `audit_rules_kernel_module_loading_delete`

#### Rationale:

- This is to make those rules compliant with DISA's STIG requirementes OL08-00-030360, OL08-00-030380 and OL08-00-030390 for ol8; and RHEL-08-030360 and RHEL-08-030390 for rhel8

### DIFF
--- a/linux_os/guide/system/auditing/auditd_configure_rules/audit_kernel_module_loading/audit_rules_kernel_module_loading_delete/ansible/shared.yml
+++ b/linux_os/guide/system/auditing/auditd_configure_rules/audit_kernel_module_loading/audit_rules_kernel_module_loading_delete/ansible/shared.yml
@@ -4,6 +4,12 @@
 # disruption = low
 # strategy = configure
 
+{{% if product in ["ol8", "rhel8"] %}}
+{{% set auid_filters = "-F auid>=" ~ auid ~ " -F auid!=unset" %}}
+{{% else %}}
+{{% set auid_filters = "" %}}
+{{% endif %}}
+
 # What architecture are we on?
 
 - name: Set architecture for audit delete_module tasks
@@ -15,7 +21,7 @@
     {{{ ansible_audit_augenrules_add_syscall_rule(
       action_arch_filters="-a always,exit -F arch=b32",
       other_filters="",
-      auid_filters="",
+      auid_filters=auid_filters,
       syscalls=["delete_module"],
       key="module-change",
       syscall_grouping=[],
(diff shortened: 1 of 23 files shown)


ggbecker:
I'm pushing for this change to be applied as well for STIG EL7. This way we don't need the jinja macros and it makes the content more accurate. Unfortunately the change takes some time to propagate so we will need to live with these ifs for some time.

vojtapolasek (changes requested):
Hello, thank you for the changes, they look sane.
However, it would be good if you could modify tests as well. You will have to create another set of tests. One will be for OL8 and RHEL8,  other will be for the rest.
You can see example of such tests here:
https://github.com/ComplianceAsCode/content/tree/master/linux_os/guide/system/auditing/grub2_audit_argument/tests
Notice the
```
# platform = Red Hat Enterprise Linux 7
```
and similar lines at the top of test scenarios.

Xeicker:
> Hello, thank you for the changes, they look sane. However, it would be good if you could modify tests as well. You will have to create another set of tests. One will be for OL8 and RHEL8, other will be for the rest. You can see example of such tests here: https://github.com/ComplianceAsCode/content/tree/master/linux_os/guide/system/auditing/grub2_audit_argument/tests Notice the
> 
> ```
> # platform = Red Hat Enterprise Linux 7
> ```
> 
> and similar lines at the top of test scenarios.

I was considering to use JINJA instead, as in https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/dhcp/disabling_dhcp_server/service_dhcpd_disabled/tests/service_enabled.fail.sh

Because in this case there are many platforms that should be included in previous tests and only RHEL8 and OL8 for the new ones. So the JINJA approach would be easier

vojtapolasek:
OK, that is a solution as well. Let's try and see.

vojtapolasek:
Hello,
thank you for tests. There is one more thing to get done - test scenarios have to account for the case when the directory /etc/audit/rules.d does not exist. Can you add some condition at the begining of scenarios to create it?
Thank you.

vojtapolasek (approved):
I am approving this. thank you for all the changes. the failing tests are not related to the rules you have modified, it might be problemin the algorithm which selects rules to test.
Let's wait for the Openshift CI.

vojtapolasek:
There are temporary problems with Openshift CI and rule modifications do not touch OCP content. Merging.

PR #8328, opened 2022-03-10
make macros-bash.jinja more bash -e safe

brett060102:
In testing the generated remediation scripts created in ./build/bash (i.e
PRODUCT-cis.sh, PRODUCT-stig.sh,...)

We found issues running the the scrpits with error on exit "-e" with the
bash_fix_audit_syscall_rule macro in macros-bash.jinja

#### Description:

- There two general types of codings that caused problems.
1: use of grep outside of test i.e:
        grep -q -- "-F" <<< "$extra_fields"
        if [ $? -ne 0 ]
        then
            candidate_rules+=("$s_rule")
        fi
   grep will exit with 1 whenever search string is not found and will cause
   a "-x" abort, unless grep is within as test.
   Since, case is expected can recode as: grep str1 str2 || action

2: use test on undefined string: i.e.
   other_string=$([[ {{{ other_filters }}} ]] && echo " {{{ other_filters }}}")
   which genates:
   other_string=$([[ $OTHER_FILTERS ]] && echo " $OTHER_FILTERS")
   since it is not required that other_filters has a value, can fix with "|| /bin/true" i.e.
   other_string=$([[ {{{ other_filters }}} ]] && echo " {{{ other_filters }}}") || /bin/true

### DIFF
--- a/shared/macros-bash.jinja
+++ b/shared/macros-bash.jinja
@@ -1542,11 +1542,7 @@ do
         # Strip all the options and fields we know of,
         # than check if there was any field left over
         extra_fields=$(sed -E -e "s/{{{ action_arch_filters }}}//"  -e "s#{{{ other_filters }}}##" -e "s/{{{ auid_filters }}}//" -e "s/((:?-S [[:alnum:],]+)+)//g" -e "s/-F key=\w+|-k \w+//"<<< "$s_rule")
-        grep -q -- "-F" <<< "$extra_fields"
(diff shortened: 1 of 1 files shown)


brett060102:
I see that the ci/prow/e2e-aws-ocp4-moderate-node — Job failed.
and look at the errors, but I can't see how could be related to my change.
Retesting

jan-cerny (approved):

PR #8329, opened 2022-03-10
Update sshd_oval_check related default constant

Xeicker:
#### Description:

- Update DEFAULT_SSH_DISTRIBUTED_CONFIG from a string value to boolean

#### Rationale:

- The variable was defined as string, and this caused the jinja code behave unexpectedly in if statements.


jan-cerny:
You're right that it's broken, but unfortunately, I think that this fix isn't sufficient, because there is more inconsistencies:

* this constant `DEFAULT_SSH_DISTRIBUTED_CONFIG` is a default value for the product property `sshd_distributed_config`, but that property is set in multiple places to a quoted string `"true"`:
  * products/fedora/product.yml:18
  * products/rhel9/product.yml:18 
* gets passed to Jinja macro `ansible_sshd_set` (defined at shared/macros-ansible.jinja:177) and in the definition of that macro it has the default value of this parameter set to a quoted string `"false"`
* linux_os/guide/services/ssh/ssh_server/sshd_rekey_limit/oval/shared.xml doesn't compare the variable to `"true"` in 2 occurences but it handles that as a bool. I guess this is the source of your problem.
* used in Jinja macro `bash_sshd_remediation` where a quoted string is expected by the code
* used in Jinja macro `sshd_oval_check` where a quoted string is expected by the code
* and more

I have no idea why we use quoted strings instead of boolean values. @yuumasato or @ggbecker Can you shed some light on why we don't use booleans and why we convert them to strings by quoting?

But, I think it would be easier in your case to revert your change in constants.py and instead fix the code in linux_os/guide/services/ssh/ssh_server/sshd_rekey_limit/oval/shared.xml to use quoted strings in every condition ie. changing both `{{%- if sshd_distributed_config %}}` to `{{%- if sshd_distributed_config == "true" %}}` there.

Xeicker:
I've seen this other approach, not sure if it is applicable here since there is already a default definition.
`data["missing_parameter_pass"] = parse_template_boolean_value(data, parameter="missing_parameter_pass", default_value=False)`
this is under the preprocess function of a template.py file

Xeicker:
Fixed the rule sshd_rekey_limit. But the issue i saw was comming from `sshd_oval_check` macro. I would really appreciate to know why the usage of strings instead of booleans

jan-cerny:
sshd_oval_check suffer from the same problem, there is also sometimes `{{%- if sshd_distributed_config %}}` and sometimes `{{%- if sshd_distributed_config == "true" %}}` there

jan-cerny:
> sshd_oval_check suffer from the same problem, there is also sometimes `{{%- if sshd_distributed_config %}}` and sometimes `{{%- if sshd_distributed_config == "true" %}}` there

I see you fixed that. Great!

yuumasato:
I don't remember exactly why we went with it as a string. 
From Jinja perspective, it should be fine to use boolean values: https://jinja.palletsprojects.com/en/3.0.x/templates/#if

@jan-cerny My guess is that I was concerned with how yaml would load the booleans, I remember there was a quirk when loading them. Do you remember what was it?
`sshd_distributed_config` is also defined in `product.yml` files:
https://github.com/ComplianceAsCode/content/commit/e04685101ee#diff-558a4dc22d2e5ac522756895e9629dd2ed729d42d814265f72d84d323612389fR18

jan-cerny:
@yuumasato

> My guess is that I was concerned with how yaml would load the booleans, I remember there was a quirk when loading them. Do you remember what was it?

I now recalled that this quirk is our deliberate decision.

We are using a modified YAML parser. By default, in YAML strings `true` and `false` (without quotes) are by default converted to boolean values. But we have code that overrrides this behavior so that YAML strings `true` and `false` (without quotes) are not converted and are considered strings. The code is here: https://github.com/ComplianceAsCode/content/blob/1a72555b2e90b41b8c18d5361ed2b5255b22a636/ssg/yaml.py#L29 and it was introduced by https://github.com/ComplianceAsCode/content/pull/2737.

But the reasoning in that PR seems quite vague for me.

So if I understand it well, if he would use `true` or `false` without quotes in these YAML files that he changes it would be handled as strings and therefore he would have to compare against strings (with quotes) wherever he uses it.

jan-cerny (approved):

PR #8331, opened 2022-03-11
OCP: SC-7(10) Control Response

Vincent056:
Added control response for SC-7(10)


controls/nist_ocp4.yml
```diff
@@ -13530,13 +13530,27 @@ controls:
     https://issues.redhat.com/browse/CMP-478
   rules: []
 - id: SC-7(10)
-  status: pending
+  status: automated
```
rhmdnd, line 13533: This seems like it would be a good candidate for a `partial` status (something to track in the future) since this appears to be a pretty hard check to fully automate (at least based on my interpretation of the control).

controls/nist_ocp4.yml
```diff
@@ -13530,13 +13530,27 @@ controls:
...
 
-    https://issues.redhat.com/browse/CMP-482
-  rules: []
+    NetworkPolicies are on by default for the cluster when using the OCP CNI
+    plugin. [1][2] Cluster admin must add policies to relevant projects.
+
+    Depending on the SDN implementation, there will be different logs
+    and interfaces to monitor the network flows.
+    
+    Application level firewall can be provided by underlying cloud's provider,
```
rhmdnd, line 13544: nit: provider.

underlying cloud provider?

rhmdnd (commented):
Only a couple minor comments and a question (which probably doesn't need to be addressed here).

controls/nist_ocp4.yml
Vincent056: I agree, I think we could only check for part of it

jhrozek (approved):
LGTM, looks like Lance's comments have been adressed.

PR #8332, opened 2022-03-11
OCP: Add vaiabled support for kubelet_configure_tls_cipher_suites

Vincent056:
User can choose which cipher to be used by setting two variable, var_kubelet_tls_cipher_suites_regex, and var_kubelet_tls_cipher_suites


jhrozek (commented):
Nice work, thank you. Just one small grammar nit and then we're good to go.

jhrozek (approved):
/lgtm
sorry for the late review, I thought I had already acked the PR

PR #8334, opened 2022-03-11
Enforce rounds parameter presence

Xeicker:
#### Description:

- Modified accounts_password_pam_unix_rounds_password_auth & accounts_password_pam_unix_rounds_system_auth OVAL so it doesn't allow to pass with default settings

#### Rationale:

- Enforce explicit parameter declaration and not rely on default


jan-cerny:
@Mab879 or @ggbecker Is this something that is applicable also to RHEL 8 and/or 9?

ggbecker:
> @Mab879 or @ggbecker Is this something that is applicable also to RHEL 8 and/or 9?

to certain extent, yes. Having the parameter explicitly set and not relying on the default seems to be the trend nowadays. On the other hand, the default parameter will be compared with the variable anyways, in case of ANSSI profile, it will return failure since it uses rounds>5000(default value)

Xeicker:
Closed PR since latest OL8 STIG (v1r2) no longer requires these rules

PR #8336, opened 2022-03-11
Update oval name resolution

Xeicker:
#### Description:

- Update network_configure_name_resolution OVAL behavior

#### Rationale:

- These criteria introduced are part of the DISA requirements under the STIG IDs that this rule references

> stigid@ol7: OL07-00-040600
    stigid@ol8: OL08-00-010680
    stigid@rhel7: RHEL-07-040600
    stigid@rhel8: RHEL-08-010680


jan-cerny (changes requested):
Please describe this in rule.yml of this rule to make the description and metadata aligned with the actual check. Also, please add test scenarios.

jan-cerny (approved):
Thank you!

PR #8337, opened 2022-03-11
Fix bash remediation of disable_users_coredumps

Xeicker:
#### Description:

- Update regex in sed command to differentiate commented lines. And included another sed command to remove unnecesary conf lines

#### Rationale:

- When the confline were commented out, the previous version fix the line but keeps it commented. Also avoid that limits.d/*.conf files override main configuration

### DIFF
--- a/linux_os/guide/system/permissions/restrictions/coredumps/disable_users_coredumps/bash/shared.sh
+++ b/linux_os/guide/system/permissions/restrictions/coredumps/disable_users_coredumps/bash/shared.sh
@@ -1,8 +1,10 @@
 # platform = Red Hat Virtualization 4,multi_platform_rhel,multi_platform_ol
 SECURITY_LIMITS_FILE="/etc/security/limits.conf"
 
-if grep -qE '\*\s+hard\s+core' $SECURITY_LIMITS_FILE; then
+if grep -qE '^\s*\*\s+hard\s+core' $SECURITY_LIMITS_FILE; then
         sed -ri 's/(hard\s+core\s+)[[:digit:]]+/\1 0/' $SECURITY_LIMITS_FILE
 else
         echo "*     hard   core    0" >> $SECURITY_LIMITS_FILE
 fi
+
+sudo sed -ri '/^\s*\*\s+hard\s+core/d' /etc/security/limits.d/*.conf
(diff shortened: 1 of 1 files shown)


linux_os/guide/system/permissions/restrictions/coredumps/disable_users_coredumps/bash/shared.sh
```diff
@@ -1,8 +1,10 @@
...
 SECURITY_LIMITS_FILE="/etc/security/limits.conf"
 
-if grep -qE '\*\s+hard\s+core' $SECURITY_LIMITS_FILE; then
+if grep -qE '^\s*\*\s+hard\s+core' $SECURITY_LIMITS_FILE; then
         sed -ri 's/(hard\s+core\s+)[[:digit:]]+/\1 0/' $SECURITY_LIMITS_FILE
 else
         echo "*     hard   core    0" >> $SECURITY_LIMITS_FILE
 fi
+
+sudo sed -ri '/^\s*\*\s+hard\s+core/d' /etc/security/limits.d/*.conf
```
jan-cerny, line 10: Don't use `sudo` in remediation script. Users are expected to run them under sudo or as root anyway.

jan-cerny (changes requested):

jan-cerny:
There was another problem that was hidden under the sudo call and now it got revealed as we can see from the test suite output:

```
Remediating rule 1/1: 'xccdf_org.ssgproject.content_rule_disable_users_coredumps'
+ rpm --quiet -q pam
+ SECURITY_LIMITS_FILE=/etc/security/limits.conf
+ grep -qE '^\s*\*\s+hard\s+core' /etc/security/limits.conf
+ sed -ri 's/(hard\s+core\s+)[[:digit:]]+/\1 0/' /etc/security/limits.conf
+ sed -ri '/^\s*\*\s+hard\s+core/d' '/etc/security/limits.d/*.conf'
sed: can't read /etc/security/limits.d/*.conf: No such file or directory
```

Your code doesn't handle a situation when the directory /etc/security/limits.d/ is empty. The glob /etc/security/limits.d/*.conf doesn't expand.

jan-cerny (approved):

PR #8343, opened 2022-03-14
Remove shared_xccdf-removeaux.xslt

jan-cerny:
The XSLT template shared_xccdf-removaux.xslt only removes
xccdf:Group[@id="srg_support"] elements and copies all other
elements and attributes. But there are no "srg_support" elements
in any xccdf-linked.xml file. And nothing reads or uses these elements.
So the template effectively just copies the file. So we can remove it.


jan-cerny:
I have rebased on master

vojtapolasek (approved):
LGTM, merging

PR #8348, opened 2022-03-14
OCP: cluster version history bug fix

Vincent056:
This pr is to address BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2053602, the rule was failing because the initial OpenShift install will always have verified as false, and this pr addresses the issue by skipping the first entry in the history list.


jhrozek (approved):
LGTM
nice work

PR #8351, opened 2022-03-15
Remove one "oscap xccdf resolve" step

jan-cerny:
In this step the "oscap xccdf resolve" is superfluous because it
doesn't do any significant changes there.

What the removed "oscap xccdf resolve" does:
* shuffles XML namespace prefixes, but that doesn't change the actual
  namespace, so we don't need it
* orders attributes of some elements, but ordering of attributes doesn't
  matter in XML
* sets explicitly the reboot attribute of xccdf:fix element to its
  default value (false), but as it's the default we don't need it
* sets the resolved attribute back to true which was changed forcefully
  to false in unselect_empty_xccdf_groups.py to trick the oscap command,
  so we can instead remove that from unselect_empty_xccdf_groups.py)

Other XCCDF resolving steps are not effective in this build phase.

Notice that there still remain other occurrences of "oscap xccdf
resolve" elsewhere.

### DIFF
--- a/build-scripts/unselect_empty_xccdf_groups.py
+++ b/build-scripts/unselect_empty_xccdf_groups.py
@@ -99,9 +99,6 @@ def main():
             "Make sure the input file is a resolved XCCDF Benchmark."
         )
 
-    # force another oscap resolve to fix namespace prefixes
-    root_element.set("resolved", "0")
-
     affected_profiles = []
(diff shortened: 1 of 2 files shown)


matejak (approved):
LGTM, it works, so let's merge it well ahead of the next upstream release to iron out possible regressions.

PR #8364, opened 2022-03-16
Build data stream without OpenSCAP

jan-cerny:
#### Description:
Instead of using `oscap ds sds-compose` we will use our custom Python script to compose an SCAP source data stream from individual SCAP components during the build process.

#### Rationale:
Speeds up the build. Removes dependency on `oscap` in this part of the build process. Notice that there are still other occurrences of `oscap` across the build system.

For more information, please read commit message of every commit.


build-scripts/compose_ds.py
```diff
@@ -0,0 +1,88 @@
...
+
+def add_component(
+        ds_collection, component_ref_parent, component_file_name,
+        dependencies=None):
+    component_id = "scap_%s_comp_%s" % (
+        ID_NS, os.path.basename(component_file_name))
+    component = ET.SubElement(
+        ds_collection, "{%s}component" % datastream_namespace)
+    component.set("id", component_id)
+    component.set("timestamp", "2022-03-16T14:46:10")
```
jan-cerny, line 34: Please add a real timestamp.

jan-cerny:
Unfortunately, not much, it stays roughly the same, because the removed command took 1.3 s, and the new script manages the task in 0.9 s, so it can save max 0.4 s. The benefit is perhaps in removing dependency on `openscap`.

jan-cerny:
I have rebased this on the top of current master branch because #8357 has been merged.

PR #8369, opened 2022-03-17
Remove update_sds_version.py

jan-cerny:
#### Description:
Remove update_sds_version.py and instead generates both SCAP 1.2 and SCAP 1.3 source data stream in a single step (single command). It's done by extending script compose_ds.py which has been introduced in #8364.

#### Rationale:
Avoid loading the SCAP source data stream from the disc more times than needed. Simplify and streamline CMakeLists.

For more information, please read commit message of every commit.


vojtapolasek (approved):
Looks good, thank you for simplifying of the build system.

PR #8373, opened 2022-03-17
OCP : Update control response for AC-6(8)

TrilokGeer:
Added control response for AC-6(8) w.r.t SCC.

### DIFF
--- a/controls/nist_ocp4.yml
+++ b/controls/nist_ocp4.yml
@@ -1090,11 +1090,15 @@ controls:
   levels:
   - high
 - id: AC-6(8)
-  status: pending
+  status: supported
   notes: |-
-    A control response is planned. Progress can be tracked via:
+    Red Hat OpenShift provides security context constraints (SCC) [1] that control permissions for actions that a pod can perform and what resources a pod can access. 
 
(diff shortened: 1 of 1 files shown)


controls/nist_ocp4.yml
```diff
@@ -1090,11 +1090,15 @@ controls:
   levels:
   - high
 - id: AC-6(8)
-  status: pending
+  status: supported
   notes: |-
-    A control response is planned. Progress can be tracked via:
+    Red Hat OpenShift provides security context constraints (SCC) [1] that control permissions for actions that a pod can perform and what resources a pod can access.
```
Vincent056, line 1095: In addition to what's here, I think maybe it is also good to add they can use RBAC, we can use the control response from CM-5 here.

https://github.com/ComplianceAsCode/content/blob/29ea4c3ed70d46afbe364544eee883eb07ee10d7/controls/nist_ocp4.yml#L4591

TrilokGeer: Thanks @Vincent056 for the review, updated the control response to include RBAC.

Vincent056 (approved):

PR #8376, opened 2022-03-17
SRG-OS-000038-GPOS-00016

Mab879:
#### Description:

- Update SRG-OS-000038-GPOS-00016 for RHEL9 STIG
- Add macro to describe for how to enable a service for fix text
- Add fix to service_auditd_enabled
- Add fix text to package_audit_installed

#### Rationale:

RHEL9 STIG

### DIFF
--- a/linux_os/guide/system/auditing/package_audit_installed/rule.yml
+++ b/linux_os/guide/system/auditing/package_audit_installed/rule.yml
@@ -39,6 +39,12 @@ ocil_clause: 'the package is not installed'
 
 ocil: '{{{ ocil_package("audit") }}}'
 
+fix: |-
+    Install the audit service (if the audit service is not already installed) with the following command:
+
+    {{{ package_install("audit") }}}
+
+
 template:
     name: package_installed
     vars:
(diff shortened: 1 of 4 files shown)


linux_os/guide/system/auditing/package_audit_installed/rule.yml
```diff
@@ -39,6 +39,12 @@ ocil_clause: 'the package is not installed'
 
 ocil: '{{{ ocil_package("audit") }}}'
 
+fix: |-
+    Install the audit service (if the audit service is not already installed) with the following command:
+
+    $ {{{ package_install("audit") }}}
```
jan-cerny, line 45: This macro already contains a dollar, so you will have 2 dollars.

The generated code is:

```
fix: "Install the audit service (if the audit service is not already installed) with\
    \ the following command:\n\n$ \n$ sudo dnf install audit"
```

```suggestion
    {{{ package_install("audit") }}}
```

jan-cerny (changes requested):

jan-cerny (approved):

PR #8386, opened 2022-03-18
Turn all wireless interfaces off

freddieRv:
#### Description:

- Update bash remediation for `wireless_disable_interfaces` so that both WIFI and WWAN wireless interfaces are turned off

#### Rationale:

- This is an effort to align the `wireless_disable_interfaces` rule with the DISA STIG OL08-00-040110 requirement


jan-cerny (approved):
It seems to be the case also for RHEL 8: https://www.stigviewer.com/stig/red_hat_enterprise_linux_8/2021-03-04/finding/V-230506

PR #8391, opened 2022-03-21
Update SRG-OS-000074-GPOS-00042 for RHEL9

Mab879:
#### Description:

- Update SRG-OS-000074-GPOS-00042 for RHEL9
- Add fix text for package_vsftpd_removed

#### Rationale:
RHEL 9 STIG

### DIFF
--- a/controls/stig_rhel9/SRG-OS-000074-GPOS-00042.yml
+++ b/controls/stig_rhel9/SRG-OS-000074-GPOS-00042.yml
@@ -2,8 +2,7 @@ controls:
     -   id: SRG-OS-000074-GPOS-00042
         levels:
             - medium
-        title: The operating system must transmit only encrypted representations of passwords.
+        title: {{{ full_name }}} must transmit only encrypted representations of passwords.
         rules:
             - package_vsftpd_removed
-            - sshd_allow_only_protocol2
         status: automated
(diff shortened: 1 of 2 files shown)


controls/stig_rhel9/SRG-OS-000074-GPOS-00042.yml
```diff
@@ -2,8 +2,7 @@ controls:
     -   id: SRG-OS-000074-GPOS-00042
         levels:
             - medium
-        title: The operating system must transmit only encrypted representations of passwords.
+        title: {{{ full_name }}} must transmit only encrypted representations of passwords.
         rules:
             - package_vsftpd_removed
-            - sshd_allow_only_protocol2
```
jan-cerny, line 8: When you remove this rule ID from here, should you remove the SRG ID from its rule.yml?

Mab879: The reason why it was remove is because this option does nothing in RHEL9 (and the same could be said about RHEL8 as well).  Since [OpenSSH 7.6](https://www.openssh.com/txt/release-7.6) (Released 2017-10-03) OpenSSH has only support for SSH protocol version 2. So this SRG item is still valid for RHEL7.  Do we want to remove it from RHEL9 then?

jan-cerny: Removing it from RHEL 9 sounds logical to me.

Mab879: Since the rule is already in the RHEL 8 STIG, I think this is should be a RHEL9+ change.

PR #8397, opened 2022-03-22
Introduce new abrt related rules to replace obsolete packages

### LINKED ISSUE #8286: Some abrt rules are not applicable on RHEL8
#### Description of problem:
Rule `package_abrt-plugin-logger_removed` should not be present on RHEL-8 as `abrt-plugin-logger` RPM is not in RHEL8 composes (it is selected in `ospp` and `stig` profiles). Other abrt rules need to be checked as well, for example `package_abrt-plugin-rhtsupport_removed` is also obsolete, `abrt-plugin-rhtsupport` RPM doesn't exist and there is rather `libreport-plugin-rhtsupport`.
Fortunately these are only minor issues the rules are about package removal and it only manifests when using SSGTS test scenarios.

master

RHEL8

ggbecker:
#### Description:

- Introduce new abrt related rules to replace obsolete packages

- Fixes #8286


products/rhel8/profiles/ospp.profile
```diff
@@ -210,6 +210,8 @@ selections:
     - package_abrt-addon-ccpp_removed
     - package_abrt-plugin-sosreport_removed
     - package_abrt-cli_removed
+    - package_libreport-plugin-rhtsupport_removed
+    - package_libreport-plugin-logger_removed
```
yuumasato, line 214: Does it makes sense to add these two rules when their old equivalents are not selected?
- package_abrt-plugin-logger_removed
- package_abrt-plugin-rhtsupport_removed

Same for the STIG profiles.

ggbecker: I believe so, the package is basically the same with a different name if I'm not mistaken. The only caveat for STIG is that their recommendation is to remove all `abrt*` package which does not include these new packages.

yuumasato (approved):
LGTM, just pending tests.

yuumasato:
Not sure why `libreport-plugin-rhtsupport` was not installed, it was available in VMs.

PR #8398, opened 2022-03-22
Use UID field for bash remediation of homedirs

cipherboy:
Per documentation, /etc/passwd has the first field as the username, the
second field as the password, the third as UID, and the fourth as GID.
The existing bash remediation used the fourth field (GID) as the UID_MIN
check, which while likely nominally functional, isn't necessarily
correct. Instead, the third field (UID) should be used for comparison.

See also: https://docs.oracle.com/cd/E86824_01/html/E54775/passwd-4.html

```
Signed-off-by: Alexander Scheel <alex.scheel@hashicorp.com>
```


cipherboy:
Sorry folks, not sure why GH UI defaulted to opening the branch on the CaC repo directly :o

Mab879 (approved):

PR #8400, opened 2022-03-22
Add centos stream dockerfiles and use them in GHA SSGTS

ggbecker:
#### Description:

- Adds dockerfiles for Centos Stream 8/9
- Add new workflows to test content on those OSes.

### DIFF
--- a/.github/workflows/ssgts-cs8.yaml
+++ b/.github/workflows/ssgts-cs8.yaml
@@ -96,7 +96,7 @@ jobs:
       - name: Get oscap-ssh
         if: ${{ steps.ctf.outputs.CTF_OUTPUT_SIZE != '0' }}
         run: |
-          wget https://raw.githubusercontent.com/OpenSCAP/openscap/maint-1.2/utils/oscap-ssh
+          wget https://raw.githubusercontent.com/OpenSCAP/openscap/maint-1.3/utils/oscap-ssh
           sudo chmod 755 oscap-ssh
           sudo mv -v oscap-ssh /usr/local/bin
           sudo chown root:root /usr/local/bin/oscap-ssh
--- a/.github/workflows/ssgts-cs8.yaml
+++ b/.github/workflows/ssgts-cs8.yaml
@@ -1,7 +1,9 @@
-name: SSGTS Fedora
+name: SSGTS CS8
 on:
   pull_request:
     branches: [ master, 'stabilization*' ]
+env:
+  DATASTREAM: ssg-centos8-ds.xml
 jobs:
   build-content:
     name: Build Content
@@ -52,12 +54,12 @@ jobs:
           prop_path: 'product'
       - name: Build product
         if: ${{ steps.ctf.outputs.CTF_OUTPUT_SIZE != '0' }}
-        run: ./build_product ${{steps.product.outputs.prop}} --datastream-only
+        run: ./build_product rhel8 --derivatives
       - uses: actions/upload-artifact@v2
         if: ${{ steps.ctf.outputs.CTF_OUTPUT_SIZE != '0' }}
         with:
-          name: ssg-${{steps.product.outputs.prop}}-ds.xml
-          path: build/ssg-${{steps.product.outputs.prop}}-ds.xml
+          name: ${{ env.DATASTREAM }}
+          path: build/${{ env.DATASTREAM }}
   validate-ubuntu:
     name: Run Tests
     needs: build-content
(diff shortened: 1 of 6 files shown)


ggbecker:
test rule in container doesn't add the cpe item into the datastream, so it relies on OpenSCAP's dictionary which may miss items for recent OSes for example CentOS Stream 9 CPE, so all test scenarios return not applicable.

If the built datastream is specifically for CentOS it should work, but on the other hand improving the test_suite.py to add the cpe item alongside the platform item as well should solve the problem even for future releases.

.github/workflows/ssgts-cs8.yaml
```diff
@@ -0,0 +1,178 @@
+name: SSGTS CS8
+on:
+  pull_request:
+    branches: [ master, 'stabilization*' ]
+jobs:
+  build-content:
+    name: Build Content
+    runs-on: ubuntu-latest
+    container:
+      image: fedora:latest
```
matusmarhefka, line 10: This is no longer needed in CentOS jobs right? If so we should remove it to save time pulling the image from the registry.

.github/workflows/ssgts-cs8.yaml
```diff
@@ -0,0 +1,178 @@
...
+        if: ${{ steps.ctf.outputs.CTF_OUTPUT_SIZE != '0' }}
+        run: ssh-keygen -N '' -t rsa -f ~/.ssh/id_rsa
+      - name: Build test suite container
+        if: ${{ steps.ctf.outputs.CTF_OUTPUT_SIZE != '0' }}
+        run: podman build --build-arg "CLIENT_PUBLIC_KEY=$(cat ~/.ssh/id_rsa.pub)" -t ssg_test_suite -f test_suite-cs8
+        working-directory: ./Dockerfiles
+      - name: Get oscap-ssh
+        if: ${{ steps.ctf.outputs.CTF_OUTPUT_SIZE != '0' }}
+        run: |
+          wget https://raw.githubusercontent.com/OpenSCAP/openscap/maint-1.2/utils/oscap-ssh
```
matusmarhefka, line 94: We should rather use `maint-1.3` branch here.

Dockerfiles/test_suite-cs8
```diff
@@ -0,0 +1,24 @@
+# This Dockerfile is a minimal example for a Centos Stream 9 SSG test suite target container.
```
matusmarhefka, line 1: `Centos Stream 8`

.github/workflows/ssgts-cs8.yaml
ggbecker: We still need fedora here as this is the step we build the content. Otherwise it would build the content on ubuntu with an outdated openscap. The place where the image is pulled for SSGTS is in a different place.

.github/workflows/ssgts-cs8.yaml
ggbecker: I'm going to change and see how it goes.

Dockerfiles/test_suite-cs8
ggbecker: fixed

.github/workflows/ssgts-cs8.yaml
```diff
@@ -0,0 +1,178 @@
...
+        with:
+          path: 'output.json'
+          prop_path: 'ansible'
+      - uses: actions/download-artifact@v2
+        if: ${{ steps.ctf.outputs.CTF_OUTPUT_SIZE != '0' }}
+        with:
+          name: ssg-${{steps.product.outputs.prop}}-ds.xml
+      - name: Run tests in a container - Bash
+        if: ${{steps.bash.outputs.prop == 'True' && steps.ctf.outputs.CTF_OUTPUT_SIZE != '0' }}
+        run: tests/test_rule_in_container.sh --dontclean --logdir logs_bash --remediate-using bash --name ssg_test_suite --datastream ssg-${{steps.product.outputs.prop}}-ds.xml ${{join(steps.rules.outputs.prop)}}
```
matusmarhefka, line 133: I noticed that ctf returns rhel8 data stream when testing on cs9 container. What about not using data stream based on ctf for now and instead hardcode it based on a container on which we run tests?

ggbecker: I believe that's the best approach here.

ggbecker: I should get to this soon :)

ggbecker:
@matusmarhefka I've updated the Centos Stream 9 job to use the cs9 derivative datastream. Let's see how it goes.

ggbecker:
@matusmarhefka it should be final now.

matusmarhefka (approved):
Thanks, it works great! Failing tests with Ansible remediations for the rule `sshd_enable_strictmodes` are unrelated to this PR.

PR #8404, opened 2022-03-22
File owners, groupowners, permissions should be able to recurse and file_regex

### LINKED ISSUE #8303: RHEL 7 - file_ownership_library_dirs remediation fails due to `chown` file dereference/access
#### Description of problem:
During kickstart installation of E8 profile, the `file_ownership_library_dirs` remediation fails due `chown` error with dereference or with access.

4bca873

RHEL 7.9

#### Steps to Reproduce:

1. Install machine using E8 kickstart
2. Check HTML report created by OAA during installation

or

1. `$ python3 tests/test_suite.py profile --libvirt qemu:///system test_suite_vm --datastream build/ssg-rhel7-ds.xml --mode online --remediate-using oscap xccdf_org.ssgproject.content_profile_e8`

#### Actual Results:
Remediation output:
```
chown: cannot dereference '/lib/debug/usr/.dwz': No such file or directory
chown: cannot dereference '/lib/modules/3.10.0-1160.el7.x86_64/build': No such file or directory
chown: cannot dereference '/lib/modules/3.10.0-1160.el7.x86_64/source': No such file or directory
chown: cannot access '/lib/firmware/brcm/brcmfmac43430a0-sdio.ONDA-V80': No such file or directory
chown: cannot access 'PLUS.txt': No such file or directory
chown: cannot access '/lib/firmware/brcm/brcmfmac43455-sdio.MINIX-NEO': No such file or directory
chown: cannot access 'Z83-4.txt': No such file or directory
basename: extra operand '(dev).py'
Try 'basename --help' for more information.
chown: cannot access '/lib/python2.7/site-packages/setuptools/script': No such file or directory
chown: cannot access 'template.py': No such file or directory
basename: extra operand '(dev).pyc'
Try 'basename --help' for more information.
...
```

#### Expected Results:
Remediation passes

yuumasato:
#### Description:

- Rules `file_ownership_library_dirs` and `file_permissions_library_dirs` should recurse and check files down in the tree.
  The rules were only checking the first level of directory.
- Fix handling of symlinks in the remdiation.
  Now it doesn't try to de-reference the symlink, and remediation changes ownership of the symlink.
- Fix handling of files with spaces.

#### Rationale:

- `recurse` and `file_regex` are not mutually exclusive.
- Fixes #8303

#### Notes
- This was found because in RHEL7, the first level of `/lib' doesn't have any file, only directories:
```
[root@localhost ~]# ls /lib
binfmt.d  debug  dracut  firewalld  firmware  games  grub  kbd  kde3  kde4  kdump  kernel  locale  modprobe.d  modules  modules-load.d  NetworkManager  polkit-1  python2.7  rpm  sendmail  sendmail.postfix  sse2  sysctl.d  systemd  tmpfiles.d  tuned  udev  yum-plugins
```

### DIFF
--- a/linux_os/guide/system/permissions/files/permissions_within_important_dirs/file_ownership_library_dirs/rule.yml
+++ b/linux_os/guide/system/permissions/files/permissions_within_important_dirs/file_ownership_library_dirs/rule.yml
@@ -69,5 +69,6 @@ template:
             - /lib64/
             - /usr/lib/
             - /usr/lib64/
+        recursive: 'true'
         file_regex: ^.*$
         fileuid: '0'
--- a/linux_os/guide/system/permissions/files/permissions_within_important_dirs/file_ownership_library_dirs/tests/incorrect_symlink.fail.sh
+++ b/linux_os/guide/system/permissions/files/permissions_within_important_dirs/file_ownership_library_dirs/tests/incorrect_symlink.fail.sh
@@ -1,0 +1,16 @@
+# platform = multi_platform_sle,multi_platform_rhel,multi_platform_fedora,multi_platform_ubuntu
+
+useradd user_test
+
+TESTDIR="/usr/lib/"
+
+# The remediation performs a 'find' followed by a 'chwon'
+# While 'find' doesn't follow symlinks by default, 'chown' does follow,
+# so 'chown' will try to change owner of a non existent file while 'find'
+# pointed out that the symlink has wrong owner.
+ln -s $TESTDIR/mising_test_file $TESTDIR/faulty_symlink
+chown -h user_test $TESTDIR/faulty_symlink
+
+# The Check ignores symlink, so we need to put a reason to run the remediations
+touch $TESTDIR/test_me
+chown user_test $TESTDIR/test_me
--- a/shared/templates/file_groupowner/ansible.template
+++ b/shared/templates/file_groupowner/ansible.template
@@ -5,13 +5,18 @@
 # disruption = low
 
 {{% for path in FILEPATH %}}
-{{% if IS_DIRECTORY and FILE_REGEX %}}
+{{% if IS_DIRECTORY %}}
+{{% if FILE_REGEX %}}
 
-- name: Find {{{ path }}} file(s) matching {{{ FILE_REGEX[loop.index0] }}}
+- name: Find {{{ path }}} file(s) matching {{{ FILE_REGEX[loop.index0] }}}{{% if RECURSIVE %}} recursively{{% endif %}}
+
   find:
     paths: "{{{ path }}}"
     patterns: {{{ FILE_REGEX[loop.index0] }}}
     use_regex: yes
+{{% if RECURSIVE %}}
+    recurse: yes
+{{% endif %}}
     hidden: yes
   register: files_found
 
@@ -22,15 +27,18 @@
   with_items:
     - "{{ files_found.files }}"
 
-{{% elif IS_DIRECTORY and RECURSIVE %}}
+{{% else %}}
 
-- name: Ensure group owner on {{{ path }}} recursively
+- name: Ensure group owner on {{{ path }}}{{% if RECURSIVE %}} recursively{{% endif %}}
   file:
     path: "{{{ path }}}"
     state: directory
+{{% if RECURSIVE %}}
     recurse: yes
+{{% endif %}}
     group: "{{{ FILEGID }}}"
 
+{{% endif %}}
 {{% else %}}
 
 - name: Test for existence {{{ path }}}
--- a/shared/templates/file_groupowner/ansible.template
+++ b/shared/templates/file_groupowner/ansible.template
@@ -24,6 +24,7 @@
   file:
     path: "{{ item.path }}"
     group: "{{{ FILEGID }}}"
+  when: item.gid != {{{ FILEGID }}}
   with_items:
     - "{{ files_found.files }}"
 
--- a/shared/templates/file_permissions/oval.template
+++ b/shared/templates/file_permissions/oval.template
@@ -46,7 +46,7 @@
 
     {{%- if IS_DIRECTORY %}}
       {{%- if RECURSIVE %}}
-      <unix:path operation="pattern match">{{{ filepath[:-1] }}}</unix:path>
+      <unix:path operation="pattern match">^{{{ filepath[:-1] }}}</unix:path>
       {{%- else %}}
       <unix:path>{{{ filepath[:-1] }}}</unix:path>
       {{%- endif %}}
--- a/shared/templates/file_permissions/oval.template
+++ b/shared/templates/file_permissions/oval.template
@@ -45,14 +45,14 @@
   <unix:file_object comment="{{{ filepath }}}" id="object_file_permissions{{{ FILEID }}}_{{{ loop.index0 }}}" version="1">
 
     {{%- if IS_DIRECTORY %}}
-      {{%- if FILE_REGEX %}}
-      <unix:path>{{{ filepath[:-1] }}}</unix:path>
-      <unix:filename operation="pattern match">{{{ FILE_REGEX[loop.index0] }}}</unix:filename>
-      {{%- elif RECURSIVE %}}
+      {{%- if RECURSIVE %}}
       <unix:path operation="pattern match">{{{ filepath[:-1] }}}</unix:path>
-      <unix:filename xsi:nil="true" />
       {{%- else %}}
       <unix:path>{{{ filepath[:-1] }}}</unix:path>
+      {{%- endif %}}
+      {{%- if FILE_REGEX %}}
+      <unix:filename operation="pattern match">{{{ FILE_REGEX[loop.index0] }}}</unix:filename>
+      {{%- else %}}
       <unix:filename xsi:nil="true" />
       {{%- endif %}}
     {{%- else %}}
--- a/docs/templates/template_reference.md
+++ b/docs/templates/template_reference.md
@@ -259,6 +259,8 @@
 
     -   **filepath** - File path to be checked. If the file path ends
         with `/` it describes a directory. Can also be a list of paths.
+        If **file_regex** is not specified, the rule will only check
+        and remediate directories.
 
     -   **filepath_is_regex** - If set to `"true"` the OVAL will
         consider the value of **filepath** as a regular expression.
@@ -294,6 +296,8 @@ they must be of the same length.
 
     -   **filepath** - File path to be checked. If the file path ends
         with `/` it describes a directory. Can also be a list of paths.
+        If **file_regex** is not specified, the rule will only check
+        and remediate directories.
 
     -   **filepath_is_regex** - If set to `"true"` the OVAL will
         consider the value of **filepath** as a regular expression.
@@ -329,6 +333,8 @@ they must be of the same length.
 
     -   **filepath** - File path to be checked. If the file path ends
         with `/` it describes a directory. Can also be a list of paths.
+        If **file_regex** is not specified, the rule will only check
+        and remediate directories.
 
     -   **filepath_is_regex** - If set to `"true"` the OVAL will
         consider the value of **filepath** as a regular expression.
--- a/linux_os/guide/system/permissions/files/permissions_within_important_dirs/file_ownership_library_dirs/tests/incorrect_owner_within_dir.fail.sh
+++ b/linux_os/guide/system/permissions/files/permissions_within_important_dirs/file_ownership_library_dirs/tests/incorrect_owner_within_dir.fail.sh
@@ -1,0 +1,9 @@
+# platform = multi_platform_sle,multi_platform_rhel,multi_platform_fedora,multi_platform_ubuntu
+
+useradd user_test
+
+TESTDIR="/usr/lib/dir/"
+
+mkdir $TESTDIR
+touch $TESTDIR/test_me
+chown user_test $TESTDIR/test_me
--- a/linux_os/guide/system/permissions/files/permissions_within_important_dirs/file_permissions_library_dirs/rule.yml
+++ b/linux_os/guide/system/permissions/files/permissions_within_important_dirs/file_permissions_library_dirs/rule.yml
@@ -70,5 +70,6 @@ template:
             - /lib64/
             - /usr/lib/
             - /usr/lib64/
+        recursive: 'true'
         file_regex: ^.*$
         filemode: '0755'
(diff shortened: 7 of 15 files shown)


shared/templates/file_permissions/oval.template
```diff
@@ -45,14 +45,14 @@
   <unix:file_object comment="{{{ filepath }}}" id="object_file_permissions{{{ FILEID }}}_{{{ loop.index0 }}}" version="1">
 
     {{%- if IS_DIRECTORY %}}
-      {{%- if FILE_REGEX %}}
-      <unix:path>{{{ filepath[:-1] }}}</unix:path>
-      <unix:filename operation="pattern match">{{{ FILE_REGEX[loop.index0] }}}</unix:filename>
-      {{%- elif RECURSIVE %}}
+      {{%- if RECURSIVE %}}
       <unix:path operation="pattern match">{{{ filepath[:-1] }}}</unix:path>
```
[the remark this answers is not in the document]
yuumasato: This was a question for me too. I tested this and it doesn't match paths like `/tmp/lib', '/somthing/lib`, although they maybe should...?
Only when I do <unix:path operation="pattern match">.*/lib</unix:path> it matches the paths mentioned above.
Maybe this is an unexpected implementation behavior from OpenSCAP?

I can add  '^` anchor in the `path`.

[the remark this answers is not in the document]
yuumasato: @jan-cerny I see in DEVEL logs, :)
```
313878 I: oscap:     I will run file_probe_main: [oscap(1960):probe_worker(7f63487f8700):worker.c:1113:probe_worker]
313879 I: oscap:     The pattern '/lib' doesn't contain a leading caret - added. All paths with the 'pattern match' operation must begin with a caret. [oscap(1960):probe_worker(7f63487f8700):oval_fts.c:538:process_pattern_match]
313880 D: oscap:     Enabling partial match optimization using pattern: '^/lib'. [oscap(1960):probe_worker(7f63487f8700):oval_fts.c:649:process_pattern_match]
```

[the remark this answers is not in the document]
jan-cerny: When I run `oscap` evaluation of the rule in verbose mode I get this interesting info message:

```
I: oscap:     The pattern '/usr/lib' doesn't contain a leading caret - added. All paths with the 'pattern match' operation must begin with a caret.
```

From the code around it seems like our custom optimization, I don't see this in OVAL specification. So I guess it doesn't work like this in other scanners.

So probably if you want to be 100 % right you should add the caret.

yuumasato:
CC @dodys

shared/templates/file_permissions/oval.template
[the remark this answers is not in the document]
jan-cerny: mid-air collision detected!

linux_os/guide/system/permissions/files/permissions_within_important_dirs/file_ownership_library_dirs/rule.yml
```diff
@@ -69,5 +69,6 @@ template:
             - /lib64/
             - /usr/lib/
             - /usr/lib64/
+        recursive: 'true'
```
dodys, line 72: is this really needed or is it an issue related to PR #8194 ?

yuumasato: It is necessary, the check is not going down the directory tree.
Check this test scenario: https://github.com/ComplianceAsCode/content/blob/d5a23b41fad45249f125258225737f96300113fb/linux_os/guide/system/permissions/files/permissions_within_important_dirs/file_ownership_library_dirs/tests/incorrect_owner_within_dir.fail.sh

yuumasato: I think we are looking at the same problem, but proposed different approaches.

shared/templates/file_permissions/oval.template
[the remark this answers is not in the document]
yuumasato: Addresssed in [16f7b22](https://github.com/ComplianceAsCode/content/pull/8404/commits/16f7b2292e7e0b4c79756532116a8c276799864a)

linux_os/guide/system/permissions/files/permissions_within_important_dirs/file_ownership_library_dirs/tests/incorrect_symlink.fail.sh
```diff
@@ -0,0 +1,17 @@
+# platform = multi_platform_sle,multi_platform_rhel,multi_platform_fedora,multi_platform_ubuntu
+
+useradd user_test
+
+TESTDIR="/usr/lib/dir/"
```
Mab879, line 5: This variable appears to be unused.

shared/templates/file_groupowner/ansible.template
```diff
@@ -5,32 +5,41 @@
 # disruption = low
 
 {{% for path in FILEPATH %}}
-{{% if IS_DIRECTORY and FILE_REGEX %}}
+{{% if IS_DIRECTORY %}}
+{{% if FILE_REGEX %}}
+
+- name: Find {{{ path }}} file(s) matching {{{ FILE_REGEX[loop.index0] }}}{{% if RECURSIVE %}} recursively{{% endif %}}
```
Mab879, line 11: Should this also get symlinks? My interruption of the [docs](https://docs.ansible.com/ansible/2.9/modules/find_module.html#parameter-file_type) says that this will not get symlinks.

Mab879 (changes requested):

shared/templates/file_groupowner/ansible.template
yuumasato: Hmm, I'm no sure, but my guess is no.
The state of symlink handkling is a mess and confusing to me.

The OVAL doesn't report symlinks as objects, but I don't know if it will follow them.
The Bash remediation doesn't follow symlinks by default, but we make it follow with `-L`, and we make changes to the symlinks when they are found. (but OVAL doesn't care about them)
The Ansible remediation doesn't follow symlinks by default and we don't make them follow.

And looking at the rule description I'm not sure what is the expectation.
If there is a link from `/usr/lib/` to somewhere else, should we go there, check and remediate stuff?

yuumasato: I think there  a lot of question to answer regarding symlinks:
- Should we check and report the files/dirs they point to?
- Should we check the symlinks themselves?
- Should we follow them?

Besides https://github.com/ComplianceAsCode/content/pull/8404/commits/c715e1239885932960a4848d7bc2152aa0b0eed5, this PR doesn't do much about the symlink situation.
So the situation will stay almost the same... I think this subject should be revisited in a future PR.

Mab879: Sounds good. Agree we can figure out the symlink issue later.

linux_os/guide/system/permissions/files/permissions_within_important_dirs/file_ownership_library_dirs/rule.yml
dodys: maybe I'm confused (I'll be honest, it's some many rules touching permissions that I have to re-read it all the time), but this rule specifically has the file_regex, shouldn't it be working recursively as you added to the documentation?
I ran a test and can see that the bash (not sure on ansible though) is correctly going through all the files. Isn't it just a matter of fixing the oval?

yuumasato: > but this rule specifically has the file_regex, shouldn't it be working recursively as you added to the documentation?

Hmm, my addition in the documentation is about what will be checked, directories vs files. It doesn't address recursion.
a. If `filepath` is a directory and `file_regex` is not specified, the rule will check and remediate the first level of directories in the `filepath`.
b. If `filepath` is a directory and `file_regex` is specified, the rule will check and remediate files in the `filepath`.

If along with case `a.`, recursion is set, the rule will check directories while recursing down.
If along with case `b.`, recursion is set, the rule will check files while recursing down.

> I ran a test and can see that the bash (not sure on ansible though) is correctly going through all the files.

Yes, the remediation is going down the directory tree, but should it happen when I didn't set `recursive`?

> Isn't it just a matter of fixing the oval?

Yes, the OVAL was the part that initially needed fixing, but then I addressed inconsistencies between what the OVAL does and the remediation does.
I think if the check recurses, the remediation should too. But if the OVAL doesn't recurse, the remediations shouldn't recurse.

dodys: > Yes, the remediation is going down the directory tree, but should it happen when I didn't set `recursive`?

hmm got your point now.

I believe we have more rules that will need the recursive flag then, besides the two rules you touched.
I actually need to go through Ubuntu 20.04 STIG and check all the rules and what is needed on which rule.

I also believe that PR #8194 is still needed because we are not dealing with the fact that a filepath might be a symlink

yuumasato: I think we should have more examples in the template documentation, but that should be done in another PR.

yuumasato: > I also believe that PR #8194 is still needed because we are not dealing with the fact that a filepath might be a symlink

Yeah, in RHEL `/lib` is a symlink to `/usr/lib`. 
@dodys Do you mean when the symlink itself is missing, or when the path pointed to by the symlink is missing?
It seems to me that `missing_file_pass` will affect the first case, when the symlink itself is absent.
I think [symlink_test](https://oval-community-guidelines.readthedocs.io/en/latest/oval-schema-documentation/unix-definitions-schema.html?highlight=recurse_direction#symlink-test) will be needed to check if the file pointed to is missing.

Out of a review item with Matthew I filed https://github.com/ComplianceAsCode/content/issues/8412.
The expected behaviour of the round around symlinks is not clear to me.

dodys: > Yeah, in RHEL `/lib` is a symlink to `/usr/lib`. @dodys Do you mean when the symlink itself is missing, or when the path pointed to by the symlink is missing? It seems to me that `missing_file_pass` will affect the first case, when the symlink itself is absent. I think [symlink_test](https://oval-community-guidelines.readthedocs.io/en/latest/oval-schema-documentation/unix-definitions-schema.html?highlight=recurse_direction#symlink-test) will be needed to check if the file pointed to is missing.
> 

No, my point is that with the mentioned commit id, it started to filter out symlinks in the oval. Therefore, if `/lib` is a symlink it won't be checked. If it won't be checked than the number of files evaluated is 0 for that test, thus evaluating to false and the whole rule will fail as the oval is set to all_exist. By passing `missing_file_pass` in the rule, then it changes from `all_exist` to `any_exist` and allow the rule to pass. It is a quick fix. 

> Out of a review item with Matthew I filed #8412. The expected behaviour of the rule around around symlinks is not clear to me.

I don't really remember if most of the permissions/ownership/group ownership rules really mention about symlinks. If oval is set to exclude symlinks (as it is right now), then I don't see the point (and feel free to correct me here) of passing -h in the remediation.

yuumasato: > Therefore, if `/lib` is a symlink it won't be checked. If it won't be checked than the number of files evaluated is 0 for that test, thus evaluating to `false` and the whole rule will fail as the oval is set to `all_exist`.

From what I could test with changes in this PR  (using `--oval-results`), in RHEL, even when `/lib` is a symlink the scanner was able to collect objects.
So actually, could it be that the scanner follows symlinks, but doesn't report them as collected objects, :thinking: ?

Attached is the oval-results file:
[rhel7-oval-results.zip](https://github.com/ComplianceAsCode/content/files/8340665/rhel7-oval-results.zip)

yuumasato: > I don't really remember if most/any of the permissions/ownership/group ownership rules really mention about symlinks. 

I don't think symlinks are mentioned anywhere. 

> If oval is set to exclude symlinks (as it is right now), then I don't see the point (and feel free to correct me here) of passing -h in the remediation.

I see, it is strange to change the owner of the symlink when the check ignores them.
The intent of the change was to avoid dereferencing erors when the symlink is broken. (I made that commit when I was still not aware of the symlink mess)
With `-h` it means that `chown` will not affect the file pointed to, but as the the `find` command doesn't follow symlinks by default ([and `-L` is not set](https://github.com/ComplianceAsCode/content/pull/8404/files#diff-dd85d2fec512415762ebc6bd5d4f47792cea008a4b21940a20514ea3b6a87448R16)), the file pointed to by the symlink wouldn't be changed anyway...but now we avoid dereferencing errors!

And I see inconsistency [here](https://github.com/ComplianceAsCode/content/pull/8404/files#diff-dd85d2fec512415762ebc6bd5d4f47792cea008a4b21940a20514ea3b6a87448R23) too.

To better align with the check the remediation should filter out symlinks before doing `chown`... But then we are not sure if we actually want to filter out the symlinks...

I'm aware that bdc59897f25e4541d9e9bd2db5eb13ef11702252 doesn't fix the symlinking situation, it just avoids errors in `chwon` command with the current behaviour.
This PR was to improve inconsistency with `recurse` and `file_regex` and I stumbled upon `symlink` and `chwon` inconsistency, :joy:

ggbecker (approved):
LGTM

linux_os/guide/system/permissions/files/permissions_within_important_dirs/file_ownership_library_dirs/rule.yml
dodys: I've just checked here my test from yesterday on master,
so `/lib` and `/lib64` are both symlinks to `/usr/lib` and `/usr/lib64` respectively.
One of the rules that I see failing `file_ownership_library_dirs` will fail for `/lib64` as the only file inside it is a .so that is a symlink to a `.so` under `/usr/lib64`.
It doesn't fail for `/lib` as it "finds" some files there, which are actually files under `/usr/lib`.

It is tricky and I need to dig deeper on those rules, just not having the time lately.
I have to test with your changes, it might be that all this recursive and changes might fix some of the weirdness I saw in the failing tests.

yuumasato: Yes, it is tricky. I had to take some time to wrap my head around what the rules do.

Let us know how the tests on your side go.

PR #8416, opened 2022-03-24
Update SRG-OS-000437-GPOS-00194 for RHEL9 STIG

Mab879:
#### Description:

- Update SRG-OS-000437-GPOS-00194 for RHEL9 STIG
- Add fix text to clean_components_post_updating

#### Rationale:
RHEL9 STIG

### DIFF
--- a/linux_os/guide/system/software/updating/clean_components_post_updating/rule.yml
+++ b/linux_os/guide/system/software/updating/clean_components_post_updating/rule.yml
@@ -64,6 +64,8 @@ ocil_clause: |-
     {{% endif %}}
 
 ocil: |-
+    Verify {{{ full_name }}} removes all software components after updated versions have been installed.
+
     {{% if product in ["sle12", "sle15"] %}}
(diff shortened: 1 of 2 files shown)


linux_os/guide/system/software/updating/clean_components_post_updating/rule.yml
```diff
@@ -83,6 +83,12 @@ ocil: |-
     <pre>clean_requirements_on_remove=1</pre>
     {{% endif %}}
 
+{{% if 'rhel' in product %}}
+fix: |-
+    Edit the file "/etc/yum.conf" by adding or editing the following line:
```
jan-cerny, line 88: You should use Jinja variable `pkg_manager_config_file` instead of the specific file path because on RHEL 9 the fix isn't aligned with rule description because on RHEL 9 the config file is `/etc/dnf/dnf.conf`.

jan-cerny (changes requested):

jan-cerny (approved):
Thanks

PR #8420, opened 2022-03-24
File permissions library dirs shall act on files when 'file_regex' is set

yuumasato:
#### Description:

- Change the remediation of templates `file_owner`, `file_groupowner` and `file_permission` to act only on regular files when `file_regex` is set,
  - Following the symlinks under `/lib`, `lib64`, `/usr/lib` and `/usr/lib64` and making changes can have catastrophic consequences. 

#### Rationale:

- Fixes issues after applying remediation with `file_permissions_library_dirs`.
- After #8404, rule `file_permissions_library_dirs` started to recursively navigate down the directory of trees, and it found symlinks to binaries installed in the system:
```
'/lib/.build-id/a4/67cb9c8fa7306d41b96a820b0178f3a9c66055' -> '../../../../usr/bin/passwd'
'/lib/.build-id/a4/8e8ae0d029dbbc1c1b0bb0fcea424860a6c412' -> '../../../../usr/bin/sudo'
'/lib/.build-id/a4/2c53d4543f5c0bb8db47d65e4b766d12f3b7bd' -> '../../../../usr/lib64/python3.9/lib-dynload/_lsprof.cpython-39-x86_64-linux-gnu.so'
```
  This became visible when `sudo` command stopped working after remediation was applied:
```
[admin@localhost ~]# ls -lh /usr/bin/sudo
-rwxr-xr-x. 1 root root 182K Aug 26  2021 /usr/bin/sudo
....
[admin@localhost ~]# sudo ls
sudo: /usr/bin/sudo must be owned by uid 0 and have the setuid bit set
```


yuumasato:
@dodys This reverts the `chmod -h` change, and changes the `find` command to return only regular files.

vojtapolasek (approved):
LGTM, thank you for quick fix

PR #8393, opened 2022-03-21
RHEL7 - exclude the user 'nfsnobody' from accounts_users_home_files_*

yuumasato:
#### Description:

- Adds a filter to exclude the user `nfsnobody` on RHEL7 systems.

#### Rationale:

- Although we already exclude the user with username 'nobody', in some systems (RHEL7) the user 'nobody' has UID 99, and the user 'nfsnobody' has UID 65534.

- Fixes #8352

Interesting reference: https://fedoraproject.org/wiki/Changes/RenameNobodyUser

PR #8426, opened 2022-03-25
Exclude user `nfsnobody` when checking home directories

### LINKED ISSUE #8419: RHEL7 - `file_*_home_directories` rules fail due to `/var/lib/nfs/`
#### Description of problem:
`file_groupownership_home_directories`, `file_ownership_home_directories`, and `file_permissions_home_directories` fails due to wrong permissions on `/var/lib/nfs/` directory. The directory should not be checked at all.

stabilization-v0.1.61 branch

RHEL 7

#### Steps to Reproduce:

1. `python3 tests/test_suite.py profile --libvirt qemu:///system test_suite_vm --datastream /tmp/ssg-rhel7-ds.xml xccdf_org.ssgproject.content_profile_stig_gui`

#### Actual Results:
`file_*_home_directories` fail

#### Expected Results:
`file_*_home_directories` pass

Fix should be similar to #8393 that fixed `accounts_users_home_files_*` rules.

mildas:
#### Description:
Filter user `nfsnobody` on RHEL7 systems.

#### Rationale:
Similar to #8393 fix

Fixes #8419

### DIFF
--- a/linux_os/guide/system/accounts/accounts-session/file_permissions_home_directories/oval/shared.xml
+++ b/linux_os/guide/system/accounts/accounts-session/file_permissions_home_directories/oval/shared.xml
@@ -12,11 +12,20 @@
   <unix:password_object id="object_file_permissions_home_directories_objects" version="1">
     <unix:username datatype="string" operation="not equal">nobody</unix:username>
     <filter action="include">state_file_permissions_home_directories_interactive_uids</filter>
+  {{%- if product == 'rhel7' %}}
+    <filter action="exclude">state_file_permissions_home_files_permissions_nfsnobody</filter>
+  {{%- endif %}}
   </unix:password_object>
 
   <unix:password_state id="state_file_permissions_home_directories_interactive_uids" version="1">
     <unix:user_id datatype="int" operation="greater than or equal">{{{ uid_min }}}</unix:user_id>
   </unix:password_state>
+
+{{%- if product == 'rhel7' %}}
+  <unix:password_state id="state_file_permissions_home_files_permissions_nfsnobody" version="1">
+    <unix:username datatype="string" operation="equals">nfsnobody</unix:username>
+  </unix:password_state>
+{{%- endif %}}
 
   <!-- #### prepare for test_file_permissions_home_directories #### -->
   <local_variable id="var_file_permissions_home_directories_dirs" datatype="string" version="1"
--- a/linux_os/guide/system/accounts/accounts-session/file_permissions_home_directories/oval/shared.xml
+++ b/linux_os/guide/system/accounts/accounts-session/file_permissions_home_directories/oval/shared.xml
@@ -10,22 +10,18 @@
   <!-- For detailed comments about logic used in this OVAL, check the
        "file_ownership_home_directories" rule. -->
   <unix:password_object id="object_file_permissions_home_directories_objects" version="1">
-    <unix:username datatype="string" operation="not equal">nobody</unix:username>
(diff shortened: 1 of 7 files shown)


linux_os/guide/system/accounts/accounts-session/file_permissions_home_directories/oval/shared.xml
```diff
@@ -12,12 +12,21 @@
...
+  {{%- endif %}}
   </unix:password_object>
 
   <unix:password_state id="state_file_permissions_home_directories_interactive_uids" version="1">
     <unix:user_id datatype="int" operation="greater than or equal">{{{ uid_min }}}</unix:user_id>
   </unix:password_state>
 
+{{%- if product == 'rhel7' %}}
+  <unix:password_state id="state_file_permissions_home_files_permissions_nfsnobody" version="1">
+    <unix:username datatype="string" operation="equals">nfsnobody</unix:username>
```
ggbecker, line 26:
```suggestion
    <unix:username datatype="string" operation="equals">^(nobody|nfsnobody)$</unix:username>
```

Can't we do something like this and remove the `nobody` from line 13 and make this filter applicable to all products?

ggbecker: nfsnobody shouldn't be expected to be a regular user even in other platforms, even though it should be technically possible

yuumasato: We can, but I didn't want other products to have more users excluded than necessary.

yuumasato: Hmm, maybe it is fine then.

mildas: What is the conclusion? Should add `nobody` there or not?

ggbecker: Let's do the following:

Line 13:
`<unix:username datatype="string" operation="pattern match">.*</unix:username>`

Then we keep the filter exclude for all platforms but we assign the list of users to a jinja variable ([example](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_kmod/rule.yml#L0-L1)) according to the product, for example:

rhel7 and ol7 - "nobody|nfsnobody"
other products - "nobody"

then

```suggestion
    <unix:username datatype="string" operation="pattern match">^({{{ user_list }}})$</unix:username>
```

yuumasato: > rhel7 and ol7 - "nobody|nfsnobody"
> other products - "nobody"

I think we need achors in the regex, like:
`^(nobody|nfsnobody)$`

yuumasato: Ahh, I see the anchors in the xml now, :)

mildas: Where to define the variable with user list? I don't think `rule.yml` is a good place as it would be basically redundant.

mildas:
I've put the variable to `shared/macros-oval.jinja` because I see it related only to OVAL check.

ggbecker (approved):

PR #8431, opened 2022-03-25
Revert changes from OL8 STIG profile that cause conflict with master

ggbecker:
#### Description:

- Revert changes from OL8 STIG profile that cause conflict with master.


yuumasato (approved):
LGTM, the conflicts are resolved.

PR #8433, opened 2022-03-25
Update template auditd_lineinfile and 3 auditd related rules

Xeicker:
#### Description:

- Update `auditd_lineinfile` to be able to allow more than one value. Take advantage of this in the rules `auditd_data_disk_error_action`, `auditd_data_disk_full_action` and `auditd_name_format`

#### Rationale:

- The STIG Ids asociated with these rules allow more than one value, so OVAL would be giving a false negatives in a few scenarios
- OL08-00-030040 allows "SYSLOG", "SINGLE", or "HALT" as value of disk_error_action
- OL08-00-030060 allows "SYSLOG", "SINGLE", or "HALT" as value of disk_full_action
- OL08-00-030062 allows "hostname", "fqd", or "numeric" as value of name_format


linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_data_disk_full_action/rule.yml
```diff
@@ -56,3 +56,18 @@ ocil: |-
...
+    name: auditd_lineinfile
+    vars:
+        missing_parameter_pass: 'false'
+        parameter: disk_full_action
+        rule_id: {{{ rule_id }}}
+        value: halt
+        value@sle15: syslog
+        value@rhcos4: syslog
+        value_regex@ol8: 'syslog|single|halt'
+        value_regex@rhel8: 'syslog|single|halt'
```
jan-cerny, line 70: The rule is parametrized by variable `var_auditd_disk_full_action` (see https://github.com/ComplianceAsCode/content/blob/da0aba874fc23ce2acdbadbc46ddf93d46d313f3/linux_os/guide/system/auditing/configure_auditd_data_retention/var_auditd_disk_full_action.var). The OVAL check you removed was using this variable. The template doesn't use the variable so the generated OVAL won't use it. At the same time Bash and Ansible will use it. That introduces an inconsistency. Users that will be creating their profiles might change this value and they will wonder why the OVAL check didn't change and isn't affected. So this isn't acceptable for us.

I think that you have 2 options:

1. Don't use the template here, keep a static OVAL file instead.

2. Modify the `auditd_lineinfile` template so that the template could accept a variable. You can get inspired for example by the `grub2_bootloader_argument` template https://github.com/ComplianceAsCode/content/tree/master/shared/templates/grub2_bootloader_argument and for example rule `grub2_slub_debug_argument`.

jan-cerny (changes requested):

linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_data_disk_full_action/rule.yml
Xeicker: I think it is better to use the template, I fixed it so it uses the existing variable if value_regex is not given

linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_data_disk_error_action/rule.yml
```diff
@@ -48,3 +48,16 @@ ocil: |-
     switch to single-user mode, execute a script,
     or halt when the disk errors:
     <pre>disk_error_action single</pre>
+
+template:
+    name: auditd_lineinfile
+    vars:
+        missing_parameter_pass: 'false'
+        parameter: disk_error_action
+        rule_id: {{{ rule_id }}}
```
jan-cerny, line 57: Don't put the rule_id key here. A rule_id variable is available in the template engine by default.

linux_os/guide/system/auditing/configure_auditd_data_retention/auditd_data_disk_error_action/rule.yml
```diff
@@ -48,3 +48,16 @@ ocil: |-
...
+        var_name: var_auditd_disk_error_action
+        value_regex@ol8: 'syslog|single|halt'
+        value_regex@rhel8: 'syslog|single|halt'
```
jan-cerny, line 60: You should really use the variable everywhere. This creates an inconsistency between rule.yml, OVAL and bash remediation on these 2 platforms. You should add these as another options into the linux_os/guide/system/auditing/configure_auditd_data_retention/var_auditd_disk_error_action.var and then assign the options in the profile file.

Then the question is how the template will recognize that the currently chosen variable option is a regular expression.

PR #8434, opened 2022-03-27
Bump Fedora version, rawhide is now 37, drop 33

maage:
#### Description:

Fedora 33 EOL 2021-12-01

#### Rationale:

Support current Fedora release cycle.


jan-cerny (approved):
Thank you!

PR #8435, opened 2022-03-28
Add new template to check kernel build configurations

yuumasato:
#### Description:

- Adds a new template (`kernel_build_config`) that checks the configurations used to build the currently installed kernel.
  - The template checks the `config-$(uname -r)` files available in `/boot/`.
  - No remediation is available as one needs to recompile and install the kernel.
- When `value="n"` the template checks for the absence of the config.
  Unfortunately this makes the test scenarios a bit confusing, as a scenario that is a `pass` for "n", is a `fail` for "y"


yuumasato:
I've made the changes I wanted in 4122a75d68021e7c4672556fb85bd65354a66ee6

shared/templates/kernel_build_config/oval.template
```diff
@@ -0,0 +1,30 @@
...
+  id="test_kernel_{{{ CONFIG}}}" version="1">
+    <ind:object object_ref="object_kernel_{{{ CONFIG }}}" />
+{{% if EXISTENCE != "none_exist" %}}
+    <ind:state state_ref="state_kernel_{{{ CONFIG }}}" />
+{{% endif %}}
+  </ind:textfilecontent54_test>
+
+  <ind:textfilecontent54_object id="object_kernel_{{{ CONFIG }}}" version="1">
+    <ind:filepath operation="pattern match">^/boot/config-.*$</ind:filepath>
+    <ind:pattern operation="pattern match">^{{{ CONFIG | upper }}}=(.*)$</ind:pattern>
```
ggbecker, line 21: The `upper` function should be at least documented. Or maybe, remove it and use directly the input from rule.yml files (with correct upper case already, it should be more transparent).

docs/templates/template_reference.md
```diff
@@ -409,6 +409,19 @@ template:
...
+#### kernel_build_config
+This template checks the configuration used to build the kernel by checking the `/boot/config-*` files.
+The template only generates a check and no remediation.
+
+- Parameters:
+
+    - **config** - The kernel configuration to check
+    - **value** - The value the configuration should have
+      When **value** is `"n"`, the check will pass when the config is absent or
+      commented out in the `/boot/config-*` files.
```
ggbecker, line 421: I prefer having a explicit attribute for this absence condition. For example:

`state: absent` similar as it's done in Ansible. Otherwise it can be very confusing when only looking at the rule.yml

shared/macros.jinja
```diff
@@ -1745,3 +1745,44 @@ Fill in <tt>GRUBENV_FILE_LOCATION</tt> based on information above.
 
     $ systemctl enable --now {{{ service }}}
 {{%- endmacro -%}}
+
+{{#
```
ggbecker, line 1749: attribute `value` is not being used in `kernel_build_config_describe_config` and `kernel_build_config_ocil`. The macros can be clever by using `value` to decide if the configuration should be present or not in the boot config file. Even better if a different attribute is used to decide if the line should be absent or not similarly as suggested in the template data.

shared/templates/kernel_build_config/oval.template
```diff
@@ -0,0 +1,30 @@
+<def-group>
+  <definition class="compliance" id="{{{ _RULE_ID }}}"
+  version="1">
+    {{{ oval_metadata("The kernel " + CONFIG + " should have value" + VALUE, affected_platforms=["multi_platform_all"]) }}}
+    <criteria>
+      <criterion comment="Check build configuration of installed kernels"
+      test_ref="test_kernel_{{{ CONFIG}}}" />
```
ggbecker, line 7:
```suggestion
      test_ref="test_kernel_{{{ CONFIG }}}" />
```

shared/templates/kernel_build_config/oval.template
```diff
@@ -0,0 +1,30 @@
...
+    </criteria>
+  </definition>
+
+  <ind:textfilecontent54_test check="all" check_existence="{{{ EXISTENCE }}}" comment="Check /boot/config-.* files for {{{ CONFIG }}}={{{ VALUE }}}"
+  id="test_kernel_{{{ CONFIG}}}" version="1">
```
ggbecker, line 12:
```suggestion
  id="test_kernel_{{{ CONFIG }}}" version="1">
```

shared/macros.jinja
```diff
@@ -1745,3 +1745,44 @@ Fill in <tt>GRUBENV_FILE_LOCATION</tt> based on information above.
...
+
+:param config: The kernel config parameter
+:type config: str
+:param value: The value for the given config
+:type value: str
+
+#}}
+{{% macro kernel_build_config_ocil(config, value) -%}}
+    To determine the config value the kernel was built with, run the following command:
+    <pre>$ grep {{{ config | upper}}} /boot/config.*</pre>
```
ggbecker, line 1787:
```suggestion
    <pre>$ grep {{{ config | upper }}} /boot/config.*</pre>
```

shared/templates/kernel_build_config/oval.template
ggbecker: If using the upper case, you can do the reverse by using `lower` function in the identifiers or have the template.py create a new attribute coming from `config` with `lower`

docs/templates/template_reference.md
```diff
@@ -409,6 +409,19 @@ template:
     arg_name: audit
 ```
 
+#### kernel_build_config
+This template checks the configuration used to build the kernel by checking the `/boot/config-*` files.
+The template only generates a check and no remediation.
```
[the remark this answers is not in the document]
yuumasato: My intention was to say that there won't be remediations ever for this template. I'll make it more explicit that there re compiling the kernel is the only way.

shared/templates/kernel_build_config/oval.template
yuumasato: I lean towards having the input in the rule lower case format and make all capital letters when needed. It is annoying to write with all caps.

yuumasato: Hmm, anyway, input in upper case seems more natural for these configs.

docs/templates/template_reference.md
yuumasato: I'm afraid that having two tightly coupled parameters (`value` and `state`) will be confusing.
Hmm, basically when `value` is `n` the `state` has to be `absent`, and vice-versa. 

https://unix.stackexchange.com/questions/305540/should-settings-missing-in-a-kernels-config-be-interpreted-as-setting-n-or

yuumasato: I lean towards keeping just `value`, and the check does whatever it needs to do.

docs/templates/template_reference.md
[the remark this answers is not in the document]
yuumasato: Updated.

shared/macros.jinja
```diff
@@ -1745,3 +1745,44 @@ Fill in <tt>GRUBENV_FILE_LOCATION</tt> based on information above.
...
+:type config: str
+:param value: The value for the given config
+:type value: str
+
+#}}
+{{% macro kernel_build_config_describe_config(config, value) -%}}
+    The configuration that was used to build kernel is available at <tt>/boot/config-*</tt>.
+    To check the configuration value for <tt>{{{ config }}}</tt>, run the following command:
+    <tt>grep {{{ config }}} /boot/config-*</tt>
+{{%- endmacro %}}
```
ggbecker, line 1762: I still think we should update these two macros to handle the case where value is set to `n`. For example,

```suggestion
{{% macro kernel_build_config_describe_config(config, value) -%}}
    The configuration that was used to build kernel is available at <tt>/boot/config-*</tt>.
    To check the configuration value for <tt>{{{ config }}}</tt>, run the following command:
    <tt>grep {{{ config }}} /boot/config-*</tt>
    
    {{% if value == 'n' %}}
    The result should return empty or a commented line
    {{% endif %}}}
{{%- endmacro %}}
```

ggbecker (commented):
is it the case that these new rules are applicable to `machine` only?

shared/macros.jinja
yuumasato: Forgot about that

yuumasato: Updated both macros in [10017fb](https://github.com/ComplianceAsCode/content/pull/8435/commits/10017fb29dae9bd3958ac786a426d9f83682f2b0)

shared/templates/kernel_build_config/template.py
```diff
@@ -0,0 +1,6 @@
+def preprocess(data, lang):
+    if data['value'] == 'n':
+        # A kernel configuration that is not set is not present in the config file
+        # or is commented out
+        data['shoud_check_absence'] = True
```
ggbecker, line 5:
```suggestion
        data['should_check_absence'] = True
```

shared/templates/kernel_build_config/tests/fail1.fail.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
ggbecker, line 1:
```suggestion
#!/bin/bash
# remediation = none
```

shared/templates/kernel_build_config/tests/fail2.fail.sh
```diff
@@ -0,0 +1,20 @@
+# remediation = none
+#!/bin/bash
```
ggbecker, line 2:
```suggestion
#!/bin/bash
# remediation = none
```

nitpick

ggbecker (changes requested):
The test shared/templates/kernel_build_config/tests/fail2.fail.sh is failing for rule `kernel_config_debug_fs`. It's passing instead of failing.

shared/templates/kernel_build_config/oval.template
```diff
@@ -0,0 +1,30 @@
...
+    <ind:filepath operation="pattern match">^/boot/config-.*$</ind:filepath>
+    <ind:pattern operation="pattern match">^{{{ CONFIG }}}="?{{{ VALUE }}}"?$</ind:pattern>
+    <ind:instance datatype="int" operation="greater than or equal">1</ind:instance>
+  </ind:textfilecontent54_object>
+
+{{% if SHOULD_CHECK_ABSENCE %}}
+  <ind:textfilecontent54_test check="all" check_existence="none_exist" comment="Check /boot/config-.* files for {{{ CONFIG }}}={{{ VALUE }}}"
+  id="test_kernel_{{{ CONFIG | lower }}}_absence" version="1">
+    <ind:object object_ref="object_kernel_{{{ CONFIG | lower }}}" />
+  </ind:textfilecontent54_test>
```
ggbecker, line 28: This test should check if a parameter is not configured in any way. You can't reuse the object above since it looks for the whole line `kernel_config=n` for example.

This test needs to check if the `config` is absent/commented out.

I believe this was not manifesting before because of the `shoud` typo.

ggbecker (approved):

PR #8438, opened 2022-03-28
Update grub2_uefi_password & grub2_uefi_admin_username

Xeicker:
#### Description:

- Removed for all products the criterion that allows OVAL to pass if <grub2_boot_path>/grub.cfg is missing. And remove unnecessary test accordingly.
- Update OVAL criteria for OL8 and RHEL8 to only look for the GRUB2_PASSWORD configuration in the file <grub2_boot_path>/user.cfg
- Set grub2_uefi_boot_path in OL8 product.yml

#### Rationale:

- The absence of grub.cfg only make sense if grub is not installed, this situation is already managed by the grub CPE
- For grub2_uefi_password:
  - It is better to check only the user.cfg file since this is the one generated by the command grub2-setpassword
  - The existence and validity of superusers is already addressed by Update grub2_uefi_admin_username and it is not required in the STIG IDs that grub2_uefi_password covers
-  The default grub2_uefi_boot_path was wrong

### DIFF
--- a/linux_os/guide/system/bootloader-grub2/uefi/grub2_uefi_password/oval/shared.xml
+++ b/linux_os/guide/system/bootloader-grub2/uefi/grub2_uefi_password/oval/shared.xml
@@ -6,7 +6,6 @@
       {{% if product in ["ol8", "rhel8"] %}}
       <criterion comment="make sure a password is defined in {{{ grub2_uefi_boot_path }}}/user.cfg" test_ref="test_grub2_uefi_password_usercfg" />
       {{% else %}}
-      {{{ oval_file_absent_criterion(grub2_uefi_boot_path + "/grub.cfg") }}}
       <criteria operator="AND">
         <criteria comment="check both files to account for procedure change in documenation" operator="OR">
           <criterion comment="make sure a password is defined in {{{ grub2_uefi_boot_path }}}/user.cfg" test_ref="test_grub2_uefi_password_usercfg" />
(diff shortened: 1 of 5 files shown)


linux_os/guide/system/bootloader-grub2/uefi/grub2_uefi_password/oval/shared.xml
```diff
@@ -3,19 +3,20 @@
     {{{ oval_metadata("The UEFI grub2 boot loader should have password protection enabled.") }}}
 
     <criteria operator="OR">
-      {{{ oval_file_absent_criterion(grub2_uefi_boot_path + "/grub.cfg") }}}
+      {{% if product in ["ol8", "rhel8"] %}}
```
jan-cerny, line 6: Why is RHEL 8 and OL 8 different? Does it apply to other Linux distributions as well?

Xeicker: I hadn't reviewed other distros. This change is based on DISA's STIG profile documentation, and apparently this also applies to el7. Do you want me to make the change to apply to all products?

This rule is used only in STIG profiles in rhel and ol

jan-cerny: OK, thanks for reply. When I read the rationale that the criteria is now handled by applicability check (grub cpe) I think that this rationale would apply for any distribution. So please remove the conditions and make the change apply on all products. As you say the rule is used only in RHEL and OL STIG I feel that it's a conservative change.

jan-cerny (approved):
Thanks!

PR #8443, opened 2022-03-28
Update regex to detect nfs filesystems in fstab

Xeicker:
#### Description:

- Update regex in mount_option_remote_filesystems template OVAL

#### Rationale:

- Regex didn't match mounted filesystems with UUID= or LABEL= method.
Also needed to add the character \ to allow escaped fs files


jan-cerny (changes requested):
In my opinion this is again something that should require a test scenario, otherwise somebody else in future might change the regex and break your use case accidentally.

jan-cerny (approved):
LGTM, Thanks for adding the test.

PR #8449, opened 2022-03-28
Introduce CPE for tftp-server

Xeicker:
#### Description:

- Introduce CPE for tftp-server
- Apply the CPE to rule tftpd_uses_secure_mode

#### Rationale:

- This because DISA requires this rule to not be applicable when tftp-server is not installed. This in
  - OL08-00-040350
  - RHEL-08-040350
  - OL07-00-040720
  - RHEL-07-040720


jan-cerny (approved):
Amazing, works for me. Thanks!

PR #8450, opened 2022-03-28
Update accounts_password template OVAL for ol8

Xeicker:
#### Description:

- Update accounts_password OVAL template for OL8 so it looks for the configuration also in /etc/security/pwquality.conf.d/*.conf

#### Rationale:

- Ol8 needs to check /etc/security/pwquality.conf.d/*.conf files also, as states the STIG IDs:
  * OL08-00-020300
  * OL08-00-020280
  * OL08-00-020230
  * OL08-00-020170
  * OL08-00-020160
  * OL08-00-020150
  * OL08-00-020140
  * OL08-00-020130
  * OL08-00-020120
  * OL08-00-020110


jan-cerny (approved):

ggbecker:
This change has high probability to be ported to RHEL as well even though it's not explicitly mentioned in RHEL STIG.

PR #8451, opened 2022-03-28
Update configure_ssh_crypto_policy regex

Xeicker:
#### Description:

- Changed regexes in configure_ssh_crypto_policy rule so it is case insensitive

#### Rationale:

- DISA, in STIG ID OL08-00-010287, mentions this conf in lower case. Also sshd related man pages mention case inensitiveness

### DIFF
--- a/linux_os/guide/system/software/integrity/crypto/configure_ssh_crypto_policy/ansible/shared.yml
+++ b/linux_os/guide/system/software/integrity/crypto/configure_ssh_crypto_policy/ansible/shared.yml
@@ -8,4 +8,4 @@
   lineinfile:
     dest: /etc/sysconfig/sshd
     state: absent
-    regexp: ^\s*CRYPTO_POLICY.*$
+    regexp: ^\s*(?i)CRYPTO_POLICY.*$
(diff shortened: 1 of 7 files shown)


jan-cerny (changes requested):
Please add a test scenario for this rule that will cover a case insensitive option.

jan-cerny (changes requested):
Thanks for adding the test scenarios! It was surely a good investment because it uncovered that you need to change the Ansible remediation as well.

You can do something like this:
```
--- a/linux_os/guide/system/software/integrity/crypto/configure_ssh_crypto_policy/ansible/shared.yml
+++ b/linux_os/guide/system/software/integrity/crypto/configure_ssh_crypto_policy/ansible/shared.yml
@@ -8,4 +8,4 @@
   lineinfile:
     dest: /etc/sysconfig/sshd
     state: absent
-    regexp: ^\s*CRYPTO_POLICY.*$
+    regexp: '^(?i)\s*CRYPTO_POLICY.*$'
```

jan-cerny (approved):

PR #8454, opened 2022-03-29
Optimize library dirs rules

yuumasato:
#### Description:

- Ditches use of Ansible `find` module in favor of the CLI utility `find`.
  - Main advantage of CLI `find` is that we can filter out the compliant files.
- This PR also expands `allow_stricter_permissions` to the remediations.
  - This should speed up even more the Ansible and Bash remediations.
  - Most of the files in the library dirs have permissions stricter (0644) than the rule threshold (0755).
 
#### Rationale:

- The Ansible remediation is taking hours to go through the library dirs because the Ansible `find` module returns a list of all files for the `file` module to iterate. With CLI `find` command, the `file` module only iterates through the files that need fixing.

#### Notes::
- If this approach looks good, I'll remove Ansible `find` from `file_owner` and `file_groupowner` too.

### DIFF
--- a/shared/templates/file_permissions/template.py
+++ b/shared/templates/file_permissions/template.py
@@ -78,11 +78,23 @@ def preprocess(data, lang):
             mode_int = mode_int >> 1
 
         search_mode = ''
+        fix_mode = ''
         for k in mode_dict:
             if mode_dict[k] != '':
                 if search_mode != '':
                     search_mode += ','
                 search_mode += "{}+{}".format(k, mode_dict[k])
 
+                if data['allow_stricter_permissions']:
+                    if fix_mode != '':
+                        fix_mode += ','
+                    fix_mode += "{}-{}".format(k, mode_dict[k])
+
         data["search_mode"] = search_mode
+        if data['allow_stricter_permissions']:
+            # overwrite "filemode" with a subtractive symbolic mode
+            # e.g., if filemode was "0755", now it is "u-s,g-ws,o-wt"
+            # This allows files to keep the permission in which they are already stricter
+            data["filemode"] = fix_mode
+
     return data
--- a/shared/templates/file_permissions/template.py
+++ b/shared/templates/file_permissions/template.py
@@ -58,4 +58,31 @@ def preprocess(data, lang):
                     + field + ">\n" + mode_str)
             mode_int = mode_int >> 1
         data["statemode"] = mode_str.rstrip("\n")
+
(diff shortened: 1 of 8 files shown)


shared/templates/file_permissions/template.py
```diff
@@ -58,4 +58,43 @@ def preprocess(data, lang):
...
+        for k in mode_dict:
+            if mode_dict[k] != '':
+                if search_mode != '':
+                    search_mode += ','
+                search_mode += "{}+{}".format(k, mode_dict[k])
+
+                if data['allow_stricter_permissions']:
+                    if fix_mode != '':
+                        fix_mode += ','
+                    fix_mode += "{}-{}".format(k, mode_dict[k])
```
ggbecker, line 91: This definitely needs to be refactored to multiple functions as they will be used by other templates as well

ggbecker (commented):
overall it looks good to me. let's see what tests have to say about it

shared/templates/file_permissions/template.py
yuumasato: Could you give example of templates that could benefit from this?

ggbecker:
Shellcheck test revealed this type of problem:

```
In /__w/content/content/build/ubuntu2004/fixes/bash/file_permissions_sshd_private_key.sh line 9:
find -L /etc/ssh/ -maxdepth 1 -perm /u+xs,g+xwrs,o+xwrt -type f -regex ^.*_key$ -exec chmod u-xs,g-xwrs,o-xwrt {} \;
                                                                       ^------^ SC2061: Quote the parameter to -regex so the shell won't interpret it.
```

yuumasato:
> Shellcheck test revealed this type of problem:
> 
> ```
> In /__w/content/content/build/ubuntu2004/fixes/bash/file_permissions_sshd_private_key.sh line 9:
> find -L /etc/ssh/ -maxdepth 1 -perm /u+xs,g+xwrs,o+xwrt -type f -regex ^.*_key$ -exec chmod u-xs,g-xwrs,o-xwrt {} \;
>                                                                        ^------^ SC2061: Quote the parameter to -regex so the shell won't interpret it.
> ```

Did it for Ansible, but forgot for Bash, :)

shared/templates/file_permissions/template.py
yuumasato: Should be split now.

yuumasato:
All templates have been updated.

ggbecker (approved):

PR #8455, opened 2022-03-29
Add utils/convert_srg_export_to_xlsx.py

Mab879:
#### Description:

This script creates an XLSX spreadsheet for DISA submission.

#### Rationale:
Less manual work

### DIFF
--- a/utils/convert_srg_export_to_xlsx.py
+++ b/utils/convert_srg_export_to_xlsx.py
@@ -57,7 +57,7 @@ def setup_headers(sheet: openpyxl.worksheet.worksheet.Worksheet) -> None:
 def format_cells(sheet: openpyxl.worksheet.worksheet.Worksheet):
     for row in sheet.iter_rows():
         for cell in row:
-            cell.alignment = cell.alignment = Alignment(wrap_text=True, vertical='top')
+            cell.alignment = Alignment(wrap_text=True, vertical='top')
 
 
 def setup_row(sheet: openpyxl.worksheet.worksheet.Worksheet, row: dict, row_num: int) -> None:
@@ -78,7 +78,7 @@ def setup_row(sheet: openpyxl.worksheet.worksheet.Worksheet, row: dict, row_num:
     sheet[f'O{row_num}'] = row['Mitigation']
     sheet[f'P{row_num}'] = row['Artifact Description']
     sheet[f'Q{row_num}'] = row['Status Justification']
-    sheet.row_dimensions[row_num].height = 42
+    sheet.row_dimensions[row_num].height = 130
 
     # freeze header row represented by A1
     # A2 is required because it freezes everything before it
--- a/utils/convert_srg_export_to_xlsx.py
+++ b/utils/convert_srg_export_to_xlsx.py
@@ -54,7 +54,7 @@ def setup_headers(sheet: openpyxl.worksheet.worksheet.Worksheet) -> None:
         cell.font = Font(bold=True, name='Calibri')
 
 
-def format_cells(sheet: openpyxl.worksheet.worksheet.Worksheet, row_num: int):
+def format_cells(sheet: openpyxl.worksheet.worksheet.Worksheet):
     for row in sheet.iter_rows():
         for cell in row:
             cell.alignment = cell.alignment = Alignment(wrap_text=True, vertical='top')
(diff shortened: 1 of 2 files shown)


utils/convert_srg_export_to_xlsx.py
```diff
@@ -0,0 +1,113 @@
...
+    sheet[f'J{row_num}'] = row['SRG Check']
+    sheet[f'K{row_num}'] = row['Check']
+    sheet[f'L{row_num}'] = row['SRG Fix']
+    sheet[f'M{row_num}'] = row['Fix']
+    sheet[f'N{row_num}'] = row['Severity']
+    sheet[f'O{row_num}'] = row['Mitigation']
+    sheet[f'P{row_num}'] = row['Artifact Description']
+    sheet[f'Q{row_num}'] = row['Status Justification']
+    sheet.row_dimensions[row_num].height = 42
+
```
ggbecker, line 82:
```suggestion

    # freeze header row represented by A1
    # A2 is required because it freezes everything before it
    # in this case we only want A1 row to be frozen
    sheet.freeze_panes = "A2"
```

utils/convert_srg_export_to_xlsx.py
```diff
@@ -0,0 +1,113 @@
...
+    for row in sheet.iter_rows():
+        for cell in row:
+            cell.alignment = cell.alignment = Alignment(wrap_text=True, vertical='top')
+
+
+def setup_row(sheet: openpyxl.worksheet.worksheet.Worksheet, row: dict, row_num: int) -> None:
+    sheet[f'A{row_num}'] = row['IA Control']
+    sheet[f'B{row_num}'] = row['CCI']
+    sheet[f'C{row_num}'] = row['SRGID']
+    sheet[f'D{row_num}'] = row['STIGID']
```
ggbecker, line 67:
```suggestion
    sheet[f'D{row_num}'] = row.get('STIGID', '')
```

In my tests it complained row didn't have `STIGID`. Maybe we should reconsider how this is populated. Using `get` is one of the solutions.

Mab879: I have rebased on master, which now has this field in the export.

ggbecker (changes requested):
Doesn't it make sense to embed this into the script that generates the csv file?

We can then make that script to output different formats, such as csv, xlsx, html, json, etc.

utils/convert_srg_export_to_xlsx.py
```diff
@@ -0,0 +1,118 @@
...
+                        help=f"Path to SSG root directory (defaults to {SSG_ROOT})")
+    parser.add_argument('input', type=str)
+    return parser.parse_args()
+
+
+def main():
+    args = parse_args()
+    xlsx = openpyxl.Workbook()
+    sheet = xlsx.active
+    sheet.name = 'Sheet0'
```
ggbecker, line 104: Let's add something meaningful here. For example, `<Product> SRG mapping`, then product can be taken from the user input. Probably doable when this script gets integrated to the srg_export.py which has the product as input already.

utils/convert_srg_export_to_xlsx.py
```diff
@@ -0,0 +1,103 @@
...
+    sheet['P1'] = 'Artifact Description'
+    sheet['Q1'] = 'Status Justification'
+    for cell in list(sheet.iter_rows(max_row=1))[0]:
+        cell.font = Font(bold=True, name='Calibri')
+
+
+def format_cells(sheet: openpyxl.worksheet.worksheet.Worksheet):
+    for row in sheet.iter_rows():
+        for cell in row:
+            cell.alignment = cell.alignment = Alignment(wrap_text=True, vertical='top')
```
ggbecker, line 60:
```suggestion
            cell.alignment = Alignment(wrap_text=True, vertical='top')
```

utils/convert_srg_export_to_xlsx.py
```diff
@@ -0,0 +1,103 @@
...
+    sheet[f'I{row_num}'] = row['Status']
+    sheet[f'J{row_num}'] = row['SRG Check']
+    sheet[f'K{row_num}'] = row['Check']
+    sheet[f'L{row_num}'] = row['SRG Fix']
+    sheet[f'M{row_num}'] = row['Fix']
+    sheet[f'N{row_num}'] = row['Severity']
+    sheet[f'O{row_num}'] = row['Mitigation']
+    sheet[f'P{row_num}'] = row['Artifact Description']
+    sheet[f'Q{row_num}'] = row['Status Justification']
+    sheet.row_dimensions[row_num].height = 42
```
ggbecker, line 81:
```suggestion
    sheet.row_dimensions[row_num].height = 130
```

ggbecker (changes requested):

ggbecker (approved):

PR #8456, opened 2022-03-29
Performance improvements for file permission and ownership templates

### LINKED ISSUE #8417: RHEL9 - `file_permissions_library_dirs` passes after remediation, but fails after next reboot
#### Description of problem:
In RHEL9 STIG without GUI, `file_permissions_library_dirs` fails after machine reboot. 

stabilization-v0.1.61 branch

RHEL 9

#### Steps to Reproduce:

1. `python3 tests/test_suite.py profile --libvirt qemu:///system test_suite_vm --datastream /tmp/ssg-rhel9-ds.xml xccdf_org.ssgproject.content_profile_stig`

#### Actual Results:
xccdf_org.ssgproject.content_rule_file_permissions_library_dirs - fail 

#### Expected Results:
xccdf_org.ssgproject.content_rule_file_permissions_library_dirs - pass

OVAL details:
<details>
<h4><span class="label label-primary">Testing mode of /lib/</span> 
        <span class="label label-default">oval:ssg-test_file_permissions_library_dirs_0:tst:1</span> 
        <span class="label label-success">true</span></h4><h5>Following items have been found on the system:</h5>

Path | Type | UID | GID | Size (B) | Permissions
-- | -- | -- | -- | -- | --
/lib/polkit-1/polkit-agent-helper-1 | regular | 0 | 0 | 20416 | rwsr-xr-x

<h4><span class="label label-primary">Testing mode of /lib64/</span> 
        <span class="label label-default">oval:ssg-test_file_permissions_library_dirs_1:tst:1</span> 
        <span class="label label-danger">false</span></h4><h5>No items have been found conforming to the following objects:</h5><h5>Object <strong><abbr title="/lib64/">oval:ssg-object_file_permissions_library_dirs_1:obj:1</abbr></strong> of type
                <strong>file_object</strong></h5>

Path | Filename | Filter | Filter
-- | -- | -- | --
^/lib64 | ^.*$ | oval:ssg-state_file_permissions_library_dirs_1_mode_not_0755:ste:1 | oval:ssg-exclude_symlinks__library_dirs:ste:1

<h4><span class="label label-primary">Testing mode of /usr/lib/</span> 
        <span class="label label-default">oval:ssg-test_file_permissions_library_dirs_2:tst:1</span> 
        <span class="label label-success">true</span></h4><h5>Following items have been found on the system:</h5>

Path | Type | UID | GID | Size (B) | Permissions
-- | -- | -- | -- | -- | --
/usr/lib/polkit-1/polkit-agent-helper-1 | regular | 0 | 0 | 20416 | rwsr-xr-x

<h4><span class="label label-primary">Testing mode of /usr/lib64/</span> 
        <span class="label label-default">oval:ssg-test_file_permissions_library_dirs_3:tst:1</span> 
        <span class="label label-danger">false</span></h4><h5>No items have been found conforming to the following objects:</h5><h5>Object <strong><abbr title="/usr/lib64/">oval:ssg-object_file_permissions_library_dirs_3:obj:1</abbr></strong> of type
                <strong>file_object</strong></h5>

Path | Filename | Filter | Filter
-- | -- | -- | --
^/usr/lib64 | ^.*$ | oval:ssg-state_file_permissions_library_dirs_3_mode_not_0755:ste:1 | oval:ssg-exclude_symlinks__library_dirs:ste:1
</details>

Check the OVAL results. It seems that `polkit` permissions are causing the failure.

marcusburghardt:
These templates used to hit performance issues when directories were
recursively assessed. The OVAL logic was slightly changed to improve the
performance around 50% or more for some rules.

- Fixes #8417

### DIFF
--- a/shared/templates/file_groupowner/oval.template
+++ b/shared/templates/file_groupowner/oval.template
@@ -13,23 +13,12 @@
    {{% endif %}}
     </criteria>
   </definition>
-  {{%- if MISSING_FILE_PASS -%}}
-    {{# Any number of files can exist, from zero to any #}}
-    {{% set FILE_EXISTENCE = "any_exist" %}}
-  {{%- else -%}}
-    {{# All defined files must exist. When using regex, at least one file must match #}}
-    {{% set FILE_EXISTENCE = "all_exist" %}}
-  {{%- endif -%}}
-
 
   {{% for filepath in FILEPATH %}}
-  <unix:file_test check="all" check_existence="{{{ FILE_EXISTENCE }}}" comment="Testing group ownership of {{{ filepath }}}" id="test_file_groupowner{{{ FILEID }}}_{{{ loop.index0 }}}" version="1">
+  <unix:file_test check="all" check_existence="none_exist" comment="Testing group ownership of {{{ filepath }}}" id="test_file_groupowner{{{ FILEID }}}_{{{ loop.index0 }}}" version="1">
     <unix:object object_ref="object_file_groupowner{{{ FILEID }}}_{{{ loop.index0 }}}" />
-    <unix:state state_ref="state_file_groupowner{{{ FILEID }}}_gid_{{{ FILEGID }}}_{{{ loop.index0 }}}" />
   </unix:file_test>
-  <unix:file_state id="state_file_groupowner{{{ FILEID }}}_gid_{{{ FILEGID }}}_{{{ loop.index0 }}}" version="1">
-    <unix:group_id datatype="int">{{{ FILEGID }}}</unix:group_id>
-  </unix:file_state>
+
   <unix:file_object comment="{{{ filepath }}}" id="object_file_groupowner{{{ FILEID }}}_{{{ loop.index0 }}}" version="1">
     {{%- if IS_DIRECTORY -%}}
       {{%- if RECURSIVE %}}
@@ -46,8 +35,14 @@
       <unix:filepath{{% if FILEPATH_IS_REGEX %}} operation="pattern match"{{% endif %}}>{{{ filepath }}}</unix:filepath>
     {{%- endif %}}
     <filter action="exclude">symlink_file_groupowner{{{ FILEID }}}_uid_{{{ FILEGID }}}</filter>
+    <filter action="exclude">state_file_groupowner{{{ FILEID }}}_gid_{{{ FILEGID }}}_{{{ loop.index0 }}}</filter>
   </unix:file_object>
+
+  <unix:file_state id="state_file_groupowner{{{ FILEID }}}_gid_{{{ FILEGID }}}_{{{ loop.index0 }}}" version="1">
+    <unix:group_id datatype="int">{{{ FILEGID }}}</unix:group_id>
+  </unix:file_state>
   {{% endfor %}}
+
   <unix:file_state id="symlink_file_groupowner{{{ FILEID }}}_uid_{{{ FILEGID }}}" version="1">
     <unix:type operation="equals">symbolic link</unix:type>
   </unix:file_state>
--- a/shared/templates/file_permissions/oval.template
+++ b/shared/templates/file_permissions/oval.template
@@ -1,19 +1,3 @@
-{{%- if ALLOW_STRICTER_PERMISSIONS %}}
-    {{#
-        When allowing stricter permissions, the check looks for files/directories
-        that have any extra permission bit set. The permissions mode (STATEMODE)
-        is negated (inverted) and becomes a mask for permissions that should not be set.
-    #}}
-  {{% set FILE_EXISTENCE = "at_least_one_exists" %}}
-{{%- else -%}}
-  {{%- if MISSING_FILE_PASS -%}}
-    {{# Any number of files can exist, from zero to any #}}
-    {{% set FILE_EXISTENCE = "any_exist" %}}
-  {{%- else -%}}
-    {{# All defined files must exist. When using regex, at least one file must match #}}
-    {{% set FILE_EXISTENCE = "all_exist" %}}
-  {{%- endif -%}}
-{{%- endif -%}}
 <def-group>
   <definition class="compliance" id="{{{ _RULE_ID }}}" version="1">
   {{% if FILEPATH is not string %}}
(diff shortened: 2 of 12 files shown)


shared/templates/file_groupowner/oval.template
```diff
@@ -13,23 +13,12 @@
    {{% endif %}}
     </criteria>
   </definition>
-  {{%- if MISSING_FILE_PASS -%}}
```
jan-cerny, line 16: This OVAL template now doesn't contain this parameter at all. How does it work now?

The docs says:

```
 -   **missing_file_pass** - If set to `"true"` the OVAL check will 
     pass when file is absent. Default value is `"false"`.
```

I have an impression that it now doesn't work when this is set to `false`. When it's set to `false` and the given file doesn't exist the rule should fail. But with this PR change if the given file doesn't exist, the `check_existence="none_exists"` is satisfied, therefore the rule will pass. Is my understanding correct?

Now the question is how to solve this change. I assume that the `missing_file_pass` set to `false` makes sense only when looking for specific files and not for recursive search. For the recursive search (i.e. rules with `recursive: true`) we usually don't care if something exists. But, our problem is mainly for recursively searching rules because in these rules we have the problem of collecting large amounts of item. We don't need this optimization for rules that match a single file or a couple of file in the same directory. Therefore we could have 2 versions of the OVAL code: one for recursive searching with the filter optimization and one for the other cases. We will document that the `missing_file_pass = false` is mutually exclusive with `recursive = true`. WDYT?

jan-cerny:
Measuring the time using ssg test suite isn't the best approach. When we discussed that we were concerned of 2 specific things:
* size of the generated OVAL results 
* time of the XML validation

mildas:
E8 profile
Stabilization branch - 94MB OVAL results, XML validation 7.5h
This PR - 32MB OVAL results, XML validation 24min

It is much better now. Not so great as it was in previous release - 7MB results and <1min validation - but I don't think these old values are realistic anymore.

marcusburghardt:
> Measuring the time using ssg test suite isn't the best approach. When we discussed that we were concerned of 2 specific things:
> 
> * size of the generated OVAL results
> * time of the XML validation

The size of the OVAL results generated for these affected templates directly reflects the total time, both to generate the file and to process the XML. Therefore, I preferred to report the total time here. But in fact, XML file sizes have also been reduced following a direct proportion of the total time which, according to my local tests during the development, varies for each rule.

shared/templates/file_groupowner/oval.template
marcusburghardt: We should remove this parameter. I didn't propose to remove from documentation because it should also be removed from many `rule.yml` files and I don't think this is mandatory for now since a behavior change is not expected. We can gradually deprecate this parameter for these affected templates.

These templates intend to assess the permissions and ownership (uid and gid) of files. Doesn't make any sense to me this assessment for files which are not present. If a file actually doesn't exist, it is quite logical that it won't have problems with permissions and ownership. The idea of the parameter `missing_file_pass` only makes sense for templates and rules which actually want to asses file existence and not their properties, like `file_existence` template or the `banner_etc_motd` rule, for example.

Also, seems that this parameter was introduced in these templates due to the design of the old OVAL logic where an empty file_object would affect the results based on the `check_existence=` parameter. In other words, this parameter is useless for these templates if we used `check_existence=none_exist` for the `file_test`.

yuumasato: I agree with Jan here, the rule passing when a specific file is absent is a behaviour changes we do not want.
We have a specific parameter that allows this behaviour, and I think we should keep it, (if we want to remove it, it should be done in master). 

> Doesn't make any sense to me this assessment for files which are not present.

I think it depends on the expectation of the rule. There may be rules that don't care if it is missing, while others may care if the file is missing.
From my PoV, this is a flexibility of the template that we want to keep.

shared/templates/file_groupowner/oval.template
```diff
@@ -13,23 +13,12 @@
...
-    {{# Any number of files can exist, from zero to any #}}
-    {{% set FILE_EXISTENCE = "any_exist" %}}
-  {{%- else -%}}
-    {{# All defined files must exist. When using regex, at least one file must match #}}
-    {{% set FILE_EXISTENCE = "all_exist" %}}
-  {{%- endif -%}}
-
 
   {{% for filepath in FILEPATH %}}
-  <unix:file_test check="all" check_existence="{{{ FILE_EXISTENCE }}}" comment="Testing group ownership of {{{ filepath }}}" id="test_file_groupowner{{{ FILEID }}}_{{{ loop.index0 }}}" version="1">
```
yuumasato, line 26: What we can do to allow the `missing_file_pass` behavior is to keep the Jinja conditionals above, so that `FILE_EXISTENCE = 'any_exist` when `MISSING_FILE_PASS` is true.

And wrap the state `state_file_groupowner{{{ FILEID }}}_gid_{{{ FILEGID }}}_{{{ loop.index0 }}}` in Jinja condiitonals so that the state is included in the test when `MISSING_FILE_PASS`.

This way, any collected object that is not compliant will still be checked against the state, and obviously fail.
If no object is collected the rule will pass.

yuumasato (changes requested):
The perfomance improvement we were looking for was not so much on the OVAL evaluations, but on the number of objects collected.

It is clear that the exclude filters are the way to do that, :)

Now, I'd prefer to keep the behaviour of `missing_file_pass` and `allow_stricter_permission` in the templates for now. If we want to remove them, we should do so in master branch first, this Pr should focus on fixing the performance bug.

shared/templates/file_permissions/oval.template
```diff
@@ -1,19 +1,3 @@
...
-  {{% set FILE_EXISTENCE = "at_least_one_exists" %}}
-{{%- else -%}}
-  {{%- if MISSING_FILE_PASS -%}}
-    {{# Any number of files can exist, from zero to any #}}
-    {{% set FILE_EXISTENCE = "any_exist" %}}
-  {{%- else -%}}
-    {{# All defined files must exist. When using regex, at least one file must match #}}
-    {{% set FILE_EXISTENCE = "all_exist" %}}
-  {{%- endif -%}}
-{{%- endif -%}}
```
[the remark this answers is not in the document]
marcusburghardt: @yuumasato , this behavior was not changed, but the logic is now only treated in the `template.py` instead of two places (OVAL and template.py) to be in alignment with the exclusion filters.

shared/templates/file_groupowner/oval.template
yuumasato: This doesn't work, :sweat_smile:

PR #8457, opened 2022-03-29
Improve pam_faillock macro remediation to fix wrong control

marcusburghardt:
Old versions of the pam_faillock remediation created a pam_faillock.so line
with the control "[default=die]", which is not the proper control. The
remediation was already updated for new implementations but the bash
remediation was not updating the control if pam_faillock.so was once
enabled using the old remediation. This patch fix this.

### DIFF
--- a/shared/macros-bash.jinja
+++ b/shared/macros-bash.jinja
@@ -1026,6 +1026,7 @@ if ! grep -qE '^\s*auth\s+required\s+pam_faillock\.so\s+(preauth silent|authfail
     sed -i --follow-symlinks '/^auth.*sufficient.*pam_unix.so.*/a auth        required      pam_faillock.so authfail' "{{{ pam_file }}}"
     sed -i --follow-symlinks '/^account.*required.*pam_unix.so.*/i account     required      pam_faillock.so' "{{{ pam_file }}}"
 fi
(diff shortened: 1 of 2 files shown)


shared/macros-bash.jinja
```diff
@@ -1026,6 +1026,7 @@ if ! grep -qE '^\s*auth\s+required\s+pam_faillock\.so\s+(preauth silent|authfail
     sed -i --follow-symlinks '/^auth.*sufficient.*pam_unix.so.*/a auth        required      pam_faillock.so authfail' "{{{ pam_file }}}"
     sed -i --follow-symlinks '/^account.*required.*pam_unix.so.*/i account     required      pam_faillock.so' "{{{ pam_file }}}"
 fi
+sed -Ei 's/(auth.*)(\[default=die\])(.*pam_faillock.so)/\1required     \3/g'
```
jan-cerny, line 1029: missing file path of the file to be modified

jan-cerny (changes requested):

shared/macros-bash.jinja
marcusburghardt: Thanks for catching this.

mildas (approved):
Tested on RHEL7 by remediating with old SSG and then remediating with new SSG. PAM rules are fixed now.

jan-cerny (approved):

PR #8458, opened 2022-03-30
Fix same shadow field bug in tests

cipherboy:
My fix from #8398 was incomplete as I hadn't noticed the same bug in the
SSG test suite. This is now fixed.

`Signed-off-by: Alexander Scheel <alexander.m.scheel@gmail.com>`


Mab879 (approved):
Thanks for the PR!

PR #8461, opened 2022-03-30
OCP4: Remove oval check from file_groupowner_proxy_kubeconfig and fil…

Vincent056:
We need to remove the oval check from `file_groupowner_proxy_kubeconfig` and `file_owner_proxy_kubeconfig` because it was meant to check a file on the node instead of inside a pod. The oval check won't work on checking file permission inside a pod.


jhrozek (approved):
/lgtm

PR #8463, opened 2022-03-31
Use template generated content instead of custom artifacts for rule `dir_permissions_library_dirs`

ggbecker:
#### Description:
- Use templated generated content instead of custom artifacts.
  - The [`dir_permissions_library_dirs`](https://github.com/ComplianceAsCode/content/blob/9b7bd315b17036ff23591dd4368f506e7727038c/linux_os/guide/system/permissions/files/permissions_within_important_dirs/dir_permissions_library_dirs/rule.yml#L64) is already using the template thus it's
not necessary to keep the custom data.


yuumasato (approved):

yuumasato:
Thanks ,the templated content is in better shape.

PR #8469, opened 2022-03-31
Cleanups

maage:
#### Description:

Select only one shebang for python files. And maybe xml files should not be executables.

#### Rationale:

Reduce variations.

Shebang format could be any of the three, but I think only one variation should be used. Now counts are somewhat even. One of my bones env did not have python, so I selected next common one.

     15 #!/usr/bin/env python3
     20 #!/usr/bin/python3
     42 #!/usr/bin/env python

But in the end I'm totally fine with any of those.


jan-cerny (approved):
Thanks, yes, `#!/usr/bin/python3` is in line with the guidelines: https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_shebangs

PR #8474, opened 2022-04-01
Add STIG references during rules resolution

jan-cerny:
#### Description:
Add STIG references during rules resolution instead of adding them in a dedicated CMake command.

#### Rationale:
Prevents parsing and saving the whole XML tree of XCCDF just because of adding references.

### DIFF
--- a/ssg/build_stig.py
+++ b/ssg/build_stig.py
@@ -4,10 +4,10 @@ from __future__ import print_function
 import sys
 
 from .xml import ElementTree as ET
-from .constants import XCCDF11_NS, stig_ns, stig_refs
+from .constants import XCCDF11_NS
 
 
-def get_versions(reference_file_name):
+def map_versions_to_rule_ids(reference_file_name):
     try:
         reference_root = ET.parse(reference_file_name)
     except IOError:
--- a/ssg/build_yaml.py
+++ b/ssg/build_yaml.py
@@ -1890,7 +1890,7 @@ class BuildLoader(DirectoryLoader):
             self.sce_metadata = json.load(open(sce_metadata_path, 'r'))
         self.stig_references = None
         if stig_reference_path:
-            self.stig_references = ssg.build_stig.get_versions(stig_reference_path)
+            self.stig_references = ssg.build_stig.map_versions_to_rule_ids(stig_reference_path)
 
     def _process_values(self):
         for value_yaml in self.value_files:


ssg/build_stig.py
```diff
@@ -7,14 +7,7 @@
 from .constants import XCCDF11_NS, stig_ns, stig_refs
```
matejak, line 7:
```suggestion
from .constants import XCCDF11_NS
```

Those other symbols are not used.

ssg/build_stig.py
```diff
@@ -7,14 +7,7 @@
...
 
-def add_references(input_file_name, reference_file_name):
-    """
-    For a given input XCCDF file and DISA STIG reference XCCDF file, process
-    all STIG references in the rules from input XCCDF file and correctly link
-    them to the corresponding reference rule.
-
-    Returns the updated ElementTree containing updated reference elements.
-    """
+def get_versions(reference_file_name):
```
matejak, line 10: The name is misleading, the function returns a mapping of version strings (versions of what?) to rule IDs.

ssg/build_yaml.py
```diff
@@ -1374,6 +1376,15 @@ def normalize(self, product):
...
             raise RuntimeError(msg)
 
+    def add_stig_references(self, stig_references):
+        stig_id = self.references.get("stigid", None)
+        if not stig_id:
+            return
+        reference = stig_references.get(stig_id, None)
+        if not reference:
+            return
+        self.references["stigref"] = reference
```
matejak, line 1386: Question: Do you think that the `stigref` and `stigid` strings would be worth promoting to constants?

matejak (changes requested):
Thank you for the PR, I don't observe any regressions, and it achieves the current functionality in a more elegant way, which is great.

I have just easy-to-do style suggestions.

ssg/build_yaml.py
jan-cerny: I don't think so, they are different for each rule.

jan-cerny:
I have applied some style suggestions.

ssg/build_yaml.py
matejak: I meant the literal `stigref` string. But after discussing and thinking about it, it wouldn't bring significant added value to the project.

matejak:
After looking into the reference adding process, it seems to be a bit weird, but it's not because of this PR, but it got introduced a long ago.
So thank you for the improvement!

PR #8478, opened 2022-04-04
Define CPE and detection logic for OCP4 on HyperShift

tmishina:
#### Description:

- Add platform definition and detection logic for HyperShift management cluster

#### Rationale:

OpenShift clusters can be deployed as a hosted cluster, one of the variations of OpenShift installation, using [HyperShift](https://github.com/openshift/hypershift) technology.  HyperShift can create a hosted cluster not only on an OpenShift cluster but also on a vanilla Kubernetes cluster. If the underlying cluster is vanilla Kubernetes, Compliance Operator does not perform any OCP4-related check (because the underlying cluster does not contain OpenShift version resource `/apis/config.openshift.io/v1/clusteroperators/openshift-apiserver`) while it hosts OpenShift components such as API server, kubelet and Etcd inside of a hosted cluster namespace.

This PR provides a new platform definition called `ocp4-on-hypershift`. The trigger of the definition is existence of a custom resource of HyperShift `/kubernetes-api-resources/apis/apiextensions.k8s.io/v1/customresourcedefinitions/hostedclusters.hypershift.openshift.io`. When this CRD exists in the cluster, then the platform is marked as `ocp4-on-hypershift`, as an variation of OpenShift, and OCP4-related rules will be ignited.

### DIFF
--- a/products/ocp4/product.yml
+++ b/products/ocp4/product.yml
@@ -126,6 +126,10 @@ cpes:
       title: "Red Hat OpenShift Container Platform 4 on SDN"
       check_id: installed_app_is_ocp4_on_openshiftsdn
 
+  - ocp4-on-hypershift:
+      name: "cpe:/a:redhat:openshift_container_platform_on_hypershift:4"
+      title: "Red Hat OpenShift Container Platform 4 on HyperShift"
+      check_id: installed_app_is_ocp4_on_hypershift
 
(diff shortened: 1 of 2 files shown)


tmishina:
Dear reviewers, I'd appreciate it if you give comments or suggestions about how I can write the test for this kind of feature request. Thanks.

Vincent056 (approved):
I think you will also need to add this path to CO in cmd/manager/scap.go

PR #8484, opened 2022-04-04
Set Benchmark's resolved attribute to true

jan-cerny:
#### Description:
As we generate an XCCDF that is in a resolved form since its creation we should set the resolved element to true.

#### Rationale:

Produces warning `WARNING: Processing an unresolved XCCDF document. This may have unexpected results.`, but shouldn't have any impact.


Mab879 (approved):
Builds and SCAPVal passes.

PR #8490, opened 2022-04-04
Update chronyd_or_ntpd_set_maxpoll

Xeicker:
#### Description:
- Remove the possibility that the rule passes if there is no server/pool set
- Update rule description and tests accordingly
- Introduce service enabled criteria
#### Rationale:
- All STIG profiles (the only places where this rule appears) state that the system should be comparing internal information system clocks at least every N hours. And this wont happen if no server is set\
- The service enabled criteria was introduced to differentiate which systems need one conf or the other


vojtapolasek:
Hello, I am not sure if using the "service_chronyd_enabled" is a good approach. I would rather decide based on the fact if the package is installed. One reason is that remediations DO NOT start the service, therefore if the service is stopped, it will be never remediated correctly - the rule will report error instead of pass.

Xeicker:
Isn't it better to modify the remediation? Because if the service is not running then the system wouldn't be comparing internal  clocks with server information. am I wrong?

vojtapolasek:
I understand your point but I think the way to go is to achieve it by using rule service_chronyd_enabled or something like that. You can add the rule to the benchmark and you should achieve the same result. The reason is that we would like to stick to the "one rule - one configuration change" principle. What do you think?

Xeicker:
Well yes, you are right, I changed that

yuumasato (changes requested):
Nack, I think the package installed checks should not be done at the oval level, but with platforms. And the rule already has these platforms.

Rule `chronyd_or_ntpd_set_maxpoll` is about ensuring that any `server` or `pool` directive has the `maxpoll` option, so it should still continue passing if no server or pool entries exist.
To ensure the presence of some server or pool use any of the rules:
- chronyd_or_ntpd_specify_remote_server
- chronyd_specify_remote_server
- ntpd_specify_remote_server

vojtapolasek:
Hello, I was wondering (probably in the same way as you did) why the rule should pass when no server is set. Here is the explanation:
https://github.com/ComplianceAsCode/content/pull/8187
Read the discussion and tell me what you think.
I think you can reach the goal by using chronyd_or_ntpd_set_maxpoll together with chronyd_server_directive.
What do you think?

yuumasato:
After reading the PR, the rule you may actually be looking for is [chronyd_server_directive](https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/services/ntp/chronyd_server_directive/rule.yml)

vojtapolasek (approved):
Looks good now.

yuumasato (approved):

PR #8492, opened 2022-04-05
Add rules for kernel module security

yuumasato:
#### Description:

- Adds rules to check for the following kernel build configs:
  - CONFIG_STRICT_MODULE_RWX
  - CONFIG_MODULE_SIG
  - CONFIG_MODULE_SIG_FORCE
  - CONFIG_MODULE_SIG_ALL
  - CONFIG_MODULE_SIG_SHA512
  - CONFIG_MODULE_SIG_HASH
  - CONFIG_MODULE_SIG_KEY

- This also adds support for XCCDF variables for `kernel_build_config` template.
#### Rationale:

- These rules check if the kernel configurations regarding modules signature creation and verification.


yuumasato:
I'm thinking whether ` CONFIG_MODULE_SIG_HASH` and `CONFIG_MODULE_SIG_HASH` should use a variable. Thoughts?

ggbecker:
> I'm thinking whether ` CONFIG_MODULE_SIG_HASH` and `CONFIG_MODULE_SIG_HASH` should use a variable. Thoughts?

that's probably the right approach as crypto algorithms usually get obsolete over time.

linux_os/guide/system/kernel_build_config/kernel_config_module_sig_hash/rule.yml
```diff
@@ -0,0 +1,33 @@
...
+ocil_clause: 'the kernel was not built with the required value'
+
+ocil: |-
+    {{{ kernel_build_config_ocil("CONFIG_MODULE_SIG_HASH", "sha512") | indent(4) }}}
+
+template:
+    name: kernel_build_config
+    vars:
+        config: CONFIG_MODULE_SIG_HASH
+        value: 'sha512'
```
ggbecker, line 32:
```suggestion
        value: '"sha512"'
```

In the configuration file there are the double quotes, is it necessary? the same for the signing_key in the rule below

yuumasato: I don't think we need to add the value with double quotes, the rule only does checking and the OVAL accounts for the presence of double quotes.
But If we were remediating, I'd think we need those.

ggbecker: That's true, I forgot about the double quotes in the regex

docs/templates/template_reference.md
```diff
@@ -419,6 +419,7 @@ The only way to remediate is to recompile and reinstall the kernel, so no remedi
     - **value** - The value the configuration should have
       When **value** is `"n"`, the check will pass when the config is absent, commented out or has the
       value `n` in the `/boot/config-*` files.
+    - **variable** - The variable to get the value from.
```
ggbecker, line 422:
```suggestion
    - **variable** - The variable to get the value from. Cannot be used if value is set and vice-versa.
```
Please add something like this to explain that both parameter can't be used at the same time.

linux_os/guide/system/kernel_build_config/kernel_config_module_sig_force/rule.yml
```diff
@@ -0,0 +1,33 @@
+documentation_complete: true
+
+title: 'Require modules to be validly signed'
+
+description: |-
+    Reject unsigned modules or signed modules for which we don't have a key.
```
ggbecker, line 6:
```suggestion
    Reject unsigned modules or signed modules with an unknown key.
```

linux_os/guide/system/kernel_build_config/kernel_config_module_sig_force/rule.yml
```diff
@@ -0,0 +1,33 @@
...
+
+    {{{ kernel_build_config_describe_config("CONFIG_MODULE_SIG_FORCE", "y") | indent(4) }}}
+
+rationale: |-
+    Prevent loading modules that are unsigned or signed with a key that we don't know about.
```
ggbecker, line 11:
```suggestion
    Prevent loading modules that are unsigned or signed with an unknown key.
```

linux_os/guide/system/kernel_build_config/kernel_config_module_sig_hash/rule.yml
```diff
@@ -0,0 +1,33 @@
+documentation_complete: true
+
+title: 'Specify the hash to use when signing modules'
+
+description: |-
+    This configures the kernel to build and sign modules using SHA512 as the hash function.
```
ggbecker, line 6:
```suggestion
    This configures the kernel to build and sign modules using xccdf_value("var_kernel_config_module_sig_hash") as the hash function.
```
Something like this

linux_os/guide/system/kernel_build_config/var_kernel_config_module_sig_hash.var
```diff
@@ -0,0 +1,20 @@
...
+    The hash function to use when signing modules during kernel build process.
+
+type: string
+
+operator: equals
+
+interactive: false
+
+options:
+    default: sha256
```
ggbecker, line 15:
```suggestion
    default: sha512
```
is there a specific reason to choose the sha256 here?

linux_os/guide/system/kernel_build_config/var_kernel_config_module_sig_key.var
```diff
@@ -0,0 +1,16 @@
+documentation_complete: true
+
+title: 'Keys and certificates for kernel module signing'
```
ggbecker, line 3:
```suggestion
title: 'Key and certificate for kernel module signing'
```

linux_os/guide/system/kernel_build_config/var_kernel_config_module_sig_key.var
```diff
@@ -0,0 +1,16 @@
...
+
+description: |-
+    The keys and certificate to use when signing modules during kernel build process.
```
ggbecker, line 6:
```suggestion
    The private key and certificate to use when signing modules during kernel build process.
```

I'd mention something related to the PKCS#11 here as well from the rule description:
```
on systems where the OpenSSL ENGINE_pkcs11
    is functional — a PKCS#11 URI as defined by RFC7512. In the latter case, the PKCS#11 URI should
    reference both a certificate and a private key.
```

shared/templates/kernel_build_config/oval.template
```diff
@@ -17,29 +17,34 @@
...
+  comment="Check /boot/config-.* files for {{{ CONFIG }}}{{{ '=' + VALUE if VALUE else " according to " + VARIABLE }}}"
   id="test_kernel_{{{ CONFIG | lower }}}" version="1">
     <ind:object object_ref="object_kernel_{{{ CONFIG | lower }}}" />
+    <ind:state state_ref="state_kernel_{{{ CONFIG | lower }}}" />
   </ind:textfilecontent54_test>
 
   <ind:textfilecontent54_object id="object_kernel_{{{ CONFIG | lower }}}" version="1">
     <ind:filepath operation="pattern match">^/boot/config-.*$</ind:filepath>
-    <ind:pattern operation="pattern match">^{{{ CONFIG }}}="?{{{ VALUE }}}"?$</ind:pattern>
+    <ind:pattern operation="pattern match">^{{{ CONFIG }}}=(.*)$</ind:pattern>
```
ggbecker, line 28: I believe this will break the double quotes matching
```suggestion
    <ind:pattern operation="pattern match">^{{{ CONFIG }}}="?(.*)"?$</ind:pattern>
```

linux_os/guide/system/kernel_build_config/var_kernel_config_module_sig_hash.var
yuumasato: No special reason, RHEL7 and RHEL8 use `sha256` but RHEL9 uses `sha512`.
We can default to `sha512`, no problem.

shared/templates/kernel_build_config/oval.template
yuumasato: I've moved the double quote matching to https://github.com/ComplianceAsCode/content/pull/8492/files/da198a111e3ab8a168b091d6b9be3aabc543f139#diff-5e630ca5dad12e93ce7d1ed3511391ddb0e02ea9d95d9116b1c390746342d51cR38.
But I see now that the when the value comes from a string, we don't match the double quotes.

ggbecker (commented):
what about adding a test scenario to cover the double quotes regex, something that will echo to the file:

echo "{{{ CONFIG }}}=\\"{{{ VALUE }}}\\"" >> "$file"

yuumasato:
> what about adding a test scenario to cover the double quotes regex, something that will echo to the file:
> 
> echo "{{{ CONFIG }}}=\"{{{ VALUE }}}\"" >> "$file"

Test added in 5c45ea16c25d3459f544ffda2dcf73eff8f0f2c0

ggbecker (approved):

PR #8494, opened 2022-04-05
Update SRG-OS-000104-GPOS-00051 for RHEL9 STIG

Mab879:
#### Description:

- Add fix text to rules
- Add group_unique_id to RHEL 9 STIG

#### Rationale:

RHEL 9 STIG


jan-cerny (changes requested):
Rule gid_passwd_group_same doesn't have a fix.

jan-cerny (approved):

PR #8497, opened 2022-04-06
Update SRG-OS-000096-GPOS-00050 for RHEL 9 STIG 

jan-cerny:
#### Description:

add missing fix texts, missing Ansible, improve OCIL

#### Rationale:

RHEL 9 STIG


controls/stig_rhel9/SRG-OS-000096-GPOS-00050.yml
```diff
@@ -2,9 +2,10 @@ controls:
...
         levels:
             - medium
-        title: The operating system must be configured to prohibit or restrict the use of
-            functions, ports, protocols, and/or services, as defined in the PPSM CAL and vulnerability
-            assessments.
+        title: {{{ full_name }}} must be configured to prohibit or restrict the
+          use of functions, ports, protocols, and/or services, as defined in
+          the Ports, Protocols, and Services Management (PPSM) Category
+          Assignments List (CAL) and vulnerability assessments.
         rules:
```
jan-cerny, line 9: @Mab879 @ggbecker We have rule `configure_firewalld_ports` listed here. This rule looks like a generic rule about configuring the firewall, but in fact its OVAL check and bash remediation are specifically about opening SSHD ports.
Then, the rule description starts in a generic way but at the end it tells you to enable sshd port. Moreover, in our project there exists also rule `firewalld_sshd_port_enabled`.  So I don't know how to write the fix, because if I describe opening sshd ports then it won't be in line with OCIL. If I make everything in the rule aligned with the bash and OVAL checks, then I would basically create a rule identical with  `firewalld_sshd_port_enabled`. So should we keep the rule `configure_firewalld_ports` in this control file? And how should I change the OCIL, fix, and description?

Link to configure_firewalld_ports directory: https://github.com/ComplianceAsCode/content/tree/master/linux_os/guide/system/network/network-firewalld/ruleset_modifications/configure_firewalld_ports/oval

Mab879: I think we should keep it. In fact, it might be better if we can make it generic. Looking at the [RHEL 8 STIG](https://stigs.mab879.com/stigs/3/controls/RHEL-08-040030) it looks a generic "configure the firewall sanely" rule would fit nicely here.

Mab879: This might be best left to new rule, but I don't think there is a way to easy automate "The firewall should be configured sanely", so there might be no OVAL for this rule I'm thinking of.

jan-cerny: Yes, but the "new rule" already exists, it's the firewalld_sshd_port_enabled.

jan-cerny: To summarize:

* The requirement "The firewall should be configured sanely" isn't possible to be automated by SCAP because services and ports differ on different systems.
* At the same time there is rule "configure_firewalld_ports" that tries to implement it.
* The rule "configure_firewalld_ports" is internally inconsistent - its rationale and description are about generic firewall configuration, but OVAL and bash check and configure that sshd service is enabled in firewall.
* But we also have rule firewalld_sshd_port_enabled which checks  that sshd service is enabled in firewall.
* Our STIG workflow recommends us that OCIL and fix should be aligned with bash and OVAL which means if we create OCIL and fix for rule configure_firewalld_ports according to this rule's existing bash and OVAL the rule would be even more internally inconsistent - it would be basically a copy of firewalld_sshd_port_enabled with a misleading generic title.

So my proposal would be:
* remove OVAL and bash from configure_firewalld_ports, leaving it notchecked.
* add firewalld_sshd_port_enabled to this SRG

is that OK?

Mab879: Your proposal seems reasonable to me. That should help create more consistent experience with these rules.

Mab879:
@jan-cerny looks like this PR needs a rebase. Not sure why the boot hasn't told you about that.

jan-cerny:
I have add fixtext to configure_firewalld_ports

linux_os/guide/system/network/network-firewalld/ruleset_modifications/configure_firewalld_ports/rule.yml
```diff
@@ -63,3 +61,10 @@ ocil: |-
     Inspect the list of enabled firewall ports and verify they are configured
     correctly by running the following command:
     <pre>$ sudo firewall-cmd --list-all</pre>
+
+fixtext: |-
+    Configure {{{ full_name }}} to allow approved services to have access to the system.
+    To open a port for a service, configure firewalld using the following command:
+    $ sudo firewall-cmd --permanent --add-port=port_number/tcp
```
marcusburghardt, line 68: Is the standard to show commands with `sudo`? I saw cases with and without `sudo`. We should agree in a standard and use everywhere. What do you think?

marcusburghardt (changes requested):

linux_os/guide/system/network/network-firewalld/ruleset_modifications/configure_firewalld_ports/rule.yml
jan-cerny: I think that we should use sudo everywhere in OCILs and fixtexts. The RHEL 8 table that Becker posted https://complianceascode.github.io/content/tables/table-rhel8-stig-manual.html uses sudo almost everywhere,

marcusburghardt (approved):
LGTM

marcusburghardt:
@Mab879 , I didn't merge so you also have opportunity to update your last review. Thanks

Mab879 (approved):

Mab879:
Tests pass locally.

PR #8477, opened 2022-04-01
Kernel memory configs

yuumasato:
#### Description:

Adds rules to check for the following  kernel build configs:
- CONFIG_ARCH_HAS_STRICT_KERNEL_RWX=y
- CONFIG_DEBUG_WX=y
- CONFIG_DEBUG_FS=n
- CONFIG_STACKPROTECTOR=y
- CONFIG_STACKPROTECTOR_STRONG=y
- CONFIG_SCHED_STACK_END_CHECK=y
- CONFIG_HARDENED_USERCOPY=y
- #CONFIG_HARDENED_USERCOPY_FALLBACK is not set
- CONFIG_VMAP_STACK=y
- CONFIG_REFCOUNT_FULL=y
- CONFIG_FORTIFY_SOURCE = y
- #CONFIG_ACPI_CUSTOM_METHOD is not set
- #CONFIG_DEVKMEM is not set
- #CONFIG_PROC_KCORE is not set
- #CONFIG_COMPAT_VDSO is not set
- CONFIG_SECURITY_DMESG_RESTRICT=y
- CONFIG_RETPOLINE=y
- CONFIG_LEGACY_VSYSCALL_NONE=y
- CONFIG_LEGACY_VSYSCALL_EMULATE=n
- CONFIG_LEGACY_VSYSCALL_XONLY=n
- CONFIG_X86_VSYSCALL_EMULATION=n

PR #8504, opened 2022-04-07
Add rules for 64b kernel

yuumasato:
#### Description:

Add rules for 64b kernel
- CONFIG_DEFAULT_MMAP_MIN_ADDR=65536
- CONFIG_RANDOMIZE_BASE=y
- CONFIG_RANDOMIZE_MEMORY=y
- CONFIG_PAGE_TABLE_ISOLATION=y
- CONFIG_IA32_EMULATION=n
- CONFIG_MODIFY_LDT_SYSCALL=n


ggbecker (changes requested):
The PR is missing the rule for: CONFIG_LEGACY_VSYSCALL_NONE=y

and there is a problem with the template when using integer as input from the rule that covers: CONFIG_DEFAULT_MMAP_MIN_ADDR=65536

```
 {{{ oval_metadata("The kernel " + CONFIG + " should have value " + (VALUE if VALUE else "according to " + VARIABLE), affected_platforms=["multi_platform_all"]) }}}
TypeError: can only concatenate str (not "int") to str
```

yuumasato:
> The PR is missing the rule for: CONFIG_LEGACY_VSYSCALL_NONE=y

A rule for `CONFIG_LEGACY_VSYSCALL_NONE` already existed:
https://github.com/ComplianceAsCode/content/pull/8477/files#diff-1772ec1b30ff6f5083eb1f1b3f4f7711f94ae57c4c61830adde5c8a3f5f8a617

I forgot to remove it from commit and description. But they should be correct now.
 
> and there is a problem with the template when using integer as input from the rule that covers: CONFIG_DEFAULT_MMAP_MIN_ADDR=65536
> 
> ```
>  {{{ oval_metadata("The kernel " + CONFIG + " should have value " + (VALUE if VALUE else "according to " + VARIABLE), affected_platforms=["multi_platform_all"]) }}}
> TypeError: can only concatenate str (not "int") to str
> ```

I have wrapped the value in quotes so they are interpreted as strings.

ggbecker (approved):

PR #8516, opened 2022-04-08
Update SRG-OS-000125-GPOS-00065 for RHEL 9 STIG

jan-cerny:
#### Description:
* Use correct separator macro sshd_lineinfile_fix The sshd_config file doesn't use = as a separator, it uses
whitespace.
* Add fix text

#### Rationale:

RHEL 9 STIG


Mab879 (approved):
Waving SSGTS as the tests also fail on master.

Issue #8513, opened 2022-04-08
OCIL clause of package installed is inverted


ggbecker:
I believe OCIL questions should always be asked in a negative form, for example

`Is the case that `**the package is not installed**`?`

https://github.com/ComplianceAsCode/content/blob/3753283b2f00f3dcbf396bd5f66e7053e390dc12/shared/macros.jinja#L694

https://github.com/ComplianceAsCode/content/blob/3753283b2f00f3dcbf396bd5f66e7053e390dc12/shared/macros.jinja#L708

This leads to a wrong question to be rendered in the STIG Spreadsheet creation for example

```
Run the following command to determine if the  chrony  package is installed:
 $ rpm -q chrony 

If the package is installed, then this is a finding.
```

**Update1**: This macro is mostly used for package removed rules. But some rules that require installing the package are also using it and leads to wrong ocil data.

Mab879:
It should be a negative.  On a compliant system the OCIL question "Is it the case that" should be false statement.

ggbecker:
Fixed via #8523

PR #8523, opened 2022-04-08
Fix ocil_clause of package installed rules

ggbecker:
#### Description:

- Fix ocil_clause of package installed rules.
  - The complete entry would generate an ocil_clause for package removed
instead package installed.


linux_os/guide/system/accounts/accounts-pam/package_pam_pwquality_installed/rule.yml
```diff
@@ -28,9 +28,9 @@ references:
 ocil_clause: 'the package is not installed'
 
 {{% if 'ubuntu' not in product %}}
-{{{ complete_ocil_entry_package(package="libpwquality") }}}
+{{{ ocil_package(package="libpwquality") }}}
```
Mab879, line 31: It looks like you will need the `ocil` key here.

Mab879 (changes requested):

ggbecker:
I believe it's missing some singles quotes around some of these macros

linux_os/guide/system/accounts/accounts-pam/package_pam_pwquality_installed/rule.yml
ggbecker: that might be it

ggbecker: if it's inline then it would need the single quotes. But multiline doesn't require it

linux_os/guide/system/accounts/accounts-pam/package_pam_pwquality_installed/rule.yml
```diff
@@ -27,10 +27,11 @@ references:
 
 ocil_clause: 'the package is not installed'
 
+ocil: |-
 {{% if 'ubuntu' not in product %}}
```
Mab879, line 31:
```suggestion
{{%- if 'ubuntu' not in product %}}
```

To remove the blank at the start of the OCIL block.

Mab879 (changes requested):

linux_os/guide/system/accounts/accounts-pam/package_pam_pwquality_installed/rule.yml
ggbecker: fixed by rebase.

Mab879 (approved):

yuumasato:
Does it make sense to improve the macro to handle both cases based on a parameter?

ggbecker:
> 

yes. I'm going to create an issue to track this and this PR actually closes #8513

PR #8524, opened 2022-04-08
Check for a CNI that supports network policies

rhmdnd:
Before, configure_network_policies was a manual check that told the user
to make sure they were using a CNI that supported network policies. This
information is actually available through the K8S operator API.

Let's update the rule so that we query the appropriate resource and
filter it so we can check for CNIs that support network policies.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2072431

### DIFF
--- a/applications/openshift/networking/configure_network_policies/rule.yml
+++ b/applications/openshift/networking/configure_network_policies/rule.yml
@@ -24,9 +24,29 @@ ocil: |-
     <pre>$ oc explain networkpolicy</pre>
     The resulting output should be an explanation of the NetworkPolicy resource.
 
+{{% set api_path = '/apis/operator.openshift.io/v1/networks/cluster' %}}
+{{% set jqfilter = '[.spec.defaultNetwork.type]' %}}
+
 references:
     cis@ocp4: 5.3.1
     nerc-cip: CIP-003-8 R6,CIP-004-6 R3,CIP-007-3 R6.1
     nist: CM-6,CM-6(1)
     pcidss: Req-1.1.4,Req-1.2,Req-2.2
     srg: SRG-APP-000516-CTR-001325,SRG-APP-000516-CTR-001330,SRG-APP-000516-CTR-001335
+
+warnings:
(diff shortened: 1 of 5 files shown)


applications/openshift/networking/configure_network_policies/rule.yml
```diff
@@ -24,9 +24,29 @@ ocil: |-
...
+  name: yamlfile_value
+  vars:
+    ocp_data: "true"
+    filepath: |-
+      {{{ openshift_filtered_path(api_path, jqfilter) }}}
+    yamlpath: "[:]"
+    check_existence: "any_exist"
+    entity_check: "all"
+    values:
+      - value: "OpenShiftSDN"
```
jhrozek, line 51: Is there a reason not to add OVN from the start? Also, could we add the e2e test blob with the expected test result? (tests/ocp4/e2e.yml)?

jhrozek (commented):
One nit added, but otherwise looks good. We can't merge the rule until we make CO upstream release I guess..

applications/openshift/networking/configure_network_policies/rule.yml
rhmdnd: Good call, I added `OVN`. I also added three different unit tests exercises the various cases. Working on the e2e test now.

Vincent056:
```
Fetching URI: '/apis/operator.openshift.io/v1/networks/cluster'
debug: Encountered non-fatal error to be persisted in the scan: networks.operator.openshift.io "cluster" is forbidden: User "system:serviceaccount:openshift-compliance:api-resource-collector" cannot get resource "networks" in API group "operator.openshift.io" at the cluster scope
```
I think you might need to add that permission to the compliance operator

rhmdnd:
> I think you might need to add that permission to the compliance operator

Yeah - we just landed that, but we might need a new compliance-operator release.

https://github.com/openshift/compliance-operator/pull/815

Vincent056 (approved):

PR #8529, opened 2022-04-11
Organize fix text macros

### LINKED ISSUE #8520: Split the fix text macros into new file
Create a new macro file for the `fix` macros and move existing fix text macros to the file.

yuumasato:
#### Description:

- Move the `shared/macros-*.jinja` files into `shared/macros/` directory.
- Separate `fix` macros into their own `macros-fixtext.jinja` file.
- Standardize the `fix` macro naming.
- Rename `Rules`'s and `Controls`'s `fix` key to `fixtext` 
- Change build system to load any  `*.jinja` macro file in `shared/macros/`

- Fixes #8520


ssg/constants.py
```diff
@@ -60,19 +60,21 @@
...
-    __file__)), "shared", "macros-kubernetes.jinja")
+    __file__)), "shared", "macros", "macros-kubernetes.jinja")
 JINJA_MACROS_OVAL_DEFINITIONS = os.path.join(os.path.dirname(os.path.dirname(
-    __file__)), "shared", "macros-oval.jinja")
+    __file__)), "shared", "macros", "macros-oval.jinja")
 JINJA_MACROS_BASH_DEFINITIONS = os.path.join(os.path.dirname(os.path.dirname(
-    __file__)), "shared", "macros-bash.jinja")
+    __file__)), "shared", "macros", "macros-bash.jinja")
+JINJA_MACROS_FIXTEXT_DEFINITIONS = os.path.join(os.path.dirname(os.path.dirname(
+    __file__)), "shared", "macros", "macros-fixtext.jinja")
```
yuumasato, line 77: I wonder if we need to specify every jinja file we load.
We could simply load every `*.jinja` file from `shared/macros`, thoughts?

linux_os/guide/services/fapolicyd/service_fapolicyd_enabled/rule.yml
```diff
@@ -33,7 +33,7 @@ ocil: |-
     {{{ ocil_service_enabled(service="fapolicyd") }}}
 
 fix: |-
-    {{{ service_enabled_fix("fapolicyd") }}}
+    {{{ fix_service_enabled("fapolicyd") }}}
```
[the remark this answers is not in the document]
Mab879: At the moment I believe that correct. Moving to `fixtext` would be fine.

yuumasato:
@mildas The content test filtering is unhappy with the Jinja macro changes. :stuck_out_tongue:

ssg/constants.py
jan-cerny: that's  a great idea

yuumasato: Done in 2827f07f9c10126a90b9d5338b7ee2dbb3d0b8b5

linux_os/guide/services/fapolicyd/service_fapolicyd_enabled/rule.yml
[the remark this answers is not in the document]
yuumasato: Done in 4423a667299d97e0a186cdded8ca883cf7fcecca

yuumasato:
@Mab879 @ggbecker After I rebased to master after #8533, `utils/create_srg_export.py` started fail on missing `pandas` package. After installing `python3-pandas` the `fixtext`s started to appear correctly to me.
Please, double check if the export is working as expected.

ggbecker (changes requested):
Please update the control files to use `fixtext` instead of just `fix`

https://github.com/ComplianceAsCode/content/blob/master/ssg/controls.py#L103

ggbecker:
> @mildas The content test filtering is unhappy with the Jinja macro changes. stuck_out_tongue

I opened https://github.com/mildas/content-test-filtering/pull/31 to handle this issue when files are moved but no changes are made.

ggbecker (changes requested):
I have just noticed that the fact of moving macros around should be reflected in these files as well for readthedocs purposes: https://github.com/ComplianceAsCode/content/tree/master/docs/jinja_macros

Also, why not to remove the `macros-` prefix from the files since they are now in a folder called `macros.

ggbecker (approved):

yuumasato:
I have decided to drop the loading of any `*.jinja` macros from `shared/macros`. At the current state of the macro files there can be problems with the order of macros being loaded.

I'll propose a new PR with the loading and organization of the macros.

ggbecker (approved):

PR #8534, opened 2022-04-12
typo: kernel_config_slub_debug Enale -> Enable

maage:
#### Description:

Typo and change checked value to match what is wanted, aka CONFIG_SLUB_DEBUG should be enabled (y).

#### Rationale:

Typo and fix of logic.

### DIFF
--- a/linux_os/guide/system/kernel_build_config/kernel_config_slub_debug/rule.yml
+++ b/linux_os/guide/system/kernel_build_config/kernel_config_slub_debug/rule.yml
@@ -6,7 +6,7 @@ description: |-
     SLUB has extensive debug support features and this allows the allocator validation checking to
     be enabled.
 
-    {{{ kernel_build_config_describe_config("CONFIG_SLUB_DEBUG", "n") | indent(4) }}}
+    {{{ kernel_build_config_describe_config("CONFIG_SLUB_DEBUG", "y") | indent(4) }}}
 
 rationale: |-
     This activates the checking of the memory allocator structures and resets to zero the zones
@@ -25,11 +25,11 @@ identifiers:
 ocil_clause: 'the kernel was not built with the required value'
 
 ocil: |-
-    {{{ kernel_build_config_ocil("CONFIG_SLUB_DEBUG", "n") | indent(4) }}}
+    {{{ kernel_build_config_ocil("CONFIG_SLUB_DEBUG", "y") | indent(4) }}}
 
 template:
     name: kernel_build_config
     vars:
         config: CONFIG_SLUB_DEBUG
-        value: 'n'
+        value: 'y'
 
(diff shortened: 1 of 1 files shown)


ggbecker:
> Now that I check rule.yml, I am confused if this rule is actually about enable or disable. Rationale hints about activating, so this would be y, but value is n. Anyways title is either wrong or has typo.

it looks like it has wrong value, it should probably be `y`. @yuumasato FYI

yuumasato:
> > Now that I check rule.yml, I am confused if this rule is actually about enable or disable. Rationale hints about activating, so this would be y, but value is n. Anyways title is either wrong or has typo.
> 
> it looks like it has wrong value, it should probably be `y`. @yuumasato FYI

It should be enabled, so`y`.
@maage Could you also fix the value in the `rule.yml`? Much appreciated.

maage:
@yuumasato Now fixed in secondary commit, thanks.

ggbecker:
@maage Out of curiosity, how did you find the typo in this rule which later ended up in discovering a more critical issue? This rule is brand new and it's not being used in any of the available profiles yet.

maage:
@ggbecker I've build own rpm because I wanted to modify some rules under Fedora 35. For example some audit rules do not work used regexp parser has quirks. Some other rules are not flexible enough to make sense in my point of view, like xccdf_org.ssgproject.content_rule_accounts_polyinstantiated_tmp. I also wondered why some rules were made differently than others, xccdf_org.ssgproject.content_rule_sshd_rekey_limit vs xccdf_org.ssgproject.content_rule_sshd_set_login_grace_time and if it would be easy to fix (=not so). It evolved into situation where I have this "enable all sane rules" tailoring file. It was easy to enable all rules in "Kernel Configuration" when using scap-workbench (one click). I was then checking results to see what rules to drop and what to keep. I then noticed typo and wondered why this rule made no sense.

ggbecker:
> @ggbecker I've build own rpm because I wanted to modify some rules under Fedora 35. For example some audit rules do not work used regexp parser has quirks. Some other rules are not flexible enough to make sense in my point of view, like xccdf_org.ssgproject.content_rule_accounts_polyinstantiated_tmp. I also wondered why some rules were made differently than others, xccdf_org.ssgproject.content_rule_sshd_rekey_limit vs xccdf_org.ssgproject.content_rule_sshd_set_login_grace_time and if it would be easy to fix (=not so). It evolved into situation where I have this "enable all sane rules" tailoring file. It was easy to enable all rules in "Kernel Configuration" when using scap-workbench (one click). I was then checking results to see what rules to drop and what to keep. I then noticed typo and wondered why this rule made no sense.

If you think some rules can be improved, feel free to open more pull requests even if you are not 100% sure about it. We can then discuss better when looking directly at the code. Thanks for the explanation, it's good to understand how people use the project.

PR #8535, opened 2022-04-12
Update SRG-OS-000134-GPOS-00068 for RHEL 9 STIG

jan-cerny:
#### Description:
* adds fix elements
* add rules selinux_state and package_policycoreutils installed

#### Rationale:

RHEL 9 STIG


jan-cerny:
Question: RHEL 8 STIG has in addition SELinux enforcing mode and policycoreutils under this SRG. Why RHEL 9 control file doesn't have these 2 items as well?

jan-cerny:
Question: RHEL 8 STIG has in addition SELinux enforcing mode and policycoreutils under this SRG. Why RHEL 9 control file doesn't have these 2 items as well?

jan-cerny:
Question: RHEL 8 STIG has in addition SELinux enforcing mode and policycoreutils under this SRG. Why RHEL 9 control file doesn't have these 2 items as well?

Mab879 (approved):

PR #8539, opened 2022-04-13
Add json output function to controleval script

Vincent056:
In order to get data for the control dashboard, this pr adds a function for `controleval.py` script to output control stats to a JSON file.

### DIFF
--- a/utils/controleval.py
+++ b/utils/controleval.py
@@ -2,6 +2,7 @@
 import collections
 import argparse
 import os
+import json
 
 # NOTE: This is not to be confused with the https://pypi.org/project/ssg/
 # package. The ssg package we're referencing here is actually a relative import
@@ -82,19 +83,34 @@ def calculate_stats(ctrls):
     print_specific_stat("Applicable", applicable, total)
     print_specific_stat("Assessed", assessed, applicable)
     print()
-    print_specific_stat("Automated", ctrlstats[controls.Status.AUTOMATED], applicable)
-    print_specific_stat("Manual", ctrlstats[controls.Status.MANUAL], applicable)
-    print_specific_stat("Supported", ctrlstats[controls.Status.SUPPORTED], applicable)
-    print_specific_stat("Documentation", ctrlstats[controls.Status.DOCUMENTATION], applicable)
-    print_specific_stat("Inherently Met", ctrlstats[controls.Status.INHERENTLY_MET], applicable)
-    print_specific_stat("Does Not Meet", ctrlstats[controls.Status.DOES_NOT_MEET], applicable)
-    print_specific_stat("Partial", ctrlstats[controls.Status.PARTIAL], applicable)
+    print_specific_stat("Automated",
+                        ctrlstats[controls.Status.AUTOMATED],
+                        applicable)
+    print_specific_stat("Manual",
+                        ctrlstats[controls.Status.MANUAL],
+                        applicable)
+    print_specific_stat("Supported",
+                        ctrlstats[controls.Status.SUPPORTED],
+                        applicable)
+    print_specific_stat("Documentation",
+                        ctrlstats[controls.Status.DOCUMENTATION],
+                        applicable)
+    print_specific_stat("Inherently Met",
+                        ctrlstats[controls.Status.INHERENTLY_MET],
+                        applicable)
+    print_specific_stat("Does Not Meet",
+                        ctrlstats[controls.Status.DOES_NOT_MEET],
+                        applicable)
+    print_specific_stat("Partial",
+                        ctrlstats[controls.Status.PARTIAL],
+                        applicable)
 
     applicablelist = ctrls - ctrllist[controls.Status.NOT_APPLICABLE]
     assessedlist = set().union(ctrllist[controls.Status.AUTOMATED]).union(ctrllist[controls.Status.SUPPORTED])\
         .union(ctrllist[controls.Status.DOCUMENTATION]).union(ctrllist[controls.Status.INHERENTLY_MET])\
         .union(ctrllist[controls.Status.PARTIAL]).union(ctrllist[controls.Status.MANUAL])
-    print("Missing:", ", ".join(sorted(str(c.id) for c in applicablelist - assessedlist)))
+    print("Missing:", ", ".join(sorted(str(c.id)
+          for c in applicablelist - assessedlist)))
 
 
 def print_specific_stat(stat, current, total):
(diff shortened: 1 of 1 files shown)


utils/controleval.py
```diff
@@ -108,14 +130,71 @@ def print_specific_stat(stat, current, total):
...
+        ctrllist[controls.Status.SUPPORTED])
+    data["addressed_controls"]["automated"] = get_id_array(
+        ctrllist[controls.Status.AUTOMATED])
+    data["addressed_controls"]["notapplicable"] = get_id_array(
+        ctrllist[controls.Status.NOT_APPLICABLE])
+    data["addressed_controls"]["partial"] = get_id_array(
+        ctrllist[controls.Status.PARTIAL])
+    data["addressed_controls"]["pending"] = get_id_array(
+        ctrllist[controls.Status.PENDING])
+    print(data)
```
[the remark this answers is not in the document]
Vincent056: thanks for the review! that's a good catch, just fixed it

utils/controleval.py
```diff
@@ -82,19 +89,34 @@ def calculate_stats(ctrls):
...
     applicablelist = ctrls - ctrllist[controls.Status.NOT_APPLICABLE]
     assessedlist = set().union(ctrllist[controls.Status.AUTOMATED]).union(ctrllist[controls.Status.SUPPORTED])\
         .union(ctrllist[controls.Status.DOCUMENTATION]).union(ctrllist[controls.Status.INHERENTLY_MET])\
         .union(ctrllist[controls.Status.PARTIAL]).union(ctrllist[controls.Status.MANUAL])
-    print("Missing:", ", ".join(sorted(str(c.id) for c in applicablelist - assessedlist)))
+    print("Missing:", ", ".join(sorted(str(c.id)
+          for c in applicablelist - assessedlist)))
 
 
 def print_specific_stat(stat, current, total):
```
rhmdnd, line 122: Should this be included in the JSON formatted output?

utils/controleval.py
```diff
@@ -108,14 +130,72 @@ def print_specific_stat(stat, current, total):
...
+    else:
+        calculate_stats(ctrls)
+
+
+def print_to_json(ctrls, product, id, level, output_file):
+    total = len(ctrls)
+    data = dict()
+    ctrllist = collections.defaultdict(set)
+
+    if total == 0:
```
rhmdnd, line 149: This doesn't seem JSON-specific, perhaps we move it up to the stats() method?

utils/controleval.py
rhmdnd: Do we want to add the percentages to the JSON representation? Or is the thing consuming the JSON going to do that if needed?

utils/controleval.py
Vincent056: thanks that makes sense

utils/controleval.py
Vincent056: The percentage will be calculated at the other end

utils/controleval.py
```diff
@@ -138,6 +219,11 @@ def create_parser():
         required=True)
     statsparser.add_argument(
         "-p", "--product", help="product to check has required references")
+    statsparser.add_argument(
+        "-o",
+        "--output_json_file",
+        type=argparse.FileType('w'),
+        help="json file output")
```
rhmdnd, line 226: By doing this for JSON and not for the non-JSON path, are we setting the precedence that a user should be able to write any output from this tool to a file?

Vincent056: that's a good point, how does it look now?

rhmdnd: Today - I don't think this script supports writing to a file natively (but you can surely redirect it as the caller)

Vincent056: Sorry, it might be a stupid question, could you elaborate on that?

rhmdnd: Not at all - I should have been more explicit in my initial comment.

What I meant here was to leave out the support for writing to a file and print the JSON output to stdout, just like how we print the report to today.

That would allow someone to redirect the output to a file if they want to persist it:

```
$ controleval.py ... -o json | report.json
```

Or, a caller can pipe it into some other useful utility, like `jq` if they want to query specific details of the summary:

```
$ controleval.py ... -o json | jq .
```

utils/controleval.py
```diff
@@ -108,14 +130,73 @@ def print_specific_stat(stat, current, total):
...
+        ctrllist[controls.Status.AUTOMATED])
+    data["addressed_controls"]["notapplicable"] = get_id_array(
+        ctrllist[controls.Status.NOT_APPLICABLE])
+    data["addressed_controls"]["partial"] = get_id_array(
+        ctrllist[controls.Status.PARTIAL])
+    data["addressed_controls"]["pending"] = get_id_array(
+        ctrllist[controls.Status.PENDING])
+    data["addressed_controls"]["notassessed"] = get_id_array(
+        applicablelist - assessedlist)
+    print(json.dumps(data, cls=setEncoder), file=output_file)
```
rhmdnd, line 188: You could remove the file support here and let print output to stdout.

utils/controleval.py
Vincent056: thanks for the explanation, that makes sense!

utils/controleval.py
```diff
@@ -33,6 +35,11 @@
 SSG_ROOT = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
 
 
+class setEncoder(json.JSONEncoder):
```
rhmdnd, line 38: I don't think this is necessary anymore? I was able to remove it and the stdout functionality worked fine.

utils/controleval.py
```diff
@@ -108,14 +130,73 @@ def print_specific_stat(stat, current, total):
...
+        ctrllist[controls.Status.AUTOMATED])
+    data["addressed_controls"]["notapplicable"] = get_id_array(
+        ctrllist[controls.Status.NOT_APPLICABLE])
+    data["addressed_controls"]["partial"] = get_id_array(
+        ctrllist[controls.Status.PARTIAL])
+    data["addressed_controls"]["pending"] = get_id_array(
+        ctrllist[controls.Status.PENDING])
+    data["addressed_controls"]["notassessed"] = get_id_array(
+        applicablelist - assessedlist)
+    print(json.dumps(data, cls=setEncoder))
```
rhmdnd, line 188: Same comment here, I was able to do the following and it worked:

```python
print(json.dumps(data))
```

utils/controleval.py
```diff
@@ -108,14 +130,73 @@ def print_specific_stat(stat, current, total):
...
     validate_args(ctrlmgr, args)
     ctrls = set(ctrlmgr.get_all_controls_of_level(args.id, args.level))
-    calculate_stats(ctrls)
+    total = len(ctrls)
+
+    if total == 0:
+        print("No controls founds with the given inputs. Maybe try another level.")
+        exit(1)
+
+    if args.output_format is not None:
```
rhmdnd, line 139: I think we can reduce lines 139 and 140 into a single if statement:

```python
if args.output_format == 'json':
```

utils/controleval.py
```diff
@@ -2,6 +2,8 @@
 import collections
 import argparse
 import os
+import json
+from re import A
```
rhmdnd, line 6: I think we can remove this. Doesn't look like it's used anywhere.

utils/controleval.py
```diff
@@ -138,6 +213,11 @@ def create_parser():
         required=True)
     statsparser.add_argument(
         "-p", "--product", help="product to check has required references")
+    statsparser.add_argument(
+        "-o",
+        "--output_format",
```
rhmdnd, line 218: I should have suggested this earlier, sorry for not catching it. But using `-` instead of `_` is probably more consistent here.

I think my original suggestion used the underscore instead of a dash.

Vincent056: I think that's a good suggestion, thanks for the throughout reviews!

rhmdnd (approved):
/lgtm thank you for sticking with me on all the comments and questions

PR #8541, opened 2022-04-13
utils/compare_results.py returns 1 if results are not the same

mildas:
#### Description:
Return 1 if something is missing or result is not the same.

#### Rationale:
For testing purposes, there should be easy way how to identify results are not the same.

### DIFF
--- a/utils/compare_results.py
+++ b/utils/compare_results.py
@@ -168,6 +168,7 @@ def process_stig_results(base_results: dict, target_results: dict,
 
 
 def do_compare(base_results: dict, target_results: dict) -> None:
+    return_code = 0
     same_status = list()
     missing_in_target = list()
     different_results = dict()
@@ -175,12 +176,15 @@ def do_compare(base_results: dict, target_results: dict) -> None:
         target_result = target_results.get(base_result_id)
         if not target_result:
             missing_in_target.append(base_result_id)
+            return_code = 1
             continue
         if base_result != target_result:
             different_results[base_result_id] = (base_result, target_result)
+            return_code = 1
         else:
             same_status.append(base_result_id)
     print_summary(base_results, different_results, missing_in_target, same_status)
+    return return_code
 
 
 def match_results(base_tree: ElementTree.ElementTree, target_tree: ElementTree.ElementTree):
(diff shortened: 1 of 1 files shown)


ggbecker (approved):

utils/compare_results.py
```diff
@@ -206,7 +209,7 @@ def main():
     check_file(args.target)
     base_tree = ssg.xml.open_xml(args.base)
     target_tree = ssg.xml.open_xml(args.target)
-    match_results(base_tree, target_tree)
+    return match_results(base_tree, target_tree)
```
ggbecker, line 212:
```suggestion
    return_code = match_results(base_tree, target_tree)
    exit(return_code)
```

ggbecker (changes requested):
actually, don't we need to use the sys.exit() function to return exit codes?

utils/compare_results.py
mildas: Ah, true :D

Mab879 (approved):

matejak (changes requested):
Would it be possible to refactor the `do_compare` function? It performs analysis (low-level), it serves a formatted user summary (high-level), and now it got a return code functionality, which is something that can be deduced from the analysis. Do you think that you could create a couple of small functions that adhere to the Clean Code style?

I am quite sure that the proposed code works, but at the same time it is small and simple enough to make it nice.

mildas:
@matejak During creation of data structure, some functions seemed as good candidate to be a method. The refactoring ended up a bit bigger than what you might meant. Check it out. This might a be a good "clean code training" for me.

utils/compare_results.py
```diff
@@ -46,6 +46,32 @@ def get_wining_status(cls, current_status: str, proposed: str) -> str:
         return proposed
 
 
+class Comparison:
+    same_status = list()
+    missing_in_target = list()
+    different_results = dict()
```
matejak, line 52: Those are declared as class variables - I suggest to move that under `__init__` in a similar way as e.g. `base_results`. The [dataclass](https://docs.python.org/3/library/dataclasses.html) syntax is similar, but data classes are a relatively new Python feature, so we can't use it yet.

utils/compare_results.py
```diff
@@ -46,6 +46,32 @@ def get_wining_status(cls, current_status: str, proposed: str) -> str:
...
+
+    def __init__(self, base_results, target_results):
+        self.base_results = base_results
+        self.target_results = target_results
+
+    def do_compare(self):
```
matejak, line 58: Do you think that it would make sense to declare this as private and also call it from init? There is not much of useful stuff one could do with a `Comparison` that is just initialized and not processed.

matejak:
I really like those changes. I just suggest to move variable initialization under init, and perhaps to do something else as well, but overall I think that the code is pretty clean now.

matejak (approved):

matejak:
Cool, thanks! @ggbecker I think that the `exit(...)` good enough to return a value.

ggbecker (approved):

PR #8543, opened 2022-04-13
rsyslog: Fix array creation when path has wildcard

lonicerae:
This patch fixes the issue that the array is expanded to wildcard path instead of its elements.

#### Description:

- A simple test case as follows:

    /etc/rsyslog.conf
    include(file="/etc/rsyslog.d/*.conf" mode="optional")

    /etc/rsyslog.d/custom1.conf
    local1.*    /tmp/local1.out

    /etc/rsyslog.d/custom2.conf
    local2.*    /tmp/local2.out

#### Rationale:

- With current code, the RSYSLOG_INCLUDE or RSYSLOG_INCLUDE_CONFIG will be read as a string '/etc/rsyslog.d/*.conf' in the `for` loop.
- This patch ensures the wildcard path is expanded to suitable files and then added to the array.

### DIFF
--- a/linux_os/guide/system/logging/ensure_rsyslog_log_file_configuration/rsyslog_files_permissions/bash/shared.sh
+++ b/linux_os/guide/system/logging/ensure_rsyslog_log_file_configuration/rsyslog_files_permissions/bash/shared.sh
@@ -5,8 +5,8 @@
 RSYSLOG_ETC_CONFIG="/etc/rsyslog.conf"
 # * And also the log file paths listed after rsyslog's $IncludeConfig directive
 #   (store the result into array for the case there's shell glob used as value of IncludeConfig)
-readarray -t RSYSLOG_INCLUDE_CONFIG < <(grep -e "\$IncludeConfig[[:space:]]\+[^[:space:];]\+" /etc/rsyslog.conf | cut -d ' ' -f 2)
-readarray -t RSYSLOG_INCLUDE < <(awk '/)/{f=0} /include\(/{f=1} f{nf=gensub("^(include\\(|\\s*)file=\"(\\S+)\".*","\\2",1); if($0!=nf){print nf}}' /etc/rsyslog.conf)
+readarray -t RSYSLOG_INCLUDE_CONFIG < <(printf '%s\n' $(grep -e "\$IncludeConfig[[:space:]]\+[^[:space:];]\+" /etc/rsyslog.conf | cut -d ' ' -f 2))
+readarray -t RSYSLOG_INCLUDE < <(printf '%s\n' $(awk '/)/{f=0} /include\(/{f=1} f{nf=gensub("^(include\\(|\\s*)file=\"(\\S+)\".*","\\2",1); if($0!=nf){print nf}}' /etc/rsyslog.conf))
 
 # Declare an array to hold the final list of different log file paths
 declare -a LOG_FILE_PATHS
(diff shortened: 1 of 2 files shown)


rmetrich (approved):
LGTM, tested on RHEL7 (old style $IncludeConfig directive) and Fedora 35 (new style include directive).

vojtapolasek (changes requested):
Hello, thank you very much for this bugfix. Could you please also add test scenario which uses the wildcard?
You can get inspired for example by:
https://github.com/ComplianceAsCode/content/blob/master/linux_os/guide/system/logging/ensure_rsyslog_log_file_configuration/rsyslog_files_permissions/tests/include_perms_0600_IncludeConfig_perms_0601.fail.sh

lonicerae:
@vojtapolasek Thank you! Just added sample configs in that file, please help to review it.

vojtapolasek:
Do you think you could create a separate test for this case? I am not sure if your modifications prove that it works, because you are including *.conf files which have been already includes just few lines above.

vojtapolasek:
Do you think you could create a separate test for this case? I am not sure if your modifications prove that it works, because you are including *.conf files which have been already includes just few lines above.

lonicerae:
@vojtapolasek Please help to review the test case as well. Thanks!

vojtapolasek:
Hello, is there anything I can help you with to get this moving?

vojtapolasek:
I believe the problem is fixed by this PR which got merged recently:
https://github.com/ComplianceAsCode/content/pull/8601
Therefore I am closing this PR. Feel free to reopen if you think that it is not fixing your situation.

lonicerae:
Hello. Sorry I was busy these days. Thanks for sharing the PR https://github.com/ComplianceAsCode/content/pull/8601. It seems that it always includes all config files under the directory `/etc/rsyslog.d/`, this brings in incorrect checkings if the directive `include(file="/etc/rsyslog.d/*.conf" mode="optional")` etc. is disabled or has other settings. So I think that PR does not following the logic of the original code.

In fact the `include()` directive could be much more complicated, my friend and me both think current code may not cover all cases. However, my current commit could at least handle the issue I listed and keep the correct logic.

I'll try to complete the test file. It doesn't matter whether my code can be accepted or not. I just do it for making things more clear. Thanks.

lonicerae:
Maybe the following example can help understanding the differences:

* 8543
```
# readarray -t NEW_INC < <(awk '/)/{f=0} /include\(/{f=1} f{nf=gensub("^(include\\(|\\s*)file=\"(\\S+)\".*","\\2",1); if($0!=nf){print nf}}' /etc/rsyslog.conf)

# echo "${NEW_INC[@]}"
/etc/rsyslog.d/*/*.conf

# readarray -t RSYSLOG_INCLUDE < <(for INCPATH in "${NEW_INC[@]}"; do eval printf '%s\\n' "${INCPATH}"; done)

# echo "${RSYSLOG_INCLUDE[@]}"
/etc/rsyslog.d/aaa/91-test-local-7.conf
```

* 8601
```
# readarray -t RSYSLOG_INCLUDE < <(awk '/)/{f=0} /include\(/{f=1} f{nf=gensub("^(include\\(|\\s*)file=\"(\\S+)\".*","\\2",1); if($0!=nf){print nf}}' /etc/rsyslog.conf)

# RSYSLOG_CONFIGS=()

# RSYSLOG_CONFIGS=("${RSYSLOG_INCLUDE[@]}")

# echo "${RSYSLOG_CONFIGS[@]}"
/etc/rsyslog.d/*/*.conf

# RSYSLOG_FILES=()

# for ENTRY in "${RSYSLOG_CONFIGS[@]}"; do mapfile -t FINDOUT < <(find "$(dirname "${ENTRY}")" -maxdepth 1 -name "$(basename "${ENTRY}")"); RSYSLOG_FILES+=("${FINDOUT[@]}"); done
find: ‘/etc/rsyslog.d/*’: No such file or directory

# echo "${RSYSLOG_FILES[@]}"

```

PR #8548, opened 2022-04-13
Update accounts_umask_etc_profile remediation regex to handle comments

brett060102:
The current regex in the bash and ansible remediation won't fire
if the umask line is commented out.

#### Description:

- Handle cases when umask line is commented out.

#### Rationale:

- Should remediate if umask line is missing or commented out.


vojtapolasek (changes requested):
Looks good. Could you please add a test with commented umask?

brett060102:
Test has been added.

vojtapolasek (approved):
LGTM, thank you

PR #8550, opened 2022-04-13
SUSE audit_rules_privileged_commands_kmod add bash remediation SLE12/15

brett060102:
Ansible remediations existed, but not bash.

#### Description:

- Bash remediations were missing.

#### Rationale:

- Ansible is not officially support on all sle instamces

### DIFF
--- a/linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_kmod/bash/sle12.sh
+++ b/linux_os/guide/system/auditing/auditd_configure_rules/audit_privileged_commands/audit_rules_privileged_commands_kmod/bash/sle12.sh
@@ -1,0 +1,5 @@
+# platform = multi_platform_sle
+
+# Perform the remediation for both possible tools: 'auditctl' and 'augenrules'
+{{{ bash_fix_audit_watch_rule("auditctl", "/usr/bin/kmod", "x", "modules") }}}
+{{{ bash_fix_audit_watch_rule("augenrules", "/usr/bin/kmod", "x", "modules") }}}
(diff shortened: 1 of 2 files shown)


Mab879 (changes requested):
You can combine these files to the `shared.sh` and just add the `# platform = multi_platform_sle` line to get the same effect

brett060102:
@Mab879 I thought that having as shared.yml and share.sh would supersede the template in the rule.yml. Are you saying that will only happen for sle, if that is the only item in the platform spec?

brett060102:
@Mab879  updated to combine ansible and basj sle15 and sle12 remediation files into singe shared. I am concerned about interaction with the template in the rule.yml.

But I can always revert if that is an issue.

Mab879:
> @Mab879 updated to combine ansible and basj sle15 and sle12 remediation files into singe shared. I am concerned about interaction with the template in the rule.yml.
> 
> But I can always revert if that is an issue.

Ah thanks for catching that, looks like a revert is necessary. 

I assume you know this but `multi_platform_sle` is in the template.

brett060102:
Change reverted.  Did know that sle is supported for the template, but it was not remediating correctly. I don't remember why.
I will come back to this when I have more time to track this down. Must likely as an effort to get rid of all of the sle12 and sl15 specific files.

brett060102:
@Mab879 This one is ready for re-review.

brett060102:
@Mab879 Can I please get a re-review on this one?
Thanks

Mab879 (approved):
Thanks for the PR. 

Waving the SSGTS failure due to no tests.

